Issue #3, opened 2018-10-21
Migrate from time to Chrono


jasonwilliams:
the time crate is deprecated and Chrono seems to be the replacement. We should move to this

**Time**
https://crates.io/crates/time
https://docs.rs/time/0.1.40/time/

**Chrono**
https://crates.io/crates/chrono
https://docs.rs/chrono/0.4.6/chrono/

**More Info**
https://github.com/rust-lang-deprecated/time/issues/136

JasonShin:
I don't know Rust but I'm keen on helping it out

jasonwilliams:
this has now been fixed thanks @AFRUITPIE !

PR #8, opened 2019-03-10
migrate from 'time' to 'chrono', fixes issue #3

AFRUITPIE:
This should have the same time formatting as before, let me know if you see anything that doesn't look quite right.


jasonwilliams:
Will try and take a look at this next week, thanks so much for the PR!

PR #27, opened 2019-06-21
Allow numbers without decimal digits to be followed by a dot

topaxi:
Decimal digits are optional in JavaScript, this means the following is valid:

```js
let foo = 10.;
```

Which makes it possible to make the following property accesses (which look quite awkward):

```js
let foostr = 10..toString();
```

I haven't made any tests for the second case as I'm not sure where and how I'd do it. Also, I have almost zero knowledge about lexers/parsers and the like, so if there is a better way to achieve this, let me know :)

(I also don't know if this is spec compliant 🤷‍♂️)

### DIFF
--- a/src/lib/syntax/lexer.rs
+++ b/src/lib/syntax/lexer.rs
@@ -391,14 +391,26 @@ impl<'a> Lexer<'a> {
                 }
                 _ if ch.is_digit(10) => {
                     let mut buf = ch.to_string();
-                    loop {
+                    'digitloop: loop {
                         // There might not be a next character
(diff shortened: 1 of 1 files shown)


jasonwilliams:
Hi @topaxi 

You may be interested to know we have a test suite for numbers:
https://github.com/jasonwilliams/boa/blob/master/src/lib/syntax/lexer.rs#L886-L901

One could be added with a decimal point on the end to see how the lexer reacts.

> I also don't know if this is spec compliant

Yep:
https://tc39.es/ecma262/#prod-StrUnsignedDecimalLiteral

Issue #35, opened 2019-06-28
Read from input instead of hardcoded to file


jasonwilliams:
Running boa will always read from the same file, at some point we will need to read from any file that is pased through.

Parsing arguments example:
https://doc.rust-lang.org/beta/rust-by-example/std_misc/arg/matching.html

https://github.com/jasonwilliams/boa/blob/master/src/bin/bin.rs#L6 can be changed to read from the command line argument (similar to NodeJS)

Maybe `tests/js/test.js` can be read as a fallback?

jasonwilliams:
This would really help with https://github.com/jasonwilliams/boa/issues/12

callumquick:
Gonna take this one since I think it will also help with testing or setting up better automated tests for the "methods of objects" we've been implementing (check behaviour is as expected with Javascript directly). Also, any closer we get to Test 262 the better.

PR #41, opened 2019-06-30
Add ability to specify a file to read from to the Boa binary (#35)

callumquick:
Adds argument parsing in the main function of bin.rs which enables specifying a CLI argument as the file to read from: if not argument is specified, it still defaults to tests/js/test.js. #35 

Also adds and prints usage information on incorrect CLI options (and rudimentary error reporting, for the case the file specified doesn't exists, by implict `Result<E, T>` conversion):

```
Usage:
boa [file.js]
    Interpret and execute file.js
    (if no file given, defaults to tests/js/test.js")
```

Where "boa" is the debug binary created by `cargo`: `target/debug/bin`.

### DIFF
--- a/src/bin/bin.rs
+++ b/src/bin/bin.rs
@@ -1,8 +1,37 @@
 extern crate boa;
 use boa::exec;
 use std::fs::read_to_string;
+use std::env;
+use std::process::exit;
 
-pub fn main() {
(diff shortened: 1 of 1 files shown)


jasonwilliams:
Wow that was quick! Will take a look

Issue #40, opened 2019-06-30
Tests for parser.rs


jasonwilliams:
Similar to the tests for the [lexer.rs](https://github.com/jasonwilliams/boa/blob/master/src/lib/syntax/lexer.rs#L570-L931) we would like tests for parser.rs 

## Creating test scenarios
The parser is used here: https://github.com/jasonwilliams/boa/blob/master/src/lib/lib.rs#L33 it accepts a vector of tokens. We then want to make sure the expr it returns matches what we want

### Generating tokens
I often write the JS i want to test here: https://github.com/jasonwilliams/boa/blob/master/tests/js/test.js
Then i can output the tokens, by adding dbg!(&tokens) here: https://github.com/jasonwilliams/boa/blob/master/src/lib/lib.rs#L31

You can now pass these into the parser and test the output.

jasonwilliams:
An alternative could be to test from js files and read from them to the parser, it would be easier for developers to follow than a list of hand crafted tokens

PR #42, opened 2019-06-30
Parser tests

Razican:
Hi!

This is my first contribution to the project, great work! I'm working on doing test for the parser (#40), and I've just created the first changes. Note that I ran `cargo fmt`, so there are minor format changes in other places.

I wanted to create the PR asap, so that you could review it and let me know if I'm following the proper procedure. Real changes are in the `src/lib/syntax/parser.rs` file, where I added a `tests` module. For now, it just contains tests for strings, arrays and declarations.

I found some errors in the parser, though:
 - Trailing commas in array assignments make the parser fail ([spec link][array_assignment]).
 - `const` declarations work with empty expressions when they shouldnt ([spec_link][empty_const]).

For this last error, I would suggest to change the definition of `syntax::ast::expr::ExprDef::ConstDeclExpr` to directly not allow an `Option` in the expression.

Let me know if you like the design, and I'll implement the tests for the rest of the expressions!

[array_assignment]: https://tc39.es/ecma262/#prod-ArrayAssignmentPattern
[empty_const]: https://tc39.es/ecma262/#sec-let-and-const-declarations-static-semantics-early-errors

### DIFF
--- a/src/lib/syntax/parser.rs
+++ b/src/lib/syntax/parser.rs
@@ -770,10 +770,6 @@ mod tests {
         let mut lexer = Lexer::new(js);
         lexer.lex().unwrap();
 
-        dbg!(Parser::new(lexer.tokens).parse_all().unwrap());
-        let mut lexer = Lexer::new(js);
-        lexer.lex().unwrap();
-
         assert_eq!(
             Parser::new(lexer.tokens).parse_all().unwrap(),
             Expr::new(ExprDef::BlockExpr(expr.into()))
@@ -990,23 +986,17 @@ mod tests {
 
     #[test]
     fn check_operations() {
-        use crate::syntax::ast::constant::Const;
+        use crate::syntax::ast::{constant::Const, op::BinOp};
 
-        fn create_bin_op(op: NumOp, exp1: Expr, exp2: Expr) -> Expr {
-            use crate::syntax::ast::op::BinOp;
-
-            Expr::new(ExprDef::BinOpExpr(
-                BinOp::Num(op),
-                Box::new(exp1),
-                Box::new(exp2),
-            ))
+        fn create_bin_op(op: BinOp, exp1: Expr, exp2: Expr) -> Expr {
+            Expr::new(ExprDef::BinOpExpr(op, Box::new(exp1), Box::new(exp2)))
         }
 
         // Check numeric operations
         check_parser(
             "a + b",
             &[create_bin_op(
-                NumOp::Add,
+                BinOp::Num(NumOp::Add),
                 Expr::new(ExprDef::LocalExpr(String::from("a"))),
                 Expr::new(ExprDef::LocalExpr(String::from("b"))),
             )],
@@ -1014,7 +1004,7 @@ mod tests {
         check_parser(
             "a+1",
             &[create_bin_op(
-                NumOp::Add,
+                BinOp::Num(NumOp::Add),
                 Expr::new(ExprDef::LocalExpr(String::from("a"))),
                 Expr::new(ExprDef::ConstExpr(Const::Num(1.0))),
             )],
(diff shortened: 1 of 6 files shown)


jasonwilliams:
That's aweome @Razican i will take a look, thanks

src/lib/syntax/parser.rs
```diff
@@ -757,3 +757,180 @@ impl Parser {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::syntax::{
+        ast::expr::{Expr, ExprDef},
+        lexer::Lexer,
+    };
+
+    fn check_parser(js: &str, expr: &[Expr]) {
```
jasonwilliams: I like this idea, makes the further tests much cleaner and easier to understand

jasonwilliams:
Looks like some contributors didn't have cargo fmt on when they PR'd, i wonder if there's a way to enforce this

Razican:
> Looks like some contributors didn't have cargo fmt on when they PR'd, i wonder if there's a way to enforce this

Actually there is!
I use rustfmt in CI for my other repos, such as [SUPER](https://github.com/SUPERAndroidAnalyzer/super)

I just add the rustup component in Travis CI and then check the format: [travis-helper.sh#L23](https://github.com/SUPERAndroidAnalyzer/super/blob/master/travis-helper.sh#L23)

I fixed the links to the docs in the PR, by the way, I forgot to add them.

jasonwilliams:
That's good to know!

jasonwilliams:
Yes i like the design, carry on, it looks good so far

jasonwilliams:
> I found some errors in the parser, though:
> - Trailing commas in array assignments make the parser fail ([spec link][array_assignment]).
> - `const` declarations work with empty expressions when they shouldnt ([spec_link][empty_const]).
>
>For this last error, I would suggest to change the definition of `syntax::ast::expr::ExprDef::ConstDeclExpr` to directly not allow an `Option` in the expression.

Issues can be raised for these

jasonwilliams:
It might be an idea to merge what you have so far to get Travis working properly again

Razican:
> It might be an idea to merge what you have so far to get Travis working properly again.
> Let me know if you're happy to do that

Sure, give me until the end of the day (CEST time) to push the last tests I did parsing numeric binary operations and we can merge, and I can keep working on it later.

By the way, is there an specification on how should the parser parse complex mathematical operations? For example:

`a + d*(b-3) + 1`

I see that it will generate something like this in the parser:

`a + (1 + (d * (b-3)))`

Is there any reason why it changes the order of the last addition?

jasonwilliams:
I think the order here might just be wrong: https://github.com/jasonwilliams/boa/blob/master/src/lib/syntax/parser.rs#L725-L729

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

Can't see anything in the spec

Razican:
I pushed another commit with more parser tests. I set up a small function to create binary operations and check numeric binary operators.

jasonwilliams:
latest commit is failing on some code formatting

Razican:
Fixed, it seems that when I rebased there was some formatting issue.

src/lib/syntax/parser.rs
```diff
@@ -757,3 +757,354 @@ impl Parser {
...
+        let mut lexer = Lexer::new(js);
+        lexer.lex().unwrap();
+
+        dbg!(Parser::new(lexer.tokens).parse_all().unwrap());
```
jasonwilliams: Is there a reason this is wrapped in `dbg!` ?

Razican: Sorry, it was left there from my debugging. It's removed now, and I added bitwise operators.

Razican:
Hi, I think this is ready for merge now.

PR #44, opened 2019-06-30
add fmt check in travis fixes #43

### LINKED ISSUE #43: enforce cargo fmt
It might be an idea to have this, so some PR's are formatting other PR's, it should make it easier for those reviewing.

https://github.com/rust-lang/rustfmt is here.

`cargo fmt` needs to be added to a commit hook. We can add something to CONTRIBUTING.md hinting that `rustfmt` needs to be installed by `rustup component add rustfmt`

### potential example?
https://github.com/rust-analyzer/rust-analyzer/pull/271

jasonwilliams:


jasonwilliams:
@Razican this seems to be enough to do the job

Razican:
> @Razican this seems to be enough to do the job

Yes, this should do it. In the future if tests become too long you can change the way it's done to improve efficiency. In SUPER we had to divide the tests to execute some parts in parallel, it was taking more than 2 hours to test everything.

Issue #36, opened 2019-06-29
Implement Array prototype methods


jasonwilliams:
Similar to https://github.com/boa-dev/boa/issues/13 but for the Array 

- [x] Array.of() (#1127)
- [x] Array.prototype.concat()
- [x] Array.prototype.copyWithin()
- [x] Array.prototype.entries()
- [x] Array.prototype.every()
- [x] Array.prototype.fill()
- [x] Array.prototype.filter()
- [x] Array.prototype.find()
- [x] Array.prototype.findIndex()
- [x] Array.prototype.flat()
- [x] Array.prototype.flatMap()
- [x] Array.prototype.forEach()
- [x] Array.prototype.includes()
- [x] Array.prototype.indexOf()
- [x] Array.prototype.join()
- [x] Array.prototype.keys()
- [x] Array.prototype.lastIndexOf()
- [x] Array.prototype.map()
- [x] Array.prototype.pop()
- [x] Array.prototype.push()
- [x] Array.prototype.reduce()
- [x] Array.prototype.reduceRight()
- [x] Array.prototype.reverse()
- [x] Array.prototype.shift()
- [x] Array.prototype.slice()
- [x] Array.prototype.some()
- [x] Array.prototype.sort()
- [x] Array.prototype.splice()
- [ ] Array.prototype.toLocaleString()
- [x] Array.prototype.toString()
- [x] Array.prototype.unshift()
- [x] Array.prototype.values()
- [x] Array.prototype[@@iterator]()

Implementation happens here: https://github.com/boa-dev/boa/blob/master/boa/src/builtins/array/mod.rs

callumquick:
@jasonwilliams For the array, what is a Javascript array represented as in Boa? From my investigation appears to be as an Object, so in the underlying Rust a hashmap mapping from index to value: is that accurate? 

Does that mean in Boa there's no Javascript difference between arrays and objects? Is that intentional or just because Array isn't implemented yet? Should we have an "Array" entry in the ValueData enum?

jasonwilliams:
That's a very good point, and yes, what you said is accurate. There is no Array primitive value in JavaScript, so we can't have an array datatype in the value enum.  (for now) An Array is just an Object with a prototype of Array methods.

V8 do some optimisations with arrays with a backing store: https://v8.dev/blog/fast-properties
I'm not sure we need to worry about fast property access just yet.

For now we just need to create a global constructor (Array), similar to what we do with strings.
[...]

PR #51, opened 2019-07-01
Implement bare Array object with prototype field

callumquick:
This PR implements the basis of the Array object (and its internal representation) similarly to String. This is part of the work towards completion of Array prototype methods #36. 

It uses an internal Object to store the mapping between indices and Values. This may need review (does this still work with Array indexing etc), and otherwise an alternative approach will need to be developed through this PR.


callumquick:
This commit will likely clash, since formatting changes are needed to get Travis CI check of cargo fmt to pass from now on.

PR #52, opened 2019-07-01
Implement Array.prototype: concat(), push(), pop() and join()

callumquick:
This implementation of concat() introduces a helper function for further Array prototype work: it is usually desirable to take all input Arrays and construct and entirely new Array from some combination/altering of their values.

Starts the work of implementing Array.prototype (#36).

Further useful helper functions may convert from Array objects to underlying HashMap or even to Vec directly (by taking all indexed values).

Hopefully this implementation of concat() allows other people to start working on the Array prototype.

Sorry for duplicated commits/messy history, not sure how it got like that.

### DIFF
--- a/src/lib/js/array.rs
+++ b/src/lib/js/array.rs
@@ -28,18 +28,14 @@ pub fn make_array(this: Value, _: Value, args: Vec<Value>) -> ResultValue {
     // between indices and values): this creates an Object with no prototype
     this.set_field_slice("length", to_value(0i32));
     match args.len() {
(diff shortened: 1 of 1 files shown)


callumquick:
Added implementations for Array push(), pop() and join(): introduces simple new helper function for adding to arrays in case it is useful for implementing other methods

Common pattern is emerging for methods that return Arrays or iterates parts of them: may wish to get a helper that takes an array and a slice description and returns the equivalent Rust Vec for manipulation

PR #53, opened 2019-07-02
Use 'dyn' since trait objects without an explicit 'dyn' are deprecated

Atul9:


jasonwilliams:
Thanks @Atul9 
btw we now have a discord channel,  https://discord.gg/tUFFk9Y

PR #55, opened 2019-07-03
Boxing fixes #29

### LINKED ISSUE #29: Automatic Boxing and Unboxing of Primitive Values
As spotted in https://github.com/jasonwilliams/boa/issues/13#issuecomment-504763415 string methods on string literals won't work because automatic Boxing and Unboxing is not supported in Boa yet. This is something we'll need to think about adding, it could be worth seeing how spidermonkey or V8 do this

jasonwilliams:
Allows Boxing and Unboxing of primitives values when accessing methods


jasonwilliams:
Closing, methods need to be changed to accept primitive values instead of promoting to objects at this point

Issue #13, opened 2019-04-04
Implement String prototype methods


jasonwilliams:
## String prototype methods
These methods would be added to the String prototype object which is implemented here:
https://github.com/boa-dev/boa/blob/master/boa/src/builtins/string/mod.rs

Spec: https://tc39.es/ecma262/#sec-properties-of-the-string-prototype-object

**Implementation Example?**
See [charAt()](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/string/mod.rs#L77-L119) and charCodeAt() for a good example

- [x] charAt()
- [x] codePointAt()
- [x] charCodeAt()
- [x] concat() 
- [x] endsWith()
- [x] includes()
- [x] indexOf()
- [x] lastIndexOf()
- [ ] localeCompare()
- [x] match()
- [x] matchAll()
- [x] normalize()
- [x] padEnd()
- [x] padStart()
- [x] repeat()
- [x] replace()
- [x] replaceAll()
- [x] search()
- [x] slice()
- [x] split()
- [x] startsWith()
- [x] substring()
- [ ] toLocaleLowerCase()
- [ ] toLocaleUpperCase()
- [x] toLowerCase()
- [x] toString()
- [x] toUpperCase()
- [x] trim()
- [x] trimEnd()
- [x] trimStart()
- [x] toString()
- [x] valueOf()
- [x] [@@iterator]()

callumquick:
I have started to write some implementations for concat() and repeat() but have encountered an issue: currently none of the String prototype methods work, even existing charAt and charCodeAt:

Using:
```
let a = "Hello world".chatAt(2);
console.log(a);
```

Results in the buffer holding only: "Error: undefined"

Is this a known issue or needs to be resolved before prototype methods can be added?

jasonwilliams:
Hey @callumquick hmm that’s interesting, I will take a look, I wonder if some of the changes have caused this to no longer work.

callumquick:
Thanks @jasonwilliams, generating the string via `new String(..)` is working with charAt(), charCodeAt(). Will now test with my new implementations.

callumquick:
On the topic of the boxing of the primitive string, the only current issue with concat(), repeat(), slice() and the other methods that should return strings is that through `to_value()` they come out as "string literals" again. This means you can't do the following:
```
let a = new String("Hello world");
a.concat("! ").repeat(2);
```
[...]

PR #56, opened 2019-07-06
Implement String.prototype.{padStart(), padEnd()}

callumquick:
Addition of new String.prototype methods padStart() and padEnd():
https://tc39.es/ecma262/#sec-string.prototype.padend
https://tc39.es/ecma262/#sec-string.prototype.padstart

Completes more of the outstanding tasks from #13.


jasonwilliams:
You're on fire!

PR #58, opened 2019-07-07
Parse array elisions and trailing commas

### LINKED ISSUE #45: Trailing commas in array assignments make the parser fail
### Specification Link
https://tc39.es/ecma262/#prod-ArrayAssignmentPattern

### Example JS
```js
let foo = [1, 2, 3,];
```

### Expected
Parser should ignore the trailing comma

### Related to
https://github.com/jasonwilliams/boa/pull/42/files#diff-9f9158af1c9e00d4a93319c5918d9b97R816

### Contributing

zgotsch:
Hey Jason, just saw your JSConf EU talk, it was great!

I have added support for parsing arrays which include elisions as well as trailing commas. The elision is parsed into an "undefined", but I'm not sure if that's semantically the best thing to do. Perhaps a new expression should be used to indicate that an element has been elided or the metadata should be stored on the array node. Fixes #45.

### DIFF
--- a/src/lib/syntax/parser.rs
+++ b/src/lib/syntax/parser.rs
@@ -407,41 +407,40 @@ impl Parser {
                 }
             }
             TokenData::Punctuator(Punctuator::OpenBracket) => {
-                let mut array: Vec<Expr> = Vec::new();
-                let mut expect_comma_or_end = self.get_token(self.pos)?.data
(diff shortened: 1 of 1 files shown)


Razican:
Hi!
It seems that the pull request is failing due to a formatting issue. Could you run `cargo fmt`?

jasonwilliams:
Great work @zgotsch !

PR #59, opened 2019-07-07
Fixed a bunch of Clippy issues

Razican:
Hi!
I ran Clippy and I saw a lot of warnings and errors. I fixed many of them, but there are still some concerns, especially with casting and with some handling of `Value`, that being a Gc object poses some issues with clones.

This also fixes one bug, previously, this code would print "b5":
```javascript
let a = "b";
let c = 5;

c + a;
```

This has been fixed.

It also improves the performance by reducig the amount of heap used, and the many clones around.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ rust:
   - beta
 cache: cargo
 before_script:
-  - rustup component add rustfmt
+  - rustup component add rustfmt clippy
 script:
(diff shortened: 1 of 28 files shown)


Razican:
I added the CI checks for Clippy and merged the latest master changes. As you will see in the CI, there will be a ton of Clippy warnings to fix still.

PR #60, opened 2019-07-07
Constants must have a value on declaration

### LINKED ISSUE #46: empty `const` declaration should fail
## Specification:
https://tc39.es/ecma262/#sec-let-and-const-declarations-static-semantics-early-errors
_It is a Syntax Error if Initializer is not present and IsConstantDeclaration of the LexicalDeclaration containing this LexicalBinding is true._

### Expected
```js
const a = 5, b, c = 6;"
```
Should fail, binding `b` is not initialised

### Related to:
https://github.com/jasonwilliams/boa/pull/42/files#diff-9f9158af1c9e00d4a93319c5918d9b97R924

### Parser const decl location
https://github.com/jasonwilliams/boa/blob/master/src/lib/syntax/parser.rs#L79-L133 

### Contributing

Razican:
This fixes #46. I just modified constant declarations so that it's impossible to have a constant without a value. I also uncommented the test that was failing.

### DIFF
--- a/src/lib/syntax/ast/expr.rs
+++ b/src/lib/syntax/ast/expr.rs
@@ -32,7 +32,7 @@ pub enum ExprDef {
     /// Make a constant value
     ConstExpr(Const),
     /// Const declaration
-    ConstDeclExpr(Vec<(String, Option<Expr>)>),
+    ConstDeclExpr(Vec<(String, Expr)>),
     /// Construct an object from the function and arg{
     ConstructExpr(Box<Expr>, Vec<Expr>),
     /// Run several expressions from top-to-bottom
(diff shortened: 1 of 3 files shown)


src/lib/syntax/ast/expr.rs
```diff
@@ -200,6 +198,13 @@ impl Display for ExprDef {
                 }
                 f.write_str("")
             }
+            ExprDef::ConstDeclExpr(ref vars) => {
+                f.write_str("var ")?;
```
jasonwilliams: does this need to be `f.write_str("const ")?;` ?

Razican: Ups, yes, I copied it from above.

Razican:
I think this should be done now :)

PR #62, opened 2019-07-07
Clean dist

### LINKED ISSUE #48: dist folder should empty before webpack generates files
`yarn serve` or `yarn build` should clean out `dist` folder before running.

I believe this can be done by using https://github.com/johnagan/clean-webpack-plugin. I couldn't get it working last time i tried it.

zgotsch:
The easy thing worked for me.

Also removed /dist from the repo, since it's generated. Not sure about your thoughts on that.

Fixes #48


jasonwilliams:
Yeah its annoying but it needs to be there for the github pages thing to work  because its just a static site. I can probbaly host that bit myself and just pull every so often. I can accept this PR when ive done that

zgotsch:
Ah didn't realize there was a github pages deployment. Probably you could make a `gh-pages` branch and the only change there would be that `/dist` isn't gitignored. Then you get the added benefit of the ability to do a "deploy" to the demo, which is a rebase and build and push. Still a lot more work though.

jasonwilliams:
Yeah you're right, thanks for this.

PR #63, opened 2019-07-08
Unit testing for Array and String

GalAster:
- **Passed:**
- [x] Array.prototype.join()
- [x] Array.prototype.concat()
- [x] String.prototype.repeat()

- **Need fix:**
- [ ] String.prototype.length
- [ ] String.prototype.concat()

### DIFF
--- a/src/lib/js/array.rs
+++ b/src/lib/js/array.rs
@@ -183,21 +183,21 @@ mod tests {
         //TODO: array display formatter
         let mut engine = Executor::new();
         let init = r#"
-        let empty = [ ];
-        let one = [1];
+        let empty = new Array();
+        let one = new Array(1);
         "#;
         forward(&mut engine, init);
         // Empty ++ Empty
-        let ee = dbg!(forward(&mut engine, "empty.concat(empty)"));
+        let ee = forward(&mut engine, "empty.concat(empty)");
         //assert_eq!(ee, String::from(""));
         // Empty ++ NonEmpty
-        let en = dbg!(forward(&mut engine, "empty.concat(one)"));
+        let en = forward(&mut engine, "empty.concat(one)");
         //assert_eq!(en, String::from("a"));
         // NonEmpty ++ Empty
-        let ne = dbg!(forward(&mut engine, "one.concat(empty)"));
+        let ne = forward(&mut engine, "one.concat(empty)");
         //assert_eq!(ne, String::from("a.b.c"));
         // NonEmpty ++ NonEmpty
-        let nn = dbg!(forward(&mut engine, "one.concat(one)"));
+        let nn = forward(&mut engine, "one.concat(one)");
         //assert_eq!(nn, String::from("a.b.c"));
     }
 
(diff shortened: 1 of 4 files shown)


GalAster:
Bug: @jasonwilliams 

```ts
const s = new String('\ud834\udf06');
console.log(s.length)
```

Should be `2` but return `4`.

GalAster:
TODO: 

- [ ] Format Array display form.

I don't know how to do this.

GalAster:
Bug:

```ts
const hello = "Hello, ";
const world = "world! ";
hello.concat(world)
```

GalAster:
Please review: @jasonwilliams 

1. May conflict with #69 , I use a step by step vm in unit testing.

2. CI build failed because of `The command "cargo clippy --verbose" exited with 101.`

It is caused by the code in #56, I can't modify it.

3. I tried to fix String.prototype.length, but this problem is a bit complicated.

`this_str.len()` counts UTF8 length, I changed as `this_str.chars().count()` to count unicode surrogate pair length.

But js string length is UTF16/USC2 length, still different on some special cases.

src/lib/js/array.rs
```diff
@@ -167,7 +167,57 @@ pub fn _create(global: &Value) -> Value {
...
+
+    #[test]
+    fn concat() {
+        //TODO: array display formatter
+        let mut engine = Executor::new();
+        let init = r#"
+        let empty = [ ];
+        let one = [1];
+        "#;
+        forward(&mut engine, init);
```
jasonwilliams: Why `forward` as a name?

src/lib/lib.rs
```diff
@@ -44,22 +45,28 @@ extern "C" {
...
 
-pub fn exec(src: &str) -> String {
+fn parser_expr(src: &str) -> Expr {
     let mut lexer = Lexer::new(src);
     lexer.lex().unwrap();
     let tokens = lexer.tokens;
+    Parser::new(tokens).parse_all().unwrap()
+}
 
+pub fn forward(engine: &mut Interpreter, src: &str) -> String {
```
jasonwilliams: why `forward` as a name?

src/lib/js/string.rs
```diff
@@ -492,10 +493,111 @@ pub fn init(global: &Value) {
...
+    fn starts_with() {
+        let mut engine = Executor::new();
+        let init = r#"
+        const empty = new String('');
+        const en = new String('english');
+        const zh = new String('中文');
+        "#;
+        forward(&mut engine, init);
+        let pass = String::from("true");
+        assert_eq!(dbg!(forward(&mut engine, "empty.startsWith('')")), pass);
```
jasonwilliams: im assuming the `dbg!()` are just for you to test?

src/lib/lib.rs
jasonwilliams: Would also be worth some comments above this function definition to explain what its purpose is. (compared to exec)

src/lib/js/string.rs
```diff
@@ -492,10 +493,111 @@ pub fn init(global: &Value) {
...
+        assert_eq!(
+            dbg!(forward(&mut engine, "en.repeat(1)")),
+            String::from("english")
+        );
+        assert_eq!(
+            dbg!(forward(&mut engine, "zh.repeat(2)")),
+            String::from("中文中文")
+        );
+    }
+
```
jasonwilliams: Good tests!

src/lib/js/string.rs
GalAster: Well, not necessary indeed.

src/lib/lib.rs
GalAster: Let me see, it was call `run_frame` in RustPython.

I'm not sure what this should be called, `forward` sounds like walk a small step, which means the step by step interpretation.

jasonwilliams:
@GalAster you looking for me to merge this or you wanting to carry on?

GalAster:
@jasonwilliams It's ready to merge, I'll do the rest after Array format complete.

PR #67, opened 2019-07-08
Implementation of Array.prototype.{shift(), unshift(), reverse()}

callumquick:
The PR implements the following prototype methods for Array objects:

* Array.prototype.reverse ( ) - The elements of the array are rearranged so as to reverse their order. The object is returned as the result of the call.
* Array.prototype.shift ( ) - The first element of the array is removed from the array and returned.
* Array.prototype.unshift ( ...items ) - The arguments are prepended to the start of the array, such that their order within the array is the same as the order in which they appear in the argument list.

More items to tick off the list for #36 Implement Array prototype methods.
This PR may conflict with testing effort being made as part of #63 Unit testing for Array and String.

### DIFF
--- a/src/lib/js/array.rs
+++ b/src/lib/js/array.rs
@@ -158,8 +158,8 @@ pub fn reverse(this: Value, _: Value, _: Vec<Value>) -> ResultValue {
     for lower in 0..middle {
         let upper = len - lower - 1;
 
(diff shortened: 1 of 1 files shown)


jasonwilliams:
works well! really conforms to the spec also

PR #68, opened 2019-07-08
String and Array specific console.log formats

callumquick:
This PR creates specific formatting for Array and String objects, and lays the foundation for defining arbitrary console.log formats for other objects like Number/Boolean.

Defines a "type" field which must go in the "private object" of any JS native type (or any object which desires to have its type known internally to the engine), which console.log reads to determine how to display the object.

**Known caveat: objects parsed directly like arrays will not be recognised as an Array unless defined through explicity `new Array(...)` call. Suggestions welcome to solve this.**

Defines a new function which (potentially recursively) checks all objects given to console.log, and their sub-objects etc, and stringifies them for display. It gives String and Array special printing syntax (just the primitive value of the string for String, the [...] display syntax for Array).

For example, the following JS:
```js
console.log(new String("foo"));
console.log(new Array(0,1,2,3,4));
console.log({"foo": "bar", "arr": new Array(0,1)});
```
Will give:
```
foo
[0, 1, 2, 3, 4]
{foo: bar, arr: [0, 1]}
```
However following will not work as intended (however this is the current behaviour anyway):
```js
console.log([0,1,2,3,4]);
```
Gives the array as an object, because "type" private field is not set because array is parsed rather than explicitly constructed:
```
{0: 0, 1: 1, 2: 2, 3: 3, 4: 4, length: 5}
```


callumquick:
@jasonwilliams I have refactored the PR to use your new Object system: by doing this and setting the Object kind both on object constructor call and parsing (e.g. for ArrayDeclExpr), it now correctly console.logs Strings and Arrays.

E.g. following:
```js
console.log([0, 1, 2, "foo", new String("bar"), [0, 1], new Array(2, 3)]);
```
Now gives:
```
[0, 1, 2, foo, bar, [0, 1], [2, 3]]
```
Instead of object representation previously:
```
{0: 0, 1: 1, 2: 2, 3: "foo", 4: {__proto__ ...}, ...}
```

callumquick:
@jasonwilliams I've fixed the conflicts with this one: should be fine to merge for better console.log behaviour

PR #70, opened 2019-07-09
WIP: adding custom trace

jasonwilliams:
@Manishearth any idea why https://github.com/jasonwilliams/boa/blob/custom_trace/src/lib/js/function.rs#L63-L65 is failing?

Seems to work fine in these locations: 
https://github.com/search?l=Rust&q=custom_trace%21&type=Code

The error could be coming from wasm-bindgen
Failing here: https://travis-ci.com/jasonwilliams/boa/jobs/214417648#L566-L579
and here: https://travis-ci.com/jasonwilliams/boa/jobs/214417648#L680-L708

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,12 +4,12 @@
 name = "Boa"
 version = "0.2.1"
 dependencies = [
- "chrono 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


Manishearth:
You don't need to write it inside an impl block. THe macro generates the impl block.

PR #91, opened 2019-08-24
implements toLowerCase, toUpperCase, substring, substr and valueOf St…

arbroween:
…ring prototype methods

Hey there !
I implemented toLowerCase, toUpperCase, substring, substr and valueOf, to get more familiar with the code base as suggested on the project Discord. They are part of #13.

This is my first PR and also I am not an expert in Rust, but I tried to implement the algorithms in the tc39 specification as much as I could and I also followed the style of the already implemented String prototype methods.

I don't know if the implementation semantics are exactly the ones expected, but they should be close.

Any feedback is very welcomed!
Tell me if there is something I can improve


jasonwilliams:
Hey @arbroween looks good, are you able to rebase?

jasonwilliams:
ping @arbroween

PR #94, opened 2019-09-01
Add support for regex literals

999eagle:
This PR aims to add full support for regex literals.

Things still to do:
- [X] Implement regex literals in lexer.
- [X] Implement regex literals in parser.
- [X] Add global `RegExp` object.
- [ ] Implement regex-related functions on `String` and `RegExp`.
  - [X] `RegExp.test(str)`
  - [X] `RegExp.toString()`
  - [X] `RegExp.exec(str)`
  - [ ] `String.match(regexp)`
  - [ ] `String.matchAll(regexp)`
  - [ ] `String.replace(regexp|substr, newSubstr|function)`
  - [ ] `String.search(regexp)`

### DIFF
--- a/src/lib/syntax/ast/token.rs
+++ b/src/lib/syntax/ast/token.rs
@@ -59,7 +59,7 @@ pub enum TokenData {
     /// A string literal
     StringLiteral(String),
     /// A regular expression, consisting of body and flags
-    RegularExpression(String, String),
+    RegularExpressionLiteral(String, String),
     /// A comment
     Comment(String),
 }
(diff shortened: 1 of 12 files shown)


src/lib/syntax/lexer.rs
```diff
@@ -466,14 +467,58 @@ impl<'a> Lexer<'a> {
...
+                                                ch => body.push(ch),
+                                            }
+                                        }
+                                        Some(ch) => body.push(ch),
+                                    }
+                                }
+                                if regex {
+                                    // body was parsed, now look for flags
+                                    let flags = self.take_char_while(char::is_alphabetic)?;
+                                    self.push_token(TokenData::RegularExpression(body, flags));
```
jasonwilliams: I wonder if this should be called "RegularExpressionLiteral" to match up with the other token names?
The specification also refers to them as [RegularExpressionLiteral](https://tc39.es/ecma262/#prod-RegularExpressionLiteral)s too

src/lib/syntax/ast/token.rs
```diff
@@ -58,8 +58,8 @@ pub enum TokenData {
     Punctuator(Punctuator),
     /// A string literal
     StringLiteral(String),
-    /// A regular expression
-    RegularExpression(String),
+    /// A regular expression, consisting of body and flags
+    RegularExpression(String, String),
```
jasonwilliams: I wonder if this should be called "RegularExpressionLiteral" to match up with the other token names?
The specification also refers to them as [RegularExpressionLiteral](https://tc39.es/ecma262/#prod-RegularExpressionLiteral)s too

jasonwilliams:
@999eagle I've taken your list and turned it into a milestone:
https://github.com/jasonwilliams/boa/milestone/3

Issue #95, opened 2019-09-02
Implement missing assign operators


oll3:
Missing implementation of assign operators:
AssignMul: x *= y
AssignPow: x **= y
AssignAdd: x += y
AssignSub: x -= y
AssignMod: x %= y
AssignOr: x |= y
AssignAnd: x &= y
AssignXor: x ^= y
AssignLeftSh: x <<= y
AssignRightSh: x >>= y

PR #97, opened 2019-09-07
Implement assign operators (#95)

oll3:
* Operators implemented: +=, -=, /=, *=, %=, &=, |=, ^=, <<=, >>=

* Add parser tests for new operators

### DIFF
--- a/src/lib/exec.rs
+++ b/src/lib/exec.rs
@@ -11,7 +11,7 @@ use crate::{
     syntax::ast::{
         constant::Const,
         expr::{Expr, ExprDef},
-        op::{BinOp, BitOp, CompOp, LogOp, NumOp, UnaryOp},
+        op::{AssignOp, BinOp, BitOp, CompOp, LogOp, NumOp, UnaryOp},
     },
 };
 use gc::{Gc, GcCell};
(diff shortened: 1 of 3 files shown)


src/lib/exec.rs
```diff
@@ -41,6 +41,21 @@ pub struct InterpreterBuilder {
     global: Value,
 }
 
+fn exec_assign_op(op: &AssignOp, v_a: ValueData, v_b: ValueData) -> Gc<ValueData> {
```
jasonwilliams: Should this be `Value` instead of `Gc<ValueData>` ?

oll3: Missed that, but fixed now :)

src/lib/exec.rs
```diff
@@ -265,6 +280,28 @@ impl Executor for Interpreter {
...
+                match a.def {
+                    ExprDef::Local(ref name) => {
+                        let v_r_a = self.environment.get_binding_value(&name);
+                        let v_a = (*v_r_a).clone();
+                        value = exec_assign_op(op, v_a, v_b);
+                        self.environment
+                            .set_mutable_binding(&name, value.clone(), true);
+                    }
+                    ExprDef::GetConstField(ref obj, ref field) => {
+                        let v_r_a = self.run(obj)?;
```
jorendorff: The spec says the left hand side of compound assignment is evaluated first. In `f().x += g()`, first `f` is called, then `g`.

oll3: Thanks for pointing out, will have a look at it later.

Issue #128, opened 2019-10-02
[Easy] - consistency with how objects are added to Global


jasonwilliams:
object, console, math, function, json should follow the same pattern as String, Regex below
https://github.com/jasonwilliams/boa/blob/master/src/lib/realm.rs#L55-L59

## Example to follow:
Here is an example of Boolean being changed to `create_constructor` 
https://github.com/jasonwilliams/boa/commit/e35b66b41213cf519e85353b83b0952a95c4eda0#diff-9c85558881ecea17184e97fb52b0882b

DomParfitt:
I'd be interested in having a go at this.

PR #129, opened 2019-10-03
WIP - Update JSON Global

DomParfitt:
Updates the JSON object to be consistent with other global objects per #128.

@jasonwilliams are you able to have a quick look and let me know if I'm on the right track? I've been using the existing code and the ECMA spec for the [JSON object](https://tc39.es/ecma262/#sec-json-object) but I'm not hugely familiar with the ES internals.


jasonwilliams:
I've made some changes to the benchmarks Action, could you try rebasing and pushing again?

DomParfitt:
Thanks @jasonwilliams, I've rebased the changes from the upstream onto my branch but it seems that the benchmarks are still failing. I'm not really sure where to start looking for this.

DomParfitt:
Also, I was considering doing the change for each global as its own PR just so they're a bit more digestible. Are you happy for me to do it this way or would you rather they all be done as a single PR?

jasonwilliams:
yeah, that makes sense

jasonwilliams:
@DomParfitt looking good, im not sure if you need to rebase, some changes in there i think are merged now

Issue #114, opened 2019-09-27
String.match(regexp)


jasonwilliams:
`match` function would need to be added to the string object here:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/string.rs

It would be very similar to:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/regexp.rs#L212-L251

Spec:
https://tc39.es/ecma262/#sec-string.prototype.match

muskuloes:
Hi, I found this via  the hacktoberfest tag. I'll like to work on it.

jasonwilliams:
@muskuloes ok assigned

PR #131, opened 2019-10-04
implement string.match(regexp)

muskuloes:
an implementation of strings.match from issue #114

### DIFF
--- a/src/lib/js/string.rs
+++ b/src/lib/js/string.rs
@@ -4,12 +4,11 @@ use crate::{
         function::NativeFunctionData,
         object::{Object, ObjectKind, PROTOTYPE},
         property::Property,
+        regexp::{make_regexp, r#match as regexp_match},
         value::{from_value, to_value, ResultValue, Value, ValueData},
     },
 };
 use gc::Gc;
-use regex::Regex;
-use std::ops::Deref;
 use std::{
     cmp::{max, min},
     f64::NAN,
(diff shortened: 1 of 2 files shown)


jasonwilliams:
Thanks, im just looking into the benchmarks issue so bear with me

jasonwilliams:
I've made some changes to the benchmarks Action, could you try rebasing and pushing again?

muskuloes:
The benchmarks still failed.

jasonwilliams:
Ugh... 
bear with me, might be a couple of days

src/lib/js/string.rs
```diff
@@ -413,6 +415,81 @@ pub fn last_index_of(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Res
...
+    let mut regex_body = String::new();
+    let mut regex_flags = String::new();
+    match args[0].deref() {
+        ValueData::String(ref body) => {
+            // first argument is a string -> use it as regex pattern
+            regex_body = body.into();
+        }
+        ValueData::Object(ref obj) => {
+            let slots = &*obj.borrow().internal_slots;
+            if slots.get("RegExpMatcher").is_some() {
```
jasonwilliams: You could use .get_internal_slot() here, which gives you either the value or `undefined`

jasonwilliams: i think you also want to handle the case where RegExpMatcher isn't there.
(An object was passed but it wasn't a RegExp)

src/lib/js/string.rs
```diff
@@ -413,6 +415,81 @@ pub fn last_index_of(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Res
...
+                // first argument is another `RegExp` object, so copy its pattern and flags
+                if let Some(body) = slots.get("OriginalSource") {
+                    regex_body = from_value(body.clone()).unwrap();
+                }
+                if let Some(flags) = slots.get("OriginalFlags") {
+                    regex_flags = from_value(flags.clone()).unwrap();
+                }
+            }
```
jasonwilliams: You could probably just return undefined here if RegExpMatcher isn't there

muskuloes:
Hi, should I move the core functionality to `regexp.rs` as per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@match as well?

jasonwilliams:
Yeap, although we don't have support for Symbols yet

muskuloes:
Yes. I will. I'll push the update over the weekend.

muskuloes:
PR updated

jasonwilliams:
This is really awesome, thanks so much, looks clean and the tests are great

PR #134, opened 2019-10-05
[Lexer]add single and multiple line test

Emanon42:


jasonwilliams:
looks like the test fail as its struggling to parse your test case.
For rust formatting i think you can just run `cargo fmt --all` in the terminal and it should format your changes.

PR #138, opened 2019-10-06
String match all fixes #115

### LINKED ISSUE #115: String.matchAll(regexp)
`matchAll` function would need to be added to the string object here:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/string.rs

It would be very similar to:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/regexp.rs#L212-L251

Spec:
https://tc39.es/ecma262/#sec-string.prototype.matchAll

Notes:

bojan88:
Implementation for #115.

Since `matchAll` is [part of RegExp prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@matchAll), and `String` objecti is calling it internally, I tried to follow that with my implementation.

There is still some work needed for `matchAll` to be finished. The method should return a JS iterator, but with this implementation it's returning an array. From what I understand, we don't have iterators implemented yet.

### DIFF
--- a/src/lib/js/regexp.rs
+++ b/src/lib/js/regexp.rs
@@ -264,8 +264,8 @@ pub fn match_all(this: &Value, arg_str: String) -> ResultValue {
     let matches: Vec<Value> = this.with_internal_state_ref(|regex: &RegExp| {
         let mut matches = Vec::new();
 
-        'find_loop: for m in regex.matcher.find_iter(&arg_str) {
-            for caps in regex.matcher.captures_iter(&m.as_str()) {
+        for m in regex.matcher.find_iter(&arg_str) {
+            if let Some(caps) = regex.matcher.captures(&m.as_str()) {
                 let match_vec = caps
                     .iter()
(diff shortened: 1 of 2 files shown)


jasonwilliams:
Hi @bojan88 

Using [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) as an example

```js
let regexp = /t(e)(st(\d?))/g;
let str = "test1test2";

str.matchAll(regexp)[0][1];
```
Should return 'e' instead i get 'undefined'

bojan88:
Thanks.
I completely forgot to handle capture groups.
I will update the PR.

@jasonwilliams do you have any other requests or suggestions?

jasonwilliams:
@bojan88 its looking ok to me, just the conflict now

bojan88:
@jasonwilliams the conflict is fixed. Thanks.

PR #139, opened 2019-10-07
init string.replace work

louisgjohnson:
## TODO
Check it matches specs.
Add ability to pass function as second param.
Write some tests

### DIFF
--- a/src/lib/js/string.rs
+++ b/src/lib/js/string.rs
@@ -1,3 +1,4 @@
+use std::ops::Deref;
 use crate::{
     exec::Interpreter,
     js::{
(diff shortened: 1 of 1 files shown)


louisgjohnson:
BTW, for time writing rust so if I'm doing anything weird or wrong just let me know!

jasonwilliams:
How is this going @louisgjohnson  ?

louisgjohnson:
> How is this going @louisgjohnson ?

Hey sorry mate, I've been occupied with work and trying to learn some more rust. I'll just back onto this issue soon.

Is there any examples of how I would go about handling a JS function as a param?

jasonwilliams:
@louisgjohnson 
> Is there any examples of how I would go about handling a JS function as a param?

Look at some of the array methods
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/array.rs#L408-L436

jasonwilliams:
This has now been added

PR #144, opened 2019-10-08
Updated Math Global

DomParfitt:
Updates the `Math` object to be consistent with other global objects per #128.


jasonwilliams:
sorry i merged Console and caused a conflict

DomParfitt:
No problem, I figured breaking it up like this would probably cause some conflicts but thankfully they're quick and easy to resolve :slightly_smiling_face:

PR #146, opened 2019-10-11
WIP: starting to refactor Objects fixes #141

### LINKED ISSUE #141: Implement Function Objects
The goal of this issue is to have a function object which more closely resembles https://tc39.es/ecma262/#sec-ecmascript-function-objects.

I created function_object (which will hopefully replace function).

In the specification function objects are actually just ordinary objects with some additional methods and internal slots. I originally attempted to do the same but ran into issues.

I needed to deviate from the spec a little bit here…

## Deviation
Implementing https://tc39.es/ecma262/#sec-ordinaryfunctioncreate is tricky, we can't just take an object and bolt on properties + methods on to it, Rust doesn't work in that way, its not a dynamic language and also doesn't have inheritance. 
In the spec Functions are basically objects except they have quite a few additional internal slots, and some additional internal methods (Call, Construct).

Now internal slots can be handled in Object Create (where we basically pass through the slots we want on that object) that isn't a big issue.
The problem is Internal Methods.

We can't just add `pub fn call`  and `pub fn construct` to every object, as 90% of the objects won't need those methods. 

We can't have an internal slot that’s a [[Call]] function either because the built-in and dynamic functions have different types. Another reason for not going down this route is because we would need to be checking if the current object is a function all over the place, (loads of if ObjectKind::Function).

At this point its better to just have function be its own thing, I think call can be a method which will either iterate through an expression (dynamic function) or call a Rust built-in function (builtin function), I think we can abstract this inside of call, the end user won't care so long as it returns a \<Value\>.

With this in mind we can create a new `function function::fromDynamicFunction([expr])` or `function::fromBuiltIn([rust_func])`

So thats my rambling, would be interested to here from @Razican @jorendorff @IovoslavIovchev as to what you think is best here?

I made a start here:
https://github.com/jasonwilliams/boa/pull/255

jasonwilliams:


jasonwilliams:
superseeded by https://github.com/jasonwilliams/boa/pull/255

PR #150, opened 2019-10-12
Implement this expression

sanxiyn:
This is my first PR here!

After playing with `boa` for a while, I noticed `this` is unimplemented. I tried to implement it with as minimal changes as possible, so it may not fit the overall architecture. If so, just tell me how should I update.

### DIFF
--- a/src/lib/environment/environment_record_trait.rs
+++ b/src/lib/environment/environment_record_trait.rs
@@ -10,7 +10,7 @@
 //!
 use crate::{
     environment::lexical_environment::{Environment, EnvironmentType},
-    js::value::Value,
+    js::value::{self, Value},
 };
 use gc::{Finalize, Trace};
 use std::fmt::Debug;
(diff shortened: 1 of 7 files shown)


src/lib/environment/environment_record_trait.rs
```diff
@@ -60,6 +60,11 @@ pub trait EnvironmentRecordTrait: Debug + Trace + Finalize {
     /// Return true if it does and false if it does not.
     fn has_this_binding(&self) -> bool;
 
+    fn get_this_binding(&self) -> Value {
+        debug_assert!(self.has_this_binding());
```
jasonwilliams: not sure if the assert is needed?
It won't even compile if a `has_this_binding()` isn't implemented

jasonwilliams:
hmm not sure about this, `this` is just a variable like any other, and [we create it](https://github.com/jasonwilliams/boa/blob/master/src/lib/exec.rs#L328) before entering a function. How will we deal with this.something() with your change?

src/lib/environment/environment_record_trait.rs
sanxiyn: `has_this_binding` can return false for some environment records, although the interpreter should never call `get_this_binding` for them. I am asserting that invariant here.

sanxiyn:
Okay I will add a test with constructors.

sanxiyn:
Now with better understanding, I reimplemented this in #185.

PR #153, opened 2019-10-13
Fix clippy warnings

Korpen:
This PR aims to fix all the current clippy warnings in the project. Some of them were very straight-forward, but some required a decision to be made, below follows some design decisions:

Some of the clippy warnings were in tests, most of these I disabled since I thought fixing them would make the tests harder to read and since it's test code and not production code then a panic caused by, for example, integer overflow is okay.

Many warnings were caused by integer arithmetic to which there are several built-in methods that can be used instead. For these cases I decided to use methods that wrap in the case of an overflow. In all the cases I saw it did not seem like an overflow should ever be encountered.

Another common issue was using unwrap instead of handling None or Err manually. For most of these cases I replaced the unwrap with expect and provided a contextual error message. For some, where it made more sense, I used a match-block to explicitly handle None/Err.


jasonwilliams:
@Korpen `cargo fmt` looks like its failing
Great work btw

Issue #154, opened 2019-10-14
Lexer doesn't handle all white spaces


evomassiny:
Hello,

By toying with this crate I stumble across a slight bug in the lexer, 
using the current boa version, any non-`" "` white space character triggers a lexing error,
this code:
```
fn main() {
    // mind the '\t' in the js expression
    let mut lexer = boa::syntax::lexer::Lexer::new("\tconsole.log('foo');");
    let lex_res = lexer.lex().expect("Lexing failed");
}
```
fails and produce this stacktrace: 
```
Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/bug_demo`
thread 'main' panicked at '1:1: Unexpected '    '', /home/some_user/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.4.0/src/lib/syntax/lexer.rs:576:23
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:214
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:339
   9: boa::syntax::lexer::Lexer::lex
             at /home/some_user/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.4.0/src/lib/syntax/lexer.rs:576
  10: pac_proxy::main
             at src/main.rs:34
  11: std::rt::lang_start::{{closure}}
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54/src/libstd/rt.rs:64
  12: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:49
[...]

PR #155, opened 2019-10-14
Handle white spaces as described in ECMAScript specs

evomassiny:
Hello,

This solves the issue:  #154
I've simply added all whitespace characters defined in the ECMAScript specification in the main `fn lex()` loop.
I've followed [this part](https://www.ecma-international.org/ecma-262/6.0/#sec-white-space) of the specification, according to the code points defined [here](https://www.compart.com/en/unicode/category/Zs#UNC_CAT).

I hope this helps,
thank you for your time,

Evomassiny


evomassiny:
thanks for your suggestion, I've edited it accordingly.

PR #156, opened 2019-10-14
Add Array.prototype.find method fixes #165

### LINKED ISSUE #165: Implement Array.prototype.find()
Array methods are implemented here, for e.g here's concat:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/array.rs#L78-L109

**Specification:**
https://tc39.es/ecma262/#sec-array.prototype.find

**Contributing & Debugging**
https://github.com/jasonwilliams/boa#contributing

letmutx:


src/lib/exec.rs
```diff
@@ -441,7 +441,7 @@ impl Executor for Interpreter {
 
 impl Interpreter {
     /// https://tc39.es/ecma262/#sec-call
-    fn call(&mut self, f: &Value, v: &Value, arguments_list: Vec<Value>) -> ResultValue {
+    pub(crate) fn call(&mut self, f: &Value, v: &Value, arguments_list: Vec<Value>) -> ResultValue {
```
jasonwilliams: What's this change?

letmutx: `find` method takes a callback to be applied on each element on the array. To call the function, we need access to this function in `array.rs`. So, I made it public in the crate. Usage [here](https://github.com/jasonwilliams/boa/pull/156/files#diff-c364bd20ac819c860ea950e0d684c3f0R276).

jasonwilliams: Ok not seen that syntax before pub(crate)

PR #161, opened 2019-10-16
Enable obj[key] = value syntax.

letmutx:


src/lib/js/value.rs
```diff
@@ -402,6 +402,17 @@ impl ValueData {
     pub fn set_field(&self, field: String, val: Value) -> Value {
         match *self {
             ValueData::Object(ref obj) => {
+                if obj.borrow().kind == ObjectKind::Array {
```
sanxiyn: You can use `if let` here to avoid deriving `Eq` for `ObjectKind`.

letmutx:
Rebased with master

Issue #170, opened 2019-10-19
Why was the Display implementation for Object removed?


IovoslavIovchev:
Hello,

While playing around with this project I came accross:
```rust
#[test]
fn concat() {
    //TODO: array display formatter
```
and, after some digging, I found that at some point in the past `Display` _was_ implemented for `ValueData::Object` (as far as I know, arrays in `boa` are objects).

So why exactly was it removed? Was it because the previous implementation was too verbose?

jasonwilliams:
Hey @IovoslavIovchev I can answer this one.
Basically we were experiencing stack overflows when printing objects.

After investigation I saw that the printer prints every property and slot for an object.
This creates an infinite loop because the prototype hidden slot of String will point to Object so all of Objects properties get printed and so on.

I think for now I changed object to just print {} (I tried reducing it to just properties but I can’t remember why that didn’t work)

Anyway, there is some work to be done need to match whatever the other implementations do

jasonwilliams:
The output was getting too big and verbose. If there is a nice way to pretty print objects I’m for it.

jasonwilliams:
If you wish to experiment you can try adding it back, then try printing a string object

IovoslavIovchev:
I tried various forms of `String`s, `Object`s and `Array`s but none of them caused a stack overflow (I'm using 2427b3b, by the way).

Did a stack overflow occur every time, or just in particular cases?

IovoslavIovchev:
One way to avoid verbosity (and a stack overflow) would be to only print own properties and not print the prototype, if that is something that might be deemed acceptable.

Additionally, pretty printing can be achieved by simply printing each new property on a new line.

jasonwilliams:
Sure, it’s a good start

I think all objects should attempt to call their own .toString() counterpart before being displayed anyway.

IovoslavIovchev:
I will take a look at re-implementing `Display` then.

jasonwilliams:
Thanks @IovoslavIovchev

IovoslavIovchev:
Done in #211

PR #171, opened 2019-10-19
Implement Display for Objects

IovoslavIovchev:
**NB:** Due to the fact that the main binary uses the `dbg!` macro, pretty printing, while present, does not the make output, well.. _pretty_ (e.g. new lines are rendered as `\n`).


IovoslavIovchev:
#170

jasonwilliams:
This may be out of scope for your PR, but be weary of circular references.
Printing an object A which has a property B which has a property back to A I think causes an infinite loop.

IovoslavIovchev:
This implementation indeed does not take circular dependencies into consideration. 

I believe I have an idea how that can be achieved, so I will take some more time to look into it and deliver everything in one PR.

IovoslavIovchev:
@jasonwilliams could you take a look?

IovoslavIovchev:
I decided to not filter out the prototype slot, as I still have not found a scenario which causes a stack overflow.
I propose we leave it as is, and if needed, remove it later.

jasonwilliams:
Do we want to somehow share this logic with `console.log` https://github.com/jasonwilliams/boa/blob/master/src/lib/js/console.rs ?

IovoslavIovchev:
We definitely could do that, only we need to find a way to escape code duplication for printing properties.

jasonwilliams:
Not sure if you're aware but the newlines aren't being parsed
This is the output i get
```
[src/bin/bin.rs:57] exec(&buffer) = "{\n\n    __proto__: {\n        valueOf: function() { [native code] },\n        constructor: {\n            prototype: \'[Cycle]\'\n            call: function() { [native code] },\n            construct: function() { [native code] }\n        },\n        toString: function() { [native code] }\n        BooleanData: false,\n        extensible: true,\n        __proto__: undefined\n    },\n    BooleanData: false\n}"
```

IovoslavIovchev:
@jasonwilliams As I have mentioned in the PR description, this is because of the `dbg!` macro. It does not render new lines or tabs, etc. but rather `\n`, `\t`, etc.

If you used `println!` rather than `dgb!` on ln 57 in `src/bin/bin.rs`, you would see the output beautified. Hope that helps.

jasonwilliams:
@IovoslavIovchev oh my bad.
Should we change bin.rs to print() instead of dbg! then?

IovoslavIovchev:
Well, that really depends. In my opinion we should not, as bin.rs is used for debugging purposes anyway, and it is better to see that the output is what you actually expect.

The downside obviously is that when it wraps (the way it does in your case) it gets really hard to read.

IovoslavIovchev:
Additionally, one slight mistake I made was I joined the properties and slots before printing and that resulted in
```
        valueOf: function() { [native code] },     
        toString: function() { [native code] }     
        BooleanData: false,
```
as you can see there is no comma after `toString`.

However, I have fixed it now. You can review the latest change.

jasonwilliams:
Sure, you also have a conflict

IovoslavIovchev:
@jasonwilliams Regarding sharing some functionality with `console.log`, the way I see it, we could either create a macro (I am hoping this way we would be able to eliminate _even_ the code duplication for slots and properties) or have two separate functions that would return `Vec<String>`s, one of which we would be able to reuse for `console.log`.

What are your thoughts on that? I am open to any suggestions.

jasonwilliams:
I’m liking the sound of the macro idea

IovoslavIovchev:
I will take a look at what can be done

jasonwilliams:
@IovoslavIovchev here is an example of circular deps printing forever
https://gist.github.com/jasonwilliams/62616cf8a1529411f681e67aec847fa0

That's the global object, its all fine until it hits `String` which has a prototype, and the prototype has a `constructor` property which points back to `String`, then `String`s "prototype property points back to the prototype object, and it goes on infinitely

jasonwilliams:
@IovoslavIovchev is this PR still WIP?

IovoslavIovchev:
Yes, it is. Unfortunately, I haven't found an adequate way to do what I want, and am still researching.

jasonwilliams:
No worries, was just curious

jasonwilliams:
You need any help with this?

IovoslavIovchev:
I hit a wall with `prototype`s. When the execution reaches `length` (for instance when printing `Array.prototype`) I get a `None` and the program panics.

src/lib/builtins/array.rs
```diff
@@ -698,25 +700,25 @@ mod tests {
...
-        let _ne = forward(&mut engine, "one.concat(empty)");
-        //assert_eq!(ne, String::from("a.b.c"));
-        // NonEmpty ++ NonEmpty
-        let _nn = forward(&mut engine, "one.concat(one)");
-        //assert_eq!(nn, String::from("a.b.c"));
+        // let realm = Realm::create();
+        // let mut engine = Executor::new(realm);
+        // let init = r#"
+        // var empty = new Array();
+        // var one = new Array(1);
```
IovoslavIovchev: Currently, `new Array(1)` produces `[1]` (an array with one element, which is `1`), when, in fact, it should produce an array with `length` 1 and its first element should be `undefined`.

IovoslavIovchev:
@jasonwilliams could you take a look again?

src/lib/builtins/array.rs
```diff
@@ -326,14 +326,14 @@
~            "missing callback when calling function Array.prototype.every".to_string(),
~        ));
~    }
~    let callback = &args[0];
~    let this_arg = if args.len() > 1 {
~        args[1].clone()
~    } else {
```
jasonwilliams: Looks like `Array.prototype.length` shouldn't have a getter according to the spec

src/lib/builtins/array.rs
jasonwilliams: We will pick this up in https://github.com/jasonwilliams/boa/issues/209

src/lib/builtins/string.rs
```diff
@@ -741,7 +741,9 @@ pub fn create_constructor(global: &Value) -> Value {
 
     // Create prototype
     let proto = ValueData::new_obj(Some(global));
-    let prop = Property::default().get(to_value(get_string_length as NativeFunctionData));
+    let prop = Property::default()
+        .get(to_value(get_string_length as NativeFunctionData))
```
jasonwilliams: `String.prototype.length` does not have a getter according to the spec https://tc39.es/ecma262/#sec-properties-of-the-string-prototype-object

PR #173, opened 2019-10-20
Implement block scoped variable declarations

### LINKED ISSUE #39: Block scope not being used on variable allocation
When we step into a new function, we create a `new_function_environment` https://github.com/jasonwilliams/boa/blob/master/src/lib/exec.rs#L110

Because we don't do the same for blocks (if,while etc) any let values we create in blocks are added to the current environment, which will be the outer function most likely. This means the following JS works..

```js
function outer() {
  var foo = "foo";
  if (true) {
    let bar = "bar";
  }

  return bar; // "bar"
}
```

###  What does the specification say?
_When a Block or CaseBlock is evaluated a new declarative Environment Record is created and bindings for each block scoped variable, constant, function, or class declared in the block are instantiated in the Environment Record._

https://tc39.es/ecma262/#sec-blockdeclarationinstantiation

### high level task
The process here is to create a `new_declarative_environment` when we hit a block and set that to the current environment, popping it off when we reach the end of our block. A potential bug this will cause is putting `var` declarations into the same scope, so we need to make sure `vars` still go into the function scope above.

### low level

- [new_declaration_environment](https://github.com/jasonwilliams/boa/blob/master/src/lib/environment/lexical_environment.rs#L157) is already exposed here. Bring it into exec.rs so that it can be used similar to how `new_function_environment` is used.
- lexical_environment.rs will need a way of setting bindings to the nearest function environment, rather than the current environment (to support `var`)

barskern:
`const` and `let` are now scoped to the block, while `var` is scoped to
the surronding function (or global).

**NB!** Another bigger change is that all tests have been changed to use `var`
instead of `let` or `const`. This is because every time `forward` is
called with some new Javascript to evaluate, we parse it as a block,
hence variables can only persist across calls to `forward` if they are
defined using `var`. I assume it is intentional to parse each call as a
separate block, because a block is the only `ExprDef` which can contain
multiple statements.

Closes #39

### DIFF
--- a/src/lib/environment/lexical_environment.rs
+++ b/src/lib/environment/lexical_environment.rs
@@ -202,6 +202,11 @@ impl LexicalEnvironment {
             .expect("Could not get mutable reference to back object")
     }
 
+    pub fn has_binding(&self, name: &str) -> bool {
+        self.environments()
+            .any(|env| env.borrow().has_binding(name))
+    }
+
     pub fn get_binding_value(&mut self, name: &str) -> Value {
         self.environments()
             .find(|env| env.borrow().has_binding(name))
@@ -310,6 +315,32 @@ mod tests {
             var bar = "bar";
           }
           bar == "bar";
+        "#;
+
+        assert_eq!(&exec(scenario), "true");
+    }
+
+    #[test]
+    fn set_outer_var_in_blockscope() {
+        let scenario = r#"
+          var bar;
+          {
+            bar = "foo";
+          }
+          bar == "foo";
+        "#;
+
+        assert_eq!(&exec(scenario), "true");
+    }
+
+    #[test]
+    fn set_outer_let_in_blockscope() {
+        let scenario = r#"
+          let bar;
+          {
+            bar = "foo";
+          }
+          bar == "foo";
         "#;
 
         assert_eq!(&exec(scenario), "true");
--- a/src/lib/environment/lexical_environment.rs
+++ b/src/lib/environment/lexical_environment.rs
@@ -99,6 +99,12 @@ impl LexicalEnvironment {
         self.environment_stack.pop_back();
     }
 
+    pub fn environments(&self) -> impl Iterator<Item = Environment> {
+        std::iter::successors(Some(self.get_current_environment_ref().clone()), |env| {
+            env.borrow().get_outer_environment()
+        })
+    }
+
     pub fn get_global_object(&self) -> Option<Value> {
         self.environment_stack
             .get(0)
@@ -116,16 +122,14 @@ impl LexicalEnvironment {
             VariableScope::Function => {
                 // Find the first function or global environment (from the top of the stack)
                 let env = self
-                    .environment_stack
-                    .iter_mut()
-                    .rev()
+                    .environments()
                     .find(|env| match env.borrow().get_environment_type() {
                         EnvironmentType::Function | EnvironmentType::Global => true,
                         _ => false,
                     })
                     .expect("No function or global environment");
 
-                env.borrow_mut().create_mutable_binding(name, deletion)
+                env.borrow_mut().create_mutable_binding(name, deletion);
             }
         }
     }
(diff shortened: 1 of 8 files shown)


src/lib/environment/lexical_environment.rs
```diff
@@ -98,25 +107,78 @@ impl LexicalEnvironment {
...
+    pub fn create_mutable_binding(&mut self, name: String, deletion: bool, scope: VariableScope) {
+        match scope {
+            VariableScope::Block => self
+                .get_current_environment()
+                .borrow_mut()
+                .create_mutable_binding(name, deletion),
+            VariableScope::Function => {
+                // Find the first function or global environment (from the top of the stack)
+                let env = self
+                    .environment_stack
```
jasonwilliams: This is great, although worth pointing out.
I want to move away from using the stack at some point.

Instead every environment has an `outer_env` pointer to the environment above them
https://github.com/jasonwilliams/boa/blob/master/src/lib/environment/declarative_environment_record.rs#L167-L169

It will be more flexible in future if we can rely on this.
The stack will only get us so far, but traversing an environment tree should rely on traversing it's parent_env until you reach null.

jasonwilliams: Trait methods for every environment record:
https://github.com/jasonwilliams/boa/blob/master/src/lib/environment/environment_record_trait.rs#L71-L75

barskern: Ah okay, then I will refactor the code to reflect this. Perhaps I could add a function to the trait which returns an iterator over the parent environments?

jasonwilliams: @barskern i don't think you need that, there's only 1 place that's going to be iterating up the environments and that's here:
https://github.com/jasonwilliams/boa/blob/master/src/lib/environment/lexical_environment.rs#L141-L173

Setting here:
https://github.com/jasonwilliams/boa/blob/master/src/lib/environment/lexical_environment.rs#L113-L115

environments themselves shouldn't care about iterating (that's the job of the `lexical_environment` module

barskern:
@jasonwilliams I spent some time and I managed to got it working, however a few of the tests were failing. After a bit of debugging I found out that I wasn't able to get from an declerative environment to its parent, and then I found out that it unconditionally returns `None`. Is there a reason for this?

https://github.com/jasonwilliams/boa/blob/a87f04f952cd4623aa096650af08ea87677550be/src/lib/environment/declarative_environment_record.rs#L163-L165

jasonwilliams:
That’s weird, only global should return None. I’m not sure, it should return a parent env if there is one

barskern:
@jasonwilliams Ah okay, thats what I was thinking aswell. I have changed it now so that it will return the parent if there is one.

src/lib/environment/lexical_environment.rs
```diff
@@ -90,6 +99,12 @@ impl LexicalEnvironment {
         self.environment_stack.pop_back();
     }
 
+    pub fn environments(&self) -> impl Iterator<Item = Environment> {
+        std::iter::successors(Some(self.get_current_environment_ref().clone()), |env| {
+            env.borrow().get_outer_environment()
+        })
+    }
+
```
jasonwilliams: nice utility function

jasonwilliams:
I see what you mean about parse_all() creating a new block each time. 
I suppose we do want this, so yeah they will have to change, that's a shame.

src/lib/environment/lexical_environment.rs
barskern: Thank you, I was quite happy with it :smile:

barskern:
> I see what you mean about parse_all() creating a new block each time.
> I suppose we do want this, so yeah they will have to change, that's a shame.

Yeah I see your point. One could rewrite tests to rather defined them as functions and then rather call the functions in the next invocation of `forward`. Kinda like the following:

```rust
let init = r#"
const empty = new Array();
const one = new Array(1);

function ee() {
	return empty.concat(empty);
}
"#;
forward(&mut engine, init);

let _ee = forward(&mut engine, "ee()");
```

But doing this would require rewriting all the tests and does bring some extra lines per test.

However as you also say it makes sense to parse the top level into a `Block` no matter what, as one could imagine that one wanted to have subsequent calls to `forward` be run in a separate block as to not intervene with each other.

PR #181, opened 2019-10-20
Array.prototype.fill method implementation

bojan88:
Implementation for https://github.com/jasonwilliams/boa/issues/164.

### DIFF
--- a/src/lib/builtins/array.rs
+++ b/src/lib/builtins/array.rs
@@ -3,7 +3,7 @@ use crate::{
         function::NativeFunctionData,
         object::{Object, ObjectKind, PROTOTYPE},
         property::Property,
-        value::{from_value, to_value, ResultValue, Value, ValueData, undefined},
+        value::{from_value, to_value, undefined, ResultValue, Value, ValueData},
     },
     exec::Interpreter,
 };
--- a/src/lib/builtins/array.rs
+++ b/src/lib/builtins/array.rs
@@ -479,18 +479,20 @@ pub fn fill(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
     let len: i32 = from_value(this.get_field_slice("length")).expect("Could not get argument");
     let default_value = Gc::new(ValueData::Undefined);
     let value = args.get(0).unwrap_or(&default_value);
-    let relateive_start: i32 = match args.get(1) {
-        Some(val) => val.to_num() as i32,
-        None => 0,
-    };
-    let relative_end: i32 = match args.get(2) {
-        Some(val) => val.to_num() as i32,
-        None => len,
-    };
-    let k = if relateive_start < 0 {
-        cmp::max(len + relateive_start, 0)
+    let relative_start = args.get(1).unwrap_or(&default_value).to_num() as i32;
+    let relative_end_val = args.get(2).unwrap_or(&default_value);
+    println!("{:?}", args.get(1).unwrap_or(&default_value));
+    let relative_end = if relative_end_val.is_undefined() {
+        len
+    } else if relative_end_val.is_null() {
+        0
     } else {
-        cmp::min(relateive_start, len)
+        relative_end_val.to_num() as i32
+    };
+    let k = if relative_start < 0 {
+        cmp::max(len + relative_start, 0)
+    } else {
+        cmp::min(relative_start, len)
     };
     let fin = if relative_end < 0 {
         cmp::max(len + relative_end, 0)
@@ -976,11 +978,12 @@ mod tests {
             String::from("4,2,3")
         );
 
-        forward(&mut engine, "a = [1, 2, 3];");
-        assert_eq!(
-            forward(&mut engine, "a.fill(4, NaN, NaN).join()"),
-            String::from("1,2,3")
-        );
+        // TODO: uncomment when NaN support is added
+        // forward(&mut engine, "a = [1, 2, 3];");
+        // assert_eq!(
+        //     forward(&mut engine, "a.fill(4, NaN, NaN).join()"),
+        //     String::from("1,2,3")
+        // );
 
         forward(&mut engine, "a = [1, 2, 3];");
         assert_eq!(
@@ -994,8 +997,27 @@ mod tests {
             String::from("1,4,3")
         );
 
+        forward(&mut engine, "a = [1, 2, 3];");
         assert_eq!(
             forward(&mut engine, "a.fill(4, 'str').join()"),
+            String::from("4,4,4")
+        );
+
+        forward(&mut engine, "a = [1, 2, 3];");
+        assert_eq!(
+            forward(&mut engine, "a.fill(4, 'str', 'str').join()"),
+            String::from("1,2,3")
+        );
+
+        forward(&mut engine, "a = [1, 2, 3];");
+        assert_eq!(
+            forward(&mut engine, "a.fill(4, undefined, null).join()"),
+            String::from("1,2,3")
+        );
+
+        forward(&mut engine, "a = [1, 2, 3];");
+        assert_eq!(
+            forward(&mut engine, "a.fill(4, undefined, undefined).join()"),
             String::from("4,4,4")
         );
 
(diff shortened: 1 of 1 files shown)


src/lib/builtins/array.rs
```diff
@@ -472,6 +472,43 @@ pub fn find_index(this: &Value, args: &[Value], interpreter: &mut Interpreter) -
...
+///
+/// The method fills (modifies) all the elements of an array from start index (default 0)
+/// to an end index (default array length) with a static value. It returns the modified array
+/// <https://tc39.es/ecma262/#sec-array.prototype.fill>
+pub fn fill(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let len: i32 = from_value(this.get_field_slice("length")).expect("Could not get argument");
+    let default_value = Gc::new(ValueData::Undefined);
+    let value = args.get(0).unwrap_or(&default_value);
+    let relateive_start: i32 = match args.get(1) {
+        Some(val) => from_value(val.clone()).expect("Could not get argument"),
```
letmutx: FWIW, the following also works in Firefox's js engine.
```
var a = [1, 2, 3, 1, 2, 5, 7, 11, 13, 17];
console.log(a.fill(4, "1.2", "3.4"));
```

src/lib/builtins/array.rs
```diff
@@ -472,6 +472,54 @@ pub fn find_index(this: &Value, args: &[Value], interpreter: &mut Interpreter) -
...
+pub fn fill(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let len: i32 = from_value(this.get_field_slice("length")).expect("Could not get argument");
+    let default_value = Gc::new(ValueData::Undefined);
+    let value = args.get(0).unwrap_or(&default_value);
+    let relateive_start: i32 = match args.get(1) {
+        Some(val) => {
+            if val.is_string() {
+                val.to_num() as i32
+            } else {
+                val.to_int()
```
bojan88: @letmutx I had problem here using only `val.to_int()` - it was converting every float to zero (e.g. `'2.5'` -> `0`)
So I had to introduce this `val.to_num() as i32` conversion

letmutx: The following seems to work. 
```
diff --git a/src/lib/builtins/array.rs b/src/lib/builtins/array.rs
index fe3a4aa..1e6caa8 100644
--- a/src/lib/builtins/array.rs
+++ b/src/lib/builtins/array.rs
@@ -482,23 +482,11 @@ pub fn fill(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
     let default_value = Gc::new(ValueData::Undefined);
     let value = args.get(0).unwrap_or(&default_value);
     let relateive_start: i32 = match args.get(1) {
-        Some(val) => {
-            if val.is_string() {
-                val.to_num() as i32
-            } else {
-                val.to_int()
-            }
-        }
+        Some(val) => val.to_num() as i32,
         None => 0,
     };
     let relative_end: i32 = match args.get(2) {
-        Some(val) => {
-            if val.is_string() {
-                val.to_num() as i32
-            } else {
-                val.to_int()
-            }
-        }
+        Some(val) => val.to_num() as i32,
         None => len,
     };
     let k = if relateive_start < 0 {
```

src/lib/builtins/array.rs
```diff
@@ -500,10 +512,9 @@ pub fn fill(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
         cmp::min(relative_end, len)
     };
 
-    while k < fin {
-        let pk = k.to_string();
+    for i in k..fin {
+        let pk = i.to_string();
         this.set_field_slice(&pk, value.clone());
```
letmutx: This should be `set_field`. Using `set_field_slice` does an extra allocation here.

src/lib/builtins/array.rs
bojan88: I saw that `val.to_num() as i32` is converting `ValueData::Undefined` to a very small negative number `-2147483648`.
But seems like the tests are all succeeding. 

I'm going to add this change

letmutx: `undefined` should behave like `None` case. Taken from Firefox
```
var a = [1, 2, 3];
a.fill(4, undefined, undefined);
> Array(3) [ 4, 4, 4 ]
​```

Comment (bojan88) [reply src/lib/builtins/array.rs]: @letmutx thanks for discovering these issues.
I updated updated the code and added more tests. I also noticed that `NaN` is still not implemented so I commented out tests using that value

**Comment by jasonwilliams** [comment]
sorry @bojan88 there's a conflict now

**Comment by bojan88** [comment]
No problem. I resolved the conflict

File: src/lib/builtins/array.rs [hunk 4]
```diff
@@ -470,6 +470,40 @@ pub fn find_index(this: &Value, args: &[Value], interpreter: &mut Interpreter) -
...
 }
 
+/// Array.prototype.fill ( value[, start[, end]] )
+///
+/// The method fills (modifies) all the elements of an array from start index (default 0)
+/// to an end index (default array length) with a static value. It returns the modified array
+/// <https://tc39.es/ecma262/#sec-array.prototype.fill>
+pub fn fill(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let len: i32 = from_value(this.get_field_slice("length")).expect("Could not get argument");
+    let default_value = Gc::new(ValueData::Undefined);
```
IovoslavIovchev: Use the [`undefined()`](https://github.com/jasonwilliams/boa/blob/master/src/lib/builtins/value.rs#L23) function here.

src/lib/builtins/array.rs
```diff
@@ -470,6 +470,40 @@ pub fn find_index(this: &Value, args: &[Value], interpreter: &mut Interpreter) -
...
+    let value = args.get(0).unwrap_or(&default_value);
+    let relative_start = args.get(1).unwrap_or(&default_value).to_num() as i32;
+    let relative_end_val = args.get(2).unwrap_or(&default_value);
+    let relative_end = if relative_end_val.is_undefined() {
+        len
+    } else {
+        relative_end_val.to_num() as i32
+    };
+    let k = if relative_start < 0 {
```
IovoslavIovchev: Could you give a more descriptive name than `k`?

src/lib/builtins/array.rs
```diff
@@ -573,6 +607,7 @@ pub fn create_constructor(global: &Value) -> Value {
     array_prototype.set_field_slice("includes", includes_func);
     array_prototype.set_field_slice("indexOf", index_of_func);
     array_prototype.set_field_slice("lastIndexOf", last_index_of_func);
+    array_prototype.set_field_slice("fill", to_value(fill as NativeFunctionData));
```
IovoslavIovchev: You need to set the `length` property of a function.

bojan88: @IovoslavIovchev can you please be more specific? I'm afraid I don't understand this comment

src/lib/builtins/array.rs
bojan88: Can you please propose any better name? What do you think about `start`?
The value is named `k` both in tc39 spec and in MDN polyfill example

IovoslavIovchev: Yeah, `start` is okay.

src/lib/builtins/array.rs
IovoslavIovchev: `Function`s have their `length` property set to the number of required parameters they have (e.g. those with no default value).

You can take a look at how it's currently done [here](https://github.com/bojan88/boa/blob/1728795ff784ffabe3ad88369a9bd2fe7a5b4f3a/src/lib/builtins/array.rs#L596).

bojan88: Thanks @IovoslavIovchev. I didn't know that `length` property existed on this function/object.

src/lib/builtins/array.rs
```diff
@@ -936,6 +973,109 @@ mod tests {
...
+        // make sure the array is modified
+        assert_eq!(forward(&mut engine, "a.join()"), String::from("4,4,4"));
+
+        forward(&mut engine, "a = [1, 2, 3];");
+        assert_eq!(
+            forward(&mut engine, "a.fill(4, '1').join()"),
+            String::from("1,4,4")
+        );
+
+        forward(&mut engine, "a = [1, 2, 3];");
```
IovoslavIovchev: It seems you added the same test twice. Remove this one.

Issue #116, opened 2019-09-27
String.replace(regexp|substr, newSubstr|function)


jasonwilliams:
`replace` function would need to be added to the string object here:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/string.rs

It would be very similar to:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/regexp.rs#L212-L251

Spec:
https://tc39.es/ecma262/#sec-string.prototype.replace

Notes:
https://github.com/jasonwilliams/boa/blob/master/docs/debugging.md

louisgjohnson:
Hey mate, can I have a go of this?

thiagoarrais:
@louisgjohnson: still working on this? I would like to help with it.

cemremengu:
@jasonwilliams I think this is already added (as I understand from the dicussions)

RageKnify:
@jasonwilliams you implemented this in #217, but as I noted in #629 , there is a bug with the current implementation:
Copied from Chrome:
```js
> let str = 'boa';
> str.replace(/.*/, "??$&??")
"??boa??"
> str.replace(".*", "??$&??")
"boa"
```
While on `boa` both would return `"??boa??"` because the string `".*"` is converted to a RegEx pattern.

Looking at the spec: https://tc39.es/ecma262/#sec-string.prototype.replace
I believe on point 2 if `searchValue` is a `RegExp` object it should give `RegExp.prototype [ @@replace ]` as `replacer`. Which would then be invoked. `@@replace` exists, but `RegExp.prototype [ @@replace ]` isn't implemented, I think the code in `String.replace` is an aproximation of it.

After #629 gets merged I'd like to look into implementing `RegExp.prototype[ @@replace ]` and making String.replace spec compliant.

jedel1043:
Closing, as the bug that blocked this is solved. If we overlooked something we can open another issue with the specific problem.

PR #190, opened 2019-10-24
Add String.replace

thiagoarrais:
Work in progress pull request for fixing #116 

The implementation isn't complete yet, but I intend to incorporate changes based on feedback. For starters, just let me know if this is the structure you expect. I'll try to evolve the `replace` method to be spec-compliant.

Related question: does this project use tests? Where should I add them if so?

TODO List:

- [x] Do not interpret string pattern as regex
- [x] Support regex pattern
- [x] Support funtion replacement
- [x] ~~Include function length (related: #193)~~

### DIFF
--- a/src/lib/builtins/regexp.rs
+++ b/src/lib/builtins/regexp.rs
@@ -14,7 +14,7 @@ use crate::{
 };
 
 #[derive(Debug)]
-struct RegExp {
+pub struct RegExp {
     /// Regex matcher.
     matcher: Regex,
     /// Update last_index, set if global or sticky flags are set.
@@ -36,6 +36,12 @@ struct RegExp {
 }
 
 impl InternalState for RegExp {}
+
+impl RegExp {
+    pub fn get_matcher(&self) -> &Regex {
+        &self.matcher
+    }
+}
 
 fn get_argument<T: FromValue>(args: &[Value], idx: usize) -> Result<T, Value> {
     match args.get(idx) {
--- a/src/lib/builtins/string.rs
+++ b/src/lib/builtins/string.rs
@@ -9,10 +9,7 @@ use crate::{
     exec::Interpreter,
 };
 use gc::Gc;
-use regex::{
-    Captures,
-    Regex,
-};
+use regex::{Captures, Regex};
 use std::{
     cmp::{max, min},
     f64::NAN,
--- a/src/lib/builtins/string.rs
+++ b/src/lib/builtins/string.rs
@@ -3,12 +3,13 @@ use crate::{
         function::NativeFunctionData,
         object::{Object, ObjectKind, PROTOTYPE},
         property::Property,
-        regexp::{make_regexp, match_all as regexp_match_all, r#match as regexp_match},
-        value::{from_value, to_value, ResultValue, Value, ValueData},
+        regexp::{make_regexp, match_all as regexp_match_all, r#match as regexp_match, RegExp},
+        value::{from_value, to_value, undefined, ResultValue, Value, ValueData},
     },
     exec::Interpreter,
 };
 use gc::Gc;
+use regex::Captures;
 use std::{
     cmp::{max, min},
     f64::NAN,
@@ -722,6 +723,60 @@ pub fn match_all(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultV
     regexp_match_all(&re, ctx.value_to_rust_string(this))
 }
 
+/// Return a String where the matches of a pattern in the original string is replaced with a given
+/// replacement.
+/// <https://tc39.es/ecma262/#sec-string.prototype.replace>
+pub fn replace(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
+    // First we get it the actual string a private field stored on the object only the engine has access to.
+    // Then we convert it into a Rust String
+    let this_str: &str = &ctx.value_to_rust_string(this);
+
+    // Make a Rust Regex from the first argument
+    let pattern = get_argument(args, 0);
+    if pattern.is_undefined() {
+        return Ok(to_value(this_str));
+    }
+
+    let pattern_re = if let Some(re_internal_state) = pattern.get_internal_state() {
+        if let Some(pattern_jsre) = re_internal_state.downcast_ref::<RegExp>() {
+            pattern_jsre.get_matcher().clone()
+        } else {
+            let pattern_str = &ctx.value_to_rust_string(&pattern);
+            regex::Regex::new(&regex::escape(pattern_str)).unwrap()
+        }
+    } else {
+        let pattern_str = &ctx.value_to_rust_string(&pattern);
+        regex::Regex::new(&regex::escape(pattern_str)).unwrap()
+    };
+
+    // Make a replacement closure from the second argument
+    let replacement = get_argument(args, 1);
+
+    let replacement_cls = if replacement.is_function() {
+        Box::new(|caps: &Captures| {
+            let m = caps.get(0).unwrap().as_str();
+            let args = vec![to_value(m)];
+            let rs_value = ctx
+                .call(&replacement, &undefined(), args)
+                .unwrap_or_else(|_| undefined());
+            ctx.value_to_rust_string(&rs_value)
+        }) as Box<dyn FnMut((&Captures)) -> (String)>
+    } else {
+        Box::new(|_: &Captures| ctx.value_to_rust_string(&replacement))
+    };
+
+    // Replace the pattern by the replacement closure
+    let result_str = pattern_re.replace_all(this_str, replacement_cls);
+    Ok(to_value(result_str.into_owned()))
+}
+
+fn get_argument(args: &[Value], idx: usize) -> Value {
+    match args.get(idx) {
+        Some(arg) => arg.clone(),
+        None => undefined(),
+    }
+}
+
 /// Create a new `String` object
 pub fn create_constructor(global: &Value) -> Value {
     // Create constructor function object
@@ -760,6 +815,7 @@ pub fn create_constructor(global: &Value) -> Value {
     make_builtin_fn!(substr, named "substr", with length 2, of proto);
     make_builtin_fn!(value_of, named "valueOf", of proto);
     make_builtin_fn!(match_all, named "matchAll", with length 1, of proto);
+    make_builtin_fn!(replace, named "replace", with length 2, of proto);
 
     let string = to_value(string_constructor);
     proto.set_field_slice("constructor", string.clone());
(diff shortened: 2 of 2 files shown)


jasonwilliams:
> Related question: does this project use tests? Where should I add them if so?
Yes, there are quite a few tests in this project, each file (or module) has the tests embedded with the code. So for String, you will see the tests at the bottom of the file. Im sure there are enough tests here for you to familiar yourself with.
https://github.com/jasonwilliams/boa/blob/master/src/lib/builtins/string.rs#L781-L1044

jasonwilliams:
Im not sure if using regex directly will work, as the rules may differ from rust to JS.
We have a RegExp object which has `match` here:
https://github.com/jasonwilliams/boa/blob/master/src/lib/builtins/regexp.rs#L257-L274

Maybe you could use that?

Secondly incase you're unaware your rustfmt is failing:
https://github.com/jasonwilliams/boa/pull/190/checks?check_run_id=273620005

You can run `cargo fmt --verbose` in the project

thiagoarrais:
Nice, @jasonwilliams! I'll add tests (sorry, hadn't located them just out of Rust-noobness) and rustfmt sometime today.

About RegExp, I started with that, but Rust complained about it not being exported or something. But my final version will most likely use that.

Thanks for the feedback!

thiagoarrais:
This is taking some shape now.

I have included a `get_argument` function inspired by but different than [`regexp::get_argument`](https://github.com/jasonwilliams/boa/blob/master/src/lib/builtins/regexp.rs#L40). It is designed to mimic JS' behaviour of interpreting missing parameters as `undefined`. I suspect such a function already exists somewhere in the codebase. Care to help me find it?

src/lib/builtins/string.rs
```diff
@@ -766,6 +798,7 @@ pub fn create_constructor(global: &Value) -> Value {
     proto.set_field_slice("substr", to_value(substr as NativeFunctionData));
     proto.set_field_slice("valueOf", to_value(value_of as NativeFunctionData));
     proto.set_field_slice("matchAll", to_value(match_all as NativeFunctionData));
+    proto.set_field_slice("replace", to_value(replace as NativeFunctionData));
```
IovoslavIovchev: You need to set the [`length`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) property of the function as well.

thiagoarrais: I don't know if I understood that. Do you mean `str.replace.length` should return `2`? If so, would you say that `str.charAt.length` should return `1` then?

IovoslavIovchev: Yes, to both your questions. A `function`'s length represents the number of "formal" parameters, e.g. those with no default value.

thiagoarrais: Do the others functions in `String` already support that in boa or are you suggesting that they should? I'm locally getting `str.charAt.length == undefined`, but I may have an outdated branch...

IovoslavIovchev: Unfortunately, some functions in boa don't have their length set because it is easily missable.

I have suggested an approach in #193 and would very much like we have something like that, so we don't miss like things like that.

thiagoarrais: Nice. I'm working on other stuff here, but I'll make sure to check where that PR of yours goes.

src/lib/builtins/regexp.rs
```diff
@@ -14,9 +14,9 @@ use crate::{
 };
 
 #[derive(Debug)]
-struct RegExp {
+pub struct RegExp {
     /// Regex matcher.
-    matcher: Regex,
+    pub matcher: Regex,
```
thiagoarrais: Need some help here. I'm feeling a bit uneasy about publishing this. What is the idiomatic way to access regex funcionality from outside the regexp module?

IovoslavIovchev: Do you intend to modify the value in any way?

The most "popular" and obvious way would be to have an immutable getter (and a mutable one only if _really_ needed).

thiagoarrais: I don't intend to modify the value. I was considering getter vs. something like adding a `replace_all` delegating method. The second option could be used to deal with differences between JS and Rust regex engines.

IovoslavIovchev: What exactly is your idea for `replace_all`?

src/lib/builtins/string.rs
```diff
@@ -728,6 +728,41 @@ pub fn match_all(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultV
...
+/// replacement.
+/// <https://tc39.es/ecma262/#sec-string.prototype.replace>
+pub fn replace(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
+    // First we get it the actual string a private field stored on the object only the engine has access to.
+    // Then we convert it into a Rust String by wrapping it in from_value
+    let this_str: &str = &ctx.value_to_rust_string(this);
+    // We then do the same thing for the other args
+    // TODO: what if there are missing arguments?
+    let pattern = get_argument(args, 0);
+    if undefined() == pattern {
```
IovoslavIovchev: If I am not mistaken, there was an `is_undefined` method.
So you can try `pattern.is_undefined()` instead.

thiagoarrais: Good catch!

thiagoarrais:
I think this is ready for another round of review.

My main concern here is about @IovoslavIovchev's suggestion on `length`. It doesn't seem to work, even after adapting for #206. I'm assuming this is unrelated to this change because some other functions in `String` don't return a valid length. But I'm open to adapting the code if needed.

thiagoarrais:
The problem with `length` seems related to #210. This should be good to go. Please let me know if you need anything else.

src/lib/builtins/string.rs
```diff
@@ -728,6 +729,60 @@ pub fn match_all(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultV
...
+/// <https://tc39.es/ecma262/#sec-string.prototype.replace>
+pub fn replace(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
+    // First we get it the actual string a private field stored on the object only the engine has access to.
+    // Then we convert it into a Rust String
+    let this_str: &str = &ctx.value_to_rust_string(this);
+
+    // Make a Rust Regex from the first argument
+    let pattern = get_argument(args, 0);
+    if pattern.is_undefined() {
+        return Ok(to_value(this_str));
```
IovoslavIovchev:
```suggestion
        return Ok(this.clone());
```

src/lib/builtins/string.rs
```diff
@@ -728,6 +729,60 @@ pub fn match_all(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultV
     regexp_match_all(&re, ctx.value_to_rust_string(this))
 }
 
+/// Return a String where the matches of a pattern in the original string is replaced with a given
+/// replacement.
+/// <https://tc39.es/ecma262/#sec-string.prototype.replace>
+pub fn replace(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
+    // First we get it the actual string a private field stored on the object only the engine has access to.
+    // Then we convert it into a Rust String
+    let this_str: &str = &ctx.value_to_rust_string(this);
```
IovoslavIovchev: Move this after the `if` statement below.

src/lib/builtins/string.rs
IovoslavIovchev: This is just a small optimisation to avoid calling `ctx.value_to_rust_string` if the regex is `undefined`

src/lib/builtins/string.rs
```diff
@@ -728,6 +729,60 @@ pub fn match_all(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultV
...
+
+    let replacement_cls = if replacement.is_function() {
+        Box::new(|caps: &Captures| {
+            let m = caps.get(0).unwrap().as_str();
+            let args = vec![to_value(m)];
+            let rs_value = ctx
+                .call(&replacement, &undefined(), args)
+                .unwrap_or_else(|_| undefined());
+            ctx.value_to_rust_string(&rs_value)
+        }) as Box<dyn FnMut((&Captures)) -> (String)>
```
IovoslavIovchev: Remove the parentheses, so it becomes

```suggestion
        }) as Box<dyn FnMut(&Captures) -> String>
```

src/lib/builtins/string.rs
```diff
@@ -728,6 +729,60 @@ pub fn match_all(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultV
...
+    } else {
+        Box::new(|_: &Captures| ctx.value_to_rust_string(&replacement))
+    };
+
+    // Replace the pattern by the replacement closure
+    let result_str = pattern_re.replace_all(this_str, replacement_cls);
+    Ok(to_value(result_str.into_owned()))
+}
+
+fn get_argument(args: &[Value], idx: usize) -> Value {
```
IovoslavIovchev: This is a bit more general, so it would be better you move it somewhere, where it can be widely accessed.

I am thinking _exec.rs_. What do you think?

src/lib/builtins/string.rs
```diff
@@ -1041,4 +1097,41 @@ mod tests {
...
+        var regexResult = str.replace(/dog./, 'monkey.');
+        var funResult1 = str.replace(/dog/, function(word) { return word.toUpperCase(); })
+        var funResult2 = str.replace('dog', function(word) { return word.toUpperCase(); })
+        var intResult = 'abc4e'.replace(4, 'd')
+        var exceptional1 = str.replace();
+        var exceptional2 = str.replace('dog');
+        "#;
+
+        forward(&mut engine, init);
+        assert_eq!(forward(&mut engine, "result1"), "The quick brown fox jumps over the lazy monkey. If the monkey reacted, was it really lazy?");
```
IovoslavIovchev:
```suggestion
        assert_eq!(forward(&mut engine, "result1"), "The quick brown fox jumps over the lazy monkey. If the dog reacted, was it really lazy?");
```

src/lib/builtins/string.rs
```diff
@@ -1041,4 +1097,41 @@ mod tests {
...
+        assert_eq!(forward(&mut engine, "result2"), "The quick brown fox jumps over the lazy monkey. If the dog reacted, was it really lazy?");
+        assert_eq!(forward(&mut engine, "regexResult"), "The quick brown fox jumps over the lazy monkey. If the monkey.reacted, was it really lazy?");
```
IovoslavIovchev:
```suggestion
        assert_eq!(forward(&mut engine, "regexResult"), "The quick brown fox jumps over the lazy monkey. If the dog reacted, was it really lazy?");
```

src/lib/builtins/string.rs
```diff
@@ -1041,4 +1097,41 @@ mod tests {
...
+        assert_eq!(
+            forward(&mut engine, "funResult1"),
+            "The quick brown fox jumps over the lazy DOG. If the DOG reacted, was it really lazy?"
```
IovoslavIovchev:
```suggestion
            "The quick brown fox jumps over the lazy DOG. If the dog reacted, was it really lazy?"
```

src/lib/builtins/string.rs
```diff
@@ -1041,4 +1097,41 @@ mod tests {
...
+        );
+        assert_eq!(
+            forward(&mut engine, "funResult2"),
+            "The quick brown fox jumps over the lazy DOG. If the DOG reacted, was it really lazy?"
```
IovoslavIovchev:
```suggestion
            "The quick brown fox jumps over the lazy DOG. If the dog reacted, was it really lazy?"
```

src/lib/builtins/string.rs
```diff
@@ -1041,4 +1097,41 @@ mod tests {
...
+        );
+        assert_eq!(forward(&mut engine, "intResult"), "abcde");
+        assert_eq!(
+            forward(&mut engine, "exceptional1"),
+            "The quick brown fox jumps over the lazy dog. If the dog reacted, was it really lazy?"
+        );
+        assert_eq!(forward(&mut engine, "exceptional2"), "The quick brown fox jumps over the lazy undefined. If the undefined reacted, was it really lazy?");
```
IovoslavIovchev:
```suggestion
        assert_eq!(forward(&mut engine, "exceptional2"), "The quick brown fox jumps over the lazy undefined. If the dog reacted, was it really lazy?");
```

src/lib/builtins/string.rs
```diff
@@ -1041,4 +1097,41 @@ mod tests {
...
+        assert_eq!(forward(&mut engine, "exceptional2"), "The quick brown fox jumps over the lazy undefined. If the undefined reacted, was it really lazy?");
```
IovoslavIovchev: Remove this line, it's a duplicate

IovoslavIovchev:
The current implementation replaces all occurrences of the first argument when it's a `String`, when, in fact, it should only replace the first occurrence. See [the spec](https://tc39.es/ecma262/#sec-string.prototype.replace) for more info.

src/lib/builtins/string.rs
thiagoarrais: I actually wonder if this _already_ exists somewhere. But I can't find it...

IovoslavIovchev: I know something similar exists in _regexp.rs_ but nothing else.

IovoslavIovchev:
@thiagoarrais do you need any help with this?

thiagoarrais:
I think I've addressed most of your comments, @IovoslavIovchev . The one thing I couldn't tackle yet is the spec breakage regarding replacing only the first occurrence. I plan on getting back to this some time, but I'm a little short on time. If you could help on that, it would be greatly appreciated.

Maybe you can give some general direction on a good approach? I'd also include a commit of yours here if you're willing to branch from this PR.

IovoslavIovchev:
@thiagoarrais no worries. I will take a look.

jasonwilliams:
This has now been added

PR #191, opened 2019-10-25
[WIP] - addition of Symbols

jasonwilliams:


src/lib/builtins/symbol.rs
```diff
@@ -0,0 +1,65 @@
...
+    let mut sym_instance = Object::default();
+
+    // Set description which should either be undefined or a string
+    let desc_string = match args.get(0) {
+        Some(value) => to_value(value.to_string()),
+        None => Gc::new(ValueData::Undefined),
+    };
+
+    sym_instance.set_internal_slot("Description", desc_string);
+    sym_instance.set_internal_slot("SymbolData", to_value(random::<i32>()));
```
IovoslavIovchev: What is this slot used for?

jasonwilliams: All symbols need to be unique and have some sort of ID.
Symbols can be used as field names in objects so the SymbolData number would be used there in sym_properties

src/lib/builtins/console.rs
```diff
@@ -94,6 +94,19 @@ fn log_string_from(x: Value) -> String {
             }
             s
         }
+        ValueData::Symbol(ref sym) => {
+            // Create empty formatted string to start writing to
+            let mut s = String::new();
+            let desc: Value = sym.borrow().get_internal_slot("Description");
+            match *desc {
```
IovoslavIovchev: Why don't you have the match without the `write!`s, so it directly returns the result of the `format!`? This way you also don't need `s`.

Is there a specific reason?

src/lib/builtins/object.rs
```diff
@@ -61,10 +62,11 @@ impl Object {
 
     /// Utility function to get an immutable internal slot or Null
     pub fn get_internal_slot(&self, name: &str) -> Value {
-        match self.internal_slots.get(name) {
+        let result = match self.internal_slots.get(name) {
```
IovoslavIovchev: Why was this added?

src/lib/builtins/array.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{
```
IovoslavIovchev: Are the changes in this file related to the implementation for `Symbols`, or are they just refactoring?

jasonwilliams: Both really.
I needed to do some refactoring around the contruction of arrays

jasonwilliams:
This is weird, these tests look cached or something, its all passing locally

src/lib/builtins/console.rs
```diff
@@ -94,6 +94,14 @@ fn log_string_from(x: Value) -> String {
             }
             s
         }
+        ValueData::Symbol(ref sym) => {
+            // Create empty formatted string to start writing to
```
IovoslavIovchev: This comment is not needed anymore

PR #192, opened 2019-10-26
Implement Array.prototype.forEach

xSke:


src/lib/builtins/array.rs
```diff
@@ -929,4 +961,27 @@ mod tests {
...
+    #[test]
+    fn for_each() {
+        let realm = Realm::create();
+        let mut engine = Executor::new(realm);
+        let init = r#"
+        var a = [2, 3, 4, 5];
+        var sum = 0;
+        var indexSum = 0;
+        var listLengthSum = 0;
+        function callingCallback(item, index, list) {
```
IovoslavIovchev: You could test using `this` inside the callback.

IovoslavIovchev: Actually, come to think of it, I am not sure `this` has been implemented. However, you could add a _TODO_ for when it is.

src/lib/builtins/array.rs
```diff
@@ -470,6 +470,37 @@ pub fn find_index(this: &Value, args: &[Value], interpreter: &mut Interpreter) -
...
+            "Missing argument for Array.prototype.forEach".to_string(),
+        ));
+    }
+
+    let callback_arg = args.get(0).expect("Could not get `callbackFn` argument.");
+
+    let this_arg = args
+        .get(1)
+        .cloned()
+        .unwrap_or_else(|| Gc::new(ValueData::Undefined));
```
bojan88: I think you can replace this line with `.unwrap_or(&Gc::new(ValueData::Undefined));

IovoslavIovchev: `clippy` is going to complain that you are not using `unwrap_or_else()`.

IovoslavIovchev: You can use the `undefined()` function, defined in _src/lib/builtins/value.rs_

src/lib/builtins/array.rs
```diff
@@ -470,6 +470,37 @@ pub fn find_index(this: &Value, args: &[Value], interpreter: &mut Interpreter) -
...
+
+    let length: i32 =
+        from_value(this.get_field_slice("length")).expect("Could not get `length` property.");
+
+    for i in 0..length {
+        let element = this.get_field(&i.to_string());
+        let arguments = vec![element.clone(), to_value(i), this.clone()];
+
+        interpreter.call(callback_arg, &this_arg, arguments)?;
```
letmutx: `callback` should not be called for undefined values. Also, can you please add tests for cases where array is modified in the callback? For ex. new elements are added to array, elements inserted in the middle, elements deleted.

MDN has a few examples on how this function should work in such cases [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).

IovoslavIovchev:
@xSke do you need any help on this?

src/lib/builtins/array.rs
jasonwilliams: Im not sure this is quite right, it does run for undefined values, just not values that have been deleted

jasonwilliams:
Replace by https://github.com/jasonwilliams/boa/pull/268 
Thanks @xSke

PR #196, opened 2019-10-31
Add Array.prototype.reduce method

letmutx:


src/lib/builtins/array.rs
```diff
@@ -998,4 +1031,31 @@ mod tests {
...
+    #[test]
+    fn reduce() {
+        let realm = Realm::create();
+        let mut engine = Executor::new(realm);
+        let init = r#"
+        function callback(acc, val) {
+            return 5;
+        }
+
+        var empty = [];
```
IovoslavIovchev: Add a test for calling `reduce` on an empty `Array` with no initial value.

src/lib/builtins/array.rs
```diff
@@ -534,6 +534,38 @@ pub fn slice(this: &Value, args: &[Value], interpreter: &mut Interpreter) -> Res
...
 }
 
+/// Array.prototype.reduce ( callback, [initialValue] );
+///
+/// The reduce() method executes the callback once for each assigned value
+/// present in the array
+/// <https://tc39.es/ecma262/#sec-array.prototype.reduce>
+pub fn reduce(this: &Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
+    let mut len: i32 = from_value(this.get_field_slice("length")).unwrap();
+    // TODO: Should throw TypeError if args.len() == 1 and length(arr) == 0
```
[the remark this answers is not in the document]
letmutx: I don't think exceptions handling is implemented yet.

src/lib/builtins/array.rs
letmutx: Done!

src/lib/builtins/array.rs
```diff
@@ -574,6 +606,7 @@ pub fn create_constructor(global: &Value) -> Value {
     array_prototype.set_field_slice("indexOf", index_of_func);
     array_prototype.set_field_slice("lastIndexOf", last_index_of_func);
     array_prototype.set_field_slice("slice", to_value(slice as NativeFunctionData));
+    array_prototype.set_field_slice("reduce", to_value(reduce as NativeFunctionData));
```
IovoslavIovchev: You forgot to set the `length` of the function

src/lib/builtins/array.rs
IovoslavIovchev: According to the spec:
> It is a TypeError if the array contains no elements and initialValue is not provided.

---

So, you can comment out the test and add a _TODO_ but the test _should not_ expect `undefined`.

src/lib/builtins/array.rs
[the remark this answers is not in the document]
jasonwilliams: Yeah we really need to get into that, I should start an issue

letmutx:
Tests break only after merging into master.

IovoslavIovchev:
@letmutx do you need any help with this?

letmutx:
Please re-assign to someone else!

Razican:
I think we should close this if it's not going to get implemented, right?

HalidOdat:
> I think we should close this if it's not going to get implemented, right?

I think so too.
This is stoping other people from implementing it!

PR #197, opened 2019-10-31
Fix early return

evomassiny:
This PR fixes https://github.com/jasonwilliams/boa/issues/195

Currently the `Interpreter` handles early returns by setting an `is_return` flag when it reaches a return expression. When executing a block expression, the interpreter will skip any sub-expressions if the flag is set, and then unset it.

This bug occurred when the return expression was located inside a block nested in the function's one,
to fix this, this PR unset the `is_return` flag only when leaving the "root block" of a function.
I used the EnvironementType of the block environment to identify such "root blocks".

I hope this helps

### DIFF
--- a/src/lib/exec.rs
+++ b/src/lib/exec.rs
@@ -5,7 +5,7 @@ use crate::{
         value::{from_value, to_value, ResultValue, Value, ValueData},
     },
     environment::lexical_environment::{
-        new_declarative_environment, new_function_environment, VariableScope,
+        new_declarative_environment, new_function_environment, EnvironmentType, VariableScope,
     },
     realm::Realm,
     syntax::ast::{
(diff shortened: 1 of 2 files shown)


src/lib/exec.rs
```diff
@@ -86,15 +86,22 @@ impl Executor for Interpreter {
...
                     }
                     if e == es.last().expect("unable to get last value") {
                         obj = val;
                     }
                 }
 
-                self.realm.environment.pop();
+                // pop the block env
+                let block_env = self.realm.environment.pop();
+                // clear the early return flag `self.is_return`
```
IovoslavIovchev: Insert a new line before this comment

IovoslavIovchev:
Looks very good. Good job fixing it so quickly!

src/lib/exec.rs
evomassiny: done :)

jasonwilliams:
Thanks for this @evomassiny

PR #200, opened 2019-11-01
Fix early return (again)

evomassiny:
Hi,

I did not test my PR well enough yesterday, and this morning I've discovered a bug caused by it: when a function call is made inside a function body, a "return" in the nested call will also quit the outer function call.

I used the EnvironementType associated with a block to identify the end of a function call, it turns out that this approach doesn't work with nested functions call (I don't fully understand why, but in some cases the environment type associated with a function block could be "Declarative", not "Function").

To fix it, I've moved the `self.is_return = false` right after the function evaluation.

Sorry for the inconvenience

### DIFF
--- a/src/lib/exec.rs
+++ b/src/lib/exec.rs
@@ -5,7 +5,7 @@ use crate::{
         value::{from_value, to_value, ResultValue, Value, ValueData},
     },
     environment::lexical_environment::{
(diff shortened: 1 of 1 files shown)


IovoslavIovchev:
Good job catching this so quickly and thanks for raising.

PR #203, opened 2019-11-03
parse function arguments

jasonwilliams:
Move argument parsing into separate function to be used as a utility

### DIFF
--- a/src/lib/syntax/parser.rs
+++ b/src/lib/syntax/parser.rs
@@ -67,6 +67,32 @@ impl Parser {
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/lib/syntax/parser.rs
```diff
@@ -67,6 +67,32 @@ impl Parser {
         }
     }
 
+    fn parse_function_parameters(&mut self) -> Result<Vec<String>, ParseError> {
+        self.expect_punc(Punctuator::OpenParen, "function parameters ( expected")?;
+        let mut args: Vec<String> = Vec::new();
```
IovoslavIovchev: Can't the type be inferred?

Issue #151, opened 2019-10-12
Rest / Spread operator


jasonwilliams:
This is needed

robertmaloney:
Can I pick this up?

jasonwilliams:
Sure

jasonwilliams:
@robertmaloney everything ok on this?

robertmaloney:
Apologies for being slow with this. I already have function spread and working on wrapping arrow functions, and I think that'd be a good point to get a PR in.

jasonwilliams:
This is now working for arrays and function arguments, objects are still not supported but i will open a new ticket for that

PR #205, opened 2019-11-06
Rest params for function declarations

robertmaloney:
Part of #151 

### TODO
* [ ] tests for arrow functions
* [ ] rebase

### DIFF
--- a/src/lib/builtins/function.rs
+++ b/src/lib/builtins/function.rs
@@ -35,14 +35,14 @@ pub struct RegularFunction {
     pub object: Object,
     /// This function's expression
     pub expr: Expr,
-    /// The argument names of the function
-    pub args: Vec<String>,
+    /// The argument declarations of the function
+    pub args: Vec<Expr>,
 }
 
 impl RegularFunction {
     /// Make a new regular function
     #[allow(clippy::cast_possible_wrap)]
-    pub fn new(expr: Expr, args: Vec<String>) -> Self {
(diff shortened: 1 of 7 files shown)


src/lib/js/value.rs
```diff
@@ -592,7 +592,15 @@ impl Display for ValueData {
             ValueData::Function(ref v) => match *v.borrow() {
                 Function::NativeFunc(_) => write!(f, "function() {{ [native code] }}"),
                 Function::RegularFunc(ref rf) => {
-                    write!(f, "function({}){}", rf.args.join(", "), rf.expr)
+                    write!(f, "function(")?;
+                    let last_index = rf.args.len() - 1;
+                    for (index, arg) in rf.args.iter().enumerate() {
```
IovoslavIovchev: This loop can be rewritten using `.iter().join()`.

robertmaloney: An attempt was made, but I'm having trouble converting `rf.args` into a format that supports `join`.

jasonwilliams:
@robertmaloney this is great, do you know why the tests are failing?

jasonwilliams:
Fixed test,
Looks like this doesn't handle spread syntax as an argument to a function, looking into that

jasonwilliams:
This has already been implemented
https://github.com/jasonwilliams/boa/blob/master/src/lib/exec.rs#L823-L848

Thanks anyway @robertmaloney !

robertmaloney:
Happy to see it taken care of, thanks for closing!

Issue #193, opened 2019-10-27
Create a macro for properly setting builtin member functions


IovoslavIovchev:
When creating a new member function for some `builtin`, you need to make sure the `create_constructor()` function contains this:
```rust
let some_func = to_value(func as NativeFunctionData);
some_func.set_field_slice("length", to_value(1_i32));

// .. and then
some_prototype.set_field_slice("someFunc", some_func); 
``` 

Writing these two lines for each new function can prove tedious _and_ error prone, as the second line can be missed when needed ([functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) have their `length` property reflect the number of arguments they accept).

What I propose is the following macro:
```rust
make_fn!(some_func, named "someFunc", with length 1, of some_prototype);
```
which will essentially contain the 3 aforementioned boilerplate lines.

What I am saying should definitely not be set in stone. It is only supposed to start a discussion on the topic.

jasonwilliams:
This is a good idea, although i wonder if it will overlap [Function Objects ](https://tc39.es/ecma262/#sec-createbuiltinfunction)

IovoslavIovchev:
Depends on what you mean by "functions we come across in code".
My intention is for it to only be used for reducing boilerplate in _buil-tins_.

It is a more humble version of `CreateBuiltinFunction` (of which I found no implementation, apologies if I am missed anything). Of course, we could expand the scope a bit here and fully implement `CreateBuiltinFunction`.

IovoslavIovchev:
Great. I believe it will be a good first issue.

Will mark it as such, if anyone is interested.

Stupremee:
I can take care of this

Stupremee:
This issue can be closed I guess

IovoslavIovchev:
Done as part of #206

thiagoarrais:
I'm getting `undefined` when I do `'abc'.matchAll.length` in `tests.js`. Even after upgrading to 7698873. Am I doing something wrong?

Stupremee:
I think a new issue for your issue would be better @thiagoarrais

thiagoarrais:
Sure!

PR #206, opened 2019-11-09
Implement macro for setting builtin functions

Stupremee:
# Description
This pull request introduces a new macro called `make_fn` in the builtins module,
which can be used to simplify setting a builtin function and replaces every "old" way with the new macro.

# Example
```rust
let some_func = to_value(func as NativeFunctionData);
some_func.set_field_slice("length", to_value(1_i32));
some_prototype.set_field_slice("someFunc", some_func); 
```
becomes
```rust
make_fn!(some_func, named "someFunc", with length 1, of some_prototype);
```
and
```rust
some_prototype.set_field_slice("someFunc", (some_func as NativeFunctionData)); 
```
becomes
```rust
make_fn!(some_func, named "someFunc", of some_prototype);
```

This change was requested #193


jasonwilliams:
Do we want to call this macro make_builtin_fn so it’s explicit that’s what it’s for? 
As we can also make dynamic functions (js code) later on.

Stupremee:
Why does some functions, e.g. `findIndex`or `find`, in the array primitive doesn't set the `length` field?

IovoslavIovchev:
@Stupremee basically, that is where the idea for this macro came from. Some functions do not set the `length` when they should simply because it is something that is easy to be missed during implementation.

Stupremee:
Ah okay. So I should set the `length` field for every method that doesn't set it?

src/lib/builtins/mod.rs
```diff
@@ -1,3 +1,15 @@
+/// Macro to create a new member function of a prototype
+macro_rules! make_builtin_fn {
+    ($fn:ident, named $name:expr, with length $l:tt, of $p:ident) => {
+        let $fn = to_value($fn as NativeFunctionData);
+        $fn.set_field_slice("length", to_value($l));
+        $p.set_field_slice($name, $fn);
+    };
+    ($fn:ident, named $name:expr, of $p:ident) => {
+        $p.set_field_slice($name, to_value($fn as NativeFunctionData));
```
IovoslavIovchev: Are there any functions that do not have their `length` set (e.g. it's `undefined`)? If so, this line is fine.

But if not, I think it would be better if you set the `length` to `0` by default.

IovoslavIovchev:
@Stupremee yep

src/lib/builtins/math.rs
```diff
@@ -203,24 +203,24 @@ pub fn create_constructor(global: &Value) -> Value {
...
-    math.set_field_slice("random", to_value(_random as NativeFunctionData));
-    math.set_field_slice("round", to_value(round as NativeFunctionData));
-    math.set_field_slice("sin", to_value(sin as NativeFunctionData));
-    math.set_field_slice("sqrt", to_value(sqrt as NativeFunctionData));
-    math.set_field_slice("tan", to_value(tan as NativeFunctionData));
+    make_builtin_fn!(abs, named "abs", with length 1, of math);
+    make_builtin_fn!(acos, named "acos", with length 1, of math);
+    make_builtin_fn!(asin, named "asin", with length 1, of math);
+    make_builtin_fn!(atan, named "atan", with length 1, of math);
+    make_builtin_fn!(atan2, named "atan2", with length 1, of math);
```
IovoslavIovchev: The `length` should be 2.

IovoslavIovchev:
```suggestion
    make_builtin_fn!(atan2, named "atan2", with length 2, of math);
```

src/lib/builtins/number.rs
```diff
@@ -145,18 +145,12 @@ pub fn create_constructor(global: &Value) -> Value {
...
-    );
-    number_prototype.set_field_slice("toFixed", to_value(to_fixed as NativeFunctionData));
-    number_prototype.set_field_slice(
-        "toLocaleString",
-        to_value(to_locale_string as NativeFunctionData),
-    );
-    number_prototype.set_field_slice("toPrecision", to_value(to_precision as NativeFunctionData));
-    number_prototype.set_field_slice("toString", to_value(to_string as NativeFunctionData));
-    number_prototype.set_field_slice("valueOf", to_value(value_of as NativeFunctionData));
+    make_builtin_fn!(to_exponential, named "toExponential", of number_prototype);
```
IovoslavIovchev:
```suggestion
    make_builtin_fn!(to_exponential, named "toExponential", with length 1, of number_prototype);
```

src/lib/builtins/number.rs
```diff
@@ -145,18 +145,12 @@ pub fn create_constructor(global: &Value) -> Value {
...
+    make_builtin_fn!(to_fixed, named "toFixed", with length 1, of number_prototype);
+    make_builtin_fn!(to_locale_string, named "toLocaleString", of number_prototype);
+    make_builtin_fn!(to_precision, named "toPrecision", with length 1, of number_prototype);
+    make_builtin_fn!(to_string, named "toString", of number_prototype);
```
IovoslavIovchev:
```suggestion
    make_builtin_fn!(to_string, named "toString", with length 1, of number_prototype);
```

src/lib/builtins/string.rs
```diff
@@ -744,28 +744,28 @@ pub fn create_constructor(global: &Value) -> Value {
...
+    make_builtin_fn!(concat, named "concat", with length 1, of proto);
+    make_builtin_fn!(repeat, named "repeat", with length 1, of proto);
+    make_builtin_fn!(slice, named "slice", with length 2, of proto);
+    make_builtin_fn!(starts_with, named "startsWith", with length 1, of proto);
+    make_builtin_fn!(ends_with, named "endsWith", with length 1, of proto);
+    make_builtin_fn!(includes, named "includes", with length 1, of proto);
+    make_builtin_fn!(index_of, named "indexOf", with length 1, of proto);
+    make_builtin_fn!(last_index_of, named "lastIndexOf", with length 1, of proto);
+    make_builtin_fn!(r#match, named "match", with length 1, of proto);
+    make_builtin_fn!(pad_end, named "padEnd", with length 2, of proto);
```
IovoslavIovchev:
```suggestion
    make_builtin_fn!(pad_end, named "padEnd", with length 1, of proto);
```

src/lib/builtins/string.rs
```diff
@@ -744,28 +744,28 @@ pub fn create_constructor(global: &Value) -> Value {
...
+    make_builtin_fn!(pad_start, named "padStart", with length 2, of proto);
```
IovoslavIovchev:
```suggestion
    make_builtin_fn!(pad_start, named "padStart", with length 1, of proto);
```

src/lib/builtins/mod.rs
```diff
@@ -1,3 +1,16 @@
+/// Macro to create a new member function of a prototype
+/// If no length is provided, the length will be set to 0.
+macro_rules! make_builtin_fn {
+    ($fn:ident, named $name:expr, with length $l:tt, of $p:ident) => {
```
IovoslavIovchev: Is there a reason why `$l` is a `tt`?

Stupremee: Because it’s the best way to represent numbers in macros

PR #215, opened 2019-12-07
WIP: Create constant values via the prototype constructor

### LINKED ISSUE #210: builtinfun.length undefined
I'm getting `undefined` when I do `'abc'.matchAll.length`. This is incompatible with Chrome's implementation (for example), where I get `1`.

Testing with master with the following `tests.js`:

```js
console.log('abc'.matchAll.length);
```

Here's the output for `cargo run`:

```
$ cargo run
   Compiling Boa v0.5.0 (/data/boa)
    Finished dev [unoptimized + debuginfo] target(s) in 3.20s
     Running `target/debug/boa`
undefined
[src/bin/bin.rs:57] exec(&buffer) = "undefined"
```

Related to #193 and #206

Stupremee:
This Pull Request changes the evaluation of `Const` values (Num, Bool, String) to use the specific prototype constructor to set all required properties and methods.

Resolves #210

### DIFF
--- a/src/lib/exec.rs
+++ b/src/lib/exec.rs
@@ -66,13 +66,44 @@ impl Executor for Interpreter {
         match expr.def {
             ExprDef::Const(Const::Null) => Ok(to_value(None::<()>)),
             ExprDef::Const(Const::Undefined) => Ok(Gc::new(ValueData::Undefined)),
-            ExprDef::Const(Const::Num(num)) => Ok(to_value(num)),
+            ExprDef::Const(Const::Num(num)) => {
+                let value = to_value(num);
+                let proto = self
+                    .realm
+                    .environment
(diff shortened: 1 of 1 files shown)


Stupremee:
I ran into more issues, if a number value gets converted to an object, the `to_num` method doesn't work and so all `BinOp` operations are not working correctly.

jasonwilliams:
@Stupremee won't this create Object instances instead of native types?
```js
let a = 1;
typeof 1; // object
```

on your branch

Stupremee:
Yep, that's the problem I didn't think about that earlier.

Razican:
Hello, what is the state of this?

jasonwilliams:
I'm going to close this PR as its quite outdated now

PR #217, opened 2020-01-11
String.prototype.replace()

jasonwilliams:


src/lib/builtins/string.rs
```diff
@@ -322,6 +324,117 @@ pub fn includes(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
     Ok(to_value(this_string.contains(&search_string)))
 }
 
+/// Return either the string itself or the string of the regex equivalent
+fn get_regex_string(value: &Value) -> String {
+    let mut regex_body = String::new();
```
IovoslavIovchev: You don't need this, you can return directly from the `match`

PR #218, opened 2020-01-12
feat: Implement Array.prototype.toString

cisen:
I am a rust newbie and my code may not be good. But I will continue to improve this code.

### DIFF
--- a/src/lib/builtins/array.rs
+++ b/src/lib/builtins/array.rs
@@ -211,7 +211,7 @@ pub fn to_string(this: &Value, _args: &[Value], _ctx: &mut Interpreter) -> Resul
             ValueData::String(ref s) => (*s).clone(),
             _ => "".to_string(),
         },
-        Err(v) => format!("{}: {}", "error", v.to_string()),
+        Err(v) => format!("error: {}", v),
     };
     Ok(to_value(match_string))
 }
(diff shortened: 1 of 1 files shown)


src/lib/builtins/array.rs
```diff
@@ -198,6 +198,24 @@ pub fn join(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
...
+/// other kinds of objects for use as a method.
+/// <https://tc39.es/ecma262/#sec-array.prototype.tostring>
+pub fn to_string(this: &Value, _args: &[Value], _ctx: &mut Interpreter) -> ResultValue {
+    let join_result = join(this, &[to_value(",")], _ctx);
+    let match_string = match join_result {
+        Ok(v) => match *v {
+            ValueData::String(ref s) => (*s).clone(),
+            _ => "".to_string(),
+        },
+        Err(v) => format!("{}: {}", "error", v.to_string()),
```
IovoslavIovchev: No need for `"error"` to be a separate param, you can change the template to `"error: {}"`.

Also, no need for `.to_string()`, `v` will be formatted accordingly to its value anyway.

IovoslavIovchev:
Looks good, sync you branch with master so the pipeline runs, and we are all set

PR #220, opened 2020-01-13
[README.md] Adjust headings importance and update Usage section

croraf:
Adjust headings in readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,8 @@ see [CHANGELOG](./CHANGELOG.md)
 
 ## Usage
 
-- Clone this repo and position yourself in its folder root
-- Create a JS file that will be executed, for example `test.js`, and fill it with some JS code
-- Build and run with `cargo run -- test.js`
+- Clone this repo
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,22 +64,21 @@ See Milestones
...
 
-### Usage
+## Usage
 
-- Check out this project
-- Build `cargo build`
-- `cargo run`
-- You can make changes to tests/js/test.js and build again
-- If any JS doesn't work then it's a bug! Please raise an issue
+- Clone this repo and position yourself in its folder root
```
IovoslavIovchev: I think that being in the root of the project is usually implied, so you can remove this

README.md
```diff
@@ -64,22 +64,21 @@ See Milestones
...
+- Create a JS file that will be executed, for example `test.js`, and fill it with some JS code
```
IovoslavIovchev: IMO, you can change this to something like
```suggestion
- Run with `cargo run -- test.js`, where `test.js` is an existing JS file
```

PR #222, opened 2020-01-14
Fixed clippy errors

IovoslavIovchev:


jasonwilliams:
Thanks @IovoslavIovchev 
I don't know why these became a problem now

PR #227, opened 2020-01-19
fix: Array.prototype.toString should be called by ES value

cisen:


src/lib/builtins/array.rs
```diff
@@ -205,7 +205,34 @@ pub fn join(this: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
...
 /// <https://tc39.es/ecma262/#sec-array.prototype.tostring>
 pub fn to_string(this: &Value, _args: &[Value], _ctx: &mut Interpreter) -> ResultValue {
-    let join_result = join(this, &[to_value(",")], _ctx);
+    let method_name = "join";
+    let arguments = vec![to_value(",")];
+    // 2.
+    let method: Value =
+        from_value(this.get_field_slice(method_name)).expect("failed to get Array.prototype.join");
+    // 3.
+    if !method.is_function() {
```
IovoslavIovchev: Wouldn't it be better if you changed the value of `method` and remove the duplicated logic from the `if` statement?

IovoslavIovchev:
Looks good

PR #229, opened 2020-01-20
Replace error throwing with panic in "Lexer::next()"

croraf:
Am not sure about those macros at the top of the file. I just removed ? in three of them that were throwing an error.

### DIFF
--- a/src/lib/syntax/lexer.rs
+++ b/src/lib/syntax/lexer.rs
@@ -18,7 +18,7 @@ macro_rules! vop {
         let preview = $this.preview_next().unwrap();
         match preview {
             '=' => {
-                $this.next()?;
+                $this.next();
                 $assign_op
             }
             _ => $op,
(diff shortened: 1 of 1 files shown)


src/lib/syntax/lexer.rs
```diff
@@ -143,18 +143,18 @@ impl<'a> Lexer<'a> {
     }
 
     /// next fetches the next token and return it, or a LexerError if there are no more.
-    fn next(&mut self) -> Result<char, LexerError> {
+    fn next(&mut self) -> char {
         match self.buffer.next() {
-            Some(ch) => Ok(ch),
-            None => Err(LexerError::new("finished")),
+            Some(ch) => ch,
+            None => panic!(),
```
jasonwilliams: Could we put a message in here?
Like "More more characters to consume from input stream"

croraf: Done. I also tested (on single line comment and on `/*asdads`) that when this is not satisfied that it panics correctly.

PR #231, opened 2020-01-21
Fix lexing of "0_" token

croraf:
Several changes regarding 0_ token.
- preview_next None check before analyzing the x,b,o token (because it can be end of stream)
- Addition of 0X and 0B along with 0x 0b.
- replace of if/else with match
- commenting-out non-strict legacy octal token, which errors out in strict mode
- previous bullet will error out on the last part added which is 11.8.3 clause from spec

### DIFF
--- a/src/lib/syntax/lexer.rs
+++ b/src/lib/syntax/lexer.rs
@@ -322,52 +322,83 @@ impl<'a> Lexer<'a> {
                 }
                 '0' => {
                     let mut buf = String::new();
-                    let num = if self.next_is('x') {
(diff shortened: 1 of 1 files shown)


croraf:
Regarding failed tests: syntax::lexer::tests::numbers fails due to non-strict assumption in specific test suite. Why others fail I dont know

croraf:
I fixed the issues I had with my PR. 

- I made a small refactor of the 0_ code (extracted some stuff). 
- Addition of 0X and 0B along with 0x 0b.
- I also covered some stuff more explicitly hopefully solving some undetected bugs

Issue #209, opened 2019-11-17
Calling `new Array` with 1 argument doesn't work properly


IovoslavIovchev:
The [spec](https://tc39.es/ecma262/#sec-array-len) describes that `new Array(5)` should create an array with 5 empty elements/slots. Currently, it produces `[5]`.

nasso:
I'm working on it!

nasso:
@nasso Woops nevermind didn't see the above PR!

PR #233, opened 2020-01-23
fix: 209 Calling `new Array` with 1 argument doesn't work properly

cisen:
#209

### DIFF
--- a/src/lib/builtins/array.rs
+++ b/src/lib/builtins/array.rs
@@ -85,19 +85,15 @@ pub fn make_array(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result
     // between indices and values): this creates an Object with no prototype
 
     // Create length
-    let length = if args.len() > 1 {
-        Property::new()
-            .value(to_value(args.len() as i32))
-            .writable(true)
-            .configurable(false)
-            .enumerable(false)
-    } else {
-        Property::new()
-            .value(to_value(args[0].clone()))
-            .writable(true)
-            .configurable(false)
-            .enumerable(false)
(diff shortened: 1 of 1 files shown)


src/lib/builtins/array.rs
```diff
@@ -85,11 +85,19 @@ pub fn make_array(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result
     // between indices and values): this creates an Object with no prototype
 
     // Create length
-    let length = Property::new()
-        .value(to_value(args.len() as i32))
-        .writable(true)
-        .configurable(false)
-        .enumerable(false);
+    let length = if args.len() > 1 {
```
IovoslavIovchev: You can extract just the length and write the logic for creating the property once, e.g.
```rust
let n = if args.len() > 1 { args.len() } ...
```

cisen:
> LGTM! Maybe a `match` statement would be a better fit instead of the `if` statements and would also allow you to recapture the value inside the pattern match.
> Not a big change but would make it easier to read and express intent better IMO.

Thanks for your advice! I will improve it later. Because of Chinese New Year and my hometown's slow network，It will take a little time.

IovoslavIovchev:
Looks good but can you look for some commented tests for this or add some if they don't exist?

cisen:
> Looks good but can you look for some commented tests for this or add some if they don't exist?

Ok, I will add it later.

cisen:
There is still some problems when I add tests like [jerryscripts'](https://github.com/jerryscript-project/jerryscript/blob/master/tests/jerry/array.js). It will will take more times to fix this problems.

PR #241, opened 2020-02-02
Made "structopt" dependency optional under "cli" feature flag

### LINKED ISSUE #239: Make "structopt" dependency optional
The `structopt` dependency seems to be only used for the binaries. It would be nice if it could be disabled if this engine was used as a library. Maybe it could be put behind a `cli` feature flag or similar and enabled for the binaries only?

adumbidiot:
I made structopt optional and put it under a "cli" feature. I left it on by default, so users have to turn off default features to deactivate it. I believe this also makes it so no changes are needed when running the binary.

Resolves #239

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,8 @@ edition = "2018"
 default-run = "boa"
 
 [features]
-default = ["wasm-bindgen"]
+default = ["wasm-bindgen", "cli"]
+cli = ["structopt"]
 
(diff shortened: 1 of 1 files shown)


jasonwilliams:
Going to hold this for now due to https://github.com/jasonwilliams/boa/issues/239#issuecomment-581846565

adumbidiot:
Closing due to #247

PR #242, opened 2020-02-03
Fixed some panics in the lexer

adumbidiot:
I went through and fixed some of the immediately obvious panics in the lexer, though more work still needs to be done. I removed all `unwrap` calls and replaced them with `expect`. I also made `parser_expr` return a `Result<Expr, String>` and edited `forward` and `forward_val` accordingly.

### DIFF
--- a/src/lib/lib.rs
+++ b/src/lib/lib.rs
@@ -10,27 +10,34 @@ pub mod syntax;
 #[cfg(feature = "wasm-bindgen")]
 mod wasm;
 
+#[cfg(feature = "wasm-bindgen")]
+pub use crate::wasm::*;
 use crate::{
     builtins::value::ResultValue,
     exec::{Executor, Interpreter},
     realm::Realm,
     syntax::{ast::expr::Expr, lexer::Lexer, parser::Parser},
 };
-#[cfg(feature = "wasm-bindgen")]
-pub use wasm::*;
 
-fn parser_expr(src: &str) -> Expr {
+fn parser_expr(src: &str) -> Result<Expr, String> {
     let mut lexer = Lexer::new(src);
-    lexer.lex().expect("lexing failed");
+    lexer.lex().map_err(|e| e.to_string())?;
     let tokens = lexer.tokens;
-    Parser::new(tokens).parse_all().expect("parsing failed")
+    Parser::new(tokens)
+        .parse_all()
+        .map_err(|_e| "parsing failed".to_string())
 }
 
 /// Execute the code using an existing Interpreter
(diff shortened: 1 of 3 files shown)


adumbidiot:
Is there something wrong with the `rustfmt` config for this repo? My rustfmt didn't appear to work while I was working on this and the CI didn't catch some bad indentations.

jasonwilliams:
Are you running the same version of rust as the CI? (rustc 1.41.0) ?
Also you can try `cargo clean` which usually helps

adumbidiot:
Yeah I'm using 1.41.0. `cargo clean` doesn't help.

src/lib/lib.rs
```diff
@@ -10,27 +10,34 @@ pub mod syntax;
...
     syntax::{ast::expr::Expr, lexer::Lexer, parser::Parser},
 };
-#[cfg(feature = "wasm-bindgen")]
-pub use wasm::*;
 
-fn parser_expr(src: &str) -> Expr {
+fn parser_expr(src: &str) -> Result<Expr, String> {
     let mut lexer = Lexer::new(src);
-    lexer.lex().expect("lexing failed");
+    lexer.lex().map_err(|e| e.to_string())?;
```
jasonwilliams: I would have this say SyntaxError: {e.to_string()}

src/lib/lib.rs
```diff
@@ -10,27 +10,34 @@ pub mod syntax;
...
     let tokens = lexer.tokens;
-    Parser::new(tokens).parse_all().expect("parsing failed")
+    Parser::new(tokens)
+        .parse_all()
+        .map_err(|_e| "parsing failed".to_string())
```
jasonwilliams: Then this can say ParsingError: {e.to_string()}

src/lib/lib.rs
```diff
@@ -10,27 +10,34 @@ pub mod syntax;
...
 
 /// Execute the code using an existing Interpreter
 /// The str is consumed and the state of the Interpreter is changed
 pub fn forward(engine: &mut Interpreter, src: &str) -> String {
     // Setup executor
-    let expr = parser_expr(src);
+    let expr = match parser_expr(src) {
+        Ok(v) => v,
+        Err(s) => {
+            return format!("{}: {}", "Error", s);
```
jasonwilliams: Then "Error" can be removed from here, and instead just return `s`

src/lib/lib.rs
adumbidiot: I had to give `ParseError` a basic `Display` impl, but I figured that it was needed eventually and should be able to be customized better later.

PR #244, opened 2020-02-05
Fixed more Lexer Panics

adumbidiot:
I tried fixing more lexer panics. I think I got most of them. I think the errors could be better, but I think we should consider a new `LexerError` that includes file positions later on.


jasonwilliams:
You able to rebase?

adumbidiot:
Done

adumbidiot:
Is that CI error my fault?

jasonwilliams:
no, i have no idea why thats failing now

PR #245, opened 2020-02-05
Fixed parsing of floats with scientific notation

### LINKED ISSUE #204: Number(<float>e<int>) should work
It looks like scientific notation numbers starting with a float compile to either `undefined` (positive float base) and when used in Number() triggers a panic.

For example:

```js
Number(1e1) // Works
Number(-1e1) // Works
Number(1e-1) // Works
Number(-1e-1) // Works
Number(1.0e1) // Fails
Number(-1.0e1) // Fails
Number(1.0e-1) // Fails
Number(-1.0e-1) // Fails
```

This behavior works in Node.js as expected:

```js
> Number(1.0e1)
10
> Number(1.0e-1)
0.1
> Number(-1.0e1)
-10
> Number(-1.0e-1)
-0.1
```

`Number()` is able to parse String inputs correctly, but raw scientific notation numbers with a float base trigger a panic. This may be a parsing issue as the errors look like this:

```
$ cargo run # Parsing `Number(1.0e1)`
     Running `target/debug/boa`
thread 'main' panicked at 'parsing failed: Expected([Punctuator(Comma), Punctuator(CloseParen)], Token { data: Identifier("e1"), pos: Position { column_number: 9, line_number: 5 } }, "function call arguments")', src/libcore/result.rs:1084:5
```

refs #182 implementing #34

adumbidiot:
Fixes scientific notation with floats. Resolves #204

### DIFF
--- a/src/lib/syntax/lexer.rs
+++ b/src/lib/syntax/lexer.rs
@@ -1016,7 +1016,9 @@ mod tests {
 
     #[test]
     fn numbers() {
-        let mut lexer = Lexer::new("1 2 0x34 056 7.89 42. 5e3 5e+3 5e-3 0b10 0O123 0999");
+        let mut lexer = Lexer::new(
+            "1 2 0x34 056 7.89 42. 5e3 5e+3 5e-3 0b10 0O123 0999 1.0e1 1.0e-1 1.0E1 1E1",
+        );
         lexer.lex().expect("failed to lex");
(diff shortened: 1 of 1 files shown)


adumbidiot:
The number parser has a lot of bugs and probably should be rewritten in the future

jasonwilliams:
@adumbidiot  thanks for this, yes i think the parsing around digits needs some work

Issue #140, opened 2019-10-08
Streamline Boa Shell into the same binary with arguments


jasonwilliams:
We have 2 binaries at the moment, Boa and Boa Shell.
Boa Shell is the debugging shell which will eventually be used in test262 https://github.com/jasonwilliams/boa/issues/74

Its possible to use boa with a shell option instead -s

This issue is for that

Arrises from:
https://github.com/jasonwilliams/boa/pull/107#issuecomment-533140009

jasonwilliams:
Closing superseeded by https://github.com/jasonwilliams/boa/issues/239

PR #247, opened 2020-02-05
Moved to a workspace architecture

### LINKED ISSUE #239: Make "structopt" dependency optional

Razican:
This fixes #239, and superseeds #241.

I also took the opportunity to generate the `Cargo.lock` file with the new Rust 1.41+ format, and add a couple of build profiles. Everything seems to be working fine, except for the wasm feature.

When building it like that, I seem to be getting a `./pkg/` directory, with an empty `index.js` file. This `./pkg/` directory seems not to be ignored by gitignore. Do you know how this works? I'm not familiarized with Rust-wasm setups.


adumbidiot:
Was this PR supposed to address #140 too?

adumbidiot:
Looks like `wasm-pack` [doesn't support workspaces](https://github.com/rustwasm/wasm-pack/issues/642).

adumbidiot:
I take that back. We can manually specify the crate to wasmify in the webpack plugin config and change the outdir. Reference: https://github.com/wasm-tool/wasm-pack-plugin.

Razican:
> Was this PR supposed to address #140 too?

Actually, not, but let's do that too!

> I take that back. We can manually specify the crate to wasmify in the webpack plugin config and change the outdir. Reference: [wasm-tool/wasm-pack-plugin](https://github.com/wasm-tool/wasm-pack-plugin).
> 
> The `pkg` dir was empty before because wasm-pack failed.

I will check that asap and update the PR :) thanks for the pointer!

Razican:
The build error is due to the binary being run in the benchmarks by default (where criterion is not available). This can be solved by passing `-p Boa` to `cargo bench`, done in the criterion compare action (https://github.com/jasonwilliams/criterion-compare-action/blob/master/main.js#L14-L18).

This should probably be added to the benchmarks GitHub action too, and it should be generic enough to accept an input "package" parameter.

jasonwilliams:
Seems ok from what i can see, unfortunately Rust Analyzer seems to fail now things are in workspaces, i don't know if there's a change i need to do there :/

Razican:
Checking at the output of the action run, it seems that the change is not being run, it's still running `cargo bench -- --save-baseline changes` :/

It's like it had some cache for the action. I'm guessing that [versioning][action-versioning] it should help. I will add the commit hash, to check it just in case.

[action-versioning]: https://github.com/actions/toolkit/blob/master/docs/action-versioning.md

Razican:
It seems it doesn't work. It's using the minified version in `dist/`, so that needs to be re-generated.

jasonwilliams:
That makes sense, I’ll try it again when I have time

jasonwilliams:
built and updated action

jasonwilliams:
seems to be weirdly cached, im going to try referencing the commit

jasonwilliams:
Looks like we've finally fixed it @Razican

Razican:
Cool! we can merge then. Still, it's weird it gets cached. We can maybe try switching to the master branch in the future :)

PR #254, opened 2020-02-19
Initial template literal lexer implementation

croraf:
Starting to implement. Would like this to be merged now.

### DIFF
--- a/boa/src/syntax/ast/token.rs
+++ b/boa/src/syntax/ast/token.rs
@@ -58,6 +58,8 @@ pub enum TokenData {
     Punctuator(Punctuator),
     /// A string literal
     StringLiteral(String),
+    /// Template litteral
+    TemplateLiteral(String),
     /// A regular expression, consisting of body and flags
     RegularExpressionLiteral(String, String),
     /// A comment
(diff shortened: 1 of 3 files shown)


croraf:
I also slipped in a move of self.next() into the match of the lexer itself.

1 line of this I adjusted in first commit, and the rest in the second commit.

croraf:
So this PR consists of two separate things. 1 is start of implementation of template literal, 2 is putting the self.next inside the match construct.

jasonwilliams:
@croraf this would need some tests to show it working

croraf:
I added 2 tests.

jasonwilliams:
@croraf thanks, I’ll have a play with this

croraf:
This will be split in 2 PRs and closed.

Razican:
I think the template lexing has been implemented in #455, right? Can we close this?

croraf:
I wanted to do a minor refactor of lexer that is present as part of this PR but otherwise the PR is obsolete.

croraf:
From this only remains the small `match` construct refactor in main lexer loop. Instead of getting the next character before `match` I put it as a part of the `match`. Check boa/src/syntax/lexer/mod.rs in the changed files.

If you like it I will make a separate PR with it?

Razican:
> If you like it I will make a separate PR with it?

All that will soon change with #432. I think it's best to invest on that by doing PRs to that branch :)

croraf:
@Razican I thought so. I will close this then.

PR #255, opened 2020-02-20
Refactor old function with new function object

### LINKED ISSUE #141: Implement Function Objects

jasonwilliams:
Fixes #141 

- [x] Implement fmt::Display for new function
- [x] Fix isFunction()
- [x] Better documentation for functions
- [x] Fall back to construct if call is not defined
- [x] all functions point to Function.prototype internally
- [x] leaner make_constructer_fn!()
- [x] deprecation of set_internal_method()
- [x] deprecation of set_method()
- [x] construct / call duplication
- [x] typeof is both a Node and a method in Value

### DIFF
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -1,134 +1,327 @@
-//! This module implements the global `Function` object.
+//! This module implements the global `Function` object as well as creates Native Functions.
 //!
-//! `Every JavaScript `function` is actually a `Function` object.
+//! Objects wrap `Function`s and expose them via call/construct slots.
+//!
+//! `The `Function` object is used for matching text with a pattern.
 //!
 //! More information:
 //!  - [ECMAScript reference][spec]
 //!  - [MDN documentation][mdn]
 //!
-//! [spec]: https://tc39.es/ecma262/#sec-function-object
+//! [spec]: https://tc39.es/ecma262/#sec-function-objects
 //! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function
-
-#[cfg(test)]
-mod tests;
 
 use crate::{
     builtins::{
-        object::{internal_methods_trait::ObjectInternalMethods, Object},
+        array,
+        object::{Object, ObjectInternalMethods, ObjectKind, PROTOTYPE},
         property::Property,
         value::{to_value, ResultValue, Value, ValueData},
     },
-    exec::Interpreter,
+    environment::lexical_environment::{new_function_environment, Environment},
+    exec::Executor,
     syntax::ast::node::{FormalParameter, Node},
+    Interpreter,
 };
-use gc::{custom_trace, Gc};
+use gc::{unsafe_empty_trace, Gc, Trace};
 use gc_derive::{Finalize, Trace};
 use std::fmt::{self, Debug};
-use std::ops::Deref;
 
-/// fn(this, arguments, ctx)
-pub type NativeFunctionData = fn(&Value, &[Value], &mut Interpreter) -> ResultValue;
+/// _fn(this, arguments, ctx) -> ResultValue_ - The signature of a built-in function
+pub type NativeFunctionData = fn(&mut Value, &[Value], &mut Interpreter) -> ResultValue;
 
-/// A Javascript `Function` object instance.
-///
-/// A member of the Object type that may be invoked as a subroutine
-///
-/// In our implementation, Function is extending Object by holding an object field which some extra data
-#[derive(Trace, Finalize, Debug, Clone)]
-pub enum Function {
-    /// A native javascript function
-    NativeFunc(NativeFunction),
-    /// A regular javascript function
-    RegularFunc(RegularFunction),
+/// Sets the ConstructorKind
+#[derive(Debug, Copy, Clone)]
+pub enum ConstructorKind {
+    Base,
+    Derived,
 }
 
-/// Represents a regular javascript function in memory.
+/// Defines how this references are interpreted within the formal parameters and code body of the function.
+///
+/// Arrow functions don't define a `this` and thus are lexical, `function`s do define a this and thus are NonLexical
 #[derive(Trace, Finalize, Debug, Clone)]
-pub struct RegularFunction {
-    /// The fields associated with the function
-    pub object: Object,
-    /// This function's expression
-    pub node: Node,
-    /// The argument declarations of the function
-    pub args: Vec<Node>,
+pub enum ThisMode {
+    Lexical,
+    NonLexical,
 }
 
-impl RegularFunction {
-    /// Make a new regular function
-    #[allow(clippy::cast_possible_wrap)]
-    pub fn new(node: Node, f_args: Vec<FormalParameter>) -> Self {
-        let mut args = vec![];
-        for i in f_args {
-            let node = if let Some(init) = &i.init {
-                init.deref().clone()
-            } else {
-                Node::Local(i.name.clone())
-            };
-            args.push(node);
+/// FunctionBody is specific to this interpreter, it will either be Rust code or JavaScript code (AST Node)
+#[derive(Clone, Finalize)]
+pub enum FunctionBody {
+    BuiltIn(NativeFunctionData),
+    Ordinary(Node),
+}
+
+impl Debug for FunctionBody {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::BuiltIn(_) => write!(f, "native code"),
+            Self::Ordinary(node) => write!(f, "{}", node),
         }
-
-        let mut object = Object::default();
-        object.properties.insert(
-            "arguments".to_string(),
-            Property::default().value(Gc::new(ValueData::Integer(args.len() as i32))),
-        );
-        Self { object, node, args }
     }
 }
 
-/// Represents a native javascript function in memory
-#[derive(Finalize, Clone)]
-pub struct NativeFunction {
-    /// The fields associated with the function
-    pub object: Object,
-    /// The callable function data
-    pub data: NativeFunctionData,
+/// `Trace` implementation for `FunctionBody`.
+///
+/// This is indeed safe, but we need to mark this as an empty trace because neither
+// `NativeFunctionData` nor Node hold any GC'd objects, but Gc doesn't know that. So we need to
+/// signal it manually. `rust-gc` does not have a `Trace` implementation for `fn(_, _, _)`.
+///
+/// <https://github.com/Manishearth/rust-gc/blob/master/gc/src/trace.rs>
+unsafe impl Trace for FunctionBody {
+    unsafe_empty_trace!();
 }
 
-impl NativeFunction {
-    /// Make a new native function with the given function data
-    pub fn new(data: NativeFunctionData) -> Self {
-        let object = Object::default();
-        Self { object, data }
+/// Signal what sort of function this is
+#[derive(Clone, Debug, Copy, Finalize)]
+pub enum FunctionKind {
+    BuiltIn,
+    Ordinary,
+}
+
+/// Waiting on <https://github.com/Manishearth/rust-gc/issues/87> until we can derive Copy
+unsafe impl Trace for FunctionKind {
+    unsafe_empty_trace!();
+}
+
+/// Boa representation of a Function Object.
+///
+/// <https://tc39.es/ecma262/#sec-ecmascript-function-objects>
+#[derive(Trace, Finalize, Clone)]
+pub struct Function {
+    /// Call/Construct Function body
+    pub body: FunctionBody,
+    /// Formal Paramaters
+    pub params: Vec<FormalParameter>,
+    /// This Mode
+    pub this_mode: ThisMode,
+    /// Function kind
+    pub kind: FunctionKind,
+    // Environment, built-in functions don't need Environments
+    pub environment: Option<Environment>,
+}
+
+impl Function {
+    /// This will create an ordinary function object
+    ///
+    /// <https://tc39.es/ecma262/#sec-ordinaryfunctioncreate>
+    pub fn create_ordinary(
+        parameter_list: Vec<FormalParameter>,
+        scope: Environment,
+        body: FunctionBody,
+        this_mode: ThisMode,
+    ) -> Self {
+        Self {
+            body,
+            environment: Some(scope),
+            params: parameter_list,
+            kind: FunctionKind::Ordinary,
+            this_mode,
+        }
+    }
+
+    /// This will create a built-in function object
+    ///
+    /// <https://tc39.es/ecma262/#sec-createbuiltinfunction>
+    pub fn create_builtin(parameter_list: Vec<FormalParameter>, body: FunctionBody) -> Self {
+        Self {
+            body,
+            params: parameter_list,
+            this_mode: ThisMode::NonLexical,
+            kind: FunctionKind::BuiltIn,
+            environment: None,
+        }
+    }
+
+    /// This will handle calls for both ordinary and built-in functions
+    ///
+    /// <https://tc39.es/ecma262/#sec-prepareforordinarycall>
+    /// <https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist>
+    pub fn call(
+        &self,
+        this: &mut Value, // represents a pointer to this function object wrapped in a GC (not a `this` JS object)
+        args_list: &[Value],
+        interpreter: &mut Interpreter,
+        this_obj: &mut Value,
+    ) -> ResultValue {
+        match self.kind {
+            FunctionKind::BuiltIn => match &self.body {
+                FunctionBody::BuiltIn(func) => func(this_obj, args_list, interpreter),
+                FunctionBody::Ordinary(_) => {
+                    panic!("Builtin function should not have Ordinary Function body")
+                }
+            },
+            FunctionKind::Ordinary => {
+                // Create a new Function environment who's parent is set to the scope of the function declaration (self.environment)
+                // <https://tc39.es/ecma262/#sec-prepareforordinarycall>
+                let local_env = new_function_environment(
+                    this.clone(),
+                    this_obj.clone(),
+                    Some(self.environment.as_ref().unwrap().clone()),
+                );
+
+                // Add argument bindings to the function environment
+                for i in 0..self.params.len() {
+                    let param = self.params.get(i).expect("Could not get param");
+                    // Rest Parameters
+                    if param.is_rest_param {
+                        self.add_rest_param(param, i, args_list, interpreter, &local_env);
+                        break;
+                    }
+
+                    let value = args_list.get(i).expect("Could not get value");
+                    self.add_arguments_to_environment(param, value.clone(), &local_env);
+                }
+
+                // Add arguments object
+                let arguments_obj = create_unmapped_arguments_object(args_list);
+                local_env
+                    .borrow_mut()
+                    .create_mutable_binding("arguments".to_string(), false);
+                local_env
+                    .borrow_mut()
+                    .initialize_binding("arguments", arguments_obj);
+
+                interpreter.realm.environment.push(local_env);
+
+                // Call body should be set before reaching here
+                let result = match &self.body {
+                    FunctionBody::Ordinary(ref body) => interpreter.run(body),
+                    _ => panic!("Ordinary function should not have BuiltIn Function body"),
+                };
+
+                // local_env gets dropped here, its no longer needed
+                interpreter.realm.environment.pop();
+                result
+            }
+        }
+    }
+
+    pub fn construct(
+        &self,
+        this: &mut Value, // represents a pointer to this function object wrapped in a GC (not a `this` JS object)
+        args_list: &[Value],
+        interpreter: &mut Interpreter,
+        this_obj: &mut Value,
+    ) -> ResultValue {
+        match self.kind {
+            FunctionKind::BuiltIn => match &self.body {
+                FunctionBody::BuiltIn(func) => func(this_obj, args_list, interpreter),
+                FunctionBody::Ordinary(_) => {
+                    panic!("Builtin function should not have Ordinary Function body")
+                }
+            },
+            FunctionKind::Ordinary => {
+                // Create a new Function environment who's parent is set to the scope of the function declaration (self.environment)
+                // <https://tc39.es/ecma262/#sec-prepareforordinarycall>
+                let local_env = new_function_environment(
+                    this.clone(),
+                    this_obj.clone(),
+                    Some(self.environment.as_ref().unwrap().clone()),
+                );
+
+                // Add argument bindings to the function environment
+                for (i, param) in self.params.iter().enumerate() {
+                    // Rest Parameters
+                    if param.is_rest_param {
+                        self.add_rest_param(param, i, args_list, interpreter, &local_env);
+                        break;
+                    }
+
+                    let value = args_list.get(i).expect("Could not get value");
+                    self.add_arguments_to_environment(param, value.clone(), &local_env);
+                }
+
+                // Add arguments object
+                let arguments_obj = create_unmapped_arguments_object(args_list);
+                local_env
+                    .borrow_mut()
+                    .create_mutable_binding("arguments".to_string(), false);
+                local_env
+                    .borrow_mut()
+                    .initialize_binding("arguments", arguments_obj);
+
+                interpreter.realm.environment.push(local_env);
+
+                // Call body should be set before reaching here
+                let result = match &self.body {
+                    FunctionBody::Ordinary(ref body) => interpreter.run(body),
+                    _ => panic!("Ordinary function should not have BuiltIn Function body"),
+                };
+
+                // local_env gets dropped here, its no longer needed
+                interpreter.realm.environment.pop();
+                result
+            }
+        }
+    }
+
+    // Adds the final rest parameters to the Environment as an array
+    fn add_rest_param(
+        &self,
+        param: &FormalParameter,
+        index: usize,
+        args_list: &[Value],
+        interpreter: &mut Interpreter,
+        local_env: &Environment,
+    ) {
+        // Create array of values
+        let array = array::new_array(interpreter).unwrap();
+        array::add_to_array_object(&array, &args_list[index..]).unwrap();
+
+        // Create binding
+        local_env
+            .borrow_mut()
+            .create_mutable_binding(param.name.clone(), false);
+
+        // Set Binding to value
+        local_env
+            .borrow_mut()
+            .initialize_binding(&param.name, array);
+    }
+
+    // Adds an argument to the environment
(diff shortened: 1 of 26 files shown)


Razican:
I think this would be a nice addition to Boa 0.8, and it could potentially solve #266. What extra work is needed here?

jasonwilliams:
I’ll dig this up again, I remember being blocked on something, leave it with me for now

jasonwilliams:
Most of the properties are there, now its a case of replacing the old function with this one

jasonwilliams:
Almost there, i hit a pretty weird error

```js
function hello() {
  console.log("hello");
}

const a = hello();
a;

```

    Finished dev [optimized + debuginfo] target(s) in 0.10s
     Running `target\debug\boa.exe ./tests/js/test.js`     
hello
undefinedthread '<unnamed>' panicked at 'assertion failed: finalizer_safe()', <::std::macros::panic macros>:2:4
stack backtrace:
   0: std::sys_common::backtrace::_print::{{impl}}::fmt
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\backtrace.rs:59
   1: core::fmt::write
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libcore\fmt\mod.rs:1052
   2: std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\io\mod.rs:1426
   3: std::panicking::default_hook::{{closure}}
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:204
   4: std::panicking::default_hook
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:224
   5: std::panicking::rust_panic_with_hook
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:472
   6: std::panicking::begin_panic<str*>
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libstd\panicking.rs:399 
   7: gc::Gc<boa::builtins::value::ValueData>::inner<boa::builtins::value::ValueData>       
             at <::std::macros::panic macros>:2
   8: gc::{{impl}}::drop<boa::builtins::value::ValueData>
             at C:\Users\jasew\.cargo\registry\src\github.com-1ecc6299db9ec823\gc-0.3.3\src\lib.rs:263
   9: hashbrown::raw::{{impl}}::drop<(alloc::string::String, gc::Gc<boa::builtins::value::ValueData>)>
             at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\hashbrown-0.6.2\src\raw\mod.rs:1041
  10: core::ptr::drop_in_place<alloc::boxed::Box<std::collections::hash::map::HashMap<alloc::string::String, gc::Gc<boa::builtins::value::ValueData>, std::collections::hash::map::RandomState>>>
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libcore\ptr\mod.rs:174
  11: core::ptr::drop_in_place<boa::builtins::function_object::Function>
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libcore\ptr\mod.rs:174
  12: core::ptr::drop_in_place<alloc::boxed::Box<gc::gc::GcBox<Trace>>>
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libcore\ptr\mod.rs:174
  13: gc::gc::{{impl}}::drop
             at C:\Users\jasew\.cargo\registry\src\github.com-1ecc6299db9ec823\gc-0.3.3\src\gc.rs:36
  14: std::thread::local::fast::destroy_value<core::cell::RefCell<gc::gc::GcState>>
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libstd\thread\local.rs:459
  15: std::sys_common::thread_local::register_dtor_fallback::run_dtors
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\thread_local.rs:260
  16: std::sys::windows::thread_local::on_tls_callback
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys\windows\thread_local.rs:198
  17: RtlActivateActivationContextUnsafeFast
  18: RtlActivateActivationContextUnsafeFast
  19: LdrShutdownProcess
  20: RtlExitUserProcess
  21: ExitProcess
  22: o_free
  23: exit
  24: __scrt_common_main_seh
             at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:295
  25: BaseThreadInitThunk
  26: RtlUserThreadStart
thread '<unnamed>' panicked at 'assertion failed: finalizer_safe()', <::std::macros::panic macros>:2:4
stack backtrace:
   0:     0x7ff78c2d46f9 - std::sys_common::backtrace::_print::{{impl}}::fmt
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\backtrace.rs:59
   1:     0x7ff78c2e376b - core::fmt::write
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libcore\fmt\mod.rs:1052
   2:     0x7ff78c2d1ed4 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\io\mod.rs:1426
   3:     0x7ff78c2d719c - std::panicking::default_hook::{{closure}}
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:204
   4:     0x7ff78c2d6dec - std::panicking::default_hook
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:224
   5:     0x7ff78c2d78ef - std::panicking::rust_panic_with_hook
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:472
   6:     0x7ff78c2c6f36 - std::panicking::begin_panic<str*>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libstd\panicking.rs:399
   7:     0x7ff78c34f078 - gc::Gc<gc::GcCell<alloc::boxed::Box<EnvironmentRecordTrait>>>::inner<gc::GcCell<alloc::boxed::Box<EnvironmentRecordTrait>>>
                               at <::std::macros::panic macros>:2
   8:     0x7ff78c350f79 - gc::{{impl}}::drop<boa::builtins::value::ValueData>
                               at C:\Users\jasew\.cargo\registry\src\github.com-1ecc6299db9ec823\gc-0.3.3\src\lib.rs:263
   9:     0x7ff78c34fe1b - core::ptr::drop_in_place<boa::builtins::function_object::Function>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libcore\ptr\mod.rs:174
  10:     0x7ffe416b1030 - <unknown>
  11:     0x7ffe416b35d6 - is_exception_typeof
  12:     0x7ffe416bbb64 - _C_specific_handler
  13:     0x7ffe416b3029 - is_exception_typeof
  14:     0x7ffe416bc021 - _CxxFrameHandler3
  15:     0x7ffe5ff2124f - _chkstk
  16:     0x7ffe5fe9d9b2 - RtlUnwindEx
  17:     0x7ffe5ff0c699 - _C_specific_handler
  18:     0x7ffe5ff211cf - _chkstk
  19:     0x7ffe5feea209 - RtlRaiseException
  20:     0x7ffe5fee9fc3 - RtlRaiseException
  21:     0x7ffe5db3a799 - RaiseException
  22:     0x7ffe416b4890 - CxxThrowException
  23:     0x7ff78c394631 - panic_unwind::imp::panic
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libpanic_unwind\seh.rs:315
  24:     0x7ff78c3945b9 - panic_unwind::__rust_start_panic
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libpanic_unwind\lib.rs:111
  25:     0x7ff78c2d79b8 - std::panicking::rust_panic
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:524
  26:     0x7ff78c2d799d - std::panicking::rust_panic_with_hook
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:494
  27:     0x7ff78c2c6f36 - std::panicking::begin_panic<str*>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libstd\panicking.rs:399
  28:     0x7ff78c30a058 - gc::Gc<boa::builtins::value::ValueData>::inner<boa::builtins::value::ValueData>
                               at <::std::macros::panic macros>:2
  29:     0x7ff78c30aa29 - gc::{{impl}}::drop<boa::builtins::value::ValueData>
                               at C:\Users\jasew\.cargo\registry\src\github.com-1ecc6299db9ec823\gc-0.3.3\src\lib.rs:263
  30:     0x7ff78c30b40d - hashbrown::raw::{{impl}}::drop<(alloc::string::String, gc::Gc<boa::builtins::value::ValueData>)>
                               at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\hashbrown-0.6.2\src\raw\mod.rs:1041
  31:     0x7ff78c3502fe - core::ptr::drop_in_place<alloc::boxed::Box<std::collections::hash::map::HashMap<alloc::string::String, gc::Gc<boa::builtins::value::ValueData>, std::collections::hash::map::RandomState>>>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libcore\ptr\mod.rs:174
  32:     0x7ff78c34fdab - core::ptr::drop_in_place<boa::builtins::function_object::Function>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libcore\ptr\mod.rs:174
  33:     0x7ff78c2c7715 - core::ptr::drop_in_place<alloc::boxed::Box<gc::gc::GcBox<Trace>>>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libcore\ptr\mod.rs:174
  34:     0x7ff78c2c790d - gc::gc::{{impl}}::drop
                               at C:\Users\jasew\.cargo\registry\src\github.com-1ecc6299db9ec823\gc-0.3.3\src\gc.rs:36
  35:     0x7ff78c392bbe - std::thread::local::fast::destroy_value<core::cell::RefCell<gc::gc::GcState>>
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\src\libstd\thread\local.rs:459
  36:     0x7ff78c2d5cae - std::sys_common::thread_local::register_dtor_fallback::run_dtors
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\thread_local.rs:260
  37:     0x7ff78c2d95a8 - std::sys::windows::thread_local::on_tls_callback
                               at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys\windows\thread_local.rs:198
  38:     0x7ffe5fea5021 - RtlActivateActivationContextUnsafeFast
  39:     0x7ffe5fea5103 - RtlActivateActivationContextUnsafeFast
  40:     0x7ffe5feeab30 - LdrShutdownProcess
  41:     0x7ffe5feea92d - RtlExitUserProcess
  42:     0x7ffe5ef6cd8a - ExitProcess
  43:     0x7ffe5de4ae38 - o_free
  44:     0x7ffe5de486ef - exit
  45:     0x7ff78c41ae87 - __scrt_common_main_seh
                               at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:295
  46:     0x7ffe5ef67bd4 - BaseThreadInitThunk
  47:     0x7ffe5feece51 - RtlUserThreadStart
thread panicked while panicking. aborting.
```

**Comment by HalidOdat** [comment]
I will take a look, to see whats causing this. :)

**Comment by HalidOdat** [comment]
Well...
I found were the bug originates but not what causes it.

Its in an unsafe block in `lib.rs` line: 263
I looks to be related to garbage collection when dropping garbage collected values.

I don't know much about the `gc` crate.

Hope that helps. :)

**Comment by HalidOdat** [comment]
```js
function hello() {
  console.log("hello");
}

const a = hello();
a;
```
The above code runs and returns `undefined` and then panics.

```text
								    The panic.
										|
		 ____________________________________________________________________
		 \/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/
undefinedthread '<unnamed>' panicked at 'assertion failed: finalizer_safe()', 
^^^^^^^^^
   |
The returned value.
```

jasonwilliams:
Hey thanks @HalidOdat 
hmmm yeah i noticed that too!

Ill take a look and see what im doing with Gc

jasonwilliams:
Sorted it

Tests remaining
```
    builtins::array::tests::every
    builtins::array::tests::find_index
    builtins::array::tests::for_each
    builtins::array::tests::for_each_push_value
    builtins::array::tests::some
    builtins::symbol::tests::check_symbol_constructor_is_function
    exec::tests::test_short_circuit_evaluation
```

HalidOdat:
Why do we have `Function` and `FunctionObj`?

jasonwilliams:
> Why do we have `function` and `function_object`?

function will be replaced and function_object will be called function, i just need them both right now while i transition in this PR.

once all tests pass and im happy with it ill get rid of all the old function code

jasonwilliams:
Ugh I’m going to have to refactor this, I forgot a single function can be both a call a construct call. So in the function struct I think Instead of body I’ll need call_body and construct_body

jasonwilliams:
I'm leaning towards the idea of Object having a call: Option<Function> and a construct: Option<Function> added on to them, and Function can remove all the overlap with objects (internal_slots, properties).

jasonwilliams:
hmm not sure why im hitting this so much

```js
var str = new String("The Quick Brown Fox Jumps Over The Lazy Dog");
str;
```

boa/src/builtins/function_object.rs
```diff
@@ -0,0 +1,341 @@
...
+impl Function {
+    /// This will create an ordinary function object
+    ///
+    /// <https://tc39.es/ecma262/#sec-ordinaryfunctioncreate>
+    pub fn create_ordinary(
+        parameter_list: Vec<FormalParameter>,
+        scope: Environment,
+        body: FunctionBody,
+        this_mode: ThisMode,
+    ) -> Function {
```
Razican, line 105:
```suggestion
    ) -> Self {
```

boa/src/builtins/function_object.rs
```diff
@@ -0,0 +1,341 @@
...
+        let func = Function {
+            body,
+            environment: Some(scope),
+            params: parameter_list,
+            kind: FunctionKind::Ordinary,
+            this_mode,
+        };
+
+        func
```
Razican, line 114:
```suggestion
        Self {
            body,
            environment: Some(scope),
            params: parameter_list,
            kind: FunctionKind::Ordinary,
            this_mode,
        }
```

boa/src/exec/mod.rs
```diff
@@ -583,88 +602,27 @@ impl Interpreter {
...
-                    self.realm
-                        .environment
-                        .initialize_binding("arguments", arguments_obj);
-
-                    let result = self.run(&data.node);
-                    self.realm.environment.pop();
-                    result
+            ValueData::Object(ref obj) => match obj.borrow_mut().call {
+                Some(ref func) => {
+                    return func.call(&mut f.clone(), &arguments_list, self, this);
```
Razican, line 616:
```suggestion
                    func.call(&mut f.clone(), &arguments_list, self, this)
```

jasonwilliams:
Thanks for taking a look @Razican 
I'm down to 7 tests failing now.

Im hitting another borrowing issue when this code is ran

```js
var sym = Symbol();
sym;
```

Looks like get_prop() returns a borrowed Object that then screws things up

HalidOdat:
I will take a look :+1:

HalidOdat:
After some debugging I found were the bug originates and what is causing it.
The bug is cause when we `borrow_mut()` in `boa/src/exec/mod.rs` line `614`
and latter on we `borrow_mut()` again.
```rust
    pub(crate) fn call(
        &mut self,
        f: &Value,
        this: &mut Value,
        arguments_list: Vec<Value>,
    ) -> ResultValue {
        // All functions should be objects, and eventually will be.
        // During this transition call will support both native functions and function objects
        match (*f).deref() {
            ValueData::Object(ref obj) => match obj.borrow_mut().call { // <<<<<-------- here line 614
                Some(ref func) => {
                    return func.call(&mut f.clone(), &arguments_list, self, this);
                }
                None => panic!("Expected function"),
            },
            _ => Err(Gc::new(ValueData::Undefined)),
        }
    } 
```
Changing the line to:
```rust
ValueData::Object(ref obj) => match obj.clone().borrow().call
```
Fixes the crash and print the output `Symbol()`

HalidOdat:
Also changing the line:
```rust
ValueData::Object(ref o) => (*o.deref().borrow_mut())
```
to:
```rust
ValueData::Object(ref o) => (*o.deref().clone().borrow_mut())
```
Drops the failing tests to 3 and there are no more `GcCell` errors in tests any more.

jasonwilliams:
Thanks a lot @HalidOdat !

HalidOdat:
For now the `cloning` fixes will suffice, but in the long run we need something more permanent.

HalidOdat:
The 3 other tests are failing because `FunctionObject` is not implemented in some areas like:
file `boa/src/exec/mod.rs` line `293` in `Node::ArrowFunctionDecl`.
file `boa/src/exec/mod.rs` line `417` in `Node::New`.
These are the places were `FunctionObject` is not implement, hence the failing tests.

Hope that helps.

jasonwilliams:
Yeah it does help thanks!
There’s light at the end of the tunnel on this now.

jasonwilliams:
all tests passing, now for tidying up

jasonwilliams:
We have TypeOf as both a Node and a variant of unaryOp
Im guessing we should pick which one it is?

jasonwilliams:
Raised https://github.com/jasonwilliams/boa/issues/360 for duplication, i think it'll need more discussion after this PR

Razican:
Wow, great job!

It seems that execution time has gone up by more than 50%. I will try to see if there is any improvement to make in performance.

Razican:
I did some profiling, and I noticed that our biggest pain point is with our heavy usage of `HashMap` and even the GC itself. #356 should help with this. I couldn't really find more pain points with profiling because this is hidding everything else.

I will re-check the code changes just in case.

boa/src/builtins/math/mod.rs
```diff
@@ -478,7 +475,7 @@ pub fn sin(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
 ///
 /// [spec]: https://tc39.es/ecma262/#sec-math.sinh
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sinh
-pub fn sinh(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+pub fn sinh(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
```
Razican, line 478: Just for my understanding, why do we need all these to be mutable?

boa/src/builtins/mod.rs
```diff
@@ -5,15 +5,82 @@
 /// If no length is provided, the length will be set to 0.
 macro_rules! make_builtin_fn {
     ($fn:ident, named $name:expr, with length $l:tt, of $p:ident) => {
-        let $fn = to_value($fn as NativeFunctionData);
-        $fn.set_field_slice("length", to_value($l));
-        $p.set_field_slice($name, $fn);
+        let func = crate::builtins::function::Function::create_builtin(
+            vec![],
+            crate::builtins::function::FunctionBody::BuiltIn($fn),
+    );
```
Razican, line 11:
```suggestion
        );
```

boa/src/builtins/object/internal_methods_trait.rs
```diff
@@ -168,6 +168,111 @@ pub trait ObjectInternalMethods {
         }
     }
 
+    #[allow(clippy::option_unwrap_used)]
```
Razican, line 171: Can we use `expect()` instead?

boa/src/builtins/math/mod.rs
jasonwilliams: The first argument `this` can be altered by Constructor functions. 
Its up to them to change the `this` value however they want.

Usually `this` starts out being a plain ordinary JS object, and String() will add a [[StringData]] internal slot to it. This mutates the value.

Why this wasn't a compile error before i don't know, but something in my refactoring meant that rust now moans if you mutate it.
imo it should have always been an issue but we got away with it before.

jasonwilliams: For regular functions it doesn't need to be mutable, as the first argument is a reference to the function itself (wrapped in a Value).
I think in future we can have 2 separate signatures for Construct and Func, for now they use the same.

jasonwilliams:
> I did some profiling, and I noticed that our biggest pain point is with our heavy usage of `HashMap` and even the GC itself. #356 should help with this. I couldn't really find more pain points with profiling because this is hidding everything else.
> 
> I will re-check the code changes just in case.

Oh nice this is interesting.
How did you profile it? Just `perf`?

I think ill raise an issue around how to deal with GC going forward.

boa/src/builtins/function/mod.rs
```diff
@@ -1,134 +1,318 @@
...
-#[derive(Trace, Finalize, Debug, Clone)]
-pub struct RegularFunction {
-    /// The fields associated with the function
-    pub object: Object,
-    /// This function's expression
-    pub node: Node,
-    /// The argument declarations of the function
-    pub args: Vec<Node>,
+/// FunctionBody is specific to this interpreter, it will either be Rust code or JavaScript code (AST Node)
+#[derive(Clone, Finalize)]
```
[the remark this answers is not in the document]
jasonwilliams: It's defined on line 76.
We can't derive it because `Gc` doesn't natively support fn(_, _, _).
Seeing that FunctionBody doesn't hold other GC values its safe to mark it as an unsafe_empty_trace

boa/src/builtins/string/tests.rs
```diff
@@ -84,19 +84,19 @@ fn repeat() {
...
-        String::from("中文中文")
-    );
+    // assert_eq!(
+    //     forward(&mut engine, "en.repeat(1)"),
+    //     String::from("english")
+    // );
+    // assert_eq!(
+    //     forward(&mut engine, "zh.repeat(2)"),
+    //     String::from("中文中文")
+    // );
```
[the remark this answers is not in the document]
jasonwilliams: oops, thanks for spotting that

jasonwilliams:
@Razican the function object is now in a box, the size_of Value has gone from 280 to 8

boa/src/builtins/function/mod.rs
```diff
@@ -1,134 +1,318 @@
...
 
-/// Represents a native javascript function in memory
-#[derive(Finalize, Clone)]
-pub struct NativeFunction {
-    /// The fields associated with the function
-    pub object: Object,
-    /// The callable function data
-    pub data: NativeFunctionData,
+/// Signal what sort of function this is
+#[derive(Clone, Trace, Debug, Finalize)]
```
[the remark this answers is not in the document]
jasonwilliams: @Razican i get
`the trait `Copy` may not be implemented for this type; the type has a destructor
Copy not allowed on types with destructorsrustc(E0184)` here

Razican:
The error in the benchmarks is because I actually fixed the JavaScript code in the benchmark and now we reach https://github.com/jasonwilliams/boa/blob/master/boa/src/exec/mod.rs#L574 since for loops are not implemented. Maybe we should remove that benchmark until we actually have the for loop execution implemented.

boa/src/builtins/function/mod.rs
```diff
@@ -1,134 +1,310 @@
...
-/// A Javascript `Function` object instance.
-///
-/// A member of the Object type that may be invoked as a subroutine
+/// Sets the ConstructorKind
+#[derive(Debug, Copy, Clone)]
+pub enum ConstructorKind {
+    Base,
+    Derived,
+}
+/// Defines how this references are interpreted within the formal parameters and code body of the function.
```
Razican, line 39:
```suggestion
}

/// Defines how this references are interpreted within the formal parameters and code body of the function.
```

boa/src/builtins/function/mod.rs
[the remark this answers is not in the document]
Razican: I see, it was implemented in a different place and I missed it! maybe it would make sense to move it closer to the definition of the `struct`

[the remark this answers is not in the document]
jasonwilliams: Its just underneath now, after  Debug for FunctionBody

HalidOdat:
If it's possible could you wait. I would like to see what I can optimize. :)

HalidOdat:
- Changed `Interpreter::call` to take a slice instead of `Vec`.
- Removed some unnecessary `cloning` in `Array` and `String` object methods.
- Fixed a bug were we didn't have `length` for ordinary and arrow functions.

Issue #12, opened 2019-03-21
setup test262 harness


jasonwilliams:
It would be good to start running these https://github.com/tc39/test262/tree/master/test/language

#### Project
https://github.com/jasonwilliams/boa/projects/2

#### Tracking of blocking issues:
 - [x] Exception handling (#377)
 - [x] Function expressions (#266)
 - [x] `this` keyword (#264)
 - [x] Hoisting variables (#376)
 - [x] Goal symbols when lexing (#294)

jasonwilliams:
@vitkarpov if this is too much we could start of by doing what rapidus does for now and start with a few js files to test from https://github.com/maekawatoshiki/rapidus/tree/new-gen/test

jasonwilliams:
A good start could be to have a test file, which pulls in some of the pre-defined bindings before loading a file.  Assert would be a start
https://github.com/tc39/test262/blob/master/INTERPRETING.md#test262-defined-bindings

Razican:
I would like to re-open this. I don't think we are so far away from it, I would say we have to implement the following:
 - [ ] Exception handling (#377)
 - [ ] Function expressions (#266)
 - [ ] `this` keyword (#264)
 - [ ] Hoisting variables (#376)

After that, we can implement an automated process to give stats on what we support, and we can start tackling syntax step by step.

Lan2u:
Now that we have goal symbol support (in the form of #486) is it worth giving this a go - do we have any progress already in terms of running test262 on boa?

HalidOdat:
> Now that we have goal symbol support (in the form of #486) is it worth giving this a go - do we have any progress already in terms of running test262 on boa (like a rust test runner)?

No progress yet in terms of setting up the harness of the tests

Lan2u:
> > Now that we have goal symbol support (in the form of #486) is it worth giving this a go - do we have any progress already in terms of running test262 on boa (like a rust test runner)?
> 
> No progress yet in terms of setting up the harness of the tests

Ok cool, I can potentially start looking at this

PR #256, opened 2020-02-24
Fixed comments lexing

Razican:
This fixes a bug in comment lexing, where the initial `/` was not being included in the comment token, and the line number and column numbers were wrong. This would break the numbering of a whole file if there was a multiline comment.

Also, I have added the line and column numbers to the parsing errors, and properly printed them in case of failure, instead of panicking with a debug message.

This should make debugging much easier.


Razican:
Oh, and I forgot to add that this PR should allow to parse multiple files in the CLI, so that we can add one after the other to the executor. This helps in #12. Please, let me know if you want any of this as a different PR.

PR #257, opened 2020-02-24
Fixed empty returns

### LINKED ISSUE #251: Parser error returning from a function.
the is not parsing correctly if there is a empty return in a function.

**POC:**

```js
function addClass(elem){
    return
}
```

```rust
use boa::syntax::lexer::Lexer;
use boa::syntax::parser::Parser;
use std::fs;

fn main() {
    let buffer = fs::read_to_string("test.js").unwrap();
    let mut lexer = Lexer::new(&buffer);
    lexer.lex().expect("failed to lex");
    let mut parser = Parser::new(lexer.tokens);
    println!("{:?}", parser.parse_all());
}
```

**Error:**
```
Err(Expected([], Token { data: Punctuator(CloseBlock), pos: Position { column_number: 1, line_number: 3 } }, "script"))
```

Razican:
This fixes #251, the parsing of empty returns. I added some tests to make sure it doesn't break again.

### DIFF
--- a/boa/src/syntax/parser/mod.rs
+++ b/boa/src/syntax/parser/mod.rs
@@ -222,7 +222,13 @@ impl Parser {
 
                 Ok(Expr::new(ExprDef::ConstDecl(vars)))
             }
-            Keyword::Return => Ok(Expr::new(ExprDef::Return(Some(Box::new(self.parse()?))))),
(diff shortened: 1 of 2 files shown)


Razican:
I rebased the changes with the latest master branch.

PR #258, opened 2020-02-24
Moved test modules to their own files

Razican:
Hi, I've noticed that files are very long, more than 1,000 lines in some cases, and test modules are in the same files as the implementation.

I think it would be nicer to divide modules between tests and the actual implementation, so that files are smaller in size. I have made this small refactor, that I think it should improve the readability of the code / tests, and even improve compile times a bit.

Still, I think it would be nice to modularize a bit the code, so that we don't have such big files.


jasonwilliams:
I'm fine with this, thanks it seems good to me.
Its a good problem to have when you've got so many tests they need separating

PR #261, opened 2020-02-25
Improved parser error messages

Razican:
Parser error messages now also include the line and the column where the offending lexer tokens started. This should make it easier to debug JavaScript code, and also the limitations of the engine itself.

I also took the opportunity to idiomatize some derives. Since derives
are part of the definition of a structure/type, they should go with
the definition, and not with the comments.

### DIFF
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -55,8 +55,8 @@ impl RegularFunction {
     }
 }
 
-#[derive(Finalize, Clone)]
 /// Represents a native javascript function in memory
+#[derive(Finalize, Clone)]
 pub struct NativeFunction {
     /// The fields associated with the function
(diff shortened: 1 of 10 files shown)


Razican:
Hi,

Razican:
Hi, I re-based to master.
The change on comments/derives was to make them more idiomatic. The `#[derive]` attribute is an attribute of the struct/enum whatever, not on the comment, so even if it works the other way, it separates functionality wit some comments, and the comments are also related to the derived trait.

Checking the official documentation, this is the common practice: https://doc.rust-lang.org/rust-by-example/trait/derive.html

jasonwilliams:
Not ignoring this PR, I’m just working on refactoring the parser right now, so some of these are on hold

jasonwilliams:
Can we close this PR now then?

PR #262, opened 2020-02-25
Implement Array.prototype.filter method

Nickforall:
I implemented the `Array.prototype.filter` method according to the spec. Took the implementation of `map` as a reference, let me know if I missed something.

### DIFF
--- a/boa/src/builtins/array/tests.rs
+++ b/boa/src/builtins/array/tests.rs
@@ -675,3 +675,70 @@ fn for_each_push_value() {
     assert_eq!(forward(&mut engine, "a[6]"), "6");
     assert_eq!(forward(&mut engine, "a[7]"), "8");
 }
+
+#[test]
+fn filter() {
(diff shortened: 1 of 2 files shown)


Nickforall:
Rebased ✅

Nickforall:
Rebased again due to conflicts from #268 😛

boa/src/builtins/array/tests.rs
```diff
@@ -675,3 +675,69 @@ fn for_each_push_value() {
     assert_eq!(forward(&mut engine, "a[6]"), "6");
     assert_eq!(forward(&mut engine, "a[7]"), "8");
 }
+
```
jasonwilliams: @Nickforall you don't have a #[test] attribute above the filter() function here like the others

Nickforall: Whoops, must have happened while fixing merge conflicts

PR #263, opened 2020-02-25
[WIP] Added Try/Catch parsing

Razican:
I added parsing functionality for `try`/`catch`/`finally` blocks. Code is still WIP, and it should go to another function at least. Also, currently, if an execution finds a try/catch block, it will just panic, which is not nice. But at least, we are now able to parse the [`assert.js`][assert_file] file of the test262 ECMAScript suite.

The [`sta.js`][sta_file] file is still not parsed due to:
```
ParsingError: Unexpected keyword 'this' at line 14, col 3
```
I opened #264 for that.

I did a small idiomatization in the `get_token()` function, so that it uses idiomatic Rust functions instead of an if block. Functionality should remain the same.

When adding the `Display` implementation of the `try`/`catch`/`finally` blocks, I also improved the display of other blocks, to give a nicer view. Still, it's not indenting expressions, so there is some work to do there.

Finally, I made some modifications to the CLI so that it will only print the result of the last file. Before, it would print intermediate artifacts such as previously defined functions and more. Still not sure how will multiple file execution actually work.

[assert_file]: https://github.com/tc39/test262/blob/master/harness/assert.js
[sta_file]: https://github.com/tc39/test262/blob/master/harness/sta.js


boa/src/syntax/ast/expr.rs
```diff
@@ -117,9 +119,9 @@ impl Display for ExprDef {
         match *self {
             ExprDef::Const(ref c) => write!(f, "{}", c),
             ExprDef::Block(ref block) => {
-                write!(f, "{{")?;
+                writeln!(f, "{{")?;
                 for expr in block.iter() {
-                    write!(f, "{};", expr)?;
+                    writeln!(f, "{};", expr)?;
```
Razican: We should add indenting.

boa/src/syntax/parser.rs
```diff
@@ -347,6 +350,79 @@ impl Parser {
                     Box::new(block),
                 )))
             }
+            Keyword::Try => {
```
Razican: We should move this to a new function, and add the ECMAScript reference: https://tc39.es/ecma262/#sec-try-statement

boa/src/exec.rs
```diff
@@ -507,6 +507,9 @@ impl Executor for Interpreter {
                     ValueData::Function(_) => "function",
                 }))
             }
+            ExprDef::CatchBlock(_, _) | ExprDef::TryCatch(_, _, _) => {
+                todo!("try/catch block execution")
```
Razican: This should not panic, we should ideally check the result of the inner expression and pass the exception to the catch block.

boa/src/exec.rs
```diff
@@ -507,6 +507,39 @@ impl Executor for Interpreter {
...
+                            self.realm.environment.create_mutable_binding(
+                                catch_bind.clone(),
+                                false,
+                                VariableScope::Block,
+                            );
+
+                            self.realm.environment.initialize_binding(catch_bind, val);
+                        }
+
+                        todo!("run the catch block");
```
Razican: Not sure if I'm doing this right, @jasonwilliams. The `catch_expr` here is a `(String, ExprDef::Block)` tuple, and I'm not sure how can I pass the binding value to the catch block. Am I doing this right? Should the type of expression be different?

boa/src/syntax/parser.rs
jasonwilliams: I agree, i want to move things into functions in this file

boa/src/exec.rs
jasonwilliams: Seems like you've done it right to me, the catch_bind is the parameter name right?
That should be initiated on the new block.

You'll also need to pop that environment when you've finished with it but there's a lot of refactoring needed in that anyway

jasonwilliams:
Will have another look through his when I catch some time

boa/src/syntax/parser.rs
Razican: This has been done in the last commit.

boa/src/exec.rs
Razican: About this, the catch block is a `Block` type expression (as it has been implemented right now) but I need to pass a variable to it. Is this the correct way to do it?

Razican:
Leaving the catch block parsing/execution, this PR is ready for review.

I want to note that I did pretty big changes in expression printing. I needed this to properly debug everything, and it seems it's working nicely. It adds proper indenting and pretty formatting.

You can try it out by executing the `assert.js` file of the test suite. It will return the last function.

jasonwilliams:
Doing a parser rewrite at the moment, so this will most likely need to change.
I will keep this here for now and see how i can get it in.

Hopefully the new parser is more easier to reason with and add features to

Razican:
Let's close this PR and try it differently with the new parser :) I think maybe we can take some ideas for the execution.

Issue #74, opened 2019-07-10
Feature Request: Boa Shell


rwaldron:
Similar to SpiderMonkey's `jsshell`, or V8's `d8`, this is needed to create an `eshost` agent.

`./target/debug/bin` is almost that, but needs: 

1. output written to stdout
    1. only outputs the result of the last evaluation.
1. errors written to stderr

Things needed in the shell runtime itself: 

1. `print()` function in the global scope (or similar). SpiderMonkey's JavaScript shell defines this as: 
    > Evaluates the expression(s) and displays the result(s) on `stdout`, separated by spaces (" ") and terminated by a newline ("\n").
1. `globalThis` or similar mechanism
1. Some way to create a new realm and global object, eg. jsshell's `newGlobal()` or d8's `Realm.global(realmId)` (which requires some way of making a `realmId`, ie. `Realm.create()` or `Realm.createAllowCrossRealmAccess()`)
1. Some way to evaluate a string of source code, eg. jsshell's `evaluate(code)` or d8's `Realm.eval(realmId, code)`

jasonwilliams:
Thanks @rwaldron im sure we can get onto this.

jasonwilliams:
@rwaldron what would happen if you pass an object to print(), would it call `toString()` on that object?

jasonwilliams:
> Some way to create a new realm and global object, eg. jsshell's newGlobal() or d8's Realm.global(realmId) (which requires some way of making a realmId, ie. Realm.create() or Realm.createAllowCrossRealmAccess())

@rwaldron do you have an example of Test262 using this somewhere so we can see for context?

Razican:
I think that something that might be missing here is the ability to create multiline blocks in the shell, if we want to complete it further. But for the rest, this is mostly done, right?

HalidOdat:
> I think that something that might be missing here is the ability to create multiline blocks in the shell, if we want to complete it further. But for the rest, this is mostly done, right?

We have pretty much implemented all of the feature shell request.

Maybe we should close this issue in favor of the new issue #341

jasonwilliams:
I’m happy to close this alao

PR #267, opened 2020-03-03
Add print & REPL functionality to CLI

### LINKED ISSUE #142: Create Print() function
Arises from:
https://github.com/jasonwilliams/boa/issues/74

```
1. `print()` function in the global scope (or similar). SpiderMonkey's JavaScript shell defines this as: 
    > Evaluates the expression(s) and displays the result(s) on `stdout`, separated by spaces (" ") and terminated by a newline ("\n").
```

For Boa Shell we need a print function, the implementation is virtually the same as https://github.com/jasonwilliams/boa/blob/master/src/lib/js/console.rs#L11-L98

It might be worth abstracting the above logic into its own function, and have console.log() and print() use it.

print() should only be available in [shell](https://github.com/jasonwilliams/boa/blob/master/src/bin/shell.rs) 

So you will need to add it to the realm there.
You may need to create a helper function in realm which lets you add Objects to the global scope.

#### Spidermonkey equivalent
https://github.com/mozilla/gecko-dev/blob/master/js/src/shell/js.cpp#L2687-L2724

JohnDoneth:
I believe this should close #142 and should make some progress towards #74.

If this PR (gets accepted / is satisfactory) I can continue working on other features needed to complete #74.

### DIFF
--- a/boa_cli/src/main.rs
+++ b/boa_cli/src/main.rs
@@ -11,12 +11,9 @@ use structopt::StructOpt;
 #[derive(Debug, StructOpt)]
 #[structopt(author, about)]
 struct Opt {
-    /// The javascript file to be evaluated.
+    /// The JavaScript file(s) to be evaluated.
     #[structopt(name = "FILE", parse(from_os_str))]
     files: Vec<PathBuf>,
-    /// Open a boa shell.
-    #[structopt(short, long)]
-    shell: bool,
 }
 pub fn main() -> Result<(), std::io::Error> {
     let args = Opt::from_args();
@@ -34,7 +31,7 @@ pub fn main() -> Result<(), std::io::Error> {
         }
     }
 
-    if args.shell || args.files.is_empty() {
+    if args.files.is_empty() {
         loop {
             let mut buffer = String::new();
 
--- a/boa_cli/src/main.rs
+++ b/boa_cli/src/main.rs
@@ -3,6 +3,7 @@
 #![allow(clippy::cognitive_complexity)]
 
 use boa::{exec::Executor, forward_val, realm::Realm};
+use std::io;
 use std::{fs::read_to_string, path::PathBuf};
 use structopt::StructOpt;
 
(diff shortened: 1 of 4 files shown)


jasonwilliams:
Hey this looks great!

boa_cli/src/main.rs
```diff
@@ -2,29 +2,30 @@
...
 use std::{fs::read_to_string, path::PathBuf};
 use structopt::StructOpt;
-
 /// CLI configuration for Boa.
 #[derive(Debug, StructOpt)]
 #[structopt(author, about)]
 struct Opt {
     /// The javascript file to be evaluated.
-    #[structopt(name = "FILE", parse(from_os_str), default_value = "tests/js/test.js")]
+    #[structopt(name = "FILE", parse(from_os_str))]
```
jasonwilliams: So if i understand it by default boa_cli will now be a shell, and we need to pass the file if we want to keep the same functionality as before?
Im personally OK with that as it matches up with Node etc

boa_cli/src/main.rs
```diff
@@ -2,29 +2,30 @@
...
     files: Vec<PathBuf>,
-    /// Open a boa shell (WIP).
+    /// Open a boa shell.
```
jasonwilliams: Does this option make sense now that shell is the default?

boa_cli/src/main.rs
JohnDoneth: Yup! It should be functionally equivalent. I just tested running node locally to check and passing no arguments (just calling `node`) opens an interactive shell.

boa_cli/src/main.rs
JohnDoneth: It is now removed!

jasonwilliams:
Do you mind updating the readme and maybe docs/debugging.md so that new users know to pass a file?

Maybe show the argument boa takes or something like https://github.com/ogham/exa/blob/master/README.md#options or https://github.com/sharkdp/fd/blob/master/README.md#command-line-options

PR #268, opened 2020-03-06
Addition of forEach()

jasonwilliams:
* Taken from @xSke's older branch
* Also some fixes to the debugger
* Tests added

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,15 +11,15 @@
       "windows": {
         "program": "${workspaceFolder}/target/debug/boa_cli.exe"
       },
-      "osx": {
-        "program": "${workspaceFolder}/target/debug/boa_cli"
-      },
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
```diff
@@ -8,7 +8,7 @@
       "type": "lldb",
       "request": "launch",
       "name": "Launch",
-      "program": "${workspaceFolder}/target/debug/boa_cli",
+      "program": "${workspaceFolder}/target/debug/boa_cli.exe",
```
Nickforall: I suppose unix-like systems don't need the `.exe` extension? VSCode has support for platform specific properties in launch.json: https://code.visualstudio.com/Docs/editor/debugging#_platformspecific-properties

jasonwilliams: Yes you’re right well spotted!
I’ll change that, thanks

jasonwilliams: Fixed!

PR #270, opened 2020-03-08
Fix weird behavior with JSON builtin

### LINKED ISSUE #269: Weird behavior with JSON builtin
The JSON built in currently has a few issues that are thankfully easy to fix.Currently the `JSON.stringify` and `JSON.parse` are undefined and expect you to specify the `JSON` builtin's prototype.

The second issue is that the function `JSON.stringify` *(accessed through `JSON.prototype.stringify`)* returns unexpected output.

Example:

```javascript
JSON.prototype.stringify({ number: 1 }) // Currently evaluates to "{ "extensible": true }"
```

This issue seems to be caused by [this line](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L594) using the `internal_slots` instead of the object's properties.

If you want I could submit a PR to fix these issues.

vlakreeh:
Fixes #269, the code to serialize arrays is a bit hacky so the array stays in the correct order.

### DIFF
--- a/boa/src/builtins/json/tests.rs
+++ b/boa/src/builtins/json/tests.rs
@@ -1,0 +1,90 @@
+use super::*;
+use crate::{exec::Executor, forward, forward_val, realm::Realm};
+use serde_json::{json, Value};
+
+#[test]
+fn check_json_global() {
+    let global = ValueData::new_obj(None);
+    let json_global = create_constructor(&global);
+
+    assert!(json_global.is_object());
+
+    let parse_function = json_global.get_prop("parse").unwrap();
+    assert!(parse_function.value.as_ref().unwrap().is_function());
+    let stringify_function = json_global.get_prop("stringify").unwrap();
+    assert!(stringify_function.value.as_ref().unwrap().is_function());
+}
+
+#[test]
+fn test_object_parse() {
+    let realm = Realm::create();
+    let mut engine = Executor::new(realm);
+
+    let parsed = forward_val(
+        &mut engine,
+        r#"JSON.parse("{\"number\": 1, \"bool\": false, \"string\": \"text\"}")"#,
+    )
+    .unwrap();
+    assert!(parsed.is_object());
+
+    let parsed_number = parsed.get_prop("number").unwrap();
+    let parsed_number_val = parsed_number.value.as_ref().unwrap();
+    assert!(parsed_number_val.is_num());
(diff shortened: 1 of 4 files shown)


jasonwilliams:
@vlakreeh could you make a json folder instead? with a mod.rs and a tests.rs similar to the other builtins, then maybe a test could be added to check it actually works as it should be

vlakreeh:
Sure, I'll get that on that soon.

vlakreeh:
While trying to write the test for the stringify function, I found a bug a where all whole numbers were being serialized as floats. I made a quick and hack-y fix for this because because I wasn't quite sure how to go about fixing it and it seemed like it was out of scope of just JSON serializing.

Also I wasn't quite sure if there was a cleaner way I could have written the unit tests, getting the value of a property seemed like a bit of a pain point.

boa/src/builtins/value/mod.rs
```diff
@@ -582,26 +582,66 @@ impl ValueData {
...
+
+                JSONValue::Array(values)
+            }
             ValueData::String(ref str) => JSONValue::String(str.clone()),
-            ValueData::Number(num) => JSONValue::Number(
-                JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
-            ),
+            ValueData::Number(num) => {
+                // TODO: This is because for some reason we are never getting a `ValueData::Integer` value
+                // this is a quick and dirty fix
```
Razican, line 636: Hmmm could we check why is this happening?

boa/src/builtins/json/tests.rs
```diff
@@ -0,0 +1,90 @@
...
+    forward(&mut engine, init);
+
+    let stringified = forward(&mut engine, "JSON.stringify(object)");
+    let parsed_object: Value = serde_json::from_str(&stringified).unwrap();
+
+    assert!(parsed_object.is_object());
+    assert_eq!(parsed_object.get("string").unwrap(), &json!("text"));
+    assert_eq!(parsed_object.get("number").unwrap(), &json!(1));
+    assert_eq!(parsed_object.get("bool").unwrap(), &json!(true));
+    assert_eq!(parsed_object.get("array").unwrap(), &json!([1, 2, 3]));
```
Razican, line 89: This might be easier if we just write JavaScript and get the results of direct property access, but I think it's OK like this too :)

Razican:
This has been superseeded by #325.

PR #278, opened 2020-03-17
Added the ability to dump the token stream or ast in bin.

HalidOdat:
The dump functionality works both for files and REPL.

With --dump-tokens or -t for short it dumps the token stream to stdout  and --dump-ast or -a for short to dump the ast to stdout.

The dumping of tokens and ast is mutually exclusive and when dumping it wont run the code.


Razican:
Seems nice, maybe it would be interesting to print tokes/ast in a given format (JSON for example) in order for other tools to analyze it?

HalidOdat:
Hmmm. That seems like a really nice idea. Thanks :+1:

HalidOdat:
With the last commit 84d41057fb530e3c5d0327c17baf4c7afcf3c9bb, I changed --dump-tokens and --dump-ast to optionally take a format type, by default it is Debug format which internally uses the std::fmt::Debug trait. But you can change it by providing a format type like: Debug, Json, JsonMinified. The input value is case insensitive.

Example: 
```bash
cargo run -- test.js --dump-tokens=Debug # debug output (std::fmt::Debug).
``` 
or
```bash
cargo run -- test.js --dump-tokens=Json # pretty printed json.
```
or
```bash
cargo run -- test.js --dump-tokens=JsonMinified # minified json.
```
Its the same with --dump-ast.

jasonwilliams:
This looks great @HalidOdat i will take a proper look soon but no issues so far

Razican:
It looks great, yes! Maybe the `serde` dependency could be an optional dependency in the lib, maybe behind a feature flag? This would reduce the dependency list for users of the lib that don't need the JSON representation.

And I would say that if we want output in JSON, it should be minified by default, with the option to make it pretty if desired.

HalidOdat:
Thanks for the input. I will see what I can do.

HalidOdat:
With this commit 8a3417684b55c3e4d76bc0415cc192095f11b18e, I made `serde` _serialization_ and _deserialization_ to be a **feature** of boa library it can be activated with the feature flag `serde-ast`.

And with this commit 96c74ae34377f11e187831bf043b592fe629b191, I changed Json format to be minified by default. Removed the JsonMineified format and added JsonPretty, which as the name implies it dumps with pretty print JSON format.

Razican:
Looks perfect to me, great job!

HalidOdat:
Thanks!

jasonwilliams:
Maybe we can get this in before the parser change then rebase the parser branch on top of this?

HalidOdat:
I think so, too.

PR #280, opened 2020-03-20
Implemented the Array.prototype.some method.

HalidOdat:
I implemented the `Array.prototype.some` based on the [spec](https://tc39.es/ecma262/#sec-array.prototype.some). I used `Array.prototype.every` method as a reference point, since they are so similar.

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -778,6 +778,43 @@ pub fn filter(this: &Value, args: &[Value], interpreter: &mut Interpreter) -> Re
     construct_array(&new, &values)
 }
 
+/// Array.prototype.some ( callbackfn [ , thisArg ] )
+///
+/// The some method tests whether at least one element in the array passes
+/// the test implemented by the provided callback function. It returns a Boolean value,
+/// true if the callback function returns a truthy value for at least one element
+/// in the array. Otherwise, false.
+///
+/// Caution: Calling this method on an empty array returns false for any condition!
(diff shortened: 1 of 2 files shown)


boa/src/builtins/array/mod.rs
```diff
@@ -778,6 +778,43 @@ pub fn filter(this: &Value, args: &[Value], interpreter: &mut Interpreter) -> Re
...
+    let max_len: i32 = from_value(this.get_field_slice("length")).unwrap();
+    let mut len = max_len;
+    while i < len {
+        let element = this.get_field_slice(&i.to_string());
+        let arguments = vec![element.clone(), to_value(i), this.clone()];
+        let result = interpreter.call(callback, &this_arg, arguments)?.is_true();
+        if result {
+            return Ok(to_value(true));
+        }
+        len = min(max_len, from_value(this.get_field_slice("length")).unwrap());
```
Razican, line 812: Why do we update the length here?

HalidOdat: Its because the callback can mutate the array and we need to update it.

HalidOdat:
I think this branch is ready to merge. :)

jasonwilliams:
At some point https://tc39.es/ecma262/#sec-iscallable will need to be implemented to check the passed in arg is a function, but i think that can be done separately considering there's an update to how functions will work

HalidOdat:
Added comment describing the updating of the length.

Issue #273, opened 2020-03-09
new Class().method() has incorrect precedence


jcdickinson:
The precedence in `new Class().method()` isn't correct. This expression is parsed as `new (Class().method())`, which is incorrect, the correct parse tree would be equivalent to `(new Class()).method()`.

A real-world example is `new Date().getTime()`.

jasonwilliams:
Tokens for `new Date().getTime()`
```
[boa\src\lib.rs:26] &tokens = [
    Token {
        data: Keyword(
            New,
        ),
        pos: Position {        
            column_number: 1,  
            line_number: 1,    
        },
    },
    Token {
        data: Identifier(      
            "Date",
        ),
        pos: Position {        
            column_number: 5,  
            line_number: 1,    
        },
    },
    Token {
        data: Punctuator(      
            OpenParen,
        ),
        pos: Position {        
            column_number: 9,  
            line_number: 1,    
        },
    },
    Token {
        data: Punctuator(      
            CloseParen,        
        ),
        pos: Position {       
            column_number: 10,
            line_number: 1,   
        },
    },
    Token {
        data: Punctuator(     
            Dot,
        ),
        pos: Position {       
            column_number: 11,
            line_number: 1,   
        },
    },
    Token {
        data: Identifier(     
            "getTime",        
        ),
        pos: Position {       
            column_number: 12,
            line_number: 1,   
        },
    },
    Token {
        data: Punctuator(     
            OpenParen,        
        ),
        pos: Position {       
            column_number: 19,
            line_number: 1,
        },
    },
    Token {
        data: Punctuator(
            CloseParen,
        ),
        pos: Position {
            column_number: 20,
            line_number: 1,
        },
    },
    Token {
        data: Punctuator(
            Semicolon,
        ),
        pos: Position {
            column_number: 21,
            line_number: 1,
        },
    },
]
```

Tokens looks fine

jasonwilliams:
Upon deeper inspection the issue is around here: https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/parser/mod.rs#L688-L721

Normally it just carries on, which is why it continues parsing until the method.
There needs to be some signal to stop after the first call.

jasonwilliams:
if we stopped after here https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/parser/mod.rs#L719 it would break other things such as function chaining

jasonwilliams:
[...]

Issue #240, opened 2020-02-02
Parser Panic


adumbidiot:
When putting the following in test.js:
```var`a = 1;```
 running the command `cargo run -- test.js` will give the following panic:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target\debug\boa.exe test.js`
thread 'main' panicked at '1:4: Unexpected '`'', src/lib\syntax\lexer.rs:620:22
stack backtrace:
   0:     0x7ff636d9a689 - backtrace::backtrace::trace_unsynchronized
                               at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.40\src\backtrace\mod.rs:66
   1:     0x7ff636d9a689 - std::sys_common::backtrace::_print_fmt
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7\/src\libstd\sys_common\backtrace.rs:77
   2:     0x7ff636d9a689 - std::sys_common::backtrace::_print::{{impl}}::fmt
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7\/src\libstd\sys_common\backtrace.rs:59
   3:     0x7ff636db007b - core::fmt::write
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7\/src\libcore\fmt\mod.rs:1052
   4:     0x7ff636d98244 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7\/src\libstd\io\mod.rs:1428
   5:     0x7ff636d9cfbc - std::sys_common::backtrace::_print
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7\/src\libstd\sys_common\backtrace.rs:62
   6:     0x7ff636d9cfbc - std::sys_common::backtrace::print
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7\/src\libstd\sys_common\backtrace.rs:49
   7:     0x7ff636d9cfbc - std::panicking::default_hook::{{closure}}
                               at /rustc/212b2c7da87f3086af535b33a9ca6b5242f2d5a7\/src\libstd\panicking.rs:204
   8:     0x7ff636d9cc0c - std::panicking::default_hook
[...]

PR #281, opened 2020-03-23
Parser fixes #225 #240 #273

### LINKED ISSUE #225: Reimplement the parser
In its current implementation, the parser works recursively, which is fine if you are doing simple expression parsing. But once you start parsing a long series of tokens, a stack overflow may occur (such is the case in #162).

What I propose is to start reimplementing the parser in small logical chunks. This includes:
- eliminating recursive calls and making the parsing iterative
- trying to keep the expressions as linear as possible (e.g. binary operators being represented in [reverse polish notation](https://en.wikipedia.org/wiki/Reverse_Polish_notation), instead of a tree)

---

As I said, the work can be split into small chunks, such as:
- [ ] `var`
- [ ] `let`
- [ ] `const`
- [ ] `return`
- [ ] functions
- [ ] blocks
- [ ] `new`
- [ ] `if`, `else if`, `else` & ternary
- [ ] loops
- [ ] objects
- [ ] arrays
- [ ] `switch`
- [ ] `typeof`
- [ ] binary operators
- [ ] unary operators
- [ ] regular expressions

> Note: this list might not be exhaustive

jasonwilliams:


boa/src/builtins/function/mod.rs
```diff
@@ -37,15 +37,21 @@ pub struct RegularFunction {
     /// The fields associated with the function
     pub object: Object,
     /// This function's expression
-    pub expr: Expr,
+    pub expr: Node,
```
Razican: I would change the name of this to `node`, if it's going to contain a node.

boa/src/syntax/ast/node.rs
```diff
@@ -0,0 +1,300 @@
...
+            Node::If(_, _, _) => 15,
+            // 16 should be yield
+            Node::Assign(_, _) => 17,
+            _ => 19,
+        }
+    }
+}
+
+impl Display for Node {
+    fn fmt(&self, f: &mut Formatter) -> Result {
```
Razican: We should probably add some indentation and proper formatting here, as I did in #263. Maybe I can port those changes to this branch, if it's ok?

boa/src/syntax/ast/node.rs
```diff
@@ -0,0 +1,300 @@
...
+        if !first {
+            f.write_str(", ")?;
+        }
+        first = false;
+        Display::fmt(e, f)?;
+    }
+    Ok(())
+}
+
+// https://tc39.es/ecma262/#prod-FormalParameter
```
Razican: I think this should be added as a doc comment, and not only as a comment. Giving some explanations on the implementation and adding the link for further clarifications on what this is.

boa/src/syntax/ast/op.rs
```diff
@@ -66,8 +67,14 @@ pub enum UnaryOp {
     Not,
     /// `~a` - bitwise-not of the value
     Tilde,
+    /// `type of` - Get the type of object
```
Razican: Isn't the operator `typeof`? 
```suggestion
    /// `typeof` - Get the type of object
```

boa/src/syntax/ast/op.rs
```diff
@@ -40,15 +40,16 @@ impl Display for NumOp {
...
+                NumOp::Exp => "**",
                 NumOp::Mod => "%",
             }
         )
     }
 }
 
 #[derive(Clone, Debug, Trace, Finalize, PartialEq)]
-/// A unary operation on a single value
+/// A unary operation on a single value
```
Razican:
```suggestion
/// A unary operation on a single value
```

Could this be due to a missing `cargo fmt`?

boa/src/syntax/parser/mod.rs
```diff
@@ -43,12 +51,16 @@ impl fmt::Display for ParseError {
                 write!(f, "Expected expression '{}', got '{}'", expected, actual)
             }
             ParseError::UnexpectedKeyword(keyword) => write!(f, "Unexpected keyword: {}", keyword),
+            ParseError::Unexpected(tok, msg) => write!(f, "Unexpected Token {} {}", tok, msg),
             ParseError::AbruptEnd => write!(f, "Abrupt End"),
+            ParseError::General(msg) => write!(f, "{}", msg),
+            ParseError::NormalEOF => write!(f, "EOF"),
+            ParseError::RangeError => write!(f, "RangeError!"),
```
Razican: I think we should also add line/column information for at least `UnexpectedKeyword` and `Unexpected`, and maybe also in `AbruptEnd` and `RangeError`, since they don't give very good debugging information.

boa/src/syntax/parser/mod.rs
```diff
@@ -83,27 +124,144 @@ impl Parser {
...
+            Ok(tok) => {
+                if tok.kind == kind {
+                    Some(self.get_next_token().unwrap())
+                } else {
+                    None
+                }
+            }
+            Err(_) => None,
+        }
+    }
```
Razican: Maybe we can implement the `Iterator` trait in the parser? Many of these methods seem pretty similar to the ones used by that trait, and could maybe make it richer.

boa/src/syntax/parser/mod.rs
```diff
@@ -716,201 +874,1487 @@ impl Parser {
...
+        if curr_tk.kind == tk {
             Ok(())
         } else {
             Err(ParseError::Expected(vec![tk], curr_tk, routine))
         }
     }
 
-    /// Returns an error if the next symbol is not the punctuator `p`
-    #[inline(always)]
+    /// Returns an error if the next symbol is not the punctuator `p`
```
Razican:
```suggestion
    /// Returns an error if the next symbol is not the punctuator `p`
```

boa/src/syntax/parser/mod.rs
```diff
@@ -716,201 +874,1487 @@ impl Parser {
...
+        let params = self.read_formal_parameters()?;
+
+        self.expect_punc(Punctuator::OpenBlock, "expect '{'");
+
+        let body = self.read_block()?;
+
+        Ok(Node::FunctionDecl(name, params, Box::new(body)))
+    }
+
+    /// https://tc39.github.io/ecma262/#prod-ArrayLiteral
```
Razican:
```suggestion
    /// <https://tc39.github.io/ecma262/#prod-ArrayLiteral>
```

boa/src/syntax/parser/mod.rs
```diff
@@ -716,201 +874,1487 @@ impl Parser {
...
+            } else {
+                elements.push(self.read_assignment_expression()?);
+            }
+            self.next_if(TokenKind::Punctuator(Punctuator::Comma));
+        }
+
+        Ok(Node::ArrayDecl(elements))
+    }
+
+    /// https://tc39.github.io/ecma262/#prod-ObjectLiteral
```
Razican:
```suggestion
    /// <https://tc39.github.io/ecma262/#prod-ObjectLiteral>
```

boa/src/syntax/parser/mod.rs
```diff
@@ -716,201 +874,1487 @@ impl Parser {
...
+                    self.get_current_token()?,
+                    "expect ',' or '}'.",
+                ));
+            }
+        }
+
+        Ok(Node::Object(elements))
+    }
+
+    /// https://tc39.github.io/ecma262/#prod-PropertyDefinition
```
Razican:
```suggestion
    /// <https://tc39.github.io/ecma262/#prod-PropertyDefinition>
```

boa/src/wasm.rs
```diff
@@ -25,7 +25,7 @@ pub fn evaluate(src: &str) -> String {
     let tokens = lexer.tokens;
 
     // Setup executor
-    let expr: Expr;
+    let expr: Node;
```
Razican: I would call this `node`.

boa/src/exec/mod.rs
```diff
@@ -68,18 +68,18 @@ impl Executor for Interpreter {
...
 
     #[allow(clippy::match_same_arms)]
-    fn run(&mut self, expr: &Expr) -> ResultValue {
-        match expr.def {
-            ExprDef::Const(Const::Null) => Ok(to_value(None::<()>)),
-            ExprDef::Const(Const::Undefined) => Ok(Gc::new(ValueData::Undefined)),
-            ExprDef::Const(Const::Num(num)) => Ok(to_value(num)),
-            ExprDef::Const(Const::Int(num)) => Ok(to_value(num)),
+    fn run(&mut self, expr: &Node) -> ResultValue {
+        match *expr {
```
HalidOdat: Changed from expr to node.
```suggestion
    fn run(&mut self, node: &Node) -> ResultValue {
        match *node {
```

boa/src/syntax/ast/node.rs
HalidOdat:
```suggestion
/// "Formal parameter" is a fancy way of saying "function parameter".
///
/// In the declaration of a function, the parameters must be identifiers,
/// not any value like numbers, strings, or objects.
///```javascript
///function foo(formalParametar1, formalParametar2) {
///}
///```
/// https://tc39.es/ecma262/#prod-FormalParameter
```

boa/src/builtins/function/mod.rs
```diff
@@ -37,15 +37,21 @@ pub struct RegularFunction {
...
     /// The argument declarations of the function
-    pub args: Vec<Expr>,
+    pub args: Vec<Node>,
 }
 
 impl RegularFunction {
     /// Make a new regular function
     #[allow(clippy::cast_possible_wrap)]
-    pub fn new(expr: Expr, args: Vec<Expr>) -> Self {
+    pub fn new(expr: Node, f_args: Vec<FormalParameter>) -> Self {
```
HalidOdat: maybe change the `expr: Node` parameter to `node: Node`.

boa/src/syntax/ast/node.rs
Razican: This is great!

The link should be between `<` and `>`, and maybe have some explanatory text, before:
`for more information, please check <https://tc39.es/ecma262/#prod-FormalParameter>`

I think this is a great opportunity to greatly improve the documentation of the parser :)

boa/src/syntax/ast/node.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// Return a string representing the type of the given expression
+    TypeOf(Box<Node>),
+    /// Try / Catch
+    Try(
+        Box<Node>,
+        Option<Box<Node>>,
+        Option<Box<Node>>,
+        Option<Box<Node>>,
+    ),
+    This,
```
HalidOdat: Added documentation for `This`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this
```suggestion
    /// A property of an execution context (global, function or eval) that, in non–strict mode, is always a reference to an object and in strict mode can be any value.
    This,
```

boa/src/syntax/ast/token.rs
```diff
@@ -6,24 +6,24 @@ use std::fmt::{Debug, Display, Formatter, Result};
 #[derive(Debug)]
 pub struct Token {
     /// The token Data
```
HalidOdat:
```suggestion
    /// The token kind
```

boa/src/syntax/ast/node.rs
Razican: I would add a small explanation and then the full explanation just below, since Rust documents this in one line, and the extra explanation in a spoiler:

```rust
    /// The `this` keyword.
    ///
    /// A property of an execution context (global, function or eval) that, in non–strict mode, is
    /// always a reference to an object and in strict mode can be any value.
```

boa/src/syntax/ast/node.rs
HalidOdat: Agreed.

boa/src/builtins/function/mod.rs
jasonwilliams: > This looks very nice! Do you want me to add the changes from #263?

Hold off for now, when i get further in to the point a majority of this works then yep, right now its still a bit of a mess

boa/src/syntax/ast/op.rs
jasonwilliams: oh, maybe, sometimes it runs on save and sometimes it doesn't (i think if there's errors)

jasonwilliams: This is intentional.

boa/src/syntax/ast/node.rs
jasonwilliams: Sure!

boa/src/syntax/ast/op.rs
jasonwilliams: If there's a better way of doing this im open to it, ive just found adding spaces is the only way to get it to work for me

HalidOdat: You can use `\` (backslash) at the end with no trailing spaces, to do a newline. Even though it is not a part of the markdown  specification. `cargo doc` markdown parser parses it as a newline, which is what we want.

For Example:
```text
hello\
world
```
_**is:**_
hello\
world

Here is the comment I got the idea from:
https://gist.github.com/shaunlebron/746476e6e7a4d698b373#gistcomment-2271765

HalidOdat: Or alternatively you can just use `<br>`, since in markdown you can use HTML.

jasonwilliams: `\` seems to work with rust-analyzer too, Thanks im happy to use this in future

boa/src/syntax/ast/op.rs
```diff
@@ -66,8 +67,14 @@ pub enum UnaryOp {
...
     /// `~a` - bitwise-not of the value
     Tilde,
+    /// `typeof` - Get the type of object
+    TypeOf,
     /// `...a` - spread an iterable value
     Spread,
+    /// Delete
+    Delete,
+    /// Void
+    Void,
```
HalidOdat: Added documentation to `Delete` and `Void`.
```suggestion
    /// The JavaScript `delete` operator removes a property from an object.
    ///
    /// Unlike what common belief suggests, the delete operator has nothing to do with
    /// directly freeing memory. Memory management is done indirectly via breaking references.
    /// If no more references to the same property are held, it is eventually released automatically. 
    /// 
    /// The `delete` operator returns `true` for all cases except when the property is an
    /// [own](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty)
    /// [non-configurable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cant_delete)
    /// property, in which case, `false` is returned in non-strict mode.
    ///
    /// For more information, please check: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete>
    Delete,
    
    /// The `void` operator evaluates the given `expression` and then returns `undefined`.
    ///
    /// This operator allows evaluating expressions that produce a value into places where an
    /// expression that evaluates to `undefined` is desired.
    /// The `void` operator is often used merely to obtain the `undefined` primitive value, usually using `void(0)`
    /// (which is equivalent to `void 0`). In these cases, the global variable undefined can be used.
    ///
    /// When using an [immediately-invoked function expression](https://developer.mozilla.org/en-US/docs/Glossary/IIFE),
    /// `void` can be used to force the function keyword to be treated as an expression instead of a declaration.
    ///
    /// For more information, please check: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void>
    Void,
```

boa/src/builtins/function/mod.rs
jasonwilliams: Done!

jasonwilliams:
@Razican @HalidOdat  thanks for the commit suggestions, ive also rebased against master.
I think all the methods are there now but its far from working.

For example, im trying (in my test.js file) `console.log("test");` and i get `ParsingError: Unexpected Token . unexpected token.`

So i think now its the point of just making adjustments on whats there now.
Feel free to play with it, i think this will be some work 😂

boa/src/syntax/ast/node.rs
```diff
@@ -0,0 +1,315 @@
...
+}
+
+pub type FormalParameters = Vec<FormalParameter>;
+
+impl FormalParameter {
+    pub fn new(name: String, init: Option<Box<Node>>, is_rest_param: bool) -> FormalParameter {
+        FormalParameter {
+            name: name,
+            init: init,
+            is_rest_param: is_rest_param,
```
Razican: This solves the clippy error.

```suggestion
            name,
            init,
            is_rest_param,
```

boa/src/syntax/ast/op.rs
```diff
@@ -40,15 +40,16 @@ impl Display for NumOp {
...
                 NumOp::Mod => "%",
             }
         )
     }
 }
 
 #[derive(Clone, Debug, Trace, Finalize, PartialEq)]
-/// A unary operation on a single value
+/// A unary operation on a single value   
+/// <https://tc39.es/ecma262/#prod-UnaryExpression>
```
Razican: In this particular case, I would leave the title as-is, and add the link with the explanation just below:
```suggestion
/// A unary operation on a single value
///
/// For more information, please check: <https://tc39.es/ecma262/#prod-UnaryExpression>
```

boa/src/syntax/lexer/mod.rs
```diff
@@ -379,7 +379,7 @@ impl<'a> Lexer<'a> {
 
                     let num = match self.preview_next() {
                         None => {
-                            self.push_token(TokenData::NumericLiteral(0 as f64));
+                            self.push_token(TokenKind::NumericLiteral(0 as f64));
```
Razican:
```suggestion
                            self.push_token(TokenKind::NumericLiteral(0_f64));
```

boa/src/syntax/lexer/mod.rs
```diff
@@ -423,7 +423,7 @@ impl<'a> Lexer<'a> {
                         }
                     };
 
-                    self.push_token(TokenData::NumericLiteral(num));
+                    self.push_token(TokenKind::NumericLiteral(num as f64));
```
Razican: Why is this casting to `f64` needed?

boa/src/syntax/ast/op.rs
Razican: I added a suggestion to improve this. If we use the idiomatic way of Rust to display a summary in one line and the explanation in the second one, we can forget about this limitation :)

boa/src/syntax/lexer/mod.rs
jasonwilliams: hmm good question, now i can't remember, i think an error happened, if it works as num we should revert it

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1015,1330 @@ impl Parser {
...
+                if is_const {
+                    return Err(ParseError::Expected(
+                        vec![TokenKind::Identifier(String::from("Expression"))],
+                        token,
+                        "Expected Expression for Const!",
+                    ));
+                } else {
+                    list.push(Node::LetDecl(vec![(name, None)]));
+                }
+            }
```
Razican: This should fix the clippy error:
```suggestion
            } else if is_const {
                return Err(ParseError::Expected(
                    vec![TokenKind::Identifier(String::from("Expression"))],
                    token,
                    "Expected Expression for Const!",
                ));
            } else {
                list.push(Node::LetDecl(vec![(name, None)]));
            }
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1015,1330 @@ impl Parser {
...
+    /// <https://tc39.es/ecma262/#prod-MemberExpression>
+    fn read_member_expression(&mut self) -> Result<Node, ParseError> {
+        let mut lhs = if self.peek_skip_lineterminator()?.kind == TokenKind::Keyword(Keyword::New) {
+            self.next_skip_lineterminator()?;
+            let lhs = self.read_member_expression()?;
+            self.expect_punc(Punctuator::OpenParen, "expect '('.");
+            let args = self.read_arguments()?;
+            let call_node = Node::Call(Box::new(lhs), args);
+            let new_node = Node::New(Box::new(call_node));
+            new_node
```
Razican: This should fix the clippy error:
```suggestion
            Node::New(Box::new(call_node))
```

boa/src/syntax/parser/mod.rs
```diff
@@ -83,27 +118,133 @@ impl Parser {
         }
     }
 
-    fn parse_function_parameters(&mut self) -> Result<Vec<Expr>, ParseError> {
+    /// consume the next token and increment position
+    fn get_next_token(&mut self) -> Result<Token, ParseError> {
+        self.pos += 1;
+        self.get_token(self.pos)
+    }
```
HalidOdat: Found the parsing bug. The problem is that we are getting the token after incrementing the pos variable. So we are always offset by one in the token stream. Now it parses the `console.log("test")` correctly.
```suggestion
    fn get_next_token(&mut self) -> Result<Token, ParseError> {
        let token = self.get_token(self.pos);
        self.pos += 1;
        token
    }
```

jasonwilliams: haha wow was that it?

HalidOdat: I found a bug in the interpreter:
The code `console.log("hello");` should run. But the interpreter panics.
```text
thread 'main' panicked at 'not yet implemented', boa/src/exec/mod.rs:515:18
```
Even though the AST is correct:
```
[boa/src/lib.rs:58] &expr = StatementList(
    [
        Call(
            GetConstField(
                Local(
                    "console",
                ),
                "log",
            ),
            [
                Const(
                    String(
                        "hello",
                    ),
                ),
            ],
        ),
    ],
)
```

HalidOdat: > haha wow was that it?

That was what caused the code `console.log("test")` to not parse.

HalidOdat:
When running `cargo build` or `cargo check`:
```
warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1230:9
     |
1230 |         self.expect(TokenKind::Punctuator(Punctuator::OpenParen), "expected '('");
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `#[warn(unused_must_use)]` on by default
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1234:9
     |
1234 | /         self.expect(
1235 | |             TokenKind::Punctuator(Punctuator::OpenBracket),
1236 | |             "expected '{'",
1237 | |         );
     | |__________^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1425:9
     |
1425 |         self.expect(TokenKind::Punctuator(Punctuator::OpenParen), "expected '('");
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1429:9
     |
1429 | /         self.expect(
1430 | |             TokenKind::Punctuator(Punctuator::CloseParen),
1431 | |             "expected ')'",
1432 | |         );
     | |__________^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1442:9
     |
1442 |         self.expect_punc(Punctuator::OpenBlock, "Expected open brace {");
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1450:13
     |
1450 |             self.expect_punc(Punctuator::OpenParen, "Expected opening parenthesis (");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1459:13
     |
1459 |             self.expect_punc(Punctuator::CloseParen, "Expected )");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1460:13
     |
1460 |             self.expect_punc(Punctuator::OpenBlock, "Expected {");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1474:13
     |
1474 |             self.expect_punc(Punctuator::OpenBlock, "Expected {");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1522:13
     |
1522 | /             self.expect(
1523 | |                 TokenKind::Punctuator(Punctuator::CloseParen),
1524 | |                 "expected ')'",
1525 | |             );
     | |______________^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1707:21
     |
1707 |                     self.expect_punc(Punctuator::Colon, "expect ':'");
     |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1726:13
     |
1726 |             self.expect_punc(Punctuator::OpenParen, "exect '('");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:1782:13
     |
1782 |             self.expect(TokenKind::Punctuator(Punctuator::Comma), "expect ','");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:2045:13
     |
2045 |             self.expect_punc(Punctuator::OpenParen, "expect '('.");
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:2189:17
     |
2189 |                 self.expect_punc(Punctuator::CloseParen, "Expect punc");
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:2214:9
     |
2214 |         self.expect_punc(Punctuator::OpenParen, "expect '('");
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled

warning: unused `std::result::Result` that must be used
    --> boa/src/syntax/parser/mod.rs:2218:9
     |
2218 |         self.expect_punc(Punctuator::OpenBlock, "expect '{'");
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled
```
why is there not a `?` for error propagation?

jasonwilliams:
> why is there not a `?` for error propagation?

I think there should be, a lot of it was written without running the code against it, so i probably missed loads of areas where i didn't add `?`

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+            name
+        } else {
+            return Err(ParseError::Expected(
+                vec![TokenKind::Identifier(String::from("function name"))],
+                token,
+                "expected function name",
+            ));
+        };
+
+        self.expect(TokenKind::Punctuator(Punctuator::OpenParen), "expected '('");
```
HalidOdat: Added `?`.
```suggestion
        self.expect(TokenKind::Punctuator(Punctuator::OpenParen), "expected '('")?;
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+
+        let params = self.read_formal_parameters()?;
+
+        self.expect(
+            TokenKind::Punctuator(Punctuator::OpenBracket),
+            "expected '{'",
+        );
```
HalidOdat: Added `?`.
```suggestion
        )?;
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+                self.step_back();
+            }
+        }
+
+        Ok(Node::If(Box::new(cond), Box::new(then_), None))
+    }
+
+    /// <https://tc39.es/ecma262/#sec-while-statement>
+    fn read_while_statement(&mut self) -> Result<Node, ParseError> {
+        self.expect(TokenKind::Punctuator(Punctuator::OpenParen), "expected '('");
```
HalidOdat: Added `?`.
```suggestion
        self.expect(TokenKind::Punctuator(Punctuator::OpenParen), "expected '('")?;
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+
+        let cond = self.read_expression()?;
+
+        self.expect(
+            TokenKind::Punctuator(Punctuator::CloseParen),
+            "expected ')'",
+        );
```
HalidOdat: Added `?`.
```suggestion
        )?;
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+
+        let body = self.read_statement()?;
+
+        Ok(Node::WhileLoop(Box::new(cond), Box::new(body)))
+    }
+
+    /// <https://tc39.es/ecma262/#sec-try-statement>
+    fn read_try_statement(&mut self) -> Result<Node, ParseError> {
+        // TRY
+        self.expect_punc(Punctuator::OpenBlock, "Expected open brace {");
```
HalidOdat: Added `?`.
```suggestion
        self.expect_punc(Punctuator::OpenBlock, "Expected open brace {")?;
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+        let try_clause = self.read_block_statement()?;
+        let is_catch = self
+            .next_if_skip_lineterminator(TokenKind::Keyword(Keyword::Catch))
+            .unwrap_or(false);
+
+        // CATCH
+        let (catch, param) = if is_catch {
+            self.expect_punc(Punctuator::OpenParen, "Expected opening parenthesis (");
```
HalidOdat: Added `?`.
```suggestion
            self.expect_punc(Punctuator::OpenParen, "Expected opening parenthesis (")?;
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+            // TODO: should accept BindingPattern
+            let tok = self.get_next_token()?;
+            let catch_param = match tok.kind {
+                TokenKind::Identifier(s) => Node::Local(s),
+                _ => {
+                    return Err(ParseError::Unexpected(tok, "expected identifier."));
+                }
+            };
+            self.expect_punc(Punctuator::CloseParen, "Expected )");
```
HalidOdat: Added `?`.
```suggestion
            self.expect_punc(Punctuator::CloseParen, "Expected )")?;
```

boa/src/syntax/parser/mod.rs
```diff
@@ -876,41 +1016,1329 @@ impl Parser {
...
+            self.expect_punc(Punctuator::OpenBlock, "Expected {");
```
HalidOdat: Added `?`.
```suggestion
            self.expect_punc(Punctuator::OpenBlock, "Expected {")?;
```

boa/src/exec/mod.rs
```diff
@@ -510,6 +512,7 @@ impl Executor for Interpreter {
                     ValueData::Function(_) => "function",
                 }))
             }
```
HalidOdat: The reason the interpreter paniced is because the StatementList was not implemented.
Now the code `console.log("hello")` should run.
```suggestion
            },
            Node::StatementList(ref list) => {
                for (i, item) in list.iter().enumerate() {
                    if i + 1 == list.len() {
                        return self.run(item);
                    }
                    self.run(item)?;
                }

                Ok(Gc::new(ValueData::Undefined))
            },
```

jasonwilliams:
its so great to see the new parser work.
ran some tests, so far it seems the array ones fail.
```
test builtins::array::tests::concat ... ok
test builtins::array::tests::find ... FAILED
test builtins::array::tests::filter ... FAILED
test builtins::array::tests::index_of ... FAILED
test builtins::array::tests::fill ... FAILED
test builtins::array::tests::last_index_of ... FAILED
test builtins::boolean::tests::check_boolean_constructor_is_function ... ok
test builtins::array::tests::find_index ... FAILED
test builtins::array::tests::every ... FAILED
test builtins::array::tests::pop ... FAILED
test builtins::array::tests::reverse ... FAILED
test builtins::array::tests::inclues_value ... FAILED
```

looks like a starting point if anyone wanted to look at that

HalidOdat:
I will do some refactoring, documentation, bug fixing and in general some improvement in code quality.

Hopefully I will a create pull request today or tomorrow!

Razican:
To add to this, in #283 I improved the parser error messages to see what error this was giving. I also made sure all errors in the tests display any potential result.

I think that the `forward()` function should return a `Result` to be able to better handle errors.

jasonwilliams:
This branch has been rebased against master

boa/src/syntax/ast/op.rs
jasonwilliams: > I added a suggestion to improve this

I might have missed this, but what was the suggestion?

jasonwilliams:
peek(0) now checks the current token instead of jumping ahead by one.
Still a lot of tests failing, but arrays don't go into infinite loops anymore

Razican:
After reading @HalidOdat's comment in #283, I propose to have an inner `ParserCursor` structure, that has very basic operations:

 - `next()`: will increment position in 1, and then return it, if any (so `Option<Token>` should be the return type.
 - `peek(skip: usize)`: will return the token in `position + 1 + skip` (also `Option<token>`
 - `peek_skip(skip: Fn)`: will return the next token skipping any token that returns `true` when applying the function.
 - `current()`: will return the current token. An API decision can be made to see if this will always return a token, or if it will be `Option<Token>` too.
 - `prev()`: will decrement position in 1, and then return the current token, if it exists (so `None` if we already were at the beginning).
 - `peek_prev(skip: usize)`: the same as `prev()` but without moving the cursor, and being able to skip some.
 - `peek_prev_skip(skip: Fn)`: The same as `peek_prev()` but using a function to skip tokens instead of a count.

Then, once we implement this, we can see if we use all of the methods or not. The parser would only be able to use these methods (we can enforce it by moving this structure to a submodule, with private fields. This would also clean up the huge `parser` module.

Some other functions to consider would be `seek()` that would move the cursor to a new position.

These functions are heavily inspired by `Iterator`, `Peekable` and `Cursor` in the standard library.

HalidOdat:
FINALLY NO INFINITE LOOP:
```
test result: FAILED. 83 passed; 43 failed; 1 ignored; 0 measured; 0 filtered out
```
BTW: Some tests are wrong. Some parser tests use `Block` instead of `StatementList` so there is probably more passing tests.

jasonwilliams:
> next(): will increment position in 1, and then return it, if any (so Option<Token> should be the return type.

This is how it was originally, I believe the bug we had was that the first time next() is called, it jumps over the first token, so we changed it so next() selects the current token, *then* advances by 1.

HalidOdat:
I think these functions should return an `Option<&Token>`. So we do not have to `clone` every single time we get a token. And only `clone` on demand.

jasonwilliams:
@HalidOdat issue created here:
https://github.com/jasonwilliams/boa/issues/284

So its not forgotten about

jasonwilliams:
Im going to look at the `check_declarations` tests

jasonwilliams:
```
failures:
    builtins::array::tests::every
    builtins::array::tests::fill
    builtins::array::tests::filter
    builtins::array::tests::find
    builtins::array::tests::find_index
    builtins::array::tests::for_each
    builtins::array::tests::for_each_push_value
    builtins::array::tests::is_array
    builtins::array::tests::map
    builtins::boolean::tests::construct_and_call
    builtins::boolean::tests::constructor_gives_true_instance
    builtins::boolean::tests::instances_have_correct_proto_set
    builtins::function::tests::check_arguments_object
    builtins::number::tests::call_number
    builtins::number::tests::value_of
    builtins::regexp::tests::test_constructors
    builtins::regexp::tests::test_exec
    builtins::regexp::tests::test_last_index
    builtins::regexp::tests::test_to_string
    builtins::string::tests::concat
    builtins::string::tests::concat
    builtins::string::tests::construct_and_call
    builtins::string::tests::ends_with
    builtins::string::tests::match_all
    builtins::string::tests::repeat
    builtins::string::tests::replace_with_function
    builtins::string::tests::starts_with
    builtins::string::tests::test_match
    exec::tests::array_rest_with_arguments
    exec::tests::object_field_set
    exec::tests::spread_with_arguments
    exec::tests::test_early_return
    exec::tests::test_short_circuit_evaluation
    exec::tests::test_tilde_operator
    syntax::lexer::tests::check_line_numbers
    syntax::lexer::tests::check_single_line_comment
    syntax::parser::tests::check_function_declarations
    syntax::parser::tests::check_object_short_function
    syntax::parser::tests::check_object_short_function_arguments
    syntax::parser::tests::check_operations

test result: FAILED. 86 passed; 40 failed; 1 ignored; 0 measured; 0 filtered out
```

jasonwilliams:
@HalidOdat @Razican what do you think about this function?
https://github.com/jasonwilliams/boa/pull/281/commits/e18f0c223a57ca259044cbcc189fb952aff3e7aa#diff-3438145cf749502a52e328ca05335fe1R1229-R1273

Should it stay as it is? or split read const/let separately

Razican:
> Should it stay as it is? or split reading const/let declarations separately.
> In the spec they look as one.

I think it's ok for now. The function is small enough so that you see what it's doing, at least for now.

jasonwilliams:
@Razican I messaged you on discord

HalidOdat:
I have good news. I managed to fix the parser errors. But there are other tests that fail. I will create a pull request as soon as possible.
```
test result: FAILED. 92 passed; 35 failed; 1 ignored; 0 measured; 0 filtered out
```

HalidOdat:
Why are we treating `comments` as tokens. Shouldn't the `Lexer` ignore them? :confused:

jasonwilliams:
Yes I don’t see any reason why comments need to make their way towards the parser, so feel free to change that.

Razican:
> Yes I don’t see any reason why comments need to make their way towards the parser, so feel free to change that.

About this, why do we need line breaks in the parser? I currently can't find a reason for it.

HalidOdat:
The line breaks/terminators are needed because the `javascript` specification requires it.
Its needed because of [Automatic Semicolon Insertion](http://www.bradoncode.com/blog/2015/08/26/javascript-semi-colon-insertion/).

jasonwilliams:
> The line breaks/terminators are needed because the `javascript` specification requires it.
> Its needed because of [Automatic Semicolon Insertion](http://www.bradoncode.com/blog/2015/08/26/javascript-semi-colon-insertion/).

Yeah pretty much this.
Most functions in the new parser have a spec link above, if you read some of them, they state to check for a line break and others don’t care. So we need to know. I don’t have an example right now but if you check some of them they’re there.

HalidOdat:
Why do we have Node::Construct and Node::New? Shouldn't it just be Node::New? :confused:

jasonwilliams:
Ouch did you see those benchmarks?

Yes it should be just Node::New 
We don’t need both.

HalidOdat:
> Ouch did you see those benchmarks?
> 
> Yes it should be just Node::New or Node::Construct
> We don’t need both.
> 
> Parsing the expression is now twice as fast (yay).
> 
> The hello world parsing is a lot slower (more than twice as slow as before)

I saw them. I think it is because we have to `clone` the token every single time we get one and for a string like `"Hello world"` it is expensive. This is were string interning  and using `&Token` would be performant.

Razican:
> I saw them. I think it is because we have to `clone` the token every single time we get one and for a string like `"Hello world"` it is expensive. This is were string interning and using `&Token` would be performant.

I'm working on this. I have the design, but I'm fixing all the compile errors here and there. I will have something to present tomorrow :)

boa/src/syntax/parser/mod.rs
```diff
@@ -83,834 +159,1428 @@ impl Parser {
...
+        self.read_expression()
+    }
+
+    /// <https://tc39.es/ecma262/#sec-break-statement>
+    fn read_break_statement(&mut self) -> ParseResult {
+        let tok = self.get_next_token()?;
+        match tok.kind {
+            TokenKind::LineTerminator
+            | TokenKind::Punctuator(Punctuator::Semicolon)
+            | TokenKind::Punctuator(Punctuator::CloseParen) => {
```
Razican: Shouldn't this be `Punctuator::CloseBlock`?

```suggestion
            | TokenKind::Punctuator(Punctuator::CloseBlock) => {
```

HalidOdat: I think so, too.

HalidOdat: The only context were `break` is allowed, is in `{}` braces.

boa/src/syntax/parser/mod.rs
```diff
@@ -83,834 +159,1428 @@ impl Parser {
...
+                "expected ')'",
+            )?;
+            Some(Box::new(step))
+        };
+
+        let body = Box::new(self.read_statement()?);
+
+        let for_node = Node::ForLoop(init, cond, step, body);
+
+        Ok(Node::Block(vec![for_node]))
```
Razican: Why does the for statement return a block with a `for` node?

HalidOdat:
I'm working on better `Parser` documentation. :memo:

jasonwilliams:
> > Ouch did you see those benchmarks?
> > Yes it should be just Node::New or Node::Construct
> > We don’t need both.
> > Parsing the expression is now twice as fast (yay).
> > The hello world parsing is a lot slower (more than twice as slow as before)
> 
> I saw them. I think it is because we have to `clone` the token every single time we get one and for a string like `"Hello world"` it is expensive. This is were string interning and using `&Token` would be performant.

Although true we didn't have string interning before though, so (lack of) that wouldn't have caused the drop. For `&Token` were we not cloning token before also?

HalidOdat:
It could also just be, that the parser is more advanced now. The previous parser was more primitive. And we did not account for things like line terminators.

I will look into it, when I get some time. :+1:

jasonwilliams:
> I found a way. If you take a &[Token] instead of an owning Vec<Token> in Cursor and you annotate the returning references with 'a, so &'a Token in functions that return it like next, next_if, etc. (also in Parser) you can eliminate the clone.
BTW I'm still new to rust lifetimes so there are probably better ways of doing this.

Do we want an issue for this or is this something @HalidOdat is adding to this pr?

HalidOdat:
> > I found a way. If you take a &[Token] instead of an owning Vec in Cursor and you annotate the returning references with 'a, so &'a Token in functions that return it like next, next_if, etc. (also in Parser) you can eliminate the clone.
> > BTW I'm still new to rust lifetimes so there are probably better ways of doing this.
> 
> Do we want an issue for this or is this something @HalidOdat is adding to this pr?

I'm on it. :+1:

HalidOdat:
Yes! Performance has increased. :tada:

Razican:
Something I would like to do with the new parser would be to avoid having more than 1,700 lines in one file, and divide it in small parsers.

So, we could have empty structures that could implement a `Parse` trait, that would only receive the cursor and have a `parse()` method. We could divide each section in the spec into its own module, with better documentation.

What do you think?

boa/src/syntax/parser/mod.rs
```diff
@@ -1,916 +1,1733 @@
...
+        //                     MethodDefinitionKind::Set
+        //                 },
+        //                 func_name,
+        //                 f,
+        //             ));
+        //         }
+        //     }
+
+        //     return Ok(PropertyDefinition::IdentifierReference(name));
+        // }
```
Razican: What should we do with this? I'm guessing it's related to `CoverInitializedName` parsing, right?

boa/src/syntax/parser/mod.rs
```diff
@@ -1,916 +1,1733 @@
...
+    }
+
+    /// Reads an initializer of variables.
+    ///
+    /// Note: it will expect the `=` token to have been already read.
+    ///
+    /// More information: <https://tc39.es/ecma262/#prod-Initializer>
+    fn read_initializer(&mut self) -> ParseResult {
+        self.read_assignment_expression()
+    }
```
Razican: I added some information in the docs here, but it feels a bit contra-intuitive. The `Initializer` is always a `=` character followed by an `AssignmentExpression`. So I think that if we want to divide the logic, we shouldn't ask for the calling code to already parse part of the initializer (the `=` token, in this case).

I think we should change it to expect reading the `=` token, and then the assignment expression.

More info: https://tc39.es/ecma262/#prod-Initializer

boa/src/syntax/ast/op.rs
Razican: > > I added a suggestion to improve this
> 
> I might have missed this, but what was the suggestion?

I think the diff is outdated here, feel free to resolve the conversation. I think the new pattern I used when documenting other functions of the parser was to add the small summary in the first line, then an empty line, and the rest in the next lines.

This is similar to how the Rust standard library does it: https://doc.rust-lang.org/stable/src/std/path.rs.html#2446-2470

boa/src/syntax/parser/mod.rs
jasonwilliams: Its not `CoverInitializedName` you're close, its `MethodDefinition` https://tc39.es/ecma262/#prod-MethodDefinition

We could park it (along with this code) an an issue for now? 
Or we tackle it as part of this PR

jasonwilliams:
> Something I would like to do with the new parser would be to avoid having more than 1,700 lines in one file, and divide it in small parsers.
> 
> So, we could have empty structures that could implement a `Parse` trait, that would only receive the cursor and have a `parse()` method. We could divide each section in the spec into its own module, with better documentation.
> 
> What do you think?

Sounds interesting, what do you mean by 'empty structures' though?
How do we know how best to break the parser up?
Should this be done now or as a follow up?

boa/src/syntax/parser/mod.rs
Razican: We can park it, but I would remove the code, and add an explanation on what should be achieved here. Something like:
```TODO
	// TODO: `MethodDefinition` parsing
	// <https://tc39.es/ecma262/#prod-MethodDefinition>
```

jasonwilliams: https://github.com/jasonwilliams/boa/issues/292 there you go

Razican:
> Sounds interesting, what do you mean by 'empty structures' though?

We can do something like this:
```rust
/// This trait contains the `parse()` method.
///
/// TODO: some extra docs here.
/// This enables using `ParserName::parse()` to parse parts of the spec.
trait Parse {
	fn parse(cursor: &Cursor<'_>) -> ParseResult;
}

/// Statement list item.
/// 
/// TODO: some extra docs here.
/// More information: <https://tc39.es/ecma262/#prod-StatementListItem>
#[derive(Debug, Clone, Copy)]
struct StatementListItem;

impl Parse for StatementListItem {
	fn parse(cursor: &Cursor<'_>) -> ParseResult {
		unimplemented!("State List Item parsing");
	}
}
```

And then call it like this:
```rust
let item = StatementListItem::parse()?;
```

> How do we know how best to break the parser up?

This is a tricky question. I think we should have a structure per each item in the spec, but many of them could share modules. Maybe it makes sense having a sub-module for [expressions][spec-expressions], another one for [statements and declarations][spec-stm-decl], another one for [functions and classes][spec-func-class] and another one for [scripts and modules][spec-script-module]. Maybe others for error handling and language extensions too, just keeping the sections in the spec.

Then, we could sub-divide these modules in the sub-sections in the specification itself. I think it would make it easier to navigate once you have the spec. This, along with some careful documentation would help also newcomers to the project.

> Should this be done now or as a follow up?

Good question. I'm happy either way. This might not take so much time, though, since the parser is already written. On the other hand, we might need to do some extra work, though, to separate everything in smaller pieces.

There is also the fact of learning what to parse if an spec can be of multiple types. Should we just try parsing, check the error and try another parser? I'm talking about cases like [this one][spec-cover-parent]. Should we try to parse an expression? or a binding identifier? should the parent parser peek next tokens to see what options we have?

[spec-expressions]: https://tc39.es/ecma262/#sec-ecmascript-language-expressions
[spec-stm-decl]: https://tc39.es/ecma262/#sec-ecmascript-language-statements-and-declarations
[spec-func-class]: https://tc39.es/ecma262/#sec-ecmascript-language-functions-and-classes
[spec-script-module]: https://tc39.es/ecma262/#sec-ecmascript-language-scripts-and-modules
[spec-cover-parent]: https://tc39.es/ecma262/#prod-CoverParenthesizedExpressionAndArrowParameterList

boa/src/syntax/parser/mod.rs
Razican: What do you guys think about this? The `AbruptEnd` variant will only be triggered when the parser has no extra tokens to retrieve (EOF, until we mess things up with streams xD), so no need to add line information here. `General` is only used once, and I added position information, and I already added position information to `UnexpectedKeyword`. Should we add it to `RangeError`?

I think we should improve the error messages of `RangeError` and `AbruptEnd`, though, and we might want to revisit the need of `General`, that it's only used once.

jasonwilliams:
I would open up an issue for discussion as i think we've already hit quite a big milestone here. Might be an idea to ship it in its current form and then start talking about how to break it up.

I also think @IovoslavIovchev may want to weigh in on this considering he talked about breaking up the parser too here: https://github.com/jasonwilliams/boa/issues/225

Razican:
> I would open up an issue for discussion as i think we've already hit quite a big milestone here. Might be an idea to ship it in its current form and then start talking about how to break it up.

Sure, let's do this. I would wait for #291 to land, though, since includes an extra improvement.

> I also think @IovoslavIovchev may want to weigh in on this considering he talked about breaking up the parser too here: #225 

I'm not sure if this is the same thing. The parser would still be recursive in this case. I actually wouldn't know how to make a parser iterative. In an easy way at least.

HalidOdat:
:tada: The performance improves yet again :tada:

Razican:
This looks really impressive! Great work everyone, I think it's ready for merging. I'm a bit curious about the expression parsing benchmark that shows the new bench being `-106%` the previous one. The real number is 34%, which is very promising!

Even for the Hello World parsing,the reduction is significant, and the rest of the benchmarks don't change (except maybe for the lexer benchmark a bit in the Hello World).

And the internal code looks much cleaner. I will start working in the parser segmentation for Boa 0.8 :)

jasonwilliams:
> This looks really impressive! Great work everyone, I think it's ready for merging. I'm a bit curious about the expression parsing benchmark that shows the new bench being -106% the previous one. The real number is 34%, which is very promising!

Yeah im not sure why its doing that either, very odd!

> the rest of the benchmarks don't change (except maybe for the lexer benchmark a bit in the Hello World).
The other benchmarks don't include lexing or parsing (they only start from execution) so i wouldn't expect them to change.

HalidOdat:
> How did you come to this number?

You calculate the percentage manually by using the two completion times from benchmark for `master` and `parser` branches.

`(100% * parser_time) / master_time` - to calculate the percentage.

Hope that helps :+1:

jasonwilliams:
The reason i asked was because i thought that was more just getting the percentage of parser_time from master_time rather than getting the percentage drop from master down to parser, which i think would be bigger than 32% right? (the difference between the two)

HalidOdat:
> Would a starting point of 11.4us down to 3.7us be a 67% decrease?
Yes it is a 67% decrease.

jasonwilliams:
Alright, lets do this

PR #282, opened 2020-03-24
Added indentation and better Display implementation for AST nodes

Razican:
This PR is to the `parser` branch, in order to keep developing the new parser in #281 .

I added indentation for the `fmt::Display` implementation in AST nodes, and also improved greatly how the final code looks. It should look pretty close to how a programmer would write it.

I also took the opportunity to fix a bunch of warnings. Some of them were showing in the builds, some of them were silenced. I added some extra lints in order to find them.

There are only a couple of warnings left: the usage of the `dbg!()` macro, that in my personal opinion should be a warning (we should remove the usage of the macro before merging into production) and a couple of unused methods, that should be removed in the case we end up not needing them.

Let me know if you want any change.

### DIFF
--- a/boa/src/builtins/value/mod.rs
+++ b/boa/src/builtins/value/mod.rs
@@ -850,10 +850,9 @@ impl Display for ValueData {
                 Function::NativeFunc(_) => write!(f, "function() {{ [native code] }}"),
                 Function::RegularFunc(ref rf) => {
                     write!(f, "function{}(", if rf.args.is_empty() { "" } else { " " })?;
-                    let last_index = rf.args.len() - 1;
                     for (index, arg) in rf.args.iter().enumerate() {
                         write!(f, "{}", arg)?;
-                        if index != last_index {
+                        if index + 1 != rf.args.len() {
                             write!(f, ", ")?;
                         }
                     }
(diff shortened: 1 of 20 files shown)


HalidOdat:
Very nice! The code looks much cleaner.

boa/src/builtins/value/mod.rs
```diff
@@ -855,15 +849,15 @@ impl Display for ValueData {
...
                 Function::RegularFunc(ref rf) => {
-                    write!(f, "function(")?;
+                    write!(f, "function{}(", if rf.args.is_empty() { "" } else { " " })?;
                     let last_index = rf.args.len() - 1;
                     for (index, arg) in rf.args.iter().enumerate() {
                         write!(f, "{}", arg)?;
                         if index != last_index {
                             write!(f, ", ")?;
                         }
                     }
```
HalidOdat: There is a bug here. When `rf.args.len() == 0`, then if you do `0usize - 1`, you get:
```
thread 'main' panicked at 'attempt to subtract with overflow', boa/src/builtins/value/mod.rs:859:38
```
```suggestion
                    write!(f, "function{}(", if rf.args.is_empty() { "" } else { " " })?;
                    for (index, arg) in rf.args.iter().enumerate() {
                        write!(f, "{}", arg)?;
                        if index + 1 != rf.args.len() {
                            write!(f, ", ")?;
                        }
                    }
```

jasonwilliams:
Yeah this looks great to me, happy for me to merge or you wanted to do more changes?

Razican:
> Happy for me to merge or you wanted to do more changes?
You can merge :) I'll probably do more PRs during the week to add more things, but for now I think it's good to go!

PR #283, opened 2020-03-25
New parser fixes tests failing against #281

Razican:
I have ported #261 to the new parser (#281) (this should close #261). And I'm working on fixing the tests that don't pass. I'm getting this error:
```
ParsingError: Unexpected Token 'empty' at line 1, col 13
```

When parsing:
```javascript
        var empty = [ ];
        var one = ["a"];
        var many = ["a", "b", "c"];
        var duplicates = ["a", "b", "c", "a", "b"];
```

So, I'll try to fix that variable creation tomorrow if possible. Feel free to add suggestions.

### DIFF
--- a/boa/src/syntax/parser/mod.rs
+++ b/boa/src/syntax/parser/mod.rs
@@ -6,6 +6,7 @@ use crate::syntax::ast::{
     keyword::Keyword,
     node::{FormalParameter, FormalParameters, Node, PropertyDefinition},
     op::{AssignOp, BinOp, BitOp, CompOp, LogOp, NumOp, Operator, UnaryOp},
+    pos::Position,
     punc::Punctuator,
     token::{Token, TokenKind},
 };
@@ -17,11 +18,11 @@ pub enum ParseError {
     /// When it expected a certain kind of token, but got another as part of something
     Expected(Vec<TokenKind>, Token, &'static str),
     /// When it expected a certain expression, but got another
-    ExpectedExpr(&'static str, Node),
+    ExpectedExpr(&'static str, Node, Position),
     /// When it didn't expect this keyword
-    UnexpectedKeyword(Keyword),
+    UnexpectedKeyword(Keyword, Position),
     /// When a token is unexpected
-    Unexpected(Token, &'static str),
+    Unexpected(Token, Option<&'static str>),
     /// When there is an abrupt end to the parsing
     AbruptEnd,
     /// Out of range error, attempting to set a position where there is no token
@@ -47,11 +48,28 @@ impl fmt::Display for ParseError {
                 actual.pos.line_number,
                 actual.pos.column_number
             ),
-            ParseError::ExpectedExpr(expected, actual) => {
-                write!(f, "Expected expression '{}', got '{}'", expected, actual)
-            }
-            ParseError::UnexpectedKeyword(keyword) => write!(f, "Unexpected keyword: {}", keyword),
-            ParseError::Unexpected(tok, msg) => write!(f, "Unexpected Token {} {}", tok, msg),
+            ParseError::ExpectedExpr(expected, actual, pos) => write!(
+                f,
+                "Expected expression '{}', got '{}' at line {}, col {}",
+                expected, actual, pos.line_number, pos.column_number
+            ),
+            ParseError::UnexpectedKeyword(keyword, pos) => write!(
+                f,
+                "Unexpected keyword: '{}' at line {}, col {}",
+                keyword, pos.line_number, pos.column_number
+            ),
+            ParseError::Unexpected(tok, msg) => write!(
+                f,
+                "Unexpected Token '{}'{} at line {}, col {}",
+                tok,
+                if let Some(m) = msg {
+                    format!(", {}", m)
+                } else {
+                    String::new()
+                },
+                tok.pos.line_number,
+                tok.pos.column_number
+            ),
             ParseError::AbruptEnd => write!(f, "Abrupt End"),
             ParseError::General(msg) => write!(f, "{}", msg),
             ParseError::NormalEOF => write!(f, "EOF"),
@@ -370,12 +388,16 @@ impl Parser {
                 _ => Ok(Node::Return(Some(Box::new(self.parse()?)))),
             },
             Keyword::New => {
+                let start_pos = self.pos;
                 let call = self.parse()?;
                 match call {
                     Node::Call(ref func, ref args) => {
                         Ok(Node::Construct(func.clone(), args.clone()))
                     }
-                    _ => Err(ParseError::ExpectedExpr("constructor", call)),
+                    _ => {
+                        let token = self.get_token(start_pos)?;
+                        Err(ParseError::ExpectedExpr("constructor", call, token.pos))
+                    }
                 }
             }
             Keyword::TypeOf => Ok(Node::TypeOf(Box::new(self.parse()?))),
@@ -490,7 +512,10 @@ impl Parser {
                 let block = self.parse()?;
                 Ok(Node::FunctionDecl(name, args, Box::new(block)))
             }
-            _ => Err(ParseError::UnexpectedKeyword(keyword)),
+            _ => {
+                let token = self.get_token(self.pos - 1)?; // Gets the offending token
+                Err(ParseError::UnexpectedKeyword(keyword, token.pos))
+            }
         }
     }
 
@@ -501,7 +526,7 @@ impl Parser {
         }
         let token = self.get_token(self.pos)?;
         self.pos += 1;
-        let expr: Node = match token.kind {
+        let node: Node = match token.kind {
             TokenKind::Punctuator(Punctuator::Semicolon) | TokenKind::Comment(_)
                 if self.pos < self.tokens.len() =>
             {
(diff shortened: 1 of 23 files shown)


jasonwilliams:
This branch has been rebased against master

jasonwilliams:
Let me know if there’s anyway I can help, not picked up anything yet

Razican:
I fixed this particular error. It turns out that it was reading new lines and not skipping them, so it didn't understand that this was a variable declaration. But I found a new one:
```
ParsingError: Expected one of ';' or 'line terminator', got '=' in routine 'variable declaration' at line 2, col 19
```

jasonwilliams:
Interesting, I’m already getting the feeling this parser is easier to reason with than before

boa/src/syntax/parser/mod.rs
```diff
@@ -202,8 +244,9 @@ impl Parser {
...
     }
 
-    /// Peek the next token and if it is ``kind``, get the next token, return true.
-    /// Otherwise, return false.
+    /// Peek the next token, if it's of `kind` type.
+    ///
+    /// When the next token is a `kind` token, get the token, otherwise return `None`.
     fn next_if(&mut self, kind: TokenKind) -> Option<Token> {
         match self.peek(0) {
             Ok(tok) => {
```
HalidOdat: The bug is here. It parses it correctly `var x = [1, 2, 3];` now.
```suggestion
    /// Gets the next token, if the current token is of `kind` type.
    ///
    /// When the current token is a `kind` token, get the next token, otherwise return `None`.
    fn next_if(&mut self, kind: TokenKind) -> Option<Token> {
        match self.get_current_token() {
            Ok(tok) => {
```

HalidOdat: The previous implementation was more like `if_next` rather than `next_if`!

Any thoughts on the matter? :thinking:

jasonwilliams: Interesting,
So next_if should only consume the next token if it matches, if it doesn’t then it shouldn’t move forward. That was the idea behind it, I guess getting the cure token is fine, but if it matches you should move forward 1.
(On my phone atm) doesn’t peek(0) and get current token effectively do the same thing?

HalidOdat: The implementation of peek is:
```rust
/// peeks at the next token
fn peek(&self, num: usize) -> Result<Token, ParseError> {
   self.get_token(self.pos + 1 + num)
}
```
so even if it is `peek(0)` it gives the next token, not the current.

jasonwilliams: What you have is more correct now that next gets the current token then moves forward

jasonwilliams: So yes you’re right now that next gas changed we need to change this too

HalidOdat: I think we should change the methods `next_if` to `matches` and `next_if_skip_lineterminator` to `matches_skip_lineterminator`. Which better aligns with the implementation of the functions.

jasonwilliams: Matches is a little bit ambiguous to me, it’s not clear that it would advance the parser if there’s a match, it sounds like a function that just checks if there’s a match between 2 things or not.

next_if was clear in the sense it would advance if the value was correct

HalidOdat: > Matches is a little too ambiguous for me, it’s not clear that it would advance the parser if there’s a match, it sounds like a function that just checks if there’s a match between 2 things or not.
> 
> next_if was clear in the sense it would advance if the value was correct

I see! Thats a good point.

Razican:
Cool, this seems to fix this one too. I got a new one with this code:
```javascript
function comp(a) {
    return a == "a";
}
var many = ["a", "b", "c"];
```

It gives:
```
ParsingError: Expected token '[', got '{' in routine 'function declaration' at line 2, col 26
```

You can try it by running `cargo test builtins::array::tests::find`. I will try to fix it during the day.

I'm also getting a panic when running another test, so I'm guessing there is something else wrong here:
```
thread 'builtins::array::tests::find_index' panicked at 'failed to convert to object: Undefined', boa/src/exec/mod.rs:128:35
```

I think we are messing things too much with so much "peek" or "next_if" and so on. Maybe this could be simplified? It seems to give a lot of of-by-one issues, among others. Maybe one options is to implement or to at least mimic the `Iterator` traits from the standard library.

For many of our needs, we could use the [`Iterator`][iterator] trait and the [`Peekable`][peekable] structure, or other implementations in the [`std::iter`][iter] module.

[iterator]: https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html
[peekable]: https://doc.rust-lang.org/stable/std/iter/struct.Peekable.html
[iter]: https://doc.rust-lang.org/stable/std/iter/index.html

jasonwilliams:
> I think we are messing things too much with so much "peek" or "next_if" and so on. Maybe this could be simplified? It seems to give a lot of of-by-one issues, among others. Maybe one options is to implement or to at least mimic the Iterator traits from the standard library.

> For many of our needs, we could use the Iterator trait and the Peekable structure, or other implementations in the std::iter module.

I do want to do this, and was attempting to do this previously but ran into some issues, so i put it on the back burner for now (i think because some methods want to go backwards etc).
I put it outside of the scope of the New Parser because migrating to this and changing to an iterator at the same time cause confusion.

jasonwilliams:
> I think we are messing things too much with so much "peek" or "next_if" and so on. Maybe this could be simplified? It seems to give a lot of of-by-one issues, among others.

Yeah we need to be on the same mental model for this.
So in the old parser getting next was incrementing in the array and returning a value.
In this, getting next pops-off whatever is first in the array, so the other methods need to match up.

HalidOdat:
> Cool, this seems to fix this one too. I got a new one with this code:
> 
> ```js
> function comp(a) {
>     return a == "a";
> }
> var many = ["a", "b", "c"];
> ```
> 
> It gives:
> 
> ```
> ParsingError: Expected token '[', got '{' in routine 'function declaration' at line 2, col 26
> ```

There is a bug when parsing function declarations in function: `read_function_declaration`. Before parsing function body.

```rust
self.expect(
    TokenKind::Punctuator(Punctuator::OpenBracket),
        "expected '{'",
)?;
```
It should be:
```rust
self.expect(
    TokenKind::Punctuator(Punctuator::OpenBlock),
        "expected '{'",
)?;
```

Razican:
I added the fix by @HalidOdat and now that seems to work. Still, it seems that it gets stuck in some kind of infinite loop. So I will put this as ready for review, but we need to check that.

jasonwilliams:
Looks good to me

HalidOdat:
Its from the `Array.prototype.every` test.
More specifically its this code:
```javascript
var empty = [];

var array = [11, 23, 45];

function callback(element) {
    return element > 10;
}

function callback2(element) {
    return element < 10;
}

var appendArray = [1, 2, 3, 4];

function appendingCallback(elem, index, arr) {
    arr.push('new');
    return elem !== "new";
}
```
I'm not sure exactly what part of this code causes the infinite loop.

jasonwilliams:
Would be nice if there's some way of logging each state it ends up in without affecting production builds at all.
Like a logger, Similar to assert_debug() which i believe doesn't get compiled in release builds.

HalidOdat:
This code also goes in an infinite loop.
```javascript
var empty = [ ];
var one = ["a"];
var many = ["a", "b", "c"];
var duplicates = ["a", "b", "c", "a", "b"];
```
I am confused were the origin of the bug is. :confused:

jasonwilliams:
For `var many = ["a", "b", "c"];`
it seems to be skipping over the `[` for the array and trying to assign `a` directly to `many`

Call Stack
```
read_assignment_expression 
read_initializer 
read_variable_declaration 
read_variable_declaration_list 
read_variable_statement 
read_statement 
read_statement_list_item 
read_statements 
read_statement_list 
parse_all 
parser_expr 
forward_val 
main 
```

jasonwilliams:
If we merged this back into the main parser branch i can fix peek and that should fix quite a few errors

Razican:
> If we merged this back into the main parser branch i can fix peek and that should fix quite a few errors

Sure, let's merge this and keep working on the other branch :)

HalidOdat:
The bug that causes the infinite looping is in the function `variable_declaration_continuation`.
The bug was that it did not move forward by one when a new line-terminator, semicolon, or comma was found and it checked the next token not the current one.
```rust
fn variable_declaration_continuation(&mut self) -> Result<bool, ParseError> {
    let mut newline_found = false;

    loop {
        match self.get_current_token() { // <----
            Ok(tok) => match tok.kind {
                TokenKind::LineTerminator => newline_found = true,
                TokenKind::Punctuator(Punctuator::Semicolon) => {
                    return Ok(false);
                }
                TokenKind::Punctuator(Punctuator::Comma) => {
                    return Ok(true);
                }
                _ if newline_found => return Ok(false),
                _ => break,
            },
            Err(_) => return Ok(false),
        }
        self.pos += 1; // <-----
    }

    Err(ParseError::Expected(
        vec![
            TokenKind::Punctuator(Punctuator::Semicolon),
            TokenKind::LineTerminator,
        ],
        self.get_current_token()?,
        "variable declaration",
    ))
}
```

Now it does not cause an infinite loop.
```
test result: FAILED. 63 passed; 63 failed; 1 ignored; 0 measured; 0 filtered out
```
WERE HALF WAY THERE!!! :grinning:

PR #286, opened 2020-03-30
Fixed all tests

HalidOdat:
Fixed all tests from the new parser #281
```
test result: ok. 127 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
```
:tada: :tada: :tada:

### DIFF
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -356,7 +356,12 @@ impl Executor for Interpreter {
                 }
                 _ => Ok(Gc::new(ValueData::Undefined)),
             },
-            Node::Construct(ref callee, ref args) => {
+            Node::New(ref call) => {
+                let (callee, args) = match call.as_ref() {
+                    Node::Call(callee, args) => (callee, args),
+                    _ => unreachable!("Node::New(ref call): 'call' must only be Node::Call type."),
+                };
+
                 let func_object = self.run(callee)?;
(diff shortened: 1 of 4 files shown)


boa/src/exec/mod.rs
```diff
@@ -538,7 +558,10 @@ impl Executor for Interpreter {
 
                 Ok(obj)
             }
-            _ => unimplemented!(),
+            Node::Spread(ref node) => {
+                Ok(Gc::new((*self.run(node)?).clone())) // for now we can do nothing but return the value as-is
```
Razican: I would add a `TODO` somewhere here to search for it easier.

boa/src/exec/mod.rs
```diff
@@ -591,7 +614,7 @@ impl Interpreter {
                                     .environment
                                     .initialize_binding(name, expr.clone());
                             }
-                            Node::UnaryOp(UnaryOp::Spread, ref expr) => {
+                            Node::Spread(ref expr) => {
```
Razican: Why was this moved out of `UnaryOp`?

HalidOdat: In the standard/specification `Spread` is not a part of the unary expression.
see: https://tc39.es/ecma262/#prod-UnaryExpression

jasonwilliams:
Wow, good work!

PR #287, opened 2020-03-30
Added cursor to parser

### LINKED ISSUE #284: Reference tokens in Parser
https://github.com/jasonwilliams/boa/pull/281#issuecomment-605117720
I think these functions should return an Option<&Token>. So we do not have to clone every single time we get a token. And only clone on demand.

Razican:
This adds a simple cursor structure to the new parser (#281), which enables doing simpler operations, and making it easier to understand.

This also brings token referencing to the parser, to avoid so much cloning. Most of the cloning only happens when we need to return an error, in which case, it doesn't matter if it takes us a bit more. But for the rest, I think this should be faster. This closes #284.

I also took the opportunity to improve a bit further the documentation and the parser errors.

### DIFF
--- a/boa/src/syntax/parser/cursor.rs
+++ b/boa/src/syntax/parser/cursor.rs
@@ -28,16 +28,8 @@ impl Cursor {
     }
 
     /// Moves the cursor to the given position.
-    ///
     /// This is intended to be used *always* with `Cursor::pos()`.
     ///
-    /// # Example:
-    /// ```
-    /// # let mut cursor = Cursor::new(Vec::new());
-    /// let pos_save = cursor.pos();
-    /// // Do some stuff that might change the cursor position...
-    /// cursor.seek(pos_save);
-    /// ```
     pub(super) fn seek(&mut self, pos: usize) {
         self.pos = pos
     }
--- a/boa/src/syntax/parser/cursor.rs
+++ b/boa/src/syntax/parser/cursor.rs
@@ -6,7 +6,7 @@ use crate::syntax::ast::token::Token;
 ///
 /// This internal structure gives basic testable operations to the parser.
 #[derive(Debug, Clone, Default)]
-pub struct Cursor {
+pub(super) struct Cursor {
     /// The tokens being input.
     tokens: Vec<Token>,
     /// The current position within the tokens.
@@ -15,15 +15,35 @@ pub struct Cursor {
 
 impl Cursor {
     /// Creates a new cursor.
-    pub fn new(tokens: Vec<Token>) -> Self {
+    pub(super) fn new(tokens: Vec<Token>) -> Self {
         Self {
             tokens,
             ..Self::default()
         }
     }
 
+    /// Retrieves the current position of the cursor in the token stream.
+    pub(super) fn pos(&self) -> usize {
+        self.pos
+    }
+
+    /// Moves the cursor to the given position.
+    ///
+    /// This is intended to be used *always* with `Cursor::pos()`.
+    ///
+    /// # Example:
+    /// ```
+    /// # let mut cursor = Cursor::new(Vec::new());
+    /// let pos_save = cursor.pos();
+    /// // Do some stuff that might change the cursor position...
+    /// cursor.seek(pos_save);
+    /// ```
+    pub(super) fn seek(&mut self, pos: usize) {
+        self.pos = pos
+    }
+
     /// Moves the cursor to the next token and returns the token.
-    pub fn next(&mut self) -> Option<&Token> {
+    pub(super) fn next(&mut self) -> Option<&Token> {
         let token = self.tokens.get(self.pos);
 
         if self.pos != self.tokens.len() {
@@ -34,11 +54,13 @@ impl Cursor {
     }
 
     /// Moves the cursor to the next token after skipping tokens based on the predicate.
-    pub fn next_skip<P>(&mut self, skip: P) -> Option<&Token>
+    pub(super) fn next_skip<P>(&mut self, mut skip: P) -> Option<&Token>
     where
         P: FnMut(&Token) -> bool,
     {
-        while let Some(token) = self.next() {
+        while let Some(token) = self.tokens.get(self.pos) {
+            self.pos += 1;
+
             if !skip(token) {
                 return Some(token);
             }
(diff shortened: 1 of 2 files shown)


boa/src/syntax/parser/mod.rs
```diff
@@ -105,27 +119,25 @@ pub type ParseResult = Result<Node, ParseError>;
...
-    pos: usize,
+    /// Cursor in the parser, the internal structure used to read tokens.
+    cursor: Cursor,
 }
 
 macro_rules! expression { ( $name:ident, $lower:ident, [ $( $op:path ),* ] ) => {
     fn $name (&mut self) -> ParseResult {
         let mut lhs = self. $lower ()?;
-        while let Ok(tok) = self.peek_skip_lineterminator() {
+        while let Some(tok) = self.peek_skip_lineterminator().cloned() {
```
Razican: Could we somehow avoid this? It seems that Rust gives a "multiple mutable reference" error when you return a read-only reference from a mutable reference more than once. Is this something the compiler should understand it's safe?

boa/src/syntax/parser/mod.rs
```diff
@@ -1258,41 +1318,56 @@ impl Parser {
...
             let args = self.read_arguments()?;
             let call_node = Node::Call(Box::new(lhs), args);
 
             Node::New(Box::new(call_node))
         } else {
             self.read_primary_expression()?
         };
-        while let Ok(tok) = self.peek_skip_lineterminator() {
-            match tok.kind {
+        while let Some(tok) = self.peek_skip_lineterminator().cloned() {
```
Razican: Could we somehow avoid this? It seems that Rust gives a "multiple mutable reference" error when you return a read-only reference from a mutable reference more than once. Is this something the compiler should understand it's safe?

boa/src/syntax/parser/mod.rs
```diff
@@ -1305,44 +1380,62 @@ impl Parser {
...
+                .ok_or(ParseError::AbruptEnd)?;
+            return Err(ParseError::Expected(
+                vec![TokenKind::Punctuator(Punctuator::OpenParen)],
+                next_token.clone(),
+                Some("call expression"),
+            ));
         }
 
-        while let Ok(tok) = self.peek_skip_lineterminator() {
+        while let Some(tok) = self.peek_skip_lineterminator().cloned() {
```
Razican: Could we somehow avoid this? It seems that Rust gives a "multiple mutable reference" error when you return a read-only reference from a mutable reference more than once. Is this something the compiler should understand it's safe?

Razican:
It seems that this new implementation is giving a test error:
```
---- exec::tests::spread_with_arguments stdout ----
thread 'exec::tests::spread_with_arguments' panicked at 'assertion failed: `(left == right)`
  left: `"undefined"`,
 right: `"1"`', boa/src/exec/tests.rs:55:5
```

boa/src/syntax/parser/cursor.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// This is intended to be used *always* with `Cursor::pos()`.
+    ///
+    /// # Example:
+    /// ```
+    /// # let mut cursor = Cursor::new(Vec::new());
+    /// let pos_save = cursor.pos();
+    /// // Do some stuff that might change the cursor position...
+    /// cursor.seek(pos_save);
+    /// ```
+    pub(super) fn seek(&mut self, pos: usize) {
```
Razican: Maybe it would make sense to abstract over this and provide a stack with breakpoints. So that we just say `cursor.create_breakpoint()` and then `cursor.to_last_breakpoint()`. But we need to remember to go back in all cases, or to remove the breakpoint if we won't use it anymore.

jasonwilliams: So its worth pointing out that in future i would like to have the lexer send the parser tokens through a channel (most likely Crossbeam), this will allow us to have concurrent parsing (or streaming parsing as V8/Spidermonkey call it).

So you need to imaging we may not even have all the tokens yet, in which case `seek` may fail.

I don't think we need to change things around now, but its worth building with that feature in mind.

jasonwilliams: Moved into an issue here:
https://github.com/jasonwilliams/boa/issues/288

jasonwilliams: > Maybe it would make sense to abstract over this and provide a stack with breakpoints. So that we just say `cursor.create_breakpoint()` and then `cursor.to_last_breakpoint()`. But we need to remember to go back in all cases, or to remove the breakpoint if we won't use it anymore.

I think `seek()`  is fine for now

boa/src/syntax/parser/mod.rs
jasonwilliams: Hmm without checking this out and playing with it I’m not sure. Will need to take a look.

boa/src/syntax/parser/mod.rs
```diff
@@ -1354,100 +1447,106 @@ impl Parser {
...
-            if self
-                .next_if(TokenKind::Punctuator(Punctuator::Spread))
-                .is_some()
-            {
-                args.push(Node::Spread(Box::new(self.read_assignment_expression()?)));
-            } else {
-                args.push(self.read_assignment_expression()?);
             }
+
+            args.push(self.read_assignment_expression()?);
```
HalidOdat: > It seems that this new implementation is giving a test error:
> 
> ```
> ---- exec::tests::spread_with_arguments stdout ----
> thread 'exec::tests::spread_with_arguments' panicked at 'assertion failed: `(left == right)`
>   left: `"undefined"`,
>  right: `"1"`', boa/src/exec/tests.rs:55:5
> ```
This will fix the failing test.
```suggestion

            if self
                .next_if(TokenKind::Punctuator(Punctuator::Spread))
                .is_some()
            {
                args.push(Node::Spread(Box::new(self.read_assignment_expression()?)));
            } else {
                args.push(self.read_assignment_expression()?);
            }
```

Razican:
Thanks @HalidOdat! This fixes the test, and I think this is ready for a merge :)

boa/src/syntax/parser/mod.rs
HalidOdat: > Could we somehow avoid this? It seems that Rust gives a "multiple mutable reference" error when you return a read-only reference from a mutable reference more than once. Is this something the compiler should understand it's safe?

I found a way. If you take a `&[Token]` instead of an owning `Vec<Token>` in `Cursor` and you annotate the returning references with `'a`, so `&'a Token` in functions that return it like `next`, `next_if`, etc. (also in Parser) you can eliminate the `clone`. 
BTW I'm still new to rust lifetimes so there are probably better ways of doing this.

jasonwilliams: Yeah I think this is correct, its the same thing I would have done.

PR #290, opened 2020-03-31
Added more build profiles

Razican:
This will add build profiles for benchmarks and tests. It will also add a thin optimization layer to development builds, and make sure that benchmarks are run with the same build profiles as the release builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,44 +6,24 @@ members = [
 
 # The release profile, used for `cargo build`.
 [profile.dev]
+# Enables thin local LTO and some optimizations.
 opt-level = 1
-debug = true
-debug-assertions = true
-overflow-checks = true
-lto = false
-panic = 'unwind'
-incremental = true
-rpath = false
 
 # The release profile, used for `cargo build --release`.
 [profile.release]
-opt-level = 3
-debug = false
-debug-assertions = false
-overflow-checks = false
-lto = true
-panic = 'unwind'
-incremental = false
+# Enables "fat" LTO, for faster release builds
+lto = "fat"
+# Makes sure that all code is compiled together, for LTO
 codegen-units = 1
-rpath = false
 
 # The test profile, used for `cargo test`.
 [profile.test]
+# Enables thin local LTO and some optimizations.
 opt-level = 1
-debug = true
-debug-assertions = true
-overflow-checks = true
-lto = false
-incremental = true
-rpath = false
 
 # The benchmark profile, used for `cargo bench`.
 [profile.bench]
-opt-level = 3
-debug = false
-debug-assertions = false
-overflow-checks = false
-lto = true
-incremental = false
+# Enables "fat" LTO, for faster benchmark builds
+lto = "fat"
+# Makes sure that all code is compiled together, for LTO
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -6,23 +6,44 @@ members = [
...
 debug = false
-rpath = false
-codegen-units = 1
+debug-assertions = false
+overflow-checks = false
 lto = true
+panic = 'unwind'
+incremental = false
+codegen-units = 1
+rpath = false
```
jasonwilliams: I have a feeling these can go as they're most likely the default anyway, same with opt-level, in fact which ones here are different from the default?

Razican: Here we have the default profiles: https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles

I did maintain some of the defaults, just to make things more clear. I can remove redundant options if you prefer it :)

jasonwilliams: Yeah I think it would be good to remove redundant options, when I see options here I assume we’ve changed them for a reason, so default ones seem odd.

Razican:
I modified the configuration to only show the changed options from the defaults, and to add some comments on why the defaults were changed.

Cargo.toml
```diff
@@ -6,23 +6,24 @@ members = [
...
-incremental = true
-opt-level = 0
-debug = true
-rpath = false
-lto = false
-debug-assertions = true
-overflow-checks = true
-panic = 'unwind'
+# Enables thin local LTO and some optimizations.
+opt-level = 1
```
jasonwilliams, line 10: I’m not sure for dev we want any optimisations, also setting 1 turns off the debug asserts which I think we’re using to get better feedback if something is wrong on dev

Razican: I was trying to find where it said that this would disable debug assertions, but I couldn't find it.

Razican:
Let me know if this is good to go, @jasonwilliams

jasonwilliams:
I will need to test this with the debugger, having an opt-level of 1 might remove some of the information for the LLDB debugger to show properly

jasonwilliams:
Testing this today with the debugger, i can still see all the values i had before, so a level of 1 seems ok, but we still have the debug_assert issue https://doc.rust-lang.org/cargo/reference/profiles.html

Razican:
> Testing this today with the debugger, i can still see all the values i had before, so a level of 1 seems ok, but we still have the debug_assert issue [doc.rust-lang.org/cargo/reference/profiles.html](https://doc.rust-lang.org/cargo/reference/profiles.html)

If I understand the reference properly, the `debug-assertions` is turned on automatically if the optimization level is 0, but if not, it will check for that configuration option, so it should work as expected.

I just checked by adding a `debug_assert!(false)` in the main function and running this with this configuration, and it panics as expected. If I run in `--release` mode, it doesn't panic.

This small optimization level is very fast at compile time, but gives a ton of performance improvements that will be interesting when running a lot of tests, or running example workloads that take long to execute.

Razican:
Let me know, @jasonwilliams if this is good to go, or if we don't want the optimizations on debug/test builds :)

PR #291, opened 2020-03-31
Changed Cursor to take in a &[Token].

HalidOdat:
Changed `Cursor` to take in a `&[Token]` and removed unnecessary *cloning* in expression parsing for new parser #281.


Razican:
Yes! I had not thought about this :) This should remove all unnecessary cloning, and maybe we can even improve in other places as well :)

Razican:
I think the error in the build is due to the need to update the benchmarks code.

HalidOdat:
Yes I fixed it I will push the code right now.

HalidOdat:
I think this branch is ready to merge now. :smile:

PR #293, opened 2020-04-01
Better documentation

HalidOdat:
This is for improving the documentation of the `boa` project.

Any feedback on how we can improve the documentation of the code, would be very helpful.

## Syntax
 - [x] lexer
 - [x] ast
 - - [x] token
 - - [x] constant
 - - [x] keyword
 - - [x] node
   - - [x] Node
   - - [x] MethodDefinitionKind
   - - [x] PropertyDefinition
 - - [x] op
   - - [x] NumOp
   - - [x] UnaryOp
   - - [x] BitOp
   - - [x] CompOp
   - - [x] LogOp
   - - [x] AssignOp
   - - [x] BinOp
 - - [x] pos
 - - [x] punc

## Builtins
 - [x] array
 - [x] boolean
 - [x] function
 - [x] math
 - [x] number
 - [x] object
 - [x] regex
 - [x] string
 - [x] symbol
 - [x] value
 - [x] console
 - [x] error
 - [x] json
 - [x] property


jasonwilliams:
On a similar note I would be interested in having a blog on here, with the typical rust theme (But it doesn’t have to be) that can be statically generated. Then shown from github.io

Right now I’m using my own site

jasonwilliams:
These additions look good, are there any other areas you wanted help on or planned to fill in?

HalidOdat:
> These additions look good, are there any other areas you wanted help on or planned to fill in?

I would like to document all of `syntax` module before merging.

Razican:
> On a similar note I would be interested in having a blog on here, with the typical rust theme (But it doesn’t have to be) that can be statically generated. Then shown from github.io
> 
> Right now I’m using my own site.
> 
> I’ll make an issue for it.

We can add automatic documentation generation to the GitHub pipelines that upload the docs to GitHub pages.

HalidOdat:
> > On a similar note I would be interested in having a blog on here, with the typical rust theme (But it doesn’t have to be) that can be statically generated. Then shown from github.io
> > Right now I’m using my own site.
> > I’ll make an issue for it.
> 
> We can add automatic documentation generation to the GitHub pipelines that upload the docs to GitHub pages.

Thats amazing! I did not know that such a feature existed.

jasonwilliams:
> We can add automatic documentation generation to the GitHub pipelines that upload the docs to GitHub pages.

We already have the docs.rs documentation which automatically generates on every release, I don’t understand what the github pages would add? I may be missing the idea here though.

boa/src/syntax/ast/op.rs
```diff
@@ -199,13 +529,38 @@ impl Display for CompOp {
     }
 }
 
-/// A logical operation between 2 boolean values
+/// Logical operators are typically used with Boolean (logical) values; when they are, they return a Boolean value.
+///
+/// However, the && and || operators actually return the value of one of the specified operands,
```
Razican, line 534:
```suggestion
/// However, the `&&` and `||` operators actually return the value of one of the specified operands,
```

boa/src/syntax/ast/op.rs
```diff
@@ -199,13 +529,38 @@ impl Display for CompOp {
...
+///
+/// More information:
+///  - ECMAScript reference: <https://tc39.es/ecma262/#sec-binary-logical-operators>.
+///  - MDN documentation:
+/// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Logical>
 #[cfg_attr(feature = "serde-ast", derive(Serialize, Deserialize))]
 #[derive(Clone, Debug, Trace, Finalize, PartialEq)]
 pub enum LogOp {
-    /// `a && b` - Logical and
+    /// The logical AND operator returns the value of the first operand if it can becoerced into `false`;
```
Razican, line 544:
```suggestion
    /// The logical AND operator returns the value of the first operand if it can be coerced into `false`;
```

boa/src/syntax/ast/op.rs
```diff
@@ -158,25 +395,118 @@ impl Display for BitOp {
...
+/// The operands can be numerical, string, logical, or object values. Strings are compared based on standard
+/// lexicographical ordering, using Unicode values. In most cases, if the two operands are not of the same type,
+/// JavaScript attempts to convert them to an appropriate type for the comparison. This behavior generally results in
+/// comparing the operands numerically. The sole exceptions to type conversion within comparisons involve the `===` and `!==`
+/// operators, which perform strict equality and inequality comparisons. These operators do not attempt to convert the operands
+/// to compatible types before checking equality.
+///
+/// More information:
+///  - MDN documentation:
+/// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Comparison>
```
Razican, line 409: Maybe we could add the "Testing and Comparison Operations" link as an spec link: https://tc39.es/ecma262/#sec-testing-and-comparison-operations

boa/src/syntax/ast/op.rs
```diff
@@ -16,21 +16,79 @@ pub trait Operator {
     }
 }
 
-/// A numeric operation between 2 values
+/// Arithmetic operators take numerical values (either literals or variables)
+/// as their operands and return a single numerical value.
+///
+/// More information:
+///     - MDN documentation:
+/// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Arithmetic>
```
Razican, line 24: Given that links are too long, maybe we could shorten them:
```suggestion
///     - [MDN documentation][mdn]
/// 
/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Arithmetic
```

Also, do we have a spec link for this? I tried to find something suitable, but didn't find anything :/

boa/src/syntax/ast/op.rs
HalidOdat: That might be the best choice.

Razican:
> We already have the docs.rs documentation which automatically generates on every release, I don’t understand what the github pages would add? I may be missing the idea here though.

`docs.rs` documentation will not include private items such as these ones, only public documentation. It might be interesting to have an internal page that adds documentation for all private interfaces as well.

boa/src/syntax/ast/op.rs
HalidOdat: I couldn't find a spec link.

HalidOdat:
> > We already have the docs.rs documentation which automatically generates on every release, I don’t understand what the github pages would add? I may be missing the idea here though.
> 
> `docs.rs` documentation will not include private items such as these ones, only public documentation. It might be interesting to have an internal page that adds documentation for all private interfaces as well.

For newcomers to the project, it would be great!

Razican:
About this particular PR, I would put the links without showing the full URL, as we did for the first case, since URLs are too long in my opinion. I'm doing the same with the new parser subdivision.

boa/src/syntax/ast/op.rs
```diff
@@ -51,32 +116,150 @@ impl Display for NumOp {
...
+    ///  - [ECMAScript reference](https://tc39.es/ecma262/#sec-bitwise-not-operator)
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_NOT
     Tilde,
-    /// `typeof` - Get the type of object
+
+    /// The `typeof` operator returns a string indicating the type of the unevaluated operand.
+    ///
+    /// Syntax: `typeof expression` or `typeof(expression)`
```
Razican, line 246: Since we are using this syntax above...
```suggestion
    /// Syntax: `typeof x` or `typeof(x)`
```

boa/src/syntax/ast/op.rs
```diff
@@ -51,32 +116,150 @@ impl Display for NumOp {
...
+    /// More information:
+    ///  - [ECMAScript reference](https://tc39.es/ecma262/#sec-typeof-operator)
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof
     TypeOf,
+
     /// The JavaScript `delete` operator removes a property from an object.
     ///
+    /// Syntax: `delete expression`
```
Razican, line 261: Same as above:
```suggestion
    /// Syntax: `delete x`
```

boa/src/syntax/ast/op.rs
```diff
@@ -86,11 +269,17 @@ pub enum UnaryOp {
...
+    /// More information:
+    ///  - [ECMAScript reference](https://tc39.es/ecma262/#sec-delete-operator)
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete
     Delete,
 
     /// The `void` operator evaluates the given `expression` and then returns `undefined`.
     ///
+    /// Syntax: `void expression`
```
Razican, line 281: Same as above:
```suggestion
    /// Syntax: `void x`
```

Razican:
> I would like to document all of `syntax` module before merging.

About this, I would leave the `parser` module for a new PR. I'm currently working on dividing the parser in smaller parsers in new modules in the branch `parser_modularization`, and I'm adding full documentation there.

HalidOdat:
> 

> > I would like to document all of `syntax` module before merging.
> 
> About this, I would leave the `parser` module for a new PR. I'm currently working on dividing the parser in smaller parsers in new modules in the `parser_modularization` branch, and I'm adding full documentation there.

HalidOdat:
I accidentally hit the `close pull request`, thats a facepalm moment right there. :man_facepalming:

HalidOdat:
I don't think `pos` needs more documentation. any thoughts?

Razican:
> I don't think `pos` needs more documentation. any thoughts?

I would add some module-level documentation in both `op` and `pos`. Something like:

```rust
//! This module implements the `Pos` structure, which represents a position in the source code.
```

HalidOdat:
@Razican and @jasonwilliams I would like your feedback on this.
The documentation I put in `Keyword` should I put it in `Node`?
If so What should i put in `Keyword`?

jasonwilliams:
Will try and look today

Razican:
> The documentation I put in `Keyword` should I put it in `Node`?

I've been reading the documentation a bit, and maybe it makes more sense in `Node`, yes.

> If so what should i put in `Keyword`?
> Or should I write the same documentation in Node as well.

I wouldn't duplicate documentation, it seems easy to mess up. Maybe we could say something like this:

```rust
/// The else keyword.
///
/// This keyword precedes an `else` block related to an `if` block. For more information:
/// - [If `Node` documentation][link_to_if_node]
/// - [MDN documentation][same_link_as_now]
/// - [ECMAScript specification][same_link_as_now]
```

What do you think?

HalidOdat:
Thanks for the feedback! That really helped! :)

jasonwilliams:
Have you been running `cargo doc` ?
There have been a lot of failures. 

I've added `missing_doc_code_examples` to allowed but there's still some errors

jasonwilliams:
I've been looking at this through the rendered doc that cargo doc generated.
I agree with what @Razican suggested

HalidOdat:
Why do we have `GetConstField` and `GetField`?. They do the same operation

jasonwilliams:
> Why do we have `GetConstField` and `GetField`?. They do the same operation

From what i understand they don't.
GetConstField will fetch you the field from the string property of an object

Example:
`foo.bar` bar is already a string, so it just does a simple lookup

getField will evaluate what's inside the square brackets then fetches the resolved value as a field
`foo[someValue]`

someValue will be called first, then the (string) result of that it used for the look up.

HalidOdat:
Interesting. So It is an optimization?

jasonwilliams:
Are there any I can start to fill in?

HalidOdat:
> Are there any I can start to fill in?

If you could fill in `lexer` that be awesome. :)

Razican:
I think this is almost ready for a merge, right? We could merge it as-is, and maybe create an extra issue with the remaining bits :)

HalidOdat:
> I think this is almost ready for a merge, right? We could merge it as-is, and maybe create an extra issue with the remaining bits :)

I think so too. the `Environment` and `Exec` has some documentation already.

Razican:
We now have the development docs in https://jasonwilliams.github.io/boa/doc/

We should probably include a link in the Readme in this PR, as it will greatly improve it :)

HalidOdat:
I think it's ready to merge now. :)

boa/src/builtins/regexp/mod.rs
```diff
@@ -269,16 +404,33 @@ pub fn r#match(this: &Value, arg: String, ctx: &mut Interpreter) -> ResultValue
...
+///
+/// The `[@@matchAll]` method returns all matches of the regular expression against a string.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-regexp-prototype-matchall
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@matchAll
+// TODO: it's returning an array, it should return an iterator
```
Razican, line 433: We should probably add issues to the repo with these things.

HalidOdat: Yep. I thought of sifting through the code to find the `// TODO:` and `// FIXME:` and creating issues for them.

jasonwilliams:
Did we want to add a frontpage explanation of syntax, builtins and the ones missing?

image

jasonwilliams:
Looks great!
Thanks for the monsterous effort on that

PR #298, opened 2020-04-02
Use jemallocator

Razican:
I have changed the system allocator for Jemallocator as the global allocator for the binary and the benchmarks. Jemalloc makes the binary bigger, but the performance should increase nicely, which is something we probably want in a compiler.

My benchmarks show around 10% improvement or more in some cases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,6 +7,7 @@ dependencies = [
  "criterion",
  "gc",
  "gc_derive",
+ "jemallocator",
  "rand",
  "regex",
  "serde",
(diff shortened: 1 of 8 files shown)


jasonwilliams:
What’s the reason Jemallocator makes the binaries bigger?
(I know very little about jemalloc)

jasonwilliams:
Why are the benchmarks stuck hmm

Razican:
I cancelled, but it seems I cannot run it again. Still, it seems it passed two times, right?

Razican:
Should we merge this?

Issue #294, opened 2020-04-01
The lexer doesn't take into account goal symbols


Razican:
Getting the code from the `harness/assert.js` file in the test262 suite, this fails to parse:
```javascript
function assert(mustBeTrue, message) {
  if (mustBeTrue === true) {
    return;
  }

  if (message === undefined) {
    message = 'Expected true but got ' + assert._toString(mustBeTrue);
  }
  $ERROR(message);
}

assert._isSameValue = function (a, b) {
  if (a === b) {
    // Handle +/-0 vs. -/+0
    return a !== 0 || 1 / a === 1 / b;
  }

  // Handle NaN vs. NaN
  return a !== a && b !== b;
};
```

With the error:
```
ParsingError: Expected one of ';', '}' or 'line terminator', got 'b' at line 24, col 24
```

Interestingly, the error is given in column 24 (marked here):
```
    return a !== 0 || 1 / a === 1 / b;
                       ^
```

But I'm guessing the issue is with the last `b`. I think this is a regression, since this particular file parsed fine before the rewrite, but it might have been something new in the file. Once fixed, we need to add a test for it.

Razican:
I'll work on fixing this. Ideally, this should be fixed before Boa 0.7.

HalidOdat:
After some debugging the error seems to be coming form:
```javascript
1 / a === 1 / b
```

Checking the token stream: `cargo run -- --dump-tokens`
```
[
    Token {
        kind: NumericLiteral(
            1.0,
        ),
        pos: Position {
            column_number: 1,
            line_number: 1,
        },
    },
    Token {
        kind: RegularExpressionLiteral(
            " a === 1 ",
            "",
        ),
        pos: Position {
            column_number: 3,
            line_number: 1,
        },
    },
	...
]
```
It seems to be a lexer bug not a parser one. It parses `/ a === 1 /` as a regex.

Hope that helps. :)

jasonwilliams:
Good find! 
I do like the ast and token output we have now

jasonwilliams:
Do we need to refactor the lexer now? 😂

HalidOdat:
I'm not exactly sure what the lexer is supposed to do in this situation since.
since `/ a === 1 /` is a regex. but not in this context `1 / a === 1 / b`.

HalidOdat:
we probably need a context aware lexer or something like that. any thoughts?

Razican:
[...]

PR #304, opened 2020-04-10
Modularized parser

Razican:
This is still work in progress, but tests pass, and even though I need to divide the code further, we now have multiple different structures. I also need to document a bunch of stuff.

I also modified the parser so that it would better respect the specification. So for example, an initializer expects to first read a `=` symbol, as in the [spec][initializer].

I added more tests for `break` `throw` and `continue` parsing, that I think they were failing with automatic semicolon insertion.

About this last part, it gets done at the `Cursor` level. I think I changed all semicolon expects for the new `expect_semicolon()` function. This will automatically insert semicolons in all the situations where the [spec][semicolon] says so. The only exception is the semicolon after the `do while` statement, so this should fail:

```javascript
var i = 0;
do {console.log("hello");} while(i++ < 10) console.log("end");
```

In any case, do-while parsing is not done, so this doesn't affect us. It's now detected with a panic, though.

In any case, this is advanced enough so that you can give it a look and let me know what you think.

[initializer]: https://tc39.es/ecma262/#prod-Initializer
[semicolon]: https://tc39.es/ecma262/#sec-automatic-semicolon-insertion


boa/src/syntax/parser/expression/mod.rs
```diff
@@ -0,0 +1,875 @@
...
+};
+
+/// Generates an expression parser.
+///
+/// This macro has 2 mandatory identifiers:
+///  - The `$name` identifier will contain the name of the parsing structure.
+///  - The `$lower` identifier will contain the parser for lower level expressions.
+///
+/// Those exressions are divided by the punctuators passed as the third (and following) parameters.
+macro_rules! expression { ( $name:ident, $lower:ident, $( $op:path ),* ) => {
```
HalidOdat, line 30: why did you remove `[`...`]` form expression?

HalidOdat:
The tests are failing because you need to import `MethodDefinitionKind`, `PropertyDefinition`, `AssignOp`, `BinOp`, `NumOp` and `UnaryOp` in `test.rs`

HalidOdat:
It might be a good idea to refactor the `AST` too, to reflect the changes.
So we don't have one node type `Node`, but a `Statement`, `Declaration` and `Expression` node type. Just like [V8 AST](https://github.com/v8/v8/blob/master/src/ast/ast.h).

Any thoughts @Razican and @jasonwilliams? :thinking:

Razican:
> It might be a good idea to refactor the `AST` too, to reflect the changes.
> So we don't have one node type `Node`, but a `Statement`, `Declaration` and `Expression` node type. Just like [V8 AST](https://github.com/v8/v8/blob/master/src/ast/ast.h).
> 
> Any thoughts @Razican and @jasonwilliams? :thinking:  

When working on iterative statements, I did notice that we were not following the spec 100%, as we directly parse them instead of using an intermediate parser for iterative statements.

It can be a good idea to match the V8 engine in AST node types.

boa/src/syntax/parser/expression/mod.rs
```diff
@@ -0,0 +1,875 @@
...
+            while let Some(tok) = cursor.peek_skip_lineterminator() {
+                match tok.kind {
+                    // Parse assign expression
+                    TokenKind::Punctuator(ref op) if op == &Punctuator::Assign => {
+                        let _ = cursor.next_skip_lineterminator().expect("token disappeared");
+                        lhs = Node::Assign(
+                            Box::new(lhs),
+                            Box::new($lower::parse(cursor)?)
+                        )
+                    }
```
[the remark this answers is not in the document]
Razican: Good question, I almost didn't change anything from what we had before. I will try to see if these things can be improved.

boa/src/syntax/parser/expression/mod.rs
```diff
@@ -0,0 +1,875 @@
...
+            TokenKind::Punctuator(Punctuator::OpenParen) => {
+                if let Some(node) = ArrowFunction::try_parse(cursor) {
+                    return Ok(node);
+                }
+            }
+            _ => {}
+        }
+
+        let mut lhs = ConditionalExpression::parse(cursor)?;
+        // let mut lhs = self.read_block()?;
```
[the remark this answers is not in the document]
Razican: I'll check what's happening here :)

boa/src/syntax/parser/expression/mod.rs
Razican: I thought this made it easier to understand. What do you think? It's pretty easy to put it back, but I wanted to try how it looked like this.

HalidOdat: I see. I think separating the parameters of the `expression` macro is better. So its not a continuous line of parameters.
I would changed even more to be:
```rust
macro_rules! expression {
	(
	  name: $name:ident,
	  call: $lower:ident,
	  matches: [$( $op:path ),*]
	)
	...
}
```
This way it is clear to see that the name of the expression, what it will call and what it matches against.
What do you think? @Razican 
Am I overdoing it a bit?

jasonwilliams:
> > It might be a good idea to refactor the `AST` too, to reflect the changes.
> > So we don't have one node type `Node`, but a `Statement`, `Declaration` and `Expression` node type. Just like [V8 AST](https://github.com/v8/v8/blob/master/src/ast/ast.h).
> > Any thoughts @Razican and @jasonwilliams? 🤔
> 
> When working on iterative statements, I did notice that we were not following the spec 100%, as we directly parse them instead of using an intermediate parser for iterative statements.
> 
> It can be a good idea to match the V8 engine in AST node types.

Yeah i agree with this.
I think what we had was good for getting us to this point, we didn't have any sort of pre-parser or intermediate parser so i think it was just easier to parse as we saw it. Obviously the end game would to be more closer to how V8 deals with it.

Razican:
I will review that. I'm not sure if it would make sense for the `TokenParser` trait to be a method instead of a member function (and therefore accept `&self`.

This would allow not having dangling functions and all parsers would implement `TokenParser`. Then an extra thing that we could add here would be to be able to return a `Result<dyn Executable, ParseError>`, where `Executable` would be a trait that gets passed an executor and executes the node. So we can implement it for AST nodes and also for things like `Vec<Node>` or so.

BTW, I added a `try_parse()` method for `TokenParser` that just tries to parse something, and then goes back to where it was if there was an error parsing it. Currently done by arrow function parsing in expressions, if I'm not mistaken.

Then, there is another thing to be considered: The fact that the lexer doesn't take into account goal symbols (#294). This could be solved by passing the `Lexer` as a parameter to the `Cursor` constructor, and adding a method in the `Cursor` to set the current goal.

This would mean that we no longer can benchmark lexing and parsing separately, as they would happen at the same time, but I think it makes sense.

What do you think?

boa/src/syntax/parser/expression/mod.rs
Razican: Hmmm the thing is that the `name` would be a `struct` or an `enum` to implement `TokenParser` in, so maybe a better name would suit. I will implement it this way for now, though, to see how it goes.

boa/src/syntax/parser/expression/mod.rs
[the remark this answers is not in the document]
Razican: Could it be related to the `Expression` definition in the [spec][spec]? Maybe @jasonwilliams knows more.

[spec]: https://tc39.es/ecma262/#prod-Expression

boa/src/syntax/parser/expression/mod.rs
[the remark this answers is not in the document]
Razican: The issue with this is that I'm not 100% sure on how is this done. An `AssignmentExpression` can be one of these:

 - `ConditionalExpression`
 - `YieldExpression`
 - `ArrowFunction`
 - `AsyncArrowFunction`
 - `LeftHandSideExpression` `=` `AssignmentExpression`
 - `LeftHandSideExpression` `AssignmentOperator` `AssignmentExpression`

The thing is that we do test for `ArrowFunction` at the beginning, and then, we directly parse a `ConditionalExpression`, but I don't understand why. Can this be explained a bit more, @jasonwilliams? maybe I can add some comments once I understand it properly.

boa/src/syntax/parser/expression/mod.rs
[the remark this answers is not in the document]
jasonwilliams: eugh not sure now i think it made sense at the time, will need to take a deeper look

boa/src/syntax/parser/expression/mod.rs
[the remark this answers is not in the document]
jasonwilliams: im not sure i understand what the question is here?

jasonwilliams:
> I will review that. I'm not sure if it would make sense for the `TokenParser` trait to be a method instead of a member function (and therefore accept `&self`.

I don't mind too much what you have now, but i'd like to see this alternative you mention (even if its just for 1 function) to get an idea of what you mean.

> This would allow not having dangling functions and all parsers would implement `TokenParser`. Then an extra thing that we could add here would be to be able to return a `Result<dyn Executable, ParseError>`, where `Executable` would be a trait that gets passed an executor and executes the node. So we can implement it for AST nodes and also for things like `Vec<Node>` or so.
> 
> BTW, I added a `try_parse()` method for `TokenParser` that just tries to parse something, and then goes back to where it was if there was an error parsing it. Currently done by arrow function parsing in expressions, if I'm not mistaken.

Looks good!
> Then, there is another thing to be considered: The fact that the lexer doesn't take into account goal symbols (#294). This could be solved by passing the `Lexer` as a parameter to the `Cursor` constructor, and adding a method in the `Cursor` to set the current goal.

Does the specification discus much about goal symbols?
Would we send the lexer the current state we're in? Then the lexer has context of what we're looking for.

> This would mean that we no longer can benchmark lexing and parsing separately, as they would happen at the same time, but I think it makes sense.
> 
> What do you think?

Yeah if there's no solution i'm fine with changing the benchmarks

Razican:
> > I will review that. I'm not sure if it would make sense for the `TokenParser` trait to be a method instead of a member function (and therefore accept `&self`.
> 
> I don't mind too much what you have now, but i'd like to see this alternative you mention (even if its just for 1 function) to get an idea of what you mean.

The idea is that we would change how the parsers are called now:
```rust
let expr = ExpressionParser::parse(cursor)?;
```

To something like this:
```rust
let expr = ExpressionParser(cursor).parse()?;
```

But this would allow having functions that require some extra information, such as `read_binding_list()` to actually be a `TokenParser`:

```rust
let binding_list = BindingList(cursor, is_const).parse()?;
```

where `BindingList` would be defined like this:

```rust
struct BindingList<'b>(&'b Cursor<'_>, bool);
```
> Would we send the lexer the current state we're in? Then the lexer has context of what we're looking for.

Yes, I think that the lexer's `next()` function should be a function that receives the `goal_symbol` as a parameter, and it should be one of the symbols declared in the spec. I'm not so sure yet how to find out the current symbol in the parser.

Also, I think the cursor should always skip line terminators, and have the `expect_semicolon()` function insert the actual semicolon if we are in a case of automatic semicolon insertion. This is mostly done, except for the line terminator skipping part. I think it would make the parser easier to understand without sacrificing the lexer.

The only special case where this might be a bit difficult would be the `)` token after the `while` condition in a do-while statement. This is pretty difficult to know from the lexer/cursor directly, so I would suggest adding a `bool` as a parameter to that function, to make sure it checks if the previous token is a `)` in the case of a do-while.

boa/src/syntax/parser/expression/mod.rs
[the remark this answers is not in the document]
Razican: It's just that the `read_block()` call is commented out, not sure why, and more generally, how does this parsing work? It seems to only check the conditional expression, but it should check more things, right?

boa/src/syntax/parser/expression/mod.rs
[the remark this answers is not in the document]
Razican: It seems that removing this solved #308 in #311.

Razican:
> The idea is that we would change how the parsers are called now:

The only thing I'm worried about is if this will make the code less performant, having to create an object first and calling its methods, instead of calling a static function with some arguments. I can try it as soon as I finish documenting this first iteration.

Razican:
I added some nice methods (WIP) to easily create AST nodes. This will help make tests more clear, and should also help with the parser implementation. So you can now create a local node like this:

```rust
Node::local("a")
```

instead of:
```rust
Node::Local(String::from("a"));
```

and you can create a `Break` node like this:
```rust
Node::break_node("label");
```

instead of:
```rust
Node::Break(Some(String::from("label")));
```

Also no more need to add `Box::new()` everywhere, so it should make things cleaner.

Razican:
I implemented automatic semicolon insertion after do-while statements. This means we have less things left for this to be mergeable:

 - [ ] Add proper documentation to all parsers.
 - [ ] Make parsers return something `Executable`.
 - [ ] Make the executor be able to execute anything `Executable`.
 - [ ] Divide parser in smaller submodules.

Razican:
Something I see from the benchmarks is that more than 99% of the time is spent on execution. I think we should improve our interpreter :)

jasonwilliams:
I’m going to continue playing with measureme which should give us some fine grained measurements for whichever functions we want to time. We should then be able to get a chrome dev-tools style flame graph of what’s happening.

That’s the theory anyway

Razican:
I think this is now ready for review. I have some documentation left in some parsers, but this looks the way I think it should look.

All parsers now implement `TokenParser`, which has an associated type to determine the result it should be generating. It will also require the parser to be `Sized`, and parsing consumes the parser. Still, Almost all parsers implement `Copy`.

Then, I added some type-checked booleans that represent the `Yield`, `Await`, `In` and so on in the spec. This controls whether a parser will allow these keywords inside of it ([example](https://tc39.es/ecma262/#prod-Declaration)). This can add some extra time to the parsing, but it prepares it for many future implementations of the spec.

I moved tests to submodule, and noticed that we don't have tests for parsing `try...catch` statements, or `if` statements, for example. I divided some tests and added some more, so we now have 184 tests.

I also added some nice `Into` implementations and removed specialized versions of the `expect()` function in the cursor, in order to use generics.

I think this can be merged as is (with any further suggestion) in order to allow for new PRs to use the new modular parser.

In the future, I would like for the `Cursor` to ignore `LineTerminator`s on `next()`, `peek()` or `expect()`, since that should be the default behaviour. And we already have `expect_semicolon()`, that will do automatic semicolon insertion. This would need the addition of `expect_no_lineterminator()` and `peek_semicolon()`. The first would make sure that the next token is not a line terminator, while the later would check if the next would be a semicolon, inserting the semicolon automatically if needed.

Razican:
Hmmm for loop parser benchmarks are a bit worrying. Maybe we can somehow profile what's happening. Other parsing benchmarks don't seem to be affected.

Razican:
I didn't have the time yet to profile this, but if it's related to the new booleans, we can move them to compile-time checks and add generics.  That should solve the creation of new structures on demand with the cost of less efficient cache usage. But with these benchmarks, I think it will be worth it.

Razican:
It seems I fixed the issue with the performance, so it's ready for review :)

boa/src/syntax/parser/function_class/mod.rs
```diff
@@ -0,0 +1,18 @@
+//! Functions and classes parsing.
```
HalidOdat, line 1: Why is there a `function_class` module? Shouldn't it just be `function_declaration`?

Furthermore shouldn't  `function_declaration` be in a module `hoistable_declaration` in `declaration` module?

[Declaration spec][spec]

[spec]: https://tc39.es/ecma262/#prod-Declaration

Razican:
> Looks awesome! The comments are for my understanding. :)

So, I divided the modules taking into account the index of the spec, depending on where something was defined. For the case of [`ArrowFunction`][arrow_function], for example, it's defined in section 14.2, and inside of section 14 (functions and classes). So I put it inside of the `function_class` module, and inside of the `arrow_function` submodule.

That's also the reason why the `function_class` module exists (even though we don't have any parsing done for classes yet.

[arrow_function]: https://tc39.es/ecma262/#prod-ArrowFunction

HalidOdat:
> So, I divided the modules taking into account the index of the spec, depending on where something was defined. For the case of [`ArrowFunction`](https://tc39.es/ecma262/#prod-ArrowFunction), for example, it's defined in section 14.2, and inside of section 14 (functions and classes). So I put it inside of the `function_class` module, and inside of the `arrow_function` submodule.
> 
> That's also the reason why the `function_class` module exists (even though we don't have any parsing done for classes yet.

But wouldn't it be clearer to divide depending on the parsing rules.
Like:
```
 Declaration[Yield, Await]:
	HoistableDeclaration[?Yield, ?Await, ~Default]
    ClassDeclaration[?Yield, ?Await, ~Default]
    LexicalDeclaration[+In, ?Yield, ?Await]
``` 
to:
```
declaration module
	hoistable submodule
    class submodule
    lexical submodule
```

Razican:
> But wouldn't it be clearer to divide depending on the parsing rules/productions.

The problem with this is that the production can be used by many rules. For example, [`IterationStatement`](https://tc39.es/ecma262/#prod-IterationStatement) also has a `LexicalDeclaration`. The only place where they are not repeated is where they are defined, so I used that.

Do you think we could find a clear way of doing this division directly like this?

HalidOdat:
> > But wouldn't it be clearer to divide depending on the parsing rules/productions.
> 
> The problem with this is that the production can be used by many rules. For example, [`IterationStatement`](https://tc39.es/ecma262/#prod-IterationStatement) also has a `LexicalDeclaration`. The only place where they are not repeated is where they are defined, so I used that.
> 
> Do you think we could find a clear way of doing this division directly like this?

Hmmm. I think in the case of [`IterationStatement`](https://tc39.es/ecma262/#prod-IterationStatement) using `LexicalDeclaration`.
`Lexical`**`Declaration`** should still go in `declaration` module. since it is a declaration and we do the same with other production rules.

What do you think of this solution? @Razican

Razican:
> What do you think of this solution? @Razican

It's good for this particular case, yes, I will try to do it like this, and see where it brings me.

Razican:
I made the requested changes, and some more. Now the cursor ignores new lines by default, but we can `expect_no_lineterminator()` and `expect/peek_semicolon()`. The first will mak sure there is no line terminator, and the second will do automatic semicolon insertion.

This still needs a rebase, that will take me a bit since there are some changes to the parser in master.

Razican:
BTW, I almost forgot. I changed the debugger executable name, that had not been changed after the binary name update. Then, I had to revert the optimizations for dev builds, as it was optimizing out some edge cases that were interesting from the debugging perspective.

boa/src/syntax/parser/expression/assignment_operator/mod.rs
```diff
@@ -0,0 +1,122 @@
+//! Assignment operator parsing.
```
HalidOdat, line 1: Shouldn't this be `assignment_expression` instead of `assignment_operator`?

Maybe we could shortened down to just `assignment`, since it's in `expression` module.

boa/src/syntax/parser/expression/conditional_operator.rs
```diff
@@ -0,0 +1,73 @@
+//! Conditional operator parsing.
```
HalidOdat, line 1: shouldn't this live in `expression/assignment`?
[spec](https://tc39.es/ecma262/#prod-AssignmentExpression)

boa/src/syntax/parser/expression/exponentiation_operator.rs
```diff
@@ -0,0 +1,92 @@
+//! Exponentiation operator parsing.
```
HalidOdat, line 1: shouldn't this live in `expression/assignment`

boa/src/syntax/parser/expression/lhs_expression/call_expression.rs
```diff
@@ -0,0 +1,93 @@
+use super::arguments::Arguments;
```
HalidOdat, line 1: Missing documentation.
```suggestion
//! Call expression parsing.
//!
//! More information:
//!  - [MDN documentation][mdn]
//!  - [ECMAScript specification][spec]
//!
//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions
//! [spec]: https://tc39.es/ecma262/#prod-CallExpression

use super::arguments::Arguments;
```

boa/src/syntax/parser/expression/unary_operator.rs
```diff
@@ -0,0 +1,77 @@
+//! Unary operator parsing.
```
HalidOdat, line 1: Maybe we should call it just `unary` module instead of `unary_expression`. It already is in `expression` module so the expression part is implied.

boa/src/syntax/parser/statement/block.rs
```diff
@@ -0,0 +1,201 @@
+//! <https://tc39.es/ecma262/#sec-block>
```
HalidOdat, line 1: Better documentation.
```suggestion
//! Block statement parsing.
//!
//! More information:
//!  - [MDN documentation][mdn]
//!  - [ECMAScript specification][spec]
//!
//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block
//! [spec]: https://tc39.es/ecma262/#sec-block
```

boa/src/syntax/parser/statement/break_statement/mod.rs
```diff
@@ -0,0 +1,69 @@
+#[cfg(test)]
```
HalidOdat, line 1: Missing documentation.
```suggestion
//! Break expression parsing.
//!
//! More information:
//!  - [MDN documentation][mdn]
//!  - [ECMAScript specification][spec]
//!
//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break
//! [spec]: https://tc39.es/ecma262/#prod-BreakStatement

#[cfg(test)]
```

boa/src/syntax/parser/statement/continue_statement/mod.rs
```diff
@@ -0,0 +1,69 @@
+#[cfg(test)]
```
HalidOdat, line 1: Missing documentation.
```suggestion
//! Continue expression parsing.
//!
//! More information:
//!  - [MDN documentation][mdn]
//!  - [ECMAScript specification][spec]
//!
//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/continue
//! [spec]: https://tc39.es/ecma262/#prod-ContinueStatement

#[cfg(test)]
```

boa/src/syntax/parser/statement/declaration/hoistable_declaration.rs
```diff
@@ -0,0 +1,118 @@
+//! Hoistable declaration parsing.
```
HalidOdat, line 1: Maybe we should call this just `hoistable` module. It already lives in `declaration` module.

boa/src/syntax/parser/statement/declaration/lexical_declaration.rs
```diff
@@ -0,0 +1,172 @@
+//! Lexical declaration parsing.
```
HalidOdat, line 1: Maybe we should call this just `lexical` module. It already lives in declaration module, so is implied that it's a declaration.

boa/src/syntax/parser/statement/declaration/mod.rs
```diff
@@ -0,0 +1,55 @@
+//! <https://tc39.es/ecma262/#sec-declarations-and-the-variable-statement>
```
HalidOdat, line 1: Missing Documentation.
```suggestion
//! Declaration parsing.
//!
//! More information:
//!  - [MDN documentation][mdn]
//!  - [ECMAScript specification][spec]
//!
//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements#Declarations
//! [spec]:https://tc39.es/ecma262/#sec-declarations-and-the-variable-statement
```

boa/src/syntax/parser/statement/iteration/do_while_statement.rs
```diff
@@ -0,0 +1,73 @@
+use crate::syntax::{
```
HalidOdat, line 1: Missing documentation.
```suggestion
//! Do-while statement parsing.
//!
//! More information:
//!  - [MDN documentation][mdn]
//!  - [ECMAScript specification][spec]
//!
//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/do...while
//! [spec]: https://tc39.es/ecma262/#sec-do-while-statement

use crate::syntax::{
```

boa/src/syntax/parser/statement/iteration/for_statement.rs
```diff
@@ -0,0 +1,93 @@
+use crate::syntax::{
```
HalidOdat, line 1: Missing documentation.
```suggestion
//! For statement parsing.
//!
//! More information:
//!  - [MDN documentation][mdn]
//!  - [ECMAScript specification][spec]
//!
//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for
//! [spec]: https://tc39.es/ecma262/#sec-for-statement

use crate::syntax::{
```

Razican:
I did al the requested changes, @HalidOdat, I will rebase today. Still, the `expression` module is a bit strange, since it has a tree of expressions.

HalidOdat:
> Still, the `expression` module is a bit strange, since it has a tree of expressions.

Well I think that's good, the parser in a way should reflect what it generates.
And it's easier now to look at the spec productions and reason were something is supposed to go or be.

In the future we should refactor the AST too.

Razican:
> Edit: The lexer probably needs a refactor too.

Yes, I'm working a bit on that in a couple of places, with string interning and trying to parse source code byte by byte.

Razican:
Rebased and ready for a merge, @HalidOdat, @jasonwilliams :D

Razican:
Rebased and ready :)

boa/src/syntax/parser/mod.rs
```diff
@@ -1,1748 +1,137 @@
...
 
-    /// <https://tc39.es/ecma262/#prod-NewExpression>
-    fn read_new_expression(&mut self, first_member_expr: Node) -> ParseResult {
-        Ok(first_member_expr)
+impl From<bool> for AllowReturn {
+    fn from(allow: bool) -> Self {
+        Self(allow)
     }
+}
+/// Boolean representing if the parser should allow a `default` keyword.
```
jasonwilliams, line 81:
```suggestion

```

boa/src/syntax/parser/expression/assignment/arrow_function.rs
```diff
@@ -0,0 +1,161 @@
...
+    ) -> Self
+    where
+        I: Into<AllowIn>,
+        Y: Into<AllowYield>,
+        A: Into<AllowAwait>,
+    {
+        Self {
+            allow_in: allow_in.into(),
+            allow_yield: allow_yield.into(),
+            allow_await: allow_await.into(),
```
jasonwilliams, line 53: How come into() is needed here? (just for me to understand)

Razican: It's not realy needed, but I used generics in the function signature: `I: Into<AllowIn>` and so on. This allows us to create a parser with booleans too: `ArrowFunction(true, false, true)`.

We could just remove the generics from the `new()` functions and pass the wrapping structures, which would improve compile times, and is probably just as clear / developer friendly.

Issue #302, opened 2020-04-08
Implement do-while loop


HalidOdat:
Do-while loops are very similar to regular while loops, so it shouldn't be too hard to implement in `parser` and `exec`.

https://github.com/jasonwilliams/boa/blob/master/CONTRIBUTING.md

ptasz3k:
Since i'm trying to learn rust yet another time i will gladly take this issue.

PR #306, opened 2020-04-12
Implement do..while loop

ptasz3k:
Quick implementation. Any comments are welcome, as i don't know rust almost at all.

Solves #302


boa/src/syntax/parser/mod.rs
```diff
@@ -633,6 +634,33 @@ impl<'a> Parser<'a> {
...
 
+    /// https://tc39.es/ecma262/#sec-do-while-statement
+    fn read_do_while_statement(&mut self) -> ParseResult {
+        let body = self.read_statement()?;
+
+        let next_token = self
+            .peek_skip_lineterminator()
+            .ok_or(ParseError::AbruptEnd)?;
+
+        if next_token.kind != TokenKind::Keyword(Keyword::While) {
```
Razican, line 645: I think this could be changed for a `self.expect()`. This method does what you did here, but without having to repeat all the code :)

boa/src/exec/mod.rs
```diff
@@ -165,6 +165,13 @@ impl Executor for Interpreter {
                 }
                 Ok(result)
             }
+            Node::DoWhileLoop(ref expr, ref cond) => {
+                let mut result = self.run(expr)?;
+                while self.run(cond)?.borrow().is_true() {
+                    result = self.run(expr)?;
+                }
+                Ok(result)
+            }
```
HalidOdat, line 174: Maybe consider changing `expr` to `body`. Just so it's easier to read the code.
```suggestion
            Node::DoWhileLoop(ref body, ref cond) => {
                let mut result = self.run(body)?;
                while self.run(cond)?.borrow().is_true() {
                    result = self.run(body)?;
                }
                Ok(result)
            }
```

boa/src/syntax/parser/mod.rs
ptasz3k: @Razican Hey, thanks for pointing me to `expect()`. It simplifies things a lot, but to make that solution work i had to change `expect()` too to make it skip newlines. It didn't break any test, but i don't know how good the coverage is. I will update pr in a moment.

Razican: Hmmm it's better not to change `expect()` as it's used in many places. One option would be to first skip the new lines and then call `expect()`, since `expect() will check just the next token.

To skip the new lines, you can write:
```rust
self.cursor.skip(|tk| tk.kind == TokenKind::Line Terminator);
```

I might have the syntax a bit wrong, but it will point you to the right direction.

ptasz3k: Ok, the situation is a bit more complicated :). 

* we have two functions that get token after skipping tokens that fulfill predicate: `self.next_skip_lineterminator()` (that consumes token after line terminators as well, so next token after calling it is the one _after_ `while`) and `self.peek_skip_lineterminator()` (does not move `self.cursor.pos` at all, so if there are line terminators next token will be the first of them).
* `self.expect()` checks next token, so if we use `self.next_skip_lineterminator()` it won't get `while` at all (as it was already consumed), and if we use `self.peek_skip_lineterminator()` and there are any newlines before while it will fail, as next token is newline, not `while`.

I see three solutions, if i should use `expect`:
* make `expect()` skip newlines (previous one, may break existing code),
* creating new function `self.peek_next_token_skip_lineterminator()` that sets `cursor.pos` on last line terminator, so the next token is the one after it, and calling it before `expect()`
* new function `expect_skip_lineterminator()` (as in first solution, but leaving current `expect()` intact),

or i can stick to the very first one (which was shamelessly ripped off of `try..catch.finally` parser :)).

Razican: Oh, true! I created the function I mentioned for a PR that is not in yet. You can use the first version you had, and we can update it later :)

ptasz3k: Done! :)

boa/src/syntax/parser/tests.rs
```diff
@@ -725,4 +725,21 @@ fn check_do_while() {
             Box::new(Node::Const(Const::Bool(true))),
         )],
     );
+
```
Razican, line 728:
```suggestion
}

```

PR #311, opened 2020-04-13
Fixed assign operator precedence

### LINKED ISSUE #308: Assignment operator not working in tests (a = a +1)
While implementing `do..while` loop i've noticed strange boa behavior. This test

```
#[test]
fn test_while_loop() {
    let src = r#"
        let a = 0;
        while (a < 10) a = a + 1;
        a
    "#;
    assert_eq!(exec(src), String::from("10"));
}
```
will run forever. Changing loop body to `a += 1` makes test pass. Seems like using `a = a + 1` does not assign rvalue to lvalue. This test fails too:

```
#[test]
fn test_addition() {
    let src = r#"
        let a = 0;
        a = a + 1;
        a
    "#;
    assert_eq!(exec(src), String::from("1"));
}
```

I haven't looked at it yet, as i have limited debug possibilities now (vs code in wsl remote mode does not support lldb and i can't build boa on native windows cause of some problem with jemalloc compilation). Waiting for wsl2.

And second thought: maybe it would be interesting to implement some timeout in tests, i found something looking good here: https://users.rust-lang.org/t/limit-execution-time-of-test/6687/3

HalidOdat:
This PR fixes #308.
The bug was caused because we were parsing assign operator in every expression in `expression` macro. Which broke the assign operator precedence.

I also added tests for it, in `parser` and `exec`.


HalidOdat:
Resolved the merge conflicts. I think it is ready to merge. :)

PR #313, opened 2020-04-13
adding test for #273

jasonwilliams:
Test for constructor precedence in the parser.
It can be quite verbose to read, i'm not sure if there's a better way of testing the AST output


Razican:
> It can be quite verbose to read, i'm not sure if there's a better way of testing the AST output

Maybe we could add a macro to parser tests to generate AST easier. I can try that in a new PR.

PR #315, opened 2020-04-13
console functions

ptasz3k:
Ok, so here is `console.assert()`. According to [this spec](https://developer.mozilla.org/en-US/docs/Web/API/Console) it should print stack trace also. I haven't found any way to get it and its implementation is way above my skills for now. Maybe later :).

I would love to get some feedback on the code quality. If everything is acceptable I'll implement another functions.

There will be a need to keep `Console` object state for interpreter instance (map for `count()`, indent level and names for `group()` etc.). Where is the best place for it? `Interpreter`?


boa/src/builtins/console.rs
```diff
@@ -32,11 +32,34 @@ pub fn error(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
     Ok(Gc::new(ValueData::Undefined))
 }
 
+/// Print a javascript value to the standard error only if first argument evaluates to false or
+/// there were no arguments
```
Razican, line 36: I would do as Rust suggest, to include a small explanation of one line at the beginning and then maybe some extra information in the 3rd line of comment, so that it shows nicer when generating the documentation.

I would add the link to the spec: https://console.spec.whatwg.org/#assert

boa/src/exec/tests.rs
```diff
@@ -247,3 +247,23 @@ fn test_short_circuit_evaluation() {
...
     assert_eq!(exec(short_circuit_eval), String::from("1"));
 }
+
+#[test]
+fn test_console_assert() {
+    let multi_args = r#"
+        console.assert(12 > 13, "we should see that!", "that one too!", "that is not normal :/");
+    "#;
+    exec(multi_args);
+
```
Razican, line 257: I would add some confirmation on what the output is.

boa/src/exec/tests.rs
```diff
@@ -247,3 +247,23 @@ fn test_short_circuit_evaluation() {
...
+    exec(condition_only);
+
+    let false_string =
+        r#"console.assert("false", "if you can see that, assert's not working properly");"#;
+    exec(false_string);
+
+    let no_arguments = "console.assert();";
+    exec(no_arguments);
+
+    assert_eq!(true, true);
```
Razican, line 268: Why is this?

boa/src/exec/tests.rs
ptasz3k: Is there a way to check side effects of function in unit tests in rust? It would be cool, but i don't know how to do it.

boa/src/exec/tests.rs
ptasz3k: I'm debugging from the unit tests, starting them with `target/debug/boa-[hash].exe console_assert --nocapture` and setting breakpoints in interesting places.

As I don't know any way to test side effects of the function from test i end it with always true assert to make it look nice ;). Those tests are written only for the sake of the debugging, i'll remove them eventually.

boa/src/builtins/console.rs
ptasz3k: Ok, I'll do it.

boa/src/builtins/console.rs
```diff
@@ -32,11 +32,34 @@ pub fn error(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
...
+        let msg = if raw_args.len() <= 1 {
+            String::new()
+        } else {
+            let msgs: Vec<String> = FromIterator::from_iter(raw_args[1..].iter().map(|x| {
+                from_value::<String>(x.clone()).expect("Could not convert value to string")
+            }));
+            msgs.join(" ")
+        };
+        eprintln!("Assertion failed: {}", msg);
+    }
```
HalidOdat, line 52: Eliminating the unnecessary allocations.
```suggestion
    let assertion = if !args.is_empty() {
        from_value::<bool>(args[0].clone()).expect("Could not convert to bool.")
    } else {
        false
    };

    if !assertion {
        eprint!("Assertion failed: ");
        for (i, item) in args.iter().skip(1).enumerate() {
            eprint!("{}", item);
            if i + 1 != args.len() {
                eprint!(" ");
            }
        }
        eprintln!();
    }
```

boa/src/exec/tests.rs
Razican: Hmmm you mean the output to stdout/stderr? Not directly. As you say, this should output there, and that cannot be checked. In the future, we might have a Logger/Printer class that would be created with a writer for errors and so on.

boa/src/builtins/console.rs
ptasz3k: Of course, thank You! It's much cleaner code!

boa/src/builtins/console.rs
```diff
@@ -32,11 +32,36 @@ pub fn error(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
     Ok(Gc::new(ValueData::Undefined))
 }
 
+/// Print a javascript value to the standard error if first argument evaluates to false or there were no arguments
+/// 
+/// https://console.spec.whatwg.org/#assert
```
Razican, line 37: Actually, I meant something like this:
```suggestion
/// `console.assert(condition, ...data)`
///
/// Prints a JavaScript value to the standard error if first argument evaluates to `false` or there
/// were no arguments.
/// 
/// More information: <https://console.spec.whatwg.org/#assert>
```

boa/src/builtins/console.rs
```diff
@@ -32,11 +32,36 @@ pub fn error(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
...
+pub fn assert(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let assertion = if !args.is_empty() {
+        from_value::<bool>(args[0].clone()).expect("Could not convert to bool.")
+    } else {
+        false
+    };
```
Razican, line 43: We can improve this even further:
```suggestion
    let assertion = args.get(0).cloned().map(|val| from_value::<bool>(val).expect("Could not convert to bool.")).unwrap_or_default();
```

This of course needs a `cargo fmt`. Still, I find it a bit legacy that we use the `from_value()` function, instead of the `TryFrom` and `From` traits from the standard library. I will try to do a PR in the future to improve this. Ideally, we should also allow doing this without cloning, to avoid some allocations.

boa/src/builtins/console.rs
```diff
@@ -32,11 +32,36 @@ pub fn error(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
...
+
+    if !assertion {
+        eprint!("Assertion failed: ");
+        for (i, item) in args.iter().skip(1).enumerate() {
+            eprint!("{}", item);
+            if i + 1 != args.len() {
+                eprint!(" ");
+            }
+        }
```
Razican, line 52: I think we can do this even easier to understand :)
```suggestion
        eprint!("Assertion failed:");
        for message in args.iter().skip(1) {
            eprint!(" {}", message);
        }
```

boa/src/builtins/console.rs
ptasz3k: Ok, I will write docs like that in the future.

One more question: should I make one pr per implemented function or wait until everything will be done?

boa/src/builtins/console.rs
ptasz3k: Looks a bit cryptic to me, but surely it's because of my lack of knowledge. I will read about it.

boa/src/builtins/console.rs
ptasz3k: :)

boa/src/builtins/console.rs
Razican: > Looks a bit cryptic to me, but surely it's because of my lack of knowledge. I will read about it.

What it does is instead of checking if we have an argument value first, it just tries to get it. Rust already checks bounds on slice access, and returns an `Option`.

We need the cloned value in case it has any, so that's why we call `cloned()`.

Then, we need to convert it to a `bool`, so we call `map()` with a closure that does the conversion.

Finally, since up to now we have only modified the value if it was `Option::Some(_)`, we must tell it what to do in the case it was `None`. So we call `unwrap_or_default()`. This will return the `bool` value of the `map()` function if it's `Some`, and use the `bool::default()` if it's `None`. The default value for a boolean in Rust is `false`.

ptasz3k: Thanks for explanation! So out-of-bound access to slice with `get` will return `None` that will be then passed all the way to `unwrap_or_default`, which will return `false`. Clever!

ptasz3k:
`console.count()` and `console.countReset()` implemented as private field in console object. According to spec this should be `Map` rather, but since it isn't implemented yet i decided to use `Object`.

I have some doubts if i should not emulate it rather with native HashMap<string, i32>, it would be better performance and memory wise.

boa/src/builtins/console.rs
```diff
@@ -56,12 +56,63 @@ pub fn assert(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
...
 /// Create a new `console` object
 pub fn create_constructor(global: &Value) -> Value {
     let console = ValueData::new_obj(Some(global));
     console.set_field_slice("log", to_value(log as NativeFunctionData));
     console.set_field_slice("error", to_value(error as NativeFunctionData));
     console.set_field_slice("exception", to_value(error as NativeFunctionData));
     console.set_field_slice("assert", to_value(assert as NativeFunctionData));
+    console.set_field_slice("count", to_value(count as NativeFunctionData));
+    console.set_field_slice("countReset", to_value(count_reset as NativeFunctionData));
+    console.set_internal_slot("countMap", to_value(ValueData::new_obj(Some(global))));
```
ptasz3k, line 116: I'm not sure if i should pass Some(global) here.

Razican:
Hmmm maybe @jasonwilliams has some insights on this.

boa/src/builtins/console.rs
jasonwilliams: You only need to if you want the object to have its prototype set to `Object.prototype` im not sure if in this case you do, ill need to see how countMap is used.

ptasz3k:
I've looked how it's done in v8 and they're implementing state for all console functions as native c++ datatypes (ie. https://github.com/v8/v8/blob/master/src/inspector/v8-console-message.h#L137). I'll go that way too.

I'll define a struct for all the state that console object needs and put it temporarily in Interpreter. It should eventually end up in console/logger abstract context that @Razican mentioned before.

Is it ok?

boa/src/builtins/console.rs
ptasz3k: I think that we shouldn't use js object for that job at all, so it really doesn't matter :).

Razican:
> Is it ok?

Ok, let's try it and see how it looks like. Maybe we can think about changes later :)

jasonwilliams:
i believe we have https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/object/internal_state.rs for that sort of thing

ptasz3k:
Thanks for pointing me to the right direction, i think it is starting to take some shape slowly.

ptasz3k:
I've implemented formatter and console.group functions. To make it work properly we should make sure that **all** console printing is done thru console object to keep group indenation. I'm not sure also if keeping state as internal in console javascript object is ok, since we'll need to get info about group indenation on each print. Effectively it'll be group stack item count multiplied by indent level. It would be much easier if we would keep the console state explicitly somewhere in `Interpreter`.

Formatter is implemented in iterative way, as opposed to recursive  one that the spec says about and returns ready, prepared string instead of `a list of objects suitable for printing`, but it gives correct results. Hope it's ok. `%c` format is not implemented, and in case of '%o' and '%O' i don't know what format is correct.

I need some hints about the way to implement the Printer (console abstraction where all boa console output should go). Or maybe it is not necessary for now? If so, i'll implement the rest of console functions, write some tests for formatter to make sure it works correctly with utf-8 strings and the code will be ready to pull.

ptasz3k:
@Razican @HalidOdat  All functions without timestamp and profile* are implemented. I think it's ready to be reviewed and pulled in.

boa/src/builtins/value/mod.rs
```diff
@@ -767,7 +767,7 @@ pub(crate) fn log_string_from(x: &ValueData, print_internals: bool) -> String {
 }
 
 /// A helper function for specifically printing object values
-fn display_obj(v: &ValueData, print_internals: bool) -> String {
+pub fn display_obj(v: &ValueData, print_internals: bool) -> String {
```
Razican, line 770: I would restrict this to the crate:
```suggestion
pub(crate) fn display_obj(v: &ValueData, print_internals: bool) -> String {
```

boa/src/builtins/console/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+pub fn logger(msg: LogMessage, console_state: &ConsoleState) {
+    let indent = 2 * console_state.groups.len();
+
+    match msg {
+        LogMessage::Error(msg) => {
+            eprint!("{}", " ".repeat(indent));
+            eprintln!("{}", msg);
+        }
```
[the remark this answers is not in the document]
ptasz3k: Of course, thanks!

PR #317, opened 2020-04-13
Profiler using measureme

jasonwilliams:
fixes https://github.com/jasonwilliams/boa/issues/296


boa/src/lib.rs
```diff
@@ -77,13 +79,20 @@ pub fn forward(engine: &mut Interpreter, src: &str) -> String {
 /// If the interpreter fails parsing an error value is returned instead (error object)
 pub fn forward_val(engine: &mut Interpreter, src: &str) -> ResultValue {
     // Setup executor
-    match parser_expr(src) {
+    MyProfiler::init();
+    let _ = MyProfiler::global().start_event("Main");
```
wesleywiser, line 83:
```suggestion
    let main_timer = MyProfiler::global().start_event("Main");
```

boa/src/lib.rs
```diff
@@ -77,13 +79,20 @@ pub fn forward(engine: &mut Interpreter, src: &str) -> String {
...
+    let result = match parser_expr(src) {
         Ok(expr) => engine.run(&expr),
         Err(e) => {
             eprintln!("{}", e);
             std::process::exit(1);
         }
-    }
+    };
+
+    MyProfiler::global().drop();
```
wesleywiser, line 94:
```suggestion
    drop(main_timer);
    MyProfiler::global().drop();
```

jasonwilliams:
@Razican @HalidOdat There's some great progress happening with this.
image

Razican:
> @Razican @HalidOdat There's some great progress happening with this.
> image

I'm loving this. This should give nice insights on where do we spend our time.

abhijeetbhagat:
Is that flamegraph displayed in a Chrome devtools window?

Razican:
> Is that flamegraph displayed in a Chrome devtools window?

Yep! :)

abhijeetbhagat:
Ok, an SVG file loaded in devtools?

Razican:
> Ok, an SVG file loaded in devtools?

Yes, the implementation is based in this: https://blog.rust-lang.org/inside-rust/2020/02/25/intro-rustc-self-profile.html

jasonwilliams:
@Razican You asked about the realm profile image
image

Razican:
> @Razican You asked about the realm profile image
> image
> 
> You should be able to load this into Chrome
> https://gist.github.com/jasonwilliams/b1caeef708b36205f279e7342fbfbcf9

This looks like a clear improvement path. Could we also have internal metrics for each of the intrinsics creation?

jasonwilliams:
> > @Razican You asked about the realm profile image
> > image
> > You should be able to load this into Chrome
> > https://gist.github.com/jasonwilliams/b1caeef708b36205f279e7342fbfbcf9
> 
> This looks like a clear improvement path. Could we also have internal metrics for each of the intrinsics creation?

Yeah we can go as deep as we want really, its up to us what we want to trace

Razican:
> Yeah we can go as deep as we want really, its up to us what we want to trace

In this case, I would add smaller steps in things that take too long, like this one. We can then review what we could do to improve this :)

jasonwilliams:
image

latest gist shows each function being created
https://gist.github.com/jasonwilliams/b20b9b89fa965e2a3fe7a58206aba257

jasonwilliams:
Over 60% of time is spent in `set_field`
image

HalidOdat:
> 60% of time is spent in `set_field` during execution
> image

I thought as much, I tested with removing `.set_field` and directly adding to the objects properties and it speed things significantly

jasonwilliams:
> I thought as much, I tested with removing .set_field and directly adding to the objects properties and it speed things significantly

That's interesting, do we have a clean way of refactoring this?

HalidOdat:
> > I thought as much, I tested with removing `.set_field` and directly adding to the objects properties and it speed things significantly
> 
> That's interesting, do we have a clean way of refactoring this?

The method that I used was only for testing, it not clean at all. I just extracted `properties` field for the `Object` and called `.insert()`.

I think I'll try to do something more permanent in #419

boa/src/lib.rs
```diff
@@ -71,15 +74,23 @@ pub fn forward(engine: &mut Interpreter, src: &str) -> String {
 /// The str is consumed and the state of the Interpreter is changed
 /// Similar to `forward`, except the current value is returned instad of the string
 /// If the interpreter fails parsing an error value is returned instead (error object)
+#[allow(clippy::unit_arg, clippy::drop_copy)]
```
Razican, line 77: Just for my understanding, why is this needed here? Is the timer `Copy` or the profiler?

boa/src/profiler.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    pub fn default() -> BoaProfiler {
+        let profiler = Profiler::new(Path::new("./my_trace")).unwrap();
+        BoaProfiler { profiler }
+    }
+
+    // init creates a global instance of BoaProfiler which can be used across the whole application
+    pub fn init() {
+        let profiler = Self::default();
+        unsafe {
```
Razican, line 48: I think we can remove this u safety by using `get_or_init()`, I will give this a look tomorrow :)

docs/profiling.md
```diff
@@ -0,0 +1,27 @@
...
+
+![Example](img/profiler.png)
+
+It's possible to get a full profile of Boa in action.  
+Sometimes this is needed to figure out where it is spending most of it's time.
+
+We use a crate called [measureme](https://github.com/rust-lang/measureme), which helps us keep track of timing functions during runtime.
+
+## How To Use
+
```
HalidOdat, line 11: We should add fill in the how to use section

boa/src/profiler.rs
```diff
@@ -0,0 +1,103 @@
...
+#[cfg(windows)]
+#[cfg(feature = "profiler")]
+type SerializationSink = measureme::FileSerializationSink;
+#[cfg(feature = "profiler")]
+pub struct BoaProfiler {
+    profiler: Profiler<SerializationSink>,
+}
+
+#[cfg(feature = "profiler")]
+pub static mut INSTANCE: OnceCell<BoaProfiler> = OnceCell::new();
```
brian-gavin, line 28: Does this need to be `pub`, considering that it is accessed safely via `BoaProfiler::global()`?

jasonwilliams: Thanks @brian-gavin thats correct, changed now

docs/profiling.md
jasonwilliams: The how-to is under the prerequesites, ive reodered now

boa/src/lib.rs
jasonwilliams:
```
note: argument has type ()
  --> boa\src\lib.rs:90:10
   |
90 |     drop(main_timer);
```

So it looks like `()` implements Copy and when we don't compile the profiler, we're just dropping `()` which triggers this linting warning.

boa/src/profiler.rs
brian-gavin: Thanks!

boa/src/lib.rs
Razican: This could be solved by only calling `drop()` with the feature active. We can add an artificial scope here and add the `#[cfg]` flag.

boa/src/profiler.rs
```diff
@@ -0,0 +1,93 @@
...
+        let id = EventId::from_label(self.profiler.alloc_string(label));
+        let thread_id = Self::thread_id_to_u32(current().id());
+        self.profiler
+            .start_recording_interval_event(kind, id, thread_id)
+    }
+
+    pub fn default() -> BoaProfiler {
+        let profiler = Profiler::new(Path::new("./my_trace")).unwrap();
+        BoaProfiler { profiler }
+    }
```
Razican, line 43: We should implement the `Default` trait, and any function returning `BoaProfiler` in `BoaProfiler` should return `Self`, in case we change its name in the future, to make it easier.

boa/src/profiler.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn drop(&self) {
+        // In order to drop the INSTANCE we need to get ownership of it, which isn't possible on a static unless you make it a mutable static
+        // mutating statics is unsafe, so we need to wrap it as so.
+        // This is actually safe though because init and drop are only called at the beginning and end of the application
+        unsafe {
+            INSTANCE
+                .take()
+                .expect("Could not take back profiler instance");
+        }
+    }
```
Razican, line 58: I'm afraid this is not safe. If we call `forward()` more than once (which we sometimes do), this would be undefined behaviour.

I think the best solution is not to use a `static` global here. We could use a `Mutex` to avoid this, but I think this would reduce performance a lot. I think the best option is to conditionally have a `BoaProfiler` in the `Lexer`, `Parser` and `Interpreter`.

In the future, we could have a global `Boa` object with this field. In any case, I think the three of them should have a `start_event()` that would be no-op if the feature is not active. And all event starts should call to the `Lexer`, `Parser` or `Interpreter`.

PR #318, opened 2020-04-14
Added continuous integration for windows

HalidOdat:
I added continuous integration for windows.
 - Rename `Check` **to** `Check on Linux`
 - Added `Check on Windows`
 - Rename `Test Suite` **to** `Test Suite on Linux`
 - Added `Test Suite on Windows`

I did not add `Rust Fmt on Windows` and `Rust Clippy on Windows`, since there are not os specific.

what do you think? @Razican @jasonwilliams 

Also should I add for benchmark too?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,8 +9,8 @@ on:
 name: Continuous integration
 
 jobs:
-  check:
-    name: Check
+  check_on_linux:
+    name: Check on Linux
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


Razican:
Could be interesting to add benchmarks, yes. And I think we should have at least one benchmark of both lexing **and** parsing, that could be checked in the future when we merge both. We can take this opportunity to do that.

I would also add a benchmark of a bit more complex program too than a `hello world`.

Razican:
Oh, and I think we should upgrade the `checkout` action to `v2`. API is the same, but it's faster.

jasonwilliams:
> Oh, and I think we should upgrade the `checkout` action to `v2`. API is the same, but it's faster.

I actually did this but it caused problems so I had to revert.
Basically v2 only checks out the branch the action is running on, so when benchmarks switch to master it fails, v1 was more like “git pull”.

I’m sure this is fixable, but if you hit the same issue that seems to be what it is.
I’m guessing with v2 you need to pass some flag to say you want the other branches too or master included.

HalidOdat:
I upgraded CI to checkout@v2 and it seems to be working fine.

I think this is ready to merge. :)

Razican:
I think it would be interesting to benchmark on Windows too, since for example Jemallocator is Linux-only.

HalidOdat:
> I think it would be interesting to benchmark on Windows too, since for example Jemallocator is Linux-only.

I think we should do this in another PR. :)

BTW I'm not very knowledgeable of how the benchmarks are set or github actions for that matter.

PR #320, opened 2020-04-15
implement "this"

### LINKED ISSUE #264: Implement the "this" keyword
The `this` [keyword][this_keyword] returns an expression dependent on the context where it's being used. This context is resolved by the [`ResolveThisBinding()`][resolve_this] abstract operation. This abstract operation will get the current environment record, and return the `this` binding, if any.

The MDN [documentation][mdn_docs] also explains how this keyword will resolve to a different expression depending on how it's called.

To implement this, we require changes to the parser, where keywords are parsed ([parser.rs#L112][struct_parsing]), which will probably require a new `ExprDef`. Executing that expression in the executor, it should return the object representing the `this` binding.

[this_keyword]: https://tc39.es/ecma262/#sec-this-keyword
[resolve_this]: https://tc39.es/ecma262/#sec-resolvethisbinding
[mdn_docs]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this
[struct_parsing]: https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/parser.rs#L112

jasonwilliams:
Fixes https://github.com/jasonwilliams/boa/issues/264
Fixes https://github.com/jasonwilliams/boa/issues/360

### DIFF
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -18,7 +18,7 @@ use crate::{
     realm::Realm,
     syntax::ast::{
         constant::Const,
-        node::{MethodDefinitionKind, Node, PropertyDefinition},
+        node::{FormalParameter, MethodDefinitionKind, Node, PropertyDefinition},
         op::{AssignOp, BinOp, BitOp, CompOp, LogOp, NumOp, UnaryOp},
     },
 };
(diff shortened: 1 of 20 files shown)


Razican:
Hmmm I will give this a look during the weekend and see if I find out how it can be done :)

Razican:
This should no longer be blocked on function objects.

About the environment trait, for now, I think this is good, but we might want to have multiple traits, depending on what should each environment have. For example, a global environment shouldn't have a parent environment, so it doesn't need to implement that function (and we avoid the `Option` there). But I think this is out of the scope of this PR.

jasonwilliams:
This is ready for review, @Razican clippy is moaning about https://github.com/jasonwilliams/boa/pull/320/files#diff-d7bd296d9ca534161e198b8c141ca51fR611 im not sure what to do about it.

Razican:
> This is ready for review, @Razican clippy is moaning about https://github.com/jasonwilliams/boa/pull/320/files#diff-d7bd296d9ca534161e198b8c141ca51fR611 im not sure what to do about it.

Let me check this tomorrow or on Friday :)

boa/src/exec/mod.rs
```diff
@@ -670,6 +605,46 @@ impl Interpreter {
         &self.realm
     }
 
+    /// Utility to create a function Value for Function Declarations, Arrow Functions or Function Expressions
+    pub(crate) fn create_function(
+        &mut self,
+        args: &Box<[FormalParameter]>,
```
Razican, line 611: So, we have two options here. Either we receive a `Box<[FormalParameter]>` or a `&[FormalParameter]`. If you take a `Box`, you can directly pass it to `FunctionObject::create_ordinary` and this would probably reflect better how the function works internally, but you would need to clone it on call.

If this is not possible, you can just receive a `&[FormalParameter]` and then do a `to_vec().into_boxed_slice()`, or something like that.

jasonwilliams: Thanks, will try both options later

PR #323, opened 2020-04-16
Added more benchmarks

### LINKED ISSUE #322: Benchmarks are failing in master
Since the v0.7.0 release, all benchmarks are failing in master ([link][benches]). The error it gives is that the `Cargo.lock` file would be overwritten by the clone. This seems related to the version change in `Cargo.toml`, if we didn't do a `cargo update` just after.

So I think the solution is to just do a `cargo update` and commit.

[benches]: https://github.com/jasonwilliams/boa/actions?query=workflow%3A%22Boa+Benchmarks+on+master%22

Razican:
We now have more benchmarks, and parsing benchmarks also take into
account lexing, since in the future, both will be done together.

I also changed a bit the structure of the benchmark files, so now it should be easier to understand. We have 3 files: `lexer.rs`, `parser.rs` and `exec.rs`, and each has its benchmarks.

I also upgraded and updated dependencies, so this should also fix #322.


Razican:
I think this is ready, but this will require one PR to the `gh-pages` branch, to retroactively update the benchmark names, since I changed the names of the benchmarks to make them easier to understand and make them more standardized.

PR #324, opened 2020-04-16
Renamed old benchmarks

Razican:
This updates old benchmarks so that they show in the same graphs in the benchmarks page. Note that this will show a huge worsening in parsing, since we now benchmark lexing and parsing together (they will work at the same time _soon_).

### DIFF
--- a/dev/bench/data.js
+++ b/dev/bench/data.js
@@ -32,7 +32,7 @@ window.BENCHMARK_DATA = {
             "unit": "us"
           },
           {
-            "name": "Symbol creation (Execution)",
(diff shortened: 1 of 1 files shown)


Razican:
I just noticed that some benchmarks don't have a name, somehow. I'm checking that.

Razican:
I finally found the issue. It turns out that the GitHub action expect for the name of the benchmark and the result to be in only one line, but for long names, Criterion divides this in two. I will change the name so that it properly shows in one line, but we might want to improve this by using the Criterion data in JSON in the future, in the action.

PR #325, opened 2020-04-16
Put JSON functions on the object, not the prototype

coolreader18:


Razican:
Hi @HalidOdat is this a duplicate? Could we link the duplicate issue?

HalidOdat:
> Hi @HalidOdat is this a duplicate? Could we link the duplicate issue?

This PR solves the same issue but it needs a rebase #270

boa/src/builtins/json.rs
```diff
@@ -23,20 +22,39 @@ pub fn parse(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
 /// Process a Javascript object into a JSON string
 pub fn stringify(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
     let obj = args.get(0).expect("cannot get argument for JSON.stringify");
-    let json = obj.to_json();
-    Ok(to_value(to_string_pretty(&json).expect("")))
+    let json: String = obj.to_json().to_string();
```
Razican, line 25: Doesn't type inference work here?
```suggestion
    let json = obj.to_json().to_string();
```

boa/src/builtins/json.rs
```diff
@@ -23,20 +22,39 @@ pub fn parse(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
...
-    json_value.set_field_slice(PROTOTYPE, prototype);
-    json_value
+    to_value(json)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::exec::Executor;
+    use crate::forward;
+    use crate::realm::Realm;
```
Razican, line 43: I think it's more clear if we use this syntax for imports. It makes it clear all we import from the crate:
```suggestion
    use crate::{
    	exec::Executor,
    	forward,
    	realm::Realm,
    };

```

boa/src/builtins/json.rs
```diff
@@ -1,10 +1,9 @@
 use crate::builtins::function::NativeFunctionData;
-use crate::builtins::object::{Object, ObjectKind, PROTOTYPE};
 /// The JSON Object
 /// <https://tc39.es/ecma262/#sec-json-object>
```
Razican, line 3: Why do we have comments in an import? I think these should be module level documentation, and all imports should be together.

PR #328, opened 2020-04-17
fix #209 "Calling Array with one argument"

### LINKED ISSUE #209: Calling `new Array` with 1 argument doesn't work properly

HalidOdat:
This is supposed to fix #209.

While trying to fix this bug I encountered another bug were we only get `Const::Num` and never `Const::Int`

This PR is blocked until it is fixed.

I will create another PR addressing that bug.

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -118,7 +118,7 @@ pub fn make_array(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result
     match args.len() {
         1 if args[0].is_integer() => {
             length = from_value::<i32>(args[0].clone()).expect("Could not convert argument to i32");
-            // TODO: It should not create an array of undefinds, but an empty array ("holy" array in V8) with length `n`.
+            // TODO: It should not create an array of undefineds, but an empty array ("holy" array in V8) with length `n`.
             for n in 0..length {
                 this.set_field_slice(&n.to_string(), Gc::new(ValueData::Undefined));
             }
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -101,15 +101,6 @@ pub fn make_array(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result
     // Make a new Object which will internally represent the Array (mapping
     // between indices and values): this creates an Object with no prototype
 
-    // Create length
-    let length = Property::new()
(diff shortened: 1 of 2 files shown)


Razican:
> While trying to fix this bug I encountered another bug were we only get `Const::Num` and never `Const::Int`
> 
> This PR is blocked until it is fixed.

This was mentioned in #270: https://github.com/jasonwilliams/boa/pull/270/files#r401575688

HalidOdat:
> > While trying to fix this bug I encountered another bug were we only get `Const::Num` and never `Const::Int`
> > This PR is blocked until it is fixed.
> 
> This was mentioned in #270: [#270 (files)](https://github.com/jasonwilliams/boa/pull/270/files#r401575688)

It seems that we never output a token of type `i32` the only numerical value is of floating type `f64`.

I think v8 has a `i32` token type `smi` (small intiger) right?

Razican:
> I will change `Token` to have different numerical types.

Nice. I see that they also have a "big integer" is this part of the spec in some way?

HalidOdat:
> > I will change `Token` to have different numerical types.
> 
> Nice. I see that they also have a "big integer" is this part of the spec in some way?

Yes. I checked the specs for [`NumericLiteral`]([`NumericLiteral`](https://tc39.es/ecma262/#prod-NonDecimalIntegerLiteral)) and BigInt is one of them.

boa/src/builtins/array/mod.rs
```diff
@@ -122,11 +113,36 @@ pub fn make_array(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result
...
+        1 if args[0].is_integer() => {
+            length = from_value::<i32>(args[0].clone()).expect("Could not convert argument to i32");
+            // TODO: It should not create an array of undefinds, but an empty array ("holy" array in V8) with length `n`.
+            for n in 0..length {
+                this.set_field_slice(&n.to_string(), Gc::new(ValueData::Undefined));
+            }
+        }
+        1 if args[0].is_double() => {
+            // TODO: throw `RangeError`.
+            unimplemented!("RangeError: invalid array length");
```
Razican, line 128: I would just use `panic!()` for now, since it's not that the feature is unimplemented, it's just that we cannot throw errors.

Issue #329, opened 2020-04-17
After #326, we cannot generate documentation for the boa library


Razican:
So, it seems that now that the CLI binary is called `boa`, the documentation is being output in `target/doc/boa` for both the binary and the library, and rustdoc just ignores the library.

So we need to either revert that PR, change the name of the library crate (to `boa_core` or something similar), or find a way to avoid this through configuration options.

HalidOdat:
Hmmm. I will look into it to find a better solution. :)

HalidOdat:
We could: 
 - disable documentation on the binary with `doc = false`
 - revert the changes.

What do you think? @Razican

Razican:
Hmmm maybe not generating documentation for the binary makes sense. There is basically no documentation anyway, and the code is simple enough (at least for now).

Let's see what @jasonwilliams thinks.

jasonwilliams:
I agree with @Razican I’m ok with not generating documentation for the binary, that would be my choice right now.

PR #330, opened 2020-04-17
Improved CI workflows

Razican:
- Added caching for all workflows (except fmt, that doesn't need it)
 - Added development documentation upload.
 - Added documentation compilation, to make sure our docs compile.
 - We now use the "minimal" profile in the Rust toolchain, to improve performance.

This is blocked on #329.


Razican:
I think this is ready for review. This requires a new _secret_, the `secrets.USERNAME`, that should be the user related to the current `secrets.PERSONAL_GITHUB_TOKEN`.

I have configured it for now to just commit to a testing branch, since I'm not 100% sure that it will respect the history of the branch, even though what I've seen from the GitHub action shows that it will.

We can have it like this for a couple of commits in master (I will do an intermediate commit in the testing branch) and if everything is fine, we can switch the branch to use `gh-pages`.

Razican:
Let me know, @jasonwilliams, when the secrets are ready (or if you want any change).

PR #338, opened 2020-04-19
Fix #331 "We only get `Const::Num`, never `Const::Int`"

### LINKED ISSUE #331: We only get `Const::Num`, never `Const::Int`
This is a bug in the lexer.

Currently we only have one numerical token type `NumericalLiteral` which is of type `f64`, hence all numerical values are converted to `f64` and we only get `Cont::Num`.

This bug blocks #328

HalidOdat:
This will close #331 and unblock #328

### DIFF
--- a/boa/src/builtins/value/mod.rs
+++ b/boa/src/builtins/value/mod.rs
@@ -48,7 +48,7 @@ pub enum ValueData {
     /// `String` - A UTF-8 string, such as `"Hello, world"`
     String(String),
     /// `Number` - A 64-bit floating point number, such as `3.1415`
-    Number(f64),
+    Rational(f64),
     /// `Number` - A 32-bit integer, such as `42`
     Integer(i32),
     /// `Object` - An object, such as `Math`, represented by a binary tree of string keys to Javascript values
@@ -150,7 +150,21 @@ impl ValueData {
     /// Returns true if the value is a 64-bit floating-point number
     pub fn is_double(&self) -> bool {
         match *self {
-            Self::Number(_) => true,
+            Self::Rational(_) => true,
+            _ => false,
+        }
+    }
+
+    /// Returns true if the value is integer.
+    #[allow(clippy::float_cmp)]
+    pub fn is_integer(&self) -> bool {
+        // If it can fit in a i32 and the trucated version is
+        // equal to the original then it is an integer.
+        let is_racional_intiger = |n: f64| n == ((n as i32) as f64);
+
+        match *self {
+            Self::Integer(_) => true,
+            Self::Rational(n) if is_racional_intiger(n) => true,
             _ => false,
         }
     }
@@ -183,7 +197,7 @@ impl ValueData {
         match *self {
             Self::Object(_) => true,
             Self::String(ref s) if !s.is_empty() => true,
-            Self::Number(n) if n != 0.0 && !n.is_nan() => true,
+            Self::Rational(n) if n != 0.0 && !n.is_nan() => true,
             Self::Integer(n) if n != 0 => true,
             Self::Boolean(v) => v,
             _ => false,
@@ -198,9 +212,9 @@ impl ValueData {
                 Ok(num) => num,
                 Err(_) => NAN,
             },
-            Self::Number(num) => num,
+            Self::Rational(num) => num,
             Self::Boolean(true) => 1.0,
-            Self::Boolean(false) | Self::Null => 0.0,
+            Self::Boolean(false) | ValueData::Null => 0.0,
             Self::Integer(num) => f64::from(num),
         }
     }
@@ -218,7 +232,7 @@ impl ValueData {
                 Ok(num) => num,
                 Err(_) => 0,
             },
-            Self::Number(num) => num as i32,
+            Self::Rational(num) => num as i32,
             Self::Boolean(true) => 1,
             Self::Integer(num) => num,
         }
@@ -556,7 +570,7 @@ impl ValueData {
     pub fn from_json(json: JSONValue) -> Self {
         match json {
             JSONValue::Number(v) => {
-                Self::Number(v.as_f64().expect("Could not convert value to f64"))
+                Self::Rational(v.as_f64().expect("Could not convert value to f64"))
             }
             JSONValue::String(v) => Self::String(v),
             JSONValue::Bool(v) => Self::Boolean(v),
@@ -607,7 +621,7 @@ impl ValueData {
                 JSONValue::Object(new_obj)
             }
             Self::String(ref str) => JSONValue::String(str.clone()),
-            Self::Number(num) => JSONValue::Number(
+            Self::Rational(num) => JSONValue::Number(
                 JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
             ),
             Self::Integer(val) => JSONValue::Number(JSONNumber::from(val)),
@@ -619,7 +633,7 @@ impl ValueData {
     /// https://tc39.es/ecma262/#sec-typeof-operator
     pub fn get_type(&self) -> &'static str {
         match *self {
-            Self::Number(_) | Self::Integer(_) => "number",
+            Self::Rational(_) | Self::Integer(_) => "number",
             Self::String(_) => "string",
             Self::Boolean(_) => "boolean",
             Self::Symbol(_) => "symbol",
@@ -637,7 +651,7 @@ impl ValueData {
     }
 
     pub fn as_num_to_power(&self, other: Self) -> Self {
-        Self::Number(self.to_num().powf(other.to_num()))
+        Self::Rational(self.to_num().powf(other.to_num()))
     }
 }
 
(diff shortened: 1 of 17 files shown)


jasonwilliams:
Thanks, will try and get round to looking at this soon

boa/src/syntax/lexer/mod.rs
```diff
@@ -399,16 +399,18 @@ impl<'a> Lexer<'a> {
                                     }
                                 }
                             }
+                            dbg!(&gone_decimal);
```
Razican, line 402: This should be removed before merging.

boa/src/syntax/lexer/mod.rs
```diff
@@ -399,16 +399,18 @@ impl<'a> Lexer<'a> {
...
                             if gone_decimal {
-                                f64::from_str(&buf).map_err(|_e| LexerError::new("Could not convert value to f64"))?
+                                NumericLiteral::Number(f64::from_str(&buf).map_err(|_e| LexerError::new("Could not convert value to f64"))?)
                             } else if buf.is_empty() {
-                                0.0
+                                NumericLiteral::Integer(0)
                             } else {
-                                (u64::from_str_radix(&buf, 8).map_err(|_e| LexerError::new("Could not convert value to u64"))?) as f64
+                                NumericLiteral::Integer(
+                                u64::from_str_radix(&buf, 8).map_err(|_e| {dbg!(&_e); LexerError::new("Could not convert value to u64")})?)
```
Razican, line 409: The `dbg!()` call should be removed before merging.
```suggestion
                                u64::from_str_radix(&buf, 8).map_err(|_e| LexerError::new("Could not convert value to u64"))?)
```

boa/src/syntax/lexer/mod.rs
```diff
@@ -469,9 +474,19 @@ impl<'a> Lexer<'a> {
...
                     self.push_token(TokenKind::NumericLiteral(
-                        f64::from_str(&buf).map_err(|_| LexerError::new("Could not convert value to f64"))?,
+                        if is_floating_point {
+                            // TODO make this a bit more safe -------------------------------VVVV
+                            NumericLiteral::Number(
+                                f64::from_str(&buf).map_err(|_| LexerError::new("Could not convert value to f64"))?
+                            )
+                        } else {
+                            NumericLiteral::Integer(
+                                u64::from_str(&buf).map_err(|_e| {dbg!(&_e); dbg!(&is_floating_point); LexerError::new("Could not convert value to u64")})?
```
Razican, line 486: The `dbg!()` macro should be removed before merging.
```suggestion
                                u64::from_str(&buf).map_err(|_e| LexerError::new("Could not convert value to u64"))?
```

boa/src/syntax/ast/token.rs
```diff
@@ -42,6 +42,17 @@ impl Debug for VecToken {
     }
 }
 
+/// Represents the type differenct types of numeric literals.
+#[cfg_attr(feature = "serde-ast", derive(Serialize, Deserialize))]
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum NumericLiteral {
+    /// A floating point number
+    Number(f64),
```
Razican, line 50: I would probably change this `Number` for `Real` or something like that, to differentiate from an `Integer`.

boa/src/syntax/ast/token.rs
```diff
@@ -42,6 +42,17 @@ impl Debug for VecToken {
...
+    /// A integer
+    Integer(u64),
+    // TODO: Add BigInt
```
Razican, line 53: How big is a `BigInt`?

HalidOdat: BigInt can hold any arbitrary large integer. All the numerical are implemented in the software level.

[num-bigint](https://crates.io/crates/num-bigint) crate

boa/src/syntax/ast/token.rs
HalidOdat: In V8 they have `Number` and `Smi`, so should I do it like V8 or `Real` and `Integer`?

Razican: > In V8 they have `Number` and `Smi`, so should I do it like V8 or `Real` and `Integer`?

I think it's more accurate if we call `Real` or even `Rational` to non-integers.

HalidOdat:
I reworked the number lexing and fixed some things:
 - Added some documentation to `lexer` and `token`
 - Fixed a bug were it did not accept `09.5e2`.
 - Made it easy to add `strict mode` for number literal (in strict mode `034` and `098` is an error).
 - Made it easy to add `BigInt` support (including other base BigInts).
 - Added `Value::is_integer` to check if a number is an integer.
 - Renamed `ValueData::Number(f64)` to `ValueData::Racional(f64)`
 - Fixed position bug related to numeric tokens.
 - Added some tests for edge case with implicit octal and hexadecimals followed by a dot
 - Split numerical lexing to a different function.
 - Removed `read_integer_in_base`.
 
This **unblocks** #328 

I think now it's ready to review. :)

boa/src/builtins/value/mod.rs
```diff
@@ -43,7 +43,7 @@ pub enum ValueData {
     /// `String` - A UTF-8 string, such as `"Hello, world"`
     String(String),
     /// `Number` - A 64-bit floating point number, such as `3.1415`
-    Number(f64),
+    Racional(f64),
```
Razican, line 46: Why is this `Racional` instead of `Rational`?

boa/src/syntax/lexer/tests.rs
```diff
@@ -336,7 +336,6 @@ fn test_two_divisions_in_expression() {
     let s = "    return a !== 0 || 1 / a === 1 / b;";
     let mut lexer = Lexer::new(s);
     lexer.lex().expect("failed to lex");
-    dbg!(&lexer.tokens);
```
Razican, line 339: This `dbg!` here was to be able to debug the token generation from the lexer when we activate this test (when we solve the goal-oriented lexer). We can remove it, but not sure if there is a need.

boa/src/syntax/ast/token.rs
```diff
@@ -42,28 +46,53 @@ impl Debug for VecToken {
     }
 }
 
+/// Represents the type differenct types of numeric literals.
+#[cfg_attr(feature = "serde-ast", derive(Serialize, Deserialize))]
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum NumericLiteral {
+    /// A floating point number
+    Racional(f64),
```
Razican, line 54: Again, isn't the correct word `Rational`?

boa/src/syntax/lexer/tests.rs
```diff
@@ -443,55 +541,88 @@ fn test_regex_literal_flags() {
...
-    assert_eq!(lexer.tokens[0].kind, TokenKind::NumericLiteral(1000.0));
+    assert_eq!(
+        lexer.tokens[0].kind,
+        TokenKind::NumericLiteral(NumericLiteral::Racional(1000.0))
+    );
     assert_eq!(lexer.tokens[1].kind, TokenKind::Punctuator(Punctuator::Add));
     assert_eq!(
         lexer.tokens[2].kind,
-        TokenKind::NumericLiteral(100_000_000_000.0)
+        TokenKind::NumericLiteral(NumericLiteral::Racional(100_000_000_000.0))
```
Razican, line 626: If we start seeing this a bit too wordy, we could implement `From<f64>` and `From<i32>` for `NumericLiteral`, and then add a `fn numeric_literal<L>(lit: L) -> Self where L: Into<NumericLiteral> {}` to `TokenKind`, which would directly allow us to pass a `f64` or a `i32`. I think I already did something similar in the modular parser PR.

boa/src/builtins/value/mod.rs
HalidOdat: > Why is this `Racional` instead of `Rational`?

Good catch! I wouldn't have noticed if I wasn't concentrated on it. It's because in my native language that's
how you spell `Rational` in my mind it didn't seem wrong at the time.

You're right it should be `Rational`. :)

boa/src/syntax/lexer/mod.rs
```diff
@@ -229,6 +223,225 @@ impl<'a> Lexer<'a> {
...
+                *self = match *self {
+                    Self::Rational => unreachable!("can not convert rational number to BigInt"),
+                    Self::Integer(base) => Self::BigInt(base),
+                    Self::BigInt(base) => Self::BigInt(base),
+                };
+            }
+        }
+
+        // TODO: Setup strict mode.
+        let strict_mode = false;
```
HalidOdat, line 263: Currently I have `strict_mode` set to always be `false`. In strict mode decimals with leading zeros and implicit octal are not allowed.

I seems like the parser should take in a lexer, because `'use strict';` it only applies to function or program scope and the lexer does not know what a scope is.

boa/src/builtins/value/mod.rs
Razican: It happens to me too. In Spanish it's spelled "racional" too haha

boa/src/syntax/lexer/mod.rs
Razican: Yep, this is going to be interesting to see once we have the lexer as an iterator tied to the cursor.

Razican:
I would wait for #304 to land before this.

HalidOdat:
I will rebase today!

Issue #292, opened 2020-03-31
MethodDefinition parsing


jasonwilliams:
We currently don't parse getters and setters on object literals.
There needs to be a new parsing section for MethodDefinitions.

We already have readObjectLiteral, so the logic would need to be called from within there.
https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/parser/mod.rs#L1599

This would either need to be rebased on https://github.com/jasonwilliams/boa/pull/281 or on master once #281 is in.

**Spec**:
https://tc39.es/ecma262/#prod-MethodDefinition

Previous Rust Code (you don't have to use this)
```rust

        TODO need to revisit this
        if let TokenKind::Identifier(name) = tok.kind {
            if name == "get" || name == "set" {
                let may_identifier = self.peek_skip_lineterminator();
                if may_identifier.is_ok()
                    && matches!(may_identifier.unwrap().kind, TokenKind::Identifier(_))
                {
                    let f = self.read_function_expression()?;
                    let func_name = if let NodeBase::FunctionExpr(ref name, _, _) = f.base {
                        name.clone().unwrap()
                    } else {
                        panic!()
                    };
                    return Ok(PropertyDefinition::MethodDefinition(
                        if name == "get" {
                            MethodDefinitionKind::Get
                        } else {
                            MethodDefinitionKind::Set
                        },
                        func_name,
                        f,
                    ));
                }
            }

            return Ok(PropertyDefinition::IdentifierReference(name));
        }
```

muskuloes:
I'll like to work on this.

HalidOdat:
> I'll like to work on this.

Go for it. :+1:

muskuloes:
I think the link in the description should point to a specific file but it points to a whole pull request with several files and I don't know where to look. Also, Is the code snippet in the description the part I should refactor?

Razican:
> I think the link in the description should point to a specific file but it points to a whole pull request with several files and I don't know where to look. Also, Is the code snippet in the description the part I should refactor?

Good catch, I updated the link in the description. It should be https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/parser/mod.rs#L1599

muskuloes:
[...]

PR #339, opened 2020-04-20
Method parsing

muskuloes:
Issue #292


boa/src/syntax/parser/mod.rs
```diff
@@ -1728,4 +1687,52 @@ impl<'a> Parser<'a> {
...
+        let (methodkind, prop_name, params) = match identifier {
+            "get" | "set" => {
+                let prop_name = self.get_object_property_name()?;
+                self.expect(
+                    TokenKind::Punctuator(Punctuator::OpenParen),
+                    Some("property method definition"),
+                )?;
+                let params = self.read_formal_parameters()?;
+                if identifier == "get" {
+                    if !params.is_empty() {
```
muskuloes, line 1705: What is the best way to handle errors here? Getters should have no arguments, while setters take only one.

jasonwilliams: If in doubt try putting an argument in a getter on a typical js environment and see what sort of error you get

muskuloes: Ok yep, I updated the PR.

PR #342, opened 2020-04-22
Fixing master build

Razican:
This should fix the build in the `master` branch. It will also properly upload a `doc/` folder to the `gh-pages` branch (for now using a `test` branch). This also removes the need to have a `secrets.USERNAME` in the repo.

It seems that the old action was broken, and this should be working better. I had to do a trick not to have a mess in the files in the branch, but I think it should be fine. Let's try it.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Upload documentation
         uses: crazy-max/ghaction-github-pages@v1
         with:
-          target_branch: gh-pages-test
+          target_branch: gh-pages
           keep_history: true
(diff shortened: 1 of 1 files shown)


Razican:
If It's ok, I can merge this.

jasonwilliams:
> This also removes the need to have a secrets.USERNAME in the repo.

Could you explain this a bit more?

Razican:
> > This also removes the need to have a secrets.USERNAME in the repo.
> 
> Could you explain this a bit more?

Yep, in #330 we added a new secret, but this uses a new GitHub action that no longer needs it.

Issue #88, opened 2019-08-21
implement in operator


arbroween:
The javascript built-in `in` operator is missing in the implementation of the lexer and also in the later stages. It is used to test if a property is in an object or its prototype chain.

It should probably be added to the lexer in this method, somewhere in the big match statement https://github.com/jasonwilliams/boa/blob/master/src/lib/syntax/lexer.rs#L207-L533

The operator's syntax is defined here https://tc39.es/ecma262/#sec-relational-operators in the spec and its semantics here below https://tc39.es/ecma262/#sec-relational-operators-runtime-semantics-evaluation

jasonwilliams:
Thanks for raising @arbroween

arbroween:
You're welcome :+1: 
I followed @jorendorff 's suggestion on the project's discord :)

arbroween:
Also following its suggestion this issue could be labelled as `good first issue`

brian-gavin:
Hey, I'm interested in doing this for Hacktoberfest!

jasonwilliams:
@brian-gavin assigned

PR #350, opened 2020-04-26
feat(boa): in operator

brian-gavin:
For parsing:

The expression! macro is extended to support a Keyword typed value in its
op arguments. This followed the spec that some expressions are separated
by Punctuators (like == etc.) but others are separated by Keywords (like
this one, or others such as instanceof).

This needed an implementation of PartialEq for Keyword and Punctuator in
order to fix a type error caused by the repeated match guard.

An as_binop method was added to Keyword to mimic the one in Punctuator.

This may be better suited to a trait implementation of TryFrom for the
BinOp enum, with implementations for TryFrom<Keyword> and
TryFrom<Punctuator>.

For interpreting:

Some abstract operations were implemented that are used in the semantics
of the operator in the spec.

An as_object cast method was addedto ValueData in order to get a ref to
that data as an object in order to use the Object methods on a
ValueData value.

Tests:

Add unit tests to interpreter and parser for the new operator.

Solves #88


Razican:
I'm waiting for #304 to land before proceeding with this, just to avoid more merge conflicts :)

brian-gavin:
Wow, that's quite the change - no worries!

Razican:
#304 has been merged :D but now this gives a lot of merge conflicts. Could you rebase?

boa/src/syntax/ast/keyword.rs
```diff
@@ -433,6 +434,21 @@ pub enum Keyword {
...
+impl Keyword {
+    pub fn as_binop(self) -> Option<BinOp> {
+        match self {
+            Keyword::In => Some(BinOp::Comp(CompOp::In)),
+            _ => None,
+        }
+    }
+}
+
+impl Into<BinOp> for Keyword {
```
Razican, line 446: An `Into` trait must not fail, I think that with the `as_binop()` method is enough.

boa/src/syntax/ast/op.rs
```diff
@@ -579,6 +579,8 @@ pub enum CompOp {
     /// [spec]: https://tc39.es/ecma262/#prod-RelationalExpression
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators#Less_than_or_equal_operator
     LessThanOrEqual,
+    /// `a in b` - If key `a` is in object `b`
```
Razican, line 582: Could we add similar documentation here to previous variants of the enum?

boa/src/syntax/ast/punc.rs
```diff
@@ -169,6 +169,12 @@ impl Punctuator {
     }
 }
 
+impl Into<BinOp> for Punctuator {
+    fn into(self) -> BinOp {
+        self.as_binop().expect("Could not get binary operation")
```
Razican, line 174: Once again, this should not fail. We can implement `TryInto` if we still need something like this.

brian-gavin: Ah yeah, `TryInto` is definitely more correct here.

The reasoning for impl'ing these traits is to abstract the main portion of the `parse` method exposed by the `expression!` macro (in which `as_binop` is called) into a generic helper that is generic over the `op` type - since before it was only a Punctuator, it now needs to have Keyword support.

So, both Punctuator and Keyword are impl'ing `Into<BinOp>` (or now TryInto<BinOp>) in order to support both Keyword and Punctuator in the main `match` over the token.

This method and usage can be found [here](https://github.com/jasonwilliams/boa/pull/350/files#diff-805f77066eac33b82808bcee94de20d7R73) - I think that my explanation will make more sense while also seeing the implementation 😄

boa/src/exec/mod.rs
```diff
@@ -739,6 +747,31 @@ impl Interpreter {
...
+            if !Property::is_property_key(key) {
+                false
+            } else {
+                obj.has_property(key)
+            }
+        } else {
+            false
+        }
+    }
+
```
HalidOdat, line 774: We only check for the objects property, we should also check the property chain.

`"toString" in {}` should return `true`.

boa/src/syntax/parser/expression/mod.rs
```diff
@@ -28,6 +47,23 @@ use crate::syntax::ast::{node::Node, punc::Punctuator, token::TokenKind};
...
 macro_rules! expression { ($name:ident, $lower:ident, [$( $op:path ),*], [$( $low_param:ident ),*] ) => {
+    impl $name {
+        // Helper method to build a BinOp in the parse method
+        fn build_op<O, L>(&self, op: O, lhs: L, cursor: &mut Cursor<'_>) -> ParseResult
+        where
+            O: TryInto<BinOp>,
+            <O as TryInto<BinOp>>::Error: Debug, // for call to `expect`
+            L: Into<Box<Node>>
+        {
+            let _ = cursor.next().expect("token disappeared");
```
Razican, line 58: Moving the cursor in this function seems a bit strange. This function should expect that the cursor is ready to parse. It' s easier to reason that way. The `cursor.next()` if needed, should be done before calling it.

boa/src/syntax/parser/expression/mod.rs
```diff
@@ -36,12 +72,10 @@ macro_rules! expression { ($name:ident, $lower:ident, [$( $op:path ),*], [$( $lo
             while let Some(tok) = cursor.peek(0) {
                 match tok.kind {
                     TokenKind::Punctuator(op) if $( op == $op )||* => {
-                        let _ = cursor.next().expect("token disappeared");
-                        lhs = Node::bin_op(
-                            op.as_binop().expect("could not get binary operation"),
-                            lhs,
-                            $lower::new($( self.$low_param ),*).parse(cursor)?
-                        )
+                        lhs = self.build_op(op, lhs, cursor)?;
```
Razican, line 75: I understand the idea of adding the helper function, but this adds lots of code to optimise, with pretty complex generics, that might slow things down when compiling, and it's only two statements. Would it make sense duplicating the code in the two branches of the `match`?

boa/src/exec/mod.rs
brian-gavin: I'm under the assumption that `has_property` should be going up the prototype chain.

I've looked into it, and [it is](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/object/internal_methods_trait.rs#L37). It seems like my assumption is broken somewhere down the line.

For this empty object, it makes a call to [get_prototype_of](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/object/internal_methods_trait.rs#L181) which calls `get_internal_slot` with the `PROTOTYPE` constant (which is `"prototype"`. There is no `prototype` field in the internal slots of this object, so this example is evaluating to `false`.

However, the base Object is within the `__proto__` internal slot.

I would assume that `get_prototype_of` would return with the base object that all other object inherit from when there are no other prototype.

I'm confused on this point because the standard is a little cryptic to me: as [\[\[GetPrototypeOf\]\]()](https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-getprototypeof) just calls [OrdinaryGetPrototypeOf](https://tc39.es/ecma262/#sec-ordinarygetprototypeof) which simply returns `O.[[Prototype]]`.

From looking at the rest of the codebase, it isn't clear to me whether `INSTANCE_PROTOTYPE` (`"__proto__"`) or `PROTOTYPE` should be used when the standard says `[[Prototype]]`.

boa/src/syntax/parser/expression/mod.rs
brian-gavin: I have decided to remove the helper - you're right, it's too complicated for just two statements, haha.

boa/src/exec/mod.rs
jasonwilliams: > From looking at the rest of the codebase, it isn't clear to me whether INSTANCE_PROTOTYPE ("__proto__") or PROTOTYPE should be used when the standard says [[Prototype]].

When the standard says [[Prototype]] they mean internal slot, so that’s “__proto__” for us (or PROTOTYPE_INSTANCE). 

“prototype” is a regular property so would not be in double brackets.

Does this help?

jasonwilliams: > For this empty object, it makes a call to get_prototype_of which calls get_internal_slot with the PROTOTYPE constant (which is "prototype". There is no prototype field in the internal slots of this object, so this example is evaluating to false.

Yeah I’m sure this is wrong, it should be checking for \_\_proto\_\_ not prototype

HalidOdat: > > For this empty object, it makes a call to get_prototype_of which calls get_internal_slot with the PROTOTYPE constant (which is "prototype". There is no prototype field in the internal slots of this object, so this example is evaluating to false.
> 
> Yeah I’m sure this is wrong, it should be checking for __proto__ not prototype

Changing from `PROTOTYPE` to `INSTANCE_PROTOTYPE` now gives the desirable output `true` for `"toString" in {}`

brian-gavin: I see - thanks a lot you two! I'll add the fix for INSTANCE_PROTOTYPE.

Issue #333, opened 2020-04-18
Make WASM an opt-in feature


Razican:
Currently, webassembly is an opt-out feature. While we use it for the web example, I think in the future it will be more common for Boa to be used as a library than as webassembly.

In this scenario, I think it's best to make the feature opt-in, so that we don't need to use `default-features = false` when loading the crate as a library.

This would mean that we need somehow to be able to select a feature in wasm-pack, or that we need a different crate only exporting boa as wasm, in the workspace, something like `boa-wasm`.

I would go for this second option, as it also allows us to add `#![forbid(unsafe_code)]` at the crate level in `boa`.

What do you think?

HalidOdat:
> This would mean that we need somehow to be able to select a feature in wasm-pack, or that we need a different crate only exporting boa as wasm, in the workspace, something like `boa-wasm`.

I also think this is a good idea as it would allow us to move the most of the webassembly stuff in that directory, making think more organized.

And not having to specify that the `default-features = false` is awesome as the default features should be features that every one needs by default.

jasonwilliams:
Yep agree with the above it should be opt in rather than opt out.

muskuloes:
I would like to take this one. But will only be able to work on it from next week, if that's ok.

Razican:
> I would like to take this one. But will only be able to work on it from next week, if that's ok.

Sure, go ahead, let us know when you start! :)

muskuloes:
I can start now.

PR #352, opened 2020-04-28
create boa-wasm package

muskuloes:
#333

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,7 +12,6 @@ dependencies = [
  "regex",
  "serde",
  "serde_json",
- "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


muskuloes:
The cli uses the `serde-ast` feature from `boa` lib.

HalidOdat:
> The cli uses the `serde-ast` feature from `boa` lib.

Removing `default-features = false` should not interfere with not non-default features like `serde-ast`.

muskuloes:
> > The cli uses the `serde-ast` feature from `boa` lib.
> 
> Removing `default-features = false` should not interfere with non-default features like `serde-ast`.

You are right. I miss understood the documentation on that :sweat_smile:

PR #358, opened 2020-05-01
Feature `BigInt`

### LINKED ISSUE #121: [Feature] - BigInt
https://tc39.es/ecma262/#sec-bigint-objects

HalidOdat:
This PR resolves #121

### DIFF
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -24,9 +24,16 @@ use crate::{
 #[cfg(test)]
 mod tests;
 
-/// `BigInt()` function.
+/// `BigInt()`
 ///
-/// More Information https://tc39.es/ecma262/#sec-number-constructor-number-value
+/// The `BigInt()` constructor is used to create BigInt objects.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-bigint-objects
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/BigInt
 pub fn make_bigint(_this: &mut Value, args: &[Value], _ctx: &mut Interpreter) -> ResultValue {
     let data = match args.get(0) {
         Some(ref value) => {
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -1,0 +1,99 @@
+//! This module implements the global `BigInt` object.
+//!
+//! `BigInt` is a built-in object that provides a way to represent whole numbers larger
+//! than the largest number JavaScript can reliably represent with the Number primitive
+//! and represented by the `Number.MAX_SAFE_INTEGER` constant.
+//! `BigInt` can be used for arbitrarily large integers.
+//!
+//! More information:
+//!  - [ECMAScript reference][spec]
+//!  - [MDN documentation][mdn]
+//!
+//! [spec]: https://tc39.es/ecma262/#sec-bigint-objects
+//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+
+use crate::{
+    builtins::{
+        function::make_constructor_fn,
+        value::{ResultValue, Value},
+    },
+    exec::Interpreter,
+    syntax::ast::bigint::BigInt,
+};
+
+#[cfg(test)]
+mod tests;
+
+/// `BigInt()` function.
+///
+/// More Information https://tc39.es/ecma262/#sec-number-constructor-number-value
+pub fn make_bigint(_this: &mut Value, args: &[Value], _ctx: &mut Interpreter) -> ResultValue {
+    let data = match args.get(0) {
+        Some(ref value) => {
+            if let Some(bigint) = value.to_bigint() {
+                Value::from(bigint)
+            } else {
+                panic!("RangeError: The value cannot be converted to a BigInt because it is not an integer");
+            }
+        }
+        None => Value::from(BigInt::from(0)),
+    };
+    Ok(data)
+}
+
+/// `BigInt.prototype.toString( [radix] )`
+///
+/// The `toString()` method returns a string representing the specified BigInt object.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-bigint.prototype.tostring
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toString
+pub fn to_string(this: &mut Value, args: &[Value], _ctx: &mut Interpreter) -> ResultValue {
+    let radix = if !args.is_empty() {
+        args[0].to_integer()
+    } else {
+        10
+    };
+    if radix < 2 && radix > 36 {
+        panic!("RangeError: toString() radix argument must be between 2 and 36");
+    }
+    Ok(Value::from(
+        this.to_bigint().unwrap().to_str_radix(radix as u32),
+    ))
+}
+
+// /// `BigInt.prototype.valueOf()`
+// ///
+// /// The `valueOf()` method returns the wrapped primitive value of a Number object.
+// ///
+// /// More information:
+// ///  - [ECMAScript reference][spec]
+// ///  - [MDN documentation][mdn]
+// ///
+/// [spec]: https://tc39.es/ecma262/#sec-bigint.prototype.valueof
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/valueOf
+pub fn value_of(this: &mut Value, _args: &[Value], _ctx: &mut Interpreter) -> ResultValue {
+    Ok(Value::from(
+        this.to_bigint().expect("BigInt.prototype.valueOf"),
+    ))
+}
+
+/// Create a new `Number` object
+pub fn create(global: &Value) -> Value {
+    let prototype = Value::new_object(Some(global));
+    prototype.set_internal_slot("BigIntData", Value::from(BigInt::from(0)));
+
+    make_builtin_fn!(to_string, named "toString", with length 1, of prototype);
+    make_builtin_fn!(value_of, named "valueOf", of prototype);
+
+    make_constructor_fn(make_bigint, global, prototype)
+}
+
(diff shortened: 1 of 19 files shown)


Razican:
Hey, how is this going? can we help?

HalidOdat:
> Hey, how is this going? can we help?

The reason I didn't update this PR is because I was preoccupied with refactoring. I wanted to refactor things before contributing to the mess. (Already have some progress locally)

Speaking of help, I have a question about which crate should use.
The best candidates are currently `rug` and `num-bigint`.
Pros of `rug`:
 - It's very **optimized**, (internally it uses `gmp-sys`)
 - It's got every thing I need for implementing BigInt.

Cons of `rug`:
 - It has and dependency on an c library (slower compile-time)
 - On windows it requires `msys2` to be preinstalled as a compilation dependency.

Pros of `num-bigint`:
 - It requires no c dependencies.
 - Faster compilation times.

Cons of `num-bigint`:
 - It's not the fastest implementation (they don't claim either)
 - requires `num-trait` crate to get some needed functionality.

What should I use? what is the best choice? @Razican

Razican:
> > Hey, how is this going? can we help?
> 
> The reason I didn't update this PR is because I was preoccupied with refactoring. I wanted to refactor things before contributing to the mess. (Already have some progress locally)

This sounds very reasonable, thanks a lot for that!

> Speaking of help, I have a question about which crate should use.
> The best candidates are currently `rug` and `num-bigint`.
> Pros of `rug`:
> 
>     * It's very **optimized**, (internally it uses `gmp-sys`)
> 
>     * It's got every thing I need for implementing BigInt.
> 
> 
> Cons of `rug`:
> 
>     * It has and dependency on an c library (slower compile-time)
> 
>     * On windows it requires `msys2` to be preinstalled as a compilation dependency.
> 
> 
> Pros of `num-bigint`:
> 
>     * It requires no c dependencies.
> 
>     * Faster compilation times.
> 
> 
> Cons of `num-bigint`:
> 
>     * It's not the fastest implementation (they don't claim either)
> 
>     * requires `num-trait` crate to get some needed functionality.
> 
> 
> What should I use? what is the best choice? @Razican

I think we should go for `num-bigint`. One of the things we are trying to develop is a pure-Rust engine, even if it's not currently 100% optimised. Maybe in the future we can contribute to `num-bigint` to improve performance, what do you think, @jasonwilliams ?

About using extra crates, I think it's ok, we only use it because we really need it, and dependencies are small enough I think so that they won't add so much burden on compile times.

HalidOdat:
> I think we should go for `num-bigint`. One of the things we are trying to develop is a pure-Rust engine, even if it's not currently 100% optimised. Maybe in the future we can contribute to `num-bigint` to improve performance, what do you think, @jasonwilliams ?
> 
> About using extra crates, I think it's ok, we only use it because we really need it, and dependencies are small enough I think so that they won't add so much burden on compile times.

Thanks for the feedback that really helps.

In the future if `num-bigint` is not sufficient for us, we could implement our own.
V8 has a blog about this: <https://v8.dev/blog/bigint>

HalidOdat:
While trying to implement this, I ran into a problem with the abstract equality, so this is blocked until #395 lands.

jasonwilliams:
> I think we should go for num-bigint. One of the things we are trying to develop is a pure-Rust engine, even if it's not currently 100% optimised. Maybe in the future we can contribute to num-bigint to improve performance, what do you think, @jasonwilliams ?

agree 100%

HalidOdat:
After #320 got merged `[[call]]` and `[[construct]]` became one function, but `BigInt` Object it can be call, but it cannot be constructed (`new BigInt` should throw an exception).

So how am I supposed to do this? @jasonwilliams

boa/src/builtins/bigint/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    ))
+}
+
+/// Create a new `Number` object
+pub fn create(global: &Value) -> Value {
+    let prototype = Value::new_object(Some(global));
+    prototype.set_internal_slot("BigIntData", Value::from(BigInt::from(0)));
+
+    make_builtin_fn!(to_string, named "toString", with length 1, of prototype);
+    // make_builtin_fn!(value_of, named "valueOf", of prototype);
```
Razican, line 95: Is this not ready yet?

boa/src/builtins/value/tests.rs
```diff
@@ -90,7 +90,7 @@ fn abstract_equality_comparison() {
     assert_eq!(forward(&mut engine, "'-0' == 0"), "true");
 
     // https://github.com/jasonwilliams/boa/issues/393
```
Razican, line 92: Do we need this comment here?

boa/src/exec/mod.rs
```diff
@@ -741,7 +743,9 @@ impl Interpreter {
             _ => input.clone(),
         }
     }
+
     /// to_string() converts a value into a String
```
Razican, line 747: This should either be:
```suggestion
    /// `to_string()` converts a value into a `String`.
```
 or:

```suggestion
    /// Converts a value into a `String`.
```

boa/src/builtins/bigint/mod.rs
HalidOdat: Yes. it is, forgot to remove the comment. :)

boa/src/builtins/value/tests.rs
HalidOdat: Probably not, its unnecessary.

boa/src/builtins/bigint/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    exec::Interpreter,
+    syntax::ast::bigint::BigInt,
+};
+
+#[cfg(test)]
+mod tests;
+
+/// `BigInt()` function.
+///
+/// More Information https://tc39.es/ecma262/#sec-number-constructor-number-value
```
jasonwilliams, line 29: This link points to the `Number` constructor, but the BigInt constructor is here:
https://tc39.es/ecma262/#sec-bigint-constructor

boa/src/builtins/value/mod.rs
```diff
@@ -163,6 +174,8 @@ pub enum ValueData {
     Rational(f64),
     /// `Number` - A 32-bit integer, such as `42`
     Integer(i32),
+    /// ---
```
jasonwilliams, line 177: was there supposed to be a comment here stating what BigInt is?

boa/src/builtins/value/operations.rs
```diff
@@ -44,32 +53,131 @@ impl Value {
...
+/// where x and y are ECMAScript language values, produces true or false.
+/// Such a comparison is performed as follows:
+///
+/// https://tc39.es/ecma262/#sec-samevalue
+/// strict mode currently compares the pointers
+pub fn same_value(x: &Value, y: &Value, strict: bool) -> bool {
+    if strict {
+        // Do both Values point to the same underlying valueData?
+        let x_ptr = Gc::into_raw(x.0.clone());
+        let y_ptr = Gc::into_raw(y.0.clone());
```
jasonwilliams, line 136: is `into_raw` safe? I thought it consumes the GC?
https://docs.rs/gc/0.3.4/gc/struct.Gc.html#method.into_raw

jasonwilliams: response from Manish

```
It's a leak because this GC object is permanently rooted after this
This is safe, but it's a leak
Instead, dereference the GC<T> to get an &T and convert that to a *const T
This is not what into_raw is for
```

Manishearth: It's safe, but it will permanently root the object.

Convert the `Gc<T>` into an `&T` via `&*` and then convert that to a `*const T` for comparison

boa/src/builtins/value/operations.rs
```diff
@@ -44,32 +53,129 @@ impl Value {
...
+
+            // 13. Return false.
             _ => false,
         }
     }
 }
 
+/// This function takes a string and conversts it to BigInt type.
+///
+/// If the result is NaN than None is returned.
```
Razican, line 112:
```suggestion
/// If the result is `NaN` than `None` is returned.
```

boa/src/syntax/ast/bigint.rs
```diff
@@ -0,0 +1,243 @@
...
+impl std::ops::DerefMut for BigInt {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl Finalize for BigInt {}
+unsafe impl Trace for BigInt {
+    unsafe_empty_trace!();
+}
```
Razican, line 243: We should mark this with a comment, since this would be unsound if `num_bigint::BigInt` implements `Trace` in the future.

boa/src/syntax/ast/constant.rs
```diff
@@ -64,6 +65,8 @@ pub enum Const {
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Numeric_literals
     Int(i32),
 
+    BigInt(BigInt),
+
```
Razican, line 69: I would add a comment here, as with other variants.

boa/src/builtins/value/operations.rs
```diff
@@ -44,32 +53,129 @@ impl Value {
...
+/// The internal comparison abstract operation SameValue(x, y),
+/// where x and y are ECMAScript language values, produces true or false.
+/// Such a comparison is performed as follows:
+///
+/// https://tc39.es/ecma262/#sec-samevalue
+/// strict mode currently compares the pointers
+pub fn same_value(x: &Value, y: &Value, strict: bool) -> bool {
+    if strict {
+        // Do both Values point to the same underlying valueData?
+        return std::ptr::eq(x.data(), y.data());
```
jasonwilliams, line 135: Nice!
Where does `data()` come from? 
I didn't see it in Gc's documentation

jasonwilliams: Also was there a reason you didn't use https://docs.rs/gc/0.3.4/gc/struct.Gc.html#method.ptr_eq ?

HalidOdat: > Where does `data()` come from?
> I didn't see it in Gc's documentation

I abstracted Value as a `struct Value(Gc<ValueData>)` and added a helper function `Value::data()` which returns a `&ValueData` in #383 instead of doing `&**x` you can write just `x.data()`.

> Also was there a reason you didn't use [docs.rs/gc/0.3.4/gc/struct.Gc.html#method.ptr_eq](https://docs.rs/gc/0.3.4/gc/struct.Gc.html#method.ptr_eq) ?

Because `Value` is abstracted as a `struct` you can not pass is as a `Gc<ValueData>`. This functionally does the same thing

PR #355, opened 2020-04-28
Removed *.bundle.js from editor WebWorkers

subhankar-panda:
Fixes #314

PR #362, opened 2020-05-01
Remove Monaco Editor Webpack Plugin and Manually Vendor Editor Workers

### LINKED ISSUE #314: Live demo has console errors in browser (monaco editor)
image

I'm not sure why monaco can't find a web worker.
You can test this locally though via running `yarn serve` then going to `localhost:8080`

subhankar-panda:
Follows the "manual approach" listed [here](https://github.com/Microsoft/monaco-editor/blob/master/docs/integrate-esm.md#option-2-using-plain-webpack).

Fixes #314.

The issue with #355 was that it was never able to find the ts worker - this manually tells webpack where to look for the files


jasonwilliams:
Hey @subhankar-panda  thanks for this!
Do you know if there's anything we can do about getting rid of all the language-support we don't need.

When you run `yarn build` then look in `dist` folder it generates a lot of js files, for all the different syntaxes it supports, is there a way to just have javascript?

PR #368, opened 2020-05-02
Changed HashMap and HashSet for Fx versions

### LINKED ISSUE #356: Use Fx Hash to speed up hash maps in the compiler
We currently use standard library hash maps ([`std::collections::HashMap`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html)) in Boa to organize all of our internal slots and more. This is not very performant, and [Fx Hash](https://crates.io/crates/fxhash) has just a bit lower DoS resistant, while providing much better performance.

We do not care about DoS in a compiler, usually, so this should improve our performance greatly, especially during the execution.

Razican:
This fixes #356. Note that this requires the git version of `rust-gc` until https://github.com/Manishearth/rust-gc/issues/92 gets done. An upgrade to rust-gc would also include new types implementing `Trace` OOB and a very reduced dependency tree without having to depend on many `syn` and `quote` versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,8 +252,8 @@ checksum = "5f2a4a2034423744d2cc7ca2068453168dcdb82c438419e639a26bd87839c674"
 
 [[package]]
 name = "gc"
-version = "0.3.3"
-source = "git+https://github.com/Manishearth/rust-gc.git#87bb44e0e3b5824df1669415a2766b9ee34dc9a4"
+version = "0.3.4"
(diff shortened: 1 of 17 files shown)


jasonwilliams:
So you know when Manish plans to release a version?

Razican:
> So you know when Manish plans to release a version?

Nope, I will try to contact him during next week to see his plans.

Razican:
This is now ready for review and merge!

HalidOdat:
Everything looks perfect to me!

PR #370, opened 2020-05-05
Optimised all `Vec<Node>` in `Node` to be `Box<[Node]>`

### LINKED ISSUE #369: Change `Vec<Node>` in `Node` to `Box<[Node]>`
We currently use `Vec<Node>` in multiple places in our [`Node`][node_impl] implementation. A `Vec` occupies more than a `Box<[T]>`, since it needs to store the capacity. We will not modify those vectors, so we could call `into_boxed_slice()` in them an just have smaller enum variants.

The only issue is that `Box<[T]>` doesn't implement `Trace`, so this is blocked on https://github.com/Manishearth/rust-gc/issues/89.

[node_impl]: https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/ast/node.rs#L16-L941

Razican:
This closes #369 but it's blocked on `rust-gc` releasing a new version, which should happen anytime now, since it has already been tagged.

### DIFF
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -92,7 +92,7 @@ pub struct Function {
     /// Call/Construct Function body
     pub body: FunctionBody,
     /// Formal Paramaters
-    pub params: Vec<FormalParameter>,
+    pub params: Box<[FormalParameter]>,
     /// This Mode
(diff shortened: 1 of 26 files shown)


Razican:
This is ready for review / merge, now that the new `gc` has been released in crates.io :)

PR #372, opened 2020-05-06
Code cleanup

HalidOdat:
This PR is for cleaning the code-base mostly `builtins`, `exec` and `environment`.

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -102,13 +102,13 @@ pub fn make_array(this: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Re
     // between indices and values): this creates an Object with no prototype
 
     // Set Prototype
-    let array_prototype = ctx
+    let prototype = ctx
         .realm
         .global_obj
         .get_field_slice("Array")
         .get_field_slice(PROTOTYPE);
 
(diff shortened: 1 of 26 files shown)


HalidOdat:
For now I think this PR is ready :)

I plan do another PR that will abstract/refactor the use of `Gc<ValueData>`.

Encapsulating `Gc<ValueData>` in a `struct Value(Gc<ValueData>)`, instead of it being a type alias. (currently it's `type Value = Gc<ValueData>`)
which will allow us to abstract and clean the code a lot, instead of `Gc::new<ValueData::Undefined>` or a free function `undefined()` (which is kind of ambiguous) we could have a static `Value::undefined()` and `Value::null()` and so on.

This would also allow us to use standard `From` trait.

This it will take some time, since `value` alias is used everywhere as `Gc<ValueData>`. changing it generates around +300 errors in +20 files. :sweat_smile: 

HalidOdat:
> This looks very good to me! I like your idea of using a `Value(Gc<ValueData>)` type in a following PR, that would greatly help with the idiomatization of the code. But it will take some time to fix those 1,000 erors, so let's land this asap :)

After some work I dropped the errors down to around 60 in 10 files, It might be easier than I thought initially, Lets hope thats the case. It might be the calm before the storm.

PR #373, opened 2020-05-06
Introduce PropertyKey for field acces, fix #172 (quotes around displayed strings)

### LINKED ISSUE #172: There are no quotes around a String value
Provided you have something like:
```rust
const o = { str: 'String' };

o;
```
the output would be:
```
"{ str: String }"
```

The issue is due to the [implementation](https://github.com/jasonwilliams/boa/blob/master/src/lib/js/value.rs#L579) of `Display` for `ValueData::String`, which simply displays the value of the `String`.

> Note that currently in `master` objects are displayed simpy as `{}` but you can see the issue in the branch for #171.

RageKnify:


RageKnify:
I have no idea how, but changing the `fmt` method for `ValueData::String` broke something related to building `Array`s, now `new Array()` returns `undefined`.
Any ideia on what I might be doing wrong?

Razican:
> I have no idea how, but changing the `fmt` method for `ValueData::String` broke something related to building `Array`s, now `new Array()` returns `undefined`.
> Any ideia on what I might be doing wrong?

This might be related to the fact that we use `ToString` in many places, and `ToString` is derived from `Display`. Maybe we somehow need to differentiate between our internal representation of a `String` and what we want to print in the screen on a `console.log()` for example.

@HalidOdat @jasonwilliams what do you think?

jasonwilliams:
> Maybe we somehow need to differentiate between our internal representation of a String and what we want to print in the screen on a console.log() for example.

Yes we certainly want this, even for debugging purposes.

Razican:
> > Maybe we somehow need to differentiate between our internal representation of a String and what we want to print in the screen on a console.log() for example.
> 
> Yes we certainly want this, even for debugging purposes.

The question is why is this giving an error. Are we using a `toString()` somewhere to convert to a `String` type of value? I'll try to check it.

IovoslavIovchev:
It doesn't seem to be related to `toString()`. For some reason when the change is made all of the globals are `undefined`. The issue might be the bindings don't get initialized at all or `get_bindings_value` isn't returning the proper value (both are wild guesses).

Razican:
After reviewing what's happening here, I saw that many tests fail:
```
test result: FAILED. 111 passed; 89 failed; 6 ignored; 0 measured; 0 filtered out
```

So it seems that we use `to_string()` in many places to convert a string value to a string. This should probably be fixed somehow, but I don't think it will be easy. The current task is to fix the display of these values in the console, and I think that can be done by creating a `ValueDisplay` structure, that would also implement `Display`, and that we only use to show this in the console. It would just be a wrapper over `Value`, and could be created by calling a `display()` method in the `Value`.

What do you think?

Razican:
Does this work now that #381 was merged?

RageKnify:
> Does this work now that #381 was merged?

No, the samer error remains, global objects are `undefined`.
`a = 0;` works, and I can do `a+1`.
But `a = new Array()` still returns `undefined`.

RageKnify:
I have some free time again, I'd like to fix this.
Should I try to implement the solution proposed by @Razican ?

Create a wrapper struct, implementing `Display` that would be created upon calling `display()` on an instance of  `Value`?
I'd then change the line: https://github.com/boa-dev/boa/blob/d837e040c9f12ceac4284f4fb557808f006f742d/boa_cli/src/main.rs#L211
to call `v.display()` instead, yes?

Razican:
> I have some free time again, I'd like to fix this.
> Should I try to implement the solution proposed by @Razican ?
> 
> Create a wrapper struct, implementing `Display` that would be created upon calling `display()` on an instance of  `Value`?
> I'd then change the line: https://github.com/boa-dev/boa/blob/d837e040c9f12ceac4284f4fb557808f006f742d/boa_cli/src/main.rs#L211
> to call `v.display()` instead, yes?

I think this is the best solution for now, yes. What do you think, @HalidOdat, @jasonwilliams ?

HalidOdat:
> I think this is the best solution for now, yes. What do you think, @HalidOdat, @jasonwilliams ?

I think so too!

RageKnify:
Since  `log_string_from` receives a `&ValueData` I made it a wrapper over `&ValueData` instead of over `&Value` so that it could be used in `log_string_from`.
https://github.com/boa-dev/boa/blob/d837e040c9f12ceac4284f4fb557808f006f742d/boa/src/builtins/value/mod.rs#L833

I don't in which places `.display()` should be invoked besides the `main.rs` of `boa_cli`.

Let me know what can be improved!

boa/src/builtins/value/mod.rs
```diff
@@ -982,3 +996,21 @@ impl Display for ValueData {
         }
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct DisplayValueData<'vd>(pub(crate) &'vd ValueData);
```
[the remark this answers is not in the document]
RageKnify: So change everything related to logging to use `&Value` instead, including the new struct, right?
Change its name to `DisplayValue`.
Using `pub(crate)` was blindly following other declaraitons in the file, I'll take it out.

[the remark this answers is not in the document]
HalidOdat: > So change everything related to logging to use `&Value` instead, including the new struct, right?
> Change its name to `ValueDisplay`.
> Using `pub(crate)` was blindly following other declaraitons in the file, I'll take it out.

Yes. pretty much!

RageKnify:
> About the `log_string_from` I think we should change it to take in a `&Value` instead of `&ValueData`

I'm having an issue with this, the `Display` implementation of `ValueData` relies on `log_string_from`, and it doesn't have a `&Value` to give, it only has `&ValueData`, what should I do?
https://github.com/boa-dev/boa/blob/c8218dd91ef3181e048e7a2659a4fbf8d53c7174/boa/src/builtins/value/mod.rs#L979

HalidOdat:
> I'm having an issue with this, the `Display` implementation of `ValueData` relies on `log_string_from`, and it doesn't have a `&Value` to give, it only has `&ValueData`, what should I do?

Hmmm. For now lets leave `log_string_from` how it is.

RageKnify:
I have pushed it, but it leaves an issue, objects that have fields of type `String` won't be printed correctly:
```js
let a = {0: "0"};
a
```
should display `0: "0"` but instead shows `0: 0`

HalidOdat:
The problem is that we implement `Display` trait for `Value` (`Value`s fmt function calls `ValueData`s fmt function) and with display we we get `ToString` trait as an extra and we use that `to_string` which is meant for displaying as the [ECMASCript `ToString` function](https://tc39.es/ecma262/#sec-tostring). these are not the same that is why we get this weird behavior. The logic should be separate.

The `ToString`s to_string method is not spec compliant!!

I'll create a PR addressing this issue ASAP.

Razican:
I think this can just change the `Display` implementation. It will also need some tests.

RageKnify:
> I think this can just change the `Display` implementation. It will also need some tests.

Are you proposing just changing the `Display` implementation for `ValueData::String` like I tried originally?
I tried again, `write!(f, "\"{}\"", v),` but a bunch of tests fail still and displaying objects becomes weird, I got:
```js
let c = {0: '0'}
undefined
c
{
""0"": "0",
__proto__: undefined,
extensible: true
}
```
Trying to figure out why the "index" gets surrounded by `""`.

HalidOdat:
> > I think this can just change the `Display` implementation. It will also need some tests.
> 
> Are you proposing just changing the `Display` implementation for `ValueData::String` like I tried originally?
> I tried again, `write!(f, "\"{}\"", v),` but a bunch of tests fail still and displaying objects becomes weird, I got:
> 
> ```js
> let c = {0: '0'}
> undefined
> c
> {
> ""0"": "0",
> __proto__: undefined,
> extensible: true
> }
> ```
> 
> Trying to figure out why the "index" gets surrounded by `""`.

I believe this is caused because `.set_field` and `.get_field` use `Display`s `to_string`

Razican:
> I believe this is caused because `.set_field` and `.get_field` use `Display`s `to_string`

I don't think they should, right?

HalidOdat:
> I don't think they should, right?

`set_field` and `get_field` is like `object[property]`, where if the property is a `Value` (of any type) and it's converted to an string with `ToString` and that string is used as a key. but how we use it's like `object.property`. where `property` is always guaranteed to be a string.

I'm not sure what to do in  this scenario? should we create a new function `set_slice` or change the behaviour of `set_field`.

What do you think? @Razican

Razican:
> > I don't think they should, right?
> 
> `set_field` and `get_field` is like `object[property]`, where if the property is a `Value` (of any type) and it's converted to an string with `ToString` and that string is used as a key. but how we use it's like `object.property`. where `property` is always guaranteed to be a string.
> 
> I'm not sure what to do in  this scenario? should we create a new function `set_slice` or change the behaviour of `set_field`.
> 
> What do you think? @Razican 

That `ToString` is the `interpreter.to_string()`, right? We just need to pass it as an extra argument to `get/set_field()`, right?

HalidOdat:
> That `ToString` is the `interpreter.to_string()`, right? We just need to pass it as an extra argument to `get/set_field()`, right?

yes. thats probably the best option.

Razican:
> > That `ToString` is the `interpreter.to_string()`, right? We just need to pass it as an extra argument to `get/set_field()`, right?
> 
> yes. thats probably the best option.

We could maybe have a fast path for `&str` though.

RageKnify:
@HalidOdat 's refactor of `Value` simplified this a lot, I'll work on tests tomorrow, I was thinking:
- plain string
- object with a string as the value of a field
- array with a string (which I've just now Identified as a bug and found the fix)

Any more ideas?

HalidOdat:
> I'd also like tips on where to place the tests?

I think the best place to put it is in `value/tests.rs`

RageKnify:
I don't love the way the test for the object is written, feels bad to hardcode so much text, if there's any ideas to improve it I'll try.

Changing only the implementation of the `fmt` for `Value::String` doesn't work because `Value::to_string` is used in places that should use `Interpreter::to_string()` instead according to @Razican and @HalidOdat.
I couldn't figure out where `get_field` and `set_field` had the missuses of `Value::to_string`, but if you could guide me I can try to fix that, which would simplify this PR.

Current solution involves having an extra method `display()` and creating a struct that wraps `Value`.

HalidOdat:
> I don't love the way the test for the object is written, feels bad to hardcode so much text, if there's any ideas to improve it I'll try.

Hmmm. put an `#[ignore]` attribute on the object printing test for now, #507 should make the tests much nicer.

> I couldn't figure out where `get_field` and `set_field` had the missuses of `Value::to_string`, but if you could guide me I can try to fix that, which would simplify this PR.

we call `Value::to_string()` `for set_field` here: <https://github.com/boa-dev/boa/blob/master/boa/src/builtins/value/mod.rs#L694> we need to make `set_field` take a `Interpreter` like builtin functions and call the interpreters `to_string` method.

The `get_field` methods does not seem to  call `Value::to_string`, (it should) we can leave it as is for now.

RageKnify:
> we need to make set_field take a Interpreter like builtin functions and call the interpreters to_string method.

I'm trying to work on that, but I had to add `&mut Interpreter` to a lot of functions that down their call graphs invoked `Value::set_field` and ended up stuck with 2 problems:
`from(value: &Property) -> Value`, it invokes `property.set_field()` which would need an interpreter as the last argument, but `from` must have only the 1 argument.
```rust
impl From<&Property> for Value {
    fn from(value: &Property) -> Value {
        let property = Value::new_object(None);
        property.set_field("configurable", Value::from(value.configurable));
        property.set_field("enumerable", Value::from(value.enumerable));
        property.set_field("writable", Value::from(value.writable));
        property.set_field("value", value.value.clone().unwrap_or_else(Value::null));
        property.set_field("get", value.get.clone().unwrap_or_else(Value::null));
        property.set_field("set", value.set.clone().unwrap_or_else(Value::null));
        property
    }
}
```
`Error::create`, invokes `set_field`, so I added `& Interpreter` to its parameters, but `Error::create` is invoked by `Error::init` which is only supposed to have 1 parameter.

I guess the first issue might be solvable by having "fast path for &str" like @Razican proposed, the second one seems more complex.

Razican:
> I guess the first issue might be solvable by having "fast path for &str" like @Razican proposed, the second one seems more complex.

I think that the solution is to not implement `From<&Property>` for `Value`, and just have a method in `Property` that converts it to value, with the extra parameter.

RageKnify:
I was wondering if having 2 different methods would be acceptable, one that receives `&str` for the field and doesn't need a reference to the interpreter and the second that receives `Into<Value>` for the field and uses a reference to the interpreter to invoke `Interpreter.to_string(field)`. The signatures would be:

```rust
    pub fn set_str_field<V>(&self, field: &str, val: V) -> Value
    where
        V: Into<Value>,
    {}

    /// Set the field in the value
    /// Field could be a Symbol, so we need to accept a Value (not a string)
    pub fn set_field<F, V>(&self, field: F, val: V, interpreter: &mut Interpreter) -> Value
    where
        F: Into<Value>,
        V: Into<Value>,
    {}
```

RageKnify:
I've found the issue was deeper, `Object::set` receives `field: Value` and invokes `field.to_string()`, so here `length` becomes `"length"` for example, for `[]`.
https://github.com/boa-dev/boa/blob/fcb4b8b1f1135a42a96df85465f666c7cc5e9389/boa/src/builtins/object/internal_methods.rs#L147
I'm not sure how to go around this issue...

Razican:
> I'm not sure how to go around this issue...

We have one first option, which is to pass an interpreter to `Object::set()`, but I'm wondering, do we need to pass a `Value` as a field, or could we pass a `String`/`&str`?

If we can, then we just need to convert it to a `String` before calling `Object::set`.

RageKnify:
> We have one first option, which is to pass an interpreter to Object::set(), but I'm wondering, do we need to pass a Value as a field, or could we pass a String/&str?

I think it needs to be a value, `Object::set` invokes `Object::get_own_property` with the `&Value` and `get_own_property` behaves differentely for `Value::String` and `Value::Symbol`, invoking `self.properties()` or `self.symbol_properties()` respectively.

Razican:
> > We have one first option, which is to pass an interpreter to Object::set(), but I'm wondering, do we need to pass a Value as a field, or could we pass a String/&str?
> 
> I think it needs to be a value, `Object::set` invokes `Object::get_own_property` with the `&Value` and `get_own_property` behaves differentely for `Value::String` and `Value::Symbol`, invoking `self.properties()` or `self.symbol_properties()` respectively.

One option would be to have a `PropertyKey` enum, that would either be a `String` (or equivalent `Box<str>` or a `Symbol`, and can be created from a `Value`, or from a `String` (or anything `Into<String>`). This way, we can use the interpreter in the conversion, and then the `Object` doesn't need the interpreter.

RageKnify:
I followed @Razican proposal for a `PropertyKey` enum, and propagated the change for functions that dealt with those. There were a lot of tests that needed updating, some also lead to bugs in places that were using `Value::to_string` where `Interpreter::to_string` was the appropriate function.
I managed to fix all but 1, `builtins::string::tests::replace_with_function`, I'm especially confused because it seems like I didn't change anything directly related to it. But for some reason the call into the `RegEx` library returns `None` when `Some` is expected.
I'll look at it again tomorrow, but I was hoping you'd have some idea...

HalidOdat:
> I followed @Razican's proposal for a `PropertyKey` enum, and propagated the change for functions that dealt with those.

This is definitely a step in the right direction :)

I don't we need a `.set_str_field`, why not just have `set_field` with the following signiture:
```rust
fn set_field(&mut self, field: F, value: V)
where
	F: Into<PropertyKey>,
	V: Into<Value>
{
	// ...
}
```


there are only a small amount of places where we call `set_field` with `Value` as field (`GetField`) and we call `.to_property_key()` this is done in the spec so we should do the same in this case :)
And also since we take a `PropertyKey` directly we don't need the interpreter anymore.

HalidOdat:
@RageKnify I did the suggested changes here, https://github.com/RageKnify/boa/pull/1

RageKnify:
Sorry, forgot to look at my notifications and tried doing what was done already. 
I just left a comment to remove the `dbg!()` statements, they were just for you to find the bug, correct? @HalidOdat

RageKnify:
I'll do some final touch ups and add some tests to increase coverage.

RageKnify:
One thing I wanted to double check, when you applied your proposed changes you changed `Value::set_field` to return the `field.into<Value>()` rather than the `value` with which it was set @HalidOdat, on `master` it returns the `value`: https://github.com/boa-dev/boa/blob/5d4d8fe79446e7b18bd9334995706eba537333f4/boa/src/builtins/value/mod.rs#L658

Can I go ahead and set it to return the value while using a clone of the value as the argument for `Object::set` like it is in `master`?

HalidOdat:
> Can I go ahead and set it to return the value while using a clone of the value as the argument for `Object::set` like it is in `master`?

Good catch. Yes. I guess I made a mistake :sweat_smile:

RageKnify:
I wasn't sure how to test the more internal functions, if you think tests should be added and have any ideas let me know.

HalidOdat:
> I wasn't sure how to test the more internal functions, if you think tests should be added and have any ideas let me know.

What exactly do you mean by internal functions? if you mean `display_obj`, I don't think we need to, if it fails the other tests fails so it is covered.

boa/src/builtins/property/mod.rs
```diff
@@ -284,3 +278,75 @@ impl<'a> From<&'a Value> for Property {
         }
     }
 }
+
+#[derive(Trace, Finalize, Debug, Clone)]
+pub enum PropertyKey {
```
HalidOdat, line 283: Some documentation would be nice

boa/src/builtins/property/mod.rs
```diff
@@ -284,3 +278,75 @@ impl<'a> From<&'a Value> for Property {
...
+    String(RcString),
+    Symbol(RcSymbol),
+}
+
+impl From<RcString> for PropertyKey {
+    fn from(string: RcString) -> PropertyKey {
+        PropertyKey::String(string)
+    }
+}
```
HalidOdat, line 292:
```suggestion
impl From<RcString> for PropertyKey {
    #[inline]
    fn from(string: RcString) -> PropertyKey {
        PropertyKey::String(string)
    }
}
```
These are very small functions and annotating it with `#[inline]` helps the compiler out, when optimizing.

boa/src/builtins/value/conversions.rs
```diff
@@ -44,6 +44,12 @@ impl From<RcString> for Value {
     }
 }
 
+impl From<RcSymbol> for Value {
+    fn from(value: RcSymbol) -> Self {
+        Value::Symbol(value)
+    }
+}
```
HalidOdat, line 51: Here too, anywhere where the function is small it should be `#[inline]`

RageKnify:
@HalidOdat should I inline the conversion functions that match on enums or only the ones without branching?

HalidOdat:
> @HalidOdat should I inline the conversion functions that match on enums or only the ones without branching

Yes. If the function contains one condition/match (small) it is worth inlining than having function call. #[inline] is a hint to the compiler.

HalidOdat:
<details>
<summary>Benchmarks</summary>

| (index) | name | changesDuration | masterDuration | difference |
| ------- | --------------------------------------- | -------------------- | ------------------- | ---------- |
| 0 | Arithmetic operations (Execution) | 368.8±23.77ns | **365.7±19.90ns** | +1.0 |
| 1 | Arithmetic operations (Full) | 176.4±10.00µs | 176.6±8.54µs | 0.0 |
| 2 | Array access (Execution) | **8.6±0.37µs** | 14.3±0.72µs | -40 |
| 3 | Array access (Full) | **190.3±10.93µs** | 201.0±10.18µs | -5.7 |
| 4 | Array creation (Execution) | **3.2±0.13ms** | 3.7±0.15ms | -13 |
| 5 | Array creation (Full) | **3.5±0.14ms** | 4.1±0.14ms | -14 |
| 6 | Array pop (Execution) | **1146.1±68.21µs** | 1451.0±62.37µs | -21 |
| 7 | Array pop (Full) | **1460.1±56.31µs** | 1818.3±103.17µs | -20 |
| 8 | Boolean Object Access (Execution) | **4.3±0.29µs** | 4.4±0.25µs | -2.9 |
| 9 | Boolean Object Access (Full) | **187.7±8.66µs** | 192.6±12.48µs | -2.9 |
| 10 | Create Realm | 164.3±35.47µs | **159.0±14.39µs** | +3.0 |
| 11 | Dynamic Object Property Access (Execution) | **5.5±0.27µs** | 7.0±0.38µs | -21 |
| 12 | Dynamic Object Property Access (Full) | **187.6±9.25µs** | 196.0±9.41µs | -3.8 |
| 13 | Expression (Lexer) | 2.4±0.16µs | 2.4±0.11µs | 0.0 |
| 14 | Expression (Parser) | 5.4±0.27µs | **5.2±0.27µs** | +4.0 |
| 15 | Fibonacci (Execution) | **1042.6±57.07µs** | 1097.6±64.93µs | -4.8 |
| 16 | Fibonacci (Full) | **1265.7±71.49µs** | 1314.8±74.64µs | -3.8 |
| 17 | For loop (Execution) | **23.6±1.36µs** | 23.8±1.22µs | -0.99 |
| 18 | For loop (Full) | **204.9±11.34µs** | 206.4±9.65µs | -0.99 |
| 19 | For loop (Lexer) | **5.3±0.26µs** | 5.3±0.27µs | -0.99 |

</details>

As expected there is a nice increase in performance `~40` in array access and a `~20` dynamic object property access this is because we don't take the take a `Value` as the key but `PropertyKey` and we consume the string index in array instead of cloning :)

HalidOdat:
This is good to go. Lets merge this

Thanks for the contribution @RageKnify :)

Issue #365, opened 2020-05-02
Implement for loop execution


Razican:
The parser generates [`Node::ForLoop`][for_node] when it encounters a for loop, but this is not implemented in the executor, so it panics.

This can be implemented by adding another match arm to the [`run()`][exec_run] function in the `exec/mod.rs` file.

[for_node]: https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/ast/node.rs#L236-L252
[exec_run]: https://github.com/jasonwilliams/boa/blob/master/boa/src/exec/mod.rs#L73-L576

akryvomaz:
Hi, I would like to implement it

Razican:
> Hi, I'd like to implement it

Sure! Let us know if you need any mentoring :)

PR #374, opened 2020-05-06
Implement for loop

akryvomaz:
#365

### DIFF
--- a/boa/src/exec/tests.rs
+++ b/boa/src/exec/tests.rs
@@ -305,6 +305,44 @@ fn test_do_while_post_inc() {
 }
 
 #[test]
+fn test_for_loop() {
+    let simple = r#"
+        let a = 0;
+        for (let i = 0; i < 10; i = i + 1) {
+            a = a + i;
+        }
+
+        a
+        "#;
+    assert_eq!(exec(simple), String::from("45"));
+
+    let without_init_and_inc_step = r#"
+        let a = 0;
+        let i = 0;
+        for (;i < 10;) {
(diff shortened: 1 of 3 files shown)


akryvomaz:
@Razican do you mean something like this for initialization test?
```
let a = ''
for (let b = 'hello, ', c = 'world'; a.length < 1;) {
    a = b + c
}
assert_eq!(a, 'hello, world')
```
Do you have an idea for a good step test?

Razican:
> @Razican do you mean something like this for initialization test?
> ```
> let a = ''
> for (let b = 'hello, ', c = 'world'; a.length < 1;) {
>     a = b + c
> }
> assert_eq!(a, 'hello, world')
> ```
> Do you have an idea for a good step test?

Something we could do for both would be something like this:

```javascript
let a = "hello";
let b = '';
for (var i = 0; i < a.length; i = i + 1) {
    b = b + a[i];
}
b;
```
Which should equal to `"hello"`. Note that `i++` is not implemented yet.

boa/src/exec/tests.rs
```diff
@@ -307,14 +307,15 @@ fn test_do_while_post_inc() {
...
-        for (let i = 0; i < 10; i = i + 1) {
-            a = a + i;
+        const a = ['h', 'e', 'l', 'l', 'o'];
+        let b = '';
+        for (let i = 0; i < a.length; i = i + 1) {
+            b = b + a[i];
         }
-
-        a
+
```
Razican, line 315:
```suggestion

```

Issue #321, opened 2020-04-16
Implement unary operator execution


Razican:
I just noticed that we haven't implemented all unary operators in execution.

 - [x] `UnaryOp::IncrementPost` (`a++`)
 - [x] `UnaryOp::IncrementPre` (`++a`)
 - [x] `UnaryOp::DecrementPost` (`a--`)
 - [x] `UnaryOp::DecrementPre` (`--a`)
 - [x] `UnaryOp::Minus` (`-a`)
 - [x] `UnaryOp::Plus` (`+a`)
 - [x] `UnaryOp::Not` (`!a`)
 - [x] `UnaryOp::Tilde` (`~a`)
 - [ ] `UnaryOp::Delete` (`delete`)
 - [ ] `UnaryOp::Void` (`void`)

This is implemented in `boa/src/exec/mod.rs`, around line [296][lines].

I think that the "pre" increment/decrement operators, along with the `typeof` and `void` operators should be easy to implement, and good for newcomers to Boa.

The "post" increment/decrement operators will be challenging, as we currently always fully evaluate an AST node before continuing, and these need to be evaluated afterwards. The `delete` operator will be challenging too, I think.

After this is done, in #304 I have implemented some tests that are currently ignored, and we can probably start running them.

[lines]: https://github.com/jasonwilliams/boa/blob/master/boa/src/exec/mod.rs#L296-L314

jasonwilliams:
Did you want to close https://github.com/jasonwilliams/boa/issues/118 in favour of this?

Razican:
> Did you want to close #118 in favour of this?

Actually, I didn't notice that one, but I think there are two things in there. That issue showed that the parsing was broken. That has been solved. Now the issue is not a bug, it's that we don't have that implemented.

So I would say we should close the issue, as the bug is already fixed, and that we should implement the execution.

I already did a small change in #304, and changed the `unreachable!()` for an `unimplemented!()`. It should give a better error message.

akryvomaz:
I think it makes sense to remove `UnaryOp::TypeOf` form the list because of the #359

Razican:
> I think it makes sense to remove `UnaryOp::TypeOf` form the list because of the #359

Yep, let's remove it!

PR #380, opened 2020-05-07
Implement unary increment and decrement

akryvomaz:
#321

### DIFF
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -357,6 +357,20 @@ impl Executor for Interpreter {
                 Ok(match *op {
                     UnaryOp::Minus => to_value(-v_a.to_num()),
                     UnaryOp::Plus => to_value(v_a.to_num()),
+                    UnaryOp::IncrementPost => {
+                        self.set_value(a.deref(), to_value(v_a.to_num() + 1.0))?;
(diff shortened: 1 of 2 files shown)


boa/src/exec/mod.rs
```diff
@@ -357,6 +357,20 @@ impl Executor for Interpreter {
                 Ok(match *op {
                     UnaryOp::Minus => to_value(-v_a.to_num()),
                     UnaryOp::Plus => to_value(v_a.to_num()),
+                    UnaryOp::IncrementPost => {
+                        self.set_value(a.deref(), to_value(v_a.to_num() + 1.0))?;
```
Razican, line 361: Can we use `*a` instead of `a.deref()` here?
Also, can we pass the `op` without the `ref` in the match? I think `UnaryOp` implements `Copy`, and if not, it probably should :)

akryvomaz: `a` here is `&Box<Node>`, so instead of `a.deref()` it will be `&*a`. `deref` looks cleaner to me, what do you think?

Razican: Ok, perfect to me :)

Razican:
This needs a rebase after #372 :)

PR #381, opened 2020-05-07
Implement toString

dvtkrlbs:


dvtkrlbs:
I implemented the initial version but the problem is i made some mistake while translating c++ to c (to further use c2rust) and in some cases produces wrong values for example `Number(90).toString(16)` should return `'5a'` but instead it returns `'50'`

boa/src/builtins/number/mod.rs
```diff
@@ -159,6 +159,145 @@ pub fn to_precision(this: &mut Value, args: &[Value], _ctx: &mut Interpreter) ->
...
+        if integer <= 0f64 {
+            break;
+        }
+    }
+    // Add sign and terminate string.
+    if negative {
+        integer_cursor -= 1;
+        buffer[integer_cursor] = b'-';
+    }
+    assert!(fraction_cursor < BUF_SIZE);
```
dvtkrlbs, line 296: I think the real bug is here but i am not so sure. The problem is this part is broken from the very beginning.

boa/src/builtins/number/tests.rs
```diff
@@ -177,12 +178,38 @@ fn to_string() {
...
+    eprintln!("{}", forward(&mut engine, init));
+    let zero_with_radix = forward(&mut engine, "zero_with_radix");
+    let num_smaller_than_radix = forward(&mut engine, "num_smaller_than_radix");
+    let num_with_radix = forward(&mut engine, "num_with_radix");
+    let fraction_with_radix = forward(&mut engine, "fraction_with_radix");
+
+    assert_eq!(zero_with_radix, String::from("0"));
+    assert_eq!(num_smaller_than_radix, String::from("9"));
+    // assert_eq!(num_with_radix, String::from("5a"));
+    assert_eq!(fraction_with_radix, String::from("5a.e66666666668"));
```
dvtkrlbs, line 212: This test shows that the issue is in the integer part.

boa/src/builtins/number/mod.rs
```diff
@@ -159,6 +159,145 @@ pub fn to_precision(this: &mut Value, args: &[Value], _ctx: &mut Interpreter) ->
...
+    // let mut exp = 0;
+    // unsafe {
+    //     frexp(integer / radix as f64, &mut exp);
+    // }
+
+    while frexp(integer / radix as f64).1 > 0 {
+        integer /= radix as f64;
+        integer_cursor -= 1;
+        buffer[integer_cursor] = b'0';
+    }
```
dvtkrlbs, line 280: The problematic part is here probably my frexp implementation fails here

boa/src/builtins/number/mod.rs
```diff
@@ -159,6 +159,145 @@ pub fn to_precision(this: &mut Value, args: &[Value], _ctx: &mut Interpreter) ->
...
+    } else {
+        let lg = s.abs().log2();
+        let x = (lg - lg.floor() - 1.0).exp2();
+        let exp = lg.floor() + 1.0;
+        (s.signum() * x, exp as isize)
+    }
+}
+
+// https://chromium.googlesource.com/v8/v8/+/refs/heads/master/src/numbers/conversions.cc#1230
+pub fn num_to_string(mut value: f64, radix: u32) -> String {
```
Razican, line 194: Since `radix` is always between 2 and 36, we can just pass a `u8`.
```suggestion
pub fn num_to_string(mut value: f64, radix: u8) -> String {
```

boa/src/builtins/number/mod.rs
```diff
@@ -159,6 +159,145 @@ pub fn to_precision(this: &mut Value, args: &[Value], _ctx: &mut Interpreter) ->
...
+    assert!(radix >= 2);
+    assert!(radix <= 36);
+    assert!(value.is_finite());
```
Razican, line 197: I think these should be debug asserts, since we should already be checking before sending it here:
```suggestion
    debug_assert!(radix >= 2, "radix too small");
    debug_assert!(radix <= 36, "radix too big");
    debug_assert!(value.is_finite(), "non-finite number found");
```

boa/src/builtins/number/mod.rs
```diff
@@ -159,6 +159,145 @@ pub fn to_precision(this: &mut Value, args: &[Value], _ctx: &mut Interpreter) ->
...
+
+    // We only compute fractional digits up to the input double's precision.
+    let mut delta = 0.5 * (next_after(value, f64::MAX) - value);
+    delta = next_after(0.0, f64::MAX).max(delta);
+    assert!(delta > 0.0);
+    if fraction >= delta {
+        // Insert decimal point.
+        let fresh0 = fraction_cursor;
+        fraction_cursor += 1;
+        buffer[fresh0] = b'.';
```
Razican, line 226: In Rust, this indexing is **very** slow, since it will do bounds checking every time it accesses the buffer. I think we can somehow iterate through it using iterators instead of direct access, but then, the carry-overs can be problematic. Maybe @HalidOdat has some ideas on how to improve this performance.

boa/src/builtins/number/mod.rs
Razican: We can avoid using this `frexp()` function by using the [`num-traits`](https://crates.io/crates/num-traits) crate. We will probably use it anyways for #358, so it makes sense to use it here. The function we want is [`FloatCore::integer_decode`](https://docs.rs/num-traits/0.2.11/num_traits/float/trait.FloatCore.html#tymethod.integer_decode).

boa/src/builtins/number/mod.rs
```diff
@@ -159,6 +159,145 @@ pub fn to_precision(this: &mut Value, args: &[Value], _ctx: &mut Interpreter) ->
...
+        }
+    }
+    // Compute integer digits. Fill unrepresented digits with zero.
+    //  while (Double(integer / radix).Exponent() > 0) {
+    // let mut exp = 0;
+    // unsafe {
+    //     frexp(integer / radix as f64, &mut exp);
+    // }
+
+    while frexp(integer / radix as f64).1 > 0 {
```
Razican, line 276: Try to use `f64::from(radix)` instead of `radix as f64`, this will make it easier to track errors in the future, and prevents the compilation of invalid conversions. There are multiple occurrences of this.
```suggestion
    while frexp(integer / f64::from(radix)).1 > 0 {
```

One extra option is to generate the `f64::from(radix)` once at the beginning, and re-use it. This should be more performant.

boa/src/builtins/number/mod.rs
Razican: Something that can be done here to at least increase the iteration for the integer part is this:
```rust
let (int_buf, frac_buf) = buffer.split_at_mut(BUF_SIZE / 2);
```

Then, in the integer iteration:
```rust
    let mut int_iter = int_buf.into_iter().rev();
    for c in int_iter {
        let remainder = integer % (radix as f64);
        *c = std::char::from_digit(remainder as u32, radix).unwrap() as u8;
        integer = (integer - remainder) / radix as f64;
        if integer <= 0_f64 {
            break;
        }
    }
    // Add sign and terminate string.
    if negative {
        *int_buf.next().expect("buffer ended") = b'-';
    }
```

This might need some `mut`-s somewhere around, and we need to save the index somehow, maybe by calling `enumerate()` in the iterator, but it should work. Maybe we can do something similar with the fractional part.

boa/src/builtins/number/mod.rs
dvtkrlbs: i left them as assert since these kind of asserts allow llvm to do further optimizations. I dont know if they apply here but for array length checks they eliminate the bound checks from llvm.

dvtkrlbs:
Ok using `FloatCore::integer_decode`actually fixed the bug. 😄

boa/src/builtins/number/mod.rs
dvtkrlbs: We may also be able to replace some of the while loop using `take_while`.

dvtkrlbs: Fraction part is hard to refactor since there is a second pass to the buffer and implementation sometimes increments and sometimes decrements the cursor. Maybe we can do something with `Rev` but it will be complicated.

boa/src/builtins/number/tests.rs
```diff
@@ -177,12 +178,255 @@ fn to_string() {
...
+    assert_eq!(
+        String::from("0.01"),
+        forward(&mut engine, "Number(0.01).toString()")
+    );
+    assert_eq!(
+        String::from("0.0123"),
+        forward(&mut engine, "Number(0.0123).toString()")
+    );
+    // assert_eq!(String::from("111111111111111110000"), forward(&mut engine, "Number(111111111111111111111).toString()"));
+    // assert_eq!(String::from("1.1111111111111111e+21"), forward(&mut engine, "Number(1111111111111111111111).toString()"));
```
dvtkrlbs, line 256: This tests fails i think because rusts default formatting does not match the js spec.

boa/src/builtins/number/mod.rs
dvtkrlbs: Something like bidir_iter might work better but problem is they use slice.get internally so we still pay for the bound checks. We might use unsafe get to not pay for the bound check since we check the bounds anyway.

boa/src/builtins/number/tests.rs
HalidOdat: > This tests fails i think because rusts default formatting does not match the js spec.

Yes. The rust formatter is not JavaScript spec compliant. Even using `format!("{:e}", num)` isn't sufficient.
We probably need too write our own f64 parser, like we did the one with bases. Here is V8 [DoubleToCString](https://chromium.googlesource.com/v8/v8/+/refs/heads/master/src/numbers/conversions.cc#938)

boa/src/builtins/number/mod.rs
Razican: I think it's good enough for now :) we can maybe improve it further in the future.

dvtkrlbs:
I was trying to resolve the conflicts and now it seems every number is interpreted as NaN. I dont it is a issue from the to_string method i just debug the value after calling to_number on it and it just prints NaN.

boa/src/builtins/number/tests.rs
```diff
@@ -177,12 +178,255 @@ fn to_string() {
...
     let exp_string = forward(&mut engine, "exp_string");
     let neg_string = forward(&mut engine, "neg_string");
+    let nan_string = forward(&mut engine, "nan_string");
 
     assert_eq!(default_string, String::from("0"));
     assert_eq!(int_string, String::from("123"));
     assert_eq!(float_string, String::from("1.234"));
     assert_eq!(exp_string, String::from("12000"));
     assert_eq!(neg_string, String::from("-1.2"));
+    assert_eq!(nan_string, String::from("NaN"));
```
Razican, line 188: I think we should make this easier to understand like this:
```suggestion
    assert_eq!(&default_string, "0");
    assert_eq!(&int_string, "123");
    assert_eq!(&float_string, "1.234");
    assert_eq!(&exp_string, "12000");
    assert_eq!(&neg_string, "-1.2");
    assert_eq!(&nan_string, "NaN");
```

I think we should do this in all tests.

dvtkrlbs:
Ok my bad i apparently called to_number on this without calling to_num

boa/src/builtins/number/tests.rs
```diff
@@ -177,12 +178,255 @@ fn to_string() {
...
+    assert_eq!(
+        String::from("0.000001"),
+        forward(&mut engine, "Number(0.000001).toString()")
+    );
+    // assert_eq!(String::from("1e-7"), forward(&mut engine, "Number(0.0000001).toString()"));
+    // assert_eq!(String::from("1.2e-7"), forward(&mut engine, "Number(0.00000012).toString()"));
+    // assert_eq!(String::from("1.23e-7"), forward(&mut engine, "Number(0.000000123).toString()"));
+    // assert_eq!(String::from("1e-8"), forward(&mut engine, "Number(0.00000001).toString()"));
+    // assert_eq!(String::from("1.2e-8"), forward(&mut engine, "Number(0.000000012).toString()"));
+    // assert_eq!(String::from("1.23e-8"), forward(&mut engine, "Number(0.0000000123).toString()"));
```
Razican, line 271: I'm guessing these tests don't pass wither, right? If that's the case, I would isolate them in a different test and add the `#[ignore]` attribute to it, to know that we must fix them in the future.

boa/src/builtins/number/tests.rs
```diff
@@ -177,12 +178,255 @@ fn to_string() {
...
+
+    assert_eq!(
+        String::from("NaN"),
+        forward(&mut engine, "Number(NaN).toString(16)")
```
Razican, line 275: I would change all these to have this syntax:
```suggestion
        "NaN",
        &forward(&mut engine, "Number(NaN).toString(16)")
```

boa/src/builtins/number/tests.rs
dvtkrlbs: I think for comprehensive tests like we may start to implement something like v8's [mjsunit](https://chromium.googlesource.com/v8/v8/+/3.30.3/test/mjsunit/number-tostring.js?autodive=0). It does not need to be something fancy. Just some file discovery and running them on a basic test suite running them. Also it would be cool to continue running these kind of tests even a previous assert fails. It is kinda annoying to rerun tests on every fix than seeing that a following assert fails.

PR #383, opened 2020-05-08
Value refactor

HalidOdat:


boa/src/builtins/value/mod.rs
```diff
@@ -34,11 +35,77 @@ pub use operations::*;
...
+
+    #[inline]
+    pub fn null() -> Self {
+        Self(Gc::new(ValueData::Null))
+    }
+
+    #[inline]
+    pub fn string(value: String) -> Self {
+        Self(Gc::new(ValueData::String(value)))
+    }
```
Razican, line 55: I think we should use `Into<String>` here, so that we can avoid all those `to_string()` when calling it, and we can just write `Value::string("hello")`. This will be a no-op for a `String`.

```suggestion
    pub fn string<S>(value: S) -> Self where S: Into<String> {
        Self(Gc::new(ValueData::String(value.into())))
    }
```

boa/src/builtins/value/mod.rs
```diff
@@ -69,10 +136,10 @@ impl ValueData {
             let obj_proto = glob.get_field_slice("Object").get_field_slice(PROTOTYPE);
 
             let obj = Object::create(obj_proto);
-            Gc::new(Self::Object(Box::new(GcCell::new(obj))))
+            Value(Gc::new(Self::Object(Box::new(GcCell::new(obj)))))
```
Razican, line 139: Should we have a `Value::object()`?

boa/src/builtins/array/mod.rs
```diff
@@ -756,11 +752,11 @@ pub fn find_index(this: &mut Value, args: &[Value], interpreter: &mut Interprete
         let result = interpreter.call(predicate_arg, &mut this_arg, &arguments)?;
 
         if result.is_true() {
-            return Ok(Gc::new(ValueData::Rational(f64::from(i))));
+            return Ok(Value::rational(f64::from(i)));
         }
     }
 
-    Ok(Gc::new(ValueData::Rational(f64::from(-1))))
+    Ok(Value::rational(f64::from(-1)))
```
Razican, line 759:
```suggestion
    Ok(Value::rational(-1_f64))
```

boa/src/builtins/array/mod.rs
```diff
@@ -756,11 +752,11 @@ pub fn find_index(this: &mut Value, args: &[Value], interpreter: &mut Interprete
         let result = interpreter.call(predicate_arg, &mut this_arg, &arguments)?;
 
         if result.is_true() {
-            return Ok(Gc::new(ValueData::Rational(f64::from(i))));
+            return Ok(Value::rational(f64::from(i)));
```
Razican, line 755: Maybe it makes sense using `Into<f64>` for `Value::rational()` too.

HalidOdat:
> The next big thing to do is to remove `FromValue` and `ToValue` traits and implement `TryFrom` and `From` traits.

Yeah! I already have some progress locally for `From` trait still needs some work though

boa/src/builtins/value/mod.rs
HalidOdat: Probably should add one :)

Razican:
Wow, performance gains, especially in execution, are impressive!

HalidOdat:
> Wow, performance gains, especially in execution, are impressive!

I think it's because I removed a **lot** of unnecessary `cloning`

HalidOdat:
I think this is ready merge now :)

I plan to do more cleanup/refactoring of builtin methods in another PR.

boa/src/builtins/value/mod.rs
```diff
@@ -34,11 +35,123 @@ pub use operations::*;
...
+
+    /// Creates a new value from a `JSONValue`.
+    #[inline]
+    pub fn from_json(json: JSONValue) -> Self {
+        Self(Gc::new(ValueData::from_json(json)))
+    }
+
+    pub fn as_num_to_power(&self, other: Self) -> Self {
+        Self::rational(self.to_number().powf(other.to_number()))
+    }
```
Razican, line 116:
```suggestion
    /// Helper function to convert the `Value` to a number and compute its power.
    pub fn as_num_to_power(&self, other: Self) -> Self {
        Self::rational(self.to_number().powf(other.to_number()))
    }
```

boa/src/builtins/value/mod.rs
```diff
@@ -34,11 +35,123 @@ pub use operations::*;
...
+    #[inline]
+    pub fn data(&self) -> &ValueData {
+        &*self.0
+    }
+
+    /// Creates a new value from a `JSONValue`.
+    #[inline]
+    pub fn from_json(json: JSONValue) -> Self {
+        Self(Gc::new(ValueData::from_json(json)))
+    }
```
Razican, line 112: Shouldn't this be a `From<JSONValue>` iplementation?
Maybe `ValueData` should also implement `From<JSONValue>`.

boa/src/builtins/value/mod.rs
HalidOdat: Thanks!

boa/src/builtins/value/mod.rs
HalidOdat: Value implements From<JSONValue>. Should ValueData implemented too?

Razican: Hmmm if it implements `From<JSONValue>`, why do we need this function? `Value::from(json)` and `Value::from_josn(json)` would be duplicates, right?

I think it makes sense implementing it in `ValueData` too, that way `From<JSONValue>` could be like this:

```rust
fn from(json: JSONValue) -> Self {
	Self(Gc::new(json.into()))
}
```

HalidOdat: `Value::from` uses `Value::from_json` to convert the value, you are right it's duplicate function with the same job.
I think we should remove `Value::from_json` and only have one function.

Razican: I think so too :)

boa/src/builtins/value/mod.rs
```diff
@@ -23,6 +23,7 @@ use std::{
     ops::{Add, BitAnd, BitOr, BitXor, Deref, DerefMut, Div, Mul, Not, Rem, Shl, Shr, Sub},
     str::FromStr,
 };
+// use std::borrow::{Borrow, BorrowMut};
```
Razican, line 26:
```suggestion

```

Issue #363, opened 2020-05-02
How to deal with Garbage Collection


jasonwilliams:
## Current state of play
We currently use [GC](https://github.com/Manishearth/rust-gc) to wrap objects which need to live as long as something is referencing it. These are all JSValues. To make things easier we wrap all [JS Values](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L39-L60) in a GC and give that the type [`Value`](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L33).

## Why not use RC?
Javascript has a lot of circular references, we could end up with memory leaks because objects will always have a reference count of 1 or more.
Constructors have a property to the prototype, and the prototype object has a property pointing back to the constructor.

More info:
https://doc.rust-lang.org/book/ch15-06-reference-cycles.html

### Environments
Environments also need to be wrapped in a Gc because multiple function objects can reference the same environment, any function declaration references the outer env. So declaring more than 1 function in the same scope means  they'll both have the same parent.
If a closure is returned the outer function environment needs to stay alive throughout the duration of the returned function.

![Blank Diagram (1)](https://user-images.githubusercontent.com/936006/80867983-6fac5b80-8c8f-11ea-9112-3257850ef24b.png)

@Razican has some ideas on how to improve Environment GC.

Razican:
Yep, so around 4% of our instructions in the code are due to us using a garbage collector. The GC will once in a while for through all the garbage collected objects and decide if they should be freed or not, which takes long: for structures/enums with members, it has to go through all its mebers one by one.

This can be alleviated firs by only deriving `Trace` in types where a GC'd object might leave, so all `Copy` types should not be derived, but we should implement an empty trace on them, for example. This could be made easier once rust-gc implements https://github.com/Manishearth/rust-gc/issues/87.

Then, environments have a clear hierarchy, so we don't really need a GC. Once the parent environment goes out of the scope, all inner environments no longer need to exist. If we use a GC, they will at some point be freed by the GC, but we don't need to wait for that, this can be reasoned at compile time by Rust by using `Rc` (or `Arc` if we go multithreaded at one point).

In a hierarchical tree, the Rust guide for Rc gives us a nice usage:

> A cycle between Rc pointers will never be deallocated. For this reason, Weak is used to break cycles. For example, a tree could have strong Rc pointers from parent nodes to children, and Weak pointers from children back to their parents.

[...]

PR #387, opened 2020-05-09
Mitigate unsoundness in `InternalState`

Razican:
This comes from https://github.com/boa-dev/boa/issues/363#issuecomment-626170196. It turns out that implementing an empty trace was not safe if the contained type was not `Trace`.

I'm still having issues, though, since we cannot downcast a non `Any` type. Adding `Any` to the `InternalState` trait bounds doesn't help either. Any idea on how to fix this, @HalidOdat or @jasonwilliams ?

This still has the issue that `Regex` does not implement `Trace`, and in the future it could, which would make it unsound again. In any case, I think this is a good compromise for now. In the future we will probably need to change the `Regex` for a different thing, in any case, since it doesn't implement everything in the JavaScript regular expressions.

### DIFF
--- a/boa/src/builtins/object/internal_state.rs
+++ b/boa/src/builtins/object/internal_state.rs
@@ -1,5 +1,6 @@
 //! Implementations for storing normal rust structs inside any object as internal state.
 
+use gc::{unsafe_empty_trace, Finalize, Trace};
 use std::{
     any::Any,
     fmt::{self, Debug},
@@ -7,13 +8,11 @@ use std::{
     rc::Rc,
 };
 
-use gc::{unsafe_empty_trace, Finalize, Trace};
-
 /// Wrapper around `Rc` to implement `Trace` and `Finalize`.
 #[derive(Clone)]
 pub struct InternalStateCell {
     /// The internal state.
-    state: Rc<dyn Any>,
+    state: Rc<dyn InternalState + 'static>,
 }
 
 impl Finalize for InternalStateCell {}
@@ -23,7 +22,7 @@ unsafe impl Trace for InternalStateCell {
 }
 
 impl Deref for InternalStateCell {
-    type Target = dyn Any;
+    type Target = dyn InternalState;
     fn deref(&self) -> &Self::Target {
         Deref::deref(&self.state)
     }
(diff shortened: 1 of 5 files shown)


boa/src/builtins/object/internal_state.rs
```diff
@@ -45,20 +44,22 @@ impl Debug for InternalStateCell {
...
-    pub fn downcast_mut<T: Any + InternalState>(&mut self) -> Option<&mut T> {
-        self.deref_mut().downcast_mut::<T>()
+    pub fn downcast_mut<T: InternalState>(&mut self) -> Option<&mut T> {
+        self.state.downcast_mut::<T>()
     }
 }
 
 /// This trait must be implemented by all structs used for internal state.
-pub trait InternalState: Debug {}
+pub trait InternalState: Debug + Trace + Any {}
```
Razican, line 65: It seems that this is not enough to be able to call `downcast_mut()` on `self.state`. It gives the following error:
```
method not found in `&(dyn builtins::object::internal_state::InternalState + 'static)`
```

Razican: After some refactoring, I'm getting this error:
```
error[E0605]: non-primitive cast: `&dyn builtins::object::internal_state::InternalState` as `&(dyn std::any::Any + 'static)`
  --> boa/src/builtins/object/internal_state.rs:44:9
   |
44 |         (state as &dyn Any).downcast_ref::<T>()
   |         ^^^^^^^^^^^^^^^^^^^
   |
   = note: an `as` expression can only be used to convert between primitive types. Consider using the `From` trait

error[E0605]: non-primitive cast: `&mut (dyn builtins::object::internal_state::InternalState + 'static)` as `&mut (dyn std::any::Any + 'static)`
  --> boa/src/builtins/object/internal_state.rs:51:9
   |
51 |         (state as &mut dyn Any).downcast_mut::<T>()
   |         ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: an `as` expression can only be used to convert between primitive types. Consider using the `From` trait

```

HalidOdat:
Superseded by #626

PR #388, opened 2020-05-09
Implement unary void, delete operators

### LINKED ISSUE #321: Implement unary operator execution

akryvomaz:
This Pull Request fixes/closes #321 

It changes the following:
 - Add implementation for unary void operator
 - Add implementation for unary delete operator

### DIFF
--- a/boa/src/exec/tests.rs
+++ b/boa/src/exec/tests.rs
@@ -434,6 +434,65 @@ fn unary_post() {
     assert_eq!(&exec(execs_after_dec), "true");
 }
 
+#[test]
+fn unary_void() {
+    let void_should_return_undefined = r#"
+        const a = 0;
+        void a;
+    "#;
+    assert_eq!(&exec(void_should_return_undefined), "undefined");
+
+    let void_invocation = r#"
+        let a = 0;
+        const test = () => a = 42;
+        const b = void test() + '';
+        a + b
+    "#;
+    assert_eq!(&exec(void_invocation), "42undefined");
+}
+
+#[test]
+fn unary_delete() {
+    let delete_var = r#"
+        let a = 5;
+        const b = delete a + '';
+        a + b
+    "#;
+    assert_eq!(&exec(const_delete), "5false");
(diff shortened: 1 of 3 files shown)


boa/src/exec/tests.rs
```diff
@@ -486,11 +486,6 @@ fn unary_delete() {
         delete a
     "#;
     assert_eq!(&exec(delete_object), "false");
-
-    let delete_recursive = r#"
-        delete delete delete 1
-    "#;
-    assert_eq!(&exec(delete_object), "true");
```
Razican: Hmmm why doesn't this work?

akryvomaz: The handler for `Node::Const` was missing

PR #389, opened 2020-05-09
Added `BindingIdentifier` parsing.

Razican:
This adds `BindingIdentifier` parsing ([spec](https://tc39.es/ecma262/#prod-BindingIdentifier)). This allows creating bindings with `yield` and `await` as identifiers.

I added some tests. Note that this should only be possible in non-strict mode, but we do not currently support strict/non-strict modes. That's something we should probably do in the future.

I was a bit unsure on what did `IdentifierReference` and `LabelIdentifier` productions meant with those `[~Yield]` and `[~Await]`. Do they work the same way as `BindingIdentifier`?

I also improved readability of some tests.

### DIFF
--- a/boa/src/exec/tests.rs
+++ b/boa/src/exec/tests.rs
@@ -10,9 +10,7 @@ fn empty_let_decl_undefined() {
         a == undefined;
         "#;
 
-    let pass = String::from("true");
-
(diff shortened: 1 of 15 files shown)


Razican:
Note that the real benchmark is the first one xD this is ready for review.

Issue #376, opened 2020-05-07
Hoisting variables


hello2dj:
now not founding Identifier will always be undefined, for `function` or  `var` this is wrong.

[moreinfo](https://www.ecma-international.org/ecma-262/10.0/index.html#prod-HoistableDeclaration)

example:
```
// v8
console.log(A); // [Function: A]
function A() {}
```

```
// boa
console.log(A); // undefined
function A() {}
```

Razican:
This seems to be related to the fact that we execute nodes in order, instead of having a special category for this. These functions should be evaluated first, and then start evaluating the rest of statements.

jasonwilliams:
I don't think v0.9.0 is realistic for this, for hoisting to work would require a big refactor of how execution works, including having some pre-parsing step.

We're no where near that point yet.

Razican:
> I don't think v0.9.0 is realistic for this, for hoisting to work would require a big refactor of how execution works, including having some pre-parsing step.
> 
> We're no where near that point yet.

Actually, this seems to be already working in #392.

jasonwilliams:
oh ok interesting.
Should we be doing something more closely related to https://v8.dev/blog/preparser#variable-allocation? is reordering sound?

Razican:
> Should we be doing something more closely related to https://v8.dev/blog/preparser#variable-allocation?

Definitelly, but I think it's a reasonable workaround in the meantime.

> is reordering sound?
It should be, it's  slower, as we will be doing a sort, but it's always maintaining ordering, so it's sound as far as I can tell.

Issue #377, opened 2020-05-07
Implement exception handling


Razican:
We currently already parse `try/catch/finally` and `throw` (even if we have a small issue in #375). and most of our functions already return a `ResultValue`.

We should therefore add support for exceptions by making sure that `ResultValue` returns an exception if it returns an `Err` variant, and then implement the execution of try/catch/finally and `throw`.

Issue #366, opened 2020-05-02
Divide the `run()` function in smaller portions


Razican:
Currently, the `run()` function is around 500 lines long. This is going to quickly become unmaintainable, so we need to divide this.

The first thing that comes to my mind to improve this is to just create new function for each `match` arm, but as we saw in the parser, we can probably do this better and easier to understand.

One option is to have an `Executable` trait, as I proposed in #304. This trait would just be as simple as this:

```rust
trait Executable {
	fn run<E>(interpreter: &mut E) -> ResultValue where E: Executor;
}
```

Then, I think that we should change the `Node` enum to be just a `Box<dyn Executable>`. And each `Node` type should be a struct where we could use `Box` to box dynamic types, or we could use compile-time generics to have less allocations (but longer compile times).

Then we implement this trait individually for each of the node types, which is easily extensible, and we reduce the `run()` function to:

```rust
fn run(&mut self, node: &Node) -> ResultValue {
	node.run(&mut self)
}
```

Tracking:

 - [x] `Node::ArrayDecl`
 - [x] `Node::ArrowFunctionDecl`
 - [x] `Node::Assign`
 - [x] `Node::BinOp`
 - [x] `Node::Block`
 - [x] `Node::Break`
 - [x] `Node::Call`
 - [x] `Node::ConditionalOp`
 - [x] `Node::Const`
 - [x] `Node::ConstDeclList`
 - [x] `Node::Continue`
 - [x] `Node::DoWhileLoop`
 - [x] `Node::FunctionDecl`
 - [x] `Node::FunctionExpr`
 - [x] `Node::GetConstField`
 - [x] `Node::GetField`
 - [x] `Node::ForLoop`
 - [x] `Node::If`
 - [x] `Node::LetDeclList`
 - [x] `Node::Identifier`
 - [x] `Node::New`
 - [x] `Node::Object`
 - [x] `Node::Return`
 - [x] `Node::Switch`
 - [x] `Node::Spread`
 - [x] `StatementList`
 - [x] `Node::Throw`
 - [x] `Node::Try`
 - [x] `Node::This`
 - [x] `Node::UnaryOp`
 - [x] `Node::VarDeclList`
 - [x] `Node::WhileLoop`

HalidOdat:
Hmmm. I think doing this at compile time is better than having to do it at runtime.
Using `Box<dyn Trait>` has to allocate and use dynamic dispatch (`vtable`) which would worsen the execution time.

[...]

Issue #336, opened 2020-04-18
Optimize the `TokenKind` enum


Razican:
As `maciejh` mentioned on [Reddit][reddit_comment], the `TokenKind` enumeration is too big for what it contains. This will be partly solved by string interning (#279), but it can be further optimized by replacing the regex flags by a more optimized structure.

The comment:
> Your `TokenKind` enum is 7 words wide (56 bytes on 64bit arch), you can easily drop this by 3 words if you turn Regex flags into an inline buffer (you can probably manage with `[u8; 7]`, one byte for length, 6 for content). Edit: shave another word by turning `String`s into `Box<str>`s.

[reddit_comment]: https://www.reddit.com/r/rust/comments/g3m5i7/boa_releases_v07_with_a_much_faster_parser/fnsng09

PR #392, opened 2020-05-11
Execution and Node modularization

Razican:
The idea behind this work is to divide the `exec` and `ast::node` modules in smaller code pieces. I also want to solve #376 in the process, and maybe #377.

This is the progress as of now:

 - [x] `Node::ArrayDecl`
 - [x] `Node::ArrowFunctionDecl`
 - [x] `Node::Assign`
 - [x] `Node::BinOp`
 - [x] `Node::Block`
 - [ ] `Node::Break`
 - [x] `Node::Call`
 - [ ] `Node::ConditionalOp`
 - [x] `Node::Const`
 - [x] `Node::ConstDecl` => `Node::ConstDeclList`
 - [ ] `Node::Continue`
 - [ ] `Node::DoWhileLoop`
 - [x] `Node::FunctionDecl`
 - [x] `Node::FunctionExpr`
 - [ ] `Node::GetConstField`
 - [ ] `Node::GetField`
 - [x] `Node::ForLoop`
 - [ ] `Node::If`
 - [x] `Node::LetDecl` => `Node::LetDeclList`
 - [x] `Node::Local` => `Node::Identifier`
 - [x] `Node::New`
 - [ ] `Node::Object`
 - [ ] `Node::Return`
 - [ ] `Node::Switch`
 - [ ] `Node::Spread`
 - [x] `Node::StatementList` _(no longer a node)_
 - [ ] `Node::Throw`
 - [x] `Node::Try`
 - [x] `Node::This` _(no change)_
 - [x] `Node::UnaryOp`
 - [x] `Node::VarDecl` => `Node::VarDeclList`
 - [ ] `Node::WhileLoop`

I also added the `Executable` trait, and I'm modifying the `Executor` trait in order for external libraries to provide their own executors. We could also just remove the `Executor` trait if we don't want to offer this option.


Razican:
A small update on the changes I made:
I removed the `StatementList` node for now, and simplified some nodes to have less indirections.

This has added a bit of extra complexity in the parser, which brings the times up a bit, but they are clearly minor compared to the huge gains in execution. This should improve further as I continue modularizing more nodes.

I also removed the `Executor` trait for now. I don't think it makes sense for other crates to create their own Boa interpreters. It was almost impossible to use it before that way anyways. This has simplified the execution and reduces compile times.

I also added a preliminary version of hoisting, but it's not 100% spec compliant yet.

Razican:
Some more changes:
 - Tokens now have a `Span` instead of a `Position`. This should make it easier to implement #300, or even better diagnostics of errors if we get into that.
 - Positions and spans are now properly checked on creation: no more positions with 0 as the column or line number, no more spans that end before they start. This brought up **many** errors with token positions that should now be fixed.
 - Positions use now `NonZeroU32` for column and line numbers instead of `u32`. This means that a whole position now occupies only 64 bits, instead of 128, and enter in a word in most recent processors. It also brings some optimisations if we use `Option<Position>`. This should make a `Token` a bit smaller, and thus easier to pass around, and should give some performance optimisations.
 - Constants no longer have a `String` if they are of type `String`, they will have a `Box<str>`, this gives us some space optimisation in both heap an stack. #279 should help with this too, by making all tokens `Copy`.
 - The helper structure used to lex numbers is now smaller and `Copy`, which should improve performance in this area.
 - Hoisting is now 100% spec compliant (!), at least with the tests I did.
 - `Node::Local` is now `Node::Identifier`. I thought this name made more sense, and it's the name V8 uses.
 - All non-mutable `String`-s have been replaced by `Box<str>` in `Node`. This will reduce the size of a `Node`.

Still many things to do, but this modularization is already showing big performance improvements. In the future, I would like to modify the `Assign` node and others to allow only for 3 types of nodes in the left hand side. This would avoid one boxing and therefore improve performance.

Razican:
@jasonwilliams it seems that after the rebase to the `this` implementation, I have a couple of tests failing, but I'm having a bit of trouble finding out why. Could you check and see if you can find the issue?

jasonwilliams:
construct isn't working properly, its returning an object but now the fields `this.a ='a'` are not there. 

Some js to try:
```js
function Foo() {
  this.a = "a";
}
var bar = new Foo();
bar.a;
```

You can inspect this https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/function/mod.rs#L268 and se it has no properties set, where as it did before.

boa/src/exec/statement_list.rs
```diff
@@ -0,0 +1,37 @@
...
+};
+
+impl Executable for StatementList {
+    fn run(&self, interpreter: &mut Interpreter) -> ResultValue {
+        {
+            let env = &mut interpreter.realm_mut().environment;
+            env.push(new_declarative_environment(Some(
+                env.get_current_environment_ref().clone(),
+            )));
+        }
```
jasonwilliams, line 17: This is most likely the problem, you shouldn't be making a new environment here.
The function already creates a new environment then immediately calls this, which means we're 2 layers deep instead of 1.

If we're inside a block then that will be Node::Block, however Node::Call or Node::Construct will create a function environment then call StatementList.

Seeing that `this` comes from the inner most environment, the `this` value is lost.

boa/src/builtins/function/mod.rs
```diff
@@ -260,7 +257,7 @@ impl Function {
 
                 // Call body should be set before reaching here
                 let _ = match &self.body {
-                    FunctionBody::Ordinary(ref body) => interpreter.run(body),
+                    FunctionBody::Ordinary(ref body) => body.run(interpreter),
```
jasonwilliams, line 260: I new environment was already created with `this` set.
body.run calls StatementList which sets a `declarativeEnvironment` again, so `this` is lost.

Razican:
> **You have 2 options here:**
> 
>  1. The quick fix for now is to not push/pop an environment inside of `StatementList`, however in future declarative environments will need to  get `this` from a parent environment.
> 
>  2. If we want to fix this properly we need to check if the environment has a `this` binding, and if not go one level up.
>        [tc39.es/ecma262/#sec-declarative-environment-records-hasthisbinding](https://tc39.es/ecma262/#sec-declarative-environment-records-hasthisbinding)
> 
> 
> declarative environment records [don't have a `this` binding](https://github.com/jasonwilliams/boa/blob/master/boa/src/environment/declarative_environment_record.rs#L148-L150) [](/jasonwilliams/boa/blob/HEAD@{2020-05-19T10:16:22Z}/boa/src/environment/declarative_environment_record.rs#L148-L150)

I implemented the first option for now, but about the second option, should we only go one level up, or should we go as many levels as needed up to the function environment? I guess it's the second one, right?

> **Open question**
> Do statementLists needs to create an env?
> They're either called by functions (which have an env) or global (which also sets up an env), so is it redundant?

About this, I think about `StatementList` as just a list of statements, so no semantic meaning to it, just `&[Node]` so I don't think we need any environment to this. Environments should be created by `Block` or `Function` nodes, I believe.

jasonwilliams:
> I implemented the first option for now, but about the second option, should we only go one level up, or should we go as many levels as needed up to the function environment? I guess it's the second one, right?

Just like with any variable binding, it should just keep going until there is a `this` value.
So there should be a loop which keeps traversing up the tree until either `get_this_binding()` is true or `parent_env` is `None`

Razican:
> Just like with any variable binding, it should just keep going until there is a `this` value.
> So there should be a loop which keeps traversing up the tree until either `get_this_binding()` is true or `parent_env` is `None`

This seems to pass all the tests, is it correct? [https://github.com/jasonwilliams/boa/pull/392/commits/c8209f16bd330a463a45b3e6ecae569df68bc2b2 ](https://github.com/jasonwilliams/boa/pull/392/commits/c8209f16bd330a463a45b3e6ecae569df68bc2b2)

HalidOdat:
> This seems to pass all the tests, is it correct? [https://github.com/jasonwilliams/boa/pull/392/commits/c8209f16bd330a463a45b3e6ecae569df68bc2b2 ](https://github.com/jasonwilliams/boa/pull/392/commits/c8209f16bd330a463a45b3e6ecae569df68bc2b2)

Correct me if I'm wrong, the environments are stored like a stack right? (first in, last out)

So would it make more sense to search in reverse, so instead of `find` use `rfind`. It's a potential bug with `find` right?

Razican:
> > This seems to pass all the tests, is it correct? [https://github.com/jasonwilliams/boa/pull/392/commits/c8209f16bd330a463a45b3e6ecae569df68bc2b2 ](https://github.com/jasonwilliams/boa/pull/392/commits/c8209f16bd330a463a45b3e6ecae569df68bc2b2)
> 
> Correct me if I'm wrong, the environments are stored like a stack right? (first in, last out)
> 
> So would it make more sense to search in reverse, so instead of `find` use `rfind`. It's a potential bug with `find` right?

Hmmm in thory, calling `environments()` will first bring the current environment, then the parent environments, in reverse order. Still, I think we should use the reverse iterator in the `VecDeque` structure.

> Edit: This edge case does not work!
> 
> Example:
> 
> ```js
> function outer() {
> 	this.outerProperty = 5;
> 	function inner() {
> 		this.innerProperty = 4;
> 	}
> 
> 	let innerInstance = new inner();
> }
> 
> new outer()
> ```
> 
> This returns an instance of `outer` with property of `innerProperty` which is wrong.

Hmmm we should check why is this happening.

Razican:
@jasonwilliams, @HalidOdat I think this is ready for review, in the sense that there are some nodes left, but the architecture is there. I'm starting to see that future future PRs will make merges very difficult, so I think we can land a preliminary version of this, and then modularize the rest of the nodes (and the tests) in another PR.

This PR notably fixes #376 (with a workaround), #336 and #377.

It adds a `Span` to the tokens, so it will be relativelly easy to implement #300. And of course, it does great progress towards #366, I will add the mising nodes to that issue.

Let me know what you think :)

PS: I also modified all static strings to be `Box<str>`, but this required a change in `rust-gc`. We would need for `rust-gc` to publish a new version in crates.io for this to be mergeable. We will need the help of @Manishearth for this :)

jasonwilliams:
Will find some time to take a look at this

> PS: I also modified all static strings to be Box<str>, but this required a change in rust-gc. We would need for rust-gc to publish a new version in crates.io for this to be mergeable. We will need the help of @Manishearth for this :)

So we need that to be merged before this then, is there a link to that PR or issue?

Razican:
> > PS: I also modified all static strings to be Box<str>, but this required a change in rust-gc. We would need for rust-gc to publish a new version in crates.io for this to be mergeable. We will need the help of @Manishearth for this :)
> 
> So we need that to be merged before this then, is there a link to that PR or issue?

It was already merged, we only need a release in crates.io :D

Manishearth:
Published

Razican:
> Published

Great! I updated the dependencies, without the Git dependency, so this is ready to be merged!

PR #396, opened 2020-05-13
Corrected impl of typeof

### LINKED ISSUE #359: TypeOf failing
TypeOf is defined as both a Node:
https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/ast/node.rs#L421

And a UnaryOp Variant:
https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/parser/expression/unary.rs#L58-L60

looking at the spec it seems the former is what we should keep and we should remove it from being a unary variant. 

Right now the variant is being used but there's no implementation on the [exec side](https://github.com/jasonwilliams/boa/blob/master/boa/src/exec/mod.rs#L314)

Make some JS, like this to test.
```
const a = String();
typeof a;
```

I'm not sure we will need https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L631-L651 once Node::TypeOf is being used.

**Spec:**
https://tc39.es/ecma262/#sec-typeof-operator

**Debugging**

RestitutorOrbis:
Corrected impl of typeof to follow specification by using relevant
AST node directly instead of using UnaryOp

This Pull Request fixes #359.

It changes the following:
 - Adjusts implementation of TypeOf to use Node::TypeOf instead of Node::UnaryOp

### DIFF
--- a/boa/src/builtins/value/operations.rs
+++ b/boa/src/builtins/value/operations.rs
@@ -19,6 +19,14 @@ impl Value {
             return number::equals(f64::from(self), f64::from(other));
         }
 
+        //Null has to be handled specially because "typeof null" returns object and if we managed
+        //this without a special case we would compare self and other as if they were actually
+        //objects which unfortunately fails
+        //Specification Link: https://tc39.es/ecma262/#sec-typeof-operator
+        if self.is_null() {
+            return true;
+        }
+
         same_value_non_number(self, other)
     }
 
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -362,11 +362,10 @@ impl Executor for Interpreter {
                         | Node::FunctionExpr(_, _, _)
                         | Node::New(_)
                         | Node::Object(_)
-                        | Node::TypeOf(_)
                         | Node::UnaryOp(_, _) => Value::boolean(true),
                         _ => panic!("SyntaxError: wrong delete argument {}", node),
                     },
-                    _ => unimplemented!(),
+                    UnaryOp::TypeOf => Value::from(v_a.get_type()),
                 })
             }
             Node::BinOp(BinOp::Bit(ref op), ref a, ref b) => {
@@ -548,25 +547,6 @@ impl Executor for Interpreter {
                     self.realm.environment.initialize_binding(&name, val);
                 }
                 Ok(Value::undefined())
-            }
-            Node::TypeOf(ref val_e) => {
-                let val = self.run(val_e)?;
-                Ok(Value::from(match *val {
-                    ValueData::Undefined => "undefined",
-                    ValueData::Symbol(_) => "symbol",
-                    ValueData::Null => "object",
-                    ValueData::Boolean(_) => "boolean",
-                    ValueData::Rational(_) | ValueData::Integer(_) => "number",
-                    ValueData::String(_) => "string",
-                    ValueData::Object(ref o) => {
-                        if o.deref().borrow().is_callable() {
-                            "function"
-                        } else {
-                            "object"
-                        }
-                    }
-                    ValueData::BigInt(_) => "bigint",
-                }))
             }
             Node::StatementList(ref list) => {
                 let mut obj = Value::null();
--- a/boa/src/exec/tests.rs
+++ b/boa/src/exec/tests.rs
@@ -386,6 +386,63 @@ fn unary_pre() {
 }
 
 #[test]
+fn unary_typeof() {
+    let typeof_string = r#"
+        const a = String();
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_string), "string");
+
+    let typeof_int = r#"
+        let a = 5;
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_int), "number");
+
+    let typeof_rational = r#"
+        let a = 0.5;
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_rational), "number");
+
+    let typeof_undefined = r#"
+        let a = undefined;
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_undefined), "undefined");
+
+    let typeof_boolean = r#"
+        let a = true;
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_boolean), "boolean");
+
+    let typeof_null = r#"
+        let a = null;
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_null), "object");
+
+    let typeof_object = r#"
+        let a = {};
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_object), "object");
+
+    let typeof_symbol = r#"
+        let a = Symbol();
+        typeof a;
+    "#;
+    assert_eq!(&exec(typeof_symbol), "symbol");
+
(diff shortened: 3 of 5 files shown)


RestitutorOrbis:
Does this code look alright? I ask because I'm encountering the following bizarre compile error and I'm wondering if there's an alternative route I can pursue:
```bash
error: internal compiler error: src/librustc/ty/steal.rs:34: attempted to read from stolen value

thread 'rustc' panicked at 'Box<Any>', src/librustc_errors/lib.rs:875:9

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.43.0 (4fb7144ed 2020-04-20) running on x86_64-apple-darwin

note: compiler flags: -C debuginfo=2 -C incremental --crate-type cdylib --crate-type lib

note: some of the compiler flags provided by cargo are hidden

thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `LLVMing`,
 right: `Codegenning`', <::std::macros::panic macros>:5:6

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.43.0 (4fb7144ed 2020-04-20) running on x86_64-apple-darwin

note: compiler flags: -C debuginfo=2 -C incremental --crate-type cdylib --crate-type lib

note: some of the compiler flags provided by cargo are hidden

error: aborting due to previous error

error: could not compile `Boa`.
```

Razican:
That is in fact a bug inside the Rust compiler. You should report it, as the error says.

jasonwilliams:
> Hi! thanks for the contribution. After some discussion in #359, it seems that it's best if we implement this as a unary expression and remove the `Node`, then move the `exec` part to the unary expression execution.
> 
> We should also add tests to this.

Yeah sorry about that its may fault i was under the impression it was a Node expression and not a unary one.

boa/src/exec/mod.rs
```diff
@@ -420,10 +420,24 @@ impl Executor for Interpreter {
...
+                        ValueData::String(_) => "string",
+                        ValueData::Object(ref o) => {
+                            if o.deref().borrow().is_callable() {
+                                "function"
+                            } else {
+                                "object"
+                            }
+                        }
+                    }),
                     _ => unimplemented!(),
```
Razican, line 441: It seems that we already implemented all variants, so this is unreachable :D
```suggestion
```

boa/src/exec/tests.rs
```diff
@@ -385,6 +385,62 @@ fn unary_pre() {
...
+    let undefined_type_result = String::from("undefined");
+    let boolean_type_result = String::from("boolean");
+    let object_type_result = String::from("object");
+    let function_type_result = String::from("function");
+    let symbol_type_result = String::from("symbol");
+    let typeof_string = r#"
+        const a = String();
+        typeof a;
+    "#;
+    assert_eq!(exec(typeof_string), string_type_result);
```
Razican, line 401: I think this is more clear if we do it like this:
```suggestion
    assert_eq!(&exec(typeof_string), "string");

```
Also adding a new line after each assert, to separate them.

boa/src/syntax/ast/node.rs
```diff
@@ -469,22 +469,7 @@ pub enum Node {
...
-    ///  - [ECMAScript reference][spec]
-    ///  - [MDN documentation][mdn]
-    ///
-    /// [spec]: https://tc39.es/ecma262/#sec-typeof-operator
-    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof
-    TypeOf(Box<Node>),
-
-    /// The `try...catch` statement marks a block of statements to try and specifies a response
-    /// should an exception be thrown.
+    /// The `try...catch` statement marks a block of statements to try and specifies a response should an exception be thrown.
```
Razican, line 472: Hmmm why was this moved to one line? Rust recommends having 100 characters per line, was it too short?

RestitutorOrbis: My IDE must have done this, I'll revert it

boa/src/exec/mod.rs
```diff
@@ -420,11 +420,24 @@ impl Executor for Interpreter {
...
+                        ValueData::Rational(_) | ValueData::Integer(_) => "number",
+                        ValueData::String(_) => "string",
+                        ValueData::Object(ref o) => {
+                            if o.deref().borrow().is_callable() {
+                                "function"
+                            } else {
+                                "object"
+                            }
+                        }
+                    }),
```
HalidOdat, line 440: Why do we do this here when we can just call `value.get_type()` which returns a `&'static str` it seems like were duping code.

jasonwilliams: I think i mentioned in the original issue that we don't need both https://github.com/jasonwilliams/boa/issues/359#issue-610887558

jasonwilliams: as @HalidOdat said maybe we can just call get_type() from here

boa/src/builtins/value/operations.rs
```diff
@@ -18,6 +18,13 @@ impl Value {
             return number::equals(f64::from(self), f64::from(other));
         }
 
+        //Null has to be handled specially because "typeof null" returns object and if we managed
+        //this without a special case we would compare self and other as if they were actually
+        //objects which unfortunately fails
+        if self.is_null() {
+            return true;
+        }
+
```
Razican, line 27: Can we add a test for this, and a link to the spec?

RestitutorOrbis: Isn't boa/src/builtins/value/tests.rs:58 already a test for this?

RestitutorOrbis:
I'm having difficulty replicating the errors in CI on my local development environments (Windows and Mac OSX). I'm not entirely sure where to proceed from here so I would love to have some suggestions!

jasonwilliams:
> I'm having difficulty replicating the errors in CI on my local development environments (Windows and Mac OSX). I'm not entirely sure where to proceed from here so I would love to have some suggestions!

Looks like you've found a regression introduced by my ["implement this"](https://github.com/jasonwilliams/boa/pull/320) change

I will look at this.

RestitutorOrbis:
> > I'm having difficulty replicating the errors in CI on my local development environments (Windows and Mac OSX). I'm not entirely sure where to proceed from here so I would love to have some suggestions!
> 
> Looks like you've found a regression introduced by my ["implement this"](https://github.com/jasonwilliams/boa/pull/320) change
> 
> I will look at this.

Thanks!

jasonwilliams:
Lets see if https://github.com/jasonwilliams/boa/pull/407 passes and ill merge that, String wasn't defaulting to empty when being called as `const a = String()`

The fact that you made a test for this is great as we may not have captured that before

jasonwilliams:
https://github.com/jasonwilliams/boa/pull/407 merged

boa/src/builtins/value/operations.rs
Razican: Yes! Perfect, rhanks! :)

PR #397, opened 2020-05-13
fix `NaN` is lexed as identifier, not as a number

### LINKED ISSUE #393: `NaN` is lexed as identifier, not as a number
**Describe the bug**
When lexing `NaN`, we don't get a number, we get an identifier.

**To Reproduce**
Run this JavaScript with `cargo run -- -t {file}.js`:
```javascript
let a = NaN;
```

**Expected behavior**
The token for `NaN` should be :
```
Token {
    kind: NumericLiteral(
        Rational(NaN),
    ),
    pos: Position {
        column_number: 9,
        line_number: 1,
    },
},
```

Instead, we are getting:
```
Token {
    kind: Identifier(
        "NaN",
    ),
    pos: Position {
        column_number: 9,
        line_number: 1,
    },
},
```

This is controlled in https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/lexer/mod.rs#L591-L602. Appart from checking for `true`, `false` and `null`, we should also check for `NaN` and assign the correct rational, containing `f64::NaN`.

attliaLin:
This Pull Request fixes/closes #393.

It changes the following:
 - make NaN a number
 - add test for it

### DIFF
--- a/boa/src/syntax/lexer/mod.rs
+++ b/boa/src/syntax/lexer/mod.rs
@@ -592,6 +592,7 @@ impl<'a> Lexer<'a> {
                         "true" => TokenKind::BooleanLiteral(true),
                         "false" => TokenKind::BooleanLiteral(false),
                         "null" => TokenKind::NullLiteral,
+                        "NaN" => TokenKind::NumericLiteral(NumericLiteral::Rational(f64::NAN)),
                         slice => {
                             if let Ok(keyword) = FromStr::from_str(slice) {
                                 TokenKind::Keyword(keyword)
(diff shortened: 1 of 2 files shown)


jasonwilliams:
yep, looks good

simonbuchan:
This is fine as a hack, but NaN is actually specified as a global variable, like undefined, not a keyword, like null. (JavaScript is terrible.)
https://www.ecma-international.org/ecma-262/10.0/index.html#sec-global-object

Razican:
> This is fine as a hack, but NaN is actually specified as a global variable, like undefined, not a keyword, like null. (JavaScript is terrible.)
> https://www.ecma-international.org/ecma-262/10.0/index.html#sec-global-object

Yep, we will have to improve this in the future, but first, we need to improve the lexer, the current one is a bit difficult to follow, it became too big.

I opened #421 regarding this.

PR #395, opened 2020-05-12
impl abstract-equality-comparison

hello2dj:
This Pull Request fixes/closes https://github.com/jasonwilliams/boa/issues/351 https://github.com/jasonwilliams/boa/issues/357

It changes the following:
 - impl abstract-equality-comparison recursive version except BigInt & NaN
 - add tests
 - remove PartialEq on Value
 - same_value & same_value_zero impl

PR #399, opened 2020-05-13
v0.8

jasonwilliams:
@Razican @HalidOdat i made a start on the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 Feature Enhancements:
 
+- [FEATURE #121](https://github.com/jasonwilliams/boa/issues/121):
+  `BigInt` Implemented (@HalidOdat)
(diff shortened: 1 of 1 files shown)


HalidOdat:
We should definitively put Parser modularization #304 by @Razican

Razican:
I think this looks pretty good. Not sure how detailed this should be, I think it includes all PRs. I would wait for the rest of the 0.8-marked PRs to land and we could publish it afterwards.

I wouldn't add new ones to 0.8 (we could already create the 0.9 milestone and add issues/PRs to it)

Razican:
We should add #388, #395, #397, #362. Also a mention of the CoC, even if it's not a feature nor a fix would be nice.

In general, I think that we should include almost all of these: https://github.com/jasonwilliams/boa/pulls?q=is%3Apr+sort%3Aupdated-desc+milestone%3Av0.8.0+

PR #402, opened 2020-05-14
Json stringify replacer

### LINKED ISSUE #345: Implement the optional `replacer` parameter in `JSON.stringify( value[, replacer [, space] ] )`
The `JSON.stringify(value[, replacer[, space]])` takes in three arguments.

The `replacer` argument is a **function** that alters the behavior of the `stringification` process, or an **array** of `String` and `Number` objects that serve as a whitelist for selecting/filtering the properties of the value object to be included in the `JSON` string. If this value is `null` or not provided, all properties of the object are included in the resulting `JSON` string.

Example:
```JavaScript
function replacer(key, value) {
  // Filtering out properties
  if (typeof value === 'string') {
    return undefined;
  }
  return value;
}

var foo = {foundation: 'Mozilla', model: 'box', week: 45, transport: 'car', month: 7};
JSON.stringify(foo, replacer);
// '{"week":45,"month":7}'
```
More information:
 - [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)
 - [ECMAScript reference](https://tc39.es/ecma262/#sec-json.stringify)

### LINKED ISSUE #403: `Value::to_json()` does not handle `undefined` correctly
I'm currently working on the JSON.stringify replacer and I noticed something that caught me a little bit off guard. When I call `someValue.to_json()` it is not removing properties whose values are undefined. Since `value.to_json()` and `JSON.stringify(...)` are both taking things and turning them into "JSON", I would expect them to behave similarly. MDN gives this blurb on the JSON.stringify(...) page:
> undefined, Functions, and Symbols are not valid JSON values. If any such values are encountered during conversion they are either omitted (when found in an object) or changed to null (when found in an array). JSON.stringify() can return undefined when passing in "pure" values like JSON.stringify(function(){}) or JSON.stringify(undefined).

However, Boa's to_json function blindly turns null, symbols and undefined all into json null values.

https://github.com/jasonwilliams/boa/blob/402041d43251017ab772fe6f9920e41f496a359d/boa/src/builtins/value/mod.rs#L685

If this is very intentional behavior I can certainly do some checks in the `JSON.stringify()` to not add properties whose values are undefined, but thought I would pose the question since putting it in the to_json function would reproduce that behavior throughout the codebase.

- [ ] does not remove key/value pairs whose value is undefined
- [ ] does not remove key/value pairs whose value is a function
- [ ] does not remove key/value pairs whose value is a symbol
- [ ] json stringify arrays does not work properly
e.g.
```javascript
const y = [1, 2, 3];
console.log(JSON.stringify(y));
// Boa logs
// {"0":1,"1":2,"2":3,"length":3} --- INCORRECT

// Should log
// "[1, 2, 3]"

### PR DESCRIPTION
**Pull request by n14little**
Closes #345 
Fixes #403 

Additionally, wrote and ignored some tests that should be addressed once #405 is fixed.

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -99,14 +99,13 @@ pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -
                         key.to_owned(),
                         Property::default().value(
                             interpreter
-                                .call(replacer, &mut this_arg, &[Value::string(key), val.clone()])
-                                .expect("failed to get returned value from replacer function"),
+                                .call(replacer, &mut this_arg, &[Value::string(key), val.clone()])?,
                         ),
                     );
                 }
-                Value::from(object_to_return.to_json().to_string())
+                Ok(Value::from(object_to_return.to_json().to_string()))
             })
-            .ok_or_else(Value::undefined)
+            .ok_or_else(Value::undefined)?
     } else if replacer_as_object.kind == ObjectKind::Array {
         let mut obj_to_return =
             serde_json::Map::with_capacity(replacer_as_object.properties.len() - 1);
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -82,13 +82,15 @@ pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -
             let borrowed_derefed_replacer = derefed_obj.borrow();
             if borrowed_derefed_replacer.kind == ObjectKind::Array {
                 for (key, value) in borrowed_derefed_replacer.properties.iter() {
+                    if key == "length" {
+                        continue;
+                    }
+
                     if let Some(Value(x)) = &value.value {
-                        if key != "length" {
-                            object_to_return.set_property(
-                                x.to_string(),
-                                object.get_property(&x.to_string()).unwrap(),
-                            );
-                        }
+                        object_to_return.set_property(
+                            x.to_string(),
+                            object.get_property(&x.to_string()).unwrap(),
+                        );
                     }
                 }
             } else if borrowed_derefed_replacer.kind == ObjectKind::Function {
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -73,63 +73,62 @@ pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -
         None => return Ok(Value::undefined()),
         Some(obj) => obj,
     };
-        let replacer = match args.get(1) {
-            Some(replacer) if replacer.is_object() => replacer,
-            _ => return Ok(Value::from(object.to_json().to_string())),
-        };
+    let replacer = match args.get(1) {
+        Some(replacer) if replacer.is_object() => replacer,
+        _ => return Ok(Value::from(object.to_json().to_string())),
+    };
 
-        let replacer_as_object = replacer
+    let replacer_as_object = replacer
+        .as_object()
+        .expect("JSON.stringify replacer was an object");
+    if replacer_as_object.is_callable() {
+        object
             .as_object()
-            .expect("JSON.stringify replacer was an object");
-        if replacer_as_object.is_callable() {
-            object
-                .as_object()
-                .map(|obj| {
-                    let object_to_return = Value::new_object(None);
-                    for (key, val) in obj.properties.iter().filter_map(|(key, v)| {
-                        if let Some(value) = &v.value {
-                            Some((key, value))
-                        } else {
-                            None
-                        }
-                    }) {
-                        let mut this_arg = object.clone();
-                        object_to_return.set_property(
-                            key.to_owned(),
-                            Property::default().value(interpreter.call(
-                                replacer,
-                                &mut this_arg,
-                                &[Value::string(key), val.clone()],
-                            )?),
-                        );
+            .map(|obj| {
+                let object_to_return = Value::new_object(None);
+                for (key, val) in obj.properties.iter().filter_map(|(key, v)| {
+                    if let Some(value) = &v.value {
+                        Some((key, value))
+                    } else {
+                        None
                     }
-                    Ok(Value::from(object_to_return.to_json().to_string()))
-                })
-                .ok_or_else(Value::undefined)?
-        } else if replacer_as_object.kind == ObjectKind::Array {
-            let mut obj_to_return =
-                serde_json::Map::with_capacity(replacer_as_object.properties.len() - 1);
-            let fields = replacer_as_object.properties.keys().filter_map(|key| {
-                if key == "length" {
-                    None
-                } else {
-                    Some(replacer.get_field(key.to_string()))
+                }) {
+                    let mut this_arg = object.clone();
+                    object_to_return.set_property(
+                        key.to_owned(),
+                        Property::default().value(interpreter.call(
+                            replacer,
+                            &mut this_arg,
+                            &[Value::string(key), val.clone()],
+                        )?),
+                    );
                 }
-            });
-            for field in fields {
-                if let Some(value) = object
-                    .get_property(&field.to_string())
-                    .map(|prop| prop.value.as_ref().map(|v| v.to_json()))
-                    .flatten()
-                {
-                    obj_to_return.insert(field.to_string(), value);
-                }
+                Ok(Value::from(object_to_return.to_json().to_string()))
+            })
+            .ok_or_else(Value::undefined)?
+    } else if replacer_as_object.kind == ObjectKind::Array {
+        let mut obj_to_return =
+            serde_json::Map::with_capacity(replacer_as_object.properties.len() - 1);
+        let fields = replacer_as_object.properties.keys().filter_map(|key| {
+            if key == "length" {
+                None
+            } else {
+                Some(replacer.get_field(key.to_string()))
             }
-            Ok(Value::from(JSONValue::Object(obj_to_return).to_string()))
-        } else {
-            Ok(Value::from(object.to_json().to_string()))
+        });
+        for field in fields {
+            if let Some(value) = object
+                .get_property(&field.to_string())
+                .map(|prop| prop.value.as_ref().map(|v| v.to_json()))
+                .flatten()
+            {
+                obj_to_return.insert(field.to_string(), value);
+            }
         }
-
+        Ok(Value::from(JSONValue::Object(obj_to_return).to_string()))
+    } else {
+        Ok(Value::from(object.to_json().to_string()))
+    }
 }
 
 /// Create a new `JSON` object.
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -17,11 +17,10 @@ use crate::builtins::{
     function::make_builtin_fn,
     object::ObjectKind,
     property::Property,
-    value::{ResultValue, Value, ValueData},
+    value::{ResultValue, Value},
 };
 use crate::exec::Interpreter;
 use serde_json::{self, Value as JSONValue};
-use std::ops::Deref;
 
 #[cfg(test)]
 mod tests;
@@ -135,21 +134,6 @@ pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -
     } else {
         Ok(Value::from(object.to_json().to_string()))
     }
-
-    // let object_to_return = Value::new_object(None);
-    // if let Some(arg) = args.get(1) {
-    //     if let ValueData::Object(ref obj) = arg.data() {
-    //         let replacer = (*obj).deref().borrow();
-    //         match replacer.kind {
-    //             ObjectKind::Function => {
-
-    //             }
-    //             _ => panic!("JSON.stringify replacer can only be an array or function"),
-    //         }
-    //         return Ok(Value::from(object_to_return.to_json().to_string()));
-    //     }
-    // }
-    // panic!("cannot get replacer for JSON.stringify");
 }
 
 /// Create a new `JSON` object.
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -71,27 +71,22 @@ pub fn stringify(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultVa
     let object = args.get(0).expect("cannot get argument for JSON.stringify");
     let object_to_return = Value::new_object(None);
     if let Some(arg) = args.get(1) {
-        match arg.data() {
-            ValueData::Object(ref obj) => {
-                let derefed_obj = (*obj).deref();
-                let borrowed_derefed_obj = derefed_obj.borrow();
-                if borrowed_derefed_obj.kind == ObjectKind::Array {
-                    for (key, value) in borrowed_derefed_obj.properties.iter() {
-                        if let Some(Value(x)) = &value.value {
-                            if key != "length" {
-                                object_to_return.set_property(
-                                    x.to_string(),
-                                    object.get_property(&x.to_string()).unwrap(),
-                                );
-                            }
+        if let ValueData::Object(ref obj) = arg.data() {
+            let derefed_obj = (*obj).deref();
+            let borrowed_derefed_obj = derefed_obj.borrow();
+            if borrowed_derefed_obj.kind == ObjectKind::Array {
+                for (key, value) in borrowed_derefed_obj.properties.iter() {
+                    if let Some(Value(x)) = &value.value {
+                        if key != "length" {
+                            object_to_return.set_property(
+                                x.to_string(),
+                                object.get_property(&x.to_string()).unwrap(),
+                            );
                         }
                     }
-                    return Ok(Value::from(object_to_return.to_json().to_string()));
-                } else {
-                    panic!("replacer only supports arrays at this time");
                 }
-            }
-            _ => {
+                return Ok(Value::from(object_to_return.to_json().to_string()));
+            } else {
                 panic!("replacer only supports arrays at this time");
             }
         }
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -68,66 +68,65 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue 
 /// [spec]: https://tc39.es/ecma262/#sec-json.stringify
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
 pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
-    let object = match args.get(0) {
-        Some(obj) => obj,
-        None => return Ok(Value::undefined()),
-    };
+    if let Some(object) = args.get(0) {
+        let replacer = match args.get(1) {
+            Some(replacer) if replacer.is_object() => replacer,
+            _ => return Ok(Value::from(object.to_json().to_string())),
+        };
 
-    let replacer = match args.get(1) {
-        Some(replacer) if replacer.is_object() => replacer,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: boa/src/builtins/json/mod.rs [hunk 1]
```diff
@@ -63,8 +67,35 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+                            }
+                        }
+                    }
+                    return Ok(Value::from(object_to_return.to_json().to_string()));
+                } else {
+                    panic!("replacer only supports arrays at this time");
+                }
+            }
+            _ => {
+                panic!("replacer only supports arrays at this time");
```
Razican, line 94:
```suggestion
                unimplemented!("replacer only supports arrays at this time");
```

boa/src/builtins/json/mod.rs
```diff
@@ -63,8 +67,35 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+                                object_to_return.set_property(
+                                    x.to_string(),
+                                    object.get_property(&x.to_string()).unwrap(),
+                                );
+                            }
+                        }
+                    }
+                    return Ok(Value::from(object_to_return.to_json().to_string()));
+                } else {
+                    panic!("replacer only supports arrays at this time");
```
Razican, line 90:
```suggestion
                    unimplemented!("replacer only supports arrays at this time");
```

boa/src/builtins/json/mod.rs
```diff
@@ -63,8 +67,35 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
 /// [spec]: https://tc39.es/ecma262/#sec-json.stringify
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
 pub fn stringify(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
-    let obj = args.get(0).expect("cannot get argument for JSON.stringify");
-    let json = obj.to_json().to_string();
+    let object = args.get(0).expect("cannot get argument for JSON.stringify");
+    let object_to_return = Value::new_object(None);
+    if let Some(arg) = args.get(1) {
+        match arg.data() {
```
Razican, line 73: The warning here is just that it would be better to do:
```rust
if let ValueData::Object(ref obj) = arg.data() {
    // Deal with it
} else {
   unimplemented!("...");
}
```

n14little: :heavy_check_mark:

boa/src/builtins/json/mod.rs
n14little: I used the unimplemented macro per your suggestion where it is able to get the replacer argument. I did, however, keep the panic for when it fails to get the replacer argument. Thoughts?

n14little:
@Razican --- The error on this test feels like a bug, would you agree b22e5b9 ?

n14little:
Once I get the symbol thing figured out, I should be able to wrap this up pretty quickly.

Razican:
> @Razican --- The error on the test in commit [b22e5b9](https://github.com/jasonwilliams/boa/commit/b22e5b935f71cbbbb5f88be2f3b53f04bfb7c7d4) feels like a bug, would you agree?

Hmm this depends. Does the test panic if you run it against master? if so, we should open an issue about it.

Razican:
> @Razican -- Yes, it happens on master. Just opened an issue #405 

Good, so comment that out and we can progress with this issue :)

boa/src/builtins/json/mod.rs
```diff
@@ -62,10 +67,62 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+                            }
+                        }
+                    }
+                }
+                _ => panic!("JSON.stringify replacer can only be an array or function"),
+            }
+            return Ok(Value::from(object_to_return.to_json().to_string()));
+        }
+    }
+    panic!("cannot get replacer for JSON.stringify");
```
n14little, line 125: I'm really not in love with the perceived complexity of this function. Any suggestions on making it more succinct would be greatly appreciated, but I'm pretty new to Rust.

Razican: Let me check if this can be improved :)

Razican:
I checked [the spec][spec] and thought of an arguably simpler design:
```rust
pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
    let object = args.get(0).expect("cannot get argument for JSON.stringify");

    if args.len() == 1 {
        return Ok(Value::from(object.to_json().to_string()));
    }

    let replacer = args.get(1).expect("first element disappeared");
    if !replacer.is_object() {
        return Ok(Value::from(object.to_json().to_string()));
    }

    let replacer = replacer.as_object().expect("replacer was an object");
    if replacer.is_callable() {
        unimplemented!("replacer function");
    } else if replacer.kind == ObjectKind::Array {
        let mut obj_to_return = serde_json::Map::with_capacity(replacer.properties.len() - 1);
        for field in replacer.properties.iter().filter_map(|(key, prop)| {
            if key == "length" {
                None
            } else {
                // TODO: this should be the abstract operation `Get`
                prop.value.as_ref().map(Value::to_string)
            }
        }) {
            if let Some(value) = object
                .get_property(&field)
                .map(|prop| prop.value.as_ref().map(|v| v.to_json()))
                .flatten()
            {
                obj_to_return.insert(field, value);
            }
        }
        Ok(Value::from(JSONValue::Object(obj_to_return).to_string()))
    } else {
        Ok(Value::from(object.to_json().to_string()))
    }
}
```

Note that I didn't implement the function replacer and that I repeated `Ok(Value::from(object.to_json().to_string()))` too many times, so this should probably be replaced by a function or something.

The idea behind this is to avoid all that nesting. Then, instead of directly looping through all properties, I filtered the property iterator and mapped it to just return the string values.

Then, I just start creating a `JSONValue`, since there is no need to create an object to then convert it to JSON. I add only properties that exist and have a value, not sure if this this is 100% spec compliant, I just skimmed through it.

I did a bit of a heavy use of the [`Option` API][option] there, I recommend you to check it. Let me know if there is something unclear :)

[spec]: https://tc39.es/ecma262/#sec-json.stringify
[option]: https://doc.rust-lang.org/stable/std/option/enum.Option.html

n14little:
@Razican -- thanks for taking the time to write something up. I've implemented the array replacer functionality per your suggestion. Still working on the function replacer.

> The idea behind this is to avoid all that nesting.

That is what I was hoping to get rid of.

n14little:
Made an attempt at using Options in the function replacer branch.

n14little:
@Razican thoughts on how this is looking?

Razican:
> @Razican thoughts on how this is looking?

I will check this today :)

boa/src/builtins/json/mod.rs
```diff
@@ -62,10 +66,73 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+    let replacer_as_object = replacer
+        .as_object()
+        .expect("JSON.stringify replacer was an object");
+    if replacer_as_object.is_callable() {
+        object
+            .as_object()
+            .map(|obj| {
+                let object_to_return = Value::new_object(None);
+                for (key, val) in obj.properties.iter() {
+                    if let Some(value) = &val.value {
```
Razican, line 90: There is no real need for checking this, we can just filter the properties by running:
```rust
for (key, val) in obj.properties.iter().filter(|(_, val)|val.value.is_some()) {}
```

boa/src/builtins/json/mod.rs
```diff
@@ -62,10 +66,73 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+                        let mut this_arg = object.clone();
+                        object_to_return.set_property(
+                            String::from(key),
```
Razican, line 93: It was good before, but I think this gives a better semantic meaning:
```suggestion
                            key.to_owned(),
```

boa/src/builtins/value/mod.rs
```diff
@@ -723,16 +723,33 @@ impl ValueData {
...
-                let new_obj = obj
-                    .borrow()
-                    .properties
-                    .iter()
-                    .map(|(k, _)| (k.clone(), self.get_field_slice(k).to_json()))
-                    .collect::<Map<String, JSONValue>>();
-                JSONValue::Object(new_obj)
+                if obj.borrow().kind == ObjectKind::Array {
+                    let mut arr: Vec<JSONValue> = Vec::new();
+                    obj.borrow().properties.iter().for_each(|(k, _)| {
```
Razican, line 731: I think it would make sense to only iterate over `properties.keys()`, since we don't use the values.

boa/src/builtins/value/mod.rs
```diff
@@ -723,16 +723,33 @@ impl ValueData {
...
+                                arr.push(JSONValue::Null);
+                            } else {
+                                arr.push(self.get_field_slice(k).to_json());
+                            }
+                        }
+                    });
+                    JSONValue::Array(arr)
+                } else {
+                    let mut new_obj = Map::new();
+                    obj.borrow().properties.iter().for_each(|(k, _)| {
```
Razican, line 744: Same here, we should only iterate over `keys()`.

boa/src/builtins/json/mod.rs
```diff
@@ -62,10 +66,73 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+            serde_json::Map::with_capacity(replacer_as_object.properties.len() - 1);
+        let fields = replacer_as_object
+            .properties
+            .iter()
+            .filter_map(|(key, prop)| {
+                if key == "length" {
+                    None
+                } else {
+                    // TODO: this should be the abstract operation `Get`
+                    prop.value.as_ref().map(Value::to_string)
```
Razican, line 120: We should change this to do `get_field()` as we do in `to_json()`, and iterate only over the keys.

n14little:
@Razican --- addressed all of your comments.

boa/src/builtins/json/mod.rs
```diff
@@ -65,10 +67,70 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+
+    let replacer_as_object = replacer
+        .as_object()
+        .expect("JSON.stringify replacer was an object");
+    if replacer_as_object.is_callable() {
+        object
+            .as_object()
+            .map(|obj| {
+                let object_to_return = Value::new_object(None);
+                for (key, val) in obj.properties.iter().filter(|(_, v)|v.value.is_some()) {
```
Razican, line 90: Since we are only using the `v.value` here, and it must be `Some()`, I would suggest using `.filter_map()` instead of `.filter()` and returning `Some((key, value))` with an `if let Some(value) = v.value`, and `None` otherwise. That way we can just call `value.clone()` when passing the third parameter to `call()` below.

boa/src/builtins/json/mod.rs
```diff
@@ -65,10 +67,70 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+                    object_to_return.set_property(
+                        key.to_owned(),
+                        Property::default().value(
+                            interpreter
+                                .call(
+                                    replacer,
+                                    &mut this_arg,
+                                    &[Value::string(key), Value::from(val.value.as_ref())],
+                                )
+                                .expect("failed to get returned value from replacer function"),
```
Razican, line 101: We now have exception handling in place, so if this call fails, it should throw an exception. I think we can use the `?` operator here, but we need to return an `Ok(Value::from...)` below, and the `else` should be `Ok(Value::undefined())`. Then, we should be able to use the `?` operator again to propagate the exception.

We should also add a test for this, with a `call` that throws an exception.

boa/src/builtins/json/tests.rs
```diff
@@ -16,3 +16,155 @@ fn json_sanity() {
...
+    let actual = forward(
+        &mut engine,
+        r#"JSON.stringify({x: 1}, (key, value) => {
+         throw new TypeError("type error")
+     })"#,
+    );
+    let expected = forward(&mut engine, r#"'Error: undefined'"#);
+
+    assert_eq!(actual, expected);
+}
```
n14little, line 170: Not entirely sure how to test this. I followed the same pattern that has been applied in this file and just captured the output of what I think you were suggesting in the production code.

Razican: We don't have a `TypeError` yet, so that's why this is returning `undefined`. What we could do is to just put it inside of a `try..catch`:

```javascript
	let a = 0;
    try {
		JSON.stringify({x: 1}, (key, value) => {
			throw 1;
    	});
	} catch (e) {
		a = e;
	}
	a;
```

This should output `1`.

n14little: @Razican -- Done :heavy_check_mark:

boa/src/builtins/json/mod.rs
```diff
@@ -65,10 +67,69 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
 ///
 /// [spec]: https://tc39.es/ecma262/#sec-json.stringify
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
-pub fn stringify(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
-    let obj = args.get(0).expect("cannot get argument for JSON.stringify");
-    let json = obj.to_json().to_string();
-    Ok(Value::from(json))
+pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
+    let object = args.get(0).expect("cannot get argument for JSON.stringify");
```
HalidOdat, line 71:
```suggestion
    let object = match args.get(0) {
        Some(object) => object,
        None => return Ok(Value::undefined()),
    };
```
Calling `JSON.stringify()` should not panic if no value has supplied

boa/src/builtins/json/mod.rs
```diff
@@ -65,10 +67,69 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+
+    if args.len() == 1 {
+        return Ok(Value::from(object.to_json().to_string()));
+    }
+
+    let replacer = args.get(1).expect("cannot get JSON.stringify replacer");
+    if !replacer.is_object() {
+        return Ok(Value::from(object.to_json().to_string()));
+    }
```
HalidOdat, line 80:
```suggestion
    let replacer = match args.get(1) {
        Some(replacer) if replacer.is_object() => replacer,
        _ => return Ok(Value::from(object.to_json().to_string())),
    };
```
Some code cleanup

boa/src/builtins/json/mod.rs
n14little: :heavy_check_mark:

boa/src/builtins/json/mod.rs
n14little: :heavy_check_mark:

n14little:
@HalidOdat -- addressed your comments. Added a test for the case where nothing is passed to stringify.

boa/src/builtins/json/mod.rs
```diff
@@ -65,10 +67,68 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
 ///
 /// [spec]: https://tc39.es/ecma262/#sec-json.stringify
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
-pub fn stringify(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
-    let obj = args.get(0).expect("cannot get argument for JSON.stringify");
-    let json = obj.to_json().to_string();
-    Ok(Value::from(json))
+pub fn stringify(_: &mut Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
+    let object = match args.get(0) {
```
Razican, line 71: Can we replace this with an `if let`? Should improve readability.

boa/src/builtins/json/mod.rs
```diff
@@ -65,10 +67,68 @@ pub fn parse(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue
...
+    let replacer_as_object = replacer
+        .as_object()
+        .expect("JSON.stringify replacer was an object");
+    if replacer_as_object.is_callable() {
+        object
+            .as_object()
+            .map(|obj| {
+                let object_to_return = Value::new_object(None);
+                for (key, val) in obj.properties.iter().filter_map(|(key, v)| {
+                    if let Some(value) = &v.value {
```
Razican, line 90: We should use `Option::map()` here, that does exactly this.

boa/src/builtins/value/mod.rs
```diff
@@ -737,6 +754,9 @@ impl ValueData {
                 // TODO: throw TypeError
                 panic!("TypeError: \"BigInt value can't be serialized in JSON\"");
             }
+            Self::Symbol(_) | Self::Undefined => {
+                panic!("Symbols and Undefined JSON Values depend on parent type");
```
Razican, line 758: Just for my understanding, is this in the spec? Should we check this before calling? Or should this throw an exception?

boa/src/builtins/json/mod.rs
n14little: @Razican I changed it, but I'd have to say that I think the match is better on readability. The else block, IMO, is too far removed from the `if`.

boa/src/builtins/value/mod.rs
n14little: @Razican 
> Just for my understanding, is this in the spec? Should we check this before calling?
According to NOTE 5 in the spec, JSON.stringify should return undefined. However, symbols, functions, undefined do not have corresponding counterparts, and thus, it depends on what their parent is. I did everything I could to keep that knowledge in the to_json function.

> Or should this throw an exception?
I am not sure. I can look that up later, unless somebody else reviewing knows that answer. cc: @HalidOdat

HalidOdat: From my understanding when calling `JSON.stringify` with `symbol` or `undefined` it should return `undefined`. A workaround would be to check if it is `undefined` or `symbol` in `JSON.stringify` and return undefined before calling `Value::to_json`.

What do you think ? @Razican @jasonwilliams

boa/src/builtins/json/mod.rs
n14little: Reverted back to match because there were some uncovered branches that needed to be added. Let me know your thoughts.

boa/src/builtins/value/mod.rs
n14little: @HalidOdat You are definitely correct WRT JSON.stringify returning undefined for symbols, functions and undefined inputs. I added some tests and updated to pass those tests. The question does still remain here on whether or not to panic. My thought process behind panicking was that if this branch is hit, it is because the calling function needs to defend against these values. Thus, when this panics we'll just need to create a bug and address it.

Razican: > @HalidOdat You are definitely correct WRT `JSON.stringify` returning undefined for symbols, functions and undefined inputs. I added some tests and updated to pass those tests. The question does still remain here on whether or not to panic. My thought process behind panicking was that if this branch is hit, it is because the calling function needs to defend against these values. Thus, when this panics we'll just need to create a bug and address it.

I would use `unreachable!()` for this.

boa/src/builtins/json/mod.rs
n14little: :heavy_check_mark:

boa/src/builtins/value/mod.rs
n14little: :heavy_check_mark:

n14little:
@Razican @HalidOdat -- I think that I have all comments addressed.

PR #404, opened 2020-05-18
Fix for 0 length new String

### LINKED ISSUE #398: String() does not set length on instance
**Describe the bug**
Calling new String('some text') does not set the instance

E.g
```
let a = new String("1234");
a.length;
```

Should return 4

**To Reproduce**
Put the above into a test.js file and run it, you should get 0

**Expected behavior**
it should return 4

[spec]: https://tc39.es/ecma262/#sec-stringcreate

Implementation is here:
Master: https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/string/mod.rs#L31-L49

The string passed in lives at args.get(0), you can call len() then convert it back into a JSValue by calling `to_value()` on it.
There's an example commented out here:
https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/string/mod.rs#L36-L37

```rust
 let a: String = from_value(args.get(0).expect("failed to get argument for String method").clone()).unwrap();
 this.set_field_slice("length", to_value(a.len() as i32));
```

tylermorten:
This Pull Request fixes/closes #398 .

It changes the following:
 - Adds the field length with the proper length of the string to the context
 - Adds a test for testing the fields is set correctly.

### DIFF
--- a/boa/src/builtins/string/tests.rs
+++ b/boa/src/builtins/string/tests.rs
@@ -52,6 +52,21 @@ fn new_string_has_length() {
 }
 
 #[test]
+fn new_utf8_string_has_length() {
+    let realm = Realm::create();
+    let mut engine = Executor::new(realm);
+    let init = r#"
+        let a = new String("中文");
+        a
+        "#;
+
+    eprintln!("{}", forward(&mut engine, init));
+
+    let empty = String::from("2");
+    assert_eq!(forward(&mut engine, "a.length"), empty);
+}
+
+#[test]
 fn concat() {
     let realm = Realm::create();
     let mut engine = Executor::new(realm);
--- a/boa/src/builtins/string/tests.rs
+++ b/boa/src/builtins/string/tests.rs
@@ -45,10 +45,8 @@ fn new_string_has_length() {
         a
         "#;
 
-    eprintln!("{}", forward(&mut engine, init));
-
-    let empty = String::from("4");
-    assert_eq!(forward(&mut engine, "a.length"), empty);
+    forward(&mut engine, init);
+    assert_eq!(forward(&mut engine, "a.length"), "4");
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


tylermorten:
Thanks!

For the UTF-8 encoding, I'm guessing you want to support the count of the characters - not the byte length, correct?  It appears that when I check the length of a String in node it gives me back the character count, not the byte count.  I believe that the default for std::String len() returns the byte count...perhaps chars().count() would work better for this implementation?

Razican:
> perhaps chars().count() would work better for this implementation?

I think so, yes, I saw you already implemented this, let me give it another look :)

boa/src/builtins/string/tests.rs
```diff
@@ -36,6 +36,36 @@ fn check_string_constructor_is_function() {
...
+    let mut engine = Executor::new(realm);
+    let init = r#"
+        let a = new String("1234");
+        a
+        "#;
+
+    eprintln!("{}", forward(&mut engine, init));
+
+    let empty = String::from("4");
+    assert_eq!(forward(&mut engine, "a.length"), empty);
```
Razican, line 51: We don't need to print everything here, it will slow down testing. If this fails in the future, we can debug then. I also made the final assert easier to read :)
```suggestion
    forward(&mut engine, init);
    assert_eq!(&forward(&mut engine, "a.length"), "4");
```

tylermorten: Ahh!  Got it. I was just copying other tests as a template.  I'll put in your suggestions and push again.  Thanks!

jasonwilliams:
Sorry for slow response but yes it looks good!

Thanks @tylermorton

tylermorten:
Thank you!  I'll continue to look through issues to see if there is anything else I can help on.

PR #406, opened 2020-05-19
Dependency upgrade

Razican:
I upgraded all dependencies both for our wasm demo and for our core in Rust. This (with some potential extra upgrades until the release) should get merged for 0.8.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,9 +76,9 @@ dependencies = [
 
 [[package]]
 name = "bstr"
-version = "0.2.12"
+version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2889e6d50f394968c8bf4240dc3f2a7eb4680844d27308f798229ac9d4725f41"
+checksum = "31accafdb70df7871592c058eca3985b71104e15ac32f64706022c58867da931"
 dependencies = [
  "lazy_static",
  "memchr",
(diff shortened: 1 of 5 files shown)


Razican:
I found 3 issues with this:
 - `text-encoding` is [deprecated](https://www.npmjs.com/package/text-encoding), we should use [`fast-text-encoding`](https://www.npmjs.com/package/fast-text-encoding), but it seems it's not compatible with wasm-pack.
 - `copy-webpack-plugin` breaks when upgrading it to `^6.0.0`. This was only released 4 days ago, so it's not a big deal, but it seems it requires a change on how we configure it.
 - We have no CI whatsoever checking our WASM target, if I'm not mistaken, we should include a `wasm-pack test` command with maybe some tests to make sure that our WASM target builds and performs as expected.

Razican:
This is ready for a merge!

jasonwilliams:
> We have no CI whatsoever checking our WASM target, if I'm not mistaken, we should include a wasm-pack test command with maybe some tests to make sure that our WASM target builds and performs as expected.

Have you made an issue for this?

Razican:
> > We have no CI whatsoever checking our WASM target, if I'm not mistaken, we should include a wasm-pack test command with maybe some tests to make sure that our WASM target builds and performs as expected.
> 
> Have you made an issue for this?

Yep, #408.

PR #409, opened 2020-05-22
Fix function object constructable/callable

HalidOdat:
It changes the following:
 - Removed unnesecary `FunctionKind` and cleanup
 - Added `name` property to global objects. example: BigInt.name === "BigInt"
 - Added `length` property to global objects. example: Boolean.length === 1
 - Made arrow functions only callable.
 - Fixes function objects constructability and callability.

Not all function objects are constructable or callable for example:
|        Type        | constructable 				|  callable |
| ------------------ | ---------------------------- |-----------|
| Arrow Functions    |    false  				    |   true    |
| Function Decl/Expr |    true   				    |   true    |
| Classes            |    true       				|   false   |
| Builtins           | false (unless explicit true) |   true    |

More informations see: <https://stackoverflow.com/a/40922715>

### DIFF
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -14,7 +14,7 @@
 use crate::{
     builtins::{
         array::Array,
-        object::{Object, ObjectInternalMethods, ObjectKind, PROTOTYPE},
+        object::{Object, ObjectInternalMethods, ObjectKind, INSTANCE_PROTOTYPE, PROTOTYPE},
         property::Property,
         value::{ResultValue, Value},
     },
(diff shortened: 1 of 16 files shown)


boa/src/builtins/function/mod.rs
```diff
@@ -372,19 +378,25 @@ pub fn make_function(this: &mut Value, _: &[Value], _: &mut Interpreter) -> Resu
...
-    make_constructor_fn(make_function, global, prototype)
+    make_constructor_fn("Function", 1, make_function, global, prototype, true)
 }
 
 /// Creates a new constructor function
 ///
 /// This utility function handling linking the new Constructor to the prototype.
 /// So far this is only used by internal functions
-pub fn make_constructor_fn(body: NativeFunctionData, global: &Value, proto: Value) -> Value {
+pub fn make_constructor_fn(
```
jasonwilliams, line 388: The comments above may need updating

HalidOdat:
If everything is good with this PR, could we merge it. It has some changes that would make it easier to implement #419 .

PR #410, opened 2020-05-22
Add support for the reviver function to JSON.parse

### LINKED ISSUE #344: Implement the optional `reviver` parameter in `JSON.parse( text [, reviver] )`
Currently the `JSON.parse( text[, reviver] )` method only supports the first parameter which takes in a `String` and converts into an `Object`.
But, there is an optional function argument that `JSON.parse( text[, reviver] )` method takes. It prescribes how the value originally produced by parsing is **transformed**, before being returned.

Example:
```javascript
JSON.parse('{"1": 1, "2": 2, "3": {"4": 4, "5": {"6": 6}}}', (key, value) => {
  console.log(key); // log the current property name, the last is "".
  return value;     // return the unchanged property value.
});
```

More information:
 - [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse)
 - [ECMAScript reference](https://tc39.es/ecma262/#sec-json.parse)

You can check the [`Array.prototype.every`](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/array/mod.rs#L491) implementation to see how to deal with JavaScript callback functions.

abhijeetbhagat:
This Pull Request tries to fix #344.

This PR needs lots of changes to the way we extract object while recursing. I tried using `match` on the `Value` using `as_object()` but this results in a borrow problem since i am passing a `&mut Value` to `walk`. So at the moment, i am cloning the unwrapped object after checking if it is safe to do so.

Need feedback on that.


Razican:
Thanks for the contribution! This will collide with #402, though. We should probably wait for that to land.

abhijeetbhagat:
@Razican Sure! Meanwhile, can you please take a look at the changes?

Razican:
> @Razican Sure! Meanwhile, can you please take a look at the changes?

Yep, I will check this today :)

boa/src/builtins/json/mod.rs
```diff
@@ -33,19 +33,67 @@ mod tests;
...
+                            walk(callback, interpreter, &mut holder, Value::from(""))
+                        } else {
+                            Ok(j)
+                        }
+                    }
+                    _ => Ok(j),
+                }
+            } else {
+                Ok(j)
+            }
```
HalidOdat, line 61:
```suggestion
            match args.get(1) {
                Some(callback) if callback.is_function() => {
                    let mut holder = Value::new_object(None);
                    holder.set_field(Value::from(""), j);
                    walk(callback, interpreter, &mut holder, Value::from(""))
                }
                _ => Ok(j)
            }
```
Simplified the `JSON.parse` function.

boa/src/builtins/json/mod.rs
```diff
@@ -33,19 +33,67 @@ mod tests;
...
+                            value.set_field(Value::from(key.as_str()), v);
+                        } else {
+                            value.remove_property(key.as_str());
+                        }
+                    }
+                    Err(_v) => {}
+                }
+            }
+        }
+    }
```
HalidOdat, line 93:
```suggestion
    if let Some(obj) = value.clone().as_object() {
        for key in obj.properties.keys() {
            let v = walk(callback, interpreter, &mut value, Value::from(key.as_str()));
            match v {
                Ok(v) if v.is_undefined() => {
                    value.set_field(Value::from(key.as_str()), v);
                }
                Ok(_) => {
                    value.remove_property(key.as_str());
                }
                Err(_) => {}
            }
        }
    }
```
Simplified the `walk` function.

abhijeetbhagat: Wow! your changes have simplified the monstrosity i wrote above.
However, there are two problems -
1. a clippy warning for clone on line 75 above: `cloned value is neither consumed nor mutated`
2. panic that says, `'GcCell<T> already borrowed'` with this redacted trace:
```
  12: std::panicking::begin_panic_fmt
             at src\libstd/panicking.rs:373
  13: gc::GcCell<T>::borrow_mut
             at boa\src/lib.rs:1
  14: boa::builtins::value::ValueData::set_field
             at boa\src\builtins\value/mod.rs:616
  15: boa::builtins::json::walk
             at boa\src\builtins\json/mod.rs:72
  16: boa::builtins::json::parse
             at boa\src\builtins\json/mod.rs:50
```

abhijeetbhagat: To be honest, i'd tried using `if let`  with clone combination but because of the above panic, I gave up and was going to ask you how to fix it.

Razican: I think the issue might be in the difference between cloning first and then getting the object or first getting the object and then cloning. So maybe we could change this:

```rust
    if let Some(obj) = value.clone().as_object() {
        for key in obj.properties.keys() {
```

for this:
```rust
    if let Some(obj) = value.as_object() {
        let obj = obj.clone();
        for key in obj.properties.keys() {
```

abhijeetbhagat: Unfortunately, `value.as_object()` borrows `value` immutably  and when that borrow is still active, we cant borrow it mutably and pass to walk: `let v = walk(callback, interpreter, &mut value, Value::from(key.as_str()));`

HalidOdat:
```rust
if let Some(obj) = value.as_object().map(|x| x.clone()) {
```
I think this should fix it, right?

abhijeetbhagat: It did! Learnt a lot of things today! Shout out to you both @Razican and @HalidOdat

HalidOdat: while refactoring I made a mistake in line 79
```rust
                Ok(v) if v.is_undefined() => {
```

it should be 
```rust
                Ok(v) if !v.is_undefined() => {
```

I have updated the comment :)

abhijeetbhagat: Clippy has not liked our sneakiness though: `You are needlessly cloning iterator elements`
We also can't use  `cloned()` as clippy suggests [here](https://rust-lang.github.io/rust-clippy/master/index.html#map_clone).

abhijeetbhagat: > while refactoring I made a mistake at line 79
> 
> ```rust
>                 Ok(v) if v.is_undefined() => {
> ```
> 
> It should be
> 
> ```rust
>                 Ok(v) if !v.is_undefined() => {
> ```
> 
> I updated the comment :)

Oh yeah, i took care of that already.

HalidOdat: We could use allow attribute to stop clippys complaining!
```rust
#[allow(clippy::map_clone)]
fn walk(
    callback: &Value,
    interpreter: &mut Interpreter,
    holder: &mut Value,
    key: Value,
) -> ResultValue {
```

abhijeetbhagat: @HalidOdat Brilliant! I'll keep that in mind the next time I wrestle with clippy :)

Razican: > We could use allow attribute to stop clippys complaining!
> ```rust
> #[allow(clippy::map_clone)]
> fn walk(
>     callback: &Value,
>     interpreter: &mut Interpreter,
>     holder: &mut Value,
>     key: Value,
> ) -> ResultValue {
> ```

Actually, I think it's best if we substitute the `.map(|x| x.clone())` by just `.cloned()`, that should be equivalent, right?

We should try to avoid allowing clippy lints if possible.

HalidOdat: Sadly `.cloned()` doesn't work it gives this error:
```
error[E0599]: no method named `cloned` found for enum `std::option::Option<gc::GcCellRef<'_, builtins::object::Object>>` in the current scope
   --> boa/src/builtins/json/mod.rs:68:42
    |
68  |       if let Some(obj) = value.as_object().cloned() {
    |                                            ^^^^^^ method not found in `std::option::Option<gc::GcCellRef<'_, builtins::object::Object>>`
    |
    = note: the method `cloned` exists but the following trait bounds were not satisfied:
            `std::option::Option<gc::GcCellRef<'_, builtins::object::Object>>: std::iter::Iterator`
            which is required by `&mut std::option::Option<gc::GcCellRef<'_, builtins::object::Object>>: std::iter::Iterator`

```

HalidOdat: Found a way to avoid allowing clippy lint:
```rust
    let object = value.as_object().as_deref().cloned();
    if let Some(obj) = object {
```

abhijeetbhagat: This one nailed it.

HalidOdat:
The next thing is adding test, so we insure it does not break in the future.

abhijeetbhagat:
Yes, working on it.

abhijeetbhagat:
@Razican @HalidOdat review?

boa/src/builtins/json/mod.rs
```diff
@@ -38,19 +38,59 @@ mod tests;
...
+            }
+        }
         Err(err) => Err(Value::from(err.to_string())),
     }
 }
 
+/// This is a translation of the [Polyfill implementation][polyfill]
+/// This function recursively walks the structure. passing each key-value pair to the reviver function
+/// for possible transformation
+/// [polyfill]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
```
HalidOdat, line 67:
```suggestion
/// This is a translation of the [Polyfill implementation][polyfill]
///
/// This function recursively walks the structure. passing each key-value pair to the reviver function
/// for possible transformation
///
/// [polyfill]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
```
Separate function title from  function description

boa/src/builtins/json/tests.rs
```diff
@@ -189,3 +189,49 @@ fn json_stringify_return_undefined() {
...
+            if (typeof v == 'number') {
+                return v * 2;
+            } else {
+                v
+        }})"#,
+    );
+    assert_eq!(
+        result,
+        "{\n    3: 8,\n    2: 6,\n    0: 2,\n    1: 4,\n    length: 4,\n    extensible: true\n}"
+    );
```
HalidOdat, line 209: I think a better way of doing this is to call `forward_val` which returns a `Value` and call `.get_field` and then call `to_number()` on them to get a number that you can assert against

abhijeetbhagat: Done

boa/src/builtins/json/mod.rs
abhijeetbhagat: Done

abhijeetbhagat:
@HalidOdat review?

Issue #378, opened 2020-05-07
Code coverage report


Razican:
Currently, we have no coverage report for Boa. How much are we testing? To know this, we should probably use a coverage platform (such as [CodeCov](https://codecov.io/)) and a code coverage reporting tool, such as [Tarpaulin](https://github.com/xd009642/tarpaulin).

There are also some interesting GitHub [actions](https://actions-rs.github.io/#code-coverage) for this.

PR #411, opened 2020-05-22
Add code coverage: tarpaulin / codecov

Lan2u:
Relates to issue #378.

Added tarpaulin action to the test suite on Linux section of the CI, the produced report is then submitted to codecov.io. I used my own personal token for testing the codecov part but have removed it for this commit (it can be re-added into the actions file - this is why that stage currently fails).


Lan2u:
I haven't done much with this except get tarpaulin to run, this is my first attempt to work on an issue as part of an open-source project so if I have missed the mark please let me know.

.github/workflows/rust.yml
```diff
@@ -82,7 +82,16 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: test
-          args: -v
+          args: -v
```
Razican, line 85:
```suggestion
          args: -v
```

.github/workflows/rust.yml
```diff
@@ -82,7 +82,16 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: test
-          args: -v
+          args: -v  
+      - name: Run cargo-tarpaulin
+        uses: actions-rs/tarpaulin@v0.1
+        with:
+          version: '0.12.4'
```
Razican, line 89: We don't need to specify a version here. The documentation mentions that if you don't specify anything, you will be getting the latest version.

.github/workflows/rust.yml
```diff
@@ -82,7 +82,16 @@ jobs:
...
+          args: '-- --test-threads 1'
```
Razican, line 90: Why do we need to specify this?

.github/workflows/rust.yml
```diff
@@ -82,7 +82,16 @@ jobs:
...
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@v1.0.2
+        with:
+          token:
```
Razican, line 94: We don't need to specify a token, since this is a public repository. This can be removed :)

.github/workflows/rust.yml
Lan2u: Removed

.github/workflows/rust.yml
Lan2u: Tried without and it seems to be working fine so removed.

.github/workflows/rust.yml
Lan2u: I removed this but it seems to still fail? Perhaps it is because I'm trying from my fork of the repo.

Razican: I think this is because we still don't have the project in CodeCov. Public repos shouldn't need a token.

Lan2u:
> It seems you modified the `Cargo.lock` file. Could we avoid this? It's giving merge conflicts, but there should be no modification to that file :)

Done, sorry I thought I already removed it.

Razican:
Ok, let's wait for @jasonwilliams to see if we can have Boa in codecov so that this works :)

jasonwilliams:
I've now added the repo, give it a try and let us know

Lan2u:
> I've now added the repo, give it a try and let us know

Tried again but still codecov fails saying it needs a token

jasonwilliams:
@Lan2u could you try again?

Lan2u:
> @Lan2u could you try again?

Tried again, the same result unfortunately :(

Razican:
I think it just needs to be initialized. We could try to merge this without the token and see if it works.

jasonwilliams:
Lets see what happens

jasonwilliams:
@Lan2u did you want to make a new PR against master?

Lan2u:
> still not working https://github.com/boa-dev/boa/runs/727333120?check_suite_focus=true

Perhaps related to https://community.codecov.io/t/github-public-repository-issues/1360? 

Will try and investigate further

Lan2u:
Seems to be fixed, Pull request #436

Issue #401, opened 2020-05-13
implement Map()


jasonwilliams:
**ECMASCript feature**
Map objects are collections of key/value pairs where both the keys and values may be arbitrary ECMAScript language values. A distinct key value may only occur in one key/value pair within the Map's collection. Distinct key values are discriminated using the SameValueZero comparison algorithm.

I would like to see `Map` implemented. [ECMAScript specification][spec].

[spec]: https://tc39.es/ecma262/#sec-map-objects

**Example code**
```javascript
let myMap = new Map()

let keyString = 'a string'
let keyObj    = {}
let keyFunc   = function() {}

// setting the values
myMap.set(keyString, "value associated with 'a string'")
myMap.set(keyObj, 'value associated with keyObj')
myMap.set(keyFunc, 'value associated with keyFunc')

myMap.size              // 3

// getting the values
myMap.get(keyString)    // "value associated with 'a string'"
myMap.get(keyObj)       // "value associated with keyObj"
myMap.get(keyFunc)      // "value associated with keyFunc"

myMap.get('a string')    // "value associated with 'a string'"
                         // because keyString === 'a string'
myMap.get({})            // undefined, because keyObj !== {}
myMap.get(function() {}) // undefined, because keyFunc !== function () {}
```

**Example to work from**
Array is implemented here: https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/array/mod.rs

**Contributing**
https://github.com/jasonwilliams/boa/blob/master/CONTRIBUTING.md

hello2dj:
I can take a stap at this

jasonwilliams:
Assigned

hello2dj:
Oh sorry, I've been very busy recently. I'll start as soon as possible. 
Before I start to do it. I need some help, [List](https://tc39.es/ecma262/#sec-list-and-record-specification-type) Type I will use Array(https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/array/mod.rs). Someone has other suggestions？

hello2dj:
https://github.com/boa-dev/boa/pull/419  I will wait for this PR.

hello2dj:
sorry, I'm too busy to have time to do this. I will unassign me.

joshwd36:
I'm happy to take a look at this

HalidOdat:
> I'm happy to take a look at this

Sure! Tell us if you need any help or have any question :)

joshwd36:
[...]

Issue #400, opened 2020-05-13
Implement Set()


jasonwilliams:
**ECMASCript feature**
Set objects are collections of ECMAScript language values. A distinct value may only occur once as an element of a Set's collection. Distinct values are discriminated using the SameValueZero comparison algorithm.

I would like to see `Set` implemented. [ECMAScript specification][spec].

[spec]: https://tc39.es/ecma262/#sec-set-objects

**Example code**
```javascript
let mySet = new Set()

mySet.add(1)           // Set [ 1 ]
mySet.add(5)           // Set [ 1, 5 ]
mySet.add(5)           // Set [ 1, 5 ]
mySet.add('some text') // Set [ 1, 5, 'some text' ]
let o = {a: 1, b: 2}
mySet.add(o)

mySet.add({a: 1, b: 2})   // o is referencing a different object, so this is okay

mySet.has(1)              // true
mySet.has(3)              // false, since 3 has not been added to the set
mySet.has(5)              // true
mySet.has(Math.sqrt(25))  // true
mySet.has('Some Text'.toLowerCase()) // true
mySet.has(o)       // true

mySet.size         // 5

mySet.delete(5)    // removes 5 from the set
mySet.has(5)       // false, 5 has been removed

mySet.size         // 4, since we just removed one value

console.log(mySet)
// logs Set(4) [ 1, "some text", {…}, {…} ] in Firefox
// logs Set(4) { 1, "some text", {…}, {…} } in Chrome
```

**Example to work from**
Array is implemented here: https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/array/mod.rs

**Contributing**
https://github.com/jasonwilliams/boa/blob/master/CONTRIBUTING.md

brian-gavin:
Hey all, I'm interested in picking this up.

jasonwilliams:
Assigned

brian-gavin:
Have a WIP local implementation, but relies on `internal_slot`, so will be waiting for #419 to be completed before raising a PR for this one.

HalidOdat:
> Have a WIP local implementation, but relies on `internal_slot`, so will be waiting for #419 to be completed before raising a PR for this one.

HalidOdat:
> Have a WIP local implementation, but relies on `internal_slot`, so will be waiting for #419 to be completed before raising a PR for this one.

Closed the issue accidentally :sweat_smile: .

[...]

Issue #364, opened 2020-05-02
Symbol can be an Object type instead of a native type


jasonwilliams:
Symbol is currently a native type on [ValueData](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L40-L60).

It doesn't need to be, it can be an Object with it's kind set to Symbol.

then in places like [is_symbol()](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L109-L115) just match on Object then check the objectKind. We do something similar, example here:
https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L731-L781

An example of creating an instance object, take a look at Boolean:
https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/boolean/mod.rs#L26-L38

**Spec**
https://tc39.es/ecma262/#sec-symbol-objects

**Debugging**
https://github.com/jasonwilliams/boa/blob/master/docs/debugging.md

HalidOdat:
> Symbol is currently a native type on [ValueData](https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L40-L60) [](/jasonwilliams/boa/blob/HEAD@{2020-05-02T16:21:08Z}/boa/src/builtins/value/mod.rs#L40-L60)
>
> It doesn't need to be, it can be an Object with it's kind set to Symbol.

Isn't it more performant to separate it?

Razican:
> Isn't it more performant to separate it?

It could be, yes, but then, it might make sense to also specialize Arrays, Strings and other type of objects. I'm not against it, since it could allow us to probably use other kind of Rust inner types, but maybe there should be an `Object` enum?

HalidOdat:
> It could be, yes, but then, it might make sense to also specialize Arrays, Strings and other type of objects. I'm not against it, since it could allow us to probably use other kind of Rust inner types, but maybe there should be an `Object` enum?

In v8 they specialize the type like `String` for primitive and `StringObject` for string `Object`.
This helps speed up things since we don't have to look in an internal slot for the data, instead we check the native Rust `String` type.
And so on for the other types.

Razican:
That would be a great addition. Still, I do think that `Symbol` is a type o object and not a different type of JS value. So maybe the `Object` enum makes sense.

HalidOdat:
[...]

PR #419, opened 2020-05-25
Object specialization

HalidOdat:
It changes the following:
 - Specializes objects
 - Implements `Hash` trait for `Value`
 - Implements `PartialEq` trait based on the `SameValueZero` algorithm for `Value` (required by `Hash` trait)
 - Implements `Eq` trait for `Value` (required by `Hash` trait)
 - Fixed a bug when comparing object pointers in `same_value_non_number`
 - Deprecated `ObjectInternalMethods`
 - Fixed Symbol primitive
 - Added `construct_type_error` and `construct_range_error`
 - Made `Number` objects constructor and methods more specification compliant.
 - Abstracted Object name and function object length.
 - Changed `init` functions to return `(&str, Value)`

The problem with the current implementation is that we store the data in an internal slot. The internal slots are stored as a `HashMap`, which means slow reads and writes. Additionally they are stored as a `Value` which means we need additional checks to unwrap the Value into a primitive (example `f64`, `String`) to access the data. Furthermore we are restricted as to what type of data we can hold, for example if we wanted to implement `Set` object with native type `HashSet`, we can not do so in the internal slots as they are specified in the standard (it's only an implementation detail).
This PR aims to allow use of non-`Value` types in objects as well as removing the need for internals slots for data storage (`NumberData`, `StringData`, etc), removing redundant checks for data access. It should increase performance greatly for object access and a bit for object creation. (I'm not sure the benchmarks will show it though, since we do not have benchmarks for objects).

Specialized objects:
 - [x] Number
 - [x] String
 - [x] Function
 - [ ] Array
 - [x] Symbol
 - [x] Boolean
 - [x] BigInt
 - [ ] RegExp

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -15,7 +15,7 @@ mod tests;
 use super::function::{make_builtin_fn, make_constructor_fn};
 use crate::{
     builtins::{
-        object::{ObjectKind, INSTANCE_PROTOTYPE, PROTOTYPE},
+        object::{ObjectData, INSTANCE_PROTOTYPE, PROTOTYPE},
         property::Property,
         value::{same_value_zero, ResultValue, Value, ValueData},
     },
@@ -42,7 +42,7 @@ impl Array {
                 .get_global_object()
                 .expect("Could not get global object"),
         ));
-        array.set_kind(ObjectKind::Array);
+        array.set_data(ObjectData::Array);
         array.borrow().set_internal_slot(
             INSTANCE_PROTOTYPE,
             interpreter
@@ -117,7 +117,7 @@ impl Array {
         this.set_internal_slot(INSTANCE_PROTOTYPE, prototype);
         // This value is used by console.log and other routines to match Object type
         // to its Javascript Identifier (global constructor method name)
-        this.set_kind(ObjectKind::Array);
+        this.set_data(ObjectData::Array);
 
         // add our arguments in
         let mut length = args.len() as i32;
@@ -176,7 +176,7 @@ impl Array {
                     // 1.
                     ValueData::Object(ref obj) => {
                         // 2.
-                        if (*obj).deref().borrow().kind == ObjectKind::Array {
+                        if let ObjectData::Array = (*obj).deref().borrow().data {
                             return Ok(value_true);
                         }
                         Ok(value_false)
@@ -1008,7 +1008,7 @@ impl Array {
         let prototype = Value::new_object(None);
         let length = Property::default().value(Value::from(0));
 
-        prototype.set_property_slice("length", length);
+        prototype.set_property("length", length);
 
         make_builtin_fn(Self::concat, "concat", &prototype, 1);
         make_builtin_fn(Self::push, "push", &prototype, 1);
@@ -1043,6 +1043,8 @@ impl Array {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("array", "init");
-        global.set_field("Array", Self::create(global));
+
+        let array = Self::create(global);
+        global.as_object_mut().unwrap().insert_field("Array", array);
     }
 }
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -1041,10 +1041,9 @@ impl Array {
 
     /// Initialise the `Array` object on the global object.
     #[inline]
-    pub(crate) fn init(global: &Value) {
+    pub(crate) fn init(global: &Value) -> (&str, Value) {
         let _timer = BoaProfiler::global().start_event("array", "init");
 
-        let array = Self::create(global);
-        global.as_object_mut().unwrap().insert_field("Array", array);
+        ("Array", Self::create(global))
     }
 }
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -36,7 +36,7 @@ impl Array {
     pub(crate) const NAME: &'static str = "Array";
 
     /// The amount of arguments this function object takes.
-    pub(crate) const LENGTH: i32 = 1;
+    pub(crate) const LENGTH: usize = 1;
 
     /// Creates a new `Array` instance.
     pub(crate) fn new_array(interpreter: &Interpreter) -> ResultValue {
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -25,7 +25,6 @@ use crate::{
 use std::{
     borrow::Borrow,
     cmp::{max, min},
-    ops::Deref,
 };
 
 /// JavaScript `Array` built-in implementation.
@@ -33,6 +32,12 @@ use std::{
 pub(crate) struct Array;
 
 impl Array {
+    /// The name of the object.
+    pub(crate) const NAME: &'static str = "Array";
+
+    /// The amount of arguments this function object takes.
+    pub(crate) const LENGTH: i32 = 1;
+
     /// Creates a new `Array` instance.
     pub(crate) fn new_array(interpreter: &Interpreter) -> ResultValue {
         let array = Value::new_object(Some(
@@ -167,25 +172,9 @@ impl Array {
         args: &[Value],
         _interpreter: &mut Interpreter,
     ) -> ResultValue {
-        let value_true = Value::boolean(true);
-        let value_false = Value::boolean(false);
-
-        match args.get(0) {
-            Some(arg) => {
-                match arg.data() {
-                    // 1.
-                    ValueData::Object(ref obj) => {
-                        // 2.
-                        if let ObjectData::Array = (*obj).deref().borrow().data {
-                            return Ok(value_true);
-                        }
-                        Ok(value_false)
-                    }
-                    // 3.
-                    _ => Ok(value_false),
-                }
-            }
-            None => Ok(value_false),
+        match args.get(0).and_then(|x| x.as_object()) {
+            Some(object) => Ok(Value::from(object.is_array())),
+            None => Ok(Value::from(false)),
         }
     }
 
@@ -1031,7 +1020,14 @@ impl Array {
         make_builtin_fn(Self::slice, "slice", &prototype, 2);
         make_builtin_fn(Self::some, "some", &prototype, 2);
 
-        let array = make_constructor_fn("Array", 1, Self::make_array, global, prototype, true);
+        let array = make_constructor_fn(
+            Self::NAME,
+            Self::LENGTH,
+            Self::make_array,
+            global,
+            prototype,
+            true,
+        );
 
         // Static Methods
         make_builtin_fn(Self::is_array, "isArray", &array, 1);
@@ -1042,8 +1038,8 @@ impl Array {
     /// Initialise the `Array` object on the global object.
     #[inline]
     pub(crate) fn init(global: &Value) -> (&str, Value) {
-        let _timer = BoaProfiler::global().start_event("array", "init");
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
-        ("Array", Self::create(global))
+        (Self::NAME, Self::create(global))
     }
 }
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -15,6 +15,7 @@
 use crate::{
     builtins::{
         function::{make_builtin_fn, make_constructor_fn},
+        object::ObjectData,
         value::{ResultValue, Value, ValueData},
     },
     exec::Interpreter,
@@ -62,9 +63,8 @@ impl BigInt {
             // 2. If Type(value) is Object and value has a [[BigIntData]] internal slot, then
             //    a. Assert: Type(value.[[BigIntData]]) is BigInt.
             //    b. Return value.[[BigIntData]].
-            ValueData::Object(_) => {
-                let bigint = value.get_internal_slot("BigIntData");
-                if let ValueData::BigInt(bigint) = bigint.data() {
+            ValueData::Object(ref object) => {
+                if let ObjectData::BigInt(ref bigint) = object.borrow().data {
                     return Ok(bigint.clone());
                 }
             }
@@ -86,16 +86,12 @@ impl BigInt {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-bigint-objects
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/BigInt
-    pub(crate) fn make_bigint(
-        _this: &mut Value,
-        args: &[Value],
-        ctx: &mut Interpreter,
-    ) -> ResultValue {
+    pub(crate) fn make_bigint(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
         let data = match args.get(0) {
-            Some(ref value) => Value::from(ctx.to_bigint(value)?),
-            None => Value::from(Self::from(0)),
+            Some(ref value) => ctx.to_bigint(value)?,
+            None => Self::from(0),
         };
-        Ok(data)
+        Ok(Value::from(data))
     }
 
     /// `BigInt.prototype.toString( [radix] )`
@@ -213,7 +209,6 @@ impl BigInt {
     /// Create a new `Number` object
     pub(crate) fn create(global: &Value) -> Value {
         let prototype = Value::new_object(Some(global));
-        prototype.set_internal_slot("BigIntData", Value::from(Self::from(0)));
 
         make_builtin_fn(Self::to_string, "toString", &prototype, 1);
         make_builtin_fn(Self::value_of, "valueOf", &prototype, 0);
@@ -230,7 +225,12 @@ impl BigInt {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("bigint", "init");
-        global.set_field("BigInt", Self::create(global));
+
+        let bigint = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
+            .insert_field("BigInt", bigint);
     }
 }
 
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -44,6 +44,12 @@ mod tests;
 pub struct BigInt(num_bigint::BigInt);
 
 impl BigInt {
+    /// The name of the object.
+    pub(crate) const NAME: &'static str = "BigInt";
+
+    /// The amount of arguments this function object takes.
+    pub(crate) const LENGTH: i32 = 1;
+
     /// The abstract operation thisBigIntValue takes argument value.
     ///
     /// The phrase “this BigInt value” within the specification of a method refers to the
@@ -213,7 +219,14 @@ impl BigInt {
         make_builtin_fn(Self::to_string, "toString", &prototype, 1);
         make_builtin_fn(Self::value_of, "valueOf", &prototype, 0);
 
-        let big_int = make_constructor_fn("BigInt", 1, Self::make_bigint, global, prototype, false);
+        let big_int = make_constructor_fn(
+            Self::NAME,
+            Self::LENGTH,
+            Self::make_bigint,
+            global,
+            prototype,
+            false,
+        );
 
         make_builtin_fn(Self::as_int_n, "asIntN", &big_int, 2);
         make_builtin_fn(Self::as_uint_n, "asUintN", &big_int, 2);
@@ -224,9 +237,9 @@ impl BigInt {
     /// Initialise the `BigInt` object on the global object.
     #[inline]
     pub(crate) fn init(global: &Value) -> (&str, Value) {
-        let _timer = BoaProfiler::global().start_event("bigint", "init");
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
-        ("BigInt", Self::create(global))
+        (Self::NAME, Self::create(global))
     }
 }
 
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -223,14 +223,10 @@ impl BigInt {
 
     /// Initialise the `BigInt` object on the global object.
     #[inline]
-    pub(crate) fn init(global: &Value) {
+    pub(crate) fn init(global: &Value) -> (&str, Value) {
         let _timer = BoaProfiler::global().start_event("bigint", "init");
 
-        let bigint = Self::create(global);
-        global
-            .as_object_mut()
-            .unwrap()
-            .insert_field("BigInt", bigint);
+        ("BigInt", Self::create(global))
     }
 }
 
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -78,8 +78,7 @@ impl BigInt {
         }
 
         // 3. Throw a TypeError exception.
-        ctx.throw_type_error("'this' is not a BigInt")?;
-        unreachable!();
+        Err(ctx.construct_type_error("'this' is not a BigInt"))
     }
 
     /// `BigInt()`
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -48,7 +48,7 @@ impl BigInt {
     pub(crate) const NAME: &'static str = "BigInt";
 
     /// The amount of arguments this function object takes.
-    pub(crate) const LENGTH: i32 = 1;
+    pub(crate) const LENGTH: usize = 1;
 
     /// The abstract operation thisBigIntValue takes argument value.
     ///
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -27,6 +27,12 @@ use crate::{
 pub(crate) struct Boolean;
 
 impl Boolean {
+    /// The name of the object.
+    pub(crate) const NAME: &'static str = "Boolean";
+
+    /// The amount of arguments this function object takes.
+    pub(crate) const LENGTH: i32 = 1;
+
     /// An Utility function used to get the internal [[BooleanData]].
     ///
     /// More information:
@@ -101,8 +107,8 @@ impl Boolean {
         make_builtin_fn(Self::value_of, "valueOf", &prototype, 0);
 
         make_constructor_fn(
-            "Boolean",
-            1,
+            Self::NAME,
+            Self::LENGTH,
             Self::construct_boolean,
             global,
             prototype,
@@ -113,8 +119,8 @@ impl Boolean {
     /// Initialise the `Boolean` object on the global object.
     #[inline]
     pub(crate) fn init(global: &Value) -> (&str, Value) {
-        let _timer = BoaProfiler::global().start_event("boolean", "init");
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
-        ("Boolean", Self::create(global))
+        (Self::NAME, Self::create(global))
     }
 }
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -45,8 +45,9 @@ impl Boolean {
             _ => {}
         }
 
-        ctx.throw_type_error("'this' is not a boolean")?;
-        unreachable!();
+        Err(ctx
+            .throw_type_error("'this' is not a boolean")
+            .expect_err("throw_type_error() did not return an error"))
     }
 
     /// `[[Construct]]` Create a new boolean object
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -15,19 +15,40 @@ mod tests;
 use super::function::{make_builtin_fn, make_constructor_fn};
 use crate::{
     builtins::{
-        object::{internal_methods_trait::ObjectInternalMethods, ObjectKind},
+        object::ObjectData,
         value::{ResultValue, Value, ValueData},
     },
     exec::Interpreter,
     BoaProfiler,
 };
-use std::{borrow::Borrow, ops::Deref};
 
 /// Boolean implementation.
 #[derive(Debug, Clone, Copy)]
 pub(crate) struct Boolean;
 
 impl Boolean {
+    /// An Utility function used to get the internal [[BooleanData]].
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-thisbooleanvalue
+    fn this_boolean_value(value: &Value, ctx: &mut Interpreter) -> Result<bool, Value> {
+        match value.data() {
+            ValueData::Boolean(boolean) => return Ok(*boolean),
+            ValueData::Object(ref object) => {
+                let object = object.borrow();
+                if let Some(boolean) = object.as_boolean() {
+                    return Ok(boolean);
+                }
+            }
+            _ => {}
+        }
+
+        ctx.throw_type_error("'this' is not a boolean")?;
+        unreachable!();
+    }
+
     /// `[[Construct]]` Create a new boolean object
     ///
     /// `[[Call]]` Creates a new boolean primitive
@@ -36,19 +57,11 @@ impl Boolean {
         args: &[Value],
         _: &mut Interpreter,
     ) -> ResultValue {
-        this.set_kind(ObjectKind::Boolean);
-
         // Get the argument, if any
-        if let Some(ref value) = args.get(0) {
-            this.set_internal_slot("BooleanData", Self::to_boolean(value));
-        } else {
-            this.set_internal_slot("BooleanData", Self::to_boolean(&Value::from(false)));
-        }
+        let data = args.get(0).map(|x| x.to_boolean()).unwrap_or(false);
+        this.set_data(ObjectData::Boolean(data));
 
-        match args.get(0) {
-            Some(ref value) => Ok(Self::to_boolean(value)),
-            None => Ok(Self::to_boolean(&Value::from(false))),
-        }
+        Ok(Value::from(data))
     }
 
     /// The `toString()` method returns a string representing the specified `Boolean` object.
@@ -60,9 +73,9 @@ impl Boolean {
     /// [spec]: https://tc39.es/ecma262/#sec-boolean-object
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/toString
     #[allow(clippy::wrong_self_convention)]
-    pub(crate) fn to_string(this: &mut Value, _: &[Value], _: &mut Interpreter) -> ResultValue {
-        let b = Self::this_boolean_value(this);
-        Ok(Value::from(b.to_string()))
+    pub(crate) fn to_string(this: &mut Value, _: &[Value], ctx: &mut Interpreter) -> ResultValue {
+        let boolean = Self::this_boolean_value(this, ctx)?;
+        Ok(Value::from(boolean.to_string()))
     }
 
     /// The valueOf() method returns the primitive value of a `Boolean` object.
@@ -73,37 +86,8 @@ impl Boolean {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-boolean.prototype.valueof
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/valueOf
-    pub(crate) fn value_of(this: &mut Value, _: &[Value], _: &mut Interpreter) -> ResultValue {
-        Ok(Self::this_boolean_value(this))
-    }
-
-    // === Utility Functions ===
-    /// [toBoolean](https://tc39.es/ecma262/#sec-toboolean)
-    /// Creates a new boolean value from the input
-    #[allow(clippy::wrong_self_convention)]
-    pub(crate) fn to_boolean(value: &Value) -> Value {
-        match *value.deref().borrow() {
-            ValueData::Object(_) => Value::from(true),
-            ValueData::String(ref s) if !s.is_empty() => Value::from(true),
-            ValueData::Rational(n) if n != 0.0 && !n.is_nan() => Value::from(true),
-            ValueData::Integer(n) if n != 0 => Value::from(true),
-            ValueData::Boolean(v) => Value::from(v),
-            _ => Value::from(false),
-        }
-    }
-
-    /// An Utility function used to get the internal BooleanData.
-    ///
-    /// More information:
-    ///  - [ECMAScript reference][spec]
-    ///
-    /// [spec]: https://tc39.es/ecma262/#sec-thisbooleanvalue
-    pub(crate) fn this_boolean_value(value: &Value) -> Value {
-        match *value.deref().borrow() {
-            ValueData::Boolean(v) => Value::from(v),
-            ValueData::Object(ref v) => (v).deref().borrow().get_internal_slot("BooleanData"),
-            _ => Value::from(false),
-        }
+    pub(crate) fn value_of(this: &mut Value, _: &[Value], ctx: &mut Interpreter) -> ResultValue {
+        Ok(Value::from(Self::this_boolean_value(this, ctx)?))
     }
 
     /// Create a new `Boolean` object.
@@ -111,7 +95,6 @@ impl Boolean {
         // Create Prototype
         // https://tc39.es/ecma262/#sec-properties-of-the-boolean-prototype-object
         let prototype = Value::new_object(Some(global));
-        prototype.set_internal_slot("BooleanData", Self::to_boolean(&Value::from(false)));
 
         make_builtin_fn(Self::to_string, "toString", &prototype, 0);
         make_builtin_fn(Self::value_of, "valueOf", &prototype, 0);
@@ -130,6 +113,11 @@ impl Boolean {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("boolean", "init");
-        global.set_field("Boolean", Self::create(global));
+
+        let boolean = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
+            .insert_field("Boolean", boolean);
     }
 }
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -111,13 +111,9 @@ impl Boolean {
 
     /// Initialise the `Boolean` object on the global object.
     #[inline]
-    pub(crate) fn init(global: &Value) {
+    pub(crate) fn init(global: &Value) -> (&str, Value) {
         let _timer = BoaProfiler::global().start_event("boolean", "init");
 
-        let boolean = Self::create(global);
-        global
-            .as_object_mut()
-            .unwrap()
-            .insert_field("Boolean", boolean);
+        ("Boolean", Self::create(global))
     }
 }
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -51,9 +51,7 @@ impl Boolean {
             _ => {}
         }
 
-        Err(ctx
-            .throw_type_error("'this' is not a boolean")
-            .expect_err("throw_type_error() did not return an error"))
+        Err(ctx.construct_type_error("'this' is not a boolean"))
     }
 
     /// `[[Construct]]` Create a new boolean object
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -31,7 +31,7 @@ impl Boolean {
     pub(crate) const NAME: &'static str = "Boolean";
 
     /// The amount of arguments this function object takes.
-    pub(crate) const LENGTH: i32 = 1;
+    pub(crate) const LENGTH: usize = 1;
 
     /// An Utility function used to get the internal [[BooleanData]].
     ///
@@ -93,6 +93,7 @@ impl Boolean {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-boolean.prototype.valueof
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/valueOf
+    #[inline]
     pub(crate) fn value_of(this: &mut Value, _: &[Value], ctx: &mut Interpreter) -> ResultValue {
         Ok(Value::from(Self::this_boolean_value(this, ctx)?))
     }
--- a/boa/src/builtins/console/mod.rs
+++ b/boa/src/builtins/console/mod.rs
@@ -556,5 +556,10 @@ pub fn create(global: &Value) -> Value {
 #[inline]
 pub fn init(global: &Value) {
     let _timer = BoaProfiler::global().start_event("console", "init");
-    global.set_field("console", create(global));
+
+    let console = create(global);
+    global
+        .as_object_mut()
+        .unwrap()
+        .insert_field("console", console);
 }
--- a/boa/src/builtins/console/mod.rs
+++ b/boa/src/builtins/console/mod.rs
@@ -554,12 +554,8 @@ pub fn create(global: &Value) -> Value {
 
 /// Initialise the `console` object on the global object.
 #[inline]
-pub fn init(global: &Value) {
+pub fn init(global: &Value) -> (&str, Value) {
     let _timer = BoaProfiler::global().start_event("console", "init");
 
-    let console = create(global);
-    global
-        .as_object_mut()
-        .unwrap()
-        .insert_field("console", console);
+    ("console", create(global))
 }
--- a/boa/src/builtins/error/mod.rs
+++ b/boa/src/builtins/error/mod.rs
@@ -35,6 +35,12 @@ pub(crate) use self::range::RangeError;
 pub(crate) struct Error;
 
 impl Error {
+    /// The name of the object.
+    pub(crate) const NAME: &'static str = "Error";
+
+    /// The amount of arguments this function object takes.
+    pub(crate) const LENGTH: i32 = 1;
+
     /// Create a new error object.
     pub(crate) fn make_error(this: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
         if !args.is_empty() {
@@ -77,13 +83,20 @@ impl Error {
 
         make_builtin_fn(Self::to_string, "toString", &prototype, 0);
 
-        make_constructor_fn("Error", 1, Self::make_error, global, prototype, true)
+        make_constructor_fn(
+            Self::NAME,
+            Self::LENGTH,
+            Self::make_error,
+            global,
+            prototype,
+            true,
+        )
     }
 
     /// Initialise the global object with the `Error` object.
     pub(crate) fn init(global: &Value) -> (&str, Value) {
-        let _timer = BoaProfiler::global().start_event("error", "init");
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
-        ("Error", Self::create(global))
+        (Self::NAME, Self::create(global))
     }
 }
--- a/boa/src/builtins/error/mod.rs
+++ b/boa/src/builtins/error/mod.rs
@@ -81,10 +81,9 @@ impl Error {
     }
 
     /// Initialise the global object with the `Error` object.
-    pub(crate) fn init(global: &Value) {
+    pub(crate) fn init(global: &Value) -> (&str, Value) {
         let _timer = BoaProfiler::global().start_event("error", "init");
 
-        let error = Self::create(global);
-        global.as_object_mut().unwrap().insert_field("Error", error);
+        ("Error", Self::create(global))
     }
 }
--- a/boa/src/builtins/error/mod.rs
+++ b/boa/src/builtins/error/mod.rs
@@ -13,7 +13,7 @@
 use crate::{
     builtins::{
         function::{make_builtin_fn, make_constructor_fn},
-        object::ObjectKind,
+        object::ObjectData,
         value::{ResultValue, Value},
     },
     exec::Interpreter,
@@ -49,7 +49,7 @@ impl Error {
         }
         // This value is used by console.log and other routines to match Object type
         // to its Javascript Identifier (global constructor method name)
-        this.set_kind(ObjectKind::Error);
+        this.set_data(ObjectData::Error);
         Err(this.clone())
     }
 
@@ -83,6 +83,8 @@ impl Error {
     /// Initialise the global object with the `Error` object.
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("error", "init");
-        global.set_field("Error", Self::create(global));
+
+        let error = Self::create(global);
+        global.as_object_mut().unwrap().insert_field("Error", error);
     }
 }
--- a/boa/src/builtins/error/mod.rs
+++ b/boa/src/builtins/error/mod.rs
@@ -39,7 +39,7 @@ impl Error {
     pub(crate) const NAME: &'static str = "Error";
 
     /// The amount of arguments this function object takes.
-    pub(crate) const LENGTH: i32 = 1;
+    pub(crate) const LENGTH: usize = 1;
 
     /// Create a new error object.
     pub(crate) fn make_error(this: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
@@ -94,6 +94,7 @@ impl Error {
     }
 
     /// Initialise the global object with the `Error` object.
+    #[inline]
     pub(crate) fn init(global: &Value) -> (&str, Value) {
         let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
--- a/boa/src/builtins/error/range.rs
+++ b/boa/src/builtins/error/range.rs
@@ -13,7 +13,7 @@ use crate::{
     builtins::{
         function::make_builtin_fn,
         function::make_constructor_fn,
-        object::ObjectKind,
+        object::ObjectData,
         value::{ResultValue, Value},
     },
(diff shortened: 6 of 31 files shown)


HalidOdat:
@Razican and @jasonwilliams I'm having a bit of trouble with the garbage collector.
I'm getting this error when construction/calling `Array`:
```
thread 'main' panicked at 'GcCell<T> already borrowed', <::std::macros::panic macros>:5:6
```

Any help would be useful?

Razican:
> @Razican and @jasonwilliams I'm having a bit of trouble with the garbage collector.
> I'm getting this error when construction/calling `Array`:
> 
> ```
> thread 'main' panicked at 'GcCell<T> already borrowed', <::std::macros::panic macros>:5:6
> ```
> 
> Any help would be useful

I'm not very used to deal with the Gc, but this seems to be that two calls to `borrow()` were made one after the other without a `clone()` or a scope ending in the middle.

boa/src/builtins/value/mod.rs
```diff
@@ -614,7 +614,7 @@ impl ValueData {
         let val = val.into();
 
         if let Self::Object(ref obj) = *self {
-            if let ObjectData::Array = obj.borrow().data {
+            if obj.borrow().data.is_array() {
```
Razican, line 617: I would add this function also to the `Object`, in order to not need to do `obj.data.is_array()` and just do `obj.is_array()`. They can be aliases, though.

HalidOdat:
I did some benchmarks on Object creation and access

I think the other execution benchmarks are wrong, we use `exec` function to benchmark execution, but exec does not just execute the the string, it lexes, parsers, creates the interpreter and initializes/creates the realm and then it executes the code, aren't we polluting the result of the benchmark? Shouldn't we isolate what were executing as much as possible, right?
With `exec` function, I saw a `~4%`, when I isolated the execution I saw a `~20%` increase for object creation/access.

You can check the benchmark [here](https://gist.github.com/HalidOdat/3d4b8f136e125e00c0cc9fbc52a65a57)

What do you think @Razican and @jasonwilliams ?

Razican:
> I did some benchmarks on Object creation and access
> 
> I think the other execution benchmarks are wrong, we use `exec` function to benchmark execution, but `exec` does not just execute the string, it lexes, parsers, creates the interpreter and initializes/creates the realm and then it executes the code, aren't we polluting the result of the benchmark? Shouldn't we isolate what were executing as much as possible, right?
> With `exec` function, I saw a `~4%`, when I isolated the execution I saw a `~20%` increase for object creation/access.
> 
> You can check the benchmark [here](https://gist.github.com/HalidOdat/3d4b8f136e125e00c0cc9fbc52a65a57)
> 
> What do you think @Razican and @jasonwilliams ?

@jasonwilliams noticed that most of the execution time is spent on Realm creation, so I think we should isolate the actual execution, yes, in order to see if we improve or worsen there.

jasonwilliams:
I think we need to make some benchmark issues and get them in before this so we can have some before/after

Razican:
We should review the benchmarks action, since it seems it's not giving accurate results for the percentages:

 - 195.0±2.10µs vs 453.0±7.55µs is a 57% improvement, not a -32%
 - 207.8±2.80µs vs 468.6±6.77µs is a 55.7% improvement, not -26%

We might either want % of change, in which case, the first would be a -57%, and second would be a -55.7%, or a new vs old, in which case the new one would be 45% for the first one and the second would be 44.3%.

I'm more in favour for the second one, so a 101% that is being shown today, would be replaced by a +1%.

jasonwilliams:
> It's looking amazing. One thing about benchmarks:
> 
> We should review the benchmarks action, since it seems it's not giving accurate results for the percentages:
> 
> * 195.0±2.10µs vs 453.0±7.55µs is a 57% improvement, not a -32%
> * 207.8±2.80µs vs 468.6±6.77µs is a 55.7% improvement, not -26%
> 
> We might either want % of change, in which case, the first would be a -57%, and second would be a -55.7%, or a new vs old, in which case the new one would be 45% for the first one and the second would be 44.3%.
> 
> I'm more in favour for the second one, so a 101% that is being shown today, would be replaced by a +1%.

This is done by the criterion-compare-action, ive moved it into boa-dev (as its only purpose was for boa anyway)

https://github.com/boa-dev/criterion-compare-action/issues/1
I've made an issue for this and we can sort it out over there.

Razican:
I think this should also take #364 into account and remove `Symbol` from ValueData.

HalidOdat:
> I think this should also take #364 into account and remove `Symbol` from ValueData.

I think so too!
From my understanding `Symbol` is like `Function`, both objects with special behaviour. I was having trouble specializing it because it's
separate.

jasonwilliams:
> I think this should also take #364 into account and remove `Symbol` from ValueData.

I was wrong when i said remove it in that issue, it is a primitive, and i think we can make it pretty lightweight, it only needs to hold a unique identifier (a number or a hash), i think it should be ok on on ValueData

HalidOdat:
> I was wrong when i said remove it in that issue, it is a primitive, and i think we can make it pretty lightweight, it only needs to hold a unique identifier (a number or a hash), i think it should be ok on on ValueData

Interesting. but `Symbols` can have a protoype so is there a object version too @jasonwilliams

jasonwilliams:
> > I was wrong when i said remove it in that issue, it is a primitive, and i think we can make it pretty lightweight, it only needs to hold a unique identifier (a number or a hash), i think it should be ok on on ValueData
> 
> Interesting. but `Symbols` can have a protoype so is there a object version too @jasonwilliams

That's correct, i was referring to the instance of Symbol.
`Symbol()` is like calling `String()` without new, it just returns a primitive type (that's why it doesn't have a `new` associated with it

HalidOdat:
> > > I was wrong when i said remove it in that issue, it is a primitive, and i think we can make it pretty lightweight, it only needs to hold a unique identifier (a number or a hash), i think it should be ok on on ValueData
> > 
> > 
> > Interesting. but `Symbols` can have a protoype so is there a object version too @jasonwilliams
> 
> That's correct, i was referring to the instance of Symbol rather than the Constructor for them.
> 
> `Symbol()` is like calling `String()` without new, it just returns a primitive type (that's why it doesn't have a `new` associated with it).
> 
> The instance is very lightweight

I see. so I should remove the Box<Gc<Object>> from ValueData::Symbol and add something lightweight like the description and hash.

jasonwilliams:
> > > > I was wrong when i said remove it in that issue, it is a primitive, and i think we can make it pretty lightweight, it only needs to hold a unique identifier (a number or a hash), i think it should be ok on on ValueData
> > > 
> > > 
> > > Interesting. but `Symbols` can have a protoype so is there a object version too @jasonwilliams
> > 
> > 
> > That's correct, i was referring to the instance of Symbol rather than the Constructor for them.
> > `Symbol()` is like calling `String()` without new, it just returns a primitive type (that's why it doesn't have a `new` associated with it).
> > The instance is very lightweight
> 
> I see. so I should remove the Box<Gc> from ValueData::Symbol and add something lightweight like the description and hash. and in the Object I can specialize the ObjectData::Symbol

Yep

jasonwilliams:
For extra info:
This is the Spidermonkey implementation:
https://github.com/mozilla/gecko-dev/blob/master/js/src/vm/SymbolType.cpp#L22-L32 and it hashes on line 44

HalidOdat:
I think we should merge this, even though I haven't specialized all the types, I don't want to block #400 and #401, we can work on it at the same time.

https://github.com/Manishearth/rust-gc/pull/103 would make it much easier to specialize the other types like `Array` in a much cleaner way.

Razican:
> I think we should merge this, even though I haven't specialized all the types, I don't want to block #400 and #401, we can work on it at the same time.

I will give this a look tomorrow. Ideally, I would like to merge #431 before this, to see the real change in execution times.

HalidOdat:
Lets merge this now that we have the benchmarks! This should unblock #400 and #401.

Note: I did not specialize all objects Like `RegExp` and `Array`. I intend to do them later.

Razican:
> Lets merge this now that we have the benchmarks! This should unblock #400 and #401.

I will review this today :) great work!

> I did not specialize objects (`RegExp` and `Array`) these will take some time because they require to rewrite all the methods. I intend to do them later.

Actually, for `RegExp`, we probably want to implement a 100% ES compliant engine. I had the link to a crate in Rust, I will create an issue for this.

> BTW: benchmarks have double `-` negative sign, this is weird
> 
> Edit: created PR to fix the benchmarks double sign https://github.com/boa-dev/criterion-compare-action/pull/7

Merged!

boa/src/builtins/array/mod.rs
```diff
@@ -1043,6 +1043,8 @@ impl Array {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("array", "init");
-        global.set_field("Array", Self::create(global));
+
+        let array = Self::create(global);
+        global.as_object_mut().unwrap().insert_field("Array", array);
```
Razican, line 1048: I would use `expect()` instead of `unwrap()` here.

Some extra optimisation that could be done here would be for the `init` function to return a list of `<&str, Value>` pairs, and then add that to the global object in the calling function. This would allow doing this with multiple threads in the future, so that the initialization gets done in paralel. This should probably go in a new PR, though.

boa/src/builtins/bigint/mod.rs
```diff
@@ -230,7 +225,12 @@ impl BigInt {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("bigint", "init");
-        global.set_field("BigInt", Self::create(global));
+
+        let bigint = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
```
Razican, line 232: Once again, I would use `expect()` here.

boa/src/builtins/boolean/mod.rs
```diff
@@ -15,19 +15,40 @@ mod tests;
...
+                let object = object.borrow();
+                if let Some(boolean) = object.as_boolean() {
+                    return Ok(boolean);
+                }
+            }
+            _ => {}
+        }
+
+        ctx.throw_type_error("'this' is not a boolean")?;
+        unreachable!();
```
Razican, line 49:
```suggestion
        Err(ctx.throw_type_error("'this' is not a boolean").expect_err("throw_type_error() did not return an error"))
```

boa/src/builtins/boolean/mod.rs
```diff
@@ -130,6 +113,11 @@ impl Boolean {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("boolean", "init");
-        global.set_field("Boolean", Self::create(global));
+
+        let boolean = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
```
Razican, line 120: I would use `expect()` here.

boa/src/builtins/console/mod.rs
```diff
@@ -556,5 +556,10 @@ pub fn create(global: &Value) -> Value {
 #[inline]
 pub fn init(global: &Value) {
     let _timer = BoaProfiler::global().start_event("console", "init");
-    global.set_field("console", create(global));
+
+    let console = create(global);
+    global
+        .as_object_mut()
+        .unwrap()
```
Razican, line 563: Once again, I think `expect()` gives better insights in case of panic.

boa/src/builtins/error/mod.rs
```diff
@@ -83,6 +83,8 @@ impl Error {
     /// Initialise the global object with the `Error` object.
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("error", "init");
-        global.set_field("Error", Self::create(global));
+
+        let error = Self::create(global);
+        global.as_object_mut().unwrap().insert_field("Error", error);
```
Razican, line 88: `unwrap()` -> `expect()`

boa/src/builtins/error/range.rs
```diff
@@ -73,6 +73,11 @@ impl RangeError {
     /// Initialise the global object with the `RangeError` object.
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("rangeerror", "init");
-        global.set_field("RangeError", Self::create(global));
+
+        let range_error = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
```
Razican, line 80: `unwrap()` -> `expect()`

boa/src/builtins/function/mod.rs
```diff
@@ -390,43 +405,49 @@ pub fn make_constructor_fn(
...
         .configurable(false)
         .enumerable(false);
-    constructor_val.set_property_slice("name", name);
+    constructor.insert_property("name", name);
+
+    let constructor = Value::from(constructor);
+
+    prototype
+        .as_object_mut()
+        .unwrap()
```
Razican, line 442: `unwrap()` -> `expect()`

boa/src/builtins/function/mod.rs
```diff
@@ -390,43 +405,49 @@ pub fn make_constructor_fn(
...
+        .insert_field("constructor", constructor.clone());
 
-    constructor_val
+    constructor
+        .as_object_mut()
+        .unwrap()
```
Razican, line 447: `unwrap()` -> `expect()`

boa/src/builtins/function/mod.rs
```diff
@@ -451,23 +472,26 @@ pub fn make_builtin_fn<N>(function: NativeFunctionData, name: N, parent: &Value,
...
 
-    let new_func_obj = Value::from(new_func);
-    new_func_obj.set_field("length", length);
+    let mut function = Object::function(Function::builtin(Vec::new(), function));
+    function.insert_field("length", Value::from(length));
 
-    parent.set_field(Value::from(name_copy), new_func_obj);
+    parent
+        .as_object_mut()
+        .unwrap()
```
Razican, line 483: `unwrap()` -> `expect()`

boa/src/builtins/function/mod.rs
```diff
@@ -451,23 +472,26 @@ pub fn make_builtin_fn<N>(function: NativeFunctionData, name: N, parent: &Value,
...
 /// Initialise the `Function` object on the global object.
 #[inline]
 pub fn init(global: &Value) {
     let _timer = BoaProfiler::global().start_event("function", "init");
-    global.set_field("Function", create(global));
+
+    let function = create(global);
+    global
+        .as_object_mut()
+        .unwrap()
```
Razican, line 495: `unwrap()` -> `expect()`

boa/src/builtins/math/mod.rs
```diff
@@ -511,14 +511,17 @@ pub fn create(global: &Value) -> Value {
...
-    math.set_field("E", Value::from(f64::consts::E));
-    math.set_field("LN2", Value::from(f64::consts::LN_2));
-    math.set_field("LN10", Value::from(f64::consts::LN_10));
-    math.set_field("LOG2E", Value::from(f64::consts::LOG2_E));
-    math.set_field("LOG10E", Value::from(f64::consts::LOG10_E));
-    math.set_field("SQRT1_2", Value::from(0.5_f64.sqrt()));
-    math.set_field("SQRT2", Value::from(f64::consts::SQRT_2));
-    math.set_field("PI", Value::from(f64::consts::PI));
+    {
+        let mut properties = math.as_object_mut().unwrap();
```
Razican, line 515: `unwrap()` -> `expect()`

boa/src/builtins/mod.rs
```diff
@@ -21,28 +21,29 @@ pub(crate) use self::{
...
     math::init(global);
     Number::init(global);
     object::init(global);
-    function::init(global);
     RegExp::init(global);
     String::init(global);
-    symbol::init(global);
+    Symbol::init(global);
     console::init(global);
+
```
Razican, line 46:
```suggestion

    // Global error types
```

boa/src/builtins/number/mod.rs
```diff
@@ -47,19 +44,22 @@ impl Number {
...
                     Value::from(0)
                 }
             }
             ValueData::Symbol(_) | ValueData::Undefined => Value::from(f64::NAN),
-            ValueData::Integer(i) => Value::from(f64::from(i)),
-            ValueData::Object(ref o) => (o).deref().borrow().get_internal_slot("NumberData"),
+            ValueData::Integer(i) => Value::from(f64::from(*i)),
+            ValueData::Object(ref o) => match (o).deref().borrow().data {
+                ObjectData::Number(num) => Value::from(num),
+                _ => unreachable!(),
```
Razican, line 59: I would add a small text in the `unreachable!("")` macro for the panic, in case it ever happens.

boa/src/builtins/number/mod.rs
```diff
@@ -556,14 +554,17 @@ impl Number {
...
-        number.set_field("EPSILON", Value::from(f64::EPSILON));
-        number.set_field("MAX_SAFE_INTEGER", Value::from(9_007_199_254_740_991_f64));
-        number.set_field("MIN_SAFE_INTEGER", Value::from(-9_007_199_254_740_991_f64));
-        number.set_field("MAX_VALUE", Value::from(f64::MAX));
-        number.set_field("MIN_VALUE", Value::from(f64::MIN));
-        number.set_field("NEGATIVE_INFINITY", Value::from(f64::NEG_INFINITY));
-        number.set_field("POSITIVE_INFINITY", Value::from(f64::INFINITY));
-        number.set_field("NaN", Value::from(f64::NAN));
+        {
+            let mut properties = number.as_object_mut().unwrap();
```
Razican, line 558: `unwrap()` -> `expect()`

boa/src/builtins/number/mod.rs
```diff
@@ -572,7 +573,12 @@ impl Number {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("number", "init");
-        global.set_field("Number", Self::create(global));
+
+        let number = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
```
Razican, line 580: `unwrap()` -> `expect()`

boa/src/builtins/object/internal_methods.rs
```diff
@@ -57,7 +46,8 @@ pub trait ObjectInternalMethods {
     ///  - [ECMAScript reference][spec]
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-isextensible
-    fn is_extensible(&self) -> bool {
+    #[inline]
+    pub fn is_extensible(&self) -> bool {
```
Razican, line 50: Hmmm do we need this to be public outside of the `boa` crate? Same with other functions here. Maybe this is more in line with #445, though.

boa/src/builtins/regexp/mod.rs
```diff
@@ -472,7 +470,10 @@ impl RegExp {
     pub(crate) fn create(global: &Value) -> Value {
         // Create prototype
         let prototype = Value::new_object(Some(global));
-        prototype.set_field("lastIndex", Value::from(0));
+        prototype
+            .as_object_mut()
+            .unwrap()
```
Razican, line 475: `unwrap()` -> `expect()`

boa/src/builtins/regexp/mod.rs
```diff
@@ -493,6 +494,10 @@ impl RegExp {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("regexp", "init");
-        global.set_field("RegExp", Self::create(global));
+        let regexp = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
```
Razican, line 500: `unwrap()` -> `expect()`

boa/src/builtins/string/mod.rs
```diff
@@ -36,46 +36,52 @@ use std::{
...
+                let object = object.borrow();
+                if let Some(string) = object.as_string() {
+                    return Ok(string.clone());
+                }
+            }
+            _ => {}
+        }
+
+        ctx.throw_type_error("'this' is not a string")?;
+        unreachable!();
```
Razican, line 52:
```suggestion
        Err(ctx.throw_type_error("'this' is not a string").expect_err("throw_type_error() did not return an error"))
```

boa/src/builtins/string/mod.rs
```diff
@@ -36,46 +36,52 @@ use std::{
...
         // This value is used by console.log and other routines to match Obexpecty"failed to parse argument for String method"pe
         // to its Javascript Identifier (global constructor method name)
-        let s = args.get(0).unwrap_or(&Value::string("")).clone();
-        let length_str = s.to_string().chars().count();
-
-        this.set_field("length", Value::from(length_str as i32));
+        let string = match args.get(0) {
+            Some(ref value) => ctx.to_string(value)?,
+            None => StdString::new(),
+        };
```
Razican, line 69:
```suggestion
        let string = match args.get(0).map(|value| ctx.to_string(value)).unwrap_or_else(|| Ok(StdString::new()))?;
```

boa/src/builtins/string/mod.rs
```diff
@@ -36,46 +36,52 @@ use std::{
...
-        }
+        this.set_data(ObjectData::String(string.clone()));
 
-        Ok(Value::from(arg.to_string()))
+        Ok(Value::from(string))
     }
 
     /// Get the string value to a primitive string
     #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_string(this: &mut Value, _: &[Value], ctx: &mut Interpreter) -> ResultValue {
```
Razican, line 82: This should only affect debug builds, but I think it makes sense.
```suggestion
    #[allow(clippy::wrong_self_convention)]
    #[inline]
    pub(crate) fn to_string(this: &mut Value, _: &[Value], ctx: &mut Interpreter) -> ResultValue {
```

boa/src/builtins/string/mod.rs
```diff
@@ -1079,6 +1087,10 @@ impl String {
     #[inline]
     pub(crate) fn init(global: &Value) {
         let _timer = BoaProfiler::global().start_event("string", "init");
-        global.set_field("String", Self::create(global));
+        let string = Self::create(global);
+        global
+            .as_object_mut()
+            .unwrap()
```
Razican, line 1093: `unwrap()` -> `expect()`

boa/src/builtins/string/mod.rs
HalidOdat: we can't use the `?` try operator in lambdas.

boa/src/builtins/object/internal_methods.rs
HalidOdat: Should I make it `pub(crate)`?

boa/src/builtins/array/mod.rs
HalidOdat: Made init functions to return a `(&str, Value)` and then just insert in global object. [https://github.com/boa-dev/boa/pull/419/commits/85791b85bc54104c1e7e56b4f18342e034aa1e82 ](https://github.com/boa-dev/boa/pull/419/commits/85791b85bc54104c1e7e56b4f18342e034aa1e82)

boa/src/builtins/number/mod.rs
HalidOdat: That was not spec compliant, made it more spec compliant here [https://github.com/boa-dev/boa/pull/419/commits/49e104b755f321bac6c4d636f74497190a827318 ](https://github.com/boa-dev/boa/pull/419/commits/49e104b755f321bac6c4d636f74497190a827318)

boa/src/builtins/mod.rs
```diff
@@ -22,30 +22,42 @@ pub(crate) use self::{
...
+        String::init(global),
+        Symbol::init(global),
+        console::init(global),
+        // Global error types.
+        Error::init(global),
+        RangeError::init(global),
+        TypeError::init(global),
+    ];
+
+    let mut global_object = global.as_object_mut().expect("global object");
```
Razican, line 59: I'm guessing that the `global` object is always an object, right? So why don't we just pass the "object" around instead of the `Value`?

boa/src/builtins/mod.rs
```diff
@@ -22,30 +22,42 @@ pub(crate) use self::{
...
-    object::init(global);
-    function::init(global);
-    RegExp::init(global);
-    String::init(global);
-    symbol::init(global);
-    console::init(global);
-    Error::init(global);
-    RangeError::init(global);
-    TypeError::init(global);
+    let globals = vec![
```
Razican, line 38: No need to put this in the heap, right? :)
```suggestion
    let globals = [
```

boa/src/builtins/math/mod.rs
```diff
@@ -19,542 +19,550 @@ use crate::{
...
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-math.abs
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/abs
+    pub(crate) fn abs(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+        Ok(Value::from(if args.is_empty() {
+            f64::NAN
+        } else {
+            f64::from(args.get(0).expect("Could not get argument")).abs()
+        }))
```
Razican, line 47: I think this is more clear, right? We also avoid one panic path:
```suggestion
        Ok(args.get(0).map(|val| f64::from(val).abs()).unwrap_or(f64::NAN).into())
```

boa/src/builtins/math/mod.rs
```diff
@@ -19,542 +19,550 @@ use crate::{
...
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-math.acos
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/acos
+    pub(crate) fn acos(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+        Ok(Value::from(if args.is_empty() {
+            f64::NAN
+        } else {
+            f64::from(args.get(0).expect("Could not get argument")).acos()
+        }))
```
Razican, line 63: I think we should do this in all functions in `math`:
```suggestion
        Ok(args.get(0).map(|val| f64::from(val).acos().unwrap_or(f64::NAN).into())
```

boa/src/builtins/object/internal_methods.rs
Razican: Hmmm let's leave it like this for now and re-visit it when we revamp the API.

boa/src/builtins/object/internal_methods.rs
```diff
@@ -282,25 +280,162 @@ pub trait ObjectInternalMethods {
...
+                            d.writable = v.writable;
+                        } else {
+                            debug_assert!(v.is_accessor_descriptor());
+                            d.get = v.get.clone();
+                            d.set = v.set.clone();
+                        }
+                        d.enumerable = v.enumerable;
+                        d.configurable = v.configurable;
+                        d
+                    }
```
Razican, line 315: Do you think a `map(||...).unwrap_or_else(Property::default)` would be easier to understand? or an `if let`?

boa/src/builtins/object/mod.rs
```diff
@@ -47,322 +46,85 @@ pub static PROTOTYPE: &str = "prototype";
...
-        write!(f, "}}")
-    }
+/// Defines the different types of objects.
+#[derive(Debug, Trace, Finalize, Clone)]
+pub enum ObjectData {
+    Array,
+    BigInt(BigInt),
+    Boolean(bool),
+    Function(Function),
+    String(String),
```
Razican, line 70: Is a string mutable? If not, we can probably use `Box<str>` here.

boa/src/builtins/string/mod.rs
Razican: But in this case, the `?` operator would be outside of the lambda, right?

boa/src/exec/mod.rs
```diff
@@ -477,6 +529,25 @@ impl Interpreter {
...
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [table]: https://tc39.es/ecma262/#table-14
+    /// [spec]: https://tc39.es/ecma262/#sec-requireobjectcoercible
+    pub fn require_object_coercible<'a>(&mut self, value: &'a Value) -> Result<&'a Value, Value> {
+        if value.is_null_or_undefined() {
+            self.throw_type_error("cannot convert null or undefined to Object")?;
+            unreachable!();
```
Razican, line 547: I would use `expect_err()` too.

boa/src/builtins/symbol/mod.rs
```diff
@@ -20,87 +20,95 @@ mod tests;
...
+            }
+            _ => {}
+        }
 
-    Ok(Value(Gc::new(ValueData::Symbol(Box::new(GcCell::new(
-        sym_instance,
-    ))))))
-}
+        ctx.throw_type_error("'this' is not a Symbol")?;
+        unreachable!();
```
Razican, line 62: I would use `expect_err()` here.

boa/src/builtins/value/display.rs
```diff
@@ -124,14 +111,10 @@ pub(crate) fn log_string_from(x: &ValueData, print_internals: bool) -> String {
...
-            match *desc {
-                ValueData::String(ref st) => format!("Symbol(\"{}\")", st.to_string()),
-                _ => String::from("Symbol()"),
-            }
-        }
-
+        ValueData::Symbol(ref symbol) => match symbol.description() {
+            Some(description) => format!("Symbol({})", description),
+            None => String::from("Symbol()"),
+        },
```
Razican, line 117: Do you think that this would be easier to understand?
```rust
symbol.description.map(|desc| format!("Symbol({})", desc)).unwrap_or_else(||String::from("Symbol()"))
```

boa/src/builtins/value/mod.rs
```diff
@@ -558,20 +554,16 @@ impl ValueData {
...
+            .as_object()
+            .and_then(|x| match x.internal_slots().get(field) {
+                Some(value) => Some(value.clone()),
+                None => None,
+            });
+
+        match property {
+            Some(value) => value,
             None => Value::undefined(),
         }
```
Razican, line 568: I think this is more clear:
```suggestion
        property.unwrap_or_else(Value::undefined)
```

boa/src/builtins/value/mod.rs
```diff
@@ -616,19 +608,17 @@ impl ValueData {
...
     /// Check whether an object has an internal state set.
     pub fn has_internal_state(&self) -> bool {
-        if let Self::Object(ref obj) = *self {
-            obj.borrow().state.is_some()
-        } else {
-            false
+        match self.as_object() {
+            Some(object) => object.state().is_some(),
+            None => false,
         }
```
Razican, line 614: I'm not sure if a `.map().unwrap_or()` would be easier to understand, what do you think?

boa/src/exec/mod.rs
```diff
@@ -250,13 +259,13 @@ impl Interpreter {
...
-        match *value.deref().borrow() {
+        match *value.data() {
             ValueData::Null => Ok(0.0),
             ValueData::Undefined => Ok(f64::NAN),
             ValueData::Boolean(b) => Ok(if b { 1.0 } else { 0.0 }),
-            ValueData::String(ref string) => match string.parse::<f64>() {
+            ValueData::String(ref string) => match string.parse() {
                 Ok(number) => Ok(number),
-                Err(_) => Ok(0.0),
+                Err(_) => Ok(f64::NAN),
```
Razican, line 268: Here, I would use this:
```rust
string.parse().unwrap_or(f64::NAN)
```

boa/src/builtins/string/mod.rs
HalidOdat: > But in this case, the `?` operator would be outside of the lambda, right?

I guess so, but i think it becomes less readable in this case, having all the logic in one line.

boa/src/builtins/mod.rs
HalidOdat: Hmmm. That's a good point! I will see what I can do :)

boa/src/builtins/mod.rs
HalidOdat: There is a problem with using slices it does not implement `IntoIterator` trait, it implements `IntoIterator` for `&[_]` but this will give me a reference to a `Value`, we don't don't want to copy the value.

boa/src/builtins/object/mod.rs
HalidOdat: Yes. exactly I was planing of doing this storage optimization in #465 :)

boa/src/exec/mod.rs
HalidOdat: Hmmm. I think we should have a version that just constructs the error, maybe something like `construct_type_error` this would return a `Value` so we don't have to do this weird thing with `unreachable!()`

So we would do:
```rust
fn some_function(&self) -> Result<AnyNonValueType, Value> {
	// ... logic here ... 
	
	Err(self.construct_type_error("some message")) 
}
```

But when we have a `ResultValue` we use the `throw_type_error` method

What do you think? @Razican

boa/src/builtins/math/mod.rs
HalidOdat: Hmmm. why don't we use `map_or` and `map_or_else`

so this would be:
```rust
Ok(args.get(0).map_or(f64::NAN, |val| f64::from(val).abs()).into())
```
I think in this case its more readable than the first and second option.

What do you think? @Razican

boa/src/builtins/value/display.rs
HalidOdat:
```rust
symbol.description.map_or_else(|| String::from("Symbol()"), |desc| format!("Symbol({})", desc))
```
I think with `map_or_else` its much better, right?

boa/src/builtins/value/mod.rs
HalidOdat: we could use `std::matches!()` macro

boa/src/builtins/object/internal_methods.rs
HalidOdat:
```rust
self.properties().get(st).map_or_else(Property::default, || {
	// ...
})
```
This is with `map_or_else`

boa/src/builtins/value/mod.rs
HalidOdat: In this case it's much cleaner :)

boa/src/builtins/math/mod.rs
HalidOdat: I think this would be some good `good fist issue`s

boa/src/builtins/mod.rs
Razican: But we can call `into_iter()` on an array, right?

boa/src/builtins/math/mod.rs
Razican: Yes, these functions are more clear :)

boa/src/exec/mod.rs
Razican: Yes, definitely, this looks much nicer :)

boa/src/builtins/value/display.rs
Razican: It should s, yes :)

boa/src/builtins/value/mod.rs
Razican: > we could use `matches!()` macro

Yes!

boa/src/builtins/mod.rs
HalidOdat: > But we can call `into_iter()` on an array, right?

Sadly `into_iter` for arrays is the same as `iter` which yields a reference to the value, not the value itself

Razican: And what about doing an array of functions and then calling them in the for?

HalidOdat: > And what about doing an array of functions and then calling them in the for?

So, function pointers that could work :)

HalidOdat: `rustc` on release mode with optimizations can remove the heap allocation of the `Vec` and store it in the stack

boa/src/builtins/object/internal_methods.rs
```diff
@@ -282,25 +280,162 @@ pub trait ObjectInternalMethods {
...
-    /// Utility function to get an immutable internal slot or Null
-    fn get_internal_slot(&self, name: &str) -> Value;
+    /// Helper function to get an immutable internal slot or `Null`.
+    #[inline]
+    pub fn get_internal_slot(&self, name: &str) -> Value {
+        let _timer = BoaProfiler::global().start_event("Object::get_internal_slot", "object");
+        match self.internal_slots.get(name) {
+            Some(v) => v.clone(),
+            None => Value::null(),
+        }
```
Razican, line 398: What about:
```suggestion
        self.internal_slots.get(name).cloned().unwrap_or_else(Value::null)
```

boa/src/builtins/object/internal_methods.rs
```diff
@@ -282,25 +280,162 @@ pub trait ObjectInternalMethods {
...
 
-    fn remove_property(&mut self, name: &str);
+    /// Helper function for property removal.
+    #[inline]
+    pub fn remove_property(&mut self, name: &str) {
+        self.properties.remove(name);
+    }
+
+    #[inline]
+    pub fn insert_field<N>(&mut self, name: N, value: Value) -> Option<Property>
```
Razican, line 423: I think we should add some documentation here.

boa/src/builtins/object/internal_methods.rs
```diff
@@ -282,25 +280,162 @@ pub trait ObjectInternalMethods {
...
+            Property::default()
+                .value(value)
+                .writable(true)
+                .configurable(true)
+                .enumerable(true),
+        )
+    }
+
+    #[inline]
+    pub fn get_field(&self, name: &str) -> Option<&Value> {
```
Razican, line 438: I think we should add some documentation here.

boa/src/builtins/object/mod.rs
```diff
@@ -47,322 +46,85 @@ pub static PROTOTYPE: &str = "prototype";
...
 
         object.set_internal_slot("extensible", Value::from(true));
         object
     }
+}
+
+impl Object {
+    pub fn new() -> Self {
+        Default::default()
+    }
```
Razican, line 116: I wouldn't add a `new()` function that is just a `default()`.

boa/src/builtins/object/mod.rs
```diff
@@ -490,58 +226,168 @@ impl Object {
...
-                Self::Symbol => "Symbol",
-                Self::Error => "Error",
-                Self::Ordinary => "Ordinary",
-                Self::Boolean => "Boolean",
-                Self::Number => "Number",
-                Self::BigInt => "BigInt",
-            }
-        )
+    #[inline]
+    pub fn as_string(&self) -> Option<&String> {
```
Razican, line 255: I think we should use `&str` here, instead of `&String`. This abstracts over the underlying implementation, that can be easily swapped by a `Box<str>` in the future. For that, we need to use `string.as_str()` later.

boa/src/builtins/value/hash.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl PartialEq for Value {
+    fn eq(&self, other: &Self) -> bool {
+        same_value_zero(self, other)
+    }
+}
+
+impl Eq for Value {}
+
+#[derive(PartialEq, Eq, Hash)]
```
Razican, line 14:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```

boa/src/builtins/value/hash.rs
```diff
@@ -0,0 +1,53 @@
...
+struct UndefinedHashable;
+
+#[derive(PartialEq, Eq, Hash)]
```
Razican, line 17:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```

boa/src/builtins/value/hash.rs
```diff
@@ -0,0 +1,53 @@
...
+struct NullHashable;
+
+struct RationalHashable(f64);
```
Razican, line 20:
```suggestion
#[derive(Debug, Clone, Copy)]
struct RationalHashable(f64);
```

boa/src/builtins/value/hash.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::builtins::Number;
+use std::hash::{Hash, Hasher};
+
+impl PartialEq for Value {
+    fn eq(&self, other: &Self) -> bool {
+        same_value_zero(self, other)
+    }
+}
+
+impl Eq for Value {}
```
Razican, line 12: This could be derived, right?

boa/src/builtins/value/hash.rs
```diff
@@ -0,0 +1,53 @@
...
+struct RationalHashable(f64);
+
+impl PartialEq for RationalHashable {
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        Number::same_value(self.0, other.0)
+    }
+}
+
+impl Eq for RationalHashable {}
```
Razican, line 29: This could be derived, right?

boa/src/builtins/value/mod.rs
```diff
@@ -374,41 +385,40 @@ impl ValueData {
...
+            Self::BigInt(bigint) => Some(bigint),
+            _ => None,
         }
     }
 
-    /// Returns true if the value is true
+    /// Returns true if the value is true.
     ///
-    /// [toBoolean](https://tc39.es/ecma262/#sec-toboolean)
+    /// [toBoolean](https://tc39.es/ecma262/#sec-toboolean
```
Razican, line 421:
```suggestion
    /// [toBoolean](https://tc39.es/ecma262/#sec-toboolean)
```

boa/src/builtins/value/mod.rs
```diff
@@ -492,36 +506,22 @@ impl ValueData {
...
-                let hash = obj.clone();
-                hash.into_inner()
+            Self::Object(ref object) => {
+                let object = object.borrow();
+                match object.properties().get(field) {
+                    Some(value) => Some(value.clone()),
+                    None => match object.internal_slots().get(INSTANCE_PROTOTYPE) {
+                        Some(value) => value.get_property(field),
+                        None => None,
+                    },
```
Razican, line 521:
```suggestion
                    None => object.internal_slots().get(INSTANCE_PROTOTYPE).map(|value| value.get_property(field)),
```

boa/src/builtins/value/mod.rs
```diff
@@ -492,36 +506,22 @@ impl ValueData {
...
+            Self::Undefined => None,
+            Self::String(ref s) if field == "length" => {
+                Some(Property::default().value(Value::from(s.chars().count())))
             }
-            Self::Symbol(ref obj) => {
-                let hash = obj.clone();
-                hash.into_inner()
+            Self::Object(ref object) => {
+                let object = object.borrow();
+                match object.properties().get(field) {
```
Razican, line 516: Here we could also use `.cloned().unwrap_or_else()`. What do you think?

boa/src/builtins/value/mod.rs
```diff
@@ -558,22 +554,15 @@ impl ValueData {
...
-        match obj.internal_slots.get(field) {
-            Some(val) => val.clone(),
-            None => Value::undefined(),
-        }
+        let property = self
+            .as_object()
+            .and_then(|x| match x.internal_slots().get(field) {
+                Some(value) => Some(value.clone()),
+                None => None,
+            });
```
Razican, line 563:
```suggestion
            .and_then(|x| x.internal_slots().get(field).cloned());
```

boa/src/builtins/value/mod.rs
```diff
@@ -616,19 +605,14 @@ impl ValueData {
...
     /// Get the internal state of an object.
     pub fn get_internal_state(&self) -> Option<InternalStateCell> {
-        if let Self::Object(ref obj) = *self {
-            obj.borrow().state.as_ref().cloned()
-        } else {
-            None
+        match self.as_object() {
+            Some(object) => object.state().clone(),
+            None => None,
         }
```
Razican, line 616:
```suggestion
        self.as_object().map(|object| object.state().clone())
```

boa/src/exec/mod.rs
```diff
@@ -87,6 +89,12 @@ impl Interpreter {
         &mut self.realm
     }
 
+    pub(crate) fn generate_hash(&mut self) -> u32 {
```
Razican, line 92: We should add some documentation here.

boa/src/builtins/value/hash.rs
HalidOdat: Sadly it can't `#[derive(Eq)]` requires that its fields have implemented `Eq` and `f64` does not.

boa/src/builtins/value/hash.rs
HalidOdat: Sadly it can't #[derive(Eq)] requires that its fields have implemented Eq and f64 does not. (`ValueData::Number(f64)`)

boa/src/builtins/object/mod.rs
HalidOdat: This was taken from the `Rust API guideline` in `Predictability` section `Constructors are static, inherent methods (C-CTOR)` 

> Note that it is common and expected for types to implement both Default and a new constructor. For types that have both, they should have the same behavior. Either one may be implemented in terms of the other.

I think eventually will have to remove them, there should because it does not create an object with the prototype, also the rule only applies if its a part of the api.

boa/src/builtins/object/mod.rs
```diff
@@ -47,322 +46,85 @@ pub static PROTOTYPE: &str = "prototype";
...
-        for (key, _) in self.properties.iter() {
-            writeln!(f, "\t\t{}", key)?;
-        }
-        writeln!(f, "\t }}")?;
-        write!(f, "}}")
-    }
+/// Defines the different types of objects.
+#[derive(Debug, Trace, Finalize, Clone)]
+pub enum ObjectData {
+    Array,
```
jasonwilliams, line 66: Just for context. why is it called `ObjectData`? 
Looks like it's still used for Type definition.

boa/src/builtins/object/mod.rs
jasonwilliams: Yeah i think this should just implement `Default` and we move to using that

boa/src/builtins/array/mod.rs
```diff
@@ -33,6 +33,12 @@ use std::{
 pub(crate) struct Array;
 
 impl Array {
+    /// The name of the object.
+    pub(crate) const NAME: &'static str = "Array";
+
+    /// The amount of arguments this function object takes.
+    pub(crate) const LENGTH: i32 = 1;
```
jasonwilliams, line 40: How come an i32 is chosen for the length?
Is there a reason it's signed?

boa/src/builtins/object/internal_methods.rs
jasonwilliams: Yeah i don't think we need to change this right now

boa/src/builtins/array/mod.rs
HalidOdat: It's because `make_constructor_fn` and `make_builtin_fn` take a i32 as length, I think I'll change them to take in a `usize` so we don't have this problem

boa/src/builtins/object/mod.rs
HalidOdat: Because it also holds data its not just type definitions, it's like `ValuData`, the doc comment is misleading,

boa/src/builtins/mod.rs
```diff
@@ -22,30 +22,42 @@ pub(crate) use self::{
...
     number::Number,
     regexp::RegExp,
     string::String,
+    symbol::Symbol,
     value::{ResultValue, Value},
 };
 
 /// Initializes builtin objects and functions
 #[inline]
 pub fn init(global: &Value) {
```
Razican, line 37: I think this should be mostly easy to implement:
```suggestion
pub fn init(global: &mut Object) {
```

We would need to use `Object::default()` in the Realm constructor [here](https://github.com/boa-dev/boa/blob/master/boa/src/realm.rs#L35-L54). And then implement `From<Object>` for `Value`.

Finally, I would change the `new_object()` function [here](https://github.com/boa-dev/boa/blob/73d7a64bb4da6040e3424297082d47debb3c91b8/boa/src/builtins/value/mod.rs#L135-L145) to not accept an `Option`, but to require always a global object.

This would allow implementing initialization with a very fast implementation, even doing it multi-threaded at some point.

Then, I think that we should create a trait or a type with a function that receives a `&mut Object` as a parameter and returns a `(&'static str, Value)`, this would help with creating an external API.

boa/src/builtins/mod.rs
Razican: Check my comment above, I think this could be solved by not using the Gc'd object, and just using the object.

boa/src/builtins/object/internal_methods.rs
Razican: We should do this, right?

boa/src/builtins/object/mod.rs
Razican: I don't know, `clippy` seems to recommend not having a `new()` function without arguents, and using a `default()` instead. In any case, I think we should mark it as `#[inline]`.

boa/src/builtins/value/mod.rs
```diff
@@ -558,22 +554,10 @@ impl ValueData {
     /// Returns a copy of the Property.
     pub fn get_internal_slot(&self, field: &str) -> Value {
```
Razican, line 555:
```suggestion
    #[inline]
    pub fn get_internal_slot(&self, field: &str) -> Value {
```

boa/src/builtins/value/mod.rs
```diff
@@ -616,20 +600,13 @@ impl ValueData {
 
     /// Check whether an object has an internal state set.
     pub fn has_internal_state(&self) -> bool {
```
Razican, line 602:
```suggestion
    #[inline]
    pub fn has_internal_state(&self) -> bool {
```

boa/src/builtins/value/mod.rs
```diff
@@ -616,20 +600,13 @@ impl ValueData {
...
-        if let Self::Object(ref obj) = *self {
-            obj.borrow().state.is_some()
-        } else {
-            false
-        }
+        matches!(self.as_object(), Some(object) if object.state().is_some())
     }
 
     /// Get the internal state of an object.
     pub fn get_internal_state(&self) -> Option<InternalStateCell> {
```
Razican, line 607:
```suggestion
    #[inline]
    pub fn get_internal_state(&self) -> Option<InternalStateCell> {
```

boa/src/builtins/value/mod.rs
```diff
@@ -722,53 +699,49 @@ impl ValueData {
...
+            (*obj.deref().borrow_mut()).data = data;
         }
     }
 
-    /// Set the property in the value
-    pub fn set_property(&self, field: String, prop: Property) -> Property {
-        if let Self::Object(ref obj) = *self {
-            obj.borrow_mut().properties.insert(field, prop.clone());
+    /// Set the property in the value.
+    pub fn set_property<S>(&self, field: S, property: Property) -> Property
```
Razican, line 720:
```suggestion
    #[inline]
    pub fn set_property<S>(&self, field: S, property: Property) -> Property
```

boa/src/builtins/value/mod.rs
```diff
@@ -722,53 +699,49 @@ impl ValueData {
...
-    }
-
-    /// Set the property in the value
-    pub fn set_property_slice(&self, field: &str, prop: Property) -> Property {
-        self.set_property(field.to_string(), prop)
+        property
     }
 
     /// Set internal state of an Object. Discards the previous state if it was set.
     pub fn set_internal_state<T: Any + InternalState>(&self, state: T) {
```
Razican, line 733:
```suggestion
    #[inline]
    pub fn set_internal_state<T: Any + InternalState>(&self, state: T) {
```

boa/src/builtins/value/tests.rs
```diff
@@ -91,6 +94,66 @@ fn abstract_equality_comparison() {
...
+    assert_eq!(
+        forward(
+            &mut engine,
+            "Number.NEGAVIVE_INFINITY === Number.NEGAVIVE_INFINITY"
+        ),
+        "true"
+    );
+}
+
+fn hash_value(value: &Value) -> u64 {
```
Razican, line 114:
```suggestion
/// Helper function to get the hash of a `Value`.
fn hash_value(value: &Value) -> u64 {
```

boa/src/exec/mod.rs
```diff
@@ -87,6 +89,15 @@ impl Interpreter {
         &mut self.realm
     }
 
+    /// Generates a new `Symbol` internal hash.
+    ///
+    /// This currently is an incremented value.
+    pub(crate) fn generate_hash(&mut self) -> u32 {
```
Razican, line 95:
```suggestion
    #[inline]
    pub(crate) fn generate_hash(&mut self) -> u32 {
```

boa/src/exec/mod.rs
```diff
@@ -494,11 +549,29 @@ impl Interpreter {
     }
 
     pub(crate) fn set_current_state(&mut self, new_state: InterpreterState) {
```
Razican, line 551:
```suggestion
    #[inline]
    pub(crate) fn set_current_state(&mut self, new_state: InterpreterState) {
```

boa/src/exec/mod.rs
```diff
@@ -494,11 +549,29 @@ impl Interpreter {
...
-        self.current_state = new_state
+        self.state = new_state
     }
 
     pub(crate) fn get_current_state(&self) -> &InterpreterState {
```
Razican, line 555:
```suggestion
    #[inline]
    pub(crate) fn get_current_state(&self) -> &InterpreterState {
```

boa/src/exec/mod.rs
```diff
@@ -494,11 +549,29 @@ impl Interpreter {
...
-        &self.current_state
+        &self.state
+    }
+    /// Check if the `Value` can be converted to an `Object`
```
Razican, line 558:
```suggestion

    /// Check if the `Value` can be converted to an `Object`
```

boa/src/exec/mod.rs
Razican: Should we do this in this PR or in a different one?

boa/src/exec/mod.rs
```diff
@@ -494,11 +549,29 @@ impl Interpreter {
...
+    /// The abstract operation `RequireObjectCoercible` takes argument argument.
+    /// It throws an error if argument is a value that cannot be converted to an Object using `ToObject`.
+    /// It is defined by [Table 15][table]
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [table]: https://tc39.es/ecma262/#table-14
+    /// [spec]: https://tc39.es/ecma262/#sec-requireobjectcoercible
+    pub fn require_object_coercible<'a>(&mut self, value: &'a Value) -> Result<&'a Value, Value> {
```
Razican, line 569:
```suggestion
    #[inline]
    pub fn require_object_coercible<'a>(&mut self, value: &'a Value) -> Result<&'a Value, Value> {
```

boa/src/builtins/boolean/mod.rs
```diff
@@ -73,52 +93,22 @@ impl Boolean {
...
-    ///  - [ECMAScript reference][spec]
-    ///
-    /// [spec]: https://tc39.es/ecma262/#sec-thisbooleanvalue
-    pub(crate) fn this_boolean_value(value: &Value) -> Value {
-        match *value.deref().borrow() {
-            ValueData::Boolean(v) => Value::from(v),
-            ValueData::Object(ref v) => (v).deref().borrow().get_internal_slot("BooleanData"),
-            _ => Value::from(false),
-        }
+    pub(crate) fn value_of(this: &mut Value, _: &[Value], ctx: &mut Interpreter) -> ResultValue {
```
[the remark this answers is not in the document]
HalidOdat: Correct me if I'm wrong, but the compiler has no way of inlineing it when its a used as a function pointer the only way it can inline is if you call it directly. I will add it anyway :)

boa/src/builtins/mod.rs
HalidOdat: yes. this would make somethings easier but what if we wanted an object to reference the global object (for example `globalThis`), this would make it really hard, right?

boa/src/builtins/array/mod.rs
```diff
@@ -167,25 +172,9 @@ impl Array {
...
-                    }
-                    // 3.
-                    _ => Ok(value_false),
-                }
-            }
-            None => Ok(value_false),
+        match args.get(0).and_then(|x| x.as_object()) {
+            Some(object) => Ok(Value::from(object.is_array())),
+            None => Ok(Value::from(false)),
         }
```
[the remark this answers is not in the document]
HalidOdat: I'm not sure, to me the first option is more readable than the second one.

boa/src/builtins/object/internal_methods.rs
HalidOdat: sorry forgot about it :sweat_smile: . I'll do it now.

boa/src/exec/mod.rs
HalidOdat: I guess we can do this in this PR.

HalidOdat:
We still don't have benchmarks for the specialized objects :sweat_smile: , Like `Number` object not primitive, the benchmarks for string don't test `String` object only string primitives. although the speed boost is really nice :)

jasonwilliams:
We should do a before and after with the profiler, it would be nice to sss

HalidOdat:
Here is for master branch: https://gist.github.com/HalidOdat/c346ea541830a80f0042a716c6d9ecea
master-profiler

-------------

And for `feature/specialized-objects` branch: https://gist.github.com/HalidOdat/c46f9b6ca1c0e1a00aa29b514b4c3a69
sp-objects

boa/src/builtins/mod.rs
Razican: > yes. this would make somethings easier but what if we wanted an object to reference the global object (for example `globalThis`), this would make it really hard, right?

Hmmm yes, this would be more difficult, like if we want to have `top/window/globalThis`, all pointing to the global object.

I think we can merge it as it is now, and maybe think on a design in the future :)

HalidOdat:
Lets merge this :)

PR #420, opened 2020-05-25
Number constants

### LINKED ISSUE #414: Implement `Number` object constants
**ECMASCript feature**

Implement `Number` object constants:
 - [ ] `Number.EPSILON`
 - [ ] `Number.MAX_SAFE_INTEGER`
 - [ ] `Number.MIN_SAFE_INTEGER`
 - [ ] `Number.MAX_VALUE`
 - [ ] `Number.MIN_VALUE`
 - [ ] `Number.NEGATIVE_INFINITY`
 - [ ] `Number.POSITIVE_INFINITY`
 - [ ] `Number.NaN`

This should be very straightforward, to see how we implement constants on objects check [`Math`][math-implementation] object.

For more information check:
 - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number)
 - [ECMAScript](https://tc39.es/ecma262/#sec-number-object)

[math-implementation]: https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/math/mod.rs#L509

Lan2u:
This Pull Request fixes #414 .

It changes the following:
 - Adds Number builtin constants of EPSILON, MAX_SAFE_INTEGER, MIN_SAFE_INTEGER, MAX_VALUE, MIN_VALUE, NEGATIVE_INFINITY, POSITIVE_INFINITY, NAN.

The constant NaN is expressed currently as NAN because using the lowercase a was causing the parser to interpret it as an identifier not a constant.


boa/src/builtins/number/mod.rs
```diff
@@ -417,7 +417,23 @@ impl Number {
         make_builtin_fn(Self::to_string, "toString", &prototype, 1);
         make_builtin_fn(Self::value_of, "valueOf", &prototype, 0);
 
-        make_constructor_fn(Self::make_number, global, prototype)
+        let number = make_constructor_fn(Self::make_number, global, prototype);
+
+        // Constants from:
+        // Draft ECMA-262 / May 22, 2020 ECMAScript® 2021 Language Specification
+        // Section 20.1.2 Properties of the Number Constructor
+        number.set_field("EPSILON", Value::from(2.220_446_049_250_313_080_847_263_336_181_6E-16_f64));
```
HalidOdat, line 425:
```suggestion
        number.set_field("EPSILON", Value::from(std::f64::EPSILON));
```
Here we can use the native `std::f64::EPSILON`

boa/src/builtins/number/mod.rs
```diff
@@ -417,7 +417,23 @@ impl Number {
...
+        number.set_field("MAX_SAFE_INTEGER", Value::from(9_007_199_254_740_991_f64));
+        number.set_field("MIN_SAFE_INTEGER", Value::from(-9_007_199_254_740_991_f64));
+        number.set_field("MAX_VALUE", Value::from(1.797_693_134_862_315_7E+308_f64));
+        number.set_field("MIN_VALUE", Value::from(5E-324_f64));
```
HalidOdat, line 429:
```suggestion
        number.set_field("MAX_VALUE", Value::from(std::f64::MAX));
        number.set_field("MIN_VALUE", Value::from(std::f64::MIN));
```
We can use `std::f64::MAX` and `std::f64::MIN` here.

boa/src/builtins/number/mod.rs
```diff
@@ -417,7 +417,23 @@ impl Number {
...
+
+        // Strings used for representing the special numerical values, could be potentially replaced with a different type.
+        number.set_field("NEGATIVE_INFINITY", Value::from("-Infinity"));
+        number.set_field("POSITIVE_INFINITY", Value::from("Infinity"));
+        number.set_field("NAN", Value::from("NAN"));
```
HalidOdat, line 434:
```suggestion
        number.set_field("NEGATIVE_INFINITY", Value::from(std::f64::NEG_INFINITY));
        number.set_field("POSITIVE_INFINITY", Value::from(std::f64::INFINITY));
        number.set_field("NAN", Value::from(std::f64::NAN));
```

Here too, we can use the native constants.

Lan2u: I actually made this change before seeing this comment (so not re-committing suggestion here, is included in commit 82e1822).

Lan2u:
> This looks good! Check my comments to see on how we can improve it :)

Thanks for the help, changes made :) 

With regards to Number.NaN vs Number.NAN is this something which should be looked at as part of this or separately? (unless I've made a mistake of course).

HalidOdat:
> With regards to Number.NaN vs Number.NAN is this something which should be looked at as part of this or separately? (unless I've made a mistake of course).

About this, it should be treated separately, it would require changes to the lexer and parser. For now you can comment it out in the Number creation and test and add a `TODO` with an explanation.

boa/src/builtins/number/mod.rs
```diff
@@ -417,7 +417,23 @@ impl Number {
         make_builtin_fn(Self::to_string, "toString", &prototype, 1);
         make_builtin_fn(Self::value_of, "valueOf", &prototype, 0);
 
-        make_constructor_fn(Self::make_number, global, prototype)
+        let number = make_constructor_fn(Self::make_number, global, prototype);
+
+        // Constants from:
+        // Draft ECMA-262 / May 22, 2020 ECMAScript® 2021 Language Specification
+        // Section 20.1.2 Properties of the Number Constructor
```
Razican, line 424: Can we change this for a link?

boa/src/builtins/number/mod.rs
```diff
@@ -417,7 +417,23 @@ impl Number {
...
+        number.set_field("EPSILON", Value::from(std::f64::EPSILON));
+        number.set_field("MAX_SAFE_INTEGER", Value::from(9_007_199_254_740_991_f64));
+        number.set_field("MIN_SAFE_INTEGER", Value::from(-9_007_199_254_740_991_f64));
+        number.set_field("MAX_VALUE", Value::from(std::f64::MAX));
+        number.set_field("MIN_VALUE", Value::from(std::f64::MIN));
+        number.set_field("NEGATIVE_INFINITY", Value::from(f64::NEG_INFINITY));
+        number.set_field("POSITIVE_INFINITY", Value::from(f64::INFINITY));
+
+        // TODO: Currently parsing/lexing Number.NaN is not supported due to the inclusion of the lower case 'a'.
+        // number.set_field("NaN", Value::from(f64::NAN));
```
Razican, line 434: I would still add this here, so that when we fix the lexer this will just work.

HalidOdat: > I would still add this here

Yep. We could add it! That's probably the best option.

boa/src/builtins/number/mod.rs
Lan2u: Removed section and added link

boa/src/builtins/number/mod.rs
Lan2u: Will add it back :)

Razican: I created #421 related to this.

PR #422, opened 2020-05-26
Divide the `run()` function

Lan2u:
This Pull Request relates to #366.

The following AST nodes have been moved into their own files.
- [x] Break
- [x] ConditionalOp
- [x] Continue
- [x] DoWhileLoop
- [x] GetConstField
- [x] GetField
- [x] If
- [x] Object
- [x] Return
- [x] Switch
- [x] Spread
- [x] Throw
- [x] WhileLoop


boa/src/syntax/ast/node/object.rs
```diff
@@ -0,0 +1,88 @@
...
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#prod-ObjectLiteral
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer
+/// [object]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object
+/// [primitive]: https://developer.mozilla.org/en-US/docs/Glossary/primitive
+
```
Razican, line 31:
```suggestion
```

boa/src/syntax/parser/expression/primary/object_initializer/mod.rs
```diff
@@ -82,7 +82,7 @@ impl TokenParser for ObjectLiteral {
             }
         }
 
-        Ok(Node::object(elements))
+        Ok(Object::new(elements).into())
```
Razican, line 85: I would suggest that this returns an `Object`, instead of a Node. We should only convert to a `Node` once it's required.

boa/src/syntax/ast/node/object.rs
```diff
@@ -0,0 +1,88 @@
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct Object {
+    properties: Box<[PropertyDefinition]>,
+}
+
+impl Object {
+    /// Creates an `Object` AST node.
+    pub fn new<D>(props: D) -> Self
```
Razican, line 41: Maybe we should implement the `From` trait instead of the `new()` function. How do we do it for other "list" type nodes?

boa/src/syntax/parser/expression/primary/object_initializer/mod.rs
Lan2u: Won't it need to be returned as a Node because of the type of ParseResult? (<Node, ParseError>)

boa/src/syntax/ast/node/object.rs
Lan2u: Is the reason for this that From is more useful / can be used in more situations?

boa/src/syntax/parser/expression/primary/object_initializer/mod.rs
Razican: > Won't it need to be returned as a Node because of the type of ParseResult? (<Node, ParseError>)

We can change that by setting the `Output` to an `Object`, and then changing the function signature :)

boa/src/syntax/ast/node/object.rs
Razican: > Is the reason for this that From is more useful / can be used in more situations?

It's just an standard trait, that can help on function signatures making them more generic, if needed.

Lan2u: Ok cool, changed :)

boa/src/syntax/parser/expression/primary/object_initializer/mod.rs
Lan2u: Done

boa/src/exec/switch/mod.rs
```diff
@@ -0,0 +1,34 @@
+use super::{Executable, Interpreter};
+use crate::{
+    builtins::value::{ResultValue, Value},
+    syntax::ast::node::Switch,
+};
+
+impl Executable for Switch {
```
Razican, line 7: If I understand this correctly, this implements the `switch` statement execution, right? This was not previously present, right?

I guess that the only thing left is to `break` when needed, right? Could we add a `// TODO` here?

boa/src/exec/switch/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+            let block = &tup.1;
+            if val.strict_equals(&cond.run(interpreter)?) {
+                matched = true;
+                let last_expr = block.last().expect("Block has no expressions");
+                for expr in block.iter() {
+                    let e_result = expr.run(interpreter)?;
+                    if expr == last_expr {
+                        result = e_result;
+                    }
+                }
```
Razican, line 24: The block itself should be `Executable`, right? so we could just run it.

boa/src/exec/throw/mod.rs
```diff
@@ -0,0 +1,8 @@
+use super::{Executable, Interpreter};
+use crate::{builtins::value::ResultValue, syntax::ast::node::Throw};
+
+impl Executable for Throw {
+    fn run(&self, interpreter: &mut Interpreter) -> ResultValue {
```
Razican, line 5: This is just so that debug builds inline this function, it should be small enough so that it doesn't cause any build delays, and should help.
```suggestion
impl Executable for Throw {
    #[inline]
    fn run(&self, interpreter: &mut Interpreter) -> ResultValue {
```

boa/src/syntax/ast/node/break_node.rs
```diff
@@ -0,0 +1,63 @@
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct Break {
+    label: Option<Box<str>>,
+}
+
+impl Break {
+    pub fn label(&self) -> &Option<Box<str>> {
+        &self.label
+    }
```
Razican, line 31: This should ideally return an `Option<&str>`. I think this will do it:
```suggestion
    pub fn label(&self) -> Option<&str> {
        self.label.as_ref().map(Box::as_ref)
    }
```

boa/src/syntax/ast/node/conditional.rs
```diff
@@ -0,0 +1,151 @@
...
+        &self.cond
+    }
+
+    pub fn body(&self) -> &Node {
+        &self.body
+    }
+
+    pub fn else_node(&self) -> &Option<Box<Node>> {
+        &self.else_node
+    }
```
Razican, line 43: This too should return an `Option<&Node>`:
```suggestion
    pub fn else_node(&self) -> Option<&Node> {
        self.else_node.as_ref().map(Box::as_ref)
    }
```

boa/src/syntax/ast/node/iteration.rs
```diff
@@ -137,3 +137,175 @@ impl InnerForLoop {
...
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct Continue {
+    label: Option<Box<str>>,
+}
+
+impl Continue {
+    pub fn label(&self) -> &Option<Box<str>> {
+        &self.label
+    }
```
Razican, line 279: This should also return an `Option<&str>`:
```suggestion
    pub fn label(&self) -> Option<&str> {
        self.label.as_ref().map(Box::as_ref)
    }
```

boa/src/syntax/ast/node/mod.rs
```diff
@@ -330,18 +163,7 @@ pub enum Node {
...
-    /// The condition is evaluated before executing the statement.
-    ///
-    /// More information:
-    ///  - [ECMAScript reference][spec]
-    ///  - [MDN documentation][mdn]
-    ///
-    /// [spec]: https://tc39.es/ecma262/#prod-grammar-notation-WhileStatement
-    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/while
-    WhileLoop(Box<Node>, Box<Node>),
+    WhileLoop(WhileLoop),
```
Razican, line 166: This would need some documentation.

boa/src/syntax/ast/node/return_smt.rs
```diff
@@ -0,0 +1,63 @@
...
+#[cfg_attr(feature = "serde", serde(transparent))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct Return {
+    expr: Option<Box<Node>>,
+}
+
+impl Return {
+    pub fn expr(&self) -> &Option<Box<Node>> {
+        &self.expr
+    }
```
Razican, line 36: This should also return Option<&Node>:
```suggestion
    pub fn expr(&self) -> Option<&Node> {
        self.expr.as_ref().map(Box::as_ref)
    }
```

boa/src/syntax/ast/node/switch.rs
```diff
@@ -0,0 +1,85 @@
...
+        &self.val
+    }
+
+    pub fn cases(&self) -> &[(Node, Box<[Node]>)] {
+        &self.cases
+    }
+
+    pub fn default(&self) -> &Option<Box<Node>> {
+        &self.default
+    }
```
Razican, line 45: This should also return an `Option<&Node>`:
```suggestion
    pub fn default(&self) -> Option<&Node> {
        self.default.as_ref().map(Box::as_ref)
    }
```

boa/src/syntax/ast/node/switch.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::fmt;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct Switch {
+    val: Box<Node>,
+    cases: Box<[(Node, Box<[Node]>)]>,
```
Razican, line 14: It's best if we implement a `Case` struct here, so that `cases` is just `Box<[Case]>`. This `Case` struct should have two members:
 - The case expression
 - The `StatementList`

boa/src/syntax/ast/node/return_smt.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Creates a `Return` AST node.
+    pub fn new<E, OE>(expr: OE) -> Self
+    where
+        E: Into<Node>,
+        OE: Into<Option<E>>,
+    {
+        Self {
+            expr: expr.into().map(E::into).map(Box::new),
```
Razican, line 45: For consistency with other `new()` functions, even though it's equivalent:
```suggestion
            expr: Box::new(expr.into().map(E::into)),
```

boa/src/syntax/ast/node/switch.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    pub fn default(&self) -> &Option<Box<Node>> {
+        &self.default
+    }
+
+    pub fn new(
+        val: Box<Node>,
+        cases: Box<[(Node, Box<[Node]>)]>,
+        default: Option<Box<Node>>,
```
Razican, line 50: We should do this more generic, like we do with others, so that we can just pas something that's `Into<Node>` and then we box it inside this function.

boa/src/syntax/parser/statement/if_stm/mod.rs
```diff
@@ -68,6 +68,6 @@ impl TokenParser for IfStatement {
             _ => None,
         };
 
-        Ok(Node::if_node::<_, _, Node, _>(cond, then_stm, else_stm))
+        Ok(If::new::<_, _, Node, _>(cond, then_stm, else_stm).into())
```
Razican, line 71: This should just return an `If`, not a `Node`.

boa/src/syntax/parser/statement/return_stm/mod.rs
```diff
@@ -51,13 +51,13 @@ impl TokenParser for ReturnStatement {
...
-            return Ok(Node::Return(None));
+            return Ok(Return::new::<Node, Option<_>>(None).into());
         }
 
         let expr = Expression::new(true, self.allow_yield, self.allow_await).parse(cursor)?;
 
         cursor.expect_semicolon(false, "return statement")?;
 
-        Ok(Node::return_node(expr))
+        Ok(Return::new(expr).into())
```
Razican, line 61: This should also return a `Return` instead of a `Node`.

boa/src/syntax/parser/statement/switch/mod.rs
```diff
@@ -54,7 +54,10 @@ impl TokenParser for SwitchStatement {
         let (cases, default) =
             CaseBlock::new(self.allow_yield, self.allow_await, self.allow_return).parse(cursor)?;
 
-        Ok(Node::switch::<_, _, _, Node>(condition, cases, default))
+        match default {
+            Some(s) => Ok(Switch::new(Box::new(condition), cases, Some(Box::new(s))).into()),
+            None => Ok(Switch::new(Box::new(condition), cases, None).into()),
+        }
```
Razican, line 60: A couple of things here. It should return a `Switch` instead of a `Node`. And we don't need a `match here:
```suggestion
        Ok(Switch::new(condition, cases, default))
```

This example above requires some generics in the `new()` function of `Switch`, as explained above.

boa/src/syntax/parser/statement/throw/mod.rs
```diff
@@ -49,6 +49,6 @@ impl TokenParser for ThrowStatement {
             }
         }
 
-        Ok(Node::throw(expr))
+        Ok(Throw::new(expr).into())
```
Razican, line 52: This should just return a `Throw`, not a `Node`.

boa/src/exec/switch/mod.rs
Lan2u: This was already present (it was within the node file), I've added a TODO and will create an issue so that it can be fixed.

boa/src/exec/switch/mod.rs
Lan2u: Won't this have problems when it comes to implementing break because if we execute block then how do we know if it ended with break and therefore should immediately return from switch.run()?

Razican: Breaks should be implemented the same way to a return, so we should be able to do that here.

boa/src/syntax/ast/node/switch.rs
Lan2u: I tried like this,
`/// Creates a `Switch` AST node.
    pub fn new<V, C, D>(val: V, cases: C, default: D) -> Self
    where
        V: Into<Node>,
        C: Into<Box<[(V, Box<[V]>)]>>,
        D: Into<Option<V>>,
    {
        // expr: expr.into().map(E::into).map(Box::new),
        Self {
            val: Box::new(val.into()),
            cases: cases.into(),
            default: default.into().map(V::into).map(Box::new),
        }
    }`

but I'm struggling to think how to convert cases correctly.

I tried an as_ref, then map but not sure how to go from the map back into the box, do you have any suggestions?

boa/src/syntax/parser/statement/if_stm/mod.rs
Lan2u: Done :)

boa/src/syntax/parser/statement/return_stm/mod.rs
Lan2u: Done (also getting the other areas which have similar problems)

boa/src/syntax/ast/node/switch.rs
Razican: I would do it like this:
```rust
/// Creates a `Switch` AST node.
pub fn new<V, C, D>(val: V, cases: C, default: Option<V>) -> Self
where
    V: Into<Node>,
    C: Into<Box<[Case]>>,
{
    // expr: expr.into().map(E::into).map(Box::new),
    Self {
        val: Box::new(val.into()),
        cases: cases.into(),
        default: default.map(V::into).map(Box::new),
     }
}
```

Then, you need a `Case` struct like this one:
```rust
struct Case {
	condition: Node,
	statements: StatementList,
}
```

boa/src/syntax/parser/statement/throw/mod.rs
Lan2u: Fixed, along with similar instances for other types.

boa/src/syntax/parser/statement/switch/mod.rs
Lan2u: Fixed as part of commit ea45b261 (to be merged), found that there was already a Case struct created for a similar purpose as part of SwitchStatement parsing so removed that struct and replaced it with the new Case struct.

boa/src/syntax/ast/node/switch.rs
Lan2u: Fixed as part of commit ea45b26 (to be merged), found that there was already a Case struct created for a similar purpose as part of SwitchStatement parsing so removed that struct and replaced it with the new Case struct.

boa/src/syntax/parser/statement/switch/mod.rs
Lan2u: Fixed

boa/src/exec/switch/mod.rs
Lan2u: Ah ok so with a flag?

boa/src/exec/switch/mod.rs
Lan2u: Created issue #439

boa/src/exec/switch/mod.rs
Razican: > Ah ok so with a flag? 

Yep, in the interpreter. At some point there will be flags for awaiting, yielding and more.

boa/src/syntax/ast/node/switch.rs
Razican: > Fixed as part of commit ea45b26 (to be merged), found that there was already a Case struct created for a similar purpose as part of SwitchStatement parsing so removed that struct and replaced it with the new Case struct.

We can have both, a `Case` parser and the produced `Case` Node, as we do with `Block` for example, if it makes sense.

boa/src/syntax/ast/node/switch.rs
Lan2u: Done :)

boa/src/syntax/ast/node/return_smt.rs
Lan2u: Found that with this change the code won't compile because expr type is Option<Box<Node>> and Box::new(...) creates Box<Option<Node>>

Razican: Yes, good catch.

jasonwilliams:
You have quite a few conflicts. @Lan2u

Lan2u:
> You have quite a few conflicts. @Lan2u

It appears these were all related to the addition of the BoaProfiler to the master branch today, I have gone through and fixed all the conflicts :)

PR #423, opened 2020-05-27
Replacement of dead links

pedropaulosuzuki:
After updating the README.md (#416), I realized that some other files still pointed to the old repository (https://github.com/jasonwilliams/boa) and needed to be updated, mainly the CHANGELOG.md. Other files changed were the CONTRIBUTING.md and both Cargo.toml files from boa-cli and boa-wasm.

It changes the following:
 - Replacement of dead links since the change from <https://github.com/jasonwilliams/boa> to <https://github.com/boa-dev/boa>.


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-# [# 0.8.0 (2020-05-23) - BigInt, Modularized Parser, Faster Hashing](https://github.com/jasonwilliams/boa/compare/v0.7.0...HEAD)
+# [# 0.8.0 (2020-05-23) - BigInt, Modularized Parser, Faster Hashing](https://github.com/boa-dev/boa/compare/v0.7.0...HEAD)
```
Razican, line 3: This should probably be changed for the diff between 0.7 and 0.8, not the head.

pedropaulosuzuki: > 
> 
> Looks good! Do we have the correct link for the repo in boa/Cargo.toml?

My bad. Now it does.

> 
> 
> This should probably be changed for the diff between 0.7 and 0.8, not the head.

Fixed!

PR #425, opened 2020-05-30
Specification compliant `ToString` (`to_string`)

HalidOdat:
This PR makes `ToString` (`to_string`) spec compliant.

#### What was wrong with the previous implementation?
The previous implementation used to `ToString` trait which was automatically implemented for `Display`, the problem is that we use different logic for displaying and converting to `String` type, for example, converting a string value to a `String` type with displays `to_string` is `"some string value"` (don't get the quotes, because the is a bug #373) but with the spec compliant function it's `some string value` without the quotes.
Additionally the spec compliant version should call `to_primitive` on objects and if it was to throw an exception it would have to propagate it.

This PR should unblock #373 .

Changes:
 - Moved `Value` display logic to `builtins/value/display.rs`
 - Added `to_string` to interpreter
 - Change most use of displays `to_string` => interpreters `to_string`


boa/src/builtins/bigint/mod.rs
```diff
@@ -64,6 +64,18 @@ impl BigInt {
         Ok(data)
     }
 
+    #[inline]
+    #[allow(clippy::wrong_self_convention)]
```
jasonwilliams, line 68: Whats this clippy allow for?

HalidOdat: It's because clippy says that `to_`, `as_` functions should take a reference to `Self`

HalidOdat: I will change this when I make `ToBigInt` (to_bigint) spec compliant, but for no this will suffice

boa/src/builtins/array/mod.rs
```diff
@@ -362,13 +362,10 @@ impl Array {
...
-            Ok(v) => match *v {
-                ValueData::String(ref s) => (*s).clone(),
-                _ => "".to_string(),
-            },
-            Err(v) => format!("error: {}", v),
+        let join = ctx.call(&method, this, &arguments)?;
+        let match_string = match join.data() {
+            ValueData::String(ref s) => s.clone(),
+            _ => "".to_owned(),
         };
```
Razican, line 369: I would change this for an `if let`, and I think we don't need to clone here, right?:
```suggestion
        let match_string = if let ValueData::String(ref s) = join.data() {
            s
        } else {
            ""
        };
```

boa/src/builtins/string/mod.rs
```diff
@@ -70,8 +70,11 @@ impl String {
     #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_string(this: &mut Value, _: &[Value], _: &mut Interpreter) -> ResultValue {
         // Get String from String Object and send it back as a new value
-        let primitive_val = this.get_internal_slot("StringData");
-        Ok(Value::from(format!("{}", primitive_val)))
+        match this.get_internal_slot("StringData").data() {
+            ValueData::String(ref string) => Ok(Value::from(string.clone())),
+            // Throw expection here:
+            _ => panic!("TypeError: this is not a string"),
+        }
```
Razican, line 77: Ideally, we should `throw` the `TypeError` here, as we do with the `RangeError` in other places. Also, this could be replaced by an `if let` and reduce indentation:
```suggestion
        if let ValueData::String(ref string) = this.get_internal_slot("StringData").data() {
            Ok(Value::from(string.clone()))
        } else {
            // TODO: Throw expection here:
            panic!("TypeError: this is not a string")
        }
```

boa/src/builtins/string/mod.rs
```diff
@@ -701,9 +704,7 @@ impl String {
 
         let fill_string = match args.len() {
             1 => None,
-            _ => Some(StdString::from(
-                args.get(1).expect("Could not get argument"),
-            )),
+            _ => Some(ctx.to_string(args.get(1).expect("Could not get argument"))?),
         };
```
Razican, line 708: I would change this for an `if let` too.

boa/src/builtins/value/display.rs
```diff
@@ -0,0 +1,225 @@
...
+        print_obj_value!(impl properties, $obj, |(key, val)| {
+            let v = &val
+                .value
+                .as_ref()
+                .expect("Could not get the property's value");
+
+            format!(
+                "{}{}: {}",
+                String::from_utf8(vec![b' '; $indent])
+                                .expect("Could not create indentation string"),
```
Razican, line 51: I would use [`str::repeat`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.repeat) here. It should remove the `expect()` branch, the UTF-8 checking, and we could personalize the indentation by either using 2 spaces, 4 spaces or a tab, with a constant (or multiple constants).

boa/src/builtins/value/display.rs
```diff
@@ -0,0 +1,225 @@
...
+
+            props
+        }
+    };
+    (internals of $obj:expr, $display_fn:ident, $indent:expr, $encounters:expr) => {
+        print_obj_value!(impl internal_slots, $obj, |(key, val)| {
+            format!(
+                "{}{}: {}",
+                String::from_utf8(vec![b' '; $indent])
+                                .expect("Could not create indentation string"),
```
Razican, line 35: I would use [`str::repeat`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.repeat) here. It should remove the `expect()` branch, the UTF-8 checking, and we could personalize the indentation by either using 2 spaces, 4 spaces or a tab, with a constant (or multiple constants).

boa/src/builtins/value/display.rs
```diff
@@ -0,0 +1,225 @@
...
+/// - A HashSet with the addresses of the already printed objects for the current branch
+///      (used to avoid infinite loops when there are cyclic deps)
+macro_rules! print_obj_value {
+    (all of $obj:expr, $display_fn:ident, $indent:expr, $encounters:expr) => {
+        {
+            let mut internals = print_obj_value!(internals of $obj, $display_fn, $indent, $encounters);
+            let mut props = print_obj_value!(props of $obj, $display_fn, $indent, $encounters, true);
+
+            props.reserve(internals.len());
+
```
Razican, line 24:
```suggestion
```

boa/src/builtins/value/mod.rs
```diff
@@ -374,6 +370,8 @@ impl ValueData {
         }
     }
 
+    // pub fn to_string(&self, ctx: &mut Interpreter) {}
+
```
Razican, line 374: What is happening here? Should we add a `TODO` or should it be removed?

boa/src/exec/mod.rs
```diff
@@ -343,9 +327,9 @@ impl Interpreter {
             ValueData::Object(_) => {
                 let prim_value = self.to_primitive(&mut (value.clone()), Some("number"));
                 self.to_string(&prim_value)
-                    .to_string()
+                    .expect("string")
```
Razican, line 330: Could we add a more explanatory message here?

boa/src/builtins/value/mod.rs
HalidOdat: I wasn't sure wether to add it to Value of Interpreter. I removed it.

boa/src/exec/mod.rs
HalidOdat: I would have liked to use the `?` operator, but the `to_number` is not spec compliant when I make `to_number` spec compliant i will use `?`.

boa/src/builtins/value/display.rs
HalidOdat: I think we should use `fromat!("{:>width}: {}", key, ..., width = $ident)` this eliminates the unnecessary string construction

boa/src/builtins/string/mod.rs
HalidOdat: I my opinion using `match` in this scenario is better with the `if let` we get 80 something characters in one line (not including indent)

Issue #430, opened 2020-05-31
[Feature Request] Add benchmarks for long file lexing + parsing


Razican:
**Current situation**
We currently don't have any benchmark checking how fast we lex and parse big files, but our current approach of reading the whole file into memory and only when we finish start lexing, and only when we finish lexing start parsing, all in one thread is far from optimal, especially for big files.

**Benchmark explanation**
We need a benchmark where we load a big JS file from disk (at least 5kB) and we lex + parse it.

Note that this file must only contain syntax that Boa understands. It can be a nonsense file, though, copy-pasting the same code multiple times.

It's important that the whole process is inside of the benchmark function, including reading the file from the disk, lexing it and parsing it.

Issue #429, opened 2020-05-31
[Feature Request] - Create benchmark for RegExp manipulation


jasonwilliams:
In order to test the upcoming [optimisation](https://github.com/boa-dev/boa/pull/419) work we're doing around objects it would be good to have a benchmark for various objects and manipulating them.

In future we may change our underlying RegExp dependency, so it would be good to see what the effects are.

## Example
Benchmarks live [here](https://github.com/boa-dev/boa/tree/master/boa/benches)
an array creation and access would be categorised under `exec`.

If exec benchmarks become too big we can compartmentalise them later on.

You can see the fibonacci benchmark here:
https://github.com/boa-dev/boa/blob/master/boa/benches/exec.rs#L49-L66

2 benchmarks could be made, 1 for using a RegExp object, another for creating one.
We could introduce more benchmarks here for various methods commonly used on RegExp objects.

in order to run the benchmarks you can run `cargo bench` in the root folder

## Contributing
https://github.com/boa-dev/boa/blob/master/CONTRIBUTING.md
https://github.com/boa-dev/boa/blob/master/docs/debugging.md

Issue #427, opened 2020-05-31
[Feature Request] - Create benchmark for Object creation and access


jasonwilliams:
In order to test the upcoming [optimisation](https://github.com/boa-dev/boa/pull/419) work we're doing around objects it would be good to have a benchmark for objects so we can see a before and after.

## Example
Benchmarks live [here](https://github.com/boa-dev/boa/tree/master/boa/benches)
an object creation and access would be categorised under `exec`.

You can see the fibonacci benchmark here:
https://github.com/boa-dev/boa/blob/master/boa/benches/exec.rs#L49-L66

2 benchmarks could be made, 1 for accessing a value from an object, another for creating an object.

in order to run the benchmarks you can run `cargo bench` in the root folder

## Contributing
https://github.com/boa-dev/boa/blob/master/CONTRIBUTING.md
https://github.com/boa-dev/boa/blob/master/docs/debugging.md

PR #431, opened 2020-05-31
Execution benchmarks now only take execution into account

Razican:
This Pull Request tries to change the execution benchmarks in order to only benchmark the actual execution. This should give us much more interesting information on the actual execution.

To be determined:
 - Should we rename the benchmarks in order to not show a huge drop on execution time in the benchmarks web page?
 - Should we maintain old benchmarks of Realm + exec too and just duplicate the benchmarks?

I can also add benchmarks for #427, #429 and #430 if you'd like.

### DIFF
--- a/boa/benches/exec.rs
+++ b/boa/benches/exec.rs
@@ -1,6 +1,6 @@
 //! Benchmarks of the whole execution engine in Boa.
 
-use boa::{exec, realm::Realm};
+use boa::{exec::Interpreter, realm::Realm, Executable, Lexer, Parser};
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 
 #[cfg(all(target_arch = "x86_64", target_os = "linux", target_env = "gnu"))]
@@ -11,9 +11,9 @@ use criterion::{black_box, criterion_group, criterion_main, Criterion};
 static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
 static SYMBOL_CREATION: &str = r#"
-let a = Symbol();
-let b = Symbol();
-let c = Symbol();
+(function () {
+    return Symbol();
+})();
 "#;
 
 fn create_realm(c: &mut Criterion) {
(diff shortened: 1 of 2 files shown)


boa/benches/exec.rs
```diff
@@ -11,24 +11,29 @@ use criterion::{black_box, criterion_group, criterion_main, Criterion};
...
 
 fn symbol_creation(c: &mut Criterion) {
+    // Create new Realm
+    let realm = Realm::create();
+    let mut engine = Interpreter::new(realm);
+    forward(&mut engine, SYMBOL_CREATION); // In order to initialize variables
+
     c.bench_function("Symbols (Execution)", move |b| {
-        b.iter(|| exec(black_box(SYMBOL_CREATION)))
+        b.iter(|| forward(&mut engine, black_box(SYMBOL_CREATION)))
```
HalidOdat, line 30: The `forward` function also lexes and parser, which would mess with the result benchmarks

jasonwilliams:
> Should we rename the benchmarks in order to not show a huge drop on execution time in the benchmarks web page?
Should we maintain old benchmarks of Realm + exec too and just duplicate the benchmarks?

I'm personally ok with having the drop, there's enough of us with context as to what happened. When we merge to master we can mention why in the commit and it will show on the benchmarks page when you click through the commit hash

boa/benches/exec.rs
jasonwilliams: `forward` should be more accurate as its not creating a new interprerter and realm all over again, but yes it will still lex and parse the new input, i don't see how you can avoid that.

Razican: I think we can avoid this by lexing and parsing before. Let me see if I can improve it.

Razican:
This is ready to go! :D

HalidOdat:
The Symbol benchmark `-2365%` :rofl:

jasonwilliams:
Do you think we need some documentation here?
Like a readme in this directory

boa/benches/exec.rs
```diff
@@ -11,57 +11,83 @@ use criterion::{black_box, criterion_group, criterion_main, Criterion};
...
+    let mut engine = Interpreter::new(realm);
+
+    let mut lexer = Lexer::new(black_box(FOR_LOOP));
+    lexer.lex().expect("failed to lex");
+
+    let nodes = Parser::new(&black_box(lexer.tokens)).parse_all().unwrap();
+
     c.bench_function("For loop (Execution)", move |b| {
-        b.iter(|| exec(black_box(FOR_LOOP)))
+        b.iter(|| black_box(&nodes).run(&mut engine).unwrap())
```
jasonwilliams, line 62: Maybe we want to leave a comment here to say what's happening

boa/benches/exec.rs
```diff
@@ -11,57 +11,83 @@ use criterion::{black_box, criterion_group, criterion_main, Criterion};
...
+    let mut engine = Interpreter::new(realm);
+
+    let mut lexer = Lexer::new(black_box(FIBONACCI));
+    lexer.lex().expect("failed to lex");
+
+    let nodes = Parser::new(&black_box(lexer.tokens)).parse_all().unwrap();
+
     c.bench_function("Fibonacci (Execution)", move |b| {
-        b.iter(|| exec(black_box(FIBONACCI)))
+        b.iter(|| black_box(&nodes).run(&mut engine).unwrap())
```
jasonwilliams, line 90: Same as above

Razican:
> Do you think we need some documentation here?
> Like a readme in this directory

I will add it.

Do you think we should retroactively fix older benchmarks? This could be done by subtracting the Realm creation, the lexing and the parsing from the execution benchmarks programmatically.

HalidOdat:
> Do you think we should retroactively fix older benchmarks? This could be done by subtracting the Realm creation, the lexing and the parsing from the execution benchmarks programmatically.

You mean the benchmark graphs, so we don't get a misleading drop in benchmarks, right? If so, I think we should if it's possible.

Razican:
These benchmarks should be using the latest changes we merged yesterday in the master branch of criterion compare, right?

> > Do you think we should retroactively fix older benchmarks? This could be done by subtracting the Realm creation, the lexing and the parsing from the execution benchmarks programmatically.
> 
> You mean the benchmark graphs, so we don't get a misleading drop in benchmarks, right? If so, I think we should if it's possible.

Yes, I meant that. I think it would be nicer. We could use the opportunity to normalize lexer + parser benchmarks.

HalidOdat:
> These benchmarks should be using the latest changes we merged yesterday in the master branch of criterion compare, right?

Yes. I was thinking about that too, I re-ran the benchmarks, the same thing happened. They should have been updated. :confused:

jasonwilliams:
> > Do you think we need some documentation here?
> > Like a readme in this directory
> 
> I will add it.
> 
> Do you think we should retroactively fix older benchmarks? This could be done by subtracting the Realm creation, the lexing and the parsing from the execution benchmarks programmatically.

Yeah if we can we should, do you have a plan of how to do this?

Razican:
> > > Do you think we need some documentation here?
> > > Like a readme in this directory
> > 
> > I will add it.
> > 
> > Do you think we should retroactively fix older benchmarks? This could be done by subtracting the Realm creation, the lexing and the parsing from the execution benchmarks programmatically.
> 
> Yeah if we can we should, do you have a plan of how to do this?

Yep, I'm writing a script to calculate real execution benchmarks by subtracting realm execution + parsing + lexing, if available, if not, these two can be calculated approximately.

Razican:
I think this can be merged, right, @jasonwilliams? Probably before #458.

PR #441, opened 2020-06-01
Optimise type comparisons

### LINKED ISSUE #435: Optimize type comparisons
**Current situation**
Currently, if the interpreter wants to know what type a value is, it has to call `get_type()`, which returns a `&str`, and compare it to another `&str`. This takes a pretty log time.

It would be much faster to just compare values in an enumeration, which should be at most 1-word size, and can therefore be stored in the stack easily.

We sill need the `&str` version of the type, though, since for example, the `typeof` operator needs to display it. I think the best option would be for the `Type` enum to implement a function called `as_str()` so that we can get the `&str` value if needed.

So, the idea is to have a `Type` enum, that is returned by the `get_type()` function [here][get_type], and then, that enum to have an `as_str()` function with a similar code to the current function.

[get_type]: https://github.com/boa-dev/boa/blob/82908dfdf457a1e21922bf6684bc77bbe8132423/boa/src/builtins/value/mod.rs#L766-L783

Lan2u:
This Pull Request fixes #435.

It changes the following:
 - Add a new Type enum which is returned from the get_type() function replacing the string used previously.
 - Updates usages.
 
I've made the mistake of basing this of the branch used for pull request #422 so depending on if/when that PR is merged this PR can be merged. If it's going to be a pain to merge I can re-do it based on master.


boa/src/builtins/value/val_type.rs
```diff
@@ -0,0 +1,32 @@
...
+            Self::Boolean => write!(f, "boolean"),
+            Self::Symbol => write!(f, "symbol"),
+            Self::Null => write!(f, "object"),
+            Self::Undefined => write!(f, "undefined"),
+            Self::Function => write!(f, "function"),
+            Self::Object => write!(f, "object"),
+            Self::BigInt => write!(f, "bigint"),
+        }
+    }
+}
```
HalidOdat, line 32: Maybe a better way of stringifying it would be to have a method on `Type` `as_str` which returns a `&str` so we don't have to construct a `String` when we want to get a string representation of `Type` and then in `Display` we call `as_str`

boa/src/builtins/value/mod.rs
```diff
@@ -762,25 +766,27 @@ impl ValueData {
...
+                    Type::Function
                 } else {
-                    "object"
+                    Type::Object
                 }
             }
-            Self::BigInt(_) => "bigint",
+            Self::BigInt(_) => Type::BigInt,
         }
     }
```
HalidOdat, line 791: Maybe, it would be nicer to put `get_type` in `val_type` module

boa/src/builtins/value/val_type.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Possible types of val as defined at https://tc39.es/ecma262/#sec-typeof-operator.
+/// Note that an object which implements call is referred to here as 'Function'.
+#[derive(Eq, PartialEq, Debug, Clone, Copy)]
+pub enum Type {
+    Undefined,
+    Null,
+    Boolean,
+    Number,
+    Str,
```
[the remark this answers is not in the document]
Lan2u: Done

boa/src/builtins/value/val_type.rs
Lan2u: Done

boa/src/builtins/value/mod.rs
Lan2u: Done, commit 93273ae

boa/src/exec/conditional/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+impl Executable for If {
+    fn run(&self, interpreter: &mut Interpreter) -> ResultValue {
+        Ok(if self.cond().run(interpreter)?.borrow().is_true() {
+            self.body().run(interpreter)?
+        } else {
+            match self.else_node() {
+                Some(ref else_e) => else_e.run(interpreter)?,
+                None => Value::undefined(),
+            }
+        })
```
Razican, line 17: Could this be reduced to an if let at the level of the else?
```suggestion
        } else if let Some(ref else_e) = self.else_node() {
            else_e.run(interpreter)?
        } else {
        	Value::undefined()
        })
```

boa/src/builtins/value/equality.rs
```diff
@@ -200,16 +200,16 @@ pub fn same_value_zero(x: &Value, y: &Value) -> bool {
...
-        "undefined" => true,
-        "null" => true,
-        "string" => {
+        Type::Undefined => true,
+        Type::Null => true,
+        Type::String => {
             if x.to_string() == y.to_string() {
                 return true;
             }
             false
```
Razican, line 209: This can be reduced further:
```suggestion
            x.to_string() == y.to_string()
```

boa/src/exec/conditional/mod.rs
Lan2u: Done :)

boa/src/builtins/value/equality.rs
Lan2u: Fixed

boa/src/syntax/parser/statement/if_stm/mod.rs
```diff
@@ -2,6 +2,7 @@
 mod tests;
 
 use super::Statement;
+
```
Razican, line 5: Why this new line?

boa/src/syntax/parser/statement/declaration/mod.rs
```diff
@@ -13,6 +13,7 @@ mod lexical;
 mod tests;
 
 use self::{hoistable::HoistableDeclaration, lexical::LexicalDeclaration};
+
```
Razican, line 16: Why this new line?

boa/src/syntax/parser/statement/if_stm/mod.rs
Lan2u: I imagine rustfmt did that as it is not something I did directly

boa/src/syntax/parser/statement/declaration/mod.rs
Lan2u: Rustfmt I think?

PR #442, opened 2020-06-02
Added `TypeError` implementation

HalidOdat:
It changes the following:
 - Added `TypeError` implementation.
 - Fixed `RangeError` and `Error` throwing.
 - Moved throwing code to `Interpreter`.
 - Added `throw_range_error()` and `throw_type_error()` to `Interpreter`.

Now with this PR it should be easier to throw exceptions, for example:
```rust
fn function_that_throws(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
	if args.is_empty() {
		return ctx.throw_range_error("Some message");
	}
	
	ctx.throw_type_error("Some other message")
}
```

This is more or less the way `neon` crate does it.

### DIFF
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -420,14 +420,14 @@ impl Number {
 
         // Constants from:
         // https://tc39.es/ecma262/#sec-properties-of-the-number-constructor
-        number.set_field("EPSILON", Value::from(std::f64::EPSILON));
+        number.set_field("EPSILON", Value::from(f64::EPSILON));
         number.set_field("MAX_SAFE_INTEGER", Value::from(9_007_199_254_740_991_f64));
         number.set_field("MIN_SAFE_INTEGER", Value::from(-9_007_199_254_740_991_f64));
-        number.set_field("MAX_VALUE", Value::from(std::f64::MAX));
-        number.set_field("MIN_VALUE", Value::from(std::f64::MIN));
-        number.set_field("NEGATIVE_INFINITY", Value::from(std::f64::NEG_INFINITY));
-        number.set_field("POSITIVE_INFINITY", Value::from(std::f64::INFINITY));
-        number.set_field("NaN", Value::from(std::f64::NAN));
+        number.set_field("MAX_VALUE", Value::from(f64::MAX));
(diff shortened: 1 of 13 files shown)


boa/src/builtins/number/mod.rs
```diff
@@ -428,9 +425,9 @@ impl Number {
         number.set_field("MIN_SAFE_INTEGER", Value::from(-9_007_199_254_740_991_f64));
         number.set_field("MAX_VALUE", Value::from(std::f64::MAX));
         number.set_field("MIN_VALUE", Value::from(std::f64::MIN));
-        number.set_field("NEGATIVE_INFINITY", Value::from(f64::NEG_INFINITY));
-        number.set_field("POSITIVE_INFINITY", Value::from(f64::INFINITY));
-        number.set_field("NaN", Value::from(f64::NAN));
+        number.set_field("NEGATIVE_INFINITY", Value::from(std::f64::NEG_INFINITY));
+        number.set_field("POSITIVE_INFINITY", Value::from(std::f64::INFINITY));
+        number.set_field("NaN", Value::from(std::f64::NAN));
```
Razican, line 430: Why was this changed? I think we should use the built-in constants instead of the module-level constants. They were stabilized in 1.42 or 1.43 and the others are more or less deprecated in idiomatic code.

boa/src/exec/mod.rs
```diff
@@ -140,7 +142,10 @@ impl Interpreter {
             ValueData::Rational(rational) => Ok(Number::to_native_string(*rational)),
             ValueData::Integer(integer) => Ok(integer.to_string()),
             ValueData::String(string) => Ok(string.clone()),
-            ValueData::Symbol(_) => panic!("TypeError exception."),
+            ValueData::Symbol(_) => {
+                self.throw_type_error("TypeError: \"can't convert symbol to string\"")?;
```
Razican, line 146: Should we just add the `TypeError` part to all the `TypeError` messages when printing?

HalidOdat: Yes. you are right. The message will be `TypeError: "TypeError: \"can't convert symbol to string\""`, I'll change it :)

Issue #439, opened 2020-06-01
Implement break handling in switch statements


Lan2u:
Currently, the Switch Executable implementation [at](https://github.com/boa-dev/boa/blob/7733da27c7b4c2b766b67cd2eab41afbb2f8f08e/boa/src/exec/switch/mod.rs#L8-L29) doesn't support break statements as part of a switch statement.

**Example code**
```javascript
let a = "A";
let b = "B";
let c = "hi";
switch (c) {
    case 'hi':
        a = 'hi';
        break;
    case 'hi':
        b = 'hi';
        break;
}
console.log(a);
console.log(b);
```
The expected output is:
'hi'
'B'

Note that the code above won't work until issue #301 is also resolved.

Lan2u:
I can do this as an extension to PR #422

Razican:
> I can do this as an extension to PR #422

Sure, go ahead!

Issue #301, opened 2020-04-03
Implementing the switch statement in the new parser


HalidOdat:
Currently the switch statement is not implemented in the new parser.

Lan2u:
Will give this a go as part of issue #439.

PR #451, opened 2020-06-03
Switch impl

Lan2u:
This Pull Request implements #439 and #301.

It changes the following:
 - Implements parsing and running of switch statements

A function parse_generalised in statement list was created as part of this and is a more generalised version of StatementList::parse() which allows statement lists to be terminated by tokens other than '}'. It's a separate function for now so that the performance impact of replacing parse can be established. 

This pull request is based of PR #422.


boa/src/exec/mod.rs
```diff
@@ -48,8 +49,12 @@ pub trait Executable {
 /// A Javascript intepreter
 #[derive(Debug)]
 pub struct Interpreter {
-    /// Wether it's running a return statement.
+    /// Whether it's running a return statement.
     is_return: bool,
+
+    /// Whether it's running a break statement.
+    is_break: bool,
```
Razican, line 56: I don't think we can be executing `break` and `return` at the same time. We should probably just have a flag that states `current_state`, which could be `Execution`, `Break`, `Return`, and maybe more in the future, such as `Yield`, `Continue`, `Await`...

This also adds us the option to add a `label` to `break` and `continue` once we implement labeled statements.

Then, we should add a couple of functions `is_return()`, `is_break()`, that would just check this.

boa/src/exec/break_node/mod.rs
```diff
@@ -0,0 +1,12 @@
+use super::{Executable, Interpreter};
+use crate::{
+    builtins::value::{ResultValue, Value},
+    syntax::ast::node::Break,
+};
+
+impl Executable for Break {
+    fn run(&self, interpreter: &mut Interpreter) -> ResultValue {
+        interpreter.is_break = true;
```
Razican, line 9: Here, we could have a `break(label: &str)` function (signature might differ) in `interpreter`, which would change the state.

boa/src/exec/mod.rs
Lan2u: I agree, would it be best to do this as part of this PR or in a separate one?

Razican: I think we should do it here. since we only had one flag before, but this adds a second one.

Lan2u: Done as part of the most recent commit, the new InterpreterState enum is exposed so that the internal label (or for new states various other information) can be accessed, additionally, I've made it so the label of a break/return node is stored in the interpreter when they run.

boa/src/exec/return_smt/mod.rs
```diff
@@ -11,7 +11,7 @@ impl Executable for Return {
             None => Ok(Value::undefined()),
         };
         // Set flag for return
-        interpreter.is_return = true;
+        interpreter.set_current_state(InterpreterState::Return(self.label().map(String::from)));
         result
```
Razican, line 15: Hmmm why does `return` have a label? It should have a returning node, right?

PR #455, opened 2020-06-04
[TemplateLiteral] Basic lexer implementation

croraf:
Very basic template literal implementation, of lexer only.

### DIFF
--- a/boa/src/syntax/ast/token.rs
+++ b/boa/src/syntax/ast/token.rs
@@ -239,6 +239,8 @@ pub enum TokenKind {
     /// A string literal.
     StringLiteral(Box<str>),
 
+    TemplateLiteral(Box<str>),
+
     /// A regular expression, consisting of body and flags.
     RegularExpressionLiteral(Box<str>, RegExpFlags),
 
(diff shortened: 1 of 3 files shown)


boa/src/syntax/lexer/tests.rs
```diff
@@ -37,6 +37,29 @@ fn check_string() {
     assert_eq!(lexer.tokens[1].kind, TokenKind::string_literal("bbb"));
 }
 
+#[test]
+fn check_template_literal_simple() {
+    let s = "`I'm a template literal`";
```
jasonwilliams, line 42: You shouldn’t need double quotes here

boa/src/syntax/lexer/tests.rs
```diff
@@ -37,6 +37,29 @@ fn check_string() {
...
+    lexer.lex().expect("failed to lex");
+    assert_eq!(
+        lexer.tokens[0].kind,
+        TokenKind::template_literal("I'm a template literal")
+    );
+}
+
+#[test]
+fn check_template_literal_unterminated() {
+    let s = "`I'm a template";
```
jasonwilliams, line 53: Same here

boa/src/syntax/ast/token.rs
```diff
@@ -239,6 +239,8 @@ pub enum TokenKind {
     /// A string literal.
     StringLiteral(Box<str>),
 
+    TemplateLiteral(Box<str>),
```
Razican, line 242: This would need some documentation. Ideally also a link to the spec.

croraf: Not sure what you mean.

Razican: At least, the variant should be documented like this:
```suggestion
    /// A template literal.
    TemplateLiteral(Box<str>),
```

croraf: Commenting a variable named TemplateLiteral with a commend saying "template literal" is bad practice. I guess the comment with the possible link to the spec can go in boa/src/syntax/lexer/mod.rs. I put it there if you check that file in this PR.

PR #459, opened 2020-06-06
parseInt, parseFloat implementation

### LINKED ISSUE #447: parseInt, parseFloat missing
I don't see a list of unsupported language features, so I'm adding this. Functions `parseInt`, `parseFloat` are not defined.

**To Reproduce**
Start CLI, enter
```javascript
parseInt(10)
```

**Expected behavior**
`10` should be returned. Actual result: `undefined`. 
- boa version: master (87aea64c)
 - OS: Ubuntu 
 - Version: 20.04
 - Rustc version:  1.43.1 (8d69840ab 2020-05-04)

Lan2u:
This Pull Request fixes/closes #447.

It changes the following:
 - Adds an implementation of the builtin parseInt and parseFloat functions
 - Associated tests


boa/src/builtins/number/mod.rs
```diff
@@ -405,6 +408,99 @@ impl Number {
...
+            match val.data() {
+                ValueData::String(s) => {
+                    if let Ok(f) = f64::from_str(s) {
+                        Ok(Value::rational(f))
+                    } else {
+                        // String can't be parsed.
+                        Ok(Value::from(f64::NAN))
+                    }
+                }
+                ValueData::Integer(i) => Ok(Value::rational(*i as f64)),
```
Razican, line 491: I think integers should be conserved as such. This is just an optimisation, an integer is just a representation of a "number".

boa/src/builtins/number/mod.rs
```diff
@@ -405,6 +408,99 @@ impl Number {
...
+    ) -> ResultValue {
+        if args.len() > PARSE_FLOAT_MAX_ARG_COUNT {
+            // Too many arguments.
+            return Err(Value::undefined());
+        }
+
+        if let Some(val) = args.get(0) {
+            match val.data() {
+                ValueData::String(s) => {
+                    if let Ok(f) = f64::from_str(s) {
```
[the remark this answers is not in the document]
Lan2u: I don't think the ::<64> can be avoided as otherwise, rust doesn't seem to be able to infer the type however I've removed FromStr + associated import.

The f64 import is kept to allow using f64::NAN.

boa/src/builtins/number/mod.rs
Lan2u: Fixed :)

boa/src/builtins/number/mod.rs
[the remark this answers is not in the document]
Razican: > The f64 import is kept to allow using f64::NAN

In theory, this is no longer needed since Rust 1.43:

https://doc.rust-lang.org/stable/std/primitive.f64.html#associatedconstant.NAN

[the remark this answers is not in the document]
Lan2u: I updated my rust version and yeah it seems you are correct, fixed in latest commit :)

boa/src/builtins/number/mod.rs
```diff
@@ -405,6 +411,132 @@ impl Number {
...
+                ValueData::Integer(i) => Ok(Value::integer(*i)),
+                ValueData::Rational(f) => Ok(Value::integer(*f as i32)),
+                _ => {
+                    // Wrong argument type to parseInt.
+                    Ok(Value::from(f64::NAN))
+                }
+            }
+        } else {
+            // Not enough arguments to parseInt.
+            Err(Value::undefined())
```
Razican, line 484: Note that if no arguments are passed, then the first argument is `undefined`, and therefore, it should return `NaN`, without throwing an exception (`Err` throws an exception with the given value).

boa/src/builtins/number/mod.rs
```diff
@@ -405,6 +411,132 @@ impl Number {
...
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
+    pub(crate) fn parse_int(
+        _this: &mut Value,
+        args: &[Value],
+        _ctx: &mut Interpreter,
+    ) -> ResultValue {
+        if args.len() > PARSE_INT_MAX_ARG_COUNT {
+            // Too many arguments.
+            return Err(Value::undefined());
+        }
```
Razican, line 436: Where does the spec say this? If they are more arguments, I think we should just ignore them, as other interpreters do.

boa/src/builtins/number/mod.rs
```diff
@@ -405,6 +411,132 @@ impl Number {
...
+                ValueData::Integer(i) => Ok(Value::integer(*i)),
+                ValueData::Rational(f) => Ok(Value::rational(*f)),
+                _ => {
+                    // Wrong argument type to parseFloat.
+                    Ok(Value::from(f64::NAN))
+                }
+            }
+        } else {
+            // Not enough arguments to parseFloat.
+            Err(Value::undefined())
```
Razican, line 536: I think this is the same case as with `parseInt`, we should just return `Ok(Value::from(f64::NAN))`

boa/src/builtins/number/mod.rs
```diff
@@ -405,6 +411,132 @@ impl Number {
...
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat
+    pub(crate) fn parse_float(
+        _this: &mut Value,
+        args: &[Value],
+        _ctx: &mut Interpreter,
+    ) -> ResultValue {
+        if args.len() > PARSE_FLOAT_MAX_ARG_COUNT {
+            // Too many arguments.
+            return Err(Value::undefined());
+        }
```
Razican, line 511: Once again, I didn't see this in the spec.

boa/src/builtins/number/mod.rs
Lan2u: Ah ok, I wasn't sure how to handle too many/few arguments, I've made the changes in the latest commit

boa/src/builtins/number/mod.rs
Lan2u: Addressed in latest commit

boa/src/builtins/number/mod.rs
Lan2u: Addressed in latest commit

boa/src/builtins/number/mod.rs
Lan2u: Addressed in latest commit

boa/src/builtins/function/mod.rs
```diff
@@ -431,6 +431,21 @@ pub fn make_constructor_fn(
 
 /// Macro to create a new member function of a prototype.
```
HalidOdat, line 432:
```suggestion
/// Creates a new member function of a `Object` or `prototype`.
```

Issue #452, opened 2020-06-03
Objects that are created with `ValuData::from_json()` do not contain `prototype` instance (`__proto__`).


HalidOdat:
**Describe the bug**

Objects that are created with `ValuData::from_json()` do not contain `prototype` instance (`__proto__`).

**To Reproduce**
```js
JSON.parse("{x: 0}")
```

**Actual behavior**
```
{
	x: 0,
    extensible: true
}
```

**Expected behavior**
```
{
	x: 0,
    extensible: true,
	__proto__: {
		...  
	}
}
```

**Build environment (please complete the following information):**
 - OS: Manjaro Linux
 - Target triple: x86_64-unknown-linux-gnu
 - Rustc version:  rustc 1.43.1 (8d69840ab 2020-05-04)

**Additional context**

For this to be fixed we need `ValuData::from_json()` to take in an `Interpreter` (context) get the `prototype` and construct the `Object`.

n14little:
@HalidOdat it looks like in order to get the `Interpreter` to `ValueData::from_json()` we would need to have access to the `Interpreter` in this conversion: https://github.com/boa-dev/boa/blob/bb2b6f638cba59269694d44424d16c9a5ede6a81/boa/src/builtins/value/conversions.rs#L191-L195
But, from my understanding, that is not possible. It seems that this conversion implementing `From<JSONValue> for Value` just have to go away? Is that correct, or is there something I'm missing? I'm pretty new to Rust.

HalidOdat:
> But, from my understanding, that is not possible. It seems that this conversion implementing `From<JSONValue> for Value` just have to go away? Is that correct, or is there something I'm missing? I'm pretty new to Rust.

No. That's right. We should remove it and use `ValueData::from_json()` with it taking an `Interpreter`.

n14little:
I am working on this to get myself going on some other issues assigned to me.

PR #462, opened 2020-06-07
Add proto object to from json

n14little:
This Pull Request addresses objects created from JSON parse #452.

It changes the following:
 - Removes `impl From<JSONValue> for Value { ... }` conversion.
 - Removes `ValueData::from_json` and `ValueData::to_json`
 - Adds `Value::from_json` and `Value::to_json`
 - Adds the object prototype to object created from JSON objects in `Value::from_json()`

It does not yet address Booleans, Arrays, Strings, Numbers or Null.

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -107,7 +107,8 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
     };
     let replacer = match args.get(1) {
         Some(replacer) if replacer.is_object() => replacer,
-        _ => return Ok(Value::from(object.to_json().to_string())),
+        // TODO: handle the error case
+        _ => return Ok(Value::from(object.to_json(ctx).unwrap().to_string())),
     };
 
     let replacer_as_object = replacer
@@ -133,7 +134,10 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
                         )?),
                     );
                 }
-                Ok(Value::from(object_to_return.to_json().to_string()))
+                // TODO: Handle the error case
+                Ok(Value::from(
+                    object_to_return.to_json(ctx).unwrap().to_string(),
+                ))
             })
             .ok_or_else(Value::undefined)?
     } else if replacer_as_object.kind == ObjectKind::Array {
@@ -149,7 +153,8 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
         for field in fields {
             if let Some(value) = object
                 .get_property(&ctx.to_string(&field)?)
-                .map(|prop| prop.value.as_ref().map(|v| v.to_json()))
+                // TODO: handle the error case
+                .map(|prop| prop.value.as_ref().map(|v| v.to_json(ctx).unwrap()))
                 .flatten()
             {
                 obj_to_return.insert(field.to_string(), value);
@@ -157,7 +162,8 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
         }
         Ok(Value::from(JSONValue::Object(obj_to_return).to_string()))
     } else {
-        Ok(Value::from(object.to_json().to_string()))
+        // TODO: handle the error case
+        Ok(Value::from(object.to_json(ctx).unwrap().to_string()))
     }
 }
 
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -17,9 +17,10 @@ use crate::builtins::{
     function::make_builtin_fn,
     object::ObjectKind,
     property::Property,
-    value::{ResultValue, Value},
+    value::{ResultValue, Value, ValueData},
 };
 use crate::{exec::Interpreter, BoaProfiler};
+use gc::Gc;
 use serde_json::{self, Value as JSONValue};
 
 #[cfg(test)]
@@ -42,7 +43,7 @@ pub fn parse(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValu
         &ctx.to_string(args.get(0).expect("cannot get argument for JSON.parse"))?,
     ) {
         Ok(json) => {
-            let j = Value::from(json);
+            let j = Value(Gc::new(ValueData::from_json(json)));
             match args.get(1) {
                 Some(reviver) if reviver.is_function() => {
                     let mut holder = Value::new_object(None);
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -19,7 +19,8 @@ use crate::builtins::{
     property::Property,
     value::{ResultValue, Value, ValueData},
 };
-use crate::{exec::Interpreter, BoaProfiler};
+use crate::exec::Interpreter;
+use crate::profiler::BoaProfiler;
 use gc::Gc;
 use serde_json::{self, Value as JSONValue};
 
@@ -43,7 +44,7 @@ pub fn parse(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValu
         &ctx.to_string(args.get(0).expect("cannot get argument for JSON.parse"))?,
     ) {
         Ok(json) => {
-            let j = Value(Gc::new(ValueData::from_json(json)));
+            let j = Value(Gc::new(ValueData::from_json(json, ctx)));
             match args.get(1) {
                 Some(reviver) if reviver.is_function() => {
                     let mut holder = Value::new_object(None);
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -19,8 +19,7 @@ use crate::builtins::{
     property::Property,
     value::{ResultValue, Value},
 };
-use crate::exec::Interpreter;
-use crate::profiler::BoaProfiler;
+use crate::{exec::Interpreter, BoaProfiler};
 use serde_json::{self, Value as JSONValue};
 
 #[cfg(test)]
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -17,11 +17,10 @@ use crate::builtins::{
     function::make_builtin_fn,
     object::ObjectKind,
     property::Property,
-    value::{ResultValue, Value, ValueData},
+    value::{ResultValue, Value},
 };
 use crate::exec::Interpreter;
 use crate::profiler::BoaProfiler;
-use gc::Gc;
 use serde_json::{self, Value as JSONValue};
 
 #[cfg(test)]
@@ -44,7 +43,7 @@ pub fn parse(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValu
         &ctx.to_string(args.get(0).expect("cannot get argument for JSON.parse"))?,
     ) {
         Ok(json) => {
-            let j = Value(Gc::new(ValueData::from_json(json, ctx)));
+            let j = Value::from_json(json, ctx);
             match args.get(1) {
                 Some(reviver) if reviver.is_function() => {
                     let mut holder = Value::new_object(None);
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -107,8 +107,7 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
     };
     let replacer = match args.get(1) {
         Some(replacer) if replacer.is_object() => replacer,
-        // TODO: handle the error case
-        _ => return Ok(Value::from(object.to_json(ctx).unwrap().to_string())),
+        _ => return Ok(Value::from(object.to_json(ctx)?.to_string())),
     };
 
     let replacer_as_object = replacer
@@ -134,9 +133,8 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
                         )?),
                     );
                 }
-                // TODO: Handle the error case
                 Ok(Value::from(
-                    object_to_return.to_json(ctx).unwrap().to_string(),
+                    object_to_return.to_json(ctx)?.to_string(),
                 ))
             })
             .ok_or_else(Value::undefined)?
@@ -153,17 +151,16 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
         for field in fields {
             if let Some(value) = object
                 .get_property(&ctx.to_string(&field)?)
-                // TODO: handle the error case
-                .map(|prop| prop.value.as_ref().map(|v| v.to_json(ctx).unwrap()))
+                .map(|prop| prop.value.as_ref().map(|v| v.to_json(ctx)))
                 .flatten()
+                .transpose()?
             {
                 obj_to_return.insert(field.to_string(), value);
             }
         }
         Ok(Value::from(JSONValue::Object(obj_to_return).to_string()))
     } else {
-        // TODO: handle the error case
-        Ok(Value::from(object.to_json(ctx).unwrap().to_string()))
+        Ok(Value::from(object.to_json(ctx)?.to_string()))
     }
 }
 
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -133,9 +133,7 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
                         )?),
                     );
                 }
-                Ok(Value::from(
-                    object_to_return.to_json(ctx)?.to_string(),
-                ))
+                Ok(Value::from(object_to_return.to_json(ctx)?.to_string()))
             })
             .ok_or_else(Value::undefined)?
     } else if replacer_as_object.kind == ObjectKind::Array {
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -149,8 +149,7 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
         for field in fields {
             if let Some(value) = object
                 .get_property(&ctx.to_string(&field)?)
-                .map(|prop| prop.value.as_ref().map(|v| v.to_json(ctx)))
-                .flatten()
+                .and_then(|prop| prop.value.as_ref().map(|v| v.to_json(ctx)))
                 .transpose()?
             {
                 obj_to_return.insert(field.to_string(), value);
--- a/boa/src/builtins/value/mod.rs
+++ b/boa/src/builtins/value/mod.rs
@@ -685,7 +685,7 @@ impl ValueData {
                     new_obj.properties.insert(
                         idx.to_string(),
                         Property::default()
-                            .value(Value::from(json.clone()))
+                            .value(Value(Gc::new(ValueData::from_json(json.clone()))))
                             .writable(true)
                             .configurable(true),
                     );
@@ -702,7 +702,7 @@ impl ValueData {
                     new_obj.properties.insert(
                         key.clone(),
                         Property::default()
-                            .value(Value::from(json.clone()))
+                            .value(Value(Gc::new(ValueData::from_json(json.clone()))))
                             .writable(true)
                             .configurable(true),
                     );
--- a/boa/src/builtins/value/mod.rs
+++ b/boa/src/builtins/value/mod.rs
@@ -202,6 +202,52 @@ impl Value {
             JSONValue::Null => Self::null(),
         }
     }
+
+    /// Conversts the `Value` to `JSON`.
+    pub fn to_json(&self) -> JSONValue {
+        match *self.data() {
+            ValueData::Null => JSONValue::Null,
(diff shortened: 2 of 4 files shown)


boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +150,95 @@ impl Value {
 
         Self::object(object)
     }
+
+    /// Convert from a JSON value to a JS value
+    pub fn from_json(json: JSONValue, interpreter: &mut Interpreter) -> Self {
+        match json {
+            JSONValue::Number(v) => {
+                Self::rational(v.as_f64().expect("Could not convert value to f64"))
+            }
```
Razican, line 159: We have a small optimisation here that we could use. If the number is an integer, we store is a `Self::integer`, which helps with performance.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +150,95 @@ impl Value {
...
+            JSONValue::String(v) => Self::string(v),
+            JSONValue::Bool(v) => Self::boolean(v),
+            JSONValue::Array(vs) => {
+                let mut new_obj = Object::default();
+                for (idx, json) in vs.iter().enumerate() {
```
Razican, line 164: I think we can consume the value and avoid the `json.clone()` below, by calling `into_iter()` instead of `iter()`. Is this possible? We would need to stract the length before, though, to use it below.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +150,95 @@ impl Value {
...
+                }
+                new_obj.properties.insert(
+                    "length".to_string(),
+                    Property::default().value(Self::from(vs.len())),
+                );
+                Self::object(new_obj)
+            }
+            JSONValue::Object(obj) => {
+                let new_obj = Value::new_object(Some(&interpreter.realm.global_obj));
+                for (key, json) in obj.iter() {
```
Razican, line 181: I think we can avoid cloning the key here by calling `into_iter()` and consuming the object.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +150,95 @@ impl Value {
...
+                }
+            }
+            ValueData::String(ref str) => JSONValue::String(str.clone()),
+            ValueData::Rational(num) => JSONValue::Number(
+                JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
+            ),
+            ValueData::Integer(val) => JSONValue::Number(JSONNumber::from(val)),
+            ValueData::BigInt(_) => {
+                // TODO: throw TypeError
+                panic!("TypeError: \"BigInt value can't be serialized in JSON\"");
```
Razican, line 235: Now that we have implemented `TypeError` throwing, I think it's a good time for `to_json()` to receive the interpreter as an argument and throw the error.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +150,95 @@ impl Value {
...
+                        if !value.is_undefined() && !value.is_function() {
+                            new_obj.insert(key, value.to_json());
+                        }
+                    });
+                    JSONValue::Object(new_obj)
+                }
+            }
+            ValueData::String(ref str) => JSONValue::String(str.clone()),
+            ValueData::Rational(num) => JSONValue::Number(
+                JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
```
Razican, line 230: I'm not 100% comfortable with this panic. What if the number is infinite or NaN?

n14little: I'm not sure. I'll have to check the spec. Most of this PR was just pulled over from previous implementation.

boa/src/builtins/value/mod.rs
n14little: Serde Json only stores numbers in 64 bits, but boa supports i32 and BigInt. How would you prefer to handle? Check to see if i64 can fit into i32; otherwise, use big int?

boa/src/builtins/value/mod.rs
n14little: :heavy_check_mark:

boa/src/builtins/value/mod.rs
n14little: :heavy_check_mark:

boa/src/builtins/value/mod.rs
Razican: Yes, I think that's the best option. Do a `try_from()` and if it fails, go for a rational.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +150,96 @@ impl Value {
...
+    /// Convert from a JSON value to a JS value
+    pub fn from_json(json: JSONValue, interpreter: &mut Interpreter) -> Self {
+        match json {
+            JSONValue::Number(v) => {
+                Self::rational(v.as_f64().expect("Could not convert value to f64"))
+            }
+            JSONValue::String(v) => Self::string(v),
+            JSONValue::Bool(v) => Self::boolean(v),
+            JSONValue::Array(vs) => {
+                let mut new_obj = Object::default();
```
HalidOdat, line 163: Array should have the `prototype` too.

n14little: Correct, I haven't done arrays, numbers, booleans, strings or null. I was able to get and set the prototype from the production code, but wasn't able to use the assertion similar to the one that I have now for the test code. As soon as I see some failing tests for the right reasons I'll address the rest of the JSON types.

boa/src/builtins/value/mod.rs
n14little: I'm having a hard time finding anything about it in the spec. Popping open the dev tools in Firefox gives me results, but I can't find any mention of the behavior that I'm seeing in the spec. Here is what Firefox provides:
```javascript
JSON.stringify({ x: NaN }) // "{\"x\":null}"
JSON.stringify({ x: Number.EPSILON }) // "{\"x\":2.220446049250313e-16}"
JSON.stringify({ x: Number.MAX_SAFE_INTEGER }) // "{\"x\":9007199254740991}"
JSON.stringify({ x: Number.MIN_SAFE_INTEGER }) // "{\"x\":-9007199254740991}"
JSON.stringify({ x: Number.MAX_VALUE }) // "{\"x\":1.7976931348623157e+308}"
JSON.stringify({ x: Number.MIN_VALUE }) // "{\"x\":5e-324}"
JSON.stringify({ x: Number.POSITIVE_INFINITY }) // "{\"x\":null}"
JSON.stringify({ x: Number.NEGATIVE_INFINITY }) // "{\"x\":null}"
```

n14little: I would venture to guess that it is buried in [SerializeJSONProperty](https://tc39.es/ecma262/#sec-serializejsonproperty).

The JSON object could probably benefit from a stricter adherence to the implementation described in the spec. It seems that JSON object might follow the intent of the spec, but not the implementation.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +151,103 @@ impl Value {
...
+                    );
+                }
+                new_obj
+            }
+            JSONValue::Null => Self::null(),
+        }
+    }
+
+    /// Conversts the `Value` to `JSON`.
+    pub fn to_json(&self, interpreter: &mut Interpreter) -> Result<JSONValue, Value> {
```
n14little, line 204: Is this the signature that you would expect to see to handle throwing the type error? It feels a little bit odd to me.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +151,103 @@ impl Value {
...
+                    Ok(JSONValue::Object(new_obj))
+                }
+            }
+            ValueData::String(ref str) => Ok(JSONValue::String(str.clone())),
+            ValueData::Rational(num) => Ok(JSONValue::Number(
+                JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
+            )),
+            ValueData::Integer(val) => Ok(JSONValue::Number(JSONNumber::from(val))),
+            ValueData::BigInt(_) => {
+                Err(interpreter.throw_type_error("BigInt value can't be serialized in JSON")?)
```
n14little, line 243: Is this how you would expect that the type errors be thrown?

HalidOdat: we would just do:
```
interpreter.throw_type_error("BigInt value can't be serialized in JSON")?;
unreachable();
```
The `throw_type_error` should always give `Err` enum field so the code bellow should be `unreachable`

boa/src/builtins/value/mod.rs
HalidOdat: Yes. this is right. we can call it `ctx` (short for context/interpreter) instead of `interpreter`

boa/src/builtins/value/mod.rs
HalidOdat: > Correct, I haven't done arrays, numbers, booleans, strings or null. I was able to get and set the prototype from the production code, but wasn't able to use the assertion similar to the one that I have now for the test code. As soon as I see some failing tests for the right reasons I'll address the rest of the JSON types.

Maybe we should make this into a separate PR.

boa/src/builtins/value/mod.rs
n14little: I should have been a little more clear, is the returned value in this signature what you would expect. That is, the `Result<JSONValue, Value>`

HalidOdat: > I should have been a little more clear, is the returned value in this signature what you would expect? That is, the `Result<JSONValue, Value>`

Yes. it seems right. it does not mater what the `Ok(_)` case is as long as the `Err(_)` case is `Value` we can throw.

boa/src/builtins/value/mod.rs
n14little: I'm not seeing how:

> The throw_type_error should always give Err enum field so the code bellow should be unreachable

is the case given that `Interpreter::throw_type_error` returns a `ResultValue` https://github.com/boa-dev/boa/blob/863abb37474d6e39ccb81e13fc56aa2739aa8d2c/boa/src/exec/exception.rs#L25-L35

Can you help me understand?

HalidOdat: When we construct a new `TypeError` or any `Error` type it will always give a `Err()`

Here is https://github.com/boa-dev/boa/blob/master/boa/src/builtins/error/range.rs#L43 `RangeError`

boa/src/builtins/value/mod.rs
n14little: I'm okay with that. In fact, I was going to ask you to edit the issue with a bullet list for each JSON type.

boa/src/builtins/value/mod.rs
n14little: Okay, so the compiler doesn't necessarily know that it will always be an error and that is why I wasn't getting a clean build.

Just for my understanding, how come `ResultValue` was used if the `Ok(...)` portion is never used?

HalidOdat: > Just for my understanding, how come `ResultValue` was used if the `Ok(...)` portion is never used?

This is just so it's nicer to throw errors, so it builtin functions we have `ResultValue` as the return type and we can do:
```rust 
fn name_of_builtin(this: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
	ctx.throw_type_error("message")
}
```
Instead of:
```rust 
fn name_of_builtin(this: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
	ctx.throw_type_error("message")?;
	unreachable();
}
```

boa/src/builtins/value/mod.rs
n14little: @Razican please take note of this thread. I noticed that you linked the PR to close the issue. I had modified the language in the PR description so that the merge would not close the issue. Please advise on whether or not this is acceptable practice for future reference.

boa/src/builtins/json/mod.rs
```diff
@@ -107,7 +107,8 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
     };
     let replacer = match args.get(1) {
         Some(replacer) if replacer.is_object() => replacer,
-        _ => return Ok(Value::from(object.to_json().to_string())),
+        // TODO: handle the error case
+        _ => return Ok(Value::from(object.to_json(ctx).unwrap().to_string())),
```
Razican, line 111: I think we can use `?` instead of `unwrap()` here, and the exception will be propagated.

boa/src/builtins/json/mod.rs
```diff
@@ -133,7 +134,10 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
                         )?),
                     );
                 }
-                Ok(Value::from(object_to_return.to_json().to_string()))
+                // TODO: Handle the error case
+                Ok(Value::from(
+                    object_to_return.to_json(ctx).unwrap().to_string(),
```
Razican, line 139: Same here, just use `?` instead of `unwrap()`.

boa/src/builtins/json/mod.rs
```diff
@@ -149,15 +153,17 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
...
                 .flatten()
             {
                 obj_to_return.insert(field.to_string(), value);
             }
         }
         Ok(Value::from(JSONValue::Object(obj_to_return).to_string()))
     } else {
-        Ok(Value::from(object.to_json().to_string()))
+        // TODO: handle the error case
+        Ok(Value::from(object.to_json(ctx).unwrap().to_string()))
```
Razican, line 166: Here too, we can use `?` instead of `unwrap()`.

boa/src/builtins/json/mod.rs
```diff
@@ -149,15 +153,17 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
         for field in fields {
             if let Some(value) = object
                 .get_property(&ctx.to_string(&field)?)
-                .map(|prop| prop.value.as_ref().map(|v| v.to_json()))
+                // TODO: handle the error case
+                .map(|prop| prop.value.as_ref().map(|v| v.to_json(ctx).unwrap()))
```
Razican, line 157: This is a bit trickier, since it's an `Option::map()`, but we can just return the `to_json()`, and after the `map()`, we call `transpose()` and then use `?`, to propagate the exception.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +151,103 @@ impl Value {
...
+                if obj.borrow().kind == ObjectKind::Array {
+                    let mut arr: Vec<JSONValue> = Vec::new();
+                    obj.borrow().properties.keys().for_each(|k| {
+                        if k != "length" {
+                            let value = self.get_field(k.to_string());
+                            if value.is_undefined() || value.is_function() {
+                                arr.push(JSONValue::Null);
+                            } else {
+                                arr.push(
+                                    self.get_field(k.to_string()).to_json(interpreter).unwrap(),
```
Razican, line 218: Here too, we should use `?` instead of `unwrap()`.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +151,103 @@ impl Value {
...
+                    });
+                    Ok(JSONValue::Array(arr))
+                } else {
+                    let mut new_obj = Map::new();
+                    obj.borrow().properties.keys().for_each(|k| {
+                        let key = k.clone();
+                        let value = self.get_field(k.to_string());
+                        if !value.is_undefined() && !value.is_function() {
+                            // TODO: Handle the errors
+                            new_obj.insert(key, value.to_json(interpreter).unwrap());
```
Razican, line 231: Another place where `?` should be used instead of `unwrap()`.

boa/src/builtins/value/mod.rs
Razican: OK, so I would suggest using `null` as default instead of panicking, and I'll try to find why is Firefox returning that.

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +151,103 @@ impl Value {
...
+            ValueData::String(ref str) => Ok(JSONValue::String(str.clone())),
+            ValueData::Rational(num) => Ok(JSONValue::Number(
+                JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
+            )),
+            ValueData::Integer(val) => Ok(JSONValue::Number(JSONNumber::from(val))),
+            ValueData::BigInt(_) => {
+                Err(interpreter.throw_type_error("BigInt value can't be serialized in JSON")?)
+            }
+            ValueData::Symbol(_) | ValueData::Undefined => {
+                // TODO: What should this return?
```
Razican, line 246: We should check this, I'm afraid of compiler panics xD

boa/src/builtins/value/mod.rs
n14little: I'm struggling with this one. It doesn't seem that we can use the `?` operator because we're inside of a for_each.

boa/src/builtins/value/mod.rs
n14little: Same here about being inside a for_each.

boa/src/builtins/value/mod.rs
n14little: Okay, that makes sense! Thanks @HalidOdat

boa/src/builtins/value/mod.rs
```diff
@@ -149,6 +151,103 @@ impl Value {
...
+            ValueData::String(ref str) => Ok(JSONValue::String(str.clone())),
+            ValueData::Rational(num) => Ok(JSONValue::Number(
+                JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
+            )),
+            ValueData::Integer(val) => Ok(JSONValue::Number(JSONNumber::from(val))),
+            ValueData::BigInt(_) => Err(interpreter
+                .throw_type_error("BigInt value can't be serialized in JSON")
+                .expect_err("throw_type_error should always return an error")),
+            ValueData::Symbol(_) | ValueData::Undefined => {
+                unreachable!("Symbols and Undefined JSON Values depend on parent type");
```
n14little, line 247: Replaying @Razican comment to not lose sight of it:

> We should check this, I'm afraid of compiler panics xD

**It went outdated.

boa/src/builtins/value/mod.rs
Razican: I would just use a `for`, it seems more accurate in this case:
```rust
for k in obj.borrow().properties.keys() {
    // ...
}
```

boa/src/builtins/value/mod.rs
Razican: OK, so this is handled by the "stringify` function, so it's expected that this panics, since it should never arrive here with any of these values.

boa/src/builtins/value/mod.rs
Razican: Yes, it's a good practise, sorry for the trouble. I have removed it.

boa/src/builtins/value/mod.rs
Razican: I would change the `for_each()` for a `for k in obj.borrow().properties.keys()` here too.

n14little: :heavy_check_mark:

boa/src/builtins/value/mod.rs
n14little: :heavy_check_mark:

boa/src/builtins/value/mod.rs
n14little: returning null. :heavy_check_mark:

n14little:
I think that this is ready to go now assuming that everybody is okay with not addressing the other prototypes for JSON types other than object (i.e. arrays, numbers, null, string, boolean).

boa/src/builtins/value/mod.rs
```diff
@@ -208,37 +208,39 @@ impl Value {
...
             ValueData::String(ref str) => Ok(JSONValue::String(str.clone())),
-            ValueData::Rational(num) => Ok(JSONValue::Number(
-                JSONNumber::from_f64(num).expect("Could not convert to JSONNumber"),
-            )),
+            ValueData::Rational(num) => {
+                if let Some(number) = JSONNumber::from_f64(num) {
+                    Ok(JSONValue::Number(number))
+                } else {
+                    Ok(JSONValue::Null)
+                }
```
Razican, line 240: I would use this:
```suggestion
                JSONNumber::from_f64(num).map(JsonValue::Number).unwrap_or(JSONValue::Null)
```

n14little: The suggestion wasn't quite right, but I got it updated.

boa/src/builtins/json/mod.rs
```diff
@@ -149,15 +149,16 @@ pub fn stringify(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Result
         for field in fields {
             if let Some(value) = object
                 .get_property(&ctx.to_string(&field)?)
-                .map(|prop| prop.value.as_ref().map(|v| v.to_json()))
+                .map(|prop| prop.value.as_ref().map(|v| v.to_json(ctx)))
                 .flatten()
```
HalidOdat, line 153: Here we can use the `Option::and_then` method so we avoid the `.flatten()`
```suggestion
                .and_then(|prop| prop.value.as_ref().map(|v| v.to_json(ctx)))
```

PR #468, opened 2020-06-09
Add BigInt.asIntN() and BigInt.asUintN() functions

### LINKED ISSUE #415: Implement `BigInt` static methods
**ECMASCript feature**

Implement `BigInt` static methods:
 - [ ] [`BigInt.asIntN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asIntN)
 - [ ] [`BigInt.asUintN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUntN)

More information:
 - [ECMAScript `BigInt` object](https://tc39.es/ecma262/#sec-bigint.asintn)
 - [MDN `BigInt` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)

Tropid:
This Pull Request fixes/closes #415.

It changes the following:
 - Adds a `BigInt.asUintN()` function
 - Adds a `BigInt.asIntN()` function
 - Adds a `to_index()` function to the interpreter (https://tc39.es/ecma262/#sec-toindex)
- Adds a `to_integer()` function to the interpreter (https://tc39.es/ecma262/#sec-tointeger, this is probably related to `ValueData::to_integer` which seems to behave different than the function described in the standard, it also returns a 32 bit integer instead of a 64 bit integer)
- Adds a `to_number()` to the interpreter (https://tc39.es/ecma262/#sec-tonumber, this is also similar to `value_to_rust_number` which behaves a bit different)
- Also adds a few tests (mostly from test262) for `to_index`, `to_integer`, `BigInt.asIntN`, `BigInt.asUintN`


boa/src/exec/mod.rs
```diff
@@ -200,6 +200,65 @@ impl Interpreter {
...
+        match *value.deref().borrow() {
+            ValueData::Null => Ok(0.0),
+            ValueData::Undefined => Ok(f64::NAN),
+            ValueData::Boolean(b) => Ok(if b { 1.0 } else { 0.0 }),
+            ValueData::String(ref string) => match string.parse::<f64>() {
+                Ok(number) => Ok(number),
+                Err(_) => Ok(0.0),
+            }, // this is probably not 100% correct, see https://tc39.es/ecma262/#sec-tonumber-applied-to-the-string-type
+            ValueData::Rational(number) => Ok(number),
+            ValueData::Integer(integer) => Ok(integer as f64),
```
HalidOdat, line 246:
```suggestion
            ValueData::Integer(integer) => Ok(f64::from(*integer)),
```
It's better if we use from conversion rather than `as` conversions, the `i32` can fit in `f64`

PR #470, opened 2020-06-10
Implement PreferredType enum

### LINKED ISSUE #466: Change `ToPrimitive()` (`to_primitive()`) hint to be an enum, instead of string
**What's wrong with the current implementation of `to_primitive`?**
The problem is that we use a string (`&str`) as the hint, so when we use `to_primitive` we always have to do a string comparison, which is not efficient.

To fix this we need to create a new enum `PreferredType` which would contain `String`, `Number` and `Default` field and use it to compare it instead of `&str`

More information:
 - [ECMAScript specification](https://tc39.es/ecma262/#sec-toprimitive)
 - [Boa `to_primitive` implementation](https://github.com/boa-dev/boa/blob/master/boa/src/exec/mod.rs#L256-L283)

Colin-Suckow:
This Pull Request fixes/closes #466 .

Implements a new enum called 'PreferredType'. This PR also fixes the existing calls to 'to_primitive' by replacing the 'Option<&str>' with an 'Option<PreferredType>'.

### DIFF
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -45,7 +45,8 @@ pub trait Executable {
     fn run(&self, interpreter: &mut Interpreter) -> ResultValue;
 }
 
-enum PreferredType {
+#[derive(Clone, Copy, Debug)]
+pub enum PreferredType {
     String,
     Number,
     Default,
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -45,6 +45,12 @@ pub trait Executable {
     fn run(&self, interpreter: &mut Interpreter) -> ResultValue;
 }
 
+enum PreferredType {
+    String,
+    Number,
+    Default,
+}
+
 /// A Javascript intepreter
 #[derive(Debug)]
 pub struct Interpreter {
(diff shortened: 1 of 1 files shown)


boa/src/exec/mod.rs
```diff
@@ -256,17 +263,17 @@ impl Interpreter {
     pub(crate) fn to_primitive(
         &mut self,
         input: &mut Value,
-        preferred_type: Option<&str>,
+        preferred_type: Option<PreferredType>,
```
HalidOdat, line 266: I think we should take in just `PreferredType` instead of `Option<PreferredType>` because when it is `None` it's `"default"`

boa/src/exec/mod.rs
```diff
@@ -256,17 +263,17 @@ impl Interpreter {
...
             ValueData::Object(_) => {
                 hint = match preferred_type {
                     None => "default",
                     Some(pt) => match pt {
-                        "string" => "string",
-                        "number" => "number",
-                        _ => "default",
+                        PreferredType::String => "string",
+                        PreferredType::Number => "number",
+                        PreferredType::Default => "default",
```
HalidOdat, line 276: Maybe we should make `ordinary_to_primitive` method take in `PreferredType` so we don't have to convert the enum to `&str`.

PR #471, opened 2020-06-10
create boolean objects with prototype

n14little:
This Pull Request address boolean values for #452 .

It changes the following:
 - Adds the prototype to boolean values created in `Value::from_json`

### DIFF
--- a/boa/src/builtins/json/tests.rs
+++ b/boa/src/builtins/json/tests.rs
@@ -19,7 +19,7 @@ fn json_sanity() {
     assert_eq!(
         forward(
             &mut engine,
-            r#"JSON.stringify({aaa: 'bbb'}) == '{"aaa":"bbb"}'"#
+            r#"JSON.stringify({aaa: 'bbb', ccc: true}) == '{"aaa":"bbb","ccc":true}'"#
         ),
         "true"
     );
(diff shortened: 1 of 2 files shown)


HalidOdat:
Hmmm. Checking what `JSON.parse("true")` on firefox and node returns is `"boolean"` not a `Boolean` object. I'm not sure if we can ever get an `Boolean` object from `parse`.

n14little:
> BTW I don't think that there can be a null or undefined object.

I'm not following.

HalidOdat:
> > BTW I don't think that there can be a null or undefined object.
> 
> I'm not following.

`null` and `undefined` are primitives that cannot be converted to Objects

n14little:
Where are null and undefined being converted into objects? Are you referring something that `JSONValue::Bool(v)` might be doing?

HalidOdat:
> Where are null and undefined being converted into objects? Are you referring something that `JSONValue::Bool(v)` might be doing?

Sorry I misunderstood the comment in #462 :sweat_smile:

Razican:
> Hmmm. Checking what `JSON.parse("true")` on firefox and node returns is `"boolean"` primitive not a `Boolean` object. I'm not sure if we can ever get an `Boolean` object from `parse`.

I guess this is the only thing left, right?

PR #474, opened 2020-06-10
Catchup new_lexer to master

Lan2u:


Lan2u:
Lots of changes on master, might be worth merging master into new_lexer now so that there are less conflicts down the line

PR #477, opened 2020-06-10
New Lexer: Minimal amount to allow compiling

Lan2u:
Updated various external references, disabled a large number of tests to allow the code to compile after changes in (#432) , acts as a base so I can measure progress (by tests passing) as I start reimplementing stuff.


boa/src/syntax/lexer/template.rs
```diff
@@ -0,0 +1,65 @@
...
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: 
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
+#[derive(Debug, Clone, Copy)]
+pub(super) struct TemplateLiteral {
+
+}
```
Razican, line 29:
```suggestion
pub(super) struct TemplateLiteral;
```

boa/src/syntax/lexer/error.rs
```diff
@@ -14,7 +14,7 @@ impl From<io::Error> for Error {
 
 impl Error {
     /// Creates a new syntax error.
-    pub(super) fn syntax<M>(err: M) -> Self
+    pub(crate) fn syntax<M>(err: M) -> Self
```
Razican, line 17: Why is this public at the crate level? I think we should only be creating lexer errors in the lexer.

boa/src/syntax/parser/tests.rs
```diff
@@ -34,10 +40,16 @@ where
...
-    let mut lexer = Lexer::new(js);
-    lexer.lex().expect("failed to lex");
+    let mut lexer = Lexer::new(js.as_bytes());
+    // lexer.lex().expect("failed to lex");
+
+    let mut tokens = Vec::new();
+
+    for token in lexer {
+        tokens.push(token.expect("failed to lex"));
+    }
```
Razican, line 50: This can be improved further:
```suggestion
    let tokens = lexer.collect::<Result<Vec<_>, _>().expect("failed to lex");
```

boa_cli/src/main.rs
```diff
@@ -115,11 +116,16 @@ arg_enum! {
...
-    lexer.lex().map_err(|e| format!("SyntaxError: {}", e))?;
-    Ok(lexer.tokens)
+    // Goes through and lexes entire given string.
+    let mut tokens = Vec::new();
+
+    for token in lexer {
+        tokens.push(token.map_err(|e| format!("Lexing Error: {}", e))?);
+    }
+
+    Ok(tokens)
```
Razican, line 128:
```suggestion
    lexer.collect::<Result<Vec<_>, _>().map_err(|e| format!("Lexing Error: {}", e))
```

boa_wasm/src/lib.rs
```diff
@@ -3,12 +3,15 @@ use wasm_bindgen::prelude::*;
...
-        .map_err(|e| JsValue::from(format!("Syntax Error: {}", e)))?;
+    let mut lexer = Lexer::new(src.as_bytes());
 
-    let tokens = lexer.tokens;
+    // Goes through and lexes entire given string.
+    let mut tokens = Vec::new();
+
+    for token in lexer {
+        tokens.push(token.map_err(|e| format!("Lexing Error: {}", e))?);
+    }
```
Razican, line 13:
```suggestion
    let tokens = lexer.collect::<Result<Vec<_>, _>().map_err(|e| format!("Lexing Error: {}", e))?;
```

boa/src/syntax/lexer/error.rs
Lan2u: I agree, the syntax function is currently used elsewhere (outside crate) and so this change was needed

Lan2u: I decided it was easier just to make it super for now and then go back and make the change later (starting at a baseline of code which compiles / can be tested)

Lan2u: This is where it is used: https://github.com/boa-dev/boa/blob/496278c46025d2a5ce1a13a4012bcc2ce28ded04/boa/src/syntax/ast/token.rs#L117-L130 

I'm wondering if the token.rs file should actually be in the lexer directory rather than ast because the lexer is the one generating tokens.

Lan2u: Commit f115892 shows the result of moving the token file

PR #485, opened 2020-06-12
updated issue #484

### LINKED ISSUE #484: `ToObject` (`to_object`) throw `undefined` when converting `ValueData::Integer` to `Number` object
**Describe the bug**

The bug is here: <https://github.com/boa-dev/boa/blob/master/boa/src/exec/mod.rs#L384>. When the `ValueData` is `Integer` it returns `Err(undefined)` this is wrong. 

We should have the same logic as the `Rational` case: <https://github.com/boa-dev/boa/blob/master/boa/src/exec/mod.rs#L398-L407>.

gvasquez11:
This Pull Request fixes/closes #484 

It changes the following:
 - Re-wrote when **ValueData** is **Integer**
 - Follows the same logic as the **Rational** case
 - Added a test case in builtins/number/tests.rs to handle integer number primitive to number object

### DIFF
--- a/boa/src/builtins/number/tests.rs
+++ b/boa/src/builtins/number/tests.rs
@@ -10,12 +10,12 @@ use crate::{
 #[test]
 fn integer_number_primitive_to_number_object() {
     let realm = Realm::create();
-    let mut engine = Interpreter::new(realm);   
+    let mut engine = Interpreter::new(realm);
 
     let scenario = r#"
         (100).toString() === "100"
     "#;
-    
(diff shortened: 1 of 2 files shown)


gvasquez11:
Thank you! I'm just trying to fix the cargo issue so all test past.

HalidOdat:
> Thank you! I'm just trying to fix the cargo issue so all test past.

You need to run `crago fmt` there just rust style fixes :)

Issue #456, opened 2020-06-04
[lexer] in token position counting end boundary should be excluded


croraf:
For example if you have JS code 
```
let a = 5;
```
One of the tokens the lexer will output it will say: `Token: Identifier "a", start position: 5, end position: 5`

Instead it should say `... end position: 6`

Razican:
If I understand the title right, if the boundary should be excluded, then `5` is correct, right? `6` would also include the boundary.

So, in that example, the 5th token is the `a` token. The first position where that token appears is the `5`, and the last position where that token appears is the `5` too.

For `let` for example, the first position where it appears is `1`, and the last one is `3`.

croraf:
No. The right boundary should be excluded meaning that if the right boundary is 5, the character at position 5 is not part of the token.

So for `let` you will have boundaries 1 and 4, including characters at position 1, 2 and 3. Excluding the character at position 4.

Razican:
> No. The right boundary should be excluded meaning that if the right boundary is 5, the character at position 5 is not part of the token.
> 
> So for `let` you will have boundaries 1 and 4, including characters at position 1, 2 and 3. Excluding the character at position 4.

So, then, for `a`, boundaries should be `5` and `6`, with the `6` excluded, right?

croraf:
Yes

Razican:
> Yes

Sorry, then I don't understand the bug. The span is set correctly, from position 5 to position 5, right?

IovoslavIovchev:
> 
> 
> > Yes
> 
> Sorry, then I don't understand the bug. The span is set correctly, from position 5 to position 5, right?

I think the issue is that the end boundary should be exclusive, e.g. in the example `a`'s span is `[5, 5]`, when it should be `[5, 6)`.

Lan2u:
This will be fixed as part of #486.

PR #486, opened 2020-06-12
New lexer

Lan2u:
This PR includes the new lexer + parser made to work with goal symbols.

Addresses: #294 primarily and #456 as a side

Hopefully once this is finished we can complete #12


boa/src/syntax/lexer/template.rs
```diff
@@ -23,7 +23,7 @@ macro_rules! template_match {
 pub(super) struct TemplateLiteral;
 
 impl TemplateLiteral {
-    /// Creates a new string literal lexer.
+    /// Creates a new template literal lexer.
     pub(super) fn new() -> Self {
         Self {}
     }
```
Razican, line 29: I don't think we need a `new()` function for a structure with no members, right?

Lan2u: Isn't it best to always provide a new function to hide the internal structure of a struct?

Razican: Hmmm if it's a unit structure, I don't think it's needed.

boa/src/syntax/lexer/cursor.rs
```diff
@@ -15,25 +15,24 @@ impl<R> Cursor<R> {
...
         let current_line = self.pos.line_number();
         let next_column = self.pos.column_number() + 1;
         self.pos = Position::new(current_line, next_column);
     }
 
-    /// Advances the position to the next column.
+    /// Advances the position to the next line.
     #[inline]
-    fn next_line(&mut self) {
+    pub(super) fn next_line(&mut self) {
```
Razican, line 28: I don't think this should be called from `super`. This should only be called from the cursor, when finding a "next line" character

boa/src/syntax/lexer/cursor.rs
```diff
@@ -15,25 +15,24 @@ impl<R> Cursor<R> {
...
         let next_line = self.pos.line_number() + 1;
         self.pos = Position::new(next_line, 1);
     }
 
     /// Performs a carriage return to modify the position in the source.
     #[inline]
-    fn carriage_return(&mut self) {
+    pub(super) fn carriage_return(&mut self) {
```
Razican, line 35: I don't think this should be called from `super`. This should only be called from the cursor, when encountering a `\r` character.

boa/src/syntax/lexer/cursor.rs
```diff
@@ -70,31 +69,94 @@ where
...
+                        buf.push(ch);
+                    }
+                }
+            }
+        }
     }
 
     /// Retrieves the given number of characters and adds them to the buffer.
-    pub(super) fn take(&mut self, count: usize, buf: &mut String) -> io::Result<()> {
+    pub(super) fn _take(&mut self, _count: usize, _buf: &mut String) -> io::Result<()> {
```
Razican, line 147: There is no need to replace all these with an underscore (especially the function name, should never start with an underscore).

Warnings are expected while in development.

boa/src/syntax/lexer/cursor.rs
```diff
@@ -70,31 +69,94 @@ where
...
         unimplemented!()
     }
 
     /// It will fill the buffer with checked ASCII bytes.
-    pub(super) fn fill_bytes(&mut self, buf: &[u8]) -> io::Result<()> {
+    pub(super) fn fill_bytes(&mut self, _buf: &mut [u8]) -> io::Result<()> {
```
Razican, line 152: There is no need to replace all these with an underscore (especially the function name, should never start with an underscore).

Warnings are expected while in development.

boa/src/syntax/lexer/cursor.rs
```diff
@@ -70,31 +69,94 @@ where
...
         unimplemented!()
     }
 
     /// Retrieves the next character as an ASCII character.
     ///
     /// It will make sure that the next character is an ASCII byte, or return an error otherwise.
-    pub(super) fn next_as_byte(&mut self) -> Option<io::Result<u8>> {
+    pub(super) fn _next_as_byte(&mut self) -> Option<io::Result<u8>> {
```
Razican, line 159: We should not start function names with an underscore.

boa/src/syntax/lexer/cursor.rs
```diff
@@ -113,8 +175,11 @@ where
         };
 
         match chr {
-            Some(Ok('\r')) => self.carriage_return(),
-            Some(Ok('\r')) | Some(Ok('\u{2028}')) | Some(Ok('\u{2029}')) => self.next_line(),
+            Some(Ok('\r')) => {
+                self.carriage_return();
+                self.next_line()
```
Razican, line 180: Why was this changed? A carriage return does not change the line number, AFAIK. Also, a `next_line()` call also does a carriage return, so the first carriage return wouldn't make sense.

boa/src/syntax/lexer/error.rs
```diff
@@ -20,13 +22,26 @@ impl Error {
...
+    ///
+    /// This error is used to represent the case where a piece of javascript
+    /// cannot be lexed/parsed because it is in invalid when strict mdoe is
+    /// enabled.
+    pub(super) fn strict<M>(err: M) -> Self
+    where
+        M: Into<Box<str>>,
+    {
+        Self::StrictMode(err.into())
+    }
```
Razican, line 36: Could we add a link to where this error is defined in the spec?

boa/src/syntax/lexer/identifier.rs
```diff
@@ -0,0 +1,71 @@
...
+
+impl<R> Tokenizer<R> for Identifier {
+    fn lex(&mut self, cursor: &mut Cursor<R>, start_pos: Position) -> Result<Token, Error>
+    where
+        R: Read,
+    {
+        let mut buf = self.init.to_string();
+
+        loop {
+            match cursor.peek() {
```
Razican, line 38: We should use a `while let Some(c) = cursor.peek() {`.
Then, we can just do `let c = c?;` and forget about errors.

boa/src/syntax/lexer/identifier.rs
```diff
@@ -0,0 +1,71 @@
...
+                Some(Err(_e)) => {
+                    // TODO handle error.
+                }
+            }
+        }
+        let tk = match buf.as_str() {
+            "true" => TokenKind::BooleanLiteral(true),
+            "false" => TokenKind::BooleanLiteral(false),
+            "null" => TokenKind::NullLiteral,
+            "NaN" => TokenKind::NumericLiteral(Numeric::Rational(f64::NAN)),
```
Razican, line 59: This has been changed in production `NaN` is not a number, but an identifier in the global object.

boa/src/syntax/lexer/lexer_old.rs
```diff
@@ -477,7 +477,7 @@ impl<'a> Lexer<'a> {
     }
 
     /// Runs the lexer until completion, returning a [LexerError] if there's a syntax issue, or an empty unit result
-    ///
+    //
```
Razican, line 480: This change should be reverted.

boa/src/syntax/lexer/mod.rs
```diff
@@ -3,23 +3,41 @@
...
 
+mod number;
+
+mod operator;
+
+mod spread;
+
+mod regex;
+
+mod identifier;
```
Razican, line 27: I see that you added a lot of space here. Is there any reason for that?

boa/src/syntax/lexer/mod.rs
```diff
@@ -3,23 +3,41 @@
...
+
 // Temporary disabled while lexer in progress.
```
Razican, line 29: I'm guessing this si no longer true, right?

boa/src/syntax/lexer/mod.rs
```diff
@@ -56,7 +74,7 @@ impl<R> Lexer<R> {
     }
 
     /// Sets the goal symbol for the lexer.
-    pub(crate) fn set_goal(&mut self, elm: InputElement) {
+    pub(crate) fn _set_goal(&mut self, elm: InputElement) {
```
Razican, line 77: We should not use functions starting with an underscore.

boa/src/syntax/lexer/mod.rs
```diff
@@ -81,14 +99,16 @@ where
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum InputElement {
     Div,
-    RegExp,
-    RegExpOrTemplateTail,
-    TemplateTail,
+    _RegExp,
+    _RegExpOrTemplateTail,
+    _TemplateTail,
```
Razican, line 104: We should not start variants with underscores.

boa/src/syntax/lexer/mod.rs
```diff
@@ -81,14 +99,16 @@ where
...
 }
 
 impl Default for InputElement {
     fn default() -> Self {
-        todo!("what is the default input element?")
+        InputElement::Div
+        // Decided on InputElementDiv as default for now based on documentation from
+        // <https://tc39.es/ecma262/#sec-ecmascript-language-lexical-grammar>
```
Razican, line 111: If I understand the spec correctly, the default one cannot be `Div`, since by default, if you find `/` as the first parameter for the lexer, this means the beginning of a RegExp literal:
```javascript
/hello/.test("hello");
```

boa/src/syntax/lexer/regex.rs
```diff
@@ -0,0 +1,109 @@
...
+/// [mdn]:
+#[derive(Debug, Clone, Copy)]
+pub(super) struct RegexLiteral;
+
+impl RegexLiteral {
+    /// Creates a new regex literal lexer.
+    pub(super) fn new() -> Self {
+        Self {}
+    }
+}
```
Razican, line 26: I think we don't need this.

boa/src/syntax/lexer/regex.rs
```diff
@@ -0,0 +1,109 @@
...
+                        }
+                        _ => body.push(c),
+                    }
+                }
+            }
+        }
+
+        unimplemented!(
+            "Regex handling, requires ability to peek more than a single character ahead"
+        );
```
Razican, line 87: So, the thing here is that this depends on the goal symbol. A `/` in a `Div` goal symbol is always a dividion, no need to check further. A `/` in an `InputElementRegExp` goal symbol is always the start / end of a regular expression, so no need to peek further.

This is the main objective of this change, to allow using goal symbols.

boa/src/syntax/lexer/spread.rs
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+impl<R> Tokenizer<R> for SpreadLiteral {
+    fn lex(&mut self, cursor: &mut Cursor<R>, start_pos: Position) -> Result<Token, Error>
+    where
+        R: Read,
+    {
+        // . or ...
+        match cursor.next_is('.') {
```
Razican, line 25: I think this can be done easier by using `if cursor.next_is('.')? {`.

boa/src/syntax/lexer/token.rs
```diff
@@ -60,7 +60,7 @@ impl Display for Token {
 /// Represents the type differenct types of numeric literals.
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Clone, PartialEq, Debug)]
-pub enum NumericLiteral {
+pub enum Numeric {
```
Razican, line 63: Why was this name changed? It's still a numeric literal, right?

boa/src/syntax/lexer/token.rs
```diff
@@ -247,6 +247,9 @@ pub enum TokenKind {
 
     /// Indicates the end of a line (`\n`).
     LineTerminator,
+
+    /// Indicates a comment, the content isn't stored.
+    Comment,
```
Razican, line 252: I don't think we need comments as tokens. A comment can only generate a line terminator token.

ignore_temp/number_grammar_based.rs
```diff
@@ -0,0 +1,261 @@
...
+            return Ok(token); // Parsed successfully.
+        }
+        if let Ok(token) = DecimalBigIntegerLiteral::new(self.init).lex(cursor, start_pos) {
+            return Ok(token); // Parsed successfully.
+        } 
+        if let Ok(token) = NonDecimalIntegerLiteral::new(self.init).lex(cursor, start_pos) {
+            return Ok(token); // Parsed successfully.
+        }
+
+        Err(Error::Reverted())
```
Razican, line 78: Just for my understanding, why are we returning this?

ignore_temp/number_grammar_based.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+} 
+
+impl<R> Tokenizer<R> for DecimalLiteral {
+    fn lex(&mut self, cursor: &mut Cursor<R>, start_pos: Position) -> Result<Token, Error>
+    where
+        R: Read,
+    {
+
+
```
Razican, line 107:
```suggestion

```

ignore_temp/number_grammar_based.rs
```diff
@@ -0,0 +1,261 @@
...
+            Ok(dil_token) => {
+                // DecimalIntegerLiteral
+
+                if cursor.next_is('.')? {
+                    // DecimalIntegerLiteral.
+                    
+                    // Consume the '.'
+                    cursor.next();
+
+
```
Razican, line 119:
```suggestion

```

boa/src/syntax/lexer/cursor.rs
Lan2u: Yeah makes sense, when I moved the code from the old lexer over (which did make these calls) I made adjustments so that it would work while I got other parts working.

boa/src/syntax/lexer/cursor.rs
Lan2u: Will revert before marking branch as ready, how come function names should never start with an underscore? (I thought it was just used to mark a function as unused currently?)

boa/src/syntax/lexer/cursor.rs
Razican: OK, For the future, we should restrict it, I think that way we can abstract all that information from the lexer itself.

boa/src/syntax/lexer/cursor.rs
Lan2u: > Why was this changed? A carriage return does not change the line number, AFAIK. Also, a `next_line()` call also does a carriage return, so the first carriage return wouldn't make sense.

I took this table in the spec to indicate a carriage return was a line terminator https://tc39.es/ecma262/#sec-line-terminators perhaps I misunderstood. 

I didn't notice the next_line did a carriage return so yeah that can be fixed

boa/src/syntax/lexer/cursor.rs
Razican: Well, a function should be used or not present. If we are in the process of developing a new feature, then it's understandable that we get "unused" warnings, but we should not silence them, so that we know where to look before moving code to the master branch.

boa/src/syntax/lexer/error.rs
Lan2u: You mean https://tc39.es/ecma262/#sec-strict-mode-code + https://tc39.es/ecma262/#prod-annexB-LegacyOctalIntegerLiteral ?

boa/src/syntax/lexer/lexer_old.rs
Lan2u: Yep, will revert.

boa/src/syntax/lexer/mod.rs
Lan2u: I had macro_use between them so I could use the literal_start!() type macros but then scrapped that idea for now as I couldn't get it working if there are multiple possible start-symbols. 

I shall clean these up

boa/src/syntax/lexer/mod.rs
Lan2u: Correct

boa/src/syntax/lexer/mod.rs
Lan2u: Hmmm, to be honest, I'm not really sure what should be default - I'll look more into this.

boa/src/syntax/lexer/cursor.rs
Razican: The thing is that even if a `\r` is considered a "line terminator token", when we report lines, we have to take into account how do editors show this. I think a `\r` character just moves the current column to number `1`.

boa/src/syntax/lexer/regex.rs
Lan2u: Yeah, it can be shortened as I don't need to store anything in the struct.

boa/src/syntax/lexer/regex.rs
Lan2u: Yeah, this code is basically just straight moved over and I haven't really looked at regex much yet so this is still todo

boa/src/syntax/lexer/spread.rs
Lan2u: Yes definitely

boa/src/syntax/lexer/error.rs
Razican: Checking https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_octal

It seems it's just a normal syntax error, right?

boa/src/syntax/lexer/token.rs
Lan2u: I think it is easier to understand named like this as it is wrapped inside a NumericLiteral as it seemed a bit weird having a numeric literal in a numeric literal

boa/src/syntax/lexer/token.rs
Lan2u: I needed something to return from the comment lex impl and this seemed to be the cleanest way.

Razican: Hmm we could just return `()`, right?

ignore_temp/number_grammar_based.rs
Lan2u: I was playing around with writing the lexer like one I did at university (in Haskell) where you tried to lex basically using the grammar directly and then if you failed one possibility you reverted and then tried the next. 

I don't think that is necessarily the best way to do it here because it is less performant and requires making a copy of the data for each call to .lex() (or using a stack to push the data back onto) but I gave it a go mostly out of curiosity.

ignore_temp/number_grammar_based.rs
Lan2u: I'll be removing the ignore_temp/ folder before marking the branch as ready

boa/src/syntax/lexer/token.rs
Lan2u: > Hmm we could just return `()`, right?

It expects a token because of the trait signature

boa/src/syntax/lexer/error.rs
Lan2u: Doesn't this mean that we then can't differentiate what caused the syntax error? Would it perhaps be better to have the syntax error to have a caused by field or something?

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,351 @@
...
+            NumericKind::Integer(base) => {
+                if let Ok(num) = i32::from_str_radix(&buf, base as u32) {
+                    if exp_str == "" {
+                        Numeric::Integer(num)
+                    } else {
+                        let n = i32::from_str(&exp_str).map_err(|_| Error::syntax("Could not convert value to f64"))?;
+                    
+                        Numeric::Integer(
+                            num * i32::pow(10, n as u32)
+                        )
```
Lan2u, line 321: Here I return an integer (if possible) even if exponent notation is used. 

Previously a rational was returned but I believe an integer is more performant - thoughts?

Lan2u: Actually disregard this is probably a bad idea as it might not fit.

boa/src/syntax/lexer/tests.rs
```diff
@@ -264,283 +152,387 @@ fn check_keywords() {
...
 
-    let mut lexer = Lexer::new(s);
-    lexer.lex().expect("failed to lex");
+    let mut lexer = Lexer::new(s.as_bytes());
 
-    assert_eq!(lexer.tokens[0].span(), span((1, 1), (1, 1)));
-    assert_eq!(lexer.tokens[1].span(), span((1, 2), (2, 1)));
-    assert_eq!(lexer.tokens[2].span(), span((2, 1), (2, 1)));
-    assert_eq!(lexer.tokens[3].span(), span((2, 2), (3, 1)));
+    assert_eq!(lexer.next().unwrap().unwrap().span(), span((1, 1), (1, 1)));
```
Lan2u, line 271: Here it seems the old lexer used inclusive end bounds (issue #456 ) and that the new line symbols were returned as tokens? 

Am I right is saying that we should be using exclusive bounds and that control characters such as new line shouldn't be produced by lexer.next()?  (And therefore that this test is wrong and needs updating)

boa/src/syntax/lexer/mod.rs
Lan2u: I have changed this to InputElementRegExpOrTemplateTail based on the spec at https://tc39.es/ecma262/#sec-ecmascript-language-lexical-grammar as using Div has the problems as you say and InputElementRegExp can't be used as it would allow the code to contain a singular RightBracePunctuator which testing on js-fiddle it doesn't.

boa/src/syntax/lexer/tests.rs
```diff
@@ -264,283 +152,438 @@ fn check_keywords() {
...
+        TokenKind::Punctuator(Punctuator::StrictEq),
+        TokenKind::NumericLiteral(Numeric::Integer(1)),
+        TokenKind::Punctuator(Punctuator::Div),
+        TokenKind::Identifier("b".into()),
+        TokenKind::Punctuator(Punctuator::Semicolon)
+    ];
 
-    assert_eq!(lexer.tokens[11].span(), span((1, 37), (1, 37)));
+    expect_tokens(&mut lexer, &expected);
 }
```
Lan2u, line 275: @Razican 

I've re-written this test to show that the exactly the right tokens are returned and like this, the test passes (now that goal symbols have been partially added in). 

Does this test look ok to you? / does it show that the desired result is being produced or should we be checking span's aswell?

boa/src/syntax/lexer/tests.rs
Razican: We should use exclusive bounds, but line terminators must be produced by `lexer.next()`, since even if in most cases they are just white space, they have some roles in automatic semicolon insertion.

boa/src/syntax/lexer/tests.rs
Razican: This is good for me. Still, since we need to specify the goal symbol, this test will need to change, as the lexer will need to be called only from the parser.

ignore_temp/number_grammar_based.rs
Razican: Yep, we need to be careful with performance. This should only be used if there is no other option.

In theory, for any given goal symbol, the maximum amount of non-conclusive characters is 3, for HTML comments vs `a <!-10` for example. There is some information on this on the related links of #294, the original issue we are trying to solve.

This for example is an interesting read: https://v8.dev/blog/understanding-ecmascript-part-3

boa/src/syntax/lexer/token.rs
Razican: Hmmm actually, all comments return a line terminator, right?
Well, a single line comment, in theory won't be consuming the rightmost new line, but we can assume there will be one (a new line at the end of a file doesn't matter much):

https://tc39.es/ecma262/#sec-comments

Lan2u: Do we return line terminators? I thought we kinda just ignored them

boa/src/syntax/lexer/tests.rs
Lan2u: Ah right ok, I think I'll need to add some tests in for this to check line terminators are returned (in most tests currently they aren't).

boa/src/syntax/lexer/tests.rs
Lan2u: Can't we sometimes infer the goal symbol? For example there are times when only a certain goal symbol is valid and I've added some of these into the code

Razican: Maybe we can, yes, do you have an example?

boa/src/syntax/lexer/number.rs
Razican: Actually, I think it's a very good idea, but we need to check the arithmetic by using `checked_mul()` and `checked_pow()`.

Then, I would use `try_from()` if possible to convert the exponent. If everything is ok, then, generate an integer, if not, use floating point arithmetics.

boa/src/syntax/lexer/tests.rs
Lan2u: `
_ if next_chr.is_digit(10) => {
                let result = NumberLiteral::new(next_chr, strict_mode).lex(&mut self.cursor, start);
                // A regex may not directly follow a NumericLiteral but a DivPunctuator may.
                // Note that the goal cannot be set to InputElementTemplateTail at this point as a TemplateSubstitutionTail would be invalid.
                self.set_goal(InputElement::Div);
                result
            }
            _ if next_chr.is_alphabetic() || next_chr == '$' || next_chr == '_' => {
                let result = Identifier::new(next_chr).lex(&mut self.cursor, start);

                // A regex may not directly follow an Identifier but a DivPunctuator may.
                // Note that the goal cannot be set to InputElementTemplateTail at this point as a TemplateSubstitutionTail would be invalid.
                self.set_goal(InputElement::Div);
                result
            }

`

boa/src/syntax/lexer/number.rs
Lan2u: Done in recent commit

Lan2u:
Found that non-english characters in string literals such as in the new "non_english_str" test were not being parsed as UTF-8 successfully, this is fixed in #c1faef7.

Razican:
> Found that non-english characters in string literals such as in the new "non_english_str" test were not being parsed as UTF-8 successfully, this is fixed in #c1faef7.

Hmmm I was trying to do it without having to go to the heap. Maybe I should check it again, but I think something similar to that should work.

Lan2u:
> > Found that non-english characters in string literals such as in the new "non_english_str" test were not being parsed as UTF-8 successfully, this is fixed in #c1faef7.
> 
> Hmmm I was trying to do it without having to go to the heap. Maybe I should check it again, but I think something similar to that should work.

My next step with stuff is going to be modifying the parser to try and get the lexer/parser/goal-symbol relationship working - do we have any progress on this already?

boa/src/syntax/lexer/tests.rs
Razican: Yep, most of the times this is what we want. There are only two cases where we depend on the goal symbol: The `/` character, and the template tail.

Razican:
> My next step with stuff is going to be modifying the parser to try and get the lexer/parser/goal-symbol relationship working - do we have any progress on this already?

We have not progress on this yet. What we should probably do is to have the cursor do all the work, and have a `set_goal_symbol()` method that would change the goal symbol. The cursor should call `next()` on the lexer, so it should own the lexer.

Lan2u:
> > My next step with stuff is going to be modifying the parser to try and get the lexer/parser/goal-symbol relationship working - do we have any progress on this already?
> 
> We have not progress on this yet. What we should probably do is to have the cursor do all the work, and have a `set_goal_symbol()` method that would change the goal symbol. The cursor should call `next()` on the lexer, so it should own the lexer.

Currently I've gone for making the parser object do it (so the parser owns the lexer and has next/peek etc.)

Lan2u:
I disabled the Expression (lexer) benchmark because I believe it needs to be modified to utilize the parser and the lexer otherwise it won't lex due to not knowing the right goal symbols to use.

boa/src/syntax/lexer/template.rs
Lan2u: Fixed

boa/src/syntax/lexer/cursor.rs
Lan2u: Fixed

boa/src/syntax/lexer/cursor.rs
Lan2u: Fixed

boa/src/syntax/lexer/cursor.rs
Lan2u: Fixed

boa/src/syntax/lexer/cursor.rs
Lan2u: Fixed

boa/src/syntax/lexer/cursor.rs
Lan2u: Fixed

boa/src/syntax/lexer/token.rs
Lan2u: Are you happy with this change or do you think it should be kept as NumericLiteral?

boa/src/syntax/lexer/token.rs
Lan2u: What about the case https://jsfiddle.net/op7t13dc/

boa/src/syntax/lexer/identifier.rs
Lan2u: Because the error returned is a reference (&) the ? notation fails.

Lan2u: > We should use a `while let Some(c) = cursor.peek() {`.
> Then, we can just do `let c = c?;` and forget about errors.

Because .peek() returns a & reference to the error the ? notation fails (I do think if possible we should try and make peek return Result<&Token, Error> so ? can be used).

boa/src/syntax/lexer/identifier.rs
Lan2u: Will need to fix this during the merge

boa/src/syntax/lexer/lexer_old.rs
Lan2u: Have deleted the file as it is no longer needed (it is in the github history).

boa/src/syntax/lexer/number.rs
```diff
@@ -138,84 +138,80 @@ impl<R> Tokenizer<R> for NumberLiteral {
...
-                    return Ok(Token::new(
-                        TokenKind::NumericLiteral(Numeric::BigInt(0.into())),
-                        Span::new(start_pos, cursor.pos()),
-                    ));
-                }
-                Some(Ok(ch)) => {
-                    if ch.is_digit(8) {
-                        // LegacyOctalIntegerLiteral
-                        if self.strict_mode {
-                            // LegacyOctalIntegerLiteral is forbidden with strict mode true.
```
Lan2u: @Razican 

I modified the lexer cursor peek to return  Option<Result<char, io::Error>> rather than  Option<&io::Result<char>>. 

This means that char is copied on return from peek however it provides the following:
 1. We don't have to store an error in peeked, if there is an io error it is returned immediately and can be dealt with (otherwise you jam peeked with an error value which should have been handled already and might now be fixed).
 
 2. We can remove areas where the result of peek is compared in a match branch using lots of Some(Ok()) such as the example here - this should hopefully provide a small performance boost which should offset any losses from copying char.

Lan2u:
I have merged master into this branch so after I have completed some fixes this branch will be ready for review/merge.

boa/src/syntax/parser/error.rs
```diff
@@ -1,9 +1,6 @@
 //! Error and result implementation for the parser.
-use crate::syntax::ast::{
-    position::Position,
-    token::{Token, TokenKind},
-    Node,
-};
+use crate::syntax::ast::{position::Position, Node};
+use crate::syntax::lexer::{Error as LexError, Token, TokenKind};
```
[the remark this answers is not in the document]
Lan2u: I went for LexError instead of LexerError to match the usage of ParseError instead of ParserError

boa/src/syntax/lexer/comment.rs
```diff
@@ -1,15 +1,80 @@
...
+/// Lexes a block (multi-line) comment.
+///
+/// Assumes that the initial '/*' is already consumed.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]:
+/// [mdn]:
```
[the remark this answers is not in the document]
Lan2u: Fixed

boa/src/syntax/parser/expression/assignment/mod.rs
```diff
@@ -105,25 +113,29 @@ impl TokenParser for AssignmentExpression {
...
 
+        cursor.set_goal(InputElement::Div);
+
         let mut lhs = ConditionalExpression::new(self.allow_in, self.allow_yield, self.allow_await)
             .parse(cursor)?;
 
-        if let Some(tok) = cursor.next() {
-            match tok.kind {
+        if let Some(tok) = cursor.peek() {
+            match tok?.kind() {
```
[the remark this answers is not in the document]
Lan2u: These statements (I believe?) have a different meaning, .peek() returns an Option and we don't want to ? the option we want to ? the token inside the option.

boa/src/syntax/lexer/identifier.rs
Lan2u: Yeah fixed

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,374 @@
...
+                    Numeric::Rational(
+                        r
+                    )
+                } else {
+                    let n = f64::from_str(&exp_str).map_err(|_| Error::syntax("Could not convert value to f64"))?;
+                    Numeric::Rational(
+                        r * f64::powf(10.0, n)
+                    )
+                }
+            }
```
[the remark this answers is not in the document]
Lan2u: Yes, actually the number lexing is kinda a mess now I look back so it is cleaned up in recent commit

boa/src/syntax/lexer/comment.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Lexes a single line comment.
+///
+/// Assumes that the initial '//' is already consumed.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]:
+/// [mdn]:
```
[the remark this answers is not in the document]
Lan2u: Fixed

boa/src/syntax/lexer/error.rs
Lan2u: Removed as per spec

boa/src/syntax/lexer/identifier.rs
```diff
@@ -0,0 +1,71 @@
...
+/// Identifier or keyword lexing.
+///
+/// This currently includes boolean/NaN lexing.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]:
+/// [mdn]:
```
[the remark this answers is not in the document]
Lan2u: Fixed

boa/src/syntax/lexer/identifier.rs
Lan2u: Now that .peek() returns an Error rather than &Error this change has been made.

boa/src/syntax/parser/expression/assignment/mod.rs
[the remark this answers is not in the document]
HalidOdat: I see. then we can use `.transpose()` this will convert `Option<io::Result<u8>>` into `io::Result<Option<u8>>` and then we can use the try operator.
```rust
	if let Some(tok) = cursor.peek().transpose()? {
            match tok.kind() {
```

Or should we restructure the way we return from `.next()` and `.peek()` instead of  
`Option<io::Result<u8>>`, we return a `Option<u8>` this would simplefy the code, do we really need to have the io error or can we ignore it?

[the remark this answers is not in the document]
Lan2u: We need the IO error because a call to parser::Cursor.peek() may invoke the lexer which may trigger an IO error.

Re: transpose I'll make the change

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,349 @@
...
+                Numeric::BigInt(
+                    BigInt::from_string_radix(&buf, base as u32).expect("Could not convert to BigInt")
+                    )
+            }
+            NumericKind::Rational /* base: 10 */ => {
+                match f64::from_str(&buf) {
+                    Ok(val) => Numeric::Rational(val),
+                    Err(e) => return Err(Error::syntax(format!("Unable to parse rational number, reason: {}", e.to_string())))
+                }
+            }
```
HalidOdat, line 292: Here we should use `expect` becasue we want it to crash if it fails to lex, we already checked if it was valid so this cannot fail, if it does that means our floating point lexing logic is faulty. this can become a very hard bug.

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,349 @@
...
+                } else {
+                    let b = f64::from(base);
+                    let mut result = 0.0_f64;
+                    for c in buf.chars() {
+                        if let Some(val) = c.to_digit(base as u32) {
+                            let digit = f64::from(val);
+                            result = result * b + digit;
+                        } else {
+                            return Err(Error::syntax("Unrecognised numerical digit encountered"));
+                        }
```
HalidOdat, line 327: Here too, we should use `expect`

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,349 @@
...
+                    }
+
+                    if exp_str.is_empty() {
+                        Numeric::Rational(
+                            result
+                        )
+                    } else {
+                        let n = i32::from_str(&exp_str).map_err(|_| Error::syntax("Could not convert value to f64"))?;
```
HalidOdat, line 335: Here too.

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,349 @@
...
+                    TokenKind::NumericLiteral(Numeric::Integer(0)),
+                    Span::new(start_pos, cursor.pos()),
+                ));
+            }
+        }
+
+        // Consume digits until a non-digit character is encountered or all the characters are consumed.
+        cursor.take_until_pred(&mut buf, &|c: char| c.is_digit(kind.base()))?;
+
+        let exp_str = &mut String::new();
```
HalidOdat, line 218: why do we need `exp_str` we can just append to `buf`, right? and use `buf`

Lan2u: from_str_radix can't handle exponent notation in the form 10E2 as it cannot differentiate between that and hex so I think they need to be separate

boa/src/syntax/lexer/number.rs
Lan2u: Ah ok, this makes sense - will make changes

boa/src/syntax/lexer/number.rs
HalidOdat: > `from_str_radix` can't handle exponent notation in the form 10E2 as it cannot differentiate between that and hex so I think they need to be separate

why not do something like for integer case:
```
if let Ok(integer) = i32::from_str_radix(&buf, base) {
	// return integer;
} else {
	let b = f64::from(base);
    let mut result = 0.0_f64;
    for c in buf.chars() {
        let digit = f64::from(c.to_digit(base as u32));
    	result = result * b + digit;
    }
	// return result
}
```

Lan2u: Will this mean that if we have an Exponent the result is always a rational?

HalidOdat: Yes. It will. we could have a case to see if it fits in a `i32` like:
```rust
if let Ok(integer) = i32::from_str_radix(&buf, base) {
	// return integer;
} else {
	let b = f64::from(base);
    let mut result = 0.0_f64;
    for c in buf.chars() {
        let digit = f64::from(c.to_digit(base as u32));
    	result = result * b + digit;
    }
	if let Ok(integer) = i32::try_from(result) {
		// return integer;
	}
	// return result 
}
```

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,337 @@
...
+        None => {
+            return Err(Error::syntax("No exponential value found"));
+        }
+    }
+
+    // Consume the decimal digits.
+    cursor.take_until_pred(buf, &|c: char| c.is_digit(kind.base()))?;
+
+    Ok(())
+}
```
HalidOdat, line 107:
```suggestion
    match cursor.next().transpose()? {
        Some('+') => {
            buf.push('+');
            if !cursor.next_is_pred(&|c: char| c.is_digit(kind.base()))? {
                // A digit must follow the + or - symbol.
                return Err(Error::syntax("No digit found after + symbol"));
            }
        }
        Some('-') => {
            buf.push('-');
            if !cursor.next_is_pred(&|c: char| c.is_digit(kind.base()))? {
                // A digit must follow the + or - symbol.
                return Err(Error::syntax("No digit found after - symbol"));
            }
        }
        Some(c) if c.is_digit(kind.base()) => {
            buf.push(c);
        }
        Some(c) => {
            return Err(Error::syntax(format!(
                "When lexing exponential value found unexpected char: '{}'",
                c
            )));
        }
        None => {
            return Err(Error::syntax("No exponential value found"));
        }
    }

    // Consume the decimal digits.
    cursor.take_until_pred(buf, &|c: char| c.is_digit(kind.base()))?;

    Ok(())
}
```
We can use `.transpose()?` here to simplify the cases

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,337 @@
...
+        let mut buf = self.init.to_string();
+
+        // Default assume the number is a base 10 integer.
+        let mut kind = NumericKind::Integer(10);
+
+        let c = cursor.peek();
+
+        if self.init == '0' {
+            if let Some(ch) = c {
+                match ch? {
```
HalidOdat, line 141:
```suggestion
        let c = cursor.peek().transpose()?;

        if self.init == '0' {
            if let Some(ch) = c {
                match ch {
```
Here too.

boa/src/syntax/lexer/number.rs
Lan2u: We decided previously (earlier comment) that we will try and keep the number as an integer if possible even if an exponential is used for performance reasons

boa/src/syntax/lexer/number.rs
Lan2u: I wonder if I might just flip the cursor return to be Result<Option<>> - this might improve performance in most cases as transpose won't be needed

boa/src/syntax/lexer/number.rs
HalidOdat: Well this is the Integer case and we already see if it fits in a `i32` here:
```rust
if let Ok(integer) = i32::from_str_radix(&buf, base) {
	// return integer;
```

and if it does not fit we retrun a rational:
```rust
} else {
	let b = f64::from(base);
    let mut result = 0.0_f64;
    for c in buf.chars() {
        let digit = f64::from(c.to_digit(base as u32));
    	result = result * b + digit;
    }
	// return result
}
```

We already try to fit it in an integer (`i32`), so I'm confused about the exponential part you are saying :confused:

Lan2u: from_str_radix doesn't support e/E so if you wrote 10E2 then you could treat it as a rational and get 1000.0 or treat the 10 and the 2 separately and get 1000

boa/src/syntax/lexer/number.rs
HalidOdat: > I wonder if I might just flip the cursor return to be `Result<Option<>>` - this might improve performance in most cases as transpose won't be needed

Yes. this would make the code much more cleaner

Lan2u: > > I wonder if I might just flip the cursor return to be `Result<Option<>>` - this might improve performance in most cases as transpose won't be needed
> 
> Yes. this would make the code much more cleaner

The tradeoff comes from not being able to implement iterator

boa/src/syntax/lexer/number.rs
HalidOdat: > from_str_radix doesn't support e/E so if you wrote 10E2 then you could treat it as a rational and get 1000.0 or treat the 10 and the 2 separately and get 1000

Let me try to explain this a bit better...

we have the `NumericKind` struct which contains the numeric kinds, if a number is for example:
 - `1000` => integer kind (can fit in an `i32`)
 - `0xfff` => integer kind (can fit in an `i32`)
 - `0xffffffffffffffff` => integer kind (can't fit in an `i32`, we store it as f64)
 - `1000000000000000000000000000` => integer kind (can't fit in an `i32`, we store it as f64)
 - `1000.0` => rational kind because it contains a `.`
 - `1000e10` => rational kind because it contains a `e`/`E`

as you can see in the example we can only get an integer kind if we can't get `e` (exponentianl notation) only when the kind is a `NumericKind::Ratioanl` so if we wanted to store as integer in cases where it can fit in an `i32` for example
 - `1000.0`
 - `100e2`
 - `1000.00`

these all can fit in an `i32` we need to in the `Rational` case:
```rust
NumericKind::Rational => {
	let number = f64::from_str(&buf).unwrap();
	if let Ok(integer) = i32::try_from(number) {
		// return integer because it can fit in i32
	}

	// return number because it cant fit in an i32
}
```

boa/src/syntax/lexer/number.rs
HalidOdat: If we don`t use other interator methods, I think it's  best if we implement `.next` returning `io::Result<Option<...>>` this would be much better.

boa/src/syntax/lexer/number.rs
Lan2u: I think I get what you are saying - so this would replace the below code?


                       let n = i32::from_str(&exp_str).map_err(|_| Error::syntax("Could not convert value to f64"))?;

                        if n < 0 { // A negative exponent is expected to produce a decimal value.

                            Numeric::Rational(

                                (num as f64) * f64::powi(10.0, n)

                            )

                        } else if let Some(exp) = i32::checked_pow(10, n as u32) {
                               if let Some(val) = i32::checked_mul(num, exp) {
                                   Numeric::Integer(val)
                               } else {
                                    Numeric::Rational(
                                        (num as f64) * (exp as f64)
                                    )
                               }
                        } else {
                            Numeric::Rational(
                                (num as f64) * f64::powi(10.0, n)
                            )
                        }`

boa/src/syntax/lexer/number.rs
Lan2u: Ok I can go switch stuff around which should clean up the code a fair amount

boa/src/syntax/lexer/number.rs
HalidOdat: First we need to remove `exp_str` and append to `buf` where we appended to `exp_str`

In the `NumericKind::Rational` case:
```rust
NumericKind::Rational => {
	let number = f64::from_str(&buf).unwrap();
	if let Ok(integer) = i32::try_from(number) {
		// return integer because it can fit in i32
	}

	// return number because it cant fit in an i32
}
```
In the `NumericKind::Integer` case:
```rust
NumericKind::Integer(base) => {
	if let Ok(integer) = i32::from_str_radix(&buf, base) {
		// return integer;
		} else {
			let b = f64::from(base);
    		let mut result = 0.0_f64;
   			for c in buf.chars() {
        		let digit = f64::from(c.to_digit(base as u32));
    			result = result * b + digit;
    		}
			// return result
		}
	}
}
```

Lan2u: I didn't find a definition for i32::try_from(num) where num is f64, is there something I'm missing?

HalidOdat: > I didn't find a definition for i32::try_from(num) where num is f64, is there something I'm missing?

Sorry I thought that it existed, in that case we can do something like:
```rust
NumericKind::Rational => {
	let number = f64::from_str(&buf).unwrap();
	if ((number as i32) as f64) == number {
		return number as i32 // because it can fit in i32
	}

	// return number because it cant fit in an i32
}
```
here we check if converting to `i32` does not change the value if it doesn't that means it can fit in a `i32`

Lan2u: Ok this makes sense and the reasoning is because it makes the code cleaner by only requiring buf?

HalidOdat: > Ok this makes sense and the reasoning is because it makes the code cleaner by only requiring buf?

Yes. that is one of the reason but also it is more efficient than having two separate allocations

Lan2u: > > Ok this makes sense and the reasoning is because it makes the code cleaner by only requiring buf?
> 
> Yes. that is one of the reason but also it is more efficient than having two separate allocations

ok cool, I shall make the changes once I've finished switching cursor to Result<Option<>> :)

Lan2u: Done

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,334 @@
...
+                        // Remove the initial '0' from buffer.
+                        cursor.next();
+                        buf.pop();
+
+                        // HexIntegerLiteral
+                        kind = NumericKind::Integer(16);
+                    }
+                    'o' | 'O' => {
+                        // Remove the initial '0' from buffer.
+                        cursor.next();
```
HalidOdat:
```suggestion
                        cursor.next()?;
```
Here too.

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,334 @@
...
+        let mut kind = NumericKind::Integer(10);
+
+        let c = cursor.peek();
+
+        if self.init == '0' {
+            if let Some(ch) = c? {
+                match ch {
+                    'x' | 'X' => {
+                        // Remove the initial '0' from buffer.
+                        cursor.next();
```
HalidOdat: we should handle this, because the `next` method can fail we should propagate it, when there is an io error
```suggestion
                        cursor.next()?;
```

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,334 @@
...
+                    //                 Numeric::Rational(
+                    //                     (num as f64) * (exp as f64)
+                    //                 )
+                    //            }
+                    //     } else {
+                    //         Numeric::Rational(
+                    //             (num as f64) * f64::powi(10.0, n)
+                    //         )
+                    //     }
+                    // }
```
HalidOdat:
```suggestion
```
we don't need these comments, right?

Lan2u: Yes, this + handling the errors is coming in the next commit (just finishing it off now)

boa/benches/lexer.rs
```diff
@@ -17,9 +17,10 @@ static EXPRESSION: &str = r#"
 fn expression_lexer(c: &mut Criterion) {
     c.bench_function("Expression (Lexer)", move |b| {
         b.iter(|| {
-            let mut lexer = Lexer::new(black_box(EXPRESSION));
+            let mut lexer = Lexer::new(black_box(EXPRESSION.as_bytes()));
 
-            lexer.lex()
+            // Goes through and lexes entire given string.
+            while lexer.next().expect("Failed to lex").is_some() {}
```
HalidOdat, line 23: we could have a method in lexer `collect` that would return `Result<Vec<Token>, LexError>`

boa/src/syntax/parser/expression/update.rs
```diff
@@ -37,14 +40,17 @@ impl UpdateExpression {
...
+    R: Read,
+{
     type Output = Node;
 
-    fn parse(self, cursor: &mut Cursor<'_>) -> ParseResult {
-        let tok = cursor.peek(0).ok_or(ParseError::AbruptEnd)?;
-        match tok.kind {
+    fn parse(self, cursor: &mut Cursor<R>) -> ParseResult {
+        let tok = cursor.peek()?.ok_or(ParseError::AbruptEnd);
+        match tok?.kind() {
```
HalidOdat, line 51:
```suggestion
        let tok = cursor.peek()?.ok_or(ParseError::AbruptEnd)?;
        match tok.kind() {
```

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,309 @@
...
+/// Number literal lexing.
+///
+/// Assumes the digit is consumed by the cursor (stored in init).
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-literals-numeric-literals
+/// [mdn]:
```
HalidOdat, line 17: Missing doc link

boa/src/syntax/lexer/operator.rs
```diff
@@ -0,0 +1,156 @@
...
+/// Operator lexing.
+///
+/// Assumes that the cursor has already consumed the operator starting symbol (stored in init).
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]:
+/// [mdn]:
```
HalidOdat, line 81: Missing doc links

boa/src/syntax/lexer/number.rs
Lan2u: Fixed in latest

boa/src/syntax/lexer/operator.rs
```diff
@@ -0,0 +1,159 @@
...
+/// Assumes that the cursor has already consumed the operator starting symbol (stored in init).
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-multiplicative-operators &
+///         https://tc39.es/ecma262/#sec-additive-operators &
+///         https://tc39.es/ecma262/#sec-relational-operators &
+///         https://tc39.es/ecma262/#sec-bitwise-shift-operators
```
HalidOdat, line 83: I don't think this is valid markdown we should use one link

Lan2u: Fixed

Lan2u:
As part of this PR I had to disable the expression_lexer benchmark as it relies on parsing a Divide symbol (which will fail as the lexer on its own cannot infer the required goal symbol), in-order to re-enable an expression benchmark I see a few options:

1. Remove the divide in favour of a multiply or other operation (easiest, keeps goal symbols internal)
2. Make setting the goal symbol on the lexer a public method and then set it before the benchmark runs (means the benchmark remains unchanged so comparisons to old benchmarks is easier)
3. (kinda the non-option) We remove the benchmark entirely

Thoughts?

HalidOdat:
> Thoughts?

I would go with the option 3, lexing and parsing is one operation now because their linked (`goal symbols`).

boa/benches/parser.rs
```diff
@@ -100,11 +85,7 @@ fn long_file_parser(c: &mut Criterion) {
...
             let file_str = fs::read_to_string(FILE_NAME)
                 .unwrap_or_else(|_| panic!("could not read {}", FILE_NAME));
-
-            let mut lexer = Lexer::new(black_box(&file_str));
-            lexer.lex().expect("failed to lex");
-
-            Parser::new(&black_box(lexer.tokens)).parse_all()
+            Parser::new(black_box(file_str.as_bytes())).parse_all()
         })
     });
```
HalidOdat, line 90: we don't have to `read_to_string` we can just pass the `File` because it implements `Read` trait:
```rust
   let file = std::fs::File::open(FILE_NAME).expect("Could not open file");
    c.bench_function("Long file (Parser)", move |b| {
		b.iter(|| {
            Parser::new(black_box(file)).parse_all()
        })
    });
```

Lan2u: Done

jasonwilliams:
Agree option 3

Razican:
> Thoughts?

Yes, that was the reason behind us adding the lexing part to parsing benchmarks.

Sorry I didn't have time these last weeks to review this. Hopefully I will get some time these days. This is an important step!

boa/src/syntax/lexer/cursor.rs
```diff
@@ -0,0 +1,243 @@
...
+                // next_is_pred will return false if the next value is None so the None case should already be handled.
+                unreachable!();
+            }
+        }
+    }
+
+    /// It will fill the buffer with checked ASCII bytes.
+    pub(super) fn fill_bytes(&mut self, buf: &mut [u8]) -> io::Result<()> {
+        unimplemented!("Lexer::cursor::fill_bytes {:?}", buf)
+    }
```
Razican, line 137: Do we need this?

boa/src/syntax/lexer/comment.rs
```diff
@@ -0,0 +1,66 @@
...
+            if ch == '\n' {
+                break;
+            }
+        }
+        Ok(Token::new(
+            TokenKind::Comment,
+            Span::new(start_pos, cursor.pos()),
+        ))
+    }
+}
```
Razican, line 34: Checking at the spec:
> Because a single-line comment can contain any Unicode code point except a LineTerminator code point, and because of the general rule that a token is always as long as possible, a single-line comment always consists of all code points from the // marker to the end of the line. However, the LineTerminator at the end of the line is not considered to be part of the single-line comment; it is recognized separately by the lexical grammar and becomes part of the stream of input elements for the syntactic grammar. This point is very important, because it implies that the presence or absence of single-line comments does not affect the process of automatic semicolon insertion (see 11.9).

I don't think we should emit a `Comment` token kind, but I don't think we should emit a `LineTerminator` either, since it's not part of the comment.

The only place where this is relevant is if we have a JS file where its last line is a single line comment, and doesn't have a `\n` in the end. We should just discard the comment, and there should be no `LineTerminator` being produced (since this would affect automatic semicolon insertion).

So, to make things simpler, we can make this ignore the last `\n` by only peeking it, and leaving the lexer produce a `LineTerminator` after this one, and just return the comment.
The issue with the current implementation is that it will remove the line terminator from the lexer list of tokens.

This:
```javascript
let a = 10 // Comment
let b = 20;
```

Will be equivalent to this:
```javascript
let a = 10 let b = 20;
```
But it should be equivalent to this:
```javascript
let a = 10
let b = 20;
```

Note that the correct representation will correctly create two variables, since the line break between the `a` and `b` declaration will create a semicolon automatically. In the current implementation, this will just fail, since we cannot have a `let` in another `let` declaration.

I would add this as a test case.

boa/src/syntax/lexer/comment.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Lexes a single line comment.
+///
+/// Assumes that the initial '//' is already consumed.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-comments
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar
```
Razican, line 17: I would put the comment in the structure instead of in the implementation, so that it shows nicely when generating the documentation.

boa/src/syntax/lexer/comment.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Assumes that the initial '/*' is already consumed.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-comments
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar
+pub(super) struct BlockComment;
+impl<R> Tokenizer<R> for BlockComment {
```
Razican, line 47:
```suggestion

impl<R> Tokenizer<R> for BlockComment {
```

boa/src/syntax/lexer/comment.rs
```diff
@@ -0,0 +1,66 @@
...
+///
+/// Assumes that the initial '/*' is already consumed.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-comments
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar
+pub(super) struct BlockComment;
```
Razican, line 46: I would change the name to `MultiLineComment`, as in the spec.

boa/src/syntax/lexer/comment.rs
```diff
@@ -0,0 +1,66 @@
...
+            } else {
+                return Err(Error::syntax("unterminated multiline comment"));
+            }
+        }
+        Ok(Token::new(
+            TokenKind::Comment,
+            Span::new(start_pos, cursor.pos()),
+        ))
+    }
+}
```
Razican, line 66: As per the spec:
> Comments behave like white space and are discarded except that, if a MultiLineComment contains a line terminator code point, then the entire comment is considered to be a LineTerminator for purposes of parsing by the syntactic grammar.

So, most of the time, this just works, but there is a particular case where this is not spec compliant:

```javascript
let a = 10 /* Test
Multiline
Comment
*/ let b = 10;
```

This should be equivalent to:
```javascript
let a = 10
let b = 10;
```

But once again, it happens the same as in single line comments, where the new line doesn't get inserted, and therefore we don't have an automatic semicolon insertion.

Note that this new line should only be inserted when the comment has one new line character in it `\n`, so this should not insert a new line:
```javascript
let a = 10; /* Test comment */ let b = 10;
```

So, basically, we need to check if we find a `\n` character inside the multiline comment. If we do, then the returned token should be a `LineTerminator`, if not, then we return a `Comment`, that will later be discarded.

Once again, I would add this as a test case.

Razican:
About benchmarks, It seems that the parser is now 2x-3x slower. Except, as expected, for long files, where we get a huge speed bump (from 6.1±0.25ms to 742.7±54.50ns), which is a **87.8%** speed improvement.

I will review the parser, especially the parser cursor, to see if further improvements can be made.

boa/src/syntax/lexer/cursor.rs
Lan2u: I don't think so, its from the string_literal code and I assumed you were using it for a specific reason so I left it

Razican: Are we still using it in string literals, or did you find a better way to do it?

boa/src/syntax/lexer/comment.rs
Lan2u: > About benchmarks, It seems that the parser is now **2x-3x** slower. Except, as expected, for long files, where we get a huge speed bump (from 6.1±0.25ms to 742.7±54.50ns), which is an **87.8%** speed improvement.
> 
> I will review the parser, especially the parser cursor, to see if further improvements can be made.

It would be interesting to see a benchmark with a lot of goal symbol switches vs one without

Lan2u:
> About benchmarks, It seems that the parser is now **2x-3x** slower. Except, as expected, for long files, where we get a huge speed bump (from 6.1±0.25ms to 742.7±54.50ns), which is an **87.8%** speed improvement.
> 
> I will review the parser, especially the parser cursor, to see if further improvements can be made.

> > About benchmarks, It seems that the parser is now **2x-3x** slower. Except, as expected, for long files, where we get a huge speed bump (from 6.1±0.25ms to 742.7±54.50ns), which is an **87.8%** speed improvement.
> > I will review the parser, especially the parser cursor, to see if further improvements can be made.
> 
> It would be interesting to see a benchmark with a lot of goal symbol switches vs one without

Lan2u:
> About benchmarks, It seems that the parser is now **2x-3x** slower. Except, as expected, for long files, where we get a huge speed bump (from 6.1±0.25ms to 742.7±54.50ns), which is an **87.8%** speed improvement.
> 
> I will review the parser, especially the parser cursor, to see if further improvements can be made.

I suspect that mechanism used to allow peeking 2 tokens ahead might be a good place to start (it might be faster to replace the VecDeque with just an array of a certain size)

boa/src/syntax/lexer/cursor.rs
Lan2u: > Are we still using it in string literals, or did you find a better way to do it?

I didn't really change the string literal stuff so its still there but equally all the tests pass so it can't be doing anything major at the moment - will need to look back into the source code and see

boa/src/syntax/lexer/tests.rs
```diff
@@ -264,270 +151,479 @@ fn check_keywords() {
...
+    );
 }
 
 #[test]
-#[ignore]
-fn two_divisions_in_expression() {
-    let s = "    return a !== 0 || 1 / a === 1 / b;";
-    let mut lexer = Lexer::new(s);
-    lexer.lex().expect("failed to lex");
-    // dbg!(&lexer.tokens);
```
Razican, line 356: I see that this was removed, but this was the main reason behind all this. This test is directly taken from test262. We should have it somewhere.

Razican:
> It would be interesting to see a benchmark with a lot of goal symbol switches vs one without

Do you have an example code that would cause this?

> I suspect that mechanism used to allow peeking 2 tokens ahead might be a good place to start (it might be faster to replace the VecDeque with just an array of a certain size)

Using an array in the stack would be very helpful, I think. What do we use the `VecDeque` for?

boa/src/syntax/lexer/tests.rs
Lan2u: It is, I moved it (and also split it into multiple tests so I could narrow down exactly the behaviour) - I'll go find them

Lan2u:
> > It would be interesting to see a benchmark with a lot of goal symbol switches vs one without
> 
> Do you have an example code that would cause this?
> 
> > I suspect that mechanism used to allow peeking 2 tokens ahead might be a good place to start (it might be faster to replace the VecDeque with just an array of a certain size)
> 
> Using an array in the stack would be very helpful, I think. What do we use the `VecDeque` for?

VecDeque is basically a queue implementation which is used to store peeked elements - it is also used in one case for the 'push_back' method which pushes an element back onto the peek queue. 

We only ever need to skip a single element when peeking ahead but because of the possibility of push_back() we need the buffer peeked array to be of at least size 3 (peek + peek_skip + push_back).

boa/src/syntax/lexer/tests.rs
Lan2u: Moved to line 100 - 131 in boa/src/syntax/parser/tests.rs (since they involve the parser I thought they were better placed there).

boa/src/syntax/lexer/comment.rs
Lan2u: Test case added + modified lexer so single line comments no longer consume the last line terminator.

Lan2u:
> > It would be interesting to see a benchmark with a lot of goal symbol switches vs one without
> 
> Do you have an example code that would cause this?
> 
> > I suspect that mechanism used to allow peeking 2 tokens ahead might be a good place to start (it might be faster to replace the VecDeque with just an array of a certain size)
> 
> Using an array in the stack would be very helpful, I think. What do we use the `VecDeque` for?

`
function foo(regex, num) {}

let i = 0;
while (i < 1000000) {
	foo(/ab+c/, 5.0/5);
  i++;
}

`

boa/src/syntax/lexer/comment.rs
Lan2u: Done

boa/src/syntax/lexer/comment.rs
Lan2u: Done

boa/src/syntax/lexer/comment.rs
Lan2u: Done

boa/src/syntax/lexer/cursor.rs
Lan2u: Actually I think we might need it - the only usage seems to require the raw u8 bytes rather than a rust char (and so therefore the regular .next() method won't work).

boa/src/syntax/lexer/comment.rs
Lan2u: Done (added 2 test cases in parser/tests)

boa/src/syntax/lexer/comment.rs
```diff
@@ -53,18 +53,30 @@ impl<R> Tokenizer<R> for MultiLineComment {
...
                 return Err(Error::syntax("unterminated multiline comment"));
             }
         }
-        Ok(Token::new(
-            TokenKind::Comment,
-            Span::new(start_pos, cursor.pos()),
-        ))
+
+        if ret.is_some() {
+            return Ok(ret.unwrap());
```
Razican, line 74:
```suggestion
        if let Some(ret) = ret {
            Ok(ret)
```

Razican:
> ```js
> function foo(regex, num) {}
> 
> let i = 0;
> while (i < 1000000) {
> 	foo(/ab+c/, 5.0/5);
>   i++;
> }
> ```

I will add this as a parsing benchmark, so that we can see differences.

Razican:
@Lan2u could you rebase the branch to get the new benchmarks and compare? I think we will get nice insights there :)

Lan2u:
> @Lan2u could you rebase the branch to get the new benchmarks and compare? I think we will get nice insights there :)

Done (I think)

Razican:
Should we close this in favour of #559?

Issue #463, opened 2020-06-07
[executor] Use of undefined variable should throw an error


croraf:
**Describe the bug**
```js
console.log(d);
console.log(d());
d;
d();
```

Any of the above statements should throw an error `ReferenceError: d is not defined`

Razican:
For anyone willing to work on this, this has two sides:
 implemented in [here][ident_exec] for identifiers execution. This will go to the Realm environment record and call [get_binding_value()][env_binding_val]. This will either return the value, or `Undefined`.

This is not 100% spec compliant. The [spec][ident_spec] says that this should call the "abstract" operation [`ResolveBinding`][resolve_binding], with no environment pre-set. Then, this should call to [`GetIdentifierReference`][get_ident_ref] (Note that Boa doesn't implement strict mode yet). We currently don't have `Reference` implemented, either, so, the current implementation does a fine job returning an `undefined`, but we lack information on the binding name. The main issue here, is that we should be calling `GetValue` on this returned reference, if I'm not mistaken.

As we can see in the formal [specification][get_value_spec] of this function, we should throw a `ReferenceError` if the value is not defined. Ideally, we should implement `Reference` and its `get_value()` function that would be 100% spec compliant.

Note that this type of error is not implemented in Boa yet, and we only have the general `Error`, the `TypeError` and the `RangeError`, as you can see [here][errors].

To implement a new one, you can base your implementation in the [`RangeError` implementation][range_error], for example, and then add a helper to the interpreter in the [exceptions list][exceptions] to easily throw one if needed.

A bit of extra information about `ReferenceError` can be found in the [spec][reference_error_spec].

About the calling of an undefined variable, is implemented [here][call]. This is calling `interpreter.call`, which is implemented [here][interp_call] with its spec [here][call_spec]. This is not spec compliant, but we can improve it greatly, and even make it spec compliant. It should first not check `func` to se if it's callable, since we recently implemented `is_callable()` on values. This should throw a `TypeError`, not panic, in the case that the value is not callable.

But we are missing here the `ReferenceError`. This is coming from [here][call_expr_spec], from the second part, that does `CallExpression: CallExpression Arguments`. Here, when we evaluate the call expression, we get the undefined `d` Reference, and should be calling `get_value()` on it, which should throw the `ReferenceError`.

I think we should first fix the `d;` case by implementing `Reference` and `get_value()`, and then, fix the `Call` implementation.

[...]

PR #488, opened 2020-06-12
[ReferenceError] complete solution

croraf:
#463


croraf:
I think basically this might be it (after all checks pass). I planned to do some tests. But I don't know how to assert the error output from the script:

- I tried `try/catch` but seems not to be working. 

- `instanceof` is also not implemented to test if ReferenceError is thrown. 

- for any error thrown we currently print the error object that is thrown, instead of error message

Razican:
> I tried `try/catch` but seems not to be working.

`try/catch` should be working. If it's not, then it's a bug. I think that the part of `try/catch` returning something is the part that is not 100% spec compliant, but exceptions can be thrown and catched, and finally blocks ran.

> for any error thrown we currently print the error object that is thrown, instead of error message

This might be something to look at, since depending on which function you use to execute the code, this will happen. If we use `exec()` or `forward()`, then yes, we return the string representation, but if we use `forward_val()`, then we return a `Result<Value, Value>`, and here, we could check if it's an error, and if it is, we can get more information, such as it being an Object and probably its instance information.

> Also several tests currently fail, because they were written to expect wrong thing, for example:
> 
> ```rust
>         scenario = r#"
>           {
>             let bar = "bar";
>           }
>           bar == undefined;
>         "#;
> 
>         assert_eq!(&exec(scenario), "true");
> ```
> 
> This is incorrect assertion. `bar` is not undefined but a ReferenceError should is thrown.

For these cases, just fix the test or remove it, since it's not spec compliant.

croraf:
Try/Catch is working. The issue was in something else. I thus managed to unignore 4 test cases.

PR #490, opened 2020-06-13
[instanceof] implementation

croraf:

It changes the following:
 - 
 - 
 -

### DIFF
--- a/boa/src/exec/operator/mod.rs
+++ b/boa/src/exec/operator/mod.rs
@@ -97,6 +97,17 @@ impl Executable for BinOp {
                         let key = interpreter.to_property_key(&mut v_a)?;
                         interpreter.has_property(&mut v_b, &key)
                     }
+                    CompOp::InstanceOf => {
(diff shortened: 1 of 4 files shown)


boa/src/exec/operator/mod.rs
```diff
@@ -97,6 +97,17 @@
...
+                    CompOp::InstanceOf => {
+                        if !v_b.is_object() {
+                            return interpreter.throw_type_error(format!(
+                                "right-hand side of 'instanceof' should be an object, got {}",
+                                v_b.get_type().as_str()
+                            ));
+                        }
+                        return interpreter.throw_type_error(format!(
+                            "'instanceof' evaluation not yet implemented"
+                        ));
```
jasonwilliams, line 109: We don't support well-known symbols yet, so you can skip the `@@hasInstance` logic and just do the https://tc39.es/ecma262/#sec-ordinaryhasinstance algorithm

Razican:
This has been superseeded by #559.

Issue #464, opened 2020-06-08
Possibly use something like rustyline for the cli


IovoslavIovchev:
[rustyline](https://crates.io/crates/rustyline/) offers some niceties out of the box (history, vi mode, etc.), which would improve the experience when using the boa-cli.

Note that this issue is not about adding `rustyline` specifically but rather about adding features that crates like `rustyline` (and possibly others) already offer and would be nice to have.

IovoslavIovchev:
I can take a look into adding `rustyline` if there are no other suggestions.

HalidOdat:
> I can take a look into adding `rustyline` if there are no other suggestions.

Awesome! as an example use of `rustyline` you can check https://github.com/PaddiM8/kalk

HalidOdat:
Cloasing this issue.

PR #492, opened 2020-06-14
rustyline for the cli

IovoslavIovchev:
This PR addresses #464


boa_cli/src/main.rs
```diff
@@ -196,26 +198,47 @@ pub fn main() -> Result<(), std::io::Error> {
...
+                Err(ReadlineError::Interrupted) | Err(ReadlineError::Eof) => break,
+
+                Ok(line) => {
+                    editor.add_history_entry(&line);
+
+                    if args.has_dump_flag() {
+                        match dump(&line, &args) {
+                            Ok(_) => {}
+                            Err(e) => eprintln!("{}", e),
+                        }
```
Razican, line 225:
```suggestion
                        if let Err(e) = dump(&line, &args)_{
                            eprintln!("{}", e);
                        }
```

IovoslavIovchev:
> This looks very good. I have a couple of questions though. Should we use a history file for the console? And, should we print errors in red, for instance?

If you are wondering whether we should have a history feature for the REPL in the first place, maybe for now it could be an optional feature with conditional compilation?
As for the second question, I think it leads to a broader question - how much UX do we want for the REPL? There are definitely quite a few things that could be added. Maybe we should open a master issue with a checklist or something.

Issue #413, opened 2020-05-24
Default number formatting is not spec compliant


HalidOdat:
**Describe the bug**

Currently we use the default native Rust formatter to format floating-point numbers, which is not spec compliant.

**To Reproduce**
Steps to reproduce the issue, or JavaScript code that causes this failure.

This JavaScript code reproduces the issue:
```javascript
0.0000001
3.e50
```

**Expected behavior**

In this scenario we should get `1e-7` and `3e+50` back.
Actual output is `0.0000001` and `300000000000000000000000000000000000000000000000000`. 

**Build environment (please complete the following information):**
 - OS: Manjaro Linux
 - Target triple: x86_64-unknown-linux-gnu
 - Rustc version: 1.43.1 (8d69840ab 2020-05-04)

**Additional context**

We need to write our own f64 formatter like V8 `DoubleToCString` (see V8 implementation link bellow)

You can find more information:
 - [ECMAScript spec](https://tc39.es/ecma262/#sec-numeric-types-number-tostring)
 - [V8 implementation](https://chromium.googlesource.com/v8/v8/+/refs/heads/master/src/numbers/conversions.cc#938)
 - [V8 Tests](https://chromium.googlesource.com/v8/v8/+/4.3.18/test/mjsunit/number-tostring.js?autodive=0%2F%2F%2F%2F#30)

Tropid:
I can take this if nobody else wants it :)

HalidOdat:
> I can take this if nobody else wants it :)

Sure! Tell us if you need any help, or if you have any questions

Issue #449, opened 2020-06-02
JSON.stringify sometimes changes values from int to float


tomasol:
When using functions like `Math.round` or `Math.floor`, type of the attribute seems seems to be float. This is not observable when inspecting the object, but is visible using `JSON.stringify`.
**To Reproduce**
Start CLI, enter
```javascript
JSON.stringify({foo:Math.round(1)})
```
Response: `{"foo":1.0}`. 
**Expected behavior**
Expected response `{"foo":1}`.

**Build environment (please complete the following information):**
- boa version: master (87aea64c)
 - OS: Ubuntu 
 - Version: 20.04
 - Target triple: x86_64-unknown-linux-gnu
 - Rustc version:  1.43.1 (8d69840ab 2020-05-04)

HalidOdat:
This might be related to #413

croraf:
A simpler (seems to be related) example is:
```
JSON.stringify({foo: 1.0})
```
which outputs `{"foo":1.0}` instead of `{"foo":1}`

PR #493, opened 2020-06-14
Use correct exponential representation for rational values

Tropid:
This Pull Request addresses #413

It changes the following:
- Uses Rust exponential display for numbers which should be represented in scientific notation
- Adds a few test cases

This is a pretty simple way to deal with the problem at hand. It just uses Rusts exponential formatting for values that require the scientific notation. It has a few problems, though:

- I'm not sure if the boundary conditions work correctly in every circumstance (probably not)
- It has to parse the string to replace the default output of the rust exponential formatting (`.replace("e", "+e")`)

I looked at various documents regarding converting doubles to decimal representation. There is V8s implementation (as linked in the Issue) which seems to use different algorithms based on the number (see `DoubleToAscii()` here: https://chromium.googlesource.com/v8/v8/+/refs/heads/master/src/numbers/dtoa.cc). The other algorithm is called `BignumDtoa` in V8.
Rust itself uses a combination of the Grisu and the Dragon families of algorithms (see https://github.com/rust-lang/rust/blob/master/src/libcore/num/flt2dec/mod.rs#L49).

Also there seems to be a new algorithm which may yield better results:
- https://pldi18.sigplan.org/details/pldi-2018-papers/20/Ry-Fast-Float-to-String-Conversion
- https://github.com/rust-lang/rust/issues/52811

I see three options to continue this issue:

1) Use this and maybe replace it later on to squeeze out more performance

This is a rather simple/easy to understand solution. It has a rather advanced implementation (Rust std/core library). It needs not much maintenance as it uses the std library. It's probably good enough for the first time/most use cases.

Also we can later replace it with more advanced algorithms.

2) Implement `BignumDtoa` from V8. This seems to be the general purpose function to format any double (V8 tries faster methods before falling back to this one).

3) (Re-)Implement the full amount of different algorithms

This seems rather complicated (at least to me) and difficult to maintain/tweak the performance to reach at least the same as the std implementation.

4) ?

I'd be glad to hear some more input on this topic.


HalidOdat:
Hmmm. I'm fine with having this for the time being, what we had before was not spec compliant at all. And in the future we can improve it.

Tropid:
This Ryu thing looks really interesting, though.

It looks like the original author of Grisu3 (which seems to be used for fast conversion in V8) even said Ryu is faster than Grisu3:
https://www.hackernext.com/story?id=20181832

I'd really be interested in implementing this algorithm.

croraf:
Cool. I think we do need this asap. The stringification of numbers must be compliant with https://tc39.es/ecma262/#sec-numeric-types-number-tostring. This will likely also solve https://github.com/boa-dev/boa/issues/449.

What is the current status you are in @Tropid ?

Tropid:
@croraf sorry, I forgot to reply after we talked about it in the Discord chat.

I just updated the pull request. It now uses the ryu algorithm to print the numbers.

I forked the dtolnay/ryu project/crate and adjusted it for ECMAScript formatting. This *draft* PR adds a dependency to my fork. I'd suggest for Boa to fork this project itself and cherry-pick my changes or something like that, so that the repository is accessible by Boa and can be updated regularly.

@HalidOdat What are your opinions on that?

HalidOdat:
> @HalidOdat What are your opinions on that?

Good job! @Tropid.

I agree, we may also want to publish this to crates.io, since `ryu-ecmascript` looks to be unmaintained, has not been updated in 2 years.

Razican:
> What do you think @Razican @jasonwilliams

Let's do this :)

Razican:
The error is giving is due to a missing `cargo update`.

PR #495, opened 2020-06-15
Added `globalThis` property

### LINKED ISSUE #120: Implement global `globalThis` property
**ECMASCript feature**

We can add this by adding a new module like `NaN` https://github.com/boa-dev/boa/tree/master/boa/src/builtins/nan and then adding it to the builtin `init` function in `builtins/mod.rs` the value it should contain should be `global.clone()` (this is the value passed to init function)

More information
    - [MDN][mdn]

[spec]: https://tc39.es/ecma262/#sec-globalthis
[mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

zanayr:
This Pull Request fixes/closes #120.

It changes the following:
 - adds the globalThis property

I'm not really sure if I did this right, I followed the suggestions left on #120.

> We can add this by adding a new module like NaN https://github.com/boa-dev/boa/tree/master/boa/src/builtins/nan and then adding it to the builtin init function in builtins/mod.rs the value it should contain should be global.clone() (this is the value passed to init function)

This is my first time, be gentle.. Please let me know if there are any problems, i'd love to fix/learn from them.

Best,


boa/src/builtins/mod.rs
```diff
@@ -6,6 +6,7 @@ pub mod boolean;
 pub mod console;
 pub mod error;
 pub mod function;
+pub mod globalThis;
```
HalidOdat, line 9: The rust language use `snake_case` for module names, same for variables, and `RascalCase` for type definitions. for more information on this you can see: <https://rust-lang.github.io/api-guidelines/naming.html>
The language is very strict about naming conventions so it does not compile the code.

So this we need to make this snake case so this would be `global_this`, instead of `globalThis`
Note the file name will have to change too.

and in the `init` function in `builtins/mod.rs` we call `global_this::init()`, instead of `globalThis::init()`

boa/src/builtins/globalThis/mod.rs
```diff
@@ -0,0 +1,11 @@
+#[cfg(test)]
+mod tests;
+
+use crate::{builtins::value::Value, BoaProfiler};
+
+/// Initialize the `globalThis` property on the global object.
+#[inline]
+pub fn init(global: &Value) {
+    let _timer = BoaProfiler::global().start_event("globalThis", "init");
+    global.set_field("globalThis", Value::from(global.clone()));
```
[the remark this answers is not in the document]
zanayr: Cool! Thanks, I'll get this sorted right away!

boa/src/builtins/mod.rs
zanayr: ... I'm embarrassed.. I knew this, but I think I got too excited. My apologies, I'm fixing it now.

zanayr:
I guess I don't know how to write the test for this module.
I understand what the test is doing, it's executing the scenario as JS, but i'm not sure what to check it against? Window? if I did `r#"typeof globalThis;"#` and compared it to `"object"` would that work? Sorry for all the questions... 😅

boa/src/builtins/global_this/tests.rs
```diff
@@ -0,0 +1,10 @@
+use crate::exec;
+
+#[test]
+fn global_this_exists_on_global_object_and_evaluates_to_global_this_value() {
+    let scenario = r#"
+        globalThis;
```
HalidOdat, line 6: The test is failing because the `globalThis` is the global object, it's not the string value `"globalThis"`, when we call `global_this::init` we set the field "globalThis" (this is the binding name not the value) and the value is the global object (that holds objects like `Number`, `String`, etc), a better way of checking if it exists would be to check `globalThis !== undefined` this evaluates to `"true"`

HalidOdat:
> I did `r#"typeof globalThis;"#` and compared it to `"object"` would that work?

Yes. that would work that is probably a better option to what I gave.

boa/src/builtins/global_this/tests.rs
zanayr: Ah, ok! I understand! Thank you for your patience, @HalidOdat! :)

HalidOdat:
The failed benchmarks are not related to this PR, it has to do with `criterion-compare-action`

PR #498, opened 2020-06-16
`Value` refactor

### LINKED ISSUE #465: Refactoring Value (decouple `Gc` from `Value`)
**Whats wrong with the current implementation?**
The current implementation after #383 coupled the `Gc` and `Value` together and the operations were promoted to the gc level, the problem with this is that `Gc` values are stored in the heap which means a performance penalty. Furthermore we need to do some checks before accessing the underlying `ValueData`.

**Changes**
 - Move `Value` logic/operations **=>** `ValueData` 
 - Rename `ValueData` **=>** `Value` (it makes more sense to call it `Value`)
 - Move `Gc<...>` to `Value::Object` field (`Value::Object(Gc<GcCell<Object>>)`)

If you have any suggestions on how to improve it, or even how other engines to it, please comment :)

I would like to start working on this preferably after #419 because the rebase is going to be a nightmare

HalidOdat:
This Pull Request fixes/closes #465 

It changes the following:
 - Moved `Gc<...>` **=>** `Value::Object(Gc<GcCell<Object>>)`
 - Renamed `ValueData` **=>** `Value`
 - Added reference counted `RcString`, `RcSymbol`, `RcBigInt`
 - Refactor `Value::String(String)` **=>** `Value::String(RcString)`
 - Refactor `Value::Symbol(Symbol)` **=>** `Value::Symbol(RcSymbol)`
 - Refactor `Value::BigInt(BigInt)` **=>** `Value::BigInt(RcBigInt)`
 - Abstracted `Gc<GcCell<Object>>` **=>** `GcObject`
 - Dropped `Value` size from `40` bytes **=>** `24` bytes
 - Changed function signature `this`, from `&mut Value` **=>** `&Value`
 - Removed `Interpreter::value_to_rust_number()`
 - Removed unnecessary `Box`s in global environment
 - Extracted `extensible` for `internal_slots`
 - Made `to_primitive`, `ordinary_to_primitive` and `Value::equals` throw errors.
 - Removed `strict` parameter of `SameValue` (it is not dependent on strict mode)

### DIFF
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -16,7 +16,7 @@ use crate::{
         array::Array,
         object::{Object, ObjectData, INSTANCE_PROTOTYPE, PROTOTYPE},
         property::Property,
-        value::{ResultValue, Value},
+        value::{RcString, ResultValue, Value},
     },
     environment::function_environment_record::BindingStatus,
     environment::lexical_environment::{new_function_environment, Environment},
@@ -28,7 +28,7 @@ use gc::{unsafe_empty_trace, Finalize, Trace};
 use std::fmt::{self, Debug};
 
 /// _fn(this, arguments, ctx) -> ResultValue_ - The signature of a built-in function
-pub type NativeFunctionData = fn(&mut Value, &[Value], &mut Interpreter) -> ResultValue;
+pub type NativeFunctionData = fn(&Value, &[Value], &mut Interpreter) -> ResultValue;
 
 /// Sets the ConstructorKind
 #[derive(Debug, Copy, Clone)]
@@ -178,7 +178,7 @@ impl Function {
     pub fn call(
         &self,
         function: Value, // represents a pointer to this function object wrapped in a GC (not a `this` JS object)
-        this: &mut Value,
+        this: &Value,
         args_list: &[Value],
         interpreter: &mut Interpreter,
     ) -> ResultValue {
@@ -236,7 +236,7 @@ impl Function {
     pub fn construct(
         &self,
         function: Value, // represents a pointer to this function object wrapped in a GC (not a `this` JS object)
-        this: &mut Value,
+        this: &Value,
         args_list: &[Value],
         interpreter: &mut Interpreter,
     ) -> ResultValue {
@@ -376,7 +376,8 @@ pub fn create_unmapped_arguments_object(arguments_list: &[Value]) -> Value {
             .writable(true)
             .configurable(true);
 
-        obj.properties_mut().insert(index.to_string(), prop);
+        obj.properties_mut()
+            .insert(RcString::from(index.to_string()), prop);
         index += 1;
     }
 
@@ -386,7 +387,7 @@ pub fn create_unmapped_arguments_object(arguments_list: &[Value]) -> Value {
 /// Create new function `[[Construct]]`
 ///
 // This gets called when a new Function() is created.
-pub fn make_function(this: &mut Value, _: &[Value], _: &mut Interpreter) -> ResultValue {
+pub fn make_function(this: &Value, _: &[Value], _: &mut Interpreter) -> ResultValue {
     this.set_data(ObjectData::Function(Function::builtin(
         Vec::new(),
         |_, _, _| Ok(Value::undefined()),
--- a/boa/src/builtins/object/gcobject.rs
+++ b/boa/src/builtins/object/gcobject.rs
@@ -1,0 +1,73 @@
+//! This module implements the `GcObject` structure.
+//!
+//! The `GcObject` is a garbage collected Object.
+
+use super::Object;
+use gc::{Finalize, Gc, GcCell, GcCellRef, GcCellRefMut, Trace};
+use std::fmt::{self, Display};
+
+/// Garbage collected `Object`.
+#[derive(Debug, Trace, Finalize, Clone)]
+pub struct GcObject(Gc<GcCell<Object>>);
+
+impl GcObject {
+    #[inline]
+    pub(crate) fn new(object: Object) -> Self {
+        Self(Gc::new(GcCell::new(object)))
+    }
+
+    #[inline]
+    pub fn borrow(&self) -> GcCellRef<'_, Object> {
+        self.try_borrow().expect("Object already mutably borrowed")
+    }
+
+    #[inline]
+    pub fn borrow_mut(&self) -> GcCellRefMut<'_, Object> {
+        self.try_borrow_mut().expect("Object already borrowed")
+    }
+
+    #[inline]
+    pub fn try_borrow(&self) -> Result<GcCellRef<'_, Object>, BorrowError> {
+        self.0.try_borrow().map_err(|_| BorrowError)
+    }
+
+    #[inline]
+    pub fn try_borrow_mut(&self) -> Result<GcCellRefMut<'_, Object>, BorrowMutError> {
+        self.0.try_borrow_mut().map_err(|_| BorrowMutError)
+    }
+
+    /// Checks if the garbage collected memory is the same.
+    #[inline]
+    pub fn equals(lhs: &Self, rhs: &Self) -> bool {
+        std::ptr::eq(lhs.as_ref(), rhs.as_ref())
+    }
+}
+
+impl AsRef<GcCell<Object>> for GcObject {
+    #[inline]
+    fn as_ref(&self) -> &GcCell<Object> {
+        &*self.0
+    }
+}
+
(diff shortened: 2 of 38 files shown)


HalidOdat:
Woww. The benchmarks are awesome. I did not expect that big of an improvement from an initial refactor of `Value`. There is still a lot to do from #465.

HalidOdat:
Apparently `Trace` is not implemented for `Rc`, made a PR https://github.com/Manishearth/rust-gc/pull/106

HalidOdat:
Size of `Value` (there is only `Value` now, no `ValueData`) has dropped from `40` bytes to 24 bytes

HalidOdat:
This is ready for review :) @Razican @jasonwilliams

HalidOdat:
I did a profile of `master` and `refactor/value` branches, here is the code that was executed: <https://gist.github.com/HalidOdat/a11478d0d67a7150a6faf69c2126d717>

The profile for `master` is here: <https://gist.github.com/HalidOdat/795d1af25d706a4fad5bd381ac2dfb8c>

master
---------
And for `refactor/value` here: <https://gist.github.com/HalidOdat/b5c7372d5d43912110fb6c46d84c3ea6>
refactor_value

There is a `13ms` difference between them.

BTW: The gap seems to disappear if I run it a lot with very little time in between, which makes me think that the the file is being cached, that's why sometimes it appears, and other times not.

boa/src/builtins/regexp/mod.rs
```diff
@@ -103,8 +97,8 @@ impl RegExp {
         match args.get(1) {
             None => {}
             Some(flags) => {
-                if let ValueData::String(flags) = flags.deref() {
-                    regex_flags = flags.into();
+                if let Value::String(flags) = flags {
+                    regex_flags = flags.to_string();
                 }
             }
         }
```
Razican, line 104: I would use an `if let Some(flags) = args.get(1) {` here.

boa/src/builtins/regexp/mod.rs
```diff
@@ -68,32 +66,28 @@ impl RegExp {
...
     /// Create a new `RegExp`
-    pub(crate) fn make_regexp(
-        this: &mut Value,
-        args: &[Value],
-        ctx: &mut Interpreter,
-    ) -> ResultValue {
+    pub(crate) fn make_regexp(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
         if args.is_empty() {
             return Err(Value::undefined());
         }
```
Razican, line 72: I think we should use this opportunity to get the value. Something like:
```rust
let arg = match args.get(0).ok_or_else(Value::undefined)?;
```

And then, down below, we could use:
```rust
match *arg {
```

This would avoid that indexing, and the clippy lint exception.

boa/src/builtins/string/mod.rs
```diff
@@ -647,10 +630,9 @@ impl String {
     /// [spec]: https://tc39.es/ecma262/#sec-string.prototype.match
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match
     /// [regex]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
-    pub(crate) fn r#match(this: &mut Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
-        let mut re =
-            RegExp::make_regexp(&mut Value::from(Object::default()), &[args[0].clone()], ctx)?;
-        RegExp::r#match(&mut re, ctx.to_string(this)?, ctx)
+    pub(crate) fn r#match(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
+        let re = RegExp::make_regexp(&Value::from(Object::default()), &[args[0].clone()], ctx)?;
```
Razican, line 634: Am I wrong to think that this will panic if no arguments are passed to this function?

boa/src/builtins/string/mod.rs
```diff
@@ -722,7 +704,12 @@ impl String {
             None
```
Razican, line 704: This part here is not convincing me. I know it's not part of this request, but I think we could refactor this to make it more clear. Something like:
```rust
let fill_string = args.get(1).map(|arg| ctx.to_string(arg)).transpose()?;
```

boa/src/builtins/string/mod.rs
```diff
@@ -722,7 +704,12 @@ impl String {
...
         };
 
-        Self::string_pad(primitive_val, max_length, fill_string, false)
+        Self::string_pad(
+            primitive_val.to_string(),
+            max_length,
+            fill_string.map(|x| x.to_string()),
```
Razican, line 710: I think we can just pass the `to_string()` function to the `map()` here, right? Instead of creating the closure. Passing `Rc::to_string` (if I'm not mistaken on what this is xD)

boa/src/builtins/string/mod.rs
```diff
@@ -756,7 +739,12 @@ impl String {
             _ => Some(ctx.to_string(args.get(1).expect("Could not get argument"))?),
```
Razican, line 739: Here too, I think we can use `args.get(1).map().transpose()?;`

boa/src/builtins/string/mod.rs
```diff
@@ -756,7 +739,12 @@ impl String {
...
         };
 
-        Self::string_pad(primitive_val, max_length, fill_string, true)
+        Self::string_pad(
+            primitive_val.to_string(),
+            max_length,
+            fill_string.map(|x| x.to_string()),
```
Razican, line 745: Once again, we should pass the function and not create a closure if possible.

boa/src/builtins/function/mod.rs
```diff
@@ -28,7 +28,7 @@ use gc::{unsafe_empty_trace, Finalize, Trace};
 use std::fmt::{self, Debug};
 
 /// _fn(this, arguments, ctx) -> ResultValue_ - The signature of a built-in function
-pub type NativeFunctionData = fn(&mut Value, &[Value], &mut Interpreter) -> ResultValue;
+pub type NativeFunctionData = fn(&Value, &[Value], &mut Interpreter) -> ResultValue;
```
jasonwilliams, line 31: Whats the reason behind the change here?

HalidOdat: > Whats the reason behind the change here?

It's because the `Value` does not need to be mutable itself, except `Object` but object is covered in a `GcCell<...>` it has interior mutability, so we can `borrow_mut` the `Object` with a immutable reference to `Value`.

This eliminated some places where we were awkwardly cloning the value and passing it to a function because the builtin function required it (`&mut value.clone()`), instead of just passing the value direclty.

boa/src/builtins/object/gcobject.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub(crate) fn new(object: Object) -> Self {
+        Self(Gc::new(GcCell::new(object)))
+    }
+
+    #[inline]
+    pub fn borrow(&self) -> GcCellRef<'_, Object> {
+        match self.try_borrow() {
+            Ok(object) => object,
+            Err(e) => panic!("{}", e),
+        }
```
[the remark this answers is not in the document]
HalidOdat: It's not exactly, because we use `panic!("{}", e)`, this displays the structure with the `Display` trait (the implementation displays the error "object is already mutually borrowed")

The `std::cell::RefCell` `try_borrow` and `try_borrow_mut` just calls `.expect()` as well as having `Display` implementation for the error types `BorrowError` and `BorrowMutError`, I think I'll do the same.

`RefCell`s `borrow` method: https://doc.rust-lang.org/src/core/cell.rs.html#797-799

boa/src/builtins/string/mod.rs
HalidOdat: Yes. it definitely will, out of bounds panic. I did not check that code, there are many places were it should not panic or just throw an exception, I did not focus on these parts, I wanted to do it in a separate PR.

boa/src/builtins/string/mod.rs
HalidOdat: Yes. we can. but I think we can make this even better by accepting a `RcString` instead of `StdString` this will eliminate a `String::clone`, also we can clean it even more if we pass in a `RcString` direclty instead of `Option<RcString>`, because the will string is defaulted if it's `None` to `" "` also in `pad_string` it makes more sense for the fill string that we are padding to be specified.

boa/src/builtins/string/mod.rs
HalidOdat: The bottom comment will remove this also.

HalidOdat:
Lets merge this :)

PR #499, opened 2020-06-16
add Infinity gloabal property (#480)

### LINKED ISSUE #480: Implement global `Infinity` property
**ECMASCript feature**

`Infinity` is not a number, but a global property.

We can add this by adding a new module like `NaN` https://github.com/boa-dev/boa/tree/master/boa/src/builtins/nan and then adding it to the builtin `init` function in `builtins/mod.rs` the value it should contain should be `Value::from(f64::INFINITY)`

More information
    - [MDN][mdn]

[spec]: https://tc39.es/ecma262/#sec-value-properties-of-the-global-object-infinity
[mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Infinity

AnirudhKonduru:
This Pull Request fixes/closes #480.

It changes the following:
 - adds the Infinity global property

### DIFF
--- a/boa/src/builtins/infinity/mod.rs
+++ b/boa/src/builtins/infinity/mod.rs
@@ -1,0 +1,32 @@
+//! This module implements the global `Infinity` property.
+//!
+//! The global property `Infinity` is a numeric value representing infinity.
+//!
+//! More information:
+//!  - [MDN documentation][mdn]
+//!  - [ECMAScript reference][spec]
+//!
+//! [spec]: https://tc39.es/ecma262/#sec-value-properties-of-the-global-object-infinity
+//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Infinity
+
+#[cfg(test)]
+mod tests;
+
+use crate::{builtins::value::Value, BoaProfiler};
+
+/// JavaScript global `Infinity` property.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub(crate) struct Infinity;
+
+impl Infinity {
+    /// The binding name of the property.
+    pub(crate) const NAME: &'static str = "Infinity";
+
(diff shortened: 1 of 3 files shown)


boa/src/builtins/infinity/mod.rs
```diff
@@ -0,0 +1,33 @@
+//! This module implements the global `Infinity` property.
+//!
+//! The global `Infinity` is a property of the global object. In other words,
+//! it is a variable in global scope.
```
HalidOdat, line 4: A better doc would be:
```suggestion
//! The global property `Infinity` is a numeric value representing infinity.
```

boa/src/builtins/infinity/tests.rs
```diff
@@ -0,0 +1,10 @@
+use crate::exec;
+
+#[test]
```
HalidOdat, line 3: Another test would be to test if `Number.POSITIVE_INFINITY === Infinity` should evaluate to `true`

AnirudhKonduru: thanks! I've added 
`
#[test]
fn infinity_exists_and_equals_to_number_positive_infinity_value() {
    let scenario = r#"
        Number.POSITIVE_INFINITY === Infinity;
        "#;

    assert_eq!(&exec(scenario), "true");
}`

looks fine?

boa/src/builtins/infinity/mod.rs
AnirudhKonduru: fixed it, thanks!

boa/src/builtins/infinity/tests.rs
HalidOdat: > looks fine?

Yes. Looks good :)

AnirudhKonduru:
my first PR 😄 @HalidOdat thank you! the info on #480 made it super easy.

HalidOdat:
Lets merge this :)

PR #501, opened 2020-06-16
Added undefined property to global scope

### LINKED ISSUE #481: Implement global `undefined` property
**ECMASCript feature**

`undefined` is not a keyword, but a global property.

We can add this by adding a new module like `NaN` https://github.com/boa-dev/boa/tree/master/boa/src/builtins/nan and then adding it to the builtin `init` function in `builtins/mod.rs` the value it should contain should be `Value::undefined()`

More information
    - [MDN][mdn]

[spec]: https://tc39.es/ecma262/#sec-undefined
[mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined

croraf:
Please merge #488 first. Only the last commit is actually for this feature.

Should close #481 when completed


croraf:
I've made some changes on how `get_binding_value` works. I made it to return an `Option` instead of `Value`. This is important because we should not return `Value::Undefined` when the variable (`Node::Identifier`) is not found, instead error should be explicitly signaled (with `Option(None)` in this case).

boa/src/exec/identifier/mod.rs
```diff
@@ -1,21 +1,17 @@
...
-            ValueData::Undefined => Err(interpreter
+        let reference = interpreter
+            .realm()
+            .environment
+            .get_binding_value(self.as_ref());
+        match reference {
+            Some(value) => Ok(value),
+            None => Err(interpreter
                 .throw_reference_error(self.as_ref())
                 .expect_err("throw_reference_error() must return an error")),
```
HalidOdat, line 14: It's clearer if we use `.construct_reference_error()`, which return `Value` sow we can return `Err(interpreter.construct_reference_error(msg))`

boa/src/exec/operator/mod.rs
```diff
@@ -118,7 +118,8 @@ impl Executable for BinOp {
                     let v_a = interpreter
                         .realm()
                         .environment
-                        .get_binding_value(name.as_ref());
+                        .get_binding_value(name.as_ref())
+                        .expect("Identifier was not initialized");
```
HalidOdat, line 122: Should we not throw a `ReferenceError` here?

boa/src/exec/identifier/mod.rs
Razican: I think we should use `ok_or_else()` here, to make it more readable, with the function you mention as the parameter.

croraf: Applied

boa/src/exec/operator/mod.rs
croraf: You are correct. I made a fix, check latest commit.

boa/src/exec/identifier/mod.rs
```diff
@@ -1,21 +1,16 @@
...
-            _ => Ok(reference),
-        }
+        interpreter
+            .realm()
+            .environment
+            .get_binding_value(self.as_ref())
+            .ok_or_else(|| {
+                interpreter
+                    .throw_reference_error(self.as_ref())
+                    .expect_err("throw_reference_error() must return an error")
```
HalidOdat, line 13: Here we can use the `.construct_reference_error()` so we don't have to call `.expect_err()`
```suggestion
                interpreter.construct_reference_error(self.as_ref())
```

boa/src/exec/operator/mod.rs
```diff
@@ -118,7 +120,12 @@ impl Executable for BinOp {
                     let v_a = interpreter
                         .realm()
                         .environment
-                        .get_binding_value(name.as_ref());
+                        .get_binding_value(name.as_ref())
+                        .ok_or_else(|| {
+                            interpreter
+                                .throw_reference_error(name.as_ref())
+                                .expect_err("throw_reference_error() must return an error")
+                        })?;
```
HalidOdat, line 128: Here too

croraf:
I've done the replacement. So you can merge if everything else is OK.

Not really sure why there exists the `throw_reference_error"` function then at all?

HalidOdat:
> Not really sure why there exists the `throw_reference_error"` function at all? Should it be deleted?

For a log time we only had the `.throw_some_error` which return a `ResultValue` which made throwing errors very easy in builtin functions, but a bit awkward for functions that return `Result<NonValue, Value>`, like `to_number()` which returns `Result<f64, Value>` we had to call `unrechable!` or use `.expect("throw_type_error should always throw")` which made this a bit convoluted.

So we added them in #419 to make it easier, we can use the `.construct_some_error()` in builtin functions, but we have to specify wether it we are throwing with `Err()` or returning an a `Value` with Ok(), so the `.throw_some_error()` are just a convenience functions.

Here we decided to add them: https://github.com/boa-dev/boa/pull/419#discussion_r439063497

PR #515, opened 2020-06-20
Implement Object.is() method issue #513

### LINKED ISSUE #513: Implement `Object.is()` method
**ECMASCript feature**
Implement the `Object.is()` method, the `is` method uses the `SameValue` algorithm (implementation is [here](https://github.com/boa-dev/boa/blob/a7cffe3c7c/boa/src/builtins/value/equality.rs#L144))

The builtin Object functions are [here](https://github.com/boa-dev/boa/blob/a7cffe3c7c/boa/src/builtins/object/mod.rs#L410-L494), after implementing it should be added to `Object` [here](https://github.com/boa-dev/boa/blob/a7cffe3c7c/boa/src/builtins/object/mod.rs#L513) like `defineProperty`

**Example code**

```javascript
Object.is('foo', 'foo');     // true

Object.is('foo', 'bar');     // false
Object.is([], []);           // false

var foo = { a: 1 };
var bar = { a: 1 };
Object.is(foo, foo);         // true
Object.is(foo, bar);         // false

Object.is(null, null);       // true

// Special Cases
Object.is(0, -0);            // false
Object.is(-0, -0);           // true
Object.is(NaN, 0/0);         // true
```

More information:
 - [ECMAScript Specification](https://tc39.es/ecma262/#sec-object.is)
 - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)

tylermorten:
This Pull Request fixes/closes #513.

It changes the following:
 - Implements Object.is() method per issue request specs (w/o strict ?).
 - Adds unit test for example code.

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -31,6 +31,8 @@ use std::{
 };
 
 use super::function::{make_builtin_fn, make_constructor_fn};
+use crate::builtins::value::same_value;
+use crate::builtins::Number;
 pub use internal_state::{InternalState, InternalStateCell};
 
 pub mod internal_methods;
@@ -406,6 +408,34 @@ pub fn make_object(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Resu
     Ok(object)
 }
 
+/// Uses the SameValue algorithm to check equality of objects
+pub fn is(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let x = args.get(0).expect("Cannot get object").clone();
+    let y = args.get(1).expect("Cannot get object").clone();
+
+    if x.get_type() != y.get_type() {
+        return Ok(Value::boolean(false));
+    }
+
+    match (x.data(), y.data()) {
+        (ValueData::BigInt(x), ValueData::BigInt(y)) => {
+            Ok(Value::boolean(BigInt::same_value(x, y)))
+        }
+        (ValueData::Rational(x), ValueData::Rational(y)) => {
+            Ok(Value::boolean(Number::same_value(*x, *y)))
+        }
+        (ValueData::Rational(x), ValueData::Integer(y)) => {
+            Ok(Value::boolean(Number::same_value(*x, f64::from(*y))))
+        }
+        (ValueData::Integer(x), ValueData::Rational(y)) => {
+            Ok(Value::boolean(Number::same_value(f64::from(*x), *y)))
+        }
+        (ValueData::Integer(x), ValueData::Integer(y)) => Ok(Value::boolean(x == y)),
+
+        (_, _) => Ok(Value::boolean(same_value(&x, &y, false))),
+    }
+}
+
 /// Get the `prototype` of an object.
 pub fn get_prototype_of(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
     let obj = args.get(0).expect("Cannot get object");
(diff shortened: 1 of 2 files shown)


boa/src/builtins/object/mod.rs
```diff
@@ -406,6 +408,28 @@ pub fn make_object(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Resu
...
+        (ValueData::BigInt(x), ValueData::BigInt(y)) => BigInt::same_value(x, y),
+        (ValueData::Rational(x), ValueData::Rational(y)) => Number::same_value(*x, *y),
+        (ValueData::Rational(x), ValueData::Integer(y)) => Number::same_value(*x, f64::from(*y)),
+        (ValueData::Integer(x), ValueData::Rational(y)) => Number::same_value(f64::from(*x), *y),
+        (ValueData::Integer(x), ValueData::Integer(y)) => x == y,
+
+        (_, _) => same_value(&x, &y, false),
+    };
+
+    Ok(Value::boolean(result))
```
HalidOdat, line 430: we can use the `same_value` function here we don't have to reimplement it.

I don't know why `same_value` has a `strict` parameter it is not dependent on strict mode

boa/src/builtins/object/mod.rs
```diff
@@ -406,6 +408,28 @@ pub fn make_object(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Resu
     Ok(object)
 }
 
+/// Uses the SameValue algorithm to check equality of objects
+pub fn is(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let x = args.get(0).expect("Cannot get object").clone();
+    let y = args.get(1).expect("Cannot get object").clone();
```
HalidOdat, line 414:
```suggestion
    let x = args.get(0).cloned().unwap_or_else(Value::undefined);
    let y = args.get(1).cloned().unwap_or_else(Value::undefined);
```
Here if a value is not specified it should be `undefined`, we should not panic if one or two of the parameters was not sepecified.

tylermorten:
Got it- I'm on it.  Should have seen that same_value does everything I needed it to do :)

boa/src/builtins/object/mod.rs
```diff
@@ -406,6 +407,20 @@ pub fn make_object(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Resu
...
 }
 
+/// Uses the SameValue algorithm to check equality of objects
+pub fn is(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let x = args.get(0).cloned().unwrap_or_else(Value::undefined);
+    let y = args.get(1).cloned().unwrap_or_else(Value::undefined);
+
+    if x.get_type() != y.get_type() {
+        return Ok(Value::boolean(false));
+    }
```
HalidOdat:
```suggestion
```
We don't have to check if the values are the same type, because `same_value` checks it https://github.com/boa-dev/boa/blob/a7cffe3c7c/boa/src/builtins/value/equality.rs#L151-L153

boa/src/builtins/object/mod.rs
```diff
@@ -406,6 +407,20 @@ pub fn make_object(_: &mut Value, args: &[Value], ctx: &mut Interpreter) -> Resu
     Ok(object)
 }
 
+/// Uses the SameValue algorithm to check equality of objects
+pub fn is(_: &mut Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let x = args.get(0).cloned().unwrap_or_else(Value::undefined);
+    let y = args.get(1).cloned().unwrap_or_else(Value::undefined);
```
HalidOdat, line 413: Maybe we should have a test case for when the arguments are not specified, like the previous comment.

boa/src/builtins/object/mod.rs
HalidOdat: Sorry, I did not see that you removed it :sweat_smile:

PR #517, opened 2020-06-21
New Lexer: Updating parser to use new lexer

Lan2u:
Now that the new lexer has been created the parser must be updated to work with it more closely to allow using GoalSymbols. 

This is part of PR #486 and #432.


Razican:
@Lan2u does this include #486? If so, we can maybe close that PR.

Lan2u:
I have removed the dowhile flag to peek_semicolon as this behaviour only occurs in a single place and by handling it directly within the do_while it removes the need for the cursor to do a .back().

Lan2u:
Merging all the new lexer branches together into a single pull request

PR #518, opened 2020-06-22
TypeError when to_object is passed null or undefined

### LINKED ISSUE #483: `ToObject` (`to_object`) should throw `TypeError` when there is `null` and `undefined`
**Describe the bug**

The problem is that when we try to convert `null` and `undefined` we throw `undefined` this is wrong.

The bug is here: <https://github.com/boa-dev/boa/blob/master/boa/src/exec/mod.rs#L384-L386>
See conversion table for `to_object`: <https://tc39.es/ecma262/#sec-toobject>

To fix this we need to throw a type error with the `Interpreter::throw_type_error()` method the message should be something like `"cannot convert 'null' or 'undefined' to object"`

zanayr:
This Pull Request fixes/closes #483.

It changes the following:
 - Changed the `to_object` implementation to throw a `TypeError` when passed an `undefined` or `null` value.

Best,

### DIFF
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -424,7 +424,7 @@ impl Interpreter {
     /// https://tc39.es/ecma262/#sec-toobject
     #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_object(&mut self, value: &Value) -> ResultValue {
-    match value {
+        match value {
             Value::Undefined | Value::Null => {
                 self.throw_type_error("cannot convert 'null' or 'undefined' to object")
             }
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -414,7 +414,7 @@ impl Interpreter {
     #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_object(&mut self, value: &Value) -> ResultValue {
         match value.data() {
-            ValueData::Undefined | ValueData::Null => Err(Value::undefined()),
+            ValueData::Undefined | ValueData::Null => self.throw_type_error("cannot convert 'null' or 'undefined' to object"),
             ValueData::Boolean(boolean) => {
                 let proto = self
                     .realm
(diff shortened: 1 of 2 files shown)


zanayr:
I think I did this correctly, not sure why I'm getting a large red rectangle and it failed that check. Not finding any useful information..

Please let me know if there is something i need to change/fix.

Best,

Razican:
> I think I did this correctly, not sure why I'm getting a large red rectangle and it failed that check. Not finding any useful information..
> 
> Please let me know if there is something i need to change/fix.
> 
> Best,

This check is related to the test coverage of the code. We should add a test that checks that this works as expected.

zanayr:
Ok! I will fix that tonight.

zanayr:
@Razican I didn't find an existing to_object test, so I worte this test:
```
#[test]
fn to_object() {
    let realm = Realm::create();
    let mut engine = Interpreter::new(realm);
    
    assert!(engine.to_object(&Value::undefined()).is_err());
    assert!(engine.to_object(&Value::null()).is_err());
}
```
My only question is where to put it? in the test.rs file in the `src/exec` directory?

HalidOdat:
> My only question is where to put it? in the test.rs file in the `src/exec` directory?

Yes. that is the best place.

For the test I think it's best if we test that it's an `Object` rather than a thrown `Value`, because this test will pass before the fix too.
```rust
#[test]
fn to_object() {
    let realm = Realm::create();
    let mut engine = Interpreter::new(realm);
    
    assert!(engine.to_object(&Value::undefined()).unwrap_err().is_object());
    assert!(engine.to_object(&Value::null()).unwrap_err().is_object());
}
```

zanayr:
I think i messed something up... 😅

boa/src/exec/mod.rs
```diff
@@ -424,8 +424,10 @@ impl Interpreter {
     /// https://tc39.es/ecma262/#sec-toobject
     #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_object(&mut self, value: &Value) -> ResultValue {
-        match value {
-            Value::Undefined | Value::Null => Err(Value::undefined()),
+    match value.data() {
```
HalidOdat, line 427:
```suggestion
    match value {
```

#498 merged `Value` and `ValueData` together so we dont have to call `.data()`, this should fix it :)

PR #520, opened 2020-06-23
Fix all `Value` operations and add unsigned shift right

HalidOdat:
**What is the problem with the current implementation of `Value`s operations?**
The problem is that they are not spec compliant for objects, for example `ToNumber` (`to_number`) for objects it is supposed to call `to_primitive` and because it does not the code `new Number(6) + 3` is broken and does not give the expected output `9`. Also because `to_number` is called by `Value`s operators, this means that the `Value` operations can throw exceptions and it requires context and should return a `ResultValue`.

It changes the following:
 - Removed `Value::is_true()` (this is a duplicate of `to_boolean`)
 - Added `to_int32`, `to_uint32` and `to_length` to `Interpreter`
 - Made `to_integer` spec compliant
 - Made `Value` operators => functions (these functions take in an `Interpreter`, and can throw in certain cases, see example) 
 - Made all `Value` operations spec compliant.
 - Added fast paths for all `Value` operations

The below code is fixed with this PR:
```js
Number.prototype.valueOf = null;
Number.prototype.toString = null;

new Number(5) + 5 // this should throw, because it calls `to_primitive`
```

I'm not sure if I should move `to_boolean` to `Interpreter`, because it's the only one that can't throw and does not need context

### DIFF
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -26,6 +26,8 @@ use num_traits::float::FloatCore;
 
 mod conversions;
 
+pub(crate) use conversions::{f64_to_int32, f64_to_uint32};
+
 #[cfg(test)]
 mod tests;
 
@@ -33,7 +35,7 @@ const BUF_SIZE: usize = 2200;
 
 /// `Number` implementation.
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct Number(f64);
+pub(crate) struct Number;
 
 /// Maximum number of arguments expected to the builtin parseInt() function.
 const PARSE_INT_MAX_ARG_COUNT: usize = 2;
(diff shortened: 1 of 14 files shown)


Razican:
> I'm not sure if I should move `to_boolean` to `Interpreter`, because it's the only one that can't throw and does not need context

I would move it. It makes sense having everything that's related working similarly and having the same endpoint. If not, it could get a bit messy.

boa/src/builtins/number/mod.rs
```diff
@@ -27,11 +24,16 @@ use crate::{
...
+
+#[cfg(test)]
+mod tests;
+
 const BUF_SIZE: usize = 2200;
 
 /// `Number` implementation.
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct Number;
+pub(crate) struct Number(f64);
```
Razican, line 36: Do you think we could/should have a fast path for integers here?

HalidOdat: > Do you think we could/should have a fast path for integers here?

If you mean to make the Number Object have a integer and rational, I don't think this would be beneficial since every number operation converts the object to a primitive, it is not like we are doing with primitives, and it would complicate the code.

I have some progress locally were I made a fast path for when the value is a `Value::Integer` for `to_int32` and `to_uint32`.

HalidOdat:
This PR fixes all `Value` operations, and adds unsigned shift left.

The `Value` `to_` methods are still there but I removed there dependency from `Value` operations and in future PR I intead to remove them entirely.

This PR is ready so should we put this in `v0.9` or leave it for `v0.10`

PR #522, opened 2020-06-23
Regress

### LINKED ISSUE #478: Use ECMAScript syntax for regular expressions
**Current situation**
Currently, we use regular expressions from the `regex` crate. They are extremely fast, but they don't support the full ECMAScript syntax like backreferences or lookaround assertions.

**Solution**
We should use a library that does support all this syntax. A nice library that was just released and seems to support this is [`regress`](https://crates.io/crates/regress).

We should remove our dependency on `regex` and use `regress`, and see how the performance gets affected.

neeldug:
This Pull Request fixes/closes #478.

It changes the following:
 - Uses regress instead of regex
Current issues:
-Refactoring regexp/mod.rs with regress.


jasonwilliams:
Hows this going @neeldug did you need help?

neeldug:
Yes sorry, was hitting a few issues with certain methods used that aren’t available in Regress, any help would be much appreciated.

neeldug:
Hey @jasonwilliams I've been stuck on a few areas where boa uses capture group methods which are currently unimplemented in regress, specifically the replace method, and so I'm unsure of how to proceed on this.

neeldug:
Oh, and I almost forgot to ask, [here](https://github.com/boa-dev/boa/blob/e49be361fa7bc6434f7558ff8da2292f05b83beb/boa/src/builtins/regexp/mod.rs#L296), find_at() is used but this is unimplemented, however, I was wondering whether this was done as a performance benefit or whether it was done to avoid previous matches, if the latter should I consider splitting into a substring?

PR #523, opened 2020-06-24
clean up the rest of the math methods

### LINKED ISSUE #514: Cleanup `Math` methods
Many of the Math builtin functions are a bit convoluted for example:
```rust
pub(crate) fn abs(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
    Ok(Value::from(if args.is_empty() {
       f64::NAN
    } else {
        f64::from(args.get(0).expect("Could not get argument")).abs()
    }))
}
```
This is also inefficient because we check if there is an argument and then another check is done when we call `.expect()`.

A better way of doing this would be to use `.map_or` which would simplify the code a lot:
```rust
pub(crate) fn abs(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
	Ok(args.get(0).map_or(f64::NAN, |x| f64::from(x).abs()).into())
}
```

n14little:
This Pull Request fixes/closes #514

It changes the following:
 - Refactors the following methods:
   - `Math::atan2`
   - `Math::log`
   - `Math::log10`
   - `Math::log2`
   - `Math::min`
   - `Math::pow`
   - `Math::sign`

### DIFF
--- a/boa/src/builtins/math/mod.rs
+++ b/boa/src/builtins/math/mod.rs
@@ -134,12 +134,10 @@ impl Math {
     /// [spec]: https://tc39.es/ecma262/#sec-math.atan2
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2
     pub(crate) fn atan2(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
-        Ok(Value::from(if args.is_empty() {
-            f64::NAN
-        } else {
-            f64::from(args.get(0).expect("Could not get argument"))
-                .atan2(args.get(1).expect("Could not get argument").to_number())
-        }))
+        match (args.get(0), args.get(1)) {
+            (Some(y), Some(x)) => Ok(Value::from(f64::from(y).atan2(f64::from(x)))),
+            _ => Ok(Value::from(f64::NAN)),
+        }
     }
 
     /// Get the cubic root of a number.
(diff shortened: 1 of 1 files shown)


boa/src/builtins/math/mod.rs
```diff
@@ -134,12 +134,10 @@ impl Math {
...
-        Ok(Value::from(if args.is_empty() {
-            f64::NAN
-        } else {
-            f64::from(args.get(0).expect("Could not get argument"))
-                .atan2(args.get(1).expect("Could not get argument").to_number())
-        }))
+        match (args.get(0), args.get(1)) {
+            (Some(y), Some(x)) => Ok(Value::from(f64::from(y).atan2(f64::from(x)))),
+            _ => Ok(Value::from(f64::NAN)),
+        }
```
HalidOdat, line 140:
```suggestion
        Ok(match (args.get(0), args.get(1)) {
            (Some(y), Some(x)) => f64::from(y).atan2(f64::from(x)),
            _ => f64::NAN,
        }.into())
```
I think it would be clearer if we take the `Ok(Value::from(...))` and move it to the match level.

boa/src/builtins/math/mod.rs
```diff
@@ -326,13 +326,12 @@ impl Math {
...
-            num.powf(power)
-        } else {
-            f64::NAN
-        }))
+        match (args.get(0), args.get(1)) {
+            (Some(base), Some(exponent)) => {
+                Ok(Value::from(f64::from(base).powf(f64::from(exponent))))
+            }
+            _ => Ok(Value::from(f64::NAN)),
+        }
```
HalidOdat, line 334: Here too.

boa/src/builtins/math/mod.rs
n14little: :heavy_check_mark:

boa/src/builtins/math/mod.rs
n14little: :heavy_check_mark:

PR #526, opened 2020-06-25
Fix sending `this` value to function environments

### LINKED ISSUE #511: [Call] Usage of "this" in methods is not supported
**Describe the bug**
```js
const foo = {
  a: 3,
  bar: function () {return this.a + 5}
};

foo.bar()
```

should return 8. Instead `thread 'main' panicked at 'Reference Error: Unitialised binding for this function', boa/src/environment/function_environment_record.rs:115:17` happens

jasonwilliams:
* Arrow functions are lexical and have no this, they should also pass `None` for initial `this` value
* Call `bind_this_value` to properly set up `this` binding in functions
* tasks.json should not have problem matcher due to this issue https://github.com/rust-analyzer/rust-analyzer/issues/4768

This Pull Request fixes/closes #511.


.editorConfig
```diff
@@ -3,3 +3,9 @@ root = true
 [{Makefile,**.mk}]
 # Use tabs for indentation (Makefiles require tabs)
 indent_style = tab
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
Razican, line 11: I would add a new line at the end of this file.

jasonwilliams: Yeah i don't know why my VSCode is not respecting that file for this haha

PR #525, opened 2020-06-24
add remaining math methods (#524)

mr-rodgers:
This Pull Request fixes/closes #524.

It changes the following (work-in-progress):

 - [x] Math.clz32()
 - [x] Math.expm1()
 - [x] Math.fround()
 - [x] Math.hypot()
 - [x] Math.log1p()
 - [x] Math.imul()

PR #528, opened 2020-06-25
v0.9 changelog

jasonwilliams:


jasonwilliams:
@HalidOdat i wonder if we can get better before / after images now there's been more work done

jasonwilliams:
We could add a blog repo to this organisation, and clone https://github.com/rust-community/rust-lang-blog-boilerplate

HalidOdat:
> @HalidOdat i wonder if we can get better before / after images now there's been more work done

Made a simple bash script for generating the profiles with specified branches/commits, I can go as back as the time mesureme was introduced to boa and as forward as master `HEAD`. Is there a specific range you want me to profile?

HalidOdat:
Did a before and after from commit `32b0741c` to master `HEAD`
**`32b0741c`:** (when #317 was merged, `58ms`)

32b0741c

and master **`HEAD`** (`15ms`):
master

HalidOdat:
Did a benchmark of `Realm creation` and its a `~70%` improvement :)

HalidOdat:
Maybe we should have a section for `Internal improvements` rather than everything being a feature, for example #465 does not add a new feature for the user to use, it is only an optimization and makes it easier to add other native types.

What do you think? @jasonwilliams @Razican

pedropaulosuzuki:
Could #525 be packed into v0.9? Would be pretty nice to have feature-complete Math object in the new release, since we got bigint in the last v0.8 release, and from the PR it seems to be ready/close to ready.

> Maybe we should have a section for `Internal improvements` rather than everything being a feature, for example #465 does not add a new feature for the user to use, it is only an optimization.

I think this makes a lot of sense, so that users can find what's new with ease and also see what's new on the internals of Boa, and what has been done to make it so much faster on each update.

> Did a benchmark of `Realm creation` and its a `~70%` improvement :)

That's amazing!!

HalidOdat:
> Could #525 be packed into v0.9? Would be pretty nice to have feature-complete Math object in the new release, since we got bigint in the last v0.8 release, and from the PR it seems to be ready/close to ready.

Sure! I don't see a reason why it shouldn't go in `v0.9` :)

CHANGELOG.md
```diff
@@ -1,8 +1,117 @@
...
+- [BUG #409](https://github.com/boa-dev/boa/pull/409):
+  Fix function object constructable/callable (@HalidOdat)
+- [BUG #403](https://github.com/boa-dev/boa/issues/403)
+  `Value::to_json()` does not handle `undefined` correctly (@n14little)
+- [BUG #443](https://github.com/boa-dev/boa/issues/443):
+  HasOwnProperty should call GetOwnProperty and not GetProperty (@n14little)
+- [BUG #210](https://github.com/boa-dev/boa/issues/210):
+  builtinfun.length undefined (@Croraf)
+- [BUG #466](https://github.com/boa-dev/boa/issues/466):
+  Change `ToPrimitive()` (`to_primitive()`) hint to be an enum, instead of string (@HalidOdat)
```
HalidOdat, line 83:
```suggestion
  Change `ToPrimitive()` (`to_primitive()`) hint to be an enum, instead of string (@Colin-Suckow)
```
I was not the one who created the PR, I just gave some advice, I don't think I should get all the credit :)

jasonwilliams:
I’m hoping to set up a blog so we can move a lot of content over to there, then I can update the change log

jasonwilliams:
we're now up and running on https://boa-dev.github.io so we just need to write up the content

jasonwilliams:
internel improvements have been added

jasonwilliams:
I plan to release tomorrow, just need to finish blog post now

PR #529, opened 2020-06-25
Refactor exec/expression into exec/call and exec/new

croraf:
Minor refactor. Also added BoaProfiled to `impl Executable for New`

### DIFF
--- a/boa/src/exec/new/mod.rs
+++ b/boa/src/exec/new/mod.rs
@@ -1,0 +1,39 @@
+use super::{Executable, Interpreter};
+use crate::{
+    builtins::{
+        object::{ObjectData, INSTANCE_PROTOTYPE, PROTOTYPE},
+        value::{ResultValue, Value},
+    },
+    syntax::ast::node::New,
+    BoaProfiler,
+};
+
+impl Executable for New {
+    fn run(&self, interpreter: &mut Interpreter) -> ResultValue {
+        let _timer = BoaProfiler::global().start_event("New", "exec");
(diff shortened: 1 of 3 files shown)


boa/src/exec/new/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+    BoaProfiler,
+};
+
+impl Executable for New {
+    fn run(&self, interpreter: &mut Interpreter) -> ResultValue {
+        let _timer = BoaProfiler::global().start_event("New", "exec");
+        // let (callee, args) = match call.as_ref() {
+        //     Node::Call(callee, args) => (callee, args),
+        //     _ => unreachable!("Node::New(ref call): 'call' must only be Node::Call type."),
+        // };
```
Razican, line 17: Can this be removed now? I think a `New` node should always have a `Call`, so it's checked at the type level, right?

croraf: I don't understand your comment. Do you mean to remove the comments and leave the timer? 

The comments were here before my work, if you want me to remove them I will?

Razican: Yes, sorry, I selected the timer too, but I meant the comments. a `New` node will now always have a `Call`, so no need to check for it.

I was wondering, though, should we move the New/Call AST nodes too from the expression module in `syntax::ast`?

croraf: I think this would be good, yes.

croraf: Check PR #542

Issue #533, opened 2020-06-27
[exec - function] function.prototype doesn't have own constructor property pointing to this function


croraf:
**Describe the bug**

`function.prototype` should have an own `constructor` property pointing back to that function.

**To Reproduce**
```js
const a = function asd() {};
a
```

**Expected behavior**
Should output this:

```js
{
  constructor: {
    name: asd,
      prototype: [Cycle],
        length: 0,
          __proto__: {
      constructor: {
        name: Function,
          prototype: [Cycle],
            length: 1,
              __proto__: undefined
      },
      __proto__: undefined
    }
  },
  __proto__: {
    constructor: {
      setPrototypeOf: {
        length: 2
      },
      prototype: [Cycle],
        name: Object,
          length: 1,
            defineProperty: {
        length: 3
      },
      getPrototypeOf: {
        length: 1
      },
      is: {
        length: 2
      },
      __proto__: {
        constructor: {
          name: Function,
            prototype: [Cycle],
              length: 1,
                __proto__: undefined
        },
        __proto__: undefined
      }
    },
    hasOwnProperty: {
      length: 0
    },
    propertyIsEnumerable: {
      length: 0
    },
    toString: {
      length: 0
    }
  }
}
```
Instead top level `constructor` field is not present in the response.

PR #534, opened 2020-06-27
[exec - function] Add constructor property to function.prototype

croraf:
Intended to solve this #533 .
It is actually a couple of lines of change, but I based it on refactor PR which should be merged first: https://github.com/boa-dev/boa/pull/529

I basically added `proto.set_field("constructor", val.clone());` to `function.prototype`. `val` here is:

```rust
let new_func = Object::function(func);
let val = Value::from(new_func);
```
https://github.com/boa-dev/boa/pull/534/files#diff-d7bd296d9ca534161e198b8c141ca51fR140-L149

Someone has to doublecheck if setting the `val.clone()` here is correct.


RageKnify:
@croraf I think #642 fixed this #533, this PR can be closed, right?

HalidOdat:
> @croraf I think #642 fixed this #533, this PR can be closed, right?

Yep #642 did add that, so closing this.

PR #536, opened 2020-06-28
Feature `SyntaxError`

HalidOdat:
It changes the following:
 - Implement `SyntaxError` object
 - Add `.construct_syntax_error()` and `.throw_syntax_error()`
 - Made error objects construction use `to_string()` (spec compliant and some cleanup)
 - Fix missing `name` field in `prototype` of error objects

### DIFF
--- a/boa/src/builtins/error/mod.rs
+++ b/boa/src/builtins/error/mod.rs
@@ -20,17 +20,19 @@ use crate::{
     profiler::BoaProfiler,
 };
 
-// mod eval;
 pub(crate) mod range;
 pub(crate) mod reference;
 pub(crate) mod syntax;
 pub(crate) mod r#type;
-// mod uri;
+// pub(crate) mod eval;
+// pub(crate) mod uri;
 
 pub(crate) use self::r#type::TypeError;
 pub(crate) use self::range::RangeError;
 pub(crate) use self::reference::ReferenceError;
 pub(crate) use self::syntax::SyntaxError;
+// pub(crate) use self::eval::EvalError;
+// pub(crate) use self::uri::UriError;
 
 /// Built-in `Error` object.
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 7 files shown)


boa/src/builtins/error/syntax.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// [spec]: https://tc39.es/ecma262/#sec-error.prototype.tostring
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString
+    #[allow(clippy::wrong_self_convention)]
+    pub(crate) fn to_string(this: &Value, _: &[Value], _: &mut Interpreter) -> ResultValue {
+        let name = this.get_field("name");
+        let message = this.get_field("message");
+        Ok(format!("{}: {}", name, message).into())
+    }
+
+    /// Create a new `RangeError` object.
```
n14little, line 65: Shouldn't this be SyntaxError?

boa/src/builtins/error/mod.rs
```diff
@@ -77,7 +72,7 @@ impl Error {
         Ok(Value::from(format!("{}: {}", name, message)))
     }
 
-    /// Create a new `Error` object.
+    /// Create a new `SyntaxError` object.
```
Razican, line 75: I think this was correct before, right?

boa/src/builtins/error/mod.rs
```diff
@@ -20,16 +20,19 @@ use crate::{
...
 
-// mod eval;
 pub(crate) mod range;
 pub(crate) mod reference;
-// mod syntax;
+pub(crate) mod syntax;
 pub(crate) mod r#type;
-// mod uri;
+// pub(crate) mod eval;
+// pub(crate) mod uri;
```
Razican, line 28: Why is this commented out?

HalidOdat: > Why is this commented out?

Because we haven't implemented them yet

PR #537, opened 2020-06-30
Upgraded dependencies before the 0.9 release

Razican:
This is just a minor dependency upgrade for the upcoming 0.9 release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,9 +150,9 @@ dependencies = [
 
 [[package]]
 name = "criterion"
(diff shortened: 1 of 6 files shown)


jasonwilliams:
Looks like it broke the benchmarks, what changed in Criterion?

Razican:
> Looks like it broke the benchmarks, what changed in Criterion?

Not sure, but it seems that it broke critcmp, so I'm reverting that upgrade. We can deal with it in 0.10.

PR #543, opened 2020-07-03
[builtins - object] Object.create

croraf:
Implements Object.create builtin method

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -414,12 +414,18 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
 ///
 /// https://tc39.es/ecma262/#sec-object.create
 /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
-pub fn create_builtin(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+pub fn create_builtin(_: &Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
     let __proto__ = args.get(0).cloned().unwrap_or_else(Value::undefined);
-    Ok(Value::new_object_from_prototype(
-        __proto__,
-        ObjectData::Ordinary,
-    ))
+    match __proto__ {
+        Value::Object(_) | Value::Null => Ok(Value::new_object_from_prototype(
+            __proto__,
+            ObjectData::Ordinary,
+        )),
+        _ => interpreter.throw_type_error(format!(
+            "Object prototype may only be an Object or null: {}",
+            __proto__
+        )),
+    }
 }
 
 /// Uses the SameValue algorithm to check equality of objects
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -410,12 +410,25 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
     Ok(object)
 }
 
-/// Creates a new object from the provided prototype
+
+/// `Object.create( proto, [propertiesObject] )`
 ///
-/// https://tc39.es/ecma262/#sec-object.create
-/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
+/// Creates a new object from the provided prototype.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-object.create
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
 pub fn create_builtin(_: &Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
     let __proto__ = args.get(0).cloned().unwrap_or_else(Value::undefined);
+    let properties = args.get(1).cloned().unwrap_or_else(Value::undefined);
+
+    if properties != Value::Undefined {
+        unimplemented!("propertiesObject argument of Object.create")
+    }
+
     match __proto__ {
         Value::Object(_) | Value::Null => Ok(Value::new_object_from_prototype(
             __proto__,
@@ -536,7 +549,7 @@ pub fn create(global: &Value) -> Value {
     let object = make_constructor_fn("Object", 1, make_object, global, prototype, true);
 
     // static methods of the builtin Object
-    make_builtin_fn(create_builtin, "create", &object, 1);
+    make_builtin_fn(create_builtin, "create", &object, 2);
     make_builtin_fn(set_prototype_of, "setPrototypeOf", &object, 2);
     make_builtin_fn(get_prototype_of, "getPrototypeOf", &object, 1);
     make_builtin_fn(define_property, "defineProperty", &object, 3);
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -411,7 +411,7 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
 }
 
 /// Creates a new object from the provided prototype
-/// 
+///
 /// https://tc39.es/ecma262/#sec-object.create
 /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
 pub fn create_builtin(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -411,6 +411,9 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
 }
 
 /// Creates a new object from the provided prototype
+/// 
+/// https://tc39.es/ecma262/#sec-object.create
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
 pub fn create_builtin(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
     let __proto__ = args.get(0).cloned().unwrap_or_else(Value::undefined);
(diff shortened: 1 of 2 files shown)


boa/src/builtins/object/mod.rs
```diff
@@ -410,6 +410,16 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
...
 }
 
+/// Creates a new object from the provided prototype
+///
+/// https://tc39.es/ecma262/#sec-object.create
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
+pub fn create_builtin(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let __proto__ = args.get(0).cloned().unwrap_or_else(Value::undefined);
+    let new_object = Value::new_object_from_prototype(__proto__, ObjectData::Ordinary);
+    Ok(new_object.into())
```
HalidOdat, line 420:
```suggestion
    Ok(Value::new_object_from_prototype(__proto__, ObjectData::Ordinary))
```
Fix clippy error

boa/src/builtins/object/mod.rs
```diff
@@ -410,6 +410,18 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
     Ok(object)
 }
 
+/// Creates a new object from the provided prototype
+///
+/// https://tc39.es/ecma262/#sec-object.create
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
+pub fn create_builtin(_: &Value, args: &[Value], _: &mut Interpreter) -> ResultValue {
+    let __proto__ = args.get(0).cloned().unwrap_or_else(Value::undefined);
```
jasonwilliams, line 418: I think we need to make sure `__proto__` has the type `object` or `null` before moving on.

jasonwilliams, line 418: I think we need to make sure __proto__ has the type object or null before moving on.

croraf: Added

boa/src/builtins/object/mod.rs
```diff
@@ -410,6 +410,24 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
...
+        Value::Object(_) | Value::Null => Ok(Value::new_object_from_prototype(
+            __proto__,
+            ObjectData::Ordinary,
+        )),
+        _ => interpreter.throw_type_error(format!(
+            "Object prototype may only be an Object or null: {}",
+            __proto__
+        )),
+    }
+}
```
HalidOdat, line 429: We should rename this to `prototype`, rather than it being `__proto__`.

Even in the standard it says:
> The create function creates a new object with a specified prototype. When the create function is called, the following steps are taken:

croraf: Both `prototype` and `__proto__` point to the "prototype". 

`prototype` is the property of the creator function (so used in that context), while `__proto__` is the property of the "child object" (so used in that context).  

That's why I think `__proto__` is better here.

HalidOdat: > Both `prototype` and `__proto__` point to the "prototype".
> 
> `prototype` is the property of the creator function (so used in that context), while `__proto__` is the property of the "child object" (so used in that context).
> 
> That's why I think `__proto__` is better here.

I know both point to the same thing, but It's more semantically meaningful when you call `Object.create(String.prototype)` we create an object from the prototype, rather than the prototype instance.

HalidOdat, line 429: The `Object.create()` method can also take arguments ([properties](https://tc39.es/ecma262/#sec-object.create)) that are defined on the created object.

croraf: Maybe I missed the point of your last comment. But you wouldn't call `Object.create`  with the functions prototype. Instead you would do `new SomeFunction ()`. I think you call `Object.create` directly with the  __proto__ instance.

croraf: Very good point. I think this requires the functionality for the `Object.defineProperties()` which is currently not implemented, so I think this should be implemented first or at the same time.

HalidOdat: > Maybe I missed the point of your last comment, but you wouldn't call `Object.create` with the function's prototype. Instead you would do `new SomeFunction ()`. I think in vast majority of cases you call `Object.create` directly with the `__proto__` instance.

Sorry, I'll try to explain this a bit better, I'll take an examle from mdn:
```js
const person = {
  isHuman: false,
  printIntroduction: function() {
    console.log(`My name is ${this.name}. Am I human? ${this.isHuman}`);
  }
};

const me = Object.create(person);
```
In this case we create a new object with the `prototype` (`person`), and when the object is created the `[[prototype]]` (prototype instance) of the new object is set to `prototype`, it's set in `step. 2` in <https://tc39.es/ecma262/#sec-object.create> which in boa is `Value::new_object_from_prototype()` in creates the instance of the `prototype` (`__proto__`)

HalidOdat: > Very good point. I think this requires the functionality for the `Object.defineProperties()` which is currently not implemented, so I think this should be implemented first. Would it be acceptable that I put a placeholder and open a new issue to make `Object.defineProperties()` and mention in that issue that after it is done the `Object.create()` should be improved with this functionality and some test-cases?

Yes that's fine, maybe we should have a case for when there are more arguments than one with `unimplemented!("propertiesObject is not implemented")`

boa/src/builtins/object/mod.rs
```diff
@@ -410,6 +410,24 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
     Ok(object)
 }
 
+/// Creates a new object from the provided prototype
+///
+/// https://tc39.es/ecma262/#sec-object.create
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
```
HalidOdat, line 416: we should change the documentation here to be like `Object.hasOwnProperty()` (`has_own_property`), I know some don't have it like `Object.is()` we should also update them, we can change it in this PR or in another one for `is`

boa/src/builtins/object/mod.rs
croraf: OK, but it is not clear to me that naming it `prototype` locally in this function is any better than `__proto__`. Even in https://tc39.es/ecma262/#sec-ordinaryobjectcreate the parameter is called _proto_. So this might even be a bit better. (Although in the function we are talking about they call it as `O`). 

The function we call from this point `pub fn new_object_from_prototype(proto: Value, data: ObjectData)` also uses the argument `proto`, as in the spec. (although the name of this function is not aligned with the spec's one).

HalidOdat: > OK, but it is not clear to me that naming it `prototype` locally in this function is any better than `__proto__`. Even in [tc39.es/ecma262/#sec-ordinaryobjectcreate](https://tc39.es/ecma262/#sec-ordinaryobjectcreate) the parameter is called _proto_. So this might even be a bit better. (Although in the function we are talking about they name this parameter as `O`).

Yes. the spec is confusing at times, I think it's `O` (as in object) because the prototoype can be any object (like the example above with `person` object) and technically `null` is an object (`typeof null == "object"`)

You can know it's an instance if there is `[[Prototype]]`, otherwise assume it's `prototype` (the spec is not always consistent).

Some good informative blogs are the V8 `Understanding the ECMAScript spec` there are currently 4 parts here is part one https://v8.dev/blog/understanding-ecmascript-part-1

croraf: So you still think it is important to rename this local variable from `__proto__` to `prototype`? If so, I'll change it.

HalidOdat: > So you still think it is important to rename this local variable from `__proto__` to `prototype`? If so, I'll change it.

Yes. It makes more sense to name it `prototype`.

boa/src/builtins/object/mod.rs
```diff
@@ -410,6 +410,37 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
...
+///
+/// [spec]: https://tc39.es/ecma262/#sec-object.create
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
+pub fn create_builtin(_: &Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
+    let __proto__ = args.get(0).cloned().unwrap_or_else(Value::undefined);
+    let properties = args.get(1).cloned().unwrap_or_else(Value::undefined);
+
+    if properties != Value::Undefined {
+        unimplemented!("propertiesObject argument of Object.create")
+    }
```
HalidOdat, line 429: we can just check if there are more than one rather than copying the value:
```suggestion
    if args.len() > 1 {
        unimplemented!("propertiesObject argument of Object.create")
    }
```

croraf: This would also fail `Object.create({}, undefined)`, which passes in the current implementation? And also cases like: 
image

Although contrived examples, I think we can leave the current implementation.

HalidOdat: I see. we can leave it as is for now.

boa/src/builtins/object/mod.rs
```diff
@@ -410,6 +410,36 @@ pub fn make_object(_: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultVa
...
+///
+/// Creates a new object from the provided prototype.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-object.create
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
+pub fn create_builtin(_: &Value, args: &[Value], interpreter: &mut Interpreter) -> ResultValue {
```
HalidOdat, line 423: we should rebase on the latest commit on master, `init` and `create` has been merged into just init which would allow us to call this `create` instead of `create_builtin`

PR #547, opened 2020-07-04
Merged `create` into `init` for builtins

HalidOdat:
It changes the following:
 - Merged `create` into `init`

### DIFF
--- a/boa/src/builtins/symbol/tests.rs
+++ b/boa/src/builtins/symbol/tests.rs
@@ -1,12 +1,4 @@
-use super::*;
 use crate::{exec::Interpreter, forward, forward_val, realm::Realm};
-
-#[test]
-fn check_symbol_constructor_is_function() {
-    let global = Value::new_object(None);
-    let symbol_constructor = Symbol::create(&global);
-    assert_eq!(symbol_constructor.is_function(), true);
-}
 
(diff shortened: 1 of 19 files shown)


boa/src/builtins/symbol/tests.rs
```diff
@@ -1,13 +1,5 @@
-use super::*;
 use crate::{exec::Interpreter, forward, forward_val, realm::Realm};
 
-#[test]
-fn check_symbol_constructor_is_function() {
-    let global = Value::new_object(None);
-    let symbol_constructor = Symbol::create(&global);
-    assert_eq!(symbol_constructor.is_function(), true);
-}
-
```
jasonwilliams, line 10: Is there a reason you've removed these tests?

HalidOdat: > Is there a reason you've removed these tests?

It's because `create` function was removed.

jasonwilliams: Do we need to care that init returns a function type like before?

HalidOdat: Not really, if init was to not return a function for example symbol all the other tests calling symbol would fail so that scenario is covered.

PR #549, opened 2020-07-05
label statement implementation

jasonwilliams:
fixes #549 

@Razican it looks like the [Statement parse](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/parser/statement/mod.rs#L93-L176) will need to be passed an optional `label` argument which then is also passed into all of the sub statements, what do you think? 

~Also what is the Output type for labelled Statement? As it technically, it is allowed to wrap any statement~ I think this could be Node

Try with 
```js
let str = "";

loop1: for (let i = 0; i < 5; i++) {
  loop2: for (let b = 0; b < 5; b++) {
    if (b === 2) {
      break loop1;
    }
    str = str + b;
  }
  console.log("never reaches here");
  str = str + i;
}

console.log(str);
// expected output: "01
```


Razican:
> @Razican it looks like the [Statement parse](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/parser/statement/mod.rs?rgh-link-date=2020-07-05T12%3A53%3A35Z#L93-L176) will need to be passed an optional `label` argument which then is also passed into all of the sub statements, what do you think?

I think we should have a `LabeledStatement` AST node, that would contain a `Box<str>` as the label, and a `Node` as the statement. This would allow not having to pass this label to all the sub statements. If we are parsing a labeled statement, we take the label and store it in a string, then we parse the statement as we did before, and get the `Node`.

> ~Also what is the Output type for labelled Statement? As it technically, it is allowed to wrap any statement~ I think this could be Node

I think it should be the `LabeledStatement` type I explained above.

boa/src/syntax/parser/statement/labelled_stm/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+/// Labelled Statement Parsing
+///
+/// More information
+/// - [MDN documentation][mdn]
+/// - [ECMAScript specification][spec]
+///
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label
+/// [spec]: https://tc39.es/ecma262/#sec-labelled-statements
+#[derive(Debug, Clone, Copy)]
+pub(super) struct Label {
```
Razican, line 20: I would keep the name of the `LabelledStatement` as in the spec. This parser would return a `LabelledStatement` node, with a label and a `Node`.

jasonwilliams:
@Razican i can't think of how we would do this without either [passing a label into the run](https://github.com/jasonwilliams/boa/blob/master/boa/src/exec/iteration/mod.rs#L16) method, or passing the label when Label creates a statement (sub node) https://github.com/jasonwilliams/boa/blob/feature/548/boa/src/syntax/parser/statement/labelled_stm/mod.rs#L49

jcdickinson:
A much simpler implementation would be to add the label to the loop statement AST nodes. Then, the loops can simply:

```rust
InterpreterState::Break(label) => {
    // Loops 'consume' breaks, only if the label is not present, or does not match the current loop.
    match label {
        None => interpreter.set_current_state(InterpreterState::Executing),
        Some(_) if label == self.label() => interpreter.set_current_state(InterpreterState::Executing),
        _ => {}
    }
    break;
}
// ... Likewise for Continue.
```

I'm pretty certain that would work. This uses the native/Rust stack as the label stack, completely avoiding that allocation. Obviously, this allows invalid JS:

```js
label1: while(true) {
  break label2;
}
```

So the validity of labels would need to be checked in the parser.

Razican:
The issue with this approach is that it only allows labelled loops, but, functions and other statements can also have a label: https://tc39.es/ecma262/#sec-labelled-statements

jasonwilliams:
The direction I’ve gone in is that Nodes which currently use a label will have a label property and use it as needed. So not too different from your example.
During parsing the label_statement will set the label on that node and return that node.

If we can set the label on the statement then there’s no need for a label node during execution, also if we need to add labels to new nodes in future it will be easy.

boa/src/syntax/ast/node/labelled_smt.rs
```diff
@@ -0,0 +1,38 @@
+use super::Node;
+use gc::{Finalize, Trace};
+use std::fmt;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct Label {
```
Razican, line 10: I would call this with the same name as in the spec: `LabelledStatement`, especially because it could be seen as if this were just the label, without the statement.

boa/src/syntax/ast/node/mod.rs
```diff
@@ -116,6 +118,9 @@ pub enum Node {
     /// An 'if' statement. [More information](./conditional/struct.If.html).
     If(If),
 
+    /// A labelled Statement.
+    Label(Label),
```
Razican, line 122: Here too, I would use the same name as in the spec: `LabelledStatement`. In fact, I think we should change other node variant names too, but in a different PR.

boa/src/syntax/parser/statement/iteration/for_statement.rs
```diff
@@ -31,7 +31,7 @@ use crate::{
 ///
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for
 /// [spec]: https://tc39.es/ecma262/#sec-for-statement
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
```
Razican, line 34: Why is this no longer `Copy`?

boa/src/syntax/parser/statement/labelled_stm/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::{
+    syntax::{
+        ast::{node::Label, Punctuator},
+        parser::{
+            cursor::Cursor, error::ParseError, AllowAwait, AllowReturn, AllowYield, TokenParser,
+        },
+    },
+    BoaProfiler,
+};
+/// Labelled Statement Parsing
```
Razican, line 11:
```suggestion

/// Labelled Statement Parsing
```

boa/src/syntax/ast/node/labelled_smt.rs
Razican, line 10: This would need some module-level, struct-level and method-level documentation.

Razican (commented):
After the last comment, I thought we would have stuff such as an `Option<Box<str>>` for a label in some nodes, but the code here still has a `Label` node. I did a review with some comments though.

I do like the idea of having an optional label in the statements that allow it, that would make everything easier, and I don't think it will grow memory usage too much.

boa/src/syntax/parser/statement/iteration/for_statement.rs
jasonwilliams: hmm don't know why that was removed, adding back in

jasonwilliams:
Label Node is no longer needed.
Instead Labelled_stmt sets the label on the stmt during parsing.

Then execution remains as it was before (no change there).

`Label` can't set the label as a Node during exec because that would require the node to be mutable and would require changing all of the traits we have for exec.

boa/src/syntax/parser/statement/mod.rs
```diff
@@ -170,8 +172,27 @@ where
...
+                        self.allow_yield,
+                        self.allow_await,
+                        self.allow_return,
+                    )
+                    .parse(cursor)
+                    .map(Node::from);
+                }
+
+                ExpressionStatement::new(self.allow_yield, self.allow_await).parse(cursor)
+            }
```
jasonwilliams, line 195: @Razican Is there a nicer way to achieve this?

boa/src/syntax/ast/node/iteration.rs
```diff
@@ -21,6 +21,7 @@ use serde::{Deserialize, Serialize};
 pub struct ForLoop {
     #[cfg_attr(feature = "serde", serde(flatten))]
     inner: Box<InnerForLoop>,
+    pub(crate) label: Option<Box<str>>,
```
HalidOdat, line 24: Its better to create a getter for this `.label()` that returns a `Option<&str>`

boa/src/exec/iteration/mod.rs
```diff
@@ -34,8 +34,22 @@ impl Executable for ForLoop {
...
             match interpreter.executor().get_current_state() {
-                InterpreterState::Break(_label) => {
-                    // TODO break to label.
+                InterpreterState::Break(label) => {
+                    // If a label is set we want to break the current block and still keep state as Break if the label is a block above
+                    if let Some(stmt_label) = &self.label {
+                        if let Some(brk_label) = label {
+                            // We have a label, but not for the current statement
+                            // break without resetting to executings
+                            if stmt_label.to_string() != *brk_label {
```
HalidOdat, line 43: Why do we do `.to_string()` here? It seems like an unnecessary string allocation.

jasonwilliams: Yeah this was annoying, `stmt_label` is a `Box<str>` and `brk_label` is a `&String`. 
Ideally `brk_label` should be a `Box<str>` but it was not possible to make that happen, i can't remember why i will have to look at it again to find out

HalidOdat: I don't we why we can't change `InterpreterState::Break(Option<String>)` to be `InterpreterState::Break(Option<Box<str>>)` this should also be more space efficient.

Razican: As far as I understand, `Box<str>` and `&String` can be compared by doing:
```rust
if &stm_label != brk_label.as_str() {
    todo!()
}
```

And the `&` in the `stm_label` might be optional, or it might require a `stm_label.as_ref()`.

jasonwilliams: > I don't we why we can't change InterpreterState::Break(Option<String>) to be InterpreterState::Break(Option<Box<str>>) this should also be more space efficient.

Yep thats what i originally started with

jasonwilliams: The problem starts here: https://github.com/boa-dev/boa/blob/master/boa/src/exec/break_node/mod.rs#L14
I can't change this to `Box<str>` because ownership of the string will be moved (and Break needs to keep ownership of the label). So at the very minimum it needs to be `&Box<str>` but this isn't easy because it means the whole of `Interpreter` and `Context` will need lifetimes added to it because it's holding a burrowed value.

jasonwilliams: > As far as I understand, `Box<str>` and `&String` can be compared by doing:
> 
> ```rust
> if &stm_label != brk_label.as_str() {
>     todo!()
> }
> ```
> 
> And the `&` in the `stm_label` might be optional, or it might require a `stm_label.as_ref()`.

That doesn't work
image

HalidOdat: > The problem starts here: [https://github.com/boa-dev/boa/blob/master/boa/src/exec/break_node/mod.rs#L14](https://github.com/boa-dev/boa/blob/master/boa/src/exec/break_node/mod.rs?rgh-link-date=2020-09-25T17%3A27%3A55Z#L14)
> I can't change this to `Box<str>` because ownership of the string will be moved (and Break needs to keep ownership of the label). So at the very minimum it needs to be `&Box<str>` but this isn't easy because it means the whole of `Interpreter` and `Context` will need lifetimes added to it because it's holding a burrowed value.
> 
> The reason it works fine now is because we're creating a new string everytime we run Break. Another option would be to wrap it in a Box instead

Just change it to `Box::from`:
```
            .set_current_state(InterpreterState::Break(self.label().map(Box::from)));
```
Box<str> implements From for &str

boa/src/syntax/ast/node/iteration.rs
jasonwilliams: This means the &str will need to be compared to the `Box<&str>` that break provides

tester/src/exec.rs
```diff
@@ -83,6 +83,7 @@ impl Test {
     /// Runs the test.
     pub(crate) fn run(&self, harness: &Harness) -> TestResult {
         // println!("Starting `{}`", self.name);
+        dbg!(&self.content);
```
HalidOdat, line 86:
```suggestion
```

we should remove the debug stuff

jasonwilliams: thanks that was committed by accident

HalidOdat (approved):

PR #550, opened 2020-07-05
Initial implementation of Map()

### LINKED ISSUE #401: implement Map()

joshwd36:
This Pull Request closes #401.

It changes the following:
 - Adds the Map object, backed by a HashMap.
 - Adds most of the associated methods.
 - Adds a display implementation for the map.

There are a few things that currently aren't implemented:
- [ ] Methods that return an iterator.
- [x] The map is not ordered by insertion. Currently this only affects the display output and the forEach method.
- [x] The Map() constructor can be called as a function to construct a map, rather than throwing an exception.
- [x] The data is stored as a HashMap in ObjectData. This makes cloning ObjectData quite expensive now, and also means actions such as adding data clone the HashMap. It should probably be placed in a wrapper containing an Rc<RefCell<_>>.
- [x] Adding the map to itself seems to work, but the display implementation I have overflows the stack when called.


joshwd36:
I'm not entirely sure what to do about the mutable key types lint. I think it should be OK because the values are cloned when they're placed in the HashMap and then again when they're removed, so it shouldn't be possible to modify that particular copy, but I'm probably missing something

joshwd36:
> 
> 
> I'm not entirely sure what to do about the mutable key types lint. I think it should be OK because the values are cloned when they're placed in the HashMap and then again when they're removed, so it shouldn't be possible to modify that particular copy, but I'm probably missing something

I've just done some additional testing and it seems the current implementation works. I tried creating an object and using it as the key in a map, then modifying the object and using it to get from the map, and it still got the value.

HalidOdat:
> The map is not ordered by insertion. Currently this only affects the display output and the forEach method.

About this I don't think `HashMap` is suficient we need something that keeps it order of insertion and can be indexed like an array, I think we need the [indexmap](https://crates.io/crates/indexmap) crate it provides the [`IndexMap`](https://docs.rs/indexmap/1.4.0/indexmap/map/struct.IndexMap.html) it seems perfect for the job and it also provides a `IndexSet` to which is needed for `Set` #400

HalidOdat:
> The data is stored as a HashMap in ObjectData. This makes cloning ObjectData quite expensive now, and also means actions such as adding data clone the HashMap. It should probably be placed in a wrapper containing an Rc<RefCell<_>>.

when we clone a `GcObject` which `Value` has (`Value::Object(GcObject)`), It does not do a deep clone of the object, only a reference to the object is cloned. 

The way we extracting the data is a bit awkward though.

joshwd36:
> 
> 
> > The map is not ordered by insertion. Currently this only affects the display output and the forEach method.
> 
> About this I don't think `HashMap` is sufficient we need something that keeps it order of insertion and can be indexed like an array, I think we need the [indexmap](https://crates.io/crates/indexmap) crate it provides the [`IndexMap`](https://docs.rs/indexmap/1.4.0/indexmap/map/struct.IndexMap.html) it seems perfect for the job and it also provides a `IndexSet` to which is needed for `Set` #400

That seems like it could work well. Unfortunately, there are a few places that might be a bit inefficient, but it would at least be correct. In order to maintain insertion order with removals, we'd have to use the [shift_remove](https://docs.rs/indexmap/1.4.0/indexmap/map/struct.IndexMap.html#method.shift_remove) function, which is O(n). Likewise, when inserting a duplicate key the original would have to be removed so that the new one is placed at the back, rather than swapped in. For now, though, this would be good, and I'll work on swapping over.
> 
> 
> > The data is stored as a HashMap in ObjectData. This makes cloning ObjectData quite expensive now, and also means actions such as adding data clone the HashMap. It should probably be placed in a wrapper containing an Rc<RefCell<_>>.
> 
> when we clone a `GcObject` which `Value` has (`Value::Object(GcObject)`), It does not do a deep clone of the object, only a reference to the object is cloned, in a way it is like an `Rc`
> 
> The way we extracting the data is a bit awkward though.
Ah, I missed how the GcObject works, that's good to hear. Just out of interest, I notice that other objects such as BigInt and String use Rc wrapper types. Is there a reason for this?

But yes, modifying the data does feel a bit awkward, as updating the data requires cloning the hashmap and replacing the original. Using a RefCell could work, as could instance methods taking a mutable reference, although that might be quite a large architectural change.

HalidOdat:
> Unfortunately, there are a few places that might be a bit inefficient, but it would at least be correct. In order to maintain insertion order with removals, we'd have to use the [shift_remove](https://docs.rs/indexmap/1.4.0/indexmap/map/struct.IndexMap.html#method.shift_remove) function, which is O(n).

Sadly yes. but that's what the spec requires. 

> likewise, when inserting a duplicate key the original would have to be removed so that the new one is placed at the back, rather than swapped in. For now, though, this would be good, and I'll work on swapping over.

for insertion it replaces or appends if there isn't a duplicate key, right? https://tc39.es/ecma262/#sec-map.prototype.set step 4.a.i `Set p.[[Value]] to value.`

> For now, though, this would be good, and I'll work on swapping over.

Yes. for now we can use `HashMap`

> Ah, I missed how the GcObject works, that's good to hear. Just out of interest, I notice that other objects such as BigInt and String use Rc wrapper types. Is there a reason for this?

It's because the `Value` primitives `Value::String`, `Value::BigInt`, `Value::Symbol` contain an `Rc` to make the `Value` trivially copyable so we don't have to clone a String or BigInt every time we clone object since they are immutable, this also shrank the size of `Value` from 40 bytes to 24 bytes, and it to also shrink the size of `ObjectData` and so we don't have to clone the String when we convert a string primitive to a string object. see #465 and #498 for more information.

HalidOdat:
> But yes, modifying the data does feel a bit awkward, as updating the data requires cloning the hashmap and replacing the original. Using a RefCell could work, as could instance methods taking a mutable reference, although that might be quite a large architectural change.

Yes the way we do it not is far from ideal.

joshwd36:
> 
> 
> > likewise, when inserting a duplicate key the original would have to be removed so that the new one is placed at the back, rather than swapped in. For now, though, this would be good, and I'll work on swapping over.
> 
> for insertion it replaces or appends if there isn't a duplicate key, right? https://tc39.es/ecma262/#sec-map.prototype.set step 4.a.i `Set p.[[Value]] to value.`
> 
Yes, sorry, you're completely right. That does simplify things then.
> 
> 
> > But yes, modifying the data does feel a bit awkward, as updating the data requires cloning the hashmap and replacing the original. Using a RefCell could work, as could instance methods taking a mutable reference, although that might be quite a large architectural change.
> 
> Yes the way we do it is far from ideal.
> 
> Edit: We can use `Rc<RefCell>` and in the future when we refactor `Object` we can change it :)

I don't think RefCell would work because Trace isn't implemented for it. GcCell works though. I also don't think Rc should be necessary, as there should only ever be one owner of the data.

boa/src/builtins/value/display.rs
```diff
@@ -102,6 +102,32 @@ pub(crate) fn log_string_from(x: &Value, print_internals: bool) -> String {
...
+                        return String::from("Map(0)");
+                    }
+
+                    let mappings = map
+                        .borrow()
+                        .iter()
+                        .map(|(key, value)| {
+                            let key = log_string_from(key, print_internals);
+                            let value = log_string_from(value, print_internals);
+                            format!("{} → {}", key, value)
```
HalidOdat, line 125: To stop the overflow we need to check if the value is an `Map` object, if it is we print object `Map(sizeOfMap)` and we don't expand it any more, like SpiderMonkey does it:
```js
let x = new Map();
x.set("y", x);
x
```
This prints `Map { y → Map(1) }`

joshwd36: I don't think this would work with all cases. For example
```javascript
let x = new Map();
let y = new Array();
y.push(x);
x.set("z", y);
x
```
will still overflow. Could we potentially fix this by only printing the top level items of collections, and have a boolean argument in log_string_from, like `log_string_from(x: &Value, print_internals: bool, top_level: bool) which is set to true when called outside of the method and set to false when called inside the method?

HalidOdat: > will still overflow. Could we potentially fix this by only printing the top level items of collections, and have a boolean argument in log_string_from, like `log_string_from(x: &Value, print_internals: bool, top_level: bool) which is set to true when called outside of the method and set to false when called inside the method?

Yes. would be a better solution for all cases printing something like:
```js
Map { y → Array(sizeOfArray) }
```

It is also nicer than having all the field of an array expanded. like:
```js
let x = new Map();
let y = new Array(1, 2, 3, 4, 5, 6, 7, 8);
x.set("z", y);
x.set("w", y);
x
```
this would be to long with the current implementation:
```js
Map { z → [1, 2, 3, 4, 5, 6, 7, 8], w  → [1, 2, 3, 4, 5, 6, 7, 8] }
```
instead it would be:
```js
Map { z → Array(8), w  → Array(8) }
```
This is mush cleaner :)

joshwd36: Great! I'll have a go at implementing that.

HalidOdat:
> The Map() constructor can be called as a function to construct a map, rather than throwing an exception.

For this we need to set this  to false for `Map` function: <https://github.com/boa-dev/boa/blob/master/boa/src/builtins/function/mod.rs#L109>

Maybe we should make <https://github.com/boa-dev/boa/blob/master/boa/src/builtins/function/mod.rs#L416-L467> accept another parameter or an enum with field `Constructable` `Callable` `ConstructableAndEnumerable` or maybe a `bitfield` and do `CONSTRUCTABLE | CALLABLE` and set depending on the argument the callable and constructable fields respectively like we do with constructable <https://github.com/boa-dev/boa/blob/master/boa/src/builtins/function/mod.rs#L429>

I think the bitfield would be much nicer and it would shrink the size of `Function` struct too.

HalidOdat:
> Methods that return an iterator.

About this, we don't have iterator anywhere, also we don't support well-known symbols (`Symbol.iterator`)

If you want we can leave this to be a separate PR, implementing `MapIterator` object. :)

boa/src/builtins/map/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+        )
+    }
+
+    /// Initialise the `Map` object on the global object.
+    #[inline]
+    pub(crate) fn init(global: &Value) -> (&str, Value) {
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
+
+        (Self::NAME, Self::create(global))
+    }
```
HalidOdat, line 327:
```suggestion
    /// Initialise the `Map` object on the global object.
    pub(crate) fn init(global: &Value) -> Value {
    	let _timer = BoaProfiler::global().start_event(Self::NAME, "init");

        // Create prototype
        let prototype = Value::new_object(Some(global));

        make_builtin_fn(Self::set, "set", &prototype, 2);
        make_builtin_fn(Self::delete, "delete", &prototype, 1);
        make_builtin_fn(Self::get, "get", &prototype, 1);
        make_builtin_fn(Self::clear, "clear", &prototype, 0);
        make_builtin_fn(Self::has, "has", &prototype, 1);
        make_builtin_fn(Self::for_each, "forEach", &prototype, 1);

        let map_object = make_constructor_fn(
            Self::NAME,
            Self::LENGTH,
            Self::make_map,
            global,
            prototype,
            true,
        );
        
        (Self::NAME, map_object)
    }
```
#547 Marged `create` into `init` because init was always empty and we did not need an extra function.

HalidOdat: Sorry, `init` should return a `(&str, Value)` :sweat_smile:

joshwd36:
> 
> 
> > Methods that return an iterator.
> 
> About this, we don't have iterator anywhere, also we don't support well-known symbols (`Symbol.iterator`)
> 
> If you want we can leave this to be a separate PR, implementing `MapIterator` object, if you want :)

That's probably best, it would probably pollute the PR quite a lot, and I'm not sure whether I'd be able to implement it, although I'll have a go

Razican:
I see many changes in the diff not directly related to this PR, why is that? Could we rebase?

joshwd36:
> 
> 
> I see many changes in the diff not directly related to this PR, why is that? Could we rebase?

I think this is due to how github is displaying the diffs, as described [here](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch). I changed the base branch away and back as described and it seems to be better now.

boa/src/builtins/object/mod.rs
```diff
@@ -67,6 +68,7 @@ pub struct Object {
 #[derive(Debug, Trace, Finalize, Clone)]
 pub enum ObjectData {
     Array,
+    Map(GcCell<OrderedMap<Value, Value>>),
```
HalidOdat, line 71: I don't think we need to `GcCell` since we expanded `get_map` in to it's methods, this would eliminate the checks from `GcCell`.

boa/src/builtins/map/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            if let Some(map) = object.as_map_ref() {
+                let mut map = map.borrow_mut();
+                map.insert(key, value);
+                map.len()
+            } else {
+                return Err(ctx.construct_type_error("'this' is not a Map"));
+            }
+        } else {
+            return Err(ctx.construct_type_error("'this' is not a Map"));
+        };
```
HalidOdat, line 78:
```suggestion
        let size = if let Value::Object(ref object) = this {
            let object = object.borrow_mut();
            if let Some(ref mut map) = object.as_map_ref() {
                map.insert(key, value);
                map.len()
            } else {
                return Err(ctx.construct_type_error("'this' is not a Map"));
            }
        } else {
            return Err(ctx.construct_type_error("'this' is not a Map"));
        };
```
Wouldn't this work without the `GcCell`?

joshwd36: Doesn't as_map_ref only return an immutable reference though? The GcCell allows us to use interior mutability to modify it without cloning the data and then replacing the objectdata, which is what it was doing previously

joshwd36: What we could do is move the GcCell into the OrderedMap so that the modification methods work on an immutable reference.

HalidOdat: > Doesn't as_map_ref only return an immutable reference though?

Yes. it does currently, but we can change it to take a muttable self and return a mutable map.

> The GcCell allows us to use interior mutability to modify it without cloning the data and then replacing the objectdata, which is what it was doing previously

The way the `Value` is setup `Value::Object` contains a `GcObject` this is just a wrapper for `Gc<GcCell<Object>>` so we can mutate value with immutable reference, so it already has `interior mutability`, this is why we can set a field to `this` with an immutable reference, like `set_size()` method

joshwd36: Ah yes, I didn't see that. I'll do that then

PR #553, opened 2020-07-08
Refactor Property Descriptor flags

### LINKED ISSUE #497: Refactoring `Property` flags
Make the flags instead of Option<bool> we combine all the flags into one `u8` and store and access the bits this should shrink the size of `Property`.

HalidOdat:
This Pull Request fixes/closes #497 .

This PR changes:
 - Added `Attribute` bitflags for defineing when the property is `WRITABLE`, `CONFIGURABLE` or `ENUMBERABLE`
 - changed `Option<bool>` for `writable`, `enumerable` and `configurable` flags to take one byte.
 - Added `data_descriptor` constructor for `Property`.
 - Added `has_*`, `set_*` and `*_or` methods for `property`
 - Made flag selection more ergonomic for example:

```rust
some_function(Attribute::READONLY | Attribute::ENUMBERABLE | Attribute::CONFIGURABLE);
```
over
```rust
some_function(Some(false), Some(true), Some(true));
```


boa/src/builtins/object/internal_methods.rs
```diff
@@ -279,43 +271,35 @@ impl Object {
...
                             d.set = v.set.clone();
                         }
-                        d.enumerable = v.enumerable;
-                        d.configurable = v.configurable;
+                        d.attribute = v.attribute;
                         d
                     })
             }
-            _ => Property::default(),
+            _ => unreachable!(""),
```
Razican, line 302: I would add a panic message here, just in case.

boa/src/builtins/property/mod.rs
```diff
@@ -53,71 +54,136 @@ pub struct Property {
...
+        value.is_string() || value.is_symbol()
     }
 
     /// Make a new property with the given value
     /// The difference between New and Default:
     ///
     /// New: zeros everything to make an empty object
     /// Default: Defaults according to the spec
+    #[inline]
     pub fn new() -> Self {
```
Razican, line 68: It seems this `new()` function is basically a default, right? We can derive Default and just not implement `new()` (or make `new()` just an alias).

boa/src/builtins/object/internal_methods.rs
HalidOdat: Yes forgot to add it :sweat_smile:

Issue #552, opened 2020-07-08
[builtins - Object]  Object.defineProperties() not implemented


croraf:
Also, as per: https://github.com/boa-dev/boa/pull/543#discussion_r450857804 , after this is done, `Object.create` should be improved to handle second argument (the object properties) along with some tests.

Razican:
Hi, this seems to be a new enhancement request, not something that is buggy. Could you please fill the template we have for this? There are no links to the spec or proper explanations on what this means.

croraf:
Which explaination you need on "Object.defineProperties() not implemented"?

Razican:
> It is not clear which of these should be selected in this case. Perhaps "Feature request" description should be improved.

Done in #554.

> Which explaination you need on "Object.defineProperties() not implemented"?

Do you have a link to the spec where this is explained? A code example that you would like to make work?

Also, you mention `Object.create`. What is this second parameter? Do you have a link to the spec? An example code? You mention we should add new tests. Which tests? why?

croraf:
> Do you have a link to the spec where this is explained? A code example that you would like to make work?
> 

 It should not be my job as a end-user and reported to suggest the way how to implement it, and provide suggestions where to find the information. This should be at the reporters discretion on add this or not. The issue and request is perfectly clear as is.


> Also, you mention `Object.create`. What is this second parameter? Do you have a link to the spec? An example code? You mention we should add new tests. Which tests? why?

Although I think this is also clear, I added a link to the related closed issue.

Razican:
> It should not be my job as a end-user and issue reported to suggest the way how to implement it, and provide suggestions where to find the information. The reporter should have a discretion to add this or not. The issue report is perfectly clear as is.

Sure, but understand that the more detail you give, the more possibilities you have for us to take this into account.

In any case, let me add some information for future developers.

This is defined in the spec [here](https://tc39.es/ecma262/#sec-object.defineproperties). This should be added to the `Object` global object in [this](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/object/mod.rs) file, similarly to the [`Object.create()`](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/object/mod.rs#L413-L441) method.

Then, this needs to be added as a property of the `Object` in the [end of the file](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/object/mod.rs#L554), and we could take the opportunity to fix the `Object.create()` method above by taking into account the second parameter, instead of panicking.

[...]

PR #554, opened 2020-07-08
Improved the description of the issue templates

Razican:
This PR changes the description of the feature request and custom issue templates.

It should avoid issues like #552, where a new ECMAScript feature is requested, but we have no links to the spec or no description of the feature, with no code example that would be affected.

### DIFF
--- a/.github/ISSUE_TEMPLATE/custom.md
+++ b/.github/ISSUE_TEMPLATE/custom.md
@@ -1,6 +1,6 @@
 ---
 name: Custom
-about: Open an issue in the repo that is neither a bug or a feature
+about: Open an issue in the repo that is neither a bug or a feature, such a new idea
 title: ''
 labels: ''
 assignees: ''
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,6 +1,6 @@
 ---
 name: "\U0001F680 Feature request"
-about: Suggest an idea for this project
+about: Suggest a new ECMAScript feature to be implemented, or a new capability of the engine.
 title: ''
 labels: enhancement
 assignees: ''


croraf:
I think reporting of incorrect, deficient or missing JS features should go under Bugs. 

The functionality of boa is to follow the latest ECMAScript, and any deviation is **a bug**. It is not up to the end user (reported) to know if the implementation is incorrect, deficient or completely missing from Boa implementation.

Lan2u:
> I think reporting of incorrect, deficient or missing JS features should all go under Bugs.
> 
> The functionality of boa is to follow the latest ECMAScript, and any deviation is **a bug**. It is not up to the end user (reporter) to know if the implementation is incorrect, deficient or completely missing from Boa implementation.
> 
> (ofc, if the reporter is aware of these details, it is good that they put them in the additional notes in the issue report)

Lan2u:
I think this is good especially since we will hopefully have test262 up and running soon and so, therefore, will have a good list of features which need implementing.

croraf:
Would anything that doesn't conform with the test262 be an unimplemented feature then? What will be a bug then?

HalidOdat:
> Would anything that doesn't conform with the test262 be an unimplemented feature then? What will be a bug then?

If something like `String.prototype.charAt` is not implemented then it's a feature request, but if it is implemented but does not work as expected than its a bug

croraf:
How would someone using boa and reporting a bug know if it is unimplemented or not working as expected?

Lan2u:
> Would anything that doesn't conform with the test262 be an unimplemented feature then? What will be a bug then?

Pretty much, I suspect we will have a significant amount of areas which will need addressing.

> How would someone using boa and reporting a bug know if it is unimplemented or not working as expected?
> 
> For example, I'm testing:
> 
> ```
> Object.defineProperties({}, {prop1: {value: '5'}})
> ```
> 
> And boa outputs me `undefined`. But I inspected the code, and I see `defineProperties` is not added on the `Object` builtin (or at least it seems to me it is not added).

I suppose if they aren't involved with the boa project then making that distinction falls to us to re-classify if required.

croraf:
And also the example of granularity:

`Object` builtin is implemented but `` is not, so is this a new feature, or wrong behavior of `Object`,
or further
`Object.create` builtin is implemented but handling of its second optional argument `additionalProperties` is not, so is this a new feature, or wrong behavior of `Object.create`.
You can go further if for example `Object.create` with second argument of `additionalProperties` does not have implemented the handling of some nested cases.

Razican:
> I think reporting of incorrect, deficient or missing JS features should all go under Bugs.

The project has started pretty recently, at least with the current level of development speed. For example 8 months ago, most of the standard was not implemented. If you want something in the standard to be prioritized for implementation, then it's a feature request.

If we add a bug for each unimplemented feature, then `await` is a bug, since it doesn't work, modules don't work, so that's another bug. `for..of` / `for..in` don't work, so two more bugs. Oh, and generators are not implemented either, so another bug.

We can't even run _test262_, so most things are feature requests IMHO. Once we have that running, and we iron out bugs in our current implementation, we will have a list of non-implemented things, and there we can more clearly see the difference between a new feature or a bug.

We don't always implement the full thing when implementing a new feature, and we leave some things unimplemented. Your own #543 PR is an example of this. We have the unimplemented second parameter in `create()`. The `create()` function was implemented in a "best effort" basis. I don't think that the implementation has a bug and you didn't comply with the spec properly. There was nothing, then there was something. Probably not all, but you did progress, and that's what we need. I don't believe that's a bug.

> And also the example of granularity

I think that if the `create` function is not implemented, then implementing it is a new feature. If a function receives one, two or three parameters, but we only have an implementation for the case with one parameter, implementing another one would be a new feature.

If we have an implementation for one parameter, but in an edge case, it doesn't work as expected, then it's a bug.

It's ok, until we don't have the *test262* ready, the creator of the issue doesn't have to know if it's a bug or a feature. We can tag it later and add a comment explaining it, but if the creator of the issue has some experience with the project, they can maybe know if it's a new feature or a bug.

PR #555, opened 2020-07-08
Implement Array.prototype.reduce

benjaminflin:
I implemented `Array.prototype.reduce` according to the [spec](https://tc39.es/ecma262/#sec-array.prototype.reduce). Let me know if I missed anything as this is my first time contributing.

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -127,9 +127,9 @@ impl Array {
             1 if args[0].is_integer() => {
                 length = i32::from(&args[0]);
                 // TODO: It should not create an array of undefineds, but an empty array ("holy" array in V8) with length `n`.
-                for n in 0..length {
-                    this.set_field(n.to_string(), Value::undefined());
-                }
+                // for n in 0..length {
+                //     this.set_field(n.to_string(), Value::undefined());
+                // }
             }
             1 if args[0].is_double() => {
                 return ctx.throw_range_error("invalid array length");
@@ -976,17 +976,13 @@ impl Array {
         let this = interpreter.to_object(this)?;
         let callback = match args.get(0) {
             Some(value) if value.is_function() => value,
-            _ => {
-                return interpreter.throw_type_error(
-                    "missing callback when calling function Array.prototype.reduce",
-                )
-            }
+            _ => return interpreter.throw_type_error("Reduce was called without a callback"),
         };
         let initial_value = args.get(1).cloned().unwrap_or_else(Value::undefined);
         let mut length = interpreter.to_length(&this.get_field("length"))?;
         if length == 0 && initial_value.is_undefined() {
             return interpreter
-                .throw_type_error("Array contains no elements and initial value is not provided");
+                .throw_type_error("Reduce was called on an empty array and with no initial value");
         }
         let mut k = 0;
         let mut accumulator = if initial_value.is_undefined() {
@@ -1000,7 +996,7 @@ impl Array {
             }
             if !k_present {
                 return interpreter.throw_type_error(
-                    "Array contains no elements and initial value is not provided",
+                    "Reduce was called on an empty array and with no initial value",
                 );
             }
             let result = this.get_field(k.to_string());
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -973,18 +973,20 @@ impl Array {
         args: &[Value],
         interpreter: &mut Interpreter,
     ) -> ResultValue {
-        if args.is_empty() {
-            return Err(Value::from(
-                "missing callback when calling function Array.prototype.reduce",
-            ));
-        }
-        let callback = args.get(0).cloned().unwrap_or_else(Value::undefined);
+        let this = interpreter.to_object(this)?;
+        let callback = match args.get(0) {
+            Some(value) if value.is_function() => value,
+            _ => {
+                return interpreter.throw_type_error(
+                    "missing callback when calling function Array.prototype.reduce",
+                )
+            }
+        };
         let initial_value = args.get(1).cloned().unwrap_or_else(Value::undefined);
         let mut length = i32::from(&this.get_field("length"));
         if length == 0 && initial_value.is_undefined() {
-            return Err(Value::from(
-                "Array contains no elements and initial value is not provided",
-            ));
+            return interpreter
+                .throw_type_error("Array contains no elements and initial value is not provided");
         }
 
         let mut k = 0;
@@ -998,9 +1000,9 @@ impl Array {
                 k += 1;
             }
             if !k_present {
-                return Err(Value::from(
+                return interpreter.throw_type_error(
                     "Array contains no elements and initial value is not provided",
-                ));
+                );
             }
             let result = this.get_field(k.to_string());
             k += 1;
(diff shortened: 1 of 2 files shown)


boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,77 @@ impl Array {
...
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-array.prototype.reduce
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce
+    pub(crate) fn reduce(
+        this: &Value,
+        args: &[Value],
+        interpreter: &mut Interpreter,
+    ) -> ResultValue {
+        if args.is_empty() {
```
HalidOdat, line 976: we should call `interperter.to_object(this)?` as described in spec step 1

boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,77 @@ impl Array {
...
+            return Err(Value::from(
+                "missing callback when calling function Array.prototype.reduce",
+            ));
+        }
+        let callback = args.get(0).cloned().unwrap_or_else(Value::undefined);
```
HalidOdat, line 981:
```suggestion
		let callback = match args.get(0) {
			Some(value) if value.is_funtion() => value,
			_ => return interperter.throw_type_error("missing callback when calling function Array.prototype.reduce"),
		};
```

```rust
Err(Value::from("missing callback when calling function Array.prototype.reduce"));
```
This throws a string not a `TypeError`, to throw a we use the `Interpreter::throw_type_error(msg)` also if callback is undefined the a `TypeError` should be thrown

boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,77 @@ impl Array {
...
+        let initial_value = args.get(1).cloned().unwrap_or_else(Value::undefined);
+        let mut length = i32::from(&this.get_field("length"));
+        if length == 0 && initial_value.is_undefined() {
+            return Err(Value::from(
+                "Array contains no elements and initial value is not provided",
+            ));
```
HalidOdat, line 987:
```suggestion
            return interpreter.throw_type_error(
                "Array contains no elements and initial value is not provided",
            );
```
Here we should throw `TypeError` with `.throw_type_error()`

boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,77 @@ impl Array {
...
+                if this.has_field(&k.to_string()) {
+                    k_present = true;
+                    break;
+                }
+                k += 1;
+            }
+            if !k_present {
+                return Err(Value::from(
+                    "Array contains no elements and initial value is not provided",
+                ));
```
HalidOdat, line 1003:
```suggestion
                return interpreter.throw_type_error(
                    "Array contains no elements and initial value is not provided",
                );
```
Here too.

boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,77 @@ impl Array {
...
+                let arguments = [
+                    accumulator,
+                    this.get_field(k.to_string()),
+                    Value::integer(k),
+                    this.clone(),
+                ];
+                match interpreter.call(&callback, &Value::undefined(), &arguments) {
+                    Ok(value) => accumulator = value,
+                    Err(e) => return Err(e),
+                }
```
HalidOdat, line 1022:
```suggestion
                accumulator = interpreter.call(&callback, &Value::undefined(), &arguments)?;
```
It's cleaner if use the try operator `?`

boa/src/builtins/array/mod.rs
benjaminflin:
```suggestion
        interpreter.to_object(this)?;
        if args.is_empty() {
```
Replying to attach a suggestion

boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,76 @@ impl Array {
...
+        let callback = match args.get(0) {
+            Some(value) if value.is_function() => value,
+            _ => {
+                return interpreter.throw_type_error(
+                    "missing callback when calling function Array.prototype.reduce",
+                )
+            }
+        };
+        let initial_value = args.get(1).cloned().unwrap_or_else(Value::undefined);
+        let mut length = i32::from(&this.get_field("length"));
```
HalidOdat, line 986:
```suggestion
        let mut length = interpereter.to_length(&this.get_field("length"))?;
```
We should use `.to_length()` method instead of `i32::from` because it is not spec compliant and we are in the process of removing it. :)

boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,76 @@ impl Array {
...
+            if this.has_field(&k.to_string()) {
+                let arguments = [
+                    accumulator,
+                    this.get_field(k.to_string()),
+                    Value::integer(k),
+                    this.clone(),
+                ];
+                accumulator = interpreter.call(&callback, &Value::undefined(), &arguments)?;
+                // reduce may change the length of the array
+                length = min(length, i32::from(&this.get_field("length")));
```
HalidOdat, line 1023:
```suggestion
                // reduce may change the length of the array
                length = min(length, interpreter.to_length(&this.get_field("length"))?);
```
also, I understand the updating of length, but I don't see this in the spec. Am I missing something?

benjaminflin: For some reason resizing the array by setting the length does not cause array accesses beyond the bounds of the array to be undefined. This means `has_field` can still return `true` for an index beyond `arr.length - 1`. IMO, this is probably a bug.

```javascript
let arr = [1, 2, 3, 4, 5];
arr.length = 3;
console.log(arr[4]); // should be undefined, but instead is 5 
```

This means that for now, I'll have to manually update the length. Even if array accesses worked as intended, it might be a good idea to keep track of length to prevent unnecessary iteration.

HalidOdat: I see. I believe that it a bug. Is because the `internal object methods` are not set right, see https://tc39.es/ecma262/#sec-array-exotic-objects-defineownproperty-p-desc

> Even if array accesses worked as intended, it might be a good idea to keep track of length to prevent unnecessary iteration.

In that case we should leave it as is, a more descriptive comment would be nice to have :)

HalidOdat: Also here we should use the `.to_lenght()` method:
```
length = min(length, interpreter.to_length(&this.get_field("length"))?);
```

boa/src/builtins/array/mod.rs
```diff
@@ -957,6 +957,77 @@ impl Array {
...
+            while k < length {
+                if this.has_field(&k.to_string()) {
+                    k_present = true;
+                    break;
+                }
+                k += 1;
+            }
+            if !k_present {
+                return interpreter.throw_type_error(
+                    "Array contains no elements and initial value is not provided",
```
Razican, line 1003: I think this error message is not correct for this case, right?

benjaminflin: I'm pretty sure handles the case when an array of a given size > 0 has no defined elements. This should be indistinguishable from calling `reduce` on an array of length 0, right? E.g.,

```javascript
var arr = [0,1];
delete arr[0];
delete arr[1];
console.log(arr.length); // 2
array.reduce((acc, x) => acc + x); // TypeError: Array contains no elements and initial value is not provided
```
should be the same as

```javascript
var arr = [];
array.reduce((acc, x) => acc + x); // TypeError: Array contains no elements and initial value is not provided
```
I confirmed that chrome gives the same error message in either case. 

I'll be sure to add these as well as other error cases to the tests.

Razican: True :) yes, error messages should be as descriptive as possible. I love Rust's error messages, we can probably learn from them.

PR #556, opened 2020-07-08
Added benchmark for goal symbol switching

Razican:
As suggested in https://github.com/boa-dev/boa/pull/486#issuecomment-655682389, this adds a new benchmark to check parsing of code with multiple goal symbol switching.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,9 +144,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.56"
+version = "1.0.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Lan2u:
I believe the mechanism that this benchmark relies on for lexing pre-goal-symbols is based on trial and error so hopefully, once new_lexer is ready (which has much less trial/error and back/forward) the performance should increase.

Razican:
> I believe the mechanism that this benchmark relies on for lexing pre-goal-symbols is based on trial and error so hopefully, once new_lexer is ready (which has much less trial/error and back/forward) the performance should increase.

We don't currently switch goal symbols, since we assume we have no context. I think the new lexer will be a bit slower, since we need to tell the lexer what to lex. On the other hand, as you say, once the goal symbol is set, the options get reduced and there is no need to backtrack, so it might be faster. But the current lexer just fails and decides to lex whatever it finds first.

Issue #288, opened 2020-03-30
Streaming Parsing (from Lexer to Parser)


jasonwilliams:
The way Boa works right now is that the Lexer will run till completion. Then once an array of tokens is filled this is sent over the parser.

However the parser doesn't need to wait idle whilst the Lexer is running, the parser can begin working through the tokens as they come through.. 

Completely inspired from how Go parses its templates, Rob talks about it in more detail here:
https://youtu.be/HxaD_trXwRE?t=960

### Parser Cursor

Does the cursor need to change? 
Can it work in its current state? @Razican is working on the updated cursor right now.
https://github.com/jasonwilliams/boa/pull/287

### Can Rust's Async/Await help here?

Maybe, i'm not sure.

### Diagram

![Blank Diagram (1)](https://user-images.githubusercontent.com/936006/77943456-26927180-72b5-11ea-8f9b-fd25186d4882.png)

@Razican @HalidOdat thoughts?

Razican:
About the cursor, it might need to return a `Result` on each call, but for the rest, the external API could stay without much change.

Razican:
Something we might need would be to add an `EOF` token to the stream, if we don't want to keep the parser waiting for more tokens indefinitely. Or, on the other hand, we might want exactly that.

Doing this in multiple threads could be done by lexing in one thread, parsing in another and executing in another. Doing concurrent parser might prove to be more difficult, though.

jasonwilliams:
The lexer can close the channel once it's finished, this will signal to the parser there is nothing more to come down the pipe.

jasonwilliams:
It might be due to the way parsing and lexing are related (goal symbols) this may not be possible.

Related:
https://github.com/jasonwilliams/boa/issues/294

Razican:
> It might be due to the way parsing and lexing are related (goal symbols) this may not be possible.

I think that what can be done is for the parser to request lexing of new tokens to the lexer via calls to `next()` in the cursor.

jasonwilliams:
I may close this seeing as we can't do it that way dur to the lexer needing context

Razican:
> I may close this seeing as we can't do it that way dur to the lexer needing context

Actually, I think this is now more important than ever!

Instead of "streaming" we could do "lazy" lexing, and making the parser iterate over it, while changing context as needed. @maciejhirsz proposed to use [Logos](https://crates.io/crates/logos) for this, and I think it makes sense. Not sure if there was any progress in that front.

[...]

PR #559, opened 2020-07-09
New lexer

Lan2u:
This is the same as the previous branch but based on a branch on the boa repo ( we are troubleshooting why there is no benchmark comment).

Closes:
#294 
#456 
#288 (at least with reference to using streams but not multithreaded)
#588


Razican:
> Benchmark for [65fe426](https://github.com/boa-dev/boa/commit/65fe426d81fcace74e480aae77f1fe4c082b629f)

Hmm it seems that didn't improve the benchmarks :(

HalidOdat:
> Hmm it seems that didn't improve the benchmarks :(

We need to find where the bottleneck is, we might be able do this by using [flamegraph](https://crates.io/crates/flamegraph) since it gives a more verbose output (also including `String::clone` for example) and using it to parse a large file directly (without execution).

Razican:
This is an example of the parsing in this branch:

<details>
<summary>Chrome flamegraph</summary>


Screenshot from 2020-07-13 17-53-53

</details>

I'm not sure if it can be seen properly, but there are 4 things in the lexer that take too much time:
 - Identifiers
 - Numeric Literals
 - Regex Literals
 - String Literals

Apart from that, calling `peek()` and `next()` in the parser cursor also takes a lot of time. I think it's related to the heap usage with the `VecDeque`.

I've tried to simplify those lexing parts, but I'm not getting anywhere.

Lan2u:
> This is an example of the parsing in this branch:
> 
> Chrome flamegraph
> Screenshot from 2020-07-13 17-53-53
> 
> I'm not sure if it can be seen properly, but there are 4 things in the lexer that take too much time:
> 
> * Identifiers
> * Numeric Literals
> * Regex Literals
> * String Literals
> 
> Apart from that, calling `peek()` and `next()` in the parser cursor also takes a lot of time. I think it's related to the heap usage with the `VecDeque`.
> 
> I've tried to simplify those lexing parts, but I'm not getting anywhere.

I'm looking into removing VecDeque

Razican:
One thing I found out is that the parser cursor no longer skips line terminators by default, which is breaking the execution of `assert.js` in test262.

Lan2u:
> One thing I found out is that the parser cursor no longer skips line terminators by default, which is breaking the execution of `assert.js` in test262.

Do u have an example of the output?

Razican:
> > One thing I found out is that the parser cursor no longer skips line terminators by default, which is breaking the execution of `assert.js` in test262.
> 
> Do u have an example of the output?

Yep:
```
Parsing Error: unexpected token 'line terminator', primary expression at line 70, col 76
```
The file is this one: https://github.com/tc39/test262/blob/master/harness/assert.js

If you check how `peek()` and `next()` are currently implemented in the master branch, they skip line terminators by default: https://github.com/boa-dev/boa/blob/master/boa/src/syntax/parser/cursor.rs#L66

This is the expected behaviour, since line terminators are supposed to be treated as whitespace except in semicolon insertion or where line terminators are not allowed.

Lan2u:
> > > One thing I found out is that the parser cursor no longer skips line terminators by default, which is breaking the execution of `assert.js` in test262.
> > 
> > 
> > Do u have an example of the output?
> 
> Yep:
> 
> ```
> Parsing Error: unexpected token 'line terminator', primary expression at line 70, col 76
> ```
> 
> The file is this one: https://github.com/tc39/test262/blob/master/harness/assert.js
> 
> If you check how `peek()` and `next()` are currently implemented in the master branch, they skip line terminators by default: https://github.com/boa-dev/boa/blob/master/boa/src/syntax/parser/cursor.rs#L66
> 
> This is the expected behaviour, since line terminators are supposed to be treated as whitespace except in semicolon insertion or where line terminators are not allowed.

Yeah so currently in this branch when parsing you need to explicitly 'skip' the line terminators (and therefore not skip them when you need to check for their presence). The above error can be fixed by adding those skip checks where appropriate into primary expression.

Razican:
> Yeah so currently in this branch when parsing you need to explicitly 'skip' the line terminators (and therefore not skip them when you need to check for their presence). The above error can be fixed by adding those skip checks where appropriate into primary expression.

This should not work like this. Parsers should not care about line terminators, if they are not explicitly checking that there isn't any.

Lan2u:
> > Yeah so currently in this branch when parsing you need to explicitly 'skip' the line terminators (and therefore not skip them when you need to check for their presence). The above error can be fixed by adding those skip checks where appropriate into primary expression.
> 
> This should not work like this. Parsers should not care about line terminators, if they are not explicitly checking that there isn't any.

It seems to be that in lots of places they do care, that said I'll have a look at putting the auto skipping back in.

Razican:
About the peeking buffer, I think it would be best to abstract that to a different crate, and probably use a ring buffer, like `ArrayVecDeque`or so. Something like merging [`arrayvec`](https://crates.io/crates/arrayvec) with `VecDeque`. We could even contribute to that crate for this.

Lan2u:
> About the peeking buffer, I think it would be best to abstract that to a different crate, and probably use a ring buffer, like `ArrayVecDeque`or so. Something like merging [`arrayvec`](https://crates.io/crates/arrayvec) with `VecDeque`. We could even contribute to that crate for this.

I did it just using a fixed sized array (as a ring buffer) - unless I'm mistaken this is probably the fastest mechanism.

Lan2u:
> > Yeah so currently in this branch when parsing you need to explicitly 'skip' the line terminators (and therefore not skip them when you need to check for their presence). The above error can be fixed by adding those skip checks where appropriate into primary expression.
> 
> This should not work like this. Parsers should not care about line terminators, if they are not explicitly checking that there isn't any.

I've been working on this for awhile but a problem which I haven't been able to solve yet is what happens if we call peek saying we don't care about line terminators and then start parsing a new statement which does (it won't find the line terminator because the previous peek call discarded it).

Lan2u:
> > > Yeah so currently in this branch when parsing you need to explicitly 'skip' the line terminators (and therefore not skip them when you need to check for their presence). The above error can be fixed by adding those skip checks where appropriate into primary expression.
> > 
> > 
> > This should not work like this. Parsers should not care about line terminators, if they are not explicitly checking that there isn't any.
> 
> I've been working on this for awhile but a problem which I haven't been able to solve yet is what happens if we call peek saying we don't care about line terminators and then start parsing a new statement which does (it won't find the line terminator because the previous peek call discarded it).

This lead me to try with peek to by default return line terminators but this has basically resulted in me coming right back round to the start.

Lan2u:
> This is an example of the parsing in this branch:
> 
> Chrome flamegraph
> I'm not sure if it can be seen properly, but there are 4 things in the lexer that take too much time:
> 
> * Identifiers
> * Numeric Literals
> * Regex Literals
> * String Literals
> 
> Apart from that, calling `peek()` and `next()` in the parser cursor also takes a lot of time. I think it's related to the heap usage with the `VecDeque`.
> 
> I've tried to simplify those lexing parts, but I'm not getting anywhere.

When you ran flamegraph what javascript input did you use?

Razican:
> When you ran flamegraph what javascript input did you use? 

I used the for loop and goal symbols benchmarks.

Lan2u:
> > When you ran flamegraph what javascript input did you use?
> 
> I used the for loop and goal symbols benchmarks.

Do you have any details or a reference for how you ran it? I've been struggling to get good results.

Razican:
> > > When you ran flamegraph what javascript input did you use?
> > 
> > I used the for loop and goal symbols benchmarks.
> 
> Do you have any details or a reference for how you ran it? I've been struggling to get good results.

I followed the profiling guide of the repo. It's not a flamegraph per se, it's using measureme with the boa profiler. Check the profiling guide (linked from the readme) to see if you can make it work.

Lan2u:
I've found that now that the comma operator has been implemented the code 
"
(b++, b);
"
cannot be parsed by boa due to the increment operator.

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        while self.buf_size <= skip_n {
+            // Need to keep peeking more values.
+            self.peeked[(self.back_index + self.buf_size) % PEEK_BUF_SIZE] =
+                self.lexer.next(skip_line_terminators)?;
+
+            self.buf_size += 1;
+        }
+
+        // Have now peeked ahead the right number of spaces so can fetch the value directly.
+        Ok(self.peeked[(self.back_index + skip_n) % PEEK_BUF_SIZE].clone())
```
Razican, line 188: I wouldn't `clone()` here. A `peek()` should return a reference to the token if the token is not `Copy` (which will probably be with the future string interner).

HalidOdat:
There is a bug related to binary operators like `+`, `||` etc where they don't skip line terminators and as result of that the following code fails:
```js
"Hello, " +
"world"
```
or 
```js
"Hello, "
+
"world"
```
but this works fine:
```js
"Hello, " + "world"
```
This is the cause for the stack overflow in `test262` branch

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn from(lexer: Lexer<R>) -> Self {
+        Self {
+            lexer,
+            peeked: [
+                None::<Token>,
+                None::<Token>,
+                None::<Token>,
+                None::<Token>,
+                None::<Token>,
+            ],
```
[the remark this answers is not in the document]
Lan2u: This doesn't work as Token is not copy

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+where
+    R: Read,
+{
+    #[inline(always)]
+    fn from(reader: R) -> Self {
+        Lexer::new(reader).into()
+    }
+}
+
+impl<R> BufferedLexer<R>
```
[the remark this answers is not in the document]
Lan2u: I like the idea of reducing multiple line terminators down to just a single one - this would simplify things alot.

boa/src/syntax/parser/cursor/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+pub(super) struct Cursor<R> {
+    buffered_lexer: BufferedLexer<R>,
+}
+
+impl<R> Cursor<R>
+where
+    R: Read,
+{
+    /// Creates a new cursor.
+    #[inline(always)]
```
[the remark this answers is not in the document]
Lan2u: Ah right, I didn't know this - will correct

boa/src/syntax/parser/cursor/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    #[inline(always)]
+    pub(super) fn lex_regex(&mut self, start: Position) -> Result<Token, ParseError> {
+        self.buffered_lexer.lex_regex(start)
+    }
+
+    #[inline(always)]
+    pub(super) fn next(
+        &mut self,
+        skip_line_terminators: bool,
```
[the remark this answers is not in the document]
Lan2u: Currently, in the majority of cases, we do not skip line terminators - but as you say in most cases we should be.

The problem is described above but basically there is an issue that if we peek(true) then we delete a line terminator which might be needed later. I think (but I haven't checked yet though) that your suggestion above (collapse line terminators and then look at index+1) will solve this but I need to check as it might lead to the possibility where a line terminator is left 'stuck' in the buffer longer than it should which later causes a problem.

boa/src/syntax/parser/cursor/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub(super) fn peek(
+        &mut self,
+        skip_n: usize,
+        skip_line_terminators: bool,
+    ) -> Result<Option<Token>, ParseError> {
+        self.buffered_lexer.peek(skip_n, skip_line_terminators)
+    }
+
+    #[inline(always)]
+    pub(super) fn push_back(&mut self, token: Token) {
```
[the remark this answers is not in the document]
Lan2u: #593 but I think there might be a better way

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
Lan2u: Fair point, 

> There is a bug related to binary operators like `+`, `||` etc where they don't skip line terminators and as result of that the following code fails:
> 
> ```js
> "Hello, " +
> "world"
> ```
> 
> or
> 
> ```js
> "Hello, "
> +
> "world"
> ```
> 
> With the error message:
> 
> ```
> Parsing Error: unexpected token 'line terminator', primary expression at line 1, col 11
> ```
> 
> but this works fine:
> 
> ```js
> "Hello, " + "world"
> ```
> 
> This is the cause for the stack overflow in `test262` branch

This should be a fairly easy fix (I probably shouldn't say that...)

Lan2u:
> There is a bug related to binary operators like `+`, `||` etc where they don't skip line terminators and as result of that the following code fails:
> 
> ```js
> "Hello, " +
> "world"
> ```
> 
> or
> 
> ```js
> "Hello, "
> +
> "world"
> ```
> 
> With the error message:
> 
> ```
> Parsing Error: unexpected token 'line terminator', primary expression at line 1, col 11
> ```
> 
> but this works fine:
> 
> ```js
> "Hello, " + "world"
> ```
> 
> This is the cause for the stack overflow in `test262` branch

This should be a fairly easy fix - I'll take a look.

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// The maximum number of tokens which can be peeked ahead.
+const MAX_PEEK_SKIP: usize = 3;
+
+#[derive(Debug)]
+pub(super) struct BufferedLexer<R> {
+    lexer: Lexer<R>,
+    peeked: [Option<Token>; PEEK_BUF_SIZE],
+    buf_size: usize,
+    back_index: usize,
```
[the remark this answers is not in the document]
Lan2u: My reasoning was that always working forwards like this simplifies the code as you don't need to deal with going backwards from 0.

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+            self.buf_size += 1;
+        }
+
+        // Have now peeked ahead the right number of spaces so can fetch the value directly.
+        Ok(self.peeked[(self.back_index + skip_n) % PEEK_BUF_SIZE].clone())
+    }
+
+    /// Takes the given token and pushes it back onto the parser token queue.
+    ///
+    /// Note: it pushes it at the the front so the token will be returned on next .peek().
```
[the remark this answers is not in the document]
Lan2u: Push_front() definitely makes more sense now that you say it, I originally called it push_back because it was 'pushing back onto the buffer'.

Razican:
> This should be a fairly easy fix - I'll take a look.

This should be fixed in #632.

Razican:
I want to check the assembly code being generated in the ring buffer. We might be getting bound checks due to indexing. If that's so, we will need to use some unsafe code, and some extra tests.

boa/src/syntax/parser/tests.rs
```diff
@@ -12,7 +12,6 @@ use crate::syntax::ast::{
 
 /// Checks that the given JavaScript string gives the expected expression.
 #[allow(clippy::unwrap_used)]
-#[track_caller]
```
Razican: This is ok for now, but we want this as soon as Rust 1.46 is released (27th August). It helps a lot when debugging tests.

Lan2u: Is it coming in 1.46? That is good

Lan2u:
> I want to check the assembly code being generated in the ring buffer. We might be getting bound checks due to indexing. If that's so, we will need to use some unsafe code, and some extra tests.

Aren't bound checks a good thing? Or do we want to remove them for performance?

Razican:
> > I want to check the assembly code being generated in the ring buffer. We might be getting bound checks due to indexing. If that's so, we will need to use some unsafe code, and some extra tests.
> 
> Aren't bound checks a good thing? Or do we want to remove them for performance?

The thing is, we have an index that should always be in bounds in the peeked token array. We cannot force that with a `usize`, but we can if we create something like an `Index{inner: usize}` that is private and can only be incremented/decremented using its public methods.

We add a bunch of tests there, and we no longer need bounds checking. This muight improve performance.

On the other hand, the Rust compiler might already be able to prove that the index never goes out of bounds. If that's so, it will have already removed the bounds checking, so this might not be needed.

Razican:
I think this is ready for review. In the future we can improve it, but I'm happy with this interface.

Lan2u:
I'm going to give stuff a last once over but I'd be keen to get this merged soon

Razican:
Note that this is 257 commits, so when merging, we must squash xD

boa/src/lib.rs
```diff
@@ -89,7 +89,8 @@ pub fn forward_val(engine: &mut Interpreter, src: &str) -> Result<Value> {
         Ok(expr) => expr.run(engine),
         Err(e) => {
             eprintln!("{}", e);
-            std::process::exit(1);
+            panic!(); // Panic instead of exit means that tests will continue.
+                      // std::process::exit(1);
```
Razican, line 93: I don't think we need to explain this.
```suggestion
            panic!();
```

Razican (commented):
Something I have noticed is that some files have CRLF line endings, which is giving nasty diffs. When editing, make sure to use an editor that respects the `.editorConfig` file, or set line endings to LF. I will try to fix this in a commit.

boa/src/lib.rs
Lan2u: This is left over from when I did this before we added it to master hence the extra comment

Razican:
The benchmark errors are because we don't switch context on lexing benchmarks. This is normal, only-lexing has no meaning anymore with this PR, so I think we should just remove the lexing benchmarks. What do you think?

boa/src/syntax/parser/statement/iteration/while_statement.rs
```diff
@@ -44,12 +46,17 @@ impl WhileStatement {
...
+    R: Read,
+{
     type Output = WhileLoop;
 
-    fn parse(self, cursor: &mut Cursor<'_>) -> Result<Self::Output, ParseError> {
+    fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
         let _timer = BoaProfiler::global().start_event("WhileStatement", "Parsing");
         cursor.expect(Keyword::While, "while statement")?;
+
+        // Line terminators can exist between a While and the condition.
```
Razican, line 59:
```suggestion
```

boa/src/syntax/parser/statement/if_stm/mod.rs
```diff
@@ -62,9 +68,11 @@ impl TokenParser for IfStatement {
...
             Statement::new(self.allow_yield, self.allow_await, self.allow_return).parse(cursor)?;
 
-        let else_stm = match cursor.peek(0) {
-            Some(else_tok) if else_tok.kind == TokenKind::Keyword(Keyword::Else) => {
-                cursor.next();
+        let else_tok = cursor.peek(0)?;
+
+        let else_stm = match else_tok {
+            Some(_) if else_tok.unwrap().kind() == &TokenKind::Keyword(Keyword::Else) => {
+                cursor.next()?.expect("Else token vanished");
```
Razican, line 75:
```suggestion
                cursor.next()?.expect("else token vanished");
```

Lan2u:
> The benchmark errors are because we don't switch context on lexing benchmarks. This is normal, only-lexing has no meaning anymore with this PR, so I think we should just remove the lexing benchmarks. What do you think?

No point keeping them now that parser/lexer are combined.

HalidOdat (approved):
This looks awesome!!

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,315 @@
...
+///  - [ECMAScript Specification][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-literals-numeric-literals
+fn check_after_numeric_literal<R>(cursor: &mut Cursor<R>) -> Result<(), Error>
+where
+    R: Read,
+{
+    let pred = |ch: char| ch.is_ascii_alphabetic() || ch == '$' || ch == '_' || ch.is_ascii_digit();
+    if cursor.next_is_pred(&pred)? {
+        Err(Error::syntax("NumericLiteral token must not be followed by IdentifierStart nor DecimalDigit characters"))
```
Razican, line 120: I think this error is a bit hard to understand by users.

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,315 @@
+use super::{Cursor, Error, TokenKind, Tokenizer};
```
Razican, line 1: I think we should add module-level documentation everywhere.

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+                mem::replace(&mut self.peeked[self.read_index], None)
+            } else {
+                // We only store 1 contiguous line terminator, so if the one at `self.read_index`
+                // was a line terminator, we know that the next won't be one.
+                self.read_index = (self.read_index + 1) % PEEK_BUF_SIZE;
+                if self.read_index == self.write_index {
+                    self.fill()?;
+                }
+
+                mem::replace(&mut self.peeked[self.read_index], None)
```
Razican, line 158: I think we should replace these by `take()`. We don't have benchmarks to decide otherwise, and it would be more clear for other developers.

boa/src/syntax/parser/cursor/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.buffered_lexer.lex_regex(start)
+    }
+
+    #[inline]
+    pub(super) fn next(&mut self) -> Result<Option<Token>, ParseError> {
+        self.buffered_lexer.next(true)
+    }
+
+    #[inline]
+    pub(super) fn peek(&mut self, skip_n: usize) -> Result<Option<&Token>, ParseError> {
```
Razican, line 55: These need documentation. Actually, it makes more sense to have it here than on the buffer, since this is the public interface. I wouldn't remove it from there, though.

boa/src/syntax/lexer/operator.rs
```diff
@@ -0,0 +1,163 @@
...
+            Some('=') => {
+                $cursor.next_char()?.expect("= token vanished");
+                $cursor.next_column();
+                $assign_op
+            }
+            Some(_) => $op,
+        }
+    });
+    ($cursor:ident, $assign_op:expr, $op:expr, {$($case:pat => $block:expr), +}) => ({
+        // let punc = $cursor.peek().ok_or_else(|| Error::syntax("could not preview next value"))?;
```
Razican, line 27: Why is this commented out?

boa/src/syntax/lexer/operator.rs
```diff
@@ -0,0 +1,163 @@
...
+};
+use std::io::Read;
+
+/// `vop` tests the next token to see if we're on an assign operation of just a plain binary operation.
+///
+/// If the next value is not an assignment operation it will pattern match  the provided values and return the corresponding token.
+macro_rules! vop {
+    ($cursor:ident, $assign_op:expr, $op:expr) => ({
+        match $cursor.peek()? {
+            None => Err(Error::syntax("Abrupt end: could not preview next value as part of operator")),
```
Razican, line 17:
```suggestion
            None => Err(Error::syntax("abrupt end - could not preview next value as part of the operator")),
```

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,315 @@
...
+                            Span::new(start_pos, cursor.pos()),
+                        ));
+                    }
+                    ch => {
+                        if ch.is_digit(8) {
+                            // LegacyOctalIntegerLiteral
+                            if self.strict_mode {
+                                // LegacyOctalIntegerLiteral is forbidden with strict mode true.
+                                return Err(Error::syntax(
+                                    "Implicit octal literals are not allowed in strict mode.",
```
Razican, line 182:
```suggestion
                                    "implicit octal literals are not allowed in strict mode",
```

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,315 @@
...
+
+                                kind = NumericKind::Integer(8);
+                            }
+                        } else if ch.is_digit(10) {
+                            // Indicates a numerical digit comes after then 0 but it isn't an octal digit
+                            // so therefore this must be a number with an unneeded leading 0. This is
+                            // forbidden in strict mode.
+                            if self.strict_mode {
+                                return Err(Error::syntax(
+                                    "Leading 0's are not allowed in strict mode.",
```
Razican, line 198:
```suggestion
                                    "leading 0's are not allowed in strict mode",
```

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,315 @@
...
+                }
+            },
+            NumericKind::Integer(base) => {
+                if let Ok(num) = i32::from_str_radix(&buf, base) {
+                    Numeric::Integer(num)
+                } else {
+                    let b = f64::from(base);
+                    let mut result = 0.0_f64;
+                    for c in buf.chars() {
+                        let digit = f64::from(c.to_digit(base).expect("Couldn't parse digit after already checking validity"));
```
Razican, line 302:
```suggestion
                        let digit = f64::from(c.to_digit(base).expect("could not parse digit after already checking validity"));
```

boa/src/syntax/lexer/number.rs
```diff
@@ -0,0 +1,315 @@
...
+///
+/// More information:
+///  - [ECMAScript Specification][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-literals-numeric-literals
+fn check_after_numeric_literal<R>(cursor: &mut Cursor<R>) -> Result<(), Error>
+where
+    R: Read,
+{
+    let pred = |ch: char| ch.is_ascii_alphabetic() || ch == '$' || ch == '_' || ch.is_ascii_digit();
```
Razican, line 118:
```suggestion
    let pred = |ch: char| ch.is_ascii_alphanumeric() || ch == '$' || ch == '_';
```

boa/src/syntax/lexer/operator.rs
```diff
@@ -0,0 +1,163 @@
...
+                cursor,
+                start_pos,
+                vop!(cursor, Ok(Punctuator::StrictNotEq), Ok(Punctuator::NotEq)),
+                Ok(Punctuator::Not)
+            ),
+            '~' => Ok(Token::new(
+                Punctuator::Neg.into(),
+                Span::new(start_pos, cursor.pos()),
+            )),
+            _ => unimplemented!(),
```
Razican, line 160:
```suggestion
            op => unimplemented!("operator {}", op),
```

boa/src/syntax/lexer/mod.rs
```diff
@@ -3,869 +3,261 @@
...
+    operator::Operator,
+    regex::RegexLiteral,
+    spread::SpreadLiteral,
+    string::StringLiteral,
+    template::TemplateLiteral,
 };
+use crate::syntax::ast::{Punctuator, Span};
+pub use crate::{profiler::BoaProfiler, syntax::ast::Position};
+pub use error::Error;
+pub use token::{Token, TokenKind};
```
Razican, line 36:
```suggestion
```

boa/src/syntax/lexer/mod.rs
```diff
@@ -3,869 +3,261 @@
...
+    /// Sets the goal symbol for the lexer.
+    #[inline]
+    pub(crate) fn set_goal(&mut self, elm: InputElement) {
+        self.goal_symbol = elm;
     }
 
-    /// Preview the next character but don't actually increment
-    fn preview_next(&mut self) -> Option<char> {
-        self.buffer.peek().copied()
+    #[inline]
```
Razican, line 77:
```suggestion
    /// Gets the goal symbol the lexer is currently using.
    #[inline]
```

boa/src/syntax/lexer/mod.rs
```diff
@@ -3,869 +3,261 @@
...
-    ///     lexer.lex()
-    /// }
-    /// ```
-    pub fn lex(&mut self) -> Result<(), LexerError> {
-        let _timer = BoaProfiler::global().start_event("lex", "lexing");
-        loop {
-            // Check if we've reached the end
-            if self.preview_next().is_none() {
-                return Ok(());
+    #[allow(clippy::should_implement_trait)] // We intentionally don't implement Iterator trait as Result<Option> is cleaner to handle.
```
Razican, line 148:
```suggestion
    /// Retrieves the next token from the lexer.
    // We intentionally don't implement Iterator trait as Result<Option> is cleaner to handle.
    #[allow(clippy::should_implement_trait)] 
```

boa/src/syntax/lexer/mod.rs
```diff
@@ -3,869 +3,261 @@
...
+                Punctuator::Question.into(),
+                Span::new(start, self.cursor.pos()),
+            )),
+            '/' => self.lex_slash_token(start),
+            '=' | '*' | '+' | '-' | '%' | '|' | '&' | '^' | '<' | '>' | '!' | '~' => {
+                Operator::new(next_chr).lex(&mut self.cursor, start)
+            }
+            _ => {
+                let details = format!(
+                    "Unexpected '{}' at line {}, column {}",
```
Razican, line 230:
```suggestion
                    "unexpected '{}' at line {}, column {}",
```

boa/src/syntax/lexer/regex.rs
```diff
@@ -0,0 +1,82 @@
...
+                        }
+                        _ => body.push(c),
+                    }
+                }
+            }
+        }
+
+        // body was parsed, now look for flags
+        let mut flags = String::new();
+        cursor.take_until_pred(&mut flags, &char::is_alphabetic)?;
```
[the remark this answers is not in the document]
Lan2u: Yeah it should be take_while_pred

boa/src/syntax/lexer/operator.rs
Lan2u: It is handled by the match below

boa/src/syntax/lexer/spread.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        // . or ...
+        if cursor.next_is('.')? {
+            if cursor.next_is('.')? {
+                Ok(Token::new(
+                    Punctuator::Spread.into(),
+                    Span::new(start_pos, cursor.pos()),
+                ))
+            } else {
+                Err(Error::syntax("Expecting Token ."))
```
Lan2u, line 46:
```suggestion
                Err(Error::syntax("Expecting Token '.' as part of spread"))
```

Razican (commented):
I think we should add a test for the lexing error positions.

Lan2u:
> I think we should add a test for the lexing error positions.

Added some :)

boa/src/syntax/lexer/cursor.rs
```diff
@@ -0,0 +1,286 @@
...
+                    io::ErrorKind::UnexpectedEof,
+                    "unexpected EOF when filling buffer",
+                )
+            })?;
+        }
+        Ok(())
+    }
+
+    /// Retrieves the next UTF-8 checked character.
+    #[inline]
```
Razican, line 214: I wouldn't inline this by default. It's a pretty big function, I would leave LLVM to do the heuristics, what do you think?
```suggestion
```

boa/src/syntax/lexer/tests.rs
```diff
@@ -594,23 +595,32 @@ fn illegal_following_numeric_literal() {
...
-        lexer.next().is_err(),
-        "DecimalDigit following NumericLiteral not rejected as expected"
-    );
+    match lexer.next() {
+        Err(Error::Syntax(_, pos)) => assert_eq!(pos, Position::new(1, 5)),
+        _ => assert!(
+            false,
+            "DecimalDigit following NumericLiteral not rejected as expected"
+        ),
+    }
```
Razican, line 604: I think we should use `expect_err()` in these, makes things much cleaner:
```suggestion
    let err = lexer.next().expect_err("DecimalDigit following NumericLiteral not rejected as expected");
    if let Error::Syntax(_, pos) = err {
        assert_eq!(pos, Position::new(1, 5))
    } else {
        panic!("invalid error type");
    }
```

Razican (approved):
This looks good. I would just change a couple of things as per my comments. If you think they are reasonable, I can do the change myself today. I think we can merge this today, if the team is fine with it :)

jasonwilliams:
I will try to take a look today

boa/src/syntax/lexer/cursor.rs
Lan2u: Yeah makes sense

jasonwilliams (approved):

HalidOdat (approved):

HalidOdat:
> I think we can merge this today, if the team is fine with it :)

Lets merge this :)

PR #560, opened 2020-07-10
Added benchmarks for full program execution

Razican:
This adds benchmarks for full program execution.

Sometimes we have benchmarks that tell us things like "the parser is now 50% slower", but what does that mean for the whole program execution? Maybe it's just a 1% speed reduction.

Also, we have no benchmarks to check improvements if for example, we create the realm in a new thread in parallel to lexing + parsing. This should reduce at least a bit the full program execution time, but there is no way to measure it.

With this PR, all execution benchmarks get duplicated by using our `exec()` function, that creates a realm, lexes and parses and finally executes everything. This should give us insights on how each change affects to a full workflow.

I also took the opportunity to remove some non-needed black boxes from old benchmarks. They were just preventing optimizations in pre-benchmark code, so they were not useful.

Ideally, this should land before #486/#559 in order to see the change in that PR.

### DIFF
--- a/boa/Cargo.toml
+++ b/boa/Cargo.toml
@@ -51,3 +51,7 @@ harness = false
 [[bench]]
 name = "exec"
 harness = false
+
+[[bench]]
+name = "full"
+harness = false
--- a/boa/benches/exec.rs
+++ b/boa/benches/exec.rs
@@ -26,11 +26,11 @@ fn symbol_creation(c: &mut Criterion) {
     let mut engine = Interpreter::new(realm);
 
     // Lex all the tokens.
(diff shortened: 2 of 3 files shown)


Lan2u:
Is it potentially also worth adding some more examples of 'less abstract' javascript?

Razican:
> Is it potentially also worth adding some more examples of 'less abstract' javascript? 

What do you mean? If you find interesting stuff to benchmark, feel free to add it :)

Lan2u:
> > Is it potentially also worth adding some more examples of 'less abstract' javascript?
> 
> What do you mean? If you find interesting stuff to benchmark, feel free to add it :)

I'll see if I can find some good examples

PR #561, opened 2020-07-11
Throw a type error when a non-object is called

### LINKED ISSUE #510: [Call] calling an undefined method does not throw
**To Reproduce**
```js
'bar'.foo()
```
or
```js
let bar = {}
bar.foo()
```

**Expected behavior**
Should throw TypeError 
`TypeError: 'bar'.foo is not a function`
`TypeError: bar.foo is not a function`

Instead `undefined` is returned.

joshwd36:
This Pull Request fixes #510.

Previously if a value other than an object was called as a function undefined would be returned. Now throws a type error.


joshwd36:
> 
> 
> Looks good! could we have a test for this, maybe something like:
> 
> ```js
> let x = false;
> try {
> 	undefined();
> } catch(e) {
> 	x = true;
> }
> ```
> 
> and we assert that `x == true`

Yes, I'll do that. I was wondering what the best way to test exceptions would be.

croraf:
> > Looks good! could we have a test for this, maybe something like:
> > ```js
> > let x = false;
> > try {
> > 	undefined();
> > } catch(e) {
> > 	x = true;
> > }
> > ```
> > 
> > 
> > and we assert that `x == true`
> 
> Yes, I'll do that. I was wondering what the best way to test exceptions would be.

I don't like this. I've been thinking about that when I had the same issue. I concluded that it is much better to do the following:

```js
let x = false;
try {
	undefined();
} catch(e) {
	e.message;
}
```

and then assert that the return of this is the correct string (the error message).

joshwd36:
That's a good point, and is actually what I ended up doing with the tests as I saw that had been used in the tests file already.

croraf:
Oh, cool. I see now.

I was thinking before how to add to the message "what is not a function", like Chrome does:
image

But this seamed difficult to do for me, as the subject on which the `Call` is invoked can be a wide variety of things:

`(some_complex_and_long_expression_that_returns_a_function)()`

and Chrome has some way on printing that out in a short form.

Would be worth to explore how to do this, perhaps as a separate Issue. What do you think?

HalidOdat:
```js
try {
	//expression that throws the error here;
} catch(e) {
	e.message;
}
```
This gives us `msg`

Yes. this this is a step in the right direction, but we still aren't testing what type of error it is, which is the most important part of testing (the spec does not specify what error message there should be, only what type of error it is and the test262 don't test for messages either), I think we can improve this by:
```js
try {
	//expression that throws the error here;
} catch(e) {
	e.toString();
}
```
This gives use `TypeError: msg`

What do you think?

> But to me this seamed as a difficult task, as the subject on which the `Call` is invoked can be a wide variety of things:
> 
> `(some_complex_and_long_expression_that_returns_a_function)()`
> 
> and Chrome has some way on printing that out in a short form.
> 
> Would be worth to explore how to do this, perhaps as a separate Issue. What do you think? (Perhaps it turns out it is not that difficult as I think.)

Maybe this could be done by keeping a span of the source code for function call node and when we error we concat the source span and the message description `is not a function` to get `someExpression is not a function`.

joshwd36:
> 
> 
> ```js
> try {
> 	//expression that throws the error here;
> } catch(e) {
> 	e.message;
> }
> ```
> 
> This gives us `msg`
> 
> Yes. this this is a step in the right direction, but we still aren't testing what type of error it is, which is the most important part of testing, I think we can improve this by:
> 
> ```js
> try {
> 	//expression that throws the error here;
> } catch(e) {
> 	e.toString();
> }
> ```
> 
> This gives use `TypeError: msg`
> 
> What do you think?
> 

That's a good point. I'll make a seperate PR for that because there are already a few tests that use the previous method, and this one has already been merged.

PR #562, opened 2020-07-12
Add missing Number methods.

joshwd36:
This Pull Request adds the missing Number and Global methods:
 - isFinite()
 - isNaN()
 - Number.isFinite()
 - Number.isInteger()
 - Number.isNaN()
 - Number.isSafeInteger()

### DIFF
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -576,9 +576,9 @@ impl Number {
     ) -> ResultValue {
         if let Some(val) = args.get(0) {
             let number = ctx.to_number(val)?;
-            Ok(Value::Boolean(number.is_finite()))
+            Ok(number.is_finite().into())
         } else {
-            Ok(Value::Boolean(false))
+            Ok(false.into())
         }
     }
 
@@ -603,9 +603,9 @@ impl Number {
     ) -> ResultValue {
         if let Some(val) = args.get(0) {
             let number = ctx.to_number(val)?;
-            Ok(Value::Boolean(number.is_nan()))
+            Ok(number.is_nan().into())
         } else {
-            Ok(Value::Boolean(true))
+            Ok(true.into())
         }
     }
 
@@ -628,7 +628,7 @@ impl Number {
         args: &[Value],
         _ctx: &mut Interpreter,
     ) -> ResultValue {
-        Ok(Value::Boolean(if let Some(val) = args.get(0) {
+        Ok(Value::from(if let Some(val) = args.get(0) {
             match val {
                 Value::Integer(_) => true,
                 Value::Rational(number) => number.is_finite(),
@@ -676,7 +676,7 @@ impl Number {
         args: &[Value],
         _ctx: &mut Interpreter,
     ) -> ResultValue {
-        Ok(Value::Boolean(if let Some(val) = args.get(0) {
+        Ok(Value::from(if let Some(val) = args.get(0) {
             match val {
                 Value::Integer(_) => false,
                 Value::Rational(number) => number.is_nan(),
@@ -706,14 +706,13 @@ impl Number {
         args: &[Value],
         _ctx: &mut Interpreter,
     ) -> ResultValue {
-        Ok((match args.get(0) {
+        Ok(Value::from(match args.get(0) {
             Some(Value::Integer(_)) => true,
             Some(Value::Rational(number)) if Self::is_float_integer(*number) => {
                 number.abs() <= Number::MAX_SAFE_INTEGER
             }
             _ => false,
-        })
-        .into())
+        }))
     }
 
     /// Checks if the argument is a finite integer Number value.
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -632,14 +632,6 @@ impl Number {
             match val {
                 Value::Integer(_) => true,
                 Value::Rational(number) => number.is_finite(),
-                Value::Object(ref obj) => {
-                    let obj = obj.borrow();
-                    if let Some(number) = obj.as_number() {
-                        number.is_finite()
-                    } else {
-                        false
-                    }
-                }
                 _ => false,
             }
         } else {
@@ -662,11 +654,7 @@ impl Number {
         args: &[Value],
         _ctx: &mut Interpreter,
     ) -> ResultValue {
-        Ok(Value::Boolean(if let Some(val) = args.get(0) {
-            Number::is_integer(val).is_some()
-        } else {
-            false
-        }))
+        Ok(args.get(0).map_or(false, Self::is_integer).into())
     }
 
     /// `Number.isNaN( number )`
@@ -692,14 +680,6 @@ impl Number {
             match val {
                 Value::Integer(_) => false,
                 Value::Rational(number) => number.is_nan(),
-                Value::Object(ref obj) => {
-                    let obj = obj.borrow();
-                    if let Some(number) = obj.as_number() {
-                        number.is_nan()
-                    } else {
-                        false
-                    }
-                }
                 _ => false,
             }
         } else {
@@ -726,15 +706,14 @@ impl Number {
         args: &[Value],
         _ctx: &mut Interpreter,
     ) -> ResultValue {
-        Ok(Value::Boolean(if let Some(val) = args.get(0) {
-            if let Some(integer) = Number::is_integer(val) {
-                integer.abs() <= Number::MAX_SAFE_INTEGER
-            } else {
-                false
+        Ok((match args.get(0) {
+            Some(Value::Integer(_)) => true,
+            Some(Value::Rational(number)) if Self::is_float_integer(*number) => {
+                number.abs() <= Number::MAX_SAFE_INTEGER
             }
-        } else {
-            false
-        }))
+            _ => false,
+        })
+        .into())
     }
 
     /// Checks if the argument is a finite integer Number value.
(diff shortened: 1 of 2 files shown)


boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,221 @@ impl Number {
...
+                        number.is_finite()
+                    } else {
+                        false
+                    }
+                }
+                _ => false,
+            }
+        } else {
+            false
+        }))
```
HalidOdat, line 647:
```suggestion
        if let Some(Value::Boolean(number)) = args.get(0) {
            return Ok(number.is_finite().into());
        }
        
        Ok(Value::from(false))
```
The `Number.isFinite()` does not check for number object <https://tc39.es/ecma262/#sec-number.isfinite>

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,221 @@ impl Number {
...
+    }
```
HalidOdat, line 708:
```suggestion
    pub(crate) fn number_is_nan(
        _this: &Value,
        args: &[Value],
        _ctx: &mut Interpreter,
    ) -> ResultValue {
        if let Some(Value::Boolean(number)) = args.get(0) {
            return Ok(number.is_nan().into());
        }
        
        Ok(Value::from(false))
    }
```
same here `Number.isNaN()` does not check for number object only primitive number https://tc39.es/ecma262/#sec-isnan-number

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,221 @@ impl Number {
...
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
```
HalidOdat, line 771:
```suggestion
    pub(crate) fn is_integer(val: &Value) -> bool {
        match val {
            Value::Integer(number) => true,
            Value::Rational(number) => {
                number.is_finite() && number.abs().floor() == number.abs()
            }
            _ => false,
        }
    }
```
`isInteger` does not check for number object <https://tc39.es/ecma262/#sec-isinteger>, also it it returns a `bool` I don't see the reason for `Option<f64>` could you explain this.

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,221 @@ impl Number {
...
+        Ok(Value::Boolean(if let Some(val) = args.get(0) {
+            if let Some(integer) = Number::is_integer(val) {
+                integer.abs() <= Number::MAX_SAFE_INTEGER
+            } else {
+                false
+            }
+        } else {
+            false
+        }))
+    }
```
HalidOdat, line 738:
```suggestion
    pub(crate) fn is_safe_integer(
        _this: &Value,
        args: &[Value],
        _ctx: &mut Interpreter,
    ) -> ResultValue {
        if let Some(val) = args.get(0) {
            if Self::is_integer(val) {
                return Ok(Value::from(integer.abs() <= Number::MAX_SAFE_INTEGER));
            }
        }
        
        Ok(Value::from(false))
    }
```

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,221 @@ impl Number {
...
+    pub(crate) fn number_is_integer(
+        _this: &Value,
+        args: &[Value],
+        _ctx: &mut Interpreter,
+    ) -> ResultValue {
+        Ok(Value::Boolean(if let Some(val) = args.get(0) {
+            Number::is_integer(val).is_some()
+        } else {
+            false
+        }))
```
HalidOdat, line 669:
```suggestion
        Ok(args.get(0).map_or(false, Self::is_integer).into())
```
simplify method

boa/src/builtins/number/mod.rs
joshwd36: I might be missing something in the spec, but from what I can see it should consider Number objects. In both chrome and firefox the following snippets are valid:
```javascript
Number.isFinite(Number(5)) // true
Number.isFinite(Number(Infinity)) // false
```

HalidOdat: > I might be missing something in the spec, but from what I can see it should consider Number objects. In both chrome and firefox the following snippets are valid:
> 
> ```js
> Number.isFinite(Number(5)) // true
> Number.isFinite(Number(Infinity)) // false
> ```

The `Number` function has two sides:
```js
Number(10) // 10 number primitive is returned.
new Number(10) // `Number { 10 }` number object is returned with value 10
```

joshwd36: Ah, OK, I wasn't aware of the difference. I'll make those changes then.

boa/src/builtins/number/mod.rs
joshwd36: The reason I wrote it like that is that the is_safe_integer() method then uses the value as a float to compare to MAX_SAFE_INTEGER, so it would save matching the value again.

HalidOdat: > The reason I wrote it like that is that the is_safe_integer() method then uses the value as a float to compare to MAX_SAFE_INTEGER, so it would save matching the value again.

Yes. but it would be better if do something like:
```rust
#[inline]
pub(crate) fn is_integer(number: f64) -> bool {
	number.is_finite() && number.abs().floor() == number.abs()
}
```

and in the `is_safe_integer()`:
```rust
	pub(crate) fn is_safe_integer(
        _this: &Value,
        args: &[Value],
        _ctx: &mut Interpreter,
    ) -> ResultValue {
        match args.get(0) {
			Some(Value::Integer(_)) => true,
			Some(Value::Number(number)) if Self::is_integer(number) => {
				number.abs() <= Number::MAX_SAFE_INTEGER
			}
			_ => false 
		}
    }
```

The other way was slightly more inefficient because when it was an `Value::Integer` (is is always a safe integer) we only need to return true, instead of conversion it to `f64` and than doing:
```
if Self::is_integer(number) => {
	number.abs() <= Number::MAX_SAFE_INTEGER
}
```
this to check if the already safe integer is a safe integer. also it removes the check from having an `Option`

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,188 @@ impl Number {
...
+    pub(crate) fn global_is_finite(
+        _this: &Value,
+        args: &[Value],
+        ctx: &mut Interpreter,
+    ) -> ResultValue {
+        if let Some(val) = args.get(0) {
+            let number = ctx.to_number(val)?;
+            Ok(Value::Boolean(number.is_finite()))
+        } else {
+            Ok(Value::Boolean(false))
```
HalidOdat, line 581:
```suggestion
            Ok(number.is_finite().into())
        } else {
            Ok(false.into())
```
Prefer using the `from`/`into` functions over direct `enum` field access when constructing a `Value`, this makes it easier to refacor `Value` in the future, since we have a layer of abstraction.

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,188 @@ impl Number {
...
+    pub(crate) fn global_is_nan(
+        _this: &Value,
+        args: &[Value],
+        ctx: &mut Interpreter,
+    ) -> ResultValue {
+        if let Some(val) = args.get(0) {
+            let number = ctx.to_number(val)?;
+            Ok(Value::Boolean(number.is_nan()))
+        } else {
+            Ok(Value::Boolean(true))
```
HalidOdat, line 608: here too.

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,188 @@ impl Number {
...
+    ) -> ResultValue {
+        Ok(Value::Boolean(if let Some(val) = args.get(0) {
+            match val {
+                Value::Integer(_) => true,
+                Value::Rational(number) => number.is_finite(),
+                _ => false,
+            }
+        } else {
+            false
+        }))
```
HalidOdat, line 639:
```suggestion
        Ok(Value::from(if let Some(val) = args.get(0) {
            match val {
                Value::Integer(_) => true,
                Value::Rational(number) => number.is_finite(),
                _ => false,
            }
        } else {
            false
        }))
```

boa/src/builtins/number/tests.rs
```diff
@@ -677,3 +677,178 @@ fn parse_float_too_many_args() {
...
+    assert_eq!("true", &forward(&mut engine, "isNaN(undefined)"));
+    assert_eq!("true", &forward(&mut engine, "isNaN({})"));
+    assert_eq!("false", &forward(&mut engine, "isNaN(true)"));
+    assert_eq!("false", &forward(&mut engine, "isNaN(null)"));
+    assert_eq!("false", &forward(&mut engine, "isNaN(37)"));
+    assert_eq!("false", &forward(&mut engine, "isNaN('37')"));
+    assert_eq!("false", &forward(&mut engine, "isNaN('37.37')"));
+    assert_eq!("true", &forward(&mut engine, "isNaN('37,5')"));
+    assert_eq!("true", &forward(&mut engine, "isNaN('123ABC')"));
+    // Incorrect due to ToNumber implementation inconsistencies.
```
HalidOdat, line 713: String to number with the `to_number` function is not 100% spec compliant: <https://github.com/boa-dev/boa/blob/master/boa/src/exec/mod.rs#L329>

To fix this we should trim and if it is a empty string return 0 else we parse as we currently do now.

joshwd36: Should this be done in this PR or should there be a separate one to try and make to_number fully spec compliant?

HalidOdat: > Should this be done in this PR or should there be a separate one to try and make to_number fully spec compliant?

Probably in a separate PR is better :)

boa/src/builtins/number/mod.rs
```diff
@@ -555,6 +555,187 @@ impl Number {
...
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-number.isinteger
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger
+    pub(crate) fn number_is_integer(
+        _this: &Value,
+        args: &[Value],
+        _ctx: &mut Interpreter,
+    ) -> ResultValue {
+        Ok(args.get(0).map_or(false, Self::is_integer).into())
```
HalidOdat, line 657:
```suggestion
                Ok(Value::from(match args.get(0) {
            Some(Value::Integer(_)) => true,
            Some(Value::Rational(number)) => Self::is_integer(number), 
            _ => false,
        }))
```
Its better if we do it this way this gives us a fast path for when the value is a `Value::Integer(_)` since it is always an integer because it is stored as a `Value::Integer(i32)`.

joshwd36: Doesn't is_integer have a fast path for Value::Integer(_) already? Doing this would save a function call as is_integer() might not be inlined, but without that they should already be the same.

HalidOdat: > Doesn't is_integer have a fast path for Value::Integer(_) already? Doing this would save a function call as is_integer() might not be inlined, but without that they should already be the same.

Ow sorry, I was mistaken :sweat_smile: , 

> Doing this would save a function call as is_integer() might not be inlined

I would suggest putting `#[inline]` on `is_integer` and is `is_float_integer` this will help with inlining, this is just a hint to the compiler.

joshwd36: It was already on is_float_integer, but I've done that for is_integer now :)

PR #566, opened 2020-07-15
Remove vec-queue

Lan2u:
Initial / prototype implementation just to see if this is having any impact.

Still todo:
- Replace the pushed_back field with just using an array of size 4
- Stop using .clone() on token and start using references as much as possible


Lan2u:
Initial benchmark seems to indicate an improvement, will continue to refine this.

Issue #507, opened 2020-06-18
`ReferenceError` on cli prints full obj properties instead of short message


AnirudhKonduru:
On the cli, simply running `ReferenceError`, `console`, `console.log` and many other global objects/functions prints the full object and its properties of what the script returns instead of a short message like `function: ReferenceError` or `global object: console`. 

### to reproduce
Run `console`, `console.log`, `ReferenceError` on the cli

**Expected behavior**
Print a shorter description like `function: ReferenceError` or `global object: console`

The spec define not defineewhat to /how to print what the script returns.

**Build environment (please complete the following information):**
 - OS: [Windows10]
 - Version: [e.g. 64]
 - Target triple: [x86_64-uwp-windows-msvc]
 - Rustc version: [rustc 1.44.0 (49cae5576 2020-06-01)]

HalidOdat:
We should discus how we should print the objects and errors (uncaught errors) we are not bound by the spec so we can print whatever we want!

I would like for objects to be printed all objects (not `Error`s): `ObjectName { MaybeValueOrEmpty }`, for example:
```js
new String("Hello") // => String { "Hello" }
new Boolean(true) // => Boolean { true }
```

For errors like type error, like `TypeError`
```js
new TypeError(msg) // => Uncaught TypeError: msg (maybe in red)
throw "Hello"; // =>  Uncaught "Hello" (this is how node REPL does it)
```

HalidOdat:
Additionally we cold have a flag for displaying all the object properties like we do now for debugging purposes.

AnirudhKonduru:
Sound great to me. Perhaps for uncaught Errors we could print the stack trace also? (maybe just 2-3 levels of it on the cli for brevity).

PR #572, opened 2020-07-16
Spec Compliant `Number.prototype.toString()`, better `Number` object formating and `-0`

HalidOdat:
This Pull Request is related to #507 .

It changes the following:
 - Made number object printing `Number { numberValue }`.
 - `-0` is now `-0` instead of positive `0` this is done by V8 and SpiderMonkey, even though it does against the tostring spec, in REPL we are not bound by the ECMAScript spec so we can do this, it makes it easier to know if the number is a `-0` or a positive `0`.
 - Made `Number.prototype.toString()` spec compliant
 - Enabled `Number.prototype.toString()` tests

```js
new Number(1000); // -> Number { 1000 }
```

### DIFF
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -361,18 +361,8 @@ impl Number {
 
     #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_native_string(x: f64) -> String {
-        if x == -0. {
-            return "0".to_owned();
-        } else if x.is_nan() {
-            return "NaN".to_owned();
-        } else if x.is_infinite() && x.is_sign_positive() {
-            return "Infinity".to_owned();
-        } else if x.is_infinite() && x.is_sign_negative() {
-            return "-Infinity".to_owned();
-        }
-
-        // FIXME: This is not spec compliant.
-        format!("{}", x)
+        let mut buffer = ryu_js::Buffer::new();
+        buffer.format(x).to_string()
     }
 
     /// `Number.prototype.toString( [radix] )`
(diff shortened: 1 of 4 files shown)


boa/src/builtins/number/mod.rs
```diff
@@ -400,6 +390,13 @@ impl Number {
                 .throw_range_error("radix must be an integer at least 2 and no greater than 36");
         }
 
+        // 5. If radixNumber = 10, return ! ToString(x).
+        // This part should use exponential notations for long integer numbers commented tests
+        if radix == 10 {
+            // return Ok(to_value(format!("{}", Self::to_number(this).to_num())));
```
Razican, line 396: Should we remove these comments?

boa/src/builtins/value/display.rs
```diff
@@ -67,6 +67,14 @@ pub(crate) fn log_string_from(x: &Value, print_internals: bool) -> String {
             match v.borrow().data {
                 ObjectData::String(ref string) => format!("String {{ \"{}\" }}", string),
                 ObjectData::Boolean(boolean) => format!("Boolean {{ {} }}", boolean),
+                ObjectData::Number(rational) => {
```
Razican, line 70: Why doesn't this use `format_rational() to reduce code duplication?

boa/src/builtins/number/mod.rs
HalidOdat: > Should we remove these comments?

Yes. the comments are deprecated, but maybe we should keep the `5. If radixNumber = 10, return ! ToString(x).` because it is taken from the spec.

boa/src/builtins/value/display.rs
HalidOdat: > Why doesn't this use `format_rational() to reduce code duplication?

Yes. I wanted to use `format_rational` but it takes a `std::Formatter` and the other `log_*` functions don't maybe I should change them too so they take a `std::Formatter`

HalidOdat: > Yes. I wanted to use `format_rational` but it takes a `std::Formatter` and the other `log_*` functions don't maybe I should change them too so they take a `std::Formatter`

I think I'll leave it as is for now, the way things are displayed is a mess, and trying to change it to use `std::Formatter` would require an almost complete rewrite of it and I think it would be best to do this in a separate PR.

PR #574, opened 2020-07-17
parsing errors kill the parent process, return an error instead.

graham:

It changes the following:
  This isn't currently an issue, I wasn't able to determine based on the docs if this is intended, but while using Boa, it crashed the application I was running. Returning an error here seems more idiomatic, but there may be reasoning I'm unaware of.

Happy to improve if this isn't enough, or feel free to reject if there is some other reason I'm unaware of.

Love the project!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,6 +7,7 @@ dependencies = [
  "bitflags",
  "criterion",
  "gc",
+ "indexmap",
  "jemallocator",
  "measureme",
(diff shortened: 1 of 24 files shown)


HalidOdat:
Hey @graham, sorry for the late response.

I believe this has been fixed in the `test262` branch

HalidOdat:
Closing this since it's fixed in `test262` branch

PR #580, opened 2020-07-20
Extracted `__proto__` from internal slots

### LINKED ISSUE #578: Extract `prototype` from internal slots
The problem is that `prototype` instance is stored in the `internal_slots` the internal slots are stored as `HashMap<String, Value>`, this is slow, every time we want to access prototype we have to do hash lookup and the use it.

Changes:
 - Extract `__proto__` to `Object`, this should optimize prototype chain lookup

HalidOdat:
This Pull Request fixes/closes #578 .

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -15,7 +15,7 @@ mod tests;
 use super::function::{make_builtin_fn, make_constructor_fn};
 use crate::{
     builtins::{
-        object::{ObjectData, INSTANCE_PROTOTYPE, PROTOTYPE},
+        object::{ObjectData, PROTOTYPE},
         property::Property,
         value::{same_value_zero, ResultValue, Value},
     },
@@ -48,8 +48,7 @@ impl Array {
                 .expect("Could not get global object"),
         ));
         array.set_data(ObjectData::Array);
-        array.borrow().set_internal_slot(
-            INSTANCE_PROTOTYPE,
+        array.as_object_mut().expect("array object").set_prototype(
             interpreter
                 .realm()
                 .environment
@@ -116,7 +115,9 @@ impl Array {
         // Set Prototype
         let prototype = ctx.realm.global_obj.get_field("Array").get_field(PROTOTYPE);
 
-        this.set_internal_slot(INSTANCE_PROTOTYPE, prototype);
+        this.as_object_mut()
+            .expect("this should be an array object")
+            .set_prototype(prototype);
         // This value is used by console.log and other routines to match Object type
         // to its Javascript Identifier (global constructor method name)
         this.set_data(ObjectData::Array);
@@ -961,7 +962,7 @@ impl Array {
     ///
     /// The reduce method traverses left to right starting from the first defined value in the array,
     /// accumulating a value using a given callback function. It returns the accumulated value.
-    ///  
+    ///
     /// More information:
     ///  - [ECMAScript reference][spec]
     ///  - [MDN documentation][mdn]
--- a/boa/src/builtins/boolean/tests.rs
+++ b/boa/src/builtins/boolean/tests.rs
@@ -1,7 +1,7 @@
 use crate::{builtins::value::same_value, exec::Interpreter, forward, forward_val, realm::Realm};
 
 /// Test the correct type is returned from call and construct
-#[allow(clippy::result_unwrap_used)]
+#[allow(clippy::unwrap_used)]
 #[test]
 fn construct_and_call() {
     let realm = Realm::create();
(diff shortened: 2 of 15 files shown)


boa/src/builtins/array/mod.rs
```diff
@@ -48,16 +48,13 @@ impl Array {
...
-            INSTANCE_PROTOTYPE,
-            interpreter
-                .realm()
-                .environment
-                .get_binding_value("Array")
-                .expect("Array was not initialized")
-                .borrow()
-                .get_field(PROTOTYPE),
-        );
+        array.as_object_mut().unwrap().prototype = interpreter
```
Razican, line 51: I would use `expect()` here, saying something like "an array was not an object" or something like that.

boa/src/builtins/array/mod.rs
```diff
@@ -48,16 +48,13 @@ impl Array {
...
+            .realm()
+            .environment
+            .get_binding_value("Array")
+            .expect("Array was not initialized")
+            .borrow()
+            .get_field(PROTOTYPE);
```
Razican, line 57: We could have a `prototype()` getter here, and the `get_field()` use that function if the passed string is the correct one. I think this would make it faster most of the time.

boa/src/builtins/boolean/tests.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{builtins::value::same_value, exec::Interpreter, forward, forward_val, realm::Realm};
 
 /// Test the correct type is returned from call and construct
-#[allow(clippy::result_unwrap_used)]
+#[allow(clippy::unwrap_used)]
```
Razican, line 4: I think we should change the unwraps for `expect()`.

boa/src/builtins/object/mod.rs
```diff
@@ -58,6 +58,8 @@ pub struct Object {
     properties: FxHashMap<RcString, Property>,
     /// Symbol Properties
     symbol_properties: FxHashMap<u32, Property>,
+    /// Instance prototype `__proto__`.
+    pub(crate) prototype: Value,
```
Razican, line 62: I think it would be best if we used getters and setters instead of doing this public, in case we want to add logging or change how setting/getting the prototype works.

boa/src/builtins/object/mod.rs
```diff
@@ -130,6 +133,7 @@ impl Object {
             internal_slots: FxHashMap::default(),
             properties: FxHashMap::default(),
             symbol_properties: FxHashMap::default(),
+            prototype: Value::null(),
```
Razican, line 136: Hmmm I see many fields here have the default value. Can't we use the `..Self::default()` notation?

boa/src/exec/mod.rs
```diff
@@ -149,10 +149,10 @@ impl Interpreter {
             callable,
         );
 
-        let new_func = Object::function(func);
+        let mut new_func = Object::function(func);
 
+        new_func.prototype = function_prototype.clone();
```
Razican, line 154: Here I see we create a function and then assign it a prototype. I would do this with a `set_prototype()` setter as I explained before, but if we are using this a lot, we could maybe have a constructor that accepts a prototype too.

boa/src/exec/new/mod.rs
```diff
@@ -23,7 +23,8 @@ impl Executable for New {
         }
         let this = Value::new_object(None);
         // Create a blank object, then set its __proto__ property to the [Constructor].prototype
-        this.set_internal_slot(INSTANCE_PROTOTYPE, func_object.get_field(PROTOTYPE));
+        this.as_object_mut().expect("this is an object").prototype =
+            func_object.get_field(PROTOTYPE);
```
Razican, line 27: I'm guessing the `expect()` here should say the opposite, right? If it fails to unwrap, then "this was not an object".

boa/src/exec/tests.rs
```diff
@@ -762,7 +762,7 @@ mod in_operator {
         "#;
         forward(&mut engine, scenario);
         let a = forward_val(&mut engine, "bar").unwrap();
-        assert!(a.get_internal_slot(INSTANCE_PROTOTYPE).is_object(), true);
+        assert!(a.as_object().unwrap().prototype().is_object(), true);
     }
```
Razican, line 766: We don't need the `true` in the assert here I think.

boa/src/builtins/object/mod.rs
HalidOdat: We can use the ..Self::default() notation but, eventually we will have to remove them they are not spec compliant.

boa/src/exec/mod.rs
HalidOdat: It is probably better for the constructor to take it, function objects should always have a prototype instance.

boa/src/builtins/array/mod.rs
HalidOdat: I think we could store the builtin objects like `Array`, `String`, `Number` etc, in the interpreter or Realm as fields as a `GcObject` this will eliminate the Object lookup an we could do something like:
```rust
interperter.realm().array_object().prototype()
``` 
So we would cache the them.

HalidOdat:
> Maybe we could extend this to other things in the future too :)

Yes. I planing to deprecate the `intenal_slots` I already did this for `extensible` and now `__proto__` and in a future PR I will remove them entirely :)

PR #581, opened 2020-07-20
Added CLEAN_JS and MINI_JS benches

### LINKED ISSUE #564: Add benchmarks for "uglified" JS
Currently, all of our benchmarks take non-uglified JavaScript as input. This is fine for some scenarios, but we should also have benchmarks for "uglified" or "minified" JavaScript.

This usually creates smaller variable and function names, and removes spaces and new lines, which should speed up parsing and even a bit the execution.

neeldug:
These benches are arbitrary code which is subject to change,
functionally the programs are identical.

This Pull Request fixes/closes #564 .

It changes the following:
 - Exec and Full benchmarks to run with minified and _clean_ code for comparison.
 - 
 - 

Specifically went for slightly longer function for visible change when minifying code with whitespace and character savings.


Razican:
> Should this really go in exec? this to me seems like we are benchmarking the lexer/parser and not execution.

IF I'm not mistaken, the part inside of the `bench()` iterator is the execution (or the full thing), so they are in the correct place, but it would be very nice to add parsing benchmarks too.

neeldug:
> > Should this really go in exec? this to me seems like we are benchmarking the lexer/parser and not execution.
> 
> If I'm not mistaken, the part inside of the `bench()` iterator is the execution (or the full thing), so they are in the correct place, but it would be very nice to add parsing benchmarks too.

I can implement this in the same PR, if this is wanted?

Razican:
> I can implement this in the same PR, if this is wanted?

Yes, please, if you can, that would be more complete :)

Razican:
I think this is blocked until #640 lands, since it will need a refactor.

Lan2u:
#640 is now merged so this is unblocked :)

Razican (approved):
This is good from my side, but let's wait for #646 to land in order to not fail the tests.

PR #587, opened 2020-07-22
Implement Well-Known Symbols

### LINKED ISSUE #585: Implement Well-Known Symbols
It would be useful to implement the well-known symbols, in order to do things like properly implement iterators (#576). [ECMAScript specification](https://tc39.es/ecma262/#sec-well-known-symbols).

```javascript
Symbol.iterator;
```
The expected output is `Symbol(Symbol.iterator)`.

The simplest way of doing this would be to have the symbols initialised in the `init` function of `Symbol`. However, symbols use a hash (currently just a counter) in order to differentiate them, and generating a hash is currently a method on `Interpreter`, which is not passed to the init function. It was suggested that it might be possible to pass the interpreter to the init functions, but this might not be possible, as the init functions are called in the setup of Realm, in the `create_intrinsics` function, and a Realm is used to create an Interpreter so the Interpreter won't exist when the Symbol object is initialised.

One way to fix this would be to move the hash counter to Realm and pass the Realm to the init functions. This could potentially be quite disruptive, as currently the global object, which is a part of Realm, is passed to the init functions, and this would have to be replaced with the Realm in order to avoid borrowing the same thing twice, meaning every init function would have to be changed in order to accommodate this. However, I think this would be the cleanest solution in the long term.

Another less disruptive, but dirtier way of doing this would be to hardcode the hash values of the well-known symbols and set the hash counter in Interpreter to start at the next number. As there are a fixed number (for now) of well known symbols it should be possible to do this without the possibility of collisions. However, this is likely to lead to bugs if more well-known symbols are added or the hash is used for something else as well, and on the whole is not a clean way of solving this.

joshwd36:
This Pull Request closes #585.

It changes the following:
 - Moves create_intrinsics from Realm to Interpreter
 - Changes the signature of the `init` function from `pub(crate) fn init(global: &Value) -> (&str, Value)` to `pub(crate) fn init(interpreter: &mut Interpreter) -> (&'static str, Value)`, allowing the init functions to reference the interpreter
 - Add the Well-Known Symbols to the Symbol object

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -1115,7 +1115,7 @@ impl Array {
     /// Initialise the `Array` object on the global object.
     #[inline]
     pub(crate) fn init(interpreter: &mut Interpreter) -> (&'static str, Value) {
-        let global = &interpreter.realm.global_obj;
+        let global = interpreter.global();
         let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
         // Create prototype
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -1114,7 +1114,7 @@ impl Array {
 
     /// Initialise the `Array` object on the global object.
     #[inline]
-    pub(crate) fn init(interpreter: &mut Interpreter) -> (&str, Value) {
+    pub(crate) fn init(interpreter: &mut Interpreter) -> (&'static str, Value) {
         let global = &interpreter.realm.global_obj;
         let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -1114,7 +1114,8 @@ impl Array {
 
     /// Initialise the `Array` object on the global object.
     #[inline]
-    pub(crate) fn init(global: &Value) -> (&str, Value) {
+    pub(crate) fn init(interpreter: &mut Interpreter) -> (&str, Value) {
+        let global = &interpreter.realm.global_obj;
         let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
         // Create prototype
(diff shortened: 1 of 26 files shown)


boa/src/builtins/array/mod.rs
```diff
@@ -1114,7 +1114,8 @@ impl Array {
 
     /// Initialise the `Array` object on the global object.
     #[inline]
-    pub(crate) fn init(global: &Value) -> (&str, Value) {
+    pub(crate) fn init(interpreter: &mut Interpreter) -> (&'static str, Value) {
+        let global = &interpreter.realm.global_obj;
```
HalidOdat, line 1118: It would make the code cleaner  if we used a getter function here that would return a reference of the global object. `Interpreter::global()`, instead of `&interpreter.realm.global_obj`

boa/src/builtins/symbol/mod.rs
```diff
@@ -100,7 +100,40 @@ impl Symbol {
...
+            Some("Symbol.hasInstance".into()),
+            interpreter.generate_hash(),
+        );
+        let symbol_is_concat_spreadable = Symbol(
+            Some("Symbol.isConcatSpreadable".into()),
+            interpreter.generate_hash(),
+        );
+        let symbol_iterator = Symbol(Some("Symbol.iterator".into()), interpreter.generate_hash());
+        let symbol_match = Symbol(Some("Symbol.match".into()), interpreter.generate_hash());
+        let symbol_match_all = Symbol(Some("Symbol.matchAll".into()), interpreter.generate_hash());
```
Lan2u, line 118: I can't see the documentation for this in the spec?

joshwd36: They're listed here https://www.ecma-international.org/ecma-262/10.0/index.html#sec-well-known-symbols. I can add a comment later

Lan2u: I meant specifically for matchAll?

joshwd36: Ah, you're right in that it isn't in the ecma-international spec, however it is in the tc39 one (https://tc39.es/ecma262/#sec-well-known-symbols) which I must have referenced when implementing it. I'm not sure which one I should be going off so I can update it if needed

Lan2u: Ah right it must be new in the 2021 spec - keep it and if possible could you please update the link to the new spec 😃

PR #592, opened 2020-07-24
`RegExp` specialization

HalidOdat:
It changes the following:
 - Removed Dependency of Internal State
 - Removed Dependency of Internal slots
 - Merged create into init (like all the other objects)
 - Specialized `RegExp` object
 - Removed wrong methods that should be getters (but we don't support getter so I disabled them for now)

BTW: This is not to make the methods spec compliant, this PR only focus on removing the dependency (I tried not to change the old behaviour). Making the methods spec compliant should be done in a separate PR.

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -19,7 +19,7 @@ use crate::{
         map::ordered_map::OrderedMap,
         property::Property,
         value::{RcBigInt, RcString, RcSymbol, ResultValue, Value},
-        BigInt,
+        BigInt, RegExp,
     },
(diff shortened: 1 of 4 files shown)


boa/src/builtins/object/mod.rs
```diff
@@ -381,6 +383,20 @@ impl Object {
         }
     }
 
+    /// Checks if it a `BigInt` object.
```
Lan2u, line 386: Typo?

HalidOdat: Copy paste mistake

PR #593, opened 2020-07-24
Assignment to non-assignable expression should throw a `SyntaxError` 

### LINKED ISSUE #588: Assignment to non-assignable expression should throw a `SyntaxError`
**Describe the bug**
Assignment to non-assignable expression should throw a `SyntaxError`.

**To Reproduce**

```javascript
3 = 5 
```

**Expected behavior**
Explain what you expected to happen, and what is happening instead.

Running this code, should throw a `SyntaxError`.

SpiderMonkey: `SyntaxError: invalid assignment left-hand side`
V8: `Uncaught SyntaxError: Invalid left-hand side in assignment`

instead we get `5`

 - OS: Manjaro Linux
 - Rustc version: rustc 1.45.0 (5c1f21c3b 2020-07-13)

Lan2u:
This Pull Request fixes/closes #588.

It changes the following:
 - The parser will now raise a SyntaxError if a value is assigned to a constant.

Note:
A syntax error is usually an error raised by the lexer (its a LexError) but in this case the parser raises the error because the lexer doesn't have the scope to look at tokens - this should be fine because the lexer/parser are now tightly coupled.


boa/src/exec/tests.rs
```diff
@@ -1235,3 +1235,135 @@ fn not_a_function() {
...
+
+#[test]
+fn assignment_to_non_assignable() {
+    // Relates to the behaviour described at
+    // https://tc39.es/ecma262/#sec-assignment-operators-static-semantics-early-errors
+    let realm = Realm::create();
+    let mut engine = Interpreter::new(realm);
+
+    const ERR_MSG: &str =
+        "Parsing Error: Syntax Error: Syntax Error: Invalid left-hand side in assignment";
```
HalidOdat, line 1247: Why is it `"Parsing Error: Syntax Error: Syntax Error: Invalid left-hand side in assignment"`?

shouldn't a `SyntaxError` be thrown with `.throw_syntax_error()` when there is a syntax error. #293 should make it easier to test for syntax error since it prints it `SyntaxError: message`, not all the fields for the error object.

boa/src/exec/tests.rs
```diff
@@ -1235,3 +1235,135 @@ fn not_a_function() {
...
+    
+    5;
+
+
+
+
+
+
+    a
+    "#;
```
HalidOdat, line 1341: Hmmm. this to me seems like we are testing the lexer/parser rather than the execution behaviour

Lan2u: Good point, will move

boa/src/exec/tests.rs
Lan2u: This error originates in the parser/lexer though so it is before the interpreter (and therefore .throw_syntax_error)?

HalidOdat: In `forward` we parse and we also have the Interpreter as an parameter so we can call `.throw_syntax_error()`. but this is kind of pointless since return a `String`.

Why does `forward` return a string it does not make sense to me, it should `forwrad` the value (`ResultValue`) not convert it to a string, `forward_val` does that, which does not make sense either.
The default behaviour should be to forward the value and if we need a string we will convert it.

^ this is kind of unrelated to this issue, but could we at least match the out push string `SyntaxError: message`, and not `"Parsing Error: Syntax Error: Syntax Error: Invalid left-hand side in assignment"`

Lan2u: Ok yeah I understand what you mean, I'll switch the hardcoded expected message to instead just look for the message to contain 'SyntaxError: '.

Lan2u:
Blocked on #559

PR #595, opened 2020-07-27
Added syntax highlighting for strings

HalidOdat:
It changes the following:
 - Added syntax highlighting for strings (`""`, `''`)
 - Split rustyline helper to a separate file (`helper.rs`)
 - Made colors for highlighting constants.

### DIFF
--- a/boa_cli/src/main.rs
+++ b/boa_cli/src/main.rs
@@ -345,9 +345,11 @@ impl Highlighter for LineHighlighter {
         let mut coloured = line.to_string();
 
         let reg = Regex::new(
-            r"(?x)
+            r#"(?x)
             (?P<identifier>[$A-z_]+[$A-z_0-9]*) |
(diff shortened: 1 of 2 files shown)


boa_cli/src/main.rs
```diff
@@ -362,8 +364,12 @@ impl Highlighter for LineHighlighter {
                         }
                         _ => cap.as_str().to_string(),
                     }
-                } else if let Some(cap) = caps.name("op") {
+                } else if let Some(cap) = caps.name("string_double_quote") {
+                    cap.as_str().green().to_string()
+                } else if let Some(cap) = caps.name("string_single_quote") {
                     cap.as_str().green().to_string()
+                } else if let Some(cap) = caps.name("op") {
+                    cap.as_str().truecolor(214, 95, 26).to_string()
```
Lan2u, line 372: Would it be better to have the colours defined as constants at the top of the file?

PR #596, opened 2020-07-27
Date Implementation

### LINKED ISSUE #252: Date() is needed https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

jcdickinson:
This Pull Request fixes/closes #252.

It changes the following:
 - Add `[Global]` internal slot that contains nothing. This is used to differentiate between constructor calls and function calls in the `Date` constructor.
 - Add `Date` according to spec.


boa/src/builtins/date/mod.rs
```diff
@@ -114,6 +124,34 @@ impl Date {
         }
     }
 
+    fn map_this_time_value(
```
jcdickinson, line 127: This whole method seems needlessly complex. Am I missing something?

boa/src/builtins/date/tests.rs
```diff
@@ -0,0 +1,1428 @@
...
+        "new Date(Date.UTC(2020, 06, 08, 09, 16, 15, 779)).valueOf()",
+    )
+    .expect("Successful eval");
+    assert_eq!(Value::number(1594199775779f64), actual);
+
+    Ok(())
+}
+
+#[test]
+fn date_neg() -> Result<(), Box<dyn std::error::Error>> {
```
jcdickinson, line 1401: It looks like `Value::Date` is never used, because `Date` is a constructed type (unlike the primitives). This means that all the JSON and primitive logic goes unused, and e.g. `Object([Slot] = Date)` always becomes `NaN`.

Should I just delete `Value::Date` and update the `Value::Object` callsites to also inspect the internal slot?

Affects this test and the following test.

jcdickinson: One option would be to implement `TryFrom<ObjectData> for Value`, and recursively handle `Object` that way?

HalidOdat: > It looks like `Value::Date` is never used, because `Date` is a constructed type (unlike the primitives)
> Should I just delete `Value::Date`.

Yes. we should delete the `Value::Date`as it is not a primitive. 

Im not sure what you mean by:
> update `Value::Object` callsites to also inspect the internal slot?

The Date object internal slots should be stored as native types in `ObjectData`

boa/src/builtins/object/mod.rs
```diff
@@ -78,6 +78,8 @@ pub enum ObjectData {
     Symbol(RcSymbol),
     Error,
     Ordinary,
+    Date(RcDate),
```
HalidOdat, line 81: I don't need `RcData` at all, since it just stores a `Rc<Data>` and the thing that `Data` stores is copyable, so why not just store `ObjectData::Data(Data)` and copy it if we need it instead of keeping a reference counted pointer to an allocation.

jcdickinson: Good point! I should watch what I copy :).

boa/src/builtins/value/val_type.rs
```diff
@@ -28,6 +29,7 @@ impl Type {
             Self::Function => "function",
             Self::Object => "object",
             Self::BigInt => "bigint",
+            Self::Date => "date",
```
HalidOdat, line 32: We should remove this, there isn't a `"date"` primitive.

boa/src/builtins/date/tests.rs
HalidOdat: > One option would be to implement `TryFrom<ObjectData> for Value`, and recursively handle `Object` that way?

Could you explain a bit more about this, why do we need and what we are trying to solve

boa/src/builtins/value/val_type.rs
jcdickinson: That resolves my above concerns, I have removed `Date::Value` entirely and changed `to_json` to match the spec (it must call `toJSON()` if that function exists), as well as `neg` calling `to_numeric_number`.

jcdickinson:
Marking this as ready for review. This will likely need to be revisited when test262 works.

boa/src/builtins/date/mod.rs
```diff
@@ -0,0 +1,1421 @@
...
+        *month = if *month < 0 {
+            *year -= 1;
+            11 + (*month + 1) % 12
+        } else {
+            *month % 12
+        }
+    }
+
+    #[inline]
+    fn fix_day(year: &mut i32, month: &mut i32, day: &mut i32) {
```
Razican, line 68: Maybe this would be more efficient by passing the values and returning a result, since they should be easily copied in the stack. Maybe this would be easier by passing a custom `Date` structure around, with these three fields, and having the "fix" functions as methods.

boa/src/builtins/date/mod.rs
```diff
@@ -0,0 +1,1421 @@
...
+impl Display for Date {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.to_local() {
+            Some(v) => write!(f, "{}", v.format("%a %b %d %Y %H:%M:%S GMT%:z")),
+            _ => write!(f, "Invalid Date"),
+        }
+    }
+}
+
+unsafe impl Trace for Date {
```
Razican, line 196: I would add a comment here, explaining that this is only safe if `chrono` never implements `Trace` for `NaiveDateTime`.

boa/src/builtins/date/mod.rs
```diff
@@ -0,0 +1,1421 @@
...
+        self.to_utc()
+            .map_or(f64::NAN, |dt| dt.timestamp_millis() as f64)
+    }
+
+    pub fn to_json_string(&self) -> String {
+        self.to_utc()
+            // RFC 3389 uses +0.00 for UTC, where JS expects Z, so we can't use the built-in chrono function.
+            .map(|f| f.format("%Y-%m-%dT%H:%M:%S.%3fZ").to_string())
+            .unwrap_or_else(|| "Invalid Date".to_string())
+    }
```
Razican, line 236: I think we should improve documentation in these functions.

boa/src/builtins/date/mod.rs
jcdickinson: I have solved this by making `Date` the "custom structure." There's a new `set_components` method that all the other setters call.

Issue #557, opened 2020-07-09
Setting array length does not delete elements.


benjaminflin:
**Describe the bug**
According to the [spec](https://tc39.es/ecma262/#sec-properties-of-array-instances-length), shortening `array.length` should delete elements in the array. Currently, array elements that are accessed beyond the array length are not necessarily `undefined`.

**To Reproduce**
```javascript
let arr = [1, 2, 3, 4, 5];
arr.length = 3;
arr[4]; // 5
```

**Expected behavior**
`arr[4]` should be `undefined` but instead is `5`.

**Build environment**
 - OS: macOS Mojave 
 - Version: 10.14.3
 - Target triple: x86_64-apple-darwin
 - Rustc version: rustc 1.46.0-nightly (f455e46ea 2020-06-20)

Found in #555

54k1:
Hey I'm new to this codebase and decided to take a look at this. Please correct me if I'm wrong.
I have seen how arrays are implemented. Basically it is a dictionary with array indices as properties (as inferred from [this](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/array/mod.rs#L92-L106) builtin).

Object properties are separate from ObjectData and the function responsible for setting properties directly mutates the dictionary (properties field of struct Object) regardless of ObjectData. I think here's where the problem lies. We might need more information while setting properties or perhaps delegate this action to the data (ObjectData) field of struct Object.

This would require perhaps a new trait which constraits objects (array, etc.) to have get/set properties (among others). The default impl of the set property could be the one that exists [here](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/value/mod.rs#L681-L691). For special cases, this can be overridden.

Could you highlight the problems with this? Thanks.

HalidOdat:
Hi @54k1 

> I think here's where the problem lies. We might need more information while setting properties or perhaps delegate this action to the data (ObjectData) field of struct Object.

[...]

PR #600, opened 2020-07-31
Fix internal object methods

### LINKED ISSUE #591: Fix Internal Object Methods
**What is wrong with the current implementation?**
Currently are only set for ordinary objects, but not for array, string etc. Every object in the spec gets these internal methods an example is `[[DefineOwnProperty]]` for ordinary objects it calls `OrdinaryDefineOwnProperty` but for Array `[[DefineOwnProperty]]` is different when the property that is being defined is "length" it calls `ArraySetLength` this is what shrinks the array. This is what causes #557 bug.

We should avoid trait objects since it forces dynamic dispatch, and it also is not a viable solution.      
With trait objects we are trying to solve an inheritance problem, rust does not have a way to describe a base class where we have base class (with the the properties) and all the object (Array, etc) inherit from it. Also this would force dynamic dispatch this would slow down the code and it would not allow some compiler optimizations, we know the exact number of object types we don`t have do do this.

**My proposed changes:**
This is how I intended to do it, we have the internal method for example `[[set]]`, we also have the type definitions of what the object is `Array`, `Map`, etc. We have all we need to make this a static dispatch. So something like this:
```rust
impl Object {
	fn ordinary_set(&self /* , ... other parameters */) -> bool {
		// implementation
	}
	fn set(&mut self /* , ... other parameters */ ) -> bool {
		match self.data {
			ObjectData::Array => {
				// Arrays [[set]]
			}
			ObjectData::String(string) => {
				// Strings [[set]]
			}
			// This is for ordinay objects and for objects
			// that don't define a special implementation of [[set]]
			_ => self.ordinary_set(/* , ... args*/)
		}
	}
} 
```
This way of implementing allows the compiler to do some optimization that can't be done in dynamic dispatch. 

What do you think? any other better ideas? or improvements?

HalidOdat:
This PR closes #591 , #557, #602 

- Ordinary internal methods:
  - [ ] [OrdinarySet][ordinary-set]
  - [ ] [OrdinaryGet][ordinary-get]
  - [x] [OrdinaryDelete][ordinary-delete]
  - [ ] [OrdinaryToPrimitive][ordinary-to-primitive]
  - [ ] [OrdinaryHasInstance][ordinary-has-instance]
  - [x] [OrdinaryHasProperty][ordinary-has-property]
  - [x] [OrdinaryIsExtensible][ordinary-is-extensible]
  - [x] [OrdinaryPreventExtensions][ordinary-prevent-extensions]
  - [x] [OrdinaryGetPrototypeOf][ordinary-get-prototype-of]
  - [x] [OrdinarySetPrototypeOf][ordinary-set-prototype-of]
  - [x] [OrdinaryGetOwnProperty][ordinary-get-own-property]
  - [x] [OrdinaryDefineOwnProperty][ordinary-define-own-property]
- ...

[ordinary-set]: https://tc39.es/ecma262/#sec-ordinaryset
[ordinary-get]: https://tc39.es/ecma262/#sec-ordinaryget
[ordinary-delete]: https://tc39.es/ecma262/#sec-ordinarydelete
[ordinary-to-primitive]: https://tc39.es/ecma262/#sec-ordinarytoprimitive
[ordinary-has-instance]: https://tc39.es/ecma262/#sec-ordinaryhasinstance
[ordinary-has-property]: https://tc39.es/ecma262/#sec-ordinaryhasproperty
[ordinary-is-extensible]: https://tc39.es/ecma262/#sec-ordinaryisextensible
[ordinary-prevent-extensions]: https://tc39.es/ecma262/#sec-ordinarypreventextensions
[ordinary-get-prototype-of]: https://tc39.es/ecma262/#sec-ordinarygetprototypeof
[ordinary-set-prototype-of]: https://tc39.es/ecma262/#sec-ordinarysetprototypeof
[ordinary-get-own-property]: https://tc39.es/ecma262/#sec-ordinarygetownproperty
[ordinary-define-own-property]: https://tc39.es/ecma262/#sec-ordinarydefineownproperty

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -15,7 +15,7 @@
 
 use crate::builtins::{
     function::make_builtin_fn,
-    property::Property,
+    property::{Property, PropertyKey},
     value::{ResultValue, Value},
 };
 use crate::{exec::Interpreter, BoaProfiler};
@@ -54,7 +54,7 @@ impl Json {
                     Some(reviver) if reviver.is_function() => {
                         let mut holder = Value::new_object(None);
                         holder.set_field("", j);
-                        Self::walk(reviver, ctx, &mut holder, Value::from(""))
+                        Self::walk(reviver, ctx, &mut holder, &"".into())
                     }
                     _ => Ok(j),
                 }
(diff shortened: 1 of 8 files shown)


HalidOdat:
I think I'll start working on this again.

HalidOdat:
Closing this since it has too many conflict to be salvageable, and most of the stuff has already been implemented in other PRs like #602

PR #604, opened 2020-08-06
Set __proto__ of function instances

### LINKED ISSUE #531: [builtins] Builtin function doesnt have constructor property
**To Reproduce**
`[].findIndex.constructor` 

**Expected behavior**
Should return a `Function` constructor. Instead `undefined` is returned.

Custom functions do have `constructor` property which returns correctly.

54k1:
This Pull Request fixes/closes #531 

It changes the following:
- make_builtin_fn now takes &Interpreter as an argument.
- Call sites are updated according to the new function signature of make_builtin_fn

### DIFF
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -525,9 +525,8 @@ pub fn make_builtin_fn<N>(
     let name = name.into();
     let _timer = BoaProfiler::global().start_event(&format!("make_builtin_fn: {}", &name), "init");
 
-    // FIXME: function needs the Function prototype set.
-    let mut function = Object::function(Function::builtin(Vec::new(), function), Value::null());
-    function.set_prototype(
+    let mut function = Object::function(
+        Function::builtin(Vec::new(), function),
         interpreter
             .global()
             .get_field("Function")
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -513,8 +513,13 @@ pub fn make_constructor_fn(
 ///     some other number of arguments.
 ///
 /// If no length is provided, the length will be set to 0.
-pub fn make_builtin_fn<N>(function: NativeFunctionData, name: N, parent: &Value, length: usize)
-where
+pub fn make_builtin_fn<N>(
+    function: NativeFunctionData,
+    name: N,
+    parent: &Value,
+    length: usize,
+    interpreter: &Interpreter,
+) where
     N: Into<String>,
 {
     let name = name.into();
@@ -522,6 +527,13 @@ where
 
     // FIXME: function needs the Function prototype set.
     let mut function = Object::function(Function::builtin(Vec::new(), function), Value::null());
+    function.set_prototype(
+        interpreter
+            .global()
+            .get_field("Function")
+            .get_field("prototype"),
+    );
+
     function.insert_field("length", Value::from(length));
 
     parent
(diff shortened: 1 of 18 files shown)


boa/src/builtins/function/mod.rs
```diff
@@ -513,15 +513,27 @@ pub fn make_constructor_fn(
...
     let _timer = BoaProfiler::global().start_event(&format!("make_builtin_fn: {}", &name), "init");
 
     // FIXME: function needs the Function prototype set.
     let mut function = Object::function(Function::builtin(Vec::new(), function), Value::null());
+    function.set_prototype(
+        interpreter
+            .global()
+            .get_field("Function")
+            .get_field("prototype"),
+    );
```
HalidOdat, line 535: We can remove the `set_prototype` call, instead we initialize it in the `Object::function()` since it takes the function and the prototype we can pass it instead of `Value::null()`

Also we can now remove the `FIXME` comment since we fixed the issue.

54k1: Thanks for the review.

boa/src/builtins/function/mod.rs
```diff
@@ -513,15 +513,26 @@ pub fn make_constructor_fn(
...
     let _timer = BoaProfiler::global().start_event(&format!("make_builtin_fn: {}", &name), "init");
 
-    // FIXME: function needs the Function prototype set.
-    let mut function = Object::function(Function::builtin(Vec::new(), function), Value::null());
+    let mut function = Object::function(
+        Function::builtin(Vec::new(), function),
+        interpreter
+            .global()
+            .get_field("Function")
+            .get_field("prototype"),
```
Razican, line 533: I'm not super up to date with this, but isn't this similar to the `prototype()` function?

https://github.com/boa-dev/boa/blob/8fd5533460af8c0c424f6ddf4f3d45ed84815f8f/boa/src/builtins/object/mod.rs#L432

54k1: Yes, it looks like the final `get_field` does the same thing as the `prototype()` function.
Thanks for your comment.

HalidOdat: > I'm not super up to date with this, but isn't this similar to the `prototype()` function?

Its not the same. the `.prototype()` function does not return the `prototype` it returns the prototype instance so for `Function` it would be `Object.prototype`, not `Function.protoype`

It is confusing, it should be called `prototype_instance()` or something like that, since that is what is does

HalidOdat: Created an issue about the naming of `prototype` #607

54k1: > for `Function` object it would be `Object.prototype`, not `Function.protoype`

Shouldn't `Function.__proto__ === Function.prototype`?

HalidOdat: > Shouldn't `Function.__proto__ === Function.prototype`?

Actually you are right, sorry about that :sweat_smile: , the `Function` object has the `__proto__` set to `Function.prototype` since it is a function and it inherits from it self (`Function`). so in this case it would be the same.
Screenshot_2020-08-07_18-48-30

54k1: This turns out to be a 'special case'. :)

54k1: Since @HalidOdat mentioned about updating parts of the code to cache frequently used objects such as `Function`, `Object`, etc. can we leave it this way for now?

HalidOdat: > Since @HalidOdat mentioned about updating parts of the code to cache frequently used objects such as `Function`, `Object`, etc. can we leave it this way for now?

Yes. I was just suggesting a possible optimization that we can do in the future.

PR #609, opened 2020-08-07
Fix infinite recursion for Debug::fmt

### LINKED ISSUE #608: `Debug::fmt` Causes Causes a Stack Overflow
**Describe the bug**
For any object that contains `prototype`, `Debug::fmt` will cause a stack overflow. This is especially frustrating when some tests fail, as it can mask the reason for the failure.

**To Reproduce**

Evaluate the following JS and `println!("{:?}", result)`.
```javascript
new Date()
```

**Expected behavior**

Print the debug structure of the value.

**Actual behavior**

A stack overflow occurs.

 - OS: Windows
 - Version: 10
 - Target triple: `x86_64-unknown-windows-msvc`
 - Rustc version: `1.45.1 (c367798cf 2020-07-26)`

jcdickinson:
This Pull Request fixes/closes #608.

It changes the following:
 - Adds a test for stack overflow
 - Removes prototype from the `Debug::fmt` for `Object`

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -48,7 +48,7 @@ pub static PROTOTYPE: &str = "prototype";
 // pub static INSTANCE_PROTOTYPE: &str = "__proto__";
 
 /// The internal representation of an JavaScript object.
-#[derive(Debug, Trace, Finalize, Clone)]
+#[derive(Trace, Finalize, Clone)]
 pub struct Object {
     /// The type of the object.
     pub data: ObjectData,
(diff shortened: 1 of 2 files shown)


boa/src/builtins/object/mod.rs
```diff
@@ -64,6 +64,18 @@ pub struct Object {
     extensible: bool,
 }
 
+impl Debug for Object {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Object")
+            .field("data", &self.data)
+            .field("properties", &self.properties)
+            .field("symbol_properties", &self.symbol_properties)
```
HalidOdat, line 72: Couldn't any of these overflow for example:
```js
let x = {};
x.itSelf = x;
```
the `itSelf` property is contained in `properties` and this would cause a stack overflow, right?

jcdickinson: `prototype` _always_ overflows, I thought it valuable to get a fix in for that (because I suspect it is breaking the test262 PR) and then figure out how to approach this in a general way after some discussion.

Razican:
Why was this closed, @jcdickinson? I would be happy to have an patial fix for stack overflows.

jcdickinson:
@Razican I'll do a proper fix.

PR #624, opened 2020-08-11
Fix global objects attributes

### LINKED ISSUE #606: Altering vital properties of builtins
**Describe the bug**
Some internal properties (which must not be allowed to set by the user) are being set.

Ex:
```javascript
>> Object.prototype = 12
12
>> Object.prototype
>> new Object()
thread 'main' panicked at 'assertion failed: prototype.is_null() || prototype.is_object()', boa/src/builtins/object/mod.rs:433:9
```

**Expected behavior**
Such things must not alter the contents of a property. For example, in the case of `Object.prototype`, the [spec](https://tc39.es/ecma262/#sec-object.prototype) tells the property must be NOT (writable, enumerable, configurable). But it [looks like](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/function/mod.rs#L490-L493) the property is being set using the [insert_field](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/object/internal_methods.rs#L373) method where the attributes of the property are (WRITABLE, ENUMERABLE, CONFIGURABLE).

54k1:
This Pull Request fixes/closes #606 .

It changes the following:
1. Rename insert_property to insert.
2. Remove insert_field.
3. Add new insert method which takes Value, Attribute and inserts a property.


boa/src/builtins/function/mod.rs
```diff
@@ -472,25 +472,25 @@ pub fn make_constructor_fn(
         length.into(),
         Attribute::READONLY | Attribute::NON_ENUMERABLE | Attribute::PERMANENT,
     );
-    constructor.insert_property("length", length);
+    constructor.insert("length", length);
```
HalidOdat, line 475: Here we should use `.insert_property()` :
```rust
    constructor.insert_property("length", length, Attribute::READONLY | Attribute::NON_ENUMERABLE | Attribute::PERMANENT);
```

boa/src/builtins/function/mod.rs
```diff
@@ -472,25 +472,25 @@ pub fn make_constructor_fn(
...
 
     let name = Property::data_descriptor(
         name.into(),
         Attribute::READONLY | Attribute::NON_ENUMERABLE | Attribute::PERMANENT,
     );
-    constructor.insert_property("name", name);
+    constructor.insert("name", name);
```
HalidOdat, line 481: same here

boa/src/builtins/math/mod.rs
```diff
@@ -648,14 +649,15 @@ impl Math {
...
-            properties.insert_field("PI", Value::from(f64::consts::PI));
+            let attribute = Attribute::default();
+            properties.insert_property("E", Value::from(f64::consts::E), attribute);
+            properties.insert_property("LN2", Value::from(f64::consts::LN_2), attribute);
+            properties.insert_property("LN10", Value::from(f64::consts::LN_10), attribute);
+            properties.insert_property("LOG2E", Value::from(f64::consts::LOG2_E), attribute);
+            properties.insert_property("LOG10E", Value::from(f64::consts::LOG10_E), attribute);
+            properties.insert_property("SQRT1_2", Value::from(0.5_f64.sqrt()), attribute);
+            properties.insert_property("SQRT2", Value::from(f64::consts::SQRT_2), attribute);
+            properties.insert_property("PI", Value::from(f64::consts::PI), attribute);
```
HalidOdat, line 660:
```suggestion
            properties.insert_property("E", f64::consts::E, attribute);
            properties.insert_property("LN2", f64::consts::LN_2, attribute);
            properties.insert_property("LN10", f64::consts::LN_10, attribute);
            properties.insert_property("LOG2E", f64::consts::LOG2_E, attribute);
            properties.insert_property("LOG10E", f64::consts::LOG10_E, attribute);
            properties.insert_property("SQRT1_2", 0.5_f64.sqrt(), attribute);
            properties.insert_property("SQRT2", f64::consts::SQRT_2, attribute);
            properties.insert_property("PI", f64::consts::PI, attribute);
```
since `insert_property` calls `Into<Value>` we dont need to explicitly convert it to `Value`, this looks much nicer.

boa/src/builtins/object/internal_methods.rs
```diff
@@ -365,21 +365,24 @@ impl Object {
...
+        N: Into<PropertyKey>,
+        V: Into<Value>,
     {
         self.properties.insert(
-            name.into(),
-            Property::data_descriptor(
-                value,
-                Attribute::WRITABLE | Attribute::ENUMERABLE | Attribute::CONFIGURABLE,
-            ),
+            (&name.into()).into(),
```
HalidOdat, line 384: Why do we do this cant we just do:
```suggestion
            name.into(),
```

boa/src/builtins/object/internal_methods.rs
```diff
@@ -352,7 +352,7 @@ impl Object {
 
     /// Helper function for property insertion.
     #[inline]
-    pub(crate) fn insert_property<N>(&mut self, name: N, p: Property)
+    pub(crate) fn insert<N>(&mut self, name: N, p: Property)
```
HalidOdat, line 355: I think we should make `.insert` to be a private function to `Object`.
```suggestion
    fn insert<N>(&mut self, name: N, p: Property)
```

boa/src/builtins/object/internal_methods.rs
54k1: `name.into()` changes it to `PropertyKey`. `(&name.into()).into()` changes it to `RcString`.

54k1: Is there any way to get around that?

HalidOdat: Oh yeah. sorry forgot I fixed this in #603. this is a bug it converts a the `Symbol` into a string and stores in the properties.

Hmmm. In this leave it as is for now.

boa/src/builtins/mod.rs
```diff
@@ -81,7 +82,9 @@ pub fn init(interpreter: &mut Interpreter) {
         let global = interpreter.global();
         match global {
             Value::Object(ref global_object) => {
-                global_object.borrow_mut().insert_field(name, value);
+                global_object
+                    .borrow_mut()
+                    .insert_property(name, value, Attribute::default());
```
HalidOdat, line 87: Hmmm. There is a problem with this, the problem is that we don't set the right propety attributes for the global object/properties an example of this is the property `undefined` is (READONLY, NON_ENUMERABLE, PERMANENT) but `Object` global object is (`WRITABLE`, `NON_ENUMERABLE`, `CONFIGURABLE`). but here we are setting all as `READONLY`, `NON_ENUMERABLE`, `PERMANENT`.

One way of doing this would be to return a `(&'staitc, Value, Attribute)` with the specified attribute, another way is to have `init` functions not return anything and set the property in the `init` functions

54k1: Thanks for spotting this.

boa/src/builtins/array/mod.rs
```diff
@@ -1204,6 +1204,11 @@ impl Array {
         // Static Methods
         make_builtin_fn(Self::is_array, "isArray", &array, 1, interpreter);
 
-        (Self::NAME, array)
+        let mut global = interpreter.global().as_object_mut().expect("Expect object");
+        global.borrow_mut().insert_property(
+            Self::NAME,
+            array,
+            Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
+        );
```
HalidOdat, line 1212:
```suggestion
        let mut global = interpreter.global().as_object_mut().expect("Expect object");
        global.borrow_mut().insert_property(
            Self::NAME,
            array,
            Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
        );
```
I think the first solution would be nicer, since we dont have to repeat the code and we dont have to do extra checks, so we should return `(&'static str, Value, Attribute)` we should only have to do:
```
fn init(...) -> (&'static str, Value, Attribute) {
 	// ...
	let attribute = Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE;
	(Self::NAME, value, attribute)
}
```

boa/src/builtins/nan/mod.rs
```diff
@@ -25,9 +27,14 @@ impl NaN {
...
-    pub(crate) fn init(_interpreter: &mut Interpreter) -> (&'static str, Value) {
+    pub(crate) fn init(interpreter: &mut Interpreter) {
         let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
-        (Self::NAME, Value::from(f64::NAN))
+        let mut global = interpreter.global().as_object_mut().expect("Expect object");
+        global.borrow_mut().insert_property(
+            Self::NAME,
+            Value::from(f64::NAN),
+            Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
```
HalidOdat, line 37: `NaN` should be `Attribute::READONLY | Attribute::NON_ENUMERABLE | Attribute::PERMANENT`

boa/src/builtins/infinity/mod.rs
```diff
@@ -24,9 +26,13 @@ impl Infinity {
...
-    pub(crate) fn init(_interpreter: &mut Interpreter) -> (&'static str, Value) {
+    pub(crate) fn init(interpreter: &mut Interpreter) {
         let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
-
-        (Self::NAME, Value::from(f64::INFINITY))
+        let mut global = interpreter.global().as_object_mut().expect("Expect object");
+        global.borrow_mut().insert_property(
+            Self::NAME,
+            Value::from(f64::INFINITY),
+            Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
```
HalidOdat, line 35: `Infinity` should be the same as `undefined` and `NaN`

HalidOdat (approved):
This looks ready to merge, just needs a rebase :)

boa/src/builtins/value/mod.rs
```diff
@@ -486,7 +486,8 @@ impl Value {
         let _timer = BoaProfiler::global().start_event("Value::update_property", "value");
 
         if let Some(ref mut object) = self.as_object_mut() {
-            object.insert_property(field, new_property);
+            // FIXME overwrite the property
```
HalidOdat, line 489: Could you explain this comment? doesn't insert overwrite it?

boa/src/builtins/object/internal_methods.rs
```diff
@@ -336,17 +336,19 @@ impl Object {
...
             PropertyKey::String(ref string) => {
-                self.string_properties.insert(string.clone(), property)
+                self.string_properties.insert(string.clone(), property);
             }
             PropertyKey::Symbol(ref symbol) => {
-                self.symbol_properties.insert(symbol.clone(), property)
+                self.symbol_properties.insert(symbol.clone(), property);
             }
         }
     }
```
HalidOdat, line 354: it would be better if `.insert` returned an `Option<Property>`

boa/src/builtins/object/internal_methods.rs
```diff
@@ -361,21 +363,30 @@ impl Object {
...
+        match key.into() {
+            PropertyKey::Index(index) => self.indexed_properties.insert(index, property),
+            PropertyKey::String(ref string) => {
+                self.string_properties.insert(string.clone(), property)
+            }
+            PropertyKey::Symbol(ref symbol) => {
+                self.symbol_properties.insert(symbol.clone(), property)
+            }
+        }
     }
```
HalidOdat, line 391: If insert retruns an `Option<Property>` we could do:
```rust
let property = Property::data_descriptor(value.into(), attribute);
self.insert(key.into(), property)
```
This would eliminate some code duplication.

HalidOdat (commented):
This looks almost ready :) Check my comments on how to improve it

Also as @Lan2u said having some test to prove that it the object attributes are set would be nice, something like:
```js
Infinity = 3;
```
and asserting that Infinity did not change (this tests that its readonly), another test is to do: 
```js
delete Infinity;
```
And asserting its sill there (tests that it is permanent)
```js
globalThis.propertyIsEnumerable("Infinity")
```
and asserting that it is false (tests that it is non_enumerable)

HalidOdat:
Closing this in favor of #722

Issue #279, opened 2020-03-17
String interning


HalidOdat:
Will there be a string interner for more affective memory use and string comparisons.

Like [rustc](https://github.com/rust-lang/rust/blob/master/src/librustc_span/symbol.rs) does it.

HalidOdat:
More specifically [rustc interner](https://github.com/rust-lang/rust/blob/master/src/librustc_span/symbol.rs#L986-L1037).

Rustc Symbol doc:
```text
An interned string.

Internally, a `Symbol` is implemented as an index, and all operations
(including hashing, equality, and ordering) operate on that index. The use
of `rustc_index::newtype_index!` means that `Option<Symbol>` only takes up 4 bytes
...
```
Which makes comparisons between symbols trivial, it is just a comparison between two u32.

The cost is when interning a string, since it requires to do a HashMap lookup.
Getting the string from a symbol is trivial since it is only indexing into an array(Vec).

I found a implementation of this [string-interner](https://crates.io/crates/string-interner)

jasonwilliams:
Yep, this is interesting.
The plan was always to move to something more efficient but start with strings first. (just to make it easier for implementation against the spec)

I had a conversation the with Spidermonkey devs about how we use strings for internal slots, but later want to move to something more faster. I think they do something like this.

I don't know very much about https://crates.io/crates/string-interner so this will be new to me, but im happy to let this issue be the conversation point.

jasonwilliams:
https://matklad.github.io/2020/03/22/fast-simple-rust-interner.html

HalidOdat:
Nice! This is pretty mush how `rustc` does it. and how [string-interner](https://crates.io/crates/string-interner) crate is implemented.

Razican:
I think that the [`string-interner`][string_interner] crate is the way to go. It already provides a really nice, safe interner, and they are even discussing the optimisations laid out in the post you mentioned in https://github.com/Robbepop/string-interner/issues/16.

This could probably allow avoiding `String`s in `TokenKind`, which would make it `Copy` pretty easily, and reduce the memory footprint and improve the CPU cache usage.

[string_interner]: https://crates.io/crates/string-interner

Razican:
I will give this a look, and maybe #336 too, at the same time.

HalidOdat:
Hi @Razican 

Correct me if I'm wrong!!!

String interning only has a good effect if the string that's it is interning is supposed to be immutable, like identifiers.
[...]

PR #625, opened 2020-08-11
Moved value operations from `Interpreter` to `Value`

HalidOdat:
This PR moves the value operation function from `Interpreter` to `Value`. There operations don't make sense to live in interpreter for example:
```rust
context.to_primitive(value, PreferredType::Number);
```
When you read this line its _"context to primitive with value and preferred type number"_ this makes it seem like we are converting the context to a primitive, which is wrong! instead it should be:
```rust
value.to_primitive(context, PreferredType::Number);
```
When you read this line its _"value to primitive with context and preferred type number"_ this makes sense and allows us to remove the `#[allow(clippy::wrong_self_convention)]` 

Changes the following:
 - Moved `Intepreter::to_primitive()` => `Value::to_primitive()`
 - Moved `exec::PreferredType` to `value::PreferredType` 
 - Added `Value::display()` which implements `Display` (this is done because it conflicts with `Value::to_string(context)`)
 - Moved `Interpreter::to_string()` to `Value::to_string()`
 - Moved `Interpreter::to_object()` to `Value::to_object()`
 - Moved `Interpreter::to_property_key()` to `Value::to_property_key()`
 - Moved `Interpreter::to_numeric()` to `Value::to_numeric()`
 - Moved `Interpreter::to_uint32()` to `Value::to_uint32()`
 - Moved `Interpreter::to_int32()` to `Value::to_int32()`
 - Moved `Interpreter::to_index()` to `Value::to_index()`
 - Moved `Interpreter::to_length()` to `Value::to_length()`
 - Moved `Interpreter::to_integer()` to `Value::to_integer()`
 - Moved `Interpreter::to_number()` to `Value::to_number()`
 - Moved `Interpreter::to_number()` to `Value::to_number()`


boa/src/builtins/date/tests.rs
```diff
@@ -1,10 +1,12 @@
+#![allow(clippy::zero_prefixed_literal)]
```
Razican, line 1: Why do we need this?

boa/src/builtins/number/mod.rs
```diff
@@ -383,7 +383,11 @@ impl Number {
 
         // 2. If radix is undefined, let radixNumber be 10.
         // 3. Else, let radixNumber be ? ToInteger(radix).
-        let radix = args.get(0).map_or(10, |arg| arg.to_integer()) as u8;
+        let radix = args
+            .get(0)
+            .map(|arg| arg.to_integer(ctx))
+            .transpose()?
+            .unwrap_or(10.0) as u8;
```
Razican, line 390: I would change this to do a `map(|radix| radix as u8).unwrap_or(10);` That way we would avoid a conversion from `{float}` to `u8` in most cases.

boa/src/builtins/value/display.rs
```diff
@@ -1,5 +1,11 @@
 use super::*;
 
+/// This object is used for displaying a `Value`.
+#[derive(Debug, Clone, Copy)]
+pub struct ValueDisplay<'value> {
+    pub(super) value: &'value Value,
```
Razican, line 6: I like this idea. In the future, this will also need a `ctx` or something with an interner, probably.

Why does it need to be `pub(super)`?

boa/src/builtins/date/tests.rs
HalidOdat: This was generating `145` errors in my IDE, Many `Date` tests are as follows `IsDate(2020, 07, 02)` the include a `0` prefix in months and days.

boa/src/builtins/value/display.rs
HalidOdat: > Why does it need to be `pub(super)`?

Its only exposed to `Value`.

> In the future, this will also need a `ctx` or something with an interner, probably.

why would this need `ctx` this displays the `Value` without context for `Display` trait, also why interner>

HalidOdat: In the future we could add a `.display_with_settings()`
```rust
struct DispplaySettings {
	use_double_quote: bool,
	indent: String,
}
```
that would the Value with different setting such as `''` instead of `"` use different indent (or specify a tab), etc.

Razican: > > Why does it need to be `pub(super)`?
> 
> Its only exposed to `Value`.

Is it used?

> > In the future, this will also need a `ctx` or something with an interner, probably.
> 
> why would this need `ctx`, this displays the `Value` without context for `Display` trait, also why interner?

If we start interning immutable strings, we might use them in `Value`, right? At least we should use them in `TokenKind` for sure. So when displaying them, we might need the interner to resolve the indexes.

HalidOdat: > Is it used?

Yes, since it is defined in `value/display.rs` we can't do `ValueDispaly { value: self }` in `value/mod.rs`

> If we start interning immutable strings, we might use them in `Value`, right? At least we should use them in `TokenKind` for sure. So when displaying them, we might need the interner to resolve the indexes.

Yes. the interner would be used for `TokenKind` but only for identifiers, and `Value` would not know what a interned string is. If we use it for stings with have the same issue: https://github.com/boa-dev/boa/issues/279#issuecomment-621945984

HalidOdat:
If everything is ok with this PR could we merge this, it's blocking me on some PRs

boa/src/builtins/value/display.rs
Razican: > > Is it used?
> 
> Yes, since it is defined in `value/display.rs` we can't do `ValueDispaly { value: self }` in `value/mod.rs`

Hmmm not sure if we should instead be using a `new()` function in `ValueDisplay` if we only need it public for creation.

The PR can be merged as is, see if it makes sense to change that, but from my side, it's good :)

HalidOdat: > Hmmm not sure if we should instead be using a `new()` function in `ValueDisplay` if we only need it public for creation.

Since the only way `ValueDisplay` can be created is with `.display()` and we don't construct it anywere and it's not public for construction, I not see the reason why it would need a `new` constructor. the standard library does the same.

HalidOdat:
Lets merge this merging this :)

Issue #445, opened 2020-06-02
New Public API


Razican:
Related to #440.

**Current situation**
The current public API of Boa, that can be found [here][boa-docs] for Boa 0.8.0 and can be seen offline running `cargo doc --open` for the `master` branch, is far from optimal. We have 3 global functions:
 - `exec()`: That creates a new lexer, parser, Realm and interpreter, and isolates the execution of a given JavaScript string, and return a string with the result.
 - `forward()`: That will receive an already initialized interpreter and a JavaScript string and will lex, parse and execute this code in the initialized interpreter, and return the result as a string.
 - `forward_val()`: That will work the same way as `forward()`  but will return a `ResultValue`, which will indicate if the execution was successful or not.

Then, we have the following modules public:
 - `syntax`: which includes:
    - `ast`: the AST implementation (nodes, constants, keywords, builtin types, positions, spans, tokens, operations, punctuators...), with some re-exports to make life easier.
    - `lexer`: The tokenizer/lexer implementation, that turns a string into a vector of tokens. (this is changing in #432)
    - `parser`: The parser implementation, with its error and a couple of parsers (`Script` and `ScriptBody`), which turns a vector of tokens into a `StatementList`, that can then be executed with an interpreter.
 - `realm`, which includes the implementation of a `Realm`, an interesting helper function to register a global function, and makes public the global object, environment, and the lexical environment. This in theory allows anyone to add stuff to the global environment, but it's a bit cumbersome.
 - `builtins`: Where we have the builtin objects such as `Array`, `Number`, `JSON` and more, and also the implementations of `Value` and object properties. And the initialization function.
 - `environment`, which has modules for each of the environments and the global trait, everything exposed because the `realm` has public members so that they can be modified.
 - `exec`: We only contain the `Interpreter`, with just two public functions: `new()` and `to_string()`. The first creates a new interpreter with a given `Realm`, and the second one converts a `Value` to a string. It also contains the `Executable` trait, that allows calling `run()` on any AST node passing the interpreter so that it runs it.

Finally, we also have some re-exports to make things easier to use.

[boa-docs]: https://docs.rs/Boa/0.8.0/boa/

**Problem description**
When a user wants to use Boa for their projects, I can imagine the following use cases:
 - They want to interpret JavaScript from a file or a stored String and get the result.
 - They want to interpret JavaScript that is coming from a socket with a `Read` interface and get the result by streaming the bytes.
[...]

PR #626, opened 2020-08-13
Refactor `Function`

HalidOdat:
This PR is a preparation for the function design discussed in #445 

It changes the following:
 - Moved `Function::call()` => `GcObject::call()` this is done because we need the object that is being called to be garbage collected, because we put it in the environment. This was awkward before since we pass a clone of object and the access the `Function` and call it and pass the function object. 
 - Moved `Function::construct()` => `GcObject::construct()` same as `Function::call()`
 - Made `Function` be an enum since many fields that are for ordinary functions where there for builtins.
 - Removed `InternalState` since we will replace it with a `NativeObject`.
 - Put `RegExp` in box because it was really big (136 bytes)
 - Added `Interpreter::create_builtin_function(name, length, function pointer) -> GcObject`
 - Added `Interpreter::register_global_function(name, length, function pointer)`
 - Shrink `Function` from `56` bytes to `48` bytes
 - Shrink `ObjectData` from `144` bytes to `80` bytes
 - Shrink `Object` from `272` bytes to `192` bytes


Razican:
If I'm not mistaken, this would de facto close #387 and give a nice solution to https://github.com/boa-dev/boa/issues/363#issuecomment-626170196, right?

HalidOdat:
> If I'm not mistaken, this would de facto close #387 and give a nice solution to [#363 (comment)](https://github.com/boa-dev/boa/issues/363#issuecomment-626170196), right?

I guess so, forgot there was an PR to fix it. This PR only removes it but there will also be PR to add `NativeObject` after some discussion. I have a proof of concept locally. I'll create a PR to see and discuss on it.

jasonwilliams:
Looks good so far from what i can see

HalidOdat:
> Im guessing this and the NativeObject will need to go in together?

No. Not necessarily they accomplish different goals. This PR does some prepwork for the new function design, and #627 (`NativeObject`) introduces the native objects

boa/src/builtins/function/mod.rs
```diff
@@ -18,84 +18,43 @@ use crate::{
...
     syntax::ast::node::{FormalParameter, StatementList},
     BoaProfiler, Result,
 };
 use bitflags::bitflags;
 use gc::{unsafe_empty_trace, Finalize, Trace};
 use std::fmt::{self, Debug};
 
-/// _fn(this, arguments, ctx) -> Result<Value>_ - The signature of a built-in function
-pub type NativeFunctionData = fn(&Value, &[Value], &mut Interpreter) -> Result<Value>;
+/// _fn(this, arguments, ctx) -> ResultValue_ - The signature of a built-in function
```
Razican, line 30: I think this comment needs to be updated back again.
```suggestion
/// _fn(this, arguments, ctx) -> Result<Value>_ - The signature of a built-in function
```

Issue #446, opened 2020-06-02
Add examples


Razican:
**Current situation**
We currently have no code examples to use the API externally both in the API documentation and in the nonexistent `examples` folder.

**What do we need?**
As mentioned in #445, we should be creating a new API, but in any case, we would need to create examples for some common use cases of Boa. I'm extracting here part of that discussion:

> When a user wants to use Boa for their projects, I can imagine the following use cases:
>  - They want to interpret JavaScript from a file or a stored String and get the result.
>  - They want to interpret JavaScript that is coming from a socket with a `Read` interface and get the result by streaming the bytes.
>  - They want to extend our engine by providing new global functions and objects, so that with the JavaScript they can either perform more rich things in their setup (for example by providing an API for a browser window), or because they want to add JavaScript helpers for common actions in their setup. This also includes things as being the scripting language of a complex platform.
>  - They want to change the `global` object, for example, because they want it to be the `window` or a `frame`.

I think we need at least one example of each, that gets tested on each commit, in the `examples` directory of the `boa` crate. Then, I would add parts of those examples to the library-level documentation, in order to make life easier for our users.

elasmojs:
@Razican  I can help with this.

Razican:
> @Razican I can help with this.

Go ahead! Thanks :)

elasmojs:
@Razican the examples are completed except for point 4, i am not sure if I have figured that out. I will raise a pull request now

jasonwilliams:
@Razican should this be on hold until we know what the public API will look like?

Razican:
I think so, yes. But the public API is already pretty good, so some examples can be added.

Issue #551, opened 2020-07-08
Implement an easily replaceable Printer for `console`


Razican:
**Current situation**
We currently have the [`Console`](https://console.spec.whatwg.org/) standard implemented, but we don't implement it with the common [`Logger`](https://console.spec.whatwg.org/#logger), [`Printer`](https://console.spec.whatwg.org/#printer) and [`Formatter`](https://console.spec.whatwg.org/#formatter) operations. This prevents crates using Boa from creating their own printers that will for example not print anything but save the result to a file or a database.

**Proposed feature**
We should implement an easily replaceable `Printer` that would either be passed to the Realm / Interpreter, so that by default, the `console` will work as it does now, but we should have a way of creating the `console` object to use a custom `Printer`.

**How do we do this?**
We should have a `printer` object that implements the `Printer` trait, which would be something like:
```rust
trait Printer {
    fn print(&mut self, logLevel: LogLevel, args: &[Value]);
}
```

Maybe the trait needs to be modified, and we would need a `LogLevel` enum, but I think the implementation should go in this direction. The idea behind having the logger use `&mut self` is because we might want a printer that stores the logs in memory, or in a buffer, or that uses `Write` to write to a log file.

What I'm not very sure is on what the `options` should be in the `Printer`. Maybe we can just skip that argument for now.

Issue #440, opened 2020-06-01
Boa prelude (A convenience module that re-exports the most commonly-used Boa APIs)


HalidOdat:
#### What would the prelude contain?
Maybe `Interpreter`, `Value`, what else?

#### How to use (just an example)
```rust
use boa::prelude::*; // This would import all the commonly-used things

fn main() {
	let number = Value::number(3.1415);
	
	let realm = Realm::create();
	let mut engine = Interpreter::new(realm);
	let result = forward(&mut engine, "Math.PI");
}
```

jasonwilliams:
Would be useful for tests, so we could start with that as a use case.

PR #627, opened 2020-08-13
Feature native class objects (`NativeObject` and `Class` traits)

HalidOdat:
This is related to #445 and #446 

It changes the following:
 - Added `NativeObject` trait
 - If an type implements `Debug`, `Any` + `Trace` it automatically is an `NativeObject`
 - It calls `Trace` trait so this is not unsound
 - Removed `internal state`
 - Added example
 - Added `Class` for native class creation. see example in `boa/examples`.

### DIFF
--- a/boa/examples/classes.rs
+++ b/boa/examples/classes.rs
@@ -1,0 +1,84 @@
+use boa::builtins::object::{Class, ClassBuilder};
+use boa::builtins::value::*;
+use boa::exec::*;
+use boa::realm::Realm;
+use boa::*;
+
+use gc::{Finalize, Trace};
+
+#[derive(Debug, Trace, Finalize)]
+struct Person {
+    name: String,
+    age: u32,
+}
+
+impl Person {
+    fn say_hello(this: &Value, _: &[Value], ctx: &mut Interpreter) -> Result<Value> {
+        if let Some(object) = this.as_object() {
+            if let Some(person) = object.downcast_ref::<Person>() {
+                println!(
+                    "Hello my name is {}, I'm {} years old",
+                    person.name, person.age
+                );
+                return Ok(Value::undefined());
+            }
+        }
+        ctx.throw_type_error("'this' is not a Person object")
+    }
+}
+
+impl Class for Person {
+    const NAME: &'static str = "Person";
+    const LENGTH: usize = 2;
+
+    fn constructor(_this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result<Self> {
+        let name = args.get(0).cloned().unwrap_or_default().to_string(ctx)?;
+        let age = args.get(1).cloned().unwrap_or_default().to_u32(ctx)?;
+
+        let person = Person {
+            name: name.to_string(),
+            age,
+        };
+
+        Ok(person)
+    }
+
+    fn methods(class: &mut ClassBuilder) -> Result<()> {
+        class.method("sayHello", 0, Self::say_hello);
+        class.static_method("is", 1, |_this, args, _ctx| {
+            if let Some(arg) = args.get(0) {
+                if let Some(object) = arg.as_object() {
+                    if object.is::<Person>() {
+                        return Ok(true.into());
+                    }
+                }
+            }
+            Ok(false.into())
+        });
+
+        Ok(())
+    }
+}
+
+fn main() {
+    let realm = Realm::create();
+    let mut context = Interpreter::new(realm);
+
+    context.register_global_class::<Person>().unwrap();
+
+    forward_val(
+        &mut context,
+        r"
+		let person = new Person('John', 19);
+		person.sayHello();
+
+		if (Person.is(person)) {
+			console.log('person is a Person class instance.');
+		}
+		if (!Person.is('Hello')) {
+			console.log('\'Hello\' string is not a Person class instance.');
+		}
+	",
+    )
+    .unwrap();
+}
--- a/boa/examples/classes.rs
+++ b/boa/examples/classes.rs
@@ -1,8 +1,13 @@
-use boa::builtins::object::{Class, ClassBuilder};
-use boa::builtins::value::*;
-use boa::exec::*;
-use boa::realm::Realm;
-use boa::*;
+use boa::{
+    builtins::{
+        object::{Class, ClassBuilder},
+        value::Value,
+    },
+    exec::Interpreter,
+    forward_val,
+    realm::Realm,
+    Result,
+};
 
 use gc::{Finalize, Trace};
 
--- a/boa/examples/classes.rs
+++ b/boa/examples/classes.rs
@@ -84,7 +84,7 @@ impl Class for Person {
     }
 
     /// This is where the object is intitialized.
-    fn methods(class: &mut ClassBuilder) -> Result<()> {
+    fn init(class: &mut ClassBuilder) -> Result<()> {
         // we add a inheritable method `sayHello` with length `0` the amount of args it takes.
         //
         // This function is added to `Person.prototype.sayHello()`
--- a/boa/examples/classes.rs
+++ b/boa/examples/classes.rs
@@ -12,56 +12,105 @@ use boa::{
 
 use gc::{Finalize, Trace};
 
+// We create a new struct that is going to represent a person.
+//
+// We derive `Debug`, `Trace` and `Finalize`, It automatically implements `NativeObject`
+// so we can pass it an object in JavaScript.
+//
+// The fields of the sturct are not accesable by JavaScript unless accessors are created for them.
+/// This  Represents a Person.
 #[derive(Debug, Trace, Finalize)]
 struct Person {
+    /// The name of the person.
     name: String,
+    /// The age of the preson.
     age: u32,
 }
 
+// Here we implement a static method for Person that matches the `NativeFunction` signiture.
+//
+// NOTE: The function does not have to be implemented of Person it can be a free function,
+// or any function that matches that signature.
 impl Person {
+    /// This function says hello
     fn say_hello(this: &Value, _: &[Value], ctx: &mut Interpreter) -> Result<Value> {
+        // We check if this is an object.
         if let Some(object) = this.as_object() {
+            // If it is we downcast the type to type `Person`.
             if let Some(person) = object.downcast_ref::<Person>() {
+                // we print the message to stdout.
                 println!(
                     "Hello my name is {}, I'm {} years old",
-                    person.name, person.age
+                    person.name,
+                    person.age // Here we can access the native rust fields of Person struct.
                 );
                 return Ok(Value::undefined());
             }
         }
+        // If `this` was not an object or the type was not an native object `Person`,
+        // we throw a `TypeError`.
         ctx.throw_type_error("'this' is not a Person object")
     }
 }
 
 impl Class for Person {
+    // we set the binging name of this function to be `"Person"`.
+    // It does not have to be `"Person"` it can be any string.
     const NAME: &'static str = "Person";
+    // We set the length to `2` since we accept 2 arguments in the constructor.
+    //
+    // This is the same as `Object.length`.
+    // NOTE: If this is not defiend that the default is `0`.
     const LENGTH: usize = 2;
 
+    // This is what is called when we do `new Person()`
     fn constructor(_this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result<Self> {
+        // we get the first arguemnt of undefined if the first one is unavalable and call `to_string`.
+        //
+        // This is equivalent to `String(arg)`.
         let name = args.get(0).cloned().unwrap_or_default().to_string(ctx)?;
+        // we get the second arguemnt of undefined if the first one is unavalable and call `to_u32`.
+        //
+        // This is equivalent to `arg | 0`.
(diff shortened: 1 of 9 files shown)


Razican:
> What do you think? @Razican

This looks great! But we need something to add methods to the object, constructors and so on. Maybe we could have some `call()` methods in the trait? they would receive a context and a method name, and then the trait implementor does its thing.

HalidOdat:
> > What do you think? @Razican
> 
> This looks great! But we need something to add methods to the object, constructors and so on. Maybe we could have some `call()` methods in the trait? they would receive a context and a method name, and then the trait implementor does its thing.

Yep. right now it's very minimal! It would be nice to have the context setup first.

HalidOdat:
> But we need something to add methods to the object, constructors and so on. Maybe we could have some `call()` methods in the trait? they would receive a context and a method name, and then the trait implementor does its thing.

About this. This is more of a class than an object, so `NativeObject` trait would allow passing the object around. For the constructor and methods we need another trait I called it `Class` with the following fields:
```rust
pub trait Class: NativeObject {
	const NAME: &'static str; // This is the binging name of the class
	const LENGTH: usize = 0; // Class `length` (the amount of arguments it takes) default is 0

	/// The constructor of the class
	fn constructor(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue;
	/// Initialize class methods
	fn methods(class: &mut ClassBuilder<'a>) -> ResultValue;
}
```

Now with [`0e93042` (#627)](https://github.com/boa-dev/boa/pull/627/commits/0e93042ab67b46dbd74efe2e7a9740dee5e42de5) (latest commit) we can do this:
```rust
struct Person {
    name: String,
    age: u32,
}

impl Person {
    fn say_hello(this: &Value, _: &[Value], ctx: &mut Interpreter) -> ResultValue {
        if let Some(object) = this.as_object() {
            if let Some(person) = object.downcast_ref::<Person>() {
                println!(
                    "Hello my name is {}, I'm {} years old",
                    person.name, person.age
                );
                return Ok(Value::undefined());
            }
        }
        ctx.throw_type_error("'this' is not a Person object")
    }
}

impl Class for Person {
    const NAME: &'static str = "Person";
    const LENGTH: usize = 2;

    fn constructor(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
        let name = ctx.to_string(&args.get(0).cloned().unwrap_or_default())?;
        let age = ctx.to_uint32(&args.get(1).cloned().unwrap_or_default())?;

        let person = Person {
            name: name.to_string(),
            age,
        };

        this.set_data(ObjectData::NativeObject(Box::new(person)));
    	Ok(this.clone())
    }

    fn methods(class: &mut ClassBuilder) -> ResultValue {
        class.method("sayHello", 0, Self::say_hello);
        class.static_method("iAmAStaticMethod", 0, |_, _, _| {
        	println!("Hi o/");
        	Ok(Value::undefined())
        });

        Ok(Value::undefined())
    }
}

fn main() {
    let realm = Realm::create();
    let mut context = Interpreter::new(realm);

    context.register_global_class::<Person>().unwrap();
    forward(
        &mut context,
        r"
		let person = new Person('John', 19);
		person.sayHello();
		Person.iAmAStaticMethod();
	");
}
```

What do you think? @Razican @jasonwilliams @Lan2u

Razican:
> What do you think? @Razican @jasonwilliams @Lan2u

I love this! It's very ergonomic and easy to use :)

About the API itself, I just have these of questions:
 - Should the API for `register_global_class()` use static or dynamic dispatch? The static dispatch might add too much code duplication that add some cache misses, but I don't think it will be relevant (this is usually only called when setting up the environment).
 - Do all classes/objects have a constructor? (didn't read the spec here)
 - What about the difference on things like `new String()` and `String()`?
 - I think we shouldn't use `println!()` to print stuff in the methods. We should add a [`Logger`](https://console.spec.whatwg.org/#logger) in the context to do things such as `console.log()` or `console.info()`. This `Logger` should be customizable by API users in case they want to provide a logger that writes to a log file or a database (instead to stdout) - #551. Probably not directly related to this PR, but something to have in mind.
 - If we would like to offer a C/C++ API, how could we wrap this?

HalidOdat:
>     * Should the API for `register_global_class()` use static or dynamic dispatch? The static dispatch might add too much code duplication that add some cache misses, but I don't think it will be relevant (this is usually only called when setting up the environment).

I don't think this is going to be an issue either, I can make `ClassBuilder::new` a non-generic since that is the biggest function and all the other generic functions are small.

>     * Do all classes/objects have a constructor? (didn't read the spec here)

Did not read the spec, but the constructor is what is called when we do `new Object()` so there has to be a function that is called ( in `builins/string/mod.rs` it's `make_string` this is the constructor when we do `new String()`)

>     * What about the difference on things like `new String()` and `String()`?

As far as I know class constructors always we require `new` keyword:
```js
class X {}
x() // Uncaught TypeError: class constructors must be invoked with 'new'
```
I could add a `CALLABLE` to `Class` trait.
```rust
pub trait Class: NativeObject {
	const CALLABLE: bool = false; // this would default to false to match with javascript class.
}
```
we could have the same for `CONSTRUCTABLE`

>     * If we would like to offer a C/C++ API, how could we wrap this?

I'm not that experienced with rust to C ffi. but this class is a wrapper around what we do in for example `Array` (calling `make_constructor_fn` and `make_builtin_fn`) we could expose those functions to C.

Razican:
> > Do all classes/objects have a constructor? (didn't read the spec here)
> 
> Did not read the spec either smile , but the constructor is what is called when we do `new Object()` so there has to be a function that is called ( in `builins/string/mod.rs` it's `make_string` this is the constructor when we do `new String()`)
> 
> > What about the difference on things like `new String()` and `String()`?
> 
> As far as I know class constructors always require `new` keyword:
> 
> ```js
> class X {}
> x() // Uncaught TypeError: class constructors must be invoked with 'new'
> ```
> 
> I could add a `CALLABLE` to `Class` trait.
> 
> ```rust
> pub trait Class: NativeObject {
> 	const CALLABLE: bool = false; // this would default to false to match with javascript class.
> }
> ```
> 
> we could have the same for `CONSTRUCTABLE`

I think @jasonwilliams can give us some insight here, since he worked with this at some point.

HalidOdat:
Now in the constructor, instead of this:
```rust
fn constructor(this: &Value, args: &[Value], ctx: &mut Interpreter) -> ResultValue {
    let name = ctx.to_string(&args.get(0).cloned().unwrap_or_default())?;
    let age = ctx.to_uint32(&args.get(1).cloned().unwrap_or_default())?;

    let person = Person {
        name: name.to_string(),
 	    age,
    };

    this.set_data(ObjectData::NativeObject(Box::new(person)));
  	Ok(this.clone())
}
```
Instead of having to manually call `.set_data()` (if we don't this is a bug and can cause a serious headache) not it is enfforeced by the type system. (Result<Self, Value>):
```
// _this is still avalable if we want to set properties (string, symbol) of the class instance.
fn constructor(_this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result<Self, Value> {
    let name = ctx.to_string(&args.get(0).cloned().unwrap_or_default())?;
    let age = ctx.to_uint32(&args.get(1).cloned().unwrap_or_default())?;

    let person = Person {
        name: name.to_string(),
 	    age,
    };
  	Ok(person)
}
```

Razican:
> what do you think?

Much better! This also allows us to change the `set_data()` API in the future without it being a breaking change :)

jasonwilliams:
> Do all classes/objects have a constructor? (didn't read the spec here)

No, constructors and objects are completely separate. A constructor is just a function that’s given an object (this) to then populate and return. 
Usually constructors just set the prototype of the object.

You can have objects from scratch with no constructor at all.
Classes are just syntactic sugar so they behave the same.

This is where things get interesting!
The constructor and the function are the same! I thought they were 2 separate implementations but they’re the same function.
The way it works is that the function sets the correct value to `this` but also returns a value.
Let’s use [`String`](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/string/mod.rs#L70) as an example..

For ‘new String()’:  `this` is provided and the function sets `this` to be the string instance and that is used ( the return value is ignored)

For ‘String()’: the return value is used and returned to the assignment, all of the `this` stuff is ignored.

boa/src/builtins/property/attribute/mod.rs
```diff
@@ -20,6 +20,9 @@ bitflags! {
         /// None of the flags are present.
         const NONE = 0b0000_0000;
 
+        /// All the flags set (`WRITABLE`, `ENUMERABLE`, `CONFIGURABLE`).
+        const ALL = 0b0011_1111;
+
```
Razican, line 25: If I'm not mistaken, we can use the `Attribute::empty()` and `Attribute::all()` for this, right? So we don't need these two.

[More information](https://docs.rs/bitflags/1.2.1/bitflags/example_generated/struct.Flags.html)

boa/examples/classes.rs
```diff
@@ -0,0 +1,84 @@
+use boa::builtins::object::{Class, ClassBuilder};
+use boa::builtins::value::*;
+use boa::exec::*;
+use boa::realm::Realm;
+use boa::*;
```
Razican, line 5: I don't think we should use `*` imports. Also, I think it is easier to understand if we indent it:

```suggestion
use boa::{
    builtins::{
        value::*,
        object::{Class, ClassBuilder}
    },
    exec::*,
    realm::Realm,
    *,
};
```

Razican (commented):
It's looking pretty good :) Do you think we should use this for our own internal objects?

boa/src/builtins/object/internal_methods.rs
```diff
@@ -298,45 +296,45 @@ impl Object {
...
+    //                 .as_object()
+    //                 .expect("prototype should be null or object")
+    //                 .prototype
+    //                 .clone();
+    //             p = prototype;
+    //         }
+    //     }
+    //     self.prototype = val;
+    //     true
+    // }
```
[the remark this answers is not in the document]
HalidOdat: It's not used anywhere, its fixed in #600 so I left it as if for now.

HalidOdat:
> It's looking pretty good :) Do you think we should use this for our own internal objects?

It is is possible just needs some changes also some missing things are the ability to add static properties and prototype properties

Razican:
> > It's looking pretty good :) Do you think we should use this for our own internal objects?
> 
> It is is possible just needs some changes also some missing things are the ability to add static properties and prototype properties,

In that case, I think we should leave that for a future PR and merge this as soon as you feel it's ready :)

boa/examples/classes.rs
HalidOdat: > I don't think we should use `*` imports. Also, I think it is easier to understand if we indent it:

This is where the `prelude` #440 would be really nice.

HalidOdat:
Added the ability to add inherited and static properties as discussed previously.

boa/src/builtins/object/mod.rs
```diff
@@ -45,8 +48,214 @@ mod tests;
...
+}
+
+/// Native class.
+pub trait Class: NativeObject + Sized {
+    /// The binding name of the object.
+    const NAME: &'static str;
+    /// The amount of arguments the class `constructor` takes, default is `0`.
+    const LENGTH: usize = 0;
+    /// The attibutes the class will be binded with, default is `writable`, `enumerable`, `configurable`.
+    const ATTRIBUTE: Attribute = Attribute::all();
```
Razican, line 76: Should this be `Attribute::default()`?

Razican (approved):
Looks good from my side, just a comment for my understanding :)

boa/src/builtins/object/mod.rs
HalidOdat: The `Attribute::default()` is readonly non_enumerable and permanent to match with the specs default.

boa/src/builtins/object/mod.rs
```diff
@@ -45,8 +48,214 @@ mod tests;
...
+
+    /// The constructor of the class.
+    fn constructor(this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result<Self>;
+
+    /// Initializes the internals and the methods of the class.
+    fn methods(class: &mut ClassBuilder<'_>) -> Result<()>;
```
HalidOdat, line 82: Should this method be called `init` instead of `methods` since its not just the methods that are being initialized.

Razican: It makes sense, yes!

boa/src/builtins/object/mod.rs
```diff
@@ -45,8 +48,214 @@ mod tests;
 /// Static `prototype`, usually set on constructors as a key to point to their respective prototype object.
 pub static PROTOTYPE: &str = "prototype";
 
+/// This trait allows Rust types to be passed around as objects.
+///
+/// This is automatically implemented, when a type implements `Debug`, `Any` and `Trace`.
+pub trait NativeObject: Debug + Any + Trace {
+    fn as_any(&self) -> &dyn Any;
+    fn as_mut_any(&mut self) -> &mut dyn Any;
```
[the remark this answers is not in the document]
Razican: If it's a public trait, I think it should be fully documented, right?

PR #629, opened 2020-08-13
Fix: String.prototype.replace substitutions, closes #610

### LINKED ISSUE #610: Error in RegExp
Hi,

The following code snippet works incorrectly for RegExp

```js
let re = /(\w+)\s(\w+)/
let str = 'John Smith'
let newstr = str.replace(re, '$2, $1')
console.log(newstr)
```

Expected: `Smith, John`
Got: `Smith, $1`

thanks

RageKnify:
This Pull Request fixes/closes #610 .

It changes the following:
 - Change the substitution logic in String.prototype.replace so that it more closely follows: https://tc39.es/ecma262/#sec-getsubstitution


boa/src/builtins/string/mod.rs
```diff
@@ -486,7 +486,7 @@ impl String {
                     }
 
                     // Capture $1, $2, $3 etc
-                    if re.is_match(&result) {
+                    while re.is_match(&result) {
```
Razican, line 489: I think this adds too much iteration, when we have the [`replace_all()`](https://docs.rs/regex/1.3.9/regex/struct.Regex.html#method.replace_all) method in `Regex`. I think we should still use an `if`, here, but in line 494, use `replace_all()` instead.

RageKnify: I think we'd still need `while` though, `replace_all()` will only work for all ocurrences of `$x`. `$y` will need to be replaced in a different iteration, no?
I think the code block might need to be re-ordered, I think `$$1` will become `$1` and then be replaced by the capture group when it should just be `$1`.

Razican: Hmmm I guess we need to iterate through the captures, but there should be only one check for the match, since the algorithm is not recursive. You would have the issue you mention

RageKnify:
I'm not gonna be working for a week, but I found what I think was an old implementation of the method from V8 in js.
The main change is that we need to go through the replacement string left to right copying characters to the result string and for `$` we check if it is part of one of the patterns described in the table (there's a link to the spec in a nearby comment), and if so replace it accordingly.

From a quick glance I think we need to invoke `chars` on the result string to iterate over them and then peek ahead when we spot `$`?

RageKnify:
I ended up not fixing the other bug, my intention was to follow the spec, but it requires `RegExp.prototype [ @@replace ]` to be implemented for invocations of `String.prototype.replace` to work and right now I don't believe it is defined/implemented I think we have an approximation of the implementation here in `String.prototype.replace`.

I think `symbol_properties` are working fine, we're just missing the implementation of the function and the initialization of it at startup.

Another thing to keep in mind is `regress`, which was mentioned in #478 I believe `regress`'s API is lacking right now but it would allow `boa` to not worry about implementing RegEx. Currently `regress` isn't very active, but we could try to get it going, probably try to identify which methods we would need and go from there.

RageKnify:
When I get back I'll implement tests for `$$`, `$&`, `$``, `$'`, `$nn`.

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +479,90 @@ impl String {
...
+                            let third_is_digit = third.map(|ch| ch.is_digit(10)).unwrap_or(false);
+
+                            match (second, third) {
+                                (Some('$'), _) => {
+                                    // $$
+                                    result.push('$');
+                                }
+                                (Some('&'), _) => {
+                                    // $&
+                                    let matched = caps.get(0).expect("cannot get matched value");
```
Razican, line 502: I would not panic here. We should return an error if this goes wrong, right? What does the spec say?

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +479,90 @@ impl String {
...
+
+                    let m = caps.len();
+
+                    while let Some(first) = chars.next() {
+                        if first == '$' {
+                            let second = chars.next();
+                            let second_is_digit = second.map(|ch| ch.is_digit(10)).unwrap_or(false);
+                            // we use peek so that it is still in the iterator if not used
+                            let third = if second_is_digit { chars.peek() } else { None };
+                            let third_is_digit = third.map(|ch| ch.is_digit(10)).unwrap_or(false);
```
Razican, line 493: I see that this works for `$1` and for `$99`, but what about `$255`? Should that work too?

Razican (commented):
It looks pretty good :) check my comments, they are mostly for my understanding. Also, in the future, we will want to move to [`regress`](https://crates.io/crates/regress) for regular expressions, as in #478, and it might in fact prove to be easier to implement with it.

boa/src/builtins/string/mod.rs
RageKnify: It does not. But it's following the spec, the [table](https://tc39.es/ecma262/#table-45) only mentions 1 and 2 digit numbers.

boa/src/builtins/string/mod.rs
RageKnify: By following the spec we wouldn't get this far if there is no match, https://tc39.es/ecma262/#sec-string.prototype.replace point 9 returns the original string if there is no match.
I'm gonna take a look into how to adapt this.

RageKnify: https://github.com/RageKnify/boa/blob/a8222ced381fcab93a721323b94ff6b54427f061/boa/src/builtins/string/mod.rs#L471
It paniced earlier in the function when trying to get the match, I'll introduce a match statement on the option and a test covering this scenario.

Razican (approved):
This is good to know from my side now!

RageKnify:
@jasonwilliams @HalidOdat @Lan2u sorry to tag, but it's been waiting for a 2nd review for a week

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +482,90 @@ impl String {
...
```
HalidOdat, line 496:
```suggestion
                            let third_is_digit = third.map_or(false, |ch| ch.is_digit(10));
```
Calling `map_or` removes the call to `unwrap_or`

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +482,90 @@ impl String {
...
+                                (Some('&'), _) => {
+                                    // $&
+                                    let matched = caps.get(0).expect("cannot get matched value");
+                                    result.push_str(matched.as_str());
+                                }
+                                (Some('`'), _) => {
+                                    // $`
+                                    let start_of_match = mat.start();
+                                    let slice = &primitive_val[..start_of_match];
+                                    result.push_str(slice);
```
HalidOdat, line 512:
```suggestion
                                    result.push_str(&primitive_val);
```
This should work since `RcString` implements `Borrow` with type `&str` otherwise,
Calling `.as_str()` on `RcString` returns a `&str` this makes the code more readable

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +482,90 @@ impl String {
...
+                                }
+                                (Some('\''), _) => {
+                                    // $'
+                                    let end_of_match = mat.end();
+                                    let slice = &primitive_val[end_of_match..];
+                                    result.push_str(slice);
```
HalidOdat, line 518:
```suggestion
                                    result.push_str(&primitive_val);
```
Same here

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +482,90 @@ impl String {
...
+                                }
+                                (_, _) if second_is_digit && third_is_digit => {
+                                    // $nn
+                                    let tens = second.unwrap().to_digit(10).unwrap() as usize;
+                                    let units = third.unwrap().to_digit(10).unwrap() as usize;
+                                    let nn = 10 * tens + units;
```
HalidOdat, line 524:
```suggestion
                                (Some(second), Some(third)) if second_is_digit && third_is_digit => {
                                    // $nn
                                    let tens = second.to_digit(10).unwrap() as usize;
                                    let units = third.to_digit(10).unwrap() as usize;
                                    let nn = 10 * tens + units;
```
Since we know that `second` and `third` are `char`s (not `None`) we can eliminate and `unwrap`

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +482,90 @@ impl String {
...
+                                    }
+                                }
+                                (_, _) if second_is_digit => {
+                                    // $n
+                                    let n = second.unwrap().to_digit(10).unwrap() as usize;
+                                    if n == 0 || n > m {
+                                        [Some(first), second]
+                                            .iter()
+                                            .flatten()
+                                            .for_each(|ch: &char| result.push(*ch));
```
HalidOdat, line 546:
```suggestion
                                (Some(second), _) if second_is_digit => {
                                    // $n
                                    let n = second.to_digit(10).unwrap() as usize;
                                    if n == 0 || n > m {
                                        result.push(first);
                                        result.push(second);
```
Same here we can really simplefy the code and also eliminate a lot of unnesecary calls

HalidOdat (commented):
This looks good! check my comments to see on how we might improve it :) 

Sorry for the long delay I haven't had that much free time this week.

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +482,88 @@ impl String {
...
+                                {
+                                    // $nn
+                                    let tens = second.to_digit(10).unwrap() as usize;
+                                    let units = third.to_digit(10).unwrap() as usize;
+                                    let nn = 10 * tens + units;
+                                    if nn == 0 || nn > m {
+                                        [Some(first), Some(second), chars.next()]
+                                            .iter()
+                                            .flatten()
+                                            .for_each(|ch: &char| result.push(*ch));
```
HalidOdat, line 529:
```suggestion
                                        result.push(first);
                                        result.push(second);
                                        if Some(ch) = chars.next() {
                                           result.push(ch))
                                        }
```
Doing it like this should be more readable and more  efficient

boa/src/builtins/string/mod.rs
```diff
@@ -479,39 +482,88 @@ impl String {
...
+                                (Some('<'), _) => {
+                                    // $<
+                                    todo!("named capture groups")
+                                }
+                                _ => {
+                                    // $?, ? is none of the above
+                                    // we can consume second because it isn't $
+                                    result.push(first);
+                                    second.iter().for_each(|ch: &char| result.push(*ch));
+                                }
```
HalidOdat, line 562:
```suggestion
                                (_, _) => {
                                    // $?, ? is none of the above
                                    // we can consume second because it isn't $
                                    result.push(first);
                                    if let Some(second) = second {
                                    	result.push(second);
                                    }
                                }
```
This is more readable, If second is garanteed to be a `char` (not `None`) we can do:
```rust
                                (Some(second), _) => {
                                    // $?, ? is none of the above
                                    // we can consume second because it isn't $
                                    result.push(first);
                                  	result.push(second);
                                }
```

HalidOdat (approved):

PR #630, opened 2020-08-14
Fix Stack Overflow for `Debug::fmt`

### LINKED ISSUE #608: `Debug::fmt` Causes Causes a Stack Overflow

jcdickinson:
This Pull Request fixes/closes #608.

It changes the following:
 - Adds RecursionLimiter. This tracks the `GcObject` pointers that have been visited during a `Debug::fmt` call.
 - Adds a test for the stack overflow.

The same object cannot appear twice in the output, by design. JS object graphs tend the be huge, and comprehending a massive data dump is nearly impossible. If the same object appears twice, even if not recursive, it is always elided. The following graph:

```js
let a = {};
a.c = a;
let b = {
 d: a, e: a
};
debug(b);
```

Would print:
```js
{
  d: {
    c: ... // Recursion protected
  },
  e: ... // Will never appear twice
}
```

### DIFF
--- a/boa/src/builtins/object/gcobject.rs
+++ b/boa/src/builtins/object/gcobject.rs
@@ -102,7 +102,8 @@ impl Drop for RecursionLimiter {
 
 impl RecursionLimiter {
     thread_local! {
-        pub static VISITED: RefCell<HashSet<usize>> = RefCell::new(HashSet::new());
+        /// The list of pointers to `GcObject` that have been visited during the current `Debug::fmt` graph.
+        static VISITED: RefCell<HashSet<usize>> = RefCell::new(HashSet::new());
     }
 
     /// Determines if the specified `GcObject` has been visited, and returns a struct that will free it when dropped.
--- a/boa/src/builtins/object/gcobject.rs
+++ b/boa/src/builtins/object/gcobject.rs
@@ -4,10 +4,14 @@
 
 use super::Object;
 use gc::{Finalize, Gc, GcCell, GcCellRef, GcCellRefMut, Trace};
-use std::fmt::{self, Display};
+use std::{
+    cell::RefCell,
+    collections::HashSet,
+    fmt::{self, Debug, Display},
+};
 
 /// Garbage collected `Object`.
-#[derive(Debug, Trace, Finalize, Clone)]
+#[derive(Trace, Finalize, Clone)]
 pub struct GcObject(Gc<GcCell<Object>>);
 
 impl GcObject {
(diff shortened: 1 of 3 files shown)


boa/src/builtins/object/gcobject.rs
```diff
@@ -61,6 +65,55 @@ impl Display for BorrowError {
...
+}
+
+impl Debug for GcObject {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::fmt::Result {
+        let limiter = RecursionLimiter::new(&self);
+
+        if limiter.first {
+            f.debug_tuple("GcObject").field(&self.0).finish()
+        } else {
+            f.write_str("...")
```
Razican, line 112: I think this makes it clear this is an object:
```suggestion
            f.write_str("{ ... }")
```

boa/src/builtins/object/gcobject.rs
```diff
@@ -61,6 +65,55 @@ impl Display for BorrowError {
     }
 }
 
+struct RecursionLimiter {
+    free: bool,
+    first: bool,
+}
```
Razican, line 71: Can we add some documentation here? And derive `Debug`, `Clone` and `Copy`, just in case.

boa/src/builtins/object/gcobject.rs
```diff
@@ -61,6 +65,55 @@ impl Display for BorrowError {
...
+
+impl RecursionLimiter {
+    thread_local! {
+        pub static VISITED: RefCell<HashSet<usize>> = RefCell::new(HashSet::new());
+    }
+
+    fn new(o: &GcObject) -> Self {
```
Razican, line 78: I would add a bit of documentation to this function, explaining what it does.

boa/src/builtins/object/gcobject.rs
jcdickinson: `Copy` can't be implemented because `RecursionLimiter` implements `Drop`.

boa/src/builtins/object/gcobject.rs
```diff
@@ -61,6 +65,75 @@ impl Display for BorrowError {
...
+        if self.free {
+            Self::VISITED.with(|hs| hs.borrow_mut().clear());
+        }
+    }
+}
+
+impl RecursionLimiter {
+    thread_local! {
+        pub static VISITED: RefCell<HashSet<usize>> = RefCell::new(HashSet::new());
+    }
```
Razican, line 106: I think we should add some documentation on what this is and how it works :)

jcdickinson:
Windows check failed with no logs, not sure if I can rerun it?

HalidOdat:
> Windows check failed with no logs, not sure if I can rerun it?

That is strange. Ill rerun it :)

PR #632, opened 2020-08-14
Fixes for the new lexer

Razican:
This is a first step towards fixing the new lexer. Parsers no longer care if they are skipping line terminators or not. Also, this has greatly reduced clones, and added some performance tricks in the
buffered lexer.

It's still failing 40 tests, but I think I can fix them during the weekend :)

### DIFF
--- a/boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
+++ b/boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
@@ -122,6 +122,10 @@ where
             self.read_index, self.write_index,
             "we reached the read index with the write index"
         );
+        debug_assert!(
+            self.read_index < PEEK_BUF_SIZE,
+            "read index went out of bounds"
+        );
 
         Ok(())
     }
@@ -140,10 +144,6 @@ where
         if self.read_index == self.write_index {
             self.fill()?;
         }
-        assert!(
-            self.read_index < PEEK_BUF_SIZE,
-            "read index went out of bounds"
-        );
 
         dbg!(&self.peeked, self.read_index, self.write_index);
         if let Some(ref token) = self.peeked[self.read_index] {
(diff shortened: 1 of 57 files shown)


Razican:
This is ready for merging on the new lexer branch. I was able to reduce performance overhead greatly. So now, instead of being up to x2.7 times slower, it's just up to x1.5 slower.

I was also able to reduce execution time somehow, I only modified the switch statement parsing and execution, and the Switch node, so that might have improved the memory representation of `Node`.

I also noticed some code duplication in `StatementList`, `Block` and `Switch` execution, that would be solved by implementing an executor for `&[Node]`, but that's something for a future PR.

After merging this, we need to squash all commits, rebase with master, and do a final review, but I think this is almost there!

As soon as we do that, we can rebase #567 and see if we can at least make it work without breaking the whole thing xD

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
```diff
@@ -58,152 +69,163 @@ where
...
-                            let src_index = (self.back_index + j) % PEEK_BUF_SIZE; // Src index for the swap.
-                            self.peeked[dst_index] = self.peeked[src_index].take();
-                            dst_index = src_index;
-                        }
-
-                        self.buf_size -= 1;
-                    }
+        if let Some(ref token) = self.peeked[self.read_index] {
+            let tok = if !skip_line_terminators || token.kind() != &TokenKind::LineTerminator {
+                mem::replace(&mut self.peeked[self.read_index], None)
```
Lan2u, line 150: Is mem::replace faster than swap?

boa/src/syntax/parser/statement/iteration/do_while_statement.rs
```diff
@@ -63,53 +63,31 @@ where
...
-        cursor.expect(Punctuator::CloseParen, "do while statement", true)?;
+        cursor.expect(Punctuator::CloseParen, "do while statement")?;
 
-        expect_semicolon_dowhile(cursor)?;
+        cursor.expect_semicolon(true, "do while statement")?;
 
         Ok(DoWhileLoop::new(body, cond))
     }
 }
-/// Checks that the next token is a semicolon with regards to the automatic semicolon insertion rules
```
Lan2u, line 94: Surely if this is a unique case to do_while it should stay within the do_while file?

Razican: Hmmm you are right, that would reduce a bit the overhead in the rest of the checks.

boa/src/syntax/parser/cursor/buffered_lexer/mod.rs
Razican: I need to check that. I'm not sure.

PR #636, opened 2020-08-15
Initial fix for const assignment panic

### LINKED ISSUE #622: Assignment to const results in panic
Hi

Assignment to const causes panic instead of expected TypeError

const number = 42;
    
    try {
        number = 99;
    } catch (err) {
        console.log(err);
        // expected output: TypeError: invalid assignment to const `number'
        // Note - error messages will vary depending on browser
    }
    
    console.log(number);
    // expected output: 42

thanks

54k1:
This Pull Request will fix #622.

This isn't complete in any sense. Just creating a draft PR so that I can get some comments on the approach.

I've defined `ErrorKind` in [environment/mod.rs](https://github.com/54k1/boa/blob/fix/622/boa/src/environment/mod.rs#L11). I would like some comments on how to go about propagating `Err`s.

I've added `ErrorKind` which gives a hint about the actual error. The error finally needs to be constructed using the [to_error](https://github.com/54k1/boa/blob/fix/622/boa/src/environment/mod.rs#L20) method by passing the interpreter. Due to this, at [several call sites](https://github.com/54k1/boa/blob/fix/622/boa/src/exec/declaration/mod.rs#L29) a match statement is required. I would like to know if there's a cleaner way to propagate the error at call sites.

@HalidOdat could you share your thoughts? Thanks!


boa/src/environment/declarative_environment_record.rs
```diff
@@ -56,9 +57,10 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
...
             },
         );
+        Ok(())
     }
 
-    fn create_immutable_binding(&mut self, name: String, strict: bool) -> bool {
+    fn create_immutable_binding(&mut self, name: String, strict: bool) -> Result<(), ErrorKind> {
         if self.env_rec.contains_key(&name) {
             // TODO: change this when error handling comes into play
             panic!("Identifier {} has already been declared", name);
```
HalidOdat, line 66: Why is this still a `panic`

boa/src/environment/declarative_environment_record.rs
```diff
@@ -73,32 +75,45 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
...
+        mut strict: bool,
+    ) -> Result<(), ErrorKind> {
         if self.env_rec.get(name).is_none() {
             if strict {
                 // TODO: change this when error handling comes into play
-                panic!("Reference Error: Cannot set mutable binding for {}", name);
+                return Err(ErrorKind::ReferenceError(format!(
+                    "Reference Error: Cannot set mutable binding for {}",
+                    name
+                )));
```
HalidOdat, line 111:
```suggestion
                return Err(ErrorKind::ReferenceError(format!(
                    "Cannot set mutable binding for {}",
                    name
                )));
```
Why is `RefecenceError` mentioned twice

boa/src/environment/declarative_environment_record.rs
```diff
@@ -73,32 +75,45 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
...
             }
 
             self.create_mutable_binding(name.to_owned(), true);
```
HalidOdat, line 114:
```suggestion
            self.create_mutable_binding(name.to_owned(), true)?;
```
The error should be propagated with `?` operator.

boa/src/environment/declarative_environment_record.rs
```diff
@@ -73,32 +75,45 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
...
             self.initialize_binding(name, value);
```
HalidOdat, line 115: same here:
```suggestion
            self.initialize_binding(name, value)?;
```

boa/src/environment/declarative_environment_record.rs
```diff
@@ -107,27 +122,35 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
         }
         if record.value.is_none() {
             // TODO: change this when error handling comes into play
-            panic!("Reference Error: Cannot set mutable binding for {}", name);
+            return Err(ErrorKind::ReferenceError(format!(
+                "Reference Error: Cannot set mutable binding for {}",
+                name
+            )));
```
HalidOdat, line 128:
```suggestion
            return Err(ErrorKind::ReferenceError(format!(
                "Cannot set mutable binding for {}",
                name
            )));
```

boa/src/environment/declarative_environment_record.rs
```diff
@@ -107,27 +122,35 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
...
 
         if record.mutable {
             record.value = Some(value);
         } else if strict {
             // TODO: change this when error handling comes into play
-            panic!("TypeError: Cannot mutate an immutable binding {}", name);
+            return Err(ErrorKind::TypeError(format!(
+                "TypeError: Cannot mutate an immutable binding {}",
+                name
+            )));
```
HalidOdat, line 138:
```suggestion
            return Err(ErrorKind::TypeError(format!(
                "Cannot mutate an immutable binding {}",
                name
            )));
```

boa/src/environment/declarative_environment_record.rs
```diff
@@ -107,27 +122,35 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
...
-                .clone()
+            if let Some(ref val) = binding.value {
+                Ok(val.clone())
+            } else {
+                Err(ErrorKind::ReferenceError("".to_string()))
+            }
         } else {
             // TODO: change this when error handling comes into play
-            panic!("ReferenceError: Cannot get binding value for {}", name);
+            panic!("Cannot get binding value for {}", name);
```
HalidOdat, line 153: Why is this still a panic?

boa/src/environment/function_environment_record.rs
```diff
@@ -102,24 +106,27 @@ impl EnvironmentRecordTrait for FunctionEnvironmentRecord {
...
             BindingStatus::Lexical => {
                 // TODO: change this when error handling comes into play
                 panic!("There is no this for a lexical function record");
             }
             BindingStatus::Uninitialized => {
                 // TODO: change this when error handling comes into play
-                panic!("Reference Error: Uninitialised binding for this function");
+                Err(ErrorKind::ReferenceError(format!(
+                    "Reference Error: Uninitialised binding for this function"
+                )))
```
HalidOdat, line 122:
```suggestion
                Err(ErrorKind::ReferenceError(format!(
                    "Uninitialised binding for this function"
                )))
```

boa/src/environment/function_environment_record.rs
```diff
@@ -135,62 +142,82 @@ impl EnvironmentRecordTrait for FunctionEnvironmentRecord {
...
-                .clone()
+            if let Some(ref val) = binding.value {
+                Ok(val.clone())
+            } else {
+                Err(ErrorKind::ReferenceError("".to_string()))
+            }
         } else {
             // TODO: change this when error handling comes into play
-            panic!("ReferenceError: Cannot get binding value for {}", name);
+            panic!("Cannot get binding value for {}", name);
```
HalidOdat, line 220: why is this still a panic

boa/src/environment/lexical_environment.rs
```diff
@@ -213,9 +224,17 @@ impl LexicalEnvironment {
...
             .find(|env| env.borrow().has_binding(name))
             .map(|env| env.borrow().get_binding_value(name, false))
+        {
+            Some(r) => match r {
+                Ok(v) => Some(v),
+                Err(_) => None,
+            },
+            None => None,
+        }
     }
```
HalidOdat, line 238: We should be returning `Result<Value, ErrorKind>` here

boa/src/environment/mod.rs
```diff
@@ -6,3 +6,31 @@ pub mod function_environment_record;
...
+
+#[macro_export]
+macro_rules! try_error {
+    ($e: expr) => {
+        match $e {
+            Err(e) => return Err(e.to_error(interpreter)),
+            _ => (),
+        }
+    };
+}
```
HalidOdat, line 36: why is it exported we don't use it out side of the crate

boa/src/environment/declarative_environment_record.rs
54k1: The [spec](https://tc39.es/ecma262/#sec-declarative-environment-records-getbindingvalue-n-s) calls assert (env_rec contains the binding).

boa/src/environment/function_environment_record.rs
54k1: Will fix these in next commit.

boa/src/environment/mod.rs
54k1: I shouldn't have added this. This was just a temporary thing.

boa/src/environment/declarative_environment_record.rs
54k1: Yes. Thanks for pointing out. Actually I've ignored the compiler warnings for now since I just wanted to know from you how to deal with errors in the `exec` module.

54k1: In exec, one possible way is to call the method on environment and further call `.or_else` on the Result type to map `ErrorKind` to an actual error object.

```Rust
env.create_mutable_binding(
    param.to_owned(),
    false,
    VariableScope::Block,
 ).or_else(|e| Err(e.to_error(interpreter)))?;
```

boa/src/environment/declarative_environment_record.rs
HalidOdat: > The [spec](https://tc39.es/ecma262/#sec-declarative-environment-records-getbindingvalue-n-s) calls `assert(env_rec contains the binding)`.

Hmmm. In that case leave it as a `panic!`

boa/src/environment/function_environment_record.rs
```diff
@@ -59,7 +60,7 @@ pub struct FunctionEnvironmentRecord {
 }
 
 impl FunctionEnvironmentRecord {
-    pub fn bind_this_value(&mut self, value: Value) {
+    pub fn bind_this_value(&mut self, value: Value) -> Result<Value, ErrorKind> {
```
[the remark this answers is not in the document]
54k1: The [spec](https://tc39.es/ecma262/#sec-bindthisvalue) tells to return the Value as well. Although in this case you might already have `value` in the scope to use again. What should we do in this case?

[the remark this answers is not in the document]
HalidOdat: > The [spec](https://tc39.es/ecma262/#sec-bindthisvalue) tells to return the `value` as well. Although in this case you will already have `value` in the scope to use again. What should we do in this case?

It's best to do as the spec does.

54k1:
@HalidOdat I think there's more work to do. I'll mark this PR as ready for review as soon as I'm done.

54k1:
I shall re-PR this. Too many bad commits.

PR #637, opened 2020-08-16
Feature `boa::Result<T>`

HalidOdat:
Related to #445 

This PRs removes the `ResultValue` and adds a generic type alias `Result<T>` where the `Err` is `Value`, this is common practice and is done in many places in the standard library, serde, etc (example `std::fmt::Result`, `std::io::Result`, `serde::Result`)

It changes the following:
 - Removed `ResultValue`.
 - Added type alias `type Result<T> = StdResult<T, Value>`


HalidOdat:

Razican:
It just needs a rebase :)

PR #640, opened 2020-08-16
Benchmark refactor

### LINKED ISSUE #639: Refactor benchmarks to pull from a module containing all constants.
I think we should reduce the code repetition in the benchmarking, by creating a module containing all of the strings and importing the pertinent variables where necessary. This should allow for easier addition of benchmarks in future development. This is really just a small change with no performance benefit, but just easier readability for benchmarks.

neeldug:
This Pull Request fixes/closes #639 .

It changes the following:
 - Moves all test scripts to separate module in `consts.rs`
 - Removes all `static definitions` from various benchmarks
 
Potentially could move ALLOC to consts.rs, however, am unsure about this.


neeldug:
Can't quite workout why this is failing now, any help would be greatly appreciated!

HalidOdat:
> Can't quite work out why this is failing now, any help would be greatly appreciated!

Hi @neeldug. The problem is that `benches/conts` is interpreted as benchmark, every file like `boa/examples/` is treated as a executable.

neeldug:
> > Can't quite work out why this is failing now, any help would be greatly appreciated!
> 
> Hi @neeldug. The problem is that `benches/conts` is interpreted as benchmark, every file like `boa/examples/` is treated as a executable.

Ahh I see, thanks, will try something else then.

neeldug:
> examples ones

But wouldn't this result in the code repetition that was initially the goal to avoid? In having the consts being separated by type?

Razican:
Something that can be done is to have one file per each, and one `constants.rs` file at the same level, and include it with [`include!()`](https://doc.rust-lang.org/stable/std/macro.include.html) at the top of each benchmark.

neeldug:
> Something that can be done is to have one file per each, and one `constants.rs` file at the same level, and include it with [`include!()`](https://doc.rust-lang.org/stable/std/macro.include.html) at the top of each benchmark.

So would you suggest having each static variable to be defined in a separate file, then using the include macro in a separate constants.rs, and then include those in each of the benchmarking files?

Razican:
> > Something that can be done is to have one file per each, and one `constants.rs` file at the same level, and include it with [`include!()`](https://doc.rust-lang.org/stable/std/macro.include.html) at the top of each benchmark.
> 
> So would you suggest having each static variable to be defined in a separate file, then using the include macro in a separate constants.rs, and then include those in each of the benchmarking files?

We should try that, yes.

neeldug:
> > > Something that can be done is to have one file per each, and one `constants.rs` file at the same level, and include it with [`include!()`](https://doc.rust-lang.org/stable/std/macro.include.html) at the top of each benchmark.
> > 
> > 
> > So would you suggest having each static variable to be defined in a separate file, then using the include macro in a separate constants.rs, and then include those in each of the benchmarking files?
> 
> We should try that, yes.

So is this what you meant? Or also using include macros in each of the benchmarking files? Because this seemingly didn't work.

neeldug:
Latest commit is fully functional, however, since it no longer uses `use`, IDE linking to definitions is lost, this is a minor issue. Although, it's also meaning that every variable is being imported into every benchmark, which also a minor issue.

Razican:
> Latest commit is fully functional, however, since it no longer uses `use`, IDE linking to definitions is lost, this is a minor issue. Although, it's also meaning that every variable is being imported into every benchmark, which also a minor issue.

I had thought of using one file for all, but what you did is even better, and it brings some interesting posibilities. We can use JavaScript (`.js`) files for each (instead of `.rs`), and use normal JavaScript syntax, then, we can use [`include_str!()`](https://doc.rust-lang.org/stable/std/macro.include_str.html) to include those strings either in a constant or directly when passing the data to the lexer in each test.

What do you think?

> EDIT: Works on local machine but strangely not CI, very strange behaviour.

The error is giving is indeed strange. Not sure if it's related (probably isn't), but I saw that you added a line between the Jemallocator static and its configuration. Could we remove that?

neeldug:
> I had thought of using one file for all, but what you did is even better, and it brings some interesting posibilities. We can use JavaScript (`.js`) files for each (instead of `.rs`), and use normal JavaScript syntax, then, we can use [`include_str!()`](https://doc.rust-lang.org/stable/std/macro.include_str.html) to include those strings either in a constant or directly when passing the data to the lexer in each test.
> 
> What do you think?

Actually, this seems very reasonable, and in fact seems better than the previous implementation since, it means raw JavaScript benchmarks can be written without having to deal with any rust formatting. But for the `include_str!()` macro, does the macro format it as a string literal, i.e. removing the necessity for dealing with r#""#?

> The error is giving is indeed strange. Not sure if it's related (probably isn't), but I saw that you added a line between the Jemallocator static and its configuration. Could we remove that?

Have removed the line, and will see on the next commit

neeldug:
Strangely enough, same issue arises... Also numerous warnings spit out due to constants.rs being treated as a bench itself.

Razican:
We don't need that file. We can just include the code samples directly on each benchmark.

neeldug:
Ahh, okay. Should I add all the import_str's at the top or next to each call?

Razican:
> Ahh, okay. Should I add all the import_str's at the top or next to each call?

I think it's better at the call site, since we only use each once, right?

Razican (approved):
This is good to go from my side :)

Lan2u (approved):
Looks great

Issue #641, opened 2020-08-18
Test new_instance_should_point_to_prototype is not checked correctly


54k1:
**Describe the bug**
The [test](https://github.com/boa-dev/boa/blob/master/boa/src/exec/tests.rs#L753-L765) just checks if the prototype of the instance is an object using the `is_object` method.

Consider the following js snippet (similar to the example in the test):
```JavaScript
function Foo () {}
var bar = new Foo();
```

The test currently asserts `typeof bar.__proto__ === "object"`
Rather it should assert `bar.__proto__ === Foo.prototype`.

I guess the point of the test was to just check if `bar.__proto__` was an object. But it's good to go one step further to assert that the prototype is correctly assigned to be `Foo.prototype`

PR #642, opened 2020-08-18
Fix 532, 533, 641

### LINKED ISSUE #532: [builtins - Object] Object.getPrototypeOf returning incorrectly
**To Reproduce**
```js
const c = function a (){};
const d = new c();

Object.getPrototypeOf(d)
```

**Expected behavior**
Should return the prototype object whose constructor field points to `function a(){}`.

`undefined` is returned instead.

54k1:
This Pull Request fixes #532, #533, & #641.

### DIFF
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -163,6 +163,10 @@ impl Interpreter {
         let new_func = Object::function(func, function_prototype);
 
         let val = Value::from(new_func);
+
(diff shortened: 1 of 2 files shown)


Razican (approved):
Thanks for the contribution!
I'm not an expert on how this should work, so I leave that for @jasonwilliams, @HalidOdat and @Lan2u.

From my side, the code looks perfect :)

HalidOdat (approved):

PR #643, opened 2020-08-19
Change default return type from null to undefined

54k1:
This PR changes the default result of `statement_list` from `null` to `undefined`.


Razican:
Thanks for the contribution! Could we add a link to the spec and a test for this?

Razican (approved):
Thanks! I think this is good to go :) We should in the future check for those "Completion" records in the spec too, which will be needed to implement async/await, and would simplify the flow for break/continue/return.

Issue #631, opened 2020-08-14
Unexpected result when applying typeof to undefined value


elasmojs:
Hi

The following code throws error which is not the expected behavior,

console.log(typeof undeclaredVariable);

Expected: "undefined"
Got: error throw

This causes issues when parsing external libraries such which use this code pattern to determine module loader type (Node vs Browser vs Require). Code snippet link [tinycolor.js](https://github.com/bgrins/TinyColor/blob/master/tinycolor.js) line# 1183

thanks

PR #645, opened 2020-08-19
Fix undefined constant expression evaluation

### LINKED ISSUE #644: `undefined` constants panic on execution
The following code panics when being run:
```javascript
let a = undefined;
```

This is because we have not implemented the execution of undefined [here](https://github.com/boa-dev/boa/blob/master/boa/src/exec/mod.rs#L366-L410). The solution is simple enough as a good first issue:

You will need to add a match arm, that converts a `Node::Const(Const::Undefined)` to an `Ok(Value::Undefined)`.

This could also be related to #631.

jcdickinson:
This Pull Request fixes/closes #644 and #631.

It changes the following:
 - Adds support for `undefined` expr
 - Makes the `Node` match exhaustive (I have `ConditionalOp` and `Continue` as a local WIP, but they don't really belong in this PR)


Razican (approved):
Thanks! Looks good from my side :)

HalidOdat (approved):

PR #646, opened 2020-08-19
Test + fmt fix

Lan2u:
The last few PR's introduced a failing test + a rust fmt issue.

Note the 'as u128' is because usize cannot fit the value 100000000000 in all environments.

This should fix both but I'm a little unsure about the test so it would be great if you could check it @HalidOdat ?

### DIFF
--- a/boa/src/exec/tests.rs
+++ b/boa/src/exec/tests.rs
@@ -966,7 +966,7 @@ fn to_length() {
     assert_eq!(
         Value::number(100000000000.0)
             .to_length(&mut engine)
-            .unwrap() as u128,
+            .unwrap() as u64,
         100000000000
     );
     assert_eq!(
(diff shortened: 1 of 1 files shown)


boa/src/exec/tests.rs
```diff
@@ -967,7 +966,7 @@ fn to_length() {
     assert_eq!(
         Value::number(100000000000.0)
             .to_length(&mut engine)
-            .unwrap(),
+            .unwrap() as u128,
```
HalidOdat, line 969: we return a `usize` from `to_length` so we are already casting as `usize` so this does not do anything.

HalidOdat: we should return a `u64` instead of `usize` since its clamped at 0 .. `2**53 - 1`

Lan2u: Changed :)

PR #651, opened 2020-08-20
Added some formatting improvements and defaults

Razican:
This PR is a proposal to improve the automatic check of formatting of the code, and an initial reformatting, mostly done automatically, but I did review it, so it should mostly be fine.

The idea is to add the following:
 - Format the Rust code in the comments too.
 - Format the `macro_rules!()` macros.
 - Make sure strings don't go past the maximum line width.
 - Merge imports in a hierarchy.
 - Use LF (Unix) as a line ending style (already enforced by `.editorconfig`, and `.gitattributes`, but as an extra check.
 - Use `//` and `///` for comments, instead of `/* */`.
 - Do not use `#[doc]` attributes except if needed (for setting the website and image of the docs).
 - Do not repeate field names in structure initialization (already checked by clippy, but autoformatting could help in PRs)
 - Do not use the `try!()` macro, and always use `?`.
 - Wrap comments to the default line width (except for long links).

Most of these options are unstable, since they sometimes break stuff, so I ran `cargo +nightly fmt` with all these on and then commented the unstable ones out. The whole documentation can be found [here](https://rust-lang.github.io/rustfmt/).

What do you think? Is it reasonable? Many of these things won't be checked until rustfmt makes them stable, but I think it's good to at least have the project this way, and then new PRs would probably adhere to this.

Let me know :)

_Note: the PR is done against the `new_lexer` branch to see the proper diff. It will be done against `master` once the new lexer is squashed and merged._


boa/src/builtins/date/tests.rs
```diff
@@ -1006,7 +1008,8 @@ fn date_proto_set_utc_full_year() -> Result<(), Box<dyn std::error::Error>> {
 
     let actual = forward_dt_utc(
         &mut engine,
-        "dt = new Date(Date.UTC(2020, 06, 08, 09, 16, 15, 779)); dt.setUTCFullYear(2012, 8, 10); dt",
+        "dt = new Date(Date.UTC(2020, 06, 08, 09, 16, 15, 779)); dt.setUTCFullYear(2012, 8, 10); \
+         dt",
```
HalidOdat, line 1012: This to me, seems very weird and remind me of the C++ macros (that i hate). for me is less readable.

HalidOdat (commented):
This looks great, especially the comments :), check my comment comment

jasonwilliams:
I was removing old branches and i think this one was in the bunch, could you do a new PR or reopen if this is still valid? apologies if it is

Razican:
> I was removing old branches and i think this one was in the bunch, could you do a new PR or reopen if this is still valid? apologies if it is

I will try to re-create it. It needed an update anyways. The internet branches were also WIP, I'll check if I still have a local copy somewhere.

PR #653, opened 2020-08-20
Fixes unit test for `builtins::tests::expm1`. #652

### LINKED ISSUE #652: Precision issue with expm
**Describe the bug**
On running cargo test, builtins::math::tests::expm1 fails its test, the stack trace shows that this is simply down to a precision issue with either the [test](https://github.com/boa-dev/boa/blob/c5b708b2ef0df5b98f2ebd945596ac3a555707ae/boa/src/builtins/math/tests.rs#L285), or the [implementation](https://github.com/boa-dev/boa/blob/d025207fc1967afbe91e317c5ae3dd98bdc29211/boa/src/builtins/math/mod.rs#L283).

**To Reproduce**
Run `cargo test` on MacOS/(probably others). 

**Expected behavior**
builtins::math::tests::expm1 test should pass.

 - OS: MacOS
 - Version: Catalina 10.15.6
 - Target triple: x86_64-Darwin
 - Rustc version: 1.45.2

**Additional context**
You can find more information in [spec](https://tc39.es/ecma262/#sec-math.expm1), or [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/expm1)

neeldug:
Adds extra digit for test cast c, so cargo test no longer fails.

This Pull Request fixes/closes #652.

It changes the following:
 - Modifies testcase c for `expm` implementation for extra digit of precision.

### DIFF
--- a/boa/src/builtins/math/tests.rs
+++ b/boa/src/builtins/math/tests.rs
@@ -305,7 +305,7 @@ fn expm1() {
 
     assert_eq!(a, String::from("NaN"));
     assert_eq!(b, String::from("NaN"));
-    assert_eq!(c.to_number(&mut engine).unwrap(), 1.718_281_828_459_045);
+    assert_eq!(c.to_number(&mut engine).unwrap(), 1.718_281_828_459_045_3);
     assert_eq!(d.to_number(&mut engine).unwrap(), -0.632_120_558_828_557_7);
     assert_eq!(e.to_number(&mut engine).unwrap(), 0_f64);
     assert_eq!(f.to_number(&mut engine).unwrap(), 6.389_056_098_930_65);


neeldug:
Strangely enough seems to have different precisions for different OS's, but only on `expm`...

Razican:
> Strangely enough seems to have different precisions for different OS's, but only on `expm`... Seems to be an upstream issue with `std::f64`.
> 
> Can probably close this now since appears to be an issue with Rust's std library.

If this is the case, you should report it in the rust repo.

neeldug:
> If this is the case, you should report it in the rust repo.

Yep, done so at [here](https://github.com/rust-lang/rust/issues/75748)

PR #656, opened 2020-08-22
Feature `Context`

HalidOdat:
This PR is related to #445 

It changes the following:
 - Add `Context`
 - Changed function signature to take a `&mut Context`.
 - Added `Context::eval()` instead of `forward_val`, `forward`, `exec`.
 - Moved stuff that should be in `Context` like `symbol_count` to `Context`.


HalidOdat:

Lan2u (approved):

Lan2u:
Why don't we fix the deprecation warnings as part of this merge request?

HalidOdat:
> Why don't we fix the deprecation warnings as part of this merge request?
> 
> Actually maybe disregard that, 1k+ warnings :O

I way that we can remove this is to set them to be `pub(crate)` this way it we can use them in the tests (for now, so we can refactor them later) and we will be able to remove the `#[allow(deprecated)]`

Lan2u:
> > Why don't we fix the deprecation warnings as part of this merge request?
> > Actually maybe disregard that, 1k+ warnings :O
> 
> A way that we can remove this is to set them to be `pub(crate)` this way it we can use them in the tests (for now, so we can refactor them later) and we will be able to remove the `#[allow(deprecated)]`

I think keeping deprecated is best as it describes the situation well.

HalidOdat:
> I think keeping deprecated is best as it describes the situation well

Yep. I'll change it back. It would be much nicer if `#[deprecated]` did not give warnings for internal tests.

Lan2u:
> > I think keeping deprecated is best as it describes the situation well
> 
> It would be much nicer if `#[deprecated]` did not give warnings for internal code/tests, sadly it does not. I made it only compile in test mode so we don't have to allow deprecated.
> 
> What do you think? @Lan2u

Yeah, it is a nice solution and to be honest, just having a forward/forward_val/exec which internally uses the new context::exec() is not a bad longer-term solution as to me this is just the same as having a test helper function.

PR #658, opened 2020-08-23
Cache standard objects

### LINKED ISSUE #613: Cache standard objects (`Array`, `Object`, etc) and there prototypes
The problem is that when we access a standard object we use `.get_field()` to get it from the global object, the problem is that we have to do a hashmap lookup and `.get_field()` does other checks that are not necessary.

we have to do:
```rust
interpreter.global().get_field("Object")
```
or to get the prototype we have to do two HashMap lookups
```rust
interpreter.global().get_field("Object").get_field("prototype")
```

**Proposed changes**
Store the common standard objects in the realm/interpreter as struct fields maybe:
```rust
struct StandardObject {
	object: GcObject,
	prototype: GcObject,
}
struct StandardObjects {
	object_object: StandardObject,
	array_object: StandardObject,
	function_object: StandardObject,
	error_object: StandardObject,
	// ... other standard objects ...
}
```
This way if we need to access the prototype of for example `Function` object we do something like:
```rust
interpreter.standard_objects.function_object.prototype
```
This way we avoid all the unnecessary `HashMap` lookups and the checks from `.get_field()`

54k1:
This PR attempts to fix #613.

- Following structures have been added (as suggested by @HalidOdat):
```Rust
// Cache constructor object (function Object()), etc
struct StandardConstructor {
    pub constructor: GcObject,
    pub prototype: GcObject
}
// Cache standard objects such as constructors
struct StandardObjects {
    pub object: StandardConstructor,
    pub function: StandardConstructor,
    // And so on..
}
```
- Interpreter now has a field `standard_objects` of type `StandardObjects`.
    - This way, global lookups can be eliminated wherever `Interpreter` is in scope.
- Method `unwrap_object` has been added to fetch the `GcObject` in `Value`.

Every builtin sets its cache field (in `interpreter.standard_objects`) in their `init` functions.

### DIFF
--- a/boa/src/builtins/object/tests.rs
+++ b/boa/src/builtins/object/tests.rs
@@ -142,3 +142,20 @@ fn object_property_is_enumerable() {
     );
     assert_eq!(forward(&mut engine, r#"x.propertyIsEnumerable()"#), "false",)
 }
+
+#[test]
+fn custom() {
+    let realm = Realm::create();
+    let mut engine = Interpreter::new(realm);
+    assert_eq!(
+        forward(&mut engine, r#"Object.getPrototypeOf(Object.prototype)"#),
+        "null"
+    );
+    assert_eq!(
+        forward(
+            &mut engine,
+            r#"Object.getPrototypeOf(Function.prototype) === Object.prototype"#
+        ),
+        "true"
+    );
+}
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -66,6 +66,9 @@ pub struct Interpreter {
 
     /// console object state.
     console: Console,
+
+    /// Holds speical objects and their prototypes (Object, Function, etc)
+    pub standard_objects: StandardObjects,
 }
 
 impl Interpreter {
@@ -76,6 +79,7 @@ impl Interpreter {
             realm,
             symbol_count: 0,
             console: Console::default(),
+            standard_objects: StandardObjects::default(),
         };
 
         // Add new builtIns to Interpreter Realm
@@ -349,11 +353,12 @@ impl Interpreter {
         RcSymbol::from(Symbol::new(self.generate_hash(), description))
     }
 
-    /// Construct an empty object.
+    /// Construct an empty object with __proto__: Object.prototype
     #[inline]
     pub fn construct_object(&self) -> GcObject {
-        let object_prototype = self.global().get_field("Object").get_field(PROTOTYPE);
-        GcObject::new(Object::create(object_prototype))
+        GcObject::new(Object::create(
+            self.standard_objects.object.prototype.clone().into(),
+        ))
     }
 }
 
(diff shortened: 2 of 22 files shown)


boa/src/builtins/object/tests.rs
```diff
@@ -142,3 +142,20 @@ fn object_property_is_enumerable() {
     );
     assert_eq!(forward(&mut engine, r#"x.propertyIsEnumerable()"#), "false",)
 }
+
+#[test]
+fn custom() {
```
HalidOdat, line 147: Could we give this test a better name?

boa/src/exec/mod.rs
```diff
@@ -409,3 +414,43 @@ impl Executable for Node {
...
+}
+
+impl Default for StandardConstructor {
+    fn default() -> Self {
+        Self {
+            constructor: GcObject::new(Object::default()),
+            prototype: GcObject::new(Object::default()), // __proto__ === null
+        }
+    }
+}
```
HalidOdat, line 432: There is a problem with this when we initialize Standard Objects we create some dummy garbage collected objects on the need then we immediatly drop them when we assign to them. This is inefficient.

54k1: True. Is there any way to then initialize the StandardObjects structure. Since we are putting it in the `Interpreter`, we need to initialize it before initializing the builtins.
Using `Option<StandardConstructor>` is an option. But subsequent accesses would require unwrapping the `Option`, which is an unnecessary check.

HalidOdat: > True. Is there any way to then initialize the StandardObjects structure. Since we are putting it in the `Interpreter`, we need to initialize it before initializing the builtins.
> Using `Option<StandardConstructor>` is an option. But subsequent accesses would require unwrapping the `Option`, which is an unnecessary check.

Hmmm. yep that is a problem. how about we initialize them inplace so the default the `make_constructor_fn` instead of returning `GcObject` we pass the constructor.
The second option is to go with `Option` which gives unnecessary check but it is possible to remove this check with a little unsafe code. with https://doc.rust-lang.org/stable/std/hint/fn.unreachable_unchecked.html

What so you think of this? @54k1 @Razican

Razican: The question is, why do we need `StandardConstructor` to implement `Default`? Can we do it in a different way?

If there is no other option, and we know this will only be an optional thing on creation as default, what about a builder pattern? We create a `StandardConstructorBuilder`, and use an `Option`. Then we call `build()` which returns a `StandardConstructor` that has no optional field.

That would avoid using `unsafe`. But if we really want to use `unsafe`, using [`MaybeUninit`](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html) would reduce the pointer size a bit, compared to `Option`.

54k1: @Razican, we need the default since we are embedding the `StandardObjects` in the `Interpreter` which must be initialized. About the builder pattern, how would you access newly created objects?

@HalidOdat, the default allocation is redundant, but there would not be too many allocations such that this becomes a bottleneck. What do you think?

Sorry about the late response. Had some school stuff to do.

HalidOdat: > @HalidOdat, the default allocation is redundant, but there would not be too many allocations such that this becomes a bottleneck. What do you think?

It's probably not that big of a bottleneck, but I would like for us to find a nicer way of doing this in the future but for now I guess it fine :)

HalidOdat:
Closing this in favor of #722

PR #659, opened 2020-08-23
Error handling in environment

### LINKED ISSUE #622: Assignment to const results in panic

54k1:
This PR introduces error handling in `environment`.
This Pull Request fixes/closes #622.

- Environment methods (which are prone to errors) now return a Result<X, ErrorKind>, where X is either `()` or `Value`.
- `ErrorKind` enum gives a hint to the other modules about the Error. According to the spec, there only are `ReferenceError` and `TypeError` generated from environment. And so there only are two fields in the `ErrorKind` enum.
- Since ErrorKind is returned, any code using environment must map data of type `ErrorKind` to type `Error`. This is achieved using `to_error` method. For instance, `env.get_this_binding().map_err(|e| e.to_error(ctx))?;`
- `panic!`s and `assert!`s that are present in some environment methods correspond to assert statements in the spec. Some things have to be guaranteed (such as prior existence of a binding) and if they fail, they may lead to `panic!`.


HalidOdat (commented):
This looks good! we should add some test cases where a panic was being caused.

54k1:
@HalidOdat I think it would be better to add tests after there's support for strict mode added (After that, some implementation in exec(Assignment Operator) needs to be fixed).

- The Assignment operator currently creates a new binding in the current scope if no binding was found in the environment stack.
This is erroneous since in functions, any assignment without a prior declaration must result in addition of new entry to the global scope (in non-strict mode). In strict mode however, it must throw a `ReferenceError`.

Current behaviour: Assignments without prior declarations are added to the function scope.
```JavaScript
>> function f () {
x = 12;
}
>> f()
>> x // Must be visible in global
Uncaught: "ReferenceError": "x is not defined"
```

Correct behaviour:
```JavaScript
>> function f () {
    x = 12;
}
>> f()
undefined
>> x
12
// Ok
```

```JavaScript
>> function f () {
    'use strict';
    x = 12;
}
>> f() // Can't do this in strict mode
Uncaught ReferenceError: assignment to undeclared variable x
```

- There's also another issue which is that the following must throw a `SyntaxError`.
```JavaScript
let x = 12;
const x = 12;
```
What do you think?

54k1:
I'm taking a re-look at this.

Razican:
Hi! It seems this requires a rebase :) what is the state of the PR?

boa/src/builtins/function/mod.rs
```diff
@@ -124,12 +124,14 @@ impl Function {
         // Create binding
         local_env
             .borrow_mut()
-            .create_mutable_binding(param.name().to_owned(), false);
+            .create_mutable_binding(param.name().to_owned(), false)
+            .unwrap();
```
Razican, line 128: I would prefer to use `expect()` in all these instead of `unwrap()`.

boa/src/environment/declarative_environment_record.rs
```diff
@@ -41,9 +42,8 @@ impl EnvironmentRecordTrait for DeclarativeEnvironmentRecord {
         self.env_rec.contains_key(name)
     }
 
-    fn create_mutable_binding(&mut self, name: String, deletion: bool) {
+    fn create_mutable_binding(&mut self, name: String, deletion: bool) -> Result<(), ErrorKind> {
         if self.env_rec.contains_key(&name) {
-            // TODO: change this when error handling comes into play
             panic!("Identifier {} has already been declared", name);
```
Razican, line 47: Should we maintain this panic?

boa/src/environment/function_environment_record.rs
```diff
@@ -60,37 +61,43 @@ pub struct FunctionEnvironmentRecord {
...
         }
     }
+
+    pub fn get_super_base(&self) -> Value {
+        let home = &self.home_object;
+        if home.is_undefined() {
+            Value::Undefined
+        } else {
+            assert!(home.is_object());
+            home.as_object().unwrap().prototype_instance().clone()
```
Razican, line 89: I think there is no need to clone this, and should remove the clippy warning.
```suggestion
            home.as_object().unwrap().prototype_instance()
```

boa/src/environment/mod.rs
```diff
@@ -6,3 +6,21 @@ pub mod function_environment_record;
 pub mod global_environment_record;
 pub mod lexical_environment;
 pub mod object_environment_record;
+
+#[derive(Debug)]
+pub enum ErrorKind {
+    ReferenceError(String),
+    TypeError(String),
```
Razican, line 13: I think we could use a `Box<str>` here to reduce a bit the memory footprint.

boa/src/object/gcobject.rs
```diff
@@ -266,7 +270,8 @@ impl GcObject {
 
                 // local_env gets dropped here, its no longer needed
                 let binding = ctx.realm_mut().environment.get_this_binding();
-                Ok(binding)
+                binding.map_err(|e| e.to_error(ctx))
+                // Ok(binding)
```
Razican, line 274: Why is this commented out?

boa/src/syntax/ast/node/identifier/mod.rs
```diff
@@ -36,7 +36,8 @@ impl Executable for Identifier {
             .realm()
             .environment
             .get_binding_value(self.as_ref())
-            .ok_or_else(|| interpreter.construct_reference_error(self.as_ref()))
+            .map_err(|e| e.to_error(interpreter))
+        // .ok_or_else(|| interpreter.construct_reference_error(self.as_ref()))
```
Razican, line 40: Why is this commented out?

Razican (commented):
This is looking very good. The CI error is unerlated, and seems to be related to the new Test262 conformance comments.

It should be writing this comment, but it's failing:
| Total | 78,413 | 78,413 | 0 |
| Passed | 18,189 | 18,193 | **+4** |
| Ignored | 15,507 | 15,507 | 0 |
| Failed | 44,717 | 44,713 | **-4** |
| Panics | 13,519 | 13,485 | **-34** |
| Conformance | 23.20 | 23.20 | **+0.01%** |

boa/src/environment/declarative_environment_record.rs
54k1: Some `panic!`s in the environment module correspond to `assert` in the spec.

boa/src/object/gcobject.rs
54k1: Oops. This is an error.

boa/src/environment/mod.rs
54k1: The Error messages are however dynamic (depend on variable names, etc.) which is why we had used a String earlier. Could you please elaborate a bit more?

boa/src/environment/function_environment_record.rs
54k1: `prototype_instance()` method returns a `&Value` and not `Value`. So we should be cloning it right?

boa/src/syntax/ast/node/identifier/mod.rs
54k1: I should remove the comment. Thanks.

boa/src/environment/mod.rs
Razican: Yep, a `String` contains a pointer, a length and a capacity. The capacity indicates how many bytes are allocated for the string, while the length indicates how many are in use. This is useful to grow the string without needing to allocate, but once we know that the String will not change anymore, it's better to drop the excess capacity and have just a pointer with length, a `Box<str>`.

This can be constructed using [`String::into_boxed_str()`](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.into_boxed_str) you can find more information on the standard library documentation :).

boa/src/environment/declarative_environment_record.rs
Razican: In that case, I think it's better to use assert:
```rust
assert!(!self.env_rec.contains_key(&name), "identifier {} has already been declared", name);
```

boa/src/environment/function_environment_record.rs
Razican: The `as_object()` method returns a `GcObject`, and this `GcObject` returns a `Value` (already cloned), so no need to clone. Also, I would recommend to use `expect()` instead of `unwrap()`.

boa/src/environment/declarative_environment_record.rs
54k1: What would you suggest for this?

```Rust
if let Some (binding) = self.env_rec.get (name) {
    ...
} else {
   panic! ("Cannot get binding value");
}
```

Razican: I think that one is good as it is now :)

boa/src/environment/mod.rs
54k1: Finally, when we are ready to construct the actual Error object, the API currently expects a template which can convert to a String as an argument (for the message).
```Rust
pub fn construct_reference_error<M> (&mut self, msg: M) -> Value
where
    M: Into<String> {
...
}
```
Would a `Box<str>` be useful in this case?

Razican: Most (if not all) of the things that can be converted to a `String` can be converted to a `Box<str>`, so changing it to `Into<Box<str>>` could make the trick.

boa/src/builtins/object/gcobject.rs
```diff
@@ -0,0 +1,309 @@
+//! This module implements the `GcObject` structure.
```
HalidOdat, line 1: We should remove this file, it seems to be a duplicate of ` boa/src/object/gcobject.rs` and was move in a previous commit

HalidOdat (approved):
Looks good to me! :)

Razican:
I would spect this to be able to remove some of the panics listed in #817. Namely:

<details>
      <summary>
        <code>Identifier x has already been declared</code>
        <a href="https://github.com/boa-dev/boa/issues/659">#659</a>, count: 58
      </summary>
      Example:
      <pre>thread 'main' panicked at 'Identifier x has already been declared', boa/src/environment/function_environment_record.rs:94:13</pre>
    </details>

and 

<details>
      <summary>
        <code>Binding already exists!</code>
        <a href="https://github.com/boa-dev/boa/issues/659">#659</a>, count: 30
      </summary>
      Example:
      <pre>thread 'main' panicked at 'Binding already exists!', boa/src/environment/global_environment_record.rs:107:13</pre>
    </details>

and

<details>
      <summary>
        <code>TypeError: Cannot mutate an immutable binding x</code>
        <a href="https://github.com/boa-dev/boa/issues/659">#659</a>, count: 2
      </summary>
      Example:
      <pre>thread 'main' panicked at 'TypeError: Cannot mutate an immutable binding x', boa/src/environment/declarative_environment_record.rs:117:13</pre>
    </details>

and

```
thread 'main' panicked at 'TypeError: Cannot mutate an immutable binding i', boa/src/environment/declarative_environment_record.rs:117:13
```

But I think they still exist in this branch, right?

Razican:
Hi @54k1, did you have the chance to review why are those environment panics still happening?

Razican (approved):
This looks good to me. Only needs a `cargo fmt`.

PR #660, opened 2020-08-25
Make `GcObject::contruct` not take `'this'`

HalidOdat:
It changes the following:
 - It does not make sense for `GcObject::contruct()` to take `'this'` since it constructs it.


Razican (approved):
Looks good!

Lan2u (approved):

PR #661, opened 2020-08-25
Add documentation to `GcObject` methods

HalidOdat:
It changes the following:
 - Added documentation to `GcObject` methods
 - Added type aliases `Ref` and `RefMut` to abstract the gc.


Razican (approved):
Looks good! thanks!

PR #665, opened 2020-08-26
Added approx_eq! macro for expm1 tests.

### LINKED ISSUE #664: `builtins::math::tests::expm1` MacOS Fix
**Describe the bug**
This is a bug which is inherent throughout all languages due to the non-trivial floating point arithmetic varying implementations across platforms, this will result in certain platforms failing tests for various float-based operations that are non-trivial. 

**To Reproduce**
Run `cargo test` on MacOS, and probably a few other OS's.

**Expected behavior**
All tests should pass, as these failures are simply due to floating point inaccuracies.

 - OS: MacOS
 - Version: 10.15.6
 - Target triple: Darwin
 - Rustc version: rustc 1.45.2 (d3fb005a3 2020-07-31)

**Additional context**
This should be solvable with use of a tolerance or by using some approximately equal assertion, just quickly looking over these, [`approx`](https://docs.rs/approx/0.3.2/approx/) or [`float-cmp`](https://docs.rs/float-cmp/0.8.0/float_cmp/), the latter seems much more often used, this may also be used in other cases where tests fail in future cases for floating point arithmetic.

neeldug:
Added approx_eq! macro for expm1 tests due to floating point arithmetic
inaccuracies, using default ULP & epsilon values. approx_eq! macro does
not work for NaN values, however, for tests this should be okay anyway!
Solves #664.

This Pull Request fixes/closes #664.

It changes the following:
 - `builtins::math::tests::expm1` strict equals assertion to approx_eq assertion

### DIFF
--- a/boa/Cargo.toml
+++ b/boa/Cargo.toml
@@ -26,6 +26,7 @@ bitflags = "1.2.1"
 indexmap = "1.4.0"
 ryu-js = "0.2.0"
 chrono = "0.4"
+float-cmp = { version = "0.8.0", features = ["std"] }
 
 # Optional Dependencies
 serde = { version = "1.0.114", features = ["derive"], optional = true }
--- a/boa/Cargo.toml
+++ b/boa/Cargo.toml
@@ -26,7 +26,7 @@ bitflags = "1.2.1"
 indexmap = "1.4.0"
 ryu-js = "0.2.0"
 chrono = "0.4"
-float-cmp = { version = "0.8.0", features = ["std"] }
+float-cmp = "0.8.0"
 
 # Optional Dependencies
 serde = { version = "1.0.114", features = ["derive"], optional = true }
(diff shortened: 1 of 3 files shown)


boa/Cargo.toml
```diff
@@ -26,6 +26,7 @@ bitflags = "1.2.1"
 indexmap = "1.4.0"
 ryu-js = "0.2.0"
 chrono = "0.4"
+float-cmp = "0.8.0"
```
Razican, line 29: This should not be a dependency, it should be a `dev-dependency`, in order to not add unused dependencies to our main build.

Razican (changes requested):
Thanks! it looks perfect. You only need to move the new dependency to the `[dev-dependencies]` section.

Razican (approved):
This is good to go from my side!

HalidOdat (approved):

PR #666, opened 2020-08-28
Some cleanup

Razican:
This Pull Request cleans up a bit the repositoy, especially regarding the WASM demo.

It changes the following:
 - Adds [Prettier][prettier] formatting
 - Adds checks in the CI for both, the JS/MD/YML styles and the fact that the WASM target builds.
 - Upgrades dependencies for both the Rust crates and the boa WASM demo.
 - Removes the `Makefile` that we were not using.
 - Adds `#[track_caller]` in the parser test functions, so that we know what test snippet is failing, instead of all panics pointing to the checker functions. This is a new feature of Rust 1.46, so this raises the minimum supported Rust version.

[prettier]: https://prettier.io/


Razican:
It seems that the wasm-pack build is failing, but can't figure out why.

HalidOdat (commented):
This looks good! :) maybe we should leave the wasm build out for now until we discover why its not building.

What do you think?

Razican:
> This looks good! :) maybe we should leave the wasm build out for now until we discover why its not building.
> 
> What do you think?

I would like to do some extra debugging during this week, before merging.

Razican:
Fixed by forcing the `wasm-pack` path :)

boa/benches/bench_scripts/array_pop.js
```diff
@@ -1,24 +1,221 @@
...
-                     27, 29, 2828, 234, 23, 56, 32, 56, 67, 77, 32, 45, 93,
-                     17, 28, 83, 62, 99, 36, 28, 93, 27, 29, 2828, 234, 23,
-                     56, 32, 56, 67, 77, 32, 45, 93, 17, 28, 83, 62, 99, 36,
-                     28, 93, 27, 29, 2828, 234, 23, 56, 32, 56, 67, 77, 32,
-                     45, 93, 17, 28, 83, 62, 99, 36, 28, 93, 27, 29, 2828, 234,
-                     23, 56, 32, 56, 67, 77, 32, 45, 93, 17, 28, 83, 62, 99,
-                     36, 28, 93, 27, 29, 2828, 234, 23, 56, 32, 56, 67, 77, 32,
-                     45, 93, 17, 28, 83, 62, 99, 36, 28];
+(function () {
+  let testArray = [
```
Razican, line 2: I am not 100% sure of this array formatting, but I guess it's the standard JS formatting.

HalidOdat (approved):

PR #667, opened 2020-08-30
Fix panic when calling function that mutates itself

### LINKED ISSUE #663: Calling a function that mutates itself causes borrow panic
This is caused when we call a function we borrow the function object and we do not release the hold until function ends. see `GcObject::call` / `GcObject::construct`

**To Reproduce**

```javascript
function x() {
	x.y = 3;
}
x();
```

**Expected behavior**
Not panic and x to have a property `y` with value `3`

**Actual behavior**
Panics:
```
thread 'main' panicked at 'Object already borrowed: BorrowMutError', boa/src/builtins/object/gcobject.rs:41:9
```

dvtkrlbs:
Fixes #663 

This commit fixes the panic when a function mutates itself. The issue happens since we borrow the object and when we call the statementList of the object we borrow_mut the object again to mutate it.
It changes the following:
 - This adds a RcStatementList to cheaply clone StatementList
 - It calls the run on the body of the function after dropping the first borrow by returning only the body after setting up the function local env

Thank you @HalidOdat for their help 😃

### DIFF
--- a/boa/src/builtins/function/tests.rs
+++ b/boa/src/builtins/function/tests.rs
@@ -1,12 +1,10 @@
-use crate::exec::Executor;
-use crate::realm::Realm;
-use crate::{builtins::value::from_value, forward, forward_val};
+use crate::{exec::Interpreter, forward, forward_val, realm::Realm};
 
 #[allow(clippy::float_cmp)]
 #[test]
 fn check_arguments_object() {
     let realm = Realm::create();
-    let mut engine = Executor::new(realm);
+    let mut engine = Interpreter::new(realm);
     let init = r#"
         function jason(a, b) {
             return arguments[0];
@@ -15,11 +13,33 @@ fn check_arguments_object() {
         "#;
 
     eprintln!("{}", forward(&mut engine, init));
-    let expected_return_val = 100;
+
     let return_val = forward_val(&mut engine, "val").expect("value expected");
     assert_eq!(return_val.is_integer(), true);
     assert_eq!(
-        from_value::<i32>(return_val).expect("Could not convert value to i32"),
-        expected_return_val
+        return_val
+            .to_i32(&mut engine)
+            .expect("Could not convert value to i32"),
+        100
+    );
+}
+
+#[test]
+fn check_self_mutating_func() {
+    let realm = Realm::create();
+    let mut engine = Interpreter::new(realm);
+    let func = r#"
+        function x() {
+	        x.y = 3;
+        }
+        x();
+        "#;
+    eprintln!("{}", forward(&mut engine, func));
+    let y = forward_val(&mut engine, "x.y").expect("value expected");
+    assert_eq!(y.is_integer(), true);
+    assert_eq!(
+        y.to_i32(&mut engine)
+            .expect("Could not convert value to i32"),
+        3
     );
 }
--- a/boa/src/builtins/object/gcobject.rs
+++ b/boa/src/builtins/object/gcobject.rs
@@ -5,12 +5,13 @@
 use super::{Object, PROTOTYPE};
 use crate::{
     builtins::{
-        function::{create_unmapped_arguments_object, BuiltInFunction, Function},
+        function::{create_unmapped_arguments_object, BuiltInFunction, Function, NativeFunction},
         Value,
     },
     environment::{
         function_environment_record::BindingStatus, lexical_environment::new_function_environment,
     },
+    syntax::ast::node::statement_list::RcStatementList,
     Executable, Interpreter, Result,
 };
 use gc::{Finalize, Gc, GcCell, GcCellRef, GcCellRefMut, Trace};
@@ -31,6 +32,13 @@ pub type RefMut<'object> = GcCellRefMut<'object, Object>;
 /// Garbage collected `Object`.
 #[derive(Trace, Finalize, Clone)]
 pub struct GcObject(Gc<GcCell<Object>>);
+
+// This is needed for the call method since we cannot mutate the function itself since we
+// already borrow it so we get the function body clone it then drop the borrow and run the body
+enum FunctionBody {
+    BuiltIn(NativeFunction),
+    Ordinary(RcStatementList),
+}
 
 impl GcObject {
     /// Create a new `GcObject` from a `Object`.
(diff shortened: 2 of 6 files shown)


boa/src/builtins/object/gcobject.rs
```diff
@@ -99,11 +106,12 @@ impl GcObject {
     // <https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist>
     pub fn call(&self, this: &Value, args: &[Value], ctx: &mut Interpreter) -> Result<Value> {
         let this_function_object = self.clone();
-        let object = self.borrow();
-        if let Some(function) = object.as_function() {
+        let f_body: FunctionBody = if let Some(function) = self.borrow().as_function() {
```
HalidOdat, line 109:
```suggestion
        let body = if let Some(function) = self.borrow().as_function() {
```

If type annotation is not needed its better to leave it to the compiler.

boa/src/builtins/object/gcobject.rs
```diff
@@ -152,18 +160,30 @@ impl GcObject {
                         ctx.realm.environment.push(local_env);
 
                         // Call body should be set before reaching here
-                        let result = body.run(ctx);
+                        // let result = body.run(ctx);
+                        //
+                        // // local_env gets dropped here, its no longer needed
+                        // ctx.realm.environment.pop();
+                        // result
```
HalidOdat, line 167:
```suggestion
```

This comment does not serve any purpose.

boa/src/builtins/object/gcobject.rs
```diff
@@ -152,18 +160,30 @@ impl GcObject {
                         ctx.realm.environment.push(local_env);
```
HalidOdat, line 160: I would move this to the bottom when we call the function, since its easier to follow the code when its in one place.

boa/src/syntax/ast/node/statement_list.rs
```diff
@@ -62,3 +64,24 @@ impl fmt::Display for StatementList {
...
+}
+
+impl From<StatementList> for RcStatementList {
+    #[inline]
+    fn from(statementlist: StatementList) -> Self {
+        Self(Rc::from(statementlist))
+    }
+}
+
+unsafe impl Trace for RcStatementList {
```
HalidOdat, line 85: It would be nice to add a comment on why it is `unsafe` why is this code ok with `SAFETY: `

boa/src/syntax/ast/node/statement_list.rs
```diff
@@ -62,3 +64,24 @@ impl fmt::Display for StatementList {
         self.display(f, 0)
     }
 }
+
+#[derive(Clone, Debug, Finalize, PartialEq)]
+pub struct RcStatementList(Rc<StatementList>);
```
HalidOdat, line 69: This does not need to be `pub` it can be `pub(crate)`. Also some documentation would be nice.

boa/src/builtins/object/gcobject.rs
```diff
@@ -32,6 +34,11 @@ pub type RefMut<'object> = GcCellRefMut<'object, Object>;
 #[derive(Trace, Finalize, Clone)]
 pub struct GcObject(Gc<GcCell<Object>>);
 
+enum FunctionBody {
```
HalidOdat, line 37: Some documentation would be nice on this, even if its not public.

boa/src/builtins/object/gcobject.rs
```diff
@@ -3,6 +3,7 @@
 //! The `GcObject` is a garbage collected Object.
 
 use super::{Object, PROTOTYPE};
+use crate::builtins::function::NativeFunction;
 use crate::{
     builtins::{
         function::{create_unmapped_arguments_object, BuiltInFunction, Function},
```
HalidOdat, line 9:
```suggestion
use crate::{
    builtins::{
        function::{create_unmapped_arguments_object, BuiltInFunction, Function, NativeFunction},
```

HalidOdat (commented):
Also some tests showing that it does not panic any more and give the expected behaviour

boa/src/syntax/ast/node/statement_list.rs
dvtkrlbs: hmm we cannot export the crate nor we can use it on any of the builtins. i cannot seem to mark it pub(crate) and just export it as pub(crate)

HalidOdat: > hmm we cannot export the crate nor we can use it on any of the builtins. i cannot seem to mark it pub(crate) and just export it as pub(crate)

My bad, the `Function` enum is pub so its fields have to be pub too. In this case leave it as pub.

dvtkrlbs: Oh ok i leaked that type through the Function enum now i fixed it

dvtkrlbs:
I added the test case from the issue

dvtkrlbs:
Oh wait it should not return the edited value from the call statement

dvtkrlbs:
Also this code snippet
```js
function y() {}
function x() { y.y = 3; } // Boa panics here with the error binding already exists (this happens in master lemme try with my branch)
// y.y is undefined here
x()
// y.y is 3 here
```

boa/src/builtins/object/gcobject.rs
```diff
@@ -151,19 +160,24 @@ impl GcObject {
 
                         ctx.realm.environment.push(local_env);
 
-                        // Call body should be set before reaching here
-                        let result = body.run(ctx);
-
-                        // local_env gets dropped here, its no longer needed
-                        ctx.realm.environment.pop();
-                        result
+                        FunctionBody::Ordinary(RcStatementList::from(body.clone()))
```
HalidOdat, line 163: It's better to expose the `RcStatementList` rather than copy the whole function body ast every time we call.

boa/src/builtins/function/tests.rs
```diff
@@ -15,11 +13,36 @@ fn check_arguments_object() {
...
+
+    let return_val = forward_val(&mut engine, "a").expect("value expected");
+    assert_eq!(return_val.is_integer(), true);
+    assert_eq!(
+        return_val
+            .to_i32(&mut engine)
+            .expect("Could not convert value to i32"),
+        3
     );
 }
```
HalidOdat, line 48:
```suggestion
#[test]
fn self_mutating_function() {
    let realm = Realm::create();
    let mut engine = Interpreter::new(realm);
    let func = r#"
        function x() {
	        x.y = 3;
        }
        x();
        "#;

    eprintln!("{}", forward(&mut engine, func));

    let y = forward_val(&mut engine, "x.y").expect("value expected");
    assert_eq!(y.is_integer(), true);
    assert_eq!(
        y.to_i32(&mut engine)
            .expect("Could not convert value to i32"),
        3
    );
}
```
This seems like we are testing the return value of the function and not the function object itself

HalidOdat:
> Oh wait it should not return the edited value from the call statement. I have no idea why that happens. When we call a self mutating function somehow that mutated value leaks and gets returned from the function. It should not happen this way.

That is strange. does it happen on master too. if so then report an issue. this bug does not seem related to this PR

dvtkrlbs:
```js
>> function x() {}
undefined
>> function y() { x.y = 3 }
undefined
>> x.y
undefined
>> y()
3
```
happens on master too. so i will apply your suggestions and then this should be ready to merge since this issue is irrelevant to my pr

dvtkrlbs:
Opened issue #672

HalidOdat (approved):

Razican (approved):
Looks good to me :) Thanks for the fix!

PR #668, opened 2020-08-30
Fix clippy on Nightly

dvtkrlbs:
This PR just fixes clippy

It changes the following:
 - It changes match blocks that returns booleans with two arms and if let blocks to the matches! macro 

Ok i accidently tried this with nightly clippy and fixed a lint error for it. So this pr could be for nothing. But i can make it a draft PR and we can revisit it after this lint comes to the stable (if it ever does).


Razican:
> Also why clippy is not enabled on CI?

Isn't it? I thought we had an specific test, that's odd.

dvtkrlbs:
Oh there is. How come it was not failing for this lint ?

Lan2u (approved):

Lan2u (approved):
Approved but only once this is stable.

Razican (approved):
I'm happy to merge this as-is. The `matches!()` macro is already stable, it's the lint the only thing that is unstable, so merging this would future-proof the project. What do everyone else think?

Lan2u:
> I'm happy to merge this as-is. The `matches!()` macro is already stable, it's the lint the only thing that is unstable, so merging this would future-proof the project. What do everyone else think?

If @dvtkrlbs can resolve the merge conflict then I'm happy to merge this

HalidOdat (approved):

Issue #576, opened 2020-07-18
Implement for...of syntax for Iterables


joshwd36:
As more collections are added, it would be useful to implement the `for...of` syntax for Iterables, and adding iterables to the String, Array, and Map types. [ECMAScript specification][spec].

[spec]: https://www.ecma-international.org/ecma-262/10.0/index.html#sec-for-in-and-for-of-statements

This code should now work and give the expected result:
```javascript
let a = [1, 2, 3];
for (i of a) {
    console.log(i);
}
```
The expected output is:
```javascript
1
2
3
```

I'm happy to have a go at this myself, although as this will be my first time adding syntax features, I'm wondering whether I should wait for #559 to land or just go right ahead with the current lexer?

Lan2u:
You could make a branch-based from #559 if you wanted to get started right away, I expect that most of the work regarding this would be in the parser (syntax/parser/statement/iteration) and exec (exec/iteration/mod.rs)

dvtkrlbs:
I can take a look at this

joshwd36:
> I can take a look at this. Does this include the for-in statement too?

I've made a start on it already, but haven't been able to work on it recently. I should have more time this week though so I'll try and finish it then, if not I'm happy to transfer what I've done over to you.

jcdickinson:
Dropping #671 for whoever wants to take this on. See my comments there, as there is a tricky design question. My current thoughts are:

```rust
pub enum IterableForLoopKind {
    In,
    Of
}
struct InnerIterableForLoop {
    kind: IterableForLoopKind,
    variable: SingleVarLetOrConstDecl, // or something
    object: Node,
    body: Node
}
```

Note that #670 would also be closed, as `for..in` is simply a `for..of` with a built-in iterator.

PR #671, opened 2020-08-31
for..in WIP

jcdickinson:
This Pull Request fixes/closes 670.

It changes the following:
 - Adds `for..in` support

### DIFF
--- a/boa/src/syntax/ast/node/iteration.rs
+++ b/boa/src/syntax/ast/node/iteration.rs
@@ -138,6 +138,103 @@ impl InnerForLoop {
     }
 }
 
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct ForInLoop {
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    inner: Box<InnerForInLoop>,
+}
+
+impl ForInLoop {
+    /// Creates a new for loop AST node.
+    pub(in crate::syntax) fn new<V, O, B>(variable: V, object: O, body: B) -> Self
+    where
+        V: Into<Node>,
+        O: Into<Node>,
+        B: Into<Node>,
+    {
+        Self {
+            inner: Box::new(InnerForInLoop::new(variable, object, body)),
+        }
+    }
+
+    pub fn variable(&self) -> &Node {
+        self.inner.variable()
+    }
+
+    pub fn object(&self) -> &Node {
+        self.inner.object()
+    }
+
+    /// Gets the body of the for loop.
+    pub fn body(&self) -> &Node {
+        self.inner.body()
+    }
+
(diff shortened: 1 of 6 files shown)


boa/src/syntax/ast/node/iteration.rs
```diff
@@ -138,6 +138,103 @@ impl InnerForLoop {
...
+    fn from(for_in_loop: ForInLoop) -> Self {
+        Self::ForInLoop(for_in_loop)
+    }
+}
+
+/// Inner structure to avoid multiple indirections in the heap.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+struct InnerForInLoop {
+    variable: Node,
```
jcdickinson, line 203: This really smells.

I think we might need an AST node that can represent a single `const`, `let` or `var`. Thoughts?

boa/src/syntax/ast/node/iteration.rs
```diff
@@ -138,6 +138,103 @@ impl InnerForLoop {
...
+impl From<ForInLoop> for Node {
+    fn from(for_in_loop: ForInLoop) -> Self {
+        Self::ForInLoop(for_in_loop)
+    }
+}
+
+/// Inner structure to avoid multiple indirections in the heap.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+struct InnerForInLoop {
```
jcdickinson, line 202: I'm wondering if this should be `IterableForLoop` containing an enum (`In, Of`), because both are basically the same syntax and `In` is supposed to use an iterator internally (according to the spec).

boa/src/syntax/parser/statement/iteration/for_statement.rs
```diff
@@ -88,42 +88,64 @@ where
...
+                cursor.expect(
+                    TokenKind::Punctuator(Punctuator::CloseParen),
+                    "for..in statement",
+                )?;
+                let body =
+                    Statement::new(self.allow_yield, self.allow_await, self.allow_return).parse(cursor)?;
+                Ok(ForInLoop::new(variable, object, body).into())
             }
-            Some(tok) if tok.kind() == &TokenKind::identifier("of") => {
+            Some(tok) if tok.kind() == &TokenKind::identifier("of") && can_iterate => {
```
jcdickinson, line 112: Is `TokenKind::identifier("of")` a placeholder, or intentional? i.e. Should there be `TokenKind::Keyword(Keyword::Of)`?

boa/src/syntax/ast/node/iteration.rs
jcdickinson: This will probably be really difficult for execution, too.

boa/src/syntax/ast/node/iteration.rs
jcdickinson: This would block this implementation on #576.

jcdickinson:
Dropping in favor of the existing work done for #576.

PR #673, opened 2020-09-01
Add `#[track_caller]` to `GcObject` methods that can panic

HalidOdat:
It changes the following:
 - Added `#[track_caller]` to `GcObject` methods that panic. This should make finding borrow error much easier.


Razican (approved):
Looks good, thanks!

PR #693, opened 2020-09-11
Feat: Rename Object::prototype() and Object::set_prototype()

### LINKED ISSUE #607: Rename `Object::prototype()` and `Object::set_prototype()` to `Object::prototype_instance()` and `Object::set_prototype_instance()`
The `Object::prototype()` returns the prototype instance `__proto__`, so it makes more sense to call it `Object::prototype_instance()`. they are is located [here](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/object/mod.rs#L432-L439)

RageKnify:
This Pull Request fixes/closes #607.

It changes the following:

- Rename `Object::prototype()` to `Object::prototype_instance()`
- Rename `Object::set_prototype()` to `Object::set_prototype_instance()`

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -42,14 +42,17 @@ impl Array {
                 .expect("Could not get global object"),
         ));
         array.set_data(ObjectData::Array);
-        array.as_object_mut().expect("array object").set_prototype(
-            interpreter
-                .realm()
-                .environment
-                .get_binding_value("Array")
-                .expect("Array was not initialized")
-                .get_field(PROTOTYPE),
-        );
+        array
+            .as_object_mut()
+            .expect("array object")
+            .set_prototype_instance(
+                interpreter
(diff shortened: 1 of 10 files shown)


HalidOdat (approved):

RageKnify:
I don't know why the benchmark check failed.
But I've found the reason for the bad coverage results, as mentioned in xd009642/tarpaulin#351, `tarpaulin` mistakenly considers chained method calls as not hit if split among multiple lines, resulting in situations like the following where lines 337-339 and 341-343 are considered as not being covered by the tests:
wrong_coverage
In this case `cargo fmt` increased the number of lines because of the new longer method name and it increased the number of lines considered to not be covered.

HalidOdat:
> I don't know why the benchmark check failed.

The benchmarks are failing because of an unrelated issue (criterion 0.3.3 is not compatible with critcmp https://github.com/BurntSushi/critcmp/pull/6).

PR #695, opened 2020-09-13
Fix CI Commit failure

neeldug:
Unsure as to whether this should be committed or not, but should fix
failing CI

It changes the following:

- `test262.yml`'s commit section to add test262 as a tracked file


HalidOdat (approved):

Razican (approved):
Ooops, yes, I forgot this!

PR #704, opened 2020-09-17
Implementation of for...of loops

### LINKED ISSUE #576: Implement for...of syntax for Iterables

joshwd36:
This Pull Request fixes/closes #576.

It changes the following:

- Adds the array iterator object and `values()` function to arrays.
- Sets the `Symbol.iterator` parameter to array pointing to the `values()` function. Also adds the `get_well_known_symbol()` function to `Context`.
- Adds support for a limited subsection of for...of loops.

Currently for...of only works if the iterating variable is declared before the loop and is referenced without another declaration. e.g.
``` js
var i;
for (i of [1, 2, 3]) {
    console.log(i)
}
```
I'm unsure of the approach to take for implementing the rest of the loop. The main problem is that the initial part of a for statement (`for (...`) is treated differently depending on whether it is an ordinary for loop or a for...of/in loop. My current idea is to loop through each token until either a semicolon or in/of is found, collecting the tokens into a vec and using that as a cursor to parse the initial element depending on what was found. This does seem a bit hacky though, so I'd be open to other ideas.

A slightly more minor note is that I'm unsure of whether my implementation for `get_well_known_symbol` is correct as there seems to be a lot of indirection required to get the symbol object.


boa/src/context.rs
```diff
@@ -495,4 +495,17 @@ impl Context {
...
+    where
+        K: Into<PropertyKey>,
+    {
+        let global_object = self.global_object();
+        let symbol = global_object.get_field("Symbol");
+        let symbol_iterator: Value = (&symbol.get_property(name)?).into();
+        let symbol_iterator = symbol_iterator.as_object()?;
+        let symbol_iterator = symbol_iterator.get_string_property("value")?;
+        let symbol_iterator = symbol_iterator.value.as_ref()?.as_symbol()?;
+        Some(symbol_iterator)
```
HalidOdat, line 509:
```suggestion
        let global_object = self.global_object();
        let symbol = global_object.get_field("Symbol");
        let symbol_iterator: Value = (&symbol.get_property(name)?).into();
        let symbol_iterator = symbol_iterator.as_object()?;
        let symbol_iterator = symbol_iterator.get_string_property("value")?;
        let symbol_iterator = symbol_iterator.value.as_ref()?.as_symbol()?;
        Some(symbol_iterator)
```

Caching the well known symbols should fix this problem, I'll do a PR for this.

HalidOdat (commented):
This looks good! I just skimmed through the code I'll do a proper review later :)

boa/src/builtins/array/array_iterator.rs
```diff
@@ -0,0 +1,141 @@
...
+pub enum ArrayIterationKind {
+    Key,
+    Value,
+    KeyAndValue,
+}
+
+#[derive(Debug, Clone, Finalize, Trace)]
+pub struct ArrayIterator {
+    array: Value,
+    next_index: i32,
```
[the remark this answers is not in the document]
joshwd36: This is actually something that seems a bit inconsistent in array. I think currently array length is stored as an integer field in the object, which is an `i32`, and most methods in array use `as_number` and cast to `i32`. However, some methods use the `to_length` function which returns a usize, although I feel like that must result in some loss of precision .

boa/src/context.rs
HalidOdat: Created the PR to help us with this #706

joshwd36:
This now works on most types of for...of loop. The one thing I'm unsure of is changing the behaviour of const declarations, as it feels a bit less neat, but I couldn't think of a different way to do it while still having the lhs of the for...of loop be parsed in the same way as an ordinary for loop.

boa/src/builtins/array/array_iterator.rs
```diff
@@ -0,0 +1,149 @@
...
+        make_builtin_fn(Self::next, "next", &array_iterator, 0, ctx);
+        let mut function = Object::function(
+            Function::BuiltIn(
+                BuiltInFunction(|v, _, _| Ok(v.clone())),
+                FunctionFlags::CALLABLE,
+            ),
+            ctx.global_object()
+                .get_field("Function")
+                .get_field("prototype"),
+        );
```
HalidOdat, line 68: shouldn't this point to array iterator `prototype` and not `Object.prototype`
Also we set it to `Object.prototype` here:
```rust
let array_iterator = Value::new_object(Some(
            &ctx.realm()
                .environment
                .get_global_object()
                .expect("Could not get global object"),
        ));
```
and again here:
```rust
array_iterator
            .as_object_mut()
            .expect("array iterator object")
            .set_prototype_instance(
                ctx.realm()
                    .environment
                    .get_binding_value("Object")
                    .expect("Object was not initialized")
                    .borrow()
                    .get_field(PROTOTYPE),
            );
```


This should be in `<Array Iterator>.prototype.next`. This should only be done in the builtin initialization phase and access that prototype  when we need it. we can store it in `Context` for now and initialize it and use it where we need it.
```rust
make_builtin_fn(Self::next, "next", &array_iterator, 0, ctx);
```

joshwd36: Could you clarify the implementation of this? Do you mean having an init function that creates a prototype instance that is then set as the prototype instance when an Array Iterator object is created?

joshwd36: I tihnk I've figured it out. Would it be as a cache of iterator prototypes held in context, similar to the cache of well known symbols?

HalidOdat: > I think I've figured it out. Would it be as a cache of iterator prototypes held in context, similar to the cache of well known symbols?

Sorry for not explaning it that much, but its pretty much the same we would initialize it in the builtins initialization phase. and use it when we construct a Array iterator

joshwd36: I've implemented something, would you be able to tell me if it's the right sort of thing?

HalidOdat: > I've implemented something, would you be able to tell me if it's the right sort of thing?

Yes. This looks very good! Definitely in the right direction :)

boa/src/builtins/array/array_iterator.rs
```diff
@@ -0,0 +1,149 @@
...
+                        Ok(create_iter_result_object(ctx, element_value, false))
+                    }
+                    ArrayIterationKind::KeyAndValue => {
+                        let element_value = array_iterator.array.get_field(index);
+                        let result = Array::make_array(
+                            &Value::new_object(Some(
+                                &ctx.realm()
+                                    .environment
+                                    .get_global_object()
+                                    .expect("Could not get global object"),
```
HalidOdat, line 109:
```suggestion
                                ctx.global_object()
```
we can call the `Context::global_object()` to get the global object instead of searching through the environment

boa/src/builtins/iterable/mod.rs
```diff
@@ -0,0 +1,141 @@
+use crate::builtins::function::{BuiltInFunction, Function, FunctionFlags};
+use crate::builtins::ArrayIterator;
+use crate::object::{Object, PROTOTYPE};
+use crate::BoaProfiler;
+use crate::{property::Property, Context, Value};
+
+#[derive(Debug, Default)]
+pub struct IteratorPrototypes {
+    iterator_prototype: Value,
+    array_iterator: Value,
```
HalidOdat, line 10:
```suggestion
    iterator_prototype: GcObject,
    array_iterator: GcObject,
```
It would be better if these were `GcObject` since:
 - The prototype can only be an object (fits the GcObject) while `Value` can be other variants
 - `Value` is 24 bytes a but `GcObject` is 8

boa/src/object/mod.rs
```diff
@@ -445,4 +472,8 @@ impl Object {
             _ => None,
         }
     }
+
+    pub fn get_string_property(&self, key: &str) -> Option<&Property> {
+        self.string_properties.get(key)
+    }
```
HalidOdat, line 478: I don't thick we need this any more

boa/src/builtins/iterable/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/// CreateIterResultObject( value, done )
+///
+/// Generates an object supporting the IteratorResult interface.
+pub fn create_iter_result_object(ctx: &mut Context, value: Value, done: bool) -> Value {
+    let object = Value::new_object(Some(
+        &ctx.realm()
+            .environment
+            .get_global_object()
+            .expect("Could not get global object"),
```
HalidOdat, line 57:
```suggestion
        ctx.global_object()
```

boa/src/builtins/string/string_iterator.rs
```diff
@@ -0,0 +1,94 @@
...
+            string,
+            next_index: 0,
+        }
+    }
+
+    pub fn create_string_iterator(ctx: &mut Context, string: Value) -> Result<Value> {
+        let string_iterator = Value::new_object(Some(
+            &ctx.realm()
+                .environment
+                .get_global_object()
```
HalidOdat, line 28:
```suggestion
            ctx.global_object()
```

HalidOdat (commented):
This is almost ready to merge :)

joshwd36:
> 
> 
> This is almost ready to merge :)

Great! In terms of implementing iterators for objects, there is still map to do. Should I implement this now (shouldn't take too long) or just leave it. Also, what should I do about the array index in the array iterator?

HalidOdat:
> Should I implement this now (shouldn't take too long) or just leave it.

Sure. if you want we can implement it in this :)

> Also, what should I do about the array index in the array iterator?

I think changing it to u32 is the best option, the [array index](https://tc39.es/ecma262/#array-index) can only be `+0 ≤ i < 2^32 - 1` so it can fit. but i32 will overflow into negative

joshwd36:
> I think changing it to u32 is the best option, the [array index](https://tc39.es/ecma262/#array-index) can only be `+0 ≤ i < 2^32 - 1` so it can fit. but i32 will overflow into negative

Currently the only issue with that is that it sometimes has to be placed back into a value. Should that be done as the Number type rather than Integer then?

HalidOdat:
> Currently the only issue with that is that it sometimes has to be placed back into a value. Should that be done as the Number type rather than Integer then?

If value can fit in an Value::Integer(i32) it will be placed but if not it will be converted to `f64` (which it can fit) `Value::Rational(f64)`, so we don't really have to worry how its stored. <https://github.com/boa-dev/boa/blob/master/boa/src/value/conversions.rs#L76-L85>

HalidOdat (approved):

boa/src/builtins/iterable/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+    let object = Value::new_object(Some(ctx.global_object()));
+    let value_property = Property::default().value(value);
+    let done_property = Property::default().value(Value::boolean(done));
+    object.set_property("value", value_property);
+    object.set_property("done", done_property);
+    object
+}
+
+/// Get an iterator record
+pub fn get_iterator(ctx: &mut Context, iterable: Value) -> Result<IteratorRecord, Value> {
```
RageKnify, line 62: Any reason to not use `boa::Result`? Should be imported at the start, no?

boa/src/builtins/iterable/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    /// Get the next value in the iterator
+    ///
+    /// More information:
+    ///  - [ECMA reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-iteratornext
+    pub(crate) fn next(&self, ctx: &mut Context) -> Result<IteratorResult, Value> {
```
RageKnify, line 121: Same as above.

boa/src/builtins/iterable/mod.rs
joshwd36: That's a good catch, thank you

RageKnify (approved):

PR #706, opened 2020-09-18
Cache well known symbols

HalidOdat:
https://github.com/boa-dev/boa/pull/704#discussion_r490840748

It changes the following:
 - Add `Context::well_known_symbols()` method.
 - Add `WellKnownSymbols` struct
 - Add getters for the symbols to `WellKnownSymbols` struct

With this PR we can now do:
```rust

let iterator = context.well_known_symbols().iterator_symbol();
```
instead of:
```rust

let iterator = context.global_object().get_field("Symbol").get_field("iterator").as_symbol().unwrap();
```
Also there is a problem with this ^, what if we override `Symbol` object like:
```js
globalThis.Symbol = 10;
```
Now the above code will panic, we should have a way to access the well known symbols without being tied to the global bindings.

What do you think? @joshwd36

### DIFF
--- a/boa/src/builtins/symbol/mod.rs
+++ b/boa/src/builtins/symbol/mod.rs
@@ -26,6 +26,7 @@ use crate::{
 };
 use gc::{Finalize, Trace};
 
+/// A structure that contains the JavaScript well known symbols.
 #[derive(Debug, Clone)]
 pub struct WellKnownSymbols {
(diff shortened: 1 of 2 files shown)


joshwd36:
> What do you think? @joshwd36

I think this is a good way of doing it, definitely makes getting symbols from the Rust side a lot easier.

PR #711, opened 2020-09-23
Add support for >>>=

### LINKED ISSUE #621: Parse error for unsigned right shift assignment
Hi,

Following code results in parse error;
```js
let a = 5; //  00000000000000000000000000000101

a >>>= 2;  //  00000000000000000000000000000001
console.log(a);
// expected output: 1

let b = -5; // -00000000000000000000000000000101

b >>>= 2;   //  00111111111111111111111111111110
console.log(b);
// expected output: 1073741822
```
thanks

arpit-saxena:
This Pull Request fixes #621 

It changes the following:

- Adds support for unsigned right shift assignment operator (>>>=)
- Adds a test for the operator

### DIFF
--- a/boa/src/exec/operator/mod.rs
+++ b/boa/src/exec/operator/mod.rs
@@ -173,6 +173,7 @@ impl BinOp {
             AssignOp::Xor => x.bitxor(&y, interpreter),
             AssignOp::Shl => x.shl(&y, interpreter),
             AssignOp::Shr => x.shr(&y, interpreter),
+            AssignOp::Ushr => x.ushr(&y, interpreter),
         }
     }
 }
--- a/boa/src/syntax/ast/op.rs
+++ b/boa/src/syntax/ast/op.rs
@@ -938,7 +938,18 @@ pub enum AssignOp {
     /// [spec]: https://tc39.es/ecma262/#prod-AssignmentOperator
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Assignment_Operators#Right_shift_assignment
     Shr,
-    // TODO: Add UShl (unsigned shift left).
+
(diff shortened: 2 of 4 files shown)


arpit-saxena:
https://github.com/boa-dev/boa/blob/3f6a8ef06bceb0171ffb9f6ca9037d813ab1bc6b/boa/src/exec/tests.rs#L1302-L1316
This test was ignored due to >>>= being unimplemented. It looks like it is outdated though given the error messages are a bit different than what it looks for. The error messages all begin with `Uncaught "TypeError": `, should I add this to the test?

HalidOdat (approved):

HalidOdat:
> This test was ignored due to >>>= being unimplemented. It looks like it is outdated though given the error messages are a bit different than what it looks for. The error messages all begin with `Uncaught "TypeError": `, should I add this to the test?

This test is for invalid assign operations like `2 += 2` this should generate a syntax error but since we don't have all the operators implemented we ignore it for now. we should have a test for `>>>=` execution like:
```js
let x = -1000;
x >>>= 1;
```
and asserting that `x` equals `2147483148` or something like this.

arpit-saxena:
> we should have a test for `>>>=` execution like:
> 
> ```js
> let x = -1000;
> x >>>= 1;
> ```
> 
> and asserting that `x` equals `2147483148` or something like this.

I can add that but if I do that, I think there should tests similar to this for other operators too, like `>>=`, and I can't find any of them currently. Is there a need to add tests for all the operators?

HalidOdat:
> I can add that but if I do that, I think there should tests similar to this for other operators too, like `>>=`, and I can't find any of them currently. Is there a need to add tests for all the operators?

There should be some tests for some operators in `value/tests.rs` and in `exec/tests.rs`.

> Is there a need to add tests for all the operators?

More tests is always nice to have.

Issue #577, opened 2020-07-19
Decouple `Object` from `Value`


HalidOdat:
**What is the problem with the current implementation?**
The problem is that we treat `Value` as an `Object` which is not always the case. for example `Value` has a `.set_field()` which sets an object field, this does not make sense for a couple of reasons:
 - what if `Value` is not an object should calling `.set_field` panic?, should it ignore it? should it return an option or result? or something else? (we ignore non object values which can become a really hard bug to debug), as we can see it does not make sense for `Value` to have `Object` behaviour
 - This also forces us to do checks if the value is an object.

If we need object behaviour we can call `as_object` method.

Changes:
 - Move all object behaivour from `Value` to `Object`
 - Places were we need (or places were the type is only an object for example: the global object) should be `GcObject`s
 - ~~function like `.to_object` should return a `GcObject` not `Value`~~ done

Doing these thing should remove unnecessary checks and make the API more intuitive

Any suggestions on how to improve this?

RageKnify:
Hey @HalidOdat I found this interesting and am trying to tackle it.
To ensure I'm following your "vision", methods from builtins like `Array` would stop receiving `&Value` as `this` and instead receive `&Object`/`&mut Object`, right?

jasonwilliams:
@HalidOdat how are you handling boxing, unboxing of primitive values? I think I left set_field and get_field on value so things like `“hello world ”.trim()` would still work

HalidOdat:
> @HalidOdat how are you handling boxing, unboxing of primitive values? I think I left set_field and get_field on value so things like `“hello world ”.trim()` would still work

We store some primitives on the stack (number, boolean, undefined, null) and string, bigint, symbol on the heap with Rcs (since they can't have cycles, less work for the gc) and object with `Gc`. we disused this in #465 .

Why would removing `.set_field` and `.get_field` cause `“hello world ”.trim()` to break?

jasonwilliams:
Ok sounds good.

> Why would removing .set_field and .get_field cause “hello world ”.trim() to break?

I think originally the unboxing happened inside of get_field() where the value was checked and upgraded to an object of it was a primitive. `.trim()` would have called get_field(“trim”)

HalidOdat:
> Hey @HalidOdat I found this interesting and am trying to tackle it.
> To ensure I'm following your "vision", methods from builtins like `Array` would stop receiving `&Value` as `this` and instead receive `&Object`/`&mut Object`, right?

[...]

PR #712, opened 2020-09-24
Refactor: Value::to_object to return GcObject

RageKnify:
This Pull Request starts the work on #577.

It changes the following:

- `Value::to_object` now returns `Result<GcObject>` rather than `Result<Value>`
- Move `Value::new_object_from_prototype` to `Object::new_object_from_prototype`
- Minor changes in other files to accomodate the new API

I assume those minor changes will be cleaned up once the separation of the `Object` and `Value` API is complete.

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -960,7 +960,8 @@ impl Array {
     /// [spec]: https://tc39.es/ecma262/#sec-array.prototype.reduce
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce
     pub(crate) fn reduce(this: &Value, args: &[Value], interpreter: &mut Context) -> Result<Value> {
-        let this = this.to_object(interpreter)?;
+        let obj_this = this.to_object(interpreter)?;
+        let this = Value::Object(obj_this);
         let callback = match args.get(0) {
             Some(value) if value.is_function() => value,
             _ => return interpreter.throw_type_error("Reduce was called without a callback"),
@@ -1022,7 +1023,8 @@ impl Array {
         args: &[Value],
         interpreter: &mut Context,
     ) -> Result<Value> {
-        let this = this.to_object(interpreter)?;
+        let obj_this = this.to_object(interpreter)?;
+        let this = Value::Object(obj_this);
         let callback = match args.get(0) {
             Some(value) if value.is_function() => value,
             _ => return interpreter.throw_type_error("reduceRight was called without a callback"),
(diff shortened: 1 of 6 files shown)


Lan2u (approved):

RageKnify:
There's one thing I started doubting after implementing, it may be more appropriate for the return type to be `Object` and then to wrap it in a `GcObject` only when required.

boa/src/builtins/array/mod.rs
```diff
@@ -960,7 +960,8 @@ impl Array {
     /// [spec]: https://tc39.es/ecma262/#sec-array.prototype.reduce
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce
     pub(crate) fn reduce(this: &Value, args: &[Value], interpreter: &mut Context) -> Result<Value> {
-        let this = this.to_object(interpreter)?;
+        let obj_this = this.to_object(interpreter)?;
+        let this = Value::Object(obj_this);
```
HalidOdat, line 964:
```suggestion
        let this = this.to_object(interpreter)?.into();
```
we can `Into<Value>` for this.

boa/src/builtins/array/mod.rs
```diff
@@ -1022,7 +1023,8 @@ impl Array {
         args: &[Value],
         interpreter: &mut Context,
     ) -> Result<Value> {
-        let this = this.to_object(interpreter)?;
+        let obj_this = this.to_object(interpreter)?;
+        let this = Value::Object(obj_this);
```
HalidOdat, line 1027:
```suggestion
        let this = this.to_object(interpreter)?.into();
```
same here

boa/src/builtins/object/mod.rs
```diff
@@ -33,7 +33,7 @@ impl Object {
     pub fn make_object(_: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         if let Some(arg) = args.get(0) {
             if !arg.is_null_or_undefined() {
-                return arg.to_object(ctx);
+                return Ok(Value::Object(arg.to_object(ctx)?));
```
HalidOdat, line 36:
```suggestion
                return Ok(arg.to_object(ctx)?.into());
```

boa/src/object/mod.rs
```diff
@@ -401,6 +401,14 @@ impl Object {
         self.prototype = prototype
     }
 
+    /// Similar to `Value::new_object`, but you can pass a prototype to create from, plus a kind
+    pub fn new_object_from_prototype(proto: Value, data: ObjectData) -> GcObject {
```
HalidOdat, line 405:
```suggestion
    #[inline]
    pub fn new_object_from_prototype(proto: Value, data: ObjectData) -> GcObject {
```
we should add inline so it can be inlined in other crates

HalidOdat:
> There's one thing I started doubting after implementing, it may be more appropriate for the return type to be `Object` and then to wrap it in a `GcObject` only when required. What do you think @HalidOdat ?

You mean `new_object_from_prototype`, right? I thick so too, but for `.to_object` it has to return `GcObject`
Also I think it's best to rename `new_object_from_prototype` to `with_prototype` so when we construct it its
```rust
let object = Object::with_prototype(Value::null(), ...);
```

HalidOdat (approved):

PR #714, opened 2020-09-25
perf(boa): test ahash perf change

neeldug:
opportunities to further improve performance are to use the trait
`ahash::CallHasher` for optimised hashing

It changes the following:

- Switches FxHash current usage with aHash

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,6 +4,7 @@
 name = "Boa"
 version = "0.9.0"
 dependencies = [
+ "ahash",
  "bitflags",
  "chrono",
  "criterion",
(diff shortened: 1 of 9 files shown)


Lan2u:
The benchmarks indicate this has a negative effect on performance - did you do any local benchmarking? What were the results?

neeldug:
Yeah I saw this and it seems to be a similar reason as to why rustc avoided using ahash and stuck with FxHash instead, local benchmarks show similar results with less variation, I’m gonna try 1 change to see if that improves performance at all.

neeldug:
aHash doesn't seem to be the right way to move on this, gonna close this for now!

Issue #716, opened 2020-09-26
Strict Mode Lex/Parse


Lan2u:
**ECMASCript feature**
As per https://tc39.es/ecma262/#sec-strict-mode-code and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode ECMAScript can be processed using either strict or unstrict mode. When parsing/lexing we currently only account for the strict case in a few select areas such as number lexing however in-order for boa to support strict mode fully we need to expand this to other areas.

**Example code**
For example, the following code is processed by boa fine when it should produce a SyntaxError as 'implements' is a reserved keyword in strict mode code.

```javascript
    'use strict';
    var implements;
```

Initially, I'm planning to focus on the lexing/parsing side and then the interpreting side can come after.

- [x] Strict mode number lexing/parsing
- [x] Strict mode dis-allowed keywords
- [x] Strict mode identifier lexing/parsing
- [x] Strict mode directive-prologue parsing
- [x] Strict mode function argument parsing

Will add more as I come across them

PR #717, opened 2020-09-26
Strict Mode Lex/Parse

Lan2u:
This Pull Request addresses #716.

It changes the following:
- Add strict mode handling to lex/parsing code


Razican:
One thing that is needed is a way for knowing if we are in strict mode from the parser, since we do need to change some parsing semantics.

Lan2u:
> One thing that is needed is a way for knowing if we are in strict mode from the parser, since we do need to change some parsing semantics.

For the parser this should be fine as the use strict directive is read at the parser level, my thought was that we can pass strict_mode as an argument between parse functions with a flag on the parser used if global strict mode is enabled. 

What might be more complicated is enforcing stuff like "strict mode makes attempts to delete undeletable properties throw" as we may not be able to do this during parsing.

Razican:
> For the parser this should be fine as the use strict directive is read at the parser level, my thought was that we can pass strict_mode as an argument between parse functions with a flag on the parser used if global strict mode is enabled. 

We already pass the lexer as an argument, and the strict mode is relevant when lexing. I would add a `strict_mode` attribute in the lexer, that can be set and retrieved by the parser, similarly to goal symbols.

> What might be more complicated is enforcing stuff like "strict mode makes attempts to delete undeletable properties throw" as we may not be able to do this during parsing.

I think that we should have a `strict_mode` flag in the interpreter, that could be retrieved from each node executors.

If I understand the strict mode properly, it starts to take effect once it is declared, but I'm not sure about the scope. If it has a function scope, then the parser must be able to set it to false after parsing the function, and the executor should be able to check the string literal to set its own flag on execution.

Lan2u:
> > For the parser this should be fine as the use strict directive is read at the parser level, my thought was that we can pass strict_mode as an argument between parse functions with a flag on the parser used if global strict mode is enabled.
> 
> We already pass the lexer as an argument, and the strict mode is relevant when lexing. I would add a `strict_mode` attribute in the lexer, that can be set and retrieved by the parser, similarly to goal symbols.
> 
> > What might be more complicated is enforcing stuff like "strict mode makes attempts to delete undeletable properties throw" as we may not be able to do this during parsing.
> 
> I think that we should have a `strict_mode` flag in the interpreter, that could be retrieved from each node executors.
> 
> If I understand the strict mode properly, it starts to take effect once it is declared, but I'm not sure about the scope. If it has a function scope, then the parser must be able to set it to false after parsing the function, and the executor should be able to check the string literal to set its own flag on execution.

I put strict_mode as a flag to lex() which I think might work better than trying to make sure we set a flag on the parser when we exit function scope. This way if the entire script is strict mode then we just set it to true at the top level and it carries through and if only a function is strict mode then we just set the flag to true when parsing the function statement list and this way it won't accidentally apply to too wide a scope. Also to me having it included in lex/parse means that if someone later wants to add another thing to lex/parse it is apparent that the strict mode case needs to be handled and how to do so is easy to understand.

boa/src/syntax/parser/expression/mod.rs
```diff
@@ -61,7 +61,7 @@ macro_rules! expression { ($name:ident, $lower:ident, [$( $op:path ),*], [$( $lo
     {
         type Output = Node;
 
-        fn parse(self, cursor: &mut Cursor<R>) -> ParseResult {
+        fn parse(self, cursor: &mut Cursor<R>)-> ParseResult {
```
RageKnify, line 64: Surprised that `rustfmt` removed the space before the arrow. But it's passing CI so it must have...

boa/src/syntax/parser/mod.rs
```diff
@@ -121,10 +121,19 @@ where
...
     fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
-        if cursor.peek(0)?.is_some() {
-            ScriptBody.parse(cursor)
-        } else {
-            Ok(StatementList::from(Vec::new()))
+        match cursor.peek(0)? {
+            Some(tok) => {
+                match tok.kind() {
+                    TokenKind::StringLiteral(string) | TokenKind::TemplateLiteral(string) => {
+                        if string == &"use strict".into() {
```
RageKnify, line 128: Could this line be changed to `string.as_str() == "use strict"`? I think it looks cleaner.

boa/src/syntax/parser/statement/variable/mod.rs
```diff
@@ -171,6 +171,8 @@ where
 
         let name = BindingIdentifier::new(self.allow_yield, self.allow_await).parse(cursor)?;
 
+        dbg!(name.clone());
+
```
RageKnify, line 175: Left behind after debugging.

RageKnify (approved):
Looks good to me (other than the forgotten `dbg!()`), but it's the first time I look inside the `syntax` module, definitely wait for someone who knows more about it.

boa/src/syntax/parser/mod.rs
Lan2u: .as_str won't work because string is of type Box<str> but .to_string() will and I agree it would be cleaner

boa/src/syntax/parser/statement/variable/mod.rs
Lan2u: Good spot!

boa/src/syntax/parser/expression/mod.rs
Lan2u: Huh, that is a bit strange

boa/src/syntax/parser/expression/unary.rs
```diff
@@ -61,11 +61,23 @@ where
     fn parse(self, cursor: &mut Cursor<R>) -> ParseResult {
         let _timer = BoaProfiler::global().start_event("UnaryExpression", "Parsing");
 
-        let tok = cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?;
+        // TODO: can we avoid cloning?
+        let tok = cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?.clone();
```
HalidOdat, line 65: why do we have to clone here?

boa/src/syntax/parser/mod.rs
RageKnify: But `.to_string()` will result in an allocation, right? I'm looking into how to avoid it.

boa/src/syntax/parser/expression/unary.rs
Lan2u: Because we later use the token() span as part of the error message, I've changed this to an alternative mechanism which avoids the clone (still needs an implicit copy).

boa/src/syntax/parser/mod.rs
RageKnify: `&*string == "use strict"` should work fine.
Taken from here: https://stackoverflow.com/a/32085042

Lan2u: Changed to .as_ref()

Lan2u: > `&*string == "use strict"` should work fine.
> Taken from here: https://stackoverflow.com/a/32085042

Hmm, I wonder what the performance difference is to .as_ref() - &* seems like strange notation!

boa/src/syntax/parser/expression/unary.rs
HalidOdat: should be fine start is a usize at worst a u64 so I don't think this is ever going to be an issue :)

HalidOdat (approved):
Looks perfect to me!

boa/src/syntax/parser/mod.rs
RageKnify: It is strange notation, but I think the codegen should be equal.
My concern was avoiding alocation, and I don think `.as_ref()` is clearer.

PR #722, opened 2020-09-27
`ConstructorBuilder`, `ObjectInitializer`, cache standard objects and fix global object attributes

### LINKED ISSUE #613: Cache standard objects (`Array`, `Object`, etc) and there prototypes

HalidOdat:
This Pull Request fixes/closes #613.

It changes the following:
 - Add BuiltIn trait for builtins
 - Fixed global properties/objects/object methods/object properties attributes
 - Made `ClassBuilder` retun `&mut Self`
 - Add ConstructorBuilder for constructing constructors (builder pattern)
 - Add ObjectInitializer for constructing objects (builder pattern) 
 - Add FunctionBuilder for constructing function objects (builder pattern) 
 - Cache core standard objects

### DIFF
--- a/boa/src/builtins/error/reference.rs
+++ b/boa/src/builtins/error/reference.rs
@@ -10,24 +10,50 @@
 //! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ReferenceError
 
 use crate::{
-    builtins::{function::make_builtin_fn, function::make_constructor_fn},
-    object::ObjectData,
+    builtins::BuiltIn,
+    object::{ConstructorBuilder, ObjectData},
     profiler::BoaProfiler,
+    property::Attribute,
     Context, Result, Value,
 };
 
 #[derive(Debug, Clone, Copy)]
 pub(crate) struct ReferenceError;
 
-impl ReferenceError {
-    /// The name of the object.
-    pub(crate) const NAME: &'static str = "ReferenceError";
+impl BuiltIn for ReferenceError {
+    const NAME: &'static str = "ReferenceError";
 
+    fn attribute() -> Attribute {
+        Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE
+    }
+
+    fn init(context: &mut Context) -> (&'static str, Value, Attribute) {
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
+
+        let error_prototype = context.standard_objects().error_object().prototype();
+        let attribute = Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE;
+        let reference_error_object = ConstructorBuilder::with_standard_object(
+            context,
+            Self::constructor,
+            context.standard_objects().reference_error_object().clone(),
+        )
+        .name(Self::NAME)
+        .length(Self::LENGTH)
+        .inherit(error_prototype.into())
+        .property("name", Self::NAME, attribute)
+        .property("message", "", attribute)
+        .build();
+
+        (Self::NAME, reference_error_object.into(), Self::attribute())
+    }
+}
+
+impl ReferenceError {
     /// The amount of arguments this function object takes.
     pub(crate) const LENGTH: usize = 1;
 
     /// Create a new error object.
-    pub(crate) fn make_error(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
+    pub(crate) fn constructor(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         if let Some(message) = args.get(0) {
             this.set_field("message", message.to_string(ctx)?);
         }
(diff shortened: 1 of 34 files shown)


HalidOdat:

Created some issues to track what was been fixed/improved, and looking at how many there are... maybe I overdid it :sweat_smile:

boa/src/object/internal_methods.rs
```diff
@@ -336,7 +336,7 @@ impl Object {
 
     /// Helper function for property insertion.
     #[inline]
-    pub(crate) fn insert_property<K>(&mut self, key: K, property: Property) -> Option<Property>
+    pub(crate) fn insert<K>(&mut self, key: K, property: Property) -> Option<Property>
```
RageKnify, line 339: `Object::insert`  inserts a `Property`, but `Object::insert_property`, inserts `V: Into<Value>`.
Seems pretty confusing to me, should switch them up, no?

boa/src/context.rs
```diff
@@ -27,11 +27,130 @@ use crate::{
...
+}
+
 /// Javascript context. It is the primary way to interact with the runtime.
 ///
-/// For each `Context` instance a new instance of runtime is created.
-/// It means that it is safe to use different contexts in different threads,
-/// but each `Context` instance must be used only from a single thread.
+/// `Context`s constructed in a thread share the same runtime, therefore it
+/// is possible to share objects from one context to another context, but they
+/// have to in the same thread.
```
RageKnify, line 153: \*have to ***be*** in the same thread.

boa/src/class.rs
```diff
@@ -108,157 +107,81 @@ impl<T: Class> ClassConstructor for T {
...
-                .get_field("prototype"),
-        );
-
-        function.insert_field("length", Value::from(length));
-        function.insert_field("name", Value::from(name.as_str()));
-
-        self.prototype
-            .borrow_mut()
-            .insert_field(name, Value::from(function));
+        // TODO: \/
```
RageKnify, line 135: Can you detail the `TODO` or was it from when you were implementing it?

boa/src/builtins/map/mod.rs
```diff
@@ -15,11 +15,92 @@ mod tests;
...
+    /// Create a new map
+    pub(crate) fn constructor(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
+        // Make a new Object which will internally represent the Array (mapping
+        // between indices and values): this creates an Object with no prototype
+
+        // Set Prototype
+        let prototype = ctx.global_object().get_field("Map").get_field(PROTOTYPE);
+
+        this.as_object_mut()
+            .expect("this is array object")
```
RageKnify, line 56: Should be `"this is map object"`, right?

boa/src/builtins/error/type.rs
```diff
@@ -16,24 +16,50 @@
...
+    fn attribute() -> Attribute {
+        Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE
+    }
+
+    fn init(context: &mut Context) -> (&'static str, Value, Attribute) {
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
+        let error_prototype = context.standard_objects().error_object().prototype();
+        let attribute = Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE;
+        let range_error_object = ConstructorBuilder::with_standard_object(
```
RageKnify, line 41: `range_error_object` should be `type_error_object`, right?

boa/src/builtins/error/syntax.rs
```diff
@@ -12,25 +12,51 @@
...
+    fn attribute() -> Attribute {
+        Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE
+    }
+
+    fn init(context: &mut Context) -> (&'static str, Value, Attribute) {
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
+        let error_prototype = context.standard_objects().error_object().prototype();
+        let attribute = Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE;
+        let range_error_object = ConstructorBuilder::with_standard_object(
```
RageKnify, line 38: `range_error_object` should be `syntax_error_object`, like in `type.rs`

boa/src/builtins/error/reference.rs
```diff
@@ -10,24 +10,50 @@
...
+    fn attribute() -> Attribute {
+        Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE
+    }
+
+    fn init(context: &mut Context) -> (&'static str, Value, Attribute) {
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
+        let error_prototype = context.standard_objects().error_object().prototype();
+        let attribute = Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE;
+        let range_error_object = ConstructorBuilder::with_standard_object(
```
RageKnify, line 35: `range_error_object` should be `reference_error_object`

RageKnify (changes requested):
There are some typos and a few things I didn't fully understand.

boa/src/builtins/mod.rs
```diff
@@ -27,61 +27,72 @@ pub(crate) use self::{
...
 
 /// Initializes builtin objects and functions
 #[inline]
-pub fn init(interpreter: &mut Context) {
-    let globals = [
-        // The `Function` global must be initialized before other types.
-        function::init,
-        Object::init,
+pub fn init(context: &mut Context) {
+    let globals2 = [
```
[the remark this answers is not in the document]
HalidOdat: > Any reason to change the name from `globals` to `globals2`?

What was when I transitioning to the cached standard objects, I'll change it :)

boa/src/object/internal_methods.rs
HalidOdat: > `Object::insert` inserts a `Property`, but `Object::insert_property`, inserts `V: Into<Value>`.
> Seems pretty confusing to me, should switch them up, no?

Its is kind of confuzing but `inset` does not insert a property it inserts a Property Descriptor (this is why `Property` should be called `PropertyDescriptor` which is the way object as stored as a `HashMap<PropertyKey, PropertyDescriptor>`), and `insert_property` inserts a property like `x: "foo"`

boa/src/context.rs
HalidOdat: > *have to _**be**_ in the same thread.

typo :sweat_smile:

boa/src/class.rs
HalidOdat: > Can you detail the `TODO` or was it from when you were implementing it?

It was something I wanted to implement, I'll do it now :)

boa/src/builtins/map/mod.rs
HalidOdat: > Should be `"this is map object"`, right?

It was there before i started changing things, I'll change it

boa/src/builtins/error/syntax.rs
HalidOdat: same here

boa/src/builtins/error/reference.rs
HalidOdat: and here

boa/src/builtins/error/type.rs
HalidOdat: Copy paste mistake

RageKnify (approved):

JohnDoneth:
I like the switch to builders instead of calling `make_builtin_fn` and like functions a bunch. Code is clear and concise! 👍

PR #728, opened 2020-09-28
Fix hanging test262 bug by setting the length on String objects

### LINKED ISSUE #720: length is not updating properly on string within while loop
**Describe the bug**
while loops do not seem to be working properly. The condition doesn't seem to re-run causing infinite loops in some places. Tested on 01b5cec04f079b7efe9fe643b310a7e9ce2a415a

**To Reproduce**

```javascript
function addZeros() {
    var s = "";
    while (s.length < 4) {
      s = "0" + s;
    }
    return s;
  }

  addZeros()

```

`s` is being extended but its still looping forever

This is what is breaking test262

jasonwilliams:
- Set a length on new String objects
- Add a test

This Pull Request fixes/closes #720.

### DIFF
--- a/boa/src/value/mod.rs
+++ b/boa/src/value/mod.rs
@@ -713,10 +713,10 @@ impl Value {
                     .expect("String was not initialized")
                     .get_field(PROTOTYPE);
 
-                Ok(GcObject::new(Object::with_prototype(
-                    proto,
-                    ObjectData::String(string.clone()),
-                )))
+                let mut obj = Object::with_prototype(proto, ObjectData::String(string.clone()));
+                // Make sure the correct length is set on our new string object
+                obj.set(PropertyKey::from("length"), string.len().into());
+                Ok(GcObject::new(obj))
             }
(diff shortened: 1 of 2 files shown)


boa/src/value/mod.rs
```diff
@@ -713,10 +713,10 @@ impl Value {
                     .expect("String was not initialized")
                     .get_field(PROTOTYPE);
 
-                Ok(GcObject::new(Object::with_prototype(
-                    proto,
-                    ObjectData::String(string.clone()),
-                )))
+                let mut obj = Object::with_prototype(proto, ObjectData::String(string.clone()));
+                // Make sure the correct length is set on our new string object
+                obj.set(PropertyKey::from("length"), string.chars().count().into());
```
HalidOdat, line 718:
```suggestion
                obj.set("length".into(), string.chars().count().into());
```
It's cleaner if we use `.into()` instead of `from`

HalidOdat (approved):

PR #735, opened 2020-09-29
Move exec implementations together with AST node structs

### LINKED ISSUE #721: Move executable implementations with the rest of the struct impl
It can be confusing to have the `Executable` implementation live elsewhere.
An example is  `Return`  which lives under the `node` directory https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/ast/node/return_smt.rs 

However clicking through to Return from [`exec`](https://github.com/boa-dev/boa/blob/master/boa/src/exec/mod.rs#L110) doesn't take you to the executable implementation, instead it takes you to the original implementation above.
The executable implementation lives here:
https://github.com/boa-dev/boa/blob/master/boa/src/exec/return_smt/mod.rs and should just in with the rest of `Return` to make developer experience easier when navigating the code base.

Same for all of the Nodes we have.

## Steps
Move [Executable implementations](https://github.com/boa-dev/boa/tree/master/boa/src/exec) to be in the same file as the struct, usually here: https://github.com/jasonwilliams/boa/tree/master/boa/src/syntax/ast/node

georgeroman:
Fixes #721

It changes the following:

- Moved `Executable` implementations for nodes in the same files as node declarations
- Broke down some files (`ast/node/declaration.rs`, `ast/node/iteration.rs`, ...) into multiple files (one per node)


HalidOdat (approved):

georgeroman:
Can I ask for a faster review on this? As a lot of files were moved, it's hard to always fix conflicts and keep this PR in sync.

boa/src/exec/statement_list.rs
```diff
@@ -0,0 +1,42 @@
...
+                }
+            }
+            if i + 1 == self.statements().len() {
+                obj = val;
+            }
+        }
+
+        Ok(obj)
+    }
+}
```
RageKnify: I think you accidentaly re-introduced this file while merging @georgeroman, you've moved it here: https://github.com/georgeroman/boa/blob/afe9daf3fa07cdd6346f29907df56b6fd7bd7c78/boa/src/syntax/ast/node/statement_list/mod.rs#L59-L95
I think you should delete this file, right?

RageKnify: GitHub says my comment is outdated, but I can see the file here: https://github.com/georgeroman/boa/blob/move_exec_to_ast/boa/src/exec/statement_list.rs

georgeroman: Yes, sorry, I forgot to delete it.

georgeroman:
Looks like an unrelated CI fail (https://github.com/boa-dev/boa/pull/735/checks?check_run_id=1198128772).

RageKnify (approved):

PR #737, opened 2020-09-29
Fix output of Object.toString()

### LINKED ISSUE #729: obj.toString() should be "[object Object]" S8.6.2_A3
During Test262 we have https://github.com/tc39/test262/blob/master/test/language/types/object/S8.6.2_A3.js failing.
You can test this by setting your js file to...

```
let obj = {};
obj.toString();
```

The output should be "[object Object]"

The issue here is we're using the internal display rather than what the JS developer should be seeing.
https://github.com/boa-dev/boa/blob/master/boa/src/builtins/object/mod.rs#L120 could be changed to "[object Object]"

We will need to implement a way for us to debug the insides of objects, but this should be left to another method, like `print()` etc

RageKnify:
This Pull Request fixes/closes #729.

It changes the following:

- `Object::to_string`returns a `Value::String` respecting the spec: https://tc39.es/ecma262/#sec-object.prototype.tostring

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -115,9 +115,9 @@ impl Object {
     /// [spec]: https://tc39.es/ecma262/#sec-object.prototype.tostring
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString
     #[allow(clippy::wrong_self_convention)]
-    pub fn to_string(this: &Value, _: &[Value], _: &mut Context) -> Result<Value> {
+    pub fn to_string(_: &Value, _: &[Value], _: &mut Context) -> Result<Value> {
         // FIXME: it should not display the object.
-        Ok(this.display().to_string().into())
+        Ok("[object Object]".into())
     }
 
     /// `Object.prototype.hasOwnPrototype( property )`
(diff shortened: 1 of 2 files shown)


boa/src/builtins/object/mod.rs
```diff
@@ -115,9 +115,9 @@ impl Object {
     /// [spec]: https://tc39.es/ecma262/#sec-object.prototype.tostring
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString
     #[allow(clippy::wrong_self_convention)]
-    pub fn to_string(this: &Value, _: &[Value], _: &mut Context) -> Result<Value> {
+    pub fn to_string(_: &Value, _: &[Value], _: &mut Context) -> Result<Value> {
         // FIXME: it should not display the object.
```
Razican, line 119: I think we can remove this comment, right?

Razican (changes requested):
I like this approach, but now that we are at it, I think we should implement it to be spec compliant:
https://tc39.es/ecma262/#sec-object.prototype.tostring

boa/src/builtins/object/mod.rs
RageKnify: Guess so.

RageKnify:
> I like this approach, but now that we are at it, I think we should implement it to be spec compliant:
> https://tc39.es/ecma262/#sec-object.prototype.tostring

Yeah, my bad to not look for it.

RageKnify:
I assume one of the bugs in `tarpaulin` makes it so lines 129 through 136 in `boa/src/builtins/object/mod.rs` are considered not covered since they are covered by the introduced test.

Razican (approved):
Looks good from my side! Thanks!

HalidOdat (approved):

RageKnify:
Output for `[object Error]` is currently not tested, I think I'll include it in #749 where I'll introduce the changes to the error constructors discussed in Discord.

PR #746, opened 2020-10-01
Add Object.defineProperties and handle props argument in Object.create

### LINKED ISSUE #552: [builtins - Object]  Object.defineProperties() not implemented

dvtkrlbs:
This Pull Request fixes/closes #552.

It changes the following:

- It adds define_properties to both internal and external Object struct

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -85,7 +85,7 @@ impl Object {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-object.create
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
-    pub fn create(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
+    pub fn create(_: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         let prototype = args.get(0).cloned().unwrap_or_else(Value::undefined);
         let properties = args.get(1).cloned().unwrap_or_else(Value::undefined);
 
@@ -103,7 +103,7 @@ impl Object {
         };
 
         if !properties.is_undefined() {
-            return Object::define_properties(this, &[obj, properties], ctx);
+            return Object::define_properties(&Value::Undefined, &[obj, properties], ctx);
         }
 
         Ok(obj)
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -85,7 +85,7 @@ impl Object {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-object.create
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
-    pub fn create(this: &Value, args: &[Value], mut ctx: &mut Context) -> Result<Value> {
+    pub fn create(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         let prototype = args.get(0).cloned().unwrap_or_else(Value::undefined);
         let properties = args.get(1).cloned().unwrap_or_else(Value::undefined);
 
@@ -102,8 +102,8 @@ impl Object {
             }
         };
 
-        if properties != Value::Undefined {
-            return Object::define_properties(this, &[obj, properties], &mut ctx);
+        if !properties.is_undefined() {
+            return Object::define_properties(this, &[obj, properties], ctx);
         }
 
         Ok(obj)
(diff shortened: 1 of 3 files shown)


dvtkrlbs:
I am looking for feedback on this. I might have missed some stuff since spec is kinda confusing

dvtkrlbs:
Ok i think this is ready for review now. I know that i did not add any tests but i have no idea on how to test this.

boa/src/object/internal_methods.rs
```diff
@@ -282,6 +282,52 @@ impl Object {
...
+        let mut prop_keys = Vec::with_capacity(size);
+
+        self.index_properties()
+            .for_each(|(p, _)| prop_keys.push(PropertyKey::from(*p)));
+        self.string_properties()
+            .for_each(|(p, _)| prop_keys.push(PropertyKey::from(p.as_str())));
+        self.symbol_properties()
+            .for_each(|(p, _)| prop_keys.push(PropertyKey::from(p.clone())));
+
+        prop_keys
```
HalidOdat, line 304:
```suggestion
        self.keys().collect()
```
we can use the `keys` iterator method

boa/src/object/internal_methods.rs
```diff
@@ -282,6 +282,52 @@ impl Object {
...
+    }
+
+    /// The abstract operation ObjectDefineProperties
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-object.defineproperties
+    pub fn define_properties(&mut self, props: &Self) {
```
HalidOdat, line 313:
```suggestion
    pub fn define_properties(&mut self, properties: Value) {
```
The properties field can be any value, and `Value::to_object()` is called on it.

boa/src/builtins/object/mod.rs
```diff
@@ -104,6 +108,21 @@ impl Object {
...
+    pub fn define_properties(_: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
+        let val = args.get(0).expect("Cannot get object");
+        if let Some(mut obj) = val.as_object_mut() {
+            let props = args
+                .get(1)
+                .expect("Cannot get props")
+                .as_object_mut()
+                .unwrap();
+            obj.define_properties(&props);
+            Ok(val.clone())
```
HalidOdat, line 121:
```suggestion
            let props = args.get(1).cloned().unwrap_or_else(Value::undefined);
            obj.define_properties(&props);
            Ok(val.clone())
```
we should not panic here, with `.expect`

dvtkrlbs:
> ```suggestion
    pub fn define_properties(&mut self, properties: Value) {
```
The properties field can be any value, and `Value::to_object()` is called on it. 

That also requires a context should i pass one as an argument?

File: boa/src/object/internal_methods.rs (hunk 2 above)
Comment (dvtkrlbs) [reply boa/src/object/internal_methods.rs:313]: That also requires a context should i pass one as an argument?

File: boa/src/builtins/object/mod.rs [hunk 4]
```diff
@@ -104,6 +108,21 @@ impl Object {
         Ok(Value::undefined())
     }
 
+    /// Define multiple properties
+    pub fn define_properties(_: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
+        let val = args.get(0).expect("Cannot get object");
+        if let Some(mut obj) = val.as_object_mut() {
```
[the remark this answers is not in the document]
dvtkrlbs: We need to return the object from this function if nothing fails. But we cannot get a reference to the val object this way. Should i do just `Ok(args.get(0).unwrap())`.

boa/src/object/internal_methods.rs
HalidOdat: > That also requires a context should i pass one as an argument?

yes. we need to pass it, and return Result.

boa/src/builtins/object/mod.rs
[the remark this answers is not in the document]
HalidOdat: > We need to return the object from this function if nothing fails. But we cannot get a reference to the val object this way. Should i do just `Ok(args.get(0).unwrap())`.

```rust
let arg = args.get(0).cloned().unwrap_or_else(Value::undefined);
if let Some(mut obj) = arg.as_object_mut() {
	// ...
	Ok(arg.clone())
}
```
This was we dont panic and we remove `unwrap`

boa/src/object/internal_methods.rs
dvtkrlbs: Hmm what do you mean it should return a result. I understand that it can return error but what should it return if it succeeds. According to the spec it should return the object but i could not find any way to return it to the value all methods wants a ownership.

HalidOdat: > Hmm what do you mean it should return a result. I understand that it can return error but what should it return if it succeeds. According to the spec it should return the object but i could not find any way to return it to the value all methods wants a ownership.

We can return `Result<()>`.

HalidOdat (approved):

RageKnify (approved):

RageKnify:
@dvtkrlbs for a test you can look at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties for an idea of how to test it.
I'd say to write the test in `boa/src/builtins/object/tests.rs`.

RageKnify (changes requested):
Adding 1 test should be enough to improve the coverage values.

boa/src/builtins/object/mod.rs
```diff
@@ -84,24 +85,28 @@ impl Object {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-object.create
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
-    pub fn create(_: &Value, args: &[Value], interpreter: &mut Context) -> Result<Value> {
+    pub fn create(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
```
jasonwilliams, line 88: I'm a bit confused why `create` accepts a `this` argument. Isn't it a static method?

RageKnify: I think it's because it needs to invoke `Object::define_properties`, which will need a `this` because it is also a BuiltIn function.

jasonwilliams: But looking at line 158 it doesn’t look like it does anything with the first value

RageKnify: Would you preffer to not have `this` in `create` and invoke `define_properties` with `&Value::Undefined`, or some other non-relevant value?

RageKnify (approved):

HalidOdat (approved):

PR #749, opened 2020-10-01
Fix Error constructors to return rather than throw

RageKnify:
On `master` `let e = new Error('test');` doesn't work and results in an uncaught error, this PR fixes that by making it so the Error constructors aren't seen as throwing their return value by `boa`.

Originally opened because `SyntaxError.toString` and `Error.toString` were using `Value::display` while all other error types use `Value::to_string`.
I don't think this has any effect right now, but I think the errors should all conform to the same pattern.

Has an example you can see `TypeError`'s implementation below:
https://github.com/boa-dev/boa/blob/e83a03ad6f14c05b4829802f38d2c4cfbe916759/boa/src/builtins/error/type.rs#L58-L63

It changes the following:
- Change error constructors to wrap their return in `Ok()` rather than `Err()`
- ~~Change conversion of `name` and `message` to use `Value::to_string` rather than `Value::display` in `SyntaxError.toString`~~
- With the new Builin patterns Error.prototype.toString is only defined once, for the generic Error and all its "subclasses" use that implementation, I've made that version spec compliant.


boa/src/builtins/error/mod.rs
```diff
@@ -65,14 +65,11 @@ impl Error {
...
-        Ok(Value::from(format!(
-            "{}: {}",
-            name.display(),
-            message.display()
-        )))
+    pub(crate) fn to_string(this: &Value, _: &[Value], ctx: &mut Context) -> Result<Value> {
+        let name = this.get_field("name").to_string(ctx)?;
+        let message = this.get_field("message").to_string(ctx)?;
+
+        Ok(Value::from(format!("{}: {}", name, message)))
```
HalidOdat, line 72: Here there should also be a check if it undefined we set message to be `"Error"` as described in spec <https://tc39.es/ecma262/#sec-error.prototype.tostring>

boa/src/builtins/error/syntax.rs
```diff
@@ -52,11 +52,11 @@ impl SyntaxError {
...
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString
     #[allow(clippy::wrong_self_convention)]
-    pub(crate) fn to_string(this: &Value, _: &[Value], _: &mut Context) -> Result<Value> {
-        let name = this.get_field("name");
-        let message = this.get_field("message");
-        // FIXME: This should not use `.display()`
-        Ok(format!("{}: {}", name.display(), message.display()).into())
+    pub(crate) fn to_string(this: &Value, _: &[Value], ctx: &mut Context) -> Result<Value> {
+        let name = this.get_field("name").to_string(ctx)?;
+        let message = this.get_field("message").to_string(ctx)?;
```
HalidOdat, line 57: same here

boa/src/builtins/error/mod.rs
RageKnify: Than I should change other Errors too, right? Like TypeError and ReferenceError.

HalidOdat: > Than I should change other Errors too, right? Like TypeError and ReferenceError.

Yes. pretty much

RageKnify: Having all 5 error types have the exact same `to_string` function feels really bad, any idea on how to "fix" this?

HalidOdat: > Having all 5 error types have the exact same `to_string` function feels really bad, any idea on how to "fix" this?

This is not spec compliant and there should only be one implementation in `Error.prototype.toString` and all the other errors should **inherit** `Error`s prototype. But I wanted to leave this for when we merge #722 since without it's a bit hackish.

This should be very easy in #722 which introduces two things `ConstructorBuilder` (a builder for creating constructor objects like `Object`, `Array`, `Error`, etc) and cached core standard objects like `Error` and it's `prototype`. The `ConstructorBuilder` has a builder method called `inherit` what that's a prototype that will be put in the new objects prototype as `$OBJECT.prototype.__proto__`

I would suggest waiting for #722.

RageKnify: Sounds good, #722 is pretty big, I'll make reviewing it my priority.

HalidOdat:
This should now be unblocked since #722 has been merged :)

RageKnify:
@HalidOdat on Discord we discussed using `?` when invoking `construct_type_error` since it will now return result, but in situations where I know it won't Error, becuase I'm giving `&str` for the message, should I use `expect` instead? Using `?` might generate unnecessary code.

There are some places where I'm forced to use `expect` because I'm inside closures which don't return `Result`.

HalidOdat:
> @HalidOdat on Discord we discussed using `?` when invoking `construct_type_error` since it will now return result, but in situations where I know it won't Error, becuase I'm giving `&str` for the message, should I use `expect` instead? Using `?` might generate unnecessary code.
> 
> There are some places where I'm forced to use `expect` because I'm inside closures which don't return `Result`.

If we know that it `construct_type_error` will not give use a `Err`, then we can return `Value` and call `.expect` in the `construct_type_error` function.

RageKnify:
Ok, that's my opinion too, there's no need for `?` if we know it's safe to use `expect`, I'll make sure all my changes use `expect` where possible.

Razican:
> Ok, that's my opinion too, there's no need for `?` if we know it's safe to use `expect`, I'll make sure all my changes use `expect` where possible.

The idea here is that if at any point in our code, we should never ever reach a branch, it's good to panic. Wether that branch is an `Err()`, a `None` or their opposites (or other things, of course).

A panic should mean "we messed up, this is a bug in Boa 100% for sure", and no JavaScript code should ever cause a panic (we still have to solve this part).

But in this case, I think `expect()` is the way to go in many places.

boa/src/context.rs
```diff
@@ -283,7 +283,7 @@ impl Context {
     }
 
     /// Constructs a `RangeError` with the specified message.
-    pub fn construct_range_error<M>(&mut self, message: M) -> Value
+    pub fn construct_range_error<M>(&mut self, message: M) -> Result<Value>
```
HalidOdat, line 286: I think it would be better if we returned Value (like we did before) and and in the body we will call `.expect()` instead of `.expect_err()` since the construction should not fail it makes sense to return `Value`, instead of `Result<Value>`

RageKnify: I can see that `construct_range_error` is only ever called with valid messages, is that always the case for all `construct_$TYPE_error` functions?

RageKnify: Now that I think about it, the `construct_$TYPE_error` messages could never fail because their message will be converted to `String` so it will always be a valid message. I'll fix it ASAP.

HalidOdat: > I can see that `construct_range_error` is only ever called with valid messages, is that always the case for all `construct_$TYPE_error` functions?

Yes. they have pretty much the same constructor, so this should not fail

boa/src/builtins/error/mod.rs
```diff
@@ -93,8 +96,33 @@ impl Error {
...
+            message_to_string.as_str()
+        };
+
+        if name == "" {
+            Ok(Value::from(message))
+        } else if message == "" {
+            Ok(Value::from(name))
+        } else {
+            Ok(Value::from(format!("{}: {}", name, message)))
+        }
```
HalidOdat, line 126:
```suggestion
        if name.is_empty() {
            Ok(message.into())
        } else if message.is_empty() {
            Ok(name.into())
        } else {
            Ok(format!("{}: {}", name, message).into())
        }
```
better to check with `.is_empty()` than to compare ` == ""`, also it looks cleaner with `.into()`, instead of `from`

HalidOdat (approved):

Lan2u (approved):

PR #770, opened 2020-10-02
Support for symbols as property keys for `Object.defineProperty`

### LINKED ISSUE #732: Symbols cannot be defined as a property
```js
var obj = {};
var s = Symbol('1');
Object.defineProperty(obj, s);
```
Outputs: `Uncaught "TypeError": "can't convert symbol to string"`

The problem starts here:
https://github.com/boa-dev/boa/blob/master/boa/src/builtins/object/mod.rs#L101 we try to coarce the key into a `string` type. We shouldn't need to do this as Property's already support Symbols as keys. We should pass the argument straight through to PropertyKey::from() and let it sort it out.

PropertyKey is defined here:
https://github.com/boa-dev/boa/blob/master/boa/src/property/mod.rs#L290-L294

georgeroman:
Fixes #732


boa/src/property/mod.rs
```diff
@@ -293,6 +293,17 @@ pub enum PropertyKey {
...
+impl From<&Value> for PropertyKey {
+    #[inline]
+    fn from(value: &Value) -> PropertyKey {
+        match value {
+            Value::String(ref string) => string.clone().into(),
+            Value::Symbol(ref symbol) => symbol.clone().into(),
+            _ => panic!("Cannot be used as property"),
+        }
+    }
+}
```
HalidOdat, line 305: We don't need this we have a spec compliant (ToPropertyKey) `Value::to_property_key` method that handles all the invariants of `Value`.

georgeroman:
Addressed the comments.

boa/src/builtins/object/mod.rs
```diff
@@ -98,7 +98,11 @@ impl Object {
     /// Define a property in an object
     pub fn define_property(_: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         let obj = args.get(0).expect("Cannot get object");
-        let prop = args.get(1).expect("Cannot get object").to_string(ctx)?;
+        let prop = args
+            .get(1)
+            .expect("Cannot get object")
+            .to_property_key(ctx)
+            .expect("Cannot be used as property key");
```
HalidOdat, line 105:
```suggestion
            .to_property_key(ctx)?;
```
We should use the `?` try operator to propagate the error if one is thrown, instead of panicking.

HalidOdat (approved):

PR #774, opened 2020-10-03
Implement `regress`

### LINKED ISSUE #478: Use ECMAScript syntax for regular expressions

neeldug:
This Pull Request fixes/closes #478.

It changes the following:

- Implements `regress` crate in place of the `regex` crate


HalidOdat (approved):

Lan2u (approved):

Razican:
It seems that this reduces our conformance for something like 19.01% to 18.99%. some of the issues I find are that things like this one fail:
https://github.com/tc39/test262/blob/main/test/language/literals/regexp/S7.8.5_A2.4_T1.js

It seems that regress is not 100% spec compliant. We should check issues like this one and report them upstream. I will also add some comments, an invalid regex should never panic.

boa/src/builtins/regexp/mod.rs
```diff
@@ -137,34 +136,26 @@ impl RegExp {
...
         }
-        // the `regex` crate uses '(?{flags})` inside the pattern to enable flags
-        if !pattern.is_empty() {
-            pattern = format!("(?{})", pattern);
-        }
-        pattern.push_str(regex_body.as_str());
 
-        let matcher = Regex::new(pattern.as_str()).expect("failed to create matcher");
+        let matcher = Regex::newf(regex_body.as_str(), Flags::from(sorted_flags.as_str()))
+            .expect("failed to create matcher");
```
Razican, line 158: I don't think a panic here is spec-compliant. We should never panic if we get an invalid regular expression. I know this was happening before, but It's best if we can avoid it, now that we are working on this.

Could we implement the constructor in compliance with the spec?
https://tc39.es/ecma262/#sec-regexp-constructor

I don't think we should block the PR on this, but if it's merged as-is, we should create an issue about this.

RageKnify: I'd say to open an issue mentioning lack of spec compliability for `RegExp`s, there's a bit of work to be done in both `boa`'s side and `regress`'s.

Razican: Let's do that :)

Razican (approved):

PR #775, opened 2020-10-03
Change ClassBuilder methods to return &mut self

### LINKED ISSUE #760: `ClassBuilder` methods should return `&mut Self`
Following the builder pattern `ClassBuilder` should return `&mut Self` so it should be possible to chain the calls

JohnDoneth:
This Pull Request closes #760.

It changes the following:
Allows `ClassBuilder` methods to be chained to follow the builder pattern.


HalidOdat:
Hey @JohnDoneth. I'm kind of confused why this PR exist? already #722 fixed this issue

PR #777, opened 2020-10-03
Fix cyclic JSON.stringify / primitive conversion stack overflows

### LINKED ISSUE #545: JSON.stringify should throw TypeError("cyclic object value") when a circular reference is found.
**Describe the bug**
JSON.stringify should throw a TypeError when it encounters a circular reference. Currently, the stack overflows.

**To Reproduce**
```javascript
let a = [];
a[0] = a;
JSON.stringify(a);
```

**Expected behavior**
A TypeError should be thrown. See NOTE 1 in the [spec](https://tc39.es/ecma262/#sec-json.stringify)

 - OS: Ubuntu Linux
 - Version: 20.04
 - Rustc version: rustc 1.43.1 (8d69840ab 2020-05-04)

**Additional context**
You can find more information on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Exceptions).

vgel:
This Pull Request fixes/closes #545

It changes the following:

- Refactor gcobject::RecursionLimiter so it can be used outside of its module for liveness tracking
- Use that struct in Value::to_json to throw a TypeError, per the spec
- Also use it in Context::ordinary_to_primitive, where we return a default value
- Adds tests for above

The biggest change is probably `ordinary_to_primitive`. Since the spec doesn't mention what to do, I went and looked at what v8 and SpiderMonkey do. They return a default value based on the type hint: either `0.` or `""`. (See example here: https://repl.it/repls/IvoryCircularCertification#index.js) So, I did the same thing. We could alternatively throw a TypeError, but that would both diverge from the spec (which doesn't mention anything) and existing implementations.


boa/src/context.rs
```diff
@@ -380,10 +392,26 @@ impl Context {
...
+        // TODO: #776 this should also accept Type::Function
         debug_assert!(o.get_type() == Type::Object);
         // 2. Assert: Type(hint) is String and its value is either "string" or "number".
         debug_assert!(hint == PreferredType::String || hint == PreferredType::Number);
 
+        // Diverge from the spec here to make sure we aren't going to overflow the stack by converting
+        // a recursive structure
+        // We can follow v8 & SpiderMonkey's lead and return a default value for the hint in this situation
+        // (see https://repl.it/repls/IvoryCircularCertification#index.js)
+        let obj = o.as_gc_object().unwrap(); // UNWRAP: Asserted type above
```
RageKnify, line 404: Mostly stylistic, but I think using `.expect("Asserted type above")` is preferable to `.unwrap()` with a comment.

RageKnify:
One thing I think we could also change is the signature of `OrdinaryToPrimitive`, it is only called in 2 places on the spec, and in both cases it receives values whose types has been asserted to be `Object`, so we could for example receive `GcObject`, but I leave that decision to @HalidOdat who knows more about how we represent Objects.

boa/src/context.rs
```diff
@@ -380,10 +392,26 @@ impl Context {
...
+        let recursion_limiter = RecursionLimiter::new(&obj);
+        if recursion_limiter.live {
+            // we're in a recursive object, bail
+            return Ok(match hint {
+                PreferredType::Number => Value::number(0.),
+                PreferredType::String => Value::string(""),
```
HalidOdat, line 410:
```suggestion
                PreferredType::Number => Value::from(0),
                PreferredType::String => Value::from(""),
```

HalidOdat (commented):
This looks ready to merge, just some small changes :)

HalidOdat:
> One thing I think we could also change is the signature of `OrdinaryToPrimitive`, it is only called in 2 places on the spec, and in both cases it receives values whose types has been asserted to be `Object`, so we could for example receive `GcObject`, but I leave that decision to @HalidOdat who knows more about how we represent Objects.

Yes. I would say its better if we move it to GcObject as `pub(crate)`, the `Context` it is not a good place for it.

boa/src/value/mod.rs
```diff
@@ -268,6 +243,37 @@ impl Value {
         }
     }
 
+    /// Converts an object to JSON, checking for reference cycles and throwing a TypeError if one is found
+    fn object_to_json(&self, obj: &GcObject, interpreter: &mut Context) -> Result<JSONValue> {
```
RageKnify, line 247: I think maybe this method should instead be an associated function of the `JSON` object, the `obj` is enough to get it's fields, there's no need to have the original `Value`. What do you think @HalidOdat ?

HalidOdat: > I think maybe this method should instead be an associated function of the `JSON` struct, the `obj` is enough to get it's fields, there's no need to have the original `Value`. What do you think @HalidOdat Its?

It's better if we move this to `GcObject`, so we can take a &self

vgel: Done!

boa/src/context.rs
vgel: Done!

boa/src/context.rs
vgel: Unwrap isn't necessary anymore because the signature now takes a GcObject directly.

RageKnify:
Looks like you accidentaly broke a lot of tests @vgel. Not sure what you did, but you broke a lot of stuf.

vgel:
I'm not sure -- the tests seemed to run fine until I merged master. Will look into it.

boa/src/object/gcobject.rs
```diff
@@ -267,6 +269,111 @@ impl GcObject {
...
+
+        // 5. For each name in methodNames in List order, do
+        let this = Value::from(self.clone());
+        for name in &method_names {
+            // a. Let method be ? Get(O, name).
+            let method: Value = this.get_field(*name);
+            // b. If IsCallable(method) is true, then
+            if method.is_function() {
+                // i. Let result be ? Call(method, O).
+                let result = self.call(&this, &[], interpreter)?;
```
HalidOdat, line 333:
```suggestion
                let result = self.call(&method, &[], interpreter)?;
```
I think @vgel this might be the problem we should be calling the method not this

vgel: Yep! should have checked here, had to figure that out myself with `git bisect` :sweat_smile:

HalidOdat (approved):
Looks good to me!

RageKnify (approved):

PR #782, opened 2020-10-03
Throw TypeError if regexp is passed to startsWith, endsWith, includes

### LINKED ISSUE #755: Throw TypeError if regex is passed to String.prototype.startsWith
In our implementation of `String.prototype.startsWith` we're not type checking the argument and blindly coercing it into a string.
https://github.com/boa-dev/boa/blob/master/boa/src/builtins/string/mod.rs#L301

We should match on the value (or if let) and if its a regular expression throw a type error:

```rust
ctx.throw_type_error("First argument to String.prototype.startsWith must not be a regular expression")?;
```

There should be a test also

Same here: https://github.com/boa-dev/boa/blob/master/boa/src/builtins/string/mod.rs#L344
and here: https://github.com/boa-dev/boa/blob/master/boa/src/builtins/string/mod.rs#L390

Example JS
```
const str1 = 'Saturday night plans';
console.log(str1.startsWith(/saturday/));
```

pt2121:
This Pull Request fixes/closes #755.

It changes the following:

- Updating `startsWith`, `endsWith`, `includes` to throw `TypeError` if regex is passed in.
- Adding tests.

### DIFF
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -331,11 +331,15 @@ impl String {
         // Then we convert it into a Rust String by wrapping it in from_value
         let primitive_val = this.to_string(ctx)?;
 
-        // TODO: Should throw TypeError if pattern is regular expression
-        let search_string = args
-            .get(0)
-            .expect("failed to get argument for String method")
-            .to_string(ctx)?;
+        let arg = args.get(0).cloned().unwrap_or_else(Value::undefined);
+
+        if Self::is_regexp_object(&arg) {
+            ctx.throw_type_error(
+                "First argument to String.prototype.startsWith must not be a regular expression",
+            )?;
+        }
+
+        let search_string = arg.to_string(ctx)?;
 
         let length = primitive_val.chars().count() as i32;
         let search_length = search_string.chars().count() as i32;
@@ -374,11 +378,15 @@ impl String {
         // Then we convert it into a Rust String by wrapping it in from_value
         let primitive_val = this.to_string(ctx)?;
 
-        // TODO: Should throw TypeError if search_string is regular expression
-        let search_string = args
-            .get(0)
-            .expect("failed to get argument for String method")
-            .to_string(ctx)?;
+        let arg = args.get(0).cloned().unwrap_or_else(Value::undefined);
+
+        if Self::is_regexp_object(&arg) {
+            ctx.throw_type_error(
+                "First argument to String.prototype.endsWith must not be a regular expression",
+            )?;
+        }
+
+        let search_string = arg.to_string(ctx)?;
 
         let length = primitive_val.chars().count() as i32;
         let search_length = search_string.chars().count() as i32;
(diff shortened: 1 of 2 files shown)


boa/src/builtins/string/mod.rs
```diff
@@ -462,6 +481,13 @@ impl String {
         }
     }
 
+    fn is_regex_string(value: &Value) -> bool {
```
HalidOdat, line 484:
```suggestion
    fn is_regexp_object(value: &Value) -> bool {
```
I think `is_regexp_object` is a better name, since a regex is an object, not a string

boa/src/builtins/string/mod.rs
```diff
@@ -331,11 +331,17 @@ impl String {
         // Then we convert it into a Rust String by wrapping it in from_value
         let primitive_val = this.to_string(ctx)?;
 
-        // TODO: Should throw TypeError if pattern is regular expression
-        let search_string = args
+        let arg = args
             .get(0)
-            .expect("failed to get argument for String method")
-            .to_string(ctx)?;
+            .expect("failed to get argument for String method");
```
HalidOdat, line 336: its better if we move this past the regexp check with the string conversion:

```
		if Self::is_regex_string(arg) {
            ctx.throw_type_error(
                "First argument to String.prototype.startsWith must not be a regular expression",
            )?;
        }

        let search_string = args.get(0).cloned().unwrap_or_else(Value::undefined).to_string(ctx)?;
```
also we shouldn't panic if the argument was not provided but we should create undefined in its place

boa/src/builtins/string/mod.rs
pt2121: Sounds good. will update

boa/src/builtins/string/mod.rs
pt2121: How about
```
        let arg = args
            .get(0)
            .cloned()
            .unwrap_or_else(Value::undefined);

        if Self::is_regexp_object(&arg) {
            ctx.throw_type_error(
                "First argument to String.prototype.startsWith must not be a regular expression",
            )?;
        }

        let search_string = arg.to_string(ctx)?;
```

`is_regexp_object` takes `&Value` so we'd need to `get(0)` before calling it.

HalidOdat (approved):

boa/src/builtins/string/mod.rs
```diff
@@ -374,11 +378,16 @@ impl String {
...
 
+        let arg = args.get(0).cloned().unwrap_or_else(Value::undefined);
+
+        if Self::is_regexp_object(&arg) {
+            ctx.throw_type_error(
+                "First argument to String.prototype.endsWith must not be a regular expression",
+            )?;
+        }
+
         // TODO: Should throw TypeError if search_string is regular expression
```
RageKnify, line 389: You forgot to remove this `TODO` comment.

RageKnify (approved):

PR #785, opened 2020-10-04
Fix zero argument panic in JSON.parse()

### LINKED ISSUE #784: Panic when no arguments are given to JSON.parse
**Describe the bug**
A panic occurs when no arguments are passed to `JSON.parse`.

**To Reproduce**
```js
JSON.parse()
```

**Expected behavior**
Expected: a JavaScript error is raised (Chrome seems to raise a `SyntaxError`)
image

- OS: Windows 10
- Version: V1909, Build 18363.1082
- Target triple: `x86_64-pc-windows-msvc` (also happens with the WASM example on the website)
- Rustc version: `rustc 1.46.0 (04488afe3 2020-08-24)`

**Additional context**
It's caused by an `.expect` in the `builtins/json/mod.rs` file.

More info on `JSON.parse`: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse)

JohnDoneth:
This Pull Request closes #784.

It changes the following:
Instead of panicking when an argument is not provided to `JSON.parse()`, a syntax error is raised like other JavaScript implementations.

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -61,12 +61,12 @@ impl Json {
     /// [spec]: https://tc39.es/ecma262/#sec-json.parse
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
     pub(crate) fn parse(_: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
-        match serde_json::from_str::<JSONValue>(
-            &args
-                .get(0)
-                .expect("cannot get argument for JSON.parse")
-                .to_string(ctx)?,
-        ) {
+        let arg = match args.get(0) {
+            Some(arg) => arg.to_string(ctx)?,
+            None => return ctx.throw_syntax_error("JSON.parse(): Expected argument"),
+        };
+
+        match serde_json::from_str::<JSONValue>(&arg) {
             Ok(json) => {
                 let j = Value::from_json(json, ctx);
                 match args.get(1) {
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -79,7 +79,7 @@ impl Json {
                     _ => Ok(j),
                 }
             }
-            Err(err) => Err(Value::from(err.to_string())),
+            Err(err) => ctx.throw_syntax_error(err.to_string()),
         }
     }
 
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -61,10 +61,11 @@ impl Json {
     /// [spec]: https://tc39.es/ecma262/#sec-json.parse
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
     pub(crate) fn parse(_: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
-        let arg = match args.get(0) {
-            Some(arg) => arg.to_string(ctx)?,
-            None => return ctx.throw_syntax_error("JSON.parse(): Expected argument"),
-        };
+        let arg = args
+            .get(0)
+            .cloned()
+            .unwrap_or_else(Value::undefined)
+            .to_string(ctx)?;
 
         match serde_json::from_str::<JSONValue>(&arg) {
             Ok(json) => {
--- a/boa/src/builtins/json/tests.rs
+++ b/boa/src/builtins/json/tests.rs
@@ -315,3 +315,10 @@ fn json_fields_should_be_enumerable() {
     assert_eq!(actual_object, expected);
     assert_eq!(actual_array_index, expected);
 }
+
+#[test]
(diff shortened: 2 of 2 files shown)


croraf:
@JohnDoneth Two things:

1. As far as I see, the spec doesn't explicitly say what should happen when argument is not passed. I guess somewhere before it says that it is considered as `undefined`. Would like to find where is that said.

2. From what I see about how Chrome behaves, is that it doesn't have special handling of this case. It is just that no parameter is considered as "undefined" (what I mention in bullet 1) and then the JSON parser throws internally during parsing which gets propagated through the stack.

boa/src/builtins/json/mod.rs
```diff
@@ -61,12 +61,12 @@ impl Json {
...
-        match serde_json::from_str::<JSONValue>(
-            &args
-                .get(0)
-                .expect("cannot get argument for JSON.parse")
-                .to_string(ctx)?,
-        ) {
+        let arg = match args.get(0) {
+            Some(arg) => arg.to_string(ctx)?,
+            None => return ctx.throw_syntax_error("JSON.parse(): Expected argument"),
+        };
```
HalidOdat, line 67:
```suggestion
        let arg = match args.get(0).cloned().unwrap_or_else(Value::undefined).to_string()?;
```
As @croraf said the spec does not say that a `TypeError` should be thrown when not argument is supplied.

JohnDoneth: Thanks for the review. I updated the code to match the suggestion.

JohnDoneth:
New output with changes made:
```
>> JSON.parse()
Uncaught: "expected value at line 1 column 1"
```

boa/src/builtins/json/mod.rs
croraf: @HalidOdat But he did throw the `SyntaxError`, so his first version was basically fine, and according to spec. But your solution is perhaps slightly more elegant.

HalidOdat: > @HalidOdat But he did throw the `SyntaxError`, so his first version was basically fine, and according to spec. But your solution is perhaps slightly more elegant.

Sorry, My bad, thought it was a `TypeError`

HalidOdat:
> New output with changes made:
> 
> ```
> >> JSON.parse()
> Uncaught: "expected value at line 1 column 1"
> ```

This matches more with what other engines are doing :)

HalidOdat (approved):
Looks good to me! :)

croraf:
@JohnDoneth Thanks. Perhaps not related to the task (but a general error), but in Firefox it shows:

> Uncaught SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data

and in Chrome

> Uncaught SyntaxError: Unexpected token u in JSON at position 0

It would be nice to have the error class and message in the output.
@HalidOdat Is this a common boa issue?

RageKnify:
I don't know if all errors in JSON::parse should be syntax errors (I'd appreciate if someone else can look into it), but for the Value that gets thrown to be an actual Error with a message this line: https://github.com/JohnDoneth/boa/blob/cfe64ae60065ca36e9bc45a7e7d46a29282abbfb/boa/src/builtins/json/mod.rs#L82
Needs to be changed to instead invoke `ctx.throw_syntax_error(err.to_string)`.

HalidOdat:
> I don't know if all errors in JSON::parse should be syntax errors (I'd appreciate if someone else can look into it), but for the Value that gets thrown to be an actual Error with a message this line: [JohnDoneth/boa@`cfe64ae`/boa/src/builtins/json/mod.rs#L82](https://github.com/JohnDoneth/boa/blob/cfe64ae60065ca36e9bc45a7e7d46a29282abbfb/boa/src/builtins/json/mod.rs#L82)
> Needs to be changed to instead invoke `ctx.throw_syntax_error(err.to_string)`.

Good catch! @RageKnify, Yep. https://tc39.es/ecma262/#sec-json.parse step 2. `SyntaxError` should be thrown and that's why the name of the error, because an error is not being thrown, we are throwing a string.

JohnDoneth:
There we go!
```
>> JSON.parse()
Uncaught: "SyntaxError": "expected value at line 1 column 1"
```
Thanks for spotting that @RageKnify

croraf:
Also we should have a test for that :D

croraf:
Although don't know what is the policy for that now when we will have Test-262 test suite, because this test will be a duplicate of https://github.com/tc39/test262/blob/main/test/built-ins/JSON/parse/text-non-string-primitive.js

HalidOdat:
> Although don't know what is the policy for the tests, now that we have Test-262 test suite, because this test will be a duplicate of

Yes some tests are duplicates but that is fine the rust tests catch the bigger bugs but the test262 catch all bugs and is not practical to run the test262 tests for when you want to see if a small case passes since takes like 10m for them to finish, compared to `cargo test` around 1 sec for the tests to finish.

Lan2u (approved):

RageKnify (approved):

PR #794, opened 2020-10-05
Refactor `PropertyDescriptor`

HalidOdat:
This PR adds a nice API for creating Property Descriptor (`DataDescriptor` and `AccessorDescriptor`) by making illegal state unrepresentable, for example `get` and `value` in the same property descriptor (which is not allowed, but was possible with the old implementation), which we needed dynamic checks to prove that it was not the case (performance penalty), but now we leverage the Type System for this as well as it makes implementing #591 and accessors so much easier.

It changes the following:
 - Rename `Property` => `PropertyDescriptor`
 - Add `DataDescriptor` for defining data descriptors
 - Add `AccessorDescriptor` for defining accessor descriptors
 - Make `PropertyDescriptor` an enum
 - Remove `HAS_*` form `Attribute`
 - Remove `Value::update_property` related to #577

### DIFF
--- a/boa/src/property/mod.rs
+++ b/boa/src/property/mod.rs
@@ -24,133 +24,188 @@ use std::{convert::TryFrom, fmt};
 mod attribute;
 pub use attribute::Attribute;
 
+/// A data descriptor is a property that has a value, which may or may not be writable.
+///
+/// More information:
+/// - [MDN documentation][mdn]
+/// - [ECMAScript reference][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-property-descriptor-specification-type
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty
 #[derive(Debug, Clone, Trace, Finalize)]
 pub struct DataDescriptor {
     value: Value,
-    attribute: Attribute,
+    attributes: Attribute,
 }
 
 impl DataDescriptor {
+    /// Create a new `DataDescriptor`.
     #[inline]
-    pub fn new<V>(value: V, attribute: Attribute) -> Self
+    pub fn new<V>(value: V, attributes: Attribute) -> Self
     where
         V: Into<Value>,
     {
         Self {
             value: value.into(),
-            attribute,
+            attributes,
         }
     }
 
+    /// Return the `value` of the data descriptor.
     #[inline]
     pub fn value(&self) -> Value {
         self.value.clone()
     }
 
+    /// Return the attributes of the descriptor.
     #[inline]
     pub fn attributes(&self) -> Attribute {
-        self.attribute
+        self.attributes
     }
 
+    /// Check whether the descriptor is configurable.
     #[inline]
     pub fn configurable(&self) -> bool {
-        self.attribute.configurable()
+        self.attributes.configurable()
     }
 
+    /// Set whether the descriptor is configurable.
     #[inline]
     pub fn set_configurable(&mut self, configurable: bool) {
-        self.attribute.set_configurable(configurable)
+        self.attributes.set_configurable(configurable)
     }
 
-    /// Set enumerable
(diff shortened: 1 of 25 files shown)


HalidOdat:
The is a boost from 10% to 20% in performance

boa/src/value/mod.rs
```diff
@@ -465,11 +466,14 @@ impl Value {
...
+    where
+        P: Into<PropertyDescriptor>,
+    {
         let _timer = BoaProfiler::global().start_event("Value::update_property", "value");
 
         if let Some(ref mut object) = self.as_object_mut() {
-            object.insert(field, new_property);
+            object.insert(field, new_property.into());
         }
     }
```
RageKnify, line 478: Since we're touching this code might as well change the type of `field` to be `K: Into<PropertyKey>`.

boa/src/property/mod.rs
```diff
@@ -14,172 +14,220 @@
...
+    /// Return the getter if it exists.
     #[inline]
-    pub fn get(mut self, get: Value) -> Self {
-        self.get = Some(get);
-        self
+    pub fn getter(&self) -> Option<&GcObject> {
+        self.get.as_ref()
     }
 
+    /// Return the getter if it exists.
```
RageKnify, line 150: `Return the getter` -> `Return the setter`

boa/src/property/mod.rs
```diff
@@ -199,82 +256,42 @@ impl Property {
...
-
-        let enumerable = value.get_field("enumerable");
-        if !enumerable.is_undefined() {
-            attribute.set_enumerable(bool::from(&enumerable));
-        }
 
-        let configurable = value.get_field("configurable");
-        if !configurable.is_undefined() {
-            attribute.set_configurable(bool::from(&configurable));
+    /// Check whether the descriptor is enumerable.
```
RageKnify, line 280: `is enumerable.` -> `is configurable.`

RageKnify:
You're leaving a lot of `TODO`s and `FIXME`s related to `AccessorDescriptor`s, to ease solving them I think you should create an issue and mention the issue number in those comments. That way when implementing we know where to search.

boa/src/value/mod.rs
HalidOdat: > Since we're touching this code might as well change the type of `field` to be `K: Into<PropertyKey>`.

I think its best if I remove this function its used in a very small amount of paces, and we should be slowly removing the Value object behaviour #577

HalidOdat:
> You're leaving a lot of `TODO`s and `FIXME`s related to `AccessorDescriptor`s, to ease solving them.

My intention with this PR was not to fix the already present bugs (even though I did fix some), but to make them more visible, places where there were hard to identify bugs (like accessor not being called, properties being ignored when they shouldn't) now panic (with unwrap/expect) so its easier to know were there located.

> I think you should create an issue and mention the issue number in those comments. That way when implementing we know what to search for.

Will create the issue once this has been merged :)

boa/src/value/mod.rs
```diff
@@ -183,15 +183,16 @@ impl Value {
...
                         idx.to_string(),
-                        Property::data_descriptor(
+                        DataDescriptor::new(
                             Self::from_json(json, context),
                             Attribute::WRITABLE | Attribute::ENUMERABLE | Attribute::CONFIGURABLE,
                         ),
                     );
                 }
                 new_obj.set_property(
                     "length".to_string(),
```
RageKnify, line 193: Can you remove the `.to_string()`? I know you didn't change this line, but its an unnecessary allocation.

Razican (approved):
I love the new benchmarks. Check the comment by @RageKnify and see if a re-base fixes the tests, but for me it's good to go (also, I'm not super familiar with this, so take my review with a gran of salt) :)

RageKnify (approved):
Other than the allocation everything seems alright, I'd already reviwed everything yesterday and the last 2 `commit`s didn't change all that much.

PR #800, opened 2020-10-05
Feature `console` crate feature

### LINKED ISSUE #718: Split WebAPI (environment) built-in objects from the ES builtin objects.
Split WebAPI (environment) objects from the ES builtin objects.

Perhaps delegate the implementation of environment objects to some other library.

HalidOdat:
This Pull Request fixes/closes #718 .

It changes the following:
- Put `console` object behind a feature flag 
- Documented features

### DIFF
--- a/boa/src/lib.rs
+++ b/boa/src/lib.rs
@@ -2,7 +2,7 @@
 This is an experimental Javascript lexer, parser and compiler written in Rust. Currently, it has support for some of the language.
 
 # Crate Features
- - **serde** - Enables serialization and deserelization of the AST (Abstracr Syntax Tree).
+ - **serde** - Enables serialization and deserialization of the AST (Abstract Syntax Tree).
  - **console** - Enables `boa`s WHATWG `console` object implementation.
  - **profiler** - Enables profiling with measureme (this is mostly internal).
 
--- a/boa/src/lib.rs
+++ b/boa/src/lib.rs
@@ -1,4 +1,12 @@
-//! This is an experimental Javascript lexer, parser and compiler written in Rust. Currently, it has support for some of the language.
+/*!
+This is an experimental Javascript lexer, parser and compiler written in Rust. Currently, it has support for some of the language.
+
+# Crate Features
+ - **serde** - Enables serialization and deserelization of the AST (Abstracr Syntax Tree).
+ - **console** - Enables `boa`s WHATWG `console` object implementation.
+ - **profiler** - Enables profiling with measureme (this is mostly internal).
+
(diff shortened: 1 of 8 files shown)


boa/src/lib.rs
```diff
@@ -1,4 +1,12 @@
-//! This is an experimental Javascript lexer, parser and compiler written in Rust. Currently, it has support for some of the language.
+/*!
+This is an experimental Javascript lexer, parser and compiler written in Rust. Currently, it has support for some of the language.
+
+# Crate Features
+ - **serde** - Enables serialization and deserelization of the AST (Abstracr Syntax Tree).
```
RageKnify, line 5: Typos: 'deserelization' -> 'deserialization', 'Abstracr' -> 'Abstract'

RageKnify (approved):
Just the 2 typos, the rest looks fine.

Lan2u (approved):

croraf:
I don't think this is the correct way to go. Console should be injected into the context (during context creation or dynamically) from `boa-cli` or any other user of `boa`. It should not be a part of `boa` so the feature flag should apply on the `boa-cli`.

HalidOdat:
> I don't think this is the correct way to go. Console should be injected into the context (during context creation or dynamically) from `boa-cli` or any other user of `boa`. It should not be a part of `boa` so the feature flag should apply on the `boa-cli`.

The `console` feature is off by default, if the user  whats a quick default implementation of the console spec that is that prints as string they can enable it, but if the user whats a more advanced console object like firefox/chrome console object where it is interactive and you can click things (not just a string) they can define their own implementation.

Razican:
> > I don't think this is the correct way to go. Console should be injected into the context (during context creation or dynamically) from `boa-cli` or any other user of `boa`. It should not be a part of `boa` so the feature flag should apply on the `boa-cli`.
> 
> The `console` feature is off by default, if the user whats a quick default implementation of the console spec that prints as string they can enable it, but if the user whats a more advanced console object like firefox/chrome console object where it is interactive and you can click things (not just a string) they can define their own implementation.

Should this be a separate crate?

HalidOdat:
> Should this be a separate crate?

> @HalidOdat If the user wants a quick default implementation of console which he doesnt have, he should use the console crate we aslo provide and initialize the boa instance providing this crate. This is negligible more work comparing to enabling a feature.
> 
> And `boa-cli` will do that by default. (so the feature can be on the `boa-cli`, enabled by default)

Yes. This is probably the best choic, Im reopening the issue :)

PR #804, opened 2020-10-05
Feature `EvalError`

HalidOdat:
It changes the following:
- Add `EvalError`
- Add tests

### DIFF
--- a/boa/src/builtins/error/eval.rs
+++ b/boa/src/builtins/error/eval.rs
@@ -1,0 +1,69 @@
+//! This module implements the global `EvalError` object.
+//!
+//! Indicates an error regarding the global `eval()` function.
+//! This exception is not thrown by JavaScript anymore, however
+//! the `EvalError` object remains for compatibility.
+//!
+//! More information:
+//!  - [MDN documentation][mdn]
+//!  - [ECMAScript reference][spec]
+//!
+//! [spec]: https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-evalerror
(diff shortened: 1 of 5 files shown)


RageKnify (approved):

RageKnify:
You can extend the test in `boa/src/builtins/error/tests.rs` to include `EvalError`.

boa/src/builtins/error/eval.rs
```diff
@@ -0,0 +1,69 @@
...
+    property::Attribute,
+    Context, Result, Value,
+};
+
+/// JavaScript `EvalError` impleentation.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct EvalError;
+
+impl BuiltIn for EvalError {
+    const NAME: &'static str = "EvalError";
```
Razican, line 27: We don't need to specify lifetimes in consts, right? This would maybe be clearer:
```suggestion
    const NAME: &str = "EvalError";
```

boa/src/builtins/error/eval.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    fn attribute() -> Attribute {
```
Razican, line 29: Should/could we have these as const functions?

HalidOdat: Making It const means that the the insides must be const and the bitwise OR `|` operator is not on bitflag structs

boa/src/builtins/error/eval.rs
HalidOdat: This gives a missing lifetime parameter error.

Razican (approved):

PR #805, opened 2020-10-05
Feature `Function.prototype.call`

### LINKED ISSUE #797: Implement Function.prototype.call
**ECMASCript feature**
I noticed recently that Function.prototype.call exists and is not implemented.
[ECMAScript specification][spec].

[spec]: https://tc39.es/ecma262/#sec-function.prototype.call

**Example code**
```javascript
let e = new Error();
Object.prototype.toString.call(e)
```
The expected output is `"[object Error]"`

RageKnify:
This Pull Request fixes/closes #797.
It changes the following:

-Implement Function.prototype.call
-Add a test to ensure it works

### DIFF
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -312,6 +312,25 @@ impl BuiltInFunctionObject {
     fn prototype(_: &Value, _: &[Value], _: &mut Context) -> Result<Value> {
         Ok(Value::undefined())
     }
+
+    /// `Function.prototype.call`
+    ///
+    /// The call() method invokes self with the first argument as the `this` value.
+    ///
+    /// More information:
+    ///  - [MDN documentation][mdn]
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-function.prototype.call
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call
+    fn call(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+        let func = this;
+        if !func.is_function() {
+            return context.throw_type_error("TODO: this is not a function");
+        }
+        let this_arg: Value = args.get(0).map_or_else(Value::default, Value::clone);
+        context.call(func, &this_arg, &args[1..])
+    }
 }
 
 impl BuiltIn for BuiltInFunctionObject {
@@ -339,6 +358,7 @@ impl BuiltIn for BuiltInFunctionObject {
         )
         .name(Self::NAME)
         .length(Self::LENGTH)
+        .method(Self::call, "call", 1)
         .build();
 
         (Self::NAME, function_object.into(), Self::attribute())
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -324,12 +324,13 @@ impl BuiltInFunctionObject {
     /// [spec]: https://tc39.es/ecma262/#sec-function.prototype.call
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call
     fn call(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
-        let func = this;
-        if !func.is_function() {
-            return context.throw_type_error("TODO: this is not a function");
+        if !this.is_function() {
+            return context.throw_type_error(format!("{} is not a function", this.display()));
         }
-        let this_arg: Value = args.get(0).map_or_else(Value::default, Value::clone);
-        context.call(func, &this_arg, &args[1..])
+        let this_arg: Value = args.get(0).cloned().unwrap_or_default();
+        // TODO?: 3. Perform PrepareForTailCall
+        let start = if !args.is_empty() { 1 } else { 0 };
+        context.call(this, &this_arg, &args[start..])
     }
 }
 
--- a/boa/src/builtins/function/tests.rs
+++ b/boa/src/builtins/function/tests.rs
@@ -125,4 +125,13 @@ fn function_prototype_call() {
     let value = forward_val(&mut engine, func).unwrap();
     assert!(value.is_string());
     assert_eq!(value.as_string().unwrap(), "[object Error]");
+
+    let throw = r#"
+        let call = Function.prototype.call;
+        call(call)
+        "#;
+    let value = forward_val(&mut engine, throw).unwrap_err();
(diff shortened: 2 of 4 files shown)


RageKnify:
Forgot to end the error message and write a test for it, I'm doing it right now.

boa/src/builtins/function/mod.rs
```diff
@@ -312,6 +312,25 @@ impl BuiltInFunctionObject {
...
+    ///
+    /// More information:
+    ///  - [MDN documentation][mdn]
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-function.prototype.call
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call
+    fn call(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+        let func = this;
+        if !func.is_function() {
```
HalidOdat, line 328:
```suggestion
        if !this.is_function() {
```

boa/src/builtins/function/mod.rs
```diff
@@ -312,6 +312,25 @@ impl BuiltInFunctionObject {
...
+            return context.throw_type_error("TODO: this is not a function");
```
HalidOdat, line 329: \Why the `TODO`?

boa/src/builtins/function/mod.rs
```diff
@@ -312,6 +312,25 @@ impl BuiltInFunctionObject {
...
+        }
+        let this_arg: Value = args.get(0).map_or_else(Value::default, Value::clone);
```
HalidOdat, line 331:
```suggestion
        let this_arg = args.get(0).cloned().unwrap_or_default();
```
This does the same thing but its much clearer IMO

boa/src/builtins/function/mod.rs
```diff
@@ -312,6 +312,25 @@ impl BuiltInFunctionObject {
...
+        context.call(func, &this_arg, &args[1..])
```
HalidOdat, line 332: What if the first argument is not present, this will panic

RageKnify: Does `let start = if !args.is_empty() { 1 } else { 0 };` look ok for you? Then I do `&args[start..]`.

HalidOdat: > Does `let start = if !args.is_empty() { 1 } else { 0 };` look ok for you? Then I do `&args[start..]`.

Yes, that would work. :)

HalidOdat:
Maybe we should test it with more that one argumet

RageKnify:
> Maybe we should test it with more that one argumet

Reasonable, any idea for a function to test?

HalidOdat:
> Reasonable, any idea for a function to test?

We can test any function that takes 1 or more arumunts ( not including `this` ), how about `Number.prototype.toString( [radix] )` it takes an optional base we can make the base 16 and assert that the number is in hexadecimal notation.

RageKnify:
I came up with:
```js
function f(a, b){
this.a = a;
this.b = b;
}
let o = {a: 0, b: 0}
f.call(o, 1, 2)
```
Then we check `o.a == 1` and `o.b == 2`, what do you think @HalidOdat ?

HalidOdat:
> Then we check `o.a == 1` and `o.b == 2`, what do you think @HalidOdat ?

Yes. This looks good :)

HalidOdat (approved):
Looks perfect to me!

Razican (approved):
Looks good to me too :D

Razican:
Somehow it seems we didn't see that this was failing the test262 suite. I think we should revert it until we find the issue. What do you think @HalidOdat @RageKnify ?

boa/src/builtins/function/mod.rs
```diff
@@ -312,6 +312,26 @@ impl BuiltInFunctionObject {
...
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-function.prototype.call
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call
+    fn call(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+        if !this.is_function() {
+            return context.throw_type_error(format!("{} is not a function", this.display()));
+        }
+        let this_arg: Value = args.get(0).cloned().unwrap_or_default();
+        // TODO?: 3. Perform PrepareForTailCall
```
HalidOdat, line 331: I think it maybe the missing `// TODO?: 3. Perform PrepareForTailCall`.

HalidOdat:
> I think we should revert it until we find the issue. What do you think @HalidOdat @RageKnify ?

Yes. I think so too.

RageKnify:
I'd like to identify which test of test262 is causing the failure. But I'm assuming its a long recursion that requires TailCall optimization, not sure how `boa` would handle that...

Razican:
> I'd like to identify which test of test262 is causing the failure. But I'm assuming its a long recursion that requires TailCall optimization, not sure how `boa` would handle that...

Something that can be done to see the last test to be executed is to uncomment [this line](https://github.com/boa-dev/boa/blob/master/boa_tester/src/exec.rs#L88) locally and run it like this.

The error message was this:
```
memory allocation of 51539607528 bytes failed/home/runner/work/_temp/445e69cc-9b3d-4bcd-887f-44df8e8c1c9b.sh: line 2:  3393 Aborted                 (core dumped) cargo run --release --bin boa_tester -- -o ../gh-pages/test262
```

RageKnify:
I've identified the test, `test/built-ins/Array/prototype/map/15.4.4.19-3-28.js`:
```js
// Copyright (c) 2012 Ecma International.  All rights reserved.
// This code is governed by the BSD license found in the LICENSE file.

/*---
esid: sec-array.prototype.map
description: Array.prototype.map - value of 'length' is boundary value (2^32)
---*/

function callbackfn(val, idx, obj) {
  return val > 10;
}

var obj = {
  0: 12,
  length: 4294967296
};
assert.throws(RangeError, function() {
  var newArr = Array.prototype.map.call(obj, callbackfn);
});
```
I don't think the problem is recursion, it is `Array.prototype.map` not being spec compliant, it's trying to create a huge `Vec` and dies.
https://tc39.es/ecma262/#sec-array.prototype.map
Point 6.c (`HasProperty(O, Pk)`) isn't respected in:
https://github.com/boa-dev/boa/blob/eb1fd280d59a0284c02c3ed9ceba606ea1e68c25/boa/src/builtins/array/mod.rs#L598-L607

PR #806, opened 2020-10-05
Feature `URIError`

HalidOdat:
It changes the following:
- Add `URIError`
- Add tests

### DIFF
--- a/boa/src/builtins/error/uri.rs
+++ b/boa/src/builtins/error/uri.rs
@@ -1,0 +1,68 @@
+//! This module implements the global `URIError` object.
+//!
+//! The `URIError` object represents an error when a global URI handling
+//! function was used in a wrong way.
+//!
+//! More information:
+//!  - [MDN documentation][mdn]
+//!  - [ECMAScript reference][spec]
+//!
+//! [spec]: https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-urierror
+//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/URIError
+
+use crate::{
+    builtins::BuiltIn,
+    object::{ConstructorBuilder, ObjectData},
+    profiler::BoaProfiler,
+    property::Attribute,
+    Context, Result, Value,
+};
+
+/// JavaScript `URIError` impleentation.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct UriError;
+
+impl BuiltIn for UriError {
+    const NAME: &'static str = "URIError";
+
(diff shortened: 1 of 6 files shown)


RageKnify (approved):

boa/src/builtins/error/uri.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Create a new error object.
+    pub(crate) fn constructor(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
+        if let Some(message) = args.get(0) {
+            this.set_field("message", message.to_string(ctx)?);
+        }
+
+        // This value is used by console.log and other routines to match Object type
+        // to its Javascript Identifier (global constructor method name)
+        this.set_data(ObjectData::Error);
```
Lan2u, line 65: Isn't this missing the fileName + lineNumber parameters? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/URIError/URIError

HalidOdat: That is a non-standard feature: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/lineNumber its not in the spec https://tc39.es/ecma262/#sec-constructor-properties-of-the-global-object-urierror

here is what mdn says:
> Non-standard
> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

boa/src/builtins/error/uri.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    fn attribute() -> Attribute {
+        Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE
+    }
+
+    fn init(context: &mut Context) -> (&'static str, Value, Attribute) {
+        let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
+
+        let error_prototype = context.standard_objects().error_object().prototype();
+        let attribute = Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE;
```
Razican, line 36: Why don't we use `Self::attribute()` here?

Razican (approved):

boa/src/builtins/error/uri.rs
HalidOdat: Its because I did not want to mix the binding attributes of the global object `URIError` (`Self::attribute()`) and the binding attributes of the `name` and `message` property, this makes it clear that they are separate.

Issue #793, opened 2020-10-05
Spread operator on a non-iterable panics


dapper-gh:
**Describe the bug**
A panic occurs when running the spread operator on a non-iterable.
![image](https://user-images.githubusercontent.com/31050943/95032422-7567fc00-0688-11eb-9f29-59389020342b.png)
The panic occurs at line `42` in file `syntax/ast/node/array/mod.rs`.

**To Reproduce**
```js
[...5]
```

**Expected behavior**
A JavaScript error is raised. Chrome seems to raise a `TypeError`, as seen below.
![image](https://user-images.githubusercontent.com/31050943/95032192-a98eed00-0687-11eb-8d01-e87a0a1c580a.png)
Additionally, the [spec](https://tc39.es/ecma262/#sec-getiterator) states that a `TypeError` should be raised.

**Build environment (please complete the following information):**

- OS: `Windows 10`
- Version: `Version 1909 (OS Build 18363.1082)`
- Target triple: `x86_64-pc-windows-msvc`
- Rustc version: `rustc 1.46.0 (04488afe3 2020-08-24)`

**Additional context**
MDN: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)
Spec: [https://tc39.es/ecma262/#sec-array-initializer](https://tc39.es/ecma262/#sec-array-initializer)

croraf:
I'll try this. I'll maybe need help on discord though.

Issue #810, opened 2020-10-07
Map.prototype[@@iterator]() method not implemented


croraf:
**ECMASCript feature**
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@iterator

**Example code**
```js
let first = new Map([["1", "one"], ["2", "two"]]);
first[Symbol.iterator]
```

**Context**

This is blocking proper fix of https://github.com/boa-dev/boa/issues/793
After this is done the `merge()` test should be unignored from boa/src/builtins/map/tests.rs

PR #811, opened 2020-10-07
[exec Array] implement spread operator using iterator

croraf:
This Pull Request fixes https://github.com/boa-dev/boa/issues/793

It changes the following:

- Makes spread operator to use iterator as per spec https://tc39.es/ecma262/#sec-runtime-semantics-arrayaccumulation
- Breaks `Map` spreading, because Map does not implement iterator. That's why I opened https://github.com/boa-dev/boa/issues/810

### DIFF
--- a/boa/src/builtins/map/tests.rs
+++ b/boa/src/builtins/map/tests.rs
@@ -43,7 +43,10 @@ fn clone() {
     assert_eq!(result, "2");
 }
 
+// depends on the https://github.com/boa-dev/boa/issues/810
 #[test]
+#[ignore]
+
 fn merge() {
     let mut engine = Context::new();
     let init = r#"
--- a/boa/src/builtins/map/tests.rs
+++ b/boa/src/builtins/map/tests.rs
@@ -43,10 +43,9 @@ fn clone() {
     assert_eq!(result, "2");
 }
 
-// depends on the https://github.com/boa-dev/boa/issues/810
+// TODO depends on the https://github.com/boa-dev/boa/issues/810
 #[test]
 #[ignore]
-
 fn merge() {
     let mut engine = Context::new();
     let init = r#"
--- a/boa/src/syntax/ast/node/array/mod.rs
+++ b/boa/src/syntax/ast/node/array/mod.rs
@@ -39,14 +39,27 @@ impl Executable for ArrayDecl {
         for elem in self.as_ref() {
             if let Node::Spread(ref x) = elem {
                 let val = x.run(interpreter)?;
-                let mut vals = interpreter.extract_array_properties(&val).unwrap();
-                elements.append(&mut vals);
-                continue; // Don't push array after spread
+                let iterator_record =
+                    super::super::super::super::builtins::iterable::get_iterator(interpreter, val)?;
+                //not sure what to do with this next_index mentioned in the spec
+                //it is mentioned that it has to be returned from somewhere
+                //https://tc39.es/ecma262/#sec-runtime-semantics-arrayaccumulation
+                //let mut next_index = 0;
+                loop {
+                    let next = iterator_record.next(interpreter)?;
+                    if next.is_done() {
+                        break;
+                    }
+                    let next_value = next.value();
+                    //next_index += 1;
+                    elements.push(next_value.clone());
+                }
+            } else {
+                elements.push(elem.run(interpreter)?);
             }
-            elements.push(elem.run(interpreter)?);
         }
-        Array::add_to_array_object(&array, &elements)?;
 
+        Array::add_to_array_object(&array, &elements)?;
         Ok(array)
     }
 }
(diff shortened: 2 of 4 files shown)


Razican:
Hey! thanks for this contribution. It seems it requires a re-base to fix one issue we were having in the test262 test suite :)

boa/src/builtins/map/tests.rs
```diff
@@ -43,7 +43,10 @@ fn clone() {
     assert_eq!(result, "2");
 }
 
+// depends on the https://github.com/boa-dev/boa/issues/810
 #[test]
+#[ignore]
+
```
RageKnify, line 49: Was this empty line an accident?

boa/src/builtins/map/tests.rs
```diff
@@ -43,7 +43,10 @@ fn clone() {
     assert_eq!(result, "2");
 }
 
+// depends on the https://github.com/boa-dev/boa/issues/810
```
Razican, line 46: I would add a TODO in this coment, so that is easier to find.

boa/src/syntax/ast/node/array/mod.rs
```diff
@@ -39,14 +39,27 @@ impl Executable for ArrayDecl {
...
                 let val = x.run(interpreter)?;
-                let mut vals = interpreter.extract_array_properties(&val).unwrap();
-                elements.append(&mut vals);
-                continue; // Don't push array after spread
+                let iterator_record =
+                    super::super::super::super::builtins::iterable::get_iterator(interpreter, val)?;
+                //not sure what to do with this next_index mentioned in the spec
+                //it is mentioned that it has to be returned from somewhere
+                //https://tc39.es/ecma262/#sec-runtime-semantics-arrayaccumulation
+                //let mut next_index = 0;
```
Razican, line 47: This is basically because we don't follow the spec with arrays. The thing is, arrays have a number of integer-indexed elements, and a `nextIndex`. this `nextIndex` will be the same as `array.length`, and it's defined here: https://tc39.es/ecma262/#sec-array-initializer-runtime-semantics-evaluation

But, due to our array being a vector (which causes performance issues), we don't have this defined. This is partially the reason behind ##816. This means that if we for example create a huge empty array (which should not allocate), say `new Array(50000)`, we will have huge performance issues.

This also means that if we do something like `let a = [5,6,,,6,7,,8];`, we will have an `undefined` value in the non-defined slots, but they should instead be empty slots, and here, `nextIndex` would be 8 (same as length).

Razican (approved):
Looks pretty good to me. Check if the mentioned changes make sense.

boa/src/syntax/ast/node/array/mod.rs
```diff
@@ -39,14 +39,27 @@ impl Executable for ArrayDecl {
...
+                loop {
+                    let next = iterator_record.next(interpreter)?;
+                    if next.is_done() {
+                        break;
+                    }
+                    let next_value = next.value();
+                    //next_index += 1;
```
RageKnify, line 54: The lines with `next_index` can be removed, right?

boa/src/builtins/map/tests.rs
croraf: Yes. I will remove it.

boa/src/syntax/ast/node/array/mod.rs
croraf: Check the previous comment from Razican https://github.com/boa-dev/boa/pull/811#discussion_r502460858

boa/src/syntax/ast/node/array/mod.rs
croraf: What do you suggest? Should we leave this comment and commented next_index, or modify the comment, or remove completly?

Razican: I would change the comment to a `TODO`, explaing this is waiting for a proper internal Array representation.

croraf: Please check the newly committed rewording.

RageKnify (approved):

boa/src/syntax/ast/node/array/mod.rs
```diff
@@ -39,14 +39,26 @@ impl Executable for ArrayDecl {
         for elem in self.as_ref() {
             if let Node::Spread(ref x) = elem {
                 let val = x.run(interpreter)?;
-                let mut vals = interpreter.extract_array_properties(&val).unwrap();
-                elements.append(&mut vals);
-                continue; // Don't push array after spread
+                let iterator_record =
+                    super::super::super::super::builtins::iterable::get_iterator(interpreter, val)?;
```
HalidOdat, line 43:
```suggestion
                let iterator_record = iterable::get_iterator(interpreter, val)?;
```
It's better if we import the iterable module on top.

HalidOdat (approved):

PR #822, opened 2020-10-08
build(gh-pages): switch bootstrap to cdn loading

### LINKED ISSUE #821: Improve our GitHub pages site by not requiring vendored code
We currently require Bootstrap and jQuery to be included in the GitHub pages branch. This is not ideal, and we should somehow just create a static page with a `node_modules` or with a webpacked script that includes all the requirements.

This should be available for both the Test262 test suite and for the benchmarks, that should then be linked from the demo and the main website.

neeldug:
deletes local copy of bootstrap and replaces it with bootstrap's
official cdn link

Closes 821

This Pull Request fixes/closes #821 .

It changes the following:

- deletes local copy of bootstrap
- formats `index.html` with prettier
- adds minor accessibility improvements, e.g. `lang` attribute, etc.


Razican:
I have merged the PR, so you can now re-base this :)

Issue #778, opened 2020-10-03
Panic on invalid utf16 instead of an Error


alexanderkjall:
**Describe the bug**

Expected an Error, not a panic.

**To Reproduce**
Can be reproduced with this program

```
fn main() {
    let _ = boa::parse("'43\\uDDDDDDDDD19");
}
```

**Expected behavior**

An Error.

**Build environment (please complete the following information):**

- OS: Ubuntu 20.04
- Version: 0.10.0
- Target triple: [e.g. x86_64-unknown-linux-gnu]
- Rustc version: 1.48.0-nightly (d006f5734 2020-08-28)

**Additional context**
 Full stacktrace:
```
thread 'main' panicked at 'Could not get next codepoint: DecodeUtf16Error { code: 56797 }', /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/syntax/lexer/string.rs:185:42
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d006f5734f49625c34d6fc33bf6b9967243abca8/library/std/src/panicking.rs:483
   1: core::panicking::panic_fmt
             at /rustc/d006f5734f49625c34d6fc33bf6b9967243abca8/library/core/src/panicking.rs:85
   2: core::option::expect_none_failed
             at /rustc/d006f5734f49625c34d6fc33bf6b9967243abca8/library/core/src/option.rs:1221
   3: core::result::Result<T,E>::expect
             at /home/capitol/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:933
   4: <boa::syntax::lexer::string::StringLiteral as boa::syntax::lexer::Tokenizer<R>>::lex
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/syntax/lexer/string.rs:182
   5: boa::syntax::lexer::Lexer<R>::next
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/syntax/lexer/mod.rs:191
   6: boa::syntax::parser::cursor::buffered_lexer::BufferedLexer<R>::fill
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/syntax/parser/cursor/buffered_lexer/mod.rs:116
[...]

PR #826, opened 2020-10-08
Fix tokenizing Unicode escape sequence in string literal

### LINKED ISSUE #808: A string containing only a Unicode escape sequence causes an error
**Describe the bug**
If a string is composed only of a Unicode escape sequence (e.g. `"\uABCD"`), then an error occurs instead of producing a string.

**To Reproduce**
```js
"\u00A7"
```
(that's the code for the section sign. It's not important, as the bug applies to any escape sequence)

**Expected behavior**
The string is valid, and is not supposed to throw an error. The syntax for string literals can be found [here](https://tc39.es/ecma262/#sec-literals-string-literals).

- OS: `Windows 10`
- Version: `Version 1909 (OS Build 18363.1082)`
- Rustc version: `rustc 1.46.0 (04488afe3 2020-08-24)`

**Additional context**
Spec: [https://tc39.es/ecma262/#sec-literals-string-literals](https://tc39.es/ecma262/#sec-literals-string-literals)
MDN: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar)

As far as I can tell, the bug is caused by an issue in [syntax/lexer/string.rs](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/lexer/string.rs) where once a Unicode escape sequence is detected, it skips checking for `"` and believes the string has finished without a `"`.

jevancc:
This Pull Request fixes/closes #808 .

It changes the following:

- Add `InnerIter::peek_iter()` and store the peeked char in `InnerIter` instance instead of `Cursor`
- Fix invalid Unicode code point in lexer test

The issue is caused by a bug in `Cursor::fill_bytes`. When `Cursor::fill_bytes` was called after `Cursor::peek`, the iter would be incremented to peek the next char but the peeked char would not be filled to the buffer. This PR introduces a new method `InnerIter::peek_char` and stores the peeked char in `InnerIter` so that it can fill the peeked char to input buffer when invoking `InnerIter::fill_bytes` correctly.

The test `syntax::lexer::tests::codepoint_with_no_braces` is updated in this PR. Since `\uD83D` is a surrogate code point, the test will panic when calling `decode_utf16` and trying to decode it. This bug should be fixed in another issue/PR to handle the invalid code point.

### DIFF
--- a/boa/src/syntax/lexer/cursor.rs
+++ b/boa/src/syntax/lexer/cursor.rs
@@ -7,7 +7,6 @@ use std::io::{self, Bytes, Error, ErrorKind, Read};
 #[derive(Debug)]
 pub(super) struct Cursor<R> {
     iter: InnerIter<R>,
-    peeked: Option<Option<char>>,
     pos: Position,
     strict_mode: bool,
 }
@@ -53,7 +52,6 @@ where
     pub(super) fn new(inner: R) -> Self {
         Self {
             iter: InnerIter::new(inner.bytes()),
-            peeked: None,
             pos: Position::new(1, 1),
             strict_mode: false,
         }
@@ -64,14 +62,7 @@ where
     pub(super) fn peek(&mut self) -> Result<Option<char>, Error> {
         let _timer = BoaProfiler::global().start_event("cursor::peek()", "Lexing");
 
-        let iter = &mut self.iter;
-        if let Some(v) = self.peeked {
-            Ok(v)
-        } else {
-            let val = iter.next_char()?;
-            self.peeked = Some(val);
-            Ok(val)
-        }
+        self.iter.peek_char()
     }
 
     /// Compares the character passed in to the next character, if they match true is returned and the buffer is incremented
@@ -81,7 +72,7 @@ where
 
         Ok(match self.peek()? {
             Some(next) if next == peek => {
-                let _ = self.peeked.take();
+                let _ = self.iter.next_char();
                 true
             }
             _ => false,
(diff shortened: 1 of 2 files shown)


boa/src/syntax/lexer/cursor.rs
```diff
@@ -283,11 +292,13 @@ where
...
-        let next_byte = self.iter.next().transpose()?;
-
-        match next_byte {
-            Some(next) if next <= 0x7F => Ok(Some(next)),
-            None => Ok(None),
+        match self.next_char() {
+            Ok(Some(chr)) if chr.is_ascii() => {
+                let mut buf = [0u8; 4];
+                chr.encode_utf8(&mut buf);
+                Ok(Some(buf[0]))
```
HalidOdat, line 299:
```suggestion
                Ok(Some(chr as u8))
```
Couldn't we just cast to u8 since we check if its ascii, this should be faster than calling `encode_utf8`

HalidOdat (commented):
Looks good to me!

boa/src/syntax/lexer/cursor.rs
jevancc: Sounds great! I've made this change.

jevancc: Besides, I can look into the invalid code point issue. Is there already an issue/PR for it?

HalidOdat (approved):
Looks good to me! :)

boa/src/syntax/lexer/cursor.rs
HalidOdat: > Besides, I can look into the invalid code point issue. Is there already an issue/PR for it?

There is an issue https://github.com/boa-dev/boa/issues/778, but not a PR to fix

Razican (approved):
It seems there are no regressions in the benchmarks, so this looks pretty good! Thanks!

PR #835, opened 2020-10-09
Move `Object` internal object methods to `GcObject`

HalidOdat:
**Why move the methods?**
Moving Object internal methods to `GcObject` because we borrow the object and we hold on to it and if an accessor (getters/setters) was to access the object it will give a borrow panic.

I expect some performance regression, since we do more borrow checks.

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -154,6 +154,7 @@ impl Json {
                 .map(|obj| {
                     let object_to_return = Value::new_object(None);
                     for (key, val) in obj
+                        .borrow()
                         .iter()
                         // FIXME: handle accessor descriptors
                         .map(|(k, v)| (k, v.as_data_descriptor().unwrap().value()))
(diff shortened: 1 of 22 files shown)


boa/src/builtins/json/mod.rs
```diff
@@ -148,12 +148,13 @@ impl Json {
         let replacer_as_object = replacer
             .as_object()
             .expect("JSON.stringify replacer was an object");
-        if replacer_as_object.is_callable() {
+        if replacer_as_object.borrow().is_callable() {
```
RageKnify, line 151: `replacer_as_object` is actually a `GcObject`, right?

HalidOdat:
Looking at the benchmarks the isn't that big of a regression, as I expected.

Razican:
> Looking at the benchmarks there isn't that big of a regression, as I expected.

I'm guessing many times the compiler can statically remove some checks.

boa/src/object/gcobject.rs
```diff
@@ -26,17 +26,19 @@ use std::{
 };
 
 /// A wrapper type for an immutably borrowed `Object`.
-pub type Ref<'object> = GcCellRef<'object, Object>;
+pub type Ref<'object, T> = GcCellRef<'object, T>;
 
 /// A wrapper type for a mutably borrowed `Object`.
-pub type RefMut<'object> = GcCellRefMut<'object, Object>;
+pub type RefMut<'object, T> = GcCellRefMut<'object, T>;
```
RageKnify, line 32: You changed `Ref` and `RefMut` to be generic but the comments and the lifetime name still mention `Object` and `'object`.

RageKnify (approved):
Other than updating the comments and the lifetime naming it looks fine.

PR #838, opened 2020-10-10
style(boa): minor cleanup

neeldug:
It changes the following:

- extracts return from if let statement
- fixes clippy error in nightly

### DIFF
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -528,7 +528,7 @@ impl Number {
                             } else {
                                 // String can't be parsed.
                                 Ok(Value::from(f64::NAN))
-                            }
+                            };
                         } else {
                             radix = 10
                         };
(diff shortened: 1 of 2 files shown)


HalidOdat (approved):

boa/src/builtins/number/mod.rs
```diff
@@ -354,7 +354,7 @@ impl Number {
                 fraction -= digit as f64;
                 // Round to even.
                 if fraction + delta > 1.0
-                    && (fraction > 0.5 || (fraction - 0.5) < f64::EPSILON && digit & 1 != 0)
+                    && (fraction > 0.5 || (fraction - 0.5).abs() < f64::EPSILON && digit & 1 != 0)
```
Lan2u, line 357: For my understanding why is this .abs()? Won't this change the behaviour of the code?

neeldug: Yeah it will but if we’re comparing to epsilon typically you do want Abs rather than the flat value, I couldn’t figure out why it wasn’t Abs in the first place.

Lan2u (approved):

Razican (approved):

PR #840, opened 2020-10-10
fix(profiler): update profiler to match current measureme api

### LINKED ISSUE #839: Profiler no longer works with measureme 9.0.0
**Describe the bug**
The profiler no longer functions as there was a breaking change in `measureme` from 0.8.0 to 9.0.0

**To Reproduce**
```
cargo run --features Boa/profiler <some.js>
```

**Expected behavior**
The profiler to produce a profile

neeldug:
moves all OS profiling to use SerializationSink as the previous types
have been deprecated

Closes #839

### DIFF
--- a/boa/src/profiler.rs
+++ b/boa/src/profiler.rs
@@ -11,17 +11,11 @@ use std::{
     thread::{current, ThreadId},
 };
 
-/// MmapSerializatioSink is faster on macOS and Linux
-/// but FileSerializationSink is faster on Windows
-#[cfg(not(windows))]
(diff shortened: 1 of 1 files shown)


HalidOdat (approved):
Looks good to me!

Lan2u (approved):

Razican:
Should we add a check in the CI to make sure this works?

neeldug:
Yeah sure, I guess we just need to ensure the profiler is able to execute with no errors on some JS file.

Issue #846, opened 2020-10-11
Map.prototype.entries() method not implemented


croraf:
**ECMASCript feature**
https://www.ecma-international.org/ecma-262/11.0/index.html#sec-map.prototype.entries
-->

**Example code**
```
const map1 = new Map();

map1.set('0', 'foo');
map1.set(1, 'bar');

const iterator1 = map1.entries();

console.log(iterator1.next().value);
// expected output: ["0", "foo"]

console.log(iterator1.next().value);
// expected output: [1, "bar"]
```

croraf:
I'm trying some implementation. I basically duplicated the ArrayIterator functionality, and tried to fit in the missing types.

PR #847, opened 2020-10-11
[builtin Map] Map.prototype.entries method and map iterator

croraf:
This Pull Request fixes/closes https://github.com/boa-dev/boa/issues/846

I will likely need some help.


boa/src/builtins/iterable/mod.rs
```diff
@@ -23,9 +25,12 @@ impl IteratorPrototypes {
             array_iterator: ArrayIterator::create_prototype(ctx, iterator_prototype.clone())
                 .as_gc_object()
                 .expect("Array Iterator Prototype is not an object"),
-            string_iterator: StringIterator::create_prototype(ctx, iterator_prototype)
+            string_iterator: StringIterator::create_prototype(ctx, iterator_prototype.clone())
                 .as_gc_object()
                 .expect("String Iterator Prototype is not an object"),
+            map_iterator: MapIterator::create_prototype(ctx, iterator_prototype.clone())
```
Razican, line 31: This should remove the clippy warning:
```suggestion
            map_iterator: MapIterator::create_prototype(ctx, iterator_prototype)
```

Razican (commented):
Seems it requires running `rustfmt`. In any case, I'm not an expert on this, maybe @HalidOdat or @RageKnify can help better :)

boa/src/builtins/iterable/mod.rs
croraf: OK, I see now where that is. Yes it was removed at first because it is not used later so clone is not necessary, but I returned it for consistency. I will remove it.

boa/src/builtins/map/mod.rs
```diff
@@ -29,6 +32,7 @@ impl BuiltIn for Map {
             .name(Self::NAME)
             .length(Self::LENGTH)
             .method(Self::set, "set", 2)
+            .method(Self::entries, "entries", 0)
```
HalidOdat, line 35: We can also implement the `keys` and `values` methods

boa/src/builtins/map/ordered_map.rs
```diff
@@ -94,6 +94,12 @@ where
         self.0.get(key)
     }
 
+    /// TODO description
+    /// TODO return Record
```
HalidOdat, line 98: Maybe some documentation

HalidOdat (commented):
Looks good! :) Also maybe some tests would be nice

boa/src/builtins/map/mod.rs
croraf: In a separate PR with their own tests.

boa/src/builtins/map/ordered_map.rs
```diff
@@ -94,6 +94,16 @@ where
         self.0.get(key)
     }
 
+    /// Get a key-value pair by index
+    /// Valid indices are 0 <= index < self.len()
+    /// Computes in O(1) time.
+    ///
+    /// TODO perhaps return Record as per
+    /// https://tc39.es/ecma262/#sec-%mapiteratorprototype%.next bullet 12.a.
```
HalidOdat, line 102:
```suggestion
    /// Get a key-value pair by index
    ///
    /// Valid indices are `0 <= index < self.len()`
    /// Computes in O(1) time.
```
We do return a record of `(key, value)` so there is no need for the todo

croraf: Is this Option<(&K, &V)> a Record?

croraf:
I've added new commit which should close #810

boa/src/builtins/map/ordered_map.rs
HalidOdat: > The Record type is used to describe data aggregations within the algorithms of this specification. A Record type value consists of one or more named fields. The value of each field is either an ECMAScript value or an abstract value represented by a name associated with the Record type. Field names are always enclosed in double brackets, for example [[Value]].

https://tc39.es/ecma262/#sec-list-and-record-specification-type

Yes. We return an tuple (key, value) and this is more idiomatic rust (since it matches with the standards HashMap). In any case whether we return a struct or a tuple this should not matter since this is just implementation detail.

boa/src/builtins/map/map_iterator.rs
```diff
@@ -0,0 +1,156 @@
...
+            ctx.throw_type_error("`this` is not an MapIterator")
+        }
+    }
+
+    /// Create the %MapIteratorPrototype% object
+    ///
+    /// More information:
+    ///  - [ECMA reference][spec]
+    ///
+    /// [spec]: TODO https://tc39.es/ecma262/#sec-%arrayiteratorprototype%-object
```
Razican, line 138: It seems that this link has a `TODO` comment, could you give it a look?

boa/src/builtins/map/ordered_map.rs
Razican, line 102: Should the `TODO` be removed?

Razican (commented):
I saw a couple of `TODO` comments, give them a look, and from my side, it's ready to merge.

HalidOdat (approved):

boa/src/builtins/map/ordered_map.rs
croraf: Removed 3 TODOs in total

RageKnify (approved):

Issue #167, opened 2019-10-17
Generate Bytecode


jasonwilliams:
Issue here to keep track of.
At some point generating bytecode would be needed.

Example from Rapidus:
https://github.com/maekawatoshiki/rapidus/blob/new-gen/src/bytecode_gen.rs#L474-L558

Spidermonkey:
https://dxr.mozilla.org/mozilla-central/source/js/src/vm/Opcodes.h

V8:
https://github.com/v8/v8/blob/master/src/compiler/opcodes.h

IovoslavIovchev:
Has any work been done on this?

jasonwilliams:
@IovoslavIovchev Nope not yet

Razican:
This will require a fundamental change on how the interpreter works, but it would be very useful to cache bytecode in the future, and to implement an optimization pass (or multiple optimization passes). Some overview on how V8 does this can be found here: https://v8.dev/blog/ignition-interpreter

jasonwilliams:
My guess is a new interpreter would be created along side the current one which reads the AST and generates bytecode, there would most likely be a switch for a while ‘—bytecode’

Design doc for ignition
https://docs.google.com/document/d/11T2CRex9hXxoJwbYqVQ32yIPMh0uouUZLdyrtmMoL44/mobilebasic#h.z0c1oogsbnnq

jasonwilliams:
Do we want to come up with a set of instructions that would work for us, that would get the ball rolling

IovoslavIovchev:
@jasonwilliams are there any ideas for the VM design/any work done? Are we aiming for a register-based VM? I reckon we need to have some idea what the VM is going to be like before we come up with the instruction set.

jasonwilliams:
> Are we aiming for a register-based VM?
Yes i would imagine it being similar to the V8 VM with registers and an accumulator

jasonwilliams:
Will attempt to base of Jason’s example here:
https://github.com/jorendorff/boa/commit/266319fbee93ef0e2467a384e237ecd5b0ac9c51#diff-8a4e02ddeeda9cccf194eab4892b1c1cR162-R167

There is an issue here though.
Operations on Values have been promoted to the Value level, so the stack needs to either use Values (which isn’t efficient because it will point to the heap for everything) or each instruction will need to box to a Value do the operation then come back down.

For now we will need to do the former, but in future we may have to refactor Values again, so that they’re not so coupled with the Gc

[...]

PR #854, opened 2020-10-12
Start on ByteCode interpreter

jasonwilliams:
This does run!

Nodes implement `CodeGen` which generates instructions onto a stack held in Context.
The `VM` will interpret the instructions from Context.

There are some issues:
- only basic instructions are added, but im working off https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Internals/Bytecode for now it should be easy to add more in.
- The Stack is a Vec, this isn't ideal (we may be able to live with it for now) but the stack should really be a fixed sized array. This isn't possible because Value can't be copied in there as it holds Rc and Gc values. Can we have fixed-sized Values that hold a pointer?
- I needed to copy the Instruction vec, i couldn't find a way to pass in both `ctx` and `ctx.instructions` together without it moaning about "mutable borrowing" twice.
- I need to change boa_cli back to the normal run.

As most of this is behind a feature flag it would be good to merge parts of it so we can work on as the main interpreter is active. Then one day we can switch over. 

for 

```js
2 + 2
```

The instructions generated are
```
[boa/src/context.rs:689] &self.instruction_stack = [
    Int32       2,
    Int32       2,
    Add,
]
4
```

We can optimize even further here and just do the addition at CodeGen level and throw 4 onto the stack

Fixes https://github.com/boa-dev/boa/issues/167
@Razican @HalidOdat

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -12,14 +12,15 @@
         "kind": "build",
         "isDefault": true
       },
-      "options": {
-        "env": {
-          "RUST_BACKTRACE": "full"
-        }
-      },
       "presentation": {
         "clear": true
-      }
+      },
+      "options": {
+        "env": {
+          "RUST_BACKTRACE": "1"
+        }
+      },
+      "problemMatcher": []
     },
     {
       "type": "process",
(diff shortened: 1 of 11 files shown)


jasonwilliams:
continuing here https://github.com/boa-dev/boa/pull/860

PR #859, opened 2020-10-13
Implement spec compliant Array constructor

### LINKED ISSUE #816: The `Array()` constructor is not spec compliant
**Describe the bug**
The `Array()` constructor is not spec compliant. When creating a huge array, for example:
```javascript
let a = new Array(50000);
```

This will try to actually insert 50k properties in the array. The [spec](https://tc39.es/ecma262/#sec-array-constructor-array) says otherwise. It should only set the length, and should not create any other property.

This happens because [our implementation](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/array/mod.rs#L100-L140) is not spec compliant. We should have 3 private constructor methods (`construct_array`, `construct_array_length` and `construct_array_items`), and they should be called from the `constructor` method. In each of them, we should have information about the exact spec, and use the actual spec implementation.

This means we will need to create an `array_create` function, that will follow [this](https://tc39.es/ecma262/#sec-arraycreate) spec.

A similar thing should be done with the rest of the `Array` prototype methods.

georgeroman:
Fixes #816


boa/src/builtins/array/mod.rs
```diff
@@ -98,8 +99,93 @@ impl Array {
...
+    /// [spec]: https://tc39.es/ecma262/#sec-arraycreate
+    fn array_create(
+        this: &Value,
+        length: u32,
+        prototype: Option<GcObject>,
+        context: &mut Context,
+    ) -> Result<Value> {
+        let prototype = match prototype {
+            Some(prototype) => prototype,
+            None => context.standard_objects().array_object().prototype(),
```
georgeroman, line 187: I'm not really sure about the correctness of this line, some help would be appreciated here. It's supposed to be step 4 from https://tc39.es/ecma262/#sec-arraycreate.

Razican: The idea here is that if `prototype` is `None`, we should set it to `%Array.prototype%`. I'm not sure if this is the best way to retrieve it, maybe @HalidOdat can confirm.

HalidOdat: > I'm not sure if this is the best way to retrieve it, maybe @HalidOdat can confirm.

Yes. this is the best way to get hold of the `Array.prototype` object

Razican (approved):

boa/src/builtins/array/mod.rs
```diff
@@ -98,8 +99,93 @@ impl Array {
...
+        this: &Value,
+        length: &Value,
+        context: &mut Context,
+    ) -> Result<Value> {
+        let prototype = context.standard_objects().array_object().prototype();
+        let array = Array::array_create(this, 0, Some(prototype), context)?;
+
+        if !length.is_number() {
+            array.set_field("0".to_string(), Value::from(length));
+            array.set_field("length".to_string(), Value::from(1));
```
RageKnify, line 138: I know the spec says to convert to `String`s, but this will just do unnecessary allocations, `0` for line 137 and `"length"` for line 138 should work fine.

boa/src/builtins/array/mod.rs
```diff
@@ -98,8 +99,93 @@ impl Array {
...
+        } else {
+            if length.is_double() {
+                return context.throw_range_error("Invalid array length");
+            }
+            array.set_field(
+                "length".to_string(),
```
RageKnify, line 144: Same thing, no need to have `.to_string()`.

boa/src/builtins/array/mod.rs
```diff
@@ -98,8 +99,93 @@ impl Array {
...
+    fn construct_array_values(
+        this: &Value,
+        items: &[Value],
+        context: &mut Context,
+    ) -> Result<Value> {
         let prototype = context.standard_objects().array_object().prototype();
+        let array = Array::array_create(this, items.len() as u32, Some(prototype), context)?;
+
+        for (k, item) in items.iter().enumerate() {
+            array.set_field(k.to_string(), item.clone());
```
RageKnify, line 167: Same thing, no need to have `.to_string()`.

boa/src/builtins/array/mod.rs
```diff
@@ -108,32 +194,10 @@ impl Array {
...
-            }
-        }
-
-        // finally create length property
         let length = DataDescriptor::new(
             length,
             Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::PERMANENT,
         );
-
         this.set_property("length".to_string(), length);
```
RageKnify, line 201: Same thing, no need to have `.to_string()`.

RageKnify (changes requested):
The spec asks for a lot of conversions to `String` which are unecessary and slow.

georgeroman:
Addressed all comments.

RageKnify (approved):

RageKnify:
@georgeroman sorry that you have to do this, but could you merge with current `master` so that all CI checks will pass please?

georgeroman:
Sure, no problem, just updated.

PR #860, opened 2020-10-13
Bytecode Interpreter (new branch)

jasonwilliams:
This does run!

Nodes implement `CodeGen` which generates instructions onto a stack held in Context.
The `VM` will interpret the instructions from Context.

There are some issues:
- only basic instructions are added, but im working off https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Internals/Bytecode for now it should be easy to add more in.
- The Stack is a Vec, this isn't ideal (we may be able to live with it for now) but the stack should really be a fixed sized array. This isn't possible because Value can't be copied in there as it holds Rc and Gc values. Can we have fixed-sized Values that hold a pointer?
- I needed to copy the Instruction vec, i couldn't find a way to pass in both `ctx` and `ctx.instructions` together without it moaning about "mutable borrowing" twice.
- I need to change boa_cli back to the normal run.

As most of this is behind a feature flag it would be good to merge parts of it so we can work on as the main interpreter is active. Then one day we can switch over. 

for 

```js
2 + 2
```

The instructions generated are
```
[boa/src/context.rs:689] &self.instruction_stack = [
    Int32       2,
    Int32       2,
    Add,
]
4
```

We can optimize even further here and just do the addition at CodeGen level and throw 4 onto the stack

Fixes https://github.com/boa-dev/boa/issues/167
@Razican @HalidOdat

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -26,7 +26,7 @@
       "type": "process",
       "label": "Cargo Run (VM)",
       "command": "cargo",
-      "args": ["run", "--features", "vm", "../tests/js/test.js"],
+      "args": ["run", "--features", "vm Boa/profiler", "../tests/js/test.js"],
       "group": {
         "kind": "build",
         "isDefault": true
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -24,6 +24,26 @@
     },
     {
       "type": "process",
+      "label": "Cargo Run (VM)",
+      "command": "cargo",
+      "args": ["run", "--features", "vm", "../tests/js/test.js"],
+      "group": {
+        "kind": "build",
+        "isDefault": true
+      },
+      "presentation": {
+        "clear": true
+      },
+      "options": {
+        "cwd": "${workspaceFolder}/boa_cli",
+        "env": {
+          "RUST_BACKTRACE": "1"
+        }
+      },
+      "problemMatcher": []
+    },
+    {
+      "type": "process",
       "label": "Cargo Run (Profiler)",
       "command": "cargo",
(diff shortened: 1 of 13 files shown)


jasonwilliams:
@HalidOdat its weird that tests are going down the "CodeGen" path when they're run with the "vm" feature off :/

jasonwilliams:
@HalidOdat i was hoping to get some sort of trace output, like this
```
$ cargo run -- --trace
> function fibo(x) { if (x<2) return 1; return fibo(x-1)+fibo(x-2)}
00020m 00000 Return                    <empty>
undefined
> fibo(3)
00009m 00000 PushInt8 3                <empty>
00015m 00002 GetValue 'fibo'           3.0
00066m 00007 Call 1                    Function
--> call function
  module_id:0 func_id:0
00007m 00000 GetValue 'x'              <empty>
00001m 00005 PushInt8 2                3.0
00013m 00007 Lt                        2.0

...

00001m 00007 Lt                        2.0
00000m 00008 JmpIfFalse 00016          true
00000m 00013 PushInt8 1                <empty>
00167m 00015 Return                    1.0
<-- return value(1.0)
  module_id:0 func_id:2
00001m 00052 Add                       1.0
00044m 00053 Return                    3.0
<-- return value(3.0)
  module_id:0 func_id:0
00000m 00012 Return                    3.0
3
> 
   |     |     |                        | 
   |     |     |                        \- value at the top of exec stack
   |     |     \-------------------------- instruction
   |     \-------------------------------- program counter
   \-------------------------------------- execution time per inst. (in microsecs)
```

So i've added Display for `Instruction`

RageKnify (approved):

PR #861, opened 2020-10-13
More bytecodes

HalidOdat:
It changes the following:
- Added `-`, `*`, `/`, `**`, `%`, `&`, `|`, `<<`, `>>`, `>>>`, `<`, `>`, `==`, `===`, `!=`, `!==`, `=>`, `<=`, `in`, `instanceof` operators
- Added unary operators `+`, `-`, `typeof`, `void`, `!`, `~`
- Added undefined, null, bool, string, bigint, rational literals
- When the `vm` flag is set we use the eval bytecode version and when its not we use the regular.

### DIFF
--- a/boa/src/vm/mod.rs
+++ b/boa/src/vm/mod.rs
@@ -1,4 +1,4 @@
-use crate::{Context, Value};
+use crate::{Context, Result, Value};
 
 pub(crate) mod compilation;
 pub(crate) mod instructions;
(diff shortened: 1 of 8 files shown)


boa/src/vm/mod.rs
```diff
@@ -37,8 +37,167 @@ impl<'a> VM<'a> {
 
                     self.stack.push(val);
                 }
+                Instruction::Sub => {
+                    let r = self.stack.pop().unwrap();
+                    let l = self.stack.pop().unwrap();
```
jasonwilliams, line 42: We have a lot of unwraps!
I wonder if we should make a self.pop method that unwraps then returns he value or panics with the correct message, then call that.

If we ever move to an array we then only need to change that method.

jasonwilliams (approved):

jasonwilliams:
@HalidOdat anything else you want to do on this or shall we merge?

HalidOdat:
> @HalidOdat anything else you want to do on this or shall we merge?

We can merge this and I can add more later on (in another PR, I don't have much free time this week).

Issue #883, opened 2020-10-16
panic at 'RangeError: Maximum BigInt size exceeded'


alexanderkjall:
**Describe the bug**
I expected something like:
`Uncaught RangeError: BigInt is too large to allocate`

Instead I got a panic

**To Reproduce**
Reproduce with:

```
fn main() {
    let mut context = boa::Context::new();
    let _ = context.eval("9722n>>-155555555555558n>>-1555");
}
```

**Expected behavior**
Firefox gives me:
`Uncaught RangeError: BigInt is too large to allocate`

**Build environment (please complete the following information):**

- OS: Ubuntu 20.04
- Version: 0.10.0
- Target triple: [e.g. x86_64-unknown-linux-gnu]
- Rustc version: rustc 1.49.0-nightly (dd7fc54eb 2020-10-15)

**Additional context**
backtrace:
```
thread 'main' panicked at 'RangeError: Maximum BigInt size exceeded', /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/builtins/bigint/operations.rs:75:9
stack backtrace:
   0: std::panicking::begin_panic
             at /home/capitol/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:505:12
   1: boa::builtins::bigint::operations::<impl core::ops::bit::Shr for boa::builtins::bigint::BigInt>::shr
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/builtins/bigint/operations.rs:75:9
   2: boa::value::operations::<impl boa::value::Value>::shr
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/value/operations.rs:329:30
   3: boa::exec::operator::<impl boa::exec::Executable for boa::syntax::ast::node::operator::BinOp>::run
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/exec/operator/mod.rs:74:35
   4: <boa::syntax::ast::node::Node as boa::exec::Executable>::run
             at /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.10.0/src/exec/mod.rs:107:36
   5: boa::exec::operator::<impl boa::exec::Executable for boa::syntax::ast::node::operator::BinOp>::run
[...]

PR #884, opened 2020-10-16
Fix some panics related to BigInt operations

georgeroman:
This should fix the panics from https://github.com/boa-dev/boa/issues/883. However, I'm not very sure if this is the right way to do it or if there is a better way to generate errors instead of panicking, so I would appreciate some feedback on it.

### DIFF
--- a/boa/src/builtins/bigint/mod.rs
+++ b/boa/src/builtins/bigint/mod.rs
@@ -173,9 +173,9 @@ impl BigInt {
     pub(crate) fn as_int_n(_this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         let (modulo, bits) = Self::calculate_as_uint_n(args, ctx)?;
 
-        if bits > 0 && modulo >= BigInt::from(2).pow(&BigInt::from(bits as i64 - 1)) {
+        if bits > 0 && modulo >= BigInt::from(2).pow(&BigInt::from(bits as i64 - 1)).unwrap() {
             Ok(Value::from(
-                modulo - BigInt::from(2).pow(&BigInt::from(bits as i64)),
+                modulo - BigInt::from(2).pow(&BigInt::from(bits as i64)).unwrap(),
             ))
         } else {
             Ok(Value::from(modulo))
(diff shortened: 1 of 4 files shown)


boa/src/builtins/bigint/mod.rs
```diff
@@ -173,9 +173,9 @@ impl BigInt {
     pub(crate) fn as_int_n(_this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         let (modulo, bits) = Self::calculate_as_uint_n(args, ctx)?;
 
-        if bits > 0 && modulo >= BigInt::from(2).pow(&BigInt::from(bits as i64 - 1)) {
+        if bits > 0 && modulo >= BigInt::from(2).pow(&BigInt::from(bits as i64 - 1)).unwrap() {
```
alexanderkjall, line 176: Coundn't you use ? instead of .unwrap() since the function returns a Result?

georgeroman: This will always succeed as the exponent is known to be positive (`bits > 0`) so I decided to simply unwrap instead mapping the `Err(String)` to an `Err(Value)` and then use `?`.

Lan2u: Could you please replace the .unwrap() with an .expect() - same outcome but it includes an error message

georgeroman: Done

boa/src/value/operations.rs
```diff
@@ -167,14 +167,22 @@ impl Value {
             (Self::Integer(x), Self::Rational(y)) => Self::rational(f64::from(*x).powf(*y)),
             (Self::Rational(x), Self::Integer(y)) => Self::rational(x.powi(*y)),
 
-            (Self::BigInt(ref a), Self::BigInt(ref b)) => Self::bigint(a.as_inner().clone().pow(b)),
+            (Self::BigInt(ref a), Self::BigInt(ref b)) => Self::bigint(
+                a.as_inner()
+                    .clone()
+                    .pow(b)
+                    .map_err(|msg| ctx.construct_range_error(&msg))?,
```
HalidOdat, line 174:
```suggestion
                    .map_err(|msg| ctx.construct_range_error(msg))?,
```
Its better if we consume the `String` rather than creating a new string from the `&str`

georgeroman: Done

HalidOdat (approved):

Razican (approved):
Looks very good, thanks!

Issue #817, opened 2020-10-08
Test262 panics


Razican:
This is a full list of panics the current ECMAScript test suite (test262) gives, 13-14 in total. We should aim for 0 panics (even before we try to implement the rest of the spec):

<ul>
  <li>
    <details>
      <summary>
        <code>Object already borrowed: BorrowMutError</code> (<a href="https://github.com/boa-dev/boa/issues/1059">#1059</a>), count: 2
      </summary>
      Example:
      <pre>thread 'main' panicked at 'Object already borrowed: BorrowMutError', boa/src/object/internal_methods.rs:198:18</pre>
    </details>
  </li>
  <li>
    <details>
      <summary>
        <code>Identifier arguments has already been declared</code> (<a href="https://github.com/boa-dev/boa/issues/1062">#1062</a>), count: 8
      </summary>
      Example:
      <pre>thread 'main' panicked at 'Identifier arguments has already been declared', boa/src/environment/declarative_environment_record.rs:52:13</pre>
    </details>
  </li>

<li>
    <details>
      <summary>
        <code>range start index 2 out of range for slice of length 0</code> (<a href="https://github.com/boa-dev/boa/issues/1241">#1241</a>), count: 2
      </summary>
      Example:
      <pre>thread 'main' panicked at 'range start index 2 out of range for slice of length 0', boa/src/builtins/function/mod.rs:123:45</pre>
    </details>
  </li>

  <li>
    <details>
      <summary>
        <code>Could not get slice</code> (<a href="https://github.com/boa-dev/boa/issues/1242">#1242</a>), count: 1
      </summary>
      Example:
      <pre>thread 'main' panicked at 'Could not get slice', boa/src/builtins/regexp/mod.rs:413:56</pre>
    </details>
  </li>
</ul>

There is another random panic that seems to not always occur, so we don't have too much information about it, or the exact test where it happens (there are multiple tests with that name):
```
thread 'main' panicked at 'Object already mutably borrowed: BorrowError', boa/src/value/type.rs:50:27
last panic was on test "returns-iterator"
```

It seems to happen [here](https://github.com/boa-dev/boa/blob/master/boa/src/value/type.rs#L50) when calling the `get_type()` function in an object, when checking if the object is a function. This happens if the object is already mutably borrowed, and this borrow will fail. Interestingly, this doesn't happen in every run.

### How to replicate / find code to reproduce:

If you run the test262 suite, you will be able to see for each of the panics which one was the panicking test. To run it:
```
[...]

PR #885, opened 2020-10-17
Fix `BigInt.prototype.toString()` radix checks

georgeroman:
Fixes some radix panics from https://github.com/boa-dev/boa/issues/817


HalidOdat (approved):

Razican (approved):
Good catch, thanks!

PR #886, opened 2020-10-17
Support running a specific test/suite in boa_tester

georgeroman:
These changes add support for running a specific test or suite in `boa_tester`. For example, specifying `-s test/built-ins/BigInt` will only run the `BigInt` suite, while specifying `-s test/built-ins/BigInt/asIntN/arithmetic.js` will only run the specified test. If nothing gets specified, it defaults to running the full suite.

### DIFF
--- a/boa_tester/src/main.rs
+++ b/boa_tester/src/main.rs
@@ -41,7 +41,7 @@ mod read;
 mod results;
 
 use self::{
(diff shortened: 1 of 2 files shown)


jasonwilliams (approved):

Razican (approved):
This is very nice indeed. Thank you!

PR #887, opened 2020-10-17
Add String.search(regexp)

### LINKED ISSUE #117: String.search(regexp)
`search` function would need to be added to the string object here:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/string.rs

It would be very similar to:
https://github.com/jasonwilliams/boa/blob/master/src/lib/js/regexp.rs#L212-L251

Spec:
https://tc39.es/ecma262/#sec-string.prototype.search

Debugging:

flix477:
This Pull Request fixes/closes #117 .

It changes the following:

- Adds String.prototype.search
- Adds RegExp.prototype[@@search]

This is my first PR here, comments are welcome!

### DIFF
--- a/boa/src/builtins/regexp/mod.rs
+++ b/boa/src/builtins/regexp/mod.rs
@@ -502,4 +502,36 @@ impl RegExp {
 
         Ok(result)
     }
+
+    /// `RegExp.prototype[ @@search ]( string )`
+    ///
+    /// The `[@@search]` method executes a search for a match between a `this` regular expression and a string.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-regexp.prototype-@@search
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@search
+    pub(crate) fn search(this: &Value, arg_str: String, ctx: &mut Context) -> Result<Value> {
+        if !this.is_object() {
+            panic!("Not an object");
+        }
+        let last_index = this.get_field("lastIndex").to_index(ctx)?;
+        if last_index != 0 {
+            this.set_field("lastIndex", Value::from(0));
+        }
+
+        let result = Self::exec(this, &[Value::from(arg_str)], ctx)?;
+        let current_last_index = this.get_field("lastIndex").to_index(ctx)?;
+        if current_last_index != last_index {
+            this.set_field("lastIndex", Value::from(current_last_index));
+        }
+
+        if result.is_null() {
+            Ok(Value::from(-1))
+        } else {
+            Ok(result.get_field("index"))
+        }
+    }
 }
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -1175,14 +1175,12 @@ impl String {
     /// [regex]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
     pub(crate) fn search(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
         let re = match args.get(0) {
-            Some(arg) if !arg.is_null() && !arg.is_undefined() => {
-                RegExp::constructor(
-                    &Value::from(Object::default()),
-                    &[Value::from(arg.to_string(ctx)?)],
-                    ctx
-                )
-            }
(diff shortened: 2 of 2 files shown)


RageKnify:
@flix477 you made this branch over an old commit of the `master` branch, could you please merge `master` into this branch so that we get accurate values for the coverage?
I don't think these 2 functions are covered and we'd probably want at least 1 test for `String.search`, that should be enough to cover most of both functions.

boa/src/builtins/regexp/mod.rs
```diff
@@ -502,4 +502,36 @@ impl RegExp {
...
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-regexp.prototype-@@search
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@search
+    pub(crate) fn search(this: &Value, arg_str: String, ctx: &mut Context) -> Result<Value> {
+        if !this.is_object() {
+            panic!("Not an object");
+        }
```
RageKnify, line 519: According to point 2. of the spec this should throw a `TypeError`.
As an example to test this: `RegExp.prototype[Symbol.search].call(0, 'test')` in Chrome throws: `TypeError: Method RegExp.prototype.@@search called on incompatible receiver 0`.
I'm not sure `.prototype` is accessible in `boa` so I don't know if we can test this.

boa/src/builtins/regexp/mod.rs
```diff
@@ -502,4 +502,36 @@ impl RegExp {
...
+    ///
+    /// The `[@@search]` method executes a search for a match between a `this` regular expression and a string.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-regexp.prototype-@@search
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@search
+    pub(crate) fn search(this: &Value, arg_str: String, ctx: &mut Context) -> Result<Value> {
```
RageKnify, line 516: Someone else can confirm, but I'm pretty sure the signature needs to be `(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value>`, all builin functions have the same signature.
This also needs to be declared as a method of the `RegExp` global object, you can see some examples here: https://github.com/boa-dev/boa/blob/2cb24427f1b5e608b453514af413004d3292c2f8/boa/src/builtins/regexp/mod.rs#L81-L83
Note that after the algorithm definition the spec mentions the `"name"` of the function should is `"[Symbol.search]"`.

boa/src/builtins/string/mod.rs
```diff
@@ -1161,6 +1162,30 @@ impl String {
...
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-string.prototype.search
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search
+    /// [regex]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
+    pub(crate) fn search(this: &Value, args: &[Value], ctx: &mut Context) -> Result<Value> {
+        let re = match args.get(0) {
+            Some(arg) if !arg.is_null() && !arg.is_undefined() => RegExp::constructor(
```
RageKnify, line 1178: Can use just 1 function to check for both cases of `null` and `undefined`:
https://github.com/boa-dev/boa/blob/2cb24427f1b5e608b453514af413004d3292c2f8/boa/src/value/mod.rs#L311-L313

RageKnify (changes requested):
Looking good, but needs some fixes. I also don't think your implementation of `String.prototype.search` is spec-compliant, rather than building a `RegExp` object regardless you should check for the `@@search` property. I don't think `GetMethod` is currently defined, but getting the property and checking whether it is a function is enough for now, probably leave a TODO in a comment mentioning it.

Razican:
Hi flix477, please, let us know the state of the changes proposed in this PR :) Check the comments by @HalidOdat and let us know if you have any doubts or want any help!

jasonwilliams:
Hey @flix477 how is this going?

jedel1043:
We should close this PR. Both `String.prototype.search` and `RegExp.prototype[@@search]` are already implemented.

PR #888, opened 2020-10-18
Fix some panics in String.prototype properties

georgeroman:
These changes address the `failed to get argument for String method` panics documented in #817.

### DIFF
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -338,7 +338,8 @@ impl String {
             .to_integer(ctx)? as i32;
         let end = args
             .get(1)
-            .unwrap_or(&Value::integer(length))
+            .cloned()
+            .unwrap_or_else(|| Value::integer(length))
             .to_integer(ctx)? as i32;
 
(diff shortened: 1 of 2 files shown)


HalidOdat (approved):

boa/src/builtins/string/mod.rs
```diff
@@ -325,20 +327,20 @@ impl String {
...
             .get(0)
-            .expect("failed to get argument for String method")
+            .cloned()
+            .unwrap_or_else(Value::undefined)
             .to_integer(ctx)? as i32;
-
         let end = args
             .get(1)
-            .expect("failed to get argument in slice")
+            .unwrap_or(&Value::integer(length))
```
Razican, line 341: Wouldn't this be more performant if it used `unwrap_or_else()`?

Razican (approved):
Looks good to me, but check my comment to see if we can improve performance a little bit.

boa/src/builtins/string/mod.rs
georgeroman: Right, done

PR #889, opened 2020-10-18
CI workflow improvements (mostly Test262)

Razican:
This requires some testing and probably some improvements/fixes, but I wanted to open a PR to get some feedback.

I have moved the PR workflows to run on `pull_request_target` instead of `pull_request`. This should allow the criterion comparisons to be posted as comments even for people outside the org, I think.

I have also added a system to compare two Test262 results. For now, the comparison is very simple, it will just print the counts of passed/ignored/failed/panics in Test262 results. But it will generate some Markdown and it should post a comment in the PR. This comment would be re-written on each change (so no new comments would be added).

I have also improved a bit how the Test262 testing is done. We now ignore some extra tests that could not be passed. We now count as 2 the tests that need to be executed twice (in strict mode and in non-strict mode). And, I added an extra parameter to the parser to directly start in strict mode, so that we can test that.

There are some things left in the tester that is probably making some tests fail:
 - We are not creating the `$262` object yet (some of its methods will be tricky to implement)
 - The `print()` global function just panics as unimplemented for now (but we are ignoring `await` tests, so it's not a big deal)
 - We don't follow 100% the negative results, since we are not really working with parsing/early/resolution/runtime errors, but should work for some cases.

This requires a small change in the results format, that I should have included in my last PR, but I forgot, so I might need to do an extra PR to fix results in the GitHub Pages.

I also took the opportunity to update the test suite and to upgrade all dependencies.


Razican:
I think this is now ready for review. I would like some feedback on this:
```
Removed some test that now pass properly from the ignore list. This
now also allows filtering out features, so I filtered out the
`TypedArray` feature that was failing all the tests. There are
other 3 features (at least) that we don't implement, but they seem
to have passing tests, so I commented them out.
```

And any general feedback about the PR.

boa_tester/src/main.rs
```diff
@@ -54,55 +55,174 @@ use std::{
...
+        }
+        Cli::Compare {
+            base,
+            new,
+            markdown,
+        } => compare_results(base.as_path(), new.as_path(), markdown),
+    }
+}
+
+/// Runst the full test suite.
```
RageKnify, line 223: `Runst -> Runs`

HalidOdat (approved):
This looks awesome! :D 

BTW shouldn't the test262 comment be posted on this PR

Razican:
> This looks awesome! :D
> 
> BTW shouldn't the test262 comment be posted on this PR

Good question, but since I have changed the `pull_request` triggers to `pull_request_target`, it will run the flows as if it were in the `master` branch. This allows adding comments by anyone (even non-project members) as far as I understand, but will only work if the trigger is in master.

Something I could try would be to temporarily set the workflow trigger to `pull_request` to see if it works. But for it to work, we need to first merge #895. But, in that case, we could not merge any other PR before this one (or we would break the CI).

HalidOdat:
> Something I could try would be to temporarily set the workflow trigger to `pull_request` to see if it works. But for it to work, we need to first merge #895. But, in that case, we could not merge any other PR before this one (or we would break the CI).

we could try this, there aren't that any PRs (besides #887 but review is still needed) that are ready to be merged anyway, even if there were we can postpone them.

Razican:
> > Something I could try would be to temporarily set the workflow trigger to `pull_request` to see if it works. But for it to work, we need to first merge #895. But, in that case, we could not merge any other PR before this one (or we would break the CI).
> 
> we could try this, there aren't any PRs (besides #887 but review is still needed) that are ready to be merged anyway, even if there were we can postpone them.

OK, let's merge it and I will change the triggers. Here are the new results, though:
```
Results:
Total tests: 78413
Passed tests: 18185
Ignored tests: 15507
Failed tests: 44721 (panics: 13519)
Conformance: 23.19%
```

Conformance is a bit higher, and we now properly count tests that we run twice.

RageKnify (approved):

HalidOdat:
@Razican the workflow seems to be failing on `Write a new comment` with error `Error: Missing comment 'body'.`

Razican:
It seems it's not capturing the full comment markdown, I will re-check that.

HalidOdat:
> It seems it's not capturing the full comment markdown, I will re-check that.

also shouldn't it be overriding the original comment?

.github/workflows/pull_request.yml
```diff
@@ -1,4 +1,4 @@
-on: [pull_request]
+on: [pull_request_target]
```
jasonwilliams, line 1: From what I understand in the documentation, anything using `pull_request_target` doesn't actually use the code PR'd. Instead it uses the base repository (master in our case). So I don't think this will work.

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target

jasonwilliams: I don't think there is a fix to the comment issue:
https://github.com/actions/first-interaction/issues/10

Razican: I think it will use the code of the PR but the workflow of `master`, if I understood it correctly.

Razican:
> > It seems it's not capturing the full comment markdown, I will re-check that.
> 
> also shouldn't it be overriding the original comment?

Yep, I have identified 2 bugs: it ignores lines after the first one in the output of the comparison, and it will not override the previous comment. I will give them a look now.

.github/workflows/pull_request.yml
jasonwilliams: oh right I see, ok cool lets try it

jasonwilliams (approved):

Razican:
I think this is ready to be merged. It looks that the comments and so on are working. I don't think we can check if the `pull_request_target` works until we actually get a Pull Request (or somebody synchronizes theirs).

PR #896, opened 2020-10-19
Fix small issue with the CI

Razican:
This fixes the Test262 uploading to our `gh-pages` branch.


HalidOdat (approved):

Razican:
It turns out @jasonwilliams was right, the `pull_request_target` action doesn't pull the changes from the PR, it will execute all the stuff from master. I updated the workflow to make sure we at least get the data for the people in boa-devs. We can still check it for the rest of the PRs by going to the details of the CI.

PR #901, opened 2020-10-20
Added "unimplemented" syntax errors

Razican:
This removes the panics in `for...in` parsing. It will still fail, but at least we don't bring the process down. This comes from the idea that "every panic is a bug", not just an unimplemented feature.


HalidOdat (approved):
Looks good to me! :)

RageKnify (approved):

PR #902, opened 2020-10-20
Fix Accessors panics

HalidOdat:
It changes the following:
- Accessors now return undefined.

### DIFF
--- a/boa/src/object/internal_methods.rs
+++ b/boa/src/object/internal_methods.rs
@@ -83,8 +83,9 @@ impl GcObject {
                 parent.get_field(key.clone())
             }
             Some(ref desc) => match desc {
-                PropertyDescriptor::Accessor(_) => todo!(),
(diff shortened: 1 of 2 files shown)


RageKnify (approved):

Razican (approved):
Thanks! Removing the panics should improve our conformance, and at at least we don't bring down the process executing the interpreter. Looks good to me, I'm a bit intrigued by the benchmarks, but I hope it's just statistical anomalies.

HalidOdat:
> Looks good to me, I'm a bit intrigued by the benchmarks, but I hope it's just statistical anomalies.

That is very strange. You are right, it's probably a statistical anomaly since there is a performance regression in parse/lex benchmarks, I haven't changed anything in the parser/lexer.

Razican:
Cool, let's merge then :D

PR #906, opened 2020-10-21
Removed a bunch of Test262 panics

Razican:
This changes the behaviour of the tester when the environment cannot be correctly set up. Instead of panicking, it just fails the test without the panic. I also removed some of the most notable panics (such as the `instanceof`), and improved the very verbose `-vv` output.

The number of panics should be greatly reduced with this.


boa_tester/src/read.rs
```diff
@@ -186,7 +186,8 @@ fn read_metadata(code: &str) -> io::Result<MetaData> {
         .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidData, "no metadata found"))?
         .get(1)
         .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidData, "no metadata found"))?
-        .as_str();
-
-    serde_yaml::from_str(yaml).map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))
+        .as_str()
+        .replace('\r', "\n");
+    dbg!(&yaml);
```
RageKnify, line 191: Was this just for debugging or is it meant to stay?

HalidOdat (approved):

RageKnify (approved):
I think I found a forgotten `dbg!()`, the rest looks good to me.

boa_tester/src/read.rs
Razican: Good catch :) removed!

Lan2u (approved):

PR #909, opened 2020-10-22
Swap the ordering of create_intrinsics() and only call if parsing is successful

jasonwilliams:
## Description

The main issue we have is that `create_intrinsics` fire regardless of syntax parsing status.
This means we spend a lot of time creating built-ins when we don't need to.

This PR swaps the order around (from intrinics then parsing, to parsing then intrinics).

I believe any test (test262) that relies on negative syntax, should fail a lot faster (70% faster).

## Main changes

* Create flag for when intriniscs are set, this is because eval can be called multiple times.
* Parsing Errors cannot utilise the realm
* Tests that don't go via forward need to call create_intrinsics()

Profiles taken below are for negative syntax

Before (15ms):
profile

After (0.7ms):
profile_after

### DIFF
--- a/boa/src/context.rs
+++ b/boa/src/context.rs
@@ -196,6 +196,9 @@ pub struct Context {
 
     /// Cached standard objects and their prototypes
     standard_objects: StandardObjects,
+
+    // Flag for when intrinsics are already set
+    intrinsics_set: bool,
 }
 
 impl Default for Context {
@@ -203,7 +206,7 @@ impl Default for Context {
         let realm = Realm::create();
         let executor = Interpreter::new();
         let (well_known_symbols, symbol_count) = WellKnownSymbols::new();
-        let mut context = Self {
+        Self {
             realm,
             executor,
             symbol_count,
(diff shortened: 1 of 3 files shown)


boa/src/context.rs
```diff
@@ -635,8 +636,15 @@ impl Context {
...
-            Err(e) => self.throw_syntax_error(e),
+            Ok(statement_list) => {
+                // Once parsing has been successful build the realm
+                self.create_intrinsics();
+                statement_list.run(self)
+            }
+            Err(e) => Err(Value::String(RcString::from(format!(
+                "SyntaxError\": \"{}",
+                e
+            )))),
```
HalidOdat, line 647: we should not be throwing a string that contains `SyntaxError` we should throw an actual `SyntaxError` object

jasonwilliams: I’ll have to look at how we can do that without requiring the realm. Right now the convenience method we have on the Context uses the realm so we can’t use that.

jasonwilliams: @HalidOdat Yeah i think there's not much we can do here then, it wont be correct if we hard code `SyntaxError` and the code modifies it.

HalidOdat:
Closing this, once we find a better way of doing this we can open it again.

PR #912, opened 2020-10-23
Feature `Context::register_global_property()`

### LINKED ISSUE #891: Enable creating a global `Object`
Currently, we can create a new object type and add it to the global scope by calling `register_global_class()`. This enables later calling `new Person()` for example, as we can see in our classes [example](https://github.com/boa-dev/boa/blob/master/boa/examples/classes.rs).

Nevertheless, this doesn't allow creating an ordinary object on a property in the global object, and add some methods to its prototype. This is required by the Test262 suite, as it can be seen [here](https://github.com/tc39/test262/blob/main/INTERPRETING.md#host-defined-functions).

For this particular case, we just need to add an ordinary (`{}`) object to the global scope, under the `$262` name, and add some functions to its prototype, such as `createRealm`, `detachArrayBuffer`...

I am not aware if this is possible to do from an outside crate, I think it's not (maybe @HalidOdat can confirm, as he did the class builder). But in any case, this should have an easy to use interface, as it happens with the global class / function creation.

HalidOdat:
This Pull Request fixes/closes #891.

It changes the following:
 - Add `Context::register_global_property`

This allows us to register any global property (primitive value, object, constructor, etc) with a property key  (symbol, string, number) and the attributes that the property will be bound to the global object. (see `register_global_property` doc test/example)

### DIFF
--- a/boa/src/context.rs
+++ b/boa/src/context.rs
@@ -10,7 +10,7 @@ use crate::{
     class::{Class, ClassBuilder},
     exec::Interpreter,
     object::{GcObject, Object, ObjectData, PROTOTYPE},
-    property::{DataDescriptor, PropertyKey},
+    property::{Attribute, DataDescriptor, PropertyKey},
     realm::Realm,
     syntax::{
         ast::{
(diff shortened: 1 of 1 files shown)


Razican (approved):
Looks very good! Just one question, could this be used to replace `register_global_function()`? Even then, I think it would make sense to have the helper, as it makes things much easier.

HalidOdat:
> Looks very good! Just one question, could this be used to replace `register_global_function()`? Even then, I think it would make sense to have the helper, as it makes things much easier.

Yes. it can replace it, we can create a function object with `FunctionBuilder` and register the function object.
And I agree we should keep it as a helper function, since it's more simple to register a function with `register_global_function`

RageKnify (approved):

PR #913, opened 2020-10-24
Added check to ignore semicolon in parser

### LINKED ISSUE #892: Empty statements not fully implemented
Currently, a code like this one will work fine:
```javascript
let a = 10;;;
```

But, this one is broken:
```javascript
;;;let a = 10;;
```

This is because if a statement list starts with an empty statement, we only check if it's a closing brace or not, bug it it's not, we just try to parse a `StatementListItem`, which will fail.

All this is implemented in [this](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/parser/statement/mod.rs#L308-L338) loop, and an equivalent one [here](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/parser/statement/mod.rs#L269-L290).

As you can see, before looping again, we do consume any repeated semicolons, so the first example works, but not the second one.

Ideally, we should add an extra branch to the first check in both loops, to see if the next item is a semicolon. If it is, just call `continue;` after consuming the token. With this, we can remove the while loop at the end of the loop.

AngelOnFira:
This Pull Request closes #892.

It changes the following:

- Empty statements are now parsed without error
- Semicolons are no longer skipped at the end of the parse loop

In the original issue, it was mentioned that there might be a problem with empty `StatementLists` as well. I couldn't find a case where it didn't work as is, but I still brought the semicolon skip up a little higher.

I'm not sure if any specific tests should be written for this, but if so I can add them.

This PR improves conformance by 0.07% :smile:

### DIFF
--- a/boa/src/syntax/parser/mod.rs
+++ b/boa/src/syntax/parser/mod.rs
@@ -9,7 +9,7 @@ mod statement;
 mod tests;
 
 pub use self::error::{ParseError, ParseResult};
-use crate::syntax::{ast::node::StatementList, ast::Punctuator, lexer::TokenKind};
+use crate::syntax::{ast::node::StatementList, lexer::TokenKind};
 
 use cursor::Cursor;
 
@@ -139,9 +139,6 @@ where
     }
 }
 
-/// The possible TokenKind which indicate the end of a case statement.
-const SCRIPT_BREAK_TOKENS: [TokenKind; 1] = [TokenKind::Punctuator(Punctuator::CloseBlock)];
-
 /// Parses a script body.
 ///
 /// More information:
@@ -158,7 +155,6 @@ where
     type Output = StatementList;
 
     fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
-        self::statement::StatementList::new(false, false, false, false, true, &SCRIPT_BREAK_TOKENS)
-            .parse(cursor)
+        self::statement::StatementList::new(false, false, false, true, &[]).parse(cursor)
     }
 }
--- a/boa/src/syntax/parser/mod.rs
+++ b/boa/src/syntax/parser/mod.rs
@@ -9,7 +9,7 @@ mod statement;
 mod tests;
 
 pub use self::error::{ParseError, ParseResult};
-use crate::syntax::{ast::node::StatementList, lexer::TokenKind};
+use crate::syntax::{ast::node::StatementList, ast::Punctuator, lexer::TokenKind};
 
 use cursor::Cursor;
 
@@ -139,6 +139,9 @@ where
     }
 }
 
+/// The possible TokenKind which indicate the end of a case statement.
+const SCRIPT_BREAK_TOKENS: [TokenKind; 1] = [TokenKind::Punctuator(Punctuator::CloseBlock)];
+
 /// Parses a script body.
 ///
 /// More information:
@@ -155,6 +158,7 @@ where
     type Output = StatementList;
 
     fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
-        self::statement::StatementList::new(false, false, false, false, true).parse(cursor)
+        self::statement::StatementList::new(false, false, false, false, true, &SCRIPT_BREAK_TOKENS)
+            .parse(cursor)
     }
 }
--- a/boa/src/syntax/parser/statement/mod.rs
+++ b/boa/src/syntax/parser/statement/mod.rs
@@ -214,15 +214,16 @@ where
 ///
 /// [spec]: https://tc39.es/ecma262/#prod-StatementList
 #[derive(Debug, Clone, Copy)]
-pub(super) struct StatementList {
+pub(super) struct StatementList<'a> {
     allow_yield: AllowYield,
     allow_await: AllowAwait,
     allow_return: AllowReturn,
     break_when_closingbraces: bool,
     in_block: bool,
+    break_nodes: &'a [TokenKind],
 }
 
-impl StatementList {
+impl<'a> StatementList<'a> {
     /// Creates a new `StatementList` parser.
     pub(super) fn new<Y, A, R>(
         allow_yield: Y,
@@ -230,6 +231,7 @@ impl StatementList {
         allow_return: R,
         break_when_closingbraces: bool,
         in_block: bool,
+        break_nodes: &'a [TokenKind],
     ) -> Self
     where
         Y: Into<AllowYield>,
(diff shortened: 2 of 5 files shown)


boa/src/syntax/parser/statement/mod.rs
```diff
@@ -214,22 +214,24 @@ where
 ///
 /// [spec]: https://tc39.es/ecma262/#prod-StatementList
 #[derive(Debug, Clone, Copy)]
-pub(super) struct StatementList {
+pub(super) struct StatementList<'a> {
     allow_yield: AllowYield,
     allow_await: AllowAwait,
     allow_return: AllowReturn,
     break_when_closingbraces: bool,
```
Lan2u, line 221: Looks really good! Just a couple of minor things which might be nice to include. 'break_when_closingbraces' no-longer seems to actually reflect the new behaviour as any of the break nodes will trigger the check - can you either rename it or look to remove it entirely.

Lan2u (commented):
Looks really good! Just a couple of minor things which might be nice to include :)

Some tests showing the empty statement parsing etc. would also be nice - if the code is going to change later I'd argue they are even more valuable

croraf:
I suggested on Discord to split this into two PRs. This one with a parse method refactor. And the new one with the EmptyStatement correct parsing solution.

For the second one the idea is not to skip the semicolon but to parse it into EmptyStatement node, as does esprima.org parser:
image

Razican:
I will review this as soon as I have time, hopefully today.

> I suggested on Discord to split this into two PRs. This one with a parse method refactor. And the new one with the EmptyStatement correct parsing solution.
> 
> For the second one the idea is not to skip the semicolon but to parse it into EmptyStatement node, as does esprima.org parser

This makes sense from a Parser-only perspective, but the issue with this is that adds extra allocations in the `StatementList`, by adding some `EmptyStatement` nodes that will need to be dynamically called `run()` during the execution, which will be just a no-op, but will increase cache misses and memory usage, note that all node variants have the same size in memory.

The current version just ignores `EmptyStatement`s, and will not add them to an `StatementList`, which is far more efficient. And having an `EmptyStatement` in the `StatementList` doesn't really provide any useful information. The only use case I can think of would be if we are trying to re-generate the original code, but we are already ignoring comments, for example, so this is currently not possible, and I think it should be the application who extracts the original code again if needed.

In my opinion we shouldn't add an `EmptyStatement` parsing and node, for performance reasons.

croraf:
> I will review this as soon as I have time, hopefully today.
> 
> > I suggested on Discord to split this into two PRs. This one with a parse method refactor. And the new one with the EmptyStatement correct parsing solution.
> > For the second one the idea is not to skip the semicolon but to parse it into EmptyStatement node, as does esprima.org parser
> 
> This makes sense from a Parser-only perspective, but the issue with this is that adds extra allocations in the `StatementList`, by adding some `EmptyStatement` nodes that will need to be dynamically called `run()` during the execution, which will be just a no-op, but will increase cache misses and memory usage, note that all node variants have the same size in memory.
> 
> The current version just ignores `EmptyStatement`s, and will not add them to an `StatementList`, which is far more efficient. And having an `EmptyStatement` in the `StatementList` doesn't really provide any useful information. The only use case I can think of would be if we are trying to re-generate the original code, but we are already ignoring comments, for example, so this is currently not possible, and I think it should be the application who extracts the original code again if needed.
> 
> In my opinion we shouldn't add an `EmptyStatement` parsing and node, for performance reasons.

We will then not conform to the parsing and evaluation described in the ES, which possibly introduces errors and makes it harder to follow.

boa/src/syntax/parser/mod.rs
```diff
@@ -139,6 +139,9 @@ where
     }
 }
 
+/// The possible TokenKind which indicate the end of a case statement.
+const SCRIPT_BREAK_TOKENS: [TokenKind; 1] = [TokenKind::Punctuator(Punctuator::CloseBlock)];
```
Razican, line 143: Is this defined somewhere? I believe a script will never end with a `}`, right? I think this list should be empty, and in that case, we should add the exception to the parser so that it doesn't fail with an `AbruptEnd` error.

boa/src/syntax/parser/statement/mod.rs
Razican: I think that if we use `}` as a `break_node`, then we don't need this anymore, and can remove it from the structure.

boa/src/syntax/parser/statement/mod.rs
```diff
@@ -214,22 +214,24 @@ where
...
-impl StatementList {
+impl<'a> StatementList<'a> {
     /// Creates a new `StatementList` parser.
     pub(super) fn new<Y, A, R>(
         allow_yield: Y,
         allow_await: A,
         allow_return: R,
         break_when_closingbraces: bool,
         in_block: bool,
+        break_nodes: &'a [TokenKind],
```
Razican, line 234: I think it makes sense to enforce this to be `&'static`, and that way we remove some extra lifetimes. I don't think it makes sense to customize this on runtime.

Razican (commented):
I like this new implementation, should solve the issue we are having with empty statements :)

Razican:
After some discussion on discord, it seems that EmptyStatements are indeed important, and they will be implemented fully in the future, as @croraf mentions. For now, I would merge this as soon as the review comments are solved.

croraf:
Can we merge just the refactor part (which is the majority of this PR), without the semicolon part?

AngelOnFira:
Sorry @croraf I'll be working through the comments today, just got hit with a ton of work this week :sweat_smile:

croraf:
@AngelOnFira No hurry, we were just discussing.

boa/src/syntax/parser/statement/mod.rs
AngelOnFira: Ok, I took it out and everything appears to still be fine.

boa/src/syntax/parser/statement/mod.rs
AngelOnFira: This is purely due to my lack of understanding Rust lifetimes past what the compiler tells me to do :sweat_smile: I'll switch it up :100:

boa/src/syntax/parser/mod.rs
AngelOnFira: No, I just added it cause I thought I needed to. I'm going to reassess each one and see what it's proper closing should be.

AngelOnFira: Ok, I think this should be resolved. I'm not sure I understand what an exception should be, I ran the tests and I'm getting the same percentage. Can you give me an example of where this might occur?

AngelOnFira:
Ok @croraf I've reverted the semicolon changes for this PR. @Razican I believe I've addressed all of your comments, but a few need a bit more discussion to confirm that they are done. Please let me know if there is anything else to take care of! :smile:

croraf:
Cool. After this bigger addressing you can make a new PR for the semicolon changes if you wish, it should be very easy to solve that.

AngelOnFira:
Oops still have to make the tests :P

AngelOnFira:
Actually I think the tests don't apply here since this is just a refactor.

AngelOnFira:
@Razican @Lan2u Just want to check in. I think this PR is ready to go, but please ping me if anything else needs to be done :smile:

Lan2u (approved):

Razican (approved):
Looks good, thanks!

  | Passed | 18,945 | 18,945 | 0 |
  | Failed | 43,923 | 43,923 | 0 |
  | Conformance | 24.16 | 24.16 | 0.00% |

croraf:
@AngelOnFira Gj. Are you willing to make another PR with the semicolon issue fix?

AngelOnFira:
@croraf ya, I'll try to do that now that this is merged.

PR #915, opened 2020-10-27
Improve lexer by make cursor iterate over bytes

### LINKED ISSUE #335: Optimize the lexer by iterating over bytes
As `maciejh` mentioned in [Reddit][reddit_comment], iterating over characters is not fast. We should iterate over bytes, but we have to take into account that identifiers, strings, comments and more can have non-ASCII characters, since JavaScript is UTF-8.

Nonetheless, we can use byte-iteration for detecting language features much faster.

The comment:
> Iterating over bytes rather than chars is inherently much faster, non-ascii UTF8 bytes are all >=128, and you might only really care about those being correct when parsing idents.

[reddit_comment]: https://www.reddit.com/r/rust/comments/g3m5i7/boa_releases_v07_with_a_much_faster_parser/fnsng09

jevancc:
This Pull Request fixes/closes #335 . Notice that **this PR does not change any behavior of the existing lexer**.

It changes the following:
- Rewrite cursor to iterate over bytes (u8) and Unicode chars (u32, code points)
- Update lexers for the new cursor

Not covered in this PR:
- Full UTF-16 support for string and regex
- Fix bug and panic of the lexer
- Read input in bytes instead of str (involving lots of changes on tests. I will create a separate PR for it once this PR is merged)


jevancc:
This PR is ready for review now. Unfortunately, it is hard to tell that this idea, i.e. iterating over bytes instead of chars, and the implementation improves the performance from the existing benchmarks. However, it is still good to have it since we can skip/handle/report the position of invalid char with the new lexer instead of panic when reading the input.

boa/src/syntax/lexer/cursor.rs
```diff
@@ -150,36 +212,74 @@ where
...
+        Ok(ch)
     }
 }
 
 /// Inner iterator for a cursor.
 #[derive(Debug)]
 struct InnerIter<R> {
     iter: Bytes<R>,
-    peeked_char: Option<Option<char>>,
+    // peeked_char: Option<Option<char>>,
```
Lan2u, line 279: left over comment

Lan2u (approved):
Thanks for this - looks good :)

boa/src/syntax/lexer/cursor.rs
jevancc: Removed. Thanks for pointing out!

HalidOdat (approved):

jasonwilliams (approved):

jasonwilliams:
This looks great to me, great work @jevancc

Razican:
Benchmark results are looking good. No big speed up, but no regressions.

Test results look good too:

  | Passed | 18,953 | 18,953 | 0 |
  | Failed | 43,915 | 43,915 | 0 |
  | Conformance | 24.17 | 24.17 | 0.00% |

No conformance changes, so I'm happy with this, it opens new doors. Thanks for your work!

PR #916, opened 2020-10-27
Code cleanup and inline

HalidOdat:
It changes the following:
- Change `Context` variable name to `context` (instead of `engine` or `interpreter`)
- Added `#[inline]`
- change gc import to use gc module


boa/src/builtins/infinity/mod.rs
```diff
@@ -25,7 +25,7 @@ impl BuiltIn for Infinity {
         Attribute::READONLY | Attribute::NON_ENUMERABLE | Attribute::PERMANENT
     }
 
-    fn init(_context: &mut Context) -> (&'static str, Value, Attribute) {
+    fn init(_: &mut Context) -> (&'static str, Value, Attribute) {
```
Razican, line 28: Why this change?

boa/src/builtins/nan/mod.rs
```diff
@@ -26,7 +26,7 @@ impl BuiltIn for NaN {
         Attribute::READONLY | Attribute::NON_ENUMERABLE | Attribute::PERMANENT
     }
 
-    fn init(_context: &mut Context) -> (&'static str, Value, Attribute) {
+    fn init(_: &mut Context) -> (&'static str, Value, Attribute) {
```
Razican, line 29: Why this change?

Razican (commented):
This looks good to me, but why was the test262 submodule changed? It looks like it was reverted to a version in September. Could we do a `git pull origin main` in there?

boa/src/builtins/infinity/mod.rs
HalidOdat: its because the context was not used to changed it to `_` like the other builtin functins

RageKnify (approved):

RageKnify:
My bad, new `commit` should work ok.

Razican (approved):

PR #919, opened 2020-10-28
Reduce the number of `Array`-related panics

jakubfijalkowski:
Hello :wave: !

This Pull Request addresses some of the panics in Test262 suite #817, most notably the ones in `Array` module.

They (most of the time) boil down to changing `get_field("length").as_number().unwrap()` to `get_field("length").to_length(context)`. I needed to add `context: &mut Context` to `construct_array` and `add_to_array_object` because of that. I've also added a little bit of `+/-Inf`, `NaN` and out-of-range checks to the code.

Basically, this reduced the number of panics in `Array` part of the test suite to 8 (and these are related to `display`).

The main problem that I see with the changes are the semi-random `usize`/`isize`/`i32`/`f64` casts, but I really don't know what would be best to do with them and would appreciate a little bit of guidance here.


Razican:

| Passed | 18,945 | 19,353 | **+408** | 

| Failed | 43,923 | 43,515 | **-408** | 

| Panics | 1,127 | 465 | **-662** | 

| Conformance | 24.16 | 24.68 | **+0.52%** |

RageKnify:
I think following the advice in [this SO post](https://stackoverflow.com/a/28280042) we should avoid conversions with `as isize` or `as usize` and use `From` or `TryFrom` instead with for example: `usize::from(x)` that makes it clearer that it is a safe conversion.
@HalidOdat @Razican what's your opinion? Since integer conversion is the sort of thing that can easily go wrong I'd rather be explicit about it.

Razican:
> I think following the advice in [this SO post](https://stackoverflow.com/a/28280042) we should avoid conversions with `as isize` or `as usize` and use `From` or `TryFrom` instead with for example: `usize::from(x)` that makes it clearer that it is a safe conversion.
> @HalidOdat @Razican what's your opinion? Since integer conversion is the sort of thing that can easily go wrong I'd rather be explicit about it.

Yes, we should not use `as`. But I'm happy to merge this as-is, since we use it widely through the codebase and that would require a different PR, I think.

RageKnify:
> Yes, we should not use `as`. But I'm happy to merge this as-is, since we use it widely through the codebase and that would require a different PR, I think.

I figured since we're changing this part of the code we could try to clean up this part of the code and as other parts are changed in PRs we can ask that those PRs remove those uses.
I preffer having safe conversions as we go since I can trust the code more if it uses safe covnersions.

Razican:
> > Yes, we should not use `as`. But I'm happy to merge this as-is, since we use it widely through the codebase and that would require a different PR, I think.
> 
> I figured since we're changing this part of the code we could try to clean up this part of the code and as other parts are changed in PRs we can ask that those PRs remove those uses.
> I preffer having safe conversions as we go since I can trust the code more if it uses safe covnersions.

OK, let's do that :)

boa/src/builtins/array/mod.rs
```diff
@@ -691,23 +691,26 @@ impl Array {
...
-                if from_idx < 0 {
-                    len + from_idx
+                if !from_idx.is_finite() {
+                    return Ok(Value::from(-1));
+                } else if from_idx < 0.0 {
+                    let k = len as isize + from_idx as isize;
+                    max(0, k) as usize
                 } else {
-                    from_idx
+                    from_idx as usize
```
Razican, line 713: We should use `usize::try_from()` here.

boa/src/builtins/array/mod.rs
```diff
@@ -744,26 +747,29 @@ impl Array {
...
 
-                if from_idx >= 0 {
-                    min(from_idx, len - 1)
+                if !from_idx.is_finite() {
+                    return Ok(Value::from(-1));
+                } else if from_idx >= 0.0 {
+                    min(from_idx as isize, len - 1)
                 } else {
-                    len + from_idx
+                    len + from_idx as isize
```
Razican, line 772: We should use `isize::try_from()` here.

boa/src/builtins/array/mod.rs
```diff
@@ -773,7 +779,8 @@ impl Array {
             let check_element = this.get_field(idx).clone();
 
             if check_element.strict_equals(&search_element) {
-                return Ok(Value::from(idx));
+                // TODO: make the cast from isize to i32 safe
+                return Ok(Value::from(idx as i32));
```
Razican, line 783: We should use `i32::try_from(idx)` here.

boa/src/builtins/array/mod.rs
```diff
@@ -846,11 +853,12 @@ impl Array {
             let result = context.call(predicate_arg, &this_arg, &arguments)?;
 
             if result.to_boolean() {
-                return Ok(Value::rational(f64::from(i)));
+                // TODO: make the case safe
+                return Ok(Value::integer(i as i32));
```
Razican, line 857: We should use `i32::try_from(i)` here.

boa/src/builtins/array/mod.rs
```diff
@@ -865,16 +873,16 @@ impl Array {
     /// [spec]: https://tc39.es/ecma262/#sec-array.prototype.fill
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill
     pub(crate) fn fill(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
-        let len: i32 = this.get_field("length").as_number().unwrap() as i32;
+        let len = this.get_field("length").to_length(context)? as isize;
```
Razican, line 876: We should use `isize::try_from()` here.

boa/src/builtins/array/mod.rs
```diff
@@ -865,16 +873,16 @@ impl Array {
...
 
         let default_value = Value::undefined();
         let value = args.get(0).unwrap_or(&default_value);
-        let relative_start = args.get(1).unwrap_or(&default_value).to_number(context)? as i32;
+        let relative_start = args.get(1).unwrap_or(&default_value).to_number(context)? as isize;
```
Razican, line 880: Another one that should be using `isize::try_from()`.

boa/src/builtins/array/mod.rs
```diff
@@ -865,16 +873,16 @@ impl Array {
...
         let relative_end_val = args.get(2).unwrap_or(&default_value);
         let relative_end = if relative_end_val.is_undefined() {
             len
         } else {
-            relative_end_val.to_number(context)? as i32
+            relative_end_val.to_number(context)? as isize
```
Razican, line 885: Another one that should be using `isize::try_from()`.

boa/src/builtins/array/mod.rs
```diff
@@ -936,26 +948,30 @@ impl Array {
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
     pub(crate) fn slice(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
         let new_array = Self::new_array(context)?;
-        let len = this.get_field("length").as_number().unwrap() as i32;
+        let len = this.get_field("length").to_length(context)? as isize;
```
Razican, line 951: Another one that should be using `isize::try_from()`.

boa/src/builtins/array/mod.rs
```diff
@@ -936,26 +948,30 @@ impl Array {
...
-            Some(v) => v.as_number().unwrap() as i32,
-            None => 0,
+            Some(v) => v.to_integer(context)?,
+            None => 0.0,
         };
         let end = match args.get(1) {
-            Some(v) => v.as_number().unwrap() as i32,
-            None => len,
+            Some(v) => v.to_integer(context)?,
+            None => len as f64,
```
Razican, line 959: This should use `f64::try_from()` I think.

boa/src/builtins/array/mod.rs
```diff
@@ -936,26 +948,30 @@ impl Array {
...
         };
 
-        let from = if start < 0 {
-            max(len.wrapping_add(start), 0)
+        let from = if !start.is_finite() {
+            0
+        } else if start < 0.0 {
+            max(len.wrapping_add(start as isize), 0)
```
Razican, line 965: Another one that should be using `isize::try_from()`.

boa/src/builtins/array/mod.rs
```diff
@@ -936,26 +948,30 @@ impl Array {
...
         } else {
-            min(start, len)
+            min(start as isize, len)
```
Razican, line 967: Another one that should be using `isize::try_from()`.

boa/src/builtins/array/mod.rs
```diff
@@ -936,26 +948,30 @@ impl Array {
...
         };
-        let to = if end < 0 {
-            max(len.wrapping_add(end), 0)
+        let to = if !end.is_finite() {
+            0
+        } else if end < 0.0 {
+            max(len.wrapping_add(end as isize), 0)
         } else {
-            min(end, len)
+            min(end as isize, len)
```
Razican, line 974: Another one that should be using `isize::try_from()`.

Razican (changes requested):
Looks very good, and I'm loving the new panic count :D 
As we mentioned in the comments, it's best if we use a `try_from()` when casting, or `from()` if possible, to avoid casting issues. Moreover, I see we use `wrapping_add()` in many places. I'm not familiar with the spec, but are we sure we should be using a wrapping addition?

boa/src/builtins/array/mod.rs
jakubfijalkowski: @Razican I will have some time this evening (hopefully) and will go through the whole file, don't need to waste time searching for all the casts. :)

jakubfijalkowski:
> Moreover, I see we use wrapping_add() in many places. I'm not familiar with the spec, but are we sure we should be using a wrapping addition?

When reading the ECMA specs I had a feeling that they assume that overflow can't happen, since everywhere they operate on JS `Number` and not on concrete types. Moreover, since this is JS, most of the time they assumed it is floating point number and handled the `Infinity` cases explicitly, which I also added in some places.

I think that the whole length/index mgmt needs to be thought over - the new ECMA draft (30.10) uses `ToIntegerOrInfinity` to handle the `Infinity`/`NaN` case (esp. `"Infinity"` interpreted as number which is quite common in the suite :) ) and I think the current code deliberately ignores that (which makes some tests to fail) or is based on some older draft (since it uses `ToInteger` that currently is not defined in the spec).

I will try to write a little bit longer summary of my findings and dig a little bit deeper in the standard to find out what they assume should be done in this case.

Razican:
This might also have to do with the fact that we use integers just as an optimization for numeric types. Working with integers is faster than working with floats, but, of course, they are not spec compliant, so sometimes we have to transform them back to floats.

boa/src/builtins/array/mod.rs
jakubfijalkowski: There is no `TryFrom<isize>` for `f64`, unfortunately. However, the cast should be safe (that is true for `isize == i32`, I'm not so sure for `isize == i64`).

jakubfijalkowski:
I went through the file and changed all the casts to `try_from`/`try_into`. There are, unfortunately, some rough edges.

Since now the conversions can fail, we need to map errors. I put the `impl From<std::num::TryFromIntError> for Value` in `value/conversions.rs`. I'm not sure whether this really is the correct place nor whether the trait should be implemented at all. Explicit conversions for every call to `try_from` seem redundant and clutter the code, but having the conversion available everywhere might be error prone.

Not all casts are possible. There is no `f64 -> isize` nor `f64 -> usize` conversions available OOTB. `num-traits` implements the safe casts, but returns `Option<T>` and not `Result<T, E>`. I wrapped these in [helper functions](https://github.com/boa-dev/boa/pull/919/files#diff-2cb9ec2c93c8ac061e6eeb85d02353c854d2b26246901361aa61646bcbfdb186R1275-R1283) but I'm not sure whether that's really necessary (because the `as` cast is [saturating](https://github.com/rust-lang/rust/issues/10184) and most of the time, this is what we expect).

I think most of the above errors will go away if a new type will be introduced that reflects the result of [`ToIntegerOrInfinity`](https://tc39.es/ecma262/#sec-tointegerorinfinity), i.e. something like:
```rust
enum IntegerOrInfinity {
    Integer(isize),
    NegativeInfinity,
    PositiveInfinity,
}
```
That way we would not need the `f64 -> isize` casts and we would be able to use integers everywhere.

Regarding the `wrapping_*` topic - I think that, most of the time, these should not be used. I can see two cases when these are used:
1) offsetting the index that we ensured is safe (i.e. we won't overflow) - in this case, normal operations would be more appropriate IMO,
2) offsetting the index, but then comparing it with `length` or `0`, e.g. `max(len.wrapping_add(start), 0)` if `start < 0`, when selecting array slice. In this case, `saturating_add` would be correct as this would ignore the fact that we operate on finite types. If we saturate, we would correctly get `0` as the start index (per standard). With wrapping, we end up with `isize::MAX - (start - isize::MAX) = -start`. If we do the same for `end`, we can construct range that points to random region in memory. :)

jakubfijalkowski:
For tests, I implemented the `ToIntegerOrInfinity` idea in a86ea85. I was able (hopefully) to make it safe and eliminate most of the casts. I haven't used it everywhere, but you can find examples in [`Array::fill`](https://github.com/boa-dev/boa/pull/919/commits/a86ea85626d82e86a0935783f7f6ebb8a262ddcc#diff-2cb9ec2c93c8ac061e6eeb85d02353c854d2b26246901361aa61646bcbfdb186R877) and [`Array::slice`](https://github.com/boa-dev/boa/pull/919/commits/a86ea85626d82e86a0935783f7f6ebb8a262ddcc#diff-2cb9ec2c93c8ac061e6eeb85d02353c854d2b26246901361aa61646bcbfdb186R944).

Tell me what do you think and I can either revert this or go and change rest of the methods. :)

boa/src/value/to_integer_or_infinity.rs
```diff
@@ -0,0 +1,99 @@
...
+            IntegerOrInfinity::Integer(i) => (i as usize).min(len),
+
+            // Special case - postive infinity. `len` is always smaller than +inf, thus from (4)
+            IntegerOrInfinity::PositiveInfinity => len,
+        }
+    }
+
+    fn offset(len: usize, i: isize) -> usize {
+        debug_assert!(i < 0);
+        if i == isize::MIN {
```
jakubfijalkowski, line 54: I'm not sure whether the condition here is worth it. Just `len.saturating_sub(i.saturating_neg() as usize)` would work almost the same here. We could also switch the `IntegerOrInfinity::Integer` case from `isize` to `i64` and then all the cases would be covered, since `Number::MAX_SAFE_INTEGER < i64::MAX`. :)

RageKnify (approved):

boa/src/value/to_integer_or_infinity.rs
```diff
@@ -0,0 +1,99 @@
+use super::*;
+
+use std::convert::TryFrom;
+
+/// Represents the result of ToIntegerOrInfinity operation
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum IntegerOrInfinity {
+    Integer(isize),
+    Undefined,
```
HalidOdat, line 9: Why `Undefined`? It can only be (+ or -) infinity or an integer, right?

boa/src/value/to_integer_or_infinity.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl Value {
+    /// Converts argument to an integer, +∞, or -∞.
+    ///
+    /// See: <https://tc39.es/ecma262/#sec-tointegerorinfinity>
+    pub fn to_integer_or_infinity(&self, context: &mut Context) -> Result<IntegerOrInfinity> {
+        // Special case - `undefined`
+        if self.is_undefined() {
+            return Ok(IntegerOrInfinity::Undefined);
+        }
```
HalidOdat, line 70: Why do we treat undefined as a special case in the spec https://tc39.es/ecma262/#sec-tointegerorinfinity the first step is `ToNumber()` call.

boa/src/value/to_integer_or_infinity.rs
```diff
@@ -0,0 +1,99 @@
+use super::*;
+
+use std::convert::TryFrom;
+
+/// Represents the result of ToIntegerOrInfinity operation
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum IntegerOrInfinity {
+    Integer(isize),
```
HalidOdat, line 8: I think `i64` should be used here, since in 32-bit arch. `isize` is 4-bytes, and cannot fit the MAX_SAFE_INTEGER

boa/src/value/mod.rs
```diff
@@ -32,6 +32,7 @@ mod operations;
 mod rcbigint;
 mod rcstring;
 mod rcsymbol;
+mod to_integer_or_infinity;
```
HalidOdat, line 35: I think there is no need to put this in a separate file, like we do with the other `Value::to_*` functions.

boa/src/value/conversions.rs
```diff
@@ -186,3 +186,9 @@ where
         }
     }
 }
+
+impl From<std::num::TryFromIntError> for Value {
+    fn from(err: std::num::TryFromIntError) -> Self {
+        Value::string(format!("{}", err))
+    }
+}
```
HalidOdat, line 194: Why would we need to convert a `TryFromIntError` to a `Value::String`?

HalidOdat (commented):
Looks good!

Check my comments on how to improve this :)

boa/src/value/to_integer_or_infinity.rs
jakubfijalkowski: Not necessarily, [ToNumber](https://tc39.es/ecma262/#sec-tonumber) returns `NaN` in case of `undefined` argument (that is - mostly when it is not specified at all) and it needs to be handled.

boa/src/value/to_integer_or_infinity.rs
jakubfijalkowski: Yeah, I think you're right. This would also simplify some conversions down the road.

boa/src/value/conversions.rs
jakubfijalkowski: `Value` is used to represent errors in most of the array-related functions and, in this PR, I use `i*::try_from(...)?` a lot there. I don't think there is a better way to return this error without adding another `Value` case or switching to non-`Value` errors (or mapping everything manually on each call site).

boa/src/value/to_integer_or_infinity.rs
HalidOdat: > Not necessarily, [ToNumber](https://tc39.es/ecma262/#sec-tonumber) returns `NaN` in case of `undefined` argument (that is - mostly when it is not specified at all) and it needs to be handled.

But in the case that `ToNumber` returns `NaN`, and in step 2 of [`ToIntegerOrInfinity`](https://tc39.es/ecma262/#sec-tointegerorinfinity) it's converted to `0`, so ToIntegerOrInfinity(undefined) returns an integer `0`.
I don't see how it could return undefined, Am I missing something?

jakubfijalkowski: Yeah, you're right. I forgot why I did that in the first place and jumped to conclusion too quickly.

The `Undefined` case has nothing to do with `ToIntegerOrInfinity` - I did add the special case here so that I can add [`as_relative_start` and `as_relative_end`](https://github.com/boa-dev/boa/pull/919/files#diff-907d77dea9f86d13f0ec081a2df726de54ad4e7ccbbb72c4edaf836a12bec639R16-R49) helpers and abstract-out the range calculation in `Array` methods that work on ranges (e.g. [`fill`](https://github.com/boa-dev/boa/pull/919/files#diff-2cb9ec2c93c8ac061e6eeb85d02353c854d2b26246901361aa61646bcbfdb186R882-R891)). Taking `fill` as an example, the algorithm in spec mentions (text in `[]` is my comment)

> Let relativeStart be ? ToIntegerOrInfinity(start). [this makes `undefined` == 0]
> (...)
> If end is undefined, let relativeEnd be len; else let relativeEnd be ? ToIntegerOrInfinity(end). [this makes `undefined` == `len`]

It is used to define the range being worked on. You can see the asymmetry there. If we move the `undefined` handling out of `ToNumber`/`ToIntegerOrInfinity` (i.e. make `ToIntegerOrInfinity` return `undefined` in case of `undefined`) to the call site, we can more or less make them symmetric and have the same calculation used in all of the implementations. That's what I did with the aforementioned helpers.

This might trade the complexity out of `array.rs` into some other module that will use `ToIntegerOrInfinity` in the future, but I think that the other cases (notably - calculating index instead of range) would work with another helper like `as_index`.

boa/src/value/conversions.rs
Razican: Hmmm I don't think we should add this to `Value`. We can use `map_err()` in the error being returned by the integer conversion, and then create a value from it, but wherever this is being used. `Value` should not be aware that we always want to convert an invalid integer conversion to a string, because we might want to convert it to an `Error` at some other point.

jakubfijalkowski:
Thanks for the comments! They really did help me get back on track. Now the changes are (I think) more targeted and don't pollute unrelated code.

To sum up the changes:
1. `IntegerOrInfinity` doesn't have the `Undefined` case to conform to standard better,
2. The `as_relative_*` methods are now in array module (this is the only place where these are used),
3. I removed the `From<TryFromIntError>` implementation from `Value`.

Razican (approved):

Razican:
Let's merge this. I think all concerns were properly treated :)

Issue #736, opened 2020-09-29
Full UTF-16 support


joshwd36:
Something I noticed when working on the string iterator (#704) is that currently strings are stored as ordinary Rust strings, which use UTF-8. However the javascript standard specifies that [strings should use UTF-16](https://tc39.es/ecma262/#sec-ecmascript-language-types-string-type). There are a few places where this difference is noticable. For example, the string `🙂` should have length 2 as it is made up of two code units, but currently shows as having length 1. Similarly, `"🙂".charAt(0)` should return some representation of the first code unit, which has a value of `\ud83d` and cannot be represented as a normal string.

There are a few ways this could be implemented:
- Continue storing strings as Rust strings, and using [`str::encode_utf16()`](https://doc.rust-lang.org/std/primitive.str.html#method.encode_utf16). This has the downside that certain operations, such as getting the length, now have to iterate through the whole string. It also complicates storing individual codepoints, such as `"🙂".charAt(0)`.
- Use the [widestring](https://crates.io/crates/widestring) crate or similar. Strings would therefore be stored as `U16String`, and only converted to rust strings on display.
- Just storing arrays of `u16`s. This probably wouldn't be a good idea as we'd probably end up reimplementing a lot of the functionality of the widestring crate.

joshwd36:
I had a bit of a go at this and it's going to be quite challenging, especially with performance. There are a number of things that are only implemented for `str` or `String`, such as parsing, regex, and son, all of which would need costly conversions between UTF-8 and UTF-16. This would suggest it might be best to just use Rust strings, but then we have the issue of storing invalid strings, which JavaScript supports.

jasonwilliams:
We might need to research how other engines deal with this issue @joshwd36, thanks for looking into it. What was he outcome  widestring crate?

joshwd36:
> 
> 
> What was he outcome widestring crate?

I used a local fork of it that I'd modified to give greater parity with `String` to do the investigation, replacing the inner value of `RcString` with it, and had a look at what errors there were. The main issue is that there is no way to have an `as_str()` method, so every time it uses a function that takes `&str` as an argument it forces an allocation to convert to a `String`. The only way I can see of mitigating that is using/rewriting libraries to use UTF-16 strings, which probably isn't feasible.
>
>
> you may find https://blog.mozilla.org/javascript/2014/07/21/slimmer-and-faster-javascript-strings-in-firefox/ interesting

[...]

PR #935, opened 2020-11-18
Implement String.prototype.codePointAt

### LINKED ISSUE #751: Implement String.prototype.codePointAt
**ECMASCript feature**
Implement `String.prototype.codePointAt`

MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt
Spec: https://tc39.es/ecma262/#sec-string.prototype.codepointat

```javascript
onst icons = '☃★♲';

console.log(icons.codePointAt(1));
// expected output: "9733"
```

It will be similar to our implementation of [`charCodeAt`](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/string/mod.rs#L136-L174)

devinus:
This Pull Request fixes/closes #751.

It changes the following:

- Adds String.prototype.codePointAt
- Adds a few more tests
- Fixes two Clippy linting suggestions

### DIFF
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -22,7 +22,7 @@ use crate::{
 };
 use regress::Regex;
 use std::{
-    char::decode_utf16,
+    char::{decode_utf16, from_u32},
     cmp::{max, min},
     f64::NAN,
     string::String as StdString,
@@ -50,11 +50,11 @@ pub(crate) fn code_point_at(string: RcString, position: i32) -> Option<(u32, u8,
 }
 
 fn is_leading_surrogate(value: u16) -> bool {
-    value >= 0xD800 && value <= 0xDBFF
+    (0xD800..=0xDBFF).contains(&value)
 }
 
 fn is_trailing_surrogate(value: u16) -> bool {
-    value >= 0xDC00 && value <= 0xDFFF
+    (0xDC00..=0xDFFF).contains(&value)
 }
 
 /// JavaScript `String` implementation.
@@ -84,6 +84,7 @@ impl BuiltIn for String {
         .property("length", 0, attribute)
         .method(Self::char_at, "charAt", 1)
         .method(Self::char_code_at, "charCodeAt", 1)
+        .method(Self::code_point_at, "codePointAt", 1)
         .method(Self::to_string, "toString", 0)
         .method(Self::concat, "concat", 1)
         .method(Self::repeat, "repeat", 1)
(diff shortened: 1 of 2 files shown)


boa/src/builtins/string/mod.rs
```diff
@@ -50,11 +50,11 @@ pub(crate) fn code_point_at(string: RcString, position: i32) -> Option<(u32, u8,
...
 
 fn is_leading_surrogate(value: u16) -> bool {
-    value >= 0xD800 && value <= 0xDBFF
+    (0xD800..0xDBFF).contains(&value)
 }
 
 fn is_trailing_surrogate(value: u16) -> bool {
-    value >= 0xDC00 && value <= 0xDFFF
+    (0xDC00..0xDFFF).contains(&value)
 }
```
RageKnify, line 58: Were these the Clippy lints you mention?

boa/src/builtins/string/mod.rs
```diff
@@ -216,6 +220,46 @@ impl String {
         ))
```
RageKnify, line 220: Calculating the length requires iterating over all `char`s and then we do it again to actually get the `nth` char.
This requires 2 iterations rather than just 1 that would user `unwrap_or('')`, right?
I know this isn't what you were implementing, but I noticed that it seems a bit off.

boa/src/builtins/string/mod.rs
```diff
@@ -238,17 +282,21 @@ impl String {
...
+        if pos < 0 || pos >= primitive_val.len() as i32 {
+            return Ok(Value::from(NAN));
+        }
+
+        // Calling .len() on a string would give the wrong result, as they are bytes not the number of unicode code points
+        // Note that this is an O(N) operation (because UTF-8 is complex) while getting the number of bytes is an O(1) operation.
+        let length = primitive_val.chars().count() as i32;
+        if pos >= length {
             return Ok(Value::from(NAN));
         }
```
RageKnify, line 301: Like in `char_at`, this doesn't actually seem like a good idea, we could just match on `primitive_val.encode_utf16().nth(pos as usize)`, right?

RageKnify (approved):
Your changes look good, there was some nearby code that I was a bit confused by that I'd like another mantainer to check on.

boa/src/builtins/string/mod.rs
devinus: @RageKnify Indeed: https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains

boa/src/builtins/string/mod.rs
devinus: @RageKnify No, it's O(1). Rust stores the size and returns that on `len()`.

boa/src/builtins/string/mod.rs
RageKnify: Oh, but I think you reduced the ranges, they included `0xDBFF` and `0xDFFF` before and now they only go up to the numbers before those. I think you should change it to use `=`, for example: `(0xD800..=0xDBFF)`, right?

boa/src/builtins/string/mod.rs
RageKnify: I'm talking about `length`, `let length = primitive_val.chars().count() as i32;`, this is O(N), is it not?

boa/src/builtins/string/mod.rs
devinus: You're right. I misread the warning. (New to Rust, and clippy for that matter.)

boa/src/builtins/string/mod.rs
devinus: Oh, right. I missed that part where you said it wasn't from my changes.

boa/src/builtins/string/mod.rs
devinus: @RageKnify Fixed and added tests around that.

boa/src/builtins/string/mod.rs
devinus: @RageKnify Updated.

devinus:
Just as a heads-up, I tried testing corner cases around unpaired surrogates until I realized that required this to be completed first: https://github.com/boa-dev/boa/issues/736

RageKnify (approved):

Razican (approved):
Thanks! It looks good to me :)

PR #959, opened 2020-12-11
Fix Unicode character escape sequence parsing

### LINKED ISSUE #827: Syntax error when using Unicode escape sequence with curly brackets in a string
**Describe the bug**
The behavior of Unicode code point escape (Unicode escape with curly brackets) is different from the spec.

**To Reproduce**
```
"\u{AA}"       // syntax error, should pass
"\u{{AA}}"     // pass, should throw syntax error
"\u{{AA}"      // freeze, should throw syntax error
"\u{AA}}"      // freeze, should pass
```

**Expected behavior**
The string should be tokenized and handled correctly. The expected behavior can be found in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#Escape_notation).
Some lexer tests should also be updated.

- OS: `Windows 10`
- Version: `OS Build 190941.508`
- Rustc version: `rustc 1.49.0-nightly (91a79fb29 2020-10-07)`

**Additional context**
It should be bugs in [`string.rs`](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/lexer/string.rs#L125) and [`cursor.rs`](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/lexer/cursor.rs).

tofpie:
This Pull Request fixes #827.

It changes the following:

- Parsing of Unicode escape sequences
- Lexer tests

### DIFF
--- a/boa/src/syntax/lexer/string.rs
+++ b/boa/src/syntax/lexer/string.rs
@@ -111,13 +111,10 @@ impl<R> Tokenizer<R> for StringLiteral {
                             b'u' => {
                                 // Support \u{X..X} (Unicode Codepoint)
                                 if cursor.next_is(b'{')? {
-                                    cursor.next_byte()?.expect("{ character vanished"); // Consume the '{'.
 
                                     // TODO: use bytes for a bit better performance (using stack)
                                     let mut code_point_buf = Vec::with_capacity(6);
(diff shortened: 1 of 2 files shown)


RageKnify (approved):

Razican (approved):
Thanks for the fix! Looks very good to me, let's merge :)

PR #960, opened 2020-12-11
Fix unresolved links in documentation

### LINKED ISSUE #957: Unresolved links in documentation
We seem to have unresolved links in the documentation, that are generating warnings in our CI:
https://github.com/boa-dev/boa/actions/runs/408553547

tofpie:
This Pull Request fixes #957.

It changes the following:
- Unresolved or misinterpreted links in the documentation

### DIFF
--- a/boa/src/syntax/ast/op.rs
+++ b/boa/src/syntax/ast/op.rs
@@ -677,11 +677,11 @@ pub enum LogOp {
     /// Syntax: `x || y`
     ///
     /// More information:
-    ///  - [ECMAScript reference](
+    ///  - [ECMAScript reference][spec]
     ///  - [MDN documentation][mdn]
     ///
-    /// [spec]: <https://tc39.es/ecma262/#prod-LogicalORExpression)>
-    /// [mdn]: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Logical_OR>
+    /// [spec]: https://tc39.es/ecma262/#prod-LogicalORExpression
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Logical_OR
     Or,
 }
 
--- a/boa/src/syntax/ast/op.rs
+++ b/boa/src/syntax/ast/op.rs
@@ -680,8 +680,8 @@ pub enum LogOp {
     ///  - [ECMAScript reference](
     ///  - [MDN documentation][mdn]
     ///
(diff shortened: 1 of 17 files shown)


boa/src/syntax/ast/op.rs
```diff
@@ -680,8 +680,8 @@ pub enum LogOp {
     ///  - [ECMAScript reference](
     ///  - [MDN documentation][mdn]
     ///
-    /// [spec]: https://tc39.es/ecma262/#prod-LogicalORExpression)
-    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Logical_OR
+    /// [spec]: <https://tc39.es/ecma262/#prod-LogicalORExpression)>
+    /// [mdn]: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Logical_OR>
```
Razican, line 684: I believe this is not correct, right? It should show `[spec]` next to `[ECMAScript reference]`, and then, the links should not have `<...>`. Also, the `spec` link should not have a closing `)`

boa/src/value/mod.rs
```diff
@@ -247,8 +247,8 @@ impl Value {
     /// For scalar types it should be false, for objects check the private field for extensibilaty.
     /// By default true.
     ///
-    /// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/seal would turn extensible to false/>
-    /// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze would also turn extensible to false/>
+    /// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/seal> would turn extensible to false
+    /// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze> would also turn extensible to false
```
Razican, line 251: I believe it's best to treat variables and variables as code, now that we are changing this.

```suggestion
    /// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/seal> would turn `extensible` to `false`
    /// <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze> would also turn `extensible` to `false`
```

Razican (commented):
Thanks for the contribution! This cleans up a lot of stuff in the docs :)

I only noticed two minor things that could be improved.

Razican (approved):
Thanks! This looks good to me :)

RageKnify (approved):

PR #961, opened 2020-12-12
Implement the optional `space` parameter in `JSON.stringify` 

### LINKED ISSUE #346: Implement the optional `space` parameter in `JSON.stringify( value[, replacer [, space] ] )`
The `JSON.stringify(value[, replacer[, space]])` takes in three arguments.

The `space` argument is a `String` or `Number` object that's used to insert white space into the output `JSON` string for readability purposes.

If this is a `Number`, it indicates the number of space characters to use as white space; this number is capped at `10` (if it is greater, the value is just `10`). Values less than `1` indicate that no space should be used.

If this is a `String`, the string (or the first `10` characters of the string, if it's longer than that) is used as white space. If this parameter is not provided (or is `null`), no white space is used.

More information:
 - [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)
 - [ECMAScript reference](https://tc39.es/ecma262/#sec-json.stringify)

tofpie:
This Pull Request fixes/closes #346 .

It changes the following:

- `JSON.stringify()` now supports `space` parameter

### DIFF
--- a/boa/src/builtins/json/tests.rs
+++ b/boa/src/builtins/json/tests.rs
@@ -208,6 +208,102 @@ fn json_stringify_fractional_numbers() {
 }
 
 #[test]
+fn json_stringify_pretty_print() {
+    let mut context = Context::new();
+
+    let actual = forward(&mut context, r#"JSON.stringify({a: "b", b: "c"}, {}, 4)"#);
+    let expected = forward(
+        &mut context,
+        r#"'{
+    "a": "b",
+    "b": "c"
+}'"#,
+    );
+    assert_eq!(actual, expected);
+}
+
+#[test]
+fn json_stringify_pretty_print_four_spaces() {
+    let mut context = Context::new();
+
+    let actual = forward(&mut context, r#"JSON.stringify({a: "b", b: "c"}, {}, 4.3)"#);
+    let expected = forward(
+        &mut context,
+        r#"'{
+    "a": "b",
+    "b": "c"
+}'"#,
+    );
+    assert_eq!(actual, expected);
+}
+
+#[test]
+fn json_stringify_pretty_print_twenty_spaces() {
+    let mut context = Context::new();
+
+    let actual = forward(
+        &mut context,
+        r#"JSON.stringify({a: "b", b: "c"}, ["a", "b"], 20)"#,
+    );
+    let expected = forward(
+        &mut context,
+        r#"'{
+          "a": "b",
+          "b": "c"
+}'"#,
+    );
+    assert_eq!(actual, expected);
+}
+
+#[test]
+fn json_stringify_pretty_print_bad_space_argument() {
+    let mut context = Context::new();
+
+    let actual = forward(
+        &mut context,
+        r#"JSON.stringify({a: "b", b: "c"}, ["a", "b"], [])"#,
+    );
+    let expected = forward(&mut context, r#"'{"a":"b","b":"c"}'"#);
+    assert_eq!(actual, expected);
+}
+
+#[test]
+fn json_stringify_pretty_print_with_string() {
+    let mut context = Context::new();
+
+    let actual = forward(
+        &mut context,
+        r#"JSON.stringify({a: "b", b: "c"}, {}, "abcd")"#,
+    );
+    let expected = forward(
+        &mut context,
+        r#"'{
+abcd"a": "b",
+abcd"b": "c"
+}'"#,
+    );
+    assert_eq!(actual, expected);
+}
+
+#[test]
+fn json_stringify_pretty_print_with_too_long_string() {
+    let mut context = Context::new();
+
+    let actual = forward(
+        &mut context,
+        r#"JSON.stringify({a: "b", b: "c"}, {}, "abcdefghijklmn")"#,
+    );
+    let expected = forward(
+        &mut context,
+        r#"'{
+abcdefghij"a": "b",
+abcdefghij"b": "c"
+}'"#,
+    );
+    assert_eq!(actual, expected);
+}
+
+#[test]
 fn json_parse_array_with_reviver() {
     let mut context = Context::new();
     let result = forward_val(
--- a/boa/src/builtins/json/tests.rs
+++ b/boa/src/builtins/json/tests.rs
@@ -211,7 +211,10 @@ fn json_stringify_fractional_numbers() {
 fn json_stringify_pretty_print() {
     let mut context = Context::new();
 
-    let actual = forward(&mut context, r#"JSON.stringify({a: "b", b: "c"}, undefined, 4)"#);
+    let actual = forward(
+        &mut context,
+        r#"JSON.stringify({a: "b", b: "c"}, undefined, 4)"#,
+    );
     let expected = forward(
         &mut context,
         r#"'{
--- a/boa/src/builtins/json/tests.rs
+++ b/boa/src/builtins/json/tests.rs
@@ -262,6 +262,24 @@ fn json_stringify_pretty_print_twenty_spaces() {
 }
 
 #[test]
+fn json_stringify_pretty_print_with_number_object() {
+    let mut context = Context::new();
+
+    let actual = forward(
+        &mut context,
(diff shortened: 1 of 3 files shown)


RageKnify (changes requested):
Looks good, but needs a fix:
`JSON.stringify({a: "b", b: "c"}, undefined, 4)` will return `{"a":"b","b":"c"}` when it should not.
The code relative to `replacer` needs to not return immediately, the implementation doesn't follow the spec closely, so I'm not sure as to how we should aproach, but it's a test case to add (I'd say to change your tests with `{}` as the replacer to use `undefined` instead).

#919 implements `ToIntegerOrInfinity`, which could be used to more closely follow the spec, but I don't think it's a big problem.

tofpie:
Thanks for your review and comments. Let's wait for #919 to be merged, so that `ToIntegerOrInfinity` can be used to follow the spec more closely.

boa/src/builtins/json/tests.rs
```diff
@@ -207,6 +207,105 @@ fn json_stringify_fractional_numbers() {
     assert_eq!(actual, expected);
 }
 
+#[test]
+fn json_stringify_pretty_print() {
+    let mut context = Context::new();
+
+    let actual = forward(&mut context, r#"JSON.stringify({a: "b", b: "c"}, {}, 4)"#);
```
RageKnify, line 214:
```suggestion
    let actual = forward(&mut context, r#"JSON.stringify({a: "b", b: "c"}, undefined, 4)"#);
```

boa/src/builtins/json/tests.rs
```diff
@@ -207,6 +207,105 @@ fn json_stringify_fractional_numbers() {
...
+    assert_eq!(actual, expected);
+}
+
+#[test]
+fn json_stringify_pretty_print_with_string() {
+    let mut context = Context::new();
+
+    let actual = forward(
+        &mut context,
+        r#"JSON.stringify({a: "b", b: "c"}, {}, "abcd")"#,
```
RageKnify, line 279:
```suggestion
        r#"JSON.stringify({a: "b", b: "c"}, undefined, "abcd")"#,
```

boa/src/builtins/json/tests.rs
```diff
@@ -207,6 +207,105 @@ fn json_stringify_fractional_numbers() {
...
+    assert_eq!(actual, expected);
+}
+
+#[test]
+fn json_stringify_pretty_print_with_too_long_string() {
+    let mut context = Context::new();
+
+    let actual = forward(
+        &mut context,
+        r#"JSON.stringify({a: "b", b: "c"}, {}, "abcdefghijklmn")"#,
```
RageKnify, line 297:
```suggestion
        r#"JSON.stringify({a: "b", b: "c"}, undefined, "abcdefghijklmn")"#,
```

RageKnify (changes requested):
Sorry to bother, but I do think we should use `undefined` in all of these cases, the spec only uses the 2nd argument when it is a function or array, I'd rather we leave it as `undefined` to be clear that it is not used.

RageKnify (approved):
Looks good, we'll just wait for #919 to use `ToIntegerOrInfinity`

tofpie:
I have run the test [space-string-object.js](https://github.com/tc39/test262/blob/3b425c6da3312d071a3ee00111039fce1774187e/test/built-ins/JSON/stringify/space-string-object.js). It passes on this branch but it also passes on `master`. I have used the following command:
```sh
cargo run --release --bin boa_tester -- run -s test/built-ins/JSON/stringify/space-string-object.js -vv
```
It is probably a dumb question, and I might have missed something, but why is it pass on `master`?

RageKnify:
That test only checks if pairs of invocations return the same result, it never checks if the results follow the spec. So we get:
```js
>> JSON.stringify(obj, null, 'xxx')
"{"a1":{"b1":[1,2,3,4],"b2":{"c1":1,"c2":2}},"a2":"a2"}"
>> JSON.stringify(obj, null, new String('xxx'))
"{"a1":{"b1":[1,2,3,4],"b2":{"c1":1,"c2":2}},"a2":"a2"}"
```
Being equal and we pass the test.

tofpie:
Thanks for the explanation. I have run [space-number.js](https://github.com/tc39/test262/blob/3b425c6da3312d071a3ee00111039fce1774187e/test/built-ins/JSON/stringify/space-number.js) and also [space-string.js](https://github.com/tc39/test262/blob/3b425c6da3312d071a3ee00111039fce1774187e/test/built-ins/JSON/stringify/space-string.js) on `master`, and they are both passed.

RageKnify:
> I have run [space-number.js](https://github.com/tc39/test262/blob/3b425c6da3312d071a3ee00111039fce1774187e/test/built-ins/JSON/stringify/space-number.js) and also [space-string.js](https://github.com/tc39/test262/blob/3b425c6da3312d071a3ee00111039fce1774187e/test/built-ins/JSON/stringify/space-string.js) on `master`, and they are both passed.

My bad for not reading the test carefully, just like the other one, it only checks if the results match, since we never use the arguments it passes the test.

tofpie:
It is  right for [space-number.js](https://github.com/tc39/test262/blob/3b425c6da3312d071a3ee00111039fce1774187e/test/built-ins/JSON/stringify/space-number.js), but [space-string.js](https://github.com/tc39/test262/blob/3b425c6da3312d071a3ee00111039fce1774187e/test/built-ins/JSON/stringify/space-string.js) makes an actual comparison with the expected indented string in the second assert. It should probably be failed on `master`.

tofpie:
I finally found why the tests are passed on `master` on my computer. It seems that there is a bug in the parsing of single line comments that occurs when the files use CRLF line endings. The end of line is not detected, so the single line comment extends until the end of the file.

In `cursor.rs` when a CR is detected, the following LF is swallowed, and only the CR is returned.
https://github.com/boa-dev/boa/blob/0c068cb35bfedb595cdad460b1380ae382807dde/boa/src/syntax/lexer/cursor.rs#L222-L230

But the tokenizer for single line comments expects a LF to end a comment, and not a CR.
https://github.com/boa-dev/boa/blob/0c068cb35bfedb595cdad460b1380ae382807dde/boa/src/syntax/lexer/comment.rs#L33-L40


As I work on Windows, and the test262 repository is cloned with CRLF endings on Windows, so all tests starting with single line comments are passed, as they are considered empty.

Fixing `cursor.rs` so that it always returns LF for end of lines would probably be the best solution, but it is also possible to fix the single line comment parser so that it detects both CR and LF as line endings. What do you think?

Razican:
> Fixing `cursor.rs` so that it always returns LF for end of lines would probably be the best solution, but it is also possible to fix the single line comment parser so that it detects both CR and LF as line endings. What do you think?

I have merged tha PR, so a rebase should fix that.

  | Passed | 18,964 | 18,963 | **-1** |
  | Failed | 43,937 | 43,938 | **+1** |
  | Panics | 1,127 | 1,128 | **+1** |
  | Conformance | 24.17 | 24.17 | **-0.00%** |

Razican (approved):
This looks good to me, It seems it follows the spec correctly, even though it's not following it step by step exactly. Still, I think it makes sense to do it this way, since the result is the same and it might be a bit more optimized.

tofpie:
In fact, using `to_integer_or_infinity` would convert a boolean to 1, and so generate an indentation with one space, and that does not follow the spec. Maybe it should stay as it is.

RageKnify:
> In fact, using `to_integer_or_infinity` would convert a boolean to 1, and so generate an indentation with one space, and that does not follow the spec. Maybe it should stay as it is.

I don't think that would be the case @tofpie the spec only enters step 6. if the type is `Number` and that wouldn't be the case for `boolean`s.

tofpie:
Thanks for the clarification. I understand now. Sorry I did not get it earlier.
I will propose something that follows the spec more closely.

RageKnify (approved):
LGTM, I'll let @Razican have a final look at the code

Razican:
Looks good to me :) thanks!

PR #962, opened 2020-12-12
Implement Number.prototype.toPrecision

### LINKED ISSUE #349: Implement `Number.prototype.toPrecision`
The `toPrecision()` method returns a string representing the `Number` object to the specified precision.

More information:
 - [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision)
 - [ECMAScript reference](https://tc39.es/ecma262/#sec-number.prototype.toprecision)

NathanRoyer:
This Pull Request closes #349.

It changes the following:

- Implementation of Number.prototype.toPrecision is complete.
- Tests of the aforementioned method were enabled.

### DIFF
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -364,9 +364,9 @@ impl Number {
         let precision_i32 = precision as i32;
 
         // 7
-        let mut prefix = String::new(); // spec: "s"
-        let mut suffix: String; // spec: "m"
-        let exponent: i32; // spec: "e"
+        let mut prefix = String::new(); // spec: 's'
+        let mut suffix: String; // spec: 'm'
+        let exponent: i32; // spec: 'e'
 
         // 8
         if this_num < 0.0 {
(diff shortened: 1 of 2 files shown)


Razican:
  | Passed | 18,964 | 18,976 | **+12** |
  | Failed | 43,937 | 43,925 | **-12** |
  | Panics | 1,127 | 1,109 | **-18** |
  | Conformance | 24.17 | 24.19 | **+0.02%** |

RageKnify:
The doc comment for `to_precision` has the wrong link to the spec, it should be https://tc39.es/ecma262/#sec-number.prototype.toprecision , can you fix it please @NathanRoyer?

NathanRoyer:
@RageKnify Done; the last commit should correct the spec and mdn links :)

RageKnify (changes requested):
Hey @NathanRoyer having comments with `// 1`, `// 2`, etc matching the spec's steps helps a lot in understanding the code, these are long functions and having each step identified makes it easy to check the "Rust translation", should be easy for you to do and it helps everyone who will one day read the function to understand it.
One thing I think can be changed to more closely follow the spec is in step 3 to use: https://github.com/boa-dev/boa/blob/8f388d501d0aed14914873ea01f11426b718cd43/boa/src/value/mod.rs#L878
That was recently implemeted and should make the first `if` a bit clearer.

You can also use the following code to get `precision` as a `Value` without needing to use `unwrap`, you will get `Value::undefined` if no argument was given.
```rust
let precision = args.get(0).cloned().unwrap_or_default();
```

And for the links you changed the wrong ones :sweat_smile: , you changed lines 13-14 but the method's links are in lines 321-322, but don't worry just place the old links back and replaced the method's.

boa/src/builtins/number/mod.rs
```diff
@@ -318,45 +318,53 @@ impl Number {
...
-        if precision_arg.is_none() || !this_num.is_finite() {
+        // 2 & 4
+        if precision_var == Value::undefined() || !this_num.is_finite() {
             return Self::to_string(this, &[], context);
         }
 
-        let precision = match precision_arg.unwrap().to_integer(context)? as i32 {
-            x @ 1..=100 => x as usize,
+        // 3
+        let precision = match precision_var.to_integer_or_infinity(context).unwrap() {
```
RageKnify, line 339: Replace the `.unwrap()` with `?`, this way it matches the spec's `? ToIntegerOrInfinity(precision)`

As an example `(3).toPrecision(Symbol.match)` will result in a panic and should instead throw the error.

boa/src/builtins/number/mod.rs
```diff
@@ -277,25 +318,99 @@ impl Number {
...
+                suffix.truncate(precision);
+                if precision > 1 {
+                    suffix.insert(1, '.');
+                }
+                suffix.push('e');
+                if great_exp {
+                    suffix.push('+');
+                }
+                suffix.push_str(&exponent.to_string());
+            }
```
RageKnify, line 392: It would be nice to have `// a`, `// b` and `// c` and inside the `if` to also include `ii`, `iii` and the rest, just like the rest of the function.

boa/src/builtins/number/mod.rs
```diff
@@ -277,25 +318,99 @@ impl Number {
...
+                if precision > e_inc as usize {
+                    suffix.insert(e_inc as usize, '.');
+                }
+            // 13
+            } else {
+                prefix.push('0');
+                prefix.push('.');
+                prefix.push_str(&"0".repeat(-e_inc as usize));
+            }
+        }
```
RageKnify, line 410: As a last thing I'd preffer if you followed the control flow from the spec, 11 is a separate `if` in the spec with an early return.

RageKnify (changes requested):
It's looking good, just a few more changes to make it easier to read it along with the spec.

RageKnify (approved):

Razican (approved):

NathanRoyer:
@Razican Hey, I just found 2 bugs earlier today, and I was discussing them on the discord forum... is it still possible to omdify the code ?

PR #964, opened 2020-12-13
Fix single line comment lexing with CRLF line ending

### LINKED ISSUE #963: Incorret lexing of single line comments
Single line comments in files using CRLF line endings result in the rest of the file being ignored.

**To Reproduce**
Since a file is required the best way to demonstrate the bug is to run with the following file: `cargo run -- test.txt` [test.txt](https://github.com/boa-dev/boa/files/5683971/test.txt)
The file contains:
```
console.log('0')
// .
console.log('1')
```

**Expected behavior**
`boa` should output:
```
0
1
undefined
```
but instead we do not get the `1`

This is probably what caused us to not see an increase in the number of passing test262 tests, a lot of them are not being lexed correctly.

Identified by @tofpie in https://github.com/boa-dev/boa/pull/961#issuecomment-743921418

tofpie:
This Pull Request fixes #963.

It changes the following:

- Lexing of single line comment

### DIFF
--- a/boa/src/syntax/lexer/comment.rs
+++ b/boa/src/syntax/lexer/comment.rs
@@ -31,11 +31,11 @@ impl<R> Tokenizer<R> for SingleLineComment {
 
         // Skip either to the end of the line or to the end of the input
         while let Some(ch) = cursor.peek()? {
-            if ch == b'\n' {
+            if ch == b'\n' || ch == b'\r' {
                 break;
             } else {
                 // Consume char.
-                cursor.next_byte()?.expect("Comment character vansihed");
+                cursor.next_byte()?.expect("Comment character vanished");
             }
         }
         Ok(Token::new(
--- a/boa/src/syntax/lexer/tests.rs
+++ b/boa/src/syntax/lexer/tests.rs
@@ -42,6 +42,21 @@ fn check_single_line_comment() {
 }
 
 #[test]
+fn check_single_line_comment_with_crlf_ending() {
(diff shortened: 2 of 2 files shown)


Lan2u:
Do you also need to update the multi-line comment code? 

I think your first idea of making the lexer always return LF for the end of a line would be the best course as it makes lexing (very slightly) more simple (relevant if in future more tokens are added).

tofpie:
I tried, but the lexer for single line comment uses `peek()` and not `next_byte()`, so it did not work.

https://github.com/boa-dev/boa/blob/0c068cb35bfedb595cdad460b1380ae382807dde/boa/src/syntax/lexer/comment.rs#L33-L40

I did not see any issues in the lexer for multi-line comment.

Razican (approved):
This looks good :) thanks

Razican:
Hmmm surprisingly, this made 6 tests fail, but maybe that's correct.

  | Passed | 18,964 | 18,958 | **-6** |
  | Failed | 43,937 | 43,943 | **+6** |
  | Conformance | 24.17 | 24.16 | **-0.01%** |

tofpie:
I have run the tests before and after the fix on Linux using WSL, and I have checked some of the newly failed tests. In fact, it seems that even when using Linux, some files from the test262 repository contains CRLF line endings. So, tests that were previously passed because they were considered empty are now failed because they contain not yet supported features (such as eval).
On the other side, some tests that explicitly test the single line comment with CRLF, and are now passed.

PR #974, opened 2020-12-16
Fix enumerable attribute on string length property

### LINKED ISSUE #973: The `length` property of a string is enumerable
**Describe the bug**
The `length` property of a string is enumerable, but it should not.

**To Reproduce**
```javascript
>> "".propertyIsEnumerable("length")
true
>> new String("").propertyIsEnumerable("length")
true
```

**Expected behavior**
`"".propertyIsEnumerable("length")` and `new String("").propertyIsEnumerable("length")` should be `false`.

**Build environment**

- OS: Windows 10
- Version: 0.10
- Rustc version: rustc 1.48.0 (7eac88abb 2020-11-16)

**Additional context**
None

tofpie:
This Pull Request fixes #973.

It changes the following:

-    The length property of a string is set to non enumerable

### DIFF
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -13,6 +13,7 @@ pub mod string_iterator;
 #[cfg(test)]
 mod tests;
 
+use crate::property::DataDescriptor;
 use crate::{
     builtins::{string::string_iterator::StringIterator, BuiltIn, RegExp},
     object::{ConstructorBuilder, Object, ObjectData},
@@ -140,9 +141,11 @@ impl String {
             None => RcString::default(),
         };
 
-        let length = string.encode_utf16().count();
-
-        this.set_field("length", Value::from(length as i32));
+        let length = DataDescriptor::new(
+            Value::from(string.encode_utf16().count() as i32),
+            Attribute::NON_ENUMERABLE,
+        );
+        this.set_property("length", length);
 
         this.set_data(ObjectData::String(string.clone()));
 
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -142,7 +142,7 @@ impl String {
         };
 
         let length = DataDescriptor::new(
-            Value::from(string.encode_utf16().count() as i32),
+            Value::from(string.encode_utf16().count()),
             Attribute::NON_ENUMERABLE,
         );
         this.set_property("length", length);
--- a/boa/src/value/mod.rs
+++ b/boa/src/value/mod.rs
@@ -664,7 +664,11 @@ impl Value {
                     ObjectData::String(string.clone()),
                 ));
                 // Make sure the correct length is set on our new string object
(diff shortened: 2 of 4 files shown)


boa/src/builtins/string/mod.rs
```diff
@@ -140,9 +141,11 @@ impl String {
             None => RcString::default(),
         };
 
-        let length = string.encode_utf16().count();
-
-        this.set_field("length", Value::from(length as i32));
+        let length = DataDescriptor::new(
+            Value::from(string.encode_utf16().count() as i32),
```
RageKnify, line 145: I know it was already there, but I think we should remove the `as i32`, `From<usize>` is implemeted for `Value` so the conversion is unnecessary.
```suggestion
            Value::from(string.encode_utf16().count()),
```

RageKnify (changes requested):
I think we could remove the unnecessary conversion since we're touching these lines.

Razican:
This looks good :)

## Test262 conformance changes:
  | Passed | 19,374 | 19,380 | **+6** |
  | Failed | 43,527 | 43,521 | **-6** |
  | Panics | 465 | 465 | 0 |
  | Conformance | 24.69 | 24.70 | **+0.01%** |

Razican (approved):

boa/src/value/mod.rs
```diff
@@ -664,7 +664,11 @@ impl Value {
                     ObjectData::String(string.clone()),
                 ));
                 // Make sure the correct length is set on our new string object
-                object.set("length".into(), string.chars().count().into());
+                object.insert_property(
+                    PropertyKey::String("length".into()),
+                    Value::from(string.chars().count()),
```
RageKnify, line 669: Just now noticed, this doesn't seem to match the `length` calculation in `boa/src/builtins/string/mod.rs`, there we use `encode_utf16().count()`, so I assume this one might give different results. I'd like someone else to back me on this though.

RageKnify: @Razican can you confirm this?

tofpie: Seems that `chars().count()` counts the code points, and `encode_utf16().count()` counts the UTF-16 code units, and I think that counting UTF-16 code units is what JavaScript does, but I am not quite sure.

Razican: As per [this](https://tc39.es/ecma262/#sec-terms-and-definitions-string-value), the inner value of a String is an array of `u16` code points (which we currently don't use, but at some point we should).

Then, as per [this](https://tc39.es/ecma262/#sec-properties-of-string-instances-length), the `length` property is the number of values (`u16` code points). So it should use `encode_utf16().count()`.

RageKnify: That's what I thought, thanks. Just change the method invoked and it should be good to `merge` @tofpie

Razican:
This is blocked on #977.

RageKnify (approved):

Razican:
#977 was merged, could you rebase?

PR #972, opened 2020-12-16
Fix enumerable attribute on array length property

tofpie:
This Pull Request fixes #971 .

It changes the following:

- The `length` property of an array is set to non enumerable

PR #976, opened 2020-12-16
Implement for..in

### LINKED ISSUE #670: Implement `for..in` for objects
**ECMASCript feature**
I would like to see `for..in` syntax and execution implemented. [ECMAScript specification][spec]. This is important for test262, as it uses `for..in` pretty extensively.

**Example code**
```js
const myObject = {
 a: 'test1',
 b: 'test2'
};
let c = '';
for(const prop in myObject) {
 c += prop + '=' + myObject[prop] + ';';
}
c
```

The expected output is `a=test1;b=test2;`.

[spec]: https://www.ecma-international.org/ecma-262/10.0/index.html#sec-for-in-and-for-of-statements

tofpie:
This Pull Request closes #670.

It changes the following:

- The tokenizer to prevent interpreting `in` as a binary operator in a `for` statement
- The parser to add `for..in` management
- The `ForOfLoop` node (renamed `ForInOfLoop`) to execute the `for..in` logic


tofpie:
This is really a WIP and I plan to add unit tests as soon as #974 and #972 are merged, but I submit it anyway to have some guidance. I am not quite sure of what I have done, especially around the lexer and parser.

@jcdickinson If you have some time, I would be happy to have your feedback

boa/src/syntax/ast/node/iteration/for_in_of_loop/mod.rs
```diff
@@ -8,68 +9,103 @@ use crate::{
...
 
+use crate::builtins::iterable::IteratorRecord;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Clone, Debug, Trace, Finalize, PartialEq)]
-pub struct ForOfLoop {
+pub enum IterationKind {
+    ITERATE,
```
jcdickinson, line 19: Nit: Rust enums are not ALL_CAPS.

tofpie: Old habits...

boa/src/builtins/object/tests.rs
```diff
@@ -279,3 +279,25 @@ fn object_define_properties() {
...
 }
+
+#[test]
+fn for_in_iterator_object() {
+    let mut context = Context::new();
+
+    let init = r#"
+        let result = [];
+        let obj = { a: "a", b: 2};
+        for (var i in obj) {
```
jcdickinson, line 290: Adds tests for const and let possibly.

jcdickinson: Also labelled breaks.

jcdickinson:
This looks good to me, but I haven't touched the codebase in a while so I'm out of touch with the new parser.

Razican:
I didn't check the code yet, but this is looking very good:

  | Passed | 19,374 | 20,409 | **+1,035** |
  | Failed | 43,527 | 42,492 | **-1,035** |
  | Panics | 465 | 477 | **+12** |
  | Conformance | 24.69 | 26.01 | **+1.32%** |

tofpie:
I have implemented break and continue in for..in, and I have added some unit tests. It is now ready for review.

boa/src/syntax/parser/error.rs
```diff
@@ -98,6 +98,7 @@ impl ParseError {
     }
 
     /// Creates a new `Unimplemented` parsing error.
+    #[allow(dead_code)]
```
Razican, line 101: Hmmm interesting, don't we have any unimplemented parsing constructs?

boa/src/syntax/ast/node/iteration/for_in_of_loop/mod.rs
```diff
@@ -8,68 +9,113 @@ use crate::{
...
 #[cfg(feature = "deser")]
 use serde::{Deserialize, Serialize};
 
 #[cfg_attr(feature = "deser", derive(Serialize, Deserialize))]
 #[derive(Clone, Debug, Trace, Finalize, PartialEq)]
-pub struct ForOfLoop {
+pub enum IterationKind {
+    Iterate,
+    Enumerate,
+}
```
Razican, line 21: This should implement `Copy`, for efficiency reasons, but that means it cannot derive `Trace`. It can implement it itself, though, using:

```rust
use crate::gc::empty_trace;

unsafe impl Trace for IterationKind {
    empty_trace!();
}
```

boa/src/syntax/ast/node/iteration/for_in_of_loop/mod.rs
```diff
@@ -8,68 +9,113 @@ use crate::{
...
+
+#[cfg_attr(feature = "deser", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug, Trace, Finalize, PartialEq)]
+pub struct ForInOfLoop {
```
Razican, line 25: I understand that the logic is very similar for `for...in` and `for...of`, but it seems that this requires some extra fields that will not be space efficient, right?

Razican (commented):
It's looking good. I did a small review, but I need to spend a bit more time with it. I'm not sure about unifying `for...in` and `for...of`, mostly because that will probably not be space efficient.

boa/src/syntax/parser/error.rs
tofpie: Certainly, but the only call to this function was in `ForInLoop`. Maybe it could be generalized?

boa/src/syntax/ast/node/iteration/for_in_of_loop/mod.rs
tofpie: My initial thought was also to add a new node, but then I read the spec and both `for..in` and `for..of` are described in the same sections. It could make things easier when adding `for await..in` and `for await..of` as they are also described in the same sections. Nevertheless, I understand this is less space efficient, and that we can add a new node for `for..in`.

Razican: Yes, even if they are described in the same section, I think it's best if we add a different node for space efficiency reasons :) As long as we follow the spec in functionality, we don't need to have the code organised exactly in the same way.

boa/src/syntax/parser/error.rs
Razican: Yes, I think it's good like this for now, at some point we should review all the unimplemented syntax and add calls to this in those places.

boa/src/syntax/ast/node/iteration/for_in_of_loop/mod.rs
tofpie: I have split `ForInOfLoop`.

Razican (approved):
This is impressive. The conformance improvement is vastly good, and I have checked the implementation, and I see improvements in many other loops. Thanks for implementing the changes I proposed!

HalidOdat (approved):

RageKnify (approved):

PR #979, opened 2020-12-18
Read file input in bytes instead of string

### LINKED ISSUE #876: Handle invalid UTF-8 chars in user input instead of panic
**Describe the bug**
When the user input has invalid UTF-8 chars, boa always panics with InvalidData error "stream did not contain valid UTF-8", no matter the invalid char is in a Unicode escape sequence, normal string char, identifier, or even comment. It seems to be a lexer-wide issue.

**To Reproduce**
The following example should work since the invalid UTF-8 chars are in the comment:
```javascript
// ����
```
Besides, the following example should throw syntax error instead of panic:
```javascript
let x� = 100;
```

**Expected behavior**
Need some help on finding the corresponding spec.
```javascript
// ����, should pass
'����' // should pass, the string stores invalid chars as bytes
let ���� = 10 // should throw syntax error
```

- OS: `Windows 10`
- Version: `OS Build 190941.508`
- Rustc version: `rustc 1.49.0-nightly (91a79fb29 2020-10-07)`

**Additional context**
Here is a sample code to generate JavaScript code with invalid UTF-8 characters. Copy and paste the sample above to boa does not work in Win10 since the invalid chars are replaced by replacement chars.
```javascript
const fs = require('fs');

const invalidTextBuf = Buffer.from('🐶🐶🐶', 'utf8').slice(1, 5);
const template= 'let _ = 100;';

const invalidjsBuf = template.split('_')
    .map(part => Buffer.from(part, 'utf8'))
    .reduce((buf, partBuf, i, array) => {
        buf = Buffer.concat([buf, partBuf]);
        if (i !== array.length - 1) {
            buf = Buffer.concat([buf, invalidTextBuf]);
        }
        return buf;
    }, Buffer.alloc(0));

fs.writeFileSync('./invalid.js', invalidjsBuf);
```

jevancc:
This Pull Request is a follow-up PR for #915, which closes #876.

This PR may also improve run-time performance slightly since it removes the unnecessary bytes to utf8 chars conversion when reading the files.

It changes the following:
- Read the file input in bytes (`boa_cli`)

Except for file input in `boa_cli`, all the other places still take input in str and convert it to bytes afterward:
#### `boa_cli` command prompt
The crate `rustyline` reads input in strings and takes care of it.
#### `boa_wasm`
`wasm_bindgen` makes JS interface and takes input in strings.
#### test/benchmark
The benchmark still uses `include_str!` to read the bench scripts. Since the files are read in compile-time, it does not affect the run-time performance. Keep it as it can also benefit because it checks if there are any invalid chars in the bench scripts in compile time.
Similar reasons for tests. Taking the input in str makes the tests simpler and clearer.

### DIFF
--- a/boa/src/context.rs
+++ b/boa/src/context.rs
@@ -677,10 +677,11 @@ impl Context {
     /// ```
     #[allow(clippy::unit_arg, clippy::drop_copy)]
     #[inline]
-    pub fn eval(&mut self, src: &str) -> Result<Value> {
+    pub fn eval<T: AsRef<[u8]>>(&mut self, src: T) -> Result<Value> {
         let main_timer = BoaProfiler::global().start_event("Main", "Main");
+        let src_bytes: &[u8] = src.as_ref();
 
-        let parsing_result = Parser::new(src.as_bytes(), false)
+        let parsing_result = Parser::new(src_bytes, false)
             .parse_all()
             .map_err(|e| e.to_string());
 
--- a/boa/src/lib.rs
+++ b/boa/src/lib.rs
@@ -78,16 +78,19 @@ pub type Result<T> = StdResult<T, Value>;
 /// It will return either the statement list AST node for the code, or a parsing error if something
 /// goes wrong.
 #[inline]
-pub fn parse(src: &str, strict_mode: bool) -> StdResult<StatementList, ParseError> {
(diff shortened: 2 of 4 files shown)


RageKnify (approved):

Razican (approved):

Razican:
Thanks!

PR #987, opened 2020-12-20
Implement property accessors

### LINKED ISSUE #986: Implement property accessors
**ECMASCript feature**
I would like to see the property accessors parsing and execution implemented. [ECMAScript specification][spec].

[spec]: https://tc39.es/ecma262/#sec-property-accessors

**Example code**
```javascript
let a = { get b() { return "c"}}
a.b
```
The expected output is `"c"`.

tofpie:
This Pull Request closes #986.

It changes the following:

- Execution of property accessors


tofpie:
@RageKnify I continue the conversation started on the issue.

In fact, property accessors are implemented both in `Value::get/set_field` and `GcObject::get/set`. As this PR has modified all the caller for `Value::get/set_field` (adding `context` and `?`), I could also replace all calls to `Value::get/set_field` by `to_object(..)?.get/set`, it would not be such a big effort. What to you think?

boa/src/builtins/map/mod.rs
```diff
@@ -96,14 +98,15 @@ impl Map {
...
-                        let len = args[0].get_field("length").to_integer(context)? as i32;
+                        let len = args[0].get_field("length", context)?.to_integer(context)? as i32;
                         for i in 0..len {
-                            let val = &args[0].get_field(i.to_string());
-                            let (key, value) = Self::get_key_value(val).ok_or_else(|| {
-                                context.construct_type_error(
-                                    "iterable for Map should have array-like objects",
-                                )
-                            })?;
+                            let val = &args[0].get_field(i.to_string(), context)?;
```
HalidOdat, line 103:
```suggestion
                            let val = &args[0].get_field(i, context)?;
```
I know it was there before, but it's better if we change so we avoid an unnecessary string allocation.

boa/src/context.rs
```diff
@@ -518,58 +510,68 @@ impl Context {
...
         if let Value::Object(ref x) = value {
             // Check if object is array
             if let ObjectData::Array = x.borrow().data {
-                let length = value.get_field("length").as_number().unwrap() as i32;
+                let length = value.get_field("length", self)?.as_number().unwrap() as i32;
                 let values = (0..length)
-                    .map(|idx| value.get_field(idx.to_string()))
-                    .collect();
-                return Ok(values);
+                    .map(|idx| value.get_field(idx.to_string(), self))
```
HalidOdat, line 534:
```suggestion
                    .map(|idx| value.get_field(idx, self))
```
Here too.

boa/src/context.rs
```diff
@@ -518,58 +510,68 @@ impl Context {
...
                                 .expect("Array was not initialized")
-                                .get_field(PROTOTYPE),
+                                .get_field(PROTOTYPE, self)?,
                         );
-                        array.set_field("0", key);
-                        array.set_field("1", value);
-                        array.set_field("length", Value::from(2));
-                        array
+                        array.set_field("0", key, self)?;
+                        array.set_field("1", value, self)?;
```
HalidOdat, line 563:
```suggestion
                        array.set_field(0, key, self)?;
                        array.set_field(1, value, self)?;
```
same here.

boa/src/property/mod.rs
```diff
@@ -118,11 +118,11 @@ impl From<DataDescriptor> for PropertyDescriptor {
 #[derive(Debug, Clone, Trace, Finalize)]
 pub struct AccessorDescriptor {
     /// The function serving as getter.
-    get: Option<GcObject>,
+    pub get: Option<GcObject>,
```
HalidOdat, line 121: Why are the fields public? we have accessors for them.

boa/src/value/mod.rs
```diff
@@ -456,19 +461,21 @@ impl Value {
...
         match self.get_property(key) {
             Some(ref desc) => match desc {
-                // TODO: Add accessors
-                PropertyDescriptor::Accessor(_) => Value::undefined(),
-                PropertyDescriptor::Data(desc) => desc.value(),
+                PropertyDescriptor::Accessor(AccessorDescriptor { get: Some(get), .. }) => {
+                    get.call(&self.clone(), &[], context)
+                }
+                PropertyDescriptor::Data(desc) => Ok(desc.value()),
+                _ => Ok(Value::undefined()),
```
HalidOdat, line 476: It would be better if didn't have to define the same implementation as `.get` but call get directly

HalidOdat (commented):
Looks good!

Check my comments, some are unrelated but it would be nice to change in this PR :)

boa/src/property/mod.rs
tofpie: I needed to put them public to use pattern matching in `GcObject::get/set`, but I can rewrite it without pattern matching.

HalidOdat: > I needed to put them public to use pattern matching in `GcObject::get/set`, but I can rewrite it without pattern matching.

In that case make it `pub(crate)` so it's only public to the crate level.

boa/src/value/mod.rs
tofpie: Maybe I could change all the calls to `Value::get/set_field` to to `to_object(..)?.get/set` as you have suggested in #577. Now that context has been added to `get/set_field`, the bulk of the work would be a search and replace. What do you think?

HalidOdat: > Maybe I could change all the calls to `Value::get/set_field` to `to_object(..)?.get/set` as you have suggested in #577. Now that context has been added to `get/set_field`, the bulk of the work would be a search and replace. What do you think?

It's not what needs to be done `.to_object()` does some other things and would not be spec compliant, the calls to `.get/.set` should be reinforced by the type system as it does not make much sense to have set/get for non-object (`Value` type) values as explained in #577.

we can merge this as it's main purpose was to add property accessors, and replace `.get_field` and `.set_field` builtin function by builtin function, through incremental changes (with many PRs), Allowing `.get_field/set_field` to take context is the first step, the end goal should be to completely remove them, it's really hard otherwise.

RageKnify:
> @RageKnify I continue the conversation started on the issue.
> 
> In fact, property accessors are implemented both in `Value::get/set_field` and `GcObject::get/set`. As this PR has modified all the calling sites for `Value::get/set_field` (adding `context` and `?`), I could also replace all calls to `Value::get/set_field` by `to_object(..)?.get/set`, it would not be such a big effort. What do you think?

Your idea is that it will be a simple transition in the future, right? Makes sense.

We will also eventually remove `Value::set_field` and `Value::get_field` since only `GcObject` will have that kind of functionality, right?

tofpie:
@HalidOdat said it is not as simple as I initially thought:
> > Maybe I could change all the calls to `Value::get/set_field` to `to_object(..)?.get/set` as you have suggested in #577. Now that context has been added to `get/set_field`, the bulk of the work would be a search and replace. What do you think?
> 
> It's not what needs to be done `.to_object()` does some other things and would not be spec compliant, the calls to `.get/.set` should be reinforced by the type system as it does not make much sense to have set/get for non-object (`Value` type) values as explained in #577.
> 
> we can merge this as it's main purpose was to add property accessors, and replace `.get_field` and `.set_field` builtin function by builtin function, through incremental changes (with many PRs), Allowing `.get_field/set_field` to take context is the first step, the end goal should be to completely remove them, it's really hard otherwise.

I guess I will leave as it is, just implementing the suggestions and add some unit tests.

boa/src/value/mod.rs
tofpie: I should have thought that if it was so easy, it would have already been done. Sorry if that suggestion seemed silly, I am quite a newbie on this code base.

HalidOdat (approved):
Looks good to me!

Once this is merged we can start removing the `.set_field` and `.get_field` functions from builtin functions :)

boa/src/builtins/map/mod.rs
```diff
@@ -2,13 +2,14 @@
...
     builtins::BuiltIn,
-    object::{ConstructorBuilder, FunctionBuilder, ObjectData, PROTOTYPE},
+    object::{ConstructorBuilder, FunctionBuilder, ObjectData},
     property::{Attribute, DataDescriptor},
     BoaProfiler, Context, Result, Value,
 };
 use ordered_map::OrderedMap;
 
 pub mod map_iterator;
+use crate::object::PROTOTYPE;
```
RageKnify, line 12: Is there a reason to change the path to `PROTOTYPE`?

boa/src/syntax/ast/node/object/mod.rs
```diff
@@ -8,6 +8,7 @@ use crate::{
 };
 use std::fmt;
 
+use crate::property::{AccessorDescriptor, Attribute, PropertyDescriptor};
```
RageKnify, line 11: This can be moved up to line 6, becoming just `property::{AccessorDescriptor, Attribute, PropertyDescriptor},`. This way it's all in 1 use statement, like in other files.

boa/src/value/display.rs
```diff
@@ -48,8 +48,9 @@ macro_rules! print_obj_value {
     };
     (props of $obj:expr, $display_fn:ident, $indent:expr, $encounters:expr, $print_internals:expr) => {
         print_obj_value!(impl $obj, |(key, val)| {
+        if val.is_data_descriptor() {
```
RageKnify, line 51: From line 51 to line 73 there's a missing indentation.

boa/src/value/mod.rs
```diff
@@ -158,13 +158,17 @@ impl Value {
...
-
-            let object = Object::create(object_prototype);
+        if global.is_some() {
+            let object = Object::create(
+                context
+                    .standard_objects()
+                    .object_object()
+                    .prototype()
+                    .into(),
+            );
```
RageKnify, line 171: This function becomes a bit weird, it receives an `Option` that is used as a boolean (the content is never checked), I think it should be refactored to possably just receive a boolean. @HalidOdat what do you think?
I'm ok with it staying as is for this PR, just noting that it could be cleaned up.

RageKnify (changes requested):
Some minor things can be cleaned up and I have some questions on `Value::new_object` current state.

boa/src/builtins/map/mod.rs
tofpie: Just my IDE doing that, when removing and auto-inserting again.

boa/src/value/mod.rs
tofpie: We could probably even remove the `global` argument. It is used with value `None` only in `Json::parse`:

https://github.com/boa-dev/boa/blob/22f8c5d0cee82262596b374832203be851d60525/boa/src/builtins/json/mod.rs#L72-L83

Maybe the call to `Value::new_object` could be replaced by `Value::object(Object::default())` in this location, or maybe it could even use an object with prototype as defined by default in `Value::new_object`

RageKnify (approved):

boa/src/value/mod.rs
HalidOdat: > Maybe the call to `Value::new_object` could be replaced by `Value::object(Object::default())` in this location, or maybe it could even use an object with prototype as defined by default in `Value::new_object`

Yes. I think we should take this approach. we replace the calls where None is passed to `Value::object(Object::default())`.

Then remove the `global: Option<Value>` so it only takes `context`:
```rust
    pub fn new_object(context: &mut Context) -> Self {
        let _timer = BoaProfiler::global().start_event("new_object", "value");

        context.construct_object().into()
    }
```

In the future we should remove it entirely, but since it is used in many places we can begin the process of deprecation.

RageKnify (approved):

HalidOdat (approved):

HalidOdat:
| Passed | 19,787 | 20,855 | **+1,068** |
| Failed | 43,121 | 42,053 | **-1,068** |
| Panics | 396 | 394 | **-2** |
| Conformance | 25.21 | 26.57 | **+1.36%** |

The conformance increase is really nice :)

PR #988, opened 2020-12-20
Add default parameter support

0x7D2B:
Modern JavaScript has support for default parameters in functions. Boa has support for the syntax but currently default parameters are ignored. For example, this code:

```javascript
function test(a, b = 2, c = a + b) {
    console.log(a, b, c);
}
test()
test(1)
```

Result produced by current Boa master:

```
undefined undefined undefined
1 undefined undefined
undefined
```

Expected result:

```
undefined 2 NaN
1 2 3
undefined
```

This PR implements support for this feature.

### DIFF
--- a/boa/src/object/gcobject.rs
+++ b/boa/src/object/gcobject.rs
@@ -145,6 +145,7 @@ impl GcObject {
                                 BindingStatus::Uninitialized
                             },
                         );
+                        context.realm_mut().environment.push(local_env.clone());
 
                         // Add argument bindings to the function environment
                         for (i, param) in params.iter().enumerate() {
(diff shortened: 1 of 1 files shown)


0x7D2B:
Need to handle scope correctly (function body has a separate scope if default parameters are used: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Scope_Effects) but can't really test this due to #989.

Razican:
If I understand this correctly, this is blocked on #989, right?

0x7D2B:
Yeah. I have a commit that sets up the right scopes in theory, but it doesn't work correctly due to issues caused by #989 (the second scope effects example given on the MDN page is a good showcase of the bug) so I can't fully finish this and add tests.

Razican:
Hello @0x7D2B is there any progress on this?

0x7D2B:
I should probably restart this one from scratch.

PR #1003, opened 2020-12-28
Improve Unicode support for identifier names

jevancc:
This PR implements the complete syntax (except UnicodeEscapeSequence) of "Names and Keywords", also known as an identifier, in the ECMAScript spec. All the identifier name related tests (except the escaped ones) in test262 are passed in this PR.

The implementation is based on Unicode® Standard Annex with Unicode version 13.0.0. The crate `unicode-general-category` v0.3.0 is added to lookup character categories, and of course it implements Unicode v13.0.0. Sad to say, there are no existing update-to-date crates that implement the identifier properties lookup functions we need, so I end up building tables in this PR.

The definition of UnicodeEscapeSequence in identifier name similar to the Unicode escape in string literal. I will work on that with a minor refactor to see if we can reuse it in both parts after this PR is merged.


RageKnify (approved):

RageKnify:
  | Passed | 22,657 | 22,691 | **+34** |
  | Failed | 40,251 | 40,217 | **-34** |
  | Panics | 364 | 364 | 0 |
  | Conformance | 28.86 | 28.91 | **+0.04%** |
  
  Numbers look good!

boa/src/syntax/lexer/mod.rs
```diff
@@ -252,6 +249,12 @@ impl<R> Lexer<R> {
                 '=' | '*' | '+' | '-' | '%' | '|' | '&' | '^' | '<' | '>' | '!' | '~' | '?' => {
                     Operator::new(next_ch as u8).lex(&mut self.cursor, start)
                 }
+                _ if c.is_digit(10) => {
+                    NumberLiteral::new(next_ch as u8).lex(&mut self.cursor, start)
+                }
+                _ if Identifier::is_identifier_start(c as u32) => {
```
Razican, line 255: Why do we do the conversion to `u32` here?

Razican (commented):
It's looking pretty good. I added a couple of comments :)

boa/src/syntax/lexer/mod.rs
jevancc: `c` here is a char, and the method `is_identifier_start` takes byte as input. I do this to make the match arms consistently processing only on `c`, but we could use `next_ch` here to prevent the type conversion.

boa/src/syntax/lexer/identifier_unicode_tables.rs
```diff
@@ -0,0 +1,370 @@
+//! This module implements the unicode lookup tables for identifier and pattern syntax.
+//!
+//! Unicode version: 13.0.0
+//!
+//! More information:
+//!  - [Unicode® Standard Annex #44][uax44]
+//!
+//! [uax44]: http://unicode.org/reports/tr44
```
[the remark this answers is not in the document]
jevancc: Actually, this file is generated by a Python script, but I don't know where in this repo to put the extra scripts so I didn't add it. It's very simple and can be easily rewritten in JS or Rust.

I am also thinking to have an additional crate to maintain the table, either to maintain an external crate or add an internal one. There are some crates like `unicode_categories` that provides char methods we need but not updated, send PR there and use it maybe a doable option. However, other engines like v8 maintain a huge table including even Unicode general categories for its use. I think the reason is to make sure the Unicode version and implementation always match the ECMAScript standard.

I don't suggest having the script to be run in every PR/commit, for 2 reasons:
- The standard update very slowly, and fetching the content from external sources and run the script may be slow. We can always manually re-run the script once we need it.
- We need to update the source URL in our script, check if the Unicode standard updates the format and modify our script, and make sure the unicode version is what we need in the latest ECMAScript standard. It's likely that we need to manually do something when the a new Unicode version published.

boa/src/syntax/lexer/mod.rs
Razican: I think it's fine as-is :) thanks for the clarification!

boa/src/syntax/lexer/identifier_unicode_tables.rs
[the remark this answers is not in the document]
Razican: For now, I guess we could have the script in the root. We might want to have a `boa_unicode` crate in the workspace providing all this. What do @jasonwilliams, @Lan2u, @RageKnify and @HalidOdat think?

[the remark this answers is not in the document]
Lan2u: If we are going to use a script to generate this then it makes sense to put it in a separate crate to keep the main boa repo simple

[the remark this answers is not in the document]
jevancc: I can create a new crate `boa_unicode` and move the table with scripts there in this PR. How about that?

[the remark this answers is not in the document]
Razican: Yes let's do that :)

boa_unicode/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Returns `true` if this value is a member of "Other_ID_Continue".
+    fn is_other_id_continue(self) -> bool;
+
+    /// Returns `true` if this value is a member of "Pattern_Syntax".
+    fn is_pattern_syntax(self) -> bool;
+
+    /// Returns `true` if this value is a member of "Pattern_White_Space".
+    fn is_pattern_whitespace(self) -> bool;
+}
```
Razican, line 30: This trait needs a bit of trait-level documentation. Mostly to explain what kinds of values implement it, why do they implement it and in what contexts is this trait useful.

boa_unicode/build_tables.js
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env node
```
Razican, line 1: I would add a section in the `CONTRIBUTING.ml` file to explain how to use this script, what dependencies it has and how to run it locally to update the tables. Also, a short explanation on what it does and why we need it would be nice. Maybe some comments in the script itself.

boa_unicode/build_tables.js
```diff
@@ -0,0 +1,92 @@
...
+  ["Other_ID_Start", "OTHER_ID_START"],
+  ["Pattern_White_Space", "PATTERN_WHITE_SPACE"],
+];
+
+const OUTPUT_FILE = path.join(__dirname, "./src/tables.rs");
+const OUTPUT_FILE_DOC_COMMENT = `
+//! This module implements the unicode lookup tables for identifier and pattern syntax.
+//! Unicode version: 13.0.0
+//!
+//! This file is generated by \`boa_unicode/build_tables.js\`.
```
Razican, line 22: I would add a small sentence like "Please, do not modify directly".

Razican (commented):
This looks much, much cleaner now. I'm only missing a bit of documentation, but for the rest, it's good to go from my side :)

boa_unicode/build_tables.js
jevancc: I prefer to add file headers in `build_table.js` or add a `README.md` under `boa_unicode` to keep the top-level `CONTRIBUTING.md` clean. Is one of them good for you?

jevancc:
@Razican I'm thinking to check the Unicode version of `unicode-general-category` in building or testing and throw warning/error when mismatch. It is an external dependency and the implementations' Unicode versions may mismatch when that crate is updated but the identifier properties tables are not.

Where do you think we can add this check?

Razican:
> @Razican I'm thinking to check the Unicode version of `unicode-general-category` in building or testing and throw warning/error when mismatch. It is an external dependency and the implementations' Unicode versions may mismatch when that crate is updated but the identifier properties tables are not.
> 
> Where do you think we can add this check?

This could be checked either in a `build.rs` file in `boa_unicode`.

jevancc:
> > @Razican I'm thinking to check the Unicode version of `unicode-general-category` in building or testing and throw warning/error when mismatch. It is an external dependency and the implementations' Unicode versions may mismatch when that crate is updated but the identifier properties tables are not.
> > Where do you think we can add this check?
> 
> This could be checked either in a `build.rs` file in `boa_unicode`.

It won't work. `build.rs` cannot access regular dependencies. It's fine to not have this check, just manually check it every time before updating the dependency `unicode-general-category`.

Razican:
> It won't work. `build.rs` cannot access regular dependencies. It's fine to not have this check, just manually check it every time before updating the dependency `unicode-general-category`.

You can also add it as a "build" dependency and check it, but it's true that those versions can get out of sync. Could we maybe add a test in `boa_unicode` that fails if versions are out of sync? That way, when we execute the test suite, it will fail.

boa_unicode/build_tables.js
Razican: Let's create a `README.md` file under `boa_unicode` then. But I would at least add a link to the `CONTRIBUTING.md` file, to make sure everything is linked, this being part of the project too.

Razican (approved):

boa_unicode/build_tables.js
jevancc: Will do :)

Razican (approved):
Thanks for the changes, this looks very good to me now :)

PR #1005, opened 2020-12-29
Fix `Object.defineProperty()`

HalidOdat:
It changes the following:
 - Fix panic if first argument is not supplied.
 - Fix panic if second argument is not supplied.
 - Fix bug when the second argument is not a object.
 - Implemented `DefinePropertyOrThrow()`

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -287,19 +287,23 @@ impl Object {
 
     /// Define a property in an object
     pub fn define_property(_: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
-        let obj = args.get(0).expect("Cannot get object");
-        let prop = args
-            .get(1)
-            .expect("Cannot get object")
-            .to_property_key(context)?;
+        let object = args.get(0).cloned().unwrap_or_else(Value::undefined);
+        if let Some(mut object) = object.as_object() {
+            let key = args
+                .get(1)
+                .unwrap_or(&Value::undefined())
+                .to_property_key(context)?;
+            let desc = args
+                .get(2)
+                .unwrap_or(&Value::undefined())
+                .to_property_descriptor(context)?;
 
-        let desc = if let Value::Object(ref object) = args.get(2).cloned().unwrap_or_default() {
-            object.to_property_descriptor(context)?
+            object.define_property_or_throw(key, desc, context)?;
+
+            Ok(object.into())
         } else {
-            return context.throw_type_error("Property description must be an object");
-        };
-        obj.set_property(prop, desc);
-        Ok(obj.clone())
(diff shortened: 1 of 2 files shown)


RageKnify (approved):

boa/src/builtins/object/mod.rs
```diff
@@ -289,19 +289,23 @@ impl Object {
...
+                .unwrap_or(&Value::undefined())
+                .to_property_key(context)?;
+            let desc = args
+                .get(2)
+                .unwrap_or(&Value::undefined())
+                .to_property_descriptor(context)?;
+
+            object.define_property_or_throw(key, desc, context)?;
+
+            Ok(Value::undefined())
```
tofpie, line 305: You could also fix this one: `Object.defineProperty` should return the object itself, and not `undefined` (see [spec][spec])

[spec]: https://tc39.es/ecma262/#sec-object.defineproperty

HalidOdat: > You could also fix this one: `Object.defineProperty` should return the object itself, and not `undefined` (see [spec](https://tc39.es/ecma262/#sec-object.defineproperty))

Nice catch. You are right, I'll fix it now :)

RageKnify:
Somehow 680579a changed 2 test262 tests from passing to panicking, the change can be seen by looking at the edit history of the comment showing the results, https://github.com/boa-dev/boa/pull/1005#issuecomment-751942679

Don't understand how that change could introduce panicking.

HalidOdat:
> Somehow [680579a](https://github.com/boa-dev/boa/commit/680579a9b6e818e0d135a353c106635693cf3a6c) changed 2 test262 tests from passing to panicking, the change can be seen by looking at the edit history of the comment showing the results, [#1005 (comment)](https://github.com/boa-dev/boa/pull/1005#issuecomment-751942679)
> 
> Don't understand how that change could introduce panicking.

It is probably be due to another unrelated bug. `Object.defineProperty` should return the object.

tofpie:
The new panic is caused by the test [15.2.3.6-4-187][panic-test].

This test calls `Object.defineProperty` on the property `length` of an `Array` setting `writable` to false but without defining a value.

As `defineProperty` now returns the object, it is displayed, and the panic occurs here, as `length` is `undefined`:
https://github.com/boa-dev/boa/blob/67ee6faa5819ba5e97ae826244a88ce370e32070/boa/src/value/display.rs#L98-L109

It seems that setting only `writable` in a property should not change its value, so the `length` should remain a number.

[panic-test]: https://github.com/tc39/test262/blob/42bf3a9f7aed3790dc1d829f290dee033df41c5b/test/built-ins/Object/defineProperty/15.2.3.6-4-187.js#L1

Issue #989, opened 2020-12-20
Functions seem to have access to their caller's scope


0x7D2B:
When calling a function, it appears to be able to access variables defined in the context it was called in, even if it's not supposed to be able to.

**To Reproduce**

```javascript
function f1() {
    return a
}
function f2() {
    let a = 1
    return f1()
}

f2()
```

`f2()` appears to return 1 instead of raising a `ReferenceError`.

**Expected behavior**

`ReferenceError` is raised 

**Build environment (please complete the following information):**

- OS: Linux
- Target triple: aarch64-unknown-linux-gnu
- Rustc version: rustc 1.47.0 (18bf6b4f0 2020-10-07), rustc 1.48.0 (7eac88abb 2020-11-16)

I also observed this issue in Boa Playground.

RageKnify:
I think you're mixing 2 different bugs together, the expeted behaviour would be for a ReferenceError to be thrown when executing `f1`.
`f2` returning `1` is caused by #672 and is a separate issue.
As an example, executing the example you provided in a Chrome console results in: 
```
VM32:2 Uncaught ReferenceError: a is not defined
    at f1 (<anonymous>:2:5)
    at f2 (<anonymous>:6:5)
    at <anonymous>:1:1
```

0x7D2B:
Hmm, maybe I had an `a` declared somewhere earlier while testing in the browser which is why it was returning undefined. I updated the test case to avoid #672 but it still returns 1 instead of raising `ReferenceError`.

0x7D2B:
I started digging deeper into the problem and I think the root cause is how lexical environments are implemented. When function environment records are created, they seem to be using the right outer environment, but when they're pushed to the LexicalEnvironment stack, [that environment gets overwritten](https://github.com/boa-dev/boa/blob/c0e7791c49e73f562efb96fd4ca4d7ff40e16443/boa/src/environment/lexical_environment.rs#L97). Then, even if that's fixed, the issue will still persist because of the way the environment stack is used in the rest of the lexical environment implementation, for example in [`get_binding_value`](https://github.com/boa-dev/boa/blob/c0e7791c49e73f562efb96fd4ca4d7ff40e16443/boa/src/environment/lexical_environment.rs#L241-L252). Because of this, the outer environment for function calls will always end up being the environment the function was called in, rather than where it was declared.

[...]

PR #1010, opened 2020-12-29
Modify environment binding behaviour of function

### LINKED ISSUE #669: Redefining a function panics with binding already exists
Redefining a function in repl panics with binding already exists error

**To Reproduce**
cargo run
`function x() {}` enter
`function x() { return 3; }` enter panics

**Expected behavior**
It should override the previous binding

Backtrace

```rust
thread 'main' panicked at 'Binding already exists!', boa/src/environment/global_environment_record.rs:108:13
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1537
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:217
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:526
  11: std::panicking::begin_panic
  12: <boa::environment::global_environment_record::GlobalEnvironmentRecord as boa::environment::environment_record_trait::EnvironmentRecordTrait>::create_mutable_binding
             at boa/src/environment/global_environment_record.rs:108
  13: boa::environment::lexical_environment::LexicalEnvironment::create_mutable_binding
             at /Users/dvtkrlbs/.cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.8.2/src/raw/mod.rs:1822
  14: boa::exec::declaration::<impl boa::exec::Executable for boa::syntax::ast::node::declaration::FunctionDecl>::run
             at boa/src/exec/declaration/mod.rs:24
  15: <boa::syntax::ast::node::Node as boa::exec::Executable>::run
             at boa/src/exec/mod.rs:389
  16: boa::exec::statement_list::<impl boa::exec::Executable for boa::syntax::ast::node::statement_list::StatementList>::run
             at boa/src/exec/statement_list.rs:15
  17: boa::forward_val
             at boa/src/lib.rs:89
  18: boa::main
             at boa_cli/src/main.rs:190
  19: std::rt::lang_start::{{closure}}
             at /Users/dvtkrlbs/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/src/libstd/rt.rs:67
  20: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:52
  21: std::panicking::try::do_call
             at src/libstd/panicking.rs:348
  22: std::panicking::try
             at src/libstd/panicking.rs:325
  23: std::panic::catch_unwind
             at src/libstd/panic.rs:394
  24: std::rt::lang_start_internal
             at src/libstd/rt.rs:51
  25: std::rt::lang_start
             at /Users/dvtkrlbs/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/src/libstd/rt.rs:67
  26: main

```

### LINKED ISSUE #858: SyntaxError not generated for duplicate identifier names with let declarations
**To Reproduce**
```JavaScript
>> let x = 12;
undefined
>> var x = 12;
undefined
```

**Expected behavior**
SyntaxError must be thrown by the parser. Environment is not supposed to deal with this according to the spec.

I think [here](https://www.ecma-international.org/ecma-262/11.0/index.html#sec-let-and-const-declarations-static-semantics-early-errors) is where the spec talks about syntax error in `let` declaration.

54k1:
This Pull Request fixes/closes #669.

If a binding for the function name already exists, we should simply override it (`set_mutable_binding`), otherwise we need to create a new binding. For this to be accurate, we need to also fix #858, since now we are simply overriding the binding (potentially `let` or `const`).

This seems to fix it, however, I haven't found this in the documentation. The implementation of `var` also seems to follow this. Could anybody provide a link to the part where this is described in the documentation?

### DIFF
--- a/boa/src/syntax/ast/node/declaration/function_decl/mod.rs
+++ b/boa/src/syntax/ast/node/declaration/function_decl/mod.rs
@@ -94,17 +94,23 @@ impl Executable for FunctionDecl {
 
         // Set the name and assign it in the current environment
         val.set_field("name", self.name());
-        context
-            .realm_mut()
-            .environment
-            .create_mutable_binding(self.name().to_owned(), false, VariableScope::Function)
-            .map_err(|e| e.to_error(context))?;
 
-        context
-            .realm_mut()
-            .environment
-            .initialize_binding(self.name(), val)
-            .map_err(|e| e.to_error(context))?;
+        let environment = &mut context.realm_mut().environment;
+        if environment.has_binding(self.name()) {
+            environment
+                .set_mutable_binding(self.name(), val, true)
+                .map_err(|e| e.to_error(context))?;
+        } else {
+            environment
+                .create_mutable_binding(self.name().to_owned(), false, VariableScope::Function)
+                .map_err(|e| e.to_error(context))?;
+
+            context
+                .realm_mut()
+                .environment
+                .initialize_binding(self.name(), val)
+                .map_err(|e| e.to_error(context))?;
+        }
 
         Ok(Value::undefined())
     }
--- a/boa/src/syntax/ast/node/declaration/mod.rs
+++ b/boa/src/syntax/ast/node/declaration/mod.rs
@@ -19,3 +19,6 @@ pub use self::{
     let_decl_list::{LetDecl, LetDeclList},
     var_decl_list::{VarDecl, VarDeclList},
 };
+
+#[cfg(test)]
+mod tests;
--- a/boa/src/syntax/ast/node/declaration/tests.rs
+++ b/boa/src/syntax/ast/node/declaration/tests.rs
@@ -1,0 +1,12 @@
+use crate::exec;
+
+#[test]
+fn duplicate_function_name() {
+    let scenario = r#"
+    function f () {}
+    function f () {return 12;}
+    f()
+    "#;
+
+    assert_eq!(&exec(scenario), "12");
+}


RageKnify:
  | Passed | 19,787 | 19,819 | **+32** |
  | Failed | 43,121 | 43,089 | **-32** |
  | Panics | 396 | 365 | **-31** |
  | Conformance | 25.21 | 25.25 | **+0.04%** |

The numbers look good.

I think adding some tests would be nice to make the desired behaviour clearer.

jasonwilliams:
Thanks for looking at this @54k1
I think currently we don't check for duplicates at parse-time nor is there a mechanism for doing so. So we don't have good early error support for that.

One place to handle that would be here (however this only accounts for when bindings are created as a list.
https://github.com/jasonwilliams/boa/blob/master/boa/src/syntax/parser/statement/declaration/lexical.rs#L145-L208

This could need a new issue, but im open to hear what the others think

> This seems to fix it, however, I haven't found this in the documentation. The implementation of var also seems to follow this. Could anybody provide a link to the part where this is described in the documentation?

Could you go into more detail here? What documentation are you referring to? The specification?

> I think adding some tests would be nice to make the desired behaviour clearer.

Yep adding some tests here would be good enough as a fix for the "environment side"

Razican (approved):
This fix looks good to me. We still need to fix #858, but I believe that is a different issue. We need to implement a pre-parsing strategy I think.

54k1:
@jasonwilliams Yes, I meant the [specification](https://tc39.es/ecma262/), sorry about the confusion
@Razican Yes, it's not directly related to #858, but assumes that `SyntaxError` is thrown for those cases mentioned in #858.
@RageKnify , I've added just a basic test. What other things can we test?

RageKnify:
I've found [`CreateGlobalFunctionBinding`](https://tc39.es/ecma262/#sec-createglobalfunctionbinding) in the spec, which seems to be the approppriate section of the spec that describes the behaviour we were looking for. 

The "call-trace" seems to be `ScriptEvaluation` -> `GlobalDeclarationInstantiation` -> `env.CreateGlobalFunctionBinding`

RageKnify:
Seems like `run` for `FunctionDecl` should be using [`GlobalEnvironmentRecord::create_global_function_binding`](https://github.com/boa-dev/boa/blob/90a8587ff5223647ba0ae0c5118039d04570a870/boa/src/environment/global_environment_record.rs#L100), but instead the bindings are created in the `LexicalEnvironment`. I'm not really familiar with these aspects, maybe @jasonwilliams or another Member have a better understanding of these concepts.

jasonwilliams:
@RageKnify that's interesting, the way functions are bound right now is we just use the containing scope. I don't know if we're supposed to be using `create_global_function_binding` for all of our functions maybe that's some research we need to do. That would in theory make runtime slower though, as every function call would need to traverse the scope chain to the top.

@54k1 you're right in that in theory we would catch this during parsing (as an early error), but we don't have functionality to check duplicates as parsing time just yet, but I see you've created an issue for that anyway.

Looks good to me

jasonwilliams (approved):

Razican:
Does this help with #989?

RageKnify:
> Does this help with #989?

No, just tested it, ran the exemple code from #989 in this branch and the bug remains.

HalidOdat (approved):

PR #1024, opened 2021-01-01
Fix delete when the property is not configurable

tofpie:
This Pull Request fixes an issue with `delete` when the property to delete is not configurable. In this case, `delete` should not remove the property from the object, and it should return `false`.

It changes the following:

- The execution of the `delete` operator

### DIFF
--- a/boa/src/syntax/ast/node/operator/unary_op/mod.rs
+++ b/boa/src/syntax/ast/node/operator/unary_op/mod.rs
@@ -90,12 +90,15 @@ impl Executable for UnaryOp {
                     get_const_field
                         .obj()
                         .run(context)?
-                        .remove_property(get_const_field.field()),
+                        .to_object(context)?
+                        .delete(&get_const_field.field().into()),
                 ),
                 Node::GetField(ref get_field) => {
                     let obj = get_field.obj().run(context)?;
                     let field = &get_field.field().run(context)?;
-                    let res = obj.remove_property(field.to_string(context)?.as_str());
+                    let res = obj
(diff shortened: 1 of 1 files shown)


Razican:
Nice conformance improvement!

  | Passed | 23,335 | 23,800 | **+465** |
  | Failed | 39,573 | 39,108 | **-465** |
  | Panics | 364 | 364 | 0 |
  | Conformance | 29.73 | 30.32 | **+0.59%** |

Razican (approved):
Looks good, thanks!

boa/src/syntax/ast/node/operator/unary_op/mod.rs
```diff
@@ -90,12 +90,15 @@ impl Executable for UnaryOp {
...
+                        .to_object(context)?
+                        .delete(&get_const_field.field().into()),
                 ),
                 Node::GetField(ref get_field) => {
                     let obj = get_field.obj().run(context)?;
                     let field = &get_field.field().run(context)?;
-                    let res = obj.remove_property(field.to_string(context)?.as_str());
+                    let res = obj
+                        .to_object(context)?
+                        .delete(&field.to_string(context)?.as_str().into());
```
HalidOdat, line 101:
```suggestion
                        .delete(&field.to_property_key(context)?);
```

I think we should be converting to a property key (`PropertyKey`), since `.to_string()` converts symbols to string and therefore we wont we able to delete symbol properties.

RageKnify (approved):

HalidOdat (approved):

PR #1026, opened 2021-01-02
Implement String.prototype.split

jevancc:
This Pull Request implements `String.prototype.split`

It changes the following:

- Add the method implementation
- Add unit tests

### DIFF
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -1213,7 +1213,7 @@ impl String {
                 .collect(),
             Some(separator) => string
                 .split(separator.as_str())
-                .map(|v| Value::from(v))
+                .map(&Value::from)
                 .take(limit)
                 .collect(),
         };
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -1224,6 +1224,7 @@ impl String {
             None => vec![Value::from(string)],
             Some(separator) if separator == "" => string
                 .encode_utf16()
+                // TODO: Support keeping invalid code point in string
                 .map(|cp| Value::from(std::string::String::from_utf16_lossy(&[cp])))
                 .take(limit)
                 .collect(),
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -1186,11 +1186,27 @@ impl String {
         } else {
             let arg0 = args
                 .get(0)
-                .expect("failed to get argument for String method");
+                .expect("Failed to get argument for String method");
 
             if arg0.is_null_or_undefined() {
                 None
             } else {
+                if let Some(splitter) =
+                    arg0.get_property(context.well_known_symbols().split_symbol())
+                {
+                    let splitter = splitter.as_data_descriptor().unwrap().value();
+                    if splitter.is_function() {
+                        let arguments = vec![
+                            Value::from(string),
+                            args.get(1)
+                                .map(|x| x.to_owned())
+                                .unwrap_or(Value::Undefined),
+                        ];
+                        return context.call(&splitter, this, &arguments);
+                    } else if !splitter.is_undefined() {
+                        return Err(Value::from("separator[Symbol.split] is not a function"));
+                    }
+                }
                 arg0.to_string(context).ok()
             }
         };
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -1179,6 +1179,7 @@ impl String {
     /// [spec]: https://tc39.es/ecma262/#sec-string.prototype.split
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split
     pub(crate) fn split(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+        let this = this.require_object_coercible(context)?;
         let string = this.to_string(context)?;
 
         let separator = if args.is_empty() {
--- a/boa/src/builtins/string/mod.rs
+++ b/boa/src/builtins/string/mod.rs
@@ -1191,13 +1191,13 @@ impl String {
 
                 match separator.get_method(context, key) {
                     Ok(splitter) => splitter.map(|splitter| {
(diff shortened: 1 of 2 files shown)


Razican:
I will check the code through the day, but the conformance increase looks really good :)

  | Passed | 23,335 | 23,468 | **+133** |
  | Failed | 39,573 | 39,440 | **-133** |
  | Panics | 364 | 367 | **+3** |
  | Conformance | 29.73 | 29.90 | **+0.17%** |

3 extra panics, though

jevancc:
@Razican The three extra panics come from `built-ins/map/prototype/keys/returns-iterator.js` (no `Set` implementation)  and `test/built-ins/string/prototype/split/cstm-split-get-err.js` (bug in `Object.defineProperty`). I have no idea why they did not panic in the master branch but here.

Razican:
> I have no idea why they did not panic in the master branch but here.

I would guess this now enables some tests to run further, which shows some panics that did not appear before because it was failing early.

boa/src/builtins/string/mod.rs
```diff
@@ -1165,6 +1166,81 @@ impl String {
...
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split
+    pub(crate) fn split(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+        let string = this.to_string(context)?;
+
+        let separator = if args.is_empty() {
+            None
+        } else {
+            let arg0 = args
+                .get(0)
+                .expect("Failed to get argument for String method");
```
Razican, line 1189: I think this this would be more idiomatic using `args.get(0).and_then(|arg0| { ... });`. It's true that the `Err` variant can make it a bit more complex, but maybe returning an `Option<Result>` and then calling `transpose()` on it could help.

boa/src/builtins/string/mod.rs
```diff
@@ -1165,6 +1166,81 @@ impl String {
...
+            }
+        };
+
+        let limit = if args.len() < 2 {
+            std::u32::MAX as usize
+        } else {
+            args.get(1)
+                .expect("Could not get the limit argument")
+                .to_integer(context)? as usize
+        };
```
Razican, line 1222: This would be more idiomatic using:
```rust
args.get(1).map(|arg| arg.to_integer(context).map(|limit|limit as usize)).transpose()?.unwrap_or(std::u32::MAX as usize);
```
Or something like it

boa/src/builtins/string/mod.rs
```diff
@@ -1165,6 +1166,81 @@ impl String {
...
+
+        let values: Vec<Value> = match separator {
+            None if limit == 0 => vec![],
+            None => vec![Value::from(string)],
+            Some(separator) if separator == "" => string
```
Razican, line 1227: I think this would be better if we used `separator.is_empty()`. It should be a bit more performant.

Razican (commented):
It's looking good! I added some sugestions to make the code a bit more idiomatic :)

boa/src/builtins/string/mod.rs
```diff
@@ -1165,6 +1166,81 @@ impl String {
...
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-string.prototype.split
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split
+    pub(crate) fn split(this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+        let string = this.to_string(context)?;
+
```
RageKnify, line 1183: We're skipping step 1 of the spec, looking at other methods in this file it seems like we do it in many of them, we can leave the others for later, but I'd like to have this one be right.
[`RequireObjectCoercible`](https://tc39.es/ecma262/#sec-requireobjectcoercible) requires us to to throw a `TypeError` if `this` is `undefined` or `null`.

I went to check which exception is thrown by Chrome: `String.prototype.split called on null or undefined`
But `RequireObjectCoercible` is also called in some RunTimeSemantics algorithms, and in that case the messages won't have a function to reffer to. Also as an example that doesn't mention the function name, `Object.fromEntries(null)` results in `TypeError: undefined is not iterable`.

For now just writing a function that will throw a `TypeError` mentioning `null` and `undefined` seems like enough to me.

RageKnify (changes requested):
Looks good, noted a few improvements to be closer to the spec.

boa/src/builtins/string/mod.rs
Razican: Should we add that function to `context`, like others?

RageKnify: Yeah, making it a method of `Context` might be best.

boa/src/builtins/string/mod.rs
jevancc: Good idea, I'll remove the outer if-else with this.

boa/src/builtins/string/mod.rs
jevancc: So do we have any existing function, not limited to `context` methods, to do this? If not, is it just adding an if block and return TypeError if `this.is_null_or_undefined()` is true?

RageKnify: Since this is a pattern that will repeat in a lot of functions I'd ask you to implement the method in `Context` (that can be just an `if` for now) and in `split` we would have:
```rust
let this = context.require_object_coercible(this)?
```
And the implementation of `require_object_coercible` can probably just be:
```rust
fn require_object_coercible(&self, value: &Value) -> Result<Value> {
    if value.is_null_or_undefined() {
        self.throw_type_error("value was null or undefined")
    } else {
        value
    }
}
```

HalidOdat: there already is a correct implementation of `RequireObjectCoercible` (`Value::require_object_coercible(context)`)

RageKnify (approved):

Razican (approved):
Looks good, let's merge it :)

PR #1027, opened 2021-01-02
wrap the regress error for now

jasonwilliams:
Regress panics on "Invalid character escape"s.
However on V8 and other JS engines these seem to be acceptable.

```js
const a = /,\;/;
```

was causing issues in Boa, due to regress panicing and us not catching.
This PR now changes it to a SyntaxError, however this is a temporary fix, ideally it should be accepted.

Raised upstream:
https://github.com/ridiculousfish/regress/issues/26

### DIFF
--- a/boa/src/builtins/regexp/tests.rs
+++ b/boa/src/builtins/regexp/tests.rs
@@ -98,3 +98,12 @@ fn to_string() {
     );
     assert_eq!(forward(&mut context, "/\\n/g.toString()"), "\"/\\n/g\"");
 }
+
+#[test]
+fn no_panic_on_invalid_character_escape() {
+    let mut context = Context::new();
+
+    // This used to panic, we now return an error
+    // The line below should not cause Boa to panic
+    forward(&mut context, r"const a = /,\;/");
+}
--- a/boa/src/builtins/regexp/mod.rs
+++ b/boa/src/builtins/regexp/mod.rs
@@ -98,7 +98,7 @@ impl RegExp {
     pub(crate) const LENGTH: usize = 2;
 
     /// Create a new `RegExp`
-    pub(crate) fn constructor(this: &Value, args: &[Value], _: &mut Context) -> Result<Value> {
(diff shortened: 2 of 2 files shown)


Razican:
Looking good!

  | Passed | 23,805 | 23,944 | **+139** |
  | Failed | 39,103 | 38,964 | **-139** |
  | Panics | 362 | 147 | **-215** |
  | Conformance | 30.33 | 30.50 | **+0.18%** |

boa/src/builtins/regexp/tests.rs
```diff
@@ -98,3 +98,13 @@ fn to_string() {
     );
     assert_eq!(forward(&mut context, "/\\n/g.toString()"), "\"/\\n/g\"");
 }
+
+#[test]
+fn no_panic_on_invalid_character_escape() {
+    let mut context = Context::new();
+
+    assert_eq!(
```
Razican, line 106: I would just try to run it, without checking that it creates a Syntax error, since that will change in the future once we fix regress. If it panics, the test will fail, if it doesn't, it will pass.

Razican (approved):
Looks good, I would just remove the check for a Syntax error in the test, in order to avoid changing it when we fix regress.

Razican:
This is linked to #817.

boa/src/builtins/regexp/tests.rs
```diff
@@ -98,3 +98,12 @@ fn to_string() {
     );
     assert_eq!(forward(&mut context, "/\\n/g.toString()"), "\"/\\n/g\"");
 }
+
+#[test]
+fn no_panic_on_invalid_character_escape() {
+    let mut context = Context::new();
+
+    // This used to panic, we now return an error
+    // The line below should now cause Boa to panic
```
Razican, line 107: "should not", right?

Razican (approved):

boa/src/builtins/regexp/tests.rs
jasonwilliams: fixed :D

Razican (approved):

RageKnify (approved):

Lan2u (approved):

PR #1030, opened 2021-01-02
VM: Implement variable declaration (var, const, and let)

### LINKED ISSUE #1028: Implement let declaration list instruction (VM)
Set your JS to this and then run
```javascript
let a = 1;
```
You should get `undefined` as expected. But it will run with no errors

Leaving the JS the same run `cargo run --features vm ../tests/js/test.js ` or `Cargo Run (VM)` on VSCode.
You will reach an "unimplemented" error, arising from here:
https://github.com/boa-dev/boa/blob/master/boa/src/vm/compilation.rs#L61

This is because `Node::LetDeclList` hasn't been implemented in this block https://github.com/boa-dev/boa/blob/master/boa/src/vm/compilation.rs#L42-L62

## Adding the instruction
You can look at the spiderMonkey Op codes for inspiration, namely `DefVar` in https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Internals/Bytecode

Then you can add `DefVar` here:
https://github.com/boa-dev/boa/blob/master/boa/src/vm/instructions.rs#L2-L60

## Implementing the instruction
Then here you can push the variable to the right environment, similar to how we do this here:
https://github.com/boa-dev/boa/blob/master/boa/src/syntax/ast/node/declaration/let_decl_list/mod.rs#L44-L48

AnnikaCodes:
This Pull Request fixes #1028.

It changes the following:

- Adds a `DefVar` VM instruction
- Includes `Context` in the function signature of `CodeGen::compile`
- Compiles `LetDeclList`s into `DefVar` instructions

### DIFF
--- a/boa/src/vm/compilation.rs
+++ b/boa/src/vm/compilation.rs
@@ -43,6 +43,28 @@ impl Compiler {
         self.pool.push(name.into().into());
         self.pool.push(value.into().into());
     }
+
+    pub fn add_deflet_instruction<N, V>(&mut self, name: N, value: V)
+    where
+        N: Into<RcString>,
+        V: Into<Value>,
+    {
+        let index = self.pool.len();
+        self.add_instruction(Instruction::DefLet(index, index + 1));
+        self.pool.push(name.into().into());
+        self.pool.push(value.into().into());
+    }
+
+    pub fn add_defconst_instruction<N, V>(&mut self, name: N, value: V)
+    where
+        N: Into<RcString>,
+        V: Into<Value>,
+    {
+        let index = self.pool.len();
+        self.add_instruction(Instruction::DefConst(index, index + 1));
+        self.pool.push(name.into().into());
+        self.pool.push(value.into().into());
+    }
 }
 
 pub(crate) trait CodeGen {
@@ -71,6 +93,16 @@ impl CodeGen for Node {
             }
             Node::BinOp(ref op) => op.compile(compiler, context),
             Node::UnaryOp(ref op) => op.compile(compiler, context),
+            Node::VarDeclList(ref list) => {
+                for var_decl in list.as_ref() {
+                    let value = match var_decl.init() {
+                        Some(v) => v.run(context).unwrap_or_else(|_| Value::undefined()),
+                        None => Value::undefined(),
+                    };
+
+                    compiler.add_defvar_instruction(var_decl.name(), value);
+                }
+            }
             Node::LetDeclList(ref list) => {
                 for let_decl in list.as_ref() {
                     let value = match let_decl.init() {
@@ -78,7 +110,17 @@ impl CodeGen for Node {
                         None => Value::undefined(),
                     };
 
-                    compiler.add_defvar_instruction(let_decl.name(), value);
+                    compiler.add_deflet_instruction(let_decl.name(), value);
+                }
+            }
+            Node::ConstDeclList(ref list) => {
+                for const_decl in list.as_ref() {
+                    let value = match const_decl.init() {
+                        Some(v) => v.run(context).unwrap_or_else(|_| Value::undefined()),
+                        None => Value::undefined(),
+                    };
+
+                    compiler.add_defconst_instruction(const_decl.name(), value);
                 }
             }
             _ => unimplemented!(),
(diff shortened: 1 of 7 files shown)


AnnikaCodes:
I made this support `var` and `const` as well.

boa/src/vm/instructions.rs
```diff
@@ -57,6 +57,10 @@ pub enum Instruction {
     Neg,
     BitNot,
     Not,
+
+    DefVar(usize, usize),
+    DefLet(usize, usize),
+    DefConst(usize, usize),
```
jasonwilliams, line 63: Could we add some documentation/comments to these? So others know what the 2 usize arguments are for
As a bonus putting these in some alphabetical order would be nice (but not essential)

boa/src/vm/mod.rs
```diff
@@ -268,6 +267,66 @@ impl<'a> VM<'a> {
...
+                        .map_err(|e| e.to_error(self.ctx))?;
+
+                    self.ctx
+                        .realm_mut()
+                        .environment
+                        .initialize_binding(&name, value)
+                        .map_err(|e| e.to_error(self.ctx))?;
+
+                    self.push(Value::undefined());
+                }
```
jasonwilliams, line 329: This is a great start.
Looking at https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Internals/Bytecode defVar doesn't add anything to the stack, so I don't think these functions need to push an `undefined` value once they're finished

boa/src/vm/instructions.rs
Razican: We could even use attributes in a struct-like variant, with names. Would make things much more clear.

boa/src/vm/compilation.rs
```diff
@@ -56,8 +91,38 @@ impl CodeGen for Node {
...
+                        None => Value::undefined(),
+                    };
+
+                    compiler.add_defvar_instruction(var_decl.name(), value);
+                }
+            }
+            Node::LetDeclList(ref list) => {
+                for let_decl in list.as_ref() {
+                    let value = match let_decl.init() {
+                        Some(v) => v.run(context).unwrap_or_else(|_| Value::undefined()),
```
jasonwilliams, line 109: This needs to be v.compile() instead of run. then you shouldn't need `context`

jasonwilliams:
```js
let a = 1;
```

generates

```
[boa/src/context.rs:761] &compiler = Compiler {
    instructions: [
        DefLet(
            0,
        ),
        One,
        InitLexical(
            0,
        ),
    ],
    pool: [
        String(
            RcString(
                "a",
            ),
        ),
    ],
}
1
```


@AnnikaCodes  if you look at the changes ive made, letDecl now doesn't use run instead it calls `compile` and that in terms puts `One` on the stack. We will need to do something similar with Var and Const.

boa/src/vm/instructions.rs
```diff
@@ -57,6 +57,11 @@ pub enum Instruction {
     Neg,
     BitNot,
     Not,
+
+    DefVar(usize),
+    DefLet(usize),
+    DefConst(usize),
+    InitLexical(usize),
```
Razican, line 64: I think these need some documentation on what the associated value is, and what do they do. We should also add documentation to the rest, but these have an extra value with them.

boa/src/vm/instructions.rs
```diff
@@ -100,6 +105,10 @@ impl std::fmt::Display for Instruction {
             Self::Neg => write!(f, "Neg"),
             Self::BitNot => write!(f, "BitNot"),
             Self::Not => write!(f, "Not"),
+            Self::DefVar(name) => write!(f, "DefVar({})", name),
+            Self::DefLet(name) => write!(f, "DefLet({})", name),
+            Self::DefConst(name) => write!(f, "DefConst({})", name),
+            Self::InitLexical(usize) => write!(f, "InitLexical({})", usize),
```
Razican, line 111: I would prefer if the binding wasn't called `usize`.

Razican (changes requested):
From my side, I think a bit of documentation is needed in the definition of the enums. Give it a look, but from the rest, I'm happy with it :)

boa/src/vm/instructions.rs
AnnikaCodes: Yeah I agree.

boa/src/vm/instructions.rs
AnnikaCodes: I added documentation for the ones in this PR

Razican (approved):

jasonwilliams (approved):
Yep looks good to me

PR #1042, opened 2021-01-05
Add receiver parameter to object internal methods

### LINKED ISSUE #1040: Object internal methods don't have Receiver parameter
A lot of object internal methods include a parameter named `Receiver` in the spec, i.e. [`[[Get]]`](https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-get-p-receiver). As an example we have the code:
```
let o = {}
let o2 = {}
Object.defineProperty(o, 't', {
  set: function(arg) {
    console.log(arg)
  },
});

Reflect.set(o, 't', 3)
Reflect.set(o, 't', 3, o2)
```
`o2.t` should remain `undefined` and `console.log(3)` should be invoked but instead it becomes `3` and `console.log` is not invoked.
This should be possible with the code in #1033 but with current implementation of `[[Get]]` it is not possible. 

`[[Get]]`: https://github.com/boa-dev/boa/blob/553ea529f357ca0b6a6ffb07e22a483483a41b66/boa/src/object/internal_methods.rs#L72-L74

`[[Set]]`: https://github.com/boa-dev/boa/blob/553ea529f357ca0b6a6ffb07e22a483483a41b66/boa/src/object/internal_methods.rs#L95-L97

Only other instances I can find in the spec are for exotic objects, which I don't think are currently implemented.

tofpie:
This Pull Request fixes #1040.

It changes the following:

- Add the `receiver` parameter `[[Get]]` and `[[Set]]` 
- Make `Array` more compliant by using `[DefineOwnProperty]]` instead of `[[Set]]` to define new indexed properties
- Add `Array` specific `[[DefineOwnProperty]]`
- Add compliant `[[SetPrototypeOf]]` (with cycle detection)
- Make `name` and `length` of functions writable


boa/src/value/display.rs
```diff
@@ -124,12 +124,9 @@ pub(crate) fn log_string_from(x: &Value, print_internals: bool, print_children:
                                 // which are part of the Array
                                 log_string_from(
                                     &v.get_own_property(&i.into())
-                                        // TODO: do this in a better way "unwrap"
-                                        .unwrap()
+                                        .and_then(|p| p.as_data_descriptor().map(|d| d.value()))
                                         // FIXME: handle accessor descriptors
```
RageKnify, line 128: Switching the order so that the comment is before the invocation like it was before.
```suggestion
                                        // FIXME: handle accessor descriptors
                                        .and_then(|p| p.as_data_descriptor().map(|d| d.value()))
```

boa/src/object/internal_methods.rs
```diff
@@ -217,6 +261,125 @@ impl GcObject {
...
+                        }
+                    }
+                }
+                Ok(true)
+            }
+            PropertyKey::Index(index) => {
+                let old_len_desc = self.get_own_property(&"length".into()).unwrap();
+                let old_len_data_desc = old_len_desc.as_data_descriptor().unwrap();
+                let old_len = old_len_data_desc.value().to_u32(context)?;
+                let index = Value::from(index).to_u32(context)?;
```
RageKnify, line 362: This line seems unnecessary, `index` was already a `u32`.

RageKnify (approved):
2 minor things, conformance looks good:

  | Passed | 24,005 | 24,308 | **+303** |
  | Failed | 38,903 | 38,600 | **-303** |
  | Panics | 90 | 90 | 0 |
  | Conformance | 30.58 | 30.97 | **+0.39%** |

RageKnify:
I'm confused, GH won't let me request reviews from @HalidOdat and @Razican so I'l tag you instead...

boa/src/object/internal_methods.rs
```diff
@@ -292,34 +455,32 @@ impl GcObject {
...
-        //             .prototype
-        //             .clone();
-        //         p = prototype;
-        //     }
-        // }
-        // self.prototype = val;
-        // true
-        todo!("Object.setPropertyOf(obj, prototype)")
+    pub fn set_prototype_of(&mut self, val: Value) -> bool {
+        debug_assert!(val.is_object() || val.is_null());
```
Razican, line 459: Should we assert this also in production builds?

Razican (approved):
Looking good for me! Check the review by @RageKnify. I just had a small question.

boa/src/object/internal_methods.rs
RageKnify: https://tc39.es/ecma262/#assert
The spec says:
> Such assertions add no additional semantic requirements and hence need not be checked by an implementation.

So I agree with this approach, I think generally we either don't do it because Rust guarantees the that the assertions are true or we can use `debug_assert!` since that will be found by tests. (Thinking about it, it won't be found by `boa_tester`, right? Since that runs in release mode.)

tofpie: Sure, it can only be detected in unit tests, as running `boa_tester` in debug mod is far too slow.

Razican: We have some places where the assertions are run in release mode too. I guess it's fine as a debug assertion though.

Razican (approved):

PR #1044, opened 2021-01-06
Removed duplicated code in vm.run using macros.

### LINKED ISSUE #1041: VM executes ge on value for gt instruction
This looks wrong:
https://github.com/boa-dev/boa/blob/0e41be114485b3492f59f48cbc28fd5a819d3e17/boa/src/vm/mod.rs#L186

because it is in the Gt branch of the match:
https://github.com/boa-dev/boa/blob/0e41be114485b3492f59f48cbc28fd5a819d3e17/boa/src/vm/mod.rs#L183

stephanemagnenat:
This Pull Request reduces code duplication in VM execution, this refactoring let me find issue #1041.

It changes the following:

- It factors out the push of the result to the stack.
- It adds a macro for binary operations.

I did not use the macro equality, in order to keep the balance with the inequality code which, due to the `!` before the `into`, would still need the macro. When using the `binop` macro, one might want to put the call in a single line instead of using `{ ... }`.

I do not know whether upstream wants this change, but I submit it just in case. I do not know if the vm feature is enabled in CI.

This PR also fixes #1041.

### DIFF
--- a/boa/src/vm/mod.rs
+++ b/boa/src/vm/mod.rs
@@ -78,162 +78,93 @@ impl<'a> VM<'a> {
 
             let _timer =
                 BoaProfiler::global().start_event(&self.instructions[self.idx].to_string(), "vm");
-            match self.instructions[self.idx] {
-                Instruction::Undefined => self.push(Value::undefined()),
-                Instruction::Null => self.push(Value::null()),
-                Instruction::True => self.push(Value::boolean(true)),
-                Instruction::False => self.push(Value::boolean(false)),
-                Instruction::Zero => self.push(Value::integer(0)),
-                Instruction::One => self.push(Value::integer(1)),
-                Instruction::Int32(i) => self.push(Value::integer(i)),
-                Instruction::Rational(r) => self.push(Value::rational(r)),
-                Instruction::String(index) => {
-                    let value = self.pool[index].clone();
-                    self.push(value)
-                }
-                Instruction::BigInt(index) => {
-                    let value = self.pool[index].clone();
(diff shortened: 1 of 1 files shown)


jasonwilliams:
@stephanemagnenat sorry for the conflict, we have a few PRs in this area at the moment

stephanemagnenat:
Do you want me to fix the conflict (by using an optional result), or is there a lot ongoing and I should rather wait that the new code is completed?

RageKnify (approved):

jasonwilliams:
@stephanemagnenat you’re free to carry on now it all went in

stephanemagnenat:
I've rebase my changes on master.

stephanemagnenat:
I was not fully sure for the naming of the macro, between `binop` (currently), and `bin_op`, the current one making vscode's spellchecking complain.

RageKnify:
> I was not fully sure for the naming of the macro, between `binop` (currently), and `bin_op`, the current one making vscode's spellchecking complain.

I agree with `code`, `bin_op` makes it easier to read as two separate words, I'd also suggest changing the name of the macro's argument from `$e` to `$op` as in **op**eration.

stephanemagnenat:
Done!

jasonwilliams (approved):
Looks good!

jasonwilliams:
Thanks for the contribution @stephanemagnenat !

jasonwilliams:
> I do not know if the vm feature is enabled in CI.

No it isn't currently but we hope to change that some point in the future

PR #1045, opened 2021-01-06
Add newTarget to construct

### LINKED ISSUE #1009: Invoking constructors changes global this instead of newTarget
**Describe the bug**
`String` used as a function call messes up with the global object.

**To Reproduce**

```javascript
String(10)
this.length
String.prototype.toString.call(this)
```

Running this code, `2` is printed as `this.length` and `"10"` is printed as `String.prototype.toString.call(this)`.

**Expected behavior**
Running this code, `this.length` should stay `undefined` and `String.prototype.toString.call(this)` should throw a `TypeError`.

- OS: Windows 10
- Version: 0.10
- Rustc version: rustc 1.48.0 (7eac88abb 2020-11-16)

**Additional context**
None

tofpie:
This Pull Request fixes #1009.

It changes the following:

- Change the `construct` method to handle `new_target`,
- Change all the built-in constructors to handle `new_target`
- Change `String` constructor to make it more compliant


Razican:
  | Passed | 24,141 | 24,187 | **+46** |
  | Failed | 38,767 | 38,721 | **-46** |
  | Panics | 90 | 90 | 0 |
  | Conformance | 30.76 | 30.81 | **+0.06%** |

boa/src/builtins/error/eval.rs
```diff
@@ -57,17 +58,27 @@ impl EvalError {
...
-        this: &Value,
+        new_target: &Value,
         args: &[Value],
         context: &mut Context,
     ) -> Result<Value> {
+        let prototype = match new_target {
+            Value::Object(obj) => match obj.get(&PROTOTYPE.into(), context)? {
+                Value::Object(ref o) => o.clone(),
+                _ => context.standard_objects().error_object().prototype(),
+            },
```
Razican, line 69: Not sure if this would be more clear:
```suggestion
            Value::Object(obj) => obj.get(&PROTOTYPE.into(), context)?.as_object().unnwrap_or_else(|| context.standard_objects().error_object().prototype()),
```

It would be applicable to the rest of the places where this happens, and we could also use it for the `new_target` match:

```rust
let prototype = new_target.as_object().map(|obj| {
    obj.get(&PROTOTYPE.into(), context)?.as_object()
}).flatten().unwrap_or_else(|| context.standard_objects().error_object().prototype());
```

This would add an extra pointer clone (which would probably be negligible and maybe optimized out). Didn't check the code, but using the [combinators](https://doc.rust-lang.org/stable/std/option/enum.Option.html) of the `Option` type many times gives a more idiomatic/clean code.

boa/src/object/gcobject.rs
```diff
@@ -199,48 +206,57 @@ impl GcObject {
...
+                        let proto = if proto.is_object() {
+                            proto
+                        } else {
+                            context
+                                .standard_objects()
+                                .object_object()
+                                .prototype()
+                                .into()
+                        };
+                        let this: Value = Object::create(proto).into();
```
Razican, line 245: I usually tend to use `From` when I need an explicit conversion to a given type. It seems more clear, and doesn't need type hints:
```suggestion
                        let this = Value::from(Object::create(proto));
```

boa/src/object/gcobject.rs
```diff
@@ -272,25 +288,23 @@ impl GcObject {
...
-            }
+            FunctionBody::BuiltInConstructor(function) => function(&new_target, args, context),
             FunctionBody::Ordinary(body) => {
                 let _ = body.run(context);
 
                 // local_env gets dropped here, its no longer needed
                 let binding = context.realm_mut().environment.get_this_binding();
                 binding.map_err(|e| e.to_error(context))
             }
+            _ => unreachable!(),
```
Razican, line 307: I would add some text in this `unreachable!()` call in case we ever mess up and end up here, in order to ease the debugging.

Razican (commented):
This is looking very good. I had some comments, though. Let me know if you think they make sense :)

boa/src/builtins/string/mod.rs
```diff
@@ -130,26 +132,48 @@ impl String {
...
         context: &mut Context,
     ) -> Result<Value> {
-        // This value is used by console.log and other routines to match Obexpecty"failed to parse argument for String method"pe
+        // This value is used by console.log and other routines to match Object type
         // to its Javascript Identifier (global constructor method name)
         let string = match args.get(0) {
+            Some(value) if value.is_symbol() && new_target.is_undefined() => {
+                Symbol::to_string(value, &[], context)?
+                    .as_string()
+                    .unwrap()
```
RageKnify, line 145:
```suggestion
                    .expect("'Symbol::to_string' returns 'Value::String'")
```

RageKnify (changes requested):
Noted some minor changes to improve readability, and minor bugs.

I'm not sure using the first argument for `new_target`when normally it would be for `this` is correct (I'd like to get @HalidOdat's opinion on this). There's `FunctionEnvironmentRecord.new_target`, but that's not really used...

RageKnify (approved):
Code looks good, still unsure about having `newTarget`be the first argument of constructors.

Razican (approved):
I'm ok having `newTarget` as first argument for now, that can be changed later if we need to or find a better solution. I'm happy to merge this :)

HalidOdat (approved):
Looks good to me! :)

PR #1048, opened 2021-01-07
VM Trace output fixes #1035

### LINKED ISSUE #1035: Improve the view for VM/Instruction output
For debugging purposes it would be useful to see what our stack and instructions look like at a particular time.
Our output is quite basic currently.

Running `let a = 1;` for example with "vm" feature set will give you:
(`cargo run --features vm ../tests/js/test.js`)
```
[boa/src/context.rs:761] &compiler = Compiler {
    instructions: [
        DefLet(
            0,
        ),
        One,
        InitLexical(
            0,
        ),
    ],
    pool: [
        String(
            RcString(
                "a",
            ),
        ),
    ],
}
```

It would be good to get some better output like the below. This would be achieved by adding some output to a buffer for each action take in the match arm here: https://github.com/boa-dev/boa/blob/master/boa/src/vm/mod.rs#L54 then printing the whole buffer when its finished.

The code for doing this should be behind a flag or an argument and not ran by default.

For timing https://doc.rust-lang.org/std/time/struct.Instant.html should do the job.

Example output below
```
$ cargo run -- --trace
> function fibo(x) { if (x<2) return 1; return fibo(x-1)+fibo(x-2)}
00020m 00000 Return                    <empty>
undefined
> fibo(3)
00009m 00000 PushInt8 3                <empty>
00015m 00002 GetValue 'fibo'           3.0
00066m 00007 Call 1                    Function
--> call function
  module_id:0 func_id:0
00007m 00000 GetValue 'x'              <empty>
00001m 00005 PushInt8 2                3.0
00013m 00007 Lt                        2.0

...

00001m 00007 Lt                        2.0
00000m 00008 JmpIfFalse 00016          true
00000m 00013 PushInt8 1                <empty>
00167m 00015 Return                    1.0
<-- return value(1.0)
  module_id:0 func_id:2
00001m 00052 Add                       1.0
00044m 00053 Return                    3.0
<-- return value(3.0)
  module_id:0 func_id:0
00000m 00012 Return                    3.0
3
> 
   |     |     |                        | 
   |     |     |                        \- value at the top of exec stack
   |     |     \-------------------------- instruction
   |     \-------------------------------- program counter
   \-------------------------------------- execution time per inst. (in microsecs)
```

jasonwilliams:
For the JS

```js
let a = 1;
let b = 2;
```

Output is:

```
VM start up time: 0μs
Time       Instr                Top Of Stack

27μs       DefLet(0)             <empty>
3μs        One                   <empty>
35μs       InitLexical(0)        1      0x7f727f41d0c0
18μs       DefLet(1)             1      0x7f727f41d0c0
4μs        Int32(2)              1      0x7f727f41d0c0
19μs       InitLexical(1)        2      0x7f727f41d0d8

Pool
0          "a" 0x7f727f41d120
1          "b" 0x7f727f41d138

Stack
0          1 0x7f727f41d0c0
1          2 0x7f727f41d0d8

2
```

Currently this is on by default.
Preferably we would pass `--trace`  to the cli to get this output, similar to other JS Cli's.

This would mean needing to add a trace option to `eval()` here https://github.com/boa-dev/boa/blob/master/boa/src/context.rs#L751


boa/src/vm/mod.rs
```diff
@@ -5,25 +5,44 @@ pub(crate) mod instructions;
...
-    pub fn new(compiler: Compiler, ctx: &'a mut Context) -> Self {
+    pub fn new(compiler: Compiler, ctx: &'a mut Context, _trace: bool) -> Self {
         Self {
             ctx,
+            idx: 0,
             instructions: compiler.instructions,
             pool: compiler.pool,
             stack: vec![],
             stack_pointer: 0,
+            is_trace: true,
```
RageKnify, line 39: I'm confused, isn't `_trace` meant to be the value used to initialize `is_trace`?

RageKnify (approved):
LGTM, just have 1 question.

boa/src/vm/mod.rs
```diff
@@ -5,25 +5,44 @@ pub(crate) mod instructions;
 
 pub use compilation::Compiler;
 pub use instructions::Instruction;
+use std::time::{Duration, Instant};
 
 // Virtual Machine.
```
Razican, line 10: We should make this a doc-comment, with 3 slashes: `///`.

boa/src/vm/mod.rs
jasonwilliams: Yes you’re right, I had some issues there I need to sort that out. I left that open because I had a question if we’re happy to have all the the new() calls have trace as an argument

boa/src/vm/mod.rs
```diff
@@ -5,25 +5,44 @@ pub(crate) mod instructions;
...
+    idx: usize,
     instructions: Vec<Instruction>,
     pool: Vec<Value>,
     stack: Vec<Value>,
     stack_pointer: usize,
+    profile: Profiler,
+    is_trace: bool,
+}
+#[derive(Debug)]
+pub struct Profiler {
```
Razican, line 23: I would add some space here between the two definitions, and add some documentation to for example explain the difference between this profiler and the general Boa profiler.

Also, do we need to modify the string below or the vectors above? If they are constant once we initialize the structure, it would be best to use `Box<[Value]>` and `Box<str>` for vectors and strings respectively, for some memory usage improvements.

boa/src/vm/mod.rs
```diff
@@ -45,12 +64,16 @@ impl<'a> VM<'a> {
...
+            if self.is_trace {
+                self.trace_print(false);
+            };
 
-        while idx < self.instructions.len() {
             let _timer =
-                BoaProfiler::global().start_event(&self.instructions[idx].to_string(), "vm");
-            match self.instructions[idx] {
+                BoaProfiler::global().start_event(&self.instructions[self.idx].to_string(), "vm");
+            match self.instructions[self.idx] {
```
Razican, line 76: Here I see we are iterating and using the index to retrieve the instruction. Couldn't we do this with iterators?

Razican (commented):
I had some comments, but overall it looks good :)

boa/src/vm/mod.rs
jasonwilliams: This has been resolved

boa/src/vm/mod.rs
jasonwilliams: I’m going to merge and come back to this

PR #1051, opened 2021-01-09
Arrow parse

### LINKED ISSUE #1032: Arrow Function Parsing fails
**ECMASCript feature**
ES6 Arrow functions

**Example code**

```javascript
let a = "Hello";
let thing = (b) => a + b;
thing(" World!");
```
The expected output is `Hello World!`.

Lan2u:
Fixes #1032 
Fixes a bug in arrow function parsing which prevented the assignment of arrow functions with a single argument using parentheses (e.g. `let foo = (a) => a` )

Arrow functions with > 1 argument or arrow functions that don't use parentheses already parse correctly

### DIFF
--- a/boa/src/syntax/parser/expression/assignment/arrow_function.rs
+++ b/boa/src/syntax/parser/expression/assignment/arrow_function.rs
@@ -71,7 +71,12 @@ where
     fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
         let _timer = BoaProfiler::global().start_event("ArrowFunction", "Parsing");
 
+        println!("Parsing arrow func");
+
         let next_token = cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?;
+
+        println!("Next token {}", next_token);
+
         let params = if let TokenKind::Punctuator(Punctuator::OpenParen) = &next_token.kind() {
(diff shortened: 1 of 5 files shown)


boa/src/syntax/parser/expression/assignment/arrow_function.rs
```diff
@@ -71,7 +71,12 @@ where
     fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
         let _timer = BoaProfiler::global().start_event("ArrowFunction", "Parsing");
 
+        println!("Parsing arrow func");
+
         let next_token = cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?;
+
+        println!("Next token {}", next_token);
+
```
Razican, line 79: It seems there are some debugging statements left.

Razican (approved):
Looks very good, thanks! Could we add also a test to check for 3 arguments? Just in case.

Also a couple of prints left.

boa/src/syntax/parser/expression/assignment/arrow_function.rs
Lan2u: Ah yes - removed

RageKnify (approved):

tofpie (approved):

PR #1066, opened 2021-01-12
Async arrow parse

### LINKED ISSUE #818: Implement `async` / `await`
**ECMASCript feature**
We currently have no support for `async` / `await` in Boa. These two keywords are used to create "promises" that will be executed asynchronously, without blocking the execution of the code that follows.

There are two types of `async` constructs:
 - Async Function Definitions: https://tc39.es/ecma262/#sec-async-function-definitions
 - Async Arrow Function Definitions: https://tc39.es/ecma262/#sec-async-arrow-function-definitions

Implementing this has to be done in two phases. First we need to be able to parse these `async` constructs, and for that, we will need new AST nodes and new parsers, that will be called when finding the `async` keyword.

After the parsing is done, we need to implement the execution. This will be a bit tricky, since we currently just execute the nodes one by one. This means we need to follow the spec in the "Runtime Semantics" for these two constructs, and implement the needed methods.

This might require further talk in #367.

Lan2u:
This Pull Request finishes the lex/parsing part of #818.

It changes the following:

- Adds async arrow function parsing


Razican:
Looking good!

A couple of things:
 - I don't think this fully closes the #818 issue, since there is no execution implemented
 - I would use a `Parse::unimplemented()` error instead of a panic for unimplemented things, that should maintain the panics to a minimum.

I will do a proper review later this week.

Lan2u:
> Looking good!
> 
> A couple of things:
> 
> * I don't think this fully closes the #818 issue, since there is no execution implemented
> * I would use a `Parse::unimplemented()` error instead of a panic for unimplemented things, that should maintain the panics to a minimum.
> 
> I will do a proper review later this week.

I put a comment on the #818 issue that it might make sense to split into multiple issues rather than one big issue but either/or is good with me.

Since this is still draft: I've left the todo!() since I plan to implement them before marking this ready

Lan2u:
I'm unsure why the test262 failures have gone up, I'd guess its to do with the lack of execution implementation but will need to look into it

Lan2u:
I've started looking at the execution stuff in the async_exec branch - I'm much less familiar with the execution side of things so I have a lot of learning to do

boa/src/syntax/parser/expression/assignment/concise_body.rs
```diff
@@ -0,0 +1,74 @@
...
+                if self.is_async {
+                    cursor.expect(Punctuator::CloseBlock, "async concise body")?;
+                } else {
+                    cursor.expect(Punctuator::CloseBlock, "concise body")?;
+                }
+
+                Ok(body)
+            }
+            _ => Ok(StatementList::from(vec![Return::new(
+                ExpressionBody::new(self.allow_in, self.is_async).parse(cursor)?,
```
Razican, line 68: Here it seems that this depends on this `self.is_async` parameter, but the spec doesn't say so.

boa/src/syntax/parser/expression/assignment/concise_body.rs
```diff
@@ -0,0 +1,74 @@
...
+where
+    R: Read,
+{
+    type Output = StatementList;
+
+    fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
+        match cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?.kind() {
+            TokenKind::Punctuator(Punctuator::OpenBlock) => {
+                let _ = cursor.next();
+                let body = FunctionBody::new(false, self.is_async).parse(cursor)?;
```
Razican, line 58: If I understand the spec correctly, both options here should be `false`.

boa/src/syntax/parser/function/mod.rs
```diff
@@ -27,30 +27,35 @@ use crate::{
 };
 use std::io::Read;
 
-/// Formal parameters parsing.
+/// UniqueFormalParameters / FormalParameters parsing.
+///
+/// Due to the similarity between these productions we use a unique flag rather than
+/// duplicating the code.
```
Razican, line 33: Actually, I would prefer having the former `FormalParameters`, and then having another `UniqueFormalParameters` structure that would basically call `FormalParameters::parse()`, and would then do a check on the generated parameter list before returning the result.

boa/src/syntax/parser/function/mod.rs
```diff
@@ -84,6 +91,14 @@ where
                 _ => FormalParameter::new(self.allow_yield, self.allow_await).parse(cursor)?,
             };
 
+            if self.unique && params.iter().any(|v| v.name() == next_param.name()) {
+                // https://tc39.es/ecma262/#prod-UniqueFormalParameters
+                return Err(ParseError::lex(LexError::Syntax(
+                    "Invalid duplicate formal parameter name".into(),
+                    pos,
+                )));
+            }
```
Razican, line 100: I don't think this should be a lexing error. It's a normal parsing error, the lexing was correct, but in this context, it's a syntax error, so a parsing error.

boa/src/syntax/parser/function/tests.rs
```diff
@@ -79,291 +75,3 @@ fn check_rest_operator() {
...
-                        None,
-                    )
-                    .into()],
-                )
-                .into(),
-            ),
-        )])
-        .into()],
-    );
-}
```
Razican, line 369: Were those tests moved?

Razican:
I'm closing this PR, as this is now very outdated. I will create issues in order to be able to re-do this again.

Razican:
> Do you think this code is relevant to implementing https://github.com/boa-dev/boa/issues/1805? I'm new to the codebase and just gathering info

Things have changed a bit, but the general idea should still be valid :) so I think it can be useful to check the diff

PR #1067, opened 2021-01-12
Refactor: Change Realm::global_object field from Value to GcObject

RageKnify:
This Pull Request works on #577.

It changes the following:

- Change Realm::global_object field from `Value` to `GcObject`
- Change `make_builin_fn`'s `parent` argument from `&Value` to `&GcObject`
- Change iterators' `create_prototype` functions to return `GcObject` rather than `Value`

### DIFF
--- a/boa/src/realm.rs
+++ b/boa/src/realm.rs
@@ -28,6 +28,7 @@ pub struct Realm {
 }
 
 impl Realm {
+    #[allow(clippy::field_reassign_with_default)]
     pub fn create() -> Self {
         let _timer = BoaProfiler::global().start_event("Realm::create", "realm");
         // Create brand new global object
--- a/boa/src/realm.rs
+++ b/boa/src/realm.rs
@@ -4,7 +4,7 @@
 //!
 //! A realm is represented in this implementation as a Realm struct with the fields specified from the spec.
 
-use crate::object::Object;
+use crate::object::{GcObject, Object, ObjectData};
 use crate::{
     environment::{
         declarative_environment_record::DeclarativeEnvironmentRecord,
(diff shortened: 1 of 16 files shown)


RageKnify:
I think the clippy lint is a false positive, I tried changing the code to respect it but it results in a `rustc` error.
Might be https://github.com/rust-lang/rust-clippy/issues/6312

boa/src/realm.rs
```diff
@@ -32,26 +32,31 @@ impl Realm {
         let _timer = BoaProfiler::global().start_event("Realm::create", "realm");
         // Create brand new global object
         // Global has no prototype to pass None to new_obj
-        let global = Value::from(Object::default());
+        let mut global = Object::default();
 
         // Allow identification of the global object easily
-        global.set_data(crate::object::ObjectData::Global);
+        global.data = ObjectData::Global;
```
tofpie, line 38: You can add attribute `#[allow(clippy::field_reassign_with_default)]` on the `create` method to make `clippy` happy.

tofpie (approved):
Looks good to me 👍

boa/src/realm.rs
Razican: Does clippy offer a better suggestion?

RageKnify: The code was triggering a clippy lint, but it seems to be a false positive, the code proposed by clippy wasn't valid.

jasonwilliams:
Sorry i think i removed this branch accidently, i was doing some cleaning up. If you have a locally copy could you pus again?

Razican (approved):
This is good to go from my side.

HalidOdat (approved):

PR #1075, opened 2021-01-17
Fix out-of-range panics of Date

### LINKED ISSUE #1054: Panic: "invalid or out-of-range date"
When running [this](https://github.com/tc39/test262/blob/main/test/annexB/built-ins/Date/prototype/setYear/time-clip.js) ECMAScript test, we get the following panic:

```
thread 'main' panicked at 'invalid or out-of-range date', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.19/src/naive/date.rs:173:51
```

This in itself is not very clear, but this comes from the [`setYear()`](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/date/mod.rs#L927-L948) function. This calls to the [`set_components()`](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/date/mod.rs#L210-L314) function which calls `NaiveDate::from_ymd()` instead of `NaiveDate::from_ymd_opt()`, so it can panic if it gets out of range.

A solution would be to change those calls and try to return `NaN` in those cases, but for it to be fully spec compliant we should implement the [`TimeClip()`](https://tc39.es/ecma262/#sec-timeclip) function, which is called from the [`setYear()`](https://tc39.es/ecma262/#sec-date.prototype.setyear) function among others.

jevancc:
This Pull Request fixes/closes #1054 

It changes the following:

- Add function `time_clip`
- Apply `time_clip` to methods of `Date`
- Add overflow checks when creating and setting `Date`

This PR resolves all known and unexplored out-of-range panics - more stable now. However, the test mentioned in the issue still failed. That is because the valid date range of `chrono::naive::NaiveDate` in Rust is smaller than the range of `Date` in JavaScript, which makes the behavior different.

### DIFF
--- a/boa/src/builtins/date/mod.rs
+++ b/boa/src/builtins/date/mod.rs
@@ -248,7 +248,8 @@ impl Date {
     ) {
         #[inline]
         fn num_days_in(year: i32, month: u32) -> Option<u32> {
-            let month = month.checked_add(1)?; // zero-based for calculations
+            let month = month + 1; // zero-based for calculations
+
             Some(
                 NaiveDate::from_ymd_opt(
                     match month {
@@ -257,7 +258,7 @@ impl Date {
                     },
                     match month {
                         12 => 1,
-                        _ => month.checked_add(1)?,
+                        _ => month + 1,
                     },
                     1,
                 )?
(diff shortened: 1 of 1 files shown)


RageKnify:
2 less panics as expected.

  | Total | 78,497 | 78,497 | 0 |
  | Passed | 24,805 | 24,805 | 0 |
  | Ignored | 15,587 | 15,587 | 0 |
  | Failed | 38,105 | 38,105 | 0 |
  | Panics | 26 | 24 | **-2** |
  | Conformance | 31.60 | 31.60 | 0.00% |

RageKnify (approved):

tofpie (approved):
Looks good to me! Thanks!

Lan2u (approved):

Issue #602, opened 2020-07-31
Indexing strings


HalidOdat:
**ECMASCript feature**
Just like arrays you can also index the characters in a string.

**Example code**
This code should now work and give the expected result:
```javascript
let x = "hello";
x[0]
```
The expected output is `'h'`. the actual output is `undefined`.

This happens because the internal slots are not set for string objects. #591

RageKnify:
I believe this was implemented in #1291, seems to be working as far as I can tell, if something is misisng we can open the issue again.

PR #1076, opened 2021-01-17
Fix internal methods

### LINKED ISSUE #591: Fix Internal Object Methods

tofpie:
This Pull Request fixes #591 and #602.

It changes the following:

- All internal methods of `GcObject` take a `&Context` and return a `Result`
- Implementation of specific internal methods for `String`
- `Context` becomes externally immutable, with only interior mutability
- `LexicalEnvironment` becomes externally immutable, with only interior mutability

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -224,7 +224,7 @@ impl Json {
                     Some(
                         replacer
                             .get_property(key, context)
-                            .unwrap()
+                            .expect("'key' is one of the keys of 'replacer'")
                             .as_ref()
                             .and_then(|p| p.as_data_descriptor())
                             .map(|d| d.value())
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -38,7 +38,7 @@ impl BuiltIn for Json {
         Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE
     }
 
-    fn init(context: &mut Context) -> (&'static str, Value, Attribute) {
+    fn init(context: &Context) -> (&'static str, Value, Attribute) {
         let _timer = BoaProfiler::global().start_event(Self::NAME, "init");
 
         let json_object = ObjectInitializer::new(context)
@@ -63,7 +63,7 @@ impl Json {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-json.parse
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
-    pub(crate) fn parse(_: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+    pub(crate) fn parse(_: &Value, args: &[Value], context: &Context) -> Result<Value> {
         let arg = args
             .get(0)
             .cloned()
@@ -94,7 +94,7 @@ impl Json {
     /// [polyfill]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
     fn walk(
         reviver: &Value,
-        context: &mut Context,
+        context: &Context,
         holder: &mut Value,
         key: &PropertyKey,
     ) -> Result<Value> {
@@ -135,7 +135,7 @@ impl Json {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-json.stringify
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
-    pub(crate) fn stringify(_: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+    pub(crate) fn stringify(_: &Value, args: &[Value], context: &Context) -> Result<Value> {
         let object = match args.get(0) {
             Some(obj) if obj.is_symbol() || obj.is_function() || obj.is_undefined() => {
                 return Ok(Value::undefined())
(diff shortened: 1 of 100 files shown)


Razican:
  | Total | 78,497 | 78,497 | 0 |
  | Passed | 24,805 | 24,857 | **+52** |
  | Ignored | 15,587 | 15,587 | 0 |
  | Failed | 38,105 | 38,053 | **-52** |
  | Panics | 26 | 26 | 0 |
  | Conformance | 31.60 | 31.67 | **+0.07%** |

boa/src/builtins/json/mod.rs
```diff
@@ -223,7 +223,8 @@ impl Json {
                 } else {
                     Some(
                         replacer
-                            .get_property(key)
+                            .get_property(key, context)
+                            .unwrap()
```
RageKnify, line 227: I think using `expect` is preferable
```suggestion
                            .expect("'key' is one of the keys of 'replacer'")```

File: boa/src/object/internal_methods.rs [hunk 2]
```diff
@@ -51,31 +73,67 @@ impl GcObject {
...
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-ordinarypreventextensions
+    #[inline]
+    pub fn ordinary_prevent_extensions(&mut self, _context: &Context) -> Result<bool> {
         self.borrow_mut().extensible = false;
-        true
+        Ok(true)
+    }
+    /// Delete a property.
```
RageKnify, line 91: Leave an empty line for clarity
```suggestion
    }
    
    /// Delete a property.
```

boa/src/object/internal_methods.rs
```diff
@@ -401,32 +535,100 @@ impl GcObject {
...
+    pub fn ordinary_own_property_keys(&self) -> Vec<PropertyKey> {
         self.borrow().keys().collect()
     }
 
+    /// OwnPropertyKeys for ordinary objects
+    ///
+    /// More information:
+    ///   - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-ordinaryownpropertykeys
```
RageKnify, line 596: The comment was copy-pasted, should mention `String Exotic Objects` instead.

RageKnify (changes requested):
Minor mistakest to fix. Can you explain the reason for:
- `Context` becomes externally immutable, with only interior mutability

Sounds, reasonable just trying to understand the reasoning.

tofpie:
All internal methods (as defined by the spec, that is `[[Get]]`, `[[Set]]` or `[[DefineOwnProperty]]` for instance) of a `GcObject` should have a parameter `&mut Context` as in some exotic objects implementation (such as `Proxy`) they can make calls to functions. If you add `&mut Context` to internal methods, then you must also add it to nearly all methods of `EnvironmentTrait` as the environment is sometimes itself an object (in `ObjectEnvironmentRecord` or `GlobalEnvironmentRecord`), but then each time the environment is modified (when declaring a variable or calling a function), then you have to borrow `Context` mutably twice, once to get the `environment` and another time when you call a method of `EnvironmentTrait`. And this is of course not allowed by the Rust compiler. And this is the same for `LexicalEnvironment`.

RageKnify (approved):

Razican (approved):

HalidOdat:
I would like to try to fix the `context`/`LexicalEnvironment` mutable borrow problem, without introducing so much interior mutability.

Hopefully I'll be done with this by tomorrow and create a PR, now that I have a little more time to spend on the project :)

Razican:
> I would like to try to fix the `context`/`LexicalEnvironment` mutable borrow problem, without introducing so much interior mutability.
> 
> Hopefully I'll be done with this by tomorrow and create a PR, now that I have a little more time to spend on the project :)

How is this going?

BTW, this needs a rebase with all the changes in master now.

HalidOdat:
> > I would like to try to fix the `context`/`LexicalEnvironment` mutable borrow problem, without introducing so much interior mutability.
> > Hopefully I'll be done with this by tomorrow and create a PR, now that I have a little more time to spend on the project :)
> 
> How is this going?

#1131 is going to fix this problem, without any additional interior mutability :)

Razican:
Now that #1131 has been merged, I think that this only needs a rebase to use the latest changes, @tofpie.

HalidOdat:
This has too many merge conflicts to try rebase this, it would be better to start over, besides most of the internal methods have already been implemented. So closing this PR.

PR #1085, opened 2021-01-20
Add primitive promotion for method calls on GetField

RageKnify:
It changes the following:

- Add primitive promotion to objects for method calls on primitives when non const `GetField` is used
- Remove a forgotten TODO comment


tofpie (approved):
This looks good to me.

Razican (approved):
Looks good, thanks!

PR #1093, opened 2021-01-24
Async Exec

Lan2u:
This Pull Request relates to async execution (async function decl, async function expr, async arrow decl).

Created as draft so I can check conformance, I've not done much before with the execution side of things so I expect much will need to change until this is merge-ready


Razican:
Note that [here](https://www.github.com/boa-dev/boa/tree/master/test_ignore.txt) we are ignoring async tests from test262, so I would expect this to require removing the async flag from the ignore list.

Lan2u:
> Note that [here](https://www.github.com/boa-dev/boa/tree/master/test_ignore.txt) we are ignoring async tests from test262, so I would expect this to require removing the async flag from the ignore list.

ah that is useful to know, thanks

Razican:
I'm closing this PR, since it's now outdated, and would need to be redone to use the VM. Feel free to re-open another PR :)

PR #1102, opened 2021-01-27
Support Unicode escape in identifier names

jevancc:
This Pull Request add the support of Unicode escapes (`\uXXXX` and `\u{ cp }`) in identifier names. An example for this is:
```javascript
let \u0078 = 10;
console.log(x); // x = 10
```

It changes the following:

- Update identifier lexer
- Throw syntax error when there are Unicode escapes in keyword
- Add True, False, Null to keywords (used only in lexing)
- Add unit tests

Some methods in `lexer/cursor.rs` are not used after this PR. Instead of removing them from the source code, I kept them for future use. The `#[allow(dead_code)]` is added to prevent clippy error.

### DIFF
--- a/boa/src/syntax/lexer/cursor.rs
+++ b/boa/src/syntax/lexer/cursor.rs
@@ -130,6 +130,7 @@ where
     /// predicate on the ascii char
     ///
     /// The buffer is not incremented.
+    #[allow(dead_code)]
     #[inline]
     pub(super) fn next_is_char_pred<F>(&mut self, pred: &F) -> io::Result<bool>
     where
@@ -191,6 +192,7 @@ where
     /// It also stops when there is no next character.
     ///
     /// Note that all characters up until the stop character are added to the buffer, including the character right before.
+    #[allow(dead_code)]
     pub(super) fn take_while_char_pred<F>(&mut self, buf: &mut Vec<u8>, pred: &F) -> io::Result<()>
     where
         F: Fn(u32) -> bool,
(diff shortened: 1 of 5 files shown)


jevancc:
This PR is ready for review. The bug related to Unicode escapes in reserved word as property is mentioned in #1117 and I prefer to address it separately. Let me know if you have any comments and suggestions.

RageKnify (approved):

HalidOdat (approved):
Looks good to me :)

boa/src/syntax/lexer/cursor.rs
```diff
@@ -130,6 +130,7 @@ where
     /// predicate on the ascii char
     ///
     /// The buffer is not incremented.
+    #[allow(dead_code)]
```
Razican, line 133: Why do we allow dead code here? If it's not being used, it should be removed, right?

Same applies to line 195.

Razican (commented):
Looks good, just a comment for my understanding on dead code.

Razican:
Would it be possible to re-base it to see conformance / benchmark changes?

Razican:
I ran this locally, and got these results:

  | Total | 78,701 | 78,701 | 0 |
  | Passed | 26,400 | 26,530 | **+130** |
  | Ignored | 15,519 | 15,519 | 0 |
  | Failed | 36,782 | 36,652 | **-130** |
  | Panics | 26 | 27 | **+1** |
  | Conformance | 33.54% | 33.71% | **+0.17%** |

PR #1103, opened 2021-01-27
Lazy evaluation for cooked template string

jevancc:
This Pull Request makes the evaluation of cooked template strings lazily, which improves the performance in some cases and fixes invalid escapes in raw template strings. The following sample is now supported
```javascript
function hello(s) { 
    return s.raw; 
}

hello`\u\x`;
```

It changes the following:

- Add struct `TemplateString` and update template related keywords
- Update tests
- Fix bug: `` `use strict`; `` can not be used to invoke strict mode


tofpie (approved):
Looks good to me! Thanks!

HalidOdat (approved):
Looks good to me! :)

RageKnify (approved):

PR #1109, opened 2021-02-02
Define all property methods of constructors

RageKnify:
This Pull Request will help work on #400.

It changes the following:

- rename `property` method in `ConstructorBuilder` to `data_property`
- rename `static_property` method in `ConstructorBuilder` to `static_data_property`
- introduce `accessor` variants similar to above functions
- introduce `property_descriptor` variants to offer more generic approach


tofpie (approved):
Looks good to me!

RageKnify:
I'm gonna try to figure out which panic was introduced and fixed it, feel free to review, just don't merge yet.

boa/src/object/mod.rs
```diff
@@ -961,6 +966,64 @@ impl<'context> ConstructorBuilder<'context> {
         self
     }
 
+    /// Add new accessor property to the constructor's prototype.
+    #[inline]
+    pub fn accessor_property<K>(
```
HalidOdat, line 971: Maybe also rename `.accessor_property()` to `.accessor()` since accessors are already properties so I don't think it's ambiguous.

HalidOdat (approved):

Issue #1092, opened 2021-01-22
Deleting elements from a map during forEach results in invalid iteration


joshwd36:
**Describe the bug**
Map iteration uses an index based approach, based on insertion order. It does this, rather than using the built-in iterator, to avoid multiple borrowing errors (#1058, #1077). However, when an element is removed from a map, all subsequent indexes are decremented and the elements shifted down a step. This means that the next element will be missed out if this deletion happens during a forEach statement.

**To Reproduce**
This JavaScript code reproduces the issue:
```javascript
let map = new Map([[0, "a"], [1, "b"], [2, "c"]]);
let index = 0;
map.forEach(function(value, key) {
    if (key === 0) {
        map.delete(0);
        map.set(3, "d");
    }
    console.log(index, key, value);
    index++;
})

```
This produces the following output:
```
0 0 a
1 2 c
2 3 d
```
This is also triggered by the [`iterates-values-deleted-then-readded`](https://github.com/tc39/test262/blob/main/test/built-ins/Map/prototype/forEach/iterates-values-deleted-then-readded.js) test.

**Expected behavior**
The correct output should be:
```
0 0 a
1 1 b
2 2 c
3 3 d
```
This is because in the [spec](https://tc39.es/ecma262/#sec-map.prototype.delete) deleting an element should actually just set the key and value to `empty`, leaving the index intact. This could be implemented by using something like the following for our implementation of OrderedMap, the backing store we use for Map:
```rust
#[derive(Hash, PartialEq, Eq)]
enum MapKey<K: Hash + Eq> {
   Key(K),
   Empty(u64), // Necessary to ensure empty keys are still unique.
}

pub struct OrderedMap<K: Hash + Eq, V, S = RandomState> {
    map: IndexMap<MapKey, Option<Value>>,
    empty_count: u64,
    valid_elements: usize // This could be stored to provide a quick way of getting the size of a map
}
```
However, this would mean that the size of a Map would never decrease unless it was cleared. I may be possible to have a method that could remove the empty elements and re-index the map, but it's unclear when it would be valid for that method to be called. Another approach could be to detect when elements are deleted during iteration, but this seems very error-prone. Other suggestions and discussions would be very welcome.

joshwd36:
[...]

PR #1111, opened 2021-02-03
Implement Set builtin object

### LINKED ISSUE #400: Implement Set()

RageKnify:
This Pull Request fixes/closes #400.

Besides **implementing `Set`** this PR also changes the following:

- Change `Value::hash` so that `1i32 == 1f64`
- Fix `Map.prototype.length` (was 1, should be 0 because the argument is optional)
- Fix msg for invoking Map without new (copied V8's message, gives more information)
- Simplify calculation of map size when displaying Maps (get size from the `OrderedMap` rather than going throught the property
- Introduce IteratorClose
- Fix switching between PropertyDescriptor kinds
- Fix deletion of accessor properties
- Add equality check for `Value::Symbol`

The change to `Value::hash` was a bandaid fix to the fact that incrementing/decrementing integers with `++`/`--` results in a float, that should still be addressed, but I think this change would still be needed.

ps: I'll leave the remaining tests failing, currently failing tests [19] can be grouped by:
- tests that expect `WeakSet` to be implemented [14]
- tests that expect correct implementation of `Set.prototype.forEach` when elements are removed (tracked in #1092 for `Map`, solution will likely be easily translated to `Set`) [2]
- test that expects `this` to be `undefined` in strict mode [1]
- tests that expect `$262` to be defined and use `$262.createRealm()` [2]
- ~~tests that expect `Reflect` to be implemented, which should pass once #1033 is merged [2]~~ (tested locally, they pass now that #1033 has been merged)


boa/src/builtins/set/set_iterator.rs
```diff
@@ -0,0 +1,149 @@
...
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-set-iterator-objects
+#[derive(Debug, Clone, Finalize, Trace)]
+pub struct SetIterator {
+    iterated_set: Value,
+    set_next_index: usize,
+    set_iteration_kind: SetIterationKind,
```
HalidOdat, line 28:
```suggestion
    next_index: usize,
    iteration_kind: SetIterationKind,
```
I think it's clearer if we don't prefix the fields with `set_` since they are already in the `SetInterator`.

boa/src/builtins/set/set_iterator.rs
```diff
@@ -0,0 +1,149 @@
...
+        set: Value,
+        kind: SetIterationKind,
+    ) -> Result<Value> {
+        let set_iterator = Value::new_object(context);
+        set_iterator.set_data(ObjectData::SetIterator(Self::new(set, kind)));
+        set_iterator
+            .as_object()
+            .expect("set iterator object")
+            .set_prototype_instance(context.iterator_prototypes().set_iterator().into());
+        Ok(set_iterator)
```
HalidOdat, line 61:
```suggestion
    ) -> Value {
        let set_iterator = Value::new_object(context);
        set_iterator.set_data(ObjectData::SetIterator(Self::new(set, kind)));
        set_iterator
            .as_object()
            .expect("set iterator object")
            .set_prototype_instance(context.iterator_prototypes().set_iterator().into());
        set_iterator
```
Since we cant throw I think its best to remove the unnecessary Result wrap.

boa/src/builtins/map/mod.rs
```diff
@@ -75,7 +74,7 @@ impl Map {
         context: &mut Context,
     ) -> Result<Value> {
         if new_target.is_undefined() {
-            return context.throw_type_error("Map requires new");
+            return context.throw_type_error("Constructor Map requires 'new'");
```
Razican, line 77: I would rephrase this a bit:

```suggestion
            return context.throw_type_error("calling the Map requires using 'new'");
```

An example of Firefox, that seems a bit more clarifying:
```
calling a builtin Map constructor without new is forbidden
```

This also requires changing the test.

boa/src/builtins/set/ordered_set.rs
```diff
@@ -0,0 +1,130 @@
...
+    V: Hash + Eq;
+
+impl<V: Eq + Hash + Trace, S: BuildHasher> Finalize for OrderedSet<V, S> {}
+unsafe impl<V: Eq + Hash + Trace, S: BuildHasher> Trace for OrderedSet<V, S> {
+    custom_trace!(this, {
+        for v in this.0.iter() {
+            mark(v);
+        }
+    });
+}
```
Razican, line 25: Since this is implementing a custom trace, I would add a comment on why this is needed and how it works.

Razican (approved):
Looks very good!! I added some suggestions, but it's good to go from my side :) congrats!

boa/src/builtins/map/mod.rs
RageKnify: I'll be changing to Firefox's style, the one I had set was Chromium's/V8's, but I agree FF's gives more context.

boa/src/builtins/set/ordered_set.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+impl<V> OrderedSet<V>
+where
+    V: Hash + Eq,
+{
+    pub fn new() -> Self {
+        OrderedSet(IndexSet::new())
+    }
```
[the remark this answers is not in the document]
RageKnify: Do you mean `OrderedSet(IndexSet::default())`? We could but I don't think we gain anything with it...

boa/src/builtins/set/mod.rs
```diff
@@ -0,0 +1,407 @@
+use crate::{
```
HalidOdat, line 1: Maybe add some documentation like other builtin global objects.

HalidOdat (approved):
Looks good to me! Jest some minor documentation things.

boa/src/builtins/set/ordered_set.rs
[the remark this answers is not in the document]
Razican: Since it's a private structure, I think it's good as-is

boa/src/builtins/set/set_iterator.rs
```diff
@@ -0,0 +1,149 @@
...
+impl SetIterator {
+    pub(crate) const NAME: &'static str = "SetIterator";
+
+    fn new(set: Value, kind: SetIterationKind) -> Self {
+        SetIterator {
+            iterated_set: set,
+            next_index: 0,
+            iteration_kind: kind,
+        }
+    }
```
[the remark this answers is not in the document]
Razican: Just something on the lines:
```
This creates a new `SetIterator`, that will start on position with index 0.
```

boa/src/builtins/set/ordered_set.rs
Razican: Hmmm maybe @HalidOdat knows about this more than me.

Razican:
The build failure is due to changes in `prettier`. A re-base should fix it (if we want, if not, I'm happy to merge it with this failure)

boa/src/builtins/set/ordered_set.rs
HalidOdat: > Hmmm maybe @HalidOdat knows about this more than me.

It is needed because `IndexSet` is does not implement `Trace`, that's why we wrap it in a struct that does implement it (`OrderedSet`). The custom trace iterates through all the values in the set and calls trace on them (the `mark(v)` part).

boa/src/builtins/set/mod.rs
```diff
@@ -0,0 +1,420 @@
...
+                .throw_type_error("calling a builtin Set constructor without new is forbidden");
+        }
+
+        // 2
+        let set_prototype = context
+            .global_object()
+            .get(&"Set".into(), context.global_object().into(), context)?
+            .get_field(PROTOTYPE, context)?
+            .as_object()
+            .expect("'Set' global property should be an object");
```
HalidOdat, line 121: we should store the constructor and prototype in the [`StandardOobjects`](https://github.com/boa-dev/boa/blob/master/boa/src/context.rs#L80-L99).

HalidOdat: then we only need to do:
```rust
let prototype = context.standard_objects().set_object().prototype();
```

boa/src/builtins/set/mod.rs
```diff
@@ -0,0 +1,420 @@
...
+        let to_string_tag = WellKnownSymbols::to_string_tag();
+
+        let values_function = FunctionBuilder::new(context, Self::values)
+            .name("values")
+            .length(0)
+            .callable(true)
+            .constructable(false)
+            .build();
+
+        let set_object = ConstructorBuilder::new(context, Self::constructor)
```
HalidOdat, line 67:
```suggestion
        let set_object = ConstructorBuilder::with_standard_object(
            context,
            Self::constructor,
            context.standard_objects().set_object().clone(),
        )
```
Here we should init the global Set object

boa/src/builtins/set/mod.rs
RageKnify: Can you explain a bit more why we want this? Is this so that people using boa can get these StandardObjects quickly from the Context instead of doing a lookup in a HashMap?

RageKnify:
The MacOS CI "failure" was just a problem on GHs side, tests have been working so far, no reason to expect them to fail now.

HalidOdat (approved):

PR #1116, opened 2021-02-04
Implement Array.of

camc:
This Pull Request checks off part of #36.

It changes the following:

- Adds Array.of static method
- Adds Array.of tests

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -99,6 +99,7 @@ impl BuiltIn for Array {
         .method(Self::entries, "entries", 0)
         // Static Methods
         .static_method(Self::is_array, "isArray", 1)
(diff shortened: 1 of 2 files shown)


RageKnify (changes requested):
Current implementation is not spec compliant, we should be able to use other constructors.
Try to follow the spec and add a comment before each step identifying it.  (Some steps can be ignored, for example `6` isn't important, you should instead iterate with `args.iter().enumerate()`).
If you need any help you can write here in the PR or talk in the Discord.

An ugly example that can be added as a test.
```
let a = Array.of.call(Date, "a", undefined, 3);
// check all of the following values
a instanceof Date;  // true
a[0]  // "a"
a[1]  // undefined
a[2]  // 3
```

PR #1126, opened 2021-02-12
Remove unnecessary wraps for non built-in functions

RageKnify:
Lint was introduced with Rust 1.50
Ignoring the lint for built-in functions because they have a fixed
signature

It changes the following:

- Remove `Result` from return types where possible
-Add `allow` clippy directives for built-in functions that since those must return `Result<Value>`

### DIFF
--- a/boa/src/builtins/array/array_iterator.rs
+++ b/boa/src/builtins/array/array_iterator.rs
@@ -49,14 +49,14 @@ impl ArrayIterator {
         context: &Context,
         array: Value,
         kind: ArrayIterationKind,
-    ) -> Result<Value> {
+    ) -> Value {
         let array_iterator = Value::new_object(context);
         array_iterator.set_data(ObjectData::ArrayIterator(Self::new(array, kind)));
         array_iterator
             .as_object()
             .expect("array iterator object")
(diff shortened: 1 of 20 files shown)


HalidOdat (commented):
Looks good!

But I think it would be better if we allow `clippy::unnecessary_wraps` this lint in the crate scope, since we have to annotate so many builtin js functions.

RageKnify:
> Looks good!
> 
> But I think it would be better if we allow `clippy::unnecessary_wraps` lint in the crate scope, since we have to annotate so many builtin js functions manually.

I feel like this is a useful lint, it helps keep functions clean and we do we have a lot of functions.
Maybe we could allow it specifically in the `buiiltins` module since that's where most of them are? What do you think of that?

HalidOdat:
> Maybe we could allow it specifically in the `buiiltins` module since that's where most of them are? What do you think of that?

Yes. this seems like the best approach, instead of global crate level :)

jasonwilliams (approved):

jasonwilliams:
Looks good to me, agree with the above, I like this lint and if we really need to avoid it we should scope that down as much as possible

HalidOdat (approved):
Looks good to me! :)

PR #634, opened 2020-08-14
Added boa examples

elasmojs:
This Pull Request fixes/closes #446 .

It changes the following:
 - root cargo.toml to add examples to the workspace
 - created examples folder with content
 - added load from string
 - added load from file
 - added handling return val using forward_val
 - added global object additions
 - added a custom module handler example

PR #1161, opened 2021-03-07
[Merged by Bors] - Added boa examples

### LINKED ISSUE #446: Add examples

jasonwilliams:
Added boa examples as per issue #446
Overtaken https://github.com/boa-dev/boa/pull/634

Somehow screwed that branch up by rebasing it and losing access
pings @elasmojs 

This Pull Request fixes/closes #446 .


RageKnify (changes requested):
Some minor things in the `Cargo.toml` files.

examples/src/loadfile.rs
```diff
@@ -0,0 +1,29 @@
...
+};
+
+pub fn run(){
+    let js_file_path = "./scripts/helloworld.js";
+    let buffer = read_to_string(js_file_path);
+
+    if buffer.is_err(){
+        println!("Error: {}", buffer.unwrap_err());
+        return;
+    }
```
RageKnify, line 16: Make it more idiomatic.
```suggestion
    if let Err(msg) = buffer {
        println!("Error: {}", msg);
        return;
    }
```

examples/src/modulehandler.rs
```diff
@@ -0,0 +1,65 @@
...
+    //Adding custom implementation that mimics 'require'
+    let requirefn = Function::builtin(Vec::new(), require);
+    ctx.global_obj.set_field("require", Value::from_func(requirefn));
+
+    //Addming custom object that mimics 'module.exports'
+    let moduleobj = Value::new_object(Some(&ctx.global_obj));
+    moduleobj.set_field("exports", Value::from(" "));
+    ctx.global_obj.set_field("module", moduleobj);
+
+    //Instantiating the engien with the execution context
```
RageKnify, line 33:
```suggestion
    //Instantiating the engine with the execution context
```

RageKnify (changes requested):
1 typo and a change to make the Rust code more idiomatic.

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "boa_wasm",
     "boa_tester",
     "boa_unicode",
+    "boa_examples",
```
RageKnify, line 8:
```suggestion
    "examples",
```
I think this line needs to have the name of the directory at the top of the repository. The other option is changing the name of the directory from `examples` to `boa_examples`.

Razican:
Hello @jasonwilliams how is this going?

jasonwilliams:
OK, from what i remember it was just the change @RageKnify pointed out.
Will try to get back on this this week

Razican:
I have updated the most simple examples, but the rest require some extra work.

boa_examples/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+    println!("\r\n");
+
+    //example that enhances the global object with custom values, objects, functions
+    enhancedglobal::run();
+    println!("\r\n");
+
+    //example that implements a custom module handler which mimics (require / module.exports) pattern
+    modulehandler::run();
+    println!("\r\n");
+}
```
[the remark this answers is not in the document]
jasonwilliams: done

RageKnify (approved):

Razican (commented):
Looks good to me, I think these examples will be useful for our users. The only thing I'm missing is a small description on the top of each example with module level comments (`!//`) explaining what the example is supposed to do.

Also, I would say we don't have an example that changes the global object, right? Do we want to have this? Is it possible with the current API?

jasonwilliams:
> Looks good to me, I think these examples will be useful for our users. The only thing I'm missing is a small description on the top of each example with module level comments (`//!`) explaining what the example is supposed to do.
> 
> Also, I would say we don't have an example that changes the global object, right? Do we want to have this? Is it possible with the current API?

How do you mean change the global object? Do you mean to add properties to it?

Razican:
> How do you mean change the global object? Do you mean to add properties to it?

I meant using "window" as a global object for example, but I guess that's just an alias, right? And then adding some attributes to it.

Razican:
@jasonwilliams do you think we could have this ready for 0.14? It would be nice to have new examples here.

Razican:
I will look at this during the week, I would love to merge it asap :)

jasonwilliams:
`gc` needed to be added as a dependency, not sure why when using boa_gc. Any ideas @Razican ?

Razican:
> `gc` needed to be added as a dependency, not sure why when using boa_gc. Any ideas @Razican ?

This is due to the way deriving works for that crate. Maybe it can be fixed with a PR to the gc crate

jasonwilliams:
@jedel1043 you happy with this?

jedel1043 (approved):
> @jedel1043 you happy with this?

Yep :)

PR #1222, opened 2021-05-04
Implement `Symbol.prototype.description` accessor

HalidOdat:
It changes the following:
- Implement `Symbol.prototype.description` accessor


RageKnify (approved):
Confusing that panics were introduced, but might be the ghost 2 that keep showing up in CI.

PR #1234, opened 2021-05-07
Separate Symbol builtin

HalidOdat:
It changes the following:
- Move non-builtin symbol stuff to `symbol` module
- Move `RcSymbol` from `value` module to `symbol` module

This is done so we don't mix the builtin Symbol stuff with the way we store Symbol. like #689


RageKnify (approved):

Razican (approved):
Looks good :) just needs a re-base.

PR #1235, opened 2021-05-08
Implement `Object.assign( target, ...sources )`

HalidOdat:


boa/src/builtins/object/mod.rs
```diff
@@ -481,4 +493,44 @@ impl Object {
...
+    /// [spec]: https://tc39.es/ecma262/#sec-object.assign
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+    pub fn assign(_: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+        let mut to = args
+            .get(0)
+            .cloned()
+            .unwrap_or_default()
+            .to_object(context)?;
+
+        if args.len() == 1 {
```
Razican, line 515: I would change this to:
```suggestion
        if args.len() <= 1 {
```

Since in the case that no argument is provided, the `to` variable will be `undefined` (which is correct), but this would be `false`, which could cause unintended behaviour.

Razican (changes requested):
I think there is a potential small bug when the function is called with no arguments, but let me know if it's not the case.

boa/src/builtins/object/mod.rs
Razican: Actually, never mind, the `to_object()` above fails with a `TypeError`, which is the expected behaviour.

Razican (approved):

PR #1243, opened 2021-05-10
Cross realm symbols

HalidOdat:
Well known Symbols are able to be used between contexts (in the same thread) interchangeably, as described in the [spec][spec] (`"Unless otherwise specified, well-known symbols values are shared by all realms"`). The spec only says this about well known symbols, but it's better to have all symbols be interchangeable rather than a few, since we will run into some hash collisions for later created symbols.

[spec]: https://tc39.es/ecma262/#sec-well-known-symbols
 
It changes the following:
- Make symbols be cross realm movable.
- Remove redundant `_symbol` suffix for well known symbol getters.
- Fix `PartialEq` for `Symbol` so it only checks equality for hashes (which are unique).
- Fix `Hash` for `Symbol` so it only hashes the `Symbol`'s hash.

The two fixes (`PartialEq` and `Hash`) might gives us a performance boots since we don't need to compare and hash the symbol description. (now we only compare and hash a u64, instead of the string description too).

### DIFF
--- a/boa/src/symbol/mod.rs
+++ b/boa/src/symbol/mod.rs
@@ -21,10 +21,20 @@ use crate::{
     gc::{Finalize, Trace},
     value::RcString,
 };
+use std::cell::Cell;
 
 pub use rcsymbol::RcSymbol;
 
 /// A structure that contains the JavaScript well known symbols.
+///
+/// # Examples
+/// ```
+///# use boa::symbol::WellKnownSymbols;
+///
+/// let iterator = WellKnownSymbols::iterator_symbol();
+/// assert_eq!(iterator.description(), Some("Symbol.iterator"));
+/// ```
+/// This is equivalent to `let iterator = Symbol.iterator` in JavaScript.
 #[derive(Debug, Clone)]
 pub struct WellKnownSymbols {
     async_iterator: RcSymbol,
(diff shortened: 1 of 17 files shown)


Razican (changes requested):
Looking very good @HalidOdat. This will be very helpful when implementing the rest of the test262 suite with the shared object that should be able to create new realms, for example.

I only added a couple of comments on documentation, and opened a question on whether we should be using the `Default` trait for an empty `new()` function, mostly related to [this](https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default) clippy lint.

boa/src/symbol/mod.rs
```diff
@@ -42,65 +55,73 @@ pub struct WellKnownSymbols {
...
+
+    /// Symbol hash.
+    ///
+    /// For now this is an incremented u64 number.
+    static SYMBOL_HASH_COUNT: Cell<u64> = Cell::new(RESERVED_SYMBOL_HASHES);
+}
+
 impl WellKnownSymbols {
-    pub(crate) fn new() -> (Self, u64) {
+    fn new() -> Self {
```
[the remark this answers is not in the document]
HalidOdat: > Maybe we should change the `new()` function and be the more idiomatic `default()` with the `Default` trait?

In this case, `new` is better since using `Default` would make it public and there should only be one instance of well known symbols.

boa/src/symbol/mod.rs
```diff
@@ -42,65 +55,78 @@ pub struct WellKnownSymbols {
     unscopables: RcSymbol,
 }
 
+/// Reserved number of symbols.
+///
+/// This is where the well known symbol live
+/// and internal engine symbols.
+const RESERVED_SYMBOL_HASHES: u64 = 128;
+
+thread_local! {
```
jasonwilliams, line 64: @HalidOdat what was the reason you went with thread_local! for holding these values? Nothing wrong with it just curious.

jasonwilliams: I ask because aren’t they supposed to be the same symbols across worker threads, web workers etc?

HalidOdat: > @HalidOdat what was the reason you went with thread_local! for holding these values? Nothing wrong with it just curious.

Well we can't pass symbols to other threads since `Symbol` is not thread safe (contains `RcString`, and store symbols in an `RcSymbol`). So I stored in a `thread_local`. We could store it in `Arc`.
But for now I think this is a good solution to detach from `Context`, we will see what we can do later.

jasonwilliams: Yup agree

PR #1081, opened 2021-01-18
Fix Array.prototype.filter

tofpie:
This Pull Request fixes issues with `Array.prototype.filter` and makes it more compliant;

It changes the following:

- Make `Array.prototype.filter` more compliant
- Add [`ArraySpeciesCreate`][spec]
- Fix `JSON` and `Math` special symbol `@@toStringTag` property

[spec]: https://tc39.es/ecma262/#sec-arrayspeciescreate

PR #1272, opened 2021-05-22
Fix Array.prototype.filter

Razican:
This is a re-base of #1081.

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -262,6 +262,51 @@ impl Array {
         Ok(array_obj_ptr)
     }
 
+    /// Utility function used to specify the creation of a new Array object using a constructor
+    /// function that is derived from original_array.
+    ///
+    /// see: <https://tc39.es/ecma262/#sec-arrayspeciescreate>
+    pub(crate) fn array_species_create(
+        original_array: &GcObject,
+        length: u32,
+        context: &mut Context,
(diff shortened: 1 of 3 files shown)


Razican:
Seems that the array creation is slower. Maybe we can improve it somehow.

boa/src/builtins/map/mod.rs
```diff
@@ -66,7 +65,7 @@ impl BuiltIn for Map {
         )
         .property(
             to_string_tag,
-            "Map",
+            Self::NAME,
```
RageKnify, line 68: For Json and Math we use the type name rather than `Self`, I'm happy with either one, but we should choose one.
```suggestion
            Map::NAME,
```

RageKnify (approved):
Minor nitpick

boa/src/builtins/map/mod.rs
Razican: But, `Self::NAME` is equivalent to `Map::NAME` here, right? since we are in the `Map` impl. Also, if we ever decide to change the name of `Map`, this would automatically continue working.

Razican: I think this will be consistent after #1276.

RageKnify: Sounds good

HalidOdat (approved):

Issue #1062, opened 2021-01-12
Panic trying to re-define the "arguments" identifier in a function


Razican:
When we try to define an identifier inside a function with the name `arguments`, this fails with a panic saying that "Identifier arguments has already been declared". You can test this with the following Test262 tests:
 - [`test/language/eval-code/direct/func-decl-fn-body-cntns-arguments-lex-bind-declare-arguments.js`](https://github.com/tc39/test262/blob/main/test/language/eval-code/direct/func-decl-fn-body-cntns-arguments-lex-bind-declare-arguments.js)
 - [`test/language/eval-code/direct/func-decl-fn-body-cntns-arguments-lex-bind-declare-arguments-and-assign.js`](https://github.com/tc39/test262/blob/main/test/language/eval-code/direct/func-decl-fn-body-cntns-arguments-lex-bind-declare-arguments-and-assign.js)
 - [`test/language/eval-code/direct/meth-fn-body-cntns-arguments-lex-bind-declare-arguments.js`](https://github.com/tc39/test262/blob/main/test/language/eval-code/direct/meth-fn-body-cntns-arguments-lex-bind-declare-arguments.js)
 - [`test/language/eval-code/direct/meth-fn-body-cntns-arguments-lex-bind-declare-arguments-and-assign.js`](https://github.com/tc39/test262/blob/main/test/language/eval-code/direct/meth-fn-body-cntns-arguments-lex-bind-declare-arguments-and-assign.js)
 - [`test/language/eval-code/direct/func-expr-fn-body-cntns-arguments-lex-bind-declare-arguments.js`](https://github.com/tc39/test262/blob/main/test/language/eval-code/direct/func-expr-fn-body-cntns-arguments-lex-bind-declare-arguments.js)
 - [`test/language/eval-code/direct/func-expr-fn-body-cntns-arguments-lex-bind-declare-arguments-and-assign.js`](https://github.com/tc39/test262/blob/main/test/language/eval-code/direct/func-expr-fn-body-cntns-arguments-lex-bind-declare-arguments-and-assign.js)
 - [`test/language/statements/function/arguments-with-arguments-lex.js`](https://github.com/tc39/test262/blob/main/test/language/statements/function/arguments-with-arguments-lex.js)
 - [`test/language/expressions/function/arguments-with-arguments-lex.js`](https://github.com/tc39/test262/blob/main/test/language/expressions/function/arguments-with-arguments-lex.js)

To run any of them, you just need to run:
```
[...]

PR #1273, opened 2021-05-22
Add default parameter support

0x7D2B:
This PR is a rewrite of #988, implementing default parameter support and merging `call` and `construct` implementations into a single internal method.

---

Modern JavaScript has support for default parameters in functions. Boa has support for the syntax but currently default parameters are ignored. For example, this code:

```javascript
function test(a, b = 2, c = a + b) {
    console.log(a, b, c);
}
test()
test(1)
```

Result produced by current Boa master:

```
undefined undefined undefined
1 undefined undefined
undefined
```

Expected result:

```
undefined 2 NaN
1 2 3
undefined
```

### DIFF
--- a/boa/src/object/gcobject.rs
+++ b/boa/src/object/gcobject.rs
@@ -202,10 +202,10 @@ impl GcObject {
                                 arguments_in_parameter_names || param.name() == "arguments";
                         }
 
-                        // An arguments object is not added in the following situations:
-                        // - In arrow functions (10.2.11.16)
-                        // - If parameter list contains `arguments` (10.2.11.17)
-                        // - If there are no default parameters and lexical names or function names contain `arguments` (10.2.11.18)
+                        // An arguments object is added when all of the following conditions are met
+                        // - If not in an arrow function (10.2.11.16)
+                        // - If the parameter list does not contain `arguments` (10.2.11.17)
+                        // - If there are default parameters or if lexical names and function names do not contain `arguments` (10.2.11.18)
                         //
                         // https://tc39.es/ecma262/#sec-functiondeclarationinstantiation
                         if !flags.is_lexical_this_mode()
--- a/boa/src/object/gcobject.rs
+++ b/boa/src/object/gcobject.rs
@@ -105,21 +105,40 @@ impl GcObject {
         std::ptr::eq(lhs.as_ref(), rhs.as_ref())
     }
 
-    /// Call this object.
+    /// Internal implementation of [`call`](#method.call) and [`construct`](#method.construct).
     ///
     /// # Panics
     ///
     /// Panics if the object is currently mutably borrowed.
-    // <https://tc39.es/ecma262/#sec-prepareforordinarycall>
-    // <https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist>
+    ///
+    /// <https://tc39.es/ecma262/#sec-prepareforordinarycall>
+    /// <https://tc39.es/ecma262/#sec-ordinarycallbindthis>
+    /// <https://tc39.es/ecma262/#sec-runtime-semantics-evaluatebody>
+    /// <https://tc39.es/ecma262/#sec-ordinarycallevaluatebody>
     #[track_caller]
-    pub fn call(&self, this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
+    fn call_construct(
+        &self,
+        this_target: &Value,
+        args: &[Value],
+        context: &mut Context,
+        construct: bool,
+    ) -> Result<Value> {
         let this_function_object = self.clone();
-        let f_body = if let Some(function) = self.borrow().as_function() {
-            if function.is_callable() {
+        let mut has_parameter_expressions = false;
+
+        let body = if let Some(function) = self.borrow().as_function() {
+            if construct && !function.is_constructable() {
+                let name = self
+                    .get(&"name".into(), self.clone().into(), context)?
+                    .display()
+                    .to_string();
+                return context.throw_type_error(format!("{} is not a constructor", name));
+            } else if !construct && !function.is_callable() {
+                return context.throw_type_error("function object is not callable");
+            } else {
                 match function {
                     Function::BuiltIn(BuiltInFunction(function), flags) => {
-                        if flags.is_constructable() {
+                        if flags.is_constructable() || construct {
                             FunctionBody::BuiltInConstructor(*function)
                         } else {
                             FunctionBody::BuiltInFunction(*function)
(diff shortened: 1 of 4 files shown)


boa/src/object/gcobject.rs
```diff
@@ -176,31 +206,109 @@ impl GcObject {
...
+                            has_parameter_expressions =
+                                has_parameter_expressions || param.init().is_some();
+
+                            let value = match args.get(i).cloned() {
+                                None | Some(Value::Undefined) => {
+                                    if let Some(init) = param.init() {
+                                        init.run(context).unwrap_or(Value::Undefined)
+                                    } else {
+                                        Value::Undefined
+                                    }
```
RageKnify, line 229: I think this can be simplified with a `map()` and `unwrap_or_default()`, something along the lines of:
`parma.init().map(|init| init.run(context)).flatten().unwrap_or_default()`

boa/src/object/gcobject.rs
```diff
@@ -176,31 +206,109 @@ impl GcObject {
...
+                            // Create a second environment when default parameter expressions are used
+                            // This prevents variables declared in the function body from being
+                            // used in default parameter initializers.
+                            // https://tc39.es/ecma262/#sec-functiondeclarationinstantiation
+                            let second_env = FunctionEnvironmentRecord::new(
+                                this_function_object,
+                                if flags.is_lexical_this_mode() {
+                                    None
+                                } else {
+                                    Some(this.clone())
```
RageKnify, line 248: Cliipy says that the `.clone()` isn't necessary.

boa/src/object/gcobject.rs
```diff
@@ -176,31 +206,109 @@ impl GcObject {
...
-            FunctionBody::BuiltInFunction(func) => func(this, args, context),
-            FunctionBody::BuiltInConstructor(func) => func(&Value::undefined(), args, context),
+        match body {
+            FunctionBody::BuiltInConstructor(function) if construct => {
+                function(&this_target, args, context)
+            }
+            FunctionBody::BuiltInFunction(_) if construct => {
+                unreachable!("Cannot have a function in construct")
+            }
+
```
RageKnify, line 277: nitpick, but this empty line looks weird.

RageKnify (changes requested):
Noted a few things that are a bit off.
I'll give a more thorough look comparing with the spec, but the numbers seem to indicate that it's great.

boa/src/object/gcobject.rs
```diff
@@ -211,111 +316,10 @@ impl GcObject {
     pub fn construct(
```
Razican, line 316: This should improve performance a bit on debug builds, since it's a 1 liner.

```suggestion
    #[inline]
    pub fn construct(
```

boa/src/object/gcobject.rs
```diff
@@ -176,31 +206,106 @@ impl GcObject {
...
 
+    /// Call this object.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the object is currently mutably borrowed.
+    // <https://tc39.es/ecma262/#sec-prepareforordinarycall>
+    // <https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist>
+    #[track_caller]
+    pub fn call(&self, this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
```
Razican, line 305: This should improve performance a bit in debug builds:

```suggestion
    #[inline]
    pub fn call(&self, this: &Value, args: &[Value], context: &mut Context) -> Result<Value> {
```

boa/src/object/gcobject.rs
```diff
@@ -211,111 +316,10 @@ impl GcObject {
     pub fn construct(
         &self,
         args: &[Value],
-        new_target: Value,
+        new_target: &Value,
```
Razican, line 319: Any reason here to pass a `Value` by reference? Wasn't it a GC value that can be cheaply cloned? Maybe @HalidOdat or @RageKnify can give some insights.

Razican (commented):
Looking good, I have a couple of questions / suggestions.

Razican:
This seems to fix #1062.

boa/src/object/gcobject.rs
RageKnify: I think it's fine to use `&Value`, should occupy less space while being passed as an argument and we can still clone when necessary.

boa/src/object/gcobject.rs
```diff
@@ -176,31 +206,106 @@ impl GcObject {
...
-        match f_body {
-            FunctionBody::BuiltInFunction(func) => func(this, args, context),
-            FunctionBody::BuiltInConstructor(func) => func(&Value::undefined(), args, context),
+        match body {
+            FunctionBody::BuiltInConstructor(function) if construct => {
+                function(&this_target, args, context)
+            }
+            FunctionBody::BuiltInFunction(_) if construct => {
+                unreachable!("Cannot have a function in construct")
+            }
```
RageKnify, line 274: Are these lines necessary? Since we know this combination (`BuiltInFuncion` and `construct`) will never happen can't we just assume it won't happen and leave only line 278 to deal with `BuiltInFuncion` bodies?

RageKnify (changes requested):
Unsure about something, other than that it looks good.

boa/src/object/gcobject.rs
0x7D2B: This change also makes it consistent with `call` which used `&Value`.

boa/src/object/gcobject.rs
0x7D2B: You're right, since the condition on `construct` on line 141 would make it always return `BuiltInConstructor`.

0x7D2B:
> This seems to fix #1062.

Only a partial fix - the issue doesn't happen when default parameters are used since `arguments` is declared in a different environment but it still happens without them.

0x7D2B:
Okay, I added arguments object handling as defined in the spec, so this should fix #1062 properly now.

boa/src/object/gcobject.rs
```diff
@@ -193,19 +193,41 @@ impl GcObject {
...
+                        // - If parameter list contains `arguments` (10.2.11.17)
+                        // - If there are no default parameters and lexical names or function names contain `arguments` (10.2.11.18)
+                        //
+                        // https://tc39.es/ecma262/#sec-functiondeclarationinstantiation
+                        if !flags.is_lexical_this_mode()
+                            && !arguments_in_parameter_names
+                            && (has_parameter_expressions
+                                || (!body.lexically_declared_names().contains("arguments")
+                                    && !body.function_declared_names().contains("arguments")))
+                        {
```
RageKnify, line 216: I think it might be a good idea to add something along the lines of "this if statement checks that none of the 3 conditions holds, and if so we add the arguments object", right now it looks a bit confusing to me that the comment describes ***when not to do something*** and then the code checks ***if we should do it***.
It's a complex expression that I think justifies commenting thoroughly.

RageKnify (changes requested):
I'd just like for the comment to be more complete to avoid confusion in the future.

boa/src/object/gcobject.rs
0x7D2B: The spec defines it as "when not to do something" and the comment was based on that. Updated to represent the code instead.

Razican (approved):

boa/src/syntax/ast/node/statement_list/mod.rs
```diff
@@ -72,6 +72,16 @@ impl StatementList {
         set
     }
 
+    pub fn function_declared_names(&self) -> HashSet<&str> {
```
Razican, line 75: I would add a bit of documentation to this function, to explain what it does.

RageKnify (approved):

PR #1276, opened 2021-05-22
Cleanup

Razican:
While doing the rebase of #1272 I found some inconsistencies, so I tried to fix them here:

 - Upgraded Rust + JS dependencies
 - Improved style of some `@@ToStringTag` definitions, using `Self::NAME` when possible
 - Updated the Test262 submodule
 - Using `Value::nan()` instead of conversions
 - Added build tests for examples

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -172,6 +172,29 @@ jobs:
           command: clippy
           args: -- --verbose
 
+  examples:
+    name: Examples
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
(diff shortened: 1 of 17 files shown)


RageKnify:
Isn't it strange that the number of passing tests went down?

RageKnify (approved):

HalidOdat (approved):

Razican:
> Isn't it strange that the number of passing tests went down?

A comparison using the same commit for Test262 shows that indeed there are 4 new failing tests. I will check what is happening. It seems that updating the Test262 they fixed some tests and this brings up the number of failures somehow.

PR #1283, opened 2021-05-25
Fix panic in regex execution

### LINKED ISSUE #1242: Regular expression execution can panic
The `Function::add_rest_param()` function can panic with the following error message:

```
thread 'main' panicked at 'Could not get slice', boa/src/builtins/regexp/mod.rs:413:56
```

This has been observed when running [this](https://github.com/tc39/test262/blob/6cf3433cf8f44c471a7e4f47e5ba36d32ebb7b69/test/built-ins/RegExp/prototype/exec/failure-g-lastindex-reset.js) ECMAScript test.

Interestingly this only happens when executing the full test suite, and not if we ask the backtrace or run the test individually.

0x7D2B:
This Pull Request fixes #1242, though that one might be accidentally triggering a different panic that's fixed in #1280


0x7D2B:
The spec actually wants ToLength and not ToIndex for lastIndex. Makes sense, right?

RageKnify (approved):
If `master` is merged into this branch we should see the panic numbers go down, right?

Razican:
> If `master` is merged into this branch we should see the panic numbers go down, right?

I think this is the panic that happens only "sometimes".

0x7D2B:
There's something non-deterministic, there was a panic that disappeared after I made a commit that only added comments. With #1280 it should fix both of the issues causing panics that I came across.

0x7D2B:
Let's see what happens to the tests.

0x7D2B:
Weird. I was getting a panic in `regress` before this.

Issue #1269, opened 2021-05-21
`Math.hypot.length` should be `2`


HalidOdat:
**Describe the bug**
`Math.hypot.length` should be `2` not `1`

**To Reproduce**
Steps to reproduce the issue, or JavaScript code that causes this failure.

This JavaScript code reproduces the issue:
```javascript
var a = Math.hypot.length;
a;
```

**Expected behavior**
Running this code, `a` should be set to `2` and printed, but `a` is instead set to `1`.

[spec](https://tc39.es/ecma262/#sec-math.hypot)

Razican:
As a hint to anyone coming to implement this, this is part of the `Math` builtin implementation :) feel free to ask questions on how this is implemented.

moadmmh:
I can fix this

PR #1288, opened 2021-05-28
Fixed hypot length bug

moadmmh:
This Pull Request fixes/closes [#1269](https://github.com/boa-dev/boa/issues/1269)

It changes the following:

- `Math.hypot.length` to be `2`


Razican (approved):
Thank you for the fix!

PR #1294, opened 2021-05-29
Don't consider panic fixes as "new failures"

Razican:
Previously, if a panic was fixed and it was now failing, it would be considered as a "new failure". This considers it a panic fix.

### DIFF
--- a/boa_tester/src/read.rs
+++ b/boa_tester/src/read.rs
@@ -103,15 +103,6 @@ pub(super) fn read_harness(test262_path: &Path) -> io::Result<Harness> {
 
 /// Reads a test suite in the given path.
 pub(super) fn read_suite(path: &Path) -> io::Result<TestSuite> {
-    use once_cell::sync::Lazy;
-    use regex::Regex;
-
-    /// Regular expression to retrieve the metadata of a test.
-    static FIXTURE_REGEX: Lazy<Regex> = Lazy::new(|| {
-        Regex::new(r#".*_FIXTURE.js(on)?"#)
-            .expect("could not compile fixture retrieval regular expression")
-    });
-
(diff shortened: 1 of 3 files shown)


Razican:
I also took the opportunity to improve a bit the code and update the Test262 suite.

Razican:
It seems that the update of the Test262 submodule is breajing this, probably due to https://github.com/tc39/test262/commit/56ca8add7dbc1efe49e5456689ddeb57ef4e296a

0x7D2B:
Can you make it so that there's a way to tell apart tests that were panicking but now pass and tests that were panicking but now fail?

Razican:
> Can you make it so that there's a way to tell apart tests that were panicking but now pass and tests that were panicking but now fail?

Something that can be done is to consider a Panic -> Passed a normal "fixed" test. In the list, it will show that it was panicking before.

0x7D2B:
I think we still want to know if a test that previously panicked now fails, since that's an improvement. Though if we get to zero panics and make sure to not merge anything that adds panics, we might not see this very often.

Razican:
> I think we still want to know if a test that previously panicked now fails, since that's an improvement. Though if we get to zero panics and make sure to not merge anything that adds panics, we might not see this very often.

Yes, the idea would be that if a "Panic" goes to "Pass", it will be considered a "new passing test", but if it goes from "panic" to "Fail", then it would be considered a "panic fix".

Razican:
This is ready for review. The extra panic seems this panic that happens only "sometimes". I'm unable to reproduce it locally.

0x7D2B (approved):

RageKnify (approved):

PR #1295, opened 2021-05-29
AST nodes now have spans

### LINKED ISSUE #300: Span nodes
It would be nice for the nodes to have a `Span` for better error messages.
A `Span` would contain a `start` `Position` and `end` `Position`.

Razican:
This Pull Request fixes/closes #300.

It changes the following:

- `Node` now has a `span`, that will start on the beginning of the first token and end on the end of the last token.
- Added the `NodeKind`enumeration, that is equivalent to the old `Node`.
- Some cleanups, that improve code by using more `Self`, among other things
- `Span` and `Position` now implement the `Default` trait. The default position is to start on line 1, character 1. A default span starts and ends on that position.
- Changed some `as_ref().to_vec()` references to `Vec` to `clone()`, that should improve performance there a bit

In general, I expect that these changes will have a negative performance impact. Nevertheless, this is needed to allow better interactivity with the parser. This is still WIP, and still doesn't build (we are still missing changes in most parsers, I'm slowly working towards it).


Razican:
I will close this, since this would require a huge rebase, and it will be easier to just start over.

PR #1302, opened 2021-06-02
fix VM branch

jasonwilliams:
The VM branch was failing to build. This uses the latest API changes made on Context.
This PR also introduces a new state to the executor, the `Error` state. This lets the VM know that an error has happened and immediately stops, leaving the error value on the top of the stack. Going forward this can be used to unwind the stack.
As the current path doesn't make use of this arm it should be backwards compatible.

### DIFF
--- a/boa/src/exec/mod.rs
+++ b/boa/src/exec/mod.rs
@@ -14,6 +14,7 @@ pub trait Executable {
 pub(crate) enum InterpreterState {
     Executing,
     Return,
+    #[cfg(feature = "vm")]
     Error,
     Break(Option<Box<str>>),
     Continue(Option<Box<str>>),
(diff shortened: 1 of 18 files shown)


Razican (changes requested):
I would add a test that at least builds this branch, to make sure it keeps working.

boa/src/exec/mod.rs
```diff
@@ -14,6 +14,7 @@ pub trait Executable {
 pub(crate) enum InterpreterState {
     Executing,
     Return,
+    Error,
```
RageKnify, line 17: Maybe we can allow the clippy lint that complains that this variant is never constructed, since it is only constructed with the `vm` feature on. Or maybe only compile it with the `vm` feature on?

RageKnify (approved):

Razican (approved):

PR #1309, opened 2021-06-08
Implement Display for function objects

### LINKED ISSUE #361: Implement Display for function objects
Calling Display (console.logging) on a function object will print its inner properties.
We need to implement Display for function objects, currently there arent any rules for function objects.

An entry would need to be made here:
https://github.com/jasonwilliams/boa/blob/master/boa/src/builtins/value/mod.rs#L728-L783

for `ObjectKind::Function`.

If the function is builtIn you can use `String() { [native code] }`
for an Ordinary function you may need to try displaying the Node inside.

You can test by adding 

```js
console.log(String);
```

then running it.

Relies on https://github.com/jasonwilliams/boa/pull/255 being merged

See also:

kvnvelasco:
This Pull Request fixes/closes #361.

Hello! 

This is my first attempt at a contribution here let me know if I'm being dumb. I decided to go ahead and stick the most naive implementation I could think of as a beginning of a discussion while I get to grips with the codebase. I also have questions.

Thanks!

It changes the following:
* Adds a special case in `Value::to_string` that handles function objects (for now)
* Adds a bunch of questions for implementation details 

Questions:
* Do all function values have a name? Right now I'm operating on the assumption that `get_field("name")` can totally fail and if it does, we proceed without a name.
* How much of the naive code in the implementation do you think should go into `impl Display for Function`?
  * Because the name of the function is located in the `Value` that contains it, not the function itself, i was thinking of exposing something like `display_function_body` instead or something. 
* How do I get at the actual backing source code that generated something? Testing this out in the browser It seems like the console.log implementation sticks the actual defined source code in there including formatting (newlines, spaces, etc).
  * Is it even sane to try to emulate that in any meaningful way?
* Why do `const x = () => {}` functions not have names?  
  
* Is there a way to get at the backing `Declr` for any of the function objects? It seems like that's the only place where arrow, expression, and function definitions are differentiated.
 
 Tested with 
 ```
 function normalFunc(arg, argb) {
    let x = 3;
}

function inline() { let x = 3 }
console.log(inline)

const arrowFunc = (arg1, arg2) => {}

console.log(normalFunc)
console.log(arrowFunc);
console.log(String)
console.log(() => {})
 ```
 
 Outputs 
 ```
 inline() {let x = 3}
normalFunc(arg, argb) {let x = 3;}
(arg1, arg2) {}
function String() {
  [native Code]
}
() {}
undefined
 ```


Razican:
  | Passed | 26,801 | 26,797 | **-4** |
  | Failed | 36,468 | 36,472 | **+4** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 33.97% | 33.96% | **-0.01%** |
  
  ```
  test/built-ins/String/S15.5.2.1_A1_T8.js [strict mode] (previously Passed)
  test/built-ins/String/S15.5.2.1_A1_T8.js (previously Passed)
  test/built-ins/String/S15.5.2.1_A1_T11.js [strict mode] (previously Passed)
  test/built-ins/String/S15.5.2.1_A1_T11.js (previously Passed)
  ```
  </details>

kvnvelasco:
Fixed 2 of the 4 broken tests.

kvnvelasco:
```
Results:
Total tests: 78897
Passed tests: 26807
Ignored tests: 15628
Failed tests: 36462 (panics: 0)
Conformance: 33.98%
```

boa/src/builtins/function/mod.rs
```diff
@@ -329,6 +329,79 @@ impl BuiltInFunctionObject {
         // TODO?: 5. PrepareForTailCall
         context.call(this, &this_arg, &arg_list)
     }
+
+    fn to_string(this: &Value, _: &[Value], context: &mut Context) -> Result<Value> {
```
Razican, line 333: You seem to not be using the `this` variable here, and it's causing a `clippy` error. You can change it to `_this` to avoid it.

boa/src/builtins/function/mod.rs
```diff
@@ -329,6 +329,79 @@ impl BuiltInFunctionObject {
...
+            if value.is_null_or_undefined() {
+                None
+            } else {
+                Some(value.to_string(context)?)
+            }
+        };
+
+        let function = {
+            let object = this.as_object().map(|object| {
+                let func = object.borrow().as_function().map(|f| f.clone());
```
Razican, line 347: Here, you are using `|f| f.clone()` as a closure, when you could use `as_function().cloned()`. This is causing a `clippy` error.

Razican (changes requested):

Razican:
> ```
> Results:
> Total tests: 78897
> Passed tests: 26807
> Ignored tests: 15628
> Failed tests: 36462 (panics: 0)
> Conformance: 33.98%
> ```

I updated te conformance. This is generated in the CI and we have to manually copy it here (until we find a better solution)

Razican (approved):

boa/src/builtins/function/mod.rs
```diff
@@ -329,6 +329,79 @@ impl BuiltInFunctionObject {
         // TODO?: 5. PrepareForTailCall
         context.call(this, &this_arg, &arg_list)
     }
+
+    #[allow(clippy::wrong_self_convention)]
+    fn to_string(this: &Value, _: &[Value], context: &mut Context) -> Result<Value> {
+        let name = {
+            // Is there a case here where if there is no name field on a value
+            // We it should default to None? Do all functions have names set?
```
RageKnify, line 337: Minor suggestion on phrasing.
```suggestion
            // Is there a case here where if there is no name field on a value
            // name should default to None? Do all functions have names set?
```
And to answer, no, not all functions have names, certain arrow functions don't have names.
As an example, with:
```js
let d = (() => {
  return () => {
    return 3.14;
  };
})
```
You get
```js
d().toString() == "() => {
  return 3.14;
}"
```

And `d().name == ""`

RageKnify (approved):
Phrasing suggestion, overall looks ok.

Issue #265, opened 2020-02-25
Support destructuring assignment


afinch7:
Support destructuring assignment syntax:
```
const object = {
  name: "text",
  size: 4,
};

const { name, size } = object;
```

vapour101:
I'd like to work on this, if that's good?

jasonwilliams:
Go for it @vapour101 !

vapour101:
Sorry, seems like I had less time to devote to this than I realised. I'm not going to be able to complete this issue.

SamuelQZQ:
@afinch7 Could I have a try?

raskad:
@SamuelQZQ Are you still working on this? I just started some basic implementation and am currently at +128 tests passing. I think I could pick this up if you are not working on it.


@davimiku you referenced #1324, that seems to implement shorthand syntax for [Object Initializers](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-object-initializer) right? I think this issue would be also about [Destructuring Binding Patterns](https://tc39.es/ecma262/#sec-destructuring-binding-patterns). Just to make sure I'm not workin on something you already started.

davimiku:
@raskad That's correct, #1324 only implements initializing objects with the short-hand notation (on the "right-hand side"), i.e.

```js
const message = "hello"
const obj = { message }
```

That PR didn't cover destructuring assignment (on the "left-hand side"), so great that you've started! I didn't start / wasn't planning on implementing destructuring assignment. 

Maybe if you don't hear from SamuelQZQ try asking a maintainer to change the assignment?

SamuelQZQ:
@raskad Go ahead. I gave up. Haha.

PR #1324, opened 2021-06-15
Implement parsing for object shorthand properties

### LINKED ISSUE #1308: Support for shorthand object literals
**Describe the bug**
The following code doesn't compile:

```js
const message = 'Hello world!';
const wrapper = { message };
```

**Expected behavior**
It should work as an equivalent of `const wrapper = { message: message };`

**Build environment:**
Boa Playground

davimiku:
This Pull Request fixes/closes #1308.

It changes the following:

- Adds to the PropertyDefinition parser to look for object shorthand properties

This takes a shortcut from [the spec](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-object-initializer), it skips the use of `IdentifierReference` from the Object Initializer portion of the spec and instead parses `prop` as if it was `prop: prop` (`PropertyName/AssignmentExpression` pair).

This allows the change to only occur in the parser, as we [currently don't have an evaluation](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/ast/node/object/mod.rs#L91) for `IdentifierReference`. The [evaluation semantics](https://tc39.es/ecma262/#sec-runtime-semantics-propertydefinitionevaluation) for `IdentifierReference` appear to be a subset of the evaluation semantics for a `PropertyName/AssignmentExpression` pair, so this shortcut is likely OK, but could be worth others checking/thinking about that.

```js
const message = "hello"
const obj1 = { message }   // parsed as if it was { message: message }
const obj2 = { message, }  // parsed as if it was { message: message }
```

### DIFF
--- a/boa/src/syntax/parser/expression/primary/object_initializer/mod.rs
+++ b/boa/src/syntax/parser/expression/primary/object_initializer/mod.rs
@@ -9,6 +9,7 @@
 
 #[cfg(test)]
 mod tests;
+use crate::syntax::ast::node::Identifier;
 use crate::syntax::lexer::TokenKind;
 use crate::{
     syntax::{
@@ -132,6 +133,27 @@ where
             let node = AssignmentExpression::new(true, self.allow_yield, self.allow_await)
                 .parse(cursor)?;
             return Ok(node::PropertyDefinition::SpreadObject(node));
+        }
+
+        // Peek for '}' or ',' to indicate shorthand property name
+        if let Some(next_token) = cursor.peek(1)? {
+            match next_token.kind() {
+                TokenKind::Punctuator(Punctuator::CloseBlock)
+                | TokenKind::Punctuator(Punctuator::Comma) => {
+                    let token = cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?;
+                    if let TokenKind::Identifier(ident) = token.kind() {
+                        // ident is both the name and value in a shorthand property
+                        let name = ident.to_string();
+                        let value = Identifier::from(ident.to_owned());
+                        cursor.next()?.expect("token vanished"); // Consume the token.
+                        return Ok(node::PropertyDefinition::property(name, value));
+                    } else {
+                        // Anything besides an identifier is a syntax error
+                        return Err(ParseError::unexpected(token.clone(), "object literal"));
+                    }
+                }
+                _ => {}
+            }
         }
 
         let prop_name = cursor.next()?.ok_or(ParseError::AbruptEnd)?.to_string();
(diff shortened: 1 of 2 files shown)


Razican:
  | Passed | 26,857 | 26,859 | **+2** |
  | Failed | 36,412 | 36,410 | **-2** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 34.04% | 34.04% | **+0.00%** |
  
  ```
  test/language/expressions/object/__proto__-permitted-dup-shorthand.js [strict mode] (previously Failed)
  test/language/expressions/object/__proto__-permitted-dup-shorthand.js (previously Failed)
  test/language/expressions/object/prop-def-id-get-error.js [strict mode] (previously Failed)
  test/language/expressions/object/prop-def-id-get-error.js (previously Failed)
  test/language/expressions/object/not-defined.js [strict mode] (previously Failed)
  test/language/expressions/object/not-defined.js (previously Failed)
  test/language/expressions/object/prop-def-id-valid.js [strict mode] (previously Failed)
  test/language/expressions/object/prop-def-id-valid.js (previously Failed)
  ```
  </details>
  
  ```
  test/language/expressions/assignment/dstr/obj-rest-not-last-element-invalid.js [strict mode] (previously Passed)
  test/language/expressions/assignment/dstr/obj-rest-not-last-element-invalid.js (previously Passed)
  test/language/statements/for-of/dstr/obj-rest-not-last-element-invalid.js [strict mode] (previously Passed)
  test/language/statements/for-of/dstr/obj-rest-not-last-element-invalid.js (previously Passed)
  test/language/statements/for-in/dstr/obj-rest-not-last-element-invalid.js [strict mode] (previously Passed)
  test/language/statements/for-in/dstr/obj-rest-not-last-element-invalid.js (previously Passed)
  ```
  </details>

davimiku:
Thanks for the test results, I need to check and fix those regressions.

davimiku:
After further review of the failed tests, in context of the regression I actually think it's good to proceed to code review. To be more precise, an example of a test that previously passed but now fails:

```js
const obj = { a: 10, b: 20, c: 30 };
const { ...rest, a } = obj;
```
(this is paraphrased from the actual test-262, which uses weirdly formatted Javascript)

In Javascript, this should be a SyntaxError, because a rest parameter must be the last parameter in the destructuring statement. Before my PR, this was a SyntaxError in Boa, after my PR this is not a SyntaxError so the test fails.

Boa hasn't yet implemented [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) ([spec](https://tc39.es/ecma262/#prod-ObjectAssignmentPattern)) in the parser (covered in #265 ), so for these tests it was basically undefined behavior either way.

For that reason, I'll move this back to Ready from Draft, hopefully that makes sense.

RageKnify (approved):

Razican (approved):
Looks very good :) thanks for the changes!

PR #1329, opened 2021-06-16
Allow Value.set_field to throw

raskad:
While working on some RegExp functions I noticed, that some tests currently fail because `Value.set_field` always silently no-ops if the property is not writable. For example [match](https://tc39.es/ecma262/#sec-regexp.prototype-@@match) sets the `lastIndex` property multiple times (6.c and 6.f.iii.3.c). It uses the abstract [7.3.4 Set ( O, P, V, Throw )](https://tc39.es/ecma262/#sec-set-o-p-v-throw) operation that takes a boolean argument `throw` and should throw if the property is not writable.

I adjusted the `Value.set_field` function to make it somewhat similar to `7.3.4 Set ( O, P, V, Throw )`. I adjusted all uses of the function to match the spec, if it could be referenced. For uses without clear spec reference, I just set throw to false, so that the current behavior is not impacted.

I'm not sure if changing `Value.set_field` is the best way forward, but I think an internal function that matches the implementation of `7.3.4 Set ( O, P, V, Throw )` is needed.

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -155,12 +155,12 @@ impl Array {
 
         if !length.is_number() {
             array.set_property(0, DataDescriptor::new(length, Attribute::all()));
-            array.set_field("length", 1, context)?;
+            array.set_field("length", 1, true, context)?;
         } else {
             if length.is_double() {
                 return context.throw_range_error("Invalid array length");
             }
-            array.set_field("length", length.to_u32(context).unwrap(), context)?;
+            array.set_field("length", length.to_u32(context).unwrap(), true, context)?;
         }
 
         Ok(array)
@@ -325,6 +325,7 @@ impl Array {
         array_ptr.set_field(
             "length",
             Value::from(orig_length.wrapping_add(add_values.len())),
+            false,
             context,
         )?;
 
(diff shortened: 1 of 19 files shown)


Razican:
  | Passed | 26,865 | 26,871 | **+6** |
  | Failed | 36,404 | 36,398 | **-6** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 34.05% | 34.06% | **+0.01%** |
  
  ```
  test/built-ins/RegExp/prototype/exec/y-fail-lastindex-no-write.js [strict mode] (previously Failed)
  test/built-ins/RegExp/prototype/exec/y-fail-lastindex-no-write.js (previously Failed)
  test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-init-err.js [strict mode] (previously Failed)
  test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-init-err.js (previously Failed)
  test/built-ins/RegExp/prototype/test/y-fail-lastindex-no-write.js [strict mode] (previously Failed)
  test/built-ins/RegExp/prototype/test/y-fail-lastindex-no-write.js (previously Failed)
  ```
  </details>

RageKnify (approved):
Seems right to me, I think most of this will change with separation of Value and Object, but it's still nice to have this to fix the tests while that isn't done.

Razican (approved):

PR #1330, opened 2021-06-16
Split intrinsics

### LINKED ISSUE #1320: [perf] - Split intrinisics into min and fullFat groups
The profile below is from the code:
```js
const a = 'hello'
const b = 'world';
a + b;
```

image

The above image shows (on VM path):
*  6.5ms on `create_intrinsics`
* 42μs microseconds lexing
* 0.44ms parsing
* 0.60ms code generation
* 63μs execution

We spend some considerable time putting the intrinsics together.  These are mainly the global objects available to use, we do this before lexing/parsing and executing primarily because parsing errors can throw a [SyntaxError](https://github.com/boa-dev/boa/blob/f8d0b66daf92c52c187c43eb3b3e18fbe71b4da0/boa/src/builtins/error/syntax.rs#L24) object.

As overloading isn't possible during parsing stage one idea is to load minimal intrinsics at the beginning, then parse, if parsing is successful load the rest of the intriniscs (group 2) before execution. 
There was the risk that SyntaxError could be overloaded, but as execution hasn't happened yet this wouldn't be possible at this point.

Another risk is [public API offerings](https://github.com/boa-dev/boa/issues/445). If a user wants to override via the API I believe we can add these in via the "minIntrinisics group" (as we don't know ahead of time what they're overriding). This means that by default we don't need to load these globals.

jasonwilliams:
See issue for full detail.

Split intrinsics into min and remaining, then only load min before executing.
* `create_full_intrinisics` is an idempotent property on context. It will only add intrinsics once even if called multiple times.
* Utility functions such as `eval`, `forward`, `forward_val` will inject full intriniscs once syntactical analysis passes

fixes #1320 

Before:
image

After:
image

### DIFF
--- a/boa/src/builtins/date/mod.rs
+++ b/boa/src/builtins/date/mod.rs
@@ -93,7 +93,7 @@ pub struct Date(Option<NaiveDateTime>);
 
 impl Display for Date {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self.to_local() {
+        match self.as_local() {
             Some(v) => write!(f, "{}", v.format("%a %b %d %Y %H:%M:%S GMT%:z")),
             _ => write!(f, "Invalid Date"),
         }
@@ -176,14 +176,14 @@ impl BuiltIn for Date {
             .method(setter_method!(set_utc_minutes(0, 1, 2)), "setUTCMinutes", 1)
             .method(setter_method!(set_utc_month(0, 1)), "setUTCMonth", 1)
             .method(setter_method!(set_utc_seconds(0, 1)), "setUTCSeconds", 1)
-            .method(getter_method!(to_date_string), "toDateString", 0)
-            .method(getter_method!(to_gmt_string), "toGMTString", 0)
-            .method(getter_method!(to_iso_string), "toISOString", 0)
-            .method(getter_method!(to_json), "toJSON", 0)
+            .method(getter_method!(as_date_string), "toDateString", 0)
+            .method(getter_method!(as_gmt_string), "toGMTString", 0)
+            .method(getter_method!(as_iso_string), "toISOString", 0)
+            .method(getter_method!(as_json), "toJSON", 0)
             // Locale strings
             .method(getter_method!(to_string), "toString", 0)
-            .method(getter_method!(to_time_string), "toTimeString", 0)
-            .method(getter_method!(to_utc_string), "toUTCString", 0)
+            .method(getter_method!(as_time_string), "toTimeString", 0)
+            .method(getter_method!(as_utc_string), "toUTCString", 0)
             .method(getter_method!(value_of), "valueOf", 0)
             .static_method(Self::now, "now", 0)
             .static_method(Self::parse, "parse", 1)
@@ -314,9 +314,9 @@ impl Date {
         }
 
         let naive = if utc {
-            self.to_utc().map(|dt| dt.naive_utc())
+            self.as_utc().map(|dt| dt.naive_utc())
         } else {
-            self.to_local().map(|dt| dt.naive_local())
+            self.as_local().map(|dt| dt.naive_local())
         };
 
         self.0 = naive.and_then(|naive| {
@@ -539,7 +539,7 @@ impl Date {
     /// [spec]: https://tc39.es/ecma262/#sec-date.prototype.getdate
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate
     pub fn get_date(&self) -> f64 {
-        self.to_local().map_or(f64::NAN, |dt| dt.day() as f64)
+        self.as_local().map_or(f64::NAN, |dt| dt.day() as f64)
     }
 
     /// `Date.prototype.getDay()`
@@ -554,7 +554,7 @@ impl Date {
     /// [spec]: https://tc39.es/ecma262/#sec-date.prototype.getday
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay
     pub fn get_day(&self) -> f64 {
-        self.to_local().map_or(f64::NAN, |dt| {
+        self.as_local().map_or(f64::NAN, |dt| {
             let weekday = dt.weekday() as u32;
             let weekday = (weekday + 1) % 7; // 0 represents Monday in Chrono
             weekday as f64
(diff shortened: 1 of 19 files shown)


boa/src/context.rs
```diff
@@ -294,10 +300,21 @@ impl Context {
...
+        let _timer = BoaProfiler::global().start_event("create_min_intrinsics", "interpreter");
+        // Create minimal intrinsics, enough for the interpreter itself to work, add global objects here
+        builtins::init_min(self);
+    }
+
+    /// Sets up the default global objects within Global
+    #[inline]
+    pub fn create_full_intrinsics(&mut self) {
+        let _timer = BoaProfiler::global().start_event("create_full_intrinsics", "interpreter");
+        // Create minimal intrinsics, enough for the interpreter itself to work, add global objects here
```
RageKnify, line 313: This comment should mention full intrinsics instead, currently it's a copy of line 305.

RageKnify (approved):
LGTM

boa/src/builtins/date/mod.rs
```diff
@@ -217,15 +217,15 @@ impl Date {
     /// Converts the `Date` to a local `DateTime`.
     ///
     /// If the `Date` is invalid (i.e. NAN), this function will return `None`.
-    pub fn to_local(&self) -> Option<DateTime<Local>> {
+    pub fn as_local(&self) -> Option<DateTime<Local>> {
```
Razican, line 220: As a convention, we are keeping the original JS name in Rust too. In this case, `to_local()`. This gives some `clippy` errors/warnings, but we are just adding `#[allow(clippy::lint_name)]` for these cases. Either we do the same here, or we change it everywhere else, but we should be consistent.

Razican (commented):
I like the idea behind the PR, and it looks overall very good. The only thing I would change were the methods where we changed the name from the convention (JS name in snake case) to a different one just to avoid clippy. I would say we should keep the convention and ignore the lint.

boa/src/builtins/date/mod.rs
jasonwilliams: I’m fully open to that, I can set ignores here

Razican: That would be my preferred option too :)

RageKnify:
I think you might  have broken some things while following clippy.

  | Passed | 26,873 | 26,837 | **-36** |
  | Failed | 36,396 | 36,432 | **+36** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 34.06% | 34.02% | **-0.05%** |
  
  ```
  test/built-ins/Array/length/define-own-prop-length-overflow-realm.js [strict mode] (previously Passed)
  test/built-ins/Array/length/define-own-prop-length-overflow-realm.js (previously Passed)
  test/built-ins/Array/prototype/concat/create-proto-from-ctor-realm-array.js [strict mode] (previously Passed)
  test/built-ins/Array/prototype/concat/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/Array/prototype/slice/create-proto-from-ctor-realm-array.js [strict mode] (previously Passed)
  test/built-ins/Array/prototype/slice/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/Array/prototype/map/create-proto-from-ctor-realm-array.js [strict mode] (previously Passed)
  test/built-ins/Array/prototype/map/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/Symbol/replace/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/replace/cross-realm.js (previously Passed)
  test/built-ins/Symbol/hasInstance/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/hasInstance/cross-realm.js (previously Passed)
  test/built-ins/Symbol/search/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/search/cross-realm.js (previously Passed)
  test/built-ins/Symbol/split/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/split/cross-realm.js (previously Passed)
  test/built-ins/Symbol/isConcatSpreadable/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/isConcatSpreadable/cross-realm.js (previously Passed)
  test/built-ins/Symbol/iterator/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/iterator/cross-realm.js (previously Passed)
  test/built-ins/Symbol/matchAll/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/matchAll/cross-realm.js (previously Passed)
  test/built-ins/Symbol/toStringTag/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/toStringTag/cross-realm.js (previously Passed)
  test/built-ins/Symbol/match/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/match/cross-realm.js (previously Passed)
  test/built-ins/Symbol/toPrimitive/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/toPrimitive/cross-realm.js (previously Passed)
  test/built-ins/Symbol/unscopables/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/unscopables/cross-realm.js (previously Passed)
  test/built-ins/Symbol/asyncIterator/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/asyncIterator/cross-realm.js (previously Passed)
  test/built-ins/Symbol/species/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/Symbol/species/cross-realm.js (previously Passed)
  test/built-ins/BigInt/prototype/valueOf/cross-realm.js [strict mode] (previously Passed)
  test/built-ins/BigInt/prototype/valueOf/cross-realm.js (previously Passed)
  ```
  </details>

Razican:
It seems that many of the broken tests are broken cross-realm symbols.

jasonwilliams:
> I think you might have broken some things while following clippy.
> 
> ### Test262 conformance changes:
> Test result	master count	PR count	difference
> Total	78,897	78,897	0
> Passed	26,873	26,837	**-36**
> Ignored	15,628	15,628	0
> Failed	36,396	36,432	**+36**
> Panics	0	0	0
> Conformance	34.06%	34.02%	**-0.05%**
> Broken tests:

Where did you see that? I don't see it in the actions output

HalidOdat:
I don't think we should split `Context::new()` since when a user creates a context all objects should be initialized for a JS context. If the user wants to only check if it parses or if he wants to create the context after it parses he can call `parse` method and execute the AST (with `.run(&mut context)`). (we could have a `eval_ast` function too)

We could have a function for creating a context with minimal objects, but that behaviour should not be the default (`Context::new/default`). (maybe something like a `ContextBuilder`)

jasonwilliams:
OK lets close this then, maybe we can optimize in other ways

PR #1333, opened 2021-06-17
adds condition to avoid triggers from dependabot

### LINKED ISSUE #1325: Don't trigger documentation/benchmarks upload if user is dependabot
Workflows triggered by Dependabot PRs will run with [read-only permissions](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/#:~:text=February%2019%2C%202021-,GitHub%20Actions%3A%20Workflows%20triggered%20by%20Dependabot%20PRs,run%20with%20read%2Donly%20permissions&text=This%20means%20they%20will%20receive,to%20the%20repository%20to%20fail.)

We should set out workflow file to not bother uploading anything when the master-merge was by dependabot. 
We should be able to use an if statement in the build block. Similar to https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#determining-when-to-use-contexts 

Maybe `${{ github.actor !== "dependabot"}}` or something

neeldug:
This Pull Request fixes/closes #1325.

It changes the following:

- Avoid documentation/benchmark master workflow if dependabot triggered

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   benchmark:
-    if: ${{ github.actor != "dependabot" }}
+    if: ${{ github.actor != "dependabot[bot]" }}
     name: Upload docs and run benchmarks
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   benchmark:
+    if: ${{ github.actor != "dependabot" }}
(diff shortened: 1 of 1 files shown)


Razican (approved):
Looks good to me, thanks!

jasonwilliams:
Looks good, i think it also tries to upload documentation too so we need it there

jasonwilliams (approved):

neeldug:
@jasonwilliams @Razican may need to be single quoted since GH actions failed on last master workflow.

PR #1339, opened 2021-06-18
Scripts should not be considered in a block

macmv:
This allows scripts like these to compile:
```js
"use strict";

function Hello() {}
```

### DIFF
--- a/boa/src/syntax/parser/function/mod.rs
+++ b/boa/src/syntax/parser/function/mod.rs
@@ -12,10 +12,7 @@ mod tests;
 
 use crate::{
     syntax::{
-        ast::{
-            node::{self},
-            Punctuator,
-        },
+        ast::{node, Punctuator},
         lexer::{InputElement, TokenKind},
         parser::{
             expression::Initializer,
@@ -25,7 +22,7 @@ use crate::{
     },
     BoaProfiler,
 };
-use std::io::Read;
+use std::{collections::HashSet, io::Read};
 
 /// Formal parameters parsing.
 ///
@@ -66,6 +63,7 @@ where
         cursor.set_goal(InputElement::RegExp);
 
         let mut params = Vec::new();
+        let mut param_names = HashSet::new();
 
         if cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?.kind()
             == &TokenKind::Punctuator(Punctuator::CloseParen)
@@ -76,6 +74,7 @@ where
         loop {
             let mut rest_param = false;
 
+            let position = cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?.span().start();
             let next_param = match cursor.peek(0)? {
                 Some(tok) if tok.kind() == &TokenKind::Punctuator(Punctuator::Spread) => {
                     rest_param = true;
@@ -83,7 +82,11 @@ where
                 }
                 _ => FormalParameter::new(self.allow_yield, self.allow_await).parse(cursor)?,
             };
+            if param_names.contains(next_param.name()) {
+                return Err(ParseError::general("duplicate parameter name", position));
+            }
 
+            param_names.insert(next_param.name().to_string());
             params.push(next_param);
 
             if cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?.kind()
--- a/boa/src/syntax/parser/function/mod.rs
+++ b/boa/src/syntax/parser/function/mod.rs
@@ -86,7 +86,7 @@ where
                 return Err(ParseError::general("duplicate parameter name", position));
             }
 
-            param_names.insert(next_param.name().to_string());
+            param_names.insert(Box::from(next_param.name()));
             params.push(next_param);
 
             if cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?.kind()
--- a/boa/src/syntax/parser/function/tests.rs
+++ b/boa/src/syntax/parser/function/tests.rs
@@ -4,7 +4,7 @@ use crate::syntax::{
         Identifier, Node, Return,
     },
     ast::op::NumOp,
-    parser::tests::check_parser,
+    parser::{tests::check_parser, Parser},
 };
 
 /// Checks basic function declaration parsing.
@@ -19,6 +19,16 @@ fn check_basic() {
         )
         .into()],
     );
+}
+
+/// Checks for duplicate parameter names.
+#[test]
+fn check_duplicates() {
+    let js = "function foo(a, a) {}";
+
+    let res = Parser::new(js.as_bytes(), false).parse_all();
+    dbg!(&res);
+    assert!(res.is_err());
 }
 
 /// Checks basic function declaration parsing with automatic semicolon insertion.
--- a/boa/src/syntax/parser/mod.rs
+++ b/boa/src/syntax/parser/mod.rs
@@ -153,6 +153,6 @@ where
     type Output = StatementList;
 
     fn parse(self, cursor: &mut Cursor<R>) -> Result<Self::Output, ParseError> {
-        self::statement::StatementList::new(false, false, false, true, &[]).parse(cursor)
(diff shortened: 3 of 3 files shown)


macmv:
Huh. This test failed:
```

    let scenario = r#"
    'use strict';
    function f(a, b, b) {}
    "#;

    let string = dbg!(forward(&mut context, scenario));

    assert!(string.starts_with("Uncaught \"SyntaxError\": "));
```
And it was passing before because functions were not allowed at all in strict mode. Fixing this now.

Razican:
  | Passed | 26,865 | 26,842 | **-23** |
  | Failed | 36,404 | 36,427 | **+23** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 34.05% | 34.02% | **-0.03%** |
  
  ```
  test/built-ins/Number/S9.3.1_A3_T2.js [strict mode] (previously Failed)
  test/built-ins/Number/S9.3.1_A3_T2.js (previously Failed)
  test/built-ins/Number/S9.3.1_A3_T1.js [strict mode] (previously Failed)
  test/built-ins/Number/S9.3.1_A3_T1.js (previously Failed)
  ```
  </details>
  
  ```
  test/language/expressions/logical-assignment/lgcl-or-assignment-operator-non-simple-lhs.js [strict mode] (previously Passed)
  test/language/expressions/logical-assignment/lgcl-nullish-assignment-operator-non-simple-lhs.js [strict mode] (previously Passed)
  test/language/expressions/logical-assignment/lgcl-and-assignment-operator-non-simple-lhs.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/parenthesized-lefthandsideexpression-assignment-assignmentexpression-1.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/direct-callexpression-arguments.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/parenthesized-primaryexpression-objectliteral.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/direct-lefthandsideexpression-assignment-assignmentexpression-1.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/direct-lefthandsideexpression-assignment-assignmentexpression-0.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/direct-lefthandsideexpression-assignment-assignmentexpression-2.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/parenthesized-lefthandsideexpression-assignment-assignmentexpression-2.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/parenthesized-callexpression-arguments.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/direct-arrowfunction-1.js [strict mode] (previously Passed)
  test/language/expressions/assignmenttargettype/parenthesized-lefthandsideexpression-assignment-assignmentexpression-0.js [strict mode] (previously Passed)
  test/language/expressions/assignment/target-assignment-inside-function.js [strict mode] (previously Passed)
  test/language/statements/async-function/dflt-params-duplicates.js [strict mode] (previously Passed)
  test/language/statements/async-function/early-errors-declaration-duplicate-parameters.js [strict mode] (previously Passed)
  test/language/statements/async-function/rest-param-strict-body.js [strict mode] (previously Passed)
  test/language/statements/async-function/early-errors-declaration-NSPL-with-USD.js [strict mode] (previously Passed)
  test/language/statements/function/param-duplicated-strict-2.js [strict mode] (previously Passed)
  test/language/statements/function/param-duplicated-strict-3.js [strict mode] (previously Passed)
  test/language/statements/function/param-duplicated-strict-1.js [strict mode] (previously Passed)
  test/language/statements/function/dflt-params-duplicates.js [strict mode] (previously Passed)
  test/language/statements/function/rest-param-strict-body.js [strict mode] (previously Passed)
  test/language/statements/function/13.1-5gs.js [strict mode] (previously Passed)
  test/language/statements/function/use-strict-with-non-simple-param.js [strict mode] (previously Passed)
  test/built-ins/Number/prototype/toExponential/this-is-0-fractiondigits-is-0.js [strict mode] (previously Passed)
  test/built-ins/Number/prototype/toExponential/this-is-0-fractiondigits-is-0.js (previously Passed)
  ```
  </details>

macmv:
I cannot believe that broke so many tests. I will do my best to fix some of them, but no promises.

macmv:
After looking over a few of these, I don't think it will be easy to make any of them pass. Most of them expected to fail, were in strict mode, and also happened to define a function. But now that they aren't failing, we get lower coverage.

Razican:
> After looking over a few of these, I don't think it will be easy to make any of them pass. Most of them expected to fail, were in strict mode, and also happened to define a function. But now that they aren't failing, we get lower coverage.

It does happen sometimes that some tests are passing because they are supposed to fail when parsing, and they were failing on parsing due to a different error. No worries, if the implementation is correct, we will merge it, even if it uncovers some hidden errors :)

boa/src/syntax/parser/function/mod.rs
```diff
@@ -76,14 +74,19 @@ where
...
+            let position = cursor.peek(0)?.ok_or(ParseError::AbruptEnd)?.span().start();
             let next_param = match cursor.peek(0)? {
                 Some(tok) if tok.kind() == &TokenKind::Punctuator(Punctuator::Spread) => {
                     rest_param = true;
                     FunctionRestParameter::new(self.allow_yield, self.allow_await).parse(cursor)?
                 }
                 _ => FormalParameter::new(self.allow_yield, self.allow_await).parse(cursor)?,
             };
+            if param_names.contains(next_param.name()) {
+                return Err(ParseError::general("duplicate parameter name", position));
```
Razican, line 86: Could we add the parameter name in the error message? That way it's a bit more explicit.

macmv:
> It does happen sometimes that some tests are passing because they are supposed to fail when parsing, and they were failing on parsing due to a different error. No worries, if the implementation is correct, we will merge it, even if it uncovers some hidden errors :)

Sounds good to me :+1:

Razican (approved):
Looks good to me! Check my comment in case it makes sense :)

boa/src/syntax/parser/function/mod.rs
macmv: I wanted to, but the `general` function requires a `&'static str`, which I cannot make from a `format!` macro. We would need to change the `ParseError` type to take an `Into<Box<str>>` in order for this to work. I'm not sure if that is within the scope of this PR.

Razican: Let's leave it like this for now then :)

boa/src/syntax/parser/function/mod.rs
```diff
@@ -76,14 +74,19 @@ where
...
+            }
 
+            param_names.insert(next_param.name().to_string());
```
RageKnify, line 89: This is creating a `HashSet<String>`, right? Is it possible to do it with just `&str` or similar? It would be nice to minimize allocations.

RageKnify (approved):
Looks good to me, just suggested a possible optimization, but I'm not sure it's possible.

boa/src/syntax/parser/function/mod.rs
Razican: Ideally, at some point we want to intern all these strings, but for now this is not possible. The only optimization that could be done reasonably fast would be to change the `Stri g` for a `Box<str>`.

Razican (approved):

boa/src/syntax/parser/function/mod.rs
RageKnify: I guess that still minimizes the space used by the HashSet...

macmv: I'll change this to a `Box<str>`. I couldn't think of a better way to check for duplicate names (except maybe storing their hashes?).

PR #1342, opened 2021-06-18
start execution context

### LINKED ISSUE #186: Create execution context and execution stack
This is a bit of a refactor.
https://tc39.es/ecma262/#sec-execution-contexts

First we need a context struct which holds information about the context.
fields would be https://tc39.es/ecma262/#table-22

Then we need to create a stack structure, this holds execution contexts in a FIFO order, and the top most one is the currently running context

This stack structure would also have methods on it:
8.3.1GetActiveScriptOrModule
8.3.2ResolveBinding
8.3.3GetThisEnvironment
8.3.4ResolveThisBinding
8.3.5GetNewTarget
8.3.6GetGlobalObject

Then the exec.rs would accept a new executionContext and run it
https://github.com/jasonwilliams/boa/blob/master/src/lib/exec.rs#L24-L29
 
**Note**
Multiple contexts can point to the same realm, so realm needs to be wrapped in a GC

jasonwilliams:
Currently the VM directly manipulates a global stack.
This PR will move to execution contexts which hold their own stack values and will manipulate those instead. The VM will then call on each execution context as each function is invoked. 
These will live on the execution context stack.

This should unblock us from `eval`, `stack traces`, `implementing functions in the VM` etc

* We may need to rename `Context` to `AgentContext` to disassociate it with an Execution Context, which is the context surrounding a running function

fixes #186

### DIFF
--- a/boa/src/vm/exec_context.rs
+++ b/boa/src/vm/exec_context.rs
@@ -9,14 +9,15 @@
 
 use crate::Value;
 
+#[derive(Debug)]
 pub struct ExecContext {
-    /// The stack will hold values as the execution context runs
+    /// The stack will hold values as the execution context runs.
     /// As these values are only temporarily held onto the stack and not accessed
     /// anywhere else they don't need to be GC'd
-    pub stack: Vec<Value>,
+    stack: Vec<Value>,
     /// Points to where in the instructions this context should start executing
     /// As the instructions already exist on another stack this only needs to reference the index.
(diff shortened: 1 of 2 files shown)


Razican:
@jasonwilliams this seems to require a rebase and a rustfmt & clippy check :) should it land in 0.13?

jasonwilliams:
This can be closed

PR #1343, opened 2021-06-19
Parallel tester

macmv:
With a quick search, I couldn't find an issue for this.

It changes the following:
- Adds `rayon` as a dependency of `boa_tester` (not sure if this is OK).
- Runs all tests in parallel

If we don't want to add `rayon` as a dependency, I can implement this by manually spawning threads.


macmv:
I screwed up git. Ignore this PR please.

PR #1338, opened 2021-06-17
fix cargo check errors

neeldug:
This Pull Request fixes/closes #1335.

It changes the following:

- moves all f64 constants NAN or Infinity to the associated version
- Parallelizes the Test262 suite

PR #1344, opened 2021-06-19
Parallel tests

macmv:
With a quick search, I couldn't find an issue for this.

It changes the following:
- Adds `rayon` as a dependency of `boa_tester` (not sure if this is OK).
- Runs all tests in parallel

If we don't want to add `rayon` as a dependency, I can implement this by manually spawning threads.


jedel1043:
Already closed by #1338.

Razican:
Thank you both for your contributions!

PR #1357, opened 2021-06-22
Add Example to Execute a Function of a Script File

schrieveslaach:
This commit adds an example code snippet that demonstrates the usage of Boa when someone wants to execute a simple function that should be embedded into an application that support scripting extensions.

For example, extension through scripting is considered in https://github.com/aixigo/PREvant/issues/41

### DIFF
--- a/boa/src/context.rs
+++ b/boa/src/context.rs
@@ -252,7 +252,7 @@ impl StandardObjects {
 ///
 /// let value = context.eval("test(arg)").unwrap();
 ///
-/// assert_eq!(value, boa::value::Value::Integer(12))
+/// assert_eq!(value.as_number(), Some(12.0))
 /// ```
 #[derive(Debug)]
 pub struct Context {
--- a/boa/src/context.rs
+++ b/boa/src/context.rs
@@ -222,6 +222,38 @@ impl StandardObjects {
 /// `Context`s constructed in a thread share the same runtime, therefore it
 /// is possible to share objects from one context to another context, but they
 /// have to be in the same thread.
+///
+/// # Examples
+///
+/// ## Execute Function of Script File
+///
+/// ```rust
+/// use boa::{Context, object::ObjectInitializer, property::Attribute};
+///
+/// let script = r#"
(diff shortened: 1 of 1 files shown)


schrieveslaach:
There were some formatting issues. I fixed them and updated the PR.

boa/src/context.rs
```diff
@@ -222,6 +222,39 @@ impl StandardObjects {
...
+///     }
+///     return 112233;
+/// }
+/// "#;
+///
+/// let statements = boa::parse(script, false).unwrap();
+///
+/// let mut context = Context::new();
+/// // Populate the script definition to the context.
+/// statements.run(&mut context);
```
HalidOdat, line 246:
```suggestion
/// use boa::{Context, object::ObjectInitializer, property::Attribute};
///
/// let script = r#"
/// function test(arg1) {
///     if(arg1 != null) {
///         return arg1.x;
///     }
///     return 112233;
/// }
/// "#;
///
/// // Populate the script definition to the context.
/// context.eval(script).unwrap();
```
we can call `eval` here too this will do the samething.

schrieveslaach: Thanks for pointing out. I'll fix this.

boa/src/context.rs
```diff
@@ -222,6 +222,38 @@ impl StandardObjects {
...
+///
+/// // Create an object that can be used in eval calls.
+/// let arg = ObjectInitializer::new(&mut context)
+///     .property("x", 12, Attribute::READONLY)
+///     .build();
+/// context.register_global_property("arg", arg, Attribute::all());
+///
+/// let value = context.eval("test(arg)").unwrap();
+///
+/// assert_eq!(value, boa::value::Value::Integer(12))
```
HalidOdat, line 255:
```suggestion
/// assert_eq!(value.as_number(), Some(12.0))
```

Better to use the the `as_` methods.

HalidOdat (commented):
Besides the small suggestion this looks good to merge :)

HalidOdat (approved):

PR #1361, opened 2021-06-24
Redesign bytecode virtual machine

HalidOdat:
The previous virtual machine was not really a bytecode vm, since did not interpret bytecode, rather it interpreted and array of fat enums, while this is easier to implement is has bad space efficiency (also bad performance), one instruction was 32 bytes (on 64-bit machine) and consequently all other instructions where 32 bytes.
So for this code:
```
PushOne
pop
```
It would take 64 bytes. But now with an actual bytecode vm we only need 2 bytes.

Also implemented some other stuff:
 - Jumping opcodes
 - `If-else`
 - `while`
 - `do-while`
 - `break`
 - `continue`
 - labeled loops
 - short-circit operators (`&&`, `||`, `??`)
 - Variable assignment
 - Object property getting and setting
 - Binary assign operators
 - throwing
 - conditional operator (`?:`)
 - Array expression (`[1, 2, 3, 4]`)
 - `this`
 - `instanceof` operator


jasonwilliams:
Looks good so far, and yes i was hoping to reduce the bytecodes down to actual codes rather than fat enums. Thanks for refactoring this.

HalidOdat:
This is ready to review/merge :)

I wanted to add more stuff like function/calls and catching error (both require a lot of change to the current execution scheme, call frames and stack unwinding). But I already added a lot in this PR, so I will leave this for separate PRs after this gets merged.

Hopefully in the near future we will completely replace the AST walker executor.

boa/src/syntax/ast/node/mod.rs
```diff
@@ -60,6 +60,7 @@ use std::{
 #[cfg(feature = "deser")]
 use serde::{Deserialize, Serialize};
 
+// TODO: This should be split into Expression and Statement.
```
Razican, line 63: I would like to know more about this, given I'm working on changing everything about a  `Node` with spans.

I'm also thinking on creating a `LabelledStatement` node, and it's execution would simply push/pop a label in a "label stack" in the context.

docs/vm.md
```diff
@@ -14,7 +14,7 @@ You need to enable this via a feature flag. If using VSCode you can run `Cargo R
 
 ## Understanding the output
 
-Once set up you should you can try some very simple javascript in your test file. For example:
+Once set up you should you can try some simple javascript in your test file. For example:
```
Razican, line 17: Typo: "you should you can"

boa/src/vm/opcode.rs
```diff
@@ -0,0 +1,614 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct InvalidOpcodeError {
+    value: u8,
+}
+
+impl std::fmt::Display for InvalidOpcodeError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "invalid opcode: {}", self.value)
```
Razican, line 594: I would format this in hexadecimal.

boa/src/vm/opcode.rs
```diff
@@ -0,0 +1,614 @@
...
+/// The opcodes of the vm.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u8)]
+pub enum Opcode {
+	/// Pop the top value from the stack.
+	///
+	/// Operands:
+	///
+	/// Stack: value **=>**
+    Pop = 0,
```
Razican, line 12: Why is `Pop` `0` but the rest doesn't have a number assigned?

boa/src/vm/opcode.rs
```diff
@@ -0,0 +1,614 @@
...
+    ///
+    /// Stack: **=>** `this`
+    This,
+
+    /// No-operation instruction, does nothing.
+    ///
+    /// Operands:
+    ///
+    /// Stack: **=>**
+    // Safety: Must be last in the list since, we use this for range checking
```
Razican, line 499: I understand this, but if we ever add more opcodes, it would make it backwards incompatible.

Maybe it would be better to assign a number to each and have a `TryFrom` implementation that checks a hardcoded number.

boa/src/vm/opcode.rs
jasonwilliams: https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations Basically it will force Rust to set the remaining variants to auto-increment the number from that point onwards.

However, their docs do say that if a number isn’t specified it will apply 0 by default, so the `= 0` may be unnecessary unless it’s supposed to be indicative

HalidOdat: As @jasonwilliams said it's not necessary, just wanted to make sure it started at `0`. I'll remove it :)

boa/src/vm/opcode.rs
HalidOdat: > I understand this, but if we ever add more opcodes, it would make it backwards incompatible.

Could you explain what you meant by this? We can add opcodes wherever we want as long as `Nop` is the last. we can even shuffle then and it will still work.

> Maybe it would be better to assign a number to each and have a `TryFrom` implementation that checks a hardcoded number.

But if we did this we would have to maintain the current opcode number (like Pop is `0`, Push is `1` etc) and adding opcodes we would have to increase this number manually, and shufulling them would be very difficult since it would require updating all the preceding hardcoded opcode values. (we may want to do this to group some opcodes by behaviour like `GetPropertyByName`, `GetPropertyByValue`, `SetPropertyByName` `SetPropertyByValue` that access properties of objects).

> Another option would be to hardcode one number per opcode and then have a constant that would be the "maximum opcode". We would need a test to make sure that we update it if we ever add an opcode.

Even with this it requires some maintenance cost like the previous solution.

boa/src/syntax/ast/node/mod.rs
HalidOdat: > I would like to know more about this, given I'm working on changing everything about a `Node` with spans.

Currently `Node` is both a expression and a statement, it would be better if `Node` was split into how the spec describes it `Expression` , `Statement` (and also `Declaration`). Why do we want this? Right now we don't differentiate between statements (like `if`, `while`, etc) and expressions (like `2 + 2`), while we could make it work like this, it makes somethings awkward like `if (cond) {}` `cond` should **only** and expression, and when we parse it we check for this, but some checks can be elided if we leverage the type system, making illegal state unrepresentable.

The AST should look something like this https://astexplorer.net/ (turn )

Razican: Makes sense, test but maybe in a different PR

boa/src/vm/opcode.rs
HalidOdat: You mean something like [v8 snapshot](https://github.com/danbev/learning-v8/blob/master/notes/snapshots.md)? that we can restore later, right?

I don't think they could be compatible between versions since we can ass more opcodes that the previous versions would not understand.

boa/src/syntax/ast/node/mod.rs
HalidOdat: Created issue for this https://github.com/boa-dev/boa/issues/1371 :)

boa/src/vm/opcode.rs
HalidOdat: Created issue for start up snapshots #1372

Razican: > You mean something like [v8 snapshot](https://github.com/danbev/learning-v8/blob/master/notes/snapshots.md)? that we can restore later, right?

Actually, I was thinking more of a Python cache. If you are running the same JS script multiple times, you can parse and compile it once, save the instructions to a file, and just execute the output instead of re-compiling every time.

boa/src/bytecompiler.rs
```diff
@@ -0,0 +1,674 @@
...
+        }
+
+        let name: RcString = name.into();
+        let index = self.code_block.names.len() as u32;
+        self.code_block.names.push(name.clone());
+        self.names_map.insert(name, index);
+        index
+    }
+
+    fn next_opcode_location(&mut self) -> u32 {
```
Razican, line 117: I would also `#[inline]` this, could bring some optimization on the bounds checking.

Razican (approved):
Looks very good :) I think this will improve how we build the VM. Dividing statements, expressions and so on in `Node` should improve the big matches.

jasonwilliams:
Will take a look today/tomorrow hopefully

jasonwilliams (approved):

HalidOdat:
Lets merge this, so other features are not blocked :)

Issue #1376, opened 2021-06-29
`Js` prefix for JavaScript rust types


HalidOdat:
`JsValue` <= `Value` (#1457)
`JsString` <= `RcString` (#1380)
`JsBigInt` <= `RcBigInt` (#1408)
`JsSymbol` <= `RcSymbol` (#1382)
`JsObject` <= `GcObject` (#1489)
`JsResult` <= `boa::Result` (#1493)

raskad:
I think that would be a great change to make things more clear.

bartlomieju:
@HalidOdat is changing `Value` to `JsValue` still desired? I could look into it, looks like fun with so many occurrences.

HalidOdat:
> @HalidOdat is changing `Value` to `JsValue` still desired? I could look into it, looks like fun with so many occurrences.

Yes. I'll change in it and create a new PR, I was doing this and #1373 at the same time that's why I did not create a PR for it. But I think it's best to rename it in a separate PR instead of doing both in one PR.

jedel1043:
Is there any other type we would want to rename?

Razican:
> Is there any other type we would want to rename?

I think it's done :)

Issue #1373, opened 2021-06-29
NaN boxed JavaScript Value


HalidOdat:
Currently a JavaScript value in Boa is defined as a fat enum. With NaN boxing we can reduce this to `8` bytes having all the variants of a JS value fit in `u64`/`f64`. With this we can reduce the memory footprint of Boa, also a JS value will be able to be put into a 64bit register.

Many JS engines use this, like spidermonkey, JavaScriptCore, etc.

For more information:
	- https://anniecherkaev.com/the-secret-life-of-nan
	- https://leonardschuetz.ch/blog/nan-boxing/
	- https://github.com/mozilla/gecko-dev/blob/master/js/public/Value.h

neeldug:
I ended up reading through the links and can see the benefit in doing NaN boxing, however, what is the memory footprint of the fat enum being used currently? Not totally sure how or whether Rust does any sort of padding or whatnot on fat enums.

HalidOdat:
> I ended up reading through the links and can see the benefit in doing NaN boxing, however, what is the memory footprint of the fat enum being used currently? Not totally sure how or whether Rust does any sort of padding or whatnot on fat enums.

Currently the size of `Value` is `24` bytes (`size_of::<Value>()`).

playXE:
If you're looking for impl of NaN boxing in Rust Starlight has two different implementations of them for handling JS values: <https://github.com/Starlight-JS/starlight/blob/dev/crates/starlight/src/vm/value.rs> :)

HalidOdat:
> If you're looking for impl of NaN boxing in Rust Starlight has two different implementations of them for handling JS values: [https://github.com/Starlight-JS/starlight/blob/dev/crates/starlight/src/vm/value.rs](https://github.com/Starlight-JS/starlight/blob/dev/crates/starlight/src/vm/value.rs?rgh-link-date=2021-07-07T06%3A42%3A40Z) :)

Awesome! thanks, for the suggestion. I have implemented NaN boxing locally, but the problem is refactoring the existing `match value` stuff (which we use a lot, this generates a couple of hundred of erros :sweat_smile: )

playXE:
What about having `ValueVariant` enum which essentially can hold all the values NaN-boxed value can encode and then add `JsValue::variant()` which returns this ValueVariant. I'm pretty sure Rust and LLVM are able to optimize this.

HalidOdat:
[...]

PR #1380, opened 2021-06-30
Custom JavaScript string rust type

### LINKED ISSUE #1377: Custom JavaScript string rust type
The current JavaScript string rust (`RcString` which holds a `Rc<str>`) is 16 bytes this is because it stores the pointer to the data and length on the stack. And this prevents #1373 from being implemented. So we need to write a custom string type which is referenced counted like `Rc<str>` and holds the length on the heap.

I should almost be done with this, and I'll create a PR soon :)

HalidOdat:
This Pull Request fixes/closes #1377 .

As per #1376 `RcString` has been renamed to `JsString`.

This is similar to `Rc<str>`. But unlike `Rc<str>` which stores the length on the stack and a pointer to the data The `JsString` length and data is stored on the heap. (`Rc<str>` keeps refcount on the heap anyway) and just an non-null pointer is kept on the stack, so its size is the size of a pointer.

Also added `JsString::concat(x, y)` function, so we don't have to do `JsString::new(format!("{}{}", x, y))` which wastes an allocation.

Moved `src/value/rcstring.rs` **=>** `src/string.rs` (like `symbol.rs`)
Reexport `JsString` to root level (because its such a common import, like `Value`)

This should unblock #1373 :)


boa/src/value/rcstring.rs
```diff
@@ -1,114 +1,448 @@
...
 
-unsafe impl Trace for RcString {
-    empty_trace!();
+    /// The number of references to the string.
+    ///
+    /// When this reaches `0` the string is deallocated.
+    refcount: Cell<usize>,
+
+    /// An empty array which is used to get the offset of string data.
+    data: [u8; 0],
```
Razican, line 24: Could we use something like [`NonNull`](https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html) for this?

HalidOdat: But we aren't pointing to any other allocation, it's only one allocation for len, refcount and string data, and size of `NonNull` is size of pointer while size of `[u8; 0]` is zero, we just use it to get the offset of the string bytes. by calling `data.as_ptr()`. This pattern is also used in C/C++ [Flexible array member](https://en.wikipedia.org/wiki/Flexible_array_member)

HalidOdat:
There seems to be an overall nice increase in performance :)

neeldug:
@HalidOdat just wondering why there should be this strong of a performance improvement, as I'd expect the memory usage to decrease, but due to heap allocations and deallocations being slower than stack shouldn't it have the adverse effect on the runtime?

HalidOdat:
> @HalidOdat just wondering why there should be this strong of a performance improvement, as I'd expect the memory usage to decrease, but due to heap allocations and deallocations being slower than stack shouldn't it have the adverse effect on the runtime?

`JsString::concat` saves one allocation so that's probably why string concatenation improved Besides that we don't really allocate/deallocate anymore than we did previously. In the benchmarks we can see that the main improvement is in arithmetic operations this is most likely do to the smaller size of `Value` (from `24` to `16`), there are probably other places that have smaller size now, like `PropertyKey`, `Symbol`, string property entries, etc.

PR #1382, opened 2021-06-30
Refactor JavaScript symbol rust type

HalidOdat:
Changes:
- Rename `Symbol` struct to `Inner`, since that is what it is the inner data of a JavaScript symbol. And it doesn't really make sense to have two symbol types.
- Rename `RcSymbol` => `JsSymbol` (#1376)
- Removed `Context::construct_symbol(description)` since it's simpler to call `JsSymbol::new(description)`.
- Reexport `JsSymbol` to root level.


boa/src/symbol.rs
```diff
@@ -264,34 +264,79 @@ impl Symbol {
...
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.inner.description {
+            Some(desc) => write!(f, "Symbol({})", desc),
+            None => write!(f, "Symbol()"),
+        }
+    }
+}
+
+impl Eq for JsSymbol {}
```
Razican, line 314: I think this can be derived.

Razican (approved):
Looks great! Thanks!

boa/src/symbol.rs
HalidOdat: Yes, I'll change it :)

PR #1384, opened 2021-07-01
fix(boa): bitwise not operation

neeldug:
It changes the following:

- follows spec more rigorously
- todo: work out why hex bigint fails

### DIFF
--- a/boa/src/syntax/ast/node/operator/unary_op/mod.rs
+++ b/boa/src/syntax/ast/node/operator/unary_op/mod.rs
@@ -76,13 +76,20 @@ impl Executable for UnaryOp {
             }
             op::UnaryOp::Not => self.target().run(context)?.not(context)?.into(),
             op::UnaryOp::Tilde => {
-                let num_v_a = self.target().run(context)?.to_number(context)?;
-                Value::from(if num_v_a.is_nan() {
-                    -1
+                let num_v_a = self.target().run(context)?.to_numeric_number(context)?;
+                if num_v_a.is_nan() {
+                    Value::from(-1)
+                } else if num_v_a.is_infinite() {
+                    Value::from(-1)
                 } else {
-                    // TODO: this is not spec compliant.
-                    !(num_v_a as i32)
-                })
+                    if self.target.run(context)?.is_bigint() {
+                        Value::from(Value::from(-num_v_a-1f64).to_bigint(context)?)
+                    }
+                    else {
+                        let temp = (num_v_a as i64) & 0x00000000ffffffff;
+                        Value::from(!(temp as i32))
+                    }
+                }
             }
             op::UnaryOp::Void => {
                 self.target().run(context)?;
--- a/boa/src/syntax/ast/node/operator/unary_op/mod.rs
+++ b/boa/src/syntax/ast/node/operator/unary_op/mod.rs
@@ -2,10 +2,11 @@ use crate::{
     exec::Executable,
     gc::{Finalize, Trace},
     syntax::ast::{node::Node, op},
-    Context, JsBigInt, Result, Value,
+    Context, Result, Value,
 };
 use std::fmt;
 
+use crate::value::Numeric;
 #[cfg(feature = "deser")]
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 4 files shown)


neeldug:
  | Passed | 26,990 | 27,000 | **+10** |
  | Failed | 36,291 | 36,281 | **-10** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 34.21% | 34.22% | **+0.01%** |
  
  ```
  test/language/expressions/bitwise-not/S9.5_A2.2_T2.js [strict mode] (previously Failed)
  test/language/expressions/bitwise-not/S9.5_A2.2_T2.js (previously Failed)
  test/language/expressions/bitwise-not/bigint.js [strict mode] (previously Failed)
  test/language/expressions/bitwise-not/bigint.js (previously Failed)
  test/language/expressions/bitwise-not/S11.4.8_A3_T2.js [strict mode] (previously Failed)
  test/language/expressions/bitwise-not/S11.4.8_A3_T2.js (previously Failed)
  test/language/expressions/bitwise-not/S9.5_A2.1_T2.js [strict mode] (previously Failed)
  test/language/expressions/bitwise-not/S9.5_A2.1_T2.js (previously Failed)
  test/language/expressions/bitwise-not/S9.5_A2.3_T2.js [strict mode] (previously Failed)
  test/language/expressions/bitwise-not/S9.5_A2.3_T2.js (previously Failed)
  ```
  </details>
  
  Currently failing tcs are bigint-primitive and ones involving eval

Razican (approved):
Thanks for the fix!

boa/src/syntax/ast/node/operator/unary_op/mod.rs
```diff
@@ -76,13 +76,16 @@ impl Executable for UnaryOp {
...
+                } else if let Some(num_bigint) = self.target.run(context)?.as_bigint() {
+                    Value::from(JsBigInt::not(num_bigint))
+                // add bigint support
                 } else {
-                    // TODO: this is not spec compliant.
-                    !(num_v_a as i32)
-                })
+                    let masked = (num_v_a as i64) & 0x00000000ffffffff; // converts float to i32 following spec to ignore MSB using mask
+                    Value::from(!(masked as i32)) // Nots i32 conversion and creates value from this
+                }
```
HalidOdat, line 88:
```
Let expr be the result of evaluating UnaryExpression.
2. Let oldValue be ? ToNumeric(? GetValue(expr)).
3. Let T be Type(oldValue).
4. Return ! T::bitwiseNOT(oldValue).
```

We should call `to_numeric` instead of `to_numeric_number` (which is a specialization of `to_numeric` for performace reasons). And then match on the numeric structure.

for bigint match arm we should call `JsBigInt::not()`.
And for number arm we should have a associated function [`Number::not(x: f64)`](https://tc39.es/ecma262/#sec-numeric-types-number-bitwiseNOT)
```
let x = f64_to_int32(x);
return !x;
```

boa/src/syntax/ast/node/operator/unary_op/mod.rs
```diff
@@ -76,13 +76,16 @@ impl Executable for UnaryOp {
             }
             op::UnaryOp::Not => self.target().run(context)?.not(context)?.into(),
             op::UnaryOp::Tilde => {
-                let num_v_a = self.target().run(context)?.to_number(context)?;
-                Value::from(if num_v_a.is_nan() {
-                    -1
+                let num_v_a = self.target().run(context)?.to_numeric_number(context)?;
+                if num_v_a.is_nan() || num_v_a.is_infinite() {
+                    Value::from(-1) // special case for inf or nan
+                } else if let Some(num_bigint) = self.target.run(context)?.as_bigint() {
```
HalidOdat, line 82: We should not call `self.target().run(context)?` multiple times since the expression could be a function with side effects like:
```js
let x = 0;
function y() {
	x += 1;
	return 1;
}

~(y());
```

boa/src/value/mod.rs
```diff
@@ -1051,3 +1051,16 @@ impl From<Numeric> for Value {
...
+    pub fn not(x: &Self) -> Value {
+        match x {
+            Numeric::Number(x) => {
+                let x = f64_to_int32(*x);
+                Value::from(!x)
+            }
+            Numeric::BigInt(x) => Value::from(JsBigInt::new(!x.as_inner())),
+        }
+    }
+}
```
HalidOdat, line 1066: I think it's best if we have separate associative function `not()` one on `JsBigInt` (like we had before) and one on `crate::builtins::Number`.

neeldug: Just wondering why that is, I figured we're only going to call not on the whole Numeric type at this point so surely matching within Numeric makes sense here?

HalidOdat: > Just wondering why that is, I figured we're only going to call not on the whole Numeric type at this point so surely matching within Numeric makes sense here?

For two reasons:
 - To comply with what the spec is doing, so we can more easily see what is going on.
 - API reasons we would like for the user to able to call `not()` on `JsBigInt` and on `Number` from the rust side.

We shouldn't have `not` on `Numeric` since to call it we it would be messy. `Numeric::not(Numeric::Number(x)).as_number().unwrap()` instead of `Number::not(x)`

HalidOdat (approved):

PR #1391, opened 2021-07-08
Lexically declared names

macmv:
Currently, lexically declared names are only partially implemented. During parsing, anything declared with `var`, `let`, `function`, or `class`, should be added to a set of `var_declared_names` and/or `lexically_declared_names`. Currently, there is a simple implementation in the [statement list](https://github.com/boa-dev/boa/blob/e5dc8fc6792dee23dc8fb92cab2cdb8e968a49d8/boa/src/syntax/parser/statement/mod.rs#L293) parser, but this does not handle things like outer scoped variables and nested scope variables.

This PR will fix the above issues by changing the following:
- It will add functions to `Node`, which will be able to generate [lexically declared names](https://tc39.es/ecma262/#sec-static-semantics-lexicallydeclarednames) and [var declared names](https://tc39.es/ecma262/#sec-static-semantics-vardeclarednames) according to the spec.
  - These functions will not end up being used at all. The `var` function recursively iterates through all children nodes, which causes O(n!) or something like that with lots of nested blocks (because for each outer block, you need to loop through all the inner ones again). This function will most likely just be used for checking if things like `arguments` is within a `Node`.
- This adds a new type called `DeclaredNames`, which is the var declared names and lexically declared names. It will be passed as a new argument to `TokenParser`, and is meant to be used like how the [spec](https://tc39.es/ecma262/#sec-block-runtime-semantics-evaluation) uses a lexical environment. This new type will only be used at parsing time, and will never be stored in any `Node` fields.
  - This is the fix to the recursion problem. For every node that is parsed, it will edit the environment depending on what type of node it is. So `VarDeclList` nodes will add all the names they encounter into the environment, while `StatementList` will look at all of those names after parsing all the nested nodes. This will allow the entire system to be fast, and not rely on many levels of recursion.
  - Also, I don't know if this is a very good name. The parameter name is `env`, so I was thinking it should be called `LexicalEnvironment` instead, but that is already used somewhere else. This type should be used to store any state information that the parser will need.
- Most of the [redeclaration](https://github.com/tc39/test262/tree/main/test/language/block-scope/syntax/redeclaration) tests should pass after this, and everything will be great, and all the problems will be solved.

### DIFF
--- a/boa/src/syntax/ast/node/mod.rs
+++ b/boa/src/syntax/ast/node/mod.rs
@@ -687,6 +687,7 @@ unsafe impl Trace for MethodDefinitionKind {
 /// any strings which may have been changed in a different indent
 /// level.
 #[cfg(test)]
+#[track_caller]
 fn test_formatting(source: &'static str) {
     // Remove preceding newline.
     let source = &source[1..];
--- a/boa/src/syntax/ast/node/mod.rs
+++ b/boa/src/syntax/ast/node/mod.rs
@@ -308,7 +308,7 @@ impl Node {
 
     /// Returns the var declared names of this node, according to
     /// the [spec](https://tc39.es/ecma262/#sec-static-semantics-vardeclarednames).
-    pub(crate) fn var_declared_names(&self) -> HashSet<&str> {
+    pub fn var_declared_names(&self) -> HashSet<&str> {
         match self {
             Node::Block(ref block) => {
                 let mut names = HashSet::new();
--- a/boa/src/syntax/ast/node/mod.rs
+++ b/boa/src/syntax/ast/node/mod.rs
@@ -54,6 +54,7 @@ use crate::{
 };
 use std::{
     cmp::Ordering,
+    collections::HashSet,
     fmt::{self, Display},
 };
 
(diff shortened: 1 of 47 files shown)


Razican:
  | Passed | 27,936 | 27,942 | **+6** |
  | Failed | 35,345 | 35,339 | **-6** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 35.41% | 35.42% | **+0.01%** |
  
  ```
  test/language/statements/for-in/head-let-bound-names-in-stmt.js [strict mode] (previously Failed)
  test/language/statements/for-in/head-let-bound-names-in-stmt.js (previously Failed)
  test/language/statements/for-in/head-const-bound-names-in-stmt.js [strict mode] (previously Failed)
  test/language/statements/for-in/head-const-bound-names-in-stmt.js (previously Failed)
  test/language/statements/for-of/head-let-bound-names-in-stmt.js [strict mode] (previously Failed)
  test/language/statements/for-of/head-let-bound-names-in-stmt.js (previously Failed)
  test/language/statements/for-of/head-const-bound-names-in-stmt.js [strict mode] (previously Failed)
  test/language/statements/for-of/head-const-bound-names-in-stmt.js (previously Failed)
  test/language/statements/for/head-let-bound-names-in-stmt.js [strict mode] (previously Failed)
  test/language/statements/for/head-let-bound-names-in-stmt.js (previously Failed)
  test/language/statements/for/head-const-bound-names-in-stmt.js [strict mode] (previously Failed)
  test/language/statements/for/head-const-bound-names-in-stmt.js (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-name-redeclaration-attempt-with-const.js [strict mode] (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-name-redeclaration-attempt-with-const.js (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-redeclaration-attempt-after-const.js [strict mode] (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-redeclaration-attempt-after-const.js (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-name-redeclaration-attempt-with-let.js [strict mode] (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-name-redeclaration-attempt-with-let.js (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-redeclaration-attempt-after-let.js [strict mode] (previously Failed)
  test/language/block-scope/syntax/redeclaration/inner-block-var-redeclaration-attempt-after-let.js (previously Failed)
  ```
  </details>
  
  ```
  test/built-ins/BigInt/constructor-from-binary-string.js [strict mode] (previously Passed)
  test/built-ins/BigInt/constructor-from-binary-string.js (previously Passed)
  test/language/statements/for-in/scope-body-lex-boundary.js [strict mode] (previously Passed)
  test/language/statements/for-in/scope-body-lex-boundary.js (previously Passed)
  test/language/statements/let/syntax/let-outer-inner-let-bindings.js [strict mode] (previously Passed)
  test/language/statements/let/syntax/let-outer-inner-let-bindings.js (previously Passed)
  test/language/statements/for-of/scope-body-lex-boundary.js [strict mode] (previously Passed)
  test/language/statements/for-of/scope-body-lex-boundary.js (previously Passed)
  test/language/statements/for/scope-head-lex-close.js [strict mode] (previously Passed)
  test/language/statements/for/scope-head-lex-close.js (previously Passed)
  test/language/statements/for/scope-head-lex-open.js [strict mode] (previously Passed)
  test/language/statements/for/scope-head-lex-open.js (previously Passed)
  test/language/statements/const/syntax/const-outer-inner-let-bindings.js [strict mode] (previously Passed)
  test/language/statements/const/syntax/const-outer-inner-let-bindings.js (previously Passed)
  ```
  </details>

croraf:
Just a heads up that several closures do share the common memory of the common scope. For example:

```ts
let i = 5;

const test = () => {
    return () => {return ++i;}
}

const inner1 = test();
const inner2 = test();

i = 6;

const a = inner1();
console.log(a);
const b = inner2();
console.log(b);
```

Here `i` is shared between inner1 closure, inner2 closure, and can be also modified outside of the closure. All three will share the same variable.

The output is:
7
8

Razican:
@macmv I just stumbled upon this looking for old PRs. This looks very good! Do you think it would be possible to re-base it?

raskad:
This was fixed by #2027

PR #1409, opened 2021-07-18
Lazy computed JsString hashes

HalidOdat:
It changes the following:
- Lazy computed hashes
- Make Environments hold `JsString` instead of `Box<str>`
- Make some nodes hold `JsString`

### DIFF
--- a/boa/src/builtins/console/mod.rs
+++ b/boa/src/builtins/console/mod.rs
@@ -487,7 +487,7 @@ impl Console {
             None => "default".into(),
         };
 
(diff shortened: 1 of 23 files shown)


Razican:
@HalidOdat I think this would need some changes with the new interner. Do we still want at least part of this?

HalidOdat:
> @HalidOdat I think this would need some changes with the new interner. Do we still want at least part of this?

I don't think this is that useful now that interner has been implemented but, I'll experiment once #1659 gets merged.

PR #1413, opened 2021-07-19
Optimized FxHasher

HalidOdat:
This PR inlines `rustc-hash` crate (it is very small ~100 LoC), and optimizes it's hashing by using some safe "unsafe".

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,7 +20,6 @@ dependencies = [
  "once_cell",
  "rand",
  "regress",
- "rustc-hash",
  "ryu-js",
  "serde",
  "serde_json",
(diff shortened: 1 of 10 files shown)


jasonwilliams:
Do you know why there’s a big jump on “Dynamic Object Property Access (Execution)”

HalidOdat:
> Do you know why there’s a big jump on “Dynamic Object Property Access (Execution)”

Not sure why, checked the assembly and it compiles to the same thing, so there shouldn't be a difference, the compiler removes all the unnecessary checks, so the unsafe code is pointless. I'll rerun the benchmarks anyway.

HalidOdat:
Since the compiler is smart and can remove all the check that I used unsafe to remove. I'm closing this PR.

PR #1425, opened 2021-07-22
Extract `PropertyMap` struct from `Object`

jedel1043:
Refactors object properties from `Object` by creating a `PropertyMap` struct.

It changes the following:

- Adds the `PropertyMap` struct
- Changes `Object` to contain a `properties` field of type `PropertyMap`
- Adds a `properties() -> &PropertyMap` function to `Object`
- Renames `iter.rs` to `property_map.rs`


RageKnify (approved):
Nice abstraction to ease the understanding of `Object`, compiler should inline where necessary.

HalidOdat (approved):
Looks good :)

just needs a rebase.

PR #1435, opened 2021-07-27
Constant `JsString`s

HalidOdat:
Every time we call `.get("length", context)` (and we do this a lot), this is bad for two reasons a new `JsString` is allocated every time, we do not utilize the `Rc` nature of `JsString` ref count and comparisons between to same `JsString` that point to the same memory is just a pointer compare.

### DIFF
--- a/boa/src/string.rs
+++ b/boa/src/string.rs
@@ -183,8 +183,7 @@ thread_local! {
                 inner: Inner::new(s),
                 _marker: PhantomData,
             };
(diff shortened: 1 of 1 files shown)


HalidOdat:
I'll try to rebase this too :)

Razican:
This is definitely a good step, but at some point we should intern all this in an interner in `Context`. It would be much more efficient.

Issue #1441, opened 2021-07-28
Functions are always callable


HalidOdat:
Currently we have flags for whether a function is callable, but this is wrong, all functions are callable but they can be optionally constructable.

Issue #1440, opened 2021-07-28
Wrong attribues for `name` and `length` function properties.


HalidOdat:
Currently for for builtins it is `WRITABLE`, `NON_ENUMERABLE`, `PERMANENT`. But it should be `READONLY`, `NON_ENUMERABLE` and `CONFIGURABLE`

https://tc39.es/ecma262/#sec-function-instances

PR #1442, opened 2021-07-28
Implement closure functions

### LINKED ISSUE #1439: Register and create closure functions
Currently we can only create function that can be coerced into a `NativeFunction`, this does not allow us to capture variables from a closure. And changing the function signature of `NativeFunction` to `Box<dyn Fn>` would be bad for performance reasons, since for every function now we have to allocate. Instead we would have another variant in `Function` enum `Closure` that would contain a `Box<dyn Fn>`

So I propose the following, we add separate functions for closures function creation.

```rust
let some_variable = ...
context.register_global_closure("name", move |_, _, _| {
	Ok(some_variable)
});
```

We would also add a way to create a closure with the `FunctionBuilder`

HalidOdat:
This Pull Request fixes/closes #1439 #1441 #1440 #1443.

It changes the following:
- Implement closure functions
- Implement a way to register through context
- Add example for closure `examples/closures.rs`

### DIFF
--- a/boa/examples/closures.rs
+++ b/boa/examples/closures.rs
@@ -1,0 +1,19 @@
+use boa::{Context, JsString};
+
+fn main() {
+    let mut context = Context::new();
+
+    let variable = JsString::new("I am a captured variable");
+
+    context
+        .register_global_closure("closure", 0, move |_, _, _| {
+            // This value is captured from main function.
+            Ok(variable.clone().into())
+        })
+        .unwrap();
+
+    assert_eq!(
+        context.eval("closure()"),
+        Ok("I am an captured variable".into())
+    );
+}
--- a/boa/examples/closures.rs
+++ b/boa/examples/closures.rs
@@ -1,19 +1,20 @@
-use boa::{Context, JsString};
+use boa::{Context, JsString, Value};
 
-fn main() {
+fn main() -> Result<(), Value> {
     let mut context = Context::new();
 
     let variable = JsString::new("I am a captured variable");
 
-    context
-        .register_global_closure("closure", 0, move |_, _, _| {
-            // This value is captured from main function.
-            Ok(variable.clone().into())
-        })
-        .unwrap();
+    // We register a global closure function that has the name 'closure' with length 0.
+    context.register_global_closure("closure", 0, move |_, _, _| {
+        // This value is captured from main function.
+        Ok(variable.clone().into())
+    })?;
 
     assert_eq!(
-        context.eval("closure()"),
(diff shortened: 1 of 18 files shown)


boa/examples/closures.rs
```diff
@@ -0,0 +1,19 @@
+use boa::{Context, JsString};
+
+fn main() {
+    let mut context = Context::new();
+
+    let variable = JsString::new("I am a captured variable");
+
+    context
+        .register_global_closure("closure", 0, move |_, _, _| {
```
jasonwilliams, line 9: I think i would be useful to let people know what the 3 arguments are that we don't need here

boa/src/context.rs
```diff
@@ -543,10 +561,28 @@ impl Context {
...
+        self.global_object().insert_property(
+            name,
+            function,
+            Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
+        );
+        Ok(())
+    }
+
+    #[inline]
+    pub fn register_global_closure<F>(&mut self, name: &str, length: usize, body: F) -> Result<()>
```
jasonwilliams, line 579: some docs above this like register_global_function has would be useful. Just to say why it exists and what usecase it may solve

boa/src/syntax/ast/node/declaration/function_decl/mod.rs
```diff
@@ -89,14 +89,12 @@ impl Executable for FunctionDecl {
     fn run(&self, context: &mut Context) -> Result<Value> {
         let _timer = BoaProfiler::global().start_event("FunctionDecl", "exec");
         let val = context.create_function(
+            self.name(),
             self.parameters().to_vec(),
             self.body().to_vec(),
-            FunctionFlags::CALLABLE | FunctionFlags::CONSTRUCTABLE,
+            FunctionFlags::CONSTRUCTABLE,
```
jasonwilliams, line 95: was CALLABLE never needed here?

HalidOdat: This PR also fixes #1441, That's why `FunctionFlags::CALLABLE` was removed, since it's implied.

boa/src/context.rs
HalidOdat: I will add documentation, I just wanted feedback first on the API

Razican (approved):
I like the API. It might have made sense to group functions and closures together, but this makes it a bit more performant for functions :)

jasonwilliams (approved):

PR #1445, opened 2021-07-29
improve map conformance without losing perf

neeldug:
Improves conformance but avoids following spec directly for performance implications.

### DIFF
--- a/boa/src/builtins/map/mod.rs
+++ b/boa/src/builtins/map/mod.rs
@@ -259,7 +259,7 @@ impl Map {
                     }
                     _ => key.clone(),
                 };
-                map.insert(key, value.into());
+                map.insert(key, value.clone());
                 map.len()
             } else {
                 return Err(context.construct_type_error("'this' is not a Map"));
--- a/boa/src/builtins/map/mod.rs
+++ b/boa/src/builtins/map/mod.rs
@@ -1,4 +1,4 @@
-//! This module implements the global `Map` objest.
+//! This module implements the global `Map` object.
 //!
 //! The JavaScript `Map` class is a global object that is used in the construction of maps; which
 //! are high-level, key-value stores.
@@ -21,7 +21,7 @@ use crate::{
     },
     property::{Attribute, PropertyDescriptor, PropertyNameKind},
     symbol::WellKnownSymbols,
-    BoaProfiler, Context, JsBigInt, JsResult, JsValue,
+    BoaProfiler, Context, JsResult, JsValue,
 };
 use ordered_map::OrderedMap;
 
@@ -257,20 +257,6 @@ impl Map {
                             key.into()
                         }
                     }
-                    JsValue::Integer(i) => {
-                        if i.is_zero() {
-                            JsValue::Integer(0)
-                        } else {
-                            key.into()
-                        }
-                    }
-                    JsValue::BigInt(b) => {
-                        if b.is_zero() {
-                            JsValue::BigInt(JsBigInt::from(0))
-                        } else {
-                            key.into()
-                        }
-                    }
                     _ => key.into(),
                 };
                 map.insert(key, value.into());
@@ -349,22 +335,6 @@ impl Map {
             JsValue::Rational(r) => {
                 if r.is_zero() {
                     temp_v = JsValue::Rational(0f64);
-                    &temp_v
-                } else {
-                    key
-                }
-            }
-            JsValue::Integer(i) => {
-                if i.is_zero() {
-                    temp_v = JsValue::Integer(0);
-                    &temp_v
-                } else {
-                    key
-                }
-            }
-            JsValue::BigInt(b) => {
-                if b.is_zero() {
-                    temp_v = JsValue::BigInt(JsBigInt::from(0));
                     &temp_v
                 } else {
                     key
--- a/boa/src/builtins/map/mod.rs
+++ b/boa/src/builtins/map/mod.rs
@@ -254,10 +254,10 @@ impl Map {
                         if r.is_zero() {
                             JsValue::Rational(0f64)
                         } else {
-                            key.into()
+                            key.clone()
                         }
                     }
-                    _ => key.into(),
+                    _ => key.clone(),
                 };
                 map.insert(key, value.into());
                 map.len()
@@ -287,15 +287,6 @@ impl Map {
         args: &[JsValue],
         context: &mut Context,
     ) -> JsResult<JsValue> {
-        // 1. Let M be the this value.
-        // 2. Perform ? RequireInternalSlot(M, [[MapData]]).
-        {
-            let obj = this.as_object().unwrap_or_default();
-            let obj = obj.borrow();
-            obj.as_map_ref().ok_or_else(|| {
-                context.construct_type_error("Map.prototype.clear called with invalid value")
-            })?;
-        }
         let key = args.get_or_undefined(0);
 
         let (deleted, size) = if let Some(object) = this.as_object() {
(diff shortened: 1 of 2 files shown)


neeldug:
  | Passed | 28,442 | 28,456 | **+14** |
  | Ignored | 15,614 | 15,614 | 0 |
  | Failed | 34,841 | 34,827 | **-14** |
  | Panics | 2 | 2 | 0 |
  | Conformance | 36.05% | 36.07% | **+0.02%** |
  
  ```
  test/built-ins/Map/prototype/get/returns-value-normalized-zero-key.js [strict mode] (previously Failed)
  test/built-ins/Map/prototype/get/returns-value-normalized-zero-key.js (previously Failed)
  test/built-ins/Map/prototype/forEach/first-argument-is-not-callable.js [strict mode] (previously Failed)
  test/built-ins/Map/prototype/forEach/first-argument-is-not-callable.js (previously Failed)
  test/built-ins/Map/prototype/set/append-new-values-normalizes-zero-key.js [strict mode] (previously Failed)
  test/built-ins/Map/prototype/set/append-new-values-normalizes-zero-key.js (previously Failed)
  test/built-ins/Map/prototype/set/replaces-a-value-normalizes-zero-key.js [strict mode] (previously Failed)
  test/built-ins/Map/prototype/set/replaces-a-value-normalizes-zero-key.js (previously Failed)
  test/built-ins/Map/prototype/clear/context-is-not-map-object.js [strict mode] (previously Failed)
  test/built-ins/Map/prototype/clear/context-is-not-map-object.js (previously Failed)
  test/built-ins/Map/prototype/clear/context-is-set-object-throws.js [strict mode] (previously Failed)
  test/built-ins/Map/prototype/clear/context-is-set-object-throws.js (previously Failed)
  test/built-ins/Map/prototype/clear/context-is-not-object.js [strict mode] (previously Failed)
  test/built-ins/Map/prototype/clear/context-is-not-object.js (previously Failed)
  ```
  </details>

Razican:
Hello @neeldug ! thank you for the contribution and sorry for the delay in the review. It seems this requires a rebase, and it seems to be failing some clippy checks.

RageKnify (approved):

neeldug:
I think this was implemented according to spec recently, only thing I could find that could be fixed is that arrays should store objects and not primitives in JS which is causing all of the `instanceof` testcases to fail.

raskad (approved):

boa/src/builtins/map/mod.rs
```diff
@@ -248,7 +249,31 @@ impl Map {
...
+                    }
+                    JsValue::BigInt(b) => {
+                        if b.is_zero() {
+                            JsValue::BigInt(JsBigInt::from(0))
+                        } else {
+                            key.into()
+                        }
+                    }
+                    _ => key.into(),
+                };
```
jedel1043, line 275: This checks for -0.0f, right? Wouldn't be enough with the `JsValue::Rational` branch?

neeldug: Yep, just realised you're totally right, no need for the extra branches, thanks!

boa/src/builtins/map/mod.rs
```diff
@@ -248,7 +249,17 @@ impl Map {
 
         let size = if let Some(object) = this.as_object() {
             if let Some(map) = object.borrow_mut().as_map_mut() {
-                map.insert(key.clone(), value.clone());
+                let key = match key {
+                    JsValue::Rational(r) => {
+                        if r.is_zero() {
+                            JsValue::Rational(0f64)
+                        } else {
+                            key.into()
```
jedel1043, line 257:
```suggestion
                            key.clone()
```
Prefer using `clone` instead of `into` when converting a `&JsValue` into a `JsValue`. Also applies for the other `into`s.
We should remove the implementation of `From<&JsValue>` for `JsValue`, really.

boa/src/builtins/map/mod.rs
```diff
@@ -276,6 +287,15 @@ impl Map {
...
     ) -> JsResult<JsValue> {
+        // 1. Let M be the this value.
+        // 2. Perform ? RequireInternalSlot(M, [[MapData]]).
+        {
+            let obj = this.as_object().unwrap_or_default();
+            let obj = obj.borrow();
+            obj.as_map_ref().ok_or_else(|| {
+                context.construct_type_error("Map.prototype.clear called with invalid value")
+            })?;
+        }
```
jedel1043, line 298: Is this check necessary? This is the exact same check that the `if let Some(map) = object.borrow_mut().as_map_mut()` takes care of.

boa/src/builtins/map/mod.rs
```diff
@@ -309,6 +329,20 @@ impl Map {
...
+                if r.is_zero() {
+                    temp_v = JsValue::Rational(0f64);
+                    &temp_v
+                } else {
+                    key
+                }
+            }
+            _ => key,
+        };
+
```
jedel1043, line 345:
```suggestion
        const JS_ZERO: &JsValue = &JsValue::Rational(0f64);

        let key = args.get_or_undefined(0);

        let key = match key {
            JsValue::Rational(r) => {
                if r.is_zero() {
                    JS_ZERO
                } else {
                    key
                }
            }
            _ => key,
        };
```

A constant is more readable than a free `temp_v`.

boa/src/builtins/map/mod.rs
```diff
@@ -333,7 +367,20 @@ impl Map {
...
+    pub(crate) fn clear(this: &JsValue, _: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        // 1. Let M be the this value.
+        // 2. Perform ? RequireInternalSlot(M, [[MapData]]).
+        {
+            let obj = this.as_object().unwrap_or_default();
+            let obj = obj.borrow();
+            obj.as_map_ref().ok_or_else(|| {
+                context.construct_type_error("Map.prototype.clear called with invalid value")
+            })?;
+        }
```
jedel1043, line 379: Same concerns as the previous one. This check could be avoided if we do `if let Some(obj) = this.as_object()` and `if let Some(map) = obj.mut_borrow().as_map_mut()` instead of `set_data`.

jedel1043 (changes requested):
Some more changes and questions.

neeldug:
Did check and it's unnecessary for delete, and for clear this should work, couldn't find a clean way to replace the reference with a new map so went into OrderedMap to create a method.

boa/src/builtins/map/mod.rs
```diff
@@ -333,12 +357,23 @@ impl Map {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-map.prototype.clear
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/clear
-    pub(crate) fn clear(this: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
-        this.set_data(ObjectData::map(OrderedMap::new()));
-
-        Self::set_size(this, 0);
```
jedel1043, line 339: Missed the `set_size` on the new version. That's why some tests now fail. You can put if after the `if let`, but we should really consider implementing the `get Map.prototype.size` accessor instead of managing the state ourselves.

boa/src/builtins/map/ordered_map.rs
```diff
@@ -142,6 +142,10 @@ impl<V> OrderedMap<V> {
         }
     }
 
+    pub fn clear(&mut self) {
+        self.map.truncate(0);
```
jedel1043, line 146: You can use `self.map.clear()`. The implementation of `clear` is internally `truncate(0)`, and both run at O(n) because they need to run the destructors of all the elements. If you're worried about the allocated memory, you can add a `shrink_to_fit`.

boa/src/builtins/map/mod.rs
```diff
@@ -248,7 +249,17 @@ impl Map {
...
+                    JsValue::Rational(r) => {
+                        if r.is_zero() {
+                            JsValue::Rational(0f64)
+                        } else {
+                            key.clone()
+                        }
+                    }
+                    _ => key.clone(),
+                };
+                map.insert(key, value.into());
```
jedel1043, line 262:
```suggestion
                map.insert(key, value.clone());
```
Missed one :)

jedel1043 (changes requested):
Small nitpicks. We can merge it after this gets fixed :D

jedel1043 (approved):
Great work!

PR #1446, opened 2021-07-30
Implement new `Value::type_of` method

### LINKED ISSUE #1436: Adding a `Map` to itself panics on the second attempt
**Describe the bug**
When adding a `Map` to itself via `set()`, the first `set()` succeeds, but the second one does reliably produce a panic.
This behavior seems to be somewhat related to inconsistent panics in the test `test/built-ins/Map/prototype/keys/returns-iterator.js`.

**To Reproduce**
```javascript
var map = new Map();
map.set(map, 1);
map.set(map, 1);
```

This code panics with the error: `thread 'main' panicked at 'Object already mutably borrowed: BorrowError', boa/src/value/type.rs:50:27`

**Expected behavior**
Should not panic and set ok.

**Build environment:**

- OS: [Linux]
- Version: [master]
- Target triple: [x86_64-unknown-linux-gnu]
- Rustc version: [rustc 1.52.0 (88f19c6da 2021-05-03)]

jedel1043:
This Pull Request fixes/closes #1436.

It changes the following:

- implements the operator `typeof` as `Value::type_of`
- changes `get_type` to be equivalent to the `Type(v)` operation in the [spec](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-data-types-and-values)
- replaces calls to `type_of` with `get_type`

### DIFF
--- a/boa/src/builtins/object/mod.rs
+++ b/boa/src/builtins/object/mod.rs
@@ -305,12 +305,12 @@ impl Object {
         if !matches!(proto.get_type(), Type::Object | Type::Null) {
             return ctx.throw_type_error(format!(
                 "expected an object or null, got {}",
-                proto.get_type().as_str()
+                proto.type_of()
             ));
         }
 
         // 3. If Type(O) is not Object, return O.
-        if obj.get_type() != Type::Object {
+        if !obj.is_object() {
             return Ok(obj);
         }
 
--- a/boa/src/syntax/ast/node/call/mod.rs
+++ b/boa/src/syntax/ast/node/call/mod.rs
@@ -4,7 +4,7 @@ use crate::{
     exec::InterpreterState,
     gc::{Finalize, Trace},
     syntax::ast::node::{join_nodes, Node},
-    value::{Type, Value},
+    value::Value,
     BoaProfiler, Context, Result,
 };
 use std::fmt;
@@ -66,7 +66,7 @@ impl Executable for Call {
         let (this, func) = match self.expr() {
             Node::GetConstField(ref get_const_field) => {
                 let mut obj = get_const_field.obj().run(context)?;
-                if obj.get_type() != Type::Object {
+                if !obj.is_object() {
                     obj = Value::Object(obj.to_object(context)?);
                 }
                 (
@@ -76,7 +76,7 @@ impl Executable for Call {
             }
             Node::GetField(ref get_field) => {
                 let mut obj = get_field.obj().run(context)?;
-                if obj.get_type() != Type::Object {
+                if !obj.is_object() {
                     obj = Value::Object(obj.to_object(context)?);
                 }
                 let field = get_field.field().run(context)?;
--- a/boa/src/syntax/ast/node/field/get_const_field/mod.rs
+++ b/boa/src/syntax/ast/node/field/get_const_field/mod.rs
@@ -2,7 +2,7 @@ use crate::{
     exec::Executable,
     gc::{Finalize, Trace},
     syntax::ast::node::Node,
-    value::{Type, Value},
+    value::Value,
     Context, Result,
 };
 use std::fmt;
@@ -65,7 +65,7 @@ impl GetConstField {
 impl Executable for GetConstField {
     fn run(&self, context: &mut Context) -> Result<Value> {
         let mut obj = self.obj().run(context)?;
-        if obj.get_type() != Type::Object {
+        if !obj.is_object() {
             obj = Value::Object(obj.to_object(context)?);
         }
 
--- a/boa/src/syntax/ast/node/field/get_field/mod.rs
+++ b/boa/src/syntax/ast/node/field/get_field/mod.rs
@@ -2,7 +2,7 @@ use crate::{
     exec::Executable,
     gc::{Finalize, Trace},
     syntax::ast::node::Node,
-    value::{Type, Value},
+    value::Value,
     Context, Result,
 };
 use std::fmt;
@@ -64,7 +64,7 @@ impl GetField {
 impl Executable for GetField {
     fn run(&self, context: &mut Context) -> Result<Value> {
         let mut obj = self.obj().run(context)?;
-        if obj.get_type() != Type::Object {
+        if !obj.is_object() {
             obj = Value::Object(obj.to_object(context)?);
         }
         let field = self.field().run(context)?;
--- a/boa/src/syntax/ast/node/template/mod.rs
+++ b/boa/src/syntax/ast/node/template/mod.rs
@@ -1,7 +1,7 @@
 //! Template literal node.
 
 use super::Node;
-use crate::{builtins::Array, exec::Executable, value::Type, BoaProfiler, Context, Result, Value};
+use crate::{builtins::Array, exec::Executable, BoaProfiler, Context, Result, Value};
 use gc::{Finalize, Trace};
 
 #[cfg(feature = "deser")]
@@ -112,7 +112,7 @@ impl Executable for TaggedTemplate {
         let (this, func) = match *self.tag {
             Node::GetConstField(ref get_const_field) => {
                 let mut obj = get_const_field.obj().run(context)?;
-                if obj.get_type() != Type::Object {
+                if !obj.is_object() {
                     obj = Value::Object(obj.to_object(context)?);
                 }
                 (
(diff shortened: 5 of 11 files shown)


Razican:
  | Passed | 28,442 | 28,442 | 0 |
  | Ignored | 15,614 | 15,614 | 0 |
  | Failed | 34,841 | 34,841 | 0 |
  | Panics | 2 | 2 | 0 |
  | Conformance | 36.05% | 36.05% | 0.00% |

HalidOdat:
The way I would solve this is by renaming `get_type` to `typeof` and make it return a `&'static str` (since that is what it is), and create a new function `get_type` that return an enum (that does not contain function, but has only the value variants, rational and integer is the same number).

jedel1043:
> The way I would solve this is by renaming `get_type` to `typeof` (since that is what it is) and make it return a `&'static str` (`"string"`, `"number"`, `"function"`, etc), and create a new function `get_type` that return an enum (that does not contain Function variant, but has only the `Value` variants, rational and integer is the same number).
> 

I should rename `get_type` to `typeof`. Also, I forgot that `std::mem::discriminant` existed lol, let me rewrite my code.

HalidOdat:
For `std::mem::discriminant` it would return a different value for `Value::Rational` and `Value::Integer` when they should be the same.

jedel1043:
> For `std::mem::discriminant` it would return a different value for `Value::Rational` and `Value::Integer` when they should be the same.

Ah no, I was thinking about creating a new internal function `cmp_types` that returns true if the discriminants are the same and defaults to `get_type` equality otherwise. I wouldn't want to have two `get_type` functions that do practically the same thing but only one is used constantly and the other one is used just to fix a bug.

RageKnify (approved):
Looks good to me, I'd prefer if `typ` would be `type_`, I understand `type` is a keyword, but I prefer the underscore over shortening the word.

HalidOdat (changes requested):
What I meant by change `get_type` to `typeof` and make it return a string was that, it should be the JavaScript `typeof` operator (like `Value::add` + op). That is used in the `typeof` unary node execution.

And create a new function `get_type()` that returns a an emun `Type` without the `Function` variant (which means it does not call borrow to know it is a function).

jedel1043:
> What I meant by change `get_type` to `typeof` and make it return a string was that, it should be the JavaScript `typeof` operator (like `Value::add` + op). That is used in the `typeof` unary node execution.
> 
> And create a new function `get_type()` that returns a an emun `Type` without the `Function` variant (which means it does not call borrow to know it is a function).

`type_of` does return a &'static str and it is used in the `typeof` ast node. Also `typ` does return an enum `Type` without `Function` objects. Do you mean to make `type_of` return a `JsString`?

boa/src/syntax/ast/node/field/get_field/mod.rs
```diff
@@ -64,7 +64,7 @@ impl GetField {
 impl Executable for GetField {
     fn run(&self, context: &mut Context) -> Result<Value> {
         let mut obj = self.obj().run(context)?;
-        if obj.get_type() != Type::Object {
+        if obj.typ() != Type::Object {
```
HalidOdat, line 67:
```suggestion
        if !obj.is_object() {
```

It's better/cleaner to use the `is_` methods instead of getting the type.

boa/src/syntax/ast/node/field/get_const_field/mod.rs
```diff
@@ -65,7 +65,7 @@ impl GetConstField {
 impl Executable for GetConstField {
     fn run(&self, context: &mut Context) -> Result<Value> {
         let mut obj = self.obj().run(context)?;
-        if obj.get_type() != Type::Object {
+        if obj.typ() != Type::Object {
```
HalidOdat, line 68:
```suggestion
        if !obj.is_object() {
```
Here too.

boa/src/syntax/ast/node/call/mod.rs
```diff
@@ -76,7 +76,7 @@ impl Executable for Call {
             }
             Node::GetField(ref get_field) => {
                 let mut obj = get_field.obj().run(context)?;
-                if obj.get_type() != Type::Object {
+                if obj.typ() != Type::Object {
```
HalidOdat, line 79:
```suggestion
                if !obj.is_object() {
```

boa/src/syntax/ast/node/call/mod.rs
```diff
@@ -66,7 +66,7 @@ impl Executable for Call {
         let (this, func) = match self.expr() {
             Node::GetConstField(ref get_const_field) => {
                 let mut obj = get_const_field.obj().run(context)?;
-                if obj.get_type() != Type::Object {
+                if obj.typ() != Type::Object {
```
HalidOdat, line 69:
```suggestion
                if !obj.is_object() {
```

boa/src/builtins/object/mod.rs
```diff
@@ -302,15 +302,15 @@ impl Object {
...
             return ctx.throw_type_error(format!(
                 "expected an object or null, got {}",
-                proto.get_type().as_str()
+                proto.type_of()
             ));
         }
 
         // 3. If Type(O) is not Object, return O.
-        if obj.get_type() != Type::Object {
+        if obj.typ() != Type::Object {
```
HalidOdat, line 313:
```suggestion
        if !obj.is_object() {
```

boa/src/syntax/ast/node/template/mod.rs
```diff
@@ -112,7 +112,7 @@ impl Executable for TaggedTemplate {
         let (this, func) = match *self.tag {
             Node::GetConstField(ref get_const_field) => {
                 let mut obj = get_const_field.obj().run(context)?;
-                if obj.get_type() != Type::Object {
+                if obj.typ() != Type::Object {
```
HalidOdat, line 115:
```suggestion
                if !obj.is_object() {
```

HalidOdat (commented):
We should remove `Type::Function` variant entirely since a function is just a object.

I think it's best to rename `type_of` to `typeof` to mach with JavaScript,  and make it return a `JsString`. (don't worry about allocations after #1435, it will always return the same string, since some common strings are cached). 

And rename `typ` to `get_type`.

jedel1043:
> I think it's best to rename `type_of` to `typeof` to mach with JavaScript, and make it return a `JsString`. (don't worry about allocations after #1435, it will always return the same string, since some common strings are cached).

Unfortunately `typeof` is a reserved word, that's why I used `type_of`. Should I leave it like that or maybe change it to `typeof_`?

Leaving that aside... yeah, I can't see the need of `Function` if we don't really use it, and returning JsString with #1435 seems like the best thing to do. I'm gonna integrate the remaining changes then. Should I remove the `AsStr` implementation? If we going to have the string representation of `Function` inside `type_of` then I don't see the benefit of having `as_str` if we can just print it with `Debug` version in case we need to check the type of a `Value`.

HalidOdat:
> Unfortunately `typeof` is a reserved word, that's why I used `type_of`. Should I leave it like that or maybe change it to `typeof_`?

ah sorry. I did not know it was a reserved keyword. then keeping it as `type_of` seems fine.

> Should I remove the `AsStr` implementation?

Hmmm. Yeah. We can remove it.

boa/src/syntax/ast/node/field/get_field/mod.rs
jedel1043: Fixed

boa/src/syntax/ast/node/field/get_const_field/mod.rs
jedel1043: Fixed

boa/src/syntax/ast/node/call/mod.rs
jedel1043: Fixed

boa/src/syntax/ast/node/call/mod.rs
jedel1043: Fixed

boa/src/builtins/object/mod.rs
jedel1043: Fixed

boa/src/syntax/ast/node/template/mod.rs
jedel1043: Fixed

HalidOdat (approved):

PR #1450, opened 2021-07-31
Fix clippy errors

HalidOdat:
This PR fixes clippy lints and makes lints warnings instead of errors. Why do we want this? It's because as clippy becomes "smarter" it's going to detect more, and they all will be errors, breaking our builds. This mostly affects nightly users as they have the latest version of clippy, but it still applies for stable (like now).

Changed CI so clippy lint warnings cause a CI failure

### DIFF
--- a/boa/src/lib.rs
+++ b/boa/src/lib.rs
@@ -13,7 +13,6 @@ This is an experimental Javascript lexer, parser and compiler written in Rust. C
     html_favicon_url = "https://raw.githubusercontent.com/jasonwilliams/boa/master/assets/logo.svg"
 )]
 #![deny(
-    unused_qualifications,
     clippy::all,
     unused_qualifications,
     unused_import_braces,
@@ -43,8 +42,6 @@ This is an experimental Javascript lexer, parser and compiler written in Rust. C
 )]
 
 pub mod bigint;
-// builtins module has a lot of built-in functions that need unnecessary_wraps
-#[allow(clippy::unnecessary_wraps)]
 pub mod builtins;
 pub mod class;
 pub mod context;
@@ -57,8 +54,6 @@ pub mod property;
 pub mod realm;
 pub mod string;
 pub mod symbol;
-// syntax module has a lot of acronyms
-#[allow(clippy::upper_case_acronyms)]
 pub mod syntax;
 pub mod value;
 
--- a/boa/src/value/tests.rs
+++ b/boa/src/value/tests.rs
@@ -10,15 +10,15 @@ use std::hash::{Hash, Hasher};
 fn is_object() {
     let context = Context::new();
     let val = Value::new_object(&context);
-    assert_eq!(val.is_object(), true);
+    assert!(val.is_object());
 }
 
 #[test]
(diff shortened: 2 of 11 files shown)


RageKnify (approved):

boa/src/lib.rs
```diff
@@ -31,7 +29,12 @@ This is an experimental Javascript lexer, parser and compiler written in Rust. C
     future_incompatible,
     nonstandard_style,
 )]
-#![warn(clippy::perf, clippy::single_match_else, clippy::dbg_macro)]
+#![warn(
+    clippy::all,
```
Razican, line 33: The problem with this is that even the ones considered as "error" by clippy will be silenced. Maybe the solution is to just not mention them (that way, errors will be errors, warnings will be warnings).

.github/workflows/rust.yml
```diff
@@ -192,7 +192,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- --verbose
+          args: -- -D warnings --verbose
```
Razican, line 195: What does this do?

boa/src/value/tests.rs
```diff
@@ -757,6 +757,8 @@ mod cyclic_conversions {
 }
 
 mod abstract_relational_comparison {
+    #![allow(clippy::bool_assert_comparison)]
```
Razican, line 760: Why are we allowing this here?

Razican (commented):
Looks pretty good. I would still not change all clippy lints to warnings though. I think we can leave them as clippy does. But I'm even in favour of leaving them as errors, even if that will break the CI on some clippy releases.

.github/workflows/rust.yml
HalidOdat: > What does this do?

It makes all clippy warnings into errors, so even warnings don't pass through the CI

boa/src/value/tests.rs
HalidOdat: It's because the macro we use uses `assert_eq($expr, $expected)` and when it's replaced it gives, and expected can be `false` or `true` depending on macro invocation, but clippy does not know this and says it should be `assert!($expr)`. And this is not correct for when we call with false since it has to be negated (assert!(!$expr))

boa/src/lib.rs
HalidOdat: But if we exclude it the ones that are neither error or warnings wont we shown...

Maybe leave it as always error and add or change the clippy CI to nightly so we catch them ahead of time and it wont fail on stable or nighlty

HalidOdat: ...Or we could just leave it as it was.

.github/workflows/rust.yml
Razican: So, this will still fail the CI even if we put everything in clippy as a warning, right? Then, I would leave `clippy::all` as `deny`.

HalidOdat: Yes. lets leave it as is.

PR #1457, opened 2021-08-06
Rename `Value` to `JsValue`

HalidOdat:
Related to #1376 

It changes the following:
- Rename `Value` to `JsValue`
- Remove `Value::number()`, `Value::integer()`, `Value::rational()` for a universal `JsValue::new()` constructor.


bartlomieju:
I believe this closes https://github.com/boa-dev/boa/issues/1376

HalidOdat:
> I believe this closes #1376

This does not close it since we still have `GcObject` => `JsObject` and we might want to change `boa::Result` as well.

Issue #803, opened 2020-10-05
Refactor the JS testing system


dvtkrlbs:
Currently most of the tests are just the same boilerplate:
```rust
#[test]
fn test() {
    let mut ctx = Context::new();
    let init = r#"some js"#;
    eprintln!("{}", forward(&mut ctx, init));

    assert_eq!(forward(&mut ctx, "value we expect from execution of that js code");
}
```
I think these codes can be extracted to js files and be executed by some test runner. V8 does this by creating js file and an out file that contains the expected output. Also another way is exposing some testing related functions like assert to js side while testing and doing this with single file. I realize that this is similiar to the 262 tests.

Lan2u:
As you say this seems to very much overlap with the work required for test 262, it makes sense to me that we could utilise the existing progress into that (e.g. the gh page display etc.) for this.

dvtkrlbs:
Yeah i agree the required work for both of them are quite similiar. I however dont think this kind of testing system needs a gh page. I thought of this to replace the testing of js exposed tests. It does not harm to have more tests. V8's test suite consists of both their own tests and 262 tests. They even use tests from Mozilla's testing suite.

Razican:
There is some debate on this. On one hand, we can reduce the boilerplate needed by providing a `check_output()` function that receives some code and the expected output, and that might work for many of our tests. This would be in charge of creating the context and doing the asserts and so on.

Then, there is the thing about moving things to JS files. We have already done this with benchmarks, and they are still compiled and executed fast (which is the main concern with the Test262 suite: it takes around 10 minutes to run). This would also allow running only one test, for example. The main issue I see with this is keeping track of the input/output files and tests, and making sure all of them are run.

Something that could be done in this regard would be to implement some kind of "test creator" that would generate code for all the tests in a given folder. This, of course requires some extra work, and a decision has to be taken on the way we implement it. This, as @Lan2u says, might take over the work done by the Test262 tester, and might be slower. We must also ensure a way for these tests to be created before every commit.

Let's see what others think @HalidOdat @jasonwilliams @RageKnify

RageKnify:
> This would also allow running only one test, for example.
`cargo` already allows this, running `cargo test $TOP::$MID::$LOWER::tests::$TEST_NAME`lets you run a single test, you can also give the name of a module to run all the tests from that module.

While there is a bit of boiler-plate around each test I don't dislike the current setup.
Not sure how we'd be able to get coverage values with the different setup.

[...]

PR #1458, opened 2021-08-08
tests: add check_output test helper

bartlomieju:
Towards https://github.com/boa-dev/boa/issues/803

### DIFF
--- a/boa/src/lib.rs
+++ b/boa/src/lib.rs
@@ -166,6 +166,12 @@ pub(crate) fn check_output(maybe_init: Option<&str>, cases: &[(&str, &str)]) {
     }
 
     for (i, (case, expected)) in cases.iter().enumerate() {
-        assert_eq!(&forward(&mut context, case), expected, "Test case {} ('{}')", i + 1, case);
+        assert_eq!(
+            &forward(&mut context, case),
+            expected,
+            "Test case {} ('{}')",
+            i + 1,
+            case
+        );
     }
 }
--- a/boa/src/lib.rs
+++ b/boa/src/lib.rs
@@ -152,3 +152,20 @@ pub(crate) fn exec<T: AsRef<[u8]>>(src: T) -> String {
         Err(error) => error.display().to_string(),
     }
 }
+
+/// Create a clean Context, optinally call `forward` if init script was provided,
+/// call `forward` for each of provided test cases and assert output from `forward`
+/// matches expected string.
+#[cfg(test)]
+#[track_caller]
+pub(crate) fn check_output(maybe_init: Option<&str>, cases: &[(&str, &str)]) {
+    let mut context = Context::new();
+
+    if let Some(init) = maybe_init {
+        forward(&mut context, init);
+    }
+
+    for (i, (case, expected)) in cases.iter().enumerate() {
+        assert_eq!(&forward(&mut context, case), expected, "Test case {} ('{}')", i + 1, case);
+    }
+}
--- a/boa/src/lib.rs
+++ b/boa/src/lib.rs
@@ -153,25 +153,34 @@ pub(crate) fn exec<T: AsRef<[u8]>>(src: T) -> String {
     }
 }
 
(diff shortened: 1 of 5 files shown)


boa/src/value/tests.rs
```diff
@@ -64,62 +64,47 @@ fn number_is_true() {
...
+            ("'-0' == 0", "true"),
+            ("0 == NaN", "false"),
+            ("'foo' == NaN", "false"),
+            ("NaN == NaN", "false"),
+            (
+                "Number.POSITIVE_INFINITY === Number.POSITIVE_INFINITY",
+                "true",
+            ),
+            (
+                "Number.NEGAVIVE_INFINITY === Number.NEGAVIVE_INFINITY",
```
jedel1043, line 104:
```suggestion
                "Number.NEGATIVE_INFINITY === Number.NEGATIVE_INFINITY",
```
Realized there's a typo here 😅, this is evaluating `undefined === undefined`.

jedel1043:
Maybe it would be good to allow the execution of intermediate instructions between tests. That way we could cleanup some tests like:
```Rust
let init = r#"
    var iterator = [1, 2, 3].keys();
    var next = iterator.next();
"#;
forward(&mut context, init);
assert_eq!(forward(&mut context, "next.value"), "0");
assert_eq!(forward(&mut context, "next.done"), "false");
forward(&mut context, "next = iterator.next()");
assert_eq!(forward(&mut context, "next.value"), "1");
assert_eq!(forward(&mut context, "next.done"), "false");
```

I'd suggest creating a `TestAction` enum with the variants `TestEq((&str, &str))` and `Execute(&str)`. This would change the previous test to:

```Rust
let init = r#"
    var iterator = [1, 2, 3].keys();
    var next = iterator.next();
"#;
check_output(
    &[
        TestAction::Execute(init),
        TestAction::TestEq(("next.value", "0")),
        TestAction::TestEq(("next.done", "false")),
        TestAction::Execute("next = iterator.next()"),
        TestAction::TestEq(("next.value", "1")),
        TestAction::TestEq(("next.done", "false")),
    ]
)
```
Notice how we don't need the `maybe_init` parameter, we just pass it as a statement to be executed.

bartlomieju:
Sure, I can rework it to your suggestion

bartlomieju:
@jedel1043 I applied your suggestion, please let me know if this is satisfactory

HalidOdat (approved):
This looks really good! We can merge this as is and refactor more in future PRs

boa/src/lib.rs
```diff
@@ -153,25 +153,43 @@ pub(crate) fn exec<T: AsRef<[u8]>>(src: T) -> String {
...
+            }
+            TestAction::TestStartsWith(case, expected) => {
+                assert!(
+                    &forward(&mut context, case).starts_with(expected),
+                    "Test case {} ('{}')",
+                    i + 1,
+                    case
+                );
+            }
+        }
```
jedel1043, line 200:
```suggestion
    let mut i = 1;
    for action in actions {
        match action {
            TestAction::Execute(src) => {
                forward(&mut context, src);
            }
            TestAction::TestEq(case, expected) => {
                assert_eq!(
                    &forward(&mut context, case),
                    expected,
                    "Test case {} ('{}')",
                    i,
                    case
                );
                i += 1;
            }
            TestAction::TestStartsWith(case, expected) => {
                assert!(
                    &forward(&mut context, case).starts_with(expected),
                    "Test case {} ('{}')",
                    i,
                    case
                );
                i += 1;
            }
        }
```

Let's skip counting between executions

jedel1043 (changes requested):
Just a small nitpick about the test counter. Everything else is pretty good! We should be able to extend `TestAction` in case we find more common use cases.

boa/src/lib.rs
bartlomieju: Applied

jedel1043 (approved):
Looks great to me!

PR #1464, opened 2021-08-12
Optimize integer negation

HalidOdat:
It changes the following:
- Negation of integer now return integer.
- Fix the `%` operator to return a number with the sign of its lhs.

### DIFF
--- a/boa/src/value/operations.rs
+++ b/boa/src/value/operations.rs
@@ -132,12 +132,19 @@ impl JsValue {
                 if *y == 0 {
                     Self::nan()
                 } else {
-                    Self::new(x % *y)
+                    match x % *y {
+                        rem if rem == 0 && *x < 0 => Self::new(-0.0),
+                        rem => Self::new(rem),
+                    }
                 }
             }
-            (Self::Rational(x), Self::Rational(y)) => Self::new(x % y),
-            (Self::Integer(x), Self::Rational(y)) => Self::new(f64::from(*x) % y),
-            (Self::Rational(x), Self::Integer(y)) => Self::new(x % f64::from(*y)),
+            (Self::Rational(x), Self::Rational(y)) => Self::new((x % y).copysign(*x)),
+            (Self::Integer(x), Self::Rational(y)) => {
+                let x = f64::from(*x);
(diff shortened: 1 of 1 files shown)


Razican (approved):

Razican:
Interestingly, it's failing 4 tests

boa/src/value/operations.rs
```diff
@@ -394,7 +394,8 @@ impl JsValue {
                 Err(_) => f64::NAN,
             }),
             Self::Rational(num) => Self::new(-num),
-            Self::Integer(num) => Self::new(-f64::from(num)),
+            Self::Integer(num) if num == 0 => Self::new(-f64::from(0)),
```
jedel1043, line 397:
```suggestion
            Self::Integer(num) if num == 0 => Self::new(f64::from(0)),
```
This is returning `-0.0f64` when `num == 0`, which is incorrect; the negation of `0i32` is `0i32`. That's probably why the tests fail.

jedel1043 (changes requested):

jedel1043 (commented):
I extracted one of the checks in the failing test:
```Javascript
if (-1 % -1 !== -0) {
  $ERROR('#2.1: -1 % -1 === 0. Actual: ' + (-1 % -1));
} else {
  if (1 / (-1 % -1) !== Number.NEGATIVE_INFINITY) {
    $ERROR('#2.2: -1 % -1 === - 0. Actual: +0');
  }
}
```
EcmaScript requires that `-1 % -1` must have the information of the sign in order to return `f64::NEGATIVE_INFINITY` in case we divide a number by `-1 % -1`.

With the changes on this commit, `-1i32 % -1i32` returns `0i32`, making `1 / (-1 % -1)  === f64::POSITIVE_INFINITY`. This fails the second check.

So, to pass this test we would need to change the `rem` operator to return a `-0.0f64` if `x < 0` in `x % y = 0`

HalidOdat:
Yup. The integer module operation is not spec compliant, ill try to look into it this week

jedel1043:
Fixed the mod operator.

raskad (approved):

PR #1485, opened 2021-08-19
Implement abstract operation `CreateArrayFromList`

jedel1043:
This Pull Request implements the abstract operation `CreateArrayFromList`.

It changes the following:

- Creates new function `Array::create_array_from_list`
- Replaces calls to other functions with calls to the above.
- Removes the function `Array::construct_array`, as it is equivalent to `create_array_from_list`

### DIFF
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -233,7 +233,7 @@ impl Array {
         array.ordinary_define_own_property(
             "length".into(),
             PropertyDescriptor::builder()
-                .value(length as f64)
+                .value(length)
                 .writable(true)
                 .enumerable(false)
                 .configurable(false)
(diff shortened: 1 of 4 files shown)


boa/src/builtins/array/mod.rs
```diff
@@ -233,6 +233,7 @@ impl Array {
         array.ordinary_define_own_property(
             "length".into(),
             PropertyDescriptor::builder()
+                // todo: check if we can refactor JsValue to hold u32
                 .value(length as f64)
```
HalidOdat, line 237:
```suggestion
                .value(length)
```

It should be able to convert `u32` to `JsValue` 

Edit: https://github.com/boa-dev/boa/blob/master/boa/src/value/conversions.rs#L58-L67

HalidOdat:
| Passed | 29,766 | 29,766 | 0 |
| Failed | 33,519 | 33,519 | 0 |
| Panics | 2 | 2 | 0 |
| Conformance | 37.73% | 37.73% | 0.00% |

boa/src/builtins/array/mod.rs
jedel1043: Nice! Let me change it then.

HalidOdat (approved):

PR #1495, opened 2021-08-23
Refactor `JSON.stringify`

raskad:
It changes the following:

- Refactor `JSON.stringify`

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -13,17 +13,20 @@
 //! [json]: https://www.json.org/json-en.html
 //! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON
 
+use std::collections::HashSet;
+
 use crate::{
-    builtins::BuiltIn,
-    object::Object,
-    object::ObjectInitializer,
-    property::{Attribute, PropertyDescriptor, PropertyKey},
+    builtins::{
+        string::{is_leading_surrogate, is_trailing_surrogate},
+        BuiltIn,
+    },
+    object::{JsObject, ObjectInitializer, RecursionLimiter},
+    property::{Attribute, PropertyKey, PropertyNameKind},
     symbol::WellKnownSymbols,
     value::IntegerOrInfinity,
-    BoaProfiler, Context, JsResult, JsValue,
+    BoaProfiler, Context, JsResult, JsString, JsValue,
 };
-use serde::Serialize;
-use serde_json::{self, ser::PrettyFormatter, Serializer, Value as JSONValue};
+use serde_json::{self, Value as JSONValue};
 
 #[cfg(test)]
 mod tests;
@@ -128,7 +131,7 @@ impl Json {
     ///
     /// This `JSON` method converts a JavaScript object or value to a JSON string.
     ///
-    /// This medhod optionally replaces values if a `replacer` function is specified or
+    /// This method optionally replaces values if a `replacer` function is specified or
     /// optionally including only the specified properties if a replacer array is specified.
     ///
     /// An optional `space` argument can be supplied of type `String` or `Number` that's used to insert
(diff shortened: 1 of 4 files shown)


raskad:
  | Passed | 31,283 | 31,318 | **+35** |
  | Failed | 32,002 | 31,967 | **-35** |
  | Panics | 2 | 2 | 0 |
  | Conformance | 39.65% | 39.69% | **+0.04%** |
  
  ```
  test/built-ins/Map/prototype/set/append-new-values-normalizes-zero-key.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-array-duplicates.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-array-duplicates.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-tojson.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-tojson.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-array-empty.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-array-empty.js (previously Failed)
  test/built-ins/JSON/stringify/value-boolean-object.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/value-boolean-object.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-arguments.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-arguments.js (previously Failed)
  test/built-ins/JSON/stringify/space-string-object.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/space-string-object.js (previously Failed)
  test/built-ins/JSON/stringify/value-number-object.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/value-number-object.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-array-undefined.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-array-undefined.js (previously Failed)
  test/built-ins/JSON/stringify/value-bigint-replacer.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/value-bigint-replacer.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-array-circular.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-array-circular.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-result-undefined.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-result-undefined.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-abrupt.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-abrupt.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-result.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-result.js (previously Failed)
  test/built-ins/JSON/stringify/value-string-object.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/value-string-object.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-object-deleted-property.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-object-deleted-property.js (previously Failed)
  test/built-ins/JSON/stringify/space-number-object.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/space-number-object.js (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-object-circular.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/replacer-function-object-circular.js (previously Failed)
  test/built-ins/JSON/stringify/value-bigint.js [strict mode] (previously Failed)
  test/built-ins/JSON/stringify/value-bigint.js (previously Failed)
  ```
  </details>
  
  ```
  test/built-ins/JSON/stringify/replacer-array-number.js [strict mode] (previously Passed)
  test/built-ins/JSON/stringify/replacer-array-number.js (previously Passed)
  ```
  </details>

boa/src/builtins/json/mod.rs
```diff
@@ -145,133 +148,542 @@ impl Json {
...
+                    for _ in 0..i {
+                        s.push(' ');
+                    }
+                    s.into()
+                }
+            }
+        // 7. Else if Type(space) is String, then
+        } else if let Some(s) = space.as_string() {
+            // a. If the length of space is 10 or less, let gap be space; otherwise let gap be the substring of space from 0 to 10.
+            if s.len() <= 10 {
```
jedel1043, line 258: Wouldn't this count the number of bytes in `s`? I think this should be encoded to utf_16 before getting its length.

raskad: True. Nice catch!

boa/src/builtins/json/mod.rs
```diff
@@ -145,133 +148,542 @@ impl Json {
...
+            }
+        // 7. Else if Type(space) is String, then
+        } else if let Some(s) = space.as_string() {
+            // a. If the length of space is 10 or less, let gap be space; otherwise let gap be the substring of space from 0 to 10.
+            if s.encode_utf16().count() <= 10 {
+                s.clone()
             } else {
-                JsValue::new("")
+                String::from_utf16_lossy(&s.encode_utf16().take(10).collect::<Vec<u16>>()).into()
             }
```
jedel1043, line 262: Now that I think about it, can't we just `take(10)` directly? the `Take` iterator should yield less than 10 elements if the iterator empties before reaching 10.

jedel1043: I think that's the reason why now it has one less passing test. Before the change that test entered the `else` branch and converted to utf16 before clamping to 10. Now it just clones itself without converting it to utf16.

raskad: I though about doing `take(10)` direly as well at some point, but dropped it for whatever reason. Just ran that and the tests look good and I think it should work as expected.

About the one less test; that should be `test/built-ins/Map/prototype/set/append-new-values-normalizes-zero-key.js` as far as I see it. It has no code relating to `JSON.stringify` and it is a `Map` test. So I think this one is probably failing most(?) of the time and only passes by accident when the underlying map structure returns items in the expected order? But I don't think it should be related to this.

boa/src/builtins/json/mod.rs
```diff
@@ -145,133 +148,538 @@ impl Json {
...
+                    // 2. Let len be ? LengthOfArrayLike(replacer).
+                    let len = replacer_obj.length_of_array_like(context)?;
+
+                    // 3. Let k be 0.
+                    let mut k = 0;
+
+                    // 4. Repeat, while k < len,
+                    while k < len {
+                        // a. Let prop be ! ToString(𝔽(k)).
+                        let prop = k.to_string();
```
HalidOdat, line 186:
```suggestion
                        let prop = k;
```
we don't have to call `to_string` on k since this would just cause an unnecessary string allocation,  we have `PropertyKey::index` we can just pass an integer to `.get`

boa/src/builtins/json/mod.rs
```diff
@@ -145,133 +148,538 @@ impl Json {
...
+        }
+
+        // 2. Append value to state.[[Stack]].
+        state.stack.push(value.clone().into());
+
+        // 3. Let stepback be state.[[Indent]].
+        let stepback = state.indent.clone();
+
+        // 4. Set state.[[Indent]] to the string-concatenation of state.[[Indent]] and state.[[Gap]].
+        state.indent = format!("{}{}", state.indent.as_str(), state.gap).into();
```
HalidOdat, line 481:
```suggestion
        state.indent = JsString::concat(state.indent, state.gap);
```
We can use `JsString::concat` which is more allocation efficient.

boa/src/builtins/json/mod.rs
```diff
@@ -145,133 +148,538 @@ impl Json {
...
```
HalidOdat, line 598:
```suggestion
        state.indent = JsString::concat(state.indent, state.gap);
```
Here too.

boa/src/builtins/json/mod.rs
raskad: That is a nice function. Can you give some guidance on when the `format!` macro is more efficient? I would guess, if the concat is not only existing `JsStrings`? How about calling it multiple time like `JsString::concat(a, JsString::concat(b, c))`?

boa/src/builtins/json/mod.rs
raskad: Done :)

boa/src/builtins/json/mod.rs
jedel1043: `format!` is less efficient because we create a new `String` in the heap, we clone both `str` into the new `String`, then we allocate a `JsString`, we clone the `String` into the JsString` and then we return it.
On the other hand, `JsString::concat` takes any two `JsString`, allocates enough space to store both in succession and then clones both into the allocated space. This saves us one `String` allocation and one clone.

HalidOdat (approved):
Looks good to me!

Razican (approved):

PR #1496, opened 2021-08-23
Fix panic on example `classes.rs`

### LINKED ISSUE #1262: Can't compile example classes.rs
Hello, thanks for the cool project.

I cannot compile the classes example:
https://github.com/boa-dev/boa/blob/master/boa/examples/classes.rs

I imported boa like this: `Boa = "0.11.0"`
I get the same error on both stable and nightly.

```
$ cargo run

    Blocking waiting for file lock on build directory
   Compiling jsint v0.1.0 (/tmp/jsint)
error[E0433]: failed to resolve: could not find `gc` in the list of imported crates
  --> src/main.rs:15:17
   |
15 | #[derive(Debug, Trace, Finalize)]
   |                 ^^^^^ could not find `gc` in the list of imported crates
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `gc` in the list of imported crates
  --> src/main.rs:15:24
   |
15 | #[derive(Debug, Trace, Finalize)]
   |                        ^^^^^^^^ could not find `gc` in the list of imported crates
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `gc` in the list of imported crates
  --> src/main.rs:15:17
   |
15 | #[derive(Debug, Trace, Finalize)]
   |                 ^^^^^ not found in `gc`
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing one of these items
   |
1  | use boa::gc::Trace;
   |
1  | use crate::Trace;
   |

error[E0433]: failed to resolve: could not find `gc` in the list of imported crates
  --> src/main.rs:15:17
   |
15 | #[derive(Debug, Trace, Finalize)]
   |                 ^^^^^ not found in `gc`
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing one of these items
   |
1  | use boa::gc::Finalize;
   |
1  | use crate::Finalize;
   |

error: aborting due to 4 previous errors

```

I get the same error every time I try to derive Trace or Finalize.
I tested with both the stable and nightly toolchain.

- OS: debian 
- Version: testing
- Target triple: Debian 5.10.24-1 (2021-03-19) x86_64 GNU/Linux
- Rustc version: rustc 1.54.0-nightly (5c0292654 2021-05-11)
- Stable rust: rustc 1.52.1 (9bc8c42bb 2021-05-09)

jedel1043:
This Pull Request fixes/closes #1262.

It changes the following:

- Initializes the prototype of a new `ClassBuilder` to `Object object`.
- Improves the documentation of the example `classes.rs`.

### DIFF
--- a/boa/src/class.rs
+++ b/boa/src/class.rs
@@ -62,7 +62,7 @@
 
 use crate::{
     builtins::function::NativeFunction,
-    object::{ConstructorBuilder, GcObject, NativeObject, ObjectData},
+    object::{ConstructorBuilder, GcObject, NativeObject, ObjectData, PROTOTYPE},
     property::{Attribute, PropertyDescriptor, PropertyKey},
     Context, JsResult, JsValue,
 };
@@ -87,7 +87,7 @@ pub trait Class: NativeObject + Sized {
 ///
 /// This is automatically implemented, when a type implements `Class`.
 pub trait ClassConstructor: Class {
-    /// The raw constructor that mathces the `NativeFunction` signature.
+    /// The raw constructor that matches the `NativeFunction` signature.
     fn raw_constructor(
         this: &JsValue,
         args: &[JsValue],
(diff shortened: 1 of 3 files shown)


Razican:
I like the changes, thank you! Could you add a test in the CI pipeline to check that it works, by running it?

Razican (approved):

jedel1043:
> I like the changes, thank you! Could you add a test in the CI pipeline to check that it works, by running it?

Okay!

Razican (approved):

RageKnify (approved):

boa/src/class.rs
```diff
@@ -116,7 +116,11 @@ impl<'context> ClassBuilder<'context> {
     where
         T: ClassConstructor,
     {
-        let mut builder = ConstructorBuilder::new(context, T::raw_constructor);
+        let mut builder = ConstructorBuilder::with_standard_object(
```
HalidOdat, line 119: We should not use `with_standard_object` since it will override the standard object in this case `Object` object.

HalidOdat (commented):
We should set the prototype instance in `raw_constructor` of this argument to be the prototype of the class name. Like we do in `Map`, `Set`

boa/src/class.rs
jedel1043: Ah, I missed that. I have to create a new object with `[[Prototype]]: Object` and set that new object as the prototype of the builder, correct?

jedel1043:
> We should set the prototype instance in `raw_constructor` of this argument to be the prototype of the class name. Like we do in `Map`, `Set`

I don't think that's optimal. We are presenting the `Class` trait as an equivalent to a Javascript `class`, and classes automatically inherit `Object` as the prototype of their prototypes. We should do this for the user.

boa/src/class.rs
HalidOdat: > Ah, I missed that. I have to create a new object with `[[Prototype]]: Object` and set that new object as the prototype of the builder, correct?

`ConstructorBuilder` creates a new constructor like `Object()`, `String()`, etc, it does not create a new object instance.

HalidOdat:
> I don't think that's optimal. We are presenting the `Class` trait as an equivalent to a Javascript `class`, and classes automatically inherit `Object` as the prototype of their prototypes. We should do this for the user.

Ummm... Yes. That what we want, the classes prototype already inherits from `Object.prototype`, so we just need to set the class _instances_ `[[prototype]]` to `className.prototype`. That is why it panics because it can't find the method in the prototype because the prototype is not set.

jedel1043:
@HalidOdat Thank you for the feedback! I didn't completely understand how we set the `[[prototype]]` of an instance. It should now set `[[prototype]]` correctly.

HalidOdat (approved):
Looks good to me :)

PR #1499, opened 2021-08-23
Updated the Test262 submodule

Razican:
This updates the Test262 submodule to have the latest tests from the official suite.


jedel1043:
Passing test:

jedel1043:
Failing test:
```Javascript
// CHECK#1
verifyNotConfigurable(Boolean, "prototype");

assert.throws(TypeError, () => {
  delete Boolean.prototype;
});
```

Previous version of the same test:
```Javascript
// CHECK#1
verifyNotConfigurable(Boolean, "prototype");

try {
  if (delete Boolean.prototype !== false) {
    $ERROR('#1: Boolean.prototype has the attribute DontDelete');
  }
} catch (e) {
  if (e instanceof Test262Error) throw e;
  assert(e instanceof TypeError);
}
```

`delete` should throw on error, but on

https://github.com/boa-dev/boa/blob/62ad20cfde0a545f35d8af07c36cd5d32a7479f4/boa/src/syntax/ast/node/operator/unary_op/mod.rs

we use `__delete__`, which doesn't throw on error. We should replace it with `DeletePropertyOrThrow`

HalidOdat (approved):

HalidOdat:
@jedel1043 I think we should leave fixing broken tests for a separate PR, we should create an issue for this to keep track of it :)

jedel1043:
> @jedel1043 I think we should leave fixing broken tests for a separate PR, we should create an issue for this to keep track of it :)

Yep, just commenting to keep track on what we need to fix

PR #1354, opened 2021-06-22
Proposal of new `Object` design

jedel1043:
Currently some tests fail because we aren't considering the many overrides of `Object` internal functions that `ExoticObjects` do.

The discussion of this issue is in #591, and the proposed solution is to implement every override that the spec requires directly inside `Object`, considering every exotic and special object implementation. This works, but it's a little detrimental for the modularity of the design. You can find a draft of this solution on #600.

This draft is a prototype of an alternative design based on traits and enum dispatch that solves the same problem.
The code is obviously WIP, so please let me know what you think can be improved.

PR #1502, opened 2021-08-23
Adjust `EnumerableOwnPropertyNames` to use all `String` type property keys

### LINKED ISSUE #1501: `Object.keys` does not work with index property keys
**Describe the bug**
`Object.keys` does not work with index property keys

**To Reproduce**

```javascript
let o = [1, 2, 3];
Object.keys(o);
```
The output is `[]`

**Expected behavior**
The output should be `[ '0', '1', '2' ]`

**Build environment:**

- OS: Linux
- Version: eaaae8544
- Target triple:x86_64-unknown-linux-gnu
- Rustc version: rustc 1.53.0 (53cb7b09b 2021-06-17)

raskad:
This Pull Request fixes/closes #1501.

It changes the following:

- Adjust `EnumerableOwnPropertyNames` to use all `String` type property keys

This looks different from the [spec](https://tc39.es/ecma262/#sec-enumerableownpropertynames), because the meaning of `String` in `4.a: If Type(key) is String, then` is different than our internal `PropertyKey` enum types `String`, `Symbol` and `Index`. Our internal `Index` type is a `String` type in the context of the spec.
This PR is only a fix for this specific function. We should change our `[[OwnPropertyKeys]]` to be spec compliant, but I think we should wait with that until #1354 is merged, because `[[OwnPropertyKeys]]` is defined multiple times for different object types.

### DIFF
--- a/boa/src/object/internal_methods.rs
+++ b/boa/src/object/internal_methods.rs
@@ -902,12 +902,10 @@ impl GcObject {
         // 4. For each element key of ownKeys, do
         for key in own_keys {
             // a. If Type(key) is String, then
-            let key_str = if let PropertyKey::Index(index) = &key {
-                Some(index.to_string().into())
-            } else if let PropertyKey::String(key_str) = &key {
-                Some(key_str.clone())
-            } else {
-                None
+            let key_str = match &key {
+                PropertyKey::String(s) => Some(s.clone()),
+                PropertyKey::Index(i) => Some(i.to_string().into()),
+                _ => None,
             };
 
             if let Some(key_str) = key_str {
--- a/boa/src/object/internal_methods.rs
+++ b/boa/src/object/internal_methods.rs
@@ -902,7 +902,15 @@ impl GcObject {
         // 4. For each element key of ownKeys, do
         for key in own_keys {
             // a. If Type(key) is String, then
-            if let PropertyKey::String(key_str) = &key {
+            let key_str = if let PropertyKey::Index(index) = &key {
+                Some(index.to_string().into())
+            } else if let PropertyKey::String(key_str) = &key {
+                Some(key_str.clone())
+            } else {
+                None
+            };
+
+            if let Some(key_str) = key_str {
                 // i. Let desc be ? O.[[GetOwnProperty]](key).
                 let desc = self.__get_own_property__(&key);
                 // ii. If desc is not undefined and desc.[[Enumerable]] is true, then
@@ -910,7 +918,7 @@ impl GcObject {
                     if desc.expect_enumerable() {
                         // 1. If kind is key, append key to properties.
                         if let PropertyNameKind::Key = kind {
-                            properties.push(key_str.clone().into())
+                            properties.push(key_str.into())
                         }
                         // 2. Else,
                         else {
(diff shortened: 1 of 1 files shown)


raskad:
The broken test is actually broken and should not have passed before ;)

  | Passed | 31,283 | 31,287 | **+4** |
  | Failed | 32,002 | 31,998 | **-4** |
  | Panics | 2 | 2 | 0 |
  | Conformance | 39.65% | 39.66% | **+0.01%** |
  
  ```
  test/built-ins/Object/keys/15.2.3.14-5-14.js [strict mode] (previously Failed)
  test/built-ins/Object/keys/15.2.3.14-5-14.js (previously Failed)
  test/built-ins/Object/keys/15.2.3.14-3-4.js [strict mode] (previously Failed)
  test/built-ins/Object/keys/15.2.3.14-3-4.js (previously Failed)
  test/built-ins/Object/keys/15.2.3.14-3-3.js [strict mode] (previously Failed)
  test/built-ins/Object/keys/15.2.3.14-3-3.js (previously Failed)
  ```
  </details>
  
  ```
  test/built-ins/Object/keys/15.2.3.14-5-11.js [strict mode] (previously Passed)
  test/built-ins/Object/keys/15.2.3.14-5-11.js (previously Passed)
  ```
  </details>

HalidOdat (approved):

boa/src/object/internal_methods.rs
```diff
@@ -902,15 +902,23 @@ impl GcObject {
...
         for key in own_keys {
             // a. If Type(key) is String, then
-            if let PropertyKey::String(key_str) = &key {
+            let key_str = if let PropertyKey::Index(index) = &key {
+                Some(index.to_string().into())
+            } else if let PropertyKey::String(key_str) = &key {
+                Some(key_str.clone())
+            } else {
+                None
+            };
```
HalidOdat, line 911:
```suggestion
            let key_str = if let PropertyKey::Index(index) = &key {
                Some(index.to_string().into())
            } else if let PropertyKey::String(key_str) = &key {
                Some(key_str.clone())
            } else {
                None
            };
```
I think a `match` here would make it more readable :)

boa/src/object/internal_methods.rs
```diff
@@ -902,15 +902,23 @@ impl GcObject {
...
+
+            if let Some(key_str) = key_str {
                 // i. Let desc be ? O.[[GetOwnProperty]](key).
                 let desc = self.__get_own_property__(&key);
                 // ii. If desc is not undefined and desc.[[Enumerable]] is true, then
                 if let Some(desc) = desc {
                     if desc.expect_enumerable() {
                         // 1. If kind is key, append key to properties.
                         if let PropertyNameKind::Key = kind {
```
HalidOdat, line 920: We could also match on `PropertyNameKind` too

boa/src/object/internal_methods.rs
raskad: True, looks way better.

boa/src/object/internal_methods.rs
raskad: Changed that in bb25549

HalidOdat (approved):

Razican (approved):
Looks very good, nice catch!

PR #1509, opened 2021-08-24
Implement `JsString::concat_array()`

HalidOdat:
There are a lot of places in the spec where we concat more strings than two, so we can avoid unnecessary string allocations caused by `format!`.

### DIFF
--- a/boa/src/string.rs
+++ b/boa/src/string.rs
@@ -63,10 +63,13 @@ impl Inner {
         unsafe { NonNull::new_unchecked(inner) }
     }
 
-    /// Concatinate two string.
+    /// Concatenate array of strings.
     #[inline]
-    fn concat(x: &str, y: &str) -> NonNull<Inner> {
-        let total_string_size = x.len() + y.len();
+    fn concat_array(strings: &[&str]) -> NonNull<Inner> {
+        let mut total_string_size = 0;
+        for string in strings {
+            total_string_size += string.len();
+        }
 
         // We get the layout of the `Inner` type and we extend by the size
         // of the string array.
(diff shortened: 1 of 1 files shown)


boa/src/string.rs
```diff
@@ -60,10 +60,13 @@ impl Inner {
...
-    /// Concatinate two string.
+    /// Concatinate array of strings.
     #[inline]
-    fn concat(x: &str, y: &str) -> NonNull<Inner> {
-        let total_string_size = x.len() + y.len();
+    fn concat_array(strings: &[&str]) -> NonNull<Inner> {
+        let mut total_string_size = 0;
+        for string in strings {
+            total_string_size += string.len();
+        }
```
jedel1043, line 69: Maybe it would be useful to accept any slice of `&[S]` where `S: AsRef<str>` to be able to pass a `&[JsString]`, but I don't know if we even concat several `JsString` anywhere in the codebase.

HalidOdat: > Maybe it would be useful to accept any slice of `&[S]` where `S: AsRef<str>` to be able to pass a `&[JsString]`.

Hmmm. I'll add that, since this is exposed to the user it should also ergonomic

> but I don't know if we even concat several JsString anywhere in the codebase.

There are some in JSON parse and stringify (#1495 ) also in `RegExp.prototype.toString ( )`, `String.prototype.replaceAll`, `Date.prototype.toUTCString ( )`, and many other places just do text search `string-concatenation` and you'll see them.

jedel1043: > There are some in JSON parse and stringify (#1495 ) also in `RegExp.prototype.toString ( )`, `String.prototype.replaceAll`, `Date.prototype.toUTCString ( )`, and many other places just do text search `string-concatenation` and you'll see them.

Ah yeah, we do concat more than two `JsString` on several places. Then yeah, we would avoid allocating `String`s with this change.

boa/src/string.rs
```diff
@@ -60,10 +60,13 @@ impl Inner {
         unsafe { NonNull::new_unchecked(inner) }
     }
 
-    /// Concatinate two string.
+    /// Concatinate array of strings.
```
Razican, line 63: It should say "concat**e**nate"

Razican (approved):
Looks good, found a typo. Should we already search for `format!("{}{}` and see if something can already be replaced?

PR #1510, opened 2021-08-24
Implement `Object.prototype.valueOf`

HalidOdat:
It changes the following:
- Implement `Object.prototype.valueOf`


Razican (approved):
Interesting how some tests failed :/ But the implementation looks definitely good. Maybe they were passing before by chance.

HalidOdat:
> Interesting how some tests failed :/ But the implementation looks definitely good. Maybe they were passing before by chance.

Yup. They are false positives.

PR #1511, opened 2021-08-24
Implement `Object.seal/isSealed/freeze/isFrozen`

HalidOdat:
It changes the following:
- Implement `Object.seal( o )`
- Implement `Object.isSealed( o )`
- Implement `Object.freeze( o )`
- Implement `Object.isFrozen( o )`
- Implement `SetIntegrityLevel` spec function
- Implement `TestIntegrityLevel` spec function
- Add `IntegrityLevel` enum 
- Make some object operations public, like `JsObject::get`, `JsObject::set`, etc

### DIFF
--- a/boa/src/object/gcobject.rs
+++ b/boa/src/object/gcobject.rs
@@ -51,7 +51,7 @@ enum FunctionBody {
 }
 
 impl JsObject {
-    /// Create a new `JsObject` from a `Object`.
+    /// Create a new `GcObject` from a `Object`.
     #[inline]
     pub fn new(object: Object) -> Self {
         Self(Gc::new(GcCell::new(object)))
--- a/boa/src/builtins/array/mod.rs
+++ b/boa/src/builtins/array/mod.rs
@@ -230,7 +230,6 @@ impl Array {
         array.borrow_mut().data = ObjectData::array();
 
         // 6. Perform ! OrdinaryDefineOwnProperty(A, "length", PropertyDescriptor { [[Value]]: 𝔽(length), [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: false }).
-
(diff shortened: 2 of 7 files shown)


boa/src/object/gcobject.rs
```diff
@@ -51,6 +51,24 @@ enum FunctionBody {
     Ordinary(RcStatementList),
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
Razican, line 54: I think we should add some documentation here.

boa/src/object/gcobject.rs
```diff
@@ -992,6 +1010,112 @@ impl GcObject {
         // 7. Return target.
         Ok(())
     }
+
+    pub fn set_integrity_level(
```
Razican, line 1014: This should also have documentation.

boa/src/object/gcobject.rs
```diff
@@ -992,6 +1010,112 @@ impl GcObject {
...
+                    }
+                }
+            }
+        }
+
+        // 8. Return true.
+        Ok(true)
+    }
+
+    pub fn test_integrity_level(
```
Razican, line 1077: Another place for documentation :)

Razican (commented):
Looks very good :) just found some definitions without documentation, and being in the public API, they should have some.

boa/src/object/gcobject.rs
jedel1043: Also, could you move this function to `object::operations`?

boa/src/object/gcobject.rs
jedel1043: This function also should be on `object::operations`

PR #1523, opened 2021-08-28
Allow moving `NativeObject` variables into closures as external captures

jedel1043:
This Pull Request follows from #1518.

It changes the following:

- Allows moving any `NativeObject + Clone` type as a parameter into any closure
- Adds test for the new feature

This unblocks the implementation of arguments exotic objects.
The new test is a good example of how this new feature works:
```Rust
let js_function = FunctionBuilder::closure_with_captures(
        &mut context,
        |_, _, context, captures| {
            println!("Called `createMessage`");
            // We obtain the `name` property of `captures.object`
            let name = captures.object.get("name", context)?;

            // We create a new message from our captured variable.
            let message = JsString::concat_array(&[
                "message from `",
                name.to_string(context)?.as_str(),
                "`: ",
                captures.greeting.as_str(),
            ]);

            println!("{}", message);

            // We convert `message` into `Jsvalue` to be able to return it.
            Ok(message.into())
        },
        // Here is where we move `clone_variable` into the closure.
        clone_variable,
    )
```

### DIFF
--- a/boa/examples/closures.rs
+++ b/boa/examples/closures.rs
@@ -1,21 +1,110 @@
-use boa::{Context, JsValue};
+// This example goes into the details on how to pass closures as functions
+// inside Rust and call them from Javascript.
+
+use boa::{
+    gc::{Finalize, Trace},
+    object::{FunctionBuilder, JsObject},
+    property::{Attribute, PropertyDescriptor},
+    Context, JsString, JsValue,
+};
 
 fn main() -> Result<(), JsValue> {
+    // We create a new `Context` to create a new Javascript executor.
     let mut context = Context::new();
 
-    let variable = "I am a captured variable";
+    // We make some operations in Rust that return a `Copy` value that we want
+    // to pass to a Javascript function.
+    let variable = 128 + 64 + 32 + 16 + 8 + 4 + 2 + 1;
 
     // We register a global closure function that has the name 'closure' with length 0.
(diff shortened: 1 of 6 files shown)


raskad (approved):
Looks good to me.
Maybe you could add the test (or something similar) to the examples? `NativeObject` is pretty abstract and it may not be clear what you can do with `closure_with_captures` without an example.

jedel1043:
> Looks good to me.
> Maybe you could add the test (or something similar) to the examples? `NativeObject` is pretty abstract and it may not be clear what you can do with `closure_with_captures` without an example.

Okay, let me add an example

jedel1043:
> Looks good to me.
> Maybe you could add the test (or something similar) to the examples? `NativeObject` is pretty abstract and it may not be clear what you can do with `closure_with_captures` without an example.

Added an example inside `closures.rs`. Also slightly modified the API to automatically downcast to the passed type. Let me know if you like the new changes.

raskad (approved):
Wow, the automatic downcast looks really nice from an API consumer perspective. We only have to make sure that the downcast never fails, otherwise the error is really confusing.

The example looks great!

raskad:
I just noticed that there are some lints in this latest version. You may want to fix those.

jedel1043:
Rebased

HalidOdat (approved):
Looks good to me! :)

PR #1525, opened 2021-08-28
Implement `Object.preventExtensions()` and `Object.isExtensible()`

HalidOdat:
It changes the following:
- Implement `Object.preventExtensions()`
- Implement `Object.isExtensible()`


jedel1043 (approved):
Checked the failing tests and they are related to the `eval` function, which we haven't implemented yet, so this should be good!

Razican (approved):
Great work @HalidOdat !!

PR #1528, opened 2021-08-29
Always return `undefined` from functions that do not `return`

raskad:
This Pull Request fixes/closes #672.

It changes the following:

- Always return `undefined` from functions that do not `return`
- Fix `JSON.parse` test

PR #1535, opened 2021-08-29
Refactor `JSON.parse`

raskad:
Fixes the new test failure from #1528.

It changes the following:

- Refactor `JSON.parse`

### DIFF
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -139,11 +139,9 @@ impl Json {
             if obj.is_array() {
                 // i. Let I be 0.
                 // ii. Let len be ? LengthOfArrayLike(val).
-                let mut i = 0;
-                let len = obj.length_of_array_like(context)? as i64;
-
                 // iii. Repeat, while I < len,
-                while i < len {
+                let len = obj.length_of_array_like(context)? as i64;
+                for i in 0..len {
                     // 1. Let prop be ! ToString(𝔽(I)).
                     // 2. Let newElement be ? InternalizeJSONProperty(val, prop, reviver).
                     let new_element = Self::internalize_json_property(
@@ -163,9 +161,6 @@ impl Json {
                         // a. Perform ? CreateDataProperty(val, prop, newElement).
                         obj.create_data_property(i, new_element, context)?;
                     }
-
-                    // 5. Set I to I + 1.
-                    i += 1;
                 }
             }
             // c. Else,
--- a/boa/src/builtins/json/mod.rs
+++ b/boa/src/builtins/json/mod.rs
@@ -19,7 +19,7 @@ use crate::{
         BuiltIn,
     },
     object::{JsObject, ObjectInitializer, RecursionLimiter},
-    property::{Attribute, PropertyKey, PropertyNameKind},
+    property::{Attribute, PropertyNameKind},
     symbol::WellKnownSymbols,
     value::IntegerOrInfinity,
     BoaProfiler, Context, JsResult, JsString, JsValue,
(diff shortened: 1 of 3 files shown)


raskad:
  | Total | 80,685 | 80,685 | 0 |
  | Passed | 32,784 | 32,790 | **+6** |
  | Ignored | 15,818 | 15,818 | 0 |
  | Failed | 32,083 | 32,077 | **-6** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 40.63% | 40.64% | **+0.01%** |
  
  <details><summary><b>Fixed tests (6):</b></summary>
  
  ```
  test/built-ins/JSON/parse/reviver-get-name-err.js [strict mode] (previously Failed)
  test/built-ins/JSON/parse/reviver-get-name-err.js (previously Failed)
  test/built-ins/JSON/parse/reviver-array-get-prop-from-prototype.js [strict mode] (previously Failed)
  test/built-ins/JSON/parse/reviver-array-get-prop-from-prototype.js (previously Failed)
  test/built-ins/JSON/parse/text-negative-zero.js [strict mode] (previously Failed)
  test/built-ins/JSON/parse/text-negative-zero.js (previously Failed)
  ```
  </details>

boa/src/builtins/json/mod.rs
```diff
@@ -72,59 +72,136 @@ impl Json {
...
+            // a. Let isArray be ? IsArray(val).
+            // b. If isArray is true, then
+            if obj.is_array() {
+                // i. Let I be 0.
+                // ii. Let len be ? LengthOfArrayLike(val).
+                let mut i = 0;
+                let len = obj.length_of_array_like(context)? as i64;
+
+                // iii. Repeat, while I < len,
+                while i < len {
```
jedel1043, line 146:
```suggestion
                for i in 0..len {
```
Better to use a range instead of an additional variable.

boa/src/builtins/json/mod.rs
```diff
@@ -72,59 +72,136 @@ impl Json {
...
+                    // 1. Let prop be ! ToString(𝔽(I)).
+                    // 2. Let newElement be ? InternalizeJSONProperty(val, prop, reviver).
+                    let new_element = Self::internalize_json_property(
+                        obj.clone(),
+                        i.to_string().into(),
```
jedel1043, line 151: Hmm... I'm doubtful if applying `to_string` to an indexed key is a good idea. Can you confirm if making `name` a `Into<PropertyKey>` instead of a `JsString` solves any tests?

jedel1043 (changes requested):

boa/src/builtins/json/mod.rs
raskad: `name` must be some kind of string for the `call` at the end of the function. If I change it into a `Into<PropertyKey>`, the `JsValue` the gets passed to the `call` function would be numeric in some cases, because `PropertyKey.into::<JsValue>()` may convert into a numeric `JsValue`.

boa/src/builtins/json/mod.rs
raskad: Done

boa/src/builtins/json/mod.rs
jedel1043: The problem I see is that when you call `internalize_json_property` recursively for an array, it will try to `Get` the property index i, but converted into a string, so it will try to get the string property "i". However, for an array we store indices as an index property, so the `Get` would return `JsValue::Undefined`, I think.

raskad: Ah I see what you mean. But we do not have to worry about this here. The `From<JsString> for PropertyKey` implementation trys to parse the string into an u32 and returning it as `PropertyKey::Index(u32)`. If that fails it returns is as an `PropertyKey::String(JsString)`.

jedel1043: I see. That's okay then. I'd like to avoid casting an index to a string and then back into an index but maybe that'll make the code less readable.

jedel1043 (approved):
Everything looks good. Good work!

HalidOdat (approved):
Looks good to me! :)

HalidOdat:
Just needs a rebase.

PR #1552, opened 2021-09-03
Implement Typed Arrays

Razican:
Hello team! Currently we have at least hundreds or thousands of tests that are failing due to not having Typed Arrays. Even the tester `$262` object is supposed to have them. This PR is supposed to end up implementing Typed Arrays, but I wanted to bring the changes to your attention before moving further (also, because I will have limited free time during some time).

This Pull Request fixes/closes #893 (when it's finished).

It changes the following:

- Moves `GetMethod()` to `JsValue`
- Creates `GetV`
- Fixes a bunch of non compliant stuff.

Feel free to comment or to even push to this branch.

PR #1561, opened 2021-09-07
Implement abstract operation `GetPrototypeFromConstructor`

jedel1043:
This Pull Request reduces the repetition on constructors when obtaining the default prototype of an object.
Extracted the pattern from #1552, thanks to @Razican :)

It changes the following:

- Implements the abstract operation `GetPrototypeFromConstructor`

- Replaces the pattern:
```Rust
let proto = new_target
        .as_object()
        .and_then(|obj| {
            obj.__get__(&PROTOTYPE.into(), obj.clone().into(), context)
                .map(|o| o.as_object())
                .transpose()
        })
        .transpose()?
        .unwrap_or_else(default_proto);
```
with
```Rust
let proto = get_prototype_from_constructor(
                new_target,
                |intrinsics| intrinsics.object_object().prototype(),
                context,
            )?;
```
This also allows for an easy replacement of  `intrinsics` with the intrinsic objects of a `realm` when we eventually implement multiple realms.

### DIFF
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -68,7 +68,7 @@ impl Boolean {
             return Ok(JsValue::new(data));
         }
         let prototype =
-            get_prototype_from_constructor(new_target, StandardObjects::object_object, context)?;
+            get_prototype_from_constructor(new_target, StandardObjects::boolean_object, context)?;
         let boolean = JsValue::new_object(context);
 
         boolean
--- a/boa/src/builtins/boolean/mod.rs
+++ b/boa/src/builtins/boolean/mod.rs
@@ -14,7 +14,8 @@ mod tests;
 
 use crate::{
     builtins::BuiltIn,
-    object::{ConstructorBuilder, ObjectData, PROTOTYPE},
+    context::StandardObjects,
+    object::{internal_methods::get_prototype_from_constructor, ConstructorBuilder, ObjectData},
     property::Attribute,
     BoaProfiler, Context, JsResult, JsValue,
 };
@@ -66,15 +67,8 @@ impl Boolean {
         if new_target.is_undefined() {
             return Ok(JsValue::new(data));
         }
-        let prototype = new_target
-            .as_object()
-            .and_then(|obj| {
-                obj.__get__(&PROTOTYPE.into(), obj.clone().into(), context)
-                    .map(|o| o.as_object())
-                    .transpose()
-            })
-            .transpose()?
-            .unwrap_or_else(|| context.standard_objects().object_object().prototype());
+        let prototype =
+            get_prototype_from_constructor(new_target, StandardObjects::object_object, context)?;
         let boolean = JsValue::new_object(context);
 
         boolean
--- a/boa/src/builtins/function/mod.rs
+++ b/boa/src/builtins/function/mod.rs
@@ -305,7 +305,7 @@ impl BuiltInFunctionObject {
         context: &mut Context,
     ) -> JsResult<JsValue> {
         let prototype =
-            get_prototype_from_constructor(new_target, StandardObjects::object_object, context)?;
+            get_prototype_from_constructor(new_target, StandardObjects::function_object, context)?;
         let this = JsValue::new_object(context);
 
         this.as_object()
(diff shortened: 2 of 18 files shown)


Razican:
Should this be implemented as part of `Context`?

raskad:
I'm not sure if the `intrinsicDefaultProto` should just be a `FnOnce(&Context) -> JsObject`. It is not probable, but I think this function could be misused with that definition. I think if we implement this, we should honor the `1. Assert: intrinsicDefaultProto is this specification's name of an intrinsic object.` somewhat. Maybe we could use an enum as the argument and do the `Context.standard_objects().some_object().prototype()` in the function?

The spec defines this as an ordinary object internal method, so I think the place is good.

boa/src/builtins/boolean/mod.rs
```diff
@@ -66,15 +66,11 @@ impl Boolean {
...
-            .and_then(|obj| {
-                obj.__get__(&PROTOTYPE.into(), obj.clone().into(), context)
-                    .map(|o| o.as_object())
-                    .transpose()
-            })
-            .transpose()?
-            .unwrap_or_else(|| context.standard_objects().object_object().prototype());
+        let prototype = get_prototype_from_constructor(
+            new_target,
+            |intrinsics| intrinsics.object_object().prototype(),
```
raskad, line 71: Should this be `boolean_object()`?

boa/src/builtins/function/mod.rs
```diff
@@ -302,15 +303,11 @@ impl BuiltInFunctionObject {
...
-            .and_then(|obj| {
-                obj.__get__(&PROTOTYPE.into(), obj.clone().into(), context)
-                    .map(|o| o.as_object())
-                    .transpose()
-            })
-            .transpose()?
-            .unwrap_or_else(|| context.standard_objects().object_object().prototype());
+        let prototype = get_prototype_from_constructor(
+            new_target,
+            |intrinsics| intrinsics.object_object().prototype(),
```
raskad, line 308: Should this be `function_object()`?

boa/src/builtins/number/mod.rs
```diff
@@ -166,15 +166,11 @@ impl Number {
...
-            .and_then(|obj| {
-                obj.__get__(&PROTOTYPE.into(), obj.clone().into(), context)
-                    .map(|o| o.as_object())
-                    .transpose()
-            })
-            .transpose()?
-            .unwrap_or_else(|| context.standard_objects().object_object().prototype());
+        let prototype = get_prototype_from_constructor(
+            new_target,
+            |intrinsics| intrinsics.object_object().prototype(),
```
raskad, line 171: Should this be `number_object()`?

boa/src/builtins/string/mod.rs
```diff
@@ -179,16 +180,11 @@ impl String {
...
-            .and_then(|obj| {
-                obj.__get__(&PROTOTYPE.into(), obj.clone().into(), context)
-                    .map(|o| o.as_object())
-                    .transpose()
-            })
-            .transpose()?
-            .unwrap_or_else(|| context.standard_objects().object_object().prototype());
+        let prototype = get_prototype_from_constructor(
+            new_target,
+            |intrinsics| intrinsics.object_object().prototype(),
```
raskad, line 185: Should this be `string_object()`?

raskad (commented):
Looks good to me.

I commented on a few places where we use `object_object()` while a specific `StandardConstructor` exists. I'm curious if changing those would result in actual changes.

jedel1043:
> Looks good to me.
> 
> I commented on a few places where we use `object_object()` while a specific `StandardConstructor` exists. I'm curious if changing those would result in actual changes.

Let's see if this works

raskad (approved):
Nice work!

jedel1043:
> Nice work!

Good catch!

Razican:
I think it would be easier to implement it by using `FnOnce(&StandardObjects) -> &StandardConstructor`. I did this in #1552 in my latest changes, but it would allow passing `StandardObjects::typed_array_object` and things like that, so directly the function.

jedel1043:
> I think it would be easier to implement it by using `FnOnce(&StandardObjects) -> &StandardConstructor`. I did this in #1552 in my latest changes, but it would allow passing `StandardObjects::typed_array_object` and things like that, so directly the function.

Yep, this is the best option.

Razican (approved):
Now it looks great! :) thanks!

PR #1567, opened 2021-09-09
Unify object creation with `empty` and `from_proto_and_data` methods

jedel1043:
Another step towards #577.

This Pull Request unifies `JsObject` creation by just having to specify an (optional) prototype and an `ObjectData` to create a new `JsObject`.

This refactors patterns like:
```Rust
let obj = context.construct_object();
obj.set_prototype_instance(prototype);
obj.borrow_mut().data = ObjectData::string("".into());
```
into a simpler:

```Rust
let obj = JsObject::from_proto_and_data(Some(prototype), ObjectData::string("".into()));
```

and the expression:
```Rust
let obj = JsObject::new(Object::Default());
```
is replaced with:
```Rust
let obj = JsObject::empty();
```

Also made the `JsObject::new` method private, because we only used it in the latter expression and in a few tests that can perfectly use `from_proto_and_data`


jedel1043:
Holding the ready for review until #1445 merges

boa/src/object/gcobject.rs
```diff
@@ -63,12 +60,32 @@ enum FunctionBody {
...
+    pub fn empty() -> Self {
+        Self::from_object(Object::default())
+    }
+
+    /// The more general form of `OrdinaryObjectCreate` and `MakeBasicObject`.
+    ///
+    /// Create a `JsObject` and automatically set its internal methods and
+    /// internal slots from the `data` provided.
+    #[inline]
+    pub fn from_proto_and_data(prototype: Option<JsObject>, data: ObjectData) -> Self {
```
Razican, line 80: I would use `Into<Option<JsObject>>` here, so that you can pass a `JsObject` without the `Some`.

RageKnify (approved):
LGTM, Razican's proposal sounds like a nice QoL improvement.

Razican (approved):
Looks good, just needs a rebase :)

PR #1568, opened 2021-09-09
Upgraded dependencies and Test262 submodule

Razican:
This just updates the Test262 submodule with the latest tests for the ongoing ECMAScript draft, and updates both JS and Rust dependencies.


jedel1043 (approved):
The failing tests are related to #1464, so this should be good!

PR #1572, opened 2021-09-11
Refactor builtin `Map` intrinsics to follow more closely the spec

jedel1043:
Changes:

- Change `OrderedMap` to reset its `empty_count` on `clear`.
- Document many methods of the `Map` object.
- Refactor some patterns to be more idiomatic.
- Implement `AddEntriesFromIterable`.
- Modify the API of `IteratorResult`.
- Move `Context` to the end of all related `Iterator` methods.

### DIFF
--- a/boa/src/builtins/iterable/mod.rs
+++ b/boa/src/builtins/iterable/mod.rs
@@ -25,22 +25,22 @@ impl IteratorPrototypes {
         let iterator_prototype = create_iterator_prototype(context);
         Self {
             array_iterator: ArrayIterator::create_prototype(
-                context,
                 iterator_prototype.clone().into(),
+                context,
             ),
-            set_iterator: SetIterator::create_prototype(context, iterator_prototype.clone().into()),
+            set_iterator: SetIterator::create_prototype(iterator_prototype.clone().into(), context),
             string_iterator: StringIterator::create_prototype(
-                context,
                 iterator_prototype.clone().into(),
+                context,
             ),
             regexp_string_iterator: RegExpStringIterator::create_prototype(
-                context,
                 iterator_prototype.clone().into(),
+                context,
             ),
-            map_iterator: MapIterator::create_prototype(context, iterator_prototype.clone().into()),
+            map_iterator: MapIterator::create_prototype(iterator_prototype.clone().into(), context),
             for_in_iterator: ForInIterator::create_prototype(
-                context,
                 iterator_prototype.clone().into(),
+                context,
             ),
             iterator_prototype,
         }
@@ -85,7 +85,7 @@ impl IteratorPrototypes {
 /// CreateIterResultObject( value, done )
 ///
 /// Generates an object supporting the IteratorResult interface.
-pub fn create_iter_result_object(context: &mut Context, value: JsValue, done: bool) -> JsValue {
+pub fn create_iter_result_object(value: JsValue, done: bool, context: &mut Context) -> JsValue {
     // 1. Assert: Type(done) is Boolean.
     // 2. Let obj be ! OrdinaryObjectCreate(%Object.prototype%).
     let obj = context.construct_object();
@@ -101,12 +101,12 @@ pub fn create_iter_result_object(context: &mut Context, value: JsValue, done: bo
 }
 
 /// Get an iterator record
-pub fn get_iterator(context: &mut Context, iterable: JsValue) -> JsResult<IteratorRecord> {
+pub fn get_iterator(iterable: &JsValue, context: &mut Context) -> JsResult<IteratorRecord> {
     let iterator_function = iterable.get_field(WellKnownSymbols::iterator(), context)?;
     if iterator_function.is_null_or_undefined() {
         return Err(context.construct_type_error("Not an iterable"));
     }
-    let iterator_object = context.call(&iterator_function, &iterable, &[])?;
+    let iterator_object = context.call(&iterator_function, iterable, &[])?;
     let next_function = iterator_object.get_field("next", context)?;
     if next_function.is_null_or_undefined() {
         return Err(context.construct_type_error("Could not find property `next`"));
@@ -158,8 +158,8 @@ impl IteratorRecord {
         let next = context.call(&self.next_function, &self.iterator_object, &[])?;
         let done = next.get_field("done", context)?.to_boolean();
 
-        let next_result = next.get_field("value", context)?;
-        Ok(IteratorResult::new(next_result, done))
+        let value = next.get_field("value", context)?;
+        Ok(IteratorResult { value, done })
     }
 
     /// Cleanup the iterator
@@ -203,20 +203,6 @@ impl IteratorRecord {
 
 #[derive(Debug)]
 pub struct IteratorResult {
-    value: JsValue,
-    done: bool,
-}
-
-impl IteratorResult {
-    fn new(value: JsValue, done: bool) -> Self {
-        Self { value, done }
-    }
-
-    pub fn is_done(&self) -> bool {
-        self.done
-    }
-
-    pub fn value(self) -> JsValue {
-        self.value
-    }
+    pub value: JsValue,
+    pub done: bool,
 }
--- a/boa/src/builtins/map/map_iterator.rs
+++ b/boa/src/builtins/map/map_iterator.rs
@@ -7,7 +7,7 @@ use crate::{
 };
 use gc::{Finalize, Trace};
 
-use super::{ordered_map::MapLock, Map};
+use super::ordered_map::MapLock;
 /// The Map Iterator object represents an iteration over a map. It implements the iterator protocol.
 ///
 /// More information:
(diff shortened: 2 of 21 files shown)


boa/src/builtins/map/mod.rs
```diff
@@ -507,28 +512,79 @@ impl Map {
...
+/// [spec]: https://tc39.es/ecma262/#sec-add-entries-from-iterable
+pub(crate) fn add_entries_from_iterable(
+    target: &JsObject,
+    iterable: &JsValue,
+    adder: &JsValue,
+    context: &mut Context,
+) -> JsResult<JsValue> {
+    // 1. If IsCallable(adder) is false, throw a TypeError exception.
+    let adder = match adder {
+        JsValue::Object(obj) if obj.is_function() => obj,
```
raskad, line 537: Same as above.
```suggestion
        JsValue::Object(obj) if obj.is_callable() => obj,
```

boa/src/builtins/map/mod.rs
```diff
@@ -419,79 +441,62 @@ impl Map {
...
+            _ => return context.throw_type_error("`this` is not a Map"),
+        };
 
-        if !callback_arg.is_function() {
-            let name = callback_arg.to_string(context)?;
-            return context.throw_type_error(format!("{} is not a function", name));
-        }
+        // 3. If IsCallable(callbackfn) is false, throw a TypeError exception.
+        let callback = match args.get_or_undefined(0) {
+            JsValue::Object(obj) if obj.is_function() => obj,
```
raskad, line 453: We should use the `is_callable` function here. At the moment it only checks if the object is a function, that may change in the future.
```suggestion
            JsValue::Object(obj) if obj.is_callable() => obj,
```

boa/src/builtins/map/mod.rs
```diff
@@ -419,79 +441,62 @@ impl Map {
...
+        // NOTE:
+        //
+        // forEach does not directly mutate the object on which it is called but
+        // the object may be mutated by the calls to callbackfn. Each entry of a
+        // map's [[MapData]] is only visited once. New keys added after the call
+        // to forEach begins are visited. A key will be revisited if it is deleted
+        // after it has been visited and then re-added before the forEach call completes.
+        // Keys that are deleted after the call to forEach begins and before being visited
+        // are not visited unless the key is added again before the forEach call completes.
+        let lock = map.borrow_mut().expect_map_mut().lock(map.clone());
```
raskad, line 473: We can prefix the `lock` variable with `_` and avoid the manual `drop` later.
```suggestion
        let _lock = map.borrow_mut().expect_map_mut().lock(map.clone());
```

boa/src/builtins/map/mod.rs
```diff
@@ -419,79 +441,62 @@ impl Map {
...
             if let Some(arguments) = arguments {
-                context.call(callback_arg, this_arg, &arguments)?;
+                // i. Perform ? Call(callbackfn, thisArg, « e.[[Value]], e.[[Key]], M »).
+                callback.call(this_arg, &arguments, context)?;
             }
 
             index += 1;
         }
 
         drop(lock);
```
raskad, line 494: See above.

```suggestion
```

raskad (changes requested):
Looks really nice.
I've got a few little things.

boa/src/builtins/map/map_iterator.rs
```diff
@@ -67,77 +72,54 @@ impl MapIterator {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-%mapiteratorprototype%.next
     pub(crate) fn next(this: &JsValue, _: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
```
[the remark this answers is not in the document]
jedel1043: I would too, but this function is a temporary hack until we implement generators, so we can't really document it.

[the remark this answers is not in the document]
raskad: Ah yes I see. Makes sense.

raskad (approved):

boa/src/object/mod.rs
```diff
@@ -603,6 +608,33 @@ impl Object {
...
     }
 
+    #[inline]
+    pub fn expect_map_mut(&mut self) -> &mut OrderedMap<JsValue> {
+        self.as_map_mut().expect("object was not a map")
+    }
+
+    #[inline]
+    pub fn is_map_iterator(&self) -> bool {
+        self.as_map_iterator_ref().is_some()
```
Razican, line 618: I think here we should use the `matches!()` macro, and that way we don't need to first extract the option and then check it's variant.

boa/src/object/mod.rs
```diff
@@ -592,6 +592,11 @@ impl Object {
         }
     }
 
+    #[inline]
+    pub fn expect_map_ref(&self) -> &OrderedMap<JsValue> {
```
Razican, line 596: I'm not a fan of this. It makes things less explicit, and only removes a few characters. It adds more functions and panics are much more difficult to debug.

At the minimum, they should have the `#[track_caller]` attribute.

boa/src/builtins/iterable/mod.rs
```diff
@@ -203,20 +203,6 @@ impl IteratorRecord {
...
-
-    pub fn is_done(&self) -> bool {
-        self.done
-    }
-
-    pub fn value(self) -> JsValue {
-        self.value
-    }
+    pub value: JsValue,
+    pub done: bool,
```
Razican, line 207: This I don't like. Is this needed? Using public members makes this structure modifiable almost anywhere and with no control, and makes it much more difficult if we ever want to change the inner structure.

The getter and setter patterns are very well established in Rust for a reason. They are less error prone and easier to maintain external compatibility.

Razican (commented):
The conformance numbers look great, and the code seems good, but I'll need some convincing in some design decisions :)

boa/src/object/mod.rs
jedel1043: Yeah, maybe that's not a great idea. Hm, I think I can modify the code to avoid expects or explicitly use `expect` where necessary.

boa/src/builtins/iterable/mod.rs
jedel1043: Using getters and setters is justifiable if we had to maintain some state internally. This struct is just a wrapper for a `(Value, bool)` to represent an iterator item, I don't see the utility on having only getters if the struct will have just `is_done` and `get_value` as methods.

raskad: While you are right that there is no functionality with this struct, I think the private fields preserve the nature of the struct. It should only be used to access the result of an iterator next call. If the fields are public, the struct is really only a wrapper for `(JsValue, bool)`. In that case we could also just work with that tuple directly instead of using a struct.

jedel1043: We could return a tuple, but using an struct makes more explicit what each type represents. That's the same reason why it's more idiomatic to represent a 2d point as a `Point {x: f64, y: f64}` instead of a `(f64, f64)`.

Razican: But, if it's not meant to be modified, we don't even need the setters, and not having them would prevent accidental mutation.

jedel1043: But it is meant to be modified. Or at least is meant to be destructured and used as an owned value, like in:

https://github.com/boa-dev/boa/blob/2f8c35db571e3a1c4dd03040d70cafcc02d69ae9/boa/src/syntax/ast/node/array/mod.rs#L54-L60

or in 

https://github.com/boa-dev/boa/blob/2f8c35db571e3a1c4dd03040d70cafcc02d69ae9/boa/src/syntax/ast/node/new/mod.rs#L61-L66

And no function takes an `IteratorResult` as a parameter, so we don't need to ensure that `IteratorResult` is not modified. If we used `IteratorResult` as a parameter on any function, I would completely agree with using getters.

raskad: Now that I think about this, is there a reason that `IteratorResult` is `pub` instead of `pub(crate)` at all? We neither return nor accept this anywhere in the public api.

RageKnify (approved):
LGTM, only exception is having the extra methods in `Object` like @Razican said. In my opinion it doesn't abstract that much code away and is one more thing people need to know about.

boa/src/builtins/iterable/mod.rs
jedel1043: > Now that I think about this, is there a reason that `IteratorResult` is `pub` instead of `pub(crate)` at all? We neither return nor accept this anywhere in the public api.

Yeah, I just copied the current visibility, but maybe we shouldn't expose it to the user.

jedel1043:
> LGTM, only exception is having the extra methods in `Object` like @Razican said. In my opinion it doesn't abstract that much code away and is one more thing people need to know about.

Ah, I forgot to push my last changes 😅

boa/src/builtins/iterable/mod.rs
Razican: The API is not very well defined yet, and we will probably need to change many visibilities. I'm OK with these changes.

Razican (approved):

PR #1584, opened 2021-09-17
Document the `Math` builtin object

jedel1043:
This Pull Request documents the `Math` builtin methods and improves some patterns with `get_or_undefined`


raskad (approved):

boa/src/builtins/math/mod.rs
```diff
@@ -415,14 +508,15 @@ impl Math {
...
-            (_, _) => 0,
-        }
-        .into())
+        // 1. Let a be ℝ(? ToUint32(x)).
+        let x = args.get_or_undefined(0).to_u32(context)?;
+
+        // 2. Let b be ℝ(? ToUint32(y)).
+        let y = args.get_or_undefined(1).to_u32(context)?;
+
+        // 3. Let product be (a × b) modulo 232.
```
RageKnify, line 517: I think you forgot the `^`?
```suggestion
        // 3. Let product be (a × b) modulo 2^32.
```

boa/src/builtins/math/mod.rs
```diff
@@ -415,14 +508,15 @@ impl Math {
...
+        // 4. If product ≥ 2^31, return 𝔽(product - 2^32); otherwise return 𝔽(product).
```
RageKnify, line 518: And here, should it not be `2^32` instead of `2^31`?
```suggestion
        // 4. If product ≥ 2^32, return 𝔽(product - 2^32); otherwise return 𝔽(product).
```

boa/src/builtins/math/mod.rs
```diff
@@ -502,11 +615,29 @@ impl Math {
     /// [spec]: https://tc39.es/ecma262/#sec-math.max
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max
     pub(crate) fn max(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        // 1. Let coerced be a new empty List.
+        // 2. For each element arg of args, do
+        // b. Append n to coerced.
+        // 3. Let highest be -∞𝔽.
         let mut max = f64::NEG_INFINITY;
```
RageKnify, line 622: Id say calling it `highest` or similar like the spec would be nicer to avoid `max.max(num)`

boa/src/builtins/math/mod.rs
```diff
@@ -502,11 +615,29 @@ impl Math {
...
         for arg in args {
+            // a. Let n be ? ToNumber(arg).
             let num = arg.to_number(context)?;
-            max = max.max(num);
+
+            max = if max.is_nan() {
+                continue;
+            } else if num.is_nan() {
+                // a. If number is NaN, return NaN.
+                f64::NAN
```
RageKnify, line 633: In these cases the loop should be short-circuiting (returning immediately) and you are not doing it, please change these both to `return Ok(f64::NAN);`

boa/src/builtins/math/mod.rs
```diff
@@ -519,11 +650,29 @@ impl Math {
...
         for arg in args {
+            // a. Let n be ? ToNumber(arg).
             let num = arg.to_number(context)?;
-            min = min.min(num);
+
+            min = if min.is_nan() {
+                continue;
+            } else if num.is_nan() {
+                // a. If number is NaN, return NaN.
+                f64::NAN
```
RageKnify, line 668: Same as above, should be short-circuiting

RageKnify (changes requested):
Just some small corrections

boa/src/builtins/math/mod.rs
raskad: I thought this too, but it is expected that `ToNumber` is called on every `arg`. There is actually a test that in the 262 suite for that behavior.

jedel1043: Yeah, I tried short circuiting but some tests failed. The reason for that is to prioritize conversion errors over a NaN result.

boa/src/builtins/math/mod.rs
jedel1043: Nope, this is to account for negative numbers, since the operation is a multiplication of signed integers

boa/src/builtins/math/mod.rs
jedel1043: Yeah, also the `min` in `Math.min`

boa/src/builtins/math/mod.rs
RageKnify: Oh, because we are doing it all in 1 loop rather than 2. I guess it's hard to know whether it is worth it though.

jedel1043: I was trying to preserve the original implementation, but I could reimplement it using a vector, and it shouldn't make too much of a difference aside from the allocation, with the advantage of being a lot more readable. What do you think?

RageKnify: I'm okay with leaving it like this, I think for most use cases this is probably the best choice.

PR #1597, opened 2021-09-29
Updated all dependencies and Test262 test suite

Razican:
This updates all the dependencies and the Test262 suite for the 0.13 release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,9 +135,9 @@ dependencies = [
 
 [[package]]
 name = "bstr"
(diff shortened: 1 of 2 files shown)


jedel1043:
I think we should just update. We can open two new issues for the failing tests and tag them as easy tasks for new contributors.

jedel1043 (commented):
I think we should just update. We can open two new issues for the failing tests and tag them as easy tasks for new contributors.

jedel1043 (approved):

PR #1610, opened 2021-10-01
VM Tidy Up

jasonwilliams:
- comment out `pushUndefined` and `Return` https://github.com/boa-dev/boa/discussions/1607#discussioncomment-1416211
- Move call frame to its own file (we may end up adding more code here later)
- heading for code blocks needed to be on its own line
- Show the difference between the VM starting up and a new Call frame being ran
- add trace to VSCode launch

### DIFF
--- a/boa/src/bytecompiler.rs
+++ b/boa/src/bytecompiler.rs
@@ -868,8 +868,9 @@ impl ByteCompiler {
 
         compiler.code_block.params = paramaters.to_owned().into_boxed_slice();
 
-        // compiler.emit(Opcode::PushUndefined, &[]);
-        // compiler.emit(Opcode::Return, &[]);
+        // TODO These are redundant if a function returns so may need to check if a function returns and adding these if it doesn't
+        compiler.emit(Opcode::PushUndefined, &[]);
+        compiler.emit(Opcode::Return, &[]);
 
         let code = Gc::new(compiler.finish());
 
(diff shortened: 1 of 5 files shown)


boa/src/bytecompiler.rs
```diff
@@ -868,8 +868,8 @@ impl ByteCompiler {
 
         compiler.code_block.params = paramaters.to_owned().into_boxed_slice();
 
-        compiler.emit(Opcode::PushUndefined, &[]);
-        compiler.emit(Opcode::Return, &[]);
+        // compiler.emit(Opcode::PushUndefined, &[]);
+        // compiler.emit(Opcode::Return, &[]);
```
HalidOdat, line 872: we can't do this without knowing if the last node is a return statement. otherwise the function will never return.

jasonwilliams: I've added it back

raskad (approved):
Looks good to me

boa/src/vm/mod.rs
```diff
@@ -521,11 +509,17 @@ impl Context {
         const OPERAND_COLUMN_WIDTH: usize = COLUMN_WIDTH;
         const NUMBER_OF_COLUMNS: usize = 4;
 
+        let msg = if !self.vm.frame().exit_on_return {
+            " Call Frame "
+        } else {
+            " VM Start"
+        };
+
```
RageKnify, line 517: Can you swap the order on this? I understand `Call Frame` is the more common one but I think having simpler logic trumps that.
```suggestion
        let msg = if self.vm.frame().exit_on_return {
            " VM Start"
        } else {
            " Call Frame "
        };

```

RageKnify (changes requested):

RageKnify (approved):

PR #1612, opened 2021-10-02
Fix broken structure links in documentation

abhishekc-sharma:
Found some broken links to some structures when perusing the documentation of Lexer and Parser.

### DIFF
--- a/boa/src/syntax/lexer/token.rs
+++ b/boa/src/syntax/lexer/token.rs
@@ -107,8 +107,6 @@ pub enum TokenKind {
     Identifier(Box<str>),
 
     /// A keyword.
-    ///
-    /// see: [`Keyword`](../keyword/enum.Keyword.html)
     Keyword(Keyword),
 
     /// A `null` literal.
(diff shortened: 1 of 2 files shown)


boa/src/syntax/lexer/token.rs
```diff
@@ -108,7 +108,7 @@ pub enum TokenKind {
 
     /// A keyword.
     ///
-    /// see: [`Keyword`](../keyword/enum.Keyword.html)
+    /// see: [`Keyword`](../../ast/keyword/enum.Keyword.html)
```
jedel1043, line 111: Can you check if using directly `Keyword` and `Punctuator` also fixes the links? Because IIRC, if the type is imported it should work.

abhishekc-sharma: Actually now that you mention it, it seems that by default rustdoc automatically links the struct inside the enum variant. The additional "see: <Link>" is not required".
Screenshot 2021-10-01 at 21-42-05 TokenKind in boa syntax lexer token - Rust

What I mean is that the green text in the above screenshot already links to the struct. Shall I remove the two explicit links that anyway seem redundant now ?

jedel1043: Yeah, you can remove them

jedel1043 (approved):
Thanks!

PR #1613, opened 2021-10-02
Implement Array.prototype.at()

### LINKED ISSUE #1596: Implement Array.prototype.at()
**ECMASCript feature**
https://tc39.es/ecma262/#sec-array.prototype.at

**Example code**
This is the same feature added for String.prototype.at() - https://github.com/boa-dev/boa/pull/1375

```javascript
const arr1 = [1,5,6,2];
arr1.at(-1) // 2
```
The expected output is `2`.

nekevss:
Hi!

This Pull Request closes #1596.

I implemented the Array.prototype.at() using the link to the ECMAScript feature provided on the issue. I also added to the method call for at() to the chain. 

ECMAScript had Array.prototype.at() just in front of the Array.prototype.concat() method, so I placed the function for at prior to that method to follow the doc order.

Just as a note: this is my first pull request. So if there is anything wrong or something I missed, feel free to let me know!


jedel1043:
Thank you for your contribution!
The code currently doesn't compile, so I'd advise you to run `cargo check` and `cargo clippy` to fix your compilation errors :)

nekevss:
Hi! I updated the branch with fixes to some errors I was able to get cargo check gave me.

This fix compiled and successfully ran a test.js file on my computer for the implementation.

jedel1043:
  | Total | 80,930 | 80,930 | 0 |
  | Passed | 33,313 | 33,335 | **+22** |
  | Ignored | 15,898 | 15,898 | 0 |
  | Failed | 31,719 | 31,697 | **-22** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 41.16% | 41.19% | **+0.03%** |
  
  <details><summary><b>Fixed tests (22):</b></summary>
  
  ```
  test/built-ins/Array/prototype/at/returns-item.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/returns-item.js (previously Failed)
  test/built-ins/Array/prototype/at/returns-undefined-for-holes-in-sparse-arrays.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/returns-undefined-for-holes-in-sparse-arrays.js (previously Failed)
  test/built-ins/Array/prototype/at/name.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/name.js (previously Failed)
  test/built-ins/Array/prototype/at/returns-undefined-for-out-of-range-index.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/returns-undefined-for-out-of-range-index.js (previously Failed)
  test/built-ins/Array/prototype/at/index-argument-tointeger.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/index-argument-tointeger.js (previously Failed)
  test/built-ins/Array/prototype/at/returns-item-relative-index.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/returns-item-relative-index.js (previously Failed)
  test/built-ins/Array/prototype/at/return-abrupt-from-this.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/return-abrupt-from-this.js (previously Failed)
  test/built-ins/Array/prototype/at/index-non-numeric-argument-tointeger-invalid.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/index-non-numeric-argument-tointeger-invalid.js (previously Failed)
  test/built-ins/Array/prototype/at/index-non-numeric-argument-tointeger.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/index-non-numeric-argument-tointeger.js (previously Failed)
  test/built-ins/Array/prototype/at/length.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/length.js (previously Failed)
  test/built-ins/Array/prototype/at/prop-desc.js [strict mode] (previously Failed)
  test/built-ins/Array/prototype/at/prop-desc.js (previously Failed)
  ```

boa/src/builtins/array/mod.rs
```diff
@@ -487,6 +488,45 @@ impl Array {
...
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-array.prototype.at
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at
+    pub(crate) fn at(this: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        //1. let O be ? ToObject(this value)
+        let obj = this.to_object(context)?;
+        //2. let len be ? LengthOfArrayLike(O)
+        let len = obj.length_of_array_like(context)? as i64;
+        //3. let relativeIndex be ? ToIntegerOrInfinity(index)
+        let relative_index = args.get(0).unwrap().to_integer_or_infinity(context)?;
```
jedel1043, line 509:
```suggestion
        let relative_index = args.get_or_undefined(0).to_integer_or_infinity(context)?;
```

This won't panic if `at` is called with no arguments, like so: `[1, 2, 3, 4].at()`

boa/src/builtins/array/mod.rs
```diff
@@ -487,6 +488,45 @@ impl Array {
...
+        let k = match relative_index {
+            //4. if relativeIndex >= 0, then let k be relativeIndex
+            IntegerOrInfinity::Integer(i) if i >= 0 => i,
+            //5. Else, let k be len + relativeIndex
+            IntegerOrInfinity::Integer(i) => len + i,
+            //handle most likely impossible case of
+            //IntegerOrInfinity::NegativeInfinity || IntegerOrInfinity::PositiveInfinity
+            //by setting k to len which will return undefined below
+            _ => len,
```
jedel1043, line 518: Instead of relying on the condition below, maybe it would be better if we early returned here:
```suggestion
            _ => return Ok(JsValue::undefined()),
```
Makes it easier to identify that only finite numbers are valid.

boa/src/builtins/array/mod.rs
```diff
@@ -487,6 +488,45 @@ impl Array {
         Ok(a.into())
     }
 
+    ///'Array.prototype.at(index)'
+    ///
+    /// When at method takes desired integer as index and returns the value at given
+    /// index. Negative integer counts backwards. If index is invalid, the at method
+    /// returns undefined.
```
jedel1043, line 495: Could you clarify this comment? You could instead take the explanation of the function on MDN and paste it here, that should be enough to explain what this does.

jedel1043 (changes requested):
Just some recommendations to improve the code. Good job!

boa/src/builtins/array/mod.rs
nekevss: I noticed this literally like 15-20 minutes ago! Completely unaware that get_or_undefined existed! Definitely making this update.

boa/src/builtins/array/mod.rs
nekevss: Definitely makes more sense to identify only finite numbers being valid

boa/src/builtins/array/mod.rs
```diff
@@ -487,6 +488,45 @@ impl Array {
...
+        let obj = this.to_object(context)?;
+        //2. let len be ? LengthOfArrayLike(O)
+        let len = obj.length_of_array_like(context)? as i64;
+        //3. let relativeIndex be ? ToIntegerOrInfinity(index)
+        let relative_index = args.get_or_undefined(0).to_integer_or_infinity(context)?;
+        let k = match relative_index {
+            //4. if relativeIndex >= 0, then let k be relativeIndex
+            IntegerOrInfinity::Integer(i) if i >= 0 => i,
+            //5. Else, let k be len + relativeIndex
+            IntegerOrInfinity::Integer(i) => len + i,
```
jedel1043, line 514: Now that I think about it, you can just use the conditions as match guards for the cases:
```suggestion
            IntegerOrInfinity::Integer(i) if i >= 0 && i < len => i,
            //5. Else, let k be len + relativeIndex
            IntegerOrInfinity::Integer(i) if i.abs() <= len => len + i,
```

That would remove the need for an if to check the validity of k.

nekevss: Whoa...that's so true...

I'll definitely add that. Match guards are so wild sometimes.

I'm going to try to add comments to note exactly the checks in case anybody else needs to read them. Hopefully, they are make sense.

nekevss:
Added all the feedback above. 

Let me know if there's any more changes! The match guard one was kind of cool not gonna lie.

Thanks for all the help and feedback! 😄

boa/src/builtins/array/mod.rs
nekevss: Took the MDN description. I thought that one was the best too, but I was a bit hesitant to just grab it.

boa/src/builtins/array/mod.rs
nekevss: Actually I think there is a small potential bug/logic gap in the above match guards.

If the input argument is len, then it would miss the first arm but trigger the second (returning essentially len + len). It looks like obj.get(k) does cover for this gap by returning undefined, but I'm making a small change to the match guard on the second arm to explicitly check that i != len.

jedel1043: Ah missed that. You can, then, check for `i < 0 && i.abs() <= len`, which describes better the condition than `i != len && i.abs() <= len`

boa/src/builtins/array/mod.rs
```diff
@@ -487,6 +488,43 @@ impl Array {
...
+        let len = obj.length_of_array_like(context)? as i64;
+        //3. let relativeIndex be ? ToIntegerOrInfinity(index)
+        let relative_index = args.get_or_undefined(0).to_integer_or_infinity(context)?;
+        let k = match relative_index {
+            //4. if relativeIndex >= 0, then let k be relativeIndex
+            //check if positive and if below length of array
+            IntegerOrInfinity::Integer(i) if i >= 0 && i < len => i,
+            //5. Else, let k be len + relativeIndex
+            //integer should be negative, so abs() and check if less than or equal to length of array
+            IntegerOrInfinity::Integer(i) if i.abs() <= len && i != len => len + i,
```
jedel1043, line 516:
```suggestion
            IntegerOrInfinity::Integer(i) if i < 0 && i.abs() <= len => len + i,
```

I think this describes the check a lot better.

jedel1043 (approved):
Looks good to me! Just a small change that you can ignore if you consider it less readable :)

jedel1043:
We usually wait for two approved reviews until a PR is merged into master, so you'll just have to wait a bit of time and your contribution should be merged soon 😄

boa/src/builtins/array/mod.rs
nekevss: Totally agree! pushed the change to the second arm of the match. Again, thanks for all the feedback!

HalidOdat (approved):

PR #1616, opened 2021-10-02
Add URI Handling Functions

### LINKED ISSUE #894: Implement URI Handling Functions
**ECMASCript feature**
We currently lack the implementation of some built-in global functions used for URI encoding/decoding. You can find these functions in the spec [here](https://tc39.es/ecma262/#sec-uri-handling-functions).

The 4 functions to implement would be these:
 - [`encodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI)
 - [`encodeURIComponent()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
 - [`decodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI)
 - [`decodeURIComponent()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent)

In those links you can find example code that should work after this implementation.

**Implementation tips**
We should probably create a new module [here](https://github.com/boa-dev/boa/tree/master/boa/src/builtins) for these functions. An example of a built-in global function implementation can be seen [here](https://github.com/boa-dev/boa/blob/master/boa/src/builtins/number/mod.rs#L83).

jtara1:
This Pull Request fixes/closes #894.

It changes the following:

- implements the builtin functions for encoding or decoding URIs

### DIFF
--- a/boa/src/builtins/uri/mod.rs
+++ b/boa/src/builtins/uri/mod.rs
@@ -1,12 +1,74 @@
 use crate::builtins::BuiltIn;
 use crate::property::Attribute;
-use crate::{Context, JsValue, JsResult};
+use crate::{Context, JsValue, JsResult, JsString};
 use crate::object::function::make_builtin_fn;
 
 use percent_encoding::{utf8_percent_encode, AsciiSet, CONTROLS};
+use crate::builtins::string::code_point_at;
 
 /// https://url.spec.whatwg.org/#fragment-percent-encode-set
 const FRAGMENT: &AsciiSet = &CONTROLS.add(b' ').add(b'"').add(b'<').add(b'>').add(b'`');
+
+fn is_uri_mark(c: &str) -> bool {
+    return match c {
+        "-" => true,
+        "_" => true,
+        "." => true,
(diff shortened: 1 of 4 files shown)


boa/src/builtins/uri/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+        r = &remaining % 16;
+    }
+
+    return to_char(r as usize) + &hex_number[..];
+}
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct Uri;
+
+impl BuiltIn for Uri {
```
jedel1043, line 76: Implementing `BuiltIn` for `Uri` would create a new property with name `Uri` on the global object, making it not spec compliant.
My recommendation would be to skip the `BuiltIn` implementation for `Uri` and just create a new associated function `init`, then call it from the initialization function of the context without adding it to the array of global inits.

PR #1618, opened 2021-10-03
Implement `Symbol.prototype.valueOf`

### LINKED ISSUE #1617: Symbol.prototype.valueOf.call(x) gives different results than x.valueOf()
**Describe the bug**

`Symbol` gives different results depending on how `valueOf` is called, which causes various tests to fail, like [this one](https://github.com/tc39/test262/blob/10ad4c159328d12ad8d12b95a8f8aefa0e4fd49b/test/built-ins/Symbol/prototype/valueOf/this-val-symbol.js) for example.

Here's an example session from boa-cli:
```
>> a = Symbol('s')
Symbol(s)
>> Symbol.prototype.valueOf.call(a) == a.valueOf()
false
>> Symbol.prototype.valueOf.call(a) == a
false
>> a.valueOf() == a
false
>>
```

**To Reproduce**
See above. `boa-cli` was run with `cd boa_cli; cargo run`.

**Expected behavior**

The `valueOf` a `Symbol` should always `==` the Symbol, no matter how `valueOf` is called. Here's an example from Node.js:

```
> a = Symbol('s')
Symbol(s)
> a.valueOf() == a
true
```

This behavior should be made the same in Boa.

- OS: Artix Linux
- Kernel Version: 5.13.12-artix1-1
- Target triple: e.g. x86_64-unknown-linux-gnu
- Rustc version: rustc 1.55.0 (c8dfcfe04 2021-09-06)

**Additional context**

hle0:
Signed-off-by: hle0 <91701075+hle0@users.noreply.github.com>

This Pull Request fixes/closes #1617 .

It changes the following:

- `valueOf` is defined differently for `Symbol` so that `Symbol('s').valueOf() == Symbol('s')`, as per the spec.

### DIFF
--- a/boa/src/builtins/symbol/mod.rs
+++ b/boa/src/builtins/symbol/mod.rs
@@ -228,7 +228,7 @@ impl Symbol {
         _: &[JsValue],
         context: &mut Context,
     ) -> JsResult<JsValue> {
-        // valueOf a symbol is itself
+        // 1. Return ? thisSymbolValue(this value).
         let symbol = Self::this_symbol_value(this, context)?;
         Ok(JsValue::Symbol(symbol))
     }
--- a/boa/src/builtins/symbol/mod.rs
+++ b/boa/src/builtins/symbol/mod.rs
@@ -127,6 +127,7 @@ impl BuiltIn for Symbol {
         .static_property("toStringTag", symbol_to_string_tag.clone(), attribute)
         .static_property("unscopables", symbol_unscopables, attribute)
         .method(Self::to_string, "toString", 0)
+        .method(Self::value_of, "valueOf", 0)
         .accessor(
             "description",
             Some(get_description),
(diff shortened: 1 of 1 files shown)


boa/src/builtins/symbol/mod.rs
```diff
@@ -212,6 +213,26 @@ impl Symbol {
...
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/valueOf
+    /// [spec]: https://tc39.es/ecma262/#sec-symbol.prototype.valueof
+    pub(crate) fn value_of(
+        this: &JsValue,
+        _: &[JsValue],
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        // valueOf a symbol is itself
+        let symbol = Self::this_symbol_value(this, context)?;
+        Ok(JsValue::Symbol(symbol))
```
jedel1043, line 233: Can you add a comment with the spec steps? Even if it's only a single one :)

jedel1043 (changes requested):
Just a small detail. Everything else looks good!

jedel1043 (approved):
Good work! We require two approved reviews to merge into master, so in a bit of time this should get merged :)

HalidOdat (approved):

PR #1620, opened 2021-10-03
Fix `Number.toExponential` and `Number.toFixed`

### LINKED ISSUE #1602: Fix `Number.toExponential` and `Number.toFixed`
The functions `toExponential` and `toFixed` of the `Number` prototype are currently not implemented spec compliant. All of the tests for those two functions should pass if they do not rely on other features that are not implemented.

To see the exact failed tests, take a look at the current results at https://boa-dev.github.io/boa/test262/

Spec references:
- https://tc39.es/ecma262/#sec-number.prototype.toexponential
- https://tc39.es/ecma262/#sec-number.prototype.tofixed

nrabulinski:
This Pull Request closes #1602.

It makes `toExponential` and `toFixed` methods on `Number` compliant with the specification.

One of the tests in `toExponential` currently fails. This is something that's blocked by https://github.com/rust-lang/rust/issues/89493, unless we were to implement the exponential string algorithm ourselves.

### DIFF
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -1164,7 +1164,7 @@ fn f64_to_exponential(n: f64) -> String {
 }
 
 /// Helper function that formats a float as a ES6-style exponential number string with a given precision.
-// We can't use the same approach as in `num_to_exponential`
+// We can't use the same approach as in `f64_to_exponential`
 // because in cases like (0.999).toExponential(0) the result will be 1e0.
 // Instead we get the index of 'e', and if the next character is not '-' we insert the plus sign
 fn f64_to_exponential_with_precision(n: f64, prec: usize) -> String {
--- a/boa/src/builtins/number/mod.rs
+++ b/boa/src/builtins/number/mod.rs
@@ -214,6 +214,21 @@ impl Number {
         }
     }
 
+    /// Helper function that formats a float as a ES6-style exponential number string with a given precision.
+    // We can't use the same approach as in `num_to_exponential`
+    // because in cases like (0.999).toExponential(0) the result will be 1e0.
+    // Instead we get the index of 'e', and if the next character is not '-' we insert the plus sign
+    fn num_to_exponential_with_precision(n: f64, prec: usize) -> String {
+        let mut res = format!("{:.*e}", prec, n);
+        let idx = res
+            .find('e')
+            .unwrap_or_else(|| unreachable!("'e' not found in exponential string"));
+        if res.as_bytes()[idx + 1] != b'-' {
+            res.insert(idx + 1, '+');
+        }
+        res
+    }
+
     /// `Number.prototype.toExponential( [fractionDigits] )`
     ///
     /// The `toExponential()` method returns a string representing the Number object in exponential notation.
@@ -227,11 +242,30 @@ impl Number {
     #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_exponential(
         this: &JsValue,
-        _: &[JsValue],
+        args: &[JsValue],
         context: &mut Context,
     ) -> JsResult<JsValue> {
         let this_num = Self::this_number_value(this, context)?;
-        let this_str_num = Self::num_to_exponential(this_num);
+        let precision = match args.get(0) {
+            Some(arg) if arg.is_undefined() => None,
+            Some(n) => Some(n.to_integer(context)? as i32),
+            None => None,
+        };
+        if !this_num.is_finite() {
+            return Ok(JsValue::new(Self::to_native_string(this_num)));
+        }
+        // Get rid of the '-' sign for -0.0
+        let this_num = if this_num == 0. { 0. } else { this_num };
+        let this_str_num = if let Some(precision) = precision {
+            if precision < 0 || precision > 100 {
+                return Err(context.construct_range_error(
+                    "toExponential() argument must be between 0 and 100",
+                ));
+            }
+            Self::num_to_exponential_with_precision(this_num, precision as usize)
+        } else {
+            Self::num_to_exponential(this_num)
+        };
         Ok(JsValue::new(this_str_num))
     }
 
@@ -254,13 +288,23 @@ impl Number {
         let this_num = Self::this_number_value(this, context)?;
         let precision = match args.get(0) {
             Some(n) => match n.to_integer(context)? as i32 {
-                x if x > 0 => n.to_integer(context)? as usize,
-                _ => 0,
+                0..=100 => n.to_integer(context)? as usize,
+                _ => {
+                    return Err(context.construct_range_error(
+                        "toFixed() digits argument must be between 0 and 100",
+                    ))
+                }
             },
             None => 0,
         };
-        let this_fixed_num = format!("{:.*}", precision, this_num);
-        Ok(JsValue::new(this_fixed_num))
+        // Get rid of the '-' sign for -0.0
+        let this_num = if this_num == 0. { 0. } else { this_num };
+        if this_num >= 1.0e21 {
+            Ok(JsValue::new(Self::num_to_exponential(this_num)))
+        } else {
+            let this_fixed_num = format!("{:.*}", precision, this_num);
+            Ok(JsValue::new(this_fixed_num))
+        }
     }
 
     /// `Number.prototype.toLocaleString( [locales [, options]] )`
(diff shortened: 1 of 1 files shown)


raskad:
  | Total | 80,930 | 80,930 | 0 |
  | Passed | 33,364 | 33,388 | **+24** |
  | Ignored | 15,898 | 15,898 | 0 |
  | Failed | 31,668 | 31,644 | **-24** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 41.23% | 41.26% | **+0.03%** |
  
  <details><summary><b>Fixed tests (24):</b></summary>
  
  ```
  test/built-ins/Number/prototype/toFixed/S15.7.4.5_A1.4_T01.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toFixed/S15.7.4.5_A1.4_T01.js (previously Failed)
  test/built-ins/Number/prototype/toFixed/range.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toFixed/range.js (previously Failed)
  test/built-ins/Number/prototype/toFixed/S15.7.4.5_A1.3_T01.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toFixed/S15.7.4.5_A1.3_T01.js (previously Failed)
  test/built-ins/Number/prototype/toFixed/S15.7.4.5_A1.3_T02.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toFixed/S15.7.4.5_A1.3_T02.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/return-abrupt-tointeger-fractiondigits-symbol.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/return-abrupt-tointeger-fractiondigits-symbol.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/this-is-0-fractiondigits-is-0.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/this-is-0-fractiondigits-is-0.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/range.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/range.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/this-is-0-fractiondigits-is-not-0.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/this-is-0-fractiondigits-is-not-0.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/return-abrupt-tointeger-fractiondigits.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/return-abrupt-tointeger-fractiondigits.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/infinity.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/infinity.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/undefined-fractiondigits.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/undefined-fractiondigits.js (previously Failed)
  test/built-ins/Number/prototype/toExponential/tointeger-fractiondigits.js [strict mode] (previously Failed)
  test/built-ins/Number/prototype/toExponential/tointeger-fractiondigits.js (previously Failed)
  ```
  </details>

boa/src/builtins/number/mod.rs
```diff
@@ -254,13 +287,23 @@ impl Number {
...
+                    return Err(context.construct_range_error(
+                        "toFixed() digits argument must be between 0 and 100",
+                    ))
+                }
             },
             None => 0,
         };
-        let this_fixed_num = format!("{:.*}", precision, this_num);
-        Ok(JsValue::new(this_fixed_num))
+        // Get rid of the '-' sign for -0.0
```
raskad, line 299: Step 6 from the spec seems to be missing here: `If x is not finite, return ! Number::toString(x).`
If you run `Infinity.toFixed()` the return is `"inf"` while it should be `"Infinity"`. I think there should be a 262 test for that, maybe we can get one added.

I would also suggest to add the spec as comments like in `parse_int`. That makes it very easy to read code and spec at the same time and see where and how they differ.

raskad (changes requested):

raskad:
Regarding the one failing test; I think we should merge this without that one.

But we may have an easier solution that to fix the issues in `std`. We already use [`ryu-js`](https://crates.io/crates/ryu-js) crate that converts numbers to strings. Conveniently the repository for the crate is in the boa-dev org: https://github.com/boa-dev/ryu-js ;)

Without looking further into it, my idea would be to implement the functionality of `num_to_exponential` and `num_to_exponential_with_precision` there. @HalidOdat I think you probably know the most about this, do you think we should try to do that, or is `ryu-js` not the right place?

boa/src/builtins/number/mod.rs
```diff
@@ -214,6 +214,21 @@ impl Number {
...
 
+    /// Helper function that formats a float as a ES6-style exponential number string with a given precision.
+    // We can't use the same approach as in `num_to_exponential`
+    // because in cases like (0.999).toExponential(0) the result will be 1e0.
+    // Instead we get the index of 'e', and if the next character is not '-' we insert the plus sign
+    fn num_to_exponential_with_precision(n: f64, prec: usize) -> String {
+        let mut res = format!("{:.*e}", prec, n);
+        let idx = res
+            .find('e')
+            .unwrap_or_else(|| unreachable!("'e' not found in exponential string"));
```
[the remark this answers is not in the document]
nrabulinski: My motivation for doing that was to eventually change it to `std::hint::unreachable_unchecked` as the resulting string **has** to contain "e" which is at most second to last, same with switching `res.as_bytes()[idx + 1]` to a `get_unchecked`. I understand and respect if that's not the direction you want to take though.

boa/src/builtins/number/mod.rs
nrabulinski: Ah, you're right, I actually planned on implementing that since there is a test for it in `toExponential` implementation and I saw it in the spec. I must've forgotten, I'm sorry.

boa/src/builtins/number/mod.rs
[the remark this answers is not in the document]
raskad: I see what you mean. I don't think we should use any non-required unsafe right now, but when our focus moves from spec compliance to performance that pattern for sure seems to be nice.

raskad (approved):
Looks good to me!

boa/src/builtins/number/mod.rs
```diff
@@ -214,6 +214,19 @@ impl Number {
         }
     }
 
+    /// Helper function that formats a float as a ES6-style exponential number string with a given precision.
+    // We can't use the same approach as in `num_to_exponential`
+    // because in cases like (0.999).toExponential(0) the result will be 1e0.
+    // Instead we get the index of 'e', and if the next character is not '-' we insert the plus sign
+    fn num_to_exponential_with_precision(n: f64, prec: usize) -> String {
```
jedel1043, line 221:
```suggestion
    fn to_exponential_with_precision(n: f64, prec: usize) -> String {
```

I'd rename `num_to_exponential` to `f64_to_exponential` and `num_to_exponential_with_precision` to `f64_to_exponential_with_precision`, and extract them as standalone functions. They don't need to be associated functions if they're just used within this module and don't receive a `&self`.

jedel1043 (changes requested):
Just a small nitpick I have about the utility functions

boa/src/builtins/number/mod.rs
nrabulinski: Sure, I agree, I just followed what was already in the codebase 😅

nrabulinski: Actually, what about other helper functions that are in the `impl Number` block? Is there a threshold when a function should be extracted and when it should stay? And would you recommend extracting it out in the same file?

jedel1043: If the function is only used internally and it doesn't need a `self` parameter, just put it as a standalone function; you can follow that rule.

We're still cleaning up some parts of the codebase, that's why there's not too much consistency about this 😅

jedel1043: Yeah, helper functions can be on the same file

nrabulinski:
Actually I just realized that there's another case which isn't covered by any 262 test and that's `(1.0).toExponential()`, which should return `1e+0` but returned `1e0` for us :smile:

jedel1043 (approved):
Great work!

jedel1043:
> Actually I just realized that there's another case which isn't covered by any 262 test and that's `(1.0).toExponential()`, which should return `1e+0` but returned `1e0` for us 😄

Oh, that can be filed on the [test262 repo](https://github.com/tc39/test262)

Issue #1609, opened 2021-10-01
Document Code Block


jasonwilliams:
There's no documentation for [code block](https://github.com/boa-dev/boa/blob/master/boa/src/vm/code_block.rs), we should add some

Issue #1608, opened 2021-10-01
Document Bytecompiler


jasonwilliams:
There's no documentation for the [bytecompiler](https://github.com/boa-dev/boa/tree/82c578b434e99a73ae95e28c968990ce20e64843/core/engine/src/bytecompiler) we should add some

postmeback:
The linked page is no longer existing. Is it good to close?

jedel1043:
That's just because we changed the location of the bytecompiler. This is still unresolved.

postmeback:
Can you paste the updated link? I want to document it.

jedel1043:
I've updated the link from the issue :)

jkrishnad:
Hey @jedel1043 is this issue still unresolved? and the page linked in here is no longer exist

jedel1043:
@JkrishnaD Yep still unresolved. I changed the link to a permalink. Do you wanna pick up this?

jkrishnad:
@jedel1043 i'll try this, so is there any reference for the docs like how it should look like.

jedel1043:
@JkrishnaD That's the fun part, we don't know 🤣

This is kind of a big task because you would have to deeply analyze the code in there to come up with useful comments explaining what each part of the bytecompiler does. 
My suggestion is to first open a PR with some module documentation on the root module file explaining the overall architecture of the VM. After that, we can discuss how to structure the rest of the documentation.

jkrishnad:
Now this is interesting 😅. So, I’ll start by adding module-level documentation explaining the VM architecture and bytecompiler flow.

jkrishnad:
Hey @jedel1043, i have gone through the bytecompiler and i observed that most of the functions, strcuts and few implementations has already have the right documentation and i think line to line documentation doesn't make any sense.

Now I have planned to add some missing documentation for the functions and impl's and also thought to add a md file in docs folder to explain the architecture and all if need.

what do you think about this?

PR #1621, opened 2021-10-03
Initial implementation for increment/decrement in VM

### LINKED ISSUE #1611: Implement increment/decrement in VM
Location of unimplemented feature:
https://github.com/boa-dev/boa/blob/master/boa/src/bytecompiler.rs#L369-L372

This would involve creating 2 new opcodes (Inc and Dec) [here](https://github.com/boa-dev/boa/blob/master/boa/src/vm/opcode.rs#L6-L534) see https://searchfox.org/mozilla-central/source/js/src/vm/Opcodes.h#649-650 for reference.

Once added the VM would need to act on it. You can copy over some of the [unary actions](https://github.com/boa-dev/boa/blob/master/boa/src/syntax/ast/node/operator/unary_op/mod.rs#L53-L127) and convert them into VM instructions like [EQ](https://github.com/boa-dev/boa/blob/master/boa/src/vm/mod.rs#L188-L193) for example

code to test:
```js
const a =  1;
a++
```

If using VSCode (Launch VM) or run with `---features vm` flag

abhishekc-sharma:
This Pull Request fixes/closes #1611

It changes the following:
- Added two new opcodes Inc and Dec to the VM for increment and decrement.
- Implementated compilation for Post/Pre Increment and Decrement operators to VM operations.
- Implemented running the Inc and Dec opcodes on the VM.

I tested my implementation on some relatively complicated expressions and I **did** notice an issue. For example the following works correctly - 

```
let a = 1, b = 2;

let foo = ++a + ++a+a++;
let bar = a++ + ++b;

console.log(foo, bar);
```

```
$ cargo run --features vm -- test.js
     Running `target/debug/boa test.js`
8 7
undefined

$ node test.js
8 7

```

But the following does not - 
```
let a = 1, b = 2;

console.log(++a + ++a+a++, a++ + ++b);
```

```
$ cargo run --features vm -- test.js
     Running `target/debug/boa test.js`
11 4
undefined

$ node test.js
8 7
```

I may be mistaken but this seems to be primarily because we evaluate arguments to function calls in reverse order and push them to the stack when making function calls - https://github.com/boa-dev/boa/blob/94992a4a8c5294df05209f71ff5758df1cb34ae3/boa/src/bytecompiler.rs#L588 . This is incorrect if the arguments' evaluation has side effects like in the case of post/pre increment/decrement.


jasonwilliams:
@abhishekc-sharma  looks good, could you explain the use of Access in your bytecode generation im not familiar with it or understand why it was used here. (im guessing because `setName` would consume the top value from the stack leaving nothing left. I wonder if Access and its methods could be documented? (sorry I know this is not to do with your PR)

> I may be mistaken but this seems to be primarily because we evaluate arguments to function calls in reverse order and push them to the stack when making function calls -

Very good spot, im not sure why they're processed in reverse here, maybe @HalidOdat knows

abhishekc-sharma:
I agree @jasonwilliams , there's very little documentation currently. These should probably be addressed as part of #1608 and #1609.

To be honest I based my implementation largely on the existing binary operators that also have a side effect similar to the increment/decrement operators like the assignment shorthand operations - https://github.com/boa-dev/boa/blob/94992a4a8c5294df05209f71ff5758df1cb34ae3/boa/src/bytecompiler.rs#L470 .

My understanding from my quick survey through the code is that the Access enum represents any kind of variable, object access. The `compile_access` method converts an AST Node representing a variable, object / attribute access to an Access instance.
And `access_set` compiles this Access to VM operations that assign to this entity the value at the top of the stack. 

Also I wasn't sure if I was supposed to add any kind of tests for this new functionality and where am I supposed to add them.

raskad (approved):
The implementation looks good to me.

abhishekc-sharma:
Done @jasonwilliams

jasonwilliams:
I think this is good to go @abhishekc-sharma in regards to tests that’s a good question, we’re currently not testing the VM path so I think we’ll need to decide how to deal with that outside of this. Maybe a new discussion.

PR #1635, opened 2021-10-05
Updated test suite and dependencies

Razican:
This PR updates the test suite with the new "Temporal" tests, along with the new regeneration of some tests. It also brings some smaller tests.

I also took the opportunity to update the dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -953,9 +953,9 @@ dependencies = [
 
 [[package]]
 name = "quote"
-version = "1.0.9"
+version = "1.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


raskad:
We probably should add `Temporal` as an ignored feature until we have started implemented it right?

jedel1043:
> We probably should add `Temporal` as an ignored feature until we have started implemented it right?

Yeah, and also we should open a new issue for that.

Razican:
> > We probably should add `Temporal` as an ignored feature until we have started implemented it right?
> 
> Yeah, and also we should open a new issue for that.

Not sure about this, It's in "experimental" stage. If someone wants to implement it, I'm fine with it, but I think I wouldn't create an issue with it. We did the same with `String.at()`: https://tc39.es/proposal-temporal/docs/

raskad (approved):

jedel1043 (approved):

PR #1640, opened 2021-10-06
Change type of prototypes of objects to `Option<JsObject>`

jedel1043:
The objective of this PR is to try to make use of the [null pointer optimization](https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#discriminant-elision-on-option-like-enums) of the [`Option`](https://doc.rust-lang.org/std/option/#representation) type, theoretically optimizing the access of the prototype of an object.

It changes the following:

- Changes the type of prototype of objects from `JsValue` to `JsObject`
- Creates a new type alias `ObjectPrototype` to avoid having to read `Option<Option<JsObject>>` in some places.  (name subject to change, maybe? I was thinking of using plain `Prototype` but there's already a `PROTOTYPE` constant, and `ObjectPrototype` doesn't sound as good because it could be confused with the intrinsic prototypes of constructor objects. I'd like to hear your suggestions on this)

### DIFF
--- a/boa/src/builtins/array_buffer/mod.rs
+++ b/boa/src/builtins/array_buffer/mod.rs
@@ -307,7 +307,7 @@ impl ArrayBuffer {
             context,
         )?;
         let obj = context.construct_object();
-        obj.set_prototype_instance(prototype.into());
+        obj.set_prototype(prototype.into());
 
         // 2. Let block be ? CreateByteDataBlock(byteLength).
         // TODO: for now just a arbitrary limit to not OOM.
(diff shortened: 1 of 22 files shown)


RageKnify (approved):
It's always nice to use the type-system to ensure invariants, should result in faster and more memory efficient code.

raskad (changes requested):
Looks good to me. Really nice idea.

jedel1043:
> Concerning the name: I think we should prefix it with `Js`. So `JsObjectPrototype` or just `JsPrototype`?

I like JsPrototype, it's clear and it doesn't overlap with the `PROTOTYPE` constant :)

raskad (approved):

PR #1653, opened 2021-10-09
Optimize the parser to use `const` booleans

jedel1043:
Testing if redesigning the parser to use `const` improves the performance.

### DIFF
--- a/boa/src/builtins/typed_array/mod.rs
+++ b/boa/src/builtins/typed_array/mod.rs
@@ -2942,10 +2942,10 @@ impl TypedArray {
         let len = values.len();
         {
             let mut o = o.borrow_mut();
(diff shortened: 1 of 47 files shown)


jedel1043:
No visible improvements on the parsing side, so closing.

Razican:
The reason this was originaly done with the `AllowYield` and so on is double:
 - Improve the performance by using generics and not depending on runtime booleans, so we have one specialized function for each parser situation
 - Make it impossible at the type level to mistakenly send an `AllowYield` to an `AllowIn`, for example, even if they are all booleans inside.

I would be very surprised if using const bools would improve performance, since this is already generating the most specific functions, as optimized as possible. It could maybe reduce cache misses, though.

PR #1668, opened 2021-10-14
Implement object literals for vm

raskad:
It changes the following:

- Implement object literals for vm
- Adds `Opcode`s for setting object getter and setter properties


jasonwilliams:
Looks good so far, I think it will need a doc comment at the top to give some explanation about this module

Issue #1558, opened 2021-09-06
Implement Async Generator Parsing


raskad:
Implement parsing for [Async Generator Function Definitions](https://tc39.es/ecma262/#sec-async-generator-function-definitions).

The following syntax should be parsed into AST nodes:
- [ ] [AsyncGeneratorMethod](https://tc39.es/ecma262/#prod-AsyncGeneratorMethod)
- [ ] [AsyncGeneratorDeclaration](https://tc39.es/ecma262/#prod-AsyncGeneratorDeclaration)
- [ ] [AsyncGeneratorExpression](https://tc39.es/ecma262/#prod-AsyncGeneratorExpression)

jedel1043:
I'm gonna add it to the async project because this needs an implementation of `Futures`.

nekevss:
Is anybody working on this currently by chance? If not, I'd like to take a try at it. Unless there are other plans around it currently.

jedel1043:
The implementation should be straightforward considering we have already implemented async and generators, so you'll just need to mirror that, following the spec to adjust it to async generators.

nekevss:
@raskad and @jedel1043 so I have a bit of a question surrounding async generators and generators and the allow_yield/allow_await bools.

Note 1 in the ECMAScript specs states "YieldExpression and AwaitExpression cannot be used within the FormalParameters of an async generator function because any expressions that are part of FormalParameters are evaluated before the resulting AsyncGenerator is in a resumable state." This note exists for the generator too, but only applies to the YieldExpression. Currently in the generator declaration implementation, we have parameters_allow_yield designated as true. Is this accurate and the value for allow_await/allow_yield should both be set to true for async generators?

To be honest, I'm not entirely certain these flags actually link to YieldExpression and AwaitExpression, but I can't entirely find where we might be checking that so I'm guessing based off the names.

jedel1043:
@nekevss Yeah, pretty much. I don't know why they're called `allow_yield/await` when they do exactly the opposite of that though, but they need to be `true` for async generators to deny the `yield` and the `await` keywords in `FormalParameters` and `FunctionBody`.

PR #1669, opened 2021-10-15
Implement Async Generator Parsing

nekevss:
This Pull Request addresses #1558.

There were a lot of files and different small changes here and there with some larger chunks as well. Let me know what you think. While I was in implementing the AsyncGeneratorMethod on the object initializer. I also added the AsyncMethod as well...hopefully that's not stepping on anyone's toes, so to speak. There were a lot of different changes across quite a few files, so please let me know if I missed anything.

It changes the following:

- Implement AsyncGeneratorExpression Parsing
- Implement AsyncGeneratorDeclaration Parsing
- Implement AsyncGeneratorMethod on Object Initializing Parsing
- Implement AsyncMethod on Object Initializing Parsing

### DIFF
--- a/boa/src/syntax/parser/statement/declaration/hoistable/async_generator_decl/mod.rs
+++ b/boa/src/syntax/parser/statement/declaration/hoistable/async_generator_decl/mod.rs
@@ -45,34 +45,42 @@ impl AsyncGeneratorDeclaration {
 }
 
 impl CallableDeclaration for AsyncGeneratorDeclaration {
+    #[inline]
     fn error_context(&self) -> &'static str {
         "async generator declaration"
     }
 
+    #[inline]
     fn is_default(&self) -> bool {
         self.is_default.0
     }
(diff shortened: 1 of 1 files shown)


nekevss:
Made the changes. I did try to rebase, but I'm not overly familiar with rebasing so hopefully that went through okay. From what I can tell it did, and cargo check and the few tests I had set up still ran fine.

raskad:
@nekevss All the failing tests that have `redeclaration` in their name can be ignored. They should be fixed by #1391.

nekevss:
I was able to make a change that fixed the tests for test/language/expressions/object/method-definition/early-errors-object-method-async-lineterminator.js. It passed when I ran the tests on my computer. I'm not entirely sure how the rest of the tests are breaking off adding in the async method parsing and async gen parsing.

I can definitely look into it further, but I might need some direction on where to look.

PR #1680, opened 2021-10-21
change that verbosity comparison to > 2

### LINKED ISSUE #1643: Add a new verbosity level to `boa_tester` to reduce the output in `-vv` mode
Currently we have 3 vervosity modes in the `boa_tester`:

 - `0`: when no `-v` flag is provided, we just show one colored dot per test, indicating if it passed or not, and the final results
 - `1`: when `-v` is provided, we show information on each test suite that ran, along with some extra information messages
 - `2`: when `-vv` is provided, it will give information on each test being run, along with information on the output in case it failed.

The issue is that there is no way of knowing the failing tests without getting the full output of the test, which can make the output difficult to read if you are only trying to find the failing test.

The idea behind this enhancement proposal is to add a new verbosity level, `-vvv`, that would include the output of failing tests that is now shown in `-vv`.

This is controlled in [here](https://github.com/boa-dev/boa/blob/main/boa_tester/src/exec/mod.rs#L255-L259). The idea is to change that verbosity comparison to `> 2`, so that it will only be shown when `-vvv` (or more) is provided.

praveenbakkal:
change that verbosity comparison to > 2 in boa_tester

This Pull Request fixes/closes 
#1643

It changes the following:

- change that verbosity comparison to > 2 in boa_tester

### DIFF
--- a/boa_tester/src/exec/mod.rs
+++ b/boa_tester/src/exec/mod.rs
@@ -252,7 +252,7 @@ impl Test {
             (TestOutcomeResult::Ignored, String::new())
         };
 
-        if verbose > 1 {
+        if verbose > 2 {
             println!("Result text:");
             println!("{}", result_text);
             println!();


praveenbakkal:
@Razican please review this PR. Let me know if additional changes are required. I went through the code and didn't see anywhere else we need to make this change. Please let me know if my understanding is not right.

PR #1685, opened 2021-10-24
[Merged by Bors] - Adding VM conformance output to PR checks

Razican:
This PR adds conformance results for the VM branch both for PRs and for the conformance results in GitHub pages (even if these are not currently being shown). I'm not 100% sure how this will really work, as I'm not very used to the syntax to concatenate strings.


Razican:
bors retry

PR #1691, opened 2021-10-27
Document CodeBlock

### LINKED ISSUE #1609: Document Code Block

TheDoctor314:
This Pull Request fixes/closes #1609.

It changes the following:

- Adds CodeBlock docs

### DIFF
--- a/boa/src/vm/code_block.rs
+++ b/boa/src/vm/code_block.rs
@@ -47,7 +47,7 @@ pub struct CodeBlock {
     /// Name of this function
     pub(crate) name: JsString,
 
-    /// The number of parameters expected.
+    /// The number of arguments expected.
     pub(crate) length: u32,
 
     /// Is this function in strict mode.
--- a/boa/src/vm/code_block.rs
+++ b/boa/src/vm/code_block.rs
@@ -34,6 +34,7 @@ unsafe impl Readable for i64 {}
 unsafe impl Readable for f32 {}
 unsafe impl Readable for f64 {}
 
+/// The internal representation of a JavaScript function.
 #[derive(Debug, Trace, Finalize)]
 pub struct CodeBlock {
     /// Name of this function
--- a/boa/src/vm/code_block.rs
+++ b/boa/src/vm/code_block.rs
@@ -20,7 +20,7 @@ use std::{convert::TryInto, fmt::Write, mem::size_of};
 
 use super::CallFrame;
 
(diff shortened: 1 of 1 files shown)


boa/src/vm/code_block.rs
```diff
@@ -34,12 +34,13 @@ unsafe impl Readable for i64 {}
...
 unsafe impl Readable for f64 {}
 
+/// The internal representation of a JavaScript function.
 #[derive(Debug, Trace, Finalize)]
 pub struct CodeBlock {
     /// Name of this function
     pub(crate) name: JsString,
 
-    // The length of this function.
+    /// The length of this function.
```
jasonwilliams, line 43: Is this the length of arguments this function takes?

jasonwilliams:
Maybe something like [this](https://github.com/boa-dev/boa/blob/main/boa/src/vm/call_frame.rs#L1-L2) at the top which says.

```rust
//! CodeBlock
//! This module is for the CodeBlock which implements a function representation in the VM

**Review by jasonwilliams — CHANGES_REQUESTED** [review]
above

**Comment by jasonwilliams** [comment]
>  And, if we could have some # Safety documentation in the Readable trait this would be perfect :)

@HalidOdat are you able to help explain the need for unsafe readable here?
https://github.com/boa-dev/boa/blob/main/boa/src/vm/code_block.rs#L26-L35

**Review by jasonwilliams — APPROVED** [review]

File: boa/src/vm/code_block.rs [hunk 2]
```diff
@@ -35,12 +39,15 @@ unsafe impl Readable for f32 {}
...
+///
+/// A CodeBlock is generated for each function compiled by the [ByteCompiler](crate::bytecompiler::ByteCompiler).
+/// It stores the bytecode and the other attributes of the function.
 #[derive(Debug, Trace, Finalize)]
 pub struct CodeBlock {
     /// Name of this function
     pub(crate) name: JsString,
 
-    /// The length of this function.
+    /// The number of parameters expected.
```
RageKnify, line 50: `parameters` should be `arguments`, the function expects to be called with N arguments because it was declared with N parameters.

RageKnify (approved):
LGTM but I think we might want to wait for @HalidOdat to say something about the documentation of the `unsafe` parts

HalidOdat (approved):

PR #1695, opened 2021-10-28
test(boa): add more coverage

neeldug:
It changes the following:

- improves coverage
- improves conformance for array_buffer
- tidies typos

### DIFF
--- a/boa/src/builtins/array/tests.rs
+++ b/boa/src/builtins/array/tests.rs
@@ -117,13 +117,15 @@ fn of() {
     assert_eq!(context.eval("a.length").unwrap(), JsValue::new(3));
 }
 
-#[ignore]
 #[test]
 fn concat() {
     let mut context = Context::new();
     let init = r#"
-    var empty = [];
-    var one = [1];
+        var empty = [];
+        var one = [1];
+        var spreadableLengthOutOfRange = {};
+        spreadableLengthOutOfRange.length = Number.MAX_SAFE_INTEGER;
+        spreadableLengthOutOfRange[Symbol.isConcatSpreadable] = true;
     "#;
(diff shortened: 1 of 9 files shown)


Razican:
Hi @neeldug it seems that the last tests are failing. Could you check those so that we can see if we can get this merged?

neeldug:
> Hi @neeldug it seems that the last tests are failing. Could you check those so that we can see if we can get this merged?

Hi, yes — sorry, I started on this a while back but I’ll take a look but from what I could see it was mishandling mutable state in the failing testcase, I’ll take a longer look at it and loop back if needed.

neeldug:
> Hi, yes — sorry, I started on this a while back but I’ll take a look but from what I could see it was mishandling mutable state in the failing testcase, I’ll take a longer look at it and loop back if needed.

Hi @Razican, I'm having a little difficulty in understanding exactly where the testcase is failing, given that currently it seems that the array in the callback function isn't being passed mutably whereas in Node, it behaves such that the array is mutable, whereas the element is not in the callback function. I'm not sure where to go about dealing with this in Boa.

Razican:
> > Hi, yes — sorry, I started on this a while back but I’ll take a look but from what I could see it was mishandling mutable state in the failing testcase, I’ll take a longer look at it and loop back if needed.
> 
> Hi @Razican, I'm having a little difficulty in understanding exactly where the testcase is failing, given that currently it seems that the array in the callback function isn't being passed mutably whereas in Node, it behaves such that the array is mutable, whereas the element is not in the callback function. I'm not sure where to go about dealing with this in Boa.

Hmmm seems something related with interior mutability. In any case, I guess we could just rebase this, ignore that test and merge it. It already improves coverage, which we really need.

PR #1697, opened 2021-11-02
[Merged by Bors] - Implement missing vm operations

raskad:
This implements most of the missing vm operations and fixes most of the panics.

The early errors for declarations in `for in` and `for of` loops are moved to the parser. The content of those `Node`s is changed accordingly.

### DIFF
--- a/boa/src/bytecompiler.rs
+++ b/boa/src/bytecompiler.rs
@@ -549,10 +549,13 @@ impl ByteCompiler {
                                 self.patch_jump(exit);
                             }
                             LogOp::Or => {
+                                self.emit_opcode(Opcode::Dup);
                                 let exit = self.jump_with_custom_opcode(Opcode::LogicalOr);
+                                self.emit_opcode(Opcode::Pop);
                                 self.compile_expr(binary.rhs(), true);
-                                self.emit(Opcode::ToBoolean, &[]);
+                                self.emit_opcode(Opcode::Dup);
                                 self.patch_jump(exit);
+                                self.emit_opcode(Opcode::Pop);
                             }
                             LogOp::Coalesce => {
                                 let exit = self.jump_with_custom_opcode(Opcode::Coalesce);
@@ -789,7 +792,7 @@ impl ByteCompiler {
             Node::ArrayDecl(array) => {
                 self.emit_opcode(Opcode::PushNewArray);
 
-                for element in array.as_ref().iter().rev() {
+                for element in array.as_ref() {
                     self.compile_expr(element, true);
                     if let Node::Spread(_) = element {
                         self.emit_opcode(Opcode::InitIterator);
@@ -812,7 +815,7 @@ impl ByteCompiler {
             Node::Call(_) => self.call(expr, use_expr),
             Node::New(_) => self.call(expr, use_expr),
             Node::TemplateLit(template_literal) => {
-                for element in template_literal.elements() {
+                for element in template_literal.elements().iter().rev() {
                     match element {
                         TemplateElement::String(s) => {
                             self.emit_push_literal(Literal::String(s.as_ref().into()))
@@ -844,11 +847,36 @@ impl ByteCompiler {
             Node::GeneratorExpr(_) => {
                 self.emit_opcode(Opcode::PushUndefined);
             }
+            // TODO: implement AsyncGeneratorExpr
+            Node::AsyncGeneratorExpr(_) => {
+                self.emit_opcode(Opcode::PushUndefined);
+            }
             // TODO: implement Yield
             Node::Yield(_) => {
                 self.emit_opcode(Opcode::PushUndefined);
             }
             Node::TaggedTemplate(template) => {
+                match template.tag() {
+                    Node::GetConstField(field) => {
+                        self.compile_expr(field.obj(), true);
+                        self.emit(Opcode::Dup, &[]);
+                        let index = self.get_or_insert_name(field.field());
+                        self.emit(Opcode::GetPropertyByName, &[index]);
+                    }
+                    Node::GetField(field) => {
+                        self.compile_expr(field.obj(), true);
+                        self.emit(Opcode::Dup, &[]);
+                        self.compile_expr(field.field(), true);
+                        self.emit(Opcode::Swap, &[]);
+                        self.emit(Opcode::GetPropertyByValue, &[]);
+                    }
+                    expr => {
+                        self.compile_expr(expr, true);
+                        self.emit_opcode(Opcode::This);
+                        self.emit_opcode(Opcode::Swap);
+                    }
+                }
+
                 for expr in template.exprs().iter().rev() {
                     self.compile_expr(expr, true);
                 }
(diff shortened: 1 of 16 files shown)


raskad:
#### Non-VM implementation
| Total | 86,464 | 86,464 | 0 |
| Passed | 40,037 | 40,113 | **+76** |
| Ignored | 18,869 | 18,869 | 0 |
| Failed | 27,558 | 27,482 | **-76** |
| Panics | 0 | 0 | 0 |
| Conformance | 46.30% | 46.39% | **+0.09%** |

#### VM implementation
| Total | 86,464 | 86,464 | 0 |
| Passed | 6,677 | 25,931 | **+19,254** |
| Ignored | 18,869 | 18,869 | 0 |
| Failed | 60,918 | 41,664 | **-19,254** |
| Panics | 60,047 | 614 | **-59,433** |
| Conformance | 7.72%25 | 29.99%25 | **+22.27%25** |

raskad:
Relevant for the review may be this change outside of the vm:

I moved the early error generation for declarations in `for in` and `for of` loops to the parser, as we have no way to throw early errors in the vm. The relevant files are:

boa/src/syntax/ast/node/iteration/for_in_loop/mod.rs
boa/src/syntax/ast/node/iteration/for_of_loop/mod.rs
boa/src/syntax/ast/node/iteration/mod.rs
boa/src/syntax/parser/statement/iteration/for_statement.rs

boa/src/vm/mod.rs
```diff
@@ -601,11 +832,7 @@ impl Context {
         const OPERAND_COLUMN_WIDTH: usize = COLUMN_WIDTH;
         const NUMBER_OF_COLUMNS: usize = 4;
 
-        let msg = if self.vm.frame().exit_on_return {
-            " VM Start"
-        } else {
-            " Call Frame "
-        };
+        let msg = " VM Start";
```
jasonwilliams, line 835: Is there a reason why you changed this?
The output shows "VM Start" twice when running because a CodeBlock also shows as "VM Start". 

This logic was to distinguish between the main VM running and a codeblock being executed

jasonwilliams (changes requested):

boa/src/vm/mod.rs
raskad: Fixed the output to show `Call Frame` and `VM Start` again. That just got lost when I removed `exit_on_return`.

Can you show an example how you think the output should look for the two other points? You mean just like this?:

```
-------------------------------------- VM Start---------------------------------------

Time         Opcode                   Operands                 Top Of Stack
1μs          PushLiteral              0                        "hello"
```

jasonwilliams (approved):

boa/src/vm/mod.rs
jasonwilliams: I would expect something like this:
```powershell
------------------------------ Compiled Output ----------------------------------------
    Location  Count   Opcode              Operands

    000000    0000    GetFunction         0000: '' (length: 0)
    000005    0001    DefInitLet          0000: 'o'
    000010    0002    PushEmptyObject     
    000011    0003    Dup                 
    000012    0004    GetFunction         0001: '' (length: 0)
    000017    0005    Swap                
    000018    0006    SetPropertyByName   0001: 'get'
    000023    0007    PushLiteral         0
    000028    0008    GetName             0000: 'o'
    000033    0009    GetName             0002: 'Object'
    000038    0010    Dup                 
    000039    0011    GetPropertyByName   0003: 'defineProperty'
    000044    0012    Call                3
    000049    0013    Pop                 
    000050    0014    GetName             0000: 'o'
    000055    0015    GetPropertyByName   0004: 'test'

Literals:
    0000: <string> "test"

Names:
    0000: o
    0001: get
    0002: Object
    0003: defineProperty
    0004: test

Functions:
    0000: name: '' (length: 0)
    0001: name: '' (length: 0)


-------------------------------------- VM Start---------------------------------------
Time         Opcode                   Operands                 Top Of Stack

42μs         GetFunction              0000: '' (length: 0)     [function]
15μs         DefInitLet               0000: 'o'                <empty>
5μs          PushEmptyObject                                   [object]
1μs          Dup                                               [object]
47μs         GetFunction              0001: '' (length: 0)     [function]
0μs          Swap                                              [object]
19μs         SetPropertyByName        0001: 'get'              [object]
1μs          PushLiteral              0                        "test"
6μs          GetName                  0000: 'o'                [function]
20μs         GetName                  0002: 'Object'           [function]
1μs          Dup                                               [function]
6μs          GetPropertyByName        0003: 'defineProperty'   [function]
45μs         Call                     3                        [function]
1μs          Pop                                               <empty>
5μs          GetName                  0000: 'o'                [function]

----------------- Call Frame '' [Func Identifier from above]---------------------------------------------
    Location  Count   Opcode              Operands

    000000    0000    PushLiteral         0
    000005    0001    Return              
    000006    0002    PushUndefined       
    000007    0003    Return              

Literals:
    0000: <string> "hello"

Names:
    <empty>

Functions:
    <empty>


-------------------------------------- Call Frame '' [Func Identifier from above]------------------------
Time         Opcode                   Operands                 Top Of Stack

1μs          PushLiteral              0                        "hello"
2μs          Return                                            "hello"
95μs         GetPropertyByName        0004: 'test'             "hello"

Stack:
0000         "hello"

"hello"
```

- The first section called "Compiled Output" or something along those lines, as its not ran at this point this is just the code generated. 
- A space between the headings and the output to make it easier to read.
- `name '' (length: 0)` being changed to `Call Frame` (do we care about the length?) If it can reference the function number its executing from that would be even more useful. so `Call Frame [001]` or something

boa/src/vm/mod.rs
```diff
@@ -310,11 +369,13 @@ impl Context {
...
             }
-            Opcode::JumpIfTrue => {
+            Opcode::JumpIfNotUndefined => {
                 let address = self.vm.read::<u32>();
-                if self.vm.pop().to_boolean() {
+                let value = self.vm.pop();
+                if !value.is_undefined() {
                     self.vm.frame_mut().pc = address as usize;
                 }
+                self.vm.push(value)
```
RageKnify, line 378: Is there a reason this line doesn't have a semicolon `;` ?

RageKnify (approved):

boa/src/vm/mod.rs
```diff
@@ -259,26 +295,49 @@ impl Context {
                 let index = self.vm.read::<u32>();
                 let name = self.vm.frame().code.variables[index as usize].clone();
 
-                self.create_mutable_binding(name.as_ref(), false, VariableScope::Function)?;
+                if !self.has_binding(name.as_ref())? {
+                    self.create_mutable_binding(name.as_ref(), false, VariableScope::Function)?;
```
RageKnify, line 299: I was looking at some failing tests and ended up comparing this to the interpreter and I think we're missing the step where we initialize variables with `undefined`, do you agree?
```suggestion
                    self.create_mutable_binding(name.as_ref(), false, VariableScope::Function)?;
                    self.initialize_binding(name.as_ref(), JsValue::undefined())?;
```

raskad: Yes that's right. I already fixed that in my current local version. I think I will just push my current fixes again. Maybe we can just get to test parity with this PR.

boa/src/bytecompiler.rs
```diff
@@ -858,24 +980,31 @@ impl ByteCompiler {
...
-                            self.emit(Opcode::DefConst, &[index]);
 
                             if let Some(expr) = decl.init() {
                                 self.compile_expr(expr, true);
-                                self.emit(Opcode::InitLexical, &[index]);
-                            };
+                                self.emit(Opcode::DefInitConst, &[index]);
+                            } else {
+                                self.emit(Opcode::DefConst, &[index]);
+                            }
```
RageKnify, line 994: I think this `else` is unreachable and we might not need the `DefConst` opcode, correct me if I'm wrong, but I think `const` declarations require an initializer.

Razican: Const declarations require an initializer always, yes. It's checked at parsing time.

RageKnify (changes requested):
Reviewing just to indicate that the `DefConst` opcode should be removed.

boa/src/vm/mod.rs
raskad: I implemented the empty lines and the switch to `Compiled Output`.

The functions ID's and the confusing output after a Call Frame returns require some more thought. I will open a issue for those, as I cannot think of a good way to solve this right now.

boa/src/bytecompiler.rs
raskad: I changed it to expect the initializer and removed to unused opcode. Nice catch.

RageKnify (approved):

PR #1704, opened 2021-11-08
[Merged by Bors] - Small test ux improvements

orndorffgrant:
Hi, first time contributor :)

I was playing with getting the test262 test runner working and found a couple small things that I thought would've made my life a bit easier. I've split them into separate commits. Just let me know if you disagree with any and we can drop that commit. And of course, let me know of any changes you'd like to see :)

The changes are:
- Adds details to the documentation on the `--suite` command to `boa_tester`. I was trying to pass a path starting with `./test262/` for a bit before I looked at the code and saw what it was doing.
- Changes the individual test output when verbosity is > 1. Because the tests are run in parallel, the "Result" line for a given test was frequently not immediately after the "Started" line. This made it hard to determine which test had failed. The new output includes the test name in the result line, and also changes the format of all the individual-test-output lines to begin with the test name.
- Adds a `--disable-parallelism` flag. Even with the adjustments to the test output, it was still a bit hard to follow. Running serially for small sub-suites produces output that can be more easily scanned IMO. I added this as a "disable" flag (as opposed to "enable parallelism") in order to maintain the current default of running in parallel.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,7 +93,9 @@ for debugging purposes by setting up the verbose flag twice, for example `-vv`. 
 each test that is executed, you can use the triple verbose (`-vvv`) flag.
 
 Finally, if you want to only run one sub-suite or even one test (to just check if you fixed/broke something specific),
-you can do it with the `-s` parameter, and then passing the path to the sub-suite or test that you want to run.
+you can do it with the `-s` parameter, and then passing the path to the sub-suite or test that you want to run. Note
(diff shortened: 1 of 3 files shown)


jedel1043 (approved):
Good job! This should make it a lot easier to test for missing features and catching bugs :)

orndorffgrant:
Thanks!

And I just noticed I missed a formatting error. Just pushed an update, hopefully it's green now

RageKnify (approved):

Razican (approved):
Looks good to me! In the future maybe we should have a log buffer in order to be able to parallelize and still get the correct output :)

PR #1718, opened 2021-11-21
[Merged by Bors] - Fix internal vm tests

raskad:
This PR fixes some vm implementation code. All our internal tests should now pass with the vm enabled.

There are only a few (~100) 262 tests left that currently break with the vm, that previously worked.


boa/src/vm/opcode.rs
```diff
@@ -729,6 +800,7 @@ impl Opcode {
             Opcode::Pop => "Pop",
             Opcode::Dup => "Dup",
             Opcode::Swap => "Swap",
+            Opcode::Swap2 => "Swap2",
```
jasonwilliams, line 803: Do we want a more descriptive name for this? "SwapFirstAndThird" or something?

jasonwilliams: Also could you give some explanation of why this is needed? Im sure its fine i just couldn't wrap my head around the examples its being used in. Thanks

jasonwilliams (approved):

jedel1043 (changes requested):
I'm seeing there's a lot of `OpCode`s that are used only once. Would it be feasible to create a `OpCode::CallRs` operation that takes a function pointer from the stack, then casts it to a `fn(&mut CodeBlock, &mut usize) -> String` and calls it with `function(self, pc)`?
Maybe it would impact the performance a bit because of the indirection, but it would reduce the complexity of `instruction_operands` and `OpCode` a lot :)

boa/src/builtins/function/mod.rs
```diff
@@ -580,7 +582,16 @@ impl BuiltInFunctionObject {
...
+            #[cfg(feature = "vm")]
+            (Function::VmOrdinary { .. }, Some(name)) if name.is_empty() => {
+                Ok("[Function (anonymous)]".into())
+            }
+            #[cfg(feature = "vm")]
+            (Function::VmOrdinary { .. }, Some(name)) => {
+                Ok(format!("[Function: {}]", &name).into())
+            }
+            #[cfg(feature = "vm")]
+            (Function::VmOrdinary { .. }, None) => Ok("[Function (anonymous)]".into()),
```
Razican, line 594: Are these strings defined in the spec? If so, could we add a link to the definition?

boa_tester/src/exec/js262.rs
```diff
@@ -87,11 +89,14 @@ fn eval_script(_this: &JsValue, args: &[JsValue], context: &mut Context) -> JsRe
...
     if let Some(source_text) = args.get(0).and_then(|val| val.as_string()) {
+        #[cfg(not(feature = "vm"))]
         match boa::parse(source_text.as_str(), false) {
             // TODO: check strict
             Err(e) => context.throw_type_error(format!("Uncaught Syntax Error: {}", e)),
-            Ok(script) => script.run(context),
+            Ok(statement_list) => statement_list.run(context),
         }
+        #[cfg(feature = "vm")]
+        context.eval(source_text.as_str())
```
Razican, line 99: Does this also throw type errors with the "Uncaught Syntax Error: " string? If not, I think it will not replicate the behaviour for parsing errors.

boa/src/vm/opcode.rs
```diff
@@ -471,6 +499,13 @@ pub enum Opcode {
     /// Stack: value, key, object **=>**
     SetPropertyByValue,
 
+    /// Defines a own property of an object by value.
+    ///
+    /// Operands:
+    ///
+    /// Stack: value, key, object **=>**
+    DefineOwnPropertyByValue,
```
HalidOdat, line 507: I think we should change the positions on this so its `object, key, value` so we don't have to call `swap2` (since its done a lot).

HalidOdat (commented):
Wow. Amazing work @raskad !  :tada:  ... I think after this PR we should remove the non-vm (ast walker)  implementation and switch to VM, since the conformance of AST walker and AM are so pretty much the same :)

boa/src/vm/opcode.rs
raskad: This was only needed for one specific thing.
Like @HalidOdat commented, I will change the `DefineOwnPropertyByValue` opcode and remove `Swap2`. This was leftover, because I implemented the `Define*` opcodes for object literals later. The previous `Set*` opcodes are used in other places, so I did not want to duplicate those.

boa_tester/src/exec/js262.rs
raskad: You're right, is does not throw a type error. I think I misread this. Nice find, I see how to best solve this best, as we do not have `StatementList::run` for the vm.

raskad:
> I'm seeing there's a lot of `OpCode`s that are used only once. Would it be feasible to create a `OpCode::CallRs` operation that takes a function pointer from the stack, then casts it to a `fn(&mut CodeBlock, &mut usize) -> String` and calls it with `function(self, pc)`? Maybe it would impact the performance a bit because of the indirection, but it would reduce the complexity of `instruction_operands` and `OpCode` a lot :)

I also though about this a bit, but I think it's probably better to just create the opcodes. I looked at how SpiderMonkey deals with some of this. For example they also have a [`CheckObjCoercible` opcode](https://searchfox.org/mozilla-central/source/js/src/vm/Opcodes.h#1409) that is also only used once.
Also I think it would only move the complexity, as it probably makes it more difficult to see how the vm works in some places.

RageKnify:
I think we're better with individual opcodes then creating a special one that actually invokes N other functions. That's just moving the indirection to a later point in execution, I think it would make it harder to understand the code and most likely not help performance.

boa/src/vm/code_block.rs
```diff
@@ -529,37 +551,46 @@ impl JsObject {
...
-                    };
+                // Push function arguments to the stack.
+                let args = match code.params.len().cmp(&args.len()) {
+                    Ordering::Greater => {
+                        let mut v = args.to_vec();
+                        v.extend(vec![JsValue::Undefined; code.params.len() - args.len()]);
+                        v
+                    }
+                    Ordering::Less | Ordering::Equal => args.to_vec(),
+                };
```
RageKnify, line 564: Same as 721 to 728, `let args = if code.params.len() > args.len() { x } else {y}`

boa/src/vm/mod.rs
```diff
@@ -736,18 +882,52 @@ impl Context {
...
                 self.vm.frame_mut().pop_env_on_return += 1;
             }
+            Opcode::PushFunctionEnvironment => {
+                let is_constructor = self.vm.frame().code.constructor;
+                let is_lexical = self.vm.frame().code.this_mode.is_lexical();
+                let current_env = self.get_current_environment();
+                let this = &self.vm.frame().this;
+
+                let new_env = FunctionEnvironmentRecord::new(
+                    this.clone().as_object().unwrap().clone(), //TODO: is this ok? this_function object on stack mb?
```
RageKnify, line 912: Is it not possible for `this` to not be an object? For example, with `Function.prototype.call`, the user can choose the `this` with which the function is invoked, wouldn't that result in possible panics?

boa/src/builtins/function/mod.rs
raskad: This is the `Function::toString` method. Most of it is implementation defined according to the spec: https://tc39.es/ecma262/#sec-function.prototype.tostring

boa/src/vm/mod.rs
raskad: Good catch. I looked at the spec a bit. `this` should always be an object as you can see in step 6 of [`OrdinaryCallBindThis`](https://tc39.es/ecma262/#sec-ordinarycallbindthis), that is called for call and construct. I added that conversion.

raskad:
@HalidOdat I've got a question for you. I want to fix the individual opcode documentation, as some of them differ. What order should the stack values have in the doc comments? e.g.:
`Stack: value, object => value, object`

What is the appropriate order / where is the top of the stack?
My initial idea is:
`Stack: top_of_stack, ..., bottom_of_stack => bottom_of_stack, ..., top_of_stack`

But I'm not sure how to best represent this.

HalidOdat:
Hmmm... IMO the `bottom_of_stack, ..., top_of_stack` is much easier to read and understand. I think the best way to represent it is `Stack: bottom_of_stack, ..., top_of_stack => bottom_of_stack, ..., top_of_stack`. The most relevant part in a stack is top element, Having the top_element at the end makes it more intuitive, since we don't have to shift the elements to the right.

```
// bottom_of_stack, ..., top_of_stack
[1] - [2] - [3]
// push 4
[1] - [2] - [3] - [4]
// pop
[1] - [2] - [3]
// pop
[1] - [2]

// vs. 

// top_of_stack, ..., bottom_of_stack
[3] - [2] - [1]
// push 4
[4] - [3] - [2] - [1]
// pop
[3] - [2] - [1]
// pop
[2] - [1]
```

RageKnify (approved):
LGTM, might want to wait for more approvals since its such a big PR

HalidOdat (approved):

Razican (approved):
LGTM :)

jasonwilliams:
@jedel1043

jedel1043 (approved):

jasonwilliams:
> Wow. Amazing work @raskad ! 🎉 ... I think after this PR we should remove the non-vm (ast walker) implementation and switch to VM, since the conformance of AST walker and VM are so pretty much the same :)

I’m also fine with this, would be good to focus on that going forward

PR #1734, opened 2021-12-06
[Merged by Bors] - Generic `JsResult<R>` in `context.throw_` methods

HalidOdat:
Previously when we had the `context.throw_` methods (like `context.thtrow_type_error()`) they were limited as to where we could call them, e.i. a function that returned `JsResult<JsValue>`. So we had to call the `context.construct_` methods with an explicit `Err()` enum wrap to throw in functions that returned non-jsvalues (which happens a lot).
Now, with this PR the throw methods have a generic `JsResult<R>` return that can return in any `JsResult<T>` returning function. Which cleans the API and makes the user experience a bit better.

```rust
return Err(context.construct_type_error("..."));
// to
return context.throw_type_error("...");
```

### DIFF
--- a/boa/src/bigint.rs
+++ b/boa/src/bigint.rs
@@ -166,7 +166,7 @@ impl JsBigInt {
         let y = if let Some(y) = y.inner.to_biguint() {
             y
         } else {
-            return Err(context.construct_range_error("BigInt negative exponent"));
+            return context.throw_range_error("BigInt negative exponent");
         };
 
         let num_bits = (x.inner.bits() as f64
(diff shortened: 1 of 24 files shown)


Razican (approved):
Thanks! For a second there I thought we already had this, but indeed, this extra generic parameter makes it much more useful, since it can basically be used everywhere.

I'm approving the PR, even if there seems to be something strange with the test suite. Could you check that? Maybe the sub-module needs updating.

HalidOdat:
> I'm approving the PR, even if there seems to be something strange with the test suite. Could you check that? Maybe the sub-module needs updating.

Yes, it was out of date, I updated it and committed it :)

raskad (approved):

RageKnify (approved):
bors r+

PR #1743, opened 2021-12-14
Switch to VM as the default execution environment

HalidOdat:
It changes the following:
- Remove AST walker code
- Remove unnecessary `Trace` and `Finalize` on AST
- Set VM as default execution


HalidOdat:
This is a duplicate of #1747 , so closing this.

PR #1746, opened 2021-12-15
[Merged by Bors] - Feature `JsArray`

HalidOdat:
This PR introduces a new API for JavaScript builtin objects in Rust (such as `Array`, `Map`, `Proxy`, etc). Rather than just expose the raw builtin functions as discussed here #1692 (though having raw API exposed may be nice as well), In this PR we introduce a very light wrapper around the raw API, for a more pleasant user experience. The wrapper implements functions that are specific to the wrapper type (for `Array` this would be methods like `pop`, `push`, etc) as well as implementing `Deref<Target = JsObject>` so we can call `JsObject` functions without converting to `JsObject` and `Into<JsValue>` for easy to `JsValue` conversions.

Please check `jsarray.rs` in the `examples`

### DIFF
--- a/boa/examples/jsarray.rs
+++ b/boa/examples/jsarray.rs
@@ -1,0 +1,107 @@
+use boa::{
+    object::{FunctionBuilder, JsArray},
+    Context, JsValue,
+};
+
+fn main() -> Result<(), JsValue> {
+    // We create a new `Context` to create a new Javascript executor.
+    let context = &mut Context::default();
+
+    // Create an empty array.
+    let array = JsArray::new(context);
+
+    assert!(array.is_empty(context)?);
+
+    array.push("Hello, world", context)?; // [ "Hello, world" ]
+    array.push(true, context)?; // [ "Hello, world", true ]
+
+    assert!(!array.is_empty(context)?);
+
+    assert_eq!(array.pop(context)?, JsValue::new(true)); // [ "Hello, world" ]
+    assert_eq!(array.pop(context)?, JsValue::new("Hello, world")); // [ ]
+    assert_eq!(array.pop(context)?, JsValue::undefined()); // [ ]
+
+    array.push(1, context)?; // [ 1 ]
+
+    assert_eq!(array.pop(context)?, JsValue::new(1)); // [ ]
+    assert_eq!(array.pop(context)?, JsValue::undefined()); // [ ]
+
+    array.push_items(
+        &[
+            JsValue::new(10),
+            JsValue::new(11),
+            JsValue::new(12),
+            JsValue::new(13),
+            JsValue::new(14),
+        ],
+        context,
+    )?; // [ 10, 11, 12, 13, 14 ]
+
+    array.reverse(context)?; // [ 14, 13, 12, 11, 10 ]
+
+    assert_eq!(array.index_of(12, None, context)?, Some(2));
+
+    // We can also use JsObject method `.get()` through the Deref trait.
+    let element = array.get(2, context)?; // array[ 0 ]
+    assert_eq!(element, JsValue::new(12));
+    // Or we can use the `.at(index)` method.
+    assert_eq!(array.at(0, context)?, JsValue::new(14)); // first element
+    assert_eq!(array.at(-1, context)?, JsValue::new(10)); // last element
+
+    // Join the array with an optional separator (default ",").
+    let joined_array = array.join(None, context)?;
+    assert_eq!(joined_array, "14,13,12,11,10");
+
+    array.fill(false, Some(1), Some(4), context)?;
+
+    let joined_array = array.join(Some("::".into()), context)?;
+    assert_eq!(joined_array, "14::false::false::false::10");
+
+    let filter_callback = FunctionBuilder::native(context, |_this, args, _context| {
+        Ok(args.get(0).cloned().unwrap_or_default().is_number().into())
+    })
+    .build();
+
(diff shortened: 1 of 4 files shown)


boa/examples/jsarray.rs
```diff
@@ -0,0 +1,53 @@
+// This example goes into the details on how to pass closures as functions
```
jasonwilliams, line 1: Looks good.
Would we move this into https://github.com/boa-dev/boa/pull/1161 once that's in?

boa/src/object/jsarray.rs
```diff
@@ -0,0 +1,145 @@
+use std::ops::Deref;
+
+use crate::{builtins::Array, object::JsObject, Context, JsResult, JsString, JsValue};
+
+// TODO: put in a better place.
+pub trait JsObjectType: Into<JsValue> + Into<JsObject> + Deref<Target = JsObject> {}
+
+#[derive(Debug, Clone)]
+pub struct JsArrayObject {
```
jasonwilliams, line 9: Might be good to document this if its going to be part of the public API

jasonwilliams (approved):
So my understanding here is this is a lightwrapper for a public API which internally calls the same things. 

Is this purely for DevEx for users or is there another reason for doing this?

Looks good though, no issues. Im assuming the src/object folder will eventually have all the external APIs?

jasonwilliams:
Sounds good to me, the exposed API could work quite well. I'm assuming src/object will eventually have all the APIs we plan to expose?

HalidOdat:
> Sounds good to me, the exposed API could work quite well. I'm assuming src/object will eventually have all the APIs we plan to expose?

Yeah, plan to add others like `JsMap`, `JsSet`, etc

RageKnify (approved):

Razican (approved):
I like this :) It just needs a re-base, and maybe fixing that TODO, but all in all, I like this approach to improve the API, it should be much easier to use from Rust.

boa/src/object/jsarray.rs
```diff
@@ -0,0 +1,344 @@
...
+
+/// JavaScript `Array` rust object.
+#[derive(Debug, Clone, Trace, Finalize)]
+pub struct JsArray {
+    inner: JsObject,
+}
+
+impl JsArray {
+    #[inline]
+    pub fn new(array: &[JsValue], context: &mut Context) -> Self {
```
jedel1043, line 21: Maybe this should have the same API as `create_array_from_list`. That way users can avoid cloning whole collections if they want to

https://github.com/boa-dev/boa/blob/b34d581db50bdef70e6598f96bcf04c19b618341/boa/src/builtins/array/mod.rs#L249-L251

jedel1043, line 21: We could copy the API of `std::Vec`, making `new` create a new, empty `JsArray`, `From` implementations for `&[JsValue]` and friends, `TryFrom<JsObject>` and a `FromIterator` implementation that uses `create_array_from_list`

jedel1043 (changes requested):
Looks good! Just some API suggestions that I would like to see implemented :)

boa/src/object/jsarray.rs
HalidOdat: > We could copy the API of `std::Vec`, making `new` create a new, empty `JsArray`, `from` methods for `&[JsValue]` and friends, `try_from::<JsObject>` and a `from_iter` that uses `create_array_from_list`

Sadly, the `From`, `TryFrom`, and `FromIterator` cannot be implemented since they only allow one argument, but we also need the `context`.

jedel1043: Maybe as associated functions?

boa/src/object/mod.rs
```diff
@@ -52,10 +52,15 @@ use self::internal_methods::{
...
 
 pub(crate) mod internal_methods;
+mod jsarray;
 mod jsobject;
 mod operations;
 mod property_map;
 
+pub use jsarray::*;
+
+pub trait JsObjectType: Into<JsValue> + Into<JsObject> + Deref<Target = JsObject> {}
```
jedel1043, line 62: Aren't the bounds a bit too restrictive? `Into<JsObject>` already implies `Into<JsValue>`, and with a blanket implementation of `impl<O: JsObjectType> Into<JsValue> for O` we can remove that bound.

The same can be said for `Into<JsObject>`, because `Deref<Target = JsObject>` implies `Into<JsObject>` by `object.deref().clone()`, but in this case I can see some implementors doing some kind of cleanup before cloning the `JsObject`.

jedel1043 (changes requested):

boa/src/object/mod.rs
HalidOdat: Hmmm.... The `JsObjectType` trait is an internal thing, that I used to ensure that types implement the required traits, and I should have made it private (`pub(crate)`), And even if we expose it. I think we should seal it.

> `impl<O: JsObjectType> Into<JsValue> for O` 

We should try not implement `Into` trait directly since implementing `From` implicitly implements it. And `impl<O: JsObjectType> From<O> for JsValue` conflicts with `impl<T: Into<JsString>> From<T> for JsValue` in `conversion.rs`

jedel1043: Ah, if it's an internal thing then that's okay, we can modify the bounds without breaking the public API :)

jedel1043 (approved):
Looks good!

Razican:
LGTM!

bors r+

PR #1752, opened 2021-12-18
Cleaning-up benchmarks

Razican:
This Pull Request does some clean-up in the benchmarks:

- Updates the Chart.js dependency
- Removes benchmark data older than 1 year
- Adds a Python script to automatically remove this data (if executed)
- Minimised the benchmark data

This is a short term solution while we improve our benchmarking.

### DIFF
--- a/dev/process.py
+++ b/dev/process.py
@@ -1,5 +1,3 @@
-#! /opt/homebrew/bin/python3
-
 import json
 from datetime import datetime, timezone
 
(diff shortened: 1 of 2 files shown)


dev/process.py
```diff
@@ -0,0 +1,24 @@
+#! /opt/homebrew/bin/python3
```
jasonwilliams, line 1: should this shebang be here? its in `/usr/local/bin/python3` on my machine

Razican: Probably not, I was using it for debugging :)

Issue #503, opened 2020-06-17
Regular expression flags should be checked at runtime, not at lexing time


LuoZijun:
CodeLine: https://github.com/boa-dev/boa/blob/e674da46e720225663be79357703511dd4a54695/boa/src/syntax/ast/token.rs#L109-L116

11.8.5 Regular Expression Literals: https://www.ecma-international.org/ecma-262/10.0/#sec-literals-regular-expression-literals

The ECMAScript spec said, `RegularExpressionFlags` is empty or `IdentifierPart`.

LuoZijun:
The runtime defined flags: g, i, m, s, u, y

[MDN] Advanced searching with flags: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Advanced_searching_with_flags_2

Razican:
Hi @LuoZijun could you explain a bit the issue you found? We are generating the regular expression flags as per the spec, right?

LuoZijun:
@Razican I mean the invalid `invalid regular expression flag` error should be decided by Runtime. Not Lexer.

Razican:
After reviewing the specification, I can see that the only static error that this should throw should be in case we find a unicode scape sequence in the flags:
https://www.ecma-international.org/ecma-262/10.0/#sec-literals-regular-expression-literals-static-semantics-early-errors

But then, on runtime, we get the Syntax Error:
https://www.ecma-international.org/ecma-262/10.0/#sec-regexpinitialize

The thing is that this would prevent some optimizations, and I can see that both V8 and Spidermonkey throw a syntax error statically, before running the code.

I tried this code to test it:
```javascript
function a() {
	return /abc/abc;
}
```

Note that this throws a Syntax Error even if it's not being executed.

In any case, is there a particular situation where this is causing pain? Do you have an example situation where this is behaving diferently to common browsers? Or any particular situation where this could be an issue?

Razican:
There are some early errors that must be taken into account: https://262.ecma-international.org/10.0/#sec-primary-expression-regular-expression-literals-static-semantics-early-errors

Razican:
After checking this thoroughly and following the Test262 suite, it turns out that these flags must be checked at parsing time (early error) and not at runtime, so I'm closing this issue.

PR #1758, opened 2021-12-22
[Merged by Bors] - Lexer string interning

### LINKED ISSUE #279: String interning

Razican:
This Pull Request is part of #279.

 It adds a string interner to Boa, which allows many types to not contain heap-allocated strings, and just contain a `NonZeroUsize` instead. This can move types to the stack (hopefully I'll be able to move `Token`, for example, maybe some `Node` types too.

Note that the internet is for now only available in the lexer. Next steps (in this PR or future ones) would include also using interning in the parser, and finally in execution. The idea is that strings should be represented with a `Sym` until they are displayed.

Talking about display. I have changed the `ParseError` type in order to not contain anything that could contain a `Sym` (basically tokens), which might be a bit faster, but what is important is that we don't depend on the interner when displaying errors.

The issue I have now is in order to display tokens. This requires the interner if we want to know identifiers, for example. The issue here is that Rust doesn't allow using a `fmt::Formatter` (only in nightly), which is making my head hurt. Maybe someone of you can find a better way of doing this.

Then, about `cursor.expect()`, this is the only place where we don't have the expected token type as a static string, so it's failing to compile. We have the option of changing the type definition of `ParseError` to contain an owned string, but maybe we can avoid this by having a `&'static str` come from a `TokenKind` with the default values, such as "identifier" for an identifier. I wanted for you to think about it and maybe we can just add that and avoid allocations there.

Oh, and this depends on the VM-only branch, so that has to be merged before :)

Another thing to check: should the interner be in its own module?

### DIFF
--- a/boa_interner/src/lib.rs
+++ b/boa_interner/src/lib.rs
@@ -1,0 +1,76 @@
+//! String interner for Boa.
+//!
+//! The idea behind using a string interner is that in most of the code, strings such as
+//! identifiers and literals are often repeated. This causes extra burden when comparing them and
+//! storing them. A string interner stores a unique `usize` symbol for each string, making sure
+//! that there are no duplicates. This makes it much easier to compare, since it's just comparing
+//! to `usize`, and also it's easier to store, since instead of a heap-allocated string, you only
+//! need to store a `usize`. This reduces memory consumption and improves performance in the
+//! compiler.
+
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/boa-dev/boa/main/assets/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/boa-dev/boa/main/assets/logo.svg"
+)]
+#![deny(
+    clippy::all,
+    unused_qualifications,
+    unused_import_braces,
+    unused_lifetimes,
+    unreachable_pub,
+    trivial_numeric_casts,
+    // rustdoc,
+    missing_debug_implementations,
+    missing_copy_implementations,
+    deprecated_in_future,
+    meta_variable_misuse,
+    non_ascii_idents,
+    rust_2018_compatibility,
+    rust_2018_idioms,
+    future_incompatible,
+    nonstandard_style,
+)]
+#![warn(clippy::perf, clippy::single_match_else, clippy::dbg_macro)]
+#![allow(
+    clippy::missing_inline_in_public_items,
+    clippy::cognitive_complexity,
+    clippy::must_use_candidate,
+    clippy::missing_errors_doc,
+    clippy::as_conversions,
+    clippy::let_unit_value,
+    rustdoc::missing_doc_code_examples
+)]
+
+use std::num::NonZeroUsize;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+use string_interner::{backend::BucketBackend, StringInterner, Symbol};
+
+/// The string interner for Boa.
+///
+/// This is a type alias that makes it easier to reference it in the code.
+pub type Interner = StringInterner<BucketBackend<Sym>>;
+
+/// The string symbol type for Boa.
+///
+/// This symbol type is internally a `NonZeroUsize`, which makes it pointer-width in size and it's
+/// optimized so that it can ocupy 1 pointer width even in an `Option` type.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+pub struct Sym {
+    value: NonZeroUsize,
+}
+
+impl Symbol for Sym {
+    #[inline]
+    fn try_from_usize(index: usize) -> Option<Self> {
+        NonZeroUsize::new(index.wrapping_add(1)).map(|value| Self { value })
+    }
+
+    #[inline]
+    fn to_usize(self) -> usize {
+        self.value.get() - 1
+    }
+}
--- a/boa_wasm/Cargo.toml
+++ b/boa_wasm/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.56"
 
 [dependencies]
 Boa = { path = "../boa", features = ["console"] }
-wasm-bindgen = "0.2.79"
+wasm-bindgen = "=0.2.78"
 getrandom = { version = "0.2.4", features = ["js"] }
 
 [lib]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7,6 +7,7 @@ name = "Boa"
 version = "0.13.0"
 dependencies = [
  "bitflags",
+ "boa_interner",
  "boa_unicode",
  "chrono",
  "criterion",
@@ -29,6 +30,17 @@ dependencies = [
  "serde",
  "serde_json",
  "unicode-normalization",
+]
+
+[[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
+dependencies = [
+ "getrandom",
+ "once_cell",
+ "version_check",
 ]
 
 [[package]]
@@ -94,6 +106,14 @@ dependencies = [
  "rustyline-derive",
  "serde_json",
  "structopt",
+]
+
+[[package]]
+name = "boa_interner"
+version = "0.13.0"
+dependencies = [
+ "serde",
+ "string-interner",
 ]
 
 [[package]]
@@ -493,6 +513,9 @@ name = "hashbrown"
 version = "0.11.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ab5ef0d4909ef3724cc8cce6ccc8572c5c817592e9285f5464f8e86f8bd3726e"
+dependencies = [
+ "ahash",
+]
 
 [[package]]
 name = "heck"
@@ -914,9 +937,9 @@ dependencies = [
 
 [[package]]
 name = "ppv-lite86"
-version = "0.2.15"
+version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ed0cfbc8191465bed66e1718596ee0b0b35d5ee1f41c5df2189d0fe8bde535ba"
+checksum = "eb9f9e6e233e5c4a35559a617bf40a4ec447db2e84c20b55a6f83167b7e57872"
 
 [[package]]
 name = "proc-macro-error"
@@ -944,18 +967,18 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.34"
+version = "1.0.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f84e92c0f7c9d58328b85a78557813e4bd845130db68d7184635344399423b1"
+checksum = "c7342d5883fbccae1cc37a2353b09c87c9b0f3afd73f5fb9bba687a1f733b029"
 dependencies = [
  "unicode-xid",
 ]
 
 [[package]]
 name = "quote"
-version = "1.0.10"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "38bc8cc6a5f2e3655e0899c1b848643b2562f853f114bfec7be120678e3ace05"
+checksum = "47aa80447ce4daf1717500037052af176af5d38cc3e571d9ec1c7353fc10c87d"
 dependencies = [
  "proc-macro2",
 ]
@@ -1103,9 +1126,9 @@ dependencies = [
 
 [[package]]
 name = "rustyline"
-version = "9.1.1"
+version = "9.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c38cfbd0a4d7df7aab7cf53732d5d43449d0300358fd15cd4e8c8468a956aca"
+checksum = "db7826789c0e25614b03e5a54a0717a86f9ff6e6e5247f92b369472869320039"
 dependencies = [
  "bitflags",
  "cfg-if",
@@ -1234,6 +1257,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d44a3643b4ff9caf57abcee9c2c621d6c03d9135e0d8b589bd9afb5992cb176a"
 
 [[package]]
+name = "string-interner"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e2531d8525b29b514d25e275a43581320d587b86db302b9a7e464bac579648"
+dependencies = [
+ "cfg-if",
+ "hashbrown",
+ "serde",
+]
+
+[[package]]
 name = "strsim"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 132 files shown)


Razican:
I will try to solve #503 with this.

Razican:
I tried to start implementing this for the parser, but the changes are huge. I created a project and will create new issues to implement the interner for the parser, the compiler and the executor, but I think this is ready for review and merge.

I did some benchmarks to see which backend was faster, and for now, I selected the fastest one in my machine. This might change once we implement the interner in more places of the engine.

Razican:
Benchmarks:
```

┌─────────┬──────────────────────────────────────────────┬───────────────────┬───────────────────────┬────────────┐
1734
│ (index) │                     name                     │  changesDuration  │    masterDuration     │ difference │
1735
├─────────┼──────────────────────────────────────────────┼───────────────────┼───────────────────────┼────────────┤
1736
│    0    │      'Arithmetic operations (Compiler)'      │ '1199.0±58.10ns'  │ '**1107.7±56.24ns**'  │   '+8.0'   │
1737
│    1    │     'Arithmetic operations (Execution)'      │   '3.0±0.25µs'    │   '**2.6±0.13µs**'    │   '+16'    │
1738
│    2    │       'Arithmetic operations (Parser)'       │   '7.5±0.30µs'    │   '**6.7±0.38µs**'    │   '+12'    │
1739
│    3    │          'Array access (Compiler)'           │ '1803.4±64.84ns'  │ '**1730.0±209.74ns**' │   '+4.0'   │
1740
│    4    │          'Array access (Execution)'          │   '14.7±0.72µs'   │   '**14.5±1.31µs**'   │   '+2.0'   │
1741
│    5    │           'Array access (Parser)'            │   '16.0±1.76µs'   │   '**14.2±1.30µs**'   │   '+13'    │
1742
│    6    │         'Array creation (Compiler)'          │   '2.9±0.11µs'    │   '**2.7±0.16µs**'    │   '+8.0'   │
1743
│    7    │         'Array creation (Execution)'         │   '3.8±0.23ms'    │   '**3.6±0.17ms**'    │   '+4.0'   │
1744
│    8    │          'Array creation (Parser)'           │   '17.4±0.99µs'   │   '**16.3±0.93µs**'   │   '+7.0'   │
1745
│    9    │            'Array pop (Compiler)'            │   '8.7±0.67µs'    │   '**8.2±0.43µs**'    │   '+6.0'   │
1746
│   10    │           'Array pop (Execution)'            │ '1660.7±138.91µs' │ '**1537.0±80.57µs**'  │   '+8.0'   │
1747
│   11    │             'Array pop (Parser)'             │  '194.8±11.57µs'  │  '**172.0±10.31µs**'  │   '+13'    │
1748
│   12    │      'Boolean Object Access (Compiler)'      │   '2.5±0.16µs'    │   '**2.3±0.11µs**'    │   '+6.0'   │
1749
│   13    │     'Boolean Object Access (Execution)'      │   '9.2±0.52µs'    │   '**8.3±0.34µs**'    │   '+12'    │
1750
│   14    │       'Boolean Object Access (Parser)'       │   '18.4±1.35µs'   │   '**16.9±1.08µs**'   │   '+9.0'   │
1751
│   15    │            'Clean js (Compiler)'             │   '6.7±0.34µs'    │   '**6.5±0.46µs**'    │   '+3.0'   │
1752
│   16    │            'Clean js (Execution)'            │ '1738.5±74.08µs'  │ '**1687.6±73.17µs**'  │   '+3.0'   │
1753
│   17    │             'Clean js (Parser)'              │   '38.0±2.51µs'   │   '**35.0±1.64µs**'   │   '+9.0'   │
1754
│   18    │                'Create Realm'                │  '436.1±19.86ns'  │  '**407.6±45.13ns**'  │   '+7.0'   │
1755
│   19    │ 'Dynamic Object Property Access (Compiler)'  │   '2.5±0.19µs'    │   '**2.2±0.11µs**'    │   '+12'    │
1756
│   20    │ 'Dynamic Object Property Access (Execution)' │   '10.1±0.62µs'   │   '**9.2±0.49µs**'    │   '+9.0'   │
1757
│   21    │  'Dynamic Object Property Access (Parser)'   │   '14.4±0.82µs'   │   '**12.5±0.69µs**'   │   '+15'    │
1758
│   22    │            'Fibonacci (Compiler)'            │   '3.7±0.17µs'    │   '**3.3±0.21µs**'    │   '+11'    │
1759
│   23    │           'Fibonacci (Execution)'            │   '3.9±0.25ms'    │   '**3.8±0.14ms**'    │   '+4.0'   │
1760
│   24    │             'Fibonacci (Parser)'             │   '22.5±1.12µs'   │   '**18.8±0.95µs**'   │   '+20'    │
1761
│   25    │            'For loop (Compiler)'             │   '3.0±0.19µs'    │   '**2.8±0.17µs**'    │   '+8.0'   │
1762
│   26    │            'For loop (Execution)'            │   '62.2±3.76µs'   │   '**56.4±3.92µs**'   │   '+10'    │
1763
│   27    │             'For loop (Parser)'              │   '19.8±1.33µs'   │   '**16.7±0.98µs**'   │   '+18'    │
1764
│   28    │             'Mini js (Compiler)'             │   '6.7±0.54µs'    │   '**6.2±0.44µs**'    │   '+9.0'   │
1765
│   29    │            'Mini js (Execution)'             │ '1601.3±62.24µs'  │ '**1569.4±68.19µs**'  │   '+2.0'   │
1766
│   30    │              'Mini js (Parser)'              │   '33.9±1.96µs'   │   '**30.5±1.99µs**'   │   '+11'    │
1767
│   31    │      'Number Object Access (Compiler)'       │   '2.1±0.10µs'    │ '**1914.9±83.96ns**'  │   '+9.0'   │
1768
│   32    │      'Number Object Access (Execution)'      │   '7.3±0.47µs'    │   '**6.8±0.63µs**'    │   '+7.0'   │
1769
│   33    │       'Number Object Access (Parser)'        │   '14.7±0.53µs'   │   '**13.1±1.23µs**'   │   '+13'    │
1770
│   34    │         'Object Creation (Compiler)'         │ '1890.2±107.99ns' │ '**1796.3±114.47ns**' │   '+5.0'   │
1771
│   35    │        'Object Creation (Execution)'         │   '9.3±0.49µs'    │   '**8.3±0.47µs**'    │   '+13'    │
1772
│   36    │          'Object Creation (Parser)'          │   '12.8±0.58µs'   │   '**10.7±0.60µs**'   │   '+19'    │
1773
│   37    │             'RegExp (Compiler)'              │   '2.2±0.11µs'    │   '**2.2±0.14µs**'    │   '+3.0'   │
1774
│   38    │             'RegExp (Execution)'             │   '17.2±1.02µs'   │   '**15.8±0.60µs**'   │   '+9.0'   │
1775
│   39    │              'RegExp (Parser)'               │   '13.9±0.71µs'   │   '**12.1±0.64µs**'   │   '+15'    │
1776
│   40    │         'RegExp Creation (Compiler)'         │ '1852.9±105.97ns' │ '**1839.4±131.38ns**' │   '+1.0'   │
1777
│   41    │        'RegExp Creation (Execution)'         │   '13.0±0.97µs'   │   '**12.3±0.71µs**'   │   '+5.0'   │
1778
│   42    │          'RegExp Creation (Parser)'          │   '10.9±0.55µs'   │   '**10.1±0.55µs**'   │   '+8.0'   │
1779
│   43    │         'RegExp Literal (Compiler)'          │   '2.2±0.08µs'    │   '**2.1±0.12µs**'    │   '+2.0'   │
1780
│   44    │         'RegExp Literal (Execution)'         │   '17.4±0.98µs'   │   '**16.1±1.05µs**'   │   '+8.0'   │
1781
│   45    │          'RegExp Literal (Parser)'           │   '10.5±0.48µs'   │   '**9.5±0.42µs**'    │   '+11'    │
1782
│   46    │     'RegExp Literal Creation (Compiler)'     │ '1814.1±82.13ns'  │ '**1758.7±74.86ns**'  │   '+3.0'   │
1783
│   47    │    'RegExp Literal Creation (Execution)'     │   '12.8±0.67µs'   │   '**12.4±0.83µs**'   │   '+4.0'   │
1784
│   48    │      'RegExp Literal Creation (Parser)'      │   '8.3±0.63µs'    │   '**7.4±0.38µs**'    │   '+11'    │
1785
│   49    │  'Static Object Property Access (Compiler)'  │   '2.0±0.13µs'    │ '**1928.4±292.73ns**' │   '+6.0'   │
1786
│   50    │ 'Static Object Property Access (Execution)'  │   '9.2±0.63µs'    │   '**8.5±0.50µs**'    │   '+8.0'   │
1787
│   51    │   'Static Object Property Access (Parser)'   │   '13.4±1.41µs'   │   '**11.7±0.91µs**'   │   '+15'    │
1788
│   52    │      'String Object Access (Compiler)'       │   '3.0±0.17µs'    │   '**2.9±0.16µs**'    │   '+5.0'   │
1789
│   53    │      'String Object Access (Execution)'      │   '11.7±0.62µs'   │   '**10.7±0.63µs**'   │   '+9.0'   │
1790
│   54    │       'String Object Access (Parser)'        │   '18.2±1.02µs'   │   '**16.1±0.65µs**'   │   '+13'    │
1791
│   55    │        'String comparison (Compiler)'        │   '3.1±0.21µs'    │   '**2.9±0.11µs**'    │   '+5.0'   │
1792
│   56    │       'String comparison (Execution)'        │   '10.5±2.90µs'   │   '**9.9±0.66µs**'    │   '+6.0'   │
1793
│   57    │         'String comparison (Parser)'         │   '14.8±0.88µs'   │   '**13.1±0.93µs**'   │   '+13'    │
1794
│   58    │      'String concatenation (Compiler)'       │   '2.2±0.15µs'    │   '**2.1±0.10µs**'    │   '+7.0'   │
1795
│   59    │      'String concatenation (Execution)'      │   '8.7±0.68µs'    │   '**8.1±0.49µs**'    │   '+7.0'   │
1796
│   60    │       'String concatenation (Parser)'        │   '9.8±0.75µs'    │   '**8.9±0.41µs**'    │   '+10'    │
1797
│   61    │           'String copy (Compiler)'           │ '1570.7±166.09ns' │ '**1492.5±90.95ns**'  │   '+5.0'   │
1798
│   62    │          'String copy (Execution)'           │   '7.8±0.53µs'    │   '**7.2±0.31µs**'    │   '+7.0'   │
1799
│   63    │            'String copy (Parser)'            │   '7.3±0.44µs'    │   '**6.6±0.43µs**'    │   '+11'    │
1800
│   64    │             'Symbols (Compiler)'             │ '1003.7±62.29ns'  │  '**935.5±59.47ns**'  │   '+7.0'   │
1801
│   65    │            'Symbols (Execution)'             │   '7.3±0.55µs'    │   '**6.9±0.40µs**'    │   '+7.0'   │
1802
│   66    │              'Symbols (Parser)'              │   '5.9±0.21µs'    │   '**4.9±0.25µs**'    │   '+19'    │
1803
│   67    │                      ''                      │     undefined     │       undefined       │   '+NaN'   │
1804
└─────────┴──────────────────────────────────────────────┴───────────────────┴───────────────────────┴────────────┘
```

RageKnify (approved):
Having the interner in a global hidden by an API might make it easier to develop `boa` (since we wouldn't need to pass its reference around everywhere), but I don't feel too strongly either way.

raskad:
> Having the interner in a global hidden by an API might make it easier to develop `boa` (since we wouldn't need to pass its reference around everywhere), but I don't feel too strongly either way.

@Razican correct me if I'm wrong, but I think that would not work, because the interner must be specific to one parsed block of code, so it can be deallocated when the code block is being dropped right?

Razican:
> > Having the interner in a global hidden by an API might make it easier to develop `boa` (since we wouldn't need to pass its reference around everywhere), but I don't feel too strongly either way.
> 
> 
> 
> @Razican correct me if I'm wrong, but I think that would not work, because the interner must be specific to one parsed block of code, so it can be deallocated when the code block is being dropped right?

If you don't de-allocate, you could use the same interner multiple times, in theory, but if the application is running for long, you might have a huge memory usage.

RageKnify:
Yeah, I guess depending on the use case you would want to drop the `Interner` to avoid a never shrinking heap structure.
If `boa` were embedded in something like node then I don't think this would matter, right? Since once everything is parsed we shouldn't intern that many more strings. 
Even in the browser, the only problem is a browser tab which keeps evaluating JS with new identifiers, right?

As I said, I can accept this choice, just not sure it makes much of a difference, if the `Interner` is part of the `Context` then the `cli` and future browsers (in the context of a given browser tab) would only have 1 at a time any ways.

Maybe for game engines or other embedding use cases it would be relevant to not use just 1 (`Interner`) since it would make sense to create new `Context`s and drop old ones when the new ones are created. In that case I can see the possibility of the heap growing because the user keeps changing the name of their variables/functions with "updates" to their script.

Razican:
I'm also thinking on servers using it as their scripting language of choice. If they can run multiple scripts, in different days or so, I think this approach could be better.

boa_interner/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+use std::num::NonZeroUsize;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+use string_interner::{backend::BucketBackend, StringInterner, Symbol};
+
+/// The string interner for Boa.
+///
+/// This is a type alias that makes it easier to reference it in the code.
+pub type Interner = StringInterner<BucketBackend<Sym>>;
```
jasonwilliams, line 53: How come you chose bucketbackend over stringbackend?

jasonwilliams (approved):
LGTM

I agree with Rageknify's concern around this being passed everywhere, but I don't see any other option from what has been discussed

raskad (approved):

boa_interner/src/lib.rs
raskad: Statics seem to be the reason. It probably makes sense that we try out the stringbackend when we use the interner in the parser.

Razican:
> Build failed:
>   * [Clippy](https://github.com/boa-dev/boa/runs/4904112785?check_suite_focus=true)

I will rebase this, it seems to fail due to Rust 1.58 lints.

boa_interner/src/lib.rs
Razican: The reason was that I tried locally all backends and this was the one giving better results, but I have no personal preference xD in the future once we use the interner everywhere we can benchmark again.

Razican:
> LGTM
> 
> I agree with Rageknify's concern around this being passed everywhere, but I don't see any other option from what has been discussed

One option here would be to have the interner in the parser, as a reference. Might make sense actually, and I might do so in the other PR, if you're OK with it.

Razican:
I had to revert the upgrade to Wasm-bindgen 0.2.79 due to https://github.com/rustwasm/wasm-bindgen/issues/2774.

boa_wasm/Cargo.toml
```diff
@@ -13,7 +13,7 @@ rust-version = "1.56"
 
 [dependencies]
 Boa = { path = "../boa", features = ["console"] }
-wasm-bindgen = "0.2.79"
+wasm-bindgen = "=0.2.78"
```
jasonwilliams, line 16: How come this is locked?

RageKnify: Razican said it just above, there's a bug in 0.2.79 that causes CI to fail.

Razican:
> > LGTM
> > I agree with Rageknify's concern around this being passed everywhere, but I don't see any other option from what has been discussed
> 
> One option here would be to have the interner in the parser, as a reference. Might make sense actually, and I might do so in the other PR, if you're OK with it.

Actually, now that I check it, the only way to do this would be to change all the methods to receive a `&mut Parser`, which would include a `&mut Interner`, but it might not be easy. I guess this should be done in a different PR.

PR #1773, opened 2022-01-03
Benchmarks page adapted to the upstream format.

Razican:
It turns out that we were still having empty benchmarks and using the upstream action changed the benchmarks format, so it was breaking out benchmarks page. This fixes both issues and also updates the JavaScript dependencies of the benchmarks page.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -1,0 +1,14 @@
+# Cleaning up benchmarks
+
+We have noticed that sometimes we store empty benchmarks (should be solved now that we use the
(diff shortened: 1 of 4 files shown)


RageKnify (approved):

jasonwilliams:
Not related to this fix, but do you mind adding a small readme file to dev folder explaining how to use the script and why we have it?

Razican:
> Not related to this fix, but do you mind adding a small readme file to dev folder explaining how to use the script and why we have it?

Done :)

jasonwilliams (approved):

PR #1774, opened 2022-01-03
[Merged by Bors] - Update the PR benchmarks action

Razican:
This is just for debugging for now. If we find out a fix, we can merge it.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -25,4 +25,4 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           branchName: ${{ github.base_ref }}
-          cwd: boa
+          cwd: ./boa
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -21,7 +21,7 @@ jobs:
             ~/.cargo/git
             ~/.cargo/registry
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - uses: boa-dev/criterion-compare-action@v3.0.1
+      - uses: boa-dev/criterion-compare-action@v3.0.2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


Razican:
This is now ready to be merged :)

.github/workflows/pull_request.yml
```diff
@@ -21,8 +21,8 @@ jobs:
             ~/.cargo/git
             ~/.cargo/registry
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - uses: boa-dev/criterion-compare-action@master
+      - uses: boa-dev/criterion-compare-action@v3.0.1
```
jedel1043, line 24: Is there a reason for pinning the action to the version 3.0.1? We could use `main` to avoid having to bump the version on every new change.

jedel1043 (commented):
Also, I realized the `criterion-compare-action` repo didn't have a version bump to 3.0.1 on the `package.json` (https://github.com/boa-dev/criterion-compare-action/blob/b12b240867eafcd503e063c17db8db3c4ab50d7e/package.json#L3)

.github/workflows/pull_request.yml
Razican: The idea behind this is to follow the nomenclature with the rest of actions. This action is now being used by more projects, not only Boa, so we should use the released versions. That way we can also develop the action in the `main` branch as needed, and release new versions with new features. Also, @dependabot will create PRs to update the action if we do it like this.

Razican:
> Also, I realized the `criterion-compare-action` repo didn't have a version bump to 3.0.1 on the `package.json` (https://github.com/boa-dev/criterion-compare-action/blob/b12b240867eafcd503e063c17db8db3c4ab50d7e/package.json#L3)

I will update this, I forgot it :/

jedel1043 (approved):
All good then!

Razican:
> All good then!

Fixed the version, released a new one and updated the GitHub action.

RageKnify (approved):

PR #1778, opened 2022-01-12
[Merged by Bors] - Adjust `compile` and `execute` to avoid clones

raskad:
It changes the following:

- Adjust the `context` methods `compile` and `execute` to avoid clones on `StatementList` and `CodeBlock`

### DIFF
--- a/boa/benches/full.rs
+++ b/boa/benches/full.rs
@@ -36,7 +36,7 @@ macro_rules! full_benchmarks {
                     let statement_list = Parser::new(CODE.as_bytes(), false).parse_all( &mut interner).expect("parsing failed");
                     c.bench_function(concat!($id, " (Compiler)"), move |b| {
                         b.iter(|| {
-                            Context::compile(black_box(statement_list.clone()));
+                            Context::compile(black_box(&statement_list));
                         })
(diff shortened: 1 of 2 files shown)


Razican (approved):

RageKnify (approved):
Maybe we should document somewhere that if the library users would like to execute a `CodeBlock` more than once they should clone the `Gc` instead of passing it to `execute` directly.

Razican:
> Maybe we should document somewhere that if the library users would like to execute a `CodeBlock` more than once they should clone the `Gc` instead of passing it to `execute` directly.

I guess this can be added to the documentation of `CodeBlock`, or to the function that executes it. As soon as that gets written we should merge this.

Razican:
> > Maybe we should document somewhere that if the library users would like to execute a `CodeBlock` more than once they should clone the `Gc` instead of passing it to `execute` directly.
> 
> I guess this can be added to the documentation of `CodeBlock`, or to the function that executes it. As soon as that gets written we should merge this.

I added the relevant documentation, LGTM :)

raskad:
Fixed a typos in the docs. Looks good to merge.

PR #1787, opened 2022-01-19
[Merged by Bors] - Copy/paste fix in Proxy error message

icecream17:
(Really small self-explanatory change - maybe the commit message could be better)

It changes the following:

When calling `new Proxy(undefined, {})` the error message now refers to the target `undefined` instead of the handler `{}`


Razican (approved):
Thanks for the fix!

RageKnify (approved):
bors r+

PR #1798, opened 2022-01-24
[Merged by Bors] - Convert `Codeblock` variables to `Sym`

raskad:
It changes the following:

- Convert `Codeblock` variables to `Sym`

PR #1801, opened 2022-01-25
Interner in the VM

Razican:
After adding the interner to the lexer and the parser, this PR extends its usage in the VM, in order to allow more optimisations.


raskad:
@Razican this one or  #1798? :D

Razican:
> @Razican this one or #1798? :D

Wow, I didn't see yours! Yours, of course, but I would add the tests change from this branch in `boa_interner`, along with the  `unwrap()` to `expect()` change in line 1547 of the bytecompiler.rs file (I will suggest this change) in the PR.

PR #1809, opened 2022-01-29
[Merged by Bors] - General code clean-up and new lint addition

Razican:
This PR adds some Clippy lints. Mainly, it adds the list of pedantic lints excluding some lints that were causing too many warnings. I also denied some useful restriction and pedantic lints, to make sure we use `Self` all the possible times (for better maintainability), and that we pass elements by reference where possible, for example, or that the documentation is properly written.

This might even have some small performance gains.

I also added a perfect hash function for the CLI keywords, which should be more efficient than a `HashSet`. This is something we could use elsewhere too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,7 +100,7 @@ dependencies = [
  "Boa",
  "colored",
  "jemallocator",
- "lazy_static",
+ "phf",
  "regex",
  "rustyline",
  "rustyline-derive",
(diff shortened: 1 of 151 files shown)


raskad (approved):
Looks really good!

I think enforcing a more consistent style across the codebase is a good idea.

HalidOdat (approved):
Looks good to me! :)

PR #1821, opened 2022-02-05
[Merged by Bors] - Reduced the size of AST nodes

Razican:
This PR boxes the `Try` and `TaggedTemplate` nodes reducing the size of the `Node` structure from 88 to 56 bytes. This should improve performance in most cases, by adding another indirection with code with `try` and templates.

### DIFF
--- a/boa_wasm/Cargo.toml
+++ b/boa_wasm/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.56"
 
 [dependencies]
 Boa = { path = "../boa", features = ["console"] }
-wasm-bindgen = "=0.2.78"
+wasm-bindgen = "=   0.2.78"
 getrandom = { version = "0.2.4", features = ["js"] }
 
 [lib]
--- a/boa_wasm/Cargo.toml
+++ b/boa_wasm/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.56"
 
 [dependencies]
 Boa = { path = "../boa", features = ["console"] }
-wasm-bindgen = "=   0.2.78"
+wasm-bindgen = "=0.2.78"
 getrandom = { version = "0.2.4", features = ["js"] }
 
 [lib]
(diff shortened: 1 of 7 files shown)


Razican:
I'm actually pretty surprised that the benchmark are not clearly showing improvements, even though there might be some small ones. There is also one benchmark that shows a 18% worsening, which is worrying me. Maybe this is just due to the GitHub Actions uncertainty (no dedicated processing power, for instance).

We should probably try to benchmark things locally or in a dedicated-core virtual server.

jasonwilliams:
@Razican have you tried running the profiler and looking at before and after? That may give you a deeper insight into what is going on.

https://github.com/boa-dev/boa/blob/main/docs/profiling.md

Razican:
> @Razican have you tried running the profiler and looking at before and after? That may give you a deeper insight into what is going on.
> 
> https://github.com/boa-dev/boa/blob/main/docs/profiling.md

I ran the profiler locally and didn't see much difference. Then I ran benchmarks, and I could clearly see a consistent 2-6% improvement in the parser and a consistent and clear 1-2% improvement in the execution. The compiler was having a 0-1% improvement, but it was not very clear.

I also checked this "Array creation (Execution)" benchmark and got a 1.8% improvement locally, so that 18% worsening is definitely not real.

boa_wasm/Cargo.toml
```diff
@@ -13,7 +13,7 @@ rust-version = "1.56"
 
 [dependencies]
 Boa = { path = "../boa", features = ["console"] }
-wasm-bindgen = "=0.2.78"
+wasm-bindgen = "=   0.2.78"
```
RageKnify, line 16: What does this do?

RageKnify (approved):
Confused about the change in `boa_wasm/Cargo.toml`, the rest looks fine.

boa_wasm/Cargo.toml
Razican: It was a typo, sorry, it has been fixed :)

HalidOdat (approved):

Issue #1788, opened 2022-01-19
`"true"[3]` is `undefined`


icecream17:
**Describe the bug**
String indexing doesn't work (gets `undefined`)

**To Reproduce**
Steps to reproduce the issue, or JavaScript code that causes this failure.

```js
"true"[3]
```

```js
"another example"[0]
```

**Expected behavior**
Should get the character (er, code point) at that index

**Build environment (please complete the following information):**

<https://boa-dev.github.io/boa/>

Razican:
This seems to be that this function is not fully spec compliant:
https://github.com/boa-dev/boa/blob/f66324cdf54357e92dfcd80f28419b365e410147/boa/src/object/internal_methods/string.rs#L144-L189

But I thought this should be working since #1354. @jedel1043 could you check this?

jedel1043:
I'll give it a look

jedel1043:
Ah, I think it's github pages, because locally it works correctly.
I checked the version published on the gh-pages branch and apparently it's the 0.10.0.
We would need to modify the action to recompile boa on each deploy.

Razican:
> Ah, I think it's github pages, because locally it works correctly.
> 
> I checked the version published on the gh-pages branch and apparently it's the 0.10.0.
> 
> We would need to modify the action to recompile boa on each deploy.

In GitHub pages we have the 0.13 version, so I guess that's the reason. 0.14 should be released reasonably soon, and that will be implemented.

jedel1043:
> In GitHub pages we have the 0.13 version, so I guess that's the reason. 0.14 should be released reasonably soon, and that will be implemented.

Hmm, that's weird. IIRC, #1354 was implemented before we released 0.13, so it should be on the release already...

Razican:
Maybe it didn't do the release in GH pages

Razican:
Indeed, the release workflow is failing, so the release in GitHub Pages is v0.10.0. I'm going to try and release v0.13, and see if there is something that can be fixed with the release workflow.

PR #1822, opened 2022-02-06
Updating the Boa playground with Boa v0.13

Razican:
It turns out, as found by #1788, that we were not releasing new versions in the Boa playground in https://boa-dev.github.io/boa/

It seems that the last version that was released was v0.10, so obviously, we were missing many new features there.

While we try to fix the release workflow for the upcoming v0.14, I'm updating GitHub pages so that it has the latest released version: v0.13.


jedel1043 (approved):
Nice! This should work in the meantime

RageKnify (approved):

PR #1828, opened 2022-02-07
Cleaning-up the playground after #1822

Razican:
Since we seem to be having some issues with the Boa playground, I re-built everything again. It seems I had done something wrong the previous time, since for example the Bootstrap version in GitHub pages is the 5.1.3, but it should be 5.1.1 (the one in Boa v0.13).

In any case, it seems there might be something wrong with the GitHub pages deployment, since the website seems to be missing [this](https://boa-dev.github.io/boa/node_modules_monaco-editor_esm_vs_basic-languages_javascript_javascript_js.js) file, but it can be found in the repository [here](https://github.com/boa-dev/boa/blob/fedb98a90e93b74e5cb1be22559cccef85654841/node_modules_monaco-editor_esm_vs_basic-languages_javascript_javascript_js.js).

This will hopefully fix it. If I serve it with `python3 -m http.server 8000` this works perfectly in my local machine.


RageKnify (approved):

jasonwilliams (approved):

jedel1043:
Hm, I tested the playground again and the problem still remains.

Razican:
> Hm, I tested the playground again and the problem still remains.

This is so strange, it works perfectly fine locally somehow, but it breaks in GitHub pages.

PR #1829, opened 2022-02-08
[Merged by Bors] - Refactor the environment for runtime performance

raskad:
This is an attempt to refactor the environments to be more performant at runtime. The idea is, to shift the dynamic hashmap environment lookups from runtime to compile time.

Currently the environments hold hashmaps that contain binding identifiers, values and additional information that is needed to identify some errors. Because bindings in outer environments are accessible from inner environments, this can lead to a traversal through all environments (in the worst case to the global environment).

This change to the environment structure pushes most of the work that is needed to access bindings to the compile time. At compile time, environments and bindings in the environments are being assigned indices. These indices are then stored instead of the `Sym` that is currently used to access bindings. At runtime, the indices are used to access bindings in a fixed size `Vec` per environment. This brings multiple benefits:
 - No hashmap access needed at runtime
 - The number of bindings per environment is known at compile time. Environments only need a single allocation, as their size is constant.
 - Potential for optimizations with `unsafe` https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked

Additionally, this changes the global object to have it's bindings directly stored on the `Realm`. This should reduce some overhead from access trough gc objects and makes some optimizations for the global object possible.

The benchmarks look not that great on the first sight. But if you look closer, I think it is apparent, that this is a positive change. The difference is most apparent on Mini and Clean as they are longer (still not near any real life js but less specific that most other benchmarks):

| Test | Base         | PR               | % |
|------|--------------|------------------|---|
| Clean js (Compiler) | **1929.1±5.37ns** | 4.1±0.02µs | **+112.53%** |
| Clean js (Execution) | 1487.4±7.50µs | **987.3±3.78µs** | **-33.62%** |

The compile time is up in all benchmarks, as expected. The percentage is huge, but if we look at the real numbers, we can see that this is an issue of orders of magnitude. While compile is up `112.53%`, the real change is `~+2µs`. Execution is only down `33.62%`, but the real time changed by `~-500µs`.


boa/src/builtins/function/arguments.rs
```diff
@@ -160,87 +161,83 @@ impl Arguments {
...
 
         // 17. Let mappedNames be a new empty List.
-        // using a set to optimize `contains`
-        let mut mapped_names = FxHashSet::default();
-
         // 12. Let parameterNames be the BoundNames of formals.
         // 13. Let numberOfParameters be the number of elements in parameterNames.
         // 18. Set index to numberOfParameters - 1.
         // 19. Repeat, while index ≥ 0,
         // a. Let name be parameterNames[index].
```
Razican, line 168: Could this somehow be mapped to the code? I guess we are changing a bit the algorithm here for performance, right?

boa/src/builtins/function/arguments.rs
```diff
@@ -160,87 +161,83 @@ impl Arguments {
...
+        let mut bindings = FxHashMap::default();
+        let mut property_index = 0;
+        'outer: for formal in formals {
+            for name in formal.names() {
+                if property_index >= len {
+                    break 'outer;
                 }
+                let binding_index = bindings.len() + 1;
+                let entry = bindings
+                    .entry(name)
```
Razican, line 181: The entry API is very nice for this :) it might be useful elsewhere in Boa too. It wasn't there when many things in the compiler where developed :)

boa/src/builtins/function/arguments.rs
```diff
@@ -160,87 +161,83 @@ impl Arguments {
...
+                // 4. Return getter.
+                FunctionBuilder::closure_with_captures(
+                    context,
+                    // 1. Let getterClosure be a new Abstract Closure with no parameters that captures
+                    // name and env and performs the following steps when called:
+                    |_, _, captures, _| Ok(captures.0.get(captures.1)),
+                    (env.clone(), *binding_index),
+                )
+                .length(0)
+                .name("")
```
Razican, line 202: Why is this empty name needed? Same happens another time below.

boa/src/builtins/function/arguments.rs
```diff
@@ -160,87 +161,83 @@ impl Arguments {
...
+                PropertyKey::from(*property_index),
+                PropertyDescriptor::builder()
+                    .set(p)
+                    .get(g)
+                    .enumerable(false)
+                    .configurable(true)
+                    .build(),
+                context,
+            )
+            .expect("[[DefineOwnProperty]] must not fail per the spec");
```
Razican, line 239: If it must not fail, can it not return a `Result`? (maybe it's out of the scope of this PR)

boa/src/builtins/number/mod.rs
```diff
@@ -83,23 +79,36 @@ impl BuiltIn for Number {
...
-        make_builtin_fn(
-            Self::parse_float,
-            "parseFloat",
-            &global,
-            PARSE_FLOAT_MAX_ARG_COUNT,
-            context,
-        );
-        make_builtin_fn(Self::global_is_finite, "isFinite", &global, 1, context);
-        make_builtin_fn(Self::global_is_nan, "isNaN", &global, 1, context);
+        fn build_in_function(
```
Razican, line 82: Shouldn't this be named `built_in_fn()`? Also, clippy is not happy about having the function definition after some statements.

In any case, if `make_builtin_fn()` is not being used, can it be replaced and have this function shared between all the different global objects?

boa/src/context.rs
```diff
@@ -726,13 +725,14 @@ impl Context {
             .constructor(true)
             .build();
 
-        self.global_object().insert_property(
-            name,
+        self.realm.global_bindings.string_property_map_mut().insert(
```
Razican, line 728: I would add some getters, things like `global_bindings()` to get the list of global bindings, and `global_bindings_mut()` to get the mutable reference, maybe directly to the context. Also, I would prefer if the `realm()` was not public.

boa/src/environment/lexical_environment.rs
```diff
@@ -4,164 +4,6 @@
 //!
 //! The following operations are used to operate upon lexical environments
 //! This is the entrypoint to lexical environments.
```
Razican, line 6: It seems that most of this file has been deleted, but tests kept. Would it make sense to just have a tests sub-module?

boa/src/environment/mod.rs
```diff
@@ -1,8 +1,3 @@
-//! Environment handling, lexical, object, function and declaritive records
+//! Environment handling, lexical, object, function and declarative records
 
-pub mod declarative_environment_record;
-pub mod environment_record_trait;
-pub mod function_environment_record;
-pub mod global_environment_record;
 pub mod lexical_environment;
-pub mod object_environment_record;
```
Razican, line 8: Maybe most of this can be removed, and everything that's left should come to a single file.

Razican (commented):
I started my review, will continue at a later time. Things are looking very, very, very good in this change! I added some comments on things where I need some background or where things could be improved, but overall, congratulations on the progress!!

boa/src/builtins/number/mod.rs
raskad: This was the only place where `make_builtin_fn()` is used on the global object. In some other places we use it for other objects. It does not work anymore for the global object, because it uses the `insert_property` helper method, that inserts on the `properties` of the object. These properties are not used for the global object anymore.

I think, I will modify or add another function like `Context::register_global_function` for this. I think the only problem with that one was, that it added the global function as a constructor.

boa/src/environment/lexical_environment.rs
raskad: Yeah, I'm not 100% sure where the code for the compile time and the runtime environments should be located now. I think it makes some sense to have all of that code close  to each other for documentation and readability. Either I will move that into the now empty environments module or into a separate file next to the realm and move the environment tests there too.

boa/src/builtins/function/arguments.rs
jedel1043: Probably the message I left was not descriptive enough. It should be changed to "Defining new own properties for a new ordinary object cannot fail" or something similar.

boa/src/environment/lexical_environment.rs
raskad: I moved all of the environment related code to a new `environments` module and removed the old one. I think it works pretty good that way

boa/src/context.rs
raskad: I also think the `Realm` should not be public, but let's leave that to a dedicated issue.

boa/src/builtins/number/mod.rs
raskad: I decided to go with `Context::register_global_builtin_function`.

boa/src/builtins/function/arguments.rs
raskad: Not really performance. The spec works with identifiers (`name`) but we do not have identifiers at runtime anymore. We have to work with the binding indices on the environments. The argument bindings are always the first few binding in the environment and we use that to decide what indices the getters and setters should have here.

For example:
```js
function f(a, b, c) {}
f(1,2,3)
```
Looks like this in the function environment:
| binding index | `arguments` property | identifier | value |
|---|---|---|---|
| 0 | arguments[0] | a | 1 |
| 1 | arguments[1] | b | 2 |
| 2 | arguments[2] | c | 3 |

This is pretty straightforward, but we can also have this:
```js
function f(a, a, c, c) {}
f(1,2,3)
```
Which translates to:
| binding index | `arguments` property | identifier | value |
|---|---|---|---|
| - | arguments[0] | - | - |
| 0 | arguments[1] | a | 2 |
| - | arguments[2] | - | - |
| 1 | arguments[3] | c | undefined |

As you can see, we have to work out the binding index based on the number of arguments and based on the parameter names.

I think I will add this comment or something similar as documentation to the code to explain how this is different from the spec.

Razican: That would be great, yes!!

boa/src/vm/call_frame.rs
```diff
@@ -15,11 +15,62 @@ pub struct CallFrame {
...
+    pub(crate) fn try_env_stack_loop_dec(&mut self) {
+        self.try_env_stack
+            .last_mut()
+            .expect("try environment stack entry must exist")
+            .num_loop_stack_entries -= 1;
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct TryStackEntry {
```
Razican, line 69: Could we add some documentation here?

boa/src/vm/call_frame.rs
```diff
@@ -15,11 +15,62 @@ pub struct CallFrame {
...
     pub(crate) finally_jump: Vec<Option<u32>>,
     pub(crate) pop_on_return: usize,
-    pub(crate) pop_env_on_return: usize,
+    pub(crate) loop_env_stack: Vec<usize>,
+    pub(crate) try_env_stack: Vec<TryStackEntry>,
     pub(crate) param_count: usize,
     pub(crate) arg_count: usize,
 }
 
+impl CallFrame {
```
Razican, line 24: I'm missing some documentation in these methods.

boa/src/bytecompiler.rs
```diff
@@ -109,6 +120,49 @@ impl<'b> ByteCompiler<'b> {
...
+                let binding = self.context.initialize_mutable_binding(name, false);
+                let index = self.get_or_insert_binding(binding);
+                self.emit(opcode, &[index]);
+            }
+            Opcode::DefInitConst => {
+                let binding = self.context.initialize_immutable_binding(name);
+                let index = self.get_or_insert_binding(binding);
+                self.emit(opcode, &[index]);
+            }
+            _ => panic!("invalid opcode passed to emit_binding"),
```
Razican, line 162: Could we add the opcode here to the panic? Also, adding the `#[track_caller]` to this function would help.

boa/src/bytecompiler.rs
```diff
@@ -343,17 +437,15 @@ impl<'b> ByteCompiler<'b> {
                 }
             } else if let Some(jump_info) = self.jump_info.last_mut() {
                 jump_info.breaks.append(&mut info.breaks);
+                jump_info.try_continues.append(&mut info.try_continues);
             }
         }
     }
 
     #[inline]
     fn compile_access<'a>(&mut self, node: &'a Node) -> Access<'a> {
```
Razican, line 446: Clippy is saying that the `self` binding is not being used anymore here, does it make sense to keep it?

boa/src/vm/code_block.rs
```diff
@@ -205,18 +221,30 @@ impl CodeBlock {
...
-            | Opcode::DeletePropertyByName
-            | Opcode::ConcatToString
-            | Opcode::CopyDataProperties => {
+            | Opcode::DeletePropertyByName => {
                 let operand = self.read::<u32>(*pc);
                 *pc += size_of::<u32>();
-                format!("{:04}: '{:?}'", operand, self.variables[operand as usize])
+                format!(
+                    "{:04}: '{}'",
+                    operand,
```
Razican, line 245:
```suggestion
                    "{operand:04}: '{}'",
```

boa/src/vm/mod.rs
```diff
@@ -298,88 +293,187 @@ impl Context {
...
+                            } => value.clone(),
+                            DescriptorKind::Accessor { get: Some(get), .. }
+                                if !get.is_undefined() =>
+                            {
+                                let get = get.clone();
+                                self.call(&get, &self.global_object().clone().into(), &[])?
+                            }
+                            _ => {
+                                return self
+                                    .throw_reference_error(format!("{} is not defined", key))
```
Razican, line 385:
```suggestion
                                    .throw_reference_error(format!("{key} is not defined"))
```

boa/src/vm/mod.rs
```diff
@@ -298,88 +293,187 @@ impl Context {
...
+                            }
+                        },
+                        _ => return self.throw_reference_error(format!("{} is not defined", key)),
```
Razican, line 388:
```suggestion
                        _ => return self.throw_reference_error(format!("{key} is not defined")),
```

boa/src/vm/mod.rs
```diff
@@ -298,88 +293,187 @@ impl Context {
...
+                    }
+                } else if let Some(value) = self.realm.environments.get_value_optional(
+                    binding_locator.environment_index(),
+                    binding_locator.binding_index(),
+                ) {
+                    value
+                } else {
+                    let name =
+                        JsString::from(self.interner().resolve_expect(binding_locator.name()));
+                    return self.throw_reference_error(format!("{} is not initialized", name));
```
Razican, line 398:
```suggestion
                    return self.throw_reference_error(format!("{name} is not initialized"));
```

boa/src/vm/mod.rs
```diff
@@ -298,88 +293,187 @@ impl Context {
...
-                    name,
+                if binding_locator.is_global() {
+                    let key: JsString = self
+                        .interner()
+                        .resolve_expect(binding_locator.name())
+                        .into();
+                    let exists = self.global_bindings_mut().contains_key(&key);
+
+                    if !exists && (self.strict() || self.vm.frame().code.strict) {
+                        return self.throw_reference_error("Binding already exists");
```
Razican, line 452: Could we add the binding name to the error message?

boa/src/vm/mod.rs
```diff
@@ -298,88 +293,187 @@ impl Context {
...
+                    let success = crate::object::internal_methods::global::global_set_no_receiver(
+                        &key.clone().into(),
+                        value,
+                        self,
+                    )?;
+
+                    if !success && (self.strict() || self.vm.frame().code.strict) {
+                        return self.throw_type_error(format!(
+                            "cannot set non-writable property: {}",
+                            key
```
Razican, line 464:
```suggestion
                            "cannot set non-writable property: {key}",
```

Razican (commented):
Looks pretty good already :) I added some comments and suggestions to improve the result. I don't think I can help a lot with the VM, since I'm still learning through it, and I'm missing some documentation, but the rest looks pretty good!

boa/src/context.rs
Razican: Could you open the issue?

boa/src/builtins/function/arguments.rs
raskad: Added!

boa/src/builtins/function/arguments.rs
raskad: Changed all of the `expect` messages in that file to that wording, as they all boil down to that.

boa/src/context.rs
```diff
@@ -719,22 +730,59 @@ impl Context {
...
+    ///
+    /// # Note
+    ///
+    /// The difference to [`Context::register_global_function`](Context::register_global_function) is,
+    /// that the function will not be `constructable`.
+    /// Usage of the function as a constructor will produce a `TypeError`.
+    #[inline]
+    pub fn register_global_builtin_function(
+        &mut self,
+        name: &str,
```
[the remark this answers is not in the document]
raskad: Because the name is a global object property, it is stored as a `JsString`. So `&str` makes sense here.

boa/src/environments.rs
```diff
@@ -0,0 +1,757 @@
...
+                    return BindingLocator::declarative(name, i, binding.index);
+                }
+                return BindingLocator::mutate_immutable(name);
+            }
+        }
+        BindingLocator::global(name)
+    }
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
raskad: True. I split it up further into `compile` and `runtime` as those are now the big logical sections.

boa/src/bytecompiler.rs
raskad: Good catch. I added a specific enum, so that this match can be exhaustive.

boa/src/bytecompiler.rs
```diff
@@ -38,6 +39,9 @@ struct JumpControlInfo {
     kind: JumpControlInfoKind,
     breaks: Vec<Label>,
     try_continues: Vec<Label>,
+    in_catch: bool,
```
Razican, line 42: At some point we could add all these booleans as bitflags, should be more efficient.

Razican (approved):
Looks good to me! LGTM!

boa/src/bytecompiler.rs
raskad: I'm not really satisfied with the current exception handling tbh. I will open an issue to think about different ways of how to handle all of that in the vm. I'm hoping that we can find a way to do it without all these fields at runtime.

boa/src/profiler.rs
```diff
@@ -31,12 +31,12 @@ impl BoaProfiler {
             .start_recording_interval_event(kind, id, thread_id)
     }
 
-    pub fn default() -> BoaProfiler {
-        let profiler = Profiler::new(Path::new("./my_trace")).unwrap();
-        BoaProfiler { profiler }
+    pub fn default() -> Self {
+        let profiler = Profiler::new(Path::new("./my_trace")).expect("must be able to crate file");
```
RageKnify, line 35:
```suggestion
        let profiler = Profiler::new(Path::new("./my_trace")).expect("must be able to create file");
```
I think it is a typo

boa/src/environments/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+//! To make the runtime more performant, environment specific behavior is split
+//! between bytecode compilation and the runtime.
+//! While the association of identifiers to values seems like a natural fit for a hashmap,
+//! lookups of the values at runtime are very expensive.
+//! Environments can also have outer environments.
+//! In the worst case, there are as many hashmap lookups, as there are environments.
+//!
+//! To avoid these costs, hashmaps are not used at runtime.
+//! At runtime, environments are represented as fixed size lists of binding values.
+//! The positions of the bindings in these lists is determined at compile time.
```
RageKnify, line 16: Here compile time refers to byte-code compilation time, right? As in not Rust compile time, I was initially confused by this and I think it might make sense to make the difference clear.

boa/src/environments/compile.rs
```diff
@@ -0,0 +1,315 @@
+use crate::{
+    environments::runtime::BindingLocator, property::PropertyDescriptor, Context, JsResult,
+    JsString, JsValue,
+};
+use boa_interner::Sym;
+use rustc_hash::FxHashMap;
+
+/// A compile time binding represents a binding at compile time in a [`CompileTimeEnvironment`].
+///
+/// It contains the binding index and a flag to indicate if this is an mutable binding or not.
```
RageKnify, line 10:
```suggestion
/// It contains the binding index and a flag to indicate if this is a mutable binding or not.
```

boa/src/environments/compile.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub(crate) fn num_bindings(&self) -> usize {
+        self.bindings.len()
+    }
+}
+
+/// The compile time environment stack contains a stack of all environments at compile time.
+///
+/// The first environment on the stack represents the global environment.
+/// This is never being deleted and is tied to the existence of the realm.
+/// All other environments are being dropped once they are not needed anymore a compile time.
```
RageKnify, line 38:
```suggestion
/// All other environments are being dropped once they are not needed anymore at compile time.
```

boa/src/environments/compile.rs
```diff
@@ -0,0 +1,315 @@
...
+pub(crate) struct CompileTimeEnvironmentStack {
+    stack: Vec<CompileTimeEnvironment>,
+}
+
+impl CompileTimeEnvironmentStack {
+    /// Creates a new compile time environment stack.
+    ///
+    /// This function should one be used once on realm creation.
+    #[inline]
+    pub(crate) fn new() -> Self {
```
RageKnify, line 49:
```suggestion
    /// Creates a new compile time environment stack.
    ///
    /// This function should only be used once, on realm creation.
    #[inline]
    pub(crate) fn new() -> Self {
```

boa/src/environments/mod.rs
raskad: Yeah, I think I wrote "bytecode copile time" in some other places. Probably also good to write that here.

raskad: I adjusted the docs to make this clear.

boa/src/environments/runtime.rs
```diff
@@ -0,0 +1,332 @@
...
+            .last()
+            .expect("global environment must always exist")
+            .clone()
+    }
+
+    /// Get the value of a binding.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the environment of binding index are out of range.
```
RageKnify, line 144:
```suggestion
    /// Panics if the environment or binding index are out of range.
```

boa/src/environments/runtime.rs
```diff
@@ -0,0 +1,332 @@
...
+            .get(binding_index)
+            .expect("binding index must be in range")
+            .clone()
+    }
+
+    /// Set the value of a binding.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the environment of binding index are out of range.
```
RageKnify, line 165:
```suggestion
    /// Panics if the environment or binding index are out of range.
```

boa/src/environments/runtime.rs
```diff
@@ -0,0 +1,332 @@
...
+            .expect("binding index must be in range");
+        *binding = Some(value);
+    }
+
+    /// Set the value of a binding if it is initialized.
+    /// Return `true` if the value has been set.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the environment of binding index are out of range.
```
RageKnify, line 190:
```suggestion
    /// Panics if the environment or binding index are out of range.
```

boa/src/environments/runtime.rs
```diff
@@ -0,0 +1,332 @@
...
+            *binding = Some(value);
+            true
+        }
+    }
+
+    /// Set the value of a binding if it is uninitialized.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the environment of binding index are out of range.
```
RageKnify, line 219:
```suggestion
    /// Panics if the environment or binding index are out of range.
```

RageKnify (approved):
LGTM, big change but the code looks good and I think the docs are at a point where other people can contribute.

PR #1830, opened 2022-02-08
NaN-boxed `JsValue`s

### LINKED ISSUE #1373: NaN boxed JavaScript Value

HalidOdat:
This Pull Request fixes/closes #1373 

This PR implements `JsValue` NaN boxing It is a work in progress, and requires a lot more changes.

### DIFF
--- a/boa_engine/src/value/mod.rs
+++ b/boa_engine/src/value/mod.rs
@@ -1,6 +1,27 @@
-//! This module implements the JavaScript Value.
+#![warn(unsafe_op_in_unsafe_fn)]
+//! This module implements the JavaScript [`JsValue`] type.
 //!
-//! Javascript values, utility methods and conversion between Javascript values and Rust values.
+//! [`JsValue`] implements several utility methods and conversions between Javascript
+//! values and Rust values.
+//!
+//! # Notes
+//!
+//! We recommend using [`JsValue::from`], [`JsValue::new`] or the [`Into::into`]
+//! trait if you need to convert from float types ([`f64`], [`f32`]) to [`JsValue`],
+//! since there are some checks implemented that convert float values to signed
+//! integer values when there's no loss of precision involved.
+//!
+//! Alternatively, you can use the [`JsValue::float64`] method if you do need a pure
+//! [`f64`] value. The constructor skips all conversions from [`f64`]s to
+//! [`i32`]s, even if the value can be represented by an [`i32`] without loss of
+//! precision.
+//!
+//! # Alternative implementations
+//!
+//! `boa` right now implements two versions of [`JsValue`]:
+//! - The default implementation using a simple `enum`.
+//! - A NaN-boxed implementation for x86-64 platforms that can be enabled using the `nan_boxing` feature.
+//! The feature flag is ignored on incompatible platforms.
 
 #[cfg(test)]
 mod tests;
@@ -12,10 +33,9 @@ use crate::{
     },
     object::{JsObject, ObjectData},
     property::{PropertyDescriptor, PropertyKey},
-    symbol::{JsSymbol, WellKnownSymbols},
+    symbol::WellKnownSymbols,
     Context, JsBigInt, JsResult, JsString,
 };
-use boa_gc::{custom_trace, Finalize, Trace};
 use boa_profiler::Profiler;
 use num_bigint::BigInt;
 use num_integer::Integer;
@@ -27,6 +47,11 @@ use std::{
     ops::Sub,
     str::FromStr,
 };
+
+mod sys;
+
+#[doc(inline)]
+pub use sys::{JsValue, JsVariant, Ref};
 
 mod conversions;
 pub(crate) mod display;
@@ -44,6 +69,7 @@ pub use hash::*;
 pub use integer::IntegerOrInfinity;
 pub use operations::*;
 pub use r#type::Type;
+pub(crate) use sys::PointerType;
 
 static TWO_E_64: Lazy<BigInt> = Lazy::new(|| {
     const TWO_E_64: u128 = 2u128.pow(64);
@@ -55,37 +81,6 @@ static TWO_E_63: Lazy<BigInt> = Lazy::new(|| {
     BigInt::from(TWO_E_63)
 });
 
-/// A Javascript value
-#[derive(Finalize, Debug, Clone)]
-pub enum JsValue {
-    /// `null` - A null value, for when a value doesn't exist.
-    Null,
-    /// `undefined` - An undefined value, for when a field or index doesn't exist.
-    Undefined,
-    /// `boolean` - A `true` / `false` value, for if a certain criteria is met.
-    Boolean(bool),
-    /// `String` - A UTF-8 string, such as `"Hello, world"`.
-    String(JsString),
-    /// `Number` - A 64-bit floating point number, such as `3.1415`
-    Rational(f64),
-    /// `Number` - A 32-bit integer, such as `42`.
-    Integer(i32),
-    /// `BigInt` - holds any arbitrary large signed integer.
-    BigInt(JsBigInt),
-    /// `Object` - An object, such as `Math`, represented by a binary tree of string keys to Javascript values.
-    Object(JsObject),
-    /// `Symbol` - A Symbol Primitive type.
-    Symbol(JsSymbol),
-}
-
-unsafe impl Trace for JsValue {
-    custom_trace! {this, {
-        if let Self::Object(o) = this {
-            mark(o);
-        }
-    }}
-}
-
 impl JsValue {
     /// Create a new [`JsValue`].
     #[inline]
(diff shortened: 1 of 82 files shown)


jedel1043:
Does https://github.com/rust-lang/rust/issues/73328 affect the WebAssembly build?

HalidOdat:
> > Does [rust-lang/rust#73328](https://github.com/rust-lang/rust/issues/73328) affect the WebAssembly build?
> 
> I don't think it would, but I'll need to test this :)

It works just fine :rocket:

boa/src/value/mod.rs
```diff
@@ -93,184 +128,791 @@ impl JsValue {
...
+                ValueTag::Symbol => {
+                    let pointer = self.as_pointer();
+                    let _ = std::mem::transmute::<_, JsSymbol>(pointer);
+                }
+                ValueTag::BigInt => {
+                    let pointer = self.as_pointer();
+                    let _ = std::mem::transmute::<_, JsBigInt>(pointer);
+                }
+                _ => {}
+            }
```
jedel1043, line 593: Seeing this I wonder if it would be worth to move the cast logic to each pointee type individually. I think it would be safer to use the `Rc::into_raw/from_raw` API instead of extensively using `transmute`. Also applies to the `NonNull::as_ptr/new/cast` API.
Should we abstract this as an unsafe trait, contained on a separate module?

boa/src/value/mod.rs
```diff
@@ -93,184 +128,791 @@ impl JsValue {
...
+                    let _ = std::mem::transmute::<_, JsString>(pointer);
+                }
+                ValueTag::Symbol => {
+                    let pointer = self.as_pointer();
+                    let _ = std::mem::transmute::<_, JsSymbol>(pointer);
+                }
+                ValueTag::BigInt => {
+                    let pointer = self.as_pointer();
+                    let _ = std::mem::transmute::<_, JsBigInt>(pointer);
+                }
```
jedel1043, line 591: Seeing this I wonder if it would be worth to move the cast logic to each pointee type individually. I think it would be safer to use the `Rc::into_raw/from_raw` API instead of extensively using `transmute`. Also applies to the `NonNull::as_ptr/new/cast` API.
Should we abstract this as an unsafe trait, contained on a separate module?

HalidOdat: > Seeing this I wonder if it would be worth to delegate the cast logic to each type individually. I think it would be safer to use the `Rc::into_raw/from_raw` API instead of extensively using `transmute`. Also applies to the `NonNull::as_ptr/new/cast` API.

I thought of using `Rc::into_raw()`/`Rc::from_raw()` but it does some unnecessary computations (offsetting the pointer than giving it for encoding and offsetting it back for decoding) so i just used transmute directly (for `Gc::from_raw` its even more costly).

> Should we abstract this as an unsafe trait, contained on a separate module?

Yeah, maybe, but this was all just me experimenting, and things will change and shift around as I think off better solutions as I go. (Any advice/suggestion is welcomed as always).

Hopefully I can rebase this today  :)

jedel1043:
Didn't want to squash the commits but the rebase was pretty painful otherwise.

boa_engine/src/value/mod.rs
```diff
@@ -88,184 +132,817 @@ impl JsValue {
...
+        let object = unsafe {
+            JsObject::from_void_ptr(self.as_pointer());
+        };
+
+        // SAFETY: Assuming the above is safe, this is safe because
+        // we currently are manipulating a [`ManuallyDrop`] that is
+        // manually dropped until this [`JsValue`] is dropped,
+        // which implies that the lifetime of `&object` can be extended
+        // to the lifetime of `&self`.
+        unsafe { std::mem::transmute(&object) }
```
HalidOdat, line 267: This is UB, we are returning a reference to a local function variable, which can be overridden.

Edit: Same for other `as_` methods.

boa_engine/src/value/mod.rs
```diff
@@ -88,184 +132,817 @@ impl JsValue {
...
 
-    /// Returns true if the value is a string.
     #[inline]
-    pub fn is_string(&self) -> bool {
-        matches!(self, Self::String(_))
+    unsafe fn root(&self) {
+        if let Some(o) = self.as_object() {
+            // SAFETY: `self.as_object()` must always return a valid `JsObject
+            unsafe { o.root() }
+        }
```
HalidOdat, line 668:
```suggestion
    unsafe fn root(&self) {
        if let Some(o) = self.as_object() {
            // SAFETY: `self.as_object()` must always return a valid `JsObject
            unsafe { o.root() }
        }
```
This will not root the object contained in `JsValue`, this will only root local object `o` (it's least significat bit is set, and does not update self)

This is why self is `Cell<u64>` and was set back after `root` and `unroot`.

boa_engine/src/value/mod.rs
```diff
@@ -88,184 +132,817 @@ impl JsValue {
...
-    /// Returns the string if the values is a string, otherwise `None`.
     #[inline]
-    pub fn as_string(&self) -> Option<&JsString> {
-        match self {
-            Self::String(ref string) => Some(string),
-            _ => None,
+    unsafe fn unroot(&self) {
+        if let Some(o) = self.as_object() {
+            // SAFETY: `self.as_object()` must always return a valid `JsObject
+            unsafe { o.unroot() }
```
HalidOdat, line 675:
```suggestion
    unsafe fn unroot(&self) {
        if let Some(o) = self.as_object() {
            // SAFETY: `self.as_object()` must always return a valid `JsObject
            unsafe { o.unroot() }
```
Same as `root`

boa_engine/src/value/mod.rs
jedel1043: AH RIGHT!
To be honest I've been thinking counfused for like 20 full minutes, but I forgot that `JsValue` now is a `Cell`, thanks.

jedel1043:
For now I undid returning refs. I'm thinking more carefully about how we can return refs without causing too much havoc, I'll make a PoC soon

HalidOdat:
> For now I undid returning refs. I'm thinking more carefully about how we can return refs without causing too much havoc, I'll make a PoC soon

The way I thought of doing it was by having a wrapper struct on `ManuallyDrop`.
```rust
#[derive(Debug)]
pub struct Ref<'a, T> {
    inner: ManuallyDrop<T>,
    _marker: PhantomData<&'a T>,
}

impl<T> Ref<'_, T> {
    #[inline]
    fn new(inner: ManuallyDrop<T>) -> Self {
        Self {
            inner,
            _marker: PhantomData,
        }
    }
}

impl<T> std::borrow::Borrow<T> for Ref<'_, T> {
    fn borrow(&self) -> &T {
        &*self.inner
    }
}

impl<T> AsRef<T> for Ref<'_, T> {
    fn as_ref(&self) -> &T {
        &*self.inner
    }
}

impl<T> std::ops::Deref for Ref<'_, T> {
    type Target = T;

    #[inline]
    fn deref(&self) -> &Self::Target {
        &*self.inner
    }
}
```
This struct must not implement `Drop`, since it would drop the `JsValue`. And anywhere we return `&JsObject`, `&JsString`, `&JsSymbol` and `JsBigInt` we return a corresponding `Ref<'_, T>` type.

HalidOdat:
>Nice! Should I make a PR with the changes or do you plan on doing it?

Sure, you can do it if you want! :) 
I'm going to be a bit busy this week, so don't think I'll have time to do it anyway 😅

jedel1043:
Is there anything else missing?

HalidOdat:
> Is there anything else missing?

Not sure, it should work. I'll debug it when I get some time this week :)

jedel1043:
@HalidOdat Fixed! There were some details about `Ref` that we didn't consider, but now it should work okay

jedel1043:
Relevant to this:

https://github.com/rust-lang/rust/issues/95228

jedel1043:
Also, this design breaks for 32-bit platforms and platforms with pointer size bigger than 48 bits. For that, I would propose using conditional compilation to replace this design with the old `JsValue` or a tagged pointer in those platforms.

HalidOdat:
> Also, this design breaks for platforms with pointer size bigger than 48 bits.

As far as I know no x86_64 CPU uses the full 64 bits of the pointer (that's why v8, spidermonkey and LuaJIT exploit this), that being said this could change (not likely in the near future at least for personal computers).

> For that, I would propose using conditional compilation to replace this design with the old `JsValue` or a tagged pointer in those platforms.

Yes, intended to put this under a feature flag, I would make this an opt-in feature

jedel1043:
> As far as I know no x86_64 CPU uses the full 64 bits of the pointer (that's why v8, spidermonkey and LuaJIT exploit this)...

Thought the same, but I did a bit more digging and apparently v8 uses tagged pointers instead of NaN boxing, meaning they allocate `f64`s on the heap. This can be seen here: https://v8docs.nodesource.com/node-0.8/d4/da0/v8_8h_source.html#l03857

However, they also implemented optimizations to unbox `f64`s on runtime.

jedel1043:
OK, this is pretty ready, I think. It's feature gated, `JsValue` doesn't implement `Send` anymore and all methods are properly documented.

HalidOdat (commented):
Looks good to me! :)

jedel1043:
Moving to 0.17, because I had a slightly safer API using a `JsValue` trait with GATs, and apparently they're about to be stabilized. https://github.com/rust-lang/rust/pull/96709

jedel1043:
Blocked since we want to first abstract `JsValue` so that we can modify its internal repr without changing the internal.

jedel1043:
Closing in favour of #4091.

PR #1832, opened 2022-02-11
[Merged by Bors] - Give the arrow function its proper name

rumpl:
With this change an arrow function name is correctly set to the name of the variable:

```javascript
const myFunction = () => {};
console.log(myFunction.name); // Prints "myFunction"
```

_Note:_ I'm still getting familiar with the codebase and am pretty new to Rust so I won't be offended if this isn't merged. I am actually surprised I had to make so many changes to give the right code the name it needed. Maybe there is a better way? I'm all ears :)

### DIFF
--- a/boa/src/bytecompiler.rs
+++ b/boa/src/bytecompiler.rs
@@ -1423,7 +1423,7 @@ impl<'b> ByteCompiler<'b> {
             ),
             Node::ArrowFunctionDecl(function) => (
                 FunctionKind::Arrow,
-                None,
+                function.name(),
                 function.params(),
                 function.body(),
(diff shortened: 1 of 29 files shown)


rumpl:
I force-pushed changes to the `syntax/parser/function/tests.rs` to make them pass

Razican (approved):
Thank you very much for the contribution! Indeed, it's a lot of changes, I guess mostly because the name retrieval is done at parsing time instead of execution time.

I think this is good, it should be faster this way, and we have other examples where this is done this way too.

Thank you for the contribution! It looks pretty good to me!

PR #1837, opened 2022-02-13
[Merged by Bors] - Rework RegExp struct to include bitflags field

### LINKED ISSUE #1819: The RegExp structure has too many booleans
As Clippy warns, the RegExp structure in [src/builtins/regexp/mod.rs](https://github.com/boa-dev/boa/blob/main/boa/src/builtins/regexp/mod.rs#L35-L61) has too many booleans (one for each flag). This is bad, because it might not get correctly optimised by the compiler. Ideally, we would want to use a bitflags structure such as the RegExpFlags one in https://github.com/boa-dev/boa/blob/60b7eb89340c266a2b4f180690e8781245b6a1c8/boa/src/syntax/lexer/regex.rs#L133-L144

For this particular case, I would recommend to just re-use that structure, and at some point, we want to create some kind of optimisation that doesn't just convert those flags to string to then run a RegExp constructor, which would again parse the flags and convert them to the RegExpFlags structure. For this we will almost certainly need a new AST node, but it's out of the scope of this issue.

aaronmunsters:
This Pull Request fixes/closes #1819.

It changes the following:

- Move the bitflags from `boa/src/syntax/lexer/regex.rs` to `boa/src/builtins/regexp/mod.rs`
- Replace the booleans in the RegExp struct to include the bitflags struct
- Update match expressions to make use of the bitflags struct

### DIFF
--- a/boa/src/builtins/regexp/mod.rs
+++ b/boa/src/builtins/regexp/mod.rs
@@ -23,8 +23,8 @@ use crate::{
     symbol::WellKnownSymbols,
     value::{IntegerOrInfinity, JsValue},
     BoaProfiler, Context, JsResult, JsString,
+    syntax::lexer::regex::RegExpFlags,
 };
-use bitflags::bitflags;
 use regexp_string_iterator::RegExpStringIterator;
 use regress::Regex;
 
@@ -1704,19 +1704,6 @@ impl RegExp {
 
         // 22. Return A.
         Ok(a.into())
-    }
-}
-
-bitflags! {
-    /// Flags of a regular expression.
-    #[derive(Default)]
-    pub struct RegExpFlags: u8 {
-        const GLOBAL = 0b0000_0001;
-        const IGNORE_CASE = 0b0000_0010;
-        const MULTILINE = 0b0000_0100;
-        const DOT_ALL = 0b0000_1000;
-        const UNICODE = 0b0001_0000;
-        const STICKY = 0b0010_0000;
     }
 }
 
--- a/boa/src/builtins/regexp/mod.rs
+++ b/boa/src/builtins/regexp/mod.rs
@@ -360,16 +360,16 @@ impl RegExp {
     }
 
     #[inline]
-    fn regexp_has_flag(this: &JsValue, flag: char, context: &mut Context) -> JsResult<JsValue> {
+    fn regexp_has_flag(this: &JsValue, flag: u8, context: &mut Context) -> JsResult<JsValue> {
         if let Some(object) = this.as_object() {
             if let Some(regexp) = object.borrow().as_regexp() {
                 return Ok(JsValue::new(match flag {
-                    'g' => regexp.flags.contains(RegExpFlags::GLOBAL),
-                    'm' => regexp.flags.contains(RegExpFlags::MULTILINE),
-                    's' => regexp.flags.contains(RegExpFlags::DOT_ALL),
-                    'i' => regexp.flags.contains(RegExpFlags::IGNORE_CASE),
-                    'u' => regexp.flags.contains(RegExpFlags::UNICODE),
-                    'y' => regexp.flags.contains(RegExpFlags::STICKY),
+                    b'g' => regexp.flags.contains(RegExpFlags::GLOBAL),
+                    b'm' => regexp.flags.contains(RegExpFlags::MULTILINE),
+                    b's' => regexp.flags.contains(RegExpFlags::DOT_ALL),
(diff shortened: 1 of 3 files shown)


raskad:
#### VM implementation
  | Total | 87,912 | 87,912 | 0 |
  | Passed | 40,892 | 40,892 | 0 |
  | Ignored | 19,929 | 19,929 | 0 |
  | Failed | 27,091 | 27,091 | 0 |
  | Panics | 0 | 0 | 0 |
  | Conformance | 46.51% | 46.51% | 0.00% |

boa/src/builtins/regexp/mod.rs
```diff
@@ -283,38 +266,33 @@ impl RegExp {
...
         // 5. If F contains any code unit other than "g", "i", "m", "s", "u", or "y"
         //    or if it contains the same code unit more than once, throw a SyntaxError exception.
-        let mut global = false;
-        let mut ignore_case = false;
-        let mut multiline = false;
-        let mut dot_all = false;
-        let mut unicode = false;
-        let mut sticky = false;
+        let mut flags = RegExpFlags::default();
         for c in f.chars() {
```
Razican, line 292: I would implement `FromStr` for `RegexFlags`, and use the version of `parse_regex_flags()` in the lexer, which iterates through bytes instead of characters, and should therefore be more efficient. Would also reduce duplicate code.

boa/src/builtins/regexp/mod.rs
```diff
@@ -391,12 +364,12 @@ impl RegExp {
         if let Some(object) = this.as_object() {
             if let Some(regexp) = object.borrow().as_regexp() {
                 return Ok(JsValue::new(match flag {
-                    'g' => regexp.global,
-                    'm' => regexp.multiline,
-                    's' => regexp.dot_all,
-                    'i' => regexp.ignore_case,
-                    'u' => regexp.unicode,
-                    'y' => regexp.sticky,
+                    'g' => regexp.flags.contains(RegExpFlags::GLOBAL),
```
Razican, line 367: Could we match bytes instead of characters here? Should be more efficient

boa/src/syntax/lexer/regex.rs
```diff
@@ -130,19 +130,6 @@ impl<R> Tokenizer<R> for RegexLiteral {
...
-        const GLOBAL = 0b0000_0001;
-        const IGNORE_CASE = 0b0000_0010;
-        const MULTILINE = 0b0000_0100;
-        const DOT_ALL = 0b0000_1000;
-        const UNICODE = 0b0001_0000;
-        const STICKY = 0b0010_0000;
-    }
-}
-
 fn parse_regex_flags(s: &str, start: Position, interner: &mut Interner) -> Result<Sym, Error> {
```
Razican, line 146: I would move this function together with the structure, and change it so that this implements the `FromStr` trait for the structure.

Razican (commented):
Looking very good! I added some suggestions :)

RageKnify:
Don't agree with moving the struct and accompanying functions from its original place.
It works against the idea of having the interpreter/environment depend on other parts of `boa`, ideally we'll be able to make the lexer and/or parser be separate crates to reduce iterative compile times. As is this PR would introduce dependencies going the other way.
Other than this the PR lgtm

Razican:
> Don't agree with moving the struct and accompanying functions from its original place.
> 
> It works against the idea of having the interpreter/environment depend on other parts of `boa`, ideally we'll be able to make the lexer and/or parser be separate crates to reduce iterative compile times. As is this PR would introduce dependencies going the other way.
> 
> Other than this the PR lgtm

Well, ideally this structure would ideally be part of regress. But I see what you mean. Maybe it should stay in the lexer.

boa/src/builtins/regexp/mod.rs
aaronmunsters: In my last commit the code matches on bytes, but this could be replaced with the `RegExpFlags` which would allow for the following reduction in matching code:

```rust
fn regexp_has_flag(this: &JsValue, flag: RegExpFlags, context: &mut Context) -> JsResult<JsValue> {
    if let Some(object) = this.as_object() {
        if let Some(regexp) = object.borrow().as_regexp() {
            return Ok(JsValue::new(regexp.flags.contains(flag)));
        }
        // ...
    }
    // ...
} 
```

Which would be more readable?

aaronmunsters:
I reworked the PR with the provided feedback. Feel free to leave new suggestions :)

RageKnify (approved):
LGTM

Razican (approved):

Razican:
This just needs running cargo fit

PR #1123, opened 2021-02-08
[Merged by Bors] - Fix string.prototype methods and add static string methods

jevancc:
This Pull Request fixes existing string prototype methods in #13 and adds static methods.

It changes the following:

- Fix bugs in existing string prototype methods and improve readability (e.g. rename variables to match the names in spec)
- Add static methods `String.raw`, `String.fromCharCode`, `String.fromCodePoint`
- Fix broken unit tests

PR #1840, opened 2022-02-15
[Merged by Bors] - Ignore wastefull `RegExp` tests

raskad:
With the implementation of `String.fromCodePoint` in #1123 some `RegExp` tests are now running for a long time. These tests check every unicode codepoint for all regexp property escape/character classes.

This not only makes the developer experience significantly worse, but also wastes cpu resources for the benefit of "completeness". I think these tests are completely useless. Ironically the unicode tables in the tests are generated - from the same data, that the unicode tables in the regex engine are also generated.

262 suite runtime:
Before: ~03:30 https://github.com/boa-dev/boa/runs/5191567446?check_suite_focus=true
After: ~31:00 https://github.com/boa-dev/boa/runs/5196405437?check_suite_focus=true

### DIFF
--- a/test_ignore.txt
+++ b/test_ignore.txt
@@ -15,3 +15,31 @@ feature:tail-call-optimization
 // These generate a stack overflow
 tco-call
 tco-member
+
+// RegExp tests that check individual codepoints.
+// They are not usefull in comparision to the cpu time they waste.
+feature:regexp-unicode-property-escapes
+character-class-non-whitespace-class-escape-plus-quantifier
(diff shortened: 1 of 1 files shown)


RageKnify (approved):
I just ran test262 on `main` and on this code and the difference was huge:
`main` = 21m13s
`ignore-wastefull-regexp-tests` = 59s
Definitely seems like a good choice.

Razican (approved):
I'm OK with this, but at some point we should run those tests, maybe before releases, or when we improve their performance.

PR #1842, opened 2022-02-16
[Merged by Bors] - Unwrap removal

Razican:
This removes all the calls to  `unwrap()` in the codebase, which made me found a couple of places where it wasn't needed, and could be improved. I also noticed we don't have dependabot updates for the test262 submodule and the interner dependencies, so I added those.

I added lints so that no new unwraps are added.

### DIFF
--- a/boa/src/syntax/parser/expression/mod.rs
+++ b/boa/src/syntax/parser/expression/mod.rs
@@ -489,9 +489,9 @@ where
     fn parse(self, cursor: &mut Cursor<R>, interner: &mut Interner) -> ParseResult {
         let _timer = BoaProfiler::global().start_event("Relation Expression", "Parsing");
 
-        if None::<InputElement>.is_some() {
-            cursor.set_goal(None::<InputElement>.unwrap());
-        }
+        // if None::<InputElement>.is_some() {
+        //     cursor.set_goal(None::<InputElement>.unwrap());
+        // }
 
         let mut lhs =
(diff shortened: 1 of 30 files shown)


boa/src/syntax/parser/expression/mod.rs
```diff
@@ -489,9 +489,9 @@ where
     fn parse(self, cursor: &mut Cursor<R>, interner: &mut Interner) -> ParseResult {
         let _timer = BoaProfiler::global().start_event("Relation Expression", "Parsing");
 
-        if None::<InputElement>.is_some() {
-            cursor.set_goal(None::<InputElement>.unwrap());
-        }
+        // if None::<InputElement>.is_some() {
+        //     cursor.set_goal(None::<InputElement>.unwrap());
+        // }
```
RageKnify, line 494: What was the point of this?

RageKnify (approved):

boa/src/syntax/parser/expression/mod.rs
Razican: Oh, leftover code, I was just surprised about the `None.is_some()` check, I will remove it

RageKnify: But shouldn't it panic unconditionally? `None` is obviously not `Some`

Razican: It panics always, but it checks before if it's some, so it never enters the branch.

PR #1849, opened 2022-02-20
[Merged by Bors] - Refactor mapped `Arguments` object

raskad:
This refactors the representation of the `[[ParameterMap]]` internal slot on the `Arguments` exotic object to be faster at runtime.

Previously `[[ParameterMap]]` was a `JsObject` like the spec describes. This can be pretty slow a runtime, because the argument getters and setters must be represented as function objects on the `[[ParameterMap]]` object. In addition to the time spend on creation and calling of those functions, every getter/setter needs a cloned gc reference to the function environment to access the bindings. This adds to the gc overhead.

The spec states that the `[[ParameterMap]]` internal slot doesn't have to be a `JsObject`. See NOTE 3 here: https://tc39.es/ecma262/#sec-arguments-exotic-objects
Leveraging this freedom, we can use a more optimized representation, that avoids any `JsObject` usage and only needs one clone of the function environment.

### DIFF
--- a/boa/src/builtins/function/arguments.rs
+++ b/boa/src/builtins/function/arguments.rs
@@ -21,14 +21,21 @@ pub struct ParameterMap {
 }
 
 impl ParameterMap {
+    /// Deletes the binding with the given index from the parameter map.
     pub(crate) fn delete(&mut self, index: usize) {
         if let Some(binding) = self.binding_indices.get_mut(index) {
             *binding = None;
         }
     }
 
-    // 10.4.4.7.1 MakeArgGetter ( name, env )
-    // https://tc39.es/ecma262/#sec-makearggetter
+    /// Get the value of the binding at the given index from the function environment.
+    ///
+    /// Note: This function is the abstract getter closure described in 10.4.4.7.1 `MakeArgGetter ( name, env )`
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-makearggetter
     pub(crate) fn get(&self, index: usize) -> Option<JsValue> {
         if let Some(Some(binding_index)) = self.binding_indices.get(index) {
             return Some(self.environment.get(*binding_index));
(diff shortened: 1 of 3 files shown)


boa/src/builtins/function/arguments.rs
```diff
@@ -2,28 +2,53 @@ use crate::{
...
+/// This struct stores all the data to access mapped function parameters in their environment.
 #[derive(Debug, Clone, Trace, Finalize)]
-pub struct MappedArguments(JsObject);
+pub struct ParameterMap {
+    binding_indices: Vec<Option<usize>>,
+    environment: Gc<DeclarativeEnvironment>,
+}
+
+impl ParameterMap {
+    pub(crate) fn delete(&mut self, index: usize) {
```
Razican, line 24:
```suggestion
    /// Deletes the binding with the given index from the parameter map.
    pub(crate) fn delete(&mut self, index: usize) {
```

boa/src/builtins/function/arguments.rs
```diff
@@ -2,28 +2,53 @@ use crate::{
...
+        if let Some(binding) = self.binding_indices.get_mut(index) {
+            *binding = None;
+        }
+    }
 
-impl MappedArguments {
-    pub(crate) fn parameter_map(&self) -> JsObject {
-        self.0.clone()
+    // 10.4.4.7.1 MakeArgGetter ( name, env )
+    // https://tc39.es/ecma262/#sec-makearggetter
```
Razican, line 31: Could we use documentation comments here?

boa/src/builtins/function/arguments.rs
```diff
@@ -2,28 +2,53 @@ use crate::{
...
+    pub(crate) fn get(&self, index: usize) -> Option<JsValue> {
+        if let Some(Some(binding_index)) = self.binding_indices.get(index) {
+            return Some(self.environment.get(*binding_index));
+        }
+        None
+    }
+
+    // 10.4.4.7.2 MakeArgSetter ( name, env )
+    // https://tc39.es/ecma262/#sec-makeargsetter
```
Razican, line 40: Here also, we should use documentation comments, and add a bit of explanation.

Razican (approved):
Looks good to me :) I think it could use some better documentation in some functions, though.

jedel1043 (approved):
Benchmarks look good! Especially the Fibonacci benchmark with a 30+% reduction in runtime. That's probably because the engine doesn't need to slowly create the mapped arguments object on each recursive call, right?

HalidOdat (approved):
Looks good to me :)

PR #1850, opened 2022-02-20
[Merged by Bors] - Handle allocation errors

### LINKED ISSUE #1847: Boa wrongly assumes that allocating never fails and writes to a null pointer
Boa assumes that `std::alloc::alloc` never fails (returning a null pointer) even though it can.

https://github.com/boa-dev/boa/blob/29cd909f88d9bcab59eedda95639756a45235d70/boa/src/string.rs#L219-L223

Quoting the [docs](https://doc.rust-lang.org/stable/core/alloc/trait.GlobalAlloc.html#tymethod.alloc):
> `unsafe fn alloc(&self, layout: Layout) -> *mut u8`
> Allocate memory as described by the given layout.
> Returns a pointer to newly-allocated memory, **or null to indicate allocation failure**.

If the allocation fails, then there will be a write on a null pointer on L223, which is considered [UB](https://doc.rust-lang.org/stable/std/ptr/index.html#safety):
> - A [null](https://doc.rust-lang.org/stable/std/ptr/fn.null.html) pointer is never valid, not even for accesses of [size zero](https://doc.rust-lang.org/stable/nomicon/exotic-sizes.html#zero-sized-types-zsts).

This also leads to violating the `NonNull::new_unchecked` invariant:
> `ptr` must be non-null.

https://github.com/boa-dev/boa/blob/29cd909f88d9bcab59eedda95639756a45235d70/boa/src/string.rs#L241

**To Reproduce**
```rs
use boa::JsString;

fn main() {
    // choose a string length that's large enough to cause an allocation error when concatenated later but
    // not too large so the initial allocation still succeeds
    let string = "a".repeat(1000 * 1000 * 1000);
    let ve = vec![&string as &str; 1000];

    // this is almost certainly going to cause an allocation error unless you've got lots of RAM
    // if it does, then this will write to a null pointer 💥
    let _ = JsString::concat_array(&ve);
}
```
Alternatively, one can override the global allocator and return null for that specific allocation: [link](https://gist.github.com/y21/64d8f2ebb9635b3e4fa95ca0a87bc497).
Implementing `GlobalAlloc` requires unsafe due to the nature of allocators being unsafe, but I'm fairly certain it's totally valid (albeit weird) for a crate to expose a safe API for an allocator like that.
Running the code under valgrind sums it up:
```sh
==22393== Process terminating with default action of signal 11 (SIGSEGV)
==22393==  Access not within mapped region at address 0x0
==22393==    at 0x110477: boa::string::JsString::concat_array
==22393==    by 0x11019F: boatest::main

[1]    22393 segmentation fault (core dumped)  valgrind ./target/release/boatest
```

**Expected behavior**
Handling null pointers, e.g. `assert!(!inner.is_null())`. Or better yet, calling [`std::alloc:handle_alloc_error`](https://doc.rust-lang.org/stable/alloc/alloc/fn.handle_alloc_error.html) as suggested in the docs

**Build environment**
- OS: Ubuntu
- Version: 20.04
- Target triple: x86_64-unknown-linux-gnu (installed)
- Rustc version: rustc 1.60.0-nightly (75d9a0ae2 2022-02-16)

y21:
Fixes #1847 by wrapping the `std::alloc::alloc()` call in `try_alloc()`, which checks that the returned pointer is non-null and handles allocation errors that way. It will now abort the process instead of executing UB in the error path

### DIFF
--- a/boa/src/string.rs
+++ b/boa/src/string.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 use rustc_hash::FxHashSet;
 use std::{
-    alloc::{alloc, dealloc, Layout},
+    alloc::{alloc, dealloc, handle_alloc_error, Layout},
     borrow::Borrow,
     cell::Cell,
     hash::{Hash, Hasher},
@@ -174,6 +174,14 @@ const MAX_CONSTANT_STRING_LENGTH: usize = {
     max
 };
 
+unsafe fn try_alloc(layout: Layout) -> *mut u8 {
+    let ptr = alloc(layout);
+    if ptr.is_null() {
+        handle_alloc_error(layout);
+    }
+    ptr
+}
+
 thread_local! {
     static CONSTANTS: FxHashSet<JsString> = {
         let mut constants = FxHashSet::default();
@@ -217,7 +225,7 @@ impl Inner {
             .expect("failed to extend memory layout");
 
         let inner = unsafe {
-            let inner = alloc(layout).cast::<Self>();
+            let inner = try_alloc(layout).cast::<Self>();
 
             // Write the first part, the Inner.
             inner.write(Self {
@@ -257,7 +265,7 @@ impl Inner {
             .expect("failed to extend memory layout");
 
         let inner = unsafe {
-            let inner = alloc(layout).cast::<Self>();
+            let inner = try_alloc(layout).cast::<Self>();
 
             // Write the first part, the Inner.
             inner.write(Self {


Razican (approved):
Thanks!

I'm not super happy to have unsafe code in the engine, though. This is the first bug we get due to this, and this workaround would abort the process, which is also not ideal. I would prefer to be able to handle the error with a `Result::Err` response, for example.

But for now, I think it's good enough. Once the UTF-16 strings get implemented, we might be able to create our own internets and use different ways for doing these things.

y21:
> ... and this workaround would abort the process, which is also not ideal. I would prefer to be able to handle the error with a Result::Err response, for example.

I think the Rust standard library suffers from the same problem. It seems like Rust std is not really "aware" of OOM errors (or allocator errors in general) and methods like `Vec::push` don't return a Result either (even though the call to realloc could fail), and instead just abort.
I agree that this isn't a great permanent solution, but at least it gets rid of the UB in a reachable path for now

RageKnify (approved):
bors r+

PR #1851, opened 2022-02-20
[Merged by Bors] - Added conversions from and to serde_json's Value type

### LINKED ISSUE #1693: Easy conversion from JsValue to serde_json::Value
In my quest to convert a `JsValue` into a Rust struct, a necessary intermediate step seems to be converting a `JsValue` into a `serde_json::Value`. This would make it very easy to then create a Rust struct. For example, this functionality might look like the following: `let my_struct: MyStruct = serde_json::from(jsValue.to_serde_json_value());`.

Without this, I either have to implement my own conversion from `JsValue` to `serde_json::Value`, or I have to `JSON.stringify` my JS value before returning it to the Rust environment, where I can then use `serde_json::from_str` to continue the conversion.

And I'm not sure this is the best way to do it, but somehow allowing `JsValue`'s to be easily converted into Rust structs is what I'm after.

Razican:
This Pull Request closes #1693.

It changes the following:

- It adds a fallible conversion from `serde_json::Value` to `JsValue`, which requires a context.
- It adds a fallible conversion from `JsValue` to `serde_json::Value`, which requires a context.
- Added examples to the documentation of both methods.
- Removed some duplicate and non-needed code that I found while doing this.

### DIFF
--- a/boa_engine/src/object/mod.rs
+++ b/boa_engine/src/object/mod.rs
@@ -456,17 +456,6 @@ impl Object {
         )
     }
 
-    #[inline]
-    pub fn as_array(&self) -> Option<()> {
-        match self.data {
-            ObjectData {
-                kind: ObjectKind::Array,
-                ..
-            } => Some(()),
-            _ => None,
-        }
-    }
-
     /// Checks if it is an `ArrayIterator` object.
     #[inline]
     pub fn is_array_iterator(&self) -> bool {
@@ -791,17 +780,6 @@ impl Object {
                 ..
             }
         )
-    }
-
-    #[inline]
-    pub fn as_error(&self) -> Option<()> {
-        match self.data {
-            ObjectData {
-                kind: ObjectKind::Error,
-                ..
-            } => Some(()),
-            _ => None,
-        }
     }
 
     /// Checks if it a Boolean object.
@@ -1159,7 +1137,10 @@ impl Object {
         &self.properties
     }
 
-    /// Helper function for property insertion.
+    /// Inserts a field in the object `properties` without checking if it's writable.
+    ///
+    /// If a field was already in the object with the same name, then a `Some` is returned
+    /// with that field's value, otherwise, `None` is returned.
     #[inline]
     pub(crate) fn insert<K, P>(&mut self, key: K, property: P) -> Option<PropertyDescriptor>
     where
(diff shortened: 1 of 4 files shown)


boa/src/value/serde_json.rs
```diff
@@ -0,0 +1,157 @@
...
+                Some(s) => Ok(Self::Rational(s as f64)),
+                None => {
+                    if let Some(i) = num.as_u64() {
+                        Ok(Self::Rational(i as f64))
+                    } else {
+                        Ok(Self::Rational(num.as_f64().ok_or_else(|| {
+                            context.construct_type_error(format!(
+                                "could not convert JSON number {num} to JsValue"
+                            ))
+                        })?))
```
jedel1043, line 54: A more functional approach should be more readable, I think.

Something like:
`num.as_i64().filter({range}).map(|i|JsValue::Integer(i as i32)).or_else(||num.as_f64()).map(JsValue::Rational).ok_or_else({error})?`

jedel1043 (changes requested):
Great! This should be pretty useful for users of `serde_json` :D

jedel1043:
Also, I think some module tests would be good to ensure conversions work OK

RageKnify (approved):
Nothing to add besides @jedel1043's notes

Razican:
> Also, I think some module tests would be good to ensure conversions work OK

I wrote the fixes and added a pretty deep test. Hopefully this should solve most situations.

boa/src/value/serde_json.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// let mut context = Context::default();
+    /// let value = JsValue::from_json(&json, &mut context).unwrap();
+    /// #
+    /// # assert_eq!(json, value.to_json(&mut context).unwrap());
+    /// ```
+    pub fn from_json(json: &Value, context: &mut Context) -> JsResult<Self> {
+        /// Biggest possible integer, as i64.
+        const MAX_INT: i64 = i32::MAX as i64;
+
+        /// Biggest possible integer, as i64.
```
jedel1043, line 40:
```suggestion
        /// Smallest possible integer, as i64.
```

jedel1043 (changes requested):

jedel1043 (approved):

PR #1852, opened 2022-02-21
[Merged by Bors] - Add assertion to check that a break label is identified at compile-time

VTCAKAVSMoACE:

- Adds a check at compile time for the existence of a break label (this should be a syntax error in the future; refactor from panics to results in compile should be a separate PR)
- Adds a test for break label existence in boa/tests

262 misses some fairly important JS parity issues and not performing this check eagerly can lead to other more severe issues during VM execution.


Razican (approved):
This looks good to me :) it's important to also get the panic to Result conversion soon, but this at least makes the VM more consistent.

RageKnify (approved):
bors r+

PR #1853, opened 2022-02-21
[Merged by Bors] - Correct pop_on_return behaviour

VTCAKAVSMoACE:

- Updates the value of `pop_on_return` after a catch as to prevent VM stack corruption
- Adds two test cases which demonstrate the issue and demonstrate that it has been fixed

I am unsure if it is possible to abuse the patch provided; one would need to catch from within an array initialisation without calling into another frame (e.g., with a lambda), which I don't think is possible.


RageKnify (changes requested):
Feels weird to not check anything after executing the JS code. We could at least check whether an exception was thrown, in the first case the result of `exec` should be `Err` and in the second case `Ok`, right? Just asserting that they got the right variants feels better than doing nothing else.

VTCAKAVSMoACE:
Sorry for the late response on this; busy Monday. :) Do the changes provided suffice?

jedel1043 (approved):
Looks good!

RageKnify (approved):
bors r+

PR #1855, opened 2022-02-21
[Merged by Bors] - Correct reference error message

aaronmunsters:
This Pull Request fixes/closes the incorrect message thrown for the following code:

```javascript
"use strict";
foo = "bar";
```

Which would throw the following before the change (incorrect):
`Uncaught "ReferenceError": "binding already exists: foo"`

And would throw the following after the change (correct):
`Uncaught "ReferenceError": "assignment to undeclared variable foo"`


Razican (approved):
Thanks for the contribution!!

jedel1043 (approved):
Thank you very much!

PR #1860, opened 2022-02-23
[Merged by Bors] - Prevent breaks without loop or switch from causing panics

VTCAKAVSMoACE:

- Replaces a panic with a syntax error when a break is used outside of a loop or switch
- Adds a test for that


boa_engine/src/tests.rs
```diff
@@ -442,6 +442,15 @@ fn test_invalid_break_target() {
     let _ = &exec(src);
 }
 
+#[test]
+fn test_invalid_break() {
+    let src = r#"
+        break;
+        "#;
+
+    assert!(matches!(Context::default().eval(src.as_bytes()), Err(_)));
```
Razican, line 451: Here too, I would use `is_err()` instead of `matches!()`, it makes things more clear, and you can directly pass a reference to `src` instead of calling `as_bytes()`, if I remember correctly.

Razican (approved):
Just a small recommendation for readability

raskad (approved):

HalidOdat (changes requested):
Hmmm... Shouldn't this be an [_early error_](https://tc39.es/ecma262/#prod-BreakStatement) . According to spec:

> 14.9.1 Static Semantics: Early Errors
> [BreakStatement](https://tc39.es/ecma262/#prod-BreakStatement): break ;
>
>    It is a Syntax Error if this [BreakStatement](https://tc39.es/ecma262/#prod-BreakStatement) is not nested, directly or indirectly (but not crossing function or static initialization block boundaries), within an [IterationStatement](https://tc39.es/ecma262/#prod-IterationStatement) or a [SwitchStatement](https://tc39.es/ecma262/#prod-SwitchStatement).

In which case it should be in the parser, not compilation.

Razican:
> Hmmm... Shouldn't this be an [_early error_](https://tc39.es/ecma262/#prod-BreakStatement) . According to spec:
> 
> > 14.9.1 Static Semantics: Early Errors
> > [BreakStatement](https://tc39.es/ecma262/#prod-BreakStatement): break ;
> > It is a Syntax Error if this [BreakStatement](https://tc39.es/ecma262/#prod-BreakStatement) is not nested, directly or indirectly (but not crossing function or static initialization block boundaries), within an [IterationStatement](https://tc39.es/ecma262/#prod-IterationStatement) or a [SwitchStatement](https://tc39.es/ecma262/#prod-SwitchStatement).
> 
> In which case it should be in the parser, not compilation.

Indeed, this should be in the parser, but that might be a bit more complex to implement, so this could be a quick win to avoid the panic for the 0.14 release.

HalidOdat:
> Indeed, this should be in the parser, but that might be a bit more complex to implement, so this could be a quick win to avoid the panic for the 0.14 release.

In that case we should put `FIXME`s in places where we have the quick fix, so we can deal with them later :)

Razican:
> > Indeed, this should be in the parser, but that might be a bit more complex to implement, so this could be a quick win to avoid the panic for the 0.14 release.
> 
> 
> 
> In that case we should put `FIXME`s in places where we have the quick fix, so we can deal with them later :)

Maybe also to open n issue about it

raskad:
> In which case it should be in the parser, not compilation.

#1829 also introduces syntax errors at compilation time. I'm not sure if it makes sense to push all of these errors to the parser. For some we would have to walk the whole AST, and we do that at compilation time anyways.

VTCAKAVSMoACE:
Are the changes requested still relevant? I'm not sure what I need to change here.

raskad:
@HalidOdat Can you give this another look so we can fix those panics on main?

PR #1861, opened 2022-02-23
[Merged by Bors] - Fix unreachable panics in compile_access

VTCAKAVSMoACE:

- More elegantly handles illegal access statements in compile_access
- Adds a slew of previously unhandled illegal access test cases

### Caveats

It is very, very likely that you will want to simply restrict unary and assignment operations in the AST. However, this prevents crashes in the meantime with a error that is just slightly less detailed than if it were implemented in AST.


Razican (approved):
Looks good to me!

boa_engine/src/tests.rs
```diff
@@ -1357,6 +1405,25 @@ fn multicharacter_bitwise_assignment_to_non_assignable() {
...
+    // Disabled - awaiting implementation.
+    let test_cases = [
+        "(()=>{})() >>>= 5",
+        "(()=>{})() &&= 5",
+        "(()=>{})() ||= 5",
+        "(()=>{})() ??= 5",
+    ];
+
+    for case in &test_cases {
+        let string = dbg!(forward(&mut context, case));
```
jedel1043, line 1421:
```suggestion
        let string = forward(&mut context, case);
```

boa_engine/src/tests.rs
```diff
@@ -1342,6 +1377,19 @@ fn multicharacter_assignment_to_non_assignable() {
...
 }
 
+#[test]
+fn multicharacter_assignment_to_non_assignable_ctd() {
+    let mut context = Context::default();
+
+    let test_cases = ["(()=>{})() **= 5", "(()=>{})() <<= 5", "(()=>{})() >>= 5"];
+
+    for case in &test_cases {
+        let string = dbg!(forward(&mut context, case));
```
jedel1043, line 1387:
```suggestion
        let string = forward(&mut context, case);
```

boa_engine/src/tests.rs
```diff
@@ -485,6 +485,18 @@ fn unary_pre() {
     assert_eq!(&exec(execs_before_dec), "true");
 }
 
+#[test]
+fn invalid_unary_access() {
+    let mut context = Context::default();
+    let test_cases = ["++[];", "[]++;", "--[];", "[]--;"];
+
+    for case in &test_cases {
+        let string = dbg!(forward(&mut context, case));
```
jedel1043, line 494:
```suggestion
        let string = forward(&mut context, case);
```

jedel1043 (changes requested):
Good job! Just some `dbg` leftovers

boa_engine/src/tests.rs
```diff
@@ -1326,6 +1338,29 @@ fn assignment_to_non_assignable() {
...
+        "(()=>{})() |= 5",
+        "(()=>{})() += 5",
+        "(()=>{})() = 5",
+    ];
+
+    for case in &test_cases {
+        let string = forward(&mut context, case);
+
+        assert!(string.starts_with("Uncaught \"SyntaxError\": "));
+    }
```
HalidOdat, line 1361: We have a better way of testing in rust with https://github.com/boa-dev/boa/blob/main/boa_engine/src/lib.rs#L151-L193
Here is an example https://github.com/boa-dev/boa/blob/main/boa_engine/src/tests.rs#L1360-L1368

HalidOdat:
> ### Caveats
> 
> It is very, very likely that you will want to simply restrict unary and assignment operations in the AST. However, this prevents crashes in the meantime with a error that is just slightly less detailed than if it were implemented in AST.

Yup. These are early errors that should be checked during parsing, and the `unreachable!()` should be unreachable if we implement them.

boa_engine/src/tests.rs
VTCAKAVSMoACE: Okay, I'll change that over; I was following the pattern I observed with similar test cases.

boa_engine/src/bytecompiler.rs
```diff
@@ -456,13 +456,13 @@ impl<'b> ByteCompiler<'b> {
...
-            Node::Identifier(name) => Access::Variable { name: name.sym() },
-            Node::GetConstField(node) => Access::ByName { node },
-            Node::GetField(node) => Access::ByValue { node },
-            Node::This => Access::This,
-            _ => unreachable!(),
+            Node::Identifier(name) => Some(Access::Variable { name: name.sym() }),
+            Node::GetConstField(node) => Some(Access::ByName { node }),
+            Node::GetField(node) => Some(Access::ByValue { node }),
+            Node::This => Some(Access::This),
+            _ => None,
```
jedel1043, line 465: This Clippy warning was already fixed, but the merge undid the fix

jedel1043 (changes requested):

boa_engine/src/bytecompiler.rs
VTCAKAVSMoACE: Ah hell, sorry. Fixing.

jedel1043 (approved):
Looks good to me!

PR #1866, opened 2022-02-23
[Merged by Bors] - Removed reference counted pointers from `JsValue` variants

### LINKED ISSUE #1864: Enable a parsed StatementList to be passed across async/await boundaries
Hi,

I'm trying to implement Boa within a Rust web server (running behind [Warp](https://github.com/seanmonstar/warp)), which has the requirement that all async code must also be Send as it *may* get processed on an alternative Tokio thread. Currently the one thing preventing a parsed [StatementList](https://boa-dev.github.io/boa/doc/boa_engine/syntax/ast/node/statement_list/struct.StatementList.html) from being sent across threads is the Rc<RawBigInt> in [src/bigint.rs](https://github.com/boa-dev/boa/blob/8c92a8fa25c1339b639a5acc5c3133ae8df670a2/boa_engine/src/bigint.rs#L24).

https://github.com/boa-dev/boa/blob/8c92a8fa25c1339b639a5acc5c3133ae8df670a2/boa_engine/src/bigint.rs#L23-L25

Removing the Rc around the RawBigInt allows the async code to build, and I don't think that it negatively affects performance when [operating on RawBigInt values](https://github.com/boa-dev/boa/blob/8c92a8fa25c1339b639a5acc5c3133ae8df670a2/boa_engine/src/bigint.rs#L216-L254) together.

Just to make it clear, this requirement **DOES NOT** have anything to do with evaluating a Boa Context in an asynchronous way, and **DOES NOT** have anything to do with async/await within the Boa Javascript runtime. Instead this would allow a parsed StatementList from being passed across async/await boundaries or across threads.

I have implemented this small change here: https://github.com/tobiemh/boa/commit/7cdc0487645f063176b261bf40904f97931b2460, but I wanted to hear thoughts on this proposal before submitting any pull request (if desired).

Razican:
This Pull Request fixes/closes #1864.

It changes the following:

- Removed `JsBigInt` from `Const` nodes, using a boxed `BigInt` instead.
- Modifies the `JsObject` variant so that it has a similar structure to other variants, where the internal structure is private.

The size of `JsValue` stays in 2 64-bit words (in a 64-bit system at least), and the size of `Const` also stays the same.

I have noticed that we clone tokens too much in the parser, so I was thinking that we should implement a by-value getter for `kind()`. Something like `kind_unwrap()`.

### DIFF
--- a/boa_engine/src/bytecompiler.rs
+++ b/boa_engine/src/bytecompiler.rs
@@ -554,7 +554,7 @@ impl<'b> ByteCompiler<'b> {
                     )),
                     Const::Int(v) => self.emit_push_integer(*v),
                     Const::Num(v) => self.emit_push_rational(*v),
-                    Const::BigInt(v) => self.emit_push_literal(Literal::BigInt(v.clone())),
+                    Const::BigInt(v) => self.emit_push_literal(Literal::BigInt(v.clone().into())),
                     Const::Bool(true) => self.emit(Opcode::PushTrue, &[]),
                     Const::Bool(false) => self.emit(Opcode::PushFalse, &[]),
                     Const::Null => self.emit(Opcode::PushNull, &[]),
--- a/boa_engine/src/bigint.rs
+++ b/boa_engine/src/bigint.rs
@@ -3,8 +3,7 @@
 use crate::{builtins::Number, Context, JsValue};
 use boa_gc::{unsafe_empty_trace, Finalize, Trace};
 use num_integer::Integer;
-use num_traits::pow::Pow;
-use num_traits::{FromPrimitive, One, ToPrimitive, Zero};
+use num_traits::{pow::Pow, FromPrimitive, One, ToPrimitive, Zero};
 use std::{
     fmt::{self, Display},
     ops::{Add, BitAnd, BitOr, BitXor, Div, Mul, Neg, Rem, Shl, Shr, Sub},
@@ -286,6 +285,15 @@ impl From<RawBigInt> for JsBigInt {
     fn from(value: RawBigInt) -> Self {
         Self {
             inner: Rc::new(value),
+        }
+    }
+}
+
+impl From<Box<RawBigInt>> for JsBigInt {
+    #[inline]
+    fn from(value: Box<RawBigInt>) -> Self {
+        Self {
+            inner: value.into(),
         }
     }
 }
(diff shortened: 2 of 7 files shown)


Razican:
Actually, I think this will reduce the performance when doing clones of `JsValue`, I think I can get a better solution for `StatementList`.

boa_engine/src/bytecompiler.rs
```diff
@@ -554,7 +554,7 @@ impl<'b> ByteCompiler<'b> {
                     )),
                     Const::Int(v) => self.emit_push_integer(*v),
                     Const::Num(v) => self.emit_push_rational(*v),
-                    Const::BigInt(v) => self.emit_push_literal(Literal::BigInt(v.clone())),
+                    Const::BigInt(v) => self.emit_push_literal(Literal::BigInt(v.clone().into())),
```
Razican, line 557: This is the only place where I see that it will actually be less performant.

jedel1043:
> Wow performance seems pretty bad

I'm curious if the cause was moving the internal `Gc<GcCell<Object>>` to `inner`.

raskad:
> I'm curious if the cause was moving the internal `Gc<GcCell<Object>>` to `inner`.

I don't see how this could have an impact. From the unsafe code guide:
`In terms of their layout, tuple structs can be understood as equivalent to a named struct with fields named 0..n-1`
https://rust-lang.github.io/unsafe-code-guidelines/layout/structs-and-tuples.html

Other than the layout is there anything that could be affected by the struct definition?

Razican:
> > I'm curious if the cause was moving the internal `Gc<GcCell<Object>>` to `inner`.
> 
> I don't see how this could have an impact. From the unsafe code guide: `In terms of their layout, tuple structs can be understood as equivalent to a named struct with fields named 0..n-1` https://rust-lang.github.io/unsafe-code-guidelines/layout/structs-and-tuples.html#struct-types
> 
> Other than the layout is there anything that could be affected by the struct definition?

Nope, that structure should be exactly the same. But the issue is that calling `clone()` on a `Box` is much more expensive than calling it on `Rc`. So, we should find where we are using it. I will try to do that by removing the `Clone` derive temporarily to see what happens.

Something I also noticed is that we clone tokens too much. We should have a `unwrap_kind()` function to extract the kind by value instead of by reference, and get ownership of it. I will also add that.

HalidOdat (approved):
Looks good from my side :)

raskad (approved):

PR #1870, opened 2022-02-26
[Merged by Bors] - Feature arrays with empty elements

HalidOdat:
This PR adds support for arrays with empty elements e.g. `["", false, , , ]`. Before we were filling the empty places with `undefined`, but this is wrong according to [spec](https://tc39.es/ecma262/#sec-runtime-semantics-arrayaccumulation) there shouldn't be undefined with a index at that place, instead only `length` is incremented. So `[,,,].length == 3` and operations like `[,,,,].indexOf(undefined) == -1`,  `[,,,,].lastIndexOf(undefined) == -1` etc.

### DIFF
--- a/boa_engine/src/bytecompiler.rs
+++ b/boa_engine/src/bytecompiler.rs
@@ -933,6 +933,11 @@ impl<'b> ByteCompiler<'b> {
                 self.emit_opcode(Opcode::PopOnReturnAdd);
 
                 for element in array.as_ref() {
+                    if let Node::Empty = element {
+                        self.emit_opcode(Opcode::PushElisionToArray);
+                        continue;
+                    }
+
                     self.compile_expr(element, true)?;
                     if let Node::Spread(_) = element {
                         self.emit_opcode(Opcode::InitIterator);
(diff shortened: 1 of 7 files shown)


raskad:
I would switch the naming from `Hole` to `Elision` to be aligned with the spec.

boa_engine/src/bytecompiler.rs
```diff
@@ -559,6 +559,7 @@ impl<'b> ByteCompiler<'b> {
                     Const::Bool(false) => self.emit(Opcode::PushFalse, &[]),
                     Const::Null => self.emit(Opcode::PushNull, &[]),
                     Const::Undefined => self.emit(Opcode::PushUndefined, &[]),
+                    Const::Elision => unreachable!(),
```
raskad, line 562: Do you think it makes sense to use `Node::Empty` as the elision node in the arraydecl to avoid this?

HalidOdat: > Do you think it makes sense to use `Node::Empty` as the elision node in the arraydecl to avoid this?

Hmmm... Yeah, `Node::Empty` is a far better choice.

raskad (approved):

Razican (approved):
Looks very good! thanks! :)

PR #1875, opened 2022-02-27
[Merged by Bors] - Implement `Object.prototype.toLocaleString()`

HalidOdat:
It changes the following:
- Implement `Object.prototype.toLocaleString()`

### DIFF
--- a/boa_engine/src/builtins/object/mod.rs
+++ b/boa_engine/src/builtins/object/mod.rs
@@ -57,7 +57,7 @@ impl BuiltIn for Object {
         .method(Self::has_own_property, "hasOwnProperty", 1)
         .method(Self::property_is_enumerable, "propertyIsEnumerable", 1)
         .method(Self::to_string, "toString", 0)
-        .method(Self::to_string, "toLocaleString", 0)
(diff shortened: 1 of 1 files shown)


jedel1043 (approved):

boa_engine/src/builtins/object/mod.rs
```diff
@@ -57,6 +57,7 @@ impl BuiltIn for Object {
         .method(Self::has_own_property, "hasOwnProperty", 1)
         .method(Self::property_is_enumerable, "propertyIsEnumerable", 1)
         .method(Self::to_string, "toString", 0)
+        .method(Self::to_string, "toLocaleString", 0)
```
Razican, line 60: Shouldn't this be `Self::to_locale_string`?

Razican (changes requested):
I think the mere is a typo when setting the method

boa_engine/src/builtins/object/mod.rs
HalidOdat: Yes, copy & paste mistake :sweat_smile:

Razican (approved):

PR #1877, opened 2022-03-01
[Merged by Bors] - Fix `Number.prototype.toString` when passing `undefined` as radix

HalidOdat:
This PR fixes the bug when passing `undefined` as radix in `Number.prototype.toString()` which was throwing a range error, instead of setting the radix to `10`

[spec](https://tc39.es/ecma262/#sec-number.prototype.tostring):
> **Note**
>
> The optional radix should be an [integral Number](https://tc39.es/ecma262/#integral-number) value in the inclusive range 2𝔽 to 36𝔽. If radix is undefined then 10𝔽 is used as the value of radix.

### DIFF
--- a/boa_engine/src/builtins/number/mod.rs
+++ b/boa_engine/src/builtins/number/mod.rs
@@ -643,21 +643,23 @@ impl Number {
         let x = Self::this_number_value(this, context)?;
 
         // 2. If radix is undefined, let radixNumber be 10.
+        let radix = args.get_or_undefined(0);
+        let radix_number = if radix.is_undefined() {
+            10.0
         // 3. Else, let radixNumber be ? ToInteger(radix).
-        let radix = args
-            .get(0)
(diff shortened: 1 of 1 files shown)


Razican (approved):
Looks good to me. I guess there shouldn't be any penalty on comparing floats in the range, instead of integers, right?

HalidOdat:
> I guess there shouldn't be any penalty on comparing floats in the range, instead of integers, right?

It shouldn't be a perf. penalty since it is equal to `f >= 2.0 && f <= 36.0`. And we have to do this anyway because `.to_integer()` returns a `f64` which is bigger than `u8` (could be even bigger than `u64`) and we where casting which could pass the condition because loss of precision from `f64 => u8` and may give the wrong output.

Razican (approved):

raskad (approved):

PR #1883, opened 2022-03-01
`%NativeError%.[[prototype]]` should be `Error` constructor

HalidOdat:
Before the `%NativeError%` objects (like `TypeError`, `ReferenceError`, etc) `[[prototype]]` field was set to `Function.prototype` but this is wrong it should be the `Error` constructor object itself.


raskad (approved):

RageKnify (approved):
Nice, I'd noticed these tests were failing but couldn't understand the whole prototype vs constructor logic.
bors r+

PR #1886, opened 2022-03-02
[Merged by Bors] - Fix `PropertyKey` to `JsValue` conversion

HalidOdat:
We store string `PropertyKey`s with two enums `String` and `Index` for performance reasons, but the spec does not differentiate between string and index property keys so before conversion to `JsValue` we have to convert to a string.

This was failing tests like `Reflect.ownKeys([true, "", 1])` because it was returning (integer numbers) `[1, 2, 3]` instead of `['1', '2', '3']`


raskad (approved):

jedel1043 (approved):
Looks good!

PR #1887, opened 2022-03-02
[Merged by Bors] - Fix `BigInt` and `Number` comparison

HalidOdat:
Fixes `BigInt` and `Number` comparison, and vice versa. Before we were removing the decimal point of the floating-point number which was causing cases like `0.000001 > 0n` (or `0n < 0.000001`) to fail.

### DIFF
--- a/boa_engine/src/value/operations.rs
+++ b/boa_engine/src/value/operations.rs
@@ -538,7 +538,7 @@ impl JsValue {
                             }
 
                             if let Ok(y) = JsBigInt::try_from(y) {
-                            	return Ok((*x < y).into());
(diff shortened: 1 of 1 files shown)


RageKnify (approved):

Razican:
@HalidOdat do you think this could be ready during the weekend, or should we move it to 0.15?

HalidOdat:
> @HalidOdat do you think this could be ready during the weekend, or should we move it to 0.15?

I will try to fix this during this weekend :)

RageKnify (approved):

Razican (approved):

PR #1890, opened 2022-03-03
[Merged by Bors] - Extract `Intrinsics` struct from `Context` and cleanup names

jedel1043:
Building up to #186, this PR extracts an `Intrinsics` struct from `Context`, facilitating a lot the extraction of a `Realm` struct.
Also, it adapts the `BuiltIn` trait to be useful for builtins that don't expose a global property on initialization (`Generator`, `TypedArray`, etc.)

It changes the following:

- Creates an `Intrinsics` struct and refactors `Context` to transfer its intrinsic related fields to `Intrinsics`.
- Renames some methods and parameters to better describe their functionality.
- Makes `BuiltIn::init` return `Option<JsValue>` to skip global property initialization if the builtin initialization returns `None`


boa_engine/Cargo.toml
```diff
@@ -57,7 +57,7 @@ fast-float = "0.2.0"
 unicode-normalization = "0.1.19"
 dyn-clone = "1.0.4"
 once_cell = "1.9.0"
-
+tap = "1.0.1"
```
HalidOdat, line 60: I don't think we really need this dependency, Since all we use it is to convert a value to a `Option`

boa_engine/src/context/intrinsics.rs
```diff
@@ -0,0 +1,368 @@
...
+impl Intrinsics {
+    /// Return the cached intrinsic objects.
+    #[inline]
+    pub fn objects(&self) -> &IntrinsicObjects {
+        &self.objects
+    }
+
+    /// Return the core standard objects.
+    #[inline]
+    pub fn standard_constructors(&self) -> &StandardConstructors {
```
HalidOdat, line 24:
```suggestion
    pub fn constructors(&self) -> &Constructors {
```
Since we have to call `.intrinsics()` before we can get the standard constructors, I think it's a bit redundant to call it `standard_constructors()` now.
```rust
context
    .intrinsics()
    .constructos()
    .type_error()
    .prototype()
```

boa_engine/src/value/conversions.rs
```diff
@@ -134,15 +134,15 @@ impl From<()> for JsValue {
     }
 }
 
-impl<T> From<Option<T>> for JsValue
+pub(crate) trait IntoOrUndefined {
+    fn into_or_undefined(self) -> JsValue;
+}
```
HalidOdat, line 139: Why do we have this trait? Is `From<Option<T>>` not sufficient?

HalidOdat (commented):
Looks good!

check my comment on how we might improve this :)

boa_engine/src/value/conversions.rs
jedel1043: Thought about it and realized `From<Option<T>>` was a bit risky to have.

When I finished refactoring the return values for the `init` function of the builtins, I waited for the error messages from the compiler. However, to my surprise, the `From<Option<T>>` impl was implicitly converting the `None` returns into `JsValue::Undefined`, and I realized having that implicit conversion could cause pretty nasty bugs.
So, I essentially made it opt-in with this trait, which clearly says in the code "yeah, I really want to convert `None` into `JsValue::Undefined`"

boa_engine/Cargo.toml
jedel1043: I think it adds clarity to the pipeline associated with the builtin constructors.
```rust
ConstructorBuilder::new()
...
... several lines of operations
...
    .build()
    .conv::<JsValue>()
    .pipe(Ok)
```
is, for me, a lot clearer than

```rust
let object = ConstructorBuilder::new()
...
... several lines of operations
...
        .build();
Ok(object.into())
```

Also, the dependency is pretty lightweight, is widely used in the Rust community (7M downloads) and it doesn't use any other dependency, which usually is our main reason against adding a dependency.

boa_engine/src/context/intrinsics.rs
jedel1043: Fixed

HalidOdat (approved):

boa_engine/src/context/intrinsics.rs
```diff
@@ -0,0 +1,368 @@
+use crate::{
+    builtins::{error::r#type::create_throw_type_error, iterable::IteratorPrototypes},
+    object::{JsObject, ObjectData},
+    Context,
+};
+
+#[derive(Debug, Default)]
+pub struct Intrinsics {
```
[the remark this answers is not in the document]
jedel1043: It's forced, since the `intrinsics` method of `Context` is public and we cannot return a non-public type from a public method.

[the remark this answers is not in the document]
raskad: True, I was more thinking if we want to expose the `Context::intrinsics` method at all. Currently we expose the `Context::standard_objects`, but is that actually useful outside of the builtins? But it's probably best to discuss this together with other public API topics.

raskad (approved):

boa_engine/src/context/intrinsics.rs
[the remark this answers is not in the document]
jedel1043: I think exposing it is necessary since this is the only way our users could inherit from builtin prototypes with ease.

What we could do is hide every other intrinsic and just leave the constructors public.

PR #1891, opened 2022-03-03
Avoid returning clones of pointer types on `JsValue` methods

jedel1043:
This Pull Request creates a new `Ref` struct (design by @HalidOdat) that represents a reference for boxed pointer types (`JsValue`, `JsString`, `JsSymbol`, `JsBigInt`), skipping unnecessary clones.

It changes the following:

- Creates a new `Ref` struct representing a reference to a pointer value.
- Makes methods of `JsValue` return `Ref<{PointerValue}>` when possible. 
- Refactors the codebase to accomodate to the new API.

### DIFF
--- a/boa_engine/src/value/mod.rs
+++ b/boa_engine/src/value/mod.rs
@@ -573,6 +573,16 @@ pub(crate) unsafe trait PointerType {
     unsafe fn into_void_ptr(ty: ManuallyDrop<Self>) -> *mut ();
 }
 
+/// Represents a reference to a boxed pointer type inside a [`JsValue`]
+///
+/// This is exclusively used to return references to [`JsString`], [`JsObject`],
+/// [`JsSymbol`] and [`JsBigInt`], since NaN boxing makes returning proper references
+/// difficult. It is mainly returned by the [`JsValue::variant`] method and the
+/// `as_` methods for checked casts to pointer types.
+///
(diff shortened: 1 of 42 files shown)


boa_engine/src/value/mod.rs
```diff
@@ -681,6 +733,8 @@ mod tests_nan_box {
 
         let bigint = value.as_bigint().unwrap();
 
+        println!("pass!");
```
Razican, line 736: Is this a leftover comment?

jedel1043: Yeah, sorry 😅

boa_engine/src/value/mod.rs
```diff
@@ -575,24 +573,61 @@ pub(crate) unsafe trait PointerType {
...
+    fn deref(&self) -> &Self::Target {
+        &*self.inner
+    }
+}
+
+impl<T: PartialEq> PartialEq<T> for Ref<'_, T> {
+    fn eq(&self, other: &T) -> bool {
+        &*self.inner == other
+    }
+}
```
HalidOdat, line 627: We should add `#[inline]` so they can be inlined in external crates

HalidOdat (approved):
This looks good! :)

Great work @jedel1043

HalidOdat:
Lets merge this into `nan-boxing` :)

PR #1898, opened 2022-03-06
Split examples into separate executables

jedel1043:
Just a general cleanup of the examples, since it is more convenient to be able to independently run any example with `cargo run --bin`.


RageKnify:
This seems to compete with #1161

jedel1043:
> This seems to compete with #1161 

I opened the PR against examples, not main

RageKnify:
> I opened the PR against examples, not main

Sorry, my bad.

RageKnify (approved):

jasonwilliams (approved):

PR #1950, opened 2022-03-17
tests web view: Change individual test links to be <a> instead of <div>, for better accessibility

### LINKED ISSUE #1949: GH Pages Conformance Suite: links to individual tests aren't links
This is a bug in boa's test262 website, not in boa itself.

**Describe the bug**
Individual tests at https://boa-dev.github.io/boa/test262/ are described a small card, specifying whether the test passed. When you click on one of these, the corresponding test is opened in a new tab. However, the cards are not links, so standard affordances like the ability to copy the destination URL aren't available.

**To Reproduce**
Visit https://boa-dev.github.io/boa/test262/ and try to Right Click > Copy Link Address on one of the test cards. It doesn't work, since it is not a link, despite acting like one :(

**Expected behavior**
The cards should be semantic links.

N/A (structure of dynamic HTML in gh-pages branch)

Nathan-Fenner:
This Pull Request fixes/closes #1949

It changes the following:

- Replace the testCard `<div />` with a `<a />`
- Replace `.click()` callback with an `href` attribute, and specify `target="_blank"` so that the link opens in a new tab by default

### DIFF
--- a/test262/results.js
+++ b/test262/results.js
@@ -295,12 +295,13 @@
           }
 
           let testCard = $(
-            `<div title="${innerTest.n}" class="card test embed-responsive ${style}"></div>`
-          ).click(() => {
-            window.open(
+            `<a title="${innerTest.n}" class="card test embed-responsive ${style}"></a>`
+          )
+            .attr(
+              "href",
               `https://github.com/tc39/test262/blob/${upstream}/${name}`
-            );
-          });
+            )
+            .attr("target", "_blank");
 
           if (innerTest.r === "P") {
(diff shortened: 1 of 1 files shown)


jedel1043 (approved):
Thank you for your contribution!
We've really leaving aside the accessibility of our webpage a bit to focus on the development of the engine, so any improvement on that aspect is very much appreciated 😁

Razican (approved):
Thank you! Yes, it's much better like this :) don't know why I did it with divs on the first place

PR #1952, opened 2022-03-17
[Merged by Bors] - add README for crates.io publish

### LINKED ISSUE #1948: Add the README to the crate in crates.io
Currently, we don't have the README linked in crates.io:
https://crates.io/crates/boa_engine

We should add it there, to give a bit more information about the crate. This might require adding the README to the package (we are excluding some stuff, so this might be some of it)

superhawk610:
This Pull Request closes #1948.

It changes the following:

- set `readme` in `boa_engine` so `README.md` will be published to crates.io
- remove unnecessary `exclude` field from `Cargo.toml` in all apps

I was unsure whether using a path outside of the workspace root was allowed for `readme` since it [doesn't get included in the release tarball](https://github.com/rust-lang/cargo/issues/5911), but this exact path is used by [juniper](https://github.com/graphql-rust/juniper/blob/master/juniper/Cargo.toml#L13) and [seems to work there](https://crates.io/crates/juniper). I believe `cargo publish` does a bit more than just uploading the tarball, including pulling the `readme` from any arbitrary path.

The default behaviour of `cargo package`/`cargo publish` if neither `exclude` or `include` is specified is to include all files from the package root, excluding

- dotfiles
- .gitignore'd files
- subpackages (any subdirectory with a `Cargo.toml` file)
- the `/target` directory

There's no need to explicitly exclude files from the parent directory since they're already excluded by default. This can be verified by running `cargo package --list` inside any workspace app:

```plain
$ cd boa_wasm
$ cargo package --list
.gitignore
Cargo.toml
Cargo.toml.orig
src/lib.rs
```

You can read more [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields).

### DIFF
--- a/boa_cli/Cargo.toml
+++ b/boa_cli/Cargo.toml
@@ -10,24 +10,6 @@ keywords = ["javascript", "compiler", "lexer", "parser", "js", "cli"]
 categories = ["command-line-utilities"]
 license = "Unlicense/MIT"
 default-run = "boa"
-exclude = [
-    "../.vscode/*",
-    "../.editorconfig",
-    "../test262/*",
-    "../node_modules/*",
-    "../target/*",
-    "../dist/*",
-    "../.github/*",
-    "../assets/*",
-    "../docs/*",
-    "../*.js",
-    "../test_ignore.txt",
-    "../yarn.lock",
-    "../package.json",
-    "../index.html",
-    "../tests/*",
-    "../.github/*",
-]
 
 [dependencies]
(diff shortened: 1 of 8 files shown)


Razican (approved):
Looks very good!! And thanks for all the clarifications :)

RageKnify (approved):
bors r+

HalidOdat (approved):

jedel1043:
bors +r

PR #1953, opened 2022-03-17
[Merged by Bors] - convert inner datetime to local in `to_date_string`

### LINKED ISSUE #1942: `date_proto_to_date_string` can fail due to timezones
**Describe the bug**
The `boa_engine` test `builtins::date::tests::date_proto_to_date_string` seems to assume a system UTC time-zone.

**To Reproduce**

On my system, set to New Zealand standard time (UTC+12:00):

```
$ cargo test -p boa_engine date_proto_to_
    Finished test [optimized + debuginfo] target(s) in 0.15s
     Running unittests (target\debug\deps\boa_engine-30e09a87720b59e0.exe)

running 7 tests
test builtins::date::tests::date_proto_to_json ... ok
test builtins::date::tests::date_proto_to_iso_string ... ok
test builtins::date::tests::date_proto_to_utc_string ... ok
test builtins::date::tests::date_proto_to_gmt_string ... ok
test builtins::date::tests::date_proto_to_string ... ok
test builtins::date::tests::date_proto_to_time_string ... ok
test builtins::date::tests::date_proto_to_date_string ... FAILED

failures:

---- builtins::date::tests::date_proto_to_date_string stdout ----
thread 'builtins::date::tests::date_proto_to_date_string' panicked at 'assertion failed: `(left == right)`
  left: `String("Wed Jul 08 2020")`,
 right: `String("Tue Jul 07 2020")`', boa_engine\src\builtins\date\tests.rs:1149:5


failures:
    builtins::date::tests::date_proto_to_date_string

test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 956 filtered out; finished in 0.01s

error: test failed, to rerun pass '-p boa_engine --lib'
```

**Expected behavior**

Tests pass in any time-zone, I guess.

- OS: Windows
- Version: 11
- Rustc version: rustc 1.59.0 (9d1b2106e 2022-02-23)

**Additional context**
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters) says:
> The parameter values are all evaluated against the local time zone, rather than UTC.

I suspect this is the issue, given that the returned time is different, despite the date values being the same as the `Date.UTC(...)` tests?

superhawk610:
This Pull Request fixes/closes #1942.

`Date.prototype.toDateString` should return a value representing the local date. The Rust `Date` inner value represents UTC time, so it should be adjusted to local time before formatting (see equivalent conversions performed by `to_string` and `to_time_string`).

To verify this is working as intended, run the test suite with your OS timezone set to `GMT+0`, then again with `GMT+10`. The test `date_proto_to_date_string` should pass for each. For me (Ubuntu via WSL), this can be done with `sudo dpkg-reconfigure tzdata`.

This PR also fixes a couple other test cases that used the wrong month value (as noted at the top of the file, JS months are 0-based while `chrono` months are 1-based).

### DIFF
--- a/boa_engine/src/builtins/date/mod.rs
+++ b/boa_engine/src/builtins/date/mod.rs
@@ -1634,7 +1634,12 @@ impl Date {
         // 4. Let t be LocalTime(tv).
         // 5. Return DateString(t).
         if let Some(t) = tv.0 {
-            Ok(t.format("%a %b %d %Y").to_string().into())
+            Ok(Local::now()
+                .timezone()
+                .from_utc_datetime(&t)
+                .format("%a %b %d %Y")
+                .to_string()
+                .into())
         } else {
             Ok(JsString::from("Invalid Date").into())
         }
--- a/boa_engine/src/builtins/date/tests.rs
+++ b/boa_engine/src/builtins/date/tests.rs
@@ -1199,7 +1199,7 @@ fn date_proto_to_string() {
         Some(JsValue::new(
             Local
                 .from_local_datetime(&NaiveDateTime::new(
-                    NaiveDate::from_ymd(2020, 6, 8),
+                    NaiveDate::from_ymd(2020, 7, 8),
                     NaiveTime::from_hms_milli(9, 16, 15, 779)
                 ))
                 .earliest()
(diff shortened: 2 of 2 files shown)


Razican (approved):
Thank you for the contribution! Indeed this makes much more sense :)

HalidOdat (approved):
Looks good to me!

RageKnify (approved):
bors r+
(there seems to be some issues with CI so this will probably take a bit)

Issue #773, opened 2020-10-02
Catching lexer/parser bugs with fuzzed input


HalidOdat:
The fuzzer would generate tests with fuzzed input and we would check what is the expected behaviour

alexanderkjall:
I uploaded the fuzzer I used here: https://github.com/alexanderkjall/boa-hongg feel free to use it however you want.

I used honggfuzz, as it's a bit better than cargo-fuzz, and have run it overnight so far.

I In addition to the bugs I have reported with it, there is also a SIGSEGV due to running out of stackspace when parsing a very nested expression, but I felt that might be out of scope?

neeldug:
It would be nice to have some automation of fuzzing going on, however, as fuzzers last indefinitely, unsure of whether using CI is applicable... I also found AFL which seems simple enough to use and can use an input to converge on bugs quicker, I’m unsure of whether honggfuzz has the same capabilities, but it might be useful to use AFL if not!

jedel1043:
Closed by #2400

PR #1954, opened 2022-03-17
Add an AST visitor/walker

VTCAKAVSMoACE:
This pull request seeks to develop an AST visitor/walker so that we can simplify a lot of the context-sensitive parsing.

### DIFF
--- a/boa_engine/src/syntax/ast/node/mod.rs
+++ b/boa_engine/src/syntax/ast/node/mod.rs
@@ -22,6 +22,7 @@ pub mod switch;
 pub mod template;
 pub mod throw;
 pub mod try_node;
+pub mod visit;
 pub mod r#yield;
 
 pub use self::{
--- a/boa_engine/src/syntax/ast/node/parameters.rs
+++ b/boa_engine/src/syntax/ast/node/parameters.rs
@@ -130,7 +130,7 @@ bitflags! {
     /// Flags for a [`FormalParameterList`].
     #[allow(clippy::unsafe_derive_deserialize)]
     #[cfg_attr(feature = "deser", derive(Serialize, Deserialize))]
-    pub(crate) struct FormalParameterListFlags: u8 {
+    pub struct FormalParameterListFlags: u8 {
         const IS_SIMPLE = 0b0000_0001;
         const HAS_DUPLICATES = 0b0000_0010;
         const HAS_REST_PARAMETER = 0b0000_0100;
--- a/boa_engine/src/syntax/ast/node/parameters.rs
+++ b/boa_engine/src/syntax/ast/node/parameters.rs
@@ -163,8 +163,8 @@ impl Default for FormalParameterListFlags {
 #[cfg_attr(feature = "deser", derive(Serialize, Deserialize))]
 #[derive(Clone, Debug, PartialEq)]
 pub struct FormalParameter {
-    declaration: Declaration,
-    is_rest_param: bool,
+    pub(crate) declaration: Declaration,
+    pub(crate) is_rest_param: bool,
 }
 
 impl FormalParameter {
--- a/boa_engine/src/syntax/ast/node/visit.rs
+++ b/boa_engine/src/syntax/ast/node/visit.rs
@@ -19,6 +19,58 @@ use crate::syntax::ast::node::{
 use crate::syntax::ast::{op, Const, Node};
 use boa_interner::Sym;
 
+/// `Visitor`s "walk" the AST at certain nodes. Useful for when you need to modify source code, but at
+/// the API level.
+///
+/// `Visitor`s have default implementations which are "reasonable", which is to say that users merely
+/// need to implement the visitor for the specific type they wish to visit. For example, let's say
+/// we want to replace all variables named "y" with a variable named "z".
+///
+/// To do so, we implement `Visitor` over a `NameReplacer` struct, then use `NameReplacer` to, well,
+/// replace the names in a parsed source.
+///
+/// ```
+/// use boa_engine::syntax::ast::node::visit::Visitor;
+/// use boa_engine::Context;
+/// use boa_interner::Sym;
+///
+/// struct NameReplacer {
+///     find: Sym,
+///     replace: Sym,
+/// }
+/// impl<'ast> Visitor<'ast> for NameReplacer {
+///     fn visit_sym_mut(&mut self, n: &'ast mut Sym) {
+///         if *n == self.find {
+///             *n = self.replace;
+///         }
+///     }
+/// }
+///
+/// // -- snip -- //
+/// # fn main() {
+/// let source = r#"
+/// let x = 5;
+/// let y = 6;
+/// x + y;
+/// "#;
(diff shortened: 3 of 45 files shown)


Razican:
I guess #773 would be solved by this, right?

raskad (commented):
There is no good way to avoid making all of the node struct field `pub(crate)` with the visitor architecture right? I assume we would need getters and setters for all of them for this to make sense? Not that I'm against `pub(crate)`, but I think we should at least discuss it.

Other than that, I think some documentation on the `Visitor` trait would be good, so that to usage is obvious.

jedel1043:
> There is no good way to avoid making all of the node struct field `pub(crate)` with the visitor architecture right? I assume we would need getters and setters for all of them for this to make sense? Not that I'm against `pub(crate)`, but I think we should at least discuss it.

I don't think it's worth the massive amount of boilerplate that change would add to the AST. Usually getters and setters are used when:
- you need to ensure non-breaking changes for the users (which we don't, since the AST is currently not exposed to our users) or
- you need to maintain certain invariants that the type system cannot maintain by itself (and I think there are very few AST types that have that requirement). We could instead find those types, use setters and getters for them, and default to `pub(crate)` for everything else.

jedel1043 (changes requested):
Looks really good! Documenting how to use the trait would be good though.
Also, did you measure the performance of the fuzzer with this?

VTCAKAVSMoACE:
@jedel1043 please confirm requested changes :)

boa_engine/src/syntax/ast/node/parameters.rs
```diff
@@ -130,7 +130,7 @@ bitflags! {
     /// Flags for a [`FormalParameterList`].
     #[allow(clippy::unsafe_derive_deserialize)]
     #[cfg_attr(feature = "deser", derive(Serialize, Deserialize))]
-    pub(crate) struct FormalParameterListFlags: u8 {
+    pub struct FormalParameterListFlags: u8 {
```
Razican, line 133: Why do we need to expose this?

boa_engine/src/syntax/ast/node/mod.rs
```diff
@@ -22,6 +22,7 @@ pub mod switch;
 pub mod template;
 pub mod throw;
 pub mod try_node;
+pub mod visit;
```
Razican, line 25: Could we add a feature flag for this? I don't think we need it in normal operation, right? and we would avoid compiling it for most of the cases. We would need to add the feature to CI, though, to make sure we still test it.

boa_engine/src/syntax/ast/node/visit.rs
```diff
@@ -0,0 +1,2317 @@
...
+/// `Visitor`s have default implementations which are "reasonable", which is to say that users merely
+/// need to implement the visitor for the specific type they wish to visit. For example, let's say
+/// we want to replace all variables named "y" with a variable named "z".
+///
+/// To do so, we implement `Visitor` over a `NameReplacer` struct, then use `NameReplacer` to, well,
+/// replace the names in a parsed source.
+///
+/// ```
+/// use boa_engine::syntax::ast::node::visit::Visitor;
+/// use boa_engine::Context;
```
Razican, line 34: We usually prefer scoped imports:
```suggestion
/// use boa_engine::{Context, syntax::ast::node::visit::Visitor};
```

boa_engine/src/syntax/ast/node/visit.rs
```diff
@@ -0,0 +1,2317 @@
...
+/// use boa_interner::Sym;
+///
+/// struct NameReplacer {
+///     find: Sym,
+///     replace: Sym,
+/// }
+/// impl<'ast> Visitor<'ast> for NameReplacer {
```
Razican, line 41:
```suggestion
///
/// impl<'ast> Visitor<'ast> for NameReplacer {
```

boa_engine/src/syntax/ast/node/visit.rs
```diff
@@ -0,0 +1,2317 @@
...
+///
+/// To do so, we implement `Visitor` over a `NameReplacer` struct, then use `NameReplacer` to, well,
+/// replace the names in a parsed source.
+///
+/// ```
+/// use boa_engine::syntax::ast::node::visit::Visitor;
+/// use boa_engine::Context;
+/// use boa_interner::Sym;
+///
+/// struct NameReplacer {
```
Razican, line 37: Even if this is just an example, I would still derive `Debug`, `Clone` and `Copy`.

boa_engine/src/syntax/ast/node/visit.rs
```diff
@@ -0,0 +1,2317 @@
...
+///     .trim_start()
+/// );
+/// # Ok(())
+/// # }
+/// ```
+pub trait Visitor<'ast> {
+    type Output;
+    type Error;
+
+    fn visit_node(&mut self, n: &'ast Node) -> Result<Self::Output, Self::Error> {
```
Razican, line 87: These functions would need some documentation, so that they show up in the implementing types, and users know how to implement the trait, especially being `pub`. All methods would need an explanation, in case they want to be used by external crates.

boa_engine/src/syntax/ast/node/visit.rs
```diff
@@ -0,0 +1,2317 @@
...
+        self.walk_node(n)
+    }
```
Razican, line 89: I've seen this in multiple places. I understand that in some cases this is done because the two functions without spacing are related to each others, but it's more clean if we add spaces, it should look better, especially once they have documentation.

```suggestion
    }

```

boa_engine/src/syntax/ast/node/visit.rs
```diff
@@ -0,0 +1,2317 @@
...
+/// x + z;
+/// "#
+///     .trim_start()
+/// );
+/// # Ok(())
+/// # }
+/// ```
+pub trait Visitor<'ast> {
+    type Output;
+    type Error;
```
Razican, line 85: These types would need some documentation.

boa_engine/src/syntax/ast/node/visit.rs
```diff
@@ -0,0 +1,2317 @@
...
+                self.visit_declaration_pattern_mut(pattern)
+            }
+            BindingPatternTypeArray::SingleNameRest { ident } => self.visit_sym_mut(ident),
+            BindingPatternTypeArray::Empty | BindingPatternTypeArray::Elision => {
+                Self::get_default_ok()
+            }
+        }
+    }
+
+    fn get_default_ok() -> Result<Self::Output, Self::Error>;
```
Razican, line 2316: This function, being mandatory for external types to implement, needs documentation on how to use it, and what it's supposed to do.

Razican (commented):
Thanks for the contribution! And sorry for my delay in reviewing it. I've made some comments, mostly related to documentation. The design is good from my point of view.

Also, due to the delay in the review, this will need a rebase.

VTCAKAVSMoACE:
Closing as the internal API has changed dramatically.

PR #1957, opened 2022-03-19
[Merged by Bors] - migrated to clap 3

### LINKED ISSUE #1955: Update CLI to use Clap 3.0
We currently use [`structopt`](https://crates.io/crates/structopt) for our CLI, but now that Clap has released its 3.0 version, the functionality has improved to the point where `structopt` is no longer needed: https://crates.io/crates/clap

We would like to move from using `structopt` to using `clap`.

manthanabc:
This Pull Request fixes/closes #1955.

It changes the following:

- changes structopt to clap

### DIFF
--- a/boa_cli/src/main.rs
+++ b/boa_cli/src/main.rs
@@ -95,7 +95,7 @@ struct Opt {
         long,
         short = 'a',
         value_name = "FORMAT",
-        case_insensitive = true,
+        ignore_case = true,
         arg_enum
     )]
     dump_ast: Option<Option<DumpFormat>>,
@@ -189,7 +189,7 @@ where
 }
 
 pub fn main() -> Result<(), std::io::Error> {
-    let args = Opt::from_args();
+    let args = Opt::parse();
 
     let mut context = Context::default();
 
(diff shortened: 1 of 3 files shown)


manthanabc:
i think i need a code review i am confused when to use clap::structopt and clap both seem to do the same thing

boa_cli/src/main.rs
```diff
@@ -64,8 +64,8 @@ use boa_interner::Interner;
 use colored::{Color, Colorize};
 use rustyline::{config::Config, error::ReadlineError, EditMode, Editor};
 use std::{fs::read, io, path::PathBuf};
-use structopt::{clap::arg_enum, StructOpt};
-
+//use structopt::{clap::arg_enum, StructOpt};
```
Razican, line 67: I guess this is no longer needed, right?

boa_cli/src/main.rs
```diff
@@ -117,6 +117,7 @@ impl Opt {
     }
 }
 
+/*
 arg_enum! {
```
Razican, line 120: If this is no longer needed, can it be removed?

Razican (commented):
Thank you!! Looking very good!

Razican:
I think `clap::structopt` might be a compatibility layer

boa_cli/src/main.rs
manthanabc: Yeah i think it should be removed.

manthanabc:
ok i removed the extra code

manthanabc:
i might try getting rid of some depreciated function as warned by clippy later tommarow

Razican (approved):
Looks good from my side, but you need to run `cargo fmt` to format it and pass the format tests :)

raskad (approved):

Issue #1920, opened 2022-03-12
The cli parameter `--dump-ast Json` panics


raskad:
**Describe the bug**
Whey trying to output the AST with the cli parameter `--dump-ast Json` or `--dump-ast JsonPretty` boa panics.

**To Reproduce**

This JavaScript code reproduces the issue:
run the following code with `cargo run -- --dump-ast Json`
```javascript
let a = 1
```

`thread 'main' panicked at 'could not convert AST to a pretty JSON string: Error("can only flatten structs and maps (got a sequence)", line: 0, column: 0)', boa_cli/src/main.rs:181:30`

**Expected behavior**
Print the AST in JSON format.

hemantku:
Hi, My name is Vineet. I would work on this issue if this is still available

raskad:
Hi @hemantku, thank you for your interest in contributing to boa!
(Un)fortunately this will already be fixed in #1959

I noticed, that we don't have many good first issues currently. If you do not find one that looks interesting to you, I recommend heading to our [conformance page](https://boa-dev.github.io/boa/test262/) and checking out some currently failing tests, for which we do not have an issue open. That may be a good starting point if you want to help us improve our test coverage :)

PR #1959, opened 2022-03-20
[Merged by Bors] - Fix panic on AST dump in JSON format

kilotaras:
Some of the fields in AST structs were both

1. Arrays
2. Marked as 'flatten'

This is illegal per serde docs (and doesn't really make sense).
The fix is to remove the attribute.

See: https://serde.rs/attr-flatten.html
Fixes: #1920


jedel1043 (approved):
Thank you! 😁

raskad (approved):
Looks good, thank you for the contribution!

Issue #1147, opened 2021-02-28
Optimize integer operations


RageKnify:
Increment operators on `Value::Integer(i32)`s result in `Value::Rational(f64)`s because they use `Value::to_number` which returns `f64`. This makes traditional `for` loops with `i++` slower because they use floating point arithmetic rather than integer arithmetic.

---
#### My idea
I thought a good solution would be to have an internal:
```rust
enum Number {
  float(f64),
  int(i32)
}
```
We would also simplify to have a single `Value` variant for number rather than the current duo of `Value::Rational` and `Value::Integer`.
This would be the new return type of `Value::to_number` using the `int` variant when possible. It would also probably be a good fit tor the `Number` variant of `ObjectData`.

Open to other ideas :bulb:

Razican:
@HalidOdat I see you assigned yourself this issue, let me know if you're still working on it, or if we can un-assign it :)

HalidOdat:
> @HalidOdat I see you assigned yourself this issue, let me know if you're still working on it, or if we can un-assign it :)

I didn't implement this because it would make #1373 harder/impossible since integer and rational has will be encoded `JsValue`  into a 8-byte value and we can't fir an enum `Number` (16-byte). It would be added just to be removed when NaN boxing gets added.

Razican:
> > @HalidOdat I see you assigned yourself this issue, let me know if you're still working on it, or if we can un-assign it :)
> 
> 
> 
> I didn't implement this because it would make #1373 harder/impossible since integer and rational has will be encoded `JsValue`  into a 8-byte value and we can't fir an enum `Number` (16-byte). It would be added just to be removed when NaN boxing gets added.

Still, having integer operations optimized would be a good thing, right? Do you see a way we could get this? An option might be to use untagged enums (unions) or somehow add different `JsValue` variants.

HalidOdat:
> Still, having integer operations optimized would be a good thing, right? Do you see a way we could get this? An option might be to use untagged enums (unions) or somehow add different `JsValue` variants.

We are still going to have integer and rational variants, we wont remove them! But they have to be flat (since the tag is encoded with the payload, all in a 64bit value), we can't nest it (the current implementation is this way anyway).
[...]

PR #1964, opened 2022-03-21
[Merged by Bors] - Preserve ints when executing int operations

### LINKED ISSUE #1962: invalid type: floating point after math operations
**Describe the bug**

I have been using the new JsValue `to_json` functionality and I'm running into some issues. I am dealing with numbers in JS that should be valid integers, whether they be u32, i32, etc. boa seems to have issues with `to_json` because it is converting them into `Rational`s for an unknown reason. This might be related to https://github.com/boa-dev/boa/issues/1961.

**To Reproduce**

```rust
fn main() {
    println!("get_u32: {}", get_u32());
}

fn get_u32() -> u32 {
    let mut context = boa_engine::Context::default();

    let return_value = context.eval(r#"
        1000000 - 500
    "#).unwrap();

    serde_json::from_value(return_value.to_json(& mut context).unwrap()).unwrap()
}
```

**Expected behavior**

This should print `get_u32: 999500` and should not error out. Instead I get this error:

```bash
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: floating point `999500`, expected u32", line: 0, column: 0)', src/main.rs:12:74
```

- OS: Ubuntu 20.04
- Version: latest main
- Target triple: wasm32-unknown-unknown or x86_64-unknown-linux-gnu
- Rustc version: rustc 1.58.1 (db9d1b20b 2022-01-20)

**Additional context**

It seems like certain math operations are causing a Rust representation of `Rational` unnecessarily.

jedel1043:
This Pull Request fixes/closes #1962.

It changes the following:

- When executing arithmetic operations on `JsValue`s, try to use integer operations and fallback to `f64` operations on error.
- Adds tests for serde_json conversions from integer operations.

### DIFF
--- a/boa_engine/src/value/operations.rs
+++ b/boa_engine/src/value/operations.rs
@@ -9,15 +9,19 @@ impl JsValue {
     pub fn add(&self, other: &Self, context: &mut Context) -> JsResult<Self> {
         Ok(match (self, other) {
             // Fast path:
-            (Self::Integer(x), Self::Integer(y)) => Self::new(f64::from(*x) + f64::from(*y)),
+            // Numeric add
+            (Self::Integer(x), Self::Integer(y)) => x
+                .checked_add(*y)
+                .map_or_else(|| Self::new(f64::from(*x) + f64::from(*y)), Self::new),
             (Self::Rational(x), Self::Rational(y)) => Self::new(x + y),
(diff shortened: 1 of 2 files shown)


HalidOdat (approved):
Looks good to me! :)

RageKnify (approved):
bors r+

Razican:
@jedel1043 does this solve #1147?

Issue #1970, opened 2022-03-22
Implement `i128` to `JsValue` conversions


Razican:
We can currently convert most of Rust's basic types to `JsValue`, but since the `i128` type was a late addition to Rust, we don't offer that conversion. You can implement it by checking how this is done for `i64`:
https://github.com/boa-dev/boa/blob/e2630faf82a27e958d0ca6c9b1c81554bcc07ff3/boa_engine/src/value/mod.rs#L1003-L1008

Shreyan11:
Just to be clear, you want to convert all i28 types to JsValue?

sudosmile:
I'd like to work on this issue

jonah-saltzman:
Could I work on this issue? It seems to have not been solved on main or the latest commit.

lastmjs:
I believe there's an open pull request for this: https://github.com/boa-dev/boa/pull/1971

Razican:
This will be handled in #2276.

Issue #1961, opened 2022-03-21
to_json u32 error


lastmjs:
**Describe the bug**

I have been using the new JsValue `to_json` functionality and I'm running into some issues. I have a valid `u32` represented in JavaScript that should be convertible to a u32 in Rust, but it isn't.

**To Reproduce**

```rust
fn main() {
    println!("get_u32: {}", get_u32());
}

fn get_u32() -> u32 {
    let mut context = boa_engine::Context::default();

    let return_value = context.eval(r#"
        4294967295
    "#).unwrap();

    serde_json::from_value(return_value.to_json(& mut context).unwrap()).unwrap()
}
```

**Expected behavior**

This should print `get_u32: 4294967295` and should not error out. Instead I get this error:

```bash
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: floating point `4294967295`, expected u32", line: 0, column: 0)', src/main.rs:12:74
```

**Build environment (please complete the following information):**

- OS: Ubuntu 20.04
- Version: latest main
- Target triple: wasm32-unknown-unknown or x86_64-unknown-linux-gnu
- Rustc version: rustc 1.58.1 (db9d1b20b 2022-01-20)

**Additional context**

I am also running into issues where JS numbers that should be representable in Rust as integers are instead `Rational`s, thus causing `to_json` to fail. I will open another issue for those problems, but I think they might be related to this issue.

lastmjs:
I certainly hope you'll support it, the `to_json` combined with `serde_json::from` is such a simple DX.

lastmjs:
Also it would be nice to have support for most of the default Rust primitive values as far as possible, perhaps with some easy interop with `bigint`.

jedel1043:
Also, consider that we don't NEED to support u32 to add this feature, we would just try to convert the `f64`s to `u32`s when converting to json, but it would be good to determine our priorities about integer support.

HalidOdat:
Hmmm... Adding `u32` integers, in addition to `i32` would be a performance penalty (besides complexity) for any number operation.

As suggested by @jedel1043 It would be far better to convert the `f64` value to `u32` in `to_json()`.

lastmjs:
I'm trying to build a workaround for this in my project, and it's proving a tad more difficult than I thought. I just want to get a reading on what I can expect from boa.

It would be ideal for boa to allow conversion to any of the following. Since `BigInt` is part of the language, I believe each conversion should be possible:

[...]

PR #1971, opened 2022-03-23
Add From<i/u128> for JsBigInt

dranikpg:
This Pull Request fixes [1970](https://github.com/boa-dev/boa/issues/1970).

It changes the following:

- Added From<i128>, From<u128> for JsBigInt
- Added From<i128>, From<u128> for Numeric

### DIFF
--- a/boa_engine/src/value/mod.rs
+++ b/boa_engine/src/value/mod.rs
@@ -1000,6 +1000,13 @@ impl From<f32> for Numeric {
     }
 }
 
+impl From<i128> for Numeric {
+    #[inline]
+    fn from(value: i128) -> Self {
+        Self::BigInt(value.into())
+    }
+}
+
 impl From<i64> for Numeric {
     #[inline]
     fn from(value: i64) -> Self {
@@ -1025,6 +1032,13 @@ impl From<i8> for Numeric {
     #[inline]
     fn from(value: i8) -> Self {
         Self::Number(value.into())
+    }
+}
+
+impl From<u128> for Numeric {
+    #[inline]
+    fn from(value: u128) -> Self {
+        Self::BigInt(value.into())
     }
 }
 
--- a/boa_engine/src/bigint.rs
+++ b/boa_engine/src/bigint.rs
@@ -370,6 +370,24 @@ impl From<u64> for JsBigInt {
     }
 }
 
+impl From<i128> for JsBigInt {
+    #[inline]
+    fn from(value: i128) -> Self {
+        Self {
+            inner: Rc::new(RawBigInt::from(value)),
+        }
+    }
+}
+
+impl From<u128> for JsBigInt {
+    #[inline]
+    fn from(value: u128) -> Self {
+        Self {
+            inner: Rc::new(RawBigInt::from(value)),
+        }
+    }
+}
+
 impl From<isize> for JsBigInt {
     #[inline]
     fn from(value: isize) -> Self {


Razican (approved):
Thank you for the contribution!! Looks great :)

jasonwilliams (approved):
Looks good to me

Razican:
> Should i128 really be clamped to f64 like i64 is?
> 
> 
> 
> ```rust
> 
> impl From<i64> for JsValue {
> 
>     #[inline]
> 
>     fn from(value: i64) -> Self {
> 
>         if let Ok(value) = i32::try_from(value) {
> 
>             Self::Integer(value)
> 
>         } else {
> 
>             Self::Rational(value as f64)
> 
>         }
> 
>     }
> 
> }
> 
> ```
> 
> 
> 
> I'd rather expect something like this, as accuracy would be very low for big values otherwise
> 
> 
> 
> ```rust
> 
> impl From<i128> for JsValue {
> 
>     #[inline]
> 
>     fn from(value: i128) -> Self {
> 
>         if let Ok(value) = i32::try_from(value) {
> 
>             Self::Integer(value)
> 
>         } else {
> 
>             Self::BigInt(value.into()) // i.e. JsBigInt::from()
> 
>         }
> 
>     }    
> 
> }
> 
> ```

I would say this can be treated in a separate issue, it's a subjective topic :)

boa_engine/src/value/mod.rs
```diff
@@ -1000,6 +1000,13 @@ impl From<f32> for Numeric {
     }
 }
 
+impl From<i128> for Numeric {
+    #[inline]
+    fn from(value: i128) -> Self {
+        Self::BigInt(value.into())
+    }
+}
+
```
HalidOdat, line 1009: This should be a number (`f64`) instead of BigInt

HalidOdat, line 1009: I'm no sure if we even want this Numeric is more of an internal thing.

Razican: Well, we are already doing this with `i64`, and `BigInt` would make sure that it doesn't lose precision:
https://github.com/boa-dev/boa/blob/e2630faf82a27e958d0ca6c9b1c81554bcc07ff3/boa_engine/src/value/mod.rs#L1003-L1008

HalidOdat: > Well, we are already doing this with i64

Ummm.... This is wrong, It should be a number, not bigint.

> BigInt would make sure that it doesn't lose precision

It's fine if we lose precision (we lose precision if we type a number value which is too big for `f64` in JavaScript, same with `From<i64>` for `JsValue`), If we don't want loss of precision we create a bigint and we can convert it into `Numeric`.
The behaviour of `From<u/iN>` should be uniform, to do some number and others bigint is confuzing.

Razican: There was some discussion here too: #1961

From my point of view, we should be able to convert an `f64` to `JsValue` or `Numeric`, and a `BigInt` too. But nothing else (at most an i32). If the user wants to convert a `i128` to a `BigInt`, they can first convert it to a `BigInt` and then convert the `BigInt` to a `Numeric` or a `JsValue`.

We shouldn't probably force anything, but give some options. Maybe a `TryFrom` implementation could be nice.

HalidOdat: > [](https://github.com/boa-dev/boa/issues/1961)
> 
> From my point of view, we should be able to convert an `f64` to `JsValue` or `Numeric`, and a `BigInt` too. But nothing else (at most an i32). If the user wants to convert a `i128` to a `BigInt`, they can first convert it to a `BigInt` and then convert the `BigInt` to a `Numeric` or a `JsValue`.
> 
> We shouldn't probably force anything, but give some options. Maybe a `TryFrom` implementation could be nice.

Yup. agreed

boa_engine/src/value/mod.rs
```diff
@@ -1000,6 +1000,13 @@ impl From<f32> for Numeric {
     }
 }
 
+impl From<i128> for Numeric {
+    #[inline]
+    fn from(value: i128) -> Self {
+        Self::BigInt(value.into())
+    }
+}
```
HalidOdat, line 1008: We should remove this, as well as `From<u128> for Numeric`. If the user wants to a bigint or number they should explicitly say so.

Besides it's out of the scope of this PR of adding `From<i/u128> for JsBigInt`

lastmjs:
Don't we want to implement the `From<i128>` trait for `JsValue`? Why aren't we adding an implementation in here: https://github.com/boa-dev/boa/blob/main/boa_engine/src/value/conversions.rs?

Razican:
This will be handled as part of #2276.

PR #1980, opened 2022-03-27
[Merged by Bors] - Retrieve feature-based results for Test262 runs

### LINKED ISSUE #1645: Retrieve feature-based results for Test262 runs
The current implementation of the Tes262 compliance results page shows the compliance test suite by test suite, which is straightforward, since we run the tests suite by suite, test by test, and then we just collect the information about it.

It would be very nice to have the option to also see it feature by feature. Some tests, such as [this one](https://github.com/tc39/test262/blob/12e9d67bb529474ad1ffddca92361512ab30c4c0/test/language/block-scope/syntax/redeclaration/async-function-name-redeclaration-attempt-with-const.js) have a list of features (only one in this case). This information is already being stored in the `MetaData` information structure [here](https://github.com/boa-dev/boa/blob/main/boa_tester/src/read.rs#L18).

In order to show these results, we first need to gather them. Currently, when we run a test suite, we get a `SuiteResult`, as you can see [here](https://github.com/boa-dev/boa/blob/main/boa_tester/src/exec/mod.rs#L16). These results are generated from the `TestResult` each test returns, as it can be seen [here](https://github.com/boa-dev/boa/blob/main/boa_tester/src/exec/mod.rs#L88). The definition of both structures can be seen in the `main.rs` file of the `boa_tester` project, [here](https://github.com/boa-dev/boa/blob/main/boa_tester/src/main.rs#L299-L329).

We should change the `SuiteResult` structure to include information on the features each test had. This would allow having feature coverage per test-suite. Since the top-most suite will include feature information for all the suites below it, we can store that information on a JSON file, as we do with the rest of the `SuiteResult`.

Note that we want to store this information in a separate JSON file, so we will need to ignore part of the `SuiteResult` structure when serializing it, and extract that information to a new structure that will be stored in a new JSON file. Similarly to what we do with the [`ReducedResultInfo` structure](https://github.com/boa-dev/boa/blob/main/boa_tester/src/results.rs#L24-L37). Note that there should be also results for "no features", for base language tests.

After this, we should be able to use that file in the results page.

NorbertGarfield:
This Pull Request fixes/closes #1645.

It changes the following:

- Add `features` field to `SuiteResult` structure
- Fetch features from `TestSuite` and propagate them via `SuiteResult`
- Add `FeaturesInfo` structure and serialize it to `features.json`

### DIFF
--- a/boa_tester/src/exec/mod.rs
+++ b/boa_tester/src/exec/mod.rs
@@ -43,6 +43,13 @@ impl TestSuite {
                 .collect()
         };
 
+        let mut features = Vec::new();
+        for test_iter in self.tests.iter() {
+            for feature_iter in test_iter.features.iter() {
+                features.push(feature_iter.to_string());
+            }
+        }
+
         if verbose != 0 {
             println!();
         }
@@ -67,6 +74,7 @@ impl TestSuite {
             passed += suite.passed;
             ignored += suite.ignored;
             panic += suite.panic;
+            features.append(&mut suite.features.clone())
         }
 
         if verbose != 0 {
@@ -95,6 +103,7 @@ impl TestSuite {
             panic,
             suites,
             tests,
+            features,
         }
     }
 }
--- a/boa_tester/src/exec/mod.rs
+++ b/boa_tester/src/exec/mod.rs
@@ -74,7 +74,7 @@ impl TestSuite {
             passed += suite.passed;
             ignored += suite.ignored;
             panic += suite.panic;
-            features.append(&mut suite.features.clone())
+            features.append(&mut suite.features.clone());
         }
 
         if verbose != 0 {
--- a/boa_tester/src/results.rs
+++ b/boa_tester/src/results.rs
@@ -50,11 +50,44 @@ impl From<ResultInfo> for ReducedResultInfo {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
+struct FeaturesInfo {
+    #[serde(rename = "c")]
+    commit: Box<str>,
+    #[serde(rename = "u")]
+    test262_commit: Box<str>,
+    #[serde(rename = "n")]
+    suite_name: Box<str>,
+    #[serde(rename = "f")]
+    features: Vec<String>,
+}
+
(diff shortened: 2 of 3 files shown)


boa_tester/src/exec/mod.rs
```diff
@@ -67,6 +74,7 @@ impl TestSuite {
             passed += suite.passed;
             ignored += suite.ignored;
             panic += suite.panic;
+            features.append(&mut suite.features.clone())
```
Razican, line 77: This is giving a clippy warning, should be fixed with this:

```suggestion
            features.append(&mut suite.features.clone());
```

boa_tester/src/results.rs
```diff
@@ -50,12 +50,45 @@ impl From<ResultInfo> for ReducedResultInfo {
...
+    commit: Box<str>,
+    #[serde(rename = "u")]
+    test262_commit: Box<str>,
+    #[serde(rename = "n")]
+    suite_name: Box<str>,
+    #[serde(rename = "f")]
+    features: Vec<String>,
+}
+
+fn remove_duplicates(features_vec: &Vec<String>) -> Vec<String> {
```
Razican, line 65: This should solve the clippy error, but might need some changes in the code to convert the slice to a vector.

```suggestion
fn remove_duplicates(features_vec: &[String]) -> Vec<String> {
```

Razican (commented):
Thanks! Looks good to me :) only those two clippy issues.

boa_tester/src/exec/mod.rs
NorbertGarfield: Thanks. Applied https://github.com/boa-dev/boa/pull/1980/files#diff-fc3d7ad7b5e64574258c9febbe56171f3309b74e0c8da35238a76002f3ee34d9R77

boa_tester/src/results.rs
NorbertGarfield: Done. https://github.com/boa-dev/boa/pull/1980/files#diff-fc3d7ad7b5e64574258c9febbe56171f3309b74e0c8da35238a76002f3ee34d9R77

Razican (approved):

jedel1043 (approved):
Looks good to me!

PR #1990, opened 2022-03-30
[Merged by Bors] - Add From<f32> for JsValue

### LINKED ISSUE #1989: Implement From<f32> for JsValue
I've noticed that the From trait for f32, i128, i16, i8, u128, u16, and u8 are not implemented for JsValue. This is my first time contributing directly to boa, and so I'm just fixing something small that I think is a problem. This is all in preparation for https://github.com/boa-dev/boa/issues/1975

lastmjs:
This Pull Request fixes/closes #1989.

It changes the following:

- Implements From<f32> for JsValue

### DIFF
--- a/boa_engine/src/value/conversions.rs
+++ b/boa_engine/src/value/conversions.rs
@@ -54,6 +54,18 @@ impl From<f64> for JsValue {
     }
 }
 
+impl From<f32> for JsValue {
+    #[allow(clippy::float_cmp)]
+    #[inline]
+    fn from(value: f32) -> Self {
+        // if value as i32 as f32 == value {
+        //     Self::Integer(value as i32)
+        // } else {
+        Self::Rational(value.into())
+        // }
+    }
+}
+
 impl From<u32> for JsValue {
     #[inline]
     fn from(value: u32) -> Self {


boa_engine/src/value/conversions.rs
```diff
@@ -54,6 +54,18 @@ impl From<f64> for JsValue {
...
 
+impl From<f32> for JsValue {
+    #[allow(clippy::float_cmp)]
+    #[inline]
+    fn from(value: f32) -> Self {
+        // if value as i32 as f32 == value {
+        //     Self::Integer(value as i32)
+        // } else {
+        Self::Rational(value.into())
+        // }
```
jedel1043, line 65: Why isn't this mirroring the `f64` conversion?

raskad: It is the same as the `f64` one no?

jedel1043: Ah, right, forgot we removed the `f64` conversion to integer

jedel1043 (approved):
Thank you!

raskad (approved):

Issue #1991, opened 2022-03-30
Integer conversion conventions


lastmjs:
I opened this issue as suggested by this comment: https://github.com/boa-dev/boa/pull/1971#issuecomment-1077297915

As I've been going through the conversion code from Rust integer types to JsValue, I've come across some conversions that don't make sense to me. I'd like to open this for discussion.

My biggest concern is `i128`, `u128`, `i64`, and `u64`. I don't see why all of those shouldn't be converted into a `JsValue::BigInt`. A `bigint` in JavaScript is the only native number type that can maintain true integer semantics and hold all possible values of `i128`, `u128`, `i64`, and `u64`, thus why should we not always convert to a `bigint`?

Razican:
The thing here is that this is a fairly complex and very subjective topic.

JavaScript has 2 numeric types:
 - "number" -> `f64` in Rust
 - "bigint" -> `BigInt` in Rust

There is nothing else in JavaScript. So, it's definitely very difficult for us to provide the conversions you mention. Since, for example, a `i32` might make sense to be converted to "number", since it won't loose precision, and it can be stored in the stack. `i128`, on the other hand, cannot be converted to JavaScript the same way as Rust uses it. Either it gets converted to `f64`, and therefore loses precision, or it gets converted to `bigint`, and this requires an allocation in the stack.

Why should Boa force any of the two options on Boa users?

We should provide a `f64` to `JsValue` and a `BigInt` to `JsValue` conversion. These two don't fail and can be represented in a straightforward way. This can be done with the `From` / `Into` traits. Boa has also a special performance optimization that allows you to use a `i32` as if it were a "number". You can already convert an arbitrary `i32` to a "number":

```rust
let num = 1256_i32;
let js_num = JsValue::Integer(num);
```

For other types, we shouldn't force anything. Users can easily (even now) convert any arbitrary Rust type to a `JsValue` (by implementing the logic, of course), and they can get a `JsValue` and decide how they can represent it in Rust. So, if I think `i128` is best represented by the "number" JavaScript type, since it's stored in the stack the same way as Rust does, I can do the following:

```rust
let num = 1256_i128;
let js_num = JsValue::Rational(num as f64);
```

If, on the other hand, I prefer the precision, I can do the following:

```rust
let num = 1256_i128;
let js_bigint = JsValue::BigInt(BigInt::from(num).into());
```

[...]

PR #1993, opened 2022-03-31
JsValue from primitives

### LINKED ISSUE #1994: Implement From<T> for JsValue where T are primitive values
I think that From<T> should be implemented in boa for as many types as possible for JsValue.

lastmjs:
This Pull Request fixes/closes #1994 and #1991 and depends on https://github.com/boa-dev/boa/pull/1971.

It changes the following:

- Adds many primitive conversions, From<T> for JsValue where T are integers
- Makes some opinions about `i128`, `u128`, `i64`, and `u64` when converting to JsValue, always converting to BigInt

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -563,9 +563,9 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.2.5"
+version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


raskad (approved):
Looks good!

Razican (changes requested):
As already mentioned in multiple comments, I don't think it's the way forward. This already silently heap-allocates with a TON of overhead many Rust stack primitives. you can already do these conversions explicitly, and there are talks for adding easier boilerplating with a trait and maybe some derives.

lastmjs:
I'm going to close this pull request in response to the comments.

PR #1995, opened 2022-03-31
IntoJsValue and TryFromJsValue

### LINKED ISSUE #1975: TryFromJsValue
Working on [Azle](https://github.com/demergent-labs/azle) which is a JS environment for the Internet Computer, I need to have a general solution for converting `JsValue`s into arbitrary types. I have been attempting to do this using `serde_json::from_value(js_value.to_json(& mut context).unwrap()).unwrap()` which works in many cases, but breaks down sometimes (for example with `undefined` and `bigint`). It's also a more complicated developer experience and probably less performant than a direct conversion from `JsValue` to an arbitrary type would be.

The below is from this comment: https://github.com/boa-dev/boa/issues/1961#issuecomment-1075673692

> This would make a lot of sense, yes. We could have a TryFromJsValue trait that would receive a context and a JsValue and try to perform the conversion. Just thinking out loud, but I could imagine a similar API to the Serializa/Deserialize APIs in serde.

I would like to start a discussion on some sort of `TryFromJsValue`. Is this something you think boa could/would want to support? If so, what should the API look like?

I would like this functionality ASAP and so if it would be accepted and I could receive some guidance, I am happy to do the work. If it's relatively simple for the maintainers to go ahead and do it then I am happy with that as well. Just hope this can be done.

Thanks!

lastmjs:
I'd love to get feedback on this, especially how I should handle errors and any other idiomatic Rust improvements you'd like me to make (I'm still kind of new to Rust).

This pull request also depends on making the Array builtin public, and in my opinion all of the builtins should be made public. I can open another pull request to make all of those public if I get the go-ahead.

This pull request also depends on implementing my opinion that `i128`, `u128`, `i64` and `u64` should always be converted into a JS `bigint` since it's the only native JS number type that will not lose precision. More discussion here: https://github.com/boa-dev/boa/issues/1991

This Pull Request fixes/closes #1975.

It changes the following:

- Adds `IntoJsValue` trait with many implementations
- Add `TryFromJsValue` trait with many implementations


boa_engine/src/value/into_js_value.rs
```diff
@@ -0,0 +1,127 @@
...
+impl<T: IntoJsValue> IntoJsValue for Vec<T> {
+    fn into_js_value(self, context: &mut Context) -> JsValue {
+        let js_values = self.into_iter().map(|item| item.into_js_value(context)).collect::<Vec<JsValue>>();
+
+        Array::create_array_from_list(
+            js_values,
+            context
+        ).into()
+    }
+}
```
jedel1043, line 115: Exposing `Array` is pretty unnecessary, since there's already an API that allows to convert any `Iterator<Item=JsValue>` into an array.

https://github.com/boa-dev/boa/blob/main/boa_engine/src/object/jsarray.rs

jedel1043 (changes requested):
I mostly agree with the changes, but I think there are some API questions that need to be discussed.

boa_engine/src/value/into_js_value.rs
lastmjs: Yeah I'm a bit confused on how to do this the right way, can you give me a little code hint?

lastmjs: Oh I think I see, from_iter would give me a JsArray and then I could get it to a JsValue from there, let me work on that

lastmjs: This has been addressed

jedel1043:
Closing in favour of #2276.

PR #1997, opened 2022-04-01
Expose raw builtin functions

HalidOdat:
This PR exposes the raw buitin functions.

It changes the following:
- Expose Builtin functions
- Add pub reexport in bultin functions
- Add disclaimer in builtin modules that the raw builtin function could change.


raskad:
I think we should make a strategic decision if we want to make the builtins public or if we want to follow the `Js*`-type approach like in https://github.com/boa-dev/boa/pull/2003 or https://github.com/boa-dev/boa/pull/1746. Doing both seems like giving users different apis for the same task.
If we make the builtins public, I do not see any direct benefit of additional `Js*` types.

HalidOdat:
Yeah, the `Js`-types are an easier and more user friendly API. So I'm closing this :)

PR #1999, opened 2022-04-02
[Merged by Bors] - fix(vm): off-by-one in code block stringification.

### LINKED ISSUE #1998: Locations in dumps of compiled code blocks are off by one
**Describe the bug**

When `boa` is printing a compiled `CodeBlock`, the locations printed correspond to the location of the wrong instruction (they are the following instruction, not the current instruction)

**To Reproduce**

```javascript
// bug.js
if (false){
    1;
} else{
    0;
}
```
Then run `cargo run -- -t bug.js`:
```
...

000001    0000    PushFalse
000006    0001    JumpIfFalse                19
000011    0002    PushDeclarativeEnvironment 0
000012    0003    PushOne
000013    0004    Pop
000014    0005    PopEnvironment
000019    0006    Jump                       27
000024    0007    PushDeclarativeEnvironment 0
000025    0008    PushZero
000026    0009    Pop
000027    0010    PopEnvironment
...
```

**Expected behavior**
In this output, you can see a few weird things:
1. The first instruction is at location 1, not 0
2. The Jump operands don't line up printed Locations - the first `JumpIfFalse` has operand 19, but the instruction listed at location 19 is another Jump.

- OS: Windows
- Version: 0.14 (on the latest `main` as of writing)
- Rustc version: rustc 1.59.0 (9d1b2106e 2022-02-23)

tsutton:
Acked-by: Taylor Sutton <tsutton125@gmail.com>

This Pull Request fixes/closes #1998 

The call to retrieve operands modifies pc, setting it to the index of
the *next* instruction. So, we save its initial value and use that
for printing.

### DIFF
--- a/boa_engine/src/vm/code_block.rs
+++ b/boa_engine/src/vm/code_block.rs
@@ -351,9 +351,10 @@ impl ToInternedString for CodeBlock {
         while pc < self.code.len() {
             let opcode: Opcode = self.code[pc].try_into().expect("invalid opcode");
             let opcode = opcode.as_str();
+            let previous_pc = pc;
             let operands = self.instruction_operands(&mut pc, interner);
             f.push_str(&format!(
-                "{pc:06}    {count:04}    {opcode:<27}{operands}\n",
+                "{previous_pc:06}    {count:04}    {opcode:<27}{operands}\n",
             ));
             count += 1;
         }


raskad (approved):
Nice find, thank you for the contribution!

jedel1043 (approved):
Thanks!

PR #2003, opened 2022-04-03
[Merged by Bors] - Feature `JsTypedArray`s

HalidOdat:
This PR adds a wrapper around the raw builtins for JavaScript typed arrays, like #1746

### DIFF
--- a/boa_engine/src/builtins/typed_array/mod.rs
+++ b/boa_engine/src/builtins/typed_array/mod.rs
@@ -114,7 +114,7 @@ macro_rules! typed_array {
             ///  - [ECMAScript reference][spec]
             ///
             /// [spec]: https://tc39.es/ecma262/#sec-typedarray
-            fn constructor(
+            pub(crate) fn constructor(
                 new_target: &JsValue,
                 args: &[JsValue],
                 context: &mut Context,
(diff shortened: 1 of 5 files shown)


boa_engine/src/builtins/typed_array/mod.rs
```diff
@@ -2951,7 +2951,7 @@ impl TypedArray {
     }
 
     /// <https://tc39.es/ecma262/#sec-initializetypedarrayfromlist>
-    fn initialize_from_list(
+    pub(crate) fn initialize_from_list(
```
lastmjs, line 2954: Why are these functions only visible to the crate? Why don't we expose them to the developer using boa? If they're in the ECMAScript standard, shouldn't they be exposed and accessible?

jedel1043: ECMAScript doesn't define `InitializeTypedArrayFromList` as an exposed API, it's just an abstract operation used internally. The API exposed by the spec are the builtin object properties and the global object properties.

boa_engine/src/object/jstypedarray.rs
```diff
@@ -0,0 +1,407 @@
...
+    value::IntoOrUndefined,
+    Context, JsResult, JsString, JsValue,
+};
+use boa_gc::{Finalize, Trace};
+use std::ops::Deref;
+
+/// JavaScript `TypedArray` rust object.
+#[derive(Debug, Clone, Trace, Finalize)]
+pub struct JsTypedArray {
+    inner: JsObject,
```
raskad, line 13: Does it make sense to have `inner` as a `JsObject`? Almost every usage of `inner` has to clone the gc pointer, wrap it into a `JsValue` and pass that reference.

HalidOdat: > Does it make sense to have `inner` as a `JsObject`? Almost every usage of `inner` has to clone the gc pointer, wrap it into a `JsValue` and pass that reference.

Hmmm.... Semantically it should be an object, but from a performance point of view it should be a `JsValue` so we don't clone and wrap it.

In all but one case (where we call `Deref<JsObject>`) we need it as `JsValue`

raskad (approved):

jedel1043 (approved):
Good work!

A bit off topic, but what are your plans for the `JsObjectType` trait?

PR #2015, opened 2022-04-07
[Merged by Bors] - Feature `JsFunction`

HalidOdat:
This PR adds `JsFunction` wrapper around JavaScript `Function` object, like #1746 
With this PR we can distinguish between regular object and function object when we need, such as accessors (because they always need to be functions), predicates in `JsArray` methods like `map`, `find`, etc. With this abstraction we leverage the type system of rust which cleans the API making intentions clear.

It changes the following:
- Make methods that take predicate/callback function take `JsFunction`s
- Make `.accessor()` and `.static_accessor()` take `Option<JsFunction>`
- Make `FunctionBuilder` return `JsFunction`
- Make `ConstructorBuilder` return `JsFunction`
- Make `ClassBuilder` return `JsFunction`

### DIFF
--- a/boa_engine/src/object/jsfunction.rs
+++ b/boa_engine/src/object/jsfunction.rs
@@ -1,0 +1,56 @@
+use crate::{
+    object::{JsObject, JsObjectType},
+    Context, JsResult, JsValue,
+};
+use boa_gc::{Finalize, Trace};
+use std::ops::Deref;
+
+/// JavaScript `Function` rust object.
+#[derive(Debug, Clone, Trace, Finalize)]
+pub struct JsFunction {
+    inner: JsObject,
+}
+
+impl JsFunction {
+    #[inline]
+    pub(crate) fn from_object_unchecked(object: JsObject) -> Self {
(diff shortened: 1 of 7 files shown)


boa_engine/src/object/jsfunction.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl JsFunction {
+    #[inline]
+    pub(crate) fn from_object_unchecked(object: JsObject) -> Self {
+        Self { inner: object }
+    }
+
+    #[inline]
+    pub fn from_object(object: JsObject, context: &mut Context) -> JsResult<Self> {
```
Razican, line 21: I think that at least public methods should have some documentation

Razican (commented):
Looks very good, but I'm missing some documentation on public methods (of `JsFunction` and other public APIs) in order to make this very usable, maybe with some example code, for instance.

Razican (approved):
This is good for me :) Let's see what the rest thinks.

raskad (approved):

PR #2021, opened 2022-04-08
[Merged by Bors] - Allow unicode escaped characters in identifiers that are keywords

raskad:

- Remove syntax error for unicode escaped characters in keywords from the lexer.
- Adjust the lexer tokens for keywords to indicate if they contain unicode escaped characters.
- Throw syntax errors in parser, when keywords cannot contain unicode escaped characters.


Razican (approved):
Looking good!

HalidOdat (approved):

PR #2023, opened 2022-04-10
implement `u128` and `i128` to `Numeric`

### LINKED ISSUE #1970: Implement `i128` to `JsValue` conversions

sudosmile:
This Pull Request fixes/closes #1970 .

It changes the following:
- Add conversion from `u128` and `i128` to `JsBigInt` type
- Add conversion from `u128` and `i128` to `Numeric` type

### DIFF
--- a/boa_engine/src/bigint.rs
+++ b/boa_engine/src/bigint.rs
@@ -370,6 +370,24 @@ impl From<u64> for JsBigInt {
     }
 }
 
+impl From<i128> for JsBigInt {
+    #[inline]
+    fn from(value: i128) -> Self {
+        Self {
+            inner: Rc::new(RawBigInt::from(value)),
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


jasonwilliams (approved):

Razican:
There are some talks going on about wether it makes sense for this to be implemented like this: https://github.com/boa-dev/boa/pull/1971/files/5d9154ee0adfc3ad7079a27c3772d201d4acf4ec#r837040291

PR #2026, opened 2022-04-12
cargo workspaces fixes #2001

### LINKED ISSUE #2001: Use Cargo Workspaces for publishing
We should setup [Cargo Workspaces](https://crates.io/crates/cargo-workspaces) for publishing next time around similar to [rust analzyer](https://github.com/rust-analyzer/rust-analyzer/blob/master/.github/workflows/publish.yml)

jasonwilliams:
This adds cargo-workspaces to our repo for easier publishing. 
(I think we may need to do a dry run of this first to test)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,18 +4,34 @@ on:
     types: [published]
 
 jobs:
-  # publish:
-  #   runs-on: ubuntu-latest
-  #   steps:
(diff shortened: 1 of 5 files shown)


raskad (approved):
Looks good to me in theory. I agree we should try a dry run somehow

Razican (approved):
Looks good to me, we'll have to test it in the next release :)

PR #2031, opened 2022-04-16
[Merged by Bors] - Implement `get RegExp.prototype.hasIndices`

HalidOdat:
It changes the following:
- Implement `get RegExp.prototype.hasIndices`
- Update Implement `get RegExp.prototype.flags` to currect spec
- Update `RegExpFlags`


jedel1043 (approved):
Nice!

raskad (changes requested):
The new flag is missing from `ToString for RegExpFlags` at in boa_engine/src/syntax/lexer/regex.rs.

That should make the missing tests pass.
For example `/./d.hasIndices` currently returns `false`.

Razican (approved):
Looks good to me :)

raskad (approved):
bors r+

PR #2036, opened 2022-04-19
[Merged by Bors] - Implement ResolveLocale helper

NorbertGarfield:
This Pull Request implements ResolveLocale abstract method. It is required for further InitializeDateTimeFormat development.

It changes the following:

- Adds several helpers to operate with locale extensions
- Adds DefaultLocale placeholder
- Implements BestAvailableLocale and locale matchers
- Implements UnicodeExtensionsComponents
- Introduces testing


raskad (changes requested):
Thanks for the contribution.

I commented on some things but I have some more general comments here.

Instead of working with `String`s, it would probably be better to use `JsString`s. We plan to switch to UTF-16 strings and these would need to be converted too.

There are some data structures in this code that the spec defines as `Records`. For example I commented on the return value of `ResolveLocale`. These are currently either `JsValue`s or `JsObject`s. I think it would be much better to have them as distinct struct types.

Also `available_locales` and `requested_locales` are `JsValue`s but you convert them to objects and then use them as array-like objects. The spec specifies to use a `List`, so it would be probably better to use a `Vec<SomeType>` instead.

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,596 @@ impl Intl {
...
+fn lookup_matcher(
+    available_locales: &JsValue,
+    requested_locales: &JsValue,
+    context: &mut Context,
+) -> MatcherRecord {
+    // 1. Let result be a new Record.
+    let mut result = MatcherRecord {
+        locale: "undefined".to_string(),
+        extension: "".to_string(),
+    };
```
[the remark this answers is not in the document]
NorbertGarfield: Done. https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R276

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,596 @@ impl Intl {
...
+
+fn trim_extensions(requested_locale: &str) -> String {
+    let trim_pos = get_leftmost_ext_pos(requested_locale);
+    requested_locale[..trim_pos].to_string()
+}
+
+fn extract_extensions(requested_locale: &str) -> String {
+    let trim_pos = get_leftmost_ext_pos(requested_locale);
+    requested_locale[trim_pos..].to_string()
+}
```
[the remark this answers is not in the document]
NorbertGarfield: I think, it is possible to search for '-u-' pattern. https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R159

Let me know if there's another way.

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,596 @@ impl Intl {
...
+/// The `ResolveLocale` abstract operation compares a BCP 47 language priority list
+/// `requestedLocales` against the locales in `availableLocales` and determines the best
+/// available language to meet the request. `availableLocales`, `requestedLocales`, and
+/// `relevantExtensionKeys` must be provided as `List` values, options and `localeData` as Records.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///
+/// [spec]: https://tc39.es/ecma402/#sec-resolvelocale
+pub fn resolve_locale(
```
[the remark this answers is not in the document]
NorbertGarfield: Done https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R513

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,596 @@ impl Intl {
...
+        lookup_matcher(available_locales, requested_locales, context)
+    } else {
+        best_fit_matcher(available_locales, requested_locales, context)
+    };
+
+    // 4. Let foundLocale be r.[[locale]].
+    let mut found_locale = r.locale;
+
+    // 5. Let result be a new Record.
+    let result = JsObject::empty();
```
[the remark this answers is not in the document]
NorbertGarfield: Done. https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R538

NorbertGarfield:
@raskad do I need to change '&str' to '&JsString' as well?

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,566 @@ impl Intl {
...
+        // j. Set result.[[<key>]] to value.
+        result.properties.insert(key.clone(), value);
+
+        // k. Append supportedExtensionAddition to supportedExtension.
+        supported_extension = JsString::concat(supported_extension, &supported_extension_addition);
+    }
+
+    // 10. If the number of elements in supportedExtension is greater than 2, then
+    let ext_elements = unicode_extension_components(&supported_extension);
+    if ext_elements.keywords.len() > 2 {
```
raskad, line 701: This seems to differ from the spec no? Should this not just be `supported_extension.len() > 2`

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,566 @@ impl Intl {
...
+                            value = JsValue::String(JsString::new(requested_value));
+                            // ii. Let supportedExtensionAddition be the string-concatenation
+                            // of "-", key, "-", and value.
+                            supported_extension_addition = JsString::new("-");
+                            supported_extension_addition =
+                                JsString::concat(supported_extension_addition, &key);
+                            supported_extension_addition =
+                                JsString::concat(supported_extension_addition, JsString::new("-"));
+                            supported_extension_addition =
+                                JsString::concat(supported_extension_addition, requested_value);
```
raskad, line 625:
```suggestion
supported_extension_addition = JsString::concat_array(&["-", &key, "-", &requested_value]);
```

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,566 @@ impl Intl {
...
+                        }
+                    }
+                // 4. Else if keyLocaleData contains "true", then
+                } else if js_arr_contains_str(&key_locale_data, "true", context) {
+                    // a. Let value be "true".
+                    value = JsValue::String(JsString::new("true"));
+                    // b. Let supportedExtensionAddition be the string-concatenation of "-" and key.
+                    supported_extension_addition = JsString::new("-");
+                    supported_extension_addition =
+                        JsString::concat(supported_extension_addition, &key);
```
raskad, line 636: Also use `JsString::concat_array` here

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,566 @@ impl Intl {
...
+
+    // TODO 7. Assert: ! IsStructurallyValidLanguageTag(locale) is true.
+    // 8. Return ! CanonicalizeUnicodeLocaleId(locale).
+    Intl::canonicalize_locale(&new_locale)
+}
+
+#[derive(Debug)]
+pub struct ResolveLocaleRecord {
+    pub locale: JsString,
+    pub properties: HashMap<JsString, JsValue>,
```
raskad, line 489: All of the `HashMap`s should be `FxHashMap`s. We use those in boa because they are significantly faster.

raskad, line 489: I think all of the `properties` keys in these three structs represent `[[LocaleData]]` slots from the spec right? I think they can be more strict to be easier to work with. The spec defines the slots here: https://tc39.es/ecma402/#sec-internal-slots.
I think `[[LocaleData]]` could probably be `FxHashMap<JsString, FxHashMap<JsString, Vec<JsString>>>`. That should also make the `js_arr_contains_str` function obsolete and remove all of the JsValue and JsObject usage.

raskad (changes requested):

raskad:
> @raskad do I need to change '&str' to '&JsString' as well?

In theory I think yes. But think we can probably fix that later. Currently the most correct thing we could do is to use `core::str::encode_utf16` everywhere. For example any `(String|str).len()` usage is in theory wrong, because it uses utf8. We would need to use `(String|str).encode_utf16().count()` to get the utf16 length.

But I think this is probably not a top priority for this right now. Let's first get some working code, get some 262 tests to pass and then we can go from there ;)

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,566 @@ impl Intl {
...
+    let mut candidate = locale.clone();
+    // 2. Repeat
+    loop {
+        // a. If availableLocales contains an element equal to candidate, return candidate.
+        if available_locales
+            .iter()
+            .any(|loc_name| candidate.eq(loc_name))
+        {
+            return candidate;
+        }
```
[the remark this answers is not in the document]
NorbertGarfield: https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R215

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,566 @@ impl Intl {
...
+                let prev_dash = tmp_candidate.rfind('-').unwrap_or(ind);
+                let trim_ind = if ind >= 2 && prev_dash == ind - 2 {
+                    ind - 2
+                } else {
+                    ind
+                };
+                // d. Let candidate be the substring of candidate from position 0, inclusive, to position pos, exclusive.
+                candidate = JsString::new(&candidate[..trim_ind]);
+            }
+            None => return JsString::new("undefined"),
```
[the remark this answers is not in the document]
NorbertGarfield: https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R234

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,566 @@ impl Intl {
...
+#[derive(Debug)]
+pub struct LocaleDataRecord {
+    pub properties: HashMap<JsString, JsValue>,
+}
+
+#[derive(Debug)]
+pub struct DateTimeFormatRecord {
+    pub locale_matcher: JsString,
+    pub properties: HashMap<JsString, JsValue>,
+}
```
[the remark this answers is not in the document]
NorbertGarfield: https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R465

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: I misinterpreted the specification. Fixed https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R647

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R577

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R584

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R584

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R467

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,512 @@ impl Intl {
...
+                // algorithm steps to replace Unicode extension values with their canonical
+                // form per Unicode Technical Standard #35 LDML § 3.2.1 Canonical Unicode
+                // Locale Identifiers, treating key as ukey and optionsValue as uvalue
+                // productions.
+
+                // 3. If optionsValue is the empty String, then
+                let options_val_str = options_value
+                    .to_string(context)
+                    .unwrap_or_else(|_| JsString::new(""))
+                    .to_string();
```
raskad, line 616: I think you should be able to do `if let Some(options_val_str) = options_value.as_string()` here.

boa_engine/src/builtins/intl/mod.rs
raskad: Yeah true that seems to be a value according to spec. Let's leave it at this and maybe we can figure out optimizations when we have more implemented and directly see how the code works together.

raskad (changes requested):
Just got my one comment left. Really nice work. Thank you for taking this on!

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: [Applied](https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R613).

raskad (approved):
Thank you, really nice work! I'm really exited for the first intl test to pass ;)

boa_engine/src/builtins/intl/tests.rs
```diff
@@ -0,0 +1,329 @@
...
+    assert_eq!(
+        locale_record.data_locale,
+        crate::builtins::intl::default_locale()
+    );
+    assert_eq!(locale_record.properties.is_empty(), true);
+
+    // available: [es-ES], requested: [es-ES]
+    let available_locales = vec![JsString::new("es-ES")];
+    let requested_locales = vec![JsString::new("es-ES")];
+    let relevant_extension_keys: Vec<JsString> = vec![];
```
Razican, line 283: While they are the same, macros require expansion, so they increase compile times. Also, using the generic in the function call reduces the duplication of code :)

```suggestion
    let relevant_extension_keys = Vec::<JsString>::new();
```

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -144,3 +147,540 @@ impl Intl {
...
+/// Trims unciode locale extensions from a given string if any.
+///
+/// For example:
+///
+/// - `ja-Jpan-JP-u-ca-japanese-hc-h12` becomes `ja-Jpan-JP`
+/// - `fr-FR` becomes `fr-FR`
+fn trim_unicode_extensions(requested_locale: &str) -> JsString {
+    let trim_pos = get_leftmost_unicode_extension_pos(requested_locale);
+    JsString::new(&requested_locale[..trim_pos])
+}
```
jedel1043, line 182: Would it be possible to replace this with the `icu` crate?
IIRC it has functionality to parse and manipulate locale strings.

NorbertGarfield:
@jedel1043 as far, as I understood, it is possible to use locale parsing, like specified [here](https://docs.rs/icu/latest/icu/locid/struct.Locale.html#examples), and then serialize language identifier and extensions to string. I would like to clarify one thing here. My understanding is that icu crate integration is still in progress.

Would you recommend including icu crate (to replace locale extension related methods) into this PR?

Thanks.

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: @jedel1043 as far, as I understood, it is possible to use locale parsing, like specified [here](https://docs.rs/icu/latest/icu/locid/struct.Locale.html#examples), and then serialize language identifier and extensions to string. I would like to clarify one thing here. My understanding is that icu crate integration is still in progress.

Would you recommend including icu crate (to replace locale extension related methods) into this PR?

Thanks.

NorbertGarfield: [Done](https://github.com/boa-dev/boa/pull/2036/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R230). Please review.

HalidOdat (approved):
Looks good to me! :)

Razican (approved):
It looks great now!

PR #2042, opened 2022-04-23
[Merged by Bors] - Improve `JsString` performance

YXL76:
It changes the following:

- The current `JsString` implementation has a lot of duplicate heap allocations. For static strings, `JsString` only needs to store the index of `CONSTANTS_ARRAY`. ~~I let `JsString` can store pointer to heap or static area. The two are distinguished by the first bit. (This implementation is rough, maybe we should put flag into `Inner`, like [arcstr](https://github.com/thomcc/arcstr/blob/70ba2fac19d3efe8d2e0231daaf74f9987c04b8a/src/arc_str.rs#L751-L757))~~
- I also added more string constants, which are always used.


YXL76:
At present, the `Rc` feature of `JsString` is very limited, just like the [function](https://github.com/boa-dev/boa/blob/5b46d1275c7abab317a5d3d96024beb838fa6ac6/boa_engine/src/object/mod.rs#L1603-1630), it always performs two heap allocations instead of increasing the reference count.

Since we always do [hashing](https://github.com/boa-dev/boa/blob/5b46d1275c7abab317a5d3d96024beb838fa6ac6/boa_engine/src/string.rs#L340-L344), wouldn't it be better to use `get_or_insert_with` instead of `get`?

YXL76:
Simplify it, now `inner` only stores the index of `CONSTANTS_ARRAY`

raskad (commented):
The principle used here is a [tagged pointer](https://en.wikipedia.org/wiki/Tagged_pointer) right? There should be an explicit mention of that in the comments, so that everyone knows what is being done.

I'm not too familiar with the platform requirements for tagged pointers. I assume this would fail on 16 bit platforms? Are there any other requirements that should be noted?
I think if we implement this, we would have to at least mention these requirements, at best prevent boa from compiling if any of the requirements are not met.

jedel1043 (changes requested):
I like the idea of this PR very much! However, I think it would be better to reuse `Inner` to be either a pointer to heap data or a `&'static str` pointer.
We could decompose the `str` pointer into its raw parts, then reuse `Inner.len` and `Inner.data` to hold that information.
Next, we would change `Inner.refcount` to be a `Cell<Option<NonZeroUSize>>`, having `None` represent a static `str` and `Some(count)` represent a heap allocated `str`, being `NonZeroUSize` so that the compiler can optimize the size of the data.
Finally, we just need to reconstruct the original `&'static str` pointer on access with `Inner.len` and `Inner.data`.

jedel1043:
Also, it would be good to check out if a trie (https://crates.io/crates/radix_trie) could be useful

boa_engine/src/value/mod.rs
```diff
@@ -278,7 +278,6 @@ impl JsValue {
     /// [spec]: https://tc39.es/ecma262/#sec-toboolean
     pub fn to_boolean(&self) -> bool {
         match *self {
-            Self::Undefined | Self::Null => false,
```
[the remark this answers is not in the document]
YXL76: It was emitted by clippy.

boa_engine/src/bigint.rs
```diff
@@ -195,7 +195,7 @@ impl JsBigInt {
             let inner = if n > 0 {
                 x.inner.as_ref().clone().shl(n as usize)
             } else {
-                x.inner.as_ref().clone().shr(n.abs() as usize)
+                x.inner.as_ref().clone().shr(n.unsigned_abs())
```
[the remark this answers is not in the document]
YXL76: It was emitted by clippy.

boa_engine/src/bigint.rs
```diff
@@ -180,7 +180,7 @@ impl JsBigInt {
             let inner = if n > 0 {
                 x.inner.as_ref().clone().shr(n as usize)
             } else {
-                x.inner.as_ref().clone().shl(n.abs() as usize)
+                x.inner.as_ref().clone().shl(n.unsigned_abs())
```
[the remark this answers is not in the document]
YXL76: It was emitted by clippy.

YXL76:
> The principle used here is a [tagged pointer](https://en.wikipedia.org/wiki/Tagged_pointer) right? There should be an explicit mention of that in the comments, so that everyone knows what is being done.
> 
> I'm not too familiar with the platform requirements for tagged pointers. I assume this would fail on 16 bit platforms? Are there any other requirements that should be noted? I think if we implement this, we would have to at least mention these requirements, at best prevent boa from compiling if any of the requirements are not met.

We can emit a compile error if the platform does not support it, so far there is no side effect of introducing this technique, since `rust-gc` also uses this technique.

https://github.com/Manishearth/rust-gc/blob/0278c6b895d53153a4eb63cfc56e59d97698e5ed/gc/src/lib.rs#L102-L107

https://github.com/Manishearth/rust-gc/blob/0278c6b895d53153a4eb63cfc56e59d97698e5ed/gc/src/lib.rs#L114-L118

boa_engine/src/string.rs
```diff
@@ -314,18 +617,81 @@ impl Inner {
...
+        self.0.get() as *mut _
+    }
+
+    /// # Safety
+    ///
+    /// It maybe a string allocated on the heap.
+    #[inline]
+    const unsafe fn get_static_unchecked(self) -> &'static str {
+        // shift right to get the index.
+        CONSTANTS_ARRAY[self.0.get() >> 1]
```
HalidOdat, line 666:
```suggestion
        CONSTANTS_ARRAY.get_unchecked(self.0.get() >> 1)
```
We should use `get_unchecked(index)`, since the index operator always does a boundary check.

YXL76:
A larger `MAX_CONSTANT_STRING_LENGTH` may have some side effects, we have to do more hash table queries, and the hit rate is low.

This goes back to my previous question, should `CONSTANTS` be a fixed size HashSet? If so, then we should strictly limit the size of `MAX_CONSTANT_STRING_LENGTH`, if not (ie as a global cache), then there are no such problems.

raskad:
> A larger `MAX_CONSTANT_STRING_LENGTH` may have some side effects, we have to do more hash table queries, and the hit rate is low.

I think this is a performance property that we would have to adjust based on real-word benchmarks. In theory a longer `MAX_CONSTANT_STRING_LENGTH` leads to more lookups at `JsString` creation time, but how does that impact the performance of the runtime overall? Do real-world programs benefit more from static strings or from less hashmap lookups?

Maybe there is some research already done in other engines that we could base a decision upon, but I think for now this is totally fine.
 
> This goes back to my previous question, should `CONSTANTS` be a fixed size HashSet? If so, then we should strictly limit the size of `MAX_CONSTANT_STRING_LENGTH`, if not (ie as a global cache), then there are no such problems.

I think inserting into `CONSTANTS` is not realistic. We either leak memory by never deleting those strings or would have to implement reference counting in `CONSTANTS` to make removing from `CONSTANTS` possible which seems like a bad idea for dealloc performance.

jedel1043:
As expected, running `MIRIFLAGS="-Zmiri-strict-provenance -Zmiri-disable-stacked-borrows" cargo +nightly miri test --package boa_engine --lib -- string::tests --skip builtins` errors with the message:

`error: Undefined Behavior: dereferencing pointer failed: 0xdb4690 is not a valid pointer`

Meaning the provenance of the pointer is not preserved. The fix is pretty easy, we just need to use `NonNull<Inner>` instead of `NonZeroUsize`, casting the tagged indexes to invalid pointers with `index as *const Inner`. That should preserve the provenance of every valid pointer.

Razican:
# Benchmarks

  <details>
    <summary>Click to view benchmark</summary>

|                  name                      |     baseDuration   |   changesDuration  | difference |
|:------------------------------------------:|:------------------:|:------------------:|:----------:|
|      Arithmetic operations (Compiler)      |  **493.7±1.24ns**  |    526.0±0.90ns    |   +7.0 %   |
|     Arithmetic operations (Execution)      |  **445.5±1.31ns**  |    487.7±1.37ns    |   +9.0 %   |
|       Arithmetic operations (Parser)       |   **6.1±0.02µs**   |     6.2±0.04µs     |   +1.0 %   |
|          Array access (Compiler)           | **1338.8±5.43ns**  |   1374.3±6.68ns    |   +3.0 %   |
|          Array access (Execution)          |     8.5±0.02µs     |     8.5±0.08µs     |   0.0  %   |
|           Array access (Parser)            |  **13.9±0.04µs**   |    14.1±0.06µs     |   +2.0 %   |
|         Array creation (Compiler)          |   1961.4±19.27ns   | **1945.9±6.06ns**  |  -0.99 %   |
|         Array creation (Execution)         |   **2.4±0.00ms**   |     2.5±0.00ms     |   +3.0 %   |
|          Array creation (Parser)           |    15.7±0.04µs     |    15.8±0.22µs     |   0.0  %   |
|            Array pop (Compiler)            |   **3.7±0.01µs**   |     3.8±0.01µs     |   +4.0 %   |
|           Array pop (Execution)            | **1075.4±13.56µs** |   1107.8±11.13µs   |   +3.0 %   |
|             Array pop (Parser)             |    157.2±0.25µs    |    157.2±0.18µs    |   0.0  %   |
|      Boolean Object Access (Compiler)      | **1110.1±5.17ns**  |   1126.9±3.66ns    |   +2.0 %   |
|     Boolean Object Access (Execution)      |     4.6±0.02µs     |   **4.5±0.01µs**   |   -2.0 %   |
|       Boolean Object Access (Parser)       |  **16.2±0.03µs**   |    16.7±0.06µs     |   +3.0 %   |
|            Clean js (Compiler)             |     3.7±0.01µs     |     3.7±0.01µs     |   0.0  %   |
|            Clean js (Execution)            |    703.3±6.39µs    |  **681.9±3.80µs**  |   -2.9 %   |
|             Clean js (Parser)              |  **33.8±0.21µs**   |    34.1±0.05µs     |   +1.0 %   |
|                Create Realm                |  **265.1±0.37ns**  |    273.5±1.32ns    |   +3.0 %   |
| Dynamic Object Property Access (Compiler)  | **1752.9±6.21ns**  |   1769.6±4.42ns    |   +1.0 %   |
| Dynamic Object Property Access (Execution) |     5.7±0.04µs     |   **5.6±0.01µs**   |  -0.99 %   |
|  Dynamic Object Property Access (Parser)   |  **12.1±0.06µs**   |    12.3±0.08µs     |   +2.0 %   |
|            Fibonacci (Compiler)            |     2.4±0.00µs     |     2.4±0.00µs     |   0.0  %   |
|           Fibonacci (Execution)            | **1400.4±2.33µs**  |   1445.6±14.19µs   |   +3.0 %   |
|             Fibonacci (Parser)             |  **18.5±0.07µs**   |    19.0±0.03µs     |   +3.0 %   |
|            For loop (Compiler)             |   **2.1±0.00µs**   |     2.2±0.03µs     |   +5.0 %   |
|            For loop (Execution)            |  **15.2±0.06µs**   |    15.4±0.07µs     |   +1.0 %   |
|             For loop (Parser)              |    16.0±0.05µs     |    16.1±0.05µs     |   0.0  %   |
|             Mini js (Compiler)             |     3.6±0.01µs     |   **3.5±0.01µs**   |   -2.0 %   |
|            Mini js (Execution)             |    657.5±4.75µs    |  **649.1±4.81µs**  |  -0.99 %   |
|              Mini js (Parser)              |  **30.0±0.12µs**   |    30.4±0.05µs     |   +1.0 %   |
|      Number Object Access (Compiler)       | **1029.0±2.34ns**  |   1060.2±4.30ns    |   +3.0 %   |
|      Number Object Access (Execution)      |   **3.5±0.02µs**   |     3.6±0.01µs     |   +2.0 %   |
|       Number Object Access (Parser)        |  **12.8±0.11µs**   |    13.0±0.04µs     |   +2.0 %   |
|         Object Creation (Compiler)         | **1434.7±3.62ns**  |   1492.4±4.71ns    |   +4.0 %   |
|        Object Creation (Execution)         |     5.2±0.02µs     |   **5.2±0.07µs**   |  -0.99 %   |
|          Object Creation (Parser)          |  **10.6±0.06µs**   |    10.8±0.05µs     |   +2.0 %   |
|             RegExp (Compiler)              |   1725.3±3.85ns    | **1691.2±29.10ns** |   -2.0 %   |
|             RegExp (Execution)             |  **11.7±0.06µs**   |    11.9±0.06µs     |   +1.0 %   |
|              RegExp (Parser)               |  **11.8±0.25µs**   |    11.9±0.06µs     |   +1.0 %   |
|         RegExp Creation (Compiler)         |   1531.4±3.00ns    | **1502.1±5.63ns**  |   -2.0 %   |
|        RegExp Creation (Execution)         |     8.8±0.02µs     |     8.8±0.04µs     |   0.0  %   |
|          RegExp Creation (Parser)          |   **9.7±0.11µs**   |     9.8±0.03µs     |   +2.0 %   |
|         RegExp Literal (Compiler)          |   1754.2±38.01ns   | **1684.2±15.68ns** |   -3.8 %   |
|         RegExp Literal (Execution)         |  **11.6±0.05µs**   |    11.9±0.06µs     |   +2.0 %   |
|          RegExp Literal (Parser)           |   **9.5±0.04µs**   |     9.7±0.02µs     |   +2.0 %   |
|     RegExp Literal Creation (Compiler)     |   1510.7±10.96ns   |   1512.1±5.35ns    |   0.0  %   |
|    RegExp Literal Creation (Execution)     |     8.9±0.03µs     |   **8.8±0.03µs**   |  -0.99 %   |
|      RegExp Literal Creation (Parser)      |   **7.5±0.05µs**   |     7.7±0.02µs     |   +2.0 %   |
|  Static Object Property Access (Compiler)  |   1486.9±4.70ns    | **1470.8±4.04ns**  |  -0.99 %   |
| Static Object Property Access (Execution)  |     5.4±0.06µs     |   **5.3±0.02µs**   |  -0.99 %   |
|   Static Object Property Access (Parser)   |  **11.4±0.06µs**   |    11.5±0.06µs     |   +1.0 %   |
|      String Object Access (Compiler)       |   1550.9±23.24ns   |   1546.9±6.29ns    |   0.0  %   |
|      String Object Access (Execution)      |   **6.1±0.02µs**   |     6.7±0.13µs     |   +9.0 %   |
|       String Object Access (Parser)        |  **15.9±0.02µs**   |    16.5±0.02µs     |   +4.0 %   |
|        String comparison (Compiler)        |     2.2±0.01µs     |   **2.2±0.01µs**   |   -3.8 %   |
|       String comparison (Execution)        |   **4.7±0.03µs**   |     4.8±0.06µs     |   +2.0 %   |
|         String comparison (Parser)         |  **12.5±0.04µs**   |    12.8±0.04µs     |   +3.0 %   |
|      String concatenation (Compiler)       |   1735.4±4.99ns    | **1673.5±3.89ns**  |   -3.8 %   |
|      String concatenation (Execution)      |     4.6±0.03µs     |     4.6±0.04µs     |   0.0  %   |
|       String concatenation (Parser)        |   **8.7±0.03µs**   |     8.9±0.01µs     |   +2.0 %   |
|           String copy (Compiler)           |   1366.5±5.89ns    | **1339.6±5.52ns**  |   -2.0 %   |
|          String copy (Execution)           |   **4.3±0.02µs**   |     4.3±0.04µs     |   +2.0 %   |
|            String copy (Parser)            |   **6.5±0.03µs**   |     6.7±0.03µs     |   +3.0 %   |
|             Symbols (Compiler)             |  **941.6±3.70ns**  |    983.6±3.43ns    |   +4.0 %   |
|            Symbols (Execution)             |     4.4±0.01µs     |   **4.3±0.03µs**   |   -2.0 %   |
|              Symbols (Parser)              |   **5.0±0.02µs**   |     5.2±0.02µs     |   +4.0 %   |

  </details>

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
 }
 
+/// It may be an index of [`CONSTANTS_ARRAY`], or a raw pointer of [`Inner`]. Use
+/// the first bit as the flag (Detail: <https://en.wikipedia.org/wiki/Tagged_pointer>).
+///
+/// When the first bit is 0, it represents the address of an [`Inner`]. When the
+/// first bit is 1, it represents an index of [`CONSTANTS_ARRAY`], and the index
+/// number is stored in higher bits.
+///
+/// It uses `NonNull`, which can get benefit from non-null optimization.
```
jedel1043, line 638:
```suggestion
/// This struct uses a technique called tagged pointer to benefit from the fact that newly
/// allocated pointers are always word aligned on 64-bits platforms, making it impossible
/// to have a LSB equal to 1. More details about this technique on the article of Wikipedia
/// about [tagged pointers][tagged_wp].
///
/// # Representation
///
/// If the LSB of the internal [`NonNull<Inner>`] is set (1), then the pointer address represents
/// an index value for [`CONSTANTS_ARRAY`], where the remaining MSBs store the index.
/// Otherwise, the whole pointer represents the address of a heap allocated [`Inner`].
///
/// It uses [`NonNull`], which guarantees that `Flag` (and subsequently [`JsString`])
/// can use the "null pointer optimization" to optimize the size of [`Option<Flag>`].
///
/// # Provenance
///
/// This struct stores a [`NonNull<Inner>`] instead of a [`NonZeroUsize`][std::num::NonZeroUsize]
/// in order to preserve the provenance of our valid heap pointers.
/// On the other hand, all index values are just casted to invalid pointers,
/// because we don't need to preserve the provenance of [`usize`] indices.
///
/// [tagged_wp]: https://en.wikipedia.org/wiki/Tagged_pointer
```

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+struct Flag(NonNull<Inner>);
+
+impl Flag {
+    #[inline]
+    fn new_heap(inner: NonNull<Inner>) -> Self {
+        Self(inner)
+    }
+
+    /// Set the first bit to 1, indicating that it is static. Store the index
+    /// in 1..63 bits.
```
jedel1043, line 650:
```suggestion
    /// Create a new static `Flag` from the index of an element inside [`CONSTANTS_ARRAY`].
```

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+    #[inline]
+    const fn new_static(idx: usize) -> Self {
+        // Safety: We already know it's not null, so this is safe.
+        Self(unsafe { NonNull::new_unchecked(((idx << 1) | 1) as *mut _) })
+    }
+
+    /// Check if the first bit is 1.
```
jedel1043, line 657:
```suggestion
    /// Check if `Flag` contains an index for [`CONSTANTS_ARRAY`].
```

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+    fn is_static(self) -> bool {
+        (self.0.as_ptr() as usize) & 1 == 1
+    }
+
+    /// Returns a reference to a string stroed on the heap, without doing
+    /// flag checking.
+    ///
+    /// # Safety
+    ///
+    /// It maybe a static string.
```
jedel1043, line 668:
```suggestion
    /// Returns a reference to a string stored on the heap,
    /// without checking if its internal pointer is valid.
    ///
    /// # Safety
    ///
    /// Calling this method with a static `Flag` results in Undefined Behaviour.
```

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+    #[inline]
+    const unsafe fn get_heap_unchecked(self) -> NonNull<Inner> {
+        self.0
+    }
+
+    /// Returns a reference to a static string, without doing flag checking.
+    ///
+    /// # Safety
+    ///
+    /// It maybe a string stroed on the heap.
```
jedel1043, line 678:
```suggestion
    /// Returns the string inside [`CONSTANTS_ARRAY`] corresponding to the
    /// index inside `Flag`, without checking its validity.
    ///
    /// # Safety
    ///
    /// Calling this method with a `Flag` storing an out of bounds index for
    /// [`CONSTANTS_ARRAY`] or a valid pointer to a heap allocated [`Inner`]
    /// results in Undefined Behaviour.
```

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+
 impl Default for JsString {
     #[inline]
     fn default() -> Self {
-        Self::new("")
+        Self::new_static(0)
     }
 }
 
+/// Data stored in [`JsString`].
```
jedel1043, line 693:
```suggestion
/// Enum representing either a reference to a heap allocated [`Inner`]
/// or a static reference to a [`str`] inside [`CONSTANTS_ARRAY`].
```

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+    // A string allocated on the heap.
+    Heap(&'a Inner),
+    // A static string slice.
+    Static(&'static str),
+}
+
 impl JsString {
+    /// Create a new JavaScript string from an index of [`CONSTANTS_ARRAY`].
+    #[inline]
+    fn new_static(idx: usize) -> Self {
```
jedel1043, line 704: This is not a safe function. It should be marked as `unsafe`.

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+impl Flag {
+    #[inline]
+    fn new_heap(inner: NonNull<Inner>) -> Self {
+        Self(inner)
+    }
+
+    /// Set the first bit to 1, indicating that it is static. Store the index
+    /// in 1..63 bits.
+    #[inline]
+    const fn new_static(idx: usize) -> Self {
```
jedel1043, line 652: This is not a safe function. It should be marked as `unsafe`.

boa_engine/src/string.rs
```diff
@@ -312,20 +618,96 @@ impl Inner {
...
+/// number is stored in higher bits.
+///
+/// It uses `NonNull`, which can get benefit from non-null optimization.
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+struct Flag(NonNull<Inner>);
+
+impl Flag {
+    #[inline]
+    fn new_heap(inner: NonNull<Inner>) -> Self {
```
jedel1043, line 645: This is not a safe function. It should be marked as unsafe.

jedel1043 (changes requested):
Some more comments about the implementation. I also tinkered a bit with the documentation to make it a bit more clearer.

boa_engine/src/string.rs
```diff
@@ -183,13 +491,11 @@ unsafe fn try_alloc(layout: Layout) -> *mut u8 {
...
+        let len = CONSTANTS_ARRAY.len();
+        let mut constants = FxHashSet::with_capacity_and_hasher(len, BuildHasherDefault::<FxHasher>::default());
 
-        for s in CONSTANTS_ARRAY.iter() {
-            let s = JsString {
-                inner: Inner::new(s),
-                _marker: PhantomData,
-            };
+        for idx in 0..len {
+            let s = JsString::new_static(idx);
```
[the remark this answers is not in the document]
YXL76: It's good, but I can't find a good way to implement this.

boa_engine/src/string.rs
```diff
@@ -358,57 +766,71 @@ impl JsString {
...
     }
 
     /// Gets the number of `JsString`s which point to this allocation.
     #[inline]
     pub fn refcount(this: &Self) -> usize {
-        this.inner().refcount.get()
+        match this.inner() {
+            InnerKind::Heap(inner) => inner.refcount.get(),
+            InnerKind::Static(_inner) => 0,
+        }
```
jedel1043, line 833: This should probably return `Option<usize>` instead, since it doesn't really makes sense to return 0 on static strings.

jedel1043 (changes requested):
Nice work! There are still some nitpicks I have, but it's looking pretty good nonetheless :)

boa_engine/src/string.rs
YXL76: The most appropriate type is `Option<NonZeroUsize>`, but I think we don't need it.

jedel1043 (approved):
Looks good to me! Let's wait until the ones who left comments approve 😄

boa_engine/src/string.rs
```diff
@@ -358,57 +767,71 @@ impl JsString {
         let x = x.as_ref();
         let y = y.as_ref();
 
-        let this = Self {
-            inner: Inner::concat_array(&[x, y]),
-            _marker: PhantomData,
-        };
+        let inner = Inner::concat_array(&[x, y]);
```
HalidOdat, line 770: There is a memory leak here. If the concatenated string is in the `CONSTANTS` (like `"Obj" + "ect"`) it will leak the heap allocated string (since no destructor is called) and return the static string.

This applies to `JsString::concat_array()`

HalidOdat (changes requested):

HalidOdat (approved):
Looks good to me! :)

raskad (approved):

PR #2044, opened 2022-04-23
[Merged by Bors] - Allow `let` as variable declaration name

raskad:

- Allow `let` as variable declaration name


Razican (approved):
Thanks!!

HalidOdat (approved):

pedropaulosuzuki:
Is there a specific test for this? Or are we using Test262 in general as our testing methods? While I see the value on using Test262, it is also good to use our in-house tests in order to catch regressions.

raskad:
> Is there a specific test for this? Or are we using Test262 in general as our testing methods? While I see the value on using Test262, it is also good to use our in-house tests in order to catch regressions.

Unless there are implementation specific tests that are not covered by the 262 suite, I think replicating the 262 tests does not make much sense. We have to run the 262 suite anyways, so we would mostly just double the test time. Also whenever the 262 suite changes, we would have to adjust our internal test suite which would probably result in the two suites diverging. But that's just my 2 cents.

PR #2069, opened 2022-05-11
[Merged by Bors] - Remove `strict` flag from `Context`

raskad:
The `Context` currently contains a `strict` flag that indicates is global strict mode is active. This is redundant to the strict flag that is set on every function and causes some non spec compliant situations. This pull request removes the strict flag from `Context` and fixes some resulting errors.

Detailed changes:

- Remove strict flag from `Context`
- Make 262 tester compliant with the strict section in [test262/INTERPRETING.md](https://github.com/tc39/test262/blob/2e7cdfbe18eae4309677033673bb4b5ac6b1de40/INTERPRETING.md#strict-mode)
- Make 262 tester compliant with the `raw` flag in [test262/INTERPRETING.md](https://github.com/tc39/test262/blob/2e7cdfbe18eae4309677033673bb4b5ac6b1de40/INTERPRETING.md#flags)
- Allow function declarations in strict mode
- Fix parser flag propagation for classes
- Move some early errors from the lexer to the parser
- Add / fix some early errors for 'arguments' and 'eval' identifier usage in strict mode
- Refactor `ArrayLiteral` parser for readability and correct early errors

### DIFF
--- a/boa_engine/src/syntax/ast/node/declaration/mod.rs
+++ b/boa_engine/src/syntax/ast/node/declaration/mod.rs
@@ -276,6 +276,114 @@ impl DeclarationPattern {
             DeclarationPattern::Array(pattern) => pattern.init(),
         }
     }
+
+    /// Returns true if the node contains a identifier reference named 'arguments'.
+    ///
+    /// More information:
+    ///  - [ECMAScript specification][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-static-semantics-containsarguments
+    #[inline]
+    pub(crate) fn contains_arguments(&self) -> bool {
+        match self {
+            DeclarationPattern::Object(pattern) => {
+                if let Some(init) = pattern.init() {
+                    if init.contains_arguments() {
+                        return true;
+                    }
+                }
+                for binding in pattern.bindings() {
+                    match binding {
+                        BindingPatternTypeObject::SingleName {
+                            property_name,
+                            default_init,
+                            ..
+                        } => {
+                            if let PropertyName::Computed(node) = property_name {
+                                if node.contains_arguments() {
+                                    return true;
+                                }
+                            }
+                            if let Some(init) = default_init {
+                                if init.contains_arguments() {
+                                    return true;
+                                }
+                            }
+                        }
+                        BindingPatternTypeObject::RestGetConstField {
+                            get_const_field, ..
+                        } => {
+                            if get_const_field.obj().contains_arguments() {
+                                return true;
+                            }
+                        }
+                        BindingPatternTypeObject::BindingPattern {
+                            ident,
+                            pattern,
+                            default_init,
+                        } => {
+                            if let PropertyName::Computed(node) = ident {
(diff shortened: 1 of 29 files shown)


HalidOdat (approved):
Looks good to me! :)

boa_engine/src/syntax/ast/node/declaration/mod.rs
```diff
@@ -276,6 +276,114 @@ impl DeclarationPattern {
...
     }
+
+    /// Returns true if the node contains a identifier reference named 'arguments'.
+    ///
+    /// More information:
+    ///  - [ECMAScript specification][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-static-semantics-containsarguments
+    #[inline]
+    pub(crate) fn contains_arguments(&self) -> bool {
```
jedel1043, line 287: Would be good to make this its own trait, akin to `ToInternedString`, to delegate the work to each node type individually

jedel1043: Ah, maybe renaming `ToInternedString` to `AstNode` and creating a default implementation for `contains_arguments` would be better

raskad: I though about this too. We could do it, but I think spreading these functions over all ast node files would probably make it unnecessarily hard to read. Especially since it it very trivial functionality. Also the spec defines this as one function, so it is easier to compare to that.
What do you think?

jedel1043: Hm, yeah, it would be very unreadable to have parts of the same function through all the module

boa_engine/src/syntax/parser/expression/update.rs
```diff
@@ -57,56 +57,66 @@ where
...
+                                position,
+                            )));
+                        } else if ident.sym() == Sym::EVAL {
+                            return Err(ParseError::lex(LexError::Syntax(
+                                "unexpected identifier 'eval' in strict mode".into(),
+                                position,
+                            )));
+                        }
+                    }
+                }
```
jedel1043, line 84: This exact code is repeated twice. I think this should be a method in `ident`

jedel1043 (changes requested):

boa_engine/src/syntax/parser/expression/update.rs
raskad: Fixed & rebased

jedel1043 (approved):
Looks great! Good job!

jedel1043:
I suppose the new failing errors are cases where the tests passed because of missing features or so?

raskad:
> I suppose the new failing errors are cases where the tests passed because of missing features or so?

Exactly. I have a fix for those in the pipeline together with some other fixes.

jedel1043:
> > I suppose the new failing errors are cases where the tests passed because of missing features or so?
> 
> Exactly. I have a fix for those in the pipeline together with some other fixes.

Nice! Let's merge this then :)

Issue #1570, opened 2021-09-10
Several `Date` tests panic on Windows


jedel1043:
**Describe the bug**
Several tests related to the Date object panic on Windows

Panicking tests:

- test\annexB\built-ins\Date\prototype\setYear\year-number-absolute.js
```
thread 'main' panicked at 'TzSpecificLocalTimeToSystemTime failed with: The parameter is incorrect. (os error 87)', ...\chrono-0.4.19\src\sys\windows.rs:127:9
```
- test\annexB\built-ins\Date\prototype\setYear\year-number-relative.js
```
thread 'main' panicked at 'TzSpecificLocalTimeToSystemTime failed with: The parameter is incorrect. (os error 87)', ...\chrono-0.4.19\src\sys\windows.rs:127:9
```
- test\built-ins\Date\prototype\setFullYear\arg-year-to-number.js
```
thread 'main' panicked at 'TzSpecificLocalTimeToSystemTime failed with: The parameter is incorrect. (os error 87)', ...\chrono-0.4.19\src\sys\windows.rs:127:9
```
- test\built-ins\Date\prototype\toString\format.js
```
thread 'main' panicked at 'SystemTimeToFileTime failed with: The parameter is incorrect. (os error 87)', ...\chrono-0.4.19\src\sys\windows.rs:117:9
```
**Build environment:**

- OS: Windows
- Version: 10
- Target triple: x86_64-pc-windows-msvc
- Rustc version: rustc 1.56.0-nightly (ad02dc46b 2021-08-26)

**Additional context**

Investigating the origin of the panic I realized the first three tests call either `setFullYear` or `setYear`, so that's probably the origin of the panic.

hle0:
I can confirm this still happens in the latest version.

The line of code that triggers the panic seems to be an unsafe call in chrono:
```rust
call!(TzSpecificLocalTimeToSystemTime(0 as *mut _, &mut sys_time, &mut utc));
```
Here's the full trace.
```
thread 'main' panicked at 'TzSpecificLocalTimeToSystemTime failed with: The parameter is incorrect. (os error 87)', C:\Users\henry\.cargo\registry\src\github.com-1ecc6299db9ec823\chrono-0.4.19\src\sys\windows.rs:127:9
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b\/library\std\src\panicking.rs:515
[...]

PR #2072, opened 2022-05-14
Implement InitializeDateTimeFormat

### LINKED ISSUE #1562: Implement missing internationalization related methods
Probably follows from #1180.

This is the list of methods we are currently missing from builtin objects because we don't have an internationalization API yet:

### String

- [ ] String.prototype.localeCompare()
- [ ] String.prototype.toLocaleLowerCase()
- [ ] String.prototype.toLocaleUpperCase()

### Date

- [ ] Date.prototype.toLocaleDateString()
- [ ] Date.prototype.toLocaleTimeString()

### General

- [ ] ***.prototype.toLocaleString()

Please comment or edit this issue if you find any other method that depends on internationalization.

NorbertGarfield:
This Pull Request partially fixes #1562.

It changes the following:

- Adds IsValidTimeZoneName
- Adds CanonicalizeTimeZoneName
- Implements BasicFormatMatcher
- Implements BestFitFormatMatcher
- Implements InitializeDateTimeFormat

### DIFF
--- a/boa_engine/src/builtins/intl/date_time_format.rs
+++ b/boa_engine/src/builtins/intl/date_time_format.rs
@@ -298,7 +298,10 @@ pub(crate) fn is_terminal(opt: &JsValue, context: &mut Context) -> bool {
             return true;
         }
 
-        if option.chars().any(|character| !character.is_alphanumeric()) {
+        if option
+            .chars()
+            .any(|character| !character.is_ascii_alphanumeric())
+        {
             return true;
         }
     }
@@ -312,7 +315,7 @@ pub(crate) fn is_terminal(opt: &JsValue, context: &mut Context) -> bool {
 ///  - [ECMAScript reference][spec]
 ///
 /// [spec]: https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots
-fn build_locale_data(context: &Context) -> LocaleDataRecord {
+fn build_locale_data(available_locales: &[JsString]) -> LocaleDataRecord {
     let mut locale_data_entry = FxHashMap::default();
     let nu_values = vec![JsString::new("arab")];
     locale_data_entry.insert(JsString::new("nu"), nu_values);
@@ -332,8 +335,10 @@ fn build_locale_data(context: &Context) -> LocaleDataRecord {
     locale_data_entry.insert(JsString::new("hourCycle"), hour_cycle_values);
 
     let mut locale_data = FxHashMap::default();
-    let default_locale_str = default_locale(context.icu().locale_canonicalizer()).to_string();
-    locale_data.insert(JsString::new(default_locale_str), locale_data_entry);
+
+    for avail_locale in available_locales {
+        locale_data.insert(avail_locale.clone(), locale_data_entry.clone());
+    }
 
     locale_data
 }
@@ -649,27 +654,15 @@ fn build_date_time_components() -> Vec<DateTimeComponents> {
         },
         DateTimeComponents {
             property: JsString::new("hour"),
-            values: vec![
-                JsString::new("narrow"),
-                JsString::new("short"),
-                JsString::new("long"),
-            ],
+            values: vec![JsString::new("2-digit"), JsString::new("numeric")],
         },
         DateTimeComponents {
             property: JsString::new("minute"),
-            values: vec![
-                JsString::new("narrow"),
-                JsString::new("short"),
-                JsString::new("long"),
-            ],
+            values: vec![JsString::new("2-digit"), JsString::new("numeric")],
         },
         DateTimeComponents {
             property: JsString::new("second"),
-            values: vec![
-                JsString::new("narrow"),
-                JsString::new("short"),
-                JsString::new("long"),
-            ],
+            values: vec![JsString::new("2-digit"), JsString::new("numeric")],
         },
         DateTimeComponents {
             property: JsString::new("fractionalSecondDigits"),
@@ -1078,6 +1071,8 @@ fn initialize_date_time_format(
     // 1. Let requestedLocales be ? CanonicalizeLocaleList(locales).
     let locales_arr = if locales.is_undefined() {
         vec![JsValue::undefined()]
+    } else if locales.is_string() {
+        vec![locales.clone()]
     } else {
         let locales_obj = locales
             .to_object(context)
@@ -1204,9 +1199,9 @@ fn initialize_date_time_format(
     opt.properties.insert(JsString::new("hc"), hour_cycle);
 
     // 16. Let localeData be %DateTimeFormat%.[[LocaleData]].
-    let locale_data = build_locale_data(context);
-    let relevant_ext_keys = build_relevant_ext_keys();
     let available_locales = build_available_locales(context);
+    let locale_data = build_locale_data(&available_locales);
+    let relevant_ext_keys = build_relevant_ext_keys();
 
     // 17. Let r be ResolveLocale(%DateTimeFormat%.[[AvailableLocales]], requestedLocales, opt,
     // %DateTimeFormat%.[[RelevantExtensionKeys]], localeData).
--- a/boa_engine/src/builtins/intl/date_time_format.rs
+++ b/boa_engine/src/builtins/intl/date_time_format.rs
@@ -267,11 +267,7 @@ pub(crate) fn to_date_time_options(
 ///  - [Unicode LDML reference][spec]
 ///
 /// [spec]: https://www.unicode.org/reports/tr35/#Unicode_locale_identifier
-pub(crate) fn is_terminal(opt: &JsValue, context: &mut Context) -> bool {
-    let opt_str = opt
-        .to_string(context)
-        .unwrap_or_else(|_| JsString::empty())
-        .to_string();
+pub(crate) fn is_terminal(opt_str: &str) -> bool {
     if opt_str.is_empty() {
         return true;
     }
@@ -298,9 +294,9 @@ pub(crate) fn is_terminal(opt: &JsValue, context: &mut Context) -> bool {
             return true;
         }
 
-        if option
+        if !option
             .chars()
-            .any(|character| !character.is_ascii_alphanumeric())
+            .all(|character| character.is_ascii_alphanumeric())
         {
             return true;
         }
@@ -317,21 +313,21 @@ pub(crate) fn is_terminal(opt: &JsValue, context: &mut Context) -> bool {
 /// [spec]: https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots
 fn build_locale_data(available_locales: &[JsString]) -> LocaleDataRecord {
     let mut locale_data_entry = FxHashMap::default();
-    let nu_values = vec![JsString::new("arab")];
+    let nu_values = Vec::from([JsString::new("arab")]);
     locale_data_entry.insert(JsString::new("nu"), nu_values);
 
-    let hc_values = vec![
+    let hc_values = Vec::from([
         JsString::new("h11"),
         JsString::new("h12"),
         JsString::new("h23"),
         JsString::new("h24"),
-    ];
+    ]);
     locale_data_entry.insert(JsString::new("hc"), hc_values);
 
-    let ca_values = vec![JsString::new("gregory")];
+    let ca_values = Vec::from([JsString::new("gregory")]);
     locale_data_entry.insert(JsString::new("ca"), ca_values);
 
-    let hour_cycle_values = vec![JsString::new("h24")];
+    let hour_cycle_values = Vec::from([JsString::new("h24")]);
     locale_data_entry.insert(JsString::new("hourCycle"), hour_cycle_values);
 
     let mut locale_data = FxHashMap::default();
@@ -426,16 +422,13 @@ pub(crate) fn canonicalize_time_zone_name(time_zone: &JsString) -> JsString {
 
 /// Converts `hour_cycle_str` to `preferences::HourCycle`
 pub(crate) fn string_to_hour_cycle(hour_cycle_str: &JsString) -> preferences::HourCycle {
-    let mut string_to_hc_map = FxHashMap::<JsString, preferences::HourCycle>::default();
-    string_to_hc_map.insert(JsString::from("h11"), preferences::HourCycle::H11);
-    string_to_hc_map.insert(JsString::from("h12"), preferences::HourCycle::H12);
-    string_to_hc_map.insert(JsString::from("h23"), preferences::HourCycle::H23);
-    string_to_hc_map.insert(JsString::from("h24"), preferences::HourCycle::H24);
-
-    let hour_cycle = string_to_hc_map
-        .get(hour_cycle_str)
-        .expect("Invalid hour cycle");
-    *hour_cycle
+    match hour_cycle_str.as_str() {
+        "h11" => preferences::HourCycle::H11,
+        "h12" => preferences::HourCycle::H12,
+        "h23" => preferences::HourCycle::H23,
+        "h24" => preferences::HourCycle::H24,
+        _ => panic!("Invalid hour cycle"),
+    }
 }
 
 /// Converts `JsValue` to `length::Date`
@@ -449,15 +442,12 @@ pub(crate) fn value_to_date_style(
     let date_style_str = date_style_val
         .to_string(context)
         .unwrap_or_else(|_| JsString::empty());
-    let mut string_to_style_map = FxHashMap::<JsString, Option<length::Date>>::default();
-    string_to_style_map.insert(JsString::from("full"), Some(length::Date::Full));
-    string_to_style_map.insert(JsString::from("long"), Some(length::Date::Long));
-    string_to_style_map.insert(JsString::from("medium"), Some(length::Date::Medium));
-    string_to_style_map.insert(JsString::from("short"), Some(length::Date::Short));
-
-    match string_to_style_map.get(&date_style_str) {
-        Some(date_style) => *date_style,
-        None => None,
+    match date_style_str.as_str() {
+        "full" => Some(length::Date::Full),
+        "long" => Some(length::Date::Long),
+        "medium" => Some(length::Date::Medium),
+        "short" => Some(length::Date::Short),
+        _ => None,
     }
 }
 
@@ -472,15 +462,12 @@ pub(crate) fn value_to_time_style(
     let time_style_str = time_style_val
         .to_string(context)
         .unwrap_or_else(|_| JsString::empty());
-    let mut string_to_style_map = FxHashMap::<JsString, Option<length::Time>>::default();
-    string_to_style_map.insert(JsString::from("full"), Some(length::Time::Full));
-    string_to_style_map.insert(JsString::from("long"), Some(length::Time::Long));
-    string_to_style_map.insert(JsString::from("medium"), Some(length::Time::Medium));
-    string_to_style_map.insert(JsString::from("short"), Some(length::Time::Short));
-
-    match string_to_style_map.get(&time_style_str) {
-        Some(time_style) => *time_style,
-        None => None,
+    match time_style_str.as_str() {
+        "full" => Some(length::Time::Full),
+        "long" => Some(length::Time::Long),
+        "medium" => Some(length::Time::Medium),
+        "short" => Some(length::Time::Short),
+        _ => None,
     }
 }
 
@@ -569,26 +556,17 @@ pub(crate) fn time_zone_to_value(maybe_tz: Option<components::TimeZoneName>) -> 
 
 /// Fetches field with name `property` from `format` bag
 fn get_format_field(format: &components::Bag, property: &str) -> JsValue {
-    if property == "weekday" {
-        text_to_value(format.weekday)
-    } else if property == "era" {
-        text_to_value(format.era)
-    } else if property == "year" {
-        year_to_value(format.year)
-    } else if property == "month" {
-        month_to_value(format.month)
-    } else if property == "day" {
-        day_to_value(format.day)
-    } else if property == "hour" {
-        numeric_to_value(format.hour)
-    } else if property == "minute" {
-        numeric_to_value(format.minute)
-    } else if property == "second" {
-        numeric_to_value(format.second)
-    } else if property == "timeZoneName" {
-        time_zone_to_value(format.time_zone_name)
-    } else {
-        JsValue::undefined()
+    match property {
+        "weekday" => text_to_value(format.weekday),
+        "era" => text_to_value(format.era),
+        "year" => year_to_value(format.year),
+        "month" => month_to_value(format.month),
+        "day" => day_to_value(format.day),
+        "hour" => numeric_to_value(format.hour),
+        "minute" => numeric_to_value(format.minute),
+        "second" => numeric_to_value(format.second),
+        "timeZoneName" => time_zone_to_value(format.time_zone_name),
+        _ => JsValue::undefined(),
     }
 }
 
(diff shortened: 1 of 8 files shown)


raskad:
#### VM implementation
  | Total | 90,186 | 90,186 | 0 |
  | Passed | 54,651 | 54,657 | **+6** |
  | Ignored | 23,273 | 23,273 | 0 |
  | Failed | 12,262 | 12,256 | **-6** |
  | Panics | 0 | 86 | **+86** |
  | Conformance | 60.60% | 60.60% | **+0.01%** |
  
  <details><summary><b>Fixed tests (12):</b></summary>
  
  ```
  test/intl402/DateTimeFormat/constructor-options-throwing-getters-fractionalSecondDigits.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-throwing-getters-fractionalSecondDigits.js (previously Failed)
  test/intl402/DateTimeFormat/timezone-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/timezone-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-calendar-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-calendar-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-calendar-numberingSystem-order.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-calendar-numberingSystem-order.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-throwing-getters-dayPeriod.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-throwing-getters-dayPeriod.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-throwing-getters-timedate-style.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-throwing-getters-timedate-style.js (previously Failed)
  ```
  </details>
  
  <details><summary><b>New panics (86):</b></summary>
  
  ```
  test/intl402/constructors-string-and-single-element-array.js [strict mode] (previously Passed)
  test/intl402/constructors-string-and-single-element-array.js (previously Passed)
  test/intl402/DateTimeFormat/constructor-options-order-timedate-style.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-order-timedate-style.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-style-conflict.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-style-conflict.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeStyle-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeStyle-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-numberingSystem-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-numberingSystem-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dateStyle-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dateStyle-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-order.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-order.js (previously Failed)
  test/intl402/DateTimeFormat/default-options-object-prototype.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/default-options-object-prototype.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-fractionalSecondDigits-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-fractionalSecondDigits-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-throwing-getters.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-throwing-getters.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-order-fractionalSecondDigits.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-order-fractionalSecondDigits.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeZoneName-valid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeZoneName-valid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeStyle-valid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeStyle-valid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeZoneName-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-timeZoneName-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dateStyle-valid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dateStyle-valid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-fractionalSecondDigits-valid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-fractionalSecondDigits-valid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dayPeriod-valid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dayPeriod-valid.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-order-dayPeriod.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-order-dayPeriod.js (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dayPeriod-invalid.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/constructor-options-dayPeriod-invalid.js (previously Failed)
  test/intl402/DateTimeFormat/legacy-regexp-statics-not-modified.js [strict mode] (previously Passed)
  test/intl402/DateTimeFormat/legacy-regexp-statics-not-modified.js (previously Passed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/pattern-on-calendar.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/pattern-on-calendar.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/fractionalSecondDigits.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/fractionalSecondDigits.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/en-US.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/en-US.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/date-same-returns-single-date.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRangeToParts/date-same-returns-single-date.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRange/fractionalSecondDigits.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRange/fractionalSecondDigits.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRange/en-US.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRange/en-US.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRange/date-same-returns-single-date.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatRange/date-same-returns-single-date.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle-default.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle-default.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/pattern-on-calendar.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/pattern-on-calendar.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/dayPeriod-narrow-en.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/dayPeriod-narrow-en.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/related-year.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/related-year.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/dayPeriod-long-en.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/dayPeriod-long-en.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/fractionalSecondDigits.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/fractionalSecondDigits.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/dayPeriod-short-en.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/formatToParts/dayPeriod-short-en.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/related-year-zh.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/related-year-zh.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/dayPeriod-narrow-en.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/dayPeriod-narrow-en.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/dayPeriod-long-en.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/dayPeriod-long-en.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/fractionalSecondDigits.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/fractionalSecondDigits.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/timedatestyle-en.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/timedatestyle-en.js (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/dayPeriod-short-en.js [strict mode] (previously Failed)
  test/intl402/DateTimeFormat/prototype/format/dayPeriod-short-en.js (previously Failed)
  test/intl402/DateTimeFormat/supportedLocalesOf/taint-Object-prototype.js [strict mode] (previously Passed)
  test/intl402/DateTimeFormat/supportedLocalesOf/taint-Object-prototype.js (previously Passed)
  test/intl402/Date/prototype/toLocaleString/default-options-object-prototype.js [strict mode] (previously Failed)
  test/intl402/Date/prototype/toLocaleString/default-options-object-prototype.js (previously Failed)
  ```

NorbertGarfield:
@raskad, thanks for the update. I fixed these panicking tests locally, but I don't know how to gather that VM implementation report. My current method is to run failing script directly via `cargo run -- /path/to/failing_test.js`.

Also, `Locale::from_bytes` call panicked when parsing "i-klingon" locale, so I had to add a fallback to my previous implementation to separate extensions from the locale id. Please let me know if you have another idea.

raskad:
> @raskad, thanks for the update. I fixed these panicking tests locally, but I don't know how to gather that VM implementation report. My current method is to run failing script directly via `cargo run -- /path/to/failing_test.js`.

You are right, we should describe how to gather this report locally.
I always do it like this:
```bash
git checkout main
cargo run --release --bin boa_tester -- run -vv -o results/main
git checkout <your-branch>
cargo run --release --bin boa_tester -- run -vv -o results/<your-branch>
cargo run --release --bin boa_tester -- compare results/mainlatest.json results/<your-branch>/latest.json
```

You may want to redirect the output of the `boa_tester run` commands to some file, because printing ~200k lines to stdout can be really slow.

I will review the PR in a bit ;)

jedel1043:
@NorbertGarfield Do note, however, that if you're on Windows you'll probably have some additional panicking tests caused by #1570. You can just ignore them in the meantime 😅

boa_engine/src/builtins/intl/date_time_format.rs
```diff
@@ -234,3 +259,1208 @@ pub(crate) fn to_date_time_options(
...
+
+    // Check that each slice is has length between 3 and 8 and all characters are alphanumeric.
+    for option in options_vec {
+        if option.len() < 3 || option.len() > 8 {
+            return true;
+        }
+
+        if option
+            .chars()
+            .any(|character| !character.is_ascii_alphanumeric())
```
Razican, line 303: Wouldn't it be better to do this using the [`all()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all) function?

```suggestion
        if !option
            .chars()
            .all(char::is_ascii_alphanumeric)
```

boa_engine/src/builtins/intl/date_time_format.rs
```diff
@@ -234,3 +259,1208 @@ pub(crate) fn to_date_time_options(
...
+    string_to_hc_map.insert(JsString::from("h11"), preferences::HourCycle::H11);
+    string_to_hc_map.insert(JsString::from("h12"), preferences::HourCycle::H12);
+    string_to_hc_map.insert(JsString::from("h23"), preferences::HourCycle::H23);
+    string_to_hc_map.insert(JsString::from("h24"), preferences::HourCycle::H24);
+
+    let hour_cycle = string_to_hc_map
+        .get(hour_cycle_str)
+        .expect("Invalid hour cycle");
+    *hour_cycle
+}
```
raskad, line 436: Using a `match` on `hour_cycle_str.as_str()` here is way more efficient than the allocations of the map and strings.

boa_engine/src/builtins/intl/date_time_format.rs
```diff
@@ -234,3 +259,1208 @@ pub(crate) fn to_date_time_options(
...
+    let mut string_to_style_map = FxHashMap::<JsString, Option<length::Date>>::default();
+    string_to_style_map.insert(JsString::from("full"), Some(length::Date::Full));
+    string_to_style_map.insert(JsString::from("long"), Some(length::Date::Long));
+    string_to_style_map.insert(JsString::from("medium"), Some(length::Date::Medium));
+    string_to_style_map.insert(JsString::from("short"), Some(length::Date::Short));
+
+    match string_to_style_map.get(&date_style_str) {
+        Some(date_style) => *date_style,
+        None => None,
+    }
```
raskad, line 458: Same here. `match` should be better.

boa_engine/src/builtins/intl/date_time_format.rs
```diff
@@ -234,3 +259,1208 @@ pub(crate) fn to_date_time_options(
...
+    let mut string_to_style_map = FxHashMap::<JsString, Option<length::Time>>::default();
+    string_to_style_map.insert(JsString::from("full"), Some(length::Time::Full));
+    string_to_style_map.insert(JsString::from("long"), Some(length::Time::Long));
+    string_to_style_map.insert(JsString::from("medium"), Some(length::Time::Medium));
+    string_to_style_map.insert(JsString::from("short"), Some(length::Time::Short));
+
+    match string_to_style_map.get(&time_style_str) {
+        Some(time_style) => *time_style,
+        None => None,
+    }
```
raskad, line 481: Same

boa_engine/src/builtins/intl/date_time_format.rs
```diff
@@ -234,3 +259,1208 @@ pub(crate) fn to_date_time_options(
...
+        numeric_to_value(format.hour)
+    } else if property == "minute" {
+        numeric_to_value(format.minute)
+    } else if property == "second" {
+        numeric_to_value(format.second)
+    } else if property == "timeZoneName" {
+        time_zone_to_value(format.time_zone_name)
+    } else {
+        JsValue::undefined()
+    }
```
raskad, line 571: Using a `match` here would be better

boa_engine/src/builtins/intl/date_time_format.rs
```diff
@@ -234,3 +259,1208 @@ pub(crate) fn to_date_time_options(
...
+
+/// `DateTimeComponents` type contains `property` map and list of `values`
+#[derive(Debug)]
+struct DateTimeComponents {
+    property: JsString,
+    values: Vec<JsString>,
+}
+
+/// Builds a list of `DateTimeComponents` which is commonly referred to as "Table 6"
+fn build_date_time_components() -> Vec<DateTimeComponents> {
```
raskad, line 590: Same a before here. Using `Vec::from([])` instead of `vec!`

raskad (changes requested):

boa_engine/src/builtins/intl/date_time_format.rs
NorbertGarfield: Done, however I used lambda since `all(char::is_ascii_alphanumeric)` failed with
```
|   expected signature of `fn(char) -> _`
|   found signature of `for<'r> fn(&'r char) -> _`
```

Is there a better approach for that?

Razican: I think it's OK this way, the issue is that the `chars()` function returns an iterator with `char` as the item, but the `is_ascii_alphanumeric` expects a reference (`&self`). I think the approach you followed is the best one :)

boa_engine/src/builtins/intl/date_time_format.rs
```diff
@@ -234,3 +259,1192 @@ pub(crate) fn to_date_time_options(
...
+        },
+    ])
+}
+
+/// Builds a list of `components::Bag` for all possible combinations of dateStyle and timeStyle
+/// ("full", "medium", "short", "long", undefined) for specified `locale` and `calendar`
+pub(crate) fn build_formats(locale: &JsString, calendar: &JsString) -> Vec<components::Bag> {
+    let locale_str = locale.to_string();
+    let locale = Locale::from_bytes(locale_str.as_bytes()).expect("Locale parsing failed");
+    let provider = InvariantDataProvider;
```
jedel1043, line 651: Just a heads up. The doc for `InvariantDataProvider` specifies this provider is:
`Sometimes useful for testing. Not intended to be used in production environments.`

Currently we cannot pull the icu data until they release the 0.6.0 version of `icu_datagen` (which should be pretty soon, looking at https://github.com/unicode-org/icu4x/issues/1861), so I'd recommend to wait a bit to integrate this function until we have the required data to create our provider :)

NorbertGarfield: As far, as I can see from this PR https://github.com/boa-dev/boa/pull/2078 lookup_match tests must be updated. Should I file a new PR to do that?

jedel1043: That should be done on that PR itself, just to have everything organized. We are still waiting on #2079 to fix the newest clippy lints. After that, I'll fix and rebase #2078 ASAP

NorbertGarfield: Just to clarify, I need to rebase to https://github.com/boa-dev/boa/pull/2083 once it is merged and then use `out_file` to get `DateTimeFormat`. Did I get it right?

jedel1043: Almost. You just directly use the `ICU4X_DATA_PROVIDER` static.

NorbertGarfield: I don't think, I understand how to use `BoaProvider` properly. Replacing

```
let provider = InvariantDataProvider;
```
with
```
let provider = context.icu().provider();
```

results in 

```
the trait `icu_provider::ResourceProvider<WeekDataV1Marker>` is not implemented for `&dyn BoaProvider`
```

jedel1043: Ah, I added to `BoaProvider` the minimal bounds to work. If there are some additional bounds required, you can add them to the definition of `BoaProvider`

NorbertGarfield: That bound is present in [this](https://github.com/boa-dev/boa/blob/518bad810974aa4cf6fcd3d864756f298d1b204d/boa_engine/src/context/icu.rs#L32) list. I assume, it has something to do with a reference to `dyn BoaProvider`. I've found several topics similar to [that](https://users.rust-lang.org/t/why-does-dyn-trait-not-implement-trait/30052) but none of them work for me.

jedel1043: Ah, I see the problem now. You're passing `&provider` to the functions, but `provider` is a `&dyn BoaProvider`, meaning the function is testing trait bounds for `&&dynBoaProvider`. Remove the reference for `&provider` and it should work again

NorbertGarfield: That's the first thing I've tried:

```
721 |     provider,
    |     ^^^^^^^^ doesn't have a size known at compile-time

```

jedel1043: That's probably a bug from the `icu` crate. I checked the function signature and the generic parameters implicitly require a `Sized` bound in them.

For now, can you manually implement the required bounds for `&dyn BoaProvider`?
I'll open an issue in the icu4x repo to address this, and we'll remove the implementations when the new version releases :)

NorbertGarfield: Sure. I'll give it a try.

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -151,6 +151,41 @@ fn best_available_locale(available_locales: &[JsString], locale: &JsString) -> O
...
+/// If no extensions found, return the length of requested locale
+fn get_leftmost_unicode_extension_pos(requested_locale: &str) -> usize {
+    let ext_sep = "-u-";
+    let src_locale = requested_locale.to_lowercase();
+    let pos = src_locale.find(ext_sep);
+    match pos {
+        Some(idx) => idx,
+        None => src_locale.len(),
+    }
+}
```
zbraniecki, line 165: `suggestion`: You should use `icu::locid::Locale` and get `LanguageIdentifier` out of it.

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -171,9 +206,11 @@ fn lookup_matcher(
     for locale_str in requested_locales {
         // a. Let noExtensionsLocale be the String value that is locale with any Unicode locale
         //    extension sequences removed.
-        let parsed_locale =
-            Locale::from_bytes(locale_str.as_bytes()).expect("Locale parsing failed");
-        let no_extensions_locale = JsString::new(parsed_locale.id.to_string());
+        let maybe_locale = Locale::from_bytes(locale_str.as_bytes());
+        let no_extensions_locale = match &maybe_locale {
+            Ok(parsed_locale) => JsString::new(parsed_locale.id.to_string()),
+            Err(_) => trim_unicode_extensions(locale_str),
```
zbraniecki, line 212: `suggestion`: if it is an invalid locale, then you shouldn't take it at all.

boa_engine/src/builtins/intl/tests.rs
```diff
@@ -154,6 +163,13 @@ fn uni_ext_comp() {
     assert_eq!(components.keywords.len(), 1);
     assert_eq!(components.keywords[0].key, "ca");
     assert_eq!(components.keywords[0].value, "islamic-civil");
+
+    let ext = JsString::new("u-ca-islamic-civil");
+    let components = unicode_extension_components(&ext);
```
zbraniecki, line 168: We give it to you - you can just use `locale.extensions.unicode` for all of that. No need to reconstruct it with your own types. https://docs.rs/icu/latest/icu/locid/extensions/unicode/index.html

zbraniecki (commented):
Hi.

First of all, this is so exciting for us. We're building ICU4X particularly to target ECMA-402 needs well and so far we've mostly been testing it in context of SpiderMonkey use in Firefox.
Your work on boa+ecma-402 is a perfect fit and I'd love us to collaborate closer to supply your needs.

For this particular PR, I left some drive-by comments, and my overarching feedback, as a maintainer of `locid` component is - you should never have to work with strings. Any place where you substr or look for `-u` or anything like that feels like either `icu_locid` should provide you that, or you're doing something wrong. Generally speaking I strongly believe the whole implementation should always operate on a `Locale` type and only stringify/parse on output/input.
If something is missing or the API we have is suboptimal, let's fix it. It's a perfect timing as we're warming up to 1.0 stability but still have wiggle room left.

jedel1043:
@zbraniecki Thank you for your review! As you can see, our `Intl` implementation using `icu4x` is still very lacking, but I hope we can be of help by being early adopters of the library!

On another note, we had a small problem with `Locale` when we implemented the [`BestAvailableLocale`](https://tc39.es/ecma402/#sec-bestavailablelocale) operation. The algorithm described there requires as steps:

```
b. Let pos be the character index of the last occurrence of "-" (U+002D) within candidate. If that character does not occur, return undefined.
c. If pos ≥ 2 and the character "-" occurs at index pos - 2 of candidate, decrease pos by 2.
d. Let candidate be the substring of candidate from position 0, inclusive, to position pos, exclusive.
```

If we translate this to a high level operation, the spec requires removing the last word from the locale. Unfortunately, `Locale` doesn't expose its internal data structures, so we have to convert to a `String` and parse again if we want to reduce the `Locale` by one word. So, would it be too much to ask for a `pop_last_word` method in `Locale`?

If that's not possible, then a way to deconstruct the internals of each extension would be good, because that would allow us to do something like: `mem::take` -> deconstruct -> modify -> recreate modified using `unchecked_*`.

zbraniecki:
Uuh, that's really unfortunate. This seems to *want* you to operate on a stringified version of Locale, which i think we should avoid. There are other issues with it (like, "sr-Cyrl" should not be shortened to "sr"), but also, it makes the unreasonable attempt to cut out subtags from extensions.

I think this operation should instead operate on `[variants, region, script, language]` array removing each one in order after removing all extensions.

I don't think we should be adding APIs to handle treating Locale as a string and make semantic operations based on positions of subtags. I think this is a change we should apply to ECMA-402 spec.  CC @sffc for validation and we can file an issue there if he doesn't point out that I am missing something obvious.

sffc:
The spec text only requires that behavior be consistent with what *would* happen if you did the string operations on the locale string. You can implement it differently, so long as the result is the same. In this case, BCP47 guarantees that variant always comes after region which always comes after script which always comes after language, and you can build that assumption into your code.

boa_engine/src/builtins/intl/date_time_format.rs
NorbertGarfield: @jedel1043 I have tried different approaches, so that's my best shot so far. I have no idea where to get [source data](https://github.com/boa-dev/boa/pull/2072/files#diff-872037c107c01bf644ede412e4802b3eefeb5a70ce595c441f75651d45111b2aR44) though.

Is there a way to cast `&dyn BoaProvider` to a concrete type? I believe that's `icu_provider_blob::StaticDataProvider` in my case.

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: Removed `get_leftmost_unicode_extension_pos` completely.

boa_engine/src/builtins/intl/mod.rs
NorbertGarfield: Done. https://github.com/boa-dev/boa/pull/2072/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R167

boa_engine/src/builtins/intl/tests.rs
NorbertGarfield: Done. https://github.com/boa-dev/boa/pull/2072/files#diff-1dd4af22d3a6142a8060d056e9d0489d84d8a4d9c68a8f6e7b359b8d3ab35499R425

boa_engine/src/builtins/intl/date_time_format.rs
jedel1043: No need, I fixed the impls on b2764e5 :)

Also, https://github.com/unicode-org/icu4x/pull/1969 has been merged, so we'll definitely remove those when 0.7.0 releases

NorbertGarfield: Great! Thanks a lot. I assume tests are failing on
```
let maybe_dtf = datetime::DateTimeFormat::<Gregorian>::try_new(locale, &provider, &options);
```

With

```
ICU4X data error: Missing data for variant or locale (key: datetime/lengths@1)
```

I'll try with another lengths shortly.

sffc: Pro tip: enable the `log_error_context` feature on the `icu_provider` crate to see more detailed error messages in the log.

zbraniecki:
> The problem are extensions, which could become invalid if that algorithm is followed as stated e.g. with hi-t-en-h0-hybrid it would try to match with hi-t-en-h0, which has an invalid t extension since the h0 key ought to have a value.

I claim with the authority given to me by the ultimate powers (Shane in this case) that you should, for all purposes, treat this as a LanguageIdentifier. It just that ICU4X is the first time we provide API surface to the spec that identifies LI as a separate thing from Locale.

I hereby recommend:

1) In this code, take LI out of Locale and in the algo just take out subtag after subtag and compare.
2) File an issue in ECMA-402 to request additional precision in the algo to cut out extensions

I'd particularly appreciate if you did (2) as I know it's easy to stop on (1) and move on :) Thank you!

jedel1043:
@zbraniecki I opened an issue in the spec and in the meantime we'll follow your suggestions and treat the input as a Language Identifier. Thank you for all your assistance!

jedel1043:
There's a new Intl API in place for implementing Intl services more easily. This is waiting on the author to decide if they want to rebase on that or if we want to do it from scratch.

NorbertGarfield:
@jedel1043, sorry for the late response. I'll try to implement it from scratch. I'm closing this PR.

PR #2076, opened 2022-05-18
[Merged by Bors] - Implement `ProxyBuilder`

### LINKED ISSUE #2075: Expose `Proxy` builtin as a public API
As per discussion in Discord:

```
[8:23 PM] Jedel: I see. We currently haven't exposed the Proxy builtin as a public API. However, we're implementing safe wrappers for every object type to make them easier to use in use cases like yours
[8:25 PM] Jedel: If you want, you can open an issue in our repo and we'll work on implementing a safe wrapper for Proxy
[8:30 PM] mikroskeem: hm, I'm not sure if that is the best approach to simply create a rust wrapper for Proxy object
[8:30 PM] mikroskeem: but then again I'm not familiar with Boa internals at all
[8:30 PM] Jedel: The difference would be that a wrapper for Proxy would make it possible to pass native functions as proxy handlers
[8:31 PM] Jedel: whereas a Proxy object inside JS uses JS functions
```

My use-case would be defining an proxy in native code, e.g to emulate `process.env[key]` for code I cannot control.

Simply put, instead of evaluating this before running vendor code:

```javascript
globalThis.process = {
  env: new Proxy({}, {
    get: (_self, key) => myBoaNativeFuncToGetEnv(key),
    //set: (_self, key, value) => { throw new Error("Not supported"); },
  }),
};
```

could be replaced with 100% Rust counterpart.

jedel1043:
This Pull Request fixes/closes #2075.

It changes the following:

- Implements a `ProxyBuilder` struct to be able to create `Proxy` objects from native function traps.
- Documents `ProxyBuilder` and adjusts some documentation.
- Fixes a clippy lint.


raskad (commented):
The changes look good to me. I like the builder approach.

Should we align this new public api with the existing ones in `boa_engine/src/object/` like `JsArray`, `JsFunction`, etc.? If not, we basically have an additional kind of public api right?

jedel1043:
> The changes look good to me. I like the builder approach.
> 
> Should we align this new public api with the existing ones in `boa_engine/src/object/` like `JsArray`, `JsFunction`, etc.? If not, we basically have an additional kind of public api right?

Yeah, I can do that. It won't have any methods since `Proxy` objects don't have methods, but it would be good to have symmetry with the other APIs

jedel1043:
Done! Will wait for #2079 to be merged, since that PR fixed the lints.

raskad (approved):

HalidOdat (approved):

PR #2082, opened 2022-05-20
[Merged by Bors] - Fixing build for changes in clippy for Rust 1.61

Razican:
This fixes the CI after the upgrade to Rust 1.61. I had to "allow" the `use_self` lint, due to https://github.com/rust-lang/rust-clippy/issues/8845 and https://github.com/rust-lang/rust-clippy/issues/6902.

It removed some false negatives, though, so I fixed some of the usage.


raskad:
Probably the same as #2079?

jedel1043:
> Probably the same as #2079?

Yep, but I would prefer having both separated. I just fixed it since it was throwing errors, I can rebase over this PR

jedel1043 (approved):

raskad (approved):

Issue #2068, opened 2022-05-10
Optional features


Razican:
Currently, the engine is provided as-is, with almost no room for customizing its features. Nevertheless, we have some situations (such as the optional #2067 feature, or the `eval()` function, or even not fully published features where it would be nice if the user of the engine could decide if they are available or not.

This can be provided by creating a feature for each of the features in the spec. Then, by default, we would include everything that is part of the spec and it's mandatory (still allowing our users to not use default features and pick each of them individually). Parts of the spec that are still only proposals (even if they are in test262) should not be included by default, but should be included when testing.

Furthermore, some users might not want all built-ins in the engine, to provide a smaller surface for their users, so we should provide an API where the global object is empty, and users can decide the list of globals that get added.

This would require extensive testing, so we would need to probably do a `cargo check` with each feature individually and `--no-default-features`.

PR #2083, opened 2022-05-22
[Merged by Bors] - Integrate ICU4X into `Intl` module

jedel1043:
This Pull Request integrates an `ICU4X` data provider API in our codebase, to make use of the internationalization APIs that this crate provides.

It changes the following:

- Creates an API for pluggable icu data providers at `Context` creation, adding an `Icu` struct to store the provider (and some other internationalization tools) at runtime.
- Slightly changes locale related functions to preserve the `Locale` type and ensure correctness. (Will make some other changes related to this).
- Integrates the `sys_locale` crate to fetch the current default locale of an user instead of always returning `en-US`.

### DIFF
--- a/boa_engine/Cargo.toml
+++ b/boa_engine/Cargo.toml
@@ -14,6 +14,15 @@ readme = "../README.md"
 [features]
 profiler = ["boa_profiler/profiler"]
 deser = ["boa_interner/serde"]
+intl = [
+    "dep:icu_locale_canonicalizer",
+    "dep:icu_locid",
+    "dep:icu_datetime",
+    "dep:icu_plurals",
+    "dep:icu_provider",
+    "dep:icu_testdata",
+    "dep:sys-locale"
+]
 
 # Enable Boa's WHATWG console object implementation.
 console = []
@@ -41,7 +50,13 @@ unicode-normalization = "0.1.19"
 dyn-clone = "1.0.5"
 once_cell = "1.12.0"
 tap = "1.0.1"
-icu = "0.6.0"
+icu_locale_canonicalizer = { version = "0.6.0", features = ["serde"], optional = true }
+icu_locid = { version = "0.6.0", features = ["serde"],  optional = true }
+icu_datetime = { version = "0.6.0", features = ["serde"], optional = true }
+icu_plurals = { version = "0.6.0", features = ["serde"], optional = true }
+icu_provider = { version = "0.6.0", optional = true }
+icu_testdata = {version = "0.6.0", optional = true}
+sys-locale = { version = "0.2.0", optional = true }
 
 [dev-dependencies]
 criterion = "0.3.5"
--- a/boa_engine/Cargo.toml
+++ b/boa_engine/Cargo.toml
@@ -2,7 +2,7 @@
 name = "boa_engine"
 version = "0.14.0"
 edition = "2021"
-rust-version = "1.58"
+rust-version = "1.60"
 authors = ["boa-dev"]
 description = "Boa is a Javascript lexer, parser and Just-in-Time compiler written in Rust. Currently, it has support for some of the language."
 repository = "https://github.com/boa-dev/boa"
--- a/boa_engine/src/builtins/intl/mod.rs
+++ b/boa_engine/src/builtins/intl/mod.rs
@@ -21,11 +21,11 @@ pub mod date_time_format;
 mod tests;
 
 use boa_profiler::Profiler;
+use icu_locale_canonicalizer::LocaleCanonicalizer;
+use icu_locid::{locale, Locale};
 use indexmap::IndexSet;
 use rustc_hash::FxHashMap;
-use tap::{Conv, Pipe};
-
-use icu::locid::Locale;
+use tap::{Conv, Pipe, TapOptional};
 
 /// JavaScript `Intl` object.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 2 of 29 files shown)


jedel1043:
Slightly related to #2068, since CLDR data increases our binary size significantly, and some users won't even use `Intl`

jedel1043:
Opened https://github.com/unicode-org/icu4x/issues/1925 to fix the two failing tests.

raskad:
Do we really have to include the data in the executable? In addition to increasing binary sizes, this also is problematic for deterministic builds.

We call `StaticDataProvider::new_from_static_blob` at runtime right? For me it seems like not having to read and parse the data at runtime would be the only benefit to including it directly.

Can we not have an api for including the data? It would then most likely be a field on the `Context` I guess? We could then use this api ourselves in the relevant tests. That would also remove the need for a `build.rs` as we would only need the data for tests.

jedel1043:
> Do we really have to include the data in the executable? In addition to increasing binary sizes, this also is problematic for deterministic builds.

Yeah, we do. The alternative would be to download the CLDR data at runtime, but that implies giving boa networking access at runtime.

jedel1043:
> Can we not have an api for including the data? It would then most likely be a field on the `Context` I guess? We could then use this api ourselves in the relevant tests. That would also remove the need for a `build.rs` as we would only need the data for tests

I think the biggest disadvantage of this would be that a user that wants to use the `Intl` object would need to learn the whole `icu_data_provider` api in order to even initialize the data, and if we do offer a fallback default data, then that's no different that having this PR but with a feature gate. (and I'm planning to do that asap)

jedel1043:
Just to complement my previous comments. If we want to preserve deterministic builds, and want the ability to include the data only on tests, a feature gate that also skips running the `build.rs` script is the simplest solution.

raskad:
In addition to the feature flag, what do you think about hosting the `icu4x_data.postcard` file ourselves @jedel1043? Generating it in `build.rs` seems to increase the build times in our ci by a pretty large amount (in the test job on linux from ~8 minutes to ~14 minutes)

jedel1043:
By hosting you mean pushing the generated file to this repo? That would increase massively the size of this repo. The size of this repo is approximately 10 MB, and by additionally hosting the postcard file, we would increase our repo size to 43MB. I don't think that's ideal. We're already caching build artifacts, which includes the postcard file. Another solution would be to add a simple check inside `build.rs` for the generated file.

Also, I'm planning on replacing this current solution with an auto-generated module using the new `crabbake` feature, but the next release will be approximately in 3 months, so we won't need a postcard file at that point.

raskad:
My idea would have been to generate the postcard files for every of our releases and push them to a separate branch or even have them in a separate git repository. That way we would sidestep the comparability and the size issue.

But apart from that we still would need a build.rs to download the files anyways. I think we should cut down on the new dependencies for that. Is there a need for `const_format` and `eyre`? Can't we just `expect` on every error in the build.rs?
Also I don't see the benefit of using `cached_path`. That crate pulls in too many dependencies that we are not using. The minimal set of dependencies would probably be this:
```toml
reqwest = { version = "0.11.10", default-features = false, features = ["rustls-tls"]}
tokio = { version = "1", features = ["full"] }
zip = { version = "0.6.2", default-features = false, features = ["deflate"] }
```

Imo the few lines of code that we save are not worth the ~50 additional dependencies.

jedel1043:
Yeah, I was also thinking using `eyre`, `const_format` and `cached-path` is a bit overkill for a script ran only once. We can also skip `tokio` by using `ureq` instead, reducing even more the number of dependencies

jedel1043:
OK, following @raskad suggestions and seeing v8 and spidermonkey made the same tradeoff about data providers, I completely removed the `build.rs` generated module and instead opted for a `Context` API to define data providers at runtime.

This is the most flexible option we currently can offer to our users, and generating the data is not that difficult with the `icu_datagen` crate.

I also took advantage of the `icu_testdata` crate to be able to offer a default provider in the meantime (and also to define a provider for us in tests), since I'd like to slowly decide on a better fitting default for users that only want to use `Intl` without doing any configuration :)

raskad (approved):
This approach looks really good to me. I would be happy to merge this. Two follow ups I see are:
* creation of an example on how to include the icu data (can probably wait until we have a good amount of `Intl` tests passing so that the feature can be used)
* Usage of icu data in our tests

boa_engine/Cargo.toml
```diff
@@ -14,6 +14,15 @@ readme = "../README.md"
 [features]
 profiler = ["boa_profiler/profiler"]
 deser = ["boa_interner/serde"]
+intl = [
+    "dep:icu_locale_canonicalizer",
```
Razican, line 18: I think this syntax was introduced fairly recently, we will probably need to increase the MSRV.

boa_engine/src/builtins/intl/mod.rs
```diff
@@ -442,9 +385,77 @@ fn insert_unicode_extension_and_canonicalize(locale: &str, extension: &str) -> J
...
+        .expect("Assert: ! IsStructurallyValidLanguageTag(locale) is true.");
+
     // 8. Return ! CanonicalizeUnicodeLocaleId(locale).
-    Intl::canonicalize_locale(&new_locale)
+    canonicalize_unicode_locale_id(&mut new_locale, canonicalizer);
+    new_locale.to_string().into()
+}
+
+fn canonicalize_locale_list(args: &[JsValue], context: &mut Context) -> JsResult<Vec<Locale>> {
+    // https://tc39.es/ecma402/#sec-canonicalizelocalelist
```
Razican, line 399: I guess we could add the link to the function documentation, right? And add an explanation of the usage :)

Razican (approved):
Looks good to me :) check my comments to see if they make sense, and we can then merge!

jedel1043:
@Razican Sorry! I addressed your comments and requested a review but I didn't see that you'd already approved the PR 😅

Razican (approved):
All good! let's merge :)

PR #2090, opened 2022-05-26
Implement `Function` constructor

raskad:

- Implement `Function` constructor
- Ignore non-standard `caller` feature in 262 tests
- Fix `Function.apply` length
- Use `TypeError` in `Function.caller` and `Function.arguments` accessors

### DIFF
--- a/boa_engine/src/builtins/function/mod.rs
+++ b/boa_engine/src/builtins/function/mod.rs
@@ -22,10 +22,12 @@ use crate::{
     object::{ConstructorBuilder, FunctionBuilder, Ref, RefMut},
     property::{Attribute, PropertyDescriptor, PropertyKey},
     symbol::WellKnownSymbols,
+    syntax::{ast::node::FormalParameterList, Parser},
     value::IntegerOrInfinity,
     Context, JsResult, JsString, JsValue,
 };
 use boa_gc::{self, Finalize, Gc, Trace};
+use boa_interner::Sym;
 use boa_profiler::Profiler;
 use dyn_clone::DynClone;
 use std::{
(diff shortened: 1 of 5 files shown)


boa_engine/src/builtins/function/mod.rs
```diff
@@ -275,23 +277,152 @@ pub struct BuiltInFunctionObject;
...
+                        .throw_syntax_error(format!("failed to parse function body: {e}"))
+                }
+            };
+
+            // Early Error: If BindingIdentifier is present and the source text matched by BindingIdentifier is strict mode code,
+            // it is a Syntax Error if the StringValue of BindingIdentifier is "eval" or "arguments".
+            if body.strict() {
+                for parameter in parameters.parameters.iter() {
+                    for name in parameter.names() {
+                        if [Sym::EVAL, Sym::ARGUMENTS].contains(&name) {
```
Razican, line 356: Since we are using symbols here, and integer comparisons are very cheap, I think it will be faster to just have two equals in an "or" comparison.

boa_engine/src/bytecompiler.rs
```diff
@@ -2906,3 +2921,10 @@ impl<'b> ByteCompiler<'b> {
         Ok(())
     }
 }
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub(crate) enum FunctionKind {
```
Razican, line 2926: Could we add some documentation to this enumeration?

Razican (commented):
Looks very good! I added some suggestions :)

Razican (approved):

RageKnify (approved):

PR #1923, opened 2022-03-13
Add initial support for promises

aaronmunsters:
It adds the following:

- A job queue (in `Context`)
- The constructor [`Promise`](https://tc39.es/ecma262/#sec-promise-executor)
- [`Promise.prototype.then`](https://tc39.es/ecma262/#sec-promise.prototype.then)
- The additional needed infrastructure
  - [`PerformPromiseThen ( promise, onFulfilled, onRejected [ , resultCapability ] )`](https://tc39.es/ecma262/#sec-performpromisethen)
  - [`TriggerPromiseReactions ( reactions, argument )`](https://tc39.es/ecma262/#sec-triggerpromisereactions)
  - [`RejectPromise ( promise, reason )`](https://tc39.es/ecma262/#sec-rejectpromise)
  - [`FulfillPromise ( promise, value )`](https://tc39.es/ecma262/#sec-fulfillpromise)
  - [`CreateResolvingFunctions ( promise )`](https://tc39.es/ecma262/#sec-createresolvingfunctions)
  - [`NewPromiseCapability ( C )`](https://tc39.es/ecma262/#sec-newpromisecapability)
  - [`NewPromiseReactionJob ( reaction, argument )`](https://tc39.es/ecma262/#sec-newpromisereactionjob)
  - [`NewPromiseResolveThenableJob ( promiseToResolve, thenable, then )`](https://tc39.es/ecma262/#sec-newpromiseresolvethenablejob)
- A test case showcasing the run-to-completion semantics.

An example program that shows the control flow with this addition is:
```javascript
new Promise((res, rej) => {
  console.log("A");
  res(undefined);
}).then((_) => console.log("B"));
console.log("C");
```
Which would output:
```
A
C
B
```

Feedback is most welcome! :wink:

PR #2107, opened 2022-06-08
Execution stack & promises

Razican:
This PR overrides #1923. It also removes the `queues` dependency added there, and rebases it to the latest `main` branch state.

It adds the following:

- A job queue (in `Context`)
- The constructor [`Promise`](https://tc39.es/ecma262/#sec-promise-executor)
- [`Promise.prototype.then`](https://tc39.es/ecma262/#sec-promise.prototype.then)
- The additional needed infrastructure
  - [`PerformPromiseThen ( promise, onFulfilled, onRejected [ , resultCapability ] )`](https://tc39.es/ecma262/#sec-performpromisethen)
  - [`TriggerPromiseReactions ( reactions, argument )`](https://tc39.es/ecma262/#sec-triggerpromisereactions)
  - [`RejectPromise ( promise, reason )`](https://tc39.es/ecma262/#sec-rejectpromise)
  - [`FulfillPromise ( promise, value )`](https://tc39.es/ecma262/#sec-fulfillpromise)
  - [`CreateResolvingFunctions ( promise )`](https://tc39.es/ecma262/#sec-createresolvingfunctions)
  - [`NewPromiseCapability ( C )`](https://tc39.es/ecma262/#sec-newpromisecapability)
  - [`NewPromiseReactionJob ( reaction, argument )`](https://tc39.es/ecma262/#sec-newpromisereactionjob)
  - [`NewPromiseResolveThenableJob ( promiseToResolve, thenable, then )`](https://tc39.es/ecma262/#sec-newpromiseresolvethenablejob)
- A test case showcasing the run-to-completion semantics.

An example program that shows the control flow with this addition is:
```javascript
new Promise((res, rej) => {
  console.log("A");
  res(undefined);
}).then((_) => console.log("B"));
console.log("C");
```
Which would output:
```
A
C
B
```

### DIFF
--- a/boa_engine/src/builtins/promise/mod.rs
+++ b/boa_engine/src/builtins/promise/mod.rs
@@ -404,7 +404,7 @@ impl Promise {
                 );
 
                 // 15. Perform HostEnqueuePromiseJob(job.[[Job]], job.[[Realm]]).
-                context.host_enqueue_promise_job(Box::new(job));
+                context.host_enqueue_promise_job(job);
 
                 // 16. Return undefined.
                 Ok(JsValue::Undefined)
@@ -575,7 +575,7 @@ impl Promise {
                 PromiseJob::new_promise_reaction_job(reaction.clone(), argument.clone(), context);
 
             // b. Perform HostEnqueuePromiseJob(job.[[Job]], job.[[Realm]]).
-            context.host_enqueue_promise_job(Box::new(job));
+            context.host_enqueue_promise_job(job);
         }
 
         // 2. Return unused.
@@ -688,7 +688,7 @@ impl Promise {
                     PromiseJob::new_promise_reaction_job(fulfill_reaction, value, context);
 
                 //   c. Perform HostEnqueuePromiseJob(fulfillJob.[[Job]], fulfillJob.[[Realm]]).
-                context.host_enqueue_promise_job(Box::new(fulfill_job));
+                context.host_enqueue_promise_job(fulfill_job);
             }
 
             // 11. Else,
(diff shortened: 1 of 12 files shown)


boa_engine/src/builtins/promise/mod.rs
```diff
@@ -0,0 +1,731 @@
+//! This module implements the global `Promise` object.
+
+#![allow(dead_code, unused_results, unused_variables)]
```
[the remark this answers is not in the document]
Razican: Fixed

boa_engine/src/builtins/promise/mod.rs
```diff
@@ -0,0 +1,905 @@
...
+#[derive(Debug)]
+struct ResolvingFunctionsRecord {
+    resolve: JsValue,
+    reject: JsValue,
+}
+
+#[derive(Debug, Trace, Finalize)]
+struct RejectResolveCaptures {
+    promise: JsObject,
+    already_resolved: Gc<boa_gc::Cell<bool>>,
```
jedel1043, line 205: This is probably too much for a shared bool variable. A `Rc<Cell<bool>>` with the `#[unsafe_ignore_trace]` attribute should do the trick

Razican: I tried this, but I got 22 tests failing. The thing is that it does fancy stuff by sharing and mutating the variable from multiple places, sometimes in a row, and while failing a promise and so on.

jedel1043: That's pretty weird, because Gc<boa_gc::Cell<bool>> and Rc<Cell<bool>> should have exactly the same runtime behaviour. Even the gc crate tells us to "Avoid using Gc where Rc or Box would be equally usable."

Razican: I pushed the changes, maybe I missed something xD

Razican: It seems it worked somehow, but not in my laptop, who knows...

Razican:
Feel free to check this up and merge it. I will create separate PRs for async/await execution and the `[[Done]]` field in iterables, based on this one.

raskad (commented):
I got some inital comments. I will probably take a detailed look later today.

jasonwilliams (approved):

raskad (approved):
I think there are some more optimizations with some types, but let's go ahead and merge it and address those in follow ups. From here we got a really nice starting point into everything async.

Issue #2067, opened 2022-05-09
Implement legacy `Object.prototype` properties


raskad:
The following `Object.prototype` properties are marked as optional and legacy in the spec but we should implement them to pass the corresponding 262 tests.

- [x] [Object.prototype.\_\_proto__](https://tc39.es/ecma262/#sec-object.prototype.__proto__)
- [x] [Object.prototype.\_\_defineGetter__](https://tc39.es/ecma262/#sec-object.prototype.__defineGetter__)
- [x] [Object.prototype.\_\_defineSetter__](https://tc39.es/ecma262/#sec-object.prototype.__defineSetter__)
- [x] [Object.prototype.\_\_lookupGetter__](https://tc39.es/ecma262/#sec-object.prototype.__lookupGetter__)
- [ ] [Object.prototype.\_\_lookupSetter__](https://tc39.es/ecma262/#sec-object.prototype.__lookupSetter__)

PR #2108, opened 2022-06-10
[Merged by Bors] - Implement `Object.prototype.__defineGetter__`

CYBAI:
This Pull Request fixes part of #2067.

It introduces the legacy `Object.prototype.__defineGetter__` function.

### DIFF
--- a/boa_engine/src/builtins/object/mod.rs
+++ b/boa_engine/src/builtins/object/mod.rs
@@ -57,6 +57,7 @@ impl BuiltIn for Object {
         .method(Self::to_locale_string, "toLocaleString", 0)
         .method(Self::value_of, "valueOf", 0)
         .method(Self::is_prototype_of, "isPrototypeOf", 1)
+        .method(Self::legacy_define_getter, "__defineGetter__", 2)
         .static_method(Self::create, "create", 2)
         .static_method(Self::set_prototype_of, "setPrototypeOf", 2)
         .static_method(Self::get_prototype_of, "getPrototypeOf", 1)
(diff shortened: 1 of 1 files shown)


Razican:

  #### VM implementation
  | Total | 90,499 | 90,499 | 0 |
  | Passed | 56,372 | 56,384 | **+12** |
  | Ignored | 23,580 | 23,580 | 0 |
  | Failed | 10,547 | 10,535 | **-12** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 62.29% | 62.30% | **+0.01%** |
  <details><summary><b>Fixed tests (12):</b></summary>

  ```
  test/built-ins/Object/prototype/__defineGetter__/this-non-obj.js [strict mode] (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/this-non-obj.js (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/length.js [strict mode] (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/length.js (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/name.js [strict mode] (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/name.js (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/prop-desc.js [strict mode] (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/prop-desc.js (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/getter-non-callable.js [strict mode] (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/getter-non-callable.js (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/define-non-configurable.js [strict mode] (previously Failed)
  test/built-ins/Object/prototype/__defineGetter__/define-non-configurable.js (previously Failed)
  ```
  </details>

boa_engine/src/builtins/object/mod.rs
```diff
@@ -115,6 +116,41 @@ impl Object {
...
+        args: &[JsValue],
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        let getter = args.get_or_undefined(1);
+
+        let obj = this.to_object(context)?;
+
+        this.as_callable().ok_or_else(|| {
+            context.construct_type_error("Object.prototype.__defineGetter__: Expecting function")
+        })?;
```
raskad, line 140: This should test if `getter` is callable not `this`. From the spec: `If IsCallable(getter) is false`.

Also we can avoid the closure by using `getter.is_callable()` and then `return context.throw_type_error()` to throw the error.

raskad (changes requested):
Thank you for your contribution :)

I only have one small difference from the spec that should be fixed. Also we like to have the steps from the spec written as comments in the code (when it makes sense). For example:

```rust
// 1. Let O be ? ToObject(this value).
let obj = this.to_object(context)?;
```

It would be great if you could add those.

boa_engine/src/builtins/object/mod.rs
CYBAI: > This should test if getter is callable not this. From the spec: If IsCallable(getter) is false.

oh! thank you so much for catching it! I now figured out why some tests doesn't pass 🙏 

> Also we can avoid the closure by using getter.is_callable() and then return context.throw_type_error() to throw the error.

Thank you for the tip! I've updated the commit.

raskad (approved):
Looks great! All tests in the `built-ins/Object/prototype/__defineGetter__` suite pass :)

jedel1043 (approved):
Looks good!

CYBAI:
Thank you for the review! 🙏

PR #2114, opened 2022-06-11
[Merged by Bors] - Only run benchmarks on PRs when a label is set

raskad:
This changes our ci benchmarks to only run when the label `run-benchmark` is set on the PR.

The motivation is to reduce the time waiting on benchmarks to run while working on PRs. Also this saves some ci minutes which is always good.
When we spot changes that we suspect impact performance, we can add the `run-benchmark` label to the PR and the benchmarks will run.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,7 +1,18 @@
-on: [pull_request]
 name: Benchmarks
+
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+
 jobs:
   runBenchmark:
+    if: contains(github.event.pull_request.labels.*.name, 'run-benchmark')
     name: run benchmark
     runs-on: ubuntu-latest
     steps:


jedel1043 (approved):
Nice! This will definitely speedup our CI process by a lot :)

jedel1043:
I saw the test conformance comment and it occured to me that it would be good to conditionally run parts of the CI only when there has been changes on `boa_engine`, `boa_gc`, `boa_interner`, etc

raskad:
> I saw the test conformance comment and it occured to me that it would be good to conditionally run parts of the CI only when there has been changes on `boa_engine`, `boa_gc`, `boa_interner`, etc

True. I think there is some discussion needed on this, as some changes outside of the Rust crates probably also have to be tested. For example if we update the 262 submodule, we should probably run ci normally.

Razican:
The thing is that a change to, for example, the Unicode dependency, could make some random stuff to fail, and performance to change in any place (almost). So it's difficult to know when it makes sense or not.

raskad:
> The thing is that a change to, for example, the Unicode dependency, could make some random stuff to fail, and performance to change in any place (almost). So it's difficult to know when it makes sense or not.

That is true, but I would argue that those cases should not be the norm and we still have the full benchmarks on main. If any change unexpectedly changes performance, we can still track it down to the commit. IMO the benefits of getting faster feedback on PRs outweigh the negatives.

jasonwilliams (approved):
I’m assuming this works at the point when you add the label?

raskad:
> I’m assuming this works at the point when you add the label?

Yes, when you add the label it gets triggered for the first time. Any push after that also triggers them like in the past.

Razican (approved):
Looks good to me :)

Issue #2098, opened 2022-06-01
Tracking issue for `JsObject` safe wrappers 


jedel1043:
This is the main tracking issue listing all the `JsObject`s for which we could implement safe wrappers, and the implementation progress for them.

- [x] `JsArray` (#1746)
- [x] `JsFunction` (#2015)
- [x] `JsProxy` (#2076)
- [ ] `JsTypedArray` (partially on #2003 , missing API to pass a `JsArrayBuffer`)
- [ ] `JsArrayBuffer`
- [ ] `JsDataView`
- [ ] `JsMap`
- [ ] `JsSet`
- [ ] `JsDate`
- [ ] `JsGenerator`
- [ ] `JsRegExp`
- [ ] `JsGenerator`
- [ ] `JsGeneratorFunction`
- [ ] `JsWeakMap` (not implemented yet)
- [ ] `JsWeakSet` (not implemented yet)
- [ ] `JsPromise` (not implemented yet)
- [ ] `JsAsyncGenerator` (not implemented yet)

Feel free leave a comment if there's something missing.

nekevss:
Hi! Would it be fine to take a crack at one or two of these or is someone currently working on them?

Razican:
> Hi! Would it be fine to take a crack at one or two of these or is someone currently working on them?

Go ahead! Let us know in which ones you'll start working on :)

nekevss:
@Razican I'll start working on JsMap if no one else has started on it

jedel1043:
> @Razican I'll start working on JsMap if no one else has started on it

Nice! I'll edit the description to assign you that work :)

anuvratsingh:
@jedel1043 can I give `JsSet` a shot?

jedel1043:
@anuvratsingh Please do! If you need some guidance, you can post a comment here or in our discord :)

nekevss:
@jedel1043 Would it be fine to work on JsArrayBuffer and JsDataView?

jedel1043:
@nekevss That would be very nice :D

HalidOdat:
Ummm... I didn't see this conversation before I implemented  `JsArrayBuffer` (#2170 ) 😅 . I needed it to be implemented so I could fully fix (in a clean way) the `Vec<u8>` to `JsTypedArray` as discussed in #2058.

Though you work on `JsDataView` if you still want it

anuvratsingh:
[...]

PR #2115, opened 2022-06-11
[Merged by Bors] - Implementation of JsMap Wrapper

nekevss:
This Pull Request related to JsMap for #2098.

Any feedback on implementing JsMapIterator would be welcome. I wasn't entirely sure if it was the right approach, but as I worked on the example file, it felt like something at least similar would be needed to use Map's .entries(), .keys(), and .values() methods.

It changes the following:

- Implements JsMap Wrapper
- Implements JsMapIterator Wrapper
- Creates JsMap example in boa_examples


raskad:
I think this looks really good. The implementation of `JsMapIterator` looks totally logical to me. Is there a reason `forEach` is missing?

One little nitpick; there are some comments that don't start uppercase and end with a dot. Just for consistency and style ;)

And we probably need a rebase, because the 262 submodule seems to be outdated here.

nekevss:
I can clean up the comments real quick. They were mostly guideposts I was using while going.

There isn't much of a reason that `forEach` was left off. To be honest, I wasn't entirely certain about adding it as it seemed to be omitted from `JsArray`, which mostly made me pause and leave it off. I can definitely add `forEach`.

The 262 submodule is probably totally my bad. Any guidance on how I can fix it on my end would be really appreciated!

nekevss:
I can clean up the comments real quick. They were mostly guideposts I was using while going.

There isn't much of a reason that `forEach` was left off. To be honest, I wasn't entirely certain about adding it as it seemed to be omitted from `JsArray`, which mostly made me pause and leave it off. I can definitely add `forEach`.

The 262 submodule is probably totally my bad. Any guidance on how I can fix it on my end would be really appreciated!

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,190 @@
+// This module is a wrapper for the Map Builtin Javascript Object
+//
+// TODO: improve Iterator object interaction: entries, keys, values
+// forEach implementation is missing
```
raskad, line 4: This comment can be remove now right?

raskad:
> The 262 submodule is probably totally my bad. Any guidance on how I can fix it on my end would be really appreciated!

You can update the 262 submodule with `git submodule update --init`. You probably have to checkout `main`, pull the current changes, update the submodule there. Then you can rebase your branch on `main` and update the submodule there too. If I'm not mistaken that should do it.

boa_engine/src/object/jsmap.rs
nekevss: Removed this comment. Think it's a bit redundant. Although those methods will probably need to be addressed as IteratorRecord is altered.

nekevss:
@raskad I updated the comments and the submodule in main and ran the rebase. There were also some changes that needed to be made to `.keys()`, `.entries()`, and `.values()` since `IteratorRecord` had changed in one of the recent pull requests.

The submodule still didn't want to update on the branch with the `git submodule update --init` command, which was one of the issues I was having. I basically ended up going into test262 and pulling from `main` there. So hopefully that should be fixed too.

raskad:
@nekevss I think the problem was that you comitted a change to the submodule. I did an interactive rebase and explicitley left out the change to the submodule. That seems to do the trick :)

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,178 @@
...
+            .get_iterator(context, None, None)?;
+        let map_iterator_object = iterator_record.iterator();
+        JsMapIterator::from_object(map_iterator_object.clone(), context)
+    }
+
+    /// Insert a new entry into the Map object
+    #[inline]
+    pub fn set<T>(&self, key: T, value: T, context: &mut Context) -> JsResult<JsValue>
+    where
+        T: Into<JsValue>,
```
Razican, line 82: This has the limitation of both the `key` and the `value` needing to be of the same type. We don't really need this limitation. I would say, we should change the definition of the function:
```suggestion
    pub fn set<K, V>(&self, key: K, value: V, context: &mut Context) -> JsResult<JsValue>
    where
        K: Into<JsValue>,
        V: Into<JsValue>,
```

Razican (commented):
Looks very good! Thanks! I'm just missing some documentation and a couple of ergonomic improvements in the API.

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,178 @@
...
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        Map::for_each(
+            &self.inner.clone().into(),
+            &[callback.into(), this_arg],
+            context,
+        )
+    }
+
+    /// Returns new Iterator object of value elements of the Map.
```
jedel1043, line 145: Same comment as the `keys` method.

jedel1043 (changes requested):
Overall a very good work! I just have some grips with the documentation side of the contribution.

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // Utility function to generate default Map object.
+    fn create_map(context: &mut Context) -> JsObject {
+        // Get default Map prototype
+        let prototype = context.intrinsics().constructors().map().prototype();
+
+        // Create a default map object with [[MapData]] as a new empty list
+        JsObject::from_proto_and_data(prototype, ObjectData::map(OrderedMap::new()))
+    }
```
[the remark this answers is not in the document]
nekevss: I mostly used this for creating the default in `new` and `from_js_iterable`. I also figured it may be useful if there was ever the desire in the future to create more "from" methods. For instance, if there is a desire to create a `from_iter` method that takes a rust iterator and creates a map. It can be split into `new` and `from_js_iterable`, but would it be worth saving for now?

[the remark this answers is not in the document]
jedel1043: Hm, it can be left as it is then :)

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,178 @@
...
+    Context, JsResult, JsValue,
+};
+
+use boa_gc::{Finalize, Trace};
+use std::ops::Deref;
+
+#[derive(Debug, Clone, Trace, Finalize)]
+pub struct JsMap {
+    inner: JsObject,
+}
```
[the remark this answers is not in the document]
nekevss: I further expanded on the introduction/documentation and added a couple basic examples that display some ways to create a `JsMap` along with different methods.

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,178 @@
...
+
+impl JsMap {
+    /// Create new Empty Map Object.
```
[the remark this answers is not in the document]
nekevss: Updated and added an example :)

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[inline]
+    pub fn new(context: &mut Context) -> Self {
+        let map = Self::create_map(context);
+        Self { inner: map }
+    }
+
+    /// Create a new map object for any object that has a `@@Iterator` field.
```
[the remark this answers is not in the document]
nekevss: I updated the documentation and added both a valid and invalid example. Let me know what you think! :)

boa_engine/src/object/jsmap.rs
nekevss: Nice catch! I updated the types for `key` and `value`

nekevss:
I made some pretty big changes to the comments and documentation for `JsMap`, building out some of the comments and documentation/examples where possible. The documentation do vary between shorthand code snippets and longer examples, but they may have caused the continuous integration check fails. Do code snippets in comments typically cause these tests to fail or should the shorthand examples be lengthened?

Also, I did shy away from going too deep into the documentation surrounding `JsMapIterator` as I believe the `next` method returns an `IteratorResult`. I wasn't certain what the discussion was around `IteratorResult` and whether that should be exposed or handled within the `next` method of `JsMapIterator`, so I thought it best to leave it open for now

jedel1043:
@nekevss Very nice work!

The thing about Rust examples in docs is that `cargo`'s test framework will try to compile and test them to verify that they're valid.
This means you need to create a `context` for each example, just like the examples inside the struct doc.
Also, if you want to hide the initialization code, check https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example for code snippets on how to do that :D

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,344 @@
...
+///     let map = JsMap::new(context);
+///
+///     // Set key-value pairs for the `JsMap`.
+///     map.set("Key-1", "Value-1", context)?;
+///     map.set("Key-2", 10, context)?;
+///
+///     assert_eq!(map.get_size(context)?, 2.into());
+///
+///     Ok(())
+/// }
```
jedel1043, line 31: You can remove the `main` definition and use `unwrap` instead of `?`, since the functionality is the emphasis of the code snippet and not the error handling.
And just to clarify: doc tests automatically wrap the snippet in a `main` function, meaning you can omit it on all the doc examples

jedel1043: Also, if you do want to preserve the error handling, the rustdoc book has an explanation on how to do that:

https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#using--in-doc-tests

It's a bit more cumbersome but it would probably look better than using tons of `unwrap`s

nekevss: Thanks for the links to the docs! I swear I was looking for those, but I think I was in a completely different book. I believe I managed to clean up all the examples and add all the `unwrap`s where they were needed.

boa_engine/src/object/jsmap.rs
```diff
@@ -164,7 +164,7 @@ impl JsMap {
     ///
     /// let some_object = JsArray::new(context);
     ///
-    /// let js_map = JsMap::from_object(some_object, context).unwrap();
+    /// let js_map = JsMap::from_object(some_object.into(), context).unwrap();
```
jedel1043, line 167: Should panic is a bit of a last resort when testing if something must throw an error. I'd recommend using `is_err` with an `assert!` instead :)
Something like: `assert!(JsMap::from_object(some_object.into(), context).is_err());`

boa_engine/src/object/jsmap.rs
```diff
@@ -0,0 +1,426 @@
...
+/// `JsMap` provides a wrapper for Boa's implementation of the Javascript `Map` object.
+///
+/// # Examples
+///
+/// Create a `JsMap` and set a new entry
+/// ```
+/// use boa_engine::{
+///    object::JsMap,
+///    Context, JsValue,
+/// };
```
jedel1043, line 21:
```suggestion
/// # use boa_engine::{
/// #     object::JsMap,
/// #     Context, JsValue,
/// # };
```

I'd recommend hiding the imports to have a cleaner example on each doc. Also applies to every example

jedel1043 (changes requested):
Very very great work! I'm very impressed with the addition of so many example usages. I just have a single nit. After that, it's ready to merge.

boa_engine/src/object/jsmap.rs
nekevss: Agreed! Went through and hid the imports to clean up the examples. :)

jedel1043 (approved):
Looks good to me!

PR #2116, opened 2022-06-13
[Merged by Bors] - Implement `super` expressions

raskad:

- Implement `super` expression parsing / execution.
- Implement early errors for `super` expressions.
- Refactor / add internal slot representation for environment and function objects.

### DIFF
--- a/boa_engine/src/builtins/eval/mod.rs
+++ b/boa_engine/src/builtins/eval/mod.rs
@@ -79,16 +79,10 @@ impl Eval {
 
         // Because of implementation details the following code differs from the spec.
 
-        // Parse the script body (11.a - 11.d)
-        // TODO: Implement errors for 11.e - 11.h
-        let parse_result = if strict {
-            context.parse_strict(x.as_bytes())
-        } else {
-            context.parse(x.as_bytes())
-        };
-        let body = match parse_result.map_err(|e| e.to_string()) {
+        // Parse the script body and handle early errors (6 - 11)
+        let body = match context.parse_eval(x.as_bytes(), direct, strict) {
             Ok(body) => body,
-            Err(e) => return context.throw_syntax_error(e),
+            Err(e) => return context.throw_syntax_error(e.to_string()),
         };
 
         // 12 - 13 are implicit in the call of `Context::compile_with_new_declarative`.
--- a/boa_engine/src/builtins/function/mod.rs
+++ b/boa_engine/src/builtins/function/mod.rs
@@ -19,7 +19,7 @@ use crate::{
         internal_methods::get_prototype_from_constructor, JsObject, NativeObject, Object,
         ObjectData,
     },
-    object::{ConstructorBuilder, FunctionBuilder, Ref, RefMut},
+    object::{ConstructorBuilder, FunctionBuilder, JsFunction, PrivateElement, Ref, RefMut},
     property::{Attribute, PropertyDescriptor, PropertyKey},
     symbol::WellKnownSymbols,
     syntax::{ast::node::FormalParameterList, Parser},
@@ -108,7 +108,13 @@ impl ThisMode {
     }
 }
 
-#[derive(Debug, PartialEq, Clone, Copy)]
+/// Represents the `[[ConstructorKind]]` internal slot of function objects.
+///
+/// More information:
+///  - [ECMAScript specification][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-ecmascript-function-objects
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub enum ConstructorKind {
     Base,
     Derived,
@@ -124,6 +130,18 @@ impl ConstructorKind {
     pub fn is_derived(&self) -> bool {
         matches!(self, Self::Derived)
     }
+}
+
+/// Record containing the field definition of classes.
+///
+/// More information:
+///  - [ECMAScript specification][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-classfielddefinition-record-specification-type
+#[derive(Clone, Debug, Trace, Finalize)]
+pub enum ClassFieldDefinition {
+    Public(PropertyKey, JsFunction),
+    Private(Sym, JsFunction),
 }
 
 /// Wrapper for `Gc<GcCell<dyn NativeObject>>` that allows passing additional
@@ -191,6 +209,19 @@ pub enum Function {
     Ordinary {
         code: Gc<crate::vm::CodeBlock>,
         environments: DeclarativeEnvironmentStack,
+
+        /// The `[[ConstructorKind]]` internal slot.
+        #[unsafe_ignore_trace]
+        constructor_kind: ConstructorKind,
+
+        /// The `[[HomeObject]]` internal slot.
+        home_object: Option<JsObject>,
+
+        /// The `[[Fields]]` internal slot.
+        fields: Vec<ClassFieldDefinition>,
+
+        /// The `[[PrivateMethods]]` internal slot.
+        private_methods: Vec<(Sym, PrivateElement)>,
     },
     Generator {
         code: Gc<crate::vm::CodeBlock>,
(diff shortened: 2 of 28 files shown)


boa_engine/src/builtins/function/mod.rs
```diff
@@ -210,6 +241,79 @@ impl Function {
...
+            None
+        }
+    }
+
+    ///  Sets the `[[HomeObject]]` slot if present.
+    pub(crate) fn set_home_object(&mut self, object: &JsObject) {
+        if let Self::Ordinary { home_object, .. } = self {
+            *home_object = Some(object.clone());
+        }
+    }
```
Razican, line 271: Here, I would pass the `JsObject` by value, since we use it like that, and can avoid a clone in some cases (even if that clone is just copying a pointer value).

boa_engine/src/builtins/eval/mod.rs
```diff
@@ -83,12 +83,10 @@ impl Eval {
...
         // TODO: Implement errors for 11.e - 11.h
-        let parse_result = if strict {
-            context.parse_strict(x.as_bytes())
-        } else {
-            context.parse(x.as_bytes())
-        };
-        let body = match parse_result.map_err(|e| e.to_string()) {
+        let body = match context
+            .parse_eval(x.as_bytes(), direct, strict)
+            .map_err(|e| e.to_string())
```
Razican, line 88: Could we avoid the closure here, and directly point to the `to_string` function?

boa_engine/src/context/mod.rs
```diff
@@ -162,14 +162,21 @@ impl Context {
         parser.parse_all(self)
     }
 
-    /// Parse the given source text in strict mode.
-    pub(crate) fn parse_strict<S>(&mut self, src: S) -> Result<StatementList, ParseError>
+    /// Parse the given source text with eval specific handling.
+    pub(crate) fn parse_eval<S>(
```
Razican, line 166: We can probably use this in `boa_tester`, in the `$262` object that gets passed to the context. It has a special `eval` function that currently does the parsing twice, but can probably benefit from this.

boa_engine/src/context/mod.rs
```diff
@@ -683,22 +690,20 @@ impl Context {
...
             finally_jump: Vec::new(),
             pop_on_return: 0,
-            loop_env_stack: vec![0],
-            try_env_stack: vec![crate::vm::TryStackEntry {
+            loop_env_stack: Vec::from([0]),
+            try_env_stack: Vec::from([crate::vm::TryStackEntry {
                 num_env: 0,
                 num_loop_stack_entries: 0,
-            }],
+            }]),
```
Razican, line 706: What is the benefit of using `Vec::from()`?

raskad: I though it may generate more efficient code, but looking at the compiler output, it seems like it's exactly the same. Probably it's better for compile times, because rustc has to do less macro expansion.

boa_engine/src/context/mod.rs
Razican: Ye, that's what I meant xD didn't remember it well. But yes, I get that. Maybe we can do the same somehow in `evalScript()`.

boa_engine/src/builtins/eval/mod.rs
```diff
@@ -81,14 +81,10 @@ impl Eval {
...
-        // TODO: Implement errors for 11.e - 11.h
-        let body = match context
-            .parse_eval(x.as_bytes(), direct, strict)
-            .map_err(|e| e.to_string())
-        {
+        // Parse the script body and handle early errors (6 - 11)
+        let body = match context.parse_eval(x.as_bytes(), direct, strict) {
             Ok(body) => body,
-            Err(e) => return context.throw_syntax_error(e),
+            Err(e) => return context.throw_syntax_error(e.to_string()),
```
Razican, line 87: That works too, yes, but my thoughts were more on using the following:
```rust
map_err(ToString::to_string)
```

instead of adding a closure in `map_err`, but maybe it wasn't possible, due to by value/by reference function signatures.

boa_engine/src/syntax/ast/node/declaration/mod.rs
```diff
@@ -229,6 +229,30 @@ impl Declaration {
             Self::Pattern(pattern) => pattern.init(),
         }
     }
+
+    /// Returns true if the node contains the given token.
```
Razican, line 233:
```suggestion
    /// Returns `true` if the node contains the given token.
```

boa_engine/src/syntax/ast/node/declaration/mod.rs
```diff
@@ -383,6 +407,99 @@ impl DeclarationPattern {
         }
         false
     }
+
+    /// Returns true if the node contains the given token.
```
Razican, line 411:
```suggestion
    /// Returns `true` if the node contains the given token.
```

boa_engine/src/syntax/ast/node/field/get_super_field/mod.rs
```diff
@@ -0,0 +1,47 @@
+use crate::syntax::ast::node::Node;
+use boa_interner::{Interner, Sym, ToInternedString};
+
+#[cfg(feature = "deser")]
+use serde::{Deserialize, Serialize};
+
+/// The super keyword is used to access fields on an object's parent.
```
Razican, line 7:
```suggestion
/// The `super` keyword is used to access fields on an object's parent.
```

boa_engine/src/syntax/ast/node/super_call/mod.rs
```diff
@@ -0,0 +1,46 @@
+use crate::syntax::ast::node::{join_nodes, Node};
+use boa_interner::{Interner, ToInternedString};
+
+#[cfg(feature = "deser")]
+use serde::{Deserialize, Serialize};
+
+///  The super keyword is used to access and call functions on an object's parent.
```
Razican, line 7:
```suggestion
/// The `super` keyword is used to access and call functions on an object's parent.
```

boa_engine/src/syntax/parser/expression/primary/object_initializer/mod.rs
```diff
@@ -173,17 +174,31 @@ where
...
 
                 let token = cursor.peek(0, interner)?.ok_or(ParseError::AbruptEnd)?;
+                let position = token.span().start();
+
                 if let TokenKind::Punctuator(Punctuator::Mul) = token.kind() {
                     let (property_name, method) =
                         AsyncGeneratorMethod::new(self.allow_yield, self.allow_await)
                             .parse(cursor, interner)?;
+
+                    // It is a Syntax Error if HasDirectSuper of MethodDefinition is true.
```
Razican, line 184:
```suggestion
                    // It is a Syntax Error if `HasDirectSuper` of `MethodDefinition` is `true`.
```

Razican: Also, should this be named `has_direct_super()` instead of `function_contains_super_call()`?

boa_engine/src/syntax/parser/expression/primary/object_initializer/mod.rs
```diff
@@ -173,17 +174,31 @@ where
...
+
                     return Ok(object::PropertyDefinition::method_definition(
                         method,
                         property_name,
                     ));
                 }
                 let (property_name, method) =
                     AsyncMethod::new(self.allow_yield, self.allow_await).parse(cursor, interner)?;
+
+                // It is a Syntax Error if HasDirectSuper of MethodDefinition is true.
```
Razican, line 197:
```suggestion
                // It is a Syntax Error if `HasDirectSuper` of `MethodDefinition` is `true`.
```

Razican: Should this comment be added to other places where this function is called?

boa_engine/src/syntax/ast/node/mod.rs
```diff
@@ -880,6 +893,361 @@ impl Node {
         }
         false
     }
+
+    /// Returns true if the node contains the given token.
```
Razican, line 897:
```suggestion
    /// Returns `true` if the node contains the given token.
```

boa_engine/src/vm/mod.rs
```diff
@@ -1481,19 +1941,22 @@ impl Context {
...
+                //    self.vm.frame().this.clone()
+                //} else {
+                //    JsValue::undefined()
+                //};
+                //
+                //self.realm.environments.push_function(
+                //    num_bindings as usize,
+                //    compile_environment,
+                //    this,
+                //);
```
Razican, line 1956: I guess we no longer need this, right?

Razican (approved):
I've reviewed it, but I'm not confident enough with the VM. Looks pretty good, but I expect a hit in Parser performance, due to the tree traversal.

I added some comments, mostly related to documentation, but now that I think of it, adding backticks everywhere might not make much sense xD just give it a look and let me know :)

boa_engine/src/syntax/parser/expression/primary/object_initializer/mod.rs
raskad: That is very true. It's basically that function in the spec.

I have not added the backticks for these comments, as they are just copied from the spec. I think it makes less sense on those.

HalidOdat (approved):
This Looks good from my side! :)

PR #2119, opened 2022-06-13
[Merged by Bors] - Prevent PR benchmarks to run on label changes

raskad:
This changes the trigger type for PR benchmarks back to the default (`opened`, `synchronize`, `reopened`).  As part of #2114 I added the `labeled` trigger type. This causes the benchmarks to run when the `run-benchmark` label is present and another label is added.
For example in #2116 I added the `run-benchmark` label while creating the PR. The benchmarks then where triggered six times; one for the PR creation (`opened`) and five times for each label that I initially added to the PR.

The only drawback is that the benchmarks are not triggered, when we just add the label, but unfortunately I don't have a clever idea on how to achieve that right now. We will have to add the label and then trigger the run via a `synchronize` (push).

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -4,11 +4,6 @@ on:
   pull_request:
     branches:
       - main
-    types:
-      - opened
-      - synchronize
-      - reopened
-      - labeled
 
 jobs:
   runBenchmark:


jedel1043 (approved):
Unfortunate. Still pretty usable though

Razican (approved):

PR #2125, opened 2022-06-15
[Merged by Bors] - Started adding the [[Done]] field to iterators

Razican:
This Pull Request adds the `[[Done]]` field to iterator records.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,7 +75,7 @@ version = "0.15.0"
 dependencies = [
  "boa_engine",
  "boa_interner",
- "clap 3.1.18",
+ "clap 3.2.5",
  "colored",
  "jemallocator",
  "phf",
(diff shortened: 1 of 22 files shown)


Razican:
I am probably missing stuff on the VM, but I'm somehow not able to make this work :(

raskad:
I got the vm code to work. Only a few tests are missing. I'm currently looking into those.

One big change is, that the `[[NextMethod]]` can be a value instead of an object. Imo the spec is just wrong with the field description of `[[NextMethod]]`. It sounds like it can only be a function: [`a function object`](https://tc39.es/ecma262/#table-iterator-record-fields). But in [`GetIterator`](https://tc39.es/ecma262/#sec-getiterator) it clearly can be a value:
```txt
5. Let nextMethod be ? GetV(iterator, "next").
6. Let iteratorRecord be the Iterator Record { [[Iterator]]: iterator, [[NextMethod]]: nextMethod, [[Done]]: false }.
```

Razican:
Thanks @raskad! I will now add the logic to the Promises :)

Razican:
The two test failures are due to changes in Test262. Not related to the PR. I think this is ready for review :)

raskad (approved):
Looks good to me, if we want to include the dependency updates here.

jedel1043 (approved):
Looks good!

PR #2131, opened 2022-06-20
Introduce NonMaxU32 for array index

CYBAI:
It changes the following:

- Introduce `NonMaxU32` type
- Use `NonMaxU32` for index property key instead of `u32`

---

While looking into the failed `built-ins/Reflect/ownKeys/return-on-corresponding-order-large-index.js` in test262, I found that the root cause is related to [`array index`](https://tc39.es/ecma262/#array-index) for property keys.

The root cause of the failure is, the test uses `u32::MAX` as key and we're ordering it wrong because we see `u32::MAX` as a [array index](https://tc39.es/ecma262/#array-index) instead of an [integer index](https://tc39.es/ecma262/#integer-index). Based on the spec, array index's range is from 0 <= n < u32::MAX so the max number is exclusive.

We can avoid transforming `u32::MAX` into property key in some functions like `From<T>` traits though, it might be forgettable and error-prone. So, I wonder it might be good to introduce the `NonMaxU32` type which is similar to `NonZeroU32` in std but excluding max instead.

Then, we can use it for the index property key

```diff
-    Index(u32),
+    Index(NonMaxU32),
```

With doing so, we can always guarantee the integer hold by Index property key is valid.

WDYT?

(On the other hand, I'm not sure where is the best place for such types. Please feel free to let me know if I should move it to other places 🙏)

### DIFF
--- a/boa_engine/src/nonmaxu32.rs
+++ b/boa_engine/src/nonmaxu32.rs
@@ -1,0 +1,180 @@
+//! This module implements `NonMaxU32`
+//! which is known not to equal `u32::MAX`.
+//!
+//! This would be useful for integers like `https://tc39.es/ecma262/#array-index`.
+
+use boa_gc::{unsafe_empty_trace, Finalize, Trace};
+use std::fmt;
+use std::ops::{BitOr, BitOrAssign};
+
+/// An integer that is known not to equal `u32::MAX`.
+///
+/// This enables some memory layout optimization.
+/// For example, `Option<NonMaxU32>` is the same size as `u32`:
+///
+/// ```rust
+/// use std::mem::size_of;
+/// use boa_engine::nonmaxu32::NonMaxU32;
+/// assert_eq!(size_of::<Option<NonMaxU32>>(), size_of::<u32>());
+/// ```
+#[derive(Copy, Clone, Eq, Finalize, PartialEq, Ord, PartialOrd, Hash)]
+#[repr(transparent)]
+pub struct NonMaxU32(u32);
+
+// Safety: `NonMaxU32` does not contain any objects which needs to be traced,
+// so this is safe.
+unsafe impl Trace for NonMaxU32 {
+    unsafe_empty_trace!();
+}
+
+/// An error type returned when a checked integral type conversion fails (mimics [`std::num::TryFromIntError`])
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TryFromIntError(());
+
+impl fmt::Display for TryFromIntError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        "out of range integral type conversion attempted".fmt(fmt)
+    }
+}
+
+impl From<core::num::TryFromIntError> for TryFromIntError {
+    fn from(_: core::num::TryFromIntError) -> Self {
+        Self(())
+    }
+}
+
+impl From<core::convert::Infallible> for TryFromIntError {
+    fn from(never: core::convert::Infallible) -> Self {
+        match never {}
+    }
+}
+
+/// An error type returned when an integer cannot be parsed (mimics [`std::num::ParseIntError`])
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct ParseIntError(());
+
+impl fmt::Display for ParseIntError {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        "unable to parse integer".fmt(fmt)
+    }
+}
+
+impl From<core::num::ParseIntError> for ParseIntError {
+    fn from(_: core::num::ParseIntError) -> Self {
+        Self(())
+    }
+}
+
+impl NonMaxU32 {
+    /// Creates a non-u32-max without checking the value.
+    ///
+    /// # Safety
+    ///
+    /// The value must not be `u32::MAX`.
+    #[inline]
+    pub const unsafe fn new_unchecked(n: u32) -> Self {
+        // SAFETY: this is guaranteed to be safe by the caller.
+        Self(n)
+    }
+
+    /// Creates a non-u32-max if the given value is not `u32::MAX`.
+    #[inline]
+    pub const fn new(n: u32) -> Option<Self> {
+        if n == u32::MAX {
+            None
+        } else {
+            // SAFETY: we just checked that there's no `u32::MAX`
+            Some(Self(n))
+        }
+    }
+
+    /// Returns the value as a primitive type.
+    #[inline]
+    pub const fn get(self) -> u32 {
+        self.0
+    }
+}
+
+impl TryFrom<u32> for NonMaxU32 {
+    type Error = TryFromIntError;
+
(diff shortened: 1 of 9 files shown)


Razican:
  
  #### VM implementation
  | Total | 90,619 | 90,619 | 0 |
  | Passed | 56,859 | 56,861 | **+2** |
  | Ignored | 13,788 | 13,788 | 0 |
  | Failed | 19,972 | 19,970 | **-2** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 62.75% | 62.75% | **+0.00%** |
  
  ```
  test/built-ins/Reflect/ownKeys/return-on-corresponding-order-large-index.js [strict mode] (previously Failed)
  test/built-ins/Reflect/ownKeys/return-on-corresponding-order-large-index.js (previously Failed)
  ```
  </details>

boa_engine/src/nonmaxu32.rs
```diff
@@ -0,0 +1,180 @@
...
+/// ```
+#[derive(Copy, Clone, Eq, Finalize, PartialEq, Ord, PartialOrd, Hash)]
+#[repr(transparent)]
+pub struct NonMaxU32(u32);
+
+// Safety: `NonMaxU32` does not contain any objects which needs to be traced,
+// so this is safe.
+unsafe impl Trace for NonMaxU32 {
+    unsafe_empty_trace!();
+}
```
Razican, line 28: I would suggest to just not implement this, and use `#[unsafe_ignore_trace]` in the `enum` (if that's allowed). It should be more performant.

boa_engine/src/nonmaxu32.rs
```diff
@@ -0,0 +1,180 @@
...
+/// For example, `Option<NonMaxU32>` is the same size as `u32`:
+///
+/// ```rust
+/// use std::mem::size_of;
+/// use boa_engine::nonmaxu32::NonMaxU32;
+/// assert_eq!(size_of::<Option<NonMaxU32>>(), size_of::<u32>());
+/// ```
+#[derive(Copy, Clone, Eq, Finalize, PartialEq, Ord, PartialOrd, Hash)]
+#[repr(transparent)]
+pub struct NonMaxU32(u32);
```
Razican, line 22: I would suggest the internal type to be not an enum type and be a structure type:
```rust
pub struct NonMaxU32 {
    inner: u32,
}
```

This would disallow direct editing of the internal structure (having to use getters and setters), and changing the internal type would be API compatible.

Talking about changing the internal type, it might be useful to use a `NonZeroU32`, which is better optimized for space. On the other hand, this would require using a `+1` when storing, and a `-1` when retrieving, which could cause a small performance penalty. We should test both options.

boa_engine/src/nonmaxu32.rs
```diff
@@ -0,0 +1,180 @@
...
+            // SAFETY: we just checked that there's no `u32::MAX`
+            Some(Self(n))
+        }
+    }
+
+    /// Returns the value as a primitive type.
+    #[inline]
+    pub const fn get(self) -> u32 {
+        self.0
+    }
```
Razican, line 95: Should we also include setters? And operations such as add, sub, mul and so on?

boa_engine/src/nonmaxu32.rs
```diff
@@ -0,0 +1,180 @@
...
+        *self = *self | rhs;
+    }
+}
+
+impl fmt::Debug for NonMaxU32 {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.get().fmt(f)
+    }
+}
```
Razican, line 173: This might be a good implementation for `Display`, but I would say the `Debug` implementation shouldn't hide the internal structure. I'm open to suggestions, though.

Razican (commented):
Looks very good! Thanks for the fix. I added some comments regarding some potential enhancements. I would like to also get inputs from the rest of the team, since some of the options might just be too much, or might not make sense.

Oh, and tests seem to be failing in MacOS and Windows, so these need to be fixed.

boa_engine/src/nonmaxu32.rs
CYBAI: Sorry for late reply 🙏 Thank you for your suggestion! I will try to work on a fix.

> On the other hand, this would require using a +1 when storing, and a -1 when retrieving, which could cause a small performance penalty.

This sounds a good idea to me. I experimented a bit using XOR to save with `NonZeroU32` like [the nonmax crate does](https://github.com/LPGhatguy/nonmax/blob/d1d4259d4c86646e4e4fce5078bdd216b7042f65/src/lib.rs#L128-L133) last week but I realized we would need to `rev()` the iterator because the xor'ed value would be in reversed order. So, I wonder just +1 and -1 would be easier.

I will try it more and share more information here or on Discord 🙏  Thank you!

boa_engine/src/nonmaxu32.rs
CYBAI: I'm good with adding these methods though, I wonder it also depends on how we want to use it. Do we want to always do calculation with `NonMaxU32` directly? 

On the other hand, I wonder it might be good to add methods like [`checked_add`](https://doc.rust-lang.org/std/num/struct.NonZeroU32.html#method.checked_add) or [`saturating_add`](https://doc.rust-lang.org/std/num/struct.NonZeroU32.html#method.saturating_add) on `NonZeroU32`.

WDYT?

boa_engine/src/nonmaxu32.rs
CYBAI: I tried to follow `NonZeroU32` debug though, it sounds fine to me to format it like `NonMaxU32 { inner: 42 }`.

boa_engine/src/nonmaxu32.rs
```diff
@@ -0,0 +1,180 @@
...
+    type Output = NonMaxU32;
+
+    #[inline]
+    fn bitor(self, rhs: NonMaxU32) -> Self::Output {
+        // Safety: since `rhs` is nonzero, the result of the
+        // bitwise-or will be nonzero regardless of the value of
+        // `self`.
+        unsafe { NonMaxU32::new_unchecked(self | rhs.get()) }
+    }
+}
```
peddermaster2, line 152: The "Safety" argument and code seems to be copied from the nonzero implementaion and is not correct. (At least the comment is wrong)

boa_engine/src/nonmaxu32.rs
RageKnify: Agree with Razican

boa_engine/src/nonmaxu32.rs
RageKnify: I think I'm ok with either approach (-1/+1 or XOR).

RageKnify (changes requested):
Need to introduce +1/-1 or XOR on accesses. fix the safety comments and correct the safety comments.

jasonwilliams:
Hey @CYBAI how is this going do you need any help?

CYBAI:
> Hey @CYBAI how is this going do you need any help?

Sorry for very late reply 🙏 I'll try to find some time to address the comments and update the PR after mid-Jan 🙇

raskad:
Added in #3321

PR #2135, opened 2022-06-20
Fix for in/of loop initializer environment

raskad:

- Add an additional environment when a for in/of loop is initialized with a `let` or `const` binding.


jedel1043 (approved):
Looks good!

HalidOdat (approved):
Looks good to me!

PR #2136, opened 2022-06-21
[Merged by Bors] - Refactor `construct` and `PromiseCapability` to preserve `JsObject` invariants 

jedel1043:
This Pull Request changes the signature of `construct` to always return `JsObject`s, and refactors `PromiseCapability` to store only `JsObject`/`JsFunction`s. This preserves the following invariants specified in the spec:

https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-ecmascript-function-objects-construct-argumentslist-newtarget
> The [[Construct]] internal method of an ECMAScript function object ... returns either a normal completion containing an Object or a throw completion ... 

https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promisecapability-records

Table 82: [PromiseCapability Record](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promisecapability-records) Fields
Field Name | Value | Meaning
-- | -- | --
[[Promise]] | an Object | An object that is usable as a promise.
[[Resolve]] | a function object | The function that is used to resolve the given promise.
[[Reject]] | a function object | The function that is used to reject the given promise.

### DIFF
--- a/boa_engine/src/builtins/array_buffer/mod.rs
+++ b/boa_engine/src/builtins/array_buffer/mod.rs
@@ -243,13 +243,9 @@ impl ArrayBuffer {
         // 16. Let new be ? Construct(ctor, « 𝔽(newLen) »).
         let new = ctor.construct(&[new_len.into()], Some(&ctor), context)?;
 
-        // 17. Perform ? RequireInternalSlot(new, [[ArrayBufferData]]).
-        let new_obj = new.as_object().cloned().ok_or_else(|| {
-            context.construct_type_error("ArrayBuffer constructor returned non-object value")
-        })?;
-
         {
-            let new_obj = new_obj.borrow();
+            let new_obj = new.borrow();
+            // 17. Perform ? RequireInternalSlot(new, [[ArrayBufferData]]).
             let new_array_buffer = new_obj.as_array_buffer().ok_or_else(|| {
                 context.construct_type_error("ArrayBuffer constructor returned invalid object")
             })?;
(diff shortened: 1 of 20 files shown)


boa_engine/src/vm/code_block.rs
```diff
@@ -1034,10 +1054,21 @@ impl JsObject {
...
-                    Ok(frame.this.clone())
+                match result {
+                    JsValue::Object(ref obj) => Ok(obj.clone()),
+                    val => {
+                        if constructor.is_base() || val.is_undefined() {
+                            Ok(frame
+                                .this
+                                .as_object()
+                                .cloned()
+                                .expect("13. Assert: Type(thisBinding) is Object."))
```
Razican, line 1065: I'm guessing all this comes directly from the spec, right? Can we add comments from it as we do in other places?

Razican (commented):
Looks very good :) I added some suggestions for improvements

boa_engine/src/builtins/array_buffer/mod.rs
```diff
@@ -264,44 +260,46 @@ impl ArrayBuffer {
             }
         }
         // 20. If SameValue(new, O) is true, throw a TypeError exception.
-        if JsValue::same_value(&new, this) {
+        if this.as_object().map(|obj| obj == &new).unwrap_or_default() {
```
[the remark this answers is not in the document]
jedel1043: Used `JsObject::equals` which is exactly the same

boa_engine/src/builtins/function/mod.rs
```diff
@@ -178,12 +183,12 @@ pub enum Function {
     Native {
         #[unsafe_ignore_trace]
         function: NativeFunctionSignature,
-        constructor: bool,
+        constructor: Option<ConstructorKind>,
```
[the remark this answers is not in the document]
jedel1043: Yep

boa_engine/src/builtins/promise/mod.rs
```diff
@@ -185,7 +185,7 @@ impl PromiseCapability {
                 }
 
                 // 9. Set promiseCapability.[[Promise]] to promise.
-                promise_capability.reject = promise;
+                promise_capability.reject = promise.into();
```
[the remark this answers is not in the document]
jedel1043: Extended the invariants for promises

boa_engine/src/vm/code_block.rs
jedel1043: I added only that comment because it was appropriate for the expect. However, the definition of `construct` differs a lot from the spec, so it cannot be documented like every other function for now. I'll reanalyze `construct` to see if it can be rewritten in a way that conforms with the spec, so that we can document the whole function

raskad (approved):
Nice change!

boa_engine/src/vm/code_block.rs
Razican: Sounds perfect to me :) thanks!!

boa_engine/src/object/jsfunction.rs
```diff
@@ -17,16 +17,14 @@ impl JsFunction {
         Self { inner: object }
     }
 
-    /// Create a [`JsFunction`] from a [`JsObject`], if the object is not a function throw a `TypeError`.
+    /// Create a [`JsFunction`] from a [`JsObject`], or return None if the object is not a function.
```
Razican, line 20:
```suggestion
    /// Create a [`JsFunction`] from a [`JsObject`], or return `None` if the object is not a function.
```

Razican (approved):
Great work! This is a very cool improvement for the internal invariants, and I really liked the enhancements for Promises.

PR #2137, opened 2022-06-21
Make all built-in constructors public

jedel1043:
This Pull Request makes all our built-in constructors public. This should ensure our users can still construct objects from Rust without having to compile and run any JS code.

### DIFF
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -36,7 +36,7 @@ use std::cmp::{max, min, Ordering};
 
 /// JavaScript `Array` built-in implementation.
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct Array;
+pub struct Array;
 
 impl BuiltIn for Array {
     const NAME: &'static str = "Array";
@@ -133,7 +133,7 @@ impl BuiltIn for Array {
 impl Array {
     const LENGTH: usize = 1;
 
-    fn constructor(
+    pub fn constructor(
         new_target: &JsValue,
         args: &[JsValue],
         context: &mut Context,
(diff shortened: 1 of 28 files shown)


jedel1043:
Marking as draft since it's missing some documentation

jedel1043:
So, I just realized we don't really need to expose the constructors because we already have them exposed in the intrinsics of `Context`:
```Rust
let constructor = context.intrinsics().constructors().$constructor_type().constructor();
let obj = constructor.construct(args, None, context)?;
```

However, it can be argued that is not as "pretty" as calling

```Rust
let obj = ConstructorType::constructor(context.intrinsics().constructors().$constructor_type().prototype(), args, context)?;
```

So exposing the constructors has still some readability value.
On the other hand, a user doesn't really need to know the internals of how constructors work in order to use the first method, but with the second method they would need to know to always pass the correct prototype every time.

raskad:
> I'm inclined for the first method, because it's easier to use and it just needs some documentation with examples, unlike the second method which exposes a big chunk of new methods. What do you think?

I completely agree with this. I think we could also build some abstraction based on that method in the future, if the readability becomes a big enough issue.

dOrgJelli:
I believe this to be a blocking issue for my use-case, where I'd like to initialize the context with a custom set of built-ins, which requires these builtin modules (structs, constructors, etc) to be accessible outside of the boa_engine itself. I posted more about this here: https://discord.com/channels/595323158140158003/595382036215365633/1031513873636917319

jedel1043:
Gonna close this because advanced users can access all builtin objects from the `Intrinsics` struct, so exposing all constructors is not necessary. We can reopen this in the future if there's a need to expose the builtins.

PR #2140, opened 2022-06-25
[Merged by Bors] - Implement `Promise.all`

raskad:

- Implement `Promise.all`

### DIFF
--- a/boa_engine/src/builtins/promise/mod.rs
+++ b/boa_engine/src/builtins/promise/mod.rs
@@ -20,7 +20,7 @@ use crate::{
     value::JsValue,
     Context, JsResult,
 };
-use boa_gc::{Finalize, Gc, Trace};
+use boa_gc::{Cell as GcCell, Finalize, Gc, Trace};
 use boa_profiler::Profiler;
 use std::{cell::Cell, rc::Rc};
 use tap::{Conv, Pipe};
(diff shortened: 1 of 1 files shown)


HalidOdat (approved):
Looks perfect to me! :)

boa_engine/src/builtins/promise/mod.rs
```diff
@@ -316,6 +317,247 @@ impl Promise {
...
+            capability_resolve: JsFunction,
+            #[unsafe_ignore_trace]
+            remaining_elements_count: Rc<Cell<i32>>,
+        }
+
+        // 1. Let values be a new empty List.
+        let values = GcCell::new(Vec::new());
+
+        // 2. Let remainingElementsCount be the Record { [[Value]]: 1 }.
+        let remaining_elements_count = Rc::new(Cell::new(1));
```
Razican, line 410: As far as I can tell, the reason behind this being inside cells is that we need to pass them as captures below, right? But we don't use them until step `j`, if I'm not mistaken. Wouldn't it be more efficient to wrap them in cells after the `loop` has completed?

Razican (approved):
Looks good, thank you!
I'm happy to merge it from my side, but I'd like to point out a potential performance optimization. Give it a look, and if it makes sense, feel free to implement it and merge.

PR #2147, opened 2022-06-28
Remove `string-interner` dependency and implement custom string `Interner`

jedel1043:
So, @raskad and myself had a short discussion about the state of #736, and we came to the conclusion that it would be a good time to implement our own string interner; partly because the `string-interner` crate is a bit unmaintained (as shown by https://github.com/Robbepop/string-interner/pull/42 and https://github.com/Robbepop/string-interner/pull/47), and partly because it would be hard to experiment with custom optimizations for UTF-16 strings. I still want to thank @Robbepop for the original implementation though, because some parts of this design have been shamelessly stolen from it 😅.

Having said that, this PR is a complete reimplementation of the interner, but with some modifications to (hopefully!) make it a bit easier to experiment with UTF-16 strings, apply optimizations, and whatnot :)


boa_interner/src/interned_str.rs
```diff
@@ -0,0 +1,70 @@
+use std::{borrow::Borrow, ptr::NonNull};
+
+/// Wrapper for an interned str pointer, required to
+/// quickly check using a hash if a string is inside an [`Interner`].
```
Razican, line 4: It seems that the link here doesn't work, as per the warning.

boa_interner/src/interned_str.rs
```diff
@@ -0,0 +1,70 @@
...
+///
+/// # Safety
+///
+/// This struct could cause Undefined Behaviour on:
+/// - Use without ensuring the referenced memory is still allocated.
+/// - Construction of an [`InternedStr`] from an invalid [`NonNull<str>`].
+///
+/// In general, this should not be used outside of an [`Interner`].
```
Razican, line 12: This link seems to also be failing.

boa_interner/src/sym.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Returns the internal value of the [`Sym`]
+    pub(super) const fn get(self) -> usize {
+        self.value.get()
+    }
+}
+
+// Safe because `Sym` implements `Copy`.
+unsafe impl Trace for Sym {
+    unsafe_empty_trace!();
+}
```
Razican, line 118: Should this be implemented, or should we force users of `Sym` to use `#[unsafe_ignore_trace]`?

Razican (approved):
This looks good :) what are the main differences with the upstream implementation?

I just noticed that some documentation links are failing (give a look to the lints in GitHub, some are even in non-modified parts of the code).

boa_interner/src/lib.rs
```diff
@@ -119,21 +131,42 @@ impl Interner {
...
+        }
+
+        let capacity = self.head.capacity();
+
+        if capacity < self.head.len() + string.len() {
+            let new_cap = (usize::max(capacity, string.len()) + 1).next_power_of_two();
+            let new_head = String::with_capacity(new_cap);
+            let old_head = std::mem::replace(&mut self.head, new_head);
+            self.full.push(old_head);
+        }
```
raskad, line 160: I think it would be nice to have a comment explaining why this is done and how the pointers into the old `head` are still valid even after pushing it to `full`. Probably that comment could also be on the fields of `Interner` directly.

boa_interner/src/sym.rs
jedel1043: Had to do some shenanigans with the implementation of `Object`, but with that `Sym` no longer depends on `gc`

boa_interner/src/interned_str.rs
jedel1043: Fixed

boa_interner/src/interned_str.rs
jedel1043: Fixed

boa_interner/src/lib.rs
jedel1043: Changed a bit the implementation to better reflect the unsafe parts, and documented some other parts with what you've mentioned

boa_interner/src/lib.rs
```diff
@@ -150,24 +183,26 @@ impl Interner {
...
-        if index <= Self::STATIC_STRINGS.len() {
-            Some(Self::STATIC_STRINGS[index - 1])
-        } else {
-            self.inner.resolve(symbol)
-        }
+        let index = symbol.get();
+
+        COMMON_STRINGS.index(index - 1).copied().or_else(|| {
+            self.spans
+                .get(symbol.get() - COMMON_STRINGS.len() - 1)
```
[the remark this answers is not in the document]
jedel1043: Yep, changed it a bit by making `index = symbol.get() - 1`, since all indexes are offset by one because of the `NonZeroUsize`, so that seemed prudent

jedel1043:
> This looks good :) what are the main differences with the upstream implementation?

Essentially I've inlined the `backend` concept inside the `Interner` itself, which makes the implementation a lot more simpler. Also, I've changed the symbol cache to use `FxHash` instead of `brownhash`, which should be a bit more performant, but I think the most important change is that, since `Interner` can see our `COMMON_STRINGS` constant, it can generate `Sym`s without having to offset them on creation by `COMMON_STRINGS.len()`. This simplifies a lot our logic.

raskad (approved):
Very nice comments explaining how this works and also really good abstractions to make the behavior clear. Really good work :)

boa_engine/src/bytecompiler.rs
```diff
@@ -2551,7 +2554,7 @@ impl<'b> ByteCompiler<'b> {
                 ConstructorKind::Base
             }),
         );
-        code.computed_field_names = Some(gc::GcCell::new(vec![]));
+        code.computed_field_names = Some(boa_gc::Cell::new(vec![]));
```
Razican, line 2557:
```suggestion
        code.computed_field_names = Some(boa_gc::Cell::new(Vec::new()));
```

boa_interner/src/lib.rs
```diff
@@ -69,46 +70,68 @@
...
 }
 
 impl Interner {
-    /// Creates a new `StringInterner` with the given initial capacity.
+    /// Creates a new [`Interner`].
     #[inline]
-    pub fn with_capacity(cap: usize) -> Self {
+    pub fn new() -> Self {
+        Self::default()
+    }
```
Razican, line 113: Do we need a `new()` function if we already have a `default()` function? This is usually frowned upon by clippy, for example.

boa_interner/src/lib.rs
```diff
@@ -69,46 +70,68 @@
...
     /// Returns the number of strings interned by the interner.
     #[inline]
     pub fn len(&self) -> usize {
-        self.inner.len()
+        COMMON_STRINGS.len() + self.spans.len()
     }
 
-    /// Returns `true` if the string interner has no interned strings.
     #[inline]
+    /// Returns `true` if the [`Interner`] contains no interned strings.
```
Razican, line 132:
```suggestion
    /// Returns `true` if the [`Interner`] contains no interned strings.
    #[inline]
```

boa_interner/src/lib.rs
```diff
@@ -180,178 +248,31 @@ impl Interner {
...
-            .find(|&(_i, s)| string == s)
-            .map(|(i, _str)| {
-                let raw = NonZeroUsize::new(i.wrapping_add(1)).expect("static array too big");
-                Sym::from_raw(raw)
+    /// Gets the symbol of the common string if one of them
+    fn get_common(string: &str) -> Option<Sym> {
+        COMMON_STRINGS.get_index(string).map(|idx|
+            // SAFETY: `idx >= 0`, since it's an `usize`,
+            // and `idx + 1 > 0` (considering no overflows, but
+            // an overflow would panic on debug anyways).
```
Razican, line 256: Even if this would panic on debug, we should probably still use [`checked_add()`](https://doc.rust-lang.org/stable/std/primitive.usize.html#method.checked_add) to make sure it's still safe in release builds.

boa_interner/src/lib.rs
```diff
@@ -180,178 +248,31 @@ impl Interner {
...
-        self.value.get() - Self::PADDING
+    /// Generates a new symbol for the provided [`str`] pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `string` points to a valid
+    /// memory inside `head` and that it won't be invalidated
+    /// by allocations and deallocations.
+    unsafe fn generate_symbol(&mut self, string: InternedStr) -> Sym {
+        let next = Sym::new(self.len() + 1).expect("Adding one makes `self.len()` always `> 0`");
```
Razican, line 269: In the `expect()` message, I would mention integer overflow, since it's the only situation when this can happen. Something like this:
```suggestion
        let next = Sym::new(self.len() + 1).expect("cannot generate interner symbol: integer overflow");
```

boa_interner/src/sym.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub(super) fn new(value: usize) -> Option<Self> {
+        NonZeroUsize::new(value).map(|value| Self { value })
+    }
+
+    #[inline]
+    /// Creates a new [`Sym`] from the provided `value`, without checking if `value` is not zero
+    ///
+    /// # Safety
+    ///
+    /// `value` must not be zero.
```
Razican, line 96: Usually, the definition of the function should be together, and the comment above it. Clippy has some lints about this if I recall correctly (we might just not be using them)

```suggestion
    /// Creates a new [`Sym`] from the provided `value`, without checking if `value` is not zero
    ///
    /// # Safety
    ///
    /// `value` must not be zero.
    #[inline]
```

boa_interner/src/sym.rs
```diff
@@ -0,0 +1,143 @@
...
+            value:
+            // SAFETY: The caller must ensure the invariants of the function.
+            unsafe {
+                NonZeroUsize::new_unchecked(value)
+            },
+        }
+    }
+
+    #[inline]
+    /// Returns the internal value of the [`Sym`]
```
Razican, line 108:
```suggestion
    /// Returns the internal value of the [`Sym`]
    #[inline]
```

Razican (commented):
Looks great :) I noticed some things that could be potentially improved. Feel free to take them into account, and we can merge this :)

boa_interner/src/lib.rs
jedel1043: > This is usually frowned upon by clippy.

Not at all! On the contrary, Clippy recommends to have a `Default` implementation when you have a `new()` method.

https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default

boa_interner/src/sym.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// # Safety
+    ///
+    /// `value` must not be zero.
+    pub(super) const unsafe fn new_unchecked(value: usize) -> Self {
+        Self {
+            value:
+            // SAFETY: The caller must ensure the invariants of the function.
+            unsafe {
+                NonZeroUsize::new_unchecked(value)
+            },
```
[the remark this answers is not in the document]
jedel1043: Technically yes, but I enabled the (soon to become an error https://github.com/rust-lang/rust/issues/71668) `unsafe_block_in_unsafe_fn` lint, since the community agrees that adding `unsafe` blocks even inside `unsafe` functions forces us to document the safety of every `unsafe` operation we do.

boa_interner/src/sym.rs
jedel1043: Couldn't find any Clippy lint related to this. (Could be a good contribution to Clippy)

boa_interner/src/lib.rs
jedel1043: Ah, I think I didn't explain well why we don't have to worry about overflows in this case

boa_interner/src/sym.rs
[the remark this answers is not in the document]
Razican: Sounds perfect :)

boa_interner/src/lib.rs
Razican: Oh, indeed, I thought there was another lint that preferred the usage of `default()` over `new()` (I think in the "restriction" or "pedantic" groups), maybe it was removed, or maybe I'm just remembering it wrong xD but it's fine for me :)

boa_interner/src/lib.rs
```diff
@@ -251,13 +261,15 @@ impl Interner {
     /// Gets the symbol of the common string if one of them
     fn get_common(string: &str) -> Option<Sym> {
         COMMON_STRINGS.get_index(string).map(|idx|
-            // SAFETY: `idx >= 0`, since it's an `usize`,
-            // and `idx + 1 > 0` (considering no overflows, but
-            // an overflow would panic on debug anyways).
+            // SAFETY: `idx >= 0`, since it's an `usize`, and `idx + 1 > 0`.
+            // In this case, we don't need to worry about overflows
+            // because `COMMON_STRINGS` would need to be of considerable
+            // size to cause an overflow, even on machines with `usize = u32`.
```
Razican, line 267: Hmmm, I still don't like this approach. To me, it looks like "since we would need many common strings to be unsafe, and we don't have so many, this is safe", but what if this grows in the future? What about 16-bit or 8-bit platforms?

Those probably have other issues of their own, but the only situation where we can ensure that this `unsafe` call is actually safe is if we have really checked the invariants beforehand, and panic if they are not filled, even if this would add some performance impact.

It's similar to the bounds checking in the standard library.

I would suggest to do some compile-time check or an assert on the length of `COMMON_STRINGS` to make sure that the amount of common strings is < `usize::MAX`, and maybe add a comment in `COMMON_STRINGS` itself, to clarify that it cannot have that value.

The thing is that while I guess having a list of length `usize::MAX + 1` would fail to compile, having it the size of `usize::MAX` would actually work, but fail here. But since that list is generated in compile time, and we know the size of `usize::MAX` at compile time, we can compare those then, and not introduce the burden in runtime. Anyways, I know it's a very small edge case, so if it's to complex to generate a `compile_error!()` or the `assert!()` at the beginning of the function doesn't get optimized away (even if `len()` is `const`), I'm OK with leaving it like this.

Razican (commented):
Looking good, still just a comment about safety in edge case scenarios.

boa_interner/src/lib.rs
jedel1043: Had to include the `static_assertions` crate for this, but now we have a compile time check in place that verifies `COMMON_STRINGS.len() < usize::MAX` :)

Razican: Great!! Thank you!!!

Razican (approved):
Looks great now :) it seems that `boa_engine/src/builtins/function/mod.rs` is giving a compile error due to `Sym` not implementing `Trace`. I guess the solution should be similar to `Object`.

jedel1043:
@Razican fixed!

boa_interner/src/sym.rs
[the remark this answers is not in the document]
jasonwilliams: bit late to this, but sounds good

PR #2153, opened 2022-06-30
[Merged by Bors] - Fix `this` in function calls

raskad:

- Fix the `this` value that is passend to function calls to be `undefined` if the function is not a property reference
- Add special handling for setting the `this` value for function calls where the function is a private property reference or a super property reference


HalidOdat (approved):

raskad:
bors +r

PR #2156, opened 2022-07-01
[Merged by Bors] - Fix remaining `Promise` bugs

raskad:

- Fix `Promise.then` length to 2.
- Fix `all`, `allSettled` and `any` element lists to actually be shared between closures.
- Fix `allSettled` reject element function to call resolve instead of reject.

The remaining failing tests for `Promise` fail because of some errors in the arrow function parsing.


Razican (approved):
Great work!!

jedel1043 (approved):
Looks good to me!

PR #2158, opened 2022-07-01
[Merged by Bors] - Implement `async function` and `await`

raskad:

- Implement `AsyncFunction` builtin object.
- Add `Async` as a function object data type.
- Implement `async function` and `await` compilation and execution.
- Parse `await` in more positions.

### DIFF
--- a/boa_engine/src/syntax/parser/expression/unary.rs
+++ b/boa_engine/src/syntax/parser/expression/unary.rs
@@ -121,7 +121,7 @@ where
                 Ok(node::UnaryOp::new(UnaryOp::Not, self.parse(cursor, interner)?).into())
             }
             TokenKind::Keyword((Keyword::Await, true)) if self.allow_await.0 => Err(
-                ParseError::general("Keyword must not contain escaped characters", token_start),
+                ParseError::general("Keyword 'await' must not contain escaped characters", token_start),
             ),
             TokenKind::Keyword((Keyword::Await, false)) if self.allow_await.0 => {
                 Ok((AwaitExpression::new(self.allow_yield).parse(cursor, interner)?).into())
--- a/boa_engine/src/syntax/parser/expression/unary.rs
+++ b/boa_engine/src/syntax/parser/expression/unary.rs
@@ -120,9 +120,12 @@ where
                 cursor.next(interner)?.expect("! token vanished"); // Consume the token.
                 Ok(node::UnaryOp::new(UnaryOp::Not, self.parse(cursor, interner)?).into())
             }
-            TokenKind::Keyword((Keyword::Await, true)) if self.allow_await.0 => Err(
-                ParseError::general("Keyword 'await' must not contain escaped characters", token_start),
-            ),
+            TokenKind::Keyword((Keyword::Await, true)) if self.allow_await.0 => {
+                Err(ParseError::general(
+                    "Keyword 'await' must not contain escaped characters",
+                    token_start,
+                ))
+            }
             TokenKind::Keyword((Keyword::Await, false)) if self.allow_await.0 => {
                 Ok((AwaitExpression::new(self.allow_yield).parse(cursor, interner)?).into())
             }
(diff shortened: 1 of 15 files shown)


boa_engine/src/vm/code_block.rs
```diff
@@ -455,13 +469,32 @@ pub(crate) fn create_function_object(code: Gc<CodeBlock>, context: &mut Context)
...
+        let promise_capability = PromiseCapability::new(
+            &context
+                .intrinsics()
+                .constructors()
+                .promise()
+                .constructor()
+                .into(),
+            context,
+        )
+        .expect("cannot  fail per spec");
```
jedel1043, line 482: Would be good to have a comment with the part of the spec that says that

boa_engine/src/vm/mod.rs
```diff
@@ -2405,6 +2539,15 @@ impl Context {
                         self.vm.push(e);
                     } else {
                         self.vm.stack.truncate(start_stack_size);
+                        if let Some(promise_capability) = promise_capability {
+                            promise_capability
+                                .reject()
+                                .call(&JsValue::undefined(), &[e.clone()], self)
+                                .expect("cannot fail per spec");
```
jedel1043, line 2546: Same as above

boa_engine/src/vm/mod.rs
```diff
@@ -2435,11 +2578,26 @@ impl Context {
         }
 
         if self.vm.stack.len() <= start_stack_size {
+            if let Some(promise_capability) = promise_capability {
+                promise_capability
+                    .resolve()
+                    .call(&JsValue::undefined(), &[], self)
+                    .expect("cannot fail per spec");
```
jedel1043, line 2585: Same as above

boa_engine/src/vm/mod.rs
```diff
@@ -2435,11 +2578,26 @@ impl Context {
...
         }
 
         let result = self.vm.pop();
         self.vm.stack.truncate(start_stack_size);
+
+        if let Some(promise_capability) = promise_capability {
+            promise_capability
+                .resolve()
+                .call(&JsValue::undefined(), &[result.clone()], self)
+                .expect("cannot fail per spec");
```
jedel1043, line 2598: Same as above

jedel1043 (approved):
Very much excited to merge this feature! Just some small nitpicks about the spec that don't block merging

boa_engine/src/syntax/parser/expression/unary.rs
```diff
@@ -120,6 +120,12 @@ where
                 cursor.next(interner)?.expect("! token vanished"); // Consume the token.
                 Ok(node::UnaryOp::new(UnaryOp::Not, self.parse(cursor, interner)?).into())
             }
+            TokenKind::Keyword((Keyword::Await, true)) if self.allow_await.0 => Err(
+                ParseError::general("Keyword must not contain escaped characters", token_start),
```
jedel1043, line 124: I saw that this error is not taken into account for other keywords. Is that exclusive to `await` (meaning it would be good to change the error message) or is that a missing error in other keywords?

raskad: We also have this error for other keywords so you are right, we can make the error more specific here.

jedel1043 (approved):
Looks good!

boa_engine/src/vm/mod.rs
```diff
@@ -2270,6 +2277,106 @@ impl Context {
...
+                )
+                .name("")
+                .length(1)
+                .build();
+
+                // 5. Let rejectedClosure be a new Abstract Closure with parameters (reason) that captures asyncContext and performs the following steps when called:
+                // 6. Let onRejected be CreateBuiltinFunction(rejectedClosure, 1, "", « »).
+                let on_rejected = FunctionBuilder::closure_with_captures(
+                    self,
+                    |_this, args, captures, context| {
```
HalidOdat, line 2333:
```suggestion
                    |_this, args, (environment, value, frame), context| {
```

I think it would be more readable if we use argument destructuring here, instead of accessing tuples through index.

HalidOdat (commented):
Besides the small nitpick it looks good :)

Great work @raskad

boa_engine/src/vm/mod.rs
raskad: Good call. Should be fixed now.

HalidOdat (approved):

PR #2162, opened 2022-07-04
[Merged by Bors] - Safe wrapper for `JsSet`

anuvratsingh:
This PR adds a safe wrapper around JavaScript `JsSet` from `builtins::set`, and is being tracked at #2098.

Implements following methods 
- [x] `Set.prototype.size`
- [x] `Set.prototype.add(value)`
- [x] `Set.prototype.clear()`
- [x] `Set.prototype.delete(value)`
- [x] `Set.prototype.has(value)`
- [x] `Set.prototype.forEach(callbackFn[, thisArg])`
Implement wrapper for `builtins::set_iterator`, to be used by following.
- [x] `Set.prototype.values()`
- [x] `Set.prototype.keys()`
- [x] `Set.prototype.entries()`

*Note: Are there any other functions that should be added?

Also adds `set_create()` and made `get_size()` public in `builtins::set`.

### DIFF
--- a/boa_engine/src/builtins/set/mod.rs
+++ b/boa_engine/src/builtins/set/mod.rs
@@ -174,8 +174,10 @@ impl Set {
             None => context.intrinsics().constructors().set().prototype(),
         };
 
-        let set = JsObject::from_proto_and_data(prototype, ObjectData::set(OrderedSet::new()));
-        Ok(set)
+        Ok(JsObject::from_proto_and_data(
+            prototype,
+            ObjectData::set(OrderedSet::new()),
+        ))
     }
 
     /// `get Set [ @@species ]`
--- a/boa_engine/src/builtins/set/mod.rs
+++ b/boa_engine/src/builtins/set/mod.rs
@@ -164,7 +164,8 @@ impl Set {
         Ok(set.into())
     }
 
-    /// Utility for constructing `Set` Objects.
+    /// Utility for constructing `Set` objects.
+    #[allow(clippy::unnecessary_wraps)]
     pub(crate) fn set_create(
         prototype: Option<JsObject>,
         context: &mut Context,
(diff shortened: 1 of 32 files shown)


jedel1043:
Ok, with #2115 we now have a simple API to handle JS iterators. Maybe it would be good if you checked out the PR and the corresponding implementation of `JsMapIterator`

boa_engine/src/builtins/set/mod.rs
```diff
@@ -164,7 +164,8 @@ impl Set {
         Ok(set.into())
     }
 
-    /// Utility for constructing `Set` Objects.
+    /// Utility for constructing `Set` objects.
+    #[allow(clippy::unnecessary_wraps)]
```
jedel1043, line 168: What is the reason for allowing this lint?
I think it would be better to return only a `JsObject` as the lint recommends

jedel1043: It's good to maintain consistency when it doesn't affect the usability of our program. However, this change impacts the usability, because now you need to handle an error on a function that cannot error.

anuvratsingh: Alright, I understand.

boa_engine/src/object/jsobject.rs
```diff
@@ -325,7 +325,7 @@ impl JsObject {
     }
 
     /// Checks if it's a `MapIterator` object
-    ///
+    ///
```
Razican, line 328: Why are some spaces added here?

boa_engine/src/builtins/set/mod.rs
```diff
@@ -164,6 +164,32 @@ impl Set {
         Ok(set.into())
     }
 
+    /// Utility for constructing `Set` objects.
+    pub(crate) fn set_create(prototype: Option<JsObject>, context: &mut Context) -> JsObject {
+        let prototype = match prototype {
+            Some(prototype) => prototype,
+            None => context.intrinsics().constructors().set().prototype(),
+        };
```
Razican, line 172: I would use `unwrap_or_else()` here, to avoid boilerplate.

boa_engine/src/object/jsobject.rs
anuvratsingh: My bad, they were left out while resolving rebase conflicts.

boa_engine/src/builtins/set/mod.rs
```diff
@@ -323,6 +349,9 @@ impl Set {
 
         let callback_arg = &args[0];
         let this_arg = args.get_or_undefined(1);
+
+        // dbg!(this_arg); // REMOVE DEBUG
```
Razican, line 353: This should probably be removed, right?

boa_engine/src/builtins/set/mod.rs
```diff
@@ -337,6 +366,7 @@ impl Set {
                 .as_object()
                 .and_then(|obj| {
                     obj.borrow().as_set_ref().map(|set| {
+                        println!("SET: {:?}", set); // REMOVE DEBUG
```
Razican, line 369: This should probably be removed, right?

Razican (changes requested):
Some leftover debug prints, but looks good!

Razican (approved):
Looks great to me now :)

raskad (approved):
Nice work. Thanks for the additional typo fixes :)

Issue #2157, opened 2022-07-01
Clean Public Call/Frame Stack API


dansteren:
I want a clean, public API to access the stack frames. Although some of the information is there, it is currently difficult/impossible to get. At the very least, the visibility needs to be changed on the existing functions/structs/etc so that they are public. But ideally there would be a nice dedicated and consistent API for working with the call stack.

**Current Behavior**
To access the current call frame you can currently do `&my_boa_context.vm.frame`, which isn't too bad. But then to get the name of the frame I have to do `&my_boa_context.interner.resolve_expect(frame.code.name.clone())`. Additionally, if I want to traverse the frame stack I have to iterate along the frames calling `current_frame.prev`. It's not very intuitive.

**Suggested Behavior**
It would be nice if there was an array/vec of frames that I could get the total size of and point to any index to easily move between frames in the stack. Specifically, it should be able to get the top-most call frame very easily. Additionally, being able to access the name of the function in an easier fashion would be ideal.

At @demergent-labs we are doing some very low level things to get [azle](https://github.com/demergent-labs/azle) working and having a clean API would be very helpful for many reasons.

Potentially Related Issues:
- https://github.com/boa-dev/boa/issues/809
- https://github.com/boa-dev/boa/issues/1711

PR #2164, opened 2022-07-04
[Merged by Bors] - Store call frames in `Vec` instead of singly-linked list

HalidOdat:
This storage method should be more cache friendly since we store in contiguous memory, besides that it should make #2157 a bit easier. Will work on that next after this gets merged :)

It changes the following:
- Remove the unneeded `prev` field in `CallFrame`
- Preallocate some space for future calls (16 slots)


raskad (approved):
Looks good. Makes working with frames much more straight forward.

Razican (approved):
Great changes!

raskad:
@HalidOdat I think you accidentally changed to an outdated version of the 262 submodule in the last push here right?

Issue #2058, opened 2022-05-03
More efficient conversion of Vec to JsArray


lastmjs:
I just want to bring this up and get some preliminary feedback. My project [Azle](https://github.com/demergent-labs/azle) is a TS/JS environment for the Internet Computer, a Wasm environment. It is relatively resource-constrained compared with other cloud solutions at the moment, and all computation is metered with a unit called cycles. Each RPC call into a canister has a cycle limit.

Users will need to deal with large arrays of `Vec<u8>` to deal with raw binary data. Right now it seems like the conversion between `Vec<u8>` and `JsArray` (to get the user's raw binary data into the canister in the RPC call) is extremely inefficient, insomuch that the cycles limit on the computation is reached and the call fails. I am reaching the limit on arrays of length between 10,000 and 25,000 (~10-20kb).

What can be done to make the conversion more efficient? And btw I am not 100% it's Boa's conversion that is causing the problem, I think my code can be optimized to some extent. But I fear the BigO of the JsArray code (like `from_iter`) will just be insufficient. What can be done if this is the case? Is there a possibility of just sharing a memory location with a boa context so that we do not need to iterate through entire arrays to convert them over?

Razican:
The thing is that a JS Array is an array of values of any type, which means that every value in the Rust array must be converted to a JS value. I don't think there is much room for improvement here (maybe pre-allocation, which we might be already doing)

But, the ideal JS type for a `Vec<u8>` is a TypedArray, which uses a Rust Vec backend to be stored. So, I would say that the next step should be to implement easy conversion from Rust Vec to TypedArray.

lastmjs:
Awesome, I will experiment with that next, thanks!

lastmjs:
I'm ready to tackle this, can you show me some code for what is currently possible converting a `Vec<u8>` into a `JsUint8Array` and vice versa? If it isn't currently possible to do this without iterating over every item, I can work on a pull request to do it efficiently.

Razican:
Most of this work would be done here: https://github.com/boa-dev/boa/blob/main/boa_engine/src/object/jstypedarray.rs

The idea is to add another `From` implementation in the `JsTypedArrayType!` macro (which should probably be renamed for Rust standard naming convention, in snake_case).

This `From` would receive a `Vec<$element>`, and create a $name object. For this, it will need to first create an `ArrayBuffer` like this:

```rust
let my_vector: Vec<u8> = vec![];
[...]

PR #2167, opened 2022-07-06
[Merged by Bors] - Dense/Packed JavaScript arrays

HalidOdat:
This PR implements an optimization done by V8 and spidermonkey. Which stores indexed properties in two class storage methods dense and sparse. The dense method stores the property in a contiguous array ( `Vec<T>` ) where the index is the property key. This storage method is the default. While on the other hand we have sparse storage method which stores them in a hash map with key `u32` (like we currently do). This storage method is a backup and is slower to access because we have to do a hash map lookup, instead of an array access.

In the dense array we can store only data descriptors that have a value field and are `writable`, `configurable` and `enumerable` (which by default array elements are). Since all the fields of the property descriptor are the same except value field, we can omit them an just store the `JsValue`s in `Vec<JsValue>` this decreases the memory consumption and because it is smaller we are less likely to have cache misses.

There are cases where we have to convert from dense to sparse (the slow case):
- If we insert index elements in a non-incremental way, like `array[1000] = 1000` (inserting an element to an index that is already occupied only replaces it does not make it sparse)
- If we delete from the middle of the array (making a hole), like `delete array[10]`  (only converts if there is actualy an element there, so calling delete on a non-existent index property will do nothing)

Once it becomes sparse is *stays* sparse there is no way to convert it again. (the computation needed to check whether it can be converted outweighs the benefits of this optimization)

I did some local benchmarks and on array creation/pop and push there is ~45% speed up and the impact _should_ be bigger the more elements we have. For example the code below on `main` takes `~21.5s` while with this optimization is `~3.5s` (both on release build).

```js
let array = [];
for (let i = 0; i < 50000; ++i) {
   array[i] = i;
}
```

In addition I also made `Array::create_array_from_list` do a direct setting of the properties (small deviation from spec but it should have the same behaviour), with this #2058 should be fixed, conversion from `Vec` to `JsArray`, not `JsTypedArray` for that I will work on next :)

### DIFF
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -266,15 +266,23 @@ impl Array {
         // 2. Let array be ! ArrayCreate(0).
         let array = Self::array_create(0, None, context)
             .expect("creating an empty array with the default prototype must not fail");
+
         // 3. Let n be 0.
         // 4. For each element e of elements, do
-        for (i, elem) in elements.into_iter().enumerate() {
-            // a. Perform ! CreateDataPropertyOrThrow(array, ! ToString(𝔽(n)), e).
-            array
-                .create_data_property_or_throw(i, elem, context)
-                .expect("new array must be extensible");
-            // b. Set n to n + 1.
-        }
+        //     a. Perform ! CreateDataPropertyOrThrow(array, ! ToString(𝔽(n)), e).
+        //     b. Set n to n + 1.
+        //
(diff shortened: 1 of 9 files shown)


jedel1043 (changes requested):
Great optimization! I just have two small comments about some parts of the code

jedel1043 (approved):
Looks good to me!

raskad (approved):
Really nice change. Great work!

Issue #2148, opened 2022-06-28
The expression `/=/` does not get parsed as a regexp


lxhom:
**Describe the bug**
The expression `/=/` does not get parsed as a regexp

**To Reproduce**

```javascript
let myRegex = /=/;
```
That throws this: "unexpected token '/=', primary expression at line 1, col 15"

**Expected behavior**
It should give me the regular expression for a single `=`. A workaround I used is just using /.{0}=/, but this should work without it.

**Build environment (please complete the following information):**

(tested it on my machine and on the playground)

- OS: Windows & Boa Playground
- Version: 0.15.0 & the playground doesn't tell me
- Target triple: x86_64-pc-windows-msvc & idk
- Rustc version: rustc 1.63.0-nightly (7466d5492 2022-06-08)

Razican:
For anyone willing to tackle this, the `/` token lexing happens here: https://github.com/boa-dev/boa/blob/main/boa_engine/src/syntax/lexer/mod.rs#L128-L187

This checks if the `/` token is the beginning of a comment (when it's followed by `/` or `*`), and if not, it will check the current lexer goal symbol. This is set by the parser, depending on what we're trying to parse.

There are some places where a `/` will only mean a division element, and others when it can only be a regular expression start. In this case, it seems it goes through the first branch (`InputElement::Div | InputElement::TemplateTail`), since it will check if the next character is an equal sign (which it is), and then say that we are getting a `Punctuator::AssignDiv` symbol.

The parser then is surprised, since the assign punctuator already appeared, and therefore, you cannot use the `/=` symbol.

It seems, that the issue here is that the parser is not properly setting the goal symbol. Once the `let myRegex =` has been consumed, the goal symbol should not accept the division operation, only the regular expression.

You should be able to follow the documentation in our [debugging section](https://github.com/boa-dev/boa/blob/main/docs/debugging.md#tokens-and-ast-nodes) to track the issue down, but feel free to ask questions on this.

PR #2169, opened 2022-07-06
Create a structure-aware fuzzer (second try)

VTCAKAVSMoACE:
This is a retry of #1902. Depends on #1954.

This PR adds ~~two~~ three experimental fuzzers which generate valid JavaScript code from Arbitrary structs. These fuzzers (or variants thereof) were used to identify all of my previous PRs and issues. It does not generate identifiers which resolve to built-in types.

I will link issues discovered by this fuzzer to this PR as they are discovered. As with last time, if you wish to test using this fuzzer, I would recommend using:

```
cargo fuzz run -s none interp_fuzzer -- -timeout=5
```


VTCAKAVSMoACE:
Note that (as part of this PR) we should add build step check to CI for fuzz with `cargo fuzz build`. Not entirely sure how to do that, though.

VTCAKAVSMoACE:
Added another fuzzer which checks for parser/interner idempotency. This should help find issues like #2148 and other tricky parser/interner bugs.

VTCAKAVSMoACE:
Closing in favour of new techniques.

Issue #1371, opened 2021-06-29
Split `Node` into `Expression`, `Statement` and `Declaration`


HalidOdat:
Currently `Node` is both a expression, statement and declaration. It would be better if `Node` was split into how the spec describes it `Expression`, `Statement` and also `Declaration`. Why do we want this? Right now we don't differentiate between statements (like `if`, `while`, etc) and expressions (like `2 + 2`), while we could make it work like this, it makes somethings awkward like `if (cond) {}`, `cond` should only an expression, and when we parse it we check for this, but some checks can be elided if we leverage the type system, making illegal state unrepresentable.

We want to match to what [astexplorer](https://astexplorer.net/#/gist/ecf73e65e156aa22ecf8f5d3c1d67613/584573b039b672ba14605eb3208bb7f0057499c7) outputs.

jtmorrisbytes:
Is anyone working on this?

Razican:
> Is anyone working on this?

I don't think so, do you want me to assign it to you?

pdogr:
If this is still available, can I work on this?

jedel1043:
> If this is still available, can I work on this?

We haven't assigned someone, so please do! Let me assign it to you

jedel1043:
I'll work on this in the next days :)

PR #2171, opened 2022-07-09
Implement arrow function parsing based on `CoverParenthesizedExpressionAndArrowParameterList`

raskad:
Previously we parsed arrow functions without the relevant cover grammar `CoverParenthesizedExpressionAndArrowParameterList`. This leads to either arrow functions or parenthesized expressions not being parsed correctly. Implementing this is a bit tricky, as the cover grammar is being parsed in `PrimaryExpression` while arrow functions are parsed in `AssignmentExpression`. This means that we have to return the covered parameter list that was parsed via `CoverParenthesizedExpressionAndArrowParameterList` in `PrimaryExpression` to `AssignmentExpression`. Fortunately this works pretty good and now the full arrow function test suite, with the exception of a few tests that require other features, passes.

- Implement `CoverParenthesizedExpressionAndArrowParameterList` parsing.
- Implement `CoverInitializedName` parsing in object literals.
- Fix a bug where an environment would be wrongly removed from the environment stack when an expression in default function parameters throws.
- Add more valid cases where on object literal can be converted to an object declaration pattern.
- Implement `Expression` parsing manually to avoid some cases where the parser would prematurely throw an error.
- Implement parsing of arrow functions via `CoverParenthesizedExpressionAndArrowParameterList`.
- Remove unneeded `AllowIn` flag on array and object declaration pattern parsers.
- Fix an of-by-one bug in the trace output.

### DIFF
--- a/boa_engine/src/syntax/ast/node/mod.rs
+++ b/boa_engine/src/syntax/ast/node/mod.rs
@@ -249,6 +249,12 @@ pub enum Node {
 
     /// A call of the super constructor. [More information](./super_call/struct.SuperCall.html).
     SuperCall(SuperCall),
+
+    /// A FormalParameterList.
+    ///
+    /// This is only used in the parser itself.
+    /// It is not a valid AST node.
+    FormalParameterList(FormalParameterList),
 }
 
 impl From<Const> for Node {
@@ -358,6 +364,7 @@ impl Node {
             Self::ClassDecl(ref decl) => decl.to_indented_string(interner, indentation),
             Self::ClassExpr(ref expr) => expr.to_indented_string(interner, indentation),
             Self::SuperCall(ref super_call) => super_call.to_interned_string(interner),
+            Self::FormalParameterList(_) => unreachable!(),
         }
     }
 
@@ -740,6 +747,7 @@ impl Node {
                                 }
                             }
                         },
+                        PropertyDefinition::CoverInitializedName(_, _) => {}
                     }
                 }
             }
@@ -1189,7 +1197,6 @@ impl Node {
             Node::Object(object) => {
                 for property in object.properties() {
                     match property {
-                        PropertyDefinition::IdentifierReference(_) => {}
                         PropertyDefinition::Property(name, init) => {
                             if let Some(node) = name.computed() {
                                 if node.contains(symbol) {
@@ -1212,6 +1219,8 @@ impl Node {
                                 }
                             }
                         }
+                        PropertyDefinition::IdentifierReference(_)
+                        | PropertyDefinition::CoverInitializedName(_, _) => {}
                     }
                 }
             }
@@ -1237,6 +1246,7 @@ impl Node {
                     }
                 }
             }
+            Node::Yield(_) if symbol == ContainsSymbol::YieldExpression => return true,
             _ => {}
         }
         false
@@ -1248,6 +1258,7 @@ impl Node {
 pub(crate) enum ContainsSymbol {
     SuperProperty,
     SuperCall,
+    YieldExpression,
 }
 
 impl ToInternedString for Node {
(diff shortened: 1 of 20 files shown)


boa_engine/src/syntax/ast/node/mod.rs
```diff
@@ -249,6 +249,12 @@ pub enum Node {
 
     /// A call of the super constructor. [More information](./super_call/struct.SuperCall.html).
     SuperCall(SuperCall),
+
+    /// A FormalParameterList.
+    ///
+    /// This is only used in the parser itself.
+    /// It is not a valid AST node.
+    FormalParameterList(FormalParameterList),
```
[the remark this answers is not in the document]
Razican: BTW, why does this need to be a `Node` if it's not used as such?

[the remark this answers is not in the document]
raskad: The problem is that we need to return it from the `PrimaryExpression` parser to the `AssignmentExpression` parser. There are almost all expression parsers in that path. If it where not in `Node` we would have an additional enum and would have to check that everytime we work with any return from any expression parser.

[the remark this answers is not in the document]
Razican: Hmmm I see, there is no perfect solution to that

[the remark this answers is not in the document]
raskad: I think it is probably worth revisiting this if we split `Node` up (see #1371). At that point there may be an efficent method of handling this in a better way.

boa_engine/src/syntax/parser/expression/primary/mod.rs
```diff
@@ -243,3 +256,304 @@ where
...
+            !cursor.peek_expect_is_line_terminator(0, interner)?
+        } else {
+            false
+        };
+
+        // If the next token is not an arrow, we know that we must parse a parenthesized expression.
+        if !is_arrow {
+            if let Some(span) = tailing_comma {
+                return Err(ParseError::unexpected(
+                    ",".to_string(),
```
Razican, line 378: Can't we pass the `Punctuator` variant in these?

boa_engine/src/syntax/parser/expression/primary/object_initializer/mod.rs
```diff
@@ -153,6 +153,25 @@ where
             }
         }
 
+        //  CoverInitializedName[?Yield, ?Await]
```
Razican, line 156: I would say, if this is a no terminal, that it should have its own parser. Would make things more modular and maintainable.

Razican (approved):
Parsing arrow functions was probably the messiest part of the parser. This looks so much better!

Thank you for fixing this :) I gave it a look from my phone and it looks pretty good, just had a few comments, but looks good from my side!

boa_engine/src/syntax/parser/expression/primary/mod.rs
```diff
@@ -243,3 +256,304 @@ where
...
+            !cursor.peek_expect_is_line_terminator(0, interner)?
+        } else {
+            false
+        };
+
+        // If the next token is not an arrow, we know that we must parse a parenthesized expression.
+        if !is_arrow {
+            if let Some(span) = tailing_comma {
+                return Err(ParseError::unexpected(
+                    Punctuator::Comma.to_string(),
```
Razican, line 378: Is the `to_string()` needed?

raskad: It was, because we take `Into<Box<str>>` in `ParseError::unexpected`. I implemented `From<Punctuator> for Box<str>` and now it works without.

Razican: Thanks! I didn't know that!

HalidOdat (approved):
Looks good to me! :)

PR #2174, opened 2022-07-11
[Merged by Bors] - Implement Generator Function Constructor

raskad:

- Modify `CreateDynamicFunction` to work with generator functions.
- Add the name `anonymus` to functions created via `CreateDynamicFunction` to comply with the spec. 
- Fix a bug in the `Yield` parser where the parser would expect a token when no token is a legal case.
- Change the `Yield::new` function to require less turbofishes.

### DIFF
--- a/boa_engine/src/builtins/function/mod.rs
+++ b/boa_engine/src/builtins/function/mod.rs
@@ -460,7 +460,7 @@ impl BuiltInFunctionObject {
         args: &[JsValue],
         context: &mut Context,
     ) -> JsResult<JsValue> {
-        Self::create_dynamic_function(new_target, args, false, context).map(Into::into)
+        Self::create_dynamic_function(new_target, args, false, false, context).map(Into::into)
     }
 
     /// `CreateDynamicFunction ( constructor, newTarget, kind, args )`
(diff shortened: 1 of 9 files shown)


boa_engine/src/builtins/function/mod.rs
```diff
@@ -473,40 +473,56 @@ impl BuiltInFunctionObject {
...
-                    for arg in args {
-                        parameters.push(arg.to_string(context)?);
+            let parameters = if args.is_empty() {
+                FormalParameterList::empty()
+            } else {
+                let mut parameters = Vec::with_capacity(args.len());
+                for arg in args {
+                    parameters.push(arg.to_string(context)?);
+                }
+                let mut parameters = parameters.join(",");
```
Razican, line 496: It seems a bit too much work here, right? Why do we need to re-convert the parameters to a string?

Especially because I see that we later call the parser. Would it be possible to have a shorthand function in the parser that doesn't need a raw `(...)` and can just work with the vector?

boa_engine/src/syntax/ast/node/yield/mod.rs
```diff
@@ -29,13 +29,12 @@ impl Yield {
     }
 
     /// Creates a `Yield` AST node.
-    pub fn new<E, OE>(expr: OE, delegate: bool) -> Self
+    pub fn new<E>(expr: Option<E>, delegate: bool) -> Self
```
Razican, line 32: The reason behind this was to allow any `E` to be used, since `T` implements `Into<Option<T>>`, so we could use the type directly without wrapping it in a `Some`. On the other hand, this increases compile time, and the benefit might not even be apparent, since for `None` you might need to turbo fish.

I guess that's the reasoning behind removing it, right?

Razican (approved):
Looking good, just a couple of questions, for clarifications mostly :) thanks!

boa_engine/src/builtins/function/mod.rs
raskad: We for sure need to convert every argument to a string individually, because this is of course valid: 
```javascript
const arg = Object("a")
typeof arg // 'object'
const f = Function(arg, "console.log(a)")
f(1) // 1
```

We probably could have a special parser for this, but tbh I don't think it's that imporant. I think (hope) that this is not used that much in production code. I will open an issue to possibly implement this if it does not cause too much code duplication in the parser.

boa_engine/src/syntax/ast/node/yield/mod.rs
raskad: Yeah, also if the turbofish would only be needed for `None` that would be fine, but for some reason we also need it in some cases for `Some`. For example here: https://github.com/boa-dev/boa/blob/56be7cec523c091e6cc49419542f985cbf344106/boa_engine/src/syntax/parser/expression/assignment/yield.rs#L72-L75

That double type specification seems really unnecessary to me, especially because the function is only used in so few places.

boa_engine/src/builtins/function/mod.rs
raskad: See https://github.com/boa-dev/boa/issues/2180

PR #2181, opened 2022-07-16
Safe wrapper for `JsDate`

anuvratsingh:
This PR adds a safe wrapper around JavaScript `JsDate` from `builtins::date`, and is being tracked at #2098.

#### Implements following methods 
- [x] `new Date()`
- [ ] `Date.prototype.getDate()`
- [ ] `Date.prototype.getDay()`
- [ ] `Date.prototype.getFullYear()`
- [ ] `Date.prototype.getHours()`
- [ ] `Date.prototype.getMilliseconds()`
- [ ] `Date.prototype.getMinutes()`
- [ ] `Date.prototype.getMonth()`
- [ ] `Date.prototype.getSeconds()`
- [ ] `Date.prototype.getTime()`
- [ ] `Date.prototype.getTimezoneOffset()`
- [ ] `Date.prototype.getUTCDate()`
- [ ] `Date.prototype.getUTCDay()`
- [ ] `Date.prototype.getUTCFullYear()`
- [ ] `Date.prototype.getUTCHours()`
- [ ] `Date.prototype.getUTCMilliseconds()`
- [ ] `Date.prototype.getUTCMinutes()`
- [ ] `Date.prototype.getUTCMonth()`
- [ ] `Date.prototype.getUTCSeconds()`
- [ ] `Date.prototype.getYear()`
- [x] `Date.now()`
- [ ] `Date.parse()`
- [x] `Date.prototype.setDate()`
- [x] `Date.prototype.setFullYear()`
- [ ] `Date.prototype.setHours()` ?
- [x] `Date.prototype.setMilliseconds()`
- [ ] `Date.prototype.setMinutes()` ?
- [x] `Date.prototype.setMonth()`
- [x] `Date.prototype.setSeconds()`
- [x] `Date.prototype.setTime()`
- [x] `Date.prototype.setUTCDate()`
- [x] `Date.prototype.setUTCFullYear()`
- [x] `Date.prototype.setUTCHours()`
- [x] `Date.prototype.setUTCMilliseconds()`
- [x] `Date.prototype.setUTCMinutes()`
- [x] `Date.prototype.setUTCMonth()`
- [x] `Date.prototype.setUTCSeconds()`
- [x] `Date.prototype.setYear()`
- [x] `Date.prototype.toDateString()`
- [ ] `Date.prototype.toGMTString()`
- [x] `Date.prototype.toISOString()`
- [x] `Date.prototype.toJSON()`
- [ ] `Date.prototype.toLocaleDateString()`
- [ ] `Date.prototype.toLocaleString()`
- [ ] `Date.prototype.toLocaleTimeString()`
- [ ] `Date.prototype.toString()`
- [x] `Date.prototype.toTimeString()`
- [ ] `Date.prototype.toUTCString()`
- [x] `Date.UTC()`
- [ ] `Date.prototype.valueOf()`

### Issues
1. `get_*()` and some other methods - They take `&self` as input internally, therefore these would require input to be `this: &JsValue, args: &[JsValue], context: &mut Context` like others.
2. `to_string()`- how can I use `Date::to_string()` rather than `alloc::string::ToString`.
3. `set_hours()` and `set_minutes()` - they subtract local timezones when setting the value, e.g.
```rust
// reference date = 2000-01-01T06:26:53.984
date.set_hours(&[23.into(), 23.into(), 23.into(), 23.into()], context)?;
// would add tiemzone(+5:30) to it
// Is                 2000-01-01T17:53:23.023
// Should be    2000-01-01T23:23:23.023
```
4. `parse()` - it uses `chrono::parse_from_rfc3339` internally, while es6 spec recommends ISO8601. And it can also parse other formats like from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) `04 Dec 1995 00:12:00 GMT` which fails. So what should be done with it.

### Possible improvements 
Right now, `object::jsdate::set_full_year()` and alike are like below, `into()` doesn't feel ergonomic.
```rust
#[inline]
pub fn set_full_year(&self, values: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
    Date::set_full_year(&self.inner.clone().into(), values, context)
}
// Usage 
date.set_full_year(&[2000.into(), 0.into(), 1.into()], context)?;
// How can something like this be made to work
#[inline]
#[inline]
    pub fn set_full_year<T>(&self, values: &[T], context: &mut Context) -> JsResult<JsValue>
    where
        T: Into<JsValue>,
    {
                                                          | expected reference `&[value::JsValue]` 
                                                          | found reference `&[T]`        
        Date::set_full_year(&self.inner.clone().into(), values, context)
    }
```


boa_engine/src/builtins/date/mod.rs
```diff
@@ -547,8 +540,19 @@ impl Date {
...
-        self.to_local().map_or(f64::NAN, |dt| f64::from(dt.day()))
+    pub fn get_date(this: &JsValue, _args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        if let Some(t) = this_time_value(this, context)?.0 {
+            let date = Local::now()
+                .timezone()
+                .from_utc_datetime(&t)
+                .format("%Y-%m-%dT%H:%M:%S%.f%:z")
+                .to_string();
+            let local = NaiveDateTime::parse_from_str(&date, "%Y-%m-%dT%H:%M:%S%.f%:z")
+                .expect("INFALLIBLE: Converting string built from `Local::now()` to NaiveDateTime");
```
raskad, line 551: Why do we have to do the conversion from a `DateTime<Local>` to a `String` and then to a `NaiveDateTime`?
Should this not yield the same result:
```rust
Local::now()
    .timezone()
    .from_utc_datetime(&t)
    .day()
```

At least on my setup the `test/built-ins/Date/prototype/getDate/` testsuite passes and I'm not on UTC.

Same for all the other non-utc get methods.

boa_engine/src/builtins/date/mod.rs
```diff
@@ -687,7 +788,16 @@ impl Date {
...
+    pub fn get_time(this: &JsValue, _args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        if let Some(t) = this_time_value(this, context)?.0 {
+            Ok(JsValue::new(t.timestamp_millis()))
+        } else {
+            Ok(JsValue::nan())
+        }
+    }
+
+    /// Utility: Internal `get_time()`
+    fn int_get_time(&self) -> f64 {
```
raskad, line 800: Looking at the call-sites of this method, it should probably be named TimeClip according to the spec (https://tc39.es/ecma262/#sec-timeclip).

boa_engine/src/builtins/date/mod.rs
```diff
@@ -1649,6 +1817,25 @@ impl Date {
...
+    ///
+    /// The `toLocaleDateString()` method returns the date portion of the given Date instance
+    /// according to language-specific conventions.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-date.prototype.tolocaledatestring
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
```
raskad, line 1830: This link seems wrong, same with all the other toLocale* mdn links

boa_engine/src/builtins/date/mod.rs
```diff
@@ -1649,6 +1817,25 @@ impl Date {
...
+    pub fn to_locale_date_string(
+        _this: &JsValue,
+        _args: &[JsValue],
+        _context: &mut Context,
+    ) -> JsResult<JsValue> {
+        unimplemented!()
```
raskad, line 1836: With this, right now if you execute `(new Date).toLocaleDateString()` boa panics. We should not add unimplemented methods that just panic.

raskad (changes requested):
I took a first look at the changes in the builts themselves and left some suggestions.

The issues that you mention all sound out of scope of this PR to me. I think we should convert all of them to issues so they can get fixed. For the purpose of this PR we just want to expose the builtins, so we can use the existing funtionalities even if they are not 100% spec compliant.

boa_engine/src/builtins/date/mod.rs
```diff
@@ -137,23 +120,26 @@ impl BuiltIn for Date {
         .method(Self::set_utc_minutes, "setUTCMinutes", 3)
         .method(Self::set_utc_month, "setUTCMonth", 2)
         .method(Self::set_utc_seconds, "setUTCSeconds", 2)
+        .method(Self::set_year, "setYear", 1)
         .method(Self::to_date_string, "toDateString", 0)
-        .method(getter_method!(to_gmt_string), "toGMTString", 0)
+        .method(Self::to_gmt_string, "toGMTString", 0)
         .method(Self::to_iso_string, "toISOString", 0)
+        // Why toJSON length is 1, does `method()`'s length signifies number of
+        // inputs? If so shouldn't it be 0?
```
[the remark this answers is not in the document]
anuvratsingh: Alright, I should also check if other functions are listed correctly.

boa_engine/src/builtins/date/mod.rs
anuvratsingh: Ohh this was unfinished part and was in there because I asked for suggestion. As suggested, `locale*` are outside the scope, so I will remove it.

boa_engine/src/builtins/date/mod.rs
anuvratsingh: Sorry, but what exactly is wrong with them? They work, link the correct thing and uses the same format as other functions.

Also, as `locale*` are out of this PR's scope, should I remove the docs or keep them?

boa_engine/src/builtins/date/mod.rs
anuvratsingh: I should have read `chrono` docs properly, as I was trying to call `day()` after `format()`. So I used a workaround.

boa_engine/src/builtins/date/mod.rs
anuvratsingh: `time_clip` already exits
```rust
#[inline]
pub fn time_clip(time: f64) -> Option<f64> {
    if time.abs() > 8.64e15 {
        None
    } else {
        Some(time)
    }
 }
```
`time_clip(f64)` and `int_get_time(self)` signatures are different and are being used at different places like [here](https://github.com/boa-dev/boa/blob/099154cb34c4f7fec6cf8d45f1f96b6797e10dd4/boa_engine/src/builtins/date/mod.rs#L321).

Let say I use `time_clip()` in `set_minutes()`, then I have to get milliseconds which is provided by `int_get_time()`.
So, Step 6 of `set_minutes` would look like this: `let u = Self::time_clip(t.int_get_time());`. And convert `Option<f64>` to `Result<JsValue, JsValue>`.

Looking at the [spec](https://tc39.es/ecma262/#sec-date.prototype.setminutes) its `LocalTime(t)`. 
On further look, steps doesn't match the spec for almost all the functions, is this intentional.

Thanks to this, I figured out the issue with `set_hours/minutes/etc.` step 6 expects `LocalTime` but `int_get_time` provides `UTC`.

boa_engine/src/builtins/date/mod.rs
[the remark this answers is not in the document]
raskad: By the way I see that you have changed the legth for the toLocale* functions to 2. Not really releveant, because we want to remove them, but just for your info; length 0 would be correct for those. The spec specifies their siganures like this: [Date.prototype.toLocaleDateString ( [ reserved1 [ , reserved2 ] ] )](https://tc39.es/ecma262/#sec-date.prototype.tolocaledatestring). Notice that the wrapping the arguments in square brackets means that they are optional. Optional arguments do not count towards the length of the function.

boa_engine/src/builtins/date/mod.rs
raskad: Alright, I think we should probably fix/refactor this. But then let's do that in another PR and keep the existing behaviour for this one.

boa_engine/src/builtins/date/mod.rs
raskad: True, I missread that. I answered the question that below.

boa_engine/src/builtins/date/mod.rs
raskad: > As suggested, `locale*` are outside the scope, so I should remove them, right?

Yes, let's either remove the functions or not add them to the object in `impl BuiltIn` for now. I think not having the function present on the object is the best way to not have the output interpreted false. Also it may be confusing in the conformance page when a few tests for the function pass but there is actually nothing implemented.

boa_engine/src/object/jsdate.rs
```diff
@@ -0,0 +1,564 @@
...
+use crate::{
+    builtins::Date,
+    object::{JsObject, JsObjectType},
+    Context, JsResult, JsValue,
+};
+
+#[derive(Debug, Clone, Trace, Finalize)]
+pub struct JsDate {
+    inner: JsObject,
+}
```
Razican, line 14: I would add some documentation to the structure itself, maybe with a small example. We should probably also add some documentation at the module level, even if it's just a one-liner.

Razican (approved):
This looks great to me :) I just recommend a bit of documentation in the `JsDate` structure itself, but it can be merged. Thank you!!

boa_engine/src/object/jsdate.rs
anuvratsingh: Thanks for the review, I have added a doc-example for `JsDate` but didn't rebase it as the code structure has changed from what I remember.

Razican: Hmm it shouldn't have changed a lot. The main difference might be that creating errors now uses `JsNativeError::...()` instead of `context.construct...()`. But checking at the merge conflicts, it seems it's just some import ordering and a different derive for `Date`.

Razican: @jedel1043 can you help here?

jedel1043: It's just the signature of `main`. It should return `JsResult<()>` instead of `Result<(), JsValue>`.

anuvratsingh: @jedel1043 thank you so much, I was really confused trying to figure out the issue.

raskad (approved):
Thanks for the contribution! Sorry for the long wait on the review.

raskad:
bors seems to have some problems with the github check status so I rebased the branch to hopefully fix that.

PR #2196, opened 2022-07-21
[Merged by Bors] - Fix length/index in `32bit` architectures

### LINKED ISSUE #2182: Panic compile to wasm32-unknown-unknown target
**Describe the bug**

Compile boa_engine to wasm32, panic when create empty array.

file: `boa_engine/src/buildins/array/mod.rs`
```rust
    pub(crate) fn array_create(
        length: usize,
        prototype: Option<JsObject>,
        context: &mut Context,
    ) -> JsResult<JsObject> {
        // 1. If length > 2^32 - 1, throw a RangeError exception.
        if length > 2usize.pow(32) - 1 {
            return context.throw_range_error("array exceeded max size");
        }
```

Change this line `if length > ` to:
```rust
if length > u32::MAX as usize {
```

**To Reproduce**
Compile boa to was32 or other 32bit targets.

Create a empty array cause panic, like:
```javascript
let a = [];
```

**Expected behavior**

`2 ^ 32` panic under 32bit arch

- OS: MacBook Pro
- Version: [e.g. 32]
- Target triple: wasm32-unknown-unknown
- Rustc version: 1.60.0 (7737e0b5c 2022-04-04)

**Additional context**
n/a

HalidOdat:
This PR fixes the length/index types from `usize` to `u64`, because in JavaScript the length can be from `0` to `2^53 - 1` it cannot be represented in a `usize`. On `64-bit` architectures this is fine because it is already a unsigned 64bit number and these changes essentially do nothing. But on 32bit architectures this was causing the length to be truncated giving an unexpected result. 

fixes/closes #2182 

It changes the following:
- Make length a `u64` to be able to represent all possible length values
- Change `JsValue::to_length()` t return `u64`
- Change `JsValue::to_index()` t return `u64`

### DIFF
--- a/boa_engine/src/builtins/array/array_iterator.rs
+++ b/boa_engine/src/builtins/array/array_iterator.rs
@@ -17,7 +17,7 @@ use boa_profiler::Profiler;
 #[derive(Debug, Clone, Finalize, Trace)]
 pub struct ArrayIterator {
     array: JsObject,
-    next_index: usize,
+    next_index: u64,
     kind: PropertyNameKind,
     done: bool,
 }
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -192,7 +192,7 @@ impl Array {
             debug_assert!(number_of_args >= 2);
 
             // b. Let array be ? ArrayCreate(numberOfArgs, proto).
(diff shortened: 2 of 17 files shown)


Razican (approved):
Looks good! Thanks!

raskad (approved):

PR #2216, opened 2022-08-03
[Merged by Bors] - Fix spread arguments in function calls

raskad:
Currently we only spread spread-expressions if they are the last argument in the function call. With this fix all arguments are spread if needed. The downside is that an array object is allocated to store all arguments if the arguments contain a spread-expression. But with dense indexed properties inplemented in #2167 this should be reasonably fast.


Razican (approved):
Looks good to me :)

HalidOdat (approved):

PR #2220, opened 2022-08-05
[Merged by Bors] - Parse class private async generator methods

raskad:
Currently async generator class methods are not being parsed correctly. In comparison to the current main branch this only fixes a few tests, but after #2200 is merged, it should be aroud 1000 additional passing tests.


Razican (approved):
Looks good! Thanks for the fix :) I will try to review the other PRs as soon as possible in the next days :)

HalidOdat (approved):
Looks perfect to me! :)

PR #2231, opened 2022-08-12
[Merged by Bors] - Fix `arguments` object iterator function

raskad:

- Create the `%Array.prototype.values%` function object once and use it in both the `Array` prototype and as the `@@iterator` value of `arguments` objects.

This fixes a test where `arguments[Symbol.iterator]` and `[][Symbol.iterator]` would not be equal.


Razican (approved):
Looks good to me :) thanks!

HalidOdat (approved):

PR #2232, opened 2022-08-13
[Merged by Bors] - Fix `yield` expression to end on line terminator

raskad:

- Fix `yield` expression to end on line terminator

### DIFF
--- a/boa_engine/src/syntax/parser/cursor/mod.rs
+++ b/boa_engine/src/syntax/parser/cursor/mod.rs
@@ -263,15 +263,15 @@ where
 
     /// Check if the peeked token is a line terminator.
     #[inline]
-    pub(super) fn peek_expect_is_line_terminator(
+    pub(super) fn peek_is_line_terminator(
         &mut self,
         skip_n: usize,
         interner: &mut Interner,
-    ) -> Result<bool, ParseError> {
+    ) -> Result<Option<bool>, ParseError> {
         if let Some(t) = self.buffered_lexer.peek(skip_n, false, interner)? {
-            Ok(t.kind() == &TokenKind::LineTerminator)
+            Ok(Some(t.kind() == &TokenKind::LineTerminator))
         } else {
-            Err(ParseError::AbruptEnd)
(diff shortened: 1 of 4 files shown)


boa_engine/src/syntax/parser/cursor/mod.rs
```diff
@@ -263,15 +263,15 @@ where
 
     /// Check if the peeked token is a line terminator.
     #[inline]
-    pub(super) fn peek_expect_is_line_terminator(
+    pub(super) fn peek_is_line_terminator(
         &mut self,
         skip_n: usize,
         interner: &mut Interner,
-    ) -> Result<bool, ParseError> {
+    ) -> Result<Option<bool>, ParseError> {
```
Razican, line 270: Since this no longer returns an error, we can remove the `Result`, right?

boa_engine/src/syntax/parser/expression/assignment/yield.rs
```diff
@@ -63,12 +63,14 @@ where
             interner,
         )?;
 
-        let token = if let Some(token) = cursor.peek(0, interner)? {
-            token
-        } else {
-            return Ok(Node::Yield(Yield::new::<Node>(None, false)));
-        };
+        match cursor.peek_is_line_terminator(0, interner)? {
```
Razican, line 66: I would use `matches!()` here to avoid an empty branch.

Razican (commented):
Looks good :) just a couple of comments that could improve the code.

boa_engine/src/syntax/parser/cursor/mod.rs
raskad: We still call `BufferedLexer::peek` and that may return a `ParseError` other than `AbruptEnd`.

Razican: True, didn't see it!

Razican (approved):

HalidOdat (approved):
Looks good to me! :)

PR #2233, opened 2022-08-13
[Merged by Bors] - Removed some unsafe_empty_trace!() calls to improve performance

### LINKED ISSUE #1615: Cleanup invalid `derive(Trace)`
Currently we have some type implementations that derive the `Trace` trait unnecessarily, such as
https://github.com/boa-dev/boa/blob/master/boa/src/syntax/ast/constant.rs

This is far from ideal, because we're not marking them as dead ends for the garbage collector to ignore them, and instead the garbage collector has to visit every field, even if the fields are also dead ends. This adds redirection and impacts on the performance of the GC.

We should reevaluate every type implemented on our codebase with a more conservative approach: deriving `Trace` and `Finalize` only when strictly necessary (e.g. structs that need to be `Trace` because they transitively have a `Gc` inside them or structs that are used inside multiple types that need to be `Trace`) and instead try to use `unsafe_ignore_trace` and `empty_trace` a lot more.

Razican:
This Pull Request fixes #1615.

It changes the following:

- Removes the `Trace` implementation from types that don't need it (except for `JsSymbol` and `JsString`, which are needed elsewere).
- Uses `#[unsafe_ignore_trace]` in places where we need to implement `Trace` for part of a structure.
- Implements a custom `Trace` in enums where deriving it is not possible, since `#[unsafe_ignore_trace]` doesn't work for enums.

### DIFF
--- a/boa_engine/src/builtins/array/array_iterator.rs
+++ b/boa_engine/src/builtins/array/array_iterator.rs
@@ -17,8 +17,11 @@ use boa_profiler::Profiler;
 #[derive(Debug, Clone, Finalize, Trace)]
 pub struct ArrayIterator {
     array: JsObject,
+    #[unsafe_ignore_trace]
     next_index: u64,
+    #[unsafe_ignore_trace]
     kind: PropertyNameKind,
+    #[unsafe_ignore_trace]
     done: bool,
 }
 
--- a/boa_engine/src/builtins/array/array_iterator.rs
+++ b/boa_engine/src/builtins/array/array_iterator.rs
@@ -17,11 +17,9 @@ use boa_profiler::Profiler;
 #[derive(Debug, Clone, Finalize, Trace)]
 pub struct ArrayIterator {
     array: JsObject,
-    #[unsafe_ignore_trace]
     next_index: u64,
     #[unsafe_ignore_trace]
     kind: PropertyNameKind,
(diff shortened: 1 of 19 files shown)


boa_engine/src/builtins/array/array_iterator.rs
```diff
@@ -17,8 +17,11 @@ use boa_profiler::Profiler;
 #[derive(Debug, Clone, Finalize, Trace)]
 pub struct ArrayIterator {
     array: JsObject,
+    #[unsafe_ignore_trace]
     next_index: u64,
+    #[unsafe_ignore_trace]
     kind: PropertyNameKind,
+    #[unsafe_ignore_trace]
```
raskad, line 24: Do we need the `#[unsafe_ignore_trace]` on `u64` and `bool`? Those should already be `unsafe_empty_trace` right?

raskad (approved):
Nice work, great to have this cleaned up.

boa_engine/src/builtins/array/array_iterator.rs
Razican: Yes, I will remove this, it will probably get optimized out by the compiler anyways.

HalidOdat (approved):

Razican:
Bors retry

PR #2234, opened 2022-08-14
[Merged by Bors] - Implement Async-from-Sync Iterator Objects

raskad:

- Implement [Async-from-Sync Iterator Objects](https://tc39.es/ecma262/#sec-async-from-sync-iterator-objects)
- Give the proper `async` hint to `GetIterator` when executing a delegate yield expression in an async generator function

### DIFF
--- a/boa_engine/src/builtins/iterable/async_from_sync_iterator.rs
+++ b/boa_engine/src/builtins/iterable/async_from_sync_iterator.rs
@@ -1,0 +1,435 @@
+use crate::{
+    builtins::{
+        iterable::{create_iter_result_object, IteratorRecord, IteratorResult},
+        promise::{if_abrupt_reject_promise, PromiseCapability},
+        JsArgs, Promise,
+    },
+    object::{FunctionBuilder, JsObject, ObjectData},
+    property::PropertyDescriptor,
+    Context, JsResult, JsValue,
+};
+use boa_gc::{Finalize, Trace};
+use boa_profiler::Profiler;
+
+/// Create the `%AsyncFromSyncIteratorPrototype%` object.
+///
+/// More information:
+///  - [ECMA reference][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-%asyncfromsynciteratorprototype%-object
+pub(crate) fn create_async_from_sync_iterator_prototype(context: &mut Context) -> JsObject {
+    let _timer = Profiler::global().start_event("AsyncFromSyncIteratorPrototype", "init");
+
+    let prototype = JsObject::from_proto_and_data(
+        context
+            .intrinsics()
+            .objects()
+            .iterator_prototypes()
+            .async_iterator_prototype(),
+        ObjectData::ordinary(),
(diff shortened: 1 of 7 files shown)


boa_engine/src/builtins/iterable/async_from_sync_iterator.rs
```diff
@@ -0,0 +1,435 @@
...
+        let r#return = sync_iterator.get_method("return", context);
+
+        // 6. IfAbruptRejectPromise(return, promiseCapability).
+        if_abrupt_reject_promise!(r#return, promise_capability, context);
+
+        let result = match (r#return, args.get(0)) {
+            // 7. If return is undefined, then
+            (None, _) => {
+                // a. Let iterResult be CreateIterResultObject(value, true).
+                let iterator_result =
```
Razican, line 209: This might better adhere to the spec if it would be called `iter_result`, but I think this name is also pretty clear.

boa_engine/src/builtins/iterable/async_from_sync_iterator.rs
```diff
@@ -0,0 +1,435 @@
...
+    /// More information:
+    ///  - [ECMA reference][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-asyncfromsynciteratorcontinuation
+    fn continuation(
+        result: &IteratorResult,
+        promise_capability: &PromiseCapability,
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        // 1. NOTE: Because promiseCapability is derived from the intrinsic %Promise%, the calls to promiseCapability.[[Reject]] entailed by the use IfAbruptRejectPromise below are guaranteed not to throw.
```
Razican, line 374: I would try to make these comments less than 100 characters wide, as per Rust formatting guidelines.

Razican (approved):
Looks very good! Sorry for not reviewing this earlier.
I only found a couple of potential changes that might make sense: a name that was changed from the spec, and some comments that are probably too wide in a single line (Rust formatting conventions prefer lines with less than 100 characters).

But I would be happy to merge it like this too, great work!!!

boa_engine/src/builtins/iterable/async_from_sync_iterator.rs
```diff
@@ -0,0 +1,440 @@
...
+
+        // 5. IfAbruptRejectPromise(value, promiseCapability).
+        if_abrupt_reject_promise!(value, promise_capability, context);
+
+        // 6. Let valueWrapper be Completion(PromiseResolve(%Promise%, value)).
+        let value_wrapper = Promise::promise_resolve(
+            context.intrinsics().constructors().promise().constructor(),
+            value,
+            context,
+        );
```
jedel1043, line 395: Would be good to do a complete survey of the strictness of our return types, because apparently `promise_resolve` only returns `JsObjects` but has a return type of `JsValue`, and that would save us an `expect`.

jedel1043 (approved):
Nice work!

PR #2245, opened 2022-08-20
[Merged by Bors] - check history file exist if not create it

### LINKED ISSUE #2223: boa-cli crashes if history file does not exist
**Describe the bug**
`boa_cli` crashes when the `.boa_history` does not exist. I think this was introduced in #1809 with the rustyline update.

**To Reproduce**
In the project directory you can change `.boa_history`'s name to `.boa_history.bak` and try to run `cargo run`, it will exit with "No such file or directory".

**Expected behavior**
The code below should be modified to create the history file if it doesn't exist yet. Not sure what is the best way to go about this but I expect rustyline to include this in its API.
https://github.com/boa-dev/boa/blob/3d6a8d5232edfb75382de5ef4be4bd895f5bee17/boa_cli/src/main.rs#L218-L221

udhaykumarbala:
This Pull Request fixes/closes #2223.

It changes the following:

- Check the .boa_history file, if it is not exist then create the file before load history function to prevent the crash

### DIFF
--- a/boa_cli/src/main.rs
+++ b/boa_cli/src/main.rs
@@ -63,7 +63,7 @@ use boa_engine::{syntax::ast::node::StatementList, Context};
 use clap::{ArgEnum, Parser};
 use colored::{Color, Colorize};
 use rustyline::{config::Config, error::ReadlineError, EditMode, Editor};
-use std::{fs::read, io, path::PathBuf};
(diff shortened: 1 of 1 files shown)


Razican (approved):
Looks good! Thanks!

raskad (approved):

PR #2246, opened 2022-08-21
[Merged by Bors] - Removed some non-needed dependencies

### LINKED ISSUE #2244: RUSTSEC-2021-0139: ansi_term is Unmaintained
> ansi_term is Unmaintained

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `ansi_term`                      |
| Version             | `0.12.1`                   |
| URL                 | [https://github.com/ogham/rust-ansi-term/issues/72](https://github.com/ogham/rust-ansi-term/issues/72) |
| Date                | 2021-08-18                         |

The maintainer has adviced this crate is deprecated and will not
receive any maintenance.

The crate does not seem to have much dependencies and may or may not be ok to use as-is.

Last release seems to have been three years ago.

## Possible Alternative(s)

 The below list has not been vetted in any way and may or may not contain alternatives;

 - [anstyle](https://github.com/epage/anstyle)
 - [nu-ansi-term](https://crates.io/crates/nu-ansi-term)
 - [yansi](https://crates.io/crates/yansi)

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0139.html) for additional details.

Razican:
This Pull Request removes two dependencies that were not really needed, and fixes #2244 by no longer having the package in the dependency tree.

It changes the following:

- The `structopt` dependency in `boa_tester` has been replaced by `clap` v3, the same way as we did in `boa_cli`. This means that we have one less dependency (at least), and that `clap` v2 is only used as a dev-dependency by `criterion` (which will probably be removed in 0.4, as per https://github.com/bheisler/criterion.rs/issues/596).
- The no-longer-updated `num-format` dependency has been removed from `boa_tester`. We were only using it to add comma thousands separator on results, so I added a simple function to do it (not very performant, but it will only be used a few times when showing results).

Looking at this, I noticed a couple of things:

 - The `csv` dependency, used by `criterion` has not been updated in more than a year, and it's using a very old `itoa` dependency. They updated the dependency in the repository in March, but unfortunately, the release is taking some more time than expected, and a tracking issue can be found here: https://github.com/BurntSushi/rust-csv/issues/271
 - `cargo update` fails, because the latest update to `tinystr` in the ICU4x breaks ICU4x 0.6. I have reported this here: https://github.com/unicode-org/icu4x/issues/2428 and their recommendation is for us to use a beta version of the library, but I don't think we should go for that, since this is a semver breakage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,28 +21,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anyhow"
 version = "1.0.62"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1485d4d2cc45e7b201ee3767015c96faa5904387c9d87c6efdd0fb511f12d305"
-
-[[package]]
-name = "arrayvec"
-version = "0.4.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cd9fd44efafa8690358b7408d253adf110036b88f55672a933f01d616ad9b1b9"
-dependencies = [
- "nodrop",
-]
 
 [[package]]
 name = "atty"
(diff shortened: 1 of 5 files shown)


raskad (approved):
Nice finds!

HalidOdat (approved):
Looks perfect to me! :)

Issue #1987, opened 2022-03-29
Uncaught TypeError when calling method on global object


lupd:
Calling a method of the global object (using `this`/ `globalThis`) results in a `TypeError`.

**To Reproduce**
This JavaScript code reproduces the issue:
```javascript
this.hasOwnProperty("length")
```

**Expected behavior**
`true` is printed

**Actual behavior**
`Uncaught "TypeError": "not a callable function"` is printed.

**Additional context**
At least one [test](https://github.com/tc39/test262/blob/main/test/built-ins/parseFloat/S15.1.2.3_A7.5.js) in the test suite requires the expected behavior above.

Razican:
It turns out that we have two bugs related to the global object:
 - When we initialize builtins, they are added to the `Global` environment record as bindings, not as properties of the global object, which means that `this` and `globalThis` are empty objects (`{}`).
 - The global object has no prototype. This is required by the spec, and as far as I can tell, it should probably be the `Object` prototype, since `hasOwnProperty()` and others are part of the `Object` prototype.

The first issue is relatively easy to fix: we can just insert properties in the global object when we initialize builtins. In theory, we should not create bindings for these, and the global environment should check if a variable is part of the global object or one of the global bindings, but removing the insertion into the global bindings breaks around 50k tests.

The second issue is trickier, since when the global object gets constructed, the `Object` built-in doesn't exist, so we cannot easily assign its prototype to it. I will publish a PR that solves the first issue, but the second one will need some extra thought.

raskad:
> * When we initialize builtins, they are added to the `Global` environment record as bindings, not as properties of the global object, which means that `this` and `globalThis` are empty objects (`{}`).
> * The global object has no prototype. This is required by the spec, and as far as I can tell, it should probably be the `Object` prototype, since `hasOwnProperty()` and others are part of the `Object` prototype.

The way the global object currently works is, that it uses special internal object methods defiend in `GLOBAL_INTERNAL_METHODS` in `boa_engine/src/object/internal_methods/global.rs`. That way it can use the global environment bindings as properties. The implementation currently completeley ignores the objects prototype. The best example is `global_has_property` where only the global property map is checked while the spec suggests that the object's prototype should be checked if no property is found on the object itself.

[...]

PR #2251, opened 2022-08-22
Fixing global object not being populated with the global object properties

Razican:
This Pull Request fixes part of #1987.

It changes the following:

- When a built-in gets initialized, not only the global binding gets updated, but the global object receives that built-in as a property.

This is a fairly inefficient way of doing this, since it clones the property and assigns it to the global object. Ideally, if I'm not mistaken, these properties shouldn't be declared in the global bindings, they should just be properties of the global binding, so we should avoid creating the global binding. Nevertheless, this, as I mentioned in #1987, this breaks a ton of tests.

The issue here is that the naming resolution at global level doesn't check the global object, unfortunately. Should this be changed?

Also, this doesn't solve the full issue, since the main problem of #1987 comes from the fact that the global object doesn't have any prototype, but as per the spec:

> has a `[[Prototype]]` internal slot whose value is [host-defined](https://tc39.es/ecma262/#host-defined).

[Full spec](https://tc39.es/ecma262/#sec-global-object)

I could not find anywhere that the prototype should be the `Object` prototype, though. This would require some further changes to be fixed.

### DIFF
--- a/boa_engine/src/builtins/mod.rs
+++ b/boa_engine/src/builtins/mod.rs
@@ -130,7 +130,11 @@ fn init_builtin<B: BuiltIn>(context: &mut Context) {
             .build();
         context
             .global_bindings_mut()
-            .insert(B::NAME.into(), property);
+            .insert(B::NAME.into(), property.clone());
+        context
+            .global_object()
+            .borrow_mut()
+            .insert(B::NAME, property);
     }
 }
 


raskad:
I this we should solve this is a different way. I put a full explanation in the linked issue: https://github.com/boa-dev/boa/issues/1987#issuecomment-1223388212

Razican:
I agree we should solve this properly :) closing this PR in favor of that approach, feel free to open a PR about it.

PR #1659, opened 2021-10-11
[Merged by Bors] - First prototype for new `JsString` using UTF-16 

jedel1043:
I think it's time to address the elephant in the room.

This Pull Request will (hopefully!) solve part of #736.

This is a complete rewrite of `JsString`, but instead of storing `u8` bytes it stores `u16` words. The `encode!` macro (renamed to `utf16!` for simplicity) from the `const-utf16` crate allows us to create UTF-16 encoded arrays at compilation time. `JsString` implements `Deref<Target=[u16]>` to unlock the slice methods and possibly make some manipulations easier. However, we would need to create our own library of utilities for `JsString`.

PR #2267, opened 2022-09-05
[Merged by Bors] - Add URI encoding and decoding functions

### LINKED ISSUE #894: Implement URI Handling Functions

Razican:
This Pull Request closes #894.

It changes the following:

- Adds the `encodeURI()`, `decodeURI()`, `encodeURIComponent()` and `decodeURIComponent()` functions
- Passes all the tests except for those depending on #1987 or on the comment below.

Things to discuss:
 - I'm unable to find in the spec information regarding the only failing tests, which relate to [this](https://github.com/tc39/test262/blob/f1870753fa616261193b99f79d996889921b3404/test/built-ins/encodeURI/S15.1.3.3_A1.1_T2.js):
   > If string.charAt(k) in [0xDC00 - 0xDFFF], throw URIError

Let me know your thoughts :)


boa_engine/src/builtins/uri/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+        // e. Set R to the string-concatenation of R and S.
+        r.extend_from_slice(&s);
+
+        // f. Set k to k + 1.
+        k += 1;
+    }
+}
+
+/// Checks if a given code unit is an hexadecimal digit represented in UTF-16.
+fn is_hexdigit(code_unit: u16) -> bool {
```
raskad, line 442: Instead of returning a bool here, this could return `Option<u8>` since we check if it is hex (in u8 range). Then the resulting `u8`s could be passed to `decode_byte` and that function would be safer, because it would not contain the `u8::try_from`.

Or even better, compine `is_hexdigit` and `decode_byte`. They are only called directly after each other and if we combine them we can avoid the double check and all potential panics.

boa_engine/src/builtins/uri/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+                    context.throw_uri_error("TO-DO")?;
+                }
+
+                // 2. If k + (3 × (n - 1)) ≥ strLen, throw a URIError exception.
+                if k + (3 * (n - 1)) > str_len {
+                    context.throw_uri_error("TO-DO")?;
+                }
+
+                // 3. Let Octets be « B ».
+                let mut octets = vec![b];
```
raskad, line 381: Same as above

boa_engine/src/builtins/uri/mod.rs
```diff
@@ -0,0 +1,553 @@
...
+            } else {
+                // viii. Else,
+                // 1. If n = 1 or n > 4, throw a URIError exception.
+                if n == 1 || n > 4 {
+                    context.throw_uri_error("TO-DO")?;
+                }
+
+                // 2. If k + (3 × (n - 1)) ≥ strLen, throw a URIError exception.
+                if k + (3 * (n - 1)) > str_len {
+                    context.throw_uri_error("TO-DO")?;
```
raskad, line 377: Some leftover todo error messages.

raskad (commented):
I left some ideas and nitpicks.
The `DC00—DFFF` range seem to be [surrogates](https://unicode-table.com/en/blocks/low-surrogates/), but I'm not sure where exactly we miss something here.
I think switching to ranges in the reserved sets is probably a good idea, but not that imporant. We could always do that later.

Really nice to get these functions done, great work :)

jedel1043:
The failing tests are probably related to the abstract operation [`Encode`](https://tc39.es/ecma262/#sec-encode):

```
i. Let cp be CodePointAt(string, k).
ii. If cp.[[IsUnpairedSurrogate]] is true, throw a URIError exception.
```

It was probably changed between spec editions, but the intent seems to be the same: rejecting URIs with unpaired surrogates.

Having said that, fixing that test is probably blocked by #1659

boa_engine/src/builtins/uri/mod.rs
Razican: I merged both functions, let me know what you think :)

Razican:
> The failing tests are probably related to the abstract operation [`Encode`](https://tc39.es/ecma262/#sec-encode):
> 
> ```
> i. Let cp be CodePointAt(string, k).
> ii. If cp.[[IsUnpairedSurrogate]] is true, throw a URIError exception.
> ```
> 
> It was probably changed between spec editions, but the intent seems to be the same: rejecting URIs with unpaired surrogates.
> 
> Having said that, fixing that test is probably blocked by #1659

I'm checking this, the issue might be with the `CodePointAt` implementation.

Razican:
> The failing tests are probably related to the abstract operation [`Encode`](https://tc39.es/ecma262/#sec-encode):
> 
> ```
> i. Let cp be CodePointAt(string, k).
> ii. If cp.[[IsUnpairedSurrogate]] is true, throw a URIError exception.
> ```
> 
> It was probably changed between spec editions, but the intent seems to be the same: rejecting URIs with unpaired surrogates.
> 
> Having said that, fixing that test is probably blocked by #1659

It turns out that the issue is in the `String.fromCharCode(...codeUnits)` function. Here, we convert the code units list to a `std::String`, which, for these cases, since the used characters are unpaired surrogates, it converts them to the UTF replacement character, and this then makes the rest fail.

I guess that not using Rust string as the backend storage would solve this (and many other issues). So we should try to implement the full UTF-16 support.

boa_engine/src/builtins/uri/mod.rs
raskad: I was thinking of doing it like this:

```rust
fn decode_hex_byte(high: u16, low: u16, context: &mut Context) -> JsResult<u8> {
    match (u8::try_from(high), u8::try_from(low)) {
        (Ok(high), Ok(low)) => {
            let high = if (b'0'..=b'9').contains(&high) {
                high - b'0'
            } else if (b'A'..=b'F').contains(&high) {
                high - b'A' + 0x0A
            } else if (b'a'..=b'f').contains(&high) {
                high - b'a' + 0x0a
            } else {
                return Err(context.construct_uri_error("invalid hexadecimal digit found"));
            };

            let low = if (b'0'..=b'9').contains(&low) {
                low - b'0'
            } else if (b'A'..=b'F').contains(&low) {
                low - b'A' + 0x0A
            } else if (b'a'..=b'f').contains(&low) {
                low - b'a' + 0x0a
            } else {
                return Err(context.construct_uri_error("invalid hexadecimal digit found"));
            };

            Ok((high << 4) + low)
        }
        _ => Err(context.construct_uri_error("invalid hexadecimal digit found")),
    }
}
```

That should avoid any double check and there is no panic anymore.

Razican: Done :)

raskad (approved):
Great work. Nice addition of the documentation to the codepoints functions!

boa_engine/src/builtins/uri/mod.rs
```diff
@@ -0,0 +1,564 @@
...
+            } else if (b'a'..=b'f').contains(&low) {
+                low - b'a' + 0x0a
+            } else {
+                return Err(context.construct_uri_error("invalid hexadecimal digit found"));
+            };
+
+            Ok((high << 4) + low)
+        }
+        _ => Err(context.construct_uri_error("invalid hexadecimal digit found")),
+    }
```
jedel1043, line 464: This would simplify a lot if we casted `high` and `low` to `u32`, then tried to parse a `char` for each of them. That would allow us to call `char::to_digit` directly.

jedel1043 (changes requested):
The implementation looks good!

jasonwilliams (approved):

boa_engine/src/builtins/uri/consts.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// More information:
+///  - [`uriReserved` in ECMAScript spec][uri_reserved]
+///  - [`uriUnescaped` in ECMAScript spec][uri_unescaped]
+///
+/// [uri_reserved]: https://tc39.es/ecma262/#prod-uriReserved
+/// [uri_unescaped]: https://tc39.es/ecma262/#prod-uriUnescaped
+pub(super) fn is_uri_reserved_or_uri_unescaped_or_number_sign(code_point: u16) -> bool {
+    code_point == NUMBER_SIGN || is_uri_unescaped(code_point) || URI_RESERVED.contains(&code_point)
+}
```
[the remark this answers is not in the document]
raskad: I think this way it's perfect. The raw ranges and codepoints are encapsuled in the one file and only accessible via the functions that have very descriptive names.

boa_engine/src/builtins/uri/mod.rs
raskad: I pushed a commit changing this to use `char::to_digit`. The generated asm looks better too as far as I can tell.

boa_engine/src/builtins/uri/consts.rs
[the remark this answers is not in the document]
jedel1043: Hm, in that case I'd add `#[inline]` to both of them, just in case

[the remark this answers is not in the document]
raskad: Done!

jedel1043 (approved):
Looks good!

PR #2177, opened 2022-07-14
Remove incorrect `cursor.set_goal()` call in `AssignmentExpression` parsing

jbarthelmess:
This Pull Request fixes/closes #2148 .

It changes the following:

- Adds a test to the lexer showing that the lexar produces correct tokens
- Removes `cursor.set_goal()` call from the AssignmentExpression parse expression

PR #2268, opened 2022-09-07
[Merged by Bors] - Fixed assignment expression parsing

### LINKED ISSUE #2148: The expression `/=/` does not get parsed as a regexp

Razican:
This Pull Request fixes/closes #2148.

It changes the following:

- When we start an assignment expression, the `/` token must be a regular expression literal. The division can only occur between expressions after the assignment operator.
- Added tests for the new behaviour, taken from #2177

This overrides #2177

### DIFF
--- a/boa_engine/src/syntax/lexer/tests.rs
+++ b/boa_engine/src/syntax/lexer/tests.rs
@@ -635,6 +635,25 @@ fn regex_literal() {
 }
 
 #[test]
+fn regex_equals_following_assignment() {
+    let mut lexer = Lexer::new(&b"const myRegex = /=/;"[..]);
+    let mut interner = Interner::default();
+
+    let expected = [
+        TokenKind::Keyword((Keyword::Const, false)),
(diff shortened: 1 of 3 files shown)


raskad (approved):
Looks good

jedel1043 (approved):
LGTM!

PR #2269, opened 2022-09-08
[Merged by Bors] - Allow some keywords as identifiers

### LINKED ISSUE #2275: Use "async" as a function identifier.
**Describe the bug**

`async` is a keyword, but also allowed as an identifier.
https://tc39.es/ecma262/#sec-keywords-and-reserved-words
> Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where Identifier is not allowed: as, async, from, get, meta, of, set, and target.

But, currently, it raises an error when calling a function named `async`.

**To Reproduce**
Run the following code:
```js
async();
```
Uncaught "SyntaxError": "expected token 'function', got '(' in async function declaration at line 1, col 6"

**Expected behavior**
It should be a reference error, or running ok if "async" function is defined.

Note that `async` should be recognized as a keyword in the following case.
```js
let fn = async () => { };
```

**Additional context**
Mentioned in https://github.com/boa-dev/boa/pull/2269

tunz:
This PR fixes #2275 

There are keywords that are allowed as identifiers.
https://tc39.es/ecma262/#sec-keywords-and-reserved-words
> Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where Identifier is not allowed: as, async, from, get, meta, of, set, and target.

This PR adds test cases for them, and fixes some cases such as
`class A { set(a, b) { } }`
`function of() { }`
`let obj = {async: true}`
`async()`

### DIFF
--- a/boa_engine/src/syntax/parser/cursor/mod.rs
+++ b/boa_engine/src/syntax/parser/cursor/mod.rs
@@ -84,6 +84,30 @@ where
     }
 
     #[inline]
+    pub(super) fn peek_or_abrupt(
+        &mut self,
+        skip_n: usize,
+        interner: &mut Interner,
+    ) -> Result<&Token, ParseError> {
+        self.buffered_lexer
+            .peek(skip_n, true, interner)?
+            .ok_or(ParseError::AbruptEnd)
+    }
+
+    #[inline]
+    pub(super) fn peek_kind(
+        &mut self,
+        skip_n: usize,
+        interner: &mut Interner,
+    ) -> Result<&TokenKind, ParseError> {
+        Ok(self
+            .buffered_lexer
+            .peek(skip_n, true, interner)?
+            .ok_or(ParseError::AbruptEnd)?
+            .kind())
+    }
+
+    #[inline]
     pub(super) fn strict_mode(&self) -> bool {
         self.buffered_lexer.strict_mode()
     }
--- a/boa_engine/src/syntax/parser/cursor/mod.rs
+++ b/boa_engine/src/syntax/parser/cursor/mod.rs
@@ -84,30 +84,6 @@ where
     }
 
     #[inline]
-    pub(super) fn peek_or_abrupt(
-        &mut self,
-        skip_n: usize,
-        interner: &mut Interner,
-    ) -> Result<&Token, ParseError> {
-        self.buffered_lexer
-            .peek(skip_n, true, interner)?
-            .ok_or(ParseError::AbruptEnd)
-    }
-
-    #[inline]
-    pub(super) fn peek_kind(
-        &mut self,
-        skip_n: usize,
-        interner: &mut Interner,
-    ) -> Result<&TokenKind, ParseError> {
-        Ok(self
-            .buffered_lexer
-            .peek(skip_n, true, interner)?
-            .ok_or(ParseError::AbruptEnd)?
-            .kind())
-    }
-
-    #[inline]
     pub(super) fn strict_mode(&self) -> bool {
         self.buffered_lexer.strict_mode()
     }
--- a/boa_engine/src/syntax/parser/expression/primary/mod.rs
+++ b/boa_engine/src/syntax/parser/expression/primary/mod.rs
@@ -100,7 +100,7 @@ where
         let tok = cursor.peek(0, interner)?.ok_or(ParseError::AbruptEnd)?;
 
         match tok.kind() {
-            TokenKind::Keyword((Keyword::This | Keyword::Async, true)) => Err(ParseError::general(
+            TokenKind::Keyword((Keyword::This, true)) => Err(ParseError::general(
                 "Keyword must not contain escaped characters",
                 tok.span().start(),
             )),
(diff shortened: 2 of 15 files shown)


Razican:
  
  #### VM implementation
  | Passed | 64,890 | 64,902 | **+12** |
  | Failed | 10,263 | 10,251 | **-12** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 70.74% | 70.75% | **+0.01%** |
  
  <details><summary><b>Fixed tests (14):</b></summary>
  
  ```
  test/built-ins/TypedArray/of/invoked-as-func.js [strict mode] (previously Failed)
  test/built-ins/TypedArray/of/invoked-as-func.js (previously Failed)
  test/built-ins/TypedArrayConstructors/of/invoked-as-func.js [strict mode] (previously Failed)
  test/built-ins/TypedArrayConstructors/of/invoked-as-func.js (previously Failed)
  test/built-ins/TypedArrayConstructors/of/BigInt/invoked-as-func.js [strict mode] (previously Failed)
  test/built-ins/TypedArrayConstructors/of/BigInt/invoked-as-func.js (previously Failed)
  test/language/statements/for-of/head-lhs-async-escaped.js [strict mode] (previously Failed)
  test/language/statements/for-of/head-lhs-async-escaped.js (previously Failed)
  test/language/statements/for-of/head-lhs-async-parens.js [strict mode] (previously Failed)
  test/language/statements/for-of/head-lhs-async-parens.js (previously Failed)
  test/language/expressions/division/no-magic-asi.js [strict mode] (previously Failed)
  test/language/expressions/division/no-magic-asi.js (previously Failed)
  test/language/expressions/division/no-magic-asi-from-block-eval.js [strict mode] (previously Failed)
  test/language/expressions/division/no-magic-asi-from-block-eval.js (previously Failed)
  ```
  </details>
  
  <details><summary><b>Broken tests (2):</b></summary>
  
  ```
  test/language/statements/for-of/head-lhs-async-invalid.js [strict mode] (previously Passed)
  test/language/statements/for-of/head-lhs-async-invalid.js (previously Passed)
  ```
  </details>

boa_engine/src/syntax/parser/cursor/mod.rs
```diff
@@ -83,6 +83,30 @@ where
...
+        skip_n: usize,
+        interner: &mut Interner,
+    ) -> Result<&TokenKind, ParseError> {
+        Ok(self
+            .buffered_lexer
+            .peek(skip_n, true, interner)?
+            .ok_or(ParseError::AbruptEnd)?
+            .kind())
+    }
+
```
Razican, line 109: I see that the `peek_or_abrupt()` function is only being used once, and the `peek_kind()` function only twice. Do we really need them? It might be easy to use the existing functions. What do you think?

boa_engine/src/syntax/parser/expression/primary/mod.rs
```diff
@@ -126,17 +126,42 @@ where
...
+                let contain_escaped_char = *contain_escaped_char;
+                let is_generator = if let Ok(kind) = cursor.peek_kind(2, interner) {
+                    kind == &TokenKind::Punctuator(Punctuator::Mul)
                 } else {
-                    AsyncFunctionExpression::new(self.name, self.allow_yield)
+                    false
+                };
+
+                // FIXME: even if the next character is OpenParen, it could be an identifier.
+                // e.g.) async();
```
Razican, line 138: Tricky parsing here. Maybe we should assume that it's an identifier, unless we encounter another identifier, for example, afterwards?

Razican (commented):
Thank you for the fix!

I added some comments for potential enhancements. It seems that this test is failing, though:
https://github.com/tc39/test262/blob/f1870753fa616261193b99f79d996889921b3404/test/language/statements/for-of/head-lhs-async-invalid.js

It seems that the async cannot be placed in a for-of. The spec for this is [here](https://tc39.es/ecma262/#sec-for-in-and-for-of-statements). It seems it requires a `var`, a `let` or a `const` for it to be valid.

boa_engine/src/syntax/parser/cursor/mod.rs
tunz: I just found these two patterns are frequently used in other places as well, so I made these functions for other uses. I just didn't want to update existing code in this PR. So, yeah it's up to you. I don't mind removing them and using existing functions.

boa_engine/src/syntax/parser/expression/primary/mod.rs
tunz: async lambda function does not have an identifier as a next token. e.g.) `async () => { }`.

boa_engine/src/syntax/parser/cursor/mod.rs
Razican: I would re-use the existing ones. The inclusion of new functions (and using them throughout the code) should come in a different PR.

boa_engine/src/syntax/parser/expression/primary/mod.rs
Razican: Makes sense, yes, thanks!

Razican:
It seems that the fix for the for...off has broken these tests, that were passing with your previous version:
 - https://github.com/tc39/test262/blob/f1870753fa616261193b99f79d996889921b3404/test/language/statements/for-of/head-lhs-async-parens.js
 - https://github.com/tc39/test262/blob/f1870753fa616261193b99f79d996889921b3404/test/language/statements/for-of/head-lhs-async-escaped.js

It turns out that the `async` keyword can be valid in a `for...of` but only if it's either surrounded with parenthesis or has escaped characters. The "escape characters" information is surely at the lexing level, but we might be missing it, I'm not sure though where is the parenthesis information getting lost.

It's not that critical, since this was broken before, but if you can find a solution that would be perfect :)

boa_engine/src/syntax/parser/expression/primary/mod.rs
tunz: filed a bug here: https://github.com/boa-dev/boa/issues/2275

Razican (approved):
It's ready to be merged from my side :) thanks!

boa_engine/src/syntax/parser/expression/primary/mod.rs
```diff
@@ -126,17 +126,47 @@ where
...
+                            if contain_escaped_char =>
+                        {
+                            let tok = cursor.peek(0, interner)?.ok_or(ParseError::AbruptEnd)?;
+                            Err(ParseError::general(
+                                "Keyword must not contain escaped characters",
+                                tok.span().start(),
+                            ))
+                        }
+                        TokenKind::Keyword((Keyword::Function, _))
+                        | TokenKind::Punctuator(Punctuator::OpenParen) => {
```
raskad, line 151: Why are we looking for OpenParen here? That would be the start of a CallExpression or an AsyncArrowFunction and neither of those are parsed in PrimaryExpression right?

I think when we have implemented the parsing of AsyncArrowFunction this pattern should probably never be valid here, because it would be parsed earlier.

tunz: I looked into it again, and it seems like I was looking at wrong code. yea, you're right. this check should've done in other place. Thanks for catching this!

raskad (approved):
Great work!

PR #2272, opened 2022-09-08
[Merged by Bors] - Added a bit more integer operation consistency to ByteDataBlock creation

Razican:
This Pull Request fixes a potential overflow when trying to convert a `u64` into a `usize` and then trying to create a byte data block.

Related to this, we seem to be using a `u64` and `i64` as a general approach for an "integer", but ECMAScript doesn't have bounds for them, so they could be as big as infinite. Should we use `u128` and `i128` to have a bigger range?

This would add a performance penalty, though, and we don't have 128-bit platforms usually, so the benefit would probably be minimal, at least when trying to allocate.


raskad (approved):
Good change. 

Regarding the interer sizes, I think we should open an issue for that and investigate what other engines do or if we can find any guidance in spec discussions.

jedel1043 (approved):
Looks good!

PR #2273, opened 2022-09-09
[Merged by Bors] - Fix property access of call expression

tunz:
Fix a syntax error when accessing literal-like property names of call expression.
e.g.) `fn().true`

### DIFF
--- a/boa_engine/src/syntax/parser/expression/left_hand_side/call.rs
+++ b/boa_engine/src/syntax/parser/expression/left_hand_side/call.rs
@@ -97,10 +97,11 @@ where
                         TokenKind::Keyword((kw, _)) => {
                             lhs = GetConstField::new(lhs, kw.to_sym(interner)).into();
                         }
-                        TokenKind::BooleanLiteral(bool) => {
-                            lhs =
-                                GetConstField::new(lhs, if *bool { Sym::TRUE } else { Sym::FALSE })
-                                    .into();
+                        TokenKind::BooleanLiteral(true) => {
+                            lhs = GetConstField::new(lhs, Sym::TRUE).into();
+                        }
+                        TokenKind::BooleanLiteral(false) => {
+                            lhs = GetConstField::new(lhs, Sym::FALSE).into();
                         }
                         TokenKind::NullLiteral => {
                             lhs = GetConstField::new(lhs, Sym::NULL).into();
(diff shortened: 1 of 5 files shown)


Razican:
  
  #### VM implementation
  | Passed | 64,890 | 64,894 | **+4** |
  | Failed | 10,263 | 10,259 | **-4** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 70.74% | 70.74% | **+0.00%** |
  
  <details><summary><b>Fixed tests (4):</b></summary>
  
  ```
  test/language/statements/class/elements/private-methods/prod-private-method-initialize-order.js [strict mode] (previously Failed)
  test/language/statements/class/elements/private-methods/prod-private-method-initialize-order.js (previously Failed)
  test/language/expressions/class/elements/private-methods/prod-private-method-initialize-order.js [strict mode] (previously Failed)
  test/language/expressions/class/elements/private-methods/prod-private-method-initialize-order.js (previously Failed)
  ```
  </details>

boa_engine/src/syntax/parser/expression/left_hand_side/call.rs
```diff
@@ -97,6 +97,25 @@ where
...
+                            match bool {
+                                true => {
+                                    lhs = GetConstField::new(lhs, Keyword::True.to_sym(interner))
+                                        .into();
+                                }
+                                false => {
+                                    lhs = GetConstField::new(lhs, Keyword::False.to_sym(interner))
+                                        .into();
+                                }
+                            };
```
Razican, line 110: Isn't it easier to use `if bool {` here? Having a match over a boolean instead of an if...else seems a bit weird. Also, I'm surprised that we need to do `Keyword::True.to_sym(interner)` instead of just using `Sym::TRUE`. `true`/`false` are pretty used strings, we might want to add them as constants. As a suggestion on the implementation:

```suggestion
                            lhs = GetConstField::new(lhs, if bool { Keyword::True } else { Keyword::False }.to_sym(interner)).into();
```

This probably needs a `cargo fmt`, though.

Razican (commented):
Looks good! Thanks for the fix :) Just a suggestion to reduce code repetition and make things more readable.

raskad (approved):
Nice work! Thank you for the additional tests and match fixes.

Razican (approved):
Looks good to me too! Thanks! LGTM

PR #2277, opened 2022-09-12
[Merged by Bors] - Fix regex literal /[/]/

tunz:
This PR fixes a case where a forward slash is located in a regex class: `let regex = /[/]/;`. In this case, the forward slash should not close the regex literal.

This fixes `test/built-ins/RegExp/regexp-class-chars.js`

### DIFF
--- a/boa_engine/src/syntax/lexer/regex.rs
+++ b/boa_engine/src/syntax/lexer/regex.rs
@@ -41,6 +41,7 @@ impl<R> Tokenizer<R> for RegexLiteral {
         let _timer = Profiler::global().start_event("RegexLiteral", "Lexing");
 
         let mut body = Vec::new();
+        let mut is_class_char = false;
 
         // Lex RegularExpressionBody.
         loop {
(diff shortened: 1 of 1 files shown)


raskad:
#### VM implementation
  | Passed | 64,940 | 64,942 | **+2** |
  | Failed | 10,213 | 10,211 | **-2** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 70.79% | 70.79% | **+0.00%** |
  
  ```
  test/built-ins/RegExp/regexp-class-chars.js [strict mode] (previously Failed)
  test/built-ins/RegExp/regexp-class-chars.js (previously Failed)
  ```
  </details>

raskad (commented):
Nice find. I think there is still a problem if an escaped `]` is placed before the end of the character class right?
`/[\]/]/` should be a valid regex but would not be lexed correctly with this approach.

tunz:
@raskad Could you elaborate? I think there is no problem with `/[\]/]/`. The escaped `]` should be handled by the backslash pattern match (line 81).

raskad:
> @raskad Could you elaborate? I think there is no problem with `/[\]/]/`. The escaped `]` should be handled by the backslash pattern match (line 81).

True, the error I was looking at is unrelated and only happens in the boa_cli.

raskad (approved):
Very nice fix, thank you!

Razican (approved):
Good catch! Thanks for the fix :)

Razican:
Bors r+

PR #2283, opened 2022-09-15
[Merged by Bors] - Create new lazy Error type

jedel1043:
This is an experiment that tries to migrate the codebase from eager `Error` objects to lazy ones.

In short words, this redefines `JsResult = Result<JsValue, JsError>`, where `JsError` is a brand new type that stores only the essential part of an error type, and only transforms those errors to `JsObject`s on demand (when having to pass them as arguments to functions or store them inside async/generators).

This change is pretty big, because it unblocks a LOT of code from having to take a `&mut Context` on each call. It also paves the road for possibly making `JsError` a proper variant of `JsValue`, which can be a pretty big optimization for try/catch.

A downside of this is that it exposes some brand new error types to our public API. However, we can now implement `Error` on `JsError`, making our `JsResult` type a bit more inline with Rust's best practices.

~Will mark this as draft, since it's missing some documentation and a lot of examples, but~ it's pretty much feature complete. As always, any comments about the design are very much appreciated!

Note: Since there are a lot of changes which are essentially just rewriting `context.throw` to `JsNativeError::%type%`, I'll leave an "index" of the most important changes here:

- [boa_engine/src/error.rs](https://github.com/boa-dev/boa/pull/2283/files#diff-f15f2715655440626eefda5c46193d29856f4949ad37380c129a8debc6b82f26)
- [boa_engine/src/builtins/error/mod.rs](https://github.com/boa-dev/boa/pull/2283/files#diff-3eb1e4b4b5c7210eb98192a5277f5a239148423c6b970c4ae05d1b267f8f1084)
- [boa_tester/src/exec/mod.rs](https://github.com/boa-dev/boa/pull/2283/files#diff-fc3d7ad7b5e64574258c9febbe56171f3309b74e0c8da35238a76002f3ee34d9)

### DIFF
--- a/boa_engine/Cargo.toml
+++ b/boa_engine/Cargo.toml
@@ -53,6 +53,7 @@ once_cell = "1.15.0"
 tap = "1.0.1"
 sptr = "0.3.2"
 static_assertions = "1.1.0"
+thiserror = "1.0.35"
 icu_locale_canonicalizer = { version = "0.6.0", features = ["serde"], optional = true }
 icu_locid = { version = "0.6.0", features = ["serde"], optional = true }
 icu_datetime = { version = "0.6.0", features = ["serde"], optional = true }
--- a/boa_engine/Cargo.toml
+++ b/boa_engine/Cargo.toml
@@ -53,7 +53,7 @@ once_cell = "1.15.0"
 tap = "1.0.1"
 sptr = "0.3.2"
 static_assertions = "1.1.0"
-thiserror = "1.0.35"
+thiserror = "1.0.37"
 icu_locale_canonicalizer = { version = "0.6.0", features = ["serde"], optional = true }
 icu_locid = { version = "0.6.0", features = ["serde"], optional = true }
 icu_datetime = { version = "0.6.0", features = ["serde"], optional = true }
--- a/boa_engine/src/builtins/error/mod.rs
+++ b/boa_engine/src/builtins/error/mod.rs
@@ -13,6 +13,7 @@
 use crate::{
     builtins::BuiltIn,
     context::intrinsics::StandardConstructors,
+    error::JsNativeError,
     js_string,
     object::{
         internal_methods::get_prototype_from_constructor, ConstructorBuilder, JsObject, ObjectData,
@@ -44,6 +45,18 @@ pub(crate) use self::syntax::SyntaxError;
 pub(crate) use self::uri::UriError;
 
 use super::JsArgs;
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum ErrorKind {
+    Aggregate,
+    Error,
+    Eval,
+    Type,
+    Range,
+    Reference,
+    Syntax,
+    Uri,
+}
 
 /// Built-in `Error` object.
 #[derive(Debug, Clone, Copy)]
@@ -109,7 +122,7 @@ impl Error {
         // 2. Let O be ? OrdinaryCreateFromConstructor(newTarget, "%Error.prototype%", « [[ErrorData]] »).
         let prototype =
             get_prototype_from_constructor(new_target, StandardConstructors::error, context)?;
-        let o = JsObject::from_proto_and_data(prototype, ObjectData::error());
+        let o = JsObject::from_proto_and_data(prototype, ObjectData::error(ErrorKind::Error));
 
         // 3. If message is not undefined, then
         let message = args.get_or_undefined(0);
@@ -149,7 +162,9 @@ impl Error {
             o
         // 2. If Type(O) is not Object, throw a TypeError exception.
         } else {
-            return context.throw_type_error("'this' is not an Object");
+            return Err(JsNativeError::typ()
+                .with_message("'this' is not an Object")
+                .into());
         };
 
         // 3. Let name be ? Get(O, "name").
--- a/boa_engine/src/builtins/error/mod.rs
+++ b/boa_engine/src/builtins/error/mod.rs
@@ -46,6 +46,18 @@ pub(crate) use self::uri::UriError;
 
 use super::JsArgs;
 
+/// The kind of a `NativeError` object, per the [ECMAScript spec][spec].
+///
+/// This is used internally to convert between [`JsObject`] and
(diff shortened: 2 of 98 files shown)


jedel1043:
Finished documenting and adding examples. Let me know if I overlooked something :)

boa_engine/src/object/jsset.rs
```diff
@@ -76,7 +77,8 @@ impl JsSet {
     {
         match Set::delete(&self.inner.clone().into(), &[value.into()], context)? {
             JsValue::Boolean(bool) => Ok(bool),
-            _ => Err(JsValue::Undefined),
+            // TODO: find a better solution to this, without impacting perf
+            _ => unreachable!("`delete` must always return a bool"),
```
Razican, line 81: I'm not 100% sure about these. Why is this like this?

boa_engine/src/vm/code_block.rs
```diff
@@ -1436,8 +1448,8 @@ impl JsObject {
                             .expect("this binding must be object")
                             .clone())
                     } else {
-                        //context.throw_type_error("Function constructor must not return non-object")
-                        context.throw_reference_error("Must call super constructor in derived class before accessing 'this' or returning from derived constructor")
+                        //Err(JsNativeError::typ().with_message("Function constructor must not return non-object").into())
```
Razican, line 1451: Why is this commented out?

boa_engine/src/error.rs
```diff
@@ -0,0 +1,787 @@
...
+
+// hiding the enum makes it a bit more difficult
+// to treat `JsError` as a proper enum, which should
+// make it a bit harder to try to match against a native error
+// without calling `try_native` first.
+#[derive(Debug, Clone, Trace, Finalize)]
+enum Repr {
+    Native(JsNativeError),
+    Opaque(JsValue),
+}
```
Razican, line 61: Even if we want to hide it, I would add some documentation explaining its usage, with triple slashes: `///`

We could also add this to that documentation.

boa_engine/src/error.rs
```diff
@@ -0,0 +1,787 @@
...
+///
+/// assert_eq!(native_error.message(), "cannot decode uri");
+/// ```
+#[derive(Debug, Clone, Trace, Finalize, Error)]
+#[error("{kind}: {message}")]
+pub struct JsNativeError {
+    /// The kind of native error (e.g. `TypeError`, `SyntaxError`, etc.)
+    pub kind: JsNativeErrorKind,
+    #[unsafe_ignore_trace]
+    message: LocalStr,
```
Razican, line 367: Do we really need this now? I understand this is more space efficient and has performance benefits, but we are adding 2 new dependencies, so it might make sense to think about this thoroughly and implement them system-wide. We also talked about small_str, and we decided in the end to intern strings, so that might be another option.

boa_engine/Cargo.toml
```diff
@@ -50,6 +50,8 @@ unicode-normalization = "0.1.22"
 dyn-clone = "1.0.9"
 once_cell = "1.15.0"
 tap = "1.0.1"
+thiserror = "1.0.35"
```
Razican, line 53: Do we really need another dependency here? I think it's only being used for one type, right? It would make sense if we extend its usage, but we might want it to be a different PR.

This happens also with `flexstr`.

Razican (commented):
I'm very happy about this PR. If we can extend this design to all native types it could increase performance A LOT.

Give a look to my comments and let me know your thoughts.

boa_engine/src/error.rs
jedel1043: Hmm yeah, maybe I prematurely optimized this. I'll replace it with a simple `Rc<str>`.

boa_engine/Cargo.toml
jedel1043: Two types, actually: `JsError` and `TryNativeError`. I'd advocate for adding `thiserror` though, since it pulls very few dependencies (which are probably on our tree already), and it could also be used to implement `Error` for `ParseError`. I'll remove `flexstr`.

boa_engine/src/object/jsset.rs
jedel1043: Yeah, the wording is a bit confusing. Reworded it to something more descriptive.

boa_engine/src/vm/code_block.rs
jedel1043: Oh, I was debugging and forgot to remove the cause of the bug 😅

boa_engine/src/builtins/error/mod.rs
```diff
@@ -43,6 +44,18 @@ pub(crate) use self::uri::UriError;
...
+pub enum ErrorKind {
+    Aggregate,
+    Error,
+    Eval,
+    Type,
+    Range,
+    Reference,
+    Syntax,
+    Uri,
+}
```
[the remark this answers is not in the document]
jedel1043: I'll just relink to `JsNativeError`, since this is mostly internal. I'll also make it crate-only, just to ensure it is always set to the correct variant.

boa_engine/src/vm/code_block.rs
jedel1043: Oh, I was debugging and forgot to remove the cause of the bug 😅

Razican (approved):
This is ready for me :) thank you!

It only needs a re-base.

boa_engine/src/error/mod.rs
```diff
@@ -0,0 +1,767 @@
...
+    /// # use boa_engine::{JsNativeError, JsNativeErrorKind};
+    /// let error = JsNativeError::uri();
+    ///
+    /// assert!(matches!(error.kind, JsNativeErrorKind::Uri));
+    /// ```
+    pub fn uri() -> Self {
+        Self::new(JsNativeErrorKind::Uri, Box::default(), None)
+    }
+
+    /// Appends a message to this error.
```
RageKnify, line 507:
```suggestion
    /// Sets the message of this error.
```
Don't like using the word append here, it gives the idea of extending a String of Vec.
Same for `with_cause` below

PR #2297, opened 2022-09-23
[Merged by Bors] - Switch to workspace inherited properties

jedel1043:
This Pull Request switches our codebase to the brand new [workspace inherited keys](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table), which allows us to define common package options that are usable within each crate's Cargo.toml file.

It also allows to share dependency versions between crates, but I defined only shared versions for our workspace members. It would be a good follow-up to lift all the shared dependencies between crates into the global Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,22 @@ members = [
     "boa_examples",
 ]
 
+[workspace.package]
+edition = "2021"
+version = "0.16.0"
+rust-version = "1.60"
+authors = ["boa-dev"]
+repository = "https://github.com/boa-dev/boa"
+license = "Unlicense/MIT"
+description = "Boa is a Javascript lexer, parser and Just-in-Time compiler written in Rust. Currently, it has support for some of the language."
+
+[workspace.dependencies]
+boa_engine = { version = "0.16.0", path = "boa_engine" }
+boa_interner = { version = "0.16.0", path = "boa_interner" }
+boa_gc = { version = "0.16.0", path = "boa_gc" }
+boa_profiler = { version = "0.16.0", path = "boa_profiler" }
+boa_unicode = { version = "0.16.0", path = "boa_unicode" }
+
 [workspace.metadata.workspaces]
 allow_branch = "main"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
 [workspace.package]
 edition = "2021"
 version = "0.16.0"
-rust-version = "1.60"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -11,6 +11,22 @@ members = [
     "boa_examples",
 ]
 
+[workspace.package]
+edition = "2021"
+version = "0.15.0"
+rust-version = "1.60"
```
Razican, line 17: We might need to specify 1.64 here, since this is only available starting on Rust 1.64 (or cargo 1.64, in fact)

jedel1043: Yep!

jedel1043:
Apparently dependabot doesn't have support for inherited dependencies (tracked in https://github.com/dependabot/dependabot-core/issues/5315), so lifting the common dependencies is blocked by that issue.

raskad (approved):
Looks good to me.
We should probably keep this change in mind for the 0.17 release, to see if it works correct with `cargo-workspaces publish`.

Cargo.toml
```diff
@@ -11,6 +11,22 @@ members = [
...
+version = "0.15.0"
+rust-version = "1.64"
+authors = ["boa-dev"]
+repository = "https://github.com/boa-dev/boa"
+license = "Unlicense/MIT"
+description = "Boa is a Javascript lexer, parser and Just-in-Time compiler written in Rust. Currently, it has support for some of the language."
+
+[workspace.dependencies]
+boa_engine = { version = "0.15.0", path = "boa_engine" }
+boa_interner = { version = "0.15.0", path = "boa_interner" }
```
jasonwilliams, line 25: What happens if version is not specified?

jedel1043: Crates.io rejects publishing the crate

jasonwilliams (approved):

jedel1043:
Should we push this into 0.16, just to see if `cargo-workspaces` work? We can rollback if there's any problem

poliorcetics:
You won't have any rust-related dependency update from dependabot now I think since it doesn't yet know where to find the version number for a `dep.workspace = true` dependency, even if that dependency is local.

I opened a PR for it https://github.com/dependabot/dependabot-core/pull/5794, it should work but it's not merged yet.

It's probably possible to regenerate the docker container and use it locally to make the necessary changes but I have no idea how to do that. I'll gladly accept people testing it and telling me if it works has expected in a variety of configs though 👍

PR #2305, opened 2022-09-25
[Merged by Bors] - Fix function property order

raskad:

- Fix function property order


jedel1043 (approved):
Very interesting! Didn't know function properties ought to have a specific order. Very good catch!

Razican (approved):
Good one! I'm also surprised about this!

PR #2306, opened 2022-09-25
[Merged by Bors] - Fix some Array spec deviations

raskad:

- Fix array constructor realm comparison.
- Fix error that ignored `ToObject`.
- Ignore `array-grouping` feature tests.

### DIFF
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -346,11 +346,16 @@ impl Array {
         let c = original_array.get("constructor", context)?;
 
         // 4. If IsConstructor(C) is true, then
-        //     a. Let thisRealm be the current Realm Record.
-        //     b. Let realmC be ? GetFunctionRealm(C).
-        //     c. If thisRealm and realmC are not the same Realm Record, then
-        //         i. If SameValue(C, realmC.[[Intrinsics]].[[%Array%]]) is true, set C to undefined.
-        // TODO: Step 4 is ignored, as there are no different realms for now
+        if let Some(c) = c.as_constructor() {
+            // a. Let thisRealm be the current Realm Record.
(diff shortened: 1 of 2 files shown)


boa_engine/src/builtins/array/mod.rs
```diff
@@ -346,11 +346,16 @@ impl Array {
...
-        //     a. Let thisRealm be the current Realm Record.
-        //     b. Let realmC be ? GetFunctionRealm(C).
-        //     c. If thisRealm and realmC are not the same Realm Record, then
-        //         i. If SameValue(C, realmC.[[Intrinsics]].[[%Array%]]) is true, set C to undefined.
-        // TODO: Step 4 is ignored, as there are no different realms for now
+        if let Some(c) = c.as_constructor() {
+            // a. Let thisRealm be the current Realm Record.
+            // b. Let realmC be ? GetFunctionRealm(C).
+            // c. If thisRealm and realmC are not the same Realm Record, then
+            if *c == context.intrinsics().constructors().array().constructor {
```
jedel1043, line 353: Would be good to add a TODO to fix this after we properly implement realms

jedel1043 (changes requested):
Just a small nitpick

boa_engine/src/builtins/array/mod.rs
raskad: I think this does exacly what the spec wants, no? Checking if the function object is that of the current realm?

jedel1043: Ahh right, `context.intrinsics()` will eventually return the current realm record, so it will be the same as checking if `c` belongs to the current realm

jedel1043 (approved):
All good!

Razican (approved):
Thanks!

Razican:
Bors r+

Issue #332, opened 2020-04-18
Any plan for the FFI?


attliaLin:
Nice work!
But as an embeddable engine, I think it would also be nice to provide some API functions for the other languages to use.

Razican:
This would be an interesting addition. I would do it in a different crate in the workspace, with only the FFI, using `boa` as a dependency.

HalidOdat:
We could use [`cbindgen`](https://github.com/eqrion/cbindgen) crate to automatically generate C/C++ FFIs.

attliaLin:
And it would be better to provide more accessibility to the engine, like https://duktape.org/api.html

bcr:
New Guy here. I'm sniffing around this. I think the problem here is going to be more one of defining scope and granularity. Like I think just implementing an `eval` is one thing, but those fine folks at duktape have a dukload of things they expose. Making it similarly comprehensive for Boa is probably... hard. But I can take a whack at `eval` in its own FFI crate.

jedel1043:
We discussed this in our triage and we agreed that we wanted to experiment with the [diplomat](https://github.com/rust-diplomat/diplomat) tool first before commiting to an FFI interface, since it makes the maintenance work more manageable.

nekevss:
Follow up on this, we are currently using `diplomat` for `temporal_rs`'s FFI.

If we were to add a FFI, I'd highly recommend it.

PR #2312, opened 2022-09-26
Start of FFI support

bcr:
This Pull Request is a starting point for #332

It changes the following:

- Adds boa_ffi crate with sample exec / string free functions

There's some more work that should probably get done, documented in the README, but I think I need some philosophical guidance about this.

- [x] Automate cbindgen from build
- [x] Integration tests to make sure it works end-to-end
- [ ] Panic handling?
- [ ] C++ as well as C bindings?
- [x] Where to put `cbindgen` output?
- [x] What `boa_engine` features do we need? I have `boa_engine = { workspace = true, features = ["deser", "console"] }`


jedel1043:
Thank you for your interest on providing a ffi to Boa! We've been postponing this work for some time and it's really helpful to have someone working on this :)

Talking more about the implementation, it would be good to investigate if https://github.com/rust-diplomat/diplomat would be a good option for us.

bcr:
Clippy seems to be humpty about marking the functions as unsafe even though the unsafe part is marked unsafe. If I make the functions unsafe then Rust gets humpty about redundantly using unsafe for the actual unsafe part. Pout.

bcr:
> Thank you for your interest on providing a ffi to Boa! We've been postponing this work for some time and it's really helpful to have someone working on this :)

Happy to try to help. I hope my approach aligns with what you want. My concern is that exposing the full functionality of boa isn't going to happen (at least with my level of ability) and I'm going to have to reduce the scope.
 
> Talking more about the implementation, it would be good to investigate if https://github.com/rust-diplomat/diplomat would be a good option for us.

I'll take a closer look. I'm using cbindgen right now. The diplomat approach looks richer, but... odder. Like they don't decorate methods using the unsurprising Rust FFI conventions, they have custom decorators. Still figuring it out.

jedel1043:
> Clippy seems to be humpty about marking the functions as unsafe even though the unsafe part is marked unsafe. If I make the functions unsafe then Rust gets humpty about redundantly using unsafe for the actual unsafe part. Pout.

Yeah, clippy tends to recommend either removing `unsafe` blocks inside `unsafe` functions or adding `#[deny(unsafe_op_in_unsafe_fn)]` to delimit unsafe blocks even inside unsafe functions. You can add `#![deny(unsafe_op_in_unsafe_fn)]` to the top of the `lib.rs` file.

bcr:
> Yeah, clippy tends to recommend either removing `unsafe` blocks inside `unsafe` functions or adding `#[deny(unsafe_op_in_unsafe_fn)]` to delimit unsafe blocks even inside unsafe functions. You can add `#![deny(unsafe_op_in_unsafe_fn)]` to the top of the `lib.rs` file.

I personally prefer to have the unsafe be as constrained as possible. In the short term I added the warning squelch at the top of the file. I'm looking to see if I can make Clippy complain less instead of making Rust complain less.

Now that I know about the Clippy check, I made sure it is clean now, which also got us documentation for the functions (because Clippy wants a `# Safety` section, which I agree with.)

Razican (commented):
This looks good to me :) I would be happy to merge it as-is if we add tests into the CI, to make sure we don't break the C API.

It's a small step, but I think this is the direction we want to follow.

bcr:
OK, let me put together something that can run under CI. I've got the C code written, it's a matter of putting in some cases where you run the C code with known input and get known output.

bcr:
So here is the state so far. Part of this may be my Rust ignorance, and I'm doing my best to research it.

## Good News
* I can build the C include file using `cbindgen` in the normal Rust build process (this is implemented on this branch)
## Not as Good News
* I can't come up with a way to build a C *executable* using the `cc` crate. It really wants to build libraries. This step is required to turn `test/boa_test.c` into an executable, which I am using as a test executable in the integration tests.
* I can't figure out how to take the header file I generate with `cbindgen` and make it accessible for someone that wants to use it. I was hoping to build the crate and include the file as part of the crate, but I couldn't get it to work. I tried generating the include file to the `OUT_DIR` for my crate and it doesn't show up in the crate. In addition, I think that `cbindgen` might have a bug where it's making a lock file and `crate publish` gets upset that something changed that shouldn't.

bcr:
It looks like #2283 introduced a breaking change that killed the CI build. I intend to rebase my branch to main and fix it.

jedel1043:
Can you rebase instead of merging? It makes it easier to review the changes, since the PR doesn't include all the changes of the merged commits

bcr:
> Can you rebase instead of merging? It makes it easier to review the changes, since the PR doesn't include all the changes of the merged commits

Whoops. 18ff5c6 already happened. So should I hard reset the branch to the previous commit? When I look at the "Files changed" for this PR, it still looks right to me.

jedel1043:
Ah, seems like it was a display bug of the android app. I reloaded the PR and it now shows the changes correctly :)

bcr:
I ended up running the build script once from Rust at the start of my tests to build the C example program used for the tests. This works on my local macOS environment.

However, this is failing in the CI right now. I added a `panic!` on nonzero result from the build script that should give some insight. The CI failures are basically "I couldn't find the executable" which means that the build script failed. So now I added a check for nonzero status from the build script to dump the `Output` and panic with the contents. I also switched from using a build script to use `make` and then I can do whatever makefile shenanigans I need to do if there are platform-specific build problems that I can't see on macOS.

bcr:
I have tarpaulin installed now, and I see the same results from the CI failure, so this is good. The essence of my current problem is that in order to build the C program for testing, I need to have access to the cdylib for my crate, which isn't built when tarpaulin runs. In my environment I didn't see this initially because I had a pre-built shared library lying around from a previous build.

***In order to run the integration tests with the C test harness, I need to have the `libboa_ffi.so` available in a consistent location so that the C program can be made using `make` and then be able to use the shared library at runtime.***

The tests are currently being run as regular Rust tests as part of the normal Rust way that it runs tests. Which apparently does not include building the dynamic library target (which hurts my head, but this is the case.)

I am thinking about the following options:

* Induce the build of the shared library when tarpaulin is run
* Run the integration tests at a later point in the CI process

I really do want to be respectful of the normal workflow for boa builds, and respectful of the Rustacean Way of Doing Things. I'm mulling over what to do next, any suggestions appreciated.

jedel1043:
Yeah, maybe trying to call a C process from a Rust test suite isn't the best idea. I'd argue that we should compile and test C scripts, maybe hooking it up with `cargo-make` to run all C tests. ICU4X has a good example of this:

Tests: https://github.com/unicode-org/icu4x/tree/main/ffi/diplomat/c/examples

CI workflow: https://github.com/unicode-org/icu4x/blob/2db2c5f3b1ee2a44ff8d9356fd4044844bd91564/.github/workflows/build-test.yml#L220-L280

`cargo make` called by the ffi workflow: https://github.com/unicode-org/icu4x/blob/2db2c5f3b1ee2a44ff8d9356fd4044844bd91564/Makefile.toml#L87-L99

script imported by the  `cargo make ci-job-ffi` command: https://github.com/unicode-org/icu4x/blob/2db2c5f3b1ee2a44ff8d9356fd4044844bd91564/tools/scripts/ffi.toml

jedel1043:
Closing this per https://github.com/boa-dev/boa/issues/332#issuecomment-1832444667

PR #2314, opened 2022-09-27
create intrinsics lazily, after we know parsing is successful. fixes #2307

### LINKED ISSUE #2307: [performance] - Opportunity to run parsing before building environment
Currently the environment is built before parsing begins. This happens when context [is created](https://github.com/boa-dev/boa/blob/main/boa_cli/src/main.rs#L186). The downside to this is if we have any parsing error we still need to build the whole env first which is wasteful as we never needed it anyway.  Another opportunity this could open up is using multithreading for parsing and building the env further down the line.

Parsing does have a dependency on the environment being around and that is for [checking duplicate bindings](https://github.com/boa-dev/boa/blob/main/boa_engine/src/syntax/parser/mod.rs#L276-L319). However this could be moved out of the parsing stage and checked after the environment is created. If this is only needed for scripts the environment shouldn't even be needed at all

jasonwilliams:
This moves intrinsics from the beginning to after parsing. With this change we can "fail fast" if there's a parsing/syntax error. Currently intrinsics need to be built before parsing the code which is wasteful.

The syntax errors are hardcoded to a JSValue as not to use the environment when we already know what the error type will be.

### Before
image

### After
image

Fixes: https://github.com/boa-dev/boa/issues/2307

### DIFF
--- a/boa_engine/src/context/mod.rs
+++ b/boa_engine/src/context/mod.rs
@@ -19,7 +19,11 @@ use crate::{
     object::{FunctionBuilder, GlobalPropertyMap, JsObject, ObjectData},
     property::{Attribute, PropertyDescriptor, PropertyKey},
     realm::Realm,
-    syntax::{ast::node::StatementList, parser::ParseError, Parser},
+    syntax::{
+        ast::node::StatementList,
+        parser::{ParseError, Script},
+        Parser,
+    },
     vm::{CallFrame, CodeBlock, FinallyReturn, GeneratorResumeKind, Vm},
     JsResult, JsValue,
 };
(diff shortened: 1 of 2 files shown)


jedel1043:
This will probably benefit fom #2283, since that'll separate error creation from the context :)

jedel1043:
@jasonwilliams Do you have some progress on this? We can probably rebase this since the new error types makes this an easier change.

jasonwilliams:
Thanks for pinging me on this, I think we can close this as it  needs a rethink. If someone creates a new error handler and we parse incorrect syntax, we would need to execute that instead of the default syntaxError which may require us having the realm initiated. So this may not work as I expected when I started this

PR #2319, opened 2022-10-01
[Merged by Bors] - Split `Node` into `Statement`, `Expression` and `Declaration`

### LINKED ISSUE #1371: Split `Node` into `Expression`, `Statement` and `Declaration`

jedel1043:
This Pull Request fixes #1371. And yeah, the number of file changes is real...

It changes the following:

- Split the ast `Node` into `Statement` and `Expression`. 
- Rewrite the parser and bytecompiler to conform to this change.
- Refactor some ast nodes into reusable structures.
- Rewrite `contains_arguments` and `contains` to ease the transition into a future ast visitor.

List of things that were apparently fixed by this refactor?:
- Implement read-assign operation for private accessors (e.g. `this.#field ||= 5`).
- `var await` declaration now allowed outside `async` functions and inside functions nested in async functions.
- Reject redeclarations of variables declared in the init list of a for loop.

Still missing some documentation adjustments, will try to do it ASAP.


raskad (approved):
Impressive work!
I saw some structs and methods missing docs, but I would definitley be fine merging this as-is to prevent needing to rebase it later.

boa_engine/src/syntax/ast/expression/access.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// Gets the original object from where to get the field from.
+    pub fn target(&self) -> &Expression {
+        &self.target
+    }
+
+    /// Gets the name of the field to retrieve.
+    pub fn field(&self) -> Sym {
+        self.field
+    }
+    #[inline]
```
Razican, line 153:
```suggestion

    #[inline]
```

boa_engine/src/syntax/ast/expression/access.rs
```diff
@@ -0,0 +1,232 @@
+use crate::syntax::ast::{expression::Expression, ContainsSymbol};
+use boa_interner::{Interner, Sym, ToInternedString};
+
+#[cfg_attr(feature = "deser", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq)]
+pub enum PropertyAccessField {
+    Const(Sym),
+    Expr(Box<Expression>),
```
Razican, line 8: Do we need these to be boxed? I guess it's a space optimization, right?

boa_engine/src/syntax/ast/expression/await.rs
```diff
@@ -0,0 +1,74 @@
...
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await
+#[cfg_attr(feature = "deser", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq)]
+pub struct Await {
+    expr: Box<Expression>,
+}
+
+impl Await {
+    /// Return the expression that should be awaited.
+    pub(crate) fn expr(&self) -> &Expression {
```
Razican, line 25:
```suggestion
    #[inline]
    pub(crate) fn expr(&self) -> &Expression {
```

boa_engine/src/syntax/ast/expression/call.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Gets the name of the function call.
+    pub fn expr(&self) -> &Expression {
+        &self.target
+    }
+
+    /// Retrieves the arguments passed to the function.
+    pub fn args(&self) -> &[Expression] {
+        &self.args
+    }
```
Razican, line 44: We seem to be adding `#[inline]` on the other methods, and these are getters/constructors, so it makes sense to inline them, right?

Same happens in other nodes.

boa_engine/src/syntax/ast/expression/literal/object/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[cfg(test)]
+mod tests;
+
+use crate::syntax::ast::expression::Expression;
+use crate::syntax::ast::ContainsSymbol;
+
+use crate::syntax::ast::property::MethodDefinition;
+use crate::syntax::ast::property::PropertyDefinition;
+use crate::syntax::ast::{block_to_string, join_nodes};
```
Razican, line 11: I would group these imports.

Razican (commented):
Impressive work! This is a great addition. I still didn't have the time to look over everything (150 files left :O) But it's looking very good.

I'd like to have your thoughts about boxing all expressions. It makes sense from a space perspective, but might be less efficient. It might make sense to box them at the `Node` level only.

boa_engine/src/syntax/ast/expression/access.rs
jedel1043: Yeah, we need to box it. `Expression` contains `PropertyAccess`, which contains `PropertyAccessField`, meaning we can't stack-allocate `PropertyAccessField` without boxing either the whole `PropertyAccessField` or `Expression`, and I think it's a lot more space and speed efficient to box the `Expression` and leave `Sym` on the stack, which is pointer-sized.

boa_engine/src/syntax/ast/expression/call.rs
jedel1043: Let me see if I can make Clippy do this for me

jedel1043:
Had to split `Statement` into `Statement` and `Declaration` since `LabelledStatement` doesn't accept `Declaration` as a labellable node. This will temporarily break some tests, but I already have another PR implementing `LabelledStatement` that fixes this.

RageKnify (approved):
Really cool change, definitely appreciate differentiating things more by having expressions and statements not be variants of the same thing.

jedel1043:
This is missing a lot of documentation, but I'll merge this as it is, since the number of changes is pretty massive and I wouldn't like to add more changes on top of that.

I'll follow up with two PRs that implement `LabelledStatement` and add the missing documentation.

bors r+

PR #2320, opened 2022-10-02
[Merged by Bors] - Fixing the output for the test diffs in PRs

Razican:
This PR fixes PR comments where negative numbers from 100 to 999 would be shown as `-,999` instead of `-999`

### DIFF
--- a/boa_tester/src/results.rs
+++ b/boa_tester/src/results.rs
@@ -264,7 +264,13 @@ pub(crate) fn compare_results(base: &Path, new: &Path, markdown: bool) {
             format!(
                 "{}{}{}{}",
                 if diff == 0 { "" } else { "**" },
-                if diff.is_positive() { "+" } else { "-" },
+                if diff.is_positive() {
+                    "+"
+                } else if diff.is_negative() {
+                    "-"
+                } else {
+                    ""
+                },
                 pretty_int(diff),
                 if diff == 0 { "" } else { "**" }
             )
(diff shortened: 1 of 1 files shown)


boa_tester/src/results.rs
```diff
@@ -264,7 +264,7 @@ pub(crate) fn compare_results(base: &Path, new: &Path, markdown: bool) {
             format!(
                 "{}{}{}{}",
                 if diff == 0 { "" } else { "**" },
-                if diff > 0 { "+" } else { "" },
+                if diff.is_positive() { "+" } else { "-" },
```
RageKnify, line 267: Think this resulted in the `-0`s,  swapping the logic might be a good idea? Check if it's negative, and if so use `-`, this way we get `+0` instead of `-0`.
```suggestion
                if diff.is_negative() { "-" } else { "+" },
```

RageKnify (changes requested):

boa_tester/src/results.rs
Razican: I added an extra check, and that way 0 will not be positive or negative :)

jedel1043 (approved):

boa_tester/src/results.rs
Razican: @RageKnify let me know if the new solution is OK and we can merge this :)

RageKnify (approved):

RageKnify (approved):

PR #2336, opened 2022-10-10
[Merged by Bors] - Upgrade clap to 4.0, add value hints for zsh and fish

Razican:
This Pull Request fixes/closes #2330.

It changes the following:

- Upgrades from clap 3 to clap 4 (note that criterion still uses clap 3, tracked here: https://github.com/bheisler/criterion.rs/issues/596)
- Updates the derive syntax with the new 4.0 syntax
- Adds hints for fish & zsh


boa_tester/src/main.rs
```diff
@@ -194,41 +194,42 @@ static IGNORED: Lazy<Ignored> = Lazy::new(|| {
...
 /// Boa test262 tester
 #[derive(Debug, Parser)]
-#[clap(author, version, about, name = "Boa test262 tester")]
+#[command(author, version, about, name = "Boa test262 tester")]
 enum Cli {
     /// Run the test suite.
     Run {
         /// Whether to show verbose output.
-        #[clap(short, long, parse(from_occurrences))]
+        #[arg(short, long, action = clap::ArgAction::Count)]
```
jedel1043, line 202: I'd just import `ArgAction`, since there are already other structs imported and using fully qualified syntax in this case doesn't do much.

jedel1043 (approved):
Looks good! Just a small nitpick that comes down to preference, really.

jedel1043:
Huh, the update probably changed some syntax of the `boa_tester` tool...

Razican:
> Huh, the update probably changed some syntax of the `boa_tester` tool...

Should be fixed now :)

jasonwilliams (approved):

PR #2363, opened 2022-10-21
[Merged by Bors] - Modified the `loadfile` example to show how Boa can read bytes

Razican:
Boa's `Context::eval()` and `Context::parse()` functions allow reading a byte slice directly, apart from a string. We were not showing this properly in our examples, so this modifies the `loadfile.rs` example to use a byte vector instead of a string, which allows us to use UTF-16 files, if I'm not mistaken.

### DIFF
--- a/boa_examples/src/bin/loadfile.rs
+++ b/boa_examples/src/bin/loadfile.rs
@@ -1,14 +1,14 @@
 // This example shows how to load, parse and execute JS code from a source file
 // (./scripts/helloworld.js)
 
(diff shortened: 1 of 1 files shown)


HalidOdat (approved):

jedel1043:
> which allows us to use UTF-16 files, if I'm not mistaken.

Currently no, since the lexer and the parser assume UTF-8. Though, this will change when we make them generic over chars/u32 instead.

jedel1043 (approved):

PR #2364, opened 2022-10-21
[Merged by Bors] - Removed the "VM Implementation" headline for Test262

Razican:
We were always getting a "VM Implementation" headline in Test262 results for PRs. This was a leftover from the time when we had 2 distinct implementations for the execution. This is no longer the case, so we can remove it to reduce clutter.


HalidOdat (approved):
Looks good to me :)

jedel1043 (approved):

PR #2365, opened 2022-10-21
[Merged by Bors] - Updated the Code of Conduct

Razican:
This updates the Code of Conduct to the Contributor Covenant v2.1 version (we were using the 2.0 version until now). Changes are very minor. As far as I can tell, the only content difference is the explicit mention of _caste_ and _color_ among the potential traits that could cause harassment.

Then it just changes the format a bit to make links a bit more user friendly, by not posting URLs directly.

But I have a question: should we update the contact method? It currently says the following:

> Instances of abusive, harassing, or otherwise unacceptable behavior may be reported to the community leaders responsible for enforcement at [discord](https://discord.gg/tUFFk9Y) by contacting _JaseW_.

Maybe we should mention the `@boa-dev` group, and that they can contact anyone there. What do you think? I would also like to update this before https://github.com/boa-dev/ryu-js/pull/22

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -60,7 +60,8 @@ representative at an online or offline event.
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
 reported to the community leaders responsible for enforcement at
-[discord](https://discord.gg/tUFFk9Y) by contacting _JaseW_.
+[discord](https://discord.gg/tUFFk9Y) by contacting anyone in the _@boa_dev_
+group (check the yellow usernames).
 All complaints will be reviewed and investigated promptly and fairly.
 
 All community leaders are obligated to respect the privacy and security of the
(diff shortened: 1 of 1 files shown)


jedel1043:
Yeah, I think it would be better if people with a report could contact any member of our core team, just in case Jason is not available at that moment.

RageKnify:
Agree, I think we can all be relied upon to handle first contact for any problem.

HalidOdat:
Yes, agreed.

I will label as https://github.com/boa-dev/ryu-js/pull/22 as blocked until this PR is merged :)

Razican:
I have updated the contact method to the whole team. I think that everyone should approve it before merging. @jasonwilliams, @HalidOdat, @RageKnify, @raskad and @jedel1043

RageKnify (approved):

raskad (approved):

HalidOdat (approved):

jedel1043 (approved):

jasonwilliams (approved):

PR #2367, opened 2022-10-22
[Merged by Bors] - Removed some duplicate code, added `ToIndentedString`

Razican:
In most cases, the `ToInternedString` was just calling `self.to_indented_string(interner, 0)`. This avoids all this duplicate code by adding a new trait, `ToIndentedString`. Any type implementing that automatically implements `ToInternedString`.

I have also added a bunch of `#[inline]` in one-liners, and some one-line documentations for some functions.

I have noticed that we also use `contains()` and `contains_arguments()` a lot. Would it make sense to create traits for this?

### DIFF
--- a/boa_interner/src/lib.rs
+++ b/boa_interner/src/lib.rs
@@ -421,15 +421,6 @@ pub trait ToInternedString {
     fn to_interned_string(&self, interner: &Interner) -> String;
 }
 
-// impl<T> ToInternedString for T
-// where
-//     T: std::fmt::Display,
-// {
-//     fn to_interned_string(&self, _interner: &Interner) -> String {
-//         self.to_string()
-//     }
-// }
-
 impl<T> ToInternedString for T
 where
     T: ToIndentedString,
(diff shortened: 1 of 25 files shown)


boa_interner/src/lib.rs
```diff
@@ -409,17 +409,32 @@ impl Interner {
...
 
+// impl<T> ToInternedString for T
+// where
+//     T: std::fmt::Display,
+// {
+//     fn to_interned_string(&self, _interner: &Interner) -> String {
+//         self.to_string()
+//     }
+// }
+
```
raskad, line 432: Some leftover commented out code here.

raskad (approved):
Looks like a great change to me!

Concerning `contains` and `contains_arguments`, I think we should leave those as-is for now. I had the idea of maybe generating the results at parse time and writing them into the AST so that we do not have to walk the AST every time we call those functions. I did not look at it seriously, but If something like that works out, we could remove all of those functions.

RageKnify (approved):
LGTM, nothing to note

Razican:
re-based :)

PR #2376, opened 2022-10-24
[Merged by Bors] - Cleanup and speed-up CI

RageKnify:
Currently we run 7 different jobs:
- tests linux
- tests macos
- tests windows
- rustfmt
- clippy
- examples
- documentation

With this change I reduced them to 4 and hopefully sped them up.

The total execution time is limited by the tests, especially linux that calculates coverage. Having separate jobs for clippy and rustfmt (which take a very small amount of time) is a waste of energy.
With this PR:
Introduced a new cargo profile, `ci`, that should create smaller sized binaries and reduce the our cache usage.
I changed the test runner for macos and windows to [nextest](https://nexte.st/), which should be faster and is specifically designed for CI.
I merged all smaller tasks in a single job, misc, the steps clearly identify what is being tested so it shouldn't affect clarity.
Switched to using the [rust-cache](https://github.com/Swatinem/rust-cache) GH action, this simplifies our work by no longer having to worry about which directories to cache, rust-cache handles all that for us.

~~The bors task should also be modified, I'll get to it as soon as I have time. I believe it should be possible for us to have a single workflow described and have it both be the normal CI and the bors test.~~


Cargo.toml
```diff
@@ -33,6 +33,12 @@ boa_macros = { version = "0.16.0", path = "boa_macros" }
 allow_branch = "main"
 
 # The release profile, used for `cargo build --release`.
```
Razican, line 35: I'm guessing this comment is not correct here, right? The release profile is the `profile.release`, and the `profile.ci` should have a different comment.

Razican (commented):
Looks good, just a small comment typo :)

Razican (approved):
Thanks for the enhancement!

jedel1043 (approved):
Looks good!

jedel1043:
Ah, you forgot to change the `bors.toml` status to the new names.

HalidOdat (approved):

RageKnify:
Yeah, I still want to modify the bors action, I'll get to it today.

PR #2377, opened 2022-10-24
[Merged by Bors] - Document the AST

jedel1043:
As promised in https://github.com/boa-dev/boa/pull/2319#issuecomment-1280095047. There are still some style inconsistencies (which require a bit more time and effort), but having the whole module documented is a lot better for clarity.

### DIFF
--- a/boa_engine/src/syntax/ast/declaration/variable.rs
+++ b/boa_engine/src/syntax/ast/declaration/variable.rs
@@ -12,18 +12,34 @@ use boa_interner::{Interner, ToInternedString};
 
 use super::Declaration;
 
-/// A [`var`][var] declaration list, also called [`VariableDeclarationList`][vardecl]
-/// in the spec.
+/// A [`var`][var] statement, also called [`VariableStatement`][varstmt] in the spec.
 ///
 /// The scope of a variable declared with `var` is its current execution context, which is either
 /// the enclosing function or, for variables declared outside any function, global. If you
 /// re-declare a JavaScript variable, it will not lose its value.
 ///
+/// Although a bit confusing, `VarDeclaration`s are not considered [`Declaration`]s by the spec.
+/// This is partly because it has very different semantics from `let` and `const` declarations, but
+/// also because a `var` statement can be labelled just like any other [`Statement`]:
+///
+/// ```javascript
+/// label: var a = 5;
+/// a;
+/// ```
+///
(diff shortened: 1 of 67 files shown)


boa_engine/src/syntax/ast/declaration/variable.rs
```diff
@@ -155,6 +171,7 @@ impl ToInternedString for VariableList {
     }
 }
 
+/// The error returned by the [`VariableList::try_from`] function.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct TryFromVariableListError(());
```
Razican, line 176: Why does this have an empty tuple inside? (Not from this PR, but it might be easier to do just this, right?
```suggestion
pub struct TryFromVariableListError;
```

boa_engine/src/syntax/ast/expression/spread.rs
```diff
@@ -24,29 +24,32 @@ use super::Expression;
...
 #[derive(Clone, Debug, PartialEq)]
 pub struct Spread {
-    val: Box<Expression>,
+    expression: Box<Expression>,
 }
 
 impl Spread {
+    /// Gets the expression to be expanded by the spread operator.
     #[inline]
     pub fn val(&self) -> &Expression {
```
Razican, line 33: I guess this function should be refactored to `expression()`, right?

boa_engine/src/syntax/ast/expression/spread.rs
```diff
@@ -24,29 +24,32 @@ use super::Expression;
...
-        &self.val
+        &self.expression
     }
 
     /// Creates a `Spread` AST Expression.
     #[inline]
     pub fn new(val: Expression) -> Self {
```
Razican, line 39: I guess here too, the name of the variable should be `expression` or something like that.

Razican (commented):
This is very good work!! Thank you so much.

I have added some comments, let me know what you think :)

boa_engine/src/syntax/ast/declaration/variable.rs
jedel1043: It's just standard Rust practices to disallow creating instances of the error outside of the crate; without this, a simple change like adding a private field would be a breaking change, for example.

HalidOdat (approved):
Looks perfect to me! :)

Razican:
Bors r+

PR #2382, opened 2022-10-26
[Merged by Bors] - Fix var collisions in strict eval calls

jedel1043:
This Pull Request allows collisions of var declarations with already existing lexical bindings if the `eval` call is strict or occurs within strict code. In short, it allows:

```Javascript
{
  let x;
  {
    eval('"use strict"; var x;');
  }
}
```

and

```Javascript
"use strict";
{
  let x;
  {
    eval('var x;');
  }
}
```

This is valid since in strict code all `eval` calls get their own function environment, making it impossible to declare a new var in the outer function environment. This change also skips poisoning environments on strict code, because `eval` cannot add new declarations for the current environment in that situation.

### DIFF
--- a/boa_engine/src/builtins/eval/mod.rs
+++ b/boa_engine/src/builtins/eval/mod.rs
@@ -11,11 +11,13 @@
 
 use crate::{
     builtins::{BuiltIn, JsArgs},
+    environments::DeclarativeEnvironment,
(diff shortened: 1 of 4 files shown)


Razican (approved):
Thank you!

raskad (approved):

PR #2383, opened 2022-10-26
Reduce documentation size in blog

jedel1043:
The documentation page of our blog is a whooping 532 MB in size. This is because we're uploading the whole documentation of all our deps instead of only our crates. This PR modifies our CI to only upload a lightweight version of our documentation, which excludes all deps (replaces all hyperlinks with crates.io links) and only builds our crates docs. This brings the total size of our docs down to 87 MB.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -29,7 +29,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: doc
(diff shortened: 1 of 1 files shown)


HalidOdat (approved):

Razican (approved):

Razican:
Bors r+

PR #2384, opened 2022-10-26
Improve performance of conformance page

jedel1043:
This Pull Request improves the performance of our conformance page.

Basically, we were creating the whole DOM structure for all our test data, which is pretty inefficient considering the size of our data and that only one suite can be displayed at once. This changes our load algorithm so that we append the required elements just before displaying them, and we remove them after hiding them.

Though, this can be further optimized by lazily initializing the DOM elements on display, and also removing them completely, which should be more memory efficient, but I'll leave that for another PR.

### DIFF
--- a/test262/index.html
+++ b/test262/index.html
@@ -98,7 +98,10 @@
           id="old-versions"
         ></section>
       </div>
-      <div class="row accordion col-md-12 collapse" id="info"></div>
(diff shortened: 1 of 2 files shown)


Razican (approved):
Thanks! Nice improvement :)

RageKnify (approved):

PR #2385, opened 2022-10-26
Prune doc from unneeded files

jedel1043:
Follow up of #2383, because this requires a manual prune of the doc directory in order to remove unused files.


jedel1043:
Will merge as it is, since asking for reviews of 5000+ files is unreasonable. If there's any error, our CI will rebuild on the next push to `main`.

PR #2387, opened 2022-10-26
Remove invalid optimization in addition

raskad:

- Remove invalid optimization in addition where `ToPrimitive` calls where skipped.


jedel1043 (approved):
Good catch!

Razican (approved):
Thanks!

PR #2390, opened 2022-10-28
[Merged by Bors] - Implement optional chains

jedel1043:
This Pull Request implements optional chains.

Example:

```Javascript
const adventurer = {
    name: 'Alice',
    cat: {
        name: 'Dinah'
    }
};

console.log(adventurer.cat?.name); // Dinah
console.log(adventurer.dog?.name); // undefined
```

Since I needed to implement `Opcode::RotateLeft`, and #2378 had an implementation for `Opcode::RotateRight`, I took the opportunity to integrate both ops into this PR (big thanks to @HalidOdat for the original implementation!).

This PR almost has 100% conformance for the `optional-chaining` test suite. However, there's this one [test](https://github.com/tc39/test262/blob/85373b4ce12a908f8fc517093d95cf2ed2f5ee6a/test/language/expressions/optional-chaining/member-expression.js) that can't be solved until we properly set function names for function expressions in object and class definitions.


RageKnify (approved):

boa_engine/src/bytecompiler/mod.rs
```diff
@@ -1443,12 +1453,164 @@ impl<'b> ByteCompiler<'b> {
...
+                    self.emit_opcode(Opcode::Pop);
+                }
+            }
             // TODO: try to remove this variant somehow
             Expression::FormalParameterList(_) => unreachable!(),
         }
         Ok(())
     }
 
+    fn compile_access_preserve_this(&mut self, access: &PropertyAccess) -> JsResult<()> {
```
Razican, line 1472: I know that we are missing documentation for most of the existing functions like this one, but would it be possible to add documentations for the new ones at least, so that #1608 doesn't grow further?

boa_engine/src/syntax/ast/expression/optional.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+    /// Gets the kind of optional chain item.
+    pub fn kind(&self) -> &OptionalItemKind {
+        &self.kind
+    }
+
+    /// Returns `true` if the item short-circuits the [`Optional`] chain when `undefined`/`null`.
+    pub fn shorted(&self) -> bool {
+        self.shorted
+    }
```
Razican, line 73: These should also be marked with `#[inline]` probably, since they are one-liners.

boa_engine/src/syntax/ast/expression/optional.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    #[inline]
+    pub(crate) fn contains(&self, symbol: ContainsSymbol) -> bool {
+        self.kind.contains(symbol)
+    }
+}
+
+impl ToInternedString for OptionalItem {
+    #[inline]
```
Razican, line 87: I would not use `#[inline]` here, might be too big of a function to inline, and probably the compiler can decide.

boa_engine/src/syntax/ast/expression/optional.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    /// Gets the target of this `Optional` expression.
+    pub fn target(&self) -> &Expression {
+        self.target.as_ref()
+    }
+
+    /// Gets the chain of accesses and calls that will be applied to the target at runtime.
+    pub fn chain(&self) -> &[OptionalItem] {
+        self.chain.as_ref()
+    }
```
Razican, line 167: I would mark these as `#[inline]`

boa_engine/src/syntax/parser/expression/left_hand_side/mod.rs
```diff
@@ -70,32 +72,50 @@ where
     type Output = Expression;
 
     fn parse(self, cursor: &mut Cursor<R>, interner: &mut Interner) -> ParseResult<Self::Output> {
+        fn is_super_call<R: Read>(
```
Razican, line 75: Is it possible to add some documentation to explain why is this needed and how it detects the super call?

Razican (commented):
Looking very good! Thanks for the implementation :)

boa_engine/src/bytecompiler/mod.rs
jedel1043: Added docs

boa_engine/src/syntax/parser/expression/left_hand_side/mod.rs
jedel1043: Added docs

HalidOdat (approved):
Looks perfect to me :)

PR #2392, opened 2022-10-29
[Merged by Bors] - Implement AST Visitor pattern (attempt #3)

VTCAKAVSMoACE:
This Pull Request closes no specific issue, but allows for analysis and post-processing passes by both internal and external developers.

It changes the following:

- Adds a Visitor trait, to be implemented by visitors of a particular node type.
- Adds `Type`Visitor traits which offer access to private members of a node.
- Adds an example which demonstrates the use of Visitor traits by walking over an AST and printing its contents.

At this time, the PR is more of a demonstration of intent rather than a full PR. Once it's in a satisfactory state, I'll mark it as not a draft.


VTCAKAVSMoACE:
cc: @jedel1043

boa_engine/src/value/mod.rs
```diff
@@ -40,8 +40,8 @@ mod r#type;
 
 pub use conversions::*;
 pub use display::ValueDisplay;
-pub use equality::*;
-pub use hash::*;
+// pub use equality::*;
+// pub use hash::*;
```
VTCAKAVSMoACE, line 44: This was causing a build failure for me locally...
```
error: unreachable `pub` item
  --> boa_engine/src/value/mod.rs:43:9
   |
43 | pub use equality::*;
   | ---     ^^^^^^^^
   | |
   | help: consider restricting its visibility: `pub(crate)`
   |
   = help: or consider exporting it for use by other crates
note: the lint level is defined here
  --> boa_engine/src/lib.rs:41:5
   |
41 |     unreachable_pub,
   |     ^^^^^^^^^^^^^^^
```

Didn't mean to commit it, but easy enough to revert.

VTCAKAVSMoACE:
Okay -- I have updated the style of visitor. Unsure how licensing here needs to work, but seeing as it's only a few lines that are exact or near-exact matches (given that you can't do them in literally any other way), I expect this is not a major concern.

boa_engine/src/syntax/ast/visitor.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// Visit a node.
+    fn visit(&mut self, node: &'a T) -> Result<(), E> {
+        Ok(())
+    }
+
+    /// Visit a node mutably.
+    fn visit_mut(&mut self, node: &'a mut T) -> Result<(), E> {
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
VTCAKAVSMoACE: Completed.

boa_engine/src/syntax/ast/visitor.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn visit_with<'a>(&'a self, visitor: &mut V) -> core::ops::ControlFlow<V::BreakTy>
+    where
+        V: Visitor<'a>;
+
+    /// Visit this node with the provided visitor mutably, allowing the visitor to modify private
+    /// fields.
+    fn visit_with_mut<'a>(&'a mut self, visitor: &mut V) -> core::ops::ControlFlow<V::BreakTy>
+    where
+        V: VisitorMut<'a>;
+}
```
jedel1043, line 83: Almost. The type variable `V` should be part of each of its functions, because there's only one way to implement `VisitWith` on a type. This design makes it possible to implement `VisitWith` several times with different visitors.

VTCAKAVSMoACE: I'm not sure I understand. What do you mean?

jedel1043:
```suggestion
/// Denotes that a type may be visited, providing a method which allows a visitor to traverse its
/// private fields.
pub trait VisitWith {
    /// Visit this node with the provided visitor.
    fn visit_with<'a, V>(&'a self, visitor: &mut V) -> core::ops::ControlFlow<V::BreakTy>
    where
        V: Visitor<'a>;

    /// Visit this node with the provided visitor mutably, allowing the visitor to modify private
    /// fields.
    fn visit_with_mut<'a, V>(&'a mut self, visitor: &mut V) -> core::ops::ControlFlow<V::BreakTy>
    where
        V: VisitorMut<'a>;
}
```

VTCAKAVSMoACE: Aha gotcha. Fixing.

Razican (commented):
I'm happy with the design :) It just needs some documentation on macros and `PrinterVisitor`, and the implementation of the implementers, but I'm happy about its look. I'd like to know the thoughts of others too.

jedel1043 (commented):
The API looks perfect! Missing implementations aside, I think it would be good to have an example on how to implement `Visitor` in the module documentation (e.g. https://docs.rs/serde/latest/serde/de/trait.Visitor.html).

VTCAKAVSMoACE:
There's no obvious way to mutably walk a FormalParameterList and change its properties... I'm going to implement this by simply walking over the parameter list itself for now, but the visibility of `new` or getters/setters may need to be added.

jedel1043:
> There's no obvious way to mutably walk a FormalParameterList and change its properties... I'm going to implement this by simply walking over the parameter list itself for now, but the visibility of `new` or getters/setters may need to be added.

That's okay, since these properties are more like a read-only "cache" of the properties of the `parameters` field. And now that I think about it, we should probably update those fields on the `visit_mut` function, but we can do it on another PR.

boa_examples/Cargo.toml
```diff
@@ -13,5 +13,6 @@ rust-version.workspace = true
 
 [dependencies]
 boa_engine = { workspace = true, features = ["console"] }
+boa_interner = { workspace = true }
```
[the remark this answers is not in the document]
VTCAKAVSMoACE: You can do that? :o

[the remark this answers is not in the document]
VTCAKAVSMoACE: Oh, I suppose it happens in the very next line... Good to know!

jedel1043:
Changes look good! It's just missing some clippy fixes.

boa_engine/src/value/mod.rs
jedel1043: If this didn't break any APIs, I think you can remove this

jedel1043 (approved):
Just remains addressing the commented out `use`s from the `value` module. Everything else looks really good. Very nice job!

jedel1043:
Ah no, I meant deleting the `use`s altogether, since those modules don't export any new types.

Razican (approved):
Everything looks good to me. The only thing here is that we might want to feature-gate this, and not compile the full visitor implementation for all cases, since we are not using it in the engine itself, if I'm not mistaken, at least for now.

Style-wise, we usually prefer `use` statements to be grouped, and in levels, but we have a bit of a mess, and this can be solved by the nightly `cargo fmt`, so that could probably be a separate thing.

In any case, I'm OK to merge this as-is and then add the feature before the 0.17 in a separate PR, for example.

VTCAKAVSMoACE:
I would offer to maybe replace some internal mechanisms. For example: https://github.com/boa-dev/boa/blob/main/boa_engine/src/syntax/ast/statement/mod.rs#L140

jedel1043:
Yeah, I was planning on replacing `contains` and `contains_arguments` with a visitor

VTCAKAVSMoACE:
For sure. Unless I'm mistaken, the code I linked becomes:
```rs
    pub(crate) fn var_declared_names(&self, vars: &mut FxHashSet<Identifier>) {
        struct IdentVisitor<'a> {
            vars: &'a mut FxHashSet<Identifier>,
        }

        impl<'a, 'ast> Visitor<'ast> for IdentVisitor<'a> {
            type BreakTy = Infallible;

            fn visit_identifier(&mut self, node: &'ast Identifier) -> ControlFlow<Self::BreakTy> {
                self.vars.insert(*node);
                ControlFlow::Continue(())
            }
        }

        let mut visitor = IdentVisitor { vars };
        visitor.visit_statement(self);
    }
```

VTCAKAVSMoACE:
Hm, actually, that causes a bunch of errors. I am not sure why.

jedel1043:
> Hm, actually, that causes a bunch of errors. I am not sure why.

Not the correct definition. That includes vars inside functions, which is incorrect since those are part of the var environment of the function. Here's the correct algorithm for `var_declared_names`:

https://tc39.es/ecma262/#sec-static-semantics-vardeclarednames

jedel1043:
Gonna merge this as it is, and I'll create some PRs to take advantage of the new visitor.

bors r+

PR #2394, opened 2022-10-31
[Merged by Bors] - Boa Gc implementation draft

nekevss:
Not sure if anyone else may be working on something more substantial/in-depth, but I thought I'd post this. 😄 

The basic rundown is that this is more of an untested (and in some ways naïve) draft than anything else. It builds rather heavily on `rust-gc`, and tries to keep plenty of the core aspects so as to not break anything too much, and also to minimize overarching changes were it to actually be merged at some point.

This implementation does add ~~a generational divide (although a little unoptimized) to the heap,~~ a GcAlloc/Collector struct with methods, and an ephemeron implementation that allows for the WeakPair and WeakGc pointers.


nekevss:
Commits are a bit of a monstrosity, but the above appears to be roughly working. Looks like there's some EcmaScript tests that I may have to take a look at, but let me know what you think 😄

jedel1043:
Just a small suggestion. Could you remove the merged commits from main, then rebase? It would make reviewing the code a bit easier, since right now it's grouping your changes with all the merged commits from the main branch.

nekevss:
Okay, I believe it's just about all cleaned up

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,575 @@
...
+
+// TODO: Determine if thread local variables are the correct approach vs an initialized structure
+thread_local!(pub static EPHEMERON_QUEUE: StdCell<Option<Vec<GcPointer>>> = StdCell::new(None));
+thread_local!(pub static GC_DROPPING: StdCell<bool> = StdCell::new(false));
+thread_local!(static BOA_GC: StdRefCell<BoaGc> = StdRefCell::new( BoaGc {
+    config: GcConfig::default(),
+    runtime: GcRuntimeData::default(),
+    adult_start: StdCell::new(None),
+    youth_start: StdCell::new(None),
+}));
```
jedel1043, line 41: What do you mean by "initialized structure"? Like, putting everything on a single `thread_local!`?

nekevss: No, probably not the right way to word what I was thinking. I meant a non-thread local struct. Basically, I was thinking something that would have to be initialized by context when I wrote that comment.

jedel1043:
So, I experimented with this PR a bit and it seems like it has a memory leak somewhere?

For example, if I run:

```bash
systemd-run --scope -p MemoryMax=2G --user cargo run --release --bin boa_tester -- run
```

The memory of the `boa_tester` process peaks at 1.9 GB:

image

Compare that to the main branch, which uses 300 MB of memory at max:

image

I'll try to run this build with Miri, maybe I can find the leak.

boa_gc/src/lib.rs
jedel1043: I think we can use thread locals for our std implementation, since they should be more performant on multi-threaded contexts (multiple Gcs can be run at the same time, because JS engines must always run on a single thread). On the other hand, we can use other mechanisms for no_std builds, because a system without an OS has no notion of threads.

raskad:
> So, I experimented with this PR a bit and it seems like it has a memory leak somewhere?

Yeah I think so too. The 262 ci currently exists with code 137 which is probably the job being OOM'd.

nekevss:
Yeah, there's definitely something. I've been looking at it all afternoon. Thought it might have to do with either `GcCellRef` or `BoaGc`'s dropping.

jedel1043:
Just for the record. I ran `systemd-run --scope -p MemoryMax=2G --user cargo +nightly miri test -p boa_engine` after cherry-picking #2412 (since Miri was detecting UB on `JsString`s instead of the garbage collector) and this printed:

```Bash
test builtins::array::tests::array_entries_simple ... error: Undefined Behavior: pointer to alloc1059445 was dereferenced after this allocation got freed
    --> /home/jedel/SDK/src/Rust/boa/boa_gc/src/pointers/gc_ptr.rs:84:18
     |
84   |         unsafe { &*self.inner_ptr() }
     |                  ^^^^^^^^^^^^^^^^^^ pointer to alloc1059445 was dereferenced after this allocation got freed
     |
     = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
     = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
     = note: BACKTRACE:
     = note: inside `boa_gc::Gc::<boa_gc::Cell<object::Object>>::inner` at /home/jedel/SDK/src/Rust/boa/boa_gc/src/pointers/gc_ptr.rs:84:18
note: inside `<boa_gc::Gc<boa_gc::Cell<object::Object>> as boa_gc::Trace>::root` at /home/jedel/SDK/src/Rust/boa/boa_gc/src/pointers/gc_ptr.rs:113:9
    --> /home/jedel/SDK/src/Rust/boa/boa_gc/src/pointers/gc_ptr.rs:113:9
     |
113  |         self.inner().root_inner();
     |         ^^^^^^^^^^^^
note: inside `object::jsobject::_DERIVE_boa_gc_Trace_FOR_JsObject::<impl boa_gc::Trace for object::jsobject::JsObject>::root::mark::<boa_gc::Gc<boa_gc::Cell<object::Object>>>` at boa_engine/src/object/jsobject.rs:30:10
    --> boa_engine/src/object/jsobject.rs:30:10
```
I'll try to come up with a minimized script that replicates the UB.

jedel1043:
Ok, I found a pretty good minimization of the problem. A simple script as:
```Rust
use boa_engine::{
    object::ObjectData,
    prelude::JsObject,
    property::PropertyDescriptor,
};

fn main() {
    let object = JsObject::from_proto_and_data(None, ObjectData::ordinary());

    for i in 0..29 {
        let function = JsObject::from_proto_and_data(None, ObjectData::ordinary());
        object.borrow_mut().insert(
            i,
            PropertyDescriptor::builder()
                .value(function)
                .writable(true)
                .enumerable(false)
                .configurable(true),
        );
    }
}
```
Throws a `malloc(): unsorted double linked list corrupted` error every time. You only have to expose the `Object::insert` method because it is crate only.

It is very interesting, because the program runs fine if you only insert 27 newly created objects. It is after you try to insert the 28th object when the gc corrupts the memory.

jedel1043:
Minimized the problem even further:
```Rust
use boa_gc::BoaAlloc;

fn main() {
    let v = BoaAlloc::new_cell(Vec::new());

    for _ in 1..=259 {
        let cell = BoaAlloc::new_cell([0u8;10]);
        v.borrow_mut().push(cell);
    }
}
```
This throws with `double free or corruption (out)`.

VTCAKAVSMoACE:
I actually see a different error here. Try using MSAN:

```bash
$ RUSTFLAGS="-Zsanitizer=memory" cargo -Zbuild-std run --target=x86_64-unknown-linux-gnu --bin gc_error
```

<details>
<summary>
MSAN crash
</summary>

```
   Compiling boa_examples v0.16.0 (/home/addisoncrump/git/boa/boa_examples)
    Finished dev [unoptimized + debuginfo] target(s) in 0.95s
     Running `/home/addisoncrump/git/boa/target/x86_64-unknown-linux-gnu/debug/gc_error`
==50838==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55af7298c4a4 in boa_gc::gc_box::GcBoxHeader::inc_roots::hc5225928b6f33fbe /home/addisoncrump/git/boa/boa_gc/src/gc_box.rs:56:12
    #1 0x55af729820d1 in boa_gc::gc_box::GcBox$LT$T$GT$::root_inner::hbde44432d819337a /home/addisoncrump/git/boa/boa_gc/src/gc_box.rs:161:9
    #2 0x55af7298952f in _$LT$boa_gc..pointers..gc_ptr..Gc$LT$T$GT$$u20$as$u20$boa_gc..trace..Trace$GT$::root::h5f98dc2014fabd3d /home/addisoncrump/git/boa/boa_gc/src/pointers/gc_ptr.rs:113:9
    #3 0x55af72979351 in _$LT$alloc..vec..Vec$LT$T$GT$$u20$as$u20$boa_gc..trace..Trace$GT$::root::mark::hb682d001a0056c52 /home/addisoncrump/git/boa/boa_gc/src/trace.rs:105:17
    #4 0x55af72977cbd in _$LT$alloc..vec..Vec$LT$T$GT$$u20$as$u20$boa_gc..trace..Trace$GT$::root::h7e0e978d6450e227 /home/addisoncrump/git/boa/boa_gc/src/trace.rs:303:13
    #5 0x55af729853fd in boa_gc::internals::cell::GcCell$LT$T$GT$::try_borrow_mut::h7d83c0d1cefa6ac4 /home/addisoncrump/git/boa/boa_gc/src/internals/cell.rs:114:17
    #6 0x55af72984fa9 in boa_gc::internals::cell::GcCell$LT$T$GT$::borrow_mut::h4fa16a01de33d44d /home/addisoncrump/git/boa/boa_gc/src/internals/cell.rs:66:15
    #7 0x55af729769d2 in gc_error::main::he780f4a299b7f5e4 /home/addisoncrump/git/boa/boa_examples/src/bin/gc_error.rs:8:9
    #8 0x55af72974c8c in core::ops::function::FnOnce::call_once::hdd5a7ec9297d7f13 /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:251:5
    #9 0x55af729842c1 in std::sys_common::backtrace::__rust_begin_short_backtrace::h908c572d2cbdd6fb /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys_common/backtrace.rs:121:18
    #10 0x55af7298bb3a in std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::hacdcc69328a4b425 /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:166:18
    #11 0x55af72afd044 in core::ops::function::impls::_$LT$impl$u20$core..ops..function..FnOnce$LT$A$GT$$u20$for$u20$$RF$F$GT$::call_once::h5e1c5deb26d62cbe /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:286:13
    #12 0x55af72c5d2fc in std::panicking::try::do_call::h4898d48b772e7a9f /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:483:40
    #13 0x55af72c6d249 in __rust_try std.e3c8d372-cgu.7
    #14 0x55af72c5cb8d in std::panicking::try::ha46d4cc94db5d020 /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:447:19
    #15 0x55af72a6790d in std::panic::catch_unwind::he21fc66129e84a2e /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panic.rs:137:14
    #16 0x55af72bbb8fe in std::rt::lang_start_internal::_$u7b$$u7b$closure$u7d$$u7d$::h242c210078bd3b3f /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:148:48
    #17 0x55af72c5d681 in std::panicking::try::do_call::heb0987c2d82e69c1 /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:483:40
    #18 0x55af72c6d249 in __rust_try std.e3c8d372-cgu.7
    #19 0x55af72c5c426 in std::panicking::try::h5a4f8aa5684fbb01 /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:447:19
    #20 0x55af72a6786d in std::panic::catch_unwind::h83016bfb06c049ae /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panic.rs:137:14
    #21 0x55af72bbb34b in std::rt::lang_start_internal::h031a9b51ee0a41ea /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:148:20
    #22 0x55af7298ba55 in std::rt::lang_start::h58633f62c16f3df2 /home/addisoncrump/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:165:17
    #23 0x55af72976c16 in main (/home/addisoncrump/git/boa/target/x86_64-unknown-linux-gnu/debug/gc_error+0x91c16) (BuildId: ca2732652103c4c8458a358058d11d9f34908a3d)
    #24 0x7f94ffbf0d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #25 0x7f94ffbf0e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #26 0x55af7290e214 in _start (/home/addisoncrump/git/boa/target/x86_64-unknown-linux-gnu/debug/gc_error+0x29214) (BuildId: ca2732652103c4c8458a358058d11d9f34908a3d)

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/addisoncrump/git/boa/boa_gc/src/gc_box.rs:56:12 in boa_gc::gc_box::GcBoxHeader::inc_roots::hc5225928b6f33fbe
Exiting
```

</details>

nekevss:
This version does work, but i removed the implementation of the youth heap. The basic mark-sweep wasn't working real well with it (and also kept causing bugs). Let me know what you think.

jedel1043:
Um, the MemorySanitizer is still throwing `==13380==WARNING: MemorySanitizer: use-of-uninitialized-value`. I think I can fix it, and I also want to improve some patterns. Do you mind if I push some changes into your branch?

jedel1043:
Miri runs okay, so I think it should be fine.

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,448 @@
...
+        let mut mark_head = head;
+        while let Some(node) = mark_head.get() {
+            if (*node.as_ptr()).header.is_ephemeron() {
+                ephemeron_queue.push(node);
+            } else if (*node.as_ptr()).header.roots() > 0 {
+                (*node.as_ptr()).trace_inner();
+            } else {
+                finalize.push(node)
+            }
+            mark_head = &(*node.as_ptr()).header.next;
```
jedel1043, line 284: Now that I think about it, the previous design probably didn't work because this only marked the nodes that were children of young nodes. This left unmarked all children of only adult nodes, causing erroneous drops.

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,448 @@
...
+}
+
+pub fn finalizer_safe() -> bool {
+    GC_DROPPING.with(|dropping| !dropping.get())
+}
+
+/// The GcAllocater handles initialization and allocation of garbage collected values.
+///
+/// The allocator can trigger a garbage collection
+pub struct BoaAlloc;
```
jedel1043, line 103: I'm not sure why do we need this structure to construct the pointers. I think following `gc`'s APIs looks better and is easier to use and compose. Maybe I'm overlooking something?

boa_gc/src/pointers/weak_pair.rs
```diff
@@ -0,0 +1,77 @@
+use crate::{
+    finalizer_safe,
+    internals::Ephemeron,
+    trace::{Finalize, Trace},
+    GcBox, EPHEMERON_QUEUE,
+};
+use std::cell::Cell;
+use std::ptr::NonNull;
+
+pub struct WeakPair<K: Trace + ?Sized + 'static, V: Trace + ?Sized + 'static> {
```
jedel1043, line 10: In my opinion, this should be named `Ephemeron` and the current ephemeron should be named `EphemeronBox`. `WeakPair` sounds like a pair of data being managed by a `WeakGc`.

boa_gc/src/lib.rs
nekevss: Yeah, I thought that too so I started running mark on the adult heap too, but even with that the nodes were still dropping. Which now that I type that out makes me think that `trace_inner` was the issue.

boa_gc/src/lib.rs
nekevss: My thought behind this primarily was to take the heap allocation and gc trigger out of `GcBox::new()` heap allocation from the structs.

I might have leaned in a bit too heavily to the idea. Maybe it would be better to call this struct from `Gc::new()` and have a `Gc::new_cell()`.

nekevss:
> Huh, apparently the warning is from the `chrono` crate instead...

Yeah, seems to be windows specific issue, but constant with the current main. The paraphrased panic output is "panicked at 'TzSpecificLocalTimeToSystemTime failed with : The parameter is incorrect' (os error 87)' /chrono-0.4.22/src/offset/local/windows.rs:273:9"

jedel1043:
Did changing the definition of `WeakGc` to an `Ephemeron<T, ()>` not work?

jedel1043:
Ah, I didn't link the PR here, oops 😅
https://github.com/nekevss/boa/pull/123

nekevss:
Sorry I didn't see those earlier! Those changes look great. Although, I'm still not entirely sure what to do with `EphemeronBox` and `GcCell` tho. And maybe that's something to readdress on a different PR?

I may be wrong, but `EphemeronBox` in its current implementation is more of of a container akin to `GcCell`. Furthermore, each container is almost always wrapped in a pointer. In short, would it make sense to implement another pointer called `GcCell`, and the current `GcCell` be renamed to `InternalCell` or `Cell` (maybe not the best struct names). That would clean up the need for `Gc::new(GcCell::new())`, and we'd be able to clean up the folder structure.

- internals
    - containers
        - internal_cell.rs
        - eph_box.rs
    - gc_box.rs
    - mod.rs

I think that better conveys what I was going for when I was adding the `new_cell()` method.

jedel1043:
Just to be sure that `GcCell` is not useful by its own, we should check the common usages of a standalone `RefCell`, since they share the same principle of interior mutability. Maybe there are some usages that we haven't though about.

jedel1043:
> I may be wrong, but EphemeronBox in its current implementation is more of of a container akin to GcCell.

I don't think so. `Cell` has a very strict meaning in the Rust world: A container that allows interior mutability, and `EphemeronBox` doesn't provide such API.

However, I do thing `EphemeronBox` isn't quite right. I'll try to think on more names for it.

nekevss:
> `Cell` has a very strict meaning in the Rust world: A container that allows interior mutability

Agreed. Just to be clear, I didn't mean that `EpemeronBox` is exactly the same as a `Cell`. More so, that in the scope of the garbage collector and it's use in Boa both are currently used as wrappers for the `GcBox`'s value field for vastly different purposes.

I think there is potential for cases where `GcCell` may be used by itself. Maybe it would've been better to describe the new pointer idea as `GcMut` vs. `Gc`.

jedel1043:
> I think there is potential for cases where `GcCell` may be used by itself, but none of those cases currently apply in Boa. Maybe it would've been better to describe the new pointer idea as `GcMut` vs. `Gc`.

I was mentioning that because if we are going to expose `GcCell` as a public API, it would be very confusing for an user to have a `Gc<GcCell<T>>` and a `GcMut<T>` that do the exact same thing. If the objective is to reduce the amount of `Gc::new(GcCell::new(T))`, maybe we could add an utility function to `Gc`, just like before? Though, I tend towards composability instead of hiding the creation of `GcCell`s behind `Gc`.

Having said all of that, it would be better to open a new discussion about the GC API, since the scope of this PR is just to pull the tree of the `gc` crate into Boa and add weak pointers.

boa_gc/src/pointers/ephemeron.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl<K: Trace + ?Sized, V: Trace> Ephemeron<K, V> {
+    pub fn new(key: &Gc<K>, value: V) -> Self {
+        Self {
+            inner_ptr: Cell::new(Allocator::new(GcBox::new_weak(EphemeronBox::new(
+                key, value,
+            )))),
+        }
+    }
+}
```
jedel1043, line 22: Shouldn't this also unroot the `value` of the ephemeron?

nekevss: We don't need to root/unroot an ephemeron, since ephemeron's are essentially viewed as rooted by default (by the weak flag) and are passed to `Collector::mark_ephemeron()`

jedel1043:
With the new tests, I ran Miri again only on the test suite and it returned:

```Bash
Undefined Behavior: pointer to alloc8449793 was dereferenced after this allocation got freed
  --> boa_gc/src/pointers/ephemeron.rs:34:18
   |
34 |         unsafe { &*self.inner_ptr().as_ptr() }
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^ pointer to alloc8449793 was dereferenced after this allocation got freed
   |
```

boa_gc/src/test/weak.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        assert_eq!(*weak.value().expect("Is live currently"), 3);
+        drop(cloned_gc);
+        force_collect();
+        assert_eq!(*weak.value().expect("WeakGc is still live here"), 3);
+
+        drop(gc_value);
+        force_collect();
+
+        assert!(weak.value().is_none())
```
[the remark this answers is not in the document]
jedel1043: Probably. It seems to be dropping ephs that are still live.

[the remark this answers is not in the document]
nekevss: Actually, I'm I was wrong on the above comment regarding roots. If we root the ephemeron by default and unroot it on drop, then we are able to extend the lifetime of the `EphemeronBox` past that of it's key. Changes were made to `mark_ephemeron()` and `sweep()`. Miri seemed okay with all the `gc` focused tests but `gc_basic_pointer_alloc`, which seems to only failing when running the tests with Miri, but passes normally.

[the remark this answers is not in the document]
jedel1043: Ah, that's another thing that I wanted to address. Miri runs all the tests on the same thread, so the collections counter preserves between tests. We can address that by making all tests create a new thread and run the actual test there.

jedel1043:
Ok, I enabled the standard lints for unsafe codebases and isolated every unsafe operation into its own block. What just remains is to document every unsafe block with an explanation on why the operation is safe. I documented some in the `lib.rs` file.

nekevss:
I was planning to go through and add a couple more examples just to cover the different variations of items being collected vs. dumped. I can add some more safety notes as well if you haven't started too.

One more thing, I saw the link to Racket's page on Ephemerons. Do we also want to link to [Hayes' paper](https://dl.acm.org/doi/10.1145/263700.263733) directly?

jedel1043:
> I can add some more safety notes as well if you haven't started too.

Please do! I'm pretty sure you have the most experience with the GC, you'll be able to provide more correct explanations on the safety of some operations.





> Do we also want to link to [Hayes' paper](https://dl.acm.org/doi/10.1145/263700.263733) directly?

Yep! While Racket's explanation is short and to the point, maybe there'll be some folks who'll want a more extensive explanation of it.

boa_gc/src/internals/eph_box.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+// SAFETY: Please see [`Trace]
+unsafe impl<K: Trace + ?Sized, V: Trace + ?Sized> Trace for EphemeronBox<K, V> {
+    /// # Safety
+    ///
+    /// Please see [`Trace`].
+    #[inline]
+    unsafe fn trace(&self) {
```
jedel1043, line 111: I'm pretty sure implementations of unsafe methods of traits don't need a `Safety` section. That section is provided by the trait definition. Also applies to the added documentation.

boa_gc/src/internals/eph_box.rs
```diff
@@ -0,0 +1,160 @@
...
+// and therefore so has the `GcBox` that `key`stores the pointer to, then we set `key`
+// to None to guarantee that we do not access freed memory.
+impl<K: Trace + ?Sized, V: Trace + ?Sized> Finalize for EphemeronBox<K, V> {
+    #[inline]
+    fn finalize(&self) {
+        self.key.set(None);
+    }
+}
+
+// SAFETY: Please see [`Trace]
```
jedel1043, line 105: The safety comments of trait implementations aren't for the users, but for other maintainers. This shouldn't redirect to `Trace`, it should explain why this implementation of the trait is sound.

jedel1043 (changes requested):

boa_gc/src/cell.rs
```diff
@@ -0,0 +1,552 @@
...
+
+impl Display for BorrowMutError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Display::fmt("GcCell<T> already borrowed", f)
+    }
+}
+
+impl<T: Trace + ?Sized> Finalize for GcCell<T> {}
+
+// SAFETY: Please see [`Trace`]. Borrowed
```
jedel1043, line 224: Missed this one

jedel1043:
So, I tried to implement the `WeakRef` builtin and I realized there's no mechanism to upgrade a `WeakGc` into a `Gc`.

nekevss:
> no mechanism to upgrade a `WeakGc` into a `Gc`. Also, should they implement `Clone`, just like `rc::Weak`

Off the top of my head, these shouldn't be overly difficult to implement for both. Although, I am a bit unclear about whether `WeakRef` is consuming the itself on the upgrade or not.

I think It does prompt a question about the API. For instance, the current `Ephemeron::new()` and `WeakGc::new()` take a `&Gc<T>`. This is due to EphemeronBox's `key` needing to point to a `GcBox`, but it also means that the initial `Gc` is still live. We could move this method to a `::from_ref()` method. And have `::new()` take a `T: Trace + ?Sized` that generates a blank `Gc` and returns a `WeakGc` to the blank.

jedel1043:
> I am a bit unclear about whether `WeakGc` is consuming the itself on the upgrade or not.

It shouldn't, just to have a symmetry with `Weak`'s API.


> We could move this method to a `::from_ref()` method. And have `::new()` take a `T: Trace + ?Sized` that generates a blank `Gc` and returns a `WeakGc` to the blank.

Conceptually, I don't think it makes much sense to have `WeakGc` create a new allocation for an immediately dropped `Gc`, it makes much more sense to always require a reference to an already allocated `Gc`.

jedel1043 (approved):
Thank you for the massive amount of work you've put on this contribution! This looks pretty much ready to be merged :)

nekevss:
> Thank you for the massive amount of work you've put on this contribution! This looks pretty much ready to be merged :)

It's no problem! It's been a lot of fun. Thanks for the all help and feedback on this as well!

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    #[inline]
+    pub(crate) fn is_ephemeron(&self) -> bool {
+        self.roots.get() & WEAK_MASK != 0
+    }
+}
+
+// NOTE: [repr(C)] is most likely unneeded here, but will keep it for now
+/// A garbage collected allocation.
+#[repr(C)]
```
Razican, line 82: Can we try this out without the `#[repr(C)]`?

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,155 @@
...
+use std::cell::Cell;
+use std::ptr::{self, NonNull};
+
+// Age and Weak Flags
+const MARK_MASK: usize = 1 << (usize::BITS - 2);
+const WEAK_MASK: usize = 1 << (usize::BITS - 1);
+const ROOTS_MASK: usize = !(MARK_MASK | WEAK_MASK);
+const ROOTS_MAX: usize = ROOTS_MASK;
+
+pub(crate) struct GcBoxHeader {
```
Razican, line 11: Would it be possible to implement `Debug` here?

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,155 @@
...
+/// A garbage collected allocation.
+#[repr(C)]
+pub(crate) struct GcBox<T: Trace + ?Sized + 'static> {
+    pub(crate) header: GcBoxHeader,
+    pub(crate) value: T,
+}
+
+impl<T: Trace> GcBox<T> {
+    pub(crate) fn new(value: T) -> Self {
+        GcBox {
```
Razican, line 90:
```suggestion
        Self {
```
This can in fact be done in many places, and can make renames easier at some point.

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub(crate) fn weak_trace_inner(&self) {
+        // SAFETY: if a `GcBox` has `weak_trace_inner` called, then the inner.
+        // value must have been deemed as reachable.
+        unsafe {
+            self.value.weak_trace();
+        }
+    }
+
+    /// Increases the root count on this `GcBox`.
+    /// Roots prevent the `GcBox` from being destroyed by the garbage collector.
```
Razican, line 136:
```suggestion
    ///
    /// Roots prevent the `GcBox` from being destroyed by the garbage collector.
```

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub(crate) fn unroot_inner(&self) {
+        self.header.dec_roots();
+    }
+
+    /// Returns a reference to the `GcBox`'s value.
+    pub(crate) fn value(&self) -> &T {
+        &self.value
+    }
+
+    pub(crate) fn is_marked(&self) -> bool {
```
Razican, line 152: This function is missing some documentation, and probably an `#[inline]` hint. I'm marking this one, since it's the only one missing the documentation in this file, but many other functions / structures / enums are missing documentation. Also, we should probably add the `#[inline]` hint to one-liners here and elsewhere.

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,370 @@
...
+
+thread_local!(static EPHEMERON_QUEUE: Cell<Option<Vec<GcPointer>>> = Cell::new(None));
+thread_local!(static GC_DROPPING: Cell<bool> = Cell::new(false));
+thread_local!(static BOA_GC: RefCell<BoaGc> = RefCell::new( BoaGc {
+    config: GcConfig::default(),
+    runtime: GcRuntimeData::default(),
+    adult_start: Cell::new(None),
+}));
+
+struct GcConfig {
```
Razican, line 76: This should implement `Debug`, `Clone` and `Copy`.

boa_gc/src/trace.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// Runs [`Finalize::finalize`] on this object and all
+    /// contained subobjects.
+    fn run_finalizer(&self);
+}
+
+/// Utility macro to define an empty implementation of [`Trace`].
+///
+/// Use this for marking types as not containing any `Trace` types.
+#[macro_export]
+macro_rules! empty_trace {
```
Razican, line 75: Why was `unsafe` removed from the macro name? It's unsafe to implement this for some types, right?

Razican (commented):
I only had the chance to review it lightly, and it's looking great :) I added some comments, mostly related to documentation.

Impressive work!

boa_gc/src/trace.rs
nekevss: This was removed when the safety notes were being added. I believe due to redundancy since `empty_trace` would be called in an `unsafe impl`

```
// SAFETY: Provides explanation why it's safe
unsafe impl Trace for DeadEnd {
    // was unsafe_empty_trace
    empty_trace()
}
```

boa_gc/src/internals/gc_box.rs
nekevss: Yeah, we should be fine. There wasn't anything implemented with layout that would need it.

boa_gc/src/internals/gc_box.rs
nekevss: Done! 😄 Let me know what you think. I was debating putting the isize of `next` into the debug as well but decided against doing it at least for now.

boa_gc/src/lib.rs
nekevss: This was added on the most recent commit. Don't know why it's not showing up as outdated or in the files 😕

raskad:
The `gc` dependency is still in the `Cargo.toml` in `boa_examples`. If we drop it from there it will be gone from the `Cargo.lock`.

boa_gc/src/test/weak.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        drop(weak);
+        force_collect();
+
+        assert_eq!(*new_gc, *new_weak.value().expect("weak should be live"));
+        assert_eq!(
+            *init_gc,
+            *new_weak.value().expect("weak_should be live still")
+        );
+    })
```
raskad, line 132:
```suggestion
    });
```

boa_gc/src/test/weak.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        let weak = WeakGc::new(&init_gc);
+
+        let new_gc = weak.upgrade().expect("Weak is still live");
+
+        drop(weak);
+        force_collect();
+
+        assert_eq!(*init_gc, *new_gc);
+    })
```
raskad, line 111:
```suggestion
    });
```

boa_gc/src/test/weak.rs
```diff
@@ -0,0 +1,133 @@
...
+fn eph_basic_alloc_dump_test() {
+    run_test(|| {
+        let gc_value = Gc::new(String::from("gc here"));
+        let _gc_two = Gc::new("hmmm");
+
+        let eph = Ephemeron::new(&gc_value, 4);
+        let _fourth = Gc::new("tail");
+
+        assert_eq!(*eph.key().expect("must be live"), String::from("gc here"));
+    })
```
raskad, line 95:
```suggestion
    });
```

boa_gc/src/test/weak.rs
```diff
@@ -0,0 +1,133 @@
...
+                    .expect("weak is still live"),
+                &String::from("foo")
+            );
+
+            drop(gc_value);
+            force_collect();
+
+            assert!(eph.key().expect("eph is still live").value().is_none());
+        }
+    })
```
raskad, line 82:
```suggestion
    });
```

boa_gc/src/internals/eph_box.rs
```diff
@@ -0,0 +1,142 @@
...
+// to None to guarantee that we do not access freed memory.
+impl<K: Trace + ?Sized, V: Trace + ?Sized> Finalize for EphemeronBox<K, V> {
+    #[inline]
+    fn finalize(&self) {
+        self.key.set(None);
+    }
+}
+
+// SAFETY: EphemeronBox implements primarly two methods of trace `Trace::is_marked_ephemeron`
+// to determine whether the key field is stored and `Trace;:weak_trace` which continues the `Trace::weak_trace()`
```
RageKnify, line 106:
```suggestion
// to determine whether the key field is stored and `Trace::weak_trace` which continues the `Trace::weak_trace()`
```

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,194 @@
...
+        f.debug_struct("GcBoxHeader")
+            .field("Roots", &self.roots())
+            .field("Weak", &self.is_ephemeron())
+            .field("Marked", &self.is_marked())
+            .finish()
+    }
+}
+
+/// A garbage collected allocation.
+pub(crate) struct GcBox<T: Trace + ?Sized + 'static> {
```
Razican, line 111: This can probably derive Debug :)

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,375 @@
...
+impl Default for GcConfig {
+    fn default() -> Self {
+        Self {
+            threshold: 1024,
+            used_space_percentage: 80,
+        }
+    }
+}
+
+#[derive(Default)]
```
Razican, line 94: This can probably also derive Debug, Clone and Copy

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,375 @@
...
+struct GcRuntimeData {
+    collections: usize,
+    bytes_allocated: usize,
+}
+
+struct BoaGc {
```
Razican, line 100: This should also implement Debug

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,375 @@
...
+    clippy::struct_excessive_bools,
+    clippy::doc_markdown,
+    clippy::semicolon_if_nothing_returned,
+    clippy::pedantic
+)]
+#![deny(
+    clippy::all,
+    clippy::cast_lossless,
+    clippy::redundant_closure_for_method_calls,
+    clippy::unnested_or_patterns,
```
raskad, line 18:
```suggestion
    clippy::redundant_closure_for_method_calls,
    clippy::use_self,
    clippy::unnested_or_patterns,
```

Let's add the `use_self` lint to make sure we catch the places that @Razican mentioned in his comment.

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,375 @@
...
+
+impl Drop for BoaGc {
+    fn drop(&mut self) {
+        Collector::dump(self);
+    }
+}
+// Whether or not the thread is currently in the sweep phase of garbage collection.
+// During this phase, attempts to dereference a `Gc<T>` pointer will trigger a panic.
+
+struct DropGuard;
```
Razican, line 114: These comments seem to be together, right? Should they be doc comments related to the struct?

Also, the guard should derive / implement Debug, and maybe Clone/Copy.

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,375 @@
...
+
+impl DropGuard {
+    fn new() -> DropGuard {
```
Razican, line 117: This is missing some docs, with information on the default value. Also, it can return Self, to make it easier to rename.

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,375 @@
...
+
+impl Drop for DropGuard {
+    fn drop(&mut self) {
+        GC_DROPPING.with(|dropping| dropping.set(false));
+    }
+}
+
+/// Returns `true` if it is safe for a type to run [`Finalize::finalize`].
+#[must_use]
+pub fn finalizer_safe() -> bool {
```
Razican, line 131: This should have an `#[inline]` hint.

boa_gc/src/lib.rs
```diff
@@ -1,6 +1,375 @@
...
+    GC_DROPPING.with(|dropping| !dropping.get())
+}
+
+/// The Allocator handles allocation of garbage collected values.
+///
+/// The allocator can trigger a garbage collection.
+struct Allocator;
```
Razican, line 138: This should at least derive/implement Debug, and maybe Clone/Copy.

Razican (commented):
Looking good! Some Debug implementations and docs missing :)

raskad (commented):
I did a technical review and I think I understand most, if not all of the mechanics. Very impressive work!

I have some initial lint comments.

boa_gc/src/cell.rs
```diff
@@ -0,0 +1,556 @@
...
+    }
+}
+
+impl<T: Trace + ?Sized + Ord> Ord for GcCell<T> {
+    #[inline]
+    #[allow(clippy::use_self)]
+    fn cmp(&self, other: &GcCell<T>) -> Ordering {
+        (*self.borrow()).cmp(&*other.borrow())
+    }
+}
```
RageKnify, line 541: Any reason against this?
```suggestion
    fn cmp(&self, other: &Self) -> Ordering {
        (*self.borrow()).cmp(&*other.borrow())
    }
}
```

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+impl<T: Trace + ?Sized> GcBox<T> {
+    /// Returns `true` if the two references refer to the same `GcBox`.
+    #[allow(clippy::use_self)]
+    pub(crate) fn ptr_eq(this: &GcBox<T>, other: &GcBox<T>) -> bool {
+        // Use .header to ignore fat pointer vtables, to work around
+        // https://github.com/rust-lang/rust/issues/46139
+        ptr::eq(&this.header, &other.header)
+    }
```
RageKnify, line 142:
```suggestion
    pub(crate) fn ptr_eq(this: &Self, other: &Self) -> bool {
        // Use .header to ignore fat pointer vtables, to work around
        // https://github.com/rust-lang/rust/issues/46139
        ptr::eq(&this.header, &other.header)
    }
```

boa_gc/src/cell.rs
nekevss: I just thought it might lack a little bit of clarity to go be comparing &self vs &Self

boa_gc/src/pointers/gc.rs
```diff
@@ -0,0 +1,276 @@
...
+        };
+        gc.set_root();
+        gc
+    }
+}
+
+impl<T: Trace + ?Sized> Gc<T> {
+    /// Returns `true` if the two `Gc`s point to the same allocation.
+    #[allow(clippy::use_self)]
+    pub fn ptr_eq(this: &Gc<T>, other: &Gc<T>) -> bool {
```
RageKnify, line 51:
```suggestion
    pub fn ptr_eq(this: &Self, other: &Self) -> bool {
```

boa_gc/src/cell.rs
RageKnify: Looking at the documentation for [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html#tymethod.cmp) this is the normal signature, I think that is why clippy recommends this.

RageKnify (approved):

boa_gc/src/internals/gc_box.rs
```diff
@@ -0,0 +1,195 @@
+use crate::Trace;
+use std::cell::Cell;
+use std::fmt;
+use std::ptr::{self, NonNull};
+
+// Age and Weak Flags
+const MARK_MASK: usize = 1 << (usize::BITS - 2);
+const WEAK_MASK: usize = 1 << (usize::BITS - 1);
+const ROOTS_MASK: usize = !(MARK_MASK | WEAK_MASK);
+const ROOTS_MAX: usize = ROOTS_MASK;
```
Razican, line 10: Why are these `usize` instead of `u8`? Also, have you considered using `bitflags!{}`?

nekevss: These are `usize` because roots is the counter for rooted objects. I think HalidOdat mentioned it above, but at `usize::BITS` that would allow us in worse case to have 2^30 roots available before a panic. If we make this a `u8`, then each header will only be able to have at best 2^8 roots.

I'm not too familiar with `bitflags!{}`. I'll take a look at it. Is there a benefit to using it that you had in mind?

boa_gc/src/cell.rs
```diff
@@ -0,0 +1,555 @@
...
+use std::cell::{Cell, UnsafeCell};
+use std::cmp::Ordering;
+use std::fmt::{self, Debug, Display};
+use std::hash::Hash;
+use std::ops::{Deref, DerefMut};
+
+use crate::trace::{Finalize, Trace};
+
+#[derive(Copy, Clone)]
+pub(crate) struct BorrowFlag(usize);
```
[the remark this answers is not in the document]
nekevss: `BorrowFlag` is also the counter for the amount of borrows a `GcCell` has like `GcBoxHeader`'s roots. The `usize` here allows for the max amount of borrows given the flag implementation, which does limit it a bit more than roots is limited in `GcBoxHeader`, switching to a `u8` would limit the amount of borrows allowed on any `Gc<GcCell>`

HalidOdat (approved):
Besides some small nitpicks, think looks good to me :) Good work @nekevss

boa_gc/src/internals/gc_box.rs
Razican: Nope, I thought this was just a flag, not a counter. bitflags is nice if you want to have binary flags in a u8. This is good :)

Razican (approved):

raskad (approved):

jedel1043:
Gonna merge this then. Again, thank you, @nekevss for all the effort you put into this PR!

bors r+

Manishearth:
FWIW, rust-gc is happy to accept more upstream PRs so that y'all can continue using it. It might be worth maintaining this as a fork that y'all can tailor that y'all occasionally contribute upstream when you think stuff is ready? I do hope the ephemeron stuff can be eventually upstreamed, i know you have a WIP pr)

The main thing is that rust-gc should be tailored in ways that ideally can be still flexibly picked up only if needed, but that shouldn't be too hard.

In the long run, I suspect boa will want a GC that doesn't need to reference count. I recommend checking out https://github.com/asajeffrey/josephine/

PR #2395, opened 2022-10-31
[Merged by Bors] - Implement delete for references

jedel1043:
This Pull Request implements `delete` for variable references:

```Javascript
x = 5;
console.log(x) // 5;
delete x;
console.log(x) // ReferenceError
```

It changes the following:

- Implements delete for references.
- Fixes tests related to deletions of function definitions inside `eval`.
- Implements an op to throw an error on super property deletion.

This puts us at a conformance of 97.98% for the `test/language/expressions/delete` suite. The last 2 failing tests are related to `with` statements ([11.4.1-4.a-5.js](https://github.com/tc39/test262/blob/b5d3192914e9b107ce4f5df4f77e625865d337c7/test/language/expressions/delete/11.4.1-4.a-5.js#L1) and [11.4.1-4.a-6.js](https://github.com/tc39/test262/blob/b5d3192914e9b107ce4f5df4f77e625865d337c7/test/language/expressions/delete/11.4.1-4.a-6.js#L18)).

### DIFF
--- a/boa_engine/src/environments/compile.rs
+++ b/boa_engine/src/environments/compile.rs
@@ -292,7 +292,7 @@ impl Context {
         &mut self,
         name: Identifier,
         function_scope: bool,
-        deletable: bool,
+        configurable: bool,
     ) {
         if !self
             .realm
@@ -316,7 +316,7 @@ impl Context {
                         .value(JsValue::Undefined)
                         .writable(true)
                         .enumerable(true)
-                        .configurable(deletable)
+                        .configurable(configurable)
                         .build(),
                 );
             }
--- a/boa_engine/src/environments/compile.rs
+++ b/boa_engine/src/environments/compile.rs
@@ -288,7 +288,12 @@ impl Context {
     ///
     /// Panics if the global environment is not function scoped.
     #[inline]
-    pub(crate) fn create_mutable_binding(&mut self, name: Identifier, function_scope: bool) {
+    pub(crate) fn create_mutable_binding(
+        &mut self,
+        name: Identifier,
+        function_scope: bool,
+        deletable: bool,
+    ) {
         if !self
             .realm
             .compile_env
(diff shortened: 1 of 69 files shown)


raskad (approved):
Great work, looks good!

boa_engine/src/environments/compile.rs
```diff
@@ -288,7 +288,12 @@ impl Context {
     ///
     /// Panics if the global environment is not function scoped.
     #[inline]
-    pub(crate) fn create_mutable_binding(&mut self, name: Identifier, function_scope: bool) {
+    pub(crate) fn create_mutable_binding(
+        &mut self,
+        name: Identifier,
+        function_scope: bool,
+        deletable: bool,
```
Razican, line 295: This seems to be used only for the `configurable` attribute. Wouldn't it make sense to call it `configurable`, so that it's more clear to understand what it means?

boa_engine/src/vm/opcode/delete/mod.rs
```diff
@@ -61,3 +61,78 @@ impl Operation for DeletePropertyByValue {
         Ok(ShouldExit::False)
     }
 }
+
+/// `DeleteName` implements the Opcode Operation for `Opcode::DeleteName`
+///
+/// Operation:
+///  - Deletes a property by value of an object
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
Razican, line 69: Do we need to implement `PartialEq`, `Eq` and `Hash` here? What benefits does it have? I see we also have it elsewhere, but I find it odd, since we never construct one of these structures.

Razican (commented):
Looks good, thanks! :) I just had a few questions :)

boa_engine/src/vm/opcode/delete/mod.rs
jedel1043: I just copied it from other Ops. I think it would be worth to delete those, because they impact compilation times.

boa_engine/src/environments/compile.rs
jedel1043: Yep :)

Razican (approved):

PR #2400, opened 2022-11-02
[Merged by Bors] - Parser Idempotency Fuzzer

VTCAKAVSMoACE:
This Pull Request offers a fuzzer which is capable of detecting faults in the parser and interner. It does so by ensuring that the parsed AST remains the same between a parsed source and the result of parsing the `to_interned_string` result of the first parsed source.

It changes the following:

- Adds a fuzzer for the parser and interner.

Any issues I raise in association with this fuzzer will link back to this fuzzer.

You may run the fuzzer using the following commands:
```bash
$ cd boa_engine
$ cargo +nightly fuzz run -s none parser-idempotency
```


boa_engine/fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "boa_engine-fuzz"
+version = "0.0.0"
+publish = false
+edition = "2021"
+
```
Razican, line 6: We can probably add the readme here, now that we have it.

boa_engine/fuzz/fuzz_targets/common.rs
```diff
@@ -0,0 +1,71 @@
...
+use boa_interner::Sym;
+use libfuzzer_sys::arbitrary;
+use libfuzzer_sys::arbitrary::{Arbitrary, Unstructured};
+use std::fmt::{Debug, Formatter};
+use std::ops::ControlFlow;
+
+pub struct FuzzData {
+    pub context: Context,
+    pub ast: StatementList,
+}
```
Razican, line 13: Even if the usage of the structure could be considered straightforward, adding some documentation could be useful.

boa_engine/fuzz/fuzz_targets/common.rs
```diff
@@ -0,0 +1,71 @@
...
+
+impl<'a> Arbitrary<'a> for FuzzData {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let mut context = Context::default();
+        let mut syms_available = Vec::with_capacity(8);
+        for c in 'a'..='h' {
+            syms_available.push(context.interner_mut().get_or_intern(&*String::from(c)));
+        }
```
Razican, line 21: Any reason to only have these symbols? Also, I see a `TODO` below requesting arbitrary string literals.

boa_engine/fuzz/fuzz_targets/parser-idempotency.rs
```diff
@@ -0,0 +1,59 @@
...
+mod common;
+
+use crate::common::FuzzData;
+use boa_engine::syntax::Parser;
+use boa_interner::ToInternedString;
+use libfuzzer_sys::fuzz_target;
+use std::error::Error;
+use std::io::Cursor;
+
+fn do_fuzz(mut data: FuzzData) -> Result<(), Box<dyn Error>> {
```
Razican, line 12: Even if the usage of the function could be considered straightforward, adding some documentation could be useful. A link to the libfuzzer documentation or to the cargo-fuzz documentation where this usage is explained would be nice.

Razican:
The merge conflict is just because the `ast` module was moved to its own crate, but the files themselves should be pretty similar.

boa_engine/fuzz/Cargo.toml
VTCAKAVSMoACE: How do you mean?

boa_engine/fuzz/fuzz_targets/common.rs
VTCAKAVSMoACE: Yes -- this creates a fixed pool of symbols to use in the AST. The AST, when generated, just throws random symbol indices in; we need to fix them to well-known (non-keyword) items in the symbol interner. Additionally, we don't want to generate them dynamically from the fuzz data because this can introduce "noise" into our sample (basically, bytes used previously to make symbols now form part of the AST or vice versa, unaligning the AST-generating bytes). This is also the reason for the TODO -- string literals from the arbitrary data could introduce quite a bit of undesirable noise, but not doing so means we don't test string parsing/interning. I decided not to for this PR because it could introduce so much noise that it would render this unusable.

boa_engine/fuzz/Cargo.toml
jedel1043: `readme = "README.md"`

jedel1043: Ah, apparently `cargo doc` automatically picks up a `README.md` on the root of the crate:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field

VTCAKAVSMoACE: Gotcha, thanks. I don't work much with the _publishing_ side of Rust. :)

jedel1043 (commented):
I think it should be better to lift the `fuzz` directory into the project root, then have a subdirectory for every type of fuzzer.

VTCAKAVSMoACE:
> I think it should be better to lift the `fuzz` directory into the project root, then have a subdirectory for every type of fuzzer.

I'm not sure if this is possible for cargo-fuzz, but I'll see. I seem to remember it having some trouble executing from a virtual workspace root.

jedel1043:
You should be able to easily. For example, on [`gfx-extras`](https://github.com/gfx-rs/gfx-extras) they do precisely that :)

VTCAKAVSMoACE:
Yup -- the issue is that you cannot *init* the fuzzer directory in the virtual workspace root, but you can move it later. Interesting.

boa_ast/Cargo.toml
```diff
@@ -13,10 +13,13 @@ rust-version.workspace = true
 [features]
 serde = ["boa_interner/serde", "dep:serde"]
 
+fuzzer-not-safe-for-production = ["arbitrary", "boa_interner/fuzzer-not-safe-for-production", "num-bigint/arbitrary"]
```
jedel1043, line 16: I think we can change this feature to something simpler like `fuzz`. We can clarify on the docs that the fuzzer is not production-ready.

boa_interner/src/sym.rs
```diff
@@ -126,10 +130,18 @@ impl Sym {
...
     pub(super) const fn get(self) -> usize {
         self.value.get()
     }
+
+    /// Returns the internal value of the [`Sym`]
+    #[cfg(feature = "fuzzer-not-safe-for-production")]
+    #[inline]
+    pub const fn get(self) -> usize {
+        self.value.get()
+    }
```
jedel1043, line 144: I'd just make the original `get` public.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "boa_engine-fuzz"
```
jedel1043, line 2:
```suggestion
name = "boa_fuzz"
```

fuzz/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+publish = false
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+
+[dependencies.boa_ast]
```
jedel1043, line 13: This can be simplified to `boa_ast = {path = "../boa_ast", features = ["fuzzer-not-safe-for-production"]}`

jedel1043 (changes requested):
Very nice work!

boa_ast/Cargo.toml
VTCAKAVSMoACE: This is more of a "traditional" feature gate name. Chromium uses something similar ("FUZZER_NOT_SAFE_FOR_PRODUCTION"). I've changed it to fuzz, it may be worthwhile for other folks to use the fuzz features.

VTCAKAVSMoACE:
Perhaps in the future we could put this into CI? It catches problems *very* fast.

boa_ast/Cargo.toml
VTCAKAVSMoACE: Completed.

boa_interner/src/sym.rs
VTCAKAVSMoACE: Completed.

fuzz/Cargo.toml
VTCAKAVSMoACE: Completed.

fuzz/Cargo.toml
VTCAKAVSMoACE: Completed, with similar simplifications for the other crates.

jedel1043:
> Perhaps in the future we could put this into CI? It catches problems _very_ fast.

Yeah! Though, we'll need a proper CI platform for it. We can't run it on every PR because it could throw random errors at any time, so a CI action that runs every hour or so could be good.

jedel1043 (changes requested):
You should rebase this. I recently pushed a changed that extracted the parser from the engine into a crate, and the API of the `Parser` type changed slightly; it now requires only a `&mut Interner` reference to work.

This change should also speedup the fuzzer, because it would skip having to initialize the builtins in order to parse.

VTCAKAVSMoACE:
> > Perhaps in the future we could put this into CI? It catches problems _very_ fast.
> 
> Yeah! Though, we'll need a proper CI platform for it. We can't run it on every PR because it could throw random errors at any time, A CI action that runs it every hour or so would be nice.

I can ask if this would be appropriate for OSS-Fuzz?

jedel1043:
Maybe! We would need to apply for it though. 
cc @jasonwilliams to hear his opinion on this

RageKnify (approved):
Ty for the awesome contribution and your perseverance having had to redo it. :heart:

fuzz/fuzz_targets/common.rs
```diff
@@ -0,0 +1,75 @@
...
+use boa_interner::Sym;
+use libfuzzer_sys::arbitrary;
+use libfuzzer_sys::arbitrary::{Arbitrary, Unstructured};
+use std::fmt::{Debug, Formatter};
+use std::ops::ControlFlow;
+
+/// Context for performing fuzzing. This structure contains both the generated AST as well as the
+/// context used to resolve the symbols therein.
+pub struct FuzzData {
+    pub context: Context,
```
jedel1043, line 15: This should be changed to an `Interner` field. The context is unused in both the parser fuzzer and the VM fuzzer (you can pass to `ContextBuilder` an `Interner`).

jedel1043 (changes requested):

fuzz/fuzz_targets/common.rs
VTCAKAVSMoACE: The VM fuzzer needs the context to track the instructions remaining.

jedel1043: Yeah, and you can create a new `Context` inside `do_fuzz` and pass it the interner of `FuzzData`, IIUC.

VTCAKAVSMoACE: Aha, got it.

jedel1043 (approved):
No more nitpicking 😆 Great job!

jedel1043:
bors+

Issue #2350, opened 2022-10-17
Runtime limits on "bad neighbors".


rhomber:
As discussed on Discord (I'm Lyran Sage). We'd love to see some features added that can prevent untrusted code from preventing the execution of other tasks. It's our goal to be able to run several user provided scripts on the same thread at the same time (probably a runtime per execution). 

To enable this to work and limit a single script from "blocking" the thread we need to have configurable limits on:

1. Loops, to prevent "while(1)" attacks.
2. Recursion or stack depth. Ideally based on stack depth.

Anything else that you guys can think of that could possibly lock up execution? I think the above two covers everything though really?

Thanks so much!

David

VTCAKAVSMoACE:
The VM fuzzer PR (#2401) adds a mechanism for restricting the number of instructions. Could probably be used to complete (1) here.

y21:
Memory limits might also be worth adding to the list (unless this already exists), otherwise one bad actor can OOM the whole process and bring all other scripts with it.

Also, I don't think limiting bytecode instructions is a good/reliable way to guard against lock ups, since this has no effect on native builtins.
For example, RegEx execution happens in native Rust code for the most part, and code abusing catastrophic backtracking will get the engine stuck in native code for a long while:
```rs
use boa_engine::Context;
fn main() {
    let mut cx = Context::builder().instructions_remaining(500).build();
    dbg!(cx.eval("/(a+)+b/.test('a'.repeat(31) + 'c');"));
}
```
This piece of JS code uses 13 instructions until it gets stuck.

There are plenty of other builtins that spend all of their time in native code and given long and complicated strings, can easily lock up the engine. Another example is `BigInt('1'.repeat(1e7))`.

HalidOdat:
If no body is working on `Loops, to prevent "while(1)" attacks.`, I'll try to solve this :)

mattsse:
with #2857 and #2904 there's a few runtime limits

ideally there's also way to limit memory which can be sidestepped with the existing limits

are there plans/ideas for restricting memory?

PR #2401, opened 2022-11-02
[Merged by Bors] - VM Fuzzer

VTCAKAVSMoACE:
This Pull Request offers a basic VM fuzzer which relies on implied oracles (namely, "does it crash or timeout?").

It changes the following:

- Adds an insns_remaining field to Context, denoting the number of instructions remaining to execute (only available when fuzzing)
- Adds a JsNativeError variant, denoting when the number of instructions has been exceeded (only available when fuzzing)
- Adds a VM fuzzer which looks for cases where Boa may crash on an input

This offers no guarantees about correctness, only assertion violations. Depends on #2400.

Any issues I raise in association with this fuzzer will link back to this fuzzer.

You may run the fuzzer using the following commands:
```bash
$ cd boa_engine
$ cargo +nightly fuzz run -s none vm-implied
```

### DIFF
--- a/boa_engine/src/context/mod.rs
+++ b/boa_engine/src/context/mod.rs
@@ -99,7 +99,7 @@ pub struct Context {
 
     /// Number of instructions remaining before a forced exit
     #[cfg(feature = "fuzz")]
-    pub(crate) insns_remaining: usize,
+    pub(crate) instructions_remaining: usize,
 
     pub(crate) vm: Vm,
 
@@ -598,7 +598,7 @@ pub struct ContextBuilder {
     #[cfg(feature = "intl")]
     icu: Option<icu::Icu>,
     #[cfg(feature = "fuzz")]
-    insns_remaining: usize,
+    instructions_remaining: usize,
 }
 
 impl ContextBuilder {
@@ -625,8 +625,8 @@ impl ContextBuilder {
     ///
     /// This function is only available if the `fuzz` feature is enabled.
     #[cfg(feature = "fuzz")]
-    pub fn insns_remaining(mut self, insns_remaining: usize) -> Self {
-        self.insns_remaining = insns_remaining;
+    pub fn instructions_remaining(mut self, instructions_remaining: usize) -> Self {
+        self.instructions_remaining = instructions_remaining;
         self
     }
 
@@ -659,7 +659,7 @@ impl ContextBuilder {
                     .expect("Failed to initialize default icu data.")
             }),
             #[cfg(feature = "fuzz")]
-            insns_remaining: self.insns_remaining,
+            instructions_remaining: self.instructions_remaining,
             promise_job_queue: VecDeque::new(),
         };
 
--- a/boa_engine/src/context/mod.rs
+++ b/boa_engine/src/context/mod.rs
@@ -97,6 +97,10 @@ pub struct Context {
     #[cfg(feature = "intl")]
     icu: icu::Icu,
 
+    /// Number of instructions remaining before a forced exit
+    #[cfg(feature = "fuzz")]
+    pub(crate) insns_remaining: usize,
+
     pub(crate) vm: Vm,
 
     pub(crate) promise_job_queue: VecDeque<JobCallback>,
@@ -593,6 +597,8 @@ pub struct ContextBuilder {
     interner: Option<Interner>,
     #[cfg(feature = "intl")]
     icu: Option<icu::Icu>,
+    #[cfg(feature = "fuzz")]
+    insns_remaining: usize,
 }
 
 impl ContextBuilder {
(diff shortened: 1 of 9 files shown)


jasonwilliams:
It would be useful to have a README.md in the `boa_engine/fuzz` directory explaining what it is and why we have it there if you can.

VTCAKAVSMoACE:
> It would be useful to have a README.md in the `boa_engine/fuzz` directory explaining what it is and why we have it there if you can.

Done -- provided some explanation for both fuzzers.

Razican:
I guess this depends on the parser fuzzer, so let's merge that first :)

boa_engine/src/error.rs
```diff
@@ -747,6 +764,10 @@ pub enum JsNativeErrorKind {
     /// [e_uri]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI
     /// [d_uri]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI
     Uri,
+    /// Error thrown when no instructions remain. Only used in a fuzzing context; not a valid JS
+    /// error variant.
+    #[cfg(feature = "fuzz")]
+    NoInstructionsRemain,
```
jedel1043, line 770: You also need to add this variant to the `builtins::error::ErrorKind` enum, because that's what the `try_native` function uses to detect the error type of an Error object.

fuzz/fuzz_targets/vm-implied.rs
```diff
@@ -3,22 +3,30 @@
...
+                    // most interesting loops will initially cause an infinite loop
+                    // this is okay as long as it is explicitly handled
+                    Corpus::Keep
+                } else {
+                    Corpus::Reject
+                }
+            }
+            Err(_) => Corpus::Reject,
+        },
+    }
```
jedel1043: Could be rewritten using let else:

```Rust
let Err(e) = eval else {
    return Corpus::Keep
};
let Ok(e) = try_native else {
    return Corpus::Reject
};
return if matches {Corpus::Keep} else {Corpus::Reject};
```

jedel1043 (changes requested):
There was an issue asking for a new feature for limits on loops: #2350. `NoInstructionsRemain` could be reused to implement this.

fuzz/fuzz_targets/vm-implied.rs
VTCAKAVSMoACE: I ended up dropping this anyways.

VTCAKAVSMoACE:
> There was an issue asking for a new feature for limits on loops: #2350. `NoInstructionsRemain` could be reused to implement this.

We are on one mind here lol

jedel1043:
Now that I think about it, you don't need to juggle with conversions of the `NoInstructionsRemain` error! The VM bails before you can execute anything that could call `to_opaque`. We just need to panic on `to_opaque` if a user tries to convert a `NoInstructionsRemain` error into an object.

VTCAKAVSMoACE:
Can the error even be manipulated? Not sure how boa makes native errors available to users.

VTCAKAVSMoACE:
> Now that I think about it, you don't need to juggle with conversions of the `NoInstructionsRemain` error! The VM bails before you can execute anything that could call `to_opaque`. We just need to panic inside `to_opaque` if a user tries to convert a `NoInstructionsRemain` error into an object.

This isn't quite true. Consider:
```js
while (true) try {} catch {}
```

The `catch` attempts to convert it into an object between instructions, so we actually do care. I previously panicked on conversion and had a bunch of false positives (including the one above).

jedel1043:
Ah, right, the check is inside `execute_instruction`. Maybe the check should be moved to the while loop inside `run`?

VTCAKAVSMoACE:
That's up to y'all; I don't really know what the preference on execution order is.

jedel1043:
I mean, a `NoInstructionsRemain` error shouldn't be circumventable from JS, for obvious reasons. I think bailing directly makes more sense for that type of error.

VTCAKAVSMoACE:
It already isn't -- it immediately throws another in any catch block.

jedel1043:
Yes, but I think it should directly quit execution instead of trying to execute a `catch` that won't be executed anyways.

boa_engine/src/context/mod.rs
```diff
@@ -97,6 +97,10 @@ pub struct Context {
     #[cfg(feature = "intl")]
     icu: icu::Icu,
 
+    /// Number of instructions remaining before a forced exit
+    #[cfg(feature = "fuzz")]
+    pub(crate) insns_remaining: usize,
```
HalidOdat, line 102:
```suggestion
    pub(crate) instructions_remaining: usize,
```
I would rename this, so its clear what it is :)

HalidOdat (commented):
Besides the small nitpick comment. This looks good to me :)

RageKnify (approved):

boa_engine/src/context/mod.rs
VTCAKAVSMoACE: Completed.

VTCAKAVSMoACE:
That is not what I thought those buttons did, whoops

HalidOdat (approved):

Razican (approved):
Thanks!! LGTM!

Razican:
@jedel1043 could you check if your comments were addressed?

PR #2402, opened 2022-11-03
[Merged by Bors] - Extract the ast to a crate

jedel1043:
This should hopefully improve our compilation times, both from a clean build and from an incremental compilation snapshot.

Next would be the parser, but it imports `Context`, so it'll require a bit more work.

The number of file changes is obviously big, but almost nothing was changed, I just moved everything to another crate and readjusted the imports of the `parser` module. (Though, I did have to change some details, because there were some functions on the ast that returned `ParseError`s, and the tests had to be moved to the parser)


boa_ast/src/expression/operator/binary/mod.rs
```diff
@@ -38,7 +38,8 @@ pub struct Binary {
 
 impl Binary {
     /// Creates a `BinOp` AST Expression.
-    pub(in crate::syntax) fn new(op: BinaryOp, lhs: Expression, rhs: Expression) -> Self {
+    #[must_use]
```
Razican, line 41: Why are we adding these attributes here? I understand this is meant to be used (like any result of any function), but it's not the same as a `Result` for example, which could contain important information of the execution.

If the user wants to warn against this, they have the `#[warn(unused)]` attribute.

Razican (approved):
Looks good! Thanks :)

We can merge it as-is, but we might want to decide what to do with the `#[must_use]` attributes. Do we want them in all getters? Is there a general guideline in the Rust community?

HalidOdat (approved):
Besides the comment by @Razican this looks perfect to me :)

boa_ast/src/expression/operator/binary/mod.rs
jedel1043: They're generally used when a method or function returns something that the user must handle to be at least useful. For example, `Vec::new` is marked as [`must_use`](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#423). False positives generally occur for types with interior mutability. but the ast doesn't contain any types with that property, so it makes sense to mark them all with `must_use`.

jedel1043: Also, if for some reason an user intentionally wants to call a `must_use` function, adding a `let _ =` before it is enough, which is arguably better for readability.

Razican: Hmm yep, I guess this is fine, but then, this should be applied to all getter and constructor methods throughout the code-base. I guess we could enable the clippy lint for this. I'm OK to merge this as-is, but we should probably have an issue or a separate PR to do this throughout the code base.

PR #2403, opened 2022-11-03
[Merged by Bors] - Replace `contains` and friends with visitors

jedel1043:
This Pull Request replaces `contains`, `contains_arguments`, `has_direct_super` and `function_contains_super` with visitors. (~1000 removed lines!)

Also, the new visitor implementation caught a bug where we weren't setting the home object of async functions, generators and async generators for methods of classes, which caused a stack overflow on `super` calls, and I think that's pretty cool!

Next is `var_declared_names`, `lexically_declared_names` and friends, which will be on another PR.

### DIFF
--- a/boa_ast/src/operations.rs
+++ b/boa_ast/src/operations.rs
@@ -1,0 +1,261 @@
+//! Definitions of various **Syntax-Directed Operations** used in the [spec].
+//!
+//! [spec]: https://tc39.es/ecma262/#sec-syntax-directed-operations
+
+use core::ops::ControlFlow;
+
+use boa_interner::Sym;
+
+use crate::{
+    expression::{access::SuperPropertyAccess, Await, Identifier, SuperCall, Yield},
+    function::{
+        ArrowFunction, AsyncFunction, AsyncGenerator, Class, ClassElement, Function, Generator,
+    },
+    property::{MethodDefinition, PropertyDefinition},
+    visitor::{VisitWith, Visitor},
+    Expression,
+};
+
+/// Represents all the possible symbols searched for by the [`Contains`][contains] operation.
+///
+/// [contains]: https://tc39.es/ecma262/#sec-syntax-directed-operations-contains
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum ContainsSymbol {
+    /// A node with the `super` keyword (`super(args)` or `super.prop`).
+    Super,
+    /// A super property access (`super.prop`).
+    SuperProperty,
+    /// A super constructor call (`super(args)`).
+    SuperCall,
+    /// A yield expression (`yield 5`).
+    YieldExpression,
+    /// An await expression (`await 4`).
+    AwaitExpression,
+    /// The new target expression (`new.target`).
+    NewTarget,
+    /// The body of a class definition.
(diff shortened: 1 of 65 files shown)


boa_ast/src/operations.rs
```diff
@@ -0,0 +1,261 @@
...
+/// More information:
+///  - [ECMAScript specification][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-static-semantics-contains
+#[must_use]
+pub fn contains<N>(node: &N, symbol: ContainsSymbol) -> bool
+where
+    N: VisitWith,
+{
+    struct ContainsVisitor(ContainsSymbol);
```
Razican, line 58: I would at least derive `Debug`, `Clone`, and `Copy`. Also, we should add some documentation here. Maybe moving it out of the function could make sense, but this way is good too.

boa_ast/src/operations.rs
```diff
@@ -0,0 +1,261 @@
...
+/// More information:
+///  - [ECMAScript specification][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-static-semantics-containsarguments
+#[must_use]
+pub fn contains_arguments<N>(node: &N) -> bool
+where
+    N: VisitWith,
+{
+    struct ContainsArgsVisitor;
```
Razican, line 195: Here too, we should derive basic functionality and add some documentation.

boa_ast/src/operations.rs
```diff
@@ -0,0 +1,261 @@
...
+            node: &'ast PropertyDefinition,
+        ) -> ControlFlow<Self::BreakTy> {
+            if let PropertyDefinition::MethodDefinition(name, _) = node {
+                name.visit_with(self)
+            } else {
+                node.visit_with(self)
+            }
+        }
+    }
+    match node.visit_with(&mut ContainsArgsVisitor) {
```
Razican, line 244: I would use [`is_break()`](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html#method.is_break) here.

boa_ast/src/statement/mod.rs
```diff
@@ -226,69 +225,20 @@ impl Statement {
         }
     }
 
-    /// Returns true if the node contains a identifier reference named 'arguments'.
+    #[inline]
```
Razican, line 228: We usually prefer to have attributes after comments, since they are part of the definition. Is there a reason to change it here?

Razican (commented):
Looks very good! I added some comments where needed. I also liked the fix with the `leading_ones()` function and the fix to the bug you found.

boa_ast/src/statement/mod.rs
jedel1043: Nope, fixed in the next commit

boa_ast/src/operations.rs
jedel1043: Added the attributes and some docs on what the visitor does.
I'm more inclined on leaving it private to the function. Having both a visitor and a `contains` function in the same module that do the exact same thing could be a bit confusing.

Razican (approved):
Thanks! LGTM!

raskad (approved):
Great stuff. Having all of the logic in one place is really nice.

PR #2406, opened 2022-11-05
[Merged by Bors] - Fix async tests result values

jedel1043:
So, there were some tests that weren't reporting the result of async evaluations correctly. This PR fixes this. It also ignores tests with the `IsHTMLDDA` feature, since we haven't implemented it.

On another note, this also changes the symbols of the test suite to 'F' (failed) and '-' (ignored), which is clearer for colorless terminals.


Razican (approved):
Thanks!

raskad (approved):

PR #2408, opened 2022-11-05
[Merged by Bors] - Rewrite scope analysis operations using visitors

jedel1043:
This PR rewrites all syntax-directed operations that find declared names and variables using visitors.

Hopefully, this should be the last step before finally being able to separate the parser from the engine.

I checked the failing [tests](https://github.com/tc39/test262/blob/85373b4ce12a908f8fc517093d95cf2ed2f5ee6a/test/language/statements/for-await-of/async-gen-decl-dstr-obj-prop-elem-target-yield-expr.js#L49) and they're apparently false positives, since they return `Promise { <rejected> ReferenceError: x is not initialized }` on the main branch.


RageKnify (approved):

jedel1043:
@Razican Applied your code suggestions here. Let me know if I missed something.

Razican (approved):
LGTM :)

PR #2409, opened 2022-11-05
[Merged by Bors] - Extract the parser into a crate

jedel1043:
Depends on #2408.

This finally separates parsing from execution, which should make compilations a bit faster.


boa_parser/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+categories = ["parser-implementations", "compilers"]
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[features]
+# serde = ["boa_interner/serde", "dep:serde"]
```
Razican, line 14: I guess this is no longer needed, right?

boa_parser/src/parser/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        Self(allow)
+    }
+}
+
+/// Boolean representing if the parser should allow a `default` keyword.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct AllowDefault(bool);
+
+impl From<bool> for AllowDefault {
+    fn from(allow: bool) -> Self {
```
Razican, line 96: I would add some `#[inline]` hints here. I know they weren't there before, but could enhance performance a bit.

boa_parser/src/parser/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        Self(allow)
+    }
+}
+
+#[derive(Debug)]
```
Razican, line 101: We should probably add some documentation to the parser.

jedel1043:
@Razican most of your comments are for #2408 😅. I'll apply them there, sorry for the confusion!

boa_ast/src/visitor.rs
```diff
@@ -14,46 +51,11 @@ macro_rules! try_break {
...
-use crate::{StatementList, StatementListItem};
-use boa_interner::Sym;
-
 /// Creates the default visit function implementation for a particular type
 macro_rules! define_visit {
     ($fn_name:ident, $type_name:ident) => {
         #[doc = concat!("Visits a `", stringify!($type_name), "` with this visitor")]
-        #[must_use]
-        fn $fn_name(&mut self, node: &'ast $type_name) -> core::ops::ControlFlow<Self::BreakTy> {
+        fn $fn_name(&mut self, node: &'ast $type_name) -> ControlFlow<Self::BreakTy> {
```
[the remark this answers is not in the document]
jedel1043: Ah, those macros are private to the module, so they don't need fully qualified paths.

boa_parser/src/parser/mod.rs
jedel1043: I'll add some docs in the crate root.

boa_engine/src/builtins/eval/mod.rs
```diff
@@ -68,28 +71,117 @@ impl Eval {
         mut strict: bool,
         context: &mut Context,
     ) -> JsResult<JsValue> {
+        #[derive(Debug, Default)]
+        #[allow(clippy::struct_excessive_bools)]
+        /// Flags used to throw early errors on invalid `eval` calls.
+        struct Flags {
```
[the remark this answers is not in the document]
jedel1043: Yep!

boa_ast/src/visitor.rs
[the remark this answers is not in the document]
Razican: OK, makes sense :)

boa_parser/src/lexer/mod.rs
```diff
@@ -236,7 +240,12 @@ impl<R> Lexer<R> {
                 )),
                 '.' => {
                     if self.cursor.peek()?.map(|c| (b'0'..=b'9').contains(&c)) == Some(true) {
-                        NumberLiteral::new(next_ch as u8).lex(&mut self.cursor, start, interner)
+                        NumberLiteral::new(
+                            next_ch
+                                .try_into()
+                                .expect("an ascii char must not fail to convert"),
+                        )
+                        .lex(&mut self.cursor, start, interner)
```
raskad, line 248: Here we could completeley avoid the conversion by doing `NumberLiteral::new(b'.').lex(&mut self.cursor, start, interner)`

boa_parser/src/lexer/mod.rs
```diff
@@ -272,17 +281,25 @@ impl<R> Lexer<R> {
...
                 }
-                _ if c.is_ascii_digit() => {
-                    NumberLiteral::new(next_ch as u8).lex(&mut self.cursor, start, interner)
-                }
+                _ if c.is_ascii_digit() => NumberLiteral::new(
+                    next_ch
+                        .try_into()
+                        .expect("an ascii char must not fail to convert"),
+                )
+                .lex(&mut self.cursor, start, interner),
```
raskad, line 302: I'm not sure about the `try_into`s here. A panic is for sure better from a safety perspective, but in these cases we can be sure that the conversion is safe. The panics also introduce a cold branch that is possibly bad for optimizations.

The safe solution without any panics would be to write out all the cases and use `b''` literals in each match branch. But I'm not sure how that would impact code generation as we then would have more different match arms.

Just some thoughts I had. I would also be fine merging this and maybe revisiting it later. What do you think?

raskad (approved):
Really nice. Good to get that `eval` code cleaned up too :)

boa_parser/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+boa_interner.workspace = true
+boa_macros.workspace = true
+boa_ast.workspace = true
+boa_profiler.workspace = true
+boa_unicode.workspace = true
+rustc-hash = "1.1.0"
+fast-float = "0.2.0"
+num-traits = "0.2.15"
+bitflags = "1.3.2"
+num-bigint = "0.4.3"
```
HalidOdat, line 23:
```suggestion
num-bigint = "0.4.3"

```
Just a new line at end of file

HalidOdat (approved):
Looks great!

Just a small nitpick :)

boa_parser/src/lexer/mod.rs
jedel1043: Fair enough :) 
I added some allows instead

PR #2410, opened 2022-11-05
[Merged by Bors] - Reduced boilerplate code in the parser

Razican:
This PR adds an `OrAbrupt` trait, with the `or_abrupt()` function. This function is equivalent to the previous `?.ok_or(ParseError::AbruptEnd)`, but it's cleaner. It's implemented for the parser cursor results types.

It also adds an `advance()` function to the parser cursor (which might be possible to optimize further), that just advances the cursor without returning any token. This shows a clearer intent in many places where it's being used.

I also used `ParseResult` in more places, since we were not using it in many places.


raskad (approved):
Great changes!

HalidOdat (approved):
Looks perfect to me!

PR #2411, opened 2022-11-05
[Merged by Bors] - Module parsing

Razican:
I'm creating this draft PR, since I wanted to have some early feedback, and because I though I would have time to finish it last week, but I got caught up with other stuff. Feel free to contribute :)

The main thing here is that I have divided `eval()`, `parse()` and similar functions so that they can decide if they are parsing scripts or modules. Let me know your thoughts.

Then, I was checking the import & export parsing, and I noticed we are using `TokenKind::Identifier` for `IdentifierName`, so I changed that name. An `Identifier` is an `IdentifierName` that isn't a `ReservedWord`. This means we should probably also adapt all `IdentifierReference`, `BindingIdentifier` and so on parsing. I already created an `Identifier` parser.

Something interesting there is that `await` is not a valid `Identifier` if the goal symbol is `Module`, as you can see in the [spec](https://tc39.es/ecma262/#prod-LabelIdentifier), but currently we don't have that information in the `InputElement` enumeration, we only have `Div`, `RegExp` and `TemplateTail`. How could we approach this?


jedel1043:
Rebased. I'll start contributing to this on the next days

jedel1043:
Parsing is done! There are still some missing details like checking `ContainsDuplicateLabels` or `AllPrivateIdentifiersValid` for modules, but those don't block merging this because they're also missing from the `Script` parsing.

Some test will obviously panic because this doesn't implement the runtime part of modules, but it would be better to do that on a follow-up PR since this one is getting big enough already.

Also, I removed the link to the modules issue to avoid Bors closing it.

raskad (approved):
Great work, I'm finde with merging the panics if we can get them fixed before the next release.

nekevss (approved):
Good work on this! 😄 I'd be fine with merging this as is as well, but fixing the panics should definitely be focused.

PR #2422, opened 2022-11-08
[Merged by Bors] - Implement instruction flowgraph generator

HalidOdat:
This PR is a WIP implementation of a vm instruction flowgraph generator

This aims to make the vm easier to debug and understand for both newcomers and experienced devs.

For example if we have the following code:
```js
let i = 0;
while (i < 10) {
    if (i == 3) {
        break;
    }
    i++;
}
```
It generates the following instructions (which is hard to read, especially jumps):
<details>

```

000000    0000    PushZero
000001    0001    DefInitLet                 0000: 'i'
000006    0002    LoopStart
000007    0003    LoopContinue
000008    0004    GetName                    0000: 'i'
000013    0005    PushInt8                   10
000015    0006    LessThan
000016    0007    JumpIfFalse                78
000021    0008    PushDeclarativeEnvironment 0, 1
000030    0009    GetName                    0000: 'i'
000035    0010    PushInt8                   3
000037    0011    Eq
000038    0012    JumpIfFalse                58
000043    0013    PushDeclarativeEnvironment 0, 0
000052    0014    Jump                       78
000057    0015    PopEnvironment
000058    0016    GetName                    0000: 'i'
000063    0017    IncPost
000064    0018    RotateRight                2
000066    0019    SetName                    0000: 'i'
000071    0020    Pop
000072    0021    PopEnvironment
000073    0022    Jump                       7
000078    0023    LoopEnd

Literals:
    <empty>

Bindings:
    0000: i

Functions:
    <empty>
```

</details>

And the flow graph is generated:
flowgraph

The beginning of the function is  marked by the `start` node (in green) and end (in red). In branching the "yes" branch is marked  in green and "no" in red.

~~This only generates in [graphviz format](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) (a widely used format) but it would be nice to also generate to a format that `mermaid.js` can understand and that could be put in articles https://github.com/boa-dev/boa-dev.github.io/issues/26~~

TODO:
  - [x] Generate graphviz format
  - [x] Generate mermaid format
  - [x] Programmatically generate colors push and pop env instructions
  - [x] Display nested functions in sub-sub-graphs.
  - [x] Put under a feature (`"flowgraph"`)
  - [x] Handle try/catch, switch instructions
  - [x] CLI option for configuring direction of flow (by default it is top down)
  - [x] Handle `Throw` instruction (requires keeping track of try blocks)
  - [x] Documentation
  - [x] Prevent node name collisions (functions with the same name)


jedel1043:
Very nice work! This is a huge improvement for debugging. I have just a small nitpick though. Can it be generated vertically instead of horizontally? I think it will read better that way.

HalidOdat:
> Very nice work! This is a huge improvement for debugging. I have just a small nitpick though. Can it be generated vertically instead of horizontally? I think it will read better that way.

Yes, that is configurable (I did it horizontaly because it looked better on github comments) I plan to add an option for that

And there are come other generation modes, where you can better see the loop:
index

HalidOdat:
Hmmm.. looking at the flowgraph it looks like there is a `PopEnvironment` which will never be executed (since nothing "flows" into it). I'm not sure if it is a but on my part or a bug in the bytecompiler, if it is on the bytecompiler this is already proving to be useful :laughing:

jedel1043:
> Hmmm.. looking at the flowgraph it looks like there is a `PopEnvironment` which will never be executed (since nothing "flows" into it). I'm not sure if it is a bug on my part or in the bytecompiler, if it is on the bytecompiler this is already proving to be useful :laughing: 

That's definitely a bug! `JumpIfFalse` is jumping to instruction 0x58, skipping the `PopEnvironment`, and the plain Jump also skips it by jumping to the end of the function.

HalidOdat:
I guess, I'll create an issue about it :)

jedel1043:
What if we coloured pairs of `PushEnvironmnent` and `PopEnvironment` with the same colour, and change the color for each pair? That way we can identify if environments are being handled correctly.

HalidOdat:
Here is a implementation of that, it assumes that the generated the push and pop instructions always are in that order even if they are interleaved (push2, push1, pop2, pop1), which should be the case generaly (if not always), unless there are some wierd jumping back and forth.
Otherwise we need to hook into the bytecompiler, because we lose context after compilation.

Also included the opcode location prefix on all nodes
<details>



</details>

This is all temporary code, will try to create an nice options for generating dot and mermaid.js formats tomorrow :)

Razican (commented):
This is extremely good. Just seeing the graphs I'm starting to understand the VM, which is already a huge step for me 😅

And it's already showing how this can spot problems!

The code needs a bit of documentation, polishing and maybe the color constant could be reviewed to be more configurable, but it's already in pretty good shape.

Thank you!

jasonwilliams:
This is great! And definitely helps with seeing how things work with the VM a lot. I’m very interested in mermaid integration as that’s supported on GitHub and VSCode now so should make collaboration easier

HalidOdat:
I still need to add many things but we can generate graphviz and mermaid diagrams

Since mermaid diagrams are supported by github comments (flipped it horizontally because its to big):

```mermaid
graph LR
    __main___i_0[PushZero]
    __main___i_1[DefInitLet 'i']
    __main___i_6[LoopStart]
    __main___i_7[LoopContinue]
    __main___i_8[GetName 'i']
    __main___i_13[PushInt8 10]
    __main___i_15[LessThan]
    __main___i_16{JumpIfFalse}
    __main___i_21[PushDeclarativeEnvironment 0, 1]
    __main___i_30[GetName 'i']
    __main___i_35[PushInt8 3]
    __main___i_37[Eq]
    __main___i_38{JumpIfFalse}
    __main___i_43[PushDeclarativeEnvironment 0, 0]
    __main___i_52{Jump}
    __main___i_57[PopEnvironment]
    __main___i_58[GetName 'i']
    __main___i_63[IncPost]
    __main___i_64[RotateRight 2]
    __main___i_66[SetName 'i']
    __main___i_71[Pop]
    __main___i_72[PopEnvironment]
    __main___i_73{Jump}
    __main___i_78[LoopEnd]
    __main___i_79{End}
    __main___i_80{Start}
    __main___i_0 -->| | __main___i_1
    __main___i_1 -->| | __main___i_6
    __main___i_6 -->| | __main___i_7
    __main___i_7 -->| | __main___i_8
    __main___i_8 -->| | __main___i_13
    __main___i_13 -->| | __main___i_15
    __main___i_15 -->| | __main___i_16
    __main___i_16 -->| YES| __main___i_78
    __main___i_16 -->| NO| __main___i_21
    __main___i_21 -->| | __main___i_30
    __main___i_30 -->| | __main___i_35
    __main___i_35 -->| | __main___i_37
    __main___i_37 -->| | __main___i_38
    __main___i_38 -->| YES| __main___i_58
    __main___i_38 -->| NO| __main___i_43
    __main___i_43 -->| | __main___i_52
    __main___i_52 -->| | __main___i_78
    __main___i_57 -->| | __main___i_58
    __main___i_58 -->| | __main___i_63
    __main___i_63 -->| | __main___i_64
    __main___i_64 -->| | __main___i_66
    __main___i_66 -->| | __main___i_71
    __main___i_71 -->| | __main___i_72
    __main___i_72 -->| | __main___i_73
    __main___i_73 -->| | __main___i_7
    __main___i_78 -->| | __main___i_79
    __main___i_80 -->| | __main___i_0

```
</details>

boa_engine/src/vm/code_block.rs
```diff
@@ -398,6 +400,355 @@ impl CodeBlock {
     }
 }
 
+impl CodeBlock {
```
Razican, line 403: I would add this in a sub-module, in order to keep files in a reasonable size. It might also make sense to have a feature-gate for it, to avoid compiling it for normal usage.

HalidOdat:
This is almost complete, besides some things that I listed on the initial comment, I'm making this "ready for review" to get some feedback in the meantime.

If you want to test it out, you can use the `--flowgraph` (or `--flowgraph=mermaid` for mermaid format) flag which outputs graphviz format by default, and pipe it to `dot` (from the graphviz package which is installed on most linux distros by default) or use an online editor like: <https://dreampuf.github.io/GraphvizOnline>
```bash
$ cargo run -- --flowgraph test.js | dot -Tpng > test.png
```
You can generate svgs with `-Tsvg`.

<details>
<summary>Example Code</summary>

<pre><code>
let i = 0;
while (i < 10) {
    if (i == 3) {
        break;
    }
    i += inc_fn();
}

function inc_fn() {
      return (function x(){return 1})();
}
</code>
</pre>

Should generate this flowgraph:

</details>

jedel1043 (approved):
Looks great. Nice addition!

jasonwilliams:
We should make sure this is mentioned in the docs https://github.com/boa-dev/boa/blob/main/docs/debugging.md or https://github.com/boa-dev/boa/blob/main/docs/profiling.md

maybe with a screenshot also

boa_cli/src/main.rs
```diff
@@ -136,6 +160,22 @@ enum DumpFormat {
...
+    Mermaid,
+}
+
+#[derive(Debug, Clone, Copy, ValueEnum)]
+enum FlowgraphDirection {
+    TopToBottom,
+    BottomToTop,
+    LeftToRight,
+    RightToLeft,
+}
```
Razican, line 177: I would add some enum-level and variant level documentation here, even if it's pretty much clear what it does.

boa_cli/src/main.rs
```diff
@@ -136,6 +160,22 @@ enum DumpFormat {
     JsonPretty,
 }
 
+#[derive(Debug, Clone, Copy, ValueEnum)]
```
Razican, line 163: I would add some enum-level documentation here. And it might be nice to add links to each format information, and where they can be used, in each variant documentation, or at the enum-level documentation.

boa_cli/src/main.rs
```diff
@@ -184,6 +224,31 @@ where
     Ok(())
 }
 
+fn generate_flowgraph(
+    context: &mut Context,
+    src: &[u8],
+    flowgraph: FlowgraphFormat,
```
Razican, line 230: I would rename this to `format`, and the `flowgraph_direction` to `direction`, just to avoid repetition and make it more clear.

boa_engine/src/vm/graph.rs
```diff
@@ -0,0 +1,873 @@
+use std::mem::size_of;
```
Razican, line 1: This module in general is missing some documentation at the module level and function/enum/struct level, to better understand what is it doing and how.

boa_engine/src/vm/graph.rs
```diff
@@ -0,0 +1,873 @@
...
+            }
+        }
+        for subgraph in &self.subgraphs {
+            subgraph.mermaid_format(result);
+        }
+        result.push_str("  end\n");
+    }
+}
+
+#[derive(Debug)]
```
Razican, line 338: This can probably also derive `Clone`, right?

boa_engine/src/vm/graph.rs
```diff
@@ -0,0 +1,873 @@
...
+
+        for subgraph in &self.subgraphs {
+            subgraph.mermaid_format(&mut result);
+        }
+        result += "\n";
+        result
+    }
+}
+
+impl CodeBlock {
```
Razican, line 400: We might want to separate this `impl` into a different file (or separate the rest into a different file), to keep file size reasonable.

Razican (commented):
Still a lot to review, but it looks great :) I added some comments related to documentation and `Clone` implementations.

HalidOdat:
Added documentation in `debugging.md` file, you can view how it looks here: <https://github.com/boa-dev/boa/blob/60d33337d322b7cf328b9237cf67a172a24d2d20/docs/debugging.md#instruction-flowgraph>

RageKnify (approved):

jasonwilliams (approved):

Razican (approved):
LGTM :)

PR #2423, opened 2022-11-08
[Merged by Bors] - Handle `__proto__` fields in object literals

raskad:

- Handle `__proto__` fields in object literals

### DIFF
--- a/boa_engine/src/vm/opcode/set/prototype.rs
+++ b/boa_engine/src/vm/opcode/set/prototype.rs
@@ -1,0 +1,36 @@
+use crate::{
+    vm::{opcode::Operation, ShouldExit},
+    Context, JsResult,
+};
+
+/// `SetPrototype` implements the Opcode Operation for `Opcode::SetPrototype`
+///
+/// Operation:
+///  - Sets the prototype of an object.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct SetPrototype;
+
+impl Operation for SetPrototype {
+    const NAME: &'static str = "SetPrototype";
+    const INSTRUCTION: &'static str = "INST - SetPrototype";
+
+    fn execute(context: &mut Context) -> JsResult<ShouldExit> {
+        let value = context.vm.pop();
+        let object = context.vm.pop();
+
+        let prototype = if let Some(prototype) = value.as_object() {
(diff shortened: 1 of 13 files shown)


HalidOdat (approved):
Looks good to me! :)

boa_interner/src/sym.rs
```diff
@@ -193,5 +196,6 @@ create_static_strings! {
     "false",
     "async",
     "of",
-    "target"
+    "target",
+    "__proto__"
```
RageKnify, line 200: Can this include a comma at the end of line? Would be nice to avoid having to modify the last line every time a new string is added.

RageKnify (approved):
Just have the question about the comma, everything else looks good

boa_interner/src/sym.rs
raskad: Unfortunately not as this is a macro call. There may be some way to modify the macro to allow it, but I'm not sure about that.

boa_engine/src/vm/opcode/set/prototype.rs
```diff
@@ -0,0 +1,36 @@
...
+        } else if value.is_null() {
+            None
+        } else {
+            return Ok(ShouldExit::False);
+        };
+
+        let object = object.as_object().expect("object is not an object");
+        object
+            .__set_prototype_of__(prototype, context)
+            .expect("cannot fail per spec");
```
Razican, line 32: Here it says that this cannot fail per the spec, but then I'm missing some link to the spec, or the spec text that says so, right?

Razican (approved):
Just a small nitpick. Mostly about documentation of the poco de. We should probably add more references to the spec in general in opcodes.

I'm happy to merge it as is too, though, great work :)

boa_interner/src/sym.rs
jedel1043: Just change the production of the macro to `( $( $s:literal ),+$(,)? )`

jasonwilliams (approved):

boa_engine/src/vm/opcode/set/prototype.rs
raskad: I will open is issue for this as it is currently a problem with all opcodes.

PR #2425, opened 2022-11-08
[Merged by Bors] - Divide byte compiler

### LINKED ISSUE #1808: Divide the byte compiler
Currently, the byte compiler is written in [a file](https://github.com/boa-dev/boa/blob/672eede30d7bcc46a58ae08acdb864678a54d46b/boa/src/bytecompiler.rs) with almost 2,000 lines. Each execution path should be divided in submodules, in order to reduce the size of the file and to make it more maintainable.

e-codes-stuff:
This Pull Request is currently unfinished but will fix/close #1808 after some review and more work

It changes the following:

- Divides byte compiler logic into separate files

I would like some review on the current code I have to know if the patterns I'm using are acceptable for the codebase, if everything looks good I will try to separate more code into different small modules to finish the work here.


raskad:
Thanks for starting to work on this!

The current changes look very much the way I would expect.

RageKnify (approved):
LGTM

e-codes-stuff:
I've moved most of the biggest logic into separated modules and reduced `src/bytecompiler/mod.rs` to about ~40% of its original size. Does this look near completion? I think there's a lot of code which feels too trivial to move into a separate module, there's a few things left that I want to fix like making all references to `ByteCompiler::compile_expr` use the `expression::compile_expr` function (which is currently wrapped by `ByteCompiler::compile_expr` to make the change not require that fix). But anything else is up to reviewer opinion.

nekevss:
This looks pretty good to me. Think it probably just needs a rebase 😄

e-codes-stuff:
I went through `git rebase --interactive` and tried to resolve everything so it can be merged into main, does this seem fine? I'm still not amazing with git so I'm unsure what it should look like in this case.

nekevss:
There test that's failing is related to Rustfmt. You just need to run `cargo fmt` 😄

e-codes-stuff:
Whoops, forgot to run that. Fixed it, anything else?

raskad (approved):
Thanks for woking on this. I have some suggestions for the new code in general, but I'd be happy to merge this as is and do the changes later.

Instead of the `compile*` functions being free functions that take `byte_compiler: &mut ByteCompiler<'_>,` as an argument, I would suggest implementing them on the `ByteCompiler`.

All of the `compile*` functions probably could use some minimal documentation, but we do not have any for that right now, so that may wait.

e-codes-stuff:
Is it possible to implement methods on `ByteCompiler` within submodules? I
was under the impression that wasn't allowed which is why I made them free
functions, I could pretty easily correct that if theres something I'm
missing though

jasonwilliams (approved):
Agree with @raskad about implementing byteCompiler but apart from that LGTM

jedel1043 (approved):

PR #2431, opened 2022-11-11
[Merged by Bors] - Fix some Date tests

jedel1043:
Found some tests on the `Date` builtin that were failing for incorrect length attributes and missing checks.

### DIFF
--- a/boa_engine/src/builtins/date/mod.rs
+++ b/boa_engine/src/builtins/date/mod.rs
@@ -153,6 +153,9 @@ impl Date {
     /// [spec]: https://tc39.es/ecma262/#sec-timeclip
     #[inline]
     pub fn time_clip(time: f64) -> Option<i64> {
+        // 1. If time is not finite, return NaN.
+        // 2. If abs(ℝ(time)) > 8.64 × 1015, return NaN.
+        // 3. Return 𝔽(! ToIntegerOrInfinity(time)).
         if time.is_nan() {
             return None;
         }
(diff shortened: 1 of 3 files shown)


boa_engine/src/builtins/date/mod.rs
```diff
@@ -152,11 +152,13 @@ impl Date {
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-timeclip
     #[inline]
-    pub fn time_clip(time: f64) -> Option<f64> {
-        if time.abs() > 8.64e15 {
-            None
-        } else {
-            Some(time)
+    pub fn time_clip(time: f64) -> Option<i64> {
```
Razican, line 155: I would add inline comments into this function, as we usually do, even if it's 3 statements:
```
If time is not [finite](https://tc39.es/ecma262/#finite), return NaN.
2. If [abs](https://tc39.es/ecma262/#eqn-abs)([ℝ](https://tc39.es/ecma262/#%E2%84%9D)(time)) > 8.64 × 1015, return NaN.
3. Return [𝔽](https://tc39.es/ecma262/#%F0%9D%94%BD)(! [ToIntegerOrInfinity](https://tc39.es/ecma262/#sec-tointegerorinfinity)(time)).
```

Razican (approved):
Looks good :) thank you! We should probably add more inline comments with the statements from the spec, but they were not there before most of the times, so we can merge it as-is.

I have marked one of the functions, but there are more.

raskad (approved):

HalidOdat (approved):
Looks perfect to me :)

PR #2432, opened 2022-11-13
[Merged by Bors] - Switch tarpaulin to llvm engine

RageKnify:
Trying this to see if it makes coverage numbers more accurate

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Run cargo-tarpaulin
         uses: actions-rs/tarpaulin@v0.1
         with:
-          args: --features intl --ignore-tests
+          args: --features intl --ignore-tests --engine llvm
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v3
 


RageKnify:
Both engines seem to mostly suffer from false negatives, this engine seems to considers 10139 more lines. Looking at the codecov.io report the differences are a little weird so I'm not sure how to feel.
Just because of this one reporting on more lines I tend to preffer it, but I want to hear others' opinion on this.
CC: @Razican @raskad @jedel1043 @HalidOdat @jasonwilliams

raskad:
After reading the relevant issue in `tarpaulin`, my understanding is that the long term plan is to use `llvm` instead of `ptrace`. From that point of view it makes sense to me to switch over. I looked trough the report diff a bit and to me it seems like the `llvm` engine is more accurate in the lines it reports a covered.

I am in favor of switching.

jedel1043:
I agree with @raskad. The LLVM backend also has the advantage of having more support for other architectures.

HalidOdat (approved):
Agreed, switching, to me seems like the right choice :)

Razican (approved):
I like the change :) tarpaulin is moving in the good direction here. If we find false positives/negatives, we should report them.

Razican:
Bors r+

PR #2433, opened 2022-11-14
[Merged by Bors] - Restructure lint lists in `boa_ast`

raskad:
This Pull Request restructures the lint deny/warn/allow lists in `boa_ast` and fixes some clippy lints in the crate.
The most relevant change should be in `boa_ast/src/lib.rs`. I went trough all the lints that are available in rustc/rustdoc/clippy with the goal to have most in the deny list, either trough groups or individually. Some clippy lints remain allowed, because they trigger false positives (I fixed them as far as possible).

I'm interested in how everyone feels about denying most lints, as it may impact individual development workflows. If we agree on how we want to structure this, I will apply the restructured lists to all other creates aswell.


jedel1043:
The one lint I'd argue that should be warn is `missing_docs`. It is very tedious having to add documentation for a prototype that could very well be deleted in the next prototype iteration.

The other lints I don't have a problem with, but let's see what the others say :)

HalidOdat (approved):
Yup, I agree with @jedel1043 about `missing_docs` being warn, I't can be quite bothersome.

Besides that this looks good :)

EDIT: Having a CI task that checks this (and other annoying to prototype lints) would be nice, that way we enforce only documented code to be pushed to `main` branch, what do you think?

Razican:
I'm OK to put `missing_docs` as a warning, but then we should forbid warnings in the CI.

boa_ast/src/function/class.rs
```diff
@@ -90,21 +90,23 @@ impl ToIndentedString for Class {
...
             return format!(
                 "class {class_name}{} {{}}",
-                if let Some(sup) = &self.super_ref {
-                    format!(" extends {}", sup.to_interned_string(interner))
-                } else {
-                    String::new()
-                }
+                self.super_ref
+                    .as_ref()
+                    .map_or_else(String::new, |sup| format!(
```
Razican, line 95: This could be a `map_or_default`

boa_ast/src/function/class.rs
```diff
@@ -90,21 +90,23 @@ impl ToIndentedString for Class {
...
         let mut buf = format!(
             "class {class_name}{} {{\n",
-            if let Some(sup) = &self.super_ref {
-                format!("extends {}", sup.to_interned_string(interner))
-            } else {
-                String::new()
-            }
+            self.super_ref
+                .as_ref()
+                .map_or_else(String::new, |sup| format!(
```
Razican, line 106: Here we could use `map_or_default()` too

Razican (commented):
Thanks for the change :) this looks very good. We should probably apply it to all the crates in the engine.

Also, as I mentioned, I would deny warnings in the CI.

RageKnify:
To deny warnings in CI I think we just need to add the `RUSTFLAGS` environment variable.
Can look at rust-analyzer's setup and copy it:
https://github.com/rust-lang/rust-analyzer/blob/45ec315e01dc8dd1146dfeb65f0ef6e5c2efed78/.github/workflows/ci.yaml#L12-L17

I definitely feel that this should be simplified, having 50 lines selecting which lints to allow or deny copied among all our crates isn't very clean. Looking at the [clippy documentation](https://doc.rust-lang.org/stable/clippy/configuration.html) it says that configuring clippy through a `clippy.toml` file is unstable, so I'm not sure that there is a way to have the same config be available to all crates.

boa_ast/src/function/class.rs
raskad: `map_or_default` does not exist for `Option`.

raskad:
I think coppying the lint list is okay for now. I will add `missing_docs` to `RUSTFLAGS` as soon as I am done with all the other crates, as it will be enforced for all crates in the ci.

RageKnify:
I don't believe there is any alternative @jasonwilliams and it's not that they are necessarily going to deprecate it but they say that they might do so in the future.

raskad:
Do we want to merge this as is based on the discussion?

jedel1043 (approved):

Razican (approved):

Razican:
Bors r+

PR #2438, opened 2022-11-14
[Merged by Bors] - Implement the `WeakRef` builtin

jedel1043:
52/60 tests passing. The remaining tests are either features not implemented ([FinalizationRegistry](https://tc39.es/ecma262/multipage/managing-memory.html#sec-finalization-registry-objects)) or features still in development ([symbols-as-weakmap-keys](https://github.com/tc39/proposal-symbols-as-weakmap-keys)).


boa_engine/src/builtins/weak/weak_ref.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// Method [`WeakRef.prototype.deref ( )`][spec].
+    ///
+    /// If the referenced object hasn't been collected, this method promotes a `WeakRef` into a
+    /// proper [`JsObject`], or returns `undefined` otherwise.
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-weakrefderef
+    pub(crate) fn deref(this: &JsValue, _: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        // Let weakRef be the this value.
```
raskad, line 96:
```suggestion
        // 1. Let weakRef be the this value.
```

boa_engine/src/builtins/weak/weak_ref.rs
```diff
@@ -0,0 +1,129 @@
...
+        // 6. Return weakRef.
+        Ok(weak_ref.into())
+    }
+
+    /// Method [`WeakRef.prototype.deref ( )`][spec].
+    ///
+    /// If the referenced object hasn't been collected, this method promotes a `WeakRef` into a
+    /// proper [`JsObject`], or returns `undefined` otherwise.
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-weakrefderef
```
raskad, line 94:
```suggestion
    /// [spec]: https://tc39.es/ecma262/#sec-weak-ref.prototype.deref
```

raskad (approved):
Looks good to me.

I think if we use a forced gc run, we could create a test to make sure that weakrefs are correctly gc'd after a microtask execution, right?

jedel1043:
> Looks good to me.
> 
> I think if we use a forced gc run, we could create a test to make sure that weakrefs are correctly gc'd after a microtask execution, right?

Yep. I think the test262 only tests that objects aren't gc'd within derefs on the same script. To test the other half of the API, we'll need to:
- Create a new `Context`.
- Run a script that uses the API.
- Call `force_collect`.
- Check that the `WeakRef` stored in the previous run returns `undefined` on deref.

I'll add some tests for it.

HalidOdat (approved):
Looks great to me! :)

boa_engine/src/builtins/weak/weak_ref.rs
```diff
@@ -0,0 +1,166 @@
...
+        })?;
+
+        let weak_ref = weak_ref.as_weak_ref().ok_or_else(|| {
+            JsNativeError::typ()
+                .with_message("WeakRef.prototype.deref: expected `this` to be a `WeakRef` object")
+        })?;
+
+        // 3. Return WeakRefDeref(weakRef).
+
+        // `WeakRefDeref`
```
Razican, line 112: I'm OK with this approach, but why does the spec move it out of the definition? Is it used elsewhere? If so, maybe we should use a macro, as we do with promises.

Razican (approved):
Thank you! Just a small question about one piece of code :)

boa_engine/src/builtins/weak/weak_ref.rs
jedel1043: The [spec](https://tc39.es/ecma262/multipage/managing-memory.html#sec-weakrefderef) says:

>This abstract operation is defined separately from WeakRef.prototype.deref strictly to make it possible to succinctly define liveness.

And the [Liveness](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-liveness) section of the spec uses `WeakRefDeref` as:

> For some set of objects S, a hypothetical WeakRef-oblivious execution with respect to S is an execution whereby the abstract operation [WeakRefDeref](https://tc39.es/ecma262/multipage/managing-memory.html#sec-weakrefderef) of a [WeakRef](https://tc39.es/ecma262/multipage/managing-memory.html#sec-weak-ref-constructor) whose referent is an element of S always returns undefined.
>
> At any point during evaluation, a set of objects S is considered live if either of the following conditions is met:
>
> - Any element in S is included in any [agent](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#agent)'s [[KeptAlive]] [List](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-list-and-record-specification-type).
> - There exists a valid future hypothetical WeakRef-oblivious execution with respect to S that observes the Object value of any object in S.

In short, it's just moved out by the spec's requirements.

Razican:
Bors r+

PR #2449, opened 2022-11-20
[Merged by Bors] - Refactor the `Date` builtin

jedel1043:
Just a general cleanup of the `Date` builtin to use slightly better patterns and to fix our warnings about deprecated functions.

About the regressed tests. It seems to be a `chrono` bug, so I opened up an issue (https://github.com/chronotope/chrono/issues/884) for it and they've already opened a PR fixing it (https://github.com/chronotope/chrono/pull/885).

However, while checking out the remaining failing tests, I realized there's a more fundamental limitation with the library. Currently, [`chrono`](https://github.com/chronotope/chrono) specifies:

> Date types are limited in about +/- 262,000 years from the common epoch.

While the [ECMAScript spec](https://tc39.es/ecma262/#sec-time-values-and-time-range) says:

> The smaller range supported by a time value as specified in this section is approximately -273,790 to 273,790 years relative to 1970.

The range allowed by the spec is barely outside of the range supported by `chrono`! This is why the remaining `Date` tests fail.
Seeing that, I would like to ping @djc and @esheppa (the maintainers of `chrono`) to ask if it would be feasible to add a feature, akin to the `large-dates` feature from the `time` crate, that expands the supported range of `chrono`.

EDIT: Filed https://github.com/chronotope/chrono/issues/886

### DIFF
--- a/boa_engine/src/builtins/date/mod.rs
+++ b/boa_engine/src/builtins/date/mod.rs
@@ -34,8 +34,6 @@ macro_rules! some_or_nan {
 
 /// Gets a mutable reference to the inner `Date` object of `val` and stores it on `var`, or returns
 /// a `TypeError` if `val` is not a `Date` object.
-///
-/// [spec]: https://tc39.es/ecma262/#sec-thistimevalue
 macro_rules! get_mut_date {
     (let $var:ident = $val:expr) => {
         let mut $var = $val
@@ -64,14 +62,17 @@ pub(super) fn this_time_value(value: &JsValue) -> JsResult<Option<NaiveDateTime>
 pub struct Date(Option<NaiveDateTime>);
 
 impl Date {
+    #[inline]
+    /// Creates a new `Date`.
     pub(crate) fn new(dt: Option<NaiveDateTime>) -> Self {
         Self(dt)
     }
+
+    /// Converts the `Date` into a `JsValue`, mapping `None` to `NaN` and `Some(datetime)` to
+    /// `JsValue::from(datetime.timestamp_millis())`.
     fn as_value(&self) -> JsValue {
-        match self.0 {
-            Some(dt) => JsValue::from(dt.timestamp_millis()),
-            None => JsValue::from(f64::NAN),
-        }
+        self.0
+            .map_or_else(|| f64::NAN.into(), |dt| dt.timestamp_millis().into())
     }
 }
 
--- a/boa_engine/src/builtins/date/mod.rs
+++ b/boa_engine/src/builtins/date/mod.rs
@@ -1,3 +1,5 @@
+mod utils;
+use utils::{make_date, make_day, make_time, replace_params, time_clip, DateParameters};
 #[cfg(test)]
 mod tests;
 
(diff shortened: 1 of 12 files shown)


djc:
Please file an issue against the chrono project instead of requesting a feature in a PR to your own project.

(Adding such a feature could make sense, I suppose.)

jedel1043:
> Please file an issue against the chrono project instead of requesting a feature in a PR to your own project.
> 
> (Adding such a feature could make sense, I suppose.)

Thanks for the quick response! Just wanted to confirm with you before opening an issue, mainly to avoid polluting the issue tracker with feature requests :)

boa_engine/src/builtins/date/mod.rs
```diff
@@ -12,45 +14,67 @@ use crate::{
...
+    };
 }
 
-macro_rules! check_normal_opt {
-    ($($v:expr),+) => {
-        $(is_zero_or_normal_opt($v.into()) &&)+ true
+/// Gets a mutable reference to the inner `Date` object of `val` and stores it on `var`, or returns
+/// a `TypeError` if `val` is not a `Date` object.
+///
+/// [spec]: https://tc39.es/ecma262/#sec-thistimevalue
```
Razican, line 38: Is this URL relevant here?

boa_engine/src/builtins/date/mod.rs
```diff
@@ -12,45 +14,67 @@ use crate::{
...
+        .and_then(|obj| obj.borrow().as_date().copied())
+        .ok_or_else(|| JsNativeError::typ().with_message("'this' is not a Date"))?
+        .0)
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Date(Option<NaiveDateTime>);
 
+impl Date {
+    pub(crate) fn new(dt: Option<NaiveDateTime>) -> Self {
```
Razican, line 67:
```suggestion
    #[inline]
    pub(crate) fn new(dt: Option<NaiveDateTime>) -> Self {
```

boa_engine/src/builtins/date/mod.rs
```diff
@@ -12,45 +14,67 @@ use crate::{
...
+        Self(dt)
+    }
+    fn as_value(&self) -> JsValue {
```
Razican, line 70:
```suggestion

    fn as_value(&self) -> JsValue {
```

boa_engine/src/builtins/date/mod.rs
```diff
@@ -12,45 +14,67 @@ use crate::{
...
+        match self.0 {
+            Some(dt) => JsValue::from(dt.timestamp_millis()),
+            None => JsValue::from(f64::NAN),
+        }
```
Razican, line 74: I would use a `map_or()` or something like that, to make this a bit more idiomatic / functional

boa_engine/src/vm/mod.rs
```diff
@@ -182,6 +182,7 @@ impl Context {
 
         while self.vm.frame().pc < self.vm.frame().code.code.len() {
             #[cfg(feature = "fuzz")]
+            #[allow(clippy::redundant_else)]
             if self.instructions_remaining == 0 {
                 return Err(JsError::from_native(JsNativeError::no_instructions_remain()));
             } else {
```
Razican, line 188: Can we reformat this code a bit to not need to allow the lint? I see two options:
 - We can invert and use a `!= 0` in the first condition, and `return` in the `else`
 - We can use a explicit scope, and not have the `else`

I'm personally more towards the second one, since it clearly marks which code will only be compiled for the "fuzz" feature.

jedel1043: Done

boa_engine/src/builtins/date/mod.rs
jedel1043: Nope :)

jedel1043:
> ... maybe this needs a minimum supported rust version bump?

We set `rust-version` to 1.65 on the workspace `Cargo.toml`, so we're already targeting the latest rustc.

HalidOdat (approved):
Looks perfect to me !

PR #2453, opened 2022-11-21
Add a hook for custom loggers

chertov:
Allows developers to replace the default logger.


jedel1043:
We discussed this in our triage and we decided to close this because we're planning to reimplement some algorithms to prettify our object display, and we don't want to expose the internals at the moment for that reason.

PR #2460, opened 2022-11-24
[Merged by Bors] - Set function names in object literal methods

raskad:

- Implement `SetFunctionName` opcode based on [`SetFunctionName`](https://tc39.es/ecma262/#sec-setfunctionname)

### DIFF
--- a/boa_engine/src/vm/code_block.rs
+++ b/boa_engine/src/vm/code_block.rs
@@ -177,7 +177,18 @@ impl CodeBlock {
         let opcode: Opcode = self.code[*pc].try_into().expect("invalid opcode");
         *pc += size_of::<Opcode>();
         match opcode {
-            Opcode::RotateLeft | Opcode::RotateRight | Opcode::SetFunctionName => {
+            Opcode::SetFunctionName => {
+                let operand = self.read::<u8>(*pc);
+                *pc += size_of::<u8>();
+                match operand {
+                    0 => "prefix: none",
+                    1 => "prefix: get",
+                    2 => "prefix: set",
+                    _ => unreachable!(),
+                }
+                .to_owned()
+            }
+            Opcode::RotateLeft | Opcode::RotateRight => {
                 let result = self.read::<u8>(*pc).to_string();
                 *pc += size_of::<u8>();
                 result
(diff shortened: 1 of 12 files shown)


boa_engine/src/vm/code_block.rs
```diff
@@ -177,7 +177,7 @@ impl CodeBlock {
         let opcode: Opcode = self.code[*pc].try_into().expect("invalid opcode");
         *pc += size_of::<Opcode>();
         match opcode {
-            Opcode::RotateLeft | Opcode::RotateRight => {
+            Opcode::RotateLeft | Opcode::RotateRight | Opcode::SetFunctionName => {
```
HalidOdat, line 180:
```suggestion
            Opcode::SetFunctionName => {
                  let operand = self.read::<u8>(*pc);
                  *pc += size_of::<u8>();
                  match operand {
                     0 => "prefix: none",
                     1 => "prefix: get",
                     2 => "prefix: set",
                  } 
            }
            Opcode::RotateLeft | Opcode::RotateRight  => {
```

I think it would be good to print the prefix, instead of a number

boa_engine/src/vm/flowgraph/mod.rs
```diff
@@ -45,7 +45,7 @@ impl CodeBlock {
 
             pc += size_of::<Opcode>();
             match opcode {
-                Opcode::RotateLeft | Opcode::RotateRight => {
+                Opcode::RotateLeft | Opcode::RotateRight | Opcode::SetFunctionName => {
```
HalidOdat, line 48: Same here.

Maybe making it a static method of `SetFunctionName` struct to  avoid code duplication.

HalidOdat (approved):
Just some small nitpicks, everything else looks perfect.

boa_parser/src/parser/expression/primary/object_initializer/mod.rs
```diff
@@ -292,8 +292,14 @@ where
 
         //  PropertyName[?Yield, ?Await] : AssignmentExpression[+In, ?Yield, ?Await]
         if cursor.next_if(Punctuator::Colon, interner)?.is_some() {
-            let value = AssignmentExpression::new(None, true, self.allow_yield, self.allow_await)
+            let name = property_name
+                .literal()
+                .filter(|name| name != &Sym::__PROTO__)
```
Razican, line 297: Since `Sym` is `Copy`, I would suggest to either dereference `name`, or use `copied()`:
```suggestion
                .filter(|name| *name != Sym::__PROTO__)
```

Even though I guess it's equivalent, it more clearly shows intent of comparing the symbol value, and not the reference.

Razican (approved):
Looking good! Thanks!

PR #2461, opened 2022-11-25
[Merged by Bors] - Fix `PartialEq` for `JsBigInt` and `f64`

raskad:

- Use the `BigInt::from_f64` function when checking `JsBigInt` and `f64` for eqality.

### DIFF
--- a/boa_engine/src/bigint.rs
+++ b/boa_engine/src/bigint.rs
@@ -453,21 +453,15 @@ impl PartialEq<JsBigInt> for i32 {
 impl PartialEq<f64> for JsBigInt {
     #[inline]
     fn eq(&self, other: &f64) -> bool {
-        if other.fract().is_zero() {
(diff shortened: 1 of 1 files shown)


boa_engine/src/bigint.rs
```diff
@@ -453,21 +453,21 @@ impl PartialEq<JsBigInt> for i32 {
...
-        self.inner.as_ref() == &RawBigInt::from(*other as i64)
     }
 }
 
 impl PartialEq<JsBigInt> for f64 {
     #[inline]
     fn eq(&self, other: &JsBigInt) -> bool {
-        if self.fract() != 0.0 {
-            return false;
+        if self.fract().is_zero() {
```
Razican, line 467: Same here, it could just be an `&&` operation.

Razican (changes requested):
Thanks for the fix! It turns out the operation doesn't need an if/else. Check the comments :)

Razican (approved):
Looks perfect now, thanks!

HalidOdat (approved):

PR #2486, opened 2022-12-16
[Merged by Bors] - Fix rust 1.66.0 lints

raskad:
This Pull Request fixes the new lints introduced in clippy 1.66.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ members = [
 [workspace.package]
 edition = "2021"
 version = "0.16.0"
-rust-version = "1.65"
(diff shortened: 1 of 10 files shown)


jedel1043 (approved):
Nice! Can you also bump the rustc version? I was planning to use a new lib feature on the Intl PR but I think it would be better to have the bump here, since it's more related.

jasonwilliams (approved):

jasonwilliams:
Clippy changes https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-166

PR #2495, opened 2022-12-20
[Merged by Bors] - [profiler] Cache StringId

tunz:
boa_profiler was allocating new strings every time even when they are the same string, which leads to a crash when `start_event` is called too much. This change updates the profiler to use a hash map cache and avoid duplicate string allocations.

### DIFF
--- a/boa_profiler/src/lib.rs
+++ b/boa_profiler/src/lib.rs
@@ -80,9 +80,15 @@
 use std::fmt::{self, Debug};
 
 #[cfg(feature = "profiler")]
-use measureme::{EventId, Profiler as MeasuremeProfiler, TimingGuard};
+use measureme::{EventId, Profiler as MeasuremeProfiler, StringId, TimingGuard};
 #[cfg(feature = "profiler")]
 use once_cell::sync::OnceCell;
+#[cfg(feature = "profiler")]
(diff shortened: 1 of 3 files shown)


jedel1043 (approved):
Looks good! I would replace the `unwrap`s with `expect`s describing why unwrapping is fine though.

boa_profiler/src/lib.rs
```diff
@@ -24,7 +24,7 @@
 //! [boa-web]: https://boa-dev.github.io/
 //! [boa-playground]: https://boa-dev.github.io/boa/playground/
 
-#![cfg_attr(not(test), forbid(clippy::unwrap_used))]
+#![cfg_attr(not(any(test, feature = "profiler")), forbid(clippy::unwrap_used))]
```
raskad, line 27: Let's replace the `unwrap`s with `expects`s, then this change can also be reverted.

raskad (changes requested):
Looks fine to me, but I'd say let's change the `unwrap`s like @jedel1043 suggested to stick to our normal lint rules.

HalidOdat (approved):
Looks good to me! :)

raskad (approved):

PR #2499, opened 2022-12-20
[Merged by Bors] - Redesign native functions and closures API

jedel1043:
This PR is a complete redesign of our current native functions and closures API.

I was a bit dissatisfied with our previous design (even though I created it 😆), because it had a lot of superfluous traits, a forced usage of `Gc<GcCell<T>>` and an overly restrictive `NativeObject` bound. This redesign, on the other hand, simplifies a lot our public API, with a simple `NativeCallable` struct that has several constructors for each type of required native function.

This new design doesn't require wrapping every capture type with `Gc<GcCell<T>>`, relaxes the trait requirement to `Trace + 'static` for captures, can be reused in both `JsObject` functions and (soonish) host defined functions, and is (in my opinion) a bit cleaner than the previous iteration. It also offers an `unsafe` API as an escape hatch for users that want to pass non-Copy closures which don't capture traceable types.

Would ask for bikeshedding about the names though, because I don't know if `NativeCallable` is the most precise name for this. Same about the constructor names; I added the `from` prefix to all of them because it's the "standard" practice, but seeing the API doesn't have any other method aside from `call`, it may be better to just remove the prefix altogether.

Let me know what you think :)


jedel1043:
Will disable the benchmark, since everything looks good.

nekevss (commented):
A lot of this looks good to me. `NativeCallable` makes sense the more I think about it. Maybe, `NativeCallableFn` to be a bit more specific. I'm not entirely sure about it being in the `function` module, and after looking at this, I'm also a bit curious about `FunctionBuilder` too, as we use the term function in a lot of different contexts that it may be ambiguous at times. Maybe `function` -> `native_function`/`native_fn` and `FunctionBuilder` -> `FunctionObjectBuilder`.

jedel1043:
> Maybe `function` -> `native_function`/`native_fn` and `FunctionBuilder` -> `FunctionObjectBuilder`.

Sounds good! Though, if we're changing `function` to `native_function` I'd also change `NativeCallable` to `NativeFunction`, since `NativeCallableFunction` sound a bit redundant.

nekevss (approved):
This looks good to me 😄

raskad (approved):
Nice changes. Great work on the docs too!

PR #2501, opened 2022-12-21
[Merged by Bors] - Use opcode table rather than match

tunz:
`execute_instruction` is heavily used. After decoding an opcode, `match` is used to find a proper `execute` function for the opcode. But, the `match` may not be able to be optimized into a table jump by rust compiler, so it may use multiple branches to find the function. When I tested with a toy program, only `enum -> &'static str` case was optimized to use a table while `enum -> function call` uses multiple branches. ([gotbolt](https://rust.godbolt.org/z/1rzK5vj6f))

This change makes the opcode to use a table explicitly. It improves the benchmark score of Richards by 1-2% (22.8 -> 23.2).

### DIFF
--- a/boa_engine/src/vm/opcode/mod.rs
+++ b/boa_engine/src/vm/opcode/mod.rs
@@ -126,32 +126,32 @@ macro_rules! generate_impl {
                 unsafe { std::mem::transmute(value) }
             }
 
-            const STR_MAP: &[&'static str] = &[
+            const NAMES: &[&'static str] = &[
                 $($Variant::NAME),*
             ];
 
             /// Name of this opcode.
             #[must_use]
             pub const fn as_str(self) -> &'static str {
-                Self::STR_MAP[self as usize]
+                Self::NAMES[self as usize]
             }
 
-            const INSTRUCTION_STR_MAP: &[&'static str] = &[
+            const INSTRUCTIONS: &[&'static str] = &[
                 $($Variant::INSTRUCTION),*
             ];
 
             /// Name of the profiler event for this opcode.
             #[must_use]
             pub const fn as_instruction_str(self) -> &'static str {
-                Self::INSTRUCTION_STR_MAP[self as usize]
+                Self::INSTRUCTIONS[self as usize]
             }
 
-            const EXECUTE_FN_MAP: &[fn(&mut Context) -> JsResult<ShouldExit>] = &[
+            const EXECUTE_FNS: &[fn(&mut Context) -> JsResult<ShouldExit>] = &[
                 $($Variant::execute),*
             ];
 
             pub(super) fn execute(self, context: &mut Context) -> JsResult<ShouldExit> {
-                Self::EXECUTE_FN_MAP[self as usize](context)
+                Self::EXECUTE_FNS[self as usize](context)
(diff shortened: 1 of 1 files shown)


boa_engine/src/vm/opcode/mod.rs
```diff
@@ -126,32 +126,44 @@ macro_rules! generate_impl {
                 unsafe { std::mem::transmute(value) }
             }
 
+            const STR_MAP: [&'static str; 256] = {
+                let mut map = [""; 256];
+                $(
+                    map[Self::$Variant as usize] = $Variant::NAME;
+                )*
+                map
+            };
```
jedel1043, line 135: I think this could be expressed as a static array of static strings, which is easier to maintain. Something like:

```
const STR_MAP: &[&str] = &[
    (<$Variant as Operation>::NAME,)*
];
```
(Not typechecked, so it may throw errors at the start)
Also applies to the other arrays.

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -126,32 +126,44 @@ macro_rules! generate_impl {
...
-                match self {
-                    $(
-                        Self::$Variant => $Variant::INSTRUCTION
-                    ),*
-                }
+                Self::INSTRUCTION_STR_MAP[self as usize]
             }
 
+            const EXECUTE_FN_MAP: [&dyn Fn(&mut Context) -> JsResult<ShouldExit>; 256] = {
+                let mut map: [&dyn Fn(&mut Context) -> JsResult<ShouldExit>; 256]  = [&|_| { unreachable!(); }; 256];
```
jedel1043, line 158: I'm pretty sure all variants only use plain functions, so this could be better expressed with an array of function pointers:
`&[fn(&mut Context) -> JsResult<ShouldExit>]`.

Also applies the previous comment about the array expression.

jedel1043 (changes requested):
Thanks for the contribution! I have some suggestions to improve maintainability and performance.

boa_engine/src/vm/opcode/mod.rs
tunz: Thanks! I was just confused if the order is preserved in the macro. It seems it is.

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -126,32 +126,32 @@ macro_rules! generate_impl {
                 unsafe { std::mem::transmute(value) }
             }
 
+            const STR_MAP: &[&'static str] = &[
```
jedel1043, line 129: Nitpick: The `MAP` suffix is a bit redundant on every variable, can you rename to `NAMES`, `INSTRUCTIONS` and `EXECUTE_FNS`?

jedel1043 (approved):
Looks good! I have just a simple nitpick, but gonna mark this as approved already😁

jasonwilliams:
This is awesome @tunz thanks, will take a look when I have time or in the new year

HalidOdat (approved):

nekevss (approved):

jedel1043:
Rebased to see if that's what's causing the fail

PR #2504, opened 2022-12-22
[Merged by Bors] - Cleanup `Context` APIs

jedel1043:
Just a general cleanup of the APIs of our `Context`.

- Reordered the `pub` and `pub(crate)/fn` methods to have a clear separation between our public and private APIs.
- Removed the call method and added it to `JsValue` instead, which semantically makes a bit more sense.
- Removed the `construct_object` method, and added an utility method `new` to `JsObject` instead.
- Rewrote some patterns I found while rewriting the calls of the removed function.

### DIFF
--- a/boa_engine/src/object/jsobject.rs
+++ b/boa_engine/src/object/jsobject.rs
@@ -35,14 +35,14 @@ pub struct JsObject {
 impl JsObject {
     /// Creates a new ordinary object with its prototype set to the `Object` prototype.
     ///
-    /// In spec jargon, this is equivalent to calling the
-    /// [`OrdinaryObjectCreate(%Object.prototype%)`][call] operation.
+    /// This is equivalent to calling the specification's abstract operation
+    /// [`OrdinaryObjectCreate(%Object.prototype%)`][call].
     ///
     /// [call]: https://tc39.es/ecma262/#sec-ordinaryobjectcreate
     #[inline]
     #[must_use]
-    pub fn new(context: &mut Context) -> Self {
-        Self::from_proto_and_data(
+    pub fn with_object_proto(context: &mut Context) -> Self {
+        Self::with_proto_and_data(
             context.intrinsics().constructors().object().prototype(),
             ObjectData::ordinary(),
         )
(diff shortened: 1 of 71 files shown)


jedel1043:
Oof, the engine didn't like the adjustments I did to the `String` builtin. I'll fix this tomorrow.

boa_engine/src/object/jsobject.rs
```diff
@@ -33,32 +33,50 @@ pub struct JsObject {
...
 
 impl JsObject {
-    /// Create a new `JsObject` from an internal `Object`.
-    fn from_object(object: Object) -> Self {
-        Self {
-            inner: Gc::new(GcCell::new(object)),
-        }
+    /// Creates a new ordinary object with its prototype set to the `Object` prototype.
+    ///
+    /// In spec jargon, this is equivalent to calling the
```
nekevss, line 38: Small suggestion on the wording 😄 

```suggestion
    /// This is the equivalent to calling the specification's abstract operation [`OrdinaryObjectCreate(%Object.prototype%)`][call].
```

*The suggestion is meant more for the entire sentence, not just the single line.

boa_engine/src/object/jsobject.rs
```diff
@@ -33,32 +33,50 @@ pub struct JsObject {
...
+    ///
+    /// In spec jargon, this is equivalent to calling the [`OrdinaryObjectCreate(null)`][call]
+    /// operation.
+    ///
+    /// [call]: https://tc39.es/ecma262/#sec-ordinaryobjectcreate
+    #[inline]
     #[must_use]
-    pub fn empty() -> Self {
-        Self::from_object(Object::default())
+    pub fn with_null_proto() -> Self {
```
nekevss, line 59: Small question here: should the method here be changed to something like `JsObject::create_null` (this is also  `JsObject::new`)? 

The way I'm thinking about this is that there are two operations that are the equivalent of `OrdinaryObjectCreate(%prototype%)`, but the methods' names don't currently link them as similar operations for `JsObject`.

nekevss (commented):
This looks awesome! Had a couple questions/potential suggestions.

boa_engine/src/object/jsobject.rs
jedel1043: Maybe `JsObject::new` -> `JsObject::with_object_proto`, `JsObject::with_null_proto` stays the same, and `JsObject::from_proto_and_data` -> `JsObject::with_proto_and_data`? That way we follow the "standard" Rust naming while also linking the names to the common operation `OrdinaryObjectCreate`.

boa_engine/src/object/jsobject.rs
jedel1043: Like it! Changed in the next commit

boa_engine/src/object/jsobject.rs
nekevss: The first two sound good to me. Not sure about the last change.

I actually think `JsObject::from_proto_and_data` makes sense, primarily because the `with_` clauses are mainly just creating an ordinary object data with the designated prototype. I believe `from_proto_and_data` is used heavily internally in connection with `OrdinaryObjectCrate`, but it also serves as a good API method for a `JsObject` to be created with the provided prototype and data.

jedel1043: Agreed. `with_` functions which are just utility wrappers around `from_proto_and_data` sounds better.

nekevss (approved):
Looks good to me 😄

HalidOdat (approved):
Looks good to me! :)

PR #2507, opened 2022-12-26
[Merged by Bors] - Fix unary operations on `this`

### LINKED ISSUE #2416: Panic when using unary operations on `this`
**Describe the bug**
Using a unary operator on `this` causes boa to panic.

**To Reproduce**
Execute the following script:

```js
++this;
```

Which causes the following error:

```
thread 'main' panicked at 'not yet implemented: access_set `this`', boa_engine/src/bytecompiler/mod.rs:828:29
```

**Expected behavior**
Boa should issue a syntax error with an invalid operand.

**Additional context**
Discovered by #2401

veera-sivarajan:
This Pull Request fixes/closes #2416.

Previously, prefix increment and decrement operations on `this` caused a panic. This PR makes the parser issue a syntax error when the operand UnaryExpression is not simple (as mentioned in https://tc39.es/ecma262/#sec-update-expressions-static-semantics-early-errors).

### DIFF
--- a/boa_parser/src/parser/expression/update.rs
+++ b/boa_parser/src/parser/expression/update.rs
@@ -56,7 +56,8 @@ impl UpdateExpression {
     }
 }
 
-// https://tc39.es/ecma262/multipage/syntax-directed-operations.html#sec-static-semantics-assignmenttargettype
+/// <https://tc39.es/ecma262/multipage/syntax-directed-operations.html#sec-static-semantics-assignmenttargettype>
+/// This function checks if the target type is simple
 fn is_simple(expr: &Expression, position: Position, strict: bool) -> ParseResult<bool> {
     match expr {
         Expression::Identifier(ident) => {
--- a/boa_parser/src/parser/expression/update.rs
+++ b/boa_parser/src/parser/expression/update.rs
@@ -21,7 +21,7 @@ use boa_ast::{
         operator::{unary::UnaryOp, Unary},
         Identifier,
     },
-    Expression, Punctuator,
+    Expression, Position, Punctuator,
 };
 use boa_interner::Interner;
 use boa_profiler::Profiler;
@@ -56,6 +56,21 @@ impl UpdateExpression {
     }
 }
 
+#[inline]
+fn is_simple(expr: &Expression, position: Position, strict: bool) -> ParseResult<bool> {
+    match expr {
+        Expression::Identifier(ident) => {
+            if strict {
+                check_strict_arguments_or_eval(*ident, position)?;
+            }
+            Ok(true)
+        }
+        Expression::PropertyAccess(_) => Ok(true),
+        _ => Ok(false),
+    }
+}
+
+// 74.54
 impl<R> TokenParser<R> for UpdateExpression
 where
     R: Read,
@@ -72,22 +87,11 @@ where
                 cursor
                     .next(interner)?
                     .expect("Punctuator::Inc token disappeared");
-                
+
                 let target = UnaryExpression::new(self.name, self.allow_yield, self.allow_await)
                     .parse(cursor, interner)?;
-                let strict = cursor.strict_mode();
                 // https://tc39.es/ecma262/#sec-update-expressions-static-semantics-early-errors
-                let simple = match &target {
-                    Expression::Identifier(_) if !strict => true,
-                    Expression::Identifier(ident) => {
-                        check_strict_arguments_or_eval(*ident, position)?;
-                        true
-                    }
-                    Expression::PropertyAccess(_) => true,
-                    _ => false,
-                };
-                
-                if !simple {
+                if !is_simple(&target, position, cursor.strict_mode())? {
                     return Err(Error::lex(LexError::Syntax(
                         "Invalid left-hand side in assignment".into(),
                         position,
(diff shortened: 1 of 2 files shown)


boa_parser/src/parser/expression/update.rs
```diff
@@ -72,11 +72,28 @@ where
...
                     .expect("Punctuator::Inc token disappeared");
-
+                
                 let target = UnaryExpression::new(self.name, self.allow_yield, self.allow_await)
                     .parse(cursor, interner)?;
-
-                if cursor.strict_mode() {
+                let strict = cursor.strict_mode();
+                // https://tc39.es/ecma262/#sec-update-expressions-static-semantics-early-errors
+                let ok = match &target {
```
jedel1043, line 80:
```suggestion
                let simple = match &target {
```

boa_parser/src/parser/expression/update.rs
```diff
@@ -72,11 +72,28 @@ where
...
+                    Expression::Identifier(_) if !strict => true,
+                    Expression::Identifier(ident)
+                        if ![Sym::EVAL, Sym::ARGUMENTS].contains(&ident.sym()) =>
+                    {
+                        true
+                    }
```
jedel1043, line 86: I'm pretty sure the eval/arguments check is already done below. Can we move the check above to the `match` in order to skip rechecking this? Maybe something like:

```Rust
Expression::Identifier(ident) =>
                    {
                        check_strict_arguments_or_eval(ident, position)?;
                        true
                    }
```

jedel1043 (commented):
Thank you for the fix! I have some improvements for it

boa_parser/src/parser/expression/update.rs
veera-sivarajan: Sure, I've moved the check inside `match` for both UnaryExpression and LeftHandSideExpression to make it uniform.

boa_parser/src/parser/expression/update.rs
```diff
@@ -72,14 +72,26 @@ where
...
                 let target = UnaryExpression::new(self.name, self.allow_yield, self.allow_await)
                     .parse(cursor, interner)?;
-
-                if cursor.strict_mode() {
-                    if let Expression::Identifier(ident) = target {
-                        check_strict_arguments_or_eval(ident, position)?;
+                let strict = cursor.strict_mode();
+                // https://tc39.es/ecma262/#sec-update-expressions-static-semantics-early-errors
+                let simple = match &target {
+                    Expression::Identifier(_) if !strict => true,
```
jedel1043, line 81: You can remove this, since the check already filters out the strict case

jedel1043 (changes requested):

boa_parser/src/parser/expression/update.rs
veera-sivarajan: Ah I'm sorry, I'm not sure if I understand this clearly. As far as I can tell, `check_strict_arguments_or_eval()`  does a simple match on the identifier's `Sym` and does not check for strict case. Also, the compiler fails a bunch of EcmaScript test cases when I remove this check.

jedel1043: Ah, didn't see that the check was not what I had in mind. What I was talking about was doing something like:

```Rust
// ... match above
if strict {
    check_strict_arguments_or_eval(..args)?;
}
true
```

veera-sivarajan: Thanks for the suggestion. Removed it and wrapped the check for simple expression into an inline function to not repeat the same code in four different places.

boa_parser/src/parser/expression/update.rs
```diff
@@ -56,6 +56,21 @@ impl UpdateExpression {
     }
 }
 
+#[inline]
```
jedel1043, line 59:
```suggestion
```

The compiler ignores the `#[inline]` annotation on functions that aren't public.

jedel1043 (changes requested):

jedel1043:
  
  | Total | 94,188 | 94,188 | 0 |
  | Passed | 70,164 | 70,164 | 0 |
  | Ignored | 18,605 | 18,605 | 0 |
  | Failed | 5,419 | 5,419 | 0 |
  | Panics | 0 | 0 | 0 |
  | Conformance | 74.49% | 74.49% | 0.00% |

jedel1043 (commented):
Implementation-wise looks good! Can you add a test case on https://github.com/boa-dev/boa/blob/main/boa_engine/src/tests.rs ? I thought the test262 suite would catch this, but apparently it doesn't...

jedel1043:
There are some failing checks. Can you run `cargo fmt`?

jedel1043 (approved):
Looks perfect to me!

boa_parser/src/parser/expression/update.rs
```diff
@@ -56,6 +56,19 @@ impl UpdateExpression {
     }
 }
 
+fn is_simple(expr: &Expression, position: Position, strict: bool) -> ParseResult<bool> {
```
Razican, line 59: Could we add a link to the spec explaining what a "simple" target type is?

veera-sivarajan: I've added a link but it explains "simple" target type in terms of syntax-directed operation. My implementation is a function that takes a few arguments. Is that ok?

Razican: I think that's OK, I would just use doc comments `///` instead of `//`, use `<` and `>` to enclose the link, and add a small explanation "this function checks if the target type is simple".

veera-sivarajan: Thanks for the suggestion. I made the changes.

Razican (approved):
It's good for me! Thank you for the changes!! :)

Razican:
Bors r+

PR #2513, opened 2023-01-04
[Merged by Bors] - Refactor some class features

raskad:
This Pull Request fixes various bugs related to classes.

The biggest changes are:

- Changed private names to be unique across multiple classes.
- Changed private name resolution to work via a visitor after a class is parsed. The way class early errors are defined makes it impossible to perform private name resolution while parsing.
- Added function names to class methods.
- Added class name binding to method function environments.
- Separated opcodes for `static` and non-`static` class method definitions to make the above operations possible.

There are still some bugs and further issues with classes but this is already a lot.

### DIFF
--- a/boa_ast/src/function/class.rs
+++ b/boa_ast/src/function/class.rs
@@ -1,6 +1,4 @@
-use core::ops::ControlFlow;
-use std::borrow::Cow;
-
+use super::Function;
 use crate::{
     block_to_string,
     expression::{Expression, Identifier},
@@ -11,8 +9,9 @@ use crate::{
     Declaration, StatementList, ToStringEscaped,
 };
 use boa_interner::{Interner, Sym, ToIndentedString, ToInternedString};
-
-use super::Function;
+use core::ops::ControlFlow;
+use std::borrow::Cow;
+use std::hash::Hash;
 
 /// A class declaration, as defined by the [spec].
 ///
@@ -27,8 +26,9 @@ use super::Function;
 pub struct Class {
     name: Option<Identifier>,
     super_ref: Option<Expression>,
-    constructor: Option<Function>,
-    elements: Box<[ClassElement]>,
+    pub(crate) constructor: Option<Function>,
+    pub(crate) elements: Box<[ClassElement]>,
+    has_binding_identifier: bool,
 }
 
 impl Class {
@@ -40,12 +40,14 @@ impl Class {
         super_ref: Option<Expression>,
         constructor: Option<Function>,
         elements: Box<[ClassElement]>,
+        has_binding_identifier: bool,
     ) -> Self {
         Self {
             name,
             super_ref,
             constructor,
             elements,
+            has_binding_identifier,
         }
     }
 
@@ -75,6 +77,13 @@ impl Class {
     #[must_use]
     pub const fn elements(&self) -> &[ClassElement] {
         &self.elements
+    }
+
+    /// Returns whether the class has a binding identifier.
+    #[inline]
+    #[must_use]
+    pub const fn has_binding_identifier(&self) -> bool {
+        self.has_binding_identifier
     }
 }
 
@@ -238,7 +247,7 @@ impl ToIndentedString for Class {
                             MethodDefinition::Set(_) => "set ",
                             _ => "",
                         },
-                        interner.resolve_expect(*name),
+                        interner.resolve_expect(name.description()),
                         match &method {
                             MethodDefinition::Get(expr)
                             | MethodDefinition::Set(expr)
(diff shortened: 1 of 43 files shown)


boa_engine/src/object/mod.rs
```diff
@@ -1773,58 +1772,119 @@ impl Object {
         self.properties.remove(key)
     }
 
+    /// Check if a private name exists.
```
jedel1043, line 1775: I'm seeing this resembles the [PrivateElementFind](https://tc39.es/ecma262/#sec-privateelementfind) operation and friends. Could this be moved to `boa_engine/src/object/operations.rs` and adapted to somewhat resemble the spec operations? Probably also `initialize_instance_elements`.

I say this because there's a host hook [HostEnsureCanAddPrivateElement](https://tc39.es/ecma262/#sec-hostensurecanaddprivateelement) that needs to be called on the [PrivateFieldAdd](https://tc39.es/ecma262/#sec-privatefieldadd) and [PrivateMethodOrAccessorAdd](https://tc39.es/ecma262/#sec-privatemethodoraccessoradd) operations to restrict the addition of new private fields for an object.

boa_ast/src/function/class.rs
```diff
@@ -505,3 +528,52 @@ impl VisitWith for ClassElement {
...
+/// [spec]: https://tc39.es/ecma262/#sec-private-names
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "fuzz", derive(arbitrary::Arbitrary))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct PrivateName {
+    /// The `[[Description]]` internal slot of the private name.
+    description: Sym,
+
+    /// The indices of the private name are included to ensure that private names are unique.
+    pub(crate) indices: (usize, usize),
```
jedel1043, line 543: Seeing the spec about private names it seems that `PrivateName`s should be globally unique, but `indices` are only unique in the context of a single `Context`, right? I was thinking that the global uniqueness property is probably to restrict the access of private names from objects that share the same `indices` but don't belong to the same `Context`, but to me it seems like this doesn't prevent that? I ask this because I don't know if you've already considered that.

jedel1043 (changes requested):
Really impressive work! I have some questions about our spec compliance though.

boa_ast/src/function/class.rs
raskad: There is still some uniqueness missing with this implementation. We indeed have to add some runtime unique identifier to separate `PrivateName` of multiple objects that are created from the same class. I think this will probably be some kind of `Context` wide index. As far as different `Context`s go, I think that would need some further work as our `Context` is currently also a `Realm`. If we have figured out how we can separate those, the `Context` should be the outermost api to the engine and imo should the the uniqueness-border.

I left the runtime uniqueness out of the PR to limit the scope somewhat.

jedel1043: It's okay! Can you add a `TODO` describing this then? It would be good to track this for when we implement proper Realms.

boa_engine/src/object/mod.rs
raskad: Good call. I've refactored those methods and they are mostly equivalent to the spec now.

boa_engine/src/object/operations.rs
```diff
@@ -656,19 +657,331 @@ impl JsObject {
...
+            Some(PrivateElement::Accessor { getter, .. }) => {
+                // 5. If entry.[[Get]] is undefined, throw a TypeError exception.
+                // 6. Let getter be entry.[[Get]].
+                let getter = if let Some(getter) = getter {
+                    getter
+                } else {
+                    return Err(JsNativeError::typ()
+                        .with_message("private property was defined without a getter")
+                        .into());
+                };
```
jedel1043, line 829: Could be shorter using `ok_or_else`

boa_engine/src/object/operations.rs
```diff
@@ -656,19 +657,331 @@ impl JsObject {
...
+    ) -> JsResult<()> {
+        // 1. Let entry be PrivateElementFind(O, P).
+        // Note: This function is inlined here for mutable access.
+        let mut object_mut = self.borrow_mut();
+        let mut entry = None;
+        for (key, value) in &mut object_mut.private_elements {
+            if key == name {
+                entry = Some(value);
+            }
+        }
```
jedel1043, line 859: Maybe something like
```suggestion
        let mut entry = object_mut.private_elements.iter_mut().find(|(key, _)| key == name);
```

jedel1043 (changes requested):
Everything looks good! Just some small nitpicks.
Also, thanks for adding the `host_ensure_can_add_private_element`! It'll make it easier to replace when the hooks API merges.

jedel1043 (approved):
Nice job!

nekevss (approved):
Looks good!

PR #2515, opened 2023-01-04
[Merged by Bors] - Add early errors for 'eval' or 'arguments' in parameters

raskad:

- Add early errors for functions to make sure that 'eval' or 'arguments' cannot be used as binding identifiers in function parameters. When the function body contains a strict directive, this also has to be accounted for.
- Fix early errors for function identifiers to make sure they cannot be 'eval' or 'arguments' when a function body contains a strict directive.

### DIFF
--- a/boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
+++ b/boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
@@ -115,13 +115,15 @@ where
 
         // Early Error: If BindingIdentifier is present and the source code matching BindingIdentifier is strict mode code,
         // it is a Syntax Error if the StringValue of BindingIdentifier is "eval" or "arguments".
-        if (cursor.strict_mode() || body.strict())
-            && [Some(Sym::EVAL), Some(Sym::ARGUMENTS)].contains(&name.map(Identifier::sym))
-        {
-            return Err(Error::lex(LexError::Syntax(
-                "unexpected identifier 'eval' or 'arguments' in strict mode".into(),
-                name_span.start(),
-            )));
+        if let Some(name) = name {
+            if (cursor.strict_mode() || body.strict())
+                && [Sym::EVAL, Sym::ARGUMENTS].contains(&name.sym())
+            {
+                return Err(Error::lex(LexError::Syntax(
+                    "unexpected identifier 'eval' or 'arguments' in strict mode".into(),
+                    name_span.start(),
+                )));
+            }
         }
 
         // Catch early error for BindingIdentifier, because strictness of the functions body is also
@@ -142,12 +144,7 @@ where
             params_start_position,
         )?;
 
-        let function = AsyncFunction::new(
-            if name.is_some() { name } else { self.name },
-            params,
-            body,
-            name.is_some(),
-        );
+        let function = AsyncFunction::new(name.or(self.name), params, body, name.is_some());
 
         if contains(&function, ContainsSymbol::Super) {
             return Err(Error::lex(LexError::Syntax(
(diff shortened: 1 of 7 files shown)


boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
```diff
@@ -124,6 +113,26 @@ where
...
+        // Early Error: If BindingIdentifier is present and the source code matching BindingIdentifier is strict mode code,
+        // it is a Syntax Error if the StringValue of BindingIdentifier is "eval" or "arguments".
+        if (cursor.strict_mode() || body.strict())
+            && [Some(Sym::EVAL), Some(Sym::ARGUMENTS)].contains(&name.map(Identifier::sym))
+        {
+            return Err(Error::lex(LexError::Syntax(
+                "unexpected identifier 'eval' or 'arguments' in strict mode".into(),
+                name_span.start(),
+            )));
+        }
```
jedel1043, line 125: Would be better to use `if let Some(name) = name` to save us the check in case the name is `None`, and to simplify the logic of the `contains` call. Also applies to the other functions' parsing.

boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
```diff
@@ -133,7 +142,12 @@ where
             params_start_position,
         )?;
 
-        let function = AsyncFunction::new(name, params, body, has_binding_identifier);
+        let function = AsyncFunction::new(
+            if name.is_some() { name } else { self.name },
```
jedel1043, line 146: `unwrap_or`? Though, this won't apply anymore if we go with the `FunctionName` suggestion.

boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
```diff
@@ -133,7 +142,12 @@ where
...
+            params,
+            body,
+            name.is_some(),
+        );
```
jedel1043, line 150: Now that I'm reviewing this I just realized, can't we encode the last parameter into the name itself? An enum like ...

```Rust
enum FunctionName {
    None,
    Direct(Identifier)
    Indirect(Identifier)
}
```

... where `Direct` represents names defined directly on the function definition (`function name()`) and `Indirect` represents names defined by variable assignments and others `let name = function()`.

jedel1043 (changes requested):
Very nice work!

boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
raskad: `unwrap_or` does not work because it takes `T`. `self.name` is also `Option<T>`. I'm not sure if there if a dedicated function on `Option` for this operation.

boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
raskad: We could but it would only make sense if we would also replace `has_binding_identifier` with that. I think this would be a good improvement for making the ast more readable. I would lift this from this PR, maybe into an issue for new contributers? Seems like a nice change to implement.

boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
jedel1043: I think it's worth. And moreso knowing `if let` chains are coming soon™.

boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
jedel1043: Ah, should be `or` then.

boa_parser/src/parser/expression/primary/async_function_expression/mod.rs
jedel1043: Yep, gonna open an issue for it.

HalidOdat (approved):

PR #2516, opened 2023-01-05
[Merged by Bors] - Pass a receiver value in property getter opcodes

raskad:

- Pass a receiver value to the object `get` function in the `GetPropertyBy*` opcodes. The receiver value may be different from the object, because `ToObject` is not called on it.


jedel1043 (approved):
Nice catch!

HalidOdat (approved):

raskad:
bors retry

PR #2517, opened 2023-01-05
[Merged by Bors] - Add regex literal early errors

raskad:

- Parse regex literals via `regress` during parsing to return errors in the regex as early syntax errors.


HalidOdat (approved):
Looks good  to me! :)

jedel1043 (approved):
Looks good! Would be good to store the parsed Regex instead of parsing it again on the engine, but that should be on a separate PR.

PR #2518, opened 2023-01-06
[Merged by Bors] - Remove `Literal::Undefined`

### LINKED ISSUE #2512: `undefined` is inconsistently treated as literal and identifier
**Describe the bug**
The `undefined` identifier is treated as both a literal (on the AST) and an identifier (on the parser) by the engine, causing bugs on parsing. 

**To Reproduce**
@VTCAKAVSMoACE's reproduction, found by our parser idempotency fuzzer:

```bash
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `38`,
 right: `39`: The number of interned symbols changed; a new string was read.
Before:
undefined;

After:
undefined;

Before (AST):
StatementList {
    statements: [
        Statement(
            Expression(
                Literal(
                    Undefined,
                ),
            ),
        ),
    ],
    strict: false,
}
After (AST):
StatementList {
    statements: [
        Statement(
            Expression(
                Identifier(
                    Identifier {
                        ident: Sym {
                            value: 39,
                        },
                    },
                ),
            ),
        ),
    ],
    strict: false,
}', fuzz_targets/parser-idempotency.rs:28:9
```

The literal `Literal::Undefined` shouldn't exist, since the engine should treat `undefined` as an identifier name, as specified by the [spec](https://tc39.es/ecma262/multipage/global-object.html#sec-undefined).

veera-sivarajan:
This Pull Request fixes/closes #2512 .

Removes `Literal::Undefined` so that `undefined` is treated as an identifier name. Ran the parser's idempotency fuzzer and ensured the bug doesn't reproduce.

### DIFF
--- a/boa_ast/src/expression/literal/mod.rs
+++ b/boa_ast/src/expression/literal/mod.rs
@@ -110,16 +110,6 @@ pub enum Literal {
     /// [spec]: https://tc39.es/ecma262/#sec-null-value
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Glossary/null
     Null,
-
-    /// The `undefined` is a primitive value automatically assigned to variables that have just been declared, or to formal arguments for which there are no actual arguments.
-    ///
-    /// More information:
-    ///  - [ECMAScript reference][spec]
-    ///  - [MDN documentation][mdn]
-    ///
-    /// [spec]: https://tc39.es/ecma262/#sec-undefined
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):

Razican (approved):
Thanks!!

raskad (approved):

PR #2520, opened 2023-01-07
[Merged by Bors] - Fix postfix operator line terminator parsing

raskad:
Postfix increment / decrement operators require that there is no line terminator between the LHS expression and  the operator. This was previously ignored.

### DIFF
--- a/boa_parser/src/parser/expression/update.rs
+++ b/boa_parser/src/parser/expression/update.rs
@@ -123,38 +123,40 @@ where
         let lhs = LeftHandSideExpression::new(self.name, self.allow_yield, self.allow_await)
             .parse(cursor, interner)?;
 
-        if let Some(tok) = cursor.peek(0, interner)? {
-            let token_start = tok.span().start();
-            match tok.kind() {
-                TokenKind::Punctuator(Punctuator::Inc) => {
-                    cursor
-                        .next(interner)?
-                        .expect("Punctuator::Inc token disappeared");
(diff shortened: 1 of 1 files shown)


boa_parser/src/parser/expression/update.rs
```diff
@@ -123,38 +123,40 @@ where
...
-                    cursor
-                        .next(interner)?
-                        .expect("Punctuator::Inc token disappeared");
-                    // https://tc39.es/ecma262/#sec-update-expressions-static-semantics-early-errors
-                    if !is_simple(&lhs, token_start, cursor.strict_mode())? {
-                        return Err(Error::lex(LexError::Syntax(
-                            "Invalid left-hand side in assignment".into(),
-                            token_start,
-                        )));
+        if !cursor.peek_is_line_terminator(0, interner)?.unwrap_or(true) {
```
jedel1043, line 126: Maybe this would be better as an early return to reduce rightward drift.

jedel1043 (commented):
Looks good! Just a small comment to improve readability.

boa_parser/src/parser/expression/update.rs
raskad: Very true, fixed!

jedel1043 (approved):
Nice!

jedel1043:
We need to fix the random tarpaulin errors...

nekevss (approved):
Nice catch 😄

PR #2521, opened 2023-01-07
[Merged by Bors] - Recognize Directive Prologues correctly

raskad:

- Recognize the `"use strict"` directive prologue correctly.
- Refactor parsers to remove a setter function.

### DIFF
--- a/boa_parser/src/parser/statement/mod.rs
+++ b/boa_parser/src/parser/statement/mod.rs
@@ -50,6 +50,7 @@ use boa_ast::{
     Keyword, Punctuator,
 };
 use boa_interner::Interner;
+use boa_macros::utf16;
 use boa_profiler::Profiler;
 use std::io::Read;
 
@@ -229,6 +230,8 @@ pub(super) struct StatementList {
     allow_await: AllowAwait,
     allow_return: AllowReturn,
     break_nodes: &'static [TokenKind],
+    directive_prologues: bool,
+    strict: bool,
 }
 
 impl StatementList {
@@ -238,6 +241,8 @@ impl StatementList {
         allow_await: A,
         allow_return: R,
         break_nodes: &'static [TokenKind],
+        directive_prologues: bool,
+        strict: bool,
     ) -> Self
     where
         Y: Into<AllowYield>,
@@ -249,6 +254,8 @@ impl StatementList {
             allow_await: allow_await.into(),
             allow_return: allow_return.into(),
             break_nodes,
+            directive_prologues,
+            strict,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


jedel1043 (approved):
Looks great!

boa_parser/src/parser/statement/mod.rs
```diff
@@ -283,6 +294,25 @@ where
...
+                        |g| g == utf16!("use strict"),
+                        true,
+                    ) {
+                        cursor.set_strict_mode(true);
+                        strict = true;
+                    }
+                } else {
+                    directive_prologues = false;
+                }
+            }
```
jedel1043, line 314: I think I found a small bug. On this PR, this will throw:

```javascript
function f(){
    "a";
    ;
    "use strict";
    let public = 5;
}
```

However, I ran that on Spidermonkey and it successfully runs:

image

This means the parser should stop considering string expressions as directives if there's an empty statement between them.

jedel1043, line 314: I think I found a small bug. On this PR, this will throw:

```javascript
function f(){
    "a";
    ;
    "use strict";
    let public = 5;
}
```

However, I ran that on Spidermonkey and it successfully runs:

image

This means the parser should stop considering string expressions as directives if there's an empty statement between them.

jedel1043 (changes requested):

boa_parser/src/parser/statement/mod.rs
jedel1043: I don't know why we are skipping consecutive semicolons though, the parser should parse that as a `Statement::Empty` statement...

jedel1043: Huh, test262 doesn't have a test to assert this. Can you add my snippet to our suite?

raskad: Yeah. I'll have to fix some other places where we do not parse semicolon correctly. That's why some of our tests are now breaking...

jedel1043: Hm yeah, the fix seems less trivial than I thought...
If you prefer, let's split this PR into two parts to avoid polluting this one with unrelated fixes.

raskad: Yeah I'm working on it now and there are some things in the for loop parsing that need to be fixed. I will split it into a new PR and push the initial changes to this one.

jedel1043 (approved):
All good then!

nekevss (approved):
Looks good to me! 😄

PR #2522, opened 2023-01-07
chore: add .devcontainer

LeoDog896:
This adds a rust [dev container](https://containers.dev/) for spinning up a dev environment locally.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,15 @@
+{
+	// See https://aka.ms/vscode-remote/devcontainer.json for format details.
+	"name": "boa",
+	"image": "mcr.microsoft.com/devcontainers/rust:0-1-bullseye",
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"rust-lang.rust-analyzer"
+			]
+		}
+	},
+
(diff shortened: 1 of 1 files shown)


raskad:
I'm not really sure how much of local development files should be in the repo, especially if none of the maintainers use the methods. For example I personally do not use Development Containers. Any other opinions or does anyone use it?

@Razican @jasonwilliams @HalidOdat @RageKnify @jedel1043 @nekevss

jasonwilliams:
I agree with @raskad we used to have a devcontainer in this repo before back in its early days but it wasn’t maintained. 

I think most contributors just use WSL or their own Linux system. Any reason why needing the container?

LeoDog896:
Honestly, I just find it easier to use to spin up a local environment without having to change any of my local computer tooling. However, given that it doesn't have much of a use since the environment setup isn't too complex, I don't see much of a use for it either. If the build process (though I doubt it) ever requires any additional tooling, however, please do let me know.

jasonwilliams:
Boa has remained pretty simple to set up. We assume anyone wanting to contribute to Boa already has some interest in Rust and thus already has the Rust toolchain installed. There isn’t much else needed.

Issue #2424, opened 2022-11-08
`JumpIfFalse` skips over `PopEnvironment` instruction


HalidOdat:
**Describe the bug**
We do not handle pushing and popping of  environments correctly in the bytecompiler.

**To Reproduce**

```javascript
let i = 0;
while (i < 10) {
    if (i == 3) {
        break;
    }
    i++;
}
```

The following instructions are generated:
```
----------------------Compiled Output: '<main>'-----------------------
Location  Count   Opcode                     Operands

000000    0000    PushZero
000001    0001    DefInitLet                 0000: 'i'
000006    0002    LoopStart
000007    0003    LoopContinue
000008    0004    GetName                    0000: 'i'
000013    0005    PushInt8                   10
000015    0006    LessThan
000016    0007    JumpIfFalse                78
000021    0008    PushDeclarativeEnvironment 0, 1
000030    0009    GetName                    0000: 'i'
000035    0010    PushInt8                   3
000037    0011    Eq
000038    0012    JumpIfFalse                58
000043    0013    PushDeclarativeEnvironment 0, 0
000052    0014    Jump                       78
000057    0015    PopEnvironment
000058    0016    GetName                    0000: 'i'
000063    0017    IncPost
000064    0018    RotateRight                2
000066    0019    SetName                    0000: 'i'
000071    0020    Pop
000072    0021    PopEnvironment
000073    0022    Jump                       7
000078    0023    LoopEnd

Literals:
    <empty>

Bindings:
    0000: i

Functions:
    <empty>
```

**Actual behavior**
The `0038`: `JumpIfFalse` instruction is jumping to instruction `58`, skipping the `0057`: `PopEnvironment`, and the plain `0073`: `Jump` also skips it by jumping to the end of the function.

**Expected behavior**
The `PopEnvironment` instruction should be executed.

**Build environment (please complete the following information):**

- OS: Arch Linux
- Version: Rolling Release (kernel version 5.15.73-1-lts)
- Target triple: x86_64-unknown-linux-gnu
- Rustc version: rustc 1.65.0 (897e37553 2022-11-02)

**Additional context**

Here is the flow graphs generated by #2422 :
<details>
<img src="https://user-images.githubusercontent.com/8566042/200647479-dbec734e-7140-49eb-9144-823895a89eb8.png">
</details>

We can see that there is a `PopEnvironment` that is never executed (since there is nothing that flows into it).

HalidOdat:
[...]

PR #2523, opened 2023-01-08
[Merged by Bors] - `Break` Opcode and `ByteCompiler` changes

nekevss:
Hi all! 😄

This Pull Request addresses #2424. There are also a few changes made to the `ByteCompiler`, the majority of which are involving `JumpControlInfo`.

It changes the following:

- Adds `Break` Opcode
- Shifts `compile_stmt` into the `statement` module. 
- Moves `JumpControlInfo` to it's own module.

### DIFF
--- a/boa_engine/src/bytecompiler/jump_control.rs
+++ b/boa_engine/src/bytecompiler/jump_control.rs
@@ -13,6 +13,7 @@ use crate::{
     bytecompiler::{ByteCompiler, Label},
     vm::Opcode,
 };
+use bitflags::bitflags;
 use boa_interner::Sym;
 use std::mem::size_of;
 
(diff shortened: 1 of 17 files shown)


boa_engine/src/bytecompiler/jump_control.rs
```diff
@@ -0,0 +1,344 @@
...
+    Loop,
+    Switch,
+    Try,
+    LabelledBlock,
+}
+
+// ---- Creation Methods ---- //
+
+impl JumpControlInfo {
+    pub(crate) const fn new() -> Self {
```
jasonwilliams, line 33: should this be ::default() instead of ::new()?

boa_engine/src/bytecompiler/jump_control.rs
```diff
@@ -0,0 +1,344 @@
+use crate::{
```
jasonwilliams, line 1: It would be good to see top level documentation for this, doesn't have to be the level of [this](https://github.com/boa-dev/boa/blob/main/boa_engine/src/environments/mod.rs#L1-L25) but similar.

jasonwilliams (approved):
The rest LGTM
Nice work!

boa_engine/src/bytecompiler/jump_control.rs
```diff
@@ -0,0 +1,366 @@
...
+// `JumpControlInfo` related methods that are implemented on `ByteCompiler`.
+impl ByteCompiler<'_, '_> {
+    /// Pushes a generic `JumpControlInfo` onto `ByteCompiler`
+    ///
+    /// Default `JumpControlInfoKind` is `JumpControlInfoKind::Loop`
+    pub(crate) fn push_new_jump_control(&mut self) {
+        self.jump_info.push(JumpControlInfo::default());
+    }
+
+    pub(crate) fn set_jump_control_label(&mut self, label: Option<Sym>) {
```
jedel1043, line 177: To avoid having a lot of duplicated setters between `ByteCompiler` and `JumpControlInfo`, what about having `ByteCompiler::current_jump_control_mut` returning `&mut JumpControlInfo`?

boa_engine/src/bytecompiler/jump_control.rs
```diff
@@ -0,0 +1,366 @@
...
+        }
+    }
+
+    pub(crate) fn push_labelled_block_control_info(&mut self, label: Sym, start_address: u32) {
+        let new_info = JumpControlInfo::default()
+            .with_kind(JumpControlInfoKind::LabelledBlock)
+            .with_label(Some(label))
+            .with_start_address(start_address);
+        self.jump_info.push(new_info);
+    }
```
jedel1043, line 349: I'm seeing that some fields on the `JumpControlInfo` are only used on some `JumpControlInfoKind`s; `jump_info.breaks` only makes sense on loops and blocks, and `has_finally` only makes sense on try blocks. Could those fields be moved into the `JumpControlInfoKind` itself?

jedel1043 (changes requested):
Great work! I opened some discussions about some improvements that could be made :)

boa_engine/src/bytecompiler/jump_control.rs
nekevss: I think it could if it was switched over to one bitflag. I was playing around with the idea when I was working on the changes initially but had decided to not do it at the time. I tried it out on the most recent commit, let me know what you think. 😄

jedel1043: Looks good! I was thinking more about using an ADT to represent the types of control infos, but on retrospect I think that'll require a lot more design work and I don't know if it'll be worth it.

jedel1043 (approved):
Nice job!

jedel1043:
bors retry

PR #2528, opened 2023-01-11
[Merged by Bors] - Prepare `Promises` for new host hooks and job queue API

jedel1043:
As part of the new modules PR, I was working on implementing the [host hooks](https://tc39.es/ecma262/#sec-host-hooks-summary) for the module import hooks and custom job queues. However, the promises module needed a bit of a refactor in order to couple with the new API. So, I thought it was a good idea to separate the promises refactor into its own PR, since the other PR is already big as it is.

- Replaced some usages of `JobCallback` with a new `NativeJob` that isn't traced by the GC, since those closures are always rooted and executed by the `Context` globally. This will also allow hosts to pass their custom jobs to the job queue, and maybe could also accept futures in the Future (pun intended 😆).
- Refactored several functions to account for the `HostPromiseRejectionTracker` hook which needs the promise `JsObject`. 
- Rewrote some patterns with newer Rust idioms.


boa_engine/src/object/operations.rs
```diff
@@ -669,6 +669,35 @@ impl JsObject {
     // todo: DefineField
 
     // todo: InitializeInstanceElements
+
+    /// Abstract operation `( V, P [ , argumentsList ] )`
```
raskad, line 673:
```suggestion
    /// Abstract operation `Invoke ( V, P [ , argumentsList ] )`
```

raskad (approved):
Nice refactor and docs!

nekevss (approved):
LGTM! Nice work! 😄

PR #2529, opened 2023-01-11
[Merged by Bors] - Implement host hooks and job queues APIs

jedel1043:
Follows from #2528, and should complement #2411 to implement the module import hooks.

~~Similarly to the Intl/ICU4X PR (#2478), this has a lot of trivial changes caused by the new lifetimes. I thought about passing the queue and the hooks by value, but it was very painful having to wrap everything with `Rc` in order to be accessible by the host.
In contrast, `&dyn` can be easily provided by the host and has the advantage of not requiring additional allocations, with the downside of adding two more lifetimes to our `Context`, but I think it's worth.~~ I was able to unify all lifetimes into the shortest one of the three, making our API just like before!

Changes:
- Added a new `HostHooks` trait and a `&dyn HostHooks` field to `Context`. This allows hosts to implement the trait for their custom type, then pass it to the context.
- Added a new `JobQueue` trait and a `&dyn JobQueue` field to our `Context`, allowing custom event loops and other fun things.
- Added two simple implementations of `JobQueue`: `IdleJobQueue` which does nothing and `SimpleJobQueue` which runs all jobs until all successfully complete or until any of them throws an error.
- Modified `boa_cli` to run all jobs until the queue is empty, even if a job returns `Err`. This also prints all errors to the user.


jedel1043:
I thought a bit more about the design of this and I think I can unify the three lifetimes into only one with some lifetime trickery, making this PR a lot more lightweight! I'll make the fix today.

boa_engine/src/job.rs
```diff
@@ -1,12 +1,37 @@
-//! Data structures for the microtask job queue.
+//! `ECMAScript` jobs and job queues.
```
nekevss, line 1: Small doc nitpick. I think this could probably be replaced with a modified version of the sentence below. Something like: "Boa's implementation and API to create and customize ECMAScript jobs and job queues."

jasonwilliams:
👀

boa_engine/src/native_function.rs
```diff
@@ -197,169 +195,3 @@ impl NativeFunction {
...
-
-    /// Calls this `GenericNativeFunction`, forwarding the arguments to the corresponding function.
-    #[inline]
-    pub fn call(&mut self, args: Args, context: &mut Context<'_>) -> Ret {
-        match self.inner {
-            GenericInner::PointerFn(f) => f(args, context),
-            GenericInner::Closure(ref mut c) => c.call(args, context),
-        }
-    }
-}
```
raskad, line 365: I think removing this is not desired, right?

jedel1043: Yep. I initially added this to implement the callbacks for the hooks, but while designing I found out that we could use a trait object with the implementations instead, so we don't need this anymore.

raskad (approved):
Nice work. A great step in the direction of making boa more configurable for users and allowing the host implementations that the spec lists.

HalidOdat (approved):
Great job @jedel1043 , This looks perfect to me! :)

PR #2530, opened 2023-01-13
[Merged by Bors] - Improve the design of ephemerons in our GC

jedel1043:
- Modifies `EphemeronBox` to be more akin to `GcBox`, with its own header, roots and markers. This also makes it more similar to [Racket's](https://docs.racket-lang.org/reference/ephemerons.html) implementation.
- Removes `EPHEMERON_QUEUE`.
- Ephemerons are now tracked on a special `weak_start` linked list, instead of `strong_start` which is where all other GC boxes live.
- Documents all unsafe blocks.
- Documents our current garbage collection algorithm. I hope this'll clarify a bit what exactly are we doing on every garbage collection.
- Renames/removes some functions.

### DIFF
--- a/boa_gc/src/lib.rs
+++ b/boa_gc/src/lib.rs
@@ -79,6 +79,7 @@
     clippy::perf,
     clippy::pedantic,
     clippy::nursery,
+    clippy::undocumented_unsafe_blocks
 )]
 #![allow(
     clippy::module_name_repetitions,
@@ -95,6 +96,7 @@ mod trace;
 pub(crate) mod internals;
 
 use boa_profiler::Profiler;
+use internals::{EphemeronBox, ErasedEphemeronBox};
 use std::{
     cell::{Cell, RefCell},
     mem,
@@ -108,13 +110,14 @@ pub use internals::GcBox;
 pub use pointers::{Ephemeron, Gc, WeakGc};
 
 type GcPointer = NonNull<GcBox<dyn Trace>>;
+type EphemeronPointer = NonNull<dyn ErasedEphemeronBox>;
 
-thread_local!(static EPHEMERON_QUEUE: Cell<Option<Vec<GcPointer>>> = Cell::new(None));
 thread_local!(static GC_DROPPING: Cell<bool> = Cell::new(false));
 thread_local!(static BOA_GC: RefCell<BoaGc> = RefCell::new( BoaGc {
     config: GcConfig::default(),
     runtime: GcRuntimeData::default(),
-    adult_start: Cell::new(None),
+    strong_start: Cell::new(None),
+    weak_start: Cell::new(None)
 }));
 
 #[derive(Debug, Clone, Copy)]
@@ -145,7 +148,8 @@ struct GcRuntimeData {
 struct BoaGc {
     config: GcConfig,
     runtime: GcRuntimeData,
-    adult_start: Cell<Option<GcPointer>>,
+    strong_start: Cell<Option<GcPointer>>,
+    weak_start: Cell<Option<EphemeronPointer>>,
 }
 
 impl Drop for BoaGc {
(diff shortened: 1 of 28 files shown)


nekevss (commented):
I'm going to try and look at this in more depth and play around with it this weekend, but I wanted to leave some early comments.

On the topic of improvements/clarity changes, I was thinking about the current names of some structs. In particular, thoughts on switching `GcCell` to `GcRefCell`, `GcCellRef` to `GcRef`, and `GcCellRefMut` to `GcRefMut`?

boa_gc/src/pointers/rootable.rs
```diff
@@ -0,0 +1,87 @@
+use std::ptr::{self, addr_of_mut, NonNull};
+
+/// A [`NonNull`] pointer with a `rooted` tag.
+///
+/// This pointer can be created only from pointers that are 2-byte aligned. In other words,
+/// the pointer must point to an address that is a multiple of 2.
+pub(crate) struct Rootable<T: ?Sized> {
```
nekevss, line 7: I really like this. It's fantastic, imo.

boa_gc/src/pointers/rootable.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// See [`NonNull::as_ref`].
+    pub(crate) unsafe fn as_ref(&self) -> &T {
+        // SAFETY: it is the caller's job to ensure the safety of this operation.
+        unsafe { self.as_ptr().as_ref() }
+    }
+}
+
+// Technically, this function is safe, since we're just modifying the address of a pointer without
+// dereferencing it.
+fn set_data_ptr<T: ?Sized, U>(mut ptr: *mut T, data: *mut U) -> *mut T {
```
nekevss, line 81: Thoughts on adding this as a function on `Rootable` so we call `Rootable::set_data_ptr`? I can't think of a use case where we would use this function without `Rootable` already being in scope and it may make it less ambiguous.

jedel1043: I'd agree if `set_data_ptr` was a `pub(crate)` function, but it is only used here and it will probably be deleted when either https://github.com/rust-lang/rust/issues/81513 or https://github.com/rust-lang/rust/pull/103702 with https://github.com/rust-lang/rust/issues/95228 stabilizes.

jedel1043:
> thoughts on switching `GcCell` to `GcRefCell`, `GcCellRef` to `GcRef`, and `GcCellRefMut` to `GcRefMut`?

I like it! The current names make `GcCell` look like it has the semantics of `Cell` but garbage collected, which is definitely not true. I'll rename them.

raskad (approved):
Nice changes. I like the renaming too.

boa_gc/src/lib.rs
```diff
@@ -222,141 +248,153 @@ impl Allocator {
...
+            };
         }
 
-        // Any left over nodes in the ephemeron queue at this point are
-        // unreachable and need to be notified/finalized.
-        finalize.extend(ephemeron_queue);
-
-        finalize
-    }
+        // === Weak mark phase ===
```
nekevss, line 336: Apologies about the delay on this. I still need to try and do some testing on this branch, but I thought I'd leave a review with some feedback/what I'm looking at. 

A lot of this looks good to me, but I'm a little concerned here on the weak mark phase since this is adding all ephemerons to the queue by default. It definitely would be a better approach than the thread_local queue if it works, but I'm worried this may end up holding onto some objects longer than expected or creating an immortal self-referencing chain. Would it be possible to add more test cases involving a chain of weak and non-weak items?

nekevss (changes requested):

boa_gc/src/lib.rs
jedel1043: Added some tests to check the correct behaviour on self-references

nekevss (approved):
Looks like it should be fine to move forward with. Nice work on the improvements. 😄

jedel1043:
bors retry

PR #2531, opened 2023-01-14
[Merged by Bors] - Clarity changes for the VM

nekevss:
Some small changes to the VM with the hopes of making it a bit more clear and concise.

It changes the following:

- Changes `code` to `code_block` and `code` to `bytecode` in `CallFrame` and `CodeBlock`, respectively.
- Adds some creation methods to `CallFrame`.
- Implements `Default` for `Vm`.

### DIFF
--- a/boa_engine/src/vm/call_frame.rs
+++ b/boa_engine/src/vm/call_frame.rs
@@ -74,6 +74,8 @@ impl CallFrame {
     }
 }
 
+// ---- `CallFrame` stack methods ---- //
+
 impl CallFrame {
     /// Tracks that one environment has been pushed in the current loop block.
     pub(crate) fn loop_env_stack_inc(&mut self) {
--- a/boa_engine/src/vm/call_frame.rs
+++ b/boa_engine/src/vm/call_frame.rs
@@ -39,11 +39,10 @@ pub struct CallFrame {
     pub(crate) async_generator: Option<JsObject>,
 }
 
-
 impl CallFrame {
     /// ---- `CallFrame` creation methods ----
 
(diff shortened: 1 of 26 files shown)


jedel1043 (approved):
Good improvements!

boa_engine/src/vm/call_frame.rs
```diff
@@ -39,6 +39,43 @@ pub struct CallFrame {
     pub(crate) async_generator: Option<JsObject>,
 }
 
+// ---- `CallFrame` creation methods ---- //
```
Razican, line 42: This can be a doc comment in the `impl`, which will appear in the docs.

boa_engine/src/vm/call_frame.rs
```diff
@@ -39,6 +39,43 @@ pub struct CallFrame {
...
+
+impl CallFrame {
+    pub(crate) fn new(code_block: Gc<CodeBlock>) -> Self {
```
Razican, line 45: Could documentation be added to these new functions?

Razican (commented):
Looks much cleaner this way, thanks! I added some comments with potential enhancements to documentation.

raskad (approved):

boa_engine/src/vm/call_frame.rs
```diff
@@ -40,6 +40,46 @@ pub struct CallFrame {
 }
 
 impl CallFrame {
+    /// ---- `CallFrame` creation methods ----
```
Razican, line 43: This should go before the impl, so that the documentation is on the list of methods, instead of only on the first method

Razican (approved):
Thanks!!

PR #2533, opened 2023-01-15
Correctly parse consecutive semicolons

raskad:

- Do not skip consecutive semicolons while parsing a `StatementList`.
- Expect semicolon in `LexicalDeclaration` and add an special case for `for` loop parsing.
- Adjust `StatementList` compilation to skip empty statements.
- Adjust/add tests to make sure consecutive semicolons  are correctly parsed.


jedel1043 (approved):
Nice fix!

nekevss (approved):
Looks good to me. 😄

Issue #2295, opened 2022-09-20
Labeled `if` statement


SheetJSDev:
A slight twist on #2243 :

```js
/* this works in commit 5d9b405 */
foo: {
  console.log('face');
  break foo;
  console.log('this will not be executed');
}
/* This elicits `Uncaught "SyntaxError": "Cannot use the undeclared label 'bar'"` */
bar: if(true){
  console.log('face');
  break bar;
  console.log('this will not be executed');
}
console.log('swap');
```

Razican:
Ideally, all these should be treated as `LabeledStatement` in general. This would be easier to implement once #1371 is implemented.

SheetJSDev:
Commit 90d310b (thanks @nekevss !) fixed this example and added a representative test case.

PR #2534, opened 2023-01-15
[Merged by Bors] - Labelled ByteCompiler Fix

nekevss:
This Pull Request addresses #2295, and another case that I came across when I was adding `Break` to the `ByteCompiler`

I did have a question that came up during this regarding the spec. We currently don't implement the [BreakableStatement](https://tc39.es/ecma262/#prod-BreakableStatement). Any thoughts on whether we should be? Especially since `BreakableStatement` seems to be a bit of a inaccurate since `LabelledStatement` is breakable too.

It changes the following:

- Moves handling of label jump out of `compile_block` and into `compile_labelled`.
- Adds a couple more tests to keep track of `LabelledStatement` breaks.

### DIFF
--- a/boa_engine/src/bytecompiler/jump_control.rs
+++ b/boa_engine/src/bytecompiler/jump_control.rs
@@ -35,7 +35,7 @@ bitflags! {
         const LOOP = 0b0000_0001;
         const SWITCH = 0b0000_0010;
         const TRY_BLOCK = 0b0000_0100;
-        const LABELLED_BLOCK = 0b0000_1000;
+        const LABELLED = 0b0000_1000;
         const IN_CATCH = 0b0001_0000;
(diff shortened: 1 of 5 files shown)


raskad:
> I did have a question that came up during this regarding the spec. We currently don't implement the [BreakableStatement](https://tc39.es/ecma262/#prod-BreakableStatement). Any thoughts on whether we should be? Especially since `BreakableStatement` seems to be a bit of a inaccurate since `LabelledStatement` is breakable too.

Since there are no early errors or specific places where execution depends on the statements being `BreakableStatement`, I think it is fine to not implement it directly.

raskad (approved):
Nice change and godd to add some more tests for `break`!

jedel1043 (approved):
Nice improvements!

PR #2539, opened 2023-01-16
Make `JsSymbol` and `JsString` thread-safe

jedel1043:
The section about `Symbol` on the [specification](https://tc39.es/ecma262/#sec-ecmascript-language-types-symbol-type) says:

> The Symbol type is the set of all non-String values that may be used as the key of an Object property ([6.1.7](https://tc39.es/ecma262/#sec-object-type)).
Each possible Symbol value is unique and immutable.

Our previous implementation of `JsSymbol` used `Rc` and a thread local `Cell<usize>`. However, this meant that two different symbols in two different threads could share the same hash, making symbols not unique.
Also, the [GlobalSymbolRegistry](https://tc39.es/ecma262/#table-globalsymbolregistry-record-fields) is meant to be shared by all realms, including realms that are not in the same thread as the main one; this also forces us to replace our current thread local global symbol registry with a thread-safe one that uses `DashMap` for concurrent access. However, the global symbol registry uses `JsString`s as keys and values, which forces us to either use `Vec<u16>` instead (wasteful and needs to allocate to convert to `JsString` on each access) or make `JsString` thread-safe with an atomic counter. For this reason, I implemented the second option.

- Makes `JsSymbol` thread-safe by using Arc instead of Rc, and making `SYMBOL_HASH_COUNT` an `AtomicU64`.
- Makes `JsString` thread-safe by using `AtomicUsize` instead of `Cell<usize>` for its ref count.
- Makes the `GLOBAL_SYMBOL_REGISTRY` truly global, using `DashMap` as our global map that is shared by all threads.
- Replaces some thread locals with thread-safe alternatives, such as static arrays and static indices.
- Various improvements to all related code for this.


boa_engine/src/symbol.rs
```diff
@@ -260,46 +117,68 @@ struct Inner {
...
-            inner: Rc::new(Inner { hash, description }),
-        }
+    pub fn new(description: Option<JsString>) -> Option<Self> {
+        let hash = get_id()?;
+        let arc = Arc::new(Inner { hash, description });
+
+        Some(Self {
+            // SAFETY: Pointers returned by `Arc::into_raw` must be non-null.
+            repr: unsafe { Tagged::from_ptr(Arc::into_raw(arc).cast_mut()) },
+        })
```
jasonwilliams, line 162: This looks interesting but not sure i understand it.
Could you explain why the JsSymbol holds a Tagged pointer over the Arc? Doesn't the Arc return a pointer anyway?

boa_engine/src/symbol.rs
```diff
@@ -319,17 +209,82 @@ impl JsSymbol {
...
+impl Clone for JsSymbol {
+    fn clone(&self) -> Self {
+        if let UnwrappedTagged::Ptr(ptr) = self.repr.unwrap() {
+            // SAFETY: the pointer returned by `self.repr` must be a valid pointer
+            // that came from an `Arc::into_raw` call.
+            unsafe {
+                let arc = Arc::from_raw(ptr.as_ptr().cast_const());
+                // Don't need the pointer since `self` is already a copyable pointer, just need to
+                // trigger the `clone` impl.
+                let _ = Arc::into_raw(arc.clone());
```
jasonwilliams, line 257: How does this work exactly? Won't the reference count revert back as soon as this line executes as _ drops immediately?

boa_engine/Cargo.toml
```diff
@@ -75,6 +75,8 @@ icu_provider = { version = "1.0.1", optional = true }
 icu_list = { version = "1.0.0", features = ["serde"], optional = true }
 writeable = { version = "0.5.0", optional = true }
 sys-locale = { version = "0.2.3", optional = true }
+dashmap = "5.4.0"
```
jasonwilliams, line 78: is this part of intl deps?

boa_engine/src/symbol.rs
jedel1043: Ah, I think I over-engineered this a bit. Let me fix this.

jedel1043: Changed it to explicitly use `mem::forget` instead.

boa_engine/Cargo.toml
jedel1043: Nope, moved it.

boa_engine/src/symbol.rs
jasonwilliams: > a JsSymbol was 8 bits in size, making it the same size as an usize

Isn't a `usize` 32bits on a 32bit machine, 64bits on a 64bit machine? i.e the pointer size? I've not known of a `usize` be only 1 byte. https://doc.rust-lang.org/std/primitive.usize.html

jedel1043: Sorry! I mistyped. Edited my comment

jedel1043: Also, I assume pointers of 64-bit in size in the comment, but also applies to 32-bit pointers because the tag itself is `usize`.

jasonwilliams (approved):
Looks good to me, we will take a performance hit switching from Rc to Arc, thats unavoidable, but probbaly explains the +5% on the `Symbols (Execution)` benchmark.

As we discussed we probably don't want to take that hit on JsStrings (using atomics) so maybe we can move the penalty into the globalSymbolRegistry.

raskad:
When I run miri (`cargo +nightly miri test -p boa_engine string::tests -- --skip builtins --skip parser`) on this branch I get some UB:

```
test string::tests::as_str ... error: Undefined Behavior: attempting a write access using <1734381> at alloc768919[0x10], but that tag does not exist in the borrow stack for this location
   --> boa_engine/src/string/mod.rs:544:13
    |
544 |             ptr::copy_nonoverlapping(string.as_ptr(), data, count);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             attempting a write access using <1734381> at alloc768919[0x10], but that tag does not exist in the borrow stack for this location
    |             this error occurs as part of an access at alloc768919[0x10..0x1a]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <1734381> would have been created here, but this is a zero-size retag ([0x10..0x10]) so the tag in question does not exist anywhere
...
```

@jedel1043 do you get the same result?

jedel1043:
@raskad I did change some code while integrating `Tagged`! This should be solved now.

boa_engine/src/tagged.rs
```diff
@@ -0,0 +1,109 @@
...
+// Right now this allows us to use the stable polyfill from the `sptr` crate, which uses
+// the same names from the unstable functions of the `std::ptr` module.
+#![allow(unstable_name_collisions)]
+
+use sptr::Strict;
+use std::ptr::NonNull;
+
+/// A pointer that can be tagged with an `usize`.
+///
+/// Only pointers with an alignment of 2-bytes are valid, and the tag must have its MSB unset,
```
nekevss, line 11: Small doc clarity. Can MSB be listed out as `most significant byte (MSB)`. Same on line16 with `least significant byte (LSB)`. I'm actually not sure we need the acronyms in parenthesis, but it can't hurt.

jedel1043: Solved and also reworded some other sentences.

boa_macros/src/lib.rs
```diff
@@ -57,11 +57,59 @@
...
+        if !keys.insert(entry) {
+            return Err(Error::new_spanned(entry, "duplicate string"));
+        }
+    }
+    Ok(())
+}
+
+/// Create an ordered static array of `&[u16]` strings from a list of [`str`] literals.
+#[proc_macro]
+pub fn static_strings(input: TokenStream) -> TokenStream {
```
HalidOdat, line 90: Is this only used internally? If its only meant for internal use we can use the `#[doc(hidden)]` attribute

HalidOdat (commented):
Looks good :)

boa_macros/src/lib.rs
jedel1043: Removed it because there were some perf problems with this approach. Reverted to a local thread map.

HalidOdat (approved):
Looks perfect to me! :)

PR #2545, opened 2023-01-19
Fix conformance page accordion with `.` in suite name

raskad:
The accordions for suite tests on our conformance page currently breaks when the suite name has a `.`. This change makes the accordion show and hide work as expected by replacing the `.` in the relevant identifiers.


jedel1043 (approved):
Good catch!

nekevss (approved):
nice 😄

PR #2546, opened 2023-01-20
[Merged by Bors] - Add early errors for escaped identifiers

raskad:

- Add early errors for escaped characters in object and class setters and getters.
- Add early errors for escaped characters in class `static`.
- Add early errors for escaped characters in `new.target`.
- Add early errors for legacy octal/decial escapes that are used in string literals before a `"use strict"` directive.

### DIFF
--- a/boa_parser/src/lexer/token.rs
+++ b/boa_parser/src/lexer/token.rs
@@ -99,7 +99,7 @@ pub enum TokenKind {
     EOF,
 
     /// An identifier.
-    Identifier((Sym, bool)),
+    Identifier((Sym, ContainsEscapeSequence)),
 
     /// A private identifier.
     PrivateIdentifier(Sym),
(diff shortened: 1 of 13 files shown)


boa_parser/src/lexer/token.rs
```diff
@@ -99,7 +99,7 @@ pub enum TokenKind {
     EOF,
 
     /// An identifier.
-    Identifier(Sym),
+    Identifier((Sym, bool)),
```
jedel1043, line 102: Would be good to have a `IsLegacyEscaped` enum instead of `bool`, because there are a lot of places where it is not obvious what the `bool` represents exactly.

jedel1043 (commented):
Good work! Just a small comment to improve readability.

boa_parser/src/lexer/token.rs
raskad: Good call, I added explicit types instead of the `bool`s.

jedel1043 (approved):
Looks great!

nekevss (approved):
Looks good to me 😄

PR #2551, opened 2023-01-20
[Merged by Bors] - Fix double property access on assignment ops

raskad:
Currently the compilation of assignment operators leads to a double object property access, both on the get and set access.

While this refactor adds special access handling instead of using the existing `access_set` and `access_get` functions, it fixes the double access and should also make the resulting code more efficient.

### DIFF
--- a/boa_engine/src/bytecompiler/expression/assign.rs
+++ b/boa_engine/src/bytecompiler/expression/assign.rs
@@ -1,9 +1,11 @@
-use boa_ast::expression::operator::{assign::AssignOp, Assign};
-
 use crate::{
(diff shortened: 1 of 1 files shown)


raskad:
> On another note, maybe we should split this function into `compile_simple_assign` and `compile_assign_op`, since they don't share any logic at all.

I thought this too. In the same change I would also refactor the AST to make it a bit typesafer. I'd like to push that in a follow up PR to keep the diff reasonable.

jedel1043 (approved):
Sounds good!

nekevss (approved):
Looks good to me 😄

PR #2553, opened 2023-01-21
[Merged by Bors] - Avoid creating `prototype` property on methods

raskad:

- Stop creating a `prototype` property on class methods by passing a flag to the relevant opcodes.

### DIFF
--- a/boa_engine/src/bytecompiler/class.rs
+++ b/boa_engine/src/bytecompiler/class.rs
@@ -138,6 +138,7 @@ impl ByteCompiler<'_, '_> {
         let index = self.code_block.functions.len() as u32;
         self.code_block.functions.push(code);
         self.emit(Opcode::GetFunction, &[index]);
+        self.emit_u8(0);
 
         self.emit_opcode(Opcode::Dup);
         if let Some(node) = class.super_ref() {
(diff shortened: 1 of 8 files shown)


boa_engine/src/bytecompiler/class.rs
```diff
@@ -138,6 +138,7 @@ impl ByteCompiler<'_, '_> {
         let index = self.code_block.functions.len() as u32;
         self.code_block.functions.push(code);
         self.emit(Opcode::GetFunction, &[index]);
+        self.emit_u8(0);
```
nekevss, line 141: Thoughts on creating an `emit` method like `emit_with_opcode_and_u8` (maybe not the best name) to bind the pushing of the two operands together?

raskad: I think we could maybe generate `emit` functions for every opcode via a macro e.g. `Opcode::GetFunction::emit`. These functions could then automatically take the correct arguments.
Do we want to create a dedicated issue for making bytecode generation more typesafe?

nekevss (approved):
The rest looks good to me :)

boa_engine/src/bytecompiler/class.rs
nekevss: Yeah, that would make sense to me.

jedel1043 (approved):
Looks good!

PR #2554, opened 2023-01-21
[Merged by Bors] - Return the correct value from a statement list

raskad:
When compiling a statement list we need to make sure that the last expression that returns a value is compiled with the `use_expr` flag. Currently we set `use_expr` on the last statement of the statement list. This leads to incorrect returns when the last statement does not return a value. This PR fixes this by looking up the last value returning expression in a statement list  and setting the `use_expr` appropriately.


HalidOdat (approved):

jedel1043 (approved):
Nice!

PR #2555, opened 2023-01-23
Cleanup intrinsics and move to realm

jedel1043:
Small (ish?) step towards having proper realm records 

- Moves `Intrinsics` to `Realm`.
- Cleans up the initialization logic of our intrinsics to not depend on `Context`, unblocking things like #2314.
- Adds hooks to initialize the global object and the global this per the corresponding [`InitializeHostDefinedRealm ( )`](https://tc39.es/ecma262/#sec-initializehostdefinedrealm) hook. Though, this is currently broken because the vm uses `GlobalPropertyMap` instead of the `JsObject` API to initialize global properties.


raskad (approved):
Amazing work!

I really like the new builtin initialization. The separated traits make it clear what exactly the builint object is and how it is used.

nekevss (approved):
Looks good to me! 😄

PR #2562, opened 2023-01-24
[Merged by Bors] - Pin tarpaulin version to 0.22

jedel1043:
Our CI broke because the new version of `tarpaulin` released more assets than the default 'travis.tar.gz'. We'll have to pin our version of `tarpaulin` until the issue gets fixed.


jedel1043:
Thanks!

raskad (approved):

Issue #1907, opened 2022-03-07
Move syntax errors from the compiler to the parser


raskad:
Currently we have some syntax errors that are being thrown in the compiler. They should be already be thrown in the parser.

The best case to search for them is to look for `throw_syntax_error` and `construct_syntax_error` in `boa_engine/src/bytecompiler.rs`. There may be state that must be added during the parsing for all of the errors to be moved to the parsing phase.

After this is fixed, `Context::compile` should not need to return a `Result` anymore.

Razican:
After #2027, what's missing here?

raskad:
There are still some syntax errors in the bytecompiler left:

* Check for undeclared labels at parse time to avoid [this](https://github.com/boa-dev/boa/blob/f0378068a0a2add5b7110addc875aa266f28cc38/boa_engine/src/bytecompiler.rs#L1609)
* Check if `continue` statements occur within a loop to avoid [this](https://github.com/boa-dev/boa/blob/f0378068a0a2add5b7110addc875aa266f28cc38/boa_engine/src/bytecompiler.rs#L1627)
* Check if `break` statements occur within loop or switch to avoid [this](https://github.com/boa-dev/boa/blob/f0378068a0a2add5b7110addc875aa266f28cc38/boa_engine/src/bytecompiler.rs#L1677)
* Check for invalid left-had-side in assignments (all others)

PR #2565, opened 2023-01-26
[Merged by Bors] - Move increment and decrement operations to `Update` expression

raskad:

- Move postfix/prefix increment and decrement operations from the `Unary` expression to a new `Update` expression.
- Add a special type for the `Update` expression target as it is very limited in comparision to an `Unary` target.
- This makes bytecode compilation more typesafe for these operations and removes syntax errors from the bytecompiler without introducing panics (see #1907).

### DIFF
--- a/boa_parser/src/parser/expression/update.rs
+++ b/boa_parser/src/parser/expression/update.rs
@@ -18,7 +18,10 @@ use crate::{
 };
 use boa_ast::{
     expression::{
-        operator::{unary::UnaryOp, Unary},
+        operator::{
+            update::{UpdateOp, UpdateTarget},
+            Update,
+        },
         Identifier,
     },
     Expression, Position, Punctuator,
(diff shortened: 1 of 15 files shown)


boa_parser/src/parser/expression/update.rs
```diff
@@ -56,18 +59,28 @@ impl UpdateExpression {
...
+///
+/// More information:
+///  - [ECMAScript specification][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-static-semantics-assignmenttargettype
+fn as_simple(
+    expr: &Expression,
+    position: Position,
+    strict: bool,
+) -> ParseResult<Option<UpdateTarget>> {
```
jedel1043, line 72: Really liked this change!

jedel1043 (approved):
Looks really nice!

nekevss (approved):

PR #2566, opened 2023-01-26
[Merged by Bors] - fuzzer: bubble up NoInstructionsRemain error instead of trying to handle as exception

Mrmaxmeier:
Hi,

the `vm-implied` fuzzer panics when executing this testcase:

```javascript
try {
    new function() {
        while (this) {}
    }();
} catch {
}
```

`internal error: entered unreachable code: The NoInstructionsRemain native error cannot be converted to an opaque type`

Handling the `NoInstructionsRemain` error upfront instead of going through the VM exception handling logic seems to work.

### DIFF
--- a/boa_engine/src/vm/mod.rs
+++ b/boa_engine/src/vm/mod.rs
@@ -10,7 +10,7 @@ use crate::{
     Context, JsResult, JsValue,
 };
 #[cfg(feature = "fuzz")]
-use crate::{JsError, JsNativeError};
+use crate::{JsError, JsNativeError, JsNativeErrorKind};
 use boa_interner::ToInternedString;
 use boa_profiler::Profiler;
 use std::{convert::TryInto, mem::size_of, time::Instant};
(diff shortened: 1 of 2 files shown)


raskad:
Hi, thanks for the contribution! This indeed seems like undesirable behaviour for the `vm-implied` fuzzer.

I think with the proposed change we would run into infinite loops while fuzzing. Take for example this code:

```js
while (true) {
  try {
    new function() {
        while (true) {}
    }();
  } catch {
  }
}
```

The error would be thrown in `Context::run` but would be catched in the outer try/catch logic and never end the execution.

I'm not 100% sure what the best solution for this problem is. Maybe instead of adding a lot of fuzzing specific logic to the vm, we could keep the panic, catch it in the fuzzer and somehow signal the fuzzer that this was a `NoInstructionsRemain` panic?

Mrmaxmeier:
> The error would be thrown in `Context::run` but would be catched in the outer try/catch logic and never end the execution.

Yes, it looks like some of the callers of `Context::run` (async, generators) will silently consume the error without bubbling it up.
That seems bad, but in practice, I don't think we'll get stuck in an infinite loop. Even if the original `NoInstructionsRemain` error is lost in generator shenannigans, we'll eventually try to execute something else and raise another `NoInstructionsRemain` error.

(Unfortunately there's no better way to do this, but running the fuzzer with your while-try-while testcase hardcoded works/terminates fine.)

raskad:
> That seems bad, but in practice, I don't think we'll get stuck in an infinite loop. Even if the original `NoInstructionsRemain` error is lost in generator shenannigans, we'll eventually try to execute something else and raise another `NoInstructionsRemain` error.

True, with that in mind this should work. @jedel1043 I think you brought up the panic on `NoInstructionsRemain` initially. Do you agree that this change should work or do you see any way in which this could fail?

jedel1043:
I think it should be ok. Though, if any problems arise, we could just make `JsError::to_opaque` throw an error on trying to convert a `JsNativeError::NoInstructionRemain`.

raskad (approved):
Thanks for the contribution!

jedel1043 (approved):
Thanks!

PR #2574, opened 2023-01-27
[Merged by Bors] - Update icu dependencies

raskad:

- Update icu dependencies
- Regenerate icu data


jedel1043 (approved):
Seems like the failing tests are either a bug in ICU4X or a bug in our Collator implementation, so this can be merged as it is and I'll investigate that.

PR #2576, opened 2023-01-29
[Merged by Bors] - Rename `check_parser` and `Identifier`

jedel1043:
Extracted from #2411 to reduce its size a bit.

This PR:
- Renames `Identifier` to `IdentifierName`, which is the name stated in the spec.
- Renames the utility function `check_parser` to `check_script_parser` to prepare for modules.
- Adds some missing `#[inline]` and rewrites some patterns.


Razican (approved):
Thanks!

raskad (approved):

jedel1043:
bors retry

Issue #1760, opened 2021-12-24
Add context information to `not a constructor` `TypeError`


raskad:
There are a few places where we throw a `TypeError` signaling that a value is `not a constructor`. Right now those errors do not contain specific information to find out why or where the error occurred.

- The produced error should contain useful information (variable name(?), type of the value (?)) that makes it easier to find out why the error occurred.
- All of the places where we throw this error should have a similar syntax for the error text.

Razican:
This also happens where we have "not a callable function".

aditsharma14:
Hi maintainers

I’d like to work on #1760 (adding more context to "not a constructor" TypeErrors).

I’ve taken some time to go through the codebase and identified the six locations where these errors are currently thrown. While looking at them, I noticed that the error messages are a bit inconsistent and don’t always include useful context (like the value’s type or name), which can make debugging harder.

Proposed approach

I’m planning to:

1) Add small helper constructors to JsNativeError to standardize how "not a constructor" errors are created.

2) Ensure error messages consistently include:

           a.The value’s type (e.g., number is not a constructor)

           b.The identifier name when available (e.g., myVar (number) is not a constructor)

           c.Clear wording for built-ins like Symbol and BigInt

3)Update all existing throw sites to use the new helpers.

4)Add tests covering the different scenarios to ensure consistent behavior.

The goal is to improve error clarity and consistency without introducing breaking changes or altering semantics.

I already have a working implementation drafted locally and would be happy to open a PR once I get confirmation that this approach aligns with the project direction.

If this sounds good, could the issue be assigned to me?

Thanks!

Issue #1313, opened 2021-06-09
Add line/character numbers to errors


macmv:
Line numbers are only in syntax errors right now. Can we add a `Position` to the `Node` type? I am thinking there would be a wrapper struct (a new `Node` type) which would contain a `Position` and a `NodeType`. The `NodeType` would be the current `Node` enum. This would mean we can store various parsing information with each node, and also produce line/character numbers when we hit undefined variables.

This would also mean that `StatmentList` would store a list of these new nodes, and so would everything else that stores a `Node`. It would add a bit of memory overhead, but other than that it shouldn't have any significant performance impact.

I can start working on this, but I would like some feedback before I go ahead with this. It seems like a large enough change that I don't think a PR with this feature would be merged out of nowhere.

Razican:
This is blocked until #1295 lands.

Razican:
Waiting for #300 to be implemented.

HalidOdat:
Once #4273 is merged this will be unblocked :)

PR #2579, opened 2023-01-30
[Merged by Bors] - Create `Source` to abstract JS code sources

jedel1043:
Slightly related to #2411 since we need an API to pass module files, but more useful for #1760, #1313 and other error reporting issues.

It changes the following:

- Introduces a new `Source` API to store the path of a provided file or `None` if the source is a plain string.
- Improves the display of `boa_tester` to show the path of the tests being run. This also enables hyperlinks to directly jump to the tested file from the VS terminal.
- Adjusts the repo to this change.

Hopefully, this will improve our error display in the future.

### DIFF
--- a/boa_parser/src/source.rs
+++ b/boa_parser/src/source.rs
@@ -15,8 +15,9 @@ pub struct Source<'path, R> {
 }
 
 impl<'bytes> Source<'static, &'bytes [u8]> {
-    /// Creates a new `Source` from any type equivalent to a slice of bytes e.g. [`&str`], [`Vec<u8>`]
-    /// <code>[Box]<[\[u8\]][slice]></code> or a plain slice <code>[&\[u8\]][slice]</code>.
+    /// Creates a new `Source` from any type equivalent to a slice of bytes e.g. [`&str`][str],
+    /// <code>[Vec]<[u8]></code>, <code>[Box]<[\[u8\]][slice]></code> or a plain slice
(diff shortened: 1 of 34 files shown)


boa_parser/src/source.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// # use boa_parser::Source;
+    /// let code = r#"var array = [5, 4, 3, 2, 1];"#;
+    /// let source = Source::from_bytes(code);
+    /// ```
+    ///
+    /// [`&str`]: str
+    /// [slice]: std::slice
```
raskad, line 30:
```suggestion
    /// Creates a new `Source` from any type equivalent to a slice of bytes e.g. [`&str`][str],
    /// <code>[Vec]<[u8]></code>,
    /// <code>[Box]<[\[u8\]][std::slice]></code>
    /// or a plain slice <code>[&\[u8\]][std::slice]</code>.
    ///
    /// # Examples
    ///
    /// ```
    /// # use boa_parser::Source;
    /// let code = r#"var array = [5, 4, 3, 2, 1];"#;
    /// let source = Source::from_bytes(code);
    /// ```
```

raskad (approved):
Just got two suggestions, otherwise this looks good

nekevss (approved):
Looks good to me 😄

PR #2582, opened 2023-02-01
[Merged by Bors] - Implement binary `in` operation with private names

raskad:

- Implement binary `in` operation with private names.
- Adding a separate `BinaryInPrivate` expression in addition to the existing `Binary` expression seems like the best way to implement this in a typesafe manner. Other methods like adding an enum for the `Binary` lhs result in having to make assertions.


nekevss (approved):

jedel1043 (approved):
Looks good!

PR #2584, opened 2023-02-02
[Merged by Bors] - Feature flag on builtins console import

nekevss:
Adding a feature flag to the console import in `builtins`. I think this should fix the failing action on the `Upload docs and run benchmarks` step


jedel1043:
To avoid this happening again, could you modify the Lint (No features) action to `cargo clippy -p boa_engine --no-default-features --all-targets`? This currently doesn't work because the `console` feature is always activated by the `boa_cli` crate.

jedel1043 (approved):
Thanks!

raskad (approved):

PR #2586, opened 2023-02-03
[Merged by Bors] - Implement `WeakSet`

raskad:

- Implement `WeakSet` buildin object.
- Supersedes #2009


boa_engine/src/builtins/weak_set/weak_ordered_set.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub fn delete(&mut self, value: &JsObject) -> bool {
+        self.inner
+            .shift_remove(&WeakSetObject(WeakGc::new(value.inner())))
+    }
+
+    /// Return `true` if an equivalent to value exists in the set.
+    pub fn contains(&self, value: &JsObject) -> bool {
+        self.inner
+            .contains(&WeakSetObject(WeakGc::new(value.inner())))
+    }
```
jedel1043, line 95: I think implementing a special `HashMap` using ephemerons directly in our GC would avoid having to create a `WeakSetObject` on each access.

jedel1043 (commented):
Thanks for the preliminar work! Though, this implementation has the problem of not freeing the inner `EphemeronBox` when the inner value invalidates. I think this could only be solved by creating a `WeakMap` structure with direct support from our GC. The GC would then track every `WeakMap` to free the invalidated boxes from the map.

Anyways, I think we could technically merge this as it is and improve the design later, but let me know what you think first :)

raskad:
> Though, this implementation has the problem of not freeing the inner `EphemeronBox` when the inner value invalidates. I think this could only be solved by creating a `WeakMap` structure with direct support from our GC. The GC would then track every `WeakMap` to free the invalidated boxes from the map.

Yes I was thinking the same thing, but I'm not sure where exactly that would fit in our gc process and how much we would have to adjust it. My first instinct is that we would want to check for any collected weak refs in `WeakMap`s at the end of the sweep phase. The big question to me is, how would we keep track of the `WeakMap`s?

jedel1043:
> The big question to me is, how would we keep track of the `WeakMap`s?

Probably by doing something similar as the strong pointers: a queue of `GcBox<dyn ErasedWeakMap>`s that automatically drops all collected weakmaps. (Though, we could also use something like a `WeakBox<dyn ErasedWeakMap>` queue to avoid having to add more collection logic, but I don't know if having to track weak boxes manually will be any better)

raskad:
@jedel1043 I implemented a `WeakMap` in the GC that is just uses `Gc` and `WeakGc` internally to avoid having much new logic. I think this should fix the leaked set/map keys. I wrote some tests and they pass miri. Let me know if you see any errors with it.
Even if it works, I'm not sure if we want to go with it or if the design should be different.

boa_gc/src/pointers/weak_map.rs
```diff
@@ -0,0 +1,35 @@
+use crate::{Allocator, Finalize, Gc, GcRefCell, Trace, WeakGc};
+use std::collections::HashMap;
+
+/// A map that holds weak references to its keys and is traced by the garbage collector.
+#[derive(Clone, Debug, Default, Trace, Finalize)]
+pub struct WeakMap<K: Trace + Sized + 'static, V: Trace + Sized + 'static> {
+    pub(crate) inner: Gc<GcRefCell<HashMap<WeakGc<K>, V>>>,
```
jedel1043, line 7: While this works, I think it would be better to directly use a `HashSet<Ephemeron<K, V>>` to optimize this, since right now it is using an `Ephemeron<K, ()>` and a `V`.

jedel1043: Hm, realized that to do this we'll probably have to implement our own `HashMap`...

jedel1043: Yeah, this should work for now. Maybe in the future we'll have to inline `HashMap` to optimize the memory of this.

boa_gc/src/internals/weak_map_box.rs
```diff
@@ -0,0 +1,47 @@
...
+/// A box that is used to track [`WeakMap`][`crate::WeakMap`]s.
+pub(crate) struct WeakMapBox<K: Trace + Sized + 'static, V: Trace + Sized + 'static> {
+    pub(crate) map: WeakGc<GcRefCell<HashMap<WeakGc<K>, V>>>,
+    pub(crate) next: Cell<Option<NonNull<dyn ErasedWeakMapBox>>>,
+}
+
+/// A trait that is used to erase the type of a [`WeakMapBox`].
+pub(crate) trait ErasedWeakMapBox {
+    /// Clear dead entries from the [`WeakMapBox`].
+    fn clear_dead_entires(&self);
```
jedel1043, line 13:
```suggestion
    fn clear_dead_entries(&self);
```

jedel1043 (approved):
Looks nice! We can improve this in the future by implementing our own `HashMap` using `Ephemeron`s, so maybe we should open an issue for it?

boa_gc/src/lib.rs
```diff
@@ -234,6 +238,32 @@ impl Allocator {
         })
     }
 
+    fn alloc_weak_map<K: Trace, V: Trace>() -> WeakMap<K, V> {
+        let _timer = Profiler::global().start_event("New WeakGc", "BoaAlloc");
```
nekevss, line 242: The profiler title should probably be "New WeakMap" here, right?

nekevss (commented):
This is looking pretty good.

boa_gc/src/lib.rs
raskad: Good catch, fixed

nekevss (approved):
Looks good to me. Nice work on this!

PR #2598, opened 2023-02-12
[Merged by Bors] - Remove Syntax Errors from Bytecompiler

### LINKED ISSUE #1907: Move syntax errors from the compiler to the parser

raskad:
This Pull Request closes #1907.

It changes the following:

- Implement several early errors relating to labels, `break` and `continue` in the parser.
- Implement an early error for invalid cover grammar of object literals in the parser.
- Remove all remaining syntax errors from the bytecompiler.

### DIFF
--- a/boa_ast/src/operations.rs
+++ b/boa_ast/src/operations.rs
@@ -849,6 +849,49 @@ pub fn class_private_name_resolver(node: &mut Class, top_level_class_index: usiz
     visitor.visit_class_mut(node).is_continue()
 }
 
+/// Errors that can occur when checking labels.
+#[derive(Debug, Clone, Copy)]
+pub enum CheckLabelsError {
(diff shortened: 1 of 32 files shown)


jedel1043 (changes requested):
Good work!

HalidOdat (approved):
Looks good to me!

jedel1043 (approved):
Looks good to me!

PR #2615, opened 2023-02-21
[Merged by Bors] - Keep Integer type for inc/dec of an integer

tunz:
Keep Integer type after running unary increment/decrement ops if possible. The integer type will be useful to take fast paths for accessing index properties. (e.g., `let a = 1; arr[++a]`)

### DIFF
--- a/boa_engine/src/vm/opcode/unary_ops/decrement.rs
+++ b/boa_engine/src/vm/opcode/unary_ops/decrement.rs
@@ -1,5 +1,5 @@
 use crate::{
-    value::Numeric,
+    value::{JsValue, Numeric},
     vm::{opcode::Operation, ShouldExit},
     Context, JsBigInt, JsResult,
 };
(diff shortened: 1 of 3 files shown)


HalidOdat (approved):
Looks good! Thanks for the contribution @tunz

boa_engine/src/vm/opcode/unary_ops/decrement.rs
```diff
@@ -17,11 +17,16 @@ impl Operation for Dec {
 
     fn execute(context: &mut Context<'_>) -> JsResult<ShouldExit> {
         let value = context.vm.pop();
-        match value.to_numeric(context)? {
-            Numeric::Number(number) => context.vm.push(number - 1f64),
-            Numeric::BigInt(bigint) => {
-                context.vm.push(JsBigInt::sub(&bigint, &JsBigInt::one()));
+        match value {
+            JsValue::Integer(number) if number > i32::MIN => {
```
nekevss, line 21: I like the idea of this optimization. Just had one question regarding this, was there a reason to not add `JsValue::BigInt` as well?

tunz: It was more like I couldn't find a reason to add BigInt here. I didn't see any fast path for BigInt when converting it to a property key, and I thought handling BigInt in the next `match` after calling `to_numeric` is not too slow.

nekevss (approved):
Looks good to me! Thanks for the contribution 😄

Issue #1542, opened 2021-08-31
Implement `CompletionRecord` type


jedel1043:
Related discussion on #1541 

The [`Completion Record`](https://tc39.es/ecma262/#sec-completion-record-specification-type) type should live in the `exec` module, inside `mod.rs` if the implementation is simple enough or in a standalone file otherwise.

This feature is being developed in the https://github.com/boa-dev/boa/tree/feature/completion_records branch.

Razican:
Should it be part of the AST? I would say this is only related to execution, since it's the completion of a execution, right?

jedel1043:
> Should it be part of the AST? I would say this is only related to execution, since it's the completion of a execution, right?

Hmm... Yeah, you're right, it should be on the `exec` module.

jedel1043:
Closing in favour of the current implementation. We can revisit this if we reassess that a more spec-compliant design is required.

Issue #1544, opened 2021-08-31
Obtain the `CompletionRecord` of every AST node evaluation


jedel1043:
Related discussion on #1541.

Follows from #1543.

We need to extensively search the [language evaluation spec](https://tc39.es/ecma262/#sec-ecmascript-language-source-code) to find instances of `CompletionRecord`s that should be returned.

This feature is being developed in the https://github.com/boa-dev/boa/tree/feature/completion_records branch.

jedel1043:
Closing in favour of the current implementation. We can revisit this if we reassess that a more spec-compliant design is required.

Issue #1543, opened 2021-08-31
Make `Executable::run` return a `JsResult<CompletionRecord>`


jedel1043:
Related discussion on #1541.

Follows from #1542.

For starters we can just always pass a `CompletionRecord` with a `[[value]]` equal to the value returned by every node evaluation, as all nodes return a `JsValue` with few exceptions like `Throw`.

This feature is being developed in the https://github.com/boa-dev/boa/tree/feature/completion_records branch.

jedel1043:
Closing in favour of the current implementation. We can revisit this if we reassess that a more spec-compliant design is required.

PR #2618, opened 2023-02-26
[Merged by Bors] - Implement `CompletionRecords` for the Vm

nekevss:
This goal of this draft is to begin to implement `CompletionRecords`. This draft switches the `Vm`'s `Opcode` from `JsResult<ShouldExit>` to `CompletionType` where the value is the top of the stack and returns a `CompletionRecord`. Primarily submitting as a draft for now for feedback. 

It changes the following:

- Adds `CompletionRecord` and `CompletionType` to `Vm`.
- Changes the evaluation loop from `while` to `loop` that returns a `CompletionType`.
- Adapts `Context::run` to handle `CompletionType` and `CompletionRecord`.
- Removes `FinallyAddresses` in favor of just checking the `env_stack`.


jedel1043:
Nice work! Does this close #1542, #1543 and #1544? If that's the case, it would be good to link them :)

nekevss:
> Does this close #1542, #1543 and #1544?

I think it does in a way. It's a bit hard to say as I think those issues may predate the move to the VM.

It does have `Context::run` return a `CompletionRecord`, but it is only a `CompletionRecord and not a `JsResult<CompletionRecord>`, which is one of the main reasons I submitted this as a draft first for feedback. Also, the `Operations` trait would now return a `CompletionType` where the top of the stack would essentially be the `[[value]]` field.

Essentially, this should allow us to use `JsResult<JsValue>` for any method or operation that returns only a Normal or Throw Completion. But for the operations or value that can return an abrupt completion, we should be able to use the `CompletionRecord`.

HalidOdat:
Looks good! Great work @nekevss , Will try to do a more thorough review tomorrow! :)

HalidOdat (approved):
Looks good to me! :)

raskad (approved):
I like the new logic.
If @jedel1043 agrees and you want to merge this as is lets do it @nekevss.

boa_engine/src/builtins/async_generator/mod.rs
```diff
@@ -527,7 +527,7 @@ impl AsyncGenerator {
         drop(generator_context_mut);
 
         // 8. Assert: result is never an abrupt completion.
-        assert!(result.is_ok());
+        assert!(!result.is_throw_completion());
```
jedel1043, line 530: `Return` is also abrupt, so `is_throw_completion` should probably be `is_abrupt_completion`

boa_engine/src/vm/completion_record.rs
```diff
@@ -0,0 +1,35 @@
+//! An implementation of a `CompletionRecord` for Boa's VM.
+
+use crate::{JsError, JsResult, JsValue};
+
+/// An implementation of the ECMAScript's `CompletionRecord` [specification] for
+/// Boa's VM output Completion and Result.
+///
+/// [specification]: https://tc39.es/ecma262/#sec-completion-record-specification-type
+#[derive(Debug, Clone)]
+pub(crate) enum CompletionRecord {
```
jedel1043, line 10: I think this should also be used in the `AsyncGenerator` logic.
More specifically, in `AsyncGeneratorRequest`, which has a `completion` field.

Maybe it could also be used in `Generator`, but I think we would need to unify the `generator_resume` and `generator_resume_abrupt` methods into one.

jedel1043 (changes requested):
Good work! The changes look almost perfect :)

boa_engine/src/vm/completion_record.rs
nekevss: There are a few places that I think we should be able to use `CompletionRecord`. The primary place being generators, but I hadn't initially made the change on this draft as there may be more changes that are needed to generator execution to adapt for the shift, which seemed like it would be really expanding the scope of these changes.

jedel1043: Completely understandable. We can fix that in another PR

boa_engine/src/builtins/async_generator/mod.rs
nekevss: Right now, the main use of `CompletionRecord::Return` in this PR is to signal a `Yield`, which would typically be a normal completion. So in this essence, `Throw` is the only true `abrupt completion` currently.

I did try to make the changes, but it started to break more tests in the test262 suite, which is a bit hard to determine whether it is related to `CompletionRecord` semantics that we aren't accounting for or `Generator` and `AsyncGenerator` execution.

jedel1043: Hm, is that ideal? Or should we try to align with the spec?
If this is the ideal behaviour, then we can just merge this as it is, but if it's not then it would be good to open an issue for it.

nekevss: I think the overarching structure here is ideal/working towards it. But if we're referring to the current implementation of `Yield`, then I'm not 100% sure. This mostly begins to implement `CompletionRecord` while maintaining the current behavior of the VM (i.e., returning `ReturnType::Normal` and separate flag for `ReturnType::Yield`).

jedel1043: Then let's just merge this and open an issue called "reevaluate the execution semantics of `CompletionRecord`" or something.

jedel1043 (approved):
Looks great to me!

PR #2619, opened 2023-02-27
[Merged by Bors] - Migrate tests to new test API

jedel1043:
This PR migrates our entire test suite to our new testing API.

It changes the following:

- Migrates tests to new test API.
- Cleans up the API to be a bit more descriptive and maintainable.
- Prettifies our test failure display to show the failed scripts.
- Splits our massive `tests.rs` file into smaller sub-suites.

Example output of a failing test:

image


boa_engine/src/lib.rs
```diff
@@ -182,85 +182,268 @@ impl JsArgs for [JsValue] {
...
-) -> JsResult<JsValue> {
-    use boa_profiler::Profiler;
+impl TestAction {
+    /// Evaluates some utility functions used in tests.
+    pub(crate) const fn run_harness() -> Self {
+        Self(Inner::RunHarness)
+    }
 
-    let main_timer = Profiler::global().start_event("Main", "Main");
+    /// Runs `source`, panickinf if the execution throws.
```
jasonwilliams, line 235: "panicking"

boa_engine/src/lib.rs
```diff
@@ -182,85 +182,268 @@ impl JsArgs for [JsValue] {
...
+    pub(crate) fn run(source: impl Into<Cow<'static, str>>) -> Self {
```
jasonwilliams, line 236: What's the reason for using `Cow` string type here instead of just the static string?

boa_engine/src/vm/tests.rs
```diff
@@ -1,33 +1,38 @@
...
+use indoc::indoc;
 
 #[test]
 fn typeof_string() {
-    let typeof_object = r#"
-        const a = "hello";
-        typeof a;
-    "#;
-    assert_eq!(&exec(typeof_object), "\"string\"");
+    run_test([TestAction::assert_eq(
```
jasonwilliams, line 6: Most of these tests are only passing 1 item in the array, do we want an API which can handle just a single test?
The name run_test doens't sound right when it's designed to take in an array of tests. But I understand this is just bikeshedding.

jasonwilliams:
Looks good so far, but improvement to what we had before!

boa_engine/src/vm/tests.rs
jedel1043: `run_test_actions` maybe?

jasonwilliams: I think that name would be describing better what is happening yeah. (Even though it is longer)

boa_engine/src/lib.rs
jedel1043: There are some tests that programmatically generate the tested scripts:
https://github.com/boa-dev/boa/blob/6c0482ecb3b741bb3931087732983680dc9184f3/boa_engine/src/tests/mod.rs#L429-L431
The other option was to take a `String` instead of a `&'static str`, but that would make almost all tests slower, so I opted for taking a `Cow` instead.

boa_engine/src/vm/tests.rs
jedel1043: > Most of these tests are only passing 1 item in the array, do we want an API which can handle just a single test?

Thought about it, but I didn't think it would be worth increasing the complexity of the API just to save two (`[` and `]`) keystrokes...

... but now that I think about it, we could implement `IntoIterator` for `TestAction`, which doesn't increase our API complexity, though it makes `run_test_actions(TestAction::exec(..))` look a bit weird.

HalidOdat (approved):
The tests look much cleaner now! Great work @jedel1043

raskad (approved):
Nice improvement!

jedel1043:
Gonna merge this since I've already resolved Jason's comments, and It'll be better to merge this ASAP to avoid merge conflicts.

PR #2656, opened 2023-03-13
[Merged by Bors] - [regexp] new tests for unicode flag

selfisekai:
the flag got introduced in regress 0.5.0

This Pull Request fixes/closes: n/a.

It changes the following:

- add test for unicode flag
- add test for native error on failing to parse (got removed in #2651 because the previous example parsing correctly now)


raskad (approved):
Nice addition. Only needs a `cargo fmt` and then we are good.

Thanks for the contribution :)

jedel1043 (approved):
Thanks!

PR #2669, opened 2023-03-15
Decouple bytecompiler from CodeBlock

HalidOdat:
This hopefully a PR in series of PRs to implement a bytecode optimizer, before that can happen there needs to be a lot of refactoring in the way we store and compile it.

This also give us some memory benefits, it reduces `CodeBlock` size from `264` **=>** `208` (removes `56` bytes).

Additionally when calling `into_boxed_slice`, If the vector has excess capacity, its items will be moved into a newly-allocated buffer with exactly the right capacity removing wasted space.


boa_engine/src/bytecompiler/mod.rs
```diff
@@ -473,10 +549,10 @@ impl<'b, 'host> ByteCompiler<'b, 'host> {
         let index = index as usize;
 
         let bytes = target.to_ne_bytes();
-        self.code_block.bytecode[index + 1] = bytes[0];
-        self.code_block.bytecode[index + 2] = bytes[1];
-        self.code_block.bytecode[index + 3] = bytes[2];
-        self.code_block.bytecode[index + 4] = bytes[3];
+        self.bytecode[index + 1] = bytes[0];
```
Razican, line 552: At some point we should check if this adds bound checking, and if we can do the bounds checking only once. Might enhance performance slightly.

boa_engine/src/vm/code_block.rs
```diff
@@ -129,21 +129,21 @@ impl CodeBlock {
     #[must_use]
     pub fn new(name: Sym, length: u32, strict: bool) -> Self {
         Self {
-            bytecode: Vec::new(),
-            literals: Vec::new(),
-            names: Vec::new(),
-            private_names: Vec::new(),
-            bindings: Vec::new(),
+            bytecode: Vec::new().into_boxed_slice(),
```
Razican, line 132: Isn't there a `Box::default()`?

Razican (approved):
All looks good :) just a couple of comments. Thanks!

raskad (approved):

boa_engine/src/bytecompiler/mod.rs
HalidOdat: From my checking it seems that it does generate a bounds check per array access, my guess is that so it can give accurate panic locations if out-of-bounds

See godbolt assembly instructions: <https://godbolt.org/z/q8azssz5K>

HalidOdat: Explicitly checking beforehand and using a little unsafe we can remove the bounds checks: <https://godbolt.org/z/bKT9ozr8s>

Razican: Sometimes in Rust, if you check the full range before all the array access with an assert, it will remove all bounds checking without needing "unsafe". We would need to test that :)

HalidOdat: Yeah, but through by testing it doesn't seem so do so <https://godbolt.org/z/q15WGh9n8> in this case...

PR #2679, opened 2023-03-16
[Merged by Bors] - Implement constant folding optimization

HalidOdat:
This PR implements an optimizer, It currently implements the [constant folding optimization][cfo]. this optimization is responsible for "folding"/evaluating constant expressions.

For example:

```js
let x = ((1 + 2 + -4) * 8) << 4
```
Generates the following instruction(s) (`cargo run -- -t`):
```
000000    0000    PushOne
000001    0001    PushInt8                   2
000003    0002    Add
000004    0003    PushInt8                   4
000006    0004    Neg
000007    0005    Add
000008    0006    PushInt8                   8
000010    0007    Mul
000011    0008    PushInt8                   4
000013    0009    ShiftLeft
000014    0010    DefInitLet                 0000: 'x'
```

With constant folding it generates the following instruction(s) (`cargo run -- -t -O`):
```
000000    0000    PushInt8                   -128
000002    0001    DefInitLet                 0000: 'x'
```

It changes the following:
- Implement ~~WIP~~ constant folding optimization, ~~only works with integers for now~~
- Add `--optimize, -O` flag to boa_cli
- Add `--optimizer-statistics` flag to boa_cli for optimizer statistics
- Add `--optimize, -O` flag to boa_tester

After I finish with this, will try to implement other optimizations :)

[cfo]: https://en.wikipedia.org/wiki/Constant_folding

### DIFF
--- a/boa_ast/src/expression/literal/mod.rs
+++ b/boa_ast/src/expression/literal/mod.rs
@@ -110,6 +110,12 @@ pub enum Literal {
     /// [spec]: https://tc39.es/ecma262/#sec-null-value
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Glossary/null
     Null,
+
+    /// This represents the JavaScript `undefined` value, it does not reference the `undefined` global variable,
+    /// it will directly evaluate to `undefined`.
+    ///
+    /// NOTE: This is used for optimizations.
+    Undefined,
 }
 
 impl From<Sym> for Literal {
@@ -173,6 +179,7 @@ impl ToInternedString for Literal {
             Self::BigInt(ref num) => num.to_string(),
             Self::Bool(v) => v.to_string(),
             Self::Null => "null".to_owned(),
+            Self::Undefined => "undefined".to_owned(),
         }
     }
 }
--- a/boa_engine/src/context/mod.rs
+++ b/boa_engine/src/context/mod.rs
@@ -23,6 +23,7 @@ use crate::{
     job::{IdleJobQueue, JobQueue, NativeJob},
     native_function::NativeFunction,
     object::{FunctionObjectBuilder, GlobalPropertyMap, JsObject},
+    optimizer::{Optimizer, OptimizerOptions, OptimizerStatistics},
     property::{Attribute, PropertyDescriptor, PropertyKey},
     realm::Realm,
     vm::{CallFrame, CodeBlock, Vm},
@@ -101,6 +102,8 @@ pub struct Context<'host> {
     host_hooks: &'host dyn HostHooks,
 
     job_queue: &'host dyn JobQueue,
+
+    optimizer_options: OptimizerOptions,
 }
 
 impl std::fmt::Debug for Context<'_> {
@@ -113,7 +116,8 @@ impl std::fmt::Debug for Context<'_> {
             .field("vm", &self.vm)
             .field("strict", &self.strict)
             .field("promise_job_queue", &"JobQueue")
-            .field("hooks", &"HostHooks");
+            .field("hooks", &"HostHooks")
+            .field("optimizer_options", &self.optimizer_options);
 
         #[cfg(feature = "intl")]
         debug.field("icu", &self.icu);
@@ -202,6 +206,15 @@ impl Context<'_> {
         result
     }
 
+    /// Applies optimizations to the [`StatementList`] inplace.
+    pub fn optimize_statement_list(
+        &mut self,
+        statement_list: &mut StatementList,
+    ) -> OptimizerStatistics {
+        let mut optimizer = Optimizer::new(self);
+        optimizer.apply(statement_list)
+    }
+
     /// Parse the given source script.
     pub fn parse_script<R: Read>(
         &mut self,
@@ -212,7 +225,11 @@ impl Context<'_> {
         if self.strict {
             parser.set_strict();
         }
-        parser.parse_script(&mut self.interner)
+        let mut result = parser.parse_script(&mut self.interner)?;
+        if !self.optimizer_options().is_empty() {
+            self.optimize_statement_list(&mut result);
+        }
+        Ok(result)
     }
 
     /// Parse the given source script.
@@ -425,6 +442,15 @@ impl Context<'_> {
     #[cfg(feature = "trace")]
     pub fn set_trace(&mut self, trace: bool) {
         self.vm.trace = trace;
+    }
+
+    /// Get optimizer options.
+    pub const fn optimizer_options(&self) -> OptimizerOptions {
+        self.optimizer_options
+    }
+    /// Enable or disable optimizations
+    pub fn set_optimizer_options(&mut self, optimizer_options: OptimizerOptions) {
+        self.optimizer_options = optimizer_options;
     }
 
     /// Changes the strictness mode of the context.
@@ -643,6 +669,7 @@ impl<'icu, 'hooks, 'queue> ContextBuilder<'icu, 'hooks, 'queue> {
             kept_alive: Vec::new(),
             host_hooks,
             job_queue: self.job_queue.unwrap_or(&IdleJobQueue),
+            optimizer_options: OptimizerOptions::OPTIMIZE_ALL,
         };
 
         builtins::set_default_global_bindings(&mut context)?;
--- a/boa_engine/src/optimizer/mod.rs
+++ b/boa_engine/src/optimizer/mod.rs
@@ -1,0 +1,135 @@
+//! Implements optimizations.
+
+pub(crate) mod pass;
+pub(crate) mod walker;
+
+use self::{pass::ConstantFolding, walker::Walker};
+use crate::Context;
+use bitflags::bitflags;
+use boa_ast::{visitor::VisitorMut, Expression, StatementList};
+use std::{fmt, ops::ControlFlow};
+
+bitflags! {
+    /// Optimizer options.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+    pub struct OptimizerOptions: u8 {
+        /// Print statistics to `stdout`.
+        const STATISTICS = 0b0000_0001;
+
+        /// Apply contant folding optimization.
+        const CONSTANT_FOLDING = 0b0000_0010;
+
+        /// Apply all optimizations.
+        const OPTIMIZE_ALL = Self::CONSTANT_FOLDING.bits();
+    }
+}
+
+/// The action to be performed after an optimization step.
+#[derive(Debug)]
+pub(crate) enum PassAction<T> {
+    /// Keep the node, do nothing.
+    Keep,
+
+    /// The node was modified inplace.
+    Modified,
+
+    /// Replace the node.
+    Replace(T),
+}
+
+/// Contains statistics about the optimizer execution.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct OptimizerStatistics {
+    /// How many times was the optimization run in total.
+    pub constant_folding_run_count: usize,
+
+    /// How many passes did the optimization run in total.
+    pub constant_folding_pass_count: usize,
+}
+
+impl fmt::Display for OptimizerStatistics {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        writeln!(f, "Optimizer {{")?;
+        writeln!(
+            f,
+            "    constant folding: {} run(s), {} pass(es) ({} mutating, {} checking)",
+            self.constant_folding_run_count,
+            self.constant_folding_pass_count,
+            self.constant_folding_pass_count
+                .saturating_sub(self.constant_folding_run_count),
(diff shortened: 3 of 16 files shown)


HalidOdat:
I enabled optimizations by default for testing, so that's why there is an improvement in benchmarks (especially 50% improvement in arithmetic execution), there also seems to be some improvement in byte compiling probably because it has to generate less.

Optimizations should probably be disabled when benchmarking

boa_ast/src/expression/literal/mod.rs
```diff
@@ -110,6 +110,16 @@ pub enum Literal {
...
+
+    /// TODO: doc
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-undefined-value
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Glossary/undefined
+    Undefined,
```
jedel1043, line 122: Uh we had to remove this because `undefined` is a valid identifier e.g. this compiles and prints `5`:
```javascript
{
    let undefined = 5;
    console.log(undefined); // 5
}
```

jedel1043 (changes requested):

boa_engine/src/optimizer/constant_folding_optimizer.rs
```diff
@@ -0,0 +1,152 @@
...
+            return PassResult::Leave;
+        };
+        let literal = match unary.op() {
+            UnaryOp::Minus => Literal::Int(-value),
+            UnaryOp::Plus => Literal::Int(value),
+            UnaryOp::Not => Literal::Bool(value == 0),
+            UnaryOp::Tilde => Literal::Int(!value),
+            UnaryOp::TypeOf => Literal::String(interner.get_or_intern("number")),
+            UnaryOp::Delete => Literal::Bool(true),
+            UnaryOp::Void => Literal::Undefined,
```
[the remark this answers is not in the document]
HalidOdat: Currently I'm only constant folding integer literals, and as mentioned in the previous reviews, `Literal::Undefined` is the an actual `undefined` value, not `undefined` global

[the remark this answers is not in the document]
jedel1043: I may not be getting something, but wouldn't that cause incorrect foldings such as:
```javascript
{
  let a = 5;
  let undefined = 10;
  console.log(void a); // prints 10, should print 5
}
```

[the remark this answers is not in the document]
jedel1043: Also, if you need to refer to the `undefined` value, you can add it to the list of static `Sym`s and refer to it with `Expression::Identifier(Sym::UNDEFINED.into())`.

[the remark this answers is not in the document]
HalidOdat: That is `Constant propagation` it is a another optimization, the are often grouped together but there different, `constant folding` optimization only deals with constant expressions.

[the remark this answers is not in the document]
HalidOdat: > Also, if you need to refer to the `undefined` value, you can add it to the list of static `Sym`s and refer to it with `Expression::Identifier(Sym::UNDEFINED.into())`.

But this gets the value from the global or local that can change, and `void 0` will always give undefined, irrelevant of other state, right?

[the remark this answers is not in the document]
jedel1043: > But this gets the value from the global or local that can change, and `void 0` will always give undefined, irrelevant of other state, right?

Yep. I was just mentioning another way to get the `undefined` binding without adding an `Undefined` literal.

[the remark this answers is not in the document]
jedel1043: > Yes, that is true, I was going to try to implement this after this optimization, since you simplify/fold the expression then you inline/propagate the constants

I mean that this current optimization will convert things such as:
```javascript
{
  let undefined = 10;
  console.log(void 5); // undefined
}
```
to
```javascript
{
  let undefined = 10;
  console.log(undefined); // 10
}
```
since `void 5` is in an expression node and constant folding will substitute it with the `undefined` binding, right?

I'm arguing that this is incorrect because it should instead evaluate to `JsValue::Undefined`, which can only be ensured if we used `void 0` directly:
```javascript
{
  let undefined = 10;
  console.log(void 0); // undefined
}
```

But now that I'm thinking it more thoroughly, wouldn't it be fine to leave the `void <integer>` as it is? Using `0` or `37373` as the rhs of the `void` operation shouldn't matter anyways.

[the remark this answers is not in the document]
HalidOdat: > since `void 5` is an expression node and constant folding will substitute it with the `undefined` binding, right?

It actually substitutes to `Literal::Undefined` witch is [mapped to `PushUndefined` opcode](https://github.com/boa-dev/boa/pull/2679/files#diff-620dfaf5e8f600ab363a3fa0d28280ed8946f91d36f86ed338f05c54cb2b19c2R36) and it pushes `JsValue::undefined()` directly, no binding lookup.

> But now that I'm thinking it more thoroughly, wouldn't it be fine to leave the `void <integer>` as it is? Using `0` or `37373` as the operand of the `void` operation shouldn't matter anyways.

It could be left like that but it would make folding more difficult also generate unneeded instructions.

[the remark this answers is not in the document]
jedel1043: > It actually substitutes to `Literal::Undefined` witch is [mapped to `PushUndefined` opcode](https://github.com/boa-dev/boa/pull/2679/files#diff-620dfaf5e8f600ab363a3fa0d28280ed8946f91d36f86ed338f05c54cb2b19c2R36) and it pushes `JsValue::undefined()` directly, no binding lookup.

That's the detail I was missing! Thanks for explaining it :)

boa_ast/src/expression/literal/mod.rs
jedel1043: We'll have to remember that this can only be used for optimization purposes and not for parsing the `undefined` token. Can you add a comment explaining it?

Razican:
> Optimizations should probably be disabled when benchmarking

Hmmm it depends what do we want to benchmark. I would like to see optimized vs non-optimized end-to-end benchmarking of real-life code.

An arithmetic expression that is compile-time known will of course be optimized away a lot.

boa_engine/src/context/mod.rs
```diff
@@ -202,6 +205,15 @@ impl Context<'_> {
         result
     }
 
+    /// Applies optimizations to the [`StatementList`] inplace.
+    pub fn optimize_statement_list(&mut self, statement_list: &mut StatementList) {
+        let mut optimizer = Optimizer::new(self);
+        let mut cfo = ConstantFoldingOptimizer::default();
+        optimizer.push_pass(&mut cfo);
```
jedel1043, line 212: Maybe it would be better to have `ConstantFoldingOptimizer` inside `Optimizer`, since it abstracts the `optimize` action from the optimization types and also makes it easier to add more optimizations in the future.

.github/workflows/test262.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         run: |
           cd boa
           mkdir ../results
-          cargo run --release --bin boa_tester -- run -v -o ../results/test262
+          cargo run --release --bin boa_tester -- run -O -v -o ../results/test262
```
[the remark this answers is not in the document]
HalidOdat: The enabled optimization in test262 is just for testing for now.

> At least on any optimizer updates, should we run the tests on both optimized and non-optimized in our CI to insure that an update to one doesn't break something in another?

Yes, completely agree! maybe we can configure to start test262 with optimizations when change to the `optimizer` module/directory happens (maybe reviewpad mentioned by @Razican  can be used for this, since it has a nice way of [checking if a directory contents are changed](https://docs.reviewpad.com/guides/built-ins/#hasfilepattern)). And we should probably run it on `main` push and assert that the non-optimized and optimized are equal, just to be sure :)

HalidOdat:
This is now ready for review/merge :) :tada: 

Also implemented statistics for optimizations which will be printed with the `--optimizer-statistics`, right now only  tracking how many times a it ran and how many passes where in total, will be more interesting when we implement other optimizations.
```bash
cargo run -- -O --optimizer-statistics
```
```js
>>> (1 + 2 * 3) + '00'
Optimizer {
    constant folding: 1 run(s), 2 pass(es) (1 mutating, 1 checking)
}
```

boa_engine/src/optimizer/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+        // TODO: Add more optimizations here...
+        // loop {
+        //     let mut changed = false;
+        //     changed |= self.run_constant_folding_pass(expr);
+        //
+        //     if !changed {
+        //         break;
+        //     }
+        // }
+        self.run_constant_folding_pass(expr);
```
jedel1043, line 99: About the optimizer pass architecture, maybe we could copy `dfdx`'s design of [`Module`](https://docs.rs/dfdx/latest/dfdx/nn/trait.Module.html), where a trait is implemented for the `(P1, P2, P3, P4)` tuple with `P1/P2/P3/P4: OptimizationPass`. That way, if we wanted to add a new optimization pass `Pass4` we just implement `OptimizationPass` for it, then we'd just need to change something like `(pass1, pass2, pass3).run(expr)` into `(pass1, pass2, pass3, pass4).run(expr)`.

HalidOdat: Hmmm. interesting! Will look into it! :)

raskad (approved):
Great work!
We should definitely adjust our ci and tests for optimized / unoptimized code, but I would say lets merge this.

boa_engine/src/context/mod.rs
```diff
@@ -101,6 +102,9 @@ pub struct Context<'host> {
     host_hooks: &'host dyn HostHooks,
 
     job_queue: &'host dyn JobQueue,
+
+    optimize: bool,
```
nekevss, line 106: Should we look at narrowing this down to a bitflag eventually?

nekevss (commented):
This looks pretty good overall! Just a couple comments. 😄

boa_engine/src/context/mod.rs
HalidOdat: Maybe, I also though of a having a struct `OptimizerOptions` where we can enable or disable certain optimizations (should make much easier to debug optimizations and see how they perform, since they can interfere with each other, that why they are run one after the other)

Maybe have a CLI flag comma separated list `--optimize=constant-folding,constant-propagation,dead-code-elimination` by default enabling all, if nothing specified

Or maybe I should leave this for when we have more than one optimization? :sweat_smile: 

What do you think about this? @nekevss

boa_engine/src/optimizer/mod.rs
HalidOdat: For this PR since it's only one optimization (seems unnecessary), I'll leave it as is for now, when we implement other optimizations we can see how well this architecture fits :)

boa_engine/src/context/mod.rs
HalidOdat: > Should we look at narrowing this down to a bitflag eventually?

I put it in bitflag in the latest commit :)

boa_engine/src/context/mod.rs
```diff
@@ -642,6 +668,7 @@ impl<'icu, 'hooks, 'queue> ContextBuilder<'icu, 'hooks, 'queue> {
             kept_alive: Vec::new(),
             host_hooks,
             job_queue: self.job_queue.unwrap_or(&IdleJobQueue),
+            optimizer_options: OptimizerOptions::empty(),
```
HalidOdat, line 671: Should it be opt-in, or opt-out?

boa_engine/src/context/mod.rs
nekevss: I like it! There's probably more things we could consider in the future like a `default` method that sets all opt-out flags. But overall, this looks good to me 😄

Razican:
Looking at the benchmarks, it seems that opting-out makes things slower than before

HalidOdat:
> Looking at the benchmarks, it seems that opting-out makes things slower than before

I disabled optimizations in the benchmarks.

The benchmarks don't seem right, benchmarks like `Create Realm` have a +13.00%  :confused: , it's not even related to this PR.

The opt-out only adds one condition check that is called once at the top level (parse_script function), its not something to cause that big of change in the benchmarks.

Razican (approved):
Looks good! :) great way to start optimizing

Razican:
@jedel1043 are your concerns solved?

jedel1043 (approved):
> @jedel1043 are your concerns solved?

Yep, no more comments from my part. We can merge this :)

PR #2680, opened 2023-03-16
[Merged by Bors] - Avoid unneeded bounds checks in bytecode address patching

HalidOdat:
As discussed in this comment https://github.com/boa-dev/boa/pull/2669#discussion_r1137618027,
`rustc` doesn't seem to optimize out the bounds checks.

### DIFF
--- a/boa_engine/src/bytecompiler/mod.rs
+++ b/boa_engine/src/bytecompiler/mod.rs
@@ -547,12 +547,17 @@ impl<'b, 'host> ByteCompiler<'b, 'host> {
         let Label { index } = label;
 
         let index = index as usize;
-
         let bytes = target.to_ne_bytes();
-        self.bytecode[index + 1] = bytes[0];
-        self.bytecode[index + 2] = bytes[1];
-        self.bytecode[index + 3] = bytes[2];
-        self.bytecode[index + 4] = bytes[3];
+
+        // This is done to avoid unneeded bounds checks.
+        assert!(self.bytecode.len() > index + std::mem::size_of::<u32>());
+        // SAFETY: We already checked that it is not out-of-bounds.
+        unsafe {
+            *self.bytecode.get_unchecked_mut(index + 1) = bytes[0];
+            *self.bytecode.get_unchecked_mut(index + 2) = bytes[1];
+            *self.bytecode.get_unchecked_mut(index + 3) = bytes[2];
+            *self.bytecode.get_unchecked_mut(index + 4) = bytes[3];
+        }
     }
 
     fn patch_jump(&mut self, label: Label) {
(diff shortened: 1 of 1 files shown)


jedel1043:
Got it to optimize the bounds check without using unsafe: https://godbolt.org/z/jdczMjW1c

boa_engine/src/bytecompiler/mod.rs
```diff
@@ -544,15 +544,16 @@ impl<'b, 'host> ByteCompiler<'b, 'host> {
...
-
         let bytes = target.to_ne_bytes();
-        self.bytecode[index + 1] = bytes[0];
-        self.bytecode[index + 2] = bytes[1];
-        self.bytecode[index + 3] = bytes[2];
-        self.bytecode[index + 4] = bytes[3];
+
+        // This is done to avoid unneeded bounds checks.
+        assert!(self.bytecode.len() > index + U32_SIZE && usize::MAX - U32_SIZE >= index);
+        self.bytecode[index..index + U32_SIZE].copy_from_slice(bytes.as_slice());
```
nekevss, line 556: Wouldn't we want index + 1 here?

HalidOdat: Yup, its offset by one, that's why all the tests are failing :sweat_smile:

jedel1043: Oops, I directly optimized https://godbolt.org/z/q15WGh9n8 without considering the original context from where it was taken 😅

jedel1043: Should be
```javascript
const U32_SIZE: usize = std::mem::size_of::<u32>();
assert!(code.len() > index + U32_SIZE && usize::MAX - U32_SIZE >= index);

let values = value.to_ne_bytes();
code[index + 1..=index + U32_SIZE].copy_from_slice(values.as_slice())
```
then

nekevss: Would it be more clear to put something like:

```rust
let slice_index = index + 1;
code[slice_index..slice_index + U32_SIZE].copy_from_slice(values.as_slice())
```

Adding the extra line may not be necessary, but it may help with readability down the line.

jedel1043: Maybe `U32_SIZE` is making the snippet a bit less readable. Something like
```Rust
// u32 (4 bytes in size) should fit inside `code`
assert!(code.len() > index + 4 && usize::MAX - 4 >= index);

let values = value.to_ne_bytes();
code[index + 1..index + 4].copy_from_slice(values.as_slice());
```
Should be fine

jedel1043: > Would it be more clear to put something like:
> 
> ```rust
> let slice_index = index + 1;
> code[slice_index..slice_index + U32_SIZE].copy_from_slice(values.as_slice())
> ```
> 
> Adding the extra line may not be necessary, but it may help with readability down the line.

This doesn't optimize the indexing :(
https://godbolt.org/z/hzbhW7e3c

HalidOdat: The previous one with the constant is fine IMO.

Razican (approved):
Looks great!

jedel1043 (approved):
Nice!

PR #2681, opened 2023-03-16
[Merged by Bors] - Fix exponent operator

HalidOdat:
When `abs(base) == 1` and `y == +/- Infinity` result should be `NaN`

Fixes the remaining failing tests in the `exponentiation` test suite

### DIFF
--- a/boa_engine/src/value/operations.rs
+++ b/boa_engine/src/value/operations.rs
@@ -195,6 +195,8 @@ impl JsValue {
     }
 
     /// Perform the binary `**` operator on the value and return the result.
+    // NOTE: There are some cases in the spec where we have to compare floats
+    #[allow(clippy::float_cmp)]
     pub fn pow(&self, other: &Self, context: &mut Context<'_>) -> JsResult<Self> {
         Ok(match (self, other) {
             // Fast path:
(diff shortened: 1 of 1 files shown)


boa_engine/src/value/operations.rs
```diff
@@ -195,22 +195,41 @@ impl JsValue {
...
-            (Self::Integer(x), Self::Rational(y)) => Self::new(f64::from(*x).powf(*y)),
+            (Self::Rational(x), Self::Rational(y)) => {
+                if x.abs() == 1.0 && y.is_infinite() {
+                    Self::nan()
+                } else {
+                    Self::new(x.powf(*y))
+                }
+            }
+            (Self::Integer(x), Self::Rational(y)) => {
+                if x.abs() == 1 && y.is_infinite() {
```
jedel1043, line 215: This will overflow if x is `i32::MIN`. It should probably use `checked_abs` instead.

jedel1043 (changes requested):
Good catch!

boa_engine/src/value/operations.rs
HalidOdat: I don't think it matters if it overflows since `i32::MIN.abs()` equals `i32::MIN` and it wont satisfy the condition, `checked_abs` might add a jump a condition.

But there is still a problem in debug mode `.abs()` panics on overflow so I should  have used `.wrapping_abs()`.

jedel1043 (approved):
LGTM

nekevss (approved):
LGTM 😄

Issue #2658, opened 2023-03-13
panic: iterator [[Done]] was not a boolean


langston-barrett:
```javascript
function AddPromise(){}tests=[{body(){}},{body(){}},{body(){async function*agf(a){for await(m of a){yield}}makeIter=function(){return{next(){return{e:this}}}}
iterTwo={[Symbol.asyncIterator]:makeIter}
const genTwo=agf(iterTwo);AddPromise(this.e,genTwo.next())
AddPromise(this.e,genTwo.next())}}]
for(const test of tests){test.body()}
```
```
undefined
thread 'main' panicked at 'iterator [[Done]] was not a boolean', boa_engine/src/vm/opcode/iteration/for_await.rs:24:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:64:14
   2: core::panicking::panic_display
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:147:5
   3: core::panicking::panic_str
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:131:5
   4: core::option::expect_failed
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/option.rs:1924:5
   5: <boa_engine::vm::opcode::iteration::for_await::ForAwaitOfLoopIterate as boa_engine::vm::opcode::Operation>::execute
   6: boa_engine::vm::<impl boa_engine::context::Context>::execute_instruction
   7: boa_engine::vm::<impl boa_engine::context::Context>::run
   8: <boa_engine::native_function::Closure<F,T> as boa_engine::native_function::TraceableClosure>::call
   9: boa_engine::vm::code_block::<impl boa_engine::object::jsobject::JsObject>::call_internal
  10: boa_engine::object::internal_methods::function::function_call{{reify.shim}}
  11: boa_engine::context::hooks::HostHooks::call_job_callback
  12: core::ops::function::FnOnce::call_once{{vtable.shim}}
  13: <boa::Jobs as boa_engine::job::JobQueue>::run_jobs
  14: boa::main
[...]

PR #2683, opened 2023-03-16
Correctly run async tests

jedel1043:
So, while working on #2658 I apparently broke 1280 tests... except I didn't! Turns out we were considering async tests that didn't call the `print` function as passed tests, but the `async` section of https://github.com/tc39/test262/blob/main/INTERPRETING.md#flags says:
> The test must not be considered complete until the implementation-defined print function has been invoked or some length of time has passed without any such invocation.

### DIFF
--- a/boa_tester/src/exec/mod.rs
+++ b/boa_tester/src/exec/mod.rs
@@ -198,8 +198,12 @@ impl Test {
 
                 context.run_jobs();
 
-                if let Err(e) = async_result.inner.borrow().as_ref() {
-                    return (false, format!("Uncaught {e}"));
+                match *async_result.inner.borrow() {
+                    UninitResult::Err(ref e) => return (false, format!("Uncaught {e}")),
+                    UninitResult::Uninit if self.flags.contains(TestFlags::ASYNC) => {
+                        return (false, "async test didn't call `print`".to_string())
+                    }
+                    _ => {}
                 }
 
                 (true, value.display().to_string())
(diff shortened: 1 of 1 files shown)


jedel1043:
  
  | Total | 94,277 | 94,277 | 0 |
  | Passed | 71,994 | 70,714 | **-1,280** |
  | Ignored | 17,324 | 17,324 | 0 |
  | Failed | 4,959 | 6,239 | **+1,280** |
  | Panics | 14 | 14 | 0 |
  | Conformance | 76.36% | 75.01% | **-1.36%** |
  
  <details><summary><b>Broken tests (1280):</b></summary>
  
  ```
  (1280 tests)
  ```

boa_tester/src/exec/mod.rs
```diff
@@ -198,8 +198,12 @@ impl Test {
 
                 context.run_jobs();
 
-                if let Err(e) = async_result.inner.borrow().as_ref() {
-                    return (false, format!("Uncaught {e}"));
+                match *async_result.inner.borrow() {
+                    UninitResult::Err(ref e) => return (false, format!("Uncaught {e}")),
+                    UninitResult::Uninit if self.flags.contains(TestFlags::ASYNC) => {
+                        return (false, "async test didn't call `print`".to_string())
```
nekevss, line 204: Should we make the failed message a little more explicit? Something like "async test did not print \"Test262:AsyncTestComplete\""

jedel1043: I like that!

boa_tester/src/exec/mod.rs
```diff
@@ -440,17 +457,26 @@ fn register_print_fn(context: &mut Context<'_>, async_result: AsyncResult) {
         Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
     );
 }
+
+#[derive(Debug, Clone, Copy, Default)]
```
Razican, line 461: Could we add documentation to the structure?

Razican (approved):
Good catch! I think we can merge it with a bit of documentation in the new result structure.

PR #2690, opened 2023-03-18
[Merged by Bors] - Add ES5 and ES6 Conformance calculation to boa_tester

### LINKED ISSUE #2629: Collect ES5 conformance information
It would be good to show our current spec conformance for ES5 exclusively, since many transpilers assume all engines support at least ES5.

# Preliminar research

We currently parse a struct that stores the full metadata of every test being run:

https://github.com/boa-dev/boa/blob/c15bf73b65a7d666e9e01be16cca9a82b45a8746/boa_tester/src/read.rs#L20-L37

Could be good to check if we can use the `es[56]?id` fields to determine if a test covers the ES5 spec or not.

We could also check how other engines that exclusively target ES5 run the test suite.

ZackMitkin:
This is calculated based the tests `es5id` or `es6id`.  

Judging by [this](https://github.com/tc39/test262/issues/1557) it's probably not the best method of doing it, but the alternative would require a lot of rework on the boa_tester so that it can pull an older version of the test262 spec which has it's own problems since there's not really an "ES5" only version.
 
I would think that if we're 100% passing on es5id's then it's safe to assume boa supports es5 (assuming test262's es5id is covering all the es5 cases)

This Pull Request fixes/closes #2629.

It changes the following:
- Store `spec_version` in TestResult based on the tests `es[6/5]id`
- Count all es5, es6 and their test outcome during `TestSuite::run()`
- Print the conformance.

I'm serializing the `spec_version` outcomes so that it can be displayed in test262 github page. I'd like to work on that too if possible. Let me know if there's anything more I should cover in this PR, I'll gladly do it :)

### DIFF
--- a/boa_tester/src/exec/mod.rs
+++ b/boa_tester/src/exec/mod.rs
@@ -3,8 +3,8 @@
 mod js262;
 
 use crate::{
-    read::ErrorType, Harness, Outcome, Phase, SuiteResult, Test, TestFlags, TestOutcomeResult,
-    TestResult, TestSuite,
+    read::ErrorType, Harness, Outcome, Phase, SpecVersion, Statistics, SuiteResult, Test,
+    TestFlags, TestOutcomeResult, TestResult, TestSuite,
 };
 use boa_engine::{
     context::ContextBuilder, job::SimpleJobQueue, native_function::NativeFunction,
@@ -58,80 +58,75 @@ impl TestSuite {
         }
 
         // Count passed tests and es specs
-        let mut passed = 0;
-        let mut ignored = 0;
-        let mut panic = 0;
-        let mut es5_passed = 0;
-        let mut es6_passed = 0;
-        let mut es5_total = 0;
-        let mut es6_total = 0;
+        let mut all = Statistics::default();
+        let mut es5 = Statistics::default();
+        let mut es6 = Statistics::default();
+
+        let mut append_stats = |spec_version: SpecVersion, f: &dyn Fn(&mut Statistics)| {
+            f(&mut all);
+            if spec_version == SpecVersion::ES5 {
+                f(&mut es5);
+            } else if spec_version == SpecVersion::ES6 {
+                f(&mut es6);
+            }
+        };
 
         for test in &tests {
             match test.result {
                 TestOutcomeResult::Passed => {
-                    passed += 1;
-                    if test.es5 {
-                        es5_passed += 1;
-                    }
-                    if test.es6 {
-                        es6_passed += 1;
-                    }
+                    append_stats(test.spec_version, &|stats| {
+                        stats.passed += 1;
+                    });
                 }
-                TestOutcomeResult::Ignored => ignored += 1,
-                TestOutcomeResult::Panic => panic += 1,
+                TestOutcomeResult::Ignored => {
+                    append_stats(test.spec_version, &|stats| {
+                        stats.ignored += 1;
+                    });
+                }
+                TestOutcomeResult::Panic => {
+                    append_stats(test.spec_version, &|stats| {
+                        stats.panic += 1;
+                    });
+                }
                 TestOutcomeResult::Failed => {}
             }
-            if test.es5 {
-                es5_total += 1;
-            }
-            if test.es6 {
-                es6_total += 1;
-            }
+            append_stats(test.spec_version, &|stats| {
+                stats.total += 1;
+            });
         }
 
         // Count total tests
-        let mut total = tests.len();
         for suite in &suites {
-            total += suite.total;
-            passed += suite.passed;
-            ignored += suite.ignored;
-            panic += suite.panic;
-            es5_total += suite.es5_total;
-            es6_total += suite.es6_total;
-            es5_passed += suite.es5_passed;
-            es6_passed += suite.es6_passed;
+            all = all + suite.all_stats.clone();
+            es5 = es5 + suite.es5_stats.clone();
+            es6 = es6 + suite.es6_stats.clone();
             features.append(&mut suite.features.clone());
         }
 
         if verbose != 0 {
             println!(
-                "Suite {} results: total: {total}, passed: {}, ignored: {}, failed: {} (panics: \
+                "Suite {} results: total: {}, passed: {}, ignored: {}, failed: {} (panics: \
                     {}{}), conformance: {:.2}%",
+                all.total,
                 self.path.display(),
-                passed.to_string().green(),
-                ignored.to_string().yellow(),
-                (total - passed - ignored).to_string().red(),
-                if panic == 0 {
+                all.passed.to_string().green(),
+                all.ignored.to_string().yellow(),
+                (all.total - all.passed - all.ignored).to_string().red(),
+                if all.panic == 0 {
                     "0".normal()
                 } else {
-                    panic.to_string().red()
+                    all.panic.to_string().red()
                 },
-                if panic == 0 { "" } else { " ⚠" }.red(),
-                (passed as f64 / total as f64) * 100.0
+                if all.panic == 0 { "" } else { " ⚠" }.red(),
+                (all.passed as f64 / all.total as f64) * 100.0
             );
         }
-
         SuiteResult {
             name: self.name.clone(),
-            total,
-            passed,
-            ignored,
-            panic,
+            all_stats: all,
+            es5_stats: es5,
+            es6_stats: es6,
             suites,
-            es5_total,
-            es6_total,
-            es5_passed,
-            es6_passed,
             tests,
             features,
         }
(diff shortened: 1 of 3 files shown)


Razican (commented):
It seems that the Test262 comparator fails because the current "main" branch doesn't have the new fields in JSON. I guess they should be optional fields.

ZackMitkin:
> It seems that the Test262 comparator fails because the current "main" branch doesn't have the new fields in JSON. I guess they should be optional fields.

Yep, just noticed too, I've just added `#[serde(default)]` to the `TestResult` for es5 and es6

boa_tester/src/main.rs
```diff
@@ -348,6 +364,10 @@ struct SuiteResult {
 struct TestResult {
     #[serde(rename = "n")]
     name: Box<str>,
+    #[serde(rename = "5", default)]
+    es5: bool,
+    #[serde(rename = "6", default)]
+    es6: bool,
```
jedel1043, line 370: I don't think we need an `es5/6` tag per se, we can manage with a field like `spec_version: u8`, which should allow filtering tests by `version <= 6` without having to do things like `es5 || es6`

jedel1043 (changes requested):
Thanks for the contribution! I just have a small suggestion for development ease :)

boa_tester/src/main.rs
ZackMitkin: sounds good, I'll fix that soon

boa_tester/src/main.rs
```diff
@@ -375,6 +397,9 @@ struct Test {
     name: Box<str>,
     description: Box<str>,
     esid: Option<Box<str>>,
+    spec_version: Option<i8>,
+    es5id: Option<Box<str>>,
+    es6id: Option<Box<str>>,
```
jedel1043, line 402: We can remove those fields, we have enough information to filter by version with `spec_version`

boa_tester/src/main.rs
```diff
@@ -375,6 +397,9 @@ struct Test {
     name: Box<str>,
     description: Box<str>,
     esid: Option<Box<str>>,
+    spec_version: Option<i8>,
```
jedel1043, line 400: No need to make it an `Option<i8>`, we can be conservative and assume any tests that don't have `es` ids are from the latest version (ES13). An `u8` should be enough.

boa_tester/src/main.rs
```diff
@@ -348,6 +364,12 @@ struct SuiteResult {
 struct TestResult {
     #[serde(rename = "n")]
     name: Box<str>,
+    #[serde(rename = "v", default)]
+    spec_version: Option<i8>,
+    #[serde(default)]
+    es5: bool,
+    #[serde(default)]
+    es6: bool,
```
jedel1043, line 372: Same comments as the ones in `Test`.

boa_tester/src/exec/mod.rs
```diff
@@ -57,17 +57,36 @@ impl TestSuite {
...
 
-        // Count passed tests
+        // Count passed tests and es specs
         let mut passed = 0;
         let mut ignored = 0;
         let mut panic = 0;
+        let mut es5_passed = 0;
+        let mut es6_passed = 0;
+        let mut es5_total = 0;
+        let mut es6_total = 0;
```
jedel1043, line 67: Seeing that this is the same as the normal counters, maybe it would be good to extract this to its own `Statistics` struct with fields `total: u32, passed: u32, ignored: u32, panic: u32`. That way we can just have a nice `stats: Statistics, es5_stats: Statistics, es6_stats: Statistics` for `SuiteResult`.

jedel1043 (changes requested):

ZackMitkin:
I've separated the `SuiteResults` stats by es-version and all stats. This will break the current CI but I can create a PR to back-fill the previous runs and fix the test262 github page sometime this week

boa_tester/src/main.rs
```diff
@@ -318,11 +330,9 @@ struct TestSuite {
...
 
-/// Outcome of a test suite.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-struct SuiteResult {
-    #[serde(rename = "n")]
-    name: Box<str>,
+/// Represents a tests statistic
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+struct Statistics {
     #[serde(rename = "c")]
```
jedel1043, line 336:
```suggestion
    #[serde(rename = "t")]
```

boa_tester/src/results.rs
```diff
@@ -41,10 +41,10 @@ impl From<ResultInfo> for ReducedResultInfo {
...
             commit: info.commit,
             test262_commit: info.test262_commit,
-            total: info.results.total,
-            passed: info.results.passed,
-            ignored: info.results.ignored,
-            panic: info.results.panic,
+            total: info.results.all_stats.total,
+            passed: info.results.all_stats.passed,
+            ignored: info.results.all_stats.ignored,
+            panic: info.results.all_stats.panic,
```
jedel1043, line 47: This should also use the new `Statistics` struct.

jedel1043:
Opened #2706 to solve the CI issue. We can merge that PR after we agree on the new data model here and this PR after that.

boa_tester/src/main.rs
```diff
@@ -390,16 +406,22 @@ enum TestOutcomeResult {
     Panic,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
jedel1043, line 409: Should also derive `Copy`. We can remove the `clone` calls that way.

boa_tester/src/main.rs
```diff
@@ -368,13 +406,22 @@ enum TestOutcomeResult {
...
```
Razican, line 409: Being able to copy this around should make it easier to compare later without needing to use reference comparison and so on.

```suggestion
#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
```

boa_tester/src/exec/mod.rs
```diff
@@ -57,53 +57,75 @@ impl TestSuite {
...
-        // Count passed tests
-        let mut passed = 0;
-        let mut ignored = 0;
-        let mut panic = 0;
+        // Count passed tests and es specs
+        let mut all = Statistics::default();
+        let mut es5 = Statistics::default();
+        let mut es6 = Statistics::default();
+
+        let mut append_stats = |spec_version: &SpecVersion, f: &dyn Fn(&mut Statistics)| {
```
Razican, line 65: If `SpecVersion` is `Copy`, you can just use it without a reference, which makes comparisons simpler without an `&`.

boa_tester/src/main.rs
jedel1043: Oops, Raz already made a comment for this 😅

boa_tester/src/exec/mod.rs
ZackMitkin: Ah, didn't know that. Thanks for the tip!

boa_tester/src/main.rs
ZackMitkin: You did it first ;)

Thanks for the tip!

jedel1043 (approved):
Thanks for the contribution! Looks good to me

HalidOdat (approved):
Looks good to me! :)

jedel1043:
Ok, gonna merge #2706 and then run CI on this.

ZackMitkin:
> Ok, gonna merge #2706 and then run CI on this.

Wouldn't it be better to merge your PR into this branch to check CI? Not sure why run it on main...

jedel1043:
> Wouldn't it be better to merge your PR into this branch to check CI? Not sure why run it on main...
Because #2706 doesn't target `main`, but `gh-pages`, so we cannot combine both PRs into one.

ZackMitkin:
```
Copying the ../gh-pages/test262/refs/heads/main/results.json file to ../results/test262/pull/results.json in order to add the results
Writing the results to ../results/test262/pull...
Error: 
   0: could not write the results to the output JSON file
   1: missing field `a5` at line 1 column 136
```

No idea what's happening here... I have seen it before though during testing. It worked fine if I wrote to a clean or non-existing dir, but if I tried to overwrite an older schema it'd fail... didn't think serde cared about the overwritten schema...

jedel1043:
Yeah, serde is very strict with deserializations. Our best option here would be to derive `Default` for `Statistics` and add a `serde(default)` to the new fields.

boa_tester/src/main.rs
```diff
@@ -318,19 +330,43 @@ struct TestSuite {
...
+        }
+    }
+}
+
+/// Outcome of a test suite.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct SuiteResult {
+    #[serde(rename = "n")]
+    name: Box<str>,
+    #[serde(rename = "a", default)]
```
jedel1043, line 364: This doesn't need a `default`, because all schemas need to have an `a` for them to be valid. We're making an exception for `a5` and `a6` because they were introduced later.

jedel1043:
All good now!

jasonwilliams (approved):
LGTM
Great contribution!

HalidOdat (approved):
Looks good to me! :)

PR #2692, opened 2023-03-19
[Merged by Bors] - Implement `with` and object environments

raskad:

- Implement `with` statement parsing, ast node, compilation and excution.
- Implement object environments that are used in the `with` statement excution.

The implementation of object environments can probably be optimized further by using more compile-time information about when object environments can exist. Maybe there could also be a separate environment stack for object environments to reduce the filtering and iteration that is needed with the current implementation.
This does not fix all tests in the `test/language/statements/with` suite yet. But for most failing tests that I have looked at we are missing other features / have bugs elsewhere.

As a note for the review:
The functions in the `impl Context` block in `boa_engine/src/environments/runtime.rs` are mostly copied / moved from the existing functions. The only change there should be the addition of the object environment logic. They had to be moved to `Context` because of borrow semantics.

### DIFF
--- a/boa_engine/src/environments/runtime.rs
+++ b/boa_engine/src/environments/runtime.rs
@@ -1,5 +1,6 @@
 use crate::{
-    environments::CompileTimeEnvironment, error::JsNativeError, object::JsObject, Context, JsValue,
+    environments::CompileTimeEnvironment, error::JsNativeError, object::JsObject, Context,
+    JsResult, JsString, JsSymbol, JsValue,
 };
 use boa_ast::expression::Identifier;
 use boa_gc::{Finalize, Gc, GcRefCell, Trace};
@@ -32,6 +33,8 @@ pub(crate) struct DeclarativeEnvironment {
     compile: Gc<GcRefCell<CompileTimeEnvironment>>,
     #[unsafe_ignore_trace]
     poisoned: Cell<bool>,
+    #[unsafe_ignore_trace]
+    with: Cell<bool>,
     slots: Option<EnvironmentSlots>,
 }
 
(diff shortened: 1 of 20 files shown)


boa_engine/src/environments/runtime.rs
```diff
@@ -852,3 +843,323 @@ impl BindingLocator {
...
+        &mut self,
+        name: Identifier,
+    ) -> JsResult<(bool, bool)> {
+        for env_index in (0..self.realm.environments.stack.len()).rev() {
+            let env = self
+                .realm
+                .environments
+                .stack
+                .get(env_index)
+                .expect("environment index must be in range");
```
HalidOdat, line 1136:
```suggestion
            let env = self.environment_expect(env_index);
```
Maybe it would be good to refactor into a utility function, this pattern seems to be in many functions

boa_engine/src/environments/runtime.rs
```diff
@@ -852,3 +843,323 @@ impl BindingLocator {
...
+                        if compile.is_function() {
+                            if let Some(b) = compile.get_binding(name) {
+                                let mut bindings = self
+                                    .realm
+                                    .environments
+                                    .stack
+                                    .get(b.environment_index)
+                                    .expect("environment index must be in range")
+                                    .as_declarative()
+                                    .expect("environment must be declarative")
```
HalidOdat, line 1087:
```suggestion
                                let mut bindings = self
                                    .declarative_environment_expect(b.environment_index)
```
I would also refactor this into a utility function

HalidOdat (approved):
Looks good! Great work @raskad

Just some small improvement suggestions

PR #2694, opened 2023-03-20
[Merged by Bors] - Fix Object constructor

raskad:

- Add the additional required condition in the Object constructor that checks if the `NewTarget` is the Object constructor itself.


jedel1043 (approved):
Nice catch!

Razican (approved):
Thanks!

PR #2709, opened 2023-03-20
Make tracing an opt-in feature

HalidOdat:
This PR adds the `"trace"` feature flag that enables vm opcode tracing (off by default), most users aren't interested in tracing/debugging that's why I think it should be made out-in.


boa_engine/src/vm/mod.rs
```diff
@@ -40,6 +44,7 @@ pub struct Vm {
     pub(crate) frames: Vec<CallFrame>,
     pub(crate) stack: Vec<JsValue>,
     pub(crate) err: Option<JsError>,
+    #[cfg(feature = "trace")]
     pub(crate) trace: bool,
```
nekevss, line 48: Do we still need the trace field when we can use `#[cfg(feature = "trace")]` and `#[cfg(not((feature = "trace"))]`?

Razican: Hmm I guess this is to enable disabling tracing for particular CLI requests even if it's enabled for other requests in the same executable.

Razican (approved):
Looks good for me!

boa_engine/src/vm/mod.rs
HalidOdat: The `trace` feature enables the ability to trace (compile time), and the `trace` field turns the trace on/off at runtime :)

clippy would also complain that there is an unused field.

nekevss (approved):
Looks good to me 😄

PR #2720, opened 2023-03-21
[Merged by Bors] - Improve tester display for multiple editions

jedel1043:
This Pull Request improves our test results display per edition and cleanups our edition detector logic.

It changes the following:

- Adds a new `edition` flag to limit the maximum edition that will be tested.
- Adds a new `versioned` flag to display all tested editions in a table.
- Adds utility methods to `SpecEdition` to detect the edition of a test and get all the available editions.
- Cleanups logic.

Output with this PR ~(We only collect ES5, ES6 and ES13 stats, so all other editions are a WIP)~:

image

~Marking as a draft since I need to determine the version of the remaining features, but feel free to review everything else.~ Finished!

### DIFF
--- a/boa_tester/src/edition.rs
+++ b/boa_tester/src/edition.rs
@@ -274,27 +274,27 @@ static FEATURE_EDITION: phf::Map<&'static str, SpecEdition> = phf::phf_map! {
 )]
 #[serde(untagged)]
 pub(crate) enum SpecEdition {
-    /// [ECMAScript 5.1 Edition][https://262.ecma-international.org/5.1]
+    /// [ECMAScript 5.1 Edition](https://262.ecma-international.org/5.1)
     ES5 = 5,
-    /// [ECMAScript 6th Edition][https://262.ecma-international.org/6.0]
+    /// [ECMAScript 6th Edition](https://262.ecma-international.org/6.0)
     ES6,
-    /// [ECMAScript 7th Edition][https://262.ecma-international.org/7.0]
+    /// [ECMAScript 7th Edition](https://262.ecma-international.org/7.0)
     ES7,
-    /// [ECMAScript 8th Edition][https://262.ecma-international.org/8.0]
+    /// [ECMAScript 8th Edition](https://262.ecma-international.org/8.0)
     ES8,
-    /// [ECMAScript 9th Edition][https://262.ecma-international.org/9.0]
+    /// [ECMAScript 9th Edition](https://262.ecma-international.org/9.0)
     ES9,
-    /// [ECMAScript 10th Edition][https://262.ecma-international.org/10.0]
+    /// [ECMAScript 10th Edition](https://262.ecma-international.org/10.0)
     ES10,
-    /// [ECMAScript 11th Edition][https://262.ecma-international.org/11.0]
+    /// [ECMAScript 11th Edition](https://262.ecma-international.org/11.0)
     ES11,
-    /// [ECMAScript 12th Edition][https://262.ecma-international.org/12.0]
+    /// [ECMAScript 12th Edition](https://262.ecma-international.org/12.0)
     ES12,
-    /// [ECMAScript 13th Edition][https://262.ecma-international.org/13.0]
+    /// [ECMAScript 13th Edition](https://262.ecma-international.org/13.0)
     ES13,
     /// The edition being worked on right now.
     ///
-    /// A draft is currently available in https://tc39.es/ecma262.
+    /// A draft is currently available in <https://tc39.es/ecma262>.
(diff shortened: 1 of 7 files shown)


boa_tester/src/edition.rs
```diff
@@ -0,0 +1,332 @@
...
+    ES5 = 5,
+    ES6,
+    ES7,
+    ES8,
+    ES9,
+    ES10,
+    ES11,
+    ES12,
+    #[default]
+    ES13,
```
Razican, line 272: I would maybe add "ESNext" here, since we track the next version of the spec, and some of the features are not yet available in the yearly spec.

boa_tester/src/edition.rs
```diff
@@ -0,0 +1,332 @@
+use std::fmt::Display;
```
Razican, line 1: We should add a bit of module-level documentation, and a bit of documentation in all structs, constants and methods.

boa_tester/src/edition.rs
```diff
@@ -0,0 +1,332 @@
...
+            Self::ES12 => "ECMAScript 12",
+            Self::ES13 => "ECMAScript 13",
+        }
+        .fmt(f)
+    }
+}
+
+impl SpecEdition {
+    pub(crate) fn from_test_metadata(metadata: &MetaData) -> Self {
+        if metadata.es5id.is_some() {
```
Razican, line 294: This is usually true, but it might happen that a test has multiple features, and some of them are ES5, while others are ES6, not sure if we're taking that into account and report the higher version among the features in the test.

boa_tester/src/edition.rs
```diff
@@ -0,0 +1,332 @@
...
+        for feature in &*metadata.features {
+            let feature_edition = FEATURE_EDITION.get(feature).copied().unwrap_or_default();
+
+            min_edition = std::cmp::max(min_edition, feature_edition);
+        }
+
+        min_edition
+    }
+
+    pub(crate) fn all_editions() -> impl Iterator<Item = Self> {
```
Razican, line 318: This can be `#[inline]`

boa_tester/src/exec/mod.rs
```diff
@@ -12,118 +12,128 @@ use boa_engine::{
...
 
         if verbose != 0 {
             println!();
         }
 
         // Count passed tests and es specs
-        let mut all = Statistics::default();
         let mut es5 = Statistics::default();
         let mut es6 = Statistics::default();
+        let mut es13 = Statistics::default();
```
Razican, line 70: I would use `esnext` here, so that we don't need to change it every year.

Razican (commented):
It's looking great! I added some suggestions

boa_tester/src/edition.rs
```diff
@@ -0,0 +1,332 @@
...
+            Self::ES6 => "ECMAScript 6",
+            Self::ES7 => "ECMAScript 7",
+            Self::ES8 => "ECMAScript 8",
+            Self::ES9 => "ECMAScript 9",
+            Self::ES10 => "ECMAScript 10",
+            Self::ES11 => "ECMAScript 11",
+            Self::ES12 => "ECMAScript 12",
+            Self::ES13 => "ECMAScript 13",
+        }
+        .fmt(f)
```
ZackMitkin, line 288:
```suggestion
       write!(f, "ECMAScript {}", *self as u8)
```

boa_tester/src/edition.rs
jedel1043: `pub(crate)` functions aren't affected by an `#[inline]`, only `pub` ones.

boa_tester/src/main.rs
```diff
@@ -406,62 +563,48 @@ enum TestOutcomeResult {
...
     {
-        let spec_version = if metadata.es5id.is_some() {
-            SpecVersion::ES5
-        } else if metadata.es6id.is_some() {
-            SpecVersion::ES6
-        } else {
-            SpecVersion::ES13
-        };
+        let edition = SpecEdition::from_test_metadata(&metadata)
+            .ok_or_else(|| eyre!("couldn't find a minimum edition from the test metadata"))?;
```
HalidOdat, line 592:
```suggestion
        let edition = SpecEdition::from_test_metadata(&metadata)
            .ok_or_else(|| eyre!("couldn't find a minimum edition from the test metadata"))?;
```
I think it would be nice to print the feature that was not found to `stdout`, so we can easily add it :)

HalidOdat (approved):
Looks good to me!

Just a small suggestion :)

Razican (approved):
Thanks!

PR #2723, opened 2023-03-22
[Merged by Bors] - Implement `Hidden classes`

HalidOdat:
This PR implements `Hidden Classes`, I named them as `Shapes` (like Spidermonkey does), calling them maps like v8 seems confusing because there already is a JS builtin, likewise with `Hidden classes` since there are already classes in JS.

There are two types of shapes:  `shared` shapes that create the transition tree, and are shared between objects, this is mainly intended for user defined objects this makes more sense because shapes can create transitions trees, doing that for the builtins seems wasteful (unless users wanted  to creating an object with the same property names and the same property attributes in the same order... which seems unlikely). That's why I added  `unique` shapes, only one object has it. This is similar to previous solution, but this architecture enables us to use inline caching.

There will probably be a performance hit until we implement inline caching.

There still a lot of work that needs to be done, on this:

- [x] Move Property Attributes to shape
- [x] Move Prototype to shape
- [x] ~~Move extensible flag  to shape~~,  On further evaluation this doesn't give any benefit (at least right now), since it isn't used by inline caching also adding one more transition.
- [x] Implement delete for unique shapes.
- [x] If the chain is too long we should probably convert it into a `unique` shape
    - [x] Figure out threshold ~~(maybe more that 256 properties ?)~~ curently set to an arbitrary number (`1024`)
- [x] Implement shared property table between shared shapes
- [x] Add  code Document
- [x] Varying size storage for  properties (get+set = 2, data = 1)
- [x] Add shapes to more object:
    - [x] ordinary object
    - [x] Arrays
    - [x] Functions
    - [x] Other builtins
- [x] Add `shapes.md` doc explaining shapes in depth with mermaid diagrams :)
- [x] Add `$boa.shape` module
- [x] `$boa.shape.id(o)`
- [x] `$boa.shape.type(o)`
- [x] `$boa.shape.same(o1, o2)`
- [x] add doc to `boa_object.md`

### DIFF
--- a/boa_engine/src/builtins/weak/weak_ref.rs
+++ b/boa_engine/src/builtins/weak/weak_ref.rs
@@ -80,7 +80,8 @@ impl BuiltInConstructor for WeakRef {
 
         // 3. Let weakRef be ? OrdinaryCreateFromConstructor(NewTarget, "%WeakRef.prototype%", « [[WeakRefTarget]] »).
         // 5. Set weakRef.[[WeakRefTarget]] to target.
-        let weak_ref = JsObject::from_proto_and_data(
+        let weak_ref = JsObject::from_proto_and_data_with_shared_shape(
+            context.root_shape(),
             get_prototype_from_constructor(new_target, StandardConstructors::weak_ref, context)?,
             ObjectData::weak_ref(WeakGc::new(target.inner())),
         );
@@ -160,7 +161,10 @@ mod tests {
                 "#},
                 |v, _| v.is_object(),
             ),
-            TestAction::inspect_context(|_| boa_gc::force_collect()),
+            TestAction::inspect_context(|context| {
+                context.clear_kept_objects();
+                boa_gc::force_collect();
+            }),
             TestAction::assert_eq("ptr.deref()", JsValue::undefined()),
         ]);
     }
--- a/boa_engine/src/context/intrinsics.rs
+++ b/boa_engine/src/context/intrinsics.rs
@@ -4,22 +4,40 @@ use boa_gc::{Finalize, Trace};
 
 use crate::{
     builtins::{iterable::IteratorPrototypes, uri::UriFunctions},
-    object::{JsFunction, JsObject, ObjectData},
+    object::{
+        shape::shared_shape::{template::ObjectTemplate, SharedShape},
+        JsFunction, JsObject, ObjectData, CONSTRUCTOR, PROTOTYPE,
+    },
+    property::{Attribute, PropertyKey},
+    JsSymbol,
 };
 
 /// The intrinsic objects and constructors.
 ///
 /// `Intrinsics` is internally stored using a `Gc`, which makes it cheapily clonable
 /// for multiple references to the same set of intrinsic objects.
-#[derive(Debug, Default, Trace, Finalize)]
+#[derive(Debug, Trace, Finalize)]
 pub struct Intrinsics {
     /// Cached standard constructors
     pub(super) constructors: StandardConstructors,
     /// Cached intrinsic objects
     pub(super) objects: IntrinsicObjects,
+    /// Cached object templates.
+    pub(super) templates: ObjectTemplates,
 }
 
 impl Intrinsics {
+    pub(crate) fn new(root_shape: &SharedShape) -> Self {
+        let constructors = StandardConstructors::default();
+        let templates = ObjectTemplates::new(root_shape, &constructors);
+
+        Self {
+            constructors,
+            objects: IntrinsicObjects::default(),
+            templates,
+        }
+    }
+
     /// Return the cached intrinsic objects.
     #[inline]
     pub const fn objects(&self) -> &IntrinsicObjects {
@@ -30,6 +48,10 @@ impl Intrinsics {
     #[inline]
     pub const fn constructors(&self) -> &StandardConstructors {
         &self.constructors
+    }
+
+    pub(crate) const fn templates(&self) -> &ObjectTemplates {
+        &self.templates
     }
 }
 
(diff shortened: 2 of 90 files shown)


boa_engine/src/object/shape/shared_shape.rs
```diff
@@ -0,0 +1,381 @@
...
+            );
+        }
+        let next_property_index = if self.previous().is_none() {
+            self.property_index()
+        } else {
+            self.property_index() + 2
+        };
+        let new_inner_shape = Inner {
+            property_table: RefCell::default(),
+            forward_property_transitions: GcRefCell::new(FxHashMap::default()),
```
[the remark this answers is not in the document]
HalidOdat: Not sure what you mean? If you are talking about`GcRefCell` and `RefCell` they don't do any computation.

BTW the latest benchmarks aren't correct (they look horrible) because I have a `DEBUG` bool flag in `SharedShape` that I forgot to turn off before pushing that prints a lot of  shape state to stdout.

boa_engine/src/object/shape/shared_shape.rs
```diff
@@ -0,0 +1,381 @@
...
+            println!(
+                "Shape: Create Transition: {} -------> {}",
+                self.property_key(),
+                key.property_key
+            );
+        }
+        let next_property_index = if self.previous().is_none() {
+            self.property_index()
+        } else {
+            self.property_index() + 2
```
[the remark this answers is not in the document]
HalidOdat: This is to account for accessors (`get`, `set`) not fitting in one `JsValue` so for now, every property is 2 `JsValue`s wide, for now. Added a fixme [here](https://github.com/boa-dev/boa/pull/2723/files#diff-b6b5585cae01e8301c94ad109a439c96c78ec04777704cd12078bdfd2f06a9f2R232-R234) this is a temporary fix.

Tried many solutions (like having a struct that can hold both accessor and data descriptor, ended up being just as inefficient memory wise) but the final solution I'm going with is to have different sized storage types (may seem like the obvious solution but it causes more branching... but the branch scenarios are not that common changing from `get`+`set` to `data` 2 => 1 the other way around 1 => 2, will cause a branch, but that requires calling `Object.defineProperty` to do that on a already existing property from accessor to data or back) in comparison delete causes a branch in almost all scenarios (so I think this is a good solution)

HalidOdat:
This is now nearing completion :) Still have some minor things left! But I think this is ready for review now!

Added code documentation as well as a markdown [`shapes.md (rendered)`](https://github.com/boa-dev/boa/blob/8dd1beaa7abc2aabcfea7da667c184a55702c8b7/docs/shapes.md) Explaining with examples (suggest reading it first, it gives a general overview of the shapes architecture :) )

The benchmarks are not looking good, shapes on their own are more of a memory optimization, will be working on inline caching, which should speed up the execution.

raskad:
> The benchmarks are not looking good, shapes on their own are more of a memory optimization, will be working on inline caching, which should speed up the execution.

Do you think it would make sense to do a very limited inline caching poc to make sure the shape implementation works with that?

HalidOdat:
Do you think it would make sense to do a very limited inline caching poc to make sure the shape implementation works with that?

I have a POC implementation locally, hopefully I can publish it soon as a separate PR, so it's easier to review :)

In the latest commit I also optimized the initialization of object and arrays, direct construction with shapes and storage, hopefully we can see an increase in performance!

docs/shapes.md
```diff
@@ -0,0 +1,193 @@
...
+    InsertX -->|Previous| ObjectPrototype
+
+    InsertY(S3: Insert Shape\n\nPrototype: Object.prototype) --> | Property Count: 2 | PropertyTable
+    InsertY -->|Previous| InsertX
+
+    O(Object o\n\nElement 0: JsValue: 100\nElement 1: JsValue: func\nElement 2: JsValue: func) --> InsertY
+```
+
+We can see that the property has been added into the property table, it has the `has_get` and `has_set` flags set,
+in the object there are two elements added, the first is the `get` function and the second is the `set` function.
```
jasonwilliams, line 110: Do we want to be careful about using _elements_ for this?
In JSC and V8 elements has the meaning of numeric slots, mainly for arrays, I think in this description we're giving the meaning to any property added? In this case get and set functions. 

Was this inspired from somewhere?

docs/shapes.md
```diff
@@ -0,0 +1,193 @@
...
+The following changes accure to the shape tree:
+
+```mermaid
+flowchart TD
+    Root(S0: Root Shape\n\nPrototype: None) --->| Property Count: 0 | PropertyTable[PropertyTable\n\nx: Slot 0, writable, configurable, enumerable\ny: Slot 1, has_get, has_set, configurable, enumerable]
+
+    ObjectPrototype(S1: Prototype Shape\n\nPrototype: Object.prototype) -->| Property Count: 0 |PropertyTable
+    ObjectPrototype -->|Previous| Root
+
+    InsertX(S2: Insert Shape\n\nPrototype: Object.prototype) --> | Property Count: 1 | PropertyTable
```
jasonwilliams, line 149: image

It may be worth showing which property was added, as currently Shape 2 and Shape 3 look the same, apart from the "property add" on the line. It doesn't seem obvious what was added.

Maybe you could put in the box ("added x", "added y") or something

docs/shapes.md
```diff
@@ -0,0 +1,193 @@
...
+    InsertYNew -->|Previous| InsertX
+
+    O(Object o\n\nElement 0: JsValue: func\nElement 1: JsValue: func) --> InsertYNew
+```
+
+**NOTE:**: `o2` has been omitted and the it's shape.
+
+When a deletion happens, we find the node in the chain where we added the property, and get it's parent (`base`),
+we also remember that what transitions happened after the property insertion, then we apply them
+one by one until we construct the chain and return the last shape in that chain.
```
jasonwilliams, line 193: I think this last one needs some tweaking, o2 has been omitted but its shape is still there, so it looks like a dangling shape with nothing referencing it.

docs/shapes.md
```diff
@@ -0,0 +1,193 @@
...
+    InsertY(S3: Insert Shape\n\nPrototype: Object.prototype) --> | Property Count: 2 | PropertyTable
+    InsertY -->|Previous| InsertX
+
+    InsertYNew(S3: Insert Shape\n\nPrototype: Object.prototype) --> | Property Count: 1 |PropertyTableNew[PropertyTable\n\ny: Slot 0, has_get, has_set, configurable, enumerable]
+    InsertYNew -->|Previous| InsertX
+
+    O(Object o\n\nElement 0: JsValue: func\nElement 1: JsValue: func) --> InsertYNew
+```
+
+**NOTE:**: `o2` has been omitted and the it's shape.
```
jasonwilliams, line 189:
```suggestion
**NOTE:**: `o2` has been omitted and then it's shape.
```

boa_engine/src/object/shape/unique_shape.rs
```diff
@@ -0,0 +1,218 @@
...
+    #[unsafe_ignore_trace]
+    property_table: RefCell<PropertyTableInner>,
+
+    /// The prototype of the shape.
+    prototype: GcRefCell<JsPrototype>,
+}
+
+/// Represents a [`Shape`] that is not shared with any other object.
+///
+/// This is useful for objects objects that are inherently unique like,
```
jasonwilliams, line 42:
```suggestion
/// This is useful for objects that are inherently unique like,
```

boa_engine/src/object/shape/slot.rs
```diff
@@ -0,0 +1,86 @@
...
+bitflags! {
+    /// Attributes of a slot.
+    #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    pub struct SlotAttribute: u8 {
+        const WRITABLE = 0b0000_0001;
+        const ENUMERABLE = 0b0000_0010;
+        const CONFIGURABLE = 0b0000_0100;
+        const HAS_GET = 0b0000_1000;
+        const HAS_SET = 0b0001_0000;
+    }
```
jasonwilliams, line 13: Nice use of bitflags here

boa_engine/src/object/shape/slot.rs
```diff
@@ -0,0 +1,86 @@
...
+
+bitflags! {
+    /// Attributes of a slot.
+    #[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    pub struct SlotAttribute: u8 {
+        const WRITABLE = 0b0000_0001;
+        const ENUMERABLE = 0b0000_0010;
+        const CONFIGURABLE = 0b0000_0100;
+        const HAS_GET = 0b0000_1000;
+        const HAS_SET = 0b0001_0000;
```
jasonwilliams, line 12: I wonder if these should be `GET` `SET` instead of the `HAS` prefix, to be consistent with the ones above and it makes more sense when looking at:

```rs
    attributes.set(SlotAttribute::GET, self.get().is_some());
    attributes.set(SlotAttribute::SET, self.set().is_some());
```

boa_engine/src/object/shape/slot.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /// Calculate next slot from previous one.
+    ///
+    /// This is needed because slots do not have the same width.
+    pub(crate) fn from_previous(
+        previous_slot: Option<Slot>,
+        new_attributes: SlotAttribute,
+    ) -> Self {
+        // If there was no previous lot then return 0 as the index.
```
jasonwilliams, line 73:
```suggestion
        // If there was no previous slot then return 0 as the index.
```

boa_engine/src/object/shape/slot.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// if they are both accessors, or both data properties.
+    pub const fn width_match(self, other: Self) -> bool {
+        self.is_accessor_descriptor() == other.is_accessor_descriptor()
+    }
+
+    /// Get the width of the slot.
+    pub fn width(self) -> u32 {
+        // accessor take 2 positions in the storage to accomodate for the `get` and `set` fields.
+        1 + u32::from(self.is_accessor_descriptor())
+    }
```
jasonwilliams, line 41: Could you explain this a bit more, im not sure i fully understand. Accessor slots will jump the index by 2 from what I can see, whats the reason this needs to happen?

docs/shapes.md
HalidOdat: I meant it as an element in the storage `vec` in the object itself. Shapes only deal with string and symbol keys, for indexed we have a separate data type ( #2167 ). I can explicitly say `in the storage array in the object` if that is clearer?

boa_engine/src/object/shape/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        Shape::unique(UniqueShape::default())
+    }
+}
+
+impl Shape {
+    /// The max transition count of a [`SharedShape`] from the root node,
+    /// before the shape will be converted into a [`UniqueShape`]
+    ///
+    /// NOTE: This only applies to [`SharedShape`].
+    const TRANSITION_COUNT_MAX: u16 = 1024;
```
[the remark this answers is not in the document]
HalidOdat: Currently it is an arbitrary number I set, it's one of the things we have to figure out.

boa_engine/src/object/mod.rs
```diff
@@ -1621,8 +1623,8 @@ impl Object {
 
     /// Gets the prototype instance of this object.
     #[inline]
-    pub const fn prototype(&self) -> &JsPrototype {
-        &self.prototype
+    pub fn prototype(&self) -> JsPrototype {
+        self.properties.shape.prototype()
```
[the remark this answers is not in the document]
HalidOdat: when a new property is added into the shape, it inherits the prototype from the previous shape. this is done so we don't traverse the shape chain every time we want to get the prototype.

boa_engine/src/object/shape/slot.rs
HalidOdat: slots are varying in length, accessors have width 2 (get and set ) while data properties have with 1 (JsValue data)

raskad:
@HalidOdat do you know what the currently failing tests are caused by?

| Total | 94,277 | 94,277 | 0 |
| Passed | 71,100 | 68,721 | **-2,379** |
| Ignored | 17,324 | 17,324 | 0 |
| Failed | 5,853 | 8,232 | **+2,379** |
| Panics | 0 | 114 | **+114** |
| Conformance | 75.42% | 72.89% | **-2.52%** |

HalidOdat:
> @HalidOdat do you know what the currently failing tests are caused by?

Ohh, Did you get this from the actions log, I didn't check the logs. Why isn't the test262 action comment updating?

In any case I'll try to find the bug and fix it!

raskad:
> Why isn't the test262 action comment updating?

When we get over some number of changing tests, the list is too large for the update process. I always run the comparision locally to be sure.

HalidOdat:
That seems like a bug, maybe we should create an issue about it, I should at least display the count

jasonwilliams (approved):

HalidOdat:
I'm setting this as ready for review/merge, there are some small things I have to do, like document some functions, fix some minor `TODO`s, and a little code cleanup. but this pretty much complete :)

I'll work more on #2767 now, which utilizes this PR :)

boa_engine/src/object/shape/shared_shape/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+    /// Getter for the previous field.
+    pub fn previous(&self) -> Option<&SharedShape> {
+        self.inner.previous.as_ref()
+    }
+    /// Get the prototype of the shape.
+    pub fn prototype(&self) -> JsPrototype {
+        self.inner.prototype.clone()
+    }
+    /// Get the property this [`SharedShape`] referes to.
```
raskad, line 143:
```suggestion
    /// Get the property this [`SharedShape`] refers to.
```

boa_engine/src/object/shape/shared_shape/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+    /// Get the flags of the shape.
+    fn flags(&self) -> ShapeFlags {
+        self.inner.flags
+    }
+    /// Getter for the [`ForwardTransition`] field.
+    fn forward_transitions(&self) -> &ForwardTransition {
+        &self.inner.forward_transitions
+    }
+    /// Check if the prototype has the given prototype.
```
raskad, line 160:
```suggestion
    /// Check if the shape has the given prototype.
```

raskad (commented):
Amazing work!

In addition to the missing docs is some places I think some functions / structs in the `shape` mod probably do not need to be `pub`. Maybe you can check that when you do the general code cleanup.

boa_engine/src/builtins/weak/weak_ref.rs
```diff
@@ -159,7 +159,12 @@ mod tests {
                 "#},
                 |v, _| v.is_object(),
             ),
-            TestAction::inspect_context(|_| boa_gc::force_collect()),
+            TestAction::inspect_context(|context| {
+                context.clear_kept_objects();
+                boa_gc::force_collect();
```
Razican, line 164: Why do we need to run this three times?

boa_engine/src/context/intrinsics.rs
```diff
@@ -2,19 +2,37 @@
...
 
 /// The intrinsic objects and constructors.
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct Intrinsics {
     /// Cached standard constructors
     pub(super) constructors: StandardConstructors,
     /// Cached intrinsic objects
     pub(super) objects: IntrinsicObjects,
+
```
Razican, line 20: We should add documentation to this, as we did with the rest

boa_engine/src/context/intrinsics.rs
```diff
@@ -928,3 +950,200 @@ impl IntrinsicObjects {
         self.intl.clone()
     }
 }
+
+#[derive(Debug)]
+pub(crate) struct ObjectTemplates {
```
Razican, line 955: I would add documentation for this structure and methods, since this is something that gets used a lot through the codebase.

This would help future contributors.

boa_engine/src/builtins/weak/weak_ref.rs
HalidOdat: This is a leftover from my testing, I think there is a bug in the way we collect weak gc pointers, from my testing when a `WeakRef` instance has reference to a shape, it's not being collected. Even with multiple `boa_gc::force_collect()` are called one after the other, the only way I got it to collect the weak ref was by creating a huge number of objects `for (let i =0; i < 10000; ++i) { ({}) }` this is why the `Weak` builtin variants don't have shapes on their instances.

Razican: Do we have an issue open about this? If not, I think we should open one, to debug and fix it if need it :)

HalidOdat: I tested it again and it seem to work fine, strange :confused:

boa_engine/src/value/tests.rs
```diff
@@ -1,3 +1,4 @@
+use boa_macros::utf16;
```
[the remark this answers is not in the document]
HalidOdat: > Is this a leftover? There is no other change in this file

Probably, but it's strange that `clippy` didn't report and unused import error

raskad (approved):
Absolutely amazing work and documentation!

There are just a few missing docs and style things left I would comment on but I'd say let's merge this and fix those later. That way the diffs are smaller and easier to review.

boa_engine/src/context/intrinsics.rs
```diff
@@ -933,3 +955,302 @@ impl IntrinsicObjects {
...
+    ///
+    /// Transitions:
+    ///
+    /// 1. `__proto__`: `Object.prototype`
+    /// 2. `"done"`: (`WRITABLE`, `CONFIGURABLE`, `ENUMERABLE`)
+    /// 3. `"value"`: (`WRITABLE`, `CONFIGURABLE`, `ENUMERABLE`)
+    pub(crate) const fn iterator_result(&self) -> &ObjectTemplate {
+        &self.iterator_result
+    }
+    /// Cached ordinary object template.
```
Razican, line 1117: I would add line breaks between functions, to keep it cleaner. This is just one, but there are more below.
```suggestion
    }
    
    /// Cached ordinary object template.
```

boa_engine/src/object/jsobject.rs
```diff
@@ -712,9 +740,9 @@ Cannot both specify accessors and a value or writable attribute",
 
     /// Inserts a field in the object `properties` without checking if it's writable.
     ///
-    /// If a field was already in the object with the same name that a `Some` is returned
-    /// with that field, otherwise None is returned.
-    pub fn insert_property<K, P>(&self, key: K, property: P) -> Option<PropertyDescriptor>
+    /// If a field was already in the object with the same name, than `true` is returned
+    /// with that field, otherwise `false` is returned.
```
Razican, line 744: Hmm, here it says "`true` is returned with that field", but if `true` is returned, then the previous field is not returned, right?
This seems to be a leftover from the previous documentation comment, which should be updated.

boa_engine/src/object/mod.rs
```diff
@@ -1839,7 +1849,7 @@ impl Object {
 
     /// Helper function for property removal.
     #[inline]
-    pub(crate) fn remove(&mut self, key: &PropertyKey) -> Option<PropertyDescriptor> {
+    pub(crate) fn remove(&mut self, key: &PropertyKey) -> bool {
```
Razican, line 1852: Here we should add a documentation comment similar to the one in `insert()`, explaining what it returns too.

boa_engine/src/vm/code_block.rs
```diff
@@ -625,41 +604,116 @@ pub(crate) fn create_function_object(
...
+    let contructor = template.create(data, storage);
+
+    if let Some(constructor_prototype) = &constructor_prototype {
+        constructor_prototype.borrow_mut().properties_mut().storage[0] = contructor.clone().into();
     }
+    contructor
+}
 
-    constructor
+pub(crate) fn create_function_object_fast(
```
Razican, line 641: Could we add some documentation to this function, to understand the difference with `create_function_object()`? Also, in `create_function_object()` we should explain when is this one preferred.

docs/shapes.md
```diff
@@ -0,0 +1,220 @@
...
+
+    PropertyTableFork(<b>PropertyTable</b>\ny: Slot 0, has_get, has_set, configurable, enumerable)
+    InsertYNew(<b>S3: Insert Shape\nProperty:</b> <i>y</i>) --> | Property Count: 1 |PropertyTableFork:::New
+    InsertYNew:::New -->|Previous| ObjectPrototype
+
+    O(<b>Object o</b>\n<b>Element 0:</b> JsValue: func\n<b>Element 1:</b> JsValue: func) --> InsertYNew
+    O:::New
+```
+
+**NOTE:**: `o2` has been omitted and it's shape.
```
Razican, line 216: A bit better wording:

```suggestion
**NOTE:**: `o2` and its shape have been omitted from the diagram.
```

Razican (commented):
It's looking very well! Still quite a bit to review, though xD

I added some suggestions :)

HalidOdat:
Currently the integration will builtin contruction is not fully complete, there are still some places that don't utilize direct shape initialization, but I have left it for #2813 which should improve buitin construction! :)

jasonwilliams:
I still want to take a second review of this also

HalidOdat:
Did some benchmarks on `main`, `shapes` and `inline-caching` branches (the current state of `inline-cahing` is a WIP)

<h3>Main</h3>

```
Richards: 18.5
DeltaBlue: 17.6
RayTrace: 68.1
----
Score (version 7): 28.1
```

<h3>Shapes</h3>

```
Richards: 19.9
DeltaBlue: 23.6
RayTrace: 80.8
----
Score (version 7): 33.6
```

<h3>Shapes + Inline Caching</h3>

```
Richards: 28.4
DeltaBlue: 29.3
RayTrace: 94.8
----
Score (version 7): 42.9
undefined
```

We can see a pretty big increase in performance, interestingly even just shapes has a big performance increase without inline caching.

raskad:
> > 
> 
> Array pop (Execution) 	660.0±16.63µs 	1029.5±3322.97µs 	+55.98%
> 
> Any idea why this one has gone up so much?

I suspect that is because the test is basically just adding many index properties to an array while creating the array literal. This process is now creating the relevant shapes which is where the overhead comes from.

HalidOdat:
```js
(function () {
  let testArray = [ /* Many elements */ ];

  while (testArray.length > 0) {
    testArray.pop();
  }
  return testArray;
})();
```

> I suspect that is because the test is basically just adding many index properties to an array while creating the array literal. This process is now creating the relevant shapes which is where the overhead comes from.

Indexed properties are special as they don't  cause a shape transition, they are stored in `IndexedProperties` enum (dense or sparse). Shapes only effect symbol and string properties.

My guess is that it's the layers of indirection, `object` -> `shape` -> `PropertyTable`, so doing a property lookup on `length`  has more overhead (which #2767 addresses). and  the `pop` method internally does a `length` property lookup too (which #2796 addresses)

jasonwilliams (approved):

jasonwilliams:
Ok thanks for looking, the rest of the PR looks fine to me, good job!

Razican:
The part that I reviewed looked great. Amazing work!!

bors r+

PR #2724, opened 2023-03-23
[Merged by Bors] - Fix super call execution order

### LINKED ISSUE #2672: panic: function object must have prototype
Found with [tree-splicer](https://github.com/langston-barrett/tree-splicer), reduced with [treereduce](https://github.com/langston-barrett/treereduce).
```js
function base() { }

class beforeSwizzle extends base {
    constructor() {
        super(Object.setPrototypeOf(beforeSwizzle, null));
    }
}

new beforeSwizzle();
```
```
undefined
undefined
undefined
undefined
thread 'main' panicked at 'function object must have prototype', boa_engine/src/vm/opcode/environment/mod.rs:113:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:64:14
   2: core::panicking::panic_display
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:147:5
   3: core::panicking::panic_str
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:131:5
   4: core::option::expect_failed
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/option.rs:1924:5
   5: <boa_engine::vm::opcode::environment::SuperCall as boa_engine::vm::opcode::Operation>::execute
   6: boa_engine::vm::<impl boa_engine::context::Context>::execute_instruction
   7: boa_engine::vm::<impl boa_engine::context::Context>::run
   8: boa_engine::object::internal_methods::function::function_construct{{reify.shim}}
   9: <boa_engine::vm::opcode::new::New as boa_engine::vm::opcode::Operation>::execute
  10: boa_engine::vm::<impl boa_engine::context::Context>::execute_instruction
  11: boa_engine::vm::<impl boa_engine::context::Context>::run
  12: boa_engine::context::Context::execute
  13: boa_engine::context::Context::eval_script
  14: boa::main
```

raskad:
This Pull Request fixes/closes #2672.

It changes the following:

- Get the super constructor and the new target before executing arguments in super calls.

### DIFF
--- a/boa_engine/src/bytecompiler/expression/mod.rs
+++ b/boa_engine/src/bytecompiler/expression/mod.rs
@@ -255,6 +255,8 @@ impl ByteCompiler<'_, '_> {
             }
             Expression::Class(class) => self.class(class, true),
             Expression::SuperCall(super_call) => {
+                self.emit_opcode(Opcode::SuperCallPrepare);
+
                 let contains_spread = super_call
                     .arguments()
                     .iter()
--- a/boa_engine/src/vm/code_block.rs
+++ b/boa_engine/src/vm/code_block.rs
@@ -447,6 +447,7 @@ impl CodeBlock {
             | Opcode::SuperCallDerived
             | Opcode::Await
             | Opcode::PushNewTarget
+            | Opcode::SuperCallPrepare
             | Opcode::CallEvalSpread
             | Opcode::CallSpread
             | Opcode::NewSpread
--- a/boa_engine/src/vm/flowgraph/mod.rs
+++ b/boa_engine/src/vm/flowgraph/mod.rs
@@ -525,6 +525,7 @@ impl CodeBlock {
                 | Opcode::CallSpread
                 | Opcode::NewSpread
                 | Opcode::SuperCallSpread
+                | Opcode::SuperCallPrepare
                 | Opcode::SetPrototype
                 | Opcode::IsObject
                 | Opcode::Nop
--- a/boa_engine/src/vm/opcode/environment/mod.rs
+++ b/boa_engine/src/vm/opcode/environment/mod.rs
@@ -84,6 +84,46 @@ impl Operation for Super {
     }
 }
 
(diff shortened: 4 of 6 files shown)


HalidOdat (approved):
Looks good to me! :)

jedel1043 (commented):
Tested this case with V8 and SM but V8 throws instead of returning `true`, and seeing that there are no test262 tests affected by this change, it seems like this behaviour is not specified by the spec. Should we open an issue on ecma262 before committing to this behaviour?

raskad:
> Tested this case with V8 and SM but V8 throws instead of returning `true`, and seeing that there are no test262 tests affected by this change, it seems like this behaviour is not specified by the spec. Should we open an issue on ecma262 before committing to this behaviour?

The two tests that I added result in the expected results in both SM and V8 for me.
The spec is also pretty good here. The section is in `SuperCall : super Arguments` https://tc39.es/ecma262/#sec-super-keyword-runtime-semantics-evaluation:

```
...
3. Let func be GetSuperConstructor().
4. Let argList be ? ArgumentListEvaluation of Arguments.
5. If IsConstructor(func) is false, throw a TypeError exception.
6. Let result be ? Construct(func, argList, newTarget).
...
```

We currently do step 4 before we fo the rest and that causes the panics / wrong results.

jedel1043:
Ah, so the spec does specify this behaviour! Then it seems like SpiderMonkey has a bug, probably...

SpiderMonkey:

image
 

V8:

image

> The spec is also pretty good here. The section is in `SuperCall : super Arguments`

Then wouldn't that invalidate the second test, since it is not throwing when calling the constructor?

jedel1043 (approved):
Good catch!

PR #2725, opened 2023-03-23
Fix remaining `Set` tests

raskad:

- Add locking for `Set`s during iteration. We already do this for `Map`s.
- Properly implement negative zero handling for `add`, `has` and `delete`.
- Refactor some `Set` functions and add spec comments.


HalidOdat (approved):
Looks good to me! :)

PR #2733, opened 2023-03-24
[Merged by Bors] - Fix update expressions getting values multiple times

raskad:
Currently update expressions get values multiple times. This can lead to multiple executions of object property getters.

This is very similar to #2551. Unfortunateley it seems like we have to sacrifice some code duplication for correctness in these cases. But that is probably for the best, as we can generate more optimized bytecode for each of these `get/set` cases.


HalidOdat (approved):
Nice catch! This looks perfect to me!

nekevss (approved):
Looks good to me 😄

PR #2738, opened 2023-03-25
[Merged by Bors] - Add AST node for parenthesized expressions

raskad:
Currently we have no explicit representation for parenthesized expressions which makes some behaviours impossible to detect. A bonus is that we can now turn AST that contains parenthesized expressions back to code.

- Add an AST node for parenthesized expressions.
- Adjust some conversions and checks to "ignore"/"expand" parenthesized expressions.
- Fix some tests that had parenthesized expressions.

### DIFF
--- a/boa_engine/src/bytecompiler/mod.rs
+++ b/boa_engine/src/bytecompiler/mod.rs
@@ -198,6 +198,7 @@ impl Access<'_> {
             Expression::Identifier(name) => Some(Access::Variable { name: *name }),
             Expression::PropertyAccess(access) => Some(Access::Property { access }),
             Expression::This => Some(Access::This),
+            Expression::Parenthesized(expr) => Self::from_expression(expr.expression()),
             _ => None,
         }
     }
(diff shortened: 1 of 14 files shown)


HalidOdat (approved):
Looks good to me!

boa_engine/src/bytecompiler/mod.rs
```diff
@@ -1239,28 +1245,36 @@ impl<'b, 'host> ByteCompiler<'b, 'host> {
...
+        let (call, mut kind) = match callable {
+            Callable::Call(call) => (call, CallKind::Call),
             Callable::New(new) => (new.call(), CallKind::New),
         };
 
-        match call.function() {
+        let mut function = call.function();
+        while let Expression::Parenthesized(p) = function {
+            function = p.expression();
+        }
```
jedel1043, line 1256: Saw this pattern on several places. Maybe it should be extracted to something like `Expression::flatten`

jedel1043 (approved):
Just a small nitpick that doesn't block merging. Nice work!

PR #2743, opened 2023-03-25
[Merged by Bors] - Added a Boa runtime

### LINKED ISSUE #718: Split WebAPI (environment) built-in objects from the ES builtin objects.

Razican:
This Pull Request fixes/closes #718.

It changes the following:

- Adds a new `boa_runtime` crate, that will only include `console` for now
- Changes the `boa_cli` crate to use the new `boa_runtime` crate for the console, instead of the `console` feature of `boa_engine`
- Removes the `console` feature in `boa_engine`
- Adds a new `boa_testing` helper crate with some useful functions for testing `boa`. This part duplicates the code from `boa_engine`, but I could not make `boa_engine` work with this crate as a dependency due to circular dependencies. Maybe doing it a bit generic could work, but didn't have enough time to check it.

To be checked: wether the WASM example works as expected with the console.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
     "boa_macros_tests",
     "boa_parser",
     "boa_profiler",
+    "boa_runtime",
     "boa_tester",
     "boa_unicode",
     "boa_wasm",
@@ -26,15 +27,16 @@ license = "Unlicense/MIT"
 description = "Boa is a Javascript lexer, parser and compiler written in Rust. Currently, it has support for some of the language."
 
 [workspace.dependencies]
-boa_engine = { version = "0.16.0", path = "boa_engine" }
-boa_interner = { version = "0.16.0", path = "boa_interner" }
-boa_gc = { version = "0.16.0", path = "boa_gc" }
-boa_profiler = { version = "0.16.0", path = "boa_profiler" }
-boa_unicode = { version = "0.16.0", path = "boa_unicode" }
-boa_macros = { version = "0.16.0", path = "boa_macros" }
 boa_ast = { version = "0.16.0", path = "boa_ast" }
-boa_parser = { version = "0.16.0", path = "boa_parser" }
+boa_engine = { version = "0.16.0", path = "boa_engine" }
+boa_gc = { version = "0.16.0", path = "boa_gc" }
 boa_icu_provider = { version = "0.16.0", path = "boa_icu_provider" }
+boa_interner = { version = "0.16.0", path = "boa_interner" }
+boa_macros = { version = "0.16.0", path = "boa_macros" }
+boa_parser = { version = "0.16.0", path = "boa_parser" }
+boa_profiler = { version = "0.16.0", path = "boa_profiler" }
+boa_runtime = { version = "0.16.0", path = "boa_runtime" }
+boa_unicode = { version = "0.16.0", path = "boa_unicode" }
 
 [workspace.metadata.workspaces]
 allow_branch = "main"
--- a/boa_runtime/src/lib.rs
+++ b/boa_runtime/src/lib.rs
@@ -1,0 +1,277 @@
+//! Example runtime for Boa
+//!
+//! This crate contains an example runtime for the `boa_engine` crate, so that it can be used as a
+//! template for runtime implementors. It contains some basic functionality that can be used by
+//! other crates.
+
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/boa-dev/boa/main/assets/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/boa-dev/boa/main/assets/logo.svg"
+)]
+#![cfg_attr(not(test), forbid(clippy::unwrap_used))]
+#![warn(missing_docs, clippy::dbg_macro)]
+#![deny(
+    // rustc lint groups https://doc.rust-lang.org/rustc/lints/groups.html
+    warnings,
+    future_incompatible,
+    let_underscore,
+    nonstandard_style,
+    rust_2018_compatibility,
+    rust_2018_idioms,
+    rust_2021_compatibility,
+    unused,
+
+    // rustc allowed-by-default lints https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html
+    macro_use_extern_crate,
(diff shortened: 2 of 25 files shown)


boa_runtime/src/lib.rs
```diff
@@ -0,0 +1,73 @@
+//! Example runtime for Boa
+//!
+//! This crate contains an example runtime for the `boa_engine` crate, so that it can be used as a
```
nekevss, line 3: Maybe this is a larger discussion, but should we be referring to this as an example runtime or as boa's basic runtime implementation and encourage implementors to use it as a template/build on it for their own needs?

Razican: It only has one builtin for now, so I'm not sure at this point 😅 but I'm open to it, of course

jasonwilliams:
Missed opportunity to call it Boa Time!?

boa_runtime/src/lib.rs
nekevss: Neither am I lol I mainly mention it because the crate is named `boa_runtime` but it only has console in it. Since there have been issues submitted around a `setTimeout` implementation, someone may see the crate and immediately think: "oh, it's a runtime" and then get confused. If you're fine with it, then I'm good to merge this.

Razican: For now, I would say, let's merge it, and add timeout and all this stuff to it slowly, see how it evolves :)

nekevss (approved):
Looks good to me :)

jedel1043 (changes requested):
I don't think having a public `boa_testing` crate is a good idea, since the API of that will evolve considerably while we work on the engine.

I'd very much prefer to move the tests of `console` into `boa_testing` and skip publishing the crate, maybe allowing us to migrate our integration tests to that crate in the future.

Razican:
> I don't think having a public `boa_testing` crate is a good idea, since the API of that will change constantly while we work on the engine.
> 
> I'd very much prefer to move the tests of `console` into `boa_testing` and skip publishing the crate, maybe allowing us to migrate our integration tests to that crate in the future.

I have skipped publishing of the crate. About moving console tests there, we could do it, but then I would like to have them under `#[cfg(test)]` so that we keep `cargo test` working.

jedel1043 (approved):
All good then!

jedel1043 (changes requested):
Oops, typo on the attribute

jedel1043:
Now that I think about it, cargo also considers dev dependencies to publish crates, so we cannot exclude `boa_testing` and also use it as a dev dependency on a published crate.

I know this because I ran `cargo publish -p boa_runtime --dry-run` and it returned:

```
no matching package named `boa_testing` found
  location searched: registry `crates-io`
```

even if `boa_testing` is only a dev dependency. We have to move the tests from `boa_runtime` to `boa_testing` if we don't want to publish it.

Razican:
> Now that I think about it, cargo also considers dev dependencies to publish crates, so we cannot exclude `boa_testing` and also use it as a dev dependency on a published crate.
> 
> I know this because I ran `cargo publish -p boa_runtime --dry-run` and it returned:
> 
> ```
> no matching package named `boa_testing` found
>   location searched: registry `crates-io`
> ```
> 
> even if `boa_testing` is only a dev dependency. We have to move the tests from `boa_runtime` to `boa_testing` if we don't want to publish it.

Hmmm, I'm also not fully confident of having "boa_testing" and "boa_tester" as crates, can confuse people. I'll try to make the crate not depend on boa_engine, and if I can't, then I would say we should duplicate the code in boa_runtime, and improve the architecture in the future if needed.

Razican:
> > Hmmm, I'm also not fully confident of having "boa_testing" and "boa_tester" as crates, can confuse people. I'll try to make the crate not depend on boa_engine, and if I can't, then I would say we should duplicate the code in boa_runtime, and improve the architecture in the future if needed.
> 
> Maybe we could just name it 'tests'? If the crate won't be published anyways, it shouldn't matter if it doesn't have the 'boa_' prefix. It also gives a pretty strong indicator that this crate is not published, only used internally.

For now, I've gone through the route of duplicating the code in the `boa_runtime` crate. This should work while we find a better solution for this where we reduce code duplication, publishing works well, and it still code coverage and `cargo test` work.

Feel free to review it again, as I have rebased the code :)

Cargo.toml
```diff
@@ -26,15 +27,16 @@ license = "Unlicense/MIT"
...
-boa_unicode = { version = "0.16.0", path = "boa_unicode" }
+boa_icu_provider = { version = "0.16.0", path = "boa_icu_provider" }
+boa_interner = { version = "0.16.0", path = "boa_interner" }
 boa_macros = { version = "0.16.0", path = "boa_macros" }
-boa_ast = { version = "0.16.0", path = "boa_ast" }
 boa_parser = { version = "0.16.0", path = "boa_parser" }
-boa_icu_provider = { version = "0.16.0", path = "boa_icu_provider" }
+boa_profiler = { version = "0.16.0", path = "boa_profiler" }
+boa_runtime = { version = "0.16.0", path = "boa_runtime" }
+boa_unicode = { version = "0.16.0", path = "boa_unicode" }
```
Razican, line 39: Here I just re-ordered everything alphabetically, as we have them in the top, and added the boa_runtime.

Razican:
I re-based and ran "cargo update", which updated ICU to 1.2.0. Unfortunately, they have broken backwards compatibility, first with this: https://github.com/unicode-org/icu4x/issues/3332, but then, I replaced the imports, and I now get a bunch of compile errors.

So, it seems that 1.2.0 is not backwards compatible with 1.1.0. I have pinned 1.1.0 for now.

This should probably be reviewed & merged :) Missing @jedel1043's confirmation

jedel1043 (approved):
We can merge as it is, we'll try to deduplicate the tests logic in the future. Nice work!

Razican:
I have rebased this, and reverted bitflags, since bitflags 2.2.0 was yanked. I think we can merge it now (if all tests pass)

HalidOdat (approved):
Looks good to me! Lets merge this! :)

PR #2751, opened 2023-03-27
[Merged by Bors] - Don't construct prototype if not needed

HalidOdat:
When the function is `async || arrow || method ` we don't need to construct the prototype (then drop it).


Razican (approved):
Good catch!

nekevss (approved):
Nice 😄

jasonwilliams (approved):

PR #2752, opened 2023-03-27
[Merged by Bors] - Shrink objects by using `ThinVec`s

HalidOdat:
The fields like `[[PrivateElements]]` are hardly used but they occupy `24` bytes (on 64-bit arch.) the `ThinVec` type stores the `len` and `cap` right before the elements (like our `JsString` implementation) and only a pointer is kept (if not used it does not allocate!), was going to use this in #2723 , since it uses a `Vec` as a dense storage, but the PR is already too big.

It changes the following:

- Shrink object from `328` to `288` bytes (40 bytes reduction)
- Add the `thin_vec` lightweight crate (single file)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -373,6 +373,7 @@ dependencies = [
  "sys-locale",
  "tap",
  "textwrap 0.16.0",
+ "thin-vec",
  "thiserror",
  "unicode-normalization",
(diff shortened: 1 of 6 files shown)


nekevss (approved):
Sounds good to me.

PR #2757, opened 2023-03-28
Shrink size of IndexedProperties

HalidOdat:
Most objects don't have indexed properties, and those who have, have dense, this PR used `ThinVec` to reduce the size of dense propeties.

It changes the following:
- Trim `16` bytes from `IndexedProperties`, this reduces all objects size


Razican (approved):
I guess this should help with performance, cool!

boa_engine/src/object/property_map.rs
```diff
@@ -44,20 +45,33 @@ enum IndexedProperties {
...
+    Sparse(Box<FxHashMap<u32, PropertyDescriptor>>),
+}
+
+unsafe impl Trace for IndexedProperties {
+    custom_trace!(this, {
+        match this {
+            Self::Dense(vec) => {
+                for elem in vec.iter() {
+                    mark(elem);
+                }
```
jedel1043, line 64: I think it'd be better if we implemented `Trace` for `ThinVec` instead, which avoids duplicating `Trace` definitions.

jedel1043 (changes requested):
Just a small nitpick. Good optimization :)

boa_engine/src/object/property_map.rs
Razican: This would require having `ThinVec` as a dependency of the GC, or adding a feature to make it optional, right?

jedel1043: Yep, if we gate it behind a "thinvec" feature it should be okay.

jasonwilliams (approved):

jasonwilliams:
Good optimization

jedel1043 (approved):
Good work!

jedel1043:
Probably needs a rebase cause of #2762

PR #2762, opened 2023-03-29
[Merged by Bors] - Fix deserialization of `SpecEdition`

jedel1043:
`boa_tester` wasn't deserializing `SpecEdition` correcly. This was because the attribute `serde(untagged)` just removes the u8 tag instead of trying to deserialize as a number. This PR fixes this using the `serde_repr` crate.

This needs #2763 to be merged for it to pass CI.

PR #2763, opened 2023-03-29
Update `latest.json` data for #2762

jedel1043:
This should be merged before merging #2762.


raskad (approved):
We have to merge this and #2762 in succession right? Otherwise the next commit to main will just overwrite this.

HalidOdat (approved):

jedel1043:
> We have to merge this and #2762 in succession right? Otherwise the next commit to main will just overwrite this.

Yep.

Issue #2736, opened 2023-03-24
Introduce `tracing` to boa


jedel1043:
We're slowly approaching an eventual 1.0 release, and the scope of the project keeps getting bigger with each new feature and optimisation. However, this means that debugging using only `dbg!` and a debugger will get harder as the time passes.

With that said, I think it's time to add something like [`tracing`](https://docs.rs/tracing/latest/tracing/) to the repo, possibly using [compile time filters](https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters) to disable the tracing in release mode.

jedel1043:
Could also be useful to use `profiling` instead, which is framework-agnostic
https://crates.io/crates/profiling

Jayant-kernel:
Hi! @jedel1043 
 I'd love to take this on as part of GSoC 2026.

I read through both the original proposal and @jedel1043's follow-up about the profiling crate. I think there's a really nice path here, so let me lay out how I'd approach this:

Step 1 — Pick the right abstraction
→The profiling crate seems like the sweet spot over raw tracing. It's zero-cost when no backend is enabled, and it lets users choose their profiler (tracy, puffin, tracing, superluminal) via feature flags — no lock-in.
→ But I'd also like to hear if there's a preference here, since tracing gives richer structured logging (spans with fields, events, subscribers) while profiling is more focused on pure instrumentation.

Step 2 — Instrument the hot paths first
→ Start with the VM execution loop — bytecode dispatch, opcode handling
→ Then GC — trace/sweep timing, allocation counts, heap size
→ Then the parser and compiler passes
→ Each instrumented with profiling::scope!() or #[profiling::function] so it's lightweight and opt-in

Step 3 — Wire up feature flags
→ Add cargo features like profile-with-tracy, profile-with-tracing, etc. that pass through to the profiling crate
→ Use [compile-time filters](https://docs.rs/tracing/latest/tracing/level_filters/index.html) so release builds have zero overhead by default (as the original issue mentioned)

Step 4 — Make it actually useful
→ Add a section to the contributor docs: "How to profile Boa" with a concrete tracy/puffin walkthrough
→ Maybe a --trace flag on boa_cli that enables a default subscriber for quick local debugging
I'd like to start with a small proof-of-concept PR — maybe just instrumenting the VM dispatch loop with profiling and a tracy feature flag — so we can discuss the ergonomics before going wider. Would that be a reasonable first step?

[...]

Issue #1924, opened 2022-03-13
Set-up benchmarking against other engines


Razican:
Currently, we run benchmarks to improve (or at least not regress) on the performance of Boa. This is a very good approach, but we are missing proper benchmarking against other JavaScript engines.

Particularly, it would be interestint to benchmark against
 - Spidermonkey
 - V8 (both in Node and in Deno)

Any other possible engine would also fit. The idea is to store these comparisons for each commit, the same way we do it for our own benchmarks. We can either create new benchmarks, or try to replicate our benchmarks in those engines, and store their values on each commit.

jasonwilliams:
Just dumping some ideas..

Hyperfine seems like it would be good for this, and maybe a separate repo under the boa-dev umbrella which can periodically run.

Something like test262 would be too much, so would need to be just a couple of JS scripts, like 3 or 4 max for now.

birdbrainswagtrain:
[This repo](https://github.com/mozilla/arewefastyet/tree/master/benchmarks) is pretty old but it contains a bunch of benchmark scripts that should be usable.

wusyong:
> [This repo](https://github.com/mozilla/arewefastyet/tree/master/benchmarks) is pretty old but it contains a bunch of benchmark scripts that should be usable.

quickJS [benchmark](https://bellard.org/quickjs/bench.html) with [v8-v7](https://github.com/mozilla/arewefastyet/tree/master/benchmarks/v8-v7). Maybe we can see some results between boa, quickJS, and v8.

wusyong:
I just modify the benchmark above so it can work with boa. Here's the result from @CYBAI 
```
Richards: 8.64
DeltaBlue: 7.19
RayTrace: 30.0
----
Score (version 7): 12.3
```

And here's V8's:

```
Richards: 26212
DeltaBlue: 60518
RayTrace: 65267
----
Score (version 7): 46956
```

I haven't tested the rest because some of them will stuck.

jasonwilliams:
Hey @wusyong great work on this, I had some questions.
Are there any current benchmarks the engines are being judged by today that Boa can be added to or is this the best option? 
What does those numbers mean and are those names [Richards] just some test cases?
What does the score mean?

wusyong:
I don't this is the best benchmark tbh. This benchmark has been archived by mozilla.
I think QuickJS choose this just for quick result. It doesn't provide metrics to keep monitor ever since.
But V8 results from @CYBAI is similar to QuickJS posted.
[...]

PR #2767, opened 2023-03-30
Implement `Inline Caching`

HalidOdat:
~~Depends on #2723~~ 

This PR i~~currently just a POC~~ implements inline caching. ~~This merges into #2723 so the changes can be easily  seen, will switch to `main`,  once `Hidden classes` gets merged.~~

~~Because the `Proxy` internal methods are not implement this will introduce a lot of panics ( will fix ), Was curious about the benchmarks since,  we don't have proxys in the benchmarks they shouldn't fail.~~ The inline caching is now internal method aware so it knows what can be cached.

It changes the following:

- [x] Fix incorrect handling of `Proxy` internal methods
- [x] Make inline caching internal method aware
- [x] Add inline caching to `GetPropertyByName`
- [x] Add inline caching to `SetPropertyByName`
- [ ] ~~Add inline caching to `GetName`~~
- [ ] ~~Add inline caching to `SetName`~~

### DIFF
--- a/boa_engine/src/object/internal_methods/mod.rs
+++ b/boa_engine/src/object/internal_methods/mod.rs
@@ -5,7 +5,12 @@
 //!
 //! [spec]: https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots
 
-use super::{JsPrototype, PROTOTYPE};
+use std::ops::{Deref, DerefMut};
+
+use super::{
+    shape::slot::{Slot, SlotAttributes},
+    JsPrototype, PROTOTYPE,
+};
 use crate::{
     context::intrinsics::{StandardConstructor, StandardConstructors},
     object::JsObject,
@@ -27,6 +32,52 @@ pub(super) mod string;
 
 pub(crate) use array::ARRAY_EXOTIC_INTERNAL_METHODS;
 pub(crate) use integer_indexed::integer_indexed_element_set;
+
+/// A lightweight wrapper around [`Context`] used in [`InternalObjectMethods`].
+#[derive(Debug)]
+pub(crate) struct InternalMethodContext<'ctx> {
+    context: &'ctx mut Context,
+    slot: Slot,
+}
+
+impl<'ctx> InternalMethodContext<'ctx> {
+    /// Create a new [`InternalMethodContext`].
+    pub(crate) fn new(context: &'ctx mut Context) -> Self {
+        Self {
+            context,
+            slot: Slot::new(),
+        }
+    }
+
+    /// Gets the [`Slot`] associated with this [`InternalMethodContext`].
+    #[inline]
+    pub(crate) fn slot(&mut self) -> &mut Slot {
+        &mut self.slot
+    }
+}
+
+impl Deref for InternalMethodContext<'_> {
+    type Target = Context;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        self.context
+    }
+}
+
+impl DerefMut for InternalMethodContext<'_> {
+    #[inline]
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.context
+    }
+}
+
+impl<'context> From<&'context mut Context> for InternalMethodContext<'context> {
+    #[inline]
+    fn from(context: &'context mut Context) -> Self {
+        Self::new(context)
+    }
+}
 
 impl JsObject {
     /// Internal method `[[GetPrototypeOf]]`
@@ -97,7 +148,7 @@ impl JsObject {
     pub(crate) fn __get_own_property__(
         &self,
         key: &PropertyKey,
-        context: &mut Context,
+        context: &mut InternalMethodContext<'_>,
     ) -> JsResult<Option<PropertyDescriptor>> {
         let _timer = Profiler::global().start_event("Object::__get_own_property__", "object");
         (self.vtable().__get_own_property__)(self, key, context)
@@ -115,7 +166,7 @@ impl JsObject {
         &self,
         key: &PropertyKey,
         desc: PropertyDescriptor,
-        context: &mut Context,
+        context: &mut InternalMethodContext<'_>,
     ) -> JsResult<bool> {
         let _timer = Profiler::global().start_event("Object::__define_own_property__", "object");
         (self.vtable().__define_own_property__)(self, key, desc, context)
@@ -132,7 +183,7 @@ impl JsObject {
     pub(crate) fn __has_property__(
         &self,
         key: &PropertyKey,
-        context: &mut Context,
+        context: &mut InternalMethodContext<'_>,
     ) -> JsResult<bool> {
         let _timer = Profiler::global().start_event("Object::__has_property__", "object");
         (self.vtable().__has_property__)(self, key, context)
@@ -150,7 +201,7 @@ impl JsObject {
         &self,
         key: &PropertyKey,
         receiver: JsValue,
-        context: &mut Context,
+        context: &mut InternalMethodContext<'_>,
     ) -> JsResult<JsValue> {
         let _timer = Profiler::global().start_event("Object::__get__", "object");
         (self.vtable().__get__)(self, key, receiver, context)
@@ -169,7 +220,7 @@ impl JsObject {
         key: PropertyKey,
         value: JsValue,
         receiver: JsValue,
-        context: &mut Context,
+        context: &mut InternalMethodContext<'_>,
     ) -> JsResult<bool> {
         let _timer = Profiler::global().start_event("Object::__set__", "object");
         (self.vtable().__set__)(self, key, value, receiver, context)
@@ -183,7 +234,11 @@ impl JsObject {
     ///  - [ECMAScript reference][spec]
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-delete-p
-    pub(crate) fn __delete__(&self, key: &PropertyKey, context: &mut Context) -> JsResult<bool> {
+    pub(crate) fn __delete__(
+        &self,
+        key: &PropertyKey,
+        context: &mut InternalMethodContext<'_>,
+    ) -> JsResult<bool> {
         let _timer = Profiler::global().start_event("Object::__delete__", "object");
         (self.vtable().__delete__)(self, key, context)
     }
@@ -215,6 +270,7 @@ impl JsObject {
     ///  - [ECMAScript reference][spec]
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist
+    #[track_caller]
     pub(crate) fn __call__(&self, argument_count: usize) -> CallValue {
         CallValue::Pending {
             func: self.vtable().__call__,
@@ -233,6 +289,7 @@ impl JsObject {
     ///  - [ECMAScript reference][spec]
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget
+    #[track_caller]
     pub(crate) fn __construct__(&self, argument_count: usize) -> CallValue {
         CallValue::Pending {
             func: self.vtable().__construct__,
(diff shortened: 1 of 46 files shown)


HalidOdat:
There is still a lot of work to be done (currently only the `GetPropertyByName` has inline caching, because I'm still developing the arch. desing), but I was curious after implementing inline caching on prototype properties, so I ran a [subset of the `QuickJS` benchmarks](https://github.com/boa-dev/boa/issues/1924#issuecomment-1166989423):

**Main**
```txt
Richards: 19.5
DeltaBlue: 20.1
RayTrace: 75.0
----
Score (version 7): 30.8
undefined
```

**With Inline caching**

```txt
Richards: 25.3
DeltaBlue: 30.5
RayTrace: 101
----
Score (version 7): 42.7
undefined
```

This is where #1605 really pays off, There are clear performance improvements.

@jedel1043 I was curious in the benchmarks in (https://github.com/boa-dev/boa/pull/2790#issuecomment-1500463155) did you edit the file, because when I ran `combined.js` the main branch It froze on one of the benchmarks.

jedel1043:
> @jedel1043 I was curious in the benchmarks in (#2790 (comment)) did you edit the file, because when I ran combined.js the main branch It froze on one of the benchmarks.

In my case it seemed like it froze, but it was just the long execution time. The only benchmark that threw errors was RegExp I think.

HalidOdat:
> In my case it seemed like it froze, but it was just the long execution time. The only benchmark that threw errors was RegExp I think.

I guess I just didn't wait long enough :sweat_smile: , I'll run them and update the comment once they finish :)

boa_engine/src/object/internal_methods/mod.rs
```diff
@@ -294,14 +302,30 @@ pub(crate) struct InternalObjectMethods {
...
+    ) -> JsResult<bool>,
+    pub(crate) __has_property__:
+        fn(&JsObject, &PropertyKey, slot: &mut Slot, &mut Context<'_>) -> JsResult<bool>,
+    pub(crate) __get__: fn(
+        &JsObject,
+        &PropertyKey,
+        JsValue,
+        slot: &mut Slot,
+        &mut Context<'_>,
+    ) -> JsResult<JsValue>,
```
jedel1043, line 326: I'm seeing a lot of usages of the `slot` argument is to pass a new `Slot` with a new `key`, which is why I think API-wise it'd be better to combine both into a `PropertyKey` that has a `slot: Cell<Slot>` property that is automatically initialized for current `Into<PropertyKey>` implementors.

HalidOdat: `slot` is an out parameter so it wouldn't work with having be merged into `PropertyKey` (we would have to change all call sites to take a reference).

And in the future we would have to pass more than just a slot, currently only direct object and it's `__proro__` properties are inline cached and a slot is sufficient for this.

We could just have methods on `JsObject` with non-sloted variant with just call internal object methods with `Slot::new()`. i.e `JsObject::__set__()` and `JsObject::__set_with_slot__` (or something like this)

jedel1043: My main worry is that at one point we will have to expose the internal methods of an object, since the spec [requires allowing custom exotic objects for hosts](https://tc39.es/ecma262/#sec-host-internal-methods-of-exotic-objects), so I'd prefer to have those be as close to the spec as possible.

> `slot` is an out parameter so it wouldn't work with having be merged into `PropertyKey` (we would have to change all call sites to take a reference).

We already take a reference for the `PropertyKey` on all property-related operations, it's just that it's an immutable one. That's why I proposed putting a `Cell<Slot>` property on `PropertyKey`, since it makes a nicer API where you can do a simpler `"key".into()` if you're not interested in using the slot, but you can also pass a reference to a variable then inspect the slot after the operation using something like `prop.slot()`. 

Another option would be to take a `&mut PropertyKey` instead. However, it doesn't make much sense semantically to have a key be mutated by access operations, but the option is there.

> We could just have methods on `JsObject` with non-sloted variant with just call internal object methods with `Slot::new()`. i.e `JsObject::__set__()` and `JsObject::__set_with_slot__` (or something like this)

If you don't agree with my previous suggestion, I'd very much prefer this, since we can hide the custom internal methods and only leave the "spec related" ones public.

HalidOdat: If we have to expose the internal methods so the user override some/all, then how about this approach: We have a light struct that contains the `&mut Context` and any other internal crate data like `Slot`.

```rust
struct InternalMethodContext<'ctx> {
    context: &'ctx mut Context,
    slot: Slot,
    // any other internal method specific data.
}

impl<'ctx> InternalMethodContext<'ctx> {
   pub fn new(&'ctx mut context) -> Self {
       Self {
           context,
           slot: Slot::new,
       }
   }
   pub(crate) fn with_slot(slot: Slot) -> Self {
       Self {
           context,
           slot,
       }
   }
   pub(crate) fn slot(&mut self) -> &mut Slot {
       &mut self.slot
   }
}

impl Deref for InternalMethodContext {
    // returns context
}
// .. DerefMut
```

This would be a drop-in replacement for the `context` parameter. And where we call the internal methods we would do `__set__(..., &mut InternalMethodContext::new(context))` (the user only provides the methods we worry about the construction of the context anyway, so the user wouldn't even have to do this.)

This would also allow us to add or remove data more easily without making it a breaking change.

boa_engine/src/object/internal_methods/mod.rs
```diff
@@ -25,6 +30,40 @@ pub(super) mod string;
 
 pub(crate) use array::ARRAY_EXOTIC_INTERNAL_METHODS;
 
+#[derive(Debug)]
+pub(crate) struct InternalMethodContext<'ctx, 'host> {
+    context: &'ctx mut Context<'host>,
+    slot: Slot,
+}
```
jedel1043, line 37: Nice! By the way, this could implement `From<Context>` to simplify calls further e.g. `obj.__get__(key, target, &mut context.into())?`

HalidOdat:
I think this is ready for review/merge :)

I left out the global `GetName` and `SetName` cause it generates to many ics, the global object changes a lot, making it not that good of a candidate for inline caching.

I will continue development on inline caching, features like lazy inline cache creation, and adding an initial phase, which would allow for one property access before promoting it to an inline cachable instruction, I think this PR is big enough already.

boa_engine/src/vm/code_block.rs
```diff
@@ -110,7 +164,8 @@ pub struct CodeBlock {
     pub(crate) params: FormalParameterList,
 
     /// Bytecode
-    pub(crate) bytecode: Box<[u8]>,
+    #[unsafe_ignore_trace]
+    pub(crate) bytecode: Box<[Cell<u8>]>,
```
raskad, line 168: As far as I can see the `Cell` is not needed here anymore right?

boa_engine/src/object/shape/slot.rs
```diff
@@ -74,4 +94,13 @@ impl Slot {
...
         }
     }
+
+    pub(crate) fn set_not_cachable_if_already_prototype(&mut self) {
+        self.attributes |= SlotAttributes::from_bits_retain(
+            (self.attributes.bits() & SlotAttributes::PROTOTYPE.bits()) << 2,
+        );
+        // if slot.attributes.contains(SlotAttributes::PROTOTYPE) {
+        //     slot.attributes |= SlotAttributes::NOT_CACHABLE;
+        // }
```
raskad, line 104:
```suggestion

```

boa_engine/src/vm/opcode/get/property.rs
```diff
@@ -25,11 +26,66 @@ impl Operation for GetPropertyByName {
...
+        let mut slot = ic.slot();
+        if slot.is_cachable() {
+            let object_borrowed = object.borrow();
+            if ic.matches(object_borrowed.shape()) {
+                // println!(
+                //     "GET: T: \"{}\" {}: {:?}",
+                //     ic.name.to_std_string_escaped(),
+                //     slot.index,
+                //     slot.attributes
+                // );
```
raskad, line 39:
```suggestion

```

boa_engine/src/vm/opcode/get/property.rs
```diff
@@ -25,11 +26,66 @@ impl Operation for GetPropertyByName {
...
+        slot = *context.slot();
+
+        // println!(
+        //     "GET: F: \"{}\" {}: {:?} - {}",
+        //     key,
+        //     slot.index,
+        //     slot.attributes,
+        //     result.type_of()
+        // );
+
```
raskad, line 80:
```suggestion

```

boa_engine/src/vm/opcode/set/property.rs
```diff
@@ -30,16 +31,81 @@ impl Operation for SetPropertyByName {
...
+        let mut slot = ic.slot();
+        if slot.is_cachable() {
+            let object_borrowed = object.borrow();
+            if ic.matches(object_borrowed.shape()) {
+                // println!(
+                //     "SET: T: \"{}\" {}: {:?}",
+                //     ic.name.to_std_string_escaped(),
+                //     slot.index,
+                //     slot.attributes
+                // );
```
raskad, line 44:
```suggestion

```

boa_engine/src/vm/opcode/set/property.rs
```diff
@@ -30,16 +31,81 @@ impl Operation for SetPropertyByName {
...
         if !succeeded && context.vm.frame().code_block.strict() {
             return Err(JsNativeError::typ()
                 .with_message(format!("cannot set non-writable property: {name}"))
                 .into());
         }
+
+        slot = *context.slot();
+
+        // println!("SET: F: \"{}\" {}: {:?}", name, slot.index, slot.attributes);
+
```
raskad, line 101:
```suggestion

```

boa_engine/src/vm/code_block.rs
HalidOdat: Yes, right now it's not needed, but I kept for implementing lazy inline cache entry generation.

I'll remove it and add it when I create the PR :)

boa_engine/src/vm/opcode/get/property.rs
HalidOdat: Intentionally kept this, because it was very useful for debugging, maybe this is a good point to introduce tracing #2736 ?

boa_engine/src/vm/code_block.rs
raskad: I did wonder if using a Vec<u8> would be more efficent in this case, since `UnsafeCell` prevents optimizations. Probably would be good to benchmark explicitly for the bytecode field.

HalidOdat: I don't think it would change much if it was a `Vec` since we would still have to have some kind of interior mutability (`Vec<Cell<u8>>`) or wrapping the Vec into a `RefCell` (but this requires keeping borrow state, which I think would be less efficient, but allowing to shrink/grow which for the lazy inline catching I don't need).

boa_engine/src/vm/opcode/get/property.rs
HalidOdat: Removed the comment :)

jedel1043:
Just throwing some ideas to the mix. `InternalObjectMethods` seems pretty similar to an `ExecutionContext`, so maybe we could transition in the future to something like that.

HalidOdat:
> maybe we could transition in the future to requiring an `ExecutionContext` on all functions and only having a function that converts a `&mut Context` into an `ExecutionContext` on demand.

You mean having an a struct for functions (like we do for internal object methods) that contains `this`, `arguments` and the `&mut Context` (all the function arguments), right?

If that's the case than I totally agree. The [`neon` crate](https://docs.rs/neon/0.10.1/neon/context/struct.CallContext.html) has this where they have multiple context type that all deref into the main `context` type. It's easier than having multiple function argument and adding/removing/changing something is really hard (requires touching every function). We could completely remove helper traits like the `JsArgs` one and have a method on that specialized context.

jedel1043:
> You mean having an a struct for functions (like we do for internal object methods) that contains `this`, `arguments` and the `&mut Context` (all the function arguments), right?

Yes! That would make it easier to preserve data between calls, and it could enable interesting optimizations.

PR #2768, opened 2023-03-31
[Merged by Bors] - Implement `escape` and `unescape`

jedel1043:
This Pull Request implements the [`escape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) and [`unescape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape) functions.

Both are technically deprecated, but they're also part of the [Additional ECMAScript Features for Web Browsers](https://tc39.es/ecma262/#sec-additional-ecmascript-features-for-web-browsers) section, so it is preferable to have them in place.

### DIFF
--- a/boa_engine/src/context/intrinsics.rs
+++ b/boa_engine/src/context/intrinsics.rs
@@ -794,7 +794,9 @@ impl Default for IntrinsicObjects {
             is_nan: JsFunction::from_object_unchecked(JsObject::default()),
             parse_float: JsFunction::from_object_unchecked(JsObject::default()),
             parse_int: JsFunction::from_object_unchecked(JsObject::default()),
+            #[cfg(feature = "annex-b")]
             escape: JsFunction::from_object_unchecked(JsObject::default()),
+            #[cfg(feature = "annex-b")]
             unescape: JsFunction::from_object_unchecked(JsObject::default()),
             #[cfg(feature = "intl")]
             intl: JsObject::default(),
--- a/boa_engine/src/context/intrinsics.rs
+++ b/boa_engine/src/context/intrinsics.rs
@@ -765,9 +765,11 @@ pub struct IntrinsicObjects {
     parse_int: JsFunction,
 
     /// [`%escape%`](https://tc39.es/ecma262/#sec-escape-string)
+    #[cfg(feature = "annex-b")]
     escape: JsFunction,
 
     /// [`%unescape%`](https://tc39.es/ecma262/#sec-unescape-string)
+    #[cfg(feature = "annex-b")]
     unescape: JsFunction,
 
     /// [`%Intl%`](https://tc39.es/ecma402/#intl-object)
@@ -903,6 +905,7 @@ impl IntrinsicObjects {
     /// Gets the [`%escape%`][spec] intrinsic function.
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-escape-string
+    #[cfg(feature = "annex-b")]
     pub fn escape(&self) -> JsFunction {
         self.escape.clone()
     }
@@ -910,6 +913,7 @@ impl IntrinsicObjects {
     /// Gets the [`%unescape%`][spec] intrinsic function.
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-unescape-string
+    #[cfg(feature = "annex-b")]
     pub fn unescape(&self) -> JsFunction {
         self.unescape.clone()
     }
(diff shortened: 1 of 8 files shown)


Razican (commented):
It looks great to me, but I would put this behind a feature flag, since this should only be enabled for browsers:

> When the ECMAScript [host](https://tc39.es/ecma262/#host) is a web browser the following additional properties of the standard built-in objects are defined.

jedel1043:
> It looks great to me, but I would put this behind a feature flag, since this should only be enabled for browsers:
> 
> > When the ECMAScript [host](https://tc39.es/ecma262/#host) is a web browser the following additional properties of the standard built-in objects are defined.

Thought about it, but I checked Node and Deno and both have these functions enabled, probably because of compatibility purposes... Though, we have plans for feature gating all builtins anyways, so maybe it should be fine to gate them, then add them to the default set of features?

Razican:
> > It looks great to me, but I would put this behind a feature flag, since this should only be enabled for browsers:
> > > When the ECMAScript [host](https://tc39.es/ecma262/#host) is a web browser the following additional properties of the standard built-in objects are defined.
> 
> Thought about it, but I checked Node and Deno and both have these functions enabled, probably because of compatibility purposes... Though, we have plans for feature gating all builtins anyways, so maybe it should be fine to gate them, then add them to the default set of features?

I would feature gate them and add them to the default set, but in the future, I can imagine having:
 - default: base + built-ins + browser (anex B) + intl
 - base: basic language, no built-ins
 - built-ins: non-anex-B built-ins
 - intl: intl module

Then, I would like to have one feature for each feature in test262, ideally, or at least one feature per built-in + async/await + generators... in case users would like to block some functionality. They should also be able to create a context with no features and cherry pick the built-ins at runtime, for example.

But this is out of this PR, of course. I would still add the feature gate, though :)

raskad:
Right now this is missing `Escape::init(&intrinsics);` and `Unescape::init(&intrinsics);` in  `Intrinsics::new()`. The functions are just empty objects, that is why only the negative tests are passing. After that there are two tests in `unescape` that are failing, but I'm currently not sure why.

jedel1043:
> Right now this is missing `Escape::init(&intrinsics);` and `Unescape::init(&intrinsics);` in  `Intrinsics::new()`. The functions are just empty objects, that is why only the negative tests are passing. After that there are two tests in `unescape` that are failing, but I'm currently not sure why.

Remind me to not push changes at 4am again 🙃 Thanks for the catch!

jedel1043:
Ok, all tests are passing for me and I gated the functions behind an `annex-b` feature.

boa_engine/src/context/intrinsics.rs
```diff
@@ -764,6 +764,12 @@ pub struct IntrinsicObjects {
     /// [`%parseInt%`](https://tc39.es/ecma262/#sec-parseint-string-radix)
     parse_int: JsFunction,
 
+    /// [`%escape%`](https://tc39.es/ecma262/#sec-escape-string)
+    escape: JsFunction,
+
+    /// [`%unescape%`](https://tc39.es/ecma262/#sec-unescape-string)
+    unescape: JsFunction,
+
```
HalidOdat, line 772: Shouldn't these also be feature gated?

boa_engine/src/context/intrinsics.rs
```diff
@@ -764,6 +764,12 @@ pub struct IntrinsicObjects {
     /// [`%parseInt%`](https://tc39.es/ecma262/#sec-parseint-string-radix)
     parse_int: JsFunction,
 
+    /// [`%escape%`](https://tc39.es/ecma262/#sec-escape-string)
+    escape: JsFunction,
```
Razican, line 768: Should this also be feature-gated?

HalidOdat: Haha, we commented the same thing :laughing:

boa_engine/src/context/intrinsics.rs
jedel1043: Yep, forgot `Intrinsics`

boa_engine/src/context/intrinsics.rs
```diff
@@ -786,6 +794,8 @@ impl Default for IntrinsicObjects {
             is_nan: JsFunction::from_object_unchecked(JsObject::default()),
             parse_float: JsFunction::from_object_unchecked(JsObject::default()),
             parse_int: JsFunction::from_object_unchecked(JsObject::default()),
+            escape: JsFunction::from_object_unchecked(JsObject::default()),
+            unescape: JsFunction::from_object_unchecked(JsObject::default()),
```
HalidOdat, line 798: Shouldn't it have the `#[cfg(feature = "annex-b")]` annotation? Wouldn't it fail compilation if the `"annex-b"` feature is not enabled.

HalidOdat (approved):
Lookis good to me! :)

raskad (approved):

PR #2772, opened 2023-04-01
[Merged by Bors] - Implement debug object for CLI

HalidOdat:
Currently some debugging stuff in JavaScript land is difficult to impossible, like triggering a GC collect, this is not impossible to do in JavaScript the way I triggered it was by creating a huge amount of object `for (let i = 0; i < 100000; ++i) { ({}) }` but this is cumbersome and not guaranteed to trigger a gc.

This PR implements `--debug-object` flag that injects the `$boa` debug object in the context, the object is separated into modules currently `gc`, `function`, `object`.

We can now do `$boa.gc.collect()`, which force triggers a GC collect.

Or sometimes I wanted a trace (the current solution is great, you can trace stuff like `>>> 1 + 1` but that is also it's limitation), it traces everything, I sometimes have a scenario and just want to trace a single function in that scenario, that's why I added the `$boa.function.trace(func, this, ...args)`  It only traces the function.

```js
>> $boa.function.trace((a, b) => a + b, undefined, 1, 2)

-------------------------Compiled Output: ''--------------------------

000000    0000    DefInitArg                 0000: 'a'
000005    0001    DefInitArg                 0001: 'b'
000010    0002    RestParameterPop           
000011    0003    GetName                    0000: 'a'
000016    0004    GetName                    0001: 'b'
000021    0005    Add                        
000022    0006    Return                     
000023    0007    PushUndefined              
000024    0008    Return                     

... (cut for brevity) ...
```

It also implements `$boa.function.flowgraph(func, options)`:
```js
$boa.function.flowgraph(func, 'graphviz')
$boa.function.flowgraph(func, { format: 'mermaid', direction: 'TopBottom' })
```

Printing the object pointer:
```js
$boa.object.id({}) // '0x566464F33'
```

It currently implements some functionality which we can grow it with our debugging needs since we are not restricted by a spec we can add whatever we want :)

I was originally going to implement this in #2723 (but the PR is too big), for shapes having functions like:
```js
$boa.shape.type({}) // Shared shape
$boa.shape.id({}) // 0x8578FG355 (objects, shape pointer)
$boa.shape.flowgraph({}) // printing the shape transition chain, like $boa.function.flowgraph
```
Shapes chains are very hard to debug once they are big... so having this type of debugging capability would make it much easier.

### DIFF
--- a/boa_cli/src/debug/function.rs
+++ b/boa_cli/src/debug/function.rs
@@ -4,6 +4,7 @@ use boa_engine::{
     vm::flowgraph::{Direction, Graph},
     Context, JsArgs, JsNativeError, JsObject, JsResult, JsValue, NativeFunction,
 };
+use boa_interner::ToInternedString;
 
 use crate::FlowgraphFormat;
 
@@ -108,6 +109,39 @@ fn flowgraph(_this: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> Js
     Ok(JsValue::new(result))
 }
 
+fn bytecode(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+    let Some(value) = args.get(0) else {
+        return Err(JsNativeError::typ()
+        .with_message("expected function argument")
+        .into());
+    };
+
+    let Some(object) = value.as_object() else {
+        return Err(JsNativeError::typ()
+        .with_message(format!("expected object, got {}", value.type_of()))
+        .into());
+    };
+    let object = object.borrow();
+    let Some(function) = object.as_function() else {
+        return Err(JsNativeError::typ()
+        .with_message("expected function object")
+        .into());
+    };
+    let code = match function {
+        Function::Ordinary { code, .. }
+        | Function::Async { code, .. }
+        | Function::Generator { code, .. }
+        | Function::AsyncGenerator { code, .. } => code,
+        Function::Native { .. } => {
+            return Err(JsNativeError::typ()
+                .with_message("native functions do not have bytecode")
+                .into())
+        }
+    };
+
+    Ok(code.to_interned_string(context.interner()).into())
+}
+
 /// Trace function.
 fn trace(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
     let value = args.get_or_undefined(0);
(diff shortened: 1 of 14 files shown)


boa_cli/src/debug/function.rs
```diff
@@ -0,0 +1,135 @@
...
+        return Err(JsNativeError::typ()
+        .with_message("expected callable object")
+        .into());
+    };
+
+    let arguments = args.get(2..).unwrap_or(&[]);
+
+    context.set_trace(true);
+    let result = callable.call(this, arguments, context);
+    context.set_trace(false);
```
jedel1043, line 126: It would be good to have a `trace` property on the functions itself, since this'll also trace called functions which may pollute the output.

boa_cli/src/debug/gc.rs
```diff
@@ -0,0 +1,13 @@
+use boa_engine::{object::ObjectInitializer, Context, JsObject, JsResult, JsValue, NativeFunction};
+
+/// Trigger grabage collection
```
raskad, line 3:
```suggestion
/// Trigger garbage collection
```

raskad (approved):
Nice addition!

HalidOdat:
Now the `.trace` doesn't return the full compiled `CodeBlock`, only what it executed (called functions are not traced as suggested by @jedel1043 )
```js
>> const add = (a, b) => a + b
>> $boa.function.trace(add, undefined, 1, 2)
5μs           DefInitArg                 0000: 'a'                  2
4μs           DefInitArg                 0001: 'b'                  <empty>
0μs           RestParameterPop                                      <empty>
3μs           GetName                    0000: 'a'                  1
1μs           GetName                    0001: 'b'                  2
2μs           Add                                                   3
1μs           Return                                                3
3
>>
```

If we want the full bytecode then, we can use the `.bytecode` function:
```js
>> $boa.function.bytecode(add)
"
------------------------Compiled Output: 'add'------------------------

000000    0000    DefInitArg                 0000: 'a'
000005    0001    DefInitArg                 0001: 'b'
000010    0002    RestParameterPop           
000011    0003    GetName                    0000: 'a'
000016    0004    GetName                    0001: 'b'
000021    0005    Add                        
000022    0006    Return                     
000023    0007    PushUndefined              
000024    0008    Return                     

Literals:
    <empty>

Bindings:
    0000: a
    0001: b

Functions:
    <empty>
"
>>
```

jasonwilliams:
https://github.com/boa-dev/boa/blob/main/docs/debugging.md will need to be updated to cover this

jedel1043 (approved):
Provided @jasonwilliams's comment is addressed, this looks great to me!

jasonwilliams:
Multiple runs give me back the same memory address, is that correct?
```
const foo = {};
$boa.object.id(foo);
```

Always returns the same address on each run for example

HalidOdat:
Added documentation [here (rendered)](https://github.com/boa-dev/boa/blob/3ad18d6080aa202523acef99a647ee1b0ec8e9e6/docs/debugging.md#debugging-through-the-debug-object-boa)

I only included a small introduction in the `debugging.md` file and put the full doc for `$boa` in `boa_object.md`, since it's too big and will become bigger as we add other functionality, like `shapes`

Also the `$boa.optimizer` module for enabling/disabling of optimizations.

PR #2783, opened 2023-04-05
Create new `Runtime` API

jedel1043:
Maybe the name isn't exactly right since #2743 exists, but I wanted to open the PR first to hear your comments about this.

This PR creates a new `Runtime` structure that holds common information that is shared between `Context`'s (`[[KeptAlive]]`, ICU data, `HostHooks`). This should technically allow us to make `Context` immutable, since `Runtime` would act as our "agent", which means it would be the one managing the `Vm` state, but I'll have to do some tests to see if it's worth doing so.

### DIFF
--- a/boa_cli/src/main.rs
+++ b/boa_cli/src/main.rs
@@ -62,11 +62,10 @@ mod helper;
 
 use boa_ast::StatementList;
 use boa_engine::{
-    context::ContextBuilder,
     job::{FutureJob, JobQueue, NativeJob},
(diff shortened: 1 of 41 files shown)


jedel1043:
Drafting since I'm reevaluating whether integrating this API is worth.

Razican:
Would it make sense to go directly for Agents?

jedel1043:
Closing this since it's out of date. We can reimplement it later if needed.

PR #2786, opened 2023-04-05
[Merged by Bors] - Fix cross-realm construction bugs

jedel1043:
This Pull Request fixes test [`assert-throws-same-realm.js`](https://github.com/tc39/test262/blob/eb44f67274bf3896fbec8814f81dd2ae02236686/test/harness/assert-throws-same-realm.js).

It changes the following:

- Handles global variables through the global object, instead of the `context`.
- Adds an `active_function` field to the vm, which is used as the `NewTarget` when certain builtins aren't called with `new`.
- Adds a `realm_intrinsics` field to `Function`.


jedel1043:
Drafting because I want to see the impact this change makes in the test suite first.

jedel1043:
Ok, this PR fixes half of the problem. The other half is that functions should also store the `global_object` and the `global_this` from the current realm, or the full realm in short. Otherwise, the function calls try to access properties from the context realm instead of their stored realm, which causes the regressions. To fix this we need to make `realm` shareable between functions.

There is a problem with this though: the realm is storing the full environment stack instead of only the global bindings. This means the vm needs to access the realm to access any binding at execution, which would be a disaster perf-wise if we wrapped the `DeclarativeEnvironmentStack` on a `GcRefCell` to make it shareable between functions.

Considering this, it'll be better to merge this as it is to reduce the size of the next PR, which will solve the second part of the problem.

boa_cli/src/debug/realm.rs
```diff
@@ -0,0 +1,14 @@
...
+/// Creates a new ECMAScript Realm and returns the global object of the realm.
+fn new_realm(_: &JsValue, _: &[JsValue], _: &mut Context<'_>) -> JsResult<JsValue> {
+    let context = &mut Context::default();
+
+    Ok(context.global_object().into())
+}
+
+pub(super) fn create_object(context: &mut Context<'_>) -> JsObject {
+    ObjectInitializer::new(context)
+        .function(NativeFunction::from_fn_ptr(new_realm), "newRealm", 0)
```
HalidOdat, line 12:
```suggestion
        .function(NativeFunction::from_fn_ptr(new_realm), "create", 0)
```
Since it's already in the `realm` module I would rename this to either `create()` or `new()`.

boa_cli/src/debug/mod.rs
```diff
@@ -35,14 +37,21 @@ fn create_boa_object(context: &mut Context<'_>) -> JsObject {
             gc_module,
             Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
         )
+        .property(
+            "realm",
+            realm_module,
+            Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::CONFIGURABLE,
+        )
```
HalidOdat, line 44: This should also be documented in `boa_object.md`

HalidOdat (commented):
Looks good! Just some small suggestions :)

Also If the added `TODO`s are not going to be addressed in this PR, maybe we should create issues about them?

jedel1043:
> Also If the added TODOs are not going to be addressed in this PR, maybe we should create issues about them?

Ah, I'll probably solve those on the follow-up PR. I just added those to be easily searchable when I start working on it.

Razican:
I'll try to use the new `realm` module in `$boa` you implement new realm creation in `boa_tester` when I return from vacation

I went through most filled and it looked good. If I have the time I'll finish the review today :)

Razican (approved):
Great work! LGTM

HalidOdat (approved):
Looks good to me! :)

PR #2790, opened 2023-04-06
[Merged by Bors] - Lift `InternalObjectMethods` from `Object`

jedel1043:
This Pull Request lifts the `InternalObjectMethods` vtable from `Object`, which should technically improve performance, since we now won't need to call `borrow` to use any of the internal methods, but let's see what the benchmarks show.

It changes the following:

- Adds a new `VTableObject` struct, containing the old `GcRefCell<Object>` and the lifted `InternalObjectMethods`.
- Changes the definition of `JsObject` to `Gc<VTableObject>`. Note that this means the `InternalObjectMethods` are accessible through the `Gc` pointer.
- Reestructures intrinsic initialization and initialization APIs to accommodate this change.

### DIFF
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -54,7 +54,7 @@ impl IntrinsicObject for Array {
 
         let values_function = BuiltInBuilder::with_object(
             realm,
-            realm.intrinsics().objects().array_prototype_values(),
+            realm.intrinsics().objects().array_prototype_values().into(),
         )
         .callable(Self::values)
         .name("values")
--- a/boa_engine/src/builtins/error/type.rs
+++ b/boa_engine/src/builtins/error/type.rs
@@ -22,12 +22,11 @@ use crate::{
     },
     context::intrinsics::{Intrinsics, StandardConstructor, StandardConstructors},
     error::JsNativeError,
-    native_function::NativeFunction,
-    object::{internal_methods::get_prototype_from_constructor, JsObject, ObjectData},
+    object::{internal_methods::get_prototype_from_constructor, JsObject, ObjectData, ObjectKind},
     property::Attribute,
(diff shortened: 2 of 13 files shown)


jedel1043:
It seems like it slightly improved perf on some tests, but I'll try to run bellard to see our current perf index.

jedel1043:
Ok, ran QuickJS's benchmarks and got the following results:

# Main

RESULT Richards 29.7
RESULT DeltaBlue 30.3
RESULT Crypto 49.5
RESULT RayTrace 110
RESULT EarleyBoyer 101
RESULT Splay 92.9
RESULT NavierStokes 14.9
SCORE 49.1

# PR

RESULT Richards 30.8
RESULT DeltaBlue 31.5
RESULT Crypto 50.3
RESULT RayTrace 112
RESULT EarleyBoyer 103
RESULT Splay 92.7
RESULT NavierStokes 14.4
SCORE 49.7

So it is safe to say the changes did improve performance.

HalidOdat (approved):
Looks good to me! :)

raskad:
Both this and #2723 change the `BuiltInBuilder`s. Do we want to try to merge #2723 first to avoid a huge rebase or is that fine @HalidOdat?

HalidOdat:
> Both this and #2723 change the `BuiltInBuilder`s. Do we want to try to merge #2723 first to avoid a huge rebase or is that fine @HalidOdat?

I think It's fine, since It's one of the things that need to be refactored so the builtins work well with shapes

jedel1043:
> I think It's fine, since It's one of the things that needs to be refactored so the builtins work well with shapes

Yeah, I'd also argue that this PR barely changes the object creation API, so it'll be easier to rebase if the smaller change is merged first.

Razican (approved):
This looks very good to me :) thanks!

PR #2796, opened 2023-04-09
Direct array length access

HalidOdat:
Depends on #2723 

This PR utilizes the change done in #2723 on how we store properties on a continues indexed array, this makes the positions of properties such as `"length"` predefined on an array, because it is constructed from an `ObjectTemplate`.

This can make access on such properties just an index lookup into an array, speeding up execution of common operations like `LengthOfArrayLike`.

Pointing to #2723 for easier review, once #2723 is merged I will rebase and point to main :)

### DIFF
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -236,6 +236,28 @@ impl BuiltInConstructor for Array {
 }
 
 impl Array {
+    /// Optimized helper function, that sets the length of the array.
+    fn set_length(o: &JsObject, len: u64, context: &mut Context<'_>) -> JsResult<()> {
+        if o.is_array() && len < (2u64.pow(32) - 1) {
+            let mut borrowed_object = o.borrow_mut();
+            if borrowed_object.properties().shape.to_addr_usize()
+                == context
+                    .intrinsics()
+                    .templates()
+                    .array()
+                    .shape()
+                    .to_addr_usize()
+            {
+                // NOTE: The "length" property is the first element.
+                borrowed_object.properties_mut().storage[0] = JsValue::new(len);
+                return Ok(());
+            }
+        }
+
+        o.set(utf16!("length"), len, true, context)?;
+        Ok(())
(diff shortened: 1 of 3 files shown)


Razican (approved):
Looks good!!

raskad:
@HalidOdat did you already check some of the increased execution benches? E.g. String concatenation, String comparison, Static Object Property Access, Object Creation...

Just from the diff, I dont see how this could cause those large increases.

boa_engine/src/context/mod.rs
```diff
@@ -721,6 +721,8 @@ impl<'icu, 'hooks, 'queue> ContextBuilder<'icu, 'hooks, 'queue> {
         let realm = Realm::create(&*host_hooks, &root_shape);
         let vm = Vm::new(realm.environment().clone());
 
+        let root_shape = SharedShape::root();
+
```
jedel1043, line 725: This creates a new `SharedShape` for the context itself, instead of reusing the `SharedShape` used to initialize the realm. Is that intended?

boa_engine/src/builtins/array/mod.rs
```diff
@@ -236,6 +236,27 @@ impl BuiltInConstructor for Array {
...
+                    .shape()
+                    .to_addr_usize()
+            {
+                // NOTE: The "length" property is the first element.
+                borrowed_object.properties_mut().storage[0] = JsValue::new(len);
+                return Ok(());
+            }
+        }
+
+        o.set(utf16!("length"), len, true, context).map(|_| ())
```
jedel1043, line 257: I think it's better to use `?` here and return `Ok(())`. Just makes it a bit more readable

jedel1043 (commented):
Great change! Just had a small question and a suggestion

boa_engine/src/context/mod.rs
HalidOdat: Nice catch! Will fix that :)

jedel1043 (approved):
Everything good!

PR #2799, opened 2023-04-10
[Merged by Bors] - Fix `String.prototype.replace()` order of `ToString` execution

HalidOdat:
Fix order of `ToString` execution, there are 3 failing tests on `.replace()` but they seem to be regex related


raskad (approved):
Nice fix and also nice optimization with the enum.

jedel1043 (approved):
Good work!

Razican (approved):
Thanks!

PR #2800, opened 2023-04-10
[Merged by Bors] - Fix `Symbol.prototype[@@iterator]`

HalidOdat:
It changes the following:
- Fixes`Symbol.prototype[@@iterator]` so it calls `RequireObjectCoercible` and then `ToString`
- Makes string iterator store a `JsString` which should be faster, instead of `JsValue` then calling `ToString` on every iteration


jedel1043 (approved):
Nice change!

raskad (approved):

Issue #1835, opened 2022-02-13
Adjust `Realm` visibility


raskad:
Currently `Realm` is a public struct and it's only public method is `crate()`. The only reason for this seems to be, so that we can benchmark the creation of a realm.

Either
1. `Realm` is fully private and we need to change how we benchmark it, or
2. we should provide some functionality for the public `Realm` e.g; changing the `Realm` of a `Context`. But is that api usefully in any way?

wusyong:
For the entry point of DOM types, I think we also have to set a `Window` instance to be realm's global object.
https://html.spec.whatwg.org/multipage/window-object.html#script-settings-for-window-objects

jedel1043:
Resolved by #2801

PR #2801, opened 2023-04-10
[Merged by Bors] - Make `Realm` shareable between functions

### LINKED ISSUE #2317: Panic while executing `eval` inside block inside function
**Describe the bug**
The interpreter panics while executing the following snippet:

```javascript
(function(){
  {
    eval("var x;")
  }
})();
```

**Additional context**
While doing some preliminar investigation, I realized functions don't push a new environment:

```
-------------------------Compiled Output: ''--------------------------

000000    0000    RestParameterPop           
000001    0001    PushDeclarativeEnvironment 0, 0
000010    0002    GetName                    0000: 'eval'
000015    0003    PushUndefined              
000016    0004    Swap                       
000017    0005    PushLiteral                0
000022    0006    CallEval                   1
000027    0007    Pop                        
000028    0008    PopEnvironment             
000029    0009    PushUndefined              
000030    0010    Return                     

Literals:
    0000: <string> "var x;"

Bindings:
    0000: eval

Functions:
    <empty>           
```

Notice how the inner block pushes a declarative environment, but the function doesn't. However, if you define argument initializers, the function pushes a function environment, and this fixes the bug:

```javascript
(function(y=5){
  {
    eval("var x;")
  }
})();
```

```
-------------------------Compiled Output: ''--------------------------

000000    0000    JumpIfNotUndefined         7
000005    0001    PushInt8                   5
000007    0002    DefInitArg                 0000: 'y'
000012    0003    RestParameterPop           
000013    0004    PushFunctionEnvironment    0, 1
000022    0005    PushDeclarativeEnvironment 0, 0
000031    0006    GetName                    0001: 'eval'
000036    0007    PushUndefined              
000037    0008    Swap                       
000038    0009    PushLiteral                0
000043    0010    CallEval                   1
000048    0011    Pop                        
000049    0012    PopEnvironment             
000050    0013    PushUndefined              
000051    0014    Return                     

Literals:
    0000: <string> "var x;"

Bindings:
    0000: y
    0001: eval

Functions:
    <empty>
```

cc @raskad and @HalidOdat which are the most familiar with the vm.

jedel1043:
This Pull Request fixes #2317 and #1835, finally giving our engine proper realms 🥳.

It changes the following:

- Extracts the compile environment stack from `Realm` and into `Vm`.
- Adjusts the bytecompiler to accommodate this change.
- Adjusts `call/construct_internal` to accommodate this change. This also coincidentally fixed #2317, which I'm pretty happy about.
- Adjusts several APIs (`NativeJob`, `Realm`) and builtins (`eval`, initializers) to accommodate this change. 
- Adjusts `JsNativeError`s to hold a reference to the Realm from which they were created. This only affects errors created within calls to function objects. Native calls don't need to set the realm because it's inherited by the next outer active function object. TLDR: `JsError` API stays the same, we just set the origin Realm of errors in `JsObject::call/construct_internal`.

### DIFF
--- a/boa_engine/src/builtins/array/array_iterator.rs
+++ b/boa_engine/src/builtins/array/array_iterator.rs
@@ -13,6 +13,7 @@ use crate::{
     error::JsNativeError,
     object::{JsObject, ObjectData},
     property::{Attribute, PropertyNameKind},
+    realm::Realm,
     symbol::JsSymbol,
     Context, JsResult,
 };
@@ -35,11 +36,17 @@ pub struct ArrayIterator {
 }
 
 impl IntrinsicObject for ArrayIterator {
-    fn init(intrinsics: &Intrinsics) {
+    fn init(realm: &Realm) {
         let _timer = Profiler::global().start_event("ArrayIterator", "init");
 
-        BuiltInBuilder::with_intrinsic::<Self>(intrinsics)
-            .prototype(intrinsics.objects().iterator_prototypes().iterator())
+        BuiltInBuilder::with_intrinsic::<Self>(realm)
+            .prototype(
+                realm
+                    .intrinsics()
+                    .objects()
+                    .iterator_prototypes()
+                    .iterator(),
+            )
             .static_method(Self::next, "next", 0)
             .static_property(
                 JsSymbol::to_string_tag(),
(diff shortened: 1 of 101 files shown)


Razican (approved):
I looked into everything but the code block, and it looks great to me :) good job!!

raskad (approved):
Greak work! Really nice refactor of the call/construct code.

PR #2805, opened 2023-04-11
[Merged by Bors] - Fix `String.prototype.substr()`

HalidOdat:
Fixes the remaining failing tests on `String.prototype.substr()`

It changes the following:

- Fixes the remaining failing tests on `String.prototype.substr()
- Move `substr` under the `"annex-b"` feature flag.


jedel1043 (approved):
Good catch!

Razican (approved):

PR #2806, opened 2023-04-11
Implement annex-b `trimLeft` and `trimRight` string methods

HalidOdat:
With this and #2805 , `annex-b/builtins/string` should have 100% conformance, also fixed the `normalize` method incorrect `length` value, which was the last failing tests in `normalize` test suite


jedel1043 (approved):
Nice!

Razican (approved):
Good catch!

PR #2808, opened 2023-04-11
[Merged by Bors] - Fix `TypedArray`s minus zero key

HalidOdat:
It changes the following:

- Fix `-0` index in typed arrays


Razican:
Hmmm are those broken tests expected?

HalidOdat:
> Hmmm are those broken tests expected?

I think this is caused because I need to check it's detached, converting this into a draft PR until it's fixed :)

HalidOdat:
> I think this is caused because I need to check if it's detached first, converting this into a draft PR until it's fixed :)

Fixed! :)

Razican (approved):
Looks good to me!! Great comments explaining deviations from the spec :)

PR #2809, opened 2023-04-11
[Merged by Bors] - Implement `TryFromJs` for `JsObject` wrappers

Razican:
This change allows using built-in `JsObject` wrappers with the `TryFromJs` logic, which makes it easier to derive it in complex objects containing known built-ins, being able to use all the power of the wrappers even inside complex structures.

It changes the following:

 * Implements `TryFromJs` for all `JsObject` wrappers ~, except for `JsProxy` and `Js<Int>Array` wrappers~ -> This has now been added.
 *  Adds checker functions for individual typed array objects.
 * Adds some missing checker functions from `Object` to `JsObject`

~The reason for not implementing it for `JsProxy` is that we don't have a way (as far as I can tell) to know if a `JsObject` is a `Proxy` object. The reason for the typed arrays (note that `JsTypedArray` implements `TryFromJs`) is that I didn't find an easy way to know which type of typed array a `JsObject` was. Do we have a way of using a `JsObject` to create a Rust `JsUint8Array` or the like?~

### DIFF
--- a/boa_engine/src/object/builtins/jsdate.rs
+++ b/boa_engine/src/object/builtins/jsdate.rs
@@ -7,6 +7,7 @@ use chrono::DateTime;
 use crate::{
     builtins::Date,
     object::{JsObject, JsObjectType, ObjectData},
+    value::TryFromJs,
     Context, JsNativeError, JsResult, JsValue,
 };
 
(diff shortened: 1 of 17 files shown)


HalidOdat (approved):
Looks perfect to me! :)

boa_engine/src/object/builtins/jsdate.rs
```diff
@@ -47,6 +48,18 @@ impl JsDate {
         Self { inner }
     }
 
+    /// Create a new `JsDataView` object from an existing object.
```
raskad, line 51:
```suggestion
    /// Create a new `JsDate` object from an existing object.
```

raskad (approved):

PR #2822, opened 2023-04-14
[Merged by Bors] - Implement `String.prototype.toLocaleUpper/LowerCase`

jedel1043:
This fixes some more ES5 tests that were failing because the functions haven't been implemented.

It changes the following:

- Adds `String::to_locale_case`, which uses ICU4X to convert strings to uppercase or lowercase.
- Refactors `String::to_uppercase` and `String::to_lowercase` into a single `String::to_case` which uses a const generic to distinguish each case.
- Adds utility functions on `JsString` to avoid code repetition.

### DIFF
--- a/boa_engine/src/builtins/string/mod.rs
+++ b/boa_engine/src/builtins/string/mod.rs
@@ -1653,7 +1653,6 @@ impl String {
     /// The value will be **converted** to a string if it isn't one.
     ///
     /// More information:
-    ///  - [ECMAScript reference][spec]
     ///  - [MDN documentation][mdn]
     ///
     /// [upper]: https://tc39.es/ecma262/#sec-string.prototype.toUppercase
--- a/boa_engine/src/builtins/string/mod.rs
+++ b/boa_engine/src/builtins/string/mod.rs
@@ -125,7 +125,7 @@ impl IntrinsicObject for String {
             .method(Self::to_case::<false>, "toLowerCase", 0)
             .method(Self::to_case::<true>, "toUpperCase", 0)
             .method(Self::to_locale_case::<false>, "toLocaleLowerCase", 0)
-            .method(Self::to_locale_case::<true>, "toLocaleLowerCase", 0)
+            .method(Self::to_locale_case::<true>, "toLocaleUpperCase", 0)
             .method(Self::substring, "substring", 2)
             .method(Self::split, "split", 2)
             .method(Self::value_of, "valueOf", 0)
@@ -1658,7 +1658,6 @@ impl String {
     /// [upper]: https://tc39.es/ecma262/#sec-string.prototype.toUppercase
     /// [lower]: https://tc39.es/ecma262/#sec-string.prototype.toLowercase
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
-    #[allow(clippy::wrong_self_convention)]
     pub(crate) fn to_case<const UPPER: bool>(
         this: &JsValue,
         _: &[JsValue],
--- a/boa_engine/src/builtins/string/mod.rs
+++ b/boa_engine/src/builtins/string/mod.rs
@@ -122,8 +122,10 @@ impl IntrinsicObject for String {
             .method(Self::pad_end, "padEnd", 1)
             .method(Self::pad_start, "padStart", 1)
             .method(Self::trim, "trim", 0)
-            .method(Self::to_lowercase, "toLowerCase", 0)
-            .method(Self::to_uppercase, "toUpperCase", 0)
+            .method(Self::to_case::<false>, "toLowerCase", 0)
+            .method(Self::to_case::<true>, "toUpperCase", 0)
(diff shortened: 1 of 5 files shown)


boa_engine/src/builtins/string/mod.rs
```diff
@@ -122,8 +122,10 @@ impl IntrinsicObject for String {
             .method(Self::pad_end, "padEnd", 1)
             .method(Self::pad_start, "padStart", 1)
             .method(Self::trim, "trim", 0)
-            .method(Self::to_lowercase, "toLowerCase", 0)
-            .method(Self::to_uppercase, "toUpperCase", 0)
+            .method(Self::to_case::<false>, "toLowerCase", 0)
+            .method(Self::to_case::<true>, "toUpperCase", 0)
+            .method(Self::to_locale_case::<false>, "toLocaleLowerCase", 0)
+            .method(Self::to_locale_case::<true>, "toLocaleLowerCase", 0)
```
raskad, line 128:
```suggestion
            .method(Self::to_locale_case::<true>, "toLocaleUpperCase", 0)
```

boa_engine/src/builtins/string/mod.rs
```diff
@@ -1644,18 +1646,20 @@ impl String {
...
     /// More information:
-    ///  - [ECMAScript reference][spec]
     ///  - [MDN documentation][mdn]
     ///
-    /// [spec]: https://tc39.es/ecma262/#sec-string.prototype.tolowercase
-    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase
+    /// [upper]: https://tc39.es/ecma262/#sec-string.prototype.toUppercase
+    /// [lower]: https://tc39.es/ecma262/#sec-string.prototype.toLowercase
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toUpperCase
     #[allow(clippy::wrong_self_convention)]
```
raskad, line 1661: I think the allow can be removed right?

raskad (commented):
Looks great. When the function name is fixed, almost all tests in the String suite pass.

boa_engine/src/builtins/string/mod.rs
jedel1043: oops 😬

boa_engine/src/builtins/string/mod.rs
jedel1043: Let me check...

raskad (approved):

boa_icu_provider/data/icudata.postcard
HalidOdat, line 1: Is there a way that we can avoid putting large binary data into git history? maybe generating this in a `build.rs`?

raskad: We started with the build.rs approach on this, but it did increase compile times to much, especially in the ci. I think for now this should work. Alternativeley, we could move `boa_icu_provider` to a separate repository to make this easier on the git history.

Also, before looking at the size of this file, I think we should probably look at the 262 test result files on the `gh-pages` branch :D

HalidOdat: We started with the build.rs approach on this, but it did increase compile times to much, especially in the ci. I think for now this should work. Alternativeley, we could move boa_icu_provider to a separate repository to make this easier on the git history.

Yes, for the long term this seems like the a good solution.

> Also, before looking at the size of this file, I think we should probably look at the 262 test result files on the gh-pages branch :D

:laughing: Yeah, It's big, but we could always edit the history on the `gh-pages` tho

HalidOdat (approved):

boa_icu_provider/data/icudata.postcard
Razican: If this file doesn't change much, we might want to use git LFS: https://docs.github.com/en/repositories/working-with-files/managing-large-files/configuring-git-large-file-storage

boa_engine/src/builtins/string/mod.rs
```diff
@@ -1666,101 +1669,95 @@ impl String {
...
-                break;
-            }
+            // 11. Return CodePointsToString(newCodePoints).
+            Ok(result.into())
         }
 
-        // 5. Let L be ! CodePointsToString(upperText).
-        // 6. Return L.
-        Ok(js_string!(upper_text).into())
+        #[cfg(not(feature = "intl"))]
```
Razican, line 1757: Should this function even exist if intl is not active?

raskad: Yes, the spec suggests that:

>An ECMAScript implementation that includes the ECMA-402 Internationalization API must >implement this method as specified in the ECMA-402 specification. If an ECMAScript >implementation does not include the ECMA-402 API the following specification of this method >is used ...


https://tc39.es/ecma262/#sec-string.prototype.tolocalelowercase

jedel1043: Yeah, it is part of the 262 spec. It's just specialized for Intl implementors.

PR #2823, opened 2023-04-14
We have a fork of Boa just for time

lastmjs:
This PR is just to show the remaining reasons for our project [Azle](https://github.com/demergent-labs/azle) to maintain a fork of Boa.

It all comes down to time. Our environment is `wasm32-unknown-unknown` which does not have an implementation of standard time, thus we need to patch certain calls with our own API calls.

It would be great if Boa allowed passing in an implementation of time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,6 +122,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.70"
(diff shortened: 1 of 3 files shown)


lastmjs:
We can close this PR once this is merged: https://github.com/boa-dev/boa/pull/2824

Razican:
Closing this in favor of #2824 :)

PR #2824, opened 2023-04-15
[Merged by Bors] - Add hooks to get the current time and timezone

jedel1043:

- Adds two new hooks to `HostHooks` to access the current UTC time and the current timezone offset.
- Replaces usages of `Local` with the host hook.
- Replaces usages of `Utc::now` and `Local::now` with the hooks.

cc @lastmjs

### DIFF
--- a/boa_engine/src/context/hooks.rs
+++ b/boa_engine/src/context/hooks.rs
@@ -1,3 +1,5 @@
+use chrono::{FixedOffset, Local, NaiveDateTime, Utc};
+
 use crate::{
     builtins::promise::OperationType,
     job::JobCallback,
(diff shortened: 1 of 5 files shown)


boa_engine/src/context/hooks.rs
```diff
@@ -1,3 +1,5 @@
+use chrono::{FixedOffset, Local, NaiveDateTime, Utc};
+
```
Razican, line 2:
```suggestion
```

Razican (approved):
It looks great to me :) Let's wait for @lastmjs to give some feedback on wether this is good enough for him to use the upstream Boa.

lastmjs:
I will try this out now and let you know

lastmjs:
I can confirm this works, all of our tests have passed.

We're now on upstream Boa with this branch, and will be on main once this is merged :)

Thanks so much, this is just great.

HalidOdat (approved):
Looks good to me! Lets get this merged! :)

PR #2826, opened 2023-04-15
[Merged by Bors] - Upgraded to ICU 1.2

Razican:
This PR upgrades ICU to 1.2.

Unfortunately we still have some breaking changes, so this is being handled in https://github.com/unicode-org/icu4x/issues/3332


boa_engine/src/context/icu.rs
```diff
@@ -4,8 +4,10 @@ use icu_collator::{Collator, CollatorError, CollatorOptions};
 use icu_list::{ListError, ListFormatter, ListLength};
 use icu_locid_transform::{LocaleCanonicalizer, LocaleExpander, LocaleTransformError};
 use icu_provider::{
-    yoke::{trait_hack::YokeTraitHack, Yokeable},
-    zerofrom::ZeroFrom,
+    prelude::{
+        yoke::{trait_hack::YokeTraitHack, Yokeable},
+        zerofrom::ZeroFrom,
+    },
```
Razican, line 10: This change is required due to backwards in-compatibility, that might be fixed in https://github.com/unicode-org/icu4x/pull/3333, but we might just want to move to explicitly import `yoke` and `zerofrom`, so that we don't rely on `#[doc(hidden)]` features.

jedel1043: Yeah, it'll be better to import both deps directly.

jedel1043:
Imported both deps directly. This is now just blocked on https://github.com/unicode-org/icu4x/pull/3335

jedel1043:
Just realized we can completely bypass the `buffer` and `any` constructors by using the `unstable` ones.
Just for the note, `unstable` does sound ominous, but I think we're in the clear, because `BoaProvider` unconditionally implements `DataProvider<M>` for all `M: KeyedDataMarker`, which means that any added or removed `P: DataProvider<M>` bounds on the constructors should never break our compilation... I think.

@sffc Can you either confirm or deny my assumptions? I can rollback to our old method if needed, just want to be sure since doing it this way simplifies a lot our logic.

sffc:
> @sffc Can you either confirm or deny my assumptions? I can rollback to our old method if needed, just want to be sure since doing it this way simplifies our logic a lot.

Yes, you have a blanket impl that leverages as_deserializing and as_downcasting, so you can use unstable constructors. The only catch is that the unstable constructors can't read from deprecated data keys, which we did for the first time in this release when we split the LocaleExpander key in three. So unless you regenerate your data, LocaleExpander will fail to construct. Using the non-unstable constructors allows you to use the old data key (they will map it for you internally).

```
impl<M> DataProvider<M> for BoaProvider<'_>
where
    M: KeyedDataMarker + 'static,
    for<'de> YokeTraitHack<<M::Yokeable as Yokeable<'de>>::Output>: Deserialize<'de>,
    for<'a> YokeTraitHack<<M::Yokeable as Yokeable<'a>>::Output>: Clone,
    M::Yokeable: ZeroFrom<'static, M::Yokeable> + MaybeSendSync,
{
    fn load(&self, req: DataRequest<'_>) -> Result<DataResponse<M>, DataError> {
        match self {
            BoaProvider::Buffer(provider) => provider.as_deserializing().load(req),
            BoaProvider::Any(provider) => provider.as_downcasting().load(req),
        }
    }
}
```

jedel1043:
> The only catch is that the unstable constructors can't read from deprecated data keys, which we did for the first time in this release when we split the LocaleExpander key in three.

Oh, that's pretty important, since I'd like to give our users the decision to upgrade their dataset when they want, instead of forcing them to regenerate when we bump `icu` to the latest version. I'll rollback that change.

This is still blocked on that PR then.

Razican:
I rebased this to the latest `main` branch and squashed commits. I also tried to "patch" cargo using the branch of the PR in ICU to see if it worked, but it seems that something is not quite right, probably some double-dependency. Let's wait for the 1.2.1 release, and then we can make this proper :)

sffc:
icu_locid_transform version 1.2.1 is released. This should unblock your upgrade

Cargo.toml
```diff
@@ -62,4 +62,4 @@ opt-level = 1
 # Enables "fat" LTO, for faster benchmark builds
 lto = "fat"
 # Makes sure that all code is compiled together, for LTO
-codegen-units = 1
+codegen-units = 1
```
Razican, line 65:
```suggestion
codegen-units = 1

```

Razican:
This looks good :) I would run a `cargo update` before merging (and should we re-generate the data file?), and add that end-of-file new line in `Cargo.toml`, but for the rest, I think we can merge it!

Thank you @sffc for your fast response!

jedel1043 (approved):
Ran datagen but git didn't detect any diff. Everything good!

HalidOdat (approved):
Looks good to me! :)

PR #2827, opened 2023-04-16
[Merged by Bors] - Direct array element access on `ByValue` instructions

HalidOdat:
Most of the time that we have a `ByValue` ( `[ value ]` syntax ) it is for arrays and the value is usually an index. This PR adds a fast path to the instructions (without calling `.borrow()` on the object to check if its an array)

For example, this code:
```js
let a = [1, 2, 3]

for (let i = 0 ; i < 10000000; ++i) {
  a[i % 3] += a[ (i + 1) % 3 ]
}
```
Using `hyperfine`, it ran `1.38` times faster on this PR.

```bash
Benchmark 1: ./boa_main test.js
  Time (mean ± σ):     16.504 s ±  0.192 s    [User: 16.440 s, System: 0.020 s]
  Range (min … max):   16.328 s … 16.938 s    10 runs

Benchmark 2: ./boa_direct_array_access test.js
  Time (mean ± σ):     11.982 s ±  0.038 s    [User: 11.939 s, System: 0.013 s]
  Range (min … max):   11.914 s … 12.035 s    10 runs

Summary
  './boa_direct_array_access test.js' ran
    1.38 ± 0.02 times faster than './boa_main test.js'
```

### DIFF
--- a/boa_engine/src/vm/opcode/get/property.rs
+++ b/boa_engine/src/vm/opcode/get/property.rs
@@ -56,6 +56,23 @@ impl Operation for GetPropertyByValue {
         };
 
         let key = key.to_property_key(context)?;
+
+        // Fast Path
+        if object.is_array() {
+            if let PropertyKey::Index(index) = &key {
+                let object_borrowed = object.borrow();
+                if let Some(element) = object_borrowed
+                    .properties()
+                    .dense_indexed_properties()
+                    .and_then(|vec| vec.get(*index as usize))
+                {
+                    context.vm.push(element.clone());
+                    return Ok(CompletionType::Normal);
+                }
+            }
+        }
+
+        // Slow path:
         let result = object.__get__(&key, value, context)?;
 
         context.vm.push(result);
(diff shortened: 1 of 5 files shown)


boa_engine/src/vm/opcode/get/property.rs
```diff
@@ -110,6 +127,24 @@ impl Operation for GetPropertyByValuePush {
...
+                    .and_then(|vec| vec.get(*index as usize))
+                {
+                    context.vm.push(key);
+                    context.vm.push(element.clone());
+                    return Ok(CompletionType::Normal);
+                }
+            }
+        }
+
+        // Slow path:
```
Razican, line 147: We should probably add this comment in line 76 for consistency

Razican (approved):
Good optimization!!

boa_engine/src/vm/opcode/get/property.rs
```diff
@@ -56,6 +57,23 @@ impl Operation for GetPropertyByValue {
         };
 
         let key = key.to_property_key(context)?;
+
+        // Fast Path
+        if std::ptr::eq(object.vtable(), &ARRAY_EXOTIC_INTERNAL_METHODS) {
```
jedel1043, line 62: Wouldn't be better to add an `as_array` method to `JsObject` that does this check internally? That should deduplicate this check.

raskad:
To me this looks like it should be faster, but the "Array creation" benchmark seems to be slower in the ci. I checked it locally and also got a pretty consistent +12% on that one. Do you know why that may be @HalidOdat?

HalidOdat:
> To me this looks like it should be faster, but the "Array creation" benchmark seems to be slower in the ci. I checked it locally and also got a pretty consistent +12% on that one. Do you know why that may be @HalidOdat?

:confused: Yes, It's weird since the benchmark before it have a `-14%` on array benchmarks, and I didn't change something that should effect performance, will rebase this and see if it persists :)

HalidOdat:
Will fix the bug then mark this as ready for review :)

raskad (approved):
Looks very good now

PR #2821, opened 2023-04-14
[Merged by Bors] - Implement async functions using generators

jedel1043:
This should hopefully fix more async/futures issues related to resuming execution in the future, since we can leverage generator logic to handle this for us.

It changes the following:

- Refactors `GeneratorContext` to handle context preparation.
- Reuses the functionality of `GeneratorContext` in `Await`.
- Removes `EarlyReturnType` in favour of a single `r#await` bool flag in `CallFrame`.

PR #2837, opened 2023-04-17
[Merged by Bors] - Improve strictness of `GeneratorState`

jedel1043:
Just some small improvements that increase the strictness of our generator state handling.

Also rollbacks the implementation of `GeneratorValidate` because I forgot to remove it after I did modifications to #2821, and it doesn't make sense to have that if it isn't used by async functions.


raskad (approved):
Nice improvement

HalidOdat (approved):
Looks good to me! :)

PR #2838, opened 2023-04-18
[Merged by Bors] - Fix sync generator yield expressions

raskad:
Depends on #2837.

- Fix the remaining `language/expressions/yield` tests.
- Align the sync generator execution more to the spec.

This breaks one async generator test. We can ignore that one as async generators are currently very broken. I will try to fix async generators next.

### DIFF
--- a/boa_engine/src/builtins/generator/mod.rs
+++ b/boa_engine/src/builtins/generator/mod.rs
@@ -189,13 +189,7 @@ impl Generator {
         context: &mut Context<'_>,
     ) -> JsResult<JsValue> {
         // 1. Return ? GeneratorResume(this value, value, empty).
-        let completion = Self::generator_resume(this, args.get_or_undefined(0).clone(), context);
-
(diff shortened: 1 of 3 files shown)


HalidOdat (approved):
Looks good to me! :)

nekevss:
Nice work! I like the changes overall, any idea why `yield-star-normal-notdone-iter-value-throws.js` broke?

raskad:
> Nice work! I like the changes overall, any idea why `yield-star-normal-notdone-iter-value-throws.js` broke?

I did not look into it too much, because `yield*` in async generators is currently completely broken. But I think it's probably just the changes to where we create the iterator result objects.

I'm currently working on a fix for async generators and proper async generator `yield*`. My branch has 1000+ tests passing currently and I think I may be on a good way to actually fix all async generator tests.

nekevss (approved):
LGTM 😄

PR #2840, opened 2023-04-18
[Merged by Bors] - Implement `Intl.Segmenter`

jedel1043:
The new ICU4X release stabilized the `icu_segmenter` component, so this PR implements `Intl.Segmenter` using that as a base.

Also, I opted for importing `itertools` instead of copy-pasting the implementation of `TupleWindows` because its design is a lot more complex than `Intersperse`, which we copy-pasted previously. Though, I disabled all `std` features of `itertools` to make it a lot more lightweight, so it shouldn't make much difference in compilation times.


raskad (approved):
Looks good!

HalidOdat (approved):
Looks good to me! :)

PR #2846, opened 2023-04-19
[Merged by Bors] - Create a unique `PromiseCapability` on each async function call

jedel1043:

- Creates a new `PromiseCapability` after every async function call instead of sharing the same capability for all calls of the same async function.


raskad (approved):

HalidOdat (approved):
Looks good to me! : )

PR #2847, opened 2023-04-19
[Merged by Bors] - Fix setting properties inside `with` blocks

jedel1043:
This preserves the semantics of the abstract operation `Set` on `with` blocks; setting non-writable properties on non-strict mode just silently fails.


HalidOdat (approved):
Good catch!

Razican (approved):
Thanks!

PR #2848, opened 2023-04-20
[Merged by Bors] - Optimize `String.prototype.normalize`

jedel1043:
We currently use `unicode_normalization` to handle the `String.prototype.normalize` method. However, the crate doesn't support UTF-16 as a first class string, so we had to do some hacks by converting the valid parts of a string to UTF-8, normalizing each one, encoding back to UTF-16 and concatenating everything with the unpaired surrogates within. All of this is obviously suboptimal for performance, which is why I leveraged the `icu_normalizer`, which does support UTF-16 input, to replace our current implementation.

Additionally, this allows users to override the default normalization data if the `intl` feature is enabled by providing the required data in the `BoaProvider` data provider.


raskad (approved):
Having the minimal data generated like this seems like a very nice solution. Looks very nice!

raskad:
@jedel1043 I did not look into it much, do you think we could do the same for the `UnicodeProperties` `ID_Start` and `ID_Continue` that we need in `boa_parser` and currently generate tables ourselves in `boa_unicode`?

jedel1043:
> @jedel1043 I did not look into it much, do you think we could do the same for the `UnicodeProperties` `ID_Start` and `ID_Continue` that we need in `boa_parser` and currently generate tables ourselves in `boa_unicode`?

Yep! There's the `icu_properties` crate that offers precisely that functionality. I'll make a PR this weekend :)

HalidOdat (approved):

PR #2857, opened 2023-04-21
Implement runtime limits for loops

HalidOdat:
Depends on #2866

Related to #2350 

This PR adds runtime limits on all loops (`while`, `do..while`, `for`, `for-in`, and `for-of`) that can be controlled at runtime, ~~this is feature gated under the `"runtime-limits"` feature flag because introduces some minor overhead due to needed iteration tracking.~~

By default when the feature is enabled it is set to `u64::MAX`, which means no runtime limit

It changes the following:
- [x] Add `RuntimeLimits` that contains the runtime limits.
- [x] Add `$boa.limits.loop` getter and setter for easier testing.
- [x] Make the error non-catchable.

Setting the limit from rust:

```rust
context.runtime_limits_mut().set_loop_iteration_limit(1024); // Set to an arbitrary number
```

Or using the `$boa` object for debugging:
```js
$boa.limits.loop = 4 // Set max loop iteration through the debug object $boa

// All the following statements throw a range error with the message:
//
// RangeError: max loop iteration limit 4 exceeded

while (1) {}
do { } while(1)
for (let i = 0; 1; ++i) {}
for (let e of [1, 2, 3, 4, 5]) {} // One above max
for (let e in [1, 2, 3, 4, 5]) {} // One above max
```

### DIFF
--- a/boa_cli/src/debug/limits.rs
+++ b/boa_cli/src/debug/limits.rs
@@ -1,0 +1,35 @@
+use boa_engine::{
+    object::{FunctionObjectBuilder, ObjectInitializer},
+    property::Attribute,
+    Context, JsArgs, JsObject, JsResult, JsValue, NativeFunction,
+};
+
+fn get_loop(_: &JsValue, _: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+    let max = context.runtime_limits().loop_iteration_limit();
+    Ok(JsValue::from(max))
+}
+
+fn set_loop(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+    let value = args.get_or_undefined(0).to_length(context)?;
+    context.runtime_limits_mut().set_loop_iteration_limit(value);
+    Ok(JsValue::undefined())
+}
+
+pub(super) fn create_object(context: &mut Context<'_>) -> JsObject {
+    let get_loop = FunctionObjectBuilder::new(context, NativeFunction::from_fn_ptr(get_loop))
+        .name("get loop")
+        .length(0)
+        .build();
+    let set_loop = FunctionObjectBuilder::new(context, NativeFunction::from_fn_ptr(set_loop))
+        .name("set loop")
+        .length(1)
+        .build();
+    ObjectInitializer::new(context)
+        .accessor(
+            "loop",
+            Some(get_loop),
+            Some(set_loop),
+            Attribute::WRITABLE | Attribute::CONFIGURABLE | Attribute::NON_ENUMERABLE,
+        )
+        .build()
+}
--- a/boa_engine/src/context/mod.rs
+++ b/boa_engine/src/context/mod.rs
@@ -35,6 +35,8 @@ use boa_interner::{Interner, Sym};
 use boa_parser::{Error as ParseError, Parser};
 use boa_profiler::Profiler;
 
+use crate::vm::RuntimeLimits;
+
 /// ECMAScript context. It is the primary way to interact with the runtime.
 ///
 /// `Context`s constructed in a thread share the same runtime, therefore it
@@ -482,30 +484,36 @@ impl<'host> Context<'host> {
 
     /// Set the value of trace on the context
     #[cfg(feature = "trace")]
+    #[inline]
     pub fn set_trace(&mut self, trace: bool) {
         self.vm.trace = trace;
     }
 
     /// Get optimizer options.
+    #[inline]
     pub const fn optimizer_options(&self) -> OptimizerOptions {
         self.optimizer_options
     }
     /// Enable or disable optimizations
+    #[inline]
     pub fn set_optimizer_options(&mut self, optimizer_options: OptimizerOptions) {
         self.optimizer_options = optimizer_options;
     }
 
     /// Changes the strictness mode of the context.
+    #[inline]
     pub fn strict(&mut self, strict: bool) {
         self.strict = strict;
     }
 
     /// Enqueues a [`NativeJob`] on the [`JobQueue`].
+    #[inline]
     pub fn enqueue_job(&mut self, job: NativeJob) {
         self.job_queue().enqueue_promise_job(job, self);
     }
 
     /// Runs all the jobs in the job queue.
+    #[inline]
     pub fn run_jobs(&mut self) {
         self.job_queue().run_jobs(self);
         self.clear_kept_objects();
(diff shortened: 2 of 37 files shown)


raskad:
I have not done a complete review, but I saw the new opcodes and though; can we not keep the limit counts in the `EnvStackEntry`s and manage them in the `LoopStart`, `LoopContinue` and `LoopEnd` opcodes?

HalidOdat:
> I have not done a complete review, but I saw the new opcodes and though; can we not keep the limit counts in the `EnvStackEntry`s and manage them in the `LoopStart`, `LoopContinue` and `LoopEnd` opcodes?

Thanks for the suggestion! It's probably the best place to put it! I realized this when I was half way through implementing the initial POC :laughing:

HalidOdat:
> Can we set the `stack_size_limit` to `1024` (or another value that currently works) in our tests and the boa_tester? That way we use the feature ourselves and we may also catch bugs that spill the stack.

It set to `1024` but It's hidden under the feature flag, Maybe this should be a opt-out feature? So we also prevent infinite loops (especially in the CI and boa_tester)

Razican:
> > Can we set the `stack_size_limit` to `1024` (or another value that currently works) in our tests and the boa_tester? That way we use the feature ourselves and we may also catch bugs that spill the stack.
> 
> It set to `1024` but It's hidden under the feature flag, Maybe this should be a opt-out feature? So we also prevent infinite loops too (especially in the CI and boa_tester)

Does it affect performance? If it's negligible, I would say we should use an `Option<NonZeroU64>>` to handle it, and make it `None` by default (or even `Some(NonZeroU64::MAX)`). Then, we  could have an `unset()` function, to disable runtime limits. I think it makes sense to limit it by default, and not have a feature on it.

boa_engine/src/error.rs
```diff
@@ -673,6 +755,11 @@ impl JsNativeError {
     /// assert!(error_obj.borrow().is_error());
     /// assert_eq!(error_obj.get("message", context).unwrap(), "error!".into())
     /// ```
+    ///
+    /// # Panics
+    ///
+    /// If the converting a [`JsNativeErrorKind::RuntimeLimit`] to an opaque object.
```
Razican, line 761: This also happens with `NoInstructionsRemain`, right?

Razican (commented):
I'm missing a test, where we run a loop without limits for 10 iterations (as an example) and it works, then we set the limit to 5 iterations, try to run it again, and check that we get the error.

boa_engine/src/vm/call_frame/env_stack.rs
```diff
@@ -3,7 +3,11 @@
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub(crate) enum EnvEntryKind {
     Global,
-    Loop,
+    Loop {
+        /// This is used to keep track of how many iterations a loop has done.
+        #[cfg(feature = "runtime-limits")]
+        iteration_count: u64,
```
nekevss, line 9: Just curious on what your thoughts might be about setting this as a `u16` vs. `u64`.

HalidOdat:
> Does it affect performance? If it's negligible, I would say we should use an Option<NonZeroU64>> to handle it, and make it None by default (or even Some(NonZeroU64::MAX)). Then, we could have an unset() function, to disable runtime limits. I think it makes sense to limit it by default, and not have a feature on it.

Yes, did some benchmarks and it wasn't showing that it had any effect on performance, and it makes sense to allow limiting by default.

boa_engine/src/error.rs
HalidOdat: When we enable `"fuzz"` feature we use with the fuzz targets and this shouldn't be reached, that why it throws an `unreachable!()`, unlike `RuntimeLimit` which uses a `panic!`.

boa_engine/src/vm/call_frame/env_stack.rs
HalidOdat: I set it because I felt like that was too restrictive `[0, 65535]`for `u16`, `u32` with `[0, 4294967295]` is better but it is possible to iterate over it (takes like a second on node with jitting), but `u64` should be more than enough for everyone `[0, 18446744073709551615]` :D

nekevss: 😆 I'd agree that u64 is more than enough. 

I mostly brought it up because of how large u64 is. It may be worth keeping track or maybe getting a poll/input to see if people are setting runtime limits past a certain point and see if we should lower to a `u32` or something. But that could just be me being overly concerned and thinking about it completely wrong.

HalidOdat:
Added an example in `boa_examples` for loop limits also removed the feature flag made it default,

I stuck with using `u64` instead of `Option<NonZeroU64>` for loop iteration count because it would introduce conditional jumps. The way we know if there is a limit set, is by checking if it's set to `u64::MAX`.  Changed the condition to check for `>` instead of `>=`, this allows us to exploit a property of comparisons and unsigned integers that `u64::MAX` is not greater than `u64::MAX` , adding 1 to `u64::MAX` overflows to `0`. Now when we increment the counter we do a `wrapping_add`.

boa_engine/src/context/mod.rs
```diff
@@ -550,6 +552,21 @@ impl<'host> Context<'host> {
     pub fn job_queue(&self) -> MaybeShared<'host, dyn JobQueue> {
         self.job_queue.clone()
     }
+
+    /// Get the [`RuntimeLimits`].
+    pub fn runtime_limits(&mut self) -> RuntimeLimits {
+        self.vm.runtime_limits
+    }
```
Razican, line 559: Do we need a mutable reference to `self` here? Wouldn't it be enough with a non-mutable reference?

boa_examples/src/bin/runtime_limits.rs
Razican, line 1: We could have this as a test, in fact.

Razican (approved):
Looks good to me :) check my suggestions and see if they make sense

boa_engine/src/vm/runtime_limits.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Max loop iterations before an error is thrown.
+    loop_iteration_limit: u64,
+}
+
+impl Default for RuntimeLimits {
+    #[inline]
+    fn default() -> Self {
+        Self {
+            loop_iteration_limit: u64::MAX,
```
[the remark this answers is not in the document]
HalidOdat: For loops I think there shouldn't be a limit set, unlike recursion and memory which cause a stack overflow or OOM, infinity loops don't crash the program.

boa_cli/src/debug/limits.rs
```diff
@@ -0,0 +1,35 @@
+use boa_engine::{
+    object::{FunctionObjectBuilder, ObjectInitializer},
+    property::Attribute,
+    Context, JsArgs, JsObject, JsResult, JsValue, NativeFunction,
+};
+
+fn get_loop(_: &JsValue, _: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+    let max = context.runtime_limits_mut().loop_iteration_limit();
```
nekevss, line 8: Should this be an immutable `runtime_limits()`?

nekevss (commented):
Everything looks just about good to me 😄

nekevss (approved):

raskad (approved):
Nice addition!

PR #2858, opened 2023-04-22
[Merged by Bors] - Catch 'eval' and 'arguments' in setter method parameter

raskad:

- Catch 'eval' and 'arguments' in setter method parameter


jedel1043 (approved):
Nice catch!

HalidOdat (approved):
Good catch! Looks good to me!

PR #2859, opened 2023-04-23
Implement `static` shapes

### LINKED ISSUE #2813: Static Shapes (`Hidden Classes`)
This is a possible optimization to improve `Context` construction, the biggest bottleneck is builtin initialization and as we add more builtins, the time needed to initialize them will increase, so I thought of the following optimization:

When we initialize builtins we always construct them in the same way, with the same property names and slot index and attributes. In a way the builtin shapes are static, It only changes when the user wants to delete or extend the builtin `Object.someNonBuiltInFunction = ...`.

Why not pre-initialize them at compile time, having a `static` `phf` table that contains the shape (its property name and descriptor attributes), and in theory a perfect hash function should be faster, since there are no collisions and also take up less space. This also has the big advantages that it can be shared between builtins from different realms,  since it's `static` it can also be shared across threads.

The only thing that has to be constructed at runtime is the object storage (`Vec<JsValue>`) which contains the values of the property names. And In the future when we implement #1372 we will be able to eliminate this overhead too :)

And if we want to extend or delete a property, we convert it into a unique shape (copy-on-write) with the changes (as we are already doing by default with all the builtins).

HalidOdat:
~~Depends on #2723~~

This Pull Request fixes/closes #2813 .

This PR is currently a WIP that implements #2813 , It currently ~~only is applied to `JSON` builtin object (because it has less properties, so easier to test with).~~ is applied to ~~some~~ all builtins objects.

~~This points to #2723 for easier review, will make it point to main one #2723 is merged!~~

Main changes:

- Added the `boa_builtins` crate that generates and stores the static shapes (this is done because `phf!` macro does not work with custom `enum`s). Maybe this is a good opportunity to try to split builtins from `boa_engine` crate to reduce compilation times.
- Because we store the shape in a `static` that can be shared across realms (and threads). I had to put the prototype in storage properties at `len() - 1` (last property), could also store it in the first but it would make #2796 a bit harder and when converting to unique on shape change we would have to shift the storage elements instead of a single `pop` (will have to do further research on what would be the best choice)
- When creating the object storage pre-allocate exact storage (to avoid `Vec` realloc)

Using the debug object we can see that the static shape of `JSON` object is shared between `Realm`s.
```js
$boa.shape.type(JSON) // "static"

// All initial JSON builtins share the static shape
let otherRealmJSON = $boa.realm.create()
$boa.shape.id(JSON) === $boa.shape.id(otherRealmJSON) // true
```

This may be a good way to identify if a global builtin object has been modified, which could allow us some interesting optimizations in some cases.

TODO:

- [ ] Operations
  - [x] Property Lookup
  - [x] Implement property insertion (convert to `unique`)
  - [x] Implement property deletion (convert to `unique`)
  - [x] Implement attribute change (convert to `unique`)
  - [x] Implement prototype change (convert to `unique`)
  - [ ] Optimize operations
- [ ] Add documentation
- [x] Add to builtin objects (eliminated `188` unique shapes per realm, replaced with reusable static shapes)
  - [x] JSON (`1`)
  - [x] Math (`1`)
  - [x] Reflect (`1`)
  - [x] Object (`2`, constructor, prototype )
  - [x] Array (`2`, constructor, prototype )
  - [x] Function (`2`, constructor, prototype )
  - [x] Date (`2`, constructor, prototype )
  - [x] Number (`2`, constructor, prototype )
  - [x] Boolean (`2`, constructor, prototype )
  - [x] BigInt (`2`, constructor, prototype )
  - [x] Symbol (`2`, constructor, prototype )
  - [x] String (`2`, constructor, prototype )
  - [x] RegExp (`2`, constructor, prototype )
  - [x] Symbol (`2`, constructor, prototype )
  - [x] Error (`2`, constructor, prototype )
  - [x] NativeErrors (Type, Reference, Syntax, Aggreagate, Range, Uri, Eval ) (`14`, constructor, prototype ) - Share the same `NATIVE_ERROR` static shape
  - [x] Set (`2`, constructor, prototype )
  - [x] Map (`2`, constructor, prototype )
  - [x] ...


Razican:
Should we already implement static shapes for all builtins, or should we just merge it as-is and enhance it later?

It's already a pretty big PR.

Also, at some point it might be useful to give this API to our users for custom builtins.

HalidOdat:
> Should we already implement static shapes for all builtins, or should we just merge it as-is and enhance it later?

We could merge it as-is, would also like some feedback on the implementation :)

In the mean time I'll work adding documentation and cleaning up the code for easier review.

raskad:
I gave this a quick review. I like most of the implementation. Everything is pretty good to read and to understand. I have some suggestion and questions:

- I think we should try to move the static shapes to static rust code and remove the build.rs. That way the code is easier to read and we have it properly in the git history. Also if we decide we want to move additional stuff to the `boa_builtins` create, that should be a lot easier.
- One thing that is a bit scary is the implicit index based relationship between the static shapes and the builtins in `boa_engine`. But I think for now this should be fine and if we can move the builtins, we can probably make that safer.

On the question on when to merge this; @HalidOdat as far as I can see, if all builtins have their static shapes, some of the builder code in `boa_engine` could be removed, right? If that is the case, maybe it makes sense to have all builtins ready.
@Razican I think the size of the PR is okay. Most LOC changes just some from the static shape definitions and the changes to all builtin builders.

HalidOdat:
> I think we should try to move the static shapes to static rust code and remove the build.rs.

Unfortunately, this isn't possible because to be able to generate the `phf` tables requires us to use `build.rs` or to have a binary generate it because of https://github.com/rust-phf/rust-phf/issues/196. It creates a special `key` value and `displacement` array that it can use.

I've been thinking of a solution, where we remove the `phf` completely, and have and with index lookups, this should be much faster than hashing and then displacement which `phf` does.

But the current and new approach that I'm experimenting would be very error prone to create without `build.rs` or a bin crate to create it.

In the code I did have a `TODO` comment putting into git history, the way I though of doing this, is like we do `boa_icu_provider` where there is an out folder that is put in git history.

an interesting idea is to have a simple manifest file (maybe toml) where it describes the buildin static shapes:

```toml
[Map.prototype]
set = { writable = true, enumerable = false, configurable = true }
# ...
```
And the `build.rs` only generates when this file is newer than the code in `out` dir. (I'm just throwing ideas :laughing: )

> One thing that is a bit scary is the implicit index based relationship between the static shapes and the builtins in boa_engine. But I think for now this should be fine and if we can move the builtins, we can probably make that safer.

Yes, that has one of the things that should be improved. currently if for example a property is missing it will trigger an debug assertion on context initialization making it easy to be catch it early.

One idea I had has to generate a wrapper builder with the static shapes in the `build.rs` using the manifest file.

```rust
boa_builtins::MapBuilder::new(realm, object)
.method_set(Self::set)
.property_name(...)
// ...
.build();
```

This would be a very light wrapper around `BuitInConstructor` where methods.

```rust
/// `Map.prototype.set()`  
///
/// `[Writable]: true` ...
///
/// Auto generated comments from manifest.
fn method_set(&mut self, ...) {
   // The index is known, we don't have to worry about in what order we call.
   const index: usize = 0;
   self.builder.method(index, ... /* pass the values */ ) // set's property on storage `vec`
}
```

> On the question on when to merge this; @HalidOdat as far as I can see, if all builtins have their static shapes, some of the builder code in boa_engine could be removed, right? If that is the case, maybe it makes sense to have all builtins ready.

Yes, the intention was to remove the old builtin constructors in favor of the current direct initialization, I have converted the majority to the new builtin builder, only have as small amount that are left (that's why it hasn't been removed)

raskad:
> In the code I did have a `TODO` comment putting into git history, the way I though of doing this, is like we do `boa_icu_provider` where there is an out folder that is put in git history.

Yeah I like that approach better than the build.rs. If we generate both the static shapes and the builders, IMO we don't need the additional wrapper because all the relevant code is generated anyway.

But for the purpose of this PR, I think the dependency between the shapes and the builder order is fine.

boa_builtins/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+pub struct StaticString {
+    index: u16,
+}
+
+impl StaticString {
+    #[inline]
+    pub fn index(self) -> u16 {
+        self.index
+    }
+}
```
jedel1043, line 35: Maybe we should extract all our value types into a `boa_values` crate. That should allow us to use `JsString` and `JsSymbol`'s const methods directly, instead of defining this.

jedel1043:
Also, I'm pretty sure we can leverage procedural macros to do the code generation for us, instead of doing it manually. We just define good ol' Rust structs with the properties of the builtin, maybe adding tags to the getters and setters or using the `bitflags` approach if only being able to use property names and types seems a bit restrictive, and the macro takes care of generating the phf and the shape for us.

jedel1043:
And now that I think about it, couldn't we just go all in and proc generate all our builtins? That would allow us to both simplify our builtin development logic and unify the static shape creation with the builtin initialization!

Razican:
> And now that I think about it, couldn't we just go all in and proc macro generate all our builtins? That would allow us to both simplify our builtin development logic and unify the static shape creation with the builtin initialization!

If we do this, it would be great if we allow this to be used by custom builtins created by our users.

HalidOdat:
Finished converting all the builtins to have static shapes, in total eliminated `188` unique shapes per `Realm`. There is only one unique shape left on the global object, since it can be overridden with `HostHooks` and changes often.

jedel1043:
Blocked it on https://github.com/boa-dev/boa/pull/2767, since it seems like they're both a bit incompatible with each other. @HalidOdat will determine if this should be closed after finishing the implementation of the other PR.

HalidOdat:
After doing some experimentation with the merged in inline cache I don't think it's worth it, so closing this!

PR #2860, opened 2023-04-23
[Merged by Bors] - Initialize `var` bindings in runtime environments with `undefined`

### LINKED ISSUE #2760: cannot access 'i' before initialization
**Describe the bug**

I am trying to use an npm package called [elliptic](https://www.npmjs.com/package/elliptic). This is a popular library, and with that knowledge it is a safe assumption that it works well in the common JS engines. I have tested it locally in Node.js and it works fine.

Unfortunately when trying to import the most basic functionality in Boa, this error is thrown:

```
cannot access 'i' before initialization
```

I have looked through the source code, and the variable referenced in the error is always declared with `var` and never `const` or `let`, so I don't think this error should be thrown.

**To Reproduce**

Just throw this code into a Boa script and execute it: https://github.com/indutny/elliptic/blob/master/dist/elliptic.min.js

The error will be thrown.

**Expected behavior**

This error should not be thrown, and the code should work just fine.

I don't believe this information is relevant.

- OS: Ubuntu
- Version: 20.04
- Target triple: wasm32-unknown-unknown and probably all triples
- Rustc version: rustc 1.66.1 (90743e729 2023-01-10)

**Additional context**

I believe this issue has a high likelihood of being related: https://github.com/boa-dev/boa/issues/1393

### LINKED ISSUE #2779: Uncaught "ReferenceError": "t is not initialized"
**Describe the bug**
compile error

**To Reproduce**
https://gist.github.com/jedel1043/38ca0c861bf838174fc7e57b92d8423a

**Expected behavior**
return true,but got an `Uncaught "ReferenceError": "t is not initialized"`

- https://boajs.dev/boa/playground/

**Additional context**

### LINKED ISSUE #2795: var hoisting in functions is not working
**Describe the bug**
Javascript supports "var hoisting", this means that `var` declaration can be located anywhere in the function/global scope. As of now it looks like var hoisting is supported only in the global scope but not in function scope.

**To Reproduce**
```js
function a() {
    tmp = "hello";
    console.log(tmp);

    var tmp;
}
a();
```

**Expected behavior**
The code snippet should declare `tmp` **before any code is executed**, then set it's value to `"hello"` and log it to the console.

[var hoisting on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting)

raskad:
Currently in draft, because it breaks some of the annexB [`Block-Level Function Declarations Web Legacy Compatibility Semantics`](https://tc39.es/ecma262/#sec-block-level-function-declarations-web-legacy-compatibility-semantics) tests.

This Pull Request fixes #2795, fixes #2779 and fixes  #2760.

It changes the following:

- Initialize `var` bindings in runtime environments with `undefined`
- Add a missing environment at class construction time.

### DIFF
--- a/boa_engine/src/bytecompiler/class.rs
+++ b/boa_engine/src/bytecompiler/class.rs
@@ -113,16 +113,13 @@ impl ByteCompiler<'_, '_> {
         self.emit(Opcode::GetFunction, &[index]);
         self.emit_u8(0);
 
-        let class_env = if let Some(class_name) = class.name() {
-            if class.has_binding_identifier() {
+        let class_env: Option<(super::Label, super::Label)> = match class.name() {
+            Some(name) if class.has_binding_identifier() => {
                 self.push_compile_environment(false);
-                self.create_immutable_binding(class_name, true);
+                self.create_immutable_binding(name, true);
                 Some(self.emit_opcode_with_two_operands(Opcode::PushDeclarativeEnvironment))
-            } else {
-                None
             }
-        } else {
-            None
+            _ => None,
         };
 
         self.emit_opcode(Opcode::Dup);
--- a/boa_engine/src/bytecompiler/class.rs
+++ b/boa_engine/src/bytecompiler/class.rs
@@ -113,6 +113,18 @@ impl ByteCompiler<'_, '_> {
         self.emit(Opcode::GetFunction, &[index]);
         self.emit_u8(0);
 
+        let class_env = if let Some(class_name) = class.name() {
+            if class.has_binding_identifier() {
+                self.push_compile_environment(false);
+                self.create_immutable_binding(class_name, true);
+                Some(self.emit_opcode_with_two_operands(Opcode::PushDeclarativeEnvironment))
+            } else {
+                None
+            }
+        } else {
+            None
+        };
+
         self.emit_opcode(Opcode::Dup);
         if let Some(node) = class.super_ref() {
             self.compile_expr(node, true);
@@ -543,6 +555,13 @@ impl ByteCompiler<'_, '_> {
         }
 
         self.emit_opcode(Opcode::Pop);
+
+        if let Some(class_env) = class_env {
+            let env_info = self.pop_compile_environment();
+            self.patch_jump_with_target(class_env.0, env_info.num_bindings as u32);
+            self.patch_jump_with_target(class_env.1, env_info.index as u32);
+            self.emit_opcode(Opcode::PopEnvironment);
+        }
 
         if !expression {
             self.emit_binding(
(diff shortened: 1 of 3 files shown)


jedel1043:
> Currently in draft, because it breaks some of the annexB [`Block-Level Function Declarations Web Legacy Compatibility Semantics`](https://tc39.es/ecma262/#sec-block-level-function-declarations-web-legacy-compatibility-semantics) tests.

I'm pretty sure we currently don't have support for that part of the annex, which probably means that the regressed tests were false positives before the change, right?

raskad:
> > Currently in draft, because it breaks some of the annexB [`Block-Level Function Declarations Web Legacy Compatibility Semantics`](https://tc39.es/ecma262/#sec-block-level-function-declarations-web-legacy-compatibility-semantics) tests.
> 
> I'm pretty sure we currently don't have support for that part of the annex, which probably means that the regressed tests were false positives, right? If that's the case, I think we can merge this with no problems.

Yeah, sort of. Im currently at +235 -291 with all broken tests annexB and some more "normal" tests fixed. I'm currently trying to make the annexB stuff work at least partially.

raskad:
There are a couple of refactors and a bit of aligning with the spec before we can fix the annexB tests. So I think @jedel1043 is right, let's go ahead and ship this change to fix the bugs.

jedel1043 (approved):
Looks good to me!

boa_engine/src/bytecompiler/class.rs
```diff
@@ -113,6 +113,18 @@ impl ByteCompiler<'_, '_> {
         self.emit(Opcode::GetFunction, &[index]);
         self.emit_u8(0);
 
+        let class_env = if let Some(class_name) = class.name() {
```
jedel1043, line 116: Could use `class.name().filter(...)` to avoid repetition.

jedel1043 (approved):

boa_engine/src/bytecompiler/class.rs
raskad: I changed it to a `match` because the `filter` looked a bit awkward.

HalidOdat (approved):
Great work!

Razican:
After this, do we need the hoistable re-ordering of nodes?

Razican (approved):

jedel1043:
> After this, do we need the hoistable re-ordering of nodes?

Probably not. We can try to remove that after we merge this and see if anything breaks.

raskad:
> After this, do we need the hoistable re-ordering of nodes?

We still need it, but I am working on refactoring the relevant parts of the compiler so that they work more like the spec. That refactor should remove the reordering. We also need to remove it to implement the annexB function syntax.

PR #2861, opened 2023-04-23
[Merged by Bors] - Allow `Date` object to store invalid `NativeDateTime`

HalidOdat:
Part of ES5. 

This PR allows `Date` objects to store an invalid `NativeDateTime` as a `i64` and check when `getMinutes` (or other methods) call and check if it's valid, like in the spec

This was failing some `Date` tests, when calling `new Date(8.64e15)` then calling `getTime()` it was returning `NaN` when it should return the passed value `8640000000000000` (as node does)


raskad (approved):
Good catch!

Razican (approved):

PR #2864, opened 2023-04-23
Implement valid `RexExp` character sets in Annex-B

### LINKED ISSUE #2794: regex: Dash character is always recognized as a range operator
**Describe the bug**
Dash (`-`) is recognized as a range operator even in situations where it shouldn't and it should be treated as a simple "match single character" rule.

**To Reproduce**
`/[a-\s]/g` should be treated as matching "a" or "-" or some whitespace ([regexr](https://regexr.com/7bo0j)). Instead the parser throws "Invalid regular expression literal"

HalidOdat:
~~Depends on https://github.com/ridiculousfish/regress/pull/62~~ has been merged :tada: 

The dependency is made to point to the bug fix branch, Once https://github.com/ridiculousfish/regress/pull/62 gets merged and published I will change to the released version :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3334,8 +3334,7 @@ dependencies = [
 [[package]]
 name = "regress"
 version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


HalidOdat:
Waiting for new version to be released, dependabot will alert so closing this PR.

PR #2866, opened 2023-04-24
[Merged by Bors] - Fix incorrect `LoopContinue` instruction in while-do loops

HalidOdat:
While working on #2857 I discovered that while generating the bytecode for `do-while` loops we were emitting an orphan `LoopContinue` that is never executed, which was preventing the error to be thrown when max loop iteration is reached.

This can more easily be identified with it's flowgraph : (`do { 1; } while(0) `)

Main:

<details>

</details>

With this PR:

<details>

</details>


Razican (approved):

nekevss (approved):
Nice catch!

PR #2876, opened 2023-04-27
Make update operations reuse the last found binding locator

jedel1043:
This Pull Request fixes update operations by reusing the lvalue accessed by the update.

It changes the following:

- Creates ops `GetNameAndLocator` and `SetNameByLocator`.
- Use new ops on update operations (`++`, `+=`, `&&=`, etc.) only when the binding is not lexical.


raskad (approved):
Really nice implementation.

Razican (approved):
Looks good to me! thanks!

PR #2878, opened 2023-04-28
[Merged by Bors] - Bugfix/new.target is not understood by the parser as an expression #2793

### LINKED ISSUE #2793: `new.target` is not understood by the parser as an expression
**Describe the bug**
Using `new.target.<something>` throws syntax error.

**To Reproduce**
```js
class A {
    constructor() {
        console.log(new.target.name);
    }
}

const a = new A(); // should log "A"
```

**Expected behavior**
The code snippet should compile and run.

The expected behavior can be achieved using a temporary variable:
```js
class A {
    constructor() {
        let tmp = new.target;
        console.log(tmp.name);
    }
}

const a = new A(); // logs "A"
```

[MDN link for `new.target`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target)

projectnoa:
This Pull Request fixes/closes #2793.

It changes the following:

- Added a condition to the boa_parser/src/parser/expression/left_hand_side/member.rs parse function match operation for the new token that allows for the operation to continue evaluating more tokens when the TARGET keyword follows it.
- Added a test to validate the fix. (Could not figure out the structure of the test suite so it's commented for now). All other tests pass.

Please let me know if there's anything else I can do to improve the fix.

### DIFF
--- a/boa_parser/src/parser/expression/left_hand_side/member.rs
+++ b/boa_parser/src/parser/expression/left_hand_side/member.rs
@@ -82,7 +82,7 @@ where
             TokenKind::Keyword((Keyword::New, false)) => {
                 cursor.advance(interner);
 
-                if cursor.next_if(Punctuator::Dot, interner)?.is_some() {
+                let lhs_new_target = if cursor.next_if(Punctuator::Dot, interner)?.is_some() {
                     let token = cursor.next(interner).or_abrupt()?;
                     match token.kind() {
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Looks great, thanks for the contribution!

nekevss (approved):
This looks good to me! 😄 Thanks for the contribution!

PR #2880, opened 2023-04-29
[Merged by Bors] - Fix `RegExp` constructor return value when pattern is a regexp

HalidOdat:
When `RegExp` constructor is called with a regexp it should return the same regexp, if `NewTarget`'s constructor is the same.


HalidOdat:
The broken test is caused by an unrelated bug that #2881 fixes

Razican (approved):
Thanks!

jedel1043 (approved):
Looks good!

bors r+

PR #2883, opened 2023-04-30
Added unit tests for `boa_ast::Keyword`

Razican:
We have a very low code coverage in our tests, and, while we have a great end-to-end testing framework setup (test262), this doesn't run unit tests for our Rust implementation.

This is the first of a series of PRs that will try to increase our code coverage so that it reaches reasonable levels.

### DIFF
--- a/boa_ast/src/keyword/tests.rs
+++ b/boa_ast/src/keyword/tests.rs
@@ -1,0 +1,354 @@
+#![allow(clippy::cognitive_complexity)]
+
+use super::*;
+
+/// Gets a list of all the keywords.
+fn all_keywords() -> impl Iterator<Item = Keyword> {
+    [
+        Keyword::Await,
+        Keyword::Async,
+        Keyword::Break,
+        Keyword::Case,
+        Keyword::Catch,
+        Keyword::Class,
+        Keyword::Continue,
+        Keyword::Const,
+        Keyword::Debugger,
+        Keyword::Default,
+        Keyword::Delete,
+        Keyword::Do,
+        Keyword::Else,
+        Keyword::Enum,
+        Keyword::Export,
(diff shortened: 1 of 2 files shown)


boa_ast/src/keyword/tests.rs
```diff
@@ -0,0 +1,354 @@
...
+            | Keyword::Void
+            | Keyword::While
+            | Keyword::With
+            | Keyword::Yield => assert!(BinaryOp::try_from(k).is_err()),
+        }
+    }
+}
+
+#[test]
+fn ut_from_str() {
```
HalidOdat, line 342: What does `ut` mean?

HalidOdat (approved):
Looks good to me! :)

Just a small question

boa_ast/src/keyword/tests.rs
Razican: haha "unit test". I tend to prefix unit tests with `ut_`, and integration tests at `it_`. I think we have not done this in Boa until now, so I can remove it if needed, but might make sense.

HalidOdat: To me it seems unneeded, because in rust tests that are defined in the source are unit tests, and tests that are defined externally in the [`tests` dir in root crate level](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) are integration tests.

nekevss (approved):

boa_ast/src/keyword/tests.rs
Razican: While it's clear during development, it might not be so much when running them.

But I get it xD I will just remove it :)

Razican: Prefix removed

PR #2890, opened 2023-05-01
[Merged by Bors] - Refactor guards into a `ContextCleanupGuard` abstraction

jedel1043:
Noticed we were using this pattern on a couple of places, so I abstracted it behind a `ContextCleanupGuard` struct.

Let me know if you remember another place where this pattern would apply.


HalidOdat (approved):
Nice refactor! Looks good to me! :)

raskad (approved):
Makes a lot of sense and makes eval code a lot safer to adjust.

PR #2891, opened 2023-05-02
Docs update for boa_runtime and console documentation

nekevss:
Hi all!

This PR is to update the docs to ease any confusion regarding adding console to a project (RE: the discussion on discord). Feel free to let me know if there are any changes you'd like to see or if a more in depth example is needed.

I can't remember if this had been decided yet, but were we publishing `boa_runtime`? I'm wondering if it would be better to link to `boajs.dev`'s docs for `boa_runtime` ~~vs. the rustdoc intra-doc link~~.

It changes the following:

- Updates the primary example in **boa_engine** to account for console being moved to **boa_runtime**
- Builds out the docs on **boa_runtime**

### DIFF
--- a/boa_engine/src/lib.rs
+++ b/boa_engine/src/lib.rs
@@ -9,6 +9,8 @@
 //!
 //! Try out the most recent release with Boa's live demo [playground][boa-playground].
 //!
+//! For information related to Web API features, please see [`boa_runtime`]
+//!
 //! # Example usage
 //!
 //! You can find multiple examples of the usage of Boa in the [`boa_examples`][examples] crate. In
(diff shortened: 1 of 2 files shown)


boa_engine/src/lib.rs
```diff
@@ -19,7 +21,12 @@
 //! ```
 //! use boa_engine::{Context, Source};
 //!
-//! let js_code = "console.log('Hello World from a JS code string!')";
+//! let js_code = r#"
```
Razican, line 24: To make sure that we don't miss this again, I would add, in line 45, a panic (hidden with `#`), so that we catch the error.

Razican (approved):
Looks great to me :) I would just add an extra panic so that we get notified if this fails for whatever reason.

Razican:
> I can't remember if this had been decided yet, but were we publishing boa_runtime? I'm wondering if it would be better to link to `boajs.dev`'s docs for `boa_runtime` ~vs. the rustdoc intra-doc link~.

I think the idea is to publish boa_runtime, in case someone wants to use it as-is. It would also allow us to improve it and to add more things to it. I also now that some of our users would like to have a runtime provided by Boa, so we can just point to that if it's in crates.io.

boa_engine/src/lib.rs
nekevss: Updated 😄

HalidOdat (approved):
Looks good to me! :)

PR #2899, opened 2023-05-05
Enable github queues and remove bors.toml

jedel1043:
Apparently we need to mark the actions needed for the github queue, so this PR does that. It also removes the `bors.toml` file since it's not needed anymore.


nekevss (approved):

Razican (approved):
Thanks!

PR #2902, opened 2023-05-05
Make `typeof` throw when accessing uninitialized variables

raskad:

- Make `typeof` throw when accessing uninitialized variables

Currently `typeof` returns `undefined` when accessing uninitialized bindings like `const`, `let`, `class`.


Razican (approved):
Nice change!

jedel1043 (approved):
Good catch :)

PR #2903, opened 2023-05-05
Fix value property assignment order

jedel1043:
This PR reorders assignments evaluation to conform to the spec... however, it seems to be undecided if the new assignment semantics are intended or a bug on the spec (see https://github.com/tc39/ecma262/issues/2659 and https://github.com/tc39/test262/issues/3407).

If it's indeed a bug on the spec, this should be closed in favour of the current semantics. Otherwise, this can be merged.

Per the previous explanation, this is blocked on https://github.com/tc39/ecma262/issues/2659)


jedel1043:
Closing since it was decided that the old behaviour was the correct one.
https://github.com/boa-dev/boa/pull/2903

PR #2904, opened 2023-05-05
Implement runtime limits for recursion

### LINKED ISSUE #809: Implement call stack size tracking
**ECMAScript feature**
Although not strictly in the ECMAScript spec (at least as far as I can find), most engines have a system to track the call stack size and throw an error if it becomes too large. This prevents stack overflows. However, Boa has not implemented this feature, hence why things like this happen:
image
If call stack size tracking was added, this would help the engine not crash, which is (clearly) a good thing.

**Example code**
```js
function a() {
  a();
}
a();
```

HalidOdat:
This Pull Request fixes/closes #809 and related to #2350 .

It changes the following:

- [x] Adds runtime limits on recursion
- [x] Adds getter & setter to `$boa.limits` module for debugging
- [x] Add test and example

Like #2857 the recursion limit can be controlled through the `.runtime_limits()`:
```rust
context.runtime_limits_mut().set_recursion_limit(123); // Set to an arbitrary number
```

Or using the `$boa` object for debugging:
```js
$boa.limits.recursion = 123

function x() {
    x()
}

x() // RuntimeLimit: maximum recursion limit 123 exceeded
```

Tried different values `400` seemed to not cause a stack-overflow, while `500` did, so I set it to `400`. This is in debug mode, In release we could probably set it to a higher value.

Removed `"tail-call-optimization"` feature from test262 ignore list, because it no longer causes a stack-overflow :)

Currently the way we do calls requires us to recurse in rust, this is not the ideal approach, we need to refactor it! That's why we are limited by our native stack size instead of heap size. This issue does not prevent this PR from being merged though.

I'll work on refactoring the way we call functions in a separate PR :)

### DIFF
--- a/boa_cli/src/debug/limits.rs
+++ b/boa_cli/src/debug/limits.rs
@@ -1,7 +1,7 @@
 use boa_engine::{
     object::{FunctionObjectBuilder, ObjectInitializer},
     property::Attribute,
-    Context, JsArgs, JsObject, JsResult, JsValue, NativeFunction,
+    Context, JsArgs, JsNativeError, JsObject, JsResult, JsValue, NativeFunction,
 };
 
 fn get_loop(_: &JsValue, _: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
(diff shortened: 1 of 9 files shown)


Razican (commented):
It looks great to me! I'm only missing the test and example, as you mention :)

Razican (approved):

raskad (approved):
Looks great!

PR #2906, opened 2023-05-05
Fix remaining object literal tests

raskad:

- Fix object literal methods to not be constructors and not contain a prototype property.
- Refactor creation of constructor functions.
- Remove unused parameters from `create_function_object`.
- Fix generator function creation to always create a prototype property.
- Fix parsing `super` after a `yield`.
- Fix `+In` in computed property names.


HalidOdat (approved):
Great work! Looks good to me! :)

jedel1043 (approved):
Looks great!

PR #2907, opened 2023-05-05
Fix switch `default` execution

HalidOdat:
Fixes switch `default` case execution, when a default case has regular switch cases below it (as in this test262 [test](https://github.com/tc39/test262/blob/72c0c5e16350a76bd41f7a1ceb7702588a2a39c6/test/language/statements/switch/S12.11_A1_T2.js) ) we execute the default case body at the end, we should execute the cases in order.

### DIFF
--- a/boa_ast/src/statement/switch.rs
+++ b/boa_ast/src/statement/switch.rs
@@ -22,7 +22,7 @@ use core::ops::ControlFlow;
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[derive(Clone, Debug, PartialEq)]
 pub struct Case {
-    condition: Expression,
+    condition: Option<Expression>,
     body: StatementList,
 }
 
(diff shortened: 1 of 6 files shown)


boa_engine/src/bytecompiler/mod.rs
```diff
@@ -295,6 +295,7 @@ pub struct ByteCompiler<'ctx, 'host> {
 impl<'ctx, 'host> ByteCompiler<'ctx, 'host> {
     /// Represents a placeholder address that will be patched later.
     const DUMMY_ADDRESS: u32 = u32::MAX;
+    const DUMMY_LABLE: Label = Label { index: u32::MAX };
```
raskad, line 298:
```suggestion
    const DUMMY_LABEL: Label = Label { index: u32::MAX };
```

raskad (approved):
Nice work! Just got a typo :)

boa_ast/src/statement/switch.rs
```diff
@@ -31,14 +31,30 @@ impl Case {
...
     #[must_use]
     pub const fn new(condition: Expression, body: StatementList) -> Self {
-        Self { condition, body }
+        Self {
+            condition: Some(condition),
+            body,
+        }
+    }
+
+    /// Creates a `Case` AST node.
```
nekevss, line 40: Can this doc highlight that this is a default case?

nekevss (commented):
Looks pretty good to me 😄 Just one doc nitpick.

nekevss (approved):
LGTM 😄 Nice work!

PR #2910, opened 2023-05-07
Implement annexB Block-Level Function Declarations

raskad:

- Implement annexB Block-Level Function Declarations
- Remove unused opcode `DefInitArg`
- Combine the similar opcodes `DefLet`, `DefInitLet` and `DefInitConst`

There are only two tests in the `annexB/language/eval-code` and `annexB/language/function-code` suites failing. I will look at fixing them after this.

### DIFF
--- a/boa_engine/src/vm/opcode/define/mod.rs
+++ b/boa_engine/src/vm/opcode/define/mod.rs
@@ -66,57 +66,26 @@ impl Operation for DefInitVar {
     }
 }
 
-/// `DefLet` implements the Opcode Operation for `Opcode::DefLet`
+/// `InitializeLexical` implements the Opcode Operation for `Opcode::InitializeLexical`
 ///
 /// Operation:
-///  - Declare `let` type variable.
+///  - Initialize a lexical binding.
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct DefLet;
+pub(crate) struct InitializeLexical;
 
-impl Operation for DefLet {
-    const NAME: &'static str = "DefLet";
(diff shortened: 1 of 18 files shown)


boa_engine/src/vm/opcode/define/mod.rs
```diff
@@ -66,57 +66,26 @@ impl Operation for DefInitVar {
...
-    const INSTRUCTION: &'static str = "INST - DefLet";
+impl Operation for InitializeLexical {
+    const NAME: &'static str = "InitializeLexical";
+    const INSTRUCTION: &'static str = "INST - InitializeLexical";
 
     fn execute(context: &mut Context<'_>) -> JsResult<CompletionType> {
         let index = context.vm.read::<u32>();
+        let value = context.vm.pop();
         let binding_locator = context.vm.frame().code_block.bindings[index as usize];
         context.vm.environments.put_declarative_value(
```
jedel1043, line 84: Hmm, maybe this should be renamed to `put_lexical_value`?

jedel1043 (approved):
Nice implementation!

HalidOdat (approved):
Very nice work!

Just requires a rebase :)

boa_engine/src/vm/opcode/define/mod.rs
raskad: Do you mean to opcode as `PutLexicalValue`?

jedel1043: No, I mean the `put_declarative_value` -> `put_lexical_value`. Just thought it was a better name for the function, since it's used inside `InitializeLexical`

PR #2915, opened 2023-05-08
Allow `let` name in for-in loop in non-strict mode

HalidOdat:
Allow `let` variable name in `for-in` loop in non-strict mode.

Fixes: https://github.com/tc39/test262/blob/72c0c5e16350a76bd41f7a1ceb7702588a2a39c6/test/language/statements/for-in/identifier-let-allowed-as-lefthandside-expression-not-strict.js


Razican (approved):
Good one!

jedel1043 (approved):

PR #2916, opened 2023-05-08
Allow escaped `let` as expression

HalidOdat:
Fixes: https://github.com/tc39/test262/blob/72c0c5e16350a76bd41f7a1ceb7702588a2a39c6/test/language/statements/let/syntax/escaped-let.js


jedel1043 (approved):
Good catch!

Razican (approved):
Thanks!

PR #2921, opened 2023-05-09
Separate declarative environment kinds

jedel1043:
This Pull Request separates declarative environments into kinds, which makes it easier to add more kinds in the future (spoiler: module environments).

PR #2922, opened 2023-05-10
Implement module execution

### LINKED ISSUE #64: Implement the ECMAScript module system
Things like implement **require/import alias**.

jedel1043:
~Depends on #2921. Will rebase onto `main` when that PR is merged.~  

This Pull Request closes #64. We finally have a module system! 🎉

It changes the following:

- Implements the module execution semantics from the spec.
- Adds a `Module` struct API from which an user can progress the loading, linking and execution of a module.
- Removes most of `Context`'s module APIs in favour of methods in `Module`.
- Adds a `ModuleLoader` trait from where a host can define how to load and cache modules.
- Adds a `ModuleEnvironment` that can contain indirect references to other environment bindings.

There's still more work to do though, like adding support for the `import()` method, fix some parsing and linking bugs, and implementing JSON modules (Stage 3).

Marking it as draft since there's a lot of documentation missing ~(and apparently some bugs)~, but I wanted to put out the PR for reviews on the implementation.


jedel1043:
By the way, the test count reduction is because we were running module tests twice: on no strict and strict mode. However, this is unnecessary for modules because they're strict by default, so we can just run them once.

jedel1043:
Ok, I checked the regressed tests and all of them have the `$DONOTEVALUATE()` function, meaning those were false positives before!

jedel1043:
All documented and added an example, this is ready for full reviews now!

raskad (commented):
I'm still reviewing the execution parts. The ast and parser changes look very good, I just found a few copy-pasted docs.

boa_engine/src/module/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+    module: Module,
+    #[unsafe_ignore_trace]
+    exports: IndexMap<JsString, Sym, BuildHasherDefault<FxHasher>>,
+}
+
+impl ModuleNamespace {
+    /// Abstract operation [`ModuleNamespaceCreate ( module, exports )`][spec].
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-modulenamespacecreate
+    pub(crate) fn create(module: Module, names: Vec<Sym>, context: &mut Context<'_>) -> JsObject {
```
HalidOdat, line 659: It would be nice to have ECMAScript docs in the code as well

boa_engine/src/module/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+
+        namespace.borrow_mut().properties_mut().insert(
+            &PropertyKey::Symbol(JsSymbol::to_string_tag()),
+            PropertyDescriptor::builder()
+                .value(js_string!("Module"))
+                .writable(false)
+                .enumerable(false)
+                .configurable(false)
+                .build(),
+        );
```
HalidOdat, line 688: We could have an `ObjectTemplate` so we direct initialize it and not have to add properties after creation.

boa_engine/src/module/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+                    .collect();
+
+                ModuleNamespace::create(self.clone(), unambiguous_names, context)
+            })
+            .clone()
+    }
+}
+
+impl PartialEq for Module {
+    fn eq(&self, other: &Self) -> bool {
```
HalidOdat, line 632: Since these are `pub` functions I think they should be annotated with `#[inline]`

boa_engine/src/module/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+        Ok(promise)
+    }
+
+    /// Abstract operation [`GetModuleNamespace ( module )`][spec].
+    ///
+    /// Gets the [**Module Namespace Object**][ns] that represents this module's exports.
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-getmodulenamespace
+    /// [ns]: https://tc39.es/ecma262/#sec-module-namespace-exotic-objects
+    pub fn namespace(&self, context: &mut Context<'_>) -> JsObject {
```
HalidOdat, line 609: It would be nice to have spec docs here too :)

HalidOdat (commented):
Looks really good!

boa_examples/src/bin/modules.rs
```diff
@@ -0,0 +1,120 @@
...
+            .canonicalize()?
+            .join("main.mjs"),
+        module.clone(),
+    );
+
+    // The lifecycle of the module is tracked using promises, which can be a bit cumbersome
+    // for simple uses but that use case is better suited by the `Module::load_link_evaluate` method.
+    // This does the full version for demonstration purposes.
+    //
+    // parse -> load -> link -> evaluate
```
nekevss, line 51: A bit nitpicky on the example docs comment, but could the comments/documentation on this section of the example be a bit more step-by-step of what's happening?

Something like:

```rust
    // 
    // The lifecycle goes through 4 main steps: parse -> load -> link -> evaluate
    //
    // Begin with the parsed module
    let promise_result = module
        // load the module
        .load(context)
        // Use a `NativeFunction` to links the module.
        .then(
            Some(
                FunctionObjectBuilder::new(
                    context,
                    NativeFunction::from_copy_closure_with_captures(
                        |_, _, module, context| {
                            module.link(context)?;
                            Ok(JsValue::undefined())
                        },
                        module.clone(),
                    ),
                )
                .build(),
            ),
            None,
            context,
        )?
        // Finally, use a `NativeFunction` to evaluate the module.
        .then(
            Some(
                FunctionObjectBuilder::new(
                    context,
                    NativeFunction::from_copy_closure_with_captures(
                        |_, _, module, context| Ok(module.evaluate(context).into()),
                        module.clone(),
                    ),
                )
                .build(),
            ),
            None,
            context,
        )?;
```

boa_interner/src/sym.rs
```diff
@@ -12,11 +13,15 @@ use core::num::NonZeroUsize;
 )]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[allow(clippy::unsafe_derive_deserialize)]
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Finalize)]
 pub struct Sym {
     value: NonZeroUsize,
 }
 
+unsafe impl Trace for Sym {
```
HalidOdat, line 21: We should probably have a `SAFETY:` comment here

boa_engine/src/object/internal_methods/namespace.rs
HalidOdat, line 1: Maybe we should rename this file to `module_namespace.rs` since it's about module namespace objects.

boa_engine/src/object/internal_methods/namespace.rs
```diff
@@ -0,0 +1,308 @@
...
+        __delete__: delete,
+        __own_property_keys__: own_property_keys,
+        ..ORDINARY_INTERNAL_METHODS
+    };
+
+/// [`[[GetPrototypeOf]] ( )`][spec].
+///
+/// [spec]: https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-getprototypeof
+#[allow(clippy::unnecessary_wraps)]
+fn get_prototype_of(_: &JsObject, _: &mut Context<'_>) -> JsResult<JsPrototype> {
```
HalidOdat, line 39:
```suggestion
fn module_namespace_exotic_get_prototype_of(_: &JsObject, _: &mut Context<'_>) -> JsResult<JsPrototype> {
```

I think we should prefix them with `module_namespace_exotic_` to be consistent with other internal methods implementation, `array_exotic_`, `string_exotic_`, etc.

boa_engine/src/object/internal_methods/immutable.rs
```diff
@@ -0,0 +1,39 @@
+use crate::{
+    object::{JsObject, JsPrototype},
+    Context, JsResult,
+};
+
+use super::{InternalObjectMethods, ORDINARY_INTERNAL_METHODS};
+
+/// Definitions of the internal object methods for [**Module Namespace Exotic Objects**][spec].
+///
+/// [spec]: https://tc39.es/ecma262/#sec-module-namespace-exotic-objects
```
HalidOdat, line 10: Docs should point to Immutable Prototype exotic object

boa_engine/src/object/internal_methods/immutable.rs
HalidOdat, line 1: I think this should be renamed to `immutable_prototype` to better match with the spec naming.

boa_engine/src/object/internal_methods/immutable.rs
```diff
@@ -0,0 +1,39 @@
...
+    InternalObjectMethods {
+        __set_prototype_of__: set_prototype_of,
+        ..ORDINARY_INTERNAL_METHODS
+    };
+
+/// [`[[SetPrototypeOf]] ( V )`][spec].
+///
+/// [spec]: https://tc39.es/ecma262/#sec-immutable-prototype-exotic-objects-setprototypeof-v
+#[allow(clippy::needless_pass_by_value)]
+pub(crate) fn set_prototype_of(
```
HalidOdat, line 21: We should probably prefix with `immutable_prototype_exotic_`

```suggestion
pub(crate) fn immutable_prototype_exotic_set_prototype_of(
```

To be consistent with other internal methods implementations

boa_engine/src/module/mod.rs
```diff
@@ -0,0 +1,734 @@
...
+impl ModuleLoader for SimpleModuleLoader {
+    fn load_imported_module(
+        &self,
+        _referrer: Referrer,
+        specifier: JsString,
+        finish_load: Box<dyn FnOnce(JsResult<Module>, &mut Context<'_>)>,
+        context: &mut Context<'_>,
+    ) {
+        let result = (|| {
+            let path = specifier
```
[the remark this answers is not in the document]
jedel1043: Because this allows deferring the load to a later point in time, making it possible to concurrently load several modules at the same time. This is deliberate, as shown by the [spec](https://tc39.es/ecma262/#sec-HostLoadImportedModule):

> The [host environment](https://tc39.es/ecma262/#host-environment) must perform [FinishLoadingImportedModule](https://tc39.es/ecma262/#sec-FinishLoadingImportedModule)(referrer, specifier, payload, result), where result is either a [normal completion containing](https://tc39.es/ecma262/#sec-completion-record-specification-type) the loaded [Module Record](https://tc39.es/ecma262/#sec-abstract-module-records) or a [throw completion](https://tc39.es/ecma262/#sec-completion-record-specification-type), either ***synchronously or asynchronously***.

Of course, the ideal implementation would be to have `async fn` in dyn traits available. Unfortunately, static `async fn` in traits are a [`WIP`](https://github.com/rust-lang/rust/issues/91611), and `async fn` in dyn traits aren't [even in the nightly compiler yet](https://github.com/rust-lang/rust/issues/107011), but we should be able to improve the API when those features arrive :)

boa_engine/src/module/source.rs
```diff
@@ -0,0 +1,1903 @@
...
+            let requested_module = self.inner.loaded_modules.borrow()[e].clone();
+
+            //     b. Let starNames be requestedModule.GetExportedNames(exportStarSet).
+            //     c. For each element n of starNames, do
+            for n in requested_module.get_exported_names(export_star_set) {
+                //         i. If SameValue(n, "default") is false, then
+                if n != Sym::DEFAULT {
+                    //             1. If exportedNames does not contain n, then
+                    //                 a. Append n to exportedNames.
+                    exported_names.insert(n);
```
HalidOdat, line 550:
```suggestion
            // a. Let requestedModule be GetImportedModule(module, e.[[ModuleRequest]]).
            let requested_module = self.inner.loaded_modules.borrow()[e].clone();

            // b. Let starNames be requestedModule.GetExportedNames(exportStarSet).
            // c. For each element n of starNames, do
            for n in requested_module.get_exported_names(export_star_set) {
                // i. If SameValue(n, "default") is false, then
                if n != Sym::DEFAULT {
                    // 1. If exportedNames does not contain n, then
                    //    a. Append n to exportedNames.
                    exported_names.insert(n);
```
It would be nice to remove the unneeded indent.

boa_engine/src/module/source.rs
```diff
@@ -0,0 +1,1903 @@
...
+    },
+}
+
+// SAFETY: This must be synced with `Status` to mark any new data added that needs to be traced.
+// This implementation is necessary to be able to transition from one state to another by destructuring,
+// which saves us some unnecessary clones.
+unsafe impl Trace for Status {
+    custom_trace!(this, {
+        match this {
+            Status::Unlinked | Status::Linking { .. } | Status::Linked { .. } => {}
```
HalidOdat, line 94: Couldn't we use the `Trace` proc macro with `#[unsafe_ignore_trace]` attribute

HalidOdat (commented):
Just some small nitpicks, the rest of the code looks great!

boa_engine/src/module/source.rs
jedel1043: Not really. The `Trace` proc macro adds a `Drop` implementation to disencourage implementing `Drop` on `Trace` implementors. However, implementing `Drop` disallows destructuring, which is pretty useful to have for state types like `Status`.

boa_engine/src/module/source.rs
jedel1043: Made a pass on all spec comments on the `module` mod and removed unneeded indent

boa_cli/src/main.rs
```diff
@@ -154,6 +156,14 @@ struct Opt {
     /// Inject debugging object `$boa`.
     #[arg(long)]
     debug_object: bool,
+
+    /// Treats the input files as modules.
+    #[arg(long, short = 'm', group = "mod")]
+    module: bool,
+
+    /// Root path from where the module resolver will try to load the modules.
+    #[arg(long, default_value_os_t = PathBuf::from("."), requires = "mod")]
```
nekevss, line 165: Maybe `root` instead of `modpath`. That would bring it more in line with `SingleModuleLoader` and a short `r` could be used as well.

nekevss (commented):
A lot of this looks fantastic! There are still some parts that I'm making my way through, but wanted to ask about the below. 😅

boa_engine/src/module/mod.rs
```diff
@@ -0,0 +1,734 @@
+//! Boa's implementation of the ECMAScript's module system.
+//!
+//! This module contains the [`Module`] type, which represents an [**Abstract Module Record**][module].
+//! Every module roughly follows the same lifecycle:
+//! - Parse using [`Module::parse`].
+//! - Load all its dependencies using [`Module::load`].
+//! - Link its dependencies together using [`Module::link`].
+//! - Evaluate the module and its dependencies using [`Module::evaluate`].
+//!
```
nekevss, line 9: `SimpleModuleLoader` should probably be mentioned earlier in the header doc, right? 

Something like:
```suggestion
//! This module contains the [`Module`] type, which represents an 
//! [**Abstract Module Record**][module], a `ModuleLoader` trait
//! for custom module loader implementations, and a default 
//! `SimpleModuleLoader`.
//!
//! Every module roughly follows the same lifecycle:
//! - Parse using [`Module::parse`].
//! - Load all its dependencies using [`Module::load`].
//! - Link its dependencies together using [`Module::link`].
//! - Evaluate the module and its dependencies using [`Module::evaluate`].
//!
```
Then we can expand on the topics below like you've currently done. 😄

boa_gc/src/internals/gc_box.rs
```diff
@@ -85,9 +85,9 @@ impl fmt::Debug for GcBoxHeader {
 
 /// A garbage collected allocation.
 #[derive(Debug)]
-pub struct GcBox<T: Trace + ?Sized + 'static> {
+pub struct GcBox<T: ?Sized + 'static> {
     pub(crate) header: GcBoxHeader,
-    value: T,
+    pub(crate) value: T,
```
raskad, line 90: I think these changes can be reverted right?

HalidOdat (approved):
Looks perfect to me! Great work!

PR #2924, opened 2023-05-10
Fix ES5 selector

### LINKED ISSUE #2886: ES5 selector in conformance website not working correctly
Some ES6 stuff appears when selecting ES5 in the conformance website, like for-of, computed property names, etc. which have the `es6id` field.

As described by @Razican, the issue is in the way we filter test:

> I think the issue is mostly here: https://github.com/boa-dev/boa/blob/main/boa_tester/src/edition.rs#L342-L357, where we first check async and module flags, then the list of features. But if there is no feature and no flag, it defaults to ES5. Which should only be the case if es5id is found, ideally (or if no es6id, esid is found)

veera-sivarajan:
This Pull Request fixes/closes #2886.

Previously, the code defaulted to ES5 without checking for es6id and esid. This PR selects ES6 if esid or es6id is found. 

Reference: [doc](https://chromium.googlesource.com/external/github.com/tc39/test262/+/HEAD/CONTRIBUTING.md#es6id).

### DIFF
--- a/boa_tester/src/edition.rs
+++ b/boa_tester/src/edition.rs
@@ -341,7 +341,7 @@ impl SpecEdition {
     pub(crate) fn from_test_metadata(metadata: &MetaData) -> Result<Self, Vec<&str>> {
         let mut min_edition = if metadata.flags.contains(&TestFlag::Async) {
             Self::ES8
-        } else if metadata.flags.contains(&TestFlag::Module) {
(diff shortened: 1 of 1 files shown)


HalidOdat (approved):
Looks good to me! :)

Razican (approved):
Thanks!

PR #2925, opened 2023-05-10
Fix tagged template creation

raskad:

- Refactor the creation of tagged templates.
- Add a unique identifier to tagged template parse nodes that is used at runtime to cache them.
- Add an identifier to the parser that is incemented when `eval` code is parsed. This is needed because tagged templates parsed by different `eval` calls are unique.

### DIFF
--- a/boa_engine/src/context/mod.rs
+++ b/boa_engine/src/context/mod.rs
@@ -112,6 +112,9 @@ pub struct Context<'host> {
 
     optimizer_options: OptimizerOptions,
     root_shape: SharedShape,
+
+    /// Unique identifier for each parser instance used during the context lifetime.
+    parser_identifier: usize,
 }
 
 impl std::fmt::Debug for Context<'_> {
@@ -656,6 +659,13 @@ impl Context<'_> {
         std::mem::swap(&mut self.realm, realm);
     }
 
+    /// Get and increment the parser identifier.
+    pub(crate) fn next_parser_identifier(&mut self) -> usize {
+        let identifier = self.parser_identifier;
+        self.parser_identifier += 1;
+        identifier
+    }
+
     /// `CanDeclareGlobalFunction ( N )`
     ///
     /// More information:
@@ -1025,6 +1035,7 @@ impl<'icu, 'hooks, 'queue> ContextBuilder<'icu, 'hooks, 'queue> {
             }),
             optimizer_options: OptimizerOptions::OPTIMIZE_ALL,
             root_shape,
+            parser_identifier: 0,
         };
 
         builtins::set_default_global_bindings(&mut context)?;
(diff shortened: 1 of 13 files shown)


boa_parser/src/parser/expression/left_hand_side/template.rs
```diff
@@ -77,13 +79,19 @@ where
                     )?;
                 }
                 TokenKind::TemplateNoSubstitution(template_string) => {
+                    let mut hasher = FxHasher::default();
+                    hasher.write_u32(start_position.line_number());
+                    hasher.write_u32(start_position.column_number());
+                    hasher.write_usize(cursor.identifier());
```
HalidOdat, line 85: There is a very small but possible possibility that it would have that same hash as a previous `TaggedTemplate`.

A solution would be to divide the `u64` identifier into two parts (`u32: count`, `u32: generation`), where the `count` starts from `0` and increments, and the generation is passed from the context as we do now.

What do you think?

HalidOdat (commented):
Looks really good! Just a small concern :)

boa_parser/src/parser/expression/left_hand_side/template.rs
raskad: Sounds good. I switched to a count and also found some more places where we have to increment the parser id.

boa_parser/src/parser/cursor/mod.rs
```diff
@@ -169,6 +179,22 @@ where
...
+
+    /// Get the identifier for a tagged template.
+    #[inline]
+    pub(super) fn tagged_template_identifier(&mut self) -> u64 {
+        let mut hasher = FxHasher::default();
+        hasher.write_u32(self.identifier);
+        hasher.write_u32(self.tagged_templates_count);
+        self.tagged_templates_count += 1;
+        hasher.finish()
+    }
```
HalidOdat, line 196:
```suggestion
    pub(super) fn tagged_template_identifier(&mut self) -> u64 {
        let identifier = self.identifier as u64;
        let count = self.tagged_templates_count as u64;

        (count << 32) | identifier
    }
```

I meant something like this, where we have the the first `32bits` for the count and the last for the generation/identifier.

raskad: Yeah true, just pushing the u64 is easier. Applied.

HalidOdat (approved):
Looks perfect to me! :)

boa_engine/src/context/mod.rs
```diff
@@ -656,6 +661,12 @@ impl Context<'_> {
         std::mem::swap(&mut self.realm, realm);
     }
 
+    /// Get and increment the parser identifier.
```
Razican, line 664: I would change the comment to "increment and get the parser identifier", since we first increment it, and then return it. To make sure that the users of the function are not surprised ;)

raskad: This should be fixed now. @Razican if you want to give it another review :)

Razican (approved):
LGTM!

PR #2928, opened 2023-05-11
Throw Syntax Errors for Keywords With Unicode Escapes

### LINKED ISSUE #2771: Keywords with Unicode escapes should throw syntax errors
**Describe the bug**
Currently, if we use a keyword with an Unicode escape, it will just work, and interpret it as the keyword itself. This is a follow-up from #1117.

**To Reproduce**
The following JavaScript code reproduces the issue:

```javascript
let x = f\u{61}lse;
```

**Expected behavior**
This should throw a syntax error. Instead, it just assigns the `false` boolean value to the `x` variable.

veera-sivarajan:
This Pull Request fixes/closes #2771 .

This PR throws a syntax error for keywords with unicode escapes. Also, it removes all traces of `bool` (used to indicate presence of unicode escapes) from `TokenKind::Keyword`.

Conformance regresses to 77.56% from 78% at e74cf369cd15cb4b3f6e14ea6d6e1a67177af94d but I don't know why.

### DIFF
--- a/boa_parser/src/lexer/identifier.rs
+++ b/boa_parser/src/lexer/identifier.rs
@@ -94,9 +94,15 @@ impl<R> Tokenizer<R> for Identifier {
             Self::take_identifier_name(cursor, start_pos, self.init)?;
 
         let token_kind = match identifier_name.parse() {
-            Ok(Keyword::True) => TokenKind::BooleanLiteral((true, ContainsEscapeSequence(contains_escaped_chars))),
-            Ok(Keyword::False) => TokenKind::BooleanLiteral((false, ContainsEscapeSequence(contains_escaped_chars))),
-            Ok(Keyword::Null) => TokenKind::NullLiteral,
+            Ok(Keyword::True) => {
+                TokenKind::BooleanLiteral((true, ContainsEscapeSequence(contains_escaped_chars)))
+            }
+            Ok(Keyword::False) => {
+                TokenKind::BooleanLiteral((false, ContainsEscapeSequence(contains_escaped_chars)))
+            }
+            Ok(Keyword::Null) => {
+                TokenKind::NullLiteral(ContainsEscapeSequence(contains_escaped_chars))
+            }
             Ok(keyword) => TokenKind::Keyword((keyword, contains_escaped_chars)),
             _ => TokenKind::IdentifierName((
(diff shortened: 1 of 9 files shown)


Razican:

| Total | 94,657 | 94,657 | 0 |
| Passed | 73,901 | 73,426 | **-475** |
| Ignored | 17,505 | 17,505 | 0 |
| Failed | 3,251 | 3,726 | **+475** |
| Panics | 0 | 0 | 0 |
| Conformance | 78.07% | 77.57% | **-0.50%** |

It seems that these changes break around 475 tests, could you check what's hapening?

boa_parser/src/lexer/identifier.rs
```diff
@@ -93,11 +93,20 @@ impl<R> Tokenizer<R> for Identifier {
...
+            return Err(Error::Syntax(
+                "Keyword must not contain escaped characters.".into(),
+                start_pos,
+            ));
+        }
+
+        let token_kind = match parsed_token {
             Ok(Keyword::True) => TokenKind::BooleanLiteral(true),
             Ok(Keyword::False) => TokenKind::BooleanLiteral(false),
             Ok(Keyword::Null) => TokenKind::NullLiteral,
```
HalidOdat, line 108: We shouldn't disallow escaped keywords in all circumstances only for expression parsing of keywords.

which happens here here, so we disallow it here.

veera-sivarajan: Oh, I thought all keywords cannot contain any unicode escape based on NOTE 1 from the [spec](https://tc39.es/ecma262/#sec-keywords-and-reserved-words). Is my interpretation incorrect?

HalidOdat: > Oh, I thought all keywords cannot contain any unicode escape based on NOTE 1 from the [spec](https://tc39.es/ecma262/#sec-keywords-and-reserved-words). Is my interpretation incorrect?

I meant our [`Keyword`](https://github.com/boa-dev/boa/blob/main/boa_ast/src/keyword/mod.rs) enum type (which doesn't conform to the spec), contains `Let`, `private`, etc. which can be identifiers and can have escape sequences.

So disallowing all of them breaks valid code. That's I suggested only disallowing some. the ones that are part of the `ReservedWords` (expect `async` and `await` as the spec says) can be identifiers.

PR #2929, opened 2023-05-13
Implement Runtime Private Environments

raskad:

- Implement runtime private environments and private names conforming with the spec.
- Remove parse-time distinction and mutation of private names.
- Adjust early error handling for private names to conform with the spec.

### DIFF
--- a/boa_engine/src/vm/code_block.rs
+++ b/boa_engine/src/vm/code_block.rs
@@ -391,6 +391,11 @@ impl CodeBlock {
                     interner.resolve_expect(self.private_names[operand as usize].description()),
                 )
             }
+            Opcode::PushPrivateEnvironment => {
+                let count = self.read::<u32>(*pc);
+                *pc += size_of::<u32>() * (count as usize + 1);
+                String::new()
+            }
             Opcode::Pop
             | Opcode::PopIfThrown
             | Opcode::Dup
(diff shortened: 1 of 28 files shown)


boa_engine/src/vm/code_block.rs
```diff
@@ -384,6 +384,11 @@ impl CodeBlock {
                     interner.resolve_expect(self.private_names[operand as usize].description()),
                 )
             }
+            Opcode::PushPrivateEnvironment => {
+                let count = self.read::<u32>(*pc);
+                *pc += size_of::<u32>() * (count as usize + 1);
+                String::new()
```
HalidOdat, line 390: It would be nice to print the operands

HalidOdat (commented):
Overall it looks good! Just had a small concern about the `id` field. The id (pointer address) is valid only for how long the Object is (because it can be reassigned to another object).

Can it be used after the object is deallocated?

raskad:
> Overall it looks good! Just had a small concern about the `id` field. The id (pointer address) is valid only for how long the Object is (because it can be reassigned to another object).
> 
> Can it be used after the object is deallocated?

I thought a bit about this. I think this should work. The private environments are only stored on function objects that are methods of classes. Those function objects always have a reference back to the class object with the `class_object` field (I think in theory the class object should be in the environment but that leads to the same outcome). So as long as the function objects with the private environments are alive, the class objects with the `id` pointer must also be alive.

HalidOdat (approved):
Looks perfect to me!

jedel1043 (approved):
Great work!

PR #2931, opened 2023-05-16
Deny Unicode Escapes in boolean and null expressions

### LINKED ISSUE #2771: Keywords with Unicode escapes should throw syntax errors

veera-sivarajan:
This Pull Request fixes/closes #2771.

This PR keeps track of `ContainsEscapeSequence` in all `TokenKind::BooleanLiteral` and `TokenKind::NullLiteral` and throws an error when it gets parsed as an expression.

Opening a different PR for this since I messed up the commit history of the previous PR.

### DIFF
--- a/boa_parser/src/lexer/token.rs
+++ b/boa_parser/src/lexer/token.rs
@@ -95,7 +95,7 @@ impl From<BigInt> for Numeric {
 #[derive(Clone, PartialEq, Debug)]
 pub enum TokenKind {
     /// A boolean literal, which is either `true` or `false`.
-    BooleanLiteral(bool),
+    BooleanLiteral((bool, ContainsEscapeSequence)),
 
     /// The end of the file.
     EOF,
@@ -118,7 +118,7 @@ pub enum TokenKind {
     /// The [`null` literal][spec].
     ///
     /// [spec]: https://tc39.es/ecma262/#prod-NullLiteral
-    NullLiteral,
+    NullLiteral(ContainsEscapeSequence),
 
     /// A numeric literal.
     NumericLiteral(Numeric),
@@ -152,7 +152,7 @@ pub enum TokenKind {
 impl From<bool> for TokenKind {
     #[inline]
     fn from(oth: bool) -> Self {
-        Self::BooleanLiteral(oth)
+        Self::BooleanLiteral((oth, ContainsEscapeSequence(false)))
     }
 }
 
@@ -182,7 +182,7 @@ impl TokenKind {
     #[inline]
     #[must_use]
     pub const fn boolean_literal(lit: bool) -> Self {
-        Self::BooleanLiteral(lit)
+        Self::BooleanLiteral((lit, ContainsEscapeSequence(false)))
     }
 
     /// Creates an `EOF` token kind.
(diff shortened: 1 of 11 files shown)


HalidOdat:
| Total | 94,657 | 94,657 | 0 |
| Passed | 73,901 | 73,901 | 0 |
| Ignored | 17,505 | 17,505 | 0 |
| Failed | 3,251 | 3,251 | 0 |
| Panics | 0 | 0 | 0 |
| Conformance | 78.07% | 78.07% | 0.00% |

raskad:
Since there seems to be not explicit test for this in the 262 suite, could you add some tests for this that make sure the parser produces syntax errors in these cases?

boa_parser/src/lexer/token.rs
```diff
@@ -95,7 +95,7 @@ impl From<BigInt> for Numeric {
 #[derive(Clone, PartialEq, Debug)]
 pub enum TokenKind {
     /// A boolean literal, which is either `true` or `false`.
-    BooleanLiteral(bool),
+    BooleanLiteral((bool, ContainsEscapeSequence)),
```
HalidOdat, line 98:
```suggestion
    BooleanLiteral(bool),
```

I think it's best to leave it as is, and throw the error when we create the `TokenKind::BooleanLiteral` (same for `TokenKind::NullLiteral`), since a boolean and null literal shouldn't have escape sequence.

HalidOdat (commented):
Looks good, check my comment to see on how we might improve this :)

boa_parser/src/lexer/token.rs
veera-sivarajan: Thanks for the suggestion.

But, if I throw an error when `Token::BooleanLiteral` is created at `boa_parser/src/lexer/identifier.rs`, it will reject the following code too:
```
let o = { f\u{61}lse: 10}; // should work
o.f\u{61}lse
```

Shouldn't we throw an error only when a `TokenKind::BooleanLiteral` or `TokenKind::NullLiteral` is parsed as a literal?

HalidOdat: ah, I see, You are right! Forgot about those cases :sweat_smile:

HalidOdat (approved):
Thanks for the contribution! @veera-sivarajan Looks perfect to me! :)

Razican:
> Since there seems to be not explicit test for this in the 262 suite, could you add some tests for this that make sure the parser produces syntax errors in these cases? 

Should we try to add such a test to Test262, or report it at least?

veera-sivarajan:
> > Since there seems to be not explicit test for this in the 262 suite, could you add some tests for this that make sure the parser produces syntax errors in these cases?
> 
> Should we try to add such a test to Test262, or report it at least?

I have opened a [PR](https://github.com/tc39/test262/pull/3831) for the same.

raskad (approved):
Really nice job, great contribution!

PR #2933, opened 2023-05-16
Fix `GetSubstitution`

HalidOdat:
Fixes `GetSubstitution` which is used in by `RegExp.prototype[@@replace]` and `String.prototype.replace()` calls `RegExp.prototype[@@replace]`.

Fixes the remaining ES5 tests on `buil-ins/String` test suite.

Because of the [bug](https://github.com/tc39/ecma262/issues/1426) in the [spec about `GetSubstitution`](https://tc39.es/ecma262/#sec-getsubstitution). I had to deviate from the spec, added a comment clarifying the deviation.

### DIFF
--- a/boa_engine/src/builtins/regexp/mod.rs
+++ b/boa_engine/src/builtins/regexp/mod.rs
@@ -1267,8 +1267,14 @@ impl RegExp {
         args: &[JsValue],
         context: &mut Context<'_>,
     ) -> JsResult<JsValue> {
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Nice work! Really like the additional notes to clarify spec deviations :)

nekevss (approved):
Looks good! Seconding that the notes updates were nice.

PR #2936, opened 2023-05-17
Allow `async` as identifier name in certain contexts

jedel1043:
Just a small change in the parser to fix a couple of tests related to `async` as an identifier name.


HalidOdat (approved):
Looks good to me!

raskad (approved):

PR #2940, opened 2023-05-18
Type safe root shape

HalidOdat:
Currently when we construct a `Realm` we take a `SharedShape` as the root, since users can create realms to ensure that the passed shape is the root shape (otherwise if a non-root shape is passed, will at best cause a panic and at worse will give unexpected behaviour), added the `RootShape` that ensures that only a root shape is passed, this is applied to many other places where a root shape is required.

It changes the following:

- Add `RootShape` wrapper around `SharedShape`
- Implement `From<UniqueShape>` for `Shape`
- Implement `From<SharedShape>` for `Shape`

### DIFF
--- a/boa_engine/benches/full.rs
+++ b/boa_engine/benches/full.rs
@@ -1,7 +1,7 @@
 //! Benchmarks of the whole execution engine in Boa.
 
 use boa_engine::{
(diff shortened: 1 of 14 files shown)


raskad (approved):
Looks very reasonable to me.

jedel1043 (approved):
Nice improvement!

PR #2951, opened 2023-05-20
Direct conversion from `u8` to `Opcode`

HalidOdat:
Optimizes opcode fetching by removing the checks that validate if it's a correct opcode. every `u8` combination matches to a valid `Opcode` variant (even if it's reserved).

**Main:**

```
Richards: 20.5
DeltaBlue: 24.8
RayTrace: 81.0
----
Score (version 7): 34.5
undefined
```

**PR:**

```
Richards: 20.6
DeltaBlue: 24.8
RayTrace: 82.1
----
Score (version 7): 34.8
undefined
```

There is a small improvement in performance.

### DIFF
--- a/boa_engine/src/vm/code_block.rs
+++ b/boa_engine/src/vm/code_block.rs
@@ -259,7 +259,7 @@ impl CodeBlock {
     /// Returns an empty `String` if no operands are present.
     #[cfg(any(feature = "trace", feature = "flowgraph"))]
     pub(crate) fn instruction_operands(&self, pc: &mut usize, interner: &Interner) -> String {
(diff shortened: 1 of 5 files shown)


jasonwilliams (approved):
LGTM

raskad:
I think we don't need `transmute` here to get the performance improvement. Even on opt-level 1 the produced code is the same when all `u8`s are matched to their enum variants: https://rust.godbolt.org/z/oxMeMMrjG.

We should also remove the `Opcode::from_raw` function since that can be done with `into()` safeley now.

jasonwilliams (approved):

raskad (approved):
Looks good

PR #2960, opened 2023-05-24
show object kind, name and address when using dbg!

jasonwilliams:
This should implement Debug a bit better for objects, previously they recursed a lot and spewed out a lot of garbage. This shows the kind, name and address for each object, which should be enough when investigating issues.

Before:
<hot mess>

After:
`&func = (Function) "fib" 0x7F9A67F17218`

### DIFF
--- a/boa_engine/src/object/jsobject.rs
+++ b/boa_engine/src/object/jsobject.rs
@@ -1114,10 +1114,9 @@ impl Debug for JsObject {
                     None => JsString::default(),
                     Some(prop) => prop
                         .value()
-                        .expect("no name property")
-                        .as_string()
-                        .expect("cannot convert to string")
-                        .clone(),
+                        .and_then(JsValue::as_string)
+                        .cloned()
+                        .unwrap_or_default()
                 };
 
                 return f.write_fmt(format_args!("({:?}) {:?} 0x{:X}", kind, name, ptr as usize));
--- a/boa_engine/src/object/jsobject.rs
+++ b/boa_engine/src/object/jsobject.rs
@@ -14,7 +14,7 @@ use crate::{
     property::{PropertyDescriptor, PropertyKey},
     string::utf16,
     value::PreferredType,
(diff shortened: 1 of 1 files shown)


boa_engine/src/object/jsobject.rs
```diff
@@ -1102,10 +1101,29 @@ impl Debug for JsObject {
...
-                .finish()
+            let ptr: *const _ = self.as_ref();
+            let obj = self.borrow();
+            let kind = obj.kind();
+            let name_prop = self
+                .borrow()
+                .properties()
+                .get(&PropertyKey::String(JsString::from("name")));
+
+            if obj.is_function() {
```
HalidOdat, line 1112:
```suggestion
            if obj.is_function() {            
                let name_prop = obj
                    .properties()
                    .get(&PropertyKey::String(JsString::from("name")));
```

Since we only use it for functions only, we can move it, also we already borrowed it so we can use the `obj`

HalidOdat (commented):
Looks good! Check the comment to see how we might improve it :)

HalidOdat (approved):
Looks good to me! :)

Just need to run `rust-fmt`

jedel1043 (approved):
Thanks! This should significantly improve the output of `dbg!`.

jasonwilliams:
I can do this in a follow up or but I left the recursion stuff in there, I gues that can be removed as it’s not possible for recursion when printing a object now (unless we want to improve the output later on with properties etc)

@HalidOdat @raskad

jedel1043:
> I can do this in a follow up or but I left the recursion stuff in there, I gues that can be removed as it’s not possible for recursion when printing a object now (unless we want to improve the output later on with properties etc)

While debugging I tend to use `object.display().internals(true)` more. This change should mostly benefit end users since it won't spit pages and pages of properties while asserting that two values are equal, for example. I think we should remove the recursion.

PR #2962, opened 2023-05-25
Remove redundant `param_count` field from `CallFrame`

HalidOdat:
Hopefully this is will be a PR in a series of PRs to refactor the way we call functions and in general and making it easier to implement some optimizations, breaking it down into smaller parts for easier review :)

It changes the following:

- Remove redundant `param_count` field from `CallFrame`
- Make `pc`, `fp`,, etc a `u32`.


nekevss (approved):
Looks good to me!

raskad (approved):

PR #2967, opened 2023-05-27
Remove unneded `num_bindings` in `Opcode`s and `CodeBlock`

HalidOdat:
Since the number of bindings is stored in the compiled envionments and when we create the environment we pass it, there is no need to store the num_bindings for the environments

It changes the following:

- Remove unneeded `function_environment_push_location` field on `CodeBlock`
- Remove unneeded `num_bindings` field on `CodeBlock`
- Remove unneeded `num_bindings` operand from `PushDeclarativeEnvironment` opcode
- Remove unneeded `num_bindings` operand from `PushFunctionEnvironment` opcode
- Make `ByteCompiler::push_compile_environment()` return only the index.
- Remove `num_bindings` field from environment stack methods that push environments.
- Move `CodeBlock` flags to `bitflags` field.

### DIFF
--- a/boa_engine/src/vm/code_block.rs
+++ b/boa_engine/src/vm/code_block.rs
@@ -52,21 +52,12 @@ unsafe impl Readable for i64 {}
 unsafe impl Readable for f32 {}
 unsafe impl Readable for f64 {}
 
-// bitflags! {
-//     #[derive(Clone, Copy, Debug, Finalize)]
-//     struct Flags: u32 {
-//         const A = 0b00000001;
-//         const B = 0b00000010;
-//         const C = 0b00000100;
-//         const ABC = Self::A.bits() | Self::B.bits() | Self::C.bits();
-//     }
-// }
-
 /// The internal representation of a JavaScript function.
 ///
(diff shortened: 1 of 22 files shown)


boa_engine/src/vm/code_block.rs
```diff
@@ -167,11 +167,36 @@ impl CodeBlock {
...
+    }
+    pub(crate) fn has_binding_identifier(&self) -> bool {
+        self.flags
+            .get()
+            .contains(CodeBlockFlags::HAS_BINDING_IDENTIFIER)
+    }
+    pub(crate) fn has_parameters_env_bindings(&self) -> bool {
+        self.flags
+            .get()
+            .contains(CodeBlockFlags::PARAMETERS_ENV_BINDINGS)
```
raskad, line 199: These functions could have done docs and spaces between them.

raskad (approved):
Looks great.

nekevss (approved):
I like the changes a lot! I'd second the above review about docs. Other than that, great work! 😄

PR #2969, opened 2023-05-28
Remove `arguments_binding` field from `CodeBlock`

HalidOdat:
This PR reorders a spec step to ensure that the `"arguments"` binding is always in the `0` position. This allows us to remove the `arguments_binding` field completely since we know in which environment it is (last pushed) and what index it has (`0`).

Removes `24` bytes from `CodeBlock` (from `184` bytes to `160` bytes)

### DIFF
--- a/boa_engine/src/builtins/function/arguments.rs
+++ b/boa_engine/src/builtins/function/arguments.rs
@@ -183,7 +183,7 @@ impl Arguments {
                 break;
             }
 
-            // NOTE(HaledOdat): Offset by +1 to account for the fitst binding ("argument").
+            // NOTE(HalidOdat): Offset by +1 to account for the first binding ("argument").
             let binding_index = bindings.len() as u32 + 1;
 
             let entry = bindings
--- a/boa_engine/src/builtins/function/arguments.rs
+++ b/boa_engine/src/builtins/function/arguments.rs
@@ -155,18 +155,22 @@ impl Arguments {
         // We use indices to access environment bindings at runtime.
         // To map to function parameters to binding indices, we use the fact, that bindings in a
         // function environment start with all of the arguments in order:
+        //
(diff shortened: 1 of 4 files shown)


HalidOdat:
**Main:**

```
Richards: 20.5
DeltaBlue: 24.7
RayTrace: 82.0
----
Score (version 7): 34.6
```

**PR:**

```
Richards: 20.7
DeltaBlue: 25.0
RayTrace: 82.8
----
Score (version 7): 35.0
undefined
```

There is an improvement in performance :)

raskad (approved):
Nice work!

jedel1043 (approved):
Looks good!

Razican:
> **Main:**
> 
> ```
> Richards: 20.5
> DeltaBlue: 24.7
> RayTrace: 82.0
> ----
> Score (version 7): 34.6
> ```
> 
> **PR:**
> 
> ```
> Richards: 20.7
> DeltaBlue: 25.0
> RayTrace: 82.8
> ----
> Score (version 7): 35.0
> undefined
> ```
> 
> There is an improvement in performance :)

What benchmarking technique are you using here? Could we document it so that we can at some point run it in the CI or manually?

HalidOdat:
> What benchmarking technique are you using here?

I used the benchmark in this comment https://github.com/boa-dev/boa/issues/1924#issuecomment-1166989423 it is a subset of the quickjs benchmarks (for the full benchmarks see https://github.com/boa-dev/boa/issues/1924#issuecomment-1437314041), I didn't use the full one because one has a very long execution time `~15min`

To run the benchmarks with the boa cli with the file. (`cargo run --bin boa --release -- bench.js`).

> Could we document it so that we can at some point run it in the CI

I don't know if the results would be that useful, at least while running with the github job runners...

PR #2976, opened 2023-05-30
Unify async iterators and iterators compilation

jedel1043:
This Pull Request fixes a bunch of iterator tests that were failing because we handled async iterators and iterators differently with respect to closes and returns.

- Unifies compilation logic for iterators and async iterators.
- Refactors `IteratorRecord` to be able to use it as iterator state at vm execution.
- Avoids using the values stack for iterator tracking, instead using an `IteratorRecord` stack.

### DIFF
--- a/boa_engine/src/bytecompiler/jump_control.rs
+++ b/boa_engine/src/bytecompiler/jump_control.rs
@@ -27,7 +27,7 @@ pub(crate) struct JumpControlInfo {
 bitflags! {
     /// A bitflag that contains the type flags and relevant booleans for `JumpControlInfo`.
     #[derive(Debug, Clone, Copy)]
-    pub(crate) struct JumpControlInfoFlags: u8 {
+    pub(crate) struct JumpControlInfoFlags: u16 {
         const LOOP = 0b0000_0001;
         const SWITCH = 0b0000_0010;
         const TRY_BLOCK = 0b0000_0100;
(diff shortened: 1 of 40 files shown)


boa_engine/src/vm/opcode/mod.rs
```diff
@@ -1421,82 +1421,120 @@ generate_impl! {
...
         ///
-        /// Stack: object **=>** iterator, next_method
+        /// Stack: object **=>**
+        ///
+        /// Iterator Stack: `iterator`
         CreateForInIterator,
 
+        /// Push iterator loop start marker.
+        ///
+        /// Operands: Exit Address: `u32`
```
raskad, line 1431: This opcode takes start and exit right?

boa_engine/src/bytecompiler/jump_control.rs
```diff
@@ -139,8 +146,12 @@ impl JumpControlInfo {
...
     }
 
-    pub(crate) const fn for_of_in_loop(&self) -> bool {
-        self.flags.contains(JumpControlInfoFlags::FOR_OF_IN_LOOP)
+    pub(crate) const fn iterator_loop(&self) -> bool {
+        self.flags.contains(JumpControlInfoFlags::ITERATOR_LOOP)
+    }
+
+    pub(crate) const fn for_await_of_loop(&self) -> bool {
+        self.flags.contains(JumpControlInfoFlags::ITERATOR_LOOP)
```
raskad, line 154: This should be `FOR_AWAIT_OF_LOOP`.

raskad (commented):
Great changes. Just have to see if my comment in jump_control breaks anything.

jedel1043:
Hmm, the last commit didn't fix any tests. I'll try to come up with a test for this.

jedel1043:
Added some tests that cover the fixed bug :)

raskad (approved):
Great stuff!

PR #2983, opened 2023-05-31
Update fuzzers and add building them to CI

addisoncrump:
This PR updates the fuzzers to the latest version of the code and adds them to the CI. It does _not_ perform the fuzz run.

Both `parser-idempotency` and `vm-implied` find crashes pretty quickly. Maybe would be worthwhile to have another round of testing with this. :)


Razican (commented):
Thanks! Indeed, we could run it for a bit and find some panics

Razican (approved):
Looks good to me!

raskad (approved):

PR #2984, opened 2023-05-31
Fix Grammar and Links in Docs

veera-sivarajan:
This PR makes a bunch of changes to the documentation:

1. Improve sentence flow and fix punctuation in `docs/debugging.md`.

2. More than 80% of users use dark mode but Boa's architecture image in `docs/vm.md` looked like this:

boa-image

This PR adds a white background to the image so that the image is clear for all users.

3. Previously, the link to SpiderMonkey's JavaScript Shell in `docs/vm.md` pointed to an invalid page. This PR fixes it by pointing it to a backup page.


Razican (approved):
Thanks!!

raskad:
The diagram png does not contain a copy of the diagram anymore right? I think we should keep that so we can easily adjust it in the future.

HalidOdat (approved):
Looks good to me! :)

PR #2986, opened 2023-06-01
Fix prompt on windows

ShaneEverittM:
It changes the following:

- Moves prompt highlighting to the helper class
- Fixes prompt width calculation on windows

### Details
In rustyline, there are two implementations of `calculate_position`, one for [windows](https://github.com/kkawakam/rustyline/blob/master/src/tty/windows.rs#L469) and one for [unix](https://github.com/kkawakam/rustyline/blob/master/src/tty/unix.rs#L1020). The one for unix support ignoring ANSI escape sequences when calculating width, but the one on windows does not, [by design](https://github.com/kkawakam/rustyline/pull/215#issuecomment-491945339).

This caused an issue where the cursor would be placed 12 columns in, instead of 3.

#### Before
before
#### After
after

### DIFF
--- a/boa_cli/src/helper.rs
+++ b/boa_cli/src/helper.rs
@@ -7,7 +7,7 @@ use rustyline::{
     validate::{MatchingBracketValidator, ValidationContext, ValidationResult, Validator},
     Completer, Helper, Hinter,
 };
-use std::borrow::Cow;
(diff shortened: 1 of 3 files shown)


HalidOdat (approved):
Nice work :)

Razican (approved):

PR #2993, opened 2023-06-03
Correctly initialize functions inside modules

jedel1043:
Small fix that ensures bindings inside functions are correctly resolved to outer functions.


raskad (approved):

HalidOdat (approved):
Nice work!

PR #2994, opened 2023-06-03
Allow `true`, `false` and `null` in object patterns

jedel1043:
This Pull Request allows using certain valid keywords inside patterns.

### DIFF
--- a/boa_engine/src/tests/mod.rs
+++ b/boa_engine/src/tests/mod.rs
@@ -481,3 +481,21 @@ fn template_literal() {
         "result: 10 and 20",
     )]);
 }
+
+#[test]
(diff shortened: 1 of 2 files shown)


Razican:
It seems that there is no test for this in Test262, so we should probably add our own, and maybe add a test to Test262.

HalidOdat (approved):
Looks good to me! +1 for the tests to test262, if not then our own :)

jedel1043:
> It seems that there is no test for this in Test262, so we should probably add our own, and maybe add a test to Test262.

Will add both a test here and a PR on Test262, just to avoid blocking this on that PR.

Razican (approved):
Thanks!

PR #2996, opened 2023-06-04
Return the correct value during a labelled break

raskad:

- Return the correct value during a labelled break


jedel1043 (approved):
Looks good!

HalidOdat (approved):
Looks good to me!

PR #3001, opened 2023-06-06
Fix panic in constructor call

raskad:

- Fix a bug in the constructor execution that leads to a panic when an arguments object is needed. In contrast to the call execution the wrong environment is being pushed.


Razican (approved):
Good catch!

jedel1043 (approved):
Good fix :)

PR #3007, opened 2023-06-07
Add convenience methods to `ModuleLoader`

jedel1043:
While implementing an example of async module fetching, I realized it would be useful to have a way to insert a module through a `&dyn ModuleLoader`. This PR provides that API.

~~Drafting because I'm still experimenting with the final design of the API.~~

### DIFF
--- a/boa_engine/src/module/mod.rs
+++ b/boa_engine/src/module/mod.rs
@@ -108,6 +108,26 @@ pub trait ModuleLoader {
         context: &mut Context<'_>,
     );
 
+    /// Registers a new module into the module loader.
+    ///
+    /// This is a convenience method for module loaders caching already parsed modules, since it
+    /// allows registering a new module through the `&dyn ModuleLoader` provided by
+    /// [`Context::module_loader`].
+    ///
+    /// Does nothing by default.
+    fn register_module(&self, _specifier: JsString, _module: Module) {}
+
+    /// Gets the module associated with the provided specifier.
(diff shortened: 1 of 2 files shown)


jedel1043:
Yeah, I think this should be enough for now. We can revisit the API if we have reports about its limitations in the future.

boa_engine/src/module/mod.rs
```diff
@@ -233,6 +253,22 @@ impl ModuleLoader for SimpleModuleLoader {
...
+        )
+    }
+
+    // TODO: Try to unify `ModuleLoader::get_module` with `SimpleModuleLoader::get`.
+    fn get_module(&self, _specifier: JsString) -> Option<Module> {
+        panic!(
+            "`SimpleModuleLoader` uses paths to cache the modules instead of specifiers.
+        To get a module, you need to use the `SimpleModuleLoader::get` method."
+        )
+    }
```
raskad, line 271: Should these really be panics or should it just be the default behaviour of doing nothing? Either way I think it probably should be documented on the `SimpleModuleLoader` struct so that users are informed right?

jedel1043: Yeah, if we're going to add the functionality in the future, maybe this should just do nothing for now. Will add docs explaining this though.

raskad (approved):

HalidOdat (approved):
Looks good to me! :)

PR #3011, opened 2023-06-08
Allow awaiting `JsPromise` from Rust code

jedel1043:
Should be mostly useful for consumers of the `JobQueue::enqueue_future_job` API, since they can now await for certain promises inside `async` blocks and functions.

Changes:
- Adds a new `JsFuture` type that implements `Future`.
- Adds a new `JsPromise::into_js_future` method.
- Rewrites the caveats section of `NativeFunction::from_async_fn`, since the old explanation was wrong.
- Adds a new `JsPromise::from_future` constructor.
- Improves some promise code.

### DIFF
--- a/boa_engine/src/object/builtins/jspromise.rs
+++ b/boa_engine/src/object/builtins/jspromise.rs
@@ -1,5 +1,7 @@
 //! A Rust API wrapper for Boa's promise Builtin ECMAScript Object
 
+use std::{future::Future, pin::Pin, task};
+
 use super::{JsArray, JsFunction};
 use crate::{
     builtins::{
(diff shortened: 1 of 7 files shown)


HalidOdat (approved):
Nice addition!

boa_engine/src/object/builtins/jspromise.rs
```diff
@@ -833,6 +891,114 @@ impl JsPromise {
...
+    ///
+    /// assert_eq!(result1, Ok(JsValue::from(10)));
+    /// assert_eq!(result2, Ok(()));
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn into_js_future(self, context: &mut Context<'_>) -> JsResult<JsFuture> {
+        // Mostly based from:
+        // https://rustwasm.github.io/wasm-bindgen/api/src/wasm_bindgen_futures/lib.rs.html#109-168
```
Razican, line 936: This link will probably change in the future, we should find a link that doesn't change (to a given version / Git commit)

boa_engine/src/object/builtins/jspromise.rs
```diff
@@ -870,3 +1036,43 @@ impl TryFromJs for JsPromise {
...
+
+#[derive(Trace, Finalize)]
+struct Inner {
+    result: Option<JsResult<JsValue>>,
+    #[unsafe_ignore_trace]
+    task: Option<task::Waker>,
+}
+
+// Taken from:
+// https://rustwasm.github.io/wasm-bindgen/api/src/wasm_bindgen_futures/lib.rs.html#171-187
```
Razican, line 1064: Here too, I would try to find a static link to the code.

Razican (approved):
Looking great! I added a couple of suggestions for two doc links. We might want to mention this in https://github.com/boa-dev/boa-dev.github.io/pull/94

PR #3017, opened 2023-06-08
Make `IntegerIndexed::byte_offset` public

CryZe:
This allows querying the full range of the `TypedArray` in the underlying `ArrayBuffer`.


jedel1043 (approved):
Thanks! I'm in favour of exposing this as a public accessor, since it's part of the properties that the spec defines for integer indexed objects.

HalidOdat (approved):
Looks good to me!

PR #3018, opened 2023-06-08
Fix panics on empty return values

raskad:
This adds and fixes some tests that caused panics on `main` and during the development of the fixes. All where found with the fuzzer. After running the fuzzer for some time I did not encounter any similar panics anymore.


HalidOdat (approved):
Nice catch!!

PR #3019, opened 2023-06-08
Add early error for `await` in class static blocks

raskad:

- Add early error for `await` in class static blocks


HalidOdat (approved):
Nice catch! :)

jedel1043 (approved):
Looking great!

PR #3021, opened 2023-06-09
Make `Context` a trait

jedel1043:
One of the most painful problems for our users right now is that any `JobQueue` and `ModuleLoader` can only be immutable by the nature of our current Context API. However, there's an alternative: to make `Context` a trait and make `JobQueue` and `ModuleLoader` a supertrait of `Context`. This gives us several advantages:
- Allows us to have mutable module loader data and job queue data, avoiding the need to wrap everything in a `RefCell`.
- Unblocks the possibility of adding `Provider` as a supertrait in the future, which is useful for users that want to append additional data onto a context and recover it inside JS calls.
- Gives freedom to the user about how to store the context itself, which could be useful if some user wants to e.g. wrap a `Context` in a `RefCell` to share it between async future jobs.
- By design, disallows using a job queue or a module loader with a different context than the one provided on initialization.

This comes with certain disadvantages however:
- Now every access to the `vm` needs to pass through a virtual method call. This could impact performance but the cache could also store the `vm` after some calls and completely remove the overhead. This would need to be tested with benchmarks.
- It makes our API more complex. I couldn't find a way to override only the `JobQueue` or only the `ModuleLoader` without having to copy the whole `DefaultContext` implementation. Though, this could be alleviated with a new `DefaultContext` API with interchangeable modules, but I'm still not sure about what the design would be.

I'm opening this PR to seek comments and suggestions to know if we should commit to this. I'll finish the implementation this weekend to be able to benchmark it against the current implementation and see if there's a performance regression.

### DIFF
--- a/boa_cli/src/debug/function.rs
+++ b/boa_cli/src/debug/function.rs
@@ -50,7 +50,11 @@ fn flowgraph_parse_direction_option(value: &JsValue) -> JsResult<Direction> {
 }
 
 /// Get functions instruction flowgraph
-fn flowgraph(_this: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+fn flowgraph(
+    _this: &JsValue,
+    args: &[JsValue],
+    context: &mut dyn Context<'_>,
+) -> JsResult<JsValue> {
     let Some(value) = args.get(0) else {
         return Err(JsNativeError::typ()
         .with_message("expected function argument")
@@ -100,7 +104,7 @@ fn flowgraph(_this: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> Js
     Ok(JsValue::new(result))
 }
 
-fn bytecode(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+fn bytecode(_: &JsValue, args: &[JsValue], context: &mut dyn Context<'_>) -> JsResult<JsValue> {
     let Some(value) = args.get(0) else {
         return Err(JsNativeError::typ()
         .with_message("expected function argument")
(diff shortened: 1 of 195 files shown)


jedel1043:
# Main
RESULT Richards 27.4
RESULT DeltaBlue 32.8
RESULT Crypto 39.6
RESULT RayTrace 108
RESULT EarleyBoyer 92.8
RESULT Splay 111
RESULT NavierStokes 15.0
SCORE 48.1

# dyn Context
RESULT Richards 26.6
RESULT DeltaBlue 31.7
RESULT Crypto 35.2
RESULT RayTrace 105
RESULT EarleyBoyer 88.9
RESULT Splay 107
RESULT NavierStokes 14.6
SCORE 45.9

So there is a small perf regression on regular scripts. Though, I would expect to see a perf increase in promises-heavy scripts and modules, but I don't think QuickJS' measures that.
I thought the performance impact would be bigger for such a significant API change, so I think this can be considered a success?

HalidOdat:
I know I'm asking something obvious, but why not have an owning `Box<dyn ModuleLoader>`? which should be accessable as `&mut` from `&mut Context<'_>`? It would be simular to `DefaultContext`.

If we need to be able to share it, then the user can pass a `Rc<ConcreteModuleLoader>`.

jedel1043:
> I know I'm asking something obvious, but why not have an owning `Box<dyn ModuleLoader>`? which should be accessable as `&mut` from `&mut Context<'_>`? It would be simular to `DefaultContext`.

Thought about it, but I don't think that's possible. The definition of `ModuleLoader` takes a reference
to the context itself:
```rust
fn load_imported_module(
    &self,
    referrer: Referrer,
    specifier: JsString,
    finish_load: Box<dyn FnOnce(JsResult<Module>, &mut Context<'_>)>,
    context: &mut Context<'_>,
);
```
This means you cannot call `load_imported_module from the `&mut dyn ModuleLoader` and at the same time pass a reference to the full context because of the double mutable borrow.

jedel1043:
Oof, benchmarks are pretty bad...
Any ideas for other designs? Maybe there are some other alternatives I'm not seeing right now.

jedel1043:
Ok, apparently there was possibilities of getting support for associated fields in traits:
https://internals.rust-lang.org/t/pre-rfc-associated-field-constant-accessors-in-traits/17756

This would completely remove the virtual calls, making the performance equal with main. I'll leave this PR open and we can revisit it in the future.

jedel1043:
Closing this since the Rust project seems to not have plans for the linked feature. We can revisit this later if the feature is added, or we can keep the current API.

PR #3022, opened 2023-06-10
Remove `[[ClassFieldInitializerName]]` from `CodeBlock`

HalidOdat:
In the entire spec the `[[ClassFieldInitializerName]]` field is only used to check that it's not empty (in `PerformEval`), so removed it and added a bitflag for it.

It changes the following:

- Remove `[[ClassFieldInitializerName]]` from `CodeBlock`
- Add `IN_CLASS_FIELD_INITIALIZER` bitflag


raskad (approved):
Nice change :)

jedel1043 (approved):
Nice optimization!

PR #3023, opened 2023-06-10
Merge `private_names` into `names` constant pool

HalidOdat:
It changes the following:

- Merges `private_names` into `names` constant pool.
- Make private names use `JsString` instead of `Sym`


raskad (approved):
Looks good!

jedel1043 (approved):
All good!

PR #3025, opened 2023-06-11
Use `Rc` instead of `Gc` for `CompileTimeEnvironment`s

HalidOdat:
This PR makes `CompileTimeEnvironment`s use `Rc` instead of `Gc` since nothing in it requires to be traced and it does not have a a cyclic reference (only a pointer to the previous).

### Main

```
PROGRESS Richards
RESULT Richards 30.8
PROGRESS DeltaBlue
RESULT DeltaBlue 35.8
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 43.2
PROGRESS RayTrace
RESULT RayTrace 121
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 102
undefined
PROGRESS RegExp
PROGRESS Splay
RESULT Splay 124
PROGRESS NavierStokes
RESULT NavierStokes 11.7
SCORE 50.6

```

### PR

```
PROGRESS Richards
RESULT Richards 31.2
PROGRESS DeltaBlue
RESULT DeltaBlue 37.3
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 44.1
PROGRESS RayTrace
RESULT RayTrace 123
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 105
undefined
PROGRESS RegExp
PROGRESS Splay
RESULT Splay 127
PROGRESS NavierStokes
RESULT NavierStokes 16.5
SCORE 54.3
```


jedel1043:
I'm in favor of using `Rc` to optimize right now, but if we had an optimal GC the current design should be faster than using reference counting, right?

jedel1043:
Could be good to have a tracking issue of our usages of `Rc` to know if after optimizing our GC we can use `Gc` instead and see if performance increases.

HalidOdat:
> I'm in favor of using `Rc` to optimize right now, but if we had an optimal GC the current design should be faster than using reference counting, right?

Maybe, if we remove reference counting from our GC and root/unrooting traversal

> Could be good to have a tracking issue of our usages of Rc to know if after optimizing our GC we can use Gc instead and see if performance increases.

Created the issue https://github.com/boa-dev/boa/issues/3034 :)

boa_engine/src/vm/code_block.rs
```diff
@@ -129,7 +134,10 @@ pub struct CodeBlock {
     pub(crate) functions: Box<[Gc<Self>]>,
 
     /// Compile time environments in this function.
-    pub(crate) compile_environments: Box<[Gc<GcRefCell<CompileTimeEnvironment>>]>,
+    ///
+    // Safety: Nothing in CompileTimeEnvironment needs tracing, so this is safe.
```
Razican, line 138: Maybe we should add a comment in `CompileTimeEnvironment` so that we don't forget and add a Gc there in the future.

boa_engine/src/environments/compile.rs
```diff
@@ -18,15 +20,19 @@ struct CompileTimeBinding {
 /// A compile time environment maps bound identifiers to their binding positions.
 ///
 /// A compile time environment also indicates, if it is a function environment.
-#[derive(Debug, Finalize, Trace)]
+#[derive(Debug, Finalize)]
 pub(crate) struct CompileTimeEnvironment {
-    outer: Option<Gc<GcRefCell<Self>>>,
+    outer: Option<Rc<RefCell<Self>>>,
     environment_index: u32,
-    #[unsafe_ignore_trace]
```
Razican, line 25: Was this wrong before?

Razican (approved):
I added some suggestions / comments, but feel free to merge

boa_engine/src/environments/compile.rs
HalidOdat: No, I just removed the `Trace` derive, since it didn't need Trace.

raskad (approved):

boa_engine/src/vm/code_block.rs
raskad: I think this should already be tracked in #3034, right?

PR #3027, opened 2023-06-11
Optimize `Swap` opcode

raskad:
Instead of using `Vec::pop` and `Vec::push` two times, we can use `slice::swap`. The explicit assert allows the compiler to reduce the panic logic further.


Razican (approved):
Good optimization!

jedel1043 (approved):

PR #3028, opened 2023-06-11
Fix class constructor return value

raskad:

- Adjust compilation of class constructors to always leave the stack with one return value.


HalidOdat (approved):
Nice catch! :)

Razican (approved):

PR #3037, opened 2023-06-13
Implement control flow graph construction

HalidOdat:
This PR is a WIP implementation of control flow graph construction from bytecode, that will be used for many optimizations, like dead-code-elimination.


jedel1043:
Nice! This'll also be really useful when we eventually implement JIT compilation

boa_cli/src/debug/optimizer.rs
```diff
@@ -43,6 +43,39 @@ fn set_statistics(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> J
     Ok(JsValue::undefined())
 }
 
+fn cfg(_: &JsValue, args: &[JsValue], _context: &mut Context<'_>) -> JsResult<JsValue> {
```
Razican, line 46: Not sure if I would call this "cfg", since that can be confused with "config", and it's also the macro name used in Rust.

Maybe "cflowg", or something that shows that this is about "control flow graphs".

boa_engine/src/optimizer/cfg.rs
```diff
@@ -0,0 +1,640 @@
...
+
+            writeln!(f, "{basic_block:#?}")?;
+        }
+
+        Ok(())
+    }
+}
+
+fn is_jump_kind_opcode(opcode: Opcode) -> bool {
+    match opcode {
```
Razican, line 513: This can probably be exchanged by a `matches!()` call, which could make it `const`

Razican (commented):
Great step :) I added a couple of comments, but I didn't do a full review yet

jedel1043:
Coincidentally, I found this crate that could simplify the optimizer logic:
https://docs.rs/egg/0.9.5/egg

HalidOdat:
> Coincidentally, I found this crate that could simplify the optimizer logic:
> https://docs.rs/egg/0.9.5/egg

Looks interesting! But I think it would be better to create our own, because JS semantics are complex and it would be easier extensibility.

I postponed this PR until #3059 is fully complete (currently refactoring generators, which are the remaining test262 failing tests).

boa_engine/src/optimizer/control_flow_graph/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+}
+
+impl Drop for ControlFlowGraph {
+    fn drop(&mut self) {
+        // NOTE: Untie BasicBlock nodes, so they can be deallocated.
+        for basic_block in &self.basic_blocks {
+            *basic_block.borrow_mut() = BasicBlock::default();
+        }
+    }
+}
```
jedel1043, line 454: Could be good to see if we can use https://lib.rs/crates/slotmap, to avoid doing things like this.

HalidOdat: Will check it out! Thank you for the suggestion! :)

jedel1043:
@HalidOdat do you still plan to work on this? Since it would require a very big rebase.

HalidOdat:
I backed it up to my own fork, we can close this very out of date.

PR #3044, opened 2023-06-16
Implement asynchronous evaluation of scripts

jedel1043:
This Pull Request implements and API to asynchronously run scripts.
Had to do some refactoring in order to share code between `run` and `run_async_with_budget`, but I ran the QuickJS benchmark and everything looks good!

~However, our engine has a big limitation right now: we cannot track the executed instructions of called functions because those internally use `run`. For this reason, this PR just tracks instructions executed in the root script being executed, but this can be implemented in the future.~ Solved by #3295

<details>
  <summary><h2>Some context</h2></summary>

We have used this pattern of freely calling `run` inside several functions because it's easy to reason about, and `Context` owns the current VM state. However, it is really unusual to have VM state inside a shared `Context`, and using `run` this way causes problems that we've already seen but ignored, like duplicate prints of functions' bytecode when using tracing.

If we can pull off the separation of the VM from the Context, we could even make the context reference we pass to every function an immutable one (`&Context<'_>`), which will unblock a LOT of patterns that we've been largely working around.

Having said that, I have some ideas on how to make this work. This will probably involve changing the return value of our functions and some generator shenanigans, but I found an interesting crate that implements them using async/await, which makes them available on stable code (https://github.com/whatisaphone/genawaiter).

I'm open to any other ideas you have! The gist of problem is: "How to bubble up JS function calls to the original `Context::run` call".

Anyways, sorry for the info dump.
</details>

### DIFF
--- a/boa_engine/src/vm/mod.rs
+++ b/boa_engine/src/vm/mod.rs
@@ -400,9 +400,9 @@ impl Context<'_> {
         #[cfg(feature = "fuzz")]
         {
             if self.instructions_remaining == 0 {
-                return CompletionRecord::Throw(JsError::from_native(
+                return ControlFlow::Break(CompletionRecord::Throw(JsError::from_native(
                     JsNativeError::no_instructions_remain(),
-                ));
+                )));
             }
             self.instructions_remaining -= 1;
         }
--- a/boa_engine/src/vm/mod.rs
+++ b/boa_engine/src/vm/mod.rs
@@ -11,7 +11,7 @@ use crate::{
 
 use boa_gc::{custom_trace, Finalize, Trace};
 use boa_profiler::Profiler;
-use std::mem::size_of;
+use std::{future::Future, mem::size_of, ops::ControlFlow, pin::Pin, task};
 
 #[cfg(feature = "trace")]
 use boa_interner::ToInternedString;
@@ -292,7 +292,10 @@ impl Context<'_> {
         );
     }
 
-    fn trace_execute_instruction(&mut self) -> JsResult<CompletionType> {
+    fn trace_execute_instruction<F>(&mut self, f: F) -> JsResult<CompletionType>
+    where
+        F: FnOnce(Opcode, &mut Context<'_>) -> JsResult<CompletionType>,
+    {
         let bytecodes = &self.vm.frame().code_block.bytecode;
         let pc = self.vm.frame().pc as usize;
         let (_, varying_operand_kind, instruction) = InstructionIterator::with_pc(bytecodes, pc)
@@ -322,7 +325,7 @@ impl Context<'_> {
         }
 
         let instant = Instant::now();
-        let result = self.execute_instruction();
+        let result = self.execute_instruction(f);
         let duration = instant.elapsed();
 
         let fp = self.vm.frames.last().map(|frame| frame.fp as usize);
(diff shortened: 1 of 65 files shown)


raskad:
@HalidOdat you mentioned in #2904 that you wanted to refactor how we call functions to not recurse on the rust side. That would probably impact the design of this feature right?

jedel1043:
> Ah, seems like @HalidOdat and me arrived at the same problem from two different but related features. We could probably solve both problems using the same solution then :)

HalidOdat:
> I'm open to any other ideas you have! The gist of problem is: "How to bubble up JS function calls to the original Context::run call?".

The solution I thought of was to have only one `run`. we have a function that is called before every run (`prepare_vm_call` or something like this) that pushes arguments onto the stack and the call frame and sets up everything, then we call `run` (which just runs instructions in a loop).

The `Opcode::Call` instructions would only prepare the vm and continue with next instruction, because the frame is pushed we automatically get the instruction from the called function.

How do we know if we should return or continue executing until we run out of call frames?

For this we can have a flag in the call frame that returns in `run` (even if there are left over frames).

This would greatly simplify our calling logic! Not having to do a clone of the passed arguments and reversing them and pushing them back again instead just using the ones that where already pushed by the `Opcodes`. Not creating separate value and environment stacks (for ordinary functions at least).

Note: This stops recursion (and causing a stack overflow) from JavaScript, You can still recurse in rust, by recursively calling a function that calls `object.call(.., context)`.

jedel1043:
That sounds perfect to avoid recursing inside JS functions. We could implement that first, then think about how to avoid recursing inside Rust functions next.

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -148,6 +148,16 @@ macro_rules! generate_impl {
...
 
+            const COSTS: [usize; Self::MAX] = [
+                $(<generate_impl!(name $Variant $(=> $mapping)?)>::COST),*
+            ];
+
+            /// Cycle cost of this opcode.
+            #[must_use]
+            pub const fn cost(self) -> usize {
+                Self::COSTS[self as usize]
+            }
```
HalidOdat, line 159: I think we should change `cost` type to something smaller like `u8`, since I don't think an Opcode should be cost more than `256`.

It would cut the cost's table size by `8` (2048 bytes to 256 bytes)

HalidOdat (approved):
Looks good!

I don't think this addition would make the call refactor any harder :)

boa_engine/src/vm/mod.rs
```diff
@@ -180,8 +180,9 @@ impl Context<'_> {
 
         let completion = loop {
             let completion = self.run_next_instruction(|opcode, context| {
-                total_cost = total_cost.checked_add(opcode.cost()).unwrap_or(usize::MAX);
-                total_cost += opcode.cost();
+                total_cost = total_cost
+                    .checked_add(usize::from(opcode.cost()))
+                    .unwrap_or(usize::MAX);
```
HalidOdat, line 185: [`.saturating_add()`](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_add) would achieve the same thing without the `unwrap_or`

HalidOdat (approved):
Looks good to me! :)

We can reduce the `spend` function table which I did in #3401, we can merge this, or merge in into this?

jedel1043:
> we can merge this first, or merge into this?

Let's just merge it into this, since they're related changes :)

nekevss (approved):
Nice work! I like the updates! 😄

Issue #2732, opened 2023-03-23
Sporadic panics on `WeakMap` tests 


jedel1043:
So, apparently our CI has some sporadic panics on the following tests:

- test/built-ins/WeakSet/prototype/delete/delete-entry-initial-iterable.js
- test/built-ins/WeakSet/prototype/has/returns-false-when-object-value-not-present.js

![image](https://user-images.githubusercontent.com/38230983/227347248-89cdb7b6-38b5-44ee-b57e-992edaf73d40.png)

However, I tried to reproduce it in my pc without success, so a platform dependent bug maybe?

**Build environment:**

- OS: Arch Linux x86_64
- Version: 6.2.7-arch1-1
- Target triple: x86_64-unknown-linux-gnu
- Rustc version: rustc 1.68.0 (2c8cc3432 2023-03-06)

langston-barrett:
Potentially related to #2670.

jedel1043:
Found a more reliable reproduction of the issue:
```js
let list = []

for (let i = 0; i < 35; i++) {
  list.push(list);
}

let ws = new WeakSet(list);
```
This snippet always panics for me, which seems to imply that the cause of this is a garbage collection pass within a borrow.

PR #3049, opened 2023-06-19
Mark header of rooted ephemerons when tracing

jedel1043:
A very subtle bug that was discovered by @tunz. Also added additional checks on the test to make sure the ephemerons are correctly cleaned up.

### DIFF
--- a/boa_gc/src/lib.rs
+++ b/boa_gc/src/lib.rs
@@ -341,7 +341,7 @@ impl Collector {
         while let Some(node) = strong.get() {
             // SAFETY: node must be valid as this phase cannot drop any node.
             let node_ref = unsafe { node.as_ref() };
-            if node_ref.header.roots() > 0 {
+            if node_ref.header.roots() != 0 {
                 // SAFETY: the reference to node must be valid as it is rooted. Passing
                 // invalid references can result in Undefined Behavior
                 unsafe {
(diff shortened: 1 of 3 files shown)


boa_gc/src/lib.rs
```diff
@@ -367,12 +367,17 @@ impl Collector {
...
         //
+        //
         // 1. Get the naive list of ephemerons that are supposedly dead or their key is dead and
         // trace all the ephemerons that have roots and their keys are live. Also remove from
         // this list the ephemerons that are marked but their value is dead.
         while let Some(eph) = weak.get() {
             // SAFETY: node must be valid as this phase cannot drop any node.
             let eph_ref = unsafe { eph.as_ref() };
+            let header = eph_ref.header();
+            if header.roots() > 0 {
```
HalidOdat, line 378:
```suggestion
            if header.roots() != 0 {
```

It's an unsigned integer so `>` only excludes `0`, I think it would be clearer to just say `!= 0`

HalidOdat (approved):
Looks good to me! just a small nitpick :)

Could be this be related to the sporadic `weakmap` panics in test262?

jedel1043:
> Could be this be related to the sporadic `weakmap` panics in test262?

Not really. I tried to run the minimized reproduction from https://github.com/boa-dev/boa/issues/2732#issuecomment-1569275298 but the engine still panics.

HalidOdat (approved):
Looks good to me! :)

Razican:
Could we add a test for the bug?

raskad (approved):

PR #3052, opened 2023-06-20
Implement ephemeron-based weak map

### LINKED ISSUE #2596: Implement a weak `HashMap` based on `Ephemeron`s
We need an internal weak `HashMap` type that is integrated with our GC. With https://github.com/boa-dev/boa/pull/2586 we are merging an initial version that uses `HashMap<WeakGc<K>, V>`. We want to optimize this by basing our `WeakMap` implementation on `Ephemeron`s. To do this we need to implement or own `HashMap` type.

For more information see the discussion in https://github.com/boa-dev/boa/pull/2586

### LINKED ISSUE #2732: Sporadic panics on `WeakMap` tests

jedel1043:
This Pull Request fixes #2596 and fixes #2732.

It changes the following:

- Implements an ephemeron-based weak map using the `hashbrown` raw API.

I only exposed `WeakMap` as public, but it would be good to decide in the future if we want to expose the full `RawWeakMap` to our users.

~Marking as draft since this just needs documentation and a clippy run, but it's feature complete.~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,6 +60,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56fc6cf8dc8c4158eed8649f9b8b0ea1518eb62b544fe9490d66fa0b349eafe9"
+
+[[package]]
 name = "android-tzdata"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


HalidOdat (approved):
Looks good to me! :)

raskad (approved):
Nice work

PR #3053, opened 2023-06-20
Simplify/Refactor exception handling and last statement value

HalidOdat:
This PR simplifies some vm/bytecompiler logic which should make it easier to implement optimizations (since the optimizer has to account for all this complexity).

It changes the following:

- Remove opcodes `PopIfThrown`, `PopOnReturnAdd`, `PopOnReturnSub`, `CatchStart`, `CatchEnd` and `CatchEnd2`, `BreakLabel` 
- Add `SetReturnValue` and `GetReturnValue` opcodes (return value is currently stored on the call frame, which is not optimal will change that once we refactor function calling in the vm).
- Remove loop value in `EnvEntry`
- Only generate Last statement `use_expr` if needed.
- Add fp (frame pointer) to try `EnvEntry` to truncate the stack instead of `pop_on_return` on call frame (which was removed).
- Remove loop value logic from `break` and `continue` opcodes


jedel1043 (approved):
Really nice clean-up!

raskad (approved):
Great refactor!

PR #3071, opened 2023-06-26
Cache `cargo-tarpaulin` binary

jedel1043:
Testing if this change improves CI times for the codecov check.

Had to use another `cargo-install` action because the original `actions-rs/cargo-install` is apparently unmaintained, and users were reporting problems with the cache (https://github.com/actions-rs/install/issues/12)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,10 +22,10 @@ jobs:
           toolchain: stable
           override: true
           profile: minimal
-      - uses: actions-rs/install@v0.1
+      - name: Install cargo-tarpaulin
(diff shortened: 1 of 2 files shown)


jasonwilliams (approved):
LGTM, shame about actions-rs being unmaintained

jedel1043:
Let's see if the binary is correctly cached...

jedel1043:
Nice! This should reduce our CI times by 3-5 minutes, which is the time it takes to compile tarpaulin

HalidOdat (approved):
Looks good to me! :)

PR #3073, opened 2023-06-26
Correctly handle finally..loop..break

### LINKED ISSUE #3054: Panic at break in an infinte loop, in finally block
**Describe the bug**
In a try-catch-finally -- in the finally block, panic in a one line break statement after an infinite loop, _if there are no braces around the break statement._ 

**To Reproduce**
These JavaScripts crash with `thread 'main' panicked at 'stack was empty', boa_engine/src/vm/opcode/control_flow/finally.rs:137:59`
```js
try {
}
catch {
} finally {
 while(true) 
  break;
}
```
or 

```js
try {
}
catch {
} finally {
 for(;;)
  break;
}
```

**Expected behavior**
Should not panic. Javascript allows one-line statements after a condition to not be encompassed by braces.

**Build environment:**
• OS: Ubuntu
• Version: 20.04
• Target triple: x86_64-unknown-linux-gnu
• Rustc version: rustc 1.70.0

**Additional context**
Note that this Javascript does _not_ panic, i.e. if braces are present. It is also clearly breaking out of the infinite loop as "in finally" does get printed. 
```js
try {
}
catch {
} finally {
 while(true) {
  break;
 }
 console.log("in finally")
}
```

dirkdev98:
The match was too greedy, being executed for 'break' abrupt completions as well.

Closes #3054

### DIFF
--- a/boa_engine/src/tests/control_flow/mod.rs
+++ b/boa_engine/src/tests/control_flow/mod.rs
@@ -195,6 +195,24 @@ fn catch_binding_finally() {
 }
 
 #[test]
+fn finally_with_loop_break() {
+    run_test_actions([TestAction::assert_eq(
+        indoc! {r#"
+            try {
+              30;
+            }
+            catch {
+            } finally {
+              while(true) {
+                break;
+              }
+            }
+        "#},
+        30,
+    )]);
+}
+
+#[test]
 fn single_case_switch() {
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Great catch!

HalidOdat (approved):
Looks good to me!

Issue #3035, opened 2023-06-12
Add binaries to GitHub releases


CanadaHonk:
Binaries in GitHub releases for the CLI shell would be nice to have to quickly look at without having to clone and compile yourself.

PR #3084, opened 2023-06-27
create binaries for the releases

jasonwilliams:
This would resolve https://github.com/boa-dev/boa/issues/3035

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,29 +71,33 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
   release-binaries:
-    name: Publish for ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
+    name: Publish binaries
     strategy:
+      fail-fast: false
       matrix:
+        build: [linux, macos, win-msvc]
         include:
-          - os: ubuntu-latest
-            artifact_name: boa
-            asset_name: boa-linux-amd64
-          - os: windows-latest
-            artifact_name: boa.exe
-            asset_name: boa-windows-amd64
-          - os: macos-latest
-            artifact_name: boa
-            asset_name: boa-macos-amd64
-
+        - build: linux
+          os: ubuntu-20.04
+          rust: stable
+          target: x86_64-unknown-linux-gnu
+        - build: macos
+          os: macos-latest
+          rust: stable
+          target: x86_64-apple-darwin
+        - build: win-msvc
+          os: windows-2019
+          rust: stable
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -63,3 +63,31 @@ jobs:
...
+        include:
+          - os: ubuntu-latest
+            artifact_name: boa
+            asset_name: boa-linux-amd64
+          - os: windows-latest
+            artifact_name: boa.exe
+            asset_name: boa-windows-amd64
+          - os: macos-latest
+            artifact_name: boa
+            asset_name: boa-macos-amd64
```
jedel1043, line 81: It'll probably be better to also set the rust target on compilation, just in case.
Example: https://github.com/crate-ci/cargo-release/blob/db00d8a078799d81546416031e2944e775be62e9/.github/workflows/post-release.yml#L55-L74

jedel1043: Also, that release script also uses the target name to name the asset itself, which is arguably better for labeling

.github/workflows/release.yml
```diff
@@ -63,3 +63,31 @@ jobs:
...
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build
+      run: cargo build --release --locked
```
HalidOdat, line 86:
```suggestion
      run: cargo build --release --locked --bin boa
```

We should probably build only the `boa` binary to avoid building all the examples.

.github/workflows/release.yml
jasonwilliams: I've made adjustments to this, the asset name may be confusing with the "boa" name in it, so ive slightly adjusted it

.github/workflows/release.yml
```diff
@@ -69,3 +69,38 @@ jobs:
...
+        - build: win-msvc
+          os: windows-2019
+          rust: stable
+          target: x86_64-pc-windows-msvc
+          asset_name: boa-windows-amd64
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build
+      run: cargo build --release --locked --bin boa
```
jedel1043, line 99:
```suggestion
      run: cargo build --target ${{ matrix.target }} --verbose --release --locked --bin boa
```

jedel1043 (approved):
Looks good!

raskad (approved):

HalidOdat (approved):
Looks good to me! :)

RageKnify (approved):

PR #3041, opened 2023-06-15
Implement Startup Snapshot

HalidOdat:
This Pull Request fixes/closes #1372 .

This is WIP implementation of startup snapshots, it needs a lot of work.

- TODO:

- [ ] Figure out optimal snapshot file format.
- [ ] Implement full serialization 
- [ ] Implement full deserialization
- [ ] Implement snapshot checksum
- [ ] Implement snapshot versioning

Issue #2626, opened 2023-03-01
Reusing `Codeblock`s between `Context`s is not property supported. 


jedel1043:
**Describe the bug**
Trying to run a `Codeblock` compiled by a `Context` in a separate `Context` causes a panic on runtime. 

**To Reproduce**
```rust
use boa_engine::{Context, Source};

fn main() {
    let script = r#"
        a = 5;
    "#;

    let context = &mut Context::default();
    let code = context.parse_script(Source::from_bytes(script)).unwrap();
    let code = context.compile_script(&code).unwrap();

    println!(
        "First execution: {}",
        context.execute(code.clone()).unwrap().display()
    );

    let context = &mut Context::default();

    println!(
        "Second execution: {}",
        context.execute(code.clone()).unwrap().display()
    );
}
```

This uses the `main` branch of Boa, but it should be also reproducible from the latest released version. 

**Expected behaviour**

Either run correctly or throw an error saying that codeblocks cannot be reused on other `Context`s.

**Current behaviour**
```
First execution: 5
thread 'main' panicked at 'string disappeared', /home/jedel/SDK/src/Rust/boa/boa_interner/src/lib.rs:422:30
... (stack trace)
```

**Build environment:**

- OS: Arch linux 
- Version: 6.1.11-arch1-1
- Target triple: stable-x86_64-unknown-linux-gnu
- Rustc version: rustc 1.67.1 (d5a82bbd2 2023-02-07)

**Additional context (no pun intented):**

This is where the fun part begins. AFAIK, most engines support using codeblocks between contexts, but we don't support that because of two reasons:

- Codeblocks use `Sym`s to internally represent variable bindings. This assumes that the containing `Context` has on its interner the binding names used in the execution.
- Our compiler currently uses the `Context` itself at compile time to check for invariants related to bindings.

To illustrate the second point, this code correctly throws:

```rust
use boa_engine::{Context, Source};

fn main() {
    let script = r#"
        let a = 50;
        {
            var a = 5;
        }
        a
    "#;

    let context = &mut Context::default();

    println!(
        "First execution: {}",
        context
            .parse_script(Source::from_bytes(script))
            .unwrap_err()
    );
}
```

While this code runs and returns `5`:

```rust
fn main() {
    let script = r#"
        {
            var a = 5;
        }
        a
    "#;

    let context = &mut Context::default();
    let code = context.parse_script(Source::from_bytes(script)).unwrap();
    let code = context.compile_script(&code).unwrap();

[...]

PR #3087, opened 2023-06-29
Implement virtual callstack for functions 

jedel1043:
This is the first prototype of a vm design that uses a virtual callstack for JS functions and Rust functions.

Right now it's full of hacks and compatibility patches in order to be able to progressively experiment and bugfix, but that can be left for later.

It currently uses the `next-gen` crate to emulate generators in stable, which is the way this design unifies JS calls with Rust calls.

## Design notes

All functions now return a `JsResult<CallResult>` struct, defined as:
```rust
pub enum CallResult<T> {
    ///
    Value(T),
    ///
    Coroutine(JsCoroutine),
    ///
    DirectCall(CallContext),
}

pub type JsCoroutine =
    Pin<Box<dyn Generator<JsResult<JsValue>, Yield = CallContext, Return = JsResult<JsValue>>>>;

pub struct CallContext {
    pub(crate) f: JsObject,
    pub(crate) this: JsValue,
    pub(crate) args: ThinVec<JsValue>,
}
```

This new return value can signal the VM about what to do on several cases:
- If a function doesn't need to call any JS functions, it can directly return `Value`, which indicates to the VM that the function finished executing and doesn't need to be pushed to the virtual callstack.
- If a function returns `DirectCall(CallContext)`, it's a tail call, meaning the VM can directly call the inner function without storing the current context of the caller.
- If a function returns `Coroutine (JsCoroutine)`, it signals the VM that the callee has inner calls to other JS functions.

The coroutine design looks a bit daunting, but it's essentially divided in steps:
1. The native Rust function returns a generator with inner yields instead of direct calls to JS functions. Each yield corresponds to a JS call that needs to be resolved by the VM.
2. The VM receives the coroutine and pushes it forward until it yields, receiving the JS function that needs to be called in order for the coroutine to progress.
3. The VM prepares the JS call and stores the coroutine in a coroutine stack for future resumption.
4. The JS function is called, returning a result.
5. The VM resumes the coroutine with the result of the JS call.
6. The VM repeats steps 2 to 5 if the coroutine yields again.
7. The coroutine returns its final result.

This bridges the gap between Rust functions and JS functions, making it possible to track both of them with a single callstack :D

cc @HalidOdat

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -430,6 +430,7 @@ dependencies = [
  "indoc",
  "itertools",
  "jemallocator",
+ "next-gen",
  "num-bigint",
  "num-integer",
  "num-traits",
@@ -2727,6 +2728,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "next-gen"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1962f0b64c859f27f9551c74afbdbec7090fa83518daf6c5eb5b31d153455beb"
+dependencies = [
+ "next-gen-proc_macros",
+ "unwind_safe",
+]
+
+[[package]]
+name = "next-gen-proc_macros"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a59395d2ffdd03894479cdd1ce4b7e0700d379d517f2d396cee2a4828707c5a0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "nibble_vec"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4478,6 +4500,12 @@ name = "unsafe-libyaml"
 version = "0.2.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1865806a559042e51ab5414598446a5871b561d21b6764f2eabb0dd481d880a6"
+
+[[package]]
+name = "unwind_safe"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0976c77def3f1f75c4ef892a292c31c0bbe9e3d0702c63044d7c76db298171a3"
 
 [[package]]
 name = "url"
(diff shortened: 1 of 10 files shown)


HalidOdat:
Took a look at the implementation looks good, but I had some concerns about complicating the the implementation of functions and that it would make #3041 impossible or very hard (since we need to store function pointer in a list that can be indexed, we can't serialize and deserialize the pointer but we can an index). using the `#[generator]` macro really obscures this and the state can't be serialized/deserialized (this may/may not apply to closure functions)

Since we have a solution for JavaScript functions (as I mentioned in #3044), So I thought we could have some (not all) built-in functions implemented in JavaScript (at least the ones that call a lot of functions, like `forEach`).

Benefits:

 - Easier to implement built-in functions
 - Once we implement inlining optimization we will be able to inline builtin functions (since they have bytecode)
 - Faster rust build-times, less code to compile.
 - No need to add generator function type.
 - Makes implementing snapshots a bit easier since we don't have to a big array of function pointers.
 - We could implement lazy compilation where it only compiles into bytecode when needed.

Disadvantages:

 - We would have to parse and compile the code at runtime (this can be mitigated by fully implementing snapshot's and the lazy compilation) 
 - Maybe? increased memory usage, instead of storing machine code, we store a `CodeBlock` (we have a lot of room to improve our bytecode) also having #2626 would allow use to share it between `Context`s
 - Less efficient functions, probably will be the case until we implement jiting, we could generate better bytecode though, having intrisics like `ToNumber` so when calling `ToNumber(arg)` it directly emits `GetName "arg"` followed by `ToNumber` opcode, like wise for very common spec function's `ToString`, `RequireObjectCoercible`, this would require small change in the bytecompiler (we already do this for direct `eval` calls that emits `EvalCall` opcode)


I checked some engines to see what other engines do and apparently they do something very simular to this, [spidermonaky has this](https://github.com/mozilla/gecko-dev/blob/master/js/src/builtin/Array.js), v8 had this too, but deprecated it in favor of creating a [domain specific language `Torque`](https://v8.dev/docs/torque-builtins)

jedel1043:
@HalidOdat Yeah, completely agree. This was mostly an experiment to see if we could leverage generator crates for the virtual stack, but seeing the many difficulties I had while implementing this, I think we'll have to wait until there's proper support for coroutines in Rust.

I'll remove the coroutine parts of this PR and only implement the virtual stack for JS to JS calls.

PR #3092, opened 2023-06-30
Clarify usage section in `README.md`

postmeback:
There were certain installation steps missing in the ReadMe file, which I added in order to get it corrected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ then go to `http://localhost:8080`.
 
 - Clone this repo.
 - Run `cargo build`
-- Run with `cargo run -- test.js` where `test.js` is an existing JS file with any JS valid code. The file has to be in the root directory.
+- Run with `cargo run -- test.js` in the project root directory where `test.js` is a path to an existing JS file with any JS valid code.
 - If any JS doesn't work then it's a bug. Please raise an [issue](https://github.com/boa-dev/boa/issues/)!
 
 ### Example
--- a/README.md
+++ b/README.md
@@ -64,7 +64,6 @@ then go to `http://localhost:8080`.
 ## Usage
 
 - Clone this repo.
-- Run `cargo build`
 - Run with `cargo run -- test.js` in the project root directory where `test.js` is a path to an existing JS file with any JS valid code.
 - If any JS doesn't work then it's a bug. Please raise an [issue](https://github.com/boa-dev/boa/issues/)!
 
--- a/README.md
+++ b/README.md
@@ -64,7 +64,8 @@ then go to `http://localhost:8080`.
 ## Usage
 
 - Clone this repo.
-- Run with `cargo run -- test.js` where `test.js` is an existing JS file with any JS valid code.
+- Run `cargo build`
+- Run with `cargo run -- test.js` where `test.js` is an existing JS file with any JS valid code. The file has to be in the root directory.
 - If any JS doesn't work then it's a bug. Please raise an [issue](https://github.com/boa-dev/boa/issues/)!
 
 ### Example
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ then go to `http://localhost:8080`.
 ## Usage
 
 - Clone this repo.
-- Run with `cargo run -- test.js` in the project root directory where `test.js` is a path to an existing JS file with any JS valid code.
+- Run with `cargo run -- test.js` in the project root directory where `test.js` is a path to an existing JS file with any valid JS code.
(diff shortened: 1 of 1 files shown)


jedel1043:
IIRC `cargo run -- test` is equivalent to `cargo build & cargo run -- test`, so I don't see why this should fail. Do you have an error log that we could look at?

postmeback:
I restarted my system, and executed "cargo run -- Hello2.js", it ran successfully. But after installation, when I was trying to do so, it was failing. Sorry, I don't have any screenshots of that.

README.md
```diff
@@ -64,7 +64,8 @@ then go to `http://localhost:8080`.
 ## Usage
 
 - Clone this repo.
-- Run with `cargo run -- test.js` where `test.js` is an existing JS file with any JS valid code.
+- Run `cargo build`
+- Run with `cargo run -- test.js` where `test.js` is an existing JS file with any JS valid code. The file has to be in the root directory.
```
postmeback, line 68: It gives the person a clear idea to where the file should be added; it's worth mentioning.

README.md
```diff
@@ -64,7 +64,8 @@ then go to `http://localhost:8080`.
 ## Usage
 
 - Clone this repo.
-- Run with `cargo run -- test.js` where `test.js` is an existing JS file with any JS valid code.
+- Run `cargo build`
```
postmeback, line 67: For the first time, to be on the safer side, I believe we should mention this step.

README.md
HalidOdat: > It gives the person a clear idea to where the file should be added; it's worth mentioning.

It doesn't have to be in the root dir, it just require a path to the file. so `cargo run -- ../some-folder-outside-root/works.js`

README.md
HalidOdat: I don't think this is needed, `cargo` should **always** build if you `run` (and it is not already built). If it doesn't, that means it is a bug in `cargo` and should be reported in the `cargo` repo https://github.com/rust-lang/cargo/

README.md
postmeback: Although, alternative suggestions are always welcome

HalidOdat: > @HalidOdat , you are correct. But if somebody changes the directory to the target folder, and then runs `cargo-run`, then it will fail. I think, for better clarity, we should be mentioning it.

This, should be obvious, `cargo` doesn't work if `Cargo.toml` is not found (applies to pretty much every build system)

> Although, alternative suggestions are always welcome

We can make it explicit:

"Run with `cargo run -- test.js` in the project root directory where `test.js` is a path to an existing JS file with any JS valid code."

postmeback: Also, let's mention, if `cargo run` fails the first time, then run  `cargo build`.

HalidOdat: > Also, let's mention, if `cargo run` fails the first time, then run `cargo build`.

https://github.com/boa-dev/boa/pull/3092#discussion_r1249588880

postmeback: @HalidOdat I would request you to put suggestions in this PR. That would be helpful.

nekevss: It may be a bit more succinct to write.

```markup
 - Run with `cargo run -- test.js` in the root directory where `test.js` is a path to any existing valid JS file.
```

Do we need to double reference JS file and valid JS code?

postmeback: Double reference is not needed.

postmeback: > It may be a bit more succinct to write.
> 
> ```
>  - Run with `cargo run -- test.js` in the root directory where `test.js` is a path to any existing valid JS file.
> ```
> 
> Do we need to double reference JS file and valid JS code?

It looks perfect.

README.md
Razican: Note that we should still remove the `cargo build` part, since it's not needed and is adds complexity to the steps.

postmeback: Removed that part.

Razican (approved):

README.md
```diff
@@ -64,7 +64,7 @@ then go to `http://localhost:8080`.
 ## Usage
 
 - Clone this repo.
-- Run with `cargo run -- test.js` where `test.js` is an existing JS file with any JS valid code.
+- Run with `cargo run -- test.js` in the project root directory where `test.js` is a path to an existing JS file with any JS valid code.
```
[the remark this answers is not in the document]
postmeback: Done

postmeback:
Hey, can this be merged ?

PR #3096, opened 2023-07-02
`TypedArray.prototype.values()` and `TypedArray.prototype[@@iterator]` should be equal

HalidOdat:
Make `TypedArray.prototype.values()` and `TypedArray.prototype[@@iterator] share the same function object.


jedel1043 (approved):
Nice catch!

Razican (approved):

Issue #2773, opened 2023-04-02
GcRefCell.borrow_mut is too expensive


tunz:
When `Object.borrow_mut()` is called, boa tries to root/unroot all the properties. So, even if we just want to set a single element, boa accesses all the other elements. I see that this overhead is taking almost 90% of run time in QuickJS NavierStokes benchmark.

In the following example, running `arr[0] = {};` is much slower when there are more elements.
```
% cat test.js
let arr = [];
for (let i = 0; i < 30000; i++)
  arr.push({});

let arr2 = [];

for (let i = 0; i < 30000; i++)
  arr[0] = {};
% time ./target/release/boa ./test.js
undefined
./target/release/boa ./test.js  11.04s user 0.02s system 99% cpu 11.146 total

% cat test2.js
let arr = [];
for (let i = 0; i < 30000; i++)
  arr.push({});

let arr2 = [];

for (let i = 0; i < 30000; i++)
  arr2[0] = {};
% time ./target/release/boa ./test2.js
undefined
./target/release/boa ./test2.js  6.57s user 0.01s system 98% cpu 6.655 total
```

I guess this design is for such case that we add a `Gc<GcRefCell<T>>` object to another `Gc<GcRefCell<T>>` object.
```rust
let v = Gc::new(GcRefCell::new(Vec::new()));
v.borrow_mut().push(Gc::new(GcRefCell::new(123)));
```
When `123` object is pushed to the `v` object, it should be unrooted because `v` is the root now. But, `Vec.push` does not know that information, and only the `GcRefMut` returned by `borrow_mut` knows that. So, what it does is just to root all the children when borrow starts, and unroot all the children again when borrow ends.

I'm not sure what's the best approach to fix this issue. I personally think we should not use `Gc` in another `Gc` object. I think we should allow `Gc` objects to keep `GcPointer` instead, and manage roots in a single place.

jedel1043:
Yeah, that's inherent of our current GC design. We have an open issue (#2631) about investigating other API designs for our GC, which should allow us to solve issues like this one.

jasonwilliams:
Hey @tunz thanks for raising this, that does indeed sound quite expensive so I’m glad you brought it up. I’d be interested in seeing if we can solve this somehow. I agree gc object within a gc object doesn’t sound ideal.

> I think we should allow Gc objects to keep GcPointer instead, and manage roots in a single place.

could you elaborate on this idea a bit more? How would this work with objects that have multiple references?

[...]

PR #3109, opened 2023-07-04
Find roots when running GC rather than runtime

tunz:
Attempt to address the issue #2773.

The existing implementation had an expensive overhead of managing root counts, especially for mutable borrow of GcRefCell. Instead of managing the root counts, this change counts the number of Gc/WeakGc handles located in Gc heap objects and total number of them. Then, we can find whether there is a root by comparing those numbers.

For example, in the following case, there are 2 `Gc`s for `Gc<i32>` type. One is allocated on the stack, and the other one is located in another Gc heap object.
```rust
#[derive(Trace, Finalize)]
struct Obj {
  Gc<i32> v,
}
let num = Gc::new(123);
let v = Gc::new(Obj { v: num.clone() });
```
Then, we'll iterate and trace the two objects when running GC, and mark the referenced `Gc`s as non root. We know that there are 2 `Gc`, but only one of them is marked as non-root. So, we can find that the remaining one is the root.

As a trade-off, this approach has an overhead in the the garbage collection time, but I believe there is a room for improvement such as tracing non roots in parallel.

Here is a result of benchmarks:
```
Richards: 25.5 -> 26.4
DeltaBlue: 28.7 -> 29.1
Crypto: 33.4 -> 38.0
RayTrace: 101 -> 107
EarleyBoyer: 88.3 -> 95.3
Splay: 104 -> 93.2
NavierStokes: 9.94 -> 75.3
---
SCORE: 41.9 -> 57.5
```

There is one regression in `Splay` benchmark because it has a lot of DeclarativeEnvironment allocation triggering GC, so it's more impacted by the overhead of GC than benefit of faster property accesses.


boa_gc/src/pointers/gc.rs
```diff
@@ -15,11 +15,10 @@ use std::{
...
 
-use super::rootable::Rootable;
-
 /// A garbage-collected pointer type over an immutable value.
 pub struct Gc<T: Trace + ?Sized + 'static> {
-    pub(crate) inner_ptr: Cell<Rootable<GcBox<T>>>,
+    pub(crate) inner_ptr: Cell<NonNull<GcBox<T>>>,
+    pub(crate) handle: Rc<GcHandle>,
     pub(crate) marker: PhantomData<Rc<T>>,
 }
```
jedel1043, line 23: Hmmm, can we add another layer of indirection to this?
We were deliberately ensuring this type is pointer sized, otherwise space optimizations such as #1830 will become impossible.

Alternatively, if equal `Gc`s share the same counter, isn't it possible to put the counter inside `GcBox` instead?

jedel1043 (changes requested):
Really liking the big performance wins! However, I have some suggestions about the design.

boa_gc/src/pointers/gc.rs
HalidOdat: +1 for @jedel1043's suggestion

Also we should be able to remove the `Cell` from `Gc<T>` internal pointer, since we don't have the old way of rooting, which should make `Option<Gc<T>>` pointer size (instead of two pointer width).

tunz: Updated to put the counters in GcBox/EphemeronBox. It's a bit slower than the previous one. I update the benchmark result in the description.

boa_gc/src/internals/ephemeron_box.rs
```diff
@@ -163,6 +141,14 @@ pub(crate) trait ErasedEphemeronBox {
...
     unsafe fn trace(&self) -> bool;
 
+    fn trace_non_roots(&self);
+
+    fn get_ref_count(&self) -> u32;
+
+    fn get_non_root_count(&self) -> u32;
+
+    fn reset_non_root_count(&self);
+
```
jedel1043, line 151: You can remove this by using the `ErasedEphemeronBox::header` method instead.

tunz:
There are too many `must_use` clippy errors in boa_engine. I guess something blocking clippy is removed. So I disabled it for now. Please let me know if we want to fix them all in this PR.

boa_gc/src/internals/gc_box.rs
```diff
@@ -5,80 +5,54 @@ use std::{
...
 ///
 /// These flags include:
-///  - Root Count
 ///  - Mark Flag Bit
 ///
 /// The next node is set by the `Allocator` during initialization and by the
 /// `Collector` during the sweep phase.
 pub(crate) struct GcBoxHeader {
-    roots: Cell<usize>,
+    marked: Cell<bool>,
```
nekevss, line 17: Any thoughts on using a bit in the `ref_count` here over a full bool?

tunz: I was hesitant to keep the mark bit in `ref_count` since there was 1~2% overhead, but it seems like putting the mark bit into `non_root_count` does not have that much of overhead. Updated to include the mark bit in `non_root_count`.

jedel1043 (approved):
Really nice work! We can resolve the lints on a follow-up PR :)

boa_gc/src/internals/ephemeron_box.rs
```diff
@@ -4,80 +4,85 @@ use std::{
...
     /// Creates a new `EphemeronBoxHeader` with a root of 1 and next set to None.
     pub(crate) fn new() -> Self {
         Self {
-            roots: Cell::new(1),
+            ref_count: Cell::new(1),
+            non_root_count: Cell::new(0),
             next: Cell::new(None),
         }
     }
+    /// Returns the `EphemeronBoxHeader`'s current ref count
```
Razican, line 34: We should leave an empty line before the comment here

HalidOdat (approved):
Great work @tunz !

PR #3113, opened 2023-07-05
Quick fix

jasonwilliams:
- remove  `--no-git-push`
- Update node version to 20
- Set correct binary name
- release_binaries task should depend on `publish`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,7 +39,7 @@ jobs:
             skip
 
   doc-publish:
-    # needs: publish
+    needs: publish
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -36,7 +36,6 @@ jobs:
             --from-git \
             --yes \
             --no-git-commit \
-            --no-git-push \
             --no-individual-tags \
```
jedel1043, line 39:
```suggestion
```

Also doesn't work with `--no-individual-tags` apparently

HalidOdat (approved):

.github/workflows/release.yml
```diff
@@ -50,7 +48,7 @@ jobs:
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
```
jedel1043, line 48:
```suggestion
        uses: baptiste0928/cargo-install@v2.1.0
        with:
          crate: wasm-pack
```

jedel1043 (approved):

PR #3117, opened 2023-07-08
Allow escaped yield and await in labelled statement

raskad:

- Allow escaped yield and await in labelled statement


HalidOdat (approved):
Nice catch! :)

jedel1043 (approved):

PR #3129, opened 2023-07-10
Implement i128/u128 to JsBigInt conversions

AlvinKuruvilla:
This commit implements trait conversions to JsValue for i128/u128. This addresses #1970

This Pull Request fixes/closes #{1970}.

It changes the following:

- Adds trait conversions for the missing i128 and u128 rust types to JsValue
-
-

### DIFF
--- a/boa_engine/src/bigint.rs
+++ b/boa_engine/src/bigint.rs
@@ -389,6 +389,7 @@ impl From<u64> for JsBigInt {
         }
     }
 }
+
 impl From<i128> for JsBigInt {
     #[inline]
     fn from(value: i128) -> Self {
@@ -406,6 +407,7 @@ impl From<u128> for JsBigInt {
         }
     }
 }
+
 impl From<isize> for JsBigInt {
     #[inline]
     fn from(value: isize) -> Self {
@@ -423,6 +425,7 @@ impl From<usize> for JsBigInt {
         }
     }
 }
+
 /// The error indicates that the conversion from [`f64`] to [`JsBigInt`] failed.
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub struct TryFromF64Error;
--- a/boa_engine/src/bigint.rs
+++ b/boa_engine/src/bigint.rs
@@ -407,22 +407,6 @@ impl From<usize> for JsBigInt {
         }
     }
 }
-impl From<i128> for JsBigInt {
-    #[inline]
-    fn from(value: i128) -> Self {
-        Self {
-            inner: Rc::new(RawBigInt::from(value)),
-        }
-    }
-}
-impl From<u128> for JsBigInt {
-    #[inline]
-    fn from(value: u128) -> Self {
-        Self {
-            inner: Rc::new(RawBigInt::from(value)),
-        }
-    }
-}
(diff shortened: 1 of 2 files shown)


AlvinKuruvilla:
Yeah that makes sense, I saw this [issue](https://github.com/boa-dev/boa/issues/1970) open and a lack of activity, so I thought it might be a good way to get my feet wet a little bit. Is there no longer any interest in doing this though?

jedel1043:
> Hmmm I'm not sure if we want to do this. There is no native type in JavaScript equivalent to i/u128. And using a BigInt is not the same thing, since it's stored in the heap and performance is different.

Note that this PR implements `From` for `JsBigInt` and `Numeric`, not `JsValue`, and both traits are also implemented for `i/u64`.

jedel1043:
I'd support merging this change. Converting to `Numeric` is a lot more explicit about allocations than directly converting to `JsValue`, and if we want to make it a lot more explicit we could specify on the `Numeric` docs that all conversions from integer types allocate.

boa_engine/src/bigint.rs
```diff
@@ -407,7 +407,22 @@ impl From<usize> for JsBigInt {
...
+    }
+}
+impl From<u128> for JsBigInt {
+    #[inline]
+    fn from(value: u128) -> Self {
+        Self {
+            inner: Rc::new(RawBigInt::from(value)),
+        }
+    }
+}
```
HalidOdat, line 425: This is fine (conversion from `u/i128` to `JsBigInt`), but I don't think we should add for `Numeric`.

AlvinKuruvilla: Okay I can make that change

AlvinKuruvilla: Resolved by the latest commit to the PR, thanks for the feedback @HalidOdat

HalidOdat: I meant that we should keep the `u/i128` to `JsBigInt` conversion, and only remove the `u/i128` to `Numeric` conversion.

AlvinKuruvilla: Oh sorry. But the `JsBigInt` conversion requires `Numeric` trait to be implemented

AlvinKuruvilla: Sorry, that was my mistake I had it backward, I re-added the trait impl for `JsBigInt`

boa_engine/src/bigint.rs
```diff
@@ -407,7 +422,6 @@ impl From<usize> for JsBigInt {
         }
     }
 }
-
```
HalidOdat, line 410: Could we add the newline back here.

boa_engine/src/bigint.rs
```diff
@@ -389,7 +389,22 @@ impl From<u64> for JsBigInt {
...
+impl From<i128> for JsBigInt {
+    #[inline]
+    fn from(value: i128) -> Self {
+        Self {
+            inner: Rc::new(RawBigInt::from(value)),
+        }
+    }
+}
+impl From<u128> for JsBigInt {
```
HalidOdat, line 400:
```suggestion
}

impl From<u128> for JsBigInt {
```

Add new line like other functions.

HalidOdat (approved):
Looks good! just some small nitpicks :)

jedel1043 (approved):

Issue #3100, opened 2023-07-03
Implement `findLast` & `findLastIndex` on TypedArray's


dirkdev98:
**ECMASCript feature**

Implement missing [`%TypedArray%.prototype.findLast ( predicate [ , thisArg ] )`](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.findlast) and [`Array.prototype.findLastIndex ( predicate [ , thisArg ] )`](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.findlastindex) features.

Implementing these should fix most of the remaining test262 tests with feature `array-find-from-last`.

**Example code**

```js
// From test/built-ins/TypedArray/prototype/findLastIndex/return-index-predicate-result-is-true.js

var sample = new Int32Array([39, 3, 9]);
var called = 0;

var result = sample.findLastIndex(function() {
  called++;
  return true;
});

assert.sameValue(result, 2, "returned true on sample[2]");
assert.sameValue(called, 1, "predicate was called once");
```

dirkdev98:
I want to implement this!

At some point the spec changed to use [`FindViaPredicate ( O, len, direction, predicate, thisArg )`](https://tc39.es/ecma262/#sec-findviapredicate) for `Array.prototype.find`, `Array.prototype.findLast`, `%TypedArray%.prototype.findLast`, et al. Do we want to refactor the existing implementations to use this? 

If so, where would be a good place to add this? Under `built-ins/array/mod.rs` as a `pub(crate) fn find_with_predicate`? I don't have any experience with Rust modules yet, so a bit unsure what the way to go would be.

HalidOdat:
> At some point the spec changed to use [`FindViaPredicate ( O, len, direction, predicate, thisArg )`](https://tc39.es/ecma262/#sec-findviapredicate) for `Array.prototype.find`, `Array.prototype.findLast`, `%TypedArray%.prototype.findLast`, et al. Do we want to refactor the existing implementations to use this?

Yes, that would be optimal, This could be done In a separate PR.

> If so, where would be a good place to add this? Under built-ins/array/mod.rs as a pub(crate) fn find_via_predicate? I don't have any experience with Rust modules yet, so a bit unsure what the way to go would be.

Yes, that's a good place :) , I usually put the helper fuctions based where it is defined in the spec then by frequency of use, since they are both used equally. array module is best.

dirkdev98:
[...]

PR #3134, opened 2023-07-11
Refactor `Array.prototype.find*` and TypedArray variants to use `FindViaPredicate`

dirkdev98:
This Pull Request references #3100 

It changes the following:

- It aligns the implementations of `Array.prototype.find{Index,Last,LastIndex}` and the existing TypedArray variants to use [`FindViaPredicate ( O, len, direction, predicate, thisArg )`](https://tc39.es/ecma262/#sec-findviapredicate)

### DIFF
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -3042,7 +3042,9 @@ pub(crate) fn find_via_predicate(
 ) -> JsResult<(JsValue, JsValue)> {
     // 1. If IsCallable(predicate) is false, throw a TypeError exception.
     if !predicate.is_callable() {
-        return Err(JsNativeError::typ().with_message("predicate is not callable" ).into());
+        return Err(JsNativeError::typ()
+            .with_message("predicate is not callable")
+            .into());
     }
 
     let indices = match direction {
@@ -3074,7 +3076,7 @@ pub(crate) fn find_via_predicate(
 
         if test_result {
             // e. If ToBoolean(testResult) is true, return the Record { [[Index]]: 𝔽(k), [[Value]]: kValue }.
-            return Ok((JsValue::new(k), k_value))
+            return Ok((JsValue::new(k), k_value));
         }
     }
 
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -36,6 +36,13 @@ pub(crate) use array_iterator::ArrayIterator;
 #[cfg(test)]
 mod tests;
 
+/// Direction for `find_via_predicate`
+#[derive(Clone, Copy, Eq, PartialEq)]
+pub(crate) enum Direction {
+    Ascending,
+    Descending,
+}
+
 /// JavaScript `Array` built-in implementation.
 #[derive(Debug, Clone, Copy)]
 pub(crate) struct Array;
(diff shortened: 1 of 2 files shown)


boa_engine/src/builtins/array/mod.rs
```diff
@@ -1519,38 +1526,20 @@ impl Array {
         // 2. Let len be ? LengthOfArrayLike(O).
         let len = o.length_of_array_like(context)?;
 
-        // 3. If IsCallable(predicate) is false, throw a TypeError exception.
         let predicate = args.get_or_undefined(0).as_callable().ok_or_else(|| {
```
dirkdev98, line 1529: This check is now obsolete since `find_via_predicate` checks if the `predicate.is_callable()`, but kept it for the completer error message in all variants. Let me know if this should be removed anyways.

dirkdev98:

| Total | 94,858 | 94,858 | 0 |
| Passed | 74,724 | 74,724 | 0 |
| Ignored | 18,828 | 18,828 | 0 |
| Failed | 1,306 | 1,306 | 0 |
| Panics | 0 | 0 | 0 |
| Conformance | 78.77% | 78.77% | 0.00% |

boa_engine/src/builtins/array/mod.rs
Razican: I wouldn't do this extra check, since the spec doesn't have it.

boa_engine/src/builtins/array/mod.rs
```diff
@@ -1628,35 +1603,20 @@ impl Array {
...
-            }
-            // e. Set k to k - 1.
+        // 3. Let findRec be ? FindViaPredicate(O, len, descending, predicate, thisArg).
+        let find_rec =
+            find_via_predicate(&o, len, Direction::Descending, predicate, this_arg, context);
+
+        // 4. Return findRec.[[Value]].
+        match find_rec {
+            Ok((_, value)) => Ok(value),
+            Err(err) => Err(err),
```
HalidOdat, line 1618:
```suggestion
        let (_, value) = find_via_predicate(&o, len, Direction::Descending, predicate, this_arg, context)?;

        // 4. Return findRec.[[Value]].
        Ok(value)
```

We can use `?` error propagation and `let` pattern matching to remove `match` here, same for other call sites

HalidOdat (commented):
Looks good just a nitpick :)

boa_engine/src/builtins/array/mod.rs
HalidOdat: I agree that the check is not needed, but it would be nice to have better error messages (especially since we don't have line numbers, yet).

Maybe we could have a parameter `caller_name: &str` on find_via_predicate that pass the caller `"Array.prototype.find"` and and construct the error message. (`format!("{caller_name}": predicate not a function)`).

dirkdev98: Seems like a good solution thanks, implemented in fd1af53

boa_engine/src/builtins/array/mod.rs
dirkdev98: Thanks for the review! Applied in fd1af53 looks much cleaner now ;)

jedel1043 (approved):
Everything looks great!

HalidOdat (approved):
Looks perfect to me! :)

PR #3171, opened 2023-07-25
Fix TypedArrayConstructors tests

raskad:
This Pull Request fixes all remaining `TypedArrayConstructors` tests.

It changes the following:

- Fix typo in TypedArray constructor.
- Update the spec and code for some TypedArray and ArrayBuffer creation functions.
- Add explicit early `drop`s to burrowed `Gc`s in TypedArray methods where needed.
- Fix logic errors in integer-indexed exotic object methods.

### DIFF
--- a/boa_engine/src/builtins/array_buffer/mod.rs
+++ b/boa_engine/src/builtins/array_buffer/mod.rs
@@ -379,7 +379,7 @@ impl ArrayBuffer {
         self.array_buffer_data.is_none()
     }
 
-    /// `25.1.2.4 CloneArrayBuffer ( srcBuffer, srcByteOffset, srcLength, cloneConstructor )`
+    /// `25.1.2.4 CloneArrayBuffer ( srcBuffer, srcByteOffset, srcLength )`
     ///
     /// More information:
     ///  - [ECMAScript reference][spec]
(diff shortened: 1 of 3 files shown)


boa_engine/src/builtins/array_buffer/mod.rs
```diff
@@ -389,39 +389,51 @@ impl ArrayBuffer {
...
+        copy_data_block_bytes(
+            target_block,
+            0,
+            src_block,
+            src_byte_offset as usize,
+            src_length as usize,
+        );
 
         // 6. Return targetBuffer.
+        drop(target_buffer_mut);
```
HalidOdat, line 436: do we need to `drop` here since it will drop anyway at the end of the function.

boa_engine/src/builtins/typed_array/mod.rs
```diff
@@ -2619,11 +2661,14 @@ impl TypedArray {
...
+                drop(target_buffer_obj_borrow);
+                drop(a_borrow);
             }
+        } else {
+            drop(a_borrow);
         }
 
-        drop(a_borrow);
-
         // 15. Return A.
```
HalidOdat, line 2672: Same here.

HalidOdat (approved):
Looks great! Just some small comments :)

boa_engine/src/builtins/array_buffer/mod.rs
raskad: I think yes if we do not want to clone `target_buffer` on return. As far as I see it `target_buffer_mut` borrows from `target_buffer` and the drop at the end of the function happens after the move for the retun of the value. For example this would be the error if we remove the drop:

```
error[E0505]: cannot move out of `target_buffer` because it is borrowed
   --> boa_engine/src/builtins/array_buffer/mod.rs:437:12
    |
398 |         let target_buffer = Self::allocate(
    |             ------------- binding `target_buffer` declared here
...
417 |         let mut target_buffer_mut = target_buffer.borrow_mut();
    |                                     -------------------------- borrow of `target_buffer` occurs here
...
437 |         Ok(target_buffer)
    |            ^^^^^^^^^^^^^ move out of `target_buffer` occurs here
438 |     }
    |     - borrow might be used here, when `target_buffer_mut` is dropped and runs the `Drop` code for type `GcRefMut`
```

jedel1043 (approved):
Nice fix!

PR #3172, opened 2023-07-25
Fix a Parser Idempotency Issue

### LINKED ISSUE #3133: Idempotency issue: braces added every time after passing through the formatter and linter
**Describe the bug**
Starting and ending braces are added within an empty code block.

**To Reproduce**

This javascript 
```js
with (this) {{
}}
```
is transformed to the following, after passing through the parser, formatter and linter:
```js
with (this) {{
    {
    }
}}
```
And then to below, after another pass through:
```js
with (this) {{
    {
        {
        }
    }
}}
```

**Expected behavior**
Multiple runs through the parser -> formatter -> linter should result in the same output. 

- OS: Ubuntu
- Version: 20.04
- Rustc version: rustc 1.70.0

**Additional context**
Found by [#4822](https://github.com/astral-sh/ruff/pull/4822)

veera-sivarajan:
This PR fixes #3133 by correctly implementing `ToIndentedString` for `with` statement.

This also replaces the existing `ToInternedString` implementation because it is implemented for all types implementing `ToIndentedString` in `boa_interner/src/lib.rs`.


raskad:
Nice fix!
@veera-sivarajan do you think it makes sense to add a test for this? I think this was only discovered via fuzzing right?

veera-sivarajan:
Yeah, this was discovered by the fuzzer but I'm not sure if it requires an additional test because the additional braces were added by how the `with` statement was formatted (`{{}}` instead of `{}`) in `to_interned_string()` which, as far as my `rust-analyzer` could parse, is called only by the fuzzer?

jedel1043:
Waiting on adding the required test to merge this.

jedel1043 (approved):
Thanks for the contribution!

HalidOdat (approved):

PR #3059, opened 2023-06-22
Refactor environment, exception handling and jumping in VM

HalidOdat:
~~Depends on #3053~~

Fixes #2424 

This PR refactors environment handling and jumping with break, continues, etc in the VM. Currently the opcodes are very complex, they dynamically search for the environments, for this we also need to keep a lot of state in the `CallFrame`, this has a performance penalty as well as being hard to implement optimizers (since it has to mirror the state, so it does not change behaviour).

Since this the bytecompiler has full knowledge of where it needs to jump, how many environments to pop, etc. This PR aims to move that opcode logic to bytecompiler, generating opcodes that are "dumber"/simple, hence easier to optimize ( using #3037 ).

~~This still needs a lot of work!~~

It changes the following:

- Removes opcodes: `FinallyStart`, `FinallyEnd`, `Break`, `Continue`, `LoopStart`, `LoopEnd`, `LabelledStart`, `LabelledEnd`, `TryStart`, `TryEnd`, `IteratorLoopStart`,`GeneratorResumeReturn`, `GeneratorSetReturn`, `GeneratorJumpOnResumeKind`
- Add opcodes: `ReThrow`, `Exception`, `JumpTable`, `JumpIfNotResumeKind`, `IteratorNextWithoutPop`, `IteratorValueWithoutPop`
- Replaces `Break`s and `Continue`s with simple unconditional `Jump`.
- Remove `EnvEntry` type and env entry stack
- Remove `abrupt_completion` record from `CallFrame`
- Remove `generator_resume_kind` from `CallFrame`
- Remove `yield` field from `CallFrame`, added `CompletionType::Yield`

TODO:

- [x] Figure out `fp` on exception handlers
- [x] Fix bytecode flowgraph creation
- [x] Add more documentation.

PR #3179, opened 2023-07-27
Refactor `Context::run()` method

HalidOdat:
~~Depends on #3059~~

This simplifies `Context::run()` and `JsObject::call_internal()` by taking some logic that is in specific types of functions (async generator or async function) and putting it in separate opcodes. This also has a performance benefit, since we only do those checks if we are using those features.

This PR is related to refactoring calling in VM.

It changes the following:

- Adds opcodes: `Generator`, `AsyncGeneratorClose`, `CreatePromiseCapability` and `CompletePromiseCapability`
- Makes `Context::run()` a simple loop that runs the opcodes, moves some logic to separate opcodes (resolving and rejecting promise capability and async generator close to separate opcode)
- Moves `JsObject::call_internal()` logic to separate opcodes (generator creation and returning promise)
- Improves trace of stack, displays the full stack per instruction execution, instead of just top of stack.

```text
>> (++i) + (i++)

------------------------------------ Call Frame -- <main> ------------------------------------
Time          Opcode                     Operands                   Stack

7μs           GetName                    0000: 'i'                  [ 0 ]
1μs           Inc                                                   [ 1 ]
0μs           Dup                                                   [ 1, 1 ]
3μs           SetName                    0000: 'i'                  [ 1 ]
1μs           GetName                    0000: 'i'                  [ 1, 1 ]
0μs           IncPost                                               [ 1, 2, 1 ]
0μs           Swap                                                  [ 2, 1, 1 ]
1μs           SetName                    0000: 'i'                  [ 1, 1 ]
1μs           Add                                                   [ 2 ]
0μs           SetReturnValue                                        [ ]
0μs           Return                                                [ ]
```


boa_engine/src/vm/opcode/generator/mod.rs
```diff
@@ -1,20 +1,172 @@
...
+        let pc = context.vm.frame().pc;
+        let mut dummy_call_frame = CallFrame::new(code_block);
+        dummy_call_frame.pc = pc;
+        let call_frame = std::mem::replace(context.vm.frame_mut(), dummy_call_frame);
+
+        let this_function_object = context
+            .vm
+            .active_function
+            .clone()
+            .expect("active functoion should be set to the generator");
```
raskad, line 50:
```suggestion
            .expect("active function should be set to the generator");
```

boa_engine/src/vm/opcode/generator/mod.rs
```diff
@@ -1,20 +1,172 @@
...
 };
 
 pub(crate) use yield_stm::*;
 
 use super::SetReturnValue;
 
+/// `Generator` implements the Opcode Operation for `Opcode::Generator`
+///
+/// Operation:
+///  - Creates the generator generator object and yields.
```
raskad, line 29: Is the double generator intended?

```suggestion
///  - Creates the generator object and yields.
```

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -1326,6 +1323,20 @@ generate_impl! {
...
         Return,
 
+        /// Close an async generator function.
+        ///
+        /// Operands:
+        ///
+        /// Stack: **=>**
+        AsyncGeneratorClose,
+
+        /// Creates the generator generator object and yields.
```
raskad, line 1333: Same as boa_engine/src/vm/opcode/generator/mod.rs

```suggestion
        /// Creates the generator object and yields.
```

boa_engine/src/environments/runtime/mod.rs
```diff
@@ -78,6 +78,18 @@ impl EnvironmentStack {
...
     }
 
+    /// Gets the current global environment.
+    pub(crate) fn global(&self) -> Gc<DeclarativeEnvironment> {
+        let env = self.stack[0].clone();
+
+        match env {
+            Environment::Declarative(ref env) => env.clone(),
+            Environment::Object(_) => {
+                unreachable!("first environment should be th global environment")
```
raskad, line 88:
```suggestion
                unreachable!("first environment should be the global environment")
```

raskad (approved):
Nice changes. I just found a few typos.

boa_engine/src/vm/mod.rs
```diff
@@ -209,260 +297,80 @@ impl Context<'_> {
...
-        self.vm
-            .frame_mut()
-            .set_frame_pointer(current_stack_length as u32);
-
-        // If the current executing function is an async function we have to resolve/reject it's promise at the end.
-        // The relevant spec section is 3. in [AsyncBlockStart](https://tc39.es/ecma262/#sec-asyncblockstart).
-        let promise_capability = self.vm.frame().promise_capability.clone();
-
-        let execution_completion = loop {
+        let mut result = Ok(CompletionType::Normal);
```
nekevss, line 307: Is there a specific benefit here to using the mutable result value vs. a scoped value?

nekevss (commented):
Hey! This looks really good for the most part (especially with that other PR re calling the function with the stack🔥).

I just had one question but it could just be a stylistic/approach difference at it's core. Still figure I'd ask though 😄.

boa_engine/src/vm/mod.rs
HalidOdat: > Is there a specific benefit here to using the mutable result value vs. a scoped value?

This was done because if we throw a `JsResult` error then we have to call `ReThrow::evaluate()` then we have to do the same thing we do when we execute an instruction, so instead of duplicate code, we just set result and continue to handle it.

HalidOdat: I inlined `ReThrow` since it's small and a lot of the logic is in `handle_exception_at`, and allows us to remove a check for generator return too, made `result` immutable :)

nekevss (approved):
Nice work on this!

PR #3180, opened 2023-07-28
Re-enable must_use clippy rule

tunz:
Following up to re-enable `must_use` clippy rule for boa_engine which was disabled in #3109.

I gave an exception to `set_prototype` because it returns false when it's not extensible but there are some code already checking if it's extensible, so they don't need to check the return value, and it looks like that's a common pattern.


HalidOdat (approved):
Looks perfect to me :)

jedel1043 (approved):

PR #3185, opened 2023-07-29
Use main stack for calling ordinary functions

HalidOdat:
~~Depends on #3179~~

Part of refactoring calling in VM, makes calling ordinary functions use the main stack, also removes return value from `CallFrame` and places the return value on the stack after the frame pointer.

### DIFF
--- a/boa_engine/src/bytecompiler/function.rs
+++ b/boa_engine/src/bytecompiler/function.rs
@@ -136,6 +136,9 @@ impl FunctionCompiler {
             // ExecuteAsyncModule ( module ): <https://tc39.es/ecma262/#sec-execute-async-module>
             compiler.emit_opcode(Opcode::CreatePromiseCapability);
 
+            // Note: We set it to one so we don't pop return value when we return.
+            compiler.current_stack_value_count += 1;
+
             // 2. Let declResult be Completion(FunctionDeclarationInstantiation(functionObject, argumentsList)).
             //
             // Note: We push an exception handler so we catch exceptions that are thrown by the
@@ -159,6 +162,10 @@ impl FunctionCompiler {
         // Note: We do handle exceptions thrown by generator body in `AsyncGeneratorStart`.
         if compiler.in_generator() {
             assert!(compiler.async_handler.is_none());
+
+            // Note: We set it to one so we don't pop return value when we return.
+            compiler.current_stack_value_count += 1;
+
             if compiler.in_async() {
                 // Patched in `ByteCompiler::finish()`.
                 compiler.async_handler = Some(compiler.push_handler());
--- a/boa_engine/src/bytecompiler/function.rs
+++ b/boa_engine/src/bytecompiler/function.rs
@@ -136,9 +136,6 @@ impl FunctionCompiler {
             // ExecuteAsyncModule ( module ): <https://tc39.es/ecma262/#sec-execute-async-module>
             compiler.emit_opcode(Opcode::CreatePromiseCapability);
 
-            // Note: We set it to one so we don't pop return value when we return.
-            compiler.current_stack_value_count += 1;
-
             // 2. Let declResult be Completion(FunctionDeclarationInstantiation(functionObject, argumentsList)).
             //
             // Note: We push an exception handler so we catch exceptions that are thrown by the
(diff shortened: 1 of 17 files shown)


boa_engine/src/vm/code_block.rs
```diff
@@ -1256,14 +1241,22 @@ impl JsObject {
 
         context.vm.push_frame(frame);
 
+        // Push return value.
+        context.vm.push(JsValue::undefined());
```
nekevss, line 1245: Would it make sense to tie pushing the return/completion value to pushing the frame onto the stack?

```rust
pub(crate) fn push_frame(&mut self, mut frame: CallFrame) {
    let current_stack_length = self.stack.len();
    frame.set_frame_pointer(current_stack_length as u32);

    self.push(JsValue::undefined());
    self.frames.push(frame);
}
```

boa_engine/src/vm/mod.rs
```diff
@@ -181,6 +181,16 @@ impl Vm {
 
         true
     }
+
+    pub(crate) fn get_return_value(&self) -> JsValue {
```
nekevss, line 185: Maybe rename `return_value` to  `completion_value`. That would align with the spec a bit more, right? Although, I'm sort of thinking this as long term it would be ideal to store the result/error value of the frame in `fp`, and that may not be the best way to approach it.

nekevss (commented):
This looks pretty great! Couple comments/questions 😄

boa_engine/src/vm/code_block.rs
HalidOdat: > Would it make sense to tie pushing the return/completion value to pushing the frame onto the stack?

Yes, thought about this, for normal function this is fine, the issue is with generators since we need to restore without adding anything.

One solution would be to have two `push_frame` functions, or maybe extract the parts into a `prepare_call_frame()`.

boa_engine/src/vm/mod.rs
HalidOdat: > Maybe rename `return_value` to `completion_value`. That would align with the spec a bit 
more, right?

But the completion value can be an exception, when a [throw happens](https://tc39.es/ecma262/#sec-completion-record-specification-type) which `return_value` cannot be, so I think return_value is more appropriate.

> Although, I'm sort of thinking this as long term it would be ideal to store the result/error value of the frame in `fp`, and that may not be the best way to approach it.

In the latest commit I moved `return_value` to `Vm` instead of per function (call frame). 

This is simular to what v8 does with the `accumulator`, implicit return value. It may be a good idea to switch to a register-based VM at some point :)

nekevss: Sorry, I may have a bit unclear on the previous comment. I think the previous approach with the value on the stack works. I was more thinking it would be great to update `JsError`s conversion to and from `JsValue`, so instead of a `pending_exception` field and a `return_value` field, we have one field (a return value or completion value, however we want to name it) on the stack. We don't have a really good way currently to preserve `JsNativeError` currently though after converting it to an opaque, which was more so what I was thinking about when I had made the comment.

> It may be a good idea to switch to a register-based VM at some point :)

I was thinking the same thing :) But also see above

boa_engine/src/bytecompiler/jump_control.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) enum JumpRecordAction {
 pub(crate) enum JumpRecordKind {
     Break,
     Continue,
-    Return,
+    Return { return_value_on_stack: bool },
```
Razican, line 69: Wouldn't this increase the size of the enum? It might be better to have a new variant

HalidOdat: [I doesn't](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7ab6df9c9c5f8852b90bd2ba26bf33c1), since a bool only needs a bit it can probably sore it in the discriminant, but even if would increase the size, I think it would be fine since we only have one per `JumpRecord`.

boa_engine/src/bytecompiler/function.rs
```diff
@@ -136,6 +136,9 @@ impl FunctionCompiler {
             // ExecuteAsyncModule ( module ): <https://tc39.es/ecma262/#sec-execute-async-module>
             compiler.emit_opcode(Opcode::CreatePromiseCapability);
 
+            // Note: We set it to one so we don't pop return value when we return.
+            compiler.current_stack_value_count += 1;
```
raskad, line 140: Are these still needed, now that `return_value` in on the `Vm`?

HalidOdat: > Are these still needed, now that `return_value` in on the `Vm`?

There was a bug that was causing a panic with some weird edge case with array destructuring with yield, this panic was fixed in the latest commit and removed `compiler.current_stack_value_count += 1;` :)

raskad (approved):
Looks good :)

nekevss (approved):

PR #3188, opened 2023-07-31
Improve bytecompiler bytecode generation.

HalidOdat:
This PR adds the `Operand` enum (which represents an instruction operand), and makes `ByteCompiler::emit` take an array of `Operand`s instead of `u32` which allows to emit varying length operands more easily.


raskad:
I like the refactor, but there seems to be a bug somewhere since -1,820 tests are failing

raskad (approved):

jedel1043 (approved):
Nice refactor!

PR #3193, opened 2023-08-01
Log traces even without message (boa_runtime)

kelbazz:
This Pull Request fixes boa_runtime console.trace not printing the trace if no argument is given.

It changes the following:
- Change the if statement scope in the trace function of Console


jedel1043 (approved):

Razican (approved):
Thanks!

PR #3194, opened 2023-08-01
Optimization fast local variables

HalidOdat:
Depends on #3185 

This PR is a WIP, it implements an optimization on function local variables, functions that do not call `eval` directly or variables that are not escaped, we can keep then on the stack after frame pointer and use and offset to index them (`index = fp + offset`). This allows us to eliminate the creation of some of the gc collected environments. Which gives us **huge** performance improvements!

#### Main

```text
PROGRESS Richards
RESULT Richards 36.0
PROGRESS DeltaBlue
RESULT DeltaBlue 42.2
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 63.2
PROGRESS RayTrace
RESULT RayTrace 147
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 136
undefined
PROGRESS RegExp
PROGRESS Splay
RESULT Splay 118
PROGRESS NavierStokes
RESULT NavierStokes 136
SCORE 84.6
```

#### PR

```
PROGRESS Richards
RESULT Richards 42.3
PROGRESS DeltaBlue
RESULT DeltaBlue 46.5
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 102
PROGRESS RayTrace
RESULT RayTrace 164
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 158
undefined
PROGRESS RegExp
PROGRESS Splay
RESULT Splay 124
PROGRESS NavierStokes
RESULT NavierStokes 246
SCORE 107
```

Almost 2x improvment on NavierStokes :)

### DIFF
--- a/boa_engine/src/bytecompiler/declarations.rs
+++ b/boa_engine/src/bytecompiler/declarations.rs
@@ -912,6 +912,14 @@ impl ByteCompiler<'_, '_> {
             }
         }
 
+        if self.can_optimize_local_variables
+            && !self
+                .flags
+                .contains(super::ByteCompilerFlags::USES_ARGUMENTS)
+        {
+            arguments_object_needed = false;
+        }
+
         // 19. If strict is true or hasParameterExpressions is false, then
         //     a. NOTE: Only a single Environment Record is needed for the parameters,
         //        since calls to eval in strict mode code cannot create new bindings which are visible outside of the eval.
--- a/boa_engine/src/bytecompiler/declarations.rs
+++ b/boa_engine/src/bytecompiler/declarations.rs
@@ -570,9 +570,13 @@ impl ByteCompiler<'_, '_> {
 
                                     // ii. Perform ! varEnv.InitializeBinding(F, undefined).
                                     let binding = self.initialize_mutable_binding(f, true);
-                                    let index = self.get_or_insert_binding(binding);
                                     self.emit_opcode(Opcode::PushUndefined);
-                                    self.emit_with_varying_operand(Opcode::DefInitVar, index);
+                                    self.get_or_insert_binding(binding).emit(
+                                        Opcode::SetLocal,
+                                        Opcode::SetGlobalName,
+                                        Opcode::DefInitVar,
+                                        self,
+                                    );
                                 }
                             }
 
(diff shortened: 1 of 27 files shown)


boa_engine/src/bytecompiler/declarations.rs
```diff
@@ -1055,15 +1086,27 @@ impl ByteCompiler<'_, '_> {
...
-                        let index = self.get_or_insert_binding(binding);
-                        self.emit(Opcode::GetName, &[Operand::U32(index)]);
+                        match self.get_or_insert_binding(binding) {
+                            EnvironmentAccess::Fast { index } => {
+                                self.emit(Opcode::GetLocal, &[Operand::U32(index)]);
+                            }
+                            EnvironmentAccess::Slow { index } => {
+                                self.emit(Opcode::GetName, &[Operand::U32(index)]);
+                            }
+                        }
```
jedel1043, line 1096: I'm seeing this snippet several times. Should this be extracted into its own method?

HalidOdat: Will do this once I have finished the optimization, currently there are some bugs which are making some test test22 fails

jedel1043:
Blocked on #3490

raskad:
@HalidOdat I would like to take a look at continuing development on local variables. We talked about having environment aware bytecode generation for this with analyzed variable scopes. Do you have any more work or ideas on this or should I just start from the current state of this PR? Also, should I keep anything in mind regarding #3798?

HalidOdat:
> Do you have any more work or ideas on this or should I just start from the current state of this PR?

The issue I have with the current implementation is that it does not account for many cases (and it's very hacked together).

My idea for a more proper solution would be to move the compile time environment logic to the parser and have the AST nodes like `*Function*`, `Block` contain a compile time environment, [LibJS](https://github.com/LadybirdBrowser/ladybird/blob/master/Userland/Libraries/LibJS/Parser.cpp) does this. It removes the need for a second pass after parsing (though makes parsing slower if we want to fail fast, but most scripts are valid so I don't think this is an issue).

The second idea was to perform a second pass after the AST generation this would generate a data structure that contains all the scope data that is passed to the compiler, and add a ID to the AST nodes that is `None` initially and filled in with the scope it points to, making mapping back to the scopes more efficient. This is a simpler approach and I was experimenting with it.

I began working on this but didn't make much progress, you can find the code on my fork (hope it helps): <https://github.com/HalidOdat/boa/tree/scope-analysis>

> Also, should I keep anything in mind regarding #3798?

Hmmm.. The logic on how to allocate register/local variables in the compiler will be shared, besides that I think both tasks can be worked on simultaneously.

I'll create a PR and extract that logic so we can merge it into `main`.

raskad:
Quick update on this. I have a working version with scope analysis at parse time. All tests pass and local variables work. I'm now in the process of cleaning up the implementation and removing some hacks.
The benchmark score increases are similar to the ones in the PR here. Overall score + ~23% for my first benchmarks.

I ran into the issue, of needing to track uninitialized variables, since we cannot track if a variable in initialized just with the stack value. Right now I'm just using a call frame local map to track this, but I think I can reduce that to an array to get some better performance.

jedel1043:
@raskad Tracking uninit variables may be something that is solved by #3037 IIRC

raskad:
Superseded by #3988

PR #3197, opened 2023-08-03
Store active runnable and active function in `CallFrame`

HalidOdat:
~~Depends on #3185~~

Part of refactoring calling in VM. In this PR active runnable and active function are moved in the `CallFrame`, this allows us to untangle some of the function calling.

It may seem like we are using more memory per ordinary function call, because of CallFrame size increase, but it actuality it's the same, before instead of storing it in the call frame we would spill it on the native rust stack which was filling the stack, contributing to smaller recursion limits.

@nekevss As I see it we already have the execution context stack, that being the `CallFrame`s (just a different name), we just have some work moving some fields that are supposed to be in  an execution context like `[[ScriptOrModule]]` which this PR does, realm needs to be moved too.

### DIFF
--- a/boa_engine/src/builtins/array/mod.rs
+++ b/boa_engine/src/builtins/array/mod.rs
@@ -174,9 +174,7 @@ impl BuiltInConstructor for Array {
         // If NewTarget is undefined, let newTarget be the active function object; else let newTarget be NewTarget.
         let new_target = &if new_target.is_undefined() {
             context
-                .vm
-                .active_function
-                .clone()
+                .active_function_object()
                 .unwrap_or_else(|| context.intrinsics().constructors().array().constructor())
                 .into()
         } else {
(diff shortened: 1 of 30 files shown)


nekevss:
I still have to take a look at the PR in a bit more detail, but I thought I'd bring it up just a small discussion point early, especially since it may have been obvious that it was coming 😆

Using `CallFrame` as an the execution stack could definitely work. It would potentially simplify things as storing everything in a singular place, but would there be any value in adding an `execution_stack` on the vm that takes a record (e.g. `Vec<ExecutionContext>`) more aligned with the [spec](https://tc39.es/ecma262/#sec-execution-contexts). This would mean also not using `swap` on calls I think. Again, that could be the wrong approach, but thought I'd bring it up nonetheless.

HalidOdat:
> Using `CallFrame` as an the execution stack could definitely work. It would potentially simplify things as storing everything in a singular place, but would there be any value in adding an `execution_stack` on the vm that takes a record (e.g. `Vec<ExecutionContext>`) more aligned with the [spec](https://tc39.es/ecma262/#sec-execution-contexts). This would mean also not using `swap` on calls I think.

In my opinion, I don't see any advantages in doing this, We already have `Vec<CallFrame>` the only difference is the name, which we can change.

nekevss (commented):
A lot of this looks pretty good to me. Just one small thing I noticed 😄

nekevss (approved):
Looks good to me 😄

raskad (approved):

PR #3198, opened 2023-08-03
Push float opcode

HalidOdat:
Depends on #3188 

This PR adds a `PushFloat` opcode that takes a `f32` value, we already have a `PushRational` that pushes a `f64` value. The reason for this additions is to reduce wasted memory usage since we can represent many `f64` in `f32` value (like `0.5`), and conversion from `f32` to `f64` is lossless.

### DIFF
--- a/boa_engine/src/bytecompiler/mod.rs
+++ b/boa_engine/src/bytecompiler/mod.rs
@@ -551,8 +551,14 @@ impl<'ctx, 'host> ByteCompiler<'ctx, 'host> {
         if f64::from(value as i32).to_bits() == value.to_bits() {
             self.emit_push_integer(value as i32);
         } else {
-            self.emit_opcode(Opcode::PushRational);
-            self.emit_u64(value.to_bits());
+            let f32_value = value as f32;
+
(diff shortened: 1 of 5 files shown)


Razican:
Do we have benchmarks? This could have some effect, right?

HalidOdat:
It does have a small effect :)

#### Main

```
Richards: 36.7
DeltaBlue: 44.0
RayTrace: 149
----
Score (version 7): 62.2
```

#### PR

```
Richards: 36.9
DeltaBlue: 44.3
RayTrace: 151
----
Score (version 7): 62.7
undefined
```

Razican (approved):
Cool change!

jedel1043 (approved):
Nice improvement

jedel1043:
Oops, thought this was based on main. Well, the change is small enough that it shouldn't be problematic to merge it alongside #3188, right?

HalidOdat:
> Oops, thought this was based on main. Well, the change is small enough that it shouldn't be problematic to merge it alongside #3188, right?

Yeah, it's fine :)

PR #3201, opened 2023-08-05
Add `Instruction` and `InstructionIterator`

HalidOdat:
This PR adds a fat enum that represents a full instruction (opcode + operands) and an iterator that takes bytecode and produces instructions.

I implemented something like this in #3037, this PR just extracts and refines it.

Why is this useful/needed?

 - It allows us to iterate over instructions without worrying, whether we are reading the operands correctly. Currently we do the same thing in `instruction_operands()` and flowgraph generation, this is duplicate code and can easily lead to mistakes (also doing it in #3037 )
 - Extracting the operands is harder when manually accessing bytecode.
 - If we have a list of instructions it's easier to do changes in it.

Note: This is not intended to replace the compact bytecode representation, that is optimized for execution and storage.

... This also makes it easier to implement an bytecode size optimization on operands, that I'm working on ;)

### DIFF
--- a/boa_engine/src/vm/opcode/mod.rs
+++ b/boa_engine/src/vm/opcode/mod.rs
@@ -1,5 +1,7 @@
+use std::iter::FusedIterator;
+
 /// The opcodes of the vm.
-use crate::{vm::CompletionType, Context, JsResult};
+use crate::{vm::CompletionType, Context, JsResult, JsValue};
 
 // Operation modules
 mod r#await;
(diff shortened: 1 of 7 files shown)


boa_engine/src/vm/opcode/mod.rs
```diff
@@ -79,23 +81,217 @@ pub(crate) use swap::*;
 pub(crate) use switch::*;
 #[doc(inline)]
 pub(crate) use templates::*;
+use thin_vec::ThinVec;
```
jedel1043, line 84: Seems weird to have this import between exports.

jedel1043 (approved):
Really nice improvement! It simplifies our logic a LOT.
Any plans for extending it to `Operation` implementers or is it just not worth?

HalidOdat:
> Any plans for extending it to `Operation` implementers or is it just not worth?

No, representation is intended for easier bytecode analysis, not really execution, the compact bytecode version is smaller and faster.

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -86,26 +88,218 @@ pub(crate) use unary_ops::*;
...
+        for _ in 0..count {
+            let item = read::<u32>(bytes, *pc);
+            *pc += std::mem::size_of::<u32>();
+            result.push(item);
+        }
+        result
+    }
+}
+
+macro_rules! generate_opcodes {
```
nekevss, line 281: Should we start writing implementation notes to break down macros?

jasonwilliams:
Whats the affect on performance with this change?

HalidOdat:
> Whats the affect on performance with this change?

There is not change in performance, this is doesn't change the way we execute bytecode. It only implements an iterator and the result being an `Instruction` (opcode + operands).

PR #3205, opened 2023-08-09
Fix switch statement `break` and `continue` return values

raskad:
This Pull Request fixes the remaining "language/statements/switch" tests with exception of the tco tests.

It changes the following:

This removes the current behaviour of setting the return value to `undefined` whenever there is a  `break` or `continue` statement at the start of a block. This was causing the return value to be unexpectedly overwritten in switch statements.
Instead this is now only implemented for `catch` and `finally` blocks where this behaviour mirrors the `UpdateEmpty` statements in the spec.

### DIFF
--- a/boa_engine/src/bytecompiler/statement/try.rs
+++ b/boa_engine/src/bytecompiler/statement/try.rs
@@ -5,7 +5,8 @@ use crate::{
 use boa_ast::{
     declaration::Binding,
     operations::bound_names,
-    statement::{Catch, Finally, Try},
+    statement::{Block, Catch, Finally, Try},
+    Statement, StatementListItem,
 };
 
 impl ByteCompiler<'_, '_> {
@@ -130,7 +131,7 @@ impl ByteCompiler<'_, '_> {
             self.emit_opcode(Opcode::Pop);
         }
 
-        self.compile_block(catch.block(), use_expr);
+        self.compile_catch_finally_block(catch.block(), use_expr);
 
         let env_index = self.pop_compile_environment();
         self.patch_jump_with_target(push_env, env_index);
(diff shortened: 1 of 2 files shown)


Razican (approved):
Looks good to me! Thanks :)

boa_engine/src/bytecompiler/statement/try.rs
```diff
@@ -162,4 +163,36 @@ impl ByteCompiler<'_, '_> {
...
+    ///
+    /// If the block contains a break or continue as the first statement,
+    /// the return value is set to undefined.
+    /// See the [ECMAScript reference][spec] for more information.
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-try-statement-runtime-semantics-evaluation
+    fn compile_catch_finally_block(&mut self, block: &Block, use_expr: bool) {
+        let mut b = block;
+
+        'outer: loop {
```
HalidOdat, line 177: there is only one loop in so i think we can remove `'outer` label, right?

boa_engine/src/bytecompiler/statement/try.rs
```diff
@@ -162,4 +163,36 @@ impl ByteCompiler<'_, '_> {
...
+                Some(StatementListItem::Statement(
+                    Statement::Break(_) | Statement::Continue(_),
+                )) => {
+                    self.emit_opcode(Opcode::PushUndefined);
+                    self.emit_opcode(Opcode::SetReturnValue);
+                    break 'outer;
+                }
+                Some(StatementListItem::Statement(Statement::Block(block))) => {
+                    b = block;
+                    continue 'outer;
```
HalidOdat, line 188: `continue` here is not needed, right?

HalidOdat (commented):
Nice change! Just some small nitpicks :)

boa_engine/src/bytecompiler/statement/try.rs
raskad: Fixed :)

HalidOdat (approved):
Looks perfect to me :)

PR #3214, opened 2023-08-14
Lazily download `test262`  repository

### LINKED ISSUE #3095: Consider removing tc39/test262 submodule and download lazily
Due to how cargo handles submodules (or rather doesn't): https://github.com/rust-lang/cargo/issues/4247

a fresh cargo build with boa as git dependency `boa { git = "..."}` will download all submodules and it takes some time to download https://github.com/tc39/test262

Perhaps it's possible to download the repo lazily instead, for example:

https://github.com/dtolnay/syn/pull/183

HalidOdat:
This Pull Request fixes/closes #3095 .

This PR makes the downloads the `test262` repository, if we run the `boa_tester` with `run` command and we do not specify the test262 directory.

There is a con with this approach, dependabot can't suggest a updates. We could implement our own checker though.

TODO:

- [x] The commit is hard coded, add a config file for test262 default commit (maybe merge `test_ignore.toml` into it).
  - [x] Add ability to specify commit on test262 clone, via CLI commands
  - [x] Warn user that the current commit is older than test262 HEAD
- [x] Better error reporting

### DIFF
--- a/boa_tester/src/main.rs
+++ b/boa_tester/src/main.rs
@@ -255,10 +255,10 @@ fn main() -> Result<()> {
                 toml::from_str(&input).wrap_err("could not decode tester config file")?
             };
 
-            let test262_commit = match (test262_commit.as_deref(), config.commit()) {
-                (Some("latest"), _) | (None, "" | "latest") => None,
-                (Some(commit), _) | (None, commit) => Some(commit),
-            };
+            let test262_commit = test262_commit
+                .as_deref()
+                .or_else(|| Some(config.commit()))
+                .filter(|s| !["", "latest"].contains(s));
 
             let test262_path = if let Some(path) = test262_path.as_deref() {
                 path
--- a/boa_tester/src/main.rs
+++ b/boa_tester/src/main.rs
@@ -91,12 +91,29 @@ use serde::{
     Deserialize, Deserializer, Serialize,
 };
 use std::{
-    fs::{self, File},
-    io::Read,
     ops::{Add, AddAssign},
     path::{Path, PathBuf},
     process::Command,
 };
+
+/// Structure that contains the configuration of the tester.
+#[derive(Debug, Deserialize)]
+struct Config {
+    commit: String,
+    ignored: Ignored,
+}
+
+impl Config {
+    /// Get the `test262` repository commit.
+    pub(crate) fn commit(&self) -> &str {
+        &self.commit
+    }
+
+    /// Get [`Ignored`] `test262` tests and features.
+    pub(crate) const fn ignored(&self) -> &Ignored {
+        &self.ignored
+    }
+}
 
 /// Structure to allow defining ignored tests, features and files that should
 /// be ignored even when reading.
@@ -161,18 +178,13 @@ enum Cli {
         #[arg(
             long,
             value_hint = ValueHint::DirPath,
-            conflicts_with = "test262_update",
             conflicts_with = "test262_commit"
         )]
         test262_path: Option<PathBuf>,
 
-        /// Specify the Test262 commit when cloning the repository.
-        #[arg(long, conflicts_with = "test262_update")]
-        test262_commit: Option<String>,
-
-        /// Update Test262 commit of the repository.
+        /// Specify the Test262 commit when cloning the repository. To checkout to the latest commit set to "latest".
         #[arg(long)]
-        test262_update: bool,
+        test262_commit: Option<String>,
 
         /// Which specific test or test suite to run. Should be a path relative to the Test262 directory: e.g. "test/language/types/number"
         #[arg(short, long, default_value = "test", value_hint = ValueHint::AnyPath)]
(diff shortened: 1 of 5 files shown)


boa_tester/src/main.rs
```diff
@@ -241,12 +241,45 @@ fn main() -> Result<()> {
...
+                    .expect("");
+
+    println!("result: {result}");
+    assert!(result.success());
+
+    println!("reset to known-good rev");
+    let result = Command::new("git")
+                    .arg("reset")
+                    .arg("--hard")
+                    .arg(TEST262_COMMIT)
```
Razican, line 268: We could just default to the latest commit in `main`, and provide a commit from the CLI. We could then hard-code this in our CI if needed.

jasonwilliams: agree, it seems odd to choose an arbitary commit. Why did we choose this one? Was it just the one we're using at the moment?

 I guess CI would have some environment variable to swap "main" out with whatever we're using.

jasonwilliams:
Good idea, I don't see there being an issue with this. It should make the whole git submodules issue easier

boa_tester/src/main.rs
gakonst: Agreed - seems fine to default to "main" and override via CLI which sources the value from CI or other source the user controls.

gakonst (commented):
Love it!! Thank you for taking this on, big devex upgrade for anyone using boa as a dep and doesn't want their clone to take minutes ^^

HalidOdat:
This is now complete, and ready for review/merge :)

Now tester warns if local test262 clone is out of sync with `origin/main`, Added `--test262-update` for automatically updating the local repository and `--test262-commit` for specifying a commit.

jedel1043:
Did you consider using [`git2`](https://docs.rs/git2/latest/git2/) instead of calling manual commands? Maybe that could further simplify the repo handling logic.

gakonst:
Hi folks - wonder if we (we = [Reth](https://github.com/paradigmxyz/reth/) dev team) can help with this PR in any way?

jedel1043 (changes requested):
This looks great implementation-wise, but I'd argue we should have a way to pin the commit downloaded by the tool for reproducibility in our CI. Maybe a file that lives in the root of the repo?

HalidOdat:
> This looks great implementation-wise, but I'd argue we should have a way to pin the commit downloaded by the tool for reproducibility in our CI. Maybe a file that lives in the root of the repo?

Maybe merge `test_ignore.toml` file into something like `test262_config.toml`, to avoid having more than one config file `boa_tester`

jedel1043:
> Maybe merge `test_ignore.toml` file into something like `test262_config.toml`, to avoid having more than one config file `boa_tester`

Great idea! We just have to create a new `ignore` section and we can put a `commit` property on the root

jasonwilliams:
Hey @gakonst what does Reth use Boa for? I’m curious

HalidOdat:
> > Maybe merge `test_ignore.toml` file into something like `test262_config.toml`, to avoid having more than one config file `boa_tester`
> 
> Great idea! We just have to create a new `ignore` section and we can put a `commit` property on the root

Implemented in the latest commit https://github.com/boa-dev/boa/pull/3214/commits/3a0d75cb799f96bb0153995569d0ffdedf569316 :)

gakonst:
> Hey @gakonst what does Reth use Boa for? I’m curious

Hi Jason, Reth (w/o getting in too many of the details around the whole thing) exposes a debugging JSON-RPC service to provide "execution traces" of the request, showing step by step what happened. These traces can be configured to run with certain "tracers", some of which are part of the node. However, there's benefit in allowing users to specify custom tracers. The way this was implemented originally in go-ethereum was by embedding a Javascript sandbox and on various parts of the tracer lifecycle to call to some client-side provided JS that augments the tracer output.

In Reth, we use boa to provide that JS sandbox to the tracer. See [Go-ethereum's docs](https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers#Javascript%20tracers) for more on this.

Hope that's helpful, happy to provide other context if needed but do not want to derail this issue.

boa_tester/src/main.rs
```diff
@@ -204,35 +232,58 @@ enum Cli {
...
+        } => {
+            let config: Config = {
+                let input = std::fs::read_to_string(config_path)?;
+                toml::from_str(&input).wrap_err("could not decode tester config file")?
+            };
+
+            let test262_commit = match (test262_commit.as_deref(), config.commit()) {
+                (Some("latest"), _) | (None, "" | "latest") => None,
+                (Some(commit), _) | (None, commit) => Some(commit),
+            };
```
jedel1043, line 261:
```suggestion
            let test262_commit = test262_commit
                .as_deref()
                .or_else(|| Some(config.commit()))
                .filter(|s| !["", "latest"].contains(s));
```

jedel1043 (approved):
Nice work! I just have a small nitpick that doesn't block merging this :)

nekevss (approved):
looks good to me!

PR #3227, opened 2023-08-19
Add `Vm` trace functionality and add trace to `boa_wasm`

nekevss:
This is the branch that was used for the `web-debugger` concept.

This branch is technically ready to be merged, but thought I'd post it as a draft for now to get feedback as it has some API changes along with quite a few changes to the `Vm` trace overall.

It changes the following:

- Search and handle the compiled bytecode trace prior to instruction execution and trace. 
- Creates a `BoaJs` object and an `evaluate_with_debug_hooks` function in `boa_wasm`
- Creates a `VmTrace` (not the biggest fan of the name here if anyone has other ideas) for handling tracing and allowing users to define handle trace themselves via "registering" a function (defaults to println).

For reference, the standard trace output would now print like the below.

```
Location  Count    Handler    Opcode                     Operands

000000    0000      none      GetName                    0000: 'greet'
000005    0001      none      PushUndefined              
000006    0002      none      Swap                       
000007    0003      none      PushLiteral                0
000012    0004      none      Call                       1
000017    0005      none      SetReturnValue             
000018    0006      none      Return                     

Literals:
    0000: <string> "World"

Bindings:
    0000: greet

Functions:
    0000: name: 'greet' (length: 1)

Handlers:
    <empty>


-----------------------Compiled Output: 'greet'-----------------------
Location  Count    Handler    Opcode                     Operands

000000    0000      none      PutLexicalValue            0000: 'targetName'
000005    0001      none      RestParameterPop           
000006    0002      none      PushLiteral                0
000011    0003      none      GetName                    0000: 'targetName'
000016    0004      none      Add                        
000017    0005      none      PushLiteral                1
000022    0006      none      Add                        
000023    0007      none      SetReturnValue             
000024    0008      none      Return                     
000025    0009      none      Return                     

Literals:
    0000: <string> "Hello, "
    0001: <string> "!"

Bindings:
    0000: targetName

Functions:
    <empty>

Handlers:
    <empty>

------------------------------------ Call Frame -- <main> ------------------------------------
Time          Opcode                     Operands                   Stack

14╬╝s          GetName                    0000: 'greet'              [ [function] ]
0╬╝s           PushUndefined                                         [ undefined, [function] ]
0╬╝s           Swap                                                  [ [function], undefined ]
1╬╝s           PushLiteral                0                          [ "World", [function], undefined ]
------------------------------------ Call Frame -- greet -------------------------------------
0╬╝s           PutLexicalValue            0000: 'targetName'         [ ]
1╬╝s           RestParameterPop                                      [ ]
0╬╝s           PushLiteral                0                          [ "Hello, " ]
1╬╝s           GetName                    0000: 'targetName'         [ "World", "Hello, " ]
5╬╝s           Add                                                   [ "Hello, World" ]
0╬╝s           PushLiteral                1                          [ "!", "Hello, World" ]
1╬╝s           Add                                                   [ "Hello, World!" ]
0╬╝s           SetReturnValue                                        [ ]
0╬╝s           Return                                                [ ]
-------------------------- Call Frame -- <Exiting greet via Return> --------------------------
140╬╝s         Call                       1                          [ "Hello, World!" ]
0╬╝s           SetReturnValue                                        [ ]
0╬╝s           Return                                                [ ]
------------------------- Call Frame -- <Exiting <main> via Return> --------------------------
"Hello, World!"

```

Let me know what you think 😄

### DIFF
--- a/boa_engine/src/vm/mod.rs
+++ b/boa_engine/src/vm/mod.rs
@@ -18,14 +18,14 @@ use boa_profiler::Profiler;
 use std::mem::size_of;
 
 #[cfg(feature = "trace")]
-use boa_interner::ToInternedString;
-#[cfg(feature = "trace")]
 use std::time::Instant;
 
 mod call_frame;
 mod code_block;
 mod completion_record;
 mod opcode;
+// TODO: #[cfg(feature = "trace")]
+mod vm_trace;
 
 mod runtime_limits;
 
@@ -33,6 +33,8 @@ mod runtime_limits;
 pub mod flowgraph;
 
 pub use runtime_limits::RuntimeLimits;
+
+use self::vm_trace::VmTrace;
 pub use {call_frame::CallFrame, code_block::CodeBlock, opcode::Opcode};
 
 pub(crate) use {
@@ -70,7 +72,7 @@ pub struct Vm {
     pub(crate) active_runnable: Option<ActiveRunnable>,
 
     #[cfg(feature = "trace")]
-    pub(crate) trace: bool,
+    pub(crate) trace: Option<VmTrace>,
 }
 
 /// Active runnable in the current vm context.
@@ -102,7 +104,7 @@ impl Vm {
             active_function: None,
             active_runnable: None,
             #[cfg(feature = "trace")]
-            trace: false,
+            trace: None,
         }
     }
 
@@ -203,44 +205,6 @@ pub(crate) enum CompletionType {
 
 #[cfg(feature = "trace")]
 impl Context<'_> {
-    const COLUMN_WIDTH: usize = 26;
-    const TIME_COLUMN_WIDTH: usize = Self::COLUMN_WIDTH / 2;
-    const OPCODE_COLUMN_WIDTH: usize = Self::COLUMN_WIDTH;
-    const OPERAND_COLUMN_WIDTH: usize = Self::COLUMN_WIDTH;
-    const NUMBER_OF_COLUMNS: usize = 4;
-
-    fn trace_call_frame(&self) {
-        let msg = if self.vm.frames.last().is_some() {
-            format!(
-                " Call Frame -- {} ",
-                self.vm.frame().code_block().name().to_std_string_escaped()
-            )
-        } else {
-            " VM Start ".to_string()
-        };
-
-        println!(
-            "{}",
-            self.vm
-                .frame()
-                .code_block
-                .to_interned_string(self.interner())
-        );
-        println!(
-            "{msg:-^width$}",
-            width = Self::COLUMN_WIDTH * Self::NUMBER_OF_COLUMNS - 10
-        );
-        println!(
-            "{:<TIME_COLUMN_WIDTH$} {:<OPCODE_COLUMN_WIDTH$} {:<OPERAND_COLUMN_WIDTH$} Stack\n",
-            "Time",
-            "Opcode",
-            "Operands",
-            TIME_COLUMN_WIDTH = Self::TIME_COLUMN_WIDTH,
-            OPCODE_COLUMN_WIDTH = Self::OPCODE_COLUMN_WIDTH,
-            OPERAND_COLUMN_WIDTH = Self::OPERAND_COLUMN_WIDTH,
-        );
-    }
-
     fn trace_execute_instruction(&mut self) -> JsResult<CompletionType> {
         let mut pc = self.vm.frame().pc as usize;
         let opcode: Opcode = self.vm.frame().code_block.read::<u8>(pc).into();
@@ -275,14 +239,9 @@ impl Context<'_> {
             stack
         };
 
-        println!(
-            "{:<TIME_COLUMN_WIDTH$} {:<OPCODE_COLUMN_WIDTH$} {operands:<OPERAND_COLUMN_WIDTH$} {stack}",
-            format!("{}μs", duration.as_micros()),
-            opcode.as_str(),
-            TIME_COLUMN_WIDTH = Self::TIME_COLUMN_WIDTH,
-            OPCODE_COLUMN_WIDTH = Self::OPCODE_COLUMN_WIDTH,
-            OPERAND_COLUMN_WIDTH = Self::OPERAND_COLUMN_WIDTH,
-        );
+        if let Some(trace) = &self.vm.trace {
+            trace.trace_instruction(duration.as_micros(), opcode.as_str(), operands, stack);
+        }
 
         result
     }
@@ -310,8 +269,14 @@ impl Context<'_> {
         let _timer = Profiler::global().start_event("run", "vm");
 
         #[cfg(feature = "trace")]
-        if self.vm.trace {
-            self.trace_call_frame();
+        if let Some(trace) = &self.vm.trace {
+            if !trace.bytecode_traced() {
+                trace.trace_compiled_bytecode(&self.vm, self.interner());
+                trace.trace_call_frame(&self.vm);
+                self.vm.trace.as_mut().expect("trace exists here").set_traced(true);
+            } else {
+                trace.trace_call_frame(&self.vm)
+            }
         }
 
         loop {
@@ -326,7 +291,7 @@ impl Context<'_> {
             }
 
             #[cfg(feature = "trace")]
-            let result = if self.vm.trace || self.vm.frame().code_block.traceable() {
+            let result = if self.vm.trace.is_some() {
                 self.trace_execute_instruction()
             } else {
                 self.execute_instruction()
--- a/boa_engine/src/vm/mod.rs
+++ b/boa_engine/src/vm/mod.rs
@@ -17,7 +17,7 @@ use boa_gc::{custom_trace, Finalize, Gc, Trace};
 use boa_profiler::Profiler;
 use std::mem::size_of;
 
-#[cfg(all(feature = "trace", not(target_arch="wasm32")))]
+#[cfg(all(feature = "trace", not(target_arch = "wasm32")))]
 use std::time::Instant;
 
 mod call_frame;
@@ -308,10 +308,8 @@ impl Context<'_> {
 
             #[cfg(feature = "trace")]
             let result = match &self.vm.trace {
-                Some(trace) if trace.should_trace() => {
-                    self.trace_execute_instruction()
-                },
-                _=>self.execute_instruction(),
+                Some(trace) if trace.should_trace() => self.trace_execute_instruction(),
+                _ => self.execute_instruction(),
             };
 
             #[cfg(not(feature = "trace"))]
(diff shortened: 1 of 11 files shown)


jedel1043:
Seeing this, I think it would be worth to create a `Tracer` trait instead of passing closures to handle tracing.

nekevss:
Switched the custom closures bits over to a trait.

It may be possible to go even further with the trait than currently implemented and selectively emit `Vm` state rather than pre-formatted messages. But I'm not exactly sure how much access we should provided, so I kept the current output as is for now.

jedel1043:
@HalidOdat and @jedel1043 will review this to see if we want to expose the instructions or if this is good enough.

boa_engine/src/vm/mod.rs
```diff
@@ -279,14 +252,12 @@ impl Context<'_> {
...
-            TIME_COLUMN_WIDTH = Self::TIME_COLUMN_WIDTH,
-            OPCODE_COLUMN_WIDTH = Self::OPCODE_COLUMN_WIDTH,
-            OPERAND_COLUMN_WIDTH = Self::OPERAND_COLUMN_WIDTH,
-        );
+        if let Some(trace) = &self.vm.trace {
+            #[cfg(not(target_arch = "wasm32"))]
+            trace.trace_instruction(duration.as_micros(), opcode.as_str(), &operands, &stack);
+            #[cfg(target_arch = "wasm32")]
+            trace.trace_instruction(opcode.as_str(), &operands, &stack);
+        }
```
jedel1043, line 260: For simplicity, I think we should just have a `Trace` type that doesn't print anything. That would allow us to remove the `Option` handling and just directly call `trace_instruction`.

boa_engine/src/vm/trace.rs
```diff
@@ -0,0 +1,289 @@
...
+impl VmTrace {
+    /// Returns if Trace type is a complete trace.
+    pub(crate) fn is_full_trace(&self) -> bool {
+        self.options.get().contains(TraceOptions::FULL_TRACE)
+    }
+
+    /// Returns if the trace is only a partial one.
+    pub fn is_partial_trace(&self) -> bool {
+        !self.is_full_trace()
+    }
```
jedel1043, line 126: Can you clarify what `full_trace` and `partial_trace` mean?

jedel1043 (commented):
Really nice API! This just needs a rebase, since it's still using the old `std::time::Instant`. I also have some suggestions and a small question.

HalidOdat (approved):
Sorry for the late review.

Besides jedel1043's comments, this looks good me! :)

boa_engine/src/vm/trace.rs
```diff
@@ -0,0 +1,289 @@
...
+        } else if self.is_partial_trace() && vm.frame().code_block().traceable() {
+            if !vm.frame().code_block().frame_traced() {
+                self.trace_current_bytecode(vm, interner);
+                vm.frame().code_block().set_frame_traced(true);
+            }
+            self.call_frame_header(vm);
+            self.activate();
+        } else {
+            self.call_frame_header(vm);
+        }
```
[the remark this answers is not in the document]
nekevss: Still need to work on this portion. The rest is rebased, so it shouldn't take too long 😄

HalidOdat (approved):
Did another review, just some small nitpicks that don't block merging. The rest looks good to me! :)

core/engine/src/vm/code_block.rs
```diff
@@ -69,10 +69,19 @@ bitflags! {
...
+    pub(crate) struct TraceFlags: u8 {
         /// Trace instruction execution to `stdout`.
-        #[cfg(feature = "trace")]
-        const TRACEABLE = 0b1000_0000_0000_0000;
+        const TRACEABLE = 0b0000_0001;
+
+        /// Has the `CodeBlock` been traced.
+        const CALLFRAME_TRACED = 0b0000_0010;
     }
 }
```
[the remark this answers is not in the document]
nekevss: Yeah, I can switch it over. I was sort of leaving at as a separate `u8`, so that we could add more options in the future and not constrain any flags added to `CodeBlockFlags`. That may have been premature though.

boa_engine/src/vm/trace.rs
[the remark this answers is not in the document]
nekevss: Added an enum (went with `TraceAction` over `ShouldTrace`), but now some of that logic is loaded into `should_trace`. Let me know what you think.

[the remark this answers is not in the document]
jedel1043: Really nice! Though, I think I didn't explain myself good enough. The goal of adding that enum was so that we can move the "should be traced partially or fully" logic from `VmTracer` to the tracers itself, by adding a `Tracer::should_trace(&self, code: Gc<CodeBlock>) -> TraceAction` method. Then, implementors would be the ones taking care of choosing when to just print the instructions vs the compiled bytecode.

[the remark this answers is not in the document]
nekevss: Should we be exposing that much of the `Vm` internals to `Tracer`? Part of the current implementation is to only expose preformatted strings in a bit more of a "controlled environment" so to speak, which included handling the logic aspect.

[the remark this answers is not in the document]
jedel1043: I'm pretty sure we already expose `CodeBlock` publicly, but users cannot do much with it aside from getting its address and name.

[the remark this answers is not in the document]
nekevss: I'm playing around with the update, and it's nicer in some places, but feels a lot more convoluted in others. I'm also not entirely sure on how to calling multiple `$boa.function.traceable()` might work as calling the function would remove the previous `Tracer`, which may cause problems on generators and async functions.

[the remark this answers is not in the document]
jedel1043: I'd say that we should try to find an architecture that allows us to always use the same `Tracer`, mostly for simplicity.

nekevss:
Made some changes to have `VmTrace` have a `Vec<Box<dyn Tracer>>` and moved `should_trace` to `Tracer`. Let me know what you think!

HalidOdat (approved):

HalidOdat:
Looks good to me! Just needs a rebase :)

core/engine/src/vm/mod.rs
```diff
@@ -76,7 +83,7 @@ pub struct Vm {
     pub(crate) realm: Realm,
 
     #[cfg(feature = "trace")]
-    pub(crate) trace: bool,
+    pub(crate) trace: VmTrace,
```
nekevss, line 86: A thought that I've had over the last day or so is whether it would be worthwhile redesigning so that the `Trace` is a trait on `Vm`.

```rust
struct Vm<T: ExecutionTrace=()> {
    ...
    trace: T
}
```
The only issue with this that I haven't totally thought through is that I'm not sure we'd be able to feature flag the fields (all though we could probably still feature flag the method calls)

core/engine/src/context/mod.rs
```diff
@@ -445,11 +449,18 @@ impl Context {
...
-        self.vm.trace = trace;
+    /// Initializes the default `Vm` trace from the context
+    pub fn init_trace(&mut self) {
+        self.vm.trace.activate_trace();
+    }
+
+    #[cfg(feature = "trace")]
+    /// Sets custom handling of trace messages.
+    pub fn set_tracer_implementation(&mut self, tracer: Box<dyn trace::Tracer>) {
+        // self.init_trace();
```
raskad, line 462:
```suggestion
```

core/engine/src/vm/code_block.rs
```diff
@@ -69,7 +69,11 @@ bitflags! {
 
         /// Trace instruction execution to `stdout`.
         #[cfg(feature = "trace")]
-        const TRACEABLE = 0b1000_0000_0000_0000;
+        const TRACEABLE = 0b0100_0000_0000_0000;
+
+        /// Stores whether the `CodeBlock` been traced.
```
raskad, line 74:
```suggestion
        /// Stores whether the `CodeBlock` has been traced.
```

core/engine/src/vm/mod.rs
raskad: I think the current design is fine. What would be the benefit of having it as a trait?

raskad (commented):
Just found some leftover comments / typos.

core/engine/src/vm/mod.rs
nekevss: Mostly I'm not sure if I'm totally in love with the `Vec<Box<dyn Tracer>>` on `VmTrace`. It works and improves the current functionality, but I'm mostly just thinking whether there may be a bit of a better design :confused: That being said, I'm not sure a theoretically better design out weighs the functionality here.

raskad: Imo the `Vec<Box<dyn Tracer>>` is totally fine for this use case.

> Using a trait would allow the trace to be defined at compile time essentially.

I'm not sure if that is a benefit. The upside would be reducing dynamic dispatch, but if you enable tracing, you probably do not care about that last bit of performance.
Being able to change the tracing at runtime also seems like a nice feature right?

nekevss: It is a nice feature! And you're right, if if someone enables tracing then they are probably not concerned about performance. But if it's switched to a trait that doesn't mean that there couldn't be an implementation of the trait that can be updated at runtime. And if we were to update this to something without dynamic dispatching, then we even provide an impl of the trait that could be updated at runtime.

The primary difference would be that it would be a user deciding to use dynamic dispatching over Boa locking the user into using it.

raskad: I would say the current implementation is good engough. If there are shortcomings we could change it. I would prefer that to adding a generic to the `Vm` struct. But if you think it is significantly better, go for it :)

jedel1043: I do think we could offer "combinators" for tracers though? Instead of having `Vec<Box<dyn Tracer>>`, why not something like `Box<dyn Tracer>`, where we `impl<T: Tracer, U: Tracer> Tracer for (T, U)`?

nekevss: :confused: Maybe. 

There's probably a better option with how `Tracer` is defined too. I was playing around with a new design and have something like:

```rust

pub trait Tracer: std::fmt::Debug {
    fn trace_instruction(&self, ctx: &TraceDecisionCtx<'_>) -> InstructionTraceAction;
    fn trace_bytecode(&self, ctx: &TraceDecisionCtx<'_>) -> BytecodeTraceAction;
    fn emit(&self, event: &TraceEvent<'_>);
}
```

Where `TraceEvent` is something like:

```rust
pub enum TraceEvent<'a> {
    NewCallFrame(&'a NewFrameEvent<'a>),
    ExitCallFrame(&'a ExitFrameEvent<'a>),
    Instruction(&'a InstructionTrace),
    ByteCode(&'a BytecodeTraceEvent),
}
```

But that design still has `VmTrace` using a `Vec`.

nekevss:
Closing since this PR is super stale, and I'm not convinced this is even a good approach anymore.

Hopefully will revisit the idea one day!

PR #3235, opened 2023-08-24
Clippy updates: add panics and etc.

nekevss:
This Pull Request is related to the failing CI on #3232.

It changes the following:

- Adds the ` # Panics` doc to the drop method.

EDIT: This kind of went a bit farther than I intended initially to just about all the lints. Comment is below.

### DIFF
--- a/ABOUT.md
+++ b/ABOUT.md
@@ -8,7 +8,7 @@ website][boa-web].
 Try out the most recent release with Boa's live demo
 [playground][boa-playground].
 
-# Boa Crates
+## Boa Crates
 
 - [**`boa_ast`**][ast] - Boa's ECMAScript Abstract Syntax Tree.
 - [**`boa_engine`**][engine] - Boa's implementation of ECMAScript builtin objects and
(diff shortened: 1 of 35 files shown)


Razican (approved):
Thanks!

nekevss:
Actually there's a lot more...clippy only feeds them one at a time apparently

nekevss:
I updated a large amount of the lints in most the crates.

There were two lints that I didn't address and marked them as allow currently:

1. There's a `Send`/``Sync` not implemented on `Arc` lint that is currently triggering on the `Inner` of `JsSymbol`
2. There's A LOT of missing `Panics` docs in `boa_engine` (50+) so I ignored them for now.

I wasn't sure if we should add panic docs on those functions or create an issue to remove the panics entirely and throw something of an `ImplementationError` or something of the sort.

Razican:
Se should at least have an issue for each of those. Ideally removing panics where possible.

jedel1043 (approved):

jedel1043:
Will merge this since it's holding back some dep updates, but I also agree that we should open issues for the pending tasks.

nekevss:
I'll submit the issues 😄

HalidOdat (approved):

PR #3253, opened 2023-09-03
Varying length operand

HalidOdat:
Depends on #3201 

Currently almost all operands are `u32` size, this is because it is enough for most cases, but we waste space since most of the time the size is smaller than `u8::MAX`.

This is an optimization that reduces that size of the operands of bytecode instructions, by default if the instruction is executed then it is assumed that the operands are `u8` type. When we need a larger size we prefix the opcode with a modifier (`Half` for `u16` or `Wide` for `u32`).

This allows us to greatly reduce the memory footprint of the generated bytecode.

V8 does this optimization, but it is not exclusive to v8 bytecode, since most CISC architectures have this as well (x86 being of of them).

### DIFF
--- a/boa_engine/src/bytecompiler/mod.rs
+++ b/boa_engine/src/bytecompiler/mod.rs
@@ -406,18 +406,18 @@ impl<'ctx, 'host> ByteCompiler<'ctx, 'host> {
             BindingOpcode::Var => {
                 let binding = self.initialize_mutable_binding(name, true);
                 let index = self.get_or_insert_binding(binding);
-                self.emit(Opcode::DefVar, &[Operand::U32(index)]);
+                self.emit_wide(Opcode::DefVar, index);
             }
             BindingOpcode::InitVar => {
                 if self.has_binding(name) {
                     match self.set_mutable_binding(name) {
                         Ok(binding) => {
                             let index = self.get_or_insert_binding(binding);
-                            self.emit(Opcode::DefInitVar, &[Operand::U32(index)]);
+                            self.emit_wide(Opcode::DefInitVar, index);
                         }
                         Err(BindingLocatorError::MutateImmutable) => {
                             let index = self.get_or_insert_name(name);
-                            self.emit(Opcode::ThrowMutateImmutable, &[Operand::U32(index)]);
+                            self.emit_wide(Opcode::ThrowMutateImmutable, index);
                         }
                         Err(BindingLocatorError::Silent) => {
                             self.emit_opcode(Opcode::Pop);
@@ -426,27 +426,27 @@ impl<'ctx, 'host> ByteCompiler<'ctx, 'host> {
                 } else {
                     let binding = self.initialize_mutable_binding(name, true);
                     let index = self.get_or_insert_binding(binding);
-                    self.emit(Opcode::DefInitVar, &[Operand::U32(index)]);
+                    self.emit_wide(Opcode::DefInitVar, index);
                 };
             }
             BindingOpcode::InitLet => {
                 let binding = self.initialize_mutable_binding(name, false);
                 let index = self.get_or_insert_binding(binding);
-                self.emit(Opcode::PutLexicalValue, &[Operand::U32(index)]);
+                self.emit_wide(Opcode::PutLexicalValue, index);
             }
             BindingOpcode::InitConst => {
                 let binding = self.initialize_immutable_binding(name);
                 let index = self.get_or_insert_binding(binding);
-                self.emit(Opcode::PutLexicalValue, &[Operand::U32(index)]);
+                self.emit_wide(Opcode::PutLexicalValue, index);
             }
             BindingOpcode::SetName => match self.set_mutable_binding(name) {
                 Ok(binding) => {
                     let index = self.get_or_insert_binding(binding);
-                    self.emit(Opcode::SetName, &[Operand::U32(index)]);
+                    self.emit_wide(Opcode::SetName, index);
                 }
                 Err(BindingLocatorError::MutateImmutable) => {
                     let index = self.get_or_insert_name(name);
-                    self.emit(Opcode::ThrowMutateImmutable, &[Operand::U32(index)]);
+                    self.emit_wide(Opcode::ThrowMutateImmutable, index);
                 }
                 Err(BindingLocatorError::Silent) => {
                     self.emit_opcode(Opcode::Pop);
@@ -464,6 +464,17 @@ impl<'ctx, 'host> ByteCompiler<'ctx, 'host> {
         self.emit_opcode(opcode);
         for operand in operands {
             self.emit_operand(*operand);
+        }
+    }
+
+    pub(crate) fn emit_wide(&mut self, opcode: Opcode, operand: u32) {
+        if let Ok(operand) = u8::try_from(operand) {
+            self.emit_opcode(opcode);
+            self.emit_u8(operand);
+        } else {
+            self.emit_opcode(Opcode::Wide);
+            self.emit_opcode(opcode);
+            self.emit_u32(operand);
         }
     }
 
@@ -533,7 +544,7 @@ impl<'ctx, 'host> ByteCompiler<'ctx, 'host> {
 
     fn emit_push_literal(&mut self, literal: Literal) {
         let index = self.get_or_insert_literal(literal);
-        self.emit(Opcode::PushLiteral, &[Operand::U32(index)]);
+        self.emit_wide(Opcode::PushLiteral, index);
     }
 
     fn emit_push_rational(&mut self, value: f64) {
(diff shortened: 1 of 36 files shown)


HalidOdat:
Ran the full quickjs benchmarks also collected the size of all the bytecode generated.

### Main

```
PROGRESS Richards
RESULT Richards 34.6
PROGRESS DeltaBlue
RESULT DeltaBlue 40.6
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 56.4
PROGRESS RayTrace
RESULT RayTrace 126
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 96.7
undefined
PROGRESS RegExp
PROGRESS Splay
RESULT Splay 103
PROGRESS NavierStokes
RESULT NavierStokes 128
SCORE 74.5
```

Total bytecode generated size: 202411B ==> 197.67KB

### PR

```
PROGRESS Richards
RESULT Richards 35.2
PROGRESS DeltaBlue
RESULT DeltaBlue 41.8
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 59.1
PROGRESS RayTrace
RESULT RayTrace 143
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 122
undefined
PROGRESS RegExp
PROGRESS Splay
RESULT Splay 108
PROGRESS NavierStokes
RESULT NavierStokes 128
SCORE 80.0
```

Total bytecode generated size: 126326B ==> ~123.37KB

This is a `1.6x` reduction in size, there are still some opcodes that are left once everything is done will make this as ready for review. But feel free to give feedback :)

boa_engine/src/bytecompiler/mod.rs
```diff
@@ -461,13 +465,46 @@ impl<'ctx, 'host> ByteCompiler<'ctx, 'host> {
...
+            VaryingOperandKind::Wide => self.emit_opcode(Opcode::Wide),
+        }
         self.emit_opcode(opcode);
         for operand in operands {
-            self.emit_operand(*operand);
+            self.emit_operand(*operand, varying_kind);
+        }
+    }
+
+    pub(crate) fn emit_varying_width(&mut self, opcode: Opcode, operand: u32) {
```
jasonwilliams, line 492: I think a comment above this function would be useful to explain why we use this over `emit`.

jasonwilliams:
This is awesome! Great optimization

boa_engine/src/vm/opcode/modifier.rs
```diff
@@ -0,0 +1,39 @@
...
+        let opcode = context.vm.read::<u8>() as usize;
+
+        Opcode::EXECUTE_FNS[256 + opcode](context)
+    }
+}
+
+/// `Wide` implements the Opcode Operation for `Opcode::Wide`
+///
+/// Operation:
+///  - [`Opcode`] prefix operand modifier, makes all varying operands of an instruction [`u16`] sized.
```
raskad, line 26:
```suggestion
///  - [`Opcode`] prefix operand modifier, makes all varying operands of an instruction [`u32`] sized.
```

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -121,16 +124,89 @@ where
...
 
+/// Represents a varying operand kind.
+#[derive(Default, Debug, Clone, Copy)]
+#[allow(missing_docs)]
+pub enum VaryingOperandKind {
+    #[default]
+    Short,
+    Half,
+    Wide,
+}
```
raskad, line 135: Is there a reason that these structs are `pub`?

I would be in favor of making `BindingOpcode`, `Instruction`, `InstructionIterator`, `Opcode` and `VaryingOperandKind` all `pub(crate)`. That way we can also remove the `missing_docs`.

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -443,6 +529,14 @@ pub(crate) trait Operation {
...
 
     fn execute(context: &mut Context<'_>) -> JsResult<CompletionType>;
+
+    fn half_execute(context: &mut Context<'_>) -> JsResult<CompletionType> {
+        Reserved::half_execute(context)
+    }
+
+    fn wide_execute(context: &mut Context<'_>) -> JsResult<CompletionType> {
+        Reserved::wide_execute(context)
+    }
```
raskad, line 539: The default impls for half and wide just exist to enable implementing `Operation` without providing all three methods right? I think a comment noting that would be nice.

raskad (approved):
Really nice work. I just have a few comments / questions.

jedel1043 (commented):
Really nice optimization work! The only suggestion I have would be to make `Opcode::Half/Wide` an operand itself. This would simplify the logic a lot, since `Operand` would remain the same, and only instructions that want to op-in into the varying operands logic would read the first operand as a `VaryingOperandKind`, then read the rest of the operands based off of that.
What do you think?

HalidOdat:
This would be a more simple approach, but it would incurred memory and performance penalties.

It would require an additional byte per varying instruction, additionally we would have to add logic for conditionally branching based on that varying kind (`Short`, `Half`, `Wide`) to read the correct sized operand.

I favored the current approach because by default, it's assumed that the operands are `Short a byte` no conditions or varying kind needed to be stored and when we prefix with the `Half`, or `Wide` modifiers we use the opcode execution table to jump directly, and the `Self::operation()` call is the last returned so it should be tail call optimized.

So we only pay a price (additional table jump) when we need more than a `Short` operand.

jedel1043:
> It would require an additional byte per varying instruction

Ahhh, because we would have to store the varying kind for all varying instructions even if the instruction is `Short`, right?
Yep, that sounds bad 😅 This design is fine then :)

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -121,16 +124,86 @@ where
     unsafe { read_unchecked(bytes, offset) }
 }
 
+/// Represents a varying operand kind.
+#[derive(Default, Debug, Clone, Copy)]
+pub(crate) enum VaryingOperandKind {
+    #[default]
+    Short,
+    Half,
+    Wide,
```
jedel1043, line 133: I don't think these names are descriptive of what they mean. `Short` varies per architecture, `Half` sounds like the operand will be half size the usual operand (which isn't true), and `Wide` doesn't really tell anything about the size of the operand.

Some alternatives:
- `U8`, `U16`, `U32` (preferred if still readable, but I think `VaryingOperand::u8(val)` is a bit weird)
- `Byte`, `BiByte`, `QuadByte` (preferred overall)
- `Byte`, `DoubleByte`, `QuadrupleByte` (a bit longer but more descriptive)

HalidOdat: The opcode is postfixed when we dump the bytecode to diffirentiate between varying kinds.

Example:

Current:
```js
000123    0509      none      PushLiteral                           2
000456    0511      none      PushLiteral.Half                      500
000789    0509      none      PushLiteral.Wide                      100000
```

Suggestion 1:
```js
000123    0509      none      PushLiteral                           2
000456    0511      none      PushLiteral.U16                       500
000789    0509      none      PushLiteral.U32                       100000
```

Suggestion 2:
```js
000123    0509      none      PushLiteral                           2
000456    0511      none      PushLiteral.BiByte                    500
000789    0509      none      PushLiteral.QuadByte                  100000
```

Suggestion 3:
```js
000123    0509      none      PushLiteral                           2
000456    0511      none      PushLiteral.DoubleByte                500
000789    0509      none      PushLiteral.QuadrupleByte             100000
```

I would prefer to use a short description `<= 5` letters to not clutter the dump too much. So I think the first suggestion is the best

jedel1043: Yep, let's go with that

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -1954,6 +2044,20 @@ generate_opcodes! {
...
+    ///
+    /// Stack: The stack changes based on the opcode that is being prefixed.
+    ModifierU16,
+
+    /// Opcode prefix modifier, [`Opcode`] prefix operand modifier, makes all [`VaryingOperand`]s of an instruction [`u32`] sized.
+    ///
+    /// Operands: opcode (operands if any).
+    ///
+    /// Stack: The stack changes based on the opcode that is being prefixed.
+    ModifierU32,
```
jedel1043, line 2059: Wouldn't be better `U16Operands` instead?

boa_engine/src/vm/opcode/mod.rs
```diff
@@ -442,7 +525,22 @@ pub(crate) trait Operation {
...
+    fn u16_execute(context: &mut Context<'_>) -> JsResult<CompletionType> {
+        Reserved::u16_execute(context)
+    }
+
+    /// Execute opcode with [`VaryingOperandKind::U32`] sized [`VaryingOperand`]s.
+    ///
+    /// By default if not implemented will call [`Reserved::u32_execute()`] that panics.
+    fn u32_execute(context: &mut Context<'_>) -> JsResult<CompletionType> {
+        Reserved::u32_execute(context)
+    }
```
jedel1043, line 543: I think it's better to use the full `execute_with_u16_operands`.

jedel1043 (approved):
Just some small nitpicks that don't block merging. Really well done!

PR #3273, opened 2023-09-12
Add ECMAScript 14 to `boa_tester`

jedel1043:
Depends on #3274.

This needs to be merged after the conformance webpage PR is merged, ~which I'll link here when I finish it~ which is linked above.

PR #3274, opened 2023-09-13
Update conformance page to ES14

jedel1043:
Related to #3273.

This PR adds a new `ES14` option to our version selector.

Had to also update the versions of the ESNext tests from 14 to 255, since ESNext is represented by `u8::MAX`. This is the cause of the big diffs on `main/latest.json` and `v0.17/latest.json`.

Since this'll probably break CI until #3273 is merged, we should wait until that PR is approved so that we can merge both consecutively.

### DIFF
--- a/test262/index.html
+++ b/test262/index.html
@@ -146,6 +146,7 @@
           <div class="form-group">
             <select class="form-select" id="info-options-es-version">
               <option value="255" selected>All</option>
+              <option value="14">ES14</option>
               <option value="13">ES13</option>
               <option value="12">ES12</option>
               <option value="11">ES11</option>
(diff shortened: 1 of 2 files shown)


test262/refs/tags/v0.17/latest.json
HalidOdat, line 1: Should this be commited/changed?

HalidOdat (approved):
Looks great to me!

Just a small comment :)

nekevss (approved):

test262/refs/tags/v0.17/latest.json
jedel1043: We technically don't need to commit it to pass CI, but I think we should, since it'll return wrong results for users who want to compare with main if we don't.

Issue #1804, opened 2022-01-29
Tracking issue for the `Temporal` builtin


Razican:
**Tracking for the `Temporal` built-in implementation**

- [x] First prototype (#3277)
- [x] Migrate Temporal to independent `boa_temporal` crate (#3461) 
- [x] Complete `Temporal.Duration` (#4070)
- [x] Complete `Temporal.PlainDate` (#4071)
- [x] Complete `Temporal.Instant` (#4074)
- [x] Complete `Temporal.PlainDateTime` (#4072)
- [x] Complete `Temporal.PlainYearMonth` (#4077)
- [x] Complete `Temporal.PlainMonthDay` (#4076)
- [x] Implement `Temporal.PlainTime` methods and operations (#4073)
- [x] ~~Implement `Temporal.TimeZone` methods and operations~~ (Removed from the Temporal Proposal Mid 2024)
- [x] Implement `Temporal.ZonedDateTime` methods and operations (#4075)
- [x] Complete `Temporal.Now` (dependent on `Temporal.Instant`, `Temporal.DateTime` and `Temporal.ZonedDateTime`)

**ECMASCript feature**
The new [`Temporal`](https://www.sitepoint.com/javascript-temporal-api-introduction/) API has been introduced to improve date handling in JavaScript, and we should add it to Boa to be up to date with the current ECMAScript standard. Tests have been added to the official ECMAScript Test Suite (test262), so we need to pass these tests to get full coverage.

The current proposal (Stage 3) specification can be found here: https://tc39.es/proposal-temporal/

In order to implement this new proposal, you can follow the example of other builtin objects, and feel free to ask for help: https://github.com/boa-dev/boa/tree/main/boa/src/builtins

**Example code**

This code should now work and give the expected result:
```javascript
// exact time since the Unix epoch on 1 Janary, 1970 UTC
Temporal.now.instant();
Temporal.now.instant().epochSeconds;
Temporal.now.instant().epochMilliseconds;

// current time zone
Temporal.now.timeZone();

// time in current location, e.g.
// 2021-09-18T04:17:48.435068431-04:00[America/New_York]
Temporal.now.zonedDateTimeISO();

// time in another time zone, e.g.
// 2021-09-18T09:17:48.438068435+01:00[Europe/London]
Temporal.now.zonedDateTimeISO('Europe/London');
```

jasonwilliams:
https://github.com/boa-dev/boa/issues/1180 Is related to this

[...]

PR #3277, opened 2023-09-14
First portion of the Temporal implementation

nekevss:
This PR is to propose the work completed so far on the `Temporal` branch for initial review and merging. The changes have become quite large (as visible via the diff 😅), but I think just about most of the building blocks are just about finished and/or close to being finished.

It changes the following:
- Adds the general structure many of the `Temporal` built-ins to varying levels `Calendar`, `Duration`, `Instant`, `PlainDate`, `PlainDateTime`, `PlainYearMonth`, and `PlainMonthDay` saw the most attention. `PlainTime`, `TimeZone`, and `ZonedDateTime` still need a decent amount of work.
- Adds parsing and very basic Parse Nodes for ISO8601 grammar.

There's still a decent amount more work to be done, but I think this is close to being a good starting point.

General design decisions that were made while working on this: the parser for iso8601 is currently separate from the `Lexer` and `Parser` for JavaScript, due to some grammar incompatibility that came up with the lexer. There is also a `BuiltinCalendar` trait that's meant to provide anyone who wants to implement any custom calendar. There are a few internal structs that are implemented with the idea of trying to prevent passing around entire `JsObjects` in favor of the native structs wherever possible. It also turned out that these structs were typically the target of the abstract operations too, so some abstract operations are implemented on the structs (I tried to leave a note in the abstract operations section when possible pointing to where the abstract operation is implemented).

There's probably a lot more I could say, but I think I'm going to leave it at this. 😆 Let me know what you think!

There are a few more things that I'd like to have completed on this before merging (although it's more of a wish list rather than necessary):

- [x] Basic Instant and Duration ISO8601 parsing (along with the Parse Nodes).
- [x] Clean up the Parse Nodes themselves (They're very stripped down at current).
- [x] A completed iso8601 Calendar implementation.

Footnote: I didn't remove `Temporal` from the `test_ignore` due to how fragmented the current progress is and the potential for false positives at this stage, but if anyone thinks it should be removed I definitely can.

Edit: Marking calendar complete but it's more of a "complete as can currently be for now".


boa_ast/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rust-version.workspace = true
 [features]
 serde = ["dep:serde", "boa_interner/serde", "bitflags/serde"]
 arbitrary = ["dep:arbitrary", "boa_interner/arbitrary", "num-bigint/arbitrary"]
+temporal = []
```
jedel1043, line 16: I'd propose changing this for an "experimental" flag instead, so that we don't need to remove the "temporal" feature when it reaches stage 4. That would also allow us to provide more implementations for other stage 3 features without adding more flags for each one of them.

.gitmodules
jedel1043, line 1: Was this intended? We already have #3214 for this, so we should definitely defer this change to that PR.

boa_engine/src/object/mod.rs
```diff
@@ -1951,20 +2346,21 @@ where
 
 /// Builder for creating native function objects
 #[derive(Debug)]
-pub struct FunctionObjectBuilder<'ctx, 'host> {
-    context: &'ctx mut Context<'host>,
+pub struct FunctionObjectBuilder {
+    realm: Realm,
```
jedel1043, line 2350: Can we extract this to its own PR? I saw that a considerable part of the PR changes were because of this, and I think if this wasn't part of the PR it'll make reviews easier.

jedel1043 (commented):
Just a preliminary review before checking the rest of the changes, but really nice work implementing all of this!

boa_engine/src/object/mod.rs
nekevss: Yeah, I think I can separate them out. I think they were part of the initial commits and mostly just kept them because they made sense. I'll try to get that completed tonight.

justingrant:
Congratulations! Temporal has a huge surface area, and seeing so much completed already in this PR is really nice to see.

I'm one of the Temporal proposal's champions. We're eager to help implementers be successful. If you have any questions about the spec, or if there's anything we can help with, then feel free to drop an issue in the [@tc39/proposal-temporal](https://github.com/tc39/proposal-temporal) repo and we'll respond quickly. Thanks again!

Are you collaborating with the folks building [ICU4X](https://github.com/unicode-org/icu4x)? (the Rust port of ICU4C the library that powers the calendar and time zone features of V8, JSC, and SpiderMonkey)

nekevss:
> seeing so much completed already in this PR is really nice to see.

Thanks! Although, I think there is still more work to be done.

> Are you collaborating with the folks building [ICU4X](https://github.com/unicode-org/icu4x)?

Not currently, and that's actually probably one of the more debatable aspects/areas to maybe improve on the current progress in this PR. Initially, my intended approach with the `BuiltinCalendar` trait was to wrap `icu_calendar`. I initially tried using `ICU4X`'s ISO calendar, but it doesn't really provide a good way to handle the proposal's `overflow` options that I was seeing besides just throwing an error on overflow, so I figured the best approach was to at least have specification ISO implementation.

jedel1043:
> the parser for iso8601 is currently separate from the `Lexer` and `Parser` for JavaScript, due to some grammar incompatibility that came up with the lexer.

I think this is okay. ICU4X [will provide a parser in the future](https://github.com/unicode-org/icu4x/issues/2127), and if they keep doing Temporal compatibility work, we'll eventually be able to replace all types with ICU4X's. So it is good to separate that to avoid integrating it too tightly to the JS parser.

sffc:
It is very much in the design of ICU4X to see the calendar-specific algorithms, including the ones for the ISO calendar, and flags like "overflow" and "constrain", land in `icu_calendar`.

ICU4X has support for all CLDR calendars, with a focus on conversion to/from ISO-8601. There is a partial implementation of date arithmetic (add/subtract/since/until) but it needs some polishing from someone who understands the Temporal spec.

I'd be very happy to mentor in this area, so that Boa can just pull in the ICU4X code for this.

nekevss:
> It is very much in the design of ICU4X to see the calendar-specific algorithms, including the ones for the ISO calendar, and flags like "overflow" and "constrain", land in `icu_calendar`.

That's great to hear! I actually thought of something that might work earlier when I was writing my reply to switch the ISO calendar implementation to `ICU4X`. I'm going to try and put it together later while making some of the other changes 😄.

boa_engine/src/object/mod.rs
nekevss: Ah that's right. The changes to `FunctionObjectBuilder` and `ObjectInitializer` stem from constructing the `Temporal.Now` object on `Temporal`. We only have access to `Realm` from `Init`, so the change was made to accomodate that. We could remove the initialization for `Temporal.Now` from this PR and submit it as a different PR as a follow up.

jedel1043: Hm, then should this use what `Intl` uses? It just references empty intrinsic objects, then the objects are initialized in `Realm` construction.

jedel1043: https://github.com/boa-dev/boa/blob/3c15e59e6717e018f826bf0b915faedff994d646/boa_engine/src/builtins/intl/mod.rs#L52-L56

nekevss: The Temporal.Now object:

 - is an [ordinary object](https://tc39.es/ecma262/#ordinary-object).
 - has a [[Prototype]] internal slot whose value is [%Object.prototype%](https://tc39.es/ecma262/#sec-properties-of-the-object-prototype-object).
 - is not a [function object](https://tc39.es/ecma262/#function-object).
 - does not have a [[Construct]] internal method; it cannot be used as a [constructor](https://tc39.es/ecma262/#constructor) with the new operator.
 - does not have a [[Call]] internal method; it cannot be invoked as a function.
 
 😕 Spec says it doesn't have a constructor

jedel1043: Yes, but you can also use `BuiltInBuilder` directly to initialize a custom intrinsic that is not a constructor, just like [`Math`](https://github.com/boa-dev/boa/blob/3c15e59e6717e018f826bf0b915faedff994d646/boa_engine/src/builtins/math/mod.rs#L35-L46)

hsivonen:
> There is a partial implementation of date arithmetic (add/subtract/since/until) but it needs some polishing from someone who understands the Temporal spec.

[The ICU4X issue for providing public methods for these operations](https://github.com/unicode-org/icu4x/issues/3964)

boa_engine/src/builtins/mod.rs
```diff
@@ -369,6 +387,12 @@ pub(crate) fn set_default_global_bindings(context: &mut Context<'_>) -> JsResult
     #[cfg(feature = "intl")]
     global_binding::<intl::Intl>(context)?;
 
+    #[cfg(feature = "experimental")]
+    {
+        global_binding::<temporal::Temporal>(context)?;
+        global_binding::<temporal::TimeZone>(context)?;
```
jedel1043, line 393: This adds a `TimeZone` property to the global object, but the spec only allows calling the constructor with `new Temporal.TimeZone`. Only adding `Temporal` to the global object should be enough.

boa_engine/src/context/intrinsics.rs
```diff
@@ -866,6 +1036,14 @@ pub struct IntrinsicObjects {
     /// [`%SegmentsPrototype%`](https://tc39.es/ecma402/#sec-%segmentsprototype%-object)
     #[cfg(feature = "intl")]
     segments_prototype: JsObject,
+
+    /// [`%Temporal%`](https://tc39.es/proposal-temporal/#sec-temporal-objects)
+    #[cfg(feature = "experimental")]
+    temporal: JsObject,
+
+    /// [`%Temporal%`](https://tc39.es/proposal-temporal/#sec-temporal-objects)
```
jedel1043, line 1044:
```suggestion
    /// [`%Temporal.Now%`](https://tc39.es/proposal-temporal/#sec-temporal-now-object)
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -1056,6 +1250,26 @@ impl IntrinsicObjects {
...
+    #[cfg(feature = "experimental")]
+    #[must_use]
+    #[inline]
+    pub fn temporal(&self) -> JsObject {
+        self.temporal.clone()
+    }
+
+    /// Gets the [`%Temporal.Now%`][spec] intrinsic object.
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-objects
```
jedel1043, line 1266:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-now-object
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -801,6 +841,136 @@ impl StandardConstructors {
     pub const fn segmenter(&self) -> &StandardConstructor {
         &self.segmenter
     }
+
+    /// Returns the `Temporal.Instant` constructor.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal.instant
```
jedel1043, line 850:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-instant-constructor
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -801,6 +841,136 @@ impl StandardConstructors {
...
+    pub const fn instant(&self) -> &StandardConstructor {
+        &self.instant
+    }
+
+    /// Returns the `Temporal.PlainDateTime` constructor.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal.plaindatetime
```
jedel1043, line 863:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-plaindatetime-constructor
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -801,6 +841,136 @@ impl StandardConstructors {
...
+    pub const fn plain_date_time(&self) -> &StandardConstructor {
+        &self.plain_date_time
+    }
+
+    /// Returns the `Temporal.PlainDate` constructor.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal.plaindate
```
jedel1043, line 876:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-plaindate-constructor
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -801,6 +841,136 @@ impl StandardConstructors {
...
+    pub const fn plain_date(&self) -> &StandardConstructor {
+        &self.plain_date
+    }
+
+    /// Returns the `Temporal.PlainTime` constructor.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal.plaintime
```
jedel1043, line 889:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-plaintime-constructor
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -801,6 +841,136 @@ impl StandardConstructors {
...
+    pub const fn plain_time(&self) -> &StandardConstructor {
+        &self.plain_time
+    }
+
+    /// Returns the `Temporal.PlainYearMonth` constructor.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal.plainyearmonth
```
jedel1043, line 902:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-plainyearmonth-constructor
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -801,6 +841,136 @@ impl StandardConstructors {
...
+    pub const fn plain_month_day(&self) -> &StandardConstructor {
+        &self.plain_month_day
+    }
+
+    /// Returns the `Temporal.TimeZone` constructor.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal=timezone-constructor
```
jedel1043, line 928:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-timezone-constructor
```

boa_engine/src/context/intrinsics.rs
```diff
@@ -801,6 +841,136 @@ impl StandardConstructors {
...
+    pub const fn zoned_date_time(&self) -> &StandardConstructor {
+        &self.zoned_date_time
+    }
+
+    /// Returns the `Temporal.Calendar` constructor.
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-zoneddatetime-constructor
```
jedel1043, line 967:
```suggestion
    /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal-calendar-constructor
```

boa_engine/src/builtins/temporal/mod.rs
```diff
@@ -0,0 +1,1149 @@
...
+    Context, JsBigInt, JsNativeError, JsNativeErrorKind, JsObject, JsResult, JsString, JsSymbol,
+    JsValue, NativeFunction,
+};
+use boa_ast::temporal::{self, UtcOffset};
+use boa_profiler::Profiler;
+
+// Relavant numeric constants
+pub(crate) const NS_PER_DAY: i64 = 86_400_000_000_000;
+pub(crate) const MICRO_PER_DAY: i64 = 8_640_000_000;
+pub(crate) const MILLI_PER_DAY: i64 = 8_600_000;
```
jedel1043, line 54:
```suggestion
pub(crate) const NANOS_PER_DAY: i64 = MICROS_PER_DAY * 1000;
pub(crate) const MICROS_PER_DAY: i64 = MILLIS_PER_DAY * 1000;
pub(crate) const MILLIS_PER_DAY: i64 = 24 * 60 * 60 * 1000; // 24 hours * 60 minutes * 60 seconds * 1000 millis
```

`MILLIS_PER_DAY` is wrong. We should probably use simple math to avoid calculation errors like this.

jedel1043 (changes requested):
Just a batch of reviews that I wanted to put out to be able to merge them. Mostly doc fixes.

boa_engine/src/builtins/temporal/mod.rs
nekevss: 😱 I missed the 4!!!

boa_parser/src/parser/mod.rs
nekevss, line 1: Just wanted to note: these are holdovers from when I was initially trying to use the pre-existing lexer and parser for Temporal. I'm going to revert these back.

boa_engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+    realm::Realm,
+    string::utf16,
+    Context, JsArgs, JsNativeError, JsObject, JsResult, JsString, JsSymbol, JsValue,
+};
+use boa_profiler::Profiler;
+
+use super::{plain_date::iso::IsoDateRecord, TemporalFields};
+
+#[derive(Debug, Clone)]
+pub(crate) struct IsoYearMonthRecord {
```
nekevss, line 17: This should be removed as well in favor of `IsoDateRecord`

nekevss (commented):
Just wanted to note some things I've noticed while going through this myself and plan to address 😄.

boa_parser/src/temporal/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+impl TemporalYearMonthString {
+    /// Parses a targeted string as a `YearMonth`
+    ///
+    /// # Errors
+    ///
+    /// The parse will error if the provided target is not valid
+    /// ISO8601 grammar.
+    pub fn parse(cursor: &mut IsoCursor) -> ParseResult<IsoParseRecord> {
+        if date_time::peek_year_month(cursor)? {
+            let ym = date_time::parse_year_month(cursor)?;
```
jedel1043, line 74: Shouldn't this try to parse a YearMonth inconditionally? According to the [spec](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporalyearmonthstring)

> If parseResult is a List of errors, throw a RangeError exception.

Also saw this in `TemporalMonthDayString`, and I think we don't need to double parse the input. The spec does that just so that `ParseISODateTime` fills the default values, but we can impl `Default` constant and do

```rust
let ym = date_time::parse_year_month(cursor)?;

// ... more checks

Ok(
    IsoParseRecord {
        year: ym.0,
        date: ym.1,
        tz_annotation,
        calendar,
        ..Default::default()
    }
)
```

boa_ast/src/temporal/mod.rs
```diff
@@ -0,0 +1,148 @@
+//! AST nodes for Temporal's implementation of ISO8601 grammar.
+
+/// TBD...
+#[derive(Default, Debug)]
+pub struct IsoParseRecord {
```
jedel1043, line 5: Maybe this should be split into three structs: `IsoDate`, `IsoTime` and `IsoDateTime`. That way, we can return `IsoTime` on `ParseTemporalTimeString` and `IsoDate` on `ParseTemporalMonthDay`/`ParseTemporalYearMonth`.

jedel1043 (changes requested):
Did a review pass on the parser code, and I have some suggestions.

boa_parser/src/temporal/mod.rs
nekevss: I think the reason I was initially peeking for a `YearMonth` / `MonthDay` is due to the record being either a `YearMonth` or an `AnnotatedDateTime`.

[Relevant Link](https://tc39.es/proposal-temporal/#prod-TemporalMonthDayString)

```
TemporalMonthDayString:
    AnnotatedMonthDay
    AnnotatedDateTime
```

I guess we could try to just straight parse `MonthDay` and `YearMonth`, and then if that errors try to parse `AnnotatedDateTime`, but I think peek is relatively quick(?), mostly just confirming general structure at certain indexes... could be wrong on that though.

jedel1043: Ahhh, I see, the parser must accept both a month-day or a full date time. This is fine then

boa_ast/src/temporal/mod.rs
nekevss: Hmmmm, the one thing that's a bit rough is that `MonthDay` and `YearMonth` can be an `AnnotatedDateTime`, so we would be omitting some of the parsed information if it was lowered to just returning an `IsoDate`.

jedel1043: Yes, but [`ParseTemporalMonthDayString`](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporalmonthdaystring) already filters out a lot of info and just returns the date. The same can be said about [`ParseTemporalTimeString`](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporaltimestring) and [`ParseTemporalYearMonth`](https://tc39.es/proposal-temporal/#sec-temporal-parsetemporalyearmonthstring)

jedel1043: Hhmm, but we need to throw `RangeError` instead of `ParseError` in the function, so maybe that would complicate things...

jedel1043: Ah, we already treat parsing errors as range errors on https://github.com/boa-dev/boa/blob/225eb4aded5a3bb28cbcec34a9e939c9a3dbd3a3/boa_engine/src/builtins/temporal/plain_date/mod.rs#L513. Are all errors thrown on parsing `RangeError`s?

nekevss: Yeah, I believe they are all range errors from what I've seen.

I was more so thinking that the error should be an ISO grammar specific error vs. using the pre-existing `ParseError`. Especially if the parser were potentially to ever move to one from `ICU4X`, then ideally the structure in the built-in may be better aligned. I just haven't gotten to that part yet 😆

I don't see a huge issue in shortening the return into smaller more selective nodes.

boa_engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -0,0 +1,999 @@
...
+            .into()
+    };
+
+    // 3. Let object be ? OrdinaryCreateFromConstructor(newTarget, "%Temporal.Duration.prototype%", « [[InitializedTemporalDuration]], [[Years]], [[Months]], [[Weeks]], [[Days]], [[Hours]], [[Minutes]], [[Seconds]], [[Milliseconds]], [[Microseconds]], [[Nanoseconds]] »).
+    let prototype =
+        get_prototype_from_constructor(&new_target, StandardConstructors::duration, context)?;
+    let mut object = prototype.borrow_mut();
+    let duration = object
+        .as_duration_mut()
+        .expect("prototype must be a Temporal.Duration.");
```
jasonwilliams, line 920: This doesn't work because the objectKind is "Ordinary" leading it to fail here.

jasonwilliams: You may still be in the middle of this though and its known this won't work

jasonwilliams (approved):
This looks really, good, still a fair few bits to iron out but its a great starting point. 

I'm happy for this to go in (bar any comments made)

boa_engine/src/builtins/date/mod.rs
```diff
@@ -7,7 +7,7 @@
 //! [spec]: https://tc39.es/ecma262/#sec-date-objects
 //! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
 
-mod utils;
+pub(crate) mod utils;
```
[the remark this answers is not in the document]
nekevss: 14.8.1 `GetUTCEpochNanoseconds` uses the preexisting `MakeDay` and `MakeTime`.

boa_engine/src/builtins/temporal/duration/mod.rs
nekevss: Oh wow. Nice catch! That was just a complete oversight on my part.

Updated it, so the next version should be good to go on this.

boa_engine/src/builtins/temporal/fields.rs
```diff
@@ -0,0 +1,563 @@
...
+            _ => month_code_integer,
+        };
+
+        self.month = Some(new_month);
+
+        Ok(())
+    }
+}
+
+fn month_code_to_integer(mc: &JsString) -> JsResult<i32> {
```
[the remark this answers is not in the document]
nekevss: In the long run, month-day fields should probably be eventually optimized down to a `u8` according to any overflow option, but I think that's more so on `IsoDateRecord`. `TemporalFields` is meant to be more of a Rust native representation of the `fields` object, so the fields mostly align with the conversion value.

boa_ast/src/temporal/mod.rs
nekevss: Cleaned them up 😄 let me know what you think

boa_engine/src/builtins/temporal/duration/record.rs
```diff
@@ -0,0 +1,1810 @@
...
+use super::super::{calendar, to_integer_if_integral, zoned_date_time};
+
+// ==== `DateDuration` ====
+
+#[derive(Debug, Default, Clone, Copy)]
+pub(crate) struct DateDuration {
+    years: f64,
+    months: f64,
+    weeks: f64,
+    days: f64,
```
jedel1043, line 20: Can you add a link to https://tc39.es/proposal-temporal/#sec-properties-of-temporal-duration-instances here and in `TimeDuration/DateTimeDuration`? It would be good to document that we're not using integers because the spec allows all float64-representable integers as valid values, and `f64` has many integers that aren't representable by `i64`.

jedel1043 (changes requested):
Another batch of reviews. I finished reviewing the parser changes and everything looks great!

boa_engine/src/builtins/temporal/options.rs
```diff
@@ -0,0 +1,402 @@
...
+    Month,
+    Week,
+    Day,
+    Hour,
+    Minute,
+    Second,
+    Millisecond,
+    Microsecond,
+    Nanosecond,
+    Auto,
```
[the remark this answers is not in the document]
nekevss: I'll take a look at it again. From what I remember, I wasn't initially wanting to add `Auto` either, but had ran into something that felt like it needed it. But it could probably also be refactored out.

boa_engine/src/bigint.rs
```diff
@@ -225,6 +225,17 @@ impl JsBigInt {
         Self::new(x.inner.as_ref().clone().add(y.inner.as_ref()))
     }
 
+    /// Utility function for performing `+` operation on more than two values.
+    #[inline]
+    #[must_use]
+    pub fn add_n(values: &[Self]) -> Self {
```
Razican, line 231: Does this need to be part of the public API?

boa_ast/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rust-version.workspace = true
 [features]
 serde = ["dep:serde", "boa_interner/serde", "bitflags/serde"]
 arbitrary = ["dep:arbitrary", "boa_interner/arbitrary", "num-bigint/arbitrary"]
+experimental = []
```
Razican, line 16: In other cases, we have used more specific feature names (such as `intl`). Should we just name the feature `temporal`?

boa_engine/src/builtins/temporal/date_equations.rs
```diff
@@ -0,0 +1,136 @@
...
+        305..=334 if in_leap_year => 10,
+        304..=333 => 10,
+        335..=366 if in_leap_year => 11,
+        334..=365 => 11,
+        _ => unreachable!(),
+    }
+}
+
+pub(crate) fn epoch_time_for_month_given_year(m: i32, y: i32) -> f64 {
+    let leap_day = i32::from(mathematical_days_in_year(y) == 366);
```
Razican, line 87: This seems a bit convoluted, right? Why not:

```rust
mathematical_days_in_year(y) - 365
```

Razican (commented):
Impressive work! I just had the chance to review part of it and add some comments.

boa_ast/Cargo.toml
jedel1043: We opted for this approach because `temporal` would be removed anyways when it reaches stage 4. Do other engines use a broad `experimental` flag for Temporal, or a specific `temporal` flag?

Razican: Makes sense... but going forward, we might want to have features specific to some JavaScript characteristics (maybe a feature per built-in, things like that), so it might not hurt to have a `temporal` feature, and `experimental` would just depend on `temporal`. Could make it easier to support other experimental features in the future.

jedel1043: That makes sense, since it allows users to reduce binary size if they need to.

nekevss: Should the feature flag update be included in this PR or a different PR?

boa_engine/src/builtins/temporal/calendar/iso.rs
```diff
@@ -0,0 +1,367 @@
...
+
+use super::BuiltinCalendar;
+
+use icu_calendar::{
+    iso::Iso,
+    week::{RelativeUnit, WeekCalculator},
+    Calendar, Date,
+};
+
+pub(crate) struct IsoCalendar;
```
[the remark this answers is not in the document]
nekevss: I think the `ISOWhatever` naming convention is relatively newer and potentially isolated just to the AST nodes. The rust notation (`IsoWhatever`) is probably better, so I'll switch them.

boa_engine/src/builtins/temporal/options.rs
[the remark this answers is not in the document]
nekevss: I was looking at the spec again, and `Auto` is a valid option value that can be provided for [`Duration`s round method options](https://github.com/tc39/proposal-temporal/blob/main/polyfill/index.d.ts), so I'm not entirely sure that it can be totally separated from the enum.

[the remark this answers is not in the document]
jedel1043: I think it can, with something like:

```Rust
enum OptionOrAuto<T> {
    Manual(T),
    Auto
}

impl<T: FromStr> FromStr for OptionOrAuto<T> {
    type Err = Error;
    fn parse(value: &str) -> Result<Self, Self::Err> {
         if value == "auto" {
            return Ok(Self::Auto);
        }

        Ok(Self::Manual(<T>::parse(value)?))
    }
}
```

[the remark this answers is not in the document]
jedel1043: With that definition, we could add something like `OptionOrAuto::resolve(t:T)` to resolve `Auto` to a specific value of `T` if needed.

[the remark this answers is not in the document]
jedel1043: We could also generalize it to an `OptionType` instead, since there could be some options that take nteger values + "auto"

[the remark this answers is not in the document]
nekevss: I can prototype it out. I'm a bit worried it's going to get a bit abstract with the nesting. We would potentially be returning `Option<OptionOrAuto<T>>` where `None`, `Auto`, or `T` may all be a valid value.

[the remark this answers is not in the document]
jedel1043: I'm proposing this assuming all "auto" options are eventually resolved to a specific `T`. If a type directly supports "auto" as a separate variant with completely different semantics as the other ones, I'd argue that it should be in `T` itself.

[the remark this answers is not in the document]
nekevss: `Auto` is technically considered a valid extra value in 3 circumstances in the proposal, `Duration.prototype.round`, `Calendar.prototype.dateUntil`, and `GetDifferenceSettings`. `dateUntil` does provide a T (Day). But the method that throws a wrench is `round` where `Auto` is used to designate the use of the computed `defaultLargestUnit` value where `largestUnitPresent` remains true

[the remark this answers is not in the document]
jedel1043: > We would potentially be returning `Option<OptionOrAuto<T>>` where `None`, `Auto`, or `T` may all be a valid value.

Maybe it looks that way thanks to the repetition of `Option`. Something like `Option<AutoValue<T>>`?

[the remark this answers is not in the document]
justingrant: We faced similar challenges in defining the TypeScript types for the Temporal API.  In case it's helpful, here's the TS types we ended up with. Each of these types defines valid inputs for units of various Temporal methods.

```ts
  export type LargestUnit<T extends DateTimeUnit> = 'auto' | T | PluralUnit<T>;
  export type SmallestUnit<T extends DateTimeUnit> = T | PluralUnit<T>;
  export type TotalUnit<T extends DateTimeUnit> = T | PluralUnit<T>;
```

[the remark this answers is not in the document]
justingrant: BTW, full set of TS types are here: https://github.com/tc39/proposal-temporal/blob/main/polyfill/index.d.ts

boa_ast/Cargo.toml
Razican: Can be done separately :)

boa_engine/src/builtins/temporal/options.rs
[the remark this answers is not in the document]
nekevss: @justingrant I've been using that polyfill as a reference! It was super useful for setting up the options initially / finally visualizing all the options 😄 (it was also linked above too). The current options were more or less pulled from that, although I tried to simplify `AssignmentOptions` and `ArithmeticOptions` to `ArithmeticOverflow` (might not be the best name).

boa_engine/src/builtins/temporal/date_equations.rs
```diff
@@ -0,0 +1,136 @@
...
+        244..=273 if in_leap_year => 8,
+        243..=272 => 8,
+        274..=304 if in_leap_year => 9,
+        273..=303 => 9,
+        305..=334 if in_leap_year => 10,
+        304..=333 => 10,
+        335..=366 if in_leap_year => 11,
+        334..=365 => 11,
+        _ => unreachable!(),
+    }
```
jedel1043, line 83: This could benefit from representing it as two arrays instead of a set of matches. That way, you can just use `binary_search` to find the position where `days` is less or equal than the value in the array:
```Rust
const DAYS: [u8; 12] = [30, 58, 89, ...];

let month = DAYS.binary_search(&days); // then unify the `Err` case with the `Ok` case. Both should always return the same index anyways.
```

boa_engine/src/builtins/temporal/mod.rs
```diff
@@ -0,0 +1,912 @@
...
+
+/// Abstract operation `ToZeroPaddedDecimalString ( n, minLength )`
+///
+/// The abstract operation `ToZeroPaddedDecimalString` takes arguments `n` (a non-negative integer)
+/// and `minLength` (a non-negative integer) and returns a String.
+fn to_zero_padded_decimal_string(n: u64, min_length: usize) -> String {
+    format!("{n:0min_length$}")
+}
+
+// TODO: 13.1 `IteratorToListOfType`
```
Razican, line 193: Is this still to-do? Seems pretty good :)

Might need some docs, though

Same for others in this file

boa_engine/src/builtins/temporal/options.rs
```diff
@@ -0,0 +1,402 @@
...
+
+    // 2. If increment is not finite, throw a RangeError exception.
+    if !increment.is_finite() {
+        return Err(JsNativeError::range()
+            .with_message("rounding increment was out of range.")
+            .into());
+    }
+
+    // 3. Let integerIncrement be truncate(ℝ(increment)).
+    let integer_increment = increment.trunc();
```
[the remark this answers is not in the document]
nekevss: Bit of an update here. So the below 109 is actually 10^9 that didn't copy over right, and I didn't notice first time through 😱  

I was looking at it last night like: huh 109 is a weird number that makes no sense in the context of the spec, and it definitely is a lot bigger.

boa_engine/src/builtins/temporal/options.rs
```diff
@@ -57,26 +57,26 @@ pub(crate) fn get_temporal_unit(
     options: &JsObject,
     key: &[u16],
     unit_group: TemporalUnitGroup,
-    required: Option<TemporalUnit>,
+    required: bool,
+    default: Option<TemporalUnit>,
     extra_values: Option<Vec<TemporalUnit>>,
     context: &mut Context<'_>,
-) -> JsResult<TemporalUnit> {
+) -> JsResult<Option<TemporalUnit>> {
```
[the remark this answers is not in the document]
nekevss: Yeah, removing required and default here in favor of `unwrap_or` and `ok_or_else` makes it a lot more ergonomic.

boa_engine/src/builtins/temporal/date_equations.rs
```diff
@@ -52,39 +54,18 @@ pub(crate) fn mathematical_in_leap_year(t: f64) -> i32 {
...
-        273..=303 => 9,
-        305..=334 if in_leap_year => 10,
-        304..=333 => 10,
-        335..=366 if in_leap_year => 11,
-        334..=365 => 11,
-        _ => unreachable!(),
+    const DAYS: [i32; 11] = [30, 58, 89, 120, 150, 181, 212, 242, 272, 303, 333];
+    let day = epoch_time_to_day_in_year(t);
+
+    let result = DAYS.binary_search(&day);
```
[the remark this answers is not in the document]
nekevss: Yeah, it should. 😕 I had it typed out too, but think I deleted it. Not sure why, but I put it back in.

jedel1043:
I think this is my last batch of reviews. There may be some patterns that can be improved, but I think we can iterate instead of trying to perfect this PR.

Just resolve my previous comments and we can merge this :)

boa_engine/src/builtins/temporal/plain_date_time/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+
+use self::iso::IsoDateTimeRecord;
+
+pub(crate) mod iso;
+
+/// The `Temporal.PlainDateTime` object.
+#[derive(Debug, Clone)]
+pub struct PlainDateTime {
+    pub(crate) inner: IsoDateTimeRecord,
+    pub(crate) calendar: JsValue,
```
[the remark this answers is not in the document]
nekevss: Yes, the `calendar` field is typically either an object representing a calendar or a string. It's a little weird because the `Instance` portion of the spec identifies the value as an object. However, `CreateTemporalDateTime` lists the value as: "calendar (a String or Object)".

When a method is called using the calendar, it typically defaults to the `Calendar-x` method, which essentially looks up the calendar and calls the method.

boa_engine/src/builtins/temporal/calendar/mod.rs
```diff
@@ -0,0 +1,2304 @@
...
+    string::{common::StaticJsStrings, utf16},
+    Context, JsArgs, JsNativeError, JsObject, JsResult, JsString, JsSymbol, JsValue,
+};
+use boa_profiler::Profiler;
+use rustc_hash::FxHashMap;
+
+mod iso;
+pub(crate) mod utils;
+
+#[cfg(feature = "experimental")]
```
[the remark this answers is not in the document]
nekevss: Probably not.

The editor I use is vscode. And if I use the builtin run test functionality, it won't run the tests without the configuration flag on the test module itself.

nekevss:
Linking #1804. 

Would it be worth while editing the original issue to make it more of a tracking issue in order to keep track of who's working on what. I plan to continue working on this, but just in case anyone else was interested in working on a part of the API, it could be useful.

Or would it be better to start a new issue altogether for tracking.

jedel1043 (commented):
I cannot imagine the massive work that was put onto understanding, designing, iterating and developing 13k lines of code for such an important feature. Really impressive work!

jedel1043 (approved):

jedel1043:
> Or would it be better to start a new issue altogether for tracking.

We can rename that issue and do a checklist with the missing APIs. You're probably the best person to list what's still missing, so I'll leave that task to you.

PR #3173, opened 2023-07-25
Added semver checks

Razican:
This Pull Request adds SemVer checks to our CI. This comes under the idea of releasing more often, and trying to release semver-compatible releases if possible.

It changes the following:

- Changes the current version to 0.17.1-dev, while we develop the new version
- Uses [cargo-semver-checks](https://crates.io/crates/cargo-semver-checks) to make sure we are semver-compatible.

PR #3290, opened 2023-09-19
Bump `rust-version` to 1.71

jedel1043:
Just to ensure our repo correctly reports our MSRV for users that want to use the `main` branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ members = [
 [workspace.package]
 edition = "2021"
 version = "0.17.0"
-rust-version = "1.66"
+rust-version = "1.71"
 authors = ["boa-dev"]
 repository = "https://github.com/boa-dev/boa"
 license = "Unlicense OR MIT"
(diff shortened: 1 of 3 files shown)


nekevss (approved):
LGTM. 

thought: Would it be worthwhile adding an MSRV test to our CI as well? I feel like we typically keep pretty close to current, but it could be useful to keep track of when we should increase the version.

jedel1043:
> Would it be worthwhile adding an MSRV test to our CI as well?

I think https://github.com/boa-dev/boa/pull/3173 does this, but we haven't got the time to review + merge it.

nekevss:
Hmmm, I think that's focused more on semver vs. msrv, right? 

As an example, I think `chrono` uses `dtolnay/rust-toolchain` to run `cargo test` on a specified version to ensure a PR also compiles and runs on that version.

jedel1043:
> Hmmm, I think that's focused more on semver vs. msrv, right?

Yeah, ignore that 😅 I was thinking that the msrv influenced semver, but I think those are orthogonal concepts.

HalidOdat (approved):
Looks good to me! :)

PR #3291, opened 2023-09-20
Added MSRV check

Razican:
This Pull Request is a follow-up of #3290.

It adds a test for the MSRV to the CI, so that we don't get similar issues in the future.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -75,7 +75,7 @@ jobs:
       # Get the rust_version from the Cargo.toml
       - name: Get rust_version
         id: rust_version
-        run: echo "::set-output name=rust_version::$(grep '^rust-version' Cargo.toml | cut -d' ' -f3 | tr -d '"')"
+        run: echo "rust_version=$(grep '^rust-version' Cargo.toml | cut -d' ' -f3 | tr -d '"')" >> $GITHUB_OUTPUT
       - uses: actions-rs/toolchain@v1.0.7
         with:
           toolchain: ${{ steps.rust_version.outputs.rust_version }}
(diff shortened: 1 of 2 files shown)


jedel1043:
Can we read the root `Cargo.toml` from CI? it would be ideal to check using the provided `rust_version` from our config.

nekevss:
Second the idea regarding reading the version in the `Cargo.toml` if possible.

Razican:
Will see how to do this this weekend :)

Razican:
@jedel1043, @nekevss this should be ready now :)

.github/workflows/rust.yml
```diff
@@ -66,6 +66,24 @@ jobs:
...
+  msrv:
+    name: Minimum supported Rust version
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      # Get the rust_version from the Cargo.toml
+      - name: Get rust_version
+        id: rust_version
+        run: echo "::set-output name=rust_version::$(grep '^rust-version' Cargo.toml | cut -d' ' -f3 | tr -d '"')"
```
nekevss, line 78: Huh, looks like `set-output` is going to be [deprecated soon](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), so maybe something more akin to below.

```suggestion
        run: echo "rust_version=$(grep '^rust-version' Cargo.toml | cut -d' ' -f3 | tr -d '"')" >> $GITHUB_OUTPUT
```

nekevss (commented):
Was double checking the run and noticed a deprecation warning. Maybe adjust it to use `$GITHUB_OUTPUT`?

jedel1043 (approved):
Nice improvement!

HalidOdat (approved):
Looks perfect to me! :)

nekevss (approved):
LGTM

PR #3292, opened 2023-09-20
Implement `Gc::new_cyclic`

jedel1043:
This is just the corresponding replacement for the [`Rc::new_cyclic`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.new_cyclic) method, which showed to be useful while I was trying to improve some patterns in our modules implementation.


HalidOdat (approved):
Looks perfect to me! :)

raskad (approved):

PR #3294, opened 2023-09-21
Implement synthetic modules

### LINKED ISSUE #3224: Implement Synthetic Modules
Right now users that want to define modules from Rust code necessarily need to use an embed script to do that. However, with the [JSON Modules proposal](https://tc39.es/proposal-json-modules) coming soon, the spec will have a way to define [Synthetic Modules](https://tc39.es/proposal-json-modules/#sec-synthetic-module-records), which covers this exact use case.

We already have some glue code on our  implementation of modules to support this, so I think we can get a bit ahead and implement Synthetic Modules.

jedel1043:
Depends on #3292.

This Pull Request closes #3224.

It changes the following:

- Implements [Synthetic Modules](https://tc39.es/proposal-json-modules/#sec-synthetic-module-records) per the JSON modules proposal.
- Improves some patterns by using `Gc::new_cyclic`.

@HalidOdat I had to do a weird hack to be able to push the `ActiveRunnable` without a `CodeBlock` available to create a `CallFrame`. Since you're currently refactoring the VM, I didn't want to modify the definition of `CallFrame` to accommodate this special case. I could leave it as it is, of course, but would there be another way to store the `ActiveRunnable` inside the callstack without having to create a dummy `CodeBlock` in the process?

### DIFF
--- a/boa_engine/src/lib.rs
+++ b/boa_engine/src/lib.rs
@@ -199,8 +199,7 @@ pub trait JsArgs {
     /// `args.get(n).cloned().unwrap_or_default()` or
     /// `args.get(n).unwrap_or(&undefined)`.
     ///
-    /// This returns a reference for efficiency, in case you only need to call methods of `JsValue`,
-    /// so try to minimize calling `clone`.
+    /// This returns a reference for efficiency, in case you only need to call methods of `JsValue`.
     fn get_or_undefined(&self, index: usize) -> &JsValue;
 }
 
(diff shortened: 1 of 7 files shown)


HalidOdat:
We could wrap `CodeBlock` in an optional, but we will have to `.expect` on every access to contents for ordinary functions.

We could take the v8 approach, and merge value stack with frame stack, have different types of call frames (though this may need some unsafe)

For now we can leave a `TODO` and resolve it once call refactor (#3295) is complete.

jedel1043:
> For now we can leave a `TODO` and resolve it once call refactor (#3295) is complete.

Sounds good! I already left a TODO on it.

raskad (approved):
Nice work and great example :)

PR #3295, opened 2023-09-21
Refactor ordinary VM calling

HalidOdat:
This PR is a WIP refactor of ordinary vm functions currently only calling is refactored (not generators, new, bind-functions, etc)

```JavaScript
>> $boa.limits.recursion = 100000
>> function fact(n) { if (n == 0) { return 1} return n * fact(n - 1) } fact(10000)
Uncaught: RuntimeLimit: Maximum call stack size exceeded
```

Now we are now no longer bound on native call stack size :)


raskad (approved):
Looks really nice.

Just to be clear, we still depend on the native call stack size when we call ordinary functions from native functions right? For example the following code will extend the call native call stack:

```js
[0].map((e) => { return ++e})
```

HalidOdat:
> Just to be clear, we still depend on the native call stack size when we call ordinary functions from native functions right? For example the following code will extend the call native call stack:
> 
> ```js
> [0].map((e) => { return ++e})
> ```

It will extend the call native stack only as much as our rust implementation of `map` takes, all the ordinary calls don't extend the call stack, including the ordinary functions that `map` calls.

So the following code will have the same effect as the previous on the native call stack:

```js
const fact = (n) => {
   return n == 0 ? 1 : (n * fact(n - 1))
}

[100].map((e) => { return fact(e) })
```

boa_engine/src/object/internal_methods/mod.rs
```diff
@@ -299,9 +297,46 @@ pub(crate) struct InternalObjectMethods {
...
+    pub(crate) __call__:
+        fn(&JsObject, argument_count: usize, &mut Context<'_>) -> JsResult<CallValue>,
     pub(crate) __construct__:
-        fn(&JsObject, &[JsValue], &JsObject, &mut Context<'_>) -> JsResult<JsObject>,
+        fn(&JsObject, argument_count: usize, &mut Context<'_>) -> JsResult<CallValue>,
+}
+
+/// The return value of an internal method (`[[Call]]` or `[[Construct]]`).
+///
+/// This is done to avoid recusion.
```
nekevss, line 308:
```suggestion
/// This is done to avoid recursion.
```

boa_engine/src/vm/call_frame/mod.rs
```diff
@@ -42,7 +43,12 @@ pub struct CallFrame {
     /// \[\[ScriptOrModule\]\]
     pub(crate) active_runnable: Option<ActiveRunnable>,
 
-    pub(crate) active_function: Option<JsObject>,
+    pub(crate) environments: EnvironmentStack,
+
+    pub(crate) realm: Realm,
+
+    pub(crate) exit_early: bool,
```
nekevss, line 50: Thought: Would it be better to have a singular `bitflag` with more states eventually over a `exit_early` and `construct`?

I'd always hoped to eventually return to `CompletionType` and remove `CompletionType::Yield` in favor of the spec's `SuspendedYield` state. This may be a bit out of scope on this PR, but I thought I'd bring it up on the off chance you had run across something while working on this or may have had some thoughts about it.

nekevss (commented):
This is looking great! Still need to review it in a bit more detail, but thought I'd add a few comments from my initial review.

boa_engine/src/error.rs
```diff
@@ -379,6 +379,13 @@ impl JsError {
         }
         self
     }
+
+    /// Is the [`JsError`] not catchable in JavaScript.
+    #[inline]
+    pub(crate) fn is_non_catchable(&self) -> bool {
```
jedel1043, line 385: Would be better to implement it as `is_catchable`, I think. It's not common in Rust APIs to implement the negative of a property.

boa_engine/src/vm/call_frame/mod.rs
```diff
@@ -57,11 +79,32 @@ impl CallFrame {
...
+    ///   --- frame pointer                                ^
+    ///  /                      __________________________/
+    /// /                      /                          \
+    /// | 0: this | 1: func | 2: arg1 | ... | (2 + N): argN |
+    ///    \            /
+    ///     ------------
+    ///     |
+    /// function prolugue
+    /// ```
+    pub(crate) const FUCNTION_PROLOGUE: usize = 2;
```
nekevss, line 97:
```suggestion
    pub(crate) const FUNCTION_PROLOGUE: usize = 2;
```

boa_engine/src/module/source.rs
```diff
@@ -1600,22 +1600,23 @@ impl SourceTextModule {
...
+        // 14. Set the LexicalEnvironment of moduleContext to module.[[Environment]].
+        // 15. Set the PrivateEnvironment of moduleContext to null.
         let call_frame = CallFrame::new(
             codeblock.clone(),
             Some(ActiveRunnable::Module(parent.clone())),
-            None,
+            envs,
+            realm.clone(),
         );
         context.vm.push_frame(call_frame);
```
nekevss, line 1615: Thought: Would it be worthwhile adding an extra method on `vm` akin to `push_frame_with_stack` or something similar.

There's a few times where the frame is pushed and then the this and function value are pushed onto the stack separate. Is there any value in coupling the frame push and the stack push in these instances.

boa_engine/src/module/synthetic.rs
```diff
@@ -305,29 +305,30 @@ impl SyntheticModule {
...
-        std::mem::swap(&mut context.vm.environments, &mut environments);
-        // 3. Set the Realm of moduleContext to module.[[Realm]].
-        context.swap_realm(&mut realm);
         // 7. Suspend the currently running execution context.
         // 8. Push moduleContext on to the execution context stack; moduleContext is now the running execution context.
         context.vm.push_frame(callframe);
 
+        // 2. Set the Function of moduleContext to null.
+        context.vm.push(JsValue::default()); // this
+        context.vm.push(JsValue::default()); // func
```
nekevss, line 322:
```suggestion
        context.vm.push(JsValue::null()); // func
```
I could be missing something. But shouldn't function for module be null over undefined given the spec context?

nekevss (commented):
Everything looks great! Nice work on this. Just came across a couple things I was curious about 😄

boa_engine/src/module/source.rs
HalidOdat: > Thought: Would it be worthwhile adding an extra method on `vm` akin to `push_frame_with_stack` or something similar.
>
> There's a few times where the frame is pushed and then the this and function value are pushed onto the stack separate. Is there any value in coupling the frame push and the stack push in these instances.

Having a utility function seems fine, maybe even `pop_frame_with_stack`, Ill add it :)

nekevss (approved):
Looks good to me!

PR #3305, opened 2023-09-25
Implement step 5 in RegExp constructor

HalidOdat:
```
5. Else if patternIsRegExp is true, then
    a. Let P be ? Get(pattern, "source").
    b. If flags is undefined, then
        i. Let F be ? Get(pattern, "flags").
    c. Else,
        i. Let F be flags.
```


jedel1043 (approved):
Nice catch!

raskad (approved):

PR #3307, opened 2023-09-25
Fix tagged template `this` in strict mode

HalidOdat:
In strict mode `this` should be undefined.

Causing test262 test to fail: [`test/language/expressions/tagged-template/call-expression-context-strict.js`](https://github.com/tc39/test262/blob/main/test/language/expressions/tagged-template/call-expression-context-strict.js)

Additionally this remove the unneeded `swap` opcode when calling functions.


nekevss (approved):

raskad (approved):
Nice find!

PR #3309, opened 2023-09-25
Replace #[deny] with #[warn]

jedel1043:
See https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html for context.

This also improves a bit the prototyping experience, since now we can compile and run with warnings even if there are missing docs or suboptimal code patterns.


nekevss (approved):

HalidOdat (commented):
Nice change! :)

HalidOdat (approved):

PR #3312, opened 2023-09-26
Fix class inherit from `null`

HalidOdat:
Fixes class inheritance from `null`, which was causing this test262 test to fail: [test/language/statements/class/subclass/class-definition-null-proto-this.js](https://github.com/tc39/test262/blob/main/test/language/statements/class/subclass/class-definition-null-proto-this.js)


jedel1043 (approved):
Nice catch!

raskad (approved):

PR #3318, opened 2023-09-27
Introduce experimental features

jedel1043:
Related to #3277

Since the Temporal PR is as big as it gets, I thought it would be good to introduce the experimental flag in a separate PR, which should reduce the review work a bit.
This also introduces the [`Promise.withResolvers`](https://github.com/tc39/proposal-promise-with-resolvers) method, just so that there's at least some feature using the flag.


raskad (approved):
Looks good

nekevss (approved):
Looks good to me. Definitely like the idea of introducing it separate from the Temporal PR.

HalidOdat (approved):
Looks good to me! :)

PR #3319, opened 2023-09-28
Remove direct conversion from `&str` to `JsValue`/`PropertyKey`.

jedel1043:
This is with the objective of avoiding hidden conversions while doing the conversion from a string literal to a `JsValue`. Removing the direct conversion to `JsValue` makes it a bit easier to realize that this is doing a costly conversion of
`string literal -> (runtime) decode UTF-8 to UTF-16 -> (runtime) JsString -> (runtime) JsValue`
and pushes the alternative of using the `js_string` macro that makes the optimal conversion
`string literal -> (compile time) &[u16] -> (runtime) JsString -> (runtime) JsValue`

This should technically improve realm initialization times, but let's see what the benchmarks show.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ test-results
 
 # dhat
 dhat-*.json
+perf.data*
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,6 @@ chrome_profiler.json
 # e2e test
 playwright-report
 test-results
+
+# dhat
+dhat-*.json
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -342,6 +342,7 @@ dependencies = [
  "boa_runtime",
  "clap",
  "colored",
+ "dhat",
  "jemallocator",
  "phf",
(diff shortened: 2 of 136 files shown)


jedel1043:
Ok, results are a bit erratic (+293.51% in parsing string copy without even touching `boa_parser`???), so I'll run them locally.

jedel1043:
Yep, some nice wins by skipping the conversions.
While `main` uses 600KB of data:

image

This PR uses 573KB:

image

Mainly because conversions use an additional 24KB of intermediary allocations:

image

It would be cool if we could also remove the additional 12KB used by the thread local cache of static strings (maybe a phf?):

image

raskad (approved):
I really like the change. The cost of conversions like this should be very clear.

nekevss (approved):
I like this! I am a bit curious about documentation on using `js_string` over `&str` for contributors and anybody consuming the API. But I'm fine with merging as is.

PR #3321, opened 2023-09-29
Add `NonMaxU32` as integer variant for `PropertyKey`

raskad:
This Pull Request supersedes #2131

It changes the following:

- Add a `NonMaxU32` that is used as the integer variant for `PropertyKey`.

### DIFF
--- a/boa_engine/src/property/mod.rs
+++ b/boa_engine/src/property/mod.rs
@@ -636,7 +636,7 @@ where
     if byte == CHAR_ZERO {
         if len == 1 {
             // SAFETY: `0` is not `u32::MAX`.
-            return Some(NonMaxU32::new_unchecked(0));
+            return unsafe { Some(NonMaxU32::new_unchecked(0)) };
         }
 
         // String "012345" is not a valid index.
@@ -660,7 +660,7 @@ where
 
         // SAFETY: `result` cannot be `u32::MAX`,
         //         because the length of the input is smaller than `MAX_CHAR_COUNT`.
-        Some(NonMaxU32::new_unchecked(result))
+        unsafe { Some(NonMaxU32::new_unchecked(result)) }
     }
 }
 
@@ -729,14 +729,14 @@ impl From<PropertyKey> for JsValue {
 impl From<u8> for PropertyKey {
     fn from(value: u8) -> Self {
         // SAFETY: `u8` can never be `u32::MAX`.
-        Self::Index(NonMaxU32::new_unchecked(value.into()))
+        unsafe { Self::Index(NonMaxU32::new_unchecked(value.into())) }
     }
 }
 
 impl From<u16> for PropertyKey {
     fn from(value: u16) -> Self {
         // SAFETY: `u16` can never be `u32::MAX`.
-        Self::Index(NonMaxU32::new_unchecked(value.into()))
+        unsafe { Self::Index(NonMaxU32::new_unchecked(value.into())) }
     }
 }
 
--- a/boa_engine/src/property/mod.rs
+++ b/boa_engine/src/property/mod.rs
@@ -603,7 +603,7 @@ pub enum PropertyKey {
 }
 
 /// Utility function for parsing [`PropertyKey`].
-fn parse_u32_index<I, T>(mut input: I) -> Option<u32>
+fn parse_u32_index<I, T>(mut input: I) -> Option<NonMaxU32>
 where
     I: Iterator<Item = T> + ExactSizeIterator + FusedIterator,
     T: Into<u16>,
@@ -635,7 +635,8 @@ where
     let byte = input.next()?.into();
     if byte == CHAR_ZERO {
         if len == 1 {
-            return Some(0);
+            // SAFETY: `0` is not `u32::MAX`.
+            return unsafe { Some(NonMaxU32::new_unchecked(0)) };
         }
 
         // String "012345" is not a valid index.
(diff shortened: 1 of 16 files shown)


boa_engine/src/property/mod.rs
```diff
@@ -674,29 +675,36 @@ impl From<&[u16]> for PropertyKey {
 impl From<JsString> for PropertyKey {
     #[inline]
     fn from(string: JsString) -> Self {
-        parse_u32_index(string.as_slice().iter().copied()).map_or(Self::String(string), Self::Index)
+        parse_u32_index(string.as_slice().iter().copied())
```
HalidOdat, line 678: I think it's better to make `parse_u32_index` return a `Option<NonMaxU32>`.

[here](https://github.com/boa-dev/boa/pull/3321/files#diff-d70662fbce40a1232f13bd61110a0252420431ec8ca66da3d07c3003b9e01a63R650-R652) we should return with max check because it can be greater or equal to u32 max. And in the else case we can return with `new_unchecked` since it will never be greater or equal to max because of the digit count check.

boa_engine/src/property/nonmaxu32.rs
```diff
@@ -0,0 +1,39 @@
...
+impl NonMaxU32 {
+    /// Creates a non-max `u32`.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the given value is not `u32::MAX`.
+    #[must_use]
+    pub const unsafe fn new_unchecked(inner: u32) -> Self {
+        // SAFETY: The caller must ensure that `inner` is not `u32::MAX`.
+        let inner = unsafe { NonZeroU32::new_unchecked(inner + 1) };
```
HalidOdat, line 18: why add one and subtract when getting?

HalidOdat (approved):
Looks good, just some comments :)

boa_engine/src/property/nonmaxu32.rs
jedel1043: Also, since this is an unsafe function and the user has to ensure the number is not `u32::MAX`, you can use `wrapping_add` in order to remove panicking code.

raskad: > why add one and subtract when getting?

Since we are using `NonZeroU32` but our type excludes `u32::MAX` instead of `0`. If we would not do the addition we would run into `NonZeroU32::new_unchecked(0)` which is undefined behaviour.

HalidOdat: Ahh, Bu why not use a plain `u32`, `NonMaxU32` can be zero?

raskad: There was a discussion over in https://github.com/boa-dev/boa/pull/2131 to use it for a potential non zero optimization. In my opinion we could also remove it and just use a `u32` since we only use `NonMaxU32` in `PropertyKey` which also has `JsString` and `JsSymbol` which are tagged pointers at `size = 8`. So optimizing would probably never work.

If we do that my only question is if we leave `NonMaxU32::new_unchecked` `unsafe`. It could not produce undefined behaviour, so in my view probably not, right?

HalidOdat: I see, thanks for the clarification :)

> If we do that my only question is if we leave `NonMaxU32::new_unchecked` `unsafe`. It could not produce undefined behaviour, so in my view probably not, right?

IMO, yes, We can leave a `debug_assert!` in it, just in case :)

HalidOdat (approved):
Looks perfect to me! :)

boa_engine/src/property/nonmaxu32.rs
```diff
@@ -0,0 +1,35 @@
...
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the given value is not `u32::MAX`.
+    #[must_use]
+    pub const fn new_unchecked(inner: u32) -> Self {
+        debug_assert!(inner != u32::MAX);
+
+        Self { inner }
+    }
```
jedel1043, line 18: I think this should be an unsafe function. My reasoning is that, in the future, we could implement an optimization that assumes that the invariant holds. However, if leave this function as safe, we'd have to make a breaking change to add `unsafe` at that moment. So, in preparation to that I'd mark it as unsafe.

raskad: I would argue that if we need to change this to be unsafe later, we just do the breaking change. In my opinion it is more important to have a clean API that does what is expected, than to avoid breaking changes.

jedel1043: It's not only a matter of doing the breaking change though. If, for example, someone tried to do an unsafe optimization assuming the invariant holds, and the invariant didn't hold, it would mean that a safe function would be the cause of UB.

jedel1043: E.g. [`String::from_utf8_unchecked`](https://doc.rust-lang.org/std/string/struct.String.html#safety-1) with it's safety comment.

raskad: Well but currently we do not have any optimizations based on the invariant, right? If we add some, we would have to mark the function as unsafe in that same change. In contrast I think there are unsafe optimizations in `String` that debend on the bytes being UTF-8.

jedel1043: > Well but currently we do not have any optimizations based on the invariant, right?

Note that the safety comment doesn't specify that there are uses of that method that unsafely assume the string is utf8.
> If this constraint is violated, it may cause memory unsafety issues with future users (sic) of the String, ...
Emphasis on "... with future uses of the String"

> If we add some, we would have to mark the function as unsafe in that same change.

I don't think it's ideal to rely on the fact that all of us have to remember to change the function to `unsafe` in the same PR.

raskad: I think the `String` documentation is not talking about "future" in the sense of functionality that may or may be not added in the future but rather the future of the created `String` itself. I think the next sentence makes it a little more clear: `[...] as the rest of the standard library assumes that Strings are valid UTF-8.`

> I don't think it's ideal to rely on the fact that all of us have to remember to change the function to unsafe in the same PR.

I very much disagree with this. When we add unsafe optimizations, we have to check that there is no way the constraints that the optimizations depend on can be violated outside of unsafe functions themselves. In this case, we would have to make sure that there is no way a `NonMaxU32` can be constructed without a `u32::MAX` value, expect from an unsafe function that documents this constraint.

Nevermind all of that, let's not get too hung up on this. If you think the function should be unsafe, I will add it back and remove the debug_assert.

jedel1043 (approved):
Nice work! And thank you for dealing with my pedantry 😅

PR #3322, opened 2023-09-29
Refactor function internal methods

HalidOdat:
This PR separates the native function from JavaScript function, their internal methods and their storage.

This allows us to remove some unneeded check as well as makes #3295 easier to implement.

### DIFF
--- a/boa_engine/src/builtins/function/mod.rs
+++ b/boa_engine/src/builtins/function/mod.rs
@@ -18,8 +18,7 @@ use crate::{
     environments::{EnvironmentStack, PrivateEnvironment},
     error::JsNativeError,
     js_string,
-    native_function::NativeFunction,
-    object::{internal_methods::get_prototype_from_constructor, JsObject, Object, ObjectData},
+    object::{internal_methods::get_prototype_from_constructor, JsObject, ObjectData},
     object::{JsFunction, PrivateElement, PrivateName},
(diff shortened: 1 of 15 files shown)


boa_engine/src/builtins/function/mod.rs
```diff
@@ -266,77 +189,65 @@ impl fmt::Debug for FunctionKind {
...
 ///
 /// `FunctionBody` is specific to this interpreter, it will either be Rust code or JavaScript code
 /// (AST Node).
 ///
 /// <https://tc39.es/ecma262/#sec-ecmascript-function-objects>
 #[derive(Debug, Trace, Finalize)]
-pub struct Function {
-    kind: FunctionKind,
-    realm: Realm,
+pub struct OrdinaryFunction {
```
HalidOdat, line 218: Maybe rename this to `ByteCodeFunction`, instead of `OrdinaryFunction`?

raskad (approved):

boa_engine/src/builtins/function/mod.rs
raskad: Could be good, but I think we have that same naming in a lot of places. If we want to do that, let's do it in a separate MR for better reviewability.

jedel1043 (approved):
Nice refactoring!

PR #3325, opened 2023-09-29
Fix anonymous function name in cover assignment

raskad:

- Fix anonymous function name in cover assignment


HalidOdat (approved):
Nice Catch! :)

jedel1043 (approved):

PR #3326, opened 2023-09-30
`icu_properties` default features to true

nekevss:
Currently there are errors when trying to run compile or run tests in `boa_parser` without the `compiled_data` feature. So just enabling the feature flag.


jedel1043 (approved):

jedel1043:
Can you add a CI action to check for this?

raskad (approved):

PR #3327, opened 2023-09-30
Fix static class element evaluation order

raskad:

- Fix the evaluation oder of static class fields and static class blocks.


HalidOdat (approved):
Nice Catch! Looks good to me! :)

jedel1043 (approved):
Very nice fix for a really subtle bug :)

PR #3331, opened 2023-10-01
Truncate environment stack on non-caught native error

HalidOdat:
When an native error is thrown and not caught it only truncates the value stack, the environment stack is not changed this causes panics if we run code than access the environment.

For example
```js
with (new Proxy({}, {has: p => false})) {a}
```

Running this code twice causes a panic (`boa file.js file.js`).

```
panicked at 'environment must be declarative', boa_engine/src/vm/opcode/define/mod.rs:114:33
```

### DIFF
--- a/boa_engine/src/vm/code_block.rs
+++ b/boa_engine/src/vm/code_block.rs
@@ -1347,6 +1347,8 @@ impl JsObject {
                     ),
                 );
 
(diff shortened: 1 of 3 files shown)


raskad (approved):
Good catch!

nekevss (approved):

PR #3333, opened 2023-10-02
Improve CI testing

jedel1043:
Just some small improvements in our CI to ensure we're testing everything.

One unfortunate thing I realized is that `cargo clippy -- -D warnings` apparently doesn't work with our current setup, somehow. I _think_ clippy is interpreting our crate global warns as overriding the command argument `-D warnings`, but it seems that this was fixed on https://github.com/rust-lang/rust-clippy/issues/5749. Maybe the bug reappeared for workspaces?

Anyways, I confirmed that the warnings do error with the [new `[lints]` section for `Cargo.toml`](https://github.com/rust-lang/rfcs/pull/3389), so this should be fixed when Rustc 1.74 releases. I'll probably open a new draft PR for that, just to have a reminder to merge the changes at that time. In the meantime, let's just ensure we check the actions for any warnings before merging any PR.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,7 +41,7 @@ jobs:
         uses: codecov/codecov-action@v3
 
   tests:
-    name: Run tests
+    name: Test
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     strategy:
@@ -70,7 +70,7 @@ jobs:
         run: cargo test --doc --profile ci --features annex-b,intl,experimental
 
   msrv:
-    name: Minimum supported Rust version
+    name: MSRV
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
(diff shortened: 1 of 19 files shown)


.github/workflows/rust.yml
```diff
@@ -106,27 +98,41 @@ jobs:
...
       - name: Lint (No features)
-        run: cargo clippy -p boa_engine --no-default-features --all-targets
+        run: cargo workspaces exec cargo clippy --no-default-features --all-targets -- -D warnings
+
       - name: Generate documentation
         run: cargo doc -v --document-private-items --all-features
-      - name: Build
-        run: cargo build --all-targets --quiet --profile ci
+
+      - name: Build (All features)
```
nekevss, line 125: Ooooh, definitely like this better. While making this large of an update to the CI. Should `Misc` be broken apart into more firmly labelled tests? For example, instead of `Misc` there could be tests labelled `Formatting and Lints` and `Crate Builds`.

jedel1043: Definitely! We can't separate build + build fuzz + test examples since those are done that way to save building time, but we can definitely split the rest into smaller jobs.

jedel1043: Oh, apparently the fuzzer CAN be separated, since it triggers a rebuild anyways because it uses the `fuzz` feature only.

nekevss (approved):

HalidOdat (approved):
Looks good to me! :)

PR #3335, opened 2023-10-02
Fix detection of runtime limits for accessors

jedel1043:
This bug was reported on our Discord server, so it doesn't have an issue associated with it.

Fixes runtime limits detection for accessors, since we were detecting those from the `Call` related ops, but property accessors are directly called from Rust code.


HalidOdat (approved):
Looks good to me! :)

boa_engine/src/vm/code_block.rs
```diff
@@ -1014,6 +1014,16 @@ impl JsObject {
         args: &[JsValue],
         context: &mut Context<'_>,
     ) -> JsResult<JsValue> {
+        if context.vm.runtime_limits.recursion_limit() <= context.vm.frames.len() {
```
nekevss, line 1017: Bit of a nit: Can an implementation note be added on this?

nekevss (approved):

PR #3332, opened 2023-10-01
Remove unused class environments

raskad:
Building on #3328 this removes a bunch of unneeded class environments that where more of a hack to get the class name binding in some places. This also enables removing the `class_object` from `FunctionKind`.

PR #3339, opened 2023-10-02
Fix `super()` construction with default parameters

HalidOdat:
Fixes the class `super()` construction error thrown when the derived class has default parameters.

For example (used as a test case):
```JavaScript
class Person {
    constructor(name) {
        this.name = name;
    }
}

class Student extends Person {
    constructor(name = 'unknown') {
        super(name);
    }
}

new Student() // Throws an error that this is not initialized.
```

This PR breaks two test262 tests:

```
test/language/eval-code/direct/arrow-fn-body-cntns-arguments-func-decl-arrow-func-declare-arguments-assign-incl-def-param-arrow-arguments.js
test/language/eval-code/direct/arrow-fn-body-cntns-arguments-var-bind-arrow-func-declare-arguments-assign-incl-def-param-arrow-arguments.js
```

But I _think_ this is due to the following steps that we haven't implement in [`FunctionDeclarationInstantiation`](https://tc39.es/ecma262/#sec-functiondeclarationinstantiation):

```
30. If strict is false, then
    30.a. Let lexEnv be NewDeclarativeEnvironment(varEnv).
    30.b. NOTE: Non-strict functions use a separate Environment Record for top-level lexical
       declarations so that a direct eval can determine whether any var scoped declarations
       introduced by the eval code conflict with pre-existing top-level lexically scoped declarations.
       This is not needed for strict functions because a strict direct eval always
       places all declarations into a new Environment Record.
```

### DIFF
--- a/boa_engine/src/bytecompiler/declarations.rs
+++ b/boa_engine/src/bytecompiler/declarations.rs
@@ -1030,8 +1030,8 @@ impl ByteCompiler<'_, '_> {
             //          visibility of declarations in the function body.
             // b. Let varEnv be NewDeclarativeEnvironment(env).
             // c. Set the VariableEnvironment of calleeContext to varEnv.
-            self.push_compile_environment(true);
-            env_label = Some(self.emit_opcode_with_operand(Opcode::PushFunctionEnvironment));
+            self.push_compile_environment(false);
+            env_label = Some(self.emit_opcode_with_operand(Opcode::PushDeclarativeEnvironment));
 
             // d. Let instantiatedVarNames be a new empty List.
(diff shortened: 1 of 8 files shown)


raskad:
> But I think this is due to the following steps that we haven't implement in [FunctionDeclarationInstantiation](https://tc39.es/ecma262/#sec-functiondeclarationinstantiation):

Yes that is correct. I have a fix for that, but it depends on #3332 and coincidentally this MR too :)

raskad (approved):

jedel1043 (approved):

PR #3340, opened 2023-10-02
Make environments opcodes use varying operands

HalidOdat:
Depends on #3332 

Makes environment push opcodes that take index use varying operands, this required a bit of a refactor so we store environments pushed starting at zero, not in reversed order.


raskad (approved):
Nice change!

jedel1043 (approved):

PR #3341, opened 2023-10-03
Improve highlighter performance

jedel1043:
This PR moves the compilation of the `Regex` used by the highlighter to the `RLHelper` itself, since this avoids having to recompile the regex every time a line is highlighted. Also adds some small improvements and lint fixes.

### DIFF
--- a/boa_cli/src/helper.rs
+++ b/boa_cli/src/helper.rs
@@ -1,6 +1,6 @@
 use colored::{Color, Colorize};
 use phf::{phf_set, Set};
-use regex::{Captures, Regex};
+use regex::{Captures, Regex, Replacer};
 use rustyline::{
     error::ReadlineError,
     highlight::Highlighter,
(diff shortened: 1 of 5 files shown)


HalidOdat (approved):
Nice! looks good to me! :)

boa_cli/src/helper.rs
```diff
@@ -139,52 +139,73 @@ static KEYWORDS: Set<&'static str> = phf_set! {
...
-        let mut coloured = line.to_string();
+struct LineHighlighter {
+    regex: Regex,
+}
 
-        let reg = Regex::new(
+impl LineHighlighter {
+    fn new() -> Self {
+        // Precompiles the regex to avoid creating it again after every highlight
+        let regex = Regex::new(
```
Razican, line 149: Wouldn't this be faster if we used `once_cell` with a static variable? It would only need to be compiled once for the whole execution (or even for multiple executions)

Razican (approved):
Just a small comment that doesn't prevent merging. We might be able to optimize the highlighter even more.

boa_cli/src/helper.rs
jedel1043: I don't think it helps in this case, because `RLHelper` is only initialized once before entering an infinite loop, so it essentially acts as a `OnceCell` already.

Razican: Sounds good, yes :) LGTM!

PR #3348, opened 2023-10-03
Fixed conformance reporting

Razican:
We received reports in our Discord saying that our conformance representation was wrong.

We count a test that needs to be run both in strict and non-strict mode as 2 tests, and if it passes in both modes, we count 2 tests as passing, while if only one of them passes, we count one as passed, one as failed.

The correct interpretation is that a test is only considered passing if both strict and no-strict modes pass (https://github.com/tc39/test262/issues/3871).

It changes the following:

- `TestResult` now contains 2 fields: `strict` and `no_strict`. `strict` is serialized as `s`, and `no_strict` is serialized as `r` (from "result")
- We don't store result strings in the `TestResult` structure, since it was being skipped and never used
- We no longer store feature information, since we were not using it, and the idea behind it was to get conformance per feature, not to have a feature list.
- Minified output by using test / suite names as keys
- Properly calculate conformance & comparisons

This requires a change in the conformance page to adhere to the new result format, and we need to migrate the current conformance results. I'll open a new PR for that.

This PR will fail when doing the result comparison, thoguh, so we might need to first merge the new result format, then this.

### DIFF
--- a/tests/tester/src/exec/mod.rs
+++ b/tests/tester/src/exec/mod.rs
@@ -178,12 +178,12 @@ impl Test {
             result.no_strict =
                 Some(self.run_once(harness, false, verbose, optimizer_options, console));
         } else {
-            if self.flags.contains(TestFlags::STRICT) && !self.flags.contains(TestFlags::RAW) {
+            if self.flags.contains(TestFlags::STRICT) {
                 result.strict =
                     Some(self.run_once(harness, true, verbose, optimizer_options, console));
             }
 
-            if self.flags.contains(TestFlags::NO_STRICT) || self.flags.contains(TestFlags::RAW) {
+            if self.flags.contains(TestFlags::NO_STRICT) {
                 result.no_strict =
                     Some(self.run_once(harness, false, verbose, optimizer_options, console));
             }


boa_tester/src/main.rs
```diff
@@ -785,11 +785,9 @@ struct TestResult {
...
     edition: SpecEdition,
     #[serde(rename = "s", default)]
-    strict: bool,
-    #[serde(skip)]
-    result_text: Box<str>,
-    #[serde(rename = "r")]
-    result: TestOutcomeResult,
+    strict: Option<TestOutcomeResult>,
+    #[serde(rename = "r", default)]
+    no_strict: Option<TestOutcomeResult>,
```
jedel1043, line 790: We could simplify this by making `TestOutcomeResult` something like:
```Rust
#[serde(tag = "t", content = "d")
enum TestOutcomeResult {
    Passed,
    Ignored,
    Failed {
        strict: bool,
        reason: String
    },
    Panic {
        strict: bool,
        reason: String,
    }
}
```
Then, we won't need separate fields for `strict` and `no_strict`, just a single `result`.

Razican: I was thinking about that, and also about the file size & JS performance in the browser. I think that having dictionaries with test/suite names as keys would decrease the size of the JSON around a 15%.

Booleans occupy a bit more. Maybe adding more variants, such as "StrictFailed" and so on could be more performant.

jedel1043: Maybe it's time to transition to a binary serialization format such as [MessagePack](https://msgpack.org/index.html)?

Another idea I had was to use [Postcard](https://crates.io/crates/postcard) and a small wasm glue code to deserialize the postcard and serialize it to JSON, then we would call that function from the conformance page.

Razican: I'll play a bit with those

Razican: I've given this a thought, and it could be a nice enhancement in the future, but I would for now just use JSON until the new website is up and running. What do you think?

jedel1043: Yep, we can revisit this when we finish the site redesign.

Putting that aside, I still think it would be better to have `FailedStrict` and `PanicStrict` for simplicity.

Razican:
I converted it back to draft, because I'm still missing:
 - [ ] ensure that the output result is the same as the new result in the website
 - [ ] ensure that "short" results only contain the last value (except for the main branch)

jedel1043:
Since we will probably do a release before this is finished, and we need to fix our tester before then, I'll open a PR with a simpler version of this that just fixes the bug without changing the data representation too much. We can apply the changes of this PR after that.

Razican:
This has been rebased. It's still missing:
 - [ ] ensure that the output result is the same as the new result in the website
 - [ ] ensure that "short" results only contain the last value (except for the main branch)

jedel1043:
Closing this since it got very updated with our current tester code.

PR #3355, opened 2023-10-05
Cleanup `get_option` and calls to the function

jedel1043:
I saw that #3277 provided better error messages when an option is required, instead of our current method of returning a generic error when `required == true`. Seeing this, I think it's just better to move that check to the caller and handle the `None` case by either throwing or setting a default.

This also removes some type annotations that aren't required.


nekevss (approved):
Looks good to me

HalidOdat (approved):

PR #3369, opened 2023-10-09
Add `Context::create_realm`

johnyob:
This Pull Request fixes/closes issue described in this [discord thread](https://discord.com/channels/595323158140158003/1152240246474412052).

It changes the following:
- Provides a function for creating Realms with the default global bindings.

### DIFF
--- a/boa_engine/src/context/mod.rs
+++ b/boa_engine/src/context/mod.rs
@@ -513,6 +513,17 @@ impl<'host> Context<'host> {
         std::mem::replace(&mut self.realm, realm)
     }
 
+    /// Create a new Realm with the default global bindings.
+    pub fn create_realm(&mut self) -> JsResult<Realm> {
+        let realm = Realm::create(&*self.host_hooks, &self.root_shape);
+
+        let old_realm = self.enter_realm(realm);
+
+        builtins::set_default_global_bindings(self)?;
+
+        Ok(self.enter_realm(old_realm))
+    }
+
     /// Get the [`RootShape`].
     #[inline]
     #[must_use]


boa_engine/src/builtins/mod.rs
```diff
@@ -276,6 +279,21 @@ impl Realm {
...
+        context: &mut Context<'_>,
+    ) -> JsResult<Self> {
+        let realm = Self::create(hooks, root_shape);
+
+        let old_realm = context.enter_realm(realm);
+
+        set_default_global_bindings(context)?;
+
+        Ok(context.enter_realm(old_realm))
+    }
```
jedel1043, line 296: Can this be moved to `Context` instead? It makes better sense if we associate the creation of a `Realm` with the `Context`, like `context.create_realm()?;` That would also allow us to make `Realm::create` crate-only, which avoids the pitfalls of using that API without a proper knowledge of the internals.

johnyob: Implemented in c5efb4118520a8d7fb392ab6dba532994073d806

jedel1043 (approved):
Thanks for the contribution! I think the new API could be used in `boa_tester` for the `createRealm` function, but we can leave it as a future improvement for now.

boa_engine/src/builtins/mod.rs
```diff
@@ -95,11 +95,14 @@ use crate::{
         weak_map::WeakMap,
         weak_set::WeakSet,
     },
-    context::intrinsics::{Intrinsics, StandardConstructor, StandardConstructors},
+    context::{
+        intrinsics::{Intrinsics, StandardConstructor, StandardConstructors},
+        HostHooks,
```
HalidOdat, line 100: Clippy error: "unused imports: `HostHooks`, `RootShape`"

HalidOdat (commented):
Besides the clippy warning, this looks perfect to me! :)

boa_engine/src/builtins/mod.rs
johnyob: Fixed

PR #3370, opened 2023-10-09
`[[HostDefined]]` Refactor

johnyob:
# Context

Currently `HostDefined` doesn't permit you to mutably borrow two objects from the `[[HostDefined]]` field since the `FxHashMap` is wrapped under a `GcRefCell`.

# Description    

This PR instead wraps all `Box<dyn NativeObject>` entries in the `FxHashMap` with a `GcRefCell`, permitting a more granular borrowing scheme for each entry.
    
Additionally, this PR takes the opportunity to provide automatic downcasting on the `insert` and `remove` methods for `HostDefined`.

### DIFF
--- a/boa_engine/src/host_defined.rs
+++ b/boa_engine/src/host_defined.rs
@@ -6,97 +6,78 @@ use rustc_hash::FxHashMap;
 
 use crate::object::NativeObject;
 
-/// Map used to store the host defined objects.
-#[doc(hidden)]
-type HostDefinedMap = FxHashMap<TypeId, Box<dyn NativeObject>>;
-
 /// This represents a `ECMASCript` specification \[`HostDefined`\] field.
 ///
 /// This allows storing types which are mapped by their [`TypeId`].
 #[derive(Default, Trace, Finalize)]
 #[allow(missing_debug_implementations)]
 pub struct HostDefined {
-    state: GcRefCell<HostDefinedMap>,
+    env: FxHashMap<TypeId, GcRefCell<Box<dyn NativeObject>>>,
+}
+
+#[allow(unsafe_op_in_unsafe_fn)]
+unsafe fn downcast_boxed_native_object_unchecked<T: NativeObject>(
+    obj: Box<dyn NativeObject>,
+) -> Box<T> {
+    let raw: *mut dyn NativeObject = Box::into_raw(obj);
+    Box::from_raw(raw as *mut T)
 }
 
 impl HostDefined {
     /// Insert a type into the [`HostDefined`].
-    ///
-    /// # Panics
-    ///
-    /// Panics if [`HostDefined`] field is borrowed.
     #[track_caller]
-    pub fn insert_default<T: NativeObject + Default>(&self) -> Option<Box<dyn NativeObject>> {
-        self.state
-            .borrow_mut()
-            .insert(TypeId::of::<T>(), Box::<T>::default())
+    pub fn insert_default<T: NativeObject + Default>(&mut self) -> Option<Box<T>> {
+        self.env
+            .insert(TypeId::of::<T>(), GcRefCell::new(Box::new(T::default())))
+            .map(|obj| unsafe { downcast_boxed_native_object_unchecked(obj.into_inner()) })
     }
 
     /// Insert a type into the [`HostDefined`].
-    ///
-    /// # Panics
-    ///
-    /// Panics if [`HostDefined`] field is borrowed.
     #[track_caller]
-    pub fn insert<T: NativeObject>(&self, value: T) -> Option<Box<dyn NativeObject>> {
-        self.state
-            .borrow_mut()
-            .insert(TypeId::of::<T>(), Box::new(value))
+    pub fn insert<T: NativeObject>(&mut self, value: T) -> Option<Box<T>> {
+        self.env
+            .insert(TypeId::of::<T>(), GcRefCell::new(Box::new(value)))
+            .map(|obj| unsafe { downcast_boxed_native_object_unchecked(obj.into_inner()) })
     }
 
     /// Check if the [`HostDefined`] has type T.
-    ///
-    /// # Panics
-    ///
-    /// Panics if [`HostDefined`] field is borrowed mutably.
     #[track_caller]
     pub fn has<T: NativeObject>(&self) -> bool {
-        self.state.borrow().contains_key(&TypeId::of::<T>())
+        self.env.contains_key(&TypeId::of::<T>())
     }
 
     /// Remove type T from [`HostDefined`], if it exists.
     ///
     /// Returns [`Some`] with the object if it exits, [`None`] otherwise.
-    ///
-    /// # Panics
-    ///
-    /// Panics if [`HostDefined`] field is borrowed.
     #[track_caller]
-    pub fn remove<T: NativeObject>(&self) -> Option<Box<dyn NativeObject>> {
-        self.state.borrow_mut().remove(&TypeId::of::<T>())
+    pub fn remove<T: NativeObject>(&mut self) -> Option<Box<T>> {
+        self.env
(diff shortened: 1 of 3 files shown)


jedel1043 (changes requested):
Downcasting on `insert` and `delete` looks great! However, I'm having second thoughts about having that amount of granularity on the API.

My problem comes from the thought that we'd be worsening the API for users that just want to store one or two types.

On the other hand, if a user wants to mutably borrow two separate types from the map, they can just put both in a wrapper type, then insert that instead, which should allow the same use case. Another workaround would be to just borrow the map multiple times. That should be a lot easier to handle than having to deal with 2 levels of borrowing
(borrowed map + borrowed type, borrowed map + mutably borrowed type, mutably borrowed map + borrowed type, mutably borrowed map + mutably borrowed type)

johnyob:
> My problem comes from the thought that we'd be worsening the API for users that just want to store one or two types.
>
> On the other hand, if a user wants to mutably borrow two separate types from the map, they can just put both in a wrapper type, then insert that instead, which should allow the same use case. Another workaround would be to just borrow the map multiple times. That should be a lot easier to handle than having to deal with 2 levels of borrowing
(borrowed map + borrowed type, borrowed map + mutably borrowed type, mutably borrowed map + borrowed type, mutably borrowed map + mutably borrowed type)

The primary pain point (for us) with the current implementation is that we cannot mutably borrow two host-defined objects (defined by separate extensions) in a single function. 

The possible pain points with the proposed implementation are:
1. Lifetimes of all host-defined objects are bounded by the lifetime of the `GcRef`/`GcRefMut` you obtain by borrowing the `host_defined` map. 

   In the context of writing a runtime, we've not encountered many issues with this. The solution to this is usually to `.borrow()` the `host_defined` map at the top of your native function. However, you do need to carefully scope the borrowed reference, or explicitly `drop` the reference (as shown in the example) when interleaving between Rust code and JavaScript code if the reference is _mutable_. 
 
 2. The 2 levels of borrowing. 

    I agree that this definitely comes with a mental overhead wrt to the borrowing semantics. But it can often result in clearer code when the semantics of borrowing are more explicit. 
 
     > Another workaround would be to just borrow the map multiple times
     
     This isn't possible if you want mutable references to _two_ (or more) entries in the host-defined map. 
  
I think with perhaps clearer documentation on the borrowing semantics (something that was still lacking with the current implementation wrt to borrowing 2 or more mutable entries in host-defined), this change could still be seen as an improvement over the current API.

jedel1043:
> This isn't possible if you want mutable references to two (or more) entries in the host-defined map.

Yeah, and I'd argue this is a design problem, not an API problem. This would still happen if this was a simple `HashMap` without any relation to Boa itself. Usually, this is resolved by shortening the lifetimes of the references or by ensuring only one reference is used at once.

Leaving that aside, I think a better solution to this would be to expose the [`get_many_mut`](https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.get_many_mut) method, which should allow to do this without having to deal with the complexity of multiple `RefCells`. This will require using `hashbrown::HashMap` instead of `std::HashMap` though, but we already have the dependency on our tree thanks to `WeakMap`, so no problems from pulling it into `boa_engine` itself.

johnyob:
(Prefixing my response: I'm still fairly new to Rust so please forgive my ignorance to existing libraries, common design patterns, etc)

> This would still happen if this was a simple HashMap without any relation to Boa itself

Would it? In Rust, what I'd expect is to be given a `&'a mut HashMap` for which I can get the mutable references for two (or more) entires, each of which has a lifetime `'ai` bounded by `'a`. This is what this the approach that this PR provides (but obviously with some foot guns due to the use of `GcRefCell`s)

> Leaving that aside, I think a better solution to this would be to expose the [get_many_mut](https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.get_many_mut) method, which should allow this without having to deal with the complexity of multiple RefCells. This will require using hashbrown::HashMap instead of std::HashMap though, but we already have the dependency on our tree thanks to WeakMap, so no problems from pulling it into boa_engine itself.

But is this possible? As I understand currently, the interface we expose with `HostDefined` is one of a _dependently-typed_ map, where the type of the entry depends on the type of the key. From what I can deduce from this `get_many_mut` method, we're not able to provide the same interface since we'd have to simply pass an array of `TypeId`s in an recieve an array of `NativeObject`s with the user doing the necessary downcasting.

jedel1043:
> Would it? In Rust, what I'd expect is to be given a `&'a mut HashMap` for which I can get the mutable references for two (or more) entires, each of which has a lifetime `'ai` bounded by `'a`. This is what this the approach that this PR provides (but obviously with some foot guns due to the use of `GcRefCell`s)

Rust doesn't allow borrowing two mutable entries to the hashmap at the same time, even if the keys aren't equal.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=958b5889ab29439c6be993473c342b49

> But is this possible? As I understand currently, the interface we expose with `HostDefined` is one of a _dependently-typed_ map, where the type of the entry depends on the type of the key. From what I can deduce from this `get_many_mut` method, we're not able to provide the same interface since we'd have to simply pass an array of `TypeId`s in an recieve an array of `NativeObject`s with the user doing the necessary downcasting.

I think this is solvable with a `NativeTuple` trait that has a `as_typeid_array` method that returns the type ids of each one of its members. We can use that to query using a tuple instead, which would result in an API such as:

```rust

use std::any::TypeId;

trait NativeTuple: Sealed {
    type BorrowMutTuple<'a>;
    fn as_type_ids() -> Vec<TypeId>; 
}

// Impls for tuples of size 1 to N

impl HostDefined {
    // More methods
    
    fn get_many_mut<Tup: NativeTuple>(&mut self) -> Tup::BorrowMutTuple<'a> {
        // Impl
    }
}
```

Or something like that, which I think it's pretty feasible. I'll try to come up with a prototype today.

johnyob:
> Rust doesn't allow borrowing two mutable entries to the hashmap at the same time, even if the keys aren't equal.

Thanks for letting me know, I forgot that `HashMap` doesn't support interior mutability an entires 😅 I've had too many `GcRefCell`s in my life recently

> I think this is solvable with a NativeTuple trait that has a `as_typeid_array` method that returns the type ids of each one of its members. We can use that to query using a tuple instead, which would result in an API such as:

Whoa, this is really neat. If you're strapped for time, I'd be happy to try and prototype it myself!

jedel1043:
> Whoa, this is really neat. If you're strapped for time, I'd be happy to try and prototype it myself! (Though I may up asking a lot of stupid questions over discord)

Sure, you can give it a go! It'll be a bit hard because of limitations from rustc around const generics,
but I think I have the trickiest part solved:
```rust
pub fn get_many_mut<Tup, const SIZE: usize>(&mut self) -> Option<Tup::BorrowedMutTuple<'_>>
    where
        Tup: Tuple<SIZE>,
    {
        let ids = Tup::as_type_ids();
        let refs: [&TypeId; SIZE] = ids.iter().collect::<Vec<_>>().try_into().expect("tuple should be of size `SIZE`");
        self.types
            .get_many_mut(refs)
            .and_then(|anys| Tup::mut_refs_from_anys(anys))
    }

pub trait Tuple<const SIZE: usize> {
    type BorrowedMutTuple<'a>;

    fn as_type_ids() -> Vec<TypeId>;

    fn mut_refs_from_anys(
        anys: [&'_ mut Box<dyn NativeObject>; SIZE],
    ) -> Option<Self::BorrowedMutTuple<'_>>;
}
```

Ideally we'd have the `SIZE` parameter as an associated const, but that's not possible at the moment (https://github.com/rust-lang/rust/issues/60551). We can use that workaround, then lift the limitation when the compiler supports it in the future.

johnyob:
Closing in favour of #3460

PR #3374, opened 2023-10-11
Fix `Function.prototype.toString()`

HalidOdat:
The remaining failing test262 test [`test/built-ins/Function/prototype/toString/method-computed-property-name.js`][test] is failing due to step 2 not being implemented. We fallback to native function to string representation.

[test]: https://github.com/tc39/test262/blob/f0c16faee2afe3cfe4d6e42aff743964e4ea9122/test/built-ins/Function/prototype/toString/method-computed-property-name.js


Razican (approved):
Good catch! LGTM

jedel1043 (approved):

PR #3375, opened 2023-10-11
Check `eval` realm before call

HalidOdat:
Fixes failing test262 test [`test/language/expressions/call/eval-realm-indirect.js`](https://github.com/tc39/test262/blob/f0c16faee2afe3cfe4d6e42aff743964e4ea9122/test/language/expressions/call/eval-realm-indirect.js)


raskad (approved):
Nice catch

jedel1043 (approved):
Nice catch!

PR #3376, opened 2023-10-12
Update feature flags to specific feature flag

nekevss:
This Pull Request addresses the feature flag discussion from #3277. 

It changes the following:

- Adds a `temporal` and `promise-resolvers` feature flag
- Sets experimental to be dependent on the feature flags

### DIFF
--- a/boa_ast/Cargo.toml
+++ b/boa_ast/Cargo.toml
@@ -13,7 +13,7 @@ rust-version.workspace = true
 [features]
 serde = ["dep:serde", "boa_interner/serde", "bitflags/serde", "num-bigint/serde"]
 arbitrary = ["dep:arbitrary", "boa_interner/arbitrary", "num-bigint/arbitrary"]
-experimental = []
+temporal = []
 
 [dependencies]
 boa_interner.workspace = true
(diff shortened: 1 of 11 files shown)


HalidOdat (approved):
Looks good to me! :)

jedel1043 (changes requested):
I don't think we need the granularity of `promise-resolvers`. For `temporal` is understandable because we're planning to gate every builtin behind a feature, but  additions to existing builtins can exist in `experimental` since they're simpler APIs.

jedel1043:
Just for the record, I do think we need a mechanism to enable experimental features individually. However, I think this can be accomplished in a backwards compatible way by using directives instead of Rust features. This would be pretty similar to how Rust handles nightly features:

- We have an array on some part of the engine with a list of the available experimental features and its current stage.
- Using directives, a user can select which experimental features they want to use, but only if the `experimental` feature is enabled.
- At execution time, we check the directives of a script/module, and if it has the corresponding feature, we inject it into the global object then remove it after execution finishes.
- Finally, for users who want to expose some features for all scripts, we can take a `features` argument on `Context` creation and that'll take care of injecting the additional features without removing them.

jedel1043 (approved):

PR #3385, opened 2023-10-14
Feature `get/set $boa.limits.stack`

HalidOdat:
~~Depends #3295~~

### DIFF
--- a/boa_cli/src/debug/limits.rs
+++ b/boa_cli/src/debug/limits.rs
@@ -16,6 +16,22 @@ fn set_loop(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResul
     Ok(JsValue::undefined())
 }
 
+fn get_stack(_: &JsValue, _: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+    let max = context.runtime_limits().stack_size_limit();
+    Ok(JsValue::from(max))
+}
+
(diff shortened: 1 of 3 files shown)


boa_cli/src/debug/limits.rs
```diff
@@ -16,6 +16,22 @@ fn set_loop(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResul
...
+fn set_stack(_: &JsValue, args: &[JsValue], context: &mut Context<'_>) -> JsResult<JsValue> {
+    let value = args.get_or_undefined(0).to_length(context)?;
+    let Ok(value) = value.try_into() else {
+        return Err(JsNativeError::range()
+            .with_message(format!("Argument {value} greater than usize::MAX"))
+            .into());
+    };
+    context.runtime_limits_mut().set_stack_size_limit(value);
+    Ok(JsValue::undefined())
+}
```
jedel1043, line 33: Would probably be better to make the access something like `stack.max_size` instead. That way we can reserve other properties from `stack` like `stack.trace` or `stack.caller`.

HalidOdat: Maybe, I grouped it based on the limits (recursion, loop iteration, and now stack), for now I think it's fine since we only have the limit get/set, we can change it in the future if we need, it's not a breaking change since this is all internal tooling.

Razican (approved):

jedel1043 (approved):

PR #3399, opened 2023-10-19
Fix compilation for targets without `AtomicU64`

### LINKED ISSUE #2599: MIPS 32-bit target without Atomic{I,U}64
https://github.com/Borber/seam/actions/runs/4154490037/jobs/7186917428#step:5:227 

How should I fix this problem

jedel1043:
Closes #2599 
Also addresses the clippy lint that we had disabled temporarily.

### DIFF
--- a/boa_engine/src/lib.rs
+++ b/boa_engine/src/lib.rs
@@ -123,7 +123,6 @@
 
     // Add temporarily - Needs addressing
     clippy::missing_panics_doc,
-    clippy::arc_with_non_send_sync,
 )]
 
 extern crate static_assertions as sa;
--- a/boa_engine/src/lib.rs
+++ b/boa_engine/src/lib.rs
@@ -125,6 +125,9 @@
     clippy::missing_panics_doc,
 )]
 
+#[cfg(not(target_has_atomic = "ptr"))]
+compile_error!("Boa requires a lock free `AtomicUsize` in order to work properly.");
+
 extern crate static_assertions as sa;
 
 pub mod bigint;
--- a/boa_engine/src/symbol.rs
+++ b/boa_engine/src/symbol.rs
@@ -33,23 +33,22 @@ use num_enum::{IntoPrimitive, TryFromPrimitive};
 
 use std::{
     hash::{Hash, Hasher},
-    sync::{
-        atomic::{AtomicU64, Ordering},
-        Arc,
-    },
+    sync::{atomic::Ordering, Arc},
 };
+
+use portable_atomic::AtomicUsize;
 
 /// Reserved number of symbols.
 ///
 /// This is where the well known symbol live
 /// and internal engine symbols.
-const RESERVED_SYMBOL_HASHES: u64 = 127;
+const RESERVED_SYMBOL_HASHES: usize = 127;
 
-fn get_id() -> Option<u64> {
+fn get_id() -> Option<usize> {
     // Symbol hash.
     //
     // For now this is an incremented u64 number.
-    static SYMBOL_HASH_COUNT: AtomicU64 = AtomicU64::new(RESERVED_SYMBOL_HASHES + 1);
+    static SYMBOL_HASH_COUNT: AtomicUsize = AtomicUsize::new(RESERVED_SYMBOL_HASHES + 1);
 
     SYMBOL_HASH_COUNT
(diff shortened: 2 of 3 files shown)


boa_engine/src/symbol.rs
```diff
@@ -33,23 +33,22 @@ use num_enum::{IntoPrimitive, TryFromPrimitive};
...
 use std::{
     hash::{Hash, Hasher},
-    sync::{
-        atomic::{AtomicU64, Ordering},
-        Arc,
-    },
+    sync::{atomic::Ordering, Arc},
 };
 
+use portable_atomic::AtomicUsize;
```
Razican, line 39: Wouldn't `std::sync::atomic::AtomicUsize` work?

Razican (approved):
Thanks!!

boa_engine/src/symbol.rs
```diff
@@ -34,7 +34,7 @@ use num_enum::{IntoPrimitive, TryFromPrimitive};
 use std::{
     hash::{Hash, Hasher},
     sync::{
-        atomic::{AtomicU64, Ordering},
+        atomic::{AtomicUsize, Ordering},
```
HalidOdat, line 37: The problem with `usize` is that on 32-bit architectures this is a `u32`, which isn't big enough, we can reach the max in about `15s`. With `u64` will take more than a century.

This effects the whole program, so all threads can't create `JsSymbol`s

jedel1043: So we use `portable_atomic::AtomicU64`? I considered that alternative, but opted for this because `AtomicU64` is a lot slower on platforms without 8 byte atomics, since they have to get a global lock to do operations.

jedel1043: I think the question would be: do we want to prioritise flexibility or speed?

jedel1043: Thinking about it more thoroughly, I don't think we need to pick one of them, we can just leave the decision to our users!

A "fast-symbols" feature that selects `AtomicUsize` if it is enabled and fallbacks to `AtomicU64` otherwise should be enough. What do you think?

jedel1043: You could also argue that symbols are not as prevalent in most JS programs, so it would be pretty rare to have `u32::MAX` symbols alive at the same time.

raskad: > I think the question would be: do we want to prioritise flexibility of use or speed?

I'm not 100% into the topic at the moment, but I'd guess 8 byte atomics are common on most platforms nowdays? I would say we go for the flexibility right now as  the speed cost would not affect most users. If we notice it to be a problem we could always add optimizations or even give the users the option to choose, right?

jedel1043: > > I think the question would be: do we want to prioritise flexibility of use or speed?
> 
> I'm not 100% into the topic at the moment, but I'd guess 8 byte atomics are common on most platforms nowdays? I would say we go for the flexibility right now as  the speed cost would not affect most users. If we notice it to be a problem we could always add optimizations or even give the users the option to choose, right?

Ok, since it seems like both of you prefer the `AtomicU64` option, I'll prioritise flexibility then. I think it shouldn't affect performance for most programs because most of them don't create a lot of symbols consecutively.

HalidOdat: Yes, lets continue with `AtomicU64`, if we need to we can change it in the future :)

boa_engine/src/lib.rs
```diff
@@ -123,9 +123,11 @@
 
     // Add temporarily - Needs addressing
     clippy::missing_panics_doc,
-    clippy::arc_with_non_send_sync,
 )]
 
+#[cfg(not(target_has_atomic = "ptr"))]
+compile_error!("Boa requires a lock free `AtomicUsize` in order to work properly.");
+
```
HalidOdat, line 130: We don't need this, right?

HalidOdat (approved):
Besides the comment below, this looks good to me! :)

boa_engine/src/lib.rs
HalidOdat: I see, forgot that `Arc` uses `AtomicUsize` internally 😅 . thanks for the clarification!

HalidOdat (approved):
Looks perfect to me! :)

PR #3401, opened 2023-10-19
Reduce spend budget execution table size

HalidOdat:
This reduces the `SPEND_FNS` from `6144 bytes` to `2048` + `256` for `COSTS` table. In total `2304` bytes.


jedel1043 (approved):
Thanks for the improvement!

PR #3406, opened 2023-10-21
Fix Array.join when the array contains itself

ahaoboy:
The documentation doesn't specify how the join function should handle cases where the array contains itself, but other engines treat it as an empty string.

```
let b = []
b[0] = b;
b[1] = b;
b[2] = b;

// expect output: ",,"
console.log(b.join(','))
```


Razican (commented):
It seems there are no Test262 tests for this behavior, so we should add a Rust test

jedel1043 (approved):
Thanks for the contribution!

HalidOdat (approved):
Looks good to me! :)

PR #3410, opened 2023-10-23
Disallow changing type of already created objects

jedel1043:
This allows us to cleanup some of our `Js` method wrappers for built-ins, like `JsPromise` which I've already cleaned up in this PR.

### DIFF
--- a/boa_engine/src/builtins/regexp/mod.rs
+++ b/boa_engine/src/builtins/regexp/mod.rs
@@ -295,6 +295,10 @@ impl RegExp {
     }
 
     /// Compiles a `RegExp` from the provided pattern and flags.
+    ///
+    /// Equivalent to the beginning of [`RegExpInitialize ( obj, pattern, flags )`][spec]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-regexpinitialize
     fn compile_native_regexp(
         pattern: &JsValue,
         flags: &JsValue,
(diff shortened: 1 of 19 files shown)


boa_engine/src/builtins/regexp/mod.rs
```diff
@@ -376,27 +338,53 @@ impl RegExp {
...
+            template.create(data, vec![0.into()])
+        } else {
+            context
+                .intrinsics()
+                .templates()
+                .regexp()
+                .create(data, vec![0.into()])
+        };
 
         // 16. Return obj.
```
Razican, line 389: It seems there are 2 16th steps somehow 😅

boa_engine/src/object/builtins/jspromise.rs
```diff
@@ -502,44 +471,42 @@ impl JsPromise {
...
     ///
     /// [`Promise.prototype.then`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then
     #[inline]
+    #[allow(clippy::return_self_not_must_use)] // Could just be used to add handlers on an existing promise
     pub fn then(
         &self,
-        on_fulfilled: Option<JsFunction>,
-        on_rejected: Option<JsFunction>,
+        on_fulfilled: Option<NativeFunction>,
+        on_rejected: Option<NativeFunction>,
```
HalidOdat, line 499: Why constrain it to only accept `NativeFunction`? If it's only to reduce the `FunctionObjectBuilder` then we could implement a conveniance method on `NativeFunction`, (maybe `to_js_function(realm)`) that would convert it.

jedel1043: Nice idea! I was just cleaning up the API since we mostly use only `NativeFunction`, but having a `to_fn_object` seems a lot more flexible and should also clean up our API :)

HalidOdat (approved):
Looks good to me! :)

raskad (approved):

PR #3413, opened 2023-10-24
Merge `CodeBlock` constant pools

HalidOdat:
This PR merges the `literals`, `names`, `functions`, `compile_environments` constant pools into one constant pool called `constants`


boa_engine/src/vm/code_block.rs
```diff
@@ -135,29 +150,15 @@ pub struct CodeBlock {
     /// Bytecode
     pub(crate) bytecode: Box<[u8]>,
 
-    /// Literals
-    pub(crate) literals: Box<[JsValue]>,
-
-    /// Property field names and private names `[[description]]`s.
-    pub(crate) names: Box<[JsString]>,
+    pub(crate) contants: ThinVec<Constant>,
```
jedel1043, line 153:
```suggestion
    pub(crate) constants: ThinVec<Constant>,
```

boa_engine/src/vm/code_block.rs
```diff
@@ -169,17 +170,14 @@ impl CodeBlock {
         flags.set(CodeBlockFlags::STRICT, strict);
         Self {
             bytecode: Box::default(),
-            literals: Box::default(),
-            names: Box::default(),
+            contants: ThinVec::default(),
```
jedel1043, line 173:
```suggestion
            constants: ThinVec::default(),
```

boa_engine/src/vm/code_block.rs
```diff
@@ -252,6 +250,33 @@ impl CodeBlock {
             .rev()
             .find(|(_, handler)| handler.contains(pc))
     }
+
+    pub(crate) fn constant_string_expect(&self, index: usize) -> JsString {
+        if let Constant::String(value) = &self.contants[index] {
+            return value.clone();
+        }
+
+        unreachable!("there should be a string constant at index {index}")
```
jedel1043, line 259: I'd change these for panics. `unreachable` gives the idea that the function has asserted that `index` points to a `Constant::String` and every other type of constant is impossible to reach, which isn't really true.

boa_engine/src/vm/flowgraph/mod.rs
```diff
@@ -526,9 +526,9 @@ impl CodeBlock {
...
         }
 
-        for function in self.functions.as_ref() {
-            let subgraph = graph.subgraph(String::new());
-            function.to_graph(interner, subgraph);
-        }
+        // for function in self.functions.as_ref() {
+        //     let subgraph = graph.subgraph(String::new());
+        //     function.to_graph(interner, subgraph);
+        // }
```
jedel1043, line 532: Missing deletion? If not, it would be good to add a `TODO` explaining why this is commented out for now.

boa_engine/src/vm/code_block.rs
```diff
@@ -252,6 +250,33 @@ impl CodeBlock {
             .rev()
             .find(|(_, handler)| handler.contains(pc))
     }
+
+    pub(crate) fn constant_string_expect(&self, index: usize) -> JsString {
+        if let Constant::String(value) = &self.constants[index] {
+            return value.clone();
+        }
+
+        panic!("there should be a string constant at index {index}")
```
nekevss, line 259: Can there be `Panic` docs added to these functions so they pass future lints.

Also, a little nitpicky aside, I don't dislike naming the functions as `_expect`. I do think it's clear, but would it be worthwhile being more explicit with `_or_panic`? I'm fine with either/or but wanted to at least mention it.

nekevss (changes requested):
Overall, this looks pretty good to me. Just had one thing regarding docs.

Definitely interested to see the new codeblock trace! 😄

boa_engine/src/vm/code_block.rs
```diff
@@ -252,6 +250,33 @@ impl CodeBlock {
...
+    }
```
raskad, line 260: A very minor optimization would be to do a `get` on the slice instead of directly accessing it via the array index notation. Then do `if let Some(Constant::String(value))`.
That way we avoid the panic branch on the bounds check and only have one panic in this function.

raskad (approved):
Look good. I only have one very minor suggestion.

boa_engine/src/vm/code_block.rs
HalidOdat: Maybe we should remove the postfix? I only added it because we have `.resolve_expect()` and `.environment_expect()` to be consistent with the naming. But there isn't a rust guideline neither does the `std` lib do this.

jedel1043: +1 to the suggestion about removing the postfix. It's not common for Rust APIs to name panicking functions with a suffix. It's much more common to panic by default, then offer `try_*` variants for non-panicking functions.

nekevss: Yeah, I was sort of thinking about the `try_*` naming convention when I was looking at it. It's much more common for us to return a `JsResult` throughout the engine as a default. So instead of adding `try_*` everywhere, adding some sort of naming convention on the panicking functions in our case might be a bit better clarity-wise.

That being said, sticking to the typical naming convention and removing the suffix is also a valid option.

nekevss (approved):
This looks good to me 😄 Thanks for adding the panic docs!

PR #3421, opened 2023-10-25
Gassification: Expose instruction limits

sam-finch-tezos:
Exposes the instruction counter and fix a bug allowing unreachable behaviour.

* If the "fuzz" feature is enabled the `instructions_remaining` counter is a public property.
* The `NoInstructionsRemaining` error can now be converted into a javascript error, preventing boa from panicking if the error is thrown async


sam-finch-tezos:
Sorry, was meant for a private fork!

PR #3424, opened 2023-10-26
`Context` independent `CodeBlock`s

### LINKED ISSUE #2626: Reusing `Codeblock`s between `Context`s is not property supported.

HalidOdat:
Fixes/Closes #2626

This allows us to share `CodeBlock`s between `Context`, which makes functions sharable between `Context`s.


HalidOdat:
Ideally we would make the interner store `JsString`s and when we `.resolve_expect()` it would return a `JsString`, this would simplify our compiler logic.

jedel1043:
Does this affect performance in any way? Just thinking that copying `Identifier` was pretty cheap, but cloning a `JsString` involves incrementing a `Cell`.

The other alternative would be to have an API similar to V8's, where scripts are "linked" to contexts, and you can unlink them and relink them to other contexts if you need to.

HalidOdat:
> Does this affect performance in any way?

Yes, there is a small perf. increase.

`main`:

```
PROGRESS Richards
RESULT Richards 35.0
PROGRESS DeltaBlue
RESULT DeltaBlue 39.8
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 54.5
PROGRESS RayTrace
RESULT RayTrace 135
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 114
```

`PR`:

```
PROGRESS Richards
RESULT Richards 36.2
PROGRESS DeltaBlue
RESULT DeltaBlue 41.9
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 55.5
PROGRESS RayTrace
RESULT RayTrace 140
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 116
```

> Just thinking that copying `Identifier` was pretty cheap, but cloning a `JsString` involves incrementing a `Cell`.

I'm fine with keeping `Identifier` in AST, we should only change the interners internal string store to be `JsString` instead of `InternedStr<u16>`, it would remove the back and forth conversions. This would require either moving `JsString` to a separate crate (probably the better choice) or in `boa_interner` crate.

> The other alternative would be to have an API similar to V8's, where scripts are "linked" to contexts, and you can unlink them and relink them to other contexts if you need to.

Sounds interesting! Is there any resource that describes this? :)

jedel1043:
> Sounds interesting! Is there any resource that describes this? :)

I don't think V8 has an example on how to use that API, but `Script` is split into an `UnboundScript` (the data that doesn't depend on the current script) and the rest of the context-dependent data. Then, `UnboundScript` has a method to bind the script to a `Context`:
https://github.com/v8/v8/blob/2465184635ae2c2237ea13afe3bb395eae7f94bd/include/v8-script.h#L62-L65

I think it's slightly more flexible, since an `UnboundScript` can be shared between many `Context`s, and you can always recover the shared `UnboundScript` from the `Script`:
https://github.com/v8/v8/blob/7fdc2728f2114b083ebbd24710a32fe4574c0c57/src/api/api.cc#L2153-L2159

jedel1043:
@jedel1043 and @raskad will review this.

boa_engine/src/builtins/eval/mod.rs
```diff
@@ -228,7 +228,7 @@ impl Eval {
         let mut var_env = var_environment.compile_env();
 
         let mut compiler = ByteCompiler::new(
-            Sym::MAIN,
+            Sym::MAIN.to_js_string(context.interner()),
```
raskad, line 231: Is there a reason to not use the `js_string!` macro directly for staticly known strings? For example in this case `js_string!("<main>")`.

Same goes for some other places in the PR where this is the case.

raskad (approved):
Really nice, I just got one comment.

PR #3428, opened 2023-10-27
example: add deno console

ahaoboy:
This Pull Request implements the console.log function using Deno.

 This is an incomplete implementation, but it covers most use cases. I don't know how to implement some complex functions like the getProxyDetails, getPromiseDetails  in Boa, so the output of promises is always 'Promise<unknown>.

boa_examples/scripts/deno/01_console.js copy from  https://github.com/denoland/deno/blob/main/ext/console/01_console.js
boa_examples/scripts/deno/00_core.js is a polyfill for deno.primordials

```js
console.log(1, 'a', false, true, Promise.resolve(1), Symbol(1))
console.log(new Set([1, 2, 3, 4]))
console.log(new Map([[1, 2], ['a', 'b'], [false, true]]))
let a = [1]
a[1] = a;
console.log(a)
let b = {a: [1,2,'a']}
b['b'] = b;
console.log(b)
console.log(new Error("console error"))
```
## deno.console
image

## boa_runtime::Console;
image

### DIFF
--- a/boa_examples/scripts/console.js
+++ b/boa_examples/scripts/console.js
@@ -1,0 +1,22 @@
+console.log(1, "a", false, true, Promise.resolve(1), Symbol(1));
+
+console.log(new Set([1, 2, 3, 4]));
+
+console.log(
+  new Map([
+    [1, 2],
+    ["a", "b"],
+    [false, true],
+  ]),
+);
+
+let a = [1];
+a[1] = a;
+console.log(a);
+
+let b = { a: [1, 2, "a"] };
+b["b"] = b;
+
+console.log(b);
+
+console.log(new Error("console error"));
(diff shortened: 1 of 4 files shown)


jedel1043:
Uhhhh, I don't think this should be an example. This is a fully fledged implementation of `console.log`, and examples are supposed to explain how to use a certain API in a simple, well documented way.

That being said, maybe we could reuse this to prettify our `console.log` implementation?

ahaoboy:
> Uhhhh, I don't think this should be an example. This is a fully fledged implementation of `console.log`, and examples are supposed to explain how to use a certain API in a simple, well documented way.
> 
> That being said, maybe we could reuse this to prettify our `console.log` implementation?

As an example, it's indeed a bit too extensive 😂. Originally, I intended to implement the complete console.log function purely in rust, but it's too complex. If you think that boa can reuse this log function, then I'll open a new PR to replace boa_runtime\src\console.

jedel1043:
> If you think that boa can reuse this log function, then I'll open a new PR to replace boa_runtime\src\console.

That would be wonderful!

jedel1043:
Closing in favour of #3468

PR #3429, opened 2023-10-27
Fixed clippy lints

Razican:
It turns out we were getting many clippy warnings, but this was not making the CI fail. This enhances the CI and fixes all the lints.


Razican:
It seems I was not using an up to date branch

PR #3432, opened 2023-10-29
Move `arguments` object creation to bytecode

HalidOdat:
Moves creation of `arguments` object creation from function internal methods, this allows us to remove the checks for `arguments` object creation for functions that don't create it and would make #3194 easier since we can have `arguments` be moves into the stack and accessed as a fast local, instead of accessing it through environments.

While doing this I discovered a bug that happens when we create a function in `eval`, which was creating the functions and initializing the function, before the `eval` environment's `PushDeclarativeEnvironment` is executed. Fixed by delaying creation of function by adding the `CreateGlobalFunctionBinding` opcode.


jedel1043 (approved):
Thanks! I think this also puts us a step closer to separating codeblocks from the context, so that's good.

raskad (approved):

PR #3440, opened 2023-10-30
Move `FunctionKind` to `CodeBlock`

HalidOdat:
Depends on #3439 

Since the kind of the function is known at compile time, we can move it to `CodeBlock`'s bitflags. This change allows us to simplify the logic of function creation. Remove `GetAsyncArrowFunction`, `GetFunctionAsync` and `GetGeneratorAsync` opcodes.


raskad (approved):
Very nice refactor!

nekevss (approved):
Looks fantastic! 😄 Really nice refactor!

PR #3450, opened 2023-11-03
Implement latin1 encoded `JsString`s

### LINKED ISSUE #3097: ASCII encoding `JsString`s
Currently `JsString` can only be `u16` even if the contents can fit into the ASCII range this requires more memory to store strings.

It would be better to have a flag in `JsString`, to know if the `data` content is an array of `u8` or `u16`.

HalidOdat:
This fixes #3097 

This memory optimization is implemented by both [spidermonkey](https://blog.mozilla.org/javascript/2014/07/21/slimmer-and-faster-javascript-strings-in-firefox/) and [v8](https://v8docs.nodesource.com/node-7.10/d2/db3/classv8_1_1_string.html#a60f87555f2f7a70a3fbd502019c5a41e).

This PR encodes strings that can be represented as [latin1](https://symbl.cc/en/unicode-table/#latin-1-supplement) as a byte array, instead of `u16` array, this cuts latin1 strings size by `2x`.

~~This may also have some interesting optimization that could be applied when we know that the string is in the ASCII range.~~ Decided to not preserve asciiness though it can work with asciiness preserved it would complicate other string optimizations. AFIK v8 does not preserve the asciiness. Preserving the asciiness could lead to some interesting optimizations but at the cost of a lot of complexity which from my testing doesn't seem to be worth it.


boa_engine/src/string/mod.rs
```diff
@@ -260,8 +362,45 @@ impl JsString {
...
+                            data = data.cast::<u16>().add(count).cast::<u8>();
+                        }
+                        (
+                            true,
+                            JsStringSliceVariant::U8NonAscii(..)
+                            | JsStringSliceVariant::U16NonAscii(_),
+                        ) => {
+                            unreachable!()
+                        }
+                    }
```
jedel1043, line 403: I think trying to recover the "asciiness" of an UTF-16 string gets too unwieldy, making it difficult to follow implementations like this. In my opinion, we should just offer two variants: ASCII and UTF-16, and if a string becomes UTF-16, it should just stay that way from that point. 

This also preemptively converts the strings to UTF-16, which I think is a plus: if the user converted some strings to UTF-16, they are probably using non-ASCII characters in some part of the code, which means they'll probably keep using UTF-16 chars in the future. Converting back to UTF-8 just does unnecessary work in this case.

HalidOdat: The problem is that if we do not keep the asciiness, then it makes access such as comparing hard (instead of just doing `JsStr == JsStr`) also hashing and any other operation, which are more common.

HalidOdat: > Yeah, that's the downside of this other approach. I think having to convert from ASCII strings to UTF-16 strings repeatedly, and traverse UTF-16 strings several times to assert that they're ASCII (e.g. trim_start, which remains cheap with the alternative approach) would be more costly than having to compare by u32 iterator.

Yes, about this I had some optimizations in mind about this, one approach was to was that we only check if what we trimmed was non-ascii (it's uncommon to have non-acsii whitespace).

Another approach I had was to return a type that can be non-ascii for ascii string and lazy evaluate to a known ascii/non-ascii type.

Both optimizations are valid and both can bee applied :)

> In addition to that, the comparison problem can be solved with a JsStr that has the UTF-16 representation and (optionally) the ASCII representation of the string, and we can offer a js_str! macro to convert from &str to JsStr.

If we concatenate two (`u8`) ascii strings it will allocate it in a `u16` array, defeating the purpose of this optimization.

> I just have a question. Why would hashing be affected?

Because hashing two strings that are both ascii and same characters but different types (`u8` and `u16`) (gives two different hashes)[https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b5df58a7e685f3f75133006727125ee3].

We would have to iterate though the ascii string and convert to `u16` and hash that.

jedel1043: Ok, if we're going with this approach, is there any way to make `JsStringSlice` internal to the implementation?

My main worry is to pollute our API with 4 types of string references with slightly different representations and semantics.

HalidOdat: > My main worry is to pollute our API with 4 types of string references with slightly different representations and semantics.

We can put in under `#[doc(hidden)]`, it's a common way to hide implementation detail, we can't make it `pub(crate)` because `js_string!()` macro uses it.

It was getting a bit messy, but still better than all the [string v8 offers](https://github.com/danbev/learning-v8/blob/master/notes/string.md) :laughing:

HalidOdat: In the future we may want to implement ConsStrings, which is a string that keeps a pointer to two other strings, which would make string concatenation cheep for big strings.

jedel1043: > We can put in under #[doc(hidden)], it's a common way to hide implementation detail, we can't make it pub(crate) because js_string!() macro uses it.

Then there's a way to make those private! We can rename the `utf16` macro to `js_str`, and the macro should convert an `&str` into its correct variant depending on whether it has non-ascii chars or not. I think that should remove the need to use `JsSliceString` on `js_string`.

jedel1043: Just realized that this also has the advantage of not having to traverse a `str` at runtime to know if it has ascii chars or not :)

jedel1043: > In the future we may want to implement ConsStrings, which is a string that keeps a pointer to two other strings, which would make string concatenation cheep for big strings.

That and inline strings, since it's pretty useless to allocate for strings of length less than 10-20 chars.

jedel1043:
> Checked locally and there is not difference between main and this PR in conformance, it seems that data repo's test262 data is not being updated?

That's weird. I checked the output of more recent PRs such as dependabot's and everything looks fine.

HalidOdat:
> That's weird. I checked the output of more recent PRs such as dependabot's and everything looks fine.

I reran the test262 workflow and now it shows the correct conformance! :)

raskad (commented):
Just got a few comments, otherwise looks great!

raskad (approved):

core/engine/src/object/mod.rs
```diff
@@ -464,6 +465,17 @@ pub struct FunctionBinding {
...
+    #[inline]
+    fn from(name: JsStr<'_>) -> Self {
+        let name: JsString = name.into();
+        Self {
+            binding: name.clone().into(),
+            name: name.clone(),
+        }
+    }
+}
+
```
jedel1043, line 478: I think for things that require converting to `JsString` we should require the `JsString` itself, instead of a `JsStr`. This mostly matches Rust's APIs around collections, which commonly take the inserted key by value instead of by reference.

This probably applies to this, set methods on objects, and in general any method that necessarily needs to take a `JsString` by value.

TLDR: We should remove implicit conversions like this, in favour of taking `JsString` by value.

jedel1043: I do see the utility of `JsStr`, but that will be more helpful to avoid allocations on all the getter methods of the engine.

jedel1043 (approved):
Really impressive work!

PR #3443, opened 2023-10-31
Temporal duration update and cleanup

nekevss:
Some more work on `temporal`, in a little bit of a smaller PR 😉 

Mostly just trying to do a bit of cleanup and small cleanups/refactors with the idea of beginning to migrate some of the temporal methods and structures into their own crates.

It changes the following:
- Updates and/or begins to update a couple of `Temporal.Duration`'s methods and abstracts to relatively recent changes to the specification.
- General nitpicky cleanups to begin isolating some Temporal APIs from Boa specifics (calendar impl's needing context)

Aside: are smaller PRs a bit more similar to this one going to be easier to manage even if they are relatively more unfinished? I'm trying to avoid the monstrosity that was the last PR.

PR #3451, opened 2023-11-03
Progress on Duration's round/total method updates

nekevss:
Follow up to #3443.

Just looking to merge more progress on some `Duration` methods. 😄

Let me know what you think!


HalidOdat (approved):
Looks good to me! :)

boa_engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -698,7 +698,107 @@ impl Duration {
...
+        // 27. Else if abs(duration.[[Hours]]) ≥ 24, set hoursToDaysConversionMayOccur to true.
+        let conversion_may_occur = if duration.inner.days() != 0.0 && zoned_relative_to.is_some() {
+            true
+        } else {
+            24f64 <= duration.inner.hours().abs()
+        };
+
+        // 28. If smallestUnit is "nanosecond" and roundingIncrement = 1, let roundingGranularityIsNoop be true; else let roundingGranularityIsNoop be false.
+        let is_noop = smallest_unit == TemporalUnit::Nanosecond
+            && (rounding_increment - 1f64).abs() < f64::EPSILON;
```
jedel1043, line 712: I'd suggest using an `u16` for the rounding increment insteaed to avoid having to use `f64` here. This is what is used on `get_digit_format_options`.

jedel1043 (commented):
Nice progress! Just have a small suggestion.

boa_engine/src/builtins/temporal/duration/mod.rs
nekevss: The one thing that's a bit rough is that the spec currently has `RoundNumberToIncrement` and `RoundNumberToIncrementAsIfPositive`, so `rounding_increment` can be negative in this instance (granted that could change depending on `Intl.NumberFormat V3`).

jedel1043: I don't think it can though? I'm pretty sure what could be negative is `x`, the number being formatted. `rounding_increment` is strictly positive thanks to the check inside `https://tc39.es/proposal-temporal/#sec-temporal-totemporalroundingincrement`

nekevss: Yep, you're 100% right. I'm going to have to keep track of this change, but I think it should be fine.

boa_engine/src/builtins/temporal/instant/mod.rs
```diff
@@ -403,7 +403,11 @@ impl Instant {
         };
 
         // 16. Perform ? ValidateTemporalRoundingIncrement(roundingIncrement, maximum, true).
-        super::validate_temporal_rounding_increment(rounding_increment.into(), maximum as f64, true)?;
+        super::validate_temporal_rounding_increment(
+            rounding_increment.into(),
+            maximum as f64,
```
jedel1043, line 408: the `increment` and `dividend` arguments of `validate_temporal_rounding_increment` should also probably change to an `u64`. I'm seeing only strictly positive numbers for it, and a `u64` should be capable of storing maximum easily.

jedel1043 (approved):
No more nitpicks :) Great work!

PR #3456, opened 2023-11-06
Simplify all extensions APIs of `Context`

jedel1043:
It seems like all our extensions usages fall in the same types:
- `Rc` for `JobQueue` and `ModuleLoader`.
- `&'static` for `HostHooks`.

This makes our API unnecessarily complex, so we should probably just commit to those representations to simplify our API for the common use case. This also removes the lifetime parameter `Context` has, which is why there are so many changes. The important changes are:
- https://github.com/boa-dev/boa/pull/3456/files#diff-a665b3b6f564521875fd0d725bffbc4f0cc84e5feefdc5fd875fd943e56311cd (boa_engine/src/context/mod.rs)
- https://github.com/boa-dev/boa/pull/3456/files#diff-febf8b33d5e09ed83ee1d3e23ce44ec020e86033d256526875ec07183b93d7c0 (boa_engine/src/context/maybe_shared.rs [deleted])
- https://github.com/boa-dev/boa/pull/3456/files#diff-872037c107c01bf644ede412e4802b3eefeb5a70ce595c441f75651d45111b2a (boa_engine/src/context/icu.rs)


nekevss (approved):
Looks good to me :)

boa_engine/src/context/mod.rs
```diff
@@ -106,13 +104,13 @@ pub struct Context<'host> {
 
     /// ICU related utilities
     #[cfg(feature = "intl")]
-    icu: icu::Icu<'host>,
+    icu: icu::Icu,
 
-    host_hooks: MaybeShared<'host, dyn HostHooks>,
+    host_hooks: &'static dyn HostHooks,
```
HalidOdat, line 109: Why isn't this `Box` too, what if the user wants to store state?

HalidOdat (approved):
Looks good! Just a nitpick :)

boa_engine/src/context/mod.rs
jedel1043: I think most uses of `HostHooks` override one or two methods without using any state at all. Using `&'static` avoids unnecessary allocations in that case, and users who want to have state inside those can just use `realm.host_defined()` or the (soon to be) `context.host_defined()`.

jedel1043: Though, I'm not attached to this optimization per se, so if you have a different opinion, I can rollback to using a `Box<dyn HostHooks>`.

jedel1043: That's completely true! Thank you for reminding me of that :)

jedel1043: Oops, forgot that we need to use `Rc` because some methods take `context` by mutable reference, and `Rc<dyn Trait>` does allocate on zero sized types :(

jedel1043: Every time we encounter this limitation again, I rethink if we should just take `Context` by immutable reference in the module loader and the host hooks...

HalidOdat: In that case, since we have `realm.host_defined()` and in the future `context.host_defined()`, I think it's fine :)

PR #3460, opened 2023-11-08
`[[HostDefined]]` Improvements

johnyob:
# Context

**Related PRs**: #3370 

Currently `HostDefined` doesn't permit you to mutably borrow two objects from the `[[HostDefined]]` field since the `FxHashMap` is wrapped under a `GcRefCell`.

# Description    

This PR implements a `get_mut_many` (from `hashbrown`'s `HashMap`) method to permit accessing several `NativeObject`s using a `NativeTuple`.

Additionally, this commit takes the opportunity to provide automatic downcasting on the `insert` and `remove` methods for `[[HostDefined]]`.

### DIFF
--- a/boa_engine/src/host_defined.rs
+++ b/boa_engine/src/host_defined.rs
@@ -1,14 +1,9 @@
 use std::any::TypeId;
 
-use boa_gc::{GcRef, GcRefCell, GcRefMut};
 use boa_macros::{Finalize, Trace};
-use rustc_hash::FxHashMap;
+use hashbrown::hash_map::HashMap;
 
 use crate::object::NativeObject;
-
-/// Map used to store the host defined objects.
-#[doc(hidden)]
-type HostDefinedMap = FxHashMap<TypeId, Box<dyn NativeObject>>;
 
 /// This represents a `ECMASCript` specification \[`HostDefined`\] field.
 ///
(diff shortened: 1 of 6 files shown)


boa_engine/src/host_defined.rs
```diff
@@ -1,111 +1,145 @@
...
 ///
 /// This allows storing types which are mapped by their [`TypeId`].
 #[derive(Default, Trace, Finalize)]
 #[allow(missing_debug_implementations)]
 pub struct HostDefined {
-    state: GcRefCell<HostDefinedMap>,
+    types: HashMap<TypeId, Box<dyn NativeObject>>,
+}
+
+#[allow(unsafe_op_in_unsafe_fn)]
```
jedel1043, line 17: We turned this lint on because rustc is planning to warn on all usages of unsafe operations outside `unsafe` blocks, even if they're inside an `unsafe` function. We should do that and document the safety invariants of the `unsafe` block instead.

boa_engine/src/host_defined.rs
```diff
@@ -1,111 +1,145 @@
...
+unsafe fn downcast_boxed_native_object_unchecked<T: NativeObject>(
```
jedel1043, line 18: Can you add this comment to the function:
```
// TODO: Track https://github.com/rust-lang/rust/issues/65991 and
// https://github.com/rust-lang/rust/issues/90850 to remove this
// when those are stabilized.

**Review by jedel1043 — CHANGES_REQUESTED** [review]
Looks really great! Just have a comment about the lint and some suggestions.

File: boa_engine/src/host_defined.rs [hunk 3]
```diff
@@ -1,111 +1,145 @@
...
     /// Returns [`Some`] with the object if it exits, [`None`] otherwise.
-    ///
-    /// # Panics
-    ///
-    /// Panics if [`HostDefined`] field is borrowed.
     #[track_caller]
-    pub fn remove<T: NativeObject>(&self) -> Option<Box<dyn NativeObject>> {
-        self.state.borrow_mut().remove(&TypeId::of::<T>())
+    pub fn remove<T: NativeObject>(&mut self) -> Option<Box<dyn NativeObject>> {
+        self.types.remove(&TypeId::of::<T>())
```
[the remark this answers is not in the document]
johnyob: Fixed

boa_examples/src/bin/host_defined.rs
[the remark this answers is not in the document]
johnyob: Added example usage of `get_many_mut`.

boa_engine/src/host_defined.rs
johnyob: I documented the internal invariants of `HostDefined` and a safety comment to `downcast_boxed_native_object_unchecked`.

boa_engine/src/host_defined.rs
johnyob: Added comment

boa_engine/src/host_defined.rs
jedel1043: Nice! You can just remove the lint and add an `unsafe` block after the `SAFETY` comment now. That should fix it. Something like:
```rust
let raw: *mut dyn NativeObject = Box::into_raw(obj);

// SAFETY: safety comment
unsafe {
    Box::from_raw(raw.cast::<T>())
}
```

johnyob: Done :)

jedel1043 (approved):
Excellent work!

HalidOdat (approved):
Nice addition! Looks good to me! :)

raskad (approved):

PR #3461, opened 2023-11-11
Migrate `Temporal` to its own crate.

nekevss:
This is the initial rough draft on separating `Temporal` into an engine agnostic crate. There's a lot in this branch approach-wise that I think is an improvement on the current implementation, and at the very least we may be able to work in to the current implementation in smaller pieces.

~~The biggest point of concern that is a potential blocker for this PR is the safety involving custom user calendars implementation (`CustomRuntimeCalendar` in `boa_engine/src/builtins/temporal/calendar/object.rs`). Currently, the implementation has `CustomRuntimeCalendar` hold a `*mut Context` to provide access to Boa's context when the user calendar is called by the temporal library.~~

Edit:

@jedel1043's suggestion of using `&mut dyn Any` and downcasting seems to work, so I'll work on cleaning this up and getting this in a more mergeable state! 😄

Feel free to leave any feedback whenever!

Also, as this is a more agnostic library, should we continue with the crate continue with the naming approach of the other crates (e.g. `boa_temporal`) or should the name be more of a generic name (e.g. `temporal_rs`).

Edit 2: Linking #1804 to this PR


boa_engine/src/builtins/temporal/calendar/object.rs
```diff
@@ -0,0 +1,299 @@
...
+
+impl CalendarProtocol for CustomRuntimeCalendar {
+    fn date_from_fields(
+        &self,
+        fields: &mut TemporalFields,
+        overflow: ArithmeticOverflow,
+    ) -> TemporalResult<TemporalDate> {
+        // Safety: Context lives for the life of the program and execution, so
+        // this should, in theory, be valid.
+        let context = unsafe { &mut *self.ctx };
```
jedel1043, line 51: I don't think this is valid. Any `JsObject` could live indefinitely right now, even after its containing context was dropped. Maybe in the future we could support it with the GC redesign, but at the moment we'll need to change our approach.

That being said, I think we can receive a `&mut dyn Any` parameter on the trait and downcast to `&mut Context` inside the implementation.

jedel1043 (changes requested):

boa_engine/src/builtins/temporal/calendar/object.rs
nekevss: Looks like that should actually work perfectly. 😄

nekevss:
Marking this as ready for review as I think this is probably a good starting point to continue from for the crate. `Date`, `Duration`, `DateTime`, and the ISO fields have been migrated. The calendar protocol has also been moved over and can support a safe user defined "`CustomRuntimeCalendar`" implementation, which at the very least makes me a lot more comfortable with moving forward with this.

General thoughts:
There are definitely some points that can be improved on with future PRs (or this one if someone wants it added) or came up while migrating this.

1. The `CalendarFields` and `Calendar.prototype.fields` implementation is not complete in this migration and I need to reason about it more, at least spec-wise. My initial thought is that there needs to be an update to the `CalendarProtocol` or a new super trait entirely which felt outside of the scope for this PR specifically.

2. It would probably be worthwhile to migrate the ISO8601 parsing and nodes over to this crate as well, but it seemed like that could be completed in another PR as well if it was deemed as needed.

3. `RoundingMode` was copied over but not removed from `builtins/options.rs` due to `RoundingMode` being used in `intl` as well, I wasn't entirely sure what the preferred approach would be as far as, `intl` using the `temporal` crate's `RoundingMode` or keep them separate.

4. As in the above edit, I wasn't entirely clear what name/version should be put on this crate. I went with `boa_temporal` and the workspace version for consistency across the project. But as this crate would be rather nascent, I'm almost wondering if it should have a separate version (e.g. `0.0.1`) that would better communicate the library's state.

jedel1043:
I think it would be good to discuss with the whole team if we want to separate this crate from Boa or not, so I'll add it to tomorrow's discussion topics.

nekevss:
Agreed 😄 I just wanted the branch to be up to date and ready if the decision was to move forward with it.

jedel1043:
@raskad @HalidOdat will try to review this in the next few days.

HalidOdat (approved):
Great work on this @nekevss! :)

There are some small things like docs, and the like but, we can leave those for future PRs :)

raskad:
@jedel1043 I think Github will not let us merge this until you have approved the PR because you requested changes.

boa_temporal/src/fields.rs
```diff
@@ -0,0 +1,487 @@
...
+            _ => month_code_integer,
+        };
+
+        self.month = Some(new_month);
+
+        Ok(())
+    }
+}
+
+fn month_code_to_integer(mc: TinyStr4) -> TemporalResult<i32> {
```
jedel1043, line 470: Could probably use `TinyAsciiStr<4>` instead, since this type has an ominous warning:

`These are temporary compatability reexports that will be removed in a future version.`

boa_temporal/src/date.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// ==== Context based API ====
+
+impl Date {
+    /// Returns the date after adding the given duration to date.
+    ///
+    /// Temporal Equivalent: 3.5.13 `AddDate ( calendar, plainDate, duration [ , options [ , dateAdd ] ] )`
+    #[inline]
+    #[cfg(feature = "context")]
```
jedel1043, line 196: I don't think adding a feature that gates a parameter is the greatest idea. If two separate crates import `boa_temporal` but one of them uses the `context` feature and the other doesn't, it would break the build for the one that doesn't (cargo unifies features). A better idea would be to remove the feature and add a `contextual_add_to_date` method that has the optional parameter, and `add_to_date` shouldn't take any `context` parameter.

jedel1043 (changes requested):
Overall a really needed crateization! Just have a concert about the "context" feature, but after that's solved we can merge.

jedel1043:
> I think Github will not let us merge this until you have approved the PR because you requested changes.

Yeah, I have a comment that definitely should be resolved before merging this.

jedel1043 (approved):
Nice work!

PR #3465, opened 2023-11-14
Make well_known_symbols functions pub

tj825:
This Pull Request makes the `well_known_symbols` functions -- `boa_engine::symbol::JsSymbol::iterator() -> JsSymbol` etc -- `pub` instead of `pub(crate)`.

This is supposed to allow embedders to get hold of well-known symbols in Rust code, e.g. to implement iterable native objects, etc.

(Have I missed some better way to do this?)


jedel1043 (approved):
Yeah, I would be very surprised if the spec removed any well known symbols in the future, so I think it is okay to expose them as public.

nekevss (approved):
Makes sense to me. 😄 Thanks for the contribution!

PR #3468, opened 2023-11-15
feat(console): reuse deno console's impl

ahaoboy:
This Pull Request  reuses deno's console implementation with the following changes: 
1. Global function __boa_no_color controls whether the output contains color information. 
2. Global function __boa_print is used for outputting characters one by one.

The testing use a hacky solution by overriding __boa_print and __boa_no_color to store colorless output into the property __hack_print_output of a global object.

An odd problem occurs when attempting to output an array containing Chinese characters using ```console.log(['你', '好'])```. However, the code runs successfully if the join() function is used before the output: ```console.log(['你', '好'].join())```.  I don't know how to fix this...


ahaoboy:
The error appears to be an issue with the RegExp, so this PR is ready to merge.

https://github.com/boa-dev/boa/issues/3474

ahaoboy:
The issue https://github.com/ridiculousfish/regress/pull/74 was fixed, but no new version has been released. I validated it locally using the latest version.

jedel1043:
We discussed this in our triage and we think having to load a whole script when loading the console is a little bit funky. Maybe this could benefit from extracting the essentials of the display algorithms and reimplementing them using native Rust code.

ahaoboy:
> We discussed this in our triage and we think having to load a whole script when loading the console is a little bit funky. Maybe this could benefit from extracting the essentials of the display algorithms and reimplementing them using native Rust code.

I haven't tested what impact introducing this huge JS file will have on size and performance. I'm also more inclined to rewrite the log function using Rust, but that will be a long-term process and beyond my capabilities.

Issue #3314, opened 2023-09-27
Refactor `Class` API


jedel1043:
Right now our `Class` and `ClassBuilder` API is a lot stricter than its equivalent `ConstructorBuilder`, which makes `ClassBuilder` a bit clunkier to use. Ideally we should integrate all `ConstructorBuilder`'s APIs with `ClassBuilder`, and also add some more features to make `ClassBuilder` a bit nicer to use. Specifically:

- [ ] Allow using any `PropertyKey` as the name of a method.
- [ ] Allow changing the base object of the generated constructor object.
- [ ] Allow deriving from other classes.
- [ ] Integrate a "class map" of sorts to easily access the activated classes (Possibly using #2952).

PR #3479, opened 2023-11-24
Add `JsNativeObject` wrapper

johnyob:
This PR adds a `JsNativeObject` wrapper around `JsObject` with the invariant that the underlying object data is a native object of some type `T`.

PR #3480, opened 2023-11-24
Add `Object::set_native_object_data`

johnyob:
Related issue: #3314 

This PR permits one to set the native object data of a `Object` using `Object::set_native_object_data`.

```rust
use boa_engine::{Context, object::ObjectInitializer};
let js_unit = ObjectInitializer::with_native((), context).build();

// Now stores 42 instead of unit
let _ = js_unit.borrow_mut().set_native_object_data(42u8);
```

### DIFF
--- a/boa_engine/src/object/mod.rs
+++ b/boa_engine/src/object/mod.rs
@@ -78,6 +78,7 @@ use boa_gc::{custom_trace, Finalize, Trace, WeakGc};
 use std::{
     any::{Any, TypeId},
     fmt::{self, Debug},
+    mem,
     ops::{Deref, DerefMut},
 };
 
@@ -1986,6 +1987,19 @@ impl Object {
     pub fn as_native_object(&self) -> Option<&dyn NativeObject> {
         match self.kind {
             ObjectKind::NativeObject(ref object) => Some(object.as_ref()),
+            _ => None,
+        }
+    }
+
+    /// Sets the native object data returning the previous data, if the
+    /// object is a 'NativeObject'.
+    #[inline]
+    pub fn set_native_object_data<T: NativeObject>(
+        &mut self,
+        data: T,
+    ) -> Option<Box<dyn NativeObject>> {
+        match self.kind {
+            ObjectKind::NativeObject(ref mut object) => Some(mem::replace(object, Box::new(data))),
             _ => None,
         }
     }


jedel1043:
I'm pretty sure this change conflicts with #3479.

In general, we ensure objects cannot change type after creation; this is to ensure APIs such as `JsNativeObject` cannot panic. However, this API in combination with #3479 would make it possible to do things such as:

```rust
let data = ();
let object = JsObject::from_proto_and_data(proto, ObjectData::native_object(data));
let wrapper: JsNativeObject<()> = JsNativeObject::from_object(object).unwrap() // cannot panic
wrapper.borrow_mut().set_native_object_data(42u8);
let unit = wrapper.as_mut() // PANIC!
```

johnyob:
I'm aware of this :pensive:

Unfortunately I'm not too sure what the best approach is. For some context, we'd like this to permit passing a unsafe `JsNativeObject<T>` for a constructor of `T` which would then use `set_native_object_data` to make it safe again.

jedel1043:
> For some context, we'd like this to permit passing a unsafe `JsNativeObject<T>` for a constructor of `T` which would then use `set_native_object_data` to make it safe again.

Must you specifically use `JsNativeObject<T>`? Or can you use `Class` APIs? I was thinking that the `JsNativeObject::new` should be integrated in `Class` instead, so that it returns a `JsNativeObject` after calling `Construct`. Would that solve your specific case?

Otherwise, do you have a link to your use case? I'll try to think on an API that solves it without having to change the type of an object.

johnyob:
> Must you specifically use JsNativeObject<T>? Or can you use Class APIs? I was thinking that the JsNativeObject::new should be integrated in Class instead, so that it returns a JsNativeObject after calling construct. Would that solve your specific case?

Yes. So for some more context, we have our own `Class` API (called `NativeClass`) which can be found here: https://github.com/trilitech/jstz/blob/86e4c1a437c859c7b4d1bdee831fe92d3155bc38/jstz_core/src/native.rs#L401. 

Integrating the creation of a `JsNativeObject` more closely with class would be suitable for our use case. 
Specifically we want to support constructors such as: https://github.com/trilitech/jstz/blob/86e4c1a437c859c7b4d1bdee831fe92d3155bc38/jstz_api/src/url/mod.rs#L75 which relies a native object reference `JsNativeObject<Url>` to pass to the `UrlSearchParams` object to permit the cyclic link between `URL` and `URLSearchParams` from the URL spec.

jedel1043:
Nice! In that case, we can close this and plan the integration of `JsNativeObject` with the `Class` trait.

Issue #3252, opened 2023-09-01
Tracking issue for implemented methods on `JsTypedArray`


jedel1043:
Implementation in https://github.com/boa-dev/boa/blob/main/boa_engine/src/object/builtins/jstypedarray.rs

- [x] at
- [x] get buffer
- [x] get byteLength
- [x] get byteOffset
- [x] constructor
- [x] copyWithin
- [x] entries
- [x] every
- [x] fill
- [x] filter
- [x] find
- [x] findIndex
- [x] findLast
- [x] findLastIndex
- [x] forEach
- [x] includes
- [x] indexOf
- [x] join
- [x] keys
- [x] lastIndexOf
- [x] get length
- [x] map
- [x] reduce
- [x] reduceRight
- [x] reverse
- [x] set
- [x] slice
- [x] some
- [x] sort
- [x] subarray
- [x] toLocaleString
- [x] toReversed
- [x] toSorted
- [x] toString
- [x] values
- [x] with
- [x] @@iterator
- [x] get @@toStringTag

### Desired features

- [x] Easily return the inner data as `&[u8]` or `&Vec<u8>`

AryaveerSR:
If I understand correctly, the methods need to call the respective method on `TypedArray` with a reference to the inner value and the arguments, do minor wrapping/unwrapping/conversions etc. and return as a `JsValue` right ?

jedel1043:
> If I understand correctly, the methods need to call the respective method on  `TypedArray` with a reference to the inner value and the arguments, do minor wrapping/unwrapping/conversions etc. and return as a `JsValue` right ?

Yes, but some methods like `toSorted` haven't been implemented for `TypedArray` yet

AngeloChecked:
Hi @jedel1043 and @AryaveerSR 
Are you actively addressing this issue? I am willing to assist if there are no concerns about potential overlap?

Just to get better the issue, are these implementations okay, or am I missing something?
```
/// Calls `TypedArray.prototype.keys()`.
pub fn keys(&self, context: &mut Context<'_>) -> JsResult<JsValue> {
    TypedArray::keys(&self.inner, &[], context)
}

/// Calls `TypedArray.prototype.entries()`.
pub fn entries(&self, context: &mut Context<'_>) -> JsResult<JsValue> {
    TypedArray::entries(&self.inner, &[], context)
}

/// Calls `TypedArray.prototype.findLast()`.
[...]

PR #3481, opened 2023-11-25
Js typed array methods

AngeloChecked:
This Pull Request pertains to issue #3252

It implements the following JsTypedArray methods:

- get buffer
- constructor
- copyWithin
- find_index
- find_last
- find_last_index
- foreach
- includes
- set
- subarray
- toLocaleString*
- toStringTag

Hi guys!
I'd like to create this draft pull request to share my work on the issue through examples and seek feedback. Is that alright with you? I'll gradually compile this post as I make progress on the issue.

### DIFF
--- a/core/engine/src/object/builtins/jstypedarray.rs
+++ b/core/engine/src/object/builtins/jstypedarray.rs
@@ -61,7 +61,27 @@ impl JsTypedArray {
         BuiltinTypedArray::at(&self.inner.clone().into(), &[index.into().into()], context)
     }
 
+
+    /// Returns the `ArrayBuffer` referenced by this typed array at construction time.
+    /// 
     /// Calls `TypedArray.prototype.buffer()`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use boa_engine::{js_string, JsResult, object::{builtins::{JsUint8Array, JsArrayBuffer}}, property::{PropertyKey}, JsValue, Context};
+    /// # fn main() -> JsResult<()> {
+    ///
+    /// let context = &mut Context::default();
+    /// let array_buffer8 = JsArrayBuffer::new(8, context)?;
+    /// let array = JsUint8Array::from_array_buffer(array_buffer8, context)?;
+    /// assert_eq!(
+    ///     array.buffer(context)?.as_object().unwrap().get(PropertyKey::String(js_string!("byteLength")), context).unwrap(), 
+    ///     JsValue::new(8)
+    /// );
+    /// # Ok(())
+    /// # }
+    /// ```
     #[inline]
     pub fn buffer(&self, context: &mut Context) -> JsResult<JsValue> {
         BuiltinTypedArray::buffer(&self.inner.clone().into(), &[], context)
--- a/core/engine/src/object/builtins/jstypedarray.rs
+++ b/core/engine/src/object/builtins/jstypedarray.rs
@@ -842,6 +842,30 @@ impl JsTypedArray {
                 .expect("`with` must always return a `TypedArray` on success"),
         })
     }
+
+    /// It is a getter that returns the same string as the typed array constructor's name. 
+    /// It returns `Ok(JsValue::Undefined)` if the this value is not one of the typed array subclasses.
+    /// 
+    /// Returns `TypedArray.prototype.toStringTag()`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use boa_engine::{JsResult, js_string, object::{builtins::{JsUint8Array}}, Context};
+    /// # fn main() -> JsResult<()> {
+    ///
+    /// let context = &mut Context::default();
+    /// let array = JsUint8Array::from_iter(vec![1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8], context)?;
+    /// let tag = array.to_string_tag(context)?.to_string(context)?;
+    /// assert_eq!(tag, js_string!("Uint8Array"));
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[inline]
+    pub fn to_string_tag(&self, context: &mut Context) -> JsResult<JsValue> {
+        BuiltinTypedArray::to_string_tag(&self.inner.clone().into(), &[], context)
+    }
 }
 
 impl From<JsTypedArray> for JsObject {
--- a/core/engine/src/object/builtins/jstypedarray.rs
+++ b/core/engine/src/object/builtins/jstypedarray.rs
@@ -439,6 +439,59 @@ impl JsTypedArray {
         }
     }
 
+    /// Executes a provided function once for each typed array element. 
+    /// 
+    /// Calls `TypedArray.prototype.forEach()`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use boa_gc::{Gc, GcRefCell};
+    /// # use boa_engine::{JsResult, object::{builtins::JsUint8Array, FunctionObjectBuilder}, NativeFunction, JsValue, Context};
+    /// # fn main() -> JsResult<()> {
+    /// let context = &mut Context::default();
+    /// let array = JsUint8Array::from_iter(vec![1, 2, 3, 4, 5], context)?;
+    /// let num_to_modify = Gc::new(GcRefCell::new(0u8));
+    ///
+    /// let js_function = FunctionObjectBuilder::new(
+    ///     context.realm(),
+    ///     NativeFunction::from_copy_closure_with_captures(
+    ///         |_, args, captures, inner_context| {
+    ///             let element = args
+    ///                 .first()
+    ///                 .cloned()
+    ///                 .unwrap_or_default()
+    ///                 .to_uint8(inner_context)
+    ///                 .expect("error at number conversion");
+    ///             *captures.borrow_mut() += element;
+    ///             Ok(JsValue::Undefined)
+    ///         },
+    ///         Gc::clone(&num_to_modify),
+    ///     ),
+    /// )
+    /// .build();
+    ///
+    /// array.for_each(js_function, None, context);
+    /// let borrow = *num_to_modify.borrow();
+    /// assert_eq!(borrow, 15u8);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[inline]
+    pub fn for_each(
+        &self,
+        callback: JsFunction,
+        this_arg: Option<JsValue>,
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        BuiltinTypedArray::foreach(
+            &self.inner.clone().into(),
+            &[callback.into(), this_arg.into_or_undefined()],
+            context,
+        )
+    }
+
     /// Calls `TypedArray.prototype.indexOf()`.
     pub fn index_of<T>(
         &self,
--- a/core/engine/src/object/builtins/jstypedarray.rs
+++ b/core/engine/src/object/builtins/jstypedarray.rs
@@ -331,6 +331,56 @@ impl JsTypedArray {
         }
     }
 
+    /// Iterates the typed array in reverse order and returns the value of 
+    /// the first element that satisfies the provided testing function. 
+    /// If no elements satisfy the testing function, `JsResult::Ok(None)` is returned.  
+    /// 
+    /// Calls `TypedArray.prototype.findLast()`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use boa_engine::{JsResult, object::{builtins::JsUint8Array, FunctionObjectBuilder}, NativeFunction, JsValue, Context};
+    /// # fn main() -> JsResult<()> {
+    /// let context = &mut Context::default();
+    /// let data: Vec<u8> = (0..=255).collect();
+    /// let array = JsUint8Array::from_iter(data, context)?;
+    ///
+    /// let lower_than_200_predicate = FunctionObjectBuilder::new(
+    ///     context.realm(),
+    ///     NativeFunction::from_fn_ptr(|_this, args, _context| {
+    ///         let element = args
+    ///             .first()
+    ///             .cloned()
+    ///             .unwrap_or_default()
+    ///             .as_number()
+    ///             .expect("error at number conversion");
+    ///         Ok(JsValue::Boolean(element < 200.0))
+    ///     }),
+    /// )
+    /// .build();
+    /// assert_eq!(
+    ///     array.find_last(lower_than_200_predicate.clone(), None, context),
+    ///     Ok(JsValue::Integer(199))
+    /// );
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[inline]
+    pub fn find_last(
+        &self,
+        predicate: JsFunction,
+        this_arg: Option<JsValue>,
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        BuiltinTypedArray::find_last(
+            &self.inner.clone().into(),
+            &[predicate.into(), this_arg.into_or_undefined()],
+            context,
+        )
+    }
+
     /// Calls `TypedArray.prototype.indexOf()`.
     pub fn index_of<T>(
         &self,
--- a/core/engine/src/object/builtins/jstypedarray.rs
+++ b/core/engine/src/object/builtins/jstypedarray.rs
@@ -155,6 +155,50 @@ impl JsTypedArray {
         Ok(self.clone())
     }
 
+    /// Returns a new typed array on the same `ArrayBuffer` store and with the same element 
+    /// types as for this typed array. 
+    /// The begin offset is inclusive and the end offset is exclusive. 
+    ///
+    /// Calls `TypedArray.prototype.subarray()`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use boa_engine::{JsResult, object::{builtins::JsUint8Array}, JsValue, Context};
+    /// # fn main() -> JsResult<()> {
+    ///
+    /// let context = &mut Context::default();
+    /// let array = JsUint8Array::from_iter(vec![1u8, 2u8, 3u8, 4u8, 5u8, 6u8, 7u8, 8u8], context)?;
+    /// let subarray2_6 = array.subarray(2, 6, context)?;
+    /// assert_eq!(subarray2_6.length(context)?, 4);
+    /// assert_eq!(subarray2_6.get(0, context)?, JsValue::new(3.0));
+    /// assert_eq!(subarray2_6.get(1, context)?, JsValue::new(4.0));
+    /// assert_eq!(subarray2_6.get(2, context)?, JsValue::new(5.0));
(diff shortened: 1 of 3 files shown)


AngeloChecked:
hi @jarredholman 
Can I rely on you to provide feedback about my work? 

What are your thoughts on my testing of the 'foreach' in this commit: [link to line](https://github.com/boa-dev/boa/pull/3481/commits/a82abb7e04b320e8fd50f0683a2e611ca65d840d#diff-65aa7cf881803282f88389947a3eac86c2edf886654a078552da8df467056b5aR97)? I heavily utilized GC and borrowing. Do you think that's alright?

boa_examples/src/bin/jstypedarray.rs
```diff
@@ -41,6 +44,80 @@ fn main() -> JsResult<()> {
...
+        NativeFunction::from_copy_closure_with_captures(
+            |_, args, captures, inner_context| {
+                let element = args
+                    .get(0)
+                    .cloned()
+                    .unwrap_or_default()
+                    .to_uint8(inner_context)
+                    .expect("error at number conversion");
+
+                captures.borrow_mut().add_assign(element);
```
jedel1043, line 108: I think this can be replaced by `*captures.borrow_mut() += element` for simplicity and to avoid importing the trait.

boa_examples/src/bin/jstypedarray.rs
```diff
@@ -41,6 +44,80 @@ fn main() -> JsResult<()> {
...
+                Ok(JsValue::Undefined)
+            },
+            Gc::clone(&num_to_modify),
+        ),
+    )
+    .build();
+
+    let _unused = array.foreach(js_function, None, context);
+
+    let borrow = *Gc::clone(&num_to_modify).borrow();
```
jedel1043, line 118: I don't think you need to clone the number again, you can just directly call `borrow` and compare with `15u8`.

boa_engine/src/object/builtins/jstypedarray.rs
```diff
@@ -273,6 +273,84 @@ impl JsTypedArray {
...
+        let index = BuiltinTypedArray::find_last_index(
+            &self.inner.clone().into(),
+            &[predicate.into(), this_arg.into_or_undefined()],
+            context,
+        )?
+        .as_number()
+        .expect("TypedArray.prototype.findLastIndex() should always return number");
+
+        #[allow(clippy::float_cmp)]
+        if index == -1.0 {
```
jedel1043, line 332: `index < 0.0` to avoid the clippy allow

boa_engine/src/object/builtins/jstypedarray.rs
```diff
@@ -273,6 +273,84 @@ impl JsTypedArray {
...
+            Ok(None)
+        } else {
+            Ok(Some(index as usize))
```
jedel1043, line 335: Should probably use `JsValue::to_index` just in case.

boa_engine/src/object/builtins/jstypedarray.rs
```diff
@@ -273,6 +273,84 @@ impl JsTypedArray {
...
     }
 
+    /// Calls `TypedArray.prototype.findIndex()`.
+    #[inline]
+    pub fn find_index(
+        &self,
+        predicate: JsFunction,
+        this_arg: Option<JsValue>,
+        context: &mut Context,
+    ) -> JsResult<Option<usize>> {
```
jedel1043, line 283: This should return `u64`, since JS considers numbers below 2^53 - 1 as valid indices. I don't think we ensure this in `JsArray` so I'll open an issue for that.

jedel1043 (changes requested):
Really nice work! Just have some suggestions.

boa_engine/src/object/builtins/jstypedarray.rs
AngeloChecked: unfortunately, `JsValue::to_index` returns a `u64` and it can't be -1

boa_engine/src/object/builtins/jstypedarray.rs
AngeloChecked: i think find methods do this because they might return -1

AngeloChecked: Could `Option::None` serve as an alternative to -1? How would `Option::None` be translated into the JavaScript world?

boa_engine/src/object/builtins/jstypedarray.rs
jedel1043: I meant after checking that `index >= 0.0`, but on second thought I think it would also complicate the implementation slightly, so I think this is fine.

boa_engine/src/object/builtins/jstypedarray.rs
jedel1043: No, I meant returning `None` if `index` is negative, and converting the `f64` to `u64` otherwise.

AngeloChecked: did it, but I'm wondering: the function api should return -1 if the element isn't found, how does `None` appear to the end user in JavaScript?

jedel1043: I don't think we have a direct conversion from `Option<T>` to `JsValue`, but if we do, it would probably map as `Some(T) -> JsValue::from(T)` and `None -> JsValue::Undefined`

AngeloChecked: Do you think we should see this as a problem now, or should we wait for later?

jedel1043: I don't think that blocks adding this? This will be exclusively a Rust API. Implementors can handle an `Option` however they want to, if they need to somehow "expose" the result to JS again.

AngeloChecked:
It appears that there's an overlap between `TypedArray.prototype.set()` and the functionality defined in [ `https://tc39.es/ecma262/#sec-set-o-p-v-throw`](https://github.com/boa-dev/boa/blob/main/boa_engine/src/object/operations.rs#L85). I'm uncertain how to resolve this conflict. I've renamed it to `typedArraySet`, but I'm open to any suggestions or considerations.

jedel1043:
> It appears that there's an overlap between [`TypedArray.prototype.set()`](https://github.com/boa-dev/boa/pull/3481/commits/3af2e0aae60d4263444cb898926c2f28cba11e82#diff-4c4cfd75fa55cac3adae8f820438e15fddcf72881ad8c052de8defab3d38e4bdR241) and the functionality defined in [ `https://tc39.es/ecma262/#sec-set-o-p-v-throw`](https://github.com/boa-dev/boa/blob/main/boa_engine/src/object/operations.rs#L85). I'm uncertain how to resolve this conflict. I've renamed it to `typed_array_set`, but I'm open to any suggestions or considerations.

Should probably be renamed to `set_values`, because the original method allows setting multiple values at the same time.

AngeloChecked:
While exploring `toLocaleString` and attempting to replicate the example from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/toLocaleString), I encountered a localization problem that prevented the code from working as expected. I'm unsure if this issue is widespread. As a temporary solution, I've commented out the example code for now.

jedel1043:
> While exploring `toLocaleString` and attempting to replicate the example from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/toLocaleString), I encountered a localization problem that prevented the code from working as expected. I'm unsure if this issue is widespread. As a temporary solution, I've commented out the example code for now.

Don't worry about that. We haven't implemented the `NumberFormat` API, meaning our numbers always format to plain numbers. When we implement that, this should be fixed.

AngeloChecked:
I want to summarize the current status of the PR:

**function implemented:**
- [x] at <- already implemented
- [x] get buffer
- [x] get byteLength <- already implemented
- [x] get byteOffset <- already implemented
- [x] constructor
- [x] copyWithin
- [x] every <- already implemented
- [x] fill <- already implemented
- [x] filter <- already implemented
- [x] find <- already implemented
- [x] findIndex 
- [x] findLast
- [x] findLastIndex
- [x] forEach
- [x] includes 
- [x] indexOf  <- already implemented
- [x] join  <- already implemented
- [x] lastIndexOf  <- already implemented
- [x] get length  <- already implemented
- [x] map  <- already implemented
- [x] reduce  <- already implemented
- [x] reduceRight  <- already implemented
- [x] reverse  <- already implemented
- [x] set
- [x] slice  <- already implemented
- [x] some  <- already implemented
- [x] sort  <- already implemented
- [x] subarray
- [x] toReversed  <- already implemented
- [x] toSorted  <- already implemented
- [x] with  <- already implemented

**function implemented but:**
- [x] toLocaleString -> currently not working
- [x] get @@toStringTag -> i don't know if its behavior is correct

**i don't know how to implement it**
- [ ] toString

**iterators are unstable**
- [ ] @https://github.com/iterator
- [ ] values
- [ ] keys
- [ ] entries

**desired features**
- [ ] Easily return the inner data as &[u8] or &Vec<u8>

---

@jedel1043 in your opinion what can be the next steps?

Can we review each implementation to align with the desired feature and reduce the reliance on `JsValue`? Any suggestions or tips on how to achieve this?

jedel1043:
> Can we review each implementation to align with the desired feature and reduce the utilization of `JsValue`?

Yep, it'll be good to check already implemented functions to see if we can improve the API on them, but that can be done in another PR.

> Any suggestions or tips on how to achieve this?

Probably to check the JS implementation to see if there are any patterns on the return types of the methods; methods returning only `JsValue::String` should return `JsString` on the object wrapper, methods returning either `undefined` or a specific type should return `Option<Type>`, etc.

jedel1043:
Also, when you feel like this is ready, just ping me so that I can change the label to `waiting-on-review`

AngeloChecked:
hi @jedel1043 
```rust
pub fn buffer(&self, context: &mut Context) -> JsResult<JsValue>
pub fn constructor(&self, context: &mut Context) -> JsResult<JsValue>
pub fn copy_within<T>( &self, target: T, start: u64, end: Option<u64>, context: &mut Context,) -> JsResult<Self> where T: Into<JsValue>
pub fn find_index( &self, predicate: JsFunction, this_arg: Option<JsValue>, context: &mut Context,) -> JsResult<Option<u64>> 
pub fn find_last( &self, predicate: JsFunction, this_arg: Option<JsValue>, context: &mut Context,) -> JsResult<JsValue>
pub fn find_last_index( &self, predicate: JsFunction, this_arg: Option<JsValue>, context: &mut Context,) -> JsResult<Option<u64>>
pub fn foreach( &self, callback: JsFunction, this_arg: Option<JsValue>, context: &mut Context,) -> JsResult<JsValue> 
pub fn includes<T>( &self, search_element: T, from_index: Option<u64>, context: &mut Context,) -> JsResult<bool> where T: Into<JsValue>
pub fn set_values( &self, source: JsValue, offset: Option<u64>, context: &mut Context,) -> JsResult<JsValue>
pub fn subarray(&self, begin: i64, end: i64, context: &mut Context) -> JsResult<Self>
```
from your point of view should any of these fn signs be improved, make the types more explicit?
otherwise, I think I'll go with the review, considering that any side improvements can be put in another card.

AngeloChecked:
hi @jedel1043,

It's been a while. What are our next steps? Did I forget to complete something specific? Can I assist with any tasks for the review?

core/engine/src/object/builtins/jstypedarray.rs
```diff
@@ -273,6 +358,104 @@ impl JsTypedArray {
...
+        if index >= 0.0 {
+            Ok(Some(index as u64))
+        } else {
+            Ok(None)
+        }
+    }
+
+    /// Calls `TypedArray.prototype.forEach()`.
+    #[inline]
+    pub fn foreach(
```
nekevss, line 424: nit: Should this be in snake case?

This is extending the API from `BuiltinTypedArray`, but we use `for_each` in other places in the engine. Would it be better to maintain consistency?

core/engine/src/object/builtins/jstypedarray.rs
```diff
@@ -273,6 +358,104 @@ impl JsTypedArray {
         )
     }
 
+    /// Calls `TypedArray.prototype.findIndex()`.
```
nekevss, line 361: nit: can there be some examples added to these docs.

Since `JsTypedArray` et al. are meant to be an externally facing API. Some more thorough examples in the documentation would be nice. Something a bit more like the examples provided in [`JsPromise`](https://boajs.dev/boa/doc/boa_engine/object/builtins/struct.JsPromise.html) or [`JsMap`](https://boajs.dev/boa/doc/boa_engine/object/builtins/struct.JsMap.html).

nekevss (changes requested):
This looks great overall! Really nice job!

I had just a couple nits.

core/engine/src/object/builtins/jstypedarray.rs
AngeloChecked: Hi @nekevss, big thanks for your support! Apologies for the delay; I got tied up with something else. Now, I'm eager to maintain continuity with this.

I used `foreach` because the internal API `BuiltinTypedArray::foreach` uses it too. But I'm flexible to rename it for consistency with our preferred naming conventions, following your guidance.

core/engine/src/object/builtins/jstypedarray.rs
AngeloChecked: Do you think it would be more meaningful to enhance the [builtins/jstypedarray.rs](https://github.com/AngeloChecked/boa/blob/JsTypedArray_methods/core/engine/src/object/builtins/jstypedarray.rs) module with standard Rust documentation, or should we focus solely on improving the code in the [examples](https://github.com/AngeloChecked/boa/blob/JsTypedArray_methods/examples/src/bin/jsarraybuffer.rs) folder? I drew a lot of inspiration from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/findIndex) for those examples, because originally I hadn't thought about it; I used them as acceptance tests to ensure the implementation was correct. However, I'm open to reconsidering them as documentation if that would be more beneficial.

core/engine/src/object/builtins/jstypedarray.rs
nekevss: @jedel1043 any thoughts on the use of snake case with `foreach` on the internal API for `BuiltinTypedArray`.

In general, I'm of the opinion for at least the external API to be closer to the `Rust` convention. Anyone who consumes the API will probably expect snake case. At minimum, we should probably be at least consistent throughout the engine on whether it's `foreach` or `for_each`.

jedel1043: I think we mostly agree on always converting camelCase JS APIs to snake_case Rust APIs. IMO, functions that differ from this convention should be treated as naming bugs.

core/engine/src/object/builtins/jstypedarray.rs
jedel1043: The convention is to have complex examples about general usage on `examples`, and simple examples as documentation comments.

jedel1043: The rule of thumb is to have examples that demonstrate HOW to use the APIs as doc comments, and have examples that demonstrate WHEN to use the APIs as code examples.

nekevss: The module level `rustdoc` documentation was more what I meant in this instance. The examples don't have to be a completely different from those in `examples`. Mostly since the Rust documentation will be auto generated into the docs, which is super handy and may be the first place someone consuming the API would go for information. For example, here's the [`JsPromise` docs](https://boajs.dev/boa/doc/boa_engine/object/builtins/struct.JsPromise.html) for reference

core/engine/src/object/builtins/jstypedarray.rs
nekevss: Sounds good! 😄 Then let's go with renaming the functions from `foreach` to `for_each` for naming.

core/engine/src/object/builtins/jstypedarray.rs
AngeloChecked: ok, [here](https://github.com/boa-dev/boa/pull/3481/commits/703411b32d8343bcadec43594a9451b0ed1864df#diff-1173b2fbe95c226df7883fdf2e2b9c0e40d2270b6ee2f58053b0f3ef38bce345R361) i created the first rust doc test

the preview looks like:
image

I intend to keep drawing considerable inspiration from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/findIndex) for comments. What are your thoughts? Should I proceed with implementing Rust doc tests throughout the module? Do you have any specific considerations or tips to share?

jedel1043: Looks really nice! I think it's a good call to adapt the examples from MDN, since it removes the effort of having to come up with descriptive usages.

As a small suggestion, I would remove the "The FindIndex method of Array instances" prefix, since it is mostly idiomatic to start a method's documentation with a verb; in this case, "Returns the index of...".

AngeloChecked: I've added doc tests to all the functions I've touched. Give me feedback!

nekevss (commented):
This is looking great! I have a one nit about the docs, but nothing else really.

core/engine/src/object/builtins/jstypedarray.rs
```diff
@@ -61,6 +61,32 @@ impl JsTypedArray {
         BuiltinTypedArray::at(&self.inner.clone().into(), &[index.into().into()], context)
     }
 
+
+    /// returns the `ArrayBuffer` referenced by this typed array at construction time.
```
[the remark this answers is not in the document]
AngeloChecked: my bad! I've corrected all the capitalization errors, and I've also sqash each commit to correspond with its respective function.

nekevss (approved):
LGTM! Thanks for contributing to Boa! 😄

EDIT: Looks like this needs a `cargo fmt -all`

AngeloChecked:
> LGTM! Thanks for contributing to Boa! 😄
> 
> EDIT: Looks like this needs a `cargo fmt -all`

Sure, please close this PR. However, if we want to discuss the points further:

**i don't know how to implement it**
- [ ] toString

**iterators are unstable**
- [ ] @https://github.com/iterator
- [ ] values
- [ ] keys
- [ ] entries

**desired features**
- [ ] Easily return the inner data as &[u8] or &Vec<u8>

--- 
I'm open to contribute. I'm also open to adding Rust test documentation to the jstypedarray module in another pull request. What do you think?

jedel1043:
> Easily return the inner data as &[u8] or &Vec

I think this should be exposed by `ArrayBuffer` and `SharedArrayBuffer` instead. We already allow getting the inner buffer of the typed array, and projecting to another inner object would mean maintaining two borrows at the same time, which is not ideal.

jedel1043 (approved):
Really nice work! It's also really helpful to have so many new doc tests for this wrapper. There are definitely some other optimizations that can be done, but having the methods in place is already a big improvement.

Issue #3487, opened 2023-11-29
Consider moving some builtins to the `NativeObject` variant of `JsObject`


jedel1043:
There are some object types like `Intl` related objects or `Temporal` objects that don't benefit from being a whole variant of `JsObject`, since they're not perf related or they're just not used in busy loops. It would be good to transition all of those to use the `NativeObject` API instead, so that we can reduce the code size of the `object` module.

PR #3491, opened 2023-11-30
Transition `Intl` types to `NativeObject` API

jedel1043:
Related to #3487.

This makes our `Object` implementation a lot simpler (saves us a lot of feature gates), so we should see if we can move other non-perf related objects to this.


HalidOdat (approved):
Nice refactor!

Always happy to see more deletions than additions :D

nekevss (approved):

PR #3492, opened 2023-12-01
Reduce `WeakGc<T>` memory usage

HalidOdat:
Makes inner storage of `WeakGc<T>` store `()` in place of `Ephemeron` value, instead of a clone of the `Gc<T>`, which is already stored in the key.

Reducing the size by the width of one pointer.

### DIFF
--- a/boa_gc/src/pointers/ephemeron.rs
+++ b/boa_gc/src/pointers/ephemeron.rs
@@ -2,7 +2,7 @@ use crate::{
     finalizer_safe,
     internals::EphemeronBox,
     trace::{Finalize, Trace},
-    Allocator, Gc,
+    Allocator, Gc, GcBox,
 };
 use std::ptr::NonNull;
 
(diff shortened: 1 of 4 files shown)


boa_gc/src/pointers/ephemeron.rs
```diff
@@ -33,6 +33,17 @@ impl<K: Trace, V: Trace + Clone> Ephemeron<K, V> {
         unsafe { self.inner_ptr.as_ref().value().cloned() }
     }
 
+    /// Gets the stored value of this `Ephemeron`, or `None` if the key was already garbage collected.
+    ///
+    /// This needs to return a clone of the value because holding a reference to it between
+    /// garbage collection passes could drop the underlying allocation, causing an Use After Free.
+    #[must_use]
+    pub fn key_ptr(&self) -> Option<NonNull<GcBox<K>>> {
```
jedel1043, line 41: This should probably return a proper `Gc<K>`, since a GC pass could invalidate the returned pointer.

raskad (approved):
Good find!

PR #3493, opened 2023-12-01
Use `Vec<T>` for keeping track of gc objects

HalidOdat:
This does not reduce memory usage, since we still keep a pointer to the gc objects, but instead it reduces the overhead of jumping through the intrusive linked list and also simplifies the gc collection functions.

---

#### Main

```
PROGRESS Richards
RESULT Richards 36.5
PROGRESS DeltaBlue
RESULT DeltaBlue 41.7
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 55.5
PROGRESS RayTrace
RESULT RayTrace 147
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 138
PROGRESS Splay
RESULT Splay 120
PROGRESS NavierStokes
RESULT NavierStokes 111
SCORE 81.0
```

#### PR

```
PROGRESS Richards
RESULT Richards 39.1
PROGRESS DeltaBlue
RESULT DeltaBlue 46.4
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 56.3
PROGRESS RayTrace
RESULT RayTrace 150
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 153
PROGRESS Splay
RESULT Splay 166
PROGRESS NavierStokes
RESULT NavierStokes 112
SCORE 88.9
```

### DIFF
--- a/boa_gc/src/internals/ephemeron_box.rs
+++ b/boa_gc/src/internals/ephemeron_box.rs
@@ -19,7 +19,6 @@ const NON_ROOTS_MAX: u32 = NON_ROOTS_MASK;
 pub(crate) struct EphemeronBoxHeader {
     ref_count: Cell<u32>,
     non_root_count: Cell<u32>,
-    pub(crate) next: Cell<Option<NonNull<dyn ErasedEphemeronBox>>>,
 }
 
 impl EphemeronBoxHeader {
@@ -28,7 +27,6 @@ impl EphemeronBoxHeader {
         Self {
             ref_count: Cell::new(1),
             non_root_count: Cell::new(0),
-            next: Cell::new(None),
         }
     }
 
--- a/boa_gc/src/internals/gc_box.rs
+++ b/boa_gc/src/internals/gc_box.rs
@@ -1,9 +1,5 @@
 use crate::Trace;
-use std::{
-    cell::Cell,
-    fmt,
-    ptr::{self, NonNull},
-};
+use std::{cell::Cell, fmt, ptr};
 
 const MARK_MASK: u32 = 1 << (u32::BITS - 1);
 const NON_ROOTS_MASK: u32 = !MARK_MASK;
@@ -20,7 +16,6 @@ const NON_ROOTS_MAX: u32 = NON_ROOTS_MASK;
 pub(crate) struct GcBoxHeader {
     ref_count: Cell<u32>,
     non_root_count: Cell<u32>,
-    pub(crate) next: Cell<Option<NonNull<GcBox<dyn Trace>>>>,
 }
 
 impl GcBoxHeader {
@@ -29,7 +24,6 @@ impl GcBoxHeader {
         Self {
             ref_count: Cell::new(1),
             non_root_count: Cell::new(0),
-            next: Cell::new(None),
         }
     }
 
--- a/boa_gc/src/internals/weak_map_box.rs
+++ b/boa_gc/src/internals/weak_map_box.rs
@@ -1,19 +1,14 @@
 use crate::{pointers::RawWeakMap, GcRefCell, Trace, WeakGc};
-use std::{cell::Cell, ptr::NonNull};
 
 /// A box that is used to track [`WeakMap`][`crate::WeakMap`]s.
 pub(crate) struct WeakMapBox<K: Trace + Sized + 'static, V: Trace + Sized + 'static> {
     pub(crate) map: WeakGc<GcRefCell<RawWeakMap<K, V>>>,
-    pub(crate) next: Cell<Option<NonNull<dyn ErasedWeakMapBox>>>,
 }
 
 /// A trait that is used to erase the type of a [`WeakMapBox`].
 pub(crate) trait ErasedWeakMapBox {
     /// Clear dead entries from the [`WeakMapBox`].
     fn clear_dead_entries(&self);
-
-    /// A pointer to the next [`WeakMapBox`].
-    fn next(&self) -> &Cell<Option<NonNull<dyn ErasedWeakMapBox>>>;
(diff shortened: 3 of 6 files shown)


jedel1043:
Um, I'm pretty sure the point of using intrusive linked lists here is to avoid having live allocations on the GC state. This does make it faster, but consider that the `Vec` will keep expanding if the number of objects keeps growing, and it'll never reclaim the allocated buffer. This can be seen using our heap profiling tools:

## Sample

```javascript
{
    let array = []
    for (let i = 0; i <= 100000; i++) {
        array.push({data: i});
    }
}

$boa.gc.collect()

{
    let array = []
    for (let i = 0; i <= 100000; i++) {
        array.push({data: i});
    }
}

$boa.gc.collect()
```

Running `cargo run --profile release-dbg --features dhat -- test.js --debug-object`

### Main

```bash
dhat: Total:     248,322,251 bytes in 1,984,424 blocks
dhat: At t-gmax: 53,298,622 bytes in 274,607 blocks
dhat: At t-end:  499,999 bytes in 2,648 blocks
```

### This PR

```bash
dhat: Total:     217,362,347 bytes in 1,984,405 blocks
dhat: At t-gmax: 57,012,374 bytes in 283,328 blocks
dhat: At t-end:  4,678,511 bytes in 2,650 blocks
```

Note how main uses only 500k bytes of heap at the end of the program, while this PR uses 4M bytes.

HalidOdat:
In that case we can [`shrink_to`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.shrink_to) with `len + some_grace_value` (maybe `len/2` or `len/3`) on every collect.

HalidOdat:
> Yeah, that's what I expected. Perf is virtually the same with that change.

Umm.. Note the "previous implementation" is the previous implementation of this PR (without https://github.com/boa-dev/boa/commit/685b29d6252c73fc66989a6873747b20b8a3ef89), not main branch, so there is still a pretty nice perf increase. (I should have added a note on it :sweat_smile: ). I wanted to show that limiting the capacity allocation does not effect the performance, oddly I saw a very slight improvement (:confused:)?

There is a ~40 score increase in splay from main's imlementation, probably because it creates so many objects, to construct the tree.

jedel1043:
> > Yeah, that's what I expected. Perf is virtually the same with that change.
> 
> Umm.. Note the "previous implementation" is the previous implementation of this PR (without https://github.com/boa-dev/boa/commit/685b29d6252c73fc66989a6873747b20b8a3ef89), not main branch, so there is still a pretty nice perf increase. (I should have added a note on it :sweat_smile: ). I wanted to show that limiting the capacity allocation does not effect the performance, oddly I saw a very slight improvement (:confused:)?
> 
> There is a ~40 score increase in splay from main's imlementation, probably because it creates so many objects, to construct the tree.

Noted 👍

I still feel it's "wrong" to use allocations to keep track of allocations, so I'd still defer to the rest of the team for more opinions on this.

HalidOdat:
This just moves the _memory_ that was used to track the gc pointers (`next` field) into a dense array so that the CPU can better cache, hence the speed improvement.

Why is this weird?

In the future, if we aspire to achieve significant improvements in speed, it may be necessary to develop our own "malloc" and take charge of managing the heap pages ourselves. Perhaps my exposure to a considerable amount of v8/Spidermonkey code and techniques has influenced my perspective of what is weird. Making this proposed change seem normal. :laughing:

jedel1043:
> Perhaps my exposure to a considerable amount of v8/Spidermonkey code and techniques has influenced my perspective of what is weird. Making this proposed change seem normal.

Ooh, does V8/SM use dynamic arrays to keep track of allocated objects? If that's the case, it'll remove my uncertainty completely.

HalidOdat:
> Ooh, does V8/SM use dynamic arrays to keep track of allocated memory? If that's the case, it'll remove my uncertainty completely.

They preallocate a large slabs of memory, they keep track of whats free instead of what is allocated. I _think_ they use an in-place free list. 

You can find some information [here](https://github.com/danbev/learning-v8/blob/master/notes/heap.md)

> Perhaps my exposure to a considerable amount of v8/Spidermonkey code and techniques has influenced my perspective of what is weird. Making this proposed change seem normal. 😆

It was an attempted joke (a note to self: written jokes can be hard to convey. bad jokes especially). They do some really hacky stuff to squeeze as much performance (and I love/hate it).

---

Anyway, I just felt like it was a waste to leave that big of a performance boost. Though I'm totally fine with closing if there is any cause for concern or a consensus has been reached :)

@raskad what are your thoughts on this?

jedel1043:
Made some research and Spidermonkey uses a mixed architecture of "Zones", which are mini heaps of memory that contain many objects and track them using linked lists. Their GC collects at the granularity of zones, and they keep track of the active zones using a `Vec` (technically a `SmallVec` that is first an array but expands to a `Vec` when exceeding 4 zones).

I think this should be enough evidence to accept this change.

jedel1043 (approved):

HalidOdat:
> I'm courious about the strict provenance argument. As far as I know miri has some strict provenance checks correct? @HalidOdat did you run miri on the boa_gc tests with this change?

Ran it with the `-Zmiri-strict-provenance` env option and miri did not complain and it ran fine.

```bash
MIRIFLAGS="-Zmiri-strict-provenance" cargo +nightly miri test -p boa_gc
```

nekevss (approved):
If SM is using a `Vec` equivalent and `miri` isn't throwing any errors, then I'm fine with moving forward with this.

raskad (approved):

PR #3496, opened 2023-12-03
Refactor vm calling convention to allow locals

HalidOdat:
This PR makes storing locals after `fp` much easier by moving the `fp` at the end of the stack that the caller sets up, making the frame pointer the in the middle, the locals start at fp and accessing the caller setup values we offset back instead.

This adds a local ( /register :wink: ) count variable to `CodeBlock`.

Checkout [`call_frame.rs`](https://github.com/boa-dev/boa/blob/93d7a48850cb6ffdeb8634e08f0e6fe88328141b/boa_engine/src/vm/call_frame/mod.rs#L84) for the new calling convention.

This makes #3194 easier and allows us to move stuff from the call frame that is only for a specific type of function (like `async` promise capability or generator function) reducing the size of the call frame and only paying for things the user uses.

Moves async generator object out of `CallFrame` to the stack. Reducing the usage of memory of a CallFrame for non-async-generator functions.

### DIFF
--- a/core/engine/src/builtins/generator/mod.rs
+++ b/core/engine/src/builtins/generator/mod.rs
@@ -20,7 +20,7 @@ use crate::{
     string::common::StaticJsStrings,
     symbol::JsSymbol,
     value::JsValue,
-    vm::{CallFrame, CompletionRecord, GeneratorResumeKind},
+    vm::{CallFrame, CallFrameFlags, CompletionRecord, GeneratorResumeKind},
     Context, JsArgs, JsData, JsError, JsResult, JsString,
 };
 use boa_gc::{custom_trace, Finalize, Trace};
@@ -74,6 +74,10 @@ impl GeneratorContext {
 
         frame.fp = CallFrame::FUNCTION_PROLOGUE + frame.argument_count;
 
+        // NOTE: Since we get a pre-built call frame with stack, and we reuse them.
+        //       So we don't need to push the locals in subsuquent calls.
+        frame.flags |= CallFrameFlags::LOCALS_ALREADY_PUSHED;
+
         Self {
             call_frame: Some(frame),
             stack,
@@ -106,6 +110,13 @@ impl GeneratorContext {
         self.call_frame = context.vm.pop_frame();
         assert!(self.call_frame.is_some());
         result
+    }
+
+    /// Returns the async generator object, if the function that this [`GeneratorContext`] is from an async generator, [`None`] otherwise.
+    pub(crate) fn async_generator_object(&self) -> Option<JsObject> {
+        self.call_frame
+            .as_ref()
+            .and_then(|frame| frame.async_generator_object(&self.stack))
     }
 }
 
--- a/core/engine/src/builtins/generator/mod.rs
+++ b/core/engine/src/builtins/generator/mod.rs
@@ -64,28 +64,20 @@ pub(crate) struct GeneratorContext {
 }
 
 impl GeneratorContext {
-    /// Creates a new `GeneratorContext` from the raw `Context` state components.
-    pub(crate) fn new(stack: Vec<JsValue>, call_frame: CallFrame) -> Self {
-        Self {
-            stack,
-            call_frame: Some(call_frame),
-        }
-    }
-
     /// Creates a new `GeneratorContext` from the current `Context` state.
     pub(crate) fn from_current(context: &mut Context) -> Self {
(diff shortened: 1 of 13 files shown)


core/engine/src/vm/call_frame/mod.rs
```diff
@@ -81,22 +80,49 @@ impl CallFrame {
 impl CallFrame {
     /// This is the size of the function prologue.
     ///
-    /// The position of the elements are relative to the [`CallFrame::fp`].
+    /// The position of the elements are relative to the [`CallFrame::fp`] (frame pointer).
+    ///
+    /// ```text
```
nekevss, line 85: General thought on this: Would it be better to store this `fp` as a `sp` (although I'm not sure it's a true stack pointer, maybe `lp` for local pointer), and keep the `fp` and make up the memory by computing the argument count via `self.sp - self.fp - Self::FUNCTION_PROLOGUE` or would computing the args count be more inefficient?

core/engine/src/builtins/generator/mod.rs
```diff
@@ -64,28 +64,24 @@ pub(crate) struct GeneratorContext {
...
-        };
+        let fp = frame.restore_fp() as usize;
+        let stack = context.vm.stack.split_off(fp);
 
-        context.vm.stack.truncate(fp);
+        frame.fp = CallFrame::FUNCTION_PROLOGUE + frame.argument_count;
 
-        this
+        // NOTE: Since we get a pre-built call frame with stack, and we reuse them.
+        //       So we don't need to push the locals in subsuquent calls.
```
nekevss, line 78:
```suggestion
        //       So we don't need to push the locals in subsequent calls.
```

nekevss (commented):
This looks great! Always glad to see the size of `CallFrame` go down! 😄 Had a couple small doc nits and one general thought that I had when reviewing.

core/engine/src/vm/call_frame/mod.rs
HalidOdat: > Would it be better to store this `fp` as a `sp`

I don't think calling it sp would be correct, sp is implict in the stack vec itself that being the length (pointing to the position that the next value will be put).

Native calling conventions the arguments passed are not part of the call frame of the callee. But it doesn't really matter, since we can make our own convention. :smile: 

The biggest reason for `fp` change is to make reduce the computation needed to access locals before we would have to do `fp + FUNCTION_PROLOGUE + arg_count + local_offset` with this change we only do `fp + local_offset`.

I kind of hinted at the next change, switching to a register based VM :smile: Registers will be used extensively so we want fast access. I think it's a good time to switch before we are too dependent on the stack version, while working on #3037 Modeling data flow of stack based VMs is much harder, and was conflicted whether to continue development, since it would make going switching much harder and it will have to be redone.

nekevss: Yepp! I noticed the register based shift!

Just for clarity, I meant keep `fp` and then `local pointer`/`sp` (whatever we call it) in exchange for `argument_count`. It would then exchanging `restore_fp` in exchange for calculating `argument_count`. I think we still compute the argument_start anyways, which can then just be `fp + FUNCTION_PROLOGUE` over `fp - argument_count`.

```rust
// Current dumbed down version of `CallFrame`
struct DummyFrame {
    fp: u32,
    argument_count: u32
}

// New version
struct NewFrame {
    fp: u32,
    lp: u32
}

// argument_count
impl NewFrame {
    // Trade off would be that we compute arg_count over compute fp
    fn arg_count(&self) -> u32 {
        self.lp - self.arg_start()
    }
    
    // Throwing this in as a utility
    fn arg_start(&self) -> u32 {
        self.fp + FUNCTION_PROLOGUE
    }
}
```

The idea was that maybe we save a bit of compute by not having to use `restore_fp` in exchange for calculating the arg_count. At least that was the thought I had (could be off on that), but thought I'd run it by you nonetheless. Of course, if that would somehow undercut the register shift you had in mind, then completely disregard this.

HalidOdat: Sorry for the delay. I decided to leave this for after we implement the register VM, added a todo so I don't forget :smiley: .

I renamed the fp into rp (register pointer), calling it register pointer, instead of local pointer for consistency with the registers that will be added into the future.

nekevss (approved):
Great work on this!

jedel1043 (approved):

PR #3505, opened 2023-12-06
Reestructure repo and CI improvements

jedel1043:
Our number of crates is slowly growing, and it's getting harder to find the exact crate that you want to open when working on the repo. This PR proposes:
- Removing the `boa_` prefix on all directories, since that's mostly superfluous.
- Separating our crates in groups:
  - `core` for the essential crates.
  - `tests` for test crates.
  - `ffi` for FFI related crates (only wasm for now).
  - `examples`, which is just a plain crate in the root for accessibility.
  - `cli` which is just a plain crate for simplicity.

This also adds some small CI improvements, like better VSCode launch and task scripts, and some cache actions that weren't caching cargo installs.


nekevss:
I actually like this change a lot! One early question: is there a reason for `core` over `crates`. I think bevy, wasmtime, rust-analyzer, and cargo all use a `crates` directory, and I had assumed, perhaps wrongly, that that was the semi "standard" after a certain point on a project.

jedel1043:
> I actually like this change a lot! One early question: is there a reason for `core` over `crates`. I think bevy, wasmtime, rust-analyzer, and cargo all use a `crates` directory, and I had assumed, perhaps wrongly, that that was the semi "standard" after a certain point on a project.

It's just personal preference, and `crates` is not really standard per se. ICU4X uses a similar structure as the proposed one, with a `components` folder with the public APIs and some secondary directories such as `utils`, `ffi`, `tools`, etc.

HalidOdat (approved):
Nice change! :)

PR #3508, opened 2023-12-06
Non recursive gc trace

### LINKED ISSUE #1848: Long reference "chain" can overflow the stack during GC trace
Very long "reference chains" (i.e. very deeply nested objects) can overflow the stack because, afaict, each object trace during GC cycles is its own function call.
It seems like there's basically no guards against stack overflows (for example even simple cases like #809 can abort the process) so this is probably lower priority, but this sounds a little trickier to guard against.

**To Reproduce**
```js
let old = {};
for (let i = 0; i < 100000000; i++) {
  old = { old };
}
```
After 10 iterations, old would be `{ old: { old: { old: { old: { old: { old: { old: { old: { old: { old: {} } } } } } } } } } }`.
After 1e8 iterations, this would be nested very deeply, so anything that makes the engine recursively visit each property will end in a stack overflow (which is what happens during a GC trace as it needs to mark everything as visited)

**Expected behavior**
Not a stack overflow. The process shouldn't abort.

**Build environment:**
- OS: Ubuntu
- Version: 20.04
- Rustc version: rustc 1.60.0-nightly (75d9a0ae2 2022-02-16)

HalidOdat:
This Pull Request fixes/closes #1848 .

It changes the following:

- Unify GcBoxHeader and EphemeronBoxHeader into GcHeader (duplicate code).
- Use a work queue to traverse the gc ojects.

The reproducible code in the issue has been made into a test for both `boa_gc` and `boa_engine` (reduced the chain object from `100_000_000` to `1_000_000`, takes too long to finish).

~~Note: Marking it as a draft. There are still some things that need to be sorted out. like the vtable extraction and construction without https://github.com/rust-lang/rust/issues/81513 will be much harder to do.~~

Reason for moving vtable to `Gc` object: Because we have `<T: Trace + ?Sized> Trace for Gc<T>` we cannot get a `NonNull<GcBox<dyn Trace>>` from `Gc<dyn Trace>` so we can enque it (we can implement a special case for `Gc<dyn Trace>` but types like `Gc<VTableObject<dyn Trace>>` cannot be done), because it would require us to have double DST, which is not possible. Moving the vtable to the `GcBox<T>` allow us to construct a `GcBox<dyn Trait>` at any time. This also has the advantage of reducing by half the `NonNull<GcBox<dyn Trace>>` pointers in the gc strong vec, also while gc collecting and tracing queue. Could also have some interesting other optimization, since we could have a `ThinGc<dyn Trace>` that is pointer width while being erased.

Solved by create a custom `VTable` with all the required function, and other metadata like size. Being able to put pretty much anything in the vtable, this approach may be more flexible than working with https://github.com/rust-lang/rust/issues/81513.
The tests run fine with strict provenance under miri.

~~Once the warnings, documentation, some code clean up is done, will mark it ready for review.~~

### DIFF
--- a/core/engine/src/builtins/function/mod.rs
+++ b/core/engine/src/builtins/function/mod.rs
@@ -142,7 +142,7 @@ pub enum ClassFieldDefinition {
 }
 
 unsafe impl Trace for ClassFieldDefinition {
-    custom_trace! {this, {
+    custom_trace! {this, mark, {
         match this {
             Self::Public(_key, func) => {
                 mark(func);
--- a/core/engine/src/builtins/generator/mod.rs
+++ b/core/engine/src/builtins/generator/mod.rs
@@ -45,7 +45,7 @@ pub(crate) enum GeneratorState {
 
 // Need to manually implement, since `Trace` adds a `Drop` impl which disallows destructuring.
 unsafe impl Trace for GeneratorState {
-    custom_trace!(this, {
+    custom_trace!(this, mark, {
         match &this {
             Self::SuspendedStart { context } | Self::SuspendedYield { context } => mark(context),
             Self::Executing | Self::Completed => {}
--- a/core/engine/src/builtins/intl/collator/mod.rs
+++ b/core/engine/src/builtins/intl/collator/mod.rs
@@ -57,7 +57,7 @@ pub(crate) struct Collator {
 
 // SAFETY: only `bound_compare` is a traceable object.
 unsafe impl Trace for Collator {
-    custom_trace!(this, mark(&this.bound_compare));
+    custom_trace!(this, mark, mark(&this.bound_compare));
 }
 
 impl Collator {
--- a/core/engine/src/builtins/map/ordered_map.rs
+++ b/core/engine/src/builtins/map/ordered_map.rs
@@ -42,7 +42,7 @@ pub struct OrderedMap<V> {
 }
 
 unsafe impl<V: Trace> Trace for OrderedMap<V> {
-    custom_trace!(this, {
+    custom_trace!(this, mark, {
         for (k, v) in &this.map {
             if let MapKey::Key(key) = k {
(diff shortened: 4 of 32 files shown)


HalidOdat:
Ran the benchmarks and the seems to be a slight speed improvement (not sure why, expected the opposite since we need to reach into the object to be able to trace it).

Also just realized that with this queue design it would be easier to implement an incremental GC since we can stop at anytime during the tracing and continue afterwards

---

#### Main

```
PROGRESS Richards
RESULT Richards 49.2
PROGRESS DeltaBlue
RESULT DeltaBlue 53.4
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 54.3
PROGRESS RayTrace
RESULT RayTrace 162
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 148
PROGRESS Splay
RESULT Splay 155
PROGRESS NavierStokes
RESULT NavierStokes 103
SCORE 91.8
```

#### PR

```
PROGRESS Richards
RESULT Richards 49.9
PROGRESS DeltaBlue
RESULT DeltaBlue 53.9
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 53.5
PROGRESS RayTrace
RESULT RayTrace 162
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 153
PROGRESS Splay
RESULT Splay 156
PROGRESS NavierStokes
RESULT NavierStokes 103
SCORE 92.3
undefined
```

jedel1043 (changes requested):
Really nice fix! And yeah, this could simplify a LOT the implementation of a concurrent GC.
Just have some critical comments about some methods of `Trace`, and the rest are just nitpicks.

jedel1043 (approved):
Good work!

raskad (approved):
Nice work!

PR #3510, opened 2023-12-06
Implment Tail Call Optimization

RageKnify:
This Pull Request implements Tail Call Optimization as described in the [spec](https://tc39.es/ecma262/#sec-tail-position-calls)

It changes the following:

- introduce a new instruction `TailCall` that can is compatible with `Call` and will perform `TCO` if possible
- introduce `ByteCompiler::optimize` that introduces optimizations to the bytecode code

While the spec says that TCO should be implemented most engines do not currently support it. Apple's JavaScriptCore is the only mainstream implementation as far as I am aware. When it was first introduced V8 implemented it but after Firefox and Internet Explorer pushed back and it was clear not all browsers would have it V8 dropped it to avoid maintaining unnecessary code.

Given that I wanted to get all of @boa-dev/maintainers opinions on implementing these parts of the spec that aren't technically required. I did it mostly as a challenge since it is a topic I find interesting so I understand if the choice is to not incorporate the code.

### DIFF
--- a/core/engine/src/bytecompiler/mod.rs
+++ b/core/engine/src/bytecompiler/mod.rs
@@ -1510,6 +1510,31 @@ impl<'ctx> ByteCompiler<'ctx> {
         }
     }
 
+    fn optimize(bytecode: &mut Vec<u8>, flags: CodeBlockFlags) {
+        // only perform Tail Call Optimisation in strict mode
+        if flags.contains(CodeBlockFlags::STRICT) {
+            // if a sequence of Call, SetReturnValue, CheckReturn, Return is found
+            // replace Call with TailCall to allow frame elimination
+            for i in 0..bytecode.len() {
+                let opcode: Opcode = bytecode[i].into();
+                if matches!(opcode, Opcode::Call) {
+                    if bytecode.len() > i + 5 {
+                        let second = bytecode[i + 2].into();
+                        let third = bytecode[i + 3].into();
+                        let fourth = bytecode[i + 4].into();
+                        if let (Opcode::SetReturnValue, Opcode::CheckReturn, Opcode::Return) =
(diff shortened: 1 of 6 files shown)


raskad:
The implementation looks pretty straight forward. I like it. @RageKnify do you know the more detailed reasons why V8 droped it as a maintanance burden? With our current VM it seems to work pretty well right? @HalidOdat do you think there could be future issues with this?

I would be in favor of merging this. If it becomes hard to maintain or blocks features in the future we could always remove for the same reason V8 did it.

jedel1043:
I'm in favour of merging this, but only if we add a way to disable the optimization.

I'm mostly worried about our recursion limit, which could be bypassed by this feature. Bonus points if we document this on the recursion limit setter/getter.

raskad:
> EDIT: alternatively, if we want to always do the optimization, we could consider it as a "pseudo-loop", and make it count for the loop limit, but I'm not sure how feasible that would be with indirect recursion.

I like this idea. If we can make that work that would be my prefered solution.

HalidOdat:
> @HalidOdat do you think there could be future issues with this?

Correct stack traces is the biggest concern (which @jedel1043 proposed a solution). I would prefer either waiting for the stack trace or we implement the JSC shadow stack (In this PR or a follow up PR) since if we implement stack traces we will have implement it anyway.

> EDIT: alternatively, if we want to always do the optimization, we could consider it as a "pseudo-loop", and make it count for the loop limit, but I'm not sure how feasible that would be with indirect recursion.

It should be easy since we have access to the previous frame that is popped and we just increment and check.

NOTE: In general I would prefer the users use the budget functions to restrict the compute, how i see it is the recursion limit is more of a call frame limit, and tail call does not add any frames.

PR #3522, opened 2023-12-18
Refactor Temporal Calendar API for `AnyCalendar` and fields

nekevss:
This PR is related to the ongoing work for `Temporal`.

The largest change on this PR is primarily updating `CalendarSlot`.

```rust
// Old CalendarSlot
enum CalendarSlot {
    Identifier(String),
    Protocol(Box<dyn CalendarProtocol>),
}

// New CalendarSlot
enum CalendarSlot<C: CalendarProtocol> {
    Builtin(AnyCalendar),
    Protocol(C),
}
```

It changes the following:

- Switches `CalendarSlot::Protocol(Box<dyn CalendarProtocol>)` to `CalendarSlot::Protocol(C)`.
- Switches `CalendarSlot::Identifier(String)` to `CalendarSlot::Builtin(AnyCalendar)`
- Updates `CalendarProtocol` to include `fields` and `mergeFields`.

I think there are still a couple things I want to double check/work through. But posting this as a draft in the meantime for any early feedback/thoughts.


core/engine/src/builtins/temporal/calendar/mod.rs
```diff
@@ -44,11 +44,11 @@ mod tests;
 // SAFETY: `Calendar` doesn't contain traceable types.
 #[boa_gc(unsafe_empty_trace)]
 pub struct Calendar {
-    slot: CalendarSlot,
+    slot: CalendarSlot<JsCustomCalendar>,
```
jedel1043, line 47: Thanks to the new design, we can remove the `unsafe_empty_trace` and implement `Trace` for `CalendarSlot<C: Trace>`.

nekevss:
Marking as ready for review as I believe the core changes for `CalendarSlot<C>` are complete. There are still a few things that could be improved on this branch, specifically around the functionality of new builtin calendars from `AnyCalendar` as well as `CalendarDescriptorFields`, `CalendarFieldKeysToIgnore`, and `CalendarResolveField` (I have a rough idea of what needs to be there, but just wanted to do some further research to confirm).

core/temporal/src/fields.rs
```diff
@@ -463,6 +480,144 @@ impl TemporalFields {
...
+                result.set_field_value(&key, &value)?;
+            }
+        }
+
+        Ok(result)
+    }
+}
+
+/// Iterator over `TemporalFields` keys.
+pub struct KeyIter {
```
HalidOdat, line 512: Just a thought: Maybe we should rename this to just `Keys` in the [standard library they use this convention](https://doc.rust-lang.org/std/collections/btree_map/struct.Keys.html?search=Keys). Same for `ValuesIter`.

HalidOdat (approved):
Great work! :)

raskad (approved):
Looks good to me. I agree with @HalidOdat's comment but that should not block the merge :)

PR #3528, opened 2023-12-20
Move `PromiseCapability` to stack

HalidOdat:
Depends on #3496 

Moves the promise capability to stack. Reduces call frame for non-async functions by 48 bytes.

### DIFF
--- a/core/engine/src/module/source.rs
+++ b/core/engine/src/module/source.rs
@@ -1327,7 +1327,7 @@ impl SourceTextModule {
 
         // 9. Perform ! module.ExecuteModule(capability).
         // 10. Return unused.
-        self.execute(Some(capability), context)
+        self.execute(Some(&capability), context)
             .expect("async modules cannot directly throw");
     }
 
@@ -1741,7 +1741,7 @@ impl SourceTextModule {
     /// [spec]: https://tc39.es/ecma262/#sec-source-text-module-record-execute-module
     fn execute(
         &self,
-        capability: Option<PromiseCapability>,
+        capability: Option<&PromiseCapability>,
         context: &mut Context,
     ) -> JsResult<()> {
         // 1. Let moduleContext be a new ECMAScript code execution context.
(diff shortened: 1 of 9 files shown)


core/engine/src/vm/call_frame/mod.rs
```diff
@@ -232,6 +298,17 @@ impl CallFrame {
         &stack[at as usize]
     }
 
+    /// Returns the local at the given index.
```
raskad, line 301:
```suggestion
    /// Sets the local at the given index.
```

raskad (approved):
Looks good.

core/engine/src/vm/call_frame/mod.rs
```diff
@@ -132,7 +134,10 @@ impl CallFrame {
     pub(crate) const FUNCTION_PROLOGUE: u32 = 2;
     pub(crate) const THIS_POSITION: u32 = 2;
     pub(crate) const FUNCTION_POSITION: u32 = 1;
-    pub(crate) const ASYNC_GENERATOR_OBJECT_REGISTER_INDEX: u32 = 0;
+    pub(crate) const PROMISE_CAPABILITY_PROMISE_REGISTER_INDEX: u32 = 0;
+    pub(crate) const PROMISE_CAPABILITY_RESOLVE_REGISTER_INDEX: u32 = 1;
+    pub(crate) const PROMISE_CAPABILITY_REJECT_REGISTER_INDEX: u32 = 2;
+    pub(crate) const ASYNC_GENERATOR_OBJECT_REGISTER_INDEX: u32 = 3;
```
jedel1043, line 140: Thought: Seeing this like this makes me wonder if we should start migrating to a register architecture, where we have a certain amount of utility registers on `Vm` so that the callee can share data with the caller.

jedel1043 (approved):
Looks great!

core/engine/src/module/source.rs
```diff
@@ -1763,21 +1763,27 @@ impl SourceTextModule {
...
         context
             .vm
             .push_frame_with_stack(callframe, JsValue::undefined(), JsValue::null());
 
+        context
+            .vm
+            .frames
+            .last()
+            .expect("there should be a frame")
+            .set_promise_capability(&mut context.vm.stack, capability);
```
jedel1043, line 1785: Thought: Seeing this like this makes me wonder if we should start migrating to a register-first architecture, where we have a certain amount of utility registers on Vm so that the callee can share data with the caller and viceversa.

PR #3538, opened 2023-12-26
Add UTF-16 input parsing

raskad:
We currently have UTF-16 handling at runtime trough our `JsString` type. But one thing that is missing is support for UTF-16 input handling itself. One case where this is relevant for boa itself it in the handling of `eval`. Because we only allow UTF-8 input, we have to convert the input of `eval` to UTF-8 before parsing it.

To solve this, I added a `ReadChar` trait to the parser crate that allows the parser to handle different input encodings. The parsing itself is done on unicode code points. To make this work I removed all of the parsing that was done on bytes, since that was assuming that we only handle UTF-8 inputs. Most of that work is in https://github.com/boa-dev/boa/commit/55752f2b683867c2787ebf528d760ed333e6e136.

I added a UTF-16 input type. To get some first positive results I also adjusted the regex parsing to work for non UTF-8 inputs. That should give us some `eval` and regex related tests that pass now.

### DIFF
--- a/core/parser/src/source/utf8.rs
+++ b/core/parser/src/source/utf8.rs
@@ -28,27 +28,26 @@ impl<R: Read> ReadChar for UTF8Input<R> {
     fn next_char(&mut self) -> io::Result<Option<u32>> {
         // Decode UTF-8
         let x = match self.next_byte()? {
-            Some(b) if b < 128 => return Ok(Some(u32::from(b))),
-            Some(b) => b,
-            None => return Ok(None),
+            Some(b) if b >= 128 => b,         // UTF-8 codepoint
+            b => return Ok(b.map(u32::from)), // ASCII or None
         };
 
         // Multibyte case follows
         // Decode from a byte combination out of: [[[x y] z] w]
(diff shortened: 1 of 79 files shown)


core/parser/src/source/utf8.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+/// Returns the value of `ch` updated with continuation byte `byte`.
+fn utf8_acc_cont_byte(ch: u32, byte: u8) -> u32 {
+    (ch << 6) | u32::from(byte & CONT_MASK)
+}
+
+fn unwrap_or_0(opt: Option<u8>) -> u8 {
+    opt.unwrap_or(0)
+}
```
jedel1043, line 77: Nitpick: This should probably be inlined, since `var.unwrap_or(0)` is not longer than `unwrap_or_0(var)`

jedel1043 (approved):
Really great to finally have proper UTF-16 parsing! I just have some nitpicks that don't block merging.

HalidOdat (approved):
Great work @raskad!

Just a nitpick :)

PR #3543, opened 2023-12-30
Update `boa_temporal` Time Zone design

nekevss:
The updates on this branch is to bring the time zone implementation in line with the design from #3522 and related to #1804.

It changes the following:

- Changes `TimeZoneSlot` to the below

```rust
pub enum TimeZoneSlot<Z: TzProtocol> {
    /// A native `TimeZone` representation.
    Tz(TimeZone),
    /// A Custom `TimeZone` that implements the `TzProtocol`.
    Protocol(Z),
}
```

- Adds the custom time zone object (with a `Trace` impl).

Notably, as further context on the below, Time Zones right now is mostly the structure/scaffolding with limited functionality. We'll need to figure out a good way to handle tz system calls and IANA mapping (whether that be using some `icu4x` provider or baking something ourselves based from the IANA tz database) to really build this out functionality wise and test it, at least imo. If anyone has any thoughts, let me know 😄

### DIFF
--- a/core/temporal/src/components/tz.rs
+++ b/core/temporal/src/components/tz.rs
@@ -16,29 +16,14 @@ use super::calendar::CalendarProtocol;
 pub const TIME_ZONE_PROPERTIES: [&str; 3] =
     ["getOffsetNanosecondsFor", "getPossibleInstantsFor", "id"];
 
-/// A clonable `TzProtocol`
-pub trait TzProtocolClone {
-    /// Clones the current `TimeZoneProtocol`.
-    fn clone_box(&self) -> Box<dyn TzProtocol>;
-}
-
-impl<P> TzProtocolClone for P
-where
-    P: 'static + TzProtocol + Clone,
-{
-    fn clone_box(&self) -> Box<dyn TzProtocol> {
-        Box::new(self.clone())
-    }
-}
-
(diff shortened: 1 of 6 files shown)


HalidOdat (approved):
Looks good to me! :)

core/temporal/src/components/tz.rs
```diff
@@ -16,29 +16,14 @@ use super::calendar::CalendarProtocol;
...
-    P: 'static + TzProtocol + Clone,
-{
-    fn clone_box(&self) -> Box<dyn TzProtocol> {
-        Box::new(self.clone())
-    }
-}
-
 /// The Time Zone Protocol that must be implemented for time zones.
-pub trait TzProtocol: TzProtocolClone {
+pub trait TzProtocol: Clone {
```
jedel1043, line 20: Thought: We should probably add an associated error type `type Error: From<TemporalError>` so that we can preserve the full engine error. Would also have to be added to `CalendarProtocol`.

jedel1043 (approved):
Good progress! I have a suggestion for the next ergonomics change, but this can be merged as is.

core/temporal/src/components/tz.rs
nekevss: That makes sense. As the change can be done on both traits. I'll update that on another PR for both traits.

PR #3564, opened 2024-01-05
Implement Date parsing according to the spec

raskad:
This PR implements parsing in `Date()` and `Date.parse()` according to the spec.

Until now we only parsed the given string in the RFC 3339 format. The spec allows for [a more relaxed format](https://tc39.es/ecma262/#sec-date-time-string-format), for example partial formats are allowed (e.g. `2024`, `2024T00:00`, ...). While the spec allows for implementation specifc parsing it also specifies that the output of `Date.toString`, `Date.toUTCString` and `Date.toISOString` must be parsable into an equal `Date` object. These formats are now also parsed (the `Date.toISOString` format is the same that the spec defines itself).

### DIFF
--- a/core/engine/src/builtins/date/utils.rs
+++ b/core/engine/src/builtins/date/utils.rs
@@ -1,6 +1,6 @@
-use chrono::{Datelike, NaiveDateTime, Timelike};
-
-use crate::{context::HostHooks, value::IntegerOrNan};
+use crate::{context::HostHooks, value::IntegerOrNan, JsString};
+use chrono::{DateTime, Datelike, Duration, NaiveDate, NaiveDateTime, Timelike};
+use std::{iter::Peekable, str::Chars};
 
 /// The absolute maximum value of a timestamp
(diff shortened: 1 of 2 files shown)


core/engine/src/builtins/date/utils.rs
```diff
@@ -198,3 +198,232 @@ pub(super) fn replace_params<const LOCAL: bool>(
...
+        }
+    }
+
+    fn next_expect(&mut self, expect: char) -> Option<()> {
+        self.input
+            .next()
+            .and_then(|c| if c == expect { Some(()) } else { None })
+    }
+
+    fn next_digit(&mut self) -> Option<u8> {
```
nekevss, line 273: Oooooh, I like this. Definitely makes other parts of parsing a lot nicer too!

nekevss (approved):
This looks good to me! I really like some aspects of it!

I was going to have a bit of a nit about `MINUS` being used, but it looks like it's not in the ECMAScript spec 😅

HalidOdat (approved):

PR #3567, opened 2024-01-05
Fix invalid return value when closing an iterator

raskad:

- Fix invalid return value when closing an iterator


HalidOdat (approved):
Nice catch!

nekevss (approved):

PR #3595, opened 2024-01-18
`Date` refactor

raskad:
This pull request refactors the `Date` builtin to align it with the spec and enable all tests to pass.

The date operations by `chrono` are removed. This is necessary, because `chrono` only allows a limited date range from `Jan 1, 262145 BCE` to `Dec 31, 262143 CE`. The spec allows for dates from `-271821-04-20T00:00:00.000Z` to `+275760-09-13T00:00:00.000Z`. Instead most datetime specific operations are now performed by the functions that the spec specifies. The only thing that is done by the `time` create is the calculation of local timzone offsets. All other usages of `chrono` are also removed so we can remove the dependency.

### DIFF
--- a/core/engine/src/builtins/date/utils.rs
+++ b/core/engine/src/builtins/date/utils.rs
@@ -1,7 +1,7 @@
 use crate::{context::HostHooks, js_string, value::IntegerOrInfinity, JsString};
 use boa_macros::utf16;
-use chrono::{DateTime, NaiveDateTime};
 use std::{iter::Peekable, str::Chars};
+use time::{macros::format_description, OffsetDateTime, PrimitiveDateTime};
 
 // Time-related Constants
 //
@@ -748,13 +748,14 @@ pub(super) fn parse_date(date: &JsString, hooks: &dyn HostHooks) -> Option<i64> 
     }
 
     // `toString` format: `Thu Jan 01 1970 00:00:00 GMT+0000`
-    if let Ok(dt) = DateTime::parse_from_str(&date, "%a %b %d %Y %H:%M:%S GMT%z") {
-        return Some(dt.naive_utc().timestamp_millis());
+    if let Ok(t) = OffsetDateTime::parse(&date, &format_description!("[weekday repr:short] [month repr:short] [day] [year] [hour]:[minute]:[second] GMT[offset_hour sign:mandatory][offset_minute][end]")) {
+        return Some(t.unix_timestamp() * 1000 + i64::from(t.millisecond()));
     }
 
     // `toUTCString` format: `Thu, 01 Jan 1970 00:00:00 GMT`
-    if let Ok(dt) = NaiveDateTime::parse_from_str(&date, "%a, %d %b %Y %H:%M:%S GMT") {
-        return Some(dt.timestamp_millis());
+    if let Ok(t) = PrimitiveDateTime::parse(&date, &format_description!("[weekday repr:short], [day] [month repr:short] [year] [hour]:[minute]:[second] GMT[end]")) {
+        let t = t.assume_utc();
+        return Some(t.unix_timestamp() * 1000 + i64::from(t.millisecond()));
     }
 
     None
--- a/core/engine/src/builtins/date/utils.rs
+++ b/core/engine/src/builtins/date/utils.rs
@@ -50,7 +50,7 @@ pub(super) fn day(t: f64) -> f64 {
 /// [spec]: https://tc39.es/ecma262/#sec-timewithinday
 pub(super) fn time_within_day(t: f64) -> f64 {
     // 1. Return 𝔽(ℝ(t) modulo ℝ(msPerDay)).
-    modulo(t, MS_PER_DAY)
+    t.rem_euclid(MS_PER_DAY)
 }
 
 /// Abstract operation `DaysInYear ( y )`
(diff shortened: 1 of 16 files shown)


core/engine/src/builtins/date/utils.rs
```diff
@@ -1,202 +1,730 @@
...
-    // Consider leap years when calculating the cumulative days added to the year from the input month
-    if (year % 4 != 0) || (year % 100 == 0 && year % 400 != 0) {
-        day += [0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334][month];
-    } else {
-        day += [0, 31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335][month];
-    }
+    // 7. Let mn be 𝔽(ℝ(m) modulo 12).
+    let mn = modulo(m, 12.0) as u8;
+
+    // 8. Find a finite time value t such that YearFromTime(t) is ym, MonthFromTime(t) is mn,
```
nekevss, line 391: I think there's an outstanding issue on `MakeDay` here. Idk if it would be worth tracking it. 

For [reference](https://github.com/tc39/ecma262/issues/1087)

core/engine/src/builtins/date/utils.rs
```diff
@@ -1,202 +1,730 @@
...
+    365
+}
+
+/// Abstract operation `DayFromYear ( y )`
+///
+/// More info:
+/// - [ECMAScript reference][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-dayfromyear
+fn day_from_year(y: f64) -> f64 {
```
nekevss, line 91: I think this is implemented as a util in the `boa_temporal` crate (`epoch_day_number_for_year`). Would it be worth it to just reuse the code in this instance?

nekevss (commented):
Was taking a look at this. A lot of it looks great! Still have a bit more to review but wanted to run a question by you.

I think some of the utils in this have shared logic in the `boa_temporal/utils`. Would it be better to share the logic from `boa_temporal/utils` where possible instead of implementing it twice.

raskad:
> I think some of the utils in this have shared logic in the `boa_temporal/utils`. Would it be better to share the logic from `boa_temporal/utils` where possible instead of implementing it twice.
> 
> EDIT: Just checked the spec, the idea is actually to unify these but the blocker is 1087, but we seem to be treating the values in most case as the mathematical anyways. So it would probably be best to unify them where possible.

My first idea was that we may just go with the different implementations for now, until things like 1087 are resolved or the issue of how these are handeled in the spec is resolved. That way the temporal code could potentially evolve if needed and we have no cross cutting effort. But you are way more up to date on how the temporal spec, if you say the spec will probably not change considerably, I'm all for sharing the logic.

One thing to keep in mind is maybe optimizations. My feeling is that there is at least some optimization potential in the Date logic. For example there are places where functions are called multiple times that could be avoided etc. I wanted to take a detailed look later on and I'm not sure how that would change the interoperability as some Date specific optimizations may require specific logic.

nekevss:
We can keep it separate for now. I think some of this logic is fairly stable, so to speak, but there are still spec changes being made. I'd be surprised if the date equations changed much more, but airing on the cautious side doesn't hurt. Especially if there may be some places for optimizations as you said.

core/engine/src/builtins/date/utils.rs
```diff
@@ -1,202 +1,730 @@
...
+    // and TimeZoneString(tv).
+    js_string!(
+        &date_string(t),
+        utf16!(" "),
+        &time_string(t),
+        &time_zone_string(t, hooks)
+    )
+}
+
+fn modulo(n: f64, m: f64) -> f64 {
```
nekevss, line 673: question: is there any benefit/difference using this modulo function vs. `rem_euclid`

core/engine/src/builtins/date/utils.rs
```diff
@@ -1,202 +1,730 @@
...
+///
+/// [spec]: https://tc39.es/ecma262/#sec-makefullyear
+pub(super) fn make_full_year(year: f64) -> f64 {
+    // 1. If year is NaN, return NaN.
+    if year.is_nan() {
+        return f64::NAN;
+    }
+
+    // 2. Let truncated be ! ToIntegerOrInfinity(year).
+    let truncated = IntegerOrInfinity::from(year);
```
nekevss, line 455: thought: is `IntegerOrInfinity` needed in this instance if we've already accounted for NaN?

nekevss (approved):
Looks good to me overall. There were a couple more things I came across, but nothing blocking.

core/engine/src/builtins/date/utils.rs
jedel1043: IIRC, `IntegerOrInfinity` also rounds to an integer, so I think it's pretty much needed here.

jedel1043 (approved):
Really like the refactor! It's a shame we cannot fully use integers because of the range of `i64`, but I think this is solved by `Temporal` in a much better way.

core/engine/src/builtins/date/utils.rs
nekevss: Yeah, I think it's just truncating.

So I was just wondering whether this step could just be `year.trunc()` in this particular instance.

jedel1043: Hmmm that is true... `IntegerOrInfinity` is mostly used to convert from `f64` to an integer, or handle the non-finite edge cases, but here it's only used for the rounding...

nekevss: It can be left as is and merged. It was just something I when I was reviewing.

I think this could mostly be expressed as the below.

```rust
// Return non-finite cases
if !year.is_finite() {
    return year
}
// truncate
let trunc = year.trunc();

// Check if value is in range.
if (0.0..=99.0).contains(&year) {
    return year 1900.0 + year
}

year

```

raskad: I think I wanted to leave this for the optimizations later. For example in this case, every result of `make_full_year` gets passed to `make_day` afterwards. So there is definitley some potential to optimize this even across spec method boundaries.

PR #3606, opened 2024-01-23
Remove allocations from `HostDefined::get_many_mut`

jedel1043:
Small improvement to that method that avoids creating a `Vec` for indexing.

I also checked the generated assembly and it seems to skip some panics thanks to the const generic, so that's good.


raskad (approved):

HalidOdat (approved):
Nice change! Looks good to me!

PR #3608, opened 2024-01-23
Fix regexp `toString` method

raskad:

- Fix regexp `toString` method to honor the spec.


HalidOdat (approved):
Nice catch!

jedel1043 (approved):

PR #3610, opened 2024-01-24
Implement getter for `ArrayBuffer` data

HalidOdat:
Implements `.data()` and `.data_mut()` for getting a reference to the internal byte data of `ArrayBuffer` without having to detach the buffer.

### DIFF
--- a/core/engine/src/object/builtins/jsarraybuffer.rs
+++ b/core/engine/src/object/builtins/jsarraybuffer.rs
@@ -3,11 +3,13 @@ use crate::{
     builtins::array_buffer::ArrayBuffer,
     context::intrinsics::StandardConstructors,
     error::JsNativeError,
-    object::{internal_methods::get_prototype_from_constructor, JsObject, JsObjectType},
+    object::{
+        internal_methods::get_prototype_from_constructor, ErasedObject, JsObject, JsObjectType,
+    },
     value::TryFromJs,
     Context, JsResult, JsValue,
 };
-use boa_gc::{Finalize, Trace};
+use boa_gc::{Finalize, GcRef, GcRefMut, Trace};
 use std::ops::Deref;
 
 /// `JsArrayBuffer` provides a wrapper for Boa's implementation of the ECMAScript `ArrayBuffer` object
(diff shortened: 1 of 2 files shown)


core/engine/src/object/builtins/jsarraybuffer.rs
```diff
@@ -195,6 +197,78 @@ impl JsArrayBuffer {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[inline]
+    #[must_use]
+    pub fn data(&self) -> Option<GcRef<'_, [u8]>> {
+        GcRef::try_map(
+            self.inner
+                .downcast_ref::<ArrayBuffer>()
+                .expect("inner must be an ArrayBuffer"),
```
jedel1043, line 230: Can we avoid the panic? It seems preferable to just return `None` in that case, since the panic shouldn't be reachable anyways.

HalidOdat: I thought of doing that, but if the invariant is broken then this does not panic, the bug will propagate somewhere else.

Though I'm fine with changing it if you want :)

A thought: It would be interesting if we had something like `JsObject<T = OridinaryObject>` storing that as the inner, that way we can avoid any checks and can access then data without `.downcast_ref()`... At that point we could make `JsArrayBuffer` a type alias to `JsObject<ArrayBuffer>` and just implement the methods on that type `impl JsObject<ArrayBuffer> { ... }`.

jedel1043: > I thought of doing that, but if the invariant is broken then this does not panic, the bug will propagate somewhere else.

We could just add a `debug_assert!` for good measure. IMHO, the advantages of not having another panic in our code outweigh a theoretical bug that we haven't encountered yet.

> A thought: It would be interesting if we had something like `JsObject<T = OridinaryObject>` storing that as the inner, that way we can avoid any checks and can access then data without `.downcast_ref()`... At that point we could make `JsArrayBuffer` a type alias to `JsObject<ArrayBuffer>` and just implement the methods on that type `impl JsObject<ArrayBuffer> { ... }`.

Yep, I'm planning to experiment a bit with that, to see how much we can leverage the new object representation.

jedel1043 (approved):
Everything looks perfect!

nekevss (approved):
Looks good to me

PR #3618, opened 2024-01-27
Implement non-erased `JsObject`s

jedel1043:
This PR implements non-erased `JsObject`s aka `JsObject<T>`. This allows accessing the inner data of an object without having to call `downcast_ref/mut` every time, which is a big ergonomics win that should avoid calling `expect` a lot of times.

This is an MVP that doesn't support calling internal methods using a `JsObject<T>`; adding support for that would require a slight redesign on the API. However, the current MVP is enough to support most of the use cases of the API in our builtin wrappers.

This also uses the new API to improve the `JsArrayBuffer` wrapper on some parts, but it still needs some more work that can be done on subsequent PRs.

Thanks to @HalidOdat for [nerd-sniping me into implementing this](https://github.com/boa-dev/boa/pull/3610#discussion_r1465893269) 🤣


HalidOdat (approved):
Great work @jedel1043! :)

nekevss (approved):
Looks good to me 😄

PR #3619, opened 2024-01-27
Fix escaping in `RegExp.prototype.source`

raskad:
This PR fixes the line terminator escaping in `RegExp.prototype.source`.


jedel1043 (approved):
Nice catch!

HalidOdat (approved):
Looks good to me! :)

PR #3627, opened 2024-01-29
Update regress to v0.8.0 and use UTF16 / UCS2 matching

raskad:

- Update regress to v0.8.0
- Use regress UTF16 / UCS2 matching
- Enable ignored regexp tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,6 +50,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


jedel1043 (approved):
Not much to review here because most of the work was done on `regress`, but it's really great to finally have proper UTF-16 support in most of the engine. Really great work!

nekevss (approved):

PR #3634, opened 2024-01-30
Implement resizable buffers

jedel1043:
A bit of a complex implementation, but happy with the new design overall. Open for suggestions though!

This implements:
- Resizable array buffers.
  - Allows `ArrayBuffer` to resize, but it's limited to a `maxByteLength` parameter passed at construction time.
  - Allows `SharedArrayBuffer` to grow (shrinking is disallowed by the spec), also limited to a `maxByteLength` parameter.
    It uses an `AtomicUsize` to track the current length, which should be enough to ensure thread-safety.
- Retrofits `Atomics`, `TypedArray` and `DataView` to use the new buffers.
- Makes `memmove`, `memcpy` and `copy_shared_to_shared` more low level, computing using pointers.
  The previous design used slices to do debug assertions before copying memory, but this made the API a bit clunkier to use, and some of the assertions weren't even needed by some callers.

EDIT: Forgot to mention that this doesn't add methods to resize `JsArrayBuffer` and `JsSharedArrayBuffer`, but this can be added later after some design work.

### DIFF
--- a/core/engine/src/builtins/array/array_iterator.rs
+++ b/core/engine/src/builtins/array/array_iterator.rs
@@ -117,15 +117,17 @@ impl ArrayIterator {
         }
 
         let len = if let Some(f) = array_iterator.array.downcast_ref::<TypedArray>() {
-            if f.is_detached() {
+            let buf = f.viewed_array_buffer().as_buffer();
+            let Some(buf) = buf
+                .bytes(std::sync::atomic::Ordering::SeqCst)
+                .filter(|buf| !f.is_out_of_bounds(buf.len()))
+            else {
                 return Err(JsNativeError::typ()
-                    .with_message(
-                        "Cannot get value from typed array that has a detached array buffer",
-                    )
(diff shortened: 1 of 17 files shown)


raskad (approved):
Great work!

Did you get the chance to run the 262 test suite in debug mode to test all of the debug assertions? Or switch them to asserts for one 262 run maybe.

jedel1043:
> Great work!
> 
> Did you get the chance to run the 262 test suite in debug mode to test all of the debug assertions? Or switch them to asserts for one 262 run maybe.

Yep, I ran the `TypedArray`, `ArrayBuffer` and `SharedArrayBuffer` suites and I caught some early bugs before opening the PR. I'll do another run of those just in case.

jedel1043:
Did another run on debug mode and everything looks good!

HalidOdat:
> Did another run on debug mode and everything looks good!

We might want to run the test262 tests in release with debug assertions in the CI, might help catch some bugs.

PR #3640, opened 2024-01-31
Remove double indirection in module types

jedel1043:
Should improve cache locality + allows allocating the whole `Module` struct in a single Gc.


HalidOdat (approved):
Nice change! Looks good to me! :)

raskad (approved):
Nice work

PR #3641, opened 2024-02-01
Fix line terminators in template strings

raskad:
This PR fixes some issues with different line terminators in templates strings.

The issues where caused by some unhandled line terminators in `raw` strings and double handling by `Cursor` of the string codepoints in the `cooked` strings.


HalidOdat (approved):
Nice catch! :)

jedel1043 (approved):

PR #3651, opened 2024-02-03
Temporal: Refactor Calendar protocol for `JsObject`s

nekevss:
This Pull Request is related to ongoing #1804.

So a little background.

There was a [test](https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/PlainDateTime/prototype/dayOfWeek/custom.js) I noticed that, in theory, we should have been passing, but weren't.

The reason we weren't was due to only passing the inner date representation through the library, and recreating the `PlainDate` once into the custom calendar impl. But this meant that we couldn't do the strict equality `compareArray` asserts.

So after some initial prototyping about an `AnyDateLike` trait. I had a sidebar with @jedel1043, and he had a good idea of using an associated trait. 😄

Let me know what you all think!

It changes the following:

- Adds `DateLike` associated Type to `CalendarProtocol`
- Updates the API for the above change.

### DIFF
--- a/core/engine/src/builtins/temporal/calendar/mod.rs
+++ b/core/engine/src/builtins/temporal/calendar/mod.rs
@@ -1074,12 +1074,7 @@ pub(crate) fn to_temporal_calendar_slot_value(
         }
 
         // c. Return temporalCalendarLike.
-        match calendar_like.clone().downcast::<Calendar>() {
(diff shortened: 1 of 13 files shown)


HalidOdat (approved):
Looks good to me! :)

jedel1043 (approved):

PR #3653, opened 2024-02-04
Simplify Temporal APIs

jedel1043:
I'm aware that this mostly undoes some API design decisions @nekevss and I did, but I was experimenting with simplifying the API and I think I prefer the simplicity of having a single contextual low level call for all methods, instead of a contextual and a non-contextual variant. If we do want to offer a protocol free implementation, I think it would be better to just make simple wrappers around the low level types.

On the other hand, this also replaces the `Any`s with an associated type inside `CalendarProtocol` itself. This wasn't possible at the beginning because I'm pretty sure we hadn't introduced the trait when we decided to pass `Any` to all method calls.

@nekevss Let me know what you think of the changes!

### DIFF
--- a/core/temporal/src/components/calendar.rs
+++ b/core/temporal/src/components/calendar.rs
@@ -8,7 +8,7 @@
 //! the calendar protocol), but it does aim to provide the necessary tools and API for
 //! implementing one.
 
-use std::{any::Any, str::FromStr};
+use std::str::FromStr;
 
 use crate::{
     components::{Date, DateTime, Duration, MonthDay, YearMonth},
@@ -88,13 +88,13 @@ impl From<&[String]> for CalendarFieldsType {
 #[derive(Debug)]
 pub enum CalendarDateLike<C: CalendarProtocol> {
     /// Represents a user-defined `Date` datelike
-    CustomDate(<<C as CalendarProtocol>::DateLikes as DateTypes<C>>::Date),
+    CustomDate(C::Date),
     /// Represents a user-defined `DateTime` datelike
-    CustomDateTime(<<C as CalendarProtocol>::DateLikes as DateTypes<C>>::DateTime),
+    CustomDateTime(C::DateTime),
     /// Represents a user-defined `YearMonth` datelike
-    CustomYearMonth(<<C as CalendarProtocol>::DateLikes as DateTypes<C>>::YearMonth),
+    CustomYearMonth(C::YearMonth),
     /// Represents a user-defined `MonthDay` datelike
-    CustomMonthDay(<<C as CalendarProtocol>::DateLikes as DateTypes<C>>::MonthDay),
+    CustomMonthDay(C::MonthDay),
     /// Represents a `DateTime<C>`.
     DateTime(DateTime<C>),
     /// Represents a `Date<C>`.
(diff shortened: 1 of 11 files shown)


core/temporal/src/components/zoneddatetime.rs
```diff
@@ -259,12 +207,12 @@ mod tests {
         )
         .unwrap();
 
-        assert_eq!(zdt.year().unwrap(), 2023);
-        assert_eq!(zdt.month().unwrap(), 11);
-        assert_eq!(zdt.day().unwrap(), 30);
-        assert_eq!(zdt.hour().unwrap(), 1);
-        assert_eq!(zdt.minute().unwrap(), 49);
-        assert_eq!(zdt.second().unwrap(), 12);
+        assert_eq!(zdt.year(&mut ()).unwrap(), 2023);
```
nekevss, line 210: issue: my main concern with this PR would be exactly this sort of API, which may be also where the faults in the previous API was.

My idea approaching this was eventually to have two distinct APIs. One for engine/custom calendar implementors and another for someone who would just want to use `Temporal` as their datetime library. I wasn't in love with `contextual_year`, but it was also meant to abstract as much of that API away from any potential users in the future that would just want to use the built-in calendars. These base API's could then be feature flagged down the line (something like `#[cfg(not(feature ="custom"))]` to remove them from the binary. `contextual_*` was meant to be a signal for any engine/custom calendar implementor. 

Granted, this all may be overthinking/premature abstraction or a wrong approach on my part. But I do think something of the sort should be considered especially if we were going to externalize the crate. Any thought regarding this?

core/temporal/src/components/calendar.rs
```diff
@@ -117,160 +117,164 @@ impl<C: CalendarProtocol> CalendarDateLike<C> {
...
-// TODO: DateTypes should implement a trait -> `ToTemporalDate`: `GetCalendarSlot`
-/// A trait for implementing `DateLike` types
-pub trait DateTypes<C: CalendarProtocol> {
+// ==== CalendarProtocol trait ====
+
+/// A trait for implementing a Builtin Calendar's Calendar Protocol in Rust.
+pub trait CalendarProtocol: Clone {
     /// A Custom `Date` Type for an associated `CalendarProtocol`. Default `Date<C>`
-    type Date: IsoDateSlots + GetCalendarSlot<C> + Clone + core::fmt::Debug;
+    type Date: IsoDateSlots + GetCalendarSlot<Self> + Clone + core::fmt::Debug;
```
nekevss, line 125: Praise: this is so much better! Let's go with this! 😄

nekevss (commented):
Overall, I think this looks great and is definitely a nice simplification. I do have one issue/discussion point.

core/temporal/src/components/zoneddatetime.rs
jedel1043: Oh wait, I remembered why I removed the methods instead of just changing them! Since now all methods take `&mut CalendarProtocol::Context`, we cannot pass a `&mut ()` to them on the general case. An idea I had was to have the current `API` but with a `C::Context: Default` bound on the impl, but that seemed very iffy because we could accidentally call any of the context-free methods on `Date<JsObject>` and it would just create a brand new `Context` for each call...

nekevss: > The main rationale for doing this was that many C and C++ APIs take a `void* data` parameter to pass context, and it's just second nature for users to pass `NULL` if they don't need additional context.

That is a good point. Hmmmmm.

As a counter, it may also be worth just avoiding that passing NULL convention as a whole, and just make sure it is extremely well documented that they should use the `context` API for that. In other words, I'm not sure making a Rust API more C-like is necessarily the best way. And over time, more and more Rust-based APIs will probably have a FFI with C/C++ so making something more "Rust-like" so to speak isn't necessarily bad and will probably be adapted to.

Granted, we should also make sure explicitly document that choice.

>   I also think we can offer high level types that are just wrappers around the "complex" date types:

Yeah, wrappers is another viable approach I had thought about. I think I'd actually prefer a wrapper approach even more as it would be nice to abstract away `Date<()>` for any users that wouldn't want or care about a custom calendar.

That sounds good to me overall. As said above, I really want to have the API in a bit more of a stable, agreed upon place. But I've mostly been trying to focus on implementing as much of the spec as possible.

I think there's probably 3 layers to consider:
 - `Date.year()` (fully abstracted)
 - `Date<Custom>.year()` (custom calendar without needed context)
 - `Date<Custom>.year_with_context(&mut Context)` (custom calendar with context)

nekevss (approved):
Looks good to me 😄 Really like the changes to simplify `CalendarProtocol`

HalidOdat (approved):
Looks good to me! I like the new design :)

PR #3669, opened 2024-02-09
Implement prototype of `NumberFormat`

jedel1043:
Implements the basics of `NumberFormat`, which includes all the digit formatting options (separators, numbering systems, significant digits/fractional digits, etc.).

ICU4X is still working on offering support for units, different notations, and `formatToParts`, but this prototype is enough to give correct results for things like:

```js
new Intl.NumberFormat("bn", {
    useGrouping: "min2",
    minimumSignificantDigits: 3,
    maximumSignificantDigits: 7
}).format(10003.1234) // outputs "১০,০০৩.১২"
```  

which is pretty nice.


core/engine/src/context/intrinsics.rs
```diff
@@ -29,15 +32,16 @@ pub struct Intrinsics {
 }
 
 impl Intrinsics {
-    pub(crate) fn new(root_shape: &RootShape) -> Self {
+    /// Creates a new set of uninitialized intrinsics.
+    pub(crate) fn uninit(root_shape: &RootShape) -> Option<Self> {
```
nekevss, line 36: thought: should a note be added to redirect anyone navigating the repo.

The API change makes sense to me, but if someone is navigating the repository for the first time. It could be useful to have a note to redirect where to look.

core/engine/src/builtins/intl/number_format/mod.rs
```diff
@@ -1,4 +1,848 @@
...
+
+        self.digit_options.format_fixed_decimal(value);
+        value.apply_sign_display(self.sign_display);
+
+        self.formatter.format(value)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(in crate::builtins::intl) struct NumberFormatLocaleOptions {
```
nekevss, line 82: question: couldn't this be `pub(super)` vs `pub(in crate::builtins::intl)`? Is there a benefit to using the latter over the former?

core/engine/src/builtins/intl/number_format/options.rs
```diff
@@ -180,3 +1208,48 @@ impl RoundingType {
         }
     }
 }
+
+#[cfg(test)]
```
nekevss, line 1212: nit: Separate the tests in to a `tests.rs` file.

With a project this large, it feels like it would be worthwhile to have the tests a bit more visible in their own file instead of in the file itself. If a potential contributor came in and wanted to add tests, they might not know that there were some tests already in place at a glance.

nekevss (approved):
This looks good to me! I did have a couple comments. Mostly nits and questions/thoughts. None of it is blocking though 😄

core/engine/src/builtins/intl/number_format/mod.rs
jedel1043: True.

core/engine/src/context/intrinsics.rs
jedel1043: Not sure what this means. Can you explain  a bit what exactly I should link?

nekevss: I meant it more of like an internal note for anyone trying familiarize themselves with the codebase. Something like:

```rust
// NOTE: Intrinsic objects are initialized by `Realm::Initialize`.  Individual initialization logic for each built-ins
// is located in their respective file `src/builtins/{object}/`.
```

Granted this could also always just be worked into the new site's contributor docs.

raskad (approved):
Great stuff. I have nothing to add.

HalidOdat (approved):
Great work! :)

PR #3675, opened 2024-02-15
Consider strict + no-strict tests as a single test

jedel1043:
A simplified version of #3364 that only fixes the conformance representation without doing anything else. Thanks to this, it should be completely retrocompatible with previous runs.


HalidOdat (approved):
Nice work! :)

PR #3682, opened 2024-02-19
Split ICU4X data generation from `boa_icu_provider`

jedel1043:
Thankss to #3681 I realized we could benefit from isolating our datagen dependencies from `boa_icu_provider`, which is why I decided to create a separate crate to run the datagen step.

This PR also adds a new flag to exclude the data from `boa_engine` if the user wants to.

# Size comparison

### Main
```
❯ ls -l core/icu_provider/data/icudata.postcard
❯ -rw-r--r-- 1 jedel users 16910774 Feb 18 22:05 core/icu_provider/data/icudata.postcard
```
### PR 
```
❯ ls -l core/icu_provider/data/icudata.postcard
❯ -rw-r--r-- 1 jedel users 16417284 Feb 19 00:29 core/icu_provider/data/icudata.postcard
```


core/engine/Cargo.toml
```diff
@@ -14,10 +14,15 @@ rust-version.workspace = true
 [features]
 profiler = ["boa_profiler/profiler"]
 deser = ["boa_interner/serde", "boa_ast/serde"]
-intl = [
+
+# Enables the `Intl` builtin object and bundles a default ICU4X data.
+intl = ["intl_core", "dep:boa_icu_provider"]
+
+# Enables only the `Intl` builtin implementation without any bundled ICU4X data provider.
+intl_core = [
```
HalidOdat, line 22: I think it would be better to keep the intl name for enabling intl functionality, and add name the flag that bakes the blob in the executable, something like `intl-blob`.

Calling it just intl to me sounds like we are activating all intl functionality, and intl_core sounds like you are enabling a sub part of the functionality.

HalidOdat (commented):
Besides the comment below this looks good to me!

core/engine/Cargo.toml
jedel1043: Will invert the names. About `intl-blob`, I think it would be better to pick a name that doesn't expose our internal representation of the ICU4X data, since we could migrate to other formats or even "baked" data in the future. Some other ideas:
- default_icu_data
- intl_data (preference)
- icu_intl

HalidOdat: `intl_data` seems good to me :)

raskad:
In the latest commit you removed the `intl_core` feature but did not revert the cfg annotations from `intl_core` to `intl`. I think that is why all intl tests are failing on the latest commit right? @jedel1043

HalidOdat (approved):

PR #3690, opened 2024-02-22
Change `HostEnsureCanCompileStrings` to the new spec

jedel1043:
PR https://github.com/tc39/ecma262/pull/3222 introduced a new `HostEnsureCanCompileStrings` that passes the body and parameters of the runtime-compiled function for additional checks.

The regressions are because the test repo hasn't been updated with the new behaviour; on the old spec,
the parameters were converted to strings before the body, which is the opposite of what the new spec states.

### DIFF
--- a/core/engine/src/builtins/eval/mod.rs
+++ b/core/engine/src/builtins/eval/mod.rs
@@ -108,10 +108,12 @@ impl Eval {
         // 3. Let evalRealm be the current Realm Record.
         // 4. NOTE: In the case of a direct eval, evalRealm is the realm of both the caller of eval
         // and of the eval function itself.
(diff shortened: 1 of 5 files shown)


jedel1043:
Opened a PR on the test262 repo: https://github.com/tc39/test262/pull/4013

raskad (approved):
Looks good. Also very nice refactor of `create_dynamic_function`.

I'm not sure if we want to wait for the discussion in https://github.com/tc39/test262/pull/4013 to be resolved. Imo we can merge this and if needed adjust later.

HalidOdat (approved):
Nice work! Agree with @raskad

PR #3692, opened 2024-02-26
Create tool to regenerate `ABOUT.md`

jedel1043:
Maybe this could also be added to CI? Still thinking about it because it's not as critical as the important tests.


raskad (approved):
Look good. I don't thing we have to add it in CI. Running it at least before every release seems enough.

nekevss (approved):
Agreed with the above comment about running it before a release.

PR #3694, opened 2024-02-26
Migrate to `temporal_rs` crate

nekevss:
This Pull Request moves migrates the current implementation over to the new `temporal_rs` crate.

It changes the following:

- Removes `boa_temporal`
- Adds functionality from `temporal_rs`


HalidOdat (approved):

nekevss:
So it looks like there's an infinite loop that's being triggered by one of the `Calendar.prototype.dateUntil` tests. I've commented out the call to `dateUntil` and returned an error. I'm going to open an issue in the `temporal_rs` repo to address the loop.

PR #3699, opened 2024-02-27
Check that `min <= max` in `clamp_finite`

### LINKED ISSUE #3697: Possible Incompleteness
In boa/core/engine/src/value/integer.rs, I think the code should check min<=max before calling clamp comments mentioned to avoid unnecessary panic.
```rust
/// # Panics
    ///
    /// Panics if `min > max`.
    #[must_use]
    pub fn clamp_finite(self, min: i64, max: i64) -> i64 {
        match self {
            Self::Integer(i) => i.clamp(min, max),
            Self::PositiveInfinity => max,
            Self::NegativeInfinity => min,
        }
    }
```

jedel1043:
Fixes #3697.
We should assert that `min <= max` even if `self` is `PositiveInfinity` or `NegativeInfinity`.

### DIFF
--- a/core/engine/src/value/integer.rs
+++ b/core/engine/src/value/integer.rs
@@ -22,6 +22,7 @@ impl IntegerOrInfinity {
     /// Panics if `min > max`.
     #[must_use]
     pub fn clamp_finite(self, min: i64, max: i64) -> i64 {
+        assert!(min <= max);
         match self {
             Self::Integer(i) => i.clamp(min, max),
             Self::PositiveInfinity => max,


raskad:
I checked all calls to `clamp_finite` and they all have `min = 0`. So we could do something like this and avoid the panics:

```rust
    pub fn clamp_finite_positive(self, max: i64) -> i64 {
        match self {
            Self::Integer(i) if i < 0 => 0,
            Self::Integer(i) if i > max => max,
            Self::Integer(i) => i,
            Self::PositiveInfinity => max,
            Self::NegativeInfinity => 0,
        }
    }
```

jedel1043:
I think that method would be useful internally to skip some panics, but we still need to expose a `clamp_finite` method for users of the `to_integer_or_infinity` method.

nekevss:
I'd be fine with a `clamp_finite_positive` method for avoiding panics. I'm not entirely sold on whether it should be exposed as `pub` off the start.

nekevss (approved):
I'm fine with moving forward with this as is for now.

raskad (approved):

PR #3703, opened 2024-02-28
Add a catch all for other categories not labelled

jasonwilliams:
Add a catch all for other categories not labelled

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -14,8 +14,6 @@ changelog:
     - title: Internal Improvements
       labels:
         - Internal
-    - title: Memory Optimizations
-      labels:
         - memory
     - title: Other Changes
       labels:
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -14,6 +14,9 @@ changelog:
     - title: Internal Improvements
       labels:
         - Internal
+    - title: Memory Optimizations
+      labels:
+        - memory
     - title: Other Changes
       labels:
         - "*"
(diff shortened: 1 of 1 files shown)


.github/release.yml
```diff
@@ -14,3 +14,6 @@ changelog:
     - title: Internal Improvements
       labels:
```
HalidOdat, line 15: There are some PRs that are labeled with the memory label we should probably add it to the internal improvements

HalidOdat (commented):
Besides the comment this looks good to me! :)

.github/release.yml
jasonwilliams: I've added a memory category

.github/release.yml
```diff
@@ -14,3 +14,9 @@ changelog:
...
         - Internal
+    - title: Memory Optimizations
+      labels:
+        - memory
```
HalidOdat, line 19:
```suggestion
        - Internal
        - memory
```

Can't it be added into as a internal improvement, instead of separate section?

jasonwilliams: Oh sorry I thought you meant you wanted it as a section, my bad

HalidOdat: :sweat_smile: On second thought just remove the memory label here, and add Internal where needed?

jasonwilliams: It won't hurt to add it to the internal improvements section, ill put it there for now and if we don't use it its not a big deal. We may want to categorise some things as memory in future

HalidOdat (approved):

jedel1043 (approved):

PR #3708, opened 2024-02-29
#3696 - WIP

### LINKED ISSUE #3696: Split `boa_tester` into crates
Now that `temporal_rs` is a thing, we need a way to test it against test262. We could reimplement `boa_tester` but it would be better to extract the test extraction part from the test runner to avoid duplicating our efforts.

PawelJastrzebski:
This Pull Request closes #3696.

It changes the following:

- todo

### DIFF
--- a/tests/tester/src/exec/mod.rs
+++ b/tests/tester/src/exec/mod.rs
@@ -3,8 +3,7 @@
 mod js262;
 
 use crate::{
-    read::ErrorType, Harness, Outcome, Phase, SpecEdition, Statistics, SuiteResult, Test,
-    TestFlags, TestOutcomeResult, TestResult, TestSuite, VersionedStats,
+    test252_parser::{ErrorType, Harness, Outcome, Phase, SpecEdition, Test, TestSuite}, Statistics, SuiteResult, TestFlags, TestOutcomeResult, TestResult, VersionedStats
 };
 use boa_engine::{
     builtins::promise::PromiseState,
--- a/tests/tester/src/git.rs
+++ b/tests/tester/src/git.rs
@@ -1,0 +1,142 @@
+use color_eyre::{eyre::bail, Result};
+use std::{path::Path, process::Command};
+
+/// Returns the commit hash and commit message of the provided branch name.
+fn get_last_branch_commit(directory: &str, branch: &str, verbose: u8) -> Result<(String, String)> {
+    if verbose > 1 {
+        println!("Getting last commit on '{branch}' branch");
+    }
+    let result = Command::new("git")
+        .arg("log")
+        .args(["-n", "1"])
+        .arg("--pretty=format:%H %s")
+        .arg(branch)
+        .current_dir(directory)
+        .output()?;
+
+    if !result.status.success() {
+        bail!(
+            "{directory} getting commit hash and message failed with return code {:?}",
+            result.status.code()
+        );
+    }
+
+    let output = std::str::from_utf8(&result.stdout)?.trim();
+
+    let (hash, message) = output
+        .split_once(' ')
+        .expect("git log output to contain hash and message");
+
+    Ok((hash.into(), message.into()))
+}
+
+fn reset_commit(directory: &str, commit: &str, verbose: u8) -> Result<()> {
+    if verbose != 0 {
+        println!("Reset {directory} to commit: {commit}...");
+    }
+
+    let result = Command::new("git")
+        .arg("reset")
+        .arg("--hard")
+        .arg(commit)
+        .current_dir(directory)
+        .status()?;
+
+    if !result.success() {
+        bail!(
+            "{directory} commit {commit} checkout failed with return code: {:?}",
+            result.code()
+        );
+    }
+
+    Ok(())
+}
+
+pub(super) fn clone(
+    directory: &str,
+    repor_url: &str, // "https://github.com/tc39/test262"
+    baranch: &str, // "origin/main"
+    commit: Option<&str>,
+    verbose: u8,
+) -> Result<()> {
+    let update = commit.is_none();
+
+    if Path::new(directory).is_dir() {
+        let (current_commit_hash, current_commit_message) =
+            get_last_branch_commit(directory, "HEAD", verbose)?;
+
+        if let Some(commit) = commit {
+            if current_commit_hash == commit {
+                return Ok(());
+            }
+        }
+
+        if verbose != 0 {
+            println!("Fetching latest {directory} commits...");
+        }
+        let result = Command::new("git")
+            .arg("fetch")
+            .current_dir(directory)
+            .status()?;
+
+        if !result.success() {
+            bail!(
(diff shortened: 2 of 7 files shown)


PawelJastrzebski (commented):
Merge comments

jedel1043 (commented):
Thank you for all the work!

It's been several days going back and forth on this PR about whether to put the feature to editions map into the new crate or the tester.

On the one hand, features are part of the test262 suite and we may include them to also get the edition of a test. On the other hand, if we include them, we would need to keep the crate pretty much updated with every new commit to the test262 repo, which would be a pretty big maintenance work for us.

I'm tending towards leaving the editions on the tester, since I don't see many use cases for it aside from engine tests. What do you think?

PawelJastrzebski:
I wonder if would be ok for tests that contains unknown features set edition to ESNext as a fallback option. I guess in that case they could be marked as ignored as well.

Thinking long term:
- test262 repo contains list of features [features.txt](https://github.com/tc39/test262/blob/main/features.txt). 
- We could read from that to find out if missing feature pass through the proposal stage, and only then break the parsing process with an error. 

image

jedel1043:
> I wonder if would be ok for tests that contains unknown features set edition to ESNext as a fallback option. I guess in that case they could be marked as ignored as well.

That would also be a good option for the short-term.

On the long term, I'm trying to push for adding this metadata to `test262` (https://github.com/tc39/test262/issues/4161). This would essentially remove the need for the map, probably making it a `HashMap<String, u8>` in the future.

tools/tc39-test262/src/edition.rs
```diff
@@ -39,12 +39,6 @@ static FEATURE_EDITION: phf::Map<&'static str, SpecEdition> = phf::phf_map! {
     // https://github.com/tc39/proposal-json-modules
     "json-modules"  => SpecEdition::ESNext,
 
-    // https://github.com/tc39/proposal-source-phase-imports
```
PawelJastrzebski: Duplicated

PawelJastrzebski:
Temporally fallback to ESNext while pending on features map on https://github.com/tc39/test262/issues/4161.

@jedel1043 Could you please take a look on this: 
https://github.com/boa-dev/boa/commit/391ac73dc72cc18da0c24cd9452611377ff9ce07#diff-91b6a7ace0b3edddaae2e4339b9cc565ce1cb47e735f2ad88805bcb533f3343aL390

I guess with that change we could merge this PR and implement features map parser later.

tools/tc39-test262/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+pub fn clone_test262(commit: Option<&str>, verbose: u8) -> color_eyre::Result<()> {
+    const TEST262_REPOSITORY: &str = "https://github.com/tc39/test262";
+    git::clone(
+        TEST262_DIRECTORY,
+        TEST262_REPOSITORY,
+        &"origin/main",
+        commit,
+        verbose,
+    )
+}
```
jedel1043, line 29: I'm not sure if we should expose a `clone_repo` function. IMO, we should provide a `test262` function that takes a struct of options and returns the whole suite, optionally cloning the repo if a path to a valid repo is not provided.

tools/tc39-test262/src/read.rs
```diff
@@ -11,85 +12,9 @@ use color_eyre::{
...
-    CanBlockIsFalse,
-    #[serde(rename = "CanBlockIsTrue")]
-    CanBlockIsTrue,
-    #[serde(rename = "non-deterministic")]
-    NonDeterministic,
-}
 
 /// Reads the Test262 defined bindings.
-pub(super) fn read_harness(test262_path: &Path) -> Result<Harness> {
+pub fn read_harness(test262_path: &Path) -> Result<Harness> {
```
jedel1043, line 17: Again, I think all of these should just be private, and the user should be able to explore the repo using methods in either `Test`, `TestSuite`, and friends.

tools/tc39-test262/src/test_files.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub expected_outcome: Outcome,
+    pub features: FxHashSet<Box<str>>,
+    pub includes: FxHashSet<Box<str>>,
+    pub locale: Locale,
+    pub ignored: bool,
+}
+
+impl Test {
+    /// Creates a new test.
+    pub fn new<N, C>(name: N, path: C, metadata: MetaData) -> color_eyre::Result<Self>
```
jedel1043, line 47: We should not use `eyre` for error handling. We should have an `Error` enum that people can match on.

jedel1043 (changes requested):
First pass of reviews. It's starting to look really nice!

tests/tester/src/exec/mod.rs
```diff
@@ -22,12 +19,41 @@ use colored::Colorize;
...
+    fn run(
+        &self,
+        harness: &Harness,
+        verbose: u8,
+        parallel: bool,
+        max_edition: SpecEdition,
+        optimizer_options: OptimizerOptions,
+        console: bool,
+    ) -> SuiteResult;
+}
```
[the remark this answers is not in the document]
PawelJastrzebski: You are right, we can remove generics.

PawelJastrzebski (commented):
Thanks for review @jedel1043

tools/tc39-test262/src/lib.rs
PawelJastrzebski: After longer consideration and refactor I not sure if reading and optionally cloning logic should be in the same function (double responsibility).  

- it would add more complexity to read(...) fn signature as we would have to provide extra information (struct) on each read level (harness, suite, test).
- if we provide it only on the suite level api in that form can be confusing and inconsistent. 

Usage in main: 

image

> in this example if path to suite would be a direct path to test (.js) test262 would skip clone and return error, causing inconsistent cli behavior.
 
 @jedel1043 please let me know what do you think about it. Maybe I'm missing something and there is a better solution to make it done.

PawelJastrzebski: @jedel1043  Thanks for clarification. 

Please take a look at this changes and let me know if there are closer to the expected result.
https://github.com/boa-dev/boa/pull/3708/commits/027fbb4f99d98cae5a9ec96a33ac542a181f8051?diff=split&w=0

Still some work is needed, but I want to make sure that this is the desired direction.

jedel1043:
Is this still being worked on? I assumed it was a WIP because you didn't re-request a review, but I'd be happy to review again if it is ready for that.

PawelJastrzebski:
Hi, @jedel1043,

Yes, it's still a work in progress. I haven’t had much time to work on it recently.

Could you take a quick look at `tests/tester/src/main.rs`? I have some doubts about the direction I've taken with the `ReadResult` enum.

If it looks good to you, I'll continue working on the error handling.

jedel1043:
Gonna close this for now since it got stale. Feel free to reopen if you want to still finish this contribution later.

PR #3725, opened 2024-03-07
Added NPM publish workflow

Razican:
This PR adds a release workflow to publish NPM packages on each release in an autonomous way.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
             skip
 
   npm_publish:
-    name: Publish crates
+    name: Publish NPM package (wasm)
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,6 +40,42 @@ jobs:
             --no-git-commit \
             skip
 
+  npm_publish:
+    name: Publish crates
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -40,6 +40,42 @@ jobs:
             --no-git-commit \
             skip
 
+  npm_publish:
+    name: Publish crates
```
nekevss, line 44: thought: Should the name of job be changed to: Publish boa_wasm.

Razican: Fixed :)

HalidOdat (approved):
Nice! looks good to me! :)

.github/workflows/release.yml
```diff
@@ -40,6 +40,42 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: wasm32-unknown-unknown
+          profile: minimal
+          override: true
```
jedel1043, line 57: Would prefer if we used Dtolnay's action, since actions-rs repos are unmaintained.

jedel1043 (changes requested):

.github/workflows/release.yml
Razican: I just noticed we are using the actions-rs actions everywhere in our codebase. Should we do this change separately, to migrate to the dtolnay's action?

jedel1043: Yep, that would be preferable.

jedel1043 (approved):
This will need a follow up PR to remove all actions from `actions-rs` from the codebase.

PR #3734, opened 2024-03-12
Close for-of iterator when the loop body throws

raskad:
Currently the iterator used in a for-of loop is not closed when the loop body throws. This PR fixes that by widening the scope of an exception handler that was already in place for some of the for-of assignment code paths.

To make this work I had to adjust the iterator close opcodes somewhat. The reason is that we may run into a situation when we attempt to close an iterator two times. See the following example:

```javascript
var iterable = {};
iterable[Symbol.iterator] = function() {
  return {
    next: function() {
      return { done: false, value: null };
    },
    return: function() {
      throw "iter-close-error";
    }
  };
};

function f() {
  for (var x of iterable) {
    return;
  }
}
f();
```

In this example we close the iterator when we `return` from `f` inside of the loop body. The iterator close code is injected during the compilation of the `return` statement. The problem is that if the `iterable.return` function throws, like in the example, we start running the exception handler code that also attempts to close the iterator. To avoid this I adjusted the iterator close opcodes.

### DIFF
--- a/core/engine/src/bytecompiler/statement/loop.rs
+++ b/core/engine/src/bytecompiler/statement/loop.rs
@@ -312,7 +312,8 @@ impl ByteCompiler<'_> {
             self.emit_with_varying_operand(Opcode::PushDeclarativeEnvironment, env_index);
         };
 
-        let mut handler_index = None;
+        let handler_index = self.push_handler();
+
         match for_of_loop.initializer() {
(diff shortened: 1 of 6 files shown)


HalidOdat (approved):
Nice catch! :)

jedel1043 (approved):

PR #3738, opened 2024-03-13
Add default value handling for destructuring property access arrays

raskad:

- Add default value handling for destructuring property access arrays


nekevss (approved):
LGTM!

PR #3743, opened 2024-03-16
Fix invalid syntax errors for allowed `let` as variable names

raskad:
This allows `let` as a variable name in some more places.


HalidOdat (approved):
Looks good to me! :)

jedel1043 (approved):

PR #3744, opened 2024-03-17
Remove array optimization path that prevents usage of setters on prototypes

raskad:
This PR removes an array optimization path in the `SetPropertyByValue` opcode.

The removed code path would insert an indexed property into the dense indexed properties of an array, if the index is the next item in the dense indexed properties.
The problem is that if the prototype of the array contains a setter at the index, this setter is ignored.

For example:
```js
Object.defineProperty(Array.prototype, '1', {
  set: function(param) {
    console.log("setter", param);
  }
});
let a = ["array"][1] = "value";
```

### DIFF
--- a/core/engine/src/vm/opcode/set/property.rs
+++ b/core/engine/src/vm/opcode/set/property.rs
@@ -1,7 +1,5 @@
-use boa_macros::utf16;
-
 use crate::{
-    builtins::{function::set_function_name, Proxy},
+    builtins::function::set_function_name,
     object::{internal_methods::InternalMethodContext, shape::slot::SlotAttributes},
     property::{PropertyDescriptor, PropertyKey},
     vm::{opcode::Operation, CompletionType},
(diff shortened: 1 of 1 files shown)


raskad:
@HalidOdat I would like to hear your opinion on the current solution of just removing the whole code path. The alternative would be to check the prototypes for an existing setter, but I think if we build that logic, we would just duplicate all of the code that the slow path does.
What do you think?

HalidOdat:
Hmmm... One thing we can do is have a bit flag on the shape, whether the object has a setter (already have bitflags [here](https://github.com/boa-dev/boa/blob/main/core/engine/src/object/shape/shared_shape/mod.rs#L43-L50) for shared shapes), and just check the flag before preforming the optimization.

This may be useful for optimizing some of the object internal methods too, since some of them check for a setter/getter.

HalidOdat:
I'll try to work on the shape flag :)

raskad:
Fixed in https://github.com/boa-dev/boa/pull/3760

PR #3745, opened 2024-03-18
Fix parsing of `async` in for-of loops

raskad:
This PR fixes a syntax error in for-of loops.

Currently the lookahead for the pattern `async of` in a for-of loop errors also in normal for loops. In a normal for loop this pattern might also be an async arrow function. To fix this I also had to fix the parsing of async arrow functions to allow `of` as a binding identifier. I added a test for that, since there seems to be no 262 test.

### DIFF
--- a/core/parser/src/parser/expression/assignment/mod.rs
+++ b/core/parser/src/parser/expression/assignment/mod.rs
@@ -141,15 +141,22 @@ where
                     1
                 };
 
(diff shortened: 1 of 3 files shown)


HalidOdat (approved):
Nice catch! :)

jedel1043 (approved):

PR #3751, opened 2024-03-19
Add missing environment creation in initial iteration of for loop

raskad:

- Add missing environment creation in initial iteration of for loop


jedel1043 (approved):
Good catch!

nekevss (approved):

PR #3752, opened 2024-03-19
Fix parsing of binding identifier in try catch parameters

raskad:

- Fix parsing of binding identifier in try catch parameters


HalidOdat (approved):
Looks great! :)

nekevss (approved):

PR #3755, opened 2024-03-19
Add a try_from_js implementation for Vec<T> (accept any Array-like)

hansl:
This PR adds an implementation of `TryFromJs` for `Vec<T: TryFromJs>` to create Rust vectors from Array-like JavaScript objects.

This is useful when using typed arguments to native functions with arrays.

### DIFF
--- a/core/engine/src/value/conversions/try_from_js.rs
+++ b/core/engine/src/value/conversions/try_from_js.rs
@@ -76,11 +76,14 @@ where
         let length = object
             .get(js_string!("length"), context)?
             .to_length(context)?;
-        let length = usize::try_from(length).map_err(|e| {
-            JsNativeError::typ()
-                .with_message(format!("could not convert length to usize: {e}"))
-                .into()
-        })?;
+        let length = match usize::try_from(length) {
+            Ok(length) => length,
+            Err(e) => {
+                return Err(JsNativeError::typ()
+                    .with_message(format!("could not convert length to usize: {e}"))
+                    .into());
+            }
+        };
         let mut vec = Vec::with_capacity(length);
         for i in 0..length {
             let value = object.get(i, context)?;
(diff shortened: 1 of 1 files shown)


core/engine/src/value/conversions/try_from_js.rs
```diff
@@ -58,6 +59,33 @@ where
...
+                return Err(JsNativeError::typ()
+                    .with_message("cannot convert value to a Vec")
+                    .into())
+            }
+        };
+
+        let length = object
+            .get(js_string!("length"), context)?
+            .to_length(context)?;
+        let mut vec = Vec::with_capacity(length as usize);
```
jedel1043, line 79: I'd avoid casting to `usize` and use `usize::try_from` with some added context in case the conversion fails.

jedel1043 (changes requested):
Thank you! I just have a small suggestion, but everything else looks good!

Razican:
The only concern here would be on how to implement this for TypedArrays, which should have a much faster conversion. Not sure if we would need to specialize or we could just add another branch.

hansl:
> The only concern here would be on how to implement this for TypedArrays

I suggest a specialized type, like `impl TryFromJs for Bytes`. There's no easy way to do that in Rust yet (see how serde handles `Vec<u8>` for example). Maybe when we have trait specialization in stable.

core/engine/src/value/conversions/try_from_js.rs
hansl: Done.

nekevss (approved):
Looks good to me! 😄 Had one nit, but that's not blocking merging.

core/engine/src/value/conversions/try_from_js.rs
```diff
@@ -250,3 +283,42 @@ impl TryFromJs for u128 {
         }
     }
 }
+
+#[test]
+fn value_into_vec() {
```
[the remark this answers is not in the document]
hansl: Done. What could I do to unblock this? It's blocking some of my follow work that I'd like to submit.

jedel1043 (approved):
Looks good! Thank you for the contribution!

PR #3758, opened 2024-03-22
Change dependabot interval to weekly

jedel1043:
It's getting overwhelming having to maintain so many dependencies in a daily basis, and some of them do two releases in succession, making the first version bump essentially useless. A weekly version bump is more maintainable in the long run, and if we require an urgent bump we can just update the version manually.


HalidOdat (approved):
Nice change! :)

raskad (approved):

PR #3760, opened 2024-03-23
Dense array storage variants for `i32` and `f64`

HalidOdat:
This PR adds dense variants for array storage (`i32` and `f64`) this is mostly a memory optimization, this optimization is also implemented by v8 (they have even more granular types like sparse `i32` and `f64` variants).

Storing number `i32` numbers in a dense array now takes `6x` less memory and `3x` less memory for `f64` (since `JsValue` is 24 bytes)

This PR also applies the changes in #3744 , because with this PR it complicates that fast path too much. Since the patch introduces a regression for setting new properties in arrays,  this PR amortizes the regression, it's pretty much the same with the current performance in quickjs benchmarks.

Even though I implemented that fast path, I was never happy with it :sweat_smile: (because it has too much duplicate logic), it was a quick way to improve performance, but I think the better option is to improve the internal object methods, which should give overall better performance.

### DIFF
--- a/core/engine/src/object/property_map.rs
+++ b/core/engine/src/object/property_map.rs
@@ -124,115 +124,63 @@ impl IndexedProperties {
             .collect()
     }
 
+    fn convert_to_sparse_and_insert(&mut self, key: u32, property: PropertyDescriptor) -> bool {
+        let mut map = match self {
+            Self::DenseI32(vec) => Self::convert_dense_to_sparse(vec),
+            Self::DenseF64(vec) => Self::convert_dense_to_sparse(vec),
+            Self::DenseElement(vec) => Self::convert_dense_to_sparse(vec),
+            Self::Sparse(map) => {
+                return map.insert(key, property).is_some();
+            }
+        };
+
+        map.insert(key, property);
+        *self = Self::Sparse(Box::new(map));
+        false
+    }
+
     /// Inserts a property descriptor with the specified key.
     fn insert(&mut self, key: u32, property: PropertyDescriptor) -> bool {
-        let mut map =
-            match self {
-                Self::DenseI32(vec) => {
-                    let len = vec.len() as u32;
-                    if key <= len
-                        && property.value().is_some()
-                        && property.writable().unwrap_or(false)
-                        && property.enumerable().unwrap_or(false)
-                        && property.configurable().unwrap_or(false)
-                    {
-                        // Fast Path: continues array access.
-                        let value = property.value().cloned().expect(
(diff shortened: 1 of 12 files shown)


raskad (approved):
Really nice optimization!

core/engine/src/object/property_map.rs
```diff
@@ -108,42 +126,170 @@ impl IndexedProperties {
...
+                        );
+
+                        // If it can fit in a i32 and the truncated version is
+                        // equal to the original then it is an integer.
+                        let is_rational_integer =
+                            |n: f64| n.to_bits() == f64::from(n as i32).to_bits();
+
+                        let value = match value {
+                            JsValue::Integer(n) => n,
+                            JsValue::Rational(n) if is_rational_integer(n) => n as i32,
```
jedel1043, line 151: Isn't this check also done when creating a `JsValue` from an f64? Why do we need to also do it here?

core/engine/src/object/property_map.rs
```diff
@@ -108,42 +126,170 @@ impl IndexedProperties {
...
-                        return false;
+        let mut map =
+            match self {
+                Self::DenseI32(vec) => {
+                    let len = vec.len() as u32;
+                    if key <= len
+                        && property.value().is_some()
+                        && property.writable().unwrap_or(false)
+                        && property.enumerable().unwrap_or(false)
+                        && property.configurable().unwrap_or(false)
```
jedel1043, line 137: Can we extract this check outside of the match? Maybe with a let else guard, so that we don't need to `expect` below on all branches.

jedel1043 (approved):
Nice work! I just have some suggestions.

core/engine/src/object/property_map.rs
HalidOdat: [Sadly no](https://github.com/boa-dev/boa/blob/main/core/engine/src/value/conversions/mod.rs#L49-L56), I thought of implementing that, the problem is that nothing prevents the user from manually constructing an `JsValue::Rational` that can fit in a `i32` because `JsValue` is a public enum.

We need to hide implementation detail from the user, we can create a struct that encapsulates the enum, that way it can only be constructed through it's respective constructors. (had similar issue when trying to keep the invariant in #3450 for `JsStr`)

I encountered this type of issue many times with `JsValue` and had to add the checks, so implementing this could enable a lot of optimizations :)

jedel1043: Could you open an issue for this? It could be a good first issue for someone.

jedel1043 (approved):

PR #3769, opened 2024-03-26
Optimize number to `PropertyKey` conversion

HalidOdat:
In previous implementation when converting a number to `PropertyKey`, we always converted it to string and only when the conversion to `PropertyKey::Index` was not possible we returned that string.

Instead of using `map_or` which expects a ready value we use the `_else` version that takes a function. Also took the oportunity to optimize the `i32` to `PropertyKey` conversion, since it's the one that's called when we convert from `JsValue::Integer`. Additionaly added a static string for `"length"` for internal array methods so we don't convert from `&[u16]` to `PropertyKey`.

Because the conversion happens very often this change resulted in big performance improvements :)

### Main

```
PROGRESS Richards
RESULT Richards 57.0
PROGRESS DeltaBlue
RESULT DeltaBlue 61.6
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 62.7
PROGRESS RayTrace
RESULT RayTrace 188
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 176
PROGRESS Splay
RESULT Splay 185
PROGRESS NavierStokes
RESULT NavierStokes 117
SCORE 107
```

### PR

```
PROGRESS Richards
RESULT Richards 58.8
PROGRESS DeltaBlue
RESULT DeltaBlue 63.7
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 71.7
PROGRESS RayTrace
RESULT RayTrace 198
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 182
PROGRESS Splay
RESULT Splay 193
PROGRESS NavierStokes
RESULT NavierStokes 158
SCORE 117
```

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -120,7 +120,7 @@ impl IntrinsicObject for Array {
                 Attribute::CONFIGURABLE,
             )
             .property(
-                utf16!("length"),
+                StaticJsStrings::LENGTH,
                 0,
                 Attribute::WRITABLE | Attribute::NON_ENUMERABLE | Attribute::PERMANENT,
             )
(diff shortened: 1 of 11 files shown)


raskad (approved):
Really nice find!

nekevss (approved):
Nice work!

PR #3771, opened 2024-03-27
Cache `this` value

HalidOdat:
The value of `this` when `this` is used in a function does not change after the initial environment chain search, so now we cache this value. This prevents the `this` value from being searched in the environment chain on every `this` usage.

There is a small increase in performance, the functions that use `this` a lot are the ones that benefit from this optimization.

### Main

```
PROGRESS Richards
RESULT Richards 58.0
PROGRESS DeltaBlue
RESULT DeltaBlue 63.9
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 69.9
PROGRESS RayTrace
RESULT RayTrace 198
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 181
PROGRESS Splay
RESULT Splay 192
PROGRESS NavierStokes
RESULT NavierStokes 150
SCORE 115
```

### PR

```
PROGRESS Richards
RESULT Richards 60.7
PROGRESS DeltaBlue
RESULT DeltaBlue 65.0
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 71.3
PROGRESS RayTrace
RESULT RayTrace 200
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 183
PROGRESS Splay
RESULT Splay 194
PROGRESS NavierStokes
RESULT NavierStokes 153
SCORE 117
```

### DIFF
--- a/core/engine/src/builtins/function/mod.rs
+++ b/core/engine/src/builtins/function/mod.rs
@@ -1054,10 +1054,16 @@ fn function_construct(
         Some(this)
     };
 
-    let frame = CallFrame::new(code.clone(), script_or_module, environments, realm)
+    let mut frame = CallFrame::new(code.clone(), script_or_module, environments, realm)
         .with_argument_count(argument_count as u32)
         .with_env_fp(env_fp)
         .with_flags(CallFrameFlags::CONSTRUCT);
+
+    // We push the `this` below so we can mark this function as having the this value
+    // cached if it's initialized.
+    frame
+        .flags
+        .set(CallFrameFlags::THIS_VALUE_CACHED, this.is_some());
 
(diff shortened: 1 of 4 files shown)


jedel1043 (approved):
Nice work!

Thought: I think in the future we could investigate if inverting the logic so that all contexts "assume" there's a `this` cached instead of the opposite is better; maybe using a `this` register on the VM that is pushed and popped from the stack when entering a new context. It would be interesting to see how many things break with this approach.

nekevss (approved):
Nice work! 😄

Edit: Second Jedel's thought

PR #3772, opened 2024-03-28
Add a boa_interop crate

hansl:
This crate will contain types and functions to help integrating boa in Rust projects, making it easier to interop between the host and the JavaScript code.

See https://github.com/boa-dev/boa/discussions/3770

### DIFF
--- a/core/gc/src/trace.rs
+++ b/core/gc/src/trace.rs
@@ -271,7 +271,6 @@ macro_rules! type_arg_tuple_based_finalize_trace_impls {
     ($(($($args:ident),*);)*) => {
         $(
             fn_finalize_trace_group!($($args),*);
-            // closure_finalize_trace_group!($($args),*);
             tuple_finalize_trace!($($args),*);
         )*
     }
--- a/core/gc/src/trace.rs
+++ b/core/gc/src/trace.rs
@@ -271,6 +271,7 @@ macro_rules! type_arg_tuple_based_finalize_trace_impls {
     ($(($($args:ident),*);)*) => {
         $(
             fn_finalize_trace_group!($($args),*);
+            // closure_finalize_trace_group!($($args),*);
             tuple_finalize_trace!($($args),*);
         )*
     }
--- a/core/interop/src/lib.rs
+++ b/core/interop/src/lib.rs
@@ -1,4 +1,5 @@
 //! Interop utilities between Boa and its host.
+
 use std::cell::RefCell;
 
 use boa_engine::module::SyntheticModuleInitializer;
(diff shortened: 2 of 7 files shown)


core/interop/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+                s.borrow_mut()();
+                Ok(JsValue::undefined())
+            })
+        }
+    }
+}
+
+#[test]
+#[allow(clippy::missing_panics_doc)]
+pub fn into_js_module() {
```
nekevss, line 58: thought: could this test be moved into a doc test

This crate is going to be heavily user facing, so having a test like this visible in rustdoc will probably be incredibly important.

core/interop/src/loaders/hashmap.rs
```diff
@@ -0,0 +1,63 @@
...
+impl Default for HashMapModuleLoader {
+    fn default() -> Self {
+        Self(GcRefCell::new(FxHashMap::default()))
+    }
+}
+
+impl HashMapModuleLoader {
+    /// Creates an empty `HashMapModuleLoader`.
+    #[must_use]
+    pub fn new() -> Self {
```
nekevss, line 23: suggestion/issue: not sure that new is needed here if all it does is call `Default`

nekevss (commented):
Nice work on this! Thanks for presenting it and taking it on!

Below is some initial feedback after a quick overview.

core/interop/src/loaders/hashmap.rs
hansl: Some people don't like to import default (it's not part of std prelude). I've seen this pattern in places, including the rust std: https://github.com/rust-lang/rust/blob/master/library/std/src/collections/hash/map.rs#L229-L234

I'm okay replacing it with `empty()` too if you want.

core/interop/src/lib.rs
hansl: This test is going to change very soon. I could make it into its own spec test, but I'll setup doc tests when I add the next layer of abstraction if you don't mind. Basically I want to add Traits that make this around ~15 lines instead of the monstrosity of boilerplate it currently is.

Basically it's very ugly and I'd rather have doctest and examples that are higher in abstraction (for the sake of the reader). WDYT?

hansl: For reference, I've created a PR that's based on this PR with the next step I want to work on: https://github.com/boa-dev/boa/pull/3773

There is a doctest for the `IntoJsFunction`, but still none for `IntoJsModule`.

core/interop/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+/// A trait to convert a type into a JS function.
+pub trait IntoJsFunction {
+    /// Converts the type into a JS function.
+    fn into_js_function(self, context: &mut Context) -> NativeFunction;
+}
+
+impl<T: FnMut() + 'static> IntoJsFunction for T {
```
jedel1043, line 43: This is probably UB, or at least it could cause a memory leak if `T` captures `Gc` types within it.

The reason we require `Copy` on `NativeFunction::from_copy_closure` is because Rust currently doesn't allow inspecting the captured variables of a closure. This makes it impossible to trace garbage collected variables that are captures of a closure. However, `Gc` is only `Clone`, making it impossible to be inside a `Fn() + Copy`.

jedel1043 (changes requested):

core/interop/src/lib.rs
hansl: Okay I see it now. That makes using closures very hard :( I need to experiment a bit.

jedel1043: I think this could probably be fine if the unsafe marker is lifted to the user, so that the user has to ensure the closures don't contain `Gc` types. This would probably imply splitting the current trait into two: a safe variant implementable by `Fn + Copy` and an unsafe variant implementable by `FnMut`.

hansl: Okay, I've added unsafe to both the trait and the method, will implement a safe version separately. PTAL.

jedel1043: Yep, seems fine to start with an unsafe trait for now. We can always adjust the design after designing the macros first.

jedel1043 (commented):
Nothing

core/gc/src/trace.rs
```diff
@@ -271,6 +271,7 @@ macro_rules! type_arg_tuple_based_finalize_trace_impls {
     ($(($($args:ident),*);)*) => {
         $(
             fn_finalize_trace_group!($($args),*);
+            // closure_finalize_trace_group!($($args),*);
```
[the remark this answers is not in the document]
hansl: Done, good catch.

HalidOdat (approved):
Great work! Looks good to me! :)

jedel1043 (approved):

PR #3773, opened 2024-03-29
Boa interop 2

hansl:
This is getting closer to where I want this crate to be, at least as a first milestone.


core/interop/src/lib.rs
```diff
@@ -39,67 +38,255 @@ impl<T: IntoIterator<Item = (JsString, NativeFunction)> + Clone> IntoJsModule fo
...
+/// f.call(&JsValue::undefined(), &[JsValue::from(1)], &mut context).unwrap();
+/// f.call(&JsValue::undefined(), &[JsValue::from(4)], &mut context).unwrap();
+/// assert_eq!(*x.borrow(), 5);
+/// ```
+///
 /// # Safety
 /// For this trait to be implemented safely, the implementing type must not contain any
 /// garbage collected objects (from [`boa_gc`]).
-pub unsafe trait IntoJsFunctionUnsafe {
+pub unsafe trait IntoJsFunctionUnsafe<Args, Ret> {
```
jedel1043, line 79: I think we can seal both `IntoJsFunction*` traits to remove the `unsafe` marker on it. Those will be provided by us anyways. That would also allow us to rename them to `IntoJsFunction` and `IntoJsFunctionCopied`, which is a bit more explanatory.

core/interop/src/into_js_function_impls.rs
```diff
@@ -0,0 +1,154 @@
...
+            #[allow(unused_variables)]
+            unsafe fn into_js_function_unsafe(self, _context: &mut Context) -> NativeFunction {
+                let s = RefCell::new(self);
+                unsafe {
+                    NativeFunction::from_closure(move |this, args, ctx| {
+                        let rest = args;
+                        $(
+                            let ($id, rest) = $t::try_from_js_argument(this, rest, ctx)?;
+                        )*
+                        let r = s.borrow_mut()( $($id,)* ctx);
```
jedel1043, line 53: Instead of panicking, it would be better to use `try_borrow_mut` and return a nice error.

core/interop/src/lib.rs
hansl: Good idea, done.

hansl: > That would also allow us to rename them to IntoJsFunction and IntoJsFunctionCopied, which is a bit more explanatory.

I'd like to keep the `Unsafe` suffix on the unsafe one though. Keeps users on their toes :)

jedel1043: That's also fine. Let's use `UnsafeIntoJsFunction` and `IntoJsFunction` then.

hansl: Done. PTAL

core/interop/src/lib.rs
```diff
@@ -40,67 +46,251 @@ impl<T: IntoIterator<Item = (JsString, NativeFunction)> + Clone> IntoJsModule fo
...
+            None => T::try_from_js(&JsValue::undefined(), context).map(|v| (v, rest)),
         }
     }
 }
 
+/// An argument that when used in a JS function will empty the list
+/// of JS arguments as `JsValue`s. This can be used for having the
+/// rest of the arguments in a function.
+#[derive(Debug, Clone)]
+pub struct JsRest(pub Vec<JsValue>);
```
jedel1043, line 160: Does this need to be a `Vec<JsValue>`? I would argue it would be better to have it be `&[JsValue]` to avoid unnecessary allocations, but I don't know if there are technical challenges that disallow this.

jedel1043 (commented):
Just have two small suggestions. Everything else looks great!

core/interop/src/lib.rs
```diff
@@ -40,67 +46,251 @@ impl<T: IntoIterator<Item = (JsString, NativeFunction)> + Clone> IntoJsModule fo
...
+/// assert_eq!(*x.borrow(), 5);
+/// ```
+pub trait UnsafeIntoJsFunction<Args, Ret>: private::IntoJsFunctionSealed<Args, Ret> {
     /// Converts the type into a JS function.
     ///
     /// # Safety
     /// This function is unsafe to ensure the callee knows the risks of using this trait.
     /// The implementing type must not contain any garbage collected objects.
-    unsafe fn into_js_function(self, context: &mut Context) -> NativeFunction;
+    unsafe fn into_js_function_unsafe(self, context: &mut Context) -> NativeFunction;
```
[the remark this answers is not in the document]
hansl: I'll rename the copy form in `..._copied` but I think we should keep `_unsafe`. I've been using these helpers in my project and sometimes it does get confusing in the error messages (mostly because the type system just name the type with the file name and lines).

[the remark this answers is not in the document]
jedel1043: Can you give some examples of that? I cannot see immediately why adding the `_unsafe` suffix would improve error messages.

core/interop/src/lib.rs
hansl: It turns out that it becomes a bit convoluted for the trait implementation. I'd need to make changes to `NativeFunction`.

jedel1043: We can do incremental improvements later then.

core/interop/src/lib.rs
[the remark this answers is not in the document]
jedel1043: I don't really want to block the PR, so I'm resolving this. We can bikeshed later.

core/interop/src/into_js_function_impls.rs
```diff
@@ -0,0 +1,180 @@
...
+        where
+            $($t: TryFromJsArgument + 'static,)*
+            R: TryIntoJsResult,
+            T: Fn($($t,)*) -> R + 'static + Copy,
+        {
+            #[allow(unused_variables)]
+            fn into_js_function_copied(self, _context: &mut Context) -> NativeFunction {
+                let s = self;
+                unsafe {
+                    NativeFunction::from_closure(move |this, args, ctx| {
```
jedel1043, line 90: In this case, I think you can use `from_copy_closure` instead.

core/interop/src/try_into_js_result_impls.rs
```diff
@@ -0,0 +1,55 @@
+//! Declare implementations of [`TryIntoJsResult`] trait for various types.
+//! We cannot rely on a generic implementation based on `TryIntoJs` due
+//! to a limitation of the Rust type system which prevents generalization
+//! of traits based on an upstream crate.
```
jedel1043, line 4: If this is the reason why we can't rely on `TryIntoJs`, maybe this should be moved to the main crate?

core/interop/src/lib.rs
```diff
@@ -40,67 +46,251 @@ impl<T: IntoIterator<Item = (JsString, NativeFunction)> + Clone> IntoJsModule fo
...
+    /// error that happened during the call to a function.
+    fn try_into_js_result(self, context: &mut Context) -> JsResult<JsValue>;
+}
+
+mod try_into_js_result_impls;
+
+/// Create a Rust value from a JS argument. This trait is used to
+/// convert arguments from JS to Rust types. It allows support
+/// for optional arguments or rest arguments.
+pub trait TryFromJsArgument: Sized {
```
jedel1043, line 130: Thought: A bit worried that this allows having two or more `JsRest` arguments on a function, but I'm really not sure what we could do to disallow this at the type level.

jedel1043 (approved):
Really nice API! I'm preemptively approving considering that we'll work on this incrementally, but left some additional comments for discussion.

core/interop/src/try_into_js_result_impls.rs
hansl: Sure, I'll make this, but probably just reimplement `TryInto<JsResult<>>` if possible. That will be simpler. I was going to seal this trait but no need, and being able to extend it is useful.

core/interop/src/lib.rs
hansl: Yes, I thought about it. Having 2+ `JsRest` basically means the other ones are empty.

What I could also do is the same trick I got with `Context`; only allow `JsRest` at the end of the arguments. WDYT?

jedel1043: That would be pretty great! And it makes sense to always require to declare the rest arguments last, so it checks out.

hansl: Here's another alternative, rename `JsRest` to `JsAll<T> for T: TryFromJs` and the first value to fail the `TryFromJs` will return. WDYT?

hansl: That way I can do `JsAll<JsString>` to get all the strings in the argument. Or `JsAll<JsValue>` to get the rest of arguments.

jedel1043: Sounds fine, but how does that solve the problem of specifying two or more rest arguments?

hansl: Alright I did both. You can still do `JsAll<i32>, JsAll<i32>` but that's user error I'd say.

jedel1043: Sounds good to me!

jedel1043 (approved):

core/interop/src/into_js_function_impls.rs
hansl: I seemed to have missed this. Done.

core/interop/src/lib.rs
```diff
@@ -40,67 +46,330 @@ impl<T: IntoIterator<Item = (JsString, NativeFunction)> + Clone> IntoJsModule fo
...
+///
+/// let result = sums.call(
+///     &JsValue::undefined(),
+///     &[JsValue::from(1), JsValue::from(2), JsValue::from(3), JsValue::Boolean(true), JsValue::from(4)],
+///     &mut context
+/// ).unwrap();
+/// assert_eq!(result, JsValue::new(6));
+/// ```
+#[derive(Debug, Clone)]
+pub struct JsAll<T: TryFromJs>(pub Vec<T>);
```
nekevss, line 235: question: does `Vec<T>` need to be `pub` here with `into_inner`.

I think it makes sense for these to be `pub(crate)` to begin unless there is a specific need for them to be `pub`.

nekevss (commented):
This is looking great! I had one small question.

core/interop/src/lib.rs
hansl: It's more a convenience. This way I can declare

```rust
fn my_js_fn(JsAll(all_numbers): JsAll<i32>) { ... } 
```

I'd rather remove `into_inner()` than remove the `pub`, tbh.

nekevss (approved):
LGTM :)

PR #3779, opened 2024-04-02
Optimize Regex match check

HalidOdat:


jedel1043:
@raskad This is a pretty big optimization because it removes O(n^2) complexity from our match implementation, but we are not sure how to adapt this to be able to work on unicode matching. Can you give this a look?

HalidOdat:
Reduced the failing tests from 48 to 4 unicode ones.

HalidOdat:
After some more debugging I think it's an issue with regresses `InputUtf16` iterator, managed to get a minimal testcase. We are trying to match against the lower part of the unicode 𝌆 this should not be possible in unicode mode since it's unicode aware. (In non-unicode mode it matches as expected).

```rs
/// 262 test/built-ins/RegExp/prototype/exec/u-lastindex-adv.js
///
/// Test case:
///
/// ```JavaScript
/// assert.sameValue(/\udf06/u.exec('\ud834\udf06'), null);
/// ```
#[test]
fn utf16_correct_unicode_scan() {
    // '𝌆' This is "Tetragram For Centre"
    // See: https://www.compart.com/en/unicode/U+1D306
    const INPUT: &[u16] = &[0xd834, 0xdf06];
    const MATCHER: &[u16] = &[0xdf06];

    let regex = Regex::from_unicode(MATCHER.iter().copied().map(u32::from), Flags::from("u"))
        .expect("valid regex");
    let m = regex.find_from_utf16(INPUT, 0).next();

    println!("{m:#?}");

    assert!(m.is_none());
}

```

There is a match:
```
Some(
    Match {
        range: 1..2,
        captures: [],
        named_captures: {},
    },
)
```

core/engine/src/builtins/regexp/mod.rs
```diff
@@ -943,72 +943,65 @@ impl RegExp {
...
                 if global || sticky {
                     // 1. Perform ? Set(R, "lastIndex", +0𝔽, true).
                     this.set(utf16!("lastIndex"), 0, true, context)?;
                 }
 
-                // ii. Return null.
                 return Ok(None);
             }
+            // i. Assert: r is a State.
+            Some(m) => {
```
jedel1043, line 986: Nit: this could be more readable with a let else.

core/engine/src/builtins/regexp/mod.rs
```diff
@@ -943,72 +943,65 @@ impl RegExp {
         // 11. If fullUnicode is true, let input be StringToCodePoints(S). Otherwise, let input be a List whose elements are the code units that are the elements of S.
         // 12. NOTE: Each element of input is considered to be a character.
 
+        // TODO: Comment spec deviation
+        // TODO: It would be better to put this in an enum.
+        //       enum Matches { Utf16(..), Ucs2(..) }
```
jedel1043, line 948: Couldn't we just do the construction and the call to `next` inline? I'm seeing those are only used in the `r` assignment.

HalidOdat: Yes, will change that, leftover when there was a loop :sweat_smile:

raskad:
The tests seem fine with the new regress version so I think the only thing missing is the todo comment

HalidOdat:
I put the comments this is ready for review/merge :)

raskad (approved):
Great find and optimization!

jedel1043 (approved):

PR #3784, opened 2024-04-02
Add an embed_module!() macro to boa_interop

hansl:
The macro creates a ModuleLoader that includes all JS files from a directory.


hansl:
@jedel1043 this would be ready for review. My use case for this is to have a build process which uses parcel, then embedding the resulting dist folder in the executable.

core/interop/src/loaders/embedded.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            if !acc.is_empty() {
+                acc.push('/');
+            }
+            acc.push_str(component);
+            Ok(acc)
+        })?;
+
+    Ok(specifier.into())
```
jedel1043, line 40: I think you can parse directly to `Path` in this case, which should allow using something like cargo's `normalize_path` impl (https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61)

hansl: We should actually have a NewType for `JsSpecifier` that matches the semantics of `Path`. But I'm too lazy. In this case I cannot use `path` as for Windows `/`s are not going to be understood. That's also why I didn't just transform into `Path` elsewhere when doing resolution.

jedel1043: > We should actually have a NewType for `JsSpecifier` that matches the semantics of `Path`

I don't think we can have that, because some embedders will want to use URLs as specifiers, so that's probably discarded.

> In this case I cannot use path as for Windows /s are not going to be understood. That's also why I didn't just transform into Path elsewhere when doing resolution.

Would this be solved if you used a strict `UnixPath` type, such as the one offered by [`typed-path`](https://crates.io/crates/typed-path)?

hansl: > URLs as specifiers, so that's probably discarded.

A `JsSpecifier` should cover everything that the JavaScript specification says can go in an import specifier. Then allow to convert to other types.

jedel1043: > A JsSpecifier should cover everything that the JavaScript specification says can go in an import specifier. Then allow to convert to other types.

When I say "cover the semantics of Path" what I mean is being able to iterate over components and such. Specifiers in JavaScript are, AFAIK, a subset of UTF-16 (can you put emojis or RTL/LTR controls in there?) and definitely a subset of OsStr.

Ah, I misunderstood. Yeah, that would be nice, maybe have some methods on there to convert from `JsSpecifier` to `JsPath`, `JsUrl`, etc.

> Maybe? I was trying to keep the number of dependencies low. If you're okay with it I can move to typed-path, it seems maintained.

Yep, I think it should be fine in this case because the crate is nicely maintained and it could possibly be added to the std library in the future.

hansl: Adding a test, I found a bug, so I'm fixing that too. Was using the old `SimpleModuleLoader` that didn't take the referrer into account.

core/interop/src/loaders/embedded.rs
```diff
@@ -0,0 +1,141 @@
...
+    ($path: literal) => {
+        embed_module!($path, max_size = 1_048_576)
+    };
+}
+
+#[derive(Debug, Clone)]
+enum EmbeddedModuleEntry {
+    Source(JsString, &'static [u8]),
+    Module(Module),
+    Error(JsError),
```
jedel1043, line 35: Can you explain why do we need an `Error` variant on the entry?
I would expect that `cache` would throw if parsing fails, instead of caching the fail.

hansl: Optimize the case where the same module is re-imported? There's no need to reparse it.

jedel1043: Yeah, but it's not common to cache the parse error because in most cases execution will bail on the first error.

hansl: Sure, removed.

jedel1043 (approved):
LGTM, just needs to appease clippy.

nekevss (approved):
LGTM :)

PR #3785, opened 2024-04-03
Add combinatorial basic module loaders in boa_interop

hansl:
These can be useful in general for creating chains of custom module loaders without much boilerplate.

### DIFF
--- a/core/interop/src/loaders/merge.rs
+++ b/core/interop/src/loaders/merge.rs
@@ -1,0 +1,57 @@
+//! A [`ModuleLoader`] that "merges" two module loaders into one. It will
+//! try to resolve using the first loader, and if it fails, it will try the
+//! second.
+//!
+//! Any errors from the first loader will be ignored, and the second loader
+//! will be used.
+
+use boa_engine::{Context, JsResult, JsString, Module};
+use boa_engine::module::{ModuleLoader, Referrer};
+
+pub struct MergeModuleLoader<First, Second> {
+    first: First,
+    second: Second,
+}
+
+impl<First: ModuleLoader, Second: ModuleLoader> From<(First, Second)>
+    for MergeModuleLoader<First, Second>
+{
+    fn from((first, second): (First, Second)) -> Self {
+        Self::new(first, second)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


core/interop/src/loaders/predicate.rs
```diff
@@ -0,0 +1,183 @@
...
+    ///
+    /// If the base is empty, that last verification will be skipped. This can be used
+    /// for module loaders that don't have an actual filesystem as their backend
+    /// (such as [`crate::loaders::HashMapModuleLoader`]).
+    ///
+    /// # Errors
+    /// This predicate will return an error if the specifier is relative but the referrer
+    /// does not have a path, or if the resolved path is outside `base`.
+    #[inline]
+    pub fn path_resolver(base: PathBuf) -> impl Fn(Option<&Path>, JsString) -> JsResult<JsString> {
```
hansl, line 102: @jedel1043 This function will fix the bug in the `SimpleModuleLoader`. I will move this code into the `SimpleModuleLoader` as a free function, and use that function here instead. It should make it easier.

I'll duplicate the tests as well.

hansl:
@jedel1043 This is ready to review when you're ready.

core/interop/src/loaders/merge.rs
```diff
@@ -0,0 +1,68 @@
...
+use boa_engine::{Context, JsResult, JsString, Module};
+
+/// A [`ModuleLoader`] that "merges" two module loaders into one. It will
+/// try to resolve using the first loader, and if it fails, it will try the
+/// second.
+#[derive(Debug)]
+pub struct MergeModuleLoader<First, Second> {
+    first: First,
+    // When resolving modules we need to clone the second loader.
+    second: Rc<Second>,
```
jedel1043, line 21: Thought: Hmmm, I don't really like this because we already wrap `MergeModuleLoader` in an `Rc`. Maybe it's time to commit to the `Rc` wrapper and change the receiver of the `ModuleLoader` methods to `Rc<Self>`.

jedel1043: Not actionable from your side, I'm just documenting my thoughts.

PR #3786, opened 2024-04-04
Add a new type Convert<> to convert values 

hansl:
This can be used to transform values in to converted versions of Rust types by using the type system and `TryFromJs`.

Also added a JsString conversion for `TryFromJs`. These methods are going to be useful when #3773 goes in (but this PR nor that one depend on each others).

### DIFF
--- a/core/engine/src/value/conversions/coerce.rs
+++ b/core/engine/src/value/conversions/coerce.rs
@@ -2,7 +2,7 @@
 //! converting.
 
 use crate::value::TryFromJs;
-use crate::{Context, JsResult, JsValue};
+use crate::{Context, JsResult, JsString, JsValue};
 use boa_engine::JsNativeError;
 
 /// A wrapper type that allows coercing a `JsValue` to a specific type.
@@ -103,7 +103,7 @@ impl TryFromJs for Coerce<JsString> {
 }
 
 impl TryFromJs for Coerce<bool> {
-    fn try_from_js(value: &JsValue, context: &mut Context) -> JsResult<Self> {
+    fn try_from_js(value: &JsValue, _context: &mut Context) -> JsResult<Self> {
         Ok(Self(value.to_boolean()))
     }
 }
--- a/core/engine/src/value/conversions/coerce.rs
+++ b/core/engine/src/value/conversions/coerce.rs
@@ -1,27 +1,46 @@
 //! Types and functions for applying Coercing rules to [`JsValue`] when
 //! converting.
 
+use boa_engine::JsNativeError;
+
 use crate::value::TryFromJs;
 use crate::{Context, JsResult, JsString, JsValue};
-use boa_engine::JsNativeError;
 
 /// A wrapper type that allows coercing a `JsValue` to a specific type.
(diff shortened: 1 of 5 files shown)


jedel1043 (commented):
I think this shouldn't be called `Coerce`, since it's not really an implicit conversion. Instead, it should be called `Cast<T>`, which makes sense as a name for an explicit conversion.

hansl:
`Coerce` (I was hesitating with `Coerced`) is the proper JavaScript term though when doing implicitly. When explicitly doing `Number("123")` it is called type conversion, so I'd be fine with `Convert`.

`Cast` in Rust is explicitly using `... as ...` which is different than JavaScript, so I'd suggest not using that word. Notably this new type can fail when the original type is utterly impossible to cast (`NaN` to `i8` for example), which is impossible with a (Rust) cast.

WDYT?

jedel1043:
Yeah, `Convert` is better.

hansl:
@jedel1043 ping for this. This is ready to review.

core/engine/src/value/conversions/coerce.rs
```diff
@@ -0,0 +1,128 @@
...
+/// let value1 = JsValue::Integer(1);
+/// let value_nan = JsValue::Rational(f64::NAN);
+/// let Convert(coerced0): Convert<bool> = Convert::try_from_js(&value0, &mut context).unwrap();
+/// let Convert(coerced1): Convert<bool> = Convert::try_from_js(&value1, &mut context).unwrap();
+/// let Convert(coerced_nan): Convert<bool> = Convert::try_from_js(&value_nan, &mut context).unwrap();
+///
+/// assert_eq!(coerced0, false);
+/// assert_eq!(coerced1, true);
+/// assert_eq!(coerced_nan, false);
+/// ```
```
jedel1043, line 39: Nit: It would be good to remove all references to "coerce" or "coercion" in both the docs and the variable names, for consistency. Not marking this as a critical change because it's not really important.

jedel1043 (approved):
Looks great! Just a small nitpick that doesn't block merging.

core/engine/src/value/conversions/coerce.rs
hansl: Done, I've also added a link to the section 9 of the spec that explains the conversion rules.

HalidOdat:
I was thinking maybe we should add this logic to the `TryFromJs` instead of having a separate struct and just using the `try_from_js` directly:

```rust
let int = i32::try_from_js(&JsValue::from(10), context)?;
```

Is there any benefit to having a separate struct?

jedel1043:
The problem is that `TryFromJs` has "lossless" semantics, which means that it only succeeds if the output type is roughly the same as the `JsValue` type (no `to_string`, `to_number`, `to_boolean`, etc).
If we want to do the change you're proposing, we would have to change the semantics of `TryFromJs`, and declare that it allows "lossy" conversions.

I tend towards not doing that because I don't really like to have implicit conversions in Rust APIs, but we could discuss with the team if this is a behaviour we want from `TryFromJs`.

raskad (approved):

HalidOdat (approved):

PR #3791, opened 2024-04-04
Properly resolve paths in SimpleModuleLoader and add path to Referrer::Script

### LINKED ISSUE #3790: `Context::get_active_script_or_module` should return a script or module when using import in a generator
This is a pretty hard problem to fix for me at the moment, as I lack a lot of the internal knowledge around realms and call frames.

In the test262 `test/language/expressions/dynamic-import/for-await-resolution-and-error-agen-yield.js`, the following code:

```js
async function * agen1() {
    yield import('./for-await-resolution-and-error-a_FIXTURE.js');
    yield import('./for-await-resolution-and-error-b_FIXTURE.js');
    yield import('./for-await-resolution-and-error-poisoned_FIXTURE.js');
}
```

Will load the call `load_imported_module` with the right `Referrer` (this file) and its path, but the second and third `yield` will have a `Referrer` of `Realm` which means they'll have no path associated with them and thus cannot import with a relative path.

These tests were passing before but were a false positive as the path was always resolved from the root of the module loader (which in this case is the parent of the script being run).

hansl:
We do have the path in parsed `Script` (if the `Source` has it), so use that if we can.

Fixes #3790.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -33,6 +33,18 @@
       ],
       "sourceLanguages": ["rust"],
       "preLaunchTask": "Cargo Build boa_cli"
+    },
+    {
+      "type": "lldb",
+      "request": "launch",
+      "name": "Debug Boa (Tester)",
+      "windows": {
+        "program": "${workspaceFolder}/target/debug/boa_tester.exe"
+      },
+      "program": "${workspaceFolder}/target/debug/boa_tester",
+      "args": ["run", "-s", "${input:testPath}", "-vvv"],
+      "sourceLanguages": ["rust"],
+      "preLaunchTask": "Cargo Build boa_tester"
     }
   ],
   "inputs": [
@@ -46,6 +58,11 @@
       "id": "modulePath",
       "description": "Relative path to the module root directory",
       "default": "debug",
+      "type": "promptString"
+    },
+    {
+      "id": "testPath",
+      "description": "Relative path to the test from the test262 directory",
       "type": "promptString"
     }
   ]
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -25,6 +25,16 @@
     },
     {
       "type": "process",
+      "label": "Cargo Build boa_tester",
+      "command": "cargo",
+      "args": ["build", "-p", "boa_tester"],
+      "group": "build",
+      "presentation": {
+        "clear": true
+      }
+    },
+    {
+      "type": "process",
       "label": "Run JS file",
       "command": "cargo",
       "args": ["run", "--bin", "boa", "${file}"],
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -438,6 +438,7 @@ dependencies = [
  "sys-locale",
  "tap",
  "temporal_rs",
+ "test-case",
  "textwrap",
  "thin-vec",
  "thiserror",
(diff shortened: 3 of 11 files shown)


hansl:
Note: This should fix many test262 that were broken by the last changes to `SimpleModuleLoader`, but it won't fix all of them. That is because some of those tests were false positive and the code is not behaving properly. I filed a bug regarding one of those newly failing test. I will have to investigate if there are more.

jedel1043:
  
  | Total | 50,268 | 50,268 | 0 |
  | Passed | 42,773 | 42,753 | **-20** |
  | Ignored | 1,391 | 1,391 | 0 |
  | Failed | 6,104 | 6,124 | **+20** |
  | Panics | 18 | 18 | 0 |
  | Conformance | 85.09% | 85.05% | **-0.04%** |
  
  <details><summary><b>Broken tests (20):</b></summary>
  
  ```
  test/language/expressions/dynamic-import/for-await-resolution-and-error-agen-yield.js (previously Passed)
  test/language/expressions/dynamic-import/custom-primitive.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/yield-assign-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/yield-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/call-expr-identifier.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/logical-or-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/logical-and-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/lhs-assign-operator-assign-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/cover-call-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/additive-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/member-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/ternary.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/await-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/call-expr-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/call-expr-arguments.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/array-literal.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/object-literal.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/lhs-eq-assign-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/cover-parenthesized-expr.js (previously Passed)
  test/language/expressions/dynamic-import/assignment-expression/identifier.js (previously Passed)
  ```
  </details>

jedel1043:
By the way, this doesn't print a nice comment because the action doesn't work on forks. If you want to see the test count, you can check it on the test262 action:
https://github.com/boa-dev/boa/actions/runs/8562670896/job/23466338447?pr=3791

jedel1043:
Fixed the remaining tests. This now just needs some small fixes on the tests and docs.

hansl:
Apparently the definition of absolute path is different on Windows. Will fix tomorrow.

Also confirming that (at least locally) the test262 results are the same after the latest commit on this PR as before my `SimpleModuleLoader` change:

CleanShot 2024-04-04 at 19 25 07@2x

jedel1043 (approved):
Really nice work!

.vscode/launch.json
```diff
@@ -33,6 +33,18 @@
...
+      "type": "lldb",
+      "request": "launch",
+      "name": "Debug Boa (Tester)",
+      "windows": {
+        "program": "${workspaceFolder}/target/debug/boa_tester.exe"
+      },
+      "program": "${workspaceFolder}/target/debug/boa_tester",
+      "args": ["run", "-s", "${input:testPath}", "-vvv"],
+      "sourceLanguages": ["rust"],
+      "preLaunchTask": "Cargo Build boa_tester"
```
HalidOdat, line 47: Nice addition! :)

HalidOdat (approved):
Nice work! :)

PR #3792, opened 2024-04-05
Cleanup tester deps and patterns

jedel1043:
Just some refactorings to our tester. Also removes the `regex` and `once_cell` dependencies from it.

Additionally, this also reexports our most important crates from `boa_engine`, to avoid having to manually import the other crates if someone's already using `boa_engine`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,9 +42,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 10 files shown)


HalidOdat (approved):
Nice work! :)

There is a small clippy error that needs to be fixed, besides that this looks good to me!

raskad (approved):
Looks good

PR #3795, opened 2024-04-07
Fix SimpleModuleLoader on Windows

hansl:
Basically I made a mistake with my previous PR and discovered it as I was building `embed_module!`.


jedel1043 (commented):
To avoid the same error in the future, maybe the test should be changed to test the path components instead of the whole path string.

hansl:
I don't think we would gain anything from checking the path components, but I'll make the change. There was a bug in the code and the tests were built to test the bug, so both were faulty. It happens.

jedel1043:
That makes a lot of sense. Thank you for explaining!

HalidOdat (approved):
Looks good to me! :)

jedel1043 (approved):

PR #3800, opened 2024-04-09
Implement Promise.try()

linusg:
Stage 2.7 as of yesterday. See: https://github.com/tc39/proposal-promise-try

### DIFF
--- a/core/engine/src/builtins/promise/mod.rs
+++ b/core/engine/src/builtins/promise/mod.rs
@@ -347,6 +347,7 @@ impl IntrinsicObject for Promise {
             .static_method(Self::race, js_string!("race"), 1)
             .static_method(Self::reject, js_string!("reject"), 1)
             .static_method(Self::resolve, js_string!("resolve"), 1)
+            .static_method(Self::r#try, js_string!("try"), 1)
             .static_method(Self::with_resolvers, js_string!("withResolvers"), 0)
             .static_accessor(
                 JsSymbol::species(),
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/promise/mod.rs
```diff
@@ -462,6 +463,57 @@ impl Promise {
...
+        // 6. Else,
+        else {
+            let value = status.unwrap();
+
+            // a. Perform ? Call(promiseCapability.[[Resolve]], undefined, « status.[[Value]] »).
+            promise_capability
+                .functions
+                .resolve
+                .call(&JsValue::undefined(), &[value], context)?;
+        }
```
HalidOdat, line 511: We could avoid the unwrap with a match, since we use both `Ok` and `Err` values.

```rust
match status {
   // 5. If status is an abrupt completion, then
   Err(err) => { .. }
   // 6. Else,
   Ok(value) => { .. }
}
```

HalidOdat (commented):
Thanks for the contribution! @linusg 

Just a small nitpick :)

core/engine/src/builtins/promise/mod.rs
linusg: Thanks, will fix this up later today! Pretty much my first time writing rust 😄

jedel1043 (approved):
Looks perfect! Thank you for the contribution :)

HalidOdat (approved):
Looks good to me! :)

PR #3803, opened 2024-04-10
 Add a JSON module loader which exports JSON values from files 

hansl:


jedel1043:
Just FYI, this is a stage 3 proposal for ECMAScript at the moment:
https://github.com/tc39/proposal-json-modules

I think we can merge this as it is, but since it's a security hole that someone could provide a `file.json` import that is a JS file instead, it would be preferable to add some TODO comments stating that this new module loader will adhere to that proposal in the future to ensure all the imported files are explicitly JSON.

hansl:
I use serde to parse the JSON file, without even touching the context. There's no security hole. I can add a test for errors.

jedel1043:
> I use serde to parse the JSON file, without even touching the context. There's no security hole. I can add a test for errors.

No, I don't mean the current code has a security hole. I mean that an embedder that, for example, downloads a file called `file.json` and passes it through their module loading pipeline could secretly dispatch that file to the JS module loader instead of the JSON loader, which is why the JSON proposal uses import attributes to ensure the downloaded file is provided directly to the JSON loader:
```js
import json from "./foo.json" with { type: "json" };
```

hansl:
I see. This Loader is not related to the TC39 proposal. To be honest, I haven't followed the evolution of JS in the last few years :)

I found a different way of fixing this in my project so I don't really need it. Ideally I was thinking this should be generalized anyway; have a transformer Bytes -> JsValue, then use a generic ModuleLoader that read the file, pass it through the transformer and give it to the application. That way we could do binary -> Uint8Array and things like that. That could also be extended to export more than just the content (e.g. image dimensions). Webpack has plugins that offer things like that, and it's useful in some cases.

Though that's probably outside of the scope of this PR.

hansl:
Going to close this in preference for making a `Module` instead. It shouldn't rely on the filesystem.

PR #3804, opened 2024-04-10
Add functions to create modules from a JSON value

hansl:


jedel1043:
I really think those functions should be static methods of `Module` instead, possibly following the steps of [`ParseJSONModule`](https://tc39.es/proposal-json-modules/#sec-parse-json-module) to be spec-compliant.

hansl:
I don't mind that, let me make that change.

core/interop/src/modules/json.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// # Errors
+/// This will return an error if the JSON string is invalid or cannot be converted.
+pub fn json_string_module(json: &str, context: &mut Context) -> boa_engine::JsResult<Module> {
+    let json_value = serde_json::from_str::<serde_json::Value>(json).map_err(|e| {
+        boa_engine::JsError::from_opaque(js_string!(format!("Failed to parse JSON: {}", e)).into())
+    })?;
+    let value: JsValue = JsValue::from_json(&json_value, context)?;
+    Ok(Module::from_value_as_default(value, context))
+}
```
jedel1043, line 19: I'd say just move this to the main crate, since we need to implement `Module::json` anyways.

hansl: Done. Had to rework the test but that's fine.

core/engine/src/module/mod.rs
```diff
@@ -223,6 +223,19 @@ impl Module {
...
+    ///
+    /// # Errors
+    /// This will return an error if the JSON string is invalid or cannot be converted.
+    pub fn from_json_as_default(json: &str, context: &mut Context) -> JsResult<Self> {
+        let json_value = serde_json::from_str::<serde_json::Value>(json).map_err(|e| {
+            JsError::from_opaque(js_string!(format!("Failed to parse JSON: {}", e)).into())
+        })?;
+        let value: JsValue = JsValue::from_json(&json_value, context)?;
+        Ok(Self::from_value_as_default(value, context))
+    }
```
jedel1043, line 237: This should use `builtins::Json::parse` instead of `serde_json`, to avoid coupling the engine with serde.

Also, `Module::parse_json` is fine. The specification declares that the function should be called `ParseJSONModule`, which assumes that the output will be on the default export, but we remove the suffix for simplicity.

core/engine/src/module/mod.rs
```diff
@@ -223,6 +223,19 @@ impl Module {
         )
     }
 
+    /// Create a module that exports a single JSON value as the default export, from its
+    /// JSON string. This required the `json` feature to be enabled.
+    ///
+    /// # Errors
+    /// This will return an error if the JSON string is invalid or cannot be converted.
```
jedel1043, line 230: Note: discuss whether this should be gated behind the experimental feature or not. I would say no because it's only accessible by embedders, and import attributes should be the ones gated.

core/engine/src/module/mod.rs
```diff
@@ -204,6 +205,37 @@ impl Module {
...
+                value,
+            ),
+            None,
+            None,
+            context,
+        )
+    }
+
+    /// Create a module that exports a single JSON value as the default export, from its
+    /// JSON string. This required the `json` feature to be enabled.
```
jedel1043, line 227:
```suggestion
    /// JSON string.
```

core/engine/src/module/mod.rs
hansl: Thanks I was looking for that builtins.

hansl: `parse_json` implies a return value of `Json` that is parsed. I did the change but I figure `json_module` would be better.

core/engine/src/module/mod.rs
jedel1043: Yeah, this is just a note for future reference.

core/engine/src/module/mod.rs
jedel1043: Not really? JSON is just the name of the syntax used to generate a `JsObject`, it doesn't really make sense to return a "Json" value; in serde-json, for example, the `serde_json::from_str` doesn't return a `Json` type, but a `Value` type.

jedel1043 (approved):
Looks perfect to me!

nekevss (approved):

PR #3810, opened 2024-04-14
Add GIFs to CI via vhs-action

nekevss:
This Pull Request is to propose adding a GIF to `boa_cli` and the `vhs-action` CI to update the GIF in CI.

With `vhs-action`, we could also look into automating more GIFs for our docs that can be built out during the artifacts CI.

The one unfortunate thing is that the GIF generation for `boa_cli` won't take in any changes from `boa_engine` as it is set to the currently released version in it's `Cargo.toml`.

I am also a bit concerned about `stefanzweifel/git-auto-commit-action@v4` as I'm not entirely sure how it will affect the commit history.

EDIT: VHS can be found [here](https://github.com/charmbracelet/vhs) for reference


.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          commit_message: Update generated VHS GIF
+          branch: main
+          commit_user_name: vhs-action 📼
+          commit_user_email: actions@github.com
+          commit_author: vhs-action 📼 <actions@github.com>
+          file_pattern: '*.gif'
```
jedel1043, line 31: Was thinking more about this, and maybe this shouldn't be a direct commit, but a check that compares the current version of the GIF with the GIF generated by the tool, and throws if both differ.

nekevss:
Closing this PR in favor of another one to clean up the git history

PR #3818, opened 2024-04-17
Create a new Specifier type for JavaScript module specifiers

hansl:
This PR does not change any of the current API. I wanted to see if people preferred doing it as part of this PR (thus larger in scope), or doing a review of this new type, then as a follow up update the related APIs.

My thoughts after playing with modules;

- There is friction between what is a Path, and what is a module specifier. To remove that friction and simplify the API surface, I suggest that all Path from Source, Module and their relative types be moved to using this new Specifier/OwnedSpecifier type.
- Paths should only be used when working with the file system. That is, a Path is only necessary in a ModuleLoader's implementation, and should never be part of the interface. That way the API is platform independent, and concerns are separated between resolution/namespacing and actually pointing at files. Same goes for URLs (Path being technically a URL). These are implementation specific and should not be part of the API. Having easy conversions from one to the other would make it trivial anyway.
- Source should have a Specifier, Scripts should take the specifier and make a copy of it, etc.
- Traces will show up just fine in a single namespace, but we should avoid storing relative specifiers (relative to what). We should see absolute paths for modules, regardless of how they were imported, so `import './path'` resolving to `/some/path.js` (see below), should show up in traces as `/some/path.js`. This would be part of the follow up PR.
- There should be an invariant that two absolute specifiers should ALWAYS resolve to the same module. AFAICT, that is true of all JavaScript module implementation (can't remember if it's in the spec). Relative specifiers should be resolved to become absolute, then follow the same invariant.
- I would suggest having a `RevoledSpecifier` or something which is guaranteed at creation to be absolute and resolved. It can `Deref<Specifier>` or `Deref<OwnedSpecifier>`. That way the type system will make sure you cannot create a module with a relative path (relative to what?!?). When creating a module, you need to pass a referrer or a resolved specifier.

Some extra curricular suggestions:

- There should be some interaction between a resolver and a loader, but these two concerns should likely be separated. I'm not going to do that, but having something like a `NodeModuleResolver` that doesn't have to also be a loader would be good.
- Loaders likely should have some awareness of what specifiers are available, and allow queries and (ideally) listing where possible. Query requires that the trait default implementation of `ModuleLoader::get_module`, and that this method be implemented correctly.
- I didn't want to get into URLs just yet (or at all). This is not a use case I care about right now. But this new type would also help there, as resolving relative to an existing specifier is made trivial.

### DIFF
--- a/core/engine/src/module/specifier.rs
+++ b/core/engine/src/module/specifier.rs
@@ -1,0 +1,703 @@
+//! A [`Specifier`] is a typed string that correspond to a module's name in the
+//! module loader. It is used to identify a module in the module loader.
+//!
+//! Specifiers are any valid JavaScript string literal. In Rust, we keep them
+//! as UTF-8 strings for performance reasons, and translate them to JavaScript
+//! strings when needed.
+//!
+//! All modules live inside a single specifier namespace. All specifiers are
+//! unique to the module they represent. This means that two different modules
+//! cannot have the same specifier.
+//!
+//! Specifiers are used to identify modules in the module loader. They are
+//! passed to the module loader to load a module, and are used to store modules
+//! in the module loader. When creating a Source, the specifier should
+//! optionally be passed as the name of the module.
+//!
+//! Specifiers are Unix-like; they use forward slashes as separators, and can
+//! be relative if they start with a `.` or `..`.
+
+use std::fmt::Display;
+use std::path::PathBuf;
+use std::{fmt, path};
+
+use crate::value::TryFromJs;
+use crate::{Context, JsError, JsNativeError, JsResult, JsString, JsValue};
+
+#[cfg(test)]
+mod tests;
+
+/// Specifier component separator.
+pub const SEPARATOR: char = '/';
+
+/// Specifier component separator as a string.
+pub const SEPARATOR_STR: &str = "/";
+
+/// Returns true if the byte is a specifier separator.
+fn is_seperator_byte(b: u8) -> bool {
+    b == SEPARATOR as u8
+}
+
+/// Internal enum for iterator state.
+#[derive(Default, Debug, Clone, Copy, PartialEq, PartialOrd)]
+enum State {
+    /// Initial value.
+    #[default]
+    Start,
+
+    /// Normal component.
+    Body,
+
+    /// Done iterating.
+    Done,
+}
+
+/// An iterator over the [`Component`]s of a [`Specifier`].
+#[derive(Debug, Clone)]
+pub struct Components<'a> {
+    // This inner has been verified to be a valid `str` at some point.
+    inner: &'a [u8],
+
+    // The state of the front iterator.
+    front: State,
+
+    // The state of the back iterator.
+    back: State,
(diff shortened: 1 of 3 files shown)


core/engine/src/module/specifier.rs
jedel1043, line 1: Isn't this just `typed_path::UnixPath/UnixPathBuf`?

jedel1043:
I'm trying to think about the implications of always having absolute paths for modules and scripts. Right now, I can think of:
- Users who don't want to use absolute paths on traces for security reasons.
- Users who want to embed Boa in a system that doesn't have a filesystem (not possible now, WIP),
  but that want to expose some relative paths to the host project for debugging purposes.
Committing to absolute paths would be very bad for both use cases.

Maybe we can just change specifiers to store a `str`, then the loaders would be the ones to decide if they want to parse unix paths/windows paths/urls?

core/engine/src/module/specifier.rs
hansl: Kind of, but different. I would have had to build pretty much the same code (maybe a bit less, so like ~100 lines) to wrap `UnixPath/Buf`.

I read their code while implementing this, as well as the std `Path`.

hansl:
> Users who don't want to use absolute paths on traces for security reasons.

Absolute to the root of the (very much virtual) namespace, not their filesystems. `Path` and FileSystem is an implementation detail. The way to go from `Specifier` to `Path` is by having a base path, like we do in `SimpleModuleLoader`. Everywhere else in Boa itself we would never deal with those base paths (which is the part that's sensitive).

> Users who want to embed Boa in a system that doesn't have a filesystem (not possible now, WIP), but that want to expose some relative paths to the host project for debugging purposes.

This would be done using the `ModuleLoader`. The base would be the host project, joining specifiers with the base path when loading/resolving modules. The root of the namespace becomes the base of the host project. But that's all up to the host itself, Boa has no knowledge of these kind of details. Boa only cares about specifiers and modules.

Basically what I'm saying is:

> `Boa` SHOULD NOT have as a concern WHERE the file is located or HOW to find it, but that there exists a map of unique `Specifier` => `Module` that can be resolved using a `ModuleLoader`, and that each unique absolute `Specifier` resolves to the same `Module`.

This is a very powerful statement, and would allow optimizations that are host independent, such as having a module cache outside of module loader, for example.

And to avoid further confusion, what I mean by "absolute `Specifier`" is not absolute in terms of filesystem (Boa should not be aware of filesystems or platform), but in an imaginary namespace that contains all modules handled by the host. Specifier can be seen as URIs in this context, and every module/source would have a Specifier that must be complete (not relative), and every such specifier must point to only one module.

hansl:
`Source::from_filepath` would likely take a `Specifier` as an additional argument. There is nothing wrong with utility functions. We could also make it separate (a `SourceExt` trait for example) to keep engine clean, but I don't think that's necessary.

jedel1043:
This discussion is getting a bit too long for a PR, and I wanted to ask some more questions about the API. I'll open a thread in Matrix.

PR #3820, opened 2024-04-18
Fix order of `ToString` call in `Function` constructor

HalidOdat:
The `ToString` method was being called on the body before the paramater list, this caused some test262 tests to fail.


nekevss (approved):

jedel1043 (approved):
Good catch!

PR #3822, opened 2024-04-21
Implement TryFromJs for Either<L, R>

hansl:


core/engine/Cargo.toml
```diff
@@ -104,6 +104,7 @@ intrusive-collections = "0.9.6"
 cfg-if = "1.0.0"
 time.workspace = true
 hashbrown.workspace = true
+either = { version = "1", optional = true }
```
jedel1043, line 107: Can you add the explicit feature? IIRC, implicit features for optional dependencies are [going away](https://github.com/rust-lang/cargo/issues/12826), so it would be good to start defaulting to explicit features.

jedel1043 (changes requested):
Looks good. Thank you!

core/engine/Cargo.toml
hansl: Done.

jedel1043 (approved):

HalidOdat (approved):
Looks good to me! :)

PR #3829, opened 2024-04-25
Decouple `Context` from `ByteCompiler`

HalidOdat:
This PR removes the `&mut Context` field from the `ByteCompiler` making it possible to put the bytecompiler in its own crate (there is still work to move `JsString` and other parts to their own crate so they can be used it both crates, but this is a step in the right direction).

~~This also fixes the some failing `eval` test262 tests.~~
EDIT: Oops, Tested with an older version of test262 result, so no bonus passing tests :sweat_smile: 

TODO:

- [x] Clean-up and documentation of spec deviation


core/engine/src/bytecompiler/declarations.rs
```diff
@@ -24,6 +25,361 @@ use boa_ast::operations::annex_b_function_declarations_names;
...
+///
+/// This diverges from the specification by separating the context from the compilation process.
+/// Many steps are skiped that are done during bytecode compilation.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-globaldeclarationinstantiation
+#[cfg(feature = "annex-b")]
+pub(crate) fn global_declaration_instantiation_context(
```
nekevss, line 59: Thought: Potentially move these out of `bytecompiler` to `script`

Now that these are separated, is there any need for these functions to still live in this module (especially if the bytecompiler is eventually moved to it's own crate).

nekevss (commented):
A couple things I missed.

Sort of general question: did you happen to benchmark this? Was sort of curious what impact the new `global` opcodes have.

Also, would it make sense for the global opcodes to occur in the instantiation functions like in the spec vs. opcodes?

HalidOdat:
> Sort of general question: did you happen to benchmark this? Was sort of curious what impact the new `global` opcodes have.

The opcodes execute only once in the execution of a script, when I checked for rust benchmarks as well as quickjs benchmarks I didn't notice any regressions.

> Also, would it make sense for the global opcodes to occur in the instantiation functions like in the spec vs. opcodes?

The problem with this is that some functions need to already be compiled and instantiated to be put into the global object, that's why the it's delayed when we execute the bytecode so we can separate them. It might be possible by having some post `_context` functions calls as well as pre calls (as we do now).

The main benefit to having opcodes for them (besides the complexity reduction) is reproducibility of `CodeBlock` execution, this will be needed when we implement `CodeBlock` serialization to disk and back, we wont need to store the AST (there is still work to be done for this, ideally all of the `_context` function should be converted to bytecode, but this is a separate issue, not related to splitting the bytecompiler).

core/engine/src/bytecompiler/declarations.rs
HalidOdat: Yup, we can put them in separate locations, the only reason I left them there was because the context and non-context variants are close to each other.

core/engine/src/bytecompiler/declarations.rs
```diff
@@ -678,17 +869,16 @@ impl ByteCompiler<'_> {
                 if !declared_function_names.contains(&name) {
                     // a. If varEnv is a Global Environment Record, then
                     if var_env.is_global() {
-                        let name = name.to_js_string(self.interner());
+                        let index = self.get_or_insert_name(name);
+                        // let name = name.to_js_string(self.interner());
```
raskad, line 873:
```suggestion
```

raskad (approved):
Nice work. :)

nekevss (approved):
Given the above, I'm fine with merging this. Nice work on this! :smile:

PR #3831, opened 2024-04-27
Separate `JsString` into its own crate

HalidOdat:
Depends on #3450 

This Pull Request is a continuation of the work on #3829 , by moving the string crate because the `ByteCompiler` depends on it.

~~There are still some issues with `js_str!` macro that constructs a `JsStr`, there is no `$crate` for proc macros unfortunately :cry:~~ Fixed by making it point to `boa_engine` otherwise you have to include `boa_string` everywhere... If there is another solution for this, would love to hear it :)


jedel1043:
Most child PRs rebase to main when its parent PR is merged, but Github decided to close this somehow ¯\\_(ツ)_/¯
(probably was out of sync)

HalidOdat:
Yeah, it's weird... It wouldn't allow me to change base manually, so I recreated the PR https://github.com/boa-dev/boa/pull/3837

PR #3837, opened 2024-05-02
Separate `JsString` into its own crate

HalidOdat:
This Pull Request is a continuation of the work on #3829 , by moving the string crate because the `ByteCompiler` depends on it.

~~There are still some issues with `js_str!` macro that constructs a `JsStr`, there is no `$crate` for proc macros unfortunately :cry:~~ Fixed by making it point to `boa_engine` otherwise you have to include `boa_string` everywhere... If there is another solution for this, would love to hear it :)


HalidOdat:
Clippy warnings fixed in #3838

core/engine/src/string.rs
```diff
@@ -0,0 +1,256 @@
...
+//   comment or doc of the unsafe block or function, respectively.
+#![deny(
+    unsafe_op_in_unsafe_fn,
+    clippy::undocumented_unsafe_blocks,
+    clippy::missing_safety_doc
+)]
+// Remove when/if https://github.com/rust-lang/rust/issues/95228 stabilizes.
+// Right now this allows us to use the stable polyfill from the `sptr` crate, which uses
+// the same names from the unstable functions of the `std::ptr` module.
+#![allow(unstable_name_collisions)]
```
raskad, line 22: These docs and rules can probably be altered since the only thing in this file is the `js_string` macro right?

core/engine/src/string.rs
```diff
@@ -0,0 +1,256 @@
...
+    //     let y = x.clone();
+
+    //     assert!(x.ptr.is_tagged());
+
+    //     assert_eq!(x.ptr.addr(), y.ptr.addr());
+
+    //     let z = js_string!();
+    //     assert_eq!(x.ptr.addr(), z.ptr.addr());
+    //     assert_eq!(y.ptr.addr(), z.ptr.addr());
+    // }
```
raskad, line 173: Leftover commented out tests.

raskad (approved):
Nice work :)

core/string/src/tagged.rs
nekevss, line 1: thought: should we look to move `Tagged<T>` into a utility crate.

Obviously this is super minor since it's really only like 100 lines, but instead of having this reimplemented in `boa_string`, should we look to move tag into a utility crate and then import it into `boa_engine` and `boa_string`?

nekevss (approved):
LGTM! Just had one passing thought when reviewing, but nothing blocking for me.

core/string/src/tagged.rs
HalidOdat: Yeah, I was going to do that, but wanted to keep the PR small. Will do that in a follow up PR :)

HalidOdat: Here is the PR for it: https://github.com/boa-dev/boa/pull/3849 :)

PR #3843, opened 2024-05-09
Adding TryFromJs implementations for tuples

hansl:
This make it easier to get the first N values from an array when de-serializing.


hansl:
No idea why the Windows action is failing.

nekevss:
If you rebase the CI should be fixed :smile: There was an issue that was fixed with #3845.

core/engine/src/value/conversions/try_from_js/tuples.rs
```diff
@@ -0,0 +1,53 @@
+//! Implementation of [`TryFromJs`] for tuples.
+//!
+//! Tuples are converted from a JavaScript array, using similar semantics to `TypeScript` tuples:
+//!     - If the tuple is shorter than the array, the extra elements are ignored.
+//!     - If the tuple is longer than the array, the extra elements are `undefined`.
+//!     - If the array is empty, all elements are `undefined`.
+//!
+//! A tuple of size 0 (unit type) is represented as any value except `null` or `undefined`.
```
jedel1043, line 8: I'm a bit conflicted about this.

On the one hand, I think it makes sense to use any non-nullable value as the unit type, because `undefined` in JS is commonly used to represent the absence of a value, which `()` doesn't represent on the Rust world.
On the other hand, it would be a lot less surprising for new users to have it be `undefined`, since that's the default value for functions that don't return, which would be equivalent to how Rust functions return `()` when the function doesn't explicitly return.

I think I'm tending towards the second option, but I would be happy to know your thoughts on this.

hansl: I'm okay either way. The first implementation did not include the empty tuple. If we have a `JsNotNull` type that would be easier, there's currently no way to verify, e.g., that a field in an object is present, without caring about its value.

jedel1043: Yeah, let's just remove the implementation for `()` for now since the rest of the implementations are really useful, and we can discuss what to do with `()` in the future.

hansl: Removed.

jedel1043 (approved):
Thank you!

jedel1043 (approved):

raskad (approved):

PR #3845, opened 2024-05-10
Switch from actions-rs/toolchain to dtolnay/rust-toolchain

raskad:
Fix the current ci errors

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -15,11 +15,9 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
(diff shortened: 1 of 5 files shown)


nekevss (approved):
Sounds good to me. `actions-rs` is archived anyways too, right? So this is probably a good switch regardless of the current CI issues.

raskad:
> Sounds good to me. `actions-rs` is archived anyways too, right? So this is probably a good switch regardless of the current CI issues.

Yes. I will switch the other remaining action-rs actions in a follow up.

jedel1043 (approved):

PR #3849, opened 2024-05-12
Split `Tagged<T>` into a utility crate

### LINKED ISSUE #4183: Extract `Tagged` into an utility crate
Implementation:
https://github.com/boa-dev/boa/blob/main/core/engine/src/tagged.rs

HalidOdat:
Depends on #3837 
Closes #4183 
This makes `Tagged<T>` be usable in `boa_string` and `boa_engine`.


nekevss:
General discussion on utility crates: how do we want to go about handling utilities in general that's going to be best for maintenance in the long term?

Should we have one general `boa_utils` crate or a `utils` directory with individual crates hosted inside it (I'm thinking of `icu4x`'s structure as an example). We could also extract `SmallMap` into the `utils` as well. Both `Tagged` and `SmallMap` could then be standalone crates unto themselves. 

There's some benefits to taking the `utils` directory approach.

1. The code is more modular.
2. Offers the libraries for generic use case for the Rust ecosystem as a whole.
3. Could provide some soft "marketing" for boa, so to speak.

There could be a pretty big negative around the maintenance complexity utility crates may cause and managing separate releases may actually end being a large headache. :confused:

Any thoughts?

HalidOdat:
Hmmm... `Tagged<T>` and `SmallMap` are very small (single file) so don't think it's going to need that much maintenance. Just a thought: we could put them in separate repos (like we do with `ryu-js`, and `temporal_rs`) may make it easier to separately publish the crates.

What does the rest of the team say? @jedel1043 @raskad @Razican

Razican:
From my side, in general, if the crate / code makes sense outside Boa, then we should make it separate, so that others can benefit from it. If not, I would keep it in.

For this case, it seems these types could be used outside of Boa, and they have no dependency on the engine, so I would make them separate if it's not too much maintenance burden (and we don't forget to maintain them 😅)

jedel1043:
I think separating small utility crates to separate repos is arguably worse for us. I would be in favour of having either a `boa_utils` or an `utils` directory here, but having to maintain a whole repo for only a single `Tagged` data structure sounds counter-productive.

jedel1043:
> From my side, in general, if the crate / code makes sense outside Boa, then we should make it separate, so that others can benefit from it. If not, I would keep it in.

We can do that and also maintain it as a separate crate on our main repo. That's what the ICU4X folks do with `writeable`, `yoke`, etc. and it's not too bad. I think it only makes sense to extract separate repos if there's a big incentive to do so.

nekevss:
As mentioned above, I'd be for the `utils` directory with any small utility crates in them.

> I'm not sure what we should name the crate because `tagged`, `tagged_ptr`, `tagged_pointer` are all taken... Any ideas?

`tagged_rs` maybe? `tag_ptr` could work although not super accurate 😕.

HalidOdat:
@boa-dev/maintainers  This is done, just needs review/merge :)

Cargo.toml
```diff
@@ -44,6 +44,7 @@ boa_parser = { version = "~0.20.0", path = "core/parser" }
 boa_profiler = { version = "~0.20.0", path = "core/profiler" }
 boa_runtime = { version = "~0.20.0", path = "core/runtime" }
 boa_string = { version = "~0.20.0", path = "core/string" }
+tag_ptr = { version = "~0.20.0", path = "utils/tagged" }
```
jedel1043, line 47: We should probably version utility crates separately from the "main" crates.

utils/tagged/Cargo.toml
jedel1043, line 1: Should we use the same name of the crate as the directory name?

utils/tagged/src/lib.rs
```diff
@@ -40,7 +42,9 @@ impl<T> Tagged<T> {
     ///
     /// - `T` must have an alignment of at least 2.
     /// - `tag` must fit inside `usize::BITS - 1` bits
-    pub(crate) const fn from_tag(tag: usize) -> Self {
+    #[inline]
```
jedel1043, line 45: Had an idea. What if we moved the assertion to compile time? Should remove the alignment requirement from the docs, and that way you cannot create invalid tagged pointers:
```rust
    const ALIGNMENT_CHECK: () = assert!(align_of::<T>() >= 2);

    /// Creates a new, tagged `Tagged` pointer from an integer.
    ///
    /// # Requirements
    ///
    /// - `tag` must fit inside `usize::BITS - 1` bits
    #[inline]
    #[must_use]
    pub const fn from_tag(tag: usize) -> Self {
        let _ = Self::ALIGNMENT_CHECK;
        let addr = (tag << 1) | 1;
        // SAFETY: `addr` is never zero, since we always set its LSB to 1
        unsafe { Self(NonNull::new_unchecked(ptr::without_provenance_mut(addr))) }
    }
```

utils/tagged/Cargo.toml
HalidOdat: Opps, forgot to update, nice catch!

utils/tagged/src/lib.rs
HalidOdat: Nice suggestion! :)

utils/tag_ptr/src/lib.rs
```diff
@@ -56,7 +64,8 @@ impl<T> Tagged<T> {
     /// # Safety
     ///
     /// - `T` must be non null.
-    pub(crate) const unsafe fn from_ptr(ptr: *mut T) -> Self {
+    #[inline]
+    pub const unsafe fn from_ptr(ptr: *mut T) -> Self {
         debug_assert!(align_of::<T>() >= 2);
```
jedel1043, line 69: You should also replace the rest of the debug_assert with the compile time check.

Cargo.toml
```diff
@@ -45,6 +45,9 @@ boa_profiler = { version = "~0.20.0", path = "core/profiler" }
 boa_runtime = { version = "~0.20.0", path = "core/runtime" }
 boa_string = { version = "~0.20.0", path = "core/string" }
 
+# Utility Repo Crates
+tag_ptr = { path = "utils/tag_ptr" }
```
jedel1043, line 49: Can you test if this doesn't require the version?
```
cargo ws publish --dry-run
```
I remember that the dependency still required the version when publishing to crates.io, but maybe things changed.

HalidOdat:
```bash
cargo ws publish --dry-run
```

Seems to work :)

```txt
info checking tag_ptr
   Compiling tag_ptr v0.1.0 (/home/arch/projects/boa/utils/tag_ptr)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
    Updating crates.io index
   Packaging tag_ptr v0.1.0 (/home/arch/projects/boa/utils/tag_ptr)
    Packaged 7 files, 14.6KiB (5.6KiB compressed)
   Uploading tag_ptr v0.1.0 (/home/arch/projects/boa/utils/tag_ptr)
warning: aborting upload due to dry run
```

jedel1043:
Removed the assertions because I realized the pointer could be unaligned even if the type must be aligned (`std::mem::read_unaligned`). That by itself cannot cause undefined behaviour though; at works some memory leaks, but unaligned pointer will just be treated as integers.

jedel1043 (approved):

PR #3856, opened 2024-05-15
add some temporal methods

jasonwilliams:
Adds methods to yearMonth and monthDay

Relies on https://github.com/boa-dev/temporal/pull/44

### DIFF
--- a/core/engine/src/builtins/temporal/duration/mod.rs
+++ b/core/engine/src/builtins/temporal/duration/mod.rs
@@ -806,7 +806,7 @@ impl Duration {
 
 // -- Duration Abstract Operations --
 
-/// 7.5.8 `ToTemporalDuration ( item )`
+/// 7.5.12 `ToTemporalDuration ( item )`
 pub(crate) fn to_temporal_duration(
     item: &JsValue,
     context: &mut Context,
--- a/core/engine/src/builtins/temporal/duration/mod.rs
+++ b/core/engine/src/builtins/temporal/duration/mod.rs
@@ -826,7 +826,7 @@ pub(crate) fn to_temporal_duration(
     Ok(result)
 }
 
-/// 7.5.9 `ToTemporalDurationRecord ( temporalDurationLike )`
+/// 7.5.13 `ToTemporalDurationRecord ( temporalDurationLike )`
 pub(crate) fn to_temporal_duration_record(
     temporal_duration_like: &JsValue,
     context: &mut Context,
@@ -858,6 +858,7 @@ pub(crate) fn to_temporal_duration_record(
     let partial = to_temporal_partial_duration(temporal_duration_like, context)?;
 
     // 5. If partial.[[Years]] is not undefined, set result.[[Years]] to partial.[[Years]].
+
     // 6. If partial.[[Months]] is not undefined, set result.[[Months]] to partial.[[Months]].
     // 7. If partial.[[Weeks]] is not undefined, set result.[[Weeks]] to partial.[[Weeks]].
     // 8. If partial.[[Days]] is not undefined, set result.[[Days]] to partial.[[Days]].
@@ -870,7 +871,7 @@ pub(crate) fn to_temporal_duration_record(
     // 15. If ! IsValidDuration(result.[[Years]], result.[[Months]], result.[[Weeks]], result.[[Days]], result.[[Hours]], result.[[Minutes]], result.[[Seconds]], result.[[Milliseconds]], result.[[Microseconds]], result.[[Nanoseconds]]) is false, then
     // a. Throw a RangeError exception.
     // 16. Return result.
-    InnerDuration::from_partial(&partial).map_err(Into::into)
+    InnerDuration::from_partial_duration(partial).map_err(Into::into)
 }
 
 /// 7.5.14 `CreateTemporalDuration ( years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds [ , newTarget ] )`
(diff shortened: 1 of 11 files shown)


nekevss (commented):
Just a little early feedback :smile:

jasonwilliams:
linked to https://github.com/boa-dev/temporal/pull/44

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -263,10 +388,40 @@ impl PlainYearMonth {
...
+                    DurationOperation::Add,
+                    &year_month,
+                    duration,
+                    context,
+                    overflow,
+                )
+                .expect("Error adding duration to year month");
+            create_temporal_year_month(year_month_result, None, context)
+        } else {
+            return Err(JsNativeError::typ()
```
nekevss, line 421: There is a `FromStr` implementation for Duration. Was the implementation not working in this scenario?

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -202,55 +210,172 @@ impl BuiltInConstructor for PlainYearMonth {
...
 impl PlainYearMonth {
-    fn get_calendar_id(_this: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
-        Err(JsNativeError::error()
-            .with_message("not yet implemented.")
-            .into())
+    // 9.2.2 `Temporal.PlainYearMonth.from ( item [ , options ] )`
+    fn from(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        let item = args.get_or_undefined(0);
+        // 1. If Type(item) is Object or Type(item) is String and item is not null, then
+        let inner = if item.is_object() {
```
nekevss, line 220: question: Is this up to date?

I currently have the specification showing the below for 9.2.2, which may actually simplify the logic here if so.

`2. If item is an Object and item has an [[InitializedTemporalYearMonth]] internal slot, then`

jasonwilliams: What's the best way to brand check these?

jedel1043: Downcasting is not really performance heavy, it's just two u64 comparisons. In this case, since you just need the inner data, you can just do something like:
```rust
if let Some(data) = item.as_object().and_then(JsObject::downcast_ref::<PlainYearMonth>) {
    // Perform ? [GetTemporalOverflowOption](https://tc39.es/proposal-temporal/#sec-temporal-gettemporaloverflowoption)(options).
    // b. Return ! [CreateTemporalYearMonth](https://tc39.es/proposal-temporal/#sec-temporal-createtemporalyearmonth)(item.[[ISOYear]], item.[[ISOMonth]], item.[[Calendar]], item.[[ISODay]]).
}
```
(not typechecked or borrow checked, maybe you'll have to clone if the call to `get_temporal_overflow_option` causes a panic)

core/engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -929,94 +928,94 @@ pub(crate) fn to_temporal_partial_duration(
...
+            context,
+        )?));
     }
 
     // 12. Let minutes be ? Get(temporalDurationLike, "minutes").
     let minutes = unknown_object.get(js_str!("minutes"), context)?;
     // 13. If minutes is not undefined, set result.[[Minutes]] to ? ToIntegerIfIntegral(minutes).
     if !minutes.is_undefined() {
-        result.set_minutes(f64::from(to_integer_if_integral(&minutes, context)?));
+        result.minutes = Some(FiniteF64::from(to_integer_if_integral(&minutes, context)?));
```
nekevss, line 972: thought: It might be nice if we had a `JsValue::map` that acted similar to `Option::map`, where Undefined is None.

core/engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -929,94 +928,94 @@ pub(crate) fn to_temporal_partial_duration(
...
-    let mut result = InnerDuration::partial();
+    let mut result = PartialDuration::default();
 
     // 3. NOTE: The following steps read properties and perform independent validation in alphabetical order.
     // 4. Let days be ? Get(temporalDurationLike, "days").
     let days = unknown_object.get(js_str!("days"), context)?;
     if !days.is_undefined() {
         // 5. If days is not undefined, set result.[[Days]] to ? ToIntegerIfIntegral(days).
-        result.set_days(f64::from(to_integer_if_integral(&days, context)?));
+        result.days = Some(FiniteF64::from(to_integer_if_integral(&days, context)?));
```
nekevss, line 938: suggestion: potentially use in-place `Option` operations.

There may be some benefit to use either `insert` or `get_or_insert` over " = Some(...) ". i.e. result.days.insert(FiniteF64::from(...));

core/engine/src/builtins/temporal/plain_month_day/mod.rs
```diff
@@ -33,6 +43,123 @@ impl PlainMonthDay {
     }
 }
 
+// ==== `Temporal.PlainMonthDay` static Methods ====
+impl PlainMonthDay {
+    // 10.2.2 Temporal.PlainMonthDay.from ( item [ , options ] )
+    fn from(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        let options = get_options_object(args.get_or_undefined(1))?;
+        let item = args.get_or_undefined(0);
+        let inner = if item.is_object() {
```
nekevss, line 52: suggestion: extract into `to_temporal_month_day` function

Technically, `ToTemporalMonthDay` is only called twice by by `from` and `equals`, so I could see the argument to not extract it. But having a to-X method implementation would be consistent with all the other built-ins and hopefully make everything a bit more readable.

core/engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -458,101 +459,79 @@ impl Duration {
         // a. Let years be temporalDurationLike.[[Years]].
         // 5. Else,
         // a. Let years be duration.[[Years]].
-        let years = if temporal_duration_like.years().is_nan() {
-            duration.inner.years()
-        } else {
-            temporal_duration_like.years()
-        };
+        let years = temporal_duration_like
```
nekevss, line 462: thought: I definitely much prefer this to the previous code!

nekevss (commented):
`PartialDuration` update looks good to me! Had a couple more comments around MonthDay. :smile:

nekevss:
Was double checking this. I think the new revision of `temporal_rs` needs to be added, and then I think this PR should be just about good to go.

core/engine/src/builtins/temporal/duration/mod.rs
jasonwilliams: Yup done

core/engine/src/builtins/temporal/duration/mod.rs
jasonwilliams: Im still wondering how to do this in a clean way, was you thinking `result.minutes = minutes.map(|minutes| FiniteF64::from(to_integer_if_integral(&minutes, context)?)` ?

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -338,3 +465,84 @@ pub(crate) fn create_temporal_year_month(
...
+        .and_then(JsObject::downcast_ref::<PlainYearMonth>)
+        .ok_or_else(|| {
+            JsNativeError::typ().with_message("this value must be a PlainYearMonth object.")
+        })?;
+
+    let inner = &year_month.inner;
+    let year_month_result = if is_addition {
+        inner
+            .add_duration(&duration, overflow)
+            .expect("Error adding duration to year month")
```
nekevss, line 508: issue: The expect here and on line 512 would cause a panic instead of returning any error that happened during addition and subtraction. They should be removed and the error returned.

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -202,55 +208,153 @@ impl BuiltInConstructor for PlainYearMonth {
...
+                        .expect("Month not found")
+                        .to_i32(context)
+                        .expect("Cannot convert month to i32"),
+                    item.get_v(js_str!("day"), context)
+                        .map_or(Some(1), |x| x.to_i32(context).ok()),
+                    calendar,
+                    overflow,
+                )?
+            }
+        } else if item.is_string() {
```
nekevss, line 251:
```suggestion
        } else if let Some(item_as_string) = item.as_string() {
```

Is something like this possible that way an expect could be avoided.

core/engine/src/builtins/temporal/plain_month_day/mod.rs
```diff
@@ -128,3 +278,50 @@ pub(crate) fn create_temporal_month_day(
...
+
+    let inner = if item.is_object() {
+        if let Some(data) = item
+            .as_object()
+            .and_then(JsObject::downcast_ref::<PlainMonthDay>)
+        {
+            data.inner.clone()
+        } else {
+            InnerMonthDay::new(
+                item.get_v(js_str!("month"), context)
```
nekevss, line 302: `to_integer_with_truncation` for `temporal/mod.rs` should be usable when getting the items value that way another `expect` can be removed.

For instance, something akin to the `PlainDate` [constructor](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/temporal/plain_date/mod.rs#L254)

core/engine/src/builtins/temporal/plain_month_day/mod.rs
```diff
@@ -128,3 +278,50 @@ pub(crate) fn create_temporal_month_day(
...
+    context: &mut Context,
+) -> JsResult<JsValue> {
+    let overflow = get_option::<ArithmeticOverflow>(options, js_str!("overflow"), context)?
+        .unwrap_or(ArithmeticOverflow::Constrain);
+
+    // get the calendar property (string) from the item object
+    let calender_id = item.get_v(js_str!("calendar"), context)?;
+    let calendar = to_temporal_calendar_slot_value(&calender_id)?;
+
+    let inner = if item.is_object() {
```
nekevss, line 294:
```suggestion
    let inner = if let Some(obj) = item.as_object() {
```

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -338,3 +465,84 @@ pub(crate) fn create_temporal_year_month(
...
+fn add_or_subtract_duration(
+    is_addition: bool,
+    this: &JsValue,
+    duration_like: &JsValue,
+    options: &JsObject,
+    context: &mut Context,
+) -> JsResult<JsValue> {
+    let duration: Duration = if duration_like.is_object() {
+        to_temporal_duration(duration_like, context)?
+    } else if duration_like.is_string() {
```
nekevss, line 479: Same as mentioned elsewhere. `if let Some(s) = item.as_string()` would ideally be used here. Doing so will remove some panics.

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -338,3 +465,84 @@ pub(crate) fn create_temporal_year_month(
...
+        Duration::from_str(
+            duration_like
+                .as_string()
+                .expect("Value passed not a string")
+                .to_std_string_escaped()
+                .as_str(),
+        )
+        .expect("Unable to parse Duration from string")
```
nekevss, line 487: issue: The errors from parsing a duration should be returned over panicking.

core/engine/src/builtins/temporal/plain_month_day/mod.rs
```diff
@@ -128,3 +278,50 @@ pub(crate) fn create_temporal_month_day(
...
+                    .expect("Cannot convert month to i32"),
+                item.get_v(js_str!("day"), context)
+                    .expect("Day not found")
+                    .to_i32(context)
+                    .expect("Cannot convert day to i32"),
+                calendar,
+                overflow,
+            )?
+        }
+    } else if item.is_string() {
```
nekevss, line 314:
```suggestion
    } else if let Some(item_string) = item.as_string() {
```

suggestion: use `as_string()` over `is_string()` to remove the `expect`

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -202,55 +208,153 @@ impl BuiltInConstructor for PlainYearMonth {
...
+                data.inner.clone()
+            } else {
+                let options = get_options_object(args.get_or_undefined(1))?;
+                let overflow = get_option(&options, js_str!("overflow"), context)?
+                    .unwrap_or(ArithmeticOverflow::Constrain);
+
+                // a. Let calendar be ? ToTemporalCalendar(item).
+                let calendar = to_temporal_calendar_slot_value(args.get_or_undefined(1))?;
+                InnerYearMonth::new(
+                    item.get_v(js_str!("year"), context)
```
nekevss, line 237:
```suggestion
                    super::to_integer_with_truncation(item.get_v(js_str!("year"), context)?, context)?,
```

suggestion: Instead of the having the expects here, `to_integer_with_truncation` can be used.

nekevss (changes requested):
This is looking good! I had a few suggestions / noticed some places where some `expects` can be removed to potentially get rid of the panics that popped up on the conformance testing.

nekevss (approved):
LGTM! Nice work! 😄

jedel1043 (approved):
Looks great!

PR #3863, opened 2024-05-20
Group dependabot updates

jedel1043:
This makes it easier to maintain dependency updates.

Only groups minor and patch updates; major version changes should definitely be reviewed in a standalone way.


HalidOdat (approved):
Looks great to me! :)

raskad (approved):

PR #3868, opened 2024-05-29
Bump ICU4X to 1.5 and cleanup Intl

jedel1043:
Some additional cleanups were made to the `Intl` services, since the ECMA-402 spec had editorial changes that hadn't been implemented.

Had to pin `temporal_rs` to the git repo until we publish a new version, but it would be better to fix the panics before doing so.

Removed some hacks that we had in place for locale resolution. We still have a small hack for `Segmenter`, but that will possibly be resolved in the future with locale-aware segmenters.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ thin-vec = "0.2.13"
 time = {version = "0.3.36", default-features = false, features = ["local-offset", "large-dates", "wasm-bindgen", "parsing", "formatting", "macros"]}
 tinystr = "0.7.5"
 log = "0.4.21"
-simple_logger = "4.3.3"
+simple_logger = "5.0.0"
 cargo_metadata = "0.18.1"
 trybuild = "1.0.95"
 rayon = "1.10.0"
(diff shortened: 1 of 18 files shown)


Cargo.toml
```diff
@@ -107,15 +107,15 @@ tap = "1.0.1"
...
+itertools = { version = "0.13.0", default-features = false }
 portable-atomic = "1.6.0"
 bytemuck = { version = "1.15.0", default-features = false }
 arrayvec = "0.7.4"
 intrusive-collections = "0.9.6"
 cfg-if = "1.0.0"
 either = "1.12.0"
 sys-locale = "0.3.1"
-temporal_rs = "0.0.2"
+temporal_rs = { git = "https://github.com/boa-dev/temporal.git" }
```
HalidOdat, line 118:
```suggestion
temporal_rs = { git = "https://github.com/boa-dev/temporal.git", rev = "61a05fbb7c72353deda72a3df0e6887d65b840d2" }
```

We may want to fix to a hash?

HalidOdat (approved):
Looks good! Check the comment if makes sense :)

jedel1043:
Managed to fix all the Intl regressions. The rest of regressions are caused by test changes for RegExp with CLDR 15.1

cc @raskad

raskad (approved):

raskad:
> The remaining regressions are caused by new test changes for RegExp with CLDR 15.1

I will look into updating regress with 15.1 :)

test262_config.toml
```diff
@@ -57,6 +57,10 @@ features = [
     # https://github.com/tc39/proposal-set-methods
     "set-methods",
 
+    # Promise.try
+    # https://github.com/tc39/proposal-promise-try
+    "promise-try",
+
```
HalidOdat, line 63: We can remove this, because it was implemented in #3800 :)

PR #3870, opened 2024-05-30
Fix base objects in `with` statements

raskad:

- Fix base objects in `with` statements


jedel1043 (approved):
Good catch!

nekevss (approved):
LGTM :smile:

PR #3872, opened 2024-06-04
Add a js_class to implement the Class trait without boilerplate

hansl:
This also adds a JsInstance that verifies that `this` is of the proper class, and an `Ignore` that ignore arguments.

The syntax is a bit special because of limitations of macro_rules. For example, the way fields are defined. It was impossible to keep both assignments (e.g. `public field = 123;`) and dynamic fields.

This reduces significantly the boilerplate for declaring classes. The example in class.rs is more than twice as long (if you remove comments) than the macro is.

### DIFF
--- a/core/interop/src/lib.rs
+++ b/core/interop/src/lib.rs
@@ -353,6 +353,7 @@ impl<T: NativeObject> JsClass<T> {
     ///
     /// This does not panic if the type is wrong, as the type is checked
     /// during the construction of the `JsClass` instance.
+    #[must_use]
     pub fn borrow(&self) -> Ref<'_, T> {
         self.inner.downcast_ref::<T>().unwrap()
     }
@@ -362,6 +363,7 @@ impl<T: NativeObject> JsClass<T> {
     /// # Panics
     ///
     /// Panics if the object is currently mutably borrowed.
+    #[must_use]
     pub fn borrow_mut(&self) -> Option<RefMut<'_, ErasedObject, T>> {
         self.inner.downcast_mut::<T>()
     }
@@ -378,7 +380,7 @@ impl<'a, T: NativeObject + 'static> TryFromJsArgument<'a> for JsClass<T> {
                 return Ok((
                     JsClass {
                         inner: object.clone(),
-                        _ty: Default::default(),
+                        _ty: PhantomData,
                     },
                     rest,
                 ));
@@ -607,6 +609,7 @@ fn class() {
     }
 
     impl Test {
+        #[allow(clippy::needless_pass_by_value)]
         fn get_value(this: JsClass<Test>) -> i32 {
             this.borrow().value
         }
--- a/core/interop/src/lib.rs
+++ b/core/interop/src/lib.rs
@@ -1,15 +1,15 @@
 //! Interop utilities between Boa and its host.
 
 use boa_engine::module::SyntheticModuleInitializer;
-use boa_engine::object::{ErasedObject, Ref, RefMut};
+use boa_engine::object::Object;
 use boa_engine::value::TryFromJs;
 use boa_engine::{
     Context, JsNativeError, JsResult, JsString, JsValue, Module, NativeFunction, NativeObject,
 };
-use std::marker::PhantomData;
 use std::ops::Deref;
 
 pub use boa_engine;
+use boa_gc::{GcRef, GcRefMut};
 pub use boa_macros;
 
 pub mod loaders;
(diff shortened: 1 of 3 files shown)


jedel1043 (commented):
Really impressive work! Definitely having a boilerplate-free way to create JsClasses is a thing we should have.

I was wondering, what is the benefit of doing it with a macro_rules vs a procedural macro? I'm just worried about using macro_rules for this since it would pretty much disable all IDE autocompletions vs a procedural macro, which is a lot nicer to IDEs.

I was envisioning something like wasm_bindgen's [`getter` and `setter` attributes](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/getter-and-setter.html)

hansl:
It was just easier for me to build a macro, and I wanted to have a more JS-ey way to describe the class. It would make sense to have a proc macro at some point, I just don't have the time yet. This reduces a lot of code already.

jedel1043:
> It was just easier for me to build a macro, and I wanted to have a more JS-ey way to describe the class. It would make sense to have a proc macro at some point, I just don't have the time yet. This reduces a lot of code already.

Fail enough. I think we can start by offering a macro rules, then switch to a proc macro in the future.

core/interop/src/lib.rs
```diff
@@ -305,6 +325,74 @@ impl<'a, T: TryFromJs> TryFromJsArgument<'a> for JsThis<T> {
...
+}
+
+/// Captures a class instance from the `this` value in a JS function. The class
+/// will be a non-mutable reference of Rust type `T`, if it is an instance of `T`.
+///
+/// To have more flexibility on the parsing of the `this` value, you can use the
+/// [`JsThis`] capture instead.
+#[derive(Debug, Clone)]
+pub struct JsClass<T: NativeObject> {
+    inner: boa_engine::JsObject,
```
jedel1043, line 343: You can use `boa_engine::JsObject<T>` and `JsObject::downcast` instead, which gives us infallible downcasts after the first downcast.

core/interop/src/lib.rs
```diff
@@ -305,6 +325,74 @@ impl<'a, T: TryFromJs> TryFromJsArgument<'a> for JsThis<T> {
...
+    ) -> JsResult<(Self, &'a [JsValue])> {
+        if let Some(object) = this.as_object() {
+            if object.downcast_ref::<T>().is_some() {
+                return Ok((
+                    JsClass {
+                        inner: object.clone(),
+                        _ty: PhantomData,
+                    },
+                    rest,
+                ));
```
jedel1043, line 386: See above. This would be a `JsObject::downcast` instead.

jedel1043 (changes requested):
Thank you! I have some suggestions that remove the `unwrap`s.

core/interop/src/lib.rs
hansl: Sure, but it downcasts to `Object<T>`, not `T` :(

hansl: I don't know how to get a `T` (or more specifically a `&mut T`) from an `Object<T>`.

Ideally I want a `GcRef<'_, T>` (and `GcRefMut<'_, _, T>`) directly.

jedel1043: Nono, there's a `JsObject::downcast` that converts from `JsObject` to `JsObject<T>, which makes it possible to  access the data between borrows without having to call `downcast_mut` or `downcast_ref`

hansl: Yes but that data is non-mut. Are you suggesting I should add this to `Object<T>`?

```rust
pub fn data_mut(&mut self) -> &mut T {
  self.data
}
```

jedel1043: After that, if you want to map to the inner data, you can do

`GcRef::map(o.borrow(), |o| &o.data)`

Same with `GcRefMut`

jedel1043: Ah right, we don't have `data_mut()`. Yeah, just add it to `Object<T>`

hansl: Uh for some reason I missed `GcRef[Mut]::map`. Probably because it's still unstable in std. Thanks!

hansl:
@jedel1043 PTAL

jedel1043 (approved):
Looks perfect to me!

nekevss (approved):
Overall, looks good to me! Thanks for all your work on this!

PR #3876, opened 2024-06-14
Fix hashbang comments by using proper goal symbols

raskad:

- Add a lexer goal symbol for hashbang comments and regexes. This goal is only used for the first lexical token of a `Script` or `Module` parse goal.


jedel1043 (approved):
LGTM! Pretty happy to follow the spec a bit more closely in this case.

nekevss (approved):

PR #3879, opened 2024-06-18
Fix AsyncGenerator to correctly handle `return` inside `then`

jedel1043:


jasonwilliams (approved):
LGTM

core/engine/src/tests/async_generator.rs
```diff
@@ -0,0 +1,121 @@
...
+                    if (count < 2) {
+                        count++;
+                        g.return();
+                    }
+                    return;
+                },
+            });
+            let first = g.next();
+            let second = g.next();
+            let ret = g.return();
```
jasonwilliams, line 105: was you meant to be asserting the `count` value here?

nekevss (approved):
Overall, LGTM :smile:

core/engine/src/tests/async_generator.rs
jedel1043: Not really, since the count is only to avoid entering an infinite loop, but I can add an assertion for that just in case

jedel1043: Added

PR #3882, opened 2024-06-27
Remove `FormalParameterList` from `CodeBlock`

HalidOdat:
This is a simple change that removes the formal parameter list from codeblock which was only needed in non-strict mode and compute the required binding indices when needed during compilation, instead of every function call. This also reduces the `CodeBlock` size by 16 bytes.


raskad (approved):

jedel1043 (approved):
Nice optimization!

PR #3883, opened 2024-06-29
Combine `HasProperty` and `Get` operations when possible

raskad:
This PR creates a new object operation that combines the `HasProperty` and `Get` operations.

In some cases in the spec `HasProperty` and `Get` are called on an object, with the same arguments, directly following each other. Both operations boil down to (in most cases) `[[GetOwnProperty]]`, just that `HasProperty` does not return the value. To avoid duplicating the exact same operations we can combine both operations.

I have some Flame Graphs from a `GetName` operation to visualize the benefit:

Before:
image
After:
image

I also ran the benchmarks:

Before:
```txt
RESULT Richards 54.3
RESULT DeltaBlue 58.2
RESULT Crypto 68.4
RESULT RayTrace 186
RESULT EarleyBoyer 170
RESULT RegExp 58.6
RESULT Splay 195
RESULT NavierStokes 159
SCORE 103
```
After:
```txt
RESULT Richards 55.1
RESULT DeltaBlue 59.6
RESULT Crypto 68.1
RESULT RayTrace 187
RESULT EarleyBoyer 176
RESULT RegExp 58.5
RESULT Splay 195
RESULT NavierStokes 160
SCORE 104
```

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -811,11 +811,9 @@ impl Array {
                 for k in 0..len {
                     // 1. Let P be ! ToString(𝔽(k)).
                     // 2. Let exists be ? HasProperty(E, P).
-                    let exists = item.has_property(k, context)?;
                     // 3. If exists is true, then
-                    if exists {
-                        // a. Let subElement be ? Get(E, P).
-                        let sub_element = item.get(k, context)?;
+                    // 3.a. Let subElement be ? Get(E, P).
+                    if let Some(sub_element) = item.try_get(k, context)? {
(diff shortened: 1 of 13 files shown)


jedel1043:
Really nice optimization! I'm seeing that all our usages of the new method are made internally, so I was curious if we could set up inline caching manually to optimise the calls even more.

jedel1043 (approved):

HalidOdat:
Yeah that was one of the things that could be optimized, though I didn't find a good way of doing it, one potentially viable idea I had was to have a global hash table on the context (with limited entries).

HalidOdat (approved):
Looks good to me! :)

PR #3888, opened 2024-07-02
Refactor call frame access to avoid panic checks

raskad:
This PR slightly changes how we store and access call frames in the vm. In addition to the `frames` stack, there is now a `frame` field that always holds the current call frame. The main reason for this is, so that we can avoid panic checks in critical paths.

I also removed some duplicate calls to `Vm::frame` and `Vm::fram_mut` and grouped them where possible. This should not be that relevant anymore, since now it's just a field access.

Here two graphs to visualize the change. These are running in `dev` profile so functions are visible. Here the bigger impact seems to actually be the deref of the `frames` vec, but I'm not sure if the actual cost in `release` looks different.

Before:
image

After:
image

And here some benchmarks. I did not run them multiple times, the impact seems a bit high to me:

Before:
```txt
RESULT Richards 55.7
RESULT DeltaBlue 57.8
RESULT Crypto 67.7
RESULT RayTrace 187
RESULT EarleyBoyer 174
RESULT RegExp 60.4
RESULT Splay 192
RESULT NavierStokes 155
SCORE 103
```

After:
```txt
RESULT Richards 59.1
RESULT DeltaBlue 59.7
RESULT Crypto 72.5
RESULT RayTrace 197
RESULT EarleyBoyer 182
RESULT RegExp 60.0
RESULT Splay 189
RESULT NavierStokes 172
SCORE 108
```

### DIFF
--- a/core/engine/src/builtins/generator/mod.rs
+++ b/core/engine/src/builtins/generator/mod.rs
@@ -96,8 +96,9 @@ impl GeneratorContext {
         let rp = frame.rp;
         context.vm.push_frame(frame);
 
-        context.vm.frame_mut().rp = rp;
-        context.vm.frame_mut().set_exit_early(true);
+        let frame = context.vm.frame_mut();
+        frame.rp = rp;
+        frame.set_exit_early(true);
 
         if let Some(value) = value {
             context.vm.push(value);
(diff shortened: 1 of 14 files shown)


HalidOdat (approved):
Nice optimization! :)

jedel1043 (approved):

PR #3890, opened 2024-07-04
Remove `Temporal.Calendar` and `Temporal.TimeZone`

jedel1043:
We will probably miss some changes when the spec is patched with the removals, but we can do a pass through `temporal` when that time comes.

Also checked all the regressions and they're failing because they try to access the removed `Temporal.Calendar` in some way or another, so that's fine.


nekevss:
Hmm, would it be worth while doing this in an intermediate stage of removing the custom functionality and waiting for the PR on the specification for the removal of the entire builtin?

jedel1043:
> Hmm, would it be worth while doing this in an intermediate stage of removing the custom functionality and waiting for the PR on the specification for the removal of the entire builtin?

That was the plan, but almost everything on the `Temporal.Calendar` and `Temporal.TimeZone` builtins depended on things that were removed in `temporal_rs`, so I had to remove all methods from those.

raskad (approved):
Looks good, I'd be ok with merging it now.

nekevss:
The custom functionality was removed from both, but weren't the core methods kept for calendars and time zones?

jedel1043:
> The custom functionality was removed from both, but weren't the core methods kept for calendars and time zones? 

On the spec side? Nope, the plan is to delete both builtins and only be able to access the calendar/timezone by IANA ID or `ca` value, respectively.

On the `temporal_rs` side? Yes, we have the core methods still, it was just easier to remove everything instead of having to patch custom versions of those methods with all the custom calendar functionality removed.

nekevss:
That makes sense. I'm just a little hesitant to preemptively remove the builtins based on the removal issues and plans without confirming with the champions the most recent consensus and timeline for the updates to the specification. Because the builtins are currently in the spec and test suite, and there is a history of things changing (even though that seems to be slowing down).

jedel1043:
I mean, there's already a PR to remove those builtins from the tests, and it has been approved already, so...
https://github.com/tc39/test262/pull/4119

nekevss (approved):
Huh, they hadn't updated the proposal repository or the specification for temporal. Given that the PR to remove the tests has been merged, then let's move forward with merging this.

HalidOdat (approved):
Looks good to me! :)

PR #3892, opened 2024-07-06
Update Temporal rounding and implement additional methods

nekevss:
This Pull Request is part of the continued work for Temporal.

It changes the following:

- Makes updates for the new rounding functionality
- Implements some additional methods from `temporal_rs`
- Some general cleanup


nekevss:
Looks like the panic was caused by an `assert`. The panic is removed by boa-dev/temporal#73

jedel1043 (approved):
Looks good!

raskad (approved):

PR #3893, opened 2024-07-06
Refactor environment stack to remove some panics

raskad:
Thie PR refactors the environment stack to avoid panics. The main change is to put the global declarative environment in a separete field instead of on the stack itself. That way we can safeley avoid panics.

~~There was only a very slight difference in benchmarks, but I think reducing the panics is worth it either way.~~
After running the benchmarks a few times they look pretty good:

Before:

```txt
RESULT Richards 60.6
RESULT DeltaBlue 63.4
RESULT Crypto 74.4
RESULT RayTrace 202
RESULT EarleyBoyer 186
RESULT RegExp 61.1
RESULT Splay 199
RESULT NavierStokes 175
SCORE 111
```

After:

```txt
RESULT Richards 65.8
RESULT DeltaBlue 69.6
RESULT Crypto 77.1
RESULT RayTrace 201
RESULT EarleyBoyer 199
RESULT RegExp 59.9
RESULT Splay 226
RESULT NavierStokes 176
SCORE 116
```

### DIFF
--- a/core/engine/src/environments/runtime/mod.rs
+++ b/core/engine/src/environments/runtime/mod.rs
@@ -299,7 +299,7 @@ impl EnvironmentStack {
     ) {
         let env = match environment {
             BindingLocatorEnvironment::GlobalObject
-            | BindingLocatorEnvironment::GlobalDecalrative => self.global(),
+            | BindingLocatorEnvironment::GlobalDeclarative => self.global(),
             BindingLocatorEnvironment::Stack(index) => self
                 .stack
                 .get(index as usize)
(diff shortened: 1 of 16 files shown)


core/engine/src/environments/runtime/mod.rs
```diff
@@ -490,6 +477,14 @@ pub(crate) enum BindingLocatorError {
     Silent,
 }
 
+/// The environment in which a binding is located.
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum BindingLocatorEnvironment {
+    GlobalObject,
+    GlobalDecalrative,
```
jedel1043, line 484:
```suggestion
    GlobalDeclarative,
```

jedel1043 (changes requested):
Really nice enhancement!

HalidOdat (approved):
Nice work, looks good to me! :)

PR #3897, opened 2024-07-07
Fix HomeObject for private class methods

raskad:

- Fix HomeObject for private class methods

### DIFF
--- a/core/engine/src/bytecompiler/class.rs
+++ b/core/engine/src/bytecompiler/class.rs
@@ -423,21 +423,33 @@ impl ByteCompiler<'_> {
                             self.emit_with_varying_operand(Opcode::PushClassPrivateSetter, index);
                         }
                         MethodDefinition::Ordinary(expr) => {
+                            self.emit(Opcode::RotateLeft, &[Operand::U8(3)]);
+                            self.emit_opcode(Opcode::Dup);
+                            self.emit(Opcode::RotateRight, &[Operand::U8(4)]);
                             self.method(expr.into());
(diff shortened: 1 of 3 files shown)


raskad:
Relevant test passing here is: `test/language/statements/class/elements/super-access-inside-a-private-method.js (previously Failed)`.

core/engine/src/bytecompiler/class.rs
```diff
@@ -423,21 +423,33 @@ impl ByteCompiler<'_> {
                             self.emit_with_varying_operand(Opcode::PushClassPrivateSetter, index);
                         }
                         MethodDefinition::Ordinary(expr) => {
+                            self.emit(Opcode::RotateLeft, &[Operand::U8(3)]);
+                            self.emit_opcode(Opcode::Dup);
+                            self.emit(Opcode::RotateRight, &[Operand::U8(4)]);
```
jedel1043, line 428: Thought: I have seen this pattern in a lot of places. Maybe it would be good to extract this operation into a common compiler method. Something like `dup_indexed` or similar.

jedel1043 (approved):
Looks good! I have a small discussion point that doesn't block merging.

core/engine/src/bytecompiler/class.rs
raskad: Totaly agree. I think this might be something for the bytecode optimizer too. I would wait for https://github.com/boa-dev/boa/pull/3798 until we work on that since that is probably gonna change some things.

nekevss (approved):
LGTM

PR #3900, opened 2024-07-08
Updates to temporal_rs version and temporal methods

nekevss:
This Pull Request is related to the ongoing work for temporal.

It changes the following:

- Bumps `temporal_rs` to `0.0.3`
- Implements more of the functionality already implemented in `temporal_rs`
- Adds `to_temporal_time` and `to_temporal_datetime` abstract operations

### DIFF
--- a/core/engine/src/builtins/temporal/mod.rs
+++ b/core/engine/src/builtins/temporal/mod.rs
@@ -254,7 +254,8 @@ pub(crate) fn to_relative_temporal_object(
         JsValue::Object(relative_to_obj) => Some(relative_to_obj.into()),
         _ => None,
     }
-    .map(|plane_date| Ok::<_, JsError>(to_temporal_date(&plane_date, None, context)?.inner))
+    .map(|plain_date| Ok::<_, JsError>(to_temporal_date(&plain_date, None, context)))
+    .transpose()?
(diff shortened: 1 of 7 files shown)


nekevss:
After a spot check, the new broken tests appear to be previous false positives on `from` method calls, so there shouldn't be any regressions.

core/engine/src/builtins/temporal/mod.rs
```diff
@@ -254,7 +254,8 @@ pub(crate) fn to_relative_temporal_object(
         JsValue::Object(relative_to_obj) => Some(relative_to_obj.into()),
         _ => None,
     }
-    .map(|plane_date| Ok::<_, JsError>(to_temporal_date(&plane_date, None, context)?.inner))
+    .map(|plain_date| Ok::<_, JsError>(to_temporal_date(&plain_date, None, context)))
```
jedel1043, line 257: Thought: The previous match could probably be rewritten to use an early return on `None`. It would simplify this part by a lot.

jedel1043 (commented):
Overall good! I just have a small suggestion.

jasonwilliams (approved):

jedel1043 (approved):
LGTM!

PR #3903, opened 2024-07-08
update versions and ABOUT files

jasonwilliams:
Bump versions to 0.19 and update ABOUT files with all the crates

### DIFF
--- a/ABOUT.md
+++ b/ABOUT.md
@@ -10,13 +10,16 @@ Try out the most recent release with Boa's live demo
 
 ## Boa Crates
 
+- [**`boa_cli`**][cli] - Boa's CLI && REPL implementation
 - [**`boa_ast`**][ast] - Boa's ECMAScript Abstract Syntax Tree.
 - [**`boa_engine`**][engine] - Boa's implementation of ECMAScript builtin objects and execution.
 - [**`boa_gc`**][gc] - Boa's garbage collector.
+- [**`boa_icu_provider`**][icu] - Boa's ICU4X data provider.
 - [**`boa_interner`**][interner] - Boa's string interner.
+- [**`boa_interop`**][interop] - Boa's interop with other languages.
+- [**`boa_macros`**][macros] - Boa's macros.
 - [**`boa_parser`**][parser] - Boa's lexer and parser.
 - [**`boa_profiler`**][profiler] - Boa's code profiler.
-- [**`boa_icu_provider`**][icu] - Boa's ICU4X data provider.
 - [**`boa_runtime`**][runtime] - Boa's WebAPI features.
 - [**`boa_string`**][string] - Boa's ECMAScript string implementation.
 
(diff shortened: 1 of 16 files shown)


ABOUT.md
```diff
@@ -10,13 +10,16 @@ Try out the most recent release with Boa's live demo
 
 ## Boa Crates
 
+- [**`boa_cli`**][cli] - Boa's CLI && REPL implementation
 - [**`boa_ast`**][ast] - Boa's ECMAScript Abstract Syntax Tree.
 - [**`boa_engine`**][engine] - Boa's implementation of ECMAScript builtin objects and execution.
 - [**`boa_gc`**][gc] - Boa's garbage collector.
+- [**`boa_icu_provider`**][icu] - Boa's ICU4X data provider.
 - [**`boa_interner`**][interner] - Boa's string interner.
+- [**`boa_interop`**][interop] - Boa's interop with other languages.
```
jedel1043, line 19: Hmm, you comment about the `boa_interop` crate made me reconsider its name, since it doesn't really have anything to do with interop between languages, just interop between API. Maybe we should change it to `boa_adapters`?

nekevss: Maybe be a bit more specific about Boa's Rust interop? I feel like FFI/`boa_ffi` would be pretty standard name wise for a crate that would interact with different languages.

jedel1043: Current ideas:

- boa_adapters
- boa_wrappers
- boa_helpers
- boa_utils
- boa_rust_interop
- boa_bindings

nekevss: My preferred options would probably be:

- boa_rust_interop
- boa_helpers
- boa_utils (maybe `boa_js_utils`?)

I'm hesitant about using `boa_utils` if we plan to actually have a `util` directory in the future for misc crates. It may cause some confusion for newer contributors.

hansl: `helpers` would have my vote. It's not just bindings (there are macros in there) and adapters has a specific design meaning which doesn't apply.

Also, it could just be that we don't publish it yet and just fold it into either existing packages and/or new packages. Like `boa_modules` to create modules, `boa_macros` for macro stuff, `boa_engine` for the `into_js_*_function`.

raskad: > Also, it could just be that we don't publish it yet and just fold it into either existing packages and/or new packages. Like boa_modules to create modules, boa_macros for macro stuff, boa_engine for the into_js_*_function.

I would second this idea. We probably will split up the `boa_engine` crate further in the near future so there may be an intuitive place for all of the funtionality.

jedel1043: > Also, it could just be that we don't publish it yet and just fold it into either existing packages and/or new packages. Like boa_modules to create modules, boa_macros for macro stuff, boa_engine for the into_js_*_function.

Huh, now that I think about this yeah, we could merge it into the engine crates for now and extract it in the future if people ask us to do that. This could also make it more discoverable in a way, because you won't need to pull a separate crate to have the nicer APIs to create classes, functions and methods.

jedel1043: Okay I think we're in agreement about merging `boa_interop` into other crates, so I'll open an issue for it.

ABOUT.md
```diff
@@ -27,8 +29,11 @@ Try out the most recent release with Boa's live demo
 [engine]: https://docs.rs/boa_engine/latest/boa_engine/index.html
 [gc]: https://docs.rs/boa_gc/latest/boa_gc/index.html
 [interner]: https://docs.rs/boa_interner/latest/boa_interner/index.html
+[interop]: https://docs.rs/boa_interop/latest/boa_interop/index.html
```
raskad, line 32: The interop link here is still in the about files here.

jasonwilliams: good spot!

jedel1043 (approved):

raskad (approved):

PR #3907, opened 2024-07-09
Implement lossless TryFromJs for integers from f64

HalidOdat:
Implement conversion from floating point values that can be converted to integer values losslessly.

The spec does not distinguish between integer numbers and floating point, so I don't think `4` should succeed and `4.0` fail, according to the spec they have the same type and are equal.

### DIFF
--- a/core/engine/src/value/conversions/try_from_js.rs
+++ b/core/engine/src/value/conversions/try_from_js.rs
@@ -156,12 +156,58 @@ impl TryFromJs for f64 {
     }
 }
 
+trait ToF64: Sized {
+    fn cast_to_f64(self) -> f64;
+    fn cast_from_f64(v: f64) -> Self;
+}
+
+macro_rules! impl_to_f64 {
+    ($type_:ident) => {
+        impl ToF64 for $type_ {
+            #[inline]
+            // NOTE: Lint only applies to types that are lossless conversion to f64.
+            #[allow(clippy::cast_lossless)]
+            fn cast_to_f64(self) -> f64 {
+                self as f64
+            }
+            #[inline]
+            fn cast_from_f64(v: f64) -> Self {
+                v as Self
+            }
+        }
+    };
+}
+
+impl_to_f64!(i8);
+impl_to_f64!(u8);
+impl_to_f64!(i16);
+impl_to_f64!(u16);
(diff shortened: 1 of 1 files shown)


core/engine/src/value/conversions/try_from_js.rs
```diff
@@ -156,6 +156,47 @@ impl TryFromJs for f64 {
     }
 }
 
+trait ToF64: Sized {
```
nekevss, line 159: thought: Is there any gain from manually implementing this vs using `NumTraits`' `ToPrimitive`/`FromPrimitive`?

nekevss (approved):
I had one potential discussion point, but it's nothing that would block merging for me.

core/engine/src/value/conversions/try_from_js.rs
HalidOdat: I started using the `ToPrimitive`/`FromPrimitive` but found out that it's not lossless :( https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a15242f1b391f3de4ba30314190e4add

jedel1043: I'm not following why we need to care about the floating point case when we're trying to convert from an integer f64 to an integer

nekevss: Huh, that's true. Then `NumTraits` definitely isnt needed

jedel1043: I think we're slightly talking over each other. I'm talking about still having the check for equality with the bits representation of the original number, but just using `To/FromPrimitive` to do the conversion to integer instead of a new trait. Hence:

```rust
fn from_f64<T: ToPrimitive + FromPrimitive>(v: f64) -> Option<T> {
    let i = T::from_f64(v)?;
    if i.to_f64()?.to_bits() != v.to_bits() {
        return None;
    }
    Ok(i)
}
```

HalidOdat: I see, yeah we can use it, `AsPrimitive` seems perfect for the job, thank for the suggestion :)

jedel1043 (approved):
LGTM!

Issue #3896, opened 2024-07-07
Boa's start-up time getting hit with GC Allocation


jasonwilliams:
Boa spends a large proportion of time in realm creation comparitive to actually running the code. 
It seems a lot of this is due to building out the global objects on the heap and storing them causing the GC to do multiple collections before the program has even started.

We may wish to look at some sort of snapshot for built-ins or staticlly storing them somewhere (switch of GC for built-ins)

You can load the below profiler into [ui.perfetto.dev](https://ui.perfetto.dev/) or chrome developer tools and take a look. 
![image](https://github.com/boa-dev/boa/assets/936006/a7a178d6-3d6a-47b9-97a0-38d150c94779)

[chrome_profiler.json](https://github.com/user-attachments/files/16120205/chrome_profiler.json)

nekevss:
Raising the threshold at the below might help.

https://github.com/boa-dev/boa/blob/fc2a6e09969772feba98eaa89aaf89ca4797e925/core/gc/src/lib.rs#L66

raskad:
> Raising the threshold at the below might help.
> 
> https://github.com/boa-dev/boa/blob/fc2a6e09969772feba98eaa89aaf89ca4797e925/core/gc/src/lib.rs#L66

Yeah I also think we should raise the default threshold and make it configurable from outside of `boa_gc`.

jasonwilliams:
```
Hello world!
undefined
dhat: Total:     611,648 bytes in 3,019 blocks
dhat: At t-gmax: 382,386 bytes in 2,283 blocks
dhat: At t-end:  354,488 bytes in 2,258 blocks
```

Also I've noticed that in DHAT we spend a lot of time growing the vector for property_table.
![image](https://github.com/boa-dev/boa/assets/936006/8dfa73ac-cea0-478e-8b80-10a4fb8a4ce2)

If we know ahead of time how many static methods are on a builtin, we should be passing this number through so it can create the `vec::with_capacity(n)`

jasonwilliams:
We’re not able to set each property_table with a capacity right now due to the fact that for each template the inner shape [is cloned](https://github.com/boa-dev/boa/blob/main/core/engine/src/object/shape/shared_shape/template.rs#L25) from the first shape. 

We may need to set new root shapes [here](https://github.com/boa-dev/boa/blob/main/core/engine/src/context/intrinsics.rs#L1397), each one having its own set of prototype storage capacity and property capacity.

@HalidOdat thoughts?

PR #3908, opened 2024-07-09
bump gc threshold

jasonwilliams:
As mentioned in https://github.com/boa-dev/boa/issues/3896, Boa's GC threshold is set too low. This was causing too many GC collections before the interpreter even started running.

We see an improvement to the 2.

```
Benchmark main: ./target/release/boa
  Time (mean ± σ): 2.2 ms ± 0.2 ms [User: 0.6 ms, System: 0.1 ms]
  Range (min … max): 2.0 ms … 3.4 ms 827 runs
```

```
Benchmark PR: ./target/release/boa
  Time (mean ± σ):       2.1 ms ±   0.1 ms    [User: 0.6 ms, System: 0.1 ms]
  Range (min … max):     1.9 ms …   2.8 ms    924 runs
```

Looking at the traces:

Before:
image

After (there are no GC collections in this trace)
image

1MB seems like a good start, we can always bump it higher or adjust after this, V8's nursery size before GC kicks in seems to be between 1-8MB,

### DIFF
--- a/core/gc/src/lib.rs
+++ b/core/gc/src/lib.rs
@@ -65,7 +65,7 @@ struct GcConfig {
 impl Default for GcConfig {
     fn default() -> Self {
         Self {
-            // Start at 1MB, the nursary for V8 is ~1-8MB and SM can be up to 16MB, so we have room to increase if needed.
+            // Start at 1MB, the nursary size for V8 is ~1-8MB and SM can be up to 16MB
             threshold: 1_000_000,
             used_space_percentage: 70,
         }
--- a/core/gc/src/lib.rs
+++ b/core/gc/src/lib.rs
@@ -65,6 +65,7 @@ struct GcConfig {
 impl Default for GcConfig {
     fn default() -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


raskad (approved):

core/gc/src/lib.rs
```diff
@@ -63,7 +65,8 @@ struct GcConfig {
 impl Default for GcConfig {
     fn default() -> Self {
         Self {
-            threshold: 1024,
+            // Start at 1MB, the nursary size for V8 is ~1-8MB and SM can be up to 16MB
+            threshold: 1_000_000,
```
nekevss, line 69: Super nitpicky, but since it would be ideal to be 2^n, would it be worthwhile for the threshold to start at `1_048_576`.

Either way, I'm fine with merging.

nekevss (approved):
Small comment, non-blocking though.

core/gc/src/lib.rs
jasonwilliams: Yes good point

PR #3911, opened 2024-07-10
 Add a way to add setters/getters in js_class! 

hansl:
And update documentation for more explanation around the macro.

### DIFF
--- a/core/interop/src/lib.rs
+++ b/core/interop/src/lib.rs
@@ -616,6 +616,7 @@ fn can_throw_exception() {
 #[test]
 fn class() {
     use boa_engine::class::{Class, ClassBuilder};
+    use boa_engine::property::Attribute;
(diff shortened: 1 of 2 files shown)


hansl:
Converting to draft as I've come up with a better interface for getter/setter. Basically:

```rust
prop name_of_prop as "nameOfProp" {
  fn getter(...) {}
  fn setter(...) {}
}
```

with getter and setter optional. That way we don't need 3 sections for properties and it's a bit simpler.

hansl:
Okay this does use the new API now.

jedel1043 (approved):
Good work! Looks perfect to me.

jedel1043:
Thought: the syntax is getting a bit too DSL-like, so I'll probably see if I can start working on that derive macro.

raskad (approved):

PR #3915, opened 2024-07-12
Add string builder to build `JsString`

CrazyboyQCD:
There are many places where `Vec` is used as a builder to create `JsString`. 
This adds `JsStringBuilder` as a more natural choice.


core/string/src/lib.rs
```diff
@@ -1144,3 +1145,348 @@ impl ToStringEscaped for [u16] {
...
+    /// Returns the total number of elements the inner `RawJsString` can hold without reallocating
+    #[inline]
+    #[must_use]
+    pub fn capacity(&self) -> usize {
+        self.cap
+    }
+
+    /// Returns the initalized bits of inner `RawJsString`.
+    #[must_use]
+    fn initialized_bits(&self) -> usize {
```
HalidOdat, line 1251: I think renaming this to something like `allocated_bytes_len()` would be a better name, feel free to suggest another name if you think it's not accurate.

core/string/src/lib.rs
```diff
@@ -1144,3 +1145,348 @@ impl ToStringEscaped for [u16] {
...
+        }
+        // SAFETY:
+        // Capacity has been expanded to be large enough to hold elements.
+        unsafe {
+            self.push_unchecked(v);
+        }
+    }
+
+    /// Push elements from slice to `JsStringBuilder` without doing capacity check.\
+    /// Unlike the standard vector, our holded element types are only `u8` and `u16`, which is [`Copy`] derived,\
```
HalidOdat, line 1357: Why do we have backslashes at the end here, if it's to add newline then you can just separate then with a blank line.

```
    /// Push elements from slice to `JsStringBuilder` without doing capacity check.
    ///
    /// Unlike the standard vector, our holded element types are only `u8` and `u16`, which is [`Copy`] derived,
```

core/string/src/lib.rs
```diff
@@ -1144,3 +1145,348 @@ impl ToStringEscaped for [u16] {
...
+                .extend(Layout::array::<T>(self.capacity()).unwrap_unchecked())
+                .unwrap_unchecked()
+                .0
+                .pad_to_align()
+        }
+    }
+
+    /// Returns the pointer of `data` of inner `RawJsString`.
+    /// # Safety
+    /// Caller should ensure that the inner is allocated.
```
HalidOdat, line 1298:
```suggestion
    /// Returns the pointer of `data` of inner `RawJsString`.
    ///
    /// # Safety
    ///
    /// Caller should ensure that the inner is allocated.
```

Minor formatting issues.

core/string/src/lib.rs
```diff
@@ -1144,3 +1145,348 @@ impl ToStringEscaped for [u16] {
...
+        unsafe {
+            ptr::copy_nonoverlapping(v.as_ptr(), self.data().add(self.len()), v.len());
+        }
+        self.len += v.len();
+    }
+
+    /// push elements from slice to `JsStringBuilder`.
+    pub fn extend_from_slice(&mut self, v: &[T]) {
+        let required_cap = self.len() + v.len();
+        if required_cap >= self.capacity() {
```
HalidOdat, line 1373:
```suggestion
        if required_cap > self.capacity() {
```

Shouldn't it be greater than if the required cap is equal to the current cap then we can copy, right?

HalidOdat (commented):
Looks good! Check my comment if they make sense :)

core/string/src/lib.rs
```diff
@@ -1144,3 +1145,348 @@ impl ToStringEscaped for [u16] {
...
+        inner: NonNull::dangling(),
+        phantom_data: PhantomData,
+    };
+
+    /// Create a new `JsStringBuilder` with capacity of zero.
+    #[inline]
+    #[must_use]
+    pub const fn new() -> Self {
+        Self::NEW
+    }
```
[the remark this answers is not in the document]
CrazyboyQCD: Sometime compiler doesn't generate good code for struct literal but for constant, I just checked this is fine for it.

core/string/src/lib.rs
CrazyboyQCD: Done.

core/string/src/lib.rs
```diff
@@ -1144,3 +1145,348 @@ impl ToStringEscaped for [u16] {
...
+
+    /// Returns the number of elements in the inner `RawJsString`, also referred to
```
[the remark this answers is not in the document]
CrazyboyQCD: Done.

core/string/src/lib.rs
CrazyboyQCD: Done.

core/string/src/lib.rs
CrazyboyQCD: Done.

core/string/src/lib.rs
CrazyboyQCD: I tested this before and found sometimes greater and equal condition gives better performance, but maybe this is just a warm-up thing. I think it's fine to just use greater condition.

CrazyboyQCD:
@HalidOdat, one more thing is that I used `realloc` instead of `alloc` before and found `alloc` has a better performance, since I can't test on other machine, you could test `realloc` to see if I'm wrong.
If you want to test it, just change the `reserve`:
```rust
    fn reserve(&mut self, new_layout: Layout) {
        // SAFETY:
        // The layout size of `RawJsString` is never zero, since it has to store
        // the length of the string and the reference count.

        let new_ptr = if self.is_dangling() {
            let new_ptr = unsafe { alloc(new_layout) };
            let Some(new_ptr) = NonNull::new(new_ptr.cast::<RawJsString>()) else {
                std::alloc::handle_alloc_error(new_layout)
            };
            new_ptr
        } else {
            use std::alloc::realloc;
            let old_ptr = self.inner.as_ptr();
            let ptr =  unsafe { realloc(ptr, new_layout, new_layout.size()) };
            let Some(new_ptr) = NonNull::new(new_ptr.cast::<RawJsString>()) else {
                std::alloc::handle_alloc_error(new_layout)
            };
            new_ptr
        };
        self.inner = new_ptr;
        let new_arr_size = new_layout.size() - DATA_OFFSET;
        self.cap = new_arr_size / Self::DATA_SIZE;
    }
```

core/string/src/lib.rs
```diff
@@ -1144,3 +1145,352 @@ impl ToStringEscaped for [u16] {
...
+    /// Caller should ensure that the inner is allocated.
+    #[must_use]
+    unsafe fn data(&self) -> *mut D {
+        // SAFETY:
+        // Caller should ensure that the inner is allocated.
+        unsafe { addr_of_mut!((*self.inner.as_ptr()).data).cast() }
+    }
+
+    #[allow(clippy::cast_ptr_alignment)]
+    fn reserve(&mut self, new_layout: Layout) {
```
HalidOdat, line 1310: Additionally `reserve()` always re-allocates, even when it's not required, for reference see std library `Vec` implementation which wraps around `RawVec`: <https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#338-357>

HalidOdat (changes requested):

HalidOdat:
```rust
            let ptr =  unsafe { realloc(ptr, new_layout, new_layout.size()) };
```

This may be unsafe because in the realloc's safety section it has the requirement that the layout passed to it is `layout must be the same layout that was used to allocate that block of memory`, so it should be the current layout.

See: <https://doc.rust-lang.org/stable/core/alloc/trait.GlobalAlloc.html#method.realloc>

HalidOdat:
I benchmarked with the current implementation (alloc) and with realloc and found that it's considerably faster, here is my benchmark code if you want to reproduce it.

```rust
fn main() {
    let mut sb = JsStringBuilder::<u8>::new();

    for _ in 0..10000 {
        sb.push(b'a');
        sb.extend_from_slice(b"bcde".as_slice());
        sb.extend([b'f', b'g'].into_iter());
    }

    let s = sb.build();
    drop(s);
}
```

```bash
hyperfine -N --warmup=10 ./sb-alloc ./sb-realloc
```

```
Benchmark 1: ./sb-alloc
  Time (mean ± σ):      10.8 ms ±   0.2 ms    [User: 5.9 ms, System: 4.6 ms]
  Range (min … max):    10.2 ms …  11.9 ms    284 runs

Benchmark 2: ./sb-realloc
  Time (mean ± σ):     728.6 µs ±  78.9 µs    [User: 565.7 µs, System: 96.5 µs]
  Range (min … max):   550.7 µs … 1324.8 µs    3970 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  ./sb-realloc ran
   14.77 ± 1.63 times faster than ./sb-alloc
```

NOTE: Changed `extend`'s reserve to be `self.reserve(Self::new_layout(self.len() + lower_bound));` to avoid the above mentioned bug.

CrazyboyQCD:
@HalidOdat 
My OS is **Windows11**, and seems `realloc` does work much better for me, what is your OS? 
And do you think I should change into `realloc`?
```shell
hyperfine -N --warmup=100  "./jsstringbuilder-realloc.exe" "./jsstringbuilder-alloc.exe"
```
```shell
Benchmark 1: ./jsstringbuilder-realloc.exe
  Time (mean ± σ):       3.0 ms ±   0.1 ms    [User: 0.8 ms, System: 1.7 ms]
  Range (min … max):     2.8 ms …   3.5 ms    1049 runs

Benchmark 2: ./jsstringbuilder-alloc.exe
  Time (mean ± σ):       3.0 ms ±   0.3 ms    [User: 0.7 ms, System: 1.7 ms]
  Range (min … max):     2.8 ms …   8.8 ms    1022 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  ./jsstringbuilder-realloc.exe ran
    1.01 ± 0.09 times faster than ./jsstringbuilder-alloc.exe
```

core/string/src/lib.rs
CrazyboyQCD: Done, but leave reserve for final shrink for less fragmentation.

CrazyboyQCD:
> Yes, there is no drawback to using realloc and even on windows (from the results you posted) there is still a small `~7%` increase in performance.

Ok

core/string/src/builder.rs
```diff
@@ -0,0 +1,423 @@
...
+        }
+        self.len += v.len();
+    }
+
+    /// push elements from slice to `JsStringBuilder`.
+    pub fn extend_from_slice(&mut self, v: &[D]) {
+        let required_cap = self.len() + v.len();
+        if required_cap > self.capacity() {
+            self.allocate(required_cap);
+        }
```
HalidOdat, line 249: Minor: Maybe we can factor out the check and call into a separate function something like `allocate_if_needed(cap: usize)`.

core/string/src/builder.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    /// Returns the capacity calculted from given layout.
+    #[must_use]
+    const fn capacity_from_layout(layout: Layout) -> usize {
+        (layout.size() - DATA_OFFSET) / Self::DATA_SIZE
+    }
+
+    /// create a new `JsStringBuilder` with specific capacity
+    #[must_use]
+    pub fn with_capacity(cap: usize) -> Self {
```
HalidOdat, line 133: We should add `#[inline]` on all public methods so the compiler can cross-crate inline them.

HalidOdat (commented):
Besides the suggested adjustments, this looks great! :)

core/string/src/builder.rs
```diff
@@ -0,0 +1,423 @@
...
+};
+
+#[doc(hidden)]
+mod private {
+    /// Inner elements represented for `JsStringBuilder`.
+    pub trait JsStringData {}
+
+    impl JsStringData for u8 {}
+    impl JsStringData for u16 {}
+}
```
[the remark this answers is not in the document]
CrazyboyQCD: Module `private` is not `pub` so it can only be only accessed in `builder.rs`, so user can't use it.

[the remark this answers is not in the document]
HalidOdat: Ahh, sorry, you are right!

Actually it may be a good to expose the `JsStringData` (but seal it) we can do this by extending a `Sealed` trait that is private and implemented on `u8` and `u16`. That way the user can use the trait in generic type bounds if they want.

```rust
mod private {
    pub trait Sealed {}

    impl Sealed for u8 {}
    impl Sealed for u16 {}
}

/// Inner elements represented for `JsStringBuilder`.
pub trait JsStringBuilder: private::Sealed {}
```

```rust
// In user code:
struct SomeUserGenericStruct<T: JsStringData> { // The user can use the trait
   builder: JsStringBuilder<T>,
   /// ...
}
```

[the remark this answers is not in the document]
CrazyboyQCD: Done.

core/string/src/builder.rs
```diff
@@ -0,0 +1,442 @@
...
+
+    impl Sealed for u8 {}
+    impl Sealed for u16 {}
+}
+
+/// Inner elements represented for `JsStringBuilder`.
+pub trait JsStringData: private::Sealed {}
+
+impl JsStringData for u8 {}
+impl JsStringData for u16 {}
```
jedel1043, line 25: I'm actually a bit worried about making the user choose the correct type of string builder instead of just handling that for them.

Can we just assume they want latin1 strings, and automatically switch to UTF-16 if they push either a character outside of the range of latin1 or an u16 value?

jedel1043 (changes requested):

core/string/src/builder.rs
CrazyboyQCD: > Can we just assume they want latin1 strings, and automatically switch to UTF-16 if they push either a character outside of the range of latin1 or an u16 value?

In `build` method there is a check for `latin1`, and with type inference I think most of the time users needn't to manually choose the specific type.

jedel1043: Yeah, but this means anyone who wants to use the API must be aware that they need to convert their `str`s into `utf16` format before pushing them. This also means that you cannot mix encodings easily e.g. push latin1, then push an u16 slice, then push more latin1. Ideally the user would be able to just push any `str` into the builder without worrying about encoding UTF-16 or checking for latin1-ness.

CrazyboyQCD: Then I think this should not put into normal `push`-like methods because `JsStringBuilder` was designed to be used as a faster, and more natural alternative of `Vec`. 
Maybe we can do something like `type MixedEncodingStringBuilder = JsStringBuilder<u16>` and provide different set of API.

jedel1043: We don't need to do that, we just need to add a new constructor for `JsStringBuilder` that assumes the UTF-16 encoding. Then, the user will have two options:
- Use `JsStringBuilder::new` without worries, but with the downside of possibly losing some performance if the input has non-latin1 bytes. AKA, easy to use, hard to misuse. They can also use this if they're really sure their input is just ascii or just latin1 characters.
` Use `JsStringBuilder::new_utf16` to squeeze some additional performance if they know the input will have characters outside latin1.

CrazyboyQCD: If we change that way, those things need to be done:
* Replace generic with a flag to determine which base to allocate.
* Add some conditional checkes for UTF-16 adding methods to change flag.
* Copy the `latin1` part when encoding is changed to `utf16`.

And those things will affect the performance, if we want to keep most of performance and code clean we could do something that restricts users' usage:
* Expose types like `Latin1StringBuilder` and `Utf16StringBuilder` with some extra `string` adding methods instead of exposing `JsStringBuilder`.
* Add some warnings to prevent something like `Latin1StringBuilder.extend_from_slice("Grüße, Jürgen ❤".as_bytes())`.

jedel1043: That sounds great! However, we still need a `JsStringBuilder` for people who only want to push `str`s without knowing if the input is Latin1 or UTF-16, and at that point it would be better to just expose some methods like `unchecked_push_latin1/utf16` on `JsStringBuilder`, then make `Latin1/Utf16StringBuilder` wrappers around that which call `unchecked_push_latin1/utf16`

CrazyboyQCD: I think a common builder like [kiesel-js's work](https://codeberg.org/kiesel-js/kiesel/src/branch/main/src/types/language/String/Builder.zig) would be better, we could export one common builder and other specified byte builders.

core/string/src/builder.rs
```diff
@@ -0,0 +1,713 @@
...
```
jedel1043, line 28: Doing a second pass I realized that maybe this abstraction is not needed? More so because pushing `u16` is always valid, but pushing `u8` needs to do latin1 checks.

Since we now use an enum for the generic `StringBuilder` that doesn't need the `JsStringData` trait, maybe we can just have different types for each builder, then just extract the common paths into utility functions.

CrazyboyQCD: > Doing a second pass I realized that maybe this abstraction is not needed?

`JsStringData` could be removed since I only export specific builders.

> maybe we can just have different types for each builder, then just extract the common paths into utility functions.

Have we already done this by using generic?

> More so because pushing u16 is always valid, but pushing u8 needs to do latin1 checks, and having separate builders for each type would allow us to reflect that in the API.

This could be done with implementations for generics seperately.

jedel1043: Yeah, I think that could potentially work.

core/string/src/builder.rs
```diff
@@ -0,0 +1,805 @@
...
+    mem::ManuallyDrop,
+    ops::AddAssign,
+    ptr::{self, addr_of_mut, NonNull},
+    str::{self},
+};
+
+/// A mutable builder to create instance of `JsString`.
+///
+#[derive(Debug)]
+pub struct JsStringBuilder<D> {
```
jedel1043, line 19: Can we add the restriction that `D: Copy`? Just to be safe, because we still want to nudge users to the correct types (`u8` or `u16`)

jedel1043 (commented):
Looking almost perfect! I just have a small final suggestion and this should be good to go

HalidOdat (approved):

PR #3917, opened 2024-07-14
Fix various parser idempotency issues and parsing errors

raskad:
This PR fixes a bunch of parser idempotency issues, but also some parsing errors where our parser was behaving non spec compliant. I found all of the issues via the `parser-idempotency` fuzzer.

There are a lot of changes here regarding the names of function expressions and their handling via `has_binding_identifier`. Since there are some more issues with that, I plan to alter our current handling to fix those and align the handling more with the spec in a follow up to this.

It changes the following:

- Fix cases where destructuring assignments would accept invalid assignment operators
- Fix ArrayLiteral formatting
- Prevent `Literal::Undefined` from being constructed by `Arbitrary` as it is only  used for optimizations
- Fix `Literal::BigInt` formatting
- Remove tracking of excluded keys in object literals from ast and get them at compile time
- Fix formatting for object literal methods
- Fix `Arbitrary` implementation for `TemplateLiteral`
- Fix formatting for unary expressions to always include a space between operator and target
- Fix formatting of function expressions that do not have a binding identifier
- Fix formatting of classes
- Fix formatting of computed property names
- Move handling of object literal function names to the compiler
- Fix required check of semicolon at the end of a `throw` statement
- Fix formatting of `TaggedTemplate`

### DIFF
--- a/core/ast/src/expression/literal/array.rs
+++ b/core/ast/src/expression/literal/array.rs
@@ -1,6 +1,6 @@
 //! Array declaration Expression.
 
-use crate::expression::operator::assign::AssignTarget;
+use crate::expression::operator::assign::{AssignOp, AssignTarget};
 use crate::expression::Expression;
 use crate::pattern::{ArrayPattern, ArrayPatternElement, Pattern};
 use crate::try_break;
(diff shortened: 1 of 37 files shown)


HalidOdat (approved):
Nice work! Just needs a rebase :)

jedel1043 (approved):

PR #3920, opened 2024-07-16
Add a js_error! macro to create opaque errors

hansl:
And as a bonus since I needed it, simplify comparison of strings and add PartialEq<&str> for JsString.

### DIFF
--- a/core/engine/src/value/conversions/mod.rs
+++ b/core/engine/src/value/conversions/mod.rs
@@ -39,24 +39,6 @@ impl From<char> for JsValue {
     }
 }
 
-impl From<&str> for JsValue {
-    #[inline]
-    fn from(value: &str) -> Self {
-        let _timer = Profiler::global().start_event("From<&str>", "value");
-
-        Self::from(js_string!(value))
-    }
-}
-
-impl From<String> for JsValue {
-    #[inline]
-    fn from(value: String) -> Self {
-        let _timer = Profiler::global().start_event("From<&str>", "value");
-
-        Self::from(JsString::from(value))
-    }
-}
-
 impl From<JsSymbol> for JsValue {
     #[inline]
     fn from(value: JsSymbol) -> Self {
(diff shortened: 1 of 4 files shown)


core/string/src/str.rs
```diff
@@ -291,6 +291,16 @@ impl PartialEq for JsStr<'_> {
...
+impl PartialEq<&str> for JsStr<'_> {
+    #[inline]
+    fn eq(&self, other: &&str) -> bool {
+        match self.variant() {
+            JsStrVariant::Latin1(v) => v == other.as_bytes(),
+            JsStrVariant::Utf16(v) => other.encode_utf16().zip(v).all(|(a, b)| a == *b),
+        }
+    }
+}
+
```
jedel1043, line 303: Can you check if `PartialEq<str>` is also implemented, and add it if it's not? Most std collections implement for both `[T]` and `&[T]` just in case someone uses `PartialEq::eq` directly.

core/engine/src/value/conversions/mod.rs
```diff
@@ -39,6 +39,24 @@ impl From<char> for JsValue {
...
+
+impl From<String> for JsValue {
+    #[inline]
+    fn from(value: String) -> Self {
+        let _timer = Profiler::global().start_event("From<&str>", "value");
+
+        Self::from(JsString::from(value))
+    }
+}
+
```
jedel1043, line 59: Uhh I'm slightly worried about this expression. If we add it, then all calls to `js_error` will fallback to runtime conversion to `UTF-16`, instead of using the nifty `js_str` which automatically converts to UTF-16 at compile time.

jedel1043: That was one of the reasons of removing `From<&str> for JsValue`

core/string/src/str.rs
hansl: Done

core/engine/src/value/conversions/mod.rs
hansl: I'm not sure there's another way? Outside of.. not implementing it I guess.

jedel1043 (approved):
Looks perfect! Thank you!

nekevss (approved):
Looks good to me :)

PR #3921, opened 2024-07-16
Add a JsError::from_rust constructor to create native errors from Rust

hansl:


raskad:
I like the addition. What do you think about the name `from_rust` instead of `from_std`? I think `Error` is moving to `core` so eventually we could switch that type over without changing the name.

hansl:
@raskad I made the change, PTAL.

raskad (approved):

jedel1043 (approved):

PR #3926, opened 2024-07-19
Fix wrong `neg` operation

CrazyboyQCD:
Negative `boolean` and `null` should return negative value


jedel1043 (commented):
Good catch! Can you add a test for it? Apparently test262 didn't have a test for this.

CrazyboyQCD:
> Good catch! Can you add a test for it? Apparently test262 didn't have a test for this.

Done.

jedel1043 (approved):
Thank you!

HalidOdat (approved):
Nice catch!

PR #3930, opened 2024-07-20
Fix temporal builtin properties

nekevss:
This PR fixes some the properties on the Temporal builtins. Primarily, it splits the builtin name and tag into 2 statics, and adjust the length property.


jedel1043 (approved):
Good catch!

raskad (approved):

PR #3932, opened 2024-07-22
A simple module loader from a function

hansl:
This will be the foundation for having a combinatoric module loader system.


core/interop/src/loaders/functions.rs
```diff
@@ -0,0 +1,63 @@
...
+/// custom [`ModuleLoader`].
+///
+/// This function cannot be `async` and must be blocking. An `async` version of
+/// this code will likely exist as a separate function in the future.
+///
+/// `F` cannot be a mutable closure as it could recursively call itself.
+#[derive(Copy, Clone)]
+pub struct FnModuleLoader<F>(F, &'static str)
+where
+    F: Fn(&JsString) -> Option<Module>;
```
jedel1043, line 16: Was thinking that maybe this should return `JsResult<Module>`. We could also have a separate loaded (`FallibleFnModuleLoader` or something), but it seems better to just have this be the fallible version already.

core/interop/src/loaders.rs
```diff
@@ -1,7 +1,31 @@
...
+    R: boa_engine::module::ModuleLoader + Clone + 'static,
+{
+    FallbackModuleLoader::new(loader, fallback)
+}
+
+/// Create a [`FsModuleLoader`] from a root path. This is a simple utility function.
+#[must_use]
+pub fn filesystem(root: std::path::PathBuf) -> FsModuleLoader {
+    FsModuleLoader::new(root)
+}
```
jedel1043, line 31: Do we need these? Seems a bit unnecessary knowing the loaders have constructors already.

core/interop/src/loaders/functions.rs
hansl: I'm thinking here that this is the simplest way to declare a module loader, without even worry about the error type and message; either the module exists or it doesn't. Even if it doesn't parse, or any other kind of error, just say it cannot be found.

I could see a better `Option<Source>` though, and show errors on parsing after the function returns the source of the module. WDYT?

core/interop/src/loaders.rs
hansl: Yeah they're just for convenience. I'll remove them.

core/interop/src/loaders/functions.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// Create a [`ModuleLoader`] from a function that takes a referrer and a path,
+/// and returns a [Module] if it exists, or an error.
+///
+/// This function cannot be `async` and must be blocking. An `async` version of
+/// this code will likely exist as a separate function in the future.
+///
+/// `F` cannot be a mutable closure as it could recursively call itself.
+#[derive(Copy, Clone)]
+pub struct FnModuleLoader<F>(F, &'static str)
```
HalidOdat, line 14: I think it would be clearer if we used named field, especially for structs with more than one field.

HalidOdat (approved):
Just a minor nitpick, besides that I looks good to me! :)

jedel1043 (approved):
Looks great! Hopefully in the future we'll be able to remove the `Clone` requirement on many of the bounds, it just requires some design work around `ModuleLoader`.

raskad (approved):

PR #3934, opened 2024-07-23
Fix destructuring assignment evaluation order

raskad:

- Fix evaluation of computed property names in declaration patterns
- Fix invalid value getter in iterator opcode
- Fix invalid setting of function binding identifier in `global_declaration_instantiation`


HalidOdat (approved):
Nice catch!

nekevss (approved):

PR #3935, opened 2024-07-23
Refactor `RawJsString`'s representation to make `JsString`s construction from string literal heap-allocation free

CrazyboyQCD:
This changes `RawJsString`'s representation, `RawJsString` in `JsString` is now a memory variant of `RawJsString` and `StaticJsString` which **holds** a string literal. It has a few benefits we can gain:
- Memory-safe (passes `miri`, thanks to @jedel1043 )
- `JsString` could be constant now
- No allocation when `JsString` is created from string literal by using `js_string` macro


nekevss (commented):
This is looking fantastic overall! Thanks for all the effort you put into this!

I still want to play around with it a bit more in depth, but thought I'd leave a couple comments.

core/string/src/lib.rs
```diff
@@ -221,6 +268,18 @@ impl<'a> IntoIterator for &'a JsString {
 }
 
 impl JsString {
+    /// Create a [`JsString`] from a static js string.
+    #[must_use]
+    pub const fn from_static_js_string(src: &'static StaticJsString) -> Self {
+        let src = ptr::from_ref(src).cast::<RawJsString>();
+        JsString {
+            // SAFETY:
```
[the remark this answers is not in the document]
jedel1043: As I see it `StaticJsString` is only constructable from a `JsStr<'static>` using `StaticJsString::new`, meaning it can never point to an invalid string. Hence `&'static StaticJsString` must be valid, which makes the call sound.

core/string/src/tests.rs
```diff
@@ -173,3 +173,28 @@ fn conversion_to_known_static_js_string() {
     assert!(string.is_some());
     assert!(string.unwrap().as_str().is_latin1());
 }
+
+#[test]
+fn from_static_js_string() {
```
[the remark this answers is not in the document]
CrazyboyQCD: > nit: can we have a few more test cases added separate from this.
> 
> I think overall 2 test cases were added: this one and the adjust refcount test. But due to the amount of unsafe code, it'd be nice to have a few more tests.

@nekevss Done.

core/string/src/lib.rs
```diff
@@ -149,37 +150,83 @@ impl CodePoint {
...
-/// The raw representation of a [`JsString`] in the heap.
-#[repr(C)]
-struct RawJsString {
-    /// Contains the flags and Latin1/UTF-16 length.
-    ///
-    /// The latin1 flag is stored in the bottom bit.
-    flags_and_len: usize,
+/// Contains the flags and Latin1/UTF-16 length.
+///
+/// The latin1 flag is stored in the bottom bit.
```
jedel1043, line 155: Nit: I would change the first line to a more "descriptive" style. Something like "A length with a tag on its bottom bit". The second line looks good.

core/string/src/lib.rs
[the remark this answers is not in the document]
jedel1043: Still needs resolution. Describing better the guarantees that `StaticJsString` gives for the safety comment should be enough.

core/string/src/lib.rs
```diff
@@ -245,24 +304,35 @@ impl JsString {
...
                 //   slice (length and starting pointer).
                 //
-                // - We aligned `h.data` on allocation, and the block is of size `h.len`, so this
+                // - We aligned `h.data()` on allocation, and the block is of size `h.len`, so this
                 //   should only generate valid reads.
                 //
                 // - The lifetime of `&Self::Target` is shorter than the lifetime of `self`, as seen
                 //   by its signature, so this doesn't outlive `self`.
+                //
+                // - The `RawJsString` created from string literal has a static lifetime `JsStr`.
```
jedel1043, line 313: Not sure what you mean by "static lifetime `JsStr`? If you mean `RawJsString` has a longer lifetime than `JsStr`, then you can put something like "A `RawJsString` created from a string literal has a static lifetime, making it always valid to reference using a `JsStr`"

core/string/src/lib.rs
```diff
@@ -871,13 +926,23 @@ impl Clone for JsString {
     #[inline]
     fn clone(&self) -> Self {
         if let UnwrappedTagged::Ptr(inner) = self.ptr.unwrap() {
-            // SAFETY: The reference count of `JsString` guarantees that `raw` is always valid.
+            // SAFETY: `NonNull` and the constructions of `JsString` guarantees that `raw` is always valid.
```
jedel1043, line 929: You also need to explain why it is safe to access the union as a `read_only`; a `Cell` is readable as an `usize` as long as we don't try to mutate the pointed variable.

core/string/src/lib.rs
```diff
@@ -896,13 +961,24 @@ impl Drop for JsString {
         if let UnwrappedTagged::Ptr(raw) = self.ptr.unwrap() {
             // See https://doc.rust-lang.org/src/alloc/sync.rs.html#1672 for details.
 
-            // SAFETY: The reference count of `JsString` guarantees that `raw` is always valid.
-            let inner = unsafe { raw.as_ref() };
-            inner.refcount.set(inner.refcount.get() - 1);
-            if inner.refcount.get() != 0 {
+            // SAFETY: `NonNull` and the constructions of `JsString` guarantees that `raw` is always valid.
```
jedel1043, line 964: Same as above.

jedel1043 (commented):
Implementation wise looks perfect! I just have some suggestions around some safety comments that could be improved.

CrazyboyQCD:
@jedel1043 , I've improved the comment, if you don't think it's good enough, please continue to point it out.

nekevss (approved):
Overall, this PR looks fantastic. Great work on this!

I saw the test added, but I still find myself wishing there were more tests. I'm not entirely sure that's not necessary to merge, imo. Maybe it would be a good follow up or first issue for someone? The docs improvements are also really nice as well.

As a separate note, @jedel1043 this may be a good candidate to build out design docs for per #3938.

jedel1043 (approved):
Looks perfect to me!

@nekevss Yeah, I'll open an issue for general improvements around documenting the current design of `JsString`.

PR #3937, opened 2024-07-27
Add more utility functions around modules and exports

hansl:
This allows to create a moderately simple example (see `tests/gcd.rs`) that can be used as a tutorial.

There are some typing issues, notably the `name` argument in `get_typed_fn`. Let's discuss.

### DIFF
--- a/core/engine/src/module/mod.rs
+++ b/core/engine/src/module/mod.rs
@@ -628,7 +628,6 @@ impl Module {
 
     /// Get an exported value from the module.
     #[inline]
-    #[must_use]
     pub fn get_value<K>(&self, name: K, context: &mut Context) -> JsResult<JsValue>
     where
         K: Into<PropertyKey>,
--- a/core/engine/src/module/mod.rs
+++ b/core/engine/src/module/mod.rs
@@ -30,6 +30,7 @@ use std::rc::Rc;
 use rustc_hash::FxHashSet;
 
 use boa_engine::js_string;
+use boa_engine::property::PropertyKey;
 use boa_gc::{Finalize, Gc, GcRefCell, Trace};
 use boa_interner::Interner;
 use boa_parser::source::ReadChar;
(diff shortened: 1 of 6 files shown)


core/engine/src/module/mod.rs
```diff
@@ -566,6 +570,20 @@ impl Module {
         promise
     }
 
+    /// Utility function that execute [`load_link_evaluate`] and then run the job queue.
+    /// This combines two operations that are often done together into a simpler
+    /// version. If you need more control over the execution, use [`load_link_evaluate`]
+    /// and [`Context::run_jobs`] separately.
+    pub fn load_link_evaluate_and_run_jobs(&self, context: &mut Context) -> JsResult<()> {
```
jedel1043, line 577: Maybe it's time to find a better name for the original function. `load_link_evaluate` sounds a bit spec-y and I think something like `run` would also be fine. That would also make the name of this function a lot shorter; `run_blocking` or something like that.

hansl: `eval_blocking`? Either way I can make the change in this PR.

jedel1043: Though about something. I think people would also want to await other promises, so maybe it should be better to add an `await_` method to `JsPromise`? That sidesteps the need for a new method for the module, and you would just call

`module.load_link_resolve(context).await_(context)?`

hansl: PTAL

core/engine/src/object/builtins/jsfunction.rs
```diff
@@ -1,11 +1,74 @@
...
 use std::ops::Deref;
 
+/// A trait for converting a tuple of Rust values into a vector of `JsValue`,
+/// to be used as arguments for a JavaScript function.
+pub trait TryIntoJsArguments {
+    /// Convert a tuple of Rust values into a vector of `JsValue`.
+    /// This is automatically implemented for tuples that implement
+    /// `TryIntoJsResult`.
+    fn into_js_args(self, cx: &mut Context) -> JsResult<Vec<JsValue>>;
+}
```
jedel1043, line 17: Praise: Loving this new API! Should help for strictness in many ways.

jedel1043 (approved):
Looks perfect to me!

core/engine/src/object/builtins/jsfunction.rs
```diff
@@ -1,11 +1,74 @@
...
+/// but will not be verified at runtime (since JavaScript doesn't
+/// actually have strong typing).
+///
+/// To create this type, use the [`JsFunction::typed`] method.
+#[derive(Debug, Clone, Trace, Finalize)]
+pub struct TypedJsFunction<A: TryIntoJsArguments, R: TryFromJs> {
+    inner: JsFunction,
+    _args: PhantomData<A>,
+    _ret: PhantomData<R>,
+}
```
jedel1043, line 48: Loving this new API! It should help for strictness in many ways.

HalidOdat (approved):
Looks great! :)

PR #3941, opened 2024-08-01
Add a Source::with_path method to set the path on a Source

hansl:
This should allow any path to be added to a Source, even when using `from_bytes` or `from_reader`.

### DIFF
--- a/core/parser/src/source/mod.rs
+++ b/core/parser/src/source/mod.rs
@@ -121,10 +121,10 @@ impl<'path, R: Read> Source<'path, UTF8Input<R>> {
 
 impl<'path, R> Source<'path, R> {
     /// Add a path to the current [`Source`] instance.
-    pub fn with_path<'path2>(self, new_path: &'path2 Path) -> Self<'path2, R> {
-        Self {
+    pub fn with_path(self, new_path: &Path) -> Source<'_, R> {
+        Source {
             reader: self.reader,
-            path: new_path,
+            path: Some(new_path),
         }
     }
 
--- a/core/parser/src/source/mod.rs
+++ b/core/parser/src/source/mod.rs
@@ -120,6 +120,14 @@ impl<'path, R: Read> Source<'path, UTF8Input<R>> {
 }
 
 impl<'path, R> Source<'path, R> {
+    /// Add a path to the current [`Source`] instance.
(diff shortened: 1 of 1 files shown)


core/parser/src/source/mod.rs
```diff
@@ -120,6 +120,14 @@ impl<'path, R: Read> Source<'path, UTF8Input<R>> {
 }
 
 impl<'path, R> Source<'path, R> {
+    /// Add a path to the current [`Source`] instance.
```
nekevss, line 123: nit: doc might be a bit unclear wording wise. As it's not adding the path to the current `Source`, but is creating a new `Source` from the combination.

```suggestion
    /// Returns a new source from the current [`Source`] instance combined with the provided `Path`.
```

hansl: It takes ownership of the source, so it's used like any `with_` method. I can use the same wording as `JsError::with_message`; `Sets the path of this Source`.

nekevss: Ah, I've been dealing with different `with_` methods recently with temporal 😆 I guess I'm a bit curious about taking `self` vs. `mut self` in this instance.

Basically, if I'm reading this correctly, this method could probably be:

```rust

// ... impl block

    pub fn with_path(mut self, new_path: &'path Path) -> Self {
        let _ = self.path.insert(new_path);
        self
    }

// cont ...

```

I think the above tends to be more in line with convention, and we also aren't potentially dropping the previous `self`, granted maybe rustc optimizes that out and it doesn't matter that much.

jedel1043: @nekevss Uhh I think your method doesn't work in general because it doesn't change the lifetime of `Source` to be the same as the lifetime of `Path`.

jedel1043 (approved):
Looks good!

core/parser/src/source/mod.rs
nekevss: True. I think I'm getting caught up on a bit of a bikeshed around the name/behavior of this method.

:confused: if we are destroying `Source` anyways, then wouldn't it be a bit clearer convention wise for the method to be a `to_` (something like `to_source_with_path`)

jedel1043: I think using `with_*` methods that take `self` is pretty common e.g. builder types. I don't see a problem with using the same convention for this.

nekevss (approved):

PR #3942, opened 2024-08-03
Implement register allocation

HalidOdat:
As discussed in #3194 this PR extracts the register allocation logic from #3798 which will be used by both PRs.

### DIFF
--- a/core/engine/src/bytecompiler/register.rs
+++ b/core/engine/src/bytecompiler/register.rs
@@ -48,12 +48,19 @@ impl Register {
 }
 
 impl Drop for Register {
+    /// This method should never be called.
+    /// It is used to detect when a register has not been deallocated.
     fn drop(&mut self) {
         if self.flags.is_persistent() {
             return;
         }
 
-        unreachable!("forgot to deallocate a register! Or a panic happend which caused Reg's drop to be called after the panic!")
+        // Prevent double panic.
+        if std::thread::panicking() {
+            return;
+        }
+
+        unreachable!("forgot to deallocate a register!")
     }
 }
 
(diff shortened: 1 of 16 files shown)


core/engine/src/bytecompiler/register.rs
```diff
@@ -0,0 +1,131 @@
...
+pub(crate) struct RegisterAllocator {
+    registers: Vec<RegisterEntry>,
+}
+
+impl RegisterAllocator {
+    pub(crate) fn alloc(&mut self) -> Register {
+        for (i, register) in self.registers.iter_mut().enumerate() {
+            if register.flags.is_used() {
+                continue;
+            }
```
jedel1043, line 70: Filter for all unused registers instead?

core/engine/src/bytecompiler/register.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl Drop for Register {
+    fn drop(&mut self) {
+        if self.flags.is_persistent() {
+            return;
+        }
+
+        unreachable!("forgot to deallocate a register! Or a panic happend which caused Reg's drop to be called after the panic!")
+    }
+}
```
jedel1043, line 58: Seeing this, I wonder if a better approach would be to use drop guards to automatically deallocate the registers, instead of having to mentally juggle when a register is still allocated or not.
You would have to add some lifetimes though, so that the guards are always dropped when you call `finish`.

jedel1043 (commented):
I have some suggestions that could maybe improve the ergonomics of the register allocator API.

core/engine/src/vm/opcode/rest_parameter/mod.rs
```diff
@@ -20,16 +20,30 @@ impl Operation for RestParameterInit {
...
-            Array::create_array_from_list(args, context)
+
+            let len = context.vm.stack.len() as u32;
+            let start = (len - rest_count - register_count) as usize;
+            let end = (len - register_count) as usize;
+
+            let args = &context.vm.stack[start..end];
+
+            let array = Array::create_array_from_list(args.iter().cloned(), context);
+            context.vm.stack.splice(start..end, [array.clone().into()]);
```
raskad, line 35: I think this should be `drain`, correct?

```suggestion
            context.vm.stack.drain(start..end);
```

raskad (changes requested):
Just got one thing that caused panics.

raskad (approved):

jedel1043 (approved):
Looks great!

PR #3943, opened 2024-08-03
Allow a custom Logger to be used as the backend for boa_runtime::Console

hansl:
There are some cleanup that could be done in the number of lines of codes, but this was the smallest change I could do to make the change.

This comes from a need to implement `Console` with using `tracing` as the backend. Please note that there is an issue here where multiple logs in the same function (e.g. `console.trace()`) would output into multiple log entries. This should probably be fixed in a following PR (as this one is already large enough).

### DIFF
--- a/core/runtime/src/lib.rs
+++ b/core/runtime/src/lib.rs
@@ -54,10 +54,7 @@
 mod console;
 
 #[doc(inline)]
-pub use console::Console;
-
-#[doc(inline)]
-pub use console::Logger;
+pub use console::{Console, Logger};
 
 #[cfg(test)]
 pub(crate) mod test {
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Looks perfect to me!

core/runtime/src/lib.rs
```diff
@@ -56,6 +56,9 @@ mod console;
 #[doc(inline)]
 pub use console::Console;
 
+#[doc(inline)]
+pub use console::Logger;
```
jasonwilliams, line 60: Can you not do `pub use console::{Console, Logger}` here?

jasonwilliams (approved):
LGTM (apart from comment), thanks @hansl

hansl:
I made the change, PTAL. Sorry I asked for a re-review from @jedel1043 by mistake.

jedel1043 (approved):

PR #3945, opened 2024-08-10
lazy builtins

jasonwilliams:
I just did this to experiment if lazy loading builtins would work and tryed it out with Array for now.

Using the js input

```js
console.log(Array);
```

Before:
image

After:
image

@jedel1043 do you have any pointers on how is best to structure this, and is the `GetName` operation `BindingLocatorEnvironment::GlobalObject` going to be a safe assumption for global builtin access?


jasonwilliams:
Closing in favour of https://github.com/boa-dev/boa/pull/3973

PR #3948, opened 2024-08-12
Split default icu data into lazily deserialized parts

jedel1043:
This should overall reduce the startup time spent on deserialization, even skipping it in some instances where the scripts only access one or two builtins.

@jasonwilliams Should be interesting to see how much this improved startup times.


jasonwilliams:
Thanks for looking at this, it doesn't look like its made any difference on the profiler unless im testing this wrong.
i've ran `cargo run --release --features boa_engine/profiler debug/script.js ` both off and on your branch. (not sure if any features need to be passed)

Before:
image 0.13ms

After:
image 0.12ms

jedel1043:
@jasonwilliams Ok, I converted a couple more utils to lazy loads, so it should be a bit faster.

jasonwilliams (approved):
Yep its basically 0 now

image

Code changes look good, I assume there's no size concerns as we would have been pulling those binaries in anyway?

nekevss (approved):
LGTM 😄

jedel1043:
> Code changes look good, I assume there's no size concerns as we would have been pulling those binaries in anyway?

Original singleton file:
```bash
❯ du -sbh -B1 core/icu_provider/data/icudata.postcard
16164195        core/icu_provider/data/icudata.postcard
```

New files:
```bash
❯ du -sbh -B1 core/icu_provider/data
16152575        core/icu_provider/data
```
Surprisingly, it's less expensive in size to split the icu data into several files.

PR #3951, opened 2024-08-14
Add a display type for JsString to allow formatting without allocations

hansl:
Splitting this into its own PR as the other PR was getting too large and opinionated. This should be fairly straightforward.


core/string/src/display.rs
```diff
@@ -0,0 +1,25 @@
...
+/// Display implementation for `JsString` that escapes unicode characters.
+#[derive(Debug)]
+pub struct JsStringDisplayEscaped<'a> {
+    pub(crate) inner: JsStr<'a>,
+}
+
+impl fmt::Display for JsStringDisplayEscaped<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.inner.variant() {
+            // SAFETY: `JsStrVariant::Latin1` is always valid latin1, so no need to check.
```
jasonwilliams, line 15: Did you mean to say "latin1 is always valid utf8" here?

jasonwilliams (approved):

core/string/src/display.rs
hansl: Fixed. PTAL.

HalidOdat (approved):
Nice addition, looks good to me! :)

core/string/src/display.rs
```diff
@@ -0,0 +1,25 @@
...
+#[derive(Debug)]
+pub struct JsStringDisplayEscaped<'a> {
+    pub(crate) inner: JsStr<'a>,
+}
+
+impl fmt::Display for JsStringDisplayEscaped<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.inner.variant() {
+            // SAFETY: `JsStrVariant::Latin1` is always valid utf8, so no need to check.
+            JsStrVariant::Latin1(v) => unsafe { std::str::from_utf8_unchecked(v) }.fmt(f),
```
HalidOdat, line 16: Actually, this is not correct! ascii (uses 7-bits) is always valid utf8, but not latin1 (uses full 8-bits).

Instead the implementation could use the `iter` method and would eliminate the match as well.

HalidOdat (changes requested):

core/string/src/display.rs
hansl: I used the variant name for a reason; `JsStrVariant::Latin1` is improper naming but should be its own fix and not part of this PR.

Actually double checked the usage of Latin1 in JsString and co, and I think it's potentially unsafe. I'll file a bug tomorrow.

> Instead the implementation could use the code_points method and would eliminate the match as well.

hansl: I did a thing; https://github.com/boa-dev/boa/pull/3958. I can fix it but it's not in scope for this PR.

core/string/src/display.rs
```diff
@@ -0,0 +1,45 @@
...
+            JsStrVariant::Latin1(v) => v
+                .iter()
+                .copied()
+                .map(char::from)
+                .try_for_each(|c| f.write_char(c)),
+            JsStrVariant::Utf16(_) => self.inner.code_points().try_for_each(|r| match r {
+                CodePoint::Unicode(c) => f.write_char(c),
+                CodePoint::UnpairedSurrogate(c) => {
+                    f.write_str("\\u")?;
+                    write!(f, "{c:04X}")
```
HalidOdat, line 25:
```suggestion
                    write!(f, "\\u{c:04X}")
```

Is this an optimization? I think we can just inline the two calls.

core/string/src/display.rs
```diff
@@ -0,0 +1,45 @@
...
+/// Display implementation for [`crate::JsString`] that escapes unicode characters.
+#[derive(Debug)]
+pub struct JsStrDisplayEscaped<'a> {
+    inner: JsStr<'a>,
+}
+
+impl fmt::Display for JsStrDisplayEscaped<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.inner.variant() {
+            // SAFETY: `JsStrVariant::Latin1` is always valid utf8, so no need to check.
```
HalidOdat, line 15:
```suggestion
            // SAFETY: `JsStrVariant::Latin1` does not contain any unpaired surrogates, so need to check.
```

core/string/src/lib.rs
```diff
@@ -150,6 +153,18 @@ impl CodePoint {
     }
 }
 
+impl std::fmt::Display for CodePoint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CodePoint::Unicode(c) => f.write_char(*c),
+            CodePoint::UnpairedSurrogate(c) => {
+                f.write_str("\\u")?;
+                write!(f, "{c:04X}")
```
HalidOdat, line 162:
```suggestion
                write!(f, "\\u{c:04X}")
```

HalidOdat (approved):
Besides the minor nitpicks this looks good to me! :)

HalidOdat (approved):
Looks perfect to me! :)

PR #3955, opened 2024-08-16
Build out partial record functionality, property bag construction, and `with` methods

nekevss:
This PR is related to the ongoing work for temporal, specifically in relation to the `with` methods and partial records.

It changes the following:

- Implements `map` on `JsValue`
- Implements `to_partial_date_record` and `to_partial_time_record`
- Builds out the `with` methods on `Time` and `Date` along with the property bags portions of `to_date` and `to_time` abstract methods

I totally blanked on making the fields of `PartialDateTime` public :sweat_smile: So that will hopefully still be coming after a patch I submitted is merged.

### DIFF
--- a/core/engine/src/builtins/temporal/mod.rs
+++ b/core/engine/src/builtins/temporal/mod.rs
@@ -279,16 +279,15 @@ pub(crate) fn is_partial_temporal_object<'value>(
     value: &'value JsValue,
     context: &mut Context,
 ) -> JsResult<Option<&'value JsObject>> {
-    // 2. If value has an [[InitializedTemporalDate]], [[InitializedTemporalDateTime]],
-    // [[InitializedTemporalMonthDay]], [[InitializedTemporalTime]],
-    // [[InitializedTemporalYearMonth]], or
-    // [[InitializedTemporalZonedDateTime]] internal slot, return false.
-
     // 1. If value is not an Object, return false.
     let Some(obj) = value.as_object() else {
         return Ok(None);
     };
 
+    // 2. If value has an [[InitializedTemporalDate]], [[InitializedTemporalDateTime]],
+    // [[InitializedTemporalMonthDay]], [[InitializedTemporalTime]],
+    // [[InitializedTemporalYearMonth]], or
+    // [[InitializedTemporalZonedDateTime]] internal slot, return false.
     if obj.is::<PlainDate>()
         || obj.is::<PlainDateTime>()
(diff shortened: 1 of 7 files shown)


core/engine/src/builtins/temporal/mod.rs
```diff
@@ -274,6 +274,47 @@ pub(crate) fn to_relative_temporal_object(
...
 
+// 13.26 IsPartialTemporalObject ( object )
+pub(crate) fn is_partial_temporal_object<'value>(
+    value: &'value JsValue,
+    context: &mut Context,
+) -> JsResult<Option<&'value JsObject>> {
+    // 2. If value has an [[InitializedTemporalDate]], [[InitializedTemporalDateTime]],
+    // [[InitializedTemporalMonthDay]], [[InitializedTemporalTime]],
+    // [[InitializedTemporalYearMonth]], or
+    // [[InitializedTemporalZonedDateTime]] internal slot, return false.
```
HalidOdat, line 285: This should be move after step one when we use the `.is<T>` method, right?

core/engine/src/builtins/temporal/plain_date/mod.rs
```diff
@@ -613,10 +618,39 @@ impl PlainDate {
...
+                .into());
+        };
+        let options = get_options_object(args.get_or_undefined(1))?;
+
+        // 4. Let resolvedOptions be ? SnapshotOwnProperties(? GetOptionsObject(options), null).
+        // 5. Let calendarRec be ? CreateCalendarMethodsRecord(temporalDate.[[Calendar]], « date-from-fields, fields, merge-fields »).
+        // 6. Let fieldsResult be ? PrepareCalendarFieldsAndFieldNames(calendarRec, temporalDate, « "day", "month", "monthCode", "year" »).
+        // 7. Let partialDate be ? PrepareTemporalFields(temporalDateLike, fieldsResult.[[FieldNames]], partial).
+        // 8. Let fields be ? CalendarMergeFields(calendarRec, fieldsResult.[[Fields]], partialDate).
+        // 9. Set fields to ? PrepareTemporalFields(fields, fieldsResult.[[FieldNames]], «»).
```
HalidOdat, line 647: Are there steps skipped, if so we may want to marked them with something like `SKIP: ` (like we do in many places).

HalidOdat (approved):
Nice work!

Just some very minor nitpicks :)

raskad (approved):

PR #3958, opened 2024-08-19
Latin1 is the wrong optimization

hansl:
This is a PR showing a new test that breaks Latin1. The first `assert!` will pass, but the `assert_eq!` will fail.

The right optimization would be to use `u8::MAX >> 1` and make it ASCII only.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,6 +489,7 @@ dependencies = [
 name = "boa_macros"
 version = "0.19.0"
 dependencies = [
+ "boa_engine",
  "proc-macro2",
  "quote",
  "syn",
--- a/core/macros/Cargo.toml
+++ b/core/macros/Cargo.toml
@@ -17,6 +17,9 @@ syn = { workspace = true, features = ["full"] }
 proc-macro2.workspace = true
 synstructure.workspace = true
 
+[dev-dependencies]
+boa_engine.workspace = true
+
 [lints]
 workspace = true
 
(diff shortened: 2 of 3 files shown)


hansl:
The error, for reference;

CleanShot 2024-08-18 at 19 12 09@2x

hansl:
CleanShot 2024-08-18 at 19 12 55@2x

jedel1043:
Looking at the test, I figured out that this is a bug in the `js_str` macro. After checking that the string literal is latin1, we use the utf8 as latin1:
```rust
if is_latin1 {
    quote! {
        ::boa_engine::string::JsStr::latin1(#utf8.as_bytes())
    }
}
```
But `utf8` is not `latin1` in the C0 - FF range, so `é` gets misinterpreted as two bytes instead of one.

I have a fix for this in #3959, so you can rebase your PR against it to add your new test.

hansl:
> But utf8 is not latin1 in the C0 - FF range

@jedel1043  Correct me if I'm wrong, but utf8 is not anything in the 80-FF range. `é` is 0x82. We should really be using `utf8` instead of `latin1`.

jedel1043:
If you're still getting lint errors, just move the test to `tests/macros/tests`. Clippy gets a bit weird when you're dealing with cyclic dependencies 😅

hansl:
It's fine, lol. I'm not planning (yet) to merge this. It's for starting the conversation.

PR #3959, opened 2024-08-19
Fix `js_str` macro to correctly handle latin1 strings

jedel1043:
Related to #3958.


HalidOdat:
I was checking it out and came to the same fix, but you beat me to it :laughing:

HalidOdat (approved):
Maybe we should also include the test that @hansl had in #3958, besides that this looks perfect to me! :)

jasonwilliams (approved):

PR #3961, opened 2024-08-19
use with_capacity to reduce re-allocations fixes #3896

### LINKED ISSUE #3896: Boa's start-up time getting hit with GC Allocation

jasonwilliams:
This PR attempts to address https://github.com/boa-dev/boa/issues/3896

We already raised the threshold, but it would be good to not re-allocate the hashmaps and vectors as we will up the property storage during startup. 

It doesn't seem to have much noticeable affect in the performance profiler but DHAT shows less `grow_one` calls from `boa_engine::object::shape::property_table::PropertyTableInner::insert (object/shape/property_table.rs:70:9)`, there are still some, we may be able to find these by checking `.len() == .capacity()` and tracing back where these are coming from

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -105,7 +105,7 @@ impl IntrinsicObject for Array {
 
         let unscopables_object = Self::unscopables_object();
 
-        BuiltInBuilder::from_standard_constructor::<Self, 41, 5>(realm)
+        BuiltInBuilder::from_standard_constructor::<Self>(realm)
             // Static Methods
             .static_method(Self::from, js_string!("from"), 1)
             .static_method(Self::is_array, js_string!("isArray"), 1)
@@ -190,6 +190,9 @@ impl BuiltInObject for Array {
 }
 
 impl BuiltInConstructor for Array {
+    const P: usize = 41;
+    const SP: usize = 5;
+
     const LENGTH: usize = 1;
 
     const STANDARD_CONSTRUCTOR: fn(&StandardConstructors) -> &StandardConstructor =
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -105,7 +105,7 @@ impl IntrinsicObject for Array {
 
         let unscopables_object = Self::unscopables_object();
 
-        BuiltInBuilder::from_standard_constructor::<Self>(realm)
+        BuiltInBuilder::from_standard_constructor::<Self>(realm, 41, 5)
             // Static Methods
             .static_method(Self::from, js_string!("from"), 1)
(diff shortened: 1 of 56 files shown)


core/engine/src/builtins/builder.rs
```diff
@@ -524,20 +524,25 @@ impl<'ctx> BuiltInBuilder<'ctx, OrdinaryObject> {
 }
 
 impl<'ctx> BuiltInBuilder<'ctx, Callable<Constructor>> {
+    /// Create a new builder for a constructor function setting the properties ahead of time for optimizations (less reallocations)
     pub(crate) fn from_standard_constructor<SC: BuiltInConstructor>(
```
nekevss, line 528: thought: is it possible to use const generics here if these values are known at compile time over a parameter?

```suggestion
    pub(crate) fn from_standard_constructor<SC: BuiltInConstructor, const M: usize, const SM: usize>(
```

HalidOdat: Hmmm... this may cause code bloat due to monomorphization.

nekevss: > Hmmm... this may cause code bloat due to monomorphization.

ooooh, that's true. They could be constants on the `BuiltInConstructor` trait instead, which may actually be nicer now that I think about it. The fields can still be left as constants, but that should remove any potential code bloat from monomorphization.

jasonwilliams: Yeah I was wondering this, if with const genetics rust will need to internally create a new concrete type for each instance. I can change to const

raskad (approved):
I'd be for merging this. I ran the benchmarks locally and there are no significant changes overall.

core/engine/src/builtins/array/mod.rs
```diff
@@ -193,6 +193,9 @@ impl BuiltInObject for Array {
 }
 
 impl BuiltInConstructor for Array {
+    const P: usize = 41;
+    const SP: usize = 5;
```
HalidOdat, line 197:
```suggestion
    const PPOPERTY_COUNT: usize = 41;
    const STATIC_PROPERTY_COUNT: usize = 5;
```

I think it would be better to add the full name instead of single letter.

HalidOdat (approved):
Besides the comment, this looks good to me! :)

PR #3966, opened 2024-08-20
Remove leftover eprintln from JsPromise::await_blocking

hansl:
We should make this a lint error.


hansl:
Actually did make it a lint error in #3967. Closing this as duplicate work.

PR #3967, opened 2024-08-21
Add clippy for denying print and eprints

hansl:
And remove usage of println or allow them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -221,8 +221,8 @@ bare_urls = "warn"
 [workspace.lints.clippy]
 # clippy allowed by default
 dbg_macro = "warn"
-print_stdout = "deny"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -226,6 +221,8 @@ bare_urls = "warn"
 [workspace.lints.clippy]
 # clippy allowed by default
 dbg_macro = "warn"
+print_stdout = "deny"
```
jedel1043, line 224: Nit: It's probably cleaner to just warn. We already have CI to deny warnings.

jedel1043 (approved):
Looks great! I just have a small nitpick.

jedel1043 (approved):

nekevss (approved):

PR #3968, opened 2024-08-21
Add a clone_inner method to allow cloning of inner data

hansl:
The issue I'm looking to solve is that I'm passing back the same object into JavaScript (think callbacks). The object is reentrant safe, but I'm ending up doing a lot of `this.borrow().clone()`. This would make it a bit easier to manage.

The JS code looks like that:

```js
const someValue = new ApiClass();
someValue.doSomethingAndCallback(() => {
  // Using a naive but straightforward borrow_mut(), Boa would fail here as
  // the `doSomethingElse` method would need to borrow_mut() again.
  someValue.doSomethingElse();
});
```

### DIFF
--- a/core/interop/src/lib.rs
+++ b/core/interop/src/lib.rs
@@ -394,7 +394,7 @@ impl<T: NativeObject + Clone> JsClass<T> {
     /// Clones the inner class instance.
     #[must_use]
     pub fn clone_inner(&self) -> T {
-        self.inner.data().clone()
+        self.inner.borrow().data().clone()
     }
 }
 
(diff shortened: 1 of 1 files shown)


core/interop/src/lib.rs
```diff
@@ -390,6 +390,14 @@ impl<T: NativeObject> JsClass<T> {
     }
 }
 
+impl<T: NativeObject + Clone> JsClass<T> {
+    /// Clones the inner class instance.
+    #[must_use]
+    pub fn clone_inner(&self) -> T {
+        self.inner.borrow().data().clone()
+    }
+}
```
jedel1043, line 399: Nit: this probably needs a `# Panics` comment specifying that it will panic if the inner data is already mutably borrowed.

jedel1043 (commented):
Looks great! I Just have a small doc nitpick.

jedel1043 (approved):

nekevss (approved):

PR #3970, opened 2024-08-26
Add support for Trace, Finalize and JsData for Convert<>

hansl:
These should have been there from the start.


nekevss (approved):

HalidOdat (approved):
Looks good to me! :)

PR #3971, opened 2024-08-26
Add a way for js_error! macro to create native errors with message

hansl:
I assume this will have a fair bit of bikeshedding on the interface.

This only supports text errors since it only adds the message (which is text only). Supports syntax similar to `format!()`, e.g. `js_error!(TypeError: "the number {} should be even", odd_number)`.

## Other Considerations and Explanation

I considered another option for the API: having a separate macro per type (like `js_typ_error!()`, `js_syntax_error!()`, etc). I decided to dismiss this and went the way of this PR instead as the code looks closer to what the message will be, since it's normally prefixed with the error. It also declutters the `use` statements as there's only one macro.

For example, the following code:

```rust
js_error!(TypeError: "not a day of the week")
```

would show the following message when printed on the console:

```text
TypeError: not a day of the week
```

Tested with boa_cli: 
CleanShot 2024-08-26 at 16 37 00@2x

### DIFF
--- a/core/engine/src/error.rs
+++ b/core/engine/src/error.rs
@@ -13,10 +13,27 @@ use boa_macros::js_str;
 use std::{error, fmt};
 use thiserror::Error;
 
-/// Create an opaque error object from a value or string literal.
+/// Create an error object from a value or string literal. Optionally the
+/// first argument of the macro can be a type of error (such as `TypeError`,
+/// `RangeError` or `InternalError`).
 ///
 /// Can be used with an expression that converts into `JsValue` or a format
 /// string with arguments.
+///
+/// # Native Errors
+///
+/// The only native error that is not buildable using this macro is
+/// [`crate::JsNativeErrorKind::AggregateError`], which requires multiple
+/// error objects available at construction.
+///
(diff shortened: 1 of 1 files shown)


nekevss:
Assuming I'm reading this right. I think my biggest concern is I'm not sure I'm thrilled about the case when args are needed for format.

Without format:
```rust
    let err = js_error!(TypeError: "{} was not a string", value.to_string());
```

With format:
```rust
    let err = js_error!(TypeError: format!("{} was not a string", value.to_string()));
```

But after looking at it, I don't like the version with format either. Overall, I think this looks pretty nice. Out of curiosity, was there any other formats that you had tried?

hansl:
I tried having one macro by type of native error. It wasn't saving that many stroke and rustfmt does the right thing here so I went for the simpler import.

raskad (approved):
I think the current interface is fine.
Personally, I'm not a huge fan of macros in general but if there is a macro to create an error, it should be only one instead of one per native type.

nekevss (approved):
I'm fine with moving forward with this as is.

PR #3973, opened 2024-08-29
lazy builtins

### LINKED ISSUE #2307: [performance] - Opportunity to run parsing before building environment

jasonwilliams:
## Lazy Builtins

So, this creates 2 new data structs which are used inside of JS Objects. `LazyBuiltIn` and `LazyPrototype`.
Instead of calling [initialize](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/mod.rs#L211) on array for example, we will setup array with a constructor that has a LazyBuiltin object and a prototype which lazy a LazyPrototype. If any properties or the constructor itself is called then we will call `init` at that point. 

For now I've started with:
- Array

I also updated the VSCode task to always choose the `debug/script.js` path, as it makes debugging much faster

Closes #2307

### DIFF
--- a/core/engine/src/builtins/builder.rs
+++ b/core/engine/src/builtins/builder.rs
@@ -3,7 +3,7 @@ use crate::{
     native_function::{NativeFunctionObject, NativeFunctionPointer},
     object::{
         shape::{property_table::PropertyTableInner, slot::SlotAttributes},
-        FunctionBinding, JsFunction, JsPrototype, CONSTRUCTOR, PROTOTYPE,
+        FunctionBinding, JsFunction, JsPrototype, LazyBuiltIn, CONSTRUCTOR, PROTOTYPE,
     },
     property::{Attribute, PropertyDescriptor, PropertyKey},
     realm::Realm,
@@ -392,12 +392,25 @@ impl BuiltInConstructorWithPrototype<'_> {
         }
 
         let mut object = self.object.borrow_mut();
-        let function = object
-            .downcast_mut::<NativeFunctionObject>()
-            .expect("Builtin must be a function object");
-        function.f = NativeFunction::from_fn_ptr(self.function);
-        function.constructor = Some(ConstructorKind::Base);
-        function.realm = Some(self.realm.clone());
+        if object.is::<NativeFunctionObject>() {
+            let function = object
+                .downcast_mut::<NativeFunctionObject>()
+                .expect("Builtin must be a function object");
+            function.f = NativeFunction::from_fn_ptr(self.function);
+            function.constructor = Some(ConstructorKind::Base);
+            function.realm = Some(self.realm.clone());
+        } else if object.is::<LazyBuiltIn>() {
+            let lazy = object
+                .downcast_mut::<LazyBuiltIn>()
+                .expect("Builtin must be a lazy object");
+            lazy.set_constructor(
+                NativeFunction::from_fn_ptr(self.function),
+                self.realm.clone(),
+            );
+        } else {
+            unreachable!("The object must be a function or a lazy object");
+        }
+
         object
             .properties_mut()
             .shape
--- a/core/engine/src/builtins/function/mod.rs
+++ b/core/engine/src/builtins/function/mod.rs
@@ -26,7 +26,7 @@ use crate::{
             get_prototype_from_constructor, CallValue, InternalObjectMethods,
             ORDINARY_INTERNAL_METHODS,
         },
-        JsData, JsFunction, JsObject, PrivateElement, PrivateName,
+        JsData, JsFunction, JsObject, LazyBuiltIn, PrivateElement, PrivateName,
     },
     property::{Attribute, PropertyDescriptor, PropertyKey},
     realm::Realm,
@@ -845,8 +845,7 @@ impl BuiltInFunctionObject {
             return Err(JsNativeError::typ().with_message("not a function").into());
         };
 
-        let object_borrow = object.borrow();
-        if object_borrow.is::<NativeFunctionObject>() {
+        if object.is::<NativeFunctionObject>() || object.is::<LazyBuiltIn>() {
             let name = {
                 // Is there a case here where if there is no name field on a value
                 // name should default to None? Do all functions have names set?
@@ -860,10 +859,11 @@ impl BuiltInFunctionObject {
             return Ok(
                 js_string!(js_str!("function "), &name, js_str!("() { [native code] }")).into(),
             );
-        } else if object_borrow.is::<Proxy>() || object_borrow.is::<BoundFunction>() {
+        } else if object.is::<Proxy>() || object.is::<BoundFunction>() {
             return Ok(js_string!("function () { [native code] }").into());
         }
 
+        let object_borrow = object.borrow();
         let function = object_borrow
             .downcast_ref::<OrdinaryFunction>()
             .ok_or_else(|| JsNativeError::typ().with_message("not a function"))?;
--- a/core/engine/src/native_function.rs
+++ b/core/engine/src/native_function.rs
@@ -343,6 +343,28 @@ pub(crate) fn native_function_call(
     argument_count: usize,
     context: &mut Context,
 ) -> JsResult<CallValue> {
+    native_function_call_inner(
+        obj,
+        &obj.downcast_ref::<NativeFunctionObject>()
+            .expect("the object should be a native function object")
+            .clone(),
+        argument_count,
+        context,
+    )
+}
+
+/// Call this object.
+///
+/// # Panics
+///
+/// Panics if the object is currently mutably borrowed.
+// <https://tc39.es/ecma262/#sec-built-in-function-objects-call-thisargument-argumentslist>
+pub(crate) fn native_function_call_inner(
+    obj: &JsObject,
+    native_function: &NativeFunctionObject,
+    argument_count: usize,
+    context: &mut Context,
+) -> JsResult<CallValue> {
     let args = context.vm.pop_n_values(argument_count);
     let _func = context.vm.pop();
     let this = context.vm.pop();
@@ -356,10 +378,7 @@ pub(crate) fn native_function_call(
         f: function,
         constructor,
         realm,
-    } = obj
-        .downcast_ref::<NativeFunctionObject>()
-        .expect("the object should be a native function object")
-        .clone();
+    } = native_function.clone();
 
     let mut realm = realm.unwrap_or_else(|| context.realm().clone());
 
@@ -380,7 +399,6 @@ pub(crate) fn native_function_call(
 
     Ok(CallValue::Complete)
 }
-
 /// Construct an instance of this object with the specified arguments.
 ///
 /// # Panics
--- a/core/engine/src/native_function.rs
+++ b/core/engine/src/native_function.rs
@@ -343,14 +343,14 @@ pub(crate) fn native_function_call(
     argument_count: usize,
     context: &mut Context,
 ) -> JsResult<CallValue> {
-    native_function_call_inner(
-        obj,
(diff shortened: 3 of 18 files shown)


core/engine/src/object/builtins/lazy_builtin.rs
```diff
@@ -0,0 +1,260 @@
...
+
+pub(crate) fn lazy_get_prototype_of(
+    obj: &JsObject,
+    context: &mut Context,
+) -> JsResult<JsPrototype> {
+    let realm = context.realm();
+    let builtin = obj.downcast_ref::<Builtin>().expect("obj is not a Builtin");
+    if builtin.is_initialized.get() {
+        (builtin.init)(realm);
+    }
```
jedel1043, line 75: Isn't this condition wrong? It should check if `is_initialized` is false, and both run init and set it to true if it enters the `if`

core/engine/src/object/builtins/lazy_builtin.rs
```diff
@@ -0,0 +1,260 @@
...
+) -> JsResult<CallValue> {
+    let realm = context.realm();
+    let builtin = obj.downcast_ref::<Builtin>().expect("obj is not a Builtin");
+    (builtin.init)(realm);
+
+    if let BuiltinKind::Constructor(constructor) = &builtin.kind {
+        return Ok(constructor.__call__(argument_count));
+    }
+
+    non_existant_call(obj, argument_count, context)
```
jedel1043, line 243: This should not be checked at runtime, but inside the `JsData` implementation. The reason is that setting an ordinary object's construct or call function to something different to `non_existant_data` will incorrectly determine that the object is callable.

core/engine/src/object/builtins/lazy_builtin.rs
```diff
@@ -0,0 +1,260 @@
...
+    __own_property_keys__: lazy_own_property_keys,
+    __call__: lazy_call,
+    __construct__: lazy_construct,
+};
+
+pub(crate) fn lazy_get_prototype_of(
+    obj: &JsObject,
+    context: &mut Context,
+) -> JsResult<JsPrototype> {
+    let realm = context.realm();
```
jedel1043, line 71: `realm` cannot be fetched here, because the initializer could be called from a different realm, which would initialize the incorrect intrinsic. We can use `Gc::new_cyclic` to store a weak reference to the `Realm` inside the builtin, then recover the strong reference inside here.

core/engine/src/object/builtins/lazy_builtin.rs
jasonwilliams: Yep the condition is wrong ive updated

core/engine/src/builtins/builder.rs
```diff
@@ -387,12 +387,6 @@ impl BuiltInConstructorWithPrototype<'_> {
         }
 
         let mut object = self.object.borrow_mut();
-        let function = object
-            .downcast_mut::<NativeFunctionObject>()
-            .expect("Builtin must be a function object");
-        function.f = NativeFunction::from_fn_ptr(self.function);
-        function.constructor = Some(ConstructorKind::Base);
-        function.realm = Some(self.realm.clone());
```
jedel1043, line 395: Oops, I suggested to remove this, but I think this is still needed.

core/engine/src/builtins/function/mod.rs
```diff
@@ -850,6 +850,13 @@ impl BuiltInFunctionObject {
             );
         } else if object_borrow.is::<Proxy>() || object_borrow.is::<BoundFunction>() {
             return Ok(js_string!("function () { [native code] }").into());
+        } else if object_borrow.is::<LazyBuiltIn>() {
+            let name = object_borrow
+                .downcast_ref::<LazyBuiltIn>()
+                .map_or_else(|| js_string!(), |built_in| built_in.name.clone());
+            return Ok(
+                js_string!(js_str!("function "), &name, js_str!("() { [native code] }")).into(),
+            );
```
jedel1043, line 859: This is not needed if you move the check for `LazyBuiltIn` to the same branch as the check for `NativeFunctionObject`. The reason is that we call `JsObject::get` inside it, which will already initialize the `name` of the builtin, printing the correct result.

core/engine/src/object/builtins/lazy_builtin.rs
```diff
@@ -0,0 +1,451 @@
...
+
+/// A builtin function. Used for lazy initialization of builtins.
+
+#[derive(Clone, Finalize)]
+pub struct LazyBuiltIn {
+    pub(crate) init: fn(&Realm),
+    pub(crate) is_initialized: Cell<bool>,
+    pub(crate) kind: BuiltinKind,
+    pub(crate) realm_inner: Option<WeakGc<RealmInner>>,
+    pub(crate) name: JsString,
```
jedel1043, line 36: We can put `init` and `realm_inner` into a single `Option` and remove `is_initialized`. That way, we can use `Option::take` (which will set the option to `None` on access) to determine if the builtin has already been initialized; if `Option::take` returns `Some`, initialize. Else, don't initialize.

core/engine/src/object/builtins/lazy_builtin.rs
```diff
@@ -0,0 +1,451 @@
...
+    realm::{Realm, RealmInner},
+    Context, JsData, JsNativeError, JsObject, JsResult, JsValue,
+};
+use boa_gc::{Finalize, Trace, WeakGc};
+use boa_string::JsString;
+use std::cell::Cell;
+
+#[derive(Debug, Clone, Trace, Finalize)]
+pub(crate) enum BuiltinKind {
+    Function(JsFunction),
```
jedel1043, line 24: Hmm, thinking a bit more about this, I think `Function` should store a `NativeFunctionObject` instead, and the builder methods should be adjusted to consider this. Else, the builder won't set the builtin to the correct value while initializing.

core/engine/src/object/builtins/lazy_builtin.rs
jasonwilliams: Done

core/engine/src/builtins/builder.rs
jasonwilliams: I've made some updates to this, take a look

core/engine/src/object/builtins/lazy_builtin.rs
```diff
@@ -0,0 +1,453 @@
...
+) -> JsResult<JsPrototype> {
+    let builtin: JsObject<LazyBuiltIn> = obj.clone().downcast().expect("obj is not a Builtin");
+    // If there is a value in here, we need to initialize the builtin
+    if builtin.borrow().data.init_and_realm.is_some() {
+        let mut builtin_borrow = builtin.borrow_mut();
+        let (init, realm_inner) = builtin_borrow
+            .data
+            .init_and_realm
+            .take()
+            .expect("init_and_realm not set");
```
jedel1043, line 101: With the new changes this can be improved. Instead of calling `init_and_realm.is_some()`, then `init_and_realm.take()`, we can do `if let Some(data) = builtin.borrow_mut().data.init_and_realm.take()`

jedel1043:
  
  | Total | 48,494 | 48,494 | 0 |
  | Passed | 43,609 | 43,581 | **-28** |
  | Ignored | 1,500 | 1,500 | 0 |
  | Failed | 3,385 | 3,413 | **+28** |
  | Panics | 0 | 2 | **+2** |
  | Conformance | 89.93% | 89.87% | **-0.06%** |
  
  <details><summary><b>Broken tests (26):</b></summary>
  
  ```
  test/language/statements/for-of/arguments-unmapped-mutation.js (previously Passed)
  test/language/statements/for-of/arguments-unmapped-aliasing.js (previously Passed)
  test/language/statements/for-of/arguments-mapped-aliasing.js (previously Passed)
  test/language/statements/for-of/arguments-mapped-mutation.js (previously Passed)
  test/language/statements/for-of/arguments-unmapped.js (previously Passed)
  test/language/statements/for-of/arguments-mapped.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-unmapped-truncation-before-exhaustion.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-mapped-iteration.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-unmapped-iteration.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-unmapped-expansion-before-exhaustion.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-mapped-expansion-after-exhaustion.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-unmapped-expansion-after-exhaustion.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-mapped-expansion-before-exhaustion.js (previously Passed)
  test/built-ins/ArrayIteratorPrototype/next/args-mapped-truncation-before-exhaustion.js (previously Passed)
  test/built-ins/String/prototype/S15.5.4_A2.js (previously Passed)
  test/built-ins/String/prototype/S15.5.4_A3.js (previously Passed)
  test/built-ins/String/prototype/S15.5.4_A1.js (previously Passed)
  test/built-ins/Array/isArray/15.4.3.2-0-5.js (previously Passed)
  test/built-ins/Array/prototype/exotic-array.js (previously Passed)
  test/built-ins/Array/prototype/filter/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/Array/prototype/concat/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/Array/prototype/slice/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/Array/prototype/map/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/Array/prototype/splice/create-proto-from-ctor-realm-array.js (previously Passed)
  test/built-ins/TypedArray/prototype/toString/BigInt/detached-buffer.js (previously Passed)
  test/built-ins/TypedArrayConstructors/ctors/no-species.js (previously Passed)
  ```
  </details>
  
  <details><summary><b>New panics (2):</b></summary>
  
  ```
  test/built-ins/Array/prototype/forEach/S15.4.4.18_A1.js (previously Passed)
  test/built-ins/Array/prototype/forEach/S15.4.4.18_A2.js (previously Passed)
  ```
  </details>

jedel1043:
Some tests regressed thanks to the changes. Is it potentially a typo/copypaste error, or do we need to go back to the drawing board?

hansl:
I'll hold any review until it's up to date with `main` and green.

core/engine/src/native_function.rs
```diff
@@ -391,20 +409,32 @@ fn native_function_construct(
...
+        &obj.downcast_ref::<NativeFunctionObject>()
+            .expect("the object should be a native function object")
+            .clone(),
+        obj.clone(),
+        argument_count,
+        context,
+    )
+}
+
+pub(crate) fn native_function_construct_inner(
```
jasonwilliams, line 423: So there's been some (minor) performance regressions with this PR and I think its related to this, I'm going to assume because its not `[inline]`'d this is now calling 2 functions instead of 1, and because its on a hotpath its actually slowed it down. Ill try adding the inline attribute and see if thats what it is, but leaving this here as a reminder.

PR #3976, opened 2024-09-02
fix: ignore `debugger` statement

shurizzle:
Right now, if I try to run js code with a `debugger` statement in it, I got the error:
`SyntaxError: unexpected token 'debugger', primary expression at line 14, col 645`.

In this PR, I’m simply ignoring the `debugger` statement. I know it’s not a definitive solution, but at least it allows the code to run.


jedel1043 (approved):
Looks great! And yeah, long term we should add a new API that's the dual of `eval` and `run` but that it also allows to stop the execution and inspect the current `VM` when we reach a `debugger` statement.

nekevss (approved):
LGTM

PR #3981, opened 2024-09-05
Add TryFromJs for TypedJsFunction and more tests

hansl:
Includes adding TryFromJs for "()".

### DIFF
--- a/core/engine/src/object/builtins/jsfunction.rs
+++ b/core/engine/src/object/builtins/jsfunction.rs
@@ -54,6 +54,11 @@ impl<A: TryIntoJsArguments, R: TryFromJs> TypedJsFunction<A, R> {
         self.inner.clone()
     }
 
+    /// Get the inner `JsFunction` without consuming this object.
+    pub fn to_js_function(&self) -> JsFunction {
+        self.inner.clone()
+    }
+
     /// Call the function with the given arguments.
     #[inline]
     pub fn call(&self, context: &mut Context, args: A) -> JsResult<R> {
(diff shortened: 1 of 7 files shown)


core/engine/src/object/builtins/jsfunction.rs
```diff
@@ -54,6 +54,12 @@ impl<A: TryIntoJsArguments, R: TryFromJs> TypedJsFunction<A, R> {
         self.inner.clone()
     }
 
+    /// Get the inner `JsFunction` without consuming this object.
+    #[must_use]
+    pub fn to_js_function(&self) -> JsFunction {
+        self.inner.clone()
+    }
```
HalidOdat, line 61:
```suggestion
    /// Get the inner `JsFunction` without consuming this object.
    #[must_use]
    pub fn as_js_function(&self) -> &JsFunction {
        &self.inner
    }
```

The `to_` implies that the convert into a type is expensive, I think it's best to to use `as_` as suggested in the [rust api guidelines](https://rust-lang.github.io/api-guidelines/naming.html?highlight=to_#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

HalidOdat (approved):
Looks great! Just a minor nitpick :)

core/engine/src/object/builtins/jsfunction.rs
hansl: It comes down to usage. This would be very different and much more boilerplate, as `JsFunction` isn't Copy: https://github.com/boa-dev/boa/pull/3981/files#diff-c88ba8c6fb9beadd76ba03bcc3e5dab23da271da9a9862ca2d8c565c37fdd6acR21-R27

hansl: I kind of think we could also `impl Into<JsFunction>`, which would likely solve both. Let me try.

hansl: Going to resolve this comment by doing the commit suggestion above, AND implementing `From<TypedJsFunction<...>> for JsFunction` if you don't need to keep the object.

HalidOdat (approved):

jedel1043 (approved):
LGTM!

PR #3984, opened 2024-09-05
Allow dead code for code that is newly detected as unused

hansl:
It is unclear if it is actually unused (`ConstructorNoProto` seems like it is fully unused, but not `Constructor`). Trying to create an MRE for a rust-lang issue, but couldn't reproduce outside of this repro.

This at least temporarily to unblock CI.

Closes #3985.


hansl:
Added #3985 to see if CI is unstuck.

jedel1043 (approved):

HalidOdat (approved):

PR #3985, opened 2024-09-06
Fix compile errors with nightly rust

theoparis:
This Pull Request fixes/closes #{3982}.

It changes the following:

- Wrap Mutex<FutexWaiters> with a type that implements Sync

Additionaly, the CI fails with `error: struct Constructor is never constructed`, not sure what I should do about that
Let me know if the struct type should have a name for the mutex instead of using `.0`


hansl:
@theoparis Ironically I have a (temporary?) fix for the compiler issue, which is blocked by this issue. I'll merge your commit in my PR to see if the CI passes.

PR #3986, opened 2024-09-07
Add an "iter()" method to Js*Array for convenience

hansl:
It might be optimizable, and in either case it should be but this at
least let people use an easy API and enjoy future improvements.


HalidOdat (approved):
Nice addition! :)

jedel1043 (approved):

PR #3994, opened 2024-09-11
Add TextEncoder, TextDecoder implementations to boa_runtime

hansl:


jedel1043 (approved):
Looks great! Seeing that we'll keep increasing the number of WebAPI things in `boa_runtime` maybe it'll be worth adding a global function `boa_runtime::register` to add all the builtins into the global object.

raskad (approved):
Looks good :)

PR #3996, opened 2024-09-12
Limit actions runs to 1 per branch and fix macos release

jedel1043:
Fixes the macos release binary action.
Also limits the amount of actions running per PR/protected branch to 1, to avoid running the actions multiple times when pushing multiple times to the same PR/branch.


raskad (approved):
Nice change

jasonwilliams (approved):
LGTM

PR #3998, opened 2024-09-19
`TryFromJs` from `JsMap` for `HashMap` & `BtreeMap`

Nikita-str:
Seems like `TryFromJs` didn't convert correctly from `JsMap` into `HashMap` / `BtreeMap`

This PR fix it

**Example**  
Next code
```Rust
use boa_engine::{value::TryFromJs, Context, JsResult, Source};
use std::collections::HashMap;

fn main() -> JsResult<()> {
    let js_code = "new Map([['a', 2], ['b', 3]])";

    let js_value = context.eval(Source::from_bytes(js_code))?;
    println!("{}", js_value.display());

    let map = HashMap::<String, i32>::try_from_js(&js_value, &mut context)?;
    println!("{map:?}");
    Ok(())
}
```
before this changes will return:
```
Map { "a" → 2, "b" → 3 }
{}
```
and after changes returns:
```
Map { "a" → 2, "b" → 3 }
{"a": 2, "b": 3}
```

**Notes**  
I don't find more fast way(than the one used in `for_each_elem_in_js_map`) to get `(key, value)`'s from `JsMap`.
Maybe there is a way?
Or maybe I should write it (for example by function like [next](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/map/map_iterator.rs#L110-L133) but that just return `Option<(JsValue, JsValue)>`)?

### DIFF
--- a/core/engine/src/object/builtins/jsmap.rs
+++ b/core/engine/src/object/builtins/jsmap.rs
@@ -406,7 +406,7 @@ impl JsMap {
     #[inline]
     pub fn rust_for_each<F>(&self, f: F) -> JsResult<()>
     where
-        F: FnMut(&JsValue, &JsValue) -> JsResult<()>,
+        F: FnMut(JsValue, JsValue) -> JsResult<()>,
     {
         let this = self.inner.clone().into();
         Map::rust_for_each(&this, f)
--- a/core/engine/src/object/builtins/jsmap.rs
+++ b/core/engine/src/object/builtins/jsmap.rs
@@ -404,12 +404,12 @@ impl JsMap {
 
     /// Executes the provided callback function for each key-value pair within the [`JsMap`].
     #[inline]
-    pub fn rust_for_each<F>(&self, f: F) -> JsResult<()>
+    pub fn for_each_native<F>(&self, f: F) -> JsResult<()>
     where
         F: FnMut(JsValue, JsValue) -> JsResult<()>,
     {
         let this = self.inner.clone().into();
-        Map::rust_for_each(&this, f)
+        Map::for_each_native(&this, f)
     }
 
     /// Returns a new [`JsMapIterator`] object that yields the `value` for each element within the [`JsMap`] in insertion order.
--- a/core/engine/src/object/builtins/jsmap.rs
+++ b/core/engine/src/object/builtins/jsmap.rs
@@ -402,6 +402,16 @@ impl JsMap {
         )
     }
 
+    /// Executes the provided callback function for each key-value pair within the [`JsMap`].
+    #[inline]
+    pub fn rust_for_each<F>(&self, f: F) -> JsResult<()>
+    where
+        F: FnMut(&JsValue, &JsValue) -> JsResult<()>,
+    {
+        let this = self.inner.clone().into();
+        Map::rust_for_each(&this, f)
+    }
+
     /// Returns a new [`JsMapIterator`] object that yields the `value` for each element within the [`JsMap`] in insertion order.
     #[inline]
     pub fn values(&self, context: &mut Context) -> JsResult<JsMapIterator> {
--- a/core/engine/src/value/conversions/try_from_js/collections.rs
+++ b/core/engine/src/value/conversions/try_from_js/collections.rs
@@ -31,7 +31,7 @@ where
                 );
                 Ok(())
             };
-            for_each_elem_in_js_map(js_map, f, context)?;
+            for_each_elem_in_js_map(&js_map, f, context)?;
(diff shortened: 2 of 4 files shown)


core/engine/src/value/conversions/try_from_js/collections.rs
```diff
@@ -62,3 +95,44 @@ where
...
+            return unexp_obj_err();
+        };
+
+        let done = iter_obj.get(js_str!("done"), context)?;
+        let Some(done) = done.as_boolean() else {
+            return unexp_obj_err();
+        };
+        if done {
+            break;
+        }
```
jedel1043, line 124: Here I think you can use [`IteratorResult`](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/iterable/mod.rs#L314) to do all the checks for you. We'll probably want to make `next` return that instead of the raw `JsValue`, but constructing it manually should be fine for now.

core/engine/src/value/conversions/try_from_js/collections.rs
```diff
@@ -62,3 +95,44 @@ where
...
+where
+    F: FnMut(JsValue, JsValue, &mut Context) -> JsResult<()>,
+{
+    let unexp_obj_err = || {
+        JsResult::Err(
+            JsNativeError::typ()
+                .with_message("MapIterator return unexpected object")
+                .into(),
+        )
+    };
```
jedel1043, line 109: On second thought, this looks like something that should replace `JsMap::for_each`, right? I think it would be a lot more ergonomic instead of our current API.

Nikita-str: ~Yes, but I actually not sure how to do it via `JsMp::for_each` because you should to create an `JsFunction` and how I can see it can be done via `NativeFunctionPointer` but you can't capture in such function outer `map` Rust object (`HashMap`/`BTreeMap`). And if you use closure that will capture the `map`, then it seems like you can't create `JsFunction`.

Actually we don't want to create extra js objects here (like `JsFunction`), and therefor maybe it makes sense to create an function in `JsMap` that will do `for_each` but for Rust fn (for `F: FnMut`), and maybe make it public only in the crate.

jedel1043: Yes, I was more talking about substituting `JsMap::for_each` with this new function, since it seems that passing `JsFunction` directly is a lot more painful overall.

Nikita-str: Done

Is `riust_for_each` acceptable name?

Just for sureness, is [this](https://github.com/Nikita-str/boa/blob/try-from-js--JsMap/core/engine/src/builtins/map/mod.rs#L479-L501) equivalent to the next:
```rust
let map = map
    .downcast_ref::<OrderedMap<JsValue>>()
    .expect("checked that `this` was a map");

map.iter().try_for_each(|(k, v)|{
    let args = [v.clone(), k.clone(), this.clone()];
    callback.call(this_arg, &args, context).map(|_|())
})?;
Ok(JsValue::undefined())
```
If they are not the same, then seems like I incorrectly impl the `fn rust_for each`

If they the same, maybe simplify it?

jedel1043: Not really sure, but I think we have pretty good coverage. You can try replacing it to see if there are any regressions on the test262 suite.

Nikita-str: No, they are not the same because `JsMap` can be changed in `for_each` and it will panic because of mut and non-mut borrowing at the same time

jedel1043 (approved):
Looks great! Thank you!

nekevss (commented):
This is looking great! Just had a small thought on the function name that I thought might be worth a little bikeshedding.

core/engine/src/object/builtins/jsmap.rs
```diff
@@ -402,6 +402,16 @@ impl JsMap {
         )
     }
 
+    /// Executes the provided callback function for each key-value pair within the [`JsMap`].
+    #[inline]
+    pub fn rust_for_each<F>(&self, f: F) -> JsResult<()>
```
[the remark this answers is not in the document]
jedel1043: Yeah, maybe that would be a better convention.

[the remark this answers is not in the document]
nekevss: I'd second `for_each_native` as well.

nekevss (approved):
Looks great! Thanks for your work on this! :smile:

PR #3999, opened 2024-09-19
`TryIntoJs` trait and derive macro for it

### LINKED ISSUE #3874: Add a `TryIntoJs` trait and derive macro to transform Rust values into `JsValue`
Basically the opposite of `TryFromJs`.

Conversions `From<T> for JsValue` are implemented for a lot of primitive values, but not as much for complex Rust types.

This is useful when calling into JavaScript functions (using `JsObject::call()`) and passing in Rust values that are not primitive.

Nikita-str:
This Pull Request closes #3874 

**Notes**
* I don't sure about conversion of `HashMap` into `JsObject`.
Should it be converted into `JsMap` or  into ordinary object `{ k1: v1, ..., kN: vN }`?
Or maybe I need to add a macro attribute to allow choose a way of conversion of `Map` types?

### DIFF
--- a/core/engine/src/value/conversions/mod.rs
+++ b/core/engine/src/value/conversions/mod.rs
@@ -7,6 +7,7 @@ use super::{JsBigInt, JsObject, JsString, JsSymbol, JsValue, Profiler};
 mod either;
 mod serde_json;
 pub(super) mod try_from_js;
+pub(super) mod try_into_js;
 
 pub(super) mod convert;
 
--- a/core/engine/src/value/conversions/try_into_js.rs
+++ b/core/engine/src/value/conversions/try_into_js.rs
@@ -211,27 +211,26 @@ mod try_into_js_tests {
     use crate::{Context, JsResult};
 
     #[test]
-    fn big_int_err() -> JsResult<()> {
-        let mut context = Context::default();
-        let context = &mut context;
-
-        fn assert<T: TryIntoJs>(int: T, context: &mut Context) {
+    fn big_int_err() {
+        fn assert<T: TryIntoJs>(int: &T, context: &mut Context) {
             let expect_err = int.try_into_js(context);
             assert!(expect_err.is_err());
         }
+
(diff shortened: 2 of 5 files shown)


HalidOdat (commented):
Why do we need to have a separate derive and implementation for it we can have impl for it, like the standard library does it: https://doc.rust-lang.org/beta/src/core/convert/mod.rs.html#788-815. Implementing `TryFromJs` should automaticaly implement `TryIntoJs`.

Am I missing something?

Nikita-str:
@HalidOdat

`TryFromJs` converts `JsValue` into Rust types, so there will no such type that converts a Rust type via `TryFromJs` into `JsValue`.

In std mod we can have an object of any type as value in `try_from`.
In [the definition of the trait](https://github.com/boa-dev/boa/blob/main/core/engine/src/value/conversions/try_from_js.rs#L14) we specifically have `value: &JsValue`.

So
* `TryFromJs` can make only `JsValue --> Rust`
* `TryIntoJs` can make only `Rust--> JsValue`

There is [try_js_into fn](https://github.com/boa-dev/boa/blob/main/core/engine/src/value/conversions/try_from_js.rs#L20-L25) that do the same as `TryFromJs`: converts `JsValue` into Rust type. (`self` is `JsValue`)

HalidOdat (approved):
Ahh, I see, yes that makes sense!

The PR looks good to me! :)

HalidOdat:
> I don't sure about conversion of HashMap into JsObject. Should it be converted into JsMap or into ordinary object { k1: v1, ..., kN: vN }? [..]

I should probably be a `JsMap` by default, it's also more space efficient, since the properties are stored in an hashmap <https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/map/ordered_map.rs#L36-L42>, rather that in the shape and object storage.

jedel1043 (approved):
Pretty impressive work! Looks really nice :)

PR #4004, opened 2024-09-24
Add a URL class to boa_runtime

hansl:
Some methods are NOT currently supported (some don't make sense outside of a browser context). They are still implemented but will throw a JavaScript Error.

Supported methods should follow the specification perfectly and are tested for it.

Notably missing is the `URLSearchParams` class which is complex with the current Boa API (making iterables is hard) and should be done in a separate PR.

### DIFF
--- a/core/runtime/src/url.rs
+++ b/core/runtime/src/url.rs
@@ -23,6 +23,7 @@ use std::fmt::Display;
 
 /// The `URL` class represents a (properly parsed) Uniform Resource Locator.
 #[derive(Debug, Clone, JsData, Trace, Finalize)]
+#[boa_gc(unsafe_no_drop)]
 pub struct Url(#[unsafe_ignore_trace] url::Url);
 
 impl Url {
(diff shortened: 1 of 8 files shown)


core/runtime/src/url.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn from(url: url::Url) -> Self {
+        Self(url)
+    }
+}
+
+impl From<Url> for url::Url {
+    fn from(url: Url) -> url::Url {
+        // Cannot avoid cloning here, unfortunately, as we would need to replace
+        // the internal URL with something else.
+        url.0.clone()
```
HalidOdat, line 78: For this we can use the `#[boa_gc(unsafe_no_drop)]` on the Url which will not generate a Drop `impl` for `Url` and the value can be moved out.

HalidOdat (approved):
Nice work! Besides the comment, this looks good to me! :)

PR #4005, opened 2024-09-24
Add context to the console `Logger` trait

hansl:
This is not backward compatible but should have minimal impact. This is also more future proof by passing the ConsoleState which is more opaque and allows us to update the way the console object is constructed and passed.


HalidOdat (approved):
Looks good to me! :)

jedel1043 (approved):

PR #4008, opened 2024-09-26
Add WPT as optional tests for boa_runtime

hansl:
To run the tests, you require adding the `wpt-tests` feature and setting the `WPT_ROOT` environment variable on the command line. This is currently depending on a pending rstest PR: https://github.com/la10736/rstest/pull/277

Example:
```bash
WPT_ROOT=$HOME/Sources/hansl/wpt cargo test -p boa_runtime --features "wpt-tests"
```

The current set of curated tests are all working.

This will be considered a draft until the URL PR (#4004) is merged. Might also need to add support for more encoding types in `TextEncoder`/`TextDecoder`.

### DIFF
--- a/core/runtime/Cargo.toml
+++ b/core/runtime/Cargo.toml
@@ -15,9 +15,12 @@ boa_engine.workspace = true
 boa_gc.workspace = true
 boa_interop.workspace = true
 rustc-hash = { workspace = true, features = ["std"] }
+url = { version = "2.5.2", optional = true }
 
 [dev-dependencies]
+boa_interop.workspace = true
 indoc.workspace = true
+rstest.workspace = true
 textwrap.workspace = true
 
 [lints]
@@ -25,3 +28,8 @@ workspace = true
 
 [package.metadata.docs.rs]
 all-features = true
+
+[features]
+default = ["all"]
+all = ["url"]
+wpt-tests = []
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
 name = "ahash"
 version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -52,9 +58,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.14"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "418c75fa768af9c03be99d17643f93f79bbba589895012a80e3452a19ddda15b"
+checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -73,27 +79,27 @@ checksum = "1bec1de6f59aedf83baf9ff929c98f2ad654b97c9510f4e70cf6f661d49fd5b1"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.4"
+version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c03a11a9034d92058ceb6ee011ce58af4a9bf61491aa7e1e59ecd24bd40d22d4"
+checksum = "eb47de1e80c2b463c735db5b217a0ddc39d612e7ac9e2e96a5aed1f57616c1cb"
 dependencies = [
  "utf8parse",
 ]
 
 [[package]]
 name = "anstyle-query"
-version = "1.1.0"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ad186efb764318d35165f1758e7dcef3b10628e26d41a44bc5550652e6804391"
+checksum = "6d36fc52c7f6c869915e99412912f22093507da8d9e942ceaf66fe4b7c14422a"
 dependencies = [
  "windows-sys 0.52.0",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.3"
+version = "3.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "61a38449feb7068f52bb06c12759005cf459ee52bb4adc1d5a7c4322d716fb19"
+checksum = "5bf74e1b6e971609db8ca7a9ce79fd5768ab6ae46441c572e46cf596f59e57f8"
 dependencies = [
  "anstyle",
  "windows-sys 0.52.0",
@@ -139,13 +145,13 @@ dependencies = [
 
 [[package]]
 name = "async-executor"
-version = "1.12.0"
+version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c8828ec6e544c02b0d6691d21ed9f9218d0384a82542855073c2a3f58304aaf0"
+checksum = "30ca9a001c1e8ba5149f91a74362376cc6bc5b919d92d988668657bd570bdcec"
 dependencies = [
  "async-task",
  "concurrent-queue",
- "fastrand 2.1.0",
+ "fastrand 2.1.1",
  "futures-lite 2.3.0",
  "slab",
 ]
@@ -163,9 +169,9 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "2.3.3"
+version = "2.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0d6baa8f0178795da0e71bc42c9e5d13261aac7ee549853162e66a241ba17964"
+checksum = "444b0228950ee6501b3568d3c93bf1176a1fdbc3b758dcd9475046d30f4dc7e8"
 dependencies = [
  "async-lock",
  "cfg-if",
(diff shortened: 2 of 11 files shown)


core/runtime/Cargo.toml
```diff
@@ -30,3 +32,10 @@ all-features = true
 [features]
 default = ["all"]
 all = ["url"]
+wpt-tests = []
+# This feature is used to disable compiling the WPT tests in the CI
+# when using `--all-features`. If this feature is enabled, we disable
+# the compilation of the WPT tests (since they will panic).
+# Clippy will be disabled for those methods as well, but they're one
+# line long.
+wpt-tests-do-not-use = []
```
jedel1043, line 41: Hmmm, I think it's slightly weird to have tests be enabled by features instead of by only the `test` cfg. Can we just have a separate crate in the workspace for wpt tests, just like `boa_tester`? We could also potentially merge both of those in the future.

tests/wpt/src/logger/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[derive(Trace, Finalize, JsData)]
+struct RecordingLoggerInner {
+    pub log: Vec<RecordingLogEvent>,
+    pub error: Vec<RecordingLogEvent>,
+}
+
+#[derive(Clone, Trace, Finalize, JsData)]
+pub(crate) struct RecordingLogger {
+    inner: Gc<GcRefCell<RecordingLoggerInner>>,
```
jedel1043: We probably don't need `Gc` here, since no inner types have traceable fields. Maybe just use `Rc<RefCell<>>` and `#[unsafe_ignore_trace]`?

tests/wpt/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+    // "Base URL about:blank cannot be a base"
+    #[exclude("url-searchparams.any.js")]
+    // "fetch is not defined"
+    #[exclude("url-origin.any.js")]
+    #[exclude("url-setters.any.js")]
+    #[exclude("url-constructor.any.js")]
+    path: PathBuf,
+) {
+    execute_test_file(&path);
+}
```
jedel1043: This seems fine for now since we only need to test a small subset of the suite, but do you plan to eventually migrate the tests to something like `boa_tester`?

I ask this because it just feels very boilerplatey having to manually declare every test we want to run.

tests/wpt/src/lib.rs
```diff
@@ -0,0 +1,395 @@
+//! Integration tests running the Web Platform Tests (WPT) for the `boa_runtime` crate.
+#![allow(unused_crate_dependencies)]
+#![cfg(feature = "wpt")]
```
jedel1043: I'm wondering if a feature is the correct approach to gate tests...
I think the Rust book recommends ignoring tests for tests that you don't want to always run, but I suppose we would want to have that functionality available.

The other approach would be to use a compiler flag instead (`#![cfg(wpt)]`), which would make things like `--all-features` still work in CI, with the downside of making the command to run the tests a bit less straightforward (`RUSTFLAGS='--cfg wpt' cargo test -p boa_wpt`).

tests/wpt/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+            "Test did not complete in 10 seconds."
+        );
+    }
+}
+
+/// Test the console with the WPT test suite.
+#[cfg(not(clippy))]
+#[rstest::rstest]
+fn console(
+    #[base_dir = "${WPT_ROOT}"]
```
jedel1043, line 355: I'm thinking that it would be better to add an internal tool to clone the wpt repo on a certain commit hash in the project root, then change `WPT_ROOT` to point to that instead. Makes it easier to keep track of the latest tested hash.

tests/wpt/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+
+    Ok(())
+}
+
+fn complete_callback__(ContextData(test_done): ContextData<TestCompletion>) {
+    test_done.done();
+}
+
+#[derive(Debug, Clone, Trace, Finalize, JsData)]
+struct TestCompletion(Gc<AtomicBool>);
```
jedel1043: This seems... wrong. `Gc` is thread-unsafe, and `AtomicBool` is used to add thread safety to things, so something's not working correctly. This should either be `Arc<AtomicBool>` if this is shared between multiple threads, or `Rc<Cell<bool>>` if this is not shared between threads.

tests/wpt/src/lib.rs
hansl: The tests could simply be `#[files("*/*.js")]`, but there is a lot more WPT than test262 that we don't support (or ever plan to). So this makes sense. Also, unit tests make sense as it hooks into the Rust test framework and allows all reporting options (like coverage) that `cargo test` supports.

tests/wpt/src/logger/mod.rs
hansl: This is old code, but now that I think about it this will likely happen a lot; my inner type does implement Trace and Finalize, so I might have thought I needed to trace it using Gc and co.

tests/wpt/src/lib.rs
hansl: That's great, I didn't know you could do that.

hansl: Oh wait, that `cfg` flag forces to recompile _everything_. I'd rather have a feature that's documented. Adding a documentation to it.

tests/wpt/src/lib.rs
hansl: I can add a `build.rs` script that does it when the tests are enabled. WDYT?

jedel1043: I don't think it's good practice to download something in the background using `build.rs`

tests/wpt/src/lib.rs
jedel1043: Oh, if there's a way to run big sets of tests then that's fine.

> unit tests make sense as it hooks into the Rust test framework and allows all reporting options (like coverage) that cargo test supports.

I'm up for integrating with `cargo test`, but let's just take note that hooking into the Rust test framework has its disadvantages, like the hacks that we need to do to make it possible to run `cargo test` without having to run the wpt suite. I think it makes sense in this case because the subset of `wpt` that we support is so small that we don't need advanced filtering/collection/categorization of tests, but this wouldn't work for test262.

tests/wpt/src/lib.rs
jedel1043: If we're going with the simple feature, then you should set the `default-members` field on the workspace `Cargo.toml` and exclude `boa_wpt` from that, to make `--all-features` work on the CI.

jedel1043: Oh wait, does that work even if we're using `cargo workspaces`? I haven't tested it yet. If it works then that's nice, otherwise we would have to modify all our CI to manually exclude `boa_wpt` from the list of tested packages.

tests/wpt/src/lib.rs
hansl: I think the build system is kind of meant to do that. For example, some packages will download and build dependencies for C++ bindings. Some build scripts are expansive (and expensive). 

For example, https://github.com/Rust-SDL2/rust-sdl2/blob/master/sdl2-sys/build.rs runs git submodule update which is also something we could do (although personally I don't like submodules, we could hide the ugly parts in scripts).

Or we could go the OpenSSL route and have a crate that's the WPT source and not rely on git to download it. I found that interesting.

tests/wpt/src/lib.rs
hansl: Yeah it doesn't really work. Let me see if I can make something without having to list all default members (which would be a foot gun later when we add new members).

tests/wpt/src/lib.rs
jedel1043: I think `build.rs` works when you absolutely need to download something for your build, but in our case we don't want to always download the repo, and putting that in the `build.rs` will force everyone to download that when they do `cargo build --all-features`, even if they're not working on `boa_runtime`.

EDIT: maybe this is not entirely true? I see that your script gates the main function which seems to work fine in this case. The recompilation argument does still apply though.

jedel1043: Another disadvantage I see is that if you wanted to bump the commit hash of wpt, you would be forced to recompile the crate, even if you didn't change anything about the code.

hansl: Adding the project to excluded removes it entirely from the workspace, but you can still run it in the directory itself. I think that's good enough, WDYT?

hansl: I updated the instructions. It's simpler now.

jedel1043: Separating the crate does seem like a nice solution for the optional tests. This doesn't really solve the recompilation problem about `build.rs` though

hansl: The build script isn't running when the workspace is built. And it doesn't run again if the tests_wpt folder is there. Cloning the repo takes about 3-4 minutes on my crappy hotel connection (it's a shallow clone), and updating the SHA takes 5-10 seconds.

The only case where it reruns is if you delete the folder, running cargo clean and rebuilding is instant.

jedel1043: Have you tried to update the SHA? Because from reading the code, it seems that changing the SHA won't trigger an update, because `build.rs` only tracks if `tests_wpt` have changed, so you would have to change something about the code in order to trigger the bump.

Maybe you just need to move `tests_wpt_config.toml` to `tests/wpt`? I'm not sure if `build.rs` tracks only rust files or every file in the specified path. Can you test that? If that works fine then a `build.rs` should work fine.

hansl: > because build.rs only tracks if tests_wpt have changed

It checks that the SHA at the HEAD of the tests_wpt/ folder is different. If it is, it will fetch and checkout the new rev.

Manually doing it on my computer (changing to a previous SHA then running build then checking the SHA):

CleanShot 2025-07-01 at 17 01 40@2x

Then running `cargo build` in the root which is instant, then in the project which takes about 13 seconds, then in the project again:

image

hansl: @jedel1043 would it be useful to use git commands instead of libgit2?

jedel1043: That's not exactly what I mean. If you change the hash in `tests_wpt_config.toml` without touching anything inside `tests_wpt`, does the `build.rs` detect that it needs to fetch the new hash?

hansl: Seems the `build` script is always run (even if nothing changes), but doesn't do anything if there's no change in the hash or the repo.

jedel1043: Sounds good then!

jedel1043 (approved):
Thank you for addressing all my suggestions!

PR #4010, opened 2024-09-28
`IntegerOrInfinity` `eq` bug fix

### LINKED ISSUE #4009: Endless loop caused by eq between IntegerOrInfinity::Integer and i64
In boa-main/core/engine/src/value/integer.rs, following code will cause endless loop.
```rust
let value: i64 = 42;
let other = IntegerOrInfinity::Integer(10);
assert!(!value.eq(&other));
```

Nikita-str:
This Pull Request fixes #4009


HalidOdat (approved):
Nice catch! Thanks for the contribution! @Nikita-str

jedel1043 (approved):

PR #4011, opened 2024-09-28
`JsValue::to_json` fix integer property keys

### LINKED ISSUE #3923: JsValue to_json removes integer keys from objects rather than stringify them.
**Describe the bug**
From 0.17 through 0.19 the to_json() function of JsValue removes integer (even integer strings) from Json objects

**To Reproduce**
0.16 code that worked:
```rust
use boa_engine::{object::ObjectInitializer, property::Attribute, Context, JsValue};

fn main() {
    let object = ObjectInitializer::new(&mut context)
        .property("0", 1, Attribute::all())
        .property("y", 1, Attribute::all())
        .build();
    let value = JsValue::from(object);
    println!("json: {}", value.to_json(&mut context).unwrap());
}

// output: json: {"0":1,"y":1}
```

0.17-0.19 code that has the bug:
```rust
use boa_engine::{object::ObjectInitializer, property::Attribute, Context, JsValue};

fn main() {

    let object = ObjectInitializer::new(&mut context)
        .property(boa_engine::js_string!("0"), 1, Attribute::all())
        .property(boa_engine::js_string!("y"), 1, Attribute::all())
        .build();

    let value = JsValue::from(object);
    println!("json: {}", value.to_json(&mut context).unwrap());
}

// output: json: {"y":1}
```

**Expected behavior**
We should be getting the same output as in 0.16: `json: {"0":1,"y":1}`

- OS: WSL Ubuntu 
- Version: 22.04.4 LTS
- Rustc version: rustc 1.79.0 (129f3b996 2024-06-10)

**Additional context**
The problem seems to stem from [this PR](https://github.com/boa-dev/boa/commit/ed3744817577d33fe944fd11d73843bfa0a64f8a#diff-b4a3ed8c289bb81ba37d3cf9ca235c10f2481f40075522cde95cb3fa86f8f560L140) on line 140 of serde_json.rs where it changed from using `obj.borrow().properties().iter()` to `obj.borrow().properties().shape.keys()` when iterating through the object. You can see the difference in the example code of:

```rust
    ...

    let object = ObjectInitializer::new(&mut context)
        .property(boa_engine::js_string!("0"), 1, Attribute::all())
        .property(boa_engine::js_string!("y"), 1, Attribute::all())
        .build();
    println!("own_property_keys: {:?}", object.own_property_keys(&mut context).unwrap());
    println!("shape.keys: {:?}", object.borrow().shape().keys());

// own_property_keys: [Index(NonMaxU32 { inner: 0 }), String("y")]
// shape.keys: [String("y")]
```

Either `shape.keys()` needs to be fixed to return the integer keys or `to_json()` should be updated to use `own_property_keys()`. I'm not sure what the correct fix is. What is the purpose of `shape.keys()`? Should it include the integer keys?

Nikita-str:
This Pull Request fixes #3923

### DIFF
--- a/core/engine/src/value/conversions/serde_json.rs
+++ b/core/engine/src/value/conversions/serde_json.rs
@@ -140,6 +140,25 @@ impl JsValue {
                     Ok(Value::Array(arr))
                 } else {
                     let mut map = Map::new();
+                    let mut value_by_prop_key = |property_key| match obj
+                        .borrow()
+                        .properties()
+                        .get(&property_key)
+                        .and_then(|x| x.value().cloned())
+                    {
+                        Some(val) => val.to_json(context),
+                        None => Ok(Value::Null),
+                    };
+
+                    for index in obj.borrow().properties().index_property_keys() {
+                        let key = index.to_string();
+
+                        let property_key = PropertyKey::from(index);
+                        let value = value_by_prop_key(property_key)?;
+
+                        map.insert(key, value);
+                    }
+
                     for property_key in obj.borrow().properties().shape.keys() {
                         let key = match &property_key {
                             PropertyKey::String(string) => string.to_std_string_escaped(),
@@ -151,15 +170,7 @@ impl JsValue {
                             }
                         };
 
-                        let value = match obj
-                            .borrow()
-                            .properties()
-                            .get(&property_key)
-                            .and_then(|x| x.value().cloned())
-                        {
-                            Some(val) => val.to_json(context)?,
-                            None => Value::Null,
-                        };
+                        let value = value_by_prop_key(property_key)?;
 
                         map.insert(key, value);
                     }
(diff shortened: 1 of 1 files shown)


core/engine/src/value/conversions/serde_json.rs
```diff
@@ -140,6 +140,25 @@ impl JsValue {
...
                     let mut map = Map::new();
+                    let mut value_by_prop_key = |property_key| match obj
+                        .borrow()
+                        .properties()
+                        .get(&property_key)
+                        .and_then(|x| x.value().cloned())
+                    {
+                        Some(val) => val.to_json(context),
+                        None => Ok(Value::Null),
+                    };
```
CrazyboyQCD, line 151: We could avoid the `clone` here with extra `map`:
```rust
     .and_then(|x| x.value().map(|v| v.to_json(context)))
{
    Some(val) => val?,
    None => Value::Null,
};
```

Nikita-str: In the worst case it's still just cloning of a smart pointer so I don't think that it has much overhead costs, but ok, done

raskad (approved):
Thank you for the fix!

HalidOdat (approved):

PR #4020, opened 2024-10-17
Some optimizations on `Error`

CrazyboyQCD:
This brings some optimizations:
- use constant to reduce binary size(https://godbolt.org/z/YEq4hW49n)
- change `message`'s type to `Cow<'static, str>` to reduce allocations

### DIFF
--- a/core/engine/src/error.rs
+++ b/core/engine/src/error.rs
@@ -734,7 +734,7 @@ impl JsNativeError {
     pub const URI: Self = Self::uri();
     #[cfg(feature = "fuzz")]
     /// Default `NoInstructionsRemain` kind `JsNativeError`.
-    pub const IS_NO_INSTRUCTIONS_REMAIN: Self = Self::is_no_instructions_remain();
+    pub const NO_INSTRUCTIONS_REMAIN: Self = Self::no_instructions_remain();
     /// Default `error` kind `JsNativeError`.
     pub const RUNTIME_LIMIT: Self = Self::runtime_limit();
 
--- a/core/engine/src/error.rs
+++ b/core/engine/src/error.rs
@@ -716,6 +716,28 @@ const fn new_str() -> Box<str> {
 }
 
 impl JsNativeError {
+    /// Default `AggregateError` kind `JsNativeError`.
+    pub const AGGREGATE: Self = Self::aggregate(Vec::new());
+    /// Default `Error` kind `JsNativeError`.
+    pub const ERROR: Self = Self::error();
+    /// Default `EvalError` kind `JsNativeError`.
+    pub const EVAL: Self = Self::eval();
+    /// Default `RangeError` kind `JsNativeError`.
+    pub const RANGE: Self = Self::range();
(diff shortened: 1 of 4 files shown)


core/engine/src/error.rs
```diff
@@ -709,9 +709,37 @@ impl fmt::Debug for JsNativeError {
     }
 }
 
+// FIXME(const-hack): replaced with Box::default once it is `const`
+const fn new_str() -> Box<str> {
+    // SAFETY: miri safe and works in stable
+    unsafe { std::mem::transmute("") }
+}
```
jedel1043, line 716: Is this guaranteed though? IIRC, `Box<T>` has the same layout as `&T` only for `T: Sized`, but for unsized types they could always change the pointer metadata layout.

If we don't want to rely on this, we can just use `Option<Box<str>>`, which should be const-compatible.

jedel1043 (commented):
Nice improvement!

core/engine/src/error.rs
CrazyboyQCD: Yep, I missed it and miri doesn't flag for it. 
And Since there are plenty of `&'static str` message, can we use `Cow<'static, str>`?

workingjubilee: Miri's default mode avoids false positives.

This means miri only errors if your program is definitely unsound according to its rules, and avoids warning for some operations that may be unsound, but are not provably unsound.

So "miri said it was okay" is not a proof that your program is sound, it is a proof that you haven't done something _monstrously_ incorrect.

CrazyboyQCD: Good to know that.

jedel1043: Using `Cow` for this sounds good

core/gc/src/trace.rs
```diff
@@ -179,7 +179,7 @@ simple_empty_finalize_trace![
     char,
     TypeId,
     String,
-    Box<str>,
```
jedel1043, line 182: Don't remove the implementation on `Box<str>` even if we don't use it. The users of `boa` would have to implement `Trace` for a wrapper around `Box<str>` if they had it on their structs that implement `Trace`.

jedel1043 (changes requested):
Looking good! Just needs a small fix.

jasonwilliams (approved):
Aside from @jedel1043's comment LGTM

jedel1043:
@jasonwilliams you pinged a random user 😅

core/gc/src/trace.rs
CrazyboyQCD: I got error from `ra` that implement `Trace` for `Cow<'static, str>` requires that `str` implements `Trace` and I tested that `Box<str>` does not affected by this.

jedel1043: Oh, I think that's because we already implement `Trace` for `Box<T>`. Makes sense.

jedel1043 (approved):
Looks perfect!

PR #4023, opened 2024-10-24
Add a display_lossy() to write a JsString lossily

hansl:
This is faster when we don't need to keep the unpaired surrogates.


jedel1043 (approved):
Looks perfect!

HalidOdat (approved):

PR #4030, opened 2024-11-06
Some string optimizations 

### LINKED ISSUE #4024: Perf: replace `js_str!` with `js_string!` where argument is to be converted to `JsString` to reduce heap allocation
Examples: `Into<PropertyKey>`, `Into<JsValue>`.
Need to change some API like [`get_option`](https://github.com/boa-dev/boa/blob/e22a703dfd7591ccfd143e1e7ade4c2d2ba55b55/core/engine/src/builtins/options.rs#L51-L68).

CrazyboyQCD:
- Closes #4024 
- Some apis changes:
  - `get_number_option`: property: JsStr<'_> -> JsString
  - `get_option`: property: JsStr<'_> -> JsString
  - `get_temporal_unit`: key: JsStr<'_> -> JsString
  - `get_typed_fn`: name: JsStr<'_> -> JsString
- Use literal in `boolean`'s `to_string`
- Remove unnecessary `to_string` in `Number::to_js_string`
- Use more efficient `LazyLock` on `RAW_STATICS_CACHE` since it is immutable

### DIFF
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -387,7 +387,7 @@ impl String {
         let cooked = args.get_or_undefined(0).to_object(context)?;
 
         // 3. Let raw be ? ToObject(? Get(cooked, "raw")).
-        let raw = cooked.get(js_str!("raw"), context)?.to_object(context)?;
+        let raw = cooked.get(js_string!("raw"), context)?.to_object(context)?;
 
         // 4. Let literalSegments be ? LengthOfArrayLike(raw).
         let literal_segments = raw.length_of_array_like(context)?;
@@ -1129,7 +1129,7 @@ impl String {
             // b. If isRegExp is true, then
             if let Some(obj) = RegExp::is_reg_exp(search_value, context)? {
                 // i. Let flags be ? Get(searchValue, "flags").
-                let flags = obj.get(js_str!("flags"), context)?;
+                let flags = obj.get(js_string!("flags"), context)?;
 
                 // ii. Perform ? RequireObjectCoercible(flags).
                 flags.require_object_coercible()?;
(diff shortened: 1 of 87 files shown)


core/engine/src/builtins/string/mod.rs
```diff
@@ -22,7 +22,10 @@ use crate::{
     value::IntegerOrInfinity,
     Context, JsArgs, JsResult, JsString, JsValue,
 };
-use boa_macros::{js_str, utf16};
+#[cfg(feature = "annex-b")]
```
hansl, line 25: Nit: move this next to line 38-39, which already has a feature. Keep the feature flags together.

core/engine/src/module/mod.rs
```diff
@@ -625,16 +624,17 @@ impl Module {
 
     /// Get an exported function, typed, from the module.
     #[inline]
+    #[allow(clippy::needless_pass_by_value)]
     pub fn get_typed_fn<A, R>(
         &self,
-        name: JsStr<'_>,
+        name: JsString,
```
hansl, line 630: Nit: This should probably be a compatible `impl Into<PropertyKey>` here, similar to `Module::get_value`.

hansl (approved):
Two non-blocking nits, but otherwise everything is pretty straightforward.

LGTM.

core/engine/src/module/mod.rs
CrazyboyQCD: `impl Into<PropertyKey>` includes integer and `JsSymbol`, which are illegal here in `ecmascript`, I think it will mislead user.

hansl (approved):
LGTM

jedel1043 (approved):
Nice work!

nekevss (approved):

PR #4038, opened 2024-11-11
`SourceText` collection & `toString()` for fns and methods

### LINKED ISSUE #3780: Function.prototyp.toString only returns a fixed `[native code]` string
**ECMASCript feature**
Function.prototyp.toString returns the actual javascript implementation in v8. [ECMAScript specification][spec]

[spec]: https://tc39.es/ecma262/#sec-function.prototype.tostring

**Example code**

```javascript
let f = () => 1 + 1;
console.log(f.toString());
```
The expected output is `() => 1 + 1`.

Nikita-str:
This Pull Request closes #3780

In current approach all `SourceText` collected (because mostly a script consist of function codes and classes) and `LinearSpan` used to retrieve it when needed (to be fair `toString()` not often called); `Gc::<SourceText>` used for storage and therefore it can be deallocated only when we have access to no function from a source anymore.

It actually solve some tests from `test262`, but there is implementation for checking `toString()` correctness that take any `[native code]` as correct, without regard to whether it should accept in concrete case or not (`assertToStringOrNativeFunction`).

Also I don't sure about `derive(arbitrary::Arbitrary)` for `core\ast\src\source.rs::Script` & `LinearSpan`, just derive is ok here?


raskad:
I have not done an in depth review yet, just an initial question. As far as I see, the positions you used for the function start and end (`LinearSpan` and `LinearPosition`) only contain the start and end line of the function, correct?

I think that would not work in cases where other stuff is in the same line, e.g.:
```js
let a = 1; function f() {return a + 1}; let b = 3;
```

In this case `f.toString()` should only be `function f() {return a + 1}` but it would return the complete line, right?

Nikita-str:
also you can read [the issue](https://github.com/boa-dev/boa/issues/3780) where it was accepted to collect the source code(it should slow down parse stage performance)

core/ast/src/source.rs
```diff
@@ -91,6 +92,11 @@ impl Script {
 
         Ok(bindings)
     }
+
+    /// Takes the source text.
+    pub fn take_source(&mut self) -> Option<SourceText> {
+        self.source.take()
+    }
```
raskad, line 99: The AST should be mostly immutable after parsing so I would change this to returning `&SourceText`. This would also be a problem if you would try to compile a `Script` multiple times.

core/ast/src/source_text.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// Constructs a new, empty `SourceText` with at least the specified capacity.
+    #[must_use]
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self {
+            source_text: Vec::with_capacity(capacity),
+        }
+    }
+    /// Get current `LinearPosition`.
+    #[must_use]
+    pub fn cur_linear_position(&self) -> LinearPosition {
```
raskad, line 21: Just for readability it would be great if you could leave an empty line between functions so the file does not look so cluttered.

core/engine/src/spanned_source_text.rs
```diff
@@ -0,0 +1,52 @@
+use boa_ast::{LinearSpan, SourceText};
+use boa_gc::{Finalize, Gc, Trace};
+
+#[derive(Trace, Finalize)]
+pub(crate) struct SourceTextInner {
+    #[unsafe_ignore_trace]
+    source_text: SourceText,
+}
```
raskad, line 8: I would suggest we rename this to `SourceText` and avoid the naming conflict, by referencing `boa_ast::SourceText` directly where we need it.
Additionally I think this should probably have an `Inner` struct that contains the `boa_ast::SourceText` and is wrapped in `Gc`. Because everywhere this is used, I think it is wrapped in a `Gc` anywas and like this we just make it easier to work with.

core/engine/src/spanned_source_text.rs
```diff
@@ -0,0 +1,52 @@
...
+impl SourceTextInner {
+    #[must_use]
+    pub(crate) fn new(code: &mut boa_ast::Script) -> Option<Gc<Self>> {
+        code.take_source()
+            .map(|source_text| Gc::new(Self { source_text }))
+    }
```
raskad, line 14: Just make this function take `boa_ast::SourceText` instead of the `Script`. It does not need anything else. Also together with the other suggestions it would just return `Self`.

core/engine/src/spanned_source_text.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/// Contains pointer to source code and span of the object.
+#[derive(Trace, Finalize, Clone)]
+pub struct SpannedSourceText {
+    gc: Gc<SourceTextInner>,
+    #[unsafe_ignore_trace]
+    span: LinearSpan,
+}
```
raskad, line 23: What do you think about leaving out this struct completeley and having the two fields exist separetley where needed (`CodeBlock` and `Bytecompiler`)? On the `Bytecompiler` we already have the source text anyway and with this we basically duplicate it.

raskad (changes requested):
Sorry for letting you wait so long for the full review. First of all thank you for the great contribution! Really nice work!

I added some specific comments, but I have one more general one. I would be in favor of Removing all the `Option`s for the various `SourceText` fields. In principle, everytime we parse, compile and execute code, we must have the source text and a span. I did a quick check an I'm 99% sure the `Option`s can be removed in all places.
In addition to the `Script` struct, the `Module` struct would also need the source text.
In some cases (e.g. the "dummy" use of the `Bytecompiler` for syntetic modules) there is really no source text, but there it is also never used. In those cases we can just use the default (empty vec).

Regarding `Arbitrary`, I think it should be a manual implementation that is just an empty source text. You can check here for en example of a manual impl: https://github.com/boa-dev/boa/blob/517ad506aeeda14fb57cd54e4caf3ed28d2688a5/core/ast/src/expression/literal/mod.rs#L121-L137

core/ast/src/source.rs
Nikita-str: Hmm...
But how and where it can happen?
Because only the place where it used is to create `SourceTextInner` in `parse` [right after](https://github.com/boa-dev/boa/blob/f000cfae071a5bb8c2036cc8fb07eb746ec3a698/core/engine/src/script.rs#L96-L101) `parse_script` returns 

Actually I can split `boa_ast::Script` structure into 2: before take the source and after.
(something like: `pub fn split(self) -> (SourceText, Script) { ... }`)

Returning a ref (`&SourceText`) would imply cloning and I guess there is no need to do it.
In worse case I can add two ways of getting it: one way will split `boa_ast::ScriptWithSourceText` into `SourceText` and `Script` and other will create a clone of a `SourceText` and return it.

(sorry for the lag in replying)

core/engine/src/spanned_source_text.rs
Nikita-str: Hmm... Actually you are right with `CodeBlock`, but it's logical related and there almost no overheads(just one smart pointer), don't you think that with these changes the relation would be more difficult to understand?

And also do we always have an outer object that contains corresponding `SourceTextInner`?
I mean with functions it is true (it is `CodeBlock`), but as I can understand `SourceTextInner` will be also used for other purpose, and I cannot predict whether we always have such an outer object.

core/ast/src/source.rs
Nikita-str: Something like `split` done:
https://github.com/boa-dev/boa/blob/1f2893c5f61cec4fc0b7e7e2834099e62a33e9c9/core/parser/src/parser/mod.rs#L38
https://github.com/boa-dev/boa/blob/1f2893c5f61cec4fc0b7e7e2834099e62a33e9c9/core/engine/src/script.rs#L96
https://github.com/boa-dev/boa/blob/1f2893c5f61cec4fc0b7e7e2834099e62a33e9c9/core/engine/src/builtins/eval/mod.rs#L134

see [this commit](https://github.com/boa-dev/boa/pull/4038/commits/1f2893c5f61cec4fc0b7e7e2834099e62a33e9c9)

Nikita-str:
If remove `Option<SourceText>` from everywhere then I don't like what will happen with `ByteCompiler::new` / `FunctionCompiler::new` in no-source cases: we should have non-init marker, and create an empty `Vec` looks not safe enough. Option directly say that there is no object. But with `Option` we can forget to init it... I can add `Option<SourceText>` as `fn new`'s  arg and there will be no way to forget init.

core/engine/src/spanned_source_text.rs
raskad: Now that I looked again, and with your new changes, we have:

`SpannedSourceText` on `CodeBlock`,
`SpannedSourceText` and `SourceText` on `ByteCompiler`,
`SourceText` and `LinearSpan` on `FunctionCompiler`

How about we just have `SpannedSourceText` on `ByteCompiler` and `FunctionCompiler` and just have a method that returns `SourceText` from `SpannedSourceText` that we use when we create a new comiler?

> And also do we always have an outer object that contains corresponding SourceTextInner?

For now I would just say let's not worry about that. We can just make `SpannedSourceText` `pub(crate)` and when we find a case where we need to make it `pub` later, we can think about it.

raskad:
> If remove Option<SourceText> from everywhere then I don't like what will happen with ByteCompiler::new / FunctionCompiler::new in no-source cases: we should have non-init marker, and create an empty Vec looks not safe enough. Option directly say that there is no object. But with Option we can forget to init it... I can add Option<SourceText> as fn new's arg and there will be no way to forget init.

I was mostly concerned with the code in the bytecompiler being a bit easier to work with. Like you mentioned, with the `Option` and the set method you always have to remember it and do the whole `if let` check.
I don't see how the empty vec is unsafe, but what do you think about changing `Gc<Inner>` in `SourceText` to `Option<Gc<Inner>>`, or alternativeley, if you implement https://github.com/boa-dev/boa/pull/4038#discussion_r1908072530, `SpannedSourceText` would be the only struct that is actually used outside of `spanned_source_text.rs`. That way you could also do the `Option` directly there.
If I'm not missing anything that should be a good solution, because we do not have the `Option`s everywhere in the bytecompiler, the type is still the same size and in the case where we have no source, we do not allocate an empty `Vec` or a `Gc`.

core/ast/src/source.rs
raskad: Nice solution!

Nikita-str:
Seems like all done.
I also pass `SourceText` to the `Module` & then pass it from the `Module` to the `ByteCompiler`, but I don't test it.

raskad:
Looks really nice! Thank you for the work! Could you do a rebase, best would be after #4117 is merged, because the CI is broken without that.

One very minor nitpick: I would like to rename `SourceText::new_pseudo` to `SourceText::new_empty` and `PSEUDO_LINEAR_SPAN` to `EMPTY_LINEAR_SPAN`. I think it describes the values a bit better.

raskad (approved):
Thank you for the contribution. Great stuff. I'd be happy to merge this.

jedel1043:
How much progress is this work towards errors with backtraces? Haven't looked at it yet, I'm just curious if this would be another step forward or if we will have to essentially rewrite everything when we have nicer errors

Nikita-str:
> How much progress is this towards errors with backtraces?

Sorry, I actually don't sure what is needed for backtraced errors.
In this PR propagated only those linear spans that are needs for functions' `toString`.
`LinearSpan`s are only needed when you need unchanged source code.
And I  don't sure, do you need unchanged source code for backtraced errors at all?
If you need then at least it will be necessary to add `LinearSpan`s to others AST nodes.
Otherwise it will be just parallel things, I suppose that you will not have to rewrite anything in any case.

jedel1043:
If we just need to increasingly add `LinearSpan` to the rest of the AST nodes, that sounds perfect! And yeah, we need unchanged source code, because even after optimization passes we need to be able to reference the original line that a bytecode instruction comes from.

PR #4047, opened 2024-11-19
bug fix: ops that stay strictly at the EOF after assigns ops are ignored

### LINKED ISSUE #3330: Slashes after assignments at the end of files are allowed
**Describe the bug**
If code in the form `<identifier> <assignment operator> <identifier> "/" <eof>` is passed to Boa, the code executes as if the "/" were not present; it should raise a SyntaxError.

This bug does not work if there are characters after the "/", including newlines or spaces. The "/" must be the last character in the input.

**To Reproduce**
In your shell, assuming GNU coreutils: `echo -n 'let a, b; a = b /' > test.js; boa test.js`
Note that `echo -n` is required to avoid a trailing newline.

**Expected behavior**
This code should throw a `SyntaxError`. The `/` character is unexpected and is not valid syntax.

- OS: Artix Linux
- Version: rolling
- Target triple: `x86_64-unknown-linux-gnu`
- Rustc version: `rustc 1.72.1 (d5c2e9c34 2023-09-13)`

**Additional context**
This bug works with most if not all of the assignment operators. For example, neither `a /= b /` nor `a += b /` raise a `SyntaxError` despite being invalid JavaScript.

Nikita-str:
This Pull Request closes #3330 

The reason of the bug was that lexer decides next Token sequence {`...`, `OP`, `EOF`} as invalid and return an error. Yes, it's invalid, but in the same time lexer cannot know that {`...`, `OP`, `SPACE`, `EOF`} is also invalid and therefore there was two different logic to catch first case.

So, for me it seems more logically correct to delegate the catching of invalid ops at the end to the parser

-----

Should I:
* remove all `Ok(...)` from `core/parser/src/lexer/operator.rs` (now there exist only pseudo `Err(..)` that needs only for type inference)?
* use strum::EnumIter to remove enumeration of almost all `Punctuator`  in `core/parser/src/lexer/tests.rs`?

-----

Hmm... now I understand that I don't know the place where errors from lexer don't propagate upwards before, seems like somewhere not used `?`
Do I need to investigate where the place?

### DIFF
--- a/core/engine/src/tests/operators.rs
+++ b/core/engine/src/tests/operators.rs
@@ -600,6 +600,44 @@ fn delete_in_strict_function_returned() {
     )]);
 }
 
+#[test]
+fn ops_at_the_end() {
+    let msg = "abrupt end";
+
+    let mut actions = vec![TestAction::assert_eq("var a, b=3; a = b ++", 3)];
+
+    let abrupt_op_sources = [
+        // there was a bug with different behavior with and without space at the end;
+        // so few lines are almost the same except for ending space
+        "var a, b=3; a = b **",
+        "var a, b=3; a = b ** ",
+        "var a, b=3; a = b *",
+        "var a, b=3; a = b * ",
+        "var a, b=3; a /= b *",
+        "var a, b=3; a /= b * ",
+        "var a, b=3; a = b /",
+        "var a, b=3; a = b / ",
+        "var a, b=3; a = b +",
+        "var a, b=3; a = b -",
+        "var a, b=3; a = b ||",
+        "var a, b=3; a = b || ",
+        "var a, b=3; a = b ==",
+        "var a, b=3; a = b ===",
+    ];
+
+    for source in abrupt_op_sources {
+        actions.push(TestAction::assert_native_error(
+            source,
+            JsNativeErrorKind::Syntax,
+            msg,
+        ));
+    }
+
+    actions.push(TestAction::assert_eq("var a, b=3; a = b --", 3));
+
+    run_test_actions(actions);
+}
+
 mod in_operator {
     use super::*;
 
--- a/core/parser/src/lexer/mod.rs
+++ b/core/parser/src/lexer/mod.rs
@@ -169,9 +169,9 @@ impl<R> Lexer<R> {
                 }
             }
         } else {
-            Err(Error::syntax(
-                "Abrupt end: Expecting Token /,*,= or regex",
-                start,
+            Ok(Token::new(
+                Punctuator::Div.into(),
+                Span::new(start, self.cursor.pos()),
             ))
         }
     }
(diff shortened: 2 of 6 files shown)


raskad:
Hi @Nikita-str thank you for investigating this. I think the tests look good like this.

> Hmm... now I understand that I don't know the place where errors from lexer don't propagate upwards before, seems like somewhere not used ?
Do I need to investigate where the place?

I did a quick check and I'm pretty sure that the place is here: https://github.com/boa-dev/boa/blob/97c44239a4edadcaec086fbd28234ebec86eefcc/core/parser/src/parser/expression/assignment/mod.rs#L110-L111
The error is ignored, because we only want to check if an arrow function follows and there may be nothing there. If you can figure out a way to propagate the error there too with your fix, or rewrite that logic, that would be great.

Nikita-str:
@raskad 
> I did a quick check and I'm pretty sure that the place is here: ...

Yeah, it is.
And just fix this line can resolve the issue, but as I say in the PR comment, it seems to me more correct to delegate catching of the error to parser. But maybe I'm wrong, so it's up for discussion.

---

Also, the problem here is that if we ignore some errors from lexer, like one that removed now from `core/parser/src/lexer/operator.rs` or `invalid regular expression flag`, the position of cursor will changed and we will get something new from `peek`/`next`.

In cur case on first call it is `Err(Lex { err: Syntax("abrupt end - could not preview next value as part of the operator", ...)})` and then (if we ignore the error) it will `Ok(None)`. This error always stay on the EOF, but with regex we can just ignore invalid regex in case of ignoring.

Thus maybe there is sense (not within this PR) to save an error in ?Lexer? and repeatedly return it?
(But you can not just clone an Error because of `IO(io::Error)`)

On the other hand, ignoring an error from lexer is error, but as you can see sometimes it's not noticeable.

raskad:
> On the other hand, ignoring an error from lexer is error, but as you can see sometimes it's not noticeable.

I would say this is definitley the problem here. We should probably think about the lexer and the parser as two separate entities. For example, they probably could be different crates. From that perspective, there should be no logic that is spread across both (like holding on to an error in the lexer).

---

Can you rebase the branch, there are some conflicts now that prevent the CI from running. Did you get a chance to run the 262 tests locally with your latest commit? I would expect some tests to fail, because arrow functions would now be parsed with line terminators between the identifier and the arrow. Or am I missing something?

Nikita-str:
Yes, you are right, there was some regression on test262, my bad.
Fixed

raskad:
Looks very good, great work!

Now that everything works as expected, I think you can remove the `#[allow(unreachable_patterns)]` and the try branches, correct? Should be possible to remove the `?` where the `vop!` macro is called in the `op!` macro. Then all the `Ok` wrapping should be uneccessary. Then we can merge this.

raskad (approved):
Looks great, thank you for finding the issue :)

jedel1043 (approved):

PR #4048, opened 2024-11-19
Bug fix: regex started with `/=` parsed as `AssignDiv`

### LINKED ISSUE #3329: RegExp /=/ not parsed as such in certain locations
**Describe the bug**
The code `() => /=/` (and certain other uses of the RegExp literal `/=/`) raises a SyntaxError.

**To Reproduce**
Execute the following JavaScript:
```javascript
() => /=/
```

**Expected behavior**
This code should be parsed as an arrow function expression with no arguments that returns a RegExp that matches the character `=`. (semantically equivalent to `() => new RegExp("=")`)

- OS: Artix Linux
- Version: rolling
- Target triple: `x86_64-unknown-linux-gnu`
- Rustc version: `rustc 1.72.1 (d5c2e9c34 2023-09-13)`

**Additional context**
Notably, `return /=/` inside of an arrow function errors, but not inside of a `function` function.
```javascript
() => {return /=/} // error
(function() {return /=/}) // fine!
```
Other uses of the literal `/=/` that error include:
```javascript
let a = 5; /=/; // should construct RegExp and set a to 5, instead errors
+/=/ // should evaluate to NaN
class A extends /=/ {} // should error out because /=/ is not a constructor, not because of syntax
```

Nikita-str:
This Pull Request closes #3329

### DIFF
--- a/core/engine/src/tests/operators.rs
+++ b/core/engine/src/tests/operators.rs
@@ -600,6 +600,15 @@ fn delete_in_strict_function_returned() {
     )]);
 }
 
+#[test]
+fn regex_slash_eq() {
+    run_test_actions([
+        TestAction::assert_eq("+/=/", JsValue::nan()),
+        TestAction::assert_eq("var a = 5; /=/; a", 5),
+        TestAction::assert_eq("x = () => /=/;\n\"a=b\".match(x())[0]", js_str!("=")),
+    ]);
+}
+
 mod in_operator {
     use super::*;
 
--- a/core/engine/src/tests/class.rs
+++ b/core/engine/src/tests/class.rs
@@ -36,3 +36,12 @@ fn class_field_initializer_name() {
         TestAction::assert_eq("c.c.name", js_str!("#c")),
     ]);
 }
+
+#[test]
+fn class_superclass_from_regex_error() {
+    run_test_actions([TestAction::assert_native_error(
+        "class A extends /=/ {}",
+        crate::JsNativeErrorKind::Type,
+        "superclass must be a constructor",
+    )]);
+}
--- a/core/parser/src/lexer/mod.rs
+++ b/core/parser/src/lexer/mod.rs
@@ -112,17 +112,20 @@ impl<R> Lexer<R> {
         }
     }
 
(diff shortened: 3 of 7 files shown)


core/engine/src/tests/operators.rs
```diff
@@ -600,6 +600,15 @@ fn delete_in_strict_function_returned() {
     )]);
 }
 
+#[test]
+fn regex_slash_eq() {
+    run_test_actions([
+        TestAction::assert_eq("+/=/", JsValue::nan()),
+        TestAction::assert_eq("var a = 5; /=/; a", 5),
+        TestAction::assert_eq("x = () => /=/;\n\"a=b\".match(x())[0]", js_str!("=")),
```
raskad, line 608: Can you expand on the `\n` here`? Is there a specific reason for it?
When I try to run `x = () => /=/;"a=b".match(x())[0]` on the current main branch I get a syntax error, in node I get `=` as expected.

raskad (approved):
Nice fix, I just have one question about one test.

core/engine/src/tests/operators.rs
Nikita-str: The syntax error is not expected here:
`/=/` is a valid regex;
so `x = () => /=/` is also valid as arrow function that returns mentioned regex;
so `x = () => /=/;"a=b".match(x())[0]` should to create arrow fn `x` that return regex and then call for `"a=b"` match function with `x()` arg that will be the same as `"a=b".match(/=/)` that return `["="]` and `[0]` for it will return `"="`.
So `x = () => /=/;"a=b".match(x())[0]` also should return `"="`

raskad: Yeah that's all clear, I was just wondering about the line break `\n`.

Nikita-str: Oh sorry, I misunderstood your question.

No, there is no special reason.
The test have such form because firstly I checked bug in js file, and this was on two diff lines so seems like I just moved new line with the test from js.

jedel1043 (approved):
Looks great! Thank you!

PR #4068, opened 2024-12-07
Build out Temporal's `ZonedDateTime` and `Now`

nekevss:
This Pull Request is related to the ongoing work for #1804

It changes the following:

- Switches Boa to the most recent revision of `temporal_rs`
- Adds a `tz_provider` field to `Context` using `FsTzdbProvider` to provide initial time zone support
- Builds out the some of the initial accessor properties and some methods for `ZonedDateTime`
- Builds out a couple methods for `Now`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -852,6 +852,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "combine"
+version = "4.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba5a308b75df32fe02788e748662718f03fde005016435c444eea572398219fd"
+dependencies = [
+ "bytes",
+ "memchr",
+]
+
+[[package]]
 name = "comfy-table"
 version = "7.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


nekevss:
Opened a PR to hopefully fix the compilation issues of this PR in boa-dev/temporal#125

jedel1043:
Nice! Did you investigate the test regressions we have?

nekevss:
I didn't dig too deep into test regression as I'm fairly certain glancing at the broken tests that they were more so false positives beforehand. But I can definitely double check.

nekevss:
Fixed the zonedDateTimeISO misspelled name. The other tests are probably somehow a regression in `temporal_rs`. Although, at a quick glance, I don't think this code was updated over the last few PRs, but I have to double check.

jedel1043:
I'd also suggest updating the test262 commit, since I think some tests could have been adjusted since our last bump.

nekevss:
So after looking at it, I'm fairly certain the remaining broken tests (prior to the test262 bump) are definitely something broken, but it's not with this PR. I'm planning on looking into it and fixing it on the `temporal_rs` side

jedel1043 (approved):
Looks good! We'll have to investigate the regexp tests, but since this didn't touch anything related, they must be updates of old tests.

raskad (approved):

PR #4079, opened 2024-12-09
Add valueOf methods to the Temporal builtins

nekevss:
This Pull Request is related to the general work on #1804

It changes the following:

- Adds or adjusts the error message for the `prototype.valueOf` method implementation to the Temporal builtins

### DIFF
--- a/core/engine/src/builtins/temporal/duration/mod.rs
+++ b/core/engine/src/builtins/temporal/duration/mod.rs
@@ -798,7 +798,7 @@ impl Duration {
 
     pub(crate) fn value_of(_this: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
         Err(JsNativeError::typ()
-            .with_message("valueOf not implemented for Temporal objects. See 'compare', 'equals', or `toString`")
+            .with_message("`valueOf` not supported by Temporal built-ins. See 'compare', 'equals', or `toString`")
             .into())
     }
 }
--- a/core/engine/src/builtins/temporal/duration/mod.rs
+++ b/core/engine/src/builtins/temporal/duration/mod.rs
@@ -191,6 +191,7 @@ impl IntrinsicObject for Duration {
             .method(Self::total, js_string!("total"), 1)
             .method(Self::to_string, js_string!("toString"), 1)
             .method(Self::to_json, js_string!("toJSON"), 0)
+            .method(Self::value_of, js_string!("valueOf"), 0)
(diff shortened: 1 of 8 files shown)


nekevss:
> adding a proper `valueOf` implementation?

What do you mean by proper `valueOf` implementation?

There's not a specific issue open, but there is a host of tracking issues that this would complete the Boa implementation for those issues. I believe this should be to spec for the [Temporal proposal](https://tc39.es/proposal-temporal/#sec-temporal.plaindate.prototype.valueof).

core/engine/src/builtins/temporal/instant/mod.rs
```diff
@@ -532,6 +533,12 @@ impl Instant {
             .with_message("not yet implemented.")
             .into())
     }
+
+    pub(crate) fn value_of(_this: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
+        Err(JsNativeError::error()
```
raskad, line 538: Should the error type not be `JsNativeError::typ`? The spec says so at least https://tc39.es/proposal-temporal/#sec-temporal.instant.prototype.valueof

nekevss: Yes, it should be. Huh, I could've sworn I typed `typ`. I'll switch those over.

hansl:
> There's not a specific issue open, but there is a host of tracking issues that this would complete the Boa implementation for those issues. I believe this should be to spec for the [Temporal proposal](https://tc39.es/proposal-temporal/#sec-temporal.plaindate.prototype.valueof).

Don't mind me, I was not looking at the right spec.

core/engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -794,6 +795,12 @@ impl Duration {
             .with_message("not yet implemented.")
             .into())
     }
+
+    pub(crate) fn value_of(_this: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
+        Err(JsNativeError::typ()
+            .with_message("valueOf not implemented for Temporal objects. See 'compare', 'equals', or `toString`")
```
jedel1043, line 801: Hmm, I'm a bit negative about the "not implemented" phrasing because it seems like we haven't implemented it yet. Maybe something like "invalid call to `valueOf` for Temporal object".

jedel1043 (approved):
LGTM!

PR #4080, opened 2024-12-10
Privatize `JsValue`'s internals and expose it through a JsVariant (with immutable reference)

hansl:
Move `JsValue` to being a struct with an internal enumeration. This prevents people from deconstructing or constructing values directly, instead they should rely on helper methods (like `as_*()`) or using the `JsVariant` which contains references to the internal value.

The direction here is that we will change the interior representation of JsValue in the near future and this new representation will not be compatible with enum deconstruction / pattern matching.

There is no deprecation strategy here; every PR that add usage of `JsValue` as an enum will have a merge conflict or a test error. Every dependents doing the same will break when they update.

Fixes #3761.

PR #4081, opened 2024-12-11
Update Temporal ToIntegerIfIntegral, ToIntegerWithTruncation, and ToPositiveIntegerWithTruncation implementation 

nekevss:
Posting this as a draft to hopefully get comments.

TLDR: I'm convinced the way we currently handle converting JsValue to integers in the Temporal implementation is wrong and this is a way to update it.

## Baseline issue

Currently, we are using `to_integer_with_truncation`, `to_positive_integer_with_truncation` and `to_integer_if_integral` to convert values into `i32`. These methods are still hold over from the initial prototype of Temporal prior to `temporal_rs` being split off. At the time, I think it was ultimately an optimistic approach, but I'm now fairly certain it's wrong long term and should probably addressed (hence the draft PR)  ASAP.

Also, every argument being truncated to `i32` negatively impacts `temporal_rs`'s API by forcing us to keep the values at `i32`. This change would allow us to change `temporal_rs`'s API, and move it away from only `i32`.

## General approach / Order of Operations

So the general approach of this PR is to first get the value and transform it to a `FiniteF64` by calling `ToNumber()` and checking that it's finite, which was essentially all the previous functions were doing anyways besides the truncation.

Once the values are made `FiniteF64`s, we then get the options object and determine the `ArithmeticOverflow` from the options object. This is needed for the observability tests which may check for the finite check throwing prior to a wrong options object throwing.

After the options object, we then truncate the value into the integer type that is expected by `temporal_rs` with the user defined `ArithmeticOverflow`. If ArithmeticOverflow::Reject and the value is not contained within T::MIN and T::MAX, we throw an early far exceeds range RangeError. If ArithemeticOverflow::Constrain, then we clamp the value to T::MAX or T::MIN.

There is also an argument to just always silently constrain the value to T::MAX or T::MIN and let `temporal_rs` handle the error. I think I may like this option more, because it does allow us to truncate early without `ArithmeticOverflow` and deferring truncation to after reading options.

Anyways, general thoughts would be appreciated. Technically, this is probably more of a `temporal_rs` leaning PR as I think the current `truncate` function in this PR should live in `temporal_rs`, but it does handle specifically the Boa's glue code to the library.

### DIFF
--- a/core/engine/src/builtins/temporal/duration/mod.rs
+++ b/core/engine/src/builtins/temporal/duration/mod.rs
@@ -23,7 +23,6 @@ use boa_profiler::Profiler;
 use temporal_rs::{
     options::{RelativeTo, RoundingIncrement, RoundingOptions, TemporalRoundingMode, TemporalUnit},
     partial::PartialDuration,
-    primitive::FiniteF64,
     Duration as InnerDuration,
 };
 
@@ -221,6 +220,7 @@ impl BuiltInConstructor for Duration {
                 .into());
         }
 
+        // TOOD: Support conversion to i64
         // 2. If years is undefined, let y be 0; else let y be ? ToIntegerIfIntegral(years).
         let years = args
             .get_or_undefined(0)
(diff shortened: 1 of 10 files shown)


raskad:
I'm all for having the more specific / representative types in the `temporal_rs` API.

> There is also an argument to just always silently constrain the value to T::MAX or T::MIN and let temporal_rs handle the error. I think I may like this option more, because it does allow us to truncate early without ArithmeticOverflow and deferring truncation to after reading options.

I'm not 100% updated on how the current APIs look, but I think instead of silently doing this, there should be helper APIs / types provided by `temporal_rs` that deal with this most of the way.
I would imagine a worflow like this:

1. (boa) Get values.
2. (temporal_rs helper) Transform values into `FiniteF64`.
3. (boa) Get options.
4. (temporal_rs helper) Transform `FiniteF64` values together with options into the expected `temporal_rs` API types.
5. (temporal_rs) "Real" temporal_rs API call with the fitting types.

Does that make sense?

nekevss:
Definitely agreed! 😄 That's why boa-dev/temporal#131 exists and this is still a draft. The helper functions were primarily to double check the approach was working how I thought it would be in Boa before committing to it.

Right now, I'm mostly waiting for reviews on the temporal_rs PR and then this can be updated with the change and switched off draft.

nekevss:
After looking at the broken tests, one appears to be a regression (case insensitive calendar ids) with a PR submitted, one is a false positive that will require a patch to remove MINUS support from `ixdtf`, and a third (duration/out-of-range) is most likely a false positive that will need to be tested and potentially updated in `temporal_rs`.

core/engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -221,67 +221,98 @@ impl BuiltInConstructor for Duration {
                 .into());
         }
 
+        // TOOD: Support conversion to i64
```
raskad, line 224: This TODO is now done, correct?

core/engine/src/builtins/temporal/duration/mod.rs
```diff
@@ -928,119 +959,162 @@ pub(crate) fn to_temporal_partial_duration(
...
     // 4. Let days be ? Get(temporalDurationLike, "days").
-    let days = unknown_object.get(js_string!("days"), context)?;
-    if !days.is_undefined() {
-        // 5. If days is not undefined, set result.[[Days]] to ? ToIntegerIfIntegral(days).
-        let _ = result
-            .days
-            .insert(FiniteF64::from(to_integer_if_integral(&days, context)?));
-    }
+    // 5. If days is not undefined, set result.[[Days]] to ? ToIntegerIfIntegral(days).
+    // TODO: Increase to i64
```
raskad, line 965: Same here?

raskad (approved):
Looks good!

core/engine/src/builtins/temporal/plain_date/mod.rs
```diff
@@ -958,47 +901,61 @@ pub(crate) fn to_partial_date_record(
...
+        })
         .transpose()?;
     let era = partial_object
         .get(js_string!("era"), context)?
         .map(|v| {
-            let JsValue::String(era) =
-                v.to_primitive(context, crate::value::PreferredType::String)?
-            else {
+            let v = v.to_primitive(context, crate::value::PreferredType::String)?;
+            let JsValue::String(era) = v else {
```
hansl, line 951: Here and everywhere; we're moving to using `as_string()` instead of pattern matching on JsValue. Could you fix this? I've fixed `main` in my PR but I'd like to not have to play catch up with everyone.

See #4080

raskad (approved):

jedel1043 (approved):
Wow, the new APIs do look a lot nicer! Nice cleanup!

PR #4082, opened 2024-12-11
Allow resizing of underlying ArrayBuffer from Rust

hansl:
I have an API that might truncate an `ArrayBuffer` passed in (or make it bigger). I have to call `resize()` from Rust for a buffer passed by JavaScript, and this is the best way I found to do that.


jedel1043:
Does it make sense to merge this after the `usize` change gets merged? I think it would be easier to rebase this than to rebase the other PR which is a lot bigger.

hansl:
Sure I'll wait for that PR to be merged then rebase this one.

hansl:
@jedel1043 Please review and ready for merge :) The other PR is back to draft for the moment.

jedel1043 (commented):
Why not call `resize` from here? It would remove a lot of duplication.

core/engine/src/builtins/array_buffer/mod.rs
```diff
@@ -554,7 +585,7 @@ impl ArrayBuffer {
     /// [`ArrayBuffer.prototype.resize ( newLength )`][spec].
     ///
     /// [spec]: https://tc39.es/ecma262/#sec-arraybuffer.prototype.resize
-    pub(crate) fn resize(
+    pub(crate) fn js_resize(
```
jedel1043, line 588: Why not call `resize` from here? It would remove a lot of duplication.

hansl: I didn't want to change the behaviour of JS code.

jedel1043 (approved):
Looks good!

nekevss (approved):

PR #4083, opened 2024-12-11
Replace all u64/i64 for length, index or offsets into usize/isize

hansl:
On platforms that are 32-bits (or 128?) this should make array and iterations faster in general. On platforms that are 64-bits, this will have no effect.

### DIFF
--- a/core/engine/src/value/mod.rs
+++ b/core/engine/src/value/mod.rs
@@ -865,7 +865,7 @@ impl JsValue {
     /// Converts a value to a non-negative integer if it is a valid integer index value.
     ///
     /// See: <https://tc39.es/ecma262/#sec-toindex>
-    pub fn to_index(&self, context: &mut Context) -> JsResult<u64> {
+    pub fn to_index(&self, context: &mut Context) -> JsResult<usize> {
         // 1. If value is undefined, then
         if self.is_undefined() {
             // a. Return 0.
@@ -890,19 +890,19 @@ impl JsValue {
         debug_assert!(0 <= clamped && clamped <= Number::MAX_SAFE_INTEGER as i64);
 
         // e. Return integer.
-        Ok(clamped as u64)
+        Ok(clamped as usize)
     }
 
     /// Converts argument to an integer suitable for use as the length of an array-like object.
     ///
     /// See: <https://tc39.es/ecma262/#sec-tolength>
-    pub fn to_length(&self, context: &mut Context) -> JsResult<u64> {
+    pub fn to_length(&self, context: &mut Context) -> JsResult<usize> {
         // 1. Let len be ? ToInteger(argument).
         // 2. If len ≤ +0, return +0.
         // 3. Return min(len, 2^53 - 1).
         Ok(self
             .to_integer_or_infinity(context)?
-            .clamp_finite(0, Number::MAX_SAFE_INTEGER as i64) as u64)
+            .clamp_finite(0, Number::MAX_SAFE_INTEGER as i64) as usize)
     }
 
     /// Abstract operation `ToIntegerOrInfinity ( argument )`
--- a/cli/src/debug/limits.rs
+++ b/cli/src/debug/limits.rs
@@ -2,7 +2,7 @@ use boa_engine::{
     js_string,
     object::{FunctionObjectBuilder, ObjectInitializer},
     property::Attribute,
-    Context, JsArgs, JsNativeError, JsObject, JsResult, JsValue, NativeFunction,
+    Context, JsArgs, JsObject, JsResult, JsValue, NativeFunction,
 };
 
 fn get_loop(_: &JsValue, _: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
@@ -12,7 +12,9 @@ fn get_loop(_: &JsValue, _: &[JsValue], context: &mut Context) -> JsResult<JsVal
 
 fn set_loop(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
     let value = args.get_or_undefined(0).to_length(context)?;
-    context.runtime_limits_mut().set_loop_iteration_limit(value);
+    context
+        .runtime_limits_mut()
+        .set_loop_iteration_limit(value as u64);
     Ok(JsValue::undefined())
 }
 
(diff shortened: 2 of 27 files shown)


jedel1043:
  
  | Total | 48,625 | 48,625 | 0 |
  | Passed | 43,616 | 43,607 | **-9** |
  | Ignored | 1,471 | 1,471 | 0 |
  | Failed | 3,538 | 3,547 | **+9** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 89.70% | 89.68% | **-0.02%** |
  
  <details><summary><b>Broken tests (9):</b></summary>
  
  ```
  test/built-ins/TypedArray/prototype/lastIndexOf/search-not-found-returns-minus-one.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/tointeger-fromindex.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/search-found-returns-index.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/fromIndex-minus-zero.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/resizable-buffer.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/BigInt/search-not-found-returns-minus-one.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/BigInt/tointeger-fromindex.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/BigInt/search-found-returns-index.js (previously Passed)
  test/built-ins/TypedArray/prototype/lastIndexOf/BigInt/fromIndex-minus-zero.js (previously Passed)
  ```
  </details>

jedel1043:
Hmm, there were some test regressions. Can you investigate them? Maybe it was just a typo somewhere.

hansl:
@jedel1043 I don't know how to manually re-run the test262 on this project, so I'll leave that to you as a reviewer. Thanks!

jedel1043:
@hansl tests already ran, it's just on the output of the "Write a new comment" step for the Test262 action. We really need to setup something so that this thing works on external repos 😅

jedel1043:
  
  | Total | 48,625 | 48,625 | 0 |
  | Passed | 43,616 | 43,616 | 0 |
  | Ignored | 1,471 | 1,471 | 0 |
  | Failed | 3,538 | 3,538 | 0 |
  | Panics | 0 | 0 | 0 |
  | Conformance | 89.70% | 89.70% | 0.00% |

jedel1043 (approved):
Looks great!

Approving this, but can you open an issue to add CI tests for a 32-bit architecture? I'd really like to potentially test that just to avoid introducing platform-dependent regressions.

hansl:
Sounds good. Let me test it on my embedded platform before merging. Will have to wait to Monday.

hansl:
Well that didn't work well...

CleanShot 2024-12-13 at 19 04 08@2x

I'll investigate on Monday (or this weekend if I have some time).

hansl:
At least 3 tests are failing in `boa_engine`, will investigate Monday.

CleanShot 2024-12-13 at 19 12 58@2x

core/engine/src/value/mod.rs
```diff
@@ -890,19 +890,19 @@ impl JsValue {
...
     }
 
     /// Converts argument to an integer suitable for use as the length of an array-like object.
     ///
     /// See: <https://tc39.es/ecma262/#sec-tolength>
-    pub fn to_length(&self, context: &mut Context) -> JsResult<u64> {
+    pub fn to_length(&self, context: &mut Context) -> JsResult<usize> {
         // 1. Let len be ? ToInteger(argument).
         // 2. If len ≤ +0, return +0.
         // 3. Return min(len, 2^53 - 1).
```
HalidOdat, line 902: We used to have it be `usize`, but we changed it because on `32-bit` systems it does not capture the full range that a JavaScript length can be. (for arrays it's fine to put it as `usize` since it's between `0..2^32-1`, but for strings and typed arrays, the upper bound is `2^53-1` (max safe representable integer in a `f64`).

Setting it to `usize` truncates the values on 32-bit platforms giving wrong results.

HalidOdat: For example, one place that `to_length` is used is [`String.padEnd()`](https://github.com/boa-dev/boa/blob/e529d702ec5cc869a08be7486b82e06d23d4466a/core/engine/src/builtins/string/mod.rs#L1524) on the `maxLength` parameter, putting `2^32-1` would work on 32-bit platform anything greater will be truncated (i.e. `2^32` will be truncated to `0`), on 64-bit platform this would be fine.

hansl: MAX_STRING_LENGTH is 3^32-1.

From the MDN:
> The language specification requires strings to have a maximum length of 253 - 1 elements, which is the upper limit for precise integers. However, a string with this length needs 16384TiB of storage, which cannot fit in any reasonable device's memory, so implementations tend to lower the threshold, which allows the string's length to be conveniently stored in a 32-bit integer.
>
> In V8 (used by Chrome and Node), the maximum length is 229 - 24 (~1GiB). On 32-bit systems, the maximum length is 228 - 16 (~512MiB).
> In Firefox, the maximum length is 230 - 2 (~2GiB). Before Firefox 65, the maximum length was 228 - 1 (~512MiB).
> In Safari, the maximum length is 231 - 1 (~4GiB).

HalidOdat: The issue isn't with allocation of the strings, rather that it breaks the math and checks in many places.

hansl: That's all fine. If there's a bug with math being broken, then we need to fix it.

I would assume test262 and/or our tests would find any discrepancies, and I'd consider any issue a blocker for this PR. I'm currently running the full test262 suite on a Rasp Pi 2 equivalent. Everything works locally on my aarch64 Mac.

hansl: If you can provide me with a valid JavaScript snippet that works on other implementations but doesn't work on this PR, I'll be more than happy to add it to tests and fix this PR.

hansl:
Okay so after running test262 a bunch, and testing in browsers and Deno, I would still like to build this PR the right way but will need to move to a different approach:

- isize/usize will be used internally for sizes, indices and offsets of strings.
- i64/u64 will be used for array.

This should still result in performance improvements on 32-bits platforms (e.g. in wasm32).

I may type the new sizes `StringUSize` / `StringISize` and `ArrayUSize` / `ArrayISize` so it becomes more clear in the code which we're dealing with.

For now this PR will be back to draft.

hansl:
If we ever want to do this, we should reopen a PR.

PR #4087, opened 2024-12-15
Skip creation of arguments object if possible

raskad:

- Skip creation of arguments object if possible
-
-

### DIFF
--- a/core/ast/src/scope.rs
+++ b/core/ast/src/scope.rs
@@ -18,6 +18,7 @@ struct Binding {
     lex: bool,
     strict: bool,
     escapes: bool,
+    accessed: bool,
 }
 
 /// A scope maps bound identifiers to their binding positions.
@@ -259,6 +260,7 @@ impl Scope {
                 .iter_mut()
                 .find(|b| &b.name == name)
             {
+                binding.accessed = true;
                 if crossed_function_border || eval_or_with {
                     binding.escapes = true;
                 }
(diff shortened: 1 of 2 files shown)


core/ast/src/scope.rs
```diff
@@ -259,6 +260,7 @@ impl Scope {
                 .iter_mut()
                 .find(|b| &b.name == name)
             {
+                binding.accessed = true;
```
hansl, line 263: Sorry I'm not familiar with this part of the code.

What if I pass the arguments object to an underlying function? Is that considered accessing bindings?

```js
function x(otherArgs) {
  console.log(otherArgs.1);
}

function y() {
  x(arguments);
}

y(3.141592);  // Will this print the right thing?

raskad: > What if I pass the arguments object to an underlying function? Is that considered accessing bindings? Or if I set a global variable to it?

Yes, that will be a binding access. The way it works is, that we search for a binding and mark it as `accessed` every time we see a binding identifier in the AST. See the visitor in the scope analyzer here: https://github.com/boa-dev/boa/blob/e529d702ec5cc869a08be7486b82e06d23d4466a/core/ast/src/scope_analyzer.rs#L88-L93

hansl (approved):
LGTM! Thanks.

hansl (approved):

jedel1043 (approved):
Looks great!

PR #4090, opened 2024-12-18
Some string cleanups

CrazyboyQCD:
- Simplify `JsString::as_str` (miri passed,  decrease the size of `cli` executable by ~5kb under release mode (28,680kb => 28,675kb) )
- Replace soft deprecated `addr_of*` macro with `& [const|mut] raw`
- Mark some `&mut self` tagged methods const


jedel1043 (approved):
Thank you! Looks good!

PR #4091, opened 2024-12-20
Use NaN-boxing on value::InnerValue

### LINKED ISSUE #1373: NaN boxed JavaScript Value

hansl:
This adds a new `InnerValue` that uses NaN-boxing to allow `JsValue` to ultimately only take 64-bits (8 bytes). This allows `JsValue` to fit into one x86_64 register which should greatly improve performance.

For more details on NaN-boxing (and its alternative, Pointer tagging), see https://piotrduperas.com/posts/nan-boxing which is a great tutorial (in C) on how to get there. A proper explanation of each tags/range of values is described in the header of the `inner.rs` module.

Fixes #1373.


hansl:
All tests in `inner.rs` are passing. Unfortunately, running `cargo test -p boa_engine` currently results in seg faults: invalid memory reference. This will remain a draft until at least all tests pass.

core/engine/src/value/inner.rs
```diff
@@ -0,0 +1,886 @@
...
+impl Drop for InnerValue {
+    fn drop(&mut self) {
+        // Drop the pointer if it is a pointer.
+        unsafe {
+            NanBitTag::drop_pointer(self.0);
+        }
+    }
+}
+
+#[test]
```
hansl, line 620: I need to add a test for `as_variant()` on all these value tests.

raskad:
Still failing some tests:

| Total | 48,625 | 48,625 | 0 |
| Passed | 43,616 | 43,606 | **-10** |
| Ignored | 1,471 | 1,471 | 0 |
| Failed | 3,538 | 3,548 | **+10** |
| Panics | 0 | 0 | 0 |
| Conformance | 89.70% | 89.68% | **-0.02%** |

<details><summary><b>Broken tests (10):</b></summary>

```
test/built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation.js (previously Passed)
test/built-ins/TypedArray/prototype/set/typedarray-arg-set-values-diff-buffer-other-type-conversions.js (previously Passed)
test/built-ins/TypedArray/prototype/set/array-arg-src-tonumber-value-conversions.js (previously Passed)
test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations.js (previously Passed)
test/built-ins/Map/valid-keys.js (previously Passed)
test/built-ins/DataView/prototype/setFloat64/set-values-return-undefined.js (previously Passed)
test/built-ins/DataView/prototype/setFloat32/set-values-return-undefined.js (previously Passed)
test/built-ins/TypedArrayConstructors/ctors/object-arg/conversion-operation.js (previously Passed)
test/built-ins/TypedArrayConstructors/internals/DefineOwnProperty/conversion-operation.js (previously Passed)
test/built-ins/TypedArrayConstructors/internals/Set/conversion-operation.js (previously Passed)
```
</details>

But the bigger issue that I see is performance. I ran the benchmarks locally and the performance was worse in all benchmarks, but in some up to almost 40%. I though a performance tradeoff vs lower memory usage seemed reasonable and expected for this change, but this seems way to extreme.

Benchmarks before:

```txt
RESULT Richards 78.5
RESULT DeltaBlue 71.8
RESULT Crypto 80.3
RESULT RayTrace 182
RESULT EarleyBoyer 227
RESULT RegExp 43.7
RESULT Splay 326
RESULT NavierStokes 187
SCORE 122
```

After:

```txt
RESULT Richards 49.9
RESULT DeltaBlue 43.7
RESULT Crypto 77.7
RESULT RayTrace 138
RESULT EarleyBoyer 161
RESULT RegExp 38.9
RESULT Splay 214
RESULT NavierStokes 163
SCORE 92.0
```

jedel1043:
@raskad @hansl It is good to note that not all memory optimizations will yield performance gains. It may be that way sometimes, but there's this balance between memory and speed that big memory optimizations (such as this one) will have to deal with.

The first step is to make every test work, then we'd have to optimize repeatedly to recover all the lost performance, and hopefully we'll have an engine that's as fast as before but using a fraction of the memory.

Having said that, we should see what the overall gains in memory are, because this should at least save a considerable amount of memory to be worth putting more work on.

hansl:
There's no rush for this, since the actual breaking change already happened :)

I'll run benchmarks on my own, and I'll try to check a profiler for memory. In the next 2 weeks or so. It's also possible that I'm missing a few easy inline, and adding some bit magic might make things faster down the line.

First things first, I'll make sure test262 has no regressions.

hansl:
@raskad BTW this should cover both; it should take less memory, AND be more performant. But I guess we pass JsValue more by reference than by value. Now that we can move JsValue for free, maybe we should reconsider some other parts of hte code as well. Later work.

[This article](https://piotrduperas.com/posts/nan-boxing) shows a 28% increase in performance for a tagged union of pointers, in JavaScriptCore (Mozilla's engine). Tough TBF we weren't using pointers.

raskad:
@hansl I did a quick perf run on the `release-dbg` profile and noticed that there are a lot of operations relating to the `Box` that the `JsObject` is now wrapped in. For example in the function `JsObject::try_get` we call `self.clone().into()` on the `JsObject` to turn in into a `JsValue`. In my baseline the `into` call only makes up 0.5% of the function runtime but now with the alloc for the `Box` it takes 6.2%. Similarly the 2.3% time to drop that `JsValue` later then takes 8.4%.
I'm totally unsure if that would explain everything, since I cant really pin it down in the inverted call stack, but it might be a starting point.

jedel1043:
> @hansl I did a quick perf run on the `release-dbg` profile and noticed that there are a lot of operations relating to the `Box` that the `JsObject` is now wrapped in. For example in the function `JsObject::try_get` we call `self.clone().into()` on the `JsObject` to turn in into a `JsValue`. In my baseline the `into` call only makes up 0.5% of the function runtime but now with the alloc for the `Box` it takes 6.2%. Similarly the 2.3% time to drop that `JsValue` later then takes 8.4%.
> I'm totally unsure if that would explain everything, since I cant really pin it down in the inverted call stack, but it might be a starting point.

That makes a lot of sense actually, because clones for `JsValue`s containing `JsObject`s now aren't as cheap as just increasing a ref count

The "proper" solution would be to make `JsObject` 8 bytes again by using manual vtables instead of dyn objects (something very similar to what we do on `GcBox`).

hansl:
@raskad If we nan boxed `Gc` and `GcBox` instead of `Box`-ing this might go away, similar to what @jedel1043 is saying, since cloning will be essentially free. I need to put the value on the heap though, so just receiving and taking a pointer to a JsObject is not an option. 

Let me make sure we don't regress on test262 then I'll look into the performance.

hansl:
Okay this one is fascinating...

If I only create 1 value in an expression, e.g. `-0` or `[-0]` or `[[[[[-0]]]]]`, it keeps the parity just fine.

If I create multiple negative zeroes in an expression, only the first one keeps the parity. `[-0, -0, -0, -0]` becomes `[-0, 0, 0, 0]`... I'm looking for a place where we would cache these, but I can't seem to find one. I'm adding tests everywhere, but I'm flailing a bit.

Cloning the negative zero keeps the parity, copying is not supported, ... It seems the bug is not in my code, but that's the only thing I changed... Any idea?

hansl:
There's something funny with the parser and/or optimizer, but in any case `[-0, -0]` results in both values not taking the same code path. Found a bug anyway, so that's great. Fixed the bug, ready for reviews.

I'm still planning on running the benchmark before the end of the year.

hansl:
Latest on my laptop:

Before:
```text
RESULT Richards 107
RESULT DeltaBlue 103
RESULT Crypto 126
RESULT RayTrace 283
RESULT EarleyBoyer 343
RESULT RegExp 47.8
RESULT Splay 454
RESULT NavierStokes 295
SCORE 175
```

After:
```text
RESULT Richards 80.9
RESULT DeltaBlue 79.6
RESULT Crypto 128
RESULT RayTrace 231
RESULT EarleyBoyer 284
RESULT RegExp 45.8
RESULT Splay 369
RESULT NavierStokes 247
SCORE 148
```

So not as dramatic as @raskad's result (though I didn't run the benchmark at that commit). This is with the latest changes of using ranges and const bit masks.

hansl:
After the last fix which should fix everything, this is the latest benchmark:

```text
RESULT Richards 85.5
RESULT DeltaBlue 84.0
RESULT Crypto 131
RESULT RayTrace 243
RESULT EarleyBoyer 303
RESULT RegExp 47.2
RESULT Splay 368
RESULT NavierStokes 251
SCORE 153
```

So a bit better. The next steps would be to remove the extra deref of pointers (bigint, object, symbol and string), which I'll argue is out of scope for this PR.

I'll cleanup some more (some magic numbers still in), but this should be ready to review soon.

hansl:

| Total | 48,625 | 48,625 | 0 |
| Passed | 43,947 | 43,947 | 0 |
| Ignored | 1,471 | 1,471 | 0 |
| Failed | 3,207 | 3,207 | 0 |
| Panics | 0 | 0 | 0 |
| Conformance | 90.38% | 90.38% | 0.00% |

core/engine/src/value/mod.rs
```diff
@@ -351,34 +295,34 @@ impl JsValue {
...
-            InnerValue::Float64(rational) if rational == ((rational as i32) as f64) => {
+        if let Some(integer) = self.0.as_integer32() {
+            Some(integer)
+        } else if let Some(rational) = self.0.as_float64() {
+            // Use this poor-man's check as `[f64::fract]` isn't const.
+            if rational == ((rational as i32) as f64) {
                 Some(rational as i32)
+            } else {
+                None
             }
```
raskad, line 306: I think we could remove this branch now, right? I ran the 262 tests locally with the f64 part removed all tests passed.

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -0,0 +1,864 @@
...
+use core::fmt;
+use static_assertions::const_assert;
+
+/// Transform an `u64` into `f64`, by its bytes. This is necessary for
+/// keeping the MSRV at 1.82, as `f64::from_bits` is not const until
+/// 1.83.
+#[inline]
+const fn f64_from_bits(bits: u64) -> f64 {
+    unsafe { std::mem::transmute(bits) }
+}
```
raskad, line 50: I would be in favor of just switching to 1.83 MSRV anyways. We always have been using mostly the latest version.

raskad (commented):
I already have some comments, but I still have to look at the real inner value code. Only skimmed it until now.
All in all I really nice work!

I ran the benchmarks again and noticed that while there are some significant variants between the individual benchmarks, the overall score is now almost on par with the main branch. Specifically the `Crypto` and `NavierStokes` benchmarks got faster. @hansl Is this trend also also something that you can confirm when you run them? Otherwise I'll have to take a look why this might be the case on my setup.

One general question regarding the feature flag: Do we plan to remove it depending on some certain features, or do we want to keep that flag? If we want to keep it, I think we probably need to think about how we can test the code with / without to be sure that everything works with both features. Also we would have to choose a default feature for our benchmarks etc.
Based on just the additional effort with all the tooling, I would tend to remove it at some point. Any other thoughts?

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -0,0 +1,864 @@
...
+//! that can represent any JavaScript value. If the integer is a non-NaN value,
+//! it will be stored as a 64-bits float. If it is a `f64::NAN` value, it will
+//! be stored as a quiet `NaN` value. Subnormal numbers are regular float.
+//!
+//! For any other type of values, the value will be stored as a 51-bits non-zero
+//! integer.
+//!
+//! In short, the memory layout of a NaN-boxed value is as follows:
+//!
+//! | Type of           | Bit Layout | Comment |
```
jasonwilliams, line 11: The table layout doesn't seem right in this comment, there's 3 headings but 4 columns underneath, do they properly match up?

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -0,0 +1,864 @@
...
+//! |-------------------|------------|---------|
+//! | `+Infinity`       | `7FF0:0000:0000:0000`    | |
+//! | `-Infinity`       | `FFF0:0000:0000:0000`    | |
+//! | `NAN` (quiet)     | `7FF8:0000:0000:0000`    | |
+//! | `NAN` (signaling) | `FFF8:0000:0000:0000`    | |
```
jasonwilliams, line 16: Why do we need a distinction here? When are they used separately?

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -0,0 +1,864 @@
...
+//! | `Integer32`       | `7FF7:0000:IIII:IIII`    | 32-bits integer. |
+//! | `BigInt`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 0` | 51-bits pointer. Assumes non-null pointer. |
+//! | `Object`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 1` | 51-bits pointer. |
+//! | `Symbol`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 2` | 51-bits pointer. |
+//! | `String`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 3` | 51-bits pointer. |
+//! | `Float64`         | Any other values.        | |
+//!
+//! Pointers have the highest bit (in the `NaN` tag) set to 1, so they
+//! can represent any value from `0x8000_0000_0000` to `0xFFFF_FFFF_FFFF`.
+//! The last 2 bits of the pointer is used to store the type of the value.
```
jasonwilliams, line 30: is this right or am I reading it wrong? Looks like the value type (for pointers) is stored from `7FF8... - 7FFF...` at the top end (when looking at the values above from line 22 to 25)

core/engine/src/builtins/array/tests.rs
```diff
@@ -953,3 +953,16 @@ fn array_sort() {
...
     ]);
 }
+
+#[test]
+fn array_of_neg_zero() {
+    run_test_actions([
+        TestAction::assert_context(|_| {
+            eprintln!("test start...");
+            true
+        }),
```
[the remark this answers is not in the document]
hansl: It was. Removed.

core/engine/src/value/inner/nan_boxed.rs
hansl: We cannot, not until Tarpaulin is supporting it. See our own comment at https://github.com/boa-dev/boa/blob/main/.github/workflows/rust.yml#L35-L36

This PR is still opened: https://github.com/xd009642/tarpaulin/issues/1642

TBH I would be a bigger advocate to keep the MSRV constant. It's always a pain on dependents to move version. Let's make it a bigger deal when we update. Those methods aren't much to maintain. I can add a TODO to it (and create an issue) so we don't easily forget to remove them when MSRV moves past 1.83.

core/engine/src/value/inner/nan_boxed.rs
hansl: There's only 3 columns in the comment. You might be confused by the pointer hexa explanation that uses `\|` to escape the pipe sign.

Here's a rendered version:

CleanShot 2024-12-31 at 18 53 56@2x

core/engine/src/value/inner/nan_boxed.rs
hansl: They are two different actual NaN values specced in IEEE 754; signaling just means that the CPU will raise an exception if used in any instruction. We should not mess with it, and it's something we can't use internally as it might lead to exceptions.

core/engine/src/value/inner/nan_boxed.rs
hansl: That's correct. If we only had 1 pointer type it would be easier, but I used a combination of pointer tagging and NaN Boxing to cover all our cases.

hansl:
>One general question regarding the feature flag: Do we plan to remove it depending on some certain features, or do we want to keep that flag? If we want to keep it, I think we probably need to think about how we can test the code with / without to be sure that everything works with both features. Also we would have to choose a default feature for our benchmarks etc.
Based on just the additional effort with all the tooling, I would tend to remove it at some point. Any other thoughts?

My plan when adding the feature flag was to be quickly able to switch between the two implementations to run benchmarks and tests for this PR.

We might decide to merge the flag in the `main` branch, e.g. if we want to have a release and test client projects with NaN boxing before actually making it the default. I did not intend it to live through a full release, but I also didn't want to commit to NaN before we were 100% committed to it, and giving us an easy revert.

core/engine/src/value/mod.rs
hansl: You'll have to explain to me what code you removed. I removed that `let Some(rational) = self.0.as_float64()` branch and got a lot of tests to fail. Is that the branch you were meaning?

hansl:
Benchmarks (using `taskpolicy` on MacOS makes CPU priority to the efficiency cores):

## Enum-based
```
$ cargo build  --release --bin boa
...
$ taskpolicy -b ./target/release/boa -O ~/Sources/boa-dev/data/bench/bench-v8/combined.js
PROGRESS Richards
RESULT Richards 19.7
PROGRESS DeltaBlue
RESULT DeltaBlue 18.5
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 21.2
PROGRESS RayTrace
RESULT RayTrace 48.8
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 58.8
PROGRESS RegExp
RESULT RegExp 10.9
PROGRESS Splay
RESULT Splay 82.3
PROGRESS NavierStokes
RESULT NavierStokes 50.4
SCORE 31.6
```

## NaN-Boxed
```
$ cargo build  --release --bin boa --features nan-box-jsvalue
...
$ taskpolicy -b ./target/release/boa -O ~/Sources/boa-dev/data/bench/bench-v8/combined.js
PROGRESS Richards
RESULT Richards 12.8
PROGRESS DeltaBlue
RESULT DeltaBlue 12.2
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 22.0
PROGRESS RayTrace
RESULT RayTrace 37.5
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 45.2
PROGRESS RegExp
RESULT RegExp 9.97
PROGRESS Splay
RESULT Splay 56.7
PROGRESS NavierStokes
RESULT NavierStokes 44.5
SCORE 24.9
```

hansl:
On the latest commit, I changed the pointer tagging to using the top bits of the pointer. This means that we lose bits on the pointer value itself (after doing some research 48 bits should be enough), but we gain on performance by having non overlapping ranges for pointers. Seems like it's speeding it up a bit overall:

```
RESULT Richards 87.9
RESULT DeltaBlue 85.6
RESULT Crypto 142
RESULT RayTrace 250
RESULT EarleyBoyer 312
RESULT RegExp 47.4
RESULT Splay 437
RESULT NavierStokes 256
SCORE 161
```

The final score is getting real close to before this PR on my machine.

core/engine/src/value/inner/nan_boxed.rs
hansl: This comment made me realize there was a bit of performance to gain on variant creation by making the pointer type tag the highest bits and using non-overlapping ranges for pointers as well.

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -19,21 +19,14 @@
...
 //! | `True`            | `7FF6:0000:0000:0001`    | |
 //! | `Integer32`       | `7FF7:0000:IIII:IIII`    | 32-bits integer. |
-//! | `BigInt`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 0` | 51-bits pointer. Assumes non-null pointer. |
-//! | `Object`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 1` | 51-bits pointer. |
-//! | `Symbol`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 2` | 51-bits pointer. |
-//! | `String`          | `7FF[8-F]:PPPP:PPPP:PPPP \| 3` | 51-bits pointer. |
+//! | `BigInt`          | `7FF8:PPPP:PPPP:PPPP`    | 49-bits pointer. Assumes non-null pointer. |
+//! | `Object`          | `7FFA:PPPP:PPPP:PPPP`    | 49-bits pointer. |
+//! | `Symbol`          | `7FFC:PPPP:PPPP:PPPP`    | 49-bits pointer. |
+//! | `String`          | `7FFE:PPPP:PPPP:PPPP`    | 49-bits pointer. |
```
jasonwilliams, line 25: So in the pointer space we only have 1 unused slot left right which is `7FF8`

hansl: `7FF8` (or `b0111_1111_1111_1000`) is used by bigint if the pointer is non-null. The 16th MSB isn't used (yet?), so pointers are actually 48 bits (I'll correct the comment).

hansl: Let me know if the commit helps clarifying this.

jasonwilliams: Yes sorry I meant to ask if 7FFF is unused and reserved

raskad (approved):
Greak work. Also very nice docs in nan_boxed!

Ran the benchmarks locally rebased on the current register enabled main branch and the difference looked ok to me.

core/engine/src/value/inner/nan_boxed.rs
jedel1043, line 1: Can we make this implementation conditional for only `x64_86` architectures? Mainly because I think this assumes a lot of things that may as well be incorrect assumptions for `arm`, `riscv`, even `wasm`. The alternative would be to add a LOT of platform-specific tests to ensure this is still valid for the rest of architectures, but that's a lot more work than just opting in on x86-64 systems.

jedel1043 (changes requested):

core/engine/src/value/inner/nan_boxed.rs
hansl: I don't mind, but this implementation actually make two arch assumptions;

1. f64 are 64 bits IEEE-754, which AFAIK is guaranteed by Rust. 
2. 48-bits pointers maximum (technically could be 49). This stands true for all current 64 bits architectures, including aarch64 which is where I developed this PR (I have a Mac as my main machine). The current ISAs enable full 64-bits only manually by programs that need it, by default it is limited to 48 bits. This is clarified here: https://en.m.wikipedia.org/wiki/64-bit_computing (also see below).

> However, not all 64-bit instruction sets support full 64-bit virtual memory addresses; x86-64 and AArch64 for example, support only 48 bits of virtual address, with the remaining 16 bits of the virtual address required to be all zeros (000...) or all ones (111...), and several 64-bit instruction sets support fewer than 64 bits of physical memory address.

ALL 32 bits architectures are compatible, of course. 

WASM with MEMORY64 (which is very rare) follows the pointer structure of its host architecture. More info: https://spidermonkey.dev/blog/2025/01/15/is-memory64-actually-worth-using.html

This leaves RISC-V and processes that enable 5-level paging extensions on Intel (https://en.m.wikipedia.org/wiki/Intel_5-level_paging). I don't mind disabling for RISC-V as a precaution and going back to the previous implementation for those processors. That means two implementations to maintain though. I had that in one of the commits above, I can just cherry pick it back. 

There is no way to feature gate on 5-level paging as it's a software trigger. 

I do assert that pointers take a max of 48 bits at runtime so people would likely get a neat panic message if they go over. So there's that. 

Having said all that, SpiderMonkey doesn't have an alternative implementation when I looked for it. They only support NaN boxing. V8 uses pointer tagging which is a different implementation with its own limitations.

jedel1043: Hmm, if it's widely supported then I don't mind leaving it like this. We can always add an alternative implementation later if we find that we need to support an architecture incompatible with this.

Can you add your comment as documentation? It would be nice to have that info on the implementation for future reference.

hansl:
@jedel1043 PTAL.

jedel1043 (approved):
Really nice implementation! Looks good

PR #4092, opened 2024-12-23
Allow local parameters if mapped arguments object is not used

raskad:

- Allow local parmeters if mapped arguments object is not used
- Add tests for scopes in `boa_ast`


jedel1043 (approved):
Nice optimization!

nekevss (approved):

hansl (approved):

PR #4098, opened 2024-12-26
Bump the `temporal_rs` version and related changes

nekevss:
This Pull Request is related to ongoing work for #1804 

This is a general bump in the `temporal_rs` version that implements `ZonedDateTime` for duration rounding, updates the general API, build out of the remaining `Now` methods, and a couple other fixes.

Some other changes made in this PR are:

  - Updates to the Temporal object's `static_property` attribute flags
  - Fixed misspelled property names `epochMilliseconds` and `epochNanoseconds`
  - Removed some unused functions and variables in the `PlainDate` module

### DIFF
--- a/core/engine/src/builtins/temporal/now.rs
+++ b/core/engine/src/builtins/temporal/now.rs
@@ -14,8 +14,9 @@ use boa_profiler::Profiler;
 use temporal_rs::{Now as NowInner, TimeZone};
 
 use super::{
-    create_temporal_datetime, create_temporal_instant, ns_max_instant, ns_min_instant,
-    time_zone::default_time_zone, to_temporal_timezone_identifier,
+    create_temporal_date, create_temporal_datetime, create_temporal_instant, create_temporal_time,
+    create_temporal_zoneddatetime, ns_max_instant, ns_min_instant, time_zone::default_time_zone,
+    to_temporal_timezone_identifier,
 };
 
 /// JavaScript `Temporal.Now` object.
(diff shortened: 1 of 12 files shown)


jedel1043:
Nice! Any reasons for the regressions? or are they just false positives? Also, do we want to bump test262 to pull the latest Temporal tests?

nekevss:
At a glance of the tests names, most if not all are going to be false positives from removing the not implemented errors, so I'd prefer to move along with these as is. Although, I'll probably start tracking the tests broken and PR # for future use as I'm starting to switch a bit more into bug hunting to shore up the tests.

I can bump the tests.

jedel1043:
Oof, those are a lot more failures than expected. I suppose those are the new spidermonkey tests that were imported. Can you check why they're failing? If it's something engine-dependent, it would be good to gate the tests.

jedel1043:
By the way, if it's getting too unwieldy for a single PR, I think it would be good to split the test262 bump into its own PR.

nekevss:
Hmmmm, I think I'd prefer to do it in a separate PR just for the separation of concerns. I was already leaning towards splitting them off last night. I'll just move them over to another branch and push that up.

core/engine/src/builtins/temporal/now.rs
```diff
@@ -83,25 +85,40 @@ impl Now {
...
         )
         .map(Into::into)
     }
 
     /// `Temporal.Now.zonedDateTime`
-    fn zoned_date_time(_: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
-        Err(JsNativeError::error()
-            .with_message("not yet implemented.")
-            .into())
+    fn zoneddatetime(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
```
jedel1043, line 96: Thought: not gonna bikeshed because I think it's not worth for internal APIs, but we should probably do a pass through our names to agree on at least a "convention" .

jedel1043 (approved):
Nice work!

raskad (approved):

core/engine/src/builtins/temporal/now.rs
nekevss: Yeah, I've been wanting to open an issue to go over naming conventions in `temporal_rs` for a few weeks now. Would be handy for Boa too

PR #4101, opened 2024-12-29
Set the array "length" property in `[[DefineOwnProperty]]` based on the array shape

raskad:
This optimization is especially useful for array literal construction. Instead of calling `OrdinaryGetOwnProperty` and `OrdinaryDefineOwnProperty` on the array object while setting the "length" property, all of that can be skipped if the shape of the array has not changed.


jedel1043 (approved):
Nice optimization!

HalidOdat (commented):
Looks perfect to me :)

HalidOdat (approved):

PR #4108, opened 2025-01-05
Cleanup CLI to use eyre + refactor patterns

### LINKED ISSUE #4085: `SimpleModuleLoader` cannot resolve relative paths on the CLI
**Describe the bug**

Trying to run this code using Boa: https://github.com/KieranP/Game-Of-Life-Implementations/tree/master/typescript

`import { World } from './world.js';` producing this error: `Uncaught SyntaxError: expected token '.', got '{' in import.meta at line 1, col 8`

`export class World {` producing this error: `Uncaught SyntaxError: unexpected token 'export', primary expression at line 3, col 1`

**To Reproduce**

Compile the Typescript to Javascript using `tsc` and then try to run `boa play.js`

**Expected behavior**

Should parse and run as expected. The code doesn't contain anything complex

- OS: Mac OS
- Version: Sequioa 15.1.1
- Target triple: arm64
- Rustc version: 1.83.0

jedel1043:
This should both make it easier to develop the CLI and also make it easier to add custom error messages in the future.

Closes #4085.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -18,15 +18,26 @@ use boa_engine::{
     optimizer::OptimizerOptions,
     script::Script,
     vm::flowgraph::{Direction, Graph},
-    Context, JsError, JsNativeError, JsResult, Source,
+    Context, JsError, Source,
 };
+use boa_parser::source::ReadChar;
 use clap::{Parser, ValueEnum, ValueHint};
+use color_eyre::{
+    eyre::{eyre, WrapErr},
+    Result, Section,
+};
 use colored::Colorize;
 use debug::init_boa_debug_object;
 use rustyline::{config::Config, error::ReadlineError, EditMode, Editor};
 use std::{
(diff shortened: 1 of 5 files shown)


cli/src/main.rs
```diff
@@ -248,96 +258,95 @@ fn generate_flowgraph(
...
+
+fn main() -> Result<()> {
+    color_eyre::config::HookBuilder::default()
+        .display_location_section(false)
+        .display_env_section(false)
+        .install()?;
+
+    if std::env::var("RUST_BACKTRACE").is_err() {
+        std::env::set_var("RUST_BACKTRACE", "0");
+    }
```
raskad, line 336: Why do we need to do this?

raskad (commented):
Looks good overall. I just have one comment.

cli/src/main.rs
jedel1043: Forgot to remove it. It was an attempt to remove the error location which we don't need for a public CLI, but it happened to be configured by using the `HookBuilder`.

raskad (approved):

nekevss (approved):

PR #4117, opened 2025-01-10
Fix CI with the latest linter errors introduced in 1.84

hansl:
To maintain backward compatibility, we had to move to using caller style calls (`sptr::Strict::addr(...)`) as clippy was erroneously triggering those as unused (and `.addr()` is not stabilized in 1.82).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,7 @@ syn = { version = "2.0.95", default-features = false }
 proc-macro2 = "1.0"
 synstructure = "0.13"
 measureme = "12.0.0"
+# TODO: when moving past 1.84, remove dependency to sptr and use `.addr()` directly.
 sptr = "0.3.2"
 paste = "1.0"
 rand = "0.8.5"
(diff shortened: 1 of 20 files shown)


raskad:
I'm for moving to 1.84 and removing the sptr dependency. Reference https://github.com/boa-dev/boa/compare/main...no-sptr

hansl:
@raskad 

> I'm for moving to 1.84 and removing the sptr dependency. Reference [main...no-sptr](https://github.com/boa-dev/boa/compare/main...no-sptr)

I can't remember where I put my other comment, but we cannot right now as Tarpaulin is stuck on 1.82. There's a comment about it: https://github.com/boa-dev/boa/blob/a23597631172a6b7ca95fe9c80384a889f380e04/.github/workflows/rust.yml#L35-L37

raskad:
Ah, true I remember now.

raskad (approved):

jedel1043 (approved):

PR #4121, opened 2025-01-12
Allow referencing `super` within initializer of static private property

### LINKED ISSUE #3055: Panic at assigning a super property as a static private class field value
**Describe the bug**
A super property assigned to a static private class field results in crash. 

**To Reproduce**
This Javascript crashes with a `thread 'main' panicked at 'super access must be in a function environment', boa_engine/src/vm/opcode/environment/mod.rs:43:18`

```js
class a{
 static #x = super.b;
}
```

**Expected behavior**
Not panic, perhaps result in an `undefined`. Similar code (such as below) where a super property is assigned to a static _not_-private class field prints out an `undefined`. However, I'm not sure if that is also a good output/behavior, should maybe throw an 'invalid super usage'

```js
class a{
 static x = super.b;
}
```

**Build environment:**
• OS: Ubuntu
• Version: 20.04
• Target triple: x86_64-unknown-linux-gnu
• Rustc version: rustc 1.70.0

jedel1043:
Fixes #3055.

### DIFF
--- a/core/engine/src/tests/class.rs
+++ b/core/engine/src/tests/class.rs
@@ -69,5 +69,6 @@ fn class_can_access_super_from_static_initializer() {
         "#}),
         TestAction::assert_eq("a.field", js_str!("super field")),
         TestAction::assert_eq("b.field", js_str!("super field")),
+        TestAction::assert_eq("c.field", js_str!("super field")),
     ]);
 }
--- a/core/engine/src/tests/class.rs
+++ b/core/engine/src/tests/class.rs
@@ -51,17 +51,23 @@ fn class_superclass_from_regex_error() {
 fn class_can_access_super_from_static_initializer() {
     run_test_actions([
         TestAction::run(indoc! {r#"
-            class b {
+            class a {
                 static field = "super field";
             }
 
-            class a extends b {
+            class b extends a {
                 static #field = super.field;
                 static get field() {
                     return this.#field;
                 }
             }
+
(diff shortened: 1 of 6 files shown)


core/engine/src/tests/class.rs
```diff
@@ -45,3 +45,23 @@ fn class_superclass_from_regex_error() {
...
+#[test]
+fn class_can_access_super_from_static_initializer() {
+    run_test_actions([
+        TestAction::run(indoc! {r#"
+            class b {
+                static field = "super field";
+            }
+
+            class a extends b {
+                static #field = super.field;
```
hansl, line 59: I'd like to see a test case for public field with super initializer. Can be part of the same test.

core/engine/src/tests/class.rs
```diff
@@ -51,17 +51,23 @@ fn class_superclass_from_regex_error() {
...
                 }
             }
+
+            class c extends a {
+                static field = super.field;
+            }
+
         "#}),
         TestAction::assert_eq("a.field", js_str!("super field")),
+        TestAction::assert_eq("b.field", js_str!("super field")),
```
hansl, line 71:
```suggestion
        TestAction::assert_eq("b.field", js_str!("super field")),
        TestAction::assert_eq("c.field", js_str!("super field")),
```

nekevss (approved):
Looks good to me!

Issue #4075, opened 2024-12-09
Tracking issue for Temporal ZonedDateTime object


nekevss:
This issue is a tracking issue for the current state Temporal.ZonedDateTime implementation in Boa and the current work towards https://github.com/boa-dev/boa/issues/1804.

| Builtin property / method | Boa Impl | temporal_rs Impl | 100% Conformance |
|---------------------------|----------|------------------|------------------|
| 6.2.1 Temporal.ZonedDateTime.prototype | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.2.2 Temporal.ZonedDateTime.from ( item [ , options ] ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.2.3 Temporal.ZonedDateTime.compare ( one, two ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.3.1 Temporal.ZonedDateTime.prototype.constructor | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.3.2 Temporal.ZonedDateTime.prototype[ %Symbol.toStringTag% ] | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.3.3 get Temporal.ZonedDateTime.prototype.calendarId | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.3.4 get Temporal.ZonedDateTime.prototype.timeZoneId | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.3.5 get Temporal.ZonedDateTime.prototype.era | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 6.3.6 get Temporal.ZonedDateTime.prototype.eraYear | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
[...]

Issue #4074, opened 2024-12-09
Tracking issue for Temporal Instant object


nekevss:
This issue is a tracking issue for the current state Temporal.Instant implementation in Boa and the current work towards https://github.com/boa-dev/boa/issues/1804.

| Builtin property / method | Boa Impl | temporal_rs Impl | 100% Conformance |
|---------------------------|----------|------------------|------------------|
| 8.2.2 Temporal.Instant.from ( item ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.2.3 Temporal.Instant.fromEpochMilliseconds ( epochMilliseconds ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.2.4 Temporal.Instant.fromEpochNanoseconds ( epochNanoseconds ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.2.5 Temporal.Instant.compare ( one, two ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.3.1 Temporal.Instant.prototype.constructor | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.3.2 Temporal.Instant.prototype[ %Symbol.toStringTag% ] | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.3.3 get Temporal.Instant.prototype.epochMilliseconds | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.3.4 get Temporal.Instant.prototype.epochNanoseconds | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 8.3.5 Temporal.Instant.prototype.add ( temporalDurationLike ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
[...]

Issue #4073, opened 2024-12-09
Tracking issue for Temporal PlainTime object


nekevss:
This issue is a tracking issue for the current state Temporal.PlainTime implementation in Boa and the current work towards https://github.com/boa-dev/boa/issues/1804.

| Builtin property / method | Boa Impl | temporal_rs Impl | 100% Conformance |
|---------------------------|----------|------------------|------------------|
| 4.2.1 Temporal.PlainTime.prototype | :white_check_mark: | N/A | :white_check_mark: |
| 4.2.2 Temporal.PlainTime.from ( item [ , options ] ) | :white_check_mark: | :white_check_mark:  | :x: |
| 4.2.3 Temporal.PlainTime.compare ( one, two ) | :white_check_mark: | :white_check_mark: | :x: |
| 4.3.1 Temporal.PlainTime.prototype.constructor | :white_check_mark: | N/A | :white_check_mark: |
| 4.3.2 Temporal.PlainTime.prototype[ %Symbol.toStringTag% ] | :white_check_mark: | N/A | :white_check_mark: |
| 4.3.3 get Temporal.PlainTime.prototype.hour | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| 4.3.4 get Temporal.PlainTime.prototype.minute | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| 4.3.5 get Temporal.PlainTime.prototype.second | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| 4.3.6 get Temporal.PlainTime.prototype.millisecond | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| 4.3.7 get Temporal.PlainTime.prototype.microsecond | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| 4.3.8 get Temporal.PlainTime.prototype.nanosecond | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| 4.3.9 Temporal.PlainTime.prototype.add ( temporalDurationLike ) | :white_check_mark: | :white_check_mark: | :x: |
| 4.3.10 Temporal.PlainTime.prototype.subtract ( temporalDurationLike ) | :white_check_mark: | :white_check_mark: | :x: |
| 4.3.11 Temporal.PlainTime.prototype.with ( temporalTimeLike [ , options ] ) | :white_check_mark: | :white_check_mark: | :x: |
[...]

Issue #4072, opened 2024-12-09
Tracking issue for Temporal DateTime object


nekevss:
This issue is a tracking issue for the current state Temporal.DateTime implementation in Boa and the current work towards https://github.com/boa-dev/boa/issues/1804.

| Builtin property / method | Boa Impl | temporal_rs Impl | 100% Conformance |
|---------------------------|----------|------------------|------------------|
| 5.2.1 Temporal.PlainDateTime.prototype | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.2.2 Temporal.PlainDateTime.from ( item [ , options ] ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.2.3 Temporal.PlainDateTime.compare ( one, two ) | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.3.1 Temporal.PlainDateTime.prototype.constructor | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.3.2 Temporal.PlainDateTime.prototype[ %Symbol.toStringTag% ] | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.3.3 get Temporal.PlainDateTime.prototype.calendarId | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.3.4 get Temporal.PlainDateTime.prototype.era | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.3.5 get Temporal.PlainDateTime.prototype.eraYear | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.3.6 get Temporal.PlainDateTime.prototype.year | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
| 5.3.7 get Temporal.PlainDateTime.prototype.month | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> | <ul><li> [ ] </li></ul> |
[...]

PR #4126, opened 2025-01-13
Implement more `toString` and `toJSON` methods on Temporal builtins

nekevss:
This Pull Request is related to ongoing work for #4072, #4073, #4074, and #4075

It changes the following:

- Bumps temporal_rs version with some various bug fixes.
- Implements `toString` and `toJSON` for `PlainDateTime`
- Implements `toString` and `toJSON` for `PlainTime` 
- Implements `toString` and `toJSON` for `Instant`
- Implements `toString` and `toJSON` for `ZonedDateTime`
- Adds `options::get_digit_options` utility method

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3541,7 +3541,7 @@ checksum = "42a4d50cdb458045afc8131fd91b64904da29548bcb63c7236e0844936c13078"
 [[package]]
 name = "temporal_rs"
 version = "0.0.4"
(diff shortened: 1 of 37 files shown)


hansl:
Any good reasons new tests broke?

nekevss:
At least 3 of them should be false positives as those methods were not previously supported. The add test is potentially a bug vs. a false positive. But that will need to be looked into and followed up on `temporal_rs`.

Either way, I'll be putting them on my broken test list :smile:

hansl (approved):

raskad (approved):

PR #4130, opened 2025-01-13
`setTimeout`, `setInterval` and `clearInterval` (and the same `clearTimeout`) implementations

hansl:
This adds non-async implementations of the `setTimeout`/`setInterval` API functions. They are by default registered in the context when registering all APIs from `boa_runtime`.

This adds a new type of jobs that implement `HostEnqueueTimeoutJob` (see https://tc39.es/ecma262/#sec-hostenqueuetimeoutjob).

This requires https://github.com/boa-dev/boa/pull/4141 to be merged first.

### DIFF
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -117,12 +117,12 @@ impl NativeJob {
 
 /// An ECMAScript [Job] that runs after a certain amount of time.
 ///
-/// This represents the [`HostEnqueueTimeoutJob`] operation from the specification.
+/// This represents the [HostEnqueueTimeoutJob] operation from the specification.
 ///
 /// [HostEnqueueTimeoutJob]: https://tc39.es/ecma262/#sec-hostenqueuetimeoutjob
 pub struct TimeoutJob {
     /// The instant this job should be run, in msec since epoch. This will be compared
-    /// to the host's [`HostHooks::utc_now`] method.
+    /// to the host's [`crate::context::HostHooks::utc_now`] method.
     timeout: i64,
     /// The job to run after the time has passed.
     job: NativeJob,
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -126,7 +126,7 @@ impl NativeJob {
 pub struct TimeoutJob {
     /// The distance in milliseconds in the future when the job should run.
     /// This will be added to the current time when the job is enqueued.
-    timeout: u64,
+    timeout: JsDuration,
     /// The job to run after the time has passed.
     job: NativeJob,
 }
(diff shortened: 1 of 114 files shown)


core/engine/src/job.rs
```diff
@@ -115,6 +116,87 @@ impl NativeJob {
...
+        F: FnOnce(&mut Context) -> JsResult<JsValue> + 'static,
+    {
+        Self::new_unchecked(
+            NativeJob::new(f),
+            timeout
+                .duration_since(std::time::UNIX_EPOCH)
+                .expect("Invalid SystemTime")
+                .as_millis() as i64,
+        )
+    }
```
jedel1043, line 167: I think we shouldn't depend so heavily on `SystemTime`. Seems weird to tie one of our job types to something that could panic on certain platforms, and users won't be able to sandbox the engine from the OS if this is in place.

Having `new(job: NativeJob, timeout: i64)` sounds enough for all usecases, and the specific executor can use the provided timeout to schedule the job as it fits.

core/engine/src/job.rs
```diff
@@ -115,6 +116,87 @@ impl NativeJob {
...
 }
 
+/// An ECMAScript [Job] that runs after a certain amount of time.
+///
+/// This represents the [`HostEnqueueTimeoutJob`] operation from the specification.
+///
+/// [HostEnqueueTimeoutJob]: https://tc39.es/ecma262/#sec-hostenqueuetimeoutjob
+pub struct TimeoutJob {
+    /// The instant this job should be run, in msec since epoch. This will be compared
+    /// to the host's [`HostHooks::utc_now`] method.
```
jedel1043, line 126: I don't think we should mention details about `SimpleJobExecutor`, since every executor can handle the timeout amount in a different way.

jedel1043 (changes requested):
Really nice implementation! I have some opinions on using `SystemTime`, but aside from that everything else looks great!

core/engine/src/job.rs
hansl: They should definitely not. People building libraries should not be tied to what an executor interpret the value as.

core/engine/src/job.rs
hansl: The point of execution needs to be an `Instant`-like point (monotonic and absolute). It cannot be left to interpretation.

jedel1043: I would be in favour of using `Duration` instead. The spec is very clear about the milliseconds` argument on [`HostEnqueueTimeoutJob`](https://tc39.es/ecma262/#sec-hostenqueuetimeoutjob):

> It schedules timeoutJob in the realm `realm` in the agent signified by `realm.[[AgentSignifier]]` to be performed after at least `milliseconds` milliseconds.

Hence, it should be fine to use `Duration` in this case.

For `utc_now`, I think we'd have to rethink our approach a bit, maybe doing a survey of what libraries do to support monotonically increasing instants.

hansl: Done. `TimeoutJob` now only holds the number of millis as a timeout, and it's the JobExecutor's job to add it to the current time and keep that internally.

nekevss (approved):
Looking great overall! Nice job!

jedel1043 (approved):
Looks really nice!

PR #4143, opened 2025-01-24
Simplify date parser

CrazyboyQCD:
- Avoid allocation if string is ascii
- Return early if string from utf16 encoded `JsString` is not ascii
- Replace `Char<'_>` with `slice::Iter<'_, u8>`
- Add `next_n_digits` helper method to save calls of `next_digit`
- Merge duplicate codes in matches

### DIFF
--- a/core/engine/src/builtins/date/utils.rs
+++ b/core/engine/src/builtins/date/utils.rs
@@ -1,6 +1,9 @@
 use crate::{context::HostHooks, js_string, value::IntegerOrInfinity, JsStr, JsString};
 use boa_macros::js_str;
-use std::{iter::Peekable, str::Chars};
+use boa_string::JsStrVariant;
+use std::iter::Peekable;
+use std::slice::Iter;
+use std::str;
 use time::{macros::format_description, OffsetDateTime, PrimitiveDateTime};
 
 // Time-related Constants
@@ -750,8 +753,16 @@ pub(super) fn pad_six(t: u32, output: &mut [u8; 6]) -> JsStr<'_> {
 /// [spec-format]: https://tc39.es/ecma262/#sec-date-time-string-format
 pub(super) fn parse_date(date: &JsString, hooks: &dyn HostHooks) -> Option<i64> {
     // All characters must be ASCII so we can return early if we find a non-ASCII character.
-    let Ok(date) = date.to_std_string() else {
-        return None;
+    let owned_js_str = date.as_str();
+    let owned_string: String;
+    let date = match owned_js_str.variant() {
+        JsStrVariant::Latin1(s) =>
+        // SAFETY: Since all characters are ASCII we can safely convert this into str.
+        unsafe { str::from_utf8_unchecked(s) },
+        JsStrVariant::Utf16(s) => {
+            owned_string = String::from_utf16(s).ok()?;
+            owned_string.as_str()
+        }
     };
 
     // Date Time String Format: 'YYYY-MM-DDTHH:mm:ss.sssZ'
--- a/core/engine/src/builtins/date/utils.rs
+++ b/core/engine/src/builtins/date/utils.rs
@@ -807,6 +807,27 @@ struct DateParser<'a> {
     offset: i64,
 }
 
+// Copied from https://github.com/RoDmitry/atoi_simd/blob/master/src/fallback.rs,
+// which is based on https://rust-malaysia.github.io/code/2020/07/11/faster-integer-parsing.html.
+#[doc(hidden)]
+#[allow(clippy::inline_always)]
+pub(in crate::builtins::date) mod fast_atoi {
+    #[inline(always)]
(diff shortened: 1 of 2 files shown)


core/engine/src/builtins/date/utils.rs
```diff
@@ -805,22 +819,39 @@ impl<'a> DateParser<'a> {
...
             if c.is_ascii_digit() {
-                Some((u32::from(c) - u32::from('0')) as u8)
+                Some(c - b'0')
             } else {
                 None
             }
         })
     }
 
+    fn next_n_digits<const N: usize>(&mut self) -> Option<[u8; N]> {
```
nekevss, line 839: suggestion: Use `next_digit` over using `next_n_digits` where possible.

As tempting as `next_n_digits` is to write, most the lengths should be relatively stable here. We can then also avoid using unsafe code.

nekevss (changes requested):
Looks good to me at a high level. I like the move from `char` to `u8` ascii. I have a suggestion.

core/engine/src/builtins/date/utils.rs
CrazyboyQCD: I put `next_digit` in `next_n_digits` now because I found `Peekable<slice::Iter>` doesn't give the len of `slice::Iter` under serveral conditions, so the length check is useless.
But I prefer using `next_n_digits` as it is not that noisy here.

core/engine/src/builtins/date/utils.rs
```diff
@@ -805,20 +819,52 @@ impl<'a> DateParser<'a> {
...
 
-    fn next_digit(&mut self) -> Option<u8> {
-        self.input.next().and_then(|c| {
-            if c.is_ascii_digit() {
-                Some((u32::from(c) - u32::from('0')) as u8)
-            } else {
-                None
+    #[allow(clippy::items_after_statements)]
+    #[allow(clippy::inline_always)]
+    fn parse_n_ascii_digits<const N: usize>(&mut self) -> Option<u64> {
```
nekevss, line 844: suggestion: add some tests for this function.

The optimizations here are kinda gnarly lol It would be nice to have some localized Rust tests just to be able to easily verify the behavior and correctness whenever someone comes back to it.

core/engine/src/builtins/date/utils.rs
```diff
@@ -750,22 +753,33 @@ pub(super) fn pad_six(t: u32, output: &mut [u8; 6]) -> JsStr<'_> {
...
 pub(super) fn parse_date(date: &JsString, hooks: &dyn HostHooks) -> Option<i64> {
     // All characters must be ASCII so we can return early if we find a non-ASCII character.
-    let Ok(date) = date.to_std_string() else {
-        return None;
+    let owned_js_str = date.as_str();
+    let owned_string: String;
+    let date = match owned_js_str.variant() {
+        JsStrVariant::Latin1(s) =>
+        // SAFETY: Since all characters are ASCII we can safely convert this into str.
+        unsafe { str::from_utf8_unchecked(s) },
```
jedel1043, line 761: This comment is incorrect though. `JsStrVariant::Latin1` could have latin1 encoded characters, which are incompatible with utf8.

core/engine/src/builtins/date/utils.rs
```diff
@@ -750,22 +753,37 @@ pub(super) fn pad_six(t: u32, output: &mut [u8; 6]) -> JsStr<'_> {
...
+            // SAFETY: Since all characters are ASCII we can safely convert this into str.
+            unsafe { str::from_utf8_unchecked(s) }
+        }
+        JsStrVariant::Utf16(s) => {
+            owned_string = String::from_utf16(s).ok()?;
+            if !owned_string.is_ascii() {
+                return None;
+            }
+            owned_string.as_str()
+        }
```
jedel1043, line 772: Funnily enough, I think this could probably use `Cow` to avoid creating an `owned_string` variable that is only used once.
```suggestion
    let date = match owned_js_str.variant() {
        JsStrVariant::Latin1(s) => {
            if !s.is_ascii() {
                return None;
            }
            // SAFETY: Since all characters are ASCII we can safely convert this into str.
            Cow::Borrowed(unsafe { str::from_utf8_unchecked(s) })
        }
        JsStrVariant::Utf16(s) => {
            date = String::from_utf16(s).ok()?;
            if !owned_string.is_ascii() {
                return None;
            }
            Cow::Owned(date)
        }
```

jedel1043 (approved):
LGTM!

nekevss (approved):
Overall looks good to me. Thanks for your work on this, especially to minimize the unsafe code.

I do have some general non-blocking concerns of the `unsafe` usage. It should be safe in this context and also should be fairly easy to test, especially given the invariants of `JsStrVariant`. But if for some reason, we ever need to undergo an unsafe audit in the future, I'm worried the unsafe usage here may come up.

PR #4154, opened 2025-02-04
Fix issues with `to_temporal_time` and `ZonedDateTime.prototype.withPlainTime`

nekevss:
This Pull Request is related to the ongoing work for temporal

It fixes a couple bugs

- Order of ops for accessing a time-like property bag 
- Uses the correct `from_partial` method from `temporal_rs`
- Properly handles an `undefined` arg provided to `withPlainTime`


raskad (approved):
Nice find

PR #4155, opened 2025-02-05
Add trace function to Logger trait

jamesthurley:
This Pull Request implements the discussion in Matrix about trace logging.

It changes the following:

- Adds a `trace` function to the `Logger` trait.
- Calls that `trace` function  from `Console.trace`.

For such a small PR, I encountered a few niggles:

## 1. Conflicting `trace` method
The `Logger` trait derives from the `Trace` trait, which already has a `trace` function, but is unrelated to trace logging. This meant I had to call the new trace function by doing:
```rust
Logger::trace(logger, formatter(args, context)?, &console.state, context)?;
```
rather than the following which is ambiguous to the compiler:
```rust
logger.trace(formatter(args, context)?, &console.state, context)?;
```

This is unsatisfactory, but I also felt that renaming the new `Logger.trace` function would make it inconsistent with the surrounding code.

## 2. Passing the stack trace
In matrix we discussed passing the stack trace into the `Logger.trace` function.
- I could pass in the `Vec<string>` of stack frame code block names, but this feels quite prescriptive with the Console deciding in how the stack trace is formatted.
- I could pass in a `Vec<&CallFrame>`, but the borrow checker points out that this means I'd be passing in both a mutable reference to `Context` and immutable references to the `CallFrame` instances inside `Context`.
- Therefore it seemed to make sense to just allow the `Logger.trace` implementation to extract the call stack from `Context` itself, and format the call frames as it sees fit, and it also keeps the `trace` signature consistent with the other logging functions.

## 3. Code block names
There were a couple of unexpected things here. If I comment out the `.filter(|name| !name.is_empty())` line which I added and run on the following test javascript:
```js
function run() {
  console.trace("Trace from 'run' function.");
  a();
}

function a() {
  b();
}

function b() {
  c();
}

function c() {
  console.trace("Trace from 'c' function.");
}

run();
```

I get the following output:
```
Trace from 'run' function.
<main>

Trace from 'c' function.
b
a
run
<main>

```
Unexpected things:
1. The blank lines under <main> are caused by a code block with an empty name, which isn't very useful to log.
2. The function that is calling `console.trace` isn't included in the stack trace.

I resolved the first by filtering out empty stack frame names, although I didn't love this solution.

I wasn't sure how to resolve the second point.

### DIFF
--- a/core/engine/src/context/mod.rs
+++ b/core/engine/src/context/mod.rs
@@ -521,9 +521,21 @@ impl Context {
     }
 
     /// Retrieves the current stack trace of the context.
+    ///
+    /// When constructing the stack trace the current frame is taken from [`Vm::frame`],
+    /// and the rest of the frames are taken from [`Vm::frames`].
+    /// The first frame is always a dummy frame (see [`Vm::frame`]).
+    ///
+    /// The stack trace is returned ordered with the most recent frames first.
     #[inline]
     pub fn stack_trace(&self) -> impl Iterator<Item = &CallFrame> {
-        self.vm.frames.iter().rev()
+        let frames: Vec<_> = self
+            .vm
+            .frames
+            .iter()
+            .chain(std::iter::once(&self.vm.frame))
+            .collect();
+        frames.into_iter().skip(1).rev()
     }
 
     /// Replaces the currently active realm with `realm`, and returns the old realm.
--- a/core/engine/src/context/mod.rs
+++ b/core/engine/src/context/mod.rs
@@ -522,19 +522,19 @@ impl Context {
 
     /// Retrieves the current stack trace of the context.
     ///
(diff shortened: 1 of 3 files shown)


jamesthurley:
I've added a test which checks the tracing output.

I think ideally the assert would be:
```
    assert_eq!(
        logs,
        indoc! { r#"
            one
            <main>
            two
            b
            a
            <main>
        "# }
```
But that would currently fail due to point 3 above, so I've modified the test to make it pass as I don't know if point 3 is easily addressable.

jedel1043:
> 1. The blank lines under are caused by a code block with an empty name, which isn't very useful to log.
> 2. The function that is calling console.trace isn't included in the stack trace.

TBH, those sound like bugs on `stacktrace` more than anything, so I wouldn't be against removing the filtering and fixing the bug after this PR.

nekevss (approved):
LGTM

PR #4156, opened 2025-02-07
Implement toLocaleString and some general cleanup

nekevss:
This Pull Request is related to the ongoing work for #1804 .

It changes the following:

- Implements the baseline `toLocaleString` stub for all types (the non ECMA-402 implementation)
- Add the intl methods `era` and `eraYear` to `PlainDate` and `PlainDateTime`
- Adds some general cleanup to `PlainYearMonth` to better align it with other Temporal builtins
- Removes some dead code from `now.rs` and `mod.rs`

### DIFF
--- a/core/engine/src/builtins/temporal/plain_year_month/mod.rs
+++ b/core/engine/src/builtins/temporal/plain_year_month/mod.rs
@@ -24,7 +24,9 @@ use temporal_rs::{
     Calendar, Duration, PlainYearMonth as InnerYearMonth,
 };
 
-use super::{calendar::to_temporal_calendar_slot_value, to_temporal_duration, DateTimeValues};
+use super::{
+    calendar::get_temporal_calendar_slot_value_with_default, to_temporal_duration, DateTimeValues,
+};
 
 /// The `Temporal.PlainYearMonth` object.
 #[derive(Debug, Clone, Trace, Finalize, JsData)]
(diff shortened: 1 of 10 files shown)


nekevss:
The regressions here appear to be from adding `era` and `eraYear` methods. I'm looking into them further. In the meantime, I'd be fine with merging this as is and fix the broken tests in subsequent iterations since supporting `era` and `eraYear` are pretty important from an ECMA-402 perspective. And the other changes clean up the general temporal code.

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -410,25 +362,35 @@ impl PlainYearMonth {
...
             .with_message("not yet implemented.")
             .into())
     }
 
+    /// 9.3.17 `Temporal.PlainYearMonth.prototype.since ( other [ , options ] )`
     fn since(_this: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
-        Err(JsNativeError::typ()
+        Err(JsNativeError::error()
             .with_message("not yet implemented.")
             .into())
```
jedel1043, line 376: Good to know we're returning `Error` instead of `TypeError` for unimplemented features to avoid false positives.

jedel1043 (approved):
LGTM!

jasonwilliams (approved):

Issue #4159, opened 2025-02-12
Add boa-macos-arm64  release file


ahaoboy:
**Describe the bug**

Macos only has boa-macos-amd64, it would be nice  if boa-macos-arm64 could be added

PR #4160, opened 2025-02-12
add macos arm64

ahaoboy:
Add boa-macos-arm64 release file

This Pull Request fixes/closes  https://github.com/boa-dev/boa/issues/4159

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,7 +91,7 @@ jobs:
           asset_name: boa-linux-amd64
           binary_name: boa
         - build: macos-amd64
-          os: macos-latest
+          os: macos-13
           rust: stable
           target: x86_64-apple-darwin
           asset_name: boa-macos-amd64
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,7 +82,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        build: [linux, macos, win-msvc]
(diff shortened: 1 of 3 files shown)


jedel1043:
If we're publishing arm64 builds, we should definitely add tests for that target on the CI.

ahaoboy:
> If we're publishing arm64 builds, we should definitely add tests for that target on the CI.

I added some CI tests, but I don't know if there are any missing

.github/workflows/release.yml
```diff
@@ -82,20 +82,26 @@ jobs:
...
         include:
         - build: linux
           os: ubuntu-20.04
           rust: stable
           target: x86_64-unknown-linux-gnu
           asset_name: boa-linux-amd64
           binary_name: boa
-        - build: macos
+        - build: macos-amd64
           os: macos-latest
```
jedel1043, line 94: Pin to `macos-13`? Seems weird to change it on tests but not here.

ahaoboy: Sorry, I missed it.

jedel1043 (approved):
Looks great!

ahaoboy:
> Looks great!

I wonder why boa doesn't use cargo-binstall format for publishing resource names?

ahaoboy:
Using the { name }-{ target } style makes it easier to install instead of compiling from source code.
```
cargo binstall boa
```

nekevss (approved):

PR #4162, opened 2025-02-15
Bump temporal_rs to Feb. 15 version + adjustments

nekevss:
This Pull Request is related to the ongoing work for #1804

This most recent version includes some bug fixes along with some API changes to `Now` in `temporal_rs`. Changes were made to accommodate the `Now` changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -371,6 +371,7 @@ dependencies = [
  "float-cmp",
  "futures-lite 2.6.0",
  "hashbrown 0.15.2",
+ "iana-time-zone",
  "icu_calendar 1.5.2",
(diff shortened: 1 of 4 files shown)


jedel1043:
Nice! Is the broken test expected?

nekevss:
It's not expected breakage, but I do think that it may have been a false positive on the test case.

The bumps included in this PR are the `Now` changes and the Calendar parsing fix (where all the fixed tests are from). Only the latter is called on the code for the broken test.

jedel1043 (approved):
If that's the case, this looks good for me.

jasonwilliams (approved):

PR #4172, opened 2025-02-22
Build out remaining method stubs for temporal

nekevss:
This Pull Request is related to ongoing work for #1804

It changes the following:

- Bumps `temporal_rs` version
- Adds stubs for all remaining methods except for `Duration.prototype.compare`, `PlainDate.prototype.toZonedDateTime` (these methods are)

The main goal here was to complete some of the remaining leg work for Temporal in Boa (outside of any remaining debugging).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3579,7 +3579,7 @@ checksum = "42a4d50cdb458045afc8131fd91b64904da29548bcb63c7236e0844936c13078"
 [[package]]
 name = "temporal_rs"
 version = "0.0.4"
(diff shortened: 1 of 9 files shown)


nekevss:
In regard to the tests, most of the legitimate passing tests should be related to branding, length, or options validation. Overall, it may not be best to put too much stock into them as the potential for false positives here is high since these are stubs.

With any luck though, there should be a decent chunk of passing tests.

jedel1043 (approved):
Looks good! Hopefully a nice chunk of the tests are true positives.

nekevss:
> Hopefully a nice chunk of the tests are true positives.

Exactly my hope, but I just don't want to put too much weight into them :laughing:

raskad (approved):

PR #4176, opened 2025-02-24
Fix JsValue::to_json with cyclic values

### LINKED ISSUE #4175: JsValue::to_json does not work with cyclic values
**Describe the bug**
`JsValue::to_json` does not work with cyclic values. For example,

```rs
let obj = JsObject::with_null_proto();
obj.create_data_property(js_string!("a"), obj.clone(), &mut context)
    .expect("should create data property");
JsValue::from(obj)
    .to_json(&mut context).unwrap();
```

gives

```
thread '...' has overflowed its stack
fatal runtime error: stack overflow
```

**To Reproduce**
Run the code snippet above.

**Expected behavior**
The same error that JSON.stringify gives in such cases:
```
Uncaught TypeError: cyclic object value
```

- OS: Ubuntu Linux
- Version: 22.04
- Rustc version: rustc 1.84.0 (9fc6b4312 2025-01-07)

**Additional context**
I already have a fix that basically aligns with JSON.stringify.

changhc:
This Pull Request fixes/closes #4175.

It changes the following:

- The actual content of `to_json` is moved to a new internal method `to_json_inner`. This internal method takes one more argument `stack` that keeps track of objects seen so far. If the current object is in the stack, it means that there's a cycle and we return a TypeError.
- `to_json` calls the internal method and returns results as usual.
- A test case is added.

### DIFF
--- a/core/engine/src/value/conversions/serde_json.rs
+++ b/core/engine/src/value/conversions/serde_json.rs
@@ -161,7 +161,7 @@ impl JsValue {
                     }
                     // Passing the object rather than its clone that was inserted to the set should be fine
                     // as they hash to the same value and therefore HashSet can still remove the clone
-                    seen_objects.remove(&obj);
+                    seen_objects.remove(obj);
                     Ok(Value::Array(arr))
                 } else {
                     let mut map = Map::new();
@@ -185,7 +185,7 @@ impl JsValue {
                         let value = value_by_prop_key(property_key, context)?;
                         map.insert(key, value);
                     }
-                    seen_objects.remove(&obj);
+                    seen_objects.remove(obj);
                     Ok(Value::Object(map))
                 }
             }
--- a/core/engine/src/value/conversions/serde_json.rs
+++ b/core/engine/src/value/conversions/serde_json.rs
@@ -7,8 +7,9 @@ use crate::{
     js_string,
     object::JsObject,
     property::{PropertyDescriptor, PropertyKey},
-    Context, JsResult, JsVariant, NativeObject,
+    Context, JsResult, JsVariant,
 };
+use indexmap::IndexSet;
 use serde_json::{Map, Value};
 
 impl JsValue {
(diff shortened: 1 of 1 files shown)


core/engine/src/value/conversions/serde_json.rs
```diff
@@ -113,6 +113,15 @@ impl JsValue {
...
     /// Panics if the `JsValue` is `Undefined`.
     pub fn to_json(&self, context: &mut Context) -> JsResult<Value> {
+        let mut stack = vec![];
+        self.to_json_inner(context, &mut stack)
+    }
+
+    fn to_json_inner(
+        &self,
+        context: &mut Context,
+        stack: &mut Vec<JsObject<dyn NativeObject>>,
```
HalidOdat, line 123:
```suggestion
        stack: &mut Vec<JsObject>,
```
Nitpick: I don't think you manually have to specify the inner type since it's default.

HalidOdat, line 123: Suggestion: You could use [`indexmap::IndexSet`](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html) (already a dependency) to speed up the check if an object is with the stack. It's set that preserves the order of inserted object so you can call `.pop()` for popping the last inserted element.

HalidOdat (approved):
Looks good to me! Just a suggestion on improving this :)

core/engine/src/value/conversions/serde_json.rs
changhc: Thanks! I thought the reason of using vec in JSON.stringify would be JSObject not being hashable, but I just realised that it implements `Hash`. We can just use a pure set/map as well (not sure IndexSet or std HashSet is faster), but I'm using IndexSet here anyway as you suggested.

HalidOdat: yes, regular `HashSet` could be used too (and it's a better choice), for some reason I though that the order mattered because it was a vec before. :sweat_smile:

changhc: Updated!

PR #4190, opened 2025-02-26
Some cleanup + order of operations fixes

nekevss:
This Pull Request related to ongoing work for #1804.

It changes the following:

- Fixes the order of operations for `to_partial_zdt`
- Removes some unused code


raskad (approved):

jedel1043:
@raskad as an FYI, we don't require two reviews to merge small PRs like this anymore (discussed this on yesterday's call), so you can merge as it is :)

PR #4191, opened 2025-02-27
Stack alloc reduce

### LINKED ISSUE #4089: `boa_tester` stack overflows in debug mode
Running the `boa_tester` in debug in the main branch aborts with a stack overflow.

image

In release mode everything seems fine, but we should still investigate.

Nikita-str:
This Pull Request closes #4089

Reduce the stack size. Mostly affected expression with nested parenthesizes.
<details>
<summary> Some examples & comparison charts</summary>

```rust
fn parenthesized_fn_test() {
    let n = ...;
    let js = "(function(){ ".repeat(n) + &" })()".repeat(n);

    let interner = &mut Interner::default();
    crate::Parser::new(crate::Source::from_bytes(&js))
        .parse_script(&boa_ast::scope::Scope::new_global(), interner)
        .expect("failed to parse");
}
```
| profile | prev max `n` <br/> without stack overflow | new max `n` <br/> without stack overflow | x |
|--|--|--|--|
| dev (`opt-level = 0`) | 11 | 79 | ~6 |
| test (`opt-level = 1`) | 50 | 160 | ~3 |
| release | 50 | 217 | ~4 |

---
```rust
fn parenthesized_sum_test() {
    let n = ...;
    let js = "(1+".repeat(n) + "41" + &")".repeat(n);

    let interner = &mut Interner::default();
    crate::Parser::new(crate::Source::from_bytes(&js))
        .parse_script(&boa_ast::scope::Scope::new_global(), interner)
        .expect("failed to parse");
}
```
| profile | prev max `n` <br/> without stack overflow | new max `n` <br/> without stack overflow | x |
|--|--|--|--|
| dev (`opt-level = 0`) | 30 | 200 | ~6.5 |
| test (`opt-level = 1`) | 105 | 390 | ~3.5 |
| release | 105 | 510 | ~4.5 |

</details>

Now there no stack overflow when running the `boa_tester` in debug(`cargo run -p boa_tester -- run -v -o test262 `). The problem was here `test262\test\language\statements\function\S13.2.1_A1_T1.js`, so you can run `cargo run -p boa_tester -- run -s test/language/statements/function/` to test the fix. For more details see the issue.

---

The reduction performed by 3 operations:
1. boxing of large objects. It's affected all `profile`s and allow to reduce the stack size by ~ 3 times;
1. In [one place](https://github.com/Nikita-str/boa/blob/df4c22fdd3db488ff23649846b86b751389ec5b9/core/parser/src/parser/expression/mod.rs#L416-L458) nested `parse` calls was unwrapped. Also affected all `profile`s;
1. [Splitting](https://github.com/Nikita-str/boa/blob/86bbd0f460210630f8b8f7bb80b745d90fffde1e/core/parser/src/parser/expression/assignment/conditional.rs#L76-L107) `parse` function. Affected only `profile.dev` (`opt-level = 0`) where the problem arose. Allow to reduce the stack size by ~ 2 times (and with boxing it becomes ~ 6 times).

---

There are some ugly changes that need to be discussed & maybe rewritten to a less ugly:
* The [unwrapping](https://github.com/Nikita-str/boa/blob/df4c22fdd3db488ff23649846b86b751389ec5b9/core/parser/src/parser/expression/mod.rs#L416-L458);
* Splitting like [this](https://github.com/Nikita-str/boa/blob/86bbd0f460210630f8b8f7bb80b745d90fffde1e/core/parser/src/parser/expression/assignment/mod.rs#L100-L102) --> [this](https://github.com/Nikita-str/boa/blob/86bbd0f460210630f8b8f7bb80b745d90fffde1e/core/parser/src/parser/expression/assignment/mod.rs#L132) --> [this](https://github.com/Nikita-str/boa/blob/86bbd0f460210630f8b8f7bb80b745d90fffde1e/core/parser/src/parser/expression/assignment/mod.rs#L113-L121) --> [this](https://github.com/Nikita-str/boa/blob/86bbd0f460210630f8b8f7bb80b745d90fffde1e/core/parser/src/parser/expression/assignment/mod.rs#L214).

All of the changes reduce the stack size, so even if they are ugly, they are still not meaningless :/

### DIFF
--- a/core/parser/src/error/mod.rs
+++ b/core/parser/src/error/mod.rs
@@ -37,9 +37,98 @@ impl From<LexError> for Error {
     }
 }
 
+impl From<LexError> for ErrorInner {
+    #[inline]
+    fn from(e: LexError) -> Self {
+        Self::lex(e)
+    }
+}
+
+/// A struct which represents errors encountered during parsing an expression.
+/// Contains boxed `ErrorInner` to reduce memory allocation on the stack.
+#[derive(Debug)]
+pub struct Error {
+    inner: Box<ErrorInner>,
+}
+
+impl Error {
+    /// Changes the context of the error, if any.
+    fn set_context(self, new_context: &'static str) -> Self {
+        Self {
+            inner: Box::new((*self.inner).set_context(new_context)),
+        }
+    }
+
+    /// Gets the context of the error, if any.
+    fn context(&self) -> Option<&'static str> {
+        self.inner.context()
+    }
+
+    /// Creates an `Expected` parsing error.
+    pub(crate) fn expected<E, F>(expected: E, found: F, span: Span, context: &'static str) -> Self
+    where
+        E: Into<Box<[String]>>,
+        F: Into<Box<str>>,
+    {
+        ErrorInner::expected(expected, found, span, context).into()
+    }
+
+    /// Creates an `Unexpected` parsing error.
+    pub(crate) fn unexpected<F, C>(found: F, span: Span, message: C) -> Self
+    where
+        F: Into<Box<str>>,
+        C: Into<Box<str>>,
+    {
+        ErrorInner::unexpected(found, span, message).into()
+    }
+
+    /// Creates a "general" parsing error.
+    pub(crate) fn general<S, P>(message: S, position: P) -> Self
+    where
+        S: Into<Box<str>>,
+        P: Into<Position>,
+    {
+        ErrorInner::general(message, position).into()
+    }
+
+    /// Creates a "general" parsing error with the specific error message for a misplaced function declaration.
+    pub(crate) fn misplaced_function_declaration(position: Position, strict: bool) -> Self {
+        ErrorInner::misplaced_function_declaration(position, strict).into()
+    }
+
+    /// Creates a "general" parsing error with the specific error message for a wrong function declaration with label.
+    pub(crate) fn wrong_labelled_function_declaration(position: Position) -> Self {
+        ErrorInner::wrong_labelled_function_declaration(position).into()
+    }
+
+    /// Creates a parsing error from a lexing error.
+    pub(crate) fn lex(e: LexError) -> Self {
+        ErrorInner::lex(e).into()
+    }
+
+    /// Creates a parsing error from a lexing error.
+    pub(crate) fn abrupt_end() -> Self {
+        ErrorInner::AbruptEnd.into()
+    }
+}
+
+impl From<ErrorInner> for Error {
+    fn from(value: ErrorInner) -> Self {
+        Self {
+            inner: Box::new(value),
+        }
+    }
+}
+
+impl From<Error> for ErrorInner {
+    fn from(value: Error) -> Self {
+        *value.inner
+    }
+}
+
 /// An enum which represents errors encountered during parsing an expression
 #[derive(Debug)]
-pub enum Error {
+pub enum ErrorInner {
     /// When it expected a certain kind of token, but got another as part of something
     Expected {
         /// The token(s) that were expected.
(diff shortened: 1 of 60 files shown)


hansl:
I would like to see the performance impact of this one, in release. Would you mind running the v8 combined benchmarks?

```sh
cargo run -p boa_cli --release -- -O $SRC/boa-dev/data/bench/bench-v8/combined.js
```

(with `$SRC` being wherever your root for `boa-dev/data` repo is)

Nikita-str:
<details>
<summary> Current results of the bench. </summary>
 
```
PROGRESS Richards
RESULT Richards 72.8
PROGRESS DeltaBlue
RESULT DeltaBlue 64.4
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 144
PROGRESS RayTrace
RESULT RayTrace 196
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 229
PROGRESS RegExp
RESULT RegExp 61.0
PROGRESS Splay
RESULT Splay 355
PROGRESS NavierStokes
RESULT NavierStokes 337
SCORE 147
undefined
```
</details>

<details>
<summary> Previous(checkout 35b44d9ce07efb45fb094887964d473911b21012) results of the bench. </summary>

```
PROGRESS Richards
RESULT Richards 76.0
PROGRESS DeltaBlue
RESULT DeltaBlue 64.8
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 140
PROGRESS RayTrace
RESULT RayTrace 190
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 230
PROGRESS RegExp
RESULT RegExp 61.2
PROGRESS Splay
RESULT Splay 357
PROGRESS NavierStokes
RESULT NavierStokes 334
SCORE 147
undefined
```
</details>

Seems like a noise error.  And it's pretty predictable, because as I can understand the bench tests the perf of js execution, and PR's changes affect only the parser stage. So is there a bench to test perf of the parser stage?

hansl:
Nice! I'll review when I have some time this weekend.

core/parser/src/error/mod.rs
```diff
@@ -37,9 +37,98 @@ impl From<LexError> for Error {
...
+    #[inline]
+    fn from(e: LexError) -> Self {
+        Self::lex(e)
+    }
+}
+
+/// A struct which represents errors encountered during parsing an expression.
+/// Contains boxed `ErrorInner` to reduce memory allocation on the stack.
+#[derive(Debug)]
+pub struct Error {
```
hansl, line 50: Does this make a difference? Note that the previous `Error` is rather small (32 bytes?), and this code should not clone errors themselves anyway.

BTW, still reviewing the rest, I haven't forgotten about you.

Nikita-str: Previous `Error` (that now named `ErrorInner`) when compiled with 8 bytes sized pointer has a size of 72 bytes:  
`Expected` variant has 3 fat pointers(`2 * sizeof(ptr)` each) + Span(`16 bytes`) =` 3 * 2 * 8 + 16` = `64 bytes` + (because it's not only variant of the enum) its maximum align (`8 bytes`) = `72 bytes`; In case of 4 bytes sized pointer = `3 * 2 * 4 + 16 + 4(align)` = `44 bytes`.

For `Result<T, E>` the allocated stack size will be:
* `Result<Box<Expr>, ErrorInner>` --> 72 bytes (no additional align because of null pointer optimization).
* `Result<Box<Expr>, Error>` --> 16 bytes.

So, to me it seems ok to have an extra boxed value in case of an error, but save at least 56 bytes on stack per a call(with low optimization it for sure will be more than 56 bytes because of stack allocation for each call in different execution branches). `... (function()( ...` has ~20 calls per one recursion, so this changes at least save 56 * 20 = 1120 bytes per one recursive repetition of the expression.

Nikita-str:
> Another question I would have is whether the new `_boxed` constructors are necessary, instead of simply replacing the regular constructors. WDYT?

Hmm... `new_boxed` shows that the return value is not `Self` but `Box<Self>`; and sometimes there is another code which does not need `Box<T>`.

PR #4192, opened 2025-02-27
Unify release workflows

### LINKED ISSUE #4186: Unify binary artifacts from releases and nightly
We publish plain binaries for releases but tars for nightly build. We should unify both into a single release format (either tar.gz or plain binaries), and use a single nomenclature for them (possibly the target triplet).

HalidOdat:
Fixes #4186


jedel1043 (approved):
LGTM!

PR #4215, opened 2025-03-23
Change signature of `from_async_fn` to allow capturing the context

jedel1043:
This allows carrying the `context` through await points, making it much more ergonomic to use for big async functions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ exclude = [
 [workspace.package]
 edition = "2021"
 version = "0.20.0"
-rust-version = "1.84.0"
+rust-version = "1.85.0"
 authors = ["boa-dev"]
 repository = "https://github.com/boa-dev/boa"
 license = "Unlicense OR MIT"
--- a/core/engine/src/native_function/mod.rs
+++ b/core/engine/src/native_function/mod.rs
@@ -3,8 +3,11 @@
 //! [`NativeFunction`] is the main type of this module, providing APIs to create native callables
 //! from native Rust functions and closures.
 
+use std::cell::RefCell;
+
 use boa_gc::{custom_trace, Finalize, Gc, Trace};
 
+use crate::job::NativeAsyncJob;
 use crate::value::JsVariant;
 use crate::{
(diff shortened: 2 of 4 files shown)


Razican:
Wow, the cool thing about this is that we can use non-desugared normal async functions:
```rust
async fn get_record(
    _this: &JsValue,
    _args: &[JsValue],
    context: &RefCell<&mut Context>,
) -> JsResult<JsValue> {
    todo!()
}
```

jedel1043:
@Razican That's always been the case if you don't use the arguments inside the function. The moment you try to get any argument, the borrow checker complains that you're capturing a lifetime for too long.

Also, I think this is only valid for edition 2021, because on edition 2024 all async functions capture the lifetime of all arguments by default.

Razican:
> @Razican That's always been the case if you don't use the arguments inside the function. The moment you try to get any argument, the borrow checker complains that you're capturing a lifetime for too long.
> 
> Also, I think this is only valid for edition 2021, because on edition 2024 all async functions capture the lifetime of all arguments by default.

Hmmm interesting, I'm able to use it with edition 2024, using the arguments and so on:

```rust
/// Gets a record from the database.
    async fn get(
        this: &JsValue,
        args: &[JsValue],
        context: &RefCell<&mut Context>,
    ) -> JsResult<JsValue> {
        let sys_id = args.get_or_undefined(0).as_string().ok_or_else(|| {
            JsNativeError::typ()
                .with_message("you must specify a record ID to fetch from the database")
        });
        let realm = context.borrow().realm().clone();

        todo!() // actually some more code that uses the host_defined() from realm and so on
    }
```

jedel1043:
@Razican Ohhhh, maybe they changed the desugaring of async functions?

HalidOdat (approved):
Looks great! This greatly improves the API's ergonomics. :)

Razican (approved):
This is really great :) I also saw some comments in Matrix about possibly hiding the `RefCell` and have `Context` use interior mutability and cheap cloning, but for now, this improves the API quite a lot.

Issue #3905, opened 2024-07-08
Merge `boa_interop` into other crates


jedel1043:
Discussion: https://github.com/boa-dev/boa/pull/3903#discussion_r1669419557

We decided to merge it for now for discoverability and because we will probably split the engine into several crates in the future.

jasonwilliams:
@jedel1043 can this be closed?

jedel1043:
Not yet, we still need to move the module loader implementations

PR #4218, opened 2025-03-27
Move interop and module utilities from boa_interop into boa_engine

hansl:
Migrated module and function interop traits and JS argument handling functionality into the `boa_engine` crate for better modularity.

Deprecated certain exports in `boa_interop` and updated references across the project to reflect the changes, to maintain backward compatibility, at least until we kill `boa_interop` for good.

Refers #3905


HalidOdat (approved):
Looks good to me! :)

PR #4219, opened 2025-03-28
 Replace NanBoxing's Box with a refcount and improve clone/drop performance

hansl:
Essentially what I'm doing here is instead of using a `Box` as the container for the `JsValue`'s inner NaN-boxed pointer, I use an RC. When cloning the value, that RC will the increment (and decrement on drop) instead of allocating a new `Box` on the heap. This should make clones of `JsValue` simply increment an integer in memory.

So given the following:

```rs
/* 1 */ let a = JsObject::new();
/* 2 */ let b = JsValue::from(a.clone());
/* 3 */ let c = b.clone();
/* 4 */ drop(b);
/* 5 */ drop(a);
/* 6 */ drop(c);
```

The refcount of the `JsObject` itself will be: 1 at line 1, 2 at line 2, still 2 at line 3 as the clone of the `JsValue` itself increment the internal counter, 2 at line 4, 1 at line 5 (`a` will drop, but the `JsValue` itself will still have an inner `JsObject`), then finally 0 and dropped at line 6.

### DIFF
--- a/core/engine/src/value/inner/nan_boxed.rs
+++ b/core/engine/src/value/inner/nan_boxed.rs
@@ -107,6 +107,30 @@
 //!
 //! The pointers are assumed to never be NULL, and as such no clash
 //! with regular NAN should happen.
+//!
+//! # Cloning / Dropping values
+//!
+//! When creating the NaN-Boxed value, a refcounted pointer is allocated
+//! on the heap. When cloning that value, the internal RC will increment
+//! (and decrement on drop) instead of allocating a new pointer. This
+//! should make clones of `JsValue` simply increment an integer in memory.
+//!
+//! Given the following:
+//!
+//! ```rs
+//! /* 1 */ let a = JsObject::new();
+//! /* 2 */ let b = JsValue::from(a.clone());
+//! /* 3 */ let c = b.clone();
+//! /* 4 */ drop(b);
+//! /* 5 */ drop(a);
+//! /* 6 */ drop(c);
+//! ```
+//!
+//! The refcount of the `JsObject` itself will be: 1 at line 1, 2 at
+//! line 2, still 2 at line 3 as the clone of the `JsValue` itself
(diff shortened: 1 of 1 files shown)


core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -296,8 +321,8 @@ mod bits {
     /// by calling `[Self::drop_pointer]`.
     #[inline(always)]
     #[allow(clippy::identity_op)]
-    pub(super) unsafe fn tag_bigint(value: Box<JsBigInt>) -> u64 {
-        let value = Box::into_raw(value) as u64;
+    pub(super) unsafe fn tag_bigint(value: Rc<JsBigInt>) -> u64 {
```
HalidOdat, line 324: Suggestion: Maybe we should pass `JsBigInt` as value directly since it's wrapped in an `Rc<RawBigInt>` already internally, and add methods for `into_raw()`/`from_raw()`.

hansl: This cannot work unless I change the interface of the JsValue to return `Rc<...>` (or similar) instead of references. To return a `&'a JsBigInt` with your suggested change I'd need to reconstruct it on the stack, and I cannot return a reference to _that_. Unless I'm missing something?

HalidOdat: I see, sorry I missed that :sweat_smile: , we could get around this by using a bit of safe "unsafe" code by adding `#[repr(transparent)]` on `JsBigInt` and `transmute`-ing `&'a Rc<RawBigInt>` into `&'a JsBigInt` JsBigInt.

For more info see: https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent

hansl: I kind of tried that. I need to spend more time on it, but I was still getting segfaults accessing freed or invalid memory.

PR #4236, opened 2025-05-07
Bump temporal_rs and add ZonedDateTime.prototype.round impl

nekevss:

Particularly exciting is that this PR removes the final "Not yet implemented" stub from Boa's side (I think `temporal_rs` still has 4-5 methods left, but majority have issues and/or PRs opened for them). 

It changes the following:

- Bumps `temporal_rs`
- Adds ZonedDateTime.prototype.round impl
- Changes related to updates in `temporal_rs`, specifically `Duration`'s move away from `f64` internally

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3637,7 +3637,7 @@ checksum = "1ac9aa371f599d22256307c24a9d748c041e548cbf599f35d890f9d365361790"
 [[package]]
 name = "temporal_provider"
 version = "0.0.6"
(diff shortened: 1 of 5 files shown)


nekevss:
Broken tests are known about and potentially addressed in follow ups. At least one is related to floating point beyond the MAX_SAFE_INTEGER with potentially all 3 being related to that.

raskad (approved):

Issue #3442, opened 2023-10-30
Cancel/Interrupt active evaluation


mattsse:
**feature**
I would like to be able to cancel an evaluation (or `JsValue::call`) after a certain timeout.
Perhaps this is already possible, but I wasn't able to find it.

boja for example has `interrupts`:

https://github.com/dop251/goja/blob/b396bb4c349df65109dea3df00fb60f6a044950d/runtime.go#L1474C9-L1482

my motivation for this is to have the option to terminate a (malicious) evaluation for example `while(1) {}`

**Example code**
Give a code example that should work after the implementation of this feature.

jedel1043:
There isn't a hook to do this at the moment, but we have plans to support this. The simplest API I can think of is a way to set a flag for the `Context` from another thread that marks it as `interrupted`. Would that be enough to support your use case?

mattsse:
cool.

> The simplest API I can think of is a way to set a flag for the Context from another thread that marks it as interrupted

yeah, this is exactly what I'm looking for, like `Interrupt(Arc<AtomicBool>)` or equivalent.

jedel1043:
> yeah, this is exactly what I'm looking for, like `Interrupt(Arc<AtomicBool>)` or equivalent.

Alright, I'll put this on our backlog then.

felipefdl:
It's difficult to use Boa for running untrusted third-party code without a way to interrupt eval.

I appreciate the simplicity and effectiveness of rquickjs's set_interrupt_handler method for handling this issue. It's a great solution for many use cases.

Here's an example showing how a timeout could be implemented using set_interrupt_handler:

```rust
let timestamp_now = std::time::SystemTime::now();

rt.set_interrupt_handler(Some(Box::new(move || {
  if let Ok(elapsed) = timestamp_now.elapsed() {
    if elapsed.as_millis() > 1000 {
      return true;
    }
  }

  false
})));
```

rib:
It could be good if any solution to interrupting evaluation didn't depend on threads, in case you might want to embed the Boa runtime in a wasm/browser environment without threads. 

[...]

PR #4237, opened 2025-05-09
Prototype support for async native functions

rib:
Based on some discussion in https://github.com/boa-dev/boa/issues/3442#issuecomment-2859413836; this experiments with enabling support for async NativeFunctions that are only async from the pov of the host, and appear as synchronous from within JavaScript.

Instead of running the async functions as a Promise via `enqueue_job`, this works by allowing Operations to be executed over multiple VM cycles, so an Operation may start some async work in one step and then further steps can poll for completion of that work and finish the Operation.

In particular this works by allowing Call Operations to return a `NormalPending` completion type that indicates that the same Call operation needs to be executed repeatedly, until it returns a `Normal` status.

In the case of a `NormalPending` status, the program counter is reset and anything that was taken off the stack is pushed back so the same Operation can be re-executed.

There is a new `NativeFunction::from_async_as_sync_with_captures()` that lets the host provide a (sync) closure that itself returns / spawns a boxed Future. This is tracked internally as an `Inner::AsyncFn`.

Whenever the function is `__call__`ed then (assuming the operation isn't already in a pending / running state) a new Future is spawned via the application's closure and the Operation enters a "pending" state.

When a NativeFunction is pending then each `__call__` will `poll()` the spawned `Future` to see if the `async` function has a result.

This effectively stalls the VM at the same Opcode while still accounting for any cycle budget and periodically yielding to the application's async runtime while waiting for an async Call Operation to finish.

Limitations / Issues
====================

## Busy Loop Polling

Even though the implementation does yield back to the application's async runtime when waiting for a NativeFunction to complete, the implementation isn't ideal because it uses a noop task Context
+ Waker when polling NativeFunction Futures. The effectively relies on the VM polling the future in a busy loop, wasting CPU time.

A better solution could be to implement a shim Waker that would flag some state on the Boa engine Context, and then adapt the Future that's used to yield the VM to the executor so that it only becomes Ready once the async NativeFunction has signalled the waker. I.e. the Waker would act like a bridge/proxy between a spawned async NativeFunction and the the Future/Task associated with the VM's async `run_async_with_budget`.

This way I think the VM could remain async runtime agnostic but would be able to actually sleep while waiting for async functions instead of entering a busy yield loop.

## Rewinding relies on coop between VM an Operation

Ideally there could be a more elegant way to reliably rewind the side effects of a Pending Operation, or else avoid the need to rewind anything and instead track partial progress for an Operation.

The support for rewiding a Call Operation partly relies on the VM handling resetting of the program counter while the operation itself handles pushing arguments back to the stack.

## Only adapts Call Operation

Currently only the Call Operation handles async NativeFunctions but there are other `Call[XYZ]` Operations that could be adapted too.

## Not compatible with composite Operations that `call()`

The ability to track pending async functions is implemented in terms of repeatedly executing an Opcode in the VM until it signals that it's not Pending.

This relies on being able to reset and re-execute the Operation (such as reverting program counter and stack changes).

There are lots of Operations that make use of `JsObject::call()` internally and they would currently trigger a panic if they called an `async` `NativeFunction` because they wouldn't not be able to "resolve()" the "AsyncPending" status that would be returned by the `call()`.

Ideally all Operations that use `__call__` or `__construct__` should support `CallValue::AsyncPending` and be fully resumable in the same way that the `Call` `Operation` is now.

This would probably be easier to achieve this with Rust Coroutines if they were stable because it would otherwise be necessary to adapt the Operation's execution into a state machine, similar to what the compiler does for an async Future. It probably wouldn't be ideal to be spawning futures directly at the level of `Operation::execute` while that would likely have a disproportionate overhead. Maybe a trade-off would be that Operations that use `call()` could be manually adapted into `match` state machines and the VM would track an integer for progressing resumable Operations - but that would notably complicate the implementation for a lot of (most?) operations.

Addresses: #3442

### DIFF
--- a/core/engine/src/native_function/mod.rs
+++ b/core/engine/src/native_function/mod.rs
@@ -4,8 +4,12 @@
 //! from native Rust functions and closures.
 
 use std::cell::RefCell;
+use std::future::Future;
+use std::pin::Pin;
+use std::rc::Rc;
 
 use boa_gc::{custom_trace, Finalize, Gc, Trace};
+use futures_lite::FutureExt as _;
 
 use crate::job::NativeAsyncJob;
 use crate::value::JsVariant;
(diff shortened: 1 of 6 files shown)


core/engine/src/native_function/mod.rs
```diff
@@ -108,6 +129,148 @@ impl JsData for NativeFunctionObject {
...
+                        AsyncCallResult::Pending
+                    }
+                    std::task::Poll::Ready(result) => {
+                        *state = AsyncCallState::None;
+                        AsyncCallResult::Ready(result)
+                    }
+                }
+            }
+        }
+    }
```
jedel1043: Sorry for the delay! I've been repeatedly looking at this PR again and I wondered, Couldn't we store a map of the "active" wakers in the `Context` itself, then try to poll all of them whenever the `Context` is trying to resume execution?

rib: No worries - it's does seem fiddly to figure out.

I agree that this current polling approach isn't good enough as is

I think some notable issues are that:
1. It doesn't feel great to be materializing a special waker + context while these are things that you would normally expect your async runtime to own (i.e. tokio)
2. I think there's probably a risk of futures polled this way not integrating with your async runtime properly. Tokio sleep did seem to work for me when testing but futures that depend on mio polling might not get handled properly - I'm not 100% sure.
3. This approach effectively leads to a non-blocking busy loop on the CPU that will keep polling for async native functions to complete

In terms of:
> Couldn't we store a map of the "active" wakers in the Context itself, then try to poll all of them whenever the Context is trying to resume execution?

I think my instinct is that, ideally this wouldn't involve any custom wakers at all.

The thing we have to poll currently is a boxed future which is currently saved in the `NativeAsyncFunction::state`, based on an assumption that any particular function can only have one active call at a time, but maybe it would make sense to store this operation state in the `Context`. It's conceptually like an internal register state for the operation.

If we could see the active/pending future from an `async` context, outside of the sync parts of the VM implementation (such as in `run_async_with_budget`) then we could do an `.await` there and avoid the need for any ad-hoc waker/context for manually polling.

That sounds like it could work, and hopefully address all the issues above.


... assuming it's possible to improve how the async native function futures are polled, I'd be curious to know what you think about the general idea of Operations having a Pending status that allows Operations to take multiple VM cycles to execute?

I think I was concerned about the possibility that other Operations may use `call` as an implementation detail.

I'm not entirely sure how practical it would be to audit the Operations that use `call` to ensure that they can all, also support returning a Pending status and be resumed after the async function finishes.

Or otherwise I'm not sure if it's feasible to assume that those operations will never call a native async function.

Without something like stable coroutines in Rust it seems like it might be a pain to have to manually support more-complex resumable operations that involve `call`s internally.

Maybe if there are only a small handful of operations that could internally call an async native function then it would be manageable to create a `match` based state machine manually to allow them to be resumable.

rib: I've taken a pass at reworking this to avoid the need for manually polling futures. The boxed futures are now stashed in the `Context` and `.await` is used from `Context::run_async_with_budget` to wait for any native async function to complete.

This should avoid async runtime compatibility concerns (from polling with a task::Context and Waker that wasn't managed by the current runtime) and should also avoid polling for native function completion in a busy loop.

Ref: https://github.com/boa-dev/boa/pull/4237/commits/b97ab83a51b50e574b488f160b9b1009c49a479c

PR #4248, opened 2025-05-14
Bump temporal_rs and remove dead code

nekevss:

It completes some general updates for `temporal_rs` that should adapt for the upcoming 0.0.7 release.

If 0.0.7 releases, it may be worthwhile to switch from a git commit to the version directly.


nekevss:
Canonicalization errors here are interesting. They need to be looked into about custom handling. Although, this is going to depend on the int-era-month-code proposal (islamic calendars were renamed to hijri to be more accurate).

The invalid offsets appear to be a regression, but general time zone parsing in `temporal_rs` is currently in flux as it is slowly vendored to `ixdtf`. So those will need to be addressed over time.

PR #4249, opened 2025-05-16
Fix Rust 1.87.0 lints

raskad:
Fix Rust 1.87.0 lints

### DIFF
--- a/.github/workflows/webassembly.yml
+++ b/.github/workflows/webassembly.yml
@@ -55,7 +55,7 @@ jobs:
         with:
           crate: wasm-pack
       - name: Build Playground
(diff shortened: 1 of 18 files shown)


raskad:
Not sure what is happening with wasm-opt right now. I would be for merging this so that other pipelines can work again and creating an issue to look into it.

nekevss (approved):
I was literally about to ask about the `--no-opt` flag. Fine by me if there's an issue opened :smile:

nekevss:
Huh, looks like `enable-bulk-memory` may need to be provided to wasm-opt somehow?

FWIW, https://github.com/rustwasm/wasm-pack/issues/1441

PR #4261, opened 2025-05-28
Add support for LZ4 compression in embedded module loader

hansl:
This results in much smaller binaries when embedding javascript. The compression is hidden behind a flag and by default will not compress, so there is no breaking change.


HalidOdat (approved):
Neat feature! Looks good to me! :D

PR #4263, opened 2025-05-29
Refactor registers to use the stack

raskad:

- Refactor the vm stack into a dedicated struct and implement stack access in functions that are more readable.
- Remove the `Registers` data structure and store registers on the stack.


raskad:
Running the benchmarks locally I noticed a bit of a performance reduction. As far as I can see, it is a result of the additional clones in some opcodes. Those add refcount increases and drop glue with refcount decreases.
When we switch to a non refcount gc, this should sort itself out.

HalidOdat (approved):
Amazing work! Really love the refactor :star_struck:

PR #4266, opened 2025-06-03
Escape function scopes on regular functions

HalidOdat:
This PR omits functions scopes that that are not needed for regular functions, currently a WIP (some test262 tests fail).

We can see a decent performance increase :partying_face: 
(Still some work to be done but those bugs shouldn't affect the benchmarks code, since it doesn't use the bugged features).

#### main

```
PROGRESS Richards
RESULT Richards 110
PROGRESS DeltaBlue
RESULT DeltaBlue 103
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 140
PROGRESS RayTrace
RESULT RayTrace 277
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 339
PROGRESS RegExp
RESULT RegExp 71.6
PROGRESS Splay
RESULT Splay 528
PROGRESS NavierStokes
RESULT NavierStokes 336
SCORE 193
undefined
```

#### PR

```
PROGRESS Richards
RESULT Richards 136
PROGRESS DeltaBlue
RESULT DeltaBlue 138
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 136
PROGRESS RayTrace
RESULT RayTrace 319
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 375
PROGRESS RegExp
RESULT RegExp 69.4
PROGRESS Splay
RESULT Splay 577
PROGRESS NavierStokes
RESULT NavierStokes 324
SCORE 211
undefined
```


HalidOdat:
This is ready for review/merge :tada: 

----

### Latest Benchmarks

#### main

```
PROGRESS Richards
RESULT Richards 107
PROGRESS DeltaBlue
RESULT DeltaBlue 101
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 140
PROGRESS RayTrace
RESULT RayTrace 276
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 336
PROGRESS RegExp
RESULT RegExp 68.8
PROGRESS Splay
RESULT Splay 519
PROGRESS NavierStokes
RESULT NavierStokes 339
SCORE 190
```

#### PR

```
PROGRESS Richards
RESULT Richards 142
PROGRESS DeltaBlue
RESULT DeltaBlue 140
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 137
PROGRESS RayTrace
RESULT RayTrace 323
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 373
PROGRESS RegExp
RESULT RegExp 71.5
PROGRESS Splay
RESULT Splay 576
PROGRESS NavierStokes
RESULT NavierStokes 337
SCORE 215
```

core/ast/src/scope.rs
```diff
@@ -277,6 +287,27 @@ impl Scope {
         }
     }
 
+    /// Escape enclosing function environment's `this`.
+    pub fn escape_this_in_enclosing_function_scope(&self) {
```
jasonwilliams, line 291: This is only called when `self.in_arrow()`, is it needed?

core/engine/src/bytecompiler/function.rs
```diff
@@ -95,6 +97,12 @@ impl FunctionCompiler {
         self
     }
 
+    /// Indicate if the function is in a `with` statement.
+    pub(crate) const fn force_function_scope(mut self, force_function_scope: bool) -> Self {
+        self.force_function_scope = force_function_scope;
```
jasonwilliams, line 102: could the property have a different name to the method? like `forced_function_scope` or something?

cli/src/debug/function.rs
```diff
@@ -110,7 +110,9 @@ fn bytecode(_: &JsValue, args: &[JsValue], _: &mut Context) -> JsResult<JsValue>
     };
     let code = function.codeblock();
 
-    Ok(js_string!(code.to_string()).into())
+    println!("{code}");
+
+    Ok(JsValue::undefined())
```
jasonwilliams, line 115: Could you explain the change made here?

jasonwilliams (approved):
This is even better than my suggestion which was to not GC these function environments. Instead they don't need to be created at all because if the bindings don't escape they are just `poppedToLocal` anyway and don't pollute the environment above.

core/ast/src/scope.rs
HalidOdat: That's true, I kept it there in case we need it elsewhere, and grouped the methods based on similar functionality.

core/engine/src/bytecompiler/function.rs
HalidOdat: I kept it that way for consistency with the other boolean methods, such as in_with() and strict(). Since Rust stores method and field names separately, it doesn't cause any issues :smile:

cli/src/debug/function.rs
HalidOdat: This is just a QoL change, it print the `CodeBlock` directly to `stdout`, instead of having to wrap it in a `console.log`

core/engine/src/bytecompiler/function.rs
jasonwilliams: It was more because force_function_scope makes it sound like it hasn’t happened yet, but by the time you check the Boolean it’s already been forced.

HalidOdat: Hmmm... the force occurs when the boolean is accessed, which overrides `requires_function_scope` in the compile() function (needed to make functions created by `Function()` have the function scope). Using the present tense seems more accurate in this context, IMO.

raskad (approved):
Really nice :)

PR #4267, opened 2025-06-04
Fix AST `contains` operation

HalidOdat:
While working on #4266, found a bug in the AST `contains` operation where we did not use the overridden method which checks for `this` and `new.target`, which always gave false even if they where there. This only happened if it was a direct expression.

It changes the following:

- Remove unknown clippy lint
- Fix contains to search using overridden `visit_expersion` for call and `with` statements
- Add tests, since test262 does not cover it.


raskad (approved):
Very nice find

PR #4271, opened 2025-06-05
Add a #[boa_class] proc macro attribute

hansl:
This transforms an `impl XYZ` block into a class. It might not be fully fool proof, and it doesn't implement everything that can be done in the Class implementation, but it is sufficient to get started.

This should reduce the code and complexity of class declaration. It supports accessors, constructors and methods, and uses the same logic for transforming arguments from `JsValue` that is used by `IntoJsFunctionCopied`. This makes it very convenient.

It also copy the `impl` block verbatim so methods inside the block can be used in Rust as regular methods.

There is an additional new functionality that isn't supported by `js_class!`: you can use `&self` or `&mut self` in your method to get a borrowed struct instance. That makes it very convenient to declare regular Rust methods. Unfortunately, it also allows for double borrows if the method calls anything into JavaScript again. I failed to find a proper way to communicate this, outside of documentation.

### DIFF
--- a/core/engine/src/value/mod.rs
+++ b/core/engine/src/value/mod.rs
@@ -29,7 +29,7 @@ pub use self::{
     variant::JsVariant,
 };
 use crate::builtins::RegExp;
-use crate::object::{JsFunction, JsPromise, JsRegExp};
+use crate::object::{ErasedObject, JsFunction, JsPromise, JsRegExp};
 use crate::{
     builtins::{
         number::{f64_to_int32, f64_to_uint32},
@@ -167,6 +167,17 @@ impl JsValue {
     #[must_use]
     pub fn as_downcast_ref<T: NativeObject>(&self) -> Option<boa_engine::object::Ref<'_, T>> {
         self.as_object().and_then(|o| o.downcast_ref::<T>())
+    }
+
+    /// Returns a downcasted mut ref object if the type matches. This is a shorthand
+    /// for `value.as_object().and_then(|o| o.downcast_ref<T>())`, which at time
+    /// can be contriving.
+    #[inline]
+    #[must_use]
+    pub fn as_downcast_mut<T: NativeObject>(
+        &self,
+    ) -> Option<boa_engine::object::RefMut<'_, ErasedObject, T>> {
+        self.as_object().and_then(|o| o.downcast_mut::<T>())
     }
 
     /// Consumes the value and return the inner object if it was an object.
--- a/core/interop/tests/class.rs
+++ b/core/interop/tests/class.rs
@@ -1,0 +1,78 @@
+//! Test for the class proc-macro.
+#![allow(unused_crate_dependencies)]
+
+use boa_engine::{js_str, js_string, Context, JsString, Source};
+use boa_macros::{boa_class, Finalize, JsData, Trace};
+
+#[derive(Clone, Trace, Finalize, JsData)]
+enum AnimalType {
+    Cat,
+    Dog,
+    Other,
+}
+
+#[derive(Clone, Trace, Finalize, JsData)]
+struct Animal {
+    ty: AnimalType,
+    age: i32,
+}
+
+#[boa_class]
+#[boa(debug)]
+impl Animal {
+    #[boa(constructor)]
+    fn new(name: String, age: i32) -> Self {
+        let ty = match name.as_str() {
+            "cat" => AnimalType::Cat,
+            "dog" => AnimalType::Dog,
+            _ => AnimalType::Other,
+        };
+
+        Self { ty, age }
+    }
+
+    #[boa(getter)]
+    fn age(&self) -> i32 {
+        self.age
+    }
+
+    #[boa(setter)]
+    #[boa(name = "age")]
+    fn set_age(&mut self, age: i32) {
+        self.age = age;
+    }
+
+    fn speak(#[boa(error = "`this` was not an animal")] &self) -> JsString {
+        match self.ty {
+            AnimalType::Cat => js_string!("meow"),
+            AnimalType::Dog => js_string!("woof"),
+            AnimalType::Other => js_string!(r"¯\_(ツ)_/¯"),
+        }
+    }
+}
+
+#[test]
+fn boa_class() {
+    let mut context = Context::default();
+
+    context.register_global_class::<Animal>().unwrap();
+
+    let result = context
+        .eval(Source::from_bytes(
+            r#"
+         let pet = new Animal("dog", 3);
+         if (pet.age !== 3) {
+            throw "age should be 3";
+         }
+         pet.age = 4;
+
+        `My pet is ${pet.age} years old. Right, buddy? - ${pet.speak()}!`
+     "#,
+        ))
+        .expect("Could not evaluate script");
+
+    assert_eq!(
+        result.as_string().unwrap(),
+        &js_str!("My pet is 4 years old. Right, buddy? - woof!")
+    );
+}
--- a/core/macros/src/class.rs
+++ b/core/macros/src/class.rs
@@ -1,0 +1,624 @@
+use proc_macro::TokenStream;
+use proc_macro2::{Span as Span2, Span, TokenStream as TokenStream2};
+use quote::{quote, ToTokens};
+use std::collections::BTreeMap;
+use std::fmt::{Display, Formatter};
+use syn::parse::{Parse, ParseStream};
+use syn::punctuated::Punctuated;
+use syn::spanned::Spanned;
+use syn::visit_mut::VisitMut;
+use syn::{
+    Attribute, Expr, ExprLit, FnArg, Ident, ImplItemFn, ItemImpl, Lit, Meta, MetaNameValue,
+    PatType, Path, Receiver, ReturnType, Token, Type,
+};
+
+type SpannedResult<T> = Result<T, (Span2, String)>;
+
+/// A function to make it easier to return error messages.
+fn error<T>(span: &impl Spanned, message: impl Display) -> SpannedResult<T> {
+    Err((span.span(), message.to_string()))
+}
+
+/// Look (and remove from AST) a `path` version of the attribute `boa`, e.g. `#[boa(something)]`.
+fn take_path_attr(attrs: &mut Vec<Attribute>, name: &str) -> bool {
+    if let Some((i, _)) = attrs
+        .iter()
+        .enumerate()
+        .filter(|(_, a)| a.meta.path().is_ident("boa"))
+        .filter_map(|(i, a)| a.meta.require_list().ok().map(|nv| (i, nv)))
+        .filter_map(|(i, m)| m.parse_args::<Path>().ok().map(|p| (i, p)))
+        .find(|(_, path)| path.is_ident(name))
+    {
+        attrs.remove(i);
+        true
+    } else {
+        false
+    }
+}
+
+/// Look (and remove from AST) for a `#[boa(name = ...)]` attribute, where `...`
+/// is a literal. The validation of the literal's type should be done separately.
+fn take_name_value_attr(attrs: &mut Vec<Attribute>, name: &str) -> Option<Lit> {
+    if let Some((i, lit)) = attrs
+        .iter()
+        .enumerate()
+        .filter(|(_, a)| a.meta.path().is_ident("boa"))
+        .filter_map(|(i, a)| a.meta.require_list().ok().map(|nv| (i, nv)))
+        .filter_map(|(i, a)| {
+            syn::parse2::<MetaNameValue>(a.tokens.to_token_stream())
+                .ok()
+                .map(|nv| (i, nv))
+        })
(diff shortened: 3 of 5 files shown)


core/engine/src/value/mod.rs
```diff
@@ -169,6 +169,17 @@ impl JsValue {
         self.as_object().and_then(|o| o.downcast_ref::<T>())
     }
 
+    /// Returns a downcasted mut ref object if the type matches. This is a shorthand
+    /// for `value.as_object().and_then(|o| o.downcast_ref<T>())`, which at time
+    /// can be contriving.
```
HalidOdat:
```suggestion
    /// Returns a downcasted mut ref object if the type matches.
    ///
    /// This is a shorthand for `value.as_object().and_then(|o| o.downcast_ref<T>())`,
    /// which at time can be contriving.
```

core/engine/src/value/mod.rs
```diff
@@ -169,6 +169,17 @@ impl JsValue {
...
+    #[inline]
+    #[must_use]
+    pub fn as_downcast_mut<T: NativeObject>(
```
HalidOdat, line 178: Suggestion: I think we should that it's only for objects.

```suggestion
    pub fn as_object_downcast_mut<T: NativeObject>(
```

core/engine/src/value/mod.rs
```diff
@@ -169,6 +169,17 @@ impl JsValue {
...
+        &self,
+    ) -> Option<boa_engine::object::RefMut<'_, ErasedObject, T>> {
+        self.as_object().and_then(|o| o.downcast_mut::<T>())
```
HalidOdat: Nitpick: Maybe we can just eliminiate the lambda entirely.
```suggestion
        self.as_object().and_then(JsObject::downcast_mut::<T>)
```

core/macros/src/lib.rs
```diff
@@ -31,6 +33,17 @@ pub fn embed_module_inner(input: TokenStream) -> TokenStream {
     embedded_module_loader::embed_module_impl(input)
 }
 
+/// Implementation of the inner iterator of the `embed_module!` macro. All
+/// arguments are required.
+///
+/// # Warning
+/// This should not be used directly as is, and instead should be used through
+/// the `embed_module!` macro in `boa_interop` for convenience.
```
HalidOdat: This doc doesn't seem correct, it looks like it's for `embed_module!` macro

core/interop/tests/class.rs
```diff
@@ -0,0 +1,101 @@
...
+            "cat" => AnimalType::Cat,
+            "dog" => AnimalType::Dog,
+            _ => AnimalType::Other,
+        };
+
+        Self { ty, age }
+    }
+
+    #[boa(name = "staticMethod")]
+    fn static_method() -> i32 {
```
HalidOdat: I think it should rename it to camelCase by default since that is JavaScript's convention and `rename` attribute should be used to override that behaviour.

Something like this should do I don't think we need to import a crate for it, especially since this is macro, compile times are important :smile: 

```rust
fn snake_to_camel(snake: &str) -> String {
    let mut camel = String::new();
    let mut upper_next = false;

    for (i, c) in snake.chars().enumerate() {
        if c == '_' {
            if upper_next {
               camel.push(c);
            }
            upper_next = true;
        } else if upper_next {
            camel.push(c.to_ascii_uppercase());
            upper_next = false;
        } else if i == 0 {
            camel.push(c.to_ascii_lowercase());
        } else {
            camel.push(c);
        }
    }

    camel
}
```

core/macros/src/class.rs
```diff
@@ -0,0 +1,703 @@
...
+            match arg {
+                Meta::NameValue(MetaNameValue {
+                    path,
+                    value: Expr::Lit(lit),
+                    ..
+                }) if path.is_ident("name") => {
+                    name = Some(match &lit.lit {
+                        Lit::Str(s) => Ok(s.value()),
+                        _ => Err(syn::Error::new(lit.span(), "Expected a string literal")),
+                    }?);
```
HalidOdat, line 768: This doesn't handle the case when name is defined twice, right?

core/macros/src/class.rs
```diff
@@ -0,0 +1,703 @@
...
+    }
+
+    fn setter(&mut self, fn_: &mut ImplItemFn) {
+        let name = match Self::name_of(fn_) {
+            Ok(name) => name,
+            Err((span, msg)) => {
+                self.error(span, msg);
+                return;
+            }
+        };
```
HalidOdat: Perhaps this should be moved into `name_of` to avoid the repetition

core/interop/tests/class.rs
HalidOdat, line 1: Maybe we should add tests for the `length` attribute with assertions for it.

core/macros/src/class.rs
```diff
@@ -0,0 +1,703 @@
...
+/// Take the last `#[boa(error = "...")]` statement if found, remove it from the list
+/// of attributes, and return the literal string.
+fn take_error_from_attrs(attrs: &mut Vec<Attribute>) -> SpannedResult<Option<String>> {
+    match take_name_value_attr(attrs, "error") {
+        None => Ok(None),
+        Some(lit) => match lit {
+            Lit::Str(s) => Ok(Some(s.value())),
+            l => Err((
+                l.span(),
+                "Invalid literal type. Was expecting a number".to_string(),
```
HalidOdat: This message seems wrong, it should be about strings, right?

core/macros/src/class.rs
```diff
@@ -0,0 +1,703 @@
...
+    fn method(name: String, fn_: &mut ImplItemFn, class_ty: &Type) -> SpannedResult<Self> {
+        if fn_.sig.asyncness.is_some() {
+            error(&fn_.sig.asyncness, "Async methods are not supported.")?;
+        }
+
+        if !fn_.sig.generics.params.is_empty() {
+            error(&fn_.sig.generics, "Generic methods are not supported.")?;
+        }
+
+        let mut has_receiver = 0;
```
HalidOdat, line 210: is it possible to have more than one receiver, if not this should probably be `bool`,  if it is we should error.

core/macros/src/class.rs
```diff
@@ -0,0 +1,703 @@
...
+                }
+                FnArg::Typed(ty) => Self::arg_from_pat_type(ty, i),
+            })
+            .collect::<SpannedResult<_>>()?;
+
+        let length =
+            take_length_from_attrs(&mut fn_.attrs)?.unwrap_or(args_decl.len() - has_receiver);
+        let fn_name = &fn_.sig.ident;
+
+        let is_static = take_path_attr(&mut fn_.attrs, "method") || has_receiver > 0;
```
HalidOdat: Isn't this suppose to be method, if it has a method attribute or it has self it should be a method, right?

HalidOdat (commented):
Nice work! Really love the new macro! :smile: 

Have some comments, see if they make any sense :)

core/engine/src/value/mod.rs
hansl: There's already an `as_downcast_ref`, we should keep the name consistent. Adding `object_` is starting to make for a long name.

core/macros/src/lib.rs
hansl: No idea how that happened, let me rewrite.

core/macros/src/class.rs
hansl: There are other failsafes for this in basic cases, but no it doesn't prevent you from going out of your way. This would be similar to setting the prototype's key `name` twice, which is also allowed.

core/interop/tests/class.rs
hansl: I was thinking about it too. I was using `serde` as an example where you need `rename` to do automatic renaming, otherwise by default it keeps snake_case. I can put a rename top-level attribute for this, WDYT?

core/macros/src/class.rs
hansl: Issue is the `return` there. I can refactor to have `Result<>` all the way.

core/macros/src/class.rs
hansl: Yeah this was renamed to the inverse of what it was. Fixed.

core/interop/tests/class.rs
hansl: I'll add more tests. Please note they're not part of the coverage report.

core/interop/tests/class.rs
hansl: Done the `rename = "camelCase"`

hansl:
@HalidOdat please take another look.

core/engine/src/value/mod.rs
HalidOdat: Your right! the we should keep it this way :smile:

core/interop/tests/class.rs
```diff
@@ -0,0 +1,112 @@
...
+        obj
+    }
+
+    #[boa(getter)]
+    fn age(&self) -> i32 {
+        self.age
+    }
+
+    #[boa(setter)]
+    #[boa(method)]
```
HalidOdat, line 60: question: does this need to marked as method, it has `&self` so should it be enough to infer it?

core/macros/src/class.rs
```diff
@@ -0,0 +1,824 @@
...
+            .as_ref()
+            .map_or_else(|| self.setter.as_ref().map(|s| &s.name), |g| Some(&g.name))
+        else {
+            return quote! {};
+        };
+        let getter = if let Some(getter) = self.getter.as_ref() {
+            let body = getter.body.clone();
+            quote! {
+                Some(
+                    boa_engine::NativeFunction::from_copy_closure( #body )
```
HalidOdat: From what I can tell (after replacing it locally running the tests), we can use `NativeFunction::from_fn_ptr()` instead of closure, which is faster and smaller in size. (should replace all instances)

core/macros/src/class.rs
```diff
@@ -0,0 +1,824 @@
...
+                if let Type::Path(path) = ty.as_ref() {
+                    let Some(t) = path.path.segments.last() else {
+                        return error(&fn_.sig.output, "Cannot infer return type.");
+                    };
+                    if t.ident == "Self" {
+                        quote! { Ok(result) }
+                    } else if t.ident == "JsResult" {
+                        quote! { result.into() }
+                    } else {
+                        return error(&fn_.sig.output, "Invalid return type.");
```
HalidOdat: Remark: Maybe we could return a more rusty error, noting/explaining what types are acceptable.

core/macros/src/class.rs
```diff
@@ -0,0 +1,824 @@
...
+
+    fn set_setter(
+        &mut self,
+        name: String,
+        fn_: &mut ImplItemFn,
+        class_ty: &Type,
+    ) -> SpannedResult<()> {
+        let setter = Function::setter(name, fn_, class_ty)?;
+        if self.setter.is_some() {
+            error(fn_, "Setter for this property already declared.")
```
HalidOdat: Nitpick: maybe the errors should print the property name that already exists.

core/macros/src/class.rs
```diff
@@ -0,0 +1,824 @@
...
+impl FromStr for RenameScheme {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.eq_ignore_ascii_case("none") {
+            Ok(Self::None)
+        } else if s.eq_ignore_ascii_case("camelcase") {
+            Ok(Self::CamelCase)
+        } else {
+            Err(format!("Invalid rename scheme: {s:?}"))
```
HalidOdat: Remark: Since they get turned into compile error messages, maybe also print the names that are acceptable.

HalidOdat (approved):
Amazing work! @hansl

Besides some nitpicks, looks good to me! :)

core/interop/tests/class.rs
hansl: Yes, because it's marked as `setter`. It's a `method` by default if 2 conditions are met:

1. &self / &mut Self, and
2. it isn't marked `constructor`, `getter`, `setter` or `static`.

core/macros/src/class.rs
hansl: Let me check, that might be nice.

hansl:
I'll need to re-run with my project and check that everything checks out (I use it for much more complicated classes than there is in the test). Once I'm sure the macro is reliable I'll merge. Thanks!

core/macros/src/class.rs
HalidOdat: Just to note, all instances of `from_copy_closure` should be replaced, https://github.com/boa-dev/boa/pull/4271/commits/3470cfc347ba6d170204cc8c7c943e806546ee30 seems to only covered 1/4 cases :smile:

hansl: Uh, you're right. Dunno why I just replaced the one. Done.

hansl:
All my bugs went away once I moved to use the latest SHA in this branch, so I'll merge as is. We can iterate.

PR #4273, opened 2025-06-06
Add `Span` to `Literal` AST node

### LINKED ISSUE #300: Span nodes

HalidOdat:
This PR partially resolves #300 by adding `Span`s to expressions.

Add spans to the following experssions:

- [ ] This
- [ ] Identifier
- [x] Literal
- [ ] RegExpLiteral
- [ ] ArrayLiteral
- [ ] ObjectLiteral
- [ ] Spread
- [ ] FunctionExpression
- [ ] ArrowFunction
- [ ] AsyncArrowFunction
- [ ] GeneratorExpression
- [ ] AsyncFunctionExpression
- [ ] AsyncGeneratorExpression
- [ ] ClassExpression
- [ ] TemplateLiteral
- [ ] PropertyAccess
- [ ] New
- [ ] Call
- [ ] SuperCall
- [ ] ImportCall
- [ ] Optional
- [ ] TaggedTemplate
- [ ] NewTarget
- [ ] ImportMeta
- [ ] Assign
- [ ] Unary
- [ ] Update
- [ ] Binary
- [ ] BinaryInPrivate
- [ ] Conditional
- [ ] Await
- [ ] Yield
- [ ] Parenthesized
- [ ] FormalParameterList?
- [ ] Debugger


core/ast/src/expression/mod.rs
```diff
@@ -277,6 +281,53 @@ impl Expression {
         }
         expression
     }
+
+    /// Get [`Span`] of the [`Expression`].
+    #[inline]
+    #[must_use]
+    // TODO: Remove lint allows after implemenation is complete.
```
nekevss: question: lint still needed?

core/ast/src/expression/literal/mod.rs
```diff
@@ -16,12 +16,121 @@ use core::ops::ControlFlow;
...
+///  - [ECMAScript reference][spec]
+///  - [MDN documentation][mdn]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-primary-expression-literals
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Literals
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[derive(Debug, Clone, PartialEq)]
+pub struct Literal {
+    kind: LiteralKind,
```
nekevss, line 42: nit (non-blocking): Wouldn't `InnerLiteral` or `LiteralRepr` be a bit better here?

Typically when I think of a `Kind`, its a lot more of a C-like enum. In this case, `LiteralKind` is the container for the literal, because the `Literal` is now the value and the span.

core/ast/src/expression/identifier.rs
```diff
@@ -38,16 +38,12 @@ pub const RESERVED_IDENTIFIERS_STRICT: [Sym; 9] = [
...
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Glossary/Identifier
-#[cfg_attr(
-    feature = "serde",
-    derive(serde::Serialize, serde::Deserialize),
-    serde(transparent)
-)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-#[repr(transparent)]
```
nekevss, line 48: question: we aren't depending on the `#[repr(transparent)]` elsewhere, correct?

We definitely need it removed in this instance, but it does sort of make me hesitate...

core/ast/src/expression/import_meta.rs
```diff
@@ -0,0 +1,64 @@
+//! Local identifier Expression.
```
nekevss: issue: file header needs to be updated.

core/ast/src/expression/new_target.rs
```diff
@@ -0,0 +1,64 @@
+//! Local identifier Expression.
```
nekevss: issue: file header needs to be updated.

core/ast/src/expression/new_target.rs
```diff
@@ -0,0 +1,64 @@
...
+use boa_interner::{Interner, ToInternedString};
+use core::ops::ControlFlow;
+
+use super::Expression;
+
+/// ECMAScript's `NewTarget` expression AST node.
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct NewTarget {
```
nekevss, line 16: thought: What should the repr here be?

This is more of an observation based off another review comment. This is essentially a new type that could have a `repr(transparent)`. I'm not saying we should add one here; in fact, I think it's worth holding off. But we may want to have a good way to designate when a type should have a certain repr and how do we then document that repr in the long term.

I don't think you should add it in this case, especially with changes that may happen in follow ups, but it may be something we should consider.

core/ast/src/expression/this.rs
```diff
@@ -0,0 +1,64 @@
+//! Local identifier Expression.
```
nekevss: issue: file header update needed.

core/ast/src/operations/mod.rs
```diff
@@ -451,21 +455,27 @@ impl IdentList for Vec<Sym> {
     }
 }
 
-impl IdentList for Vec<Identifier> {
-    fn add(&mut self, value: Sym, _function: bool) {
-        self.push(Identifier::new(value));
-    }
-}
+// impl IdentList for Vec<Identifier> {
```
nekevss: question: commented out code here intentional?

core/parser/src/parser/expression/left_hand_side/optional/tests.rs
```diff
@@ -18,16 +18,20 @@ fn simple() {
...
-                        field: PropertyAccessField::Const(
+                        field: Identifier::new(
                             interner.get_or_intern_static("name", utf16!("name")),
-                        ),
+                            Span::new((1, 4), (1, 8)),
+                        )
+                        .into(),
                     },
                     true,
+                    Span::new((1, 4), (1, 8)),
```
nekevss: thought: shouldn't this be (1, 2), (1, 3)?

Also, I'm a bit curious about whether this is inclusive or exclusive. It looks like it's exclusive. Is that documented anywhere? I might have missed it.

core/parser/src/parser/expression/primary/function_expression/tests.rs
```diff
@@ -113,27 +141,31 @@ fn check_nested_function_expression() {
 #[test]
 fn check_function_non_reserved_keyword() {
     macro_rules! genast {
-        ($keyword:literal, $interner:expr) => {
+        ($keyword:literal, $interner:expr, $function_span:expr, $name_span:expr, $body_span:expr, $literal_span:expr) => {
```
nekevss: thought: it might be nice to add some clarity names to the macro expansion to aid with understanding.

For instance:

```rust
    let ast = genast!(
        "as",
        interner,
        function: Span::new((1, 13), (1, 40)),
        name: Span::new((1, 22), (1, 24)),
        body: Span::new((1, 27), (1, 40)),
        literal: Span::new((1, 36), (1, 37))
    );
```

Although, maybe that becomes overly verbose?

core/ast/src/expression/mod.rs
HalidOdat: This seems to be an older commit, it's fixed in alter commits: https://github.com/boa-dev/boa/pull/4273/files#diff-f055b46e7262e69d18478e7b465d042d3800db2bae179b11452f0ca75a63e288R291-R295 :smile:

core/ast/src/expression/literal/mod.rs
HalidOdat: > nit (non-blocking): Wouldn't `InnerLiteral` or `LiteralRepr` be a bit better here?

Not sure... I was just trying to be consistent with what we do in the `Token` and `TokenKind` (which is a fat enum).

Let me know if I should change the inner literal name :smile:

core/ast/src/expression/identifier.rs
HalidOdat: > question: we aren't depending on the `#[repr(transparent)]` elsewhere, correct?

I don't think so as far as I can tell.. things like transmutes would give an error due to size difference.

core/ast/src/operations/mod.rs
HalidOdat: Opps, forgot to remove that commented out code... will remove it :smile:

core/ast/src/expression/literal/mod.rs
nekevss: Huh, that's a good point. I still think that it would make more sense for this to be changed to an inner literal vs. a kind. I think it makes more sense in this instance to me.

EDIT: to expand, these almost feel semantically different then what it currently is. The enum in this case is wrapping the actual literal so it can be packaged with metadata vs. strictly information. I think this probably extends to `Token` too after taking a look at it.

core/parser/src/parser/expression/left_hand_side/optional/tests.rs
HalidOdat: > Is that documented anywhere?

Yes, the doc on `Span` it's `[start, end)`: https://github.com/boa-dev/boa/pull/4273/files#diff-f055b46e7262e69d18478e7b465d042d3800db2bae179b11452f0ca75a63e288R291-R295

> thought: shouldn't this be (1, 2), (1, 3)?

Hmmm.. Since this span is for an [`OptionalOperationKind::SimplePropertyAccess`](https://github.com/boa-dev/boa/pull/4273/files/94c6173a23244e268393561d22d540b646fe89ef#diff-898be9165c43242494a5bc9d8cb53ecfa78075689d41470b6d40e1dc155c6144L16-L19) I included the operator and the field it accessed (similar to regular fields access).

EDIT: :man_facepalming: , sorry I misunderstood `(1, 2), (1, 3)` as a span... you are right! Nice catch! Will fix it! :sweat_smile: 

EDIT: Fixed in https://github.com/boa-dev/boa/pull/4273/commits/3d29391ce08a31ecb892b2a01b8fa35717fcf579

core/parser/src/parser/expression/primary/function_expression/tests.rs
HalidOdat: Done in https://github.com/boa-dev/boa/pull/4273/commits/8b439bcd0a1d2b046ba84496bf5b0f13a479cfbd :smile:

core/ast/src/expression/new_target.rs
HalidOdat: I agree that we should hold off on adding them for now -- especially since it’s hard to predict all future use cases, and we might need to include additional information later (e.g. for static analysis, similar to what we did with functions).

HalidOdat:
> Out of pure curiosity, did you happen to run any benchmarks on this branch to see what the impact of adding spans is (if there even is any)?

Ran the local parser benchmarks (criterion) and there was a performance improvement on almost all the benchmark `~10%` :partying_face:.

I think this is due to wrapping some of the very big enum variants in `Box` (done to appease clippy), which overall shrunk the memory usage, see commit: https://github.com/boa-dev/boa/pull/4273/commits/94c6173a23244e268393561d22d540b646fe89ef

NOTE: The quickjs/v8 benchmarks only benchmark runtime performance so they aren't relevant.

core/ast/src/expression/literal/mod.rs
HalidOdat: > Not sure... I was just trying to be consistent with what we do in the `Token` and `TokenKind` (which is a fat enum).

Actually after looking around it's done through out the code base of `boa_ast`, so maybe that should be done in a separate PR...? (maybe a good first issue?)

nekevss (approved):
LGTM :)

PR #4274, opened 2025-06-06
Bump ICU4X to 2.0

jedel1043:
Finally...

This PR bumps all our ICU4X dependencies to version 2.0.
This is just a version bump, so some APIs can be further improved by using the many new features introduced by ICU4X 2.0, but that should be its own PR for ease of review.

Note that this is a breaking change, since ICU4X removed the concept of `AnyProvider`s, so the only type of providers allowed at the moment are `BufferProvider`s. A bit of design work needs to be done if we want to allow providing `DataProvider`s on the API.


jedel1043:
Hour cycle related regressions are caused by a recent deprecation (and removal) of the `hc-h24` value.
Track https://github.com/unicode-org/icu4x/issues/6597 and https://github.com/tc39/ecma402/issues/1002 for a resolution on that.

jedel1043:
PR is good to go now

core/engine/src/builtins/intl/date_time_format.rs
```diff
@@ -295,3 +283,30 @@ pub(crate) fn to_date_time_options(
...
+            .and_then(|v| CalendarAlgorithm::try_from(&v).ok())
+            .ok_or_else(|| {
+                JsNativeError::range()
+                    .with_message(format!("provided calendar `{s}` is invalid"))
+                    .into()
+            })
+    }
+}
+
+impl OptionType for HourCycle {
```
nekevss, line 304: thought: potentially add a TODO comment for tracking the ongoing `HourCycle` discussion

tools/gen-icu4x-data/Cargo.toml
```diff
@@ -10,14 +10,15 @@ license.workspace = true
...
-    "networking",
-    "use_wasm",
-    "provider",
+icu_provider_export = { workspace = true, features = [
     "blob_exporter",
-    "experimental_components",
     "rayon",
 ] }
+icu_provider_source = { workspace = true, features = [
+    "networking",
```
nekevss, line 18: question: what data do we need networking for?

core/icu_provider/src/lib.rs
```diff
@@ -82,34 +103,64 @@ macro_rules! provider_from_icu_crate {
...
+}
+
+/// Gets the default data provider stored as a [`DynamicDryDataProvider<BufferMarker>`].
 ///
-/// [`BufferProvider`]: icu_provider::BufferProvider
+/// [`DynamicDryDataProvider<BufferMarker>`]: icu_provider::DynamicDryDataProvider
 #[must_use]
-pub fn buffer() -> &'static impl BufferProvider {
-    &*PROVIDER
+pub fn buffer() -> impl DynamicDryDataProvider<BufferMarker> {
```
nekevss, line 164: question: is leaving `Wrapper` private intentional here?

At first this was a little surprising, but on a second glance I believe I understood it better. Although, I do think it's a bit surprising returning something private from a public function (although, it's an interesting way of sealing the general usage, which is what I'm assuming the intent is here). Nevertheless, thought that I'd confirm.

core/engine/src/builtins/intl/date_time_format.rs
```diff
@@ -295,3 +283,30 @@ pub(crate) fn to_date_time_options(
     // 13. Return options.
     Ok(options)
 }
+
+impl OptionType for CalendarAlgorithm {
+    fn from_value(value: JsValue, context: &mut Context) -> JsResult<Self> {
+        let s = value.to_string(context)?.to_std_string_escaped();
+        Value::try_from_str(&s)
```
nekevss, line 290: thought: is it possible to use a `try_from_utf8` and `try_from_utf16`.

I don't think the constructors are available on `Value` itself, but this did come up with `temporal_rs`, where a `JsString` type can be either utf8 or utf16. We may be able to avoid the allocation of `to_std_string_escaped` if we use `JsStringVariant` instead.

tools/gen-icu4x-data/Cargo.toml
jedel1043: Fetching the CLDR data repo. Otherwise we would have to download and provide it manually.

core/icu_provider/src/lib.rs
jedel1043: Yes. This allows changing our data provider without having to change the signature of the function.

core/engine/src/builtins/intl/date_time_format.rs
jedel1043: Problem is that our strings are technically latin1, so we cannot use `try_from_utf8`, and `try_from_utf16` doesn't exist for `Value` IIRC

HalidOdat (approved):
Looks good to me! :smile:

core/engine/src/builtins/intl/date_time_format.rs
nekevss: Yeah, we'd probably have to file an issue / request for a `try_from_utf16` for `Value`. Do we actually go over the first 128 characters in our latin1 repr?

nekevss: Nevermind, just looked. Looks like we are doing latin1

jedel1043: Yeah, and I think it's hard to justify adding a `try_from_latin1` on ICU4X since locales are always ascii, but we can leave it as a future improvement for now.

nekevss: Yeah, there's no justification for such an API, especially since, that seems like something we'd need to be concerned about.

We could also just treat it as GIGO since locals are ASCII, which should align with Latin1

PR #4277, opened 2025-06-09
Add a `#[boa_module]` macro to automatically implement a Module

hansl:
This adds a `pub(super) fn boa_module(context: &mut Context) -> JsResult<Module>` function to the same module that builds the module from its content.

It understands the following:

1. `fn some_func(...) -> ... {}` will be added as functions to the module.
2. `const SOME_CONST: ... = ...` will be added as values. The type of the constant value must support `impl Into<JsValue>`.
3. `type SomeType = ...` will be added as a class. The class will be exported using the `SomeType` name (with renaming support).

Two schemes of renaming are now supported, one for classes and one for regular imports. Use `#[boa(rename_class = "...")]` to select the one you want.

A PascalCase renaming is also added to allow for classes.

Simple tests added for everything.

### DIFF
--- a/core/macros/src/utils.rs
+++ b/core/macros/src/utils.rs
@@ -119,10 +119,6 @@ impl FromStr for RenameScheme {
 }
 
 impl RenameScheme {
-    pub(crate) fn from_attrs(attrs: &mut Vec<Attribute>) -> SpannedResult<Self> {
-        Self::from_named_attrs(attrs, "rename").map(|rs| rs.unwrap_or(RenameScheme::None))
-    }
-
     pub(crate) fn from_named_attrs(
         attrs: &mut Vec<Attribute>,
         name: &str,


jedel1043:
Hmmm, I think it's REALLY surprising to have common Rust syntax be what determines what gets exported in the module...

Can we add an `#[export]` attribute instead? IMO it should be a lot more readable and less surprising overall to be a bit more explicit in this case.

hansl:
@jedel1043 

> Can we add an #[export] attribute instead? IMO it should be a lot more readable and less surprising overall to be a bit more explicit in this case.

Could you give me an example of what that would look like? You can use the current example test module as template (export a const, function and class). Keep in mind the idea is to have a function that returns the module, so we would have to add that to the source as well.

I felt that having a `mod` was easy and explicit enough with the proc_macro above. The name of the module is irrelevant and it doesn't have to be public, so this cannot interfere with "valid" code.

jedel1043:
I stand corrected. I was thinking it would be a bit clunky to have something like a private `const` used throughout multiple public methods, but I see that the correct usage would be to have private functions, types and constants in the parent module, then the child module is the one with the `boa_module` attribute, which is nicer than having to add an `#[export]` tag to every single exported item.

core/interop/tests/module.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[derive(Clone, Trace, Finalize, JsData)]
+struct Animal {
+    ty: AnimalType,
+    age: i32,
+}
+
+#[boa_class]
+#[boa(rename = "camelCase")]
```
jedel1043: I don't think these should be independent attributes. It doesn't make sense to have `#[boa(rename = "camelCase")]` without a corresponding `#[boa_class]` or `#[boa_module]`. I would make it so that you're required to provide the rename schema as an option for the `boa_class` or `boa_module` attributes:

`#[boa_class(rename = "camelCase")]`

core/interop/tests/module.rs
```diff
@@ -0,0 +1,112 @@
...
+            AnimalType::Cat => js_string!("meow"),
+            AnimalType::Dog => js_string!("woof"),
+            AnimalType::Other => js_string!(r"¯\_(ツ)_/¯"),
+        }
+    }
+}
+
+#[boa_module]
+#[boa(rename = "camelCase")]
+#[boa(rename_class = "PascalCase")]
```
jedel1043: Is it really necessary to always specify the naming schema? IMO it makes much more sense to automatically rename to the "correct" schema, then just provide a generic `#[boa(js_name = "newName")]` if the user wants to override the default name. Do you think it is too surprising to automatically change the name of exports and class members?

jedel1043 (commented):
Really impressive work! Couple of discussion points that I wanted to raise.

core/interop/tests/module.rs
hansl: `#[boa(name = "...")]` is available on all members, so this could definitely be done. I was thinking about just making `camelCase` the default for const and functions, and `PascalCase` the default for classes. WDYT?

jedel1043: Yes, that sounds nice!

hansl:
@jedel1043 Done, changed the defaults to be more sensible.

Issue #4275, opened 2025-06-07
Panic with Pointer is not 4-bits aligned or over 51-bits


andreievg:
**Describe the bug**
Getting this panic on aarch64-linux-android build running on Android device:

2025-06-06 09:19:02.933955487 [ERROR] <log_panics:130>:thread '<unnamed>' panicked at 'assertion `left == right` failed: Pointer is not 4-bits aligned or over 51-bits.
  left: 502975824280
 right: 12970367429802852760': /.cargo/git/checkouts/boa-126d820dff07bf57/6f03c21/core/engine/src/value/inner/nan_boxed.rs:379
  

**To Reproduce**

[I've create this repo](https://github.com/andreievg/boa-alignment-test), with replication instructions, test pass on v0.20, but fail on main, this is only testeable with, arm64-v8a. I can't seem to get github action running emulator for arm64-v8a, tried a few things, but none seem to work. But when i connect samsung tablet and run the test described in readme.md, it can be replicated. 

**Expected behavior**
Explain what you expected to happen, and what is happening instead.

**Build environment (please complete the following information):**

- OS: Samsung tablet, Android
- Version: On "main", 0.20 works
- Target triple: aarch64-linux-android
- Rustc version: 1.86.0

**Additional context**

@raskad, here is the issue, sorry couldn't get replication going in the github action

hansl:
Hi @andreievg. Does Android tag pointers? Some OSes do, and that would not work with our setup (since we tag pointers ourselves too).

jedel1043:
@hansl According to [Android's documentation](https://source.android.com/docs/security/test/tagged-pointers#memory-tagging-extension-readiness), yes:

> ARM's Memory Tagging Extension (MTE) helps address memory safety issues. MTE works by tagging the 56th-59th address bits of each memory allocation on the stack, heap, and globals. The hardware and instruction set automatically checks that the correct tag is used upon every memory access.

jedel1043:
IMO Android is too big of a platform to not support it, so we should offer an alternative implementation of nan-boxing only for that architecture.

andreievg:
> IMO Android is too big of a platform to not support it, so we should offer an alternative implementation of nan-boxing only for that architecture.

That would be great, i think for now we will revert to 0.20, or use main with revert to this https://github.com/boa-dev/boa/pull/4251 (since i still need the 'safe' version of to_json, for serde). Btw from our testing, 0.20 seemed to be working well for Android.

[...]

PR #4281, opened 2025-06-10
Add the legacy enum-based JsValue implementation behind a flag

hansl:
For platforms that have pointer tagging (either on MSB or LSB). This should allows us to support those platforms until a more permanent solution is sorted out.

Ref #4275.


core/engine/Cargo.toml
```diff
@@ -12,6 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [features]
+legacy-jsvalue = []
```
HalidOdat: Discussion: I think we should make this the default, and `nan-boxing` should be a feature that is enabled as a `default-feature` this way it makes it easier to add any other value encoding, in the future if we need too.

raskad: Agree. I would also name this `jsvalue-encoding-enum`, just to make it more explicit and allow for additional encodings with the same naming convention.

hansl: I don't mind but I'll ask @jasonwilliams to weigh in on this. AFAICT NaN-Boxing was always the direction we wanted. This is just to support this one platform (Android 16+), and we have a workaround for it on the app side. I would advise against having to maintain two vastly different implementations of `JsValue`, if only because we lack the bandwidth as a team.

If NaN-Boxing is the future implementation (which I think it is if only because it should be more efficient when JsValue is `Copy`), then that should be the default.

HalidOdat: I was thinking the enum version should be the default (and doesn't need a feature flag) since it's universally compatible, while `nan-boxing` could be an optional feature, as it doesn't work everywhere.

raskad: Yeah, that would also be an option. My only question is then, if we run e.g. our benchmarks with the default universally compatible features, or with those that we focus more on when going for optimizations. I would of course be for doing the later.

HalidOdat: Sorry for late response... we can make the `nan-boxing` a default feature, that way we get the performance benefit.

hansl: So you want to have a new feature `nan-boxing` and add that to the default flags, so whichever platforms want/need to use the enum based impl need to pass `no-default-features` as well as list the other default features?

HalidOdat: Yes. it's not ideal for those platforms...

AFAIK, conditionally enabling features based on arch (for the current crate) is not supported, but for dependent crates this works:

```toml
[target.'cfg(target_arch = "x86_64")'.dependencies]
boa_js_value = { version = "0.1.0", features = ["nan-boxing"] }

[target.'cfg(not(target_arch = "x86_64"))'.dependencies]
boa_js_value = { version = "0.1.0" }
```

And for architectures, that we are not sure if it will work (like the before mentioned android platform) we could omit them and let the user decide.

Taking `JsValue` out of boa would be a big refactor, for now we can just make it a default on all cases.

HalidOdat: Discussion:

Though there is a case to be made about having a reversed (`enum-jsvalue`), if we have the above implementation with nan-boxing as default. If a crate enables `nan-boxing` (because it's default) and it's consumed by a crate that compiles for a target that can't work with `nan-boxing`. Due to feature unification `nan-boxing` will be enabled, even if the consuming library has the `no-default-features`... (not an issue if `nan-boxing` is not default, or if the platform supports it always).

But I think this is a case aginst `nan-boxing` being the default... it's the performant path, but could have issues on some architectures.

I think we should treat at as a feature that could potentially be unsafe, for example if we add a feature that removes all asserts for the bytecode, making them only debug assets (such as for reading operands accessing registers, etc this will definitely increase performance, but..) will requires unsafe, even if we have some sort of finite state machine, to verify that the it will not cause some out-of-bounds read. I would still put it in a feature fag that is not enabled by default (something like `unsafe-vm`, or something like that). To give the user full control and confidence that if they are not enabled then it's safe. Same for a `jit` feature (it's inherently unsafe)

When we run benchmarks/tests etc those features should be enabled this will allow us to catch any bugs.

hansl: > But I think this is a case aginst nan-boxing being the default... it's the performant path, but could have issues on some architectures.

I think this is a case for `jsvalue-enum` being the feature, and the default being no crate feature (being nan boxing). Basically use nan boxing, unless you or a transitive crate enable enum jsvalues because the target platform doesn't support it. The happy path is performant, the fallback is compatibility. Not the other way around.

This is how crate features are meant to be used, additive-ly.

Also, we haven't discussed how many platforms are using MTE. Android is one, but I'm failing to find another one at the moment. And they use only the 56-59th bits, which means we might be able to work around that for our use case (see https://source.android.com/docs/security/test/tagged-pointers).

HalidOdat: Fair enough :smile: Let's merge this we can always change it in the future (if we need to :) )

Suggestion: Rename the feature to `enum-jsvalue` (instead of legacy which doesn't provide much information.

Suggestion: Add some documentation on the feature so user are aware of what it is and why they would want to activate it.

hansl: Changing it to `jsvalue-enum` (larger > smaller semantic scope).

PR #4282, opened 2025-06-13
Add a js_value! macro to allow creation of JsValue from JSON-like DSL

hansl:
This is as close to JSON as we can get.

This should make `impl Into<JsValue> for SomeRustStruct {}` really easy to make.

This macro can be used like so:

```rust
js_value!(
  {
    "some": "value",
    "array": [1, 2, { "this": "works too" }, [[["nested"], [] /* empty */]]],
    // Allows Rust comments!
    "bool": [true, false],
    "etc": {}
  }
)
```

The main limitation would be that if an error happens during serialization, it will panic. And it doesn't support `TryIntoJs` trait, only `Into<JsValue>`. The follow up will be to have a `js_value_ok!(...)` which would use `TryIntoJs` and return a `JsResult<JsValue>` instead of a `JsValue`.


core/engine/src/object/jsobject.rs
```diff
@@ -287,6 +287,42 @@ impl JsObject {
...
+    /// Checks that all own property keys and values are equal (recursively).
+    #[must_use]
+    #[inline]
+    pub fn deep_strict_equals(lhs: &Self, rhs: &Self, context: &mut Context) -> bool {
+        let Ok(l_keys) = lhs.own_property_keys(context) else {
+            return false;
+        };
+        let Ok(r_keys) = lhs.own_property_keys(context) else {
+            return false;
+        };
```
HalidOdat: I think it's best to return a `JsResult<bool>` and propagate the error, since calling the `own_property_keys` could have side effects.

core/engine/src/value/macros.rs
```diff
@@ -0,0 +1,49 @@
+//! Macros related to `JsValue`.
+
+/// Create a `JsValue` from a simple DSL that resembles JSON.
+///
+///
```
HalidOdat: Remark: Maybe include some of the tests as doc, showcasing the macro :)

HalidOdat (commented):
Nice work! Besides the small comments this seems good to me! :)

core/engine/src/object/jsobject.rs
```diff
@@ -287,6 +287,37 @@ impl JsObject {
...
+            let vr = rhs.get_property(key);
+
+            match (vl, vr) {
+                (None, None) => {}
+                (Some(vl), Some(vr)) => match (vl.value(), vr.value()) {
+                    (None, None) => {}
+                    (Some(lv), Some(rv)) => {
+                        if !lv.deep_strict_equals(rv, context)? {
+                            return Ok(false);
+                        }
```
jedel1043: I don't think this is the correct approach because a LOT of objects could have circular references to themselves, causing an infinite loop in that case. We could probably use a map of all seen objects to handle this?

core/engine/src/value/macros.rs
```diff
@@ -0,0 +1,61 @@
...
+///   //
+/// }, context);
+/// ```
+#[macro_export]
+macro_rules! js_value {
+    // Single pattern to simplify documentation.
+    ($($v:tt)+) => {
+        $crate::js_value_internal!($($v)+)
+    };
+}
```
jedel1043: Could we also have an equivalent `js_object` macro that skips wrapping the object with `JsValue`? Seems useful to have this alternative version for when someone just needs the object itself instead of a value.

core/engine/src/object/jsobject.rs
hansl: I'll just serialize the object, sigh.

jedel1043: I mean, we could also try to port https://www.npmjs.com/package/deep-equal, since it sounds useful to have

core/engine/src/value/macros.rs
jedel1043: Another idea I had was to make `js_value!` private and only expose `js_object!`, since `JsValue::from(f32::INFINITY)` is only 5 characters longer than using the macro. This would completely sidestep the problem of having to surround some special values like `f32::INFINITY` with parentheses.

hansl: Done.

core/engine/src/object/jsobject.rs
hansl: Actually implemented it like we do for display.

core/engine/src/value/equality.rs
```diff
@@ -1,7 +1,28 @@
...
+        match (self.as_object(), other.as_object()) {
+            (None, None) => Ok(self.strict_equals(other)),
+            (Some(x), Some(y)) => JsObject::deep_strict_equals_inner(x, y, encounters, context),
+            _ => Ok(false),
+        }
+    }
+
+    /// Deep equality comparison, strict. If the value is an object/array, also compare
+    /// the key-values. It uses `strict_equals()` for non-object values.
+    pub fn deep_strict_equals(&self, other: &Self, context: &mut Context) -> JsResult<bool> {
```
HalidOdat: Nitpick: Separate function title from description.

```suggestion
    /// Deep strict equality.
    ///
    /// If the value is an object/array, also compare the key-values.
    /// It uses `strict_equals()` for non-object values.
    pub fn deep_strict_equals(&self, other: &Self, context: &mut Context) -> JsResult<bool> {
```

core/macros/src/value.rs
```diff
@@ -0,0 +1,249 @@
...
+                let value = value.output(context)?;
+                let Some(c_ident) = context else {
+                    return Err(syn::Error::new(
+                        Span::call_site(),
+                        "Need to specify a context identifier.",
+                    ));
+                };
+
+                Ok(quote! {
+                    o.set( ::boa_engine::js_string!( #key ), #value, false, #c_ident )
```
HalidOdat: Consider using [`ObjectInitializer`](https://github.com/boa-dev/boa/blob/dddc98ffe279f21e7ed79cfd6d592389181c5f5e/core/engine/src/object/mod.rs#L610)

core/macros/src/value.rs
```diff
@@ -0,0 +1,249 @@
...
+        let Some(c_ident) = context else {
+            return Err(syn::Error::new(
+                Span::call_site(),
+                "Need to specify a context identifier.",
+            ));
+        };
+
+        Ok(quote! {
+            ::boa_engine::JsValue::from(
+                ::boa_engine::object::builtins::JsArray::from_iter( vec![ #(#items),* ], #c_ident )
```
HalidOdat:
```suggestion
                ::boa_engine::object::builtins::JsArray::from_iter( [ #(#items),* ], #c_ident )
```

Couldn't we use a static array here? It implements `IntoIterator`

core/macros/src/value.rs
```diff
@@ -0,0 +1,249 @@
...
+impl Object {
+    fn output(&self, context: Option<&Ident>) -> syn::Result<TokenStream> {
+        let fields: Vec<TokenStream> = self
+            .fields
+            .iter()
+            .map(|field| match &field.key {
+                Key::Ident(ident) => (ident.to_string(), &field.value),
+                Key::StringLiteral(literal) => (literal.value(), &field.value),
+            })
+            .map(|(key, value)| {
```
HalidOdat: We could use `::boa_engine::js_string!( #key )` for `StirngLiteral` and for ident we should probably use `PropertyKey::from`, so we can allow Symbols as itentifiers

core/macros/src/value.rs
hansl: So, not sure how that happened, but when I was doing the macro_rules version of this and this wouldn't work (neither was `&[]`) for some reason. Seems to work now, so yeah!

core/macros/src/value.rs
hansl: Thanks, I didn't know about it.

core/macros/src/value.rs
hansl: Good idea. Done.

core/macros/src/value.rs
hansl: Turns out this cannot be used, unfortunately. It borrows `context` and when using recursive objects (objects in objects) we'd have to borrow `context` as mutable twice. So I'll stick with the regular setting properties.

jedel1043: You can borrow the context again with [`ObjectInitializer::context`](https://github.com/boa-dev/boa/blob/dddc98ffe279f21e7ed79cfd6d592389181c5f5e/core/engine/src/object/mod.rs#L717-L720). Does that not work in this case?

core/macros/src/value.rs
hansl: There is a case to be made where we would like to align to JavaScript; `ident` becomes an `"ident"` key and `[expr]` resolves `expr` to a key. Doing this would prevent us from ever doing the second one. I'm not against it, just want to point it out.

HalidOdat: Makes sense, I like the new bracketed expressions :)

core/macros/src/value.rs
hansl: I changed it to use `ObjectInitializer`. This works.

hansl:
@HalidOdat @jedel1043 I think I addressed all your comments. PTAL

jedel1043 (approved):
Very nice!

PR #4284, opened 2025-06-16
Encode/decode testing for Argument types

### LINKED ISSUE #4240: Add unit tests for bytecode arguments reading
We should have unit test for writing / reading of bytecode arguments. Especially during argument reads we use `unsafe` so unit tests should ensure that every read can only panic.

Argument writing / reading is confined to [args.rs](https://github.com/boa-dev/boa/blob/main/core/engine/src/vm/opcode/args.rs).
Tests should cover every argument type and if present format.

t-h-p:
This Pull Request fixes/closes #4240.

It changes the following:

- Adds tests for encode/decode functionality for all argument types.
- Changes formatting for reads and writes of `(VaryingOperand,i32)`.


core/engine/src/vm/opcode/args.rs
```diff
@@ -286,48 +287,177 @@ impl Argument for (VaryingOperand, i16) {
...
+            }
+            Format::U32 => {
+                assert!(bytes.len() >= pos + 8, "buffer too small to read arguments");
+                let (arg1, arg2) = unsafe {
+                    (
+                        read_unchecked::<u32>(bytes, pos),
+                        read_unchecked::<f32>(bytes, pos + 4),
+                    )
+                };
+                ((arg1.into(), arg2), pos + 8)
```
HalidOdat: Couldn't we use `read` here?
```suggestion
                let ((arg1, arg2), pos) = read::<(u32, f32)>(bytes, pos);
                ((arg1.into(), arg2), pos)
```

core/engine/src/vm/opcode/args.rs
```diff
@@ -659,6 +789,8 @@ impl Argument for (u32, u64, VaryingOperand) {
     fn encode(self, bytes: &mut Vec<u8>) {
         write_u32(bytes, self.0);
         write_u64(bytes, self.1);
+        // Might want to add some formatting to this?
+        // Not sure what byte arrangment is good so I'll leave it alone for now.
```
HalidOdat: Add it at the beginning like the other `Argument`s :)

core/engine/src/vm/opcode/args.rs
```diff
@@ -807,3 +939,654 @@ impl Argument for (VaryingOperand, ThinVec<u32>) {
...
+    T::decode(r, pos)
+  }
+
+
+  //
+  // VariableOperand
+  //
+  #[test]
+  fn arg_vop_u8() {
+    test_write_arg(VaryingOperand{value: u8::MAX as u32});
```
HalidOdat: We can use [From<T>](https://github.com/boa-dev/boa/blob/dddc98ffe279f21e7ed79cfd6d592389181c5f5e/core/engine/src/vm/opcode/mod.rs#L235-L269) impl on `VaryingOperand`
```suggestion
    test_write_arg(VaryingOperand::from(u8::MAX));
```

core/engine/src/vm/opcode/args.rs
```diff
@@ -807,3 +939,654 @@ impl Argument for (VaryingOperand, ThinVec<u32>) {
...
+
+  fn test_write_arg<T: Argument>(test_data: T) -> Vec<u8> {
+    let mut v: Vec<u8> = Vec::new();
+    test_data.encode(&mut v);
+    return v
+  }
+
+  // This can fail depending on Argument/opcode combination given (too short)
+  fn test_read_arg<T: Argument>(opcode: Box<[u8]>, pos: usize) -> (T, usize) {
+    let r: &[u8] = &opcode;
```
HalidOdat:
```suggestion
  fn test_read_arg<T: Argument, B: AsRef<[u8]>>(bytecode: B, pos: usize) -> (T, usize) {
    let r: &[u8] = bytecode.as_ref();
```

We can add a generic type that can be turned into a reference so we can give an array directly.

```rust
    test_read_arg::<VaryingOperand>([0,0xFF], 0);
```

See: https://doc.rust-lang.org/std/primitive.array.html#impl-AsRef%3C%5BT%5D%3E-for-%5BT;+N%5D
See: https://doc.rust-lang.org/std/convert/trait.AsRef.html

core/engine/src/vm/opcode/args.rs
```diff
@@ -807,3 +939,654 @@ impl Argument for (VaryingOperand, ThinVec<u32>) {
...
+  use super::*;
+
+  fn test_write_arg<T: Argument>(test_data: T) -> Vec<u8> {
+    let mut v: Vec<u8> = Vec::new();
+    test_data.encode(&mut v);
+    return v
+  }
+
+  // This can fail depending on Argument/opcode combination given (too short)
+  fn test_read_arg<T: Argument>(opcode: Box<[u8]>, pos: usize) -> (T, usize) {
```
HalidOdat: If you always pass `0` to `pos` it's probably better to remove the pos argument.

core/engine/src/vm/opcode/args.rs
```diff
@@ -807,3 +939,654 @@ impl Argument for (VaryingOperand, ThinVec<u32>) {
...
+  #[test]
+  #[should_panic]
+  fn fail_arg_vop_u8() {
+    test_read_arg::<VaryingOperand>(Box::new([0]), 0);
+  }
+
+  #[test]
+  fn arg_vop_u16() {
+    test_write_arg(VaryingOperand{value: u16::MAX as u32});
+    test_read_arg::<VaryingOperand>(Box::new([1,0xFF,0xFF]), 0);
```
HalidOdat: We should check if the argument is equal after turning to bytes and back.
```suggestion
    let arg = VaryingOperand::from(u16::MAX);
    test_write_arg(arg);
    assert_eq(test_read_arg::<VaryingOperand>(Box::new([1,0xFF,0xFF]), 0), arg);
```

t-h-p:
I made the requested changes and formatted it, this should pass the checks.

HalidOdat:
A lot of rust tests seem to be failing, have you checked why?

t-h-p:
I made the most recent commit without running the full test suite or clippy, my mistake. I also found the changes that were causing most of the issues. `(u32,VaryingOperand)` and `(u32, VaryingOperand, VaryingOperand)` both break a lot of tests when formatting is added, instead of using write_u32 for all variants. Now I am trying to figure out the problem.

PR #4286, opened 2025-06-17
Add a #[boa(rename = ...)] attribute to TryFromJs and TryIntoJs derive macros

hansl:
This should reduce the amount of rename on each fields.

### DIFF
--- a/examples/src/bin/try_into_js_derive.rs
+++ b/examples/src/bin/try_into_js_derive.rs
@@ -4,6 +4,12 @@ use boa_engine::{
     Context, JsResult, JsValue, Source,
 };
 
+#[derive(TryFromJs, TryIntoJs, Debug, PartialEq, Eq)]
+#[boa(rename = "camelCase")]
+struct InnerTest {
+    hello_world_how_are_you: u8,
+}
+
 #[derive(TryIntoJs)]
 struct Test {
     x: i32,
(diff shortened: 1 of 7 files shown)


examples/src/bin/try_into_js_derive.rs
```diff
@@ -4,6 +4,12 @@ use boa_engine::{
     Context, JsResult, JsValue, Source,
 };
 
+#[derive(TryFromJs, TryIntoJs, Debug, PartialEq, Eq)]
+#[boa(rename = "camelCase")]
```
jedel1043: IMO it's pretty weird to have both `rename = "format"` and `rename = "newName"`. Can we be a bit more consistent on this? We could rename the first rename to something else (`name_scheme`, `name_format`, maybe something else).

hansl: `serde` has `rename_all` at the container, and `rename` or `alias` at the field level.

hansl: https://serde.rs/container-attrs.html#rename_all

I think consistency with the other proc macros is more important than the name itself, but I'd suggest just `name` and `rename_all` at the container. WDYT?

jedel1043: I like `rename_all` and `rename`. To the point and very easy to remember, plus consistent with Serde so it's one less thing for users to remember.

hansl: Done.

PR #4287, opened 2025-06-18
Refactor `JsObject` to always be size 8 

raskad:
This is currently a draft, since I think there are some important design decisions that need to be discussed.If you look at this, best is to review the changes commit by commit.

This PR contains two changes that can be best viewed separately.

1. [Refactor `JsObject` to always be a size 8 thin pointer (`Gc<T: Sized>`).](https://github.com/boa-dev/boa/commit/89c136ab2b0cbbbfdc7467a5b0f2c2c4ea72928b)
2. [Building on 1., remove the `Box` from the nan-boxed `JsObject` value.](https://github.com/boa-dev/boa/commit/8aa6d2af956a1ead9610c7e80c287d93ca9e0ba0)

## Motivation

I think the ideal representation for `JsObject` and `JsValue` should be a `no Drop` + `Copy` size 8 value. Currently we have a lot of drop and clone code in our runtime, caused by our ref-counting GC, as well as a lot of `Box` allocations from our current nan-box creation / clones. Hopefully we can get rid of the `Drop` and `Clone` requirements for `Gc` in the future. Then we could also move all other pointer types (`JsString`, `JsSymbol` and `JsBigInt`) to be `Gc` values, to enable `JsValue` to be `Copy`.

## Refactor `JsObject` to always be a size 8 thin pointer (`Gc<T: Sized>`)

Currently our `JsObject` type can be either typed (e.g. `JsObject<u64>`) or erased (e.g. `JsObject` == `JsObject<dyn NativeObject>)`. While there is only a `Gc` field on `JsObject`, the erased variant is a size 16 fat pointer, to store the Rust DST vtable for `dyn NativeObject`. The goal of this change is to make `JsObject` always a transparent size 8 thin pointer (`Gc`) to enable storing it directly as a nan-boxed value.

Here are the requirements for `JsObject` that I tried to keep:

* `JsObject` needs to be a size 8 thin pointer.
* Typed and erased variants should be one struct (erased being the default).
  * Possible alternative: `JsObject` could be a non generic struct and `JsObject::downcast` could return a typed struct. The positive of this approach would be, that access to the data T of the object can be controlled better, which would probably make everything safer. One drawback, that I currently see, is, that functions that should work on `JsObject<T>` would have to be either specific to one variant, or need to be implemented twice. I have not tried this approach, but I think in theory it should be possible.
* Any type that implements `NativeObject` should be able to be used as `T` for `JsObject` without any other user interactions (e.g. registering the type, manual unsafe vtable implementation, ...)

This lead me to the current design in this PR. Instead of erasing the object into a `dyn NativeObject`, the `TypeId` is stored in the object itself and used to check during casts. The default `T` is erased into a size 0 struct. I don't really like this approach, since one could access the erased struct. It might also be that some of this is inherently unsafe. Any ideas are welcome.

## Using the `JsObject` `Gc` pointer as the value for nan boxing.

At first this seems straight forward. But since there is no additional `Box` around the real value, it seems to me, that returning a reference to the `JsObject` from a nan boxed `JsValue` is impossible.
This alters the `JsValue::as_object` API to return a `Option<JsObject>` instead of `Option<&JsObject>`, which cascades out into some breaking changes in our `JsValue` APIs and some arguably less ergonomic usage patterns.
I think we should break the APIs, since when we move all pointer types and `JsValue` to being `Copy`, we basically don't want to work with references to those types anyway and always take them as values. This would of course mean a lot of breaking changes in our APIs, but I think it would be the logical thing to do.

## Benchmarks

The current changes obviously don't change anything related to the GC, but removing the additional `Box` for nan-boxed `JsObject`s seems to already make a difference. Local benchmark results:

Before

```
RESULT Richards 221
RESULT DeltaBlue 218
RESULT Crypto 208
RESULT RayTrace 485
RESULT EarleyBoyer 591
RESULT RegExp 104
RESULT Splay 829
RESULT NavierStokes 485
SCORE 324
```

After

```
RESULT Richards 276
RESULT DeltaBlue 281
RESULT Crypto 226
RESULT RayTrace 587
RESULT EarleyBoyer 721
RESULT RegExp 110
RESULT Splay 1039
RESULT NavierStokes 513
SCORE 380
```

### DIFF
--- a/core/engine/src/object/jsobject.rs
+++ b/core/engine/src/object/jsobject.rs
@@ -58,7 +58,7 @@ impl JsData for ErasedObjectData {}
 #[derive(Trace, Finalize)]
 #[boa_gc(unsafe_no_drop)]
 pub struct JsObject<T: NativeObject + ?Sized = ErasedObjectData> {
-    inner: Gc<VTableObject<T>>,
+    pub(crate) inner: Gc<VTableObject<T>>,
 }
 
 impl<T: NativeObject + ?Sized> Clone for JsObject<T> {
@@ -90,6 +90,16 @@ impl Default for JsObject {
 }
 
 impl JsObject {
+    pub(crate) fn from_raw(ptr: NonNull<GcBox<ErasedVTableObject>>) -> Self {
+        // SAFETY: The pointer is guaranteed to be valid because we just created it.
+        // `VTableObject<ErasedObjectData>` and `VTableObject<T>` have the same size and alignment.
+        unsafe {
+            JsObject {
+                inner: Gc::from_raw(ptr),
+            }
+        }
+    }
+
     /// Creates a new `JsObject` from its inner object and its vtable.
     pub(crate) fn from_object_and_vtable<T: NativeObject>(
         object: Object<T>,
--- a/core/engine/src/object/jsobject.rs
+++ b/core/engine/src/object/jsobject.rs
@@ -12,7 +12,7 @@ use super::{
     shape::RootShape,
 };
 use crate::{
-    Context, JsResult, JsString, JsValue,
+    Context, JsData, JsResult, JsString, JsValue,
     builtins::{
         array::ARRAY_EXOTIC_INTERNAL_METHODS,
         array_buffer::{ArrayBuffer, BufferObject, SharedArrayBuffer},
@@ -43,15 +43,21 @@ pub type Ref<'a, T> = boa_gc::GcRef<'a, T>;
 /// A wrapper type for a mutably borrowed type T.
 pub type RefMut<'a, T, U> = boa_gc::GcRefMut<'a, T, U>;
 
-/// An `Object` with inner data set to `dyn NativeObject`.
-pub type ErasedObject = Object<dyn NativeObject>;
+pub(crate) type ErasedVTableObject = VTableObject<ErasedObjectData>;
 
-pub(crate) type ErasedVTableObject = VTableObject<dyn NativeObject>;
+/// An `Object` with inner data set to `ErasedObjectData`.
+pub type ErasedObject = Object<ErasedObjectData>;
+
+/// A erased object data type that must never be used directly.
+#[derive(Debug, Trace, Finalize)]
+pub struct ErasedObjectData {}
+
+impl JsData for ErasedObjectData {}
 
 /// Garbage collected `Object`.
 #[derive(Trace, Finalize)]
 #[boa_gc(unsafe_no_drop)]
-pub struct JsObject<T: NativeObject + ?Sized = dyn NativeObject> {
+pub struct JsObject<T: NativeObject + ?Sized = ErasedObjectData> {
     inner: Gc<VTableObject<T>>,
 }
 
@@ -71,6 +77,8 @@ impl<T: NativeObject + ?Sized> Clone for JsObject<T> {
 #[derive(Trace, Finalize)]
 pub(crate) struct VTableObject<T: NativeObject + ?Sized> {
     #[unsafe_ignore_trace]
+    type_id: core::any::TypeId,
+    #[unsafe_ignore_trace]
     vtable: &'static InternalObjectMethods,
     object: GcRefCell<Object<T>>,
 }
@@ -87,13 +95,18 @@ impl JsObject {
         object: Object<T>,
         vtable: &'static InternalObjectMethods,
     ) -> Self {
-        let gc = Gc::new(VTableObject {
+        let ptr = Gc::into_raw(Gc::new(VTableObject {
+            type_id: core::any::TypeId::of::<T>(),
             object: GcRefCell::new(object),
             vtable,
-        });
+        }));
 
-        Self {
-            inner: coerce_gc(gc),
+        // SAFETY: The pointer is guaranteed to be valid because we just created it.
+        // `VTableObject<ErasedObjectData>` and `VTableObject<T>` have the same size and alignment.
+        unsafe {
+            JsObject {
+                inner: Gc::from_raw(ptr.cast::<GcBox<ErasedVTableObject>>()),
+            }
         }
     }
 
@@ -136,18 +149,23 @@ impl JsObject {
         data: T,
     ) -> Self {
         let internal_methods = data.internal_methods();
-        let gc = Gc::new(VTableObject {
+        let ptr = Gc::into_raw(Gc::new(VTableObject {
+            type_id: core::any::TypeId::of::<T>(),
             object: GcRefCell::new(Object {
-                data,
+                data: Box::new(data),
                 properties: PropertyMap::from_prototype_unique_shape(prototype.into()),
                 extensible: true,
                 private_elements: ThinVec::new(),
             }),
             vtable: internal_methods,
-        });
+        }));
 
-        Self {
-            inner: coerce_gc(gc),
+        // SAFETY: The pointer is guaranteed to be valid because we just created it.
+        // `VTableObject<ErasedObjectData>` and `VTableObject<T>` have the same size and alignment.
+        unsafe {
+            JsObject {
+                inner: Gc::from_raw(ptr.cast::<GcBox<ErasedVTableObject>>()),
+            }
         }
     }
 
@@ -164,9 +182,10 @@ impl JsObject {
         data: T,
     ) -> Self {
         let internal_methods = data.internal_methods();
-        let gc = Gc::new(VTableObject {
+        let ptr = Gc::into_raw(Gc::new(VTableObject {
+            type_id: core::any::TypeId::of::<T>(),
             object: GcRefCell::new(Object {
-                data,
+                data: Box::new(data),
                 properties: PropertyMap::from_prototype_with_shared_shape(
                     root_shape,
                     prototype.into(),
@@ -175,10 +194,14 @@ impl JsObject {
                 private_elements: ThinVec::new(),
             }),
             vtable: internal_methods,
-        });
+        }));
 
-        Self {
-            inner: coerce_gc(gc),
+        // SAFETY: The pointer is guaranteed to be valid because we just created it.
+        // `VTableObject<ErasedObjectData>` and `VTableObject<T>` have the same size and alignment.
+        unsafe {
+            JsObject {
+                inner: Gc::from_raw(ptr.cast::<GcBox<ErasedVTableObject>>()),
+            }
         }
     }
 
@@ -188,8 +211,8 @@ impl JsObject {
     ///
     /// Panics if the object is currently mutably borrowed.
     pub fn downcast<T: NativeObject>(self) -> Result<JsObject<T>, Self> {
-        if self.borrow().is::<T>() {
-            let ptr: NonNull<GcBox<VTableObject<dyn NativeObject>>> = Gc::into_raw(self.inner);
+        if self.is::<T>() {
+            let ptr: NonNull<GcBox<VTableObject<ErasedObjectData>>> = Gc::into_raw(self.inner);
 
             // SAFETY: the rooted `Gc` ensures we can read the inner `GcBox` in a sound way.
             #[cfg(debug_assertions)]
@@ -203,9 +226,7 @@ impl JsObject {
 
             let ptr: NonNull<GcBox<VTableObject<T>>> = ptr.cast();
 
-            // SAFETY: The conversion between an `Any` and its downcasted type must be valid.
-            // The pointer returned by `Gc::into_raw` is the same one that is passed to `Gc::from_raw`,
-            // just downcasted to the type `T`.
+            // SAFETY: We checked that `self` is of type `T`, so we can safely cast the pointer.
             let inner = unsafe { Gc::from_raw(ptr) };
 
             Ok(JsObject { inner })
@@ -225,6 +246,8 @@ impl JsObject {
 
         // SAFETY: The caller guarantees `T` is the original inner data type of the underlying
         // object.
+        // The pointer is guaranteed to be valid because we just created it.
+        // `VTableObject<ErasedObjectData>` and `VTableObject<T>` have the same size and alignment.
         unsafe {
             JsObject {
                 inner: Gc::from_raw(ptr),
@@ -241,7 +264,15 @@ impl JsObject {
     #[must_use]
     #[track_caller]
     pub fn downcast_ref<T: NativeObject>(&self) -> Option<Ref<'_, T>> {
-        Ref::try_map(self.borrow(), ErasedObject::downcast_ref)
+        if self.is::<T>() {
+            let obj = self.borrow();
+
+            // SAFETY: We have verified that the object is of type `T`, so we can safely cast it.
+            let obj = unsafe { obj.cast::<Object<T>>() };
+
+            return Some(Ref::map(obj, |r| r.data()));
+        }
+        None
     }
 
     /// Downcasts a mutable reference to the object,
@@ -253,7 +284,15 @@ impl JsObject {
     #[must_use]
     #[track_caller]
     pub fn downcast_mut<T: NativeObject>(&self) -> Option<RefMut<'_, ErasedObject, T>> {
-        RefMut::try_map(self.borrow_mut(), ErasedObject::downcast_mut)
+        if self.is::<T>() {
+            let obj = self.borrow_mut();
+
+            // SAFETY: We have verified that the object is of type `T`, so we can safely cast it.
+            let obj = unsafe { obj.cast::<Object<T>>() };
+
+            return Some(RefMut::map(obj, |c| c.data_mut()));
+        }
+        None
     }
 
     /// Checks if this object is an instance of a certain `NativeObject`.
@@ -265,7 +304,7 @@ impl JsObject {
     #[must_use]
     #[track_caller]
     pub fn is<T: NativeObject>(&self) -> bool {
-        self.borrow().is::<T>()
+        self.inner.type_id == core::any::TypeId::of::<T>()
     }
 
     /// Checks if it's an ordinary object.
@@ -620,23 +659,18 @@ Cannot both specify accessors and a value or writable attribute",
     /// Casts to a `BufferObject` if the object is an `ArrayBuffer` or a `SharedArrayBuffer`.
     #[inline]
     pub(crate) fn into_buffer_object(self) -> Result<BufferObject, JsObject> {
-        let obj = self.borrow();
-
-        if obj.is::<ArrayBuffer>() {
-            drop(obj);
+        if self.is::<ArrayBuffer>() {
             // SAFETY: We have verified that the inner data of `self` is of type `ArrayBuffer`.
             return Ok(BufferObject::Buffer(unsafe {
                 self.downcast_unchecked::<ArrayBuffer>()
             }));
         }
-        if obj.is::<SharedArrayBuffer>() {
-            drop(obj);
+        if self.is::<SharedArrayBuffer>() {
             // SAFETY: We have verified that the inner data of `self` is of type `SharedArrayBuffer`.
             return Ok(BufferObject::SharedBuffer(unsafe {
                 self.downcast_unchecked::<SharedArrayBuffer>()
             }));
         }
-        drop(obj);
 
         Err(self)
     }
@@ -653,8 +687,9 @@ impl<T: NativeObject + ?Sized> JsObject<T> {
     {
         let internal_methods = data.internal_methods();
         let inner = Gc::new(VTableObject {
+            type_id: core::any::TypeId::of::<T>(),
             object: GcRefCell::new(Object {
-                data,
+                data: Box::new(data),
                 properties: PropertyMap::from_prototype_with_shared_shape(
                     root_shape,
                     prototype.into(),
@@ -678,8 +713,9 @@ impl<T: NativeObject + ?Sized> JsObject<T> {
     {
         let internal_methods = data.internal_methods();
         let inner = Gc::new(VTableObject {
+            type_id: core::any::TypeId::of::<T>(),
             object: GcRefCell::new(Object {
-                data,
+                data: Box::new(data),
                 properties: PropertyMap::from_prototype_unique_shape(prototype.into()),
                 extensible: true,
                 private_elements: ThinVec::new(),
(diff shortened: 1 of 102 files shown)


hansl:
> This alters the JsValue::as_object API to return a Option<JsObject> instead of Option<&JsObject>, which cascades out into some breaking changes in our JsValue APIs and some arguably less ergonomic usage patterns.

I think this is a very good change, and will inevitably be done for the other "pointer" types (`BigInt`, `String`, etc). If we can make `JsObject` Copy, there shouldn't be any references to it anywhere anymore. Should improve benchmarks even further. I like this a lot.

jedel1043:
> Instead of erasing the object into a dyn NativeObject, the TypeId is stored in the object itself and used to check during casts. The default T is erased into a size 0 struct. I don't really like this approach, since one could access the erased struct.

I don't think this matters too much, because writes to zero-sized structs are replaced by no-ops IIRC. Maybe we would have to add a `#[repr(C)]` to `Object` though, so that the compiler doesn't reorder the `data` field.

I'm wondering why do we need to wrap the data with a `Box` though. Is it because of the `?Sized` requirement? Because if we're using a zero-sized type as our placeholder, I think we can remove that from `Object`.

core/engine/src/object/jsobject.rs
```diff
@@ -42,16 +42,22 @@ pub type Ref<'a, T> = boa_gc::GcRef<'a, T>;
...
+pub struct ErasedObjectData {}
+
+impl JsData for ErasedObjectData {}
 
 /// Garbage collected `Object`.
 #[derive(Trace, Finalize)]
 #[boa_gc(unsafe_no_drop)]
-pub struct JsObject<T: NativeObject + ?Sized = dyn NativeObject> {
-    inner: Gc<VTableObject<T>>,
+pub struct JsObject<T: NativeObject + ?Sized = ErasedObjectData> {
```
jedel1043: I don't think this has the correct implementation of `Trace` and `Finalize` for `JsObject<ErasedObjectData>`, because it would ignore any `Gc`s inside the data itself. I think you also need to store the Vtable of `<T as Trace>` and `<T as Finalize>` so that we can trace through a `JsObject` correctly.

jedel1043: I'm pretty sure we do something fairly similar to this in the garbage collector code

raskad:
> I don't think this matters too much, because writes to zero-sized structs are replaced by no-ops IIRC. Maybe we would have to add a `#[repr(C)]` to `Object` though, so that the compiler doesn't reorder the `data` field.

Alright, I will add some tests to check if that all works as expected.

> I'm wondering why do we need to wrap the data with a Box though. Is it because of the ?Sized requirement?

It's because of the alignment of `Object`. If we don't box `T`, the alignment would depend on the alignment of `T`. That would mean, that we cannot safely access fields on `Object` while it is erased (at least as far as I understand).

> Because if we're using a zero-sized type as our placeholder, I think we can remove that from Object.

You mean completely removing the `?Sized` requirement from `Object` and `JsObject`? That was also something I though about and I think it could be removed now, because it was only there to allow for the erased `dyn NativeObject`, correct?

core/engine/src/object/jsobject.rs
HalidOdat: Hmmm... this may not be needed, when we construct a `Gc<T>` we keep the trace and finalize function of `T` in the gc's vtable. And calling [`trace` on `Gc<T>` erases its type when it's queued](https://github.com/boa-dev/boa/blob/main/core/gc/src/pointers/gc.rs#L166-L171), and when we trace/finalize `T` we [call the vtable's trace](https://github.com/boa-dev/boa/blob/main/core/gc/src/trace.rs#L49-L53).

Casting a `Gc<T>` to `Gc<U>` of any type is safe... as long as you don't access `U` of course (assuming they having different layouts). When the object is traced/finalized/dropped then the correct `T`'s corresponding functions will be called.

So doing this should be safe :)

jedel1043: Ohhh true, that's really nice to have actually! We should definitely add a comment explaining this then, since it's pretty easy to miss

jedel1043:
> That was also something I though about and I think it could be removed now, because it was only there to allow for the erased dyn NativeObject, correct?

Yes. We could technically enable something like `Object<[u8]>` if we left the `?Sized` requirement, but I think `NativeObject` requires `Sized`, so it wouldn't be useful at all to have that if we're removing `dyn NativeObject`.

> It's because of the alignment of Object. If we don't box T, the alignment would depend on the alignment of T. That would mean, that we cannot safely access fields on Object while it is erased (at least as far as I understand).

Maybe the field is already aligned to 8 bytes thanks to the rest of the fields? We could also force `#[repr(C, align(64)]` and add padding so that the data field is always 8 byte aligned.

core/engine/src/object/jsobject.rs
HalidOdat: Yes! We could make even better if we some additional type ID data in the `Gc`'s `VTable`, we can enable easy `downcast` functionality on the `Gc` pointer level, that way we may be able to eliminate the type ID in the object vtable. :)

And space wise putting it on the static `Gc` Vtable which is per type, would be better that the object level with is per instance.

EDIT: I'll make the PR for it :smile:

raskad:
> Yes. We could technically enable something like `Object<[u8]>` if we left the `?Sized` requirement, but I think `NativeObject` requires `Sized`, so it wouldn't be useful at all to have that if we're removing `dyn NativeObject`.

Sounds good!
 
> Maybe the field is already aligned to 8 bytes thanks to the rest of the fields? We could also force `#[repr(C, align(64)]` and add padding so that the data field is always 8 byte aligned.

Can we force the struct into a specific alignment? The reference for repr C documents, that the alignment is always the alignment of the most aligned field. When I test the following code, `Object` is still 2^29 aligned:

```rust
#[repr(align(536870912))]
struct Inner {
    field: u8,
}

#[repr(C, align(64))]
struct Object<T> {
    field: T,
}

```

jedel1043:
> Can we force the struct into a specific alignment? The reference for repr C documents, that the alignment is always the alignment of the most aligned field. When I test the following code, Object is still 2^29 aligned:

Oh true, I think I got confused by how `#[repr(C)]` works. Thinking it through a bit more, to avoid the `Box` we just need to add `#[repr(C)]`, because in that case the automatic padding to align `data` would be just before `data` itself, right? Thus, the previous fields won't change their offset, making it them safe to access.

jedel1043:
> Possible alternative: JsObject could be a non generic struct and JsObject::downcast could return a typed struct. The positive of this approach would be, that access to the data T of the object can be controlled better, which would probably make everything safer. One drawback, that I currently see, is, that functions that should work on JsObject<T> would have to be either specific to one variant, or need to be implemented twice. I have not tried this approach, but I think in theory it should be possible.

Note that most of our methods are only implemented for `JsObject<dyn NativeObject>`, because most of the JS functions only take the erased version of `JsObject`. There are only a handful of methods implemented for `JsObject<T>`, so going with this approach shouldn't be that much of a problem I think.

raskad:
> Oh true, I think I got confused by how `#[repr(C)]` works. Thinking it through a bit more, to avoid the `Box` we just need to add `#[repr(C)]`, because in that case the automatic padding to align `data` would be just before `data` itself, right? Thus, the previous fields won't change their offset, making it them safe to access.

I tried that and it produced UB. I don't have the code anymore, but I think I added `repr(C)` to both the `Object` and `VTableObject`.
The offsets of the previous fields would be the same [1], but the fields inside of `Object` before `T` would not be fine, since `Object` is at a different offset.

```rust
use std::{any::TypeId, fmt::Debug};

fn main() {
    dbg!(std::mem::offset_of!(VTableObject<Erased>, t));        // 0
    dbg!(std::mem::offset_of!(VTableObject<Erased>, vtable));   // 16
    dbg!(std::mem::offset_of!(VTableObject<Erased>, data));     // 24

    dbg!(std::mem::offset_of!(VTableObject<u128>, t));          // 0
    dbg!(std::mem::offset_of!(VTableObject<u128>, vtable));     // 16
    dbg!(std::mem::offset_of!(VTableObject<u128>, data));       // 32
}

#[derive(Debug)]
struct Erased {}

#[repr(C)]
struct VTableObject<T = Erased> {
    t: TypeId,
    vtable: & 'static InternalMethods,
    data: Object<T>,
}

struct InternalMethods {
    debug: fn() -> & 'static str
}

impl Debug for InternalMethods {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.write_str((self.debug)())
    }
}

#[repr(C)]
struct Object<T = Erased> {
    properties: Vec<usize>,
    data: T,
}
```

[1]: I tried to wrap that example code with a dummy `JsObject` (just a `NonNull`) and cased it. Running it with already throws an error when accessing the `vtable` field.

jedel1043:
@raskad Ahh yeah, I didn't consider the alignment of `VTableObject`, and any other struct that wraps `Object`.

jedel1043:
The only other way I see to remove `Box` is to store the offset to the data as a pointer, then use that for access:

```rust
#[repr(C)]
struct VTableObject<T = Erased> {
    t: TypeId,
    vtable: & 'static InternalMethods,
    data_ptr: *mut Object<T>,
    data: Object<T>,
}
```

But maybe this would complicate the refactor a lot more, so it would be arguably better to use `Box` for now and leave this as a future improvement.

HalidOdat:
Hmmm... I think we can leave the removal of the box in separate PR, just so it doen't get to big :sweat_smile:....

Just wanted to give a potential solution for it: What we could do is restrict the min and max alignment that a JsData can be.

For the min we can create a wrapper struct with the `#[align(N)]` property.

```rust
// Force alignment to be the same for all `JsData` types.
#[repr(C, align(16))]
struct ObjectData<T> {
    data: T,
}
```

We can wrap object type:

```rust
#[repr(C)]
struct Object<T = Erased> {
    properties: Vec<usize>,
    data: ObjectData<T>,
}
```

As for the max we can use some const `assert!`s

```rust
impl<T> ObjectData<T> {
    const IS_VALID: bool = std::mem::align_of::<T>() <= 16;
    const ALIGNMENT_ASSERT: () = assert!(Self::IS_VALID, "Alignment of JsData must be <= 16, NOTE: you can wrap the data in a Box<T>.");

    pub fn new(value: T) -> Self {
        // force const assertion to triger when we instantiate.
        let _ = Self::ALIGNMENT_ASSERT;
        
        Self {
            data: value,
        }
    }
}
```

I couldn't get the assert to trigger when we instantiate the type, it triggers when we instantiate the `new()` function and will cause a compile error with the `assert!` message

See code (changes denoted with comments :) ): <https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4abfd21414195b7d66863ce1208fc4d5>

raskad:
Got an update on this. I rebased this on the changes from #4291 and used `GcErased` in `JsObject`. @HalidOdat I needed to switch the `*cast` functions on `GcErased` to use take a value instead of a reference, to avoid clones of `Gc<T>`. For most other scenarios, I added a `downcast_ref` function. What do you think abou this?

Building on you comment @jedel1043

> Note that most of our methods are only implemented for `JsObject<dyn NativeObject>`, because most of the JS functions only take the erased version of `JsObject`. There are only a handful of methods implemented for `JsObject<T>`, so going with this approach shouldn't be that much of a problem I think.

I added a `JsObjectTyped<T>` struct that can be used for the downcasted `JsObjects`. You where totally correct, there where only a couple of methods, that I just copied for now. The biggest breaking change would be, that users would need to change their usages. What do you all think (you can see our internal usage changes to guage the effort).

HalidOdat:
> [..] I needed to switch the `*cast` functions on `GcErased` to use take a value instead of a reference, to avoid clones of `Gc<T>`

Do we actually need this change? Could we instead call clone inside JsObject's .downcast methods?

The previous implementation had the advantage of avoiding clones (reference count increments) when accessing the type. I'm totally fine with the change either way, just curious! 😄

core/engine/src/object/jsobject.rs
```diff
@@ -82,6 +108,43 @@ impl Default for JsObject {
...
+        unsafe { self.inner.downcast_ref_unchecked::<ErasedVTableObject>() }
+    }
+
+    /// Creates a new `JsObject` from a raw pointer.
+    ///
+    /// # Safety
+    /// The caller must ensure that the pointer is valid and points to a `GcBox<ErasedVTableObject>`.
+    /// The pointer must not be null.
+    #[cfg(not(feature = "jsvalue-enum"))]
+    pub(crate) unsafe fn from_u64(value: u64) -> Self {
```
HalidOdat: Remark: (non-blocking) Maybe we should only have a `from_raw` and `into_raw` that return and take a `NonNull<GcBox<ErasedVTableObject>>`, just so we can move all the non-strict provenance code into `nan-boxing` jsvalue module.

core/engine/src/object/jsobject.rs
```diff
@@ -24,38 +24,64 @@ use crate::{
...
-pub type ErasedObject = Object<dyn NativeObject>;
+pub(crate) type ErasedVTableObject = VTableObject<ErasedObjectData>;
+
+/// An `Object` with inner data set to `ErasedObjectData`.
+pub type ErasedObject = Object<ErasedObjectData>;
 
-pub(crate) type ErasedVTableObject = VTableObject<dyn NativeObject>;
+/// A erased object data type that must never be used directly.
+#[derive(Debug, Trace, Finalize)]
+pub struct ErasedObjectData {}
```
HalidOdat:
```suggestion
pub enum ErasedObjectData {};
```

I think we can make this an enum with no variants (a non-constructable enum), this is what standard library does for [`std::convert::infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html)

We never construct it we cast into it ;)

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -311,6 +311,16 @@ mod bits {
         value_masked | MASK_OBJECT
     }
 
+    /// Returns an owned `JsObject` from a tagged value.
+    ///
+    /// # Safety
+    /// * The pointer must be a valid pointer to a `GcBox<ErasedVTableObject>`.
+    #[allow(clippy::unnecessary_box_returns)]
```
HalidOdat:
```suggestion
```
I don't think this lint is needed anymore :smile:

core/engine/src/object/jsobject.rs
```diff
@@ -1039,13 +1104,154 @@ impl<T: NativeObject + ?Sized> Debug for JsObject<T> {
...
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-iscallable
+    #[inline]
+    #[must_use]
+    #[expect(
+        unpredictable_function_pointer_comparisons,
+        reason = "can only use `ptr::fn_addr_eq` on rustc 1.85"
+    )]
+    pub fn is_callable(&self) -> bool {
+        self.inner.vtable.__call__ != ORDINARY_INTERNAL_METHODS.__call__
```
HalidOdat: nitpick (non-blocking): It doesn't change much but we could use the [`fn_addr_eq()`](https://doc.rust-lang.org/std/ptr/fn.fn_addr_eq.html). (same for `is_constructor()` below)

core/engine/src/object/jsobject.rs
```diff
@@ -865,29 +867,92 @@ impl<T: NativeObject + ?Sized> JsObject<T> {
...
 
-impl<T: NativeObject + ?Sized> AsRef<GcRefCell<Object<T>>> for JsObject<T> {
+impl<T: NativeObject> JsObjectTyped<T> {
+    /// Creates a new `JsObjectTyped` from its root shape, prototype, and data.
+    ///
+    /// Note that the returned object will not be erased to be convertible to a
+    /// `JsValue`. To erase the pointer, call [`JsObjectTyped::upcast`].
+    pub fn new<O: Into<Option<JsObject>>>(root_shape: &RootShape, prototype: O, data: T) -> Self
+    where
+        T: Sized,
```
HalidOdat: `T` is always `Sized` because of the `impl` constraint so we can remove this (applies to all functions in this `impl` block) :)

```suggestion
```

core/engine/src/object/jsobject.rs
```diff
@@ -24,38 +24,64 @@ use crate::{
...
+        }
+    }
+}
+
+/// Garbage collected typed `Object`.
+#[derive(Trace, Finalize)]
+#[boa_gc(unsafe_no_drop)]
+pub struct JsObjectTyped<T: NativeObject> {
+    pub(crate) inner: Gc<VTableObject<T>>,
 }
```
HalidOdat:
```suggestion
#[derive(Trace, Finalize)]
#[boa_gc(unsafe_no_drop)]
pub struct JsObjectTyped<T: NativeObject = ErasedObjectData> {
    pub(crate) inner: Gc<VTableObject<T>>,
}
```

Couldn't we have a default parameter here and it would eliminate the need for two `JsObject` types.

HalidOdat (commented):
Really nice work! Really like the performance boost! Check my comments if they make any sense :smile:

core/engine/src/object/jsobject.rs
raskad: Yes, we could keep it like this and avoid `JsObjectTyped`. I just wanted to try it out like this to leverage `GcErased`. As per our discussion, I personally don't think there is a big benefit to splitting the erased and typed variants, so unless someone can think of a big benefit, I would also be for keeping the default paramter.

HalidOdat: > [..] I just wanted to try it out like this to leverage GcErased. [..]

Yes, unfortunately it doesn't fit too well for this case, but we do leverage the erasure part, that being the `TypeId` with downcast methods. I guess `GcErased` can be for cases where you don't care what inner `T` is. For JsObject we do, ...at least partially, so I think it's fine.

We could still use `GcErased` in `JsObject` (if we wanted). It would just requires having a small method that we use to access inner as `VTableObject<T>` (by casting with `GcErased::cast`).

> [..] so unless someone can think of a big benefit, I would also be for keeping the default paramter [..].

Hmmm... maybe there is a benefit, reduction of code monomorfization and compile times for consuming crates? Though both points (assuming this is an issue to begin with, since the functions are small and LLVM does do code [function merging](https://llvm.org/docs/MergeFunctions.html)). Can be addressed if we wrap the non-generic code in a non-generic function, that the generic function calls.

One downside that I can see from splitting, is that we can't call the functions from `JsObject` while the object is `Typed`. Though this too can be addressed by a deref will requires some `transmute` (casting Self (`&JsObject`) to (`&JsObjectTyped<T>`) this should be safe though).

IMO, I think we should merge it back, I found the other version to be more intuitive/pretty, though I'm curious what others think @boa-dev/maintainers :smile:

raskad: > We could still use GcErased in JsObject (if we wanted). It would just requires having a small method that we used to access inner as VTableObject<T> (by casting with GcErased::cast).

And we would need a `PhantomData<T>`, right? That would be my reason just to use `Gc<T>` directly, since the code would be easier.

HalidOdat: > And we would need a `PhantomData<T>`, right? That would be my reason just to use `Gc<T>` directly, since the code would be easier.

Yes, phantom data is required to appease the compiler from the warning of unused `T`. and yes, using `Gc<T>` _will probably_ be simpler :smile:

raskad:
> Do we actually need this change? Could we instead call clone inside JsObject's .downcast methods?
> 
> The previous implementation had the advantage of avoiding clones (reference count increments) when accessing the type. I'm totally fine with the change either way, just curious! 😄

With the latest change, I only use `Gc::cast_unchecked`. As far I can see, there should be no clone with `Gc::cast_unchecked` taking `Self`.

core/engine/src/object/jsobject.rs
```diff
@@ -92,9 +119,11 @@ impl JsObject {
...
         });
 
-        Self {
-            inner: coerce_gc(gc),
-        }
+        // SAFETY: The pointer is guaranteed to be valid because we just created it.
+        // `VTableObject<ErasedObjectData>` and `VTableObject<T>` have the same size and alignment.
+        let inner = unsafe { Gc::cast_unchecked::<ErasedVTableObject>(gc) };
+
+        JsObject { inner }
```
HalidOdat:
```suggestion
        JsObject { inner: gc }.upcast()
```

We could eliminate the unsafe by using the `upcast` method which already has the unsafe code.

core/engine/src/object/jsobject.rs
```diff
@@ -138,17 +167,19 @@ impl JsObject {
...
```
HalidOdat: Same here :)

core/engine/src/object/jsobject.rs
```diff
@@ -177,9 +208,11 @@ impl JsObject {
...
```
HalidOdat: Same here :)

core/engine/src/object/jsobject.rs
```diff
@@ -188,25 +221,9 @@ impl JsObject {
...
-
-            let ptr: NonNull<GcBox<VTableObject<T>>> = ptr.cast();
-
-            // SAFETY: The conversion between an `Any` and its downcasted type must be valid.
-            // The pointer returned by `Gc::into_raw` is the same one that is passed to `Gc::from_raw`,
-            // just downcasted to the type `T`.
-            let inner = unsafe { Gc::from_raw(ptr) };
+        if self.is::<T>() {
+            // SAFETY: We have verified that the object is of type `T`, so we can safely cast it.
+            let inner = unsafe { Gc::cast_unchecked::<VTableObject<T>>(self.inner) };
```
HalidOdat: Consider using the `JsObject` downcast method :)

```suggestion
            let inner = unsafe { self.downcast_unchecked::<T>() };
```

core/engine/src/object/jsobject.rs
```diff
@@ -620,88 +651,24 @@ Cannot both specify accessors and a value or writable attribute",
...
-        let obj = self.borrow();
-
-        if obj.is::<ArrayBuffer>() {
-            drop(obj);
+        if self.is::<ArrayBuffer>() {
             // SAFETY: We have verified that the inner data of `self` is of type `ArrayBuffer`.
             return Ok(BufferObject::Buffer(unsafe {
                 self.downcast_unchecked::<ArrayBuffer>()
             }));
         }
```
HalidOdat: I think we can remove the unsafe here by chaining downcast calls :)

```suggestion
        match self.downcast::<ArrayBuffer>() {
            Ok(buffer) => return Ok(BufferObject::Buffer(buffer)),
            Err(object) => object.downcast::<SharedArrayBuffer>().map(BufferObject::SharedBuffer)
        }
```

PR #4288, opened 2025-06-20
Bump Test262 hash commit and cleanup test features

jedel1043:
Another day, another Test262 bump.

Had to cleanup a couple of feature names, because some of them were included into ES2025, and two new ones were added.


jedel1043:
> Have you glanced at the panics yet?

Yeah, they seem to be double borrows in the `slice` code. I'll patch them up in a follow up PR.

PR #4289, opened 2025-06-20
Fix panics on staging TypedArray.slice tests

jedel1043:
Follow up from #4288.

Fixes the two panics that occurred after bumping the Test262 commit hash.


HalidOdat (approved):
Looks good to me! :)

PR #4290, opened 2025-06-20
Implement small changes from Intl's 2026 spec

jedel1043:
A couple of changes were introduced in ECMA-402's latest spec. This PR implements the couple of changes that only modify already implemented builtins.


jedel1043:
Need to fix the regressions.

jedel1043:
Fixed the Intl regressions. The last regression I think it's a regress bug, but will investigate later.

PR #4291, opened 2025-06-21
Implement type erased `Gc<T>`

HalidOdat:
This is a WIP PR that implement `downcast`-ing on the `Gc<T>` pointer level by adding `TypeId` information to `Gc`'s VTable.


HalidOdat:
If this get's merged then we could use it in the #4287, see the `erased` test module.

The tests have been run and pass under miri :)

HalidOdat:
Hmmm.. the windows tests seem to be failing at doc tests only during linking... :confused: will look into it...

EDIT: Re-ran last commit's test CI tasks and it seems to fail, it seems to be unrelated to this PR...

jedel1043:
Yeah, there's a bug with MSVC's linker (https://github.com/rust-lang/rust/issues/141626)
I fixed the issue on #4290, but I could also extract the fix into its own PR for an easier review.

HalidOdat:
> [..] I could also extract the fix into its own PR for an easier review.

Nice find! ~~Yes, that be great! :smile:~~ The https://github.com/boa-dev/boa/pull/4290 PR is merged :partying_face:

raskad:
Looks very nice, I will try to rebase my PR on this and see how it works.

hansl (approved):
LGTM but I'd like to see more tests. Also merge conflicts.

jedel1043:
I'm wondering if this is useful for https://github.com/boa-dev/boa/pull/4287 though...

This erases the `T` in `Gc<T>`, but we want to erase the `T` in `Gc<Object<T>>`. Otherwise, having this would mean that we are storing the type id of both `T` and `Object<T>`

HalidOdat:
> This erases the `T` in `Gc<T>`, but we want to erase the `T` in `Gc<Object<T>>`. Otherwise, having this would mean that we are storing the type id of both `T` and `Object<T>`

We can use C-style inheritance for this (assuming proper alignment and layout). We can `cast<T>()` to any type, so a `Gc<Object<T>>` to a`Gc<Object<U>>` (this is unsafe for the mentioned reasons). In #4287 we could either store `Gc<Object<()>>` as the inner or, we can have a helper `inner` method on `JsObject` that returns a `Gc<Object<()>>` by casting a `GcErased`.

When performing an `is<T>()` or `downcast<T>()`, we just need to wrap the provided `T` with `Object<T>`, and the check should work as expected.

Unfortunately we can't eliminate all the unsafe code that was added in #4287, this mostly eliminates the `TypeId` and some unsafe in the downcast methods, while adding a feature that can potentially be used in other places?

jedel1043:
Ohhh, I saw the c inheritance example and I see what you mean. Yeah, if we just need to cast from `Object<T>` to `ObjectData` and back, then it should work nicely.

PR #4292, opened 2025-06-21
Implement backtrace information for errors

### LINKED ISSUE #1313: Add line/character numbers to errors

### LINKED ISSUE #3726: Add backtraces to errors
As part of our focus on improving our Error APIs, we need to introduce backtraces to errors generated at execution. This includes:

- [ ] Adding a `Backtrace` struct.
- [ ] Retrofitting our VM to hold the current backtrace.
- [ ] Adding names to all native functions.
- [ ] Storing the current backtrace inside every thrown `JsError`.

HalidOdat:
This Pull Request closes #1313 and closes #3726.

It changes the following:

- Add position to the `JsError` where error was thrown.
- Add position to all `Error` types where the constructor was called.
- Add shadow stack for keeping track of native and bytecode calls.
- Add limits for backtrace (`$boa.limits.backtrace`)

### DIFF
--- a/core/engine/src/bytecompiler/mod.rs
+++ b/core/engine/src/bytecompiler/mod.rs
@@ -12,7 +12,11 @@ mod register;
 mod statement;
 mod utils;
 
-use std::cell::Cell;
+use std::{
+    borrow::{Borrow, BorrowMut},
+    cell::Cell,
+    ops::{Deref, DerefMut},
+};
 
 use crate::{
     JsBigInt, JsStr, JsString, SourceText, SpannedSourceText,
@@ -405,6 +409,43 @@ pub(crate) enum BindingAccessOpcode {
     DeleteName,
     GetLocator,
     DefVar,
+}
+
+/// Manages the source position scope, push on creation, pop on drop.
+pub(crate) struct SourcePositionGuard<'a, 'b> {
+    compiler: &'a mut ByteCompiler<'b>,
+}
+impl<'a, 'b> SourcePositionGuard<'a, 'b> {
+    pub(crate) fn new(compiler: &'a mut ByteCompiler<'b>, position: Position) -> Self {
+        compiler.push_source_position(position);
+        Self { compiler }
+    }
+}
+impl Drop for SourcePositionGuard<'_, '_> {
+    fn drop(&mut self) {
+        self.pop_source_position();
+    }
+}
+impl<'a> Deref for SourcePositionGuard<'_, 'a> {
+    type Target = ByteCompiler<'a>;
+    fn deref(&self) -> &Self::Target {
+        self.compiler
+    }
+}
+impl DerefMut for SourcePositionGuard<'_, '_> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.compiler
+    }
+}
+impl<'a> Borrow<ByteCompiler<'a>> for SourcePositionGuard<'_, 'a> {
+    fn borrow(&self) -> &ByteCompiler<'a> {
+        self.compiler
+    }
+}
+impl<'a> BorrowMut<ByteCompiler<'a>> for SourcePositionGuard<'_, 'a> {
+    fn borrow_mut(&mut self) -> &mut ByteCompiler<'a> {
+        self.compiler
+    }
 }
 
 /// The [`ByteCompiler`] is used to compile ECMAScript AST from [`boa_ast`] to bytecode.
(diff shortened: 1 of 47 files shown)


HalidOdat:
```JavaScript
function x() {
  String.nonExistentFunction();
}

x();
```

```
cargo run -- test.js
Uncaught TypeError: not a callable function
    at x test.js:2:29
```

... finally... :face_holding_back_tears:

HalidOdat:
This now also resolves #3726, implemented a shadow stack that keeps tract of all calls (native and bytecode) and would make it easir to implement TCO without changing the backtraces.

A very contrived example to demonstrate the backtraces:

```JavaScript
[
  function element(arg) {
    arg(10)
  }
].forEach(
  function func(element) {
    element((value) => {
      let error = Error("Some error happend...");
      if (value == 10) {
        throw error
      }
    })
  }
)
```

```bash
cargo run -- test.js
Uncaught Error: Some error happend... (test.js:8:24)
    at <anonymous> (test.js:10:15)
    at element (test.js:3:8)
    at func (test.js:7:12)
    at forEach (native)
    at <main> (test.js:5:10)
```

NOTE: The position in `Error: Some error happend... (test.js:8:24)` is the place where the `Error` was created.

HalidOdat:
Ran the v8/quickjs benchmarks and didn't find any significant regression in performance, they were within noise range :partying_face: (this feature mostly only affect the throwing performance).

jedel1043:
> Ran the v8/quickjs benchmarks and didn't find any significant regression in performance, they were within noise range 🥳 (this feature mostly only affect the throwing performance).

Very nice! I suppose the shadow stack is cheap to build, but expensive to query for information then?

HalidOdat:
> I suppose the shadow stack is cheap to build

Yes building the stack is pretty cheap the `SourceMap` is just a wrapper for an `Rc`

> but expensive to query for information then?

Yes, this is why we don't query the shadow stack's source maps when we throw, we only taken n (were n is the backtrace limit set by the user). The actual query happens in the `to_string()` when the exceptions escapes to the root caller (i.e. `Uncaught ...`).

core/engine/src/bytecompiler/mod.rs
```diff
@@ -1036,65 +1057,71 @@ impl<'ctx> ByteCompiler<'ctx> {
...
                 self.emit_binding_access(BindingAccessOpcode::GetName, &index, dst);
             }
-            Access::Property { access } => match access {
-                PropertyAccess::Simple(access) => {
-                    let object = self.register_allocator.alloc();
-                    self.compile_expr(access.target(), &object);
+            Access::Property { access } => {
+                match access {
+                    PropertyAccess::Simple(access) => {
+                        self.push_source_position(access.field().span().start());
```
hansl: How complicated would it be to have a `SourcePositionGuard` that on `Drop` would call `pop_source_position`? Seems like it would make the code cleaner (and stronger since you can't forget to pop at the end or on `?`), but it might be impossible with the borrow checker as it is.

hansl: The way I do this is a `XyzGuard` which takes a mut to a structure, and has `DerefMut` to the structure, then use that guard to access the other methods.

core/engine/src/vm/mod.rs
```diff
@@ -666,10 +677,18 @@ impl Context {
...
         // If we hit the execution step limit, bubble up the error to the
         // (Rust) caller instead of trying to handle as an exception.
         if !err.is_catchable() {
+            if err.backtrace.is_none() {
+                err.backtrace = Some(
+                    self.vm
+                        .shadow_stack
+                        .take(self.vm.runtime_limits.backtrace_limit(), self.vm.frame.pc),
+                );
+            }
```
jedel1043, line 690: Thought: I'm wondering if this is the best place to set the backtrace info, because I don't know how this will interact with promises and generators. Did you consider setting it on the `call` and `construct` operations for function objects?

jedel1043: Oh, maybe that wouldn't work because you would not get the exact instruction where the error happened. Yeah, ignore my previous comment.

core/engine/src/bytecompiler/mod.rs
HalidOdat: Done in https://github.com/boa-dev/boa/pull/4292/commits/17a4dad24e477801496d16f40d3dba4ab57afb69 :smile:

jedel1043 (approved):
Really awesome work! Nicely done 🥳

Rafferty97:
Hey, just wanting to double check I'm not crazy - but it seems that backtraces only work for scripts at the moment, but not modules? i.e. if I run something like `cargo run -- test.js` I get a stacktrace but `cargo run -- -m test.js` does not.

HalidOdat:
@Rafferty97  You are right! The issue is that when we convert into `JsErasedError` we lose that info, I'll create a PR with a fix for that :smile:

PR #4294, opened 2025-06-23
Move the URL class to using the boa_class macro

hansl:
Instead of the js_class one which was weird to use. This is much cleaner and takes less space.

Please note we cannot move Console, since it is not a class.

### DIFF
--- a/core/runtime/src/url.rs
+++ b/core/runtime/src/url.rs
@@ -10,15 +10,15 @@
 //! [spec]: https://url.spec.whatwg.org/
 //! [mdn]: https://developer.mozilla.org/en-US/docs/Web/API/URL
 #![cfg(feature = "url")]
+#![allow(clippy::needless_pass_by_value)]
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


core/runtime/src/url.rs
```diff
@@ -58,179 +81,159 @@ impl Url {
...
-            }
-
-            fn set(this: JsClass<Url>, value: Convert<String>) {
-                url::quirks::set_hash(&mut this.borrow_mut().0, &value.0);
-            }
-        }
+    #[boa(setter)]
+    #[boa(rename = "hostname")]
+    fn set_hostname(&mut self, value: Convert<String>) {
+        let _ = url::quirks::set_hostname(&mut self.0, &value.0);
```
HalidOdat, line 104: Question: Do we not care about the error?

hansl: Setters in the spec actually never throws, and instead just fail silently. The URL is not changed. You can try this easily in your console.

PR #4304, opened 2025-06-25
Fix more Intl tests for latest ECMA402 spec

jedel1043:


tools/gen-icu4x-data/src/main.rs
```diff
@@ -60,8 +60,12 @@ fn main() -> Result<(), Box<dyn Error>> {
         .locales_for_coverage_levels([CoverageLevel::Modern])?
         .into_iter()
         .map(DataLocaleFamily::with_descendants)
-        // Some test262 tests assume the en-US locale does not fallback.
-        .chain([DataLocaleFamily::single(locale!("en-US").into())]);
+        .chain([
+            // test262 assumes the en-US locale does not fallback.
```
nekevss, line 64: nit (non-blocking): Can we link to where this is the specified behavior for test262?

nekevss (approved):
Looks good! Nice work :smile:

I had one doc nit, but it's not blocking merging

tools/gen-icu4x-data/src/main.rs
jedel1043: I can link to the tests that require those specific locales

HalidOdat (approved):

PR #4306, opened 2025-06-25
Implement native backtrace positions

HalidOdat:
This implements a native source code positions for backtraces

```JavaScript
Array.prototype.every.call([
  [10]
], element => element.forEach(
  (element) => {
    throw value
  }
))
```

```
cargo run -- test.js
Uncaught ReferenceError: value is not defined (native at core/engine/src/vm/opcode/get/name.rs:91:28)
    at <anonymous> (test.js:5:15)
    at forEach (native at core/engine/src/builtins/array/mod.rs:968:26)
    at <anonymous> (test.js:3:30)
    at every (native at core/engine/src/builtins/array/mod.rs:1375:22)
    at call (native at core/engine/src/builtins/function/mod.rs:825:14)
    at <main> (test.js:1:27)
```

### DIFF
--- a/core/engine/src/builtins/error/mod.rs
+++ b/core/engine/src/builtins/error/mod.rs
@@ -14,7 +14,7 @@ use crate::{
     Context, JsArgs, JsData, JsResult, JsString, JsValue,
     builtins::BuiltInObject,
     context::intrinsics::{Intrinsics, StandardConstructor, StandardConstructors},
-    error::JsNativeError,
+    error::{IgnoreEq, JsNativeError},
     js_string,
     object::{JsObject, internal_methods::get_prototype_from_constructor},
     property::Attribute,
@@ -130,20 +130,13 @@ pub enum ErrorKind {
 /// [`JsNativeErrorKind`][crate::error::JsNativeErrorKind].
 ///
 /// [spec]: https://tc39.es/ecma262/#sec-error-objects
-#[derive(Debug, Clone, Trace, Finalize, JsData)]
+#[derive(Debug, Clone, PartialEq, Eq, Trace, Finalize, JsData)]
 pub struct Error {
     pub(crate) tag: ErrorKind,
-    #[unsafe_ignore_trace]
-    pub(crate) position: Option<ShadowEntry>,
-}
 
-impl Eq for Error {}
-impl PartialEq for Error {
     // The position of where the Error was created does not affect equality check.
-    #[inline]
-    fn eq(&self, other: &Self) -> bool {
-        self.tag == other.tag
-    }
+    #[unsafe_ignore_trace]
+    pub(crate) position: IgnoreEq<Option<ShadowEntry>>,
 }
 
 impl Error {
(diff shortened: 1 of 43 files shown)


HalidOdat:
This should be done :partying_face: , I put the feature under the `native-backtrace` feature flag (not enabled by default, enabled by default in the our CLI).

Here is an example of with and without :)

----

```js
Array.prototype.every.call([
  [10]
], element => element.forEach(
  (element) => {
    throw value
  }
))
```

#### Without native backtraces

```
Uncaught ReferenceError: value is not defined
    at <anonymous> (test.js:5:15)
    at forEach (native)
    at <anonymous> (test.js:3:30)
    at every (native)
    at call (native)
    at <main> (test.js:1:27)
```

#### With native backtraces

```
Uncaught ReferenceError: value is not defined (native at core/engine/src/vm/opcode/get/name.rs:91:28)
    at <anonymous> (test.js:5:15)
    at forEach (native at core/engine/src/builtins/array/mod.rs:968:26)
    at <anonymous> (test.js:3:30)
    at every (native at core/engine/src/builtins/array/mod.rs:1375:22)
    at call (native at core/engine/src/builtins/function/mod.rs:825:14)
    at <main> (test.js:1:27)
```

core/engine/src/builtins/function/mod.rs
```diff
@@ -976,6 +976,11 @@ pub(crate) fn set_function_name(
 pub(crate) fn function_call(
     function_object: &JsObject,
     argument_count: usize,
+    #[allow(
+        unused_variables,
+        reason = "Only used if native-backtrace feature is enabled"
+    )]
+    native_source_info: NativeSourceInfo,
```
jedel1043: Why not use `InternalMethodContext` instead? Seems better to futureproof any other data added to the internal methods.

core/engine/src/builtins/error/mod.rs
```diff
@@ -130,20 +130,13 @@ pub enum ErrorKind {
...
 
-impl Eq for Error {}
-impl PartialEq for Error {
     // The position of where the Error was created does not affect equality check.
-    #[inline]
-    fn eq(&self, other: &Self) -> bool {
-        self.tag == other.tag
-    }
+    #[unsafe_ignore_trace]
+    pub(crate) position: IgnoreEq<Option<ShadowEntry>>,
```
jedel1043, line 139: `IgnoreEq` looks practical, but it is only used once, so I don't think it's better than just implementing `PartialEq` manually. This also avoids having a type that technically implements `Eq` in an "incorrect" way.

HalidOdat: > `IgnoreEq` looks practical, but it is only used once,

Currently used in two places the builtin `Error` and `JsNativeError`... so I'm more inclined to keep it.


> This also avoids having a type that technically implements `Eq` in an "incorrect" way.

Ohh, I didn't think of this... [Implementing `Hash` trait](https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq) with an empty implementation should fix this :smile:

HalidOdat:
> Also, the `track_caller` attributes seem really useful for debugging panics, so maybe we should add them unconditionally?

The only reason why I didn't do this was because it may introduce some small overhead, since there is an implicit caller info pointer passed around: see <https://doc.rust-lang.org/reference/attributes/codegen.html#limitations>

core/engine/src/builtins/error/mod.rs
jedel1043: Oh, didn't consider that you could implement an empty hash. Yeah, that solves all my issues with this

jedel1043:
> The only reason why I didn't do this was because it may introduce some small overhead, since there is an implicit caller info pointer passed around

In that case we can possibly enable it only on debug mode

HalidOdat:
> In that case we can possibly enable it only on debug mode

I'm leaning toward keeping it in the release build as well, the performance impact should be very minimal. Since the user explicitly enables the feature, they effectively pay only for what they use.

PR #4309, opened 2025-06-27
Fix lints for rustc 1.88

jedel1043:
New version, new lints

### DIFF
--- a/core/parser/src/parser/tests/mod.rs
+++ b/core/parser/src/parser/tests/mod.rs
@@ -23,7 +23,7 @@ use boa_ast::{
         FunctionBody, FunctionDeclaration,
     },
     scope::Scope,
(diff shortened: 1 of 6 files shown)


core/parser/src/parser/tests/mod.rs
```diff
@@ -732,14 +732,17 @@ fn hashbang_use_strict_no_with() {
 }
 
 #[test]
-#[ignore]
 fn hashbang_use_strict_with_with_statement() {
     check_script_parser(
         r#"#!\"use strict"
 
         with({}) {}
```
HalidOdat: Remark: Maybe use `indoc!`, so rust code indent doesn't affect the span possition :)

HalidOdat (approved):

PR #4310, opened 2025-06-27
Regex Improvements

jasonwilliams:
- inline the take_while_ascii_pred function
- change flag handling to stack allocation by using a fixed sized array
- Parse regex literals at an earlier opportunity when inspecting expressions (to reduce stack depth)


AdithyaLaakso:
I took your suggestion in the issues and made a new function for this issue using an arrayvec which I believe will be marginally more memory efficient in the vast majority of cases (where no flags are given) than allocating an array. Additionally, I added error handling for the case where more than 8 regex flags are added and preserved the original version of the function for future use. 

However, your pull request covers a lot of area that mine does not we should merge them!

Issue #1402, opened 2021-07-16
stack-overflow caused by deep call stack


charlesxsh:
**Describe the bug**
stack-overflow caused by deep call stack

**To Reproduce**
Input: 
`+[4446<4444^/ /g/  /g[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[[44444   ]`

Code:
```rust
fn main(){
    // let data = input

    let mut context = Context::new();
    context.eval(data);

}
```

Output
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==11135==ERROR: AddressSanitizer: stack-overflow on address 0x7ffdacb98c40 (pc 0x55d5d74cc4d5 bp 0x7ffdacb9a430 sp 0x7ffdacb98c40 T0)
    #0 0x55d5d74cc4d5 in _$LT$boa..syntax..parser..expression..assignment..exponentiation..ExponentiationExpression$u20$as$u20$boa..syntax..parser..TokenParser$LT$R$GT$$GT$::parse::he8285e05f6e8dcc5 /home/xsh/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.11.0/src/syntax/parser/expression/assignment/exponentiation.rs:84
    #1 0x55d5d74ca2dd in _$LT$boa..syntax..parser..expression..MultiplicativeExpression$u20$as$u20$boa..syntax..parser..TokenParser$LT$R$GT$$GT$::parse::he0cd2b68c53a87cd /home/xsh/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.11.0/src/syntax/parser/expression/mod.rs:82:27
    #2 0x55d5d74c7940 in _$LT$boa..syntax..parser..expression..AdditiveExpression$u20$as$u20$boa..syntax..parser..TokenParser$LT$R$GT$$GT$::parse::h4ae7e82483a469c8 /home/xsh/.cargo/registry/src/github.com-1ecc6299db9ec823/Boa-0.11.0/src/syntax/parser/expression/mod.rs:82:27
[...]

PR #4313, opened 2025-06-29
Parser rewriting

### LINKED ISSUE #4301: Parser overflows stack in WASM
_(Apologise for the duplication - accidentally created the ticket from the wrong GH account)_

**Describe the bug**
When compiled to WASM (`wasm-pack --debug`), the parser exceeds the call stack when fed certain input. I presumed it was getting stuck in an infinite recursion loop, but after compiling the exact same code in release mode, the issue vanished.

This implies that the root cause is that, in debug mode, the parser is prone to overflowing the stack simply due to being very recursive in its implementation.

**To Reproduce**
This is the minimal Rust code I found that reproduces the bug:
```
#[wasm_bindgen]
pub fn fail_case() {
    let src = r#"
        const a = function () {
            b((c = d([])))
        }
    "#;

    let context = &mut boa_engine::Context::builder().build().unwrap();
    let source = boa_engine::Source::from_bytes(src);
    let _module = boa_engine::Module::parse(source, None, context).unwrap();
}
```

Compiling to WASM and calling from JS in Chrome yields this stack trace:

Image

**Expected behavior**
Successful parse without a stack overflow

- OS: MacOS
- Version: Sequoia 15.5
- Target triple: wasm32-unknown-unknown
- Rustc version: rustc 1.87.0 (17067e9ac 2025-05-09)

Nikita-str:
This Pull Request potentially(if it will be done) closes #4301 & #4089 & #1402 and allow to forget about stack overflow in the parser.

# The main idea

The main idea is next:
* all parse nodes will return `ControlFlow` with next variants: `Done` or `SubParse{parse_node, continuation_point}`
* there is a main loop that handles `ControlFlow`s
* instead of calling a sub-parse-node when it necessary a parse node returns corresponding `ControlFlow`

So parsing will be looks like:
```txt
main loop -> parse node A
main loop <- parse node A (cmd: call sub parse node B)
main loop -> parse node B
<... maybe here parse node B calls some sub nodes through main loop ...>
main loop <- parse node B (cmd: Done)
main loop -> parse node A
<... maybe here parse node A calls again sub node B or call another sub parse node C through main loop ...>
main loop <- parse node A (cmd: Done)
main loop -> return last result (parsed `StatementList` for `ScriptBody`)
```

# Some more details & explanation

Simple rewrite case (actually this case can be simplified by adding `Pass` command  to `ControlFlow`).
Next one:
https://github.com/boa-dev/boa/blob/27bdda588c9338e73f8883f134fdc83b75e16da6/core/parser/src/parser/statement/declaration/mod.rs#L64-L90
Transforms to this:
https://github.com/boa-dev/boa/blob/27bdda588c9338e73f8883f134fdc83b75e16da6/core/parser/src/parser/statement/declaration/mod.rs#L94-L127
On first call of a parse node the `continue_point`  is equal to `0`. So we ignore first `if` and do all actions as before, except for not going deeper but return from the parse call (by calling `parse_cmd![[SUB PARSE]: node]`: it change `state` and return valid `ControlFlow`) into `ParseLoop` and then `ParseLoop` call the sub-parsing node next.

---

I don't sure -- should I name the `continue_point`s in the function, or stay it like magic numbers? (0, 1, 2, and so on; mostly used only 0 and 1 but each time they have different sense depending on which sub calls were called)

---

Some small pieces of code will be repeated often but they cannot be moved to a function due to different enum variants, so I moved them into macro_rule `parse_cmd!`, but I don't sure if such syntax is acceptable, so that's worth to discuss too.

---

Right now rewritten functions looks like:
```rust
if continue_point = 1 {
    <.. continue_point_1 case ..>
} else if continue_point = 2 {
    <.. continue_point_2 case ..>
} else if ... { <...> }

<.. continue_point_0 case (execution path on first call of a parse node) ..>
```
Maybe it's better to write as next to make the code easier to understand, I don't sure:
```rust
if continue_point = 0 {
    ..
} else if continue_point = 1 {
    ..
} else if ... { .. }
```

---

⚠️ This approach will make it harder to find sub-parse-node calls because you cannot search where `parse` called (it will be called from `ParseLoop` always), to find it you have to see where the Node created.

# Why is it a draft PR?
⚠️ This will solve the stackoverflow problem only when all parse nodes are rewritten. Because if any node call sub-parse-node outside of `ParseLoop` then it should be parsed in old way. Now only 3 parse nodes (out of ~100) have been rewritten to check if there will be any problems with this approach (it seems there will not be any). 

When all nodes will be rewritten with `parse_loop`, the `parse` function & trait impl will be removed.

---

I'm expecting some discussion around this approach, and I would prefer to wait for approval before proceeding with rewriting the rest of the nodes, since this looks like a task for several days.

### DIFF
--- a/core/parser/src/parser/parse_loop.rs
+++ b/core/parser/src/parser/parse_loop.rs
@@ -1,3 +1,4 @@
+use boa_ast::declaration::LexicalDeclaration;
 use boa_ast::Position;
 use boa_interner::Interner;
 
@@ -30,6 +31,13 @@ macro_rules! parse_cmd {
             return Err($state.general_error(concat!("expect `", stringify!($variant) ,"` node")))
         };
         ret
+    }};
+
+    // or move into `peek_node!`
+    [[PEEK NODE]: $state:ident match $($variants:ident)+] => {{
+        if !matches!($state.peek_node(), Some($(| $crate::parser::ParsedNode::$variants(_))+)) {
+            return Err($state.general_error(concat!("expect `", $(stringify!($variants), "` | `",)+ "` node")))
+        }
     }};
 
     // or move into `sub_parse!`
@@ -123,8 +131,8 @@ where R: ReadChar,
 pub(super) struct ParseState<'a, R> {
     nodes: Vec<ParsedNode>,
     saved_state: Vec<SavedState>,
-    cursor: &'a mut Cursor<R>,
-    interner: &'a mut Interner,
+    pub cursor: &'a mut Cursor<R>,
+    pub interner: &'a mut Interner,
 }
 impl<'a, R: ReadChar> ParseState<'a, R> {
     pub(super) fn new(cursor: &'a mut Cursor<R>, interner: &'a mut Interner) -> Self {
@@ -161,6 +169,9 @@ impl<'a, R: ReadChar> ParseState<'a, R> {
 
     pub(super) fn pop_node(&mut self) -> ParseResult<ParsedNode> {
         self.nodes.pop().ok_or_else(||self.general_error("expect parsed node"))
+    }
+    pub(super) fn peek_node(&mut self) -> Option<&ParsedNode> {
+        self.nodes.last()
     }
 
     pub(super) fn pop_local_state(&mut self) -> ParseResult<SavedState> {
--- a/core/parser/src/parser/statement/declaration/hoistable/mod.rs
+++ b/core/parser/src/parser/statement/declaration/hoistable/mod.rs
@@ -27,6 +27,7 @@ use crate::{
     source::ReadChar,
     Error,
 };
+use crate::{parse_cmd, parser::{ControlFlow, TokenLoopParser, parse_loop::ParseState}};
 use boa_ast::{
     self as ast,
     expression::Identifier,
@@ -122,6 +123,14 @@ where
             }
             _ => unreachable!("unknown token found: {:?}", tok),
         }
+    }
+}
+
+impl<R: ReadChar> TokenLoopParser<R> for HoistableDeclaration {
+    fn parse_loop(&mut self, state: &mut ParseState<'_, R>, _continue_point: usize) -> ParseResult<ControlFlow<R>> {
+        let (cursor, interner) = state.mut_inner();
+        let ok = self.parse(cursor, interner)?;
+        parse_cmd!([DONE]: state <= Declaration(ok))
     }
 }
 
--- a/core/parser/src/parser/statement/declaration/lexical.rs
+++ b/core/parser/src/parser/statement/declaration/lexical.rs
@@ -18,6 +18,7 @@ use crate::{
     source::ReadChar,
     Error,
 };
+use crate::{parse_cmd, parser::{ControlFlow, TokenLoopParser, parse_loop::ParseState}};
 use ast::operations::bound_names;
 use boa_ast::{self as ast, declaration::Variable, Keyword, Punctuator};
 use boa_interner::{Interner, Sym};
@@ -118,6 +119,14 @@ where
         }
 
         Ok(lexical_declaration)
+    }
+}
+
+impl<R: ReadChar> TokenLoopParser<R> for LexicalDeclaration {
+    fn parse_loop(&mut self, state: &mut ParseState<'_, R>, _continue_point: usize) -> ParseResult<ControlFlow<R>> {
+        let (cursor, interner) = state.mut_inner();
+        let ok = self.parse(cursor, interner)?;
+        parse_cmd!([DONE]: state <= Declaration(ok.into()))
     }
 }
 
--- a/core/parser/src/parser/statement/declaration/mod.rs
+++ b/core/parser/src/parser/statement/declaration/mod.rs
@@ -91,10 +91,39 @@ where
 }
 
 impl<R: ReadChar> TokenLoopParser<R> for Declaration {
-    fn parse_loop(&mut self, state: &mut ParseState<'_, R>, _continue_point: usize) -> ParseResult<ControlFlow<R>> {
-        let (cursor, interner) = state.mut_inner();
-        let ok = self.parse(cursor, interner)?;
-        parse_cmd!([DONE]: state <= Declaration(ok))
+    fn parse_loop(&mut self, state: &mut ParseState<'_, R>, continue_point: usize) -> ParseResult<ControlFlow<R>> {
+        if continue_point == 1 {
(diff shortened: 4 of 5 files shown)


jasonwilliams:
Thank you @Nikita-str for putting the time and effort into writing this up. I support moving to a trampoline style parsing with a loop and state machine. The recursive descent parsing doesn’t suit rust so much ([no tail call optimisation](https://github.com/rust-lang/rfcs/issues/2691)) and causes multiple issues.

I understand that statically it’s not as easy to find sub parse nodes, but I hope we can improve debugging but having a debug_log which prints which parser is currently running when a flag is passed (we don’t have this today).

Do you know if this can be worked on incrementally?  Or do you plan to just keep updating this? Maybe maintainers can help if they have access, we could have a check list at the top and keep rebasing as we go along (the parser doesn’t beg touched as much these days so I don’t think it will conflict that often).

im interested to hear from the other maintainers.

Nikita-str:
> this can be worked on incrementally?

I think, incrementally is not an option, because if any node A call sub-parse-node B outside of `ParseLoop` then the Loop cannot handle any subsequent calls from node B (or otherwise the Loop will return handling on wrong position), and therefore it's parsed in the old way (this is why both `parse` and `parse_loop` are now preserved for rewritten nodes). So such changes should be done for all nodes to make sense.

---

> maintainers can help if they have access

That would be good, actually. We can do it in parallel. It's a very easily separable task, and with a few extra hands this can be done in 1-2 days.

jedel1043:
I feel like this may be over engineering something that should have a very simple solution.

Seeing the original issue, the parser works fine on release mode, so there must be some set of optimizations that the compiler is doing to reduce the amount of recursion. My suggestion would be that we should try to look at the assembly of a release build, then carefully try to force the compiler to apply the same optimizations in debug mode.

Nikita-str:
> My suggestion would be that we should try to look at the assembly of a release build, then carefully try to force the compiler to apply the same optimizations in debug mode.

@jedel1043 
The problem here, is that in debug mode (with `opt-level = 0`) the stack structures take too many memory and it use up the stack too quickly. Also with `opt-level = 0` there is no auto function splitting (to make less stack allocation) & no auto passing through some functions (in code A call B then B call C, then with `opt-level = 1` (or more) it can be transformed into A call C directly).

To temporarily solve it you even need really splitting the functions (after making some structure size less -- what seems good to me; only structures' size reduction is not enough).

For more details see [this PR](https://github.com/boa-dev/boa/pull/4191): first part of commits is seems ok to me, and second part is ugly (because you need splitting the function to solve the problem with `opt-level = 0`).

---

But maybe, in current PR you need to do something like the splitting (you need to continue execution from `continue_point`)... but at least it would not be a temporary solution, but a long-term one.

jedel1043:
> The problem here, is that in debug mode (with opt-level = 0) the stack structures take too many memory and it use up the stack too quickly. Also with opt-level = 0 there is no auto function splitting (to make less stack allocation) & no auto passing through some functions (in code A call B then B call C, then with opt-level = 1 (or more) it can be transformed into A call C directly).

If the problem is the size of ast nodes, I think we should just stop having ast nodes in the stack, and use something like `oxc_allocator` to have an arena of all ast nodes. This way we don't need to deal with trampolining functions, and it also gives the nice bonus of making the drop of ast trees really cheap.

jasonwilliams:
> > The problem here, is that in debug mode (with opt-level = 0) the stack structures take too many memory and it use up the stack too quickly. Also with opt-level = 0 there is no auto function splitting (to make less stack allocation) & no auto passing through some functions (in code A call B then B call C, then with opt-level = 1 (or more) it can be transformed into A call C directly).
> 
> If the problem is the size of ast nodes, I think we should just stop having ast nodes in the stack, and use something like `oxc_allocator` to have an arena of all ast nodes. This way we don't need to deal with trampolining functions, and it also gives the nice bonus of making the drop of ast trees really cheap.

I don’t think having heap allocated nodes would help with performance of the parser as we need to visit them a lot so it would create a new problem with back and forth trips to the heap (this would need to be benchmarked).

The idea of specialized allocation for nodes on the AST also seems odd if we drop the whole tree anyway once we’ve moved onto the VM stage, unless I’m misunderstanding something.

> Seeing the original issue, the parser works fine on release mode, so there must be some set of optimizations that the compiler is doing to reduce the amount of recursion

I’m guessing the compiler is aggressively inlining during release mode so we don’t end up with deep stacks.  So if we wanted to recreate this in debug mode we would need to add `inline(always)` in a lot of places across the parser. I can’t say I’m sure if that’s a proper fix for this problem.

I’d be open to seeing how a loop & state machine fairs in performance to what we have today.

jedel1043:
> I don’t think having heap allocated nodes would help with performance of the parser as we need to visit them a lot so it would create a new problem with back and forth trips to the heap (this would need to be benchmarked).

It's not for performance though. The problem that we have right now is that our AST nodes use too much space in the stack, which gives little room for function calls. Moving every AST node to the heap makes it such that more recursive calls can be made. We would obviously still need to rewrite the parser to be smarter about recursive operators (`[[[[[]]]]]` and such, maybe it can be done with a pratt parser), but it should be a smaller change than having to convert everything to trampolines.

jasonwilliams:
> > I don’t think having heap allocated nodes would help with performance of the parser as we need to visit them a lot so it would create a new problem with back and forth trips to the heap (this would need to be benchmarked).
> 
> It's not for performance though. The problem that we have right now is that our AST nodes use too much space in the stack, which gives little room for function calls. Moving every AST node to the heap makes it such that more recursive calls can be made. We would obviously still need to rewrite the parser to be smarter about recursive operators (`[[[[[]]]]]` and such, maybe it can be done with a pratt parser), but it should be a smaller change than having to convert everything to trampolines.

Oh yeah I know the original issue isn’t about performance. I was saying that to move nodes to the heap could end up harming parser performance despite fixing this specific issue. The trade off wouldn’t be worth it.

instead, as you say we would be best finding ways to make the parser smarter around recursive operations. Or have a parser that wasn’t subject to these problems. The answer may be a bit of both.

jedel1043:
> Oh, yeah I know the original issue isn’t about performance. I was saying that to move nodes to the heap could end up harming parser performance despite fixing this specific issue. The trade off wouldn’t be worth it.

It could also improve performance! My thought is that right now any inner expression needs to be allocated on the heap, which distributes expressions throughout the heap, causing a lot of cache misses when you try to access related expressions. If we use bumpalo, related expressions will live near each other, making multiple accesses very cheap if the expressions fit in the same cache line.

PR #4314, opened 2025-06-30
cli: use `mimalloc` as global allocator on windows

cijiugechu:
It changes the following:

- use `mimalloc` on windows

## Benchmark:

main:

```
PROGRESS Richards
RESULT Richards 56.2
PROGRESS DeltaBlue
RESULT DeltaBlue 61.0
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 89.8
PROGRESS RayTrace
RESULT RayTrace 150
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 174
PROGRESS RegExp
RESULT RegExp 44.8
PROGRESS Splay
RESULT Splay 238
PROGRESS NavierStokes
RESULT NavierStokes 186
SCORE 106
undefined
```

PR:

```
PROGRESS Richards
RESULT Richards 105
PROGRESS DeltaBlue
RESULT DeltaBlue 108
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 118
PROGRESS RayTrace
RESULT RayTrace 255
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 297
PROGRESS RegExp
RESULT RegExp 59.5
PROGRESS Splay
RESULT Splay 420
PROGRESS NavierStokes
RESULT NavierStokes 268
SCORE 169
undefined
```


Cargo.lock
```diff
@@ -790,6 +791,15 @@ dependencies = [
  "error-code",
 ]
 
+[[package]]
+name = "cmake"
```
HalidOdat, line 795: Discussion: Should `mimalloc` (as well as other allocators) be feature gated? Since this adds a requirement for `cmake` to be available on the host system for it to compile.

Maybe we could feature gate all custom allocators under a `fast-allocator` feature (or something like that), it can be on by default for the CLI, that way the user can decide :)

Any thoughts on this @boa-dev/maintainers

nekevss: Benchmarks results are super interesting. Especially seeing just baseline performance gains an allocator can provide.

I'm inclined to feature gate this because of the build dependency unless we have solid proof that having cmake as a build dependency won't be too big of an issue on windows. But I'm also not going to die on the hill that it should be feature gated.

jedel1043: Agreed on having anything that requires C build tools behind features.

jasonwilliams: Yep, agree this should be feature gated due to the C build tools features

cijiugechu: > Maybe we could feature gate all custom allocators under a fast-allocator feature (or something like that), it can be on by default for the CLI, that way the user can decide :)

Considering that `jemallocator` does not depend on cmake, is it an acceptable solution to feature gate mimalloc separately?

HalidOdat: > Considering that `jemallocator` does not depend on cmake, is it an acceptable solution to feature gate mimalloc separately?

Hmm, I believe it would be best to place all of them under a `fast-allocator` feature flag. While jemalloc doesn't rely on CMake, it does depend on a C build system, specifically make and a C compiler, which are typically available by default on most distributions, so it hasn't been a major issue so far.

That said, I think it would still be a good idea to let users choose whether they want to compile with the custom allocator or stick with the default. We could enable this feature by default (by adding it to the `default` feature set), but still leave the option open for users to opt out if needed.

cijiugechu: Done

.github/workflows/release.yml
```diff
@@ -120,4 +120,4 @@ jobs:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: target/${{ matrix.target }}/release/boa${{ matrix.binary_extension }}
           asset_name: boa-${{ matrix.target }}${{ matrix.binary_extension }}
-          tag: ${{ github.ref }}
+          tag: ${{ github.ref }}
```
HalidOdat: whitespace :policeman: missing newline

```suggestion
          tag: ${{ github.ref }}

```

HalidOdat (approved):
Looks perfect to me! :smile: 

Just a very minor nitpick.

.github/workflows/release.yml
cijiugechu: Thanks for catching that :)

PR #4315, opened 2025-07-02
Bump rustc edition to 2024 and version to 1.88

jedel1043:
This PR bumps our rustc version to 1.88, and bumps our edition to 2024.

Bumping the edition comes with a lot of formatting changes, but those are pretty much isolated in 0204830334fd482cd6026e46dd11338d4c391e3c. The rest of the changes use the newly introduced if let chains to improve some snippets of code.

### DIFF
--- a/cli/src/debug/function.rs
+++ b/cli/src/debug/function.rs
@@ -59,20 +59,18 @@ fn flowgraph(_this: &JsValue, args: &[JsValue], context: &mut Context) -> JsResu
             .with_message(format!("expected object, got {}", value.type_of()))
             .into());
     };
-    let format;
+    let mut format = FlowgraphFormat::Mermaid;
     let mut direction = Direction::LeftToRight;
-    if let Some(arguments) = args.get(1).and_then(JsValue::as_object) {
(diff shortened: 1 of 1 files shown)


cli/src/debug/function.rs
```diff
@@ -59,7 +59,7 @@ fn flowgraph(_this: &JsValue, args: &[JsValue], context: &mut Context) -> JsResu
             .with_message(format!("expected object, got {}", value.type_of()))
             .into());
     };
-    let mut format = FlowgraphFormat::Mermaid;
```
nekevss: I don't think I've seen this syntax before ... Has this been around for a while?

jedel1043: Yeah, pretty much. This is only allowed if all code paths either diverge or initialize your variable.

nekevss: Yeah, that's what I was figuring. I'd assume it probably compiles to something similar to "let format = [ExpressionStatement]"

HalidOdat (approved):
Looks good to me! :)

PR #4316, opened 2025-07-02
Refactor the `Binding` modifiers to use bitflags

cijiugechu:
It changes the following:

- Use bitflags to represent modifiers instead of multiple booleans.

## Benchmark

main:

```
PROGRESS Richards
RESULT Richards 56.2
PROGRESS DeltaBlue
RESULT DeltaBlue 61.0
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 89.8
PROGRESS RayTrace
RESULT RayTrace 150
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 174
PROGRESS RegExp
RESULT RegExp 44.8
PROGRESS Splay
RESULT Splay 238
PROGRESS NavierStokes
RESULT NavierStokes 186
SCORE 106
undefined
```

PR:

```
PROGRESS Richards
RESULT Richards 57.1
PROGRESS DeltaBlue
RESULT DeltaBlue 60.9
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 88.9
PROGRESS RayTrace
RESULT RayTrace 154
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 177
PROGRESS RegExp
RESULT RegExp 45.0
PROGRESS Splay
RESULT Splay 245
PROGRESS NavierStokes
RESULT NavierStokes 189
SCORE 107
undefined
```

### DIFF
--- a/core/ast/src/scope.rs
+++ b/core/ast/src/scope.rs
@@ -31,7 +31,7 @@ impl BindingFlags {
     fn is_strict(self) -> bool {
         self.contains(BindingFlags::STRICT)
     }
-    fn is_escapes(self) -> bool {
+    fn escapes(self) -> bool {
         self.contains(BindingFlags::ESCAPES)
     }
     fn is_accessed(self) -> bool {
@@ -44,6 +44,24 @@ struct Binding {
     name: JsString,
     index: u32,
     flags: BindingFlags,
+}
+
+impl Binding {
+    fn is_mutable(&self) -> bool {
+        self.flags.is_mutable()
+    }
+    fn is_lex(&self) -> bool {
+        self.flags.is_lex()
+    }
+    fn is_strict(&self) -> bool {
+        self.flags.is_strict()
+    }
+    fn escapes(&self) -> bool {
+        self.flags.escapes()
+    }
+    fn is_accessed(&self) -> bool {
+        self.flags.is_accessed()
+    }
 }
 
 /// A scope maps bound identifiers to their binding positions.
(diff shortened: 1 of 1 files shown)


core/ast/src/scope.rs
```diff
@@ -347,14 +375,17 @@ impl Scope {
             return;
         }
         let binding_index = bindings.len() as u32;
+        let mut flags = BindingFlags::LEX;
+        if strict {
+            flags |= BindingFlags::STRICT;
+        }
```
HalidOdat: We can simplify this if we use [`Flags::set()`](https://docs.rs/bitflags/latest/bitflags/trait.Flags.html#method.set) (applies to all)
```suggestion
        flags.set(BindingFlags::STRICT, strict);
```

core/ast/src/scope.rs
```diff
@@ -2,23 +2,48 @@
...
+    fn is_mutable(self) -> bool {
+        self.contains(BindingFlags::MUTABLE)
+    }
+    fn is_lex(self) -> bool {
+        self.contains(BindingFlags::LEX)
+    }
+    fn is_strict(self) -> bool {
+        self.contains(BindingFlags::STRICT)
+    }
+    fn is_escapes(self) -> bool {
```
HalidOdat:
```suggestion
    fn escapes(self) -> bool {
```

I think we should rename this to just `escapes`, `is_escapes` sounds odd.

HalidOdat (commented):
Looks good! Just some minor suggestions :smile:

core/ast/src/scope.rs
nekevss: It may be worth while moving this up to Binding as well. It's probably going to be easier to reason about `bindings.escapes()` vs. `bindings.flags.escapes()`. The latter seems like more of an implementation detail leaking out.

PR #4318, opened 2025-07-03
Migrate temporal_rs from 0.0.9 to 0.0.10

HalidOdat:
This PR migrates `temporal_rs` to the recent `0.0.10` version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3597,9 +3597,9 @@ checksum = "1ac9aa371f599d22256307c24a9d748c041e548cbf599f35d890f9d365361790"
 
 [[package]]
 name = "temporal_rs"
-version = "0.0.9"
+version = "0.0.10"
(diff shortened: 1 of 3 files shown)


nekevss:
Huh, curious what the error messages are. If there's regressions on the `temporal_rs` side, we should file them as issues and try to fix them before next release.

HalidOdat:
Yep, currently looking into it :smile:

HalidOdat:
https://github.com/boa-dev/temporal/pull/389 should fix the regressions :partying_face:

@nekevss should we merge this or wait for a temporal_rs release?

nekevss (approved):
I'm fine with merging as is.

The only regression I'm slightly concerned about currently is the `order-of-operations.js`, but we can always look into that after the next release once the merged fix is available.

We're on a much more regular release schedule too since now with the V8 work, so it really shouldn't be too long before its fixed.

PR #4321, opened 2025-07-04
Add arm64 linux nightly build

### LINKED ISSUE #4320: linux arm64 nightlies
Having linux arm64 nightly builds would be very nice :)

Currently blocks [test262.fyi v2](https://github.com/test262-fyi/data/issues/98), currently compiling myself but it takes long, breaks often, and is not very nice to deal with :P

ty!

nekevss:
Hopefully closes #4320.

This adds arm64 linux builds to the nightly CI. I don't really have a good way to check if this does successfully build though.


jedel1043:
If you want to test this properly, you should add this target to our list of PR targets IMO

HalidOdat (approved):
Nice addition! Looks good to me! :)

PR #4325, opened 2025-07-08
Add a new `Nullable<T>` type that deserialize null to `Nullable::<T>::Null`

hansl:
Now `Option<T>` deserializes `undefined` to `None`, but keep `null` as deserialization for the inner type.

This should clarify the API and distinguish between Nullable and Optional when dealing with JavaScript values in the host.


core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+///
+/// This is used to distinguish between values in JavaScript that can be
+/// `null` (i.e. `Nullable<T>`) vs. those of which that can be `undefined`
+/// (i.e. `Option<T>`).
+///
+/// `Nullable<T>` tries to be close in API surface to the standard
+/// `Option<T>`, though with a much smaller API scope, and can be
+/// transformed into and from `Option<T>` for free.
+///
+/// # Why a new type?
```
nekevss: nit: h2 / ## here

(and potentially elsewhere)

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl<T> Nullable<T> {
+    /// Returns `true` if this value is `Nullable::Null`.
+    pub const fn is_null(&self) -> bool {
+        matches!(self, Nullable::Null)
+    }
+
+    /// Returns `true` if this value is `Nullable::NotNull`.
+    pub const fn is_not_null(&self) -> bool {
```
nekevss: thought: this may be better as `pub fn is<T: PartialEq>(&self)`

`is_not_null` could also be expressed as `!v.is_null()`, so it seems like a duplicate API, but being able to compare the interior value could be useful.

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
+//! A value that can be `null` in JavaScript, but not `undefined`.
+
+use crate::value::{TryFromJs, TryIntoJs};
+use boa_engine::{Context, JsResult, JsValue};
+
+/// A value that can be `null` in JavaScript, but not `undefined`.
```
nekevss: praise: great docs!

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+///
+/// This is used to distinguish between values in JavaScript that can be
+/// `null` (i.e. `Nullable<T>`) vs. those of which that can be `undefined`
```
HalidOdat:
```suggestion
/// `null` (i.e. [`Nullable<T>`]) vs. those of which that can be `undefined`
```

Nitpick (non-blocking): And all places where it applies. Has the advantage of creating a link in docs in the event we rename something, rust-analyzer renames them too.

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+/// (i.e. `Option<T>`).
```
HalidOdat:
```suggestion
/// (i.e. [`Option<T>`]).
```

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+/// do the conversion. This new type makes this process much simpler.
+///
+/// It also means that there is an asymmetry between `JsValue` to `Option<T>`
+/// then `Option<T>` back to `JsValue`, leading to unintuitive errors. Having
+/// a type `Nullable<T>` that acts like `Option<T>` but (de-)serialize to
+/// `null` clarifies all usage.
+///
+/// # How can I do `EitherNullOrUndefined<T>`?
+/// The best way is to use `Nullable<Option<T>>` and convert into `Option<T>`
+/// using `.flatten()`. Please note that the reverse (`Nullable<Option<T>>`)
```
HalidOdat:
```suggestion
/// using [`Nullable::flatten()`]. Please note that the reverse (`Nullable<Option<T>>`)
```

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+/// # use boa_engine::{Context, JsResult, JsValue};
+/// # use boa_engine::value::Nullable;
+/// # let context = &mut Context::default();
+/// let mut v: JsResult<Nullable<Option<u8>>> = JsValue::undefined().try_js_into(context);
+/// assert_eq!(v, Ok(Nullable::NonNull(None)));
+/// v = JsValue::null().try_js_into(context);
+/// assert_eq!(v, Ok(Nullable::Null));
+/// v = JsValue::from(42).try_js_into(context);
+/// assert_eq!(v, Ok(Nullable::NonNull(Some(42))));
+/// assert_eq!(v.unwrap().flatten(), Some(42));
```
HalidOdat:
```suggestion
/// let v: JsResult<Nullable<Option<u8>>> = JsValue::undefined().try_js_into(context);
/// assert_eq!(v, Ok(Nullable::NonNull(None)));
///
/// let v = JsValue::null().try_js_into(context);
/// assert_eq!(v, Ok(Nullable::Null));
///
/// let v = JsValue::from(42).try_js_into(context);
/// assert_eq!(v, Ok(Nullable::NonNull(Some(42))));
/// assert_eq!(v.unwrap().flatten(), Some(42));
```

nekevss (commented):
Have some general feedback. It's looking pretty good overall! Had one thought about one of the APIs.

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+pub enum Nullable<T> {
+    /// The value was not defined in JavaScript.
+    Null,
+
+    /// The value was defined in JavaScript.
+    NonNull(T),
+}
+
+impl<T> Nullable<T> {
+    /// Returns `true` if this value is `Nullable::Null`.
```
HalidOdat:
```suggestion
    /// Returns `true` if this value is [`Nullable::Null`].
```

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.into_iter()
+    }
+
+    /// Converts from `&Nullable<T>` to `Nullable<&T>`.
+    ///
+    /// # Examples
+    /// ```
+    /// # use boa_engine::value::Nullable;
+    /// let text: Nullable<String> = Nullable::NonNull("Hello, world!".to_string());
+    /// // First, cast `Nullable<String>` to `Nullable<&String>` with `as_ref`,
```
HalidOdat:
```suggestion
    /// let text: Nullable<String> = Nullable::NonNull("Hello, world!".to_string());
    ///
    /// // First, cast `Nullable<String>` to `Nullable<&String>` with `as_ref`,
```

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,314 @@
...
+    assert_eq!(v.try_into_js(context).unwrap(), JsValue::null());
+}
+
+#[test]
+fn invalid() {
+    let context = &mut Context::default();
+    let v: JsResult<Nullable<i32>> = JsValue::undefined().try_js_into(context);
+
+    assert!(v.is_err());
+}
```
HalidOdat: consider putting this in a `#[cfg(test)]` `tests` module in-file or as a separate `tests.rs` module.

HalidOdat (approved):
Looks good just some minor nitpicky comment :smile:

core/engine/src/value/conversions/nullable.rs
hansl: `PartialEq` is already implemented for it, so I don't think `is` is useful. I was trying to add the equivalent of `is_some()` and `is_none()` from `Option<T>`.

nekevss: Ah, oops. I didn't mean to leave the `PartialEq` in there. I meant more so like [`Gc::is`](https://github.com/boa-dev/boa/blob/main/core/gc/src/pointers/gc.rs#L88).

Hmmmmm, maybe `is_non_null` if you were going for a `is_some`/`is_none`

core/engine/src/value/conversions/nullable.rs
HalidOdat, line 1: It's interesting to see the other way of declaring multi-file modules, but I think we should try to be consistent with the rest of the code base and move this to `nullable` and rename it to `mod.rs`

core/engine/src/value/conversions/nullable.rs
```diff
@@ -0,0 +1,283 @@
+//! A value that can be `null` in JavaScript, but not `undefined`.
+
+use crate::value::{TryFromJs, TryIntoJs};
+use boa_engine::{Context, JsResult, JsValue};
+
+mod tests;
```
HalidOdat:
```suggestion
#[cfg(test)]
mod tests;
```

I see that the `cfg` is in the test module... which is the first time I'm seeing that, didn't know you could do that :laughing: , but I think we should be consistent with other way we declare tests modules.

(assumption): It might be slightly faster for annotation to be outside, since we don't have to check in the module to eliminate the module from build.

HalidOdat (commented):
Looks good to me! Just some minor nitpicks :smile:

HalidOdat (approved):

PR #4327, opened 2025-07-08
Simplify SourcePositionGuard creation

hansl:
Add a Spanned trait for any types/elements of an AST to be able to generate a Span, and implement that for any types that are used in the source position backtrace.

Then added a method to simplify the creation of the source position. Since it's always created from a ByteCompiler it makes sense to have it as an associated method.

### DIFF
--- a/core/ast/src/expression/access.rs
+++ b/core/ast/src/expression/access.rs
@@ -14,10 +14,10 @@
 //! [spec]: https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-property-accessors
 //! [access]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors
 
-use crate::Span;
 use crate::expression::Expression;
 use crate::function::PrivateName;
 use crate::visitor::{VisitWith, Visitor, VisitorMut};
+use crate::{Span, Spanned};
 use boa_interner::{Interner, ToInternedString};
 use core::ops::ControlFlow;
 
@@ -45,6 +45,12 @@ impl PropertyAccessField {
             Self::Const(identifier) => identifier.span(),
             Self::Expr(expression) => expression.span(),
         }
+    }
+}
+
+impl Spanned for PropertyAccessField {
+    fn span(&self) -> Span {
+        self.span()
     }
 }
 
@@ -194,6 +200,12 @@ impl SimplePropertyAccess {
     #[must_use]
     pub fn span(&self) -> Span {
         Span::new(self.target.span().start(), self.field.span().end())
+    }
+}
+
+impl Spanned for SimplePropertyAccess {
+    fn span(&self) -> Span {
+        self.span()
     }
 }
 
--- a/core/ast/src/position.rs
+++ b/core/ast/src/position.rs
@@ -203,12 +203,38 @@ impl Span {
     }
 }
 
+impl From<Span> for Option<Position> {
+    fn from(value: Span) -> Self {
+        Some(value.start)
+    }
+}
+
+impl<'a> From<&'a Span> for Option<Position> {
+    fn from(value: &'a Span) -> Self {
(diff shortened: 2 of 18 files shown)


core/ast/src/position.rs
```diff
@@ -232,6 +258,12 @@ impl fmt::Display for Span {
     }
 }
 
+/// An element of the AST or any type that can be located in the source with a Span.
+pub trait Spanned {
+    /// Returns a span from the current type.
+    fn span(&self) -> Span;
+}
```
HalidOdat: I think we should remove direct `span()` method on the type, and use this trait to get the span. having two to span methods seems odd, it does have advantage of being able to be used without trait import, IMO this isn't a strong point.

core/engine/src/bytecompiler/mod.rs
```diff
@@ -416,7 +417,10 @@ pub(crate) struct SourcePositionGuard<'a, 'b> {
     compiler: &'a mut ByteCompiler<'b>,
 }
 impl<'a, 'b> SourcePositionGuard<'a, 'b> {
-    pub(crate) fn new(compiler: &'a mut ByteCompiler<'b>, position: Position) -> Self {
+    pub(crate) fn new(
+        compiler: &'a mut ByteCompiler<'b>,
+        position: impl Into<Option<Position>>,
```
HalidOdat:
```suggestion
        position: Option<Position>,
```

I don't think we need this to be a generic, it's internal and we don't construct it anywhere were we need this `impl`.

core/ast/src/position.rs
```diff
@@ -203,6 +203,18 @@ impl Span {
...
+        Some(value.start)
+    }
+}
+
+impl<'a> From<&'a Span> for Option<Position> {
+    fn from(value: &'a Span) -> Self {
+        Some(value.start)
+    }
+}
+
```
HalidOdat: Continuation of the previous comment on `impl` on `SourcePositionGuard`. We should be able to remove these, since they wouldn't be needed.

core/ast/src/position.rs
hansl: I was going to do that but didn't want to break any other code that might be using it. I don't mind (though I prefer keeping my PRs closely scoped), let me do it.

core/ast/src/expression/access.rs
```diff
@@ -99,11 +97,8 @@ pub enum PropertyAccess {
     Super(SuperPropertyAccess),
 }
 
-impl PropertyAccess {
-    /// Get the [`Span`] of the [`PropertyAccess`] node.
-    #[inline]
-    #[must_use]
-    pub fn span(&self) -> Span {
+impl Spanned for PropertyAccess {
+    fn span(&self) -> Span {
```
HalidOdat, line 102: We should try to mark then as `#[inline]` so the compiler can cross-crate inline the code.

core/engine/src/bytecompiler/mod.rs
```diff
@@ -788,6 +789,13 @@ impl<'ctx> ByteCompiler<'ctx> {
         self.bytecode.next_opcode_location()
     }
 
+    pub(crate) fn position_guard(
+        &mut self,
+        spanned: impl Spanned,
```
HalidOdat, line 794:
```suggestion
    pub(crate) fn position_guard<T: Spanned>(
        &mut self,
        spanned: &T,
```
We may not need the `impl<T: Spanned> Spanned for &T` at `core/ast/src/position.rs:222` in  if we take `T` by reference :)

hansl: Then we need to change all passed in values by reference instead. It's easier to accept both :)

core/ast/src/expression/access.rs
hansl: Yeah I don't like how Rust forces you to do that ....  I'll do a quick sweep.

PR #4328, opened 2025-07-10
Introduce async `ModuleLoader`s

jedel1043:
This PR changes `ModuleLoader` so that it an async trait instead of a trait that takes a closure that needs to be executed. This makes things much nicer for the composing module loaders, since they can compose nicely using async await syntax.

One small optimization was to use a `RefModuleLoader` for the composing loaders instead of only `ModuleLoader`. This is because composers like `FallbackModuleLoader` are much nicer if they can take their receiver as `Rc<Self>`, but their inner loaders must take `&Self` to avoid having something like
`FallbackModuleLoader<L,R>(Rc<L>,Rc<R>)`. This is just an optimization though, so we could fallback to only having a single trait if you think having unnecessary `Rc`s inside `Rc`s is a worth tradeoff for API simplicity.

Finally, a caveat is that I hid the `DynModuleLoader` from the API using `Any`, so getting the loader is now only possible if you know the full type of the loader:
```rust
context.module_loader() // private to the crate
context.downcast_module_loader::<SimpleModuleLoader>().unwrap() // public
```
This also allows removing the `register_module` and `get_module`, which were a bit clunky because many modules used different things to index their internal caches, and most of the loaders didn't have a concept of "registering" and "getting" modules. Though, I'm not opposed to adding them back if you think they're important to have.

cc @hansl since you were very interested in having something like this.


jedel1043:
Test changes were false positives; we haven't implemented source phase imports.

core/engine/src/module/mod.rs
```diff
@@ -10,11 +10,24 @@
...
+//! will automatically implement `ModuleLoader`, and viceversa. Depending on your use case,
+//! you would want to implement one or the other:
+//!
+//! - `ModuleLoader` takes `Rc<Self>` as receiver, which makes it better for loaders that need to
+//!   asynchronously modify their internal state, since it allows moving the loader into the future
+//!   returned by [`ModuleLoader::load_imported_module`].
+//! - `RefModuleLoader` takes `&Self` as receiver, which disallows moving the loader into the future
+//!   returned by [`RefModuleLoader::load_imported_module`]. This trait is ideal for loaders that mostly
+//!   do everything in a synchronous way, or don't need to modify any internal state.
+//!
```
hansl: I think, for the sake of saving a lot of questions and maintenance, that we should pick one of those 2 and stick with it. My vote would go for `Rc<Self>`. It's self explanatory, not really unusual for async code, and will be the most future proof one.

I don't think supporting both will do us any favors. It's just another edge case to support and test for, and it doesn't provide added value. The number of times people implement their own `ModuleLoader` is minimal at best anyway (they should use one of our provided implementation). Unless there is a clear reason (e.g. performance) to keep `RefModuleLoader`, I'd say we should not introduce it.

PS: I've reviewed most of the rest, which is fine, but it highly depends on this decision as well.

jedel1043: I think this is fair. We can still optimize this in the future with some kind of "projectable" Rc pointer when the Rust team stabilizes custom receivers.

EDIT: [arbitrary_self_types](https://github.com/rust-lang/rust/issues/44874) was the correct feature

jedel1043: Rolled back to the unoptimized but simpler version.

hansl (approved):
LGTM.

hansl:
I'll make a mental note to improve on the module loaders as I move them to `boa_engine`. I think we can simplify them a lot more now (and compose them better). But that work is out of scope for this PR.

PR #4331, opened 2025-07-11
Make `JobExecutor::run_jobs_async` a plain async method

jedel1043:
This PR makes `JobExecutor::run_jobs_async` a proper async function, which greatly simplifies our API by removing all the ugly lifetimes.
The change that enabled this was to make the trait's receiver a `Rc<Self>`, which can be moved inside the future returned by the function.

In contrast to #4328, we don't really require async trait objects in this case, since the only thing that the engine does is calling `enqueue_job` and `run_jobs`, which are plain old methods. Thus, I excluded the `run_jobs_async` method from the trait's VTable with a `where Self: Sized` bound, but the method can still be called by creating the `Rc<Queue>` executor first, then passing a clone to the `ContextBuilder`.

### DIFF
--- a/examples/src/bin/module_fetch_async.rs
+++ b/examples/src/bin/module_fetch_async.rs
@@ -235,18 +235,13 @@ impl JobExecutor for Queue {
             // We have some jobs pending on the microtask queue. Try to poll the pending
             // tasks once to see if any of them finished, and run the pending microtasks
             // otherwise.
-            let Some(result) = future::poll_once(group.next()).await.flatten() else {
-                // No completed jobs. Run the microtask queue once.
-                self.drain_jobs(&mut context.borrow_mut());
-                continue;
-            };
-
-            if let Err(err) = result {
+            if let Some(Err(err)) = future::poll_once(group.next()).await.flatten() {
                 eprintln!("Uncaught {err}");
-            }
+            };
 
             // Only one macrotask can be executed before the next drain of the microtask queue.
             self.drain_jobs(&mut context.borrow_mut());
+            future::yield_now().await
         }
     }
 }
--- a/examples/src/bin/smol_event_loop.rs
+++ b/examples/src/bin/smol_event_loop.rs
@@ -123,21 +123,15 @@ impl JobExecutor for Queue {
                 continue;
             }
 
-            // We have some jobs pending on the microtask queue. Try to poll the pending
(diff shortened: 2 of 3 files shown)


examples/src/bin/module_fetch_async.rs
```diff
@@ -204,58 +204,49 @@ impl JobExecutor for Queue {
...
+            };
+
+            if let Err(err) = result {
+                eprintln!("Uncaught {err}");
             }
-        })
+
+            // Only one macrotask can be executed before the next drain of the microtask queue.
+            self.drain_jobs(&mut context.borrow_mut());
+        }
```
hansl: GitHub don't let me apply a suggestion on this, but I think you can simplify:

```rust
if let Some(Err(err)) = future::poll_once(group.next()).await.flatten() {
  eprintln!("Unchaught {err}");
}

// Only one macrotask can be executed before the next drain of the microtask queue.
self.drain_jobs(&mut context.borrow_mut());
```

examples/src/bin/smol_event_loop.rs
```diff
@@ -96,59 +95,50 @@ impl JobExecutor for Queue {
...
                 if let Err(err) = result {
                     eprintln!("Uncaught {err}");
                 }
+                continue;
+            }
 
-                // Only one macrotask can be executed before the next drain of the microtask queue.
+            // We have some jobs pending on the microtask queue. Try to poll the pending
+            // tasks once to see if any of them finished, and run the pending microtasks
+            // otherwise.
```
hansl: See above.

examples/src/bin/tokio_event_loop.rs
```diff
@@ -106,57 +105,48 @@ impl JobExecutor for Queue {
...
                 }
 
-                // Only one macrotask can be executed before the next drain of the microtask queue.
+                continue;
+            }
+
+            // We have some jobs pending on the microtask queue. Try to poll the pending
+            // tasks once to see if any of them finished, and run the pending microtasks
+            // otherwise.
+            let Some(result) = future::poll_once(group.next()).await.flatten() else {
```
hansl: See above.

examples/src/bin/module_fetch_async.rs
jedel1043: Simplified.

PR #4338, opened 2025-07-15
`Fetch` API (beta)

hansl:
Introduce an implementation of `fetch` ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)) to `boa_runtime`. It is optional and covers many basic use cases but not all.

Other house keeping and yak shaving for this PR:

- Some fixes and improvements
- The new `fetch` is added to boa's CLI, using a blocking `reqwest` client. We publish a blocking client implementation of `Fetcher` here to allow for non-async applications to use it. A non-blocking client should be added.
- Broaden the implementation of `TryIntoJsResult` to also include any types that can be `TryIntoJs` already. Before it was limited to `Into<JsValue>`.
- Add missing implementations of `TryIntoJs` for `JsPromise`, any `T: Class + Clone` (since it's a trivial implementation), and `isize`/`usize`.
- Update `JsPromise::from_future` to include an `AsyncFnOnce` that receives a context.
- Add `Nullable::unwrap_or_default()`, which is used in this PR for simplicity.
- Add an implementation of `Trace` and `Finalize` for `Either` (conditional on existing feature)
- Fix the interface of `clearTimeout` to allow for nulls (now that `Option<>` only supports `undefined`). This fixes a couple WPT.
- Clarify how to initialize `boa_runtime` now that it's becoming more complex.
- Add support for `fetch` to WPT tests. Unfortunately, it's basically failing everything and requires a webserver to run locally, so WPT will not run on CI just yet and all fetch tests are ignored for now.
- Add support for generics in `boa_module`.
- Add support for a `boa_register()` function on `boa_modules` for those who want to register the module's content on the context/realm.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -15,7 +15,7 @@ rust-version.workspace = true
 boa_engine = { workspace = true, features = ["deser", "flowgraph", "trace"] }
 boa_parser.workspace = true
 boa_gc.workspace = true
-boa_runtime = { workspace = true }
+boa_runtime.workspace = true
 rustyline = { workspace = true, features = ["derive", "with-file-history"] }
 clap = { workspace = true, features = ["derive"] }
 serde_json.workspace = true


core/engine/src/native_function/mod.rs
```diff
@@ -178,6 +178,7 @@ impl NativeFunction {
     /// Certain async functions need to be desugared for them to be compatible. For example, the
     /// following won't compile:
     ///
+    /// TODO: this does not fail anymore.
```
jedel1043: Uhh I'm confused by this because it is marked as `compile_fail` and the test seems to pass (fail to compile) on the CI...

jedel1043: Ohhh, it succeeds because our imports have a missing comma lol

core/engine/src/object/builtins/jspromise.rs
```diff
@@ -267,34 +268,35 @@ impl JsPromise {
...
+    /// let promise = JsPromise::from_future(f, context);
     ///
     /// context.run_jobs();
     ///
     /// assert_eq!(promise.state(), PromiseState::Fulfilled(JsValue::null()));
     /// ```
     ///
     /// [async_fn]: crate::native_function::NativeFunction::from_async_fn
-    pub fn from_future<Fut>(future: Fut, context: &mut Context) -> Self
+    pub fn from_future<F>(f: F, context: &mut Context) -> Self
```
jedel1043: Would probably be good to rename the function, since it's not taking a `future` anymore.

jedel1043: Actually, now that I think about it, do we want to rewrite this function, or would it be better to add a new `from_async_closure` function instead? Seems useful to have the more straightforward `from_future` for when you have a Rust future that you need to resolve which doesn't depend on taking the context.

core/engine/src/value/conversions/nullable/mod.rs
```diff
@@ -135,6 +135,17 @@ impl<T> Nullable<T> {
 
     /// Maps a [`Nullable<T>`] to [`Nullable<U>`] by applying a function to a contained
     /// value.
+    ///
+    /// ```
```
jedel1043, line 140:
```suggestion
    /// # Examples
    ///
    /// ```
```

core/runtime/src/fetch/headers.rs
```diff
@@ -0,0 +1,272 @@
...
+        .as_ref()
+        .parse()
+        .map_err(|_| js_error!("Cannot convert value to header string as it is not valid ASCII."))
+}
+
+/// A JavaScript wrapper for the `Headers` object.
+#[derive(Debug, Default, Clone, JsData, Trace, Finalize)]
+pub struct JsHeaders {
+    #[unsafe_ignore_trace]
+    headers: Rc<RefCell<HttpHeaderMap>>,
```
jedel1043, line 49: Why do you need `Rc<RefCell<>>` here? The outer `JsObject` is already wrapping a `Gc<GcRefCell<>`, so you should be able to get the data from that.

core/runtime/src/fetch/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    // Try fetching from the context first, then the current realm. Else fail.
+    let Some(fetcher) = context.get_data::<FetcherRc<T>>().cloned().or_else(|| {
+        context
+            .realm()
+            .host_defined()
+            .get::<FetcherRc<T>>()
+            .cloned()
+    }) else {
+        return Err(
+            js_error!(Error: "Implementation of fetch requires a fetcher registered in the context"),
```
jedel1043, line 85: Thought: throwing `Error` here is fine IMO, but maybe we could add some kind of custom `RuntimeError` specifically for errors coming from `boa_runtime` in the future.

core/runtime/src/fetch/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+pub fn fetch<T: Fetcher>(
+    resource: Either<JsString, JsObject>,
+    options: Option<RequestInit>,
+    context: &mut Context,
+) -> JsPromise {
+    JsPromise::from_future(
+        async move |context| fetch_inner::<T>(resource, options, context).await,
+        context,
+    )
+}
```
jedel1043: Thought: if we added some kind of `IntoAsyncJsFunction` trait, this would be much simpler; just an async function that calls `fetch_inner`. Could be a nice future improvement.

core/runtime/src/fetch/request.rs
```diff
@@ -0,0 +1,223 @@
...
+) -> JsResult<http::request::Builder> {
+    add_headers_to_builder(headers.iter().map(|(k, v)| (k, v)), builder)
+}
+
+fn add_btree_headers_to_builder(
+    headers: &BTreeMap<JsString, Convert<JsString>>,
+    builder: http::request::Builder,
+) -> JsResult<http::request::Builder> {
+    add_headers_to_builder(headers.iter(), builder)
+}
```
jedel1043: I don't really see the benefit of having this. The underlying calls to `headers.iter()` are so simple that they should be done inline.

core/runtime/src/fetch/request.rs
```diff
@@ -0,0 +1,223 @@
...
+                }
+                Either::Right(headers) => {
+                    builder = add_btree_headers_to_builder(headers, builder)?;
+                }
+            }
+        }
+
+        if let Some(Convert(ref method)) = self.method.take() {
+            builder = builder.method(method.to_std_string().map_err(
+                |_| js_error!(TypeError: "Requestion constructor: {} is an invalid method", method.to_std_string_escaped()),
```
jedel1043:
```suggestion
                |_| js_error!(TypeError: "Request constructor: {} is an invalid method", method.to_std_string_escaped()),
```

core/runtime/src/fetch/response.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    #[unsafe_ignore_trace]
+    r#type: ResponseType,
+
+    #[unsafe_ignore_trace]
+    status: Option<StatusCode>,
+
+    headers: JsHeaders,
+
+    body: Gc<Vec<u8>>,
```
jedel1043: Since this doesn't have any `Gc` pointers in it, it would be better to use `Rc` instead.

core/runtime/src/lib.rs
```diff
@@ -38,6 +36,55 @@
...
+//!             res.to_string(&mut context).unwrap().to_std_string_escaped()
+//!         );
+//!     }
+//!     Err(e) => {
+//!         // Pretty print the error
+//!         eprintln!("Uncaught {e}");
+//!         # panic!("An error occured in boa_runtime's js_code");
+//!     }
+//! };
+//! ```
```
jedel1043: Loving the new documentation here!

core/runtime/src/lib.rs
```diff
@@ -63,55 +110,34 @@ mod text;
...
-    }
-}
+pub use options::RegisterOptions;
 
 /// Register all the built-in objects and functions of the `WebAPI` runtime.
 ///
 /// # Errors
 /// This will error is any of the built-in objects or functions cannot be registered.
-pub fn register(
+pub fn register<#[cfg(feature = "fetch")] F: fetch::Fetcher, L: Logger + 'static>(
```
jedel1043: I don't think it's a good pattern to change the generics of a function depending on the enabled features. I think it's much better to gate the implementors of `Fetcher` themselves, then leave the `ErrorFetcher` as the default if the feature is disabled.

tests/wpt/src/fetcher/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::cell::RefCell;
+use std::path::PathBuf;
+use std::rc::Rc;
+use url::Url;
+
+#[derive(Debug, Clone, Trace, Finalize, JsData)]
+pub struct WptFetcher {
+    wpt_server: String,
+    wpt_root: PathBuf,
+    current_file: Gc<GcRefCell<Option<PathBuf>>>,
```
jedel1043: This doesn't require any `Gc`; all the types involved are not `Gc`-related.

core/engine/src/native_function/mod.rs
hansl: Yeah, I noticed it from my IDE and didn't have time to look deeper into it. I'll revert this.

core/runtime/src/fetch/headers.rs
hansl: `for_each` clones the actual Rust object, and that can be expensive.

core/runtime/src/fetch/mod.rs
hansl: Great suggestion, but I'd like to do this in a separate PR.

core/runtime/src/fetch/headers.rs
hansl: Actually that might not be necessary. Let me see.

core/engine/src/object/builtins/jspromise.rs
hansl: I renamed it `from_async_fn`. We are not using the old function in our code so I don't know how much we need it. I'll keep it as is and if there really is a need we can add it later. WDYT?

core/runtime/src/fetch/mod.rs
hansl: I like it. Would also make creating `async` members in `boa_class` and `boa_module` easier. Created https://github.com/boa-dev/boa/issues/4345

core/runtime/src/fetch/request.rs
hansl: Yeah they diverged at some point and I didn't finish the cleanup. Done.

core/runtime/src/lib.rs
hansl: For this I'll probably sync with you on Elements as I think I'm in the right here (though the actual code could likely improve).

What I want out of this feature is to remove ALL the code you don't need, which is a surprising amount. If the feature doesn't reduce the code compiled then we should just include fetch all the time.

The trait is dependent on `JsResponse`, `JsRequest`, `JsHeaders` etc. Basically all except the actual `fetch` function. That's a lot of code to include when the feature is disabled. So I'd like to remove all mentions of `fetch` and co from the API when the feature is not specified.

core/engine/src/object/builtins/jspromise.rs
jedel1043: Sounds reasonable.

core/runtime/src/lib.rs
hansl: Okay let me know what you think of the new API.

hansl:
@jedel1043 PTAL.

core/engine/src/value/conversions/try_from_js.rs
```diff
@@ -95,9 +97,13 @@ where
                 .into());
         };
 
-        let length = object
-            .get(js_string!("length"), context)?
-            .to_length(context)?;
+        let length = object.get(js_string!("length"), context)?;
```
HalidOdat: Suggestion: Consider using `StaticJsStrings::LENGTH` constant and prevents us from accidentally mistyping it. :smile: 
```suggestion
        let length = object.get(StaticJsStrings::LENGTH, context)?;
```

core/engine/src/value/conversions/try_into_js.rs
```diff
@@ -96,6 +106,27 @@ impl TryIntoJs for u64 {
         }
     }
 }
+impl TryIntoJs for isize {
+    fn try_into_js(&self, _context: &mut Context) -> JsResult<JsValue> {
+        let value = *self;
+        #[allow(clippy::manual_range_contains)]
+        if value < MIN_SAFE_INTEGER_I64 as isize || (MAX_SAFE_INTEGER_I64 as isize) < value {
```
HalidOdat: Question: is there a reason for ignoring the lint?
Remark: If that's the case then consider adding a small note on it :)

core/gc/src/trace.rs
```diff
@@ -549,3 +549,20 @@ mod boa_string_trace {
...
+
+    impl<L: Trace, R: Trace> Finalize for either::Either<L, R> {}
+
+    unsafe impl<L: Trace, R: Trace> Trace for either::Either<L, R> {
+        custom_trace!(this, mark, {
+            if let either::Either::Left(l) = this {
+                mark(l);
+            } else if let either::Either::Right(r) = this {
+                mark(r);
+            }
```
HalidOdat: Question: Why not use `match`? Has the advantage of being a compilation error if a variant is added.

Not that it matters in this case since it's only really going to have two variants. But it's nice to be consistent with other enums that we have.

cli/src/main.rs
```diff
@@ -454,8 +454,17 @@ fn main() -> Result<()> {
...
 fn add_runtime(context: &mut Context) {
-    boa_runtime::register(context, boa_runtime::RegisterOptions::new())
-        .expect("should not fail while registering the runtime");
+    boa_runtime::register(
+        (
+            boa_runtime::extensions::ConsoleExtension::default(),
+            boa_runtime::extensions::FetchExtension(
+                boa_runtime::fetch::BlockingReqwestFetcher::default(),
+            ),
+        ),
```
HalidOdat: Suggestion: Maybe a `&[Box<dyn RuntimeExtension>]`, this has the disadvantage of having to require `Box<T>` (only allocates for non-ZST types). But the bigger advantage of removing the tuple implementation, which allows to dynamically add/remove extensions.

core/runtime/src/extensions.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct FetchExtension<F: crate::fetch::Fetcher>(pub F);
+
+#[cfg(feature = "fetch")]
+impl<F: crate::fetch::Fetcher + Debug + 'static> RuntimeExtension for FetchExtension<F> {
+    fn register(self, realm: Option<Realm>, context: &mut Context) -> JsResult<()> {
+        crate::fetch::register(self.0, realm, context)
+    }
+}
+
+macro_rules! decl_runtime_ext_tuple {
```
HalidOdat, line 85: Suggestion(continuation): The `&[Box<dyn RuntimeExtension>]` would remove this :smile:

core/runtime/Cargo.toml
```diff
@@ -28,6 +33,9 @@ workspace = true
...
 
 [features]
-default = ["all"]
-all = ["url"]
+# Default should not add `reqwest` as it is not available on all platforms.
+default = ["fetch", "url"]
+all = ["default", "reqwest"]
 url = ["dep:url"]
+fetch = ["dep:futures-lite", "dep:http", "dep:serde_json", "boa_engine/either"]
+reqwest = ["dep:reqwest", "reqwest/blocking"]
```
HalidOdat: Question: Maybe this should be renamed to `reqwest-blocking`?

tests/wpt/Cargo.toml
```diff
@@ -9,7 +9,9 @@ boa_engine = { path = "../../core/engine" }
 boa_gc = { path = "../../core/gc" }
 boa_interop = { path = "../../core/interop" }
 boa_runtime = { path = "../../core/runtime" }
+http = "1.3.1"
 rstest = "0.25.0"
+url = { version = "2.5.4", features = [] }
```
HalidOdat: Consider putting under workspace dependencies
```suggestion
http.workspace = true
rstest.workspace = true
url.workspace = true
```

tests/wpt/src/fetcher/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+use boa_runtime::fetch::response::JsResponse;
+use boa_runtime::fetch::BlockingReqwestFetcher;
+use boa_runtime::fetch::Fetcher;
+use std::cell::RefCell;
+use std::path::PathBuf;
+use std::rc::Rc;
+use url::Url;
+
+#[derive(Debug, Clone, Trace, Finalize, JsData)]
+pub struct WptFetcher {
```
HalidOdat, line 17: remark: Is the WPT fetcher specified somewhere in the wpt repository, if so consider adding a link to it :smile:

core/runtime/src/fetch/request.rs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+
+        builder
+            .body(request_body.unwrap_or_default())
+            .map_err(|_| js_error!(Error: "Cannot construct request"))
+    }
+}
+
+/// The JavaScript Response class.
```
HalidOdat:
```suggestion
/// The JavaScript `Request` class.
```

core/runtime/src/fetch/request.rs
```diff
@@ -0,0 +1,209 @@
...
+
+impl RequestInit {
+    /// Create an [`http::request::Builder`] object and return both the
+    /// body specified by JavaScript and the builder.
+    ///
+    /// # Errors
+    /// If the body is not a valid type, an error is returned.
+    pub fn into_request_builder(
+        mut self,
+        maybe_request: Option<HttpRequest<Vec<u8>>>,
```
HalidOdat: nitpick:
```suggestion
        request: Option<HttpRequest<Vec<u8>>>,
```

core/runtime/src/fetch/request.rs
```diff
@@ -0,0 +1,209 @@
...
+    ) -> JsResult<HttpRequest<Vec<u8>>> {
+        let mut builder = HttpRequest::builder();
+        if let Some(r) = maybe_request {
+            let (parts, _body) = r.into_parts();
```
HalidOdat: nitpick:
```suggestion
        if let Some(request) = maybe_request {
            let (parts, _body) = request.into_parts();
```

core/runtime/src/fetch/headers.rs
```diff
@@ -0,0 +1,272 @@
...
+        this: JsClass<Self>,
+        callback: ForEachCallback,
+        this_arg: Option<JsValue>,
+        context: &mut Context,
+    ) -> JsResult<()> {
+        let object = this.inner().upcast();
+        let this_arg = this_arg.unwrap_or_default();
+        for (k, v) in this.clone_inner().headers.borrow().iter() {
+            let k = JsString::from(k.as_str());
+            let v = JsString::from(v.to_str().unwrap_or(""));
```
HalidOdat, line 190:
```suggestion
            let v = JsString::from(v.to_str().unwrap_or_default());
```

core/runtime/Cargo.toml
```diff
@@ -14,7 +14,12 @@ rust-version.workspace = true
 boa_engine.workspace = true
 boa_gc.workspace = true
 boa_interop.workspace = true
+either.workspace = true
+futures-lite = { workspace = true, optional = true }
+http = { version = "1.3.1", optional = true }
+reqwest = { version = "0.12.20", optional = true }
```
HalidOdat: Nitpick: Consider putting theses in workspace dependencies, so we have one place where all versions are defined, avoiding potential compilation issues related to different versions
```suggestion
http = { version = "1.3.1", optional = true }
reqwest = { version = "0.12.20", optional = true }
```

cli/src/main.rs
```diff
@@ -454,8 +454,17 @@ fn main() -> Result<()> {
...
 /// Adds the CLI runtime to the context with default options.
 fn add_runtime(context: &mut Context) {
-    boa_runtime::register(context, boa_runtime::RegisterOptions::new())
-        .expect("should not fail while registering the runtime");
+    boa_runtime::register(
+        (
+            boa_runtime::extensions::ConsoleExtension::default(),
+            boa_runtime::extensions::FetchExtension(
+                boa_runtime::fetch::BlockingReqwestFetcher::default(),
+            ),
```
HalidOdat, line 463: Important: The CLI uses `BlockingReqwestFetcher` and fails if the default feature are disabled (see build CI), this should be featured flagged that enables the `boa_runtime/fetch` and `boa_runtime/reqwest`
```suggestion
            boa_runtime::extensions::FetchExtension(
                boa_runtime::fetch::BlockingReqwestFetcher::default(),
            ),
```

core/engine/src/value/conversions/try_into_js.rs
hansl: That's just copied code from the `i64` implementation, I didn't actually look into it. I'll fix both.

cli/src/main.rs
hansl: I can add it too. I was thinking about this today and I was thinking that it'd be neat if we could call register multiple times. Let me try a small refactor.

core/runtime/src/extensions.rs
hansl: I'd like to keep the tuple as well, as it's simpler when you don't need conditions.

tests/wpt/Cargo.toml
hansl: It's not part of the workspace.

tests/wpt/src/fetcher/mod.rs
hansl: No, it uses the browser's fetch implementation, which for example understands relative paths to the script (which is distributed through the local webserver). This is the only difference between `WptFetcher` and `BlockingReqwestFetcher`.

I haven't looked too deeply for fetch, I think `NodeJS` has some WPT runner too, I should look how they resolve stuff.

core/runtime/src/fetch/request.rs
hansl: Sorry, old Haskell habits ^_^

cli/src/main.rs
hansl: Yeah added a `fetch` feature to cli that is by default but removing the default features `fetch` won't be defined.

tests/wpt/src/fetcher/mod.rs
hansl: The WPT is a strong mix of HTML, JavaScript and Python webserver and scripts... It probably will not run in our CI anytime soon :(

cli/src/main.rs
hansl: Actually you can just call `register()` on the extension itself...

```rust
boa_runtime::fetch::BlockingReqwestFetcher::default().register(None, context)
  .expect("Could not register stuff");
```

cli/Cargo.toml
```diff
@@ -15,7 +15,7 @@ rust-version.workspace = true
 boa_engine = { workspace = true, features = ["deser", "flowgraph", "trace"] }
 boa_parser.workspace = true
 boa_gc.workspace = true
-boa_runtime.workspace = true
+boa_runtime = { workspace = true }
```
HalidOdat: nit:
```suggestion
boa_runtime.workspace = true
```

PR #4339, opened 2025-07-15
Implement `Atomics.waitAsync`

jedel1043:
Phew, this was a bit of an adventure.

This implements `Atomics.waitAsync`, which is the last missing synchronization primitive we needed to implement have 100% conformance on that builtin. The implementation is not the most optimal one; we could definitely use some channels to communicate which waiters were notified instead of having to iterate through all of them, but IMO it looks fairly decent for a first pass.

### DIFF
--- a/tests/tester/src/exec/js262.rs
+++ b/tests/tester/src/exec/js262.rs
@@ -69,10 +69,14 @@ impl Drop for WorkerHandles {
 }
 
 /// Creates the object $262 in the context.
-pub(super) fn register_js262(handles: WorkerHandles, context: &mut Context) -> JsObject {
+pub(super) fn register_js262(
+    handles: WorkerHandles,
+    console: bool,
+    context: &mut Context,
+) -> JsObject {
     let global_obj = context.global_object();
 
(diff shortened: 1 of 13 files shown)


tests/tester/src/exec/js262.rs
```diff
@@ -69,10 +69,14 @@ impl Drop for WorkerHandles {
 }
 
 /// Creates the object $262 in the context.
-pub(super) fn register_js262(handles: WorkerHandles, context: &mut Context) -> JsObject {
+pub(super) fn register_js262(
+    handles: WorkerHandles,
+    console: bool,
```
nekevss, line 74: question: was the console update related to the `Atomics.waitAsync` implementation?

jedel1043: Kinda. I had a couple of bugs in the implementation that were very tricky to debug from the main thread, so I had to enable the console in the worker threads. I left it here because it may be useful to debug other multi-threaded tests in the future.

jedel1043:
Will rebase this today so that we can merge

PR #4342, opened 2025-07-17
Remove `Box` from `Object`'s data member

HalidOdat:
This PR implements removes the `Box` from `Object`'s data member (by implementing this PR comment https://github.com/boa-dev/boa/pull/4287#issuecomment-2993682209).

The main idea is to limit the alignment, for the min we use a wrapper struct (`ObjectData<T>`) that has an minimum alignment using the `#[align(N)]`, and for the max alignment we use a constant assertion, preventing compilation with a message on how to workaround it.

### DIFF
--- a/core/engine/src/object/mod.rs
+++ b/core/engine/src/object/mod.rs
@@ -161,6 +161,42 @@ impl dyn NativeObject {
     }
 }
 
+// Force alignment to be the same for all JsData types.
+#[derive(Debug, Finalize, Trace)]
+// SAFETY: This does not implement drop, so this is safe.
+#[boa_gc(unsafe_no_drop)]
+#[repr(C, align(8))]
+struct ObjectData<T: ?Sized> {
+    // MUST BE PRIVATE, should not be constructed directly. i.e. { data: ... }
+    // Because we want to trigger the assertion below.
+    //
+    // It is fine if we have deref/deref_mut to it or any access.
+    data: T,
+}
+
+impl<T: Default> Default for ObjectData<T> {
+    #[inline]
+    fn default() -> Self {
+        Self::new(T::default())
+    }
+}
+
+static_assertions::const_assert!(align_of::<Box<()>>() <= 8);
+
+impl<T> ObjectData<T> {
+    const OBJECT_DATA_ALIGNMENT_REQUIREMENT: () = assert!(
+        align_of::<T>() <= 8,
+        "Alignment of JsData must be <= 8, NOTE: you can wrap the data in a Box<T>."
+    );
+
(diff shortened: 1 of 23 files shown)


HalidOdat:
## Benchmarks

### main

```
PROGRESS Richards
RESULT Richards 184
PROGRESS DeltaBlue
RESULT DeltaBlue 189
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 145
PROGRESS RayTrace
RESULT RayTrace 398
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 480
PROGRESS RegExp
RESULT RegExp 71.0
PROGRESS Splay
RESULT Splay 738
PROGRESS NavierStokes
RESULT NavierStokes 339
SCORE 254
```

### PR

```
PROGRESS Richards
RESULT Richards 190
PROGRESS DeltaBlue
RESULT DeltaBlue 193
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 149
PROGRESS RayTrace
RESULT RayTrace 402
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 475
PROGRESS RegExp
RESULT RegExp 72.6
PROGRESS Splay
RESULT Splay 741
PROGRESS NavierStokes
RESULT NavierStokes 340
SCORE 257
```

core/engine/src/object/mod.rs
```diff
@@ -173,7 +209,7 @@ pub struct Object<T: ?Sized> {
     /// The `[[PrivateElements]]` internal slot.
     private_elements: ThinVec<(PrivateName, PrivateElement)>,
     /// The inner object data
-    pub(crate) data: Box<T>,
+    data: ObjectData<T>,
```
jedel1043, line 181: `Object` probably requires `#[repr(C)]`, right? Otherwise, the compiler could reorder a field after `ObjectData`, which would cause UB.

HalidOdat: Yes, actually it should have `#[repr(C)]` even before this change. (AFAIK) there is no guarantee for `Box`-ed implementation either (but probably unlikely that it will reorder, since it has same size and alignment)...

EDIT: Will create a PR to add that just in case :smile: 
EDIT: https://github.com/boa-dev/boa/pull/4343

jedel1043 (approved):
Looks good!

By the way, did you try to run some tests with Miri? It shouldn't throw errors now after we fixed the UB in JsValue

HalidOdat:
> By the way, did you try to run some tests with Miri? It shouldn't throw errors now after we fixed the UB in JsValue

Just ran some tests and miri seems to be happy :smile:

PR #4346, opened 2025-07-21
Fix UB in implementation of `NanBoxedValue`

jedel1043:
- Changes the representation of `NanBoxedValue` to be a `*const ()` (plus an extra `u32` for 32-bit platforms).
- Removes usages of `Box` in `NanBoxedValue` for `JsSymbol` and `JsBigInt`.

The previous implementation technically was UB, because we didn't carry the provenance of the nan-boxed pointers.

### DIFF
--- a/core/engine/src/value/inner/nan_boxed.rs
+++ b/core/engine/src/value/inner/nan_boxed.rs
@@ -1,6 +1,6 @@
 //! A NaN-boxed inner value for JavaScript values.
 //!
-//! This `[JsValue]` is a float using `NaN` values to represent inner
+//! This [`JsValue`] is a float using `NaN` values to represent inner
 //! JavaScript value.
 //!
 //! # Assumptions
@@ -48,7 +48,7 @@
 //!
 //! # Design
 //!
-//! This `[JsValue]` inner type is a NaN-boxed value, which is a 64-bits value
+//! This [`JsValue`] inner type is a NaN-boxed value, which is a 64-bits value
 //! that can represent any JavaScript value. If the integer is a non-NaN value,
 //! it will be stored as a 64-bits float. If it is a `f64::NAN` value, it will
 //! be stored as a quiet `NaN` value. Subnormal numbers are regular float.
@@ -106,27 +106,39 @@
 //! with regular NAN should happen.
 #![allow(clippy::inline_always)]
 
-use crate::{JsBigInt, JsObject, JsSymbol, JsVariant};
-use boa_gc::{Finalize, Trace, custom_trace};
-use boa_string::JsString;
+use crate::{
+    JsBigInt, JsObject, JsSymbol, JsVariant, bigint::RawBigInt, object::ErasedVTableObject,
+    symbol::RawJsSymbol,
+};
+use boa_gc::{Finalize, GcBox, Trace, custom_trace};
+use boa_string::{JsString, RawJsString};
 use core::fmt;
 use static_assertions::const_assert;
+use std::{
+    mem::ManuallyDrop,
+    ptr::{self, NonNull},
+};
 
-// We cannot NaN-box pointers larger than 64 bits.
-const_assert!(size_of::<usize>() <= size_of::<u64>());
+const _NAN_BOX_COMPAT_CHECK: () = const {
+    // We can only NaN-box pointers that are 32 or 64 bits.
+    assert!(
+        size_of::<usize>() == size_of::<u64>() || size_of::<usize>() == size_of::<u32>(),
+        "this platform is not compatible with a nan-boxed `JsValueInner`\n\
+        enable the `jsvalue-enum` feature to use the enum-based `JsValueInner`"
+    );
 
-// We cannot NaN-box pointers that are not 4-bytes aligned.
-const_assert!(align_of::<*mut ()>() >= 4);
+    // We cannot NaN-box pointers that are not 4-bytes aligned.
+    assert!(
+        align_of::<*mut ()>() >= 4,
+        "this platform is not compatible with a nan-boxed `JsValueInner`\n\
+        enable the `jsvalue-enum` feature to use the enum-based `JsValueInner`"
+    );
+};
 
 /// Internal module for bit masks and constants.
 ///
 /// All bit magic is done here.
 mod bits {
-    use crate::object::ErasedVTableObject;
-    use boa_engine::{JsBigInt, JsObject, JsSymbol};
-    use boa_gc::GcBox;
-    use boa_string::{JsString, RawJsString};
-    use std::ptr::NonNull;
 
     /// The mask for the bits that indicate if the value is a NaN-value.
     const MASK_NAN: u64 = 0x7FF0_0000_0000_0000;
@@ -183,18 +195,6 @@ mod bits {
         (value & MASK_NAN != MASK_NAN)
             || (value & MASK_KIND) == (MASK_NAN | TAG_INF)
             || (value & MASK_KIND) == (MASK_NAN | TAG_NAN)
-    }
-
-    /// Checks that a value is a valid undefined.
-    #[inline(always)]
-    pub(super) const fn is_undefined(value: u64) -> bool {
-        value == VALUE_UNDEFINED
-    }
-
-    /// Checks that a value is a valid null.
-    #[inline(always)]
-    pub(super) const fn is_null(value: u64) -> bool {
-        value == VALUE_NULL
     }
 
     /// Checks that a value is a valid integer32.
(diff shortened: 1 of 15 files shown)


hansl:
Could you provide benchmark results?

I'll review tomorrow.

jedel1043:
# PR

```
PROGRESS Richards
RESULT Richards 239
PROGRESS DeltaBlue
RESULT DeltaBlue 244
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 193
PROGRESS RayTrace
RESULT RayTrace 526
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 625
PROGRESS RegExp
RESULT RegExp 88.9
PROGRESS Splay
RESULT Splay 943
PROGRESS NavierStokes
RESULT NavierStokes 443
SCORE 329
```

# main

```
PROGRESS Richards
RESULT Richards 241
PROGRESS DeltaBlue
RESULT DeltaBlue 246
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 201
PROGRESS RayTrace
RESULT RayTrace 539
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 639
PROGRESS RegExp
RESULT RegExp 89.0
PROGRESS Splay
RESULT Splay 952
PROGRESS NavierStokes
RESULT NavierStokes 468
SCORE 336
```

Kind of a small perf regression, but it makes sense since we were exploiting UB to get more perf. I'll try to profile it to see if we can shave off some additional time though.

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -262,181 +262,56 @@ mod bits {
...
-    /// result in a panic. Also, the object is not checked for validity.
-    ///
-    /// The box is forgotten after this operation. It must be dropped separately,
-    /// by calling `[Self::drop_pointer]`.
-    #[inline(always)]
-    #[allow(clippy::identity_op)]
-    pub(super) unsafe fn tag_bigint(value: Box<JsBigInt>) -> u64 {
-        let value = Box::into_raw(value) as u64;
+    fn tag_pointer(value: *mut ()) -> u64 {
+        let value = value.addr() as u64;
```
hansl: Reading on https://doc.rust-lang.org/std/ptr/index.html#using-strict-provenance (and the section after, Exposed Provenance), it seems that it is not UB to cast from a pointer to `usize` using `addr`, `with_addr`, `expose_provenance` and `with_exposed_provenance`.

Should we try to use those instead? It seems that would keep the same performance (and minimal changes) while resolving the UB itself. Also, I have no idea how this resolves on 32-bits pointer platforms and AFAIK our CI does not test on 32-bits.

hansl: BTW I cannot seem to be able to run Miri because of UB in strings. I'll look closer tomorrow.

jedel1043: That would be less than ideal though. From the same documentation you linked:
> Rust’s model for dealing with integer-to-pointer casts is called Exposed Provenance. However, the semantics of Exposed Provenance are on much less solid footing than Strict Provenance, and at this point it is not yet clear whether a satisfying unambiguous semantics can be defined for Exposed Provenance. (If that sounds bad, be reassured that other popular languages that provide integer-to-pointer casts are not faring any better.) Furthermore, Exposed Provenance will not work (well) with tools like [Miri](https://github.com/rust-lang/miri) and [CHERI](https://www.cl.cam.ac.uk/research/security/ctsrd/cheri/).

So we would basically be giving up the possibility of using Miri to scan our code for a couple of perf points... not good.

jedel1043: > Also, I have no idea how this resolves on 32-bits pointer platforms and AFAIK our CI does not test on 32-bits.

I tested `i686-unknown-linux-gnu` with cross and it passed all tests. We should definitely add an action to run tests in that platform.

jedel1043: By the way, the command I used to test using Miri was
```
MIRIFLAGS="-Zmiri-isolation-error=warn" cargo +nightly miri test -p boa_engine value::inner::nan_boxed
```

jedel1043: Pushed a change that makes it easier to run Miri. You should be able to run it now using 
```
cargo +nightly miri test -p boa_engine value::inner::nan_boxed
```
for only the nan-boxing tests. I'm currently testing the whole value suite with
```
cargo +nightly miri test -p boa_engine value::
```
but it takes a while to run the 152 tests...

hansl: I tried to run it earlier today and got into String problems with UB.

jedel1043: Weird, I was able to run the whole value suite through Miri without any errors

hansl: On the `main` branch:

```
error: Undefined Behavior: in-bounds pointer arithmetic failed: attempting to offset pointer by -1 bytes, but got 0xdabad1[noalloc] which is a dangling pointer (it has no provenance)
   --> /Volumes/Projects/hansl/boa/core/string/src/lib.rs:542:32
    |
542 |             let ptr = unsafe { self.ptr.byte_sub(1) };
    |                                ^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
```

It works on your PR, so I think the change you did on that file helped.

hansl: Let me build and run the tests on my ARM32. This will take time.

jedel1043: > On the `main` branch:
> 
> ```
> error: Undefined Behavior: in-bounds pointer arithmetic failed: attempting to offset pointer by -1 bytes, but got 0xdabad1[noalloc] which is a dangling pointer (it has no provenance)
>    --> /Volumes/Projects/hansl/boa/core/string/src/lib.rs:542:32
>     |
> 542 |             let ptr = unsafe { self.ptr.byte_sub(1) };
>     |                                ^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
> ```
> 
> It works on your PR, so I think the change you did on that file helped.

Ahhh no, that's the UB caused by the nan-boxed implementation after casting an usize into a pointer. This PR fixes that

hansl: Got it. I'm still reviewing.

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -106,27 +106,39 @@
...
-const_assert!(align_of::<*mut ()>() >= 4);
+use std::{
+    mem::ManuallyDrop,
+    ptr::{self, NonNull},
+};
+
+const _NAN_BOX_COMPAT_CHECK: () = const {
+    // We can only NaN-box pointers that are 32 or 64 bits.
+    assert!(
+        size_of::<usize>() == size_of::<u64>() || size_of::<usize>() == size_of::<u32>(),
```
hansl, line 125:
```suggestion
        size_of::<usize>() <= size_of::<u64>(),
```

Nit, for simplification.

core/engine/src/value/inner/nan_boxed.rs
hansl: BTW, just as a non-blocking. When the documentation says:

> Furthermore, Exposed Provenance will not work (well) with tools like Miri and CHERI.

I'm fairly certain that what we're doing also does this to an extent, because of the casting and bit manipulation in `value()`. 

From https://doc.rust-lang.org/std/primitive.pointer.html#method.addr:

> This is similar to self as usize, except that the [provenance](https://doc.rust-lang.org/std/ptr/index.html#provenance) of the pointer is discarded and not [exposed](https://doc.rust-lang.org/std/ptr/index.html#exposed-provenance). This means that casting the returned address back to a pointer yields a [pointer without provenance](https://doc.rust-lang.org/std/ptr/fn.without_provenance.html), which is undefined behavior to dereference. To properly restore the lost information and obtain a dereferenceable pointer, use [with_addr](https://doc.rust-lang.org/std/primitive.pointer.html#method.with_addr) or [map_addr](https://doc.rust-lang.org/std/primitive.pointer.html#method.map_addr).

I think we should still do the `expose_provenance` / `with_exposed_provenance` on the inner pointer(s).

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -501,18 +380,37 @@ impl Clone for NanBoxedValue {
...
     const fn from_inner_unchecked(inner: u64) -> Self {
-        Self(inner)
+        Self {
+            #[cfg(target_pointer_width = "32")]
+            half: (inner >> 32) as u32,
+            ptr: ptr::without_provenance_mut(inner as usize),
+        }
+    }
+
+    #[must_use]
```
hansl, line 389:
```suggestion
    /// Returns the value contained within this structure as a `u64`.
    #[must_use]
```

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -501,18 +380,37 @@ impl Clone for NanBoxedValue {
...
+    #[inline(always)]
+    fn value(&self) -> u64 {
```
hansl, line 391: According to `std::pointer::addr()` documentation, this is the same as using `self.ptr as usize as u64`. Except that the cast is still `const`, which might fix some of the performance regression.

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -453,8 +328,12 @@ const_assert!(f64::from_bits(bits::MASK_STRING).is_nan());
 const_assert!(f64::from_bits(bits::MASK_SYMBOL).is_nan());
 const_assert!(f64::from_bits(bits::MASK_BIGINT).is_nan());
 
-/// A NaN-boxed `[JsValue]`'s inner.
-pub(crate) struct NanBoxedValue(pub u64);
+/// A NaN-boxed [`JsValue`]'s inner.
+pub(crate) struct NanBoxedValue {
+    #[cfg(target_pointer_width = "32")]
+    half: u32,
+    ptr: *mut (),
```
hansl, line 306: Have you considered `NonNull<()>` instead? Not saying it's an improvement, just curious as we are doing some conversions to NonNull.

core/engine/src/value/tests.rs
```diff
@@ -636,7 +636,7 @@ fn to_int32() {
         check_to_int32!(-3_000_000_000.5 => 1_294_967_296);
         check_to_int32!(-3_000_000_000.75 => 1_294_967_296);
 
-        let base = 2f64.powi(64);
+        let base = 18_446_744_073_709_552_000.0; // approx. 2^64
```
hansl, line 639: Did this fix anything?

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -501,18 +380,37 @@ impl Clone for NanBoxedValue {
...
-        Self(inner)
+        Self {
+            #[cfg(target_pointer_width = "32")]
+            half: (inner >> 32) as u32,
+            ptr: ptr::without_provenance_mut(inner as usize),
+        }
+    }
+
+    #[must_use]
+    #[inline(always)]
```
hansl, line 390: In stdlib code, there's a lot of #[track_caller] added when running Miri for these types of casting (basically #[cfg_attr(miri, track_caller)]). Probably to help Miri with provenance. Is this something we should consider?

core/engine/src/value/inner/nan_boxed.rs
```diff
@@ -551,103 +449,128 @@ impl NanBoxedValue {
...
 
-    /// Returns a `InnerValue` from a boxed `[JsBigInt]`.
+    /// Returns a `InnerValue` from a boxed [`JsBigInt`].
     #[must_use]
     #[inline(always)]
     pub(crate) fn bigint(value: JsBigInt) -> Self {
-        Self::from_inner_unchecked(unsafe { bits::tag_bigint(Box::new(value)) })
+        let value = value.into_raw().cast::<()>().cast_mut();
+        let addr = bits::tag_bigint(value);
+        Self {
```
hansl: Adding a function like:

```rust
#[must_use]
#[inline(always)]
fn new_unchecked(ptr: u64) -> Self {
  // ...
}
```

would greatly simplify all those constructors.

hansl (changes requested):
I'm still unconvinced that this is better than using the provenance API or simply casting `as u64`.

core/engine/src/value/inner/nan_boxed.rs
jedel1043: That would give UB for 16-bit platforms, which we don't really support right now TBH, but it seems weird to claim that it works for sizes that we don't really support.

core/engine/src/value/inner/nan_boxed.rs
jedel1043: I think there's a bit of a confusion on what that means. The documentation is telling us that calling `ptr::addr` on a pointer will consume the pointer and return an address that does not have provenance. In other words, it tells us that it is UB to try to cast an `usize` to a pointer if said `usize` was the return value of (or computed from) a call to `addr`. However, this code passes a copy of the original pointer to callers of `ptr::addr`, then uses `with_addr` to change the memory address of the original pointer without losing the provenance of it.

core/engine/src/value/inner/nan_boxed.rs
jedel1043: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f6e60366fad1e0c8c60a48fce2d49094

You cannot cast pointers into `usize` in const eval.

core/engine/src/value/inner/nan_boxed.rs
jedel1043: I don't think that would work because we need the null pointer value to represent "0.0" in f64, and using `Option<NonNull<()>>` would make it so that we would have to `transmute` instead.

core/engine/src/value/tests.rs
jedel1043: Yes, the Miri VM doesn't have the exact same semantics for f64 than the host machine in complex operations like `powi` or `sqrt`, so it was causing failures. Using the exact value that we want is a lot more reliable.

core/engine/src/value/inner/nan_boxed.rs
jedel1043: That doesn't work. It wouldn't preserve the provenance of the original pointer. However, something like this should work:
```rust
fn new_pointer(ptr: *mut (), addr: usize) -> Self {
    // ...
}
```

core/engine/src/value/inner/nan_boxed.rs
jedel1043: I don't think that's to help with provenance, I think it's to make it so that the errors reported by Miri are not shown on the functions that caused the UB, but on the callers of those functions, since it's very common to cause UB by improperly following safety invariants before calling an unsafe function.

jedel1043: Case in point, if I run `cargo +nightly miri test -p boa_engine value::inner::nan_boxed::string` on the main branch I get:

```bash
error: Undefined Behavior: in-bounds pointer arithmetic failed: attempting to offset pointer by -1 bytes, but got 0xcfd2c1[noalloc] which is a dangling pointer (it has no provenance)
   --> /home/jedel/sdk/src/boa/core/string/src/lib.rs:542:32
    |
542 |             let ptr = unsafe { self.ptr.byte_sub(1) };
    |                                ^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE on thread `value::inner::n`:
    = note: inside `boa_string::JsString::unwrap` at /home/jedel/sdk/src/boa/core/string/src/lib.rs:542:32: 542:52
```

However, if if I add `#[cfg_attr(miri, track_caller)])` to the `unwrap` function I get this:

```bash
error: Undefined Behavior: in-bounds pointer arithmetic failed: attempting to offset pointer by -1 bytes, but got 0xcfd2c1[noalloc] which is a dangling pointer (it has no provenance)
   --> /home/jedel/sdk/src/boa/core/string/src/lib.rs:556:25
    |
556 |         let ptr = match self.unwrap() {
    |                         ^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE on thread `value::inner::n`:
    = note: inside `boa_string::JsString::as_str` at /home/jedel/sdk/src/boa/core/string/src/lib.rs:556:25: 556:38
```

So the attribute seems useful to trim down Miri errors and make it a bit more obvious where to fix an UB.

core/engine/src/value/inner/nan_boxed.rs
hansl: Right. Need more coffee....

core/engine/src/value/inner/nan_boxed.rs
hansl: I'm fairly certain that Boa will never run on Arduino Uno, if only because of a lack of RAM... But fair enough.

core/engine/src/value/inner/nan_boxed.rs
jedel1043: Made some changes to simplify the construction of object-like values. PTAL

hansl (approved):
LGTM. I'll do some more exploration soon to see if we get regain performance.

HalidOdat (approved):
Awesome work! Looks great to me! :smile:

jedel1043:
Merging this to address the UB, but we can squeeze more performance in future PRs

PR #4348, opened 2025-07-21
Bump temporal_rs version to v0.0.11

nekevss:

It bumps temporal_rs to the most recent version.

It also looks like there may be some formatting changes too.


nekevss:
~~I'll have to look into the regression for the next temporal_rs release.~~ But this is a pretty big increase in conformance. it fixes roughly 106 of the remaining 188 tests

Fixed by adding era/eraYear support to PlainYearMonth

raskad (approved):

PR #4358, opened 2025-08-01
Add support for utf16(-le|be) and use the class API

hansl:
Also added passing WPT tests for text encoding.


core/runtime/src/text/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+mod tests;
+
+mod encodings;
+
+/// The [`TextDecoder`][mdn] class represents an encoder for a specific method, that is
+/// a specific character encoding, like `utf-8`.
+///
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder
+#[derive(Debug, Clone, JsData, Trace, Finalize)]
+pub struct TextDecoder(#[unsafe_ignore_trace] Rc<dyn encodings::Encoder>);
```
jedel1043: This seems a bit overengineered IMO. What is the problem with using a simple enum with the list of all encodings we support? Should be smaller and faster than dyn objects.

core/runtime/src/text/encodings.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn name(&self) -> JsString {
+        js_string!("utf-16le")
+    }
+    fn encode(&self, input: &JsString, _context: &Context) -> JsResult<Vec<u8>> {
+        let bytes = input.as_str().to_vec();
+        let len = bytes.len() * 2;
+
+        // SAFETY: The vector is u16, and we transmute to u8, so this will never fail
+        // nor have any misalignment.
+        Ok(unsafe { std::slice::from_raw_parts(bytes.as_ptr().cast::<u8>(), len).to_vec() })
```
jedel1043: We already have `bytemuck` as a dependency, no? You could use that instead of unsafe code

hansl: For the engine, not for runtime.

jedel1043: Yeah, I mean that in the sense that it is already a trusted dependency for us, so we should try to delegate unsafe code to that instead

jedel1043: also the runtime depends on the engine, so we are already transitively using that dependency

core/runtime/src/text/mod.rs
hansl: So there is a point where I'll want to add a lot of encoders (using https://crates.io/crates/encoding_rs), but I guess that's not necessary to have a vtable for it. At least not yet.

jedel1043: More so because the encoders don't carry state, they're just plain old algorithms

hansl:
@jedel1043 Addressed both your comments. PTAL.

core/runtime/src/text/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// buffer passed as a parameter.
+    ///
+    /// # Errors
+    /// Any error that arises during decoding the specific encoding.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode
+    pub fn decode(&self, buffer: JsValue, context: &mut Context) -> JsResult<JsString> {
+        // `buffer` can be an `ArrayBuffer`, a `TypedArray` or a `DataView`.
+        let bytes = if let Ok(array_buffer) = JsArrayBuffer::try_from_js(&buffer, context) {
+            JsUint8Array::from_array_buffer(array_buffer, context)?
```
HalidOdat, line 80: Suggestion: This create a new gc object of type `Uint8Array`, cosider getting the `JsArrayBuffer` and then calling the [`.data()` method](https://github.com/boa-dev/boa/blob/main/core/engine/src/object/builtins/jsarraybuffer.rs#L253) when you clone the array.

EDIT: You might not even need to clone it if you need a reference to a `[u8]` (which seems to be the case)

core/runtime/src/text/encodings.rs
```diff
@@ -0,0 +1,87 @@
...
+
+pub(crate) mod utf16be {
+    use boa_engine::{JsString, js_string};
+
+    pub(crate) fn encode(input: &JsString) -> Vec<u8> {
+        let mut bytes = input.as_str().to_vec();
+
+        // Swap the bytes.
+        for b in bytes.as_mut_slice() {
+            *b = *b >> 8 | (*b & 0xFF) << 8;
```
HalidOdat: Suggestion: <https://doc.rust-lang.org/std/primitive.u16.html#method.swap_bytes>

```suggestion
            *b = b.swap_bytes();
```

core/runtime/src/text/encodings.rs
```diff
@@ -0,0 +1,87 @@
...
+            true
+        } else {
+            false
+        };
+
+        let input: &mut [u16] = bytemuck::cast_slice_mut(input);
+
+        // Swap the bytes.
+        for b in &mut *input {
+            *b = *b >> 8 | (*b & 0xFF) << 8;
```
HalidOdat:
```suggestion
            *b = b.swap_bytes();
```

core/runtime/src/text/encodings.rs
```diff
@@ -0,0 +1,87 @@
...
+            input = &input[0..input.len() - 1];
+            true
+        } else {
+            false
+        };
+
+        let input: &[u16] = bytemuck::cast_slice(input);
+
+        if dangling {
+            JsString::from(&[JsString::from(input), js_string!("\u{fffd}")])
```
HalidOdat: Remark: Since it's used in a lot of places consider adding it to the list of well known static string in `boa_string` crate.

```suggestion
            JsString::from(&[JsString::from(input), StaticJsStrings::REPLACEMENT_CHARACTER])
```

core/runtime/src/text/encodings.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+pub(crate) mod utf16le {
+    use boa_engine::{JsString, js_string};
+
+    pub(crate) fn encode(input: &JsString) -> Vec<u8> {
+        let bytes = input.as_str().to_vec();
+        let bytes = bytes.as_slice();
+        bytemuck::cast_slice(bytes).to_vec()
```
HalidOdat: Suggestion: It would be more efficient to call `.variant()` on `JsStr`, and match on the latin1 and utf16. That way we don't do any allocations if not required.

HalidOdat (approved):
Really nice work! Just some suggestion on how we might improve it :smile:

core/runtime/src/text/mod.rs
hansl: Thank you, I was looking for that. I will need to clone in the u16be case I think. Let me check.

core/runtime/src/text/encodings.rs
hansl: Doesn't exist. Added it, let me know if I should remove it.

hansl: Actually it's a UTF-16, and static js strings seem to only work with latin1 strings. Let me know if you think there's a way to do this.

core/runtime/src/text/mod.rs
hansl: Turns out this is a bit more complicated as I have a hard time owning the data and returning a ref to it. I'll try it later as an optimization.

hansl: https://github.com/boa-dev/boa/issues/4362

PR #4363, opened 2025-08-05
Regex optimization

### LINKED ISSUE #4305: Flags for Regex shouldn’t require a trip to the heap
When regex literals are parsed the flags are collected in a [Vec on the heap](https://github.com/boa-dev/boa/blob/main/core/parser/src/lexer/regex.rs#L117). This is overkill because the majority of the time there is only ever 1 flag, it’s also impossible for there to be more than 8 flags so this can almost certainly be stack allocated.

Vecs are [lazily allocated](https://doc.rust-lang.org/std/vec/struct.Vec.html) so if there are no flags this shouldn’t incur any performance penalty. But for when there are flags it can be optimised. 

One option could be to use ArrayVec, another is to try and have the lexer generate a single string for the flags and  then pass that into [parse_regex_flags](https://github.com/boa-dev/boa/blob/main/core/parser/src/lexer/regex.rs#L152) (skipping the vec filling step).

AdithyaLaakso:
This Pull Request fixes/closes #4305

It changes the following:

- Use an ArrayVec instead of a regular Vec for processing regular expression flags, avoiding an unnecessary heap allocation and memory usage especially in the case of complex nested regexps. 
- Refactors the processing of regexps to avoid deeply nested and potentially error prone match structure
- Properly shows syntax errors in cases where more than 8 regex flags are provided (disallowed by the spec).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -518,6 +518,7 @@ dependencies = [
 name = "boa_parser"
 version = "0.20.0"
 dependencies = [
+ "arrayvec",
  "bitflags 2.9.1",
  "boa_ast",
  "boa_interner",
--- a/core/parser/Cargo.toml
+++ b/core/parser/Cargo.toml
@@ -21,6 +21,7 @@ bitflags.workspace = true
 num-bigint.workspace = true
 regress.workspace = true
 icu_properties.workspace = true
+arrayvec.workspace = true
 
 [dev-dependencies]
 indoc.workspace = true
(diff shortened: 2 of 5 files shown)


jasonwilliams:
@AdithyaLaakso is there a reason you closed this?

AdithyaLaakso:
@jasonwilliams I saw a bunch of red Xs. That would mean that tests failed, right? If so, I would not know how to fix it because it works on my machine.

PR #4364, opened 2025-08-05
Add (optionally) Float16Array and f16round() support and add JsUint8ClampedArray

hansl:
For some reason the second one was missing from Rust side.

The reason to gate this behind a feature is that f16 is niche as best when using Boa in an application. Adding dependencies based on barely used features just makes the code bloated.

Enabling the test262 tests and the feature in boa_tester passes all the relevant tests.


hansl:
  
  | Total | 50,595 | 50,595 | 0 |
  | Passed | 47,253 | 47,296 | **+43** |
  | Ignored | 2,192 | 2,143 | **-49** |
  | Failed | 1,150 | 1,156 | **+6** |
  | Panics | 0 | 0 | 0 |
  | Conformance | 93.39% | 93.48% | **+0.08%** |

hansl:
For some reason that's not what I get on my computer; I get -3 failed tests which were using `f16round`...

```
Results (ECMAScript Next):
Total tests: 50595
Passed tests: 47305
Ignored tests: 2143
Failed tests: 1147 (0 panics)
Conformance: 93.50%
```

core/engine/Cargo.toml
```diff
@@ -72,6 +72,9 @@ experimental = ["temporal"]
 # Enable binding to JS APIs for system related utilities.
 js = ["dep:web-time", "dep:getrandom"]
 
+# Enable support for Float16 typed arrays
+float16 = ["dep:half"]
```
nekevss: issue: we probably shouldn't feature flag this

As much as I agree with it being niche, it is part of the specification as of [February 2025](https://github.com/tc39/proposals/blob/main/finished-proposals.md), so we should probably unconditionally support it.

Maybe there's a world where we have a Boa specific "full" feature set, where we have a smaller default version of the engine that is partially compliant and the "full" feature supports the entire specification ... but that's probably a fairly controversial stance because we would essentially be picking and choosing which of the tc39 proposals we "support" by default.

core/engine/src/builtins/typed_array/element/mod.rs
```diff
@@ -186,6 +186,32 @@ impl From<ClampedU8> for Numeric {
     }
 }
 
+/// An `u8` that clamps instead of overflowing when converting from a `JsValue`.
```
nekevss: question: is this doc accurate?

core/engine/src/builtins/math/mod.rs
```diff
@@ -447,6 +451,35 @@ impl Math {
...
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        let num = args
+            .get_or_undefined(0)
+            //1. Let n be ? ToNumber(x).
+            .to_number(context)?;
+
+        // 2. If n is NaN, return NaN.
+        // 3. If n is one of +0𝔽, -0𝔽, +∞𝔽, or -∞𝔽, return n.
+        // 4. Let n16 be the result of converting n to IEEE 754-2019 binary16 format using roundTiesToEven mode.
```
nekevss, line 477: observation: `half`'s f16 is the IEEE 754-2008 binary16 format. I'm curious what the major differences are (don't have the time to look them up currently, probably will later.).

I'm curious if this could cause some relative non spec compliance that could be surprising and hard to find or wouldn't affect it at all. On the other hand, I haven't looked around if there is another implementation like `half`, and there's a good chance that there's no major issue.

core/engine/Cargo.toml
hansl: I could enable the flag by default, at least. But I'd keep the flag.

> choosing which of the tc39 proposals we "support" by default.

The default should be the full specification, and I do think that we should feature flag many builtins that require other crates.

core/engine/src/builtins/typed_array/element/mod.rs
hansl: If it isn't it _should_ be. `ClampedU8` was already implemented and I didn't read _that_ code.

core/engine/src/builtins/math/mod.rs
hansl: Good catch. I'll look into it later today.

hansl: It might explain the difference between local and CI test failures.

core/engine/Cargo.toml
nekevss: Fair enough stance. Feature flag built-ins with external dependencies and then list them in default, and then it's not really about choosing specification features but dependencies. I like that. Do we do that similarly elsewhere and I just haven't noticed?

If not, we should go through the built-ins and retroactively feature flag them. I'd also prefer that we document this somewhere too.

hansl: Yes I think we should go through builtins at some point (outside this PR).

We do that in a few cases; runtime does that extensively, and temporal/intl are also gated.

nekevss: > Yes I think we should go through builtins at some point (outside this PR).

Agreed, but definitely in a different PR :smile: 

> temporal/intl are also gated

temporal is gated because it's still in stage 3 + the still relatively unstable state of temporal_rs. Although, hopefully that won't be the case for too much longer. intl is because ecma402 is an extension (for lack of better terms) to ecma262 + the additional compiled data included into the binary.

We should always be flagging stage 3 proposals and list them under experimental until the proposal is accepted.

hansl: Addressed this by making it the default.

core/engine/src/builtins/math/mod.rs
hansl: Looking to the wikipedia page for IEEE754 (https://en.wikipedia.org/wiki/IEEE_754):

> The current version, IEEE 754-2019, was published in July 2019.[[1]](https://en.wikipedia.org/wiki/IEEE_754#cite_note-1) It is a minor revision of the previous version, incorporating mainly clarifications, defect fixes and new recommended operations.

From the actual summary page: https://grouper.ieee.org/groups/msc/ANSI_IEEE-Std-754-2019/background/:

>  It is believed that any existing implementation of 754-2008 conforms to 754-2019 - that's primarily achieved by making new operations "should" (optional) rather than "shall" (mandatory). However there are a couple of ways that a 754-2019 implementation might not conform to 754-2008:
>
>    The 754-2008 5.3.1 required operations {min,max}{Num,NumMag} are not mentioned in 754-2019, in favor of its new 9.5 recommended operations {minimum,maximum} and {minimum,maximum}{Number,Magnitude,MagnitudeNumber}. An implementation could implement both sets and thus conform to both standards.
>    The relaxed ordering of NaNs by the 5.10 totalOrder operation is another case where an implementation that conforms to 754-2019 might not conform to 754-2008. Conformance to both could be obtained by satisfying the stricter 754-2008 requirement d)3)iii) "lesser payload, when regarded as an integer, orders below greater payload for +NaN, reverse for -NaN."

I think the implementation of `float16` should be sufficient for Boa.

core/engine/src/builtins/typed_array/element/mod.rs
hansl: Oh that was copied. Fixed.

nekevss (approved):
Looks good to me.

PR #4365, opened 2025-08-06
Allow public access to the full position of a call frame

hansl:
Prior to this, it was impossible for an application to know the origin of an API call.

My use case is that I want to enable back trace logging on `console.trace()` or when a CLI flag is used.

### DIFF
--- a/core/engine/src/vm/call_frame/mod.rs
+++ b/core/engine/src/vm/call_frame/mod.rs
@@ -3,10 +3,9 @@
 //! This module will provides everything needed to implement the `CallFrame`
 
 use super::ActiveRunnable;
-use crate::vm::source_info::SourcePath;
 use crate::{
     JsValue, builtins::iterable::IteratorRecord, environments::EnvironmentStack, realm::Realm,
-    vm::CodeBlock,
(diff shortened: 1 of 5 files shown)


core/engine/src/vm/call_frame/mod.rs
```diff
@@ -78,6 +79,19 @@ impl CallFrame {
     pub const fn code_block(&self) -> &Gc<CodeBlock> {
         &self.code_block
     }
+
+    /// Retrieves a tuple of `(`[`JsString`]`, `[`SourcePath`]`, `[`Position`]`)` to know the
+    /// location of the call frame.
+    #[inline]
+    #[must_use]
+    pub fn position(&self) -> (JsString, SourcePath, Option<Position>) {
```
HalidOdat: Consider returning a struct with the fields

```rust
using boa_ast::Position as AstPosition;

struct SourcePosition {
    function_name: JsString,
    path: SourcePath,
    position: AstPosition,
}
```

hansl: I was too lazy lol.

PR #4366, opened 2025-08-06
Implement a `JsValueStore` as well as structuredClone and tests

hansl:
The `JsValueStore` allows storage of `JsValue`s outside of a `Context`. It follows the semantics of the [Structured Clone Algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) when created.

This API will be necessary when implementing transfer across contexts. 

For now, this PR implements the `Window.structuredClone()` function correctly (except some types are not supported yet) and passes the tests I took from its documentation (and more).


hansl:
For `postMessage`, I need a context-free variant of `JsValue` (including special objects) which would also be useful for this PR, so I'll refactor this PR to include it.

core/runtime/src/store/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+mod from;
+mod to;
+
+/// Convenience method to avoid copy-pasting the same message.
+#[inline]
+fn unsupported_type() -> JsError {
+    js_error!(Error: "DataCloneError: unsupported type for structured data")
+}
+
+/// A type to help store [`JsString`]. Because [`JsString`] relies on [`std::rc::Rc`],
```
nekevss, line 19: question: should we eventually be trying to update JsString to be Send?

core/runtime/src/store/to.rs
```diff
@@ -0,0 +1,161 @@
...
+        ValueStoreInner::Undefined => Ok(JsValue::undefined()),
+        ValueStoreInner::Boolean(b) => Ok(JsValue::from(*b)),
+        ValueStoreInner::Float(f) => Ok(JsValue::from(*f)),
+        ValueStoreInner::String(s) => Ok(JsValue::from(s.to_js_string())),
+        ValueStoreInner::BigInt(b) => Ok(JsValue::from(JsBigInt::new(b.clone()))),
+        ValueStoreInner::Object(fields) => try_fields_into_js_object(store, fields, seen, context),
+        ValueStoreInner::Map(key_values) => try_into_js_map(store, key_values, seen, context),
+        ValueStoreInner::Set(values) => try_into_js_set(store, values, seen, context),
+        ValueStoreInner::Array(items) => try_items_into_js_array(store, items, seen, context),
+        ValueStoreInner::Date(_) => unimplemented!(),
```
nekevss: nit: return a JsError with a message of "Not yet implemented."

core/runtime/src/store/from.rs
```diff
@@ -0,0 +1,266 @@
...
+            .and_then(|x| x.value().cloned());
+        if let Some(v) = v {
+            let v = try_from_js_value(&v, transfer, seen, context)?;
+            inner.push(Some(v));
+        } else {
+            inner.push(None);
+        }
+    }
+
+    unsafe {
```
nekevss, line 103: nit: add safety comment

nekevss (approved):
Overall looks good to me, but I do have a couple nits that I would prefer are addressed prior to merging.

Also, as a side note to this PR. The use of std::ptf::from_ref makes me a tad nervous but should ultimately be fine. 

We really need to start some GC experiments in the oscars repo and do some work on the redesign. I'm hoping to actually get to that soon with the temporal finishing up. I mostly mention it because it may be worth keeping in mind for runtime features implementation.

core/runtime/src/store/mod.rs
hansl: That would be quasi-impossible, and really not a goal. The reference counting itself makes this impossible to be `Send`. There was a proposal to have a separate type `JsStringSend` or similar that would be Send, but at the end of the day, this is an edge case that is probably better to handle locally.

I could see a method in `JsString` that takes `self`, consumes it, and if it's the only reference, returns the underlying vector without making a copy of it. That'd be nice.

PR #4368, opened 2025-08-07
Fix clippy lints for Rust 1.89

nekevss:
Fixes a variety of cargo clippy lints for the new version.

This also removes quite a few clones in `boa_engine`! We should probably try to watch out for that pattern in the future and make use of `std::slice::from_ref`

### DIFF
--- a/core/ast/src/function/async_function.rs
+++ b/core/ast/src/function/async_function.rs
@@ -260,9 +260,10 @@ impl ToIndentedString for AsyncFunctionExpression {
     fn to_indented_string(&self, interner: &Interner, indentation: usize) -> String {
         let mut buf = "async function".to_owned();
         if self.has_binding_identifier
-            && let Some(name) = self.name {
-                let _ = write!(buf, " {}", interner.resolve_expect(name.sym()));
-            }
+            && let Some(name) = self.name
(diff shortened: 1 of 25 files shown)


raskad (approved):

core/ast/src/function/async_function.rs
```diff
@@ -259,10 +259,10 @@ impl Spanned for AsyncFunctionExpression {
 impl ToIndentedString for AsyncFunctionExpression {
     fn to_indented_string(&self, interner: &Interner, indentation: usize) -> String {
         let mut buf = "async function".to_owned();
-        if self.has_binding_identifier {
-            if let Some(name) = self.name {
-                let _ = write!(buf, " {}", interner.resolve_expect(name.sym()));
-            }
+        if self.has_binding_identifier
+            && let Some(name) = self.name
```
hansl, line 263: This means the minimum version of Rust we support would be 1.89, but I don't see any changes to rust-version in the crate manifests.

jedel1043: let-chains were introduced in 1.88 IIRC, this just runs clippy which rewrites snippets using let-chains

PR #4372, opened 2025-08-09
Fix index access in `JumpTable` instruction

HalidOdat:
Prior to this PR we accessed `JumpTable` index by popping an element of the stack, with the introduction of #3798 the mechanisms for clearing the stack have been removed and we cannot ensure that the index is the last element. This PR switches the JumpTable instruction to use a register to access the index.

### DIFF
--- a/core/engine/src/tests/mod.rs
+++ b/core/engine/src/tests/mod.rs
@@ -539,3 +539,66 @@ fn null_bool_in_object_pattern() {
         TestAction::assert_eq("c", 100),
     ]);
 }
+
+// https://github.com/boa-dev/boa/pull/4372
+#[test]
+fn fix_index_access_in_finally_jump_table_in_loop() {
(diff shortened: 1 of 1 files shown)


hansl:
@HalidOdat I can confirm my code reaches the `unreachable!()` statement, with this PR it throws the expected `not a callable function`. Thanks!

hansl:
What else is needed for this PR and can I help?

HalidOdat:
> Not sure if you want to wait for someone with deeper knowledge of the VM to review as well, I'll leave that at your discretion.

I think it’s fine to merge this. It just removes a `PushFromRegister` opcode (transitional opcode used during the shift from a stack-based to a register VM). The same needs to be done everywhere else it’s still used :smile: 

EDIT: in the end we shouldn't have any push and pop operations on the stack, only registers should be used.

PR #4378, opened 2025-08-15
Implement `Intl.DateTimeFormat`

### LINKED ISSUE #4438: Intl: Implement `Intl.DateTimeFormat`
We need to implement `Intl.DateTimeFormat`.

This is required for both Intl and Temporal conformance numbers under the ECMA402 test suite.

## Other considerations

We should keep in mind while implementing how the implementation may be able to be externalized into a separate ECMA402 styled Intl crate.

nekevss:
Closes #4438

This pull request aims to implement `Intl.DateTimeFormat`.

Posting as a draft early as I lost my initial work when my laptop died a few weeks back :upside_down_face: So better safe then sorry currently.

There's still plenty of work to be done, but feel free to take a look early :smile:


nekevss:
This is ready for review.

Just to note, this is primarily just the `format` functionality. The rest will need to be added in follow up PRs, but hopefully this keeps the general progress a bit more incremental.

EDIT: It also does not currently support Temporal. So that will need to be added in follow ups as well.

core/engine/src/builtins/intl/date_time_format/format.rs
```diff
@@ -0,0 +1,423 @@
...
+            || self.minute.is_some()
+            || self.second.is_some()
+            || self.fractional_second_digits.is_some()
+    }
+
+    pub(super) fn to_length(&self) -> Option<Length> {
+        match (self.month, self.week_day) {
+            (Some(month), _) => Some(month.to_length()),
+            (None, Some(week_day)) => Some(week_day.to_length()),
+            _ => None,
```
sffc: Suggestion: you could check the day period and era lengths, too, if you want

core/engine/src/builtins/intl/date_time_format/format.rs
```diff
@@ -0,0 +1,423 @@
...
+            (Some(_y), _m, Some(_d), None) => Some(DateFields::YMD),
+            (Some(_y), Some(_m), None, None) => Some(DateFields::YM),
+            (Some(_y), None, None, None) => Some(DateFields::Y),
+            (None, Some(_m), _d, Some(_e)) => Some(DateFields::MDE),
+            (None, Some(_m), Some(_d), None) => Some(DateFields::MD),
+            (None, Some(_m), None, None) => Some(DateFields::M),
+            (None, None, Some(_d), Some(_e)) => Some(DateFields::DE),
+            (None, None, Some(_d), None) => Some(DateFields::D),
+            (None, None, None, Some(_e)) => Some(DateFields::E),
+            _ => None,
```
sffc: Suggestion: I think if you match on (None, None, None, None) then this match is exhaustive, which would be cool.

core/engine/src/builtins/intl/date_time_format/format.rs
```diff
@@ -0,0 +1,423 @@
...
+            self.hour,
+            self.minute,
+            self.second,
+            self.fractional_second_digits,
+        ) {
+            (_h, _m, _s, Some(digits)) => Some(TimePrecision::Subsecond(digits.into())),
+            (_h, _m, Some(_s), None) => Some(TimePrecision::Second),
+            (_h, Some(_m), None, None) => Some(TimePrecision::Minute),
+            (Some(_h), None, None, None) => Some(TimePrecision::Hour),
+            _ => None,
```
sffc: Also here.

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -0,0 +1,825 @@
...
+        // TODO: Determine if any locale_preferences processing is needed here.
+
+        options.locale_preferences = (&*locale).into();
+
+        // The below handles the [[RelevantExtensionKeys]] of DateTimeFormatters
+        // internal slots.
+        //
+        // See https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots
+
+        // Handle LDML unicode key "ca", Calendar algorithm
```
sffc, line 112: Thought: This code is a bit more verbose than I was expecting; locale preferences are supposed to make code like this easier to write. If you have ideas for new functions we could add to icu_locale to make this easier, please suggest.

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -0,0 +1,825 @@
...
+fn date_time_style_format(
+    date_style: Option<DateStyle>,
+    time_style: Option<TimeStyle>,
+) -> JsResult<CompositeFieldSet> {
+    let mut builder = FieldSetBuilder::default();
+    builder.length = match date_style {
+        Some(DateStyle::Full | DateStyle::Long) => Some(Length::Long),
+        Some(DateStyle::Medium) => Some(Length::Medium),
+        Some(DateStyle::Short) => Some(Length::Short),
+        None => None, // TODO (nekevss): What to do here?
```
sffc: Read the length from TimeStyle as fallback. If both DateStyle and TimeStyle are undefined, then handle that error case which is an error case anyway

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -0,0 +1,825 @@
...
+    };
+    builder.date_fields = match date_style {
+        Some(DateStyle::Full) => Some(DateFields::YMDE),
+        Some(DateStyle::Long | DateStyle::Medium | DateStyle::Short) => Some(DateFields::YMD),
+        // NOTE (nekevss): It is a TypeError for for DateStyle or TimeStyle
+        // to be provided with explicit format components.
```
sffc: I believe it's an error, yes. Seems to be a TypeError

sffc (commented):
Nice work!

core/engine/src/builtins/intl/date_time_format/mod.rs
nekevss: This has been pretty standard for resolution on all `Intl` builtins. There may be a way to make this resolution easier to work through, but ideally this portion of the code would at least live in a potential intermediary crate.

jedel1043: Yeah, this is an unfortunate side effect of `ECMA402` requiring that all options and extensions must be tested against the data to verify that they're "supported" in some sense, or they should be excluded from the final locale.

I guess we could add on `icu_locale` a method to do "DataProvider-aware resolution", but it seems better to have that in an ICU4X-ECMA402 crate, since it's a very niche use case.

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -0,0 +1,829 @@
...
+
+#[allow(unused)] // All is currently unused, potentially remove.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub(crate) enum FormatDefaults {
+    Date,
+    Time,
+    All,
+}
+
+impl OptionType for CalendarAlgorithm {
```
jedel1043: Might be better to move these to `options.rs`

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -0,0 +1,829 @@
...
+    builder
+        .build_composite()
+        .map_err(|e| JsNativeError::range().with_message(e.to_string()).into())
+}
+
+/// Represents the `required` and `defaults` arguments in the abstract operation
+/// `toDateTimeOptions`.
+///
+/// Since `required` and `defaults` differ only in the `any` and `all` variants,
+/// we combine both in a single variant `AnyAll`.
```
jedel1043: No variant like this though, you probably meant to have an `AnyAll` variant instead, right?

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -0,0 +1,829 @@
...
+            .map(|hc| {
+                let hc =
+                    Value::try_from_utf8(hc.as_utf8()).expect("As utf8 returns a valid subtag");
+                // Handle steps 3.a-c here
+                // c. If hour12 is not undefined, set options.[[hc]] to null.
+                if hour_12.is_some() {
+                    Ok(None)
+                } else {
+                    HourCycle::try_from(&hc).map(Some)
+                }
```
jedel1043: Seems a bit wonky to convert to `Value` then to `HourCycle`. Why not implement `TryFrom<options::HourCycle>` for `HourCycle` instead?

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -0,0 +1,829 @@
...
+            // 1. Set formatOptions.[[<prop>]] to "numeric".
+            format_options.set_time_defaults();
+        }
+        // TODO (nekevss): Do we have access to the localized formats via `icu_datetime`. Is there
+        // a specific API by which this is accessed.
+        //
+        // f. Let formats be resolvedLocaleData.[[formats]].[[<resolvedCalendar>]].
+        // TODO: Support formatMatcher for formatOptions matcher
+        // g. If formatMatcher is "basic", then
+        // i. Let bestFormat be BasicFormatMatcher(formatOptions, formats).
```
jedel1043, line 712: Thought: I think this can be fetched through the `DataProvider` APIs, but it might be better to upstream a `CompositeFieldSet::build_composite_basic_ecma402` method, since it avoids having to depend on provider internals.

cc @sffc

core/engine/src/builtins/intl/date_time_format/format.rs
jedel1043, line 1: I see a lot of additional options being defined here. Maybe this should be moved to `options.rs`?

jedel1043 (commented):
Impressive work! This seems like a very good first iteration on the datetime formatter. I have a couple of suggestions, but just small nitpicks IMO.

core/engine/src/builtins/intl/date_time_format/mod.rs
nekevss: Agreed on the upstreaming part. I've mentioned it elsewhere, but the intent is to move some of this into an ECMA402 specific intl crate.  Although, from there, who's exactly responsible for this functionality between `icu_datetime` and that intl crate is probably up for debate.

There is a lot of this initial pass and the specification implementation in an ecma402 crate in general.

core/engine/src/builtins/intl/date_time_format/mod.rs
nekevss: Huh, that should definitely be an `Any` variant.

It's been a while, but I think I tried to consolidate the two enums at one point and decided against it. Obviously forgot I had noted it in the documentation.

PR #4377, opened 2025-08-14
Improve object display

xubaiwang:
This Pull Request closes #4371.

It changes the following:

1. show constructor name when display object
   ```js
   >> class Foo { a = 1 }
   undefined
   >> const foo = new Foo()
   undefined
   >> foo
   Foo {
      a: 1
   }

   // nested works well
   >> class Bar { b = new Foo() }
   undefined
   >> const bar = new Bar()
   undefined
   >> bar
   Bar {
      b: Foo {
          a: 1
       }  // <- indent error here, not brought by this PR
   }

   // work with `"Object"` as name
   >> class Object { a = 1 }
   undefined
   >> new Object()
   Object {
      a: 1
   }

   // plain does not show constructor name
   >> const plain = { a: 123 }
   undefined
   >> plain
   {
      a: 123
   }

   // builtin special classes need further improvement but not in this PR,
   // at least now we know its type
   >> new Float32Array([1,2,3,4,5])
   Float32Array {
   
   }
   ```

2. `console.{log,error,...}` now use `JsValue::display` for non string value
   ```js
   >> console.log(123)
   123
   undefined
   >> console.log("456")
   456
   undefined
   >> console.log({a: 123, b: "456"})
   {
      a: 123,
      b: "456"
   }
   undefined
   ```
4. function and class display `[class $name]`, `[Function: $name]`
   ```js
   >> console.log(console.log)
   [Function: log]
   undefined
   >> class Foo {}
   undefined
   >> console.log(Foo)
   [class Foo]
   undefined
   >> console.log(ArrayBuffer)
   [class ArrayBuffer]  // <- FIXME: I'm not able to distinguish callable and constructor with current engine
   undefined               // as custom class `Foo` also satisfies `is_callable`
   ```
6. A `flake.nix` file that makes it easier for Nix users to setup environment and contribute to the project

PR #4381, opened 2025-08-19
Add Nix flake

xubaiwang:
Nix flake is helpful to Nix users for setting up develop environment. Nix is a popular package manager on Linux, macOS and Android.

This PR is split from https://github.com/boa-dev/boa/pull/4377 .

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,96 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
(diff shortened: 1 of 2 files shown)


xbwwj:
The lack of a development shell has largely stopped my progress on boa. To move forward, I've made the nix file into a separate [repository](https://github.com/xubaiwang/nur/blob/main/devShells/boa.nix).

Nix users can now initialize a development environment using:

```bash
nix develop github:xubaiwang/nur#boa
```

While this solution works for me, it's far less discoverable for other users.

If adding Nix files directly is not favorable, I would alternatively suggest adding a `Setup` section (with subsections for Nix and other platforms/distros like Windows, Ubuntu, etc) to README, guiding users to this external nix file.

What's your opinion on this? @nekevss

nekevss:
Hey, sorry for the delay on this from my end. It's been a busy week / couple of weeks :sweat_smile: 

My general opinion is that this could be a general positive to add to Boa's repository. While I don't think any of the core maintainers right now use Nix directly right now, it could be useful for contributors such as yourself.

Currently, I'm trying to read and learn about Nix to be able to review this better and have a better idea on how to maintain it. If any of the other maintainers already know Nix and can review, I wouldn't block merging at all.

Out of curiosity, is there a GitHub Action for running the nix flake in CI to test it?

xbwwj:
There is one GitHub Action for setting up Nix <https://github.com/marketplace/actions/install-nix>. But I don't think it need a CI test. 

The flake I wrote is for development environment only, not packaging. It's nothing different from `apt install rust openssl-dev` or `winget install rust` on other platforms.

nekevss (approved):
LGTM

PR #4383, opened 2025-08-25
Implement `Math.sumPrecise`

nekevss:
This PR implements [Math.sumPrecise](https://tc39.es/ecma262/#sec-math.sumprecise) functionality.

This uses the `xsum` crate for the implementation, so that crate has been added behind a feature flag for the dependency and the dependency has been added to `boa_engine`'s defaults.

### DIFF
--- a/core/engine/src/builtins/math/mod.rs
+++ b/core/engine/src/builtins/math/mod.rs
@@ -86,6 +86,9 @@ impl IntrinsicObject for Math {
         #[cfg(feature = "float16")]
         let builder = builder.static_method(Self::f16round, js_string!("f16round"), 1);
 
+        #[cfg(feature = "xsum")]
+        let builder = builder.static_method(Self::sum_precise, js_string!("sumPrecise"), 1);
+
         builder.build();
     }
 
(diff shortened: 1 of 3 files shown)


nekevss:
@hansl If you'd prefer a default implementation with `f64`. Just let me know and I can make the update.

Although, I do think it's a good idea to keep an opt in version with `xsum` (or one of the other summation algorithms).

hansl (approved):

hansl:
The feature flag is good. If we had to implementation without the `xsum` crate, we would actually have to reimplement the `xsum` crate here.

core/engine/src/builtins/math/mod.rs
```diff
@@ -941,4 +944,120 @@ impl Math {
...
+    /// Sum each value in an iterable
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [spec]: https://tc39.es/ecma262/#sec-math.sumprecise
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sumPrecise
+    #[cfg(feature = "xsum")]
+    pub(crate) fn sum_precise(
```
Gumichocopengin8, line 957: Hello @nekevss!

I came here from https://github.com/tc39/proposal-math-sum/issues/19#issuecomment-3225284852.
That's excellent work! and thanks for using `xsum` crate.

As the author of `xsum` crate and `Math.sumPrecise` in JavascriptCore (JSC) in WebKit, I'd like to share few comments here as I found the misusage of `xsum`.
From what I can see, the current implementation in Boa has inefficient and unnecessary code.

As mentioned earlier I also implemented `Math.sumPrecise` with `xsum` in JSC, so here are the commits, which include the perf benchmark comparison in the commit messages.

### Performance in JSC

- https://github.com/WebKit/WebKit/commit/2682d2ab17a40851f7e6bcff35126e810b0646fd
  - Replaced `Shewchuk’s` algorithm with `xsum` algorithm
- https://github.com/WebKit/WebKit/commit/d95d6a80355e27cd4091e806e21019e4bc557a1e
  - Use the large accumulator of `xsum` if array size is larger than `1,000`, otherwise, use the small accumulator.

I made 2 PRs to improve the performance, so perf stats are kind of hard to see the comparison between `xsum` and `Shewchuk’s` algorithm, but in short, `xsum` is roughly 2x faster than`Shewchuk’s` algorithm with a `10,000` sized array, and the performance benefits of `xsum` can be observed across all array sizes.

So I think Boa could adopt the similar approach like the following. Use `xsum` crate instead of `PreciseSum`, and use `XsumSmall` and `XsumLarge` instead of `XsumAuto`.

- https://github.com/WebKit/WebKit/blob/6ef7c763d18e81e3cddcf2294ee66d78fafab448/Source/JavaScriptCore/runtime/MathObject.cpp#L446-L480

Furthermore, I highly recommend to stick with `xsum` crate because `xsum` implementation is significantly larger and more complicated than `Shewchuk’s` algorithm, and `xsum` crate has the exactly the same logic as in JSC (I just rewrote C++ code into Rust code).
This is the `xsum` implementation.
- https://github.com/WebKit/WebKit/blob/b21d2072809e9d5f30797756469cec5e22fb6820/Source/WTF/wtf/PreciseSum.h
- https://github.com/WebKit/WebKit/blob/d95d6a80355e27cd4091e806e21019e4bc557a1e/Source/WTF/wtf/PreciseSum.cpp

Feel free to reach out if you have questions.

FYI, JSC enabled `Math.sumPrecise` by default. 
- https://github.com/WebKit/WebKit/commit/b21d2072809e9d5f30797756469cec5e22fb6820

nekevss: Oh wow, thanks for following up on this! I'll take a look at making updates to this based off what you've mentioned!

> FYI, JSC enabled `Math.sumPrecise` by default.

`Math.sumPrecise` feature flag will be part of the default for Boa too :smile:

PR #4411, opened 2025-09-08
Fix final `Temporal.PlainTime.from` tests

nekevss:
This Pull Request is related to ongoing work on #1804.

It changes the following:

- Removes outdated specification steps
- Adds an overflow options read for the from_str path

### DIFF
--- a/core/engine/src/builtins/temporal/plain_time/mod.rs
+++ b/core/engine/src/builtins/temporal/plain_time/mod.rs
@@ -415,25 +415,9 @@ impl PlainTime {
     /// [spec]: https://tc39.es/proposal-temporal/#sec-temporal.plaintime.from
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainTime/from
     fn from(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/temporal/plain_time/mod.rs
```diff
@@ -415,25 +415,9 @@ impl PlainTime {
...
-            // item.[[ISOSecond]], item.[[ISOMillisecond]], item.[[ISOMicrosecond]],
-            // item.[[ISONanosecond]]).
-            time.inner
-        } else {
-            to_temporal_time(item, args.get(1), context)?
-        };
-
-        // 4. Return ? ToTemporalTime(item, overflow).
-        create_temporal_time(time, None, context).map(Into::into)
+        // 1. Return ? ToTemporalTime(item, options).
```
hansl, line 418: Just curious, was the spec changed or did we not adhere to it prior to this PR?

nekevss: Technically a mixture of both would be my guess.

The spec definitely changed at some point, and we stopped adhering to it as a result. I'm not entirely sure if it was a normative change or a spec simplification that hit a bug in our older implementation with the tests. The large change in `from` impl was only one of the test failures. The not reading options update on the String path of ToTemporalTime fixed the other 3 failures.

PR #4414, opened 2025-09-08
Various quality of life improvements to the CLI

hansl:
Allow an expression to be passed from the arguments (using -e). This makes it easier to just run a single expression and returning.

Additionally, if the value returned by a _file_ is undefined, don't show it. It gets confusing to see a list of "undefined" on the command line when running a bunch of modules in batch.

Another change is merging the error printing code so the output is more consistent. As well, I added distinction between an error thrown from a statement versus from a job, so the user can see where it's coming from.

Added support for timeout and generic jobs in the CLI. Now we can use setTimeout or any generic jobs. The CLI will always wait for all jobs to be completed before exiting, which makes setInterval a bit awkward if it has no end conditin.

Finally, fixed a bug in setTimeout where errors were not propagated properly.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -512,8 +512,9 @@ struct Executor {
 }
 
 impl Executor {
-    fn is_empty(&self) -> bool {
-        self.promise_jobs.borrow().is_empty()
+    fn is_empty(&self, context: &mut Context) -> bool {
+        !context.has_pending_context_jobs()
+            && self.promise_jobs.borrow().is_empty()
             && self.async_jobs.borrow().is_empty()
             && self.timeout_jobs.borrow().is_empty()
             && self.generic_jobs.borrow().is_empty()
@@ -563,9 +564,11 @@ impl JobExecutor for Executor {
 
     fn run_jobs(self: Rc<Self>, context: &mut Context) -> JsResult<()> {
         loop {
-            if self.is_empty() {
+            if self.is_empty(context) {
                 return Ok(());
             }
+
+            context.enqueue_resolved_context_jobs();
 
             self.drain_timeout_jobs(context);
             self.drain_generic_jobs(context);
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -10,6 +10,8 @@
 mod debug;
 mod helper;
 
+use boa_engine::context::time::JsInstant;
+use boa_engine::job::{GenericJob, TimeoutJob};
 use boa_engine::{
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -468,43 +502,90 @@ fn add_runtime(context: &mut Context) {
...
+    timeout_jobs: RefCell<BTreeMap<JsInstant, TimeoutJob>>,
+    generic_jobs: RefCell<VecDeque<GenericJob>>,
+}
+
+impl Executor {
+    fn is_empty(&self) -> bool {
+        self.promise_jobs.borrow().is_empty()
+            && self.async_jobs.borrow().is_empty()
+            && self.timeout_jobs.borrow().is_empty()
+            && self.generic_jobs.borrow().is_empty()
```
jedel1043: This works for our current usage of the context since we don't use multiple contexts, but if we decide to add web workers in the future, it will break. Not calling `context.has_pending_context_jobs()` and `context.enqueue_resolved_context_jobs()` will completely skip resolving calls to `Atomics.notify`.

hansl: I'm about to add workers but they will use a separate `Context`, not just a new `Realm`. That's the only way to make them multi-threaded.

I'll add the context stuff, but I'm curious why aren't those already enqueued?

hansl: Seems like those could be `AsyncJob`s.

jedel1043: > I'll add the context stuff, but I'm curious why aren't those already enqueued?

Because having the context itself enqueue its own pending notify jobs was the simpler way to enable multi-threaded job enqueuing. The alternative was to modify the API for `JobExecutor` to take an `Arc`, but we would have to deal with Rust's strict semantics on non-Send, non-Sync jobs, and implementors of `JobExecutor` would need to use `RwLock` or some other multi-threaded state to keep track of all the jobs.

jedel1043: As a whole, it seemed much simpler to have a multi-threaded flag on the context to enqueue its notify jobs, and have `enqueue_resolved_context_jobs` check that flag to see if there are pending notify jobs.

jedel1043: > Seems like those could be AsyncJobs

Ohhhhh! Yeah, this made me realize it is possible to use an async oneshot channel to resolve an async job that does the same, which would nicely avoid having to manually track all futex waiters in the context

jedel1043 (approved):
Looks good!

PR #4415, opened 2025-09-09
Implement `Date` and `RegExp` `JsValueStore` from/to

hansl:
`DataView`s are complex as they become invalid when the buffer is transferred, and JsError should be its own thing as it's missing lots of APIs in Boa.

### DIFF
--- a/core/runtime/src/store/from.rs
+++ b/core/runtime/src/store/from.rs
@@ -49,7 +49,7 @@ fn try_from_js_object(
     Ok(new_value)
 }
 
-/// Transfer an object into a store, instead of cloning it. See [mdn].
+/// Transfer an object into a store instead of cloning it. See [mdn].
 ///
 /// Only [transferable objects][to] can be transferred. Anything else will return an
 /// error. Since any object t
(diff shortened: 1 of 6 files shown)


core/runtime/src/store/from.rs
```diff
@@ -134,6 +134,36 @@ fn clone_typed_array(
     Ok(dolly)
 }
 
+fn clone_date(
+    original: &JsObject,
+    date: &JsDate,
+    seen: &mut SeenMap,
+    context: &mut Context,
+) -> JsResult<JsValueStore> {
+    let msec_since_epoch = date
```
nekevss: nit: ms_since_epoch

Super small nit. But at a glance it could be confused for something not milliseconds.

core/runtime/src/store/to.rs
```diff
@@ -123,6 +123,51 @@ fn try_into_js_set(
     Ok(JsValue::from(set))
 }
 
+fn try_into_js_date(
+    store: &JsValueStore,
+    msec: f64,
```
nekevss: nit: epoch_ms or epoch_millis

PR #4416, opened 2025-09-09
Unblock Context::run_jobs when no timeouts need to be run

hansl:
Prior to this PR, `run_jobs()` would block as long as there would be timeout jobs, even if they were in the future. A good example is using `setInterval` would run an infinite loop in `run_jobs()`.

This PR unblock `run_jobs()` when the jobs do not need to be run right now.

The new tests would infinite loop prior to this PR.

This PR also moved the code for the CLI prompt to a separate thread so the JavaScript code isn't waiting on prompts before executing tasks. This means that intervals/timeouts/promises/etc return immediately and will output their logs and move the prompt without interupting the user.

### DIFF
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -695,7 +695,6 @@ impl JobExecutor for SimpleJobExecutor {
                 && self.async_jobs.borrow().is_empty()
                 && self.generic_jobs.borrow().is_empty()
                 && no_timeout_jobs_to_run
-                && self.timeout_jobs.borrow().is_empty()
                 && group.is_empty()
             {
                 break;
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -157,6 +157,10 @@ pub struct TimeoutJob {
     job: NativeJob,
     /// Signals if the timeout job was cancelled.
     cancelled: OnceFlag,
+    /// Signals that this job is recurring. A recurring job shouldn't be
+    /// awaited for when considering whether a run of the event loop is
+    /// done.
+    recurring: bool,
 }
 
 impl TimeoutJob {
@@ -167,6 +171,18 @@ impl TimeoutJob {
             timeout: JsDuration::from_millis(timeout_in_millis),
             job,
             cancelled: OnceFlag::new(),
+            recurring: false,
+        }
+    }
+
+    /// Create a new `TimeoutJob` that is marked as recurring.
+    #[must_use]
+    pub fn recurring(job: NativeJob, timeout_in_millis: u64) -> Self {
+        Self {
+            timeout: JsDuration::from_millis(timeout_in_millis),
+            job,
+            cancelled: OnceFlag::new(),
+            recurring: true,
         }
     }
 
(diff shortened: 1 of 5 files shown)


core/engine/src/job.rs
```diff
@@ -656,10 +678,16 @@ impl JobExecutor for SimpleJobExecutor {
         Self: Sized,
     {
         loop {
+            // There are no timeout jobs to run IIF there are no jobs to execute right now.
+            let no_timeout_jobs_to_run = {
+                let now = context.borrow().clock().now();
+                !self.timeout_jobs.borrow().iter().any(|(t, _)| &now >= t)
```
jedel1043, line 691: Now that the input reader is on a separate thread, we don't really need this anymore. We can just enqueue an async job that uses something like `smol::channel::mpmc` to receive messages, then clear the job queue when the channel gets closed by the reader thread.

hansl: Disagree. In general we should let the thread sleep if there's no jobs to run. If there is an interval job the timeout job queue will never be empty and without this change would be busy polling. I also think the input would work but the channel would not be queried as it would be running an infinite loop without this PR. 

If you're suggesting making interval/timeout an async wake up job then we need to pick an executor implementation and so far we always wanted to avoid it.

jedel1043: That's unrelated though, since we could just add a sleep in `run_jobs` directly.

jedel1043: > I also think the input would work but the channel would not be queried as it would be running an infinite loop without this PR.

I'm a bit confused by this. Can you clarify a bit? What I envision is an async task that awaits the receive call in an infinite loop, but breaks when the channel gets closed.

hansl: Oh you want the prompting to run in an async job? Interesting idea...

hansl: Okay after thinking and tinkering about it, I've come to the conclusion that using jobs like that is the wrong pattern. 

Most JavaScript engine works by having tasks separated by event queues. For example, a browser will re-layout HTML and repaint, gather events (clicks, key presses, etc) and then send those to the JavaScript engine and then run microtasks before re-layout again. They don't run half tasks, events, then half other tasks. This separation of concern makes it easy to reason about the flow of execution; it's a simple loop.

Mixing events and tasks together makes the execution timeline hard to reason about. It makes me look clever, but most of the times intuitive trumps smart, if that makes sense. I believe this is one of those case.

hansl:
@jedel1043 Did you want to discuss more of this PR?

core/engine/src/job.rs
jedel1043: I think in this case it's a tradeoff. Yes, putting everything into `run_jobs` makes it hard to know the execution timeline from inside `run_jobs`, but it makes it really easy to reason about the execution timeline from outside `run_jobs`: just a thread sending messages to an event loop that is handled entirely within the `run_jobs` function.

However, with the current design the line between both gets blurred; code outside `run_jobs` needs to take care of receiving and processing text, and needs to call `run_jobs` multiple times, not mentioning that we also need to modify `TimeoutJob` so that it can support this method of doing things. So both ways have their upsides and downsides, but I prefer the async solution because that at least reduces the API surface that we need to expose on `TimeoutJob` to support this execution model.

That being said, I also think this is good enough for now, but I may try to play with the implementation after this gets merged to see what other options we have that don't require modifying `TimeoutJob`.

jedel1043 (approved):

PR #4417, opened 2025-09-10
Implement `JsValueStore` for `SharedArrayBuffer` and `postMessage`

hansl:
This adds support for creating a `JsValueStore` with a `SharedArrayBuffer` in it (including `TypedArray` pointing to a SAB), and add an implementation of `postMessage` that sends messages from JavaScript to a handler (normally another Boa `Context`).

As well as two tests;
1. `postMessage` is in the same context as `onMessageQueue`
2. two threads have their own context and send a message from one to the other.

### DIFF
--- a/core/runtime/src/message/senders.rs
+++ b/core/runtime/src/message/senders.rs
@@ -14,7 +14,7 @@ use futures::channel::mpsc::UnboundedSender;
 /// A [`MessageSender`] that reads the `onMessageQueue` property of the global
 /// object and calls it if it is a function. Note that this does not support
 /// event listeners, only checks the (made-up) `onMessageQueue` property. It
-/// also does not check the `targetOrigin` and accept all messages.
+/// also does not check the `targetOrigin` and accepts all messages.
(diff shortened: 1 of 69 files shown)


core/runtime/src/message/senders.rs
```diff
@@ -0,0 +1,67 @@
...
+use boa_engine::{
+    Context, Finalize, JsData, JsError, JsResult, JsString, JsValue, Trace, js_string,
+};
+use futures::StreamExt;
+use futures::channel::mpsc::UnboundedSender;
+
+/// A [`MessageSender`] that reads the `onMessageQueue` property of the global
+/// object and calls it if it is a function. Note that this does not support
+/// event listeners, only checks the (made-up) `onMessageQueue` property. It
+/// also does not check the `targetOrigin` and accept all messages.
```
nekevss: spelling: accepts all messages

nekevss (approved):
Looks good overall! Nice work :)

PR #4418, opened 2025-09-10
Use async channels for Atomics.waitAsync implementation

jedel1043:
Tweaks the implementation of `Atomics.waitAsync` to stop depending on the `Context` to enqueue its notify jobs. This is accomplished by enqueuing an `AsyncJob` that awaits the response of a oneshot channel, which can be send between multiple threads more easily.

Note that this makes it so that `JobExecutor` depends a bit more on having at least a "correct" implementation of `run_jobs_async`, where we don't immediately block on each future, but intersperse polling our pending futures with running promises, timeouts and generic jobs. This generally makes implementing `JobExecutor` a bit harder, but since `Atomics.waitAsync` is already a power user feature, it seemed worth overall.

### DIFF
--- a/core/engine/src/builtins/atomics/tests.rs
+++ b/core/engine/src/builtins/atomics/tests.rs
@@ -19,9 +19,10 @@ fn waiterlist_block_indexedposition_wake() {
     let i32a = JsInt32Array::from_shared_array_buffer(buffer.clone(), context).unwrap();
 
     std::thread::scope(|s| {
-        let _threads = [0, 2].map(|idx| {
+        let mut threads = Vec::new();
+        for idx in [2, 0] {
             let buffer = inner_buffer.clone();
-            s.spawn(move || {
+            let handle = s.spawn(move || {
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/atomics/tests.rs
```diff
@@ -0,0 +1,121 @@
...
+    const NUMAGENT: i32 = 2;
+    const RUNNING: i32 = 4;
+
+    let context = &mut Context::builder().can_block(true).build().unwrap();
+    let buffer = JsSharedArrayBuffer::new(size_of::<i32>() * 5, context).unwrap();
+    let inner_buffer = buffer.inner();
+    let i32a = JsInt32Array::from_shared_array_buffer(buffer.clone(), context).unwrap();
+
+    std::thread::scope(|s| {
+        let _threads = [0, 2].map(|idx| {
```
hansl: Nit: `for idx in [0, 2] {`

hansl (approved):
LGTM. That was quite the read.

PR #4422, opened 2025-09-18
Fix 2 UBs, and cleanup `GcRefCell`

hansl:
- Empty enums are uninhabited types and referencing them is UB.
- Removed the ref in `GcRefMut<>` and `GcRef<>` in favor of `NonNull<>` for good hygiene.
- Simplified the type of `GcRefMut` by having a ref to the flags instead of keeping a ref to the `GcRefCell` itself.
- Unconstrained the types for `Deref` and `DerefMut` for `Ref`/`RefMut`.
- Little things, like using `assert_ne` instead of `assert!(... != ...)`, or typos in comments, etc.

### DIFF
--- a/core/engine/src/object/datatypes.rs
+++ b/core/engine/src/object/datatypes.rs
@@ -228,6 +228,7 @@ pub(crate) struct ObjectData<T: ?Sized> {
     // Because we want to trigger the compile-time const assertion below.
     //
     // It is fine if we have as_ref/as_mut to it or any access.
+    // TODO: see below.
     #[cfg(not(miri))]
     data: T,
 
@@ -236,6 +237,12 @@ pub(crate) struct ObjectData<T: ?Sized> {
     // objects are U before casting them. Instead of trying to get Miri to
     // behave properly, we use a `Box<T>` which always has the same size as
     // `Box<U>`.
+    //
+    // Note: The `tree-borrows` checker is fine with the above in miri, it is
+    //       only the stack borrow checker that complaints.
+    //
+    // TODO: his when we have a task runner OR when tree-borrows is
+    //       the default in Miri.
     #[cfg(miri)]
     data: Box<T>,
 }
--- a/core/engine/src/object/datatypes.rs
+++ b/core/engine/src/object/datatypes.rs
@@ -228,23 +228,7 @@ pub(crate) struct ObjectData<T: ?Sized> {
     // Because we want to trigger the compile-time const assertion below.
     //
     // It is fine if we have as_ref/as_mut to it or any access.
-    // TODO: see below.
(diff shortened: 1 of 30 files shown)


core/engine/src/object/datatypes.rs
```diff
@@ -228,7 +228,16 @@ pub(crate) struct ObjectData<T: ?Sized> {
     // Because we want to trigger the compile-time const assertion below.
     //
     // It is fine if we have as_ref/as_mut to it or any access.
+    #[cfg(not(miri))]
     data: T,
+
+    // Miri triggers a UB on `downcast_ref` and `downcast_mut` because it cannot
```
nekevss: question: Was miri flagging UB on both stack borrows and tree borrows, or just one of them?

If it passes with one of them, we should note that.

core/gc/src/cell.rs
```diff
@@ -58,13 +58,13 @@ impl BorrowFlag {
     ///  - This method will panic if the current `BorrowState` is writing.
     ///  - This method will panic after incrementing if the borrow count overflows.
     fn add_reading(self) -> Self {
-        assert!(self.borrowed() != BorrowState::Writing);
+        assert_ne!(self.borrowed(), BorrowState::Writing);
```
nekevss: nit: change to `assert!`

I'm not entirely sure of the performance characteristics here. I've also only ever really seen assert used for runtime checking and would prefer these are kept as the base assert. There may not be any particular benefit one way or the other, but I'd rather we benchmark it to have an idea first.

core/gc/src/cell.rs
```diff
@@ -472,56 +474,55 @@ impl<'a, T: ?Sized, U: ?Sized> GcRefMut<'a, T, U> {
...
+        unsafe { self.value.as_mut() }
     }
 }
 
-impl<T: ?Sized, U: ?Sized> Drop for GcRefMut<'_, T, U> {
+impl<T: ?Sized> Drop for GcRefMut<'_, T> {
     fn drop(&mut self) {
-        debug_assert!(self.gc_cell.flags.get().borrowed() == BorrowState::Writing);
-        self.gc_cell.flags.set(BorrowFlag(UNUSED));
+        debug_assert_eq!(self.borrow.get().borrowed(), BorrowState::Writing);
```
nekevss: nit: re: assert comment, but the debug variation now.

nekevss (commented):
Looks great! Had a couple nits and a question to start

core/engine/src/object/datatypes.rs
hansl: Just tree borrows. I added a note.

jedel1043 (approved):
LGTM!

PR #4426, opened 2025-09-20
Use same cache key for builds and tests

jasonwilliams:
We've been getting very little cache reuse from our cache keys, and the upstream action we're using doesn't support shared cache keys properly, so its easier to just use the basic cache action.

- Use the same cache key if os, arch and cargo.lock is the same
- Remove macos 13 as that is no longer supported by Rust

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -16,16 +16,17 @@ jobs:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             binary_extension: ""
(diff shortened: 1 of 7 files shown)


jedel1043:
> Question, should we split these by test and debug/release?

I don't think it's necessary since both output to different folders, so it would not be much different to split them.

jedel1043 (commented):
I'm wondering why we need to modify the values for `matrix` with the arch, if `runner.arch` already has the arch name. Maybe we can skip modify those by just using `runner.arch` everywhere?

PR #4427, opened 2025-09-21
Fix UB on unaligned ArrayBuffers

jedel1043:
Our implementation of `ArrayBuffer` assumes all array allocations would be aligned to at least 64 bits. This works for amd64, but for other architectures it could be a wrong assumption to make, causing UB. This PR adds the `aligned-vec` crate to always allocate buffers at 64 bit boundaries, which solves the issue.

For an easy test, just run `cargo +nightly miri test array_buffer -p boa_engine`.

### DIFF
--- a/core/engine/src/builtins/array_buffer/shared.rs
+++ b/core/engine/src/builtins/array_buffer/shared.rs
@@ -1,5 +1,5 @@
 use std::{
-    alloc,
+    ptr,
     sync::{Arc, atomic::Ordering},
 };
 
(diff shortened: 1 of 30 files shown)


core/engine/src/builtins/array_buffer/shared.rs
```diff
@@ -578,38 +590,25 @@ pub(crate) fn create_shared_byte_data_block(
...
-    // already align pointers to 8 bytes, so it's a lot of work for such a small
-    // compatibility improvement.
-    assert_eq!(buffer.as_ptr().addr() % align_of::<u64>(), 0);
+    buf.resize(size, 0);
+    buf.shrink_to_fit();
+    let (data, align, len, _) = buf.into_raw_parts();
 
     // 3. Return db.
-    Ok(buffer)
+    // SAFETY: `[u8]` must be trasparently castable to `[AtomicU8]`.
```
nekevss: spelling: transparently

hansl:
I'd like to run this PR on a 32-bit platform. Have you tried that? Are you sure Miri is able to see an alignment problem on a platform that doesn't have it. Or an assert.

jedel1043:
@hansl Ran `cargo +nightly miri test array_buffer -p boa_engine --target=i686-unknown-linux-gnu` and there were no errors reported, so it should be okay for 32-bit platforms.

PR #4428, opened 2025-09-21
Allow printing test262 comment on PRs from forks

jedel1043:
Fixes our test262 action so that it can still print the test results from PRs opened from forks.

One caveat is that any changes made to that action won't be reflected on the PR that introduces the changes, because `pull_request_target` makes it so that Github pulls code from `main` to run the action, instead of pulling the code from the PR branch.

### DIFF
--- a/.github/workflows/test262.yml
+++ b/.github/workflows/test262.yml
@@ -1,14 +1,20 @@
 name: EcmaScript official test suite (test262)
 on:
+  # SECURITY WARNING: Using `pull_request_target` can be risky!
+  # It runs workflows with full repo permissions and secrets on code from forks.
+  # If not handled carefully, malicious PRs can exploit this to steal secrets or manipulate the repo.
+  #
+  # Before merging any PR that changes this file, verify that the jobs don't contain
+  # any commands that could expose credentials or secrets.
+  #
+  # GitHub Documentation: <https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target>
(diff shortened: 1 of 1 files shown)


.github/workflows/test262.yml
```diff
@@ -1,6 +1,6 @@
 name: EcmaScript official test suite (test262)
 on:
-  pull_request:
+  pull_request_target:
```
HalidOdat, line 11: Question: IFAIK the `_target` variant runs the code "target" branch, does `actions/checkout` work correctly?

HalidOdat, line 11: It may be good to put some security warnings on the event, just so we are careful with future changes :smile: 
```suggestion
  # SECURITY WARNING: Using `pull_request_target` can be risky!
  # It runs workflows with full repo permissions and secrets on code from forks.
  # If not handled carefully, malicious PRs can exploit this to steal secrets or manipulate the repo.
  #
  # GitHub Documentation: <https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target>
  pull_request_target:
```

jedel1043: Hmmm true, this only runs against the latest commit in `main`, not the latest commit in the PR. I may have to adjust this to use `github.event.pull_request.head.sha` so that it checks out the PR commit head instead.

jedel1043:
Should be patched up now. I tested it in my personal fork (https://github.com/jedel1043/boa/pull/1) and it worked pretty well!

PR #4433, opened 2025-09-24
Update to temporal_rs to 0.1 release

nekevss:
This is related to ongoing work for #1804

This PR updates Boa to the new 0.1 release of temporal_rs and timezone_provider.

Notably, I've switched us over to `CompiledTzdbProvider` for now, which will come with an increase binary size (I think the jiff_tzdb .dat file is ~200 Kb). I am hoping to update this in the near future to be more configurable though.

### DIFF
--- a/core/engine/src/context/mod.rs
+++ b/core/engine/src/context/mod.rs
@@ -10,7 +10,9 @@ pub use hooks::{DefaultHooks, HostHooks};
 pub use icu::IcuError;
 use intrinsics::Intrinsics;
 #[cfg(feature = "temporal")]
-use timezone_provider::tzif::FsTzdbProvider;
+use temporal_rs::provider::TimeZoneProvider;
+#[cfg(feature = "temporal")]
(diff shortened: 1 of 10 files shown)


core/engine/src/context/mod.rs
```diff
@@ -106,7 +108,7 @@ pub struct Context {
     can_block: bool,
 
     #[cfg(feature = "temporal")]
-    tz_provider: FsTzdbProvider,
+    tz_provider: CompiledTzdbProvider,
```
jasonwilliams, line 111: Does this make the engine quite heavy binary wise? Id expect to see the runtime pass in data rather than the engine use compiledData directly. That being said i'd imagine tests would break if we have no dta when testing the engine? (maybe i don't understand this)

nekevss: Not particularly so. We will always have to include in some sort of data that will need to be packaged or compiled in because of windows. My plan is to expose this to be a configurable option in the future.

PR #4434, opened 2025-09-24
Improve regular expression flags parsing

### LINKED ISSUE #4305: Flags for Regex shouldn’t require a trip to the heap

hansl:
There is more to do, but this gets us ~7% improvement in the regex benchmark. This should be relevant right away.

Before:

```
RESULT RegExp 89.5
```

After:

```
RESULT RegExp 96
```

Fixes #4305

### DIFF
--- a/core/parser/src/lexer/tests.rs
+++ b/core/parser/src/lexer/tests.rs
@@ -9,7 +9,6 @@ use crate::source::ReadChar;
 use boa_ast::{Keyword, Position, Span, Spanned};
 use boa_interner::Sym;
 use boa_macros::utf16;
-use std::num::NonZeroU32;
 
 fn span(start: (u32, u32), end: (u32, u32)) -> Span {
     Span::new(Position::new(start.0, start.1), Position::new(end.0, end.1))


core/parser/src/lexer/cursor.rs
```diff
@@ -158,25 +159,25 @@ impl<R: ReadChar> Cursor<R> {
...
         F: Fn(char) -> bool,
     {
-        loop {
+        for out in arr.iter_mut() {
             if !self.next_is_ascii_pred(pred)? {
                 return Ok(());
-            } else if let Some(byte) = self.next_char()? {
-                #[allow(clippy::cast_possible_truncation)]
-                buf.push(byte as u8);
+            } else if let Some(byte) = self.next_char()?.and_then(NonZeroU32::new) {
```
jasonwilliams: don't we already know at this point it would be impossible to have a NUL character make it to here as a flag?  which means it's not possible to have a zero char.

 In which case you could save few cycles and do `NonZeroU32::new_unchecked` instead. It is more verbose but less checks

jasonwilliams: If this is always dealing with ascii, does the value of Array need to be a U32? Could it not be an Option<NonZeroU8> instead?

core/parser/src/lexer/regex.rs
```diff
@@ -189,6 +202,60 @@ bitflags! {
...
+        let mut buff = [0u32; 8];
+        let mut i = 0;
+        for (c, b) in value.iter().filter_map(|v| *v).zip(buff.iter_mut()) {
+            *b = c.into();
+            i += 1;
+        }
+
+        Self::try_from(&buff[..i])
+    }
+}
```
jasonwilliams: I wonder if this is worth the extra checks and rebuiding another array stack. Could we not have the original stack just be `[U32; MaxFlags] = [0; MaxFlags]`, then just iterate through that until you hit a 0 and stop. You could then just pass that directly into TryFrom &[u32] above. 

I may have missed some context which I don't understand though

core/parser/src/lexer/cursor.rs
hansl: You're right. `next_is_ascii_pred` allows NUL to get through, but our predicate doesn't. I'll fix.

hansl:
@jasonwilliams @jedel1043 PTAL

core/parser/src/lexer/regex.rs
```diff
@@ -114,14 +117,21 @@ impl<R> Tokenizer<R> for RegexLiteral {
...
+        let mut flags: [Option<NonZeroU32>; MAXIMUM_REGEX_FLAGS] = [None; MAXIMUM_REGEX_FLAGS];
+        let n = cursor.take_array_alphabetic(&mut flags)?;
+        if n > MAXIMUM_REGEX_FLAGS {
+            // There can only be a maximum of 8 flags.
+            return Err(Error::syntax(
+                "Invalid regular expression: too many flags",
+                start_pos,
+            ));
+        }
+        let flags: [u32; MAXIMUM_REGEX_FLAGS] = flags.map(|c| c.map_or(0, NonZeroU32::get));
```
jasonwilliams: I'm not sure why you bother with the `[Option<NoneZeroU32>; MAXIMUM_REGEX_FLAGS]` at all when you just convert it back to a `[u32; MAXIMUM_REGEX_FLAGS]`, it seems a bit pointless.

Why not just start with a `[u32; MAXIMUM_REGEX_FLAGS]` from the beginning on line 120 so you don't need to map it (so `[0; MAXIMUM..]`). The Option type doesn't look like it adds much in this procedure.

core/parser/src/lexer/regex.rs
```diff
@@ -139,7 +149,7 @@ impl<R> Tokenizer<R> for RegexLiteral {
             }
         }
 
-        if let Err(error) = Regex::from_unicode(body.into_iter(), flags_str) {
+        if let Err(error) = Regex::from_unicode(body.into_iter(), flags) {
```
jasonwilliams, line 150: not related to this PR, but it would be good to cache the result of this and reuse it when we actually come to need it, but thats outside the scope of optimizing the flag parsing

core/parser/src/lexer/regex.rs
hansl: Semantically they're not the same. They take the same bytes but are clearly typed appropriately. That should give the compiler more room to optimize. `0` is never going to be a valid value.

core/parser/src/lexer/regex.rs
hansl: This is the next PR I have in queue.

hansl:
@jasonwilliams There is a huge optimization to do here where `Cursor` handles slices of data, instead of having to operate `next_char` / `peek_char`. I wanted to look into it, but I'm not sure the parser performance is more important than runtime performance.

core/parser/src/lexer/regex.rs
hansl: As a matter of fact, I don't even think we need `u32` at all. `Option<NonZeroU8>` probably should be better.

jasonwilliams:
> @jasonwilliams There is a huge optimization to do here where `Cursor` handles slices of data, instead of having to operate `next_char` / `peek_char`. I wanted to look into it, but I'm not sure the parser performance is more important than runtime performance.

That’s fair, is there any chance you could knock up an issue about it (along with some ideas), just so we don’t forget about it. Someone else may pick it up and investigate

core/parser/src/lexer/regex.rs
jasonwilliams: I think we used u32 here because that’s what next_char() returns. Whether it’s worth truncating that or not I don’t know.

> Semantically they're not the same. They take the same bytes but are clearly typed appropriately. That should give the compiler more room to optimize. 0 is never going to be a valid value.

maybe, you’re only using the Option as an intermediate step though, you still finish up with a `[u32; MAX]` so I’m not sure how much optimizations the compiler would do with that.

hansl: That's true. I'll remove the `Option`. I started this PR with `take_array_alphabetic<N>` which makes sense to take an `Option`, but it is lost along the way. We can always revise later.

core/parser/src/lexer/regex.rs
```diff
@@ -189,6 +197,41 @@ bitflags! {
     }
 }
 
+impl TryFrom<&[u32]> for RegExpFlags {
+    type Error = String;
+
+    fn try_from(value: &[u32]) -> Result<Self, Self::Error> {
+        let mut flags = Self::default();
+        for c in value {
+            let c = char::from_u32(*c)
```
jasonwilliams, line 206: What happens when zero is passed through here? I don't see a guard against that? Maybe I'm missing something. 

The input will always have 0s unless every flag is used

hansl: The input is truncated here: https://github.com/boa-dev/boa/pull/4434/files#diff-eb8b67046d176cede8e456ccd429ebdb276e99ab024246302e2b4517f432e572R129

The contract of the monad is to not have zeros, same as the `FromStr` implementation.

jasonwilliams (approved):
LGTM

core/parser/src/lexer/regex.rs
jasonwilliams: ah yes i missed that, thanks

PR #4436, opened 2025-09-25
Implement upsert methods for Map

jasonmilad:
This Pull Request implements part of [the Upsert proposal](https://tc39.es/proposal-upsert/) for `Map`.

It changes the following:

- Adds `Map.prototype.getOrInsert` method.
- Adds `Map.prototype.getOrInsertComputed` method.
- Implements canonicalization of keys (`-0` → `+0`) per the spec draft.
- Includes new tests for both methods to ensure correct behavior.

### DIFF
--- a/core/engine/src/builtins/map/mod.rs
+++ b/core/engine/src/builtins/map/mod.rs
@@ -25,14 +25,14 @@ use crate::{
 use num_traits::Zero;
 
 use super::{
-    BuiltInBuilder, BuiltInConstructor, IntrinsicObject, iterable::if_abrupt_close_iterator,
+    BuiltInBuilder, BuiltInConstructor, IntrinsicObject, canonicalize_keyed_collection_value,
+    iterable::if_abrupt_close_iterator,
 };
 
 mod map_iterator;
 pub(crate) use map_iterator::MapIterator;
 
 pub mod ordered_map;
-use crate::value::JsVariant;
 use ordered_map::OrderedMap;
 
 #[cfg(test)]
@@ -86,6 +86,12 @@ impl IntrinsicObject for Map {
             .method(Self::keys, js_string!("keys"), 0)
             .method(Self::set, js_string!("set"), 2)
             .method(Self::values, js_string!("values"), 0)
+            .method(Self::get_or_insert, js_string!("getOrInsert"), 2)
+            .method(
+                Self::get_or_insert_computed,
+                js_string!("getOrInsertComputed"),
+                2,
+            )
             .accessor(
                 js_string!("size"),
                 Some(get_size),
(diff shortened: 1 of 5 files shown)


jasonmilad:
#4436

core/engine/src/builtins/map/mod.rs
```diff
@@ -549,6 +555,111 @@ impl Map {
...
+            .as_ref()
+            .and_then(|obj| obj.downcast_mut::<OrderedMap<JsValue>>())
+            .ok_or_else(|| JsNativeError::typ().with_message("`this` is not a Map"))?;
+
+        // 3. Set key to CanonicalizeKeyedCollectionKey(key).
+        let key = match key.variant() {
+            // If key is -0𝔽, set key to +0𝔽.
+            JsVariant::Float64(r) if r.is_zero() => JsValue::new(0),
+            _ => key.clone(),
+        };
```
hansl: We have this implemented: [`canonicalize_keyed_collection_value`](https://github.com/boa-dev/boa/blob/3bf25a692bc793a3dfaf501c935891eeb9fb8d24/core/engine/src/builtins/set/mod.rs#L127). Maybe we could move that function to a utility since it'd be used outside of set.

core/engine/src/builtins/map/mod.rs
```diff
@@ -549,6 +555,111 @@ impl Map {
...
+                callback.display()
+            ))
+        })?;
+
+        // 4. Set key to CanonicalizeKeyedCollectionKey(key).
+        let key = match key.variant() {
+            // If key is -0𝔽, set key to +0𝔽.
+            JsVariant::Float64(r) if r.is_zero() => JsValue::new(0),
+            _ => key.clone(),
+        };
```
hansl: Same as above.

core/engine/src/builtins/map/mod.rs
```diff
@@ -549,6 +555,111 @@ impl Map {
...
+            return Ok(existing.clone());
+        }
+
+        // 6. Let value be ? Call(callback, undefined, « key »).
+        // 7. Let p be the Record { [[Key]]: key, [[Value]]: value }.
+        let value = callback.call(&JsValue::undefined(), std::slice::from_ref(&key), context)?;
+        // 8. Append p to M.[[MapData]].
+        map.insert(key.clone(), value.clone());
+        // 9. Return value.
+        Ok(value)
```
hansl: The comments here seem to diverge from the spec: https://tc39.es/proposal-upsert/#sec-map.prototype.getOrInsertComputed

BTW, everytime you call a JS function, you _have_ to drop the `RefMut` instance since the callback may use it, leading to a panic (since you cannot mutably reference an object twice). There are 2 test262 that are failing with panics:

- `test262/test/built-ins/Map/prototype/getOrInsertComputed/check-state-after-callback-fn-throws.js`
- `test262/test/built-ins/Map/prototype/getOrInsertComputed/overwrites-mutation-from-callbackfn.js`

A good reference on what to do is in `Set::for_each()`: https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/set/mod.rs#L494-L523

hansl (changes requested):
Good job so far!  Two minor nits and one blocker which panics in the compliance tests.

This seems to fix some test262 compliance tests, so that's great! But also, we ignored the Upsert-specific tests, to re-enable them you need to remove these lines: https://github.com/boa-dev/boa/blob/main/test262_config.toml#L51-L54

And run the test262 suite (it might take a while):

```bash
cargo run --bin boa_tester -- run
```

You can also add the `--suite test/built-ins/Map/prototype` argument to only run Map-specific compliance tests.

core/engine/src/builtins/map/mod.rs
jasonmilad: Good point. I'm thinking where to put this utility file. We could have it at the `./builtins/utils.rs`, but another idea I had was to group all the sets and maps into a `keyed_collections` folder, and then have the `utils.rs` in there. But, seems wasteful just for one function, so not sure.

hansl: It could also just be in `builtins/mod.rs` as a free standing function.

core/engine/src/builtins/map/tests.rs
```diff
@@ -305,3 +305,94 @@ fn for_of_delete() {
...
+}
+
+#[test]
+fn get_or_insert_computed_this_is_undefined_and_key_canonicalized() {
+    run_test_actions([
+        TestAction::run(
+            r#"
+            const m = new Map();
+            let seenThis, seenKey;
+            const v = m.getOrInsertComputed(-0, function(k) { 'use strict'; seenThis = this; seenKey = k; return 'ok'; });
```
hansl, line 387: Non-blocking nit: no need for `use strict`.

test262_config.toml
```diff
@@ -5,70 +5,66 @@ commit = "a073f479f80b336256b7fc4e04700c827293e2fe"
 flags = []
 
 features = [
-    ### Unimplemented features:
+  ### Unimplemented features:
```
hansl: Just realized that our `editorconfig` didn't have 4 spaces for indentation. I made https://github.com/boa-dev/boa/pull/4446, but in the meantime, could you reconfigure your editor to have 4 spaces, then re-format this file? That way we can actually see what changed :)

core/engine/src/builtins/map/tests.rs
jasonmilad: Without `use strict`, is `this` undefined?

test262_config.toml
jasonmilad: Absolutely

hansl:
Good work @jasonmilad !  Thanks so much for the PR and the patience going through the review process.

PR #4437, opened 2025-09-25
Add a simple cargo-make to the project

hansl:
It makes it easier to run `test262` and JS benchmarks.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -11,11 +11,20 @@ workspace = false
 
 [tasks.test262]
 command = "cargo"
-args = ["run", "--release", "--bin", "boa_tester", "--", "run", "${@}"]
+args = ["run", "@@remove-empty(RELEASE_ARG)", "--bin", "boa_tester", "--", "run", "${@}"]
 workspace = false
 
 [tasks.bench-js]
 condition = { env_set = [ "BOA_DATA_ROOT" ] }
 command = "cargo"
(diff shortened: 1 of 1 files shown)


Makefile.toml
```diff
@@ -0,0 +1,21 @@
...
+command = "cargo"
+args = ["fmt", "--", "--emit=files"]
+workspace = false
+
+[tasks.test]
+command = "cargo"
+args = ["test", "${@}"]
+workspace = false
+
+[tasks.test262]
```
nekevss, line 12: thought: I'm not entirely sure that we should have one test262 command that's set to release.

Overflows will only panic in debug, so running the test suite only ever with the --release flag could hide some easy to catch bugs.

hansl: Should we have both with debug being the default? Some issues might arise in release.

nekevss: Yeah, that's fine. Assuming that they are two separate tasks. I'm not entirely sure everyone needs to be running them back-to-back in one task though. That would be like 10+ minutes of tests.

My biggest concern is that if we add a task that only ever uses release and we get comfortable just using that task, then we might end up missing something that starts panicking in debug.

hansl: Added environment profiles, which makes it easier. This whole thing could be better but likely will improve over time :)

CleanShot 2025-09-26 at 09 50 27@2x

PR #4447, opened 2025-09-26
Allow TimeZoneProvider to be configurable via the ContextBuilder

nekevss:
This is related to work on #1804

It changes the following:

- Implements a `DynamicTimeZoneProvider`.
- Adds a method for setting a custom `TimeZoneProvider` on `ContextBuilder`.

This is related to the general API, so please feel free to let me know if you think this is the wrong approach.

Assuming we're happy with the result of this PR and its merged, I would be fine removing Temporal from the `experimental` feature flags for the next release. We still need to figure out a good way to make reading the host system time zone accessible, but that's still an odd aspect of the specification itself.

### DIFF
--- a/core/engine/src/context/mod.rs
+++ b/core/engine/src/context/mod.rs
@@ -14,8 +14,6 @@ use temporal_rs::provider::TimeZoneProvider;
 #[cfg(feature = "temporal")]
 use timezone_provider::tzif::CompiledTzdbProvider;
 
-#[cfg(feature = "temporal")]
-use crate::context::time::DynamicTimeZoneProvider;
 use crate::job::Job;
 use crate::module::DynModuleLoader;
 use crate::vm::RuntimeLimits;
@@ -110,7 +108,7 @@ pub struct Context {
     can_block: bool,
 
     #[cfg(feature = "temporal")]
-    timezone_provider: DynamicTimeZoneProvider,
+    timezone_provider: Box<dyn TimeZoneProvider>,
 
     /// Intl data provider.
     #[cfg(feature = "intl")]
@@ -899,8 +897,8 @@ impl Context {
 
     /// Get the Time Zone Provider
     #[cfg(feature = "temporal")]
-    pub(crate) fn timezone_provider(&self) -> &impl TimeZoneProvider {
-        &self.timezone_provider
+    pub(crate) fn timezone_provider(&self) -> &dyn TimeZoneProvider {
+        self.timezone_provider.as_ref()
     }
 }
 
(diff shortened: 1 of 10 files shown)


nekevss:
NOTE: the tests are not from this PR, but the `upsert` PR

HalidOdat (approved):

core/engine/src/context/mod.rs
```diff
@@ -892,8 +898,8 @@ impl Context {
 
     /// Get the Time Zone Provider
     #[cfg(feature = "temporal")]
-    pub(crate) fn tz_provider(&self) -> &impl TimeZoneProvider {
-        &self.tz_provider
+    pub(crate) fn timezone_provider(&self) -> &impl TimeZoneProvider {
```
jedel1043: I'd expose this as `&dyn TimeZoneProvider` instead, which avoids the additional generic (`impl`)

core/engine/src/context/time.rs
```diff
@@ -179,6 +187,68 @@ impl Clock for FixedClock {
     }
 }
 
+/// The `DynamicTimeZoneProvider` is a wrapper type that allows users
```
jedel1043: Why not use `Box<dyn TimeZoneProvider>` directly instead? We don't need a wrapper in this case, since the provider is only accessible within the crate

jedel1043 (changes requested):
Looking nice! Though, I have some suggestions.

Also, it seems like we expose this as the `timezone_provider`, but the trait is called `TimeZoneProvider`, so it might be better to expose it as the `time_zone_provider`?

jedel1043:
Probably blocked on https://github.com/boa-dev/temporal/pull/599

jedel1043:
This is unblocked now

core/engine/src/context/mod.rs
```diff
@@ -1005,6 +1017,21 @@ impl ContextBuilder {
         Ok(self)
     }
 
+    /// Set the [`timezone_provider::provider::TimeZoneProvider`] that should be used to source
```
jedel1043, line 1020: s/timezone/time_zone
It is also in other places such as the name of the getter in `Context` or the debug implementation of `ContextBuilder` and `Context`.

nekevss: I've gone back and forth on this (partially why the name is different). I think `timezone_provider` makes sense for consistency with the crate, but there is also a good argument for `time_zone_provider` as that is what is implied by the pascal case.

jedel1043: I prefer `timezone_provider` imo, it looks nicer lol

nekevss: That's what I went with too. :rofl: Something about `time_zone_provider` just does not seem right.

jedel1043 (approved):
LGTM

PR #4448, opened 2025-09-27
Add minimal CI task to cargo-make

nekevss:
This PR adds a minimal CI task to the cargo make TOML.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -17,23 +17,19 @@ run_task.name = [
 install_crate = "rustfmt"
 command = "cargo"
 args = ["fmt", "--all", "--", "--emit=files"]
-workspace = false
 
 [tasks.test]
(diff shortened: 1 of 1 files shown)


Makefile.toml
```diff
@@ -1,7 +1,22 @@
+extend = [
+    { path = "./make/ci.toml"}
+]
+
+[config]
+default_to_workspace = false
```
hansl, line 6: Does this mean we can remove all the mentions of workspace in all the rules? Neat.

hansl (approved):

Makefile.toml
nekevss: Yeah, I think it's going to be a bit cleaner to deal with overall

EDIT: oh I forgot to remove the workspace reference from task.format

PR #4451, opened 2025-10-01
Only validate the RegExp, do not optimize/compile it

### LINKED ISSUE #4432: Improve regex performance (keep parsed result for later)
**Describe the bug**
Right now we parse and evaluate every regex twice. [Once to check that its valid](https://github.com/boa-dev/boa/blob/main/core/parser/src/lexer/regex.rs#L142-L147) then [again](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/regexp/mod.rs#L341) when we actually need to compile it.

This is not efficient as we're throwing away a compiled regex, only to rebuild the same thing again during runtime.

We can't afford to skip the first check, as the specification says they need to be checked during parsing, so the best solution is to keep the compiled regex object we created and pass it through to the AST, then in builtins we can reuse that object instead of creating a new one.

So:
- Pass the compiled regex into the [RegularExpressionLiteral](https://github.com/boa-dev/boa/blob/main/core/parser/src/lexer/regex.rs#L150-L153)
- https://github.com/boa-dev/boa/blob/main/core/parser/src/lexer/token.rs#L173 would need to support holding a compiled regex
- The compiled regex [comes out here](https://github.com/boa-dev/boa/blob/main/core/parser/src/parser/expression/primary/mod.rs#L235-L239) and is passed into AstRegex
- This is then held [here](https://github.com/boa-dev/boa/blob/main/core/ast/src/expression/regexp.rs#L32-L36)
- and so on

The other option is to find out if [regress](https://github.com/ridiculousfish/regress) have a cheap way of checking if a regex string is valid without doing a full compile

hansl:
The `regress` package does export a `try_parse` method that tries to parse without optimizing and compiling the regex itself.

Fixes #4432

### DIFF
--- a/core/parser/src/lexer/regex.rs
+++ b/core/parser/src/lexer/regex.rs
@@ -5,7 +5,7 @@ use crate::source::ReadChar;
 use bitflags::bitflags;
 use boa_ast::PositionGroup;
 use boa_interner::Interner;
-use regress::{Flags, Regex};
+use regress::Flags;
 use std::fmt::{Display, Write};
 use std::str::{self, FromStr};
 
@@ -147,12 +147,15 @@ impl<R> Tokenizer<R> for RegexLiteral {
             }
         }
 
-        if let Err(error) = Regex::from_unicode(body.into_iter(), flags) {
-            return Err(Error::syntax(
-                format!("Invalid regular expression literal: {error}"),
-                start_pos,
-            ));
-        }
+        // Only try to parse and validate, do not optimize/compile.
+        drop(
+            regress::backends::try_parse(body.into_iter(), flags.into()).map_err(|error| {
+                Error::syntax(
+                    format!("Invalid regular expression literal: {error}"),
+                    start_pos,
+                )
+            })?,
+        );
 
         Ok(Token::new_by_position_group(
             TokenKind::regular_expression_literal(


jedel1043:
Huh, I wonder if we could store that as our AST node, then run the rest of the compilation pipeline at bytecode compilation time (maybe even lazily optimizing/compiling at runtime?)

hansl:
@jedel1043 We cannot do _that_ specifically. Only internal methods of `Regex` can use the intermediate representation. We could either make a PR to regress to expose those, or keep the Regex object itself in the AST.

I tried keeping the Regex object around but it ended up being quite complicated. I have a slightly better understanding of the AST now, so I could try. But this PR could still go in (it doesn't close the issue) as an intermediate step for performance.

core/parser/src/lexer/regex.rs
```diff
@@ -147,12 +147,15 @@ impl<R> Tokenizer<R> for RegexLiteral {
...
-        if let Err(error) = Regex::from_unicode(body.into_iter(), flags) {
-            return Err(Error::syntax(
-                format!("Invalid regular expression literal: {error}"),
-                start_pos,
-            ));
-        }
+        // Only try to parse and validate, do not optimize/compile.
+        drop(
+            regress::backends::try_parse(body.into_iter(), flags.into()).map_err(|error| {
+                Error::syntax(
```
jasonwilliams, line 153: should we be returning this wrapped in an `Err` like before?

hansl: It is, this is `map_err()`, then using `?` to branch off if it's an `Err(_)`.

jasonwilliams (approved):
LGTM, I agree about the complexity of keeping the result, I looked into it aswell months ago and requires a large refactor in places to pass the value through. We can always merge this now and attempt that at a later date

PR #4455, opened 2025-10-05
Try to change the internal structure of JsString for performance

hansl:
This is draft where I run experiments. Do not merge.

### DIFF
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -803,7 +803,7 @@ impl String {
             Ok(js_string!().into())
         } else {
             // 13. Return the substring of S from from to to.
-            Ok(js_string!(string.get_expect(from..to)).into())
+            Ok(unsafe { JsString::slice_unchecked(string.clone(), from, to).into() })
         }
     }
 
--- a/core/string/src/common.rs
+++ b/core/string/src/common.rs
@@ -12,7 +12,7 @@ macro_rules! well_known_statics {
             paste!{
                 #[doc = "Gets the static `JsString` for `\"" $string "\"`."]
                 pub const $name: JsString = const {
-                    JsString::from_static_js_str(Self::find_static_js_string($string))
+                    JsString::from_static(Self::find_static_js_string($string))
                 };
             }
         )+
@@ -73,7 +73,7 @@ impl StaticJsStrings {
         // SAFETY: Type of T in is `&'static JsStr<'static>`, so this is safe.
         let ptr = unsafe { std::mem::transmute::<&JsStr<'_>, &'static JsStr<'static>>(str) };
 
-        Some(JsString::from_static_js_str(ptr))
+        Some(JsString::from_static(ptr))
     }
 
     // Some consts are only used on certain features, which triggers the unused lint.
--- a/core/string/src/lib.rs
+++ b/core/string/src/lib.rs
@@ -546,7 +546,7 @@ impl JsString {
     #[must_use]
     pub const fn from_static(src: &'static JsStr<'static>) -> Self {
         // SAFETY: A reference cannot be null, so this is safe.
-        let ptr = unsafe { NonNull::from_ref(src) };
+        let ptr = NonNull::from_ref(src);
 
         // SAFETY:
         // - Adding one to an aligned pointer will tag the pointer's last bit.
@@ -576,7 +576,7 @@ impl JsString {
     /// range.
     #[inline]
     #[must_use]
-    unsafe fn from_slice_unchecked(data: JsString, start: usize, end: usize) -> Self {
+    pub unsafe fn slice_unchecked(data: JsString, start: usize, end: usize) -> Self {
         let ptr = Box::into_raw(Box::new(SliceString { data, start, end }));
         // SAFETY: Guaranteed to be a valid pointer just allocated.
         let ptr = NonNull::new(ptr).unwrap();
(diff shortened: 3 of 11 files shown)


core/engine/src/builtins/string/mod.rs
```diff
@@ -1887,7 +1888,9 @@ impl String {
         let to = max(final_start, final_end);
 
         // 10. Return the substring of S from from to to.
-        Ok(js_string!(string.get_expect(from..to)).into())
+        // Ok(js_string!(string.get_expect(from..to)).into())
```
nekevss: issue: remove commented out code

core/string/src/common.rs
```diff
@@ -214,735 +221,730 @@ const MAX_STATIC_LENGTH: usize = {
...
+    StaticString::new(JsStr::latin1("get epochNanoseconds".as_bytes())),
+    StaticString::new(JsStr::latin1("epochSeconds".as_bytes())),
+    StaticString::new(JsStr::latin1("epochMilliseconds".as_bytes())),
+    StaticString::new(JsStr::latin1("epochMicroseconds".as_bytes())),
+    StaticString::new(JsStr::latin1("epochNanoseconds".as_bytes())),
+    StaticString::new(JsStr::latin1("subtract".as_bytes())),
+    StaticString::new(JsStr::latin1("until".as_bytes())),
+    StaticString::new(JsStr::latin1("since".as_bytes())),
+    StaticString::new(JsStr::latin1("equals".as_bytes())),
+    StaticString::new(JsStr::latin1("toZonedDateTime".as_bytes())),
```
nekevss, line 809: thought: temporal will probably always be `temporal` feature flagged because of the dependency, so should we also be feature flagging the `StaticString`s that we register.

core/string/src/lib.rs
```diff
@@ -598,11 +579,14 @@ impl JsString {
             full_count = sum;
         }
 
-        let ptr = Self::allocate_inner(full_count, latin1_encoding);
+        let ptr = Self::allocate_seq(full_count, latin1_encoding);
 
         let string = {
-            // SAFETY: `allocate_inner` guarantees that `ptr` is a valid pointer.
-            let mut data = unsafe { (&raw mut (*ptr.as_ptr()).data).cast::<u8>() };
+            // SAFETY: `allocate_inner` guarantees that `ptr` is a valid pointer to a `SequenceString`.
```
nekevss: nit: allocate_seq

core/string/src/lib.rs
```diff
@@ -641,22 +625,20 @@ impl JsString {
...
 
         StaticJsStrings::get_string(&string.as_str()).unwrap_or(string)
     }
 
-    /// Allocates a new [`RawJsString`] with an internal capacity of `str_len` chars.
+    /// Allocates a new [`SequenceString`] with an internal capacity of `str_len` chars.
     ///
     /// # Panics
     ///
     /// Panics if `try_allocate_inner` returns `Err`.
```
nekevss: nit: try_allocate_seq

core/string/src/lib.rs
```diff
@@ -732,9 +711,10 @@ impl JsString {
     /// Creates a new [`JsString`] from `data`, without checking if the string is in the interner.
     fn from_slice_skip_interning(string: JsStr<'_>) -> Self {
         let count = string.len();
-        let ptr = Self::allocate_inner(count, string.is_latin1());
+        let ptr = Self::allocate_seq(count, string.is_latin1());
 
         // SAFETY: `allocate_inner` guarantees that `ptr` is a valid pointer.
+        // let data = unsafe { ptr.as_ref().data().cast_mut() };
```
nekevss: nit: remove commented code

core/string/src/lib.rs
```diff
@@ -732,9 +711,10 @@ impl JsString {
     /// Creates a new [`JsString`] from `data`, without checking if the string is in the interner.
     fn from_slice_skip_interning(string: JsStr<'_>) -> Self {
         let count = string.len();
-        let ptr = Self::allocate_inner(count, string.is_latin1());
+        let ptr = Self::allocate_seq(count, string.is_latin1());
 
         // SAFETY: `allocate_inner` guarantees that `ptr` is a valid pointer.
```
nekevss: nit: change `allocate_inner`

core/string/src/lib.rs
```diff
@@ -477,17 +404,24 @@ impl JsString {
...
+
+    /// Consumes the [`JsString`], returning the internal pointer.
     ///
     /// To avoid a memory leak the pointer must be converted back to a `JsString` using
     /// [`JsString::from_raw`].
     #[inline]
     #[must_use]
-    pub fn into_raw(self) -> NonNull<RawJsString> {
-        ManuallyDrop::new(self).ptr
+    pub fn into_raw(self) -> NonNull<()> {
```
nekevss: thought: Is `NonNull<()>` the best type here.

I'm not sure having APIs that expose and accept a `NonNull<()>` is the best approach vs. casting to a specific new type (e.g. `NonNull<RawAnyString>``). This is not something I'm necessarily pushing for being changed. It mostly gave me a bit of a pause when reviewing this.

core/string/src/common.rs
hansl: I wouldn't worry for now. Maybe we should move these constants to boa_engine where it's the only place they're used, then we should feature flag some. I don't think boa_string should have temporal feature flags, personally.

core/string/src/lib.rs
hansl: Changed it to an opaque type.

core/string/src/common.rs
nekevss: Actually that's very true.

nekevss (approved):
Looks good to me

PR #4459, opened 2025-10-07
Implement Upsert methods for weakMap: getOrInsert and getOrInsertComputed

### LINKED ISSUE #4435: Add getOrInsert and getOrInsertCallable to map & weak_map
**ECMASCript feature**
[Upsert proposal for maps has made it to stage three. ](https://tc39.es/proposal-upsert/)This ticket is tracking that implementation.

[spec]: https://tc39.es/proposal-upsert/

**Example code**

This code should now work:
```javascript
// Currently
let counts = new Map();
if (counts.has(key)) {
  counts.set(key, counts.get(key) + 1);
} else {
  counts.set(key, 1);
}

// Using getOrInsert
let counts = new Map();
counts.set(key, counts.getOrInsert(key, 0) + 1);

// Using getOrInsertComputed
let grouped = new Map();
for (let [key, ...values] of data) {
  grouped.getOrInsertComputed(key, () => []).push(...values);
}
```

rrogerc:
This Pull Request closes #4435, implementing the [Upsert proposal](https://tc39.es/proposal-upsert/) for weak map

It changes the following:
- Adds `WeakMap.prototype.getOrInsert` method.
- Adds `WeakMap.prototype.getOrInsertComputed` method.
- Enforces object-only keys and callable callbacks.
- Includes new tests for both methods to ensure correct behavior and reentrancy.

### DIFF
--- a/core/engine/src/builtins/weak_map/mod.rs
+++ b/core/engine/src/builtins/weak_map/mod.rs
@@ -28,6 +28,9 @@ type NativeWeakMap = boa_gc::WeakMap<ErasedVTableObject, JsValue>;
 #[derive(Debug, Trace, Finalize)]
 pub(crate) struct WeakMap;
 
+#[cfg(test)]
+mod tests;
+
 impl IntrinsicObject for WeakMap {
     fn get(intrinsics: &Intrinsics) -> JsObject {
         Self::STANDARD_CONSTRUCTOR(intrinsics.constructors()).constructor()
@@ -44,6 +47,12 @@ impl IntrinsicObject for WeakMap {
             .method(Self::get, js_string!("get"), 1)
             .method(Self::has, js_string!("has"), 1)
             .method(Self::set, js_string!("set"), 2)
+            .method(Self::get_or_insert, js_string!("getOrInsert"), 2)
+            .method(
+                Self::get_or_insert_computed,
+                js_string!("getOrInsertComputed"),
+                2,
+            )
             .build();
     }
 }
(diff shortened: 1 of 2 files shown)


core/engine/src/builtins/weak_map/mod.rs
```diff
@@ -265,4 +274,129 @@ impl WeakMap {
...
+        let mut map = object
+            .as_ref()
+            .and_then(JsObject::downcast_mut::<NativeWeakMap>)
+            .ok_or_else(|| {
+                JsNativeError::typ()
+                    .with_message("WeakMap.getOrInsert: called with non-object value")
+            })?;
+
+        // 3. If CanBeHeldWeakly(key) is false, throw a TypeError exception.
+        //    In Boa this maps to: key must be an Object.
```
jedel1043: Could you change this to a `TODO` comment instead? It is true that this currently maps to `key must be an Object`, but hopefully in the future we'll have a proper `CanBeHeldWeakly` to also accept Symbols as valid keys.

core/engine/src/builtins/weak_map/mod.rs
```diff
@@ -265,4 +274,129 @@ impl WeakMap {
...
+        // 1. Let M be the this value.
+        // 2. Perform ? RequireInternalSlot(M, [[WeakMapData]]).
+        let object = this.as_object();
+        let map = object
+            .as_ref()
+            .and_then(JsObject::downcast_ref::<NativeWeakMap>)
+            .ok_or_else(|| {
+                JsNativeError::typ()
+                    .with_message("WeakMap.getOrInsertComputed: called with non-object value")
+            })?;
```
jedel1043, line 356: You can avoid downcasting multiple times by using `JsObject::downcast` instead. This will return a "typed" object that you can borrow from by using the `borrow` and `borrow_mut` methods, and the type system will know that you're working with a `JsObject` that has a `NativeWeakMap` inside.

PR #4462, opened 2025-10-10
perf: Improve `Math.sumPrecise` Performance

Gumichocopengin8:
improve `Math.sumPrecise` performance by using `XsumVariant` instead of `XsumAuto`

It changes the following:

- `XsumAuto` has some overhead to determine when to switch from `XsumSmall` to `XsumLarge`, so use `XsumVariant` instead.

### DIFF
--- a/core/engine/src/builtins/math/mod.rs
+++ b/core/engine/src/builtins/math/mod.rs
@@ -960,7 +960,7 @@ impl Math {
         context: &mut Context,
     ) -> JsResult<JsValue> {
         const ITERATION_MAX: u64 = 2u64.pow(53) - 1;
-        use xsum::{Xsum, XsumAuto};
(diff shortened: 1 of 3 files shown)


core/engine/src/builtins/math/mod.rs
```diff
@@ -973,15 +973,27 @@ impl Math {
...
+        let length = if iterator_record.iterator().is_array() {
+            iterator_record
+                .iterator()
+                .length_of_array_like(context)
+                .ok()
+        } else {
+            None
+        };
+        let mut sum = if let Some(length) = length {
+            if length <= XSUM_THRESHOLD as u64 {
```
Gumichocopengin8, line 989: The threshold can be a bigger number than `XSUM_THRESHOLD` (1,000), but conventionally used `XSUM_THRESHOLD` for now.

Gumichocopengin8:
@nekevss Could you please take a look at this? Thanks.

PR #4463, opened 2025-10-10
add changelog, set correct heading levels, update versions

jasonwilliams:


nekevss (commented):
Did we bump the version on the utility crates?

CHANGELOG.md
```diff
@@ -1,8 +1,283 @@
...
+- Update to temporal_rs to 0.1 release by @nekevss in https://github.com/boa-dev/boa/pull/4433
+- Add minimal CI task to cargo-make by @nekevss in https://github.com/boa-dev/boa/pull/4448
+- Improve regular expression flags parsing by @hansl in https://github.com/boa-dev/boa/pull/4434
+- Bump the rust-dependencies group with 4 updates by @dependabot[bot] in https://github.com/boa-dev/boa/pull/4449
+- Only validate the RegExp, do not optimize/compile it by @hansl in https://github.com/boa-dev/boa/pull/4451
+- Bump peter-evans/create-or-update-comment from 4 to 5 by @dependabot[bot] in https://github.com/boa-dev/boa/pull/4456
+- Bump peter-evans/find-comment from 3 to 4 by @dependabot[bot] in https://github.com/boa-dev/boa/pull/4457
+- Bump bytemuck from 1.23.2 to 1.24.0 in the rust-dependencies group by @dependabot[bot] in https://github.com/boa-dev/boa/pull/4458
+
+## New Contributors
```
Razican:
```suggestion
### New Contributors
```

Razican (commented):
Looks great, so many changes!

I think one header was not right, though

utils/tag_ptr/Cargo.toml
```diff
@@ -4,7 +4,7 @@ keywords = ["tag", "pointer"]
 categories = ["data-structures"]
 readme = "../../README.md"
 description = "Utility library that enables a pointer to be associated with a tag of type `usize`"
-version = "0.1.0"
```
jedel1043, line 7: I remember that we decided to version these utilities independently from the core crates, so it was fine to release 0.1.0 instead.

Cargo.toml
jedel1043: Oh, I remembered what was the missing piece. This needs a version just like the repo crates above, since currently Cargo does not resolve the version from a path dependency, it just throws an error when publishing.

.github/workflows/release.yml
```diff
@@ -85,15 +85,15 @@ jobs:
         build: [linux, macos-arm64, win-msvc]
         include:
         - build: linux
-          os: ubuntu-latest
+          os: ubuntu-20.04
```
jedel1043: Just realized we're using 20.04 to build the binaries. Do we want to try to migrate to 24.04?

jedel1043: Same with the rest of the OS

jedel1043 (approved):
Looks good, just made a small comment that doesn't block merging.

PR #4464, opened 2025-10-12
Cleanup `BuiltInConstructor` constants to ensure no additional allocations

jedel1043:
Cleanup some of the `BuiltInConstructor` constants to ensure the constants are more descriptive and the pushed properties don't exceeded the defined allocated property slots.

### DIFF
--- a/core/engine/src/builtins/builder.rs
+++ b/core/engine/src/builtins/builder.rs
@@ -182,17 +182,18 @@ pub(crate) struct BuiltInConstructorWithPrototype<'ctx> {
 
 #[allow(dead_code)]
 impl BuiltInConstructorWithPrototype<'_> {
-    /// The number of static properties that are always present in a standard constructor.
+    /// The number of storage slots that are always present in a standard
+    /// constructor object.
     ///
     /// See [`BuiltInConstructorWithPrototype::build`].
     const OWN_CONSTRUCTOR_STORAGE_SLOTS: usize = 3;
 
-    /// The number of prototype properties that are always present in a standard constructor.
+    /// The number of storage slots properties that are always present in a
+    /// standard constructor's prototype object.
     ///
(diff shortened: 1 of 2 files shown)


core/engine/src/builtins/builder.rs
```diff
@@ -167,20 +164,35 @@ pub(crate) struct BuiltInConstructorWithPrototype<'ctx> {
...
+    ///
+    /// See [`BuiltInConstructorWithPrototype::build`].
+    const OWN_CONSTRUCTOR_STORAGE_SLOTS: usize = 3;
+
+    /// The number of prototype properties that are always present in a standard constructor.
+    ///
+    /// See [`BuiltInConstructorWithPrototype::build`].
+    const OWN_PROTOTYPE_STORAGE_SLOTS: usize = 1;
+
+    // The nub
```
jasonwilliams: is this line meant to be here?

core/engine/src/builtins/builder.rs
```diff
@@ -374,9 +387,32 @@ impl BuiltInConstructorWithPrototype<'_> {
...
+                    <= self.constructor_storage_slots_expected
+                        + Self::OWN_CONSTRUCTOR_STORAGE_SLOTS,
+                "expected to allocate at most {} constructor storage slots, got {}. \
+                constant {}::CONSTRUCTOR_STORAGE_SLOTS may need to be adjusted",
+                self.constructor_storage_slots_expected,
+                self.constructor_storage.len() - Self::OWN_CONSTRUCTOR_STORAGE_SLOTS,
+                self.name.display_escaped(),
+            );
+        }
+
```
jasonwilliams, line 416: This is a great idea to have a debug assertion to check

core/engine/src/builtins/builder.rs
jedel1043: Nope, fixed it up

PR #4468, opened 2025-10-14
Use persistent vector for performance comparison

stevefan1999-personal:
I'm just trying to replace some Vec with [imbl](https://docs.rs/imbl/latest/imbl/) and I want to see if someone could benchmark it. My best guess is that it should make the performance slightly worse.

This would be a draft PR for a long time, probably this would only introduce marginal return.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,6 +116,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "archery"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70e0a5f99dfebb87bb342d0f53bb92c81842e100bbb915223e38349580e5441d"
+
+[[package]]
 name = "arrayvec"
 version = "0.7.6"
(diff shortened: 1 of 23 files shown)


jedel1043:
### PR

RESULT Richards 85.8
RESULT DeltaBlue 83.7
RESULT Crypto 74.3
RESULT RayTrace 142
RESULT EarleyBoyer 173
RESULT RegExp 67.6
RESULT Splay 505
RESULT NavierStokes 153
SCORE 127

### main

RESULT Richards 182
RESULT DeltaBlue 190
RESULT Crypto 148
RESULT RayTrace 408
RESULT EarleyBoyer 488
RESULT RegExp 75.2
RESULT Splay 724
RESULT NavierStokes 326
SCORE 255

... Yeah, it is at least 2x slower to use immutable structs.

jedel1043:
Closing this to avoid keeping stale PRs, but feel free to reopen if you find a way to optimize the implementation.

PR #4476, opened 2025-10-20
ci: add SemVer check workflow to detect API-breaking changes

### LINKED ISSUE #4471: We should enable semver checks on CI
As we approach 1.0 we should make sure we’re not making breaking changes or at the very least tracking breaking changes that are made intentionally.

To do this we should add a new CI on each PR which runs https://github.com/obi1kenobi/cargo-semver-checks.

We probably want to have a PR ready but enable this after this after the release so we can catch all PRs after.

thakazuki:
This Pull Request fixes/closes #4471.

It changes the following:

- Adds a new github workflow to check semantic versioning via "cargo-semver-checks" on pull requests.


.github/workflows/semver.yml
```diff
@@ -0,0 +1,30 @@
...
+          fetch-depth: 0
+
+      - name: Cache Cargo
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
jasonwilliams: This isn't quite right, see https://github.com/boa-dev/boa/blob/main/.github/workflows/nightly_build.yml#L38-L46

.github/workflows/semver.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Check semver
+        uses: obi1kenobi/cargo-semver-checks-action@v2
```
jasonwilliams: new line needed at the end of the file

jasonwilliams (dismissed):

.github/workflows/semver.yml
jedel1043, line 1: I don't think we need a whole file for this, seeing the check is so simple to set up. You can move it to be inside the `rust.yml` set of actions.

jedel1043 (changes requested):

.github/workflows/semver.yml
thakazuki: Understood, thx for pointing that out.

.github/workflows/semver.yml
thakazuki: Understood
I deleted the new file and moved the changes to rust.yml.
Additionally I adjusted the code to be more consistent with the other jobs.

.github/workflows/rust.yml
```diff
@@ -297,3 +297,29 @@ jobs:
           cargo run -p boa_examples --bin 2>&1 \
             | grep -E '^ ' \
             | xargs -n1 sh -c 'cargo run -p boa_examples --profile ci --bin $0 || exit 255'
+
+  runSemVerCheck:
```
jedel1043:
```suggestion
  run-semver-check:
```
To be consistent with the rest of the job names.

.github/workflows/rust.yml
```diff
@@ -297,3 +297,29 @@ jobs:
...
+    timeout-minutes: 60
+    steps:
+      - name: Set environment
+        run: |
+          target=$(rustc -vV | awk '/^host/ { print $2 }' | tr [:lower:] [:upper:] | tr '-' '_')
+          echo "CARGO_TARGET_${target}_RUSTFLAGS=$W_FLAGS" >> $GITHUB_ENV
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
```
jedel1043:
```suggestion
```
Not sure if you need to fetch all history; IIRC `cargo-semver-checks` works perfectly fine with only the single latest commit, since it checks against the version published in crates.io. However, I didn't test that. Can you check that it works, and remove these lines if that's the case?

jedel1043 (commented):
Some smaller comments, but it's looking almost ready

jasonwilliams:
@jedel1043 im imagining there will be times where we will want to make breaking changes, and we don’t want the CI to keep failing and blocking merges.

so I guess unlike other CI checks, we may want a way to “skip” this or mark it as optional.

I don’t have a solution in mind but it’s worth taking this into account

jedel1043:
@jasonwilliams Oh no worries about that. We need to explicitly add the action to the branch rules if we want to require it for PRs, so if we don't add the action, it shouldn't block merges.

.github/workflows/rust.yml
thakazuki: Just tested it, cargo-semver-checks exits with code 1 which is the expected result when it detects breaking changes. It seems to work fine without full history.

jedel1043:
Now that we published 0.21, we can try to merge this. Gonna update the branch to see that the action passes, then merge after that.
@thakazuki Thank you for the contribution!

jedel1043 (approved):

.github/workflows/rust.yml
```diff
@@ -297,3 +297,28 @@ jobs:
...
+
+  run-semver-check:
+    name: Check SemVer compatibility
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Set environment
+        run: |
+          target=$(rustc -vV | awk '/^host/ { print $2 }' | tr [:lower:] [:upper:] | tr '-' '_')
+          echo "CARGO_TARGET_${target}_RUSTFLAGS=$W_FLAGS" >> $GITHUB_ENV
```
obi1kenobi, line 309: Hey, I'm the `cargo-semver-checks` maintainer 👋

Hope you don't mind me giving you a heads-up here: the `-D warnings` set here is ineffective and explicitly overridden by `cargo-semver-checks` using `--cap-lints=allow`.

This is because `-D warnings` is indiscriminate and can cause `cargo-semver-checks` to fail to build *the baseline* version, not just the current version being checked. For example, we've had situations where:
- a dependency X published a new non-major version
- the new version caused a new warning to appear in all versions of crate Y that depends on X
- downstream crate Z, which depends on Y but doesn't know about X, cannot run `cargo-semver-checks` because of the new warnings upstream 

Wanting to deny warnings is understandable, but it's best to do that in jobs other than `cargo-semver-checks` because this job is the only one that'll need to build older versions of your crates, not just `main`.

PR #4486, opened 2025-10-24
Implement RegExp.escape

### LINKED ISSUE #4443: Implement `RegExp.escape` proposal
https://github.com/tc39/proposal-regex-escaping

alabhyajindal:
Closes #4443.

Implements a static method `escape` on the global `RegExp` object. The test cases have been added from the examples given in the [proposal](https://github.com/tc39/proposal-regex-escaping).

### DIFF
--- a/core/engine/src/builtins/regexp/mod.rs
+++ b/core/engine/src/builtins/regexp/mod.rs
@@ -787,7 +787,8 @@ impl RegExp {
 
     /// Helper function to check if a character is a WhiteSpace character
     fn is_whitespace(ch: char) -> bool {
-        matches!(ch,
+        matches!(
+            ch,
             '\u{0009}' | // <TAB>
             '\u{000B}' | // <VT>
             '\u{000C}' | // <FF>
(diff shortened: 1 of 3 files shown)


alabhyajindal:
3 tests are failing for `escape` from the `test262` directory. Marking the PR as draft, until those are resolved.

core/engine/src/builtins/regexp/mod.rs
```diff
@@ -772,6 +773,184 @@ impl RegExp {
...
+        if !arg.is_string() {
+            return Err(JsNativeError::typ()
+                .with_message("RegExp.escape requires a string argument")
+                .into());
+        }
+
+        let string = arg.to_string(context)?;
+
+        // 2. Let escaped be the empty String.
+        let mut escaped = Vec::<u16>::new();
```
jedel1043: Could probably use `CommonJsStringBuilder` here, which should be a bit more optimized.

jedel1043 (changes requested):
Nicely done! I have a small suggestion, but everything else looks good.

core/engine/src/builtins/regexp/mod.rs
alabhyajindal: Thanks! That simplified things quite a bit.

jedel1043 (approved):
Awesome, let's merge!

core/engine/src/builtins/regexp/mod.rs
```diff
@@ -772,6 +773,169 @@ impl RegExp {
...
+        let arg = args.get_or_undefined(0);
+
+        // 1. If S is not a String, throw a TypeError exception.
+        if !arg.is_string() {
+            return Err(JsNativeError::typ()
+                .with_message("RegExp.escape requires a string argument")
+                .into());
+        }
+
+        let string = arg.to_string(context)?;
```
jedel1043: You can use `JsValue::as_string` here instead.

PR #4490, opened 2025-10-24
Throw on negative values when using `Overflow::Reject` in `Temporal.PlainTime.with`

HalidOdat:
This PR fixes the last failing test on the `Temporal.PlainTime` test suite ([`throws-if-time-is-invalid-when-overflow-is-reject.js`](https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/PlainTime/prototype/with/throws-if-time-is-invalid-when-overflow-is-reject.js)).

### DIFF
--- a/core/engine/src/builtins/temporal/plain_time/mod.rs
+++ b/core/engine/src/builtins/temporal/plain_time/mod.rs
@@ -31,6 +31,9 @@ use temporal_rs::{
     primitive::FiniteF64,
 };
 
+#[cfg(test)]
+mod tests;
+
 /// The `Temporal.PlainTime` built-in implementation.
 ///
 /// More information:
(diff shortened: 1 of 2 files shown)


jedel1043:
Why not upstream this to `temporal_rs`? Is this behaviour only visible from the engine side?

HalidOdat:
> Why not upstream this to `temporal_rs`? Is this behaviour only visible from the engine side?

The issue is that `PartialTime` in `temporal_rs` expects only unsigned integers, while the specification allows negative integers too. This means the validation can't be delayed after overflow retrieval (hence the two `[Js]PartialTime`). From a Rust API design standpoint as a standalone library, restricting `PartialTime` to the narrowest valid type makes sense (less margin for error from the user).

I’d be interested to hear what @nekevss thinks as well? :smile:

nekevss:
See boa-dev/temporal#432 and boa-dev/temporal#334. So yes, the issue has been known and preferably this behavior should be upstream in `temporal_rs`. The question has been the best way to represent this in `temporal_rs`'s API.

It is pretty niche behavior, so I've delayed adding it. My goal is to ultimately land something in temporal_rs before 1.0.

PR #4495, opened 2025-10-25
Add strict types to `Promise` objects

jedel1043:
Changes the return value of `JsObject::from_proto_and_data_with_shared_shape` to not upcast to `JsObject`, which in some cases allows storing `JsObject<Promise>` directly instead of having to downcast.

This process could also be done with the rest of the builtins, especially the iterator objects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ tests_wpt
 *.events
 chrome_profiler.json
 *.mm_profdata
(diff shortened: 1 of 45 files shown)


HalidOdat (commented):
Nice change! Looks good to me! :)

HalidOdat (approved):

PR #4505, opened 2025-10-27
Optimize `JsValue::to_property_key` operation

jedel1043:
Just some small optimizations around `JsValue::to_property_key`, which is one of our hot spots on the benchmarks.

### Main results

RESULT Richards 237
RESULT DeltaBlue 243
RESULT Crypto 192
RESULT RayTrace 524
RESULT EarleyBoyer 618
RESULT RegExp 87.6
RESULT Splay 877
RESULT NavierStokes 427
SCORE 323

### PR results

RESULT Richards 232
RESULT DeltaBlue 239
RESULT Crypto 200
RESULT RayTrace 524
RESULT EarleyBoyer 623
RESULT RegExp 91.8
RESULT Splay 862
RESULT NavierStokes 467
SCORE 328


core/engine/src/value/mod.rs
```diff
@@ -595,46 +556,51 @@ impl JsValue {
...
-                    JsVariant::Symbol(symbol) => symbol.clone().into(),
-                    JsVariant::Integer32(integer) => integer.into(),
-                    _ => primitive.to_string(context)?.into(),
-                }
-            }
-            _ => self.to_string(context)?.into(),
-        })
+        match self.variant() {
+            // fast path
+            JsVariant::Integer32(integer) => Ok(integer.into()),
```
nekevss, line 608: thought: is it worth writing an internal note here and elsewhere on optimizations (beyond fast path and slow path)

Generally, I think it could be useful to document optimizations so they aren't accidentally changed in the future and easier to identify and maintain. Although, I'd concede in this case that fast path vs. slow path may be enough in this specific circumstance.

jedel1043: Added something there

nekevss (approved):
Looks good to me!

PR #4507, opened 2025-10-28
Remove `environments` and `realm` fields from the Vm

jedel1043:
Since those are accessible from the active `frame` field, we don't need to store them twice.


nekevss (approved):
Looks good!

PR #4511, opened 2025-10-29
Implement proposal nonextensible-applies-to-private 

jedel1043:
Implements the [`nonextensible-applies-to-private` proposal](https://github.com/tc39/proposal-nonextensible-applies-to-private).

Since it's such a small change and it's already on stage 3, I decided to not gate this behind the experimental feature.

### DIFF
--- a/core/engine/src/object/operations.rs
+++ b/core/engine/src/object/operations.rs
@@ -891,8 +891,7 @@ impl JsObject {
         if !self.extensible() {
             return Err(js_error!(
                 TypeError: "cannot add private field to non-extensible class instance"
-            )
-            .into());
+            ));
         }
 
         // 2. If the host is a web browser, then
@@ -906,7 +905,7 @@ impl JsObject {
 
         // 4. If entry is not empty, throw a TypeError exception.
         if entry.is_some() {
-            return Err(js_error!(TypeError: "private field already exists on prototype").into());
+            return Err(js_error!(TypeError: "private field already exists on prototype"));
         }
 
         // 5. Append PrivateElement { [[Key]]: P, [[Kind]]: field, [[Value]]: value } to O.[[PrivateElements]].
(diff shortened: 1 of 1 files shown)


core/engine/src/object/operations.rs
```diff
@@ -887,23 +888,27 @@ impl JsObject {
         value: JsValue,
         context: &mut Context,
     ) -> JsResult<()> {
-        // 1. If the host is a web browser, then
-        // a. Perform ? HostEnsureCanAddPrivateElement(O).
+        if !self.extensible() {
```
HalidOdat: Suggestion: It would be nice if we added the spec step to know where this came from. :smile: 
```suggestion
        // 1. If O.[[Extensible]] is false, throw a TypeError exception.
        //
        // NOTE: From <https://tc39.es/proposal-nonextensible-applies-to-private/#sec-privatefieldadd>
        if !self.extensible() {
```

core/engine/src/object/operations.rs
```diff
@@ -931,12 +936,25 @@ impl JsObject {
...
             PrivateElement::Method(_) | PrivateElement::Accessor { .. }
         ));
+
         let (getter, setter) = if let PrivateElement::Accessor { getter, setter } = method {
             (getter.is_some(), setter.is_some())
         } else {
             (false, false)
         };
 
+        if !self.extensible() {
```
HalidOdat: Suggestion: Same here:

```suggestion
        // 2. If O.[[Extensible]] is false, throw a TypeError exception.
        //
        // NOTE: From <https://tc39.es/proposal-nonextensible-applies-to-private/#sec-privatemethodoraccessoradd>
        if !self.extensible() {
```

core/engine/src/object/operations.rs
```diff
@@ -931,12 +936,25 @@ impl JsObject {
...
+            return if getter || setter {
+                Err(js_error!(
+                    TypeError: "cannot add private accessor to non-extensible class instance"
+                ))
+            } else {
+                Err(js_error!(
+                    TypeError: "cannot add private method to non-extensible class instance"
+                ))
+            };
```
HalidOdat, line 965: Praise: I really like the error message Improvement :smile:

HalidOdat (approved):
Looks great! Just some minor nitpicks :smile:

linusg:
Note that this changed slightly in https://github.com/tc39/proposal-nonextensible-applies-to-private/commit/52e3ccca7db513cf096f53a39f3752a01f7f159c :)

PR #4512, opened 2025-10-30
Fix typos across the tree

linkmauve:
This has been done with the help of the wonderful [typos](https://github.com/crate-ci/typos) tool, and manual review of each individual change.  A few changes are my own as well.

This Pull Request fixes/closes no issue.

It changes the following:
- Many typos have been fixed to be proper English.
- Of course not all of them.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -153,6 +153,17 @@ jobs:
       - name: Format (rustfmt)
         run: cargo fmt --all --check
 
+  typos:
+    name: Typos
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+
+      - name: Check for typos
+        uses: crate-ci/typos@v1.38.1
+
   clippy:
     name: Lint
     runs-on: ubuntu-latest
--- a/typos.toml
+++ b/typos.toml
@@ -1,0 +1,11 @@
+[files]
+extend-exclude = ["CHANGELOG.md"]
+
+[default.extend-words]
+ba = "ba"
+beng = "beng"
+fo = "fo"
+forin = "forin"
+mis = "mis"
+pn = "pn"
+tru = "tru"
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ For more information on Boa's API, feel free to check out our documentation.
 ## 🏅 Conformance
 
 To know more details about Boa's conformance surrounding the _ECMAScript_ specification,
-you can check out our _ECMASCript Test262_ test suite results [here](https://boajs.dev/conformance).
+you can check out our _ECMAScript Test262_ test suite results [here](https://boajs.dev/conformance).
 
 ## 🪚 Contributing
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -20,7 +20,7 @@ your CLI.
 
 Once installed, your good to go!
 
-To execute some JavaScript source code, navigate to the directy of your choosing and type:
+To execute some JavaScript source code, navigate to the directory of your choice and type:
 
 ```shell
     boa test.js
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -596,7 +596,7 @@ impl Executor {
 
         self.promise_jobs.borrow().is_empty()
             && self.async_jobs.borrow().is_empty()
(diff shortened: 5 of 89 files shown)


nekevss:
Thanks for fixing these!

We should make it a point to check for types in CI or have a cargo-make recipe for it. How easy is it to work with typo? I think I've heard there's a GitHub action for it, is it relatively nice to use?

linkmauve:
> Thanks for fixing these!
> 
> We should make it a point to check for types in CI or have a cargo-make recipe for it. How easy is it to work with typo? I think I've heard there's a GitHub action for it, is it relatively nice to use?

Running typos is very easy, you run the `typos` command and if it returns 2 (or any non-zero status code) it means there are newer typos.

I’ve added a `typos.toml` file which excludes all false positives I kept in the previous commit.

I have no experience with github actions though.

Also, why is the semver check failing?

jedel1043:
> why is the semver check failing?

Ignore that for now, we had some breakage with the release of ICU4X 2.1

linkmauve:
I’ve added a GitHub Actions job, and a (temporary) commit adding a typo to check that it indeed fails there, but it doesn’t seem to run here, do you have to do anything on your side to allow for a new job in CI?

.github/workflows/rust.yml
```diff
@@ -153,6 +153,17 @@ jobs:
...
 
+  typos:
+    name: Check for typos
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+
+      - name: Install Rust toolchain
```
jedel1043: Maybe not the best name for this 😅

typos.toml
```diff
@@ -0,0 +1,11 @@
...
+ba = "ba"
+beng = "beng"
+fo = "fo"
+forin = "forin"
+littl = "littl"
+mis = "mis"
+neccesary = "neccesary"
+oll = "oll"
+pn = "pn"
+tru = "tru"
```
jedel1043, line 11: Why ignore those specific words? `forin` makes sense, since it's a JS construct, but things like "neccesary" or "littl" sound like things we want to check.

linkmauve: "neccesary" is part of a ChangeLog entry about fixing that very typo, "littl" is part of the name of someone in the ChangeLog as well:
```
- Changes neccesary -> necessary by @nekevss in [#2370](https://github.com/boa-dev/boa/pull/2370)
[…]
- [FEATURE #1265](https://github.com/boa-dev/boa/pull/1265): Implement `BigInt.prototype[ @@toStringTag ]` (@n14littl)
```

And the other non-typos are part of the code. You can remove them from this file locally, run `typos` again, and see where they are used in the code.

.github/workflows/rust.yml
linkmauve: Oops, fixed.

linkmauve:
CI does fail when there is a new typo! https://github.com/boa-dev/boa/actions/runs/18965726581/job/54191104595?pr=4512

I’ve just removed the commit introducing said typo, to check for the happy path now.

Edit: I’ve done #4517 and #4518 to fix two of the other CI job failures.

typos.toml
jedel1043: Can we ignore checking typos in those specific files instead? I wouldn't want to avoid checking those words just because there are legit usages somewhere else.

linkmauve: I don’t think so, or at least I haven’t found how in the [typos reference](https://github.com/crate-ci/typos/blob/master/docs/reference.md).

jedel1043: It seems like you can:
https://github.com/crate-ci/typos/blob/master/docs/reference.md#filesextend-exclude
It's under the `[files]` section

CHANGELOG.md
jedel1043, line 1: I would suggest to avoid editing the commit names. It's unfortunate if we did miss a typo when merging a PR, but I would expect the changelog to still show the original name of each PR. This also ties with my comment about ignoring this file for typos.

typos.toml
linkmauve: I’ve now ignored CHANGELOG.md instead.

CHANGELOG.md
linkmauve: Done.

jedel1043 (approved):
Thank you very much for the contribution!

Issue #4181, opened 2025-02-26
Audit current usage of unsafe


jedel1043:
We have a significant amount of unsafe in our codebase. We should audit all of the current unsafe usage and determine if unsafe is even required in those cases.

alabhyajindal:
I would like to help here. I can start by creating a markdown file listing the files and line numbers where the `unsafe` keyword is used, and then we all could pitch in to determine which one of those could be improved? Just a suggestion - please let me know if you would like this done any other way!

jedel1043:
@alabhyajindal Awesome! I like that plan. I'll assign the issue to you.

alabhyajindal:
I've created a file listing all `unsafe` keywords using the following ripgrep command:

```
rg --line-number --heading --context=2 "unsafe" --type rust > unsafe_occurrences.txt
```

[unsafe_occurrences.txt](https://github.com/user-attachments/files/23265854/unsafe_occurrences.txt)

sarthakNITT:
@jedel1043 Is this issue still open for contributions?

jedel1043:
Sure, you can go through the list that @alabhyajindal kindly provided to check if any of those can be removed with some clever hacking

sarthakNITT:
Sure. working on it.

sarthakNITT:
After going through the list, I found 2 unsafe usages that can be safely removed: 
- `core/parser/src/lexer/number.rs:394`- `str::from_utf8_unchecked` can be replaced with safe `str::from_utf8().expect(...)` since `buf` only ever contains ASCII bytes.
- `core/engine/src/builtins/date/utils.rs:763`- same pattern; ASCII is already checked on the line above, so `from_utf8_unchecked` is unnecessary.

I'll open a PR fixing these two. The rest of the unsafe I looked at appear to be genuinely required.

0xShubhamSolanki:
Hi @jedel1043, following up on the audit in #418. After seeing the discussion in #4781, it seems to be clear that performance-critical unsafe blocks are intentional and shouldn't be replaced with slower safe alternatives.However, many of these blocks currently lack the required // SAFETY: documentation to justify their soundness. Since I've recently been working on boa_gc for #3239, I would like to focus my audit specifically on the boa_gc crate.

My plan is to:
1.Identify any unsafe blocks that can be refactored without performance regressions.
2.For the intentional unsafe blocks, document the safety guarantees to ensure long-term maintainability.

Does starting with a focused audit of boa_gc sound like a good direction?

PR #4516, opened 2025-10-31
Add file listing unsafe occurrences

alabhyajindal:
Addresses #4181.

The PR adds a file `unsafe_occurrences.txt` to the project root. It lists files where the `unsafe` keyword is being used, grouped by files in which they occur. Line numbers, and surrounding code near the keyword is included for context. 

Ripgrep was used to generate the file, using the following command:

```
rg --line-number --heading --context=2 "unsafe" --type rust > unsafe_occurrences.txt
```


jedel1043:
I would say it would be preferable to push this file as a comment under the issue.

alabhyajindal:
Okay - thanks!

PR #4513, opened 2025-10-31
Bump crates' versions to 1.0.0

jedel1043:
Since we already made a breaking change, this bumps all our crates' versions to 1.0.0. This is also a commitment to start working towards releasing a 1.0.0, which many people will be happy about :)

- Fixes clippy lints.
- Bumps our MSRV to 1.91 for future required changes.

PR #4517, opened 2025-10-31
Replace manual `Default` impls with `#[derive(Default)]`

linkmauve:
This Pull Request fixes/closes no issue.

It changes the following:
- Remove an unneeded `.to_string()`.
- Replace a bunch of manual `impl Default`s with `#[derive(Default)]`
- Make explicit the failure mode of a `Duration` subtraction.

### DIFF
--- a/core/engine/src/context/time.rs
+++ b/core/engine/src/context/time.rs
@@ -123,7 +123,10 @@ impl std::ops::Sub for JsInstant {
     #[inline]
     fn sub(self, rhs: JsInstant) -> Self::Output {
         JsDuration {
-            inner: self.inner - rhs.inner,
+            inner: self
+                .inner
+                .checked_sub(rhs.inner)
+                .expect("overflow when subtracting durations"),
         }
     }
 }


jedel1043:
This was already fixed by #4513

linkmauve:
I’ve rebased on top of main, there are still changes worth it in this PR, since I’m running clippy nightly locally.

core/engine/src/context/time.rs
```diff
@@ -123,7 +123,7 @@ impl std::ops::Sub for JsInstant {
     #[inline]
     fn sub(self, rhs: JsInstant) -> Self::Output {
         JsDuration {
-            inner: self.inner - rhs.inner,
+            inner: self.inner.checked_sub(rhs.inner).unwrap(),
```
HalidOdat: I'm fine with the panic in case of overflow ([since that is what impl of `Sub` `std::time::Duration` does](https://doc.rust-lang.org/stable/src/core/time.rs.html#1165)), but we should use `expect` instead.

```suggestion
            inner: self.inner.checked_sub(rhs.inner).expect("overflow when subtracting durations"),
```

linkmauve: Thanks @HalidOdat, I’ve applied your suggestion. :)

PR #4523, opened 2025-11-04
Unflag temporal as experimental and turn it on by default

nekevss:
This Pull Request removes temporal as an experimental feature and moves it to default.

Boa's implementation is generally complete. We have made the Time Zone Provider available for configuration on the context, and small changes should be needed moving forward for the implementation.

### DIFF
--- a/core/engine/Cargo.toml
+++ b/core/engine/Cargo.toml
@@ -12,7 +12,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [features]
-default = ["float16", "xsum"]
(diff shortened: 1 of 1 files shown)


nekevss:
Open question: would we consider this PR as grounds for closing #1804, even with some outstanding test failures and bugs that may need to be fixed?

jasonwilliams (approved):

jasonwilliams:
> Open question: would we consider this PR as grounds for closing #1804, even with some outstanding test failures and bugs that may need to be fixed?

Yes I think we have enough to close that issue now

Issue #4519, opened 2025-11-02
Confused by Scope API


rix0rrr:
Just to dabble, I'm trying to build my own bytecode VM on top of the Boa parser (to skip past the annoying things and get to the good stuff!).

I did expressions and at this point I'm getting to doing variables. It feels like I should be using the `Scope` API for that, but it's a little unclear to me how it works. 

I've parsed the AST and I'm now traversing it, querying `Scope` objects when I'm encountering identifiers.

## Is this even the right API?

First off, it feels to weird to me that I'm getting `Sym`s (interned strings) out of the AST, but then have to pass `JsString` objects into the `Scope` API rather than those same interned strings. Seems a bit weird to have to reverse the interned string and then build a new class out of it, if all we need is a unique name that we already have access to?

## Can't get to the bindings

I would like to access a `Scope`'s `bindings` for example, or its `unique_id` but none of those are exposed, so I can't. 

## I don't understand the numbering rules

Right, as I'm traversing the AST, I'm keeping a stack of `Scope`s that I push and pop whenever I'm entering/leaving a scope. For example:

```rust
            Statement::Block(block) => {
                self.enter_scope(block.scope().unwrap().clone());

                self.compile_statement_list(block.statement_list())?;

                self.leave_scope();

                Ok(())
            }
```

Then, when I encounter a variable I call `get_identifier_reference` on the closest scope to resolve that identifier:

```rust
                let locator = self
                    .current_scope()
                    .scope
                    .get_identifier_reference(ident.sym().to_js_string(&self.interner));
```

Now I need to know in which of the scopes on my stack that variable was found in, so that I know where on my stack my local variable slot will live(*). 

> (*) Haven't gotten to closures yet, I know my model will have to change a little once I start doing that.

If I could get the `unique_id` out of a scope I could check my stack to find it, but as we already saw that's not accessible. Instead, I can get the `BindingLocatorScope` (which is based off of of the `self.scope` member) which seems to do something similar, but it seems to have weird values.

If I send in the following script:

```js
var x = 3;
{
  let y = 4;
  {
    let z = 1;
    x * y *z
  }
}
```

And I call `get_identifier_reference` on every variable declaration and print it, I see the following:

```
IdentifierReference { locator: BindingLocator { name: "x", scope: 0, binding_index: 0, unique_scope_id: 0 }, lexical: false, escapes: true }
[...]

PR #4527, opened 2025-11-09
fix: make Scope's unique_id truly unique and add a getter for it

rix0rrr:
Right now, `Scope::unique_id` is always the same as the parent Scope's `index + 1`. This means that if two different scopes are created as children of the same scope, they will have the same `unique_id`, belying its name.

In this PR, add a new `ScopeContext` struct that all Scopes in an AST share; this new struct contains a container which controls the identifiers that are given out to scopes in the tree.

Also make the field accessible via a getter function, as well as on `BindingLocator`, so that `BindingLocator` and `Scope` can be correlated.

Relates to #4519.

### DIFF
--- a/core/ast/src/scope.rs
+++ b/core/ast/src/scope.rs
@@ -105,6 +105,8 @@ pub(crate) struct Inner {
     function: bool,
     // Has the `this` been accessed/escaped outside the function environment boundary.
     this_escaped: Cell<bool>,
+
+    context: Rc<ScopeContext>,
 }
 
 impl Scope {
@@ -119,6 +121,7 @@ impl Scope {
                 bindings: RefCell::default(),
                 function: true,
                 this_escaped: Cell::new(false),
+                context: Rc::default(),
             }),
         }
     }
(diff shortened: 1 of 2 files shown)


core/ast/src/scope.rs
```diff
@@ -637,7 +676,7 @@ pub enum BindingLocatorError {
 }
 
 /// The scope in which a binding is located.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
rix0rrr, line 676: Not strictly necessary for this PR, but came up in the context of writing some other unit test, and seems unobjectionable enough.

nekevss: I'm not necessarily against the derives. The tests should only require PartialEq though, right? Was there a test that needed Eq?

rix0rrr: No, but it seems this type is trivially `Eq` as well, so I didn't see a reason to leave it out. Would you rather not have it?

core/ast/src/scope.rs
```diff
@@ -105,6 +105,11 @@ pub(crate) struct Inner {
     function: bool,
     // Has the `this` been accessed/escaped outside the function environment boundary.
     this_escaped: Cell<bool>,
+
+    /// A counter for unique IDs for Scopes generated as part of this scope tree.
+    ///
+    /// Only used in the root scope.
+    unique_id_ctr: Cell<u32>,
```
HalidOdat: Suggestion: It might be cleaner to keep the Scope state separate for example, by introducing a `ScopeContext` that holds this data. It could then be passed into Scope::new, or each Scope could store an Rc<ScopeContext> (eliminates the root search for next unique id)

rix0rrr: Sure! It does simplify things a bunch. Made that change.

Issue #4515, opened 2025-10-31
Panic `The RuntimeLimit native error cannot be converted to an opaque type.` in `core/engine/src/error.rs`


qarmin:
Rust code
```
fn check_file(path: &str) {
    let Ok(file_content) = fs::read(path) else {
        return;
    };
    println!("Checking file: {path}");
    let mut context = Context::default();

    let _result = context.eval(Source::from_bytes(&file_content));
}
```
when evaluating javascript code(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
async function*asyncGeerator(){}l=asyncGeerator().return({get then(){this.then}})
```

crashes with this info:
```
thread 'main' (213719) panicked at /home/runner/.cargo/git/checkouts/boa-126d820dff07bf57/868d13f/core/engine/src/error.rs:1260:17:
The RuntimeLimit native error cannot be converted to an opaque type.
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/292be5c7c05138d753bbd4b30db7a3f1a5c914f7/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/292be5c7c05138d753bbd4b30db7a3f1a5c914f7/library/core/src/panicking.rs:80:14
   2: boa_engine::error::JsNativeError::to_opaque
             at /home/runner/.cargo/git/checkouts/boa-126d820dff07bf57/868d13f/core/engine/src/error.rs:1260:17
   3: boa_engine::error::JsError::to_opaque
             at /home/runner/.cargo/git/checkouts/boa-126d820dff07bf57/868d13f/core/engine/src/error.rs:377:34
   4: boa_engine::builtins::promise::Promise::create_resolving_functions::{{closure}}
             at /home/runner/.cargo/git/checkouts/boa-126d820dff07bf57/868d13f/core/engine/src/builtins/promise/mod.rs:2133:56
   5: <boa_engine::native_function::Closure<F,T> as boa_engine::native_function::TraceableClosure>::call
             at /home/runner/.cargo/git/checkouts/boa-126d820dff07bf57/868d13f/core/engine/src/native_function/mod.rs:69:9
   6: boa_engine::native_function::NativeFunction::call
             at /home/runner/.cargo/git/checkouts/boa-126d820dff07bf57/868d13f/core/engine/src/native_function/mod.rs:306:40
   7: boa_engine::native_function::native_function_call
[...]

PR #4534, opened 2025-11-14
Extract engine errors as their own `JsError` variant

jedel1043:
Related to #4515.

Extracts a new `EngineError` type that is a new variant of `JsError` that cannot be converted to a `JsValue` using `into_erased`.

This also makes `into_erased` fallible for `JsError`, which IMO is fine since we really want to make sure users handle engine errors gracefully.

EDIT: Unfortunately, this doesn't fully address #4515 because the stack overflows on the CI profile. Will open an issue related to that.

### DIFF
--- a/core/engine/src/vm/runtime_limits.rs
+++ b/core/engine/src/vm/runtime_limits.rs
@@ -19,7 +19,7 @@ impl Default for RuntimeLimits {
     fn default() -> Self {
         Self {
             loop_iteration: u64::MAX,
-            recursion: 2048,
+            recursion: 512,
             backtrace_limit: 50,
             stack_size: 1024 * 10,
         }
--- a/core/engine/src/vm/runtime_limits.rs
+++ b/core/engine/src/vm/runtime_limits.rs
@@ -19,7 +19,7 @@ impl Default for RuntimeLimits {
     fn default() -> Self {
         Self {
             loop_iteration: u64::MAX,
-            recursion: 512,
+            recursion: 2048,
             backtrace_limit: 50,
             stack_size: 1024 * 10,
         }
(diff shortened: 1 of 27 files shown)


core/engine/src/vm/tests.rs
```diff
@@ -498,3 +492,20 @@ fn long_object_chain_gc_trace_stack_overflow() {
         TestAction::inspect_context(|_| boa_gc::force_collect()),
     ]);
 }
+
+// See: https://github.com/boa-dev/boa/issues/4515
+#[test]
```
nekevss, line 497: thought: do we have a notation for marking these tests as TODO?

I think linking the issue is good, but it might be nice to add a `FIX:` or `TODO (fix):` prefix to tests like these so that they can be easily grepped. But maybe `#[ignore]` is enough.

core/engine/src/vm/runtime_limits.rs
```diff
@@ -19,7 +19,7 @@ impl Default for RuntimeLimits {
     fn default() -> Self {
         Self {
             loop_iteration: u64::MAX,
-            recursion: 512,
+            recursion: 2048,
```
nekevss: question: what's the reason for the bump on the recursion reason?

nekevss (approved):
Looks good to me. I had a couple comments but nothing blocking

core/engine/src/vm/runtime_limits.rs
jedel1043: Oops, that go through when I was testing things

core/engine/src/vm/tests.rs
jedel1043: Added a TODO comment

nekevss (approved):
Looks great!

HalidOdat (approved):
Looks good to me! :smile:

PR #4542, opened 2025-11-22
Add fast path for `parseInt` and `parseFloat`

HalidOdat:


core/engine/src/value/mod.rs
```diff
@@ -354,6 +354,13 @@ impl JsValue {
         self.0.is_integer32() || self.0.is_float64()
     }
 
+    /// Returns true if the value is a negative zero (`-0`).
+    #[inline]
+    #[must_use]
+    pub fn is_negative_zero(&self) -> bool {
```
nekevss: issue: I have a soft preference for this method to be kept pub(crate) vs. pub 

Does this method need to be added to the public API? I'm not sure that I'm convinced we should support this method in the public API currently vs. pub(crate). Mostly because `is_negative_zero` signals that there should be something like `is_nan` or `is_finite_number`.

HalidOdat: I wasn't sure if that was a good idea either, the only reason I did it was because it much faster than `as_number() + check`, Made it `pub(crate)` here: https://github.com/boa-dev/boa/pull/4542/commits/d355e47428ae831a111b1474416398afb113e7fa :smile:

core/engine/src/builtins/number/globals.rs
```diff
@@ -297,40 +303,71 @@ pub(crate) fn parse_float(
...
+        return Ok(JsValue::negative_infinity());
+    } else if let Some(LOWER_CASE_I_CHAR | UPPER_CASE_I_CHAR) = prefix.get(0) {
+        return Ok(JsValue::nan());
     }
+
+    let value = match trimmed_string.variant() {
+        JsStrVariant::Latin1(s) => fast_float2::parse_partial::<f64, _>(s),
+        JsStrVariant::Utf16(s) => {
+            let s = String::from_utf16_lossy(s);
+            fast_float2::parse_partial::<f64, _>(s.as_bytes())
```
nekevss, line 356: thought: it would be nice if `fast_float2` could parse directly from `UTF16`.

Maybe add a `NOTE` comment as a potential future optimization.

PR #4549, opened 2025-11-27
Implement Import Attributes proposal

### LINKED ISSUE #4440: Implement import attributes proposal
https://github.com/tc39/proposal-import-attributes

rrogerc:
This Pull Request closes #4440, implementing the [Import Attributes](https://github.com/tc39/proposal-import-attributes) proposal

It changes the following:
- Implemented the with clause (and assert fallback) for imports, re-exports, and dynamic imports.
- Updated ModuleLoader and ModuleRequest to support attributes in module resolution and caching.
- Added handling for type: "json" in SimpleModuleLoader to verify attribute behavior.

### DIFF
--- a/core/engine/src/vm/opcode/call/mod.rs
+++ b/core/engine/src/vm/opcode/call/mod.rs
@@ -532,7 +532,7 @@ impl ImportCall {
 
         // 3. Let argRef be ? Evaluation of AssignmentExpression.
         // 4. Let specifier be ? GetValue(argRef).
-        let arg = context.vm.get_register(specifier_op.into()).clone();
+        let specifier = context.vm.get_register(specifier_op.into()).clone();
 
         // Get options if provided
         let options = context.vm.get_register(options_op.into()).clone();
@@ -546,7 +546,7 @@ impl ImportCall {
         let promise = cap.promise().clone();
 
         // 6. Let specifierString be Completion(ToString(specifier)).
-        let specifier_str = match arg.to_string(context) {
+        let specifier_str = match specifier.to_string(context) {
             Ok(s) => s,
             // 7. IfAbruptRejectPromise(specifierString, promiseCapability).
             Err(err) => {


core/ast/src/declaration/mod.rs
```diff
@@ -156,3 +156,61 @@ impl VisitWith for ModuleSpecifier {
...
+
+/// An import attribute entry in an [`ImportDeclaration`].
+///
+/// This is a key-value pair, where the key is an identifier or string literal,
+/// and the value is a string literal.
+///
+/// More information:
+///  - [ECMAScript proposal][proposal]
+///
+/// [proposal]: https://tc39.es/proposal-import-attributes/#prod-ImportAttribute
```
nekevss: nit: link to the Import Attributes Record in the ECMA262 specification text

Personally, I think the actual proposal text is easier to parse for implementation purposes, so it makes sense that you're using it. But for documentation, you should link to the ECMA262 specification text (mostly because the proposal has been adopted and added to the specification)

core/parser/src/parser/statement/declaration/mod.rs
```diff
@@ -135,3 +135,130 @@ where
         Ok((*from).into())
     }
 }
+
+/// Parses an optional `with` clause for import attributes.
+///
+/// More information:
+///  - [ECMAScript Import Attributes proposal][proposal]
+///
+/// [proposal]: https://tc39.es/proposal-import-attributes/#prod-WithClause
```
nekevss: nit: re: use test262 specification link vs. proposal link

core/engine/src/module/mod.rs
```diff
@@ -57,6 +57,62 @@ mod namespace;
 mod source;
 mod synthetic;
 
+/// A module request with optional import attributes.
+///
+/// Represents a module specifier and its associated import attributes.
+/// According to the [Import Attributes proposal][proposal], the module cache key
+/// should be (referrer, specifier, attributes).
+///
+/// [proposal]: https://tc39.es/proposal-import-attributes/
```
nekevss: nit: link to test262 specification

core/engine/src/module/mod.rs
```diff
@@ -57,6 +57,62 @@ mod namespace;
...
+    pub fn new(specifier: JsString, mut attributes: Box<[(JsString, JsString)]>) -> Self {
+        // Sort attributes by key to ensure canonical cache keys.
+        attributes.sort_unstable_by(|(k1, _), (k2, _)| k1.cmp(k2));
+        Self {
+            specifier,
+            attributes,
+        }
+    }
+
+    /// Creates a new module request from only a specifier, with no attributes.
```
nekevss: nit: remove comma

core/engine/src/module/source.rs
```diff
@@ -241,14 +241,77 @@ impl std::fmt::Debug for SourceTextModule {
...
+    type BreakTy = std::convert::Infallible;
+
+    fn visit_import_declaration(
+        &mut self,
+        node: &'ast boa_ast::declaration::ImportDeclaration,
+    ) -> std::ops::ControlFlow<Self::BreakTy> {
+        let specifier = node.specifier().sym().to_js_string(self.interner);
+        let attributes = node
+            .attributes()
+            .iter()
```
nekevss: thought: this `iter().map(|attr|(JsString, JsString)).collect()` happens multiple times in the code base and may benefit from a new type

I think I've seen it at least 3 or 4 times, so it's definitely a candidate. But I'll leave that up to you all

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,74 @@ impl Operation for CallSpread {
     const COST: u8 = 3;
 }
 
+/// Parses the import attributes from the options object.
+fn parse_import_attributes(
+    specifier: JsString,
+    options: JsValue,
+    context: &mut Context,
+) -> JsResult<crate::module::ModuleRequest> {
+    if options.is_undefined() {
```
nekevss: nit: document the specification steps being addressed here with comments as done elsewhere in the codebase

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -340,69 +386,49 @@ async fn load_dyn_import(
     // 3. Let loadPromise be module.LoadRequestedModules().
     let load = module.load(&mut context.borrow_mut());
 
-    // 4. Let rejectedClosure be a new Abstract Closure with parameters (reason) that captures promiseCapability and performs the following steps when called:
```
nekevss, line 343: question: why are the specification steps being removed?

Are these no longer applicable? If they're not applicable, what are the new steps?

core/parser/src/parser/statement/declaration/mod.rs
```diff
@@ -135,3 +135,130 @@ where
...
+        };
+
+        let is_with = matches!(tok.kind(), TokenKind::Keyword((Keyword::With, _)));
+        let is_assert = if is_with {
+            false
+        } else {
+            matches!(tok.kind(), TokenKind::IdentifierName((sym, _)) if interner.resolve_expect(*sym).utf8().is_some_and(|s| s == "assert"))
+        };
+
+        // Only treat `with` / `assert` as part of a with-clause if it is
```
nekevss: question: why is `assert` still supported?

I only took a quick glance at the specification and it looks like assert was removed from the proposal.

nekevss (commented):
There's still probably some more to review, but thought I'd leave some early review comments on this.

Overall, great work on this so far! Mostly had just some general nits and questions for now.

core/parser/src/parser/statement/declaration/mod.rs
rrogerc: Oh, if you look at the bottom of the [proposal](https://github.com/tc39/proposal-import-attributes), under history it says "For compatibility with existing implementations, the assert keyword will still be supported until it's safe to remove it, if it will ever be."

Though let me know if I’ve misunderstood though!

core/engine/src/vm/opcode/call/mod.rs
rrogerc: Oh, sorry about that, I think that was a mistake on our end. I've restored the specification comments in load_dyn_import and ImportCall::operation.

Thanks for spotting this!

core/ast/src/declaration/mod.rs
rrogerc: Thanks! I've updated it to link to the official specification.

I used the link for the Imports section (#sec-imports) to match the existing convention in ImportKind and to ensure the link remains stable, since the specific anchors for the Record definition were not loading reliably.

core/engine/src/module/source.rs
rrogerc: Oh, that makes sense, I agree that centralizing would make the logic a lot cleaner. 

I’ve extracted it into a new ModuleRequest::from_ast helper method!

core/engine/src/module/mod.rs
```diff
@@ -57,6 +60,82 @@ mod namespace;
...
+///
+/// Represents a module specifier and its associated import attributes.
+/// According to the [ECMAScript specification][spec], the module cache key
+/// should be (referrer, specifier, attributes).
+///
+/// [spec]: https://tc39.es/ecma262/#sec-modulerequest-record
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Trace, Finalize)]
+pub struct ModuleRequest {
+    specifier: JsString,
+    attributes: Box<[(JsString, JsString)]>,
```
HalidOdat: Suggestion: I think it would be better to match the spec more by adding a `ImportAttribute`.

```rust
/// /* ... */
///
/// [spec]: https://tc39.es/ecma262/#table-importattribute-fields
struct ImportAttribute {
    key: JsString,
    value: JsString,
}
```

To differentiate the AST import we can rename it to `AstImportAttribute`:
```rust
use boa_ast::declaration::ImportAttribute as AstImportAttribute;
```

core/engine/src/module/loader/mod.rs
```diff
@@ -178,7 +178,7 @@ pub trait ModuleLoader: Any {
     async fn load_imported_module(
         self: Rc<Self>,
         referrer: Referrer,
-        specifier: JsString,
+        request: super::ModuleRequest,
```
HalidOdat: Nitpick:

```suggestion
        request: ModuleRequest,
```

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+    context: &mut Context,
+) -> JsResult<crate::module::ModuleRequest> {
+    // Taken from `EvaluateImportCall`
+    //
+    // <https://tc39.es/proposal-import-attributes/#sec-evaluate-import-call>
+
+    // 1. Let attributes be a new empty List.
+    // 2. If options is not undefined, then
+    if options.is_undefined() {
+        return Ok(crate::module::ModuleRequest::from_specifier(specifier));
```
HalidOdat: Nitpick:
```suggestion
        return Ok(ModuleRequest::from_specifier(specifier));
```

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+    }
+
+    // a. If Type(options) is not Object, throw a TypeError exception.
+    if let Some(options_obj) = options.as_object() {
+        // b. Let attributesObj be ? Get(options, "with").
+        let with_key = crate::js_string!("with");
+        let with_val = options_obj.get(with_key, context)?;
```
HalidOdat: Suggestion: Try matching the names of the spec steps as close as possible makes it easier adapt the spec changes and easier to review :smile:  
```suggestion
        let attribute_obj = options_obj.get(js_str!("with"), context)?;
```

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+            (assert_val, true)
+        } else {
+            (with_val, false)
+        };
+
+        // d. If attributesObj is not undefined, then
+        if attributes_val.is_undefined() {
+            Ok(crate::module::ModuleRequest::from_specifier(specifier))
+        // i. If Type(attributesObj) is not Object, throw a TypeError exception.
+        } else if let Some(attributes_obj) = attributes_val.as_object() {
```
HalidOdat: Remark: You could use let-else statement in this case, would make the spec steps follow better.

```suggestion
        // i. If Type(attributesObj) is not Object, throw a TypeError exception.
        let Some(attributes_obj) = attributes_val.as_object() else {
            // .. throw here ..
        };
```

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+    // <https://tc39.es/proposal-import-attributes/#sec-evaluate-import-call>
+
+    // 1. Let attributes be a new empty List.
+    // 2. If options is not undefined, then
+    if options.is_undefined() {
+        return Ok(crate::module::ModuleRequest::from_specifier(specifier));
+    }
+
+    // a. If Type(options) is not Object, throw a TypeError exception.
+    if let Some(options_obj) = options.as_object() {
```
HalidOdat: Remark: Same here we could use let-else statement.

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+        let (attributes_val, is_assert) = if with_val.is_undefined() {
+            let assert_key = crate::js_string!("assert");
+            let assert_val = options_obj.get(assert_key, context)?;
+            (assert_val, true)
+        } else {
+            (with_val, false)
+        };
+
+        // d. If attributesObj is not undefined, then
+        if attributes_val.is_undefined() {
```
HalidOdat: Suggestion: The spec says "is _not_ undefined" which makes the if condition seem like it's not conforming to the spec (but it actually is because the else and if body have been swapped).

In general try to match the spec as close as possible places where there is a need to deviate (i.e. performance, ease of implementation, etc) should be noted with some sort of `NOTE: ... `, just to make it clear for anyone who is sees this code.

In this case we can conform to the spec structure, without degrading performance :smile:

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+            Ok(crate::module::ModuleRequest::from_specifier(specifier))
+        // i. If Type(attributesObj) is not Object, throw a TypeError exception.
+        } else if let Some(attributes_obj) = attributes_val.as_object() {
+            // ii. Let entries be ? EnumerableOwnProperties(attributesObj, "key+value").
+            let keys = attributes_obj
```
HalidOdat: Remark:
```suggestion
            let entries = attributes_obj
```

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
 
+/// Parses the import attributes from the options object.
+fn parse_import_attributes(
+    specifier: JsString,
+    options: &JsValue,
+    context: &mut Context,
+) -> JsResult<crate::module::ModuleRequest> {
+    // Taken from `EvaluateImportCall`
+    //
+    // <https://tc39.es/proposal-import-attributes/#sec-evaluate-import-call>
```
HalidOdat: Important: Since the it's already in the spec, I'd suggest using the spec links and steps instead of the proposal: <https://tc39.es/ecma262/#sec-evaluate-import-call>

The main spec will change over time but the proposal will not.

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+        };
+
+        // d. If attributesObj is not undefined, then
+        if attributes_val.is_undefined() {
+            Ok(crate::module::ModuleRequest::from_specifier(specifier))
+        // i. If Type(attributesObj) is not Object, throw a TypeError exception.
+        } else if let Some(attributes_obj) = attributes_val.as_object() {
+            // ii. Let entries be ? EnumerableOwnProperties(attributesObj, "key+value").
+            let keys = attributes_obj
+                .enumerable_own_property_names(crate::property::PropertyNameKind::Key, context)?;
```
HalidOdat: Important: The spec says that it should be keys and values

```suggestion
          .enumerable_own_property_names(crate::property::PropertyNameKind::KeyAndValue, context)?;
```

The below code does get the value, and my guess this is this was an optimization?, but the issue is that the order of operations has changed and they can have side effects.

spec: `[[GetOwnProperty]], Get(), [[GetOwnProperty]], Get(), ...`
spec: `[[GetOwnProperty]], [[GetOwnProperty]], ... ,  Get(), Get(), ...`

If the options (is an object that is guaranteed to to be a `Proxy`, the I guess this would be fine, and we should note that deviation of the spec with a `OPTIMIZATION: `... reason why this was done ...`

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+
+            // iii. For each entry in entries, do
+            let mut attributes = Vec::with_capacity(keys.len());
+            for key in keys {
+                // 1. Let key be entry.[[Key]].
+                if !key.is_string() {
+                    continue;
+                }
+                let key_str = key.as_string().expect("key must be string").clone();
```
HalidOdat: Important: You could use let-else, if the key value being anything other than string is an impossible state then we should assert it.

```suggestion
                // 1. Let key be entry.[[Key]].
                let Some(key) = key.as_string() else {
                    unreachable!("key from enumerable_own_property_names must always be a string");
                };
```

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
...
+                let value = attributes_obj.get(key_str.clone(), context)?;
+
+                // 3. If Type(value) is not String, throw a TypeError exception.
+                if !value.is_string() {
+                    return Err(JsNativeError::typ()
+                        .with_message("import attribute value must be a string")
+                        .into());
+                }
+
+                let value_str = value.as_string().expect("value must be string").clone();
```
HalidOdat: Suggestion: Some let-else avoids the except here :smile:

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -258,6 +258,92 @@ impl Operation for CallSpread {
     const COST: u8 = 3;
 }
 
+/// Parses the import attributes from the options object.
+fn parse_import_attributes(
```
HalidOdat: Suggestion: If this function implements the `https://tc39.es/ecma262/#sec-evaluate-import-call` (` EvaluateImportCall `), then we should call it evaluate_import_call

core/engine/src/module/mod.rs
```diff
@@ -57,6 +60,82 @@ mod namespace;
...
+            attributes,
+        }
+    }
+
+    /// Creates a new module request from only a specifier with no attributes.
+    #[must_use]
+    pub fn from_specifier(specifier: JsString) -> Self {
+        Self {
+            specifier,
+            attributes: Box::new([]),
```
HalidOdat: Nitpick:
```suggestion
            attributes: Box::default(),
```

core/parser/src/parser/statement/declaration/mod.rs
nekevss: Based off tc39/proposal-import-attributes#135, I would go with we should only support `with` out of the box.

core/engine/src/module/mod.rs
rrogerc: Yep, I've implemented your suggestion!

core/engine/src/vm/opcode/call/mod.rs
rrogerc: Yep, I've refactored the function to use nested if blocks!

core/parser/src/parser/statement/declaration/mod.rs
rrogerc: Thanks for the link, we've updated it to remove it entirely!

nekevss (approved):
LGTM!

Thanks for adding all the tests for this as well 😄

nekevss:
Thanks for contributing this!!!

PR #4551, opened 2025-11-30
Implement Performance.now() and Performance.timeOrigin

### LINKED ISSUE #4491: Feature: Performance APIs (e.g. `performance.now()`)
https://developer.mozilla.org/en-US/docs/Web/API/Performance_API
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

**Example code**

```js
const start = performance.now()
slowWork()
const finish = performance.now()
console.log(finish - start)
```

HalidOdat:
Adds implementation of the W3C High Resolution Time API's `performance.now()` method and `performance.timeOrigin` getter property.

Some benchmarks use this if available, which hopefully will remove the very big fluctuations in performance measurement.

### DIFF
--- a/core/runtime/src/performance/mod.rs
+++ b/core/runtime/src/performance/mod.rs
@@ -1,0 +1,152 @@
+//! Boa's implementation of the `performance` Web API object.
+//!
+//! The `performance` object provides access to performance-related information.
+//!
+//! More information:
+//!  - [MDN documentation][mdn]
+//!  - [W3C High Resolution Time specification][spec]
+//!
+//! [spec]: https://w3c.github.io/hr-time/
+//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/API/Performance
+
+use boa_engine::{
+    Context, JsData, JsNativeError, JsResult, JsValue, NativeFunction,
+    context::time::JsInstant,
+    js_string,
+    object::{FunctionObjectBuilder, ObjectInitializer},
+    property::Attribute,
+};
+use boa_gc::{Finalize, Trace};
+
+#[cfg(test)]
+mod tests;
+
+/// The `Performance` object.
+#[derive(Debug, Trace, Finalize, JsData)]
+pub struct Performance {
+    #[unsafe_ignore_trace]
+    time_origin: JsInstant,
+}
+
+impl Performance {
+    /// Create a new `Performance` object.
+    #[must_use]
+    pub fn new(context: &Context) -> Self {
+        Self {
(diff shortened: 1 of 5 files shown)


HalidOdat:
Ideally this should inherit from `EventTarget` (not implemented) have some work on that in a separate PR.

core/runtime/src/performance/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+mod tests;
+
+/// The `Performance` object.
+#[derive(Debug, Trace, Finalize, JsData)]
+pub struct Performance {
+    #[unsafe_ignore_trace]
+    time_origin: JsInstant,
+}
+
+impl Performance {
```
hansl, line 31: Any reason you don't use the `boa_class` proc macro here? It would significantly reduce the code.

hansl:
> Ideally this should inherit from EventTarget

@HalidOdat I was starting work on class inheritance with that specifically in mind, but hit a road block and inheritance in Rust currently is broken. Do you have pointers to help?

HalidOdat:
Oh didn't know you where working on that... the way I implemented it was on the `boa_class` by adding a `#[boa(inherit)]` that can be added to a method that takes context and returns `JsResult<JsObject>` (returns the prototype to be inherited) (see [here](https://github.com/boa-dev/boa/compare/main...HalidOdat:boa:feature/runtime-event?expand=1#diff-4d94bf3ffcb058be7ca9088509b13c7880f0610aa46e513fb1df68554f0edb7eR240)).

Since we can attach custom host fields on the realm, I was thinking of adding an `RuntimeIntrinsics` ([like we do on boa_engine](https://github.com/boa-dev/boa/blob/main/core/engine/src/context/intrinsics.rs#L25)) and that can be accessed (through the context) and we can get the prototype. This avoids having to do `globalThis` gets.

hansl:
Having the inherit being an attribute on the class is more natural and closer to what JavaScript really does. Here's my PR: https://github.com/boa-dev/boa/pull/4431/changes#diff-37970e7d817d4316a093bb967d74ff7e26d9c6c01d93f5a79d3d195ba9f6155bR389

Also this allows for JS classes to inherit Rust classes which inherit JS classes, etc.

hansl:
I think the main blocker is that Rust's JsObject prototype stuff is broken and doesn't work. I could get it to work for static methods but not instance, for some reason. We also need to figure out a way to call `super` from the base class (I think I had `JsSuper` as an argument in that PR).

PR #4552, opened 2025-11-30
Implement `Error.prototype.stack` accessor property

HalidOdat:
This implements the `Error.prototype.stack` property as specified in the TC39 Error Stacks proposal (https://tc39.es/proposal-error-stacks).

### DIFF
--- a/core/engine/src/builtins/error/mod.rs
+++ b/core/engine/src/builtins/error/mod.rs
@@ -10,6 +10,8 @@
 //! [spec]: https://tc39.es/ecma262/#sec-error-objects
 //! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
 
+use std::fmt::Write;
+
 use crate::{
     Context, JsArgs, JsData, JsResult, JsString, JsValue,
     builtins::BuiltInObject,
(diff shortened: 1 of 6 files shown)


core/engine/src/builtins/error/mod.rs
```diff
@@ -263,6 +291,77 @@ impl Error {
...
+    ///
+    /// The accessor property of Error instances represents the stack trace
+    /// when the error was created.
+    ///
+    /// More information:
+    ///  - [Proposal][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-error-stacks/
+    #[allow(clippy::unnecessary_wraps)]
+    fn get_stack(this: &JsValue, _: &[JsValue], _context: &mut Context) -> JsResult<JsValue> {
```
jedel1043, line 304: We might need to gate the `get_stack` and `set_stack` methods behind the `experimental` feature until the proposal gets to stage 4

HalidOdat: Hmmm... I'm more inclined to leave it on by default, mainly because it's implemented and available by default by every engine (even before the proposal).

From the MDN doc on [`get/set .stack` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack):

> [!NOTE]
>
>**Note:** The stack property is de facto implemented by all major JavaScript engines, and [the JavaScript standards committee is looking to standardize it](https://github.com/tc39/proposal-error-stacks). You cannot rely on the precise content of the stack string due to implementation inconsistencies, but you can generally assume it exists and use it for debugging purposes.

What do you think? :smile:

jedel1043: Oh, if it is a standard API on engines then it makes sense to leave it ungated

PR #4553, opened 2025-11-30
Optimize JsStr::code_points() for Latin1 strings

Abd002:
This Pull Request optimizes the `code_points()` method for Latin1 strings.

It changes the following:

- Added `CodePointIter` enum with specialized Latin1 path
- Latin1 strings now avoid UTF-16 conversion overhead
- UTF-16 string behavior remains unchanged
- Added tests for both encoding paths


core/string/src/tests.rs
```diff
@@ -469,3 +469,19 @@ fn common_js_string_builder() {
...
 }
+
+#[test]
+fn code_points_optimization() {
+    // Test Latin1 optimization
+    let latin1_str = JsStr::latin1(b"Hello");
+    let latin1_points: Vec<CodePoint> = latin1_str.code_points().collect();
+    let expected_latin1: Vec<CodePoint> = "Hello".chars().map(CodePoint::Unicode).collect();
+    assert_eq!(latin1_points, expected_latin1);
+    assert!(latin1_str.is_latin1());
```
jedel1043: Can you add some latin1-compatible characters on this test string? Just to be sure we're properly testing latin1 and not just ASCII.

jedel1043 (commented):
I have a small suggestion about the new test. Everything else looks good!

core/string/src/tests.rs
Abd002: Thank you for the review and for the helpful suggestion! I've updated the test to include Latin1-compatible characters (like é and ï) as you recommended. This ensures we are properly testing the full Latin1 range and not just ASCII.

HalidOdat (approved):
Looks good to me! :smile:

core/string/src/str.rs
```diff
@@ -450,13 +450,10 @@ impl<'a> JsStr<'a> {
...
     /// Gets an iterator of all the Unicode codepoints of a [`JsStr`].
-    // TODO: optimize for Latin1 strings.
     #[inline]
-    pub fn code_points(&self) -> impl Iterator<Item = CodePoint> + Clone + use<'a> {
-        char::decode_utf16(self.iter()).map(|res| match res {
-            Ok(c) => CodePoint::Unicode(c),
-            Err(e) => CodePoint::UnpairedSurrogate(e.unpaired_surrogate()),
-        })
+    #[must_use]
+    pub fn code_points(&self) -> CodePointIter<'a> {
```
jedel1043: Very small nitpick: the iterator struct should be called `CodePointsIter`, to have consistency with the method name.

core/string/src/tests.rs
```diff
@@ -469,3 +469,23 @@ fn common_js_string_builder() {
...
 }
+
+#[test]
+fn code_points_optimization() {
+    // Test Latin1 optimization with extended Latin1 characters
+    let latin1_str = JsStr::latin1(b"Caf\xe9 na\xefve"); // "Café naïve" in Latin1 encoding
+    let latin1_points: Vec<CodePoint> = latin1_str.code_points().collect();
+    let expected_latin1: Vec<CodePoint> = "Café naïve".chars().map(CodePoint::Unicode).collect();
+    assert_eq!(latin1_points, expected_latin1);
+    assert!(latin1_str.is_latin1());
```
jedel1043: This assert is not really testing anything: the `JsStr` is already latin1 by construction.

core/string/src/tests.rs
```diff
@@ -469,3 +469,23 @@ fn common_js_string_builder() {
...
+
+    // Test UTF-16 behavior unchanged (including non-ASCII)
+    let utf16_str = JsStr::utf16(&[
+        0x0043, 0x0061, 0x0066, 0x00E9, // "Café"
+        0x0020, // space
+        0x006E, 0x0061, 0x00EF, 0x0076, 0x0065, // "naïve"
+    ]);
+    let utf16_points: Vec<CodePoint> = utf16_str.code_points().collect();
+    assert_eq!(latin1_points, utf16_points); // Same result for same content
+    assert!(!utf16_str.is_latin1());
```
jedel1043: Same suggestion here. The `JsStr` is already UTF-16 by construction.

jedel1043 (commented):
Final round of comments and we can review after that. Thank you for the contribution!

jedel1043:
Ah, you may have to add "Caf" to the list of excluded words in `typos.toml`.

core/string/src/str.rs
Abd002: I initially considered CodePointIter to align with the CodePoint enum type, but CodePointsIter for consistency with code_points() makes sense.

jedel1043 (approved):
LGTM!

PR #4563, opened 2025-12-09
Bump test262 to most recent commit

nekevss:
This PR bumps main to the most recent test262 commit and adds the `intl-era-monthcode` feature and the `joint-iteration` feature.


nekevss:
Small update here. There was one missing feature that needed to be added and a bug in test262 that should be addressed by tc39/test262#4773.

PR #4569, opened 2025-12-16
Add v8 benchmarks to the criterion benches

hansl:
Can be run with `cargo bench -p boa_benches`.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -13,3 +13,6 @@ test-results
 
 # For some reason Prettier likes to reformat JSON lock files.
 flake.lock
+
+# Assets that should not be checked.
+benches/scripts/v8-benches


core/engine/src/builtins/string/mod.rs
```diff
@@ -1887,7 +1888,9 @@ impl String {
         let to = max(final_start, final_end);
 
         // 10. Return the substring of S from from to to.
-        Ok(js_string!(string.get_expect(from..to)).into())
+        // Ok(js_string!(string.get_expect(from..to)).into())
```
nekevss: issue: remove commented code

benches/benches/scripts.rs
```diff
@@ -15,20 +15,29 @@ fn bench_scripts(c: &mut Criterion) {
     let scripts: Vec<_> = walkdir::WalkDir::new(&scripts_dir)
         .into_iter()
         .filter_map(|e| e.ok())
-        .filter(|e| e.path().extension().is_some_and(|ext| ext == "js"))
+        .filter(|e| {
```
nekevss, line 18: nit: small preference, but can we document with a comment the ignore file behavior.

PR #4571, opened 2025-12-19
Split SequenceString into two types for Latin1 and Utf16 

hansl:
Slice still has `is_latin1` which will be fixed in the next PR.

### DIFF
--- a/core/string/src/type.rs
+++ b/core/string/src/type.rs
@@ -3,38 +3,33 @@ use crate::vtable::SequenceString;
 use crate::{JsStr, JsStringKind};
 use std::alloc::Layout;
 
-pub(crate) mod sealed {
-    use crate::{JsStr, JsStringKind};
-    use std::alloc::Layout;
-
+mod sealed {
     /// Seal to prevent others from implementing their own string types.
     pub trait Sealed {}
-
-    /// Internal trait for crate-specific usage. Contains implementation details
-    /// that should not leak through the API.
-    #[allow(private_interfaces)]
-    pub trait InternalStringType {
-        /// The offset to the data field in the sequence string struct.
-        const DATA_OFFSET: usize;
-
-        /// The kind of string produced by this string type.
-        const KIND: JsStringKind;
-
-        /// The type of one character for this string type.
-        type Byte: Copy + Eq + 'static;
-
-        /// Create the base layout for the sequence string header.
-        fn base_layout() -> Layout;
-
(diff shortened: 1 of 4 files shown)


core/string/src/type.rs
```diff
@@ -0,0 +1,90 @@
+//! Module containing string types public and crate-specific.
+use crate::vtable::SequenceString;
+use crate::{JsStr, JsStringKind};
+use std::alloc::Layout;
+
+pub(crate) mod sealed {
```
nekevss: nit: remove `pub(crate)`

I know it's being imported below, but I think it's better to fully qualify the path, i.e. `sealed::Sealed` and `sealed::InternalStringType`. To me, it's a bit higher signal.

core/string/src/type.rs
```diff
@@ -0,0 +1,90 @@
...
+    use crate::{JsStr, JsStringKind};
+    use std::alloc::Layout;
+
+    /// Seal to prevent others from implementing their own string types.
+    pub trait Sealed {}
+
+    /// Internal trait for crate-specific usage. Contains implementation details
+    /// that should not leak through the API.
+    #[allow(private_interfaces)]
+    pub trait InternalStringType {
```
nekevss: thought: I don't think this needs to be in the private module either.

It can just be a `sealed::Sealed` super trait.

A lot of this tends to come from my experience with icu4x / temporal_rs, so I could be a bit biased on the approach towards sealing.

```rust
mod private {
    pub trait Sealed {}
}

trait TraitOne : private::Sealed { }

trait TraitTwo: private::Sealed { }
```

I think this approach is nice because it's very clear that the trait is private and sealed.

core/string/src/type.rs
```diff
@@ -0,0 +1,90 @@
...
+
+// It is good defensive programming to have [`Latin1`] `!Copy`, as it should
+// not be used as a value anyway.
+#[allow(missing_copy_implementations)]
+#[derive(Debug)]
+pub enum Latin1 {}
+
+impl Sealed for Latin1 {}
+impl StringType for Latin1 {
+    type Char = u8;
```
nekevss: nit: Shouldn't this be CodePoint rather than char since it's the numeric value vs. instead of the actual char

hansl: This is kind of a misnomer. This should be what you have when you transform a pointer into a slice. So it's the same as `Byte`. I need to fix that, let me see if I can do better.

PR #4585, opened 2026-01-23
Apply lints for rustc 1.93

jedel1043:


nekevss (approved):
LGTM.

Looks like the SemVer CI is the only blocker

Issue #3978, opened 2024-09-03
Build out debugger API for Boa's VM


nekevss:
With #3976, Boa's handling of the debugger keyword was updated to ignore debugger expressions. Currently, the behavior is for Boa to ignore any debugger statement.

A nice feature would be to further build out a debugger API for Boa's VM that can pause the VMs execution to allow users to view the current state of the VM.

PR #4586, opened 2026-01-23
Add Initial Debug Adapter Protocol (DAP) Support to Boa

albertleigh:
Hi Reviewers,

I am a fan of boa, and I think I could help with a DAP debugger for boa. 

This PR adds initial IDE debugging support to Boa CLI through the Debug Adapter Protocol (DAP), enabling developers to debug JavaScript code running in Boa using VS Code or any other DAP-compatible IDE.

Everything was hidden begind core/debugger and cli/dap features, they should not impact the current performance I believed.

And I think it would be a good time to talk about high-level stuffs like:
- we would like `core` to maintain most of the debug state like spidermonkey
- or `boa-cli` should not do too much.
- some file/struct were terrible named as Al is an alien. 🤣
- there should be a better way/place for dbg_log! marco
then we could settle places to pour more dap message and test again boa engine.

So I just put minimum changes to support debugger instruction and pause and support to resume like:

We could try it like:
1) Build a cli of dap features:  `cargo.exe build --message-format=json-diagnostic-rendered-ansi --color=always --bin boa --features default,dap --profile dev --manifest-path boa\cli\Cargo.toml`
2) Before we starting boa-cli, we could set BOA_DAP_DEBUG=1 to allow log some dap events to the console for tracing.
3) [Optional] we could first start the boa in your IDE rather than extension.js starting one for you and we have to attache debugger to its exe latter. `/boa/target/debug/boa.exe --dap 4711`
4) change directory to tools/vscode-boa-debug and `code .` there and run a light weight extension with the task `Run Extension`
5) So should have another VSC about to run the extension.js open and use this VSC open the folder tools/vscode-boa-debug/test-files, then run the task `Test Boa Debugger`. 
6) Make no mistake we should reached the happy path like the clip that I attached:

https://github.com/user-attachments/assets/654f05b1-d0de-4192-a949-157f12470683

Not sure we like this DAP idea or not, maybe too early for boa. I am not that sure. Should Al continue on it? Maybe Al could use it as another starting point and help boa better with more and more features. 

Like: 
- We could add this https://nekevss.github.io/boa-debugger/ as a side view in the extension.
- We could  gradually enhance our source line mapping to become more accurate while integrating DAP
- We could release boa via a vsc extension latter.

Let Al know whether boa like DAP ~

Thanks

ps: 
I noticed we got: 
https://github.com/boa-dev/boa/issues/3978
https://github.com/boa-dev/boa/pull/4510

Dont be so serious with the markdowns in this pr, I generate with AI
core/engine/src/debugger/README.MD

### DIFF
--- a/core/engine/src/debugger/README.md
+++ b/core/engine/src/debugger/README.md
@@ -12,17 +12,17 @@ The Boa debugger is a **SpiderMonkey-inspired** debugging system adapted for Rus
 
 Our design directly maps SpiderMonkey's debugging architecture to Boa's Rust implementation:
 
-| SpiderMonkey | Boa Equivalent | Purpose | Status |
-|--------------|----------------|---------|--------|
-| `JS::Debugger` | `Debugger` (state.rs) | Central debugger state | ⚠️ Basic (pause/resume only) |
-| `js::Breakpoint` | `Breakpoint` (breakpoint.rs) | Breakpoint metadata | ❌ Not implemented |
-| `DebuggerFrame` | `DebuggerFrame` (reflection.rs) | Call stack frame reflection | ⚠️ Basic |
-| `DebuggerScript` | `DebuggerScript` (reflection.rs) | Script/source code reference | ⚠️ Basic |
-| `DebuggerObject` | `DebuggerObject` (reflection.rs) | Safe object inspection | ⚠️ Basic |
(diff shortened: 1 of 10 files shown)


jedel1043:
Please don't open PRs with 9k new lines of code.

albertleigh:
okay 🤣

PR #4588, opened 2026-01-23
Correctly throw when DateTimeFormat receives a NaN timestamp

jedel1043:
Test `test/intl402/DateTimeFormat/prototype/format/throws-value-non-finite.js` was panicking when running the test suite in debug mode. This PR fixes the panic.

### DIFF
--- a/core/engine/src/builtins/intl/date_time_format/mod.rs
+++ b/core/engine/src/builtins/intl/date_time_format/mod.rs
@@ -323,6 +323,15 @@ impl DateTimeFormat {
 
                         // 5. Return ? FormatDateTime(dtf, x).
 
+                        // A.O 11.5.6 PartitionDateTimePattern
+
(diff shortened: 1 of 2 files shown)


hansl (approved):

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -323,6 +323,15 @@ impl DateTimeFormat {
 
                         // 5. Return ? FormatDateTime(dtf, x).
 
+                        // A.O 11.5.6 PartitionDateTimePattern
+
+                        // 1. Let x be TimeClip(x).
+                        // 2. If x is NaN, throw a RangeError exception.
+                        let x = time_clip(x);
```
nekevss, line 330: Huh, this was inlined in `ToLocalTime` 😕

PR #4589, opened 2026-01-23
Restructure Intl implementation to use ICU4X's locale preferences

jedel1043:
A pending task after migrating to ICU4X 2.0 was to use the newly introduced locale preferences to clean up our Intl implementation. This PR takes care of that by introducing the `ServicePreferences` trait and removing the `Service::resolve` method, which makes the code much more modular overall.

cc @sffc, it might be interesting to see if we can upstream parts of the `ServicePreferences` trait, like `as_unicode` or `intersection`.

### DIFF
--- a/core/engine/src/builtins/intl/collator/options.rs
+++ b/core/engine/src/builtins/intl/collator/options.rs
@@ -1,13 +1,22 @@
 use std::str::FromStr;
 
 use icu_collator::{
+    CollatorPreferences,
     options::{CaseLevel, Strength},
-    preferences::CollationCaseFirst,
+    preferences::{CollationCaseFirst, CollationType},
+};
+use icu_locale::{LanguageIdentifier, preferences::PreferenceKey};
+use icu_provider::{
+    DataMarker, DataMarkerAttributes, DryDataProvider,
+    prelude::icu_locale_core::{extensions::unicode, preferences::LocalePreferences},
 };
 
 use crate::{
     Context, JsNativeError, JsResult, JsValue,
-    builtins::options::{OptionType, ParsableOptionType},
+    builtins::{
+        intl::{ServicePreferences, locale::validate_extension},
+        options::{OptionType, ParsableOptionType},
+    },
 };
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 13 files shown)


core/engine/src/builtins/intl/collator/options.rs
```diff
@@ -97,3 +108,60 @@ impl OptionType for CollationCaseFirst {
...
+        exts
+    }
+
+    fn extended(&self, other: &Self) -> Self {
+        let mut result = *self;
+        result.extend(*other);
+        result
+    }
+
+    fn intersection(&self, other: &Self) -> Self {
```
sffc: Preferences have a merge operation already

core/engine/src/builtins/intl/collator/options.rs
```diff
@@ -97,3 +108,60 @@ impl OptionType for CollationCaseFirst {
...
+impl ServicePreferences for CollatorPreferences {
+    fn validate(&mut self, id: &LanguageIdentifier, provider: &IntlProvider) {
+        self.collation_type = self.collation_type.take().filter(|co| {
+            let attr = DataMarkerAttributes::from_str_or_panic(co.as_str());
+            co != &CollationType::Search
+                && validate_extension::<CollationMetadataV1>(id, attr, provider)
+        });
+    }
+
+    fn as_unicode(&self) -> unicode::Unicode {
```
sffc: I think in some cases we have impl From Preferences for Locale

core/engine/src/builtins/intl/collator/options.rs
```diff
@@ -97,3 +108,60 @@ impl OptionType for CollationCaseFirst {
         }
     }
 }
+
+impl ServicePreferences for CollatorPreferences {
+    fn validate(&mut self, id: &LanguageIdentifier, provider: &IntlProvider) {
+        self.collation_type = self.collation_type.take().filter(|co| {
+            let attr = DataMarkerAttributes::from_str_or_panic(co.as_str());
+            co != &CollationType::Search
```
sffc, line 116: Good, this is Ecma specific functionality so it should be here in boa

core/engine/src/builtins/intl/collator/options.rs
jedel1043: This the opposite though; it takes two preference objects and gets the preferences that are the same in both.

We need this because ECMA402 requires correctly setting up the locale with the extensions that were "used" by the locale resolution algorithm. Here I implemented that by first extending the locale prefs with the option prefs, then taking the intersection of the full prefs with the locale prefs to see which ones were not changed.

core/engine/src/builtins/intl/collator/options.rs
jedel1043: I also thought that, but after looking at the preferences code, it seems like it was commented out:
https://github.com/unicode-org/icu4x/blob/5fb31d7ae061d304088f1e79897f7435f1846da6/components/locale_core/src/preferences/mod.rs#L519-L533

core/engine/src/builtins/intl/collator/options.rs
jedel1043: Yep, extension validation like this seems to be a good candidate for a future "icu4x-ecma402" glue crate

core/engine/src/builtins/intl/collator/options.rs
sffc: Probably there just wasnt a need for it yet.

core/engine/src/builtins/intl/collator/options.rs
```diff
@@ -97,3 +108,60 @@ impl OptionType for CollationCaseFirst {
...
+                && validate_extension::<CollationMetadataV1>(id, attr, provider)
+        });
+    }
+
+    fn as_unicode(&self) -> unicode::Unicode {
+        let mut exts = unicode::Unicode::new();
+
+        if let Some(co) = self.collation_type
```
nekevss: thought: this could probably be a macro rule

A lot of these seem to be the exact same pattern. We may be able to lower a little of the boilerplate with a macro. But I'm not entirely sure whether that will make this as clear to parse.

core/engine/src/builtins/intl/locale/utils.rs
```diff
@@ -390,11 +390,29 @@ where
...
+    // - All options provided by the locale are valid.
+    // - All options provided by args are valid.
+    // - Options provided by args are extended (but not overridden) by
+    //   options provided in the locale.
+    // - Only the locale options that extended the args options are
+    //   added to the final locale.
     provider
         .locale_canonicalizer()?
         .canonicalize(&mut found_locale);
+    let mut locale_prefs = S::Preferences::from(&found_locale);
```
nekevss, line 404: question: does ordering matter here?

nekevss (approved):
This looks great! I like the new trait.

I had just a question and general thought, but nothing that's blocking

core/engine/src/builtins/intl/collator/options.rs
jedel1043: Yep, gonna see if I can improve on this.

Also, ideally all the preferences would implement `Into<Locale>`, but I'm still working on a PR in ICU4X for this.

core/engine/src/builtins/intl/locale/utils.rs
jedel1043: Yes. If we validated after extending, we might override a supported extension value in the locale with an unsupported extension value from the options, which would be worse overall. The rest of the things don't have an order though.

nekevss (approved):
Looks great!

PR #4597, opened 2026-02-02
Add Process object so as to allow process.env

### LINKED ISSUE #4492: Feature: `process.env`
Ability to access ENV variables passed into boa:

**Example code**

```js
const api_key = process.env.API_KEY
```

```bash
API_KEY=xxxxx boa -m app.js
```

randomboi404:
This Pull Request fixes/closes #4492.

It changes the following:

1) Ability to access ENV variables passed into boa:

Example code

```sh
const api_key = process.env.API_KEY
API_KEY=xxxxx boa -m app.js
```

2) Allows future addition of other process object methods of node js.

### DIFF
--- a/core/runtime/src/process/tests.rs
+++ b/core/runtime/src/process/tests.rs
@@ -62,7 +62,7 @@ fn process_property_descriptors() {
 }
 
 #[test]
-#[ignore]
+#[ignore = "Unsafe under parallel test execution as it tempers with env."]
 fn process_env_contains_variables() {
     temp_env::with_vars(
         [
@@ -90,7 +90,7 @@ fn process_env_contains_variables() {
 }
 
 #[test]
-#[ignore]
+#[ignore = "Unsafe under parallel test execution as it tempers with env."]
 fn process_env_properties_writable() {
     temp_env::with_var("TEST_VAR", Some("original"), || {
         let mut context = Context::default();
@@ -135,7 +135,7 @@ fn process_env_object_properties() {
 }
 
 #[test]
-#[ignore]
+#[ignore = "Unsafe under parallel test execution as it tempers with env."]
 fn process_env_iteration() {
     temp_env::with_vars(
         [


core/runtime/src/process/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// Name of the built-in `process` property.  
+    pub const NAME: JsString = js_string!("process");
+
+    /// Initializes the `process` built-in object.  
+    pub fn init(context: &mut Context) -> JsObject {
+        let process = JsObject::default(context.intrinsics());
+
+        // Create env object with current environment variables
+        let env = JsObject::default(context.intrinsics());
+        for (key, value) in std::env::vars() {
```
hansl, line 39: My only note right now would be that if the environment changes during the run (say using a Rust function), this would not be reflected. To resolve this you should do build a proxy object which catches property accessors. There's currently not enough examples of this, so I don't know how much you'd like to dig into it.

randomboi404: Well I tried to make it like how node.js process.env works. It fetches at runtime and you can change it by process.env.FOO = "xyz"; for your application only (won't be reflected in actual environment) but in either case, it will neither fetch nor update the original environment variables now. I think users coming from node.js will expect this behaviour too, no? What do you think?

hansl: Yeah I thought it was setting the environment, but it doesn't (a Node plugin would not show the right value).

The only thing missing is to stringify values when setting properties, but I will say it is non blocking for now. Let me review properly now that I have a computer.

core/runtime/src/process/tests.rs
```diff
@@ -0,0 +1,179 @@
+use crate::test::{TestAction, run_test_actions_with};
+use boa_engine::Context;
+use indoc::indoc;
+
+/// Test harness functions for process tests.  
+const TEST_HARNESS: &str = r#"
```
hansl, line 6: There is a new `TestAction::harness()` action which installs the harness. We haven't updated the console tests to it apparently.

The harness code can be seen in `core/runtime/assets/harness.js`.

core/runtime/src/lib.rs
```diff
@@ -145,6 +146,7 @@ pub fn register(
         StructuredCloneExtension,
         #[cfg(feature = "url")]
         extensions::UrlExtension,
+        ProcessExtension,
```
hansl, line 152: I don't think we want this one by default, considering it's a Node API and not a standard w3c one.

hansl (approved):
2 non-blocking nits for now. If you want to merge without addressing that's fine we can do it in a follow up PR.

randomboi404:
That makes sense.

Regarding the new `TestAction::harness()` since console tests haven't been migrated yet, I think it's better to let process be consistent with it and do a follow up once test harness is adopted more broadly.

Regarding it being enabled by default, I think we should let it be enabled for now and if in case we strictly don't want it enabled as it's a node api, we'll do a follow up to make it not being enabled by default.

nekevss:
I'd second having the Process extension not on by default.

I think this is a super cool extension to have as an option, but since process is not based on a web standard, we should have it be an opt in extension and not opt out.

randomboi404:
Done! I've made it optional.

nekevss (approved):
Thanks! 😄 

The feature flag approach looks good to me!

core/runtime/src/process/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        let _unused = process.define_property_or_throw(
+            js_string!("env"),
+            PropertyDescriptor::builder()
+                .value(env)
+                .writable(true)
+                .enumerable(false)
+                .configurable(true)
+                .build(),
+            context,
+        );
```
jedel1043: Why ignore the errors instead of bubbling them up?

core/runtime/src/process/tests.rs
```diff
@@ -0,0 +1,179 @@
...
+        &mut context,
+    );
+}
+
+#[test]
+fn process_env_contains_variables() {
+    unsafe {
+        std::env::set_var("TEST_VAR", "test_value");
+        std::env::set_var("ANOTHER_VAR", "another_value");
+    }
```
jedel1043: Wouldn't this cause UB if Cargo decides to run this test and the test below it concurrently?

nekevss: Oh, shoot. That's true.

It may be worth using something like `temp_env` as a dev-dependency.

core/runtime/src/process/mod.rs
randomboi404: Yup, I bubbled them up right now in new commit.

core/runtime/src/process/tests.rs
randomboi404: Ah yes, my bad for overlooking it.
Also as per @nekevss's suggestion, I'll use `temp_env` as a dev dependency and fix it.

randomboi404: Done!

jedel1043: Unfortunately, I don't think this fixes the issue, according to [this issue](https://github.com/vmx/temp-env/issues/34) in the `temp_env` repo. I see two possible solutions:
- Require running tests with nextest, but it feels "weird" to depend on another test runner to not cause UB.
- Somehow run the test in a subprocess that sets up the environment variables before running it. This is the safest way of testing this but might be much more complex to set up.

Are you open for taking a shot at the second option? If not, then I would propose ignoring the tests for now and opening an issue for fixing this later.

hansl: @jedel1043 I think there is a couple other solutions:

1. have a global mutex lock. Rust describes that `set_var` is only UB when multiple threads try to write at the same time. If threads are synchronized, there's no UB.
2. merge the two tests into the same test to avoid this case. I don't recommend this one because we might add more tests in runtime with `set_var` later on and would likely not think about updating this test suite.

This saves having to create a subprocess which is heavy.

Ref https://doc.rust-lang.org/stable/std/env/fn.set_var.html#safety

jedel1043: Hmm a lot of options to discuss then.

jedel1043: In that case, let's just merge after ignoring the new tests, and we can discuss the shape of our solution later

randomboi404: I agree that the subprocess option would be safest but it'll be quite complex to implement for what we're trying to test here.

I think we can go for the mutex route later but I'm okay with ignoring the new tests for now and create a new issue for it where we can discuss it.

Thanks for the review!

PR #4603, opened 2026-02-05
Add process.cwd()

randomboi404:
It changes the following:

1) Uses a provider pattern so as to allow custom `ProcessProvider` for embedded systems where using the default `StdProcessProvider` will not be proper.

2) Adds `process.cwd()` so as to allow:

```sh
./boa
>> process.cwd()
"/root/boa"
```

### DIFF
--- a/core/runtime/src/process/mod.rs
+++ b/core/runtime/src/process/mod.rs
@@ -11,11 +11,10 @@
 pub(crate) mod tests;
 
 use boa_engine::{
-    Context, JsData, JsError, JsObject, JsResult, JsString, JsSymbol, JsValue, js_string,
+    Context, JsData, JsObject, JsResult, JsString, JsSymbol, JsValue, js_error, js_string,
     native_function::NativeFunction, object::ObjectInitializer, property::Attribute,
 };
 use boa_gc::{Finalize, Trace};
-use std::collections::HashMap;
 use std::rc::Rc;
 
 /// A trait that can be used to forward process provider to an implementation.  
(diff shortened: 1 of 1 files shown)


core/runtime/src/process/mod.rs
```diff
@@ -1,42 +1,92 @@
...
+    /// # Errors  
+    /// Returns an error if the current directory cannot be obtained.  
+    fn cwd(&self) -> JsResult<JsString>;
 
-/// Boa's implementation of Node.js' `process` object.    
+    /// Get a `HashMap` of environment variables so as to allow env property (`process.env`)  
+    ///  
+    /// # Errors  
+    /// Returns an error if the environment variables cannot be obtained.  
+    fn env(&self) -> JsResult<HashMap<String, String>>;
```
jedel1043: I would suggest just using `impl IntoIterator<(JsString, JsString)>` as the return value here, to mirror what `std::env::vars()` returns. IMO we really don't need a fallible method here.

core/runtime/src/process/mod.rs
```diff
@@ -1,42 +1,92 @@
...
+///  
+/// Implements the [`ProcessProvider`] trait. Outputs the process properties'
+/// values on the basis of std.  
 #[derive(Debug, Trace, Finalize)]
+pub struct StdProcessProvider;
+
+impl ProcessProvider for StdProcessProvider {
+    fn cwd(&self) -> JsResult<JsString> {
+        let path = std::env::current_dir()
+            .map_err(|e| JsError::from_opaque(js_string!(e.to_string()).into()))?;
```
jedel1043: It's not that useful to return an opaque error here; just use `TypeError` and pass `e.to_string()` as the message for the error, maybe with a nice error prefix to show that this error comes from calling `std::env::current_dir()`

jedel1043 (changes requested):
Implementation wise looks great! I just have an API suggestion and a comment on the error thrown when calling `current_dir()`.

core/runtime/src/process/mod.rs
jedel1043: By the way you can easily format nice error messages with our `js_error` macro:
```
js_error!(TypeError: "nice error message. source: {error}", e.to_string());
```

core/runtime/src/process/mod.rs
randomboi404: Yup I see. Will do it!

core/runtime/src/process/mod.rs
randomboi404: Oh I see. I'll get rid of it right away!

PR #4607, opened 2026-02-06
Fix missing source position/span in errors caught by internal handlers

### LINKED ISSUE #4608: Errors caught by internal handlers (e.g. module async wrapper) lack source position/span info
## Description

When a runtime error occurs inside a module (e.g. calling a non-callable value), the resulting error contains no JavaScript source position or backtrace information. This makes it difficult to identify where the error occurred.

## Steps to Reproduce

**hello.js:**
```js
function foo() {
  undefined();
}
foo();
```

**Rust host (running as module):**
```rust
let module = Module::parse(source, Some(context.realm().clone()), &mut context).unwrap();
let promise = module.load_link_evaluate(&mut context);

let err = match promise.await_blocking(&mut context) {
    Ok(js_value) => return println!("Fulfilled: {}", js_value.display()),
    Err(e) => e,
};

println!("{err}"); // TypeError: not a callable function
```

## Expected Behavior

The error should include the source location:
```
TypeError: not a callable function (hello.js:2:3)
    at foo (hello.js:2:3)
    at <anonymous> (hello.js:4:1)
```

## Actual Behavior

```
TypeError: not a callable function
```

No position or backtrace information.

## Root Cause

In `handle_error()`, when a catchable error is intercepted by an internal exception handler (such as the module's async wrapper), the error is stored as `pending_exception` without capturing the backtrace or injecting the bytecode source position. Both only happen in `handle_throw()`, which is never reached when an internal handler catches the error first.

This affects all `JsNativeError`s from opcode handlers when run inside modules — TypeError, ReferenceError, SyntaxError, etc.

## Fix

See #4607

AlbertMarashi:
## Summary

Closes #4608

- Errors caught by internal exception handlers (e.g. the module async wrapper) now include JavaScript source position and backtrace information
- Previously, only errors that reached `handle_throw()` got this info — but internal handlers intercepted the error before that point
- Adds `JsError::inject_position()` to set the bytecode position on native errors that lack one
- Adds regression test using `undefined()` call inside a module

## Details

When a catchable error (like `TypeError: not a callable function`) was intercepted by an internal exception handler (such as the module's async wrapper), `handle_error()` stored it as `pending_exception` **without** capturing the backtrace or injecting the JavaScript source position. Both only happened in `handle_throw()`, which is never reached when an internal handler catches the error first.

This affected all `JsNativeError`s returned by opcode implementations when run inside modules, including:
- `TypeError`: not a callable function, not a constructor, Cannot destructure null/undefined, cannot set non-writable property
- `ReferenceError`: X is not defined, cannot delete a property of super
- `SyntaxError`/`TypeError` from `in` operator misuse
- And any other native error from opcode handlers

The fix moves backtrace capture and position injection into `handle_error()` **before** the exception handler check, so the information is present regardless of how the error is caught.

## Test plan

- [x] New test `test_module_error_has_source_position` — calls `undefined()` inside a module, verifies the rejected error contains line number info after round-tripping through `try_native()`
- [x] All existing module tests pass (8/8)
- [x] Full `boa_engine` test suite passes (120/120)

### DIFF
--- a/core/engine/src/builtins/error/mod.rs
+++ b/core/engine/src/builtins/error/mod.rs
@@ -20,7 +20,7 @@ use crate::{
     property::Attribute,
     realm::Realm,
     string::StaticJsStrings,
-    vm::shadow_stack::ShadowEntry,
+    vm::shadow_stack::{Backtrace, ShadowEntry},
 };
 use boa_gc::{Finalize, Trace};
 use boa_macros::js_str;
@@ -137,6 +137,12 @@ pub struct Error {
     // The position of where the Error was created does not affect equality check.
     #[unsafe_ignore_trace]
     pub(crate) position: IgnoreEq<Option<ShadowEntry>>,
+
+    // The backtrace captured when this error was thrown. Stored here so it
+    // survives the JsError → JsValue → JsError round-trip through promise
+    // rejection. Does not affect equality checks.
+    #[unsafe_ignore_trace]
+    pub(crate) backtrace: IgnoreEq<Option<Backtrace>>,
 }
 
 impl Error {
@@ -147,6 +153,7 @@ impl Error {
         Self {
             tag,
             position: IgnoreEq(None),
+            backtrace: IgnoreEq(None),
         }
     }
 
@@ -155,6 +162,7 @@ impl Error {
         Self {
             tag,
             position: IgnoreEq(entry),
+            backtrace: IgnoreEq(None),
         }
     }
 
@@ -163,6 +171,7 @@ impl Error {
         Self {
             tag,
             position: IgnoreEq(context.vm.shadow_stack.caller_position()),
+            backtrace: IgnoreEq(None),
         }
     }
 }
--- a/core/engine/src/error.rs
+++ b/core/engine/src/error.rs
@@ -377,6 +377,9 @@ impl JsError {
 
     /// Creates a new `JsError` from an opaque error `value`.
     ///
+    /// If the value is an `Error` object that contains a stored backtrace
+    /// (from a previous `into_opaque` conversion), the backtrace is recovered.
+    ///
     /// # Examples
     ///
     /// ```rust
@@ -386,10 +389,16 @@ impl JsError {
     /// assert!(error.as_opaque().is_some());
     /// ```
     #[must_use]
-    pub const fn from_opaque(value: JsValue) -> Self {
+    pub fn from_opaque(value: JsValue) -> Self {
+        // Recover the backtrace from the Error object if present,
+        // so it survives the JsError → JsValue → JsError round-trip.
+        let backtrace = value.as_object().and_then(|obj| {
+            let error = obj.downcast_ref::<Error>()?;
+            error.backtrace.0.clone()
+        });
         Self {
             inner: Repr::Opaque(value),
-            backtrace: None,
+            backtrace,
         }
     }
 
@@ -421,7 +430,18 @@ impl JsError {
     /// ```
     pub fn into_opaque(self, context: &mut Context) -> JsResult<JsValue> {
         match self.inner {
-            Repr::Native(e) => Ok(e.into_opaque(context).into()),
+            Repr::Native(e) => {
+                let obj = e.into_opaque(context);
+                // Store the backtrace in the Error object so it survives the
+                // JsError → JsValue → JsError round-trip through promise
+                // rejection.
+                if let Some(backtrace) = self.backtrace {
+                    if let Some(mut error) = obj.downcast_mut::<Error>() {
+                        error.backtrace = IgnoreEq(Some(backtrace));
+                    }
+                }
+                Ok(obj.into())
+            }
             Repr::Opaque(v) => Ok(v.clone()),
             Repr::Engine(_) => Err(self),
         }
--- a/core/engine/tests/module.rs
+++ b/core/engine/tests/module.rs
@@ -5,7 +5,7 @@ use std::rc::Rc;
 
 use boa_engine::builtins::promise::PromiseState;
 use boa_engine::module::{ModuleLoader, Referrer};
-use boa_engine::{Context, JsResult, JsString, Module, Source, js_string};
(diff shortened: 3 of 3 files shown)


core/engine/src/vm/mod.rs
```diff
@@ -699,8 +699,27 @@ impl Context {
...
+        // Always capture backtrace for catchable errors, since the error may be
+        // caught by an internal handler (e.g. module async wrapper) before
+        // handle_throw has a chance to capture it.
+        if err.backtrace.is_none() {
+            err.backtrace = Some(
+                self.vm
+                    .shadow_stack
+                    .take(self.vm.runtime_limits.backtrace_limit(), self.vm.frame.pc),
+            );
+        }
```
jedel1043: This is exactly the same code as the one inside the previous `if`. You just need to extract the code from the `if`.

core/engine/src/vm/mod.rs
```diff
@@ -699,8 +699,27 @@ impl Context {
...
+        // Inject bytecode position from the current frame into native errors
+        // that don't already have one, so the error message includes the
+        // JavaScript source location where the error occurred.
         // Note: -1 because we increment after fetching the opcode.
         let pc = self.vm.frame().pc.saturating_sub(1);
+        err.inject_position(ShadowEntry::Bytecode {
+            pc,
+            source_info: self.vm.frame.code_block().source_info.clone(),
+        });
+
```
jedel1043: This is kind of incorrect. It may work for a single level of error erasure, but the moment you have multiple levels of conversion from `JsNativeError` to `JsValue`(IIRC `try` blocks do this), you would be overriding the last position with an unrelated position in the code.

Fortunately, we already store the correct position inside `Error` objects when converting from `JsNativeError` to `JsValue`, so the correct thing to do here is to "recover" this info on our `impl Display` for `JsError`.

jedel1043: Though, we _might_ want to do something like this for "rare" errors such as when a code throws non-`Error` values as errors (`throw 5` or something), but it is not really clear to me at first glance if that would be _useful_ if the position can be overridden by any `try` block that catches the error.

jedel1043: Hmmm, I looked at the code for `JsNativeError` again and maybe this is still required? Seems like we're only setting the position for Rust's native stack, but not for the VM position itself.

jedel1043: We might just have to change the implementation of `inject_position` then, since it's currently overriding any "native stack" position with the JS position instead

AlbertMarashi:
**On the backtrace capture comment:**

> Moved the backtrace capture to the top of `handle_error`, before any handler check. Note: the duplicate in `handle_throw` is intentionally kept since `handle_throw` is called from 6 other call sites (Throw, ReThrow, await, yield, return) that don't go through `handle_error`.

**On the `inject_position` comment**:

> Updated the approach based on your feedback:
>
> - Now uses the current frame's PC directly (`ShadowEntry::Bytecode { pc, source_info }`) instead of `caller_position()`. This is correct for opcode-level errors since no new frame is pushed when e.g. `Call::operation` fails — `caller_position()` would have returned the caller's caller.
> - The `is_none()` guard ensures we never override an existing position (whether native or bytecode), so multiple conversion levels are safe.
> - Also added a fallback in `Display` for `Repr::Opaque` errors: when the backtrace is `None` (promise rejection path), it tries to recover position from the `Error` object's internal data.

Force-pushed with these changes + reset `Cargo.lock` to match main.

@jedel1043

core/engine/src/vm/mod.rs
jedel1043: Okay, I did a deep dive on the code and it seems like this is not really necessary after all, but not for the reason I expected:
- All errors created from JS will set their position correctly on the `Error()` constructor, which covers most use cases.
- All errors created by the engine itself will create a `ShadowEntry::Native` position such that the error has the position in the Rust code base that created the error. This is useful for runtime developers, but not really for end users, which is I guess the origin of #4608.

From the second reason, I think the _correct_ fix is to instead store the backtrace in `Error` when converting from `JsError` to `JsValue` if the inner `JsError` is a `JsNativeError`, and recover it back in `JsError::from_opaque` if our `JsValue` is an `Error` object. This should ensure that we can still report the "source function" that caused the error while also preserving the original native function's position where the error was created. Other types of errors might not be able to preserve the backtrace, but it doesn't matter in that case IMO because it is kind of impossible to preserve error info if our error value is "5" or "null".

Sorry for all the comments by the way! This has been kind of a pain point of our error messages for a while so I just want to have it be in a good place if possible.

AlbertMarashi: I know you posted this comment a minute before I force pushed those changes (accidentally committed to a different branch) - but could you let me know what you thought about the newly updated code then?

jedel1043: My last comment still applies to the new code, because if we store the backtrace in `Error` within `JsError::into_opaque` and recover it in `JsError::from_opaque`, we don't really need to do all the shenanigans of having to recover and inject the position of errors; only setting the backtrace of the error would take care of that for us.

AlbertMarashi: You'd have a much better picture on the internals than I do, so I have no opinions here.

Do you think you might be able to apply your changes and edits to this PR instead? Otherwise I will have to do this tomorrow or the day after @jedel1043.

jedel1043: I would prefer if you did the work because doing it myself would basically mean I would undo almost all your changes 😅

However, the change shouldn't be hard at all, you just need to extend [this Error type](https://github.com/boa-dev/boa/blob/a981f32197a7c60c66d0d09af4afa182b5410840/core/engine/src/builtins/error/mod.rs#L134-L140) with the `backtrace` data in `JsError`, then push that when converting from `JsError` to `JsValue` in `JsError::into_opaque` and recover that in `JsError::from_opaque`.

AlbertMarashi:
Sorry I am just going to create a new PR as this one has gotten a little bit messed up.

PR #4609, opened 2026-02-07
Preserve backtrace through JsError → JsValue → JsError round-trip

### LINKED ISSUE #4608: Errors caught by internal handlers (e.g. module async wrapper) lack source position/span info

AlbertMarashi:
## Summary

Closes #4608
Supersedes #4607

- Errors caught by internal exception handlers (e.g. async module evaluation) now preserve their backtrace through the `JsError → JsValue → JsError` round-trip via promise rejection
- Stores the backtrace on the `Error` object in `JsError::into_opaque` (both `Native` and `Opaque` variants) and recovers it in `JsError::from_opaque`
- Moves backtrace capture in `handle_error` before the exception handler check, so catchable errors also get a backtrace before being intercepted by internal handlers
- Adds regression tests

## Details

Previously, the backtrace was only captured for uncatchable errors (inside `if !err.is_catchable()`) and for catchable errors that reached `handle_throw()`. Errors intercepted by internal handlers (e.g. the module async wrapper) were stored as `pending_exception` _before_ either path could capture a backtrace. When these errors were later converted to `JsValue` for promise rejection, the backtrace was lost entirely.

This PR:

1. **Captures the backtrace unconditionally** in `handle_error`, before the exception handler check
2. **Stores the backtrace on the `Error` object** during `into_opaque` (for both `Repr::Native` and `Repr::Opaque` errors) so it survives conversion to `JsValue`
3. **Recovers the backtrace** in `from_opaque` by reading it back from the `Error` object

This approach was suggested by @jedel1043 in #4607 — extending the [`Error` type](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/error/mod.rs#L134-L140) with a `backtrace` field rather than injecting positions, which would be incorrect with nested `try` blocks.

## Test plan

- [x] `test_call_error_preserves_backtrace` — calls `undefined()` at module top level, verifies backtrace with exact string match
- [x] `test_nested_call_error_preserves_backtrace` — nested `foo → baz → import.meta.non_existent()`, verifies full multi-frame backtrace
- [x] `test_explicit_throw_preserves_backtrace` — explicit `throw new Error("test")` inside a function, verifies backtrace frames are present

Made with [Cursor](https://cursor.com)


AlbertMarashi:
@jedel1043 requesting review

randomboi404:
I recommend you to fix the issues in your code and make it pass the CI otherwise it can't be merged.

AlbertMarashi:
@randomboi404 @jedel1043 done

AlbertMarashi:
interesting. it appears that backtraces (maybe a feature) seems to differ on the CI...

to avoid flakiness with source code changes in the `(native)` part of the backtraces, I might strip them out or use another approach - (so ideally when the source code is changed it doesn't painstakingly cause this test to fail also)

---

Reproduced it. So:                                                                        
                                                                                            
- `cargo test -p boa_engine` — only `boa_engine`'s own features, `native-backtrace` is off →     tests pass
- `cargo test --workspace` (what CI does) — Cargo unifies features across the workspace, the CLI crate enables `native-backtrace` on `boa_engine` → `(native at file:line:col)` appears instead of `(native)` → exact string match fails

core/engine/tests/errors.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+    result.push_str(rest);
+    // Trim trailing whitespace left behind on each line
+    result
+        .lines()
+        .map(str::trim_end)
+        .collect::<Vec<_>>()
+        .join("\n")
+}
```
jedel1043: This solution seems a bit flaky. What about instead of trying to match the output of `Display`, just try to match against the error backtrace itself? That way, you can conditionally test that the first element in the backtrace is a correct native entry if `native-backtrace` is enabled, and skip this check when the feature is disabled.

AlbertMarashi: @jedel1043 should we make the `Backtrace` struct, `ShadowEntry` enum, or `JsError.backtrace` methods public APIs in that case?

Unable to access them from the tests.

jedel1043: The backtrace is `pub(crate)` so you should be able to access it from here.

core/engine/src/builtins/error/tests.rs
```diff
@@ -1,7 +1,167 @@
...
+        &mut context,
+        br"function foo() {
+    function baz() {
+        import.meta.non_existent()
+    }
+    baz()
+}
+
+foo()
+",
```
jedel1043: You should be able to use indoc here to prettify the code string so that it has the same indentation as the rest of the code. Also applies to the other tests.

AlbertMarashi:
@jedel1043 should be ready to merge now.

core/engine/src/builtins/error/tests.rs
jedel1043, line 1: These tests will only run correctly when native backtraces are enabled, so it might be worth tweaking the test assertions depending on whether the native backtraces feature is enabled or disabled

also, I think these should be in `core/engine/src/error/tests.rs` instead (migrating to a module directory if necessary), since they don't really touch the `Error` builtin.

AlbertMarashi: Done — moved backtrace tests to `core/engine/src/error/tests.rs` (converted `error.rs` to a module directory).

Regarding the feature flag: the backtrace tests currently pass both with and without `native-backtrace` enabled, because `push_native` on the shadow stack is called unconditionally — the feature only affects whether `NativeSourceInfo` carries real Rust source locations. The test assertions only check frame types and JS-level source positions, not native locations, so they're valid either way. Let me know if you'd still like them gated behind `#[cfg(feature = "native-backtrace")]`.

jedel1043: Oh nice. Then we can leave it at that. The source code locations are much more prone to change, so I think it's better to not test those.

PR #4610, opened 2026-02-07
Add tests for conditional keywords vs. reserverd keywords (#4598)

### LINKED ISSUE #4598: Add tests for conditional keywords vs. reserved keywords
In #4593, we were not properly handling non-reserved keywords. It would be nice to add unit tests for the linked PR.

Likewise, it appears that the fix did not trigger any more passes or failures in test262. It may be worth upstreaming some tests to the test262 repository

adrien-zinger:
Add a unit test to check the parser to handle correctly handling non-reserved keywords during a `let` declaration.

This Pull Request fixes/closes #4598.

It changes the following:

- Add a unit test in core parser to check handling of non-reserved keywords in a `let` declaration.

Related to #4593 introducing the change.

### DIFF
--- a/core/parser/src/parser/statement/declaration/tests.rs
+++ b/core/parser/src/parser/statement/declaration/tests.rs
@@ -251,6 +251,30 @@ fn let_declaration_no_spaces() {
     );
 }
 
+/// Checks `let` declaration with a reserved keyword as identifier.
+#[test]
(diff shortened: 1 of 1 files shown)


nekevss:
Looks good! Thanks for the contribution! 😁

PR #4616, opened 2026-02-08
feat: Implement proposal-arraybuffer-base64

magic-akari:
This Pull Request implements [proposal-arraybuffer-base64](https://github.com/tc39/proposal-arraybuffer-base64).

It changes the following:

- Add `Uint8Array.fromBase64(string, options)` static method for decoding base64 strings
- Add `Uint8Array.prototype.setFromBase64(string, options)` method for in-place base64 decoding
- Add `Uint8Array.prototype.toBase64(options)` method for base64 encoding
- Add `Uint8Array.fromHex(string)` static method for decoding hex strings
- Add `Uint8Array.prototype.setFromHex(string)` method for in-place hex decoding
- Add `Uint8Array.prototype.toHex()` method for hex encoding
- Add `simdutf` crate dependency for efficient encoding/decoding operations
- Update test262 expectations for the new proposal tests

---

- close: #4630

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -464,6 +464,7 @@ dependencies = [
  "ryu-js",
  "serde",
  "serde_json",
+ "simdutf",
  "small_btree",
  "static_assertions",
  "sys-locale",
@@ -758,9 +759,9 @@ checksum = "a2698f953def977c68f935bb0dfa959375ad4638570e969e2f1e9f433cbf1af6"
 
 [[package]]
 name = "cc"
-version = "1.2.49"
+version = "1.2.55"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "90583009037521a116abf44494efecd645ba48b6622457080f080b85544e2215"
+checksum = "47b26a0954ae34af09b50f0de26458fa95369a0d478d8236d3f93082b219bd29"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -1444,9 +1445,9 @@ dependencies = [
 
 [[package]]
 name = "find-msvc-tools"
-version = "0.1.5"
+version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3a3076410a55c90011c298b04d0cfa770b00fa04e1e3c97d3f6c9de105a03844"
+checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
 
 [[package]]
 name = "fixed_decimal"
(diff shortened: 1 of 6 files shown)


magic-akari:
1. **Stage 4 Status**: The proposal-arraybuffer-base64 has now reached Stage 4 in the TC39 process. Given its finalized status, would the Boa team be interested in accepting this implementation?

2. **WASM Support**: The current implementation uses the `simdutf` crate for efficient encoding/decoding operations. However, this crate does not support WASM targets, and WASM support may not be resolved in the short term. 
   - Is this limitation acceptable for merging?
   - Should I document this limitation somewhere (e.g., in the PR description or code comments)?

3. **Implementation Review**: Please let me know if there are any changes or improvements needed in the implementation. I'm happy to address any feedback.

jedel1043:
Wow, thank you for the great work! The implementation looks nice, but I just wanted to raise some concerns about using `simdutf`. IIUC, this is a library written in C++, right?

Historically, we've not used any C or C++ libraries for two reasons:
- Folks kind of appreciate having a JS engine that can be compiled very easily for WASM, x86, ARM, etc. without having to setup their compilation environment too much.
- Using C and C++ libraries kind of negates the benefit of having all the safety guarantees that the language offers, which has been a big selling point for the engine.

If this is the case, I wanted to ask if using something like [encoding_rs](https://docs.rs/encoding_rs/latest/encoding_rs) (or any other Rust-only encoding library) was out of the box for this implementation.

magic-akari:
Hi @jedel1043, thanks for the feedback! I've looked into the alternatives you mentioned and here's what I found:

## encoding_rs

Unfortunately, **encoding_rs doesn't support Base64 at all**. It's focused on character encoding conversions (UTF-8, legacy encodings like GBK, Shift_JIS, etc.) but Base64 is outside its scope. So this isn't a viable option.

## [data_encoding](https://docs.rs/data-encoding/latest/data_encoding/)

This crate supports basic Base64 operations but is missing two key features required by the TC39 proposal:

### 1. lastChunkHandling modes

The spec requires three modes:
- `"loose"` - allows incomplete 4-char chunks (data_encoding only supports this)
- `"strict"` - requires complete chunks, errors otherwise
- `"stop-before-partial"` - stops before incomplete chunks

### 2. read/written statistics

The spec requires `setFromBase64` to return `{ read: number, written: number }`.

`DecodePartial` does provide `read` and `written` fields, but **only when decoding errors occur**. For successful decoding, `decode_mut` only returns the written length, not the read count.

This means we can't get the `read` statistic for successful decodes, which is required by the spec.

## Current situation

I tried implementing the missing features on top of data_encoding in Boa, but even with ~300 lines of complex manual parsing code, there are still edge cases that don't work correctly. This approach doesn't seem viable.

## Possible ways forward

1. **Keep simdutf for now** - works perfectly but no WASM support
2. **Use data_encoding and implement missing features in Boa** - attempted but couldn't get all edge cases working correctly
3. **Keep looking** - maybe there's another crate I missed?

I'm also planning to try upstreaming the missing features to data_encoding, though as a third-party dependency there's uncertainty around timeline and acceptance.

What are your thoughts on the best path forward?

jedel1043:
Oh nice, thank you for the write-up with your investigation.

> The spec requires three modes:
    "loose" - allows incomplete 4-char chunks (data_encoding only supports this)
    "strict" - requires complete chunks, errors otherwise
    "stop-before-partial" - stops before incomplete chunks

If `data_encoding` only supports the first mode, I don't see any problem with only implementing the "loose" mode and coming around to implement the rest of the modes after adding the required support for it in `data_encoding`; just offering the "loose" mode already provides a lot of value.

> DecodePartial does provide read and written fields, but only when decoding errors occur. For successful decoding, decode_mut only returns the written length, not the read count.

Isn't this enough? If decoding succeeded, then `read` will always be the length of the input buffer, right? We only care about `read` when decoding fails IIUC.

jedel1043:
Closing in favour of #4630

PR #4624, opened 2026-02-10
Implement getter for Context.instructions_remaining

ShovelTime:
This implements `Context.instructions_remaining()`
for reading the remaining instruction budget of the Context, without the need for pointer arithmetic to access the field otherwise.

​This would be useful for cases where we want to monitor engine usage in a remote service for billing purposes


jedel1043 (approved):
Seems a reasonable enough addition. We might want to have discussions about "lifting" the instructions_remaining parameter out of the `fuzz` feature, but that can be done in the future.

PR #4625, opened 2026-02-12
Add feature flag for accessing host's system timezone

### LINKED ISSUE #4599: Feature flag system-time-zone access
Currently, we use `iana-time-zone`, and to access the host's system time zone by default. We should not do this.

Instead, we should have a HostTimeZone that always returns UTC by default, and a feature flagged HostTimeZone that accesses the host's time zone.

Vellumic:
### Changes:
- **Cargo.toml**: Moved `iana-time-zone` to added `system-time-zone` flag dependencies
- **core/engine/src/builtins/temporal/now.rs**:
  - Wrapped `temporal_rs::TemporalError` import with `#[cfg(feature = "system-time-zone")]`
  - `get_host_time_zone()` now uses `iana-time-zone::get_timezone()` only when flag is enabled, `TimeZone::utc_with_provider()` otherwise

### Verification:
- [x] **Automated tests**: Ran
  - [x] `cargo test --package boa_engine --all-targets --features system-time-zone` (all passed)
  - [x] `cargo test --package boa_engine --all-targets --no-default-features` (all passed)
- [x] **Manual check**: With flag
```bash
$ cargo run --features system-time-zone
>> Temporal.Now.timeZoneId()
"Australia/Sydney"
```
Without flag
```bash
$ cargo run
>> Temporal.Now.timeZoneId()
"UTC"
```

This Pull Request closes #4599

### DIFF
--- a/core/engine/Cargo.toml
+++ b/core/engine/Cargo.toml
@@ -71,7 +71,10 @@ trace = ["js"]
 annex-b = ["boa_ast/annex-b", "boa_parser/annex-b"]
 
 # Enable Boa's Temporal proposal implementation
-temporal = ["dep:icu_calendar", "dep:temporal_rs", "dep:iana-time-zone", "dep:timezone_provider", "timezone_provider/tzif"]
+temporal = ["dep:icu_calendar", "dep:temporal_rs", "dep:timezone_provider", "timezone_provider/tzif"]
+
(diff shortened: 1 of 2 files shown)


nekevss (approved):
This looks great! Thanks so much for contributing this!

PR #4630, opened 2026-02-16
feat: Implement proposal-arraybuffer-base64

### LINKED ISSUE #5170: Implement Uint8Array base64/hex encoding methods (fromBase64, fromHex, toBase64, toHex, setFromBase64, setFromHex)
## Description

Boa is missing the **Uint8Array base64 and hex encoding/decoding methods** specified in ECMAScript 2025 (Section 23.3). These methods provide native, efficient conversion between `Uint8Array` byte data and base64/hex string representations — functionality that is extremely common in real-world JavaScript (handling binary data, cryptographic digests, network protocols, etc.).

### Methods to implement

| Method | Spec Section | Type |
|---|---|---|
| `Uint8Array.fromBase64(string [, options])` | [§23.3.1.1](https://tc39.es/ecma262/#sec-uint8array.frombase64) | Static |
| `Uint8Array.fromHex(string)` | [§23.3.1.2](https://tc39.es/ecma262/#sec-uint8array.fromhex) | Static |
| `Uint8Array.prototype.toBase64([options])` | [§23.3.2.3](https://tc39.es/ecma262/#sec-uint8array.prototype.tobase64) | Instance |
| `Uint8Array.prototype.toHex()` | [§23.3.2.4](https://tc39.es/ecma262/#sec-uint8array.prototype.tohex) | Instance |
| `Uint8Array.prototype.setFromBase64(string [, options])` | [§23.3.2.1](https://tc39.es/ecma262/#sec-uint8array.prototype.setfrombase64) | Instance |
| `Uint8Array.prototype.setFromHex(string)` | [§23.3.2.2](https://tc39.es/ecma262/#sec-uint8array.prototype.setfromhex) | Instance |

### Helper abstract operations (§23.3.3)

- `ValidateUint8Array(ta)` — §23.3.3.1
- `GetUint8ArrayBytes(ta)` — §23.3.3.2
- `SetUint8ArrayBytes(into, bytes)` — §23.3.3.3
- `SkipAsciiWhitespace(string, index)` — §23.3.3.4
- `DecodeFinalBase64Chunk(chunk, throwOnExtraBits)` — §23.3.3.5
- `DecodeFullLengthBase64Chunk(chunk)` — §23.3.3.6
- `FromBase64(string, alphabet, lastChunkHandling [, maxLength])` — §23.3.3.7
- `FromHex(string [, maxLength])` — §23.3.3.8

## Expected Behavior

```js
// Encoding
const bytes = new Uint8Array([72, 101, 108, 108, 111]);
bytes.toBase64();  // "SGVsbG8="
bytes.toHex();     // "48656c6c6f"

// Decoding
Uint8Array.fromBase64("SGVsbG8=");        // Uint8Array [72, 101, 108, 108, 111]
Uint8Array.fromHex("48656c6c6f");         // Uint8Array [72, 101, 108, 108, 111]

// Options: base64url alphabet
bytes.toBase64({ alphabet: "base64url" });
Uint8Array.fromBase64("SGVsbG8", { alphabet: "base64url", lastChunkHandling: "loose" });

// In-place set operations
const target = new Uint8Array(10);
target.setFromBase64("SGVsbG8=");   // returns { read: 8, written: 5 }
target.setFromHex("48656c6c6f");    // returns { read: 10, written: 5 }
```

## Implementation Hints

- **Location**: `core/engine/src/builtins/typed_array/` — either extend the existing `Uint8Array` builtin or add a new `uint8array_extras.rs` module  
- **Rust ecosystem**: The `base64` and `hex` crates can accelerate implementation, but the spec algorithms should be followed precisely for edge cases (whitespace handling, `lastChunkHandling` option, alphabet selection)
- **Test262 coverage**: The `test262` suite includes tests under `test/built-ins/Uint8Array/fromBase64/`, `test/built-ins/Uint8Array/fromHex/`, `test/built-ins/Uint8Array/prototype/toBase64/`, `test/built-ins/Uint8Array/prototype/toHex/`, etc.
- **Other engines**: V8, SpiderMonkey, and JavaScriptCore all ship these methods

## Why this matters

Base64/hex encoding is ubiquitous in JavaScript — from `fetch` response handling to WebCrypto to working with binary protocols. Until these methods existed, JS developers had to rely on `atob`/`btoa` (which work on strings, not bytes) or third-party libraries. These spec-mandated methods are the correct, type-safe, standards-compliant way to do it.

magic-akari:
- #5170


magic-akari:
> [!NOTE]
>
>`data-encoding` gives us first-error boundaries (`read`/`written`/`position`), but `stop-before-partial` requires the last spec-acceptable full-chunk boundary, which is not always the same.
>
>For `setFromBase64`, the spec requires early return once `maxLength` is reached (and trailing garbage after that must be ignored), while `data-encoding` has no output-cap-aware early-stop decode API.
>
>Also, simple 4-character truncation is insufficient: whitespace affects `read`, padding legality depends on exact `=` placement, and `strict`/`loose`/`stop-before-partial` have different tail rules.
>
>At the moment, we do not support `stop-before-partial` semantics.
>
>If we decide to go with `data-encoding` (instead of `simdutf`), one realistic long-term path might involve forking `data-encoding` and add the API we need.

jedel1043:
Thank you, this looks nice. Just to be clear, what would be the full list of missing things from `data-encoding`'s side?

I ask this because I plan to open a discussion with the `data-encoding` maintainers to know if they're interested in adding those features.

magic-akari:
The main missing feature is `LastChunkHandlingOptions`. Currently, `data-encoding` only offers two padding modes: `None` and `Some`, both of which enforce strict padding validation.

However, the ECMA-262 spec requires three distinct modes: `strict("=")`, `loose("=")`, and `StopBeforePartial`.

Our current implementation works around this limitation by checking for the presence of `=` in the input beforehand to toggle between `Some("=")` and `None`. Having native `loose("=")` support in `data-encoding` would eliminate this hack.

There's another subtle issue: ECMA-262 disallows concatenated padded inputs, but `data-encoding` with padding enabled accepts them. A proper `loose` mode shouldn't just be a binary switch—it needs to reject concatenated padded inputs to match the spec.

As for `StopBeforePartial`, we should probably pull some test cases from Test262 to ensure our implementation is correct.

jedel1043:
Thank! I'll open an issue with them to see if they're open to implementing the features.

jedel1043:
https://github.com/ia0/data-encoding/issues/152#issuecomment-4055042614
Things went well! The maintainer plans to implement a "disallow concatenated padded inputs" mode, and they also provided a wrapper function that supports "stop-before-partial", with a plan to integrate that into data-encoding@v3.

Since they seem collaborative and responsive, I'd say we should merge this and try to push for the addition of the features we need.

ia0:
Actually I missed those requirements from https://github.com/boa-dev/boa/pull/4630#issuecomment-3908996494:

> For `setFromBase64`, the spec requires early return once `maxLength` is reached (and trailing garbage after that must be ignored), while `data-encoding` has no output-cap-aware early-stop decode API.

This is something I've never heard of yet. I've added it to the wish list for v3.

> Also, simple 4-character truncation is insufficient: whitespace affects `read`, padding legality depends on exact `=` placement, and `strict`/`loose`/`stop-before-partial` have different tail rules.

I missed the fact you don't use `BASE64` and `BASE64URL` but have custom encodings that ignore whitespaces. The work-around I gave only works without ignored characters.

Is the `maxLength` from above a hard constraint? If yes, I can try to see if there's a work-around that would fix both issues: short output and ignored characters. If there's a solution, then I can probably add it to v2 because it would be generic enough.

magic-akari:
> Is the `maxLength` from above a hard constraint?

Yes, the `setFromBase64` API performs in-place decoding into an existing byte container rather than returning a new instance. Consequently, it must account for complex boundary conditions involving the discrepancy between the decoded data volume and the destination's capacity, including error-handling protocols for overflow. These requirements are meticulously documented in ECMA-262 and validated through Test262 conformance tests.

ia0:
Thanks! I guess this is the specification: https://tc39.es/ecma262/#sec-frombase64. I'll take a look when I get time.

ia0:
Sorry for the long delay. That's the first week-end where I could get a chance to look at this.

You can find [here](https://github.com/ia0/data-encoding/blob/a68dc8c7c6b6a91c9a0919876595a91915b8a825/ecma262/src/lib.rs#L6-L182) an example implementation of `FromBase64` from ECMA-262 using `data-encoding`. If my interpretation of the specification is correct, then the implementation should be correct (I fuzzed it for 10 hours).

There's a few things to note:
- The implementation is not optimized to handle cases where `max_length` is much smaller than `6 * input.len() / 8`. This could be improved with an iterative approach if performance matters in those cases (this would probably still require `output.len()` to be greater than `max_length` by at least 3 bytes though).
- Lines 54 to 77 are not ECMA-262 specific but a way to work around the fact that `data-encoding` doesn't provide a way to decode up to a given output length. I'm not sure how to do this without sacrificing performance yet. I'll continue thinking about it though since it looks like a legitimate use-case and I'd like to have it in v3.
- Lines 79 to 151 are really the ECMA-262 specific logic and I don't think `data-encoding` can provide much more help there. But happy to look into ideas you have.

If using that implementation on your side is satisfactory, then I can create a PR to update `data-encoding` with the change helping performance and readability of the implementation.

magic-akari:
> If using that implementation on your side is satisfactory, then I can create a PR to update `data-encoding` with the change helping performance and readability of the implementation.

Thank you for the update. I have completed local testing of the ecma262 branch and confirmed that it passes all test262 test cases. I believe the implementation is now ready to move forward.

Regarding the ecma262 module - specifically the `decode` and `decode_mut` implementations - I have a question about its long-term maintenance. You previously mentioned that this is an "example implementation," but given that it has already reached a high degree of usability, will it be officially published and maintained within data-encoding? Or is the expectation for boa to implement and maintain its own version?

ia0:
Yes, my idea was for boa to copy and maintain the implementation locally. An alternative would be for me (or us) to publish and maintain (or co-maintain) a `base64-ecma262` crate with that implementation. There shouldn't be much maintenance anyway since the standard shouldn't change often. The implementation might change once data-encoding v3 is stable (in a few years) to benefit from possible performance and code-size improvements, and also simplifying the implementation from new features. I can take the maintenance responsibility for this change, that's not a problem. It's just that I don't want code specific to a particular standard in data-encoding, which should remain a generic crate (fast to compile with no dependencies).

magic-akari:
I have updated the implementation to use the ecma262 branch of data-encoding as a dependency, which is expected to pass all tests. I will update the dependency again once a new version of data-encoding is officially released.

Regarding the maintenance strategy for the ECMA-262 logic—whether it resides in a separate crate or within Boa—I will leave the final decision to @jedel1043, as I am contributing to this project as an external contributor.

magic-akari:
> There's a few things to note:
> 
> * The implementation is not optimized to handle cases where `max_length` is much smaller than `6 * input.len() / 8`. This could be improved with an iterative approach if performance matters in those cases (this would probably still require `output.len()` to be greater than `max_length` by at least 3 bytes though).
> * Lines 54 to 77 are not ECMA-262 specific but a way to work around the fact that `data-encoding` doesn't provide a way to decode up to a given output length. I'm not sure how to do this without sacrificing performance yet. I'll continue thinking about it though since it looks like a legitimate use-case and I'd like to have it in v3.
> * Lines 79 to 151 are really the ECMA-262 specific logic and I don't think `data-encoding` can provide much more help there. But happy to look into ideas you have.

Thank you for these technical notes.

jedel1043:
> Regarding the maintenance strategy for the ECMA-262 logic—whether it resides in a separate crate or within Boa—I will leave the final decision to @jedel1043, as I am contributing to this project as an external contributor.

Sounds fine to have a small crate for it (might be useful for other Rust engines such as Nova), but we can leave that for the future. In the meantime let's just bundle the code in `boa_engine`

core/engine/src/builtins/typed_array/base64.rs
```diff
@@ -0,0 +1,256 @@
...
+        while 0 < read && base.interpret_byte(input[read - 1]).is_ignored() {
+            read -= 1;
+        }
+        if extra_input == 0 {
+            break;
+        }
+        // Backtrack one symbol.
+        read -= 1;
+        extra_input -= 1;
+        debug_assert!(base.interpret_byte(input[read]).is_symbol());
```
ia0, line 71: I've merged https://github.com/ia0/data-encoding/pull/154 and decided `is_symbol()` should return an `Option` instead of a `bool` (to avoid losing possibly useful information). So you'll need the following change once the `Cargo.toml` points to the default branch, and once I'll release the change (in about 2 weeks to let it cool down).

```suggestion
        debug_assert!(base.interpret_byte(input[read]).is_symbol().is_some());
```

Cargo.toml
```diff
@@ -147,6 +147,8 @@ strum = { version = "0.28", features = ["derive"] }
 unsend = { version = "0.2.1", default-features = false }
 husky-rs = "0.3.2"
 async-channel = "2.5.0"
+data-encoding = { git = "https://github.com/ia0/data-encoding.git", branch = "ecma262"}
+data-encoding-macro = { git = "https://github.com/ia0/data-encoding.git", branch = "ecma262"}
```
ia0: I've released the change under 2.11.0 so https://github.com/boa-dev/boa/pull/4630#discussion_r3063995122 is now actionable with the change below:

```suggestion
data-encoding = "2.11.0"
data-encoding-macro = "0.1.20"
```

magic-akari:
It's ready for review.

jedel1043 (approved):
Thank you for the contribution! Looks pretty much perfect, and the code seems clean.

jedel1043:
Gonna update the branch to rerun tests and we can merge after that.

PR #4631, opened 2026-02-17
Implement BOM stripping and `ignoreBOM` option in TextDecoder

### LINKED ISSUE #4613: TextDecoder should strip BOM by default
Boa: 
```js
>> new TextDecoder().decode(Uint8Array.of(0xef, 0xbb, 0xbf)).length
1
>> new TextDecoder('utf-16le').decode(Uint8Array.of(0xff, 0xfe)).length
1
>> new TextDecoder('utf-16be').decode(Uint8Array.of(0xfe, 0xff)).length
1
>> new TextDecoder('utf-8', { ignoreBOM: true }).decode(Uint8Array.of(0xef, 0xbb, 0xbf)).length
1
>> new TextDecoder('utf-16le', { ignoreBOM: true }).decode(Uint8Array.of(0xff, 0xfe)).length
1
>> new TextDecoder('utf-16be', { ignoreBOM: true }).decode(Uint8Array.of(0xfe, 0xff)).length
1
>> new TextDecoder('utf-8', { ignoreBOM: false }).decode(Uint8Array.of(0xef, 0xbb, 0xbf)).length
1
>> new TextDecoder('utf-16le', { ignoreBOM: false }).decode(Uint8Array.of(0xff, 0xfe)).length
1
>> new TextDecoder('utf-16be', { ignoreBOM: false }).decode(Uint8Array.of(0xfe, 0xff)).length
1
```

Correct behavior:
```js
> new TextDecoder().decode(Uint8Array.of(0xef, 0xbb, 0xbf)).length
0
> new TextDecoder('utf-16le').decode(Uint8Array.of(0xff, 0xfe)).length
0
> new TextDecoder('utf-16be').decode(Uint8Array.of(0xfe, 0xff)).length
0
> new TextDecoder('utf-8', { ignoreBOM: true }).decode(Uint8Array.of(0xef, 0xbb, 0xbf)).length
1
> new TextDecoder('utf-16le', { ignoreBOM: true }).decode(Uint8Array.of(0xff, 0xfe)).length
1
> new TextDecoder('utf-16be', { ignoreBOM: true }).decode(Uint8Array.of(0xfe, 0xff)).length
1
> new TextDecoder('utf-8', { ignoreBOM: false }).decode(Uint8Array.of(0xef, 0xbb, 0xbf)).length
0
> new TextDecoder('utf-16le', { ignoreBOM: false }).decode(Uint8Array.of(0xff, 0xfe)).length
0
> new TextDecoder('utf-16be', { ignoreBOM: false }).decode(Uint8Array.of(0xfe, 0xff)).length
0
```

Spec: https://encoding.spec.whatwg.org/

apoorvdarshan:
## Summary
- Strip BOM bytes by default during decoding (UTF-8, UTF-16LE, UTF-16BE) per the WHATWG Encoding spec
- Support the `ignoreBOM` constructor option to opt out of stripping
- Add read-only `ignoreBOM` getter property

Closes #4613

## Test plan
- [x] Added tests for default BOM stripping (UTF-8, UTF-16LE, UTF-16BE)
- [x] Added tests for `ignoreBOM: true` preserving BOM for all three encodings
- [x] Added tests for `ignoreBOM: false` stripping BOM (same as default)
- [x] Added test for `ignoreBOM` getter returning correct values
- [x] All 50 `boa_runtime` tests pass with zero warnings

### DIFF
--- a/core/runtime/src/text/tests.rs
+++ b/core/runtime/src/text/tests.rs
@@ -157,19 +157,24 @@ fn roundtrip(encoding: &'static str) {
 }
 
 // Default behavior: BOM is stripped
-#[test_case("utf-8", "0xEF, 0xBB, 0xBF, 72, 105")]
-#[test_case("utf-16le", "0xFF, 0xFE, 72, 0, 105, 0")]
-#[test_case("utf-16be", "0xFE, 0xFF, 0, 72, 0, 105")]
-fn decoder_bom_default_stripped(encoding: &'static str, bytes: &'static str) {
+#[test_case("utf-8", &[0xEF, 0xBB, 0xBF, 72, 105])]
+#[test_case("utf-16le", &[0xFF, 0xFE, 72, 0, 105, 0])]
+#[test_case("utf-16be", &[0xFE, 0xFF, 0, 72, 0, 105])]
+fn decoder_bom_default_stripped(encoding: &'static str, bytes: &'static [u8]) {
     let context = &mut Context::default();
     text::register(None, context).unwrap();
+
+    let input = JsUint8Array::from_iter(bytes.iter().copied(), context).unwrap();
+    context
+        .register_global_property(js_str!("input"), input, Attribute::default())
+        .unwrap();
 
     run_test_actions_with(
         [
             TestAction::run(format!(
                 r#"
                 const d = new TextDecoder({encoding:?});
-                decoded = d.decode(Uint8Array.from([{bytes}]));
+                decoded = d.decode(input);
             "#
             )),
             TestAction::inspect_context(|context| {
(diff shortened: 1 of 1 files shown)


core/runtime/src/text/tests.rs
```diff
@@ -155,3 +155,250 @@ fn roundtrip(encoding: &'static str) {
         context,
     );
 }
+
```
jedel1043, line 158: Can this set of tests be simplified by using the `#[test_case]` macro? Above this there's a pretty good example on how to do this.

apoorvdarshan: Done! Refactored the 9 individual tests into 3 parameterized `#[test_case]` functions, following the `roundtrip` pattern above.

apoorvdarshan:
Thanks for the review! Refactored the 9 individual BOM tests into 3 parameterized `#[test_case]` functions, following the `roundtrip` pattern. Also fixed the rustfmt issue from the earlier push.

core/runtime/src/text/encodings.rs
```diff
@@ -29,7 +34,11 @@ pub(crate) mod utf16le {
         }
     }
 
-    pub(crate) fn decode(mut input: &[u8]) -> JsString {
+    pub(crate) fn decode(mut input: &[u8], strip_bom: bool) -> JsString {
+        if strip_bom {
+            input = input.strip_prefix(&[0xFF, 0xFE]).unwrap_or(input);
+        }
+
```
jedel1043: This kind of uses two different ways of doing the same thing. Maybe just use the approach in `utf16le`? Seems simpler to understand.

apoorvdarshan: Done! Updated all three decode functions to use the same `if strip_bom { input = ... }` pattern from `utf16le`. Also switched `utf16be` to use `drain` instead of reallocating with `.to_vec()`.

core/runtime/src/text/encodings.rs
```diff
@@ -59,7 +66,13 @@ pub(crate) mod utf16be {
         }
     }
 
-    pub(crate) fn decode(mut input: Vec<u8>) -> JsString {
+    pub(crate) fn decode(mut input: Vec<u8>, strip_bom: bool) -> JsString {
+        if strip_bom {
+            if input.starts_with(&[0xFE, 0xFF]) {
+                input.drain(..2);
+            }
```
jedel1043: Uhh what? As I've mentioned, just do the same as `utf16le` and strip_prefix.

jedel1043: Oh right, this uses a Vec instead of an &[u8]. Sorry! Disregard this

apoorvdarshan: No worries\! Yeah, since it takes `Vec<u8>` for the in-place byte swapping, `strip_prefix` doesn't apply here. Happy to change if you'd prefer a different approach though.

core/runtime/src/text/tests.rs
```diff
@@ -155,3 +155,124 @@ fn roundtrip(encoding: &'static str) {
...
+// Default behavior: BOM is stripped
+#[test_case("utf-8", "0xEF, 0xBB, 0xBF, 72, 105")]
+#[test_case("utf-16le", "0xFF, 0xFE, 72, 0, 105, 0")]
+#[test_case("utf-16be", "0xFE, 0xFF, 0, 72, 0, 105")]
```
jedel1043: This depends on how we format byte buffers, so it's much more prone to false negatives. Can you change those to arrays, and test by using slice or iterator equality instead?

apoorvdarshan: Done! Changed the `#[test_case]` parameters from string-formatted bytes to actual `&[u8]` slices. The tests now create a `JsUint8Array` from the slice directly in Rust and register it as a global, instead of interpolating byte strings into JS. See 177202b3.

jedel1043 (approved):
Looks great!

PR #4632, opened 2026-02-17
Replace the inner of JsStr with JsStrVariant

hansl:
A JavaScript string slice now is simply a pointer to its slice. This simplifies the code greatly. A `SliceString` is now just a string slice.

Benchmark output (`cargo bench -p boa_benches`) is within error threshold.

Combined benchmark results as well:

```
PROGRESS Richards
RESULT Richards 233
PROGRESS DeltaBlue
RESULT DeltaBlue 232
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 194
PROGRESS RayTrace
RESULT RayTrace 464
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 562
PROGRESS RegExp
RESULT RegExp 81.6
PROGRESS Splay
RESULT Splay 781
PROGRESS NavierStokes
RESULT NavierStokes 451
SCORE 307
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -162,7 +162,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Check for typos
-        uses: crate-ci/typos@v1.43.3
+        uses: crate-ci/typos@v1.43.4
 
   clippy:
     name: Lint
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -334,9 +334,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.10.0"
+version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "812e12b5285cc515a9c72a5c1d3b6d46a19dac5acfef5265968c166106e31dd3"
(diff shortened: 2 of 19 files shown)


core/string/src/vtable/slice.rs
```diff
@@ -14,34 +14,35 @@ pub(crate) struct SliceString {
...
                 refcount: slice_refcount,
-                len,
+                len: end - start,
                 kind: JsStringKind::Slice,
             },
             owned: owned.clone(),
-            data,
-            len,
-            is_latin1,
+            // Safety: this whole function is unsafe for the same invariant.
```
nekevss: question: isn't the safety of `as_static` about tracking the lifetime manually?

Wouldn't that mean we want to note the invariant of the managed lifetime here or in the comment?

hansl: It should be at the function level. Is it not? Let me check.

hansl: Nevermind, that was a bit of copy and paste. Fixed, I'll push a new commit.

hansl:
@nekevss PTAL.

hansl:
I run MIRI on `boa_string` when doing these PRs, since it's not run on CI. Maybe it should. And the test coverage is good enough for `boa_string` IMO, and that is run on CI. I'll merge.

nekevss:
> I run MIRI on `boa_string` when doing these PRs, since it's not run on CI

Just to be clear, I'm aware that you run them. I think its more that it would be nice to have it run in CI to some degree as well.

PR #4635, opened 2026-02-20
feat: implement JsArray::splice (Fixes #4548)

### LINKED ISSUE #4548: JsArray.splice method
**JsArray.splice method**
In JsArray I see slice method is implemented but not splice.

**Example code**
```rust
let js_array: JsArray = ...;
js_array.splice(...);
```

Jayant-kernel:
This PR implements the [splice](cci:1://file:///c:/Users/jayan/Desktop/boa/core/engine/src/builtins/array/mod.rs:2260:4-2395:5) method for [JsArray](cci:2://file:///c:/Users/jayan/Desktop/boa/core/engine/src/object/builtins/jsarray.rs:13:0-15:1) in the public Rust API, resolving #4548.
### Changes
- Added `JsArray::splice` in [core/engine/src/object/builtins/jsarray.rs](cci:7://file:///c:/Users/jayan/Desktop/boa/core/engine/src/object/builtins/jsarray.rs:0:0-0:0).
- The implementation delegates to the internal `Array::splice` function, which is efficient and avoids using [get("splice").call()](cci:1://file:///c:/Users/jayan/Desktop/boa/core/engine/src/builtins/array/mod.rs:187:4-189:5).
- This mirrors the behavior of `Array.prototype.splice`, allowing Rust users to insert, remove, and replace elements in a [JsArray](cci:2://file:///c:/Users/jayan/Desktop/boa/core/engine/src/object/builtins/jsarray.rs:13:0-15:1).
### Verification
- Verified that `Array::splice` is `pub(crate)` and accessible within `boa_engine`.
- Ran local tests to confirm correct behavior for adding and removing elements.
- Ran `cargo check -p boa_engine` and `cargo fmt`.

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -2268,14 +2268,24 @@ impl Array {
         args: &[JsValue],
         context: &mut Context,
     ) -> JsResult<JsValue> {
+        let start = args.first();
+        let delete_count = args.get(1);
+        let items = args.get(2..).unwrap_or_default();
+
+        Self::splice_internal(this, start, delete_count, items, context)
+    }
+
+    pub(crate) fn splice_internal(
+        this: &JsValue,
+        start: Option<&JsValue>,
+        delete_count: Option<&JsValue>,
+        items: &[JsValue],
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
         // 1. Let O be ? ToObject(this value).
         let o = this.to_object(context)?;
         // 2. Let len be ? LengthOfArrayLike(O).
         let len = o.length_of_array_like(context)?;
-
-        let start = args.first();
-        let delete_count = args.get(1);
-        let items = args.get(2..).unwrap_or_default();
 
         // 3. Let relativeStart be ? ToIntegerOrInfinity(start).
         // 4. If relativeStart = -∞, let actualStart be 0.
--- a/core/engine/src/object/builtins/jsarray.rs
+++ b/core/engine/src/object/builtins/jsarray.rs
@@ -331,6 +331,52 @@ impl JsArray {
         Self::from_object(object)
     }
 
+    /// Calls `Array.prototype.splice()`.
+    ///
+    /// Removes and/or inserts elements from the array, returning the removed elements.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use boa_engine::{Context, JsValue};
+    /// # use boa_engine::object::builtins::JsArray;
+    /// let context = &mut Context::default();
+    /// let array = JsArray::from_iter([1, 2, 3].map(JsValue::from), context);
+    ///
+    /// // Insert elements at index 1 without removing
+    /// let removed = array.splice(
+    ///     1,
+    ///     Some(0),
+    ///     &[JsValue::from(10), JsValue::from(20)],
+    ///     context,
+    /// ).unwrap();
+    ///
+    /// assert_eq!(array.length(context).unwrap(), 5);
+    /// assert_eq!(removed.length(context).unwrap(), 0);
+    /// ```
+    #[inline]
+    pub fn splice(
+        &self,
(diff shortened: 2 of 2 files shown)


Jayant-kernel:
Hi Hi @carlos-verdes
review my pr when free

core/engine/src/object/builtins/jsarray.rs
```diff
@@ -331,6 +331,28 @@ impl JsArray {
...
+    ) -> JsResult<Self> {
+        let mut args = vec![
+            start.into_or_undefined(),
+            delete_count.into_or_undefined(),
+        ];
+        args.extend_from_slice(items);
+
+        let object = Array::splice(&self.inner.clone().into(), &args, context)?
+            .as_object()
+            .expect("Array.prototype.splice should always return object");
```
jedel1043: Hmm, I don't think having to build a vec with the args is the most efficient approach. Maybe create a private method based off of `Array::splice` which receives the options directly? Then we would call that from both `JsArray::splice` and `Array::splice`.

core/engine/src/object/builtins/jsarray.rs
```diff
@@ -440,3 +462,71 @@ impl TryFromJs for JsArray {
...
+                Some(1),
+                Some(0),
+                &[JsValue::from(10), JsValue::from(20)],
+                context,
+            )
+            .unwrap();
+
+        assert_eq!(array.length(context).unwrap(), 5);
+        assert_eq!(removed.length(context).unwrap(), 0);
+    }
```
jedel1043, line 515: Can you add this as a doc comment? Seems useful to have as documentation.

core/engine/src/object/builtins/jsarray.rs
```diff
@@ -331,6 +331,28 @@ impl JsArray {
         Self::from_object(object)
     }
 
+    /// Calls `Array.prototype.splice()`.
+    #[inline]
+    pub fn splice(
+        &self,
+        start: Option<u32>,
```
jedel1043: Start is always required, so this should be a `u32`.

Jayant-kernel:
@jedel1043 
Done! Changes

jedel1043:
@Jayant-kernel can you explain why would you want to "match the file's style" and not add a documentation example? it is kind of obvious that having a doc comment code snippet is much better than just having a single line of documentation.

jedel1043:
First warning.

If you're not using an LLM, don't ignore reviews from maintainers without justifying why the review was ignored.
If you're using an LLM, don't push the changes made by the agent without checking that they properly address the review first.

Jayant-kernel:
@jedel1043 Sorry for the late reply, and sorry for the confusion earlier.

You’re right — I should have addressed the review more carefully before pushing. I was trying to keep the style consistent, but I understand now that a doc example is the better choice here.

I’ve updated the PR to add a proper documentation example for `JsArray::splice`, and I also refactored the implementation to call the internal splice path directly (`splice_internal`) instead of building a temporary args vec in the public API.

Please let me know if you’d like any changes to the example or test coverage.

jedel1043:
Fix all CI checks first.

core/engine/src/builtins/array/mod.rs
```diff
@@ -2287,8 +2297,13 @@ impl Array {
...
 
-        let actual_delete_count =
-            Self::get_delete_count(len, actual_start, start, delete_count, context)?;
+        let actual_delete_count = Self::get_delete_count(
+            len,
+            actual_start,
+            start,
+            delete_count,
+            context,
+        )?;
```
jedel1043: note to self: in the future we might need to cleanup all these utility methods to see if we can start reworking them in a way that the methods in JsArray don't need to convert to JsValue to call them.

Jayant-kernel: done

jedel1043 (approved):
Cool! Thanks for the contribution

PR #4636, opened 2026-02-20
fix: TextDecoder.decode() to respect TypedArray byteOffset and byteLength

### LINKED ISSUE #4615: TextDecoder returns irrelevant underlying buffer data
```console
>> new TextDecoder().decode(Uint8Array.of(0x42, 0x42).subarray(1))
"BB"
```

As this leaks unrelated data, this could have potential security impact in some setups, when the buffer has both private and public parts.

abhinavs1920:
This Pull Request fixes/closes #4615.

It changes the following:

## Summary

Fixed a bug where `TextDecoder.decode()` was leaking data from the underlying buffer when decoding TypedArray subarrays. The issue was that it decoded the entire ArrayBuffer instead of respecting the TypedArray's `byteOffset` and `byteLength`.

## Changes

- Updated `TextDecoder.decode()` in `core/runtime/src/text/mod.rs` to properly handle TypedArray views by extracting only the relevant slice from the buffer
- Added support for DataView inputs (which had the same issue)
- Added tests to verify the fix works correctly for both TypedArray subarrays and DataView

## Example

Before:
```js
new TextDecoder().decode(Uint8Array.of(0x42, 0x42).subarray(1))
// Returns "BB" (incorrect - leaks data from index 0)
```
After:
```js
new TextDecoder().decode(Uint8Array.of(0x42, 0x42).subarray(1))
// Returns "B" (correct - only decodes the subarray)
```

I have checked all the tests are passing and this fix does not break anything existing. Along-with the fix I have also added the test for the fix.

Also I am adding up screenshots of the tests passed.
Screenshot from 2026-02-20 11-42-22

Screenshot from 2026-02-20 11-43-24
Screenshot from 2026-02-20 11-46-25

### DIFF
--- a/core/runtime/src/text/mod.rs
+++ b/core/runtime/src/text/mod.rs
@@ -2,7 +2,7 @@
 //!
 //! See <https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API> for more information.
 
-use boa_engine::object::builtins::{JsArrayBuffer, JsTypedArray, JsUint8Array};
+use boa_engine::object::builtins::{JsArrayBuffer, JsTypedArray, JsUint8Array, JsDataView};
 use boa_engine::realm::Realm;
 use boa_engine::value::TryFromJs;
 use boa_engine::{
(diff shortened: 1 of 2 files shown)


abhinavs1920:
Heyy everyone..

I have 2 questions related to the current fix.

1. I'm currently calling `.to_vec()` to extract the buffer slice, which creates a copy. I considered using `subslice()` (like in the atomics implementation), but that would require holding a borrow across the encoding functions. Is the current approach acceptable, or should we explore a zero-copy solution?

2. I'm returning a `RangeError` when the TypedArray/DataView access is out of bounds. Should this be a `TypeError` instead, or is `RangeError` more appropriate here? I noticed the spec doesn't explicitly define this error case.

jedel1043:
I'm very sorry for the trouble! Gonna have to close this because a different person claimed the issue first. Feel free to contribute to other unclaimed issues.

PR #4637, opened 2026-02-20
Fix TextDecoder to respect TypedArray view bounds #4615

Rahul-2k4:
## Summary
This fixes `TextDecoder.decode()` for TypedArray views so decoding respects the view window (`byteOffset..byteOffset+byteLength`) instead of decoding the full backing `ArrayBuffer`.

## Security impact
Before this change, decoding a `subarray()` could include bytes outside the intended view, which may leak unrelated data in embeddings where a buffer contains both public and private regions.

## Repro (before)
```js
new TextDecoder().decode(Uint8Array.of(0x42, 0x42).subarray(1))
// "BB" (incorrect)
```

## Behavior after
```js
new TextDecoder().decode(Uint8Array.of(0x42, 0x42).subarray(1))
// "B" (correct)
```

## Changes
- `core/runtime/src/text/mod.rs`
  - In `TextDecoder::decode`, TypedArray inputs now:
    - read `byteOffset` and `byteLength`
    - slice decoded bytes to that range before decoding
- `core/runtime/src/text/tests.rs`
  - Added regression tests:
    - `decoder_js_respects_typed_array_subarray`
    - `decoder_js_respects_typed_array_subarray_end`

This Pull Request fixes/closes https://github.com/boa-dev/boa/issues/4615.

### DIFF
--- a/core/runtime/src/text/mod.rs
+++ b/core/runtime/src/text/mod.rs
@@ -90,12 +90,12 @@ impl TextDecoder {
                 JsUint8Array::from_array_buffer(JsArrayBuffer::from_object(buffer)?, context)?;
             let backing_buffer_length = bytes.length(context)?;
 
-            let Some(end) = byte_offset.checked_add(byte_length) else {
+            let Some(_end) = byte_offset
+                .checked_add(byte_length)
+                .filter(|end| *end <= backing_buffer_length)
+            else {
                 return Err(js_error!(TypeError: "Invalid TypedArray byte range."));
             };
-            if end > backing_buffer_length {
-                return Err(js_error!(TypeError: "Invalid TypedArray byte range."));
-            }
 
             bytes
                 .iter(context)


core/runtime/src/text/mod.rs
```diff
@@ -76,20 +76,34 @@ impl TextDecoder {
...
+            let byte_offset = typed_array.byte_offset(context)?;
+            let byte_length = typed_array.byte_length(context)?;
             let Some(buffer) = typed_array.buffer(context)?.as_object() else {
                 return Err(js_error!(TypeError: "Invalid buffer backing TypedArray."));
             };
-            JsUint8Array::from_array_buffer(JsArrayBuffer::from_object(buffer)?, context)?
+            let bytes =
+                JsUint8Array::from_array_buffer(JsArrayBuffer::from_object(buffer)?, context)?
+                    .iter(context)
+                    .collect::<Vec<u8>>();
```
jedel1043: Do you need to collect here? if you have the offset and end, you could just `skip()` then `take()`, right?

Rahul-2k4: Good catch. We don’t need to collect the full buffer there. I’ll change it to iterate with skip(byte_offset) + take(byte_length) and collect only the relevant bytes.

core/runtime/src/text/mod.rs
```diff
@@ -76,20 +76,38 @@ impl TextDecoder {
...
+            let byte_length = typed_array.byte_length(context)?;
             let Some(buffer) = typed_array.buffer(context)?.as_object() else {
                 return Err(js_error!(TypeError: "Invalid buffer backing TypedArray."));
             };
-            JsUint8Array::from_array_buffer(JsArrayBuffer::from_object(buffer)?, context)?
+            let bytes =
+                JsUint8Array::from_array_buffer(JsArrayBuffer::from_object(buffer)?, context)?;
+            let backing_buffer_length = bytes.length(context)?;
+
+            let Some(end) = byte_offset.checked_add(byte_length) else {
```
jedel1043: You can use `Option::filter` here to avoid the double let else

Rahul-2k4: sure i will fix so it will be more clean and minimise redundancy

zhuzhu81998:
This needs to be rebased, and I think while the comment from awhile back claims that `buffer` could be array buffer, typed array or data view, the data view part is also not actually implemented properly.

PR #4638, opened 2026-02-20
Fix/headers iterable #4611

### LINKED ISSUE #4611: Headers should be iterable
```console
% boa
>> [...new Headers([['x','y']])]
Uncaught: TypeError: value with type `object` is not iterable (native at core/engine/src/builtins/iterable/mod.rs:307:13)
    at <main> (unknown at :?:?)
>> new Map(new Headers([['x','y']]))
Uncaught: TypeError: value with type `object` is not iterable (native at core/engine/src/builtins/iterable/mod.rs:307:13)
    at <main> (unknown at :1:1)
```

```console
% node
> [...new Headers([['x','y']])]
[ [ 'x', 'y' ] ]
> new Map(new Headers([['x','y']]))
Map(1) { 'x' => 'y' }
```

Rahul-2k4:
## Summary

  This PR makes `Headers` iterable in Boa runtime so it can be consumed by iterable APIs.

This Pull Request closes #4611 

  ### What changed

  - Added a regression test for `Headers` iterability:
    - spread: `[...new Headers([["x", "y"]])]`
    - `Map` construction: `new Map(new Headers([["x", "y"]]))`
  - Wired `Headers.prototype[Symbol.iterator]` during fetch registration.
  - Exposed `entries`, `keys`, and `values` as class methods on `Headers`.

  ### Why

  `Headers` should be iterable over `[name, value]` pairs. Without `@@iterator`, spread and
  `Map` construction throw a `TypeError`.

  ## Testing

  ```bash
  cargo test -p boa_runtime fetch::tests -- --nocapture

  All fetch tests pass, including the new regression test.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,8 @@
+# Fallback to all maintainers for common directories.
+
+* @boa-dev/maintainers
+
+# Owned components
+
+core/engine/src/builtins/intl @jedel1043 @nekevss
+core/engine/src/builtins/temporal @jasonwilliams @jedel1043 @nekevss
--- a/core/ast/src/scope_analyzer.rs
+++ b/core/ast/src/scope_analyzer.rs
@@ -456,7 +456,9 @@ impl<'ast> VisitorMut<'ast> for BindingEscapeAnalyzer<'_> {
             ClassElement::PrivateFieldDefinition(field)
             | ClassElement::PrivateStaticFieldDefinition(field) => {
                 if let Some(e) = &mut field.initializer {
+                    std::mem::swap(&mut self.scope, &mut field.scope);
                     self.visit_expression_mut(e)?;
+                    std::mem::swap(&mut self.scope, &mut field.scope);
                 }
                 ControlFlow::Continue(())
             }
--- a/core/engine/src/builtins/date/tests.rs
+++ b/core/engine/src/builtins/date/tests.rs
@@ -678,27 +678,53 @@ fn date_proto_set_utc_full_year() {
 
 #[test]
 fn date_proto_set_utc_hours() {
+    // The local-time constructor `new Date(2020, 6, 8, 9, 16, 15, 779)` converts
+    // local time to UTC internally. The resulting UTC components depend on the
+    // host timezone offset, so we compute them dynamically.
+    let base = from_local(2020, 7, 8, 9, 16, 15, 779).to_offset(time::UtcOffset::UTC);
+    let base_y = base.year();
+    let base_m = base.month() as u8;
+    let base_d = base.day();
+    let base_min = base.minute();
+    let base_sec = base.second();
+    let base_ms = base.millisecond();
+
     run_test_actions([
+        // setUTCHours(11): only the UTC hour changes; date/min/sec/ms stay.
         TestAction::assert_eq(
             "new Date(2020, 6, 8, 9, 16, 15, 779).setUTCHours(11)",
-            timestamp_from_utc(2020, 7, 8, 11, 16, 15, 779),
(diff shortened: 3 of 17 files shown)


core/runtime/src/fetch/headers.rs
```diff
@@ -244,6 +245,7 @@ impl JsHeaders {
     /// Returns an iterator allowing you to go through all keys of the key/value pairs
     /// contained in this object.
     #[allow(clippy::unused_self)]
+    #[boa(method)]
```
jedel1043: Why add the `method` attribute? If the function starts with `&self`, it will automatically be exposed as a method.

Rahul-2k4: Yep  right, &self methods are already exposed auto, so [boa(method)] is redundant . I will remove the extra attribute from`keys values .entries.

jedel1043:
If you're using AI agents to do this work, maybe you should use something like [git-worktree](https://git-scm.com/docs/git-worktree) to avoid accidentally pushing unrelated commits? Just a suggestion.

Rahul-2k4:
yep sure @jedel1043 ill keep that in mind . i got little confused because of some uncommited changes of my other pr . and again thank you for the suggestion of worktree that it will keep my workspace clean . Will keep in mind and ill surely improve from next time and thanks for the quick reviews

core/runtime/src/fetch/headers.rs
jedel1043: uhhh so? nothing changed...

Rahul-2k4: my appologies , now i have correctly implemented on suggestion for the removal of redundant #[boa(method)] from entries/keys/values in core/runtime/src/fetch/headers.rs at [commit](https://github.com/boa-dev/boa/pull/4638/commits/ac29992edb8e0fe3e5140616b81a19f5bfde9706)

core/runtime/src/fetch/mod.rs
```diff
@@ -191,7 +212,33 @@ pub fn register<F: Fetcher>(
...
+
+    headers_proto.define_property_or_throw(
+        JsSymbol::iterator(),
+        PropertyDescriptor::builder()
+            .value(iterator)
+            .writable(true)
+            .enumerable(false)
+            .configurable(true),
+        context,
+    )?;
```
jedel1043, line 239: Oof, seems kind of ugly having to do all of this just to add a symbol method. Can you open an issue about adding support for symbols in the `#[boa(class)]` macro, then add a TODO here to change it later?

Rahul-2k4: ok i have raised an issue for the same

core/runtime/src/fetch/mod.rs
```diff
@@ -176,6 +178,25 @@ pub mod js_module {
...
+        })?;
+
+    let entries = headers.entries(context);
+    let entries_obj = entries.to_object(context)?;
+    let method = entries_obj
+        .get(JsSymbol::iterator(), context)?
+        .as_object()
+        .ok_or_else(|| js_error!(TypeError: "entries result is not iterable"))?;
+
+    method.call(&entries, &[], context)
```
jedel1043: I think here you should be able to wrap `headers.entries(context)` with a `JsArray` then call `values` IIRC. That should avoid having to do this dance of manually fetching the iterator method.

Rahul-2k4: Hmmmm I think that's much better approach. Nice , will resolve this now

jedel1043 (approved):
Looks great! We'll hopefully remove all the boilerplate of setting up the method symbol when #4697 gets merged.

Rahul-2k4:
@jedel1043 yep sure

PR #4641, opened 2026-02-20
Add test for response constructor with body and status

### LINKED ISSUE #4547: Response constructor has dummy implementation
**Describe the bug**

[Response](https://github.com/boa-dev/boa/blob/main/core/runtime/src/fetch/response.rs#L170) constructor has dummy implementation 

```rust
#[boa_class(rename = "Response")]
#[boa(rename_all = "camelCase")]
impl JsResponse {

    #[boa(constructor)]
    fn constructor(_body: Option<JsValue>, _options: JsResponseOptions) -> Self {
        Self::basic(js_string!(""), http::Response::new(Vec::new()))
    }
```

**To Reproduce**

  Both parameters are prefixed with _ and never used. This means:

```js
  const response = new Response('Hello World', { status: 404 });
  console.log(await response.text()); // Returns "" (empty string)
  console.log(response.status);       // Returns 200 (default)
```

**Expected behavior**
Explain what you expected to happen, and what is happening instead.

  The constructor should parse the body and options:

```js
  const response = new Response('Hello World', { status: 404 });
  console.log(await response.text()); // Should return "Hello World"
  console.log(response.status);       // Should return 404
```

- OS: macOS (Darwin)
- Version: 24.6.0
- Rustc version: rustc 1.90.0 (1159e78c4 2025-09-14)

**Additional context**

Workaround : 

```rust
        // Override Response constructor with our working implementation
        // (until Boa fixes their constructor to actually use the body parameter)
        // TODO: Remove this when https://github.com/boa-dev/boa/issues/XXXXX is fixed
        let setup_response = r#"
            // Save native Response class
            const NativeResponse = globalThis.Response;
            const nativeFetch = globalThis.fetch;

            // Override Response constructor to make it actually work
            globalThis.Response = function(body, init) {
                init = init || {};
                this.body = String(body || '');
                this.status = init.status || 200;
                this.headers = init.headers || {};
                this.text = async function() {
                    return this.body;
                };
            };

            // Keep native fetch (uses NativeResponse internally)
            globalThis.fetch = nativeFetch;
        "#;
```

psg-19:
This Pull Request fixes/closes #4547 .

Summary :- The `Response` constructor was a dummy implementation that completely ignored its `body` and `options` parameters, always creating an empty response with status 200. The constructor now properly parses the `body` argument, reads `status` and `headers` from the options object, validates the status code and returns `JsResult<Self>` to handle errors .
Also added the "hello world" test mentioned in the issue for testing.

### DIFF
--- a/core/runtime/src/fetch/tests/response.rs
+++ b/core/runtime/src/fetch/tests/response.rs
@@ -157,6 +157,29 @@ fn response_getter() {
 }
 
 #[test]
+fn response_constructor_with_body_and_status() {
+    run_test_actions([
+        TestAction::harness(),
+        TestAction::inspect_context(|ctx| register(&[], ctx)),
+        TestAction::run(
+            r#"
+                globalThis.response = (async () => {
+                    const response = new Response('Hello World', { status: 404 });
+                    assertEq(response.status, 404);
+                    assertEq(response.type, "default");
+                    const text = await response.text();
+                    assertEq(text, "Hello World");
+                })();
+            "#,
+        ),
+        TestAction::inspect_context(|ctx| {
+            let response = ctx.global_object().get(js_str!("response"), ctx).unwrap();
+            response.as_promise().unwrap().await_blocking(ctx).unwrap();
+        }),
+    ]);
+}
+
+#[test]
 fn response_redirect_default_status() {
     run_test_actions([
         TestAction::harness(),
--- a/core/runtime/src/fetch/tests/response.rs
+++ b/core/runtime/src/fetch/tests/response.rs
@@ -464,3 +464,26 @@ fn response_clone_preserves_status() {
         }),
     ]);
 }
+
+#[test]
+fn response_constructor_with_body_and_status() {
+    run_test_actions([
+        TestAction::harness(),
+        TestAction::inspect_context(|ctx| register(&[], ctx)),
+        TestAction::run(
+            r#"
+                globalThis.response = (async () => {
+                    const response = new Response('Hello World', { status: 404 });
+                    assertEq(response.status, 404);
+                    assertEq(response.type, "default");
+                    const text = await response.text();
+                    assertEq(text, "Hello World");
+                })();
+            "#,
+        ),
+        TestAction::inspect_context(|ctx| {
+            let response = ctx.global_object().get(js_str!("response"), ctx).unwrap();
+            response.as_promise().unwrap().await_blocking(ctx).unwrap();
+        }),
+    ]);
+}


core/runtime/src/fetch/response.rs
```diff
@@ -166,8 +166,45 @@ impl JsResponse {
     }
 
     #[boa(constructor)]
-    fn constructor(_body: Option<JsValue>, _options: JsResponseOptions) -> Self {
-        Self::basic(js_string!(""), http::Response::new(Vec::new()))
+    fn constructor(
+        body: Option<JsValue>,
+        options: Option<JsResponseOptions>,
+        context: &mut Context,
+    ) -> JsResult<Self> {
```
jedel1043: Can you link to the specific part of `https://fetch.spec.whatwg.org` for this? Seems kind of useful to have for future reference.

core/runtime/src/fetch/response.rs
```diff
@@ -166,8 +166,45 @@ impl JsResponse {
...
+                body_val
+                    .to_string(context)?
+                    .to_std_string_lossy()
+                    .into_bytes()
+            } else {
+                Vec::new()
+            }
+        } else {
+            Vec::new()
+        };
```
jedel1043: Oof, this seems overly complex. `to_string` is kind of already doing the first `let Some` so you should be able to do something like

```suggestion
        let body_bytes = if let Some(body) = body && !body.is_null_or_undefined() {
            body.to_string(context)?.to_std_string_lossy().into_bytes()
        } else {
            Vec::new()
        };
```

jedel1043 (changes requested):
Looking nice! I just have a couple of suggestions

core/runtime/src/fetch/response.rs
psg-19: Done

core/runtime/src/fetch/response.rs
```diff
@@ -165,9 +165,45 @@ impl JsResponse {
...
+    ///
+    /// More information:
+    ///  - [WHATWG Fetch spec][spec]
+    ///
+    /// [spec]: https://fetch.spec.whatwg.org/#dom-response
     #[boa(constructor)]
-    fn constructor(_body: Option<JsValue>, _options: JsResponseOptions) -> Self {
-        Self::basic(js_string!(""), http::Response::new(Vec::new()))
+    fn constructor(
+        body: Option<JsValue>,
```
jedel1043:
```suggestion
        body: JsValue
```
It's kinda unnecessary to use an `Option` if we're gonna check for `null_or_undefined` at the end of the day

core/runtime/src/fetch/response.rs
```diff
@@ -165,9 +165,45 @@ impl JsResponse {
...
+        options: Option<JsResponseOptions>,
+        context: &mut Context,
+    ) -> JsResult<Self> {
+        let (status_code, headers) = match options {
```
jedel1043: For this I think it would be better to derive `Default` on `JsResponseOptions` so that you can just deconstruct it:
```rust
let JsResponseOptions { status_code, headers, .. } = options.unwrap_or_default();
```

jedel1043 (changes requested):

jedel1043:
This looks good, just needs a `cargo fmt` and `cargo clippy`

core/runtime/src/fetch/response.rs
```diff
@@ -165,9 +165,43 @@ impl JsResponse {
...
+        } else {
+            body.to_string(context)?.to_std_string_lossy().into_bytes()
+        };
+
+        Ok(Self {
+            url: js_string!(""),
+            r#type: ResponseType::Basic,
+            status: Some(status),
+            headers,
+            body: Rc::new(body_bytes),
```
jedel1043: This seems to be missing a couple of checks. For example, according to the spec, if the status is not between 200 and 599, it should throw a RangeError. We should have comments that mirror what the spec requires, for easy cross-reference.

jedel1043: That was not the only missing check. The spec has a couple more on its [initialize a response](https://fetch.spec.whatwg.org/#initialize-a-response) procedure. See [this example](https://github.com/boa-dev/boa/blob/aa9a070814e0a2a02a9bc84d559466276c161eba/core/engine/src/builtins/bigint/mod.rs#L174-L208) for the style we follow for spec comments alongside implementation code.

psg-19: Hi, can you please have a look now?

jedel1043:
Needs a rebase

psg-19:
Just added the test, since the constructor was written well and better than I wrote it before when I was resolving the conflict.

jedel1043 (approved):
LGTM

PR #4644, opened 2026-02-20
Add doctests to  JsValue  methods

### LINKED ISSUE #3953: Build out doctests for `JsValue` methods in documentation
Part of #3952.

Add doctests to `JsValue`'s method documentation. The doctests should aim to be in the form of examples and display relevant ways in which the method can be used.

plasmaDestroyer:
This Pull Request fixes/closes #3953.

It changes the following:

- Adds doctest examples to 39 public methods on JsValue covering constructors, type checkers, accessors, and utility methods.
- `to_*` methods requiring `Context` were omitted to keep docs concise, happy to add in a follow-up if desired.

### DIFF
--- a/core/engine/src/value/mod.rs
+++ b/core/engine/src/value/mod.rs
@@ -642,7 +642,7 @@ impl JsValue {
     #[inline]
     #[must_use]
     pub fn is_null_or_undefined(&self) -> bool {
-        self.is_undefined() || self.is_null()
+        self.0.is_null_or_undefined()
     }
 
     /// Returns the number if the value is a finite integral Number value, otherwise `None`.
@@ -885,16 +885,9 @@ impl JsValue {
     /// assert!(!JsValue::new(false).to_boolean());
     /// ```
     #[must_use]
+    #[inline]
     pub fn to_boolean(&self) -> bool {
-        match self.variant() {
-            JsVariant::Symbol(_) | JsVariant::Object(_) => true,
-            JsVariant::String(s) if !s.is_empty() => true,
-            JsVariant::Float64(n) if n != 0.0 && !n.is_nan() => true,
-            JsVariant::Integer32(n) if n != 0 => true,
-            JsVariant::BigInt(n) if !n.is_zero() => true,
-            JsVariant::Boolean(v) => v,
-            _ => false,
-        }
+        self.0.to_boolean()
     }
 
     /// The abstract operation `ToPrimitive` takes an input argument and an optional argument
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,17 @@ jobs:
       - name: Set-up npm config for publishing
         run: npm config set -- '//registry.npmjs.org/:_authToken' "${{ secrets.NPM_TOKEN }}"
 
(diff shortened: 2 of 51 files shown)


core/engine/src/value/mod.rs
```diff
@@ -215,13 +354,40 @@ impl JsValue {
...
     /// [spec]: https://tc39.es/ecma262/#sec-iscallable
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use boa_engine::{Context, JsValue, Source};
+    ///
+    /// let mut context = Context::default();
+    /// // Evaluate a function expression to get a callable value.
+    /// let func = context.eval(Source::from_bytes(b"(function() {})")).unwrap();
```
nekevss: nit: I think this would be better served to use `NativeFunction`.

Of the top of my head, it would look something like the below.

```rust
use boa_engine::{NativeFunction, Context, JsFunction};

let native_function = NativeFunction::from_copy_closure(|_, _, _,| {
    Ok(JsValue::undefined())
});
let js_value = JsValue::from(native_function.to_js_function(context.realm()));
assert!(js_value.is_callable());
```

core/engine/src/value/mod.rs
```diff
@@ -230,13 +396,41 @@ impl JsValue {
...
     /// Returns true if the value is a constructor object.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use boa_engine::{Context, JsValue, Source};
+    ///
+    /// let mut context = Context::default();
+    /// // Classes and regular functions are constructors.
+    /// let class = context.eval(Source::from_bytes(b"(class {})")).unwrap();
```
nekevss, line 429: note: this probably doesn't have to be a rust based JS builtin

Off the top of my head, I think this is the best way to handle this currently, so it doesn't need to be updated. But it may be nice to have this a bit further built out in the future

core/engine/src/value/mod.rs
```diff
@@ -246,13 +440,39 @@ impl JsValue {
...
 
     /// Returns true if the value is a promise object.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use boa_engine::{Context, JsValue, Source};
+    ///
+    /// let mut context = Context::default();
+    /// let promise = context.eval(Source::from_bytes(b"new Promise((r) => r())")).unwrap();
```
nekevss: nit: Use `JsPromise` in this instance.

core/engine/src/value/mod.rs
```diff
@@ -268,20 +488,59 @@ impl JsValue {
...
 
     /// Returns true if the value is a regular expression object.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use boa_engine::{Context, JsValue, Source};
+    ///
+    /// let mut context = Context::default();
+    /// let regex = context.eval(Source::from_bytes(b"/abc/g")).unwrap();
```
nekevss: nit: Use `JsRegExp` here

plasmaDestroyer:
Updated per review feedback. Thanks for the pointers!

nekevss (approved):
Looks good to me 😄 Thanks for your contribution!!!

Issue #4445, opened 2025-09-25
Implement explicit resource management proposal


jedel1043:
https://github.com/tc39/proposal-explicit-resource-management

abhinavs1920:
Heyy @jedel1043 ,
Can you once look at #4649 whenever you are available :)

PR #4649, opened 2026-02-21
Add parser support for explicit resource management (using/await using)

abhinavs1920:
This Pull Request fixes/closes https://github.com/boa-dev/boa/issues/4445

This PR introduces the parser foundation for the TC39 Explicit Resource Management proposal. It’s an initial step and does not yet include runtime support, which will be added gradually in future commits. Because this is a large feature, it will be developed over multiple commits.

It focuses on parsing support for using and await using, along with proper validation, AST integration, and test coverage. Runtime behavior is intentionally not included yet, using declarations are currently compiled as regular lexical bindings, with clear TODOs marking the follow-up work.

All existing parser tests continue to pass, and 5 new tests have been added for the new syntax. An example script and a short implementation roadmap are included to make future work straightforward.

This is the first step of a larger feature and is designed to be safe, incremental, and easy to review. Runtime disposal logic and related built-ins will be added in further commits of my work.

Thanks for reviewing!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,9 +67,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -88,9 +88,9 @@ checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.7"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
+checksum = "52ce7f38b242319f7cabaa6813055467063ecdc9d355bbb4ce0c68908cd8130e"
 dependencies = [
  "utf8parse",
 ]
@@ -826,9 +826,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.60"
+version = "4.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2797f34da339ce31042b27d23607e051786132987f595b02ba4f6a6dffb7030a"
+checksum = "b193af5b67834b676abd72466a96c1024e6a6ad978a1f484bd90b85c94041351"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -836,9 +836,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.60"
+version = "4.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "24a241312cea5059b13574bb9b3861cabf758b879c15190b37b6d6fd63ab6876"
+checksum = "714a53001bf66416adb0e2ef5ac857140e7dc3a0c48fb28b2f10762fc4b5069f"
 dependencies = [
  "anstream",
  "anstyle",
@@ -848,9 +848,9 @@ dependencies = [
 
 [[package]]
 name = "clap_derive"
-version = "4.5.55"
+version = "4.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a92793da1a46a5f2a02a6f4c46c6496b28c43638adea8306fcb0caa1634f24e5"
+checksum = "1110bd8a634a1ab8cb04345d8d878267d57c3cf1b38d91b71af6686408bbca6a"
 dependencies = [
  "heck",
  "proc-macro2",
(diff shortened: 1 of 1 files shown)


core/parser/src/parser/statement/declaration/lexical.rs
```diff
@@ -79,18 +88,60 @@ where
...
+                if cursor
+                    .peek(0, interner)?
+                    .filter(|t| matches!(t.kind(), TokenKind::Keyword((Keyword::Await, false))))
+                    .is_some()
+                {
+                    return Err(Error::general(
+                        "Invalid syntax: 'using await' is not valid, use 'await using' instead",
+                        tok.span().start(),
+                    ));
+                }
```
jedel1043: Is this specified in the proposal? Can you add a link to it if that's the case?

core/parser/src/parser/statement/declaration/lexical.rs
```diff
@@ -79,18 +88,60 @@ where
...
+                        self.loop_init,
+                    )
+                    .parse(cursor, interner)?
+                } else {
+                    return Err(Error::general(
+                        "Unexpected token 'await'",
+                        tok.span().start(),
+                    ));
+                }
+            }
```
jedel1043, line 150: According to the [spec](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=3000#prod-LexicalDeclaration), this is only parsable if `allow_await` is true. Is this checked somewhere else or something?

core/parser/src/parser/statement/declaration/lexical.rs
abhinavs1920: I have removed the "using await" check - I added it thinking it would help developers, but it's not in the spec. The spec actually allows `await` as an identifier in non-async contexts, so `using await = value;` is valid (though confusing).

core/parser/src/parser/statement/declaration/lexical.rs
abhinavs1920: I have added the AllowAwait check for `await using` - I missed that the spec notation `[+Await] await using` means it's only valid in async contexts. Now it properly checks `self.allow_await` and rejects it at the top level.

abhinavs1920:
Heyy @jedel1043, I have made the required updates.
Also updated the tests to verify both behaviors. All 302 parser tests pass now.

core/parser/src/parser/statement/declaration/lexical.rs
jedel1043: confusing language constructs in my Javascript? they would never!

abhinavs1920:
So... I’m thinking of implementing the remaining implementation in separate PRs. Does that work for you, or would you prefer if I handle this iteratively in a single PR?

jedel1043:
Open up multiple PRs please, it makes reviewing much easier

core/parser/src/parser/statement/declaration/lexical.rs
abhinavs1920: :rofl:  Yep, sticking to spec behavior felt like the right call.

core/parser/src/parser/statement/declaration/mod.rs
```diff
@@ -66,27 +66,54 @@ where
...
-            TokenKind::Keyword((Keyword::Const | Keyword::Let, _)) => {
+            TokenKind::Keyword((Keyword::Const | Keyword::Let | Keyword::Using, _)) => {
                 LexicalDeclaration::new(true, self.allow_yield, self.allow_await, false)
                     .parse(cursor, interner)
                     .map(Into::into)
             }
+            TokenKind::Keyword((Keyword::Await, false)) => {
+                // Check if this is `await using`
+                if let Some(next_tok) = cursor.peek(1, interner)? {
+                    if matches!(next_tok.kind(), TokenKind::Keyword((Keyword::Using, false))) {
```
jedel1043:
```suggestion
                if let Some(next_tok) = cursor.peek(1, interner)?
                    && matches!(next_tok.kind(), TokenKind::Keyword((Keyword::Using, false))) {
```

Small nitpick

core/parser/src/parser/statement/mod.rs
```diff
@@ -436,6 +436,20 @@ where
                     .parse(cursor, interner)
                     .map(ast::StatementListItem::from)
             }
+            TokenKind::Keyword((Keyword::Await, false)) => {
+                // Check if this is `await using`
+                if let Some(next_tok) = cursor.peek(1, interner)? {
+                    if matches!(next_tok.kind(), TokenKind::Keyword((Keyword::Using, false))) {
```
jedel1043:
```suggestion
                if let Some(next_tok) = cursor.peek(1, interner)?
                    && matches!(next_tok.kind(), TokenKind::Keyword((Keyword::Using, false))) {
```

core/parser/src/parser/statement/declaration/lexical.rs
```diff
@@ -79,18 +88,57 @@ where
...
+                // `await using` is only valid when [+Await] is true
+                if !self.allow_await.0 {
+                    return Err(Error::general(
+                        "Unexpected token 'await'",
+                        tok.span().start(),
+                    ));
+                }
+                
+                // Check if this is `await using`
+                let next_tok = cursor.peek(0, interner).or_abrupt()?;
```
jedel1043: Doesn't this need to use `peek_no_skip_line_term`? According to the [spec](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=3000#prod-AwaitUsingDeclarationHead), `await using` is only valid if there are no line terminators between both keywords.

I would suggest you do a pass through the spec to see if you can find small caveats like these

jedel1043 (changes requested):

core/parser/src/parser/statement/declaration/lexical.rs
abhinavs1920: Yep, I actually pushed this in the latest update 🙂 Went through the spec properly and added peek_no_skip_line_term() checks in all three spots (lexical.rs, declaration/mod.rs, and statement/mod.rs) with explicit line terminator validation. Also added solid test coverage for the no-line-terminator rule, destructuring rejection, AllowAwait checks, and the other early errors.

abhinavs1920:
Hey @jedel1043, thanks for the review.

Could you run the CI tests so I can check if anything breaks? Also, I think we should have a setting to run the CI automatically once the changes are approved by the maintainer. This would help contributors run tests and make improvements more easily.

I’m also working on the next part of the proposal, but I need these changes to be merged first so I can open a new PR with the continued updates.

jedel1043:
> I think we should have a setting to run the CI automatically once the changes are approved by the maintainer.

That's enabled, but only if you have at least a PR merged into the project

jedel1043:
You could just run tests locally if you want early feedback :)

abhinavs1920:
> That's enabled, but only if you have at least a PR merged into the project

ohh.. that's understandable :sweat_smile:

jedel1043:
Ping me after you fix the formatting so that I can merge this

PR #4650, opened 2026-02-21
Fix UTF-16 TextDecoder unpaired surrogate handling (#4612)

### LINKED ISSUE #4612: utf-16 TextDecoder is incorrect
It should always produce well-formed strings

Instead, this is what happens on `utf-16le` and `utf-16be`:
```
✖ FAIL Common implementation mistakes > Unicode > Invalid input is replaced > utf-16le
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'ab\ud800wx'
- 'ab�wx'
     ^
✖ FAIL Common implementation mistakes > Unicode > Invalid input is replaced > utf-16be
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'ab\ud800wx'
- 'ab�wx'
     ^
```

Libraries/users might rightfully assume that TextDecoder always produces well-formed strings, and not doing so is problematic.

This might have security implications due to:
1. how hashes and signatures behave on non-well-formed strings
2. not following `fatal` option

Rahul-2k4:
## Summary
- make UTF-16 TextDecoder output well-formed strings by replacing unpaired surrogates with `U+FFFD`
- apply the fix to both `utf-16le` and `utf-16be` decode paths
- keep truncation behavior correct for odd-length input (no extra replacement when the last decoded code unit is a high surrogate)

## Tests
- add regression tests for unpaired surrogate replacement in `utf-16le` and `utf-16be`
- add regression tests for truncation edge cases in both endian modes
- verified with:
  - `cargo test -p boa_runtime decoder_utf16`
  - `cargo test -p boa_runtime`
  - `cargo fmt --all --check`
  - `cargo clippy -p boa_runtime --all-targets -- -D warnings`

Closes #4612.

### DIFF
--- a/core/runtime/src/text/tests.rs
+++ b/core/runtime/src/text/tests.rs
@@ -128,14 +128,14 @@ fn decoder_js_invalid() {
 fn utf16le_bytes(code_units: &[u16]) -> Vec<u8> {
     code_units
         .iter()
-        .flat_map(|unit| [*unit as u8, (unit >> 8) as u8])
+        .flat_map(|unit| unit.to_le_bytes())
         .collect()
 }
 
 fn utf16be_bytes(code_units: &[u16]) -> Vec<u8> {
     code_units
         .iter()
-        .flat_map(|unit| [(unit >> 8) as u8, *unit as u8])
+        .flat_map(|unit| unit.to_be_bytes())
         .collect()
 }
 


core/runtime/src/text/encodings.rs
```diff
@@ -59,28 +69,20 @@ pub(crate) mod utf16be {
...
-
-        if dangling {
-            JsString::from(&[JsString::from(&*input), js_string!("\u{FFFD}")])
-        } else {
-            JsString::from(&*input)
-        }
+        let code_units = input
+            .chunks_exact(2)
+            .map(|pair| u16::from_be_bytes([pair[0], pair[1]]))
+            .collect::<Vec<_>>();
```
jedel1043: Why collect here if you can just pass the iterator? We're already using an iterator in `decode_utf16_units`.

zhuzhu81998:
@Rahul-2k4 are you still working on this?

Rahul-2k4:
yep just finished in a while rn final commits are taking place

core/runtime/src/text/encodings.rs
Rahul-2k4: yep its done now

core/runtime/src/text/tests.rs
```diff
@@ -125,6 +125,110 @@ fn decoder_js_invalid() {
...
+                const d = new TextDecoder("utf-16le");
+                for (const { invalid, replaced } of invalid16) {
+                    const input = new Uint8Array(invalid.length * 2);
+                    for (let i = 0; i < invalid.length; i++) {
+                        input[2 * i] = invalid[i] & 0xff;
+                        input[2 * i + 1] = invalid[i] >> 8;
+                    }
+                    assertEq(d.decode(input), String.fromCharCode(...replaced));
+                }
+            "#}),
```
zhuzhu81998: in the spirit of unit tests, it would probably be better to not do this part as JS code. For example if `String.fromCharcode` has a problem this would fail even though `decode` works fine.

core/runtime/src/text/tests.rs
```diff
@@ -125,6 +125,110 @@ fn decoder_js_invalid() {
...
+                ];
+
+                const d = new TextDecoder("utf-16be");
+                for (const { invalid, replaced } of invalid16) {
+                    const input = new Uint8Array(invalid.length * 2);
+                    for (let i = 0; i < invalid.length; i++) {
+                        input[2 * i] = invalid[i] >> 8;
+                        input[2 * i + 1] = invalid[i] & 0xff;
+                    }
+                    assertEq(d.decode(input), String.fromCharCode(...replaced));
```
zhuzhu81998: same here

zhuzhu81998:
Hmm why push commits after approval and then close it before it could be merged?

PR #4651, opened 2026-02-21
Fix unsafe as casts in typed_array with try_from for 32-bit safety

### LINKED ISSUE #4646: Fix unsafe `as` casts in `typed_array`  replace with `try_from` for 32-bit safety
## Problem

The `typed_array` module has multiple `u64 as usize` casts that silently truncate on 32-bit platforms. These casts feed directly into `unsafe` buffer slicing operations, which could cause incorrect memory access.

Example from `core/engine/src/builtins/typed_array/object.rs` (line 631):
```rust
let byte_index = ((index * size) + offset) as usize;

// This value is then used in unsafe buffer access:
unsafe {
    buffer
        .subslice(byte_index..)
        .get_value(elem_type, Ordering::Relaxed)
};
```

`index`, `size`, and `offset` are all `u64`. On 32-bit targets (`usize` = 32 bits), the result silently wraps, producing wrong byte offsets that go into `unsafe` code.

This is part of a broader issue — `boa_engine` globally suppresses `clippy::cast_possible_truncation` and related lints in `lib.rs`:
```rust
// It may be worth to look if we can fix the issues highlighted by these lints.
clippy::cast_possible_truncation,
clippy::cast_sign_loss,
clippy::cast_precision_loss,
clippy::cast_possible_wrap,
```

## Affected code

- `core/engine/src/builtins/typed_array/object.rs` — 2 casts (both in `unsafe` paths)
- `core/engine/src/builtins/typed_array/builtin.rs` — 23 casts (buffer indexing, `copyWithin`, `set`, `slice`, `from`)

## Proposed fix

Replace `u64 as usize` with `usize::try_from()`, returning a `RangeError` when the value exceeds the platform's address space:

```rust
// Before:
let byte_index = ((index * size) + offset) as usize;

// After:
let byte_index = usize::try_from((index * size) + offset).map_err(|_| {
    JsNativeError::range().with_message("byte index exceeds addressable range")
})?;
```

For casts that are provably safe (e.g., `element_size` is always 1–8), add a local `#[allow]` with a safety comment instead.

I'd like to work on this. This would be the first step toward eventually re-enabling the cast safety lints crate-wide.

hardik-xi11:
This Pull Request fixes/closes #4646

It changes the following:

- Replaced u64 as usize casts with usize::try_from() in the typed_array module
- Updated `typed_array_get_element` to return `JsResult<Option<JsValue>>` 
- Added `byte_index_to_usize()` helper method

Contribution checklist
[x] Work done in a feature branch, not main.
[x] Single, focused change (typed_array 32-bit safety).
[x] Commit message references the issue and uses Fixes #4646.
[x] Pushed to your fork; PR will be from fork → boa-dev/boa.

### DIFF
--- a/core/engine/src/builtins/typed_array/builtin.rs
+++ b/core/engine/src/builtins/typed_array/builtin.rs
@@ -29,6 +29,16 @@ use crate::{
 };
 use crate::{builtins::array_buffer::utils::memmove_naive, value::JsVariant};
 
+/// Converts a byte index or length from `u64` to `usize`, returning a `RangeError` on 32-bit
+/// when the value exceeds the platform's addressable range (e.g. for buffer indexing).
+fn byte_index_to_usize(v: u64) -> JsResult<usize> {
+    usize::try_from(v).map_err(|_| {
+        JsNativeError::range()
+            .with_message("byte index exceeds addressable range")
+            .into()
+    })
+}
+
 /// The JavaScript `%TypedArray%` object.
 ///
 /// <https://tc39.es/ecma262/#sec-%typedarray%-intrinsic-object>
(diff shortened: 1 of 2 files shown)


jedel1043:
Please don't work on already claimed issues.

PR #4652, opened 2026-02-21
Implement monotonic clock

### LINKED ISSUE #4330: Implementation of `StdClock` is not monotonic.
According to [`JsInstant`'s documentation](https://github.com/boa-dev/boa/blob/e51fcd7f9637788c5e9e16f1e667f1fdfebef306/core/engine/src/context/time.rs#L1-L16):

> This type is guaranteed to be monotonic, i.e. if two instants are compared, the later one will always be greater than the earlier one. It is also always guaranteed to be greater than
or equal to the Unix epoch.

However, [`StdClock`'s implementation](https://github.com/boa-dev/boa/blob/e51fcd7f9637788c5e9e16f1e667f1fdfebef306/core/engine/src/context/time.rs#L141-L150) uses `SystemTime`, which on its [documentation](https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html) specifies:

> Distinct from the [Instant](https://doc.rust-lang.org/stable/std/time/struct.Instant.html) type, this time measurement is not monotonic. This means that you can save a file to the file system, then save another file to the file system, and the second file has a SystemTime measurement earlier than the first. In other words, an operation that happens after another operation in real time may have an earlier SystemTime!

This makes it possible to break the invariants of `JsInstant`.

abhinavs1920:
This Pull Request fixes/closes #4330.

### Summary

This PR fixes the implementation of StdClock so that it matches the guarantees documented for JsInstant:

- JsInstant is monotonic (time never goes backwards).
- JsInstant is always greater than or equal to the Unix epoch.

Previously, StdClock::now used std::time::SystemTime::now(), which is not monotonic and can move backwards if the system clock changes.

### Implementation details

- StdClock now uses a `std::time::Instant` connected to a single `SystemTime` snapshot at startup.
- On first use, we store:
  - epoch_offset = SystemTime::now() - UNIX_EPOCH
  - instant = Instant::now()
- On each now() call we compute:
  - `duration = epoch_offset.saturating_add(instant.elapsed())`
  - and wrap that in JsInstant.

This keeps JsInstant monotonic while still representing a duration since the Unix epoch.

I also added a small unit test (stdclock_is_monotonic) to assert that consecutive StdClock::now() calls never go backwards.

### DIFF
--- a/core/engine/src/context/time.rs
+++ b/core/engine/src/context/time.rs
@@ -135,18 +135,48 @@ pub trait Clock {
     fn now(&self) -> JsInstant;
 }
 
-/// A clock that uses the standard system clock.
+/// A clock that uses a monotonic source anchored to the system clock.
 #[derive(Debug, Clone, Copy, Default)]
 pub struct StdClock;
 
-impl Clock for StdClock {
-    fn now(&self) -> JsInstant {
-        let now = std::time::SystemTime::now();
-        let duration = now
-            .duration_since(std::time::UNIX_EPOCH)
-            .expect("System clock is before Unix epoch");
+impl StdClock {
+    /// Returns a `JsInstant` for the current moment, using a monotonic `Instant` anchored
+    /// to the system clock at startup. This preserves monotonicity even if the system
+    /// clock jumps backwards, while still reporting a time since the Unix epoch.
+    fn now_internal() -> JsInstant {
+        use std::sync::OnceLock;
+
+        #[derive(Debug, Clone, Copy)]
+        struct Anchor {
+            epoch_offset: std::time::Duration,
+            instant: std::time::Instant,
+        }
+
+        static ANCHOR: OnceLock<Anchor> = OnceLock::new();
+
+        let anchor = ANCHOR.get_or_init(|| {
+            let system_now = std::time::SystemTime::now();
+            let epoch_offset = system_now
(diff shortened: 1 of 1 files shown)


jedel1043 (commented):
IMO this requires an API redesign. The purpose of `JsInstant` was to have a nicer way to handle an engine clock, but it seems like combining the concept of a monotonic clock and a Unix-timestamp clock is making this a bit of a mess.

I think just having `JsInstant` be a strictly monotonic timestamp that doesn't reflect the system time would make this much simpler. @hansl anything we would want to consider for this to be feasible?

abhinavs1920:
@jedel1043 
That makes sense, I agree that mixing a monotonic clock with Unix time in JsInstant is what’s making this complicated.

If JsInstant were strictly monotonic and engine-internal, the implementation would be much cleaner and wouldn’t need to reflect system time at all. Wall-clock / epoch time could then live separately where needed (e.g. Date).

Before going further, are there any current uses or spec constraints that rely on JsInstant being tied to system/epoch time? Happy to adjust the PR or help with a redesign if that’s the preferred direction.

jedel1043:
I would suggest doing a redesign instead of merging this PR. We can discuss the shape of the new `JsInstant` on the comments of #4330.

abhinavs1920:
Yes, understood your concern thanks for pushing me to the right direction.

PR #4653, opened 2026-02-21
fix: replace unsafe u64 as usize casts in typed_array for 32-bit safety

### LINKED ISSUE #4646: Fix unsafe `as` casts in `typed_array`  replace with `try_from` for 32-bit safety

Deepak-negi11:
## Problem

`typed_array` module has unsafe `u64 as usize` casts that 
silently truncate on 32-bit platforms, feeding directly 
into unsafe buffer operations.

## Changes

- Add `to_usize()` helper in `mod.rs` that wraps 
  `usize::try_from` and returns `RangeError` on overflow
- Replace unsafe casts in `object.rs` buffer paths
- Replace unsafe casts in `builtin.rs` buffer indexing

## Testing

Existing tests pass. Safe on both 32-bit and 64-bit platforms.

Closes #4646

### DIFF
--- a/core/engine/src/builtins/typed_array/builtin.rs
+++ b/core/engine/src/builtins/typed_array/builtin.rs
@@ -2767,7 +2767,7 @@ impl BuiltinTypedArray {
         )?;
 
         // 10. Return O.
-        Ok(TypedArray::new(
+        TypedArray::new(
             // 6. Set O.[[ViewedArrayBuffer]] to data.
             BufferObject::Buffer(data),
             T::ERASED,
@@ -2777,7 +2777,7 @@ impl BuiltinTypedArray {
             Some(byte_length),
             // 9. Set O.[[ArrayLength]] to length.
             Some(length),
-        ))
+        )
     }
 
     /// <https://tc39.es/ecma262/#sec-initializetypedarrayfromlist>
@@ -2998,7 +2998,7 @@ impl BuiltinTypedArray {
                 0,
                 Some(byte_length),
                 Some(element_length),
-            ),
+            )?,
         )
         .upcast();
 
@@ -3112,7 +3112,7 @@ impl BuiltinTypedArray {
         Ok(JsObject::from_proto_and_data_with_shared_shape(
             context.root_shape(),
             proto,
-            TypedArray::new(buffer, T::ERASED, offset, byte_length, array_length),
+            TypedArray::new(buffer, T::ERASED, offset, byte_length, array_length)?,
         )
         .upcast())
     }
--- a/core/engine/src/builtins/typed_array/object.rs
+++ b/core/engine/src/builtins/typed_array/object.rs
@@ -55,20 +55,70 @@ impl JsData for TypedArray {
 }
 
 impl TypedArray {
-    pub(crate) const fn new(
+    pub(crate) fn new(
         viewed_array_buffer: BufferObject,
         kind: TypedArrayKind,
         byte_offset: u64,
         byte_length: Option<u64>,
         array_length: Option<u64>,
-    ) -> Self {
-        Self {
+    ) -> JsResult<Self> {
+        Self::validate_platform_limits(kind, byte_offset, byte_length, array_length)?;
+
+        Ok(Self {
             viewed_array_buffer,
             kind,
             byte_offset,
             byte_length,
             array_length,
+        })
+    }
+
+    fn validate_platform_limits(
+        kind: TypedArrayKind,
+        byte_offset: u64,
+        byte_length: Option<u64>,
+        array_length: Option<u64>,
+    ) -> JsResult<()> {
+        let ensure_usize = |value, name| {
+            usize::try_from(value).map_err(|_| {
+                JsNativeError::range()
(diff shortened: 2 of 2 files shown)


Deepak-negi11:
@jedel1043 review this pr and tell me if their is any issue

jedel1043:
I wonder if a better approach would be to check when constructing the `TypedArray` if the index fits within a `usize`. This would avoid having to check the index every time we want to cast to `usize`.

Deepak-negi11:
Thanks for the feedback will try to solve

core/engine/src/builtins/typed_array/builtin.rs
```diff
@@ -559,16 +560,16 @@ impl BuiltinTypedArray {
...
-            let to_byte_index = (to * element_size + byte_offset) as usize;
+            let to_byte_index = byte_index_to_usize(to * element_size + byte_offset)?;
 
             // j. Let fromByteIndex be (startIndex × elementSize) + byteOffset.
-            let from_byte_index = (from * element_size + byte_offset) as usize;
+            let from_byte_index = byte_index_to_usize(from * element_size + byte_offset)?;
 
             // k. Let countBytes be count × elementSize.
-            let mut count_bytes = (count * element_size) as usize;
+            let mut count_bytes = byte_index_to_usize(count * element_size)?;
```
zhuzhu81998, line 573: while we are here, does the spec define what should happen if these numeric operations (*, + etc.) overflow?
afaik rust release mode will silently wrap overflows, which will result in a "non-sense operation" before the cast to `usize` happens?

EDIT: I dont think it could actually overflow in this exact case, but maybe there is somewhere a case where it might occur?

zhuzhu81998: @jedel1043 in case there is no notifications.

Sharktheone:
I don't think this can ever be an issue. On 32-bit platforms, you can't ever have enough memory to be able to overflow the `as` cast. (not even with 32-bit memory extensions). Even when having no check when creating the TA, the OS will just reject your allocation, making it actually impossible to hit that case.

zhuzhu81998:
> I don't think this can ever be an issue. On 32-bit platforms, you can't ever have enough memory to be able to overflow the as cast. (not even with 32-bit memory extensions). Even when having no check when creating the TA, the OS will just reject your allocation, making it actually impossible to hit that case.

Even if the truncation when converting from `u64` to `usize` never actually happens here, i would say it is still good practice to avoid these `as` . There is no point in spending time thinking whether there is an edge case where it could overflow, if a solution is straightforward.

zhuzhu81998:
I would say we fix #4766 first, because after that they should all be `usize` even without casting. And then we see, if there are still any potentially unsafe stuffs left.

core/engine/src/builtins/typed_array/builtin.rs
jedel1043: It's... complicated.

According to the [spec](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-mathematical-operations), all of these should be mathematical values with infinite range. However, we cannot use BigInts for performance reasons, and it wouldn't even make sense to use that if realistically it's impossible to have an array with 2^64-1 elements in it.

jedel1043:
Closing this since #4787 seems like a better approach overall.

PR #4654, opened 2026-02-21
fix: skip npm publish if boa_wasm version already exists

### LINKED ISSUE #4065: Don't publish npm wasmt package if version exists
We sometimes need to re-publish the same release which means publishing the same version. Cargo will ignore publishing a crate if it already exists but npm still fails.

It would be nice to have npm not fail if the version was already uploaded to the registry

our workflow lives here:
https://github.com/boa-dev/boa/blob/main/.github/workflows/release.yml#L39-L70

I guess we would want a step that runs `npm view @boa-dev/boa_wasm version` then checks the value against the current version (from Cargo.toml) if they match exit the workflow without any errors

jkrishnad:
This Pull Request fixes/closes #4065 .

It changes the following:

- Added npm version check using `npm view @boa-dev/boa_wasm@$VERSION`
- Skip `npm publish` step when the version already exists

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,7 +79,7 @@ jobs:
 
           if npm view @boa-dev/boa_wasm@$VERSION version > dev/null 2>&1; then
             echo "Version $VERSION already published. Skipping"
-            echo "SKIP_PUBLISH=true" >> $GUTHUB_ENV
+            echo "SKIP_PUBLISH=true" >> $GITHUB_ENV
           fi
 
       - name: Publish to npm
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,7 +75,7 @@ jobs:
 
       - name: Check if the npm version already exists
         run: |
-          VERSION=$(node -p "require('./ffi/wasm/pkg/package.json').version")
+          VERSION=$(jq -r '.version' ./ffi/wasm/pkg/package.json)
 
           if npm view @boa-dev/boa_wasm@$VERSION version > dev/null 2>&1; then
             echo "Version $VERSION already published. Skipping"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -73,7 +73,17 @@ jobs:
       - name: Set-up npm config for publishing
         run: npm config set -- '//registry.npmjs.org/:_authToken' "${{ secrets.NPM_TOKEN }}"
 
+      - name: Check if the npm version already exists
+        run: |
+          VERSION=$(node -p "require('./ffi/wasm/pkg/package.json').version")
+
+          if npm view @boa-dev/boa_wasm@$VERSION version > dev/null 2>&1; then
+            echo "Version $VERSION already published. Skipping"
+            echo "SKIP_PUBLISH=true" >> $GUTHUB_ENV
```
jedel1043:
```suggestion
            echo "SKIP_PUBLISH=true" >> $GITHUB_ENV
```
typo

jkrishnad: sry ser my bad i'll fix now

.github/workflows/release.yml
```diff
@@ -73,7 +73,17 @@ jobs:
       - name: Set-up npm config for publishing
         run: npm config set -- '//registry.npmjs.org/:_authToken' "${{ secrets.NPM_TOKEN }}"
 
+      - name: Check if the npm version already exists
+        run: |
+          VERSION=$(node -p "require('./ffi/wasm/pkg/package.json').version")
```
jedel1043: Use `jq` here. It should be simpler to understand than having to call node just to access a field.

jedel1043 (approved):
Nice, thank you for the fix!

PR #4660, opened 2026-02-21
fix: use dynamic UTC expectations in date_proto_set_utc_hours test for timezone portability

### LINKED ISSUE #4678: date_proto_set_utc_hours  test fails on non-UTC timezones
## Bug

Some tests in `builtins::date::tests` use the local-time Date constructor
(`new Date(2020, 6, 8, 9, 16, 15, 779)`) but compare UTC components against
hardcoded values that only work in UTC.

For example, [date_proto_set_utc_hours](cci:1://file:///c:/Users/dayan/Projects/boa/core/engine/src/builtins/date/tests.rs:678:0-733:1) expects UTC minutes to be `16`,
but on IST (UTC+5:30), local 9:16 → UTC 3:46, so the minutes are `46`.

## Affected tests
- [date_proto_set_utc_hours](cci:1://file:///c:/Users/dayan/Projects/boa/core/engine/src/builtins/date/tests.rs:678:0-733:1)

## Expected behavior
Tests should pass regardless of the host machine's timezone.

## Fix
Compute expected UTC values dynamically using Rust's [time](cci:1://file:///c:/Users/dayan/Projects/boa/core/engine/src/builtins/date/tests.rs:61:0-76:1) crate
([from_local().to_offset(UTC)](cci:1://file:///c:/Users/dayan/Projects/boa/core/engine/src/builtins/date/tests.rs:30:0-46:1)) instead of hardcoding them.

- Rustc version: [e.g. rustc 1.43.0 (4fb7144ed 2020-04-20), running `rustc -V`]

**Additional context**

Deepak-negi11:
Fixes #4678

## Problem
Some tests in `builtins::date::tests` were constructing dates using the 
local-time `Date` constructor:
```js
new Date(2020, 6, 8, 9, 16, 15, 779)
```

This interprets arguments as **local time**, so the resulting UTC components 
vary depending on the timezone of the machine running the tests.

For example, in IST (UTC+5:30), local `9:16` becomes UTC `3:46` — the minutes 
shift by 30. This caused `date_proto_set_utc_hours` to fail because the test 
expected UTC minutes to be `16`, but they were actually `46`.

## Fix
Replaced the local-time constructor with `Date.UTC()`:
```js
new Date(Date.UTC(2020, 6, 8, 9, 16, 15, 779))
```

`Date.UTC()` always interprets its arguments as UTC directly, so the UTC 
components are consistent regardless of the machine's timezone.

## Testing
- ✅ `date_proto_set_month`
- ✅ `date_proto_set_utc_hours`

Tested on IST (UTC+5:30) where the bug was originally discovered.

### DIFF
--- a/core/engine/src/builtins/date/tests.rs
+++ b/core/engine/src/builtins/date/tests.rs
@@ -678,27 +678,53 @@ fn date_proto_set_utc_full_year() {
 
 #[test]
 fn date_proto_set_utc_hours() {
+    // The local-time constructor `new Date(2020, 6, 8, 9, 16, 15, 779)` converts
+    // local time to UTC internally. The resulting UTC components depend on the
+    // host timezone offset, so we compute them dynamically.
+    let base = from_local(2020, 7, 8, 9, 16, 15, 779).to_offset(time::UtcOffset::UTC);
+    let base_y = base.year();
+    let base_m = base.month() as u8;
+    let base_d = base.day();
+    let base_min = base.minute();
+    let base_sec = base.second();
+    let base_ms = base.millisecond();
+
     run_test_actions([
+        // setUTCHours(11): only the UTC hour changes; date/min/sec/ms stay.
         TestAction::assert_eq(
             "new Date(2020, 6, 8, 9, 16, 15, 779).setUTCHours(11)",
-            timestamp_from_utc(2020, 7, 8, 11, 16, 15, 779),
+            timestamp_from_utc(base_y, base_m, base_d, 11, base_min, base_sec, base_ms),
         ),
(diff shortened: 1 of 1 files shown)


jedel1043:
I don't think this is the correct fix; we cannot just leave local time resolution untested. A correct fix would fetch the current datetime externally (maybe with the `time` crate?), then check equality with the date returned by `Date`.

You pushed an unrelated commit btw 😅

jedel1043:
But I guess we could also apply this and have a separate set of tests for local date + setUTC.

Deepak-negi11:
@jedel1043  can you please review it again

jedel1043 (approved):
Looks great!

PR #4661, opened 2026-02-21
fix(intl): canonicalize Unicode extension 'yes' alias in Intl.getCanonicalLocales

Lokesh0224:
This PR fixes canonicalization of the Unicode extension value "yes" for the keys:

- kb
- kc
- kh
- kk
- kn

According to  [ECMA-402 9.2.1](https://tc39.es/ecma402/#sec-canonicalizelocalelist), "yes" is an alias for "true" for these
keys. Since "true" is the default value, it must be omitted in the canonical Unicode extension sequence.

Before:
Screenshot from 2026-02-22 02-02-12

After:
Screenshot from 2026-02-22 03-39-06

Test Impact:

Fixes test262 failure in:
 test/intl402/Intl/getCanonicalLocales/unicode-ext-canonicalize-yes-to-true.js

### DIFF
--- a/core/engine/src/builtins/intl/locale/utils.rs
+++ b/core/engine/src/builtins/intl/locale/utils.rs
@@ -80,6 +80,9 @@ pub(crate) fn locale_from_value(tag: &JsValue, context: &mut Context) -> JsResul
         .locale_canonicalizer()?
         .canonicalize(&mut tag);
 
+    // TODO: Remove this workaround once ICU4X supports canonicalization of
+    // Unicode extension value aliases (unicode-org/icu4x#3483).
+    // Currently ICU4X does not canonicalize "yes" → "true" for certain keys.
     let keys: Vec<_> = tag
         .extensions
         .unicode
--- a/core/engine/src/builtins/intl/locale/utils.rs
+++ b/core/engine/src/builtins/intl/locale/utils.rs
@@ -60,7 +60,6 @@ pub(crate) fn locale_from_value(tag: &JsValue, context: &mut Context) -> JsResul
     // iv. Else,
     // 1. Let tag be ? ToString(kValue).
     let tag = tag.to_string(context)?.to_std_string_escaped();
-
     if tag.contains('_') {
         return Err(JsNativeError::range()
             .with_message("locale is not a structurally valid language tag")
@@ -83,6 +82,7 @@ pub(crate) fn locale_from_value(tag: &JsValue, context: &mut Context) -> JsResul
 
     // TODO: Remove this workaround once ICU4X supports canonicalization of
     // Unicode extension value aliases (unicode-org/icu4x#3483).
+    // Currently ICU4X does not canonicalize "yes" → "true" for certain keys.
     let keys: Vec<_> = tag
         .extensions
         .unicode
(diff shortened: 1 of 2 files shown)


jedel1043:
Thanks!

We might have to track https://github.com/unicode-org/icu4x/issues/3483 then, to remove this code when ICU4X supports extension canonicalization.

Can you add a TODO linking to that issue?

Lokesh0224:
Done!

Lokesh0224:
@jedel1043 help!!, I'm also seeing failures in transformed-ext-canonical.js — is the m0 alias something we should handle similarly?

jedel1043:
If we need to add all aliases into our code, we would be playing cat and mouse with every new CLDR release 😅

I would suggest only patching up `yes` since it is the most common, kind of as a reminder to keep track of the ICU4X issue. All other extension canonicalization should eventually be fixed on ICU4X's side.

jedel1043:
What would be kind of useful is to collect a full list of all the tests that require extension canonicalization, then ignore them in [`test262_config.toml`](https://github.com/boa-dev/boa/blob/main/test262_config.toml) with a link to the issue.

Lokesh0224:
Done! I've reverted the extra alias commits and added the 6 failing extension canonicalization tests to test262_config.toml with a link to the ICU4X issue.

core/engine/src/builtins/intl/locale/utils.rs
```diff
@@ -79,6 +80,29 @@ pub(crate) fn locale_from_value(tag: &JsValue, context: &mut Context) -> JsResul
...
+        .unicode
+        .keywords
+        .iter()
+        .map(|(k, _)| k.clone())
+        .collect();
+
+    for k in keys {
+        if let Some(v) = tag.extensions.unicode.keywords.get_mut(&k) {
+            if v.to_string() == "yes" {
+                match k.as_str() {
```
jedel1043: Extract the match out of the ifs. We don't really need to do anything if the key doesn't support this canonicalization.

core/engine/src/builtins/intl/locale/utils.rs
```diff
@@ -79,6 +80,29 @@ pub(crate) fn locale_from_value(tag: &JsValue, context: &mut Context) -> JsResul
...
+        .extensions
+        .unicode
+        .keywords
+        .iter()
+        .map(|(k, _)| k.clone())
+        .collect();
+
+    for k in keys {
+        if let Some(v) = tag.extensions.unicode.keywords.get_mut(&k) {
+            if v.to_string() == "yes" {
```
jedel1043: You can combine the two ifs into a single one with `if let Some(v) = _ && v == something`.

Also maybe we can avoid creating a string? IIRC you can statically create single values, then compare directly: `v == value!("yes")`

jedel1043 (changes requested):
Nice work! I have a couple nitpicks, but everything else looks really nice.

core/engine/src/builtins/intl/locale/utils.rs
Lokesh0224: Done! Used value!("yes") for direct comparison to avoid string allocation. Thanks for the suggestion.

jedel1043 (approved):
Thank you!

core/engine/src/builtins/intl/locale/utils.rs
```diff
@@ -79,6 +80,29 @@ pub(crate) fn locale_from_value(tag: &JsValue, context: &mut Context) -> JsResul
...
+    // TODO: Remove this workaround once ICU4X supports canonicalization of
+    // Unicode extension value aliases (unicode-org/icu4x#3483).
+    // Currently ICU4X does not canonicalize "yes" → "true" for certain keys.
+    let keys: Vec<_> = tag
+        .extensions
+        .unicode
+        .keywords
+        .iter()
+        .map(|(k, _)| *k)
+        .collect();
```
jedel1043, line 92: Wait, do we require doing this? It seems like we could just iterate using `tag.extensions.unicode.keywords.iter_mut()`, right? which avoids collecting

jedel1043: Oh, checked the API for `Keywords` and it doesn't expose a way to mutably iterate. Weird

This is fine then.

PR #4664, opened 2026-02-22
fix(string): ensure separator is an object before calling Symbol.split

### LINKED ISSUE #4663: bug: String.prototype.split should not access Symbol.split on primitive separators
**Describe the bug**
I found a bug in how String.prototype.split works in Boa.
According to the ECMAScript spec, when we use split(separator), the engine should only check for Symbol.split if the separator is an Object
Right now, Boa tries to access Symbol.split even if the separator is a primitive (like 1n, true, or 100). This is wrong because if someone adds a custom function to BigInt.prototype[Symbol.split], Boa will call it, but the spec says it should ignore it for primitives.

**To Reproduce**
You can also see this by running the official Test262 test:
```
cargo run --bin boa_tester -- run -s "test/built-ins/String/prototype/split/cstm-split-on-bigint-primitive.js"
```
**Expected behavior**

Boa should check if the separator is an Object. Since 1n is a primitive, it should skip the Symbol.split check, convert 1n to the string "1", and return ["a", "b"]

**Actual behavior**
 tries to access Symbol.split on the primitive, triggers the error, and the test fails.

[spec]: 
According to [Section 22.1.3.21, step 2](https://tc39.es/ecma262/#sec-string.prototype.split), the engine should only look for @@split if the separator is an Object.

**Test Evidence:**
**Full Suite Failure:** As seen in the screenshot below, there are 4 failing tests in the split directory, resulting in 96.67% conformance.

Image

**Specific Test Failure:** Specifically, the test cstm-split-on-bigint-primitive.js fails with an F as shown here:
Image

amrkhaled104:
This Pull Request fixes #4663

**Changes:**
- Added a check to ensure the separator is an Object before looking for the @@split method.
- This prevents the engine from calling custom split functions added to primitive prototypes (like BigInt.prototype).
- Updated the logic in `core/engine/src/builtins/string/mod.rs`

**Validation:**
- I ran the official Test262 suite for String.prototype.split.
- Before fix: 116/120 tests passed (4 failed).
- After fix: 120/120 tests passed (100% conformance).

**Test Results Screenshot:**
image
I have expanded the fix to include other string methods that were also incorrectly accessing Symbol methods on primitive separators.

### DIFF
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -1915,12 +1915,14 @@ impl String {
 
         // 2. If separator is neither undefined nor null, then
         if !separator.is_null_or_undefined() {
-            // a. Let splitter be ? GetMethod(separator, @@split).
-            let splitter = separator.get_method(JsSymbol::split(), context)?;
-            // b. If splitter is not undefined, then
-            if let Some(splitter) = splitter {
-                // i. Return ? Call(splitter, separator, « O, limit »).
-                return splitter.call(separator, &[this.clone(), limit.clone()], context);
+            if let Some(separator_obj) = separator.as_object() {
+                // a. Let splitter be ? GetMethod(separator, @@split).
+                let splitter = separator_obj.get_method(JsSymbol::split(), context)?;
+                // b. If splitter is not undefined, then
+                if let Some(splitter) = splitter {
+                    // i. Return ? Call(splitter, separator, « O, limit »).
+                    return splitter.call(separator, &[this.clone(), limit.clone()], context);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/string/mod.rs
```diff
@@ -1915,12 +1916,14 @@ impl String {
 
         // 2. If separator is neither undefined nor null, then
         if !separator.is_null_or_undefined() {
-            // a. Let splitter be ? GetMethod(separator, @@split).
-            let splitter = separator.get_method(JsSymbol::split(), context)?;
-            // b. If splitter is not undefined, then
-            if let Some(splitter) = splitter {
-                // i. Return ? Call(splitter, separator, « O, limit »).
-                return splitter.call(separator, &[this.clone(), limit.clone()], context);
+            if let Some(separator_obj) = separator.as_object() {
```
jedel1043, line 1919: Kind of unrelated, but now that you're touching this part of the code, can you change this to use `let` chains and `&&`s?
Something like `if !separator.blah() && let Some(separator_obj) = blah`

Also kind of applies to the rest of the added lines since I see that a lot of them are just two nested `if`s.

jedel1043:
Already fixed in the main branch

PR #4665, opened 2026-02-22
Fix String custom matcher delegation per ECMAScript §22.1.3

mrhapile:
This Pull Request fixes/closes #{none}.

It changes the following:

- Updates `String.prototype.{match, matchAll, replace, replaceAll, search, split}` to invoke `GetMethod(@@symbol)` only when the argument is an **Object**, per ECMAScript §22.1.3.
- Replaces `!value.is_null_or_undefined()` guards with `value.is_object()` to prevent incorrect primitive delegation.
- Aligns implementation comments with current ECMAScript specification wording.

### Before:
- built-ins/String: 28 failures
- Full suite: 1131 failures

### After:
- built-ins/String: 5 failures
- Full suite: 1108 failures
- 0 panics

## Specification Alignment

According to ECMAScript §22.1.3, custom matcher delegation must only occur if the argument is an Object:

> If the argument is an Object, then let method be ? GetMethod(argument, @@symbol).

Previously, delegation was performed whenever the argument was not `null` or `undefined`, which allowed primitives to be implicitly boxed and incorrectly trigger prototype-based `Symbol.*` method delegation.

This PR corrects that behavior.

## Test262 Impact

- built-ins/String failures reduced from **28 → 5** (-23)
- Full suite failures reduced from **1131 → 1108** (-23)
- 0 panics
- No regressions observed in Array, Promise, or Object suites

The remaining 5 failures are related to RegExp `v/u` flag support and are unrelated to this change.

## Risk Assessment

- Change is isolated to `core/engine/src/builtins/string/mod.rs`
- No VM, GC, or runtime structural changes
- No observable regressions outside String builtins
- Strictly improves spec compliance

### DIFF
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -1003,8 +1003,8 @@ impl String {
         let search_value = args.get_or_undefined(0);
         let replace_value = args.get_or_undefined(1);
 
-        // 2. If searchValue is neither undefined nor null, then
-        if !search_value.is_null_or_undefined() {
+        // 2. If searchValue is an Object, then
+        if search_value.is_object() {
             // a. Let replacer be ? GetMethod(searchValue, @@replace).
             let replacer = search_value.get_method(JsSymbol::replace(), context)?;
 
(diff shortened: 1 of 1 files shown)


jedel1043 (approved):
Oh, I think this was correct before, but the spec just got updated to a more strict check. Thank you for the fix!

PR #4667, opened 2026-02-22
Fix RunFile to execute test files only once

### LINKED ISSUE #4666: bug: `RunFile` executes the same file twice in runtime test harness
## Summary
When using `TestAction::RunFile` in `core/runtime`, the file gets executed twice. This happens because `forward_file` is called twice in the `RunFile` branch.

## Steps to reproduce
1. Create a JS file with a visible side effect:
   ```js
   // test.js
   globalThis.counter = (globalThis.counter ?? 0) + 1;
   ```
2. Run the test harness with `TestAction::RunFile { path: "test.js".into() }`.
3. Observe that `counter` becomes `2`, even though the file was intended to run once.

## Expected behavior
`RunFile` should evaluate the file exactly once.

## Actual behavior
The file is evaluated twice, so any side effects run twice.

## Impact
File-based tests become flaky or incorrect. Any setup or state mutation in the file will be doubled.

## Location in code
`core/runtime/src/lib.rs`, `Inner::RunFile` branch:
```rust
if let Err(e) = forward_file(context, &path) {
    panic!("Uncaught {e} in file {path:?}");
}
forward_file(context, &path).expect("failed to run file");
```

## Suggested fix
Remove the second call and keep a single evaluation.

Nakshatra480:
Closes #4666 

## Summary
This removes the duplicate `forward_file` call in the runtime test harness so `TestAction::RunFile` runs a file exactly once.

## Changes
- Remove the second `forward_file` invocation in the `RunFile` branch.

## Tests
- cargo fmt -- --check

### DIFF
--- a/core/runtime/src/lib.rs
+++ b/core/runtime/src/lib.rs
@@ -309,7 +309,6 @@ pub(crate) mod test {
                     if let Err(e) = forward_file(context, &path) {
                         panic!("Uncaught {e} in file {path:?}");
                     }
-                    forward_file(context, &path).expect("failed to run file");
                 }
                 Inner::RunJobs => {
                     if let Err(e) = context.run_jobs() {


jedel1043 (approved):
Thank you!

PR #4669, opened 2026-02-22
ci: deny warnings and run miri in CI

### LINKED ISSUE #2366: Deny warnings and run miri in the CI
As commented by @Razican in https://github.com/boa-dev/ryu-js/pull/21#discussion_r1001881114 and https://github.com/boa-dev/ryu-js/pull/21#discussion_r1001881454, we should deny warnings and run MIRI in our CI in order to catch more bugs on PRs.

Flamki:
This Pull Request fixes/closes #2366.\n\nIt changes the following:\n\n- Keeps warning denial conditional for main (W_FLAGS / RUSTDOCFLAGS), so release branches are not blocked by warning backports.\n- Fixes the ARM Linux target rustflags key to CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUSTFLAGS.\n- Adds a dedicated miri CI job on nightly (cargo miri setup).\n- Runs opt-in Miri tests by name (miri) across the workspace, excluding oa_cli and oa_examples because their binary test harnesses use non-Miri-compatible native FFI paths (jemalloc/curl).

### DIFF
--- a/core/gc/src/test/allocation.rs
+++ b/core/gc/src/test/allocation.rs
@@ -1,7 +1,7 @@
 mod miri {
     use boa_macros::{Finalize, Trace};
 
-    use super::{Harness, run_test};
+    use super::super::{Harness, run_test};
     use crate::{Gc, GcBox, GcRefCell, force_collect};
 
     #[test]
@@ -63,5 +63,4 @@ mod miri {
             Harness::assert_empty_gc();
         });
     }
-
 }
--- a/core/gc/src/test/cell.rs
+++ b/core/gc/src/test/cell.rs
@@ -1,5 +1,5 @@
 mod miri {
-    use super::run_test;
+    use super::super::run_test;
     use crate::{Gc, GcRefCell};
 
     #[test]
@@ -13,5 +13,4 @@ mod miri {
             }
         });
     }
-
 }
--- a/core/gc/src/test/erased.rs
+++ b/core/gc/src/test/erased.rs
@@ -3,7 +3,7 @@ mod miri {
 
     use boa_macros::{Finalize, Trace};
 
-    use super::run_test;
+    use super::super::run_test;
     use crate::{Gc, GcBox, GcErased, force_collect, test::Harness};
 
     #[test]
@@ -109,5 +109,4 @@ mod miri {
             assert_eq!(*Gc::downcast::<Derived>(base).unwrap(), *derived);
         });
     }
-
 }
--- a/core/gc/src/test/weak.rs
+++ b/core/gc/src/test/weak.rs
@@ -1,7 +1,7 @@
 mod miri {
     use std::{cell::Cell, rc::Rc};
 
-    use super::run_test;
+    use super::super::run_test;
     use crate::{
         Ephemeron, Finalize, Gc, GcBox, GcRefCell, Trace, WeakGc, force_collect,
         internals::EphemeronBox, test::Harness,
@@ -333,5 +333,4 @@ mod miri {
             Harness::assert_exact_bytes_allocated(root_size);
         });
     }
-
 }
--- a/core/gc/src/test/weak_map.rs
+++ b/core/gc/src/test/weak_map.rs
@@ -1,5 +1,5 @@
 mod miri {
-    use super::run_test;
+    use super::super::run_test;
     use crate::{Gc, WeakMap, force_collect, has_weak_maps};
 
     #[test]
@@ -137,5 +137,4 @@ mod miri {
             assert!(map.contains_key(&key_copy));
         });
     }
-
 }


.github/workflows/rust.yml
```diff
@@ -66,10 +66,10 @@ jobs:
...
       RUSTUP_WINDOWS_PATH_ADD_BIN: 1
       # TODO: Make a small Rust tool that does this
-      CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_RUSTFLAGS: ${{ (github.ref == 'refs/heads/main' || github.base_ref == 'main') && '-D warnings' || '' }}
-      CARGO_TARGET_AARCH64_APPLE_DARWIN_RUSTFLAGS: ${{ (github.ref == 'refs/heads/main' || github.base_ref == 'main') && '-D warnings' || '' }}
-      CARGO_TARGET_X86_64_APPLE_DARWIN_RUSTFLAGS: ${{ (github.ref == 'refs/heads/main' || github.base_ref == 'main') && '-D warnings' || '' }}
-      CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU: ${{ (github.ref == 'refs/heads/main' || github.base_ref == 'main') && '-D warnings' || '' }}
+      CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_RUSTFLAGS: -D warnings
+      CARGO_TARGET_AARCH64_APPLE_DARWIN_RUSTFLAGS: -D warnings
+      CARGO_TARGET_X86_64_APPLE_DARWIN_RUSTFLAGS: -D warnings
+      CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUSTFLAGS: -D warnings
```
jedel1043: Oh cool, didn't notice this bug in our CI.

.github/workflows/rust.yml
```diff
@@ -107,6 +107,41 @@ jobs:
...
+            target
+            ~/.cargo/git
+            ~/.cargo/registry
+          key: ${{ runner.os }}-${{ runner.arch }}-cargo-miri-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Setup miri
+        run: cargo miri setup
+
+      - name: Run miri tests
+        run: cargo miri test -p boa_gc --lib
```
jedel1043: I think we do want to test other parts of the engine, so I would suggest running all tests that have `miri` on their name, as per the discussion in #2448. That way we can easily "opt-in" to running a test with miri on our CI.

.github/workflows/rust.yml
```diff
@@ -17,7 +17,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  W_FLAGS: ${{ (github.ref == 'refs/heads/main' || github.base_ref == 'main') && '-D warnings' || '' }}
+  W_FLAGS: -D warnings
```
jedel1043: I'd suggest leaving the conditional `-D warnings`. We generally don't want to enable warnings on release branches, because this makes it more difficult to backport bugs if we also need to push additional commits fixing all the warnings.

Flamki:
@jedel1043 Addressed the requested changes in 20774a71:\n\n- Restored conditional warning denial behavior (W_FLAGS and RUSTDOCFLAGS) so warnings are denied on main only.\n- Kept and fixed the ARM Linux env var key as CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUSTFLAGS.\n- Updated the Miri step to run opt-in tests by name: cargo miri test --workspace miri (per the #2448 discussion).\n\nPlease take another look when convenient.

Flamki:
Addressed the failing Miri job in a6af4949.\n\nRoot cause: cargo miri test --workspace miri also executed bin unit-test harnesses (including oa_cli), which uses jemallocator; Miri cannot call jemalloc FFI (_rjem_malloc).\n\nChange:\n- Updated Miri command to only run library + integration-test targets with the opt-in filter: cargo miri test --workspace --lib --tests miri.\n\nThis keeps the miri name-based opt-in behavior from #2448 while avoiding unsupported bin target execution under Miri.

Flamki:
Follow-up fix in b5c72159: Miri still failed because cargo miri test ... miri was running oa_cli's unit-test harness, and that binary uses jemallocator (_rjem_malloc), which Miri cannot execute.\n\nUpdated the command to:\n- cargo miri test --workspace --exclude boa_cli miri\n\nThis preserves the name-based opt-in (miri) across the workspace while skipping the known incompatible CLI target.

Flamki:
Another Miri follow-up in f5d63f61.\n\nAfter excluding oa_cli, Miri still failed on oa_examples (module_fetch_async) due curl_global_init FFI in isahc/curl, which Miri does not support.\n\nUpdated command to:\n- cargo miri test --workspace --exclude boa_cli --exclude boa_examples miri\n\nThis keeps name-based opt-in while excluding the two known non-Miri-compatible binary packages.

jedel1043:
Nice work! Now we can start pushing some tests to run with miri. A good candidate would be tests in `boa_gc`. My suggestion is to wrap the files containing tests with a `mod miri` e.g. a file with tests like
```rust
#[test]
fn test_something(){}

#[test]
fn test_something_else(){}
```

should become
```rust
mod miri {
  #[test]
  fn test_something(){}
  
  #[test]
  fn test_something_else(){}
}
```
you should be able to easily do that for the files in https://github.com/boa-dev/boa/tree/cb47fa5b1e924e029e5b490cd22f99b6f58b7f72/core/gc/src/test

Flamki:
Addressed the latest review feedback in 6913b538.

- Wrapped each test file under core/gc/src/test (llocation.rs, cell.rs, erased.rs, weak.rs, weak_map.rs) in mod miri { ... }.
- Kept core/gc/src/test/mod.rs unchanged since it only contains shared harness helpers.

This preserves the existing tests while making them discoverable by the cargo miri test ... miri name filter in CI.

jedel1043:
Cool! Let's see if this properly works then :)

Flamki:
Follow-up fix pushed in 5be577c3.

Root cause of the CI failures was module path resolution after wrapping GC tests in mod miri:
- inside core/gc/src/test/*, super::run_test and super::{Harness, run_test} resolved to the file module (	est::<file>), not 	est.

Fix:
- Updated imports to parent test module scope:
  - super::super::run_test
  - super::super::{Harness, run_test}
- Removed trailing blank lines before closing braces in those files to satisfy formatting.
- Rebased on latest main.

This should clear the unresolved-import compile failures across Coverage/Lint/MSRV/Miri/Test jobs.

PR #4670, opened 2026-02-22
no redundant allocation in text decoder via `ArrayBuffer::data()`

### LINKED ISSUE #4362: Use `ArrayBuffer::data()` in the text decoder as an optimization
This: https://github.com/boa-dev/boa/blob/main/core/runtime/src/text/mod.rs#L78-L90 creates a copy of the bytes, although that isn't necessary normally.

zhuzhu81998:
This Pull Request fixes #4362 .

Following the suggestion from @HalidOdat in #4358 .

Note that for `Utf16Be`, the `decode` function does in place swapping and thus requires mutability.

### DIFF
--- a/core/runtime/src/text/mod.rs
+++ b/core/runtime/src/text/mod.rs
@@ -134,7 +134,7 @@ impl TextDecoder {
         let strip_bom = !self.ignore_bom;
 
         let Some(data) = array_buffer.data() else {
-            return Err(js_error!(TypeError: "Detached ArrayBuffer"));
+            return Err(js_error!(TypeError: "cannot decode a detached ArrayBuffer"));
         };
 
         Ok(match self.encoding {


zhuzhu81998:
Just saw the work in #4650 , #4637, #4631 that all target basically the same area for different issues. And all of them still use the allocating version.
@jedel1043 it would make sense to at least merge #4631 first so that others can rebase?

core/runtime/src/text/mod.rs
```diff
@@ -70,30 +70,38 @@ impl TextDecoder {
...
-            return Err(
-                js_error!(TypeError: "Argument 1 must be an ArrayBuffer, TypedArray or DataView."),
-            );
+            return Err(js_error!(
+                TypeError: "Argument 1 must be an ArrayBuffer, TypedArray or DataView."
+            ));
+        };
+
+        let Some(data) = array_buffer.data() else {
+            return Err(js_error!(TypeError: "Detached ArrayBuffer"));
```
jedel1043:
```suggestion
            return Err(js_error!(TypeError: "cannot decode a detached ArrayBuffer"));
```

jedel1043 (approved):
Thank you for the contribution! Everything looks good, I just had a suggestion for the detached error message.

jedel1043:
I'll merge this first. #4631 needs another review pass.
EDIT: scratch this, seems like that PR got fixed sooner than I expected 😅

PR #4671, opened 2026-02-22
vm: add polymorphic inline cache for property access

### LINKED ISSUE #4656: Proposal: Polymorphic Inline Cache (PIC) for Property Access
## Background

Currently Boa uses a **monomorphic inline cache** for property access sites, storing only a single `(WeakShape, Slot)` pair. When a shape mismatch occurs, the cache is reset, discarding the previous entry.

This causes constant cache invalidation when the same property is accessed on objects with different shapes, which is very common in JavaScript. As a result, many polymorphic access sites fall back to the slow path repeatedly.

## Example

```javascript
function getX(o) { return o.x; }

getX({x:1});
getX({x:2,y:3});
getX({x:5,z:7});

```
Each object has a different shape, so the current inline cache thrashes and provides little benefit.

### Proposed Improvement

One possible improvement would be to upgrade the monomorphic inline cache to a polymorphic inline cache (PIC) that stores multiple `(Shape → Slot)` mappings per access site. Many modern engines (e.g., V8, SpiderMonkey) use small PICs (commonly 4 entries) to efficiently handle polymorphic property access.

### High-level idea:

- Cache up to N shape → slot mappings (e.g., N = 4)
- On lookup, check cached entries for a matching shape
- If the number of observed shapes exceeds N, transition to a megamorphic state and fall back to the slow path

### Expected Benefit

This could reduce inline cache thrashing and improve performance for polymorphic property accesses.

If this approach seems reasonable, I would be happy to work on an implementation.

Flamki:
This Pull Request fixes/closes #4656.

It changes the following:

- Replaces the monomorphic property inline cache with a 4-entry polymorphic inline cache (PIC) in `core/engine/src/vm/inline_cache/mod.rs`.
- Adds megamorphic fallback when more than 4 distinct shapes are observed at the same access site.
- Updates property/name opcode fast paths to use PIC matching (`GetPropertyByName`, `SetPropertyByName`, `GetNameGlobal`).
- Adds VM inline-cache tests for polymorphic caching and megamorphic transition behavior.

### DIFF
--- a/core/engine/src/vm/inline_cache/mod.rs
+++ b/core/engine/src/vm/inline_cache/mod.rs
@@ -13,76 +13,42 @@ mod tests;
 
 const PIC_CAPACITY: usize = 4;
 
+#[derive(Clone, Debug, Trace, Finalize)]
+struct PicEntry {
+    shape: WeakShape,
+
+    #[unsafe_ignore_trace]
+    slot: Slot,
+}
+
 /// An inline cache entry for a property access.
 #[derive(Clone, Debug, Trace, Finalize)]
 pub(crate) struct InlineCache {
     /// The property that is accessed.
     pub(crate) name: JsString,
 
-    /// A pointer is kept to the shape to avoid the shape from being deallocated.
-    pub(crate) shape: GcRefCell<WeakShape>,
-
-    /// The [`Slot`] of the property.
-    #[unsafe_ignore_trace]
-    pub(crate) slot: Cell<Slot>,
-
-    /// Additional PIC entries beyond the primary `shape`/`slot` entry.
-    pub(crate) secondary_shapes: GcRefCell<[WeakShape; PIC_CAPACITY - 1]>,
-
-    /// [`Slot`] values for additional PIC entries.
-    #[unsafe_ignore_trace]
-    pub(crate) secondary_slots: [Cell<Slot>; PIC_CAPACITY - 1],
-
-    /// Number of active entries in `secondary_shapes` / `secondary_slots`.
-    #[unsafe_ignore_trace]
-    pub(crate) secondary_len: Cell<u8>,
+    /// Cached `(shape, slot)` entries for this access site.
+    ///
+    /// NOTE: This should never exceed `PIC_CAPACITY`.
+    entries: GcRefCell<Vec<PicEntry>>,
 
     /// A site is megamorphic when we observe more distinct shapes than the PIC capacity.
     #[unsafe_ignore_trace]
-    pub(crate) megamorphic: Cell<bool>,
+    megamorphic: Cell<bool>,
 }
 
 impl InlineCache {
     pub(crate) fn new(name: JsString) -> Self {
         Self {
             name,
-            shape: GcRefCell::new(WeakShape::None),
-            slot: Cell::new(Slot::new()),
-            secondary_shapes: GcRefCell::new(std::array::from_fn(|_| WeakShape::None)),
-            secondary_slots: std::array::from_fn(|_| Cell::new(Slot::new())),
-            secondary_len: Cell::new(0),
+            entries: GcRefCell::new(Vec::with_capacity(PIC_CAPACITY)),
             megamorphic: Cell::new(false),
         }
     }
 
-    fn remove_secondary_entry(
-        &self,
-        index: usize,
-        secondary_len: usize,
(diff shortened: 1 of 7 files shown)


core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -23,39 +25,200 @@ pub(crate) struct InlineCache {
...
+    #[unsafe_ignore_trace]
+    pub(crate) secondary_slots: [Cell<Slot>; PIC_CAPACITY - 1],
+
+    /// Number of active entries in `secondary_shapes` / `secondary_slots`.
+    #[unsafe_ignore_trace]
+    pub(crate) secondary_len: Cell<u8>,
+
+    /// A site is megamorphic when we observe more distinct shapes than the PIC capacity.
+    #[unsafe_ignore_trace]
+    pub(crate) megamorphic: Cell<bool>,
```
jedel1043: Hmmm, this kind of gets really ugly when you have both a primary shape and a secondary shape array. What about unifying all into a dynamic array using something like `arrayvec`?

jedel1043 (changes requested):
Very nice!

Flamki:
@jedel1043 Addressed in e5d37780.\n\nI refactored the PIC layout to a single dynamic cache list instead of primary+secondary fields:\n- InlineCache now stores one entries: GcRefCell<Vec<PicEntry>> buffer (capacity PIC_CAPACITY).\n- Each PicEntry contains both WeakShape and Slot.\n- set/match_shape operate over this unified entry list, and megamorphic transition simply clears it.\n- Updated inline-cache tests to assert via entry_count() / contains_shape() instead of internal primary-shape assumptions.\n\nThis keeps the code path simpler and removes the split storage representation.

Flamki:
Addressed failing CI from the PIC refactor in 089d7d6e.\n\nRoot cause: code_block.rs disassembly still referenced removed InlineCache::shape field.\n\nFix:\n- Added InlineCache::first_shape_addr() for debug/disassembly output over the unified PIC entry list.\n- Updated disassembly formatting in code_block.rs to use that helper for GetLengthProperty, GetPropertyByName, GetPropertyByNameWithThis, SetPropertyByName, and SetPropertyByNameWithThis.\n\nThis keeps runtime PIC behavior unchanged and restores compile compatibility after the storage layout change.

core/engine/src/vm/inline_cache/mod.rs
jedel1043: Uhh I suggested using `arrayvec` (not `Vec`) because it would avoid allocating, since we only need a small stack-allocated dynamic array.

Flamki:
Follow-up update pushed in eb7c58de.

Changes:
- Rebased branch on latest main.
- Reworked PIC storage to be allocation-free and fixed-capacity ([Option<PicEntry>; 4] + len) while keeping a unified cache representation.
- Kept megamorphic transition semantics unchanged (clear entries + stop caching).
- Added assertions around irst_shape_addr() in inline-cache tests (initially 0, set after first cache fill, reset to 0 after megamorphic transition).

This keeps the PIC unified while avoiding heap allocation in the hot cache path.

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -11,51 +11,193 @@ use crate::{
...
+
+#[derive(Clone, Debug, Trace, Finalize)]
+struct PicEntries {
+    entries: [Option<PicEntry>; PIC_CAPACITY],
+
+    #[unsafe_ignore_trace]
+    len: u8,
+}
+
+impl PicEntries {
```
jedel1043, line 32: Again, you can use `arrayvec` here. It is already a dependency of the repository. No need to handwrite a dynamic array.

jedel1043:
Closing since the original issue was already claimed and it seems like you're just ignoring review comments.

Flamki:
Apologies if it came across as ignoring the arrayvec suggestion — that wasn’t my intention. I initially avoided it to keep the PIC storage representation explicit and minimal, but you’re right that arrayvec is already a dependency and is a cleaner solution than hand-rolling a fixed-capacity dynamic array.

I understand the PR has been closed since the issue was already claimed. If it makes sense, I’m happy to:

Rework the PIC storage using ArrayVec as suggested, and

Reopen or submit a fresh PR if the original assignee is no longer working on it.

Thanks again for the detailed review — the feedback is appreciated.

akash-R-A-J:
working on it.

PR #4672, opened 2026-02-22
Fix OOM in String.replace/replaceAll and add TryIntoJs regression test

### LINKED ISSUE #4360: `toString` missing on `TryIntoJs`
**Describe the bug**

`toString` method is missing from value constructed with `TryIntoJs`

**To Reproduce**

Minimal example:

```rust
use boa_engine::{Context, IntoJsFunctionCopied, Source, js_string, value::TryIntoJs};

fn main() {
    let get_foo_bar = get_foo_bar.into_js_function_copied(&mut context);
    context
        .register_global_builtin_callable(js_string!("getFooBar"), 0, get_foo_bar)
        .unwrap();
    let value = context.eval(Source::from_bytes("getFooBar()")).unwrap();
    println!(
        "{}",
        value
            .to_string(&mut context)
            .unwrap()
            .to_std_string_escaped()
    );
}

#[derive(TryIntoJs)]
struct FooBar {
    foo: usize,
    bar: usize,
}

fn get_foo_bar() -> FooBar {
    FooBar { foo: 0, bar: 1 }
}
```

The error message:

```
thread 'main' panicked at delight_cli/src/main.rs:47:14:
called `Result::unwrap()` on an `Err` value: JsError { inner: Native(JsNativeError { kind: Type, message: "cannot convert object to primitive value", cause: None, .. }), backtrace: None }
```

**Expected behavior**

It works without throwing if using `js_object`:

```rust
fn get_foo_bar(context: &mut Context) -> JsObject {
    // FooBar { foo: 0, bar: 1 }
    js_object!({"foo": 0, "bar": 1}, context)
}
// [object Object]
```

**Build environment**

- OS: NixOS Linux
- Version: 25.05
- Rustc version: rustc 1.88.0 (6b00bc388 2025-06-23)

**Additional context**

Using the git version of boa:

```toml
[[package]]
name = "boa_engine"
version = "0.20.0"
source = "git+https://github.com/boa-dev/boa#488ef539f70c7e80d86b568f5d60e6822eb8f261"
...
```

### LINKED ISSUE #4655: Unbounded allocation in `String.prototype.replace` causes OOM crash (from test262 `replace-math.js`)
Boa lacks a maximum string length bounds check or upfront length calculation when executing `String.prototype.replace` with capture groups. If a replacement results in a massive string, Boa attempts to physically allocate and build it in memory. This causes the process to rapidly consume all available system RAM until the OS's OOM killer forcefully terminates it.

This specific bug was found via the `test262` test suite: `test/staging/sm/String/replace-math.js`.

**To Reproduce**

```javascript
function puff(x, n)
{
  while(x.length < n)
    x += x;
  return x.substring(0, n);
}

var x = puff("1", 1 << 20);
var rep = puff("$1", 1 << 16);

try
{
  var y = x.replace(/(.+)/g, rep);
  assert.sameValue(y.length, Math.pow(2, 36));
}
catch (e)
{
  // OOM also acceptable
}
```

**Build environment:**

- OS: Arch Linux
- Rustc version: 1.95.0-nightly

**Additional context**
I observed this on a machine with 24GB of RAM. Running the snippet causes Boa to fill the entire 24GB in a few seconds, leading to my OS killing the process to prevent the system from completely freezing. V8 (Node.js) handles this instantly by rejecting (Caught error: Invalid string length).

apoorvdarshan:
## Summary
- Throw `RangeError` when `replace`/`replaceAll`/`RegExp[@@replace]` would produce a string exceeding `MAX_STRING_LENGTH` (`u32::MAX`), preventing OOM crashes
- Add upfront length checks (before the replacement loop) and per-iteration guards (before buffer extensions) in all three code paths
- Add regression test verifying `TryIntoJs`-derived structs have `Object.prototype` in their prototype chain (`toString`, `hasOwnProperty`)

Fixes #4655
Closes #4360

## Test plan
- [x] `replace_all_max_length` — `replaceAll` throws `RangeError` when result exceeds max length
- [x] `replace_regexp_max_length` — regex global replace throws `RangeError`
- [x] `replace_normal_works` — normal replace/replaceAll still work correctly
- [x] `try_into_js_has_object_prototype` — `TryIntoJs` structs have proper prototype chain
- [x] All existing string and regexp tests pass

### DIFF
--- a/core/engine/src/builtins/regexp/mod.rs
+++ b/core/engine/src/builtins/regexp/mod.rs
@@ -1665,6 +1665,16 @@ impl RegExp {
         // 16. If nextSourcePosition ≥ lengthS, return accumulatedResult.
         // 17. Return the string-concatenation of accumulatedResult and the substring of S from nextSourcePosition.
 
+        // Early length check: if just the replacement copies would exceed max string length,
+        // fail before allocating large buffers.
+        if let CallableOrString::ReplaceValue(ref rv) = replace_value {
+            let total_replacement_len = results
+                .len()
+                .checked_mul(rv.len())
+                .unwrap_or(usize::MAX);
+            string::String::check_string_max_length(total_replacement_len)?;
+        }
+
         // 13. Let accumulatedResult be the empty String.
         let mut accumulated_result = vec![];
 
@@ -1772,7 +1782,14 @@ impl RegExp {
                 //    In such cases, the corresponding substitution is ignored.
 
                 // ii. Set accumulatedResult to the string-concatenation of accumulatedResult, the substring of S from nextSourcePosition to position, and replacement.
-                accumulated_result.extend(s.get_expect(next_source_position..position).iter());
+                let substr = s.get_expect(next_source_position..position);
+                let new_len = accumulated_result
+                    .len()
+                    .checked_add(substr.len())
+                    .and_then(|l| l.checked_add(replacement.len()))
+                    .unwrap_or(usize::MAX);
+                string::String::check_string_max_length(new_len)?;
+                accumulated_result.extend(substr.iter());
                 accumulated_result.extend(replacement.iter());
 
                 // iii. Set nextSourcePosition to position + matchLength.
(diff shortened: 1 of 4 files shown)


jedel1043:
I really don't agree with this solution. It makes it such that we need to manually keep track of every string usage so that we can add the validation code.

A better solution IMO would be to directly modify the JsString constructor such that it is fallible, but that would require a lot more work.

apoorvdarshan:
@jedel1043 should i work on it or close this pr?

jedel1043:
I'll just close this PR. We can reopen if we see that making JsString construction fallible is not feasible.

apoorvdarshan:
@jedel1043 sounds good. is there any other issue you would like me to work on ?

PR #4673, opened 2026-02-22
fix(intl/plural_rules): apply notation option when resolving plural category

Lokesh0224:
Previously, Intl.PluralRules did not take the notation option into account when resolving the plural category. Even when notation: "compact" was provided, the plural resolution logic ignored it and determined the category based on the full number.

This PR updates resolve_plural() to check the notation kind. When "compact" notation is used, it constructs a compact representation of the number before calling category_for.

As a result, plural category selection now behaves correctly according to ECMA-402.

**Changes**

- Added handling for NotationKind::Compact inside resolve_plural().
- Construct a CompactDecimal when compact notation is used.
- Call category_for with the compact representation instead of the full number.
- Verified against the intl402/PluralRules test262 suite (100% passing).

Before

```
>> new Intl.PluralRules("fr", { notation: "compact" }).select(1500000)
"Other"
>> 

```
After
```
>> new Intl.PluralRules("fr", { notation: "compact" }).select(1500000)
"many"
>> 
```

### DIFF
--- a/core/engine/src/builtins/intl/plural_rules/mod.rs
+++ b/core/engine/src/builtins/intl/plural_rules/mod.rs
@@ -1,7 +1,7 @@
 mod options;
 
 use boa_gc::{Finalize, Trace};
-use fixed_decimal::{Decimal, SignedRoundingMode, UnsignedRoundingMode};
+use fixed_decimal::{CompactDecimal, Decimal, SignedRoundingMode, UnsignedRoundingMode};
 use icu_locale::Locale;
 use icu_plurals::{
     PluralCategory, PluralRuleType, PluralRules as NativePluralRules, PluralRulesOptions,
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/intl/plural_rules/mod.rs
```diff
@@ -481,7 +481,18 @@ fn resolve_plural(plural_rules: &PluralRules, n: f64) -> ResolvedPlural {
...
+
+    let category = if let NotationKind::Compact = plural_rules.notation {
+        // get the compact exponent from magnitude
+        let exp = (*fixed.magnitude_range().end()).max(0) as u8;
+        //instead of using full number, this constructs a compact representation (eg: 1500000 => significand = 1.5, exponent= 6)
+        let compact = CompactDecimal::from_significand_and_exponent(fixed.clone(), exp);
+        plural_rules.native.rules().category_for(&compact)
+    } else {
+        plural_rules.native.rules().category_for(&fixed)
+    };
```
jedel1043, line 497: Hm, I suppose we'd have to modify this for the other notation kinds in the future (IIRC ICU4X doesn't have scientific and engineering notation _yet_). Can you add a TODO comment with the links to issues https://github.com/unicode-org/icu4x/issues/3983 and https://github.com/unicode-org/icu4x/issues/3984?

core/engine/src/builtins/intl/plural_rules/mod.rs
```diff
@@ -481,7 +481,18 @@ fn resolve_plural(plural_rules: &PluralRules, n: f64) -> ResolvedPlural {
     // 8. Let s be res.[[FormattedString]].
     // 9. Let operands be ! GetOperands(s).
     // 10. Let p be ! PluralRuleSelect(locale, type, n, operands).
-    let category = plural_rules.native.rules().category_for(&fixed);
+
+    //let category = plural_rules.native.rules().category_for(&fixed);
```
jedel1043:
```suggestion
```

core/engine/src/builtins/intl/plural_rules/mod.rs
Lokesh0224: Done!

jedel1043 (approved):
Thank you!

PR #4680, opened 2026-02-23
Fix Headers iterators and Symbol.iterator in fetch runtime

### LINKED ISSUE #4679: Fetch Headers is not iterable and entries()/keys()/values() return arrays instead of iterators
## Summary
Headers in boa_runtime does not follow the Fetch standard. entries() returns an array, keys() / values() return arrays, and Headers itself is not iterable. This breaks common patterns like headers.entries().next() and for (const [k,v] of headers) .

## Expected behavior
- entries() returns an iterator with .next() .
- keys() / values() return iterators.
- Headers is iterable via Symbol.iterator (same as entries() ).

## Actual behavior
entries() returns a JsArray , keys() / values() return vectors, and there is no Symbol.iterator implementation.

## Where in code
core/runtime/src/fetch/headers.rs
entries() builds a JsArray , and keys() / values() return Vec<JsString> instead of iterators.

## Why this matters
This breaks compatibility with real‑world Fetch usage and libraries that expect standards‑compliant Headers iteration.

Nakshatra480:
Closes #4679 

## Summary
- Return proper iterators from Headers.keys/values/entries instead of arrays
- Wire Headers.prototype[Symbol.iterator] to entries for for...of and spread
- Add iterator tests that avoid relying on iteration order

## Testing
- cargo fmt -- --check
- cargo test -p boa_runtime

### DIFF
--- a/core/runtime/src/fetch/headers.rs
+++ b/core/runtime/src/fetch/headers.rs
@@ -4,10 +4,11 @@
 #![allow(clippy::needless_pass_by_value)]
 
 use boa_engine::interop::JsClass;
-use boa_engine::object::builtins::{JsArray, TypedJsFunction};
+use boa_engine::object::builtins::{JsArray, JsFunction, TypedJsFunction};
 use boa_engine::value::{Convert, TryFromJs};
 use boa_engine::{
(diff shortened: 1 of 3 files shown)


jedel1043:
There's already a PR open that fixes this (#4638)

PR #4687, opened 2026-02-23
Fix RequestInit clone body preservation

### LINKED ISSUE #4686: Bug: Request body is dropped when cloning with options
### Summary

### To Reproduce
```javascript
  method: "POST",
  body: "payload",
});

  headers: { "x-test": "1" },
});

fetch(withHeaders);
```

### Expected behavior

### Actual behavior

### Why this is critical

### Environment

Nakshatra480:
Closes #4686 

## Summary
- Preserve the original request body when cloning a Request with RequestInit options unless `body` is explicitly provided.
- Add regression tests for Request cloning edge cases (empty body, override body, no body).

## Changes
- Keep the existing body from `HttpRequest::into_parts()` and only overwrite it when `RequestInit.body` is set.
- Add unit tests covering cloning behavior with empty/overridden/no body.

## Testing
- `cargo test -p boa_runtime request_clone_`
- `cargo fmt -- --check`
- `cargo clippy --all-features --all-targets`

### DIFF
--- a/core/engine/src/builtins/regexp/mod.rs
+++ b/core/engine/src/builtins/regexp/mod.rs
@@ -1251,11 +1251,9 @@ impl RegExp {
         a.create_data_property_or_throw(0, matched_substr, context)
             .expect("this CreateDataPropertyOrThrow call must not fail");
 
-        let mut named_groups = match_value
+        let named_groups = match_value
             .named_groups()
             .collect::<Vec<(&str, Option<Range>)>>();
-        // Strict mode requires groups to be created in a sorted order
-        named_groups.sort_by(|(name_x, _), (name_y, _)| name_x.cmp(name_y));
 
         // Combines:
         // 26. Let groupNames be a new empty List.
--- a/core/engine/src/object/operations.rs
+++ b/core/engine/src/object/operations.rs
@@ -420,7 +420,9 @@ impl JsObject {
             context.vm.frames[frame_index + 1].set_exit_early(true);
         }
 
+        context.vm.host_call_depth += 1;
         let result = context.run().consume();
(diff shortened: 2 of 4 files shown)


Nakshatra480:
Hi @jedel1043, i have solved the issue. Can you pls review this PR.
Thanks for your time😊

Nakshatra480:
Hey @jedel1043  just wanted to bump this PR.
It fixes an issue where cloning a Request with RequestInit could accidentally drop the original body. Now the body is preserved unless RequestInit.body is explicitly provided.

The implementation keeps the body from into_parts() and only overrides it when a new one is passed in. I also added regression tests covering the edge cases around cloning and body overrides.

Would really appreciate a review whenever someone from the runtime side has time. Happy to make any adjustments needed 🙂

core/runtime/src/fetch/request.rs
```diff
@@ -66,8 +66,9 @@ impl RequestInit {
         request: Option<HttpRequest<Vec<u8>>>,
     ) -> JsResult<HttpRequest<Vec<u8>>> {
         let mut builder = HttpRequest::builder();
+        let mut request_body = None;
```
jedel1043: This kinda carried over from the original implementation, but `Vec::new` can already act as our default value. This way we can skip all the `Some` wrapping and the `unwrap_or_default` call at the end.

Nakshatra480: Thanks for the review! I've addressed this by using Vec::new() directly as the default value for request_body. This successfully removed the need for Option/Some wrapping and the unwrap_or_default() call at the end, making it much cleaner.

Nakshatra480:
> Nice catch! I have a small nitpick but everything else looks good!

@jedel1043 i have done the suggested changes, can you review this PR now.
Thanks for your feedback

PR #4691, opened 2026-02-23
test(intl): ignore Locale tests blocked by ICU4X limitations

Lokesh0224:
## Summary
Ignores 6 failing `Intl.Locale` tests that are blocked by upstream ICU4X limitations.

## Changes
- Ignores 3 tests related to `h24` hour cycle support, which ICU4X does not currently implement
- Ignores 3 tests related to Unicode extension value canonicalization (e.g. `islamicc` → `islamic-civil`)

## Related Issues
- https://github.com/unicode-org/icu4x/issues/6597 — h24 hour cycle not supported
- https://github.com/unicode-org/icu4x/issues/3483 — Unicode extension value canonicalization

These tests will be re-enabled once the upstream ICU4X issues are resolved.

### DIFF
--- a/test262_config.toml
+++ b/test262_config.toml
@@ -81,4 +81,14 @@ tests = [
     "test/intl402/Intl/getCanonicalLocales/unicode-ext-canonicalize-calendar.js",
     "test/intl402/Intl/getCanonicalLocales/unicode-ext-canonicalize-timezone.js",
     "test/intl402/Intl/getCanonicalLocales/transformed-ext-canonical.js",
+    # h24 hour cycle not supported by ICU4X.
+    # TODO: Remove when https://github.com/unicode-org/icu4x/issues/6597 is fixed.
+    "test/intl402/Locale/getters.js",
+    "test/intl402/Locale/constructor-getter-order.js",
+    "test/intl402/Locale/constructor-options-hourcycle-valid.js",
+    # Extension canonicalization not yet supported by ICU4X.
+    # TODO: Remove when https://github.com/unicode-org/icu4x/issues/3483 is fixed.
+    "test/intl402/Locale/prototype/calendar/canonicalize.js",
+    "test/intl402/Locale/constructor-non-iana-canon.js",
+    "test/intl402/Locale/constructor-options-canonicalized.js",
 ]


Lokesh0224:
While investigating other Intl.Locale failures, I found they were also blocked by the same ICU4X limitations (h24 and extension canonicalization), not actual Boa bugs. So opened this follow-up to ignore those tests as well.
Are there any issues you'd suggest I pick up next? @jedel1043

jedel1043 (approved):
Thank you! Looks good!

jedel1043:
@Lokesh0224 Since I see you're kind of interested in our Intl implementation, maybe you can pick up the DateTimeFormat TODOs?

We have a couple of those in our DTF implementation, and I think some of them could just be resolved by investigating on ICU4X's side how to do certain things e.g. https://github.com/boa-dev/boa/blob/b965589ce77b884964433e0494d005c6d6204910/core/engine/src/builtins/intl/date_time_format/options.rs#L595-L596

Lokesh0224:
Thanks for the suggestion. The calendar algorithm TODO looks like a great starting point. I'm swamped with college work for the next day or so, but I'll definitely pick this up right after — already looking forward to exploring how ICU4X handles it!

Lokesh0224:
Hey @jedel1043 , I checked with the calendar algorithm TODO. I found
ICU4X doesn't expose a DataMarkerAttributes-based validation for calendars like it does for numbering systems (DecimalSymbolsV1), so we can't use "validate_extension" here. The input is already validated via CalendarAlgorithm::try_from in mod.rs, which rejects unknown strings with a RangeError.

The missing piece — getting computed calendar options back from ICU4X for resolvedOptions() — is tracked upstream at unicode-org/icu4x#5868 (Resolved options for ECMA-402, P1 for their 2.3 release) and unicode-org/icu4x#7320 (settling what calendars are valid in AnyCalendar).

Lokesh0224:
While investigating this, I also noticed that fn resolved_options() is not implemented at all for DateTimeFormat — calling it throws a TypeError.
Screenshot from 2026-02-27 13-08-09

When i checked with other resolved_options() like NumberFormat its working as expected, 
image

jedel1043:
>  I found
ICU4X doesn't expose a DataMarkerAttributes-based validation for calendars like it does for numbering systems (DecimalSymbolsV1), so we can't use "validate_extension" here.

It should, because ICU4X also needs to validate that the data provided has the required data for the calendar, it's just mostly unstable IIRC. You can try digging through [`icu_datetime`'s provider module](https://docs.rs/icu_datetime/2.1.1/icu_datetime/provider/index.html) to see how they fetch the calendar data, and make a similar check in our side.

Lokesh0224:
@jedel1043, I dug into icu_datetime's provider module like you suggested. Found that the calendar markers like DatetimePatternsDateGregorianV1 are in the bundled data and in MARKERS. However, when I tried dry_load with various request formats (locale-keyed, root locale, with calendar extensions), all returned IdentifierNotFound even for CalendarAlgorithm::Gregory with locale "en":

**Debug output:**
Err(DataError { kind: IdentifierNotFound, marker: Some(DataMarkerId { debug: "DatetimePatternsDateGregorianV1", hash: [116, 100, 109, 104, 6, 28, 40, 45] }), str_context: None, silent: true })

Since invalid calendar strings are already rejected at the syntax level via CalendarAlgorithm::try_from, that part is covered. I wasn’t able to find a working way to perform data-level validation via dry_load, so I’m not sure if that’s the intended approach here or if I’m still keying the request incorrectly.

For now, I’ve kept validation simple: accept all syntactically valid CalendarAlgorithm variants and rely on earlier parsing to reject invalid inputs.

Am I missing something about how these calendar markers are expected to be keyed, or is there a different validation mechanism in ICU that I should be using?

[#4753](https://github.com/boa-dev/boa/pull/4753), in this pr i've updated the below comment.
 // TODO: determine the correct way to verify the calendar algorithm data.

jedel1043:
To know how to query the provider correctly for that specific data key, you might have to dig into the [`DateTimeFormatter`](https://docs.rs/icu_datetime/2.1.1/icu_datetime/struct.DateTimeFormatter.html) code, specifically its [`try_new`](https://docs.rs/icu_datetime/2.1.1/icu_datetime/struct.DateTimeFormatter.html#method.try_new) method

Lokesh0224:
@jedel1043 , Thanks for navigating me! Dug into DateTimeFormatter::try_new_internal and traced the full calendar loading path — it goes through FormattableAnyCalendarKind::from_preferences → FormattableAnyCalendar::try_new_with_buffer_provider, but both types are private to icu_datetime. 
After looking at how AnyCalendarKind::new works internally, found that CalendarAlgorithm has a TryInto<AnyCalendarKind> impl, so used that combined with AnyCalendar::try_new_with_buffer_provider as the public-facing equivalent to achieve the same validation.
Opened a PR for this: [#4790](https://github.com/boa-dev/boa/pull/4790)

PR #4692, opened 2026-02-23
 ignore regexp-v-flag tests due to library limitations

amrkhaled104:
### What I did:
I added `regexp-v-flag` to the ignored features in `test262_config.toml`. 

### Why?
I found that many tests are failing because of the `v` flag. After some debugging, I realized the problem is not in Boa itself, but in the `regress` crate that Boa uses for Regex. 

Currently, `regress` doesn't support the new **Unicode Sets (v flag)** syntax. Because of this, any test that uses the `v` flag will fail or panic during parsing. 

### Evidence:
Tests for `search`, `match`, `matchAll`, and `replace` are all failing when they try to use the `v` flag in a RegExp. We should ignore these tests for now until the `regress` library adds support for it.

**Reference:**
- Proposal: https://github.com/tc39/proposal-regexp-v-flag
- This will help keep our CI green.


amrkhaled104:
I was focusing on the string tests, but I didn’t try running the full test suite. I stopped because I noticed some related tests were already passing, so I’ll search for them and check if I might have done something wrong.

amrkhaled104:
@jedel1043 
I’ve managed to fix the test262 failures for RegExp surrogate pair matching in Non-Unicode mode.

The Issue:
In Non-Unicode mode, Boa was passing full 32-bit Code Points to regress, but at runtime, it was searching through 16-bit Code Units (UCS2). This mismatch caused failures when matching characters like 𠮷.

The Fix:
I updated compile_native_regexp to manually "flatten" the pattern into individual UTF-16 units before compilation. This ensures the matcher treats surrogate pairs as two separate 16-bit atoms, aligning with the ECMAScript spec for Non-Unicode execution.

image

 I'll provide a detailed breakdown in the Pull Request!

Issue #4688, opened 2026-02-23
Support symbol-named methods in #[boa(class)] macro


Rahul-2k4:
## Summary
`#[boa(class)]` currently supports string-named methods/properties but does not provide a direct way to expose symbol-named methods (for example `Headers.prototype[Symbol.iterator]`).

## Motivation
In PR #4638 we had to manually register `[Symbol.iterator]` in `core/runtime/src/fetch/mod.rs` by constructing a function object and defining the symbol property on the prototype after `js_module::boa_register`.

This works, but it is verbose and duplicates boilerplate that should ideally live in the class macro path.

## Requested enhancement
Add support in `#[boa(class)]` (and related derives/attributes) for symbol-keyed methods so builtins can declare these declaratively, similar to other method definitions.

## Current workaround
Manual prototype patching is used in PR #4638 with a TODO to migrate once macro support exists.

## References
- PR #4638
- `core/runtime/src/fetch/mod.rs` manual symbol wiring block

Yuvraj-cyborg:
Hey I'd like to work on this issue !

jedel1043:
Sure. Let us know if you need guidance for this!

jedel1043:
Implemented in #4697

PR #4697, opened 2026-02-23
feat(macros): support symbol-named methods in #[boa(class)]

Yuvraj-cyborg:
# Summary

This PR adds support for well-known symbol-keyed methods, statics, and accessors
in the `#[boa(class)]` macro. This allows builtins to declare them declaratively
instead of manually patching prototypes after registration.

**Closes:** #4688

---

# What Changed

## `core/macros/src/class.rs`

- Introduced a `MethodName` enum (`String | Symbol`) replacing the raw `String` name on:
  - `Function`
  - `Accessor`
  - `ClassVisitor`
- `MethodName::to_key_tokens()` now emits:
  - `js_string!(...)` for string keys
  - `JsSymbol::iterator()` (etc.) for symbol keys
- `name_of()` now:
  - Checks for `#[boa(symbol = "...")]` first
  - Validates against the 13 well-known symbol names
  - Falls through to `#[boa(rename = "...")]` or default rename scheme otherwise
- `serialize_class_impl()` now uses polymorphic key tokens for:
  - `builder.method()`
  - `builder.static_method()`
  - `builder.accessor()`
- `match` is handled via `Ident::new_raw` since it is a Rust keyword

## `core/macros/src/module.rs`

- Updated `Function::from_sig` callsite to wrap the name in `MethodName::String(...)`

## `core/runtime/src/fetch/headers.rs`

- Added `#[boa(symbol = "iterator")]` on `JsHeaders`
- Replaced the manual prototype patching TODO from #4638

## `tests/macros/tests/class.rs`

- Added `Pair` struct with:
  - `#[boa(symbol = "toPrimitive")]`
  - `#[boa(symbol = "iterator")]`
- Added `boa_class_symbol_methods` test covering:
  - Numeric coercion (`+p`)
  - Addition (`p + 3`)
  - Spread (`[...p]`)
  - `for...of`

---

# Accepted Symbol Names

The following well-known symbol names are supported:

- `asyncIterator`
- `hasInstance`
- `isConcatSpreadable`
- `iterator`
- `match`
- `matchAll`
- `replace`
- `search`
- `species`
- `split`
- `toPrimitive`
- `toStringTag`
- `unscopables`

Unknown names produce a compile-time error listing valid options.

---

# Usage Example

```rust
#[boa_class]
impl MyType {
    #[boa(symbol = "iterator")]
    fn iter(&self, ctx: &mut Context) -> JsValue {
        /* ... */
    }

    #[boa(symbol = "toPrimitive")]
    fn to_prim(&self) -> i32 {
        42
    }
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,6 +391,7 @@ dependencies = [
 name = "boa_cli"
 version = "1.0.0-dev"
 dependencies = [
+ "async-channel 2.5.0",
  "boa_engine",
  "boa_gc",
  "boa_parser",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -27,6 +27,7 @@ color-eyre.workspace = true
 cow-utils.workspace = true
 futures-concurrency.workspace = true
 futures-lite.workspace = true
+async-channel = "2.5.0"
 
 [features]
 default = [
--- a/cli/src/executor.rs
+++ b/cli/src/executor.rs
@@ -1,0 +1,143 @@
+use std::{
+    cell::RefCell,
+    collections::{BTreeMap, VecDeque},
+    mem,
+    rc::Rc,
+};
+
+use boa_engine::{
+    Context, JsResult,
+    context::time::JsInstant,
+    job::{GenericJob, Job, JobExecutor, NativeAsyncJob, PromiseJob, TimeoutJob},
+};
+use futures_concurrency::future::FutureGroup;
+use futures_lite::{StreamExt, future};
+
+use crate::{logger::SharedExternalPrinterLogger, uncaught_job_error};
+
+pub(crate) struct Executor {
+    promise_jobs: RefCell<VecDeque<PromiseJob>>,
+    async_jobs: RefCell<VecDeque<NativeAsyncJob>>,
+    timeout_jobs: RefCell<BTreeMap<JsInstant, Vec<TimeoutJob>>>,
+    generic_jobs: RefCell<VecDeque<GenericJob>>,
+
+    printer: SharedExternalPrinterLogger,
+}
+
+impl Executor {
+    pub(crate) fn new(printer: SharedExternalPrinterLogger) -> Self {
+        Self {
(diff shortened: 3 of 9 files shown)


Yuvraj-cyborg:
cc: @jedel1043

core/runtime/src/fetch/headers.rs
```diff
@@ -170,6 +170,12 @@ impl JsHeaders {
         .into()
     }
 
+    /// `Headers.prototype[Symbol.iterator]()` — delegates to `entries()`.
+    #[boa(symbol = "iterator")]
+    pub fn iterator(&self, context: &mut Context) -> JsValue {
+        self.entries(context)
+    }
```
jedel1043: This is incorrect. An `Array` is _not_ an iterator object. I would suggest just removing this to keep the current PR only focused on the new feature itself.

Yuvraj-cyborg: removed it !

Yuvraj-cyborg:
cc: @jedel1043

jedel1043:
No need to CC me, just requesting a review already pings my notifications haha

Yuvraj-cyborg:
> No need to CC me, just requesting a review already pings my notifications haha

bad habit :)

core/macros/src/class.rs
```diff
@@ -16,14 +16,79 @@ use syn::{
...
+        "iterator",
+        "match",
+        "matchAll",
+        "replace",
+        "search",
+        "species",
+        "split",
+        "toPrimitive",
+        "toStringTag",
+        "unscopables",
```
jedel1043: The set of well known symbols is continually on influx (e.g. #4649 added more symbols). I would suggest not parsing the method name and just directly forwarding it to `JsSymbol:: #ident ()`. We still do need to convert to snake_case, but you can probably use a crate to do that.

Yuvraj-cyborg:
hey @jedel1043 could you review this !!
I have made the changes

jedel1043 (approved):
Nice work!

Issue #4643, opened 2026-02-20
RegExp execution lacks backtracking limit, allowing ReDoS


akash-R-A-J:
**Describe the bug**

Boa’s RegExp implementation allows unbounded backtracking, which can cause the engine to hang or take an extremely long time on certain valid regular expression patterns.
Because the underlying `regress` engine has no execution limit or timeout, pathological patterns (e.g., nested quantifiers) can trigger exponential backtracking and block the executing thread.

**To Reproduce**
```javascript
var re = new RegExp("(a+)+$");
re.test("a".repeat(25) + "!");  // hangs or takes very long time
```
Increasing the input size increases execution time exponentially.

**Expected behavior**

Regular expression execution should be bounded so that untrusted patterns or inputs cannot freeze the engine.
Other JavaScript engines mitigate this class of issue using backtracking limits or interruption checks. Boa currently executes RegExp matching without any such guard.

**Build environment (please complete the following information):**

- OS: Windows 11
- Boa: current `main` 
- Target triple: x86_64-pc-windows-msvc
- Rustc version: 1.93.1 (01f6ddf75 2026-02-11)

**Additional context**

- This is a known class of issue ([ReDoS / CWE-1333](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS)).
- The behavior appears to originate from unbounded execution in the regress backend, which currently exposes no backtracking or execution limit API.

![Image](https://github.com/user-attachments/assets/c7d4e43a-4e01-4b13-8fbb-8f0067ae4da5)

I’m happy to work on a fix if this aligns with the project’s direction.

akash-R-A-J:
Hi @jedel1043, I'd like to work on this issue. Could you please assign it to me?

Happy to discuss the approach before starting if you'd prefer.

jedel1043:
@akash-R-A-J As you've mentioned, this requires having an API in regress' side to limit execution, so you would have to work with them first. I will mark this as blocked for now, and you can return to this issue in the future when the feature gets implemented.

akash-R-A-J:
@jedel1043 Thanks! I'll open an issue in regress and keep you posted.

PR #4698, opened 2026-02-23
regexp: mitigate ReDoS on ASCII inputs with PikeVM path

Flamki:
This Pull Request mitigates #4643.

It changes the following:
- Adds a dedicated non-optimized `regress::Regex` matcher for PikeVM execution in `RegExp` objects.
- Uses PikeVM matching for ASCII inputs in `RegExpBuiltinExec`, avoiding pathological backtracking in common ReDoS cases such as `(a+)+$`.
- Keeps existing UTF-16/UCS-2 backtracking paths for non-ASCII inputs to preserve current encoding semantics.
- Adds a regression test covering the nested-quantifier ASCII case.

Notes:
- This is a mitigation path inside Boa while the broader backtracking-budget API is still blocked in `regress`.

### DIFF
--- a/core/engine/src/builtins/regexp/mod.rs
+++ b/core/engine/src/builtins/regexp/mod.rs
@@ -42,6 +42,12 @@ mod tests;
 pub struct RegExp {
     /// Regex matcher.
     matcher: Regex,
+
+    /// Non-optimized matcher used by the PikeVM executor on ASCII inputs.
+    ///
+    /// The PikeVM backend in `regress` does not support optimized bytecode.
+    pikevm_matcher: Regex,
+
     flags: RegExpFlags,
     original_source: JsString,
     original_flags: JsString,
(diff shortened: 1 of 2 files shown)


Flamki:
Follow-up fix pushed in 63d66115.

Addresses the CI failures from the previous run:
- Added explicit lifetime args for 
egress::backends::PikeVMExecutor<'_, '_> at both ind call sites (fixes elided-lifetimes-in-paths under -D warnings).
- Updated doc comments to use backticks around PikeVM (fixes clippy::doc-markdown).
- Applied rustfmt-compliant line wrapping in the ASCII UTF-16 conversion path.

This should resolve the build/lint/fmt failures that were cascading across jobs.

Flamki:
@jedel1043 I did a full cleanup pass on `#4698` so it avoids the same issues we hit in earlier PRs.

rebased the branch onto latest `main`, kept the PikeVM ReDoS mitigation logic intact, and preserved the CI/lint fixes (explicit `PikeVMExecutor<'_, '_>` lifetimes, doc markdown fixes, and rustfmt-compliant formatting).

- also added one extra regression test to cover the ASCII-in-UTF16 execution path, not just the plain ASCII string path, so coverage is better aligned with the new matcher behavior.

Current branch head: `91d9da41`.
Ready for CI + maintainer review.

jedel1043:
According to https://github.com/ridiculousfish/regress/issues/140, V8 has the same behaviour, so I don't think this requires a mitigation.

PR #4699, opened 2026-02-23
vm: avoid stack overflow on recursive accessor calls

### LINKED ISSUE #4535: Getters and setters can cause stack overflows
**Describe the bug**
Our current implementation of getter and setter properties makes it possible to overflow the stack with bigger recursion limits.

**To Reproduce**

```javascript
async function* f() {}
f().return({
    get then() {
    this.then;
    },
});
```

Then run in debug mode (you might need to bump the default value of recursion in `RuntimeLimits` to something like 2048)

https://github.com/boa-dev/boa/blob/6200e4b5eec1284b87e902c551f09154c1113fe0/core/engine/src/vm/runtime_limits.rs#L22

```bash
cargo run --bin boa -- script.js
```

**Expected behavior**

The engine throws a recursion limit error.

- OS: Arch Linux
- Version: 6.17.3-arch2-1
- Rustc version: rustc 1.91.1

**Additional context**

When executing a property get or set, we directly run any Javascript code if the property is a getter or setter property:

https://github.com/boa-dev/boa/blob/6200e4b5eec1284b87e902c551f09154c1113fe0/core/engine/src/object/internal_methods/mod.rs#L738-L741

https://github.com/boa-dev/boa/blob/6200e4b5eec1284b87e902c551f09154c1113fe0/core/engine/src/object/internal_methods/mod.rs#L918-L920

Which runs the bytecode directly:

https://github.com/boa-dev/boa/blob/6200e4b5eec1284b87e902c551f09154c1113fe0/core/engine/src/object/operations.rs#L394-L427

This needs to be adjusted such that it bubbles up to the parent if the property is a getter, which gives a chance to the vm to prepare execution and increase its recursion count.

Flamki:
This Pull Request fixes/closes #4535.

## Summary
- Track nested host-driven VM re-entry (JsObject::call / JsObject::construct) with a new Vm::host_call_depth counter.
- Include host_call_depth in Context::check_runtime_limits() recursion checks so recursive accessor calls fail with RuntimeLimitError::Recursion instead of overflowing the native stack.
- Unignore and strengthen the regression test in m/tests.rs by exercising the async-generator 	hen getter recursion path with a higher recursion limit.

## Notes
- I could not run cargo in this environment (toolchain not available), so CI is expected to validate formatting, lint, and tests.

### DIFF
--- a/core/engine/src/object/operations.rs
+++ b/core/engine/src/object/operations.rs
@@ -420,7 +420,9 @@ impl JsObject {
             context.vm.frames[frame_index + 1].set_exit_early(true);
         }
 
+        context.vm.host_call_depth += 1;
         let result = context.run().consume();
+        context.vm.host_call_depth = context.vm.host_call_depth.saturating_sub(1);
 
         context.vm.pop_frame().expect("frame must exist");
 
@@ -474,7 +476,9 @@ impl JsObject {
             context.vm.frames[frame_index + 1].set_exit_early(true);
         }
 
+        context.vm.host_call_depth += 1;
         let result = context.run().consume();
+        context.vm.host_call_depth = context.vm.host_call_depth.saturating_sub(1);
 
         context.vm.pop_frame().expect("frame must exist");
 
--- a/core/engine/src/vm/mod.rs
+++ b/core/engine/src/vm/mod.rs
@@ -91,6 +91,13 @@ pub struct Vm {
     /// because we don't push a frame for them.
     pub(crate) native_active_function: Option<JsObject>,
 
+    /// Number of nested host calls that re-enter the VM via `Context::run()`.
+    ///
+    /// This is incremented by high-level host entry points such as
+    /// [`JsObject::call`](crate::object::JsObject::call) and
+    /// [`JsObject::construct`](crate::object::JsObject::construct).
+    pub(crate) host_call_depth: usize,
+
     pub(crate) shadow_stack: ShadowStack,
 
     #[cfg(feature = "trace")]
@@ -422,6 +429,7 @@ impl Vm {
             pending_exception: None,
             runtime_limits: RuntimeLimits::default(),
             native_active_function: None,
+            host_call_depth: 0,
             shadow_stack: ShadowStack::default(),
             #[cfg(feature = "trace")]
             trace: false,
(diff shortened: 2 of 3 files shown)


Flamki:
Follow-up hardening pushed in 9966682d.

What I changed:
- Rebased branch on latest main.
- Added a dedicated regression test for recursive setters:
  - 
ecursion_in_setter_throws_uncatchable_error
  - asserts RuntimeLimitError::Recursion for set x(v) { this.x = v; } recursion.

This complements the existing async-generator getter regression and ensures both accessor call directions are covered by runtime-limit behavior.

jedel1043:
I'd suggest measuring the performance impact of this change, because keeping track of recursive native calls sounds like it could be detrimental for perf. You can use https://github.com/boa-dev/data/blob/main/bench/bench-v8/combined.js for this.

Flamki:
@jedel1043 thanks for raising this, I took a careful pass on perf.

I built clean release binaries for:
- `upstream/main` (`549d14a5`)
- this branch (`390f0a1b`, plus the latest test-only commit)

I first tried to run the exact benchmark you suggested from `boa-dev/data` (`bench/bench-v8/combined.js`). On my Windows setup, that script stack-overflows on both revisions before producing comparable scores, so I couldn’t get a reliable canonical A/B from it locally.

To still get signal, I ran fallback release A/B hot-path measurements (11 runs each, first run dropped as warmup) focused on accessor/property/function throughput:

| Script | main mean (ms) | branch mean (ms) | delta |
|---|---:|---:|---:|
| accessor_hot.js | 3894.162 | 3629.003 | -6.81% |
| plain_prop_hot.js | 1446.026 | 1381.159 | -4.49% |
| func_hot.js | 7877.123 | 7616.737 | -3.31% |

From this local fallback measurement, I did not observe a regression from `host_call_depth` tracking.

I also lowered both recursion regression test limits to `128` for platform stability and re-ran:
`cargo test -p boa_engine vm::tests` -> `26 passed, 0 failed`.

jedel1043 (approved):
Thanks!

Maybe in the future we'll finally have coroutines to lift all the inner calls to the main evaluation loop, but in the meantime we need to work around this limitation.

PR #4700, opened 2026-02-23
fix: handle surrogate pairs in non-unicode regex patterns

amrkhaled104:
#### **Overview**

This PR fixes `test262` failures related to RegExp matching when surrogate pairs (like `𠮷`) are used without the `u` or `v` flags.

#### **The Problem**

I found that in Non-Unicode mode, Boa was compiling the regex pattern using full 32-bit **Code Points**. However, at runtime, the engine uses `find_from_ucs2` which looks at raw 16-bit **Code Units**.

For example:

* The character `𠮷` was compiled as one atom: `0x20BB7`.
* But the input string in memory is two units: `[0xD842, 0xDFB7]`.
This mismatch caused the regex to fail even if the character was clearly there.

#### **What I Changed**

1. In `core/engine/src/builtins/regexp/mod.rs`:
* I updated `compile_native_regexp` to check if the Unicode flag is missing.
* If there is no `u` or `v` flag, I manually "flatten" the pattern. I iterate through each code point and decompose it into its individual UTF-16 units (surrogates) before passing them to the `regress` matcher.
* This makes the compiled pattern match the 16-bit structure of the input string.

#### **How I Tested**

I ran the `boa_tester` with the following command:

```bash
cargo run --bin boa_tester -- run -s test/built-ins/String/prototype/match/regexp-prototype-match-v-u-flag.js

```

**Result:** All tests passed
image

and exec test 
image

---

### DIFF
--- a/core/engine/src/builtins/regexp/mod.rs
+++ b/core/engine/src/builtins/regexp/mod.rs
@@ -340,22 +340,22 @@ impl RegExp {
         // 14. If parseResult is a non-empty List of SyntaxError objects, throw a SyntaxError exception.
 
         // If u or v flag is set, fullUnicode is true — compile as full codepoints.
-        let full_unicode = flags.contains(RegExpFlags::UNICODE)
-            || flags.contains(RegExpFlags::UNICODE_SETS);
+        let full_unicode =
+            flags.contains(RegExpFlags::UNICODE) || flags.contains(RegExpFlags::UNICODE_SETS);
 
         // In non-Unicode mode, check if pattern contains named groups (?<...).
         // Named groups with astral Unicode identifiers (e.g. (?<𝑓𝑜𝑥>)) require
         // full codepoints to work correctly with regress group name handling.
-        let has_named_groups = p
-            .code_points()
-            .collect::<Vec<_>>()
-            .windows(3)
-            .any(|w| {
-                matches!(
-                    (w[0], w[1], w[2]),
-                    (CodePoint::Unicode('('), CodePoint::Unicode('?'), CodePoint::Unicode('<'))
+        let has_named_groups = p.code_points().collect::<Vec<_>>().windows(3).any(|w| {
+            matches!(
+                (w[0], w[1], w[2]),
+                (
+                    CodePoint::Unicode('('),
+                    CodePoint::Unicode('?'),
+                    CodePoint::Unicode('<')
                 )
-            });
+            )
+        });
 
         let matcher = if full_unicode || has_named_groups {
             // Unicode mode (u/v flag) OR pattern has named groups:
@@ -365,7 +365,7 @@ impl RegExp {
                     JsNativeError::syntax()
                         .with_message(format!("failed to create matcher: {}", error.text))
                 })?
-            } else {
+        } else {
             // Non-Unicode mode with no named groups:
             // compile as raw UTF-16 code units so that surrogate pairs
             // (e.g. 𠮷 = [0xD842, 0xDFB7]) are matched correctly by find_from_ucs2.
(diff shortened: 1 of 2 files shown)


amrkhaled104:
@jedel1043 why cI not work ? 
image

amrkhaled104:
@jedel1043  All 8 tests are now passing 💪

amrkhaled104:
Are there any issues you'd suggest I pick up ? @jedel1043

core/engine/src/builtins/regexp/mod.rs
```diff
@@ -338,18 +338,61 @@ impl RegExp {
...
+        let has_named_groups = p.code_points().collect::<Vec<_>>().windows(3).any(|w| {
+            matches!(
+                (w[0], w[1], w[2]),
+                (
+                    CodePoint::Unicode('('),
+                    CodePoint::Unicode('?'),
+                    CodePoint::Unicode('<')
+                )
+            )
+        });
```
jedel1043: I kind of don't understand why we need this. IIRC the only things that influence if a regex is parsed as UTF16 or UCS2 are the `u` and `v` flags, right? And that's being taken care of by the `full_unicode` check above.

amrkhaled104: @jedel1043  I want to explain the full picture. Sorry, I forgot to update the PR description after my last changes 
In the first version of the PR, I found that many Prototype tests were failing. The reason was how we handle Unicode and Non-Unicode modes 
In Non-Unicode mode, JavaScript treats 'heavy' characters (like emojis or special math symbols) as two 16-bit units, not one character. But our engine was treating them as one unit. This made the search index wrong
I used flat_map to split these characters into two units. This fixed the indexing and made the Prototype tests pass.

After that, the test `test\built-ins\RegExp\named-groups\non-unicode-property-names-valid.js` failed. I realized this is a special case Even in Non-Unicode mode, Named Groups (?<name>) need the character to stay as one 'Identity' (Code Point). If we split the name into two units, the Regex engine cannot find the group name because the name becomes 'broken' So, I added a check:
`let has_named_groups = !full_unicode && p.to_std_string_escaped().contains("(?<");`  
or 
```        let has_named_groups = p.code_points().collect::<Vec<_>>().windows(3).any(|w| {
            matches!(
                (w[0], w[1], w[2]),
                (
                    CodePoint::Unicode('('),
                    CodePoint::Unicode('?'),
                    CodePoint::Unicode('<')
                )
            )
        });
```
This code checks for the (?< sequence If the pattern has named groups, we keep the characters as Code Points (like Unicode mode). .

jedel1043: Then this edge case is not our responsibility. It should be the responsibility of regress to parse group names as unicode even if parsing without unicode support.

I would suggest opening a bug on their side reporting this, and removing the hack. The test should pass afterwards without having to hack around this.

amrkhaled104: You are right,I will update the PR

amrkhaled104: @jedel1043 I have updated the code and removed the hack. To keep the CI green while we wait for a fix in regress, should I add this specific test to the test262_config.toml ignore list?

jedel1043: Yep, please do! And if you can, add a TODO on the config file pointing to regress' issue.

amrkhaled104: Got it! I will add the test to test262_config.toml with a TODO note in this PR.

amrkhaled104:
@jedel1043  Anything else?

jedel1043 (approved):
Nope, everything looks good!

PR #4703, opened 2026-02-24
fix: Transitive prototype inline caching

### LINKED ISSUE #4265: Transitive prototype inline caching
The current implementation of inline caching only works for properties that are in the object or in it's direct prototype.

This is very limiting, as transitive properties (properties of prototype's prototype) are treated as a cache miss.

So code the code below always gives a cache miss:

```JavaScript
class X {
   getX() {
      return 42;
   }
}

class Y extends X {
   getY() {
      return 42;
   }
}

let y = new Y();
y.getY() // cacheable

let y = new Y();
y.getX() // never cacheable
```

sarthakNITT:
Description
Extended inline caching to support transitive prototype properties. Previously, only own or direct-prototype properties were cacheable, properties found deeper in the chain (e.g. grandparent prototype) always caused a cache miss. Fixed by tracking 
prototype_hops through the lookup and using it to walk the correct depth on cache hits.

This Pull Request fixes/closes #4265 .

It changes the following:

- inline_cache/mod.rs: Added prototype_hops field; updated set() and match_or_reset() to include it
- internal_methods/mod.rs: Added prototype_hops counter to InternalMethodPropertyContext; replaced set_not_cacheable_if_already_prototype() with prototype_hops += 1 in ordinary_get, ordinary_try_get, ordinary_has_property
- get/property.rs, get/name.rs, set/property.rs: Fast path walks prototype_hops levels up the chain; slow path saves hops to cache via ic.set(shape, slot, hops)
- inline_cache/tests.rs: Added test for 2-hop transitive prototype caching

### DIFF
--- a/core/engine/src/vm/inline_cache/mod.rs
+++ b/core/engine/src/vm/inline_cache/mod.rs
@@ -140,7 +140,10 @@ impl InlineCache {
         for entry in &entries[..count] {
             let cached_addr = entry.shape.to_addr_usize();
             if cached_addr == shape_addr {
-                return entry.shape.upgrade().map(|s| (s, entry.slot, entry.prototype.clone()));
+                return entry
+                    .shape
+                    .upgrade()
+                    .map(|s| (s, entry.slot, entry.prototype.clone()));
             }
         }
 


jedel1043:
Cool! Can you check the performance change from this change? You can use https://github.com/boa-dev/data/tree/main/bench/bench-v8 for that, just run the combined.js file with `cargo run --release --bin boa -- ../data/bench/bench-v8/combined.js` from the main branch and from this PR's commit head and comment the results.

sarthakNITT:
@jedel1043 Here are the performance change results

• Main branch results - 
PROGRESS Richards
RESULT Richards 233
PROGRESS DeltaBlue
RESULT DeltaBlue 242
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 216
PROGRESS RayTrace
RESULT RayTrace 433
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 561
PROGRESS RegExp
RESULT RegExp 78.1
PROGRESS Splay
RESULT Splay 754
PROGRESS NavierStokes
RESULT NavierStokes 475
SCORE 309

• Pr commit head results -
-v8/combined.js`
PROGRESS Richards
RESULT Richards 225
PROGRESS DeltaBlue
RESULT DeltaBlue 253
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 210
PROGRESS RayTrace
RESULT RayTrace 455
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 553
PROGRESS RegExp
RESULT RegExp 77.7
PROGRESS Splay
RESULT Splay 690
PROGRESS NavierStokes
RESULT NavierStokes 450
SCORE 304

jedel1043:
Thanks!

Hmm, performance barely changed. I suspect it's because we're going through the whole prototype chain in core/engine/src/vm/opcode/get/name.rs, which kind of nullifies the optimization.

For this to have a perf benefit we'd have to somehow store that prototype in the cache such that we don't need to go through the whole prototype chain to get the correct prototype.

sarthakNITT:
@jedel1043 Updated the implementation based on your feedback. Instead of storing a hop count and traversing the prototype chain on every cache hit, the inline cache now stores the owning prototype JsObject directly — so on a cache hit we jump straight to the correct prototype with zero chain traversal.

Also these are the results for main and pr head -
Main - 
v8/combined.js`
PROGRESS Richards
RESULT Richards 233
PROGRESS DeltaBlue
RESULT DeltaBlue 240
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 215
PROGRESS RayTrace
RESULT RayTrace 462
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 553
PROGRESS RegExp
RESULT RegExp 77.9
PROGRESS Splay
RESULT Splay 756
PROGRESS NavierStokes
RESULT NavierStokes 472
SCORE 310

Pr head - 310
v8/combined.js`
PROGRESS Richards
RESULT Richards 231
PROGRESS DeltaBlue
RESULT DeltaBlue 256
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 211
PROGRESS RayTrace
RESULT RayTrace 455
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 554
PROGRESS RegExp
RESULT RegExp 77.3
PROGRESS Splay
RESULT Splay 758
PROGRESS NavierStokes
RESULT NavierStokes 468
SCORE 310

jedel1043:
Hmm, performance barely changed again... Is there anything we're missing?

sarthakNITT:
I think we might be missing intermediate prototype shape validation. Currently, the cache is keyed only on the child's shape, so if an intermediate prototype in the chain changes, the cache doesn't invalidate and may be incorrectly trusted or conservatively skipped. Storing and validating the intermediate prototype's shape in the IC entry as well could make the cache safer to use in more cases and produce a real performance gain. 

Should I proceed with this?

jedel1043:
Yeah, go ahead! Also, just keep running the perf tests and adjusting things until there's a noticeable perf increase, otherwise it won't make sense to implement all this code 😅

sarthakNITT:
@jedel1043 Ran extensive perf testing to try to get a noticeable score increase.

- Stored the owning prototype directly in the IC (skip chain traversal on hit). Score stayed at 312 (same as main)
- Added debug counters: 88.1% hit rate (256M hits / 34M misses), with 25M being prototype hits, the IC is actively being used
- Implemented polymorphic IC (up to 4 shape entries per site, fixed inline array, no heap allocation). Splay jumped from 725 → 882–986, but overall score stayed at 311-312

Why overall score doesn't move: 
- The V8 suite is mixed, Splay (polymorphic tree ops) gained significantly, but monomorphic-heavy benchmarks (Richards, DeltaBlue) regressed slightly because a larger IC struct (4 entries inline) hurts cache-line efficiency for the overwhelmingly monomorphic majority. It's a wash.
- The polymorphic IC is a real improvement for poly call sites. The fundamental bottleneck for the aggregate score seems to be interpreter dispatch, not property lookup.

- Do you have any idea what else we can try here to get a visible score increase? Or should we focus on a different area of the engine?

sarthakNITT:
@jedel1043 Was going through main's `GetPropertyByName` code and noticed that prototype property IC was already implemented there (via shape.prototype() on cache hit). So our changes were improving an already-optimized path, that's why the score barely moved. Also implemented polymorphic IC (4 entries per site) which improved Splay, but overall score stayed the same since main's IC was already catching most hits. 

Any suggestions on what to target next?

jedel1043:
Huh, looking at your poly IC I found it interesting that you chose to evict pointers instead of just disabling the cache to avoid constantly caching unused shapes.

I mention this because we have a PR in the queue that implements this (#4740) but disables the cache instead of evicting shapes, so it might be worth checking if your approach is better for performance?

sarthakNITT:
@jedel1043 #4740   megamorphic disable is cleaner than my eviction, avoids thrashing at high-polymorphic sites entirely.

One gap: it doesn't store the prototype directly in `PicEntry`, so hits on inherited properties still call `shape.prototype()`. This PR has that (transitive + direct prototype caching) before the poly-IC commits.

Can I co-author on #4740 to add that else revert my commits here?. If you have any other suggestions specifically around the original IC issue, happy to work on those here instead.

jedel1043:
#4740 was merged, so you can rebase and start playing with it!

jedel1043:
Gonna close this since the optimization was apparently already there. If you still have plans of improving the polymorphic IC with transitive prototype caching, feel free to open another PR.

PR #4706, opened 2026-02-24
[bug]: Fix TimeoutJob key collision and run_jobs infinite loop

ashnaaseth2325-oss:
## Summary

This PR fixes two issues in `SimpleJobExecutor` and the CLI `Executor` related to timeout scheduling.

Timeouts were previously stored as:

```rust
timeout_jobs: RefCell<BTreeMap<JsInstant, TimeoutJob>>,
```

This allowed only one job per timestamp. Multiple `setTimeout(..., 0)` calls at the same instant caused silent overwrites via:

```rust
self.timeout_jobs.borrow_mut().insert(now + t.timeout(), t);
```

Additionally, a mismatch between the loop guard (`>= now`) and `split_off(&now)` caused `run_jobs()` to hang when a job was due exactly at `now`.

---

## Impact

* Silent loss of timeout callbacks
* `run_jobs()` infinite loop with `FixedClock`
* CLI REPL could hang
* Affects default `Context` executor

---

## Fix

1. Support multiple jobs per timestamp:

```rust
timeout_jobs: RefCell<BTreeMap<JsInstant, Vec<TimeoutJob>>>,
```

```rust
.entry(now + t.timeout())
.or_default()
.push(t);
```

2. Align due-check with `split_off` semantics (`>` instead of `>=`).

3. Added regression test ensuring two zero-delay timeouts both fire.

---

## Conclusion

This change makes timeout scheduling correct and deterministic, preventing silent job loss and infinite execution loops while preserving existing behavior and passing all current tests.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -574,7 +574,7 @@ fn add_runtime(printer: SharedExternalPrinterLogger, context: &mut Context) {
 struct Executor {
     promise_jobs: RefCell<VecDeque<PromiseJob>>,
     async_jobs: RefCell<VecDeque<NativeAsyncJob>>,
-    timeout_jobs: RefCell<BTreeMap<JsInstant, TimeoutJob>>,
+    timeout_jobs: RefCell<BTreeMap<JsInstant, Vec<TimeoutJob>>>,
     generic_jobs: RefCell<VecDeque<GenericJob>>,
 
     printer: SharedExternalPrinterLogger,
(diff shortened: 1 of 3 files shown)


ashnaaseth2325-oss:
hey @jedel1043
This should resolve the timeout collision and infinite loop while preserving spec behavior. Happy to adjust if anything looks off. Test262 has passed. Please approve the remaining workflows so the rest of CI can run. Thanks!

hansl (approved):
LGTM.

hansl:
Please fix the lint errors on the tests, and I'll merge it. Thanks!

ashnaaseth2325-oss:
Hey @hansl @jedel1043,
Thanks! I’ve fixed the lint issues and all checks are now passing.

PR #4710, opened 2026-02-25
Fix BigInt shift_right/shift_left for large shift amounts

Ansh-699:
`1000n >> 1000000000000000n` incorrectly throws `RangeError: Maximum BigInt size exceeded` instead of returning `0n`. The shift amount does not fit in `i32`, so `to_i32()` returns `None` and the code falls through to the error branch unconditionally.

Per the ECMAScript spec (and confirmed against V8/SpiderMonkey), arithmetic right-shifting by a very large positive amount should return:
- `0n` for non-negative values
- `-1n` for negative values

The same fix is applied symmetrically to `shift_left` when the shift amount is a large negative number (equivalent to a large right-shift).

**Changes:**
- `JsBigInt::shift_right`: check sign of `y` in the `None` arm — large positive `y` returns `0n`/`-1n`, large negative `y` keeps the `RangeError`
- `JsBigInt::shift_left`: mirror logic — large negative `y` returns `0n`/`-1n`, large positive `y` keeps the `RangeError`
- Updated tests: replaced `assert_native_error` with `assert_eq` for the right-shift case, added coverage for negative BigInt and zero

### DIFF
--- a/core/engine/src/bigint.rs
+++ b/core/engine/src/bigint.rs
@@ -228,21 +228,15 @@ impl JsBigInt {
             // <https://tc39.es/ecma262/#sec-numeric-types-bigint-signedRightShift>
             // implies that for very large positive y the result converges to
             // 0n (x >= 0) or -1n (x < 0). V8 and SpiderMonkey agree.
-            None => {
-                if y.inner.sign() == Sign::Minus {
-                    // x >> (large negative) is equivalent to x << (large positive), which overflows.
-                    Err(JsNativeError::range()
-                        .with_message("Maximum BigInt size exceeded")
-                        .into())
-                } else {
-                    // x >> (large positive): all bits are shifted out.
-                    if x.inner.sign() == Sign::Minus {
-                        Ok(Self::new(RawBigInt::from(-1)))
-                    } else {
-                        Ok(Self::zero())
-                    }
-                }
-            }
+            None => match (x.inner.sign(), y.inner.sign()) {
+                // x >> (large positive): all bits are shifted out.
+                (Sign::Minus, Sign::Plus) => Ok(Self::new(RawBigInt::from(-1))),
+                (_, Sign::Plus) => Ok(Self::zero()),
+                // x >> (large negative) is equivalent to x << (large positive), which overflows.
+                (_, _) => Err(JsNativeError::range()
+                    .with_message("Maximum BigInt size exceeded")
+                    .into()),
+            },
         }
     }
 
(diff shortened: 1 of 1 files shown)


jedel1043:
> Per the ECMAScript spec (and confirmed against V8/SpiderMonkey), arithmetic right-shifting by a very large positive amount should return:
    0n for non-negative values
    -1n for negative values

I would suggest adding a comment to the specific section of the spec that mentions this, just for future reference.

Ansh-699:
Per the ECMAScript specification, the Signed Right Shift Operator (`>>`) is defined here:
https://tc39.es/ecma262/#sec-signed-right-shift-operator
For BigInt operands, it is defined mathematically as:
> ℤ(floor(x / 2^y))

For sufficiently large positive `y`:
- If `x ≥ 0`, then `x / 2^y → 0` ⇒ `floor(...) = 0` ⇒ `0n`
- If `x < 0`, then `x / 2^y → 0⁻` ⇒ `floor(...) = -1` ⇒ `-1n`

Therefore, shifting by a very large positive amount must return `0n` or `-1n`, not throw a `RangeError`. This matches V8 and SpiderMonkey behavior.

jedel1043:
@Ansh-699 That sound more like it's not part of the spec.

I guess we can still apply this safeguard, but let's be very clear with a comment that this is just a best-effort safeguard, _not_ a spec-related result.

Ansh-699:
ok no worries

Ansh-699:
Updated — thanks for the feedback!

Changes in this push:
- Reworded comments to make it clear this is a **best-effort safeguard**, not a direct spec requirement. Added links to the relevant spec sections (`sec-numeric-types-bigint-signedRightShift` and `sec-numeric-types-bigint-leftShift`).
- Added tests for `shift_left` with large negative amounts (the mirror case), which should improve the patch coverage.

core/engine/src/bigint.rs
```diff
@@ -219,9 +220,29 @@ impl JsBigInt {
...
+                        .into())
+                } else {
+                    // x >> (large positive): all bits are shifted out.
+                    if x.inner.sign() == Sign::Minus {
+                        Ok(Self::new(RawBigInt::from(-1)))
+                    } else {
+                        Ok(Self::zero())
+                    }
+                }
+            }
```
jedel1043: Nitpick: maybe this would look a bit nicer with a match
```rust
match (x.inner.sign(), y.inner.sign()) {
    (Sign::Minus, Sign::Plus) => -1,
    (Sign::Plus, Sign::Plus) => 0,
    (_, Sing::Minus) => Err,
}
```

core/engine/src/bigint.rs
```diff
@@ -231,9 +252,25 @@ impl JsBigInt {
...
+                        Ok(Self::new(RawBigInt::from(-1)))
+                    } else {
+                        Ok(Self::zero())
+                    }
+                } else {
+                    // x << (large positive) overflows.
+                    Err(JsNativeError::range()
+                        .with_message("Maximum BigInt size exceeded")
+                        .into())
+                }
```
jedel1043: Same suggestion as above; a match could be easier to read overall.

jedel1043 (approved):
Looks good!

PR #4711, opened 2026-02-25
feat(cli): add --time flag to report script execution timing

### LINKED ISSUE #4709: Feature: Add --time CLI flag to report script execution timing
### Summary

The Boa CLI has no built-in way to measure how long parsing and execution take. A `--time` flag would print a timing breakdown to stderr after execution, similar to how other engines expose timing (V8's `d8 --times`, etc.).

### Expected Behavior

$ boa --time script.js Hello, World!
Parsing: 3.21ms Execution: 12.45ms Total: 15.66ms

### Proposed Implementation

- Add a `--time` (`-t`) boolean flag to [Opt](cci:2://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:97:0-169:1) in [cli/src/main.rs](cci:7://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:0:0-0:0).
- Wrap parsing/execution with `std::time::Instant` and print results to stderr.
- ~30–50 lines of change, scoped entirely to the `cli` crate.
- Fits naturally alongside existing `--trace` and `--optimizer-statistics` flags.

Nakshatra480:
### Summary

This adds a `--time` flag to the Boa CLI that prints how long parsing and execution took, similar to what other engines offer (V8's `d8 --times`, etc.).

When the flag is passed, timing info is printed to stderr so it doesn't interfere with script output on stdout:
$ boa --time script.js Hello, World!
Execution: 1.56ms

For modules, parsing and execution are measured separately:
$ boa --time -m module.mjs 42
Parsing: 1.82ms Execution: 1.27ms Total: 3.09ms

Without the flag, there's zero overhead — no `Instant::now()` calls and no extra output.

### Changes

All changes are in [cli/src/main.rs](cci:7://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:0:0-0:0):

- Added `--time` boolean flag to the [Opt](cci:2://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:97:0-169:1) struct (long flag only, since `-t` is already used by `--trace`)
- Wrapped `context.eval()` in [evaluate_expr](cci:1://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:302:0-328:1) and [evaluate_file](cci:1://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:330:0-389:1) with conditional `Instant` measurements using `args.time.then(Instant::now)`
- For the module path, `Module::parse()` and `load_link_evaluate()` are measured separately
- In [evaluate_expr](cci:1://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:302:0-328:1), `context.run_jobs()` is called unconditionally after eval so async work is always drained regardless of whether timing is enabled

No new dependencies - uses `std::time::Instant` which was already imported (as `Duration`) in the file.

Closes #4709

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -269,7 +269,11 @@ impl Drop for Counters {
         let mut total = Duration::ZERO;
         eprintln!();
         for (name, elapsed) in &counters {
-            eprintln!("{:<width$} {elapsed:.2?}", format!("{name}:"), width = max_width);
+            eprintln!(
+                "{:<width$} {elapsed:.2?}",
+                format!("{name}:"),
+                width = max_width
+            );
             total += *elapsed;
         }
         if counters.len() > 1 {


Nakshatra480:
Hey @jedel1043  just following up here 🙂

This adds a --time flag to the Boa CLI that prints parsing and execution timings to stderr (so it won’t mix with script stdout).

It measures parse vs. execution separately for modules and has no runtime overhead when the flag isn’t used.

Would appreciate a review from someone on the CLI/runtime side, happy to make any changes if needed.

cli/src/main.rs
jedel1043: Might be useful to measure only parsing and ast generation time. Can you also add timings for `dump`?

jedel1043 (changes requested):
Nice! I have a very small suggestion.

cli/src/main.rs
```diff
@@ -256,6 +261,10 @@ fn dump<R: ReadChar>(src: Source<'_, R>, args: &Opt, context: &mut Context) -> R
             };
 
         println!("{dump}");
+
+        if let Some(start) = start {
+            eprintln!("\nParsing:   {:.2?}", start.elapsed());
+        }
```
jedel1043: This should be before dumping the AST. Otherwise, you'd also measure the time it takes to print to the console

cli/src/main.rs
```diff
@@ -319,10 +328,22 @@ fn evaluate_expr(
...
             Ok(v) => printer.print(format!("{}\n", v.display())),
             Err(ref v) => printer.print(uncaught_error(v)),
         }
+
+        if let Err(err) = context.run_jobs() {
+            printer.print(uncaught_job_error(&err));
+        }
+
+        if let Some(start) = start {
+            let elapsed = start.elapsed();
```
jedel1043: This might also need to be split like in the module case. Otherwise, we'd incorrectly report the time it takes to print values and errors.

cli/src/main.rs
Nakshatra480: Thanks for the suggestion, totally makes sense to include timings for dump output.
 I’ll add parse + AST generation timings there so it’s consistent with the rest.

cli/src/main.rs
Nakshatra480: you’re right about avoiding console print time in the measurement.
 I’ll move the timing capture to happen before the dump output is generated.

cli/src/main.rs
Nakshatra480: Agreed, the timing should be split like the module path to avoid print overhead.
 I’ll mirror the parse/exec split so the numbers are accurate.

jedel1043:
I kinda feel like this pattern of `start = now`, then executing something and `elapsed = time` can probably be simplified with a couple of structs:

```rust
struct Timer<'a> {
    name: &'static str,
    now: Instant,
    counters: &'a mut Vec<(&'static str, Duration)>
}

impl Drop for Timer<'a> {
    fn drop(&mut self) {
        self.counters.push((self.name, self.now.elapsed()));
    }
}


struct Counters {
    counters: Option<Vec<(&'static str, Duration)>>
}

impl Counters {
    fn new_timer(&mut self, name: &'static str) -> Option<Timer<'_>> {
        if let Some(counters) = &mut self.counters {
            Some(Timer {
              name,
              now: Instant::now(),
              counters
            })
        } else { None }
        
    }
}

impl Drop for Counters {
    // Code to print all timers
}
```

That way, to measure something you just do like

```rust
{
    let _timer = counters.new_timer("Parsing");
    // parse
}

// more code
```

Nakshatra480:
Hey @jedel1043, thanks again for the review and the suggestions.

I’ve just pushed some updates to address your feedback:
- **`Timer` & `Counters` struct**: Implemented your suggestion! This makes the code super clean and ensures there's zero overhead when the flag isn't passed (we just use an `Option` and don't create `Instant`s).
- **Split timings everywhere**: The timing is now consistently split into `Parsing` and `Execution` (or `AST generation` for `--dump-ast`) across all execution paths (module, script, and expression).
- **Avoided print overhead**: Moved `print_timing()` to happen completely independently from result/error printing, so we aren't measuring console I/O time anymore. I also made sure that async job execution (`run_jobs()`) is accurately included in the execution timer for consistency.
- **Fixed `-t` flag conflict**: Also noticed that the short `-t` flag was accidentally reassigned from `--trace` to `--time` in the previous iteration. I’ve reverted that so `--trace` keeps its short flag, avoiding any breaking changes there.

Let me know if this looks good to you or if there's anything else you'd like me to tweak!

cli/src/main.rs
```diff
@@ -214,21 +217,87 @@ enum FlowgraphDirection {
...
+}
+
+fn print_timing(
+    counters: Counters,
+    first: &'static str,
+    second: &'static str,
+    first_label: &'static str,
+    second_label: &'static str,
+    total_label: &'static str,
+) {
```
jedel1043: My suggestion would be to put all this on the `Drop` implementation of `Counters`, so that you don't need to call it after every execution. This also allows just creating a single `Counters` at the start of the main function, and use that handle to push timers anywhere.

Also, why pass the labels manually? You already have their names, you just need to properly format them

Nakshatra480: Hey @jedel1043, I've just pushed a refactored version that implements exactly what you suggested:
- Moved all the timing output logic into an `impl Drop for Counters`. This completely eliminates the need for manual `print_timing()` calls at the end of every execution path!
- The labels are now properly formatted dynamically from the timer names themselves (e.g., `Parsing:`), with automatic column alignment based on the longest timer name inside the `counters` vector.
- Removed the old `into_vec()` and the huge `print_timing()` function altogether, which makes the whole thing a lot cleaner and way more idiomatic.

Let me know if this looks good to you or if there's anything else you'd like me to tweak!

PR #4720, opened 2026-02-25
Fix `delete` with optional chaining not actually deleting properties

### LINKED ISSUE #4719: `delete` with optional chaining (`?.`) does not actually delete the property
`delete` combined with optional chaining always returns `true` but never performs the actual deletion.

**Reproduction:**
```js
var o = { a: { b: 1 } };
console.log(delete o?.a.b); // true
console.log(o.a.b);         // 1 (should be undefined)
```

All forms are affected: `delete o?.x`, `delete o?.a["b"]`, `delete o?.a?.b`, `delete o?.[key]`.

**Expected:** The property is deleted (matches V8/SpiderMonkey behavior).
**Actual:** The expression evaluates to `true`, but the property remains.

**Root cause:** In the bytecompiler, `Access::from_expression()` has no arm for `Expression::Optional`, so `delete` on an optional chain falls through to the else branch that compiles it as a regular expression evaluation and pushes `true` without emitting any delete opcode.

Ansh-699:
## Description

`delete` with optional chaining (`?.`) returned `true` but did not actually delete the property.

### Root Cause

`Access::from_expression()` has no arm for `Expression::Optional`, so the delete compilation falls to the else branch that just evaluates the expression and pushes `true` without emitting a delete opcode.

### Fix

Added `compile_optional_delete` and `compile_optional_delete_final` methods in the bytecompiler (`mod.rs`). These walk the optional chain, compile intermediate accesses normally, and emit `DeletePropertyByName`/`DeletePropertyByValue` for the final chain item. A new `Expression::Optional` check in `unary.rs` routes delete compilation through this path.

### Example (before fix)
```js
let o = {a: {b: 1}};
delete o?.a.b; // returns true but o.a.b is still 1
```

### Example (after fix)
```js
let o = {a: {b: 1}};
delete o?.a.b; // returns true and o.a.b is undefined
```

Closes #4719

### DIFF
--- a/core/engine/src/tests/operators.rs
+++ b/core/engine/src/tests/operators.rs
@@ -271,20 +271,24 @@ fn unary_delete() {
 #[test]
 fn delete_optional_chaining() {
     run_test_actions([
-        // Basic: delete o?.prop actually removes the property.
+        // Basic: delete o?.prop actually removes the property descriptor.
         TestAction::assert(indoc! {r#"
             var o = { a: 1 };
-            delete o?.a === true && o.a === undefined && !("a" in o)
+            delete o?.a === true
+                && Object.getOwnPropertyDescriptor(o, "a") === undefined
         "#}),
-        // Chain: delete o?.a.b removes nested property, keeps sibling.
+        // Chain: delete o?.a.b removes nested property descriptor, keeps sibling.
         TestAction::assert(indoc! {r#"
             var o = { a: { b: 1, c: 2 } };
-            delete o?.a.b === true && o.a.b === undefined && o.a.c === 2
+            delete o?.a.b === true
+                && Object.getOwnPropertyDescriptor(o.a, "b") === undefined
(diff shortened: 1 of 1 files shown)


nekevss (changes requested):
This is looking pretty good overall.

Can there be a few unit tests added to test the behavior of this PR.

Ansh-699:
okie i'll add in a bit

core/engine/src/tests/operators.rs
jedel1043, line 1: Tests look kinda good, but instead of asserting that the value is undefined I'd use `Object.getOwnPropertyDescriptor` to assert that the descriptor itself gets deleted. This ensures we're properly checking deletion.

Ansh-699: okay

Ansh-699:
is it okay now ?

nekevss (approved):
LGTM

PR #4721, opened 2026-02-25
# Cache Module Namespace Binding Resolution

### LINKED ISSUE #4704: # Performance: Cache Module Namespace Binding Resolution Instead of Resolving on Every Access
**Labels:** `enhancement`, `performance`

---

##  Problem

In `namespace.rs:297`, both `[[GetOwnProperty]]` and `[[Get]]` for module namespace exotic objects perform a full `ResolveExport` walk on every single property access.

`ResolveExport`:
- Traverses the module graph
- Allocates a `HashSet` to track circular references
- Clones export names

This work is repeated on every access, even though the binding never changes after module linking.

---

rajat552:
## Problem

Both `[[GetOwnProperty]]` and `[[Get]]` on module namespace exotic objects call `ResolveExport` on every single property access.

This results in:
- Traversing the module graph repeatedly
- Allocating a fresh `HashSet` for circular reference tracking each time
- Cloning export names on every access

All of this work is repeated even though bindings never change after module linking.

---

## Why This Matters

- Module namespace accesses are common in real-world code:

  ```js
  import * as mod from "./lib";
  mod.foo;

##  Closes #4704

### DIFF
--- a/core/engine/src/module/source.rs
+++ b/core/engine/src/module/source.rs
@@ -729,21 +729,20 @@ impl SourceTextModule {
                 if resolution.module() != star_resolution.module() {
                     return Err(ResolveExportError::Ambiguous);
                 }
-                match (
-                    resolution.binding_name(),
-                    star_resolution.binding_name(),
-                ) {
-                    // 3. If resolution.[[BindingName]] is not starResolution.[[BindingName]] and either
-                    //    resolution.[[BindingName]] or starResolution.[[BindingName]] is namespace,
-                    //    return ambiguous.
+                // 3. If resolution.[[BindingName]] is not starResolution.[[BindingName]] and either
+                //    resolution.[[BindingName]] or starResolution.[[BindingName]] is namespace,
+                //    return ambiguous.
+                // 4. If resolution.[[BindingName]] is a String, starResolution.[[BindingName]] is a
+                //    String, and SameValue(resolution.[[BindingName]], starResolution.[[BindingName]])
+                //    is false, return ambiguous.
+                match (resolution.binding_name(), star_resolution.binding_name()) {
                     (BindingName::Namespace, BindingName::Name(_))
                     | (BindingName::Name(_), BindingName::Namespace) => {
+                        // See step 3 above.
                         return Err(ResolveExportError::Ambiguous);
                     }
-                    // 4. If resolution.[[BindingName]] is a String, starResolution.[[BindingName]] is a
-                    //    String, and SameValue(resolution.[[BindingName]], starResolution.[[BindingName]])
-                    //    is false, return ambiguous.
                     (BindingName::Name(res), BindingName::Name(star)) if res != star => {
+                        // See step 4 above.
                         return Err(ResolveExportError::Ambiguous);
                     }
                     _ => {}


rajat552:
@jedel1043  please review this PR let me know if there is any changes required

jedel1043:
Please fix CI first.

rajat552:
@jedel1043 i made required changes let me know if there is any changes required

core/engine/src/module/namespace.rs
```diff
@@ -26,6 +26,9 @@ pub struct ModuleNamespace {
     module: Module,
     #[unsafe_ignore_trace]
     exports: IndexSet<JsString, BuildHasherDefault<FxHasher>>,
+    /// Cached binding resolutions for each export name.
+    /// Populated once during namespace creation; bindings are immutable after linking.
+    resolved_bindings: FxHashMap<JsString, ResolvedBinding>,
```
zhuzhu81998, line 40: hmm it seems to me that you can/should add `#[unsafe_ignore_trace]` here. because the resolved bindings dont actually contain (EDIT: indepedently) gc tracked objects (same as `exports`). (perhaps check that and add a comment to briefly justify using `#[unsafe_ignore_trace]`)

core/engine/src/module/namespace.rs
```diff
@@ -62,6 +65,20 @@ impl ModuleNamespace {
         let mut exports = names.into_iter().collect::<IndexSet<_, _>>();
         exports.sort();
 
+        // Pre-resolve all export bindings and cache them.
+        // After linking, ResolveExport results are stable (per spec),
+        // so we can safely cache them to avoid repeated graph traversals.
+        let mut resolved_bindings = FxHashMap::default();
+        for name in &exports {
+            if let Ok(binding) = module.resolve_export(
+                name.clone(),
```
zhuzhu81998: I think if you modify the `resolve_export` a bit, you can avoid `clone` here. because right now there is a bit of a double cloning for this for the branch `ModuleKind::SourceText`

core/engine/src/module/namespace.rs
```diff
@@ -278,24 +298,16 @@ fn module_namespace_exotic_try_get(
         return Ok(None);
     };
 
-    // 4. Let m be O.[[Module]].
```
zhuzhu81998, line 281: can we place this spec-related comment somewhere meaningfully instead of deleting it outright?

core/engine/src/module/namespace.rs
```diff
@@ -358,24 +370,16 @@ fn module_namespace_exotic_get(
         return Ok(JsValue::undefined());
     };
 
-    // 4. Let m be O.[[Module]].
```
zhuzhu81998, line 361: some as above, try to keep this somehow

core/engine/src/module/namespace.rs
```diff
@@ -278,24 +298,16 @@ fn module_namespace_exotic_try_get(
...
-        .resolve_export(
-            export_name.clone(),
-            &mut HashSet::default(),
-            context.interner(),
-        )
-        .expect("6. Assert: binding is a ResolvedBinding Record.");
+    // 5. Let binding be the cached ResolveExport result.
+    let binding = obj
+        .get_resolved_binding(&export_name)
+        .expect("binding must be cached after namespace creation");
```
zhuzhu81998: hmm i think it would be better to have a fallback path using resolve_export in case of a cache miss (instead of directly panicking) even though i cant think of a case where it might happen right now. bonus would also be that we can keep `6. Assert: ...` comment for spec conformance.

core/engine/src/module/namespace.rs
```diff
@@ -358,24 +370,16 @@ fn module_namespace_exotic_get(
...
```
zhuzhu81998: same as above

core/engine/src/module/mod.rs
```diff
@@ -214,7 +214,7 @@ impl ModuleKind {
 /// Return value of the [`Module::resolve_export`] operation.
 ///
 /// Indicates how to access a specific export in a module.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Trace, Finalize)]
```
zhuzhu81998: you dont actually need the `Trace, Finalize`. there is no gc tracked things here. (see the comment with `#[unsafe_ignore_trace]`)

core/engine/src/module/mod.rs
```diff
@@ -224,7 +224,7 @@ pub(crate) struct ResolvedBinding {
 ///
 /// Note that a resolved binding can resolve to a single binding inside a module (`export var a = 1"`)
 /// or to a whole module namespace (`export * as ns from "mod.js"`).
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Trace, Finalize)]
```
zhuzhu81998: same as above

core/engine/src/module/source.rs
```diff
@@ -1795,9 +1792,13 @@ impl SourceTextModule {
                         .kind()
                         .as_module()
                         .expect("last environment should be the module env")
-                        .set_indirect(locator.binding_index(), export_locator.module, name),
+                        .set_indirect(
+                            locator.binding_index(),
+                            export_locator.module().clone(),
```
zhuzhu81998, line 1810: if you do the `#[unsafe_ignore_trace]` and the other corresponding things, this cloning here should be redundant?

core/engine/src/module/namespace.rs
rajat552: Good point. I’ve added #[unsafe_ignore_trace] and verified that resolved_bindings does not contain any GC-tracked objects (same reasoning as exports).

I also added a brief comment explaining why ignoring trace is safe here.

core/engine/src/module/namespace.rs
rajat552: Applied the same fallback logic here as well to ensure consistent behavior across both code paths.

rajat552:
@jedel1043 
I made required changes verify it

core/engine/src/module/mod.rs
```diff
@@ -216,8 +216,8 @@ impl ModuleKind {
 /// Indicates how to access a specific export in a module.
 #[derive(Debug, Clone)]
 pub(crate) struct ResolvedBinding {
-    module: Module,
-    binding_name: BindingName,
+    pub(crate) module: Module,
+    pub(crate) binding_name: BindingName,
```
zhuzhu81998: i fail to see the point of the extra `pub(crate)` here.

core/engine/src/module/namespace.rs
```diff
@@ -279,23 +311,28 @@ fn module_namespace_exotic_try_get(
...
+    // ResolveExport call on cache miss for robustness.
+    let binding = match obj.get_resolved_binding(&export_name) {
+        Some(b) => b.clone(),
+        None => module
+            .resolve_export(
+                &export_name,
+                &mut rustc_hash::FxHashSet::default(),
+                context.interner(),
+            )
+            .expect("export name must be resolvable"),
```
zhuzhu81998: keep the original comment please ie `6. Assert: binding is a ResolvedBinding Record.`

core/engine/src/module/namespace.rs
```diff
@@ -279,23 +311,28 @@ fn module_namespace_exotic_try_get(
     };
 
     // 4. Let m be O.[[Module]].
-    let m = obj.module();
+    let module = obj.module().clone();
```
zhuzhu81998: dont see the point of cloning here

core/engine/src/module/namespace.rs
```diff
@@ -279,23 +311,28 @@ fn module_namespace_exotic_try_get(
...
-        .resolve_export(
-            export_name.clone(),
-            &mut HashSet::default(),
-            context.interner(),
-        )
-        .expect("6. Assert: binding is a ResolvedBinding Record.");
+    // Use the pre-resolved cache when available; fall back to a fresh
+    // ResolveExport call on cache miss for robustness.
+    let binding = match obj.get_resolved_binding(&export_name) {
+        Some(b) => b.clone(),
```
zhuzhu81998: do you need to clone here? to have the same type across the match arms, just delete this clone and add a & for the `None` branch

core/engine/src/module/namespace.rs
```diff
@@ -359,23 +396,28 @@ fn module_namespace_exotic_get(
     };
 
     // 4. Let m be O.[[Module]].
-    let m = obj.module();
+    let module = obj.module().clone();
```
zhuzhu81998: same as above

core/engine/src/module/namespace.rs
```diff
@@ -359,23 +396,28 @@ fn module_namespace_exotic_get(
...
-        .resolve_export(
-            export_name.clone(),
-            &mut HashSet::default(),
-            context.interner(),
-        )
-        .expect("6. Assert: binding is a ResolvedBinding Record.");
+    // Use the pre-resolved cache when available; fall back to a fresh
+    // ResolveExport call on cache miss for robustness.
+    let binding = match obj.get_resolved_binding(&export_name) {
+        Some(b) => b.clone(),
```
zhuzhu81998: same as above try to avoid cloning

core/engine/src/module/namespace.rs
```diff
@@ -359,23 +396,28 @@ fn module_namespace_exotic_get(
...
+        None => module
+            .resolve_export(
+                &export_name,
+                &mut rustc_hash::FxHashSet::default(),
+                context.interner(),
+            )
+            .expect("export name must be resolvable"),
```
zhuzhu81998: same as above (comment).

core/engine/src/module/source.rs
```diff
@@ -1631,7 +1628,7 @@ impl SourceTextModule {
                         // deferred to initialization below
                         imports.push(ImportBinding::Namespace {
                             locator,
-                            module: resolution.module,
+                            module: resolution.module.clone(),
```
zhuzhu81998: cloning here should not be necessary here either?

zhuzhu81998:
@rajat552 are you still working on this? should be fairly simple changes.

rajat552:
@zhuzhu81998 Yes, I'm working on it. I'll submit the changes shortly.

core/engine/src/module/mod.rs
rajat552: Removed `pub(crate)` from `ResolvedBinding` fields (`module`, `binding_name`) and updated `source.rs` to use the existing accessor methods (`module()` and `binding_name_ref()`) instead of direct field access. Everything compiles and is formatted correctly.

core/engine/src/module/namespace.rs
rajat552: Moved the spec `// 6. Assert:` comment above the binding resolution and reused it as the `.expect()` message in both places in `namespace.rs`. Builds and formatting checks pass.

core/engine/src/module/namespace.rs
rajat552: Removed unnecessary `.clone()` from `obj.module()` calls in `namespace.rs` since `resolve_export` only needs `&self`. Builds and formatting checks pass.

core/engine/src/module/namespace.rs
rajat552: Replaced `b.clone()` with a reference-based approach so both match arms return `&ResolvedBinding`. Builds and formatting checks pass.

core/engine/src/module/source.rs
rajat552: Added `into_module()` to `ResolvedBinding` and replaced `resolution.module().clone()` with `resolution.into_module()` to avoid the extra clone. Builds and formatting checks pass.

core/engine/src/module/namespace.rs
rajat552: Already addressed in a previous revision — `obj.module()` in `namespace.rs` no longer uses `.clone()`.

core/engine/src/module/mod.rs
```diff
@@ -238,9 +238,14 @@ impl ResolvedBinding {
         &self.module
     }
 
-    /// Gets the binding associated with the resolved export.
-    pub(crate) fn binding_name(&self) -> BindingName {
-        self.binding_name.clone()
+    /// Consumes `self` and returns the module from which the export resolved.
+    pub(crate) fn into_module(self) -> Module {
+        self.module
```
zhuzhu81998, line 243: now it looks a lot better without those clones but what do you need this for 😂

rajat552: It's used to avoid `module().clone()`. In that path the `ResolvedBinding` is consumed anyway, so `into_module()` lets us move the module instead of cloning it.

rajat552:
Thanks for the review!

core/engine/src/module/namespace.rs
```diff
@@ -26,6 +26,16 @@ pub struct ModuleNamespace {
...
     #[unsafe_ignore_trace]
     exports: IndexSet<JsString, BuildHasherDefault<FxHasher>>,
+    /// Cached binding resolutions for each export name.
+    /// Populated once during namespace creation; bindings are immutable after linking.
+    ///
+    /// SAFETY: Every `Module` inside a `ResolvedBinding` is a transitive dependency
+    /// of the parent `module` field (which IS traced). Those modules are reachable
+    /// through `SourceTextModule::loaded_modules` / `SyntheticModule`, so tracing
+    /// them again here would be redundant. Skipping the trace avoids walking the
+    /// entire hashmap on every GC cycle.
```
jedel1043: thought: this is a good optimization but in general we kinda don't need to be smart about things like this. Our GC already ignores pointers that were already traced before.

Though I would still leave it like this since it avoids a lookup to check if the pointer is alive, but you should update the comment.

core/engine/src/module/mod.rs
```diff
@@ -238,9 +238,14 @@ impl ResolvedBinding {
...
+    }
+
+    /// Gets a reference to the binding associated with the resolved export.
+    pub(crate) const fn binding_name_ref(&self) -> &BindingName {
```
jedel1043: I would rename this to `binding_name`. Just makes the thing cleaner and if you want an owned version you can just clone from the ref.

core/engine/src/module/namespace.rs
rajat552: Thanks for the suggestion! I've updated the comment to clarify that the optimization isn't strictly required for correctness since the GC already ignores previously traced pointers, but we keep it to avoid the extra lookup.

core/engine/src/module/mod.rs
rajat552: Good point, thanks! I've renamed `binding_name_ref` to `binding_name` and updated all usages accordingly.

jedel1043 (approved):
Looks great!

jedel1043:
Ping me when you fix the formatting so that I can send this to the merge queue

rajat552:
@jedel1043 All checks passed

jedel1043:
Also thank you @zhuzhu81998 for the extensive review! It made my work much easier 😅

zhuzhu81998:
> Also thank you @zhuzhu81998 for the extensive review! It made my work much easier 😅

really getting swarmed with prs here XD

PR #4723, opened 2026-02-25
fix(fetch): don't drop body when cloning a Request with options

### LINKED ISSUE #4686: Bug: Request body is dropped when cloning with options

jakharmonika364:
Fixes #4686

When you do `new Request(existingRequest, { headers: {...} })`, the body from the original request gets lost. The issue is in [into_request_builder](cci:1://file:///e:/opensource/boa/core/runtime/src/fetch/request.rs:58:4-120:5) — `into_parts()` was splitting off the body as [_body](cci:1://file:///e:/opensource/boa/core/runtime/src/fetch/tests/request.rs:50:0-60:1) and just ignoring it.

Fix: save the original body and use it if `options.body` wasn't set.

### DIFF
--- a/core/runtime/src/fetch/request.rs
+++ b/core/runtime/src/fetch/request.rs
@@ -66,8 +66,9 @@ impl RequestInit {
         request: Option<HttpRequest<Vec<u8>>>,
     ) -> JsResult<HttpRequest<Vec<u8>>> {
         let mut builder = HttpRequest::builder();
+        let mut original_body = None;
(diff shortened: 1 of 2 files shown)


nekevss:
Duplicate of #4687

PR #3141, opened 2023-07-15
Added examples

postmeback:
This Pull Request fixes/closes #2337 

It changes the following:

Added two new functions to test `Interner::default();` & `Interner::with_capacity;`

Requesting for review

PR #4724, opened 2026-02-25
docs(interner): add code examples to boa_interner public API

### LINKED ISSUE #2337: Add code examples for `boa_interner`
The `boa_interner` crate has an `allow(rustdoc::missing_doc_code_examples)` on its root. Ideally, this crate should have some usage snippets on every `pub` function.

KaustubhOG:
Closes #2337

Added `# Examples` doc sections to all 14 public functions in`core/interner/src/lib.rs`, covering `Interner` and `JSInternedStrRef`.

PR #3141 added unit tests but left the public API without doc examples.This PR completes the original intent of the issue.

Notable considerations:
- `is_empty()` example reflects that a new interner is never empty due to pre-loaded `COMMON_STRINGS_UTF8`
- `len()` uses `initial_len` pattern for the same reason
- `into_common()` defines a local `JsString` enum since no standard type implements both `From<&str>` and `From<&[u16]>`

All 14 doctests pass (`cargo +nightly test --doc`).


core/interner/src/lib.rs
```diff
@@ -258,7 +399,6 @@ impl Interner {
 
             let utf16_index = self.utf16_interner.intern(&utf16);
 
-            // Just to check everything is okay
```
jedel1043, line 261: Uhh why did you remove the comments?

KaustubhOG: Hi @jedel1043, sorry about that!

KaustubhOG: like i also removed the 285 to 287 then 292 and 357 , i thought maybe they causes the redundant or self-explanatory meaning , if that was wrong interpretation from my end i will solve it quickly just a quick question should i rollback all comments that removed or that specidic part 246 to 261

jedel1043: Rollback the comments related to the _why_. The ones that only say things like `Just to check everything is okay` can be removed. Also the `get_common` one can be removed, since it's incomplete.

KaustubhOG: yup just a minute i will get back to you

KaustubhOG: @jedel1043 i rollback the comments have a look

KaustubhOG:
> Looks good to me!

thanks @jedel1043 tbh i enjoyed this because it was my first PR i raised in boa

PR #4725, opened 2026-02-25
fix: preserve request body when cloning with options (#4686)

### LINKED ISSUE #4686: Bug: Request body is dropped when cloning with options

LakshyaBagani:
This Pull Request fixes/closes #4686.

It changes the following:

1 => Preserve original request body in RequestInit::into_request_builder instead of discarding it with _body during                                            into_parts()
2 => Fall back to the original body when RequestInit does not provide a new one via                                                                                                                         request_body.or(original_body).unwrap_or_default()                                                                    
3 => Add #[cfg(test)] constructor for RequestInit to enable Rust-level unit tests
4 => Add 8 new tests: exact issue reproduction (E2E with echo fetcher), body override (E2E), body preservation via Rust API,  method change with body retention, clone without options, and empty-body GET clone

### DIFF
--- a/core/runtime/src/fetch/request.rs
+++ b/core/runtime/src/fetch/request.rs
@@ -56,6 +56,21 @@ pub struct RequestInit {
 }
 
 impl RequestInit {
+    /// Create a new `RequestInit` from Rust types, for use in tests.
+    #[cfg(test)]
+    #[must_use]
+    pub fn new(
+        body: Option<String>,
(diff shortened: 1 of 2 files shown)


nekevss:
Closing, duplicate of #4687

PR #4730, opened 2026-02-25
[bug]: TimeoutJob silently dropped when deadline == now in SimpleJobExecutor

ashnaaseth2325-oss:
## **Summary**

This PR fixes a boundary-condition bug in `SimpleJobExecutor::run_jobs_async` (`core/engine/src/job.rs`) where a `TimeoutJob` whose deadline is exactly equal to `now` could be silently skipped and permanently lost.

The issue was caused by inconsistent boundary semantics between the loop break-check and the timeout extraction logic. As a result, the executor could exit the event loop with `Ok(())` while a ready timeout job remained unexecuted.

This affects the default executor used by `Context::default()`.

---

## **Steps to Reproduce**

1. Use a fixed or frozen clock (e.g., `FixedClock`).
2. Schedule a timeout with delay `0` (deadline == now).
3. Ensure no other promise or timeout jobs exist.
4. Call `run_jobs()`.

The executor exits successfully, but the timeout callback never fires.

Minimal reproduction pattern:

```rust

context.enqueue_job(Job::TimeoutJob(TimeoutJob::new(
    NativeJob::new(|_| Ok(JsValue::undefined())),
    0,
)));

context.run_jobs().expect("run_jobs failed");
// Job never executed when deadline == now
```

---

## **Impact**

* Silent loss of timeout callbacks (no panic, no error).
* `setTimeout(fn, 0)` may fail in deterministic/frozen-clock scenarios.
* Async chains depending on zero-delay timers may hang.
* Breaks correctness for mock-clock based testing.
* Affects all users of `SimpleJobExecutor` (default executor).

This is particularly problematic in deterministic execution environments and test harnesses.

---

## **Root Cause**

The break condition used a strict `>` comparison:

```rust
!self.timeout_jobs.borrow().iter().any(|(t, _)| &now > t)
```

This treats a job as ready only when `now > deadline`.

However, execution uses:

```rust
let mut jobs_to_keep = timeouts_borrow.split_off(&now);
```

`split_off(&now)` keeps keys `>= now`, meaning only keys `< now` are executed.

Both sides consistently exclude `deadline == now`.

When:

* `now == deadline`
* all other queues are empty

The loop breaks early, and the job is never executed.

---

## **Fix**

1. Align readiness semantics to treat `deadline <= now` as ready.
2. Change the break-check to use `>=`.
3. Explicitly extract the entry keyed exactly at `now` after `split_off`.

Updated logic:

```rust
// Break condition
!self.timeout_jobs.borrow().iter().any(|(t, _)| &now >= t)
```

```rust
let mut jobs_to_keep = timeouts_borrow.split_off(&now);
let exact_now = jobs_to_keep.remove(&now);

let mut jobs_to_run = mem::replace(&mut *timeouts_borrow, jobs_to_keep);

if let Some(jobs) = exact_now {
    jobs_to_run.insert(now, jobs);
}
```

Additionally, regression tests were added to cover:

* `deadline == now`
* `deadline < now`
* `deadline > now`

Using `FixedClock` to ensure deterministic behavior.

---

## **Result**

* Timeout jobs fire correctly when `deadline == now`.
* Zero-delay timeouts behave deterministically.
* No silent job loss.
* No behavioral change for future deadlines.
* Async chains relying on timers no longer risk hanging.

This restores correct deadline semantics and ensures the executor behaves predictably under fixed or mock clock scenarios.

### DIFF
--- a/core/engine/src/object/property_map.rs
+++ b/core/engine/src/object/property_map.rs
@@ -386,6 +386,81 @@ impl IndexedProperties {
         }
     }
 
+    /// Pushes a value to the end of the dense indexed properties.
+    ///
+    /// Returns `true` if the push succeeded (storage is dense), `false` if
+    /// the storage is sparse and the caller should fall back to the slow path.
+    ///
+    /// Handles type transitions: `DenseI32` → `DenseF64` → `DenseElement`.
+    pub(crate) fn push_dense(&mut self, value: &JsValue) -> bool {
+        match self {
+            Self::DenseI32(vec) => {
+                if let Some(i) = value.as_i32() {
+                    vec.push(i);
+                } else if let Some(n) = value.as_number() {
+                    let mut new_vec: ThinVec<f64> = vec.iter().copied().map(f64::from).collect();
+                    new_vec.push(n);
+                    *self = Self::DenseF64(new_vec);
+                } else {
+                    let mut new_vec: ThinVec<JsValue> =
+                        vec.iter().copied().map(JsValue::from).collect();
+                    new_vec.push(value.clone());
+                    *self = Self::DenseElement(new_vec);
+                }
+                true
+            }
+            Self::DenseF64(vec) => {
+                if let Some(n) = value.as_number() {
+                    vec.push(n);
+                } else {
+                    let mut new_vec: ThinVec<JsValue> =
+                        vec.iter().copied().map(JsValue::from).collect();
+                    new_vec.push(value.clone());
+                    *self = Self::DenseElement(new_vec);
+                }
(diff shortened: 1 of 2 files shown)


hansl:
Please make your PR on top of `main` and fix only that issue.

I would also appreciate if you could explain in your own words what you're trying to accomplish and why.

ashnaaseth2325-oss:
Hi @hansl , 
Thanks for the feedback! I’ll rebase on top of main, narrow the changes to only the deadline boundary issue, and open a new PR with a clearer explanation of the fix.

PR #4732, opened 2026-02-25
feat: Implement symbols as WeakMap keys (closes #4439)

### LINKED ISSUE #4439: Implement Symbols as WeakMap keys proposal
https://github.com/tc39/proposal-symbols-as-weakmap-keys

gluzandii:
This Pull Request fixes/closes #4439

Implements the [Symbols as WeakMap keys proposal](https://github.com/tc39/proposal-symbols-as-weakmap-keys)
(Stage 4), allowing unique symbols (along with well-known symbols, ie Symbol.iterator etc) to be used as keys in `WeakMap`.

### What I have changed.

- **`can_be_held_weakly`** — A helper that is used to check if can be used as a key in weak map. Returns true for objects and symbols that are not registered (ie Symbol.for).

- **`JsSymbol::as_weak`** — Added a method that returns a `Weak<RawJsSymbol>` reference,
  enabling the weak-keyed storage needed by `NativeWeakMap`.

- **`NativeWeakMap`** — Extended with a two `symbols` map
  (`HashMap<u64, (Weak<SymbolData>, JsValue)>`), and  (`HashMap<u64, JsSymbol>`) keyed by symbol hash. One map is for user defined Symbols (ie from Symbol()) and the other is for well-known symbols (ie Symbol.iterator etc). Both together are used for storing symbols as keys for WeakMap.

All five public WeakMap
  operations (`get`, `has`, `set`, `delete`, `getOrInsert`, `getOrInsertComputed`) now handle
  both object and symbol keys through dedicated helper methods. Dead entries are purged on
  every `insert` and `delete` to prevent unbounded growth.

- **`is_well_known`** — Present in JsSymbol. Used to check if the symbol is a well-known symbol like Symbol.iterator etc.

- **`is_registered`** — Present in JsSymbol. Used to check if the symbol is a registered symbol (ie from Symbol.for).

### DIFF
--- a/core/engine/src/builtins/weak_map/mod.rs
+++ b/core/engine/src/builtins/weak_map/mod.rs
@@ -7,13 +7,13 @@
 //! [spec]: https://tc39.es/ecma262/#sec-weakmap-objects
 //! [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap
 
-use std::{collections::HashMap, fmt, sync::Weak};
+use std::{cell::RefCell, collections::HashMap, fmt, sync::Weak};
 
 use crate::{
     Context, JsArgs, JsNativeError, JsResult, JsString, JsValue,
     builtins::{
         BuiltInBuilder, BuiltInConstructor, BuiltInObject, IntrinsicObject,
-        map::add_entries_from_iterable, symbol,
+        map::add_entries_from_iterable,
     },
     context::intrinsics::{Intrinsics, StandardConstructor, StandardConstructors},
     js_string,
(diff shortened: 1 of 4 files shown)


jedel1043:
I recommend you to use the spec as the true source of truth for the implementation.

According to the note in https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-canbeheldweakly, well-known symbols are suitable for being keys in a weakmap, even if their values will never get collected.

gluzandii:
> I recommend you to use the spec as the true source of truth for the implementation.
> 
> According to the note in https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-canbeheldweakly, well-known symbols are suitable for being keys in a weakmap, even if their values will never get collected.

@jedel1043 Well-known symbols are now allowed as keys for WeakMap.

core/engine/src/builtins/weak_map/mod.rs
```diff
@@ -15,15 +17,135 @@ use crate::{
...
+// The `Weak<RawJsSymbol>` keys are Arc-based and contain no GC pointers.
+unsafe impl Trace for NativeWeakMap {
+    custom_trace!(this, mark, {
+        mark(&this.objects);
+
+        this.symbols.values().for_each(|(weak, value)| {
+            if weak.upgrade().is_some() {
+                mark(value);
+            }
+        });
```
jedel1043: This is UB.

Since a `Gc` can [access its inner `GcBox` allocation on `Drop`](https://github.com/boa-dev/boa/blob/e6114cf4fb094ce152a3277a20476e3dfb413693/core/gc/src/pointers/gc.rs#L320-L328), not tracing through the value could cause the Gc to deallocate the inner `GcBox`, making `Gc` access unallocated memory at drop.

jedel1043:
I'm gonna close this because we should at least add support for symbols in our GC for this to be supported properly. We can revisit this implementation after we address that.

PR #4734, opened 2026-02-26
feat(cli): support reading JavaScript from piped stdin

Nakshatra480:
## Summary

This PR adds support for reading JavaScript from piped stdin in the CLI.

Currently, when no file or `-e` expression is provided, Boa always drops into the interactive REPL — even if stdin is being piped from another command. This means common workflows like these don't work:

    echo 'console.log("hello")' | boa
    cat script.js | boa
    boa < script.js

## Changes

Added a check using `std::io::IsTerminal` in [main()](cci:1://file:///Users/nakshatrasharma/Desktop/boa/cli/src/main.rs:403:0-487:1) to detect when stdin is not a terminal. When piped input is detected, the CLI reads the full input and evaluates it as a script instead of launching the REPL. The normal interactive REPL behavior is completely unchanged.

## How it works

The detection happens after the existing file and `-e` checks, so the priority order is:

1. File arguments → execute files
2. `-e` expression → evaluate expression
3. Piped stdin (not a terminal) → read and evaluate stdin
4. Otherwise → launch the interactive REPL

## Testing

    # Simple expression
    echo '1 + 2' | boa
    # Output: 3

    # console.log
    echo 'console.log("hello")' | boa
    # Output: hello

    # Multi-line script
    printf 'let x = 10;\nlet y = 20;\nconsole.log(x + y);' | boa
    # Output: 30

    # File redirection
    boa < script.js


Nakshatra480:
@nekevss i have implemented the feature that i discussed in the community channel, can you review this PR.
Thanks for your time.

Nakshatra480:
Hey @nekevss @jedel1043 just following up here 🙂

These changes makes boa run JavaScript piped to stdin (for example echo '1+1' | boa or boa < script.js) instead of launching the REPL.

It detects non-terminal stdin with std::io::IsTerminal (after the existing file and -e checks), reads the piped input and evaluates it as a script, REPL stays the same when stdin is a terminal.

Would really appreciate a review. I’m happy to make any tweaks.
Thanks for your time

cli/src/main.rs
```diff
@@ -454,6 +454,18 @@ fn main() -> Result<()> {
...
         evaluate_expr(expr, &args, &mut context, &printer)?;
         return Ok(());
+    } else if !io::stdin().is_terminal() {
+        let mut input = String::new();
+        io::stdin()
+            .read_to_string(&mut input)
+            .wrap_err("could not read from stdin")?;
+        if !input.is_empty() {
+            evaluate_expr(&input, &args, &mut context, &printer)?;
+        }
```
jedel1043:
```suggestion
        if input.is_empty() {
            return Ok(());
        }
        evaluate_expr(&input, &args, &mut context, &printer)?;
```
Slightly better

jedel1043 (changes requested):

cli/src/main.rs
Nakshatra480: I have done the suggested changes.
Thanks for the feedback and your time

nekevss (changes requested):
Super cool addition! I actually really like it.

Can you add some examples of this functionality to the README.md for `boa_cli` so that this functionality is displayed somewhere.

cli/README.md
```diff
@@ -32,6 +32,14 @@ Or if you'd like to use Boa's REPL, simply type:
     boa
 ```
 
+You can also pipe JavaScript into Boa:
+
+```shell
+    echo 'console.log(1 + 2)' | boa
+    cat script.js | boa
+    boa < script.js
```
jedel1043, line 40: Oh cool that you changed the docs!

Nakshatra480: Yes i have changed the docs. 
Thanks for the feedback

jedel1043 (approved):
Looks great!

nekevss (approved):
Looks great! Thanks!

PR #4737, opened 2026-02-26
Remove optional-chain-first-expression.js from test262 ignore list

Anayo-Anyafulu:
This Pull Request is a maintenance cleanup (no associated issue).

It changes the following:
- Removes `optional-chain-first-expression.js` from the test262 ignore list
- This test was previously ignored due to a panic in Boa's bytecode emitter
- Running it against current main no longer causes a panic, so the ignore entry is no longer needed


jedel1043 (approved):
Oh cool, one less test to worry about :)

PR #4738, opened 2026-02-26
feat(builtins): implement JsWeakMap safe wrapper

KaustubhOG:
Implements `JsWeakMap` as a safe Rust wrapper around the ECMAScript `WeakMap` builtin, as part of #2098.

## Changes
- `core/engine/src/object/builtins/jsweakmap.rs` — new `JsWeakMap` wrapper with all methods
- `core/engine/src/object/builtins/mod.rs` — registered and re-exported `JsWeakMap`
- `examples/src/bin/jsweakmap.rs` — usage example

## Methods implemented
- `new` — creates an empty WeakMap
- `get` — retrieves value by key
- `set` — inserts a key-value pair
- `has` — checks key existence
- `delete` — removes a key
- `get_or_insert` — upsert with default value
- `get_or_insert_computed` — upsert with callback
- `from_object` — converts `JsObject` to `JsWeakMap`

Follows the same pattern as `JsMap` and `JsSet`.

Closes part of #2098

### DIFF
--- a/core/engine/src/object/builtins/jsweakmap.rs
+++ b/core/engine/src/object/builtins/jsweakmap.rs
@@ -5,13 +5,11 @@ use boa_gc::{Finalize, Trace};
 
 use crate::{
     Context, JsResult, JsValue,
-    builtins::weak_map::WeakMap,
+    builtins::weak_map::{NativeWeakMap, WeakMap},
     error::JsNativeError,
-    object::{ErasedVTableObject, JsObject},
+    object::JsObject,
     value::TryFromJs,
 };
-
-type NativeWeakMap = boa_gc::WeakMap<ErasedVTableObject, JsValue>;
 
 /// `JsWeakMap` provides a wrapper for Boa's implementation of the ECMAScript `WeakMap` object.
 #[derive(Debug, Clone, Trace, Finalize)]
--- a/core/engine/src/builtins/weak_map/mod.rs
+++ b/core/engine/src/builtins/weak_map/mod.rs
@@ -23,7 +23,7 @@ use crate::{
 };
 use boa_gc::{Finalize, Trace};
 
-type NativeWeakMap = boa_gc::WeakMap<ErasedVTableObject, JsValue>;
+pub(crate) type NativeWeakMap = boa_gc::WeakMap<ErasedVTableObject, JsValue>;
 
 #[derive(Debug, Trace, Finalize)]
 pub(crate) struct WeakMap;


core/engine/src/object/builtins/jsweakmap.rs
```diff
@@ -0,0 +1,159 @@
...
+            &self.inner.clone().into(),
+            &[key.clone().into(), default],
+            context,
+        )
+    }
+
+    /// Returns the existing value if the key exists; otherwise calls `callback(key)`,
+    /// inserts the result, and returns it.
+    ///
+    /// Same as JavaScript's `weakmap.getOrInsertComputed(key, callback)`.
```
jedel1043: You don't need to add all of these `Same as...` comments, the name implies that. What would be useful though is to link to the corresponding Mozilla docs explaining the method's API.

KaustubhOG: Hi @jedel1043, thank you for the feedback! I've removed all the Same as JavaScript's... comments and replaced them with MDN documentation links for every method. I've applied the same change to the JsWeakSet PR (https://github.com/boa-dev/boa/pull/4739) as well. Please let me know if there's anything else to improve!

core/engine/src/object/builtins/jsweakmap.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use crate::{
+    Context, JsResult, JsValue,
+    builtins::weak_map::WeakMap,
+    error::JsNativeError,
+    object::{ErasedVTableObject, JsObject},
+    value::TryFromJs,
+};
+
+type NativeWeakMap = boa_gc::WeakMap<ErasedVTableObject, JsValue>;
```
jedel1043: There's already a type alias for this in here:
https://github.com/boa-dev/boa/blob/c89f7e9d99c51be4d463ce2af84bd4936e3598e1/core/engine/src/builtins/weak_map/mod.rs#L26
so just make that `pub(crate)` and import it from there

KaustubhOG: Hi @jedel1043, removed the local NativeWeakMap type alias from jsweakmap.rs, made it pub(crate) in weak_map/mod.rs, and imported it from there. is it correct way to handle it or should it require other way around.

jedel1043: Nah that's the only change required.

jedel1043 (approved):
Looks great! Thanks!

core/engine/src/object/builtins/jsweakmap.rs
KaustubhOG: so should i do same changes in the weakset also ?

jedel1043: Yes please!

KaustubhOG: just a minute sirr , getting there

KaustubhOG: i made changes does it reflect in #4739  properly jedel

jedel1043: Yep, already approved that one. Thanks!

KaustubhOG:
Hi @jedel1043! Thank you for the quick reviews and guidance on  #4738   really appreciated the clear feedback.

PR #4739, opened 2026-02-26
feat(builtins): implement JsWeakSet safe wrapper

KaustubhOG:
Implements `JsWeakSet` as a safe Rust wrapper around the ECMAScript `WeakSet` builtin, as part of #2098.

## Changes
- `core/engine/src/object/builtins/jsweakset.rs` — new `JsWeakSet` wrapper with all methods
- `core/engine/src/object/builtins/mod.rs` — registered and re-exported `JsWeakSet`
- `examples/src/bin/jsweakset.rs` — usage example

## Methods implemented
- `new` — creates an empty WeakSet
- `add` — adds an object to the set
- `has` — checks if an object exists
- `delete` — removes an object
- `from_object` — converts `JsObject` to `JsWeakSet`

Follows the same pattern as `JsWeakMap` (#4738) and `JsSet`.

Closes part of #2098

### DIFF
--- a/core/engine/src/object/builtins/mod.rs
+++ b/core/engine/src/object/builtins/mod.rs
@@ -17,8 +17,8 @@ mod jsset;
 mod jsset_iterator;
 mod jssharedarraybuffer;
 mod jstypedarray;
-mod jsweakset;
 mod jsweakmap;
+mod jsweakset;
 
 pub use jsarray::*;
 pub use jsarraybuffer::*;
@@ -35,5 +35,5 @@ pub use jsset::*;
 pub use jsset_iterator::*;
 pub use jssharedarraybuffer::*;
 pub use jstypedarray::*;
-pub use jsweakset::JsWeakSet;
 pub use jsweakmap::JsWeakMap;
+pub use jsweakset::JsWeakSet;


core/engine/src/object/builtins/jsweakset.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use crate::{
+    Context, JsResult, JsValue,
+    builtins::weak_set::WeakSet,
+    error::JsNativeError,
+    object::{ErasedVTableObject, JsObject},
+    value::TryFromJs,
+};
+
+type NativeWeakSet = boa_gc::WeakMap<ErasedVTableObject, ()>;
```
jedel1043: Same comment as the `JsWeakMap`. Use the type alias from the builtins module
https://github.com/boa-dev/boa/blob/c89f7e9d99c51be4d463ce2af84bd4936e3598e1/core/engine/src/builtins/weak_set/mod.rs#L25

KaustubhOG:
> LGTM!

hey i made some changes in #4724 as you suggested to  me is it loooks fine

jedel1043:
Looks fine, just needs a `cargo fmt` run

KaustubhOG:
> Looks fine, just needs a `cargo fmt` run

Yup ,while  solving the conflicts in Jsweakmap and and weakset ,Ithe web one we uses I think it didn't do the cargo fmt , let me do Manually

KaustubhOG:
@jedel1043 did the jsgeneratorfunction and the Asyncgenerator isn't checked on original pr #2098 is it currently outof scope(if they are in scope or not yet solved I think it can be fixed with the same process if i am not mistaken ,if that's the case I loved to raise pr for it), also thanks for quick guidence in this pr. I learned alot

PR #4740, opened 2026-02-26
feat(vm): implement polymorphic inline cache (PIC) for property access

### LINKED ISSUE #4656: Proposal: Polymorphic Inline Cache (PIC) for Property Access

akash-R-A-J:
This Pull Request fixes/closes #4656.

Currently Boa uses a **monomorphic inline cache** for property access sites, storing a single `(Shape → Slot)` mapping. When the shape does not match, the cache is reset, which causes frequent cache invalidation when accessing the same property across objects with different shapes.

This PR upgrades the inline cache to a **polymorphic inline cache (PIC)** to better support polymorphic property access patterns commonly seen in JavaScript.

It changes the following:

- Replace the monomorphic inline cache with a **polymorphic inline cache (PIC)** that stores multiple `(Shape → Slot)` mappings per access site using a small fixed-capacity cache (`PIC_CAPACITY = 4`).
- Add support for a **megamorphic state**, where the cache is disabled after observing too many shapes to avoid unnecessary overhead.
- Update inline cache lookup and insertion logic to support multiple cached shapes and opportunistic cleanup of stale weak shape references.
- Update disassembly output to correctly display the cached shape information for PIC.
- Add tests covering **polymorphic** and **megamorphic** inline cache behavior.

This change reduces inline cache thrashing for polymorphic property access patterns and aligns Boa's inline cache design more closely with modern JavaScript engines.

### DIFF
--- a/core/engine/src/vm/code_block.rs
+++ b/core/engine/src/vm/code_block.rs
@@ -634,7 +634,7 @@ impl CodeBlock {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
                 format!(
                     "dst:{dst}, value:{value}, shape:0x{:x}]",
-                    ic.shape.borrow().to_addr_usize(),
+                    ic.first_shape_addr(),
                 )
             }
             Instruction::GetPropertyByName {
@@ -646,7 +646,7 @@ impl CodeBlock {
                 format!(
                     "dst:{dst}, value:{value}, ic:[name:{}, shape:0x{:x}]",
                     ic.name.to_std_string_escaped(),
-                    ic.shape.borrow().to_addr_usize(),
+                    ic.first_shape_addr(),
                 )
             }
             Instruction::GetPropertyByNameWithThis {
@@ -659,7 +659,7 @@ impl CodeBlock {
                 format!(
                     "dst:{dst}, receiver:{receiver}, value:{value}, ic:[name:{}, shape:0x{:x}]",
                     ic.name.to_std_string_escaped(),
-                    ic.shape.borrow().to_addr_usize(),
+                    ic.first_shape_addr(),
                 )
             }
             Instruction::SetPropertyByName {
@@ -670,7 +670,7 @@ impl CodeBlock {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
                 format!(
                     "object:{object}, value:{value}, ic:shape:0x{:x}",
-                    ic.shape.borrow().to_addr_usize(),
+                    ic.first_shape_addr(),
                 )
             }
             Instruction::SetPropertyByNameWithThis {
@@ -682,7 +682,7 @@ impl CodeBlock {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
                 format!(
                     "object:{object}, receiver:{receiver}, value:{value}, ic:shape:0x{:x}",
-                    ic.shape.borrow().to_addr_usize(),
+                    ic.first_shape_addr(),
                 )
             }
             Instruction::GetPropertyByValue {
--- a/core/engine/src/vm/code_block.rs
+++ b/core/engine/src/vm/code_block.rs
@@ -630,24 +630,14 @@ impl CodeBlock {
                 dst,
                 value,
                 ic_index,
-            } => {
-                let ic = &self.ic[u32::from(*ic_index) as usize];
-                format!(
-                    "dst:{dst}, value:{value}, ic:[shapes:{}]",
-                    ic.shapes_display(),
-                )
             }
-            Instruction::GetPropertyByName {
+            | Instruction::GetPropertyByName {
                 dst,
                 value,
                 ic_index,
             } => {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
-                format!(
-                    "dst:{dst}, value:{value}, ic:[name:{}, shapes:{}]",
-                    ic.name.to_std_string_escaped(),
-                    ic.shapes_display(),
-                )
+                format!("dst:{dst}, value:{value}, ic:{ic}",)
             }
             Instruction::GetPropertyByNameWithThis {
                 dst,
@@ -656,11 +646,7 @@ impl CodeBlock {
                 ic_index,
             } => {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
-                format!(
-                    "dst:{dst}, receiver:{receiver}, value:{value}, ic:[name:{}, shapes:{}]",
-                    ic.name.to_std_string_escaped(),
-                    ic.shapes_display(),
-                )
+                format!("dst:{dst}, receiver:{receiver}, value:{value}, ic:{ic}",)
             }
             Instruction::SetPropertyByName {
                 value,
@@ -668,10 +654,7 @@ impl CodeBlock {
                 ic_index,
             } => {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
-                format!(
-                    "object:{object}, value:{value}, ic:[shapes:{}]",
-                    ic.shapes_display(),
-                )
+                format!("object:{object}, value:{value}, ic:{ic}",)
             }
             Instruction::SetPropertyByNameWithThis {
                 value,
@@ -680,10 +663,7 @@ impl CodeBlock {
                 ic_index,
             } => {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
-                format!(
-                    "object:{object}, receiver:{receiver}, value:{value}, ic:[shapes:{}]",
-                    ic.shapes_display(),
-                )
+                format!("object:{object}, receiver:{receiver}, value:{value}, ic:{ic}")
             }
             Instruction::GetPropertyByValue {
                 dst,
--- a/core/engine/src/vm/code_block.rs
+++ b/core/engine/src/vm/code_block.rs
@@ -633,8 +633,8 @@ impl CodeBlock {
             } => {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
                 format!(
-                    "dst:{dst}, value:{value}, shape:0x{:x}]",
-                    ic.first_shape_addr(),
+                    "dst:{dst}, value:{value}, ic:[shapes:{}]",
+                    ic.shapes_display(),
                 )
             }
             Instruction::GetPropertyByName {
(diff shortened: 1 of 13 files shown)


akash-R-A-J:
hii @jedel1043, could you please review this pr? happy to make any changes or address feedback. thanks!

jedel1043:
Thanks!

Before reviewing anything, can you run https://github.com/boa-dev/data/blob/main/bench/bench-v8/combined.js? This should report benchmark results for your current PR's head commit, and you can compare that with a run against the main branch

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -11,51 +12,125 @@ use crate::{
...
+        while i < entries.0.len() {
+            if let Some(upgraded) = entries.0[i].shape.upgrade() {
+                if upgraded.to_addr_usize() == shape_addr {
+                    result = Some((upgraded, entries.0[i].slot));
+                    break;
+                }
+                i += 1;
+            } else {
+                // Opportunistically clean up stale weak shapes.
+                entries.0.remove(i);
```
jedel1043: I'd use [`swap_remove`](https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html#method.swap_remove) here, which avoids having to move all elements after `i`.

core/engine/src/vm/code_block.rs
```diff
@@ -634,7 +634,7 @@ impl CodeBlock {
                 let ic = &self.ic[u32::from(*ic_index) as usize];
                 format!(
                     "dst:{dst}, value:{value}, shape:0x{:x}]",
-                    ic.shape.borrow().to_addr_usize(),
+                    ic.first_shape_addr(),
```
jedel1043: IMO these should print all the shapes in the `ic`. Seems useful to known which shapes are cached.

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -11,51 +12,125 @@ use crate::{
...
+            entries.0.clear();
+        }
     }
 
-    /// Returns true, if the [`InlineCache`]'s shape matches with the given shape.
+    /// Returns the cached `(Shape, Slot)` if a matching shape exists in the inline cache.
     ///
     /// Otherwise we reset the internal weak reference to [`WeakShape::None`],
     /// so it can be deallocated by the GC.
     pub(crate) fn match_or_reset(&self, shape: &Shape) -> Option<(Shape, Slot)> {
```
jedel1043: You should change the method name, because it doesn't reflect what's really happening anymore. `get` should probably be fine. Also you need to update the doc comment with the new behaviour; we don't reset the internal weak reference anymore.

akash-R-A-J:
hey @jedel1043, I tried running the V8 combined benchmark (`combined.js`) on both the `main` branch and this PR's branch. 

Both runs currently overflow the stack on Windows when executing:
- target/release/boa.exe combined.js
- target/debug/boa.exe combined.js

I also tried increasing the stack size (RUST_MIN_STACK=33554432) and rerunning the benchmark, but it still crashes with the same stack overflow.

If there is a recommended way to run this benchmark on Windows, I'd be happy to try again and report the results.

akash-R-A-J:
hey @jedel1043, here's the report after running both branches on wsl2.

After running on main branch: 

boa_,main

After running on this PR's branch:

boa_pic

jedel1043:
Wow, Splay got a very _biiiiig_ perf boost

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -1,6 +1,7 @@
+use arrayvec::ArrayVec;
 use std::cell::Cell;
 
-use boa_gc::GcRefCell;
+use boa_gc::{GcRefCell, Trace as BoaTrace, custom_trace};
```
jedel1043: Why import `Trace` as `BoaTrace`?

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -11,51 +12,125 @@ use crate::{
...
 #[derive(Clone, Debug, Trace, Finalize)]
+pub(crate) struct PicEntry {
+    /// A weak reference is kept to the shape to avoid the shape preventing deallocation.
+    pub(crate) shape: WeakShape,
+    #[unsafe_ignore_trace]
+    pub(crate) slot: Slot,
+}
+
+#[derive(Clone, Debug, Finalize)]
+pub(crate) struct PicEntries(pub(crate) ArrayVec<PicEntry, PIC_CAPACITY>);
```
jedel1043: Maybe it would be better to implement `Trace` on `ArrayVec` in the `boa_gc` crate? Should be basically the same as the `custom_trace!` you implemented.

jedel1043 (changes requested):

core/engine/src/vm/inline_cache/mod.rs
akash-R-A-J: The alias was introduced during development while wiring up the `Trace` implementation to avoid confusion with other imports. After the refactor it was no longer necessary, so it has been removed in the latest update.

core/engine/src/vm/inline_cache/mod.rs
akash-R-A-J: That makes sense. Initially, I used `custom_trace!` on the PIC container to keep the changes local, since `ArrayVec` did not implement `Trace`. I’ve now implemented Trace for `ArrayVec` in `boa_gc` as suggested and updated the PIC implementation to use that instead.

akash-R-A-J:
Hey @jedel1043, thanks for the review and the feedback! After incorporating the suggested changes, the benchmark score improved further (179 → 192 → 206).  The largest improvement, as you noted earlier, continues to appear in the Splay benchmark.

Latest benchmark on this PR branch: 

latest_report

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -11,51 +12,116 @@ use crate::{
...
+                return;
+            }
+        }
+
+        // Add a new entry if there's space.
+        if entries.len() < PIC_CAPACITY {
+            entries.push(PicEntry {
+                shape: shape.into(),
+                slot,
+            });
```
jedel1043: Here you should use https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html#method.try_push

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -11,51 +12,116 @@ use crate::{
...
+            return;
+        }
 
-    pub(crate) fn slot(&self) -> Slot {
-        self.slot.get()
+        let mut entries = self.entries.borrow_mut();
+        let shape_addr = shape.to_addr_usize();
+
+        // If the shape already exists, update its slot.
+        // This handles cases where property transitions preserve the shape but change the slot.
```
jedel1043: Is this possible? IIRC any property transition _will_ change the shape, so it should be impossible to have two shapes that are the same without them having the same slots.

core/engine/src/vm/inline_cache/mod.rs
```diff
@@ -11,51 +12,116 @@ use crate::{
 #[cfg(test)]
 mod tests;
 
+pub(crate) const PIC_CAPACITY: usize = 4;
```
jedel1043, line 16: thought: we might have to experiment with this value in the future. Increasing it _might_ increase performance, but it might also cause more cache misses which could decrease performance.

jedel1043 (changes requested):

core/engine/src/vm/inline_cache/mod.rs
akash-R-A-J: Thanks for pointing this out! I've updated this to use `ArrayVec::try_push()`.

core/engine/src/vm/inline_cache/mod.rs
akash-R-A-J: You're right. `set()` is only called after a `get()` miss, so the shape cannot already exist in the cache. I originally added that check as a defensive guard, but since property transitions create a new shape and `set()` is only reached after a cache miss, it’s effectively dead code. I've removed that logic.

core/engine/src/vm/inline_cache/mod.rs
akash-R-A-J: Makes sense. We can experiment with different values in benchmarks later to see how it affects cache hit rates and overall performance.

jedel1043 (approved):
Nice job! We might still have to experiment in the future with evicting caches instead of making the whole cache inaccessible, but this seems like a very good starting point.

zhuzhu81998:
Note: ci fails relating to dead code still need to be fixed

akash-R-A-J:
@jedel1043 thanks for the review and feedback!

I’ve fixed the dead_code and lint errors that were causing the CI failures. One of the warnings appeared to be unrelated to the PIC changes, so I limited the fixes to the parts directly affected by this PR to avoid modifying unrelated code.

I also noted your suggestion about experimenting with cache eviction strategies and tuning `PIC_CAPACITY`. I’ll try some experiments around those and open a follow-up PR if I can get measurable improvements.

For reference, the latest local benchmark on the V8 combined.js suite showed roughly a ~3% improvement (`boa-pic vs main`) with the biggest gains appearing in the Splay benchmark.

final_perf

Also, thanks for the note @zhuzhu81998!

PR #4741, opened 2026-02-26
fix: Clear async_parent_modules after async evaluation

ashnaaseth2325-oss:
## Summary

`SourceTextModule::async_parent_modules` was never cleared after async evaluation completed.

This vector stores strong `Gc<ModuleRepr>` references to parent modules during evaluation. Since async dependency modules are cached inside `SimpleModuleLoader::module_map`, those references permanently retain the full parent module object graphs.

Over repeated evaluations, this causes **linear, unbounded heap growth**.

---

## Steps to Reproduce

1. Create a module with Top-Level Await.
2. Repeatedly evaluate another module importing it using a shared `SimpleModuleLoader`.
3. Observe heap growth over iterations.

Each run appends a new parent into:

```rust
required_module
    .async_parent_modules
    .borrow_mut()
    .push(module_self.clone());
```

The list is read during completion but never cleared:

```rust
module_src.gather_available_ancestors(&mut ancestors);
// no clear()
```

---

## Impact

* Affects long-running embedded Boa runtimes
* Leaks one full module graph per evaluation
* Heap grows linearly
* Leads to increased GC pressure and potential OOM

Triggered by any Top-Level Await dependency.

---

## Fix

Clear the vector after ancestor gathering in both success and rejection paths:

```rust
module_src.async_parent_modules.borrow_mut().clear();
```

Minimal change (1–2 lines), no API or semantic impact.

---

## Result

* Parent modules become collectible
* Heap stabilizes under repeated evaluation
* No change in JS behavior
* Prevents production memory leak in async module execution

### DIFF
--- a/core/engine/src/module/source.rs
+++ b/core/engine/src/module/source.rs
@@ -1996,6 +1996,11 @@ fn async_module_execution_fulfilled(module: &Module, context: &mut Context) -> J
     // 9. Perform GatherAvailableAncestors(module, execList).
     module_src.gather_available_ancestors(&mut ancestors);
 
+    // The async parent list has served its purpose now that the module is `Evaluated`.
+    // Holding it any longer permanently retains the Gc graphs of all parent modules
+    // (via strong Gc<ModuleRepr> references) for the lifetime of the cached dependency.
+    module_src.async_parent_modules.borrow_mut().clear();
+
     // 11. Assert: All elements of sortedExecList have their [[AsyncEvaluation]] field set to true, [[PendingAsyncDependencies]] field set to 0, and [[EvaluationError]] field set to empty.
     let mut ancestors = ancestors.into_iter().collect::<Vec<_>>();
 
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hi @jedel1043 @hansl, 
This PR fixes a GC retention issue by clearing async_parent_modules once async evaluation completes. No spec or conformance behavior changes ; Test262 results remain identical.
Thanks!

core/engine/src/module/source.rs
```diff
@@ -1996,6 +1996,11 @@ fn async_module_execution_fulfilled(module: &Module, context: &mut Context) -> J
     // 9. Perform GatherAvailableAncestors(module, execList).
     module_src.gather_available_ancestors(&mut ancestors);
 
+    // The async parent list has served its purpose now that the module is `Evaluated`.
+    // Holding it any longer permanently retains the Gc graphs of all parent modules
+    // (via strong Gc<ModuleRepr> references) for the lifetime of the cached dependency.
+    module_src.async_parent_modules.borrow_mut().clear();
```
jedel1043: Hmm, not a big fan of just clearing the `async_parent_modules` in a random place.

Can we just take the vec on the place it is used instead?

ashnaaseth2325-oss:
Hi @jedel1043 @hansl 
Refactored as suggested! The vec is now consumed at the usage site via take() , removing the need for a separate clear. Happy to adjust further if you’d like any refinements.

jedel1043 (approved):
Cool! I think this can be further improved by implementing `Trace` on `Cell<Vec<T>>` and pushing/taking the whole vec at once, but it is a very small optimization that can be done later.

ashnaaseth2325-oss:
Thanks! That makes sense. I can look into the Trace + full vec push/take optimization in a follow-up PR.

PR #4744, opened 2026-02-26
Implement URLSearchParams class and URL.searchParams

### LINKED ISSUE #4743: Implement URLSearchParams class
### Summary

### Expected behavior

```js
params.get("foo"); // "1"
params.append("baz", "3");

```

### Actual behavior

```
```

### References

apoorvdarshan:
## Summary
- Implement the `URLSearchParams` Web API class with full method support: `append`, `delete`, `get`, `getAll`, `has`, `set`, `sort`, `size`, `toString`, `forEach`, `entries`, `keys`, `values`
- Wire up `URL.searchParams` getter to return a `URLSearchParams` instance parsed from the URL's query string (previously threw "not implemented")
- Registered alongside `URL` so it's available when the `url` feature is enabled

Fixes #4743

## Test plan
- [x] `url_search_params_basic` — constructor, get, size, toString
- [x] `url_search_params_constructor_leading_question_mark` — strips leading `?`
- [x] `url_search_params_append_delete` — append/delete operations
- [x] `url_search_params_has` — has with and without value filter
- [x] `url_search_params_set` — set replaces first, removes duplicates, adds new
- [x] `url_search_params_sort` — sorts by name
- [x] `url_search_params_get_all` — returns all values for a name
- [x] `url_search_params_for_each` — callback iteration
- [x] `url_search_params_from_url` — URL.searchParams returns working params
- [x] `url_search_params_encoding` — proper URL encoding of special characters
- [x] All 4 existing URL tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/runtime/src/url/search_params.rs
+++ b/core/runtime/src/url/search_params.rs
@@ -88,13 +88,13 @@ impl UrlSearchParams {
     /// Returns all values associated with the given name.
     fn get_all(&self, name: Convert<String>, context: &mut Context) -> JsValue {
         let name = &name.0;
-        let values: Vec<JsValue> = self
-            .pairs
-            .iter()
-            .filter(|(n, _)| n == name)
-            .map(|(_, v)| JsValue::from(JsString::from(v.as_str())))
-            .collect();
-        let arr = boa_engine::object::builtins::JsArray::from_iter(values, context);
+        let arr = boa_engine::object::builtins::JsArray::from_iter(
+            self.pairs
+                .iter()
+                .filter(|(n, _)| n == name)
+                .map(|(_, v)| JsValue::from(JsString::from(v.as_str()))),
+            context,
+        );
         arr.into()
     }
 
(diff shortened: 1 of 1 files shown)


memmokx:
You could avoid unnecessary memory allocations by removing the `collect` before calling `JsArray::from_iter` for these methods: `values`, `keys`, `get_all` (`impl IntoIterator` accepts `Iterator`'s).

apoorvdarshan:
@memmokx Thanks for the suggestion! Removed the `.collect()` for `get_all`, `keys`, and `values` in 922f20ae. The `entries` method still needs `.collect()` because the inner `JsArray::from_iter` call borrows `context` mutably inside the closure, which conflicts with the outer `from_iter` also needing `context`.

jedel1043:
If you're gonna generate a PR with an LLM, at least make the effort of reading the spec to check if everything is correctly implemented per the spec.

PR #4748, opened 2026-02-27
Fix typeof on TDZ let/const variables returning "undefined" instead of throwing ReferenceError

### LINKED ISSUE #4746: typeof on TDZ variable returns "undefined" instead of throwing ReferenceError
## Summary
`typeof` on a Temporal Dead Zone (TDZ) `let`/`const` variable returns `"undefined"` instead of throwing `ReferenceError`, violating the ECMAScript spec.

### Reproduction

```js
try { console.log(typeof x); } catch(e) { console.log(e.constructor.name); }
let x = 1;
// Boa:      "undefined"
// Expected: ReferenceError (V8, SpiderMonkey, JSC all throw)
```

`typeof` should only return `"undefined"` silently for **undeclared** variables ([spec §13.5.3.1, step 2a](https://tc39.es/ecma262/#sec-typeof-operator)). For TDZ bindings, `GetValue` calls `GetBindingValue` which must throw ([spec §9.1.1.1.6](https://tc39.es/ecma262/#sec-declarative-environment-records-getbindingvalue-n-s)).

### Root Cause

In `bytecompiler/expression/unary.rs`, `typeof <identifier>` always emits `GetNameOrUndefined`, which returns `undefined` for both undeclared **and** uninitialized bindings. It should use `GetName` (which throws) for lexical bindings in TDZ, and `GetNameOrUndefined` only for undeclared/var bindings.

### Environment

- **Boa version:** latest main branch
- **OS:** macOS

Nakshatra480:
Closes #4746 

`typeof` on a `let`/`const` variable in the Temporal Dead Zone was incorrectly returning `"undefined"` instead of throwing a `ReferenceError`. The issue was in the bytecompiler - `typeof <identifier>` always emitted the `GetNameOrUndefined` opcode, which silently returns undefined for both undeclared and uninitialized bindings.

Per the spec ([§13.5.3.1](https://tc39.es/ecma262/#sec-typeof-operator)), `typeof` should only suppress the `ReferenceError` for unresolvable references (i.e., undeclared variables). For bindings that exist but are uninitialized (TDZ), `GetValue` → `GetBindingValue` must still throw ([§9.1.1.1.6](https://tc39.es/ecma262/#sec-declarative-environment-records-getbindingvalue-n-s)).

The fix checks `binding.is_lexical()` when compiling `typeof <identifier>` to pick the right opcode:
- Lexical bindings (let/const): emit `GetName` - throws `ReferenceError` if in TDZ
- Non-lexical bindings (var/undeclared): emit `GetNameOrUndefined` - returns `"undefined"` silently

Before:

```js
try { console.log(typeof x); } catch(e) { console.log(e.constructor.name); }
let x = 1;
// Boa output: "undefined"
```

After:

```js
try { console.log(typeof x); } catch(e) { console.log(e.constructor.name); }
let x = 1;
// Boa output: "ReferenceError" (matches V8, SpiderMonkey, JSC)
```

All existing tests pass.

### DIFF
--- a/core/engine/src/tests/operators.rs
+++ b/core/engine/src/tests/operators.rs
@@ -211,18 +211,42 @@ fn typeofs() {
 }
 
 #[test]
-fn typeof_tdz_throws() {
-    run_test_actions([TestAction::assert_native_error(
-        indoc! {r#"
-            function f() {
-                typeof x;
-                let x = 1;
-            }
-            f();
-        "#},
-        JsNativeErrorKind::Reference,
-        "access of uninitialized binding",
-    )]);
+fn typeof_edge_cases() {
+    run_test_actions([
+        TestAction::assert_eq("typeof notDeclared", js_str!("undefined")),
+        TestAction::assert_eq(
+            indoc! {r#"
+                function f() {
+                    return typeof x;
+                    var x = 1;
+                }
+                f();
+            "#},
+            js_str!("undefined"),
+        ),
+        TestAction::assert_native_error(
+            indoc! {r#"
+                function f() {
+                    return typeof x;
+                    let x = 1;
+                }
+                f();
+            "#},
+            JsNativeErrorKind::Reference,
+            "access of uninitialized binding",
+        ),
+        TestAction::assert_native_error(
+            indoc! {r#"
+                function f() {
+                    return typeof x;
+                    const x = 1;
+                }
+                f();
+            "#},
+            JsNativeErrorKind::Reference,
+            "access of uninitialized binding",
+        ),
+    ]);
 }
 
 #[test]


Nakshatra480:
Hey @jedel1043 @nekevss  just following up on this PR.
This fixes typeof on TDZ let/const so it throws ReferenceError instead of returning "undefined".
The bytecode compiler now emits GetName for lexical bindings and GetNameOrUndefined for non-lexical ones, matching the spec.
All tests pass and Codecov reports the modified lines are covered.
Would appreciate a review, happy to make any tweaks.

jedel1043 (commented):
Nice work!

Since it seems like test262 didn't catch this bug, can you add a regression test for it?

Nakshatra480:
@jedel1043 I’ve added a focused regression test covering typeof in TDZ scenarios (let/const), plus undeclared/var‑hoisted cases to ensure behavior stays correct.
Thanks for the feedback

PR #4749, opened 2026-02-27
fix(runtime): ignore recurring TimeoutJob in termination guard to prevent setInterval(0) infinite spin

ashnaaseth2325-oss:
## **Summary**

This PR fixes an infinite event loop bug in the async job executor where recurring `TimeoutJob`s (created by `setInterval`) were incorrectly considered when deciding whether the event loop should terminate.

The `TimeoutJob` struct already contains a `recurring` flag and an `is_recurring()` accessor:

```rust
pub fn is_recurring(&self) -> bool {
    self.recurring
}
```

However, this flag was never used inside `SimpleJobExecutor::run_jobs_async()` when evaluating the termination condition.

As a result, under the default `StdClock` (real wall clock), a zero-delay interval:

```js
setInterval(() => {}, 0);
```

would cause `context.run_jobs()` to spin forever at 100% CPU.

---

## **Root Cause**

In `run_jobs_async`, the termination guard checked whether any timeout jobs were past-due:

```rust
let no_timeout_jobs_to_run = {
    let now = context.borrow().clock().now();
    !self.timeout_jobs.borrow().iter().any(|(t, _)| &now > t)
};
```

This logic:

* Considered **all timeout jobs**
* Did **not exclude recurring jobs**
* Ignored the `recurring` flag entirely

For `setInterval(fn, 0)`:

1. The job is re-enqueued at `now + 0ms = now`
2. `StdClock::now()` advances by nanoseconds between loop iterations
3. The re-enqueued job appears immediately past-due
4. The termination condition never becomes true
5. The loop busy-spins forever

This bug was invisible in tests because they use `FixedClock`, which does not advance automatically.

---

## **Steps to Reproduce**

Using the default `ContextBuilder` (which uses `StdClock`):

```rust
context.eval(Source::from_bytes(
    r#"setInterval(() => {}, 0);"#
)).unwrap();

context.run_jobs().unwrap(); // Hangs forever (CPU spin)
```

Observed behavior:

* `run_jobs()` never returns
* CPU usage hits 100%
* Event loop never terminates

---

## **Impact**

* ❌ Infinite loop in `run_jobs()`
* ❌ 100% CPU busy-spin
* ❌ Any embedded Boa runtime using real wall clock can hang
* ❌ Legitimate JavaScript pattern (`setInterval(fn, 0)`) becomes unsafe

This affects real-world embedding scenarios and long-running runtimes.

---

## **Fix**

The termination guard now ignores recurring jobs when checking whether the loop can exit:

```rust
let no_timeout_jobs_to_run = {
    let now = context.borrow().clock().now();
    !self.timeout_jobs.borrow().iter().any(|(t, jobs)| {
        &now > t && jobs.iter().any(|job| !job.is_recurring())
    })
};
```

Now:

* Only **non-recurring** past-due jobs prevent termination
* Recurring intervals no longer block loop exit
* `is_recurring()` is finally wired into runtime logic

No changes were required to `handle()`, `enqueue_job`, or `TimeoutJob::recurring()`.

---

## **Result**

After this fix:

* ✅ `setInterval(fn, 0)` no longer pins CPU
* ✅ `run_jobs()` terminates correctly when non-recurring work is done
* ✅ Recurring intervals behave as intended
* ✅ Existing `FixedClock` tests continue to pass
* ✅ Added regression test using `StdClock` to prevent reintroduction

The event loop now correctly distinguishes between one-shot work and recurring timers.

### DIFF
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -717,19 +717,19 @@ impl JobExecutor for SimpleJobExecutor {
             // past-due. Recurring jobs (e.g. from `setInterval`) have already been dispatched
             // and re-enqueued into the future above, so they must not prevent the event loop
             // from terminating when all other work is done.
-         let has_pending_timeout_jobs = 'result: {
-    let now = context.borrow().clock().now();
+            let has_pending_timeout_jobs = 'result: {
+                let now = context.borrow().clock().now();
 
-    for (timeout, jobs) in self.timeout_jobs.borrow().iter() {
-        for job in jobs {
-            if !job.is_recurring() && &now > timeout {
-                break 'result true;
-            }
-        }
-    }
+                for (timeout, jobs) in self.timeout_jobs.borrow().iter() {
+                    for job in jobs {
+                        if !job.is_recurring() && &now > timeout {
+                            break 'result true;
+                        }
+                    }
+                }
 
-    false
-};
+                false
+            };
 
             if self.promise_jobs.borrow().is_empty()
                 && self.async_jobs.borrow().is_empty()


jedel1043:
~Uhhh what? This effectively makes it possible to hang the whole runtime if you call `setTimeout(fn, 100000000)`...~
Scratch this, I was misinterpreting the new conditional. It is really confusing though

jedel1043:
There is a very simple fix to this: just define a minimum timeout (4ms) in `enqueue_job`.

ashnaaseth2325-oss:
Hi @jedel1043 
Thanks for clarifying! I agree it’s not the clearest as written. Would you prefer I refactor it for readability or just add a brief comment explaining the intent? I’m happy to update it either way.

jedel1043:
Just make it easier to read. A comment explaining what a line of code does seems like a code smell IMO

ashnaaseth2325-oss:
Refactored the termination guard to make the condition easier to follow while keeping the behavior unchanged. Please let me know if you'd like it structured differently.

core/engine/src/job.rs
```diff
@@ -730,6 +713,37 @@ impl JobExecutor for SimpleJobExecutor {
...
+
+                let has_due_non_recurring_timeout =
+                    self.timeout_jobs.borrow().iter().any(|(t, jobs)| {
+                        let is_due = &now > t;
+                        let has_non_recurring = jobs.iter().any(|job| !job.is_recurring());
+                        is_due && has_non_recurring
+                    });
+
+                !has_due_non_recurring_timeout
+            };
```
jedel1043: This is still a bit hard to follow. What about something like this?

```rust
let has_pending_timeout_jobs = 'result: {
    let now = context.borrow().clock().now();
    for (timeout, jobs) in self.timeout_jobs.borrow().iter() {
        for job in jobs {
            if !job.is_recurring() && &now > timeout {
                break 'result true;
            }
        }
    }
    false
};

// below

if etc && !has_pending_timeout_jobs {
   break;
}
```

ashnaaseth2325-oss:
Hi @jedel1043 ,
Appreciate the suggestion! I’ve reworked the termination guard using the proposed structure and kept the behavior intact. Please let me know if you'd prefer any changes.

ashnaaseth2325-oss:
Thanks! That’s a good point about setInterval(f, 0) 
Happy to revisit the behavior if we decide to align more closely with Node in the future.

PR #4751, opened 2026-02-27
fix: handle accessor descriptors in array display

### LINKED ISSUE #4750: # Issue: Handle Accessor Descriptors in Value Display
**File:** `display.rs`

---

## Problem

When displaying JavaScript arrays, the code contained two `FIXME` comments where accessor property descriptors (getters/setters) were not handled correctly.

### 1. Array `length` Property (around line ~114)

The code used:

    .expect_value()

to read the array’s `length`.

If `length` were implemented as an accessor descriptor (getter/setter) instead of a plain data property, this would cause a runtime panic.

---

### 2. Array Element Properties (around line ~137)

When printing each element of an array, the code used:

    .value().cloned()

This returns `None` for accessor descriptors.

As a result:

- Getter/setter properties were silently displayed as `<empty>`
- This misled users into thinking the array slot was unoccupied
- The behavior was incorrect and inconsistent with JavaScript engines

---

### For `length`

Replaced:

    .expect_value()

with a safe chain using:

    .value().and_then(...)

Now:

- If the descriptor is a data property → works normally
- If it is not → gracefully falls back to `0`
- No panic occurs

---

### For Array Elements

Instead of blindly calling `.value()`, the code now checks the descriptor type:

- Data descriptor → prints actual value
- Accessor (getter only) → prints `[Getter]`
- Accessor (setter only) → prints `[Setter]`
- Getter + Setter → prints `[Getter/Setter]`
- No property → prints `<empty>`

This matches Node.js / Chrome DevTools console behavior.

---

### Before

- Two `FIXME` comments
- Potential runtime panic
- Getter/setter properties displayed incorrectly as `<empty>`

### After

- No panics
- Accurate display of accessor descriptors
- Behavior aligned with JavaScript engines
- Cleaner, more robust implementation
- Both `FIXME` comments removed

---

## Summary

This is a small, self-contained code quality improvement that:

- Prevents a possible runtime panic
- Improves correctness of array display
- Handles edge cases properly
- Removes technical debt (`FIXME`s)
- Aligns behavior with real-world JavaScript environments

rajat552:
## Summary
Fixes two FIXMEs in `display.rs` where accessor property descriptors (getters/setters) were not handled when displaying arrays.

---

## Problem

- **`length` property** — used `.expect_value()` which panics if the property is an accessor descriptor instead of a data property.
- **Array elements** — used `.value().cloned()` which returns `None` for accessor descriptors, causing getter/seter properties to incorrectly display as `<empty>`.

---

## Changes

- Replaced `.expect_value().as_number()` with:

  ```rust
  .value().and_then(|v| v.as_number())

Closes #4750

### DIFF
--- a/core/engine/src/value/display.rs
+++ b/core/engine/src/value/display.rs
@@ -152,7 +152,7 @@ fn log_array_to(
                         f.write_str(display)?;
                     }
                     DescriptorKind::Generic => {
-                        f.write_str("undefined")?;
+                        unreachable!("found generic descriptor in array")
                     }
                 }
             } else {


rajat552:
@jedel1043  review this PR let me know if is there any changes required

core/engine/src/value/display.rs
```diff
@@ -134,14 +133,22 @@ fn log_array_to(
...
+                        f.write_str("undefined")?;
+                    }
+                } else {
+                    let display = match (desc.get().is_some(), desc.set().is_some()) {
+                        (true, true) => "[Getter/Setter]",
+                        (true, false) => "[Getter]",
+                        (false, true) => "[Setter]",
+                        _ => "<empty>",
+                    };
+                    f.write_str(display)?;
```
jedel1043: For simplicity I'd recommend using [`PropertyDescriptor::kind`](https://github.com/boa-dev/boa/blob/f22ef4415d99bd63ccdf51de9b202ec99e3066e0/core/engine/src/property/mod.rs#L263) here instead.

jedel1043 (changes requested):
Thank you! I have a small suggestion to improve the code.

core/engine/src/value/display.rs
rajat552: sir I made required changes let me know if any changes required

core/engine/src/value/display.rs
```diff
@@ -134,14 +133,28 @@ fn log_array_to(
...
+                        let display = match (get.is_some(), set.is_some()) {
+                            (true, true) => "[Getter/Setter]",
+                            (true, false) => "[Getter]",
+                            (false, true) => "[Setter]",
+                            _ => "<empty>",
+                        };
+                        f.write_str(display)?;
+                    }
+                    DescriptorKind::Generic => {
+                        f.write_str("undefined")?;
```
jedel1043: We might want to panic here. A generic descriptor here would be an engine bug.

This kind reminds me that we might need to tweak descriptors to remove the `Generic` variant.

rajat552: Thanks for the suggestion.
I’ve updated the DescriptorKind::Generic case to use
unreachable!("found generic descriptor in array"), as reaching this branch would indicate an engine bug.
All 879 tests pass.

rajat552:
Thank you for the review and for merging this! I appreciate the feedback.

Issue #4742, opened 2026-02-26
Make anonymous functions identifiable in VM trace


ashddev:
This is a sub-task of [#1711](https://github.com/boa-dev/boa/issues/1711).

PR #4754, opened 2026-02-27
feat: assign auto-incremented ids to code blocks so anonynous functions can be identified in vm trace

ashddev:
This Pull Request closes [#4742](https://github.com/boa-dev/boa/issues/4742)

It changes the following:

- Adds a `debug_id` to every `CodeBlock`.
- Displays the `debug_id` for anonymous functions in the compiled output and call frame titles during tracing.

This is my first contribution to Boa, and I’m happy to make any changes based on feedback :)

### DIFF
--- a/tests/insta-bytecode/src/snapshots/insta_bytecode__compile_bytecode@basic-loop.js.snap
+++ b/tests/insta-bytecode/src/snapshots/insta_bytecode__compile_bytecode@basic-loop.js.snap
@@ -3,7 +3,7 @@ source: tests/insta-bytecode/src/lib.rs
 expression: output
 input_file: tests/insta-bytecode/scripts/basic-loop.js
 ---
----------------------------Compiled Output: '<main>'----------------------------
+-------------------------- Compiled Output: '<main>' ---------------------------
 Location     Handler      Opcode                            Operands
   000000                    PushZero                          dst:r02
   000005                    Jump                              address:000014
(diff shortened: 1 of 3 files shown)


core/engine/src/vm/code_block.rs
```diff
@@ -176,6 +180,8 @@ impl CodeBlock {
                 name,
                 SpannedSourceText::new_empty(),
             ),
+            #[cfg(feature = "trace")]
+            anon_debug_id: Cell::new(None),
```
jedel1043: I'd say for simplicitly's sake, we can just assign a incremental ID to every CodeBlock on the same thread. We do something very similar in Context:

https://github.com/boa-dev/boa/blob/f22ef4415d99bd63ccdf51de9b202ec99e3066e0/core/engine/src/context/mod.rs#L46-L48

https://github.com/boa-dev/boa/blob/f22ef4415d99bd63ccdf51de9b202ec99e3066e0/core/engine/src/context/mod.rs#L1115



Since any CodeBlock can be executed by any Context but CodeBlock's cannot be sent between threads, a thread local counter should be enough for this purpose.

ashddev: Thank you! This is a lot simpler 😅

ashddev:
Sorry about the messy git history. I ran into some trouble keeping the branch up to date with main, everything should be clean now!

core/engine/src/vm/code_block.rs
```diff
@@ -176,6 +180,8 @@ impl CodeBlock {
                 name,
                 SpannedSourceText::new_empty(),
             ),
+            #[cfg(feature = "trace")]
+            debug_id: 0,
```
jedel1043: This should probably also get a new debug_id.

What you could do here to simplify everything is to add a utility function on this file called `get_next_codeblock_id` that fetches the next codeblock id. Then you can put the thread local inside the function to avoid exposing it as public, and `bytecompiler/mod.rs` can just call that function

ashddev: Thanks, I think I've implemented this like you said. I appreciate your help!

jedel1043 (approved):
Looks great! Thank you very much!

PR #4755, opened 2026-02-28
feat(builtins): implement JsAsyncGenerator safe wrapper

KaustubhOG:
Implements the `JsAsyncGenerator` safe Rust wrapper as part of #2098.

Adds `core/engine/src/object/builtins/jsasyncgenerator.rs` with:
- `from_object(object: JsObject) -> JsResult<Self>`
- `next(value, context) -> JsResult<JsValue>`
- `return(value, context) -> JsResult<JsValue>`
- `throw(value, context) -> JsResult<JsValue>`
- `From<JsAsyncGenerator>` for `JsObject` and `JsValue`
- `Deref` to `JsObject`
- `TryFromJs` impl

Follows the same pattern as `JsGenerator`.

### DIFF
--- a/examples/src/bin/jsasyncgenerator.rs
+++ b/examples/src/bin/jsasyncgenerator.rs
@@ -1,6 +1,7 @@
 //! Example demonstrating the `JsAsyncGenerator` API wrapper.
 use boa_engine::{
-    Context, JsValue, Source, builtins::promise::PromiseState, object::builtins::JsAsyncGenerator,
+    Context, JsString, JsValue, Source, builtins::promise::PromiseState,
+    object::builtins::JsAsyncGenerator,
 };
 
 fn main() {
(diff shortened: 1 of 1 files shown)


core/engine/src/object/builtins/jsasyncgenerator.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// More information:
+    ///  - [MDN documentation][mdn]
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator/throw
+    pub fn throw<T>(&self, value: T, context: &mut Context) -> JsResult<JsValue>
+    where
+        T: Into<JsValue>,
+    {
+        AsyncGenerator::throw(&self.inner.clone().into(), &[value.into()], context)
+    }
```
jedel1043: All of these should return `JsResult<JsPromise>` to follow the protocol of `AsyncGenerator`

jedel1043 (changes requested):

core/engine/src/object/builtins/jsasyncgenerator.rs
KaustubhOG: yup i was waiting for that one, i try to handle it same way #4756 was handled , i made changes have a look @jedel1043 ?

examples/src/bin/jsasyncgenerator.rs
```diff
@@ -0,0 +1,25 @@
...
+        .eval(Source::from_bytes(
+            "async function* count() { yield 1; yield 2; yield 3; } count()",
+        ))
+        .unwrap();
+
+    let obj = result.as_object().unwrap().clone();
+    let async_gen = JsAsyncGenerator::from_object(obj).unwrap();
+
+    // next() returns a Promise
+    let promise = async_gen.next(JsValue::undefined(), &mut context).unwrap();
```
jedel1043, line 20: What you could do here is to call `context.run_jobs()` to resolve the promise, then you should be able to access the value `1` from its state.

jedel1043: That way we at least show that promises can only go forward if you call `context.run_jobs()`

KaustubhOG: yup let me handle this quickly

KaustubhOG: @jedel1043 i made some changes u suggested me , have a look at it

examples/src/bin/jsasyncgenerator.rs
```diff
@@ -0,0 +1,31 @@
...
+    drop(context.run_jobs());
+    if let PromiseState::Fulfilled(val) = promise.state() {
+        println!("next resolved with: {}", val.display());
```
jedel1043: Would be kinda better to `assert_eq` with a `JsValue::from(1)` to test that the function returns the correct value.

KaustubhOG: Hi @jedel1043, updated the example to use assert_eq! to verify the resolved promise values — JsValue::from(1) for next and JsValue::from(42) for return. The assertions pass cleanly. is it what you mean to test the function or any other way ?

jedel1043: Yep, pretty much what I meant

PR #4758, opened 2026-02-28
Add Strength Reduction and Dead Code Elimination to AST Optimizer

Psykii22:
This PR adds two new AST optimizer passes:
- **Strength Reduction**: Optimizes computationally heavy expressions 
  without side effects. Specifically:
  - Eliminates division `x / 2` in favor of multiplication `x * 0.5`.
  - Transforms exponents `x ** 2` into `x * x`. 
  Both optimizations use `std::mem::replace` instead of clone to avoid redundant memory allocations.

- **Dead Code Elimination**: Removes unreachable `if`, `while`, and `for` 
  branches/loops when the condition is a known boolean literal (after constant folding).
 handles JavaScript hoisting semantics by bailing out if the removed 
  body contains `var` declarations, `function` declarations, or generator/async variants.
also, prevents side-effect vulnerabilities in `for` loops by 
  refusing to eliminate loops that contain an initialization statement. like `for(let i=doAuth();false;i++)`
 
Also bounds all optimizer pass loops with MAX_PASS_ITERATIONS to prevent
infinite loops.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -86,6 +86,9 @@ jobs:
         with:
           toolchain: stable
 
+      - name: Install Cargo insta
+        run: cargo install --locked cargo-insta
+
       - name: Cache cargo
         uses: actions/cache@v5
         with:
@@ -106,6 +109,8 @@ jobs:
         run: cargo nextest run --profile ci --cargo-profile ci --features annex-b,intl_bundled,experimental,embedded_lz4
       - name: Test docs
         run: cargo test --doc --profile ci --features annex-b,intl_bundled,experimental
+      - name: Test bytecode output
+        run: cargo insta test -p insta-bytecode
 
   miri:
     name: Miri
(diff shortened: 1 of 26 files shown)


Psykii22:
sir while running the V8 benchmark tests, I encountered a STATUS_STACK_OVERFLOW when executing combined.js. To investigate, I cloned a fresh copy of boa-dev/boa (latest main, no modifications), built it in release mode, and reproduced the same crash.

When running with a larger stack (16 MB thread), all benchmarks completed successfully (Score: 214). also i ran some custom microbenchmarks around the specific patterns introduced by the optimisation passes, and that showed measurable performance improvements to the current boa

jedel1043:
No worries! You're probably running this on Windows, which has a much smaller stack size by default than Linux. We're kinda working towards fixing that, but than you for testing it anyways!

jedel1043 (approved):
Everything looks good! We can merge this after fixing all the CI warnings

hansl:
Hey, are those test262 still failing? That is a huge red flag that something broke.

Issue #4728, opened 2026-02-25
`TextDecoder.decode()` does not support `DataView`


zhuzhu81998:
Test code:
```js
new TextDecoder().decode(
  new DataView(new TextEncoder().encode("hello").buffer)
);
```

Firefox:
<img width="717" height="126" alt="Image" src="https://github.com/user-attachments/assets/7b8533e1-bb12-4263-835d-143621b9b2d9" />

Boa incorrectly throws
```
>> new TextDecoder().decode(
  new DataView(new TextEncoder().encode("hello").buffer)
);
Uncaught: TypeError: Argument 1 must be an ArrayBuffer, TypedArray or DataView.
```

PR #4763, opened 2026-02-28
fix: `TypedArray` offset and length not respected by `TextDecoder`

### LINKED ISSUE #4615: TextDecoder returns irrelevant underlying buffer data

zhuzhu81998:
Closes #4615 . 
Started because #4637 was closed without merging by the author while it had conflicts due to changes on the main branch from #4670 (let me know in case it was not intended to be closed).

Compared to previous PRs on this matter including #4637, this PR avoids allocating unnecessarily.

### DIFF
--- a/core/runtime/src/text/mod.rs
+++ b/core/runtime/src/text/mod.rs
@@ -161,8 +161,8 @@ impl TextDecoder {
         };
 
         Ok(match self.encoding {
-            Encoding::Utf8 => encodings::utf8::decode(&data, strip_bom),
-            Encoding::Utf16Le => encodings::utf16le::decode(&data, strip_bom),
+            Encoding::Utf8 => encodings::utf8::decode(data, strip_bom),
+            Encoding::Utf16Le => encodings::utf16le::decode(data, strip_bom),
             Encoding::Utf16Be => {
                 let owned = data.to_vec();
                 encodings::utf16be::decode(owned, strip_bom)


zhuzhu81998:
hard to believe there is no test262 for this :thinking:

jedel1043:
That's because `TextDecoder` is not part of ECMA262; it is from a separate spec (https://encoding.spec.whatwg.org/). We kinda have tests for it in https://github.com/boa-dev/boa/tree/201418a5e1506ad098522be8d308bfbea80e495a/tests/wpt, but the repo download is kinda massive so we don't run it on our CI, and IIRC we would also need to enable the tests for `TextDecoder`

core/runtime/src/text/mod.rs
jedel1043, line 139: I _think_ DataView also has a byteoffset and bytelength, so we might also have to do the same for it

core/runtime/src/text/mod.rs
```diff
@@ -117,13 +117,19 @@ impl TextDecoder {
     ///
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode
     pub fn decode(&self, buffer: JsValue, context: &mut Context) -> JsResult<JsString> {
+        let mut range = None;
         let array_buffer = if let Ok(array_buffer) = JsArrayBuffer::try_from_js(&buffer, context) {
             array_buffer
```
jedel1043, line 122: Can we just use the range `0..array_buffer.len()` here? Seems to simplify the code below, since it avoids having to use `Option`

core/runtime/src/text/mod.rs
zhuzhu81998: yeah i know. i did not realize that as i did #4728 😅 . Will open a separate pr for that ig.

however right now there is a teeny tiny problem, because `byte_offset` and `byte_length` returns usize for `TypedArray`, but for `DataView` they return u64 right now.
I think there miiiiight be a more general issue around the usage of `f64` vs `u64` vs `usize` for indices (e.g. seen in #4651 ). But will need to investigate a bit more.

jedel1043: Hmmm true. We've historically used `u64` for indices to preserve some spec-related checks, but we would need to see if we can fully migrate everyting to `usize` for simplicity

core/runtime/src/text/mod.rs
jedel1043: Resolving since this might require deciding if our indices should be `usize` or `u64`

zhuzhu81998: because we do not have the `array_buffer` at this point. what `array_buffer` is depends on the following branch check and afaik `buffer` at this point can not be used for length either.

you could avoid using the `Option` by e.g. copy pasting the code below for each of the stuff `buffer` can be, but not very nice just to avoid one more simple branching i think.

zhuzhu81998: ah i was typing as you were merging the PR XD (or rather I was confused myself why I didnt just use `array_buffer.len()` for quite awhile and i think this is not the first time i am confused because of this either...)

jedel1043: Yeah the thought was changing each branch to return (array_buffer, range) instead of only the array buffer, but it is kinda void if we have inconsistent index types

core/runtime/src/text/mod.rs
zhuzhu81998: out of curiosity, what spec-related checks?

jedel1043: I kinda don't remember actually 😅
but the justification should be on the blame for the definition of TypedArray

PR #4776, opened 2026-03-01
Fix `Date.prototype.toISOString` formatting year 0 as `-000000` instead of `0000`

### LINKED ISSUE #4775: `Date.prototype.toISOString` formats year 0 as `-000000` instead of `0000`
## Bug

`toISOString()` produces `-000000` for year 0 instead of the 4-digit `0000` format.

### Reproduce

```js
new Date("0000-06-15T00:00:00Z").toISOString()
// Actual:   "-000000-06-15T00:00:00.000Z"
// Expected: "0000-06-15T00:00:00.000Z"
```

### Cause

In `core/engine/src/builtins/date/mod.rs`, the year formatting uses `year.is_positive()` to decide the format. In Rust, `0i32.is_positive()` is `false`, so year 0 falls into the negative/extended-year branch.

### Fix

Change `year.is_positive()` to `year >= 0` in the second branch.

### Spec

[ECMA-262 §21.4.4.36](https://tc39.es/ecma262/#sec-date.prototype.toisostring) — years 0000–9999 use 4-digit format.

Ansh-699:
Fixes #4775

`toISOString()` was producing `-000000` for year 0 because `0i32.is_positive()` is `false` in Rust, causing year 0 to fall into the negative/extended-year branch.

Changed `year.is_positive()` to `year >= 0` so year 0 correctly uses the 4-digit format.

Added a unit test for the fix.

### DIFF
--- a/core/engine/src/builtins/date/mod.rs
+++ b/core/engine/src/builtins/date/mod.rs
@@ -1540,7 +1540,7 @@ impl Date {
         // 6. Return a String representation of tv in the Date Time String Format on the UTC time scale,
         //    including all format elements and the UTC offset representation "Z".
         let year = year_from_time(tv);
-        let year = if year >= 0 && year >= 10000 {
+        let year = if year >= 10000 {
             js_string!(js_str!("+"), pad_six(year.unsigned_abs(), &mut [0; 6]))
         } else if year >= 0 {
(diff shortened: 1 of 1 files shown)


nekevss (approved):
LGTM

PR #4780, opened 2026-03-01
Reject `Date.parse` hour 24 with non-zero minutes/seconds/ms

### LINKED ISSUE #4779: `Date.parse` accepts invalid hour-24 timestamps with non-zero minutes/seconds/ms
## Bug

`Date.parse` accepts strings like `"2024-01-01T24:30:00Z"` instead of returning `NaN`. Hour 24 is only valid when minutes, seconds, and milliseconds are all zero.

### Reproduce

```js
Date.parse("2024-01-01T24:00:00Z")     // valid   → 1704153600000 ✓
Date.parse("2024-01-01T24:30:00Z")     // invalid → 1704155400000 ✗ (should be NaN)
Date.parse("2024-01-01T24:00:01Z")     // invalid → 1704153601000 ✗ (should be NaN)
Date.parse("2024-01-01T24:00:00.001Z") // invalid → 1704153600001 ✗ (should be NaN)
```

### Cause

In `core/engine/src/builtins/date/utils.rs`, `parse_time` checks `hour > 24` but does not validate that when `hour == 24`, the remaining time components must all be 0.

### Spec

[ECMA-262 §21.4.1.15](https://tc39.es/ecma262/#sec-date-time-string-format) — hour 24 is only valid as `24:00:00.000` (midnight end-of-day).

Ansh-699:
Fixes #4779

`Date.parse` accepted strings like `"2024-01-01T24:30:00Z"` instead of returning `NaN`. Per spec, hour 24 is only valid as `24:00:00.000`.

Added validation in `finish()` and `finish_local()` to reject hour 24 when minute, second, or millisecond is non-zero. Added tests.

### DIFF
--- a/core/engine/src/builtins/date/tests.rs
+++ b/core/engine/src/builtins/date/tests.rs
@@ -896,3 +896,16 @@ fn date_json() {
         js_string!(r#"{"date":"2020-07-08T09:16:15.779Z"}"#),
     )]);
 }
+
+#[test]
+fn date_parse_hour24_validation() {
+    run_test_actions([
+        // 24:00:00.000 is valid (midnight end-of-day)
+        TestAction::assert("!isNaN(Date.parse('2024-01-01T24:00:00Z'))"),
+        TestAction::assert("!isNaN(Date.parse('2024-01-01T24:00:00.000Z'))"),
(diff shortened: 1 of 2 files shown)


nekevss (approved):
LGTM

PR #4781, opened 2026-03-01
refactor: replace from_utf8_unchecked with safe from_utf8

sarthakNITT:
Addresses #4181 

- Replaced `from_utf8_unchecked` with safe `from_utf8` in core/parser/src/lexer/number.rs - buf only contains ASCII bytes
- Same in core/engine/src/builtins/date/utils.rs - ASCII is already checked on the line above


nekevss:
`from_utf8` contains a UTF-8 check in it vs. not checking in `from_utf8_unchecked`. Why would we double validate here in this case if the string is ASCII?

sarthakNITT:
Sorry for the noise, the unsafe here is intentional for performance. Closing this.

PR #4785, opened 2026-03-01
Fix due-now timeout jobs in executor and CLI

### LINKED ISSUE #4782: timeout jobs scheduled for `now` can be skipped (setTimeout(0) case)
### Summary
Timeout jobs that are scheduled for the current instant ( now ) may never run. This can happen for setTimeout(fn, 0) or other jobs scheduled with a timestamp equal to the current clock. The job queues treat now as “future” due to strict > comparisons and split_off(&now) usage, so entries with timeout == now are neither executed nor kept alive by the termination checks.

### To Reproduce
1. Evaluate a script that schedules a 0ms timeout:
   ```
   setTimeout(() => console.log("tick"), 
   0);
   ```
2. Run a job loop that calls run_jobs() once after evaluation (CLI file execution or SimpleJobExecutor ).
### Expected behavior
The timeout callback should run at least once and print tick .

### Actual behavior
The callback may never run because timeout == now is treated as “not yet due,” and the job loop can exit.

### Suspected cause
The timeout queues use split_off(&now) and > comparisons. Any timeout with a key equal to now is retained in the “future” map and never counted as pending. The termination check uses now > timeout , which ignores equality.

Relevant code locations:

- SimpleJobExecutor : core/engine/src/job.rs
  - Timeout drain: split_off(&now)
  - Pending check: now > timeout
- CLI executor: cli/src/main.rs
- Example executors: examples/src/bin/smol_event_loop.rs , examples/src/bin/tokio_event_loop.rs
### Suggested fix
Treat timeout <= now as due:

- Drain with ..=now semantics (or adjust the split to include equality).
- Update pending checks to use >= rather than > so due-now jobs prevent premature termination.

### Environment
- OS: macOS (Sonoma)
- Boa version: current main

Nakshatra480:
Closes #4782 

## Summary
Fixes timeout jobs scheduled at the current instant so `setTimeout(0)` does not get skipped. Ensures due-now jobs are drained and pending checks treat equality as due.

## Changes
- Drain timeout jobs at `now` instead of treating them as future.
- Update pending timeout checks to include equality.

## Tests
- cargo test
- cargo fmt -- --check

### DIFF
--- a/core/runtime/src/interval/tests.rs
+++ b/core/runtime/src/interval/tests.rs
@@ -44,6 +44,29 @@ fn two_zero_delay_timeouts_both_fire() {
     );
 }
 
+/// Regression test for <https://github.com/boa-dev/boa/issues/4782>.
+#[test]
+fn set_timeout_zero_fires_without_clock_advance() {
+    let clock = Rc::new(FixedClock::default());
+    let context = &mut create_context(clock);
+
+    run_test_actions_with(
+        [
+            TestAction::run(indoc! {r#"
+                called = false;
+                setTimeout(() => { called = true; }, 0);
+            "#}),
+            TestAction::inspect_context(|ctx| {
+                ctx.run_jobs().unwrap();
+
+                let called = ctx.global_object().get(js_str!("called"), ctx).unwrap();
+                assert_eq!(called.as_boolean(), Some(true));
+            }),
+        ],
+        context,
+    );
+}
+
 #[test]
 fn set_timeout_basic() {
     let clock = Rc::new(FixedClock::default());


Nakshatra480:
Hey @jedel1043 @nekevss  could you please take a look and review?
- I tightened the timeout handling so due‑now jobs are always drained instead of being skipped.
- The logic is aligned in both the executor and the CLI to keep behavior consistent.
- This avoids missing immediate timeouts while keeping the change minimal and safe.
- Test262 results are unchanged overall, with one previously failing test now passing.

zhuzhu81998:
> Test262 results are unchanged overall, with one previously failing test now passing.

That one test (`test/staging/sm/Date/toISOString-01.js`) is actually fixed by #4776. It is just that the main branch conformance results is ran only once 24h and you presumably have #4776 already included in your pr.

Nakshatra480:
> > Test262 results are unchanged overall, with one previously failing test now passing.
> 
> That one test (`test/staging/sm/Date/toISOString-01.js`) is actually fixed by #4776. It is just that the main branch conformance results is ran only once 24h and you presumably have #4776 already included in your pr.

Ah, that makes sense! I was briefly confused when I saw that result. Thanks for the heads up and clarifying how the main branch conformance reporting works.

Nakshatra480:
Hi @jedel1043 i have fixed the due-now timeout jobs. Can you pls review this PR.
Thanks for your time😊

hansl:
I'd like to see a test for this PR. Do you think you could make one real quick?

Nakshatra480:
Hey @hansl the due-now timeout jobs are fixed and also added a test for this as per your feedback. Can you review this PR now?
Thank you

jedel1043:
Superseeded by #4891

Nakshatra480:
> Superseeded by #4891

Sorry for wasting your time😭

Issue #3449, opened 2023-11-03
perf issue: basic loop is 10 times slower than QJS and 100 times slower than Node


ahaoboy:
**Describe the bug**
Basic loop is very slow

**To Reproduce**
```js
const n = 10_000;

for(let i=0;i<n;i++){
    for(let j=0;j<n;j++){
    }
}
```

```bash
 hyperfine --shell zsh -- "node ./a.js" "deno run ./a.js" "boa-windows-amd64  ./a.js"   "qjs ./a.js"
```
```txt   
 Benchmark 1: node ./a.js
  Time (mean ± σ):      99.7 ms ±  11.1 ms    [User: 11.7 ms, System: 11.3 ms]
  Range (min … max):    79.0 ms … 115.5 ms    20 runs

Benchmark 2: deno run ./a.js
  Time (mean ± σ):      98.6 ms ±   9.9 ms    [User: 24.8 ms, System: 14.3 ms]
  Range (min … max):    79.8 ms … 112.1 ms    17 runs

Benchmark 3: boa-windows-amd64  ./a.js
  Time (mean ± σ):     72.532 s ±  2.410 s    [User: 47.223 s, System: 0.056 s]
  Range (min … max):   69.313 s … 75.440 s    10 runs

Benchmark 4: qjs ./a.js
  Time (mean ± σ):      1.601 s ±  0.016 s    [User: 1.033 s, System: 0.003 s]
  Range (min … max):    1.577 s …  1.629 s    10 runs

Summary
  deno run ./a.js ran
    1.01 ± 0.15 times faster than node ./a.js
   16.23 ± 1.64 times faster than qjs ./a.js
  735.58 ± 78.06 times faster than boa-windows-amd64  ./a.js
```
 
**Expected behavior**
Loop can run  faster.

**Build environment (please complete the following information):**

- OS: win11 
- Rustc version:  rustc 1.75.0-nightly (75b064d26 2023-11-01)

zhuzhu81998:
i think this specific  issue could be closed now after the newest efforts in #4864 , #4850, #4848, #4845 . boa is still slower than quickjs in general, but for this specific example the difference is not that big any more

hansl:
Correct. We are now faster than QuickJS for this particular case.

PR #4788, opened 2026-03-01
Make all Boa's API use &Context instead of &mut Context

hansl:
This was done with the help of Claude Opus 4.6 as it ended up being more than a search and replace. Tests all pass locally.

This is Proof-of-Concept for using Context as a non-mutable instance, with interior mutability. The goal is to simplify many use cases where we end up doing complicated workaround because of the borrow checker. Two examples come to mind: 1. using RefCell in executors, and 2. singling out `&mut Context` as an argument in our macros and traits for js functions (boa_class, boa_module and `*IntoJsFunction`. These will be largely simplified in a follow up PR.

This has minor noticeable improvements as we remove many clone calls that become unnecessary. It should lead to a faster VM in the future as I add more work on top of it, but it is pending merge of this work first.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -809,6 +809,8 @@ impl JobExecutor for Executor {
                     }
                 }
             }
+            context.borrow_mut().clear_kept_objects();
+            future::yield_now().await;
         }
     }
 }
--- a/.github/workflows/test262_release.yml
+++ b/.github/workflows/test262_release.yml
@@ -1,9 +1,16 @@
-name: Update Test262 Results on Release
+name: Update Test262 Results
 
 on:
   release:
     types:
       - published
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   update_test262_results:
@@ -38,7 +45,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: boa-dev/data
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.DATA_PAT }}
           path: data
 
       # Run the Test262 test suite
@@ -48,13 +55,18 @@ jobs:
           cargo run --release --bin boa_tester -- run -v -o ../data/test262
 
       # Commit and push results back to the `data` repo
-      - name: Commit and push results
+      - name: Commit results
         run: |
           cd data
           git config user.name "GitHub Actions"
           git config user.email "actions@github.com"
           git add test262
-          git commit -m "Update Test262 results for release ${{ github.ref_name }}"
-          git push
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          git commit -m "Update Test262 results ( ${{ github.ref_name }} )"
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+            # cannot use secrets.GITHUB_TOKEN since it only gives you
+            # write permissions to the current repository.
+            github_token: ${{ secrets.DATA_PAT }}
+            repository: boa-dev/data
+            directory: data
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -373,6 +373,7 @@ dependencies = [
  "num-bigint",
  "rustc-hash 2.1.1",
  "serde",
+ "strum",
 ]
 
 [[package]]
@@ -596,6 +597,7 @@ dependencies = [
  "num-traits",
  "regress",
  "rustc-hash 2.1.1",
+ "strum",
 ]
 
 [[package]]
@@ -4140,6 +4142,27 @@ name = "strsim"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7da8b5736845d9f2fcb837ea5d9e2628564b3b043a70948a3f0b778838c5fb4f"
+
+[[package]]
+name = "strum"
+version = "0.27.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "af23d6f6c1a224baef9d3f61e287d2761385a5b88fdab4eb4c6f11aeb54c4bcf"
+dependencies = [
+ "strum_macros",
+]
+
+[[package]]
+name = "strum_macros"
+version = "0.27.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7695ce3845ea4b33927c055a39dc438a45b059f7c1b3d91d38d10355fb8cbca7"
+dependencies = [
+ "heck",
+ "proc-macro2",
+ "quote",
+ "syn",
+]
 
 [[package]]
 name = "subtle"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,6 +140,7 @@ dynify = "0.1.2"
 futures-channel = "0.3.32"
 aligned-vec = "0.6.4"
 temp-env = "0.3.6"
+strum = { version = "0.27", features = ["derive"] }
 
 # ICU4X
 
--- a/core/ast/Cargo.toml
+++ b/core/ast/Cargo.toml
@@ -25,6 +25,7 @@ num-bigint.workspace = true
 serde = { workspace = true, features = ["derive"], optional = true }
 arbitrary = { workspace = true, features = ["derive"], optional = true }
 indexmap.workspace = true
+strum.workspace = true
 
 [lints]
 workspace = true
--- a/core/ast/src/punctuator/mod.rs
+++ b/core/ast/src/punctuator/mod.rs
@@ -21,7 +21,7 @@ mod tests;
 ///
 /// [spec]: https://tc39.es/ecma262/#prod-Punctuator
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+#[derive(PartialEq, Eq, Clone, Copy, Debug, strum::EnumIter)]
 pub enum Punctuator {
     /// `+`
     Add,
--- a/core/engine/src/bytecompiler/jump_control.rs
+++ b/core/engine/src/bytecompiler/jump_control.rs
@@ -347,7 +347,6 @@ impl ByteCompiler<'_> {
         let handler_index = self.handlers.len() as u32;
         let start_address = self.next_opcode_location();
 
-        // FIXME(HalidOdat): figure out value stack fp value.
         let environment_count = self.current_open_environments_count;
         self.handlers.push(Handler {
             start: start_address,
--- a/core/parser/Cargo.toml
+++ b/core/parser/Cargo.toml
@@ -24,6 +24,7 @@ icu_properties.workspace = true
 
 [dev-dependencies]
 indoc.workspace = true
+strum.workspace = true
 
 [features]
 annex-b = []
--- a/core/parser/src/lexer/tests.rs
+++ b/core/parser/src/lexer/tests.rs
@@ -9,6 +9,7 @@ use crate::source::ReadChar;
 use boa_ast::{Keyword, Position, Span, Spanned};
 use boa_interner::Sym;
 use boa_macros::utf16;
+use strum::IntoEnumIterator;
 
 fn span(start: (u32, u32), end: (u32, u32)) -> Span {
     Span::new(Position::new(start.0, start.1), Position::new(end.0, end.1))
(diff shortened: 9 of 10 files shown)


cli/src/main.rs
```diff
@@ -777,19 +777,19 @@ impl JobExecutor for Executor {
         }
     }
 
-    fn run_jobs(self: Rc<Self>, context: &mut Context) -> JsResult<()> {
+    fn run_jobs(self: Rc<Self>, context: &Context) -> JsResult<()> {
         future::block_on(self.run_jobs_async(&RefCell::new(context)))
     }
 
-    async fn run_jobs_async(self: Rc<Self>, context: &RefCell<&mut Context>) -> JsResult<()> {
+    async fn run_jobs_async(self: Rc<Self>, context: &RefCell<&Context>) -> JsResult<()> {
```
jedel1043, line 784: This kinda doesn't make sense now, and we should be able to use `&Context` for async-related things :)

core/engine/src/context/mod.rs
```diff
@@ -173,6 +173,36 @@ impl Default for Context {
...
+    /// Returns a mutable reference to the VM.
+    ///
+    /// # Safety
+    ///
+    /// This uses `UnsafeCell` internally. It is sound because `Context` is `!Sync`
+    /// and `!Send` (contains `Rc`), so it can only be accessed from a single thread.
+    /// Callers must ensure no other reference from `vm()` or `vm_mut()` is active.
+    #[inline]
+    #[allow(clippy::mut_from_ref)]
+    pub(crate) fn vm_mut(&self) -> &mut Vm {
```
jedel1043: This is unsound since two references to a `Context` could extract two mutable references to the `Vm`, causing UB

cli/src/main.rs
hansl: I know :) I even mentioned it directly in the PR description.

I just did an advanced search-and-replace, didn't touch the design yet.

Should I make this part of this PR? I wanted to discuss the merit of the approach first, before redesigning things.

core/engine/src/context/mod.rs
jedel1043: Note that this is not unsafe, but unsound. A sound API would have to return a pointer here

hansl: A pointer is not a bad idea. I was thinking if there was some zero cost type we could return here, like `VmRefMut`, instead. The would be a wrapper around a pointer that `DeferMut` or something similar. WDYT?

jedel1043: DerefMut returns `&mut Context`, so it would immediately be UB. It can only be a pointer or a pointer-like, without any visible mutable references.

hansl: Okay I'm exploring having a `with_vm_mut()` method right now.

`DerefMut` would be implemented on `VmRefMut`, which could be verified at runtime. This is the first pattern I was exploring, basically a `RefCell` specialized around `Vm`.

jedel1043: But wouldn't doing that bump into the "`RefCell` is not reentrant" problem? Or we might have to explore if we can design a sound `RefCell` that is also reentrant.

hansl: Okay so I replaced access to the Vm with `with_vm()` and `with_vm_mut()`. The reference cannot escape the closure, so that should be sound.

jedel1043: Yeah... about that...

```rust
#[test]
fn context_ub() {
    run_test_actions([TestAction::inspect_context(|context| {
        let context2 = context;

        context.with_vm_mut(move |vm| {
            context2.with_vm_mut(|vm2| {
                let a = &mut vm.frame;
                let b = &mut vm2.frame;
                a.argument_count = 0;
                b.argument_count = 0;
            })
        })
    })]);
}
```

hansl: Thanks for catching that.

I reimplemented the VM logic with a `RefCell`. I thought we needed reentrance, but apparently it passes all the tests (and Miri is currently all green). Access to the VM mutably is only done in short spans of code. I'll look deeper if we can make it even better.

jedel1043:
Oof, having to borrow the `RefCell` on each bytecode decode is tanking the performance:

### PR
RESULT Richards 233
RESULT DeltaBlue 208
RESULT Crypto 214
RESULT RayTrace 451
RESULT EarleyBoyer 538
RESULT RegExp 81.5
RESULT Splay 1024
RESULT NavierStokes 398
SCORE 309

### main

RESULT Richards 282
RESULT DeltaBlue 248
RESULT Crypto 259
RESULT RayTrace 511
RESULT EarleyBoyer 595
RESULT RegExp 81.7
RESULT Splay 1165
RESULT NavierStokes 656
SCORE 369

core/engine/src/context/mod.rs
jedel1043: It's kinda still reentrant, but since the RefCell is only being borrowed in small intervals while decoding the bytecode, you don't really get the reentrancy issue.

Downside is that it's 16% slower than the mutable context

hansl: Yes, I'm looking at passing the mutable Vm and non mutable Context now. I'm looking deeper on how the Vm and bytecode operates.

zhuzhu81998:
The performance drop looks like a deal breaker to me. The simple numeric loop from #3449 with this pr takes about 4.6s on my machine, and on main it is around 3.5s .

Additionally, do you have any concrete examples of `clone`s that can be deleted safely? It could be just even though the code compiles it just panics in runtime instead.

Maybe instead of doing `RefCell`s it would be possible to re-write some methods such that instead of taking the entire `Context`, it only takes the part that it needs? That way, even without `RefCell` you wont encounter the case where 2 separate things from `Context` are needed but you have to clone one of them, because borrow checker wont let you borrow `Context` separately
A not that meaningful example that is inside `impl Context` but does not actually have to: https://github.com/boa-dev/boa/blob/af51d4d1f1ec9aec4e1d6e46a3b29b2d8a6491f4/core/engine/src/environments/runtime/mod.rs#L594-L606 for sure does not need the entire context

hansl:
@zhuzhu81998 The `RefCell` implementation was never going to make it. Please try with the latest commit.

core/engine/src/context/mod.rs
```diff
@@ -173,6 +176,294 @@ impl Default for Context {
...
+    /// Executes a closure with a shared reference to the VM.
+    ///
+    /// The `&Vm` reference is scoped to the closure and cannot escape,
+    /// preventing aliasing with mutable VM references.
+    #[inline]
+    pub(crate) fn with_vm<R>(&self, f: impl FnOnce(&Vm) -> R) -> R {
+        // SAFETY: Context is !Sync/!Send, single-threaded access only.
+        // The reference cannot escape the closure scope.
+        f(unsafe { &*self.vm.get() })
+    }
```
jedel1043: No `&Vm` refs can exist if we're using pointers. Actually, no transitive refs to inner fields of `Vm` can exist FWIW. Otherwise it's immediate UB
```rust
#[test]
fn context_ub() {
    run_test_actions([TestAction::inspect_context(|context| {
        let context2 = context;

        context.with_vm(move |vm| {
            let frame = vm.frame();
            context2.set_pc(10);
            let pc = frame.pc;
        })
    })]);
}
```
```shell
test tests::context_ub ... error: Undefined Behavior: write access through <4878605> at alloc1864591[0x80] is forbidden
   --> core/engine/src/context/mod.rs:356:18
    |
356 |         unsafe { (*self.vm_ptr()).frame.pc = pc }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/tree-borrows.md for further information
    = help: the accessed tag <4878605> is foreign to the protected tag <4878594> (i.e., it is not a child)
    = help: this foreign write access would cause the protected tag <4878594> (currently Frozen) to become Disabled
    = help: protected tags must never be Disabled
help: the accessed tag <4878605> was created here
   --> core/engine/src/context/mod.rs:219:9
    |
219 |         self.vm.get()
    |         ^^^^^^^^^^^^^
help: the protected tag <4878594> was created here, in the initial state Frozen
   --> core/engine/src/tests/mod.rs:24:31
    |
 24 |         context.with_vm(move |vm| {
    |                               ^^
    = note: this is on thread `tests::context_`
    = note: stack backtrace:
            0: context::Context::set_pc
                at core/engine/src/context/mod.rs:356:18: 356:48
            1: tests::context_ub::{closure#0}::{closure#0}
                at core/engine/src/tests/mod.rs:26:13: 26:32
            2: context::Context::with_vm::<(), {closure@core/engine/src/tests/mod.rs:24:25: 24:34}>
                at core/engine/src/context/mod.rs:202:9: 202:38
            3: tests::context_ub::{closure#0}
                at core/engine/src/tests/mod.rs:24:9: 28:11
            4: <{closure@core/engine/src/tests/mod.rs:21:51: 21:60} as std::ops::FnOnce<(&context::Context,)>>::call_once - shim
                at /home/jedel/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5: 250:71
            5: tests::context_ub
                at core/engine/src/tests/mod.rs:21:5: 29:9
            6: tests::context_ub::{closure#0}
                at core/engine/src/tests/mod.rs:20:16: 20:16

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error; 2 warnings emitted

error: test failed, to rerun pass `-p boa_engine --lib`
```

jedel1043: We need to only use `*mut Vm` for _everything_, including read-only accesses

jedel1043: And yeah, this kinda means that the whole API of `Vm` needs to use only raw pointers to call its methods.

zhuzhu81998:
> @zhuzhu81998 The `RefCell` implementation was never going to make it. Please try with the latest commit.

well performance is back on par (at least for the simple numeric loop). but `UnsafeCell` feels ... unsafe.
I am not an expert of the code base (started barely 1 week ago), but I think I saw a few places where `Context` as an input parameter is a bit "abused". So if that can be changed, maybe we can get rid of some clones etc. without these special cells.

core/engine/src/context/mod.rs
zhuzhu81998: > We need to only use `*mut Vm` for _everything_, including read-only accesses

does this defeat the point of this pr? :thinking: it sound like it would require workaround just like `&mut Context`

jedel1043: `*mut Vm` is painful for us but for our users it's easier to pass `&Context` around. The opposite happens for `&mut Context`: easier to develop, harder to enable use cases.

This PR is basically an ergonomics improvement for users at the cost of making development harder for us, which I'm kinda in favor of, provided it's still _not too difficult_ to develop with

jedel1043: Oof, still unsound.
```rust
#[test]
fn context_ub() {
    run_test_actions([TestAction::inspect_context(|context| {
        let context2 = context;

        let realm = context2.realm();
        context.eval(Source::from_bytes(b"eval(5)"));
        let addr = realm.addr();

    })]);
}
```
```shell
test tests::context_ub ... error: Undefined Behavior: reborrow through <4878361> at alloc1864545[0x70] is forbidden
   --> core/engine/src/tests/mod.rs:27:21
    |
 27 |         let value = realm.addr();
    |                     ^^^^^ Undefined Behavior occurred here
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Tree Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/tree-borrows.md for further information
    = help: the accessed tag <4878361> has state Disabled which forbids this reborrow (acting as a child read access)
help: the accessed tag <4878361> was created here, in the initial state Frozen
   --> core/engine/src/tests/mod.rs:25:21
    |
 25 |         let realm = context2.realm();
    |                     ^^^^^^^^^^^^^^^^
help: the accessed tag <4878361> later transitioned to Disabled due to a foreign write access at offsets [0x0..0x98]
   --> core/engine/src/vm/mod.rs:508:9
    |
508 |         std::mem::swap(&mut self.frame, &mut frame);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: this transition corresponds to a loss of read permissions
    = note: this is on thread `tests::context_`
    = note: stack backtrace:
            0: tests::context_ub::{closure#0}
                at core/engine/src/tests/mod.rs:27:21: 27:26
            1: <{closure@core/engine/src/tests/mod.rs:22:51: 22:60} as std::ops::FnOnce<(&context::Context,)>>::call_once - shim
                at /home/jedel/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5: 250:71
            2: tests::context_ub
                at core/engine/src/tests/mod.rs:22:5: 29:9
            3: tests::context_ub::{closure#0}
                at core/engine/src/tests/mod.rs:21:16: 21:16

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error; 3 warnings emitted

error: test failed, to rerun pass `-p boa_engine --lib`
```

jedel1043: So now I'm seeing that we would kinda have to make the whole API of `Context` not return any references related to the Vm. Might be fine but those are kinda more clones that we originally wanted to avoid doing.

hansl:
It seems that overall the current state of the branch is that it is slightly slower than the current `main` branch. I will experiment more, but for the moment I'm just going to put this PR on ice.

jedel1043:
I'll close for now since changing all usages of `&mut Context` will naturally cause big merge conflicts. We can retry migrating to `&Context` in the future.

PR #4789, opened 2026-03-01
fix(intl): implement calendar algorithm data validation using AnyCalendarKind

Lokesh0224:
**Description:**
Resolves the TODO in ServicePreferences::validate for the LDML unicode key ca (calendar algorithm).

To find the correct way to validate calendar data, similar to how numbering system validation uses validate_extension with DecimalSymbolsV1. However, calendars don't expose a DataMarkerAttributes-based validation path like numbering systems do — attempting dry_load with calendar markers like DatetimePatternsDateGregorianV1 consistently returned IdentifierNotFound regardless of request format.

**Approach**
After digging into DateTimeFormatter::try_new_internal in ICU4X, I found that CalendarAlgorithm implements TryInto<AnyCalendarKind>, and AnyCalendar::try_new_with_buffer_provider performs the actual provider-level data validation. If the provider lacks data for the requested calendar kind, it returns an error and we clear calendar_algorithm, falling back to the locale default — consistent with how other extension keys are handled.

**Output:**
image

### DIFF
--- a/core/engine/src/builtins/intl/date_time_format/options.rs
+++ b/core/engine/src/builtins/intl/date_time_format/options.rs
@@ -593,7 +593,14 @@ impl ServicePreferences for DateTimeFormatterPreferences {
         });
 
         // Handle LDML unicode key "ca", Calendar algorithm
-        // TODO: determine the correct way to verify the calendar algorithm data.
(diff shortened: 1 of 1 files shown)


Lokesh0224:
Closing this pr because branched from the wrong base. I’ll open a clean PR.

PR #4792, opened 2026-03-01
Push test262 results on pushes to main

jedel1043:
Avoids having to wait until the next day to get updated results.

This also ensures only one push action is active at a time, since it only makes sense to push the last PR action that was in flight.

### DIFF
--- a/.github/workflows/test262_release.yml
+++ b/.github/workflows/test262_release.yml
@@ -1,9 +1,16 @@
-name: Update Test262 Results on Release
+name: Update Test262 Results
 
 on:
   release:
     types:
       - published
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   update_test262_results:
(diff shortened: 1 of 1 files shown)


zhuzhu81998:
nice!

can we maybe create a new file in the same folder with the purpose of storing the commit hash of the main branch? Something like `git rev-parse HEAD > sha.txt` should do the trick. Then the conformance bot can read it for posting.

jedel1043:
That has always been there https://github.com/boa-dev/data/blob/main/test262/refs/heads/main/results.json

The "c" key is the commit hash

It's also on the other files btw, not just on the results.json one

hansl (approved):
LGTM

jedel1043:
Time to test this bad boy

zhuzhu81998:
> That has always been there https://github.com/boa-dev/data/blob/main/test262/refs/heads/main/results.json
The "c" key is the commit hash

Did not know that! However, with the approach before this pr the hash right now seems to the commit of the boa-dev/data repo: https://github.com/boa-dev/data/commit/01dd7475337273636802a2089502a0097f5cb6f3 :thinking: .

jedel1043:
Shoot, that's true. Though this PR should fix that, so now we just need to disable the action in the data repo

PR #4795, opened 2026-03-02
Use Personal Access Token for pushing to the data repo

jedel1043:
The default github token cannot write to other repositories, so this uses a separate PAT with write permissions to `boa-dev/data`.


hansl:
How do we ensure the PAT will be kept secure?

jedel1043:
By using secrets apparently. The alternative was to use an app, but it seemed iffy to have an app for just pushing a single commit.

The secret is scoped to only being able to access the data repo, and it expires in a year for good measure. Aside from that we can't do much else

PR #4810, opened 2026-03-02
feat(builtins): implement Iterator Helpers proposal

batmnnn:
## Summary

Implements the [TC39 Iterator Helpers proposal](https://tc39.es/proposal-iterator-helpers/) in `boa_engine`, adding the `Iterator` constructor, `%IteratorHelperPrototype%`, and `%WrapForValidIteratorPrototype%` as first-class built-ins.

## Why This Matters

Iterator Helpers is a Stage 4 TC39 proposal shipping in all major engines. This brings Boa one step closer to full ES2025 compliance. The API surface requires correct prototype chain wiring, abstract-class enforcement, lazy vs eager semantics, re-entrancy guards, and proper `return()` propagation.

## Architecture

The prototype chain for an iterator helper instance:
`instance -> %IteratorHelperPrototype% -> Iterator.prototype -> %IteratorPrototype% -> Object.prototype`

### New files

| File | Purpose |
|---|---|
| `iterator_constructor.rs` | `Iterator` constructor + all prototype methods |
| `iterator_helper.rs` | `%IteratorHelperPrototype%` with stateful generator model |
| `wrap_for_valid_iterator.rs` | `%WrapForValidIteratorPrototype%` used by `Iterator.from()` |
| `tests.rs` | 39 unit tests |

### Modified files

| File | Change |
|---|---|
| `iterable/mod.rs` | Added prototype slots, test module |
| `builtins/mod.rs` | `init` calls, `global_binding`, corrected init order |
| `context/intrinsics.rs` | Added `iterator: StandardConstructor` to `StandardConstructors` |
| `core/string/src/common.rs` | Added `ITERATOR` static string |

### Implementation highlights

- **Abstract-class enforcement**: `new Iterator()` throws; subclassing works via `new.target` check
- - **`Iterator.from()`**: implements `GetIteratorFlattenable`, wraps plain iterables via `WrapForValidIterator`
- - **Lazy helpers** (`map`, `filter`, `take`, `drop`, `flatMap`): return `IteratorHelper` with `SuspendedStart -> SuspendedYield -> Completed` generator state machine; values computed lazily on `.next()` call
- - **Eager consumers** (`reduce`, `toArray`, `forEach`, `some`, `every`, `find`): drive iterator to completion with correct short-circuit for `some`/`every`/`find`
- - **`.return()` propagation**: calling `.return()` on any helper closes the underlying iterator via `IteratorClose`
- - **`constructor` accessor**: per spec, `Iterator.prototype.constructor` is a get/set accessor pair for web compatibility; inserted post-build to avoid conflict with `BuiltInBuilder`'s auto-inserted `constructor` data property
- - **Split-borrow safety**: `execute_next` restructured into per-arm scopes so Rust's borrow checker is satisfied without unsafe code
## Test Coverage
```
cargo test -p boa_engine builtins::iterable
test result: ok. 39 passed; 0 failed; 0 ignored
```

Tests cover: constructor throws, subclassing, `Iterator.from()` with arrays/strings/plain objects, all 11 prototype methods, error paths (non-callable mapper, negative/NaN limit, reduce on empty), `@@toStringTag`, `.return()` propagation, and multi-step chaining (`filter -> map -> take -> toArray`).

## Checklist

- [x] `cargo check -p boa_engine` -- 0 errors, 0 warnings
- [x] 39 unit tests pass
- [x] Spec step comments in implementation
- [x] Follows existing `BuiltInBuilder` / `JsData` patterns
- [x] No unsafe code


core/engine/src/builtins/iterable/iterator_constructor.rs
```diff
@@ -0,0 +1,803 @@
...
+    const CONSTRUCTOR_ARGUMENTS: usize = 0;
+    const STANDARD_CONSTRUCTOR: fn(&StandardConstructors) -> &StandardConstructor =
+        StandardConstructors::iterator;
+
+    /// `Iterator ( )`
+    ///
+    /// More information:
+    ///  - [ECMAScript reference][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-iterator-helpers/#sec-iterator
```
jedel1043: You might wanna update your links to point to `https://tc39.es/ecma262`, since the proposal has already been merged to the spec.

core/engine/src/builtins/iterable/iterator_constructor.rs
```diff
@@ -0,0 +1,803 @@
...
+        // Check if NewTarget is the active function object (i.e., direct `new Iterator()`)
+        let new_target_obj = new_target
+            .as_object()
+            .ok_or_else(|| JsNativeError::typ().with_message("NewTarget must be an object"))?;
+        let iterator_constructor = context.intrinsics().constructors().iterator().constructor();
+        if JsObject::equals(&new_target_obj, &iterator_constructor) {
+            return Err(JsNativeError::typ()
+                .with_message("Abstract class Iterator not directly constructable")
+                .into());
+        }
```
jedel1043: You can simplify this with `Context`'s `active_function_object` method:
```rust
        if new_target.is_undefined()
            || new_target
                == &context
                    .active_function_object()
                    .unwrap_or_else(|| context.intrinsics().constructors().iterator().constructor())
                    .into()
```

core/engine/src/builtins/iterable/iterator_constructor.rs
```diff
@@ -0,0 +1,803 @@
...
+                JsSymbol::to_string_tag(),
+                Some(get_to_string_tag),
+                Some(set_to_string_tag),
+                Attribute::CONFIGURABLE,
+            )
+            .build();
+
+        // Iterator.prototype.constructor is an accessor per the spec (web-compat requirement).
+        // We cannot add it via the builder (build() auto-inserts `constructor` as a data
+        // property), so we override it directly on the prototype after build().
```
jedel1043: `BuiltInBuilder` is internal to the engine, so we generally add support for weird edge cases like this, instead of working around its API.

core/engine/src/builtins/iterable/iterator_constructor.rs
jedel1043, line 1: This file has multiple method with no spec comments at all. I would suggest adding those comments as required.

jedel1043 (changes requested):
Looking really good!

batmnnn:
@jedel1043 Addressed all your points, fixed the accessor names, switched instances to [OrdinaryObject](cci:2://file:///Users/batman/Documents/boa/core/engine/src/builtins/object/mod.rs:48:0-48:26), simplified the new-target check, and updated all spec links.

The [BuiltInBuilder](cci:2://file:///Users/batman/Documents/boa/core/engine/src/builtins/builder.rs:140:0-145:1) one was fun to dig into, ended up adding proper [constructor_accessor()](cci:1://file:///Users/batman/Documents/boa/core/engine/src/builtins/builder.rs:392:4-409:5) support to the builder itself instead of the post-build workaround, felt like the right way to handle it.

Also caught a bug while at it: [next()](cci:1://file:///Users/batman/Documents/boa/core/engine/src/builtins/iterable/iterator_helper.rs:98:4-163:5) was reading `done` via `obj.get()` which could trigger arbitrary JS, changed [execute_next](cci:1://file:///Users/batman/Documents/boa/core/engine/src/builtins/iterable/iterator_helper.rs:165:4-478:5) to return [(JsValue, bool)](cci:1://file:///Users/batman/Documents/boa/core/engine/src/builtins/object/mod.rs:627:4-633:5) directly.

Let me know if anything else needs work!

core/engine/src/builtins/iterable/iterator_constructor.rs
```diff
@@ -0,0 +1,824 @@
...
+        // 2. If O is not an Object, throw a TypeError exception.
+        let o = this.as_object().ok_or_else(|| {
+            JsNativeError::typ().with_message("Iterator.prototype.map called on non-object")
+        })?;
+
+        // 3. If IsCallable(mapper) is false, throw a TypeError exception.
+        let mapper = args.get_or_undefined(0);
+        let mapper_obj = mapper.as_callable().ok_or_else(|| {
+            JsNativeError::typ().with_message("Iterator.prototype.map: mapper is not callable")
+        })?;
```
jedel1043: I've noticed that this is missing all the `IteratorClose` related operations. Here, for example, the spec asks you to call `IteratorClose` if the mapper is not callable:

> If [IsCallable](https://tc39.es/ecma262/#sec-iscallable)(mapper) is false, then
>   a. Let error be [ThrowCompletion](https://tc39.es/ecma262/#sec-throwcompletion) (a newly created TypeError object).
>   b. Return ? [IteratorClose](https://tc39.es/ecma262/#sec-iteratorclose)(iterated, error).

That doesn't happen only here, but throughout all prototype methods. I would suggest you to go through the spec and check line by line that you're covering it properly as specified.

jedel1043 (changes requested):

jedel1043 (approved):
Don't have anything else to comment on. Nice work!

PR #4812, opened 2026-03-02
Bump dependencies

jedel1043:


Cargo.toml
```diff
@@ -125,8 +125,8 @@ intrusive-collections = "0.10.0"
 cfg-if = "1.0.4"
 either = "1.15.0"
 sys-locale = "0.3.2"
-timezone_provider = { version = "0.1.2" }
-temporal_rs = { version = "0.1.2", default-features = false, features = ["float64_representable_durations"] }
+timezone_provider = { version = "0.2.0" }
```
nekevss, line 128: Note: it would be nice to switch to the `ZeroCompiledTzdbProvider` in the future

PR #4814, opened 2026-03-02
Remove stale FIXME comment in exception handler setup

Nakshatra480:
## Summary
Removes a stale `FIXME(HalidOdat): figure out value stack fp value.` comment from the exception handler setup in `jump_control.rs`.

After investigation, the VM's existing call frame push/pop (`push_frame`/`pop_frame`) and `handle_throw` already restore the value stack correctly via `truncate_to_frame` when unwinding across frames. The value stack frame pointer tracking mentioned in the FIXME is not needed.

## Changes
- Removed the obsolete FIXME comment in `push_handler()` (`core/engine/src/bytecompiler/jump_control.rs`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -373,6 +373,7 @@ dependencies = [
  "num-bigint",
  "rustc-hash 2.1.1",
  "serde",
+ "strum",
 ]
 
 [[package]]
@@ -596,6 +597,7 @@ dependencies = [
  "num-traits",
  "regress",
  "rustc-hash 2.1.1",
+ "strum",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


jedel1043:
Uhh I tested your new test cases against main and they seem to pass? What does this PR fix? Alternatively, do you have a test case that fails on main but passes on this PR?

Nakshatra480:
Hey @jedel1043 My bad, you're right! After looking closer at how main handles this, I realized handle_throw and truncate_to_frame actually already clean up the value stack properly when unwinding. The FIXME was just stale. I've updated the PR to simply remove the obsolete comment instead.

PR #4815, opened 2026-03-02
Fix TextDecoder to respect DataView byteOffset and byteLength

### LINKED ISSUE #4767: `TextDecoder` does not respect `DataView` offset and length
Similar to #4615 :
```js
const buffer = Uint8Array.of(0x42, 0x42).buffer;

// Create a DataView that starts at byte offset 1
const view = new DataView(buffer, 1);

console.log(new TextDecoder().decode(view));
```
Expected: `B`, actual: `BB`.

Let us handle this after #4766 .

shreyadesai20042009:
Fixes #4767

This change ensures that TextDecoder.decode correctly respects
the byteOffset and byteLength of DataView objects.

Previously, DataView inputs were decoded using the full
underlying ArrayBuffer instead of the specified view range.

Now the decoder properly slices the buffer before decoding.

### DIFF
--- a/core/runtime/src/text/mod.rs
+++ b/core/runtime/src/text/mod.rs
@@ -116,59 +116,69 @@ impl TextDecoder {
     /// Any error that arises during decoding the specific encoding.
     ///
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode
-    pub fn decode(&self, buffer: JsValue, context: &mut Context) -> JsResult<JsString> {
-        let mut range = None;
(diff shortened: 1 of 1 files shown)


core/runtime/src/text/mod.rs
```diff
@@ -116,59 +116,69 @@ impl TextDecoder {
...
+        range = Some(offset..offset + length);
 
-            JsArrayBuffer::from_object(obj)?
-        } else {
-            return Err(js_error!(
-                TypeError: "Argument 1 must be an ArrayBuffer, TypedArray or DataView."
-            ));
+        JsArrayBuffer::from_object(obj)?
+    }
+    // ✅ FIXED: Handle DataView properly
```
zhuzhu81998, line 141: looks like ai

Issue #4798, opened 2026-03-02
CLI crashes on recoverable errors due to unwrap() calls


Tomeshwari-02:
*** THE BUG ****

The CLI module contains multiple unwrap() calls that can cause the application to crash on recoverable errors instead of handling them gracefully. When operations such as file reading, network requests, or parsing encounter errors, the application panics rather than providing proper error messages or graceful fallbacks.

****REPRODUCTION *****

Steps to reproduce the issue:

Run the Boa CLI with a malformed JavaScript file.

Run the CLI with a file that has insufficient permissions.

Run the CLI with a file that does not exist.

The application will crash with a panic instead of displaying a proper error message.

******Example scenario:******

# This will crash if the file doesn't exist or has syntax errors
./boa nonexistent-file.js
Expected behavior

Instead of crashing with a panic, the application should handle errors gracefully by:

Providing informative error messages to the user.

Returning appropriate exit codes.

Continuing execution where appropriate.

Avoiding unexpected termination on recoverable errors.

The application should properly handle file I/O errors, parsing errors, and other recoverable conditions without crashing.

Build environment (please complete the following information):

OS: Any

Version: All versions

Target triple: All targets

Rustc version: All versions

Additional context

This issue negatively impacts the robustness and user experience of the CLI tool. Using unwrap() for error handling in production code is inappropriate because it leads to crashes on recoverable errors.

The fix involves replacing unwrap() calls with proper error handling using the ? operator or implementing appropriate error recovery mechanisms throughout the CLI module.

Known locations where unwrap() is used include:

File reading operations in cli/src/main.rs

Network request handling

Parsing operations that may encounter malformed input

Addressing this issue would align the project with Rust best practices for error handling and significantly improve the overall reliability of the application.

PR #4816, opened 2026-03-02
fix(cli): handle recoverable errors gracefully and remove panics #4798

Shivampal157:
This PR addresses issue #4798 by refactoring the CLI module to eliminate potential panics on recoverable errors. I have replaced several instances of .unwrap() and .expect() with proper Rust error handling patterns, specifically using the ? operator for error propagation and map_err for context-aware reporting.
Key Changes:
Error Propagation in main.rs: Converted AST serialization and runtime registration paths to return Result. This ensures that issues like circular references or initialization failures are caught and reported instead of crashing the process.
Robust Mutex Handling in logger.rs: Switched from .expect() to unwrap_or_else when locking the printer mutex. This allows the CLI to remain functional even if a previous thread left the mutex in a poisoned state.
Safe Regex in helper.rs: Handled regex compilation results gracefully. If a pattern fails to compile (e.g., due to invalid characters), the CLI now simply disables highlighting for that segment rather than panicking.
Debug Module Refactor: Updated init_boa_debug_object to propagate errors back to the caller, ensuring a clean exit on failure.
Verification & Testing:
I have manually verified these fixes against the reproduction steps provided in the issue:
Missing Files: Verified that ./boa nonexistent.js now outputs a clean "failed to read file" error with an exit code of 1.
Malformed JS: Confirmed that syntax errors are caught and reported without triggering a panic.
Permissions: Tested with restricted file access to ensure the I/O error path works as expected.

### DIFF
--- a/cli/src/debug/mod.rs
+++ b/cli/src/debug/mod.rs
@@ -2,6 +2,7 @@
 #![allow(clippy::unnecessary_wraps)]
 
 use boa_engine::{Context, JsObject, js_string, object::ObjectInitializer, property::Attribute};
+use color_eyre::Result;
 
 mod function;
 mod gc;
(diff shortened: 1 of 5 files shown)


jedel1043:
Needs a rebase first

PR #4824, opened 2026-03-02
Revert commit dc2f73695a786ac7206be84c634c18af19f7e1bd

jedel1043:
Uhh the original code was already correct; `emit_binding_access` already checks if the binding is lexical or not.

### DIFF
--- a/core/engine/src/bytecompiler/expression/unary.rs
+++ b/core/engine/src/bytecompiler/expression/unary.rs
@@ -39,12 +39,11 @@ impl ByteCompiler<'_> {
                         let identifier = identifier.to_js_string(self.interner());
                         let binding = self.lexical_scope.get_identifier_reference(identifier);
                         let index = self.get_binding(&binding);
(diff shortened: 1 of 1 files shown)


zhuzhu81998:
> emit_binding_access already checks if the binding is lexical or not.

where though? :thinking:

jedel1043:
[`get_binding`](https://github.com/jedel1043/boa/blob/dd35a0f5057cc8ee0c0091ee32bc5e6e000a7d35/core/engine/src/bytecompiler/mod.rs#L696-L698) determines the variable's TDZ before even calling `emit_binding_access`. If the variable is a `let` binding, it will pass `BindingKind::Local(None)` to `emit_binding_access`, which will [push a ReferenceError instruction for it](https://github.com/jedel1043/boa/blob/dd35a0f5057cc8ee0c0091ee32bc5e6e000a7d35/core/engine/src/bytecompiler/mod.rs#L897-L903)

Issue #3241, opened 2023-08-26
Address `Panics` in `boa_engine`


nekevss:
Related to #3235.

With Rust v1.72, `clippy` began linting any panics that have missing `panics` documentation.

Some potential approaches to addressing the panics are: remove the panics and update them to return a `RuntimeError`/`ImplementationError` or add the documentation to the function that identified as panicking, or a combination of the two of those.

#3235 disables the lint, but the panics should be addressed and the lint reenables.

jedel1043:
Seeing that some language engines such as [Rhai](https://github.com/rhaiscript/rhai) have a Don't Panic guarantee as a feature, maybe we should consider migrating to an `ImplementationError` as you've said, instead of panicking everywhere.

KaustubhOG:
Hey @jedel1043! as you suggested i added a varient , before proceeding .

KaustubhOG:
Tracking progress for converting panics in `boa_engine` to `EngineError::Panic` using `js_expect` (#4828).
> **Note:** The PRs listed below may look messy or unstructured for now  don't worry, that's intentional . The comments above (where I noted "this approach is wrong") are also being updated and revised as the approach evolved through discussion with maintainers. Everything will  be cleaned up and reorganized into a clear architecture with sections and subsections before this month ends . 

Progress so far:
- [x] #4826 — added `Panic` variant to `EngineError`
- [x] #4837 — converted `jsarray.rs` (14 panics)
- [x] #4844 — converted `jstypedarray.rs` (26 panics)
- [x] #4842 — converted `jsdataview.rs` (12 panics)
- [x] #4846  — converted `jsregexp.rs` (17 panics)
- [x] #4854   — converted `jsrproxy.rs` (13 panics)
- [x] #4875   — converted `jsrpromise.rs` (23 panics)
- [x] #4925    — converted `builtins/proxy,json,uri` (24 panics)
- [x] #4947   — converted `builtins/regexp,async-generator` (32 panics)
- [x] #5001 — converted `builtins/function/mod, function/arguments, function/bound, object/mod` (29 panics) 
- [x] #5007 — converted `builtins/array/mod, array/from_async, string/mod` (29 panics) 
- [x] #5014 —  converted `builtins/intl, builtins/promise`(26panics)
- [x] #5045 —converted `builtins/typed_array,iterable,generator,error,bigint,eval`(47panics)--------->/builtins completed
[...]

PR #4826, opened 2026-03-02
feat(error): add Panic variant to EngineError

KaustubhOG:
Part of #3241

Adds a new `Panic` variant to `EngineError` with a `message: String` field to provide context about the original panic condition.

This variant will be used to convert panics across `boa_engine` intorecoverable errors that bubble up to the host application without crashing the process, and cannot be caught from within ECMAScript code.

`Copy` was removed from `EngineError` since `String` does not implement`Copy`. A manual `unsafe impl Trace` replaces `#[boa_gc(empty_trace)]`
for the same reason.

### DIFF
--- a/core/engine/src/error/mod.rs
+++ b/core/engine/src/error/mod.rs
@@ -335,11 +335,7 @@ pub enum EngineError {
     RuntimeLimit(#[from] RuntimeLimitError),
 
     /// Error thrown when an internal panic condition is encountered.
-    ///
-    /// This error is used to convert panics into recoverable errors that bubble
-    /// up to the host application without crashing the process. It cannot be
-    /// caught from within ECMAScript code.
-    #[error("PanicError: {message}")]
+    #[error("EnginePanic: {message}")]
     Panic {
         /// The original panic message providing context about what went wrong.
         message: String,


core/engine/src/error/mod.rs
```diff
@@ -334,6 +333,23 @@ pub enum EngineError {
     /// Error thrown when a runtime limit is exceeded.
     #[error("RuntimeLimitError: {0}")]
     RuntimeLimit(#[from] RuntimeLimitError),
+
+    /// Error thrown when an internal panic condition is encountered.
+    ///
+    /// This error is used to convert panics into recoverable errors that bubble
+    /// up to the host application without crashing the process. It cannot be
+    /// caught from within ECMAScript code.
+    #[error("PanicError: {message}")]
```
jedel1043:
```suggestion
    #[error("EnginePanic: {message}")]
```

core/engine/src/error/mod.rs
```diff
@@ -334,6 +333,23 @@ pub enum EngineError {
     /// Error thrown when a runtime limit is exceeded.
     #[error("RuntimeLimitError: {0}")]
     RuntimeLimit(#[from] RuntimeLimitError),
+
+    /// Error thrown when an internal panic condition is encountered.
+    ///
+    /// This error is used to convert panics into recoverable errors that bubble
+    /// up to the host application without crashing the process. It cannot be
+    /// caught from within ECMAScript code.
```
jedel1043:
```suggestion
```
Just explaining what the error represents should be enough for documentation.

core/engine/src/error/mod.rs
```diff
@@ -644,7 +660,7 @@ impl JsError {
         }
     }
 
-    /// Gets the inner [`JsNativeError`] if the error is an engine
+    /// Gets the inner [`EngineError`] if the error is an engine
```
jedel1043, line 659: Oh, nice catch!

jedel1043 (changes requested):

core/engine/src/error/mod.rs
KaustubhOG: yup just a second

core/engine/src/error/mod.rs
KaustubhOG: thanks

KaustubhOG:
Thanks @jedel1043 for the detailed guidance ,looking forward to refactor

KaustubhOG:
> Looks great! Should be a nice starting point for the panics refactor

Yup, I will start working on the approach

PR #4828, opened 2026-03-03
Improve panic error ergonomics

jedel1043:
I was experimenting with the new PanicError and I found it really useful, but kind of a pain to use. This adjusts its definition such that it is kind of a replacement for `expect`: you just call `js_expect` on either a `JsResult` or an `Option` and it automatically creates the `PanicError`.

### Before
```rust
let promise = self
            .stack
            .get(frame.promise_capability_promise_register_index())
            .expect("stack must have a promise capability")
            .as_object()?;
```

After:

```rust
let promise = self
            .stack
            .get(frame.promise_capability_promise_register_index())
            .and_then(JsValue::as_object)
            .js_expect("stack must have a promise capability")?;
```

cc @KaustubhOG so that you're aware about the API changes

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -197,7 +197,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Check for typos
-        uses: crate-ci/typos@v1.43.5
+        uses: crate-ci/typos@v1.44.0
 
   clippy:
     name: Lint
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2640,9 +2640,9 @@ dependencies = [
 
 [[package]]
 name = "js-sys"
-version = "0.3.88"
+version = "0.3.91"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


KaustubhOG:
> I was experimenting with the new PanicError and I found it really useful, but kind of a pain to use. This adjusts its definition such that it is kind of a replacement for `expect`: you just call `js_expect` on either a `JsResult` or an `Option` and it automatically creates the `PanicError`.
> 
> ### Before
> ```rust
> let promise = self
>             .stack
>             .get(frame.promise_capability_promise_register_index())
>             .expect("stack must have a promise capability")
>             .as_object()?;
> ```
> 
> After:
> 
> ```rust
> let promise = self
>             .stack
>             .get(frame.promise_capability_promise_register_index())
>             .and_then(JsValue::as_object)
>             .js_expect("stack must have a promise capability")?;
> ```
> 
> cc @KaustubhOG so that you're aware about the API changes

Thanks for the cc! Already using js_expect in #4837

PR #4832, opened 2026-03-03
feat(intl): implement Intl.DateTimeFormat.supportedLocalesOf()

Lokesh0224:
Implements the missing supportedLocalesOf static method on Intl.DateTimeFormat, following the same pattern as remaining Intl.___.supportedLocalesOf().

**Output**
image


Lokesh0224:
@jedel1043 could you please take a look when you have time?

PR #4833, opened 2026-03-03
fix(executor): prevent CPU busy-spin when async jobs are pending

ashnaaseth2325-oss:
## Summary

This PR fixes a CPU busy-spin in `run_jobs_async` affecting both:

* `core/engine/src/job.rs`
* `cli/src/main.rs`

Previously, the executor used:

```rust
future::poll_once(group.next()).await.flatten();
future::yield_now().await;
```

When any `NativeAsyncJob` future was pending (e.g., `Atomics.waitAsync`), this combination never truly suspended the thread. The loop re-polled continuously, causing sustained 100% CPU usage until the async job resolved.

---

## Steps to Reproduce

```js
const sab = new SharedArrayBuffer(4);
const ia  = new Int32Array(sab);
Atomics.waitAsync(ia, 0, 0, 3000);
```

Running via CLI would peg one CPU core at 100% for the full duration.

---

## Fix

When no synchronous work is available, the executor now properly blocks on:

```rust
group.next().await;
```

Instead of busy-polling + yielding.

---

## Result

* CPU usage normalizes during async waits
* Executor properly parks until a future completes
* No behavioral or Test262 changes
* Improved stability for CLI and embedders

### DIFF
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -38,6 +38,7 @@ use crate::{
     realm::Realm,
 };
 use boa_gc::{Finalize, Trace};
+use futures_channel::oneshot;
 use futures_concurrency::future::FutureGroup;
 use futures_lite::{StreamExt, future};
 use portable_atomic::AtomicBool;
@@ -48,6 +49,7 @@ use std::mem;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::sync::atomic::Ordering;
+use std::task::Poll;
 use std::{cell::RefCell, collections::VecDeque, fmt::Debug, future::Future, pin::Pin};
 
 /// An ECMAScript [Job Abstract Closure].
@@ -718,7 +720,10 @@ impl JobExecutor for SimpleJobExecutor {
                 let now = context.borrow().clock().now();
                 let jobs_to_run = {
                     let mut timeout_jobs = self.timeout_jobs.borrow_mut();
-                    let mut jobs_to_keep = timeout_jobs.split_off(&now);
+                    // Use `now + 1ns` so jobs whose deadline equals `now` are
+                    // included in `jobs_to_run` rather than deferred.
+                    let split_at = now + time::Duration::from_nanos(1).into();
+                    let mut jobs_to_keep = timeout_jobs.split_off(&split_at);
                     jobs_to_keep.retain(|_, jobs| {
                         jobs.retain(|job| !job.is_cancelled());
                         !jobs.is_empty()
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @jedel1043,
This PR prevents CPU busy-spinning in the executor when async jobs are pending.
All CI checks are passing and conformance is unchanged.
Please let me know if any refinements are needed.
Thanks!

core/engine/src/job.rs
```diff
@@ -739,8 +739,22 @@ impl JobExecutor for SimpleJobExecutor {
...
+                && self.async_jobs.borrow().is_empty()
+                && self.generic_jobs.borrow().is_empty()
+                && no_timeout_jobs_to_run;
+
+            if no_sync_work && !group.is_empty() {
+                if let Some(Err(err)) = group.next().await {
+                    self.clear();
+                    return Err(err);
+                }
+            } else if let Some(Err(err)) = future::poll_once(group.next()).await.flatten() {
```
jedel1043: This kinda fails if you have a long enough timeout, right? Something like `setTimeout(f, 10000)` would make it such that the `group.next().await` path is never taken.

ashnaaseth2325-oss:
Hello @jedel1043 ,
Thanks for the catch! I’ve updated the logic so we only block when no timeout jobs are actually due to run.
All other CI checks are passing now. It looks like the test262 job hit the 60-minute timeout and was automatically cancelled, but it was progressing normally with no failures before that.
Please let me know if you’d like me to re-run it.

jedel1043:
Gonna add this to #4891 and add you as a co-author, since it's simpler to not exit on interval jobs.

ashnaaseth2325-oss:
Sounds good, thanks! Glad to see the change integrated there.

jedel1043:
Wait, this is kinda wrong. Since we take `&RefCell<&mut Context>` as argument, any other caller that also captures that can also schedule jobs. The event loop is not "isolated", so awaiting here would block the whole event loop if the pending async jobs take long enough. 
Here I have a repro of what I'm talking about:

```rust
#[test]
fn test_async_job_not_blocking_event_loop() {
    let clock = Rc::new(FixedClock::default());
    let context = &mut ContextBuilder::default()
        .clock(clock.clone())
        .build()
        .unwrap();

    run_test_actions_with(
        [TestAction::inspect_context_async(async move |ctx| {
            let executor = ctx.downcast_job_executor::<SimpleJobExecutor>().unwrap();
            let ctx = &RefCell::new(ctx);

            let mut event_loop = pin!(future::poll_once(executor.run_jobs_async(ctx)));

            // There are no jobs in our queue. Push
            // an async job that will consistently yield to the executor.
            ctx.borrow_mut().enqueue_job(
                NativeAsyncJob::new(async |_| {
                    loop {
                        future::yield_now().await;
                    }
                })
                .into(),
            );

            // Then, start the event loop
            assert!(event_loop.as_mut().await.is_none());

            let checker = Rc::new(Cell::new(false));
            {
                let checker = checker.clone();
                // At this point, the event loop should have yielded again to the async executor.
                // Thus, enqueue a generic job that should resolve in the next loop.
                let realm = ctx.borrow().realm().clone();
                ctx.borrow_mut().enqueue_job(
                    GenericJob::new(
                        move |_| {
                            checker.set(true);
                            Ok(JsValue::undefined())
                        },
                        realm,
                    )
                    .into(),
                );
            }

            // Next iteration of the event loop
            assert!(event_loop.as_mut().await.is_none());

            // At this point, our generic job should have been executed.
            assert!(checker.get());
        })],
        context,
    );
}
```
Since the event loop blocks awaiting the async job, the generic job never gets executed.

jedel1043:
Closing in favour of #4994.

PR #4836, opened 2026-03-03
Add AbortController/AbortSignal support and fetch integration

### LINKED ISSUE #4825: Add AbortController/AbortSignal and integrate with fetch
### Summary
Implement `AbortController` and `AbortSignal` in `boa_runtime` and wire `RequestInit.signal` into the Fetch API so requests can be canceled and reject with the correct abort error.

### Problem
Many modern JS libraries rely on `AbortController` to cancel in-flight fetch requests (timeouts, navigation, race conditions). Boa’s runtime currently exposes fetch but does not support abort signals, so compliant Fetch usage fails or must be polyfilled.

### Current gap
- `RequestInit` only supports `body`, `headers`, and `method`.
- No `AbortController` or `AbortSignal` runtime extension is available.

### Proposed solution
- Add global `AbortController` and `AbortSignal` classes.
- Extend `RequestInit` to accept `signal`.
- Store the signal in `Request`.
- In fetch, observe abort state and reject with an abort error when signaled.

### Acceptance criteria
- `AbortController` and `AbortSignal` are available globally.
- `RequestInit.signal` is accepted and propagated.
- `fetch()` rejects with an abort error if signaled before or during request.
- Tests cover:
  - Abort before fetch starts
  - Abort during fetch
  - Multiple listeners and repeated abort calls
  - Abort signal reuse across requests

### Notes
This aligns with the Fetch standard and enables compatibility with real-world JS code that expects abortable requests.

Nakshatra480:
## Summary
This implements the [AbortController](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/abort/mod.rs:96:0-98:1) and [AbortSignal](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/abort/mod.rs:17:0-22:1) Web APIs in the runtime and integrates them with the Fetch API, allowing requests to be correctly cancelled.

## Changes
- Created [JsAbortSignal](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/abort/mod.rs:17:0-22:1) and [JsAbortController](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/abort/mod.rs:96:0-98:1) classes inside a new [abort](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/abort/mod.rs:114:4-121:5) module.
- Added `signal: Option<JsObject>` to [RequestInit](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/fetch/request.rs:51:0-56:1) to accept the abort signal.
- Modified [fetch_inner](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/fetch/mod.rs:103:0-168:1) to check the signal's abort status both before making the HTTP request and right after it completes.
- Registered the new classes alongside fetch via [AbortControllerExtension](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/extensions.rs:116:0-116:36) (gated behind the [fetch](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/fetch/mod.rs:183:4-198:5) feature flag).
- Used `GcRefCell` for the signal's reason and listeners so they properly participate in garbage collection.

## Testing
- Added 11 new unit tests in [abort/tests.rs](cci:7://file:///Users/nakshatrasharma/Desktop/boa/core/runtime/src/abort/tests.rs:0:0-0:0) covering all acceptance criteria (initial state, abort behavior, custom/default reasons, event listeners, and idempotency).
- Ran the full test suite (`cargo test -p boa_runtime`) locally to ensure no regressions in existing fetch behavior.

## Issue
- Closes #4825

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+if ! command -v cargo-make >/dev/null 2>&1; then
+    echo "cargo-make is not installed. Install it with:"
+    echo "  cargo install cargo-make"
+    exit 1
+fi
+
+cargo make run-ci
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -440,6 +440,7 @@ dependencies = [
  "futures-lite 2.6.1",
  "getrandom 0.4.2",
  "hashbrown 0.16.1",
+ "husky-rs",
  "iana-time-zone",
  "icu_calendar",
  "icu_casemap",
@@ -1927,6 +1928,12 @@ name = "httparse"
 version = "1.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6dbf3de79e51f3d586ab4cb9d5c3e2c14aa28ed23d180cf89b4df0454a69cc87"
+
+[[package]]
+name = "husky-rs"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "62b00c51c8ea1f7e8d93caaf071948f766eb331cd55cb9871096614c9291642e"
 
 [[package]]
 name = "hyper"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -144,6 +144,7 @@ futures-channel = "0.3.32"
 aligned-vec = "0.6.4"
 temp-env = "0.3.6"
 strum = { version = "0.28", features = ["derive"] }
+husky-rs = "0.3.2"
 
 # ICU4X
 
--- a/benches/scripts/closures/create.js
+++ b/benches/scripts/closures/create.js
@@ -1,0 +1,14 @@
+// Measures closure creation overhead: allocation of function objects
+// and captured variable environments inside a hot loop.
+const kIterationCount = 100_000;
+
+function main() {
+  let sum = 0;
+  for (let i = 0; i < kIterationCount; i++) {
+    const x = i;
+    const y = i * 2;
+    const closure = (z) => x + y + z;
+    sum += closure(i + 1);
+  }
+  return sum;
+}
--- a/benches/scripts/closures/invoke.js
+++ b/benches/scripts/closures/invoke.js
@@ -1,0 +1,32 @@
+// Measures closure invocation overhead: closures are created once,
+// then called repeatedly to isolate dispatch and variable lookup cost.
+const kIterationCount = 100_000;
+
+function makeCounter(start) {
+  let count = start;
+  return {
+    increment: (n) => {
+      count += n;
+      return count;
+    },
+    decrement: (n) => {
+      count -= n;
+      return count;
+    },
+    value: () => count,
+  };
+}
+
+const counterA = makeCounter(0);
+const counterB = makeCounter(1000);
+const counterC = makeCounter(-500);
+
+function main() {
+  let sum = 0;
+  for (let i = 0; i < kIterationCount; i++) {
+    sum += counterA.increment(1);
+    sum += counterB.decrement(1);
+    sum += counterC.value();
+  }
+  return sum;
+}
--- a/benches/scripts/intl/collator-compare.js
+++ b/benches/scripts/intl/collator-compare.js
@@ -1,0 +1,8 @@
+const kIterationCount = 1000;
+const collator = new Intl.Collator("en");
+
+function main() {
+  for (let i = 0; i < kIterationCount; i++) {
+    collator.compare("apple", "banana");
+  }
+}
--- a/benches/scripts/intl/collator-construction.js
+++ b/benches/scripts/intl/collator-construction.js
@@ -1,0 +1,10 @@
+const kIterationCount = 100;
+const locales = ["en", "de", "ja", "ar", "fr", "es", "zh"];
+
+function main() {
+  for (let i = 0; i < kIterationCount; i++) {
+    for (const locale of locales) {
+      new Intl.Collator(locale);
+    }
+  }
+}
--- a/benches/scripts/intl/datetimeformat-construction.js
+++ b/benches/scripts/intl/datetimeformat-construction.js
@@ -1,0 +1,10 @@
+const kIterationCount = 100;
+const locales = ["en", "de", "ja", "ar", "fr", "es", "zh"];
+
+function main() {
+  for (let i = 0; i < kIterationCount; i++) {
+    for (const locale of locales) {
+      new Intl.DateTimeFormat(locale);
+    }
+  }
+}
--- a/benches/scripts/intl/datetimeformat-format.js
+++ b/benches/scripts/intl/datetimeformat-format.js
@@ -1,0 +1,9 @@
+const kIterationCount = 1000;
+const dtf = new Intl.DateTimeFormat("en");
+const date = new Date(2024, 0, 1);
+
+function main() {
+  for (let i = 0; i < kIterationCount; i++) {
+    dtf.format(date);
+  }
+}
--- a/benches/scripts/intl/datetimeformat-with-options.js
+++ b/benches/scripts/intl/datetimeformat-with-options.js
@@ -1,0 +1,16 @@
+const kIterationCount = 100;
+
+function main() {
+  for (let i = 0; i < kIterationCount; i++) {
+    new Intl.DateTimeFormat("en", {
+      year: "numeric",
+      month: "long",
+      day: "numeric",
+    });
+    new Intl.DateTimeFormat("en", {
+      hour: "2-digit",
+      minute: "2-digit",
+      second: "2-digit",
+    });
+  }
+}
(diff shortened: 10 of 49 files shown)


core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+#[cfg(test)]
+mod tests;
+
+/// The JavaScript `AbortSignal` class.
+#[derive(Debug, Clone, Default, JsData, Trace, Finalize)]
+pub struct JsAbortSignal {
+    #[unsafe_ignore_trace]
+    aborted: Cell<bool>,
+    #[unsafe_ignore_trace]
+    reason: RefCell<Option<JsValue>>,
```
hansl: This definitely need trace.

core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[unsafe_ignore_trace]
+    listeners: RefCell<Vec<JsFunction>>,
```
hansl: And this.

hansl (changes requested):

core/runtime/src/abort/mod.rs
Nakshatra480: Good catch! I've replaced RefCell with GcRefCell so it traces properly.

core/runtime/src/abort/mod.rs
Nakshatra480: Fixed this one as well with GcRefCell.
Thanks for the feedback😊

Nakshatra480:
Hi @hansl 

Would appreciate your thoughts on this approach. Please let me know if there’s anything else that should be adjusted or improved here.

Nakshatra480:
Hi @jedel1043 can you review this pr whenever you gets time.
Thanks

jedel1043:
I'm not in charge of this PR, so no. You've already re-requested a review from Hans, so just be patient.

Nakshatra480:
Ok i will be waiting for review by @hansl

core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub fn is_aborted(&self) -> bool {
+        self.aborted.get()
+    }
+
+    /// Returns the abort reason.
+    pub fn abort_reason(&self) -> JsValue {
+        self.reason
+            .borrow()
+            .clone()
+            .unwrap_or_else(|| js_string!("AbortError").into())
```
hansl: From https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason:

> The property is undefined when the signal has not been aborted. It can be set to a specific value when the signal is aborted, using [AbortController.abort()](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) or [AbortSignal.abort()](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/abort_static). If not explicitly set in those methods, it defaults to "AbortError" [DOMException](https://developer.mozilla.org/en-US/docs/Web/API/DOMException).

core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn signal_abort(&self, reason: JsValue, context: &mut Context) -> JsResult<()> {
+        if self.aborted.get() {
+            return Ok(());
+        }
+        self.aborted.set(true);
+        *self.reason.borrow_mut() = Some(reason);
+
+        let listeners: Vec<JsFunction> = self.listeners.borrow().clone();
+        for listener in &listeners {
+            listener.call(&JsValue::undefined(), &[], context)?;
```
hansl: The spec says that ALL listeners should be run independently. The current code would stop at the first listener that throws.

What this should do is setup an immediate job on the context for each listeners and allow them to run that way. Exceptions in jobs are handled separately. Makes sense?

I'm using this test in Node as reference:

```js
a = AbortSignal.timeout(1000);
a.addEventListener('abort', () => { throw 1; });
a.addEventListener('abort', () => { console.log(2); });
a.addEventListener('abort', () => { throw 2; });
```

The output is:

```
2
Uncaught 1
Uncaught 2
>
```

core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Debug, Clone, Default, JsData, Trace, Finalize)]
+pub struct JsAbortSignal {
+    #[unsafe_ignore_trace]
+    aborted: Cell<bool>,
+    reason: GcRefCell<Option<JsValue>>,
+    listeners: GcRefCell<Vec<JsFunction>>,
+}
+
+impl JsAbortSignal {
+    fn signal_abort(&self, reason: JsValue, context: &mut Context) -> JsResult<()> {
```
hansl: This function should likely be public, no? So hosts can use it when creating abort signals.

core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+            .clone()
+            .unwrap_or_else(|| js_string!("AbortError").into())
+    }
+}
+
+#[boa_class(rename = "AbortSignal")]
+#[boa(rename_all = "camelCase")]
+impl JsAbortSignal {
+    #[boa(constructor)]
+    fn constructor() -> Self {
```
hansl: The spec has an error in the constructor. It's illegal to construct an `AbortSignal`.

core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    fn throw_if_aborted(&self) -> JsResult<()> {
+        if self.aborted.get() {
+            Err(JsError::from_opaque(self.abort_reason()))
+        } else {
+            Ok(())
+        }
+    }
+
+    fn add_event_listener(
```
hansl, line 155: There should be a `removeEventListener` too.

core/runtime/src/abort/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+
+impl JsAbortSignal {
+    fn signal_abort(&self, reason: JsValue, context: &mut Context) -> JsResult<()> {
+        if self.aborted.get() {
+            return Ok(());
+        }
+        self.aborted.set(true);
+        *self.reason.borrow_mut() = Some(reason);
+
+        let listeners: Vec<JsFunction> = self.listeners.borrow().clone();
```
hansl: Should drain the vector of listeners so we don't keep references to functions that aren't used anymore. This can only be called once, but it might live longer if the JS code keeps a copy of it.

core/runtime/src/fetch/mod.rs
```diff
@@ -141,6 +161,9 @@ async fn fetch_inner<T: Fetcher>(
     }
 
     let response = fetcher.fetch(JsRequest::from(request), context).await?;
```
hansl: The abort signal should be sent to the fetcher as it can actually cancel it when the signal is triggered. Right now if I'm downloading a 1GB file and abort in the middle I keep downloading the whole file before aborting it.

core/runtime/src/abort/tests.rs
```diff
@@ -0,0 +1,158 @@
+use crate::test::{TestAction, run_test_actions};
```
hansl, line 1: Add a test for no reason given. It should be `undefined` per the spec.

hansl: Sorry I meant add a test for checking the reason when the signal is not aborted yet. If it has been aborted without a reason it should be `AbortError`.

Nakshatra480:
Thanks for the thorough review, @hansl , I've pushed an update addressing all your points: the reason getter is now spec-compliant, listeners are dispatched independently as non-blocking GenericJobs with dead references drained, and signal_abort is now public. I also added the missing tests, properly implemented removeEventListener using pointer equality, and wired abort cancellation down to the fetchers.

core/runtime/src/abort/mod.rs
```diff
@@ -23,27 +47,56 @@ fn make_abort_error(context: &mut Context) -> JsValue {
...
-        for listener in &listeners {
-            listener.call(&JsValue::undefined(), &[], context)?;
+        let listeners: Vec<JsFunction> = self.listeners.borrow_mut().drain(..).collect();
+
+        let realm = context.realm().clone();
+        for listener in listeners {
+            context.enqueue_job(
+                GenericJob::new(
+                    move |context| {
+                        drop(listener.call(&JsValue::undefined(), &[], context));
```
hansl: In this case, because this is a job, you likely should propagate the error. It will not prevent other jobs from running.

hansl (commented):
One last change. Good work!

core/runtime/src/abort/mod.rs
Nakshatra480: Good catch, thanks! Just pushed a fix to propagate the listener error instead of dropping it. All tests are passing.

Nakshatra480:
> One last change. Good work!

Thanks @hansl

Nakshatra480:
Thanks Jedel and Hans

PR #4837, opened 2026-03-03
fix(jsarray): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241

Converts all 14 panics in `jsarray.rs` to use `js_expect` introduced in #4828, replacing `.expect()` calls that would crash the process with recoverable `EngineError::Panic` errors that bubble up to the host application.

Changes:
- `JsArray::new` now returns `JsResult<Self>` instead of `Self`
- All 14 `.expect()` calls replaced with `.js_expect()?`
- Updated all callers of `JsArray::new` across the codebase
- Fixed doc examples in `jsmap.rs` and `jsobject.rs`

### DIFF
--- a/examples/src/bin/jsarray.rs
+++ b/examples/src/bin/jsarray.rs
@@ -11,7 +11,7 @@ fn main() -> JsResult<()> {
     let context = &mut Context::default();
 
     // Create an empty array.
-    let array = JsArray::new(context);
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Looks great, thank you!

KaustubhOG:
@jedel1043 thanks

PR #4841, opened 2026-03-03
fix: Make StdClock monotonic using std::time::Instant

### LINKED ISSUE #4330: Implementation of `StdClock` is not monotonic.

abhinavs1920:
This Pull Request fixes/closes #4330 .

Previously, `StdClock` used `SystemTime` which isn't monotonic—it can jump backward when the system clock adjusts. This broke the guarantees that `JsInstant` is supposed to provide for engine timing.

Now `StdClock` uses `std::time::Instant` with a base instant, measuring elapsed time from that point. This makes `JsInstant` truly monotonic like `std::time::Instant`, so time never goes backward for `setTimeout`, `setInterval`, and job scheduling.

For `Date` objects that need actual wall-clock time, I added a separate `system_time_millis()` method to the `Clock` trait. Clean separation of concerns.

Added tests to verify the monotonic behavior works as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2459,10 +2459,6 @@ version = "1.0.0-dev"
 dependencies = [
  "boa_engine",
  "insta",
- "cfg-if",
- "js-sys",
- "wasm-bindgen",
- "web-sys",
 ]
 
 [[package]]
--- a/core/engine/src/context/time.rs
+++ b/core/engine/src/context/time.rs
@@ -160,7 +160,7 @@ pub trait Clock {
 
 /// A clock that uses the standard monotonic clock.
 ///
-/// This clock is based on [`web_time::Instant`] (via `crate::sys::time`) which provides
+/// This clock is based on the `instant` crate's `Instant` type, which provides
 /// cross-platform monotonic time, including WASM support via `performance.now()`.
 /// Time measurements are relative to an arbitrary starting point
 /// (the first call to `now()`) and are not affected by system clock adjustments.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: "\U0001F41B Bug report"
 about: Create a report to help us improve
 title: ""
-labels: bug
+type: "Bug"
 assignees: ""
 ---
 
--- a/.github/ISSUE_TEMPLATE/custom.md
+++ b/.github/ISSUE_TEMPLATE/custom.md
@@ -1,8 +1,9 @@
 ---
 name: Custom
-about: Open an issue in the repo that is neither a bug or a feature, such a new idea
+about: Open an issue in the repo that is neither a bug or a feature.
 title: ""
 labels: ""
+type: ""
 assignees: ""
 ---
 
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,8 @@
 name: "\U0001F680 Feature request"
 about: Suggest a new ECMAScript feature to be implemented, or a new capability of the engine.
 title: ""
-labels: enhancement
+type: "Feature"
+labels: ""
 assignees: ""
 ---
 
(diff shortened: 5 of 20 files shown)


abhinavs1920:
Heyy @jedel1043 , I have refactored the API design as we discussed. Please take a look when ever you are available :)

abhinavs1920:
Heyy I have put pushed my latest commit, but it doesn't seem to be pushed here due to some github issue. 🥲

core/engine/src/context/time.rs
```diff
@@ -1,5 +1,7 @@
 //! Clock related types and functions.
 
+use instant::Instant;
```
jedel1043: We already depend on `web-time`, and we have all the bindings on `crate::sys::time`, so you should be able to import `crate::sys::time::Instant`

core/engine/Cargo.toml
```diff
@@ -139,6 +139,7 @@ either = { workspace = true, optional = true }
 static_assertions.workspace = true
 futures-channel.workspace = true
 aligned-vec.workspace = true
+instant.workspace = true
```
jedel1043: Need to remove this dependency

Cargo.toml
```diff
@@ -99,6 +99,7 @@ bus = "2.4.1"
 wasm-bindgen = { version = "0.2.97", default-features = false }
 getrandom = { version = "0.4.1", default-features = false }
 console_error_panic_hook = "0.1.7"
+instant = { version = "0.1", features = ["wasm-bindgen"] }
```
jedel1043: Also here, needs to be removed

jedel1043 (changes requested):

Cargo.lock
jedel1043, line 1: please update the lock file

abhinavs1920: Updated Cargo.lock using cargo update. Does this look right?

core/engine/src/context/time.rs
```diff
@@ -131,22 +145,62 @@ impl std::ops::Sub for JsInstant {
...
+    /// Returns the current wall-clock time in milliseconds since the Unix epoch.
+    ///
+    /// This is NOT monotonic and can go backward if the system clock is adjusted.
+    /// It should only be used for `Date` objects and other wall-clock time needs.
+    fn system_time_millis(&self) -> i64;
+}
+
+/// A clock that uses the standard monotonic clock.
+///
+/// This clock is based on [`instant::Instant`] which provides cross-platform
```
jedel1043: You should fix the documentation to reflect that we're using `web-time`.

core/engine/src/context/time.rs
```diff
@@ -131,22 +145,62 @@ impl std::ops::Sub for JsInstant {
 
 /// Implement a clock that can be used to measure time.
 pub trait Clock {
-    /// Returns the current time.
+    /// Returns the current monotonic time.
+    ///
+    /// This is guaranteed to be monotonic and should be used for measuring
+    /// durations and scheduling timeouts.
```
jedel1043: Slight adjustment here. This is not guaranteed to be monotonic, the engine assumes it is monotonic, so the implementers of the trait are the ones that need to ensure the monotonicity of the returned `JsInstant`.

abhinavs1920: Thanks for the missed detail.. :)

PR #4842, opened 2026-03-03
fix(jsdataview): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241

Converts all 12 panics in `jsdataview.rs` to use `js_expect` introduced
in #4828, replacing `.expect()` calls that would crash the process with
recoverable `EngineError::Panic` errors that bubble up to the host
application.

No public API changes — all methods already returned `JsResult<T>`,
only the internal implementation was updated.


KaustubhOG:
image
 hey @jedel1043 is that internal server error is because of github or any mistake  fromm my side

nekevss:
GitHub has been having some service issues. 500 points to GitHub

PR #4845, opened 2026-03-03
perf: eliminate unnecessary Move instructions via register forwarding

hansl:
For local variable references in binary expressions, update expressions, and for-loop iteration bindings, directly reference the local's persistent register instead of emitting a Move into a temporary. This avoids redundant copies for operations like `i < n`, `i++`, and per-iteration let binding copies when all bindings are local.

Changes:
- Add `compile_expr_operand` to ByteCompiler that returns a VaryingOperand pointing directly at a local binding's register when possible
- Refactor binary operation compilation (arithmetic, bitwise, relational) to use operand forwarding for both LHS and RHS
- Add fast path for pre-increment/decrement on local bindings in update expressions (2 ops instead of 4)
- Skip no-op per-iteration binding copy in for-loops when all bindings are local (scope_index is None)

Profiling shows handle_move dropped from 25.5% to 15.0% of execution time on a nested for-loop benchmark.

### DIFF
--- a/core/engine/src/bytecompiler/mod.rs
+++ b/core/engine/src/bytecompiler/mod.rs
@@ -1450,7 +1450,7 @@ impl<'ctx> ByteCompiler<'ctx> {
     /// directly without emitting a `Move` instruction. For all other expressions,
     /// it allocates a temporary register and compiles into it.
     ///
-    /// The caller must deallocate the returned `Register` (if `Some`) after use.
+    /// The `inner_fn` passed in will be called before the register get deallocated.
     pub(crate) fn compile_expr_operand(
         &mut self,
         expr: &Expression,


core/engine/src/bytecompiler/mod.rs
```diff
@@ -1444,6 +1444,31 @@ impl<'ctx> ByteCompiler<'ctx> {
...
+    /// Compile an expression and return the operand where the result lives.
+    ///
+    /// For local variable references, this returns the local's persistent register
+    /// directly without emitting a `Move` instruction. For all other expressions,
+    /// it allocates a temporary register and compiles into it.
+    ///
+    /// The caller must deallocate the returned `Register` (if `Some`) after use.
+    pub(crate) fn compile_expr_operand(
+        &mut self,
+        expr: &Expression,
```
jedel1043, line 1456: Here it might be cleaner to take `impl FnOnce(&mut ByteCompiler<'_>, VaryingOperand)` and call that inside the function. That way you don't require having to deallocate manually.

hansl: Done, PTAL.

core/engine/src/bytecompiler/mod.rs
```diff
@@ -1444,6 +1444,34 @@ impl<'ctx> ByteCompiler<'ctx> {
         self.compile_expr_impl(expr, dst);
     }
 
+    /// Compile an expression and return the operand where the result lives.
+    ///
+    /// For local variable references, this returns the local's persistent register
+    /// directly without emitting a `Move` instruction. For all other expressions,
+    /// it allocates a temporary register and compiles into it.
+    ///
+    /// The caller must deallocate the returned `Register` (if `Some`) after use.
```
jedel1043: Docs need to be updated

hansl:
```suggestion
    /// The `inner_fn` passed in will be called before the register get deallocated.
```

PR #4846, opened 2026-03-03
fix(jsregexp): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241

Converts all 17 panics in `jsregexp.rs` to `EngineError::Panic` using
`js_expect` introduced in #4828.

No public API signatures were changed — only 1 file modified.

Patterns converted:
- `.as_object().expect()` → `.as_object().js_expect()?`
- `.map(|v| v.as_boolean().expect())` → `.and_then(|v| v.as_boolean().js_expect().map_err(Into::into))`
- `.map(|v| v.as_string().expect()...to_std_string().expect())` → `.and_then(|v| v.as_string().js_expect()?.to_std_string().map_err(...))`
- nested `.expect()` calls in `exec()` converted using `?` inside `.and_then()`

### DIFF
--- a/core/engine/src/object/builtins/jsregexp.rs
+++ b/core/engine/src/object/builtins/jsregexp.rs
@@ -1,7 +1,8 @@
 //! A Rust API wrapper for Boa's `RegExp` Builtin ECMAScript Object
 use crate::{
-    Context, JsNativeError, JsResult, JsValue,
+    Context, JsExpect, JsNativeError, JsResult, JsValue,
     builtins::RegExp,
+    error::PanicError,
     object::{JsArray, JsObject},
     value::TryFromJs,
 };
@@ -62,7 +63,7 @@ impl JsRegExp {
     {
         let regexp = RegExp::initialize(None, &pattern.into(), &flags.into(), context)?
             .as_object()
-            .expect("RegExp::initialize must return a RegExp object")
+            .js_expect("RegExp::initialize must return a RegExp object")?
             .clone();
 
         Ok(Self { inner: regexp })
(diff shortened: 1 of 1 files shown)


core/engine/src/object/builtins/jsregexp.rs
```diff
@@ -146,11 +167,11 @@ impl JsRegExp {
...
     #[inline]
     pub fn flags(&self, context: &mut Context) -> JsResult<String> {
-        RegExp::get_flags(&self.inner.clone().into(), &[], context).map(|v| {
+        RegExp::get_flags(&self.inner.clone().into(), &[], context).and_then(|v| {
             v.as_string()
-                .expect("value must be string")
+                .js_expect("value must be string")?
                 .to_std_string()
-                .expect("flags must be a valid string")
+                .map_err(|e| JsNativeError::typ().with_message(e.to_string()).into())
```
jedel1043: Should probably use `PanicError` here right? And pass `e.to_string()` as the error panic

core/engine/src/object/builtins/jsregexp.rs
```diff
@@ -238,11 +262,11 @@ impl JsRegExp {
...
     #[inline]
     pub fn to_string(&self, context: &mut Context) -> JsResult<String> {
-        RegExp::to_string(&self.inner.clone().into(), &[], context).map(|v| {
+        RegExp::to_string(&self.inner.clone().into(), &[], context).and_then(|v| {
             v.as_string()
-                .expect("value must be a string")
+                .js_expect("value must be a string")?
                 .to_std_string()
-                .expect("to_string value must be a valid string")
+                .map_err(|e| JsNativeError::typ().with_message(e.to_string()).into())
```
nekevss: nit: JsError, not TypeError

core/engine/src/object/builtins/jsregexp.rs
```diff
@@ -171,11 +192,11 @@ impl JsRegExp {
...
     #[inline]
     pub fn source(&self, context: &mut Context) -> JsResult<String> {
-        RegExp::get_source(&self.inner.clone().into(), &[], context).map(|v| {
+        RegExp::get_source(&self.inner.clone().into(), &[], context).and_then(|v| {
             v.as_string()
-                .expect("value must be string")
+                .js_expect("value must be string")?
                 .to_std_string()
-                .expect("source must be a valid string")
+                .map_err(|e| JsNativeError::typ().with_message(e.to_string()).into())
```
nekevss: nit: JsError, not TypeError

core/engine/src/object/builtins/jsregexp.rs
nekevss: nit: JsError, not TypeError

nekevss (changes requested):

core/engine/src/object/builtins/jsregexp.rs
```diff
@@ -209,14 +234,14 @@ impl JsRegExp {
...
             if v.is_null() {
-                None
+                Ok(None)
             } else {
-                Some(
-                    JsArray::from_object(v.to_object(context).expect("v must be an array"))
-                        .expect("from_object must not fail if v is an array object"),
-                )
+                let obj = v.to_object(context).js_expect("v must be an array")?;
+                let array = JsArray::from_object(obj)
```
nekevss, line 242: I think it's possible to do this in a slightly cleaner way.

This is off the top of my head so it may not be 100% accurate. But something like:

```rust
v.to_object(context)
  .js_expect("value must be an array")
  .and_then(JsArray::from_object)
  .js_expect("from_object must not fail if value is an array object")
  .map(Some)
```

KaustubhOG: yup that looks much more cleaner , let me try to structure it like that

KaustubhOG: Thanks!  @nekevss  Tried this but it doesn't compile — after `.js_expect` the error type is 
`PanicError`, but `JsArray::from_object` returns `Result<JsArray, JsError>`, so 
`.and_then` fails due to mismatched error types. so i Kept the intermediate variables instead.

nekevss (approved):

PR #4847, opened 2026-03-03
Implement RISC-style opcodes for `super`-related accesses

jedel1043:
Instead of having two big opcodes doing a lot of things, this PR splits them into smaller opcodes, delegating the construction of the bigger operation to the bytecompiler.

The plan with this is to eventually remove the `SuperCall` opcodes, but that's for another PR.

### DIFF
--- a/core/engine/src/bytecompiler/mod.rs
+++ b/core/engine/src/bytecompiler/mod.rs
@@ -1127,7 +1127,7 @@ impl<'ctx> ByteCompiler<'ctx> {
     pub(crate) fn jump_if_neq(&mut self, lhs: &Register, rhs: &Register) -> Label {
         let index = self.next_opcode_location();
         self.bytecode
-            .emit_jump_if_neq(Self::DUMMY_ADDRESS, lhs.variable(), rhs.variable());
+            .emit_jump_if_not_equal(Self::DUMMY_ADDRESS, lhs.variable(), rhs.variable());
         Label { index }
     }
 


hansl:
I'm going to need a bit of time to see how that affects the bytecode. Having more instructions might actually be bad for performance in this case.

jedel1043:
~Kind of? The previous instructions would access the function environment multiple times, since getting `super` and getting `this` were two separate instructions. By splitting them into small but general functions, the VM just needs to access the function environment once, and the rest is just moving values between registers and calling methods~
Doing it this way was not correct, so now it's calling `This` and `GetFunctionObject`. I still think it's better because `This` caches its result, but `Super` needed to go through the environment stack every time.

nekevss:
FWIW, I like the idea of more RISC-style while mostly with the general concept of simplification being better. It may give us a better idea of what can be optimize and/or slow operations.

But I'm also not neck deep in the opcode and VM like @hansl is right now.

jedel1043:
Yep pretty much. The idea is that right now we have a lot of big opcodes that might or might not be that useful as a standalone opcode. Having more RISC style instructions would let us see which repeated patterns the bytecode has, then we can profile against that to specialize on those cases instead

hansl:
The issue is that right now we spend too much time in decoding instructions and their operands and branching off the jump table. So for this to make sense I'll have to do some benchmarking. 

I'm actually going the opposite direction on common instructions so we spend less time decoding and branching, for example removing LessThan + JumpIfFalse in favor of JumpIfNotLessThan.

jedel1043:
doing that for simple common operations like `JumpIfNotLessThan` makes sense because it's a very common operation that is used almost everywhere

`super` is more like a "superop", where it does so many things that even specializing it shouldn't considerably affect the performance (I think), because we're kinda bottle-necked by the operations it does inside. Splitting it will at least give us the opportunity to cache or optimize its smaller parts in a more granular way

PR #4850, opened 2026-03-03
perf: reduce inner loop dispatch overhead with post-increment fast path and fused comparison-branch opcodes

hansl:
This is on top of #4848.

Two optimizations targeting the inner loop hot path:

1. Post-increment fast path for local bindings (update.rs):
   j++ on locals now emits Move(dst, local) + Inc(local, local) = 2 ops
   instead of Move(dst, local) + Inc(tmp, dst) + Move(local, tmp) = 3 ops.

2. Fused comparison-branch opcodes (JumpIfNotLessThan, etc.):
   Combines relational comparison + conditional jump into a single opcode,
   eliminating one dispatch and one intermediate boolean per iteration.
   Applied automatically to for/while/do-while loop conditions.

Together these reduce the inner loop from 7 to 5 opcodes per iteration.

----

With this, it brings us within 1.1x of QuickJS on an empty nested loop benchmark:

```
Benchmark 1: ./target/release/boa /Users/hansl/a.js
  Time (mean ± σ):      1.891 s ±  0.006 s    [User: 1.878 s, System: 0.008 s]
  Range (min … max):    1.886 s …  1.907 s    10 runs

Benchmark 2: qjs /Users/hansl/a.js
  Time (mean ± σ):      1.696 s ±  0.005 s    [User: 1.688 s, System: 0.006 s]
  Range (min … max):    1.689 s …  1.705 s    10 runs

Summary
  qjs /Users/hansl/a.js ran
    1.11 ± 0.00 times faster than ./target/release/boa /Users/hansl/a.js
```

### DIFF
--- a/core/engine/src/bytecompiler/expression/update.rs
+++ b/core/engine/src/bytecompiler/expression/update.rs
@@ -44,7 +44,10 @@ impl ByteCompiler<'_> {
                 //
                 // Skip for const bindings — they must fall through to emit ThrowMutateImmutable.
                 if is_lexical
-                    && compiler.lexical_scope.set_mutable_binding(name.clone()).is_ok()
+                    && compiler
+                        .lexical_scope
+                        .set_mutable_binding(name.clone())
+                        .is_ok()
                     && let BindingKind::Local(Some(local_reg)) = &index
                 {
                     let local_op = (*local_reg).into();


core/engine/src/bytecompiler/expression/update.rs
```diff
@@ -51,6 +51,30 @@ impl ByteCompiler<'_> {
...
 
+                // Fast path: for local bindings with post-increment/decrement.
+                //
+                // Post-increment (i++):
+                //   Before: Move(dst, local); Inc(tmp, dst); Move(local, tmp) → 3 ops
+                //   After:  Move(dst, local); Inc(local, local) → 2 ops
+                //
+                // Inc(local, local) works because Inc writes new to dst AFTER old to src,
+                // so when dst==src the new value wins.
+                if post
```
jedel1043: You should be able to combine this block with the previous one instead of duplicating the if check

hansl:
@jedel1043 PTAL

core/engine/src/bytecompiler/statement/loop.rs
```diff
@@ -128,14 +128,15 @@ impl ByteCompiler<'_> {
...
+        let exit = if let Some(condition) = for_loop.condition() {
+            self.compile_condition_and_branch(condition)
         } else {
+            let value = self.register_allocator.alloc();
             self.bytecode.emit_push_true(value.variable());
-        }
-        let exit = self.jump_if_false(&value);
-        self.register_allocator.dealloc(value);
+            let exit = self.jump_if_false(&value);
+            self.register_allocator.dealloc(value);
```
jedel1043: Uhh I'm kinda confused by this. Isn't it the same as a no-op? It's setting up a register to push true in it, then jump if false, which will never happen since the register is always true.

hansl: Some refactor (from the previous code)/merge mishap, it _looks_ noop to me too. I'm checking if I can remove it and still pass the tests.

hansl:
@jedel1043 did another read through of the PR and didn't see anything outstanding. I have another PR after this one to address useless Move when the result is discarded, so please don't address it here :)

jedel1043 (approved):
LGTM!

PR #4852, opened 2026-03-04
Make jump tables fast

jedel1043:
Our current code to execute jump tables is kind of wonky; it iteratively goes through all the entries of the jump table until it finds the desired index. The intended way of using a jump table is to just use the value as an index into the table, then jump to the returned pc value.

This PR also removes the `default` argument of jump tables. The fallback is usually to just not jump and keep executing the next instruction, we don't need to explicitly add a default path.

### DIFF
--- a/core/engine/Cargo.toml
+++ b/core/engine/Cargo.toml
@@ -123,7 +123,7 @@ thiserror.workspace = true
 dashmap.workspace = true
 num_enum.workspace = true
 thin-vec.workspace = true
-itertools = { workspace = true, default-features = false }
+itertools = { workspace = true, default-features = false, features = ["use_alloc"] }
 icu_normalizer = { workspace = true, features = [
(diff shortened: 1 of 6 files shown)


jedel1043:
Kinda nice improvements all over the suite

# main

RESULT Richards 305
RESULT DeltaBlue 245
RESULT Crypto 282
RESULT RayTrace 508
RESULT EarleyBoyer 629
RESULT RegExp 78.8
RESULT Splay 1170
RESULT NavierStokes 700
SCORE 380

# pr

RESULT Richards 313
RESULT DeltaBlue 248
RESULT Crypto 287
RESULT RayTrace 531
RESULT EarleyBoyer 642
RESULT RegExp 81.9
RESULT Splay 1189
RESULT NavierStokes 711
SCORE 389

hansl (approved):

PR #4853, opened 2026-03-04
feature(display): Implement pretty printing of arguments inside of console.log (closes #1437)

### LINKED ISSUE #1437: Pretty print `Arguments` objects
**Describe the bug**
Using `console.log` on the `arguments` object inside a function produces a type error

> Uncaught "TypeError": "cannot convert object to primitive value"

**To Reproduce**
```javascript
function test() {
   console.log(arguments)
}
test('first', 'second')
```

**Expected behavior**

Running this code should log the `arguments` object to the console.

In Firefox the output looks like this:
```
Arguments { 0: "first", 1: "second", … }
```

In Node the output looks like this:
```
[Arguments] { '0': 'first', '1': 'second' }
```

- OS: Ubuntu
- Version: 20.04
- Rustc version: rustc 1.53.0 (53cb7b09b 2021-06-17)

**Additional context**
I believe that the `arguments` object is missing the `Object` prototype, so it's missing [Object.prototype.valueOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf) which `console.log` is trying to invoke.

gluzandii:
This pull request closes: #1437.

Firstly, it restructures the `core::boa_engine::value::display` module into different files. Initially the display module was a single file with everything in it, this made it hard to read and maintain, I have split it up into different modules inside of display, each pertaining to a certain data type for formatting (array, map, set etc).

Added compact formatting of certain Js types, so that it can be used in console logging the arguments object:

-  [array.rs](https://github.com/gluzandii/boa-fork/blob/feature/prettyprintarguments/core/engine/src/value/display/array.rs#L68)
-  [map.rs](https://github.com/gluzandii/boa-fork/blob/feature/prettyprintarguments/core/engine/src/value/display/map.rs#L39)
-  [set.rs](https://github.com/gluzandii/boa-fork/blob/feature/prettyprintarguments/core/engine/src/value/display/set.rs#L37)
-  [object.rs](https://github.com/gluzandii/boa-fork/blob/feature/prettyprintarguments/core/engine/src/value/display/object.rs#L159)

Along with the compact formatting methods that are used for logging arguments, there are also methods for regular logging (not compact) for the same data types as well, present in the same file.

This PR also added ability to log:
- Regex -> In the form "/{regex pattern}/{flags}"
- Date -> In the ISO 8601 format

The above 2 changes have been added for formatting in general (not only restricted to how arguments object is displayed)

Added method in [value.rs](https://github.com/gluzandii/boa-fork/blob/feature/prettyprintarguments/core/engine/src/value/display/value.rs#L163) called `log_value_compact` which is used for compact formatting of objects. This is what is internally used when the `console.log(arguments)` is called.

Compact formatting is the same as regular formatting but smartly compacts deeply nested types (like objects, arrays to just mentioning what type they are). It does this by keeping track of the depth. The current max depth is 2.

This can be changed by modifying the constant `COMPACT_DEPTH_LIMIT` constant inside of [value.rs](https://github.com/gluzandii/boa-fork/blob/feature/prettyprintarguments/core/engine/src/value/display/value.rs#L25).

### DIFF
--- a/core/engine/src/value/display/arguments.rs
+++ b/core/engine/src/value/display/arguments.rs
@@ -1,7 +1,11 @@
+use crate::builtins::function::arguments::MappedArguments;
+use crate::property::DescriptorKind;
 use crate::value::display::value;
 use crate::{JsObject, JsValue, js_string};
 use std::collections::HashSet;
 use std::fmt::{self, Write};
+
+const MAX_ARGUMENTS_TO_LOG: u32 = 100;
 
 /// Formats an Arguments object for display.
 ///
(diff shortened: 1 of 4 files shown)


gluzandii:
I force pushed since my branch was very behind the main branch, hence updated its base commit to the latest.

jedel1043:
Need to address all copilot comments first

PR #4855, opened 2026-03-04
fix(module): drop AST and source text after module initialization

ashnaaseth2325-oss:
# Summary

This PR fixes a memory retention issue in the module system where the **parsed AST (`boa_ast::Module`) and the module source text were kept in memory even after compilation finished**.

These structures are only required during the **module initialization phase** (`initialize_environment`). However, they were previously stored as permanent fields inside `ModuleCode`, meaning every loaded module kept its entire AST and source text allocated for the lifetime of the process.

In workloads that load many modules, this can lead to **steady memory growth over time**.

---

# Where the Issue Occurred

The problem originates in:

```
core/engine/src/module/source.rs
```

`ModuleCode` previously stored the full AST and source text directly:

```rust
struct ModuleCode {
    has_tla: bool,
    requested_modules: IndexSet<ModuleRequest>,
    source: boa_ast::Module,
    source_text: SourceText,
}
```

Both values are only used inside `SourceTextModule::initialize_environment()`, but they remained allocated afterward.

Since modules are cached by `SimpleModuleLoader`, the AST and source text would stay in memory indefinitely.

---

# Fix

The AST and source text are now stored as optional values so they can be **consumed and dropped once compilation is complete**.

```rust
source: RefCell<Option<boa_ast::Module>>,
source_text: RefCell<Option<SourceText>>,
```

They are extracted during initialization:

```rust
let source = self.code.source.borrow_mut().take()
    .expect("module source consumed before initialize_environment");
```

After this point, the AST and source text are no longer retained by the module.

---

# Impact

This prevents modules from permanently holding large AST structures that are no longer needed after compilation.

In module-heavy workloads (such as large dependency graphs or repeated module loading), this **keeps memory usage stable instead of growing with every loaded module**.

---

# Result

* AST memory is released immediately after module linking
* Modules retain only the data needed for execution
* Long-running Boa processes avoid unnecessary memory growth

### DIFF
--- a/core/engine/src/module/source.rs
+++ b/core/engine/src/module/source.rs
@@ -1565,13 +1565,11 @@ impl SourceTextModule {
         let source = self
             .code
             .source
-            .borrow_mut()
             .take()
             .expect("module source consumed before initialize_environment");
         let source_text = self
             .code
             .source_text
-            .borrow_mut()
             .take()
             .expect("module source_text consumed before initialize_environment");
 


core/engine/src/module/source.rs
```diff
@@ -1560,19 +1560,33 @@ impl SourceTextModule {
         // 4. Assert: realm is not undefined.
         let realm = module_self.realm().clone();
 
+        // Take the AST and source text — they are only needed during compilation.
+        // Dropping them here frees the parse tree after linking is complete.
+        let source = self
+            .code
+            .source
+            .borrow_mut()
```
hansl: You should be able to just call `take()` on the RefCell here. It is implemented when default is implemented which is the case for an Option.

ashnaaseth2325-oss:
Hello @hansl @jedel1043 
Updated the implementation to use **RefCell::take()** instead of **borrow_mut().take()**. Thanks for pointing that out.

jedel1043:
This kinda makes it impossible to re-evaluate a module again. Might be fine I guess

PR #4857, opened 2026-03-04
ci: test262 actions bot comments main commit too

zhuzhu81998:
Yet another upgrade for the test 262 actions bot. Currently, only PR commit is shown.
It is possible that the base commit on main is not immediately clear without this PR (e.g. if main has a newer commit than the PR's latest one)

Demo at https://github.com/zhuzhu81998/boa/pull/3:
image

Made possible by #4792 (so that the `latest.json` would contain the correct commit, which is read here via `jq` which is by default included in the runner environment).


jedel1043 (approved):
Thanks! This is very cool :)

Issue #2675, opened 2023-03-16
Evaluate/Reevaluate the execution semantics of `CompletionRecord`


nekevss:
As mentioned in PR #2618, we have previously been evaluating `Return` statements as normal completions and `Yield` statements a non-normal completion (`ReturnType::Yield` and now `CompletionRecord::Return`). This should be a different case with `Yield` statements being normal completions (`CompletionRecord::Normal`) and return statements being return completions (`CompletionRecord::Return`).

Completing the above change appears to require more evaluation/reevaluation of `Generator` executions and/or `CompletionRecord` semantics in the `Vm` as when the change was initially attempted, a number of `test262` tests failed.

Nakshatra480:
Hey @nekevss ! I've analyzed issue #2675 and would love to work on it. I've gone through the ECMAScript spec (§27.5 Generator Objects, §14.4 Yield) and have a clear understanding of the problem.
Root Cause: Yield is incorrectly emitting CompletionRecord::Return (non-normal) and Return is emitting CompletionRecord::Normal - both are swapped from what the spec requires.
My proposed approach - breaking this into 5 focused sub-issues:

- Audit: Map all CompletionRecord emit sites across vm/opcode/ and builtins/generator/ to understand the full scope
- Fix Yield opcode: Change it to emit CompletionRecord::Normal as per spec §27.5.3.7
- Fix Return opcode: Change it to emit CompletionRecord::Return to correctly mark generator as completed
- Update Generator drive loop: Revise the resumption logic to handle the corrected completion variants (Normal → suspend & yield value, Return → mark done)
- test262 regressions: Run generator test suite, fix failures, add Boa-specific regression tests

I'll start with the audit first and share findings before touching any opcodes, so we can align on approach early. 
Could you please assign this issue to me? Thanks!

PR #4861, opened 2026-03-04
docs(vm): audit CompletionRecord semantics for Generator Yield/Return

Nakshatra480:
This PR is the first step toward resolving #2675.

Before changing any opcode logic, this audit documents every site in the VM
where `CompletionRecord` variants are produced or consumed in Generator context,
with inline `FIXME(#2675)` comments explaining the current vs. spec-correct behavior.

No logic is changed in this PR - only documentation/comments are added.
This makes the actual fix PRs easier to review and gives maintainers visibility
into the full scope of the problem.

Refs: #2675
Spec: ECMAScript §27.5 (Generator Objects), §14.4 (Yield), §27.5.3.3, §27.5.3.7


core/engine/src/builtins/generator/mod.rs
```diff
@@ -2,6 +2,14 @@
 //!
 //! A Generator is an instance of a generator function and conforms to both the Iterator and Iterable interfaces.
 //!
+//! # `CompletionRecord` Semantics (Issue #2675)
+//!
+//! The VM currently uses `CompletionRecord` variants with inverted semantics relative to
+//! the ECMAScript spec for generator execution: `CompletionRecord::Return` is used to signal
+//! a yield suspension (not a return), and `CompletionRecord::Normal` is used for generator
+//! completion (including `return` statements). This diverges from spec §27.5.3.3 and §27.5.3.4,
+//! which define return completions and normal completions with their standard meanings.
```
jedel1043: Since we're treating this behaviour as a bug, it shouldn't really be in the documentation

Nakshatra480: Make sense i just removed it from the module-level docs.
Thanks for your feedback😊

Nakshatra480:
Hey @jedel1043 i have removed the module level docs as you suggested, can you review this PR now. Happy to address any feedback.
Thanks for your time

jedel1043:
This is kinda superseeded by #4880, right? Then we can probably close this and just fix the issue directly

PR #4864, opened 2026-03-04
perf: optimize loop update expressions and comparison fast paths

hansl:
- Add `discard` parameter to `compile_update` to skip dead Move opcodes when the result of an update expression is unused (e.g., `i++` in `for` loop final expressions)
- Return `bool` directly from comparison fast paths (`lt_fast`, `le_fast`, `gt_fast`, `ge_fast`) instead of wrapping in `JsValue`, avoiding unnecessary `.to_boolean()` calls in fused comparison-branch opcodes

### DIFF
--- a/core/engine/src/bytecompiler/expression/mod.rs
+++ b/core/engine/src/bytecompiler/expression/mod.rs
@@ -91,7 +91,7 @@ impl ByteCompiler<'_> {
                 );
             }
             Expression::Unary(unary) => self.compile_unary(unary, dst),
-            Expression::Update(update) => self.compile_update(update, dst),
+            Expression::Update(update) => self.compile_update(update, dst, false),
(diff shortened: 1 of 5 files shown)


hansl:
The results from the latest nested loop benchmark:

```
Benchmark 1: ./target/release/boa /Users/hansl/a.js
  Time (mean ± σ):      1.916 s ±  0.007 s    [User: 1.901 s, System: 0.008 s]
  Range (min … max):    1.910 s …  1.929 s    10 runs

Benchmark 2: qjs /Users/hansl/a.js
  Time (mean ± σ):      1.721 s ±  0.020 s    [User: 1.710 s, System: 0.007 s]
  Range (min … max):    1.702 s …  1.755 s    10 runs

Summary
  qjs /Users/hansl/a.js ran
    1.11 ± 0.01 times faster than ./target/release/boa /Users/hansl/a.js
```

PR #4868, opened 2026-03-05
chore: bump test262 version

nekevss:
Bumps the test262 version to a more recent commit + add new feature.

### DIFF
--- a/test262_config.toml
+++ b/test262_config.toml
@@ -7,8 +7,6 @@ flags = []
 features = [
     ### Unimplemented features:
 
-    "import-bytes",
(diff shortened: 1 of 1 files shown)


test262_config.toml
```diff
@@ -7,6 +7,8 @@ flags = []
 features = [
     ### Unimplemented features:
 
+    "import-bytes",
+    "await-dictionary",
```
jedel1043: can you move these to the pending proposals section? to keep the features organized

nekevss: Ah, my bad. Moved it.

PR #4869, opened 2026-03-05
Switch to zerocopy tzdb provider

nekevss:
This switches us off the compiled tzdb provider that parses data at runtime to the zerocopy compiled tzdb provider.

Also, it lowers our dependencies, so that's nice.

### DIFF
--- a/core/engine/src/context/mod.rs
+++ b/core/engine/src/context/mod.rs
@@ -12,7 +12,7 @@ use intrinsics::Intrinsics;
 #[cfg(any(feature = "temporal", feature = "intl"))]
 use temporal_rs::provider::TimeZoneProvider;
 #[cfg(any(feature = "temporal", feature = "intl"))]
-use timezone_provider::tzif::CompiledTzdbProvider;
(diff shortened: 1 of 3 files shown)


core/engine/src/context/mod.rs
```diff
@@ -1142,7 +1142,7 @@ impl ContextBuilder {
             timezone_provider: if let Some(provider) = self.timezone_provider {
                 provider
             } else {
-                Box::new(CompiledTzdbProvider::default())
+                Box::new(ZeroCompiledTzdbProvider::default())
```
jedel1043, line 1145: I guess we need a `temporal_bundled` feature to mirror the `intl_bundled` feature, for folks that wanna bundle their own Temporal data.

nekevss: Yeah, for some reason I thought that was done. But it doesn't look like it is.

Would you prefer that added to this PR or a follow up?

jedel1043: Can be a follow up

PR #4870, opened 2026-03-05
Implement RISC-style opcodes for iterator operations

jedel1043:
~Depends on #4865.~ merged.

The equivalent to that PR but for iterator operations. Thankfully this doesn't require new operations, which is nice because it gives us two more operation slots for free :)


hansl:
Could you add a snapshot script that shows the new opcodes?

jedel1043:
sure, let me pull up one of the scripts I used as test

nekevss (approved):
LGTM

Was going to say that it looks like the example is missing the iterator, but I think that's because my Trace PR hasn't been updated, which was what would have unblocked that 🙃

PR #4872, opened 2026-03-05
fix(ci): replace deprecated wasm-pack-action with cargo install

### LINKED ISSUE #3912: bump node version in the npm action
We get the following error when publishing a release, we may need to bump node or a dependency for the wasm build:

```
The following actions uses Node.js version which is deprecated and will be forced to run on node20: jetli/wasm-pack-action@v0.4.0. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default
````

### PR DESCRIPTION
**Pull request by tkshsbcue**
# fix(ci): Replace deprecated wasm-pack-action with cargo install

Fixes #3912

---

## Summary

This pull request resolves the Node.js deprecation warning emitted during Boa's release publishing workflow. The `npm_publish` job currently uses `jetli/wasm-pack-action@v0.4.0`, which runs on Node.js 16—a version deprecated by GitHub Actions. This change replaces that action with the official `cargo install wasm-pack` method, eliminating the deprecation warning and aligning the workflow with supported tooling.

## Problem

When a release is published, the `npm_publish` job produces the following warning:

```
The following actions uses Node.js version which is deprecated and will be forced to run on node20: 
jetli/wasm-pack-action@v0.4.0. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default
```

While GitHub indicates it will force the action to Node 20, the warning clutters CI output and signals reliance on unmaintained infrastructure. The action has not been updated to explicitly support Node 20.

## Solution

Replace the third-party `jetli/wasm-pack-action` with a direct `cargo install wasm-pack` step. This approach:

- **Eliminates the deprecation warning** — No Node 16 dependency
- **Uses official tooling** — wasm-pack is installed from [crates.io](https://crates.io/crates/wasm-pack) via Cargo
- **Reduces external dependencies** — No reliance on third-party actions that may become unmaintained
- **Maintains correctness** — Same wasm-pack binary used for the build; behavior is unchanged

## Changes

| Before | After |
|--------|-------|
| `uses: jetli/wasm-pack-action@v0.4.0` with `version: 'latest'` | `run: cargo install wasm-pack` |

No other steps in the `npm_publish` job are modified. The build command, npm publishing logic, and Node.js setup remain unchanged.

## Trade-offs

- **Build time:** `cargo install` adds approximately 2–3 minutes to the job compared to downloading a pre-built binary. Release publishes are infrequent (typically a few times per release cycle), so this is an acceptable trade-off for cleaner, more maintainable CI.

## Testing

- Workflow YAML has been validated
- No changes to build or publish logic; existing release process is preserved

## Checklist

- [x] Fixes #3912
- [x] Workflow syntax is valid
- [x] No functional changes to build or publish steps

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,7 +58,9 @@ jobs:
           targets: wasm32-unknown-unknown
 
       - name: Install wasm-pack
-        run: cargo install wasm-pack
+        uses: baptiste0928/cargo-install@v3.4.0
+        with:
+          crate: wasm-pack
 
       - name: Build boa_wasm
         run: wasm-pack build --scope boa-dev ./ffi/wasm

### CODE REVIEW

**Comment by tkshsbcue** [comment]
@nekevss can you review my PR

File: .github/workflows/release.yml [hunk 1]
```diff
@@ -58,9 +58,7 @@ jobs:
           targets: wasm32-unknown-unknown
 
       - name: Install wasm-pack
-        uses: jetli/wasm-pack-action@v0.4.0
-        with:
-          version: 'latest'
+        run: cargo install wasm-pack
```
jedel1043: Can you use `baptiste0928/cargo-install` instead? We use that action widely because it automatically caches the compiled binaries.

tkshsbcue: alright!

tkshsbcue: @jedel1043 just made those changes!

tkshsbcue:
@jedel1043 please have a look

jedel1043 (approved):
Cool, thanks

PR #4873, opened 2026-03-05
feat(json): Implement json-parse-with-source proposal (ES2026)

xcb3d:
Related: https://github.com/tc39/proposal-json-parse-with-source

It changes the following:

- Implement `JSON.rawJSON(text)`: creates frozen raw JSON objects with `[[IsRawJSON]]` internal slot and a `rawJSON` property. Validates input is a valid JSON primitive (not object/array), rejects empty and whitespace-padded strings.
- Implement `JSON.isRawJSON(value)`: checks for the `[[IsRawJSON]]` internal slot.
- Update `JSON.stringify` to serialize raw JSON objects by outputting their `rawJSON` property value directly.
- Implement `JSON.parse` reviver `context.source`: the reviver function now receives a third `context` argument. For primitive JSON values, `context` contains a [source](cci:1://file:///d:/boa/core/engine/src/builtins/json/mod.rs:188:0-193:1) property with the original source text (e.g., `"1.1e1"` instead of `"11"`). For objects/arrays, `context` is an empty object. A value-matching guard ensures [source](cci:1://file:///d:/boa/core/engine/src/builtins/json/mod.rs:188:0-193:1) is omitted when the reviver forward-modifies values during iteration.
- Add a `JsonNode` source tree builder that parses JSON to extract byte-accurate source text spans for all primitive values.
- Remove `"json-parse-with-source"` from ignored features in [test262_config.toml](cci:7://file:///d:/boa/test262_config.toml:0:0-0:0).

**Test262 Results: 165/165 (100% conformance)**

| Suite | Pass | Total |
|-------|------|-------|
| `JSON.parse` | 77 | 77 |
| `JSON.stringify` | 66 | 66 |
| `JSON.rawJSON` | 10 | 10 |
| `JSON.isRawJSON` | 6 | 6 |

### DIFF
--- a/core/engine/src/builtins/json/mod.rs
+++ b/core/engine/src/builtins/json/mod.rs
@@ -93,26 +93,9 @@ impl<'ast> boa_ast::visitor::Visitor<'ast> for JsonSourceVisitor<'_> {
         &mut self,
         node: &'ast boa_ast::expression::literal::Literal,
     ) -> std::ops::ControlFlow<Self::BreakTy> {
-        use boa_ast::expression::literal::LiteralKind;
-
         let span = node.linear_span();
-        let text = if span.is_empty() {
-            // Fallback: reconstruct source text from literal kind
-            match node.kind() {
-                LiteralKind::Null => "null".to_string(),
-                LiteralKind::Bool(true) => "true".to_string(),
-                LiteralKind::Bool(false) => "false".to_string(),
-                LiteralKind::Num(n) => n.to_string(),
-                LiteralKind::Int(n) => n.to_string(),
-                LiteralKind::String(sym) => {
-                    format!("\"{}\"", self.interner.resolve_expect(*sym))
-                }
-                _ => String::new(),
-            }
-        } else {
-            let code_points = self.source_text.get_code_points_from_span(span);
-            String::from_utf16_lossy(code_points)
-        };
+        let code_points = self.source_text.get_code_points_from_span(span);
+        let text = String::from_utf16_lossy(code_points);
         self.stack.push(JsonNode::Primitive(text));
         std::ops::ControlFlow::Continue(())
     }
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/json/mod.rs
jedel1043, line 1: Seeing this file I'm wondering if it would be better to extend our JSON parser with what we need to get the source strings. We already collect spans of everything, so it might be better than having to re-parse?

core/engine/src/builtins/json/mod.rs
```diff
@@ -16,32 +16,141 @@
...
+
+/// Build a `JsonNode` source tree by walking the AST expression tree,
+/// extracting source text from `LinearSpan` information tracked by the parser.
+/// This avoids re-parsing the JSON string a second time.
+fn build_source_node_from_expression(
+    expr: &boa_ast::Expression,
+    source_text: &boa_ast::SourceText,
+    interner: &boa_interner::Interner,
+) -> JsonNode {
+    match expr {
```
jedel1043: No need to do all this manual matching, you can use our [Visitor trait](https://github.com/boa-dev/boa/blob/main/core%2Fast%2Fsrc%2Fvisitor.rs) to recursively collect the source strings

jedel1043 (approved):
Very nice work! No more nitpicks from me

xcb3d:
Fixed the formatting issue that caused the CI failure. Could you re-enable auto-merge when you get a chance? Thanks!

core/engine/src/builtins/json/mod.rs
```diff
@@ -16,32 +16,193 @@
...
+impl<'ast> boa_ast::visitor::Visitor<'ast> for JsonSourceVisitor<'_> {
+    type BreakTy = std::convert::Infallible;
+
+    fn visit_literal(
+        &mut self,
+        node: &'ast boa_ast::expression::literal::Literal,
+    ) -> std::ops::ControlFlow<Self::BreakTy> {
+        use boa_ast::expression::literal::LiteralKind;
+
+        let span = node.linear_span();
```
jedel1043, line 96: Are those checks for `span.is_empty()` necessary? because I would consider emitting an empty span for a literal a parser bug, and we should fix that. Otherwise, just using the span directly makes the code much cleaner.

xcb3d:
Hi @jedel1043, Just checking — is this PR ready to be merged, or is there anything else I should update? Thanks!

PR #4875, opened 2026-03-05
fix(jspromise): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241

Converts all 23 panics in jspromise.rs to EngineError::Panic using
js_expect introduced in #4828.

Patterns converted:
- `.js_expect()` on internal promise operations that cannot fail
- `.js_expect().map_err(Into::into)` for methods returning `JsResult<JsPromise>`

Signature changes:
- `resolve()`, `reject()`, `from_result()` → `JsResult<Self>`
- `then()`, `catch()`, `finally()` → `JsResult<JsPromise>`
- `all()`, `all_settled()`, `any()`, `race()` → `JsResult<JsPromise>`
- `into_js_future()` → `JsResult<JsFuture>`

Files modified: 6

### DIFF
--- a/core/engine/src/builtins/array/from_async.rs
+++ b/core/engine/src/builtins/array/from_async.rs
@@ -116,7 +116,7 @@ impl Array {
                     from_array_like(Ok(JsValue::undefined()), &coroutine_state, context)?
                 {
                     // Coroutine yielded. We need to allocate it for a future execution.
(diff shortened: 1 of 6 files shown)


Razican (approved):
Great work! Thanks! :)

PR #4876, opened 2026-03-05
fix(script): drop AST after script execution

ashnaaseth2325-oss:
## Context

This PR extends the idea introduced in #4855.

PR #4855 fixed a memory retention issue for **modules**, where the parsed AST and source text were kept in memory even after module initialization.

This PR applies the same idea to **scripts**, ensuring that the parsed AST is released once script execution has completed.

# Summary

This PR prevents scripts from keeping their parsed AST after compilation.

Right now `Script::Inner` stores the full AST in the `source` field:

```rust
struct Inner {
    realm: Realm,
    source: boa_ast::Script,
    source_text: SourceText,
    codeblock: GcRefCell<Option<Gc<CodeBlock>>>,
}
```

The AST is only needed while generating bytecode in `Script::codeblock()`.
After compilation it is not used anymore, but it still stays in `Script::Inner`.

Function objects created from the script keep a reference to the script (`OrdinaryFunction::script_or_module`). Because of this, the script can stay alive for a long time, which also keeps the whole AST in memory.

# Steps to Reproduce

In a long-running program (for example a REPL or server), repeatedly run scripts that define functions:

```rust
let script = Script::parse(Source::from_bytes(src), None, &mut context)?;
script.evaluate(&mut context)?;
```

If those functions are stored on the global object, they keep the script alive.
Since the script holds the AST, memory usage keeps growing as more scripts run.

# Impact

This mostly affects long-running programs that embed Boa.

Every script that defines functions can keep its AST in memory forever.
AST structures are usually much bigger than the source code, so memory usage can grow over time.

Modules already drop their AST after linking, but scripts did not.

# Fix

`Script::Inner::source` is changed to:

```rust
source: RefCell<Option<boa_ast::Script>>
```

The AST is used during compilation and then removed once the `CodeBlock` is created:

```rust
*self.inner.source.borrow_mut() = None;
```

# Result

After this change, the AST is dropped once compilation finishes.

Scripts now behave consistently with modules after the change introduced in #4855.

**The AST is only required during parsing and compilation. Once the script finishes execution and the result is consumed, it can be safely released to reduce memory usage in long-running runtimes.**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,6 +391,7 @@ dependencies = [
 name = "boa_cli"
 version = "1.0.0-dev"
 dependencies = [
+ "async-channel 2.5.0",
  "boa_engine",
  "boa_gc",
  "boa_parser",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -27,6 +27,7 @@ color-eyre.workspace = true
 cow-utils.workspace = true
 futures-concurrency.workspace = true
 futures-lite.workspace = true
+async-channel = "2.5.0"
 
 [features]
 default = [
--- a/cli/src/executor.rs
+++ b/cli/src/executor.rs
@@ -1,0 +1,143 @@
+use std::{
+    cell::RefCell,
+    collections::{BTreeMap, VecDeque},
+    mem,
+    rc::Rc,
+};
+
+use boa_engine::{
+    Context, JsResult,
+    context::time::JsInstant,
+    job::{GenericJob, Job, JobExecutor, NativeAsyncJob, PromiseJob, TimeoutJob},
+};
+use futures_concurrency::future::FutureGroup;
+use futures_lite::{StreamExt, future};
+
+use crate::{logger::SharedExternalPrinterLogger, uncaught_job_error};
+
+pub(crate) struct Executor {
+    promise_jobs: RefCell<VecDeque<PromiseJob>>,
+    async_jobs: RefCell<VecDeque<NativeAsyncJob>>,
+    timeout_jobs: RefCell<BTreeMap<JsInstant, Vec<TimeoutJob>>>,
+    generic_jobs: RefCell<VecDeque<GenericJob>>,
+
+    printer: SharedExternalPrinterLogger,
+}
+
+impl Executor {
+    pub(crate) fn new(printer: SharedExternalPrinterLogger) -> Self {
+        Self {
+            promise_jobs: RefCell::default(),
+            async_jobs: RefCell::default(),
+            timeout_jobs: RefCell::default(),
+            generic_jobs: RefCell::default(),
+            printer,
+        }
+    }
+
+    pub(crate) fn clear(&self) {
+        self.promise_jobs.borrow_mut().clear();
+        self.async_jobs.borrow_mut().clear();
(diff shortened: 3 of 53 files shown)


ashnaaseth2325-oss:
Hello @zhuzhu81998 
#4855 handled the module path (`SourceTextModule::initialize_environment`).
This PR is trying to apply the same idea to the script execution path (`Script::evaluate_async_with_budget`), since scripts can also retain their parsed AST after execution.
So the goal here is basically to mirror the module fix but for regular scripts.

ashnaaseth2325-oss:
Hello @hansl @jedel1043 
This PR was trying to apply the same idea to scripts. It looks like dropping the AST here causes some test262 regressions (+97 failures), so the placement is probably too early.
I'm looking into where it could be safely released instead.

core/engine/src/script.rs
```diff
@@ -46,8 +49,9 @@ impl std::fmt::Debug for Script {
 #[derive(Trace, Finalize)]
 struct Inner {
     realm: Realm,
+    // Safety: `boa_ast::Script` contains no GC-managed values.
     #[unsafe_ignore_trace]
-    source: boa_ast::Script,
+    source: RefCell<Option<boa_ast::Script>>,
     source_text: SourceText,
     codeblock: GcRefCell<Option<Gc<CodeBlock>>>,
```
jedel1043: I think a better approach would be to make this item an enum like
```rust
enum ScriptPhase {
    Ast(boa_ast::Script),
    Codeblock(Gc<CodeBlock>)
}
```
Then we transition through the phases of compilation of the script in `codeblock`.
This saves some nice bytes of memory representation.

ashnaaseth2325-oss:
Hello @jedel1043 
Thanks for the suggestion! I refactored the implementation to use an enum-based phase (`ScriptPhase`) as you described, transitioning from `Ast` → `Codeblock` during compilation. The project builds and tests pass locally.

core/engine/src/script.rs
```diff
@@ -120,18 +121,18 @@ impl Script {
     ///
     /// This is a no-op if this has been called previously.
     pub fn codeblock(&self, context: &mut Context) -> JsResult<Gc<CodeBlock>> {
-        let mut codeblock = self.inner.codeblock.borrow_mut();
+        let mut phase = self.inner.phase.borrow_mut();
 
-        if let Some(codeblock) = &*codeblock {
+        if let ScriptPhase::Codeblock(codeblock) = &*phase {
             return Ok(codeblock.clone());
         }
```
jedel1043: Let's make use of Rust features :)
```rust
let phase = self.inner.phase.borrow();
let source = match &*phase {
    ScriptPhase::Codeblock(codeblock) => return Ok(codeblock.clone()),
    ScriptPhase::Ast(source) => source
}
```

jedel1043: I edited this btw, at first I used a clone but I realized that would clone the whole AST, which we don't want

ashnaaseth2325-oss:
Hello @jedel1043 
Thanks for the suggestion! I updated the code to follow the pattern you suggested (using `borrow()` and avoiding cloning the AST). I’m checking the CI failures now to see what still needs fixing.

PR #4877, opened 2026-03-05
Add regression tests for Symbol.match override and RegExp.exec mutation

mrhapile:
This Pull Request fixes/closes N/A.

It changes the following:

- Add a regression test verifying that objects implementing `Symbol.match`
  are correctly invoked by `String.prototype.match`.

- Add a regression test ensuring that `String.prototype.match` respects
  user-overridden `RegExp.exec` implementations.

- Improve localized Rust-level test coverage for String ↔ RegExp dispatch
  behavior to help detect regressions earlier during engine changes.

While these behaviors are covered by test262, adding targeted Rust-level
tests provides faster feedback when changes affect prototype dispatch or
RegExp execution routing inside the engine.

Screenshot 2026-03-05 at 8 57 43 PM


core/engine/src/builtins/string/tests.rs
```diff
@@ -916,3 +916,36 @@ fn from_code_point() {
         ),
     ]);
 }
+
+#[test]
+fn match_with_symbol_match_method() {
+    run_test_actions([
+        TestAction::run_harness(),
```
jedel1043: Copilot's right, you don't need the harness if you're not gonna compare arrays or something like that

mrhapile:
Screenshot 2026-03-06 at 4 43 24 PM
removed harness and updated the branch 

@jedel1043

PR #4879, opened 2026-03-05
Do not pull default engine features into runtime

tomasol:
The default features include `temporal` which cannot be deactivated with the previous configuration.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -10,7 +10,7 @@ repository.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-boa_engine = { workspace = true, features = ["annex-b"] }
+boa_engine = { workspace = true, features = ["annex-b", "float16", "xsum", "temporal"] }
 boa_ast.workspace = true
 boa_interner.workspace = true
 boa_gc.workspace = true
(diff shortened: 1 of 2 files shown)


nekevss:
What is the use case for not wanting to include default features into the runtime?

Also, if we are turning it off by default, shouldn't provide corresponding feature flags to turn them back on?

tomasol:
> What is the use case for not wanting to include default features into the runtime?

I want to be able to disable certain engine features like `temporal` for smaller binary size. This is possible if just depending on `engine`, however `runtime` is currently dependent on all default `engine` features as well, so they are pulled anyway. 

> Also, if we are turning it off by default, shouldn't provide corresponding feature flags to turn them back on?

I've re-enabled the `engine`'s default features as default features of `runtime`, so existing users should see no change. Only users who explicitly disable `runtime`'s default features gain the ability to also exclude `temporal` etc.

jedel1043:
I agree with this change. `boa_runtime` should be a reusable library of components that folks can use to build their runtimes first, and a "runtime" second. Not bundling unnecessary features by default seems to align with our first target.

nekevss:
I'm for this changes as well.

Part of the reason why I'm for the same change in the engine too. It's just given me pause because I'm curious if there's a way to ship temporal without any of the data compiled in the binary. But that can be solved by feature flagging the default provider.

PR #4880, opened 2026-03-05
fix(vm): fix CompletionRecord semantics for Yield and Return opcodes

Nakshatra480:
## Summary

Fixes issue #2675 — Evaluate/Reevaluate the execution semantics of `CompletionRecord`

## Root Cause

Boa's VM was previously emitting the wrong `CompletionRecord` variants for yield and return in generator contexts:

- Yield was emitting `CompletionRecord::Return` (non-normal).
- Return was emitting `CompletionRecord::Normal`.

Both were swapped from what the ECMAScript spec requires. The generator drive loop had its arms swapped to compensate, making the code confusing and fragile.

## Changes

- **Sub-issue 1 — Audit**: Audited `CompletionRecord` emit/match sites across `vm/opcode/` and `builtins/generator/`. Scope confirmed minimal — only 2 files required changes.
- **Sub-issue 2 — Fix Yield opcode**: `Context::handle_yield` now correctly emits `CompletionRecord::Normal` per spec §27.5.3.7.
- **Sub-issue 3 — Fix Return opcode**: `Context::handle_return` now correctly emits `CompletionRecord::Return` per spec §14.4.
- **Sub-issue 4 — Generator drive loop**: Swapped match arms in both `generator_resume` and `generator_resume_abrupt` so `Normal → SuspendedYield` and `Return → Completed`.

## Testing

- [x] All existing generator tests pass.
- [x] Added 3 new regression tests covering yield, return, and early-return scenarios.
- [x] Test262 generator suite passing with no regressions.

### DIFF
--- a/core/ast/src/scope.rs
+++ b/core/ast/src/scope.rs
@@ -261,6 +261,28 @@ impl Scope {
         self.inner.outer.is_none()
     }
 
+    /// Check if a binding with the given name is mutable.
+    ///
+    /// Returns `Some(true)` for mutable bindings (`let`, `var`),
+    /// `Some(false)` for immutable bindings (`const`),
+    /// or `None` if the binding is not found in this or any outer scope.
+    #[must_use]
+    pub fn is_binding_mutable(&self, name: &JsString) -> Option<bool> {
+        if let Some(binding) = self
+            .inner
+            .bindings
+            .borrow()
+            .iter()
+            .find(|b| &b.name == name)
+        {
+            Some(binding.is_mutable())
+        } else if let Some(outer) = &self.inner.outer {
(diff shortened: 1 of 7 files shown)


Nakshatra480:
Hey @jedel1043  I've now addressed all 4 sub-issues from #2675. The Yield opcode correctly emits CompletionRecord::Normal (§27.5.3.7), Return now emits CompletionRecord::Return (§14.10.1), and the generator drive loop arms are updated accordingly so Normal → SuspendedYield and Return → Completed. Also did a full audit of CompletionRecord emit sites across vm/opcode/ and builtins/generator/. Regression tests were covered separately in #4887 PR. Would really appreciate a review when you get a chance!
Thanks for your time

PR #4884, opened 2026-03-05
perf(vm): two small changes

hansl:
Eliminate frame swap on push/pop by storing all frames in Vec

Instead of keeping the current CallFrame as a separate field on Vm and
swapping it in/out on every push_frame/pop_frame, store all frames
(including a dummy sentinel at index 0) in the existing frames Vec.
frame()/frame_mut() now return the last element via unchecked indexing.

This eliminates the std::mem::swap of the entire CallFrame struct on
every function call and return, replacing it with a simple Vec push/pop.

----

Add `take_register` to avoid unnecessary clones on consumed values.

### DIFF
--- a/core/engine/src/vm/mod.rs
+++ b/core/engine/src/vm/mod.rs
@@ -385,12 +385,7 @@ impl Vm {
         // SAFETY: Register indices are determined by the bytecode compiler and are
         // guaranteed to be within the register bounds for well-formed bytecode. The
         // debug_assert above catches any compiler bugs during development.
-        unsafe {
-            std::mem::replace(
-                self.stack.stack.get_unchecked_mut(rp + index),
-                JsValue::undefined(),
-            )
-        }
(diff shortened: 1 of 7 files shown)


core/engine/src/vm/mod.rs
```diff
@@ -354,33 +340,52 @@ impl Vm {
...
+            rp + index < self.stack.stack.len(),
+            "register index out of bounds: rp {rp}, index {index}, stack len {}",
+            self.stack.stack.len()
         );
         // SAFETY: Register indices are determined by the bytecode compiler and are
         // guaranteed to be within the register bounds for well-formed bytecode. The
         // debug_assert above catches any compiler bugs during development.
         unsafe {
-            *self.registers.get_unchecked_mut(actual) = value;
+            *self.stack.stack.get_unchecked_mut(rp + index) = value;
```
nekevss, line 353: Well ... `stack.stack` is at least doubly descriptive.

nekevss (approved):

PR #4886, opened 2026-03-05
Fix timeout scheduling edge case and cancellation handling in SimpleJobExecutor

mrhapile:
This PR fixes two issues in `SimpleJobExecutor`.

1. Timeout jobs scheduled exactly at `now` could be deferred due to the
behavior of `BTreeMap::split_off`.
2. Cancelled timeout jobs could still execute because the executor
did not verify `job.is_cancelled()` before dispatch.
Screenshot 2026-03-06 at 1 27 46 AM

Fix:
- Explicitly move jobs scheduled at `now` back into the execution set.
- Add a guard preventing execution of cancelled jobs.

Two regression tests were added:
Screenshot 2026-03-06 at 1 58 06 AM
Screenshot 2026-03-06 at 1 59 03 AM

- timeout_runs_at_exact_time
- cancelled_timeout_should_not_execute

All tests pass locally (`cargo test`, `cargo clippy`).

### DIFF
--- a/core/engine/src/vm/mod.rs
+++ b/core/engine/src/vm/mod.rs
@@ -12,6 +12,7 @@ use crate::{
     object::JsFunction,
     realm::Realm,
     script::Script,
+    vm::opcode::{OPCODE_HANDLERS, OPCODE_HANDLERS_BUDGET},
 };
 use boa_gc::{Finalize, Gc, Trace, custom_trace};
 use shadow_stack::ShadowStack;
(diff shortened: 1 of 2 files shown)


core/engine/src/job.rs
```diff
@@ -696,6 +696,11 @@ impl JobExecutor for SimpleJobExecutor {
                 let now = context.borrow().clock().now();
                 let mut timeouts_borrow = self.timeout_jobs.borrow_mut();
                 let mut jobs_to_keep = timeouts_borrow.split_off(&now);
+                // `split_off` keeps keys `>= now`. We want exactly `now` to execute in the
+                // current tick, so we move it back into `timeouts_borrow` (which becomes `jobs_to_run`).
+                if let Some(jobs_at_now) = jobs_to_keep.remove(&now) {
```
hansl, line 701: If there is more than 1 job due now, this would only run one of them, right>

mrhapile: No — it will run all of them.

`timeout_jobs` is a `BTreeMap<JsInstant, Vec<TimeoutJob>>`, so `remove(&now)` returns the entire vector of jobs scheduled at that timestamp, and the executor then iterates over all of them.

jedel1043:
Might be superseeded by #4891

jedel1043:
Superseeded by #4891

PR #4891, opened 2026-03-06
Simplify executor implementations

jedel1043:
After seeing the [avalanche](https://github.com/boa-dev/boa/pull/4886) [of](https://github.com/boa-dev/boa/pull/4833) [issues](https://github.com/boa-dev/boa/pull/4785) [that](https://github.com/boa-dev/boa/pull/4749) [our](https://github.com/boa-dev/boa/issues/4782) executors have, I don't think it's worth being too smart about when to exit the loop. Folks who wanna have that behaviour can just implement their own `JobExecutor`.

Thus, this reverts the implementation to the old behaviour of blocking when having pending timeout and interval jobs. Fortunately, `run_jobs_async` exists, so we can use it in our CLI to intersperse executing jobs with executing parsing and execution.

And just for good measure, it also adds a `stop` cancellation token to `SimpleJobExecutor`, in case folks still want to use it but that also want to remotely stop the event loop's execution from another thread.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,6 +391,7 @@ dependencies = [
 name = "boa_cli"
 version = "1.0.0-dev"
 dependencies = [
+ "async-channel 2.5.0",
  "boa_engine",
  "boa_gc",
  "boa_parser",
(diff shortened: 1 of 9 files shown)


nekevss (approved):
LGTM

PR #4892, opened 2026-03-06
feat: implement tc39 Symbols as WeakMap keys

shruti2522:
with the "Symbols as WeakMap keys" proposal officially in Stage 4 (link- https://tc39.es/proposal-symbols-as-weakmap-keys/), 
I think we can implement `CanBeHeldWeakly` for Boa

It changes the following:

- Implements `can_be_held_weakly`
- `WeakMap` and `WeakSet` now keep track of Objects using the regular GC under the hood and store Symbol hashes in a separate list
- `WeakMap`, `WeakSet` and `WeakRef` all now accept non registered Symbols as valid keys
- adds tests to verify behavior with normal symbols, registered symbol and builtin well known symbols.


jedel1043:
I'm gonna close this because we should at least add support for symbols in our GC for this to be supported properly. We can revisit this implementation after we address that.

PR #4894, opened 2026-03-06
perf(parser): avoid heap allocation when interning regex flags

tkshsbcue:
# perf(parser): avoid heap allocation when interning regex flags

## Summary

Avoids heap allocation when lexing regex literals by using a stack buffer for the flags string instead of `to_string()`.

## Motivation

During regex literal lexing, `flags.to_string().as_str()` allocates a `String` on the heap for every regex literal (e.g. `/foo/gi`, `/bar/ms`). The flags string is at most 8 characters (d, g, i, m, s, u, y, v), so this allocation is unnecessary and adds overhead in a hot path.

## Changes

- Add `RegExpFlags::write_to_buf(&self, buf: &mut [u8; 8]) -> &str` that writes flags into a caller-provided buffer and returns a `&str` slice
- Update the regex lexer to use a stack buffer and `write_to_buf` instead of `flags.to_string().as_str()` when interning flags
- Flag order matches the existing `Display` impl (d, g, i, m, s, u, y, v)

## Testing

- `cargo test -p boa_parser` — all 297 tests pass
- Includes `regex_literal`, `regex_literal_flags`, `regex_literal_flags_err`


tkshsbcue:
@jedel1043 can you review my PR please 
Thanks!

jedel1043 (approved):
Thanks!

PR #4895, opened 2026-03-06
feat: implement Trace and Finalize for common std types #3239

### LINKED ISSUE #3239: Tracking issue for std types needing a `Trace` impl
- [ ] `TcpStream`

Feel free to comment if you need us to add any missing `Trace` impl

0xShubhamSolanki:
This Pull Request fixes #3239.

Description
Finally added the missing Trace and Finalize impls for common std types as discussed in #3239. This should help folks who are embedding Boa and using these types in their state.

What's included:

Basic types: Added Instant, SystemTime, Path, and PathBuf using the empty_trace! macro.

I/O & Net: Added File, FileType, TcpStream, and UdpSocket. These are wrapped in #[cfg(not(target_family = "wasm"))] so we don't break WASM builds.

Sync wrappers: Did manual Trace impls for Mutex<T> and RwLock<T>. Since these can hold traceable data, I made sure they actually visit the inner T instead of just skipping it.

Testing:
I've added a new test file core/gc/src/test/std_types.rs to verify that Mutex/RwLock actually trigger the tracer and that the leaf types don't panic.

Ran cargo test -p boa_gc locally and all 25 tests are passing.

### DIFF
--- a/core/gc/src/trace.rs
+++ b/core/gc/src/trace.rs
@@ -11,7 +11,7 @@ use std::{
     },
     path::{Path, PathBuf},
     rc::Rc,
-    sync::{Mutex, RwLock, atomic},
+    sync::atomic,
     time::{Instant, SystemTime},
 };
 
@@ -572,16 +572,6 @@ mod boa_string_trace {
     }
 
     impl Finalize for boa_string::JsString {}
-}
-
-impl<T: Trace> Finalize for Mutex<T> {}
-unsafe impl<T: Trace> Trace for Mutex<T> {
-    empty_trace!();
-}
-
-impl<T: Trace> Finalize for RwLock<T> {}
-unsafe impl<T: Trace> Trace for RwLock<T> {
-    empty_trace!();
 }
 #[cfg(feature = "either")]
 mod either_trace {
--- a/core/gc/src/test/std_types.rs
+++ b/core/gc/src/test/std_types.rs
@@ -1,27 +1,6 @@
-use crate::{Gc, Trace, Tracer};
+use crate::{Trace, Tracer};
 use std::path::PathBuf;
-use std::sync::{Mutex, RwLock};
 use std::time::Instant;
-
-#[test]
-fn test_sync_types_ignored_trace() {
-    // Note: Since `Gc` pointers are `!Send`, they cannot be safely shared across threads.
-    // Therefore, `Mutex` and `RwLock` just use `empty_trace!()` and ignore inner values.
-    let mutex = Mutex::new(Gc::new(10));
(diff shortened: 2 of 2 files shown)


core/gc/src/trace.rs
```diff
@@ -563,6 +574,23 @@ mod boa_string_trace {
...
+}
+
+impl<T: Trace> Finalize for RwLock<T> {}
+unsafe impl<T: Trace> Trace for RwLock<T> {
+    custom_trace!(this, mark, {
+        if let Ok(val) = this.read() {
+            mark(&*val);
+        }
+    });
+}
```
jedel1043: Implementing these don't really make much sense; if someone's using a `Mutex<T>`, the `T` should at least be `Send` for it to be useful, but all our Gc pointers are `!Send`. In that case, just adding `#[unsafe_ignore_trace]` makes much more sense.

0xShubhamSolanki:
Thanks for catching that, @jedel1043  you are right since Gc is !Send it doesn't make sense to trace inside a mutex or Rwlock. so i have swapped both mutex and rwlock to use empty_trace!() and adjusted the tests to verify the tracer stays empty. Just pushed the update. Let me know if it looks good now

core/gc/src/trace.rs
```diff
@@ -576,20 +576,12 @@ mod boa_string_trace {
 
 impl<T: Trace> Finalize for Mutex<T> {}
 unsafe impl<T: Trace> Trace for Mutex<T> {
-    custom_trace!(this, mark, {
-        if let Ok(val) = this.lock() {
-            mark(&*val);
-        }
-    });
+    empty_trace!();
```
jedel1043: Now it's worse, because a `Mutex<Gc<T>>` will cause UB since the inner `Gc` won't be marked as live.

I think we shouldn't implement this for `Mutex<T>`, and we should require users to mark those as `#[unsafe_ignore_trace]`

0xShubhamSolanki:
My bad @jedel1043, you're right about empty_trace. here is a recipe for UB since it hides the inner Gc.
 i have removed the mutex and rwLock implementations entirely and cleaned up the tests. This way, users are forced to handle them manually. let me know about that is it better or needs more improvement.

0xShubhamSolanki:
Hey @jedel1043, just wanted to check in on this. I have completely removed the Mutex and RwLock implementations to avoid the undefined behavior risks we discussed, and i have also cleaned up the tests accordingly.Does the current approach of only including the safe leaf std types (like instant, pathbuf etc) looks good to you? Let me know if there anything else i should tweak to get this across the finish line.

PR #4896, opened 2026-03-06
perf: detect Latin1-encodable strings at intern time instead of per-c…

### LINKED ISSUE #4881: # Optimize string handling by detecting Latin1-encodable strings during parsing
## Summary

There is a TODO in `core/engine/src/bytecompiler/mod.rs` suggesting that strings which can be encoded as **Latin1** should be identified during the parsing stage rather than checked at runtime.

Implementing this optimization would eliminate unnecessary runtime checks and improve overall performance when compiling JavaScript source code.

---

## Problem

Currently, the engine may perform runtime checks to determine whether a string can be encoded as **Latin1**.

This introduces avoidable overhead because:

- The parser already processes the string contents.
- It could determine during parsing whether all characters fall within the Latin1 range.
- If known earlier, the bytecompiler can store this information and skip runtime checks.

---

## References

 
`core/engine/src/bytecompiler/mod.rs`

Relevant TODO comment indicates this optimization opportunity.

linisha15:
This Pull Request
closes #4881

Background-

When the bytecode compiler converts an interned string (`Sym`) to a `JsString`, it needs to decide whether to store it as Latin1 (1 byte per character) or UTF-16 (2 bytes per character). Previously, this was done by scanning every character of the string on each call — even if the same string was used many times.

What changed-

- The `Interner` now checks once, at the moment a string is first stored, whether all its characters fit in Latin1 (code point ≤ U+00FF). The result is saved in a new `latin1_flags` field.

- A new `is_latin1(sym)` method lets callers read that saved result instantly, without re-scanning the string.

- `ToJsString for Sym` in both `boa_ast` and `boa_engine::bytecompiler` now calls `is_latin1()` instead of scanning the string's characters every time.

- `From<&str> for JsString` was also fixed to correctly produce a Latin1 string for characters in the U+0080–U+00FF range, not just plain ASCII.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,6 @@
+# TODO: track https://github.com/rust-lang/rust/issues/141626 for a resolution
 [target.x86_64-pc-windows-msvc]
 rustflags = ['-Csymbol-mangling-version=v0']
 
 [env]
 MIRIFLAGS = "-Zmiri-tree-borrows"
-RUST_BACKTRACE = { value = "1", force = false }


rajat552:
Hi @linisha15 ,

Thanks for working on this! I was also exploring this issue since I opened it.

I noticed that the formatting and lint checks are currently failing. You may want to run `cargo fmt` and `cargo clippy` to fix them before review.

Looking forward to the updated changes.

rajat552:
please fix CI first

linisha15:
@rajat552 Thanks for the feedback! I clicked “Update branch” to sync with the latest `main`. It looks like the CI workflows are now waiting for maintainer approval before they run again. Once they are approved, the checks should start automatically.

linisha15:
Hi @rajat552, thanks for the feedback . CI has been fixed , the formatting and lint issues are resolved in the latest commit.

linisha15:
Hi! It looks like the CI workflows are currently awaiting maintainer approval before they can run. Once they are approved, the checks should start automatically. Please let me know if anything else is needed from my side. Thanks!

.cargo/config.toml
```diff
@@ -1,6 +1,6 @@
-# TODO: track https://github.com/rust-lang/rust/issues/141626 for a resolution
```
zhuzhu81998: why delete?

.cargo/config.toml
zhuzhu81998, line 1: this file should not have been committed.

linisha15: Thanks for pointing that out! That change was unintentional. I've reverted the modification to `.cargo/config.toml` and pushed the update.

jedel1043 (approved):
Nice optimization!

PR #4905, opened 2026-03-06
Extend Response API

tomasol:
This PR allows writing simple HTTP endpoints.
In JS, it allows creating `Response` using constructor or using `json()` static method. Example:
```js
function handle(request) {
    return new Response("Hello from JS!", {
        status: 200,
        headers: { "content-type": "text/plain" },
    });
}
```
Rust code can then obtain response status, headers and body.

### DIFF
--- a/core/runtime/src/fetch/response.rs
+++ b/core/runtime/src/fetch/response.rs
@@ -206,7 +206,7 @@ pub struct JsResponseOptions {
 /// `body_with_type` is a body paired with its MIME type (if any), as produced by
 /// "extract a body". `None` means no body was provided.
 fn initialize_response(
-    init: JsResponseOptions,
+    init: &JsResponseOptions,
     body_with_type: Option<(Vec<u8>, Option<&str>)>,
 ) -> JsResult<JsResponse> {
     // Step 1: If init["status"] is not in the range 200 to 599, inclusive, throw a RangeError.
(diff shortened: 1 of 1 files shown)


core/runtime/src/fetch/response.rs
```diff
@@ -147,8 +147,32 @@ impl JsResponse {
     }
 }
 
+/// Parse an optional [`JsResponseOptions`] argument.
+fn parse_response_options(init: &JsValue, context: &mut Context) -> JsResult<JsResponseOptions> {
+    if init.is_undefined() || init.is_null() {
```
jedel1043: There is an `is_null_or_undefined` method btw

core/runtime/src/fetch/response.rs
```diff
@@ -165,13 +189,61 @@ impl JsResponse {
         Self::error()
     }
 
+    /// Creates a `Response` with a JSON-serialized body and `Content-Type: application/json`.
+    ///
+    /// See <https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static>.
```
jedel1043: Can you also add a link to the WhatWG spec? Should be this one:
https://fetch.spec.whatwg.org/#dom-response-json

core/runtime/src/fetch/response.rs
```diff
@@ -165,13 +189,62 @@ impl JsResponse {
...
+                Convert::from("content-type".to_string()),
+                Convert::from("application/json".to_string()),
+            )?;
+        }
+
+        Ok(Self::basic(
+            js_string!(""),
+            build_response(status, &headers, json_bytes)?,
+        ))
+    }
```
jedel1043, line 324: This is not following the spec by the way. For example, the status needs to be checked to be between 200 and 599 according to https://fetch.spec.whatwg.org/#initialize-a-response.

You should go through the spec and add comments on the code mirroring it, to make sure everything gets properly implemented here.

tomasol:
Thanks for review, added more checks and getters.

PR #4907, opened 2026-03-06
Add basic benchmarks to the boa_benches suite

hansl:


jedel1043:
I think you also need to update the `benches/full.rs`, right?

hansl:
These are outside the engine, and can be run using `cargo bench -p boa_benches`. 

Maybe I should merge those 2 suites.

nekevss (approved):
LGTM

PR #4908, opened 2026-03-07
fix(builtins): correct Object.prototype.__proto__ setter length to 1

Nakshatra480:
## Description

This PR fixes a bug where the [set](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/object/mod.rs:663:4-722:5) accessor for `Object.prototype.__proto__` incorrectly had `.length === 0`. The ECMAScript specification requires it to be `1`.

Per [§B.2.2.1.2 `set Object.prototype.__proto__ ( proto )`](https://tc39.es/ecma262/#sec-set-object.prototype.__proto__), the setter takes one parameter ([proto](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/object/mod.rs:724:4-754:5)), so its [length](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/object/tests.rs:4:0-7:1) property must be `1`.

In [core/engine/src/builtins/object/mod.rs](cci:7://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/object/mod.rs:0:0-0:0), the setter was registered via `BuiltInBuilder::callable()` without specifying `.length(1)`, defaulting to `0`. This PR adds the correct `.length(1)` builder configuration and a regression test to prevent recurrence.

Node.js (V8) returns `1`, confirming the expected behavior.

## Changes

- **Fix**: Added `.length(1)` to the `set __proto__` built-in builder.
- **Test**: Added [object_prototype_proto_accessor_properties](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/object/tests.rs:429:0-440:1) regression test verifying `.length` and `.name` properties on both `__proto__` accessors.

## Reproduction

In Boa CLI previously:
```js
var desc = Object.getOwnPropertyDescriptor(Object.prototype, "__proto__");
console.log(desc.set.length); // Expected: 1, Actual: 0

### DIFF
--- a/core/parser/src/lexer/string.rs
+++ b/core/parser/src/lexer/string.rs
@@ -45,10 +45,6 @@ pub(crate) enum StringTerminator {
 pub(crate) trait UTF16CodeUnitsBuffer {
     /// Encodes the code point to UTF-16 code units and push to the buffer.
     fn push_code_point(&mut self, code_point: u32);
-
-    /// Decodes the buffer into a String and replace the invalid data with the replacement character (U+FFFD).
-    #[allow(dead_code)]
-    fn to_string_lossy(&self) -> String;
 }
 
 impl UTF16CodeUnitsBuffer for Vec<u16> {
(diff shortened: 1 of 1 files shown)


Nakshatra480:
Thank you🙂

PR #4888, opened 2026-03-05
feat: improve REPL interface with banner, dot-commands, and --quiet flag

sahanaxzy:
## Summary

Improves the REPL experience by adding a welcome banner, dot-commands, and a `--quiet` flag.

Solves #4883

## Changes

- Welcome banner with version info on startup
- `.help`, `.clear`, `.load <file>` dot-commands
- `--quiet` / `-q` flag to suppress the banner
- Ctrl+C now shows exit hint instead of quitting
- Improved error formatting (bold + red labels)
- Added help descriptions for `-O` and `--optimizer-statistics`
- Updated [README.md](cci:7://file:///c:/Users/Sahana/boa/cli/README.md:0:0-0:0) with new REPL Commands section

## Files Changed
 Modified  `cli/src/main.rs`
> REPL loop, banner, dot-commands, CLI flags, error formatting

Modified  `cli/README.md`
> Updated CLI options and added REPL Commands section

PR #4911, opened 2026-03-07
Implement validator for JS syntax in CLI

jedel1043:
Before this change, when being on REPL mode, you could type something that is invalid Javascript but contains unclosed brackets, then press enter and the REPL would happily accept your input. For example:

```shell
>> function b() [
       let a = 5;
       let b = 6;
       let c = 7;
]
Uncaught: SyntaxError: expected token '{', got '[' in function declaration at line 1, col 14 (native at core/engine/src/script.rs:95:34)
```
Note that it prints an error only until the bracket is closed, even if there was already a syntax error from line 1.

What this does is to make it such that the CLI gives feedback when pressing enter if the input is already invalid JS:

```shell
>> function b() [
Uncaught SyntaxError expected token '{', got '[' in function declaration at line 1, col 14
```

### DIFF
--- a/core/parser/src/error/mod.rs
+++ b/core/parser/src/error/mod.rs
@@ -78,7 +78,7 @@ pub enum Error {
 
     /// A scope analysis error.
     ScopeAnalysis {
-        /// The error that ocurred during scope analysis.
+        /// The error that occurred during scope analysis.
         err: &'static str,
     },
 


cli/src/main.rs
```diff
@@ -361,16 +361,12 @@ fn generate_flowgraph<R: ReadChar>(
...
 }
 
 #[must_use]
 fn uncaught_job_error(error: &JsError) -> String {
-    format!(
-        "{}: {}\n",
-        "Uncaught error (during job evaluation)".red(),
-        error.to_string().red()
-    )
+    format!("{} {}\n", "Uncaught ".red(), error.to_string().red())
```
hansl: Double space. And comment above.

cli/src/main.rs
```diff
@@ -361,16 +361,12 @@ fn generate_flowgraph<R: ReadChar>(
 
 #[must_use]
 fn uncaught_error(error: &JsError) -> String {
-    format!("{}: {}\n", "Uncaught".red(), error.to_string().red())
+    format!("{} {}\n", "Uncaught".red(), error.to_string().red())
```
hansl: This is pending: https://github.com/boa-dev/boa/pull/4888 and like the way they're doing it.

jedel1043: Gotcha, gonna rebase against that PR

rajat552:
Hi @jedel1043 , I noticed that PR #4888 has already been merged.  
Do you think rebasing this branch on the latest main could fix the lint failure?  
I'd be happy to help test the changes if needed.

PR #4913, opened 2026-03-07
test(vm): add round-trip tests for bytecode argument encode/decode

tkshsbcue:
# test(vm): add round-trip tests for bytecode argument encode/decode

## Summary

Adds unit tests for the `Argument` trait's encode/decode logic in `core/engine/src/vm/opcode/args.rs`. The bytecode argument serialization uses `unsafe` via `read_unchecked` and previously had no test coverage. These tests validate round-trip correctness for all argument types used by VM opcodes.

## Motivation

The VM's bytecode format encodes instruction arguments (addresses, register operands, literals, etc.) into a compact byte stream. The `Argument` trait defines `encode` and `decode` for each type, with decoding implemented via a `read` helper that uses `read_unchecked` for unaligned reads from byte slices. This `unsafe` path had no unit tests, leaving encode/decode correctness unverified and vulnerable to regressions from layout changes, endianness assumptions, or buffer-boundary bugs. Adding round-trip tests ensures that any value passed through encode → decode returns the same logical value, and that truncated buffers panic as expected rather than producing undefined behavior.

## Changes

| Category | Description |
|----------|-------------|
| **Added** | `#[cfg(test)] mod tests` with round-trip helpers `round_trip` and `round_trip_eq` |
| **Added** | Round-trip tests for `()`, `Address`, `RegisterOperand`, `VaryingOperand` |
| **Added** | Round-trip tests for primitives: `u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`, `f32`, `f64` |
| **Added** | Round-trip tests for tuples `(u8, u8)`, `(u32, u32)`, `(Address, RegisterOperand)` |
| **Added** | Round-trip tests for `ThinVec<u32>` and `ThinVec<RegisterOperand>` |
| **Added** | `test_encoded_size_matches_type_size` — verifies encoded byte lengths match `size_of` for `u32`, `u64`, `f64` |
| **Added** | `decode_truncated_buffer_panics` — ensures decoding from a too-small buffer panics with "buffer too small" |

## Technical Details

- **File modified:** `core/engine/src/vm/opcode/args.rs`
- **Lines changed:** +114 (test module only)
- **Behavioral impact:** None — tests only; no production code changes

The `round_trip` helper encodes a value, decodes from the same buffer, and asserts equality (for types implementing `PartialEq`). The `round_trip_eq` helper accepts a custom comparison function for types like `RegisterOperand` and `VaryingOperand` that do not implement `PartialEq`, comparing via `u32::from()`. Float tests use `0.0` and `1.5` only; `NaN` is excluded because IEEE 754 defines `NaN != NaN`. The panic test uses `#[should_panic(expected = "buffer too small")]` to assert the `read` function's buffer-length check triggers correctly when decoding `u32` from a 2-byte slice.

## Testing

- [x] `cargo test -p boa_engine --lib` — all 9 new tests pass
- [x] `test_unit_round_trip`, `test_address_round_trip`, `test_register_operand_round_trip`, `test_varying_operand_round_trip`
- [x] `test_primitive_round_trips`, `test_tuple_round_trips`, `test_thin_vec_round_trip`
- [x] `test_encoded_size_matches_type_size`, `decode_truncated_buffer_panics`
- [x] `cargo build` — successful compilation

### DIFF
--- a/core/engine/src/vm/opcode/args.rs
+++ b/core/engine/src/vm/opcode/args.rs
@@ -221,7 +221,7 @@ impl_argument_for_int!(u8 u16 u32 u64 i8 i16 i32 f32 f64);
 
 #[cfg(test)]
 mod tests {
-    use super::{Argument, Address, RegisterOperand, VaryingOperand};
+    use super::{Address, Argument, RegisterOperand, VaryingOperand};
     use std::mem::size_of;
     use thin_vec::ThinVec;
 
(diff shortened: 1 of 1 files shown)


tkshsbcue:
@jedel1043 i hope you have a look into this
Thanks!

PR #4917, opened 2026-03-07
perf(bytecompiler): skip PopIntoRegister for discarded expression results

hansl:
Add `compile_expr_for_side_effects` that compiles expressions purely for side effects without storing the result. For call expressions, this emits `Pop` instead of `PopIntoRegister`, saving an instruction, a register allocation, and a JsValue clone+drop per discarded call.

Used in expression statements (when result is unused) and for-loop final expressions, which are the two main sites where expression results are compiled into temporary registers that are immediately deallocated.

----

Another optimization:

When a call/new result is needed on the stack (as an argument, return
value, etc.), the compiler previously popped into a temporary register
then immediately pushed it back. This adds a `CallResultDest::Stack`
variant that leaves the result on the stack directly, and a
`compile_expr_to_stack` method that uses it for Call/New expressions.

Also handles `Expression::Parenthesized` unwrapping so that patterns
like `(new Foo(...))` (common in the EarleyBoyer benchmark) are
properly optimized. This eliminates all 3658 adjacent same-register
PopIntoRegister+PushFromRegister pairs in the V8 benchmark suite.

### DIFF
--- a/tests/insta-bytecode/src/snapshots/insta_bytecode__double_loop_function.snap
+++ b/tests/insta-bytecode/src/snapshots/insta_bytecode__double_loop_function.snap
@@ -14,28 +14,28 @@ Location     Handler      Opcode                            Operands
   000037                    IncrementLoopIteration            
   000038                    Inc                               src:r02, dst:r02
   000041                    PushInt8                          value:20, dst:r03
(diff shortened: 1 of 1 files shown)


nekevss:
Oh nice! Looks like the bytecode integration tests need to be updated :smile:

PR #4918, opened 2026-03-07
Trivial functions optimizations

hansl:
3 optimizations:

----

Inline immediately invoked function expressions (IIFEs) at compile time,
eliminating function object allocation, call frame creation, and
environment setup overhead. Supports arrow functions and function
expressions with simple parameters, handling expression bodies,
void/side-effect bodies (no return), and bodies with a trailing return.

----

Skip frame creation for empty functions (e.g., `function f() {}`) by
detecting them at compile time and using CallValue::Complete at runtime,
similar to native function calls.

---

Skip frame creation for empty constructors (e.g., `function Foo() {}`)
that have no fields or private methods and are not derived. Creates the
object and returns it directly via CallValue::Complete, avoiding the
full frame push/pop cycle.

----

This brings the `call.js` benchmark to within 1.9x of QuickJS:

```
Benchmark 1: ./target/release/boa /Users/hansl/call.js
  Time (mean ± σ):     578.9 ms ±   1.8 ms    [User: 570.3 ms, System: 4.7 ms]
  Range (min … max):   577.2 ms … 583.3 ms    10 runs

Benchmark 2: qjs /Users/hansl/call.js
  Time (mean ± σ):     307.6 ms ±   0.3 ms    [User: 304.1 ms, System: 2.2 ms]
  Range (min … max):   307.1 ms … 308.1 ms    10 runs

Summary
  qjs /Users/hansl/call.js ran
    1.88 ± 0.01 times faster than ./target/release/boa /Users/hansl/call.js
```

### DIFF
--- a/core/engine/src/bytecompiler/inline.rs
+++ b/core/engine/src/bytecompiler/inline.rs
@@ -17,7 +17,9 @@ use boa_ast::{
     Expression, Statement, StatementListItem,
     declaration::Binding,
     function::{ArrowFunction, FunctionExpression},
-    operations::{ContainsSymbol, annex_b_function_declarations_names, contains, var_scoped_declarations},
+    operations::{
+        ContainsSymbol, annex_b_function_declarations_names, contains, var_scoped_declarations,
+    },
     visitor::{VisitWith, Visitor},
 };
 
--- a/core/engine/src/bytecompiler/mod.rs
+++ b/core/engine/src/bytecompiler/mod.rs
@@ -2340,21 +2340,21 @@ impl<'ctx> ByteCompiler<'ctx> {
     fn call(&mut self, callable: Callable<'_>, dst: CallResultDest<'_>) {
         // Try to inline IIFE calls: ((a, b) => expr)(x, y)
         // Only when the result goes into a register (the common expression case).
-        if let CallResultDest::Register(dst) = dst {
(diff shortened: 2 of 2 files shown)


core/engine/src/bytecompiler/inline.rs
```diff
@@ -0,0 +1,472 @@
...
+        }
+        if scopes.parameters_eval_scope().is_some() {
+            return false;
+        }
+
+        true
+    }
+
+    /// Check if a function expression is eligible for inlining.
+    fn is_function_expr_inlinable(&self, func: &FunctionExpression) -> bool {
```
jedel1043, line 155: This is _very_ inefficient. We're basically going up and down through the whole AST trying to find specific AST nodes, multiple times. Preferrably we should implement a visitor that bails if it finds any of the patterns we're testing for.

core/engine/src/bytecompiler/inline.rs
jedel1043, line 1: This whole file is almost just inspecting the AST and not the bytecode, so it kinda makes more sense to implement it as an AST optimization, right? Something like replacing the IIFE with a block.

hansl: Maybe. I don't know how easy it would be to build an IIFE => block AST optimization. It could be done I presume.

PR #4920, opened 2026-03-07
fix(parser): replace unimplemented! with syntax error for unknown ope…

tkshsbcue:
# fix(parser): replace unimplemented! with syntax error for unknown operators

## Summary

Replaces `unimplemented!("operator {}", op)` in the operator lexer with proper error handling. When the lexer encounters an unknown operator byte (i.e. a byte that does not start any valid ECMAScript operator), it now returns a recoverable `Error::syntax` with a clear message instead of panicking and crashing the parser process.

## Motivation

The operator lexer (`Operator` in `core/parser/src/lexer/operator.rs`) tokenizes ECMAScript operators such as `+`, `-`, `*`, `%`, `|`, `&`, `?`, `^`, `=`, `<`, `>`, `!`, and `~`. The `lex()` method matches the initial byte (`self.init`) against these known operator starts; any other byte falls through to a catch-all arm. Previously, that arm used `unimplemented!("operator {}", op)`, which panics when the lexer hits an unexpected byte—e.g. from malformed input, a non-ASCII character in an operator position, or a future operator not yet implemented. For Boa as an embedded engine or in tooling (linters, formatters, REPLs), panics are unacceptable: the host expects to receive a parse error and continue. Replacing the panic with `Err(Error::syntax(...))` aligns with Boa's stability goals ("Fewer panics, better error handling") and ensures the parser fails gracefully with a descriptive syntax error that includes the offending character and position.

## Changes

| Category   | Description |
|-----------|-------------|
| **Replaced** | `op => unimplemented!("operator {}", op)` with `op => { return Err(Error::syntax(format!("unexpected operator '{}'", char::from(op)), start_pos.position())); }` |
| **Preserved** | All existing operator arms (`*`, `+`, `-`, `%`, `|`, `&`, `?`, `^`, `=`, `<`, `>`, `!`, `~`) unchanged |

The new error message includes the unexpected character (via `char::from(op)`) and the source position (via `start_pos.position()`), making it easier for users and tooling to diagnose invalid input.

## Technical Details

- **File modified:** `core/parser/src/lexer/operator.rs`
- **Lines changed:** +6, -1
- **Behavioral impact:** Invalid operator bytes that previously caused a panic now produce `Err(Error::syntax("unexpected operator 'X'", position))`. Valid operator input is unchanged; the catch-all is only reached for bytes not matched by any operator arm.

The `lex()` method returns `Result<Token, Error>`, and `Error::syntax(msg, pos)` is the standard way to report syntax errors in the parser. The change fits the existing error-handling pattern used elsewhere in the lexer.

## Testing

- [x] `cargo clippy -p boa_parser` — no warnings
- [x] `cargo build` — successful compilation

### DIFF
--- a/core/parser/src/lexer/operator.rs
+++ b/core/parser/src/lexer/operator.rs
@@ -164,7 +164,10 @@ impl<R> Tokenizer<R> for Operator {
             }
             op => {
                 return Err(Error::syntax(
-                    format!("unexpected operator '{}'", char::from(op)),
+                    format!(
+                        "unexpected character '{}' in operator",
+                        char::from_u32(u32::from(op)).unwrap_or('\u{FFFD}')
+                    ),
                     start_pos.position(),
                 ));
             }


tkshsbcue:
Hey! (@jedel1043 & @hansl ) I noticed the operator lexer was using unimplemented!() for unknown operator bytes, which panics on invalid input. That’s rough for anyone embedding Boa—a single bad character can crash the whole process instead of getting a parse error.

This PR swaps that panic for a proper syntax error. Now when the lexer hits something it doesn’t recognize, it returns an Error::syntax with the character and position, so callers can handle it and keep going. Same behavior for valid code, just a safer path for the weird cases.

tkshsbcue:
@nekevss i hope you can review my PR
Thanks!

jedel1043:
Need to resolve the conflicts

core/parser/src/lexer/operator.rs
```diff
@@ -162,7 +162,15 @@ impl<R> Tokenizer<R> for Operator {
             b'~' => {
                 Token::new_by_position_group(Punctuator::Neg.into(), start_pos, cursor.pos_group())
             }
-            op => unimplemented!("operator {}", op),
+            op => {
+                return Err(Error::syntax(
+                    format!(
+                        "unexpected character '{}' in operator",
+                        char::from_u32(u32::from(op)).unwrap_or('\u{FFFD}')
```
jedel1043, line 169: It might be a bad idea to replace with the unknown char in this case, or at least it would make errors much harder to debug.

We should print the u32 char value in hex if the u32 is not a valid char instead.

jedel1043:
Please verify what your LLM generates before opening PRs.

tkshsbcue:
You're right, I should have reviewed this more carefully before submitting. Thanks for the feedback

PR #4921, opened 2026-03-07
Add bytecode snapshot tests using `insta` crate

nekevss:
As mentioned on Matrix, let's add some snapshot tests for the bytecode.

This PR is primarily designed to allow us to better diff and review changes to the bytecompiler.

It also introduces an extra way for us to test and produce artifacts that we can track and manage.


tests/insta-bytecode/src/lib.rs
```diff
@@ -0,0 +1,44 @@
+use std::{fs::File, path::PathBuf, process::{Command, Stdio}};
+
+pub const MANIFEST_DIR: &str = env!("CARGO_MANIFEST_DIR");
+
+pub fn js_directory() -> PathBuf {
+    PathBuf::from(MANIFEST_DIR).join("js")
+}
+
+pub fn collect_file_trace(file_path: PathBuf) -> String {
```
hansl: Was wishing you'd use the Boa library directly and not the CLI. This means we need to have it in our PATH and compiled to latest.

tests/insta-bytecode/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+    if result.status.success() {
+        return String::from_utf8_lossy(&result.stdout).to_string()
+    } else {
+        let failure_msg = String::from_utf8_lossy(&result.stderr).to_string();
+        panic!("boa failed: {}", failure_msg);
+    }
+}
+
+#[test]
+fn basic_loop() {
```
hansl: Could you have used `rstest` to generate test cases for every JS files in the `js` folder? That way you would avoid copy pasting the same code for every new JS file.

tests/insta-bytecode/src/lib.rs
nekevss: This is mostly likely short term until we have some support for trace that isn't printing to stdout.

Once we handle that, we can probably support `boa_engine` directly.

Although, I'm not entirely clear how much of a downside having boa_cli installed would be when most people have `.cargo/bin` in your path already and you can run `cargo install --path boa_cli` from root.

tests/insta-bytecode/src/lib.rs
nekevss: As far as I can tell, the function names are used to output to the snap files. If we have one function name, then we will have "test_function", "test_function_1", "test_function_2", and so on for all the tests.

tests/insta-bytecode/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+        let failure_msg = String::from_utf8_lossy(&result.stderr).to_string();
+        panic!("boa failed: {}", failure_msg);
+    }
+}
+
+macro_rules! test_case {
+    ($fn_name:ident, $js_filename:literal) => {
+        #[test]
+        fn $fn_name() {
+            let output = collect_file_trace(js_directory().join("basicLoop.js"));
```
nekevss: oh man, i stayed up way too late 😅

hansl (approved):
So much cleaner, and probably much faster. We can do enumeration of all files later. LGTM

tests/insta-bytecode/src/lib.rs
hansl: The `#[rstest]` proc macro creates function names based on the file path, so that would still work. It's a matter of ordering attributes.

Issue #4444, opened 2025-09-25
Implement iterator helpers proposal


jedel1043:
https://github.com/tc39/proposal-iterator-helpers

Gourav2026:
@jedel1043  i want to work on this , please assign this to me

sarthakNITT:
Hi @Gourav2026, Are you still working on this, or is it free to be picked up?

Gourav2026:
@sarthakNITT i am working on it bro

Gourav2026:
@jedel1043  please unassign me as i am busy for my exams
@sarthakNITT please take it over

batmnnn:
can i be assigned to this issue @jedel1043

sarthakNITT:
@Gourav2026 sure, @jedel1043 Would love to get assigned. I'd like to work on this.

I plan to implement the iterator helpers by adding the `IteratorHelperPrototype` object and all 10+ methods (map, filter, take, drop, flatMap, reduce, toArray, forEach, some, every, find) following the existing builtins pattern in Boa. Will start with the spec steps from the proposal and add tests from `test262`.

sarthakNITT:
@nekevss I've raised a pr. Please review it and let me know in case of further changes.

PR #4922, opened 2026-03-07
feat: implement Iterator Helpers proposal

sarthakNITT:
- Added Iterator global constructor with Iterator.from() static method
- Added 11 prototype methods on %IteratorPrototype%: map, filter, take, drop, flatMap, reduce, toArray, forEach, some, every, find
- Added IteratorHelperPrototype and WrapForValidIteratorPrototype intrinsic objects
- Added 13 tests covering all new methods

issue #4444


jedel1043:
Duplicate of #4810

PR #4926, opened 2026-03-07
fix(job): resolve 100% CPU usage when idling on future timeouts

ashnaaseth2325-oss:
## Summary
This PR fixes a performance bug where the `SimpleJobExecutor` pins a CPU core at 100% while waiting for a `setTimeout` or `setInterval` delay. The issue occurs because the executor doesn't have a way to "sleep" when there are no active async jobs to poll, causing it to spin in a tight loop until the timer expires.

## Steps to Reproduce
Run any script with a `setTimeout` delay:

```javascript
setTimeout(() => console.log("Done"), 5000);

```

During the 5 second wait, the engine burns 100% of a CPU core despite having no work to do.

## Problem

The current `run_jobs_async` loop has a logic gap. We prevent the loop from terminating (because a timeout is pending), but we only yield execution if there are active async futures. If the async group is empty, the engine falls into a "tight spin" ; it stays awake but has nothing to execute, leading to unnecessary high CPU usage.

## Fix

I updated the loop guards to recognize the "idle wait" state. If the sync queues are empty and no async work is active, the executor now calculates the time remaining until the next scheduled `TimeoutJob` and parks the thread until that deadline.

## Result

* CPU usage during timer delays drops from 100% to effectively 0%.
* Periodic tasks (like heartbeats) now consume negligible resources between ticks.
* The engine is now stable for long running server or embedded usage.

This fix significantly improves the efficiency of the `SimpleJobExecutor` for any host environment using standard JS timers.

### DIFF
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -49,6 +49,7 @@ use std::mem;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::sync::atomic::Ordering;
+use std::task::Poll;
 use std::{cell::RefCell, collections::VecDeque, fmt::Debug, future::Future, pin::Pin};
 
 /// An ECMAScript [Job Abstract Closure].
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @jedel1043 @hansl 
This PR fixes a busy spin in SimpleJobExecutor when waiting for future timeouts. The executor now waits for the next timeout instead of continuously polling.
All CI checks are passing. Thank You!

ashnaaseth2325-oss:
Hello @jedel1043 @hansl @nekevss 
Just a gentle follow up.
Kindly view the pr whenever you have time . Thanks!

core/engine/src/job.rs
```diff
@@ -763,7 +768,51 @@ impl JobExecutor for SimpleJobExecutor {
...
+                    .filter(|&d| d > now);
+
+                if let Some(deadline) = deadline {
+                    let dur = time::Duration::from(deadline - now);
+                    if !dur.is_zero() {
+                        let (tx, mut rx) = oneshot::channel::<()>();
+                        std::thread::spawn(move || {
+                            std::thread::sleep(dur);
+                            let _ = tx.send(());
+                        });
```
jedel1043, line 796: This is very inefficient. We are essentially creating a whole thread to just wait for `dur` milliseconds.

I don't think this is something we can fix on the `SimpleJobExecutor`. This executor should be kept as simple as possible, but to properly support this we need an async executor to sleep using an async timer.

jedel1043:
Related: #4994. Uses no additional threads by integrating with `smol`

ashnaaseth2325-oss:
Hello @jedel1043 Thanks for the feedback! That’s a fair point! Spawning a thread just to wait for dur isn’t ideal for a simple executor. I agree this is probably better handled by a proper async executor.

PR #4931, opened 2026-03-07
Add MVP of virtual machine tracer

nekevss:
This pull request implements a new `VirtualMachineTracer` trait that emits `VirtualMachineEvents` rather than immediately printing the values to stdout.

This should allow better customization of event handling.


core/engine/src/vm/trace.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// A message that emits details about a call frame
+#[derive(Debug, Clone)]
+pub struct CallFrameMessage {
+    pub bytecode: String,
+}
+
+/// A message that emits instruction execution details about a call frame
+#[derive(Debug, Clone)]
+pub struct OpcodeExecutionMessage {
```
jedel1043, line 209: It might be better to granularize this, because some tracers would need to have the individual operands and stack values instead of a big string; I'm thinking of like a web-based tracer that displays the stack in a more dynamic way, for example.

nekevss: Ah, sorry just saw this!

Thats a good idea. I'll make the update ASAP

jedel1043:
Any updates on this?

nekevss:
Still planning to eventually take another pass at this when I have a bit of time.

PR #4932, opened 2026-03-07
feat(engine): add JsUint8Array::to_vec for copying viewed bytes

### LINKED ISSUE #4722: Rust API improvement: extracting bytes from `JsTypedArray`
This is about the Rust API for `JsTypedArray`, not new JS syntax

While experimenting with embedding Boa from Rust, I noticed that creating typed arrays from iterators is straightforward, but getting the data back out as bytes means going through JS properties or manual iteration / extra JS calls This works, but feels a bit awkward compared to the rest of the Rust API ergonomics

For embedding use cases, typed arrays are often just treated as byte buffers, so a small helper on `JsTypedArray` could make this workflow simpler and avoid some boilerplate and extra allocations on the Rust side

I’m not sure if something like this already exists internally and I just missed it, but I couldn’t find a direct helper in the public API

I know the data lives behind `ArrayBuffer` / `SharedArrayBuffer` and is managed by the GC, so I’m not asking for raw pointers or a long‑lived `&[u8]` A simple copying helper seems like a reasonable starting point Initially this helper could just target `Uint8Array` (byte‑oriented buffers), with room to extend to other typed array kinds later if a byte view makes sense there

**Example code**

```rust
use boa_engine::{
    Context, JsResult,
    object::builtins::JsUint8Array,
}

fn use_bytes(context: &mut Context) -> JsResult<()> {
    let data: Vec<u8> = (0..=255).collect()
    let array = JsUint8Array::from_iter(data, context)?

    // safe API that copies the bytes
    let bytes_vec: Vec<u8> = array.to_vec_u8(context)?
    assert_eq!(bytes_vec.len(), 256)

    Ok(())
}
```

**Implementation checklist (rough)**

- [x] Add a copying helper on `JsTypedArray`, e.g. `to_vec_u8(&self, ctx: &mut Context) -> JsResult<Vec<u8>>`
- [x] Document how this works with the underlying `ArrayBuffer` / `SharedArrayBuffer` and why it’s GC safe
- [x] Add tests that round‑trip data between `Vec<u8>` and `JsTypedArray`
- [x] Update Rust API docs to show this pattern for embedding code

If this direction makes sense for the API, I’d like to work on it

alienx5499:
This Pull Request fixes/closes #4722.

It changes the following:
- Add `SliceRef::to_vec()` helper to copy buffer data (including `AtomicSlice`) into `Vec<u8>`
- Add `JsUint8Array::to_vec(&mut Context)` to copy the viewed byte range of a `JsUint8Array` into a `Vec<u8>`
- Add a small round-trip test (`uint8_array_to_vec_roundtrip`) and a doc example for `JsUint8Array::to_vec`

Testing:
- `cargo test -p boa_engine` (including the new unit test and doctest)


alienx5499:
Hi @jedel1043,
I’d appreciate a review when you get time.

alienx5499:
Hi @jedel1043,
I pushed the Clippy fix, which disabled auto merge due to branch permissions.
Could you re enable auto-merge when convenient? Thanks.

alienx5499:
Hi @jedel1043, updating the branch from my side may disable auto merge due to permissions.
Will the merge queue proceed as is, or would you prefer to update it?

jedel1043:
Don't need to update, there are no merge conflicts

Issue #4934, opened 2026-03-07
fix(display): ArrayBuffer display format inconsistency in console/debug output


sahanaxzy:
The display logic in `core/engine/src/value/display.rs` (around line 288) includes a FIXME that ArrayBuffer cannot be reliably distinguished when formatting values for `console.log` and debug output.

**Current state:** The display path uses a heuristic based on the `"name"` property, which does not correctly handle ArrayBuffer in some cases.

**Proposed fix:** Add proper handling for ArrayBuffer (and related built-ins) in the display logic so the output is correct and consistent with other engines.

**Note:** `Object.prototype.toString` already works for ArrayBuffer via `@@toStringTag`. This only affects the Display / `console.log` formatting path.

shreyadesai20042009:
Hi, I would like to work on this issue. Is it still available?

sahanaxzy:
Hi @shreyadesai20042009 
Thanks for reaching out. I’ve already opened PR #4935 that addresses this issue, and it’s currently awaiting review from the maintainers.

PR #4935, opened 2026-03-07
fixed show ArrayBuffer constructor as [Function: ArrayBuffer]

sahanaxzy:
## Summary

Fixes how constructors are shown in console/debug output so that only real class constructors use `[class Name]`, while built-ins like ArrayBuffer use `[Function: Name]`.

Solves #4934 

## Changes

Modified `core/engine/src/value/display.rs`
> only shows `[class Name]` when the value is an `OrdinaryFunction` with `is_class_constructor()`; otherwise shows callables as `[Function: Name]`.

Replace `.map(...).unwrap_or(false)` with `.is_some_and(...)` to satisfy Clippy.

### DIFF
--- a/core/engine/src/value/display.rs
+++ b/core/engine/src/value/display.rs
@@ -2,8 +2,8 @@ use super::{Display, HashSet, JsValue, JsVariant, fmt};
 use crate::{
     JsError, JsObject, JsString,
     builtins::{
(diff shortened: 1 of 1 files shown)


sahanaxzy:
@jedel1043 can you review this..

PR #4936, opened 2026-03-07
fix(gc): implement better way to handle root overload in `inc_non_root_count`

shruti2522:
It changes the following:

saturating at `NON_ROOTS_MAX` pins the counter instead of panicking. pinned counter returns true for `is_rooted` so the object is kept alive rather than freed, a leak is safer than a UAF anyways. The limit ~2^31 heap references to one object is unreachable in practice


jedel1043:
Well that would make it so that we underflow after all references get dropped. right? Then we're patching one panic with another one

shruti2522:
> Well that would make it so that we underflow after all references get dropped. right? Then we're patching one panic with another one

oh yes, missed that 😅 let me look into this more carefully

shruti2522:
I have capped `non_root_count` so it can never exceed `ref_count`. This guarantees `is_rooted` will never return false just because the counter overflowed, so we won't accidentally free a live object and cause a UAF

jedel1043:
Ping me after you address the CI errors so that I can merge

shruti2522:
> Ping me after you address the CI errors so that I can merge

@jedel1043 fixed the CI, ready for merge

PR #4938, opened 2026-03-08
feat(gc): make WeakMap::get return a reference instead of cloning

shruti2522:

It changes the following:

`WeakMap::get` now returns an `Option<GcRef<'_, V>>` instead of `Option<V>`


jedel1043:
This is unsound. If you get a reference to the inner value of the ephemeron and a garbage collection occurs, you could be left with a reference to deallocated data.

jedel1043:
Closing this for now, since it doesn't make sense to merge something that could cause UB. Feel free to reopen if you find a way to support this without causing unsoundness.

shruti2522:
I looked into this more carefully and you are correct, it is always going to be unsound in a tracing GC. My attempt was to resolve `TODO: return a reference instead of cloning`. the only workaround to safely avoid cloning would be a callback pattern (like with_get(|val| ...)), but since `JsValue` is just an O(1) RC clone it's completely unnecessary

PR #4939, opened 2026-03-08
feat(gc): add configurable GC threshold and used space percentage

shruti2522:

It changes the following:

- Exposed `GcConfig` struct and `configure_gc` to allow customizing GC limits
- Integrates GC config into `ContextBuilder` via `gc_config`


shruti2522:
> I'm not entirely sure whether we want to be adding this functionality to the current GC or who it would benefit. Ideally, we replace the GC before the next release if possible.

yeah, you're right. It's not worth adding it, I guess we can close this

PR #4940, opened 2026-03-08
fix: replace unreachable!/todo! with proper error handling and no-op

tkshsbcue:
# fix: replace unreachable!/todo! with proper error handling and no-op

## Summary

Two stability improvements: (1) Replace `unreachable!()` in the hoistable and lexical declaration parsers with proper `Error::general` so unexpected tokens produce parse errors instead of panics. (2) Implement `access_set` for `Access::This` in the bytecompiler—in sloppy mode, assignment to `this` is a no-op per ECMAScript; previously `todo!()` would panic.

## Motivation

**Parser unreachable!** — The hoistable and lexical declaration parsers used `unreachable!("unknown token found: {:?}", tok)` in their catch-all arms. In theory the caller ensures only valid tokens reach these parsers, but if a bug or edge case produces an unexpected token, the parser would panic. For embedded use and tooling, panics are unacceptable. Returning a parse error lets the host handle the failure gracefully.

**Bytecompiler todo!** — Assigning to `this` in sloppy mode (e.g. `this = 5`) is valid JavaScript; the assignment is a no-op (the RHS is evaluated for side effects only). The bytecompiler had `todo!("access_set `this`")`, so executing such code would panic. Implementing the no-op path removes this panic and aligns with the spec.

## Changes

| Category   | Description |
|-----------|-------------|
| **Parser** | `core/parser/src/parser/statement/declaration/hoistable/mod.rs`: `_ => unreachable!(...)` → `Err(Error::general("expected 'function', 'async', or 'class' in declaration", tok.span().start()))` |
| **Parser** | `core/parser/src/parser/statement/declaration/lexical.rs`: `_ => unreachable!(...)` → `Err(Error::general("expected 'let' or 'const' in lexical declaration", tok.span().start()))` |
| **Bytecompiler** | `core/engine/src/bytecompiler/mod.rs`: `Access::This => todo!(...)` → evaluate RHS via `expr_fn(self)` for side effects, then no-op (no assignment) |

## Technical Details

- **Files modified:** `core/parser/.../hoistable/mod.rs`, `core/parser/.../lexical.rs`, `core/engine/src/bytecompiler/mod.rs`
- **Behavioral impact:** Parser: unexpected tokens in declaration context now yield parse errors. Bytecompiler: `this = expr` in sloppy mode now executes without panicking (RHS evaluated, assignment ignored).
- **Strict mode:** Assignment to `this` in strict mode is a SyntaxError and is rejected by the parser, so the bytecompiler path is only reached in sloppy mode.

## Testing

- [x] `cargo test -p boa_engine --lib` — all 921 tests pass

### DIFF
--- a/core/engine/src/bytecompiler/mod.rs
+++ b/core/engine/src/bytecompiler/mod.rs
@@ -1471,7 +1471,7 @@ impl<'ctx> ByteCompiler<'ctx> {
                 },
             },
             Access::This => {
-                // In sloppy mode, assignment to `this` is a no-op per spec; we still evaluate
+                // In non-strict code, assignment to `this` is a no-op per spec; we still evaluate
                 // the RHS for side effects (e.g. `this = foo()` must call foo()).
                 let _ = expr_fn(self);
             }
--- a/core/parser/src/parser/statement/declaration/hoistable/mod.rs
+++ b/core/parser/src/parser/statement/declaration/hoistable/mod.rs
@@ -119,9 +119,15 @@ where
                     .parse(cursor, interner)
                     .map(Declaration::from)
             }
-            _ => Err(Error::general(
-                "expected 'function', 'async', or 'class' in declaration",
-                tok.span().start(),
+            _ => Err(Error::expected(
+                [
+                    Keyword::Function.to_string(),
+                    Keyword::Async.to_string(),
+                    Keyword::Class.to_string(),
+                ],
+                tok.to_string(interner),
+                tok.span(),
+                "declaration",
             )),
         }
     }
--- a/core/parser/src/parser/statement/declaration/lexical.rs
+++ b/core/parser/src/parser/statement/declaration/lexical.rs
@@ -92,9 +92,11 @@ where
             )
             .parse(cursor, interner)?,
             _ => {
-                return Err(Error::general(
-                    "expected 'let' or 'const' in lexical declaration",
-                    tok.span().start(),
+                return Err(Error::expected(
+                    [Keyword::Let.to_string(), Keyword::Const.to_string()],
+                    tok.to_string(interner),
+                    tok.span(),
+                    "lexical declaration",
                 ));
             }
         };


tkshsbcue:
@jedel1043 i hope you can review my PR
Thanks!

core/parser/src/parser/statement/declaration/hoistable/mod.rs
```diff
@@ -119,7 +119,10 @@ where
                     .parse(cursor, interner)
                     .map(Declaration::from)
             }
-            _ => unreachable!("unknown token found: {:?}", tok),
+            _ => Err(Error::general(
+                "expected 'function', 'async', or 'class' in declaration",
+                tok.span().start(),
+            )),
```
jedel1043: There's precisely an `Error::expected` method for these kinds of errors.

core/engine/src/bytecompiler/mod.rs
```diff
@@ -1470,7 +1470,11 @@ impl<'ctx> ByteCompiler<'ctx> {
                     }
                 },
             },
-            Access::This => todo!("access_set `this`"),
+            Access::This => {
+                // In sloppy mode, assignment to `this` is a no-op per spec; we still evaluate
```
jedel1043: is "sloppy mode" another name for non-strict code? I've never seen it being called like that haha

core/parser/src/parser/statement/declaration/hoistable/mod.rs
tkshsbcue: Switched to Error::expected for consistency with the rest of the parser.

core/engine/src/bytecompiler/mod.rs
tkshsbcue: Yes, “sloppy mode” is the ECMAScript term for non-strict code. I’ve updated the comment to use “non-strict” instead.

tkshsbcue:
@jedel1043 made some changes hope you can have a look

PR #4942, opened 2026-03-08
perf: batch atomic copies in ArrayBuffer using u64 chunks

xcb3d:
This Pull Request resolves the TODO comments at `core/engine/src/builtins/array_buffer/utils.rs:387` and `utils.rs:424` requesting batched atomic writes.

It changes the following:

- Add `batched_atomic_copy_forward` — copies `AtomicU8` arrays forward using aligned `AtomicU64` stores (8 bytes/op), with byte-by-byte fallback for unaligned head/tail
- Add `batched_atomic_copy_backward` — same strategy but in reverse, for overlapping `memmove` support
- Add `batched_copy_bytes_to_atomic` — optimized `*const u8` → `*const AtomicU8` copy using `ptr::read_unaligned` + aligned `AtomicU64` stores
- Add `batched_copy_atomic_to_bytes` — optimized `*const AtomicU8` → `*mut u8` copy using aligned `AtomicU64` loads + `ptr::write_unaligned`
- Update `copy_shared_to_shared`, `copy_shared_to_shared_backwards`, and `memcpy` to use the new batched helpers
- Remove the two TODO comments

All four helpers use a 3-phase pattern (head → aligned u64 body → tail) to ensure alignment safety. Source reads from `AtomicU8` arrays use individual atomic loads packed via `u64::from_ne_bytes` to avoid UB with concurrent writers.

### DIFF
--- a/core/engine/src/builtins/array_buffer/utils.rs
+++ b/core/engine/src/builtins/array_buffer/utils.rs
@@ -812,7 +812,7 @@ pub(crate) unsafe fn memmove(ptr: BytesMutPtr, from: usize, to: usize, count: us
 }
 
 #[cfg(test)]
-mod utils_tests {
+mod tests_miri {
     use super::*;
     use portable_atomic::AtomicU8;
     use std::sync::atomic::Ordering;


core/engine/src/builtins/array_buffer/utils.rs
```diff
@@ -373,8 +373,240 @@ impl<'a> From<&'a [AtomicU8]> for SliceRefMut<'a> {
...
+    // SAFETY: `dest + offset` is now 8-byte aligned (guaranteed by Phase 1).
+    // `AtomicU8` is `#[repr(transparent)]` over `u8`, so casting to `AtomicU64`
+    // is valid when properly aligned. We read source bytes individually through
+    // atomic loads to avoid UB with concurrent writers, then pack into u64.
+    #[allow(clippy::cast_ptr_alignment)]
+    unsafe {
+        let dest_u64 = dest.add(offset).cast::<AtomicU64>();
+        for i in 0..chunks {
+            let base = offset + i * BATCH_SIZE;
+            let mut bytes = [0u8; BATCH_SIZE];
```
jedel1043: This kinda negates the point of batch copying, since you're writing one by one at the end of the day.

What we should do instead is to batch write only if both buffers have the same misalignment, and fallback to normal byte copies otherwise

hansl:
Good work, but can you fix the test262 failures?

jedel1043:
failures are not related to this PR, it's just out of date

core/engine/src/builtins/array_buffer/utils.rs
```diff
@@ -373,8 +373,318 @@ impl<'a> From<&'a [AtomicU8]> for SliceRefMut<'a> {
...
+        if misalign == 0 {
+            0
+        } else {
+            (BATCH_SIZE - misalign).min(count)
+        }
+    };
+    // SAFETY: ensured by the caller — both pointers are valid for `count` bytes.
+    unsafe {
+        for _ in 0..head {
+            (*dest.add(offset)).store(
```
jedel1043: No need to create `offset`. You can just use the index of the for loop for the offset.

core/engine/src/builtins/array_buffer/utils.rs
jedel1043, line 1: There's a lot of duplicate logic in here. Maybe we could simplify this a bit by creating a function that takes a pointer and size, returning the head, middle and tail offsets required

jedel1043 (changes requested):

core/engine/src/builtins/array_buffer/utils.rs
```diff
@@ -375,6 +375,23 @@ impl<'a> From<&'a [AtomicU8]> for SliceRefMut<'a> {
...
+/// Given a pointer address and total byte count, computes the number of
+/// head bytes needed to reach 8-byte alignment, the number of aligned
+/// 8-byte chunks, and the number of remaining tail bytes.
+const fn compute_batch_offsets(ptr_addr: usize, count: usize) -> (usize, usize, usize) {
+    let misalign = ptr_addr % BATCH_SIZE;
+    let head = if misalign == 0 {
+        0
+    } else {
+        let needed = BATCH_SIZE - misalign;
+        if needed < count { needed } else { count }
```
jedel1043: where did the usize::min go?

xcb3d: .min() is a trait method from Ord, which isn't a const trait yet, so it can't be called inside the const fn compute_batch_offsets. The if needed < count { needed } else { count } is the standard workaround. Happy to add a comment noting this if you'd like!

core/engine/src/builtins/array_buffer/utils.rs
```diff
@@ -484,49 +486,50 @@ unsafe fn batched_atomic_copy_backward(src: *const AtomicU8, dest: *const Atomic
         return;
     }
 
-    let mut end = count;
+    // Compute offsets from the end: peel tail bytes to align the end, then
+    // process aligned 8-byte chunks, leaving remaining head bytes.
+    let end_addr = (dest as usize).wrapping_add(count);
```
jedel1043: Why use wrapping add specifically here? Everything else uses the usual operators.

Also, was this too complex to use `compile_batch_offsets`? I kinda see that it could be simpler though, you just start from `ptr.add(count - tail)`, then from `ptr.add(head)` and finally from `ptr`, doing something like `for i in (0..tail).rev()` on each phase

core/engine/src/builtins/array_buffer/utils.rs
jedel1043: Okay so my question is, why is the function const if it will only get called with runtime values?

xcb3d: Done! Removed the const qualifier and restored .min() — you're right, the inputs are always runtime pointer addresses so const provided no benefit. See 3fdd0705.

core/engine/src/builtins/array_buffer/utils.rs
xcb3d: You're right on both counts — I over-complicated the backward copy when it could just reuse compute_batch_offsets
 and .rev() each phase. Thanks for the guidance, fixed in b614e444!

PR #4943, opened 2026-03-08
test: improve insta snap

hansl:
This reduces the amount of work to add new bytecode snapshot tests. Now we simply need to add a single file.


nekevss (approved):
Fine by me

nekevss:
As an additional note, part of the reason I moved away from JS files is that GitHub Actions wasn't liking the file read. 

And I can't remember how to fix it if it is even possible

PR #4945, opened 2026-03-08
Add Latin1 Encoding Check in `from` Method

### LINKED ISSUE #4944: Enhancement: Add Latin1 Encoding Check in `from` Method
## Location
`core/string/src/lib.rs`

## Summary
The `from` method currently checks whether the input string is ASCII using `s.is_ascii()`. However, it does not account for Latin1 encoding, which is a single-byte character encoding that extends ASCII with additional characters.

Supporting Latin1 detection would allow the method to handle a wider range of strings efficiently, since Latin1 characters still fit within a single byte representation.

## Current Behavior
- The method only verifies whether the string is ASCII.
- Non-ASCII characters are treated as requiring a different representation, even when they fall within the Latin1 range.

## Expected Behavior
The method should also detect whether the input string is Latin1 encodable.

Latin1 encoding includes characters in the range:

rajat552:
## Summary
This pull request introduces a check for Latin1 encoding in the `from` method of the `JsString` implementation. Previously, the method only handled ASCII strings, but this update extends support to strings containing characters within the Latin1 range (`0x00` to `0xFF`).

## Changes
- Added a condition to check whether all characters in the input string fall within the Latin1 range.
- Updated the `from` method to handle Latin1-compatible strings by converting them to `JsStr::latin1`.

## Testing
- Ran the full test suite to verify the changes.
- All existing tests passed successfully, confirming that no regressions were introduced.

## Impact
This change improves string handling by allowing the `from` method to support Latin1-encoded strings. As a result, the implementation becomes more robust and capable of handling a wider range of character inputs efficiently.

Closes #4944

### DIFF
--- a/core/string/src/display.rs
+++ b/core/string/src/display.rs
@@ -19,7 +19,7 @@ impl fmt::Display for JsStrDisplayEscaped<'_> {
             JsStrVariant::Latin1(v) => v
                 .iter()
                 .copied()
-                .map(|b| char::from_u32(b as u32).unwrap())
+                .map(|b| char::from_u32(u32::from(b)).unwrap())
                 .try_for_each(|c| f.write_char(c)),
             JsStrVariant::Utf16(_) => self.inner.code_points().try_for_each(|r| match r {
                 CodePoint::Unicode(c) => f.write_char(c),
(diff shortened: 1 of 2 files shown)


rajat552:
@jedel1043 review this PR let me know if any changes required

jedel1043:
Duplicate of #4896

PR #4951, opened 2026-03-08
 Improve Root Overflow Handling in GC Header

### LINKED ISSUE #4950: # Improve Handling of Root Count Overflow in `gc_header.rs`
## Location
`core/gc/src/internals/gc_header.rs`

## Summary
The current implementation panics when the root count exceeds its maximum value (root overload). While this prevents undefined behavior, panicking is not always the best way to handle this situation in a runtime system.

A more robust approach would allow the system to handle root overflow gracefully instead of immediately terminating execution.

## Problem
Currently:

- The root counter tracks how many active roots reference a GC-managed object.
- When this counter exceeds its maximum allowed value, the implementation triggers a panic.
- This behavior can abruptly terminate the program and makes error handling difficult for callers.

## Proposed Improvement
Instead of panicking, the system could handle root overflow in a more controlled way.

Possible approaches include:

1. **Logging a warning or error**
   - Emit a warning when the root counter approaches or exceeds the limit.

2. **Returning a `Result`**
   - Modify the function to return a `Result` type so the caller can handle the overflow condition explicitly.

3. **Graceful fallback**
   - Refuse to increment the root count when the limit is reached.
   - Signal the caller that the operation failed.

rajat552:
## Summary
This change improves how root overload is handled in the garbage collector's header implementation.

Previously, when the non-root count exceeded its maximum value, the code would panic and crash the program. This update replaces the panic with a safer error-handling approach.

## Changes
- Modified the increment function to return a `Result` type instead of panicking when the root count exceeds its maximum.
- Introduced proper error propagation so callers can handle overflow situations gracefully.
- Updated all usages of the increment function to accommodate the new error handling mechanism.

## Benefits
- Prevents unexpected runtime crashes caused by root count overflow.
- Allows callers to respond to overflow conditions safely.
- Improves the robustness and reliability of the garbage collector.

## Impact
This change makes the garbage collector more resilient by replacing a panic-based failure mode with controlled error handling.

Closes #4950

### DIFF
--- a/core/gc/src/internals/ephemeron_box.rs
+++ b/core/gc/src/internals/ephemeron_box.rs
@@ -109,8 +109,8 @@ impl<K: Trace + ?Sized, V: Trace> EphemeronBox<K, V> {
     }
 
     #[inline]
-    pub(crate) fn inc_non_root_count(&self) {
-        self.header.inc_non_root_count();
+    pub(crate) fn inc_non_root_count(&self) -> Result<(), &'static str> {
+        self.header.inc_non_root_count()
     }
 }
 
--- a/core/gc/src/internals/gc_box.rs
+++ b/core/gc/src/internals/gc_box.rs
@@ -45,8 +45,8 @@ impl<T: Trace + ?Sized> GcBox<T> {
     }
 
     #[inline]
-    pub(crate) fn inc_non_root_count(&self) {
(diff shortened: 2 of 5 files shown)


jedel1043:
This is not safer, it's arguably worse. It will also cause a panic on underflow when all the Gc's get dropped.

PR #4955, opened 2026-03-08
Remove opcodes related to globals instantiation

jedel1043:
This removes a whooping 6 opcodes from our VM; mostly everything related to global initialization. Instead of that, we add the `GlobalDeclarationInstantiation` and `EvalDeclarationInstantiation` methods for `Context`, which can take any codeblock and throws an error if any global initialization fails.

### DIFF
--- a/core/engine/src/context/mod.rs
+++ b/core/engine/src/context/mod.rs
@@ -885,7 +885,11 @@ impl Context {
     /// This is the common path of the instantiations:
     /// - `EvalDeclarationInstantiation ( body, varEnv, lexEnv, privateEnv, strict )`
     /// - `GlobalDeclarationInstantiation ( script, env )`
-    fn create_globals(&mut self, codeblock: &CodeBlock, is_eval_call: bool) -> JsResult<()> {
+    fn create_globals(
+        &mut self,
+        codeblock: &CodeBlock,
+        configurable_globals: bool,
+    ) -> JsResult<()> {
         // 8. For each element d of varDeclarations, in reverse List order, do
         //    a. If d is not either a VariableDeclaration, a ForBinding, or a BindingIdentifier, then
         //       i. Assert: d is either a FunctionDeclaration, a GeneratorDeclaration, an AsyncFunctionDeclaration, or an AsyncGeneratorDeclaration.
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/eval/mod.rs
```diff
@@ -335,7 +332,7 @@ impl Eval {
         let environments = context.vm.frame().environments.clone();
         let realm = context.realm().clone();
         context.vm.push_frame_with_stack(
-            CallFrame::new(code_block, None, environments, realm)
+            CallFrame::new(code_block.clone(), None, environments, realm)
```
hansl, line 335: Is this needed? We don't use code_block anywhere after this

jedel1043: yeah we do, in the shiny new `eval_declaration_instantiation` function

I could make the context use its current active codeblock to do the initialization though, and it might be arguably cleaner

core/engine/src/bytecompiler/declarations.rs
```diff
@@ -635,14 +627,19 @@ impl ByteCompiler<'_> {
     pub(crate) fn eval_declaration_instantiation(
         &mut self,
         body: &Script,
-        #[allow(unused_variables)] strict: bool,
+        #[allow(
+            unused_variables,
+            reason = "only used when the `annex-b` feature is enabled"
```
hansl, line 632: ❤️

core/engine/src/context/mod.rs
```diff
@@ -878,6 +879,95 @@ impl Context {
 
         self.vm.stack.get_function(self.vm.frame())
     }
+
+    /// Creates all globals required to evaluate `codeblock`.
+    ///
+    /// This is the common path of the instantiations:
+    /// - `EvalDeclarationInstantiation ( body, varEnv, lexEnv, privateEnv, strict )`
+    /// - `GlobalDeclarationInstantiation ( script, env )`
+    fn create_globals(&mut self, codeblock: &CodeBlock, is_eval_call: bool) -> JsResult<()> {
```
hansl: Can we rename the `is_eval_call` to something that makes more sense when reading the code without reading the spec? Like `is_configurable`?

core/engine/src/vm/code_block.rs
```diff
@@ -151,6 +161,10 @@ pub struct CodeBlock {
 
     /// Bytecode to source code mapping.
     pub(crate) source_info: SourceInfo,
+
+    pub(crate) global_lexs: Box<[u32]>,
+    pub(crate) global_fns: Box<[GlobalFunctionBinding]>,
+    pub(crate) global_vars: Box<[u32]>,
```
hansl, line 167: This is the part that I don't like of this PR; I'm trying to reduce the amount of heap and time it takes to create `CodeBlock`s and make `Call`/`New`. These take 16 bytes each for every scope, but they'll likely be empty for most scopes.

What do you think of making these `Option<Box<GlobalInitData>>` with the `GlobalInitData` type having those 3 fields?

jedel1043: Does that make a difference? IIRC we only use `Gc<CodeBlock>`, which is already allocated. So for (probably) all usecases, it would just copy a single pointer

jedel1043: I mention this because it feels a bit weird to have a `Gc` of a `Box` of a `Box<[]>`; a lot of indirection for something that is already heap allocated

hansl: This is constant after compiler phase, right? I might have thought this was on the call frame. I rescind my concern and will approve.

PR #4956, opened 2026-03-08
fix(executor): ignore recurring TimeoutJobs in is_empty

ashnaaseth2325-oss:
# Summary

This PR fixes an issue where `SimpleJobExecutor` could fail to terminate when `setInterval` is active.

Previously, `SimpleJobExecutor::is_empty()` only checked whether the `timeout_jobs` map itself was empty:

```rust
self.timeout_jobs.borrow().is_empty()
```

Because `setInterval` continuously schedules new `TimeoutJob`s, the queue never became empty. As a result, the executor always believed work was pending and `run_jobs()` could never exit.

However, `TimeoutJob` already exposes a `recurring` flag intended to indicate jobs that should **not block executor termination**.

# Fix

Two small adjustments connect the existing design to the executor logic.

First, the executor now ignores recurring timeout jobs when determining completion:

```rust
self.timeout_jobs
    .borrow()
    .values()
    .flatten()
    .all(TimeoutJob::is_recurring)
```

Second, the initial interval scheduling now uses `TimeoutJob::recurring` to ensure consistent behavior with later reschedules.

# Result

`run_jobs()` now exits correctly once all finite (non-recurring) work has completed, while `setInterval` continues to function normally.

### DIFF
--- a/core/engine/src/job.rs
+++ b/core/engine/src/job.rs
@@ -747,11 +747,21 @@ impl JobExecutor for SimpleJobExecutor {
                 return Err(err);
             }
 
-            let jobs = mem::take(&mut *self.promise_jobs.borrow_mut());
-            for job in jobs {
-                if let Err(err) = job.call(&mut context.borrow_mut()) {
-                    self.clear();
-                    return Err(err);
+            // Drain the promise-job queue completely (Microtask Checkpoint).
+            // A single `mem::take` snapshot is not enough: each job may
+            // enqueue further promise jobs via `.then()`, and those must be
+            // processed in the same checkpoint before the event loop can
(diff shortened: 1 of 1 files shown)


jedel1043:
This was the old behaviour, but it was patched up in #4891 to return to blocking if timeout jobs are still pending. It's much easier to maintain vs having to check all edge cases of termination.

ashnaaseth2325-oss:
Thanks for the clarification!
I wasn't aware that #4891 intentionally restored the blocking behaviour when timeout jobs are pending. That makes sense regarding the edge cases around termination.
Appreciate the explanation.

Issue #4410, opened 2025-09-07
`Finalize` not called


xbwwj:
**Describe the bug**

The `finalize` impl is not called when object is garbage collected.

**To Reproduce**

A minimal `Finalize` implementation

```rust
use boa_gc::{Finalize, Gc, Trace};
// use gc::{Finalize, Gc, Trace};

fn main() {
    _ = Gc::new(FinalizePrint);
}

#[derive(Trace)]
struct FinalizePrint;

impl Finalize for FinalizePrint {
    fn finalize(&self) {
        println!("this is finalized");
    }
}
```

Running the program outputs nothing.

**Expected behavior**

The `rust_gc` crate correctly output the finalize text.

```rust
// replace boa_gc with
use gc::{Finalize, Gc, Trace};
```

```sh
$ cargo run
this is finalized
```

**Build environment (please complete the following information):**

- OS: NixOS Linux
- Version: 25
- Target triple: x86_64-unknown-linux-gnu
- Rustc version: rustc 1.89.0 (29483883e 2025-08-04)

**Additional context**

jedel1043:
Marking as not a bug. We never guarantee that our garbage collector must always run all finalizers, we just guarantee that if a value gets freed by the garbage collector, it will run its finalizer before freeing it. I'm pretty sure the bigger memory thresholds for our collector are causing the differences between `rust_gc` and our code, since objects don't get collected if you don't reach the minimum required allocated memory.

xbwwj:
@jedel1043 I think this behavior may break thread local storage (TLS) usage in other crates. 

For example, the `wgpu` crate. The `wgpu` crate requires all resources to `drop` before its internal TLS [`SNATCH_LOCK_TRACE`](https://github.com/gfx-rs/wgpu/blob/6bab36d1fd946fe6456e4c86d98459122a7547ed/wgpu-core/src/snatch.rs#L113C16-L113C33), while `boa` defer these drops to `BOA_GC` TLS. 

As Rust has no control over TLS drop order, this can lead to TLS `AccessError`. This may be the root cause of https://github.com/boa-dev/boa/issues/4370.

jedel1043:
This will technically get fixed when we rewrite the GC, since that will tie every value to a context, making it possible to drop everything related to a context when the context gets dropped.

jedel1043:
For now, the simple solution for the wgpu problem would be to call `boa_gc::force_collect()` after the context gets dropped

PR #4959, opened 2026-03-08
fix(cli): resolve TLS AccessError by managing resource lifecycle on shutdown (#4370)

### LINKED ISSUE #4370: Thread Local Storage AccessError when used with wgpu
**Describe the bug**

The program panic with Thread Local Storage AccessError when use `wgpu` struct from `JsValue`.

**To Reproduce**

The WebGPU implementation (partly) is in this branch <https://github.com/xubaiwang/boa/tree/main/core/runtime/src/webgpu>.

Running this module with boa cli:

```js
// webgpu.js
const adapter = await navigator.gpu.requestAdapter();
const device = await adapter.requestDevice();

const input = new Float32Array([1, 3, 5]);
console.log("input =", input);
console.log("input instanceof Float32Array =", input instanceof Float32Array);

const workBuffer = device.createBuffer({
  size: input.byteLength,
  usage: GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC |
    GPUBufferUsage.COPY_DST,
});
console.log("workBuffer =", workBuffer);
console.log(
  "workBuffer instanceof GPUBuffer =",
  workBuffer instanceof GPUBuffer,
);

device.queue.writeBuffer(workBuffer, 0, input);
console.log("writeBuffer success");

console.log("end");
```

```bash
❯ cargo run -- -m webgpu-min.js
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.57s
     Running `target/debug/boa -m webgpu-min.js`
input = 1,3,5
input instanceof Float32Array = true
workBuffer = [object Object]
workBuffer instanceof GPUBuffer = true
writeBuffer success
end
The application panicked (crashed).
Message:  cannot access a Thread Local Storage value during or after destruction: AccessError
Location: /nix/store/wpdjjqamyjqhi3x54ivx0dkr9sm74w1b-rust-default-1.89.0/lib/rustlib/src/rust/library/std/src/thread/local.rs:281
fatal runtime error: thread local panicked on drop, aborting
Aborted                    (core dumped) cargo run -- -m webgpu-min.js
```

**Expected behavior**

The code should run without panic (similar code passes on rquickjs).

```rust
// the rquickjs version pass
#[derive(Trace, JsLifetime, Debug, Clone)]
#[rquickjs::class(rename = "GPUDevice", rename_all = "camelCase")]
pub struct Device<'js> {
    #[qjs(skip_trace)]
    pub inner: wgpu::Device,
    #[qjs(get)]
    pub queue: Value<'js>,
}
```

So I assume the usage may be different in in `boa_gc`. But there is too little documentation on how to play with `boa_gc`.

**Build environment**

- OS: NixOS Linux
- Version: 25.05
- Rustc version: rustc 1.89.0 (29483883e 2025-08-04)

**Additional context**

To make [`GPUDevice.queue`](https://developer.mozilla.org/en-US/docs/Web/API/GPUDevice/queue) always return the same object (it's property rather than getter, not different new JsValue each time), I wrap the `queue` into a `JsValue`.

<https://github.com/xubaiwang/boa/blob/5444f706ea9da0b559fb197bd4a3796417fcea86/core/runtime/src/webgpu/device.rs#L17>

```rust
#[derive(Debug, Clone, JsData, Trace, Finalize)]
#[boa_gc(unsafe_no_drop)]
pub struct Device {
    /// Inner wgpu handle.
    #[unsafe_ignore_trace]
    pub inner: wgpu::Device,
    /// Returns the primary `GPUQueue` for the device.
    pub queue: JsValue,  // <- Value wrapped rather than `Queue` or `wgpu::Queue`
}

#[derive(Debug, JsData, Trace, Finalize)]
#[boa_gc(unsafe_no_drop)]
pub struct Queue {
    /// Inner wgpu handle.
    #[unsafe_ignore_trace]
    pub inner: wgpu::Queue,
}
```

0xShubhamSolanki:
This PR fixes the fatal `AccessError` panic that happens during the CLI's teardown phase when handling TLS-dependent resources.

As noted by @jedel1043 in #4410, the crash occurs because hardware-linked resources (such as `wgpu` structs in custom embeddings) are dropped too late during the main thread's teardown. By the time the garbage collector attempts to finalize them, the Thread Local Storage (TLS) has already been destroyed.

To fix this, we just need to ensure that garbage collection runs deterministically before the thread teardown begins. 

Changes made in `cli/src/main.rs`:
-  Bound the `Context` to a local `owned_context` variable to allow manual drop control.
-  Added an explicit `drop(owned_context)` immediately followed by `boa_gc::force_collect()` across all CLI exit paths (file execution, `-e` evaluation, stdin, and Repl termination).

By clearing the GC roots and forcing a synchronous collection block, all pending finalizers are safely executed while the thread's TLS is still fully active.

*(Note: i haven't included a dedicated webGpu test file since `wgpu` isn't part of the core engine, but this lifecycle adjustment permanently resolves the teardown issue for any external embeddings relying on TLS).*

Fixes #4370

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -541,12 +541,14 @@ fn main() -> Result<()> {
 
     let executor = Rc::new(Executor::new(printer.clone()));
     let loader = Rc::new(SimpleModuleLoader::new(&args.root).map_err(|e| eyre!(e.to_string()))?);
-    let context = &mut ContextBuilder::new()
+    // Bind the Context to an owned variable so it can be explicitly dropped
+    // before process termination, preventing TLS AccessErrors during garbage collection.
+    let mut owned_context = ContextBuilder::new()
         .job_executor(executor.clone())
         .module_loader(loader.clone())
         .build()
         .map_err(|e| eyre!(e.to_string()))?;
-
+    let context = &mut owned_context;
     // Strict mode
     context.strict(args.strict);
 
@@ -572,21 +574,37 @@ fn main() -> Result<()> {
         if let Some(ref expr) = args.expression {
             evaluate_expr(expr, &args, context, &printer)?;
         }
-
+        // Ensure context is dropped and collected to finalize resources (e.g., wgpu) safely.
+        drop(loader);
+        drop(executor);
+        drop(owned_context);
+        boa_gc::force_collect();
         return Ok(());
     } else if let Some(ref expr) = args.expression {
         evaluate_expr(expr, &args, context, &printer)?;
+        // Ensure context is dropped and collected to finalize resources (e.g., wgpu) safely.
+        drop(loader);
+        drop(executor);
+        drop(owned_context);
+        boa_gc::force_collect();
         return Ok(());
     } else if !io::stdin().is_terminal() {
         let mut input = String::new();
         io::stdin()
             .read_to_string(&mut input)
             .wrap_err("failed to read stdin")?;
-        return if input.is_empty() {
+
+        let result = if input.is_empty() {
             Ok(())
         } else {
             evaluate_expr(&input, &args, context, &printer)
         };
+        // Ensure context is dropped and collected to finalize resources (e.g., wgpu) safely.
+        drop(loader);
+        drop(executor);
+        drop(owned_context);
+        boa_gc::force_collect();
+        return result;
     }
 
     // Print the welcome banner unless --quiet is passed.
@@ -671,6 +689,10 @@ fn main() -> Result<()> {
         .map_err(|e| e.into_erased(context));
 
     handle.join().expect("failed to join thread");
+    // Explicitly drop the root context and force garbage collection.
+    // This allows hardware-linked finalizers to run while Thread Local Storage is still active.
+    drop(owned_context);
+    boa_gc::force_collect();
 
     Ok(result?)
 }


jedel1043:
Uhhhh I don't think this fixes the issue. I mean, it fixes it for _us_ (`boa_cli`) but the person who reported the issue was probably using `boa_engine`, and doing this doesn't fix the issue for _them_ and all users of `boa_engine`

0xShubhamSolanki:
Ah, that makes complete sense. I iwas too focused on fixing the CLI teardown panic and missed the broader impact on `boa_engine` consumers. You're entirely right that anyone embedding the engine would still hit this.

Since you mentioned in #4410 that `boa_gc::force_collect()` should ideally be called "AFTER" the context is dropped, doing this automatically inside the engine seems a bit tricky before the full Gc rewrite. 

How would you prefer to handle this at the boa_engine level? 
1. Should we implement `drop` for `context` and trigger `boa_gc::force_collect()` there (if it's safe to do so while the context is tearing down)?
2. Or would you prefer exposing a dedicated teardown / shutdown method on the context that consumers are encouraged to call?

happy to pivot the pr to implement the fix at the engine level based on your preference!

jedel1043:
Actually, based off of https://github.com/boa-dev/boa/issues/4370#issuecomment-3273241384, it might not even be a fix to call `force_collect`, or at least it didn't work for them. So this might require more investigation, but if my gut feeling is right, this might not be fixed until we migrate to the new GC.

jedel1043:
Closing the PR for now, need to investigate the proper fix.

PR #4960, opened 2026-03-09
fix : restore Unicode property escapes support (fixes ~153 Test262 failures)

mrhapile:
## Summary

This PR updates the `regress` dependency to a newer revision containing updated Unicode property tables.

Boa delegates RegExp parsing and matching to the `regress` crate. The previously used version lacked newer Unicode property data, which caused failures in Unicode property escape tests.

Updating the dependency restores correct behavior for `\p{...}` and `\P{...}` escapes as defined in ECMA-262.

## Specification

ECMA-262 RegExp Unicode Property Escapes  
https://tc39.es/ecma262/#sec-runtime-semantics-unicodematchproperty

## Root Cause

The previous dependency:
regress = 0.10.5

did not contain the latest Unicode property classifications required by the Test262 suite.

Since Boa forwards regex patterns directly to `regress`, missing Unicode tables in the dependency caused property escape tests to fail.

## Fix

Update the dependency to:

regress (git revision b4dc529)

This revision includes updated Unicode property tables, allowing the regex backend to correctly interpret property escapes.

No changes were required in Boa's RegExp implementation.

### Before

Screenshot 2026-03-09 at 5 19 00 AM

- Some Unicode property escapes returned incorrect results
- ~154 Test262 failures in `RegExp/property-escapes`

### After
Screenshot 2026-03-09 at 5 22 26 AM
Screenshot 2026-03-09 at 5 34 42 AM

- Unicode property escapes resolve correctly
- Property escape Test262 tests pass

## Notes

- No changes to Boa's RegExp implementation
- No GC interaction changes
- Minimal dependency update

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,6 +605,7 @@ dependencies = [
 name = "boa_runtime"
 version = "1.0.0-dev"
 dependencies = [
+ "base64",
  "boa_engine",
  "boa_gc",
  "bytemuck",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ small_btree = { version = "~0.1.0", path = "utils/small_btree" }
 
 # Shared deps
 arbitrary = "1"
+base64 = "0.22.1"
 bitflags = "2.11.0"
 clap = "4.5.59"
 colored = "3.1.1"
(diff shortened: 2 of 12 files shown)


jedel1043:
Let's not use git dependencies for dependencies that we not control. We make an exception for `temporal_rs` because it's our project, and we can just push a new release for it if we wanted to release a new version of Boa, but for the rest of the dependencies it's better to stay in released versions to avoid blocking our own release schedule.

PR #4961, opened 2026-03-09
perf: hoist loop-invariant constants out of loop conditions

hansl:
When a loop condition like `i < 10` has a literal or const operand, compile the constant into a register once before the loop rather than re-emitting PushInt8/PushInt32 on every iteration. This eliminates one opcode dispatch + register write per iteration for the common for-loop pattern.

Applies to for, while, and do-while loops. Only hoists values that are provably immutable: literal expressions and const bindings that aren't already optimized by compile_expr_operand (local or cached const bindings are already zero-cost inside the loop).


jedel1043 (approved):
Looks good!

PR #4964, opened 2026-03-09
cli: return non-zero for uncaught runtime exceptions

### LINKED ISSUE #4962: boa_cli exit code 0 on uncaught runtime exceptions
**Describe the bug**
  If boa reports an uncaught runtime exception in a script, it exits with a zero status code.

  **To Reproduce**

  ```sh
  $ echo '1/x;' > /tmp/ref.js && ./boa /tmp/ref.js; echo $?
  Uncaught: ReferenceError: x is not defined (native at core/engine/src/vm/opcode/get/name.rs:91:28)
      at <main> (/tmp/ref.js:?:?)
  0

  $ echo 'throw new Error("boom");' > /tmp/throw.js && ./boa /tmp/throw.js; echo $?
  Uncaught: Error: boom (/tmp/throw.js:1:7)
      at <main> (/tmp/throw.js:1:7)
  0

  Syntax error for comparison return non-zero code already:
  $ echo 'function bad( {}' > /tmp/syntax.js && ./boa /tmp/syntax.js; echo $?
  Error:
     0: SyntaxError: abrupt end
  1
```

**Expected behavior**
It should signal an abnormal script termination with a non-zero status code.

**Build environment**

  - OS: Linux
  - Version: 0919e3b2c00d75fd32968903a328e7b85989d5c5
  - Target triple: aarch64-unknown-linux-gnu

### PR DESCRIPTION
**Pull request by SergioChan**
This Pull Request fixes/closes #4962.

It changes the following:

- Return an error from `evaluate_file` when script execution raises an uncaught runtime exception.
- Preserve existing uncaught error printing, but propagate the failure so the CLI process exits non-zero.
- Keep successful file execution behavior unchanged.

## Validation

- Not run locally in this environment because the Rust toolchain (`cargo`) is unavailable.
- Behavioral check is directly aligned with issue reproduction: uncaught runtime errors now flow through an error return path instead of being swallowed with `Ok(())`.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -12,6 +12,9 @@ mod executor;
 mod helper;
 mod logger;
 
+#[cfg(test)]
+use assert_cmd as _;
+
 use crate::executor::Executor;
 use crate::logger::SharedExternalPrinterLogger;
 use async_channel::Sender;
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -512,7 +512,7 @@ fn evaluate_file(
         }
         Err(v) => {
             printer.print(uncaught_error(&v));
-            return Err(v.into_erased(context).into());
+            return Err(eyre!("execution failed"));
         }
     }
 
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -411,15 +411,22 @@ fn evaluate_expr(
                     let result = script.evaluate(context);
                     if let Err(err) = context.run_jobs() {
                         printer.print(uncaught_job_error(&err));
+                        return Err(err.into_erased(context).into());
                     }
                     result
                 };
                 match result {
                     Ok(v) => printer.print(format!("{}\n", v.display())),
-                    Err(ref v) => printer.print(uncaught_error(v)),
+                    Err(v) => {
+                        printer.print(uncaught_error(&v));
+                        return Err(v.into_erased(context).into());
+                    }
                 }
             }
-            Err(ref v) => printer.print(uncaught_error(v)),
+            Err(v) => {
+                printer.print(uncaught_error(&v));
+                return Err(v.into_erased(context).into());
+            }
         }
     }
 
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -503,7 +503,10 @@ fn evaluate_file(
                 println!("{}", v.display());
             }
         }
-        Err(v) => printer.print(uncaught_error(&v)),
+        Err(v) => {
+            printer.print(uncaught_error(&v));
+            return Err(v.into_erased(context).into());
+        }
     }
 
     Ok(())
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -12,9 +12,6 @@ mod executor;
 mod helper;
 mod logger;
 
-#[cfg(test)]
-use assert_cmd as _;
-
 use crate::executor::Executor;
 use crate::logger::SharedExternalPrinterLogger;
 use async_channel::Sender;
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -411,7 +411,7 @@ fn evaluate_expr(
                     let result = script.evaluate(context);
                     if let Err(err) = context.run_jobs() {
                         printer.print(uncaught_job_error(&err));
-                        return Err(err.into_erased(context).into());
+                        return Err(eyre!("execution failed"));
                     }
                     result
                 };
@@ -419,13 +419,13 @@ fn evaluate_expr(
                     Ok(v) => printer.print(format!("{}\n", v.display())),
                     Err(v) => {
                         printer.print(uncaught_error(&v));
-                        return Err(v.into_erased(context).into());
+                        return Err(eyre!("execution failed"));
                     }
                 }
             }
             Err(v) => {
                 printer.print(uncaught_error(&v));
-                return Err(v.into_erased(context).into());
+                return Err(eyre!("parsing failed"));
             }
         }
     }
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -30,6 +30,9 @@ futures-lite.workspace = true
 async-channel.workspace = true
 rustls.workspace = true
 
+[dev-dependencies]
+assert_cmd = "2.0.16"
+
 [features]
 default = [
     "boa_engine/annex-b",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -30,9 +30,6 @@ futures-lite.workspace = true
 async-channel.workspace = true
 rustls.workspace = true
 
-[dev-dependencies]
-assert_cmd = "2.0.16"
-
 [features]
 default = [
     "boa_engine/annex-b",
--- a/cli/tests/exit_code.rs
+++ b/cli/tests/exit_code.rs
@@ -1,0 +1,19 @@
+use assert_cmd::Command;
+
+#[test]
+fn stdin_uncaught_error_exits_non_zero() {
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by jedel1043** [comment]
I don't think this is fully working
```rust
➜  boa git:(pr-4964) echo 'throw Error("nooo")' | cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/boa`
Uncaught Error: Error: nooo (unknown at :1:12)
    at <main> (unknown at :1:12)
➜  boa git:(pr-4964) echo $?
0
➜  boa git:(pr-4964)
```

SergioChan:
Thanks for the repro — you were right.

The previous patch only returned a non-zero status for file execution errors. `stdin` / `-e` still went through `evaluate_expr`, which printed the uncaught error but returned `Ok(())`.

I pushed a follow-up commit that propagates those errors out of `evaluate_expr` as well:
- commit: `81d20ff`
- PR branch: `fix/issue-4962-exitcode`

What changed:
- return `Err(...)` on script parse/eval failures in `evaluate_expr`
- return `Err(...)` when `context.run_jobs()` fails in `evaluate_expr`

Validation note:
- I could not run the local CLI repro in this environment because the Rust toolchain (`cargo`) is unavailable here.
- Please rely on CI for execution validation; if you want, I can add a targeted regression test once I have a toolchain-enabled environment.

cli/src/main.rs
jedel1043, line 398: Probably also needs to be propagated

jedel1043:
Yeah a test would be ideal, using something like [`assert_cmd`](https://docs.rs/assert_cmd/latest/assert_cmd/)

jedel1043 (changes requested):

SergioChan:
Great call — I added integration regression coverage using `assert_cmd`.

### What I changed
- Added `assert_cmd` as a dev dependency for `boa_cli`.
- Added `cli/tests/exit_code.rs` with two regression tests:
  - `stdin_uncaught_error_exits_non_zero`
  - `expression_uncaught_error_exits_non_zero`
- Both assert that uncaught runtime exceptions produce a non-zero process exit code for stdin and `-e` execution paths.

### Validation
- I couldn't execute `cargo test` in this environment because the Rust toolchain (`cargo`) is unavailable here.
- The tests are committed and CI should validate them on the PR branch.

SergioChan:
Pushed follow-up commit `8dc5d133` to address CI/lint blockers from the regression test addition.

### What changed
- Added crate-level allowances in `cli/tests/exit_code.rs` for strict test-target lint settings:
  - `missing_docs`
  - `unused_crate_dependencies`
- Added a test-only import in `cli/src/main.rs`:
  - `#[cfg(test)] use assert_cmd as _;`
  This satisfies `unused-crate-dependencies` for the binary test target.

This keeps the `assert_cmd` regression tests while matching the repo's `-D warnings` policy. Please take another look.

jedel1043:
You also need to commit your changes to the `Cargo.lock`, since you added a new dependency.

SergioChan:
Thanks for the review. I鈥檓 working through the requested updates and will post a focused patch shortly.

cli/src/main.rs
```diff
@@ -413,15 +416,22 @@ fn evaluate_expr(
...
+                        return Err(err.into_erased(context).into());
                     }
                     result
                 };
                 match result {
                     Ok(v) => printer.print(format!("{}\n", v.display())),
-                    Err(ref v) => printer.print(uncaught_error(v)),
+                    Err(v) => {
+                        printer.print(uncaught_error(&v));
+                        return Err(v.into_erased(context).into());
```
jedel1043: Tested the code and directly returning the error makes it show up two times in stderr. We might have to adjust the returned error such that it doesn't reference the already printed error.

SergioChan:
Thanks for the pointer on `Cargo.lock`.

I pushed commit `b2116dfa` to avoid introducing a new dependency in `cli` tests:

- removed `assert_cmd` from `cli/Cargo.toml`
- removed the test-only `assert_cmd` import from `cli/src/main.rs`
- rewrote `cli/tests/exit_code.rs` to use `std::process::Command` + `CARGO_BIN_EXE_boa`

This keeps the same regression coverage (stdin and `-e` uncaught errors return non-zero) while avoiding lockfile churn.

Validation run in this update:
- `git diff --check` (clean)

I couldn’t run `cargo test` in this environment because the Rust toolchain is not available, so CI is needed for execution validation.

jedel1043:
You didn't address my comment though

SergioChan:
You’re absolutely right — I missed that thread. Thanks for the nudge.

I’ll address your remaining comment directly and push a follow-up patch shortly.

SergioChan:
Thanks for the follow-up — I tracked down the missed review point about duplicated stderr output.

I just pushed `62e25c9`.

### What changed
- In `evaluate_expr`, keep printing the uncaught parse/runtime/job error once via the existing printer.
- Return a concise sentinel error (`parsing failed` / `execution failed`) instead of re-wrapping the original JS error object.

This preserves the non-zero exit behavior while avoiding duplicated detailed error text in stderr.

### Validation
- `git diff --check` (clean)
- I could not run `cargo test` here because the Rust toolchain is unavailable in this environment, so CI is needed for execution validation.

SergioChan:
Quick follow-up on the outstanding changes-requested thread: commit 62e25c9 keeps the uncaught-error printing path but returns sentinel errors to avoid duplicated stderr output, while preserving non-zero exits.\n\n@jedel1043 could you re-check this when you have time and confirm whether this resolves your remaining concern?

SergioChan:
Quick follow-up on the CHANGES_REQUESTED items: I switched the -e/stdin paths to return a short eyre!(...) after printing the uncaught error so we avoid bubbling the full engine error twice on stderr, while keeping evaluate_file propagation as-is for file execution semantics.\n\nCould you re-check if this matches what you had in mind, or should I also switch the file path to a shortened propagated error?

SergioChan:
Pushed follow-up commit `17c35d59` to address the duplicate-stderr behavior in file execution and align all execution modes.

### What changed
- In `evaluate_file`, after printing the uncaught error, return a short sentinel error (`execution failed`) instead of bubbling the full JS error object again.
- Added a regression test `file_uncaught_error_exits_non_zero` in `cli/tests/exit_code.rs` to cover script-file execution explicitly.

### Why
This keeps the non-zero exit behavior from #4962 while avoiding duplicate detailed error output on stderr in the file path (matching the stdin / `-e` handling).

### Validation
- `git diff --check` (clean)
- Could not run `cargo test` here because `cargo` is unavailable in this environment, so CI is needed for execution validation.

SergioChan:
Quick update: I rebased this branch onto the latest `main` and force-pushed to refresh the PR tip (`fa38eac5`).

### What changed
- Replayed the existing fix commits on top of current `main`.
- No intentional functional changes beyond the rebase; this is to keep the branch current for review/merge.

### Validation
- `git diff --check upstream/main...HEAD` ✅
- I couldn’t run `cargo test` here because `cargo` is not available in this runtime; CI should provide execution validation.

SergioChan:
Rebased this branch onto the latest `main` again and force-pushed the refreshed head (`be429baf`).

No intentional functional changes were made in this update; this was a branch-refresh push to keep the PR current for review.

Validation in this run:
- `git diff --check upstream/main...HEAD` ✅
- `cargo test -p boa_cli --test exit_code -- --nocapture` *(not runnable here: `cargo` is unavailable in this environment)*

jedel1043 (approved):
LGTM!

zhuzhu81998:
Lint and formatting still need to be adjusted

and what is going on with the iterator test262 tests XD

PR #4973, opened 2026-03-09
fix(parser): Error::expected for const init, clearer for-of error, op…

tkshsbcue:
## Summary

Three parser improvements aligning with maintainer feedback: (1) Use `Error::expected` instead of `Error::general` for const declaration missing initializer. (2) Replace vague "unexpected token" with a descriptive message for invalid `let` in for-of. (3) Replace operator lexer `unimplemented!` with proper syntax error.

## Motivation

**Error::expected for const** — "Expected initializer for const declaration" is an "expected X, got Y" pattern. Using `Error::expected(["="], found, span, context)` matches the rest of the parser and gives a clearer message.

**For-of `let` message** — The previous "unexpected token" was vague. Replacing with "'let' cannot be used as the iterable in a for-of loop" explains the issue (mirrors the nearby `async` error).

**Operator lexer** — `unimplemented!("operator {}", op)` panics on unknown operator bytes. Replacing with `Error::syntax` returns a recoverable parse error instead.

## Changes

| File | Change |
|------|--------|
| `lexical.rs` | `Error::general("Expected initializer...")` → `Error::expected(["="], next.to_string(interner), next.span(), "const declaration")` |
| `for_statement.rs` | `Error::general("unexpected token")` → `Error::general("'let' cannot be used as the iterable in a for-of loop", position)` |
| `operator.rs` | `unimplemented!("operator {}", op)` → `Err(Error::syntax(format!("unexpected operator '{}'", char::from(op)), start_pos.position()))` |

## Testing

- [x] `cargo test -p boa_parser` — 296 tests pass
- [x] `cargo clippy -p boa_parser` — no warnings

### DIFF
--- a/core/parser/src/parser/expression/assignment/conditional.rs
+++ b/core/parser/src/parser/expression/assignment/conditional.rs
@@ -68,7 +68,7 @@ where
         if let Some(tok) = cursor.peek(0, interner)?
             && tok.kind() == &TokenKind::Punctuator(Punctuator::Question)
         {
-            let lhs = lhs.expect_expression();
+            let lhs = lhs.try_into_expression()?;
 
             cursor.advance(interner);
             let then_clause = AssignmentExpression::new(true, self.allow_yield, self.allow_await)
--- a/core/parser/src/parser/expression/fpl_or_exp.rs
+++ b/core/parser/src/parser/expression/fpl_or_exp.rs
@@ -10,15 +10,6 @@ pub(crate) enum FormalParameterListOrExpression {
 }
 
 impl FormalParameterListOrExpression {
-    pub(crate) fn expect_expression(self) -> ast::Expression {
-        match self {
-            FormalParameterListOrExpression::Expression(expr) => expr,
-            FormalParameterListOrExpression::FormalParameterList { .. } => {
-                panic!("Unexpected arrow-function arguments");
-            }
-        }
-    }
-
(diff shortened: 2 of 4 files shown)


tkshsbcue:
Hi @jedel1043 , I came across a few interesting cases while compiling. I think improving the error handling in this part could make the user experience better.

Thanks!

core/parser/src/lexer/operator.rs
```diff
@@ -162,7 +162,12 @@ impl<R> Tokenizer<R> for Operator {
             b'~' => {
                 Token::new_by_position_group(Punctuator::Neg.into(), start_pos, cursor.pos_group())
             }
-            op => unimplemented!("operator {}", op),
+            op => {
+                return Err(Error::syntax(
+                    format!("unexpected operator '{}'", char::from(op)),
+                    start_pos.position(),
+                ));
```
jedel1043: We have an `Error::unexpected` variant for cases like this.

tkshsbcue: @jedel1043 Switched to Error::unexpected per your suggestion.

tkshsbcue:
@jedel1043 i hope you can review my PR

tkshsbcue:
hey @jedel1043 if there are any fixes i need to do lemme know

jedel1043 (approved):
Thanks! This should also give much better error messages so that's good

PR #4976, opened 2026-03-09
feat(engine): add `JsArrayBuffer::to_vec` for copying bytes

### LINKED ISSUE #4974: Rust API: add `JsArrayBuffer::to_vec()` for copying bytes
## Problem

- On `main`, there is no simple copying helper on `JsArrayBuffer` to retrieve its contents as `Vec<u8>`.
- Embedders can use `JsArrayBuffer::data()` and copy from `GcRef<[u8]>`, but that adds boilerplate for a common “snapshot bytes” use case.
- We now have `JsUint8Array::to_vec(...)` and `JsSharedArrayBuffer::to_vec()`, but `JsArrayBuffer` itself is missing a matching ergonomic helper.

## Proposal

- Add a helper on `JsArrayBuffer` that copies the current buffer contents into a fresh `Vec<u8>`.
- Because `ArrayBuffer` can be detached, mirror `data()` semantics by returning `None` when detached (instead of throwing).

## API

```rust
/// Copies the contents of this `ArrayBuffer` into a new `Vec<u8>`.
///
/// Returns `None` if the buffer has been detached.
#[must_use]
pub fn to_vec(&self) -> Option<Vec<u8>>
```

Location: `impl JsArrayBuffer` in `core/engine/src/object/builtins/jsarraybuffer.rs`.

## Rationale

- Consistent with existing helpers:
  - `JsUint8Array::to_vec(...)`
  - `JsSharedArrayBuffer::to_vec()`
- Improves Rust embedding ergonomics for a common “snapshot bytes” use case.
- No raw pointers or long-lived references; mirrors existing detach semantics.

## Implementation

- Call `self.data()` and, if present, clone into a `Vec<u8>`; return `None` if detached.
- Add docs + an example (including detach behaviour).
- Add a unit test: round-trip bytes + detach → `None`.

## Checklist

- [x] Add `to_vec(&self) -> Option<Vec<u8>>` with docs.
- [x] Unit test: create buffer, call `to_vec()`, assert contents; detach and assert `None`.
- [x] Add a short doc example (if docs cover this area).

---

alienx5499:
This Pull Request fixes/closes #4974.

It changes the following:

- Add `JsArrayBuffer::to_vec(&self) -> Option<Vec<u8>>` helper to copy buffer contents into a fresh `Vec<u8>` while returning `None` when the buffer has been detached
- Add a doc comment and example for `JsArrayBuffer::to_vec`, including both normal and detached cases
- Add small unit tests (`array_buffer_to_vec_roundtrip_and_detach` and `array_buffer_to_vec_empty`) to cover round-trip, detach, and empty-buffer behavior

Testing:

- `cargo test -p boa_engine` (including the new unit tests and doctest)

### DIFF
--- a/core/engine/src/object/builtins/jsarraybuffer.rs
+++ b/core/engine/src/object/builtins/jsarraybuffer.rs
@@ -257,6 +257,35 @@ impl JsArrayBuffer {
         GcRef::try_map(self.inner.borrow(), |o| o.data().bytes())
     }
 
+    /// Copies the contents of this [`JsArrayBuffer`] into a new [`Vec<u8>`].
+    ///
+    /// Returns `None` if the buffer has been detached.
+    ///
+    /// See also [`crate::object::builtins::JsUint8Array::to_vec`] and
+    /// [`crate::object::builtins::JsSharedArrayBuffer::to_vec`].
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use boa_engine::object::builtins::{AlignedVec, JsArrayBuffer};
+    /// # use boa_engine::{Context, JsResult, JsValue};
+    /// # fn main() -> JsResult<()> {
+    /// let context = &mut Context::default();
+    /// let data = AlignedVec::from_iter(0, [1u8, 2, 3, 4]);
+    /// let buffer = JsArrayBuffer::from_byte_block(data, context)?;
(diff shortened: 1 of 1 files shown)


alienx5499:
Hi @jedel1043
This PR is ready for review. Would appreciate your feedback when you have time.
Thanks!

PR #4980, opened 2026-03-10
feat: implement [[IsHTMLDDA]] internal slot (Annex B §B.3.6)

xcb3d:
This Pull Request implements the `[[IsHTMLDDA]]` internal slot (ECMAScript Annex B §B.3.6), enabling correct handling of `document.all`-like objects.

It changes the following:

- Add `IsHTMLDDA` marker struct with a custom `[[Call]]` internal method that returns `undefined` (`builtins/is_html_dda.rs`)
- Modify `typeof` operator to return `"undefined"` for objects with `[[IsHTMLDDA]]` (`value/variant.rs`)
- Modify `ToBoolean` to return `false` for objects with `[[IsHTMLDDA]]` (`value/inner/nan_boxed.rs`, `value/inner/legacy.rs`)
- Modify abstract equality (`==`) so that `[[IsHTMLDDA]]` objects are loosely equal to `null` and `undefined` (`value/equality.rs`)
- Fix `String.prototype.{match,matchAll,replace,replaceAll,search,split}` to treat `[[IsHTMLDDA]]` objects as "undefined or null" in the spec's "neither undefined nor null" check (`builtins/string/mod.rs`)
- Add `$262.IsHTMLDDA` to the Test262 harness (`tests/tester/src/exec/js262.rs`)
- Remove `IsHTMLDDA` from ignored features in `test262_config.toml`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3613,9 +3613,9 @@ checksum = "dc897dd8d9e8bd1ed8cdad82b5966c3e0ecae09fb1907d58efaa013543185d0a"
 
 [[package]]
 name = "regress"
-version = "0.10.5"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2057b2325e68a893284d1538021ab90279adac1139957ca2a74426c6f118fb48"
+checksum = "07948de9abc2e83adbeb7543c061a5ddaf7d944afcafbdd6e6b39aeacd40504b"
 dependencies = [
  "hashbrown 0.16.1",
  "memchr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ oneshot = "0.2.1"
 phf = { version = "0.13.1", default-features = false }
 pollster = "0.4.0"
 regex = "1.12.3"
-regress = { version = "0.10.5", features = ["utf16"] }
+regress = { version = "0.11.0", features = ["utf16"] }
 reqwest = { version = "0.13.2" }
 rustc-hash = { version = "2.1.1", default-features = false }
 serde_json = "1.0.149"
--- a/core/ast/src/scope.rs
+++ b/core/ast/src/scope.rs
@@ -261,6 +261,28 @@ impl Scope {
         self.inner.outer.is_none()
     }
 
+    /// Check if a binding with the given name is mutable.
+    ///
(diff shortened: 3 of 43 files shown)


core/engine/src/builtins/string/mod.rs
```diff
@@ -1003,8 +1003,12 @@ impl String {
         let search_value = args.get_or_undefined(0);
         let replace_value = args.get_or_undefined(1);
 
-        // 2. If searchValue is an Object, then
-        if search_value.is_object() {
+        // 2. If searchValue is neither undefined nor null, then
+        if search_value.is_object()
+            && !search_value
+                .as_object()
+                .is_some_and(|o| o.is::<IsHTMLDDA>())
```
jedel1043: I don't think changing this is correct. According to the [specification](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-identity):

> From the perspective of the ECMAScript language, language values are compared for equality using the [SameValue](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-samevalue) abstract operation and the [abstract operations](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-algorithm-conventions-abstract-operations) it transitively calls.

So even if `[[IsHTMLDDA]]` objects act like `undefined`, per the spec semantics they should evaluate to not equal to `undefined`.

jedel1043 (changes requested):
We also need to gate every related to `IsHTMLDDA` behind the `annex-b` feature, because we don't wanna bundle this in builds that don't care about web compatibility.

core/engine/src/value/inner/legacy.rs
```diff
@@ -270,7 +272,16 @@ impl EnumBasedValue {
...
-            Self::Object(_) | Self::Symbol(_) => true,
+            Self::Symbol(_) => true,
+            // Objects are truthy, unless they have [[IsHTMLDDA]] (Annex B §B.3.6.1).
+            Self::Object(obj) => {
+                #[cfg(feature = "annex-b")]
+                if obj.is::<IsHTMLDDA>() {
+                    return false;
+                }
+                let _ = obj;
+                true
```
jedel1043: You should use a match guard here to avoid the weird `let _` assignment

```rust
#[cfg(feature = "annex-b")]
Self::Object(obj) if obj.is::<IsHTMLDDA>() =>
```
Also applies to the nan boxed version

jedel1043 (approved):
Looks good!

Issue #4937, opened 2026-03-07
Developer Tooling & Engine Observability


kingsmen732:
## From understanding what is the current issue:
- print statements and ad-hoc eprintln! calls are the only debugging mechanism today.No structured VM logging
- there is no way to capture a chronological record of bytecode execution, GC events, or scope resolution without    modifying source and recompiling.No opt-in tracing
- allocation counts, GC pause durations, bytecode histogram data, and similar statistics are not collected or exposed
- the existing docs do not explain how to use debug builds, enable verbose output, or interpret engine internals

## Rough idea on what to be done 
- Logging — structured, levelled messages at key execution points
- Tracing — a step-by-step record of what the engine did and when
- Metrics — hard numbers: allocations, GC pauses, bytecode frequency

## What we could be actually incorporating to make this work 
- Replace ad-hoc debug output with structured, levelled log calls
- Add a `DebugVM` wrapper around the main interpreter loop that, when enabled via a builder option or environment variable (`BOA_TRACE=1`), emits per-instruction trace events
- Maybe introduce a lightweight `BoaMetrics` struct (hidden behind the `metrics` Cargo feature) that accumulates counters and histograms during execution
-Mainly start making debugging.md and rustdocs examples to every public metrics type and how to attach the logs 

this currently aligns close to #2736 , we can try to start creating a first level implementation so later it will be helpful to start with what parameters needs to be traced and logged.

rajat552:
@kingsmen732 

This looks like a very interesting direction for improving observability in the engine.

From what I understand, introducing tracing could help capture key execution events such as bytecode execution, GC activity, and scope resolution without relying on ad-hoc debug prints. It might also make debugging and performance analysis significantly easier.

I’m interested in exploring this further. A possible first step could be introducing a lightweight tracing layer around the main interpreter loop and emitting events for opcode execution, then gradually expanding it to GC and other subsystems.

Would it make sense to start with a minimal tracing implementation (possibly using the `tracing` crate) and iterate from there?

shreyadesai20042009:
I’m interested in contributing to this. I would like to start by improving
the debugging documentation and developer tooling guides.

Saksham1o:
Hi! This looks like a very useful improvement for the developer experience around Boa.

I'd be interested in working on a first minimal step here, possibly introducing structured logging for key VM events (e.g., opcode execution) using something like the `tracing` crate, and then expanding from there.

Would that be a good starting direction for this issue?

[...]

PR #4982, opened 2026-03-10
cli: support BOA_TRACE env flag for VM trace mode

Flamki:
## Summary
This is a minimal, first-step observability improvement for #4937.

It adds an opt-in environment toggle for existing VM trace output so tracing can be enabled in non-interactive workflows (scripts/CI/wrappers) without changing source code.

## Changes
- CLI: add `BOA_TRACE` support in addition to `--trace`.
- Accept strict boolean-like values for `BOA_TRACE`:
  - truthy: `1`, `true`, `yes`, `on`
  - falsy: `0`, `false`, `no`, `off`
- Error clearly on invalid/non-UTF8 `BOA_TRACE` values.
- Keep precedence deterministic: `--trace` overrides `BOA_TRACE`.
- Add focused unit tests for env var parsing.
- Update `docs/debugging.md` with usage examples and value rules.

## Scope
- Intentionally narrow: CLI + docs only.
- No VM core tracing redesign in this PR.
- No metrics subsystem introduced yet.

## Validation
- `cargo fmt --all -- --check`
- `cargo test -p boa_cli tests::parse_bool_env_var -- --nocapture`
- `cargo clippy -p boa_cli --all-targets --no-deps -- -D warnings`

Partially addresses #4937.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -112,7 +112,11 @@ struct Opt {
     #[allow(clippy::option_option)]
     dump_ast: Option<Option<DumpFormat>>,
 
-    /// Dump the AST to stdout with the given format.
+    /// Enable VM instruction tracing.
+    ///
+    /// This can also be enabled with the `BOA_TRACE` environment variable using
+    /// one of the following values: `1`, `true`, `yes`, `on`, `0`, `false`,
+    /// `no`, `off`.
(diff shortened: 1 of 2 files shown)


jedel1043:
Kind of superseeded by #4931, and arguably much better since this assumes you have an OS to get env variables, but the other PR is customizable per the host's requirements.

PR #4984, opened 2026-03-10
Fix unicode regress update

### LINKED ISSUE #4791: Title:Emoji 16.0 sequences fail because regress uses Unicode 15.1 tables
**Describe the bug**
RegExp fails to match new Emoji 16.0 sequences (e.g., the Cook Islands flag 🇨🇶) when the v flag is used. Even with the latest engine updates, these sequences return None during execution

**To Reproduce**
Running the official Test262 suite for Emoji 16.0 or the following code in Boa:
```
const re = /^\p{RGI_Emoji}$/v;
re.test("🇨🇶"); // Returns false, but should be true
```

**Expected behavior**
The regex should correctly match all RGI Emoji sequences defined in the Unicode 16.0 standard. Currently, it only matches emojis up to version 15.1.

**Additional context**

I did a deep dive into the boa_engine and found the following:

Engine Logic: The engine correctly routes through Path: UTF16 (Unicode/V flag).

Root Cause: The issue is in the regress dependency. I checked the source code of regress v0.10.5 in gen-unicode/src/main.rs and found that the Unicode path is hardcoded to an old version:
`const UCD_PATH: &str = "/tmp/ucd-15.1.0";` https://github.com/ridiculousfish/regress/blob/master/gen-unicode/src/main.rs line 24

Debug Logs:

Image

Image

This confirms that the internal tables in regress were not updated for Unicode 16.0, causing the match to fail.

amrkhaled104:
### Pull Request Description

This Pull Request closes #4791 

**It changes the following:**

-  Updated from `v0.10.5` to `v0.11.0` to support Unicode 17.0+ emoji sequences.
-  Removed the previously ignored `rgi-emoji-16.0.js` and `rgi-emoji-17.0.js` tests from `test262_config.toml` as they now pass with the updated library.

**Verification:**

image


amrkhaled104:
@jedel1043  the CI for the Wasm demo was passing previously  

image

PR #4985, opened 2026-03-10
fix(lexer): replace unimplemented! with syntax error for unknown oper…

tkshsbcue:
# fix(lexer): replace unimplemented! with syntax error for unknown operators

Replace `unimplemented!("operator {}", op)` in the operator lexer with `Err(Error::syntax(...))`. Unknown operator bytes now produce a recoverable syntax error instead of panicking.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+if ! command -v cargo-make >/dev/null 2>&1; then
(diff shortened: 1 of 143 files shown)


jedel1043:
You don't need to merge if you don't have merge conflicts by the way, our merge queue already resolves those.

tkshsbcue:
Sorry my bad @jedel1043

PR #4991, opened 2026-03-10
perf: use linked-list EnvironmentStack for O(1) clone on function calls

hansl:
Replace the flat Vec-based EnvironmentStack with a Gc-linked-list (EnvironmentNode chain) so that cloning for closures and function calls is O(1) instead of O(depth). Remove the `global` field from EnvironmentStack entirely—global environment is now accessed via `frame.realm.environment()` using split borrows, eliminating one Gc atomic ref bump per function call. Also switch CallFrame's binding_stack from Vec to ThinVec (saves 16 bytes per frame).

Benchmarks vs main: call.js ~6% faster, closure.js ~15% faster.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -291,9 +291,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.16.0"
+version = "1.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9a7b350e3bb1767102698302bc37256cbd48422809984b98d292c40e2579aa9"
+checksum = "94bffc006df10ac2a68c83692d734a465f8ee6c5b384d8545a636f81d858f4bf"
 dependencies = [
  "aws-lc-sys",
  "zeroize",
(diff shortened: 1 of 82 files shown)


core/engine/src/environments/runtime/mod.rs
```diff
@@ -17,15 +18,44 @@ pub(crate) use self::{
...
 pub(crate) struct EnvironmentStack {
-    stack: Vec<Environment>,
-    global: Gc<DeclarativeEnvironment>,
-    private_stack: Vec<Gc<PrivateEnvironment>>,
+    /// The tip (most recently pushed) environment in the chain.
+    tip: Option<Gc<EnvironmentNode>>,
+
+    /// Number of environments in the chain (not counting global).
+    #[unsafe_ignore_trace]
+    depth: u32,
```
jedel1043, line 49: I don't immediately see how having the depth is useful. We still need to go through the environment chain

hansl: We address environments by index. Without knowing the depth, we cannot (quickly) calculate how many parents we have to go through before we have the environment we're looking for.

jedel1043: I get that, but I think it would be more efficient to shift our bindings to count from the current environment up, instead of from the first environment down. That way you don't need to store the depth of every environment, you just walk up the environment chain and subtract 1 to the current binding's scope until you reach the correct environment.

That being said, this is more of a future optimization, so I don't think it blocks this PR.

jedel1043 (approved):
Looks good!

PR #4993, opened 2026-03-11
feat(runtime): add `Request.clone()` and `Response.clone()` per Fetch Standard §7.3/§7.4

### LINKED ISSUE #4992: Add Request.clone() and Response.clone() to boa_runtime fetch
### Summary
Implement the clone() method on Request and Response objects in boa_runtime, as required by the Fetch Standard §7.3 and §7.4.

### Current Behavior
Calling request.clone() or response.clone() throws a TypeError or returns undefined in boa_runtime. Real-world JS code (e.g., middleware, service workers, retry logic) routinely clones requests and responses to allow multiple reads of the body.

### Expected Behavior
Per the Fetch Standard:
- request.clone() returns a new Request with identical method, headers, body, and signal. It throws a TypeError if the body is already disturbed/locked.
- response.clone() returns a new Response with identical status, statusText, headers, and body. It throws a TypeError if the body is already disturbed/locked.

### Acceptance Criteria
- Request.prototype.clone() is implemented per spec §7.3
- Response.prototype.clone() is implemented per spec §7.4
- Throws TypeError if body is disturbed or locked
- Cloned request/response is independent (reading one does not affect the other)
- Basic tests covering: clone with body, clone without body, clone-then-read-both, error on disturbed body

### Spec References
https://fetch.spec.whatwg.org/#dom-request-clone
https://fetch.spec.whatwg.org/#dom-response-clone

Nakshatra480:
## Summary

Closes #4992

Implements `clone()` on `Request` and `Response` objects in `boa_runtime` per Fetch Standard §7.3 and §7.4.

## Changes

- **headers.rs**: Added `pub(crate) fn deep_clone()` on `JsHeaders` to produce a fully independent header map copy.
- **request.rs**: Added `body_used: Cell<bool>` tracking field and `clone()` method. Throws `TypeError` if body is already used.
- **response.rs**: Added `body_used: Cell<bool>` tracking field and `clone()` method. Throws `TypeError` if body is already used. `bytes()`, `text()`, and `json()` now also throw `TypeError` on a consumed body and mark it as used.

## Behavior

- `request.clone()` returns a new `Request` with an independent copy of method, headers, and body. Shares `AbortSignal` when signal support is wired into `JsRequest` (follow-up).
- `response.clone()` returns a new `Response` with independent copies of status, headers, body, and url.
- Both throw `TypeError` if the body has already been consumed.

## Testing

- [x] `cargo check -p boa_runtime` passes
- [x] `cargo clippy --all-features --all-targets -- -D warnings` passes
- [x] `cargo fmt --all -- --check` passes
- [x] `cargo test -p boa_runtime` passes

## Spec References

- https://fetch.spec.whatwg.org/#dom-request-clone
- https://fetch.spec.whatwg.org/#dom-response-clone


Nakshatra480:
Hi @jedel1043 can you pls review this pr whenever you get a chance.

core/runtime/src/fetch/request.rs
```diff
@@ -126,6 +126,8 @@ impl RequestInit {
 pub struct JsRequest {
     #[unsafe_ignore_trace]
     inner: HttpRequest<Vec<u8>>,
+    #[unsafe_ignore_trace]
+    body_used: std::cell::Cell<bool>,
```
jedel1043: I don't see why we need this. Does the spec deny reusing request and response bodies? Otherwise this doesn't really affect if we're cloning or not; we always clone when passing the body to the `text` and `bytes` promises.

jedel1043 (changes requested):

core/runtime/src/fetch/request.rs
Nakshatra480: I've removed body_used entirely. The spec's "unusable" check is for ReadableStream locking, but since we use Vec<u8> it doesn't apply. It's now just a straightforward deep copy, and I added a few JS tests for it.
Thanks for feedback

jedel1043 (approved):
LGTM

Ping me when you resolve the conflicts so that we can merge

Nakshatra480:
@jedel1043 i have resolved the merge conflicts.

Nakshatra480:
Thanks Jedel

PR #4994, opened 2026-03-11
Redesign CLI executor to avoid spin looping when idle

jedel1043:
The design of our current event loop on the CLI has a couple of warts; one of them being that if there are any async jobs or timeout jobs but not any other type of job on the queue, it will start spin looping, consuming a lot of CPU resources.

This PR completely redesigns our event loop to use `smol` and cooperative async tasks to push the jobs forward, which avoids spin looping by delegating pauses and timeouts to `smol` (effectively delegating those to the OS, and OS sleeping = 0% CPU usage). This redesign also modularizes the way jobs are ran; instead of having a big loop that goes through all of the job queues at once, it extracts the running logic of each job queue into independent async tasks


core/engine/src/job.rs
```diff
@@ -127,24 +127,47 @@ impl NativeJob {
     }
 }
 
+type Callback = Box<dyn FnOnce()>;
+
 /// Flag that can only be set once.
-#[derive(Debug, Clone)]
-pub(crate) struct OnceFlag(Rc<Cell<bool>>);
+#[derive(Clone)]
+pub(crate) struct OnceFlag(Rc<Cell<Option<Callback>>>);
```
hansl: Update comment, it's not longer a flag.

core/engine/src/job.rs
```diff
@@ -127,24 +127,47 @@ impl NativeJob {
...
 impl OnceFlag {
     /// Creates a new `OnceFlag`.
     pub(crate) fn new() -> Self {
-        Self(Rc::new(Cell::new(false)))
+        Self(Rc::new(Cell::new(Some(Box::new(|| {})))))
+    }
+
+    /// Sets a callback to run when the flag is set
+    pub(crate) fn set_callback(&self, f: impl FnOnce() + 'static) {
+        self.0.set(Some(Box::new(f)));
```
hansl, line 162: Should we do a `debug_assert!(self.0.is_none())`?

core/engine/src/job.rs
```diff
@@ -127,24 +127,47 @@ impl NativeJob {
...
     }
 
     /// Sets this `OnceFlag` to `true`.
     pub(crate) fn set(&self) {
-        self.0.set(true);
+        if let Some(fun) = self.0.take() {
```
hansl, line 167: This should be named something like `cancel()` and the struct `CancellationToken` or something.

core/engine/src/job.rs
jedel1043: Yep, makes sense

core/engine/src/job.rs
```diff
@@ -127,24 +127,58 @@ impl NativeJob {
...
+    /// used.
+    pub(crate) fn set_callback(&self, f: impl FnOnce() + 'static) {
+        #[cfg(debug_assertions)]
+        {
+            let callback = self.0.take();
+            assert!(
+                callback.is_some(),
+                "setting a callback on an already used cancellation token"
+            );
+        }
```
hansl: Isn't this just:

```suggestion
        debug_assert!(self.0.take().is_none(), "setting a callback on an already used cancellation token");
```

Also, I think the logic is wrong, you want to make sure it's `None` before setting it.

jedel1043: Nah, it needs to check if it is `Some`, otherwise the token has already been used and the `Cell` has a `None` on its place

zhuzhu81998:
there are merge conflicts now?

and can we pls merge this asap. i have the tendency to forget closing the cli and my core 1 suffers for this :face_holding_back_tears:

nekevss (approved):
LGTM

We should definitely merge this as soon as possible

PR #4998, opened 2026-03-11
Fix Number.prototype.toExponential() rounding (closes #4169)

### LINKED ISSUE #4169: Fix `Number.prototype.toExponential()` value rounding
**Describe the bug**

The output of incorrect when rounding with 5 as the last digit.

```javascript
(25).toExponential(0), // Should return "3e+1", actual "2e+1"
(12345).toExponential(3) // Should return "1.235e+4", actual "1.234e+4"
```

This cause the last built-in `Number` test262 test to fail: https://github.com/tc39/test262/blob/main/test/built-ins/Number/prototype/toExponential/return-values.js

Already have the issue in the `ryu-js` crate <https://github.com/boa-dev/ryu-js/issues/31>, just leaving this issue to avoid duplicate work 😅

xcb3d:
This Pull Request fixes/closes #4169.

It changes the following:

- Replaces Rust's `format!("{n:.prec$e}")` in `f64_to_exponential_with_precision` which uses **banker's rounding** (round-half-to-even) with a manual implementation that uses **round-half-away-from-zero** to match ECMAScript semantics
- The new implementation extracts full-precision digits via `format!("{n:.100e}")`, then rounds the digit string correctly before formatting the exponential output
- Fixes `(25).toExponential(0)` returning `"2e+1"` instead of `"3e+1"` and `(12345).toExponential(3)` returning `"1.234e+4"` instead of `"1.235e+4"`
- The test262 [`return-values.js`](https://github.com/tc39/test262/blob/main/test/built-ins/Number/prototype/toExponential/return-values.js) now passes in both normal and strict mode

### DIFF
--- a/core/engine/src/builtins/number/mod.rs
+++ b/core/engine/src/builtins/number/mod.rs
@@ -921,14 +921,66 @@ fn f64_to_exponential(n: f64) -> JsString {
 }
 
 /// Helper function that formats a float as a ES6-style exponential number string with a given precision.
-// We can't use the same approach as in `f64_to_exponential`
-// because in cases like (0.999).toExponential(0) the result will be 1e0.
-// Instead we get the index of 'e', and if the next character is not '-' we insert the plus sign
+///
+/// Uses round-half-away-from-zero to match ECMAScript semantics, instead of Rust's
+/// default banker's rounding (round-half-to-even).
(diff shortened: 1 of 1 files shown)


jedel1043:
The issue specifically states avoiding hacks using `format!`. We have a WIP implementation for this on ryu-js: https://github.com/boa-dev/ryu-js/pull/50

PR #4999, opened 2026-03-11
perf: use Vec::with_capacity for known-size allocations

tkshsbcue:
perf: use `Vec::with_capacity` for known-size allocations

## Description

This PR reduces unnecessary heap reallocations by replacing `Vec::new()` with `Vec::with_capacity()` in several hot paths where the collection size is known or can be reasonably estimated upfront.

When a `Vec` is created with `Vec::new()`, it starts with zero capacity and doubles its allocation each time it runs out of space. This leads to multiple reallocations and memory copies during growth. By pre-allocating with `Vec::with_capacity()`, we avoid these intermediate allocations entirely.

## Changes

| File | Function / Context | Optimization |
|------|-------------------|--------------|
| `builtins/string/mod.rs` | `String.fromCharCode()` | Pre-allocate from `args.len()` |
| `builtins/array/mod.rs` | `Array.prototype.join()` | Pre-allocate for elements + separators (`len * 2 - 1`) |
| `builtins/array/mod.rs` | `Array.prototype.toLocaleString()` | Pre-allocate for elements + separators (`len * 2 - 1`) |
| `builtins/uri/mod.rs` | `encode()` | Pre-allocate from input string length |
| `builtins/uri/mod.rs` | `decode()` | Pre-allocate from input string length |
| `error/mod.rs` | `AggregateError` construction | Reserve capacity from errors array length |
| `bytecompiler/statement/loop.rs` | Loop binding indices | Pre-allocate from bound names count |
| `object/operations.rs` | `enumerable_own_property_names()` | Pre-allocate from `own_keys` count |

## Why this matters

These are not micro-optimizations — several of these functions sit in frequently executed paths:

- `String.fromCharCode` and `Array.join` are among the most commonly called builtins in real-world JS
- URI encode/decode processes every character in the input string
- `enumerable_own_property_names` is called during `Object.keys()`, `Object.values()`, `Object.entries()`, `JSON.stringify`, and for-in loops

## Test plan

- [x] `cargo check` passes
- [x] `cargo test -p boa_engine --lib` — all 958 tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3613,9 +3613,9 @@ checksum = "dc897dd8d9e8bd1ed8cdad82b5966c3e0ecae09fb1907d58efaa013543185d0a"
 
 [[package]]
 name = "regress"
-version = "0.10.5"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2057b2325e68a893284d1538021ab90279adac1139957ca2a74426c6f118fb48"
+checksum = "07948de9abc2e83adbeb7543c061a5ddaf7d944afcafbdd6e6b39aeacd40504b"
 dependencies = [
  "hashbrown 0.16.1",
  "memchr",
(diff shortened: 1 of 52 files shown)


tkshsbcue:
@jedel1043 i hope you can review my PR thanks!

PR #5000, opened 2026-03-11
Fix f64 to f16 conversion for subnormal boundary rounding

xcb3d:
This Pull Request fixes incorrect `f64` to `f16` (Float16) conversion that caused 9 test262 failures.

It changes the following:

- Added a custom `f64_to_f16()` function in `core/engine/src/value/mod.rs` that correctly implements IEEE 754 binary16 conversion with roundTiesToEven, fixing a subnormal boundary rounding bug in the `float16` crate v0.1.5 where `2.980232238769532e-8` was incorrectly rounded to `0` instead of `5.960464477539063e-8` (smallest f16 subnormal).
- Updated `Math.f16round` in `core/engine/src/builtins/math/mod.rs` to use `f64_to_f16()` instead of `float16::f16::from_f64()`.
- Updated `JsValue::to_f16()` in `core/engine/src/value/mod.rs` to use `f64_to_f16()`, which propagates the fix to all Float16Array operations (fill, map, set, constructors, DefineOwnProperty, DataView.setFloat16).

### DIFF
--- a/core/engine/src/builtins/math/mod.rs
+++ b/core/engine/src/builtins/math/mod.rs
@@ -480,7 +480,7 @@ impl Math {
         // 4. Let n16 be the result of converting n to IEEE 754-2019 binary16 format using roundTiesToEven mode.
         // 5. Let n64 be the result of converting n16 to IEEE 754-2019 binary64 format.
         // 6. Return the ECMAScript Number value corresponding to n64.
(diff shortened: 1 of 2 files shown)


jedel1043:
I don't think it makes sense to fix this here, it should be fixed in the [`float16` crate](https://github.com/Alexhuszagh/float16).

PR #5001, opened 2026-03-11
fix(builtins/function/mod,function/arguments,function/bound,object/mod): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241.

It changes the following:
* `core/engine/src/builtins/function/arguments.rs`: converted 7 panics (7× `js_expect` on `downcast_ref::<MappedArguments>()` exotic method calls); `binding_indices` retains `.expect("binding must exist")` as a documented 
  compiler invariant — binding existence is guaranteed by `bound_names(formals)`
* `core/engine/src/builtins/function/mod.rs`: converted 10 panics (2× `downcast_ref::<OrdinaryFunction>()`, 3× `vm.frames.last()` frame existence, 1× `to_integer_or_infinity`, 1× `define_property_or_throw` length, 1× `define_property_or_throw` name, 1× `new.target` object cast, 1× `set_function_name`) to use `js_expect`
* `core/engine/src/builtins/function/bound.rs`: converted 2 panics (2× `downcast_ref::<BoundFunction>()` exotic method calls) to use `js_expect`
* `core/engine/src/builtins/object/mod.rs`: converted 10 panics (6× `CreateDataPropertyOrThrow`
* `from_property_descriptor`, 1× `create_data_property_or_throw` in `group_by`,1× `create_data_property_or_throw` in `get_own_property_descriptors`,1× `toObject` in `to_string`, 1× `ToObject` in `assign`) to use `js_expect`

### DIFF
--- a/core/engine/src/builtins/proxy/mod.rs
+++ b/core/engine/src/builtins/proxy/mod.rs
@@ -619,7 +619,7 @@ pub(crate) fn proxy_exotic_define_own_property(
     if !trap
         .call(
             &handler.into(),
-            &[target.clone().into(), key.clone().into(), desc_obj.into()],
+            &[target.clone().into(), key.clone().into(), desc_obj],
             context,
         )?
         .to_boolean()


KaustubhOG:
Note: set_function_name was changed from -> () to -> JsResult<()>. Call sites in vm/opcode/define/class/ and vm/opcode/set/property.rs were updated only to propagate this cascade ,remaining .expect() calls in those files are left because  vm opcodes are a separate batch (after /builtins complete)  in #3241

PR #5002, opened 2026-03-11
chore: Add husky-rs git pre-push hook to clippy and fmt

hansl:
No need to do that on every commit, but pushing will help people stop pushing things that fail Lint on CI. Like me. I do that.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,17 +1,9 @@
 #!/bin/sh
 
-echo "Running cargo fmt check..."
-cargo fmt --all -- --check
-if [ $? -ne 0 ]; then
-    echo "cargo fmt check failed. Run 'cargo fmt --all' to fix formatting."
+if ! command -v cargo-make >/dev/null 2>&1; then
+    echo "cargo-make is not installed. Install it with:"
+    echo "  cargo install cargo-make"
     exit 1
 fi
 
-echo "Running cargo clippy..."
-cargo clippy --all-targets --all-features -- -D warnings
-if [ $? -ne 0 ]; then
-    echo "cargo clippy failed. Fix the warnings above before pushing."
-    exit 1
-fi
-
-echo "All checks passed."
+cargo make run-ci


.husky/pre-push
```diff
@@ -0,0 +1,17 @@
...
+    echo "cargo fmt check failed. Run 'cargo fmt --all' to fix formatting."
+    exit 1
+fi
+
+echo "Running cargo clippy..."
+cargo clippy --all-targets --all-features -- -D warnings
+if [ $? -ne 0 ]; then
+    echo "cargo clippy failed. Fix the warnings above before pushing."
+    exit 1
+fi
```
jedel1043: I would just run `cargo make run-ci` TBH. It's an additional step of having to install `cargo-make`, but we should require it for contributions IMO

zhuzhu81998: what do you think about just adding a `.devcontainer.json` ?
then you can add whatever you want in there and make  dev environment more consistent.
(at least for people (incl. me) on linux life should get easier)

zhuzhu81998: and a vscode settings.json could let vscode use cargo to format files on save automatically.

hansl: @jedel1043 done.

@zhuzhu81998 most people don't have dev containers. Husky hooks into git so it always run for sure.

alienx5499:
Hi @jedel1043,
Hi @hansl,

I ran into an issue with the new pre-push hook on macOS when the shell
environment uses GCC as the default compiler (e.g. `CC=gcc-15`).

`cargo make run-ci` builds `aws-lc-sys`, which pulls macOS SDK headers
containing Objective-C block syntax. GCC doesn't support ObjC blocks,
so the build fails with errors like:

MDItem.h:178:69: error: expected ')' before '^' token

This happens because the hook inherits `CC`/`CXX` from the user's shell.

One possible fix would be forcing Clang inside `.husky/pre-push`:

export CC=clang
export CXX=clang++

If this approach looks reasonable, I can open a PR adding this to the hook.

jedel1043:
> This happens because the hook inherits CC/CXX from the user's shell.

That sounds like something you need to fix from your end then? Because adding that in the pre-push script would force Windows contributors to use clang instead of the VS build tools

alienx5499:
Maybe we could scope it only to macOS and only when CC isn’t already set, something like:

```
if [[ "$OSTYPE" == "darwin"* ]] && [ -z "$CC" ]; then
  export CC=clang
  export CXX=clang++
fi
```

That way it wouldn’t affect Windows or Linux environments.

jedel1043:
Honestly we should just kill aws-lc-rs with fire and use ring IMO

jedel1043:
Opened #5068

PR #5003, opened 2026-03-11
fix(runtime): implement spec-compliant HeadersIterator for Headers API

### LINKED ISSUE #4989: Headers.entries(), keys(), and values() return arrays instead of iterator objects
## Describe the bug

`Headers.entries()`, `Headers.keys()`, and `Headers.values()` currently return arrays instead of iterator objects.

Because of this, calling `.next()` on them fails. For example, `headers.entries().next()` throws instead of returning an iterator result.

Interestingly, `Headers[Symbol.iterator]()` already returns a working iterator.

---

## To reproduce

```javascript
let h = new Headers([["a", "b"], ["c", "d"]]);

console.log("typeof entries:", typeof h.entries());
console.log("entries is array:", Array.isArray(h.entries()));
console.log("entries has next:", h.entries().next !== undefined);

console.log("keys has next:", h.keys().next !== undefined);
console.log("values has next:", h.values().next !== undefined);

let it = h[Symbol.iterator]();
console.log("symbol iterator next:", JSON.stringify(it.next()));
```

Output:

```
typeof entries: object
entries is array: true
entries has next: false
keys has next: false
values has next: false
symbol iterator next: {"value":["a","b"],"done":false}
```

---

## Expected behavior

`entries()`, `keys()`, and `values()` should return iterator objects with a `.next()` method.

Example:

```javascript
const it = h.entries();
it.next(); // { value: ["a","b"], done: false }
```

---

## Possible location

The behavior appears to come from:

```
core/runtime/src/fetch/headers.rs
```

`entries()` currently constructs a `JsArray`, and there is also a comment in the file mentioning that it should eventually return an iterator.

---

## Build environment

OS: Windows 11  
Target: x86_64-pc-windows-msvc  
Rust: rustc 1.94.0

---

If this is expected behavior then feel free to close, otherwise I'd be happy to help work on a fix and add tests.

iammdzaidalam:
## Summary

Fix `Headers.entries()`, `Headers.keys()`, and `Headers.values()` so they return proper iterator objects instead of arrays.

Closes #4989

## Problem

`Headers.entries()`, `Headers.keys()`, and `Headers.values()` were not returning iterator objects.

Current behavior before this change:
- `entries()` returned an array
- `keys()` and `values()` returned collected values without a callable `.next()`
- `Headers[Symbol.iterator]()` already worked through a workaround

Because of this, calls like `headers.entries().next()` failed.

## What changed

- Added a `HeadersIterator` implementation for fetch headers
- Updated `Headers.entries()`, `Headers.keys()`, and `Headers.values()` to return iterator objects
- Kept `Headers[Symbol.iterator]()` aligned with iterator behavior
- Added regression tests covering iterator behavior and `for...of` usage

## Tests

Added/updated tests for:
- `entries()` returns an iterator
- `keys()` returns an iterator
- `values()` returns an iterator
- iterator results have the correct `{ value, done }` shape
- `[Symbol.iterator]()` still works
- `for...of` works with headers iteration

Ran locally:

```bash
cargo test
```
## Specification

According to the Fetch Standard, `Headers.entries()`, `Headers.keys()`, and
`Headers.values()` must return iterator objects.

Reference:
https://fetch.spec.whatwg.org/#headers-class

This change aligns Boa's `Headers` implementation with the expected iterator
semantics used by other JavaScript runtimes.

### DIFF
--- a/core/runtime/src/fetch/mod.rs
+++ b/core/runtime/src/fetch/mod.rs
@@ -213,17 +213,13 @@ fn headers_symbol_iterator(
         || js_error!(TypeError: "`Headers.prototype[Symbol.iterator]` requires a `Headers` object"),
     )?;
 
-    if !this_object.is::<JsHeaders>() {
+    let Ok(headers) = this_object.clone().downcast::<JsHeaders>() else {
         return Err(
             js_error!(TypeError: "`Headers.prototype[Symbol.iterator]` requires a `Headers` object"),
         );
-    }
+    };
 
-    HeadersIterator::create_headers_iterator(
-        this_object.clone().downcast().expect("checked above"),
-        IterationKind::KeyAndValue,
-        context,
-    )
+    HeadersIterator::create_headers_iterator(headers, IterationKind::KeyAndValue, context)
 }
 
 /// Register the `fetch` function in the realm, as well as ALL supporting classes.
--- a/core/runtime/src/fetch/tests/headers.rs
+++ b/core/runtime/src/fetch/tests/headers.rs
@@ -119,3 +119,30 @@ fn headers_invalid_inputs_throw_type_error_objects() {
         ),
     ]);
 }
+
+#[test]
+fn headers_iterator_throws_on_invalid_this() {
+    run_test_actions([
+        TestAction::harness(),
+        TestAction::inspect_context(register),
+        TestAction::run(
+            r#"
+                try {
+                    const iterator = Headers.prototype[Symbol.iterator].call({});
+                    throw Error("expected the call above to throw");
+                } catch (e) {
+                    assert(e instanceof TypeError);
+                    assertEq(e.message, "`Headers.prototype[Symbol.iterator]` requires a `Headers` object");
+                }
+
+                try {
+                    const iterator = Headers.prototype[Symbol.iterator].call(1);
+                    throw Error("expected the call above to throw");
+                } catch (e) {
+                    assert(e instanceof TypeError);
+                    assertEq(e.message, "`Headers.prototype[Symbol.iterator]` requires a `Headers` object");
+                }
+            "#,
+        ),
+    ]);
+}


core/runtime/src/fetch/headers_iterator.rs
```diff
@@ -0,0 +1,146 @@
...
+/// [spec]: https://fetch.spec.whatwg.org/#headers-class
+#[derive(Debug, Finalize, Trace, JsData)]
+pub(crate) struct HeadersIterator {
+    iterated_headers: JsObject,
+    next_index: usize,
+    #[unsafe_ignore_trace]
+    iteration_kind: PropertyNameKind,
+}
+
+impl Class for HeadersIterator {
```
jedel1043: Why not use `#[boa_class]` here instead of a manual implementation?

iammdzaidalam: Good point.

Refactored HeadersIterator to use #[boa_class].

There was a small issue with Symbol.iterator placement during the refactor (the method ended up on the constructor instead of the prototype), which caused the for...of test to fail. Adding #[boa(method)] alongside #[boa(symbol = "iterator")] fixed that.

The headers iterator tests are passing again, including the for...of case. Pushing the updated version.

jedel1043:
Needs rebase

iammdzaidalam:
Hi @jedel1043, I've addressed the rebase, lint, and coverage issues. All checks are passing now. Would appreciate a review whenever you have time. Thanks!

core/runtime/src/fetch/mod.rs
```diff
@@ -205,64 +203,45 @@ pub mod js_module {
...
-    let headers_proto = match realm {
-        Some(realm) => realm.get_class::<JsHeaders>(),
-        None => context.get_global_class::<JsHeaders>(),
+    js_module::boa_register::<F>(realm.cloned(), context)?;
+
+    if realm.is_none() {
+        context.register_global_class::<headers_iterator::HeadersIterator>()?;
+
+        // `#[boa_class]` cannot express symbol-keyed static properties, so
+        // we set `@@toStringTag` on the prototype manually after registration.
```
jedel1043: Is this true? I would expect something like
```rust
#[boa(static, symbol = to_string_tag)]
fn to_string_tag(){}
```
to work

iammdzaidalam: You're right that symbol-keyed members are parsed, so my comment was a bit too broad.
The issue is that the macro only emits methods/accessors. `@@toStringTag` here needs to be a string-valued data property on the prototype, so it can't be expressed via `#[boa(...)]`.

Manual definition is still required here... I'll update the comment.

iammdzaidalam: The issue here is a bit more specific: #[boa_class] only emits methods/accessors (method, static_method, accessor), but @@toStringTag needs to be a string-valued data property on the prototype.

Using #[boa(symbol = "toStringTag")] would create a function instead, and #[boa(static, ...)] would also put it on the constructor.

Since the macro doesn’t currently emit property() / static_property(), this still needs to be defined manually.

jedel1043: Oh now I get what you mean. Yeah, it would be good to add support in our macro to define plain properties (including statics) in classes.

iammdzaidalam: Yep, that would make this cleaner. Thanks.

core/runtime/src/fetch/mod.rs
```diff
@@ -205,64 +203,46 @@ pub mod js_module {
...
+            .expect("just registered")
+            .prototype();
+        proto.define_property_or_throw(
+            boa_engine::JsSymbol::to_string_tag(),
+            boa_engine::property::PropertyDescriptor::builder()
+                .value(JsString::from("Headers Iterator"))
+                .configurable(true)
+                .build(),
+            context,
+        )?;
```
jedel1043: Seems like this should also be defined even if `Realm` is `Some`, right? Otherwise it will only define `Symbol.toStringTag` if you don't pass a target realm.

iammdzaidalam: Good catch.

`fetch::register` now enters the target realm before running the registration logic, so `HeadersIterator` and its `Symbol.toStringTag` are defined in that realm as well, not just the default one.

Added cross-realm coverage for `response.headers.entries()` and verified with:
- `cargo test -p boa_runtime --lib -- fetch::tests::e2e`
- `cargo test -p boa_runtime fetch`
- `cargo test -p boa_runtime`

core/runtime/src/fetch/headers_iterator.rs
```diff
@@ -0,0 +1,119 @@
...
+/// The Headers Iterator object represents an iteration over a Headers object.
+/// It implements the iterator protocol.
+///
+/// More information:
+///  - [Fetch specification][spec]
+///
+/// [spec]: https://fetch.spec.whatwg.org/#headers-class
+#[derive(Debug, Finalize, Trace, JsData)]
+pub(crate) struct HeadersIterator {
+    iterated_headers: JsObject,
```
jedel1043:
```suggestion
    iterated_headers: JsObject<Headers>,
```

Maybe? That way you wouldn't need to `downcast_ref` inside `next`.

core/runtime/src/fetch/mod.rs
```diff
@@ -205,64 +203,57 @@ pub mod js_module {
...
     }
-    js_module::boa_register::<F>(realm.clone(), context)?;
 
-    // TODO(#4688): Replace this manual `[Symbol.iterator]` wiring once `#[boa(class)]`
-    // supports symbol-named methods.
-    let headers_proto = match realm {
-        Some(realm) => realm.get_class::<JsHeaders>(),
-        None => context.get_global_class::<JsHeaders>(),
+    let register_classes = |context: &mut Context| -> JsResult<()> {
+        js_module::boa_register::<F>(None, context)?;
```
jedel1043: `boa_register` already does what you're trying to do here with respect to registering either on the realm or on the global context, so you don't need to also duplicate that. What you need to do to fetch the prototype is to call `realm.get_class().expect().prototype()` on the realm if the realm was provided, or call `context.get_global_class().expect().prototype()` if it was not provided.

jedel1043: See the old code, specifically the `let headers_proto` part

core/runtime/src/fetch/headers_iterator.rs
iammdzaidalam: Sorry for the late reply... switched to JsObject<JsHeaders> which also let me drop the downcast_ref in next(). Thanks for the suggestion.

core/runtime/src/fetch/mod.rs
iammdzaidalam: dropped the extra logic and just use boa_register now. Also fixed the prototype lookup to use the realm/global get_class().expect().prototype() path. Thanks again @jedel1043 .

core/runtime/src/fetch/headers.rs
```diff
@@ -50,6 +54,9 @@ fn to_header_value(value: impl AsRef<str>) -> JsResult<HeaderValue> {
 pub struct JsHeaders {
     #[unsafe_ignore_trace]
     headers: Rc<RefCell<HttpHeaderMap>>,
+
+    #[unsafe_ignore_trace]
+    pub(crate) realm: Option<Realm>,
```
jedel1043: This would cause a memory leak, so...

Also, I don't think we need to preserve the realm here. If you create a `JsHeaders` in a realm then create its headers iterator in a separate realm, I would think the iterator would have the realm of the currently active realm, not the original realm of the object.

iammdzaidalam: Yeah u're right, we dont need to carry the realm here ... will drop it

iammdzaidalam: Done, removed the realm field and register the iterator in the active realm on demand. Fixes the cross-realm issue. Verified with headers_iterator_with_realm.

core/runtime/src/fetch/response.rs
```diff
@@ -141,6 +141,9 @@ pub struct JsResponse {
 
     #[unsafe_ignore_trace]
     body: Rc<Vec<u8>>,
+
+    #[unsafe_ignore_trace]
+    pub(crate) realm: Option<Realm>,
```
jedel1043: Again, this is a memory leak + it's enough to use the currently active realm

iammdzaidalam: Fixed. Response no longer stores a Realm; headers now uses the active Context/class registration instead. Rebased on latest main too.

iammdzaidalam:
Hey @jedel1043, any updates here when you get a moment?

iammdzaidalam:
Hey @jedel1043, any updates?

core/runtime/src/fetch/headers_iterator.rs
```diff
@@ -0,0 +1,134 @@
...
+                .define_property_or_throw(
+                    JsSymbol::to_string_tag(),
+                    PropertyDescriptor::builder()
+                        .value(JsString::from("Headers Iterator"))
+                        .configurable(true)
+                        .build(),
+                    context,
+                )
+                .expect("failed to set @@toStringTag");
+        }
```
jedel1043: IMO we should fail here instead of trying to register the class, since we should've already registered the class in `register`.

core/runtime/src/fetch/headers_iterator.rs
```diff
@@ -0,0 +1,110 @@
...
+        let iter = Self {
+            iterated_headers: headers,
+            next_index: 0,
+            iteration_kind: kind,
+        };
+
+        let proto = context
+            .realm()
+            .get_class::<Self>()
+            .expect("Headers Iterator not registered")
```
jedel1043: Don't panic here, just return an error instead

iammdzaidalam: yep, should be an error not a panic here. fixed it now.

core/runtime/src/fetch/mod.rs
```diff
@@ -202,21 +204,26 @@ pub mod js_module {
...
+    )?;
+
+    if !this_object.is::<JsHeaders>() {
+        return Err(
+            js_error!(TypeError: "`Headers.prototype[Symbol.iterator]` requires a `Headers` object"),
+        );
+    }
+
+    HeadersIterator::create_headers_iterator(
+        this_object.clone().downcast().expect("checked above"),
```
jedel1043: You should be able to downcast instead of calling `is` then `downcast`, right? That would avoid using `expect`.

iammdzaidalam: yep. removed the redundant is check and switched to a direct downcast with proper error handling instead of expect.

iammdzaidalam:
Hey @jedel1043 , please take a look when u get free.

PR #4811, opened 2026-03-02
fix(typedarray): implement spec-compliant [[PreventExtensions]] for resizable ArrayBuffers

mrhapile:
ECMAScript §10.4.5.4 requires Integer-Indexed exotic objects (TypedArrays) to override [[PreventExtensions]].

Currently, TypedArrays fall back to ordinary behavior, which allows Object.freeze() to proceed on resizable ArrayBuffers. According to the specification, [[PreventExtensions]] must:

### BEFORE

Screenshot 2026-03-02 at 2 45 53 PM
Screenshot 2026-03-02 at 2 47 19 PM

- Throw a TypeError if the buffer is detached
- Return false if the buffer is resizable
- Otherwise delegate to OrdinaryPreventExtensions

This PR introduces a TypedArray-specific [[PreventExtensions]] override aligned with the ECMAScript specification.

As a result, Object.freeze() correctly throws when applied to a TypedArray backed by a resizable ArrayBuffer, fixing the following test262 failure:

  test/built-ins/Object/freeze/typedarray-backed-by-resizable-buffer.js

All existing passing tests remain unaffected. Full Object.seal and Object.preventExtensions suites continue to pass without regression.

This Pull Request fixes/closes #{issue_num}.

It changes the following:

### AFTER

Screenshot 2026-03-02 at 3 35 07 PM

- Implements TypedArray-specific [[PreventExtensions]] per ECMAScript §10.4.5.4
- Ensures Object.freeze() throws for resizable ArrayBuffer-backed TypedArrays
- Aligns behavior with test262 expectations without modifying integrity-level logic

PR #5004, opened 2026-03-11
Fix: Implement [[PreventExtensions]] exotic method for TypedArray

mrhapile:
### Summary

**Implements the missing TypedArray exotic [[PreventExtensions]] behavior.**

According to `ECMAScript §10.4.5.1`, if `IsTypedArrayFixedLength(O`) is false (e.g. the underlying ArrayBuffer is resizable or length-tracking), the `[[PreventExtensions]]` internal method must return false.

Boa previously inherited the ordinary [[PreventExtensions]] implementation for TypedArrays, which skipped this check and allowed operations like Object.freeze or Object.seal to incorrectly succeed.

This PR adds the correct TypedArray exotic [[PreventExtensions]] implementation.

Changes

- Removed the earlier guard added in SetIntegrityLevel.
- Implemented typed_array_exotic_prevent_extensions in
- core/engine/src/builtins/typed_array/object.rs. 
- Added the method to the InternalObjectMethods table as
- __prevent_extensions__.
- Implemented the algorithm from ECMAScript §10.4.5.1:
  - If IsTypedArrayFixedLength(O) is false → return false.
  - Otherwise delegate to OrdinaryPreventExtensions.

**Test**
Fixes the Test262 case:

`test/built-ins/Object/freeze/typedarray-backed-by-resizable-buffer.js`

Result:
image

+1 Test262 pass

The behavior now matches the `ECMAScript specification`, where `Object.freeze / Object`.seal throw due to **[[PreventExtensions]]** returning false.

### DIFF
--- a/benches/scripts/basic/closure.js
+++ b/benches/scripts/basic/closure.js
@@ -1,0 +1,18 @@
+function outer() {
+  let x = 1;
+  function middle() {
+    let y = 2;
+    function inner() {
+      return x + y;
+    }
+    return inner;
+  }
+  return middle;
+}
+let f = outer()();
+
+function main() {
+  for (let n = 0; n < 10_000_000; n++) {
+    f();
+  }
+}
--- a/core/engine/src/builtins/eval/mod.rs
+++ b/core/engine/src/builtins/eval/mod.rs
@@ -14,7 +14,7 @@ use crate::{
     builtins::{BuiltInObject, function::OrdinaryFunction},
     bytecompiler::{ByteCompiler, prepare_eval_declaration_instantiation},
     context::intrinsics::Intrinsics,
-    environments::Environment,
+    environments::SavedEnvironments,
     error::JsNativeError,
     js_string,
     object::JsObject,
@@ -90,10 +90,9 @@ impl Eval {
 
         /// Possible actions that can be executed after exiting this function to restore the environment to its
         /// original state.
-        #[derive(Debug)]
         enum EnvStackAction {
             Truncate(usize),
-            Restore(Vec<Environment>),
+            Restore(SavedEnvironments),
         }
 
         // 1. Assert: If direct is false, then strictCaller is also false.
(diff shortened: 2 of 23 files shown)


core/engine/src/builtins/typed_array/object.rs
```diff
@@ -257,6 +258,31 @@ impl TypedArray {
...
+
+        ta.viewed_array_buffer().as_buffer().is_fixed_len()
+    };
+
+    if !is_fixed_length {
+        return Ok(false);
+    }
+
+    ordinary_prevent_extensions(obj, context)
+}
```
zhuzhu81998, line 285: i believe it is necessary to add the step-by-step spec instructions  to ensure compliance (even if that might seem redundant but others have done it).

and then maybe adjust the pr title since it is technically not about object.freeze.

mrhapile:
@zhuzhu81998 should I update the pr description also !?

mrhapile:
@jedel1043 thankyou for updating the title

mrhapile:
@zhuzhu81998 @jedel1043 #4811 look over this pr also

jedel1043:
@mrhapile yeah you should update the description actually

mrhapile:
@jedel1043 pr description updated!!!!

zhuzhu81998:
> #4811 look over this pr also

i am confused. so u had a pr that will solve this problem? and why does that pr have stuffs about some other stage 4 proposal too :joy:  did you mix several ones together?

jedel1043:
Yeah, that's a bit weird.

Just FYI, misusage of LLMs could result in an org ban, so make sure to review the code generated by LLMs.

Not gonna ban you because I could see this being an honest "I forgot I opened a PR doing the same changes".

mrhapile:
😅 I had another PR open earlier while I was researching this issue and trying a possible fix, but I couldn’t get it working properly at the time. Later I revisited it, went through the spec again, and opened this PR with the corrected implementation.
sorry @jedel1043 @zhuzhu81998 for any confusion🥲

zhuzhu81998:
ooops the PR contains a mistake :joy: 
`IsTypedArrayFixedLength` is not exactly the same as if the underlying buffer is fixed length or not (and this pr only checks if the buffer is fixed length): https://tc39.es/ecma262/#sec-istypedarrayfixedlength

See #5256

mrhapile:
@zhuzhu81998 yayaayaya 🥲 I think I should go through the spec once again I am still studying the spec, can you suggest me a better resource or way to understand the spec and codebase completely

zhuzhu81998:
> @zhuzhu81998 yayaayaya 🥲 I think I should go through the spec once again I am still studying the spec, can you suggest me a better resource or way to understand the spec and codebase completely

hmm i dont really have anything sry. i just look at the spec and MDN. and look around the codebase a bit ig

PR #5006, opened 2026-03-11
feat(string): implement RopeString with thin vtable and lazy flattening

### LINKED ISSUE #5005: Optimize string concatenation by introducing Rope strings
## Problem

Repeated string concatenation in Boa currently performs **eager buffer copying**, which leads to **O(N²)** behavior in workloads that repeatedly append to a string.

### Example:

```js
let s = "";
for (let i = 0; i < N; i++) {
    s += "a";
}
```

Each iteration allocates a new buffer and copies the previous string, causing quadratic growth in work and unnecessary memory allocations.

This behavior can significantly impact performance in string-heavy workloads such as template generation, logging, or repeated concatenation patterns.

## Proposed Solution
Introduce Rope strings to the `boa_string` crate. Ropes represent concatenations as a binary tree of strings rather than immediately copying buffers.

### Key ideas:

- Lazy flattening: only produce contiguous buffers when required
- Balanced tree construction for batch concatenations
- Depth limits to avoid pathological rope trees
- Iterative traversal to avoid recursion overflow

### Expected Benefits
- Eliminate O(N²) behavior in repeated string concatenation
- Reduce temporary allocations during string building
- Maintain near-baseline performance for general workloads
- Provide a foundation for further string optimizations

akash-R-A-J:
This Pull Request fixes/closes #5005 .

This PR introduces **Rope strings** to `boa_string` and integrates them into the engine’s concatenation pipeline to eliminate the pathological **O(N²)** behavior of repeated string concatenation.

### Rope Strings

- Added `RopeString` representation (`core/string/src/vtable/rope.rs`)
- Lazy flattening: ropes are only flattened when a contiguous buffer is required
- Balanced tree construction for batch concatenations
- **Fibonacci-based** rope rebalancing to maintain bounded depth
- **Leaf sinking optimization** to reduce rope node depth during concatenation
- Depth limit (32) and small-string threshold to avoid pathological trees
- Iterative traversal to prevent recursion overflow
- **Devirtualized hot paths** (clone, drop, as_str, code_unit_at)

### Engine Integration

- `JsValue::add` now creates ropes for large concatenations (`len(lhs) + len(rhs) > 1024`)
- `ConcatToString` updated to use balanced batch concatenation
- `String.prototype.concat` refactored to use the new API

### Thin VTable & Performance Recovery

- Moved method pointers to static vtables, reducing string header size
- Devirtualized hot paths (`clone`, `drop`, `as_str`, `code_unit_at`)
- Added cached `u64` hash to speed up property lookups
- Added `ptr_eq` + hash checks to optimize equality comparisons

### Concatenation Stress Test

```javascript
main : 2.876 s ±  0.025 s
rope : 132.4 ms ±   0.3 ms
```

≈ **21.7× faster**

### V8 Combined Suite

```javascript
main : 181.562 s ±  8.539 s
rope : 181.642 s ±  9.381 s
```

## Summary

This change removes quadratic concatenation behavior while maintaining baseline performance for general workloads. The thin vtable and devirtualization ensure that ropes introduce minimal overhead when not used.


core/string/src/vtable/mod.rs
```diff
@@ -1,39 +1,57 @@
...
-    /// Kind tag to identify the string type.
-    pub kind: JsStringKind,
+    /// Get the string as a `JsStr`.
+    pub as_str: fn(NonNull<RawJsString>) -> JsStr<'static>,
+    /// Get an iterator of code points.
+    pub code_points: fn(NonNull<RawJsString>) -> CodePointsIter<'static>,
+    /// Get the code unit at the given index.
+    pub code_unit_at: fn(NonNull<RawJsString>, usize) -> Option<u16>,
+    /// Deallocate the string.
+    pub dealloc: fn(NonNull<RawJsString>),
```
jedel1043: This refactor kinda doesn't make sense. `RawJsString` is not a "raw string", it's just information about a string, and each string will have additional fields that compose the full string. I think the old single `JsStringVTable` was honestly fine.

akash-R-A-J: Thanks for the feedback! The main motivation for the `RawJsString` and split `vtable` refactor was to reduce the string header size by moving the method pointers to a static structure. The goal was to improve cache locality and reduce memory overhead for applications with a very high number of small strings.

However, I completely agree that the abstraction adds complexity. I'll revert this to use the original single `JsStringVTable` structure as the header for all string types, while keeping the Rope logic and the new hash-caching intact.

jedel1043: Oh I see it now! Since all strings only store a single &'static ref instead of multiple function pointers, it lowers the amount of memory used. Yeah, I can get behind this. I would also suggest moving `JsStringKind` to it then, because that's also a static value that is shared between all strings with the same vtable.

core/string/src/vtable/rope.rs
```diff
@@ -0,0 +1,174 @@
...
+                len: left.len() + right.len(),
+                refcount: 1,
+                kind: JsStringKind::Rope,
+                hash: 0,
+            },
+            left,
+            right,
+            flattened: OnceCell::new(),
+            depth,
+        });
```
jedel1043: You might need to rebalance the rope following some kind of heuristic. Otherwise, you kinda lose the time complexity of O(log N) to get an element, since one side could get arbitrarily large, which converts indexing into an O(N) operation.

core/string/src/vtable/mod.rs
akash-R-A-J: Thanks for the suggestion! I've moved `JsStringKind` into the `static vtable`, which allowed the `RawJsString` header to remain at exactly 32 bytes (half a cache line).

core/string/src/vtable/rope.rs
akash-R-A-J: Yes, I've addressed rebalancing in two ways:

**Batch Concatenations:** The `concat_strings_balanced` helper uses a mid-point split to build a perfectly balanced tree (`log N` depth).

**Incremental Concatenations:** I've implemented a hard depth limit of 32 in `RopeString::create`. If this limit is exceeded, the rope auto-flattens into a contiguous string. This effectively bounds indexing to `O(32)`, which is `O(1)` in practice, preventing the `O(N)` worst-case.

This keeps the implementation simple while providing a guaranteed bound on access time

core/string/src/lib.rs
```diff
@@ -604,36 +630,153 @@ impl JsString {
...
-    /// # Safety
-    /// This should only be used when the inner type has been validated via `kind()`.
-    /// Using an unvalidated inner type is undefined behaviour.
+    /// Returns the string as a [`JsStr`].
     #[inline]
-    pub(crate) unsafe fn as_inner<T>(&self) -> &T {
-        // SAFETY: Caller must ensure the type matches.
-        unsafe { self.ptr.cast::<T>().as_ref() }
+    #[must_use]
+    pub fn as_str(&self) -> JsStr<'static> {
```
jedel1043: This is unsound. It's returning a `'static` reference to a temporary `JsString`, so you could deallocate the original string and the compiler will happily let you access the `JsStr<'static>`:

```rust
let s = JsString::from_str("undefined behaviour!");
let temp = s.as_str();
drop(s)
println!("{}", temp.display_lossy()); // UB!
```

core/string/src/vtable/rope.rs
jedel1043: > The concat_strings_balanced helper uses a mid-point split to build a perfectly balanced tree (log N depth).

That's not a balanced tree because you only consider the top strings, not if the strings themselves are ropes. If you pass an array of 3 strings, where 2 strings are ropes of depth 2 and 1 string is a rope of depth 28, you will have a very unbalanced tree.

> This effectively bounds indexing to O(32), which is O(1) in practice, preventing the O(N) worst-case.

This is _not_ O(N) in practice, because if you have thousands of strings of depth 32 and you need to index through all of them, you would pay the cost of O(N).

>

core/string/src/lib.rs
akash-R-A-J: Thanks for pointing this out! you're absolutely right, that was a leftover from the initial pointer-based POC while focusing on the memory layout. I've now refactored the `vtable` to use HRTBs for `as_str` which replaces that temporary shim with a lifetime binding. It was my fault.

core/string/src/vtable/rope.rs
akash-R-A-J: The current implementation uses a depth cap to avoid pathological trees, but you're right that it doesn't guarantee logarithmic access in all cases.

Would it make sense to adopt a Fibonacci-based rebalancing heuristic (similar to classical rope implementations) so that we guarantee weight-balanced ropes instead of relying on flattening?

jedel1043: Yep, that's pretty much the standard for ropes

core/engine/src/module/loader/mod.rs
```diff
@@ -62,9 +64,9 @@ pub fn resolve_module_specifier(
 
     // On Windows, also replace `/` with `\`. JavaScript imports use `/` as path separator.
     #[cfg(target_family = "windows")]
-    let specifier = specifier.replace('/', "\\");
+    let specifier = specifier.cow_replace('/', "\\");
 
-    let short_path = Path::new(&specifier);
+    let short_path = Path::new(&*specifier);
```
zhuzhu81998, line 69: what does this have to do with rope strings?

core/string/src/vtable/mod.rs
```diff
@@ -1,39 +1,63 @@
...
 pub(crate) mod r#static;
 pub use r#static::StaticString;
 
-/// Embedded vtable for `JsString` operations. This is stored directly in each string
-/// struct (not as a reference) to eliminate one level of indirection on hot paths.
-#[derive(Debug, Copy, Clone)]
+pub(crate) mod rope;
+pub(crate) use rope::RopeString;
+
+/// Header for all `JsString` allocations.
```
zhuzhu81998, line 21: Then call this `JsStringHeader` instead of `RawJsString` perhaps?

core/string/src/vtable/mod.rs
```diff
@@ -1,39 +1,63 @@
...
+    /// Reference count for this string.
+    pub(crate) refcount: usize,
+    /// Cached hash of the string content.
+    pub(crate) hash: u64,
+}
+
+// SAFETY: We only mutate refcount and hash via atomic-casts when kind != Static.
+unsafe impl Sync for RawJsString {}
+// SAFETY: RawJsString contains only thread-safe data.
+unsafe impl Send for RawJsString {}
```
zhuzhu81998: do you need this somewhere?

core/string/src/vtable/mod.rs
```diff
@@ -1,39 +1,63 @@
...
+    pub dealloc: fn(NonNull<RawJsString>),
+
+    /// Kind tag to identify the string type. Shared across all strings of this vtable.
     pub kind: JsStringKind,
 }
+
+// SAFETY: We only mutate refcount and hash via atomic-casts when kind != Static.
+unsafe impl Sync for JsStringVTable {}
+// SAFETY: JsStringVTable contains only thread-safe data.
+unsafe impl Send for JsStringVTable {}
```
zhuzhu81998: same question: where is the need?

core/string/src/vtable/rope.rs
```diff
@@ -0,0 +1,243 @@
...
+/// A rope string that is a tree of other strings.
+#[repr(C)]
+pub(crate) struct RopeString {
+    /// Standardized header for all strings.
+    pub(crate) header: RawJsString,
+    pub(crate) left: JsString,
+    pub(crate) right: JsString,
+    flattened: OnceCell<JsString>,
+    pub(crate) depth: u8,
+}
```
zhuzhu81998, line 101: hmmm so it is my understanding that `JsString` here will most likely GC-tracked. And there is no way for GC to know that `RopeString` is holdig reference to the `left` and `right` here?
are yo doing something to prevent this?

zhuzhu81998 (changes requested):
ci shows segmentaion fault (potential cause below)

core/string/src/vtable/rope.rs
jedel1043: To clarify: JSStrings are not GC-tracked, they're independently ref-counted, so it's fine to not trace through them. I'm fairly certain the UB comes from the modifications made to `as_str` that removed the lifetime inheritance for `JsStr<'static>`.

zhuzhu81998: indeed its not gc. disabling the gc does not resolve the segmentation fault XD.

core/string/src/vtable/rope.rs
```diff
@@ -0,0 +1,275 @@
...
+};
+
+/// A rope string that is a tree of other strings.
+#[repr(C)]
+pub(crate) struct RopeString {
+    /// Standardized header for all strings.
+    pub(crate) header: JsStringHeader,
+    pub(crate) left: JsString,
+    pub(crate) right: JsString,
+    flattened: OnceLock<JsString>,
```
jedel1043: Uhh I doubt you need a thread safe OnceCell here. `JsString` implements `!Send` and `!Sync`, so it's kinda unnecessary to make this thread safe.

core/string/src/vtable/mod.rs
akash-R-A-J: sounds good, will replace `RawJsString` with the suggested `JsStringHeader`

core/string/src/vtable/rope.rs
akash-R-A-J: Thanks! I've switched it to `OnceCell` to avoid any unnecessary thread-safe overhead.

akash-R-A-J: Reverting this back to `OnceLock` as `OnceCell` panics on reentrant initialization, and since rope trees can share nodes symmetrically, flattening may recursively call `.as_str()` on the same node, leading to reentrant initialization and test262 failures.

jedel1043: Uhh if you have that behaviour that's a bug though... because if you have a children that recursively references itself, things like `code_point_at` could infinitely recurse.

akash-R-A-J:
@jedel1043 I believe all requested changes have now been addressed. CI is green across all platforms, and test262 results match `main`.

Please let me know if anything else needs adjustment.

core/string/src/vtable/rope.rs
akash-R-A-J: This is not a self reference cycle. Since `JsString` is `refcounted`, ropes can share subtrees and form DAGs. During flattening the same node may be reached through multiple paths before its cache is initialized, which triggers reentrant initialization in `OnceCell`.

jedel1043: Then I don't see the reentrancy. If you have a children that is shared between two nodes, the first access would build the flat version, and the second access would just fetch the already constructed array, no?

But I think this is all null because your code already does a DFS so it should only call `as_str` for the leaves, which are never ropes.

akash-R-A-J: Ahh, you're right that the DFS traversal only calls `as_str()` on leaf strings, not ropes, so `rope_as_str` itself shouldn't recursively reenter. The earlier panic I observed with `OnceCell` likely came from another path (possibly inside `concat_array`). I'll investigate that path more closely.

core/string/src/vtable/rope.rs
```diff
@@ -58,18 +82,23 @@ pub(crate) static ROPE_VTABLE: JsStringVTable = JsStringVTable {
...
-    // We use `OnceLock` over `OnceCell` despite `JsString` being `!Sync`.
-    // The reason is that rope flattening is structurally reentrant: `rope.as_str()` can internally
-    // trigger `.as_str()` on its children, which might be identical shared references (DAG).
-    // `OnceCell` explicitly panics on reentrant initialization, whereas `OnceLock` handles it.
-    flattened: OnceLock<JsString>,
+    // Using a raw buffer cache instead of `JsString` in `OnceCell` solves the "refcount aliasing"
+    // problem. Storing a refcounted `JsString` inside a `OnceCell` could lead to ownership cycles
+    // or use-after-free errors if shared nodes were dropped while still cached.
+    // By storing raw `Box<[u8]>` or `Box<[u16]>`, we decouple the cache from the refcounting system.
+    flattened: OnceCell<Flattened>,
```
jedel1043, line 99: Just use `Option<JsString>`...

akash-R-A-J: okay

jedel1043: Sorry it should be `Cell<Option<JsString>>`

jedel1043: Otherwise you won't be able to modify it 😅

jedel1043: Ah wait that doesn't work, you need to have something referenceable

jedel1043: Hmmm... what if we just make `as_str` return `Option<JsStr>`? That kinda makes more sense since flattening ropes should be reserved to only exceptional situations such as reaching the maximum depth.

jedel1043: This also nicely reflects in our inner API that not all strings have a `JsStr` available to use, and the caller should be the one responsible for accessing the string in other ways if that's the case

akash-R-A-J: That makes sense from an API design perspective. However, changing `as_str()` to return `Option<JsStr>` would require a fairly large refactor across the engine since many built-ins and internal methods currently rely on it always returning a `JsStr`. Would it be okay if I explore that in a separate follow-up PR?

akash-R-A-J:
Hey @jedel1043, whenever you have some time, could you please take another look at this PR?

All previously mentioned issues should now be addressed, including the rope rebalancing changes, safety fixes, and flattening adjustments discussed in the earlier review.

Here's the latest benchmark results:

final_benchmark_rope

core/engine/src/builtins/date/utils.rs
```diff
@@ -753,22 +752,22 @@ pub(super) fn pad_six(t: u32, output: &mut [u8; 6]) -> JsStr<'_> {
 /// [spec-format]: https://tc39.es/ecma262/#sec-date-time-string-format
 pub(super) fn parse_date(date: &JsString, hooks: &dyn HostHooks) -> Option<i64> {
     // All characters must be ASCII so we can return early if we find a non-ASCII character.
-    let owned_js_str = date.as_str();
-    let date = match owned_js_str.variant() {
-        JsStrVariant::Latin1(s) => {
+    let date = match date.as_flat_str() {
```
jedel1043: Do we need `as_flat_str` here? We could directly convert to a `String` if the `JsStr` is a rope, right?

core/engine/src/builtins/number/globals.rs
```diff
@@ -359,13 +358,14 @@ pub(crate) fn parse_float(
         return Ok(JsValue::nan());
     }
 
-    let value = match trimmed_string.variant() {
-        JsStrVariant::Latin1(s) => fast_float2::parse_partial::<f64, _>(s),
-        JsStrVariant::Utf16(s) => {
+    let value = match input_string.as_flat_str() {
```
jedel1043: Same here, we should convert to `String` if the str is a rope.

core/engine/src/builtins/regexp/mod.rs
```diff
@@ -1157,20 +1157,20 @@ impl RegExp {
 
         // 13.b. Let inputIndex be the index into input of the character that was obtained from element lastIndex of S.
         // 13.c. Let r be matcher(input, inputIndex).
-        let r: Option<regress::Match> = match (full_unicode, input.as_str().variant()) {
-            (true | false, JsStrVariant::Latin1(_)) => {
+        let r: Option<regress::Match> = match (full_unicode, input.as_flat_str()) {
```
jedel1043: This one's a bit more complex, but we could use an iterator just like the `JsStr::Latin1` branch

zhuzhu81998:
the failed tests are probably related.

core/engine/src/builtins/date/utils.rs
akash-R-A-J: thanks for pointing this out, have incorporated it in the latest commit.

core/engine/src/builtins/regexp/mod.rs
akash-R-A-J: got it

core/string/src/vtable/mod.rs
```diff
@@ -1,39 +1,73 @@
...
+/// Header for all `JsString` allocations.
+///
+/// This is stored at the beginning of every string allocation.
+/// By using a reference to a static vtable, we reduce the header size
+/// and improve cache locality for common operations.
+#[repr(C, align(8))]
+#[derive(Debug, Clone, Copy)]
+pub struct JsStringHeader {
+    /// Reference to the static vtable for this string kind.
+    pub vtable: &'static JsStringVTable,
```
HalidOdat, line 30: I like the thin vtable here, we could make this even more space efficient if we store the kind only and have a table of vtables (u8 that we could potentially even encode in the len or refcount), the reference to a vtable makes more sense for object that we can't determine at compile time and are not limited (but `JsString` types are limited 😄).

```rust
const VTABLES = [JsStringVTable; {num vtables}] = [
    LATIN1_VABLE,
    ROPE_VTABLE,
    // ...
];
// ...
// usage:
let dealloc = VTABLES[header.kind as usize].dealloc;
```

Keeping it by value instead of reference in the table may also help with cache locality for all JsString internal implementations.

core/string/src/vtable/rope.rs
```diff
@@ -0,0 +1,436 @@
+use crate::vtable::{JsStringHeader, JsStringVTable};
+use crate::{JsStr, JsString, JsStringKind};
+use std::cell::OnceCell;
+use std::ptr::{self, NonNull};
+
+/// Fibonacci numbers for rope balancing thresholds.
+/// `F(n) = Fib(n + 2)`. A rope of depth `n` is balanced if its length >= `F(n)`.
+static FIBONACCI_THRESHOLDS: [usize; 46] = [
```
HalidOdat, line 8: Suggestion: We can shrink this lookup table by half here (at least on 64bit systems) 😄 
```suggestion
static FIBONACCI_THRESHOLDS: [u32; 46] = [
```

core/engine/src/string.rs
```diff
@@ -62,10 +62,10 @@ macro_rules! js_string {
         $crate::string::JsString::from($s)
     };
     ( $x:expr, $y:expr ) => {
-        $crate::string::JsString::concat($crate::string::JsStr::from($x), $crate::string::JsStr::from($y))
+        $crate::string::JsString::concat(&$crate::string::JsString::from($x), &$crate::string::JsString::from($y))
```
HalidOdat, line 65: Suggestion: Recursively calling js_string here, has the benefit of creating a static string instead of two dynamic strings. (i.e. `js_string!(s, ";")`)
```suggestion
        $crate::string::JsString::concat(&crate::js_string!($x), &crate::js_string!($y))
```

core/engine/src/string.rs
```diff
@@ -62,10 +62,10 @@ macro_rules! js_string {
...
     };
     ( $( $s:expr ),+ ) => {
-        $crate::string::JsString::concat_array(&[ $( $crate::string::JsStr::from($s) ),+ ])
+        $crate::string::JsString::concat_array_strings(&[ $( $crate::string::JsString::from($s) ),+ ])
```
HalidOdat, line 68:
```suggestion
        $crate::string::JsString::concat_array_strings(&[ $( $crate::js_string!($s) ),+ ])
```

Issue #4986, opened 2026-03-10
Make call frame boundaries explicit in VM trace


ashddev:
This is a sub-task of [#1711](https://github.com/boa-dev/boa/issues/1711)

The idea is to have the call frame output look similar to:

```
----------------- Call Frame foo -----------------
Time     Opcode        Operands      Stack
1μs      PushLiteral   0             ["hello"]
3μs      Return                      ["hello"]
----------------- Call Frame bar -----------------
Time     Opcode        Operands      Stack
134μs    GetPropertyByName 0004:'test'  ["hello"]
```

MayankRaj435:
hii,i would like to work on this issue

ashddev:
Hi @MayankRaj435 sorry I'm currently working on this issue

PR #5010, opened 2026-03-12
feat(vm): explicit call frame boundaries

ashddev:
This Pull Request fixes/closes [#4986](https://github.com/boa-dev/boa/issues/4986)

It changes the following:
- Trace a new call frame whenever a new frame is pushed onto the stack.
- Ensure each code block’s compiled output is printed only once.


jedel1043:
Before:

image


After:

image

Looks very clean!

PR #5012, opened 2026-03-12
feat(error): harden JsNativeError public API surface

### LINKED ISSUE #4981: Public API surface: harden `JsNativeError`
## Problem

- `JsNativeError` is part of Boa’s public Rust API and will be very hard to change once embedders depend on it.
- The `kind` field on `JsNativeError` is currently `pub`, which:
  - encourages direct field access instead of a stable getter,
  - makes future refactors (e.g. storing additional metadata, changing representation) a potential breaking change.
- For a stable public API, we should prefer explicit constructors and getters over public fields.

## Proposal

- Treat `JsNativeError` as a stable abstraction and narrow its public surface:
  - Make the `kind` field private.
  - Add a `kind()` getter (and, if needed, additional accessors) for reading the error kind.
- Keep `JsNativeErrorKind` as the public enum (already marked `#[non_exhaustive]`), but avoid exposing the internal layout of `JsNativeError` itself.

## Scope / API surface

- File: `core/engine/src/error/mod.rs`
- Types:
  - `pub struct JsNativeError`
  - `pub enum JsNativeErrorKind` (already `#[non_exhaustive]`)

## Rationale

- **Helps stabilize the public error API**:
  - Hardens one important error type’s public surface before it is widely relied on.
  - Reduces the risk of future internal changes becoming breaking API changes.
- **SemVer‑friendly**:
  - `JsNativeErrorKind` remains `#[non_exhaustive]`, so new variants are still addable.
  - Moving from `pub` field to a getter is done early, while breaking changes are still acceptable.
- **Minimal risk**:
  - Changes are limited to the error module and call sites that read `kind` directly.
  - No change to the semantics of error construction or display.

## Implementation outline

- Change `JsNativeError` definition:
  - `pub kind: JsNativeErrorKind,` → `kind: JsNativeErrorKind,`
  - Add `pub fn kind(&self) -> &JsNativeErrorKind`.
    - Returning a reference avoids unnecessary copies and keeps flexibility if the enum grows or changes representation.
- Update internal Boa code to use the getter instead of the field where needed.
- Audit public docs and examples that mention `JsNativeError` and update them to use the getter.

## Compatibility note

- This is technically breaking for downstream users who access `err.kind` directly.
- Making this change now, as part of public API surface hardening, avoids larger long‑term breakage once embedders rely on the current layout.

## Checklist

- [ ] Make `JsNativeError::kind` private.
- [ ] Add a stable getter for the error kind on `JsNativeError`.
- [ ] Update internal call sites to use the getter instead of the public field.
- [ ] Run `cargo test -p boa_engine` and docs to ensure no regressions.
- [ ] If needed, update docs to reflect the accessor.

---

alienx5499:
This Pull Request fixes/closes #4981.

It changes the following:

* Make `JsNativeError::kind` a private field instead of `pub`.
* Add a `JsNativeError::kind(&self) -> &JsNativeErrorKind` accessor as the stable API.
* Update internal usages (engine tests, runtime tests, URI tests, tester helpers, docs/examples) to call `kind()` instead of reading the field directly.

Testing:

* `cargo test -p boa_engine`

### DIFF
--- a/core/engine/src/builtins/uri/mod.rs
+++ b/core/engine/src/builtins/uri/mod.rs
@@ -587,6 +587,6 @@ mod tests {
         let s = js_string!("%E7%9A%8");
         let err = decode(&s, |_| false).expect_err("should error on incomplete escape");
         let native = err.as_native().expect("error should be native");
-        assert!(matches!(native.kind, JsNativeErrorKind::Uri));
+        assert!(matches!(native.kind(), JsNativeErrorKind::Uri));
     }
 }
--- a/core/engine/src/error/mod.rs
+++ b/core/engine/src/error/mod.rs
@@ -202,9 +202,8 @@ macro_rules! js_error {
 ///     .with_cause(cause)
 ///     .into();
 ///
-/// assert!(native_error.as_native().is_some());
-///
-/// let kind = &native_error.as_native().unwrap().kind;
+/// let native = native_error.as_native().unwrap();
+/// let kind = native.kind();
 /// assert!(matches!(kind, JsNativeErrorKind::Type));
 /// ```
 #[derive(Debug, Clone, Trace, Finalize)]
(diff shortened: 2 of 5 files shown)


jedel1043 (approved):
Cool, thank you!

PR #5017, opened 2026-03-12
fix(string): cleanup code in String.prototype.repeat

ghost:
Overview
This PR refactors the String.prototype.repeat implementation to strictly align with Section 22.1.3.16 of the ECMA-262 specification. The previous implementation used a generic error handling match; this update introduces explicit checks for Infinity and negative values as required by the abstract operations.

Spec Compliance Audit
I have mapped the implementation to the following specification steps:

Step 3: Let n be ? ToIntegerOrInfinity(count).

Step 4: If n < 0, throw a RangeError exception. (Implemented via explicit match)

Step 5: If n is +∞, throw a RangeError exception. (Implemented via PositiveInfinity check)

Changes
Replaced the catch-all error handling with an exhaustive match on the IntegerOrInfinity enum.

Added specific RangeError messages that distinguish between negative counts and infinite counts.

Used variable shadowing to simplify the logic flow from an Enum to a primitive i64 after validation.

Verification
I have verified the fix using a local test script across the following edge cases:

JavaScript
"abc".repeat(0);        // Returns "" (Success)
"abc".repeat(-1);       // Throws RangeError: count must be non-negative (Success)
"abc".repeat(Infinity); // Throws RangeError: count must be less than infinity (Success)
Note to Maintainers
I am a prospective GSoC 2026 applicant interested in the ECMAScript Conformance project. I chose this task to demonstrate my ability to translate ECMA-262 algorithms into idiomatic Rust using Boa's internal types. I am eager to contribute more to the engine's compliance suite!

PR #5018, opened 2026-03-12
feat(string): implement OOM protection and spec-alignment for padStart/padEnd

ghost:
Overview
This PR improves the robustness of String.prototype.padStart and String.prototype.padEnd. While the basic functionality existed, the implementation lacked safeguards against massive memory allocations and didn't perfectly match the early-return execution order required by the ECMA-262 specification (Sections 22.1.3.14 & 22.1.3.15).

Spec Compliance & Safety Improvements
I have updated the logic to align with the following abstract operations:

Step 5 (Early Return): If intMaxLength is less than or equal to the current string length, the function now returns the original string immediately before processing the filler.

Step 9 (Empty Filler): If the filler string is empty, it returns the original string, preventing infinite loop logic or unnecessary allocation.

Memory Safety (OOM Protection): Added a validation check against String::MAX_STRING_LENGTH.

Previously: Attempting "a".padStart(Number.MAX_SAFE_INTEGER) could lead to a heap allocation failure/crash.

Now: It correctly throws a RangeError, matching the behavior of high-performance engines like V8.

Changes
Refactored fn pad in core/engine/src/builtins/string/mod.rs to follow spec steps 1-13 sequentially.

Integrated String::MAX_STRING_LENGTH check to ensure engine stability.

Simplified undefined argument handling for the fillString parameter.

Verification
Verified with a custom test suite covering:

Large maxLength values (Triggering RangeError).

undefined and empty string fillers.

maxLength values smaller than the target string.

Note to Maintainers
This is my second contribution as a GSoC 2026 applicant for the ECMAScript Conformance project. After my previous PR for String.prototype.repeat (#5017), I am continuing to audit the String built-ins to ensure they are both spec-compliant and memory-safe.


jedel1043 (approved):
Looks good

core/engine/src/builtins/string/mod.rs
```diff
@@ -1533,6 +1533,17 @@ impl String {
 
         // 8. Let fillLen be intMaxLength - stringLength.
         let fill_len = int_max_length - string_length;
+
+        // Check if the resulting string would exceed the maximum string length
+        if int_max_length > (Self::MAX_STRING_LENGTH as u64) {
+            return Err(JsNativeError::range()
+                .with_message(format!(
+                    "padStart/padEnd would create a string longer than the maximum allowed length ({})",
```
jedel1043:
```suggestion
                    "cannot create a string longer than the maximum allowed length ({})",
```

ghost:
Thanks for the merge, @jedel1043 !  It’s been great getting familiar with the String built-ins.

Since I'm looking to dive deeper into the engine, I’ve been exploring the core/engine/src/builtins directory. I noticed we have some interesting work happening in the Intl and Temporal modules for the v1.0.0 milestone. From a maintenance perspective, are there any specific areas in the built-ins that you feel currently need the most 'spec-alignment' cleanup or refactoring? I'd love to help tackle some of the more complex logic gaps next .

PR #5024, opened 2026-03-13
perf: use FxHashSet for O(1) name lookups in declaration instantiation

tkshsbcue:
## Description

This PR replaces `Vec` with `FxHashSet` for name-tracking collections in the bytecode compiler's declaration instantiation functions, reducing `.contains()` lookups from O(n) to O(1).

### The problem

During script/module/function compilation, the bytecode compiler tracks declared function names, variable names, and parameter bindings using `Vec<Sym>`. These vectors are repeatedly checked with `.contains()` inside loops, creating O(n^2) behavior when a script has many declarations.

There are **22 `.contains()` calls** in `declarations.rs`, many inside nested loops:

```rust
// O(n) outer loop x O(n) contains = O(n^2)
for declaration in var_declarations {
    for name in bound_names(&declaration) {
        if !declared_function_names.contains(&name) {  // O(n) per check
            if !declared_var_names.contains(&name) {   // O(n) per check
                declared_var_names.push(name);
            }
        }
    }
}
```

### The fix

Replace `Vec<Sym>` with `FxHashSet<Sym>` for all name-tracking collections used only for membership checks. Where a collection is also iterated (e.g., `declared_var_names` for emitting bindings), a parallel `FxHashSet` is maintained alongside the `Vec`.

`FxHashSet` (from `rustc_hash`) is already used elsewhere in the codebase and provides faster hashing than `std::collections::HashSet` for small integer keys like `Sym`.

### Affected functions

| Function | Collections converted |
|----------|----------------------|
| `global_declaration_instantiation_context` | `declared_function_names`, `declared_var_names` |
| `prepare_eval_declaration_instantiation` | `declared_function_names`, `lexically_declared_names` |
| `global_declaration_instantiation` | `declared_function_names`, `declared_var_names` |
| `eval_declaration_instantiation` | `declared_function_names`, `declared_var_names` |
| `function_declaration_instantiation` | `function_names`, `parameter_bindings`, `instantiated_var_names` |

## Why this matters

Declaration instantiation runs on **every script, eval, and function compilation**. Real-world JavaScript bundles routinely have hundreds or thousands of declarations. The O(n^2) behavior is particularly visible when:

- Loading large bundled scripts (webpack/rollup output with many top-level declarations)
- Using `eval()` with significant code
- Compiling functions with many local variables

## Test plan

- [x] `cargo check` passes
- [x] `cargo clippy` -- no warnings
- [x] `cargo test -p boa_engine --lib` -- all 958 tests pass


tkshsbcue:
@jedel1043 i think you might fight this interesting haha!

zhuzhu81998:
do you have benchmark results? :eyes: `cargo bench -p boa_benches` first on main and then on pr branch would be nice.

tkshsbcue:
hey hi @zhuzhu81998 these are the benchmarks i compiled

| Benchmark | main | PR | Change | Significant? |
|-----------|------|-----|--------|-------------|
| strings/slice | 1.030 ms | 1.007 ms | **-2.3%** | Yes (improved) |
| v8/deltablue | 5.052 µs | 4.973 µs | **-1.4%** | Yes (improved) |
| v8/richards | 5.018 µs | 4.957 µs | **-1.5%** | Yes (improved) |
| v8/splay | 5.019 µs | 4.939 µs | **-1.5%** | Yes (improved) |

jedel1043:
Need to fix formatting first

tkshsbcue:
hey @jedel1043 fixed formatting!

zhuzhu81998:
The names that are hashed here, they are not internal right? They come from JS?
So maybe `FxHashSet` is not ok here, since usecases of Boa probably include executing untrusted scripts (plugins etc.).

tkshsbcue:
hey @zhuzhu81998 

The keys here are Sym, which is just a NonZeroUsize  it's an integer index into the string interner, not the raw JS string itself. FxHash on sequential/near-sequential   
integers is essentially a perfect hash with no realistic collision scenario. The actual JS name strings never reach the hasher.

core/engine/src/bytecompiler/declarations.rs
```diff
@@ -718,6 +710,7 @@ impl ByteCompiler<'_> {
 
         // 12. Let declaredVarNames be a new empty List.
         let mut declared_var_names = Vec::new();
```
zhuzhu81998: do we still need this separately?

zhuzhu81998 (changes requested):

core/engine/src/bytecompiler/declarations.rs
tkshsbcue: Are you suggesting we drop the Vec and use only the FxHashSet here?

zhuzhu81998: yea unless there is a reason to keep this that i am not seeing? you dropped the `Vec` elsewhere already no?

jedel1043: We _could_ try to drop it, but IIRC the spec requires vars to be initialized in a certain order, so I don't know if using a hash set will break assumptions about this

tkshsbcue: Hmm, looking at it  the iteration at step 18 just calls CreateGlobalVarBinding for each name, and those bindings are independent of each other. Wouldn't the end result be the same regardless of order?

jedel1043: Only way to find out is to remove this and see if any tests break :)

tkshsbcue: :)

tkshsbcue: uhhh what should i do then lmao?

jedel1043: > Are you suggesting we drop the Vec and use only the FxHashSet here?

This, but use a separate commit just in case it breaks tests. Makes reversing easier

tkshsbcue: alright i will implement this and push another follow up commit !

tkshsbcue: hey @jedel1043 please have a look!

core/engine/src/bytecompiler/declarations.rs
```diff
@@ -258,12 +256,10 @@ pub(crate) fn prepare_eval_declaration_instantiation(
...
-        if !declared_function_names.contains(&name.sym()) {
+        // 2. Append fn to declaredFunctionNames.
+        if declared_function_names.insert(name.sym()) {
             // SKIP: 1. If varEnv is a Global Environment Record, then
 
-            // 2. Append fn to declaredFunctionNames.
-            declared_function_names.push(name.sym());
-
             // SKIP: 3. Insert d as the first element of functionsToInitialize.
         }
```
zhuzhu81998: do we still need this empty `if` block?

core/engine/src/bytecompiler/declarations.rs
```diff
@@ -272,7 +268,8 @@ pub(crate) fn prepare_eval_declaration_instantiation(
     // 11. If strict is false, then
     #[cfg(feature = "annex-b")]
     if !strict {
-        let lexically_declared_names = lexically_declared_names(body);
+        let lexically_declared_names: FxHashSet<Sym> =
+            lexically_declared_names(body).into_iter().collect();
```
zhuzhu81998, line 268: the doc of `lexically_declared_names` says:
```rust
/// Returns a list with the lexical bindings of a node, which may contain duplicates.
///
/// This is equivalent to the [`LexicallyDeclaredNames`][spec] syntax operation in the spec.
///
/// [spec]: https://tc39.es/ecma262/#sec-static-semantics-lexicallydeclarednames
```

what cases will result in duplicates here? 🤔

core/engine/src/bytecompiler/declarations.rs
tkshsbcue: Haha Good catch  after the Vec→FxHashSet swap the body is just // SKIP: spec-comment markers, so the if wraps nothing. I'll drop it (same applies to the matching blocks at lines 93, 422, 668, and 944) and keep just declared_function_names.insert(name.sym());.

core/engine/src/bytecompiler/declarations.rs
tkshsbcue: Duplicates can occur in the !strict branch via Annex B  e.g. sibling blocks each declaring the same function name, or repeated function bindings across nested CaseClauses. That said, dedup isn't really the goal of the change here; the loop below does .contains(&f) for every Annex B function name, so the win is O(1) lookup vs O(n) scan. I'll update the commit message / PR description to make that the stated motivation rather than "duplicates"

zhuzhu81998: my concern is whether or not the preservation of duplicates is necessary here for correctness/spec-compliance

tkshsbcue: No dedup is safe here. The only use of lexically_declared_names in this block is the .contains(&f) membership check in the Annex B loop below:
 
                                                                                                                                                                            
```
  for f in annex_b_function_declarations_names(body) {                                                                                                                       
      if !lexically_declared_names.contains(&f) {                                                                                                                            
          ...                                                                                                                                                                
      }                                                                                                                                                                      
  }   
```

zhuzhu81998 (approved):
Other than linting it looks fine.
The failed test262 tests seem unrelated.

but ig the maintainers are busy with reviewing proposals right now XD

tkshsbcue:
@jedel1043 i hope you can review this one!

jedel1043 (approved):
Nice optimization!

PR #5025, opened 2026-03-13
fix(arraybuffer): relax alignment debug assertions in batched copy routines

mrhapile:
Relax debug assertions in the batched ArrayBuffer copy helpers to allow
Phase 1 to terminate early when the copy size is smaller than the
remaining alignment padding.

Previously the code assumed Phase 1 always reaches `BATCH_SIZE`
alignment, which is not guaranteed for small copies (e.g.
`DataView.getUint8`). This caused debug builds to panic even though the
copy operation completed correctly.

This change updates the debug assertions to allow early termination
when `head == count`.

Fixes the panic observed in the Test262 test:
`test/built-ins/SharedArrayBuffer/init-zero.js`.

Screenshot 2026-03-13 at 11 05 07 AM

### DIFF
--- a/core/engine/src/builtins/array_buffer/utils.rs
+++ b/core/engine/src/builtins/array_buffer/utils.rs
@@ -428,6 +428,16 @@ unsafe fn batched_atomic_copy_forward(src: *const AtomicU8, dest: *const AtomicU
 
     let (head, chunks, tail) = compute_batch_offsets(dest as usize, count);
 
+    if chunks == 0 {
+        // SAFETY: ensured by the caller.
+        unsafe {
+            for i in 0..count {
+                (*dest.add(i)).store((*src.add(i)).load(Ordering::Relaxed), Ordering::Relaxed);
+            }
+        }
+        return;
+    }
+
     // Phase 1: Copy unaligned head bytes until both pointers are 8-byte aligned.
     // SAFETY: ensured by the caller — both pointers are valid for `count` bytes.
     unsafe {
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/array_buffer/utils.rs
```diff
@@ -439,8 +439,14 @@ unsafe fn batched_atomic_copy_forward(src: *const AtomicU8, dest: *const AtomicU
...
-        debug_assert_eq!((dest as usize + head) % BATCH_SIZE, 0);
-        debug_assert_eq!((src as usize + head) % BATCH_SIZE, 0);
+        debug_assert!(
+            head == count || (dest as usize + head) % BATCH_SIZE == 0,
+            "dest pointer failed to reach alignment after phase 1"
+        );
+        debug_assert!(
+            head == count || (src as usize + head) % BATCH_SIZE == 0,
+            "src pointer failed to reach alignment after phase 1"
+        );
```
jedel1043: I don't think this is the correct fix. The correct fix IMO is to add a check on the second `if` for `count < BATCH_SIZE`

mrhapile: Yeah I meant the initial byte-by-byte copy that moves `src/dest `until they reach `BATCH_SIZE `alignment..... In cases like `count < required_padding` (e.g. copying just 1 byte when misalign is 6), the copy finishes before alignment is reached, which is why the debug assert was firing

mrhapile: hmmmm but your suggestion about checking `count < BATCH_SIZE` before entering the batching path makes sense 😃.......I’ll try that and see if it works better.

jedel1043: Oh I think I get what you mean. In that case the fix is not to check `count < BATCH_SIZE`, but to bail if `chunks == 0`.

jedel1043: And not exactly bail, but to just fallback to copying one by one if `chunks == 0`

mrhapile: hmmmmmm I'll try adding the fallback when `chunks == 0`

mrhapile:
@jedel1043 Added the `chunks == 0` fallback and pushed the update.🙇‍♂️

PR #5027, opened 2026-03-13
fix(string): make repeat respect loop iteration runtime limit

### LINKED ISSUE #3828: Bug on set_loop_iteration_limit on ''.repeat
**Describe the bug**
When I call set_loop_iteration_limit and then evaluate code with ''.repeat, the repeat functionality does not respect the set loop iteration limit. The code below gets stuck in a loop.

**To Reproduce**

```javascript
  let arr = [];
    while (true) {
      arr.push(' '.repeat(1024 * 1024));
    }
```

**Expected behavior**
The .repeat function should stop iterating when set_loop_iteration_limit is set, similar to how it works with normal loops.

**Build environment:**

- OS: macOS
- Version: 14.4.1 (23E224)
- Rustc version: rustc 1.77.2 (25ef9e3d8 2024-04-09)

RajdeepKushwaha5:
## Description

Fixes a runtime-limits gap in `String.prototype.repeat` where large repeat counts in native builtin looping could bypass `set_loop_iteration_limit` expectations from user code.

## Problem

Issue #3828 reports that loop iteration limits are not respected in scenarios using `''.repeat(...)`, causing effectively unbounded native work even when runtime loop limits are configured.

## Solution

- Updated `String.prototype.repeat` to enforce runtime loop iteration limits during native loop work.
- Preserved existing RangeError behavior for invalid repeat counts and overflow checks.
- Added regression coverage to ensure repeat obeys runtime loop limits.

### Updated
- `core/engine/src/builtins/string/mod.rs`
  - integrated runtime loop limit guard for `repeat`
  - returns `RuntimeLimitError::LoopIteration` when limit is exceeded

### Added tests
- `core/engine/src/builtins/string/tests.rs`
  - new test: `repeat_respects_loop_runtime_limit`

## Verification

- `cargo build -p boa_engine` ✅
- `cargo test -p boa_engine repeat --lib` ✅
- `cargo test -p boa_engine loop_runtime_limit --lib` ✅
- `cargo test -p boa_engine --lib -- --quiet` ✅ (`1034 passed, 0 failed`)

## Related issue

Fixes #3828

### DIFF
--- a/core/engine/src/builtins/string/tests.rs
+++ b/core/engine/src/builtins/string/tests.rs
@@ -154,9 +154,8 @@ fn repeat_respects_loop_runtime_limit() {
         TestAction::inspect_context(|context| {
             context.runtime_limits_mut().set_loop_iteration_limit(10);
         }),
+        TestAction::assert_runtime_limit_error("'x'.repeat(11)", RuntimeLimitError::LoopIteration),
         TestAction::assert_eq("'x'.repeat(10)", js_str!("xxxxxxxxxx")),
-        TestAction::assert_eq("'x'.repeat(11)", js_str!("xxxxxxxxxxx")),
-        TestAction::assert_runtime_limit_error("'x'.repeat(12)", RuntimeLimitError::LoopIteration),
     ]);
 }
 
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -737,15 +737,14 @@ impl String {
             return Ok(js_string!().into());
         }
 
-        // Mirror VM `IncrementLoopIteration` semantics for a bulk increment.
+        // Runtime limit guard for native loops inside this builtin.
         let previous_iteration_count = context.vm.frame().loop_iteration_count;
         let max_iteration_count = context.vm.runtime_limits.loop_iteration_limit();
+        let loop_iteration_count = previous_iteration_count.saturating_add(n);
 
-        if previous_iteration_count.saturating_add(n - 1) > max_iteration_count {
+        if loop_iteration_count > max_iteration_count {
             return Err(RuntimeLimitError::LoopIteration.into());
         }
-
-        let loop_iteration_count = previous_iteration_count.saturating_add(n);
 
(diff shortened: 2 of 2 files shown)


RajdeepKushwaha5:
Addressed both review comments in follow-up commit 66eb9537.

### Changes made
- Aligned String.prototype.repeat native loop-limit guard with VM IncrementLoopIteration semantics.
- Updated repeat runtime-limit regression boundaries accordingly.

### Rationale
VM loop-limit currently throws when previous_iteration_count > max before increment, which allows up to max + 1 increments.  
The previous repeat guard was stricter by one iteration.  
This patch keeps repeat behavior consistent with current VM semantics.

### Updated test expectations (limit = 10)
- 'x'.repeat(10) -> pass
- 'x'.repeat(11) -> pass
- 'x'.repeat(12) -> RuntimeLimitError::LoopIteration

### Verification
- cargo test -p boa_engine repeat_respects_loop_runtime_limit --lib
- cargo test -p boa_engine loop_runtime_limit --lib
- cargo test -p boa_engine repeat --lib

core/engine/src/builtins/string/tests.rs
```diff
@@ -144,6 +148,18 @@ fn repeat_generic() {
...
 
+#[test]
+fn repeat_respects_loop_runtime_limit() {
+    run_test_actions([
+        TestAction::inspect_context(|context| {
+            context.runtime_limits_mut().set_loop_iteration_limit(10);
+        }),
+        TestAction::assert_eq("'x'.repeat(10)", js_str!("xxxxxxxxxx")),
+        TestAction::assert_eq("'x'.repeat(11)", js_str!("xxxxxxxxxxx")),
+        TestAction::assert_runtime_limit_error("'x'.repeat(12)", RuntimeLimitError::LoopIteration),
```
jedel1043: It was more correct as it was before.

`'x'.repeat(11)` should fail when the loop iteration limit is 10, because `'x'.repeat(1)` just returns the same string, so it doesn't contribute to the loop limit, meaning `'x'.repeat(11)` will only "loop" 10 times in total.

PR #5030, opened 2026-03-13
fix: check extensibility in DefinePrivateField opcode

xcb3d:
This Pull Request fixes the DefinePrivateField VM opcode to comply with the nonextensible-applies-to-private proposal.

It changes the following:

- DefinePrivateField now calls private_field_add() instead of append_private_element(), properly checking IsExtensible(O) per the PrivateFieldAdd spec algorithm


jedel1043 (approved):
Cool, thanks!

Not gonna ask for gating against the `experimental` feature since it's such a small change in behaviour that it doesn't seem worth gating.

Issue #4350, opened 2025-07-24
Inconsistent function vs object assignment in indirect eval


ThinerDAS:
## Description
When executing an indirect eval call (`(1,eval)`) that contains both a function declaration and a subsequent variable assignment with the same name, Boa exhibits inconsistent behavior that may not conform to ECMAScript standards.

## Reproduction Code
```javascript
t = []

var s1 = `
function core() { t.push(1) }

core.prototype.a = function () { t.push(2) }
core.prototype.b = function () { t.push(3) }
`;
var s2 = `
function core() { t.push(1) }

core.prototype.a = function () { t.push(2) }
core.prototype.b = function () { t.push(3) }
var core = new core();
`;
var s3 = `
function core() { t.push(1) }
var core = new core();
`;

function run_ctx(s) {
    (1,eval)(s);
}

function test() {
    run_ctx(s1);
    var core1 = new core()
  
    run_ctx(s2);
    var core2 = core
  
    run_ctx(s3);
    var core3 = core
    return [core1,core2,core3]
}

test()
```

## Expected Behavior
According to ECMAScript standards, when a function declaration is followed by a variable assignment that overwrites it with an instance of itself (e.g., `var core = new core()`), the final global binding should be the object instance, not the function.

All three elements in the returned array should be object instances.

And in most of my javascript environments, it is like:
```
"[object Object],[object Object],[object Object]"
```

### v0.20.0
```
"[object Object],function () { [native code] },[object Object]"
```

### git master
```
"[object Object],function () { t.push(3) },[object Object]"
```

The second element remains a function instead of becoming an object instance as expected.

## Questions
- Which behavior is conformant with the ECMAScript standard?
- Is this a bug or a specification edge case?

## Environment
- Boa version: 0.20.0 and git master
- Platform: rust 1.88

## Rust source code

```rust
use boa_engine::{Context, Source, JsValue, JsError, js_string};
use std::{env, fs, path::Path, error::Error};

fn main() -> Result<(), Box<dyn Error>> {
    let args: Vec<String> = env::args().skip(1).collect();
    if args.is_empty() {
        eprintln!("Usage: {} <file1.js> [file2.js ...]", env::args().next().unwrap());
        return Ok(());
    }

    let mut context = Context::default();
    let mut last_result = None;

    for file in args {
        let js_code = fs::read_to_string(&file)?;
        match context.eval(Source::from_bytes(&js_code).with_path(Path::new(&file))) {
            Ok(result) => {
[...]

PR #5031, opened 2026-03-13
Fix stale return value leakage in indirect eval constructor binding (#4350)

RajdeepKushwaha5:
# PR Description

## Summary
This PR fixes issue #4350, where indirect eval could leave `core` bound to a function instead of the object instance after:
- function declaration
- prototype assignments
- `var core = new core();`

In the failing case, the second value became a function (for example `function () { t.push(3) }`) instead of an object.

## Root Cause
A stale VM return accumulator value could leak into ordinary call/construct frame execution in specific eval completion paths.  
That stale value could then affect constructor return handling and produce an incorrect final binding value.

## Fix
- Reset the VM return accumulator to `undefined` when entering ordinary function call frames.
- Reset the VM return accumulator to `undefined` when entering ordinary function construct frames.
- Add a regression test that reproduces #4350 exactly.

## Validation
Passed:
1. `cargo test -p boa_engine indirect_eval_function_var_binding_4350 -- --nocapture`
2. `cargo test -p boa_engine empty_return_values -- --nocapture`
3. `cargo test -p boa_engine tests::env:: -- --nocapture`

## Result
The repro now evaluates to:
`[object Object],[object Object],[object Object]`

instead of returning a function in the second position.

### DIFF
--- a/core/engine/src/builtins/function/mod.rs
+++ b/core/engine/src/builtins/function/mod.rs
@@ -1028,6 +1028,7 @@ pub(crate) fn function_call(
     }
 
     context.vm.push_frame(frame);
+    context.vm.set_return_value(JsValue::undefined());
 
     let context = context.context();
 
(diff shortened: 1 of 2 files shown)


RajdeepKushwaha5:
> Nice catch!
Thanks for the review and approval!

PR #5035, opened 2026-03-13
feat: implement import.defer() and import.source() dynamic import syntax

xcb3d:
This Pull Request adds parsing and runtime support for `import.defer()` and `import.source()` dynamic import syntax, enabling 64 previously-ignored test262 tests.

It changes the following:

- Added `ImportPhase` enum (`Evaluation`, `Defer`, `Source`) and a `phase` field to the `ImportCall` AST node in `core/ast/src/expression/call.rs`, with proper `ToInternedString` output for each phase.
- Extended the parser in `core/parser/src/parser/expression/left_hand_side/mod.rs` to detect and parse `import.defer(expr)` and `import.source(expr)` syntax, including 4-token lookahead in `is_keyword_call` and phase-aware token consumption.
- Updated the bytecode compiler in `core/engine/src/bytecompiler/expression/mod.rs` to encode `ImportPhase` as an `IndexOperand` (0=evaluation, 1=defer, 2=source) on the `ImportCall` instruction.
- Added `phase: IndexOperand` to the `ImportCall` opcode definition in `core/engine/src/vm/opcode/mod.rs` and updated the code block display in `core/engine/src/vm/code_block.rs` to show phase names.
- Updated the VM `ImportCall` handler in `core/engine/src/vm/opcode/call/mod.rs` to decode the phase operand: source phase rejects with `SyntaxError` per `GetModuleSource()` spec (16.2.1.7.2), defer phase uses standard evaluation semantics.
- Removed `import-defer` and `source-phase-imports` from the ignored features list in `test262_config.toml`, bringing `dynamic-import/catch` suite from 112/176 to **176/176 (100% conformance)**.

### DIFF
--- a/core/engine/src/vm/opcode/call/mod.rs
+++ b/core/engine/src/vm/opcode/call/mod.rs
@@ -345,6 +345,7 @@ async fn load_dyn_import(
     referrer: Referrer,
     request: ModuleRequest,
     cap: PromiseCapability,
+    phase: u32,
     context: &RefCell<&mut Context>,
 ) -> JsResult<()> {
     let loader = context.borrow().module_loader();
@@ -409,6 +410,36 @@ async fn load_dyn_import(
                 .or_insert_with(|| module.clone());
             debug_assert_eq!(&module, entry);
         }
+    }
+
+    // When the `experimental` feature is disabled, reject any non-evaluation phase.
+    #[cfg(not(feature = "experimental"))]
+    if phase != 0 {
+        let err = JsNativeError::syntax()
+            .with_message("import.defer() and import.source() require the 'experimental' feature")
+            .into();
+        let err = JsError::into_opaque(err, &mut context.borrow_mut())?;
+        cap.reject()
+            .call(&JsValue::undefined(), &[err], &mut context.borrow_mut())
+            .expect("default `reject` function cannot throw");
+        return Ok(());
+    }
+
+    // TODO: For source phase (phase == 2), implement GetModuleSource()
+    // 16.2.1.7.2 GetModuleSource ( )
+    // Source Text Module Record provides a GetModuleSource implementation
+    // that always returns an abrupt completion indicating that a source phase import is not available.
+    // 1. Throw a SyntaxError exception.
+    #[cfg(feature = "experimental")]
+    if phase == 2 {
+        let err = JsNativeError::syntax()
+            .with_message("source phase import is not available for this module")
+            .into();
+        let err = JsError::into_opaque(err, &mut context.borrow_mut())?;
+        cap.reject()
+            .call(&JsValue::undefined(), &[err], &mut context.borrow_mut())
+            .expect("default `reject` function cannot throw");
+        return Ok(());
     }
 
     // 2. Let module be moduleCompletion.[[Value]].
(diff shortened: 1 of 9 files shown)


core/parser/src/parser/expression/left_hand_side/mod.rs
```diff
@@ -123,6 +144,43 @@ where
...
+                            cursor.advance(interner);
+                            phase
+                        } else {
+                            ImportPhase::Evaluation
+                        }
+                    } else {
+                        ImportPhase::Evaluation
+                    }
+                } else {
+                    ImportPhase::Evaluation
```
jedel1043: Is this spec compliant? I would be very surprised if the spec just made it so that `import.thing()` would fallback to evaluation.

core/engine/src/vm/opcode/call/mod.rs
```diff
@@ -411,6 +412,22 @@ async fn load_dyn_import(
         }
     }
 
+    // For source phase (phase == 2), implement GetModuleSource()
```
jedel1043:
```suggestion
    // TODO: For source phase (phase == 2), implement GetModuleSource()
```

core/engine/src/vm/opcode/call/mod.rs
jedel1043, line 1: All of the new behaviour kinda needs to be gated behind the `experimental` feature, since their corresponding proposals are still not stage 4.

My recommendation is to unconditionally throw if `phase != 0` when the feature is disabled.

jedel1043 (changes requested):

core/parser/src/parser/expression/left_hand_side/mod.rs
```diff
@@ -107,6 +107,27 @@ where
...
+                        && let Some(dot) = cursor.peek(1, interner)?
+                        && dot.kind() == &TokenKind::Punctuator(Punctuator::Dot)
+                        && let Some(ident_tok) = cursor.peek(2, interner)?
+                    {
+                        let is_phase_ident = matches!(
+                            ident_tok.kind(),
+                            TokenKind::IdentifierName((sym, _))
+                                if {
+                                    let s = interner.resolve_expect(*sym).utf8().unwrap_or("");
+                                    s == "defer" || s == "source"
```
jedel1043: Instead of having to fetch the strings here, we would just add "defer" and "source" to the list of static `Sym`s, so that you can directly match against those.

core/parser/src/parser/expression/left_hand_side/mod.rs
```diff
@@ -107,6 +107,27 @@ where
...
+                                }
+                        );
+                        if is_phase_ident
+                            && let Some(paren) = cursor.peek(3, interner)?
+                            && paren.kind() == &TokenKind::Punctuator(Punctuator::OpenParen)
+                        {
+                            return Ok(Some(keyword_token_start));
+                        }
+                    }
```
jedel1043: You're doing all of this manual handling here and again below. To keep this a bit cleaner, I would suggest not depending in `is_keyword_call` to check for `import.defer` and `import.source` calls, and doing those checks directly in the let-else matches below.

jedel1043 (changes requested):

jedel1043 (approved):
Yep, looks great! Thank you!

PR #5038, opened 2026-03-13
perf(intl): cache resolvedOptions result on DateTimeFormat

Lokesh0224:
Intl.DateTimeFormat.prototype.resolvedOptions() previously rebuilt a new JS object on every call, re-allocating strings for locale, calendar, numberingSystem, timeZone, hourCycle, dateStyle, timeStyle each time, even though none of these values can change after construction.
This PR caches the result on first call and returns the cached JsObject on subsequent calls.

**Before cached:**
Screenshot from 2026-03-13 16-08-40

**After cached: (~10x faster)**
image


jedel1043 (approved):
Nice optimization!

PR #5039, opened 2026-03-13
Implement URL.searchParams and URLSearchParams

### LINKED ISSUE #4743: Implement URLSearchParams class

Monti-27:
This Pull Request fixes/closes #4743.

It changes the following:
- adds `URLSearchParams` and implements `URL.searchParams`
- keeps `URL.searchParams` live and cached, including mutation during iteration and `forEach()`
- re-enables the excluded `url-searchparams.any.js` WPT case and adds focused runtime tests

Test plan:
- `cargo test --manifest-path tests/wpt/Cargo.toml url -- --nocapture`

### DIFF
--- a/core/runtime/src/url.rs
+++ b/core/runtime/src/url.rs
@@ -333,9 +333,6 @@ impl UrlSearchParamsIterator {
 impl UrlSearchParams {
     #[boa(constructor)]
     fn constructor(init: JsValue, context: &mut Context) -> JsResult<Self> {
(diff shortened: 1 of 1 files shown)


Monti-27:
Superseded by #5040, which contains the same change as a single squashed commit.

PR #5040, opened 2026-03-13
add URLSearchParams support

### LINKED ISSUE #4743: Implement URLSearchParams class

Monti-27:
Fixes #4743.

Adds `URLSearchParams` and wires up `URL.searchParams`.

This also fixes the constructor path for iterable inputs, handles explicit `undefined` correctly in the two-argument `has()` and `delete()` overloads, adds runtime coverage, and re-enables the skipped URL WPT case.

Tests:
- `cargo test --manifest-path tests/wpt/Cargo.toml url -- --nocapture`

### DIFF
--- a/core/runtime/src/url.rs
+++ b/core/runtime/src/url.rs
@@ -18,7 +18,7 @@
 #[cfg(test)]
 mod tests;
 
-use boa_engine::builtins::iterable::create_iter_result_object;
+use boa_engine::builtins::iterable::{IteratorRecord, create_iter_result_object};
 use boa_engine::builtins::object::OrdinaryObject;
 use boa_engine::class::Class;
 use boa_engine::interop::{JsClass, TryFromJsArgument};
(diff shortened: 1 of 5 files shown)


Monti-27:
@jedel1043  can you please review this 
thank you !!

core/runtime/src/url.rs
```diff
@@ -68,149 +724,152 @@ impl Url {
...
-    fn search_params() -> JsResult<()> {
-        Err(js_error!(Error: "URL.searchParams is not implemented"))
+    fn search_params(this: JsClass<Self>, context: &mut Context) -> JsResult<JsValue> {
+        if let Some(existing) = this.borrow().search_params.clone() {
+            return Ok(existing.into());
+        }
+
+        let params = UrlSearchParams::from_url(this.inner(), context)?;
+        this.borrow_mut().search_params = Some(params.clone());
+        Ok(params.into())
```
jedel1043: I don't think we need to lazily initialize this, right? According to the spec, we can directly initialize in `Url`'s constructor.

jedel1043 (changes requested):

core/runtime/src/url.rs
Monti-27: yeah you right, no real need for lazy init here. i’m updating it so searchParams gets created in the URL constructor itself and the getter just returns the cached object

Monti-27:
@jedel1043  updated this. searchParams is now created during URL construction and the getter just returns the same object.

core/runtime/src/url.rs
```diff
@@ -14,18 +14,671 @@
...
+    }
+
+    let Some(method) = method.as_object().filter(JsObject::is_callable) else {
+        return Err(js_error!(
+            TypeError: "value returned for property of object is not a function"
+        ));
+    };
+
+    Ok(Some(method.clone()))
+}
```
jedel1043: This [already exists](https://github.com/boa-dev/boa/blob/0698e7063866a1b40be043a1089be43d3224db4b/core/engine/src/object/operations.rs#L797-L824), so maybe just expose it as a public API

core/runtime/src/url.rs
```diff
@@ -14,18 +14,671 @@
...
+        _: &'a JsValue,
+        rest: &'a [JsValue],
+        _: &mut Context,
+    ) -> JsResult<(Self, &'a [JsValue])> {
+        match rest.split_first() {
+            Some((first, rest)) => Ok((Self(Some(first.clone())), rest)),
+            None => Ok((Self(None), rest)),
+        }
+    }
+}
```
jedel1043, line 181: Wondering about this implementation. Was `Option<JsValue>` not enough for this? It will additionally automatically map `undefined` to `None`, so if that's the correct behaviour it would be better to use `Option`

core/runtime/src/url.rs
jedel1043, line 1: Can you document this file with the specific sections of the [spec](https://url.spec.whatwg.org/#urlsearchparams) that are covered? We generally try to mirror each line of code with one or more lines from the spec

core/runtime/src/url.rs
```diff
@@ -64,153 +748,148 @@ impl Url {
...
     #[boa(rename = "search")]
     fn set_search(&mut self, value: Convert<String>) {
-        url::quirks::set_search(&mut self.0, &value.0);
+        url::quirks::set_search(&mut self.inner.borrow_mut(), &value.0);
     }
 
     #[boa(getter)]
-    fn search_params() -> JsResult<()> {
-        Err(js_error!(Error: "URL.searchParams is not implemented"))
+    fn search_params(&self) -> JsValue {
```
jedel1043:
```suggestion
    fn search_params(&self) -> JsObject {
```

jedel1043: The macro automatically converts from `JsObject` to `JsValue`

jedel1043 (changes requested):

core/runtime/src/url.rs
Monti-27: thnx switched this over to the existing get_method instead of keeping a duplicate helper here

core/runtime/src/url.rs
Monti-27: kept this one as is Option<JsValue> would treat an explicit undefined the same as a missing second argument her and for delete/has we need to keep those two cases separate

core/runtime/src/url.rs
Monti-27: yep added the relevant spec refs in the file and around the URLSearchParams constructor / methods that this patch is covering

core/runtime/src/url.rs
Monti-27: yep changed this to JsObject

Monti-27:
give me 5 mins I'll clutch this :))

Monti-27:
also fixed the clippy failure from the missing errors docs on the Rust side URL constructors

core/runtime/src/url.rs
```diff
@@ -7,25 +7,679 @@
...
+#[derive(Debug, JsData, Trace, Finalize)]
+struct UrlSearchParamsIterator {
+    search_params: JsObject<UrlSearchParams>,
+    next_index: usize,
+    #[unsafe_ignore_trace]
+    kind: UrlSearchParamsIteratorKind,
+    done: bool,
+}
+
+impl UrlSearchParamsIterator {
```
jedel1043, line 360: This could also be a class IMO

Monti-27: yeah i thought about that too. i kept it as an internal iterator object since its only created from URLSearchParams and doesn’t need separate registration but i can switch it to a class if you’d prefer that here totally your take

core/runtime/src/url.rs
```diff
@@ -7,25 +7,679 @@
...
+                TypeError: "value returned for property of object is not a function"
+            ));
+        };
+
+        Ok(Self {
+            iterator: iterator.clone(),
+            next: next.clone(),
+            done: false,
+        })
+    }
```
jedel1043: This looks like something we've already implemented in `boa_engine`. Maybe we should expose that instead of reimplementing it here?

jedel1043 (changes requested):

core/runtime/src/url.rs
Monti-27: Switched this to reuse boa_engine iterator support and removed the local runtime iterator implementation

Monti-27:
@hansl your thoughts on this

tomasol:
I find this feature very useful. Are there any blockers or can it be merged?

Monti-27:
@jedel1043 PTAL

hansl:
This LGTM, @jedel1043 should also do a last pass of review.

hansl (approved):

jedel1043 (approved):
LGTM

Monti-27:
this break after v0.22 was released. I’m updating Url to preserve its existing traits, then I’ll rerun the checks

PR #5046, opened 2026-03-13
feat(runtime): add Response.redirect() and Response.json()

### LINKED ISSUE #5042: Add Response.redirect() and Response.json() static constructors to fetch runtime
While working in `core/runtime/src/fetch/response.rs`, I noticed that two static constructors defined in the Fetch spec are currently missing from the runtime implementation.

`Response.error()` already exists and is exposed as a static constructor, but the following methods are not available yet:

- `Response.redirect(url, status)`
- `Response.json(data, init)`

Both are part of the standard Fetch API and implemented in other runtimes. Adding them would extend the existing `JsResponse` implementation without introducing new modules or touching the ongoing TC55 work.

### `Response.redirect(url, status)`

Creates a redirect response and sets the `Location` header.

Expected behavior:

- `status` must be one of `301`, `302`, `303`, `307`, `308`
- default status should be `302`
- throw `RangeError` for invalid redirect status
- construct a `Response` with empty body and a `Location` header

Spec reference  
https://fetch.spec.whatwg.org/#dom-response-redirect

### `Response.json(data, init)`

Creates a `Response` whose body is the JSON serialization of `data`.

Expected behavior:

- serialize `data` using the same JSON conversion path used by the instance `response.json()` method
- apply optional `init` fields such as `status` and `headers`
- set `Content-Type: application/json` if it is not already present

Spec reference  
https://fetch.spec.whatwg.org/#dom-response-json

## Implementation location

Both methods can be added as `#[boa(static)]` methods inside the existing `#[boa_class]` implementation for `JsResponse` in:

core/runtime/src/fetch/response.rs

This keeps the change localized to the existing fetch runtime code.

## Tests

Tests would be added under:

core/runtime/src/fetch/tests/response.rs

Proposed cases:

- `Response.redirect(url)` returns status `302` with a `Location` header
- `Response.redirect(url, 301)` uses the provided redirect status
- invalid redirect status throws `RangeError`
- `Response.json(data)` creates a response with JSON body and correct `Content-Type`
- `Response.json(data, { status })` respects the provided status
- `Response.json(data, { headers })` does not overwrite an existing `Content-Type`

If this looks reasonable, I can open a PR implementing both methods.

iammdzaidalam:
Resolves https://github.com/boa-dev/boa/issues/5042

This adds `Response.redirect()` and `Response.json()` to the fetch runtime.

`Response.error()` was already exposed as a static constructor, but the other two spec-defined constructors were still missing.

The change is localized to `core/runtime/src/fetch/response.rs` with the corresponding tests in `core/runtime/src/fetch/tests/response.rs`.

## Added

- `Response.redirect(url, status)`
  - defaults to `302`
  - validates redirect status
  - sets the `Location` header

- `Response.json(data, init)`
  - serializes the provided value as JSON
  - applies optional response init fields
  - sets `Content-Type: application/json` if it is not already present

## Tests

Added coverage for:

- default redirect status
- custom redirect status
- invalid redirect status
- JSON response body and content type
- preserving explicit `Content-Type`
- applying `init.status`

Spec:
- https://fetch.spec.whatwg.org/#dom-response-redirect
- https://fetch.spec.whatwg.org/#dom-response-json


jedel1043 (approved):
Looks good!

jedel1043:
Ping me after resolving the conflicts so that I can merge

core/runtime/src/fetch/tests/response.rs
```diff
@@ -179,6 +248,8 @@ fn response_headers_get_combines_duplicate_values_with_comma_space() {
             "#,
         ),
         TestAction::inspect_context(|ctx| {
+            let p = ctx.global_object().get(js_str!("p"), ctx).unwrap();
+            p.as_promise().unwrap().await_blocking(ctx).unwrap();
```
jedel1043: Uhh something got kinda borked in here? You have a Rust test inside a string with JS code

iammdzaidalam: Ok wait let me check

jedel1043: Maybe this was a consequence of the rebase? Seems like the test somehow merged with the JS string code instead of being a separate test

iammdzaidalam: Ah yes, my bad. Bad conflict resolution here. Part of the Rust test block got shoved inside the JS string. Fixing it now and will push the corrected patch.

iammdzaidalam: Hi @jedel1043. Fixed the issue, pushed the corrected tests.

PR #5047, opened 2026-03-13
perf(vm): new tailcalling dispatch behind feature `tailcall` (nightly required)

### LINKED ISSUE #4858: idea: opcode dispatch performance improvements
Just started looking into boa's VM. It appears that boa uses call threading for opcode dispatch via function pointers:
https://github.com/boa-dev/boa/blob/75c0a75d902172baf9ac7c28ab067e3997801a5c/core/engine/src/vm/mod.rs#L876-L891

There should be room for measurable improvements according to https://github.com/tipo159/rust-instruction-dispatch .
Running the experiments with rustc 1.93.0 on my machine (Lunar Lake, Q3 2024):
| Approach | Average of 100 results (ns) |
|--------|--------|
|direct-call-threading|2_947_553|
|switch-dispatch|2_469_013 |
|direct-tail-call-threading|1_594_078|

While tipo159/rust-instruction-dispatch does have some implementation inconsistencies (which perhaps caused switch-dispatch to be better than direct-call-threading), it would appear that direct-tail-call-threading is more superior and does not require writing ASM by hand either.

zhuzhu81998:
Closes #4858 .
I had to move the big opcode list around a bit hence the large number of diffs.
Specifically, had to split the opcode handlers generation into a separate macros so feature-conditional compilation works. And to avoid repeating the gigantic opcode list twice (for tailcall and non-tailcall handlers), i put the list into a helper macros `for_each_opcode`.

Budgeted handlers would still use the standard loop-based approach.
Tailcalling handlers support tracing as well.

On main:
```
PROGRESS Richards
RESULT Richards 407
PROGRESS DeltaBlue
RESULT DeltaBlue 359
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 350
PROGRESS RayTrace
RESULT RayTrace 623
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 785
PROGRESS RegExp
RESULT RegExp 104
PROGRESS Splay
RESULT Splay 1376
PROGRESS NavierStokes
RESULT NavierStokes 849
SCORE 484
```

With this PR and tracing enabled (`cargo run --release --bin boa --features tailcall -- ../data/bench/bench-v8/combined.js`):
```
PROGRESS Richards
RESULT Richards 459
PROGRESS DeltaBlue
RESULT DeltaBlue 390
PROGRESS Encrypt
PROGRESS Decrypt
RESULT Crypto 377
PROGRESS RayTrace
RESULT RayTrace 669
PROGRESS Earley
PROGRESS Boyer
RESULT EarleyBoyer 844
PROGRESS RegExp
RESULT RegExp 104
PROGRESS Splay
RESULT Splay 1479
PROGRESS NavierStokes
RESULT NavierStokes 971
SCORE 524
```

TODO:
 ~~- [ ] budgeted dispatch still use old call threading right now.~~ (async interrupt does not play very well with tailcalling dispatch)
 - [x] tracing is not exactly integrated right now

### DIFF
--- a/.github/workflows/pr_management.yml
+++ b/.github/workflows/pr_management.yml
@@ -16,12 +16,21 @@ jobs:
         uses: actions/github-script@v8
         with:
           script: |
-            github.rest.issues.addLabels({
-              issue_number: context.issue.number,
+            const labels = await github.rest.issues.listLabelsOnIssue({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              labels: ['Waiting On Review']
-            })
+              issue_number: context.issue.number
+            });
+
+            if (labels.data.every(label => label.name != "Waiting On Author")) {
+              github.rest.issues.addLabels({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                labels: ['Waiting On Review']
+              })
+            }
+
 
       - name: Auto Remove Label
         if: github.event.action == 'closed'
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2563,9 +2563,9 @@ checksum = "5e5032e24019045c762d3c0f28f5b6b8bbf38563a65908389bf7978758920897"
 
 [[package]]
 name = "lz4_flex"
-version = "0.12.0"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab6473172471198271ff72e9379150e9dfd70d8e533e0752a27e515b48dd375e"
+checksum = "db9a0d582c2874f68138a16ce1867e0ffde6c0bb0a0df85e1f36d04146db488a"
 dependencies = [
  "twox-hash",
 ]
@@ -2752,9 +2752,9 @@ dependencies = [
 
 [[package]]
 name = "num_enum"
-version = "0.7.5"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1207a7e20ad57b847bbddc6776b968420d38292bbfe2089accff5e19e82454c"
+checksum = "5d0bca838442ec211fa11de3a8b0e0e8f3a4522575b5c4c06ed722e005036f26"
 dependencies = [
  "num_enum_derive",
  "rustversion",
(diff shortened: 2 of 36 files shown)


core/engine/src/vm/opcode/tailcall.rs
```diff
@@ -0,0 +1,107 @@
...
+                    context.vm.frame_mut().pc = next_pc as u32;
+                    let result = $Variant::operation(args, context);
+
+                    let cr = IntoCompletionRecord::into_completion_record(result, context);
+
+                    // This match MUST be the last expression — both arms in tail position
+                    match cr {
+                        ControlFlow::Continue(()) => become context.dispatch_next(context.vm.frame().pc as usize),
+                        ControlFlow::Break(value) => value,
+                    }
```
jedel1043, line 100: I wonder... If you replace this with
```rust
If let ControlFlow::Break(value) = cr {
    return value
}

context.dispatch_next(context.vm.frame().pc as usize)
```
does the compiler tail-call optimize it without having to use `become`?

zhuzhu81998: the problem is, if you dont use become, there is no guarantee.

like, technically, llvm would in practice if optimization is turned on (not in debug build tho).

jedel1043: Oh interesting that using this method in debug builds would immediately cause a stack overflow. Makes sense why we need `become` then

zhuzhu81998: yea like `cargo test` would already fail hard XD

zhuzhu81998:
So there was a noticeable performance drop when i added the tracing part to `dispatch_next` in 2bab9b7b6017fd1b392ad7bdaa564f25e05007eb . The v8_bench score dropped from 550 to 520, even though tracing was not enabled so the path was not actually taken. And the performance is bit too big to be explained by a condition check, especially since it should be easily predictable by the cpu.

I think it might be related to how `preserve_none` works, but my experimental approach of splitting the traced path into a separate function did not really help.

It is also possible that maybe my 550 benchmark was an outlier 😂 . Will investigate a bit more.

PR #5049, opened 2026-03-13
fix(ast): prevent ContainsVisitor from incorrectly traversing ClassElement

RajdeepKushwaha5:
### Description
This PR fixes an overarching scope bug where `ContainsVisitor` would mistakenly traverse into `ClassElement`s (such as static and instance class fields) when searching the AST for instances of `eval`. This caused the engine to improperly flag an outer scope as containing an `eval` token based on the inner context of an embedded class definition.

By explicitly overriding the `visit_class_element` method in the `ContainsVisitor` to halt further downward traversal across those scope boundaries, the AST execution now correctly preserves scope isolation, fixing panics and parse failures related to `eval` expressions within static class fields.

### Why is this needed?
The ECMAScript specification relies on analyzing scopes to determine if a node contains an `eval` call. Because inner class fields and methods construct their own separate environments, they must be partitioned out of the parent node's search context for `ContainsSymbol::Eval` evaluations.

### Changes Made
- Added an empty `visit_class_element` override within `core/ast/src/visitor/contains.rs` to stop AST traversal from piercing class element bounds.
- Checked against `test262` test cases to ensure correct isolation of nested `eval` statements.

### DIFF
--- a/core/ast/src/operations/mod.rs
+++ b/core/ast/src/operations/mod.rs
@@ -233,8 +233,31 @@ where
                     }
                 }
                 ClassElement::FieldDefinition(field)
-                | ClassElement::StaticFieldDefinition(field) => field.name.visit_with(self),
-                _ => ControlFlow::Continue(()),
+                | ClassElement::StaticFieldDefinition(field) => {
+                    field.name.visit_with(self)?;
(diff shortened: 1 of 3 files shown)


jedel1043:
_Sigh_

PR #5055, opened 2026-03-14
refactor(engine): remove expect()-based panics from JsRegExp wrapper

### LINKED ISSUE #4818: refactor(engine): remove expect()-based panics from JsRegExp wrapper
While auditing the core engine's error handling, I found that the JsRegExp convenience wrapper relies on .expect() for internal invariants that are not strictly enforced by the type system. These can lead to host process crashes if a regression occurs in the engine.
Locations:
boa/core/engine/src/object/builtins/jsregexp.rs: Around L59–L68 (in JsRegExp::new).
boa/core/engine/src/object/builtins/jsregexp.rs: Boolean accessors (e.g., global, multiline) around L83–L108.
Proposed Fix:
Replace .expect() in JsRegExp::new with proper error propagation returning JsNativeError::internal().
Use ToBoolean semantics or explicit type checks instead of .as_boolean().expect() in accessors.

RajdeepKushwaha5:
# Description

This PR resolves issue #4818 by removing the panicking `.js_expect()` calls located internally in the `core/engine/src/object/builtins/jsregexp.rs` wrapper API. 

Previously, `JsRegExp` heavily relied on `.js_expect()` for internal invariants (such as assuming that string or primitive evaluation from a regular expression property will always succeed). This meant if the spec-compliant getters threw an expected error (such as a memory exception, a side effect from proxies, or an overridden getter), it would panic the entire execution context. 

We now use `JsValue::to_boolean()`, `JsValue::to_string(context)`, and `JsResult` error propagations across `JsRegExp::new`, `.has_indices()`, `.global()`, `.ignore_case()`, `.multiline()`, `.dot_all()`, `.unicode()`, `.sticky()`, `.flags()`, etc. Instead of panicking the host rust engine, this bubbles up the corresponding execution errors gracefully to Javascript. 

Resolves #4818 

## Changes Made
- Removed `js_expect` entirely from `core/engine/src/object/builtins/jsregexp.rs`.
- Leveraged `.to_boolean()` and `.to_string(context)` to fetch regex config and flag descriptors securely.
- Propagated expected failures as standard `JsNativeError` and `JsResult` objects via the `?` operator.

### DIFF
--- a/core/engine/src/module/loader/mod.rs
+++ b/core/engine/src/module/loader/mod.rs
@@ -62,7 +62,10 @@ pub fn resolve_module_specifier(
 
     // On Windows, also replace `/` with `\`. JavaScript imports use `/` as path separator.
     #[cfg(target_family = "windows")]
-    let specifier = specifier.replace('/', "\\");
+    let specifier = {
+        use cow_utils::CowUtils;
+        specifier.cow_replace('/', "\\").into_owned()
+    };
 
     let short_path = Path::new(&specifier);
 
--- a/core/engine/src/object/builtins/jsregexp.rs
+++ b/core/engine/src/object/builtins/jsregexp.rs
@@ -1,8 +1,7 @@
 //! A Rust API wrapper for Boa's `RegExp` Builtin ECMAScript Object
 use crate::{
-    Context, JsExpect, JsNativeError, JsResult, JsValue,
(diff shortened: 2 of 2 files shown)


jedel1043:
`js_expect` _already_ returns an error, but it also signals that there's a possible bug in the engine, so we don't want to get rid of the engine panic.

PR #5058, opened 2026-03-14
Improve VM instruction tracing with call depth and clearer output

Maanvi212006:
## 📝 Description
This PR introduces an instruction-level tracing feature in the Boa JavaScript engine VM.  
It allows developers to see each opcode execution, operands, stack state, and execution time.  
Additionally, call/return boundaries are highlighted for better visibility of function calls.

## Example Output :
Time          Depth   PC     Opcode                     Operands                   Stack
48μs          0       pc=0   CanDeclareGlobalFunction   index:0, dst:0             [ undefined, true, null, undefined |0| ]
...
--- entering call frame ---
108μs         1       pc=52  Call                       argument_count:2           [ undefined, undefined, 2, 1, [function], [object] |1| ... ]
...
--- leaving call frame ---
33μs          0       pc=39  Return                                              [ 3, 2, [function], null, undefined |0| ]

## 🔧 Implementation Details
- Added `trace_execute_instruction` method in `vm::mod.rs`.
- Captures:
  - `opcode` being executed
  - `operands` of the instruction
  - current `pc` (program counter)
  - execution time in microseconds
  - stack snapshot
  - call frame depth
- Marks boundaries for function calls with:
  - `--- entering call frame ---`
  - `--- leaving call frame ---`
- Formatting aligns columns for Time, Depth, PC, Opcode, Operands, and Stack.

## Motivation
This change improves visibility into the execution of JavaScript code inside Boa, which will be useful for:
- Debugging complex functions and nested calls.
- Understanding bytecode execution flow.
- Helping contributors follow Boa’s VM internals.

## 🖼️ Screenshots
boa pr 1

## Checklist
- [x] Tested locally with cargo run --features trace -- --trace.
- [x] Formatted with cargo fmt.

- [x] Verified output shows depth and stack correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -483,6 +483,7 @@ dependencies = [
  "time",
  "timezone_provider",
  "tinystr",
+ "tracing",
  "web-time",
  "writeable",
  "xsum",
--- a/core/engine/Cargo.toml
+++ b/core/engine/Cargo.toml
@@ -13,6 +13,8 @@ rust-version.workspace = true
 
 [features]
 default = ["float16", "xsum", "temporal"]
+vm-trace = []
+
 
 embedded_lz4 = ["boa_macros/embedded_lz4", "lz4_flex"]
 
@@ -92,6 +94,7 @@ xsum = ["dep:xsum"]
 native-backtrace = []
 
 [dependencies]
+tracing = "0.1"
 tag_ptr.workspace = true
 boa_interner.workspace = true
 boa_gc = { workspace = true, features = ["thin-vec", "boa_string", "arrayvec"] }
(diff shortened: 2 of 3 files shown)


jedel1043:
We already have a `trace` feature for VM tracing, so we should build off of that.

I'm not against adding `tracing` itself though, but that should be reserved for other parts of the engine: builtins, object, value, etc.

Maanvi212006:
Hi! Thanks for the feedback — that makes sense.

I see that VM tracing already exists, and modifying its output directly may not be the right approach.

Would it be more useful if I explored extending tracing to other parts of the engine, like builtins or object/value operations? For example, adding optional instrumentation hooks or logs there for better observability.

Happy to take this in a different direction if that aligns better with the project 🙂

PR #5061, opened 2026-03-14
docs: add NativeObject design document

### LINKED ISSUE #3938: Create design notes for documentation purposes
It would be really nice to have some design notes in place for both future contributors and current contributors that don't have the full knowledge about how and why some things work as they do.

- [x] #3946
- [x] #3947
- [ ] Add design documents for `JsObject`
- [ ] Add design document for hidden classes
- [ ] Add design document for `NativeObject`
- [ ] Add design document for the parser
- [ ] Add design document for the interner
- [ ] ?

yush-1018:
Resolves #3938 

Added a design doc for NativeObject. It covers how Rust types are wrapped using JsData and ObjectData, and how they interact safely with the garbage collector.

### DIFF
--- a/docs/native_object.md
+++ b/docs/native_object.md
@@ -1,0 +1,47 @@
+# `NativeObject` Design Document
+
+## Overview
+In the Boa JavaScript engine, a lot of built-in objects (like Maps, Sets, or Date) and user-defined host objects require holding underlying Rust data structures. 
+
+The `NativeObject` trait serves as a bridge between the JavaScript object system and raw Rust types. It allows any arbitrary piece of Rust data to be securely passed around, managed, and garbage-collected as a JavaScript object.
+
+## The `JsData` Trait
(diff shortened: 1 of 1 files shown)


docs/native_object.md
```diff
@@ -0,0 +1,55 @@
...
+let object = JsObject::from_proto_and_data(Some(prototype), my_data);
+```
+
+### Retrieving and Mutating the Data
+
+Once wrapped, the underlying Rust data can be accessed safely via downcasting:
+
+1. `object.downcast_ref::<T>()` -> Returns `Option<&T>` representing an immutable reference to the native object data if the types match.
+2. `object.downcast_mut::<T>()` -> Returns `Option<&mut T>` representing a mutable reference to the native object data.
+3. `object.borrow_mut()` -> Often used when working closely with garbage collected `NativeObject`s that need interior mutability inside contexts.
```
jedel1043: You might also wanna mention `downcast`, which returns a typed `JsObject<T>` that can be borrowed later without losing the type information.

jedel1043 (approved):
Looks great!

PR #5062, opened 2026-03-14
perf: optimize `JsSharedArrayBuffer::to_vec`

jedel1043:
Just a small optimization for `to_vec` that uses `memcpy`.


nekevss (approved):
LGTM

PR #5063, opened 2026-03-14
fix(object/remaining): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241. It changes the following:

`core/engine/src/object/operations.rs`: converted 4 panics (pop_frame assertions, construct value assertions) to use `js_expect`.
`core/engine/src/object/internal_methods/string.rs`: converted 1 panic (string exotic downcast assertion) to use `js_expect`.

This completes all real panics in `object/`. The remaining `.expect()`/`.unwrap()` calls in `object/` are either in `#[cfg(test)]` blocks, `///` doc comment examples, or intentional invariant assertions in non-`JsResult` contexts (borrow guards, `!` operations) and are left as-is.

### DIFF
--- a/core/engine/src/object/internal_methods/string.rs
+++ b/core/engine/src/object/internal_methods/string.rs
@@ -1,5 +1,5 @@
 use crate::{
-    Context, JsResult, JsString,
+    Context, JsExpect, JsResult, JsString,
     object::{JsData, JsObject},
     property::{PropertyDescriptor, PropertyKey},
 };
@@ -90,7 +90,7 @@ pub(crate) fn string_exotic_own_property_keys(
     // 3. Assert: Type(str) is String.
     let string = obj
         .downcast_ref::<JsString>()
-        .expect("string exotic method should only be callable from string objects")
+        .js_expect("string exotic method should only be callable from string objects")?
         .clone();
     // 4. Let len be the length of str.
     let len = string.len();
--- a/core/engine/src/object/operations.rs
+++ b/core/engine/src/object/operations.rs
@@ -458,7 +458,7 @@ impl JsObject {
         let result = context.run().consume();
         context.vm.host_call_depth = context.vm.host_call_depth.saturating_sub(1);
 
-        context.vm.pop_frame().expect("frame must exist");
+        context.vm.pop_frame().js_expect("frame must exist")?;
 
         result
     }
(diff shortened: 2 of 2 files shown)


nekevss (approved):

KaustubhOG:
**Remaining `object/` panics audit — all invariants, zero convertible**
| File(s) | Hits | Reason |
|:---|:---:|:---|
| `object/operations.rs` | 3 | `.expect("should not fail according to spec")` and two spec-guaranteed infallible ToObject/CreateDataProperty calls — spec invariants, not convertible. |
| `object/jsobject.rs` | 1 | `.expect("Object already mutably borrowed")` / `.expect("Object already borrowed")` in `borrow()`/`borrow_mut()` — GC borrow checker invariants, not convertible. |
| `object/shape/shared_shape/mod.rs` | 1 | `.expect("There should be a property")` in `property()` — shape chain invariant, property guaranteed to exist by construction. |
| `object/internal_methods/string.rs` | 1 | `.expect("string exotic method should only be callable from string objects")` in `string_get_own_property` — exotic dispatch invariant, type verified by vtable routing. |
| `object/datatypes.rs` | 1 | Compile-time `assert!` in `ObjectData::new` — alignment check, not a runtime panic, not in scope for #3241. |
| **Total** | **7** | **0 convertible** |

All remaining panics fall into: GC borrow invariants, spec-guaranteed infallible operations, shape construction invariants, exotic dispatch invariants, or compile-time assertions. None are suitable for `EngineError` conversion.

KaustubhOG:
object-top

jedel1043:
Again, those might be triggered by bugs, so it is safer to throw engine panics to catch those.

For the `borrow` set of operations, we might wanna replace those with `try_borrow` + `js_expect` for extra safety.

jedel1043:
`expect`s inside `const` blocks are fine to leave as they are, since those are not visible at execution

KaustubhOG:
> Again, those might be triggered by bugs, so it is safer to throw engine panics to catch those.
> 
> For the `borrow` set of operations, we might wanna replace those with `try_borrow` + `js_expect` for extra safety.

Ohh that's missed i thought they are not convertible  , will open a follow-up for the borrow guards (try_borrow + js_expect) and the remaining invariants. (will couple it with the /module ) then relocate this comment to 3241

PR #5064, opened 2026-03-14
fix(module/source.rs,namespace.rs,synthetic.rs): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241. It changes the following:

`core/engine/src/module/source.rs`: converted 24 panics to use `js_expect`.
`core/engine/src/module/namespace.rs`: converted 11 panics  to use `js_expect`.
`core/engine/src/module/synthetic.rs`: converted 4 panics to use `js_expect`.

This completes all real panics in `module/`. The remaining `.expect()` calls in `module/` are either in non-`JsResult` functions (`load()`, `inner_load()`, `load_link_evaluate()`, `link()`, `execute_async()`), inside `debug_assert!` blocks, or in `#[cfg(test)]` blocks and are left as-is.

### DIFF
--- a/core/engine/src/module/namespace.rs
+++ b/core/engine/src/module/namespace.rs
@@ -13,7 +13,7 @@ use crate::object::internal_methods::{
 };
 use crate::object::{JsData, JsPrototype};
 use crate::property::{PropertyDescriptor, PropertyKey};
-use crate::{Context, JsResult, JsString, JsValue, js_string, object::JsObject};
+use crate::{Context, JsExpect, JsResult, JsString, JsValue, js_string, object::JsObject};
 use crate::{JsNativeError, Module};
 
 use super::{BindingName, ResolvedBinding};
@@ -149,7 +149,7 @@ fn module_namespace_exotic_set_prototype_of(
     // 1. Return ! SetImmutablePrototype(O, V).
     Ok(
         immutable_prototype_exotic_set_prototype_of(obj, val, context)
-            .expect("this must not fail per the spec"),
+            .js_expect("this must not fail per the spec")?,
     )
 }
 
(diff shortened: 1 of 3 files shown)


nekevss (approved):

KaustubhOG:
image

KaustubhOG:
**Remaining `module/` panics audit — all invariants, zero convertible**
| File(s) | Hits | Reason |
|:---|:---:|:---|
| `module/source.rs` | 5 | `.expect("should be linking")` — state machine invariant, transition only reachable from correct state. `.expect("cannot fail for the %Promise% intrinsic")` / `.expect("%Promise% constructor must always return a Promise object")` / `.expect("async modules cannot directly throw")` — spec-guaranteed %Promise% intrinsic invariants. `panic!` on exhaustive match arm — unreachable by construction. |
| `module/synthetic.rs` | 2 | `.expect("default resolve functions cannot throw and must return a promise")` — spec-guaranteed promise invariant. `.expect("should have the module environment")` — environment set during link phase, guaranteed present at evaluate. |
| `module/namespace.rs` | 2 | Both `.expect("6. Assert: binding is a ResolvedBinding Record.")` — verbatim spec assert, binding is guaranteed non-empty by the caller's resolve step. |
| `module/loader/embedded.rs` | 1 | `.expect("Invalid compress type")` — called on a compile-time `&'static str` constant from the macro, cannot be an invalid value at runtime. |
| `module/mod.rs` (tests) | 7 | All inside `#[test]` functions — `.unwrap()` on context builder and module parse. Not in scope for #3241. |
| **Total** | **17** | **0 convertible** |

All remaining panics fall into: spec-mandated assert invariants, module state machine invariants, %Promise% intrinsic guarantees, compile-time constant `.expect`, or test assertions. None are suitable for `EngineError` conversion.

jedel1043:
Uh I'd argue these kind of state transition steps should also be engine panics, since those have the highest probability of being incorrect

Also, I think I've mentioned the spec might also have bugs, so it is safer to engine panic on spec assertions.

KaustubhOG:
> Uh I'd argue these kind of state transition steps should also be engine panics, since those have the highest probability of being incorrect
> 
> Also, I think I've mentioned the spec might also have bugs, so it is safer to engine panic on spec assertions.

Yup , just saw yours comment on the 5063 , will convert the statetranstions and spec assertions ones .

PR #5066, opened 2026-03-14
Make `JsAsyncGenerator` fully type safe

jedel1043:
Replaces the inner `JsObject` of `JsAsyncGenerator` with `JsObject<AsyncGenerator>`, which makes it fully type safe with the addition of the methods `inner_next`, `inner_return` and `inner_throw`.


HalidOdat (approved):
Nice change! 😄

PR #5068, opened 2026-03-14
Remove `aws-lc-rs` in favour of `ring`

jedel1043:
Using `aws-lc-rs` causes miscompilations in MacOS. Even then we're not using `ring` in `boa_runtime`, so folks can provide their preferred crypto library.


nekevss (approved):
Fine by me!

PR #5069, opened 2026-03-14
feat(engine): make String.prototype.repeat respect loop iteration limit

### LINKED ISSUE #5060: Runtime limits: ensure built‑in iteration helpers respect `set_loop_iteration_limit`
## Problem

- On `main`, `Context::set_loop_iteration_limit` only applies to VM‑compiled loops (`for`, `while`, `do`), but built-in methods that iterate internally in Rust may not be charged against this limit.
- `String.prototype.repeat` can iterate many times and allocate large results without decrementing the VM loop counter.
- Some built-in methods that iterate internally in Rust may bypass the VM loop-budget mechanism.

This creates a runtime safety hole: embedders think they have a loop limit, but JS code can still hang the process via “hidden” loops inside built‑ins.

## Proposal

- Charge built-in loops against the existing loop budget (e.g. via `IncrementLoopIteration::operation((), context)`) so exceeding the limit yields the same `RuntimeLimitError::LoopIteration` as VM loops.
- Apply in built-ins that iterate in Rust (e.g. `String.prototype.repeat`); use a conservative check frequency (e.g. once per iteration initially); batch checks in follow-up work only where the loop body is very small and hot.

## Rationale

- Keeps the runtime limit semantics consistent between user‑written loops and built‑ins that perform equivalent work.
- Prevents denial‑of‑service through “hidden” Rust loops even when embedders configure a loop limit.
- Reuses existing VM infrastructure (no new counters or error types).

## Scope

- Built-ins implemented with explicit Rust loops.
- No changes to VM loop accounting logic itself.
- This issue intentionally starts with `String.prototype.repeat` as a minimal entry point.

## Implementation

This issue focuses on ensuring built-ins that perform explicit Rust iteration respect the existing runtime loop budget.

- Identify native iteration loops in the targeted built-ins (starting with `String.prototype.repeat` as a concrete example).
- Insert calls to `IncrementLoopIteration::operation((), context)` in those loops.
- The goal is to ensure built-ins consume the same iteration budget as equivalent JS loops.

## Checklist

- [ ] Use `IncrementLoopIteration::operation((), context)` inside `String.prototype.repeat`.
- [ ] Identify built-ins with explicit Rust loops and ensure they respect the loop iteration limit.
- [ ] Add tests that:
  - Verify normal behavior is unchanged when no loop limit is set.
  - With a small loop limit, both a plain JS loop and the targeted built-ins throw the same runtime-limit error.
  - Confirm the thrown error matches the same `RuntimeLimitError::LoopIteration` produced by exceeding the limit in a normal JS loop.
  - Include at least one regression test for a large `repeat` count to demonstrate the limit prevents unbounded work without panicking or hanging.

---

alienx5499:
This Pull Request fixes/closes #5060.

It changes the following:

- Charge built-in iteration loops against the existing loop budget so they respect `Context::set_loop_iteration_limit`
- Use `IncrementLoopIteration::operation((), context)` inside `String.prototype.repeat` so large repeat counts consume the VM loop budget and throw `RuntimeLimitError::LoopIteration` when the limit is exceeded
- Add tests that:
  - Verify normal behavior is unchanged when no loop limit is set
  - With a small loop limit, both a plain JS loop and `String.prototype.repeat` throw the same `RuntimeLimitError::LoopIteration`
  - Include a regression test for a large repeat count to ensure the limit prevents unbounded work without panicking or hanging

Testing:

- `cargo test -p boa_engine` (including the new runtime-limit tests for built-in iteration)

### DIFF
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -703,9 +703,13 @@ impl String {
                     return Ok(js_string!().into());
                 }
                 let n = n as usize;
-                let mut result = Vec::with_capacity(n);
 
-                std::iter::repeat_n(string.as_str(), n).for_each(|s| result.push(s));
+                // Charge each repetition against the VM loop-iteration limit.
+                let mut result = Vec::new();
+                for _ in 0..n {
+                    crate::vm::opcode::IncrementLoopIteration::operation((), context)?;
+                    result.push(string.as_str());
+                }
 
                 // 6. Return the String value that is made from n copies of S appended together.
                 Ok(JsString::concat_array(&result).into())
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -705,7 +705,7 @@ impl String {
                 let n = n as usize;
 
                 // Charge each repetition against the VM loop-iteration limit.
-                let mut result = Vec::new();
+                let mut result = Vec::with_capacity(n);
                 for _ in 0..n {
                     crate::vm::opcode::IncrementLoopIteration::operation((), context)?;
                     result.push(string.as_str());
--- a/core/engine/src/builtins/string/tests.rs
+++ b/core/engine/src/builtins/string/tests.rs
@@ -104,6 +104,40 @@ fn repeat() {
 }
 
 #[test]
+fn repeat_respects_loop_limit() {
+    run_test_actions([
+        // Set a small loop iteration limit so long repeats trip it.
+        TestAction::inspect_context(|context| {
+            context.runtime_limits_mut().set_loop_iteration_limit(10);
+        }),
+        // Plain JS loop should hit the runtime limit.
+        TestAction::assert_runtime_limit_error(
+            "for (let i = 0; i < 100; ++i) {}",
+            crate::error::RuntimeLimitError::LoopIteration,
+        ),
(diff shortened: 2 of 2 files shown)


alienx5499:
Hi @jedel1043,

CI checks have passed for this PR. When you have some time, I’d appreciate a review.

This change focuses on the minimal entry point mentioned in the issue `String.prototype.repeat` to ensure builtins implemented with Rust loops correctly charge the VM loop iteration budget. If this approach looks good, I can follow up with additional PRs auditing other builtins that iterate internally.

core/engine/src/builtins/string/mod.rs
```diff
@@ -703,9 +703,13 @@ impl String {
                     return Ok(js_string!().into());
                 }
                 let n = n as usize;
-                let mut result = Vec::with_capacity(n);
 
-                std::iter::repeat_n(string.as_str(), n).for_each(|s| result.push(s));
+                // Charge each repetition against the VM loop-iteration limit.
+                let mut result = Vec::new();
```
jedel1043: You forgot to carry over the `with_capacity`

alienx5499: I’ve restored `Vec::with_capacity(n)` in the latest commit.

jedel1043 (approved):
Looks good!

PR #5070, opened 2026-03-14
fix(runtime-limits): charge built-in iteration loops against loop budget

### LINKED ISSUE #5060: Runtime limits: ensure built‑in iteration helpers respect `set_loop_iteration_limit`

linisha15:
Description

 Summary
This PR ensures built-in iteration performed in Rust respects the runtime loop budget configured via `set_loop_iteration_limit`.

What changed
- Added loop-budget charging in built-in iteration paths using `IncrementLoopIteration::operation((), context)`.
- Covered `String.prototype.repeat` and shared Array/TypedArray iteration helper paths.
- Added runtime-limit parity tests to ensure behavior matches VM loop limits.
- Added regression coverage for large repeat counts to prevent unbounded work.

 Why
Previously, loop limits applied to VM-compiled loops but not all built-in Rust loops, which could bypass runtime protection.

Validation
-Verified runtime-limit tests for:
 - VM loop limit behavior
 - Built-in parity behavior
 - Large repeat regression behavior

Closes #5060

### DIFF
--- a/.github/workflows/pr_management.yml
+++ b/.github/workflows/pr_management.yml
@@ -1,0 +1,65 @@
+name: PR Management
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, closed]
+
+jobs:
+  manage_pr:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      issues: write
+    steps:
+      - name: Auto Add Label
+        if: github.event.action == 'opened' || github.event.action == 'reopened' || github.event.action == 'synchronize'
+        uses: actions/github-script@v8
+        with:
+          script: |
+            github.rest.issues.addLabels({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: ['waiting-for-review']
+            })
+
+      - name: Auto Remove Label
+        if: github.event.action == 'closed'
+        uses: actions/github-script@v8
+        continue-on-error: true
+        with:
+          script: |
+            try {
+              await github.rest.issues.removeLabel({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                name: 'waiting-for-review'
+              });
+            } catch (error) {
+              console.log('Label "waiting-for-review" not found or could not be removed.');
+            }
+
+      - name: Auto Assign Milestone
+        if: github.event.action == 'opened'
+        uses: actions/github-script@v8
+        with:
+          script: |
+            // Fetch open milestones and assign the closest one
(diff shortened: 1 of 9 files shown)


linisha15:
Hi! @jedel1043 , please review my pr .

linisha15:
Since the issue has already been resolved in another PR, I'll close this one. Thanks!

PR #5078, opened 2026-03-14
feat(intl): implement Number.prototype.toLocaleString with Intl support

### LINKED ISSUE #5074: Implement Intl support for Number.prototype.toLocaleString
Spec: https://tc39.es/ecma402/#sup-number.prototype.tolocalestring
Mozilla Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

alienx5499:
This Pull Request fixes/closes #5074.

It changes the following:

- Implement `Number.prototype.toLocaleString([locales [, options]])` using `Intl.NumberFormat` when the `intl` feature is enabled, following the same pattern as `BigInt.prototype.toLocaleString`.
- Return `"NaN"`, `"Infinity"`, and `"-Infinity"` for non-finite values instead of going through `Decimal::try_from_f64`.
- Keep the existing fallback behaviour (simple number string conversion) when the `intl` feature is disabled.
- Add shared tests for 0, 5, -25, NaN, Infinity, -Infinity.
- Add `#[cfg(feature = "intl")]` tests for locale formatting (`en-US`, `de-DE`) and options (`useGrouping`, `minimumFractionDigits`).

Testing:

- `cargo test -p boa_engine --lib -- number::tests::to_locale_string` (without intl)
- `cargo test -p boa_engine --features intl_bundled --lib -- number::tests::to_locale_string` (with intl)

### DIFF
--- a/core/engine/src/builtins/number/mod.rs
+++ b/core/engine/src/builtins/number/mod.rs
@@ -291,23 +291,51 @@ impl Number {
     ///
     /// The `toLocaleString()` method returns a string with a language-sensitive representation of this number.
     ///
-    /// Note that while this technically conforms to the Ecma standard, it does no actual
-    /// internationalization logic.
-    ///
     /// More information:
     ///  - [ECMAScript reference][spec]
     ///  - [MDN documentation][mdn]
     ///
-    /// [spec]: https://tc39.es/ecma262/#sec-number.prototype.tolocalestring
+    /// [spec]: https://tc39.es/ecma402/#sup-number.prototype.tolocalestring
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString
     #[allow(clippy::wrong_self_convention)]
+    #[allow(
+        unused_variables,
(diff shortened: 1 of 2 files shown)


alienx5499:
Hi @jedel1043,

Whenever you have a moment, could you please take a look at this PR? I’d really appreciate your review.

Thanks!

core/engine/src/builtins/number/mod.rs
```diff
@@ -291,23 +291,51 @@ impl Number {
...
+
+        #[cfg(feature = "intl")]
+        {
+            use fixed_decimal::{Decimal, FloatPrecision};
+
+            use crate::builtins::intl::NumberFormat;
+
+            if !x.is_finite() {
+                return Ok(JsValue::new(js_string!(x)));
+            }
```
jedel1043, line 321: Note for the future: there are technically internationalized versions of non-finite f64 values, but supporting this might be blocked by https://github.com/unicode-org/icu4x/issues/862.

jedel1043 (approved):
Looks good!

PR #5079, opened 2026-03-14
Implement core disposal logic for explicit resource management

### LINKED ISSUE #5314: implement SuppressedError builtin (ES2024, required for using / await using runtime semantics)
## Summary

Boa is missing the `SuppressedError` builtin, which is part of the ECMAScript 2024
standard. This is a blocking gap for the Explicit Resource Management feature
(`using` / `await using`) - without it, the engine cannot correctly handle errors
that occur during resource disposal as required by the `DisposeResources` abstract
operation.

---

## Current Behavior

Calling `SuppressedError` in Boa throws a `ReferenceError` because the constructor
does not exist:

```js
new SuppressedError(new Error("dispose failed"), new Error("original"), "suppressed")
// ReferenceError: SuppressedError is not defined
```

Additionally, if a `using` block exits with an error and `[Symbol.dispose]()` also
throws, the engine has no mechanism to wrap both errors — the original error is
silently lost.

---

## Expected Behavior

`SuppressedError` should be available as a global builtin that inherits from
`Error.prototype`, with the following constructor signature:

```js
new SuppressedError(error, suppressed [, message])
```

And expose the following instance properties:

| Property | Description |
|----------|-------------|
| `.error` | The new error thrown during disposal |
| `.suppressed` | The original error that was suppressed |
| `.message` | Optional message string, defaults to `""` |
| `.name` | `"SuppressedError"` |
| `.stack` | Inherited from `Error.prototype` |

---

## Acceptance Criteria

- [ ] `SuppressedError` is available as a global constructor
- [ ] `new SuppressedError(error, suppressed, message)` correctly sets `.error`,
      `.suppressed`, and `.message` properties
- [ ] `SuppressedError.prototype` inherits from `Error.prototype`
- [ ] `SuppressedError.prototype.name` is `"SuppressedError"`
- [ ] `new SuppressedError() instanceof Error` returns `true`
- [ ] `new SuppressedError() instanceof SuppressedError` returns `true`
- [ ] A `SuppressedError` variant is added to `JsNativeErrorKind` so the VM can
      throw it natively from Rust-side disposal logic
- [ ] All relevant test262 tests for `SuppressedError` pass

**Implementation note:** The closest reference in the codebase is `AggregateError`
in `core/engine/src/builtins/error/aggregate.rs` - `SuppressedError` should follow
the same pattern. Files to touch:

| File | Change |
|------|--------|
| `core/engine/src/builtins/error/suppressed.rs` | **[new]** |
| `core/engine/src/builtins/error/mod.rs` | **[modify]** Register as native error |
| `core/engine/src/context/intrinsics.rs` | **[modify]** Add to standard constructors |
| `core/engine/src/error.rs` | **[modify]** Add `SuppressedError` to `JsNativeErrorKind` |

---

## Spec Reference

- [SuppressedError constructor: tc39.es/ecma262/#sec-suppressederror-constructor](https://tc39.es/ecma262/#sec-suppressederror-constructor)
- [DisposeResources abstract operation: tc39.es/ecma262/#sec-disposeresources](https://tc39.es/ecma262/#sec-disposeresources)
- [Explicit Resource Management proposal: github.com/tc39/proposal-explicit-resource-management](https://github.com/tc39/proposal-explicit-resource-management)

abhinavs1920:
This Pull Request implements part of #4445 and fixes/closes #5314.

# Make `using` actually dispose and SuppressedError builtin  ( continued work )

Follow-up to my last PR (#4649) that added parsing.

Now when a block ends, resources get cleaned up properly.

- New bytecodes: AddDisposableResource + DisposeResources + PushDisposalScope
- VM now has a disposal stack + scope tracking
- Bytecompiler adds resources after each `using` and cleans up at end of blocks

- Add `SuppressedError` builtin following `AggregateError` pattern
- Add `.error` and `.suppressed` properties to `SuppressedError
- Update tests to expect `TypeError` for non-disposable values

Works like:

```js
{
  using file = openFile();
} // ← auto disposed here

{
  using a = one();
  using b = two();
} // disposes b then a

{
  using outer = ...;
  { using inner = ...; } // inner first
} // then outer

using x = null; // just skips, no crash

### CODE REVIEW

File: core/engine/src/vm/mod.rs [hunk 1]
```diff
@@ -98,6 +98,16 @@ pub struct Vm {
...
 
+    /// Stack of disposable resources for explicit resource management.
+    ///
+    /// Resources are added via `using` declarations and disposed in reverse order (LIFO)
+    /// when the scope exits. Each entry contains (`value`, `dispose_method`, `scope_depth`).
+    pub(crate) disposal_stack: Vec<(JsValue, JsValue)>,
+
+    /// Tracks the disposal stack depth for each scope level.
+    /// When a scope exits, we dispose resources back to this depth.
+    pub(crate) disposal_scope_depths: Vec<usize>,
```
jedel1043: This shouldn't be a runtime parameter. You can statically determine how many `using` parameters to drop in the bytecompiler

jedel1043 (changes requested):

core/engine/src/vm/mod.rs
abhinavs1920: I initially thought about this like a runtime stack problem (similar to environment scopes), but that turned out to be over-engineering. The number of using bindings per block is actually fixed at parse time, so we can just bake it directly into the instruction.

core/engine/src/vm/opcode/mod.rs
```diff
@@ -2139,12 +2145,73 @@ generate_opcodes! {
     ///   - Output: dst
     CreateUnmappedArgumentsObject { dst: RegisterOperand },
 
-    /// Reserved [`Opcode`].
-    Reserved1 => Reserved,
-    /// Reserved [`Opcode`].
-    Reserved2 => Reserved,
-    /// Reserved [`Opcode`].
-    Reserved3 => Reserved,
```
jedel1043, line 2160: Please don't remove reserved instructions from the start; it makes it much more difficult to determine how many operations we have available.

core/engine/src/vm/opcode/global.rs
jedel1043, line 1: All of this was deleted on a previous PR. Why add it back?

abhinavs1920: Sorry about that! :/ 
I was looking at the opcode enum and noticed those variants were listed but I couldn't find their implementations anywhere, so I assumed they got accidentally dropped and added them back. Should have checked the git history more carefully before doing that - my bad.

core/engine/src/vm/opcode/mod.rs
abhinavs1920: Yeah that was a dumb mistake on my part. I needed space for the new opcodes and just removed from the top without thinking about what that does to the byte values of everything that follows. Removed from the end instead and restored Reserved1/2/3 where they were. Also added Reserved57/58 to keep the total at 256 since removing the global opcodes freed up slots.

jedel1043 (changes requested):
Looks great, I really like the staged approach you're doing to implement the feature.

One thing though is that we need to gate everything behind the `experimental` feature while the proposal is not in stage 4. The parser can stay as it is since that usually doesn't get gated, but the bytecompiler should throw an error instruction instead of pushing a resource to the resources stack if the feature is not enabled.

abhinavs1920:
Done, pushed the fix. Makes sense to keep unfinished proposals gated, I should've caught that myself looking at how import.defer was handled :)

Quick question while I'm here: for the next PR I'm planning to wire up disposal into the exception handling path (try-finally integration + SuppressedError). Is there anything specific about how the current handler/finally mechanism works that I should be aware of before diving in, or is the existing `try.rs` in the bytecompiler a good enough starting point?

jedel1043:
Yes, `try.rs` is half of the solution but the other half lives in https://github.com/boa-dev/boa/blob/main/core/engine/src/bytecompiler/jump_control.rs, which is our jump handler tracking code that ensures `finally` statements are always executed even when `return`ing from functions or `break`ing from loops.

I imagine you can reuse the code that we have for `try-finally` to integrate `using` statements, since those are very similar in concept

abhinavs1920:
Just went through both files. The picture is clearer now, `jump_control.rs` is basically the glue that makes sure any break, continue, or return inside a try block doesn't just jump past the finally. It does this by collecting `JumpRecords` and routing them through a `JumpTable` at the end of the finally block so each one ends up at the right destination after disposal runs.

For using, I think the approach would be: push a `JumpControlInfo` with `TRY_WITH_FINALLY` semantics when entering a block that has using declarations, wrap the block body in a handler (like `push_handler` does for try), and on exit emit `DisposeResources` in the same spot `compile_finally_stmt` runs today. That way breaks/returns/exceptions all naturally route through disposal before leaving the scope.

Does that sound like the right direction, or is there a cleaner way you had in mind :)

jedel1043:
Yep, sounds like the correct approach

abhinavs1920:
Heyyy @jedel1043, can you have a quick look when you're available? 👀

jedel1043 (changes requested):
It might be better to start integrating the `try-finally` semantics in this PR instead of postponing it to a separate phase, because a lot of the new code here will have to be adjusted (or even removed) when integrating that.

abhinavs1920:
Heyyy,
 
I was thinking more about this and I feel like the current approach with static counting + emitting `DisposeResources` at block exit isn’t going to hold up once we integrate `try/finally`.

Since disposal needs to run on all exits (exceptions, early returns, etc.), it seems like we’ll need to move closer to the jump-control / handler-based approach anyway.

What I’m considering is:
- removing the `using_count` logic from `block.rs`
- instead pushing a `JumpControlInfo` when entering a block that has `using` declarations (similar to how `try` works)
- wrapping the block body with a handler
- and routing all exits (return, break, continue, throw) through disposal via the jump table

Would push a update soon!! :)

abhinavs1920:
Quick update: everything looks good on my side, try/finally is working and all tests pass (including the new exception cases).

I did notice one issue though. Using inside function bodies works fine in tests (via eval), but it panics when run through the CLI. The panic points to [`mod.rs:90`](https://github.com/boa-dev/boa/blob/master/core/engine/src/vm/opcode/define/mod.rs#L90), and it looks like a binding index mismatch.

Since all tests pass, I’m guessing this is due to a difference between how eval and top-level code are compiled.

Let me know you thoughts on this...

jedel1043:
What is the status of this PR? From the last comment, it seems like it's still not finished since it doesn't work in our CLI

abhinavs1920:
> What is the status of this PR? From the last comment, it seems like it's still not finished since it doesn't work in our CLI

Thanks for the nudge. I will check and let you know the PR status.

PR #5080, opened 2026-03-15
feat(intl): implement Date.prototype.toLocaleString, toLocaleDateString, and toLocaleTimeString

### LINKED ISSUE #5075: Implement Date.prototype.toLocaleString and friends
Spec:
 - https://tc39.es/ecma402/#sup-date.prototype.tolocalestring
 - https://tc39.es/ecma402/#sup-date.prototype.tolocaledatestring
 - https://tc39.es/ecma402/#sup-date.prototype.tolocaletimestring
Docs:
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString

alienx5499:
This Pull Request fixes/closes #5075.

It changes the following:

- Implement `Date.prototype.toLocaleString`, `toLocaleDateString`, and `toLocaleTimeString` using `Intl.DateTimeFormat` when the `intl` feature is enabled (ECMA-402 20.4.1–20.4.3).
- Add shared helper `format_date_time_locale` in `intl/date_time_format`; expose `create_date_time_format` as `pub(crate)`; add `FormatDefaults::All` and fix default-application logic for it.
- Add `this_date_value` helper for the RequireInternalSlot / `[[DateValue]]` step in the three methods.
- Return `"Invalid Date"` for NaN in all three methods; when `intl` is disabled, fall back to `Date.prototype.toString`.
- Add `#[cfg(feature = "intl")]` tests: invalid receiver (TypeError), NaN → `"Invalid Date"`, `typeof` checks, locale differences (en-US vs de-DE) for all three methods, and options pipeline (`dateStyle: 'short'`).

## Testing

- `cargo test -p boa_engine --lib -- date::tests::date_proto_to_locale_string_intl` (without intl; test is skipped)
- `cargo test -p boa_engine --features intl_bundled --lib -- date::tests::date_proto_to_locale_string_intl` (with intl)

### DIFF
--- a/core/engine/src/builtins/date/mod.rs
+++ b/core/engine/src/builtins/date/mod.rs
@@ -1608,15 +1608,6 @@ impl Date {
         func.call(this, &[], context)
     }
 
-    /// Returns the `[[DateValue]]` internal slot (RequireInternalSlot(dateObject, `[[DateValue]]`)).
-    fn this_date_value(this: &JsValue) -> JsResult<f64> {
-        this.as_object()
-            .and_then(|obj| obj.downcast_ref::<Date>().as_deref().copied())
-            .ok_or_else(|| JsNativeError::typ().with_message("'this' is not a Date"))
-            .map(|d| d.0)
-            .map_err(Into::into)
-    }
-
     /// [`Date.prototype.toLocaleDateString()`][spec].
     ///
     /// The `toLocaleDateString()` method returns the date portion of the given Date instance according
@@ -1636,21 +1627,25 @@ impl Date {
         args: &[JsValue],
         context: &mut Context,
     ) -> JsResult<JsValue> {
-        // Let dateObject be the this value.
-        // Perform ? RequireInternalSlot(dateObject, [[DateValue]]).
-        // Let x be dateObject.[[DateValue]].
-        let t = Self::this_date_value(this)?;
-        // If x is NaN, return "Invalid Date".
-        if t.is_nan() {
-            return Ok(JsValue::new(js_string!("Invalid Date")));
-        }
-        // Let dateFormat be ? CreateDateTimeFormat(%Intl.DateTimeFormat%, locales, options, date, date).
-        // Return ! FormatDateTime(dateFormat, x).
         #[cfg(feature = "intl")]
         {
             use crate::builtins::intl::date_time_format::{
                 FormatDefaults, FormatType, format_date_time_locale,
             };
+            // 1. Let dateObject be the this value.
+            // 2. Perform ? RequireInternalSlot(dateObject, [[DateValue]]).
+            // 3. Let x be dateObject.[[DateValue]].
+            let t = this
+                .as_object()
+                .and_then(|obj| obj.downcast_ref::<Date>().as_deref().copied())
+                .ok_or_else(|| JsNativeError::typ().with_message("'this' is not a Date"))?
+                .0;
+            // 4. If x is NaN, return "Invalid Date".
+            if t.is_nan() {
+                return Ok(JsValue::new(js_string!("Invalid Date")));
+            }
+            // 5. Let dateFormat be ? CreateDateTimeFormat(%Intl.DateTimeFormat%, locales, options, date, date).
+            // 6. Return ! FormatDateTime(dateFormat, x).
             let locales = args.get_or_undefined(0);
             let options = args.get_or_undefined(1);
             format_date_time_locale(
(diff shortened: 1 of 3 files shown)


core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -967,3 +971,57 @@ fn unwrap_date_time_format(
...
+        format_type,
+        defaults,
+        context,
+    )?;
+    let format_val = dtf.get(js_string!("format"), context)?;
+    let format_fn = format_val
+        .as_callable()
+        .ok_or_else(|| JsNativeError::typ().with_message("format is not callable"))?;
+    format_fn.call(&dtf.into(), &[JsValue::from(timestamp)], context)
+}
```
jedel1043, line 1055: I don't think this is the correct approach. This is interacting with `DateTimeFormat` as if it was an external crate to be consumed, but since the creation of the formatter is not visible to the external world, we can get away with calling `create_date_time_format` directly.

My suggestion here would be to tweak `create_date_time_format` such that it returns `DateTimeFormatter` directly, then use that here to avoid creating a new `JsObject`, and move the object creation logic to the constructor instead

jedel1043: Kinda weird to do it this way in here but use the correct approach in `Number.prototype.toLocaleString` by the way

alienx5499: Thanks for the suggestion! I refactored `create_date_time_format` to return the internal `DateTimeFormat` and moved the `JsObject` creation into the constructor. `Date.prototype.toLocaleString` and related methods now format directly using the internal formatter, similar to `Number.prototype.toLocaleString`. Addressed in `725441e`.

core/engine/src/builtins/date/mod.rs
```diff
@@ -1608,6 +1608,15 @@ impl Date {
         func.call(this, &[], context)
     }
 
+    /// Returns the `[[DateValue]]` internal slot (RequireInternalSlot(dateObject, `[[DateValue]]`)).
```
nekevss: nit: accessing the inner field should remain as the current style.

core/engine/src/builtins/date/mod.rs
```diff
@@ -1635,16 +1675,47 @@ impl Date {
...
+        args: &[JsValue],
+        context: &mut Context,
     ) -> JsResult<JsValue> {
-        Err(JsError::from_opaque(JsValue::new(js_string!(
-            "Function Unimplemented]"
-        ))))
+        // Let dateObject be the this value.
+        // Perform ? RequireInternalSlot(dateObject, [[DateValue]]).
+        // Let x be dateObject.[[DateValue]].
+        let t = Self::this_date_value(this)?;
```
nekevss: issue: these steps are only applicable to the intl flagged implementation.

core/engine/src/builtins/date/mod.rs
```diff
@@ -1635,16 +1675,47 @@ impl Date {
...
-        _: &[JsValue],
-        _context: &mut Context,
+        this: &JsValue,
+        args: &[JsValue],
+        context: &mut Context,
     ) -> JsResult<JsValue> {
-        Err(JsError::from_opaque(JsValue::new(js_string!(
-            "Function Unimplemented]"
-        ))))
+        // Let dateObject be the this value.
```
nekevss: nit: style, include the step numbers alongside the spec steps

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -807,13 +805,13 @@ fn create_date_time_format(
         // 2. If value is not undefined, set needDefaults to false.
         let needs_defaults = format_options.check_dtf_type(date_time_format_type);
         // d. If needDefaults is true and defaults is either date or all, then
-        if needs_defaults && defaults != FormatDefaults::Time {
+        if needs_defaults && (defaults == FormatDefaults::All || defaults == FormatDefaults::Date) {
```
nekevss: question: why was this changed? The previous looks correct and acts in less steps

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -843,21 +841,57 @@ fn create_date_time_format(
...
+
+/// Formats a timestamp (epoch milliseconds) using the given [`DateTimeFormat`] internals.
+/// Used by the bound `format` function and by [`format_date_time_locale`] without creating a JS object.
+fn format_timestamp_with_dtf(
+    dtf: &DateTimeFormat,
+    timestamp: f64,
+    context: &mut Context,
+) -> JsResult<JsString> {
+    // FormatDateTime / PartitionDateTimePattern: TimeClip, then ToLocalTime, then format.
+    let x = time_clip(timestamp);
```
nekevss: issue: we should not be handling this logic again.

We already handle this logic in two different places. There should not be a reason to implement it a second time. Maybe you're looking for `ToLocalTime`?

nekevss (changes requested):
Noticed a couple things while reviewing this.

alienx5499:
Hi @nekevss,

I've pushed updates addressing the review comments and resolved the outdated threads. Could you please take another look when you have time?

Thanks!

jedel1043:
Deferring to @nekevss for the final review, but LGTM from my side

core/engine/src/builtins/intl/date_time_format/mod.rs
```diff
@@ -843,21 +793,53 @@ fn create_date_time_format(
...
+
+/// Formats a timestamp (epoch milliseconds) using the given [`DateTimeFormat`] internals.
+/// Callers must have already applied `TimeClip` and `NaN` check (`FormatDateTime` steps 1–2).
+/// This performs `ToLocalTime` and format only.
+fn format_timestamp_with_dtf(
+    dtf: &DateTimeFormat,
+    timestamp: f64,
+    context: &mut Context,
+) -> JsResult<JsString> {
+    let time_zone_offset = match dtf.time_zone {
```
nekevss: issue: keep the doc lines that are applied with the methods.

Adapting dtf can be incredibly complicated, and this function is probably going to need to be updated / reworked in order to support temporal. So keeping the comments in this instance is very important.

nekevss (changes requested):
I think one last issue and then this should probably be good.

alienx5499:
Hi @nekevss, I’ve pushed commit `6b270bf` addressing your comment.

It keeps the spec step documentation alongside `format_timestamp_with_dtf` and introduces a `time_zone` binding to better mirror the spec flow and reuse the value.

Would appreciate a re-review when you have time.

alienx5499:
Hi @nekevss,

Just following up on this, I’ve addressed all the previous review comments and incorporated the requested changes.

Please let me know if there’s anything else I should update or refine. Happy to iterate further.

Thanks!

jedel1043:
Please don't repeatedly ping maintainers. A single request for review is enough. We'll review when we have the time

nekevss (approved):
Generally looks good to me.

We'll have to keep in mind how this may need to expand in order to support Temporal in the future, but that's all TBD

PR #5098, opened 2026-03-15
feat(intl): implement Temporal.PlainYearMonth.prototype.toLocaleString

### LINKED ISSUE #5083: Implement Temporal.PlainYearMonth.prototype.toLocaleString
Spec: https://tc39.es/proposal-temporal/#sec-temporal.plainyearmonth.tolocalestring

alienx5499:
This Pull Request fixes/closes #5083.

It changes the following:

- Implement `Temporal.PlainYearMonth.prototype.toLocaleString` per [spec](https://tc39.es/proposal-temporal/#sec-temporal.plainyearmonth.tolocalestring). It now acts as a thin wrapper that delegates to a new, centralized `HandleDateTimeValue` dispatcher (§15.6.22) in the `Intl` module. When the caller omits `dateStyle`/`timeStyle`, default `year: "numeric"` and `month: "short"` are applied so ICU does not fall back to full date defaults (reference `day`) — fixes Test262 `default-does-not-include-day-time-and-time-zone-name.js`.
- Implement `HandleDateTimeTemporalYearMonth` (§15.6.16) to compute the Intl formatting anchor from the Temporal plain value; validate Intl `calendar` against the Temporal calendar, and model Temporal plain `timeZone` by forcing Intl `timeZone` to `"+00:00"`.
- Add tests for `PlainYearMonth.prototype.toLocaleString`: returns string, invalid receiver throws `TypeError`, and (with the `intl` feature) different locales/options affect output; plain `timeZone` is ignored/overridden, incompatible `calendar` throws, and default output excludes reference day/time/timeZoneName (regression for Test262 `default-does-not-include-day-time-and-time-zone-name.js`).

Testing

- `cargo test -p boa_engine --no-default-features --features temporal --lib plain_year_month` (Temporal-only tests)
- `cargo test -p boa_engine --features intl_bundled,temporal --lib plain_year_month` (with intl)
- `cargo run -p boa_tester -- run --suite test/intl402/Temporal/PlainYearMonth/prototype/toLocaleString/default-does-not-include-day-time-and-time-zone-name.js`

Status: rebased onto `main` after #5080; this PR now only contains the Temporal `PlainYearMonth.prototype.toLocaleString` changes for #5083.


core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -769,7 +776,30 @@ impl PlainYearMonth {
                 JsNativeError::typ().with_message("this value must be a PlainYearMonth object.")
             })?;
 
-        Ok(JsString::from(year_month.inner.to_string()).into())
+        #[cfg(feature = "intl")]
+        {
+            use crate::builtins::{
```
nekevss: issue: the implementation here should ideally follow or adapt the specification steps provided in the Temporal specification as additions to ECMA402

nekevss (changes requested):
Generally worth noting, that from what I can tell. The `T.p.toLocaleString` implementations all seem to require similar plumbing.

The specification can be found [here](https://tc39.es/proposal-temporal/#sec-temporal-handledatetimevalue)

alienx5499:
Hi @nekevss,

I’ve addressed your feedback and aligned the implementation more closely with the Temporal -> ECMA-402 specification steps, and rebased on latest main.

Since #5080 is now merged, the original blocking dependency should be resolved. Could you please take another look when you have time?

If the **Blocked** label was only due to that dependency, it may no longer be needed.

Let me know if anything still needs adjustment.

Thanks!

core/engine/src/builtins/temporal/plain_year_month/mod.rs
```diff
@@ -755,21 +759,139 @@ impl PlainYearMonth {
...
+                    options_obj.create_data_property_or_throw(
+                        js_string!("month"),
+                        JsValue::from(js_string!("short")),
+                        context,
+                    )?;
+                }
+            }
+
+            // Compute epochNs from isoDate + NoonTimeRecord (steps 2-3), then pass it to the shared
+            // ECMA-402 formatting pipeline as epoch milliseconds.
```
nekevss: issue: this should ideally not be inlined from `HandleDateTimeValue`.

The implementation here should probably stick as close to the specification as possible while also being able to expand to support the other temporal built-ins

nekevss (changes requested):
This is looking good overall and the tests appear to be promising.

This method is a little tricky, especially when it comes to the specification. There's a lot of complicated specification text, but in order to better support the other temporal built-ins and follow up work on this code path.

It's worth noting that if there's other code paths that you aren't trying to support currently. Feel free to return a result where the other branches return a `js_error!("Not yet implemented")`.

alienx5499:
Hi @nekevss,

Updated based on your feedback, moved the `Temporal.PlainYearMonth` logic out of `HandleDateTimeValue` and into a dedicated handler, keeping the dispatcher clean and closer to the spec structure.

Would appreciate a re-review when you have time.

Thanks!

alienx5499:
Hi @nekevss,

Just a quick followup, the earlier feedback has been addressed.

Happy to refine further if needed. Would appreciate a re review when you get a chance. 

Thanks!

Vellumic:
I noticed this PR uses logic that violates the ECMA-402 specification regarding `dateStyle`/`timeStyle` fallbacks. I've already submitted a separate bugfix and regression test for this issue here: #5324.

Furthermore, this PR duplicates parts of the formatting infrastructure I'm currently working on in #5199. We should avoid merging these changes to prevent double-work and immediate technical debt.

alienx5499:
@Vellumic
You’re right about the `dateStyle`/`timeStyle` fallback, that was introduced earlier in #5080 and is now corrected here, along with regression tests.

This PR focuses on implementing `Temporal.PlainYearMonth.prototype.toLocaleString` #5083 with minimal, spec-aligned changes.

Vellumic:
I see you used my fix in your code, however it doesn't fixes the general issue I was addressing earlier. Because `Temporal` was designed with legacy `Date` flaws in mind, it needs more complex parsing.

alienx5499:
@Vellumic

The change here addresses the dateStyle/timeStyle fallback behavior in the shared Date formatting path, which this PR depends on.

This PR is scoped to `Temporal.PlainYearMonth.prototype.toLocaleString` #5083. Broader Temporal formatting concerns should be handled separately if needed.

PR #5100, opened 2026-03-15
feat(vm): emit `Opcode::Debugger` for `debugger;` statements

tkshsbcue:
## Summary

Adds a dedicated `Opcode::Debugger` that the bytecompiler emits for `debugger;` statements. Previously these were silently ignored (empty match arm in the bytecompiler); now they are represented in bytecode and visible during tracing.

Currently a no-op — this lays groundwork for future debugger integration to intercept these points.

**Before:** `debugger;` produced no bytecode at all

**After:**
```
Location     Opcode        Operands
  000000     Debugger
  000001     PushOne       dst:r02
  ...
```

## Changes

- **`vm/opcode/mod.rs`** — new `Debugger` variant in `generate_opcodes!`, module declaration and re-export
- **`vm/opcode/debugger/mod.rs`** — new file, no-op operation implementation
- **`bytecompiler/statement/mod.rs`** — emit `Debugger` opcode instead of ignoring the statement
- **`vm/flowgraph/mod.rs`** — handle `Instruction::Debugger` in flowgraph generation
- **`vm/code_block.rs`** — handle `Instruction::Debugger` in operand display

## Test plan

- `cargo check` and `cargo check --features trace` pass
- `cargo test -p boa_engine` — all 183 tests pass, zero regressions
- `cargo test -p insta-bytecode` — bytecode snapshot tests pass
- Verified `debugger;` appears in trace output via `boa -t -e 'debugger; 1+1;'`

### DIFF
--- a/core/engine/src/bytecompiler/statement/mod.rs
+++ b/core/engine/src/bytecompiler/statement/mod.rs
@@ -98,7 +98,10 @@ impl ByteCompiler<'_> {
                 }
             }
             Statement::With(with) => self.compile_with(with, use_expr),
-            Statement::Empty | Statement::Debugger => {}
+            Statement::Empty => {}
+            Statement::Debugger => {
+                self.bytecode.emit_debugger();
+            }
         }
     }
 
--- a/core/engine/src/vm/code_block.rs
+++ b/core/engine/src/vm/code_block.rs
@@ -858,7 +858,8 @@ impl CodeBlock {
             | Instruction::SuperCallSpread
             | Instruction::PopPrivateEnvironment
             | Instruction::Generator
-            | Instruction::AsyncGenerator => String::new(),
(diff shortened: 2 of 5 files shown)


tkshsbcue:
Hey @jedel1043 I've been looking into the VM's bytecode pipeline and noticed `debugger;` statements were being silently dropped by the bytecompiler. Opened a small PR to emit a proper opcode for them would love your thoughts on it. I have some ideas for building on top of this that I'd love to discuss with you when you get a chance.

zhuzhu81998:
> Currently a no-op — this lays groundwork for future debugger integration to intercept these points.

i feel like in this case, it would make sense to do this later together with the debugger tooling.

jedel1043:
Yep, this should be added with the tooling itself

PR #5101, opened 2026-03-15
Implement `FinalizationRegistry`

jedel1043:
~Depends on #5107~ merged 

Funny GC with its funny semantics

anyways, this implements the last big feature we were missing from the spec. It uses the same pattern from `Atomics.waitAsync` of (ab)using async jobs to enqueue a cleanup task on a later point in time.


core/engine/src/builtins/finalization_registry/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+                    TypeError: "FinalizationRegistry.prototype.register: \
+                        invalid object type for `this`",
+                )
+            })?;
+
+        let target = args.get_or_undefined(0);
+        let held_value = args.get_or_undefined(1);
+        let unregister_token = args.get_or_undefined(2);
+
+        // 3. If CanBeHeldWeakly(target) is false, throw a TypeError exception.
```
nekevss, line 220: nit: can we inline the docs for `CanBeHeldWeakly` here.

core/engine/src/builtins/finalization_registry/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+            //   GC'd, so we don't need to worry about cleanups.
+            let _ = sender.try_send(());
+        }
+    }
+}
+
+///  A cell tracked by a [`FinalizationRegistry`].
+#[derive(Trace, Finalize)]
+pub(crate) struct RegistryCell {
+    target: Ephemeron<ErasedVTableObject, CleanupSignaler>,
```
nekevss, line 52: observation: huh, this is interesting to me.

Was it not possibly to implement this directly with `WeakGc` with a type that implements the finalizer vs. the ephemeron value type being the finalizer implementor.

nekevss (commented):
Overall looks pretty good. I had a nit and observation / question about the implementation.

None are blocking though.

nekevss (approved):

core/engine/src/builtins/finalization_registry/mod.rs
jedel1043: Don't think so. The `ErasedVTableObject` comes from JS, so you'll need to somehow "wrap" its inner data with a custom type, but we don't allow changing the type of `JsObject`s after creation

Issue #5102, opened 2026-03-15
Fix : Trace output overflow in CLI


sahanaxzy:
Right now, the trace output tries to display info in nice tables, but when the stack
gets too long, it overflows the terminal and everything looks messy and hard to follow.

We should look into ways to keep the tables clean even with large stacks — things like
truncating long entries, wrapping text, or adjusting column widths so the output stays
readable.

PR #5104, opened 2026-03-15
Improve readability of CLI trace output

sahanaxzy:
## Description

The trace output in the CLI was difficult to read when the stack grew large — table rows would overflow the terminal width, making everything look messy.

solves #5102 

### Changes

> **Group consecutive identical values**: Repeated stack entries like `undefined, undefined, undefined` are now displayed as `undefined (x3)`, reducing visual noise

> **Truncate long values**: Individual stack values are capped at 18 characters to prevent single entries from blowing out the table layout

> **Cap total stack display width**: The stack column is bounded so lines don't overflow the terminal, with a `.. (N total)` indicator showing the full stack depth

> **Empty stack handling**: Empty stacks now display `[ <empty> ]` instead of `[ ]`


jedel1043:
One suggestion that I always wanted to do was to display `undefined -> und`, `function -> func`, `object -> obj`, etc. The logic for this is that folks that are tracing through bytecode usually already know the internals of JS engines, so offering a more brief representation of all values would make it easier to show more state about the VM at once

sahanaxzy:
Thanks for the suggestion... I have updated
the stack output is lot more prettier now

Monti-27 (commented):
A couple of bugs stood out in the new stack formatter.

core/engine/src/vm/mod.rs
```diff
@@ -277,26 +277,92 @@ impl Stack {
...
+            v if v.is_callable() => return "func".to_string(),
+            v if v.is_object() => return "obj".to_string(),
+            v if v.is_undefined() => return "und".to_string(),
+            v if v.is_null() => return "null".to_string(),
+            v => v.display().to_string(),
+        };
+        if raw.len() <= Self::MAX_VALUE_LEN {
+            raw
+        } else {
+            format!("{}..", &raw[..Self::MAX_VALUE_LEN - 2])
```
Monti-27: This slices the formatted value by bytes, so a long non-ASCII value can panic here if the cutoff lands in the middle of a UTF-8 code point. A traced string with multibyte characters is enough to hit it. We need to truncate on char boundaries before appending `..`.

core/engine/src/vm/mod.rs
```diff
@@ -277,26 +277,92 @@ impl Stack {
...
+            .iter()
+            .enumerate()
+            .rev()
+            .map(|(j, v)| (Self::format_value(v), j))
+            .collect();
+
+        // Group consecutive identical values
+        let mut groups: Vec<(String, usize, Option<usize>)> = Vec::new();
+        for (val, idx) in &entries {
+            if let Some(last) = groups.last_mut()
```
Monti-27: This grouping step can swallow the frame marker. If the first value in the run has no marker, but a later identical value is the frame pointer entry, we merge it into the existing group before recording `|frame_count|`. A run of repeated `undefined` registers would end up hiding the current frame boundary completely.

core/engine/src/vm/mod.rs
```diff
@@ -277,26 +277,92 @@ impl Stack {
...
-                let _ = write!(string, " |{frame_count}|");
-            } else if i + 1 != self.stack.len() {
-                string.push(',');
+        // Build (formatted_value, original_index) pairs, top of stack first
+        let entries: Vec<(String, usize)> = self
+            .stack
+            .iter()
+            .enumerate()
+            .rev()
+            .map(|(j, v)| (Self::format_value(v), j))
```
Monti-27: We are grouping on `format_value(v)` here, not on the underlying value. That means two different long values with the same first 16 bytes collapse to the same `..` form and get rendered as `(xN)`, even though they were distinct stack entries. That changes the meaning of the trace instead of only shortening it.

core/engine/src/vm/mod.rs
```diff
@@ -277,26 +277,99 @@ impl Stack {
...
+            let separator = if let Some(fc) = marker {
+                format!(" |{fc}|")
+            } else if i + 1 < groups.len() {
+                ",".to_string()
+            } else {
+                String::new()
+            };
+
+            let addition = format!("{part}{separator} ");
+            if string.len() + addition.len() > Self::MAX_STACK_WIDTH - 10 {
```
Monti-27: Small thing here: this still reserves only 10 chars, but the suffix we append below is already longer than that. `.. (1 total) ]` is 14 chars, and it grows with the stack size. So the truncation path can still go past `MAX_STACK_WIDTH`. I think this should reserve the actual suffix length dynamically instead of using `10`.

core/engine/src/vm/mod.rs
```diff
@@ -277,26 +274,101 @@ impl Stack {
...
+            let display_val = Self::truncate_display(val);
+            let part = if *count > 1 {
+                format!("{display_val} (x{count})")
+            } else {
+                display_val
+            };
+
+            let separator = if let Some(fc) = marker {
+                format!(" |{fc}|")
+            } else if i + 1 < total && !(force_truncate && i + 1 == groups.len()) {
```
Monti-27: small thing here: this is still using total stack entries for the comma decision, so grouped output can end up with a trailing comma like [ und (x2), ]. i think this should key off rendered groups here instead.

Monti-27 (commented):
lgtm to me now. the issues i called out earlier look addressed.

Monti-27 (approved):

jedel1043 (approved):
Very nice, thank you!

jedel1043:
And thank you @Monti-27 for the thorough review

Issue #4988, opened 2026-03-10
[Tracking] WinterTC (TC55) Minimum Common Web API: gap analysis for boa_runtime


KaustubhOG:
## Overview

[WinterTC (TC55)](https://wintertc.org/) is an Ecma International Technical Committee 
working on a "Minimum Common Web API" standard — a baseline set of Web Platform APIs 
that all server-side JS runtimes (Deno, Bun, Cloudflare Workers, etc.) should implement 
for interoperability.

The [TC55 spec (December 2025)](https://min-common-api.proposal.wintertc.org/) defines 
the APIs that a conforming runtime must expose. This issue tracks the gap between what 
`boa_runtime` currently implements and what TC55 requires.

## Implemented in `boa_runtime`

- [x] `console` (optional extension)
- [x] `fetch` / `Headers` / `Request` / `Response` (feature-gated, requires custom `Fetcher` impl)
- [x] `TextEncoder` / `TextDecoder` (registered by default)
- [x] `setTimeout` / `setInterval` / `clearTimeout` / `clearInterval` (registered by default)
- [x] `structuredClone` (registered by default)
- [x] `queueMicrotask` (registered by default)
- [x] `URL` (feature-gated — `searchParams` / `createObjectURL` / `revokeObjectURL` unimplemented)
- [x] `postMessage` (optional extension, requires custom `MessageSender` impl)

### DOM
- [ ] `AbortController` / `AbortSignal`
- [ ] `Event` / `EventTarget`

### HTML
- [ ] `CustomEvent` / `ErrorEvent` / `PromiseRejectionEvent`
- [ ] `MessageChannel` / `MessagePort` / `MessageEvent`
- [ ] `atob()` / `btoa()`
- [ ] `reportError()`
- [ ] `navigator.userAgent`
- [ ] `onerror` / `onunhandledrejection` / `onrejectionhandled`
- [ ] `self`

### URL
- [ ] `URLSearchParams`
- [ ] `URLPattern`

### File API
- [ ] `Blob` / `File`

### XHR
- [ ] `FormData`

### Encoding
- [ ] `TextDecoderStream` / `TextEncoderStream`

### Compression
- [ ] `CompressionStream` / `DecompressionStream`

### Streams
- [ ] `ReadableStream` + `ReadableStreamDefaultController` + `ReadableStreamDefaultReader`
- [ ] `ReadableByteStreamController` / `ReadableStreamBYOBReader` / `ReadableStreamBYOBRequest`
[...]

PR #5105, opened 2026-03-15
feat: add boa_wintertc crate skeleton for WinterTC compliance

KaustubhOG:
relates to #4988

## Summary

Adds a new `boa_wintertc` crate under `core/wintertc/` to house the WinterTC (TC55)Minimum Common Web API implementation for Boa, as discussed in #4988.

The crate is standalone , it depends directly on `boa_engine` ,not on `boa_runtime`. The intention is that `boa_runtime` will eventually depend on `boa_wintertc` and re-expose its APIs.

## Structure

Each TC55-mandated API group gets its own module with a `register()` stub,following the same conventions as `boa_runtime`:

| Module | wintertc API | Source |
|--------|----------|--------|
| `console` | `console` | migrate from `boa_runtime` |
| `timers` | `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval` | migrate from `boa_runtime` |
| `encoding` | `TextEncoder`, `TextDecoder`, `TextEncoderStream`, `TextDecoderStream` | migrate + implement |
| `microtask` | `queueMicrotask` | migrate from `boa_runtime` |
| `clone` | `structuredClone` | migrate from `boa_runtime` |
| `base64` | `atob`, `btoa` | migrate from `boa_runtime` |
| `abort` | `AbortController`, `AbortSignal` | migrate from `boa_runtime` |
| `events` | `EventTarget`, `Event`, `CustomEvent`, `ErrorEvent`, `MessageEvent` | new |
| `url` | `URL`, `URLSearchParams` | migrate + implement |
| `fetch` | `fetch`, `Request`, `Response`, `Headers` | migrate from `boa_runtime` |

## Next Steps

- Migrate existing implementations from `boa_runtime` into their respective modules
- Implement missing APIs (`EventTarget`, `URLSearchParams`, `TextEncoderStream`, `TextDecoderStream`)
- Add `boa_wintertc` as a dependency of `boa_runtime` with re-exports

---

cc @hansl @jasonwilliams 

This is a skeleton-only draft to validate the crate architecture before.implementation begins, based on the direction discussed in #4988.

A few things I'd like confirmation on before proceeding:
1. Is `core/wintertc/` the right location?
2. Is the dependency direction correct — `boa_wintertc` standalone, `boa_runtime` depending on it?
3. Any module groupings to adjust before I start migrating implementations?


jedel1043:
One thing I'd mention is that I would prefer this to be called `boa_wintertc` or even `boa_winter` instead of `boa_tc55`. The latter is more precise, but kinda hard to find since you need to remember the exact technical committe number. In contrast, the former ones are very memorable and easy to remember.

KaustubhOG:
> One thing I'd mention is that I would prefer this to be called `boa_wintertc` or even `boa_winter` instead of `boa_tc55`. The latter is more precise, but kinda hard to find since you need to remember the exact technical committe number. In contrast, the former ones are very memorable and easy to remember.

Thanks for the feedback! `boa_wintertc` sounds precise ,  much more sensible than the committee number. I'll rename the crate .

jedel1043:
Looking really nice! To answer your questions:
> Is core/wintertc/ the right location?

Yep, pretty much.

> Is the dependency direction correct — boa_wintertc standalone, boa_runtime depending on it?

Yeah, the expectation is that `boa_runtime` will also include things like `fs` or `process`, which are less WebAPI oriented.

> Any module groupings to adjust before I start migrating implementations?

Maybe changing the `interval` module from `interval.rs` + `interval` subdirectory to subdirectory + `mod.rs`. It is the only module that has that inconsistency.

KaustubhOG:
> Looking really nice! To answer your questions:
> 
> > Is core/wintertc/ the right location?
> 
> Yep, pretty much.
> 
> > Is the dependency direction correct — boa_wintertc standalone, boa_runtime depending on it?
> 
> Yeah, the expectation is that `boa_runtime` will also include things like `fs` or `process`, which are less WebAPI oriented.
> 
> > Any module groupings to adjust before I start migrating implementations?
> 
> Maybe changing the `interval` module from `interval.rs` + `interval` subdirectory to subdirectory + `mod.rs`. It is the only module that has that inconsistency.

heyy @jedel1043  , Updated all modules to the `subdirectory/mod.rs` pattern. As implementation progresses, tests and additional files (e.g. `tests.rs`, `headers.rs`) will be added within each module's subdirectory.

Is this  structure correct?

jedel1043:
Yep, looking good

KaustubhOG:
> Yep, looking good

image

 thanks  i changed from draft to PR ,also the `boa_wintertc` isn't published in the crates.io so any baseline version i can campare with it or is it fine .

jedel1043:
To solve the lint, you should add
```
  with:
    exclude: boa_wintertc
```
to our check semver job

core/wintertc/ABOUT.md
jedel1043, line 1: Instead of adding an `ABOUT.md` here, you should modify the one we have in our root, then regenerate all `ABOUT.md` using our tool:
```shell
cargo run --bin regenerate-about
```

KaustubhOG: done! Modified the root ` ABOUT.md `and regenerated all crate docs using `cargo run --bin regenerate-about`. Also added `exclude: boa_wintert` to the semver check job to fix the CI lint failure.

core/wintertc/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[lints]
+workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+
+[features]
+default = []
+url = []
+fetch = []
```
jedel1043: Missing a new line here

.github/workflows/rust.yml
```diff
@@ -472,3 +372,5 @@ jobs:
 
       - name: Check Semver
         uses: obi1kenobi/cargo-semver-checks-action@v2
+        with:
+          exclude: boa_wintertc
```
jedel1043: Also missing a newline here

jedel1043 (changes requested):

jedel1043 (approved):
LGTM! This is a nice baseline that we can start building upon.

KaustubhOG:
Hey @jedel1043, glad to have this merged! I'm planning to work on the full boa_wintertc implementation as my GSoC 2026 project , migrating existing APIs from boa_runtime and implementing the missing ones. I have a proposal in progress around this. Would it be possible to get assigned to the tracking issue #4988?

PR #5107, opened 2026-03-16
Remove Clone requirement for Ephemeron::value

jedel1043:
This PR removes the `Clone` requirement we had for fetching `Ephemeron` values by introducing an `EphemeronValueRef` that derefs to `&V`. Soundness is preserved by storing the key of the ephemeron in the `EphemeronValueRef`, such that any collection won't be able to deallocate the value. This ensures the reference to `&V` is never invalid.

### DIFF
--- a/core/gc/src/pointers/ephemeron.rs
+++ b/core/gc/src/pointers/ephemeron.rs
@@ -5,7 +5,27 @@ use crate::{
     internals::EphemeronBox,
     trace::{Finalize, Trace},
 };
-use std::ptr::NonNull;
+use std::{ops::Deref, ptr::NonNull};
+
+/// A reference to the value held by an [`Ephemeron`].
+///
+/// This reference can be thought as a `(Gc<K>, &V)` pair, where the
+/// returned `Gc<K>` ensures that the reference `&V` is always valid until
+/// the `Gc<K>` gets dropped.
+#[derive(Debug)]
+pub struct EphemeronValueRef<'a, K: Trace + ?Sized + 'static, V> {
+    // Only required to maintain the reference `&V` alive.
+    _key: Gc<K>,
+    value: &'a V,
+}
+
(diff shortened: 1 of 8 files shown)


core/gc/src/pointers/ephemeron.rs
```diff
@@ -5,7 +5,27 @@ use crate::{
...
+
+/// A reference to the value held by an [`Ephemeron`].
+///
+/// This reference can be thought as a `(Gc<K>, &V)` pair, where the
+/// returned `Gc<K>` ensures that the reference `&V` is always valid until
+/// the `Gc<K>` gets dropped.
+#[derive(Debug)]
+pub struct EphemeronValueRef<'a, K: Trace + ?Sized + 'static, V> {
+    // Only required to maintain the reference `&V` alive.
+    _key: Gc<K>,
```
nekevss, line 18: huh, clever

core/gc/src/pointers/weak_map.rs
```diff
@@ -313,16 +313,16 @@ where
         old
     }
 
-    /// Removes a key from the map, returning the value at the key if the key
+    /// Removes a key from the map, returning `true` if the key
     /// was previously in the map. Keeps the allocated memory for reuse.
-    pub(crate) fn remove(&mut self, k: &Gc<K>) -> Option<V> {
+    pub(crate) fn remove(&mut self, k: &Gc<K>) -> bool {
```
nekevss, line 318: observation: we're deviating from most map APIs returning a bool here rather than an option.

A bool does preserve the general gist though, so 🤷‍♂️

nekevss (approved):
Fine by me.  I had one observation about the API, but I wouldn't block merging on it.

core/gc/src/pointers/weak_map.rs
jedel1043: Yes, and it also puts our API closer to ECMAScript's `WeakMap` API, so I didn't see many downsides to changing it

PR #5110, opened 2026-03-16
fix(intl): add @@toStringTag to Intl.DateTimeFormat.prototype

Lokesh0224:
**Adds the missing `@@toStringTag` property to `Intl.DateTimeFormat.prototype`.**

According to ECMA-402 11.3.2, 
The initial value of the [@@toStringTag](https://tc39.es/ecma262/#sec-well-known-symbols) property is the String value "Intl.DateTimeFormat". This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.

This was missing from the `IntrinsicObject::init` registration, causing
`Object.prototype.toString.call(new Intl.DateTimeFormat())` to return
image

### DIFF
--- a/core/engine/src/builtins/intl/date_time_format/mod.rs
+++ b/core/engine/src/builtins/intl/date_time_format/mod.rs
@@ -101,6 +101,7 @@ impl Service for DateTimeFormat {
 
 impl IntrinsicObject for DateTimeFormat {
     fn init(realm: &Realm) {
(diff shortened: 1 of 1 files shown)


Lokesh0224:
@jedel1043, While drafting my proposal i noticed this small missing implementation, so thought of fixing is right away

jedel1043 (approved):
Good catch!

PR #5115, opened 2026-03-17
Enable oscars behind feature flags in boa_gc

nekevss:
This PR adds an `unstable` and `oscars` feature flag and re-exports the oscars repo from `boa_gc`.

### DIFF
--- a/core/gc/Cargo.toml
+++ b/core/gc/Cargo.toml
@@ -22,9 +22,6 @@ either = ["dep:either"]
 # Enable default implementations of trace and finalize for the arrayvec crate
 arrayvec = ["dep:arrayvec"]
 
-experimental_mark_sweep = ["oscars", "oscars/mark_sweep"]
-experimental_mark_sweep2 = ["oscars", "oscars/mark_sweep2"]
-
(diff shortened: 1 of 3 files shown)


core/gc/Cargo.toml
```diff
@@ -22,6 +22,11 @@ either = ["dep:either"]
 # Enable default implementations of trace and finalize for the arrayvec crate
 arrayvec = ["dep:arrayvec"]
 
+oscars = ["dep:oscars", "unstable"]
+
+# Unstable feature
+unstable = []
```
jedel1043, line 28: Why the unstable feature? Is it not enough to have the oscars feature to gate the new GC?

nekevss: If we are adding an `oscars` feature flag. I'd also prefer it comes with the unstable feature flag so that if someone were to enable it, we would also be clear via flagging that the feature itself is not currently stable.

At least, that's generally my thought process on it. But also, after asking the question on Matrix. I'd probably be fine with not moving forward with this and moving to a branch.

jedel1043:
Closing since we decided to use a branch instead.

PR #5116, opened 2026-03-17
fix(engine): Prefer `js_error!` over `JsError::from_opaque`

p0lyw0lf:
In a few places, native code was throwing
`JsError::from_opaque(js_string!("some literal"))`. This is allowed, but it is not exactly standard: most users will expect thrown values to be of type `Error`.

So, where possible, let's convert code using that to instead use the provided `js_error!` macro, which does generate proper `Error` objects.

This doesn't touch `AbortError`, because (1) that's a larger code change, and (2) it might actually be relied on as a sentinel value.

### DIFF
--- a/core/engine/src/interop/into_js_function_impls.rs
+++ b/core/engine/src/interop/into_js_function_impls.rs
@@ -3,7 +3,7 @@
 use super::private::IntoJsFunctionSealed;
 use super::{IntoJsFunctionCopied, UnsafeIntoJsFunction};
 use crate::interop::{JsRest, TryFromJsArgument};
-use crate::{Context, JsError, NativeFunction, TryIntoJsResult, js_string};
+use crate::{Context, NativeFunction, TryIntoJsResult};
 use std::cell::RefCell;
 
 /// A token to represent the context argument in the function signature.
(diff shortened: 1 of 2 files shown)


core/engine/src/interop/into_js_function_impls.rs
```diff
@@ -85,7 +85,7 @@ macro_rules! impl_into_js_function {
                         match s.try_borrow_mut() {
                             Ok(mut r) => r( $($id,)* rest.into() ).try_into_js_result(ctx),
                             Err(_) => {
-                                Err(JsError::from_opaque(js_string!("recursive calls to this function not supported").into()))
+                                Err($crate::js_error!(ReferenceError: "recursive calls to this function not supported"))
```
jedel1043: These should also be plain `Error`s, right?

jedel1043 (changes requested):
Looks good, just a couple of nits

p0lyw0lf:
Hi, sorry for the delay on this! I've addressed the nit & rebased past the merge conflict.

jedel1043 (approved):
Looks good! Thank you!

PR #5119, opened 2026-03-17
normalize header values in Headers

### LINKED ISSUE #5114: Headers does not normalize header values
`Headers` is currently keeping leading and trailing HTTP whitespace in header values instead of normalizing it.

Quick repro:

```js
const headers = new Headers([["x", "  a\tb  "]]);
headers.get("x");
// Boa returns: "  a\tb  "
// Expected: "a\tb"
```

Same thing also happens through `append()` and `set()`.

The relevant code path is in [core/runtime/src/fetch/headers.rs](https://github.com/boa-dev/boa/blob/main/core/runtime/src/fetch/headers.rs).

Right now `to_header_value()` just parses the string into an `http::HeaderValue`, but it does not do the Fetch value normalization step before storing it.

There is already local WPT coverage for this in:
- `tests_wpt/fetch/api/headers/headers-normalize.any.js`

That file expects cases like:
- `" space "` -> `"space"`
- `"\ttab\t"` -> `"tab"`
- `"\r\n newLine"` -> `"newLine"`

Spec reference:
https://fetch.spec.whatwg.org/#concept-header-value-normalize

So this looks like a small fetch runtime bug with a pretty local fix surface.

Monti-27:
Fixes #5114.

This normalizes header values before storing them in `Headers`, so the constructor, `append()`, and `set()` all trim leading and trailing HTTP whitespace the same way.

Tests:
- `cargo test -p boa_runtime headers --lib`

### DIFF
--- a/core/runtime/src/fetch/headers.rs
+++ b/core/runtime/src/fetch/headers.rs
@@ -29,25 +29,10 @@ fn to_header_name(key: impl AsRef<str>) -> JsResult<HeaderName> {
         .map_err(|_| js_error!("Cannot convert key to header string as it is not valid ASCII."))
 }
 
-/// Converts a JavaScript string to a valid header value (or error).
-///
-/// # Errors
-/// If the value is not valid ASCII, an error is returned.
+/// Trims leading and trailing HTTP whitespace from a header value.
 #[inline]
 fn normalize_header_value(value: &str) -> &str {
-    let bytes = value.as_bytes();
-
-    let mut start = 0;
-    while start < bytes.len() && matches!(bytes[start], b'\t' | b'\n' | b'\r' | b' ') {
-        start += 1;
-    }
-
-    let mut end = bytes.len();
-    while end > start && matches!(bytes[end - 1], b'\t' | b'\n' | b'\r' | b' ') {
-        end -= 1;
-    }
-
-    &value[start..end]
+    value.trim_matches(|ch| matches!(ch, '\t' | '\n' | '\r' | ' '))
 }
 
 /// Converts a JavaScript string to a valid header value (or error).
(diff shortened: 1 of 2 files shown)


core/runtime/src/fetch/headers.rs
```diff
@@ -29,14 +29,34 @@ fn to_header_name(key: impl AsRef<str>) -> JsResult<HeaderName> {
...
+        start += 1;
+    }
+
+    let mut end = bytes.len();
+    while end > start && matches!(bytes[end - 1], b'\t' | b'\n' | b'\r' | b' ') {
+        end -= 1;
+    }
+
+    &value[start..end]
+}
```
jedel1043, line 36: It might be better to replace this with [`str::trim_matches`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim_matches)

jedel1043 (changes requested):

core/runtime/src/fetch/headers.rs
Monti-27: yeah makes sense updated this to use trim_matches with the same HTTP whitespace set so the behavior stays the same and is cleaner now :)

core/runtime/src/fetch/headers.rs
```diff
@@ -29,14 +29,19 @@ fn to_header_name(key: impl AsRef<str>) -> JsResult<HeaderName> {
         .map_err(|_| js_error!("Cannot convert key to header string as it is not valid ASCII."))
 }
 
+/// Trims leading and trailing HTTP whitespace from a header value.
+#[inline]
+fn normalize_header_value(value: &str) -> &str {
+    value.trim_matches(|ch| matches!(ch, '\t' | '\n' | '\r' | ' '))
```
jedel1043: No need to use a closure, you can pass an array of all the chars and it should still work

Monti-27: nice catch mb

jedel1043 (approved):
Cool, looks great!

PR #5130, opened 2026-03-18
feat: implement using/await using declarations with SuppressedError

xcb3d:
It changes the following:

- **Implement `using` and `await using` declaration parsing** with proper guards for line terminators (`using\nlet` triggers ASI), bracket tokens (`using[0]` is element access), and `for (using of of arr)` disambiguation
- **Add `SuppressedError` built-in constructor** (`core/engine/src/builtins/error/suppressed.rs`) with full prototype chain inheriting from `Error`, registered as a global binding
- **Add disposal VM opcodes** (`CreateDisposeCapability`, `AddDisposableResource`, `DisposeResources`) implementing the spec's `DisposeResources` abstract operation with proper error chaining via `SuppressedError`
- **Fix `DisposeResources` to incorporate pending exception** as the initial completion per spec, so body errors and disposal errors are properly combined in the `SuppressedError` chain
- **Add `IterableLoopInitializer::Using`** variant for `for-of` loops with immutable binding semantics
- **Block `using` declarations** in switch clause bodies and top-level scripts per spec early errors


jedel1043:
Duplicate of #5079

PR #5131, opened 2026-03-18
make TextEncoder utf-8 only

### LINKED ISSUE #5120: TextEncoder constructor accepts and uses non-UTF-8 encodings
Right now `TextEncoder` in `boa_runtime` is treating the constructor argument like a real encoding selector.

Quick repro:

```js
const enc = new TextEncoder("utf-16");
enc.encoding;                 // boa: "utf-16le"
Array.from(enc.encode("A")); // boa: [65, 0]
```

That should not happen. `TextEncoder` is supposed to stay UTF-8.

The current implementation is in `core/runtime/src/text/mod.rs`:
- `TextEncoder::constructor()` accepts `utf-16`, `utf-16le`, and `utf-16be`
- `TextEncoder::encode()` then emits UTF-16 bytes based on that choice

There is also a local runtime test in `core/runtime/src/text/tests.rs` that is currently locking this behavior in:
- `roundtrip()` includes `utf-16`, `utf-16le`, and `utf-16be`

WPT already expects the opposite in `tests_wpt/encoding/textencoder-constructor-non-utf.any.js`:
- `new TextEncoder(encoding.name).encoding` should still be `"utf-8"`

So this looks like a real spec mismatch, not just missing coverage.

Spec side:
- the Encoding API keeps `TextEncoder` fixed to UTF-8

I can work on this if that sounds good.

Monti-27:
Fixes #5120.

This makes `TextEncoder` UTF-8-only.

It now ignores passed encoding labels, always reports `utf-8`, and always encodes with UTF-8.

This also updates the local runtime tests and enables the `encoding/textencoder-constructor-non-utf.any.js` WPT file.

### DIFF
--- a/core/runtime/src/text/encodings.rs
+++ b/core/runtime/src/text/encodings.rs
@@ -22,15 +22,7 @@ pub(crate) mod utf8 {
 }
 
 pub(crate) mod utf16le {
-    use boa_engine::string::JsStrVariant;
     use boa_engine::{JsString, js_string};
-
-    pub(crate) fn encode(input: &JsString) -> Vec<u8> {
-        match input.as_str().variant() {
-            JsStrVariant::Latin1(l) => l.iter().flat_map(|c| [*c, 0]).collect(),
-            JsStrVariant::Utf16(s) => bytemuck::cast_slice(s).to_vec(),
-        }
-    }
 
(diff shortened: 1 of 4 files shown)


jedel1043 (approved):
Good catch!

PR #5132, opened 2026-03-18
fix Headers.get spacing

### LINKED ISSUE #5113: Headers.get() joins duplicate header values without a space
`Headers.get()` is currently combining duplicate header values with `","` instead of `", "`.

Quick repro:

```js
const headers = new Headers([["x", "1"], ["x", "2"]]);
headers.get("x");
// Boa returns: "1,2"
// Expected: "1, 2"
```

The current code path is in [core/runtime/src/fetch/headers.rs](https://github.com/boa-dev/boa/blob/main/core/runtime/src/fetch/headers.rs):

```rust
.fold(None, |mut acc, v| {
    let str = acc.get_or_insert_with(String::new);
    if !str.is_empty() {
        str.push(',');
    }
    str.push_str(v);
    acc
});
```

That ends up returning `"1,2"` for duplicate values.

This should use `", "` when combining values. The Fetch spec's header-list get algorithm combines values with comma + space:
https://fetch.spec.whatwg.org/#concept-header-list-get

There is also local WPT coverage that expects the spaced form:
- `tests_wpt/fetch/api/basic/header-value-combining.any.js`
- `tests_wpt/fetch/content-length/api-and-duplicate-headers.any.js`

So this looks like a small fetch runtime bug with a pretty small patch surface.

Monti-27:
Fixes #5113.

This makes `Headers.get()` join duplicate header values with `, ` instead of `,`.

It also adds runtime tests for both direct `Headers` instances and fetched response headers.

### DIFF
--- a/core/runtime/src/fetch/headers.rs
+++ b/core/runtime/src/fetch/headers.rs
@@ -208,13 +208,10 @@ impl JsHeaders {
             .get_all(name.clone())
             .into_iter()
             .map(|v| v.to_str().unwrap_or(""))
-            // Use an Option<String> to accumulate the values into a single string,
-            // if there are any. Otherwise, we return None.
-            // Cannot use `join(",")` as we need to return undefined if none is found.
             .fold(None, |mut acc, v| {
                 let str = acc.get_or_insert_with(String::new);
                 if !str.is_empty() {
-                    str.push(',');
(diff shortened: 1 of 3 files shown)


jedel1043 (approved):
Thanks!

PR #5133, opened 2026-03-18
test: add coverage for JsTypedArray toReversed, toSorted, toLocaleString

### LINKED ISSUE #3252: Tracking issue for implemented methods on `JsTypedArray`

Xavrir:
## Context

These three `JsTypedArray` wrapper methods are fully implemented but had no test coverage. This PR adds unit tests for each.

Ref #3252 -- the tracking issue checkboxes for `toLocaleString`, `toReversed`, and `toSorted` can be checked off, since the implementations already exist in both the builtins layer and the Rust wrapper.

## Tests added

- **`typed_array_to_reversed`** -- verifies `toReversed()` returns a new reversed array and leaves the original unchanged
- **`typed_array_to_sorted`** -- verifies `toSorted()` returns a new sorted array and leaves the original unchanged
- **`typed_array_to_locale_string`** -- verifies `toLocaleString()` returns a string containing the element values

All three tests pass locally: `cargo test --package boa_engine typed_array_to_`

### DIFF
--- a/core/engine/src/object/builtins/jstypedarray.rs
+++ b/core/engine/src/object/builtins/jstypedarray.rs
@@ -1,14 +1,14 @@
 //! Rust API wrappers for the `TypedArray` Builtin ECMAScript Objects
 use crate::{
-    Context, JsExpect, JsResult, JsString, JsValue,
     builtins::{
-        BuiltInConstructor,
         array_buffer::AlignedVec,
         typed_array::{BuiltinTypedArray, TypedArray, TypedArrayKind},
+        BuiltInConstructor,
     },
     error::JsNativeError,
     object::{JsArrayBuffer, JsFunction, JsObject, JsSharedArrayBuffer},
     value::{IntoOrUndefined, TryFromJs},
+    Context, JsExpect, JsResult, JsString, JsValue,
 };
 use boa_gc::{Finalize, Trace};
 use std::ops::Deref;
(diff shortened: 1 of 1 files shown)


jedel1043 (approved):
Thanks!

Xavrir:
no prob! happy to contribute

core/engine/src/object/builtins/jstypedarray.rs
```diff
@@ -1513,40 +1513,53 @@ fn typed_array_values() {
 }
 
 #[test]
-fn typed_array_iterator() {
+fn typed_array_to_reversed() {
```
jedel1043, line 1554: I think you deleted the `typed_array_iterator` test. Can you add that back?

jedel1043 (changes requested):

core/engine/src/object/builtins/jstypedarray.rs
Xavrir: So sorry about that, adding it back asap

Xavrir: Fixed in 375b13c5. I restored `typed_array_iterator` and reran `cargo test -p boa_engine typed_array_` plus `cargo make run-ci` locally.

jedel1043 (approved):
Cool, thanks

PR #5136, opened 2026-03-18
Enable tracing for property sets in VM (__set__ calls)

Maanvi212006:
This Pull Request introduces logging for property sets in the JS engine.

It changes the following:

- Adds `[TRACE] SET` logging in `__set__` for `JsObject` properties.
- Prints both key and value whenever a property is set, helping debug assignment behavior.
- Minimal code changes: only touches `core/engine/src/object/internal_methods/mod.rs`.

## Screenshot
The screenshot demonstrates the new [TRACE] SET logging added in the JS engine.
-A JavaScript object x is created: let x = {};.
-A property b is assigned the value 20: x.b = 20;.
-The [TRACE] SET log prints the property key ("b") and the value being assigned (20) twice, showing the internal call flow.

The final 20 shows that the assignment completes successfully and the value is returned as expected.

This clearly illustrates that the new logging correctly captures property set operations in the VM without affecting normal execution.

boa 2


core/engine/src/object/internal_methods/mod.rs
```diff
@@ -281,15 +281,19 @@ impl JsObject {
...
+    &self,
+    key: PropertyKey,
+    value: JsValue,
+    receiver: JsValue,
+    context: &mut InternalMethodPropertyContext<'_>,
+) -> JsResult<bool> {
+    let result = (self.vtable().__set__)(self, key.clone(), value.clone(), receiver.clone(), context);
+    #[cfg(feature = "trace")]
+    if let Ok(true) = result {
+        println!("[TRACE] SET -> key: {:?}, value: {:?}", key, value);
```
jedel1043: It's honestly fine to use `tracing` here. We just need to adjust our features so that `tracing` is disabled if the `trace` feature is not set

jedel1043: Maybe this is useful? https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters

Maanvi212006: Hi @jedel1043 ,

Thank you for your feedback. I’ve carefully reviewed your suggestions and made the following changes:

1. Trace logging updated: Replaced manual println! traces with tracing::trace! macros for proper compile-time and runtime filtering.
2. Cloning key/value for safe ownership: Fixed move/borrow issues by cloning key and value only when necessary to avoid multiple moves.
3. ordinary_set cleanup: Removed duplicate calls and unnecessary code paths, ensuring proper handling of data descriptors and accessor descriptors.
4. Feature-gated tracing: All debug output now respects the "trace" feature flag and does not affect normal execution.
5. Unused variables and imports addressed: Prefixed _context and removed unused imports to clean warnings.

Issue #4123, opened 2025-01-12
perf: use cow-utils instead


heygsc:
We can improve performance by using `cow_utils`, and automatically check in clippy.

We have done similar work before, and if necessary and possible, I would like to provide pr for boa.

Related issue: https://github.com/oxc-project/oxc/issues/5586

jasonwilliams:
Hey @heygsc 

That sounds great, we would be happy to accept a PR for this, I will assign this issue to you if you're happy to work on it.
CC @CrazyboyQCD who has done [similar work](https://github.com/boa-dev/boa/pull/4020) on this recently

PR #5141, opened 2026-03-18
fix(engine): replace str::replace with cow_replace in module loader

### LINKED ISSUE #5140: Remaining `str::replace` usage flagged by clippy::disallowed_methods in module loader
While running `cargo make run-ci`, clippy reports a `disallowed_methods` warning for:

core/engine/src/module/loader/mod.rs:65

    let specifier = specifier.replace('/', "\\");

Since #4123 migrated similar cases to `cow_utils::CowUtils::cow_replace` to avoid unnecessary allocations, this looks like a remaining instance.

I’d like to send a small PR replacing this with `cow_replace`.

iammdzaidalam:
Closes #5140

Replaces `str::replace` with `cow_utils::CowUtils::cow_replace` in the module loader to satisfy `clippy::disallowed_methods` and avoid unnecessary allocation.

Follows the pattern introduced in #4123.

### DIFF
--- a/core/engine/src/module/loader/mod.rs
+++ b/core/engine/src/module/loader/mod.rs
@@ -3,8 +3,6 @@ use std::cell::RefCell;
 use std::path::{Component, Path, PathBuf};
 use std::rc::Rc;
 
-use cow_utils::CowUtils;
-
 use dynify::dynify;
 use rustc_hash::FxHashMap;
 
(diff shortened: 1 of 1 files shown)


core/engine/src/module/loader/mod.rs
```diff
@@ -62,7 +64,7 @@ pub fn resolve_module_specifier(
 
     // On Windows, also replace `/` with `\`. JavaScript imports use `/` as path separator.
     #[cfg(target_family = "windows")]
-    let specifier = specifier.replace('/', "\\");
+    let specifier = specifier.cow_replace('/', "\\").into_owned();
```
jedel1043: It would be better to use the trait directly instead of importing
```
cow_utils::CowUtils::cow_replace(specifier, '/', "\\")
```
Also, using `cow_replace` then `into_owned` immediately after is... an interesting decision, because it's basically nullifying the use of `cow_replace`.

jedel1043 (approved):
LGTM

core/engine/src/module/loader/mod.rs
iammdzaidalam: Thanks, resolved 👍

PR #5145, opened 2026-03-19
perf(json): use FxHashSet for cycle detection in JSON.stringify

iammdzaidalam:
## Description

`JSON.stringify` currently performs cycle detection using linear scans over the active stack.

This replaces those checks with an `FxHashSet`, while keeping the existing traversal order in a `Vec`.

`StateRecord` now stores:
- `stack: Vec<JsObject>`
- `stack_set: FxHashSet<JsObject>`

Both are updated together in `serialize_json_object` and `serialize_json_array`.

Behavior remains unchanged, only the lookup strategy is different.

## Changes

- use `FxHashSet` for cycle detection
- keep `Vec` for order
- ensure both structures stay in sync during push/pop

## Tests

Added tests for:
- cyclic object
- cyclic array
- nested cyclic structures

## Notes

Ran local checks (`fmt`, `check`, `clippy`, `tests`).

Tried running `cargo bench -p boa_benches`, but the suite hit a stack overflow in `v8-benches/deltablue`, so not using it as validation for this change.

### DIFF
--- a/core/engine/src/builtins/json/mod.rs
+++ b/core/engine/src/builtins/json/mod.rs
@@ -625,13 +625,10 @@ impl Json {
         args: &[JsValue],
         context: &mut Context,
     ) -> JsResult<JsValue> {
-        // 1. Let stack be a new empty List.
-        let stack = Vec::new();
-
-        // 2. Let indent be the empty String.
+        // 1. Let indent be the empty String.
         let indent = js_string!();
 
-        // 3. Let PropertyList and ReplacerFunction be undefined.
+        // 2. Let PropertyList and ReplacerFunction be undefined.
         let mut property_list = None;
         let mut replacer_function = None;
 
@@ -749,7 +746,6 @@ impl Json {
         // 11. Let state be the Record { [[ReplacerFunction]]: ReplacerFunction, [[Stack]]: stack, [[Indent]]: indent, [[Gap]]: gap, [[PropertyList]]: PropertyList }.
         let mut state = StateRecord {
             replacer_function,
-            stack,
             stack_set: FxHashSet::default(),
             indent,
             gap,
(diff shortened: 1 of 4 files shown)


iammdzaidalam:
Hi @jedel1043, small perf change for JSON cycle detection, would appreciate a review.

zhuzhu81998:
> Tried running `cargo bench -p boa_benches`, but the suite hit a stack overflow in `v8-benches/deltablue`, so not using it as validation for this change.

a simple test that actually uses `JSON.stringify` and measures the time would be better than nothing.
not sure if the current benches use json functions at all.

iammdzaidalam:
Yeah, that makes sense.

If `boa_benches` is not exercising `JSON.stringify` directly, it is not a very useful signal for this change. I'll add a small focused benchmark for it.

iammdzaidalam:
Added focused `JSON.stringify` benchmarks for this path:
- `stringify_deep.js` (deep acyclic case)
- `stringify_circular.js` (cycle detection path)

I tried running them with `cargo bench -p boa_benches -- json/...`, but the current bench harness hits a pre-existing stack overflow before the filter takes effect. It looks like all script benches are initialized first, so unrelated `v8-benches` entries fail before reaching these.

Left that as a separate follow-up rather than expanding this PR.

zhuzhu81998:
> Added focused `JSON.stringify` benchmarks for this path:
> 
>     * `stringify_deep.js` (deep acyclic case)
> 
>     * `stringify_circular.js` (cycle detection path)
> 
> 
> I tried running them with `cargo bench -p boa_benches -- json/...`, but the current bench harness hits a pre-existing stack overflow before the filter takes effect. It looks like all script benches are initialized first, so unrelated `v8-benches` entries fail before reaching these.
> 
> Left that as a separate follow-up rather than expanding this PR.

then dont use the cargo bench. run the script manually and time it (remember to add a loop though). and compare the time between pr and main.

iammdzaidalam:
yeah, fair. will just time it manually with a loop on main vs this PR and post numbers.

iammdzaidalam:
Ran the scripts manually through `boa_cli` instead of `cargo bench`, using the same files on both `main` and this branch and timing the built binary directly.

On my machine (after warmup), timings were roughly:

Deep case:
- PR: ~14.3–17.4 ms
- main: ~15.5–18.3 ms

Circular case:
- PR: ~14.5–16.8 ms
- main: ~14.4–17.1 ms

The deep acyclic case shows a small improvement on this branch, while the circular case is roughly flat and within noise. This is just a quick manual check, not a rigorous benchmark.

core/engine/src/builtins/json/mod.rs
```diff
@@ -965,127 +968,139 @@ impl Json {
...
-        state.stack.pop();
+        let result = inner();
 
-        // 12. Set state.[[Indent]] to stepback.
-        state.indent = stepback;
+        // 11. Remove the last element of state.[[Stack]].
+        let popped = state
+            .stack
+            .pop()
+            .expect("json stringify stack should not be empty");
```
jedel1043: Do you need to preserve the old stack? AIUI, the stack is only used to check for duplicates, so what you can do is to instead call `state.stack_set.remove(&value)` and it should still work fine, right?

jedel1043: That way we can remove the `stack` field

iammdzaidalam: yep, the old stack was only being used for duplicate detection, so I removed it and now just insert/remove from stack_set around the recursive stringify paths.

iammdzaidalam: pushing after running some checks

iammdzaidalam:
### Benchmark

Ran the focused stringify scripts manually through `boa_cli --release` (10 rounds each, after warmup).

#### Deep case (acyclic object)

| Metric | main | PR | Improvement |
|--------|------|----|------------|
| Typical | ~46.6–47.0s | ~42.3–42.7s | ~9–10% faster |
| Min | ~46.6s | ~42.3s | ~9.2% faster |
| Max | ~50.3s | ~46.4s | ~7–8% faster |

#### Circular case (cycle detection)

| Metric | main | PR | Improvement |
|--------|------|----|------------|
| Typical | ~1.69–1.72s | ~1.38–1.39s | ~15–20% faster |
| Min | ~1.69s | ~1.36s | ~19.5% faster |
| Max | ~1.79s | ~1.40s | ~21.8% faster |

### Notes

- measured using the same scripts on both branches
- each run uses 10 rounds after warmup
- numbers are approximate ranges from manual runs (not a rigorous benchmark)

iammdzaidalam:
hey @zhuzhu81998 @jedel1043 , any updates when u get a moment?

jedel1043 (approved):
Thank you! Looks good!

PR #5146, opened 2026-03-19
fix(tester): correct edition mapping for ES16

### LINKED ISSUE #5053: edition.rs lists wrong ES edition for arraybuffer-transfer / set-methods
I've been looking at test262 features-to-edition mapping and noticed a couple of discrepancies:

* `arraybuffer-transfer` is part of ES2024 (ES15):
  https://262.ecma-international.org/15.0/#sec-arraybuffer.prototype.transfer
  https://github.com/boa-dev/boa/blob/main/tests/tester/src/edition.rs#L126
  
* `set-methods` is part of ES2025 (ES16):
   https://262.ecma-international.org/16.0/index.html#sec-set.prototype.union
   https://github.com/boa-dev/boa/blob/main/tests/tester/src/edition.rs#L280

Xavrir:
## Summary
- map `arraybuffer-transfer` to ES15 and `set-methods` to ES16 in the test262 feature table
- add `SpecEdition::ES16` support to tester stats and backward-compatible result deserialization
- add focused tests for the edition mappings and ES16 stats handling

## Testing
- `cargo test -p boa_tester`
- `cargo clippy -p boa_tester --all-features --all-targets`
- pre-push `cargo make run-ci`

Closes #5053


tests/tester/src/edition.rs
```diff
@@ -443,8 +446,57 @@ impl SpecEdition {
...
+    #[test]
+    fn maps_set_methods_to_es16() {
+        let metadata = metadata(&["set-methods"]);
+
+        assert_eq!(
+            SpecEdition::from_test_metadata(&metadata),
+            Ok(SpecEdition::ES16)
+        );
+    }
+}
```
jedel1043: I don't think we need tests for these. Not because tests are bad or anything, but because the editions mapping is so in-flight sometimes that it's not really useful to test if specific features map to a certain version.

Case in point, if we had tests for these, then you wouldn't have been able to change the edition of these two features without changing the tests.

tests/tester/src/main.rs
```diff
@@ -743,6 +752,66 @@ impl AddAssign for VersionedStats {
...
+            .expect("ES16 stats should exist");
+
+        assert_eq!(es16.total, es15.total);
+        assert_eq!(es16.passed, es15.passed);
+        assert_eq!(es16.ignored, es15.ignored);
+        assert_eq!(es16.panic, es15.panic);
+    }
+
+    #[test]
+    fn apply_updates_es16_statistics() {
```
jedel1043: This test doesn't really reflect how we keep count of tests; usually if ES6 has a test, that test also counts for the total in ES7 and onwards, so it doesn't make sense to have ES15 with a lower test count than ES16

jedel1043 (changes requested):

tests/tester/src/main.rs
Xavrir: Removed the extra ES16 stats tests in 50e871f0 and kept the PR focused on the edition fix plus the required tester plumbing.

tests/tester/src/edition.rs
Xavrir: Got it, I removed those mapping tests in 50e871f0. You’re right that pinning exact feature editions there would just create churn.

jedel1043 (approved):
Thank you for the fix!

PR #5147, opened 2026-03-19
docs(engine): add spec references for bytecode internals

### LINKED ISSUE #2426: Document opcode and bytecompiler spec references
In the builtins and in the parser we document our code with direct references and steps of the ECMAScript specification. This makes it easy to understand how the specification is implemented and why some code is written the way it is.
Currently this kind of documentation is missing for our opcodes and from the bytecompiler. We should add the documentation and spec references, especially in places where the spec expects operations to not fail (`!`) which is often a panic in our code.

Xavrir:
## Summary
- add module-level spec references for the bytecompiler and opcode layers
- document how spec `!` operations map to Boa's internal bytecode invariants
- clarify operand decoding assumptions in the opcode argument helpers

## Testing
- `cargo fmt --all --check`
- `cargo check -p boa_engine --all-features`
- pre-push `cargo make run-ci`

Closes #2426

### DIFF
--- a/core/engine/src/bytecompiler/expression/mod.rs
+++ b/core/engine/src/bytecompiler/expression/mod.rs
@@ -1,3 +1,12 @@
+//! Lowering for ECMAScript expressions.
+//!
+//! The functions in this module implement the bytecompiler side of expression evaluation. They
+//! preserve the source language evaluation order while translating AST nodes into opcode sequences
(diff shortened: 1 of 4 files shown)


jedel1043 (approved):
Cool! It's nice to have this explained for new contributors.

PR #5149, opened 2026-03-19
refactor(engine): encapsulate Context VM field access behind wrapper methods

hansl:
Add `context/vm_access.rs` with 45 wrapper methods that delegate to `Context`'s internal `Vm` field, then replace all ~600 direct `context.vm.*` accesses across 78 files with calls to these wrappers.

This is the first step toward extracting Context into its own crate (boa_context). By routing all external access through methods, the internal structure of Context can be changed without touching call sites.


core/engine/src/builtins/function/mod.rs
```diff
@@ -653,7 +653,7 @@ impl BuiltInFunctionObject {
             return Err(js_error!(SyntaxError: "failed to analyze function scope: {}", reason));
         }
 
-        let in_with = context.vm.frame().environments.has_object_environment();
+        let in_with = context.frame().environments.has_object_environment();
```
jedel1043, line 656: I think this shows we would also need to extract the environment related types, but those depend on `JsObject`, so...

Maybe we should start by extracting `JsObject` first?

hansl: And `JsObject` depends on `Context`.

Regardless, this PR would still be needed. And it's a simple refactor, there should be no behaviour changes. I can do a follow to address the environment types.

jedel1043: Yeah this doesn't block merging this, but we also need to think what exactly are the steps to properly extract this into its own crate

core/engine/src/context/vm_access.rs
```diff
@@ -0,0 +1,296 @@
...
+
+impl Context {
+    /// Push a value onto the VM stack.
+    pub(crate) fn stack_push<T: Into<JsValue>>(&mut self, value: T) {
+        self.vm.stack.push(value);
+    }
+
+    /// Pop a value from the VM stack.
+    #[track_caller]
+    pub(crate) fn stack_pop(&mut self) -> JsValue {
```
jedel1043, line 198: Do we need to have individual methods to pop and push from the stack here? I would expect that just returning `&mut Stack` should be enough

jedel1043: It's the same approach we follow for the EnvironmentStack, so it makes sense to also use the same approach here.

core/engine/src/context/vm_access.rs
```diff
@@ -0,0 +1,296 @@
...
+            .stack
+            .calling_convention_get_function(argument_count)
+    }
+}
+
+// === Runtime Limits ===
+
+impl Context {
+    /// Get the loop iteration limit.
+    pub(crate) fn loop_iteration_limit(&self) -> u64 {
```
jedel1043, line 284: Same suggestion here. We should return `&RuntimeLimits` instead.

jedel1043 (changes requested):

jedel1043:
Closing this since it got stale, but it will be good reference for when we finally migrate the Context type to its own crate

PR #5153, opened 2026-03-19
ci: set fail-fast to false on tests matrix to prevent platform blind spots

### LINKED ISSUE #5152: CI: set `fail-fast: false` on the tests matrix to prevent platform blind spots
I think we should improve the developer experience and testing feedback loop by explicitly adding `fail-fast: false` to our multi-OS testing matrix in [.github/workflows/rust.yml](cci:7://file:///c:/Users/risha/Desktop/boa/.github/workflows/rust.yml:0:0-0:0).

### The Problem
Currently, our `tests` job in [rust.yml](cci:7://file:///c:/Users/risha/Desktop/boa/.github/workflows/rust.yml:0:0-0:0) runs a parallel execution matrix across four different platforms (`macos-14`, `windows-latest`, `ubuntu-24.04-arm`, and `ubuntu-latest`). 

Because we haven't explicitly configured the `fail-fast` value inside the `strategy` block, GitHub Actions defaults to `fail-fast: true`. This means that if a developer pushes a bug that breaks only on Windows, but the `ubuntu-latest` runner happens to finish compiling and failing first, GitHub will **instantly force-cancel** the other three platforms mid-run. 

This creates massive blind spots: developers lose their test results for the other platforms and have no idea if their bug is platform-specific or completely universal until they push a second fix.

### The Fix
We should simply append `fail-fast: false` underneath the `strategy` block for the `tests` job.
```yaml
    strategy:
      fail-fast: false
      matrix:
        include:
          - os: macos-14
          - os: windows-latest
          - os: ubuntu-24.04-arm
          - os: ubuntu-latest

RishavTiwari25:
This Pull Request fixes/closes #5152 .

It changes the following:

- Appends the explicit `fail-fast: false` property to the `tests` matrix strategy block within [.github/workflows/rust.yml](cci:7://file:///c:/Users/risha/Desktop/boa/.github/workflows/rust.yml:0:0-0:0).
- Enhances developer diagnostic visibility by guaranteeing that every OS environment (`macos`, `windows`, `ubuntu`, `ubuntu-arm`) successfully completes its full test suite execution, even if a sibling platform fails first.
- Prevents GitHub Actions' default behavior from instantly force-cancelling the entire pipeline, eliminating "platform blind spots" where developers are unaware of platform-specific regressions.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -207,6 +207,7 @@ jobs:
     env:
       RUSTUP_WINDOWS_PATH_ADD_BIN: 1
     strategy:
+      fail-fast: false
       matrix:
         include:
           - os: macos-14


jedel1043 (approved):
Cool

PR #5156, opened 2026-03-19
refactor(builtins): replace manual loop with iterator in String.fromCharCode

Nakshatra480:
## Summary
Replaces the verbose manual `Vec` initialization and `for` loop in 
`String.fromCharCode` with a cleaner, idiomatic Rust iterator chain.

## Changes
- `core/engine/src/builtins/string/mod.rs`: Replaced manual loop with 
  `.iter().map().collect::<JsResult<Vec<_>>>()?`

## Before
```rust
let mut result = Vec::with_capacity(args.len());
for next in args {
    let next_cu = next.to_uint16(context)?;
    result.push(next_cu);
}
```

## After
```rust
let result = args
    .iter()
    .map(|next| next.to_uint16(context))
    .collect::<JsResult<Vec<_>>>()?;
```

## Notes
- No observable behavior change
- All existing `cargo test -p boa_engine` tests pass (183 passed, 0 failed)
- `cargo clippy` reports no new warnings
- `JsResult` error bubbling is preserved correctly via `.collect()`

## Spec Reference
- [String.fromCharCode — MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode)
- ECMAScript §22.1.2.1

### DIFF
--- a/core/engine/src/builtins/string/mod.rs
+++ b/core/engine/src/builtins/string/mod.rs
@@ -440,16 +440,13 @@ impl String {
         context: &mut Context,
     ) -> JsResult<JsValue> {
         // 1. Let result be the empty String.
-        let mut result = Vec::with_capacity(args.len());
-
         // 2. For each element next of codeUnits, do
-        for next in args {
-            // a. Let nextCU be the code unit whose numeric value is ℝ(? ToUint16(next)).
-            let next_cu = next.to_uint16(context)?;
-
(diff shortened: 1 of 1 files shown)


jedel1043:
Neither style looks good or bad, and I disagree that iterators are inherently more "idiomatic" than manual looping, so I would prefer that we keep the current code.

PR #5159, opened 2026-03-19
fix(ci): harden test262 PR workflow security

### LINKED ISSUE #5158: test262_pr workflow runs untrusted code under pull_request_target
The `test262_pr.yml` workflow currently uses `pull_request_target` while checking out and executing the PR's head commit:

- uses `pull_request_target`
- checks out `${{ github.event.pull_request.head.sha }}`
- runs `cargo run --release --bin boa_tester`

This effectively executes untrusted code from the PR with the permissions of the base repository.

Even though the file includes a warning comment about this, the workflow still performs the unsafe pattern. A malicious PR could exploit this to access secrets or perform unintended actions.

Suggested fix:

Split the workflow into two parts:

1. A `pull_request` workflow that runs tests on the PR code without elevated permissions
2. A follow-up `workflow_run` or similar trusted job that handles commenting/reporting

Alternatively, avoid using `pull_request_target` for any job that executes PR code directly.

This seems distinct from existing CI work, as it focuses specifically on unsafe execution patterns in workflows rather than caching, fail-fast, or CodeQL-related changes.

iammdzaidalam:
Closes #5158.

This replaces the old `test262_pr.yml` flow with a safer two-phase setup:

- `test262.yml`: runs under `pull_request` with minimal permissions and executes the test262 suite on PR code
- `test262_comment.yml`: runs under `workflow_run` in the base repository context and updates the PR comment from an uploaded artifact

Main changes:
- remove untrusted execution from `pull_request_target`
- disable persisted checkout credentials in the run workflow
- keep the trusted comment workflow isolated from PR code
- pass results through an artifact and use `body-path` for comment updates

This preserves the reporting behavior while avoiding execution of PR code in a privileged workflow context.


jedel1043 (approved):
Cool, thanks

PR #5162, opened 2026-03-19
fix TextDecoder label parsing

### LINKED ISSUE #5135: TextDecoder constructor does not normalize encoding labels
`TextDecoder` is currently matching encoding labels too strictly.

Right now these fail:

```js
new TextDecoder("UTF-8");
new TextDecoder(" utf-16 ");
new TextDecoder("\nutf-8\t");
```

Boa throws a `RangeError` for those, but plain lowercase labels like `new TextDecoder("utf-16")` work.

This looks like the constructor is matching the raw string directly in `core/runtime/src/text/mod.rs` instead of normalizing the label first.

The local WPT file already expects this to work:
- `tests_wpt/encoding/textdecoder-labels.any.js`

That test expects label matching to be:
- ASCII case insensitive
- trimmed for surrounding ASCII whitespace (`' '`, `\t`, `\n`, `\f`, `\r`)

So this seems like a separate issue from the existing `TextDecoder` bugs:
- `#4612` utf-16 decoding correctness
- `#4614` `fatal` handling
- `#4767` `DataView` bounds

This one is specifically about constructor label normalization.

Monti-27:
Fixes #5135.

This makes `TextDecoder` normalize supported encoding labels before matching them.

It now accepts supported labels with ASCII case differences, surrounding ASCII whitespace, and the supported aliases for the UTF-8 and UTF-16 encodings Boa currently implements.

Tests:
- `cargo test -p boa_runtime text --lib`

### DIFF
--- a/core/runtime/src/text/mod.rs
+++ b/core/runtime/src/text/mod.rs
@@ -34,6 +34,51 @@ pub enum Encoding {
     Utf16Be,
 }
 
+const UTF_8_LABELS: &[&str] = &[
+    "unicode-1-1-utf-8",
+    "unicode11utf8",
+    "unicode20utf8",
+    "utf-8",
+    "utf8",
+    "x-unicode20utf8",
+];
+
+const UTF_16BE_LABELS: &[&str] = &["unicodefffe", "utf-16be"];
+
+const UTF_16LE_LABELS: &[&str] = &[
+    "csunicode",
+    "iso-10646-ucs-2",
+    "ucs-2",
+    "unicode",
+    "unicodefeff",
+    "utf-16",
+    "utf-16le",
+];
+
+#[inline]
+fn resolve_text_decoder_label(label: &str) -> Option<Encoding> {
+    let label = label.trim_matches(['\u{0009}', '\u{000A}', '\u{000C}', '\u{000D}', '\u{0020}']);
+
+    if UTF_8_LABELS
+        .iter()
+        .any(|supported| label.eq_ignore_ascii_case(supported))
+    {
+        Some(Encoding::Utf8)
+    } else if UTF_16LE_LABELS
+        .iter()
+        .any(|supported| label.eq_ignore_ascii_case(supported))
+    {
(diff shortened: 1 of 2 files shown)


core/runtime/src/text/mod.rs
```diff
@@ -34,6 +34,51 @@ pub enum Encoding {
...
+
+const UTF_16LE_LABELS: &[&str] = &[
+    "csunicode",
+    "iso-10646-ucs-2",
+    "ucs-2",
+    "unicode",
+    "unicodefeff",
+    "utf-16",
+    "utf-16le",
+];
```
jedel1043: This might be simplified with an array of tuples `[(&str, Encoding)]`

core/runtime/src/text/tests.rs
```diff
@@ -288,6 +288,66 @@ fn decoder_bom_ignore_bom_false(encoding: &'static str, bytes: &'static [u8]) {
...
+    text::register(None, context).unwrap();
+
+    run_test_actions_with(
+        [
+            TestAction::run(indoc! {r#"
+                threw = false;
+                try {
+                    new TextDecoder(" utf-32 ");
+                } catch (e) {
+                    threw = e instanceof RangeError;
```
jedel1043: No need to manually check if it threw, you can do an `if !(e instanceof RangeError) { throw e }` to reraise the exception and the suite should automatically report that the run failed

jedel1043 (changes requested):
Looks good! I just have a couple of nitpicks

core/runtime/src/text/mod.rs
Monti-27: makes sense i can fold that into one small label table here

core/runtime/src/text/tests.rs
Monti-27: good point

jedel1043 (approved):
Looks great!

Monti-27:
> Looks great!

thanks mentor !!

PR #5164, opened 2026-03-19
ci: tighten workflow permissions and disable checkout credential persistence

### LINKED ISSUE #5163: CI: tighten workflow permissions and disable checkout credential persistence
Some of our GitHub Actions workflows can be tightened a bit from a least-privilege perspective.

A few read-only jobs still use the default `actions/checkout` behavior, which persists credentials in the local git config. Those jobs only need repository read access for checkout/build/test/audit steps, so they can use `persist-credentials: false`.

`nightly_build.yml` also has `contents: write` at the workflow level, but that permission only seems necessary for the `build` job.

## Proposed changes

- add `persist-credentials: false` to checkout steps in:
  - `rust.yml`
  - `pull_request.yml`
  - `webassembly.yml`
  - `security_audit.yml`
  - `nightly_build.yml`
- move `contents: write` in `nightly_build.yml` from the workflow level to the specific job that needs it
- keep workflow defaults read-only where possible

This should just be a hardening cleanup with no intended behavior change.

iammdzaidalam:
Closes #5163

## Summary

Tighten least-privilege settings in a few core CI workflows.

## Changes

- add `persist-credentials: false` to checkout steps in:
  - `rust.yml`
  - `pull_request.yml`
  - `webassembly.yml`
  - `security_audit.yml`
  - `nightly_build.yml`
- move `contents: write` in `nightly_build.yml` from the workflow level to the `build` job
- keep workflow defaults read-only where possible

## Notes

This is just a hardening cleanup with no intended behavior change.

`pull_request.yml` still keeps `fetch-depth: 0`, and `nightly_build.yml` still retains the write permission needed for release asset upload.

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -1,6 +1,6 @@
 name: Nightly Build
 permissions:
-  "contents": "write"
+  contents: read
 
 # Schedule this workflow to run at midnight every day
 on:
@@ -10,6 +10,8 @@ on:
 
 jobs:
   build:
+    permissions:
+      contents: write
     strategy:
       matrix:
         include:
@@ -29,6 +31,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable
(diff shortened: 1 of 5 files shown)


.github/workflows/test262.yml
jedel1043, line 1: Unrelated change, which probably comes from #5159

iammdzaidalam: You’re right, that came from #5159. I’ll rebase this PR to keep only the workflow-permissions changes.

iammdzaidalam: Rebased onto latest `main` and removed the unrelated change from #5159.

This should now only include the workflow permissions hardening. Thanks for pointing that out.

jedel1043 (approved):
Looks good

PR #5165, opened 2026-03-19
ci: harden checkout credential persistence and nightly permissions

### LINKED ISSUE #5163: CI: tighten workflow permissions and disable checkout credential persistence

SergioChan:
## Summary
- sets `persist-credentials: false` on every `actions/checkout` step in:
  - `.github/workflows/rust.yml`
  - `.github/workflows/pull_request.yml`
  - `.github/workflows/webassembly.yml`
  - `.github/workflows/security_audit.yml`
  - `.github/workflows/nightly_build.yml`
- scopes `contents: write` to the `build` job in `nightly_build.yml`
- keeps workflow-level permissions read-only where possible

## Validation
- `git diff --check`
- `python -c "import yaml, pathlib; files=['tmp-boa-5163/.github/workflows/rust.yml','tmp-boa-5163/.github/workflows/pull_request.yml','tmp-boa-5163/.github/workflows/webassembly.yml','tmp-boa-5163/.github/workflows/security_audit.yml','tmp-boa-5163/.github/workflows/nightly_build.yml']; [yaml.safe_load(pathlib.Path(f).read_text(encoding='utf-8')) for f in files]; print('YAML_OK')"`

Closes #5163

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -1,6 +1,6 @@
 name: Nightly Build
 permissions:
-  "contents": "write"
+  contents: read
 
 # Schedule this workflow to run at midnight every day
 on:
@@ -10,6 +10,8 @@ on:
 
 jobs:
   build:
+    permissions:
+      contents: write
     strategy:
       matrix:
         include:
@@ -29,6 +31,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable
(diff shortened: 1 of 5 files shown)


jedel1043:
Duplicate of #5164

PR #5168, opened 2026-03-19
fix(engine): use code_unit_at instead of code_point_at in URI Decode

### LINKED ISSUE #5166: decodeURI/decodeURIComponent return incorrect results for non-BMP characters (emoji)
**Describe the bug**
`decodeURI` and `decodeURIComponent` return incorrect results for non-BMP characters such as emoji (`\uD83D\uDE00`). The URI Decode abstract operation reads a Unicode code point instead of a UTF-16 code unit, but ECMAScript spec step 4.b requires reading a UTF-16 code unit directly at index `k`.

**To Reproduce**
```javascript
decodeURI('\uD83D\uDE00') === '\uD83D\uDE00'
// returns false in Boa, expected true

decodeURIComponent('\uD83D\uDE00') === '\uD83D\uDE00'
// returns false in Boa, expected true
```

Verified in Node.js:
```bash
node -e "console.log(decodeURI('\uD83D\uDE00') === '\uD83D\uDE00')"
# true

node -e "console.log(decodeURIComponent('\uD83D\uDE00') === '\uD83D\uDE00')"
# true
```

**Expected behavior**
Both `decodeURI` and `decodeURIComponent` should return `true` for non-BMP input, matching Node.js and Chrome. ECMAScript [Decode abstract operation](https://tc39.es/ecma262/#sec-decode) step 4.b states the code unit at index `k` should be read, not the code point.

**Build environment:**
- OS: Windows 11
- Version: 10.0.26200

**Additional context**
Root cause is in `core/engine/src/builtins/uri/mod.rs` where `code_point_at` is used instead of `code_unit_at`. A surrogate pair like `\uD83D\uDE00` is two UTF-16 code units, but `code_point_at` returns the combined code point `U+1F600` which then gets truncated. The spec requires processing each code unit individually.

Spec reference: https://tc39.es/ecma262/#sec-decode

HiteshShonak:
This Pull Request fixes/closes #5166.

It changes the following:

- Fixes URI Decode to read UTF-16 code units at index k instead of reading a code point and truncating to u16.
- Corrects `decodeURI` and `decodeURIComponent` behavior for non-BMP characters such as `\uD83D\uDE00`, matching ECMAScript Decode step 4.b and Node/Chrome behavior.
- Adds regression tests to keep non-BMP decode behavior correct.

Testing:

- `cargo test -p boa_engine uri`
- `cargo fmt --all`

Spec reference:
https://tc39.es/ecma262/#sec-decode


core/engine/src/builtins/uri/mod.rs
```diff
@@ -384,7 +384,7 @@ where
         }
 
         // b. Let C be the code unit at index k within string.
-        let c = string.code_point_at(k).as_u32() as u16;
+        let c = string.code_unit_at(k).js_expect("Bounds were verified")?;
```
jedel1043: I know this is kinda unrelated, but now that I look at this code it seems that we don't need to check if `k == str_len` and `js_expect` here, we can just match against the `Option` given by `code_unit_at`, then return if it is `None`, which would remove the `js_expect`. Can you do that for me?

jedel1043 (changes requested):
Looks great! I just have a very small suggestion

core/engine/src/builtins/uri/mod.rs
HiteshShonak: done! replaced the separate bounds check and js_expect with a single pattern match on code_unit_at

jedel1043 (approved):
Thank you, looks great

PR #5171, opened 2026-03-20
feat(builtins): implement `Iterator.concat`

Nakshatra480:
## Summary

This implements the `Iterator.concat` static method from the [Iterator Sequencing](https://github.com/tc39/proposal-iterator-sequencing) proposal (currently Stage 4 / ES2026). This is the last remaining static method for the [Iterator](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/iterable/mod.rs:192:0-192:27) constructor.

Behavior matches the latest spec text:
- Validates all arguments upfront before opening any iterators (checks for object and `@@iterator` presence).
- Bumps `CONSTRUCTOR_STORAGE_SLOTS` to 2 to account for the new static method alongside [from](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/iterable/iterator_constructor.rs:162:4-204:5).
- Lazily opens each iterable one at a time as previous ones are exhausted within `IteratorHelperOp::Concat`.
- Tracks opened inner iterators in a `Vec<IteratorRecord>` (`[[UnderlyingIterators]]`) for proper GC tracing.
- Forwards `.return()` calls correctly to the active inner iterator.

Added test cases to `builtins::iterable::tests` covering normal arrays, generators, error conditions, and early returns. Passes the existing `test262` suite for `Iterator.concat`.

### DIFF
--- a/core/engine/src/builtins/iterable/iterator_constructor.rs
+++ b/core/engine/src/builtins/iterable/iterator_constructor.rs
@@ -233,6 +233,7 @@ impl IteratorConstructor {
                 iterables,
                 current_index: 0,
                 inner: None,
+                opened: Vec::new(),
             },
             context,
         );
--- a/core/engine/src/builtins/iterable/iterator_constructor.rs
+++ b/core/engine/src/builtins/iterable/iterator_constructor.rs
@@ -204,15 +204,20 @@ impl IteratorConstructor {
     }
 
     fn concat(_this: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
+        // 1. Let iterables be a new empty List.
         let mut iterables = Vec::with_capacity(args.len());
 
+        // 2. For each element item of items, do
         for item in args {
+            // a. If item is not an Object, throw a TypeError exception.
             if !item.is_object() {
                 return Err(JsNativeError::typ()
                     .with_message("Iterator.concat requires iterable objects")
                     .into());
             }
 
+            // b. Let method be ? GetMethod(item, %Symbol.iterator%).
+            // c. If method is undefined, throw a TypeError exception.
             let method = item
                 .get_method(JsSymbol::iterator(), context)?
                 .ok_or_else(|| {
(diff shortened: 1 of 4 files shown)


core/engine/src/builtins/iterable/iterator_constructor.rs
```diff
@@ -202,6 +203,42 @@ impl IteratorConstructor {
...
+        let helper = IteratorHelper::create(
+            dummy_iterator,
+            IteratorHelperOp::Concat {
+                iterables,
+                current_index: 0,
+                inner: None,
+                opened: Vec::new(),
+            },
+            context,
+        );
```
jedel1043: According to the spec, this is _not_ the correct way to implement this. The spec updated `[[UnderlyingIterator]]` to `[[UnderlyingIterators]]`, so the `underlying_iterator` field from `IteratorHelper` should be updated to a `Vec` instead of using a dummy iterator record.

jedel1043 (changes requested):

core/engine/src/builtins/iterable/iterator_helper.rs
```diff
@@ -598,90 +596,23 @@ impl IteratorHelper {
...
-                    if let Some(inner_iter) = inner_to_close {
-                        inner_iter.close(Ok(JsValue::undefined()), context)?;
-                    }
-                    return Ok(create_iter_result_object(
-                        JsValue::undefined(),
-                        true,
-                        context,
-                    ));
+                for iter in &iterators {
+                    iter.close(Ok(JsValue::undefined()), context)?;
```
jedel1043: This is not correct. According to the [spec](https://tc39.es/ecma262/#sec-iteratorcloseall), all iterators should be closed in reverse order, but making sure that only the last error is the one that bubbles up to the caller.

jedel1043:
You should add spec comments that mirror what the code does. That makes checking for errors and inconsistencies much easier.

Nakshatra480:
Thanks for the review! I've refactored underlying_iterator to a Vec to match the spec's [[UnderlyingIterators]] slot, implemented the correct reverse-order last-error-wins semantics for IteratorCloseAll in r#return, and added detailed spec-step comments throughout the implementation.

jedel1043:
Cool, now just remove the ignore for iterator helpers and this proposal in the `test262_config.toml` file to check the conformance changes

Nakshatra480:
Done! The ignores for iterator-helpers and iterator-sequencing have been removed from test262_config.toml so the CI checks conformance changes now.

jedel1043:
Three new panics, but I think only one of them is related to this PR

test/built-ins/Iterator/concat/throws-typeerror-when-generator-is-running-next.js

Nakshatra480:
Good catch @jedel1043 ! We were holding a mutable borrow on the helper during the inner step_value() call, which caused a BorrowMutError panic if user code re-entered .next(). I've updated it to drop the borrow first, so it now correctly hits the Executing state guard and throws the expected TypeError.

jedel1043 (approved):
Looks good!

Nakshatra480:
Thanks Jedel

PR #5172, opened 2026-03-20
fix(runtime): honor DataView offset and length in TextDecoder.decode

### LINKED ISSUE #4767: `TextDecoder` does not respect `DataView` offset and length

HiteshShonak:
This Pull Request fixes/closes #4767.

It changes the following:

- Updates `TextDecoder.decode` for `DataView` inputs to apply the view window (`byteOffset` + `byteLength`) instead of decoding the full backing `ArrayBuffer`.
- Adds checked conversion for `DataView` offset/length and returns `RangeError` if values exceed addressable size.
- Adds a regression test to verify `new DataView(buffer, 1, 2)` decodes only that slice (expected `"CE"`).

Testing:

- `cargo test -p boa_runtime text::tests` — 23 passed, 0 failed
- `cargo fmt --all` — passed

Repro from issue now returns `"B"` instead of `"BB"`:
- `new TextDecoder().decode(new DataView(Uint8Array.of(0x42, 0x42).buffer, 1))`

Spec reference:
https://encoding.spec.whatwg.org/#dom-textdecoder-decode


jedel1043 (approved):
Thanks!

PR #5177, opened 2026-03-20
fix(fetch): throw TypeError objects for invalid Headers name/value

### LINKED ISSUE #5175: `Headers` throws non-Error value instead of `TypeError` for invalid header name/value
**Describe the bug**
`new Headers()`, `headers.append()`, and `headers.set()` throw a plain string value instead of a `TypeError` object when given an invalid header name or value. Caught exceptions do not expose standard error fields like `name` or `message` and do not behave like normal Error objects.

**To Reproduce**

```js
try {
  new Headers([["a\n", "b"]]);
} catch (e) {
  console.log(typeof e, e && e.name, e && e.message);
}
try {
  new Headers([["x-test", "a\u0000b"]]);
} catch (e) {
  console.log(typeof e, e && e.name, e && e.message);
}
```

From repo root:
```bash
cargo run --bin boa -- -e 'try { new Headers([["a\n","b"]]); } catch (e) { console.log(typeof e + " " + (e && e.name) + " " + (e && e.message)); }'
```

Boa output:
```
string undefined undefined
```

Node.js reference:
```bash
node -e 'try { new Headers([["a\n","b"]]); } catch (e) { console.log(typeof e + " " + (e && e.name) + " " + (e && e.message)); }'
# object TypeError Headers.append: "a\n" is an invalid header name.
```

**Expected behavior**
`Headers` should throw a proper `TypeError` object for invalid header names and values, matching browser and Node.js behavior. The [Fetch spec][spec] step 1 states: "If name is not a header name or value is not a header value, then throw a TypeError."

[spec]: https://fetch.spec.whatwg.org/#dom-headers-append

- OS: Windows 11
- Version: 10.0.26200.8037

**Additional context**
Root cause is in `core/runtime/src/fetch/headers.rs` where `to_header_name()` and `to_header_value()` use plain `js_error!("...")` instead of `js_error!(TypeError: "...")`.

HiteshShonak:
This Pull Request fixes/closes #5175.

It changes the following:
- `to_header_name()` and `to_header_value()` now throw `TypeError` objects 
instead of plain string values for invalid header names and values.
- All paths that use these helpers (constructor, `append()`, `set()`) are 
fixed automatically.
- Added a regression test covering invalid name and invalid value cases.

Testing:
```bash
cargo test -p boa_runtime headers -- --nocapture
```

Spec reference:
https://fetch.spec.whatwg.org/#dom-headers-append


core/runtime/src/fetch/headers.rs
```diff
@@ -25,8 +25,9 @@ pub type ForEachCallback = TypedJsFunction<(JsString, JsString, JsObject), ()>;
 /// If the key is not valid ASCII, an error is returned.
 #[inline]
 fn to_header_name(key: impl AsRef<str>) -> JsResult<HeaderName> {
-    HeaderName::from_str(key.as_ref())
-        .map_err(|_| js_error!("Cannot convert key to header string as it is not valid ASCII."))
+    HeaderName::from_str(key.as_ref()).map_err(
+        |_| js_error!(TypeError: "Cannot convert key to header string as it is not valid ASCII."),
+    )
```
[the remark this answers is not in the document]
HiteshShonak: Good point, updated to invalid header name. which is more accurate.

core/runtime/src/fetch/headers.rs
```diff
@@ -41,9 +42,9 @@ fn normalize_header_value(value: &str) -> &str {
 /// If the value is not valid ASCII, an error is returned.
 #[inline]
 fn to_header_value(value: impl AsRef<str>) -> JsResult<HeaderValue> {
-    normalize_header_value(value.as_ref())
-        .parse()
-        .map_err(|_| js_error!("Cannot convert value to header string as it is not valid ASCII."))
+    normalize_header_value(value.as_ref()).parse().map_err(
+        |_| js_error!(TypeError: "Cannot convert value to header string as it is not valid ASCII."),
+    )
```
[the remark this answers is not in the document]
HiteshShonak: Updated to invalid header value as suggested.

core/runtime/src/fetch/tests/headers.rs
```diff
@@ -84,3 +84,41 @@ fn headers_normalize_values() {
...
+                ];
+
+                for (const make of cases) {
+                    let threw = false;
+                    try {
+                        make();
+                    } catch (e) {
+                        threw = true;
+                        assertEq(typeof e, "object", "should throw an Error object");
+                        assertEq(e.name, "TypeError", "should throw TypeError");
```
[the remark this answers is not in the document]
HiteshShonak: added instanceof typeerror assertion to the test

core/runtime/src/fetch/tests/headers.rs
```diff
@@ -84,3 +84,42 @@ fn headers_normalize_values() {
...
+                    () => {
+                        const h = new Headers();
+                        h.set("x-test", "a\u0000b");
+                    },
+                ];
+
+                for (const make of cases) {
+                    let threw = false;
+                    try {
+                        make();
```
jedel1043, line 110:
```suggestion
                        make();
                        throw Error("expected the call above to throw");
```

This way you don't need the `threw` flag, since the asserts below will take care of checking the correct error type

core/runtime/src/fetch/tests/headers.rs
```diff
@@ -84,3 +84,42 @@ fn headers_normalize_values() {
...
+                    } catch (e) {
+                        threw = true;
+                        assertEq(typeof e, "object", "should throw an Error object");
+                        assertEq(e.name, "TypeError", "should throw TypeError");
```
jedel1043: Not needed. `instanceof` already does all of this.
```suggestion
```

core/runtime/src/fetch/tests/headers.rs
```diff
@@ -84,3 +84,41 @@ fn headers_normalize_values() {
...
+                        h.append("a\n", "b");
+                    },
+                    () => {
+                        const h = new Headers();
+                        h.set("x-test", "a\u0000b");
+                    },
+                ];
+
+                for (const make of cases) {
+                    let threw = false;
```
jedel1043: We don't need `threw` anymore, the assert already checks that it throws the correct error

```suggestion
```

core/runtime/src/fetch/tests/headers.rs
```diff
@@ -84,3 +84,41 @@ fn headers_normalize_values() {
...
+                    try {
+                        make();
+                        throw Error("expected the call above to throw");
+                    } catch (e) {
+                        threw = true;
+                        assert(e instanceof TypeError, "should throw TypeError instance");
+                        assert(typeof e.message === "string" && e.message.length > 0, "error message should be non-empty");
+                    }
+
+                    assert(threw, "expected Headers validation to throw");
```
jedel1043:
```suggestion
```

core/runtime/src/fetch/tests/headers.rs
```diff
@@ -84,3 +84,41 @@ fn headers_normalize_values() {
...
+                    },
+                ];
+
+                for (const make of cases) {
+                    let threw = false;
+                    try {
+                        make();
+                        throw Error("expected the call above to throw");
+                    } catch (e) {
+                        threw = true;
```
jedel1043:
```suggestion
```

jedel1043 (changes requested):
Cool!

jedel1043 (approved):
Looks great!

PR #5188, opened 2026-03-20
Fix trace output

yush-1018:
This PR addresses the trace output formatting overflow in the CLI. 

It implements truncation for both the unbounded stack representation strings and the instruction operand strings to ensure they fit neatly within the terminal columns without wrapping or breaking the table alignment.


jedel1043:
Duplicate of #5104

PR #5190, opened 2026-03-20
bench: lazily initialize script benchmarks

### LINKED ISSUE #5169: boa_benches: filtered script benchmark runs still eagerly initialize unrelated scripts
## Problem

Filtered script benchmark runs still initialize unrelated script benches first.

While working on issue #5145 and adding a focused local script benchmark, I tried running only that target with:
```
cargo bench -p boa_benches -- json/stringify_deep
```

Even with the filter, `benches/benches/scripts.rs` initializes all discovered scripts. This includes `v8-benches/deltablue`, which overflows the stack during registration and aborts the run before the requested benchmark executes.

## Repro
```
cargo bench -p boa_benches -- json/stringify_deep
```

This fails with a stack overflow in an unrelated `v8-benches/*` script.

## Evidence

Adding temporary logging during registration shows all scripts being initialized:
```
REGISTERING BENCH: basic\call-loop
...
REGISTERING BENCH: v8-benches\deltablue
...
```

So the failure happens before the filtered benchmark is reached.

## Root Cause

`scripts.rs` eagerly reads, parses, and evaluates all scripts during registration. Since Criterion filtering happens later, unrelated scripts can still fail and block targeted runs.

## Expected Behavior

A filtered run should not be blocked by unrelated script benchmarks being initialized eagerly.

iammdzaidalam:
Closes #5169

## Summary

Defer script benchmark setup until the selected benchmark actually runs.

`benches/benches/scripts.rs` was eagerly reading, parsing, compiling, and evaluating every script during registration, so filtered runs could still fail on unrelated entries before reaching the requested benchmark.

This moves that setup behind the benchmark closure and caches the prepared state per benchmark, so unmatched scripts are no longer initialized.

## Changes

- add a small `PreparedScriptBench` helper for cached per-benchmark state
- move script file reading into lazy setup
- move `Context` creation, runtime registration, parse/compile/evaluate, and `main` lookup into lazy setup
- keep benchmark discovery and existing `v8-benches` group config unchanged
- cache the prepared script once per matched benchmark so setup is not repeated during measurement

## Verification

Ran locally:

- `cargo fmt --check`
- `cargo check -p boa_benches`
- `cargo bench -p boa_benches -- --list`
- `cargo bench -p boa_benches -- call-loop`

Also temporarily added logging in the lazy init path to verify behavior:

- `call-loop` only initialized `basic/call-loop.js`
- a nonexistent filter initialized nothing
- `deltablue` only initialized `v8-benches/deltablue.js`

So filtered runs no longer initialize unrelated scripts first.


benches/benches/scripts.rs
```diff
@@ -26,49 +60,24 @@ fn bench_scripts(c: &mut Criterion) {
...
-        // Get the main function
-        let function = context
-            .global_object()
-            .get(js_string!("main"), context)
-            .unwrap_or_else(|_| panic!("No main function defined in script: {}", path.display()))
-            .as_callable()
-            .unwrap_or_else(|| panic!("'main' is not a function in script: {}", path.display()))
-            .clone();
+        group.bench_function("Execution", move |b| {
+            let prepared = prepared.get_or_insert_with(|| prepare_script_bench(&path));
```
jedel1043: I don't think we should put the initialization code inside the benchmark, it'll just pollute the results.

hansl: Was about to say. The reason we use the `main` function is to benchmark specific bits of VM, not the initialization and parsing (and optimization, etc).

iammdzaidalam: Oh, I was mainly trying to avoid the eager init issue but putting it inside the benchmark isnt the right tradeoff here... thinking of instead filtering before registration so only matching scripts get initialized, and keeping the setup outside bench_function like before

does that sound like the right direction?

jedel1043: Yeah that's a better approach

iammdzaidalam:
hey @jedel1043, Please take a look when u get free.

jedel1043 (approved):
LGTM, thanks!

iammdzaidalam:
@jedel1043 Thanks for taking another look at this! Really appreciate it 🙌

PR #5201, opened 2026-03-21
fix(fetch): reject Request body for GET and HEAD methods

### LINKED ISSUE #5200: Request constructor accepts body for GET and HEAD methods instead of throwing
**Describe the bug**
`new Request()` silently accepts a body when the method is `GET` or `HEAD`, but the Fetch Standard requires throwing a `TypeError` in this case. Boa just proceeds without any error.

**To Reproduce**
```js
new Request("https://example.com", { method: "GET", body: "x" })

new Request("https://example.com", { method: "HEAD", body: "x" })
```

Verified in Node.js:
```bash
node -e 'try { new Request("https://example.com", { method: "GET", body: "x" }); console.log("OK"); } catch (e) { console.log(e.name + ": " + e.message); }'
# TypeError: Request with GET/HEAD method cannot have body.
```

**Expected behavior**
`new Request()` should throw a `TypeError` when body is provided with `GET` or `HEAD`, matching Node.js and browser behavior.

**Build environment:**
- OS: Windows 11
- Version: 10.0.26200

**Additional context**
Root cause is in `core/runtime/src/fetch/request.rs` where `into_request_builder` sets method and body separately without checking the GET/HEAD + body restriction.

Spec reference: https://fetch.spec.whatwg.org/#dom-request

HiteshShonak:
This Pull Request fixes/closes #5200.

It changes the following:
- `into_request_builder` now checks if the method is `GET` or `HEAD` and throws a `TypeError` if a body is provided, matching the Fetch Standard.
- Added regression tests for both `GET` and `HEAD` cases.

Testing:
```bash
cargo test -p boa_runtime request -- --nocapture
```

Spec reference: https://fetch.spec.whatwg.org/#dom-request

### DIFF
--- a/core/runtime/src/fetch/request.rs
+++ b/core/runtime/src/fetch/request.rs
@@ -41,9 +41,13 @@ impl RequestInit {
         request: Option<HttpRequest<Vec<u8>>>,
     ) -> JsResult<HttpRequest<Vec<u8>>> {
         let mut builder = HttpRequest::builder();
+        let mut is_get_or_head_method = true;
+        let mut has_inherited_body = false;
         let mut request_body = Vec::new();
         if let Some(r) = request {
             let (parts, body) = r.into_parts();
+            is_get_or_head_method = matches!(parts.method, http::Method::GET | http::Method::HEAD);
+            has_inherited_body = !body.is_empty();
             builder = builder
                 .method(parts.method)
                 .uri(parts.uri)
@@ -65,7 +69,6 @@ impl RequestInit {
             let method = method.to_std_string().map_err(
                 |_| js_error!(TypeError: "Request constructor: {} is an invalid method", method.to_std_string_escaped()),
             )?;
-
             // 25. If init["method"] exists, then:
             //     1. Let method be init["method"].
             //     2. If method is not a method or method is a forbidden method, throw a TypeError.
@@ -82,7 +85,14 @@ impl RequestInit {
                 ));
             }
 
+            is_get_or_head_method =
+                method.eq_ignore_ascii_case("GET") || method.eq_ignore_ascii_case("HEAD");
+
             builder = builder.method(method.as_str());
+        }
+
+        if is_get_or_head_method && (self.body.is_some() || has_inherited_body) {
+            return Err(js_error!(TypeError: "Request with GET/HEAD method cannot have body."));
         }
 
         if let Some(body) = &self.body {
--- a/core/runtime/src/fetch/request.rs
+++ b/core/runtime/src/fetch/request.rs
@@ -25,6 +25,9 @@ pub struct RequestInit {
     signal: Option<JsObject>,
 }
 
+#[derive(Clone, Copy)]
+struct HasBody;
+
 impl RequestInit {
     /// Takes the abort signal from the options, if present.
     pub fn take_signal(&mut self) -> Option<JsObject> {
@@ -47,7 +50,7 @@ impl RequestInit {
         if let Some(r) = request {
             let (parts, body) = r.into_parts();
             is_get_or_head_method = matches!(parts.method, http::Method::GET | http::Method::HEAD);
-            has_inherited_body = !body.is_empty();
+            has_inherited_body = parts.extensions.get::<HasBody>().is_some() || !body.is_empty();
             builder = builder
                 .method(parts.method)
                 .uri(parts.uri)
@@ -109,9 +112,13 @@ impl RequestInit {
             }
         }
 
-        builder
+        let mut request = builder
             .body(request_body)
-            .map_err(|_| js_error!(Error: "Cannot construct request"))
+            .map_err(|_| js_error!(Error: "Cannot construct request"))?;
+        if self.body.is_some() || has_inherited_body {
+            request.extensions_mut().insert(HasBody);
+        }
+        Ok(request)
     }
 }
 
--- a/core/runtime/src/fetch/request.rs
+++ b/core/runtime/src/fetch/request.rs
@@ -41,14 +41,9 @@ impl RequestInit {
         request: Option<HttpRequest<Vec<u8>>>,
     ) -> JsResult<HttpRequest<Vec<u8>>> {
         let mut builder = HttpRequest::builder();
-        let mut is_get_or_head_method = true;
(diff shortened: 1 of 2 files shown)


core/runtime/src/fetch/request.rs
```diff
@@ -41,9 +41,13 @@ impl RequestInit {
         request: Option<HttpRequest<Vec<u8>>>,
     ) -> JsResult<HttpRequest<Vec<u8>>> {
         let mut builder = HttpRequest::builder();
+        let mut is_get_or_head_method = true;
+        let mut has_inherited_body = false;
         let mut request_body = Vec::new();
         if let Some(r) = request {
             let (parts, body) = r.into_parts();
+            is_get_or_head_method = matches!(parts.method, http::Method::GET | http::Method::HEAD);
+            has_inherited_body = !body.is_empty();
```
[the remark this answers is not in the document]
HiteshShonak: verified, already throws correctly for empty body too

core/runtime/src/fetch/tests/request.rs
```diff
@@ -48,6 +49,46 @@ fn request_constructor() {
...
+            } catch (e) {
+                if (!(e instanceof TypeError)) {
+                    throw e;
+                }
+            }
+        "#}),
+    ]);
+}
+
 #[test]
```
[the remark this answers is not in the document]
HiteshShonak: good catch, fixed. used a HasBody marker to track body presence and added tests for the inherited empty body case.

core/runtime/src/fetch/request.rs
```diff
@@ -41,9 +44,13 @@ impl RequestInit {
         request: Option<HttpRequest<Vec<u8>>>,
     ) -> JsResult<HttpRequest<Vec<u8>>> {
         let mut builder = HttpRequest::builder();
+        let mut is_get_or_head_method = true;
+        let mut has_inherited_body = false;
         let mut request_body = Vec::new();
         if let Some(r) = request {
             let (parts, body) = r.into_parts();
+            is_get_or_head_method = matches!(parts.method, http::Method::GET | http::Method::HEAD);
+            has_inherited_body = parts.extensions.get::<HasBody>().is_some() || !body.is_empty();
```
jedel1043: Can you link the specific part of the spec that handles inherited bodies? I was looking through it and I didn't see it doing this. The only thing it checked was the parent request and the RequestInit object.

jedel1043 (changes requested):

core/runtime/src/fetch/request.rs
jedel1043: In general I would like to see this being commented with the specific spec lines it's implementing, since it makes maintainance much easier.

HiteshShonak: hey, added spec comments to both places. the HasBody extension is there because boa stores body as Vec<u8>, so an empty string body and no body both look the same. needed a way to track if body was explicitly set, which is what inputBody is non-null checks in the spec: https://fetch.spec.whatwg.org/#dom-request

jedel1043: That sounds kinda suboptimal. What about just changing body to an `Option<Vec<u8>>`? That avoids having to use an extension to something that should be "native" in a way.

HiteshShonak: switched to `Option<Vec<u8>>` and removed the HasBody extension. let me know if this looks good

core/runtime/src/fetch/request.rs
```diff
@@ -41,9 +41,16 @@ impl RequestInit {
         request: Option<HttpRequest<Vec<u8>>>,
     ) -> JsResult<HttpRequest<Vec<u8>>> {
         let mut builder = HttpRequest::builder();
-        let mut request_body = Vec::new();
+        let mut is_get_or_head_method = true;
+        let mut inherited_is_get_or_head_method = true;
+        let mut inherited_body = None;
+        let mut request_body: Option<Vec<u8>> = None;
```
jedel1043: I'm slightly confused about using all these flags. As far as I understand you just need `inherited_body`, and the rest can be inlined into the specific part of the code using them, since they're mostly used only once.

jedel1043: `request_body` doesn't need to exist here, you can put it below by using a `let-if` block:

```rust
let request_body = if let Some(body) = &self.body {
    // ...
} else {
    input_body
};
```

core/runtime/src/fetch/request.rs
```diff
@@ -62,9 +68,25 @@ impl RequestInit {
...
+
+        // Fetch Standard §5.4 Request constructor:
+        // If either init["body"] exists and is non-null or inputBody is non-null,
+        // and request's method is GET or HEAD, then throw a TypeError.
+        // https://fetch.spec.whatwg.org/#dom-request
+        if is_get_or_head_method
+            && (self.body.is_some()
+                || inherited_body
+                    .as_ref()
+                    .is_some_and(|body| !body.is_empty() || !inherited_is_get_or_head_method))
```
jedel1043: Why `inherited_body` also checks that the body is empty? I thought this would throw even if the body is empty, since the spec is checking if the body is null, not if it's empty.
This change would also considerably simplify the code:
```rust
let is_get_or_head_method = method.eq_ignore_ascii_case("GET") || method.eq_ignore_ascii_case("HEAD");
if is_get_or_head_method && (self.body.is_some() || inherited_body.is_some() {
    // throw error
}
```

jedel1043 (changes requested):

core/runtime/src/fetch/request.rs
HiteshShonak: simplified this and inlined the extra flags, and moved request_body into the let-if block as suggested

core/runtime/src/fetch/request.rs
HiteshShonak: removed the empty-body check here and switched to the simplified null/non-null check

HiteshShonak:
I fixed the clippy issue in core/macros, but now CI is failing on core/string (while_let_loop). It looks like newer clippy rules are surfacing additional lints outside the scope of this PR.

Should I continue fixing these as they come up, or would you prefer handling them separately?

jedel1043 (approved):
LGTM! Thank you for the contribution!

PR #5203, opened 2026-03-21
fix(fetch): reject forbidden HTTP methods CONNECT, TRACE, and TRACK

### LINKED ISSUE #5202: `Request` constructor accepts forbidden HTTP methods like CONNECT, TRACE, and TRACK
**Describe the bug**
`new Request()` accepts forbidden HTTP methods like `CONNECT`, `TRACE`, and `TRACK` without throwing. the Fetch Standard explicitly forbids these methods and browsers throw a `TypeError` for them.

**To Reproduce**
```js
new Request("https://example.com", { method: "CONNECT" })

new Request("https://example.com", { method: "TRACE" })

new Request("https://example.com", { method: "TRACK" })
```

Verified in Node.js:
```bash
node -e 'try { new Request("https://example.com", { method: "CONNECT" }); console.log("OK"); } catch (e) { console.log(e.name + ": " + e.message); }'
# TypeError: 'CONNECT' HTTP method is unsupported.

node -e 'try { new Request("https://example.com", { method: "TRACE" }); console.log("OK"); } catch (e) { console.log(e.name + ": " + e.message); }'
# TypeError: 'TRACE' HTTP method is unsupported.

node -e 'try { new Request("https://example.com", { method: "TRACK" }); console.log("OK"); } catch (e) { console.log(e.name + ": " + e.message); }'
# TypeError: 'TRACK' HTTP method is unsupported.
```

**Expected behavior**
`new Request()` should throw a `TypeError` when a forbidden method is used, matching Node.js and browser behavior.

**Build environment:**
- OS: Windows 11
- Version: 10.0.26200

**Additional context**
root cause is in `core/runtime/src/fetch/request.rs` - method is set without checking against the forbidden methods list from the Fetch spec.

spec reference: https://fetch.spec.whatwg.org/#forbidden-method

HiteshShonak:
This Pull Request fixes/closes #5202.

It changes the following:
- Reject `CONNECT`, `TRACE`, and `TRACK` methods in the `Request` constructor and throw a `TypeError`, matching the Fetch Standard.
- Check is case-insensitive, so `connect`, `trace`, `track` are also rejected.
- Added regression tests for all three forbidden methods.

Testing:
```bash
cargo test -p boa_runtime request -- --nocapture
```

Spec reference: https://fetch.spec.whatwg.org/#forbidden-method


core/runtime/src/fetch/tests/request.rs
```diff
@@ -48,6 +49,66 @@ fn request_constructor() {
...
+        TestAction::inspect_context(|ctx| {
+            let fetcher = TestFetcher::default();
+            crate::fetch::register(fetcher, None, ctx).expect("failed to register fetch");
+        }),
+        TestAction::run(indoc! {r#"
+            try {
+                new Request("http://unit.test", { method: "CONNECT" });
+                throw Error("expected the call above to throw");
+            } catch (e) {
+                if (!(e instanceof TypeError)) {
```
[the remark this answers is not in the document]
HiteshShonak: good catch, added lowercase connect to the consolidated test.

core/runtime/src/fetch/tests/request.rs
```diff
@@ -48,6 +49,66 @@ fn request_constructor() {
     ]);
 }
 
+#[test]
+fn request_constructor_connect_method_throws() {
+    run_test_actions([
+        TestAction::inspect_context(|ctx| {
+            let fetcher = TestFetcher::default();
```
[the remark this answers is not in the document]
HiteshShonak: done, consolidated all four cases into one test

core/runtime/src/fetch/request.rs
```diff
@@ -62,9 +62,21 @@ impl RequestInit {
...
+                || method.eq_ignore_ascii_case("TRACE")
+                || method.eq_ignore_ascii_case("TRACK")
+            {
+                return Err(js_error!(
+                    TypeError: "'{}' HTTP method is unsupported.",
+                    method
+                ));
+            }
+
+            builder = builder.method(method.as_str());
```
jedel1043, line 85: It might be good to reference the spec steps that are implemented here (https://fetch.spec.whatwg.org/#dom-request):

```
// 25. If init["[method](https://fetch.spec.whatwg.org/#dom-requestinit-method)"] [exists](https://infra.spec.whatwg.org/#map-exists), then:
//
//     1. Let method be init["[method](https://fetch.spec.whatwg.org/#dom-requestinit-method)"].
//
//     2. If method is not a [method](https://fetch.spec.whatwg.org/#concept-method) or method is a [forbidden method](https://fetch.spec.whatwg.org/#forbidden-method), then [throw](https://webidl.spec.whatwg.org/#dfn-throw) a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).
//
//     3. [Normalize](https://fetch.spec.whatwg.org/#concept-method-normalize) method.
//
//     4. Set request’s [method](https://fetch.spec.whatwg.org/#concept-request-method) to method.

Comment (HiteshShonak) [reply core/runtime/src/fetch/request.rs:85]: added the spec comments for that check

File: core/runtime/src/fetch/request.rs [hunk 4]
```diff
@@ -62,9 +62,27 @@ impl RequestInit {
...
-            builder = builder.method(method.to_std_string().map_err(
+            let method = method.to_std_string().map_err(
                 |_| js_error!(TypeError: "Request constructor: {} is an invalid method", method.to_std_string_escaped()),
-            )?.as_str());
+            )?;
+
+            // 25. If init["method"] exists, then:
+            //     1. Let method be init["method"].
+            //     2. If method is not a method or method is a forbidden method, throw a TypeError.
+            //     3. Normalize method.
```
jedel1043, line 72: Thought: this also kinda shows we might also need to normalize methods, but that can be a separate PR.

jedel1043 (approved):
Looks great!

core/runtime/src/fetch/request.rs
HiteshShonak: yeah, that makes sense. i'll look into that separately

PR #5206, opened 2026-03-21
fix(intl): read weekday option in DateTimeFormat constructor

Lokesh0224:
`weekday` option was incorrectly spelled as `weekDay` (camelCase) in FormatOptions::try_init, causing the option getter to never be called. 
So this caused three failures:

constructor-options-order.js 
constructor-options-throwing-getters.js 
test-option-date-time-components.js 

Fix: correct the option key from "weekDay" to "weekday" per ECMA-402 [§11.5.](https://402.ecma-international.org/11.0/index.html#sec-datetimeformat-abstracts)


nekevss (approved):
Huh, nice catch!

PR #5208, opened 2026-03-21
fix(fetch): add status 205 to null body status check in Response

### LINKED ISSUE #5207: `Response` constructor accepts body for status 205 instead of throwing
**Describe the bug**
`new Response()` silently accepts a body when the status is `205`. the Fetch Standard defines `205` as a null body status, so passing a body should throw a `TypeError` - same as `204` and `304` which already throw correctly in Boa.

**To Reproduce**
```js
new Response("x", { status: 205 })
```

Verified in Node.js:
```bash
node -e 'try { new Response("x", { status: 205 }); console.log("OK"); } catch (e) { console.log(e.name + ": " + e.message); }'
# TypeError: Response constructor: Invalid response status code 205

cargo run --bin boa -- -e 'try { new Response("x", { status: 205 }); console.log("OK"); } catch (e) { console.log(e.name + ": " + e.message); }'
# OK
```

**Expected behavior**
`new Response("x", { status: 205 })` should throw a `TypeError`, matching Node.js and browser behavior. `new Response(null, { status: 205 })` should still be fine.

**Build environment:**
- OS: Windows 11
- Version: 10.0.26200

**Additional context**
Root cause is in `core/runtime/src/fetch/response.rs` - `205` is missing from the `is_null_body_status` check.

Spec reference: https://fetch.spec.whatwg.org/#null-body-status

HiteshShonak:
This Pull Request fixes/closes #5207.

It changes the following:
- Add `205` to the `is_null_body_status` check in `response.rs` so `new Response("x", { status: 205 })` throws a `TypeError`.
- Update the error message and doc comment to include `205`.
- Add a consolidated regression test covering `204`, `205`, and `304`.

Testing:
```bash
cargo test -p boa_runtime response -- --nocapture
```

Spec reference: https://fetch.spec.whatwg.org/#null-body-status

### DIFF
--- a/core/runtime/src/fetch/response.rs
+++ b/core/runtime/src/fetch/response.rs
@@ -96,11 +96,11 @@ impl TryIntoJs for ResponseType {
     }
 }
 
-/// A null body status is a status that is 101, 103, 204, or 304.
+/// A null body status is a status that is 101, 103, 204, 205, or 304.
 ///
 /// See <https://fetch.spec.whatwg.org/#null-body-status>
 fn is_null_body_status(status: u16) -> bool {
-    matches!(status, 101 | 103 | 204 | 304)
+    matches!(status, 101 | 103 | 204 | 205 | 304)
 }
 
 /// Validates that a string matches the `reason-phrase` token production.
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Good catch

PR #5212, opened 2026-03-21
fix(fetch): set correct type for Response.redirect()

### LINKED ISSUE #5211: `Response.redirect()` returns wrong `type` - `basic` instead of `default`
**Describe the bug**
`Response.redirect()` returns `"basic"` for the `type` property instead of `"default"`. the Fetch Standard says a new response type is `"default"` unless stated otherwise, and nothing in the redirect algorithm changes it.

**To Reproduce**
```js
Response.redirect("https://example.com").type
// returns "basic" in Boa, expected "default"
```

From repo root:
```bash
cargo run --bin boa -- -e 'console.log(Response.redirect("https://example.com").type);'
```

Boa output:
```
basic
```

Node.js output:
```bash
node -e 'console.log(Response.redirect("https://example.com").type);'
# default
```

**Expected behavior**
`Response.redirect().type` should return `"default"`, matching Node.js and browser behavior.

**Build environment:**
- OS: Windows 11
- Version: 10.0.26200

**Additional context**
Root cause is in `core/runtime/src/fetch/response.rs` - the redirect static method sets `ResponseType::Basic` instead of `ResponseType::Default`.

Spec reference: https://fetch.spec.whatwg.org/#dom-response-redirect

HiteshShonak:
This Pull Request fixes/closes #5211.

It changes the following:
- Change `ResponseType::Basic` to `ResponseType::Default` in `Response.redirect()` - a new response type is `"default"` unless stated otherwise per the Fetch Standard.
- Add a regression test asserting `Response.redirect().type === "default"`.

Testing:
```bash
cargo test -p boa_runtime response -- --nocapture
```

Spec reference: https://fetch.spec.whatwg.org/#dom-response-redirect


jedel1043 (approved):
Cool, thanks!

PR #5214, opened 2026-03-21
perf(store): use FxHashMap/FxHashSet for traversal seen maps

### LINKED ISSUE #5213: perf(store): Use FxHash for traversal seen maps
While reviewing the structured clone implementation in `core/runtime/src/store`, I noticed we are using `std::collections::{HashMap, HashSet}` for internal bookkeeping structures like `SeenMap` and `ReverseSeenMap`.

These maps are keyed by internal identities (mostly pointer addresses) and get hit constantly during deep or cyclic object traversal. Since these keys represent purely internal engine state, they don't need SipHash's collision resistance. 

Using the default hasher is likely doing more algorithmic work than necessary. Swapping these specific internal maps to `FxHashMap` / `FxHashSet` should reduce the overhead for traversal.

I ran a quick local benchmark using `boa_cli` with a deeply cyclic object graph over 10 rounds of `structuredClone()`, and the `FxHash` swap showed roughly a ~33% improvement in median traversal time (from ~10.4s down to ~6.8s).

I can put up a PR for this if the approach looks good.

iammdzaidalam:
Closes #5213

This switches `HashMap` / `HashSet` to `FxHashMap` / `FxHashSet` in the structured clone/store path (`core/runtime/src/store/from.rs` and `to.rs`).

#### Problem
The store path uses standard collections for traversal bookkeeping (e.g. `SeenMap`, `ReverseSeenMap`).

These maps are keyed by internal identities (pointer-like values) and are hit a lot during deep or cyclic object traversal. Using the default hasher here is probably doing more work than needed.

#### Change
Replace those maps with `FxHashMap` / `FxHashSet`.

This only affects internal bookkeeping. There is no intended semantic change.

#### Benchmark
Ran a local benchmark using `boa_cli` with repeated `structuredClone()` on a deep cyclic object graph (10 rounds, release build).

| Metric | std (main) | FxHash (this PR) | Improvement |
|--------|------------|------------------|-------------|
| Median | 10,423 ms | 6,894 ms | ~33.8% faster |
| Min | 9,231 ms | 6,050 ms | ~34.4% faster |
| Max | 12,028 ms | 8,806 ms | ~26.7% faster |

#### Notes
- no intended semantic change
- only touches internal traversal bookkeeping
- FxHash is already used in other perf-sensitive parts of the engine

#### Test plan
- `cargo test -p boa_engine -p boa_runtime` passes
- manually tested structured cloning on deep / cyclic graphs

### DIFF
--- a/core/runtime/src/store/from.rs
+++ b/core/runtime/src/store/from.rs
@@ -9,12 +9,12 @@ use boa_engine::object::builtins::{
 };
 use boa_engine::property::PropertyKey;
 use boa_engine::{Context, JsError, JsObject, JsResult, JsString, JsValue, JsVariant, js_error};
-use std::collections::{HashMap, HashSet};
+use rustc_hash::{FxHashMap, FxHashSet};
 
 /// A Map of seen objects when walking through the value. We use the address
 /// of the inner object as it is unique per JavaScript value.
 #[derive(Default)]
-pub(super) struct SeenMap(HashMap<usize, JsValueStore>);
+pub(super) struct SeenMap(FxHashMap<usize, JsValueStore>);
 
 impl SeenMap {
     fn get(&self, object: &JsObject) -> Option<JsValueStore> {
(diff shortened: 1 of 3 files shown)


jedel1043 (approved):
LGTM

PR #5215, opened 2026-03-21
feat(console): implement console.table() using comfy-table

### LINKED ISSUE #3806: Implement `console.table()` method

ParthMozarkar:
I implemented the console.table() method for the console builtin, 
this fixes #3806.

I used comfy-table for the table rendering as it was suggested in 
earlier feedback. To keep the code clean and readable I split the 
logic into three small helper functions instead of one big function:

- `property_key_display` - converts a property key to a display string
- `extract_row_cells` - extracts column names and cell values from a row
- `build_table_string` - builds the final ASCII table using comfy-table

It handles these cases:
- Plain objects
- Arrays of primitives (shown under a "Values" column)
- Arrays of objects
- Optional second argument to filter which columns to show

I also added a default `table` method to the Logger trait so none of 
the existing logger implementations need any changes.

Added tests covering all the above cases. Everything passes 
cargo fmt, cargo clippy, and cargo test.

Hope this looks good, let me know if anything needs to be changed!

### DIFF
--- a/core/runtime/Cargo.toml
+++ b/core/runtime/Cargo.toml
@@ -23,6 +23,7 @@ reqwest = { workspace = true, optional = true }
 rustc-hash = { workspace = true, features = ["std"] }
 serde_json = { workspace = true, optional = true }
 url = { workspace = true, optional = true }
+comfy-table="7"
 
 [dev-dependencies]
 indoc.workspace = true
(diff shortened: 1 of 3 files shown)


ParthMozarkar:
Sorry, I just realized this issue is already assigned to someone else. 
Closing this and will wait for another opportunity. Apologies for the noise!

PR #5217, opened 2026-03-21
fix(fetch): handle BufferSource bodies in Response

### LINKED ISSUE #5216: fix(fetch): support BufferSource bodies in Response()
While looking at `core/runtime/src/fetch/response.rs`, I noticed `Response()` body extraction still falls back to stringification for non-string inputs.

That means BufferSource-style inputs like `ArrayBuffer`, typed arrays, and `DataView` don’t get treated as raw bytes when constructing a `Response`.

For example, this currently goes through string coercion instead of byte extraction:

```js
new Response(new Uint8Array([104, 101, 108, 108, 111]))
```
A narrow fix here would be to teach extract_body(...) in response.rs to handle:

- ArrayBuffer
- typed array views
- DataView

using the visible byte slice (byteOffset / byteLength) before falling back to the existing string path.

This would keep the current string behavior intact, avoid injecting the default text content-type for binary inputs, and leave other BodyInit cases like Blob, FormData, and URLSearchParams for follow-up.

### Repro
```js
let u8a = new Uint8Array([104, 101, 108, 108, 111]);
let r = new Response(u8a);

// expected: "hello"
// current: "[object Uint8Array]"
r.text().then(console.log);

// expected: false
// current: true
console.log(r.headers.has("content-type"));
```

Happy to open a small PR scoped just to Response() if this direction makes sense.

iammdzaidalam:
Closes #5216

Adds BufferSource handling to the `Response()` constructor body extraction path.

Currently, inputs like `Uint8Array` or `DataView` fall through to stringification,
so binary bodies end up as strings (e.g. `"[object Uint8Array]"`) instead of raw bytes.

This change:

- handles `ArrayBuffer`, typed arrays, and `DataView`
- slices typed array / `DataView` inputs using `byteOffset` / `byteLength`
- uses `usize::try_from(...)` for offsets/lengths to avoid truncation on 32-bit targets
- preserves the existing string fallback for non-binary inputs
- leaves other `BodyInit` types (e.g. `Blob`, `FormData`) for follow-up

### Testing

Added tests in `core/runtime/src/fetch/tests/response.rs` covering:

- `ArrayBuffer`
- `Uint8Array`
- subarray/view slicing
- `DataView` with non-zero offset/length
- string fallback behavior
- content-type behavior (binary vs string)

### Notes

- binary bodies do not inject a default `Content-Type`
- string bodies still default to `text/plain;charset=UTF-8`


iammdzaidalam:
Hey @jedel1043, any updates here when you get a moment?

iammdzaidalam:
hey @jedel1043 @nekevss ,any updates when u get a moment?

jedel1043 (approved):
Looks great! It just needed an update against main, but we can merge now.

PR #5220, opened 2026-03-22
test: add edge case coverage for Array.prototype.map

ParthMozarkar:
I added test cases for Array.prototype.map covering some important 
edge cases that weren't tested before:

- Empty array
- Callback returning undefined  
- Sparse array (holes should be preserved)
- Identity mapping

No existing tests were removed or modified.

Raising this as a draft for review — let me know if any changes 
are needed!


hansl (approved):
We don't need to do all Array functions. Test262 are doing every function in the spec and all edge cases already.

ParthMozarkar:
> We don't need to do all Array functions. Test262 are doing every function in the spec and all edge cases already.

Thanks for the approval and the feedback! Understood  I won't add 
more Array test PRs since test262 already covers those. I'll focus 
on more meaningful contributions going forward!

jedel1043:
Closing since this has no line changes

PR #5221, opened 2026-03-22
test: add edge case coverage for Array.prototype.find

ParthMozarkar:
I added test cases for Array.prototype.find covering edge cases that 
weren't tested before:

- Empty array returns undefined
- No match returns undefined
- Returns first match not last
- Works correctly with objects
- Callback receives correct index
- Sparse array holes are visited

No existing tests were removed or modified.


core/engine/src/object/tests.rs
zhuzhu81998, line 1: i dont think this is the right place to place a test for Array

ParthMozarkar: You're right, thanks for pointing that out! I'll move the tests to 
core/engine/src/builtins/array/tests.rs where they belong...

hansl (approved):

PR #5222, opened 2026-03-22
test: add edge case coverage for Array.prototype.forEach

ParthMozarkar:
I added test cases for Array.prototype.forEach covering some edge 
cases that weren't tested before:

- Empty array  callback should never be called
- Return value should always be undefined
- Sparse array holes should be skipped by the callback
- this binding should work correctly via second argument

I made sure not to touch or remove any of the existing tests, 
only added the new ones.
Hope this looks good, let me know if any changes are needed!

### DIFF
--- a/core/engine/src/object/tests.rs
+++ b/core/engine/src/object/tests.rs
@@ -38,3 +38,36 @@ fn object_properties_return_order() {
         ),
     ]);
 }
+
+#[test]
+fn array_prototype_for_each_edge_cases() {
+    run_test_actions([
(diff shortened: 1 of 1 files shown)


hansl (approved):
This is already covered by test262. Not sure of the value of these tests.

ParthMozarkar:
> This is already covered by test262. Not sure of the value of these tests.

Thanks for the review! I understand now that test262 already covers 
these cases. Closing this and moving on to more meaningful contributions. 
Sorry for the noise!

PR #5223, opened 2026-03-22
chore(deps): replace paste with pastey

dyc3:
This Pull Request fixes/closes N/A.

It changes the following:
- replaces the `paste` crate with `pastey`

`paste` is unmaintained. `pastey` is a drop in replacement, and is currently maintained.

This PR was generated with gpt 5.4, but its pretty much just a string replacement of the crate name.

The reason I made this PR is because we want to use cargo-deny to deny this crate in our project, but cargo-deny currently doesn't allow handling transitive dependencies differently.


jedel1043 (approved):
Fair enough. We also might wanna migrate away from paste by using trait magic, but that can be a future enhancement.

PR #5226, opened 2026-03-22
fix(builtins): correct toPrecision output for subnormal floats

### LINKED ISSUE #5225: `Number.prototype.toPrecision` returns wrong output for subnormal floats like `Number.MIN_VALUE`
**Describe the bug**
`Number.prototype.toPrecision()` returns wrong output for subnormal floats. instead of the correct significant digits, boa returns a bogus value with a wrong positive exponent.

**To Reproduce**
```js
Number.MIN_VALUE.toPrecision(1)
// returns "0e+101" in Boa, expected "5e-324"
```

From repo root:
```bash
cargo run --bin boa -- -e 'console.log(Number.MIN_VALUE.toPrecision(1))'
```

Boa output:
```
0e+101
```

Node.js output:
```bash
node -e 'console.log(Number.MIN_VALUE.toPrecision(1))'
# 5e-324
```

**Expected behavior**
`toPrecision()` should return correctly rounded significant digits for all finite values including subnormals, matching Node.js and browser behavior.

**Build environment:**
- OS: Windows 11
- Version: 10.0.26200

**Additional context**
Root cause is in `core/engine/src/builtins/number/mod.rs` - `format!("{this_num:.100}")` collapses subnormal values to all zeros, causing `flt_str_to_exp` to return a wrong positive exponent.

Spec reference: https://tc39.es/ecma262/#sec-number.prototype.toprecision

HiteshShonak:
This Pull Request fixes/closes #5225.

It changes the following:
- When `format!("{this_num:.100}")` collapses a subnormal float to all zeros, fall back to scientific notation formatting to correctly extract the mantissa and exponent.
- Add regression tests for `Number.MIN_VALUE.toPrecision(1)` and `toPrecision(2)`.

Testing:
```bash
cargo test -p boa_engine number -- --nocapture
```

Spec reference: https://tc39.es/ecma262/#sec-number.prototype.toprecision


jedel1043:
Should be fixed in `ryu-js` instead, and we have an issue about this (https://github.com/boa-dev/ryu-js/issues/33).

PR #5227, opened 2026-03-22
Implement `Iterator.prototype.includes` proposal

nekevss:
This PR implements the new Stage 2.7 Iterator includes proposal and adds some tests for it as well.

It's worth noting that the current tests don't take advantage of generator functions because our iterator helpers don't appear to work correctly with them, which we should probably look into. In the meantime, an arrow function returning an iterator was used.

I left the clamping of the skippedElements value to the range of `i64`, which I believe should be sufficient since it is beyond the MAX_SAFE_INTEGER.

Also worth noting, there's not test262 tests yet AFAIK, so no unflagging is needed currently.

CC: @michaelficarra for the visibility since we were talking about this just the other day


jedel1043 (approved):
LGTM!

PR #5231, opened 2026-03-22
fix(repl): Dot commands not working.

### LINKED ISSUE #5219: Issue in the dot commands of the REPL
Upon using the DOT commands of `.help` and `.exit`, a parsing error shows up instead of actually running the. commands. 

To reproduce this, just run the boa JavaScript runtime normally. 
The command I used: `cargo run --jobs 28`

What was supposed to happen is it will exit upon using `.exit` or show the help for `.help`

- OS: macOS
- Version: 26.3.1 (a)
- Rustc version: 1.94.0 (4a4ef493e 2026-03-02)

Image

gluzandii:
This Pull Request fixes/closes #5219.

This commit fixes the parsing error that shows when using `dot` commands in the REPL.
It fixes this by preventing the passing of the `dot` commands to the JS parser.

It changes the following:
- [cli/helper.rs](https://github.com/boa-dev/boa/blob/main/cli/src/helper.rs): Prevents passing of dot commands to JS parser.

### DIFF
--- a/cli/src/helper.rs
+++ b/cli/src/helper.rs
@@ -60,6 +60,11 @@ impl Validator for RLHelper {
         &self,
         context: &mut ValidationContext<'_>,
     ) -> Result<ValidationResult, ReadlineError> {
+        // Dot commands are handled by the REPL loop, not the JS parser.
+        if context.input().trim_start().starts_with('.') {
+            return Ok(ValidationResult::Valid(None));
+        }
+
         let mut parser = boa_parser::Parser::new(Source::from_bytes(context.input()));
         if self.strict {
             parser.set_strict();


cli/src/helper.rs
```diff
@@ -60,6 +60,12 @@ impl Validator for RLHelper {
         &self,
         context: &mut ValidationContext<'_>,
     ) -> Result<ValidationResult, ReadlineError> {
+        // Dot commands are handled by the REPL loop, not the JS parser.
+        let input = context.input().trim();
+        if input == ".help" || input == ".exit" || input.starts_with(".load") || input == ".clear" {
```
jedel1043: Don't use manual matching, just check if it starts with a dot. We'll keep extending those with more commands.

No JS syntax starts with a dot, so it doesn't matter in this case.

jedel1043 (changes requested):

cli/src/helper.rs
gluzandii: Yes that makes sense, understood.

HalidOdat (approved):
Looks good! Thanks for the contribution @gluzandii

jedel1043 (approved):

PR #5232, opened 2026-03-22
fix(vm/opcode/push,environment,templates,generator,await,new): convert panics to EngineError::Panic using js_expect

KaustubhOG:
Part of #3241.

Converts all convertible `.expect()` panics to `.js_expect()?` in:

- `opcode/push/array.rs` (4)
- `opcode/push/environment.rs` (2)
- `opcode/push/class/mod.rs` (1)
- `opcode/push/class/field.rs` (8)
- `opcode/push/class/private.rs` (12)
- `opcode/environment/mod.rs` (12)
- `opcode/templates/mod.rs` (7)
- `opcode/generator/mod.rs` (6)
- `opcode/generator/yield_stm.rs` (2)
- `opcode/await/mod.rs` (5)
- `opcode/new/mod.rs` (2)

*Total: 61 panics converted.*

### DIFF
--- a/core/engine/src/vm/opcode/await/mod.rs
+++ b/core/engine/src/vm/opcode/await/mod.rs
@@ -5,6 +5,7 @@ use crate::{
         Promise, async_generator::AsyncGenerator, generator::GeneratorContext,
         promise::PromiseCapability,
     },
+    error::PanicError,
     js_string,
     native_function::NativeFunction,
     object::FunctionObjectBuilder,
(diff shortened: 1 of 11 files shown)


KaustubhOG:
To this pr i tried one new thing 
For functions returning `ControlFlow<CompletionRecord>`, `?` is not available, so `js_expect` cannot be used directly. A new pattern was introduced using `let..else` + `handle_error` + `PanicError`:
```rust
let Some(value) = some_option else {
    return context.handle_error(PanicError::new("msg").into());
};
```

This ensures panics are still converted to `EngineError::Panic` without requiring a `JsResult` context. Happy to revert these if a different approach is preferred.

jedel1043 (approved):
Looks good! And pretty smart to use `handle_error` to handle panics.

PR #5233, opened 2026-03-23
refactor: extract type-check boilerplate into a single `require_internal_slot!` macro

Nakshatra480:
### Summary

This PR introduces a new `require_internal_slot!` macro to replace the repetitive 5-line [this](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/date/tests.rs:100:0-107:1)-type-checking boilerplate found across the engine's built-in prototype methods. 

Previously, almost every prototype method (across Temporal, Date, TypedArrays, etc.) manually downcasted `this` and returned a `TypeError` if the internal slot was missing:

```rust
let object = this.as_object();
let date = object
    .as_ref()
    .and_then(JsObject::downcast_ref::<Self>)
    .ok_or_else(|| JsNativeError::typ().with_message("'this' is not a Date"))?;
```

This PR refactors those occurrences into a single, much cleaner macro call that correctly handles the necessary borrowing and lifetime scoping:

`require_internal_slot!(date = this, Date, "Date");`

### Scope of changes
While originally intended for just the Temporal built-ins, I realized this boilerplate is heavily duplicated across the entire engine. Therefore, I've broadened the scope and applied the macro to all affected built-ins, including:

- Temporal (all modules)
- Date
- ArrayBuffer & SharedArrayBuffer
- DataView
- TypedArray
- WeakRef, WeakMap, WeakSet
- Intl (Collator, Segmenter, Segments)

### Error Messages & Testing
As part of standardizing this pattern, all error messages across the engine for missing internal slots have been unified to use the generic format: TypeError: "the this object must be a {Type} object.".

All existing tests (specifically in Builtin::Date) that previously relied on custom error strings like "'this' is not a Date" have been updated to expect the new standardized message. CI tests now pass cleanly with these updates.

### DIFF
--- a/core/engine/src/builtins/mod.rs
+++ b/core/engine/src/builtins/mod.rs
@@ -1,5 +1,26 @@
 //! Boa's ECMAScript built-in object implementations, e.g. Object, String, Math, Array, etc.
 
+/// Downcasts `$this` (a `&JsValue`) to a `Ref<$type>`, returning a `TypeError` if the
+/// cast fails. The error message uses `$name` as the type name.
+///
+/// This centralizes the repeated this-type-check boilerplate found across
+/// all builtin methods. The macro introduces a hidden binding for the
+/// owned `JsObject` to ensure it outlives the returned `Ref`, and binds the
+/// result to `$var`.
+///
+/// Usage: `require_internal_slot!(dt = this, Self, "PlainDateTime");`
+macro_rules! require_internal_slot {
+    ($var:ident = $this:expr, $type:ty, $name:expr) => {
+        let __temporal_obj = $crate::JsValue::as_object($this);
+        let $var = __temporal_obj
+            .as_ref()
+            .and_then($crate::JsObject::downcast_ref::<$type>)
+            .ok_or_else(|| {
+                $crate::js_error!(TypeError: "the this object must be a {} object.", $name)
+            })?;
+    };
+}
+
 pub mod array;
 pub mod array_buffer;
 pub mod async_function;
--- a/core/engine/src/builtins/mod.rs
+++ b/core/engine/src/builtins/mod.rs
@@ -1,23 +1,24 @@
 //! Boa's ECMAScript built-in object implementations, e.g. Object, String, Math, Array, etc.
 
-/// Downcasts `$this` (a `&JsValue`) to a `Ref<$type>`, returning a `TypeError` if the
-/// cast fails. The error message uses `$name` as the type name.
+/// Downcasts `$this` (a `&JsValue`) to a `JsObject<$type>`, returning a
+/// `TypeError` if the cast fails. The error message uses `$name` as the
+/// type name.
 ///
 /// This centralizes the repeated this-type-check boilerplate found across
-/// all builtin methods. The macro introduces a hidden binding for the
-/// owned `JsObject` to ensure it outlives the returned `Ref`, and binds the
-/// result to `$var`.
+/// all builtin methods. Unlike `downcast_ref`, the returned `JsObject<T>`
+/// does **not** hold an active borrow, so callers can `.borrow()` /
+/// `.borrow_mut()` only when needed, avoiding panics from long-lived
+/// `GcRefCell` borrows.
 ///
-/// Usage: `require_internal_slot!(dt = this, Self, "PlainDateTime");`
+/// Usage: `let dt = require_internal_slot!(this, Self, "PlainDateTime");`
 macro_rules! require_internal_slot {
-    ($var:ident = $this:expr, $type:ty, $name:expr) => {
-        let __temporal_obj = $crate::JsValue::as_object($this);
-        let $var = __temporal_obj
-            .as_ref()
-            .and_then($crate::JsObject::downcast_ref::<$type>)
+    ($this:expr, $type:ty, $name:expr) => {
+        $crate::JsValue::as_object($this)
+            .map(|o| o.clone())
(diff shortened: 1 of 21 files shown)


core/engine/src/builtins/temporal/mod.rs
```diff
@@ -7,6 +7,31 @@
...
+/// Downcasts `$this` (a `&JsValue`) to a `Ref<$type>`, returning a `TypeError` if the
+/// cast fails. The error message uses `$name` as the type name.
+///
+/// This centralizes the repeated this-type-check boilerplate found across
+/// all Temporal builtin methods. The macro introduces a hidden binding for the
+/// owned `JsObject` to ensure it outlives the returned `Ref`, and binds the
+/// result to `$var`.
+///
+/// Usage: `this_temporal_object!(dt = this, Self, "PlainDateTime");`
+macro_rules! this_temporal_object {
```
nekevss: nit: rename to `require_internal_slot`

core/engine/src/builtins/temporal/mod.rs
```diff
@@ -7,6 +7,31 @@
...
+    ($var:ident = $this:expr, $type:ty, $name:expr) => {
+        let __temporal_obj = $crate::JsValue::as_object($this);
+        let $var = __temporal_obj
+            .as_ref()
+            .and_then($crate::JsObject::downcast_ref::<$type>)
+            .ok_or_else(|| {
+                $crate::JsError::from($crate::error::JsNativeError::typ().with_message(concat!(
```
nekevss: nit: use `js_error` macro here

nekevss (commented):
I do sort of like the change for how succinct it is, and it's potential for making the code more readable.

One question: why change just the temporal objects and not the built-ins in general? Had you looked at other built-ins?

core/engine/src/builtins/temporal/mod.rs
Nakshatra480: Done, renamed.

core/engine/src/builtins/temporal/mod.rs
Nakshatra480: Done, switched to js_error!(TypeError: ...).

Nakshatra480:
Hey @nekevss Thanks for the review!

To answer your question: Yes, while originally working on Temporal, I realized that the 5-line [this](cci:1://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/date/tests.rs:100:0-107:1)-type-check boilerplate is heavily duplicated across almost all built-ins. I have now expanded the scope of this PR to apply the macro across the entire engine, not just Temporal.

Here is a summary of the updates in the latest push:
1. **Scope Expanded**: Applied the macro to all affected built-ins, including [Date](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/date/mod.rs:42:0-42:21), `ArrayBuffer`, `SharedArrayBuffer`, [DataView](cci:2://file:///Users/nakshatrasharma/Desktop/boa/core/engine/src/builtins/dataview/mod.rs:38:0-42:1), `TypedArray`, `WeakRef`, `WeakMap`, `WeakSet`, and `Intl` (`Collator`, `Segmenter`, `Segments`), replacing hundreds of lines of repetitive downcasting boilerplate.
2. **Macro Renamed**: Renamed the macro to `require_internal_slot!` as suggested.
3. **`js_error!` Used**: Updated the macro to use `js_error!` internally as requested.
4. **Error Messages**: The macro standardizes the error message to `TypeError: "the this object must be a {Type} object."`. A few tests (specifically in `Builtin::Date`) that explicitly asserted the old `"'this' is not a Date"` error string have been updated to expect the new standardized message. 

All CI tests, including `cargo fmt` and `cargo clippy`, now pass cleanly. Let me know if there's anything else you'd like me to adjust!

core/engine/src/builtins/mod.rs
```diff
@@ -1,5 +1,26 @@
...
+/// owned `JsObject` to ensure it outlives the returned `Ref`, and binds the
+/// result to `$var`.
+///
+/// Usage: `require_internal_slot!(dt = this, Self, "PlainDateTime");`
+macro_rules! require_internal_slot {
+    ($var:ident = $this:expr, $type:ty, $name:expr) => {
+        let __temporal_obj = $crate::JsValue::as_object($this);
+        let $var = __temporal_obj
+            .as_ref()
+            .and_then($crate::JsObject::downcast_ref::<$type>)
```
jedel1043: I'd say if we're doing this change, I would modify it to use `JsObject::downcast` instead. That way you can use normal expr syntax `let o = require_internal_slot!()` and it also makes it such that we don't `borrow_mut` objects for whole functions, which could cause panics.

jedel1043 (changes requested):

core/engine/src/builtins/mod.rs
Nakshatra480: Thanks for the feedback! I've updated the require_internal_slot! macro to return an owned JsObject<T> using JsObject::downcast as suggested. I've also refactored all call sites across the engine to use let obj = require_internal_slot!(...) and read/write via .borrow().data() or .borrow_mut().data_mut(), which successfully prevents the long-lived GcRefCell borrows and avoids the re-entry panics.

Nakshatra480:
Hey @jedel1043 @nekevss I've pushed the requested changes using JsObject::downcast, and all builds and tests are now passing cleanly. The PR is ready for another look whenever you have time!

PR #5234, opened 2026-03-23
Support running test262 using boa cli

### LINKED ISSUE #5054: Expose full $262 object in boa_cli
**ECMASCript feature**

I'm interested in running test262 on boa eshost-style - using boa_cli shell rather than the separate integrated tester binary (boa_tester). This requires $262 object, which boa_cli doesn't directly expose -  there's $boa (with --debug-object) but 1) it requires maintaining a custom shim to map $boa <-> $262 methods which is error- and labor-prone (especially for complex logic like `$262.agent`) and 2) doesn't provide some methods, like detachArrayBuffer and now the recently added IsHTMLDDA, so there are discrepancies in test results.

Could you the expose the full internal $262 object ([js262.rs](https://github.com/boa-dev/boa/blob/main/tests/tester/src/exec/js262.rs)) in boa_cli? I suppose --debug-object flag can be reused to add both $boa and $262.

\[spec\]: https://github.com/tc39/test262/blob/main/INTERPRETING.md#host-defined-functions

**Example code**

Properties like `$262.detachArrayBuffer`, `$262.agent`, `$262.IsHTMLDDA` should be available with `boa --debug-object` (or another flag)

ivankra:
This Pull Request fixes/closes #5054

It changes the following:

- Move `tests/tester/src/exec/js262.rs` to `core/runtime/src/test262.rs` so it can be reused by both cli and boa_tester
- Add `--test262-object` flag to CLI
- Add `print()` to CLI with `--test262-object`
- Enable `[[CanBlock]]`  by default in CLI to match boa_tester as well as d8/jsc/sm shells default behavior
- Add `--no-can-block` flag to CLI to support Atomics CanBlockIsFalse tests (same flag name as d8 has)
- Auto-detect .mjs files as modules in CLI (a small optional quality-of-life improvement; matches behavior of most other major JS engine shells)

These changes make it possible to reproduce boa_tester's test262 results using external test262 harnesses with boa cli binary.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -153,6 +153,10 @@ struct Opt {
     #[arg(long)]
     debug_object: bool,
 
+    /// Inject the test262 host object `$262`.
+    #[arg(long)]
+    test262_object: bool,
+
     /// Treats the input files as modules.
     #[arg(long, short = 'm', group = "mod")]
     module: bool,
@@ -560,6 +564,14 @@ fn main() -> Result<()> {
 
     if args.debug_object {
         init_boa_debug_object(context);
+    }
+
+    if args.test262_object {
+        boa_runtime::test262::register_js262(
+            boa_runtime::test262::WorkerHandles::new(),
+            true, // register `console` in $262.agent worker threads
+            context,
+        );
     }
 
     // Configure optimizer options
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -157,6 +157,10 @@ struct Opt {
     #[arg(long)]
     test262_object: bool,
 
+    /// Disallow the main thread from blocking (e.g. `Atomics.wait`).
+    #[arg(long)]
+    no_can_block: bool,
+
     /// Treats the input files as modules.
     #[arg(long, short = 'm', group = "mod")]
     module: bool,
@@ -558,6 +562,7 @@ fn main() -> Result<()> {
     let context = &mut ContextBuilder::new()
         .job_executor(executor.clone())
         .module_loader(loader.clone())
+        .can_block(!args.no_can_block)
         .build()
         .map_err(|e| eyre!(e.to_string()))?;
 
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -289,13 +289,18 @@ impl Drop for Counters {
 ///
 /// Returns a error of type String with a error message,
 /// if the source has a syntax or parsing error.
-fn dump<R: ReadChar>(src: Source<'_, R>, args: &Opt, context: &mut Context) -> Result<()> {
+fn dump<R: ReadChar>(
+    src: Source<'_, R>,
+    args: &Opt,
+    is_module: bool,
+    context: &mut Context,
+) -> Result<()> {
     if let Some(arg) = args.dump_ast {
         let mut counters = Counters::new(args.time);
         let arg = arg.unwrap_or_default();
         let mut parser = boa_parser::Parser::new(src);
         let dump =
-            if args.module {
+            if is_module {
                 let scope = context.realm().scope().clone();
                 let module = {
                     let _timer = counters.new_timer("Parsing");
@@ -390,7 +395,7 @@ fn evaluate_expr(
     printer: &SharedExternalPrinterLogger,
 ) -> Result<()> {
     if args.has_dump_flag() {
-        dump(Source::from_bytes(line), args, context)?;
+        dump(Source::from_bytes(line), args, args.module, context)?;
     } else if let Some(flowgraph) = args.flowgraph {
         match generate_flowgraph(
             context,
@@ -437,8 +442,11 @@ fn evaluate_file(
     loader: &SimpleModuleLoader,
     printer: &SharedExternalPrinterLogger,
 ) -> Result<()> {
+    // Treat files with .mjs extension automatically as modules.
+    let is_module = args.module || file.extension().is_some_and(|ext| ext == "mjs");
+
     if args.has_dump_flag() {
-        return dump(Source::from_filepath(file)?, args, context);
+        return dump(Source::from_filepath(file)?, args, is_module, context);
     }
 
     if let Some(flowgraph) = args.flowgraph {
@@ -454,7 +462,7 @@ fn evaluate_file(
         return Ok(());
     }
 
-    if args.module {
+    if is_module {
         let source = Source::from_filepath(file)?;
         let mut counters = Counters::new(args.time);
         let module = {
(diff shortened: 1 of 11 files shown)


core/runtime/src/test262.rs
```diff
@@ -202,9 +203,7 @@ fn sleep(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsVal
 /// The `$262.agent.monotonicNow()` function.
 #[allow(clippy::unnecessary_wraps)]
 fn monotonic_now(_: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValue> {
-    let clock = START
-        .get()
-        .ok_or_else(|| JsNativeError::typ().with_message("could not get the monotonic clock"))?;
+    let clock = START.get_or_init(Instant::now);
```
jedel1043: I don't think this is correct. This will always return 0 or something close to 0 for the first call to `monotonic_now`, which could fail some tests that assume some amount of time already passed

ivankra: Thanks, I think you're right -- looking at the [spec](https://www.w3.org/TR/hr-time-2/#sec-time-origin) for DOMHighResTimeStamp, it does suggest to use worker/binary's start time as the time origin.

I moved initialization of START to the beginning of `register_js262()` ([diff](https://github.com/boa-dev/boa/compare/c0b0380b2b63ec4b0d67b74ebeb2052173133d7b..4c52207e66e1e1fe37d9c0702f4d3c77f6891303)) and tested:
```sh
$ ./target/release/boa --test262-object -e '$262.agent.sleep(100); print($262.agent.monotonicNow());'
100
undefined
```
(before it printed 0 as clock got initialized on first use)

test262 tests aren't affected, no regressions either way.

ivankra:
Fixed lint errors in lib.rs / test262.rs and rebased.

core/runtime/src/test262.rs
```diff
@@ -205,6 +226,7 @@ fn monotonic_now(_: &JsValue, _: &[JsValue], _: &mut Context) -> JsResult<JsValu
     let clock = START
         .get()
         .ok_or_else(|| JsNativeError::typ().with_message("could not get the monotonic clock"))?;
+    #[allow(clippy::cast_precision_loss)]
     Ok(JsValue::from(clock.elapsed().as_millis() as f64))
```
jedel1043: Forgot to mention that having an external clock might not be necessary anymore. We introduced a `Clock` trait that can be used to get a monotonic `now` with `context.clock().now().millis_since_epoch()`

ivankra: Ah cool, I've update the patch to use it.

ivankra:
Rebased again.

ivankra:
I ran `./target/release/boa_tester run` with default configuration (test262 commit: 72ec758c) - output is unchanged between f5e88de558e038f0ae675a012d59917a098f44b6 and  85df98d49e4a6cf373c5aec1d66e1e90ac33af54, so I think GitHub CI diff above is flaky:

```
Results (ECMAScript Next):
Total tests: 52963
Passed tests: 50745
Ignored tests: 1426
Failed tests: 792 (0 panics)
Conformance: 95.81%
```

Testing boa cli with my patches and my [own test262 harness](https://github.com/ivankra/javascript-zoo/blob/master/harness/test262.py) on the same test262 commit vs boa_tester I'm down to just a handful of diffs:
```
test/built-ins/Atomics/waitAsync/bigint/true-for-timeout.js: "AsyncTestFailure: Test timed out" vs OK
test/built-ins/Atomics/waitAsync/returns-result-object-value-is-promise-resolves-to-timed-out.js: "AsyncTestFailure: Test timed out" vs OK
test/built-ins/Atomics/waitAsync/true-for-timeout.js: "AsyncTestFailure: Test timed out" vs OK
test/language/import/import-defer/errors/syntax-error/import-defer-of-syntax-error-fails.js: "OK" vs FAIL
test/language/module-code/source-phase-import/import-source.js: "OK" vs FAIL
test/staging/sm/strict/8.7.2.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/sm/strict/8.12.7-2.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/sm/strict/10.4.2.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/sm/strict/10.6.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/sm/strict/12.10.1.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/sm/strict/15.3.5.1.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/sm/strict/15.3.5.2.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/sm/strict/15.5.5.2.js: "Test262Error: Expected SameValue(«false», «true») to be true" vs OK
test/staging/source-phase-imports/import-source-source-text-module.js: "AsyncTestFailure: Promise should be rejected with SyntaxError Expected a SyntaxError to be thrown asynchronously but got TypeError" vs OK
```

test/staging/sm/strict/ I think would be better fixed upstream in test262 repo: https://github.com/tc39/test262/pull/5008

jedel1043:
Nah the CI diff hasn't changed because actions haven't run haha

cli/src/main.rs
```diff
@@ -153,6 +153,14 @@ struct Opt {
     #[arg(long)]
     debug_object: bool,
 
+    /// Inject the test262 host object `$262`.
+    #[arg(long)]
+    test262_object: bool,
+
+    /// Disallow the main thread from blocking (e.g. `Atomics.wait`).
+    #[arg(long)]
+    no_can_block: bool,
```
jedel1043, line 162: It might be better to err on the safe side and have a `can_block` option instead, defaulting to `false` for the rest of the CLI users, which ensures they can't just block the whole thread by calling `Atomics.wait`

ivankra: I'd expect blocking the default behavior from a CLI app - that's how most javascript shells and node behave.

ivankra: node doesn't even have a flag to disable it. Only `d8` has `--no-can-block` flag - undocumented and only used for test262.

jedel1043: Oh forgot that Web Workers exist. Was gonna say that `Atomics.wait` was useless for normal CLI users, but that could change after implementing Web Workers, so we can leave it like this then

jedel1043 (approved):
Thanks!

PR #5235, opened 2026-03-23
fix(runtime): support cloning DataView objects

### LINKED ISSUE #5167: structuredClone() / postMessage() reject DataView values
Ran into this while checking structured clone support.

```js
const view = new DataView(new ArrayBuffer(4));
structuredClone(view);
```

Current result:

```js
TypeError: Data views are not supported yet.
```

This comes from `core/runtime/src/store/from.rs`, and it affects both `structuredClone()` and `postMessage()` since both go through `JsValueStore::try_from_js(...)`.

This also looks like unfinished support rather than an intentional gap:

- `JsValueStore` already has a `ValueStoreInner::DataView` variant in `core/runtime/src/store/mod.rs`
- `core/runtime/src/store/to.rs` already has a `DataView` reconstruction path

There is direct WPT coverage for cloning `DataView` values in:

- `tests_wpt/html/webappapis/structured-clone/structured-clone-battery-of-tests.js`
- `tests_wpt/html/webappapis/structured-clone/structured-clone-battery-of-tests-with-transferables.js`

So the main issue here is just that Boa rejects `DataView` values outright right now.

yush-1018:
Fixes #5167

This PR adds support for `structuredClone()` to clone `DataView` objects. Let me know if anything needs changes!

### DIFF
--- a/core/macros/src/module.rs
+++ b/core/macros/src/module.rs
@@ -187,14 +187,14 @@ fn module_impl_impl(_args: ModuleArguments, mut mod_: ItemMod) -> SpannedResult<
             | Item::TraitAlias(ItemTraitAlias { attrs, .. })
             | Item::Type(ItemType { attrs, .. })
             | Item::Union(ItemUnion { attrs, .. })
-            | Item::Use(ItemUse { attrs, .. }) => {
-                if take_path_attr(attrs, "skip") {
-                    original_module_decl = quote! {
-                        #original_module_decl
-                        #item
-                    };
-                    continue;
-                }
+            | Item::Use(ItemUse { attrs, .. })
+                if take_path_attr(attrs, "skip") =>
+            {
+                original_module_decl = quote! {
+                    #original_module_decl
+                    #item
+                };
+                continue;
             }
             _ => {}
         }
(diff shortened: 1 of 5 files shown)


core/engine/src/builtins/iterable/iterator_constructor.rs
```diff
@@ -247,6 +263,328 @@ impl IteratorConstructor {
...
+    // ==================== Static Methods — Experimental ====================
+
+    #[cfg(feature = "experimental")]
+    /// `Iterator.zip ( iterables [ , options ] )`
+    ///
+    /// More information:
+    ///  - [TC39 proposal][spec]
+    ///
+    /// [spec]: https://tc39.es/proposal-joint-iteration/#sec-iterator.zip
+    fn zip(_: &JsValue, args: &[JsValue], context: &mut Context) -> JsResult<JsValue> {
```
jedel1043: I think you pushed unrelated commits from another PR 😅

jedel1043 (changes requested):

yush-1018:
Sorry about that! I've force-pushed with only the relevant DataView fix commit now. Could you re-review? 

Thanks!

yush-1018:
Please review the PR once.

jedel1043:
Closing since the PR is stale and not in a working state.

PR #5236, opened 2026-03-23
Cleanup iterator helpers

jedel1043:
Kinda big refactor of our iterator helpers, but this makes it such that they reuse the same coroutine machinery that `Array.fromAsync` uses, removing many of the `expect` calls that were in place, which was a consequence of having to use an enum to represent all iterator helpers.

This also cleans up the intrinsic initialization code, which was kinda wrong in some places and caused a bug where the iterator returned by generator functions wouldn't have the iterator prototype setup correctly.

This puts us at 100% conformance (not counting joint iterators) for the `Iterator.prototype` suite BTW


core/engine/src/builtins/iterable/iterator_prototype.rs
```diff
@@ -0,0 +1,776 @@
...
+            // e. If ToBoolean(result) is true, return ? IteratorClose(iterated, NormalCompletion(value)).
+            if result.to_boolean() {
+                return iterated.close(Ok(value), context);
+            }
+
+            // f. Set counter to counter + 1.
+            counter += 1;
+        }
+
+        // b. If value is done, return undefined.
```
nekevss: nit: can we keep the steps in line and reference to them in comments

i.e. 

```txt
// a. step text
// b. step text - DEFERRED TO END OF FUNCTION
```

And then the end of the function would be: `// Step 7.b.`

Another option might be to keep `b` here, move `a`, and then reference that step a-b are moved in the while loop block.

core/engine/src/builtins/iterable/iterator_prototype.rs
```diff
@@ -0,0 +1,776 @@
...
+        };
+        // 5. Set iterated to ? GetIteratorDirect(O).
+        let mut iterated = get_iterator_direct(iterated.iterator(), context)?;
+
+        // 6. Let counter be 0.
+        let mut counter = 0u64;
+
+        // 7. Repeat,
+        while let Some(value) = iterated.step_value(context)? {
+            // a. Let value be ? IteratorStepValue(iterated).
```
nekevss: nit: move above the while loop

nekevss (approved):
Had a couple nits regarding the docs. But nothing blocking.

LGTM!

core/engine/src/builtins/iterable/iterator_prototype.rs
jedel1043: Adjusted the docs to preserve the original order

PR #5238, opened 2026-03-23
fix(math): fix double rounding in Math.f16round

### LINKED ISSUE #5237: Math.f16round gives wrong results for some values
I was going through the test262 conformance results and noticed that 
Math.f16round is failing the value-conversion.js test.

After digging into it, I found that Math.f16round(1.00048828125000022204) 
returns 1 instead of the correct value 1.0009765625.

The problem seems to be in how the float16 crate converts from f64 to f16 
internally — it goes through f32 first, which causes double rounding and 
produces the wrong result for certain inputs.

The fix would be to use the half crate's f16::from_f64 which is already 
in the dependency tree and does the conversion correctly without going 
through f32.

ParthMozarkar:
Hey! I noticed Math.f16round was failing the test262 value-conversion.js 
test and dug into why.

The issue is that the float16 crate's from_f64 converts through f32 
internally, which causes double rounding for certain values.

For example Math.f16round(1.00048828125000022204) was returning 1 
instead of the correct 1.0009765625.

I fixed it by switching to the half crate's f16::from_f64 which is 
already in the dependency tree and converts directly from f64 to f16 
without any intermediate f32 conversion.

All existing tests pass.

Fixes #5237


zhuzhu81998:
switching to half does not solve the problem.
on ur branch:
image

because half seems to do the same thing. the reason is probably what the float16 comment said in the FIXME:
```rust
#[inline]
#[target_feature(enable = "f16c")]
pub(super) unsafe fn f64_to_f16_x86_f16c(f: f64) -> u16 {
    // FIXME: Change to use `_mm_cvtpd_ph` when stable.
    f32_to_f16_x86_f16c(f as f32)
}

```

hansl:
Correct. `float16` is a fork of `half` because `half` had too many dependencies (crunchy and zerocopy). There is also a PR on `float16` that might fix this: https://github.com/Alexhuszagh/float16/pull/16

Have you tried it?

jedel1043:
Closing since this doesn't fix the issue, and the issue seems to be fixed by a PR upstream.

PR #5240, opened 2026-03-23
fix(math): correct Math.asinh for large finite inputs

### LINKED ISSUE #5239: Math.asinh returns Infinity for large finite inputs like 1e308 and Number.MAX_VALUE
**Describe the bug**
`Math.asinh()` returns `Infinity` for very large finite inputs. instead of a correct finite result, boa returns `Infinity` due to an internal overflow in Rust's `f64::asinh()`.

**To Reproduce**
```javascript
Math.asinh(1e308)
// returns Infinity in Boa, expected 709.889355822726
```

From repo root:
```bash
cargo run --bin boa -- -e 'console.log(Math.asinh(1e308))'
```

Boa output:
```
Infinity
```

Node.js output:
```bash
node -e 'console.log(Math.asinh(1e308))'
# 709.889355822726
```

`Number.MAX_VALUE` and negative large inputs are also affected:
```bash
cargo run --bin boa -- -e 'console.log(Math.asinh(Number.MAX_VALUE))'
# Boa:  Infinity
# Node: 710.4758600739439

cargo run --bin boa -- -e 'console.log(Math.asinh(-1e308))'
# Boa:  -Infinity
# Node: -709.889355822726
```

**Expected behavior**
`Math.asinh()` should return a correct finite result for all large finite inputs, matching Node.js and browser behavior.

**Build environment:**
* OS: Windows 11
* Version: 10.0.26200
* Target triple: x86_64-pc-windows-msvc
* Rustc version: rustc 1.94.0 (4a4ef493e 2026-03-02)

**Additional context**
Root cause is in `core/engine/src/builtins/math/mod.rs` - Rust's `f64::asinh()` overflows for inputs with absolute value above around `1.3e154`, producing `Infinity`.

Spec reference: https://tc39.es/ecma262/#sec-math.asinh

HiteshShonak:
This Pull Request fixes/closes #5239.

It changes the following:
- When the absolute value of the input is finite and greater than `1/√f64::EPSILON` (`67_108_864.0`), fall back to `n.signum() * (n.abs().ln() + LN_2)` instead of calling `f64::asinh()` directly, avoiding an internal overflow that produces `Infinity`. threshold follows the [Boost math library convention](https://www.boost.org/doc/libs/latest/libs/math/doc/html/math_toolkit/inv_hyper/asinh.html).
- Add regression tests for `Math.asinh(1e308)`, `Math.asinh(-1e308)`, and `Math.asinh(Number.MAX_VALUE)`.

Testing:
```bash
cargo test -p boa_engine math -- --nocapture
```

Spec reference: https://tc39.es/ecma262/#sec-math.asinh


core/engine/src/builtins/math/tests.rs
```diff
@@ -40,6 +40,9 @@ fn asinh() {
     run_test_actions([
         TestAction::assert_eq("Math.asinh(1)", 0.881_373_587_019_543),
         TestAction::assert_eq("Math.asinh(0)", 0.0),
+        TestAction::assert_eq("Math.asinh(1e308)", 709.889_355_822_726_f64),
+        TestAction::assert_eq("Math.asinh(-1e308)", -709.889_355_822_726_f64),
+        TestAction::assert_eq("Math.asinh(Number.MAX_VALUE)", 710.475_860_073_943_9_f64),
```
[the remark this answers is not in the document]
HiteshShonak: all platform tests passed with assert_eq. keeping it consistent with the existing asinh and acosh tests in the same file

core/engine/src/builtins/math/mod.rs
```diff
@@ -194,14 +194,18 @@ impl Math {
...
-            // 3. Return an implementation-approximated value representing the result of the inverse hyperbolic sine of ℝ(n).
-            .asinh()
-            .into())
+        // 1/√f64::EPSILON, as established by the Boost math library.
+        const ASINH_LARGE_INPUT_THRESHOLD: f64 = 67_108_864.0;
+
+        // 1. Let n be ? ToNumber(x).
+        let n = args.get_or_undefined(0).to_number(context)?;
+        if n.is_finite() && n.abs() > ASINH_LARGE_INPUT_THRESHOLD {
+            return Ok((n.signum() * (n.abs().ln() + std::f64::consts::LN_2)).into());
```
jedel1043: Do we want to use the better approximation here? (ln |2x| + 1/(4x^2))

We can decompose it to avoid overflowing into infinity:
```
ln 2x + 1/(4x^2) = ln x + ln 2 + (1/2 * 1/x)(1/2 * 1/x)
```

Veercodeprog: could you check this:
i have made updates as you suggested
https://github.com/boa-dev/boa/pull/5241

jedel1043: @Veercodeprog don't try to take over already assigned work, please.

Veercodeprog: are there any unassigned issues?

HiteshShonak: @jedel1043 updated to use the more accurate approximation ln(|x|) + ln(2) + 1/(4x²) as suggested, thanks

core/engine/src/builtins/math/mod.rs
```diff
@@ -194,14 +194,20 @@ impl Math {
...
-            .asinh()
-            .into())
+        // 1/√f64::EPSILON, as established by the Boost math library.
+        const ASINH_LARGE_INPUT_THRESHOLD: f64 = 67_108_864.0;
+
+        // 1. Let n be ? ToNumber(x).
+        let n = args.get_or_undefined(0).to_number(context)?;
+        if n.is_finite() && n.abs() > ASINH_LARGE_INPUT_THRESHOLD {
+            return Ok(
+                (n.signum() * (n.abs().ln() + std::f64::consts::LN_2 + (0.25 / (n * n)))).into(),
```
jedel1043:
```suggestion
                (n.signum() * (n.abs().ln() + std::f64::consts::LN_2 + 0.25 * n.recip() * n.recip())).into(),
```

Since `n*n` could overflow to infinity, making that factor always zero.

core/engine/src/builtins/math/mod.rs
```diff
@@ -198,14 +198,20 @@ impl Math {
...
-            .asinh()
-            .into())
+        // 1/√f64::EPSILON, as established by the Boost math library.
+        const ASINH_LARGE_INPUT_THRESHOLD: f64 = 67_108_864.0;
+
+        // 1. Let n be ? ToNumber(x).
+        let n = args.get_or_undefined(0).to_number(context)?;
+        if n.is_finite() && n.abs() > ASINH_LARGE_INPUT_THRESHOLD {
+            return Ok((n.signum()
+                * (n.abs().ln() + std::f64::consts::LN_2 + 0.25 * n.recip() * n.recip()))
```
zhuzhu81998: even tho compiler probably optimizes this any way, it is imo better to avoid doing `recip` twice.

zhuzhu81998 (changes requested):

core/engine/src/builtins/math/mod.rs
HiteshShonak: n*n overflows for large inputs, n.recip()*n.recip() sidesteps that, compiler CSE handles the rest

zhuzhu81998: what i mean is introducing a variable that saves `n.recip`...

jedel1043: I think it's clearer like this. I would be very surprised if CSE doesn't trigger here

jedel1043 (approved):
Looks great! Thanks

core/engine/src/builtins/math/mod.rs
HiteshShonak: Done, extracted n.recip() into let r for clarity

PR #5242, opened 2026-03-23
Veercodeprog/feat/5144 boa class plain properties

### LINKED ISSUE #5144: Support defining plain properties in `#[boa_class]`
Currently we only support defining methods in a `#[boa_class]` declaration, as such:
```rust
#[boa_class]
impl Animal {
    #[boa(constructor)]
    #[allow(clippy::needless_pass_by_value)]
    fn new(name: String, age: i32) -> Self {
        let ty = match name.as_str() {
            "cat" => AnimalType::Cat,
            "dog" => AnimalType::Dog,
            _ => AnimalType::Other,
        };

        Self { ty, age }
    }
}
```

However, some users will also want to define plain properties on their classes and class instances, but there are still questions on where to put those:

- `const PROPERTY: T = value()` variables, but it is kind of weird from rust's perspective, and might require special macro support.
- Enable some kind of `#[boa(js_constructor)]` that takes the already constructed object for additional initialization.
- Modify `#[boa(constructor)]` such that it can also be used to initialize properties. Might require passing something like `MaybeUninit`, or maybe passing a nullable `Object` (not JsObject) so that folks can initialize their properties, then inject the returned native type into the modified `Object`.

A lot of design decisions here, but seems feasible by the list of possible solutions. We might also need to modify the `Class` trait directly, since that's kind of the trait that provides support for `boa_class` itself.

(didn't cover static properties, but those have their own set of design decisions)

Veercodeprog:
# feat(macros): add `#[boa(js_init)]` support to `#[boa_class]`

Closes #5144 (partial — instance-side initialization only)

## Summary

Introduces the `#[boa(js_init)]` attribute to the `#[boa_class]` macro,
allowing embedders to hook into `object_constructor` without having to
hand-write a full `impl Class` block. This enables setting own data
properties on a freshly constructed `JsObject` instance — the first
concrete step toward the plain-property support discussed in #5144.

## Motivation

Currently, `#[boa_class]` only exposes `data_constructor` via
`#[boa(constructor)]`. If you want to set own JS properties after the
object exists (e.g. `instance.upcast().set("label", ...)`), you have to
drop out of the macro entirely and write `impl Class` manually. The
`object_constructor` hook already exists on the `Class` trait for
exactly this purpose — this PR just wires it up through the macro.

### `core/macros/src/class.rs`
- Parse `#[boa(js_init)]` on at most one function per `#[boa_class] impl` block.
- Validate the expected shape: no `self`, no generics or `async`, exactly
  three parameters (`instance`, `args`, `context`), return type must be
  `()`, `JsResult<()>`, or omitted (normalized to `()`).
- Emit `fn object_constructor(...)` in the generated `impl Class` when a
  `js_init` function is present; otherwise fall back to the no-op trait
  default.
- Exclude `js_init`-annotated functions from method/static registration
  so they are not accidentally exposed as JS-callable methods.

### `core/macros/src/lib.rs`
- Add `#[boa(js_init)]` to the `boa_class` attribute documentation with
  a brief usage example.

### `tests/macros/tests/class.rs`
- New `Widget` class and `boa_class_js_init_sets_own_property` test:
  `#[boa(constructor)]` stores the native `id` field; `#[boa(js_init)]`
  sets an own data property `label` from a second constructor argument,
  verifying end-to-end that instance-side initialization works.

## Scope

**Covered by this PR:**
- Instance-side property initialization via the existing
  `object_constructor` trait hook.

**Out of scope / follow-up work for #5144:**
- Prototype and static property declarations
  (`#[boa(property)]` / `#[boa(static_property)]` → `ClassBuilder::property`).
- Declarative `const`-style properties, particularly where values involve
  non-`const` types such as `JsValue`.
- Unifying `data_constructor` and `object_constructor` into a single
  constructor model (if the project pursues that direction).
- Proc-macro-level type validation for `js_init` parameters — wrong types
  still produce a compile error at the forwarded `Self::…` call site.

## Testing

```bash
# Run the new macro integration test
cargo test -p boa_macros_tests --test class

# Lint check
cargo clippy -p boa_macros --all-targets -- -D warnings

### DIFF
--- a/core/engine/src/builtins/math/mod.rs
+++ b/core/engine/src/builtins/math/mod.rs
@@ -20,14 +20,17 @@ use crate::{
 use super::{BuiltInBuilder, IntrinsicObject};
 
 /// For very large finite `|x|`, `f64::asinh` can overflow internally when forming `x²`.
-/// In that regime `asinh(x)` is well approximated by `sign(x) · (ln|x| + ln 2)`.
+/// In that regime `asinh(x) ≈ ln(2|x|) + 1/(4x²)`; the correction is computed as `(1/(2x))²`
+/// so we never square `x` (which would overflow).
 fn asinh_f64(n: f64) -> f64 {
     if n.is_nan() || n == 0.0 || n.is_infinite() {
         return n;
     }
     let ax = n.abs();
     if (ax * ax).is_infinite() {
-        n.signum() * (ax.ln() + std::f64::consts::LN_2)
+        let ln_2x = ax.ln() + std::f64::consts::LN_2;
+        let inv_2x = 0.5 / ax;
+        n.signum() * (ln_2x + inv_2x * inv_2x)
     } else {
         n.asinh()
     }
--- a/core/engine/src/builtins/math/mod.rs
+++ b/core/engine/src/builtins/math/mod.rs
@@ -19,6 +19,20 @@ use crate::{
 
 use super::{BuiltInBuilder, IntrinsicObject};
 
+/// For very large finite `|x|`, `f64::asinh` can overflow internally when forming `x²`.
+/// In that regime `asinh(x)` is well approximated by `sign(x) · (ln|x| + ln 2)`.
+fn asinh_f64(n: f64) -> f64 {
+    if n.is_nan() || n == 0.0 || n.is_infinite() {
+        return n;
+    }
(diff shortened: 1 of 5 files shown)


jedel1043:
The original issue still needs a proper design, so closing this until the issue gets properly triaged

PR #5244, opened 2026-03-23
fix(number): fix Infinity casing in Number.prototype.toExponential

### LINKED ISSUE #5243: Number.prototype.toExponential returns wrong casing for Infinity and -Infinity
Hey, I was going through the test262 conformance results and noticed 
that Number.prototype.toExponential is failing the return-values.js test.

The issue is that:
Infinity.toExponential() returns "infinity" instead of "Infinity"
(-Infinity).toExponential() returns "-infinity" instead of "-Infinity"

The spec says in step 4: "If x is not finite, return ! Number::toString(x)"
which should return "Infinity" or "-Infinity" with capital I.

The bug is that the current implementation uses JsString::from(this_num) 
which relies on Rust's default float formatting that produces lowercase 
"infinity". The fix is to explicitly return the correct strings "Infinity", 
"-Infinity" and "NaN" for non-finite values.

Happy to submit a fix for this!

ParthMozarkar:
I noticed Number.prototype.toExponential was failing the test262 
return-values.js test and dug into why.

Infinity.toExponential() was returning "infinity" instead of "Infinity"
(-Infinity).toExponential() was returning "-infinity" instead of "-Infinity"

The issue was that the non-finite check was using JsString::from(this_num) 
which relies on Rust's default float formatting that produces lowercase 
"infinity".

Fixed it by explicitly returning the correct strings "Infinity", 
"-Infinity" and "NaN" for non-finite values, matching what the 
ECMAScript spec requires.

All existing tests pass.

Fixes #5243


zhuzhu81998:
main works fine for me. and the test262 bot does not report any newly fixed tests.

image

jedel1043:
Closing since ryu-js already outputs the correct formatting.

Please verify that your PRs at least fix something.

PR #5255, opened 2026-03-24
fix(temporal): align Duration toLocaleString with spec fallback and i…

Veercodeprog:
This PR brings `Temporal.Duration.prototype.toLocaleString()` and `Temporal.Duration.prototype.valueOf()` closer to the Temporal spec.

Since Boa does not currently implement `Intl.DurationFormat`, `Duration.prototype.toLocaleString()` should follow the spec-defined fallback behavior. In practice, that means returning the same serialized output as `toJSON()`.

This change also keeps `Duration.prototype.valueOf()` as a throwing method, while updating the error message to point users to the APIs that are actually appropriate for comparison and string conversion.

## Motivation

`toLocaleString()` previously produced output equivalent to the default duration serialization, but the implementation did not clearly reflect the spec's fallback path when ECMA-402 duration formatting is unavailable.

`valueOf()` already threw as expected, but the error message was not aligned with the spec note and did not direct users toward the best alternatives.

### `Temporal.Duration.prototype.toLocaleString()`

- Make `toLocaleString()` delegate to `toJSON()`.
- Preserve the expected API shape with `locales` and `options`, even though those arguments are currently ignored.
- Update the Rustdoc to document the current behavior and clarify that `Intl.DurationFormat` is not implemented yet.

### `Temporal.Duration.prototype.valueOf()`

- Keep the method behavior unchanged in that it always throws `TypeError`.
- Update the error message to reference:
  - `Temporal.Duration.compare()`
  - `Temporal.Duration.prototype.toString()`

## Tests

Added test coverage for the following cases in `core/engine/src/builtins/temporal/duration/tests.rs`:

- `toLocaleString()` returns the same result as `toJSON()`.
- `toLocaleString(locales, options)` still returns the same result while Intl-backed formatting is not implemented.
- `valueOf()` throws `TypeError` with the updated error message.

## Testing

```bash
cargo test -p boa_engine duration_to_locale_string_matches
cargo test -p boa_engine duration_value_of_throws

Note:
This PR does not implement localized duration formatting.

Proper locale-aware output will require Intl.DurationFormat support, along with the necessary locale data plumbing. This change only fixes the spec-defined fallback behavior used when that functionality is not available yet.

### DIFF
--- a/core/engine/src/builtins/temporal/duration/mod.rs
+++ b/core/engine/src/builtins/temporal/duration/mod.rs
@@ -1115,7 +1115,15 @@ impl Duration {
         Ok(JsString::from(result).into())
     }
 
-    /// 7.3.24 `Temporal.Duration.prototype.toLocaleString ( )`
+    /// 7.3.24 `Temporal.Duration.prototype.toLocaleString ( [ locales [ , options ] ] )`
+    ///
+    /// When the implementation includes ECMA-402, this method is defined by the Intl specification
+    /// (typically via `Intl.DurationFormat`). Boa does not implement `Intl.DurationFormat` yet, so
+    /// we use the Temporal proposal **non-ECMA-402** fallback: `TemporalDurationToString(duration, auto)`
+    /// — the same string as [`Self::to_json`].
+    ///
+    /// The `locales` and `options` arguments are accepted for API compatibility but are ignored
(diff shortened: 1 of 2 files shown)


jedel1043:
Closing per @nekevss comment

PR #5262, opened 2026-03-26
fix(built-ins): modify `IsTypedArrayFixedLength` implementation to match spec

### LINKED ISSUE #5256: Fix: permit `preventExtensions` on fixed length `TypedArray`s backed by GSABs
**Describe the bug**
Passing a fixed length `TypedArray` that is backed by a `SharedArrayBuffer` into `Object.preventExtensions` causes the method to throw a `TypeError`. This goes against the ecma262 spec which permits this operation under [`sec-typedarray-preventextensions`](https://tc39.es/ecma262/#sec-typedarray-preventextensions).

I'm fairly certain that this is why test262's [`test/staging/built-ins/Object/preventExtensions`](https://github.com/tc39/test262/blob/main/test/staging/built-ins/Object/preventExtensions/preventExtensions-variable-length-typed-arrays.js) is failing. 

**To Reproduce**
```javascript
const rab = new SharedArrayBuffer(4, { maxByteLength: 8 });
const ta = new Uint8Array(rab, 0, 4);

try {
    Object.preventExtensions(ta);
    console.log("IsTypedArrayFixedLength() == true");
} catch (e) {
    console.log("IsTypedArrayFixedLength() == false | " + e.name);
}
```

**Expected behavior**
The output should be `IsTypedArrayFixedLength() == true` but it throws and outputs `IsTypedArrayFixedLength() == false | TypeError`. The relevant spec sections are [`sec-typedarray-preventextensions`](https://tc39.es/ecma262/#sec-typedarray-preventextensions) and [`sec-istypedarrayfixedlength`](https://tc39.es/ecma262/#sec-istypedarrayfixedlength)

**Build environment**
- OS: Windows 11
- Version: 10.0.22631 Build 22631
- Target triple: [e.g. x86_64-unknown-linux-msvc]

**Additional context**
The current implementation of [`typed_array_exotic_prevent_extensions`](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/typed_array/object.rs#L289) incorrectly attempts to implement `IsTypedArrayFixedLength` with the following line:
```rust
ta.viewed_array_buffer().as_buffer().is_fixed_len()
```
The `is_fixed_len` method has no means to check whether the `TypedArray` is length is `AUTO`. It also dispatches to the corresponding `is_fixed_len` method on `ArrayBuffer` or `SharedArrayBuffer` which logically makes sense but the spec only cares for the case of fixed-length `ArrayBuffer`s (and for this, `IsTypedArrayFixedLength` is `false`). The fix should merely involve adding this two checks.

AlvinThorn008:
This Pull Request fixes/closes #5256.

As required by ecma262, `Object.preventExtensions` may be applied to typed arrays without error if the `IsTypedArrayFixedLength` is not false. Prior to this commit, this check was incorrectly implemented via `ArrayBuffer::is_fixed_len` which does not/can not include an AUTO length check. This discrepancy lies in that the requirements of a fixed-length `TypedArray` slightly differ from a fixed-length arraybuffer.

It adds and replaces the checks as specified by [`sec-istypedarrayfixedlength`](https://tc39.es/ecma262/#sec-istypedarrayfixedlength). Now, [typed_array_exotic_prevent_extensions](https://github.com/boa-dev/boa/blob/main/core/engine/src/builtins/typed_array/object.rs#L289) should
- return `Ok(false)` when the `TypedArray` is auto length or when the backing `ArrayBuffer` is resizable.
- failing the above, call `ordinary_prevent_extensions` if the backing array is a `SharedArrayBuffer`.

This PR also adds a test to ensure what the title of this issue says. In addition, test262's `test/staging/built-ins/preventExtensions` should now pass completely.

### DIFF
--- a/core/engine/src/builtins/typed_array/object.rs
+++ b/core/engine/src/builtins/typed_array/object.rs
@@ -4,7 +4,7 @@ use std::sync::atomic::Ordering;
 
 use crate::{
     Context, JsExpect, JsNativeError, JsResult, JsString, JsValue,
-    builtins::array_buffer::BufferObject,
+    builtins::array_buffer::{BufferObject, BufferRef},
     object::{
         JsData, JsObject,
         internal_methods::{
@@ -295,7 +295,14 @@ pub(crate) fn typed_array_exotic_prevent_extensions(
             .downcast_ref::<TypedArray>()
             .js_expect("must be a TypedArray")?;
 
-        ta.viewed_array_buffer().as_buffer().is_fixed_len()
+        if ta.is_auto_length() {
+            return Ok(false);
+        }
+
+        match ta.viewed_array_buffer().as_buffer() {
+            BufferRef::Buffer(buf) => !buf.is_fixed_len(),
+            _ => true
+        }
     };
 
     // 1. If IsTypedArrayFixedLength(O) is false, return false.
--- a/core/engine/src/builtins/typed_array/object.rs
+++ b/core/engine/src/builtins/typed_array/object.rs
@@ -312,8 +312,8 @@ pub(crate) fn typed_array_exotic_prevent_extensions(
 ) -> JsResult<bool> {
     let is_fixed_length = {
         let ta = obj
-        .downcast_ref::<TypedArray>()
-        .js_expect("must be a TypedArray")?;
+            .downcast_ref::<TypedArray>()
+            .js_expect("must be a TypedArray")?;
 
         ta.is_fixed_length()
     }; // Note: this block ensures that the borrow of obj is dropped
--- a/core/engine/src/builtins/typed_array/object.rs
+++ b/core/engine/src/builtins/typed_array/object.rs
@@ -300,7 +300,7 @@ pub(crate) fn typed_array_exotic_prevent_extensions(
         }
 
         match ta.viewed_array_buffer().as_buffer() {
-            BufferRef::Buffer(buf) => !buf.is_fixed_len(),
+            BufferRef::Buffer(buf) => buf.is_fixed_len(),
             BufferRef::SharedBuffer(_) => true,
         }
     };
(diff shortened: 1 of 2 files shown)


zhuzhu81998 (changes requested):
I think it would be much nicer if you implement the `IsTypedArrayFixedLength` operation as a function, similar to:
https://github.com/boa-dev/boa/blob/2437b6704c4315f6ec8f0766bf229fba23e0e61e/core/engine/src/builtins/typed_array/builtin.rs#L3269-L3293

Note that you should put the spec as comment.

AlvinThorn008:
@zhuzhu81998 
That is true. I debated whether it was necessary but it's certainly more in line with the codebase's style. I'll refactor it in a couple minutes. 

Thanks

AlvinThorn008:
I've copied the format of the function you linked. Everything seems to be working as before.

core/engine/src/builtins/typed_array/object.rs
```diff
@@ -290,23 +290,41 @@ pub(crate) fn typed_array_exotic_prevent_extensions(
...
     ordinary_prevent_extensions(obj, context)
 }
 
+/// Abstract operation `IsTypedArrayFixedLength ( O )`.
+///
+/// More information:
+///  - [ECMAScript reference][spec]
+///
+/// [spec]: https://tc39.es/ecma262/#sec-istypedarrayfixedlength
+fn is_typed_array_fixed_length(obj: &JsObject) -> JsResult<bool> {
```
jedel1043: I think it's better to put this in `TypedArray`, such that we can simplify the name to `is_fixed_length` just like `ArrayBuffer.is_fixed_len()`

AlvinThorn008: Makes sense. I'll do that shortly

AlvinThorn008: I ran into a small borrowing error but it resolved now.

jedel1043 (approved):
Looks great!

AlvinThorn008:
> Looks great!

Wonderful! Is there anything else that need to be done to merge this PR?

zhuzhu81998:
> > Looks great!
> 
> Wonderful! Is there anything else that need to be done to merge this PR?

maybe change the PR title quickly? it is more about implementing `IsTypedArrayFixedLength` correctly.

PR #4787, opened 2026-03-01
refactor: use `usize` for `Array`, `ArrayBuffer`, `TypedArray`, `DataView`

zhuzhu81998:
Closes #4766 .

Many lines of changes are just deleting `as usize` really.

Issue #4766, opened 2026-03-01
refactor `DataView`, `TypedArray`, `ArrayBuffer` indices to use `usize`


zhuzhu81998:
The usage of `u64` vs `usize` is not unified for indices of the types backed by `ArrayBuffer`.
We want to use `usize` wherever possible.

For example:
https://github.com/boa-dev/boa/blob/27959b2cebc00bbd917e8a8f3782b0b6b27a3bfb/core/engine/src/object/builtins/jsdataview.rs#L180-L186

https://github.com/boa-dev/boa/blob/27959b2cebc00bbd917e8a8f3782b0b6b27a3bfb/core/engine/src/builtins/array_buffer/mod.rs#L200-L213

zhuzhu81998:
closing due to reasons mentioned in #4787 .

PR #5263, opened 2026-03-26
refactor: `TypedArray` uses `usize` instead of `u64` for internal properties

zhuzhu81998:
Part of #4766 .
2nd atttempt after #4787 .
Compared to the previous attempt, this PR limits the scope of change to `TypedArray` only (as opposed to touching `Array` as well like in #4787 ).

We also keep the spec-defined checks (for which `u64` is used).

`TypedArray` test262 tests pass on 32-bit (well 4 failed ones actually but main also fails 4 so).

If this approach is perceived as good/reasonable, i will do this for `DataView` too.

### DIFF
--- a/core/engine/src/builtins/atomics/mod.rs
+++ b/core/engine/src/builtins/atomics/mod.rs
@@ -686,6 +686,9 @@ fn validate_atomic_access(
 
     // 2. Let accessIndex be ? ToIndex(requestIndex).
     let access_index = request_index.to_index(context)?;
+    let access_index: usize = access_index
+        .try_into()
+        .map_err(|_| JsNativeError::range().with_message("index too large"))?;
 
     // 3. Assert: accessIndex ≥ 0.
     //    ensured by the type.
@@ -698,7 +701,7 @@ fn validate_atomic_access(
     }
 
     // 8. Return (accessIndex × elementSize) + offset.
-    let offset = ((access_index * kind.element_size()) + offset) as usize;
+    let offset = (access_index * kind.element_size()) + offset;
     Ok(AtomicAccess {
         byte_offset: offset,
         kind,
--- a/core/engine/src/builtins/array/array_iterator.rs
+++ b/core/engine/src/builtins/array/array_iterator.rs
@@ -119,7 +119,7 @@ impl ArrayIterator {
                     .into());
             };
 
-            f.array_length(buf.len())
+            f.array_length(buf.len()) as u64
         } else {
             array_iterator.array.length_of_array_like(context)?
         };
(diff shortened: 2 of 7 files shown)


core/engine/src/builtins/atomics/mod.rs
```diff
@@ -686,6 +686,9 @@ fn validate_atomic_access(
 
     // 2. Let accessIndex be ? ToIndex(requestIndex).
     let access_index = request_index.to_index(context)?;
+    let access_index: usize = access_index
```
nekevss, line 689: issue: access_index per spec can be an integer anywhere from 0..=2^53 -1.

Casting to a usize in this instance would make us non-compliant on 32 bit systems

core/engine/src/builtins/atomics/mod.rs
```diff
@@ -698,7 +701,7 @@ fn validate_atomic_access(
     }
 
     // 8. Return (accessIndex × elementSize) + offset.
-    let offset = ((access_index * kind.element_size()) + offset) as usize;
+    let offset = (access_index * kind.element_size()) + offset;
```
nekevss, line 704: thought: we should validate and test whether we may overflow when at a large index and element size.

nekevss (changes requested):
I'm not sure that we can actually make this change unfortunately do to spec compliance. If I'm reading correctly, we need to be in u64 land.

core/engine/src/builtins/atomics/mod.rs
zhuzhu81998: well not really. because if you look at spec point 4: "4. If accessIndex ≥ length, throw a RangeError exception."

There is no way that the length of an allocated `TypedArray` can be larger than `usize`, thus rejecting `accessIndex` that is larger than `usize` is quite spec-compliant.

core/engine/src/builtins/atomics/mod.rs
zhuzhu81998: hmm i guess we could, if it overflows then it really just means that `access_index` is out of range.

zhuzhu81998: maybe in a followup PR? right now on main, it is also unsafe cast to usize.

core/engine/src/builtins/atomics/mod.rs
nekevss: Is there an upper limit to the length of a `TypedArray` according to the spec?

If the `access_index` per specification can be up to 2^53 - 1 and we don't support up to that range as a length, then wouldn't we be non spec compliant for TypedArray on 32-bit since the specification by default is essentially setting the upper bound of the TypedArray length at 2^53 - 1.

I'd be curious if there was a bound set somewhere in the specification besides this.

jedel1043: > Is there an upper limit to the length of a TypedArray according to the spec?

This is inherited from Shared/ArrayBuffer, but the limit is essentially "whichever number makes [AllocateArrayBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-allocatearraybuffer) fail"

jedel1043: so in practice, it would be much less than `usize`

core/engine/src/builtins/atomics/mod.rs
jedel1043: Agreed that this can be a follow up because an unsafe cast to usize is not that reassuring either.

PR #5269, opened 2026-03-27
fix abortsignal event names

### LINKED ISSUE #5260: AbortSignal.addEventListener() throws on non-"abort" event names
`AbortSignal` is an `EventTarget`, so `addEventListener()` should not throw just because the event name is something other than `"abort"`.

Right now Boa throws a `TypeError` here:

```js
const ctrl = new AbortController();
ctrl.signal.addEventListener("whatever", () => {});
```

Current result:

```js
TypeError: AbortSignal only supports the 'abort' event type
```

Expected behavior:

- no throw
- the listener is just never fired, since `AbortSignal` only dispatches `abort`

This looks like a small conformance bug in the current `AbortSignal` implementation. `removeEventListener()` already quietly ignores unknown event names, so `addEventListener()` throwing here is also a bit inconsistent with the rest of the API.

Relevant implementation:
- `core/runtime/src/abort/mod.rs`

Monti-27:
Fixes #5260.

This makes AbortSignal.addEventListener ignore non abort event names instead of throwing.
It keeps the abort listener path the same and adds tests for ignored event names.

Tests:
- cargo test -p boa_runtime abort --lib
- cargo test -p boa_runtime --lib

### DIFF
--- a/core/runtime/src/abort/mod.rs
+++ b/core/runtime/src/abort/mod.rs
@@ -159,7 +159,7 @@ impl JsAbortSignal {
         context: &mut Context,
     ) -> JsResult<()> {
         if event_type.to_std_string_escaped() != "abort" {
-            return Err(js_error!(TypeError: "AbortSignal only supports the 'abort' event type"));
+            return Ok(());
         }
         if self.aborted.get() {
             callback.call(&JsValue::undefined(), &[], context)?;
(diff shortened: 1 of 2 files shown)


core/runtime/src/abort/mod.rs
```diff
@@ -159,7 +159,7 @@ impl JsAbortSignal {
         context: &mut Context,
     ) -> JsResult<()> {
         if event_type.to_std_string_escaped() != "abort" {
-            return Err(js_error!(TypeError: "AbortSignal only supports the 'abort' event type"));
+            return Ok(());
```
jedel1043: This is also kinda not conformant though. According to the spec, we should be appending all event listeners irrespective of their type to the listeners, just making sure to not push two event listeners with the same type and callback

ref: https://dom.spec.whatwg.org/#dom-eventtarget-addeventlistener

jedel1043 (approved):
LGTM! Nice work!

PR #5271, opened 2026-03-28
Update modules code to the latest spec

jedel1043:
Just a small cleanup to update some paths that changed from the spec.

### DIFF
--- a/core/engine/src/module/source.rs
+++ b/core/engine/src/module/source.rs
@@ -47,15 +47,6 @@ use super::{
     SourceText,
 };
 
-/// Information for the [**Depth-first search**] algorithm used in the
-/// [`Module::link`] and [`Module::evaluate`] methods.
-#[derive(Clone, Copy, Debug, Trace, Finalize)]
-#[boa_gc(empty_trace)]
-pub(super) struct DfsInfo {
-    dfs_index: usize,
-    dfs_ancestor_index: usize,
-}
-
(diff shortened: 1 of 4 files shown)


core/engine/src/module/source.rs
```diff
@@ -2105,45 +2136,52 @@ fn async_module_execution_fulfilled(module: &Module, context: &mut Context) -> J
...
+        }
+        // c. Else,
+        // c.i. Let result be m.ExecuteModule().
+        let result = m_src.execute(&m, None, context);
+
+        // c.ii. If result is an abrupt completion, then
+        if let Err(e) = result {
+            // eagerly convert to an opaque error so that we can exit immediately
+            // if there was an EngineError.
+            let e = &JsError::from_opaque(e.into_opaque(context)?);
```
HalidOdat, line 2149: Suggestion: Maybe, we should add a helper method on `JsError` struct `fn to_non_engine_error() -> JsResult<Self, Self>`, we don't have to do the full round trip.

HalidOdat (approved):
Looks great! 🙌

core/engine/src/module/source.rs
jedel1043: Agreed. Though, it doesn't matter too much in this case since the callee converts to an opaque error afterwards, but it might be useful for other cases, so I would leave it as a future improvement

PR #5278, opened 2026-03-28
fix module loader cache

### LINKED ISSUE #5270: SimpleModuleLoader cache ignores import attributes
`SimpleModuleLoader` currently caches modules by path only, even though `ModuleRequest` already treats import attributes as part of the request.

That means the same module path can be reused across requests with different attributes.

A concrete case is JSON modules:

1. import `data.json` with `with { type: 'json' }`
2. import the same `data.json` again without attributes

The second import should reject, but it currently succeeds because the first request already put the module in cache and the later lookup ignores attributes.

This looks to come from `SimpleModuleLoader::get_with_attributes()` and `insert_with_attributes()` both ignoring the attributes, and the cache lookup happening before the loader checks whether the current request has the right import attributes.

Relevant files:
- `core/engine/src/module/loader/mod.rs`
- `core/engine/src/module/mod.rs`

Monti-27:
Fixes #5270.

This makes SimpleModuleLoader cache modules with their import attributes instead of only by path.
It also adds a regression test for the json case where the same path is imported once with attributes and later without them.

Tests:
- cargo test -p boa_engine --test imports -- --nocapture
- cargo test -p boa_engine --test module -- --nocapture
- cargo clippy -p boa_engine --all-features --all-targets -- -D warnings


jedel1043 (approved):
Looks great!

PR #5280, opened 2026-03-28
feat(clone): add structuredClone support for Error objects and relate…

### LINKED ISSUE #5279: structuredClone and postMessage reject Error objects
Both `structuredClone` and `postMessage` currently reject `Error` objects.

I was able to reproduce this on current `main`.

`structuredClone(new Error('boom'))` throws `TypeError: Errors are not supported yet.`

`postMessage(new Error('boom'))` fails for the same reason, since both paths go through `JsValueStore`.

This looks like a real gap in the structured clone path, not just missing API wiring. `ValueStoreInner` already has an `Error` variant, but the store still rejects `Error` objects on the way in and does not reconstruct them on the way out.

Relevant files:

1. `core/runtime/src/store/from.rs`
2. `core/runtime/src/store/to.rs`
3. `core/runtime/src/store/mod.rs`

Expected behavior is for `Error` objects to be cloned through the same path instead of failing at serialization.

akshatnathani:
This Pull Request fixes/closes #5279.

It changes the following:
- Added Error object support in the JsValueStore serialization path so structuredClone(new Error("boom")) and postMessage(new Error("boom")) no longer fail during cloning.
- Implemented Error object reconstruction in the deserialization path using stored error metadata (kind, name, message, and related fields), instead of returning Not yet implemented.
- Added regression tests for both clone paths:
  - clone::tests::clones_error_objects
  - message::tests::basic_error_message
- Removed the stale #[expect(unused)] on the ValueStoreInner::Error variant, since it is now actively used.
- Verified consistency with project standards and checks:
  - cargo fmt --all -- --check
  - cargo clippy -p boa_runtime --tests -- -D warnings
  - cargo test -p boa_runtime clones_error_objects -- --nocapture
  - cargo test -p boa_runtime basic_error_message -- --nocapture

### DIFF
--- a/core/runtime/src/store/from.rs
+++ b/core/runtime/src/store/from.rs
@@ -186,9 +186,13 @@ fn clone_regexp(
 
 fn clone_error(
     original: &JsObject,
+    transfer: &HashSet<JsObject>,
     seen: &mut SeenMap,
     context: &mut Context,
 ) -> JsResult<JsValueStore> {
+    let mut store = JsValueStore::empty();
+    seen.insert(original, store.clone());
+
     let native = JsError::from_opaque(JsValue::from(original.clone()))
         .try_native(context)
         .map_err(|_| unsupported_type())?;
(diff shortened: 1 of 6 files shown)


core/runtime/src/store/from.rs
```diff
@@ -181,6 +184,50 @@ fn clone_regexp(
...
+        if value.is_undefined() {
+            Ok(js_string!())
+        } else {
+            value.to_string(context)
+        }
+    };
+
+    let name = to_optional_string("name", context)?;
+    let stack = to_optional_string("stack", context)?;
+    let cause = to_optional_string("cause", context)?;
```
Monti-27: cause is getting flattened into a string here before it goes into the store. that works for simple text causes, but it breaks real cases like new Error("boom", { cause: { code: 7 } }), where the cloned error should still keep the original object as its cause.

core/runtime/src/store/to.rs
```diff
@@ -150,6 +161,47 @@ fn try_into_js_date(
...
+fn try_into_js_error(
+    store: &JsValueStore,
+    error_data: ErrorStoreFields<'_>,
+    seen: &mut ReverseSeenMap,
+    context: &mut Context,
+) -> JsResult<JsValue> {
+    let (kind, name, message, stack, cause) = error_data;
+    let message = message.to_js_string().to_std_string_escaped();
+    let native = match kind {
+        ErrorKind::Aggregate => JsNativeError::aggregate(Vec::new()),
```
Monti-27, line 174: this rebuilds AggregateError with an empty list, so the cloned value loses its errors property on the way back out. i checked this with a quick repro and structuredClone(new AggregateError([new Error("inner")], "agg")) comes back with errors.length === 0.

core/runtime/src/store/from.rs
akshatnathani: Great catch. I fixed the cause handling so that it is preserved as structured data rather than stringified. Added regression coverage for both structuredClone and postMessage paths, and also looking into the AggregateError issue you pointed out, and will follow up with the patch/tests shortly.

core/runtime/src/store/from.rs
```diff
@@ -181,6 +184,50 @@ fn clone_regexp(
...
 }
 
+fn clone_error(
+    original: &JsObject,
+    seen: &mut SeenMap,
+    context: &mut Context,
+) -> JsResult<JsValueStore> {
+    let native = JsError::from_opaque(JsValue::from(original.clone()))
+        .try_native(context)
+        .map_err(|_| unsupported_type())?;
```
jedel1043, line 194: This should be infallible, since we know `original` is already an `Error` type.
What we should do here is to add a conversion from `JsObject<Error>` to `JsNativeError` directly, and pass `JsObject<Error>` as our original object.

core/runtime/src/store/to.rs
```diff
@@ -1,17 +1,28 @@
...
 #[derive(Default)]
 pub(super) struct ReverseSeenMap(HashMap<usize, JsObject>);
 
+type ErrorStoreFields<'a> = (
+    &'a ErrorKind,
+    &'a StringStore,
+    &'a StringStore,
+    &'a StringStore,
+    &'a StringStore,
+);
```
jedel1043, line 24: This looks kinda unmaintainable; you cannot easily distinguish which field is the name, the message, the stack or the cause at first glance. A proper struct would be better to make it harder to pass the incorrect field.

jedel1043 (changes requested):
Looks good! I just have a couple of suggestions to improve the code.

PR #5283, opened 2026-03-29
Perf/array push pop dense fast path

### LINKED ISSUE #5281: Array.prototype Operations Fast-Path Optimization
# Array.prototype Operations Fast-Path Optimization

This plan outlines the implementation of fast-path optimizations for `Array.prototype.push` and `Array.prototype.pop` when dealing with arrays that have contiguous, "dense" index storage. 

As a JS engine, it's fundamental to optimize array operations. We already perform a similar optimization for `Array.prototype.shift`, but the `push` and `pop` functions—arguably the two most commonly used methods—currently go through the fully generic (and much slower) `#get()` and `#set()` paths when computing property existence and values.

## Proposed Changes

We will introduce fast paths that inspect the internal `IndexedProperties` variant of an array. If the array is dense and continuous (which is the case 99% of the time for simple scripts), we can bypass the full object method property table lookups entirely.

#### [MODIFY] [property_map.rs](file:///home/mayank/boa/core/engine/src/object/property_map.rs)

- Add a `pop_dense` method to `IndexedProperties` that returns a `JsValue`. It checks `vec.len()` and trivially uses `pop()` on `DenseI32`, `DenseF64`, and `DenseElement`, returning the last item safely, and updates internal states.
- Ensure that `push_dense` can correctly handle multi-argument pushing without needing an additional lookup per arg if they succeed.

#### [MODIFY] [mod.rs](file:///home/mayank/boa/core/engine/src/builtins/array/mod.rs)

- Modify `Array.prototype.pop`: Add a fast path identical to the existing `shift` one. When `len > 0` and the array's dense property array matches the length `len`, borrow the properties mutably, pop the item natively via `IndexedProperties::pop_dense`, update the length uniformly, and return the item wrapped as a `JsValue`.
- Modify `Array.prototype.push`: Iterate over the arguments and use the existing `push_dense` on `IndexedProperties`. If all arg pushes fall in the dense track successfully, update length directly, returning early to save hundreds of cycles per push.

### Automated Tests
- Run `cargo test array` to verify all `push` and `pop` behaviors remain unchanged for both dense and sparse arrays.
- Run the Test262 test suite via `make test262` ensuring we haven't broken any ECMAScript specifications or edge cases (e.g. frozen arrays, multi-argument pushes spanning length limits).

### Manual Verification
- Execute typical microbenchmarks covering rapid insertions/pops to verify the expected performance gains, making sure it executes successfully and without regression.

MayankRaj435:
## Summary
This PR implements fast-path optimizations for `Array.prototype.push` and `Array.prototype.pop` when dealing with arrays that have contiguous, "dense" index storage. By bypassing the generic property map lookups and standard `[[Get]]`/`[[Set]]` internal methods, we significantly reduce the overhead for these fundamental operations.

## Related Issue
Fixes #5281

## Key Changes
- **`IndexedProperties` (property_map.rs)**:
    - Added `pop_dense()` method to natively remove elements from dense vectors (`DenseI32`, `DenseF64`, and `DenseElement`).
    - Added `len()` and `is_sparse()` helper methods to facilitate fast check logic.
- **`Array.prototype.pop` (array/mod.rs)**:
    - Introduced a fast path that checks if the object is a standard array with dense storage.
    - Directly removals the last element and updates the `length` property, avoiding string conversion and map lookups.
- **`Array.prototype.push` (array/mod.rs)**:
    - Introduced a fast path that iterates over arguments and used `push_dense`.
    - Added an `extensible` check to ensure strict compliance when adding new properties to non-extensible arrays.
    - Updates the `length` property once at the end of the operation.

## Verification
- **Unit Tests**: Passed `cargo test array` in `core/engine` (123 tests).
- **CI**: local `cargo clippy` and `cargo fmt` passed (as seen in the command logs).

## Performance Impact
Initial microbenchmarks suggest a significant reduction in cycles for `push` and `pop` on dense arrays, as we now avoid:
1. Numeric index stringification.
2. Property descriptor map lookups.
3. Multiple internal method calls per argument.

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -875,6 +875,28 @@ impl Array {
                 )
                 .into());
         }
+
+        // Small optimization for arrays using dense properties
+        if o.is_array() {
+            let mut o_borrow = o.borrow_mut();
+            let extensible = o_borrow.extensible;
+            let props = o_borrow.properties_mut();
+            if !props.indexed_properties.is_sparse()
+                && (len as usize) == props.indexed_properties.len()
+                && (extensible || args.is_empty())
+            {
+                for element in args {
+                    if !props.indexed_properties.push_dense(element) {
+                        break;
+                    }
+                    len += 1;
+                }
+                drop(o_borrow);
+                Self::set_length(&o, len, context)?;
+                return Ok(len.into());
+            }
+        }
+
         // 5. For each element E of items, do
         for element in args.iter().cloned() {
             // a. Perform ? Set(O, ! ToString(𝔽(len)), E, true).
@@ -913,6 +935,16 @@ impl Array {
             Ok(JsValue::undefined())
         // 4. Else,
         } else {
+            // Small optimization for arrays using dense properties
+            if o.is_array() {
+                let mut o_borrow = o.borrow_mut();
+                if let Some(res) = o_borrow.properties_mut().indexed_properties.pop_dense() {
+                    drop(o_borrow);
+                    Self::set_length(&o, len - 1, context)?;
+                    return Ok(res);
+                }
+            }
+
             // a. Assert: len > 0.
             // b. Let newLen be 𝔽(len - 1).
             let new_len = len - 1;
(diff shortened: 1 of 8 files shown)


MayankRaj435:
Hey maintainers ,please review it and let me know if any changes required

PR #5284, opened 2026-03-30
Add completed `PropertyDescriptor` type

HalidOdat:
Refactors property descriptor handling by replacing the builder-style PropertyDescriptor with a more explicit `CompletePropertyDescriptor` enum. This makes things clearer and more type-safe. The spec always stores complete property descriptors in objects.

Since descriptors are always complete, this allows us to skip some checks and may allow the compiler to better optimize the code. Additionally this also reduces the memory usage for sparse generic arrays, `CompletePropertyDescriptor` is 24 bytes vs. `PropertyDescriptor` 40 bytes (64bit arch).

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -26,7 +26,9 @@ use crate::{
             ordinary_get_own_property,
         },
     },
-    property::{Attribute, PropertyDescriptor, PropertyKey, PropertyNameKind},
+    property::{
+        Attribute, CompletePropertyDescriptor, PropertyDescriptor, PropertyKey, PropertyNameKind,
+    },
     realm::Realm,
     string::StaticJsStrings,
     symbol::JsSymbol,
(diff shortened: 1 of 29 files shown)


core/engine/src/builtins/array/mod.rs
```diff
@@ -3493,21 +3496,28 @@ fn array_exotic_define_own_property(
...
             // b. Assert: ! IsDataDescriptor(oldLenDesc) is true.
-            debug_assert!(old_len_desc.is_data_descriptor());
+            let CompletePropertyDescriptor::Data {
+                value,
+                writable,
+                enumerable,
+                configurable,
+            } = old_len_desc
+            else {
+                panic!("length is always a data property descriptor");
```
jedel1043: We're kinda trying to migrate to `PanicError` instead of true panics to remove all panics from the engine, so you should return that instead.

jedel1043 (changes requested):
Mostly looks great! I just have a suggestion to avoid introducing more panics in the engine.

jedel1043:
You might need to bump to v0.23 since removing `fill_with` is technically a breaking change, but it might just be possible to preserve the original public API by restricting the usage of `fill_with` to only incomplete descriptors, and only allow `PropertyDescriptor::fill_with`

PR #5287, opened 2026-03-30
Add unit tests for bytecode argument reading (`args.rs`)

### LINKED ISSUE #4240: Add unit tests for bytecode arguments reading

Cypher-CP0:
## Summary

Closes #4240

Adds unit tests for the bytecode argument reading logic in `core/engine/src/vm/opcode/args.rs`, as requested in the linked issue.

## Changes

Added the following test coverage inside the existing `#[cfg(test)] mod tests` block in `args.rs`:

### Direct `read` function tests
- `read_u8_direct` — basic u8 read
- `read_u16_little_endian` — verifies correct little-endian byte ordering for u16
- `read_u32_little_endian` — verifies correct little-endian byte ordering for u32
- `read_at_non_zero_offset` — verifies reading from a non-zero offset advances `pos` correctly

### Signed negative value tests
- `read_i8_negative` — reads -1 as i8
- `read_i16_negative` — reads -1 as i16 in little-endian
- `read_i32_negative` — reads `i32::MIN` from raw bytes

### Operand edge cases
- `register_operand_large_value` — `RegisterOperand` with `u32::MAX`
- `index_operand_mid_value` — `IndexOperand` with a mid-range value
- `address_max_value` — `Address` with `u32::MAX`
- `decode_at_non_zero_offset` — decodes the second of two back-to-back encoded values

### Primitive edge cases
- `test_i8_max`, `test_i16_min_max`, `test_i32_max`, `test_u8_mid_value`

### Float special values
- `test_f32_special_values` — NaN, +inf, -inf for f32
- `test_f64_special_values` — NaN, +inf, -inf for f64

### Additional tuple combinations
- `test_triple_tuple_round_trip` — 3-element tuples
- `test_five_tuple_round_trip` — 5-element tuples

### Additional panic/truncation tests
- `decode_u16_truncated_panics` — 1 byte buffer for u16
- `decode_u64_truncated_panics` — 4 byte buffer for u64
- `decode_at_offset_past_end_panics` — offset + size exceeds buffer length

### Additional ThinVec tests
- `test_thin_vec_single_element` — single element vec
- `test_thin_vec_index_operand` — ThinVec of IndexOperand with min/max values

### DIFF
--- a/core/engine/src/vm/opcode/args.rs
+++ b/core/engine/src/vm/opcode/args.rs
@@ -341,4 +341,203 @@ mod tests {
         let bytes = [0u8; 2];
         let _ = u32::decode(&bytes, 0);
     }
+    // --- Direct `read` function tests ---
+    // These test the low-level `read` fn directly, not just via encode/decode.
+
+    #[test]
+    fn read_u8_direct() {
+        use super::read;
+        let bytes: &[u8] = &[0x2A];
+        let (val, next) = read::<u8>(bytes, 0);
+        assert_eq!(val, 42u8);
+        assert_eq!(next, 1);
+    }
+
+    #[test]
+    fn read_u16_little_endian() {
+        // Verifies little-endian byte order: 0x0105 = 261
+        use super::read;
+        let bytes: &[u8] = &[0x05, 0x01];
+        let (val, next) = read::<u16>(bytes, 0);
+        assert_eq!(val, 261u16);
+        assert_eq!(next, 2);
+    }
+
+    #[test]
+    fn read_u32_little_endian() {
+        // 0x01000000 in little-endian is [0x00, 0x00, 0x00, 0x01]
+        use super::read;
+        let bytes: &[u8] = &[0x00, 0x00, 0x00, 0x01];
+        let (val, next) = read::<u32>(bytes, 0);
+        assert_eq!(val, 0x0100_0000u32);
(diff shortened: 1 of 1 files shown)


core/engine/src/vm/opcode/args.rs
```diff
@@ -341,4 +341,203 @@ mod tests {
...
+    }
+
+    #[test]
+    fn address_max_value() {
+        round_trip_eq(&Address::new(u32::MAX), |a, b| {
+            u32::from(*a) == u32::from(*b)
+        });
+    }
+
+    // --- More primitive edge cases ---
```
jedel1043, line 450: These should be part of the `test_primitive_round_trips` test. Having a single test for round tripping a single num sounds kind of suboptimal.

Also, many of the new tests are kinda already covered by already existing tests. Can you distill those into just the paths not covered by the current tests?

jedel1043 (changes requested):

jedel1043:
Closing since this PR got stale and the review wasn't addressed.

PR #5288, opened 2026-03-30
fix request signal

### LINKED ISSUE #5286: fetch(request) ignores Request.signal
When `fetch()` is called with a `Request` object, Boa currently seems to ignore the signal stored on that request.

A simple repro is:

```js
const ctrl = new AbortController();
const request = new Request("http://unit.test", { signal: ctrl.signal });
ctrl.abort("stop");
await fetch(request);
```

This should reject with the abort reason, but right now it fulfills.

The current path looks like it only reads `signal` from the second `fetch(..., options)` argument. When the resource is a `Request`, `fetch_inner()` clones the inner HTTP request, but `JsRequest` does not carry the signal through that path.

So the request level abort state gets lost before the fetcher sees it.

Monti-27:
Fixes #5286.

This makes fetch(request) use the signal stored on the Request object.
It also keeps fetch level signal overrides working the same way.

Tests:
- cargo test -p boa_runtime --lib

### DIFF
--- a/core/runtime/src/fetch/mod.rs
+++ b/core/runtime/src/fetch/mod.rs
@@ -122,6 +122,8 @@ async fn fetch_inner<T: Fetcher>(
 
     // The resource parsing is complicated, so we parse it in Rust here (instead of relying on
     // `TryFromJs` and friends).
+    let mut signal = signal;
+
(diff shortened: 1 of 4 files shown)


HalidOdat (approved):
Looks good to me! 😄

PR #5289, opened 2026-04-01
Modify boa_runtime's timers to use native timeout tracking and cancellation

jedel1043:
This modifies the timer code in `boa_runtime` such that it uses the recently introduced `CancellationToken`.
It also removes usages of `Rc<RefCell<>>`, which are replaces by directly querying the context for a timeout tracker, making it possible to extract a `&mut IntervalInnerState` from it`.

Finally a couple of cleanups and renames here and there


nekevss (approved):
LGTM

PR #5290, opened 2026-04-01
refactor(temporal): remove unnecessary clone in roundTo option parsing

tkshsbcue:
# PR: refactor(temporal): remove unnecessary clone in roundTo option parsing

## Summary

The `round` method in 5 Temporal builtins (`Duration`, `Instant`, `PlainDateTime`, `PlainTime`, `ZonedDateTime`) previously matched on `args.first().map(JsValue::variant)` to inspect the `roundTo` argument. In the fallback arm, the extracted `JsVariant` was converted back into a `JsValue` via `JsValue::from(round_to)` just to pass it to `get_options_object(&JsValue)` — an unnecessary round-trip that cloned the inner value.

This PR refactors all 5 call sites to work directly with `&JsValue` via `args.get_or_undefined(0)`, using `is_undefined()` and `as_string()` checks instead of variant pattern matching. This eliminates the clone while preserving identical semantics.

### Changes

- **`duration/mod.rs`** — Replaced `match` on `JsVariant` with `if`/`else` on `&JsValue`
- **`instant/mod.rs`** — Same refactor + removed unused `JsVariant` import
- **`plain_date_time/mod.rs`** — Same refactor + removed unused `JsVariant` import
- **`plain_time/mod.rs`** — Same refactor
- **`zoneddatetime/mod.rs`** — Same refactor

Resolves the `TODO: remove this clone` comments in the affected files.

## Test plan

- [x] `cargo fmt -- --check` passes
- [x] `cargo clippy --all-features --all-targets` passes
- [x] `cargo clippy --no-default-features` passes
- [x] `cargo test -p boa_engine -- temporal` passes

### DIFF
--- a/core/engine/src/builtins/temporal/duration/mod.rs
+++ b/core/engine/src/builtins/temporal/duration/mod.rs
@@ -896,33 +896,30 @@ impl Duration {
                 JsNativeError::typ().with_message("this value must be a Duration object.")
             })?;
 
-        let round_to = match args.first().map(JsValue::variant) {
-            // 3. If roundTo is undefined, then
-            None | Some(JsVariant::Undefined) => {
-                return Err(JsNativeError::typ()
-                    .with_message("roundTo cannot be undefined.")
-                    .into());
-            }
-            // 4. If Type(roundTo) is String, then
-            Some(JsVariant::String(rt)) => {
-                // a. Let paramString be roundTo.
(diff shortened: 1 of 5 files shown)


zhuzhu81998 (approved):
looks correct to me 👀

jedel1043 (approved):
Thanks!

PR #5300, opened 2026-04-05
perf(array): add dense fast path to Array.prototype.push

tkshsbcue:
### Description

`Array.prototype.push` goes through the generic `[[Set]]` path for every element — converting the index to a `PropertyKey`, cloning the receiver, calling `__get_own_property__`, then `__define_own_property__` with full descriptor validation. For dense arrays, none of this is necessary.

The `PushValueToArray` opcode (used for array literals like `[1, 2, 3]`) already has a fast path that appends directly to dense indexed storage via `push_dense()`. This PR applies the same pattern to the `Array.prototype.push` builtin.

### How it works

Before entering the generic `[[Set]]` loop, check if the object is an extensible array with dense storage. If so, push each element directly via `push_dense()` and update the length in-place. `push_dense()` handles type transitions (DenseI32 → DenseF64 → DenseElement) internally and only returns `false` for already-sparse storage, so the first push serves as a probe — if it succeeds, all subsequent pushes are guaranteed to succeed.

The fast path is skipped for:
- Non-array objects (array-like objects go through the slow path)
- Non-extensible arrays (`Object.preventExtensions`, `Object.freeze`, `Object.seal`)
- Sparse arrays (storage already transitioned to sparse)
- Arrays whose length doesn't fit in i32

The slow path is completely untouched.

### Benchmark results

Criterion microbenchmark (200 iterations × 5000 integer pushes):

```
Baseline (3 runs):          127.48 ms median
With fast path (3 runs):     64.73 ms median
```

This is a push-dominated benchmark. On a more realistic workload (pushing objects + filtering), the improvement is smaller since push is no longer the bottleneck:

```
Realistic (objects + filter, Date.now(), release build):
  Baseline:    139 ms median
  With patch:  110 ms median
```

Improvement ranges from ~20% (mixed workloads) to ~49% (push-dominated) depending on how much time is spent in push.

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -875,6 +875,44 @@ impl Array {
                 )
                 .into());
         }
+
+        // Fast path: if O is an extensible dense array whose shape still matches the
+        // default array template (guaranteeing length is writable and no custom
+        // property descriptors), push directly to indexed storage without going
+        // through the generic [[Set]] machinery. push_dense() handles type transitions
+        // (DenseI32 -> DenseF64 -> DenseElement) and only returns false for
+        // already-sparse storage, so if the first push succeeds all subsequent
+        // pushes on the same storage will too.
+        let template_shape = context
+            .intrinsics()
+            .templates()
+            .array()
+            .shape()
+            .to_addr_usize();
+        if o.is_array() && !args.is_empty() {
+            let mut o_borrowed = o.borrow_mut();
+            if o_borrowed.extensible
+                && o_borrowed.properties().shape.to_addr_usize() == template_shape
+                && let Some(len_i32) = o_borrowed.properties().storage[0].as_i32()
+                && o_borrowed
+                    .properties_mut()
+                    .indexed_properties
+                    .push_dense(&args[0])
+            {
+                let mut current_len = len_i32 + 1;
+                for element in &args[1..] {
+                    o_borrowed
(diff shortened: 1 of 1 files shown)


zhuzhu81998:
the failing tests hint on skipped checks (similar to #5076)

tkshsbcue:
You're right @zhuzhu81998 same class of bug as #5076. The shape check only verifies the array's own shape matches the default template, but doesn't guard against indexed property setters on the prototype chain. So when someone does `Object.defineProperty(Array.prototype, "0", { set: ... })`, the fast path fires anyway and bypasses `[[Set]]` entirely, skipping the setter invocation and any side effects (like freezing the array or making `length` non-writable).

I think the most practical fix is to extend the guard to also check that `Array.prototype` and `Object.prototype` still have their default template shapes if anyone has defined an indexed property on either, the shape will differ and we fall through to the slow path. Something like:

```rust
let array_proto_shape = context
    .intrinsics()
    .constructors()
    .array()
    .prototype();

// bail if Array.prototype or Object.prototype have been modified
// with indexed properties
```

The more robust long-term approach would be a realm-level "no indexed accessors on prototypes" flag (like V8's protector cells), but checking prototype shapes should be sufficient here and keeps the PR scoped.

I'll update the PR with the fix.

jedel1043:
Closing since the PR is not ready. Feel free to reopen if you want to resolve the test regresssions

PR #5302, opened 2026-04-05
fix Response.redirect URL handling

### LINKED ISSUE #5301: Response.redirect() accepts relative URLs and does not serialize Location
`Response.redirect()` currently has two spec mismatches:

```js
Response.redirect("https://example.com").headers.get("location")
// current: "https://example.com"
// expected: "https://example.com/"

try {
  Response.redirect("/foo");
  console.log("no throw");
} catch (e) {
  console.log(e.name);
}
// current: "no throw"
// expected: "TypeError"
```

The Fetch spec says `Response.redirect()` should parse the input URL and append the serialized parsed URL to the `Location` header.

So the current implementation is using the raw input string instead of the serialized parsed URL, and it also accepts relative redirect targets that should fail here.

Monti-27:
Fixes #5301.

`Response.redirect()` was validating the input with a generic URI check and then writing the original string into the `Location` header. That meant absolute URLs were not serialized, and relative URLs like `/foo` were accepted.

This changes `Response.redirect()` to parse the target as a URL and store the serialized parsed URL in `Location`. It also enables the `url` dependency under the `fetch` feature so the non-default build stays correct.

Tests cover:
- serializing parsed absolute URLs
- preserving fragments
- rejecting relative redirect targets
- rejecting values that coerce to non-URL strings

Checks:
- cargo test -p boa_runtime response -- --nocapture
- cargo test -p boa_runtime --no-default-features --features fetch response_redirect -- --nocapture

### DIFF
--- a/core/runtime/src/fetch/tests/response.rs
+++ b/core/runtime/src/fetch/tests/response.rs
@@ -180,10 +180,74 @@ fn response_redirect_custom_status_and_coercion() {
             r#"
                 const response = Response.redirect("http://example.com/", 301);
                 assertEq(response.status, 301);
+            "#,
+        ),
+    ]);
+}
 
-                // Tests Web IDL coercion of the URL parameter
-                const response2 = Response.redirect(12345);
-                assertEq(response2.headers.get("location"), "12345");
+#[test]
+fn response_redirect_rejects_non_url_after_string_coercion() {
+    run_test_actions([
+        TestAction::harness(),
+        TestAction::inspect_context(|ctx| register(&[], ctx)),
+        TestAction::run(
+            r#"
+                let threw = false;
+                try {
+                    Response.redirect(12345);
+                } catch (e) {
+                    threw = true;
+                    if (!(e instanceof TypeError)) {
+                        throw new Error("Expected TypeError, got " + e.name);
+                    }
+                }
+                if (!threw) {
+                    throw new Error("Expected TypeError, but no error was thrown");
+                }
+            "#,
+        ),
+    ]);
+}
+
+#[test]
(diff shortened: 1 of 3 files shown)


core/runtime/src/fetch/tests/response.rs
```diff
@@ -180,10 +180,74 @@ fn response_redirect_custom_status_and_coercion() {
...
+#[test]
+fn response_redirect_rejects_relative_url() {
+    run_test_actions([
+        TestAction::harness(),
+        TestAction::inspect_context(|ctx| register(&[], ctx)),
+        TestAction::run(
+            r#"
+                let threw = false;
+                try {
+                    Response.redirect("/foo");
```
zhuzhu81998, line 241: need adjusting (see comments in #5301. )
this should not throw.

Monti-27: yeah I'll push the commit by night was working on something else

Monti-27: I rechecked this more carefully. The spec parses against the current settings objects API base URL, so I think the /foo case depends on the host environment. that should explain why it works in your browser console ss while node and boa CLI both throw here

The part this PR is fixing is narrower: Location should use the serialized parsed url

@jedel1043 what do you think about keeping this PR scoped to that part ??

jedel1043: This just needs to call `Fetcher::resolve_uri`, which is the platform-dependant API for URI resolution. (unless the spec says otherwise, but since browsers and CLIs differ, I expect it to have something related to a host-dependant operation for URIs)

PR #5304, opened 2026-04-07
fix(runtime): replace unpaired surrogates with U+FFFD in UTF-16 TextD…

### LINKED ISSUE #4612: utf-16 TextDecoder is incorrect

tkshsbcue:
fix(runtime): replace unpaired surrogates with U+FFFD in UTF-16 TextDecoder

## Summary

Fixes #4612.

The UTF-16 `TextDecoder` (both LE and BE) passed raw code units directly to `JsString`, preserving unpaired surrogates (e.g. `\ud800`) instead of replacing them with U+FFFD as required by the [WHATWG Encoding Standard](https://encoding.spec.whatwg.org/).

## Changes

- Add a shared `decode_utf16_units()` helper using `std::char::decode_utf16` with `.unwrap_or('\u{FFFD}')` to produce well-formed output
- Route both `utf16le::decode` and `utf16be::decode` through this helper
- Simplify `utf16be::decode` to borrow (`&[u8]`) instead of taking ownership (`Vec<u8>`)
- Handle dangling byte edge case: only append extra U+FFFD if the last code unit is not already a replaced high surrogate

## Test plan

- [x] `decoder_utf16le_replaces_unpaired_surrogates` — lone high/low surrogates, consecutive, reversed pairs
- [x] `decoder_utf16be_replaces_unpaired_surrogates` — same for big-endian
- [x] `decoder_utf16le_dangling_byte_produces_replacement` — odd-length input
- [x] `decoder_utf16be_dangling_byte_produces_replacement` — odd-length input (BE)
- [x] All 33 existing text tests pass
- [x] `cargo fmt` and `cargo clippy` clean

### DIFF
--- a/core/runtime/src/text/tests.rs
+++ b/core/runtime/src/text/tests.rs
@@ -1,4 +1,3 @@
-use super::encodings;
 use crate::test::{TestAction, run_test_actions_with};
 use crate::text;
 use boa_engine::object::builtins::JsUint8Array;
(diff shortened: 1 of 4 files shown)


core/runtime/src/text/tests.rs
```diff
@@ -476,3 +477,68 @@ fn decoder_handle_data_view_offset_and_length() {
...
+    assert_eq!(result, expected);
+}
+
+#[test]
+fn decoder_utf16be_dangling_byte_produces_replacement() {
+    let input: &[u8] = &[0x00, 0x41, 0x42]; // 'A' (BE) + dangling byte
+    let result = encodings::utf16be::decode(input, false);
+    let expected = JsString::from(&[0x0041u16, 0xFFFD][..]);
+    assert_eq!(result, expected);
+}
```
jedel1043: Do we need these tests? Might be better to enable the utf16 decoding tests from the wpt suite instead, since those basically compile to the same kind of tests

jedel1043 (changes requested):

core/runtime/src/text/tests.rs
tkshsbcue: oh yes, I've pushed a follow-up commit that enables the upstream WPT suite instead. Specifically:                                                                      
                                                                                                                                                                             
 ->Enabled encoding/textdecoder-utf16-surrogates.any.js in tests/wpt/src/lib.rs, which covers both the replacement-mode cases and the fatal: true cases.
                     
->Since the WPT test also exercises fatal, I implemented the fatal option on TextDecoder: added it to TextDecoderOptions, stored it on the struct, exposed the fatal 
      
 ->getter, and threaded it through the decode helpers so decode() returns a TypeError on invalid input instead of inserting U+FFFD.                                           

->Dropped the custom INVALID_UTF16_CASES Rust tests since the WPT file now covers the same ground.

nekevss:
Closing due to stale PR. Feel free to reopen in the future if you come back to it.

PR #5305, opened 2026-04-07
feat(compiler): Generate single emit opcode for negative numbers

HalidOdat:
A small optimization that replaces the generated output for simple negative numbers, previously emitting `StoreInt8` followed by `Neg`, with a single `StoreInt8` instruction that directly emits the negative value.

### DIFF
--- a/core/engine/src/bytecompiler/expression/unary.rs
+++ b/core/engine/src/bytecompiler/expression/unary.rs
@@ -1,5 +1,6 @@
 use crate::bytecompiler::{Access, BindingAccessOpcode, ByteCompiler, Register, ToJsString};
 use boa_ast::Expression;
+use boa_ast::expression::literal::Number;
 use boa_ast::expression::operator::{Unary, unary::UnaryOp};
 
 impl ByteCompiler<'_> {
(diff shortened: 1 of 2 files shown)


core/engine/src/bytecompiler/expression/unary.rs
```diff
@@ -18,8 +19,19 @@ impl ByteCompiler<'_> {
...
-                self.compile_expr(unary.target(), dst);
-                self.bytecode.emit_neg(dst.variable());
+                if let Expression::Literal(literal) = unary.target().flatten()
+                    && let Some(number) = literal.kind().as_number()
+                {
+                    match number {
+                        // Handles special case -0
+                        Number::Int(0) => self.emit_store_rational(-0.0, dst),
+                        Number::Int(value) => self.emit_store_integer(-value, dst),
+                        Number::Num(value) => self.emit_store_rational(-value, dst),
```
zhuzhu81998, line 29: What about `BigInt`?

zhuzhu81998 (commented):
maybe 2 or 3 unit tests to check the -0 special case? (assuming they are not covered elsewhere already)

PR #5306, opened 2026-04-07
feat(compiler): Generate `Dec` opcode when subracting by one

HalidOdat:
This replaces the `rhs` generation for `1` for subtracting binary operations, with a `Dec` operation.


HalidOdat:
Nvm, this optimization won't work because `1n - 1` should throw instead of return `0n`

PR #5311, opened 2026-04-09
fix(object/ ): convert remaining panics to EngineError::Panic

KaustubhOG:
Part of #3241. Follow-up to #5063 addressing  review comment.

Changes:
- `operations.rs`: converted 5 panics — ToObject, CreateDataPropertyOrThrow (js_expect), private_field_add, private_method_or_accessor_add, private_set (try_borrow_mut)
- `internal_methods/mod.rs`: migrated 6 borrow call sites to try_borrow/try_borrow_mut
- `internal_methods/string.rs`: migrated 3 borrow call sites, return type of string_get_own_property changed to JsResult<Option<PropertyDescriptor>>
- `builtins/jsdataview.rs`: migrated 2 borrow call sites to try_borrow

### DIFF
--- a/core/engine/src/object/builtins/jsdataview.rs
+++ b/core/engine/src/object/builtins/jsdataview.rs
@@ -2,6 +2,7 @@
 use crate::{
     Context, JsExpect, JsNativeError, JsResult, JsValue,
     builtins::{DataView, array_buffer::BufferObject},
+    error::PanicError,
     object::{JsArrayBuffer, JsObject},
     value::TryFromJs,
 };
@@ -60,7 +61,9 @@ impl JsDataView {
         let offset = offset.unwrap_or_default();
 
         let (buf_byte_len, is_fixed_len) = {
-            let buffer = buffer.borrow();
+            let buffer = buffer
+                .try_borrow()
+                .map_err(|e| PanicError::new(e.to_string()))?;
             let buffer = buffer.data();
 
             // 4. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
@@ -110,7 +113,13 @@ impl JsDataView {
 
         // 11. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
         // 12. Set bufferByteLength to ArrayBufferByteLength(buffer, seq-cst).
-        let Some(buf_byte_len) = buffer.borrow().data().bytes().map(|s| s.len() as u64) else {
+        let Some(buf_byte_len) = buffer
+            .try_borrow()
+            .map_err(|e| PanicError::new(e.to_string()))?
+            .data()
+            .bytes()
+            .map(|s| s.len() as u64)
+        else {
             return Err(JsNativeError::typ()
                 .with_message("ArrayBuffer is detached")
                 .into());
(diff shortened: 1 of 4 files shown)


KaustubhOG:
@jedel1043 Following up on your review in #5063 regarding the `borrow()`/`borrow_mut()` guards in `jsobject.rs` (lines 820, 850).

Changing their return types from `Ref<'_, Object<T>>`/`RefMut<'_, Object<T>>` to `JsResult<...>` would cascade into hundreds of call sites across `object/`, `shape/`, and `internal_methods/`. Two possible approaches:

1. Change the foundation methods themselves and update all call sites
2. Leave the foundation methods as-is and only convert specific call sites inside `JsResult`-returning functions to use `try_borrow + js_expect` directly

does thiss idea looks fine ?

jedel1043:
> Changing their return types from Ref<'_, Object<T>>/RefMut<'_, Object<T>> to JsResult<...> would cascade into hundreds of call sites across object/, shape/, and internal_methods/. 

I was more thinking of a progressive migration to `try_borrow` and `try_borrow_mut`, leaving the panicking methods there. Users would still wanna panic sometimes, so having that API in place is still useful.

KaustubhOG:
> > Changing their return types from Ref<'_, Object>/RefMut<'_, Object> to JsResult<...> would cascade into hundreds of call sites across object/, shape/, and internal_methods/.
> 
> I was more thinking of a progressive migration to `try_borrow` and `try_borrow_mut`, leaving the panicking methods there. Users would still wanna panic sometimes, so having that API in place is still useful.

Ok. Will migrate call sites inside `JsResult`-returning functions across `object/` to use `try_borrow`/`try_borrow_mut` directly, leaving the panicking foundation methods in place. Will update this PR shortly.

KaustubhOG:
@jedel1043 Updated. Migrated all eligible `borrow`/`borrow_mut` call sites inside `JsResult`-returning functions across `object/` to use `try_borrow`/`try_borrow_mut` directly.


Remaining `.borrow()`/`.borrow_mut()` calls in `object/` are either in non-`JsResult` contexts (`bool`, `Option`, `()` returns), test blocks, or the foundation `borrow()`/`borrow_mut()` methods themselves  ,all left as-is per suggestion.

core/engine/src/object/builtins/jsdataview.rs
```diff
@@ -110,7 +113,13 @@ impl JsDataView {
 
         // 11. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
         // 12. Set bufferByteLength to ArrayBufferByteLength(buffer, seq-cst).
-        let Some(buf_byte_len) = buffer.borrow().data().bytes().map(|s| s.len() as u64) else {
+        let Some(buf_byte_len) = buffer
+            .try_borrow()
+            .map_err(|e| PanicError::new(e.to_string()))?
```
jedel1043, line 118: Instead of manually doing the conversion every time, it might be better to implement `From<Borrow/Mut/Error> for PanicError` or whichever error type is better to allow automatic `?` propagation.

jedel1043 (changes requested):

PR #5312, opened 2026-04-09
Implement support for compact formatting in NumberFormat

jedel1043:
Implements compact formatting for `NumberFormat` with ICU4X 2.2's newly introduced (but unstable) `CompactDecimalFormatter`.

Not worried about using unstable since this change only affects how we handle `NumberFormat` construction.


nekevss (approved):
LGTM!

PR #5313, opened 2026-04-10
fix(vm): reset implicit return value when pushing new call frames (#4…

### LINKED ISSUE #4485: Functions without return incorrectly get their return argument from the last evaluated expression

tkshsbcue:
# PR: fix(vm): reset implicit return value when pushing new call frames

Fixes #4485.

`Vm::return_value` is shared across all call frames but never gets reset in `push_frame()`. This means functions without an explicit `return` leak the caller's last evaluated expression as their return value:

```js
function f() {}
5;
f(); // returns 5 instead of undefined
```

What happens is `SetAccumulator` writes `5` into `vm.return_value` for the caller, then `f()` pushes a new frame without clearing it. When `f` returns, `handle_return()` picks up the stale `5`.

The fix resets `self.return_value = JsValue::undefined()` at the top of `push_frame()`.

### Generators

This was the main concern from the PR #4713 review — whether resetting in `push_frame()` would break generator resumption since `GeneratorContext::resume()` also calls `push_frame()` with `REGISTERS_ALREADY_PUSHED`.

Traced through the code and it's fine: `GeneratorYield` sets `return_value` explicitly before calling `handle_yield()`, and `handle_yield()` consumes it via `take_return_value()`. On resume, the generator runs more opcodes and will set a fresh `return_value` before the next yield/return. The reset only affects the initial state going into the frame, which is what we want.

Added generator-specific regression tests as @jedel1043 requested in #4713.

### Tests

- Plain functions leaking caller values
- Constructor implicit returns
- Cross-eval isolation
- Generator yield + return correctness
- Generator implicit return producing `undefined`
- Nested call chain independence

### DIFF
--- a/core/engine/src/tests/function.rs
+++ b/core/engine/src/tests/function.rs
@@ -19,6 +19,88 @@ fn empty_function_returns_undefined() {
 }
 
 #[test]
+fn implicit_return_does_not_leak_last_evaluated_expression() {
+    run_test_actions([TestAction::assert_eq(
+        indoc! {r#"
+            let seen;
+
+            function g(inner) {
+                seen = inner;
+            }
+
+            function f() {}
+
+            5;
+            let outer = g(f());
+            seen;
+        "#},
+        JsValue::undefined(),
+    )]);
+}
+
+#[test]
+fn implicit_constructor_return_uses_constructed_object() {
+    run_test_actions([TestAction::assert(indoc! {r#"
+        let seen;
+
+        function g(inner) {
+            seen = inner;
+        }
+
+        function f() {}
+
+        ({});
+        let outer = g(new f());
+        Object.getPrototypeOf(seen) === f.prototype;
+    "#})]);
+}
+
+#[test]
+fn implicit_return_does_not_leak_from_previous_eval() {
+    run_test_actions([
+        TestAction::run("function f() {}"),
+        TestAction::run("let seen; function g(inner) { seen = inner; }"),
+        TestAction::run("5;"),
+        TestAction::run("let outer = g(f());"),
+        TestAction::assert_eq("seen", JsValue::undefined()),
+    ]);
+}
+
+#[test]
+fn implicit_return_does_not_leak_completion_value_from_same_eval() {
+    run_test_actions([
+        TestAction::run(indoc! {r#"
+            eval(`
+                function f() {}
+                globalThis.seen = "initial";
+                function g(inner) { globalThis.seen = inner; }
+                5;
+                let outer = g(f());
+            `);
+        "#}),
+        TestAction::assert_eq("globalThis.seen", JsValue::undefined()),
+    ]);
+}
+
+#[test]
+fn implicit_constructor_return_does_not_leak_completion_value_from_same_eval() {
+    run_test_actions([
+        TestAction::run(indoc! {r#"
+            eval(`
+                function f() {}
+                globalThis.prototype_matches = false;
+                function g(inner) {
+                    globalThis.prototype_matches = Object.getPrototypeOf(inner) === f.prototype;
+                }
+                ({});
+                let outer = g(new f());
+            `);
+        "#}),
+        TestAction::assert("globalThis.prototype_matches"),
+    ]);
+}
+
+#[test]
 fn property_accessor_member_expression_dot_notation_on_function() {
     run_test_actions([TestAction::assert_eq(
         indoc! {r#"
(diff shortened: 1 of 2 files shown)


core/engine/src/tests/function.rs
```diff
@@ -198,3 +198,93 @@ fn function_constructor_nested_lexical_binding() {
...
+/// when they lack an explicit return, regardless of the caller's state.
+#[test]
+fn nested_calls_implicit_return_independent() {
+    run_test_actions([TestAction::assert(indoc! {r#"
+        function a() {}
+        function b() { 10; return a(); }
+        function c() { 20; return b(); }
+        c() === undefined;
+    "#})]);
+}
```
jedel1043: Honestly I preferred #4485's tests because it tested both that nested functions didn't leak into each other and that functions didn't leak to the outer scope. I would prefer if you just took that PR, added the additional tests and added the author as a coauthor.

jedel1043: to add them as a coauthor you just need to take the commits of that PR, add a commit with all your changes above that, push everything to this PR, then the CI will take care of adding you and them as coauthors

tkshsbcue:
@jedel1043 just took your review into account!

core/engine/src/tests/function.rs
```diff
@@ -18,6 +18,127 @@ fn empty_function_returns_undefined() {
...
+        }
+        var g = gen();
+        var a = g.next();
+        var b = g.next();
+        var c = g.next();
+        a.value === 1 && !a.done &&
+        b.value === 2 && !b.done &&
+        c.value === 3 && c.done;
+    "#})]);
+}
```
jedel1043, line 123: This is already covered by multiple other tests (test262, for example)

core/engine/src/tests/function.rs
```diff
@@ -18,6 +18,127 @@ fn empty_function_returns_undefined() {
...
+            yield 1;
+        }
+        99;
+        var g = gen();
+        var a = g.next();
+        var b = g.next();
+        a.value === 1 && !a.done &&
+        b.value === undefined && b.done;
+    "#})]);
+}
```
jedel1043, line 145: This is kinda not the test we were talking about. The test was more about "checking if generator resumption is not affected by resetting the accumulator value", which can only be tested if you call the `next` function of a generator with a call to another function as its parameter:
```js
var g = gen();
g.next(f())
```
and
```js
var g = gen();
g.next(new f())
```

Basically the same as the other tests but with the outer function being a generator instead of an ordinary function.

jedel1043 (changes requested):

core/engine/src/tests/function.rs
tkshsbcue: sorry my bad, did fix it hope you could have a look was overthinking stuff haha

tkshsbcue: hey @jedel1043 hope you can have a look once again Thanks!

jedel1043 (approved):
Cool, thanks!

PR #5320, opened 2026-04-11
fix(runtime): set Response.url and Response.redirected correctly after fetch redirects

### LINKED ISSUE #5310: fetch() keeps the original URL after redirects
`fetch()` follows redirects, but the resulting `Response` still reports the original URL and `redirected === false`.

Repro against a local server that returns `302 Location: /target#frag`:

```js
const r = await fetch("http://127.0.0.1:8131/redirect");
console.log(r.redirected, r.url);
```

Current Boa:
- `false`
- `http://127.0.0.1:8131/redirect`

Expected:
- `true`
- `http://127.0.0.1:8131/target`

This matches Node and the Fetch spec for `Response.redirected` and `Response.url`.

Cypher-CP0:
This Pull Request fixes/closes #5310.

It changes the following:
- Added a `redirected: bool` field to `JsResponse` struct in `response.rs`
- Updated `JsResponse::basic()` to accept a `redirected` parameter and store it
- Fixed the `Response.redirected` getter to return the stored value instead of always returning `false`
- Updated `fetchers.rs` to capture the final URL from reqwest before consuming
  the response body, and derive `redirected` by comparing it to the original request URL
- Updated `Response.url` to return the final URL after redirects instead of the original request URL
- Added `response_redirected_flag` test in `e2e.rs` covering both the redirected and non-redirected cases

### DIFF
--- a/core/runtime/src/fetch/response.rs
+++ b/core/runtime/src/fetch/response.rs
@@ -123,7 +123,7 @@ fn is_valid_reason_phrase(s: &str) -> bool {
 #[derive(Clone, Debug, Trace, Finalize, JsData)]
 pub struct JsResponse {
     url: JsString,
-
+    redirected: bool,
     #[unsafe_ignore_trace]
     r#type: ResponseType,
 
@@ -155,6 +155,7 @@ impl JsResponse {
 
         Self {
             url,
+            redirected,
             r#type: ResponseType::Basic,
             status,
             status_text,
@@ -170,6 +171,7 @@ impl JsResponse {
     pub fn error() -> Self {
         Self {
             url: js_string!(""),
+            redirected: false, 
             r#type: ResponseType::Error,
             // A network error's status is always 0.
             // See https://fetch.spec.whatwg.org/#concept-network-error
@@ -443,13 +445,14 @@ impl JsResponse {
     fn redirected(&self) -> bool {
         // The spec says: return true if this's response's URL list's size is greater than 1.
         // TODO: track the full URL list to implement this properly.
-        false
+        self.redirected
     }
 
     #[boa(rename = "clone")]
     fn clone_response(&self) -> Self {
         Self {
             url: self.url.clone(),
+            redirected: self.redirected,
             r#type: self.r#type,
             status: self.status,
             status_text: self.status_text.clone(),
--- a/core/runtime/src/fetch/fetchers.rs
+++ b/core/runtime/src/fetch/fetchers.rs
@@ -39,6 +39,7 @@ impl Fetcher for BlockingReqwestFetcher {
         _context: &RefCell<&mut Context>,
     ) -> JsResult<JsResponse> {
         use boa_engine::{JsError, JsString};
+        use http::response;
 
         if let Some(ref sig) = signal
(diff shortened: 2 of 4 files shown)


core/runtime/src/fetch/response.rs
```diff
@@ -443,13 +445,14 @@ impl JsResponse {
     fn redirected(&self) -> bool {
         // The spec says: return true if this's response's URL list's size is greater than 1.
         // TODO: track the full URL list to implement this properly.
-        false
+        self.redirected
```
zhuzhu81998, line 448: how does your change align with the TODO comment and spec?

Cypher-CP0: TODO was about tracking every hop in the redirect chain, but Response.redirected only needs to know if there was any redirect, not the full list. Since reqwest already follows redirects and exposes the final URL, comparing that to the original request URL gives us exactly what we need. Full URL list tracking could still be a follow-up if needed for other spec requirements.

zhuzhu81998: > Response.redirected only needs to know if there was any redirect, not the full list

But according to the comment here, spec requires the check to be done via url list, and not a "quick" check like this?

E.g. what if A redirects to B and B redirects back to A? Does your check consider scenario like this?

Cypher-CP0: Thanks for catching that!
You're right, A→B→A would break my check. What approach would you recommend, manual redirect handling to build the full URL list, or something simpler?

zhuzhu81998: Currently the response only has `url`, but according to spec and the TODO comment, url list is also required: https://fetch.spec.whatwg.org/#concept-response-url-list . So i guess you should implement that and follow the spec for where to modify it in what way.

But fetch api is in general not complete in boa, so idk how much work that actually might be

zhuzhu81998: I am guessing once you have the list, then A -> B -> A would be handled correctly because of the url list containing multiple urls even though the resulting url might be the same as the original.

HalidOdat: Yup, we need to keep a list of urls as described in the specification, since we are using `reqwest` we can create a client that [has custom redirect policy](https://docs.rs/reqwest/latest/reqwest/blocking/struct.ClientBuilder.html#method.redirect), that will collect the url attempts.

```rust
    // part of fetcher initialization:
    let url_list = Arc::new(Mutex::new(Vec::new()));
    let url_list_clone = Arc::clone(&history);

    let custom_policy = redirect::Policy::custom(move |attempt| {
        url_list_clone.lock().expect("should not be poisoned").push(attempt.url().clone());
        reqwest::redirect::Policy::default().redirect(attempt)
    });

    let client = Client::builder()
        .redirect(custom_policy)
        .build()?;

    // making the request
    let result = self.client.execute(req);
    
    // clears and takes the list for the previous request.
    let url_list = std::mem::take(&mut *self.url_list.lock().expect("should not be poisoned"));

    let response = result.map_err(JsError::from_rust)?;
```

jedel1043:
Closing since it seems like the PR is not following what we require for properly fixing this.

PR #5322, opened 2026-04-11
add Request body methods

### LINKED ISSUE #5321: Request is missing Body methods
`Request` on current `main` does not expose the `Body` mixin.

```js
const r = new Request("http://example.com", { method: "POST", body: "x" });
console.log(typeof r.text, typeof r.json, typeof r.bytes, typeof r.bodyUsed);
```

Current Boa:
- `undefined undefined undefined undefined`

Expected:
- `function function function boolean`

The Fetch spec says `Request` includes `Body`, so this should expose the body-reading methods and `bodyUsed`.

Monti-27:
Fixes #5321

This adds the Body methods currently supported by Boa to `Request`, and tracks body usage so `Request` cloning and transfer behave correctly.

Tests:
- cargo test -p boa_runtime request -- --nocapture
- cargo make run-ci

### DIFF
--- a/core/runtime/src/fetch/body.rs
+++ b/core/runtime/src/fetch/body.rs
@@ -1,0 +1,35 @@
+use boa_engine::object::builtins::{JsPromise, JsUint8Array};
+use boa_engine::{Context, JsNativeError, JsString, JsValue};
+use std::rc::Rc;
+
+pub(super) fn bytes(body: Rc<Vec<u8>>, context: &mut Context) -> JsPromise {
+    JsPromise::from_async_fn(
+        async move |context| {
+            JsUint8Array::from_iter(body.iter().copied(), &mut context.borrow_mut()).map(Into::into)
+        },
+        context,
+    )
+}
+
+pub(super) fn text(body: Rc<Vec<u8>>, context: &mut Context) -> JsPromise {
+    JsPromise::from_async_fn(
+        async move |_| {
+            let body = String::from_utf8_lossy(body.as_ref());
+            Ok(JsString::from(body).into())
+        },
+        context,
+    )
+}
+
+pub(super) fn json(body: Rc<Vec<u8>>, context: &mut Context) -> JsPromise {
+    JsPromise::from_async_fn(
+        async move |context| {
+            let json_string = String::from_utf8_lossy(body.as_ref());
+            let json = serde_json::from_str::<serde_json::Value>(&json_string)
+                .map_err(|e| JsNativeError::syntax().with_message(e.to_string()))?;
+
+            JsValue::from_json(&json, &mut context.borrow_mut())
+        },
+        context,
+    )
+}
--- a/core/runtime/src/fetch/request.rs
+++ b/core/runtime/src/fetch/request.rs
@@ -73,6 +73,12 @@ impl RequestInit {
                 |_| js_error!(TypeError: "Request constructor: {} is an invalid method", method.to_std_string_escaped()),
             )?;
 
+            // 25. If init["method"] exists, then:
+            //     1. Let method be init["method"].
+            //     2. If method is not a method or method is a forbidden method, throw a TypeError.
+            //     3. Normalize method.
+            //     4. Set request's method to method.
+            // https://fetch.spec.whatwg.org/#dom-request
             if method.eq_ignore_ascii_case("CONNECT")
                 || method.eq_ignore_ascii_case("TRACE")
                 || method.eq_ignore_ascii_case("TRACK")
@@ -87,6 +93,7 @@ impl RequestInit {
         }
 
         if let Some(body) = &self.body {
+            // TODO: add more support types.
             if let Some(body) = body.as_string() {
                 let body = body.to_std_string().map_err(
                     |_| js_error!(TypeError: "Request constructor: body is not a valid string"),
@@ -110,7 +117,7 @@ impl RequestInit {
 /// The `Request` interface of the [Fetch API][mdn] represents a resource request.
 ///
 /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API
-#[derive(Debug, JsData, Trace, Finalize)]
+#[derive(Clone, Debug, JsData, Trace, Finalize)]
 pub struct JsRequest {
     #[unsafe_ignore_trace]
     inner: HttpRequest<Vec<u8>>,
@@ -119,17 +126,6 @@ pub struct JsRequest {
     has_body: bool,
     #[unsafe_ignore_trace]
     body_used: Cell<bool>,
-}
-
-impl Clone for JsRequest {
-    fn clone(&self) -> Self {
-        Self {
-            inner: self.inner.clone(),
-            signal: self.signal.clone(),
-            has_body: self.has_body,
-            body_used: Cell::new(self.body_used.get()),
-        }
-    }
 }
 
 impl JsRequest {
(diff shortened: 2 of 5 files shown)


core/runtime/src/fetch/request.rs
```diff
@@ -66,12 +73,6 @@ impl RequestInit {
                 |_| js_error!(TypeError: "Request constructor: {} is an invalid method", method.to_std_string_escaped()),
             )?;
 
-            // 25. If init["method"] exists, then:
-            //     1. Let method be init["method"].
-            //     2. If method is not a method or method is a forbidden method, throw a TypeError.
-            //     3. Normalize method.
-            //     4. Set request's method to method.
-            // https://fetch.spec.whatwg.org/#dom-request
```
HalidOdat, line 74: Question: Why remove this comment?

core/runtime/src/fetch/request.rs
```diff
@@ -86,7 +87,6 @@ impl RequestInit {
         }
 
         if let Some(body) = &self.body {
-            // TODO: add more support types.
```
HalidOdat, line 89: Question: Why remove the todo comment? I doesn't seem like we covered all the supported types: https://fetch.spec.whatwg.org/#concept-bodyinit-extract

core/runtime/src/fetch/request.rs
```diff
@@ -110,36 +110,85 @@ impl RequestInit {
...
+impl Clone for JsRequest {
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            signal: self.signal.clone(),
+            has_body: self.has_body,
+            body_used: Cell::new(self.body_used.get()),
+        }
+    }
 }
```
HalidOdat: Question: Why manually implement `clone` trait? `Cell<bool>` implements Copy so it shouldn't be an issue.

core/runtime/src/fetch/request.rs
```diff
@@ -110,36 +110,85 @@ impl RequestInit {
...
-        (request, signal)
+        (request, signal, self.has_body)
     }
 
     /// Get a reference to the inner `http::Request` object.
     pub fn inner(&self) -> &HttpRequest<Vec<u8>> {
         &self.inner
     }
 
-    /// Get the abort signal associated with this request, if any.
```
HalidOdat, line 138: Question: Why remove this doc comment?

core/runtime/src/fetch/request.rs
```diff
@@ -203,23 +254,57 @@ impl JsRequest {
...
+        self.is_body_used()
     }
 
     #[boa(rename = "clone")]
-    fn clone_request(&self) -> Self {
-        self.clone()
+    fn clone_request(&self) -> JsResult<Self> {
+        self.ensure_body_unused()?;
+        Ok(self.clone())
+    }
```
HalidOdat, line 319: Suggestion: We should include the specification steps for this method.

```
The clone() method steps are:

    1. If this is unusable, then throw a TypeError.
    2. Let clonedRequest be the result of cloning this’s request.
    4. Assert: this’s signal is non-null.
    5. Let clonedSignal be the result of creating a dependent abort signal from « this’s signal », using AbortSignal and this’s relevant realm.
    6. Let clonedRequestObject be the result of creating a Request object, given clonedRequest, this’s headers’s guard, clonedSignal and this’s relevant realm.
    7. Return clonedRequestObject.```

File: core/runtime/src/fetch/request.rs [hunk 6]
```diff
@@ -203,23 +254,57 @@ impl JsRequest {
...
+        if !body_overridden && let Some(source_request) = source_request {
+            source_request.borrow().data().mark_body_used();
+        }
+
+        Ok(request)
+    }
+
+    #[boa(getter)]
+    fn body_used(&self) -> bool {
+        self.is_body_used()
```
HalidOdat, line 304: Important: The method should always contain the [full specification steps with a link](https://fetch.spec.whatwg.org/#dom-body-bodyused), even if they can't be implemented right now:

```
The bodyUsed getter steps are to return true if this’s body is non-null and this’s body’s stream is disturbed; otherwise false.
```

core/runtime/src/fetch/request.rs
```diff
@@ -110,36 +110,85 @@ impl RequestInit {
...
+        Ok(())
+    }
+
+    pub(crate) fn mark_body_used(&self) {
+        if self.has_body {
+            self.body_used.set(true);
+        }
+    }
+
+    fn consume_body(&self) -> JsResult<Rc<Vec<u8>>> {
```
HalidOdat, line 191: Important: We should always add the [spec link and steps](https://fetch.spec.whatwg.org/#concept-body-consume-body) to the methods, so we can map between spec and implementation, steps that cannot be implemented should be prefixed with `TODO: `:

```
The consume body algorithm, given an object that includes Body object and an algorithm that takes a byte sequence and returns a JavaScript value or throws an exception convertBytesToJSValue, runs these steps:

    If object is unusable, then return a promise rejected with a TypeError.

    Let promise be a new promise.
    Let errorSteps given error be to reject promise with error.
    Let successSteps given a byte sequence data be to resolve promise with the result of running convertBytesToJSValue with data. If that threw an exception, then run errorSteps with that exception.

    If object’s body is null, then run successSteps with an empty byte sequence.

    Otherwise, fully read object’s body given successSteps, errorSteps, and object’s relevant global object.

    Return promise.
```

core/runtime/src/fetch/request.rs
Monti-27: I dropped it while refactoring this section, but the context is still useful i restored it

core/runtime/src/fetch/request.rs
Monti-27: I had trimmed it when narrowing the body handling here, but you are right that it still applies. I restored the TODO

core/runtime/src/fetch/request.rs
Monti-27: I originally wanted to make the body_usedhandling explicit, butderive(Clone) already does the right thing here I switched it back

core/runtime/src/fetch/request.rs
Monti-27: I dropped it while reshaping the helper methods.. restored !!!

core/runtime/src/fetch/request.rs
Monti-27: Added the spec link and the getter steps

core/runtime/src/fetch/request.rs
Monti-27: Added the spec link and the consume body steps. the stream specific parts are marked as todos since this path still uses buffered bytes

core/runtime/src/fetch/request.rs
Monti-27: Added the spec link and the relevant steps

core/runtime/src/fetch/body.rs
```diff
@@ -0,0 +1,35 @@
...
+        async move |context| {
+            let json_string = String::from_utf8_lossy(body.as_ref());
+            let json = serde_json::from_str::<serde_json::Value>(&json_string)
+                .map_err(|e| JsNativeError::syntax().with_message(e.to_string()))?;
+
+            JsValue::from_json(&json, &mut context.borrow_mut())
+        },
+        context,
+    )
+}
```
jedel1043, line 35: Why are you using async functions here? I don't see any `await`s in the callbacks, so you could use a plain `JsPromise::new`

jedel1043 (changes requested):

PR #5324, opened 2026-04-12
Fix a spec violation in `Date.prototype.toLocaleString`

### LINKED ISSUE #5323: `Date.prototype.toLocaleString` behavior violates the ECMA-402
**Describe the bug**

The `Date.prototype.toLocaleString` is not working as expected, it always adds both date and time when call it with `dateStyle`/`timeStyle` option.

**To Reproduce**
```javascript
const date = new Date(Date.UTC(2024, 2, 10, 2, 30))
console.log(date.toLocaleString("en-US", {dateStyle: "short"}))
console.log(date.toLocaleString("en-US", {timeStyle: "short"}))
```

**Expected behavior**

Running this code, `3/10/24` and `2:30 AM` should be printed, but instead we get `3/10/24, 2:30:00 AM
` and `March 10, 2024 at 2:30 AM`. The expected behavior can be found in steps 30-31 of [this ECMAScript specification part][spec].

[spec]: https://tc39.es/ecma402/#sec-createdatetimeformat

Vellumic:
This Pull Request fixes/closes #5323 

**Changes**
- Eliminates the code which broke the standard behavior, when calling `Date.prototype.toLocaleString` with `dateStyle/timeStyle` options.
- Adds a regression test to ensure that `dateStyle` and `timeStyle` don't force a fallback to a combined date-time string, as per ECMA-402

**Validation**
- Run (all passed)
```
cargo test --package boa_engine --lib --features intl_bundled -- builtins::intl::date_time_format::tests --nocapture
```
- Manual check
```javascript
>> const date = new Date(Date.UTC(2024, 2, 10, 2, 30))
undefined
>> date.toLocaleString("en-US", {dateStyle: "short"})
"3/10/24"
>> date.toLocaleString("en-US", {timeStyle: "short"})
"2:30\u{202f}AM"
```

### DIFF
--- a/core/engine/src/builtins/intl/date_time_format/mod.rs
+++ b/core/engine/src/builtins/intl/date_time_format/mod.rs
@@ -1025,24 +1025,6 @@ pub(crate) fn format_date_time_locale(
     context: &mut Context,
 ) -> JsResult<JsValue> {
     let options = coerce_options_to_object(options, context)?;
-    if format_type != FormatType::Time
-        && get_option::<DateStyle>(&options, js_string!("dateStyle"), context)?.is_none()
-    {
-        options.create_data_property_or_throw(
-            js_string!("dateStyle"),
-            JsValue::from(js_string!("long")),
-            context,
-        )?;
-    }
-    if format_type != FormatType::Date
-        && get_option::<TimeStyle>(&options, js_string!("timeStyle"), context)?.is_none()
-    {
-        options.create_data_property_or_throw(
-            js_string!("timeStyle"),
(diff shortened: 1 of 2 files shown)


HalidOdat (approved):
Looks good to me! 😄

Vellumic:
Hi everyone! Just wanted to gently bump this PR. Let me know if you need any changes or refactoring from my side. Thanks!

jedel1043:
BTW, since I see that this behaviour is not covered by Test262, maybe you should consider upstreaming the test case onto the https://github.com/tc39/test262 repo

PR #5325, opened 2026-04-13
fix(builtins): replace PanicError with proper JsNativeError in JsRegExp

### LINKED ISSUE #4818: refactor(engine): remove expect()-based panics from JsRegExp wrapper

tkshsbcue:
# PR: fix(builtins): replace PanicError with proper JsNativeError in JsRegExp

Part of #3241. Closes #4818.

`JsRegExp` was using `js_expect()` everywhere, which wraps errors in `PanicError` — not great for API consumers since they can't match on it or get a meaningful message. Swapped all of them out for `JsNativeError::typ()` with descriptive messages per method.

For `exec()` I also noticed `to_object()` and `from_object()` already return `JsResult`, so the `js_expect()` wrappers were redundant — just using `?` directly now.

Dropped the `JsExpect` and `PanicError` imports since nothing in the file needs them anymore.

### DIFF
--- a/core/engine/src/object/builtins/jsregexp.rs
+++ b/core/engine/src/object/builtins/jsregexp.rs
@@ -1,8 +1,7 @@
 //! A Rust API wrapper for Boa's `RegExp` Builtin ECMAScript Object
 use crate::{
-    Context, JsExpect, JsNativeError, JsResult, JsValue,
+    Context, JsError, JsNativeError, JsResult, JsValue,
     builtins::RegExp,
-    error::PanicError,
     object::{JsArray, JsObject},
     value::TryFromJs,
 };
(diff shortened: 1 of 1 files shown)


jedel1043:
Those are panics because erroring there is a bug in the engine, and users are not supposed to match against engine bugs.

tkshsbcue:
my bad, thanks for rectification!

PR #5326, opened 2026-04-13
fix(lexer): replace Error::unexpected with Error::syntax for unknown …

tkshsbcue:
## fix(lexer): replace Error::unexpected with Error::syntax for unknown operators

Resubmit of #4985 (approved but closed before merging due to automation).

The operator lexer's catch-all arm used `Error::unexpected`, which produces a generic "expected valid operator" message. This replaces it with `Error::syntax` and a formatted message that includes the actual character, making it easier to diagnose malformed input.

Also removes the now-unused `Error::unexpected` method from the lexer error module to keep `-D warnings` clean. The parser error module has its own `unexpected` method which is unaffected.

### Changes

- `core/parser/src/lexer/operator.rs` — switch from `Error::unexpected` to `Error::syntax` with descriptive format string
- `core/parser/src/lexer/error.rs` — remove unused `unexpected` method

### DIFF
--- a/core/parser/src/lexer/error.rs
+++ b/core/parser/src/lexer/error.rs
@@ -41,20 +41,6 @@ impl Error {
     {
         Self::Syntax(err.into(), pos.into())
     }
-
-    /// Creates an "unexpected" syntax error (found X, with message).
-    #[inline]
-    pub(crate) fn unexpected<F, M, P>(found: F, message: M, pos: P) -> Self
-    where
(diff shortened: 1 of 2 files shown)


jedel1043:
Operators are u8's, so they're always valid chars. This doesn't do anything aside from making error creation more complex and changing the error message.

tkshsbcue:
sorry got it!

PR #5328, opened 2026-04-13
perf(array): add dense fast path for Array.prototype.splice

tkshsbcue:
Related to #3407.

Right now `splice` on a dense array still goes through the full generic path — individual `Get`/`Set`/`DeletePropertyOrThrow` per element. Same approach as the existing `shift` fast path, this skips all of that and operates on the `ThinVec` directly via `ThinVec::splice`.

Handles `DenseI32`, `DenseF64`, and `DenseElement`. If replacement items don't match the storage type (e.g. string into an i32 array), converts to `DenseElement` first.

The fast path only kicks in when:
- It's a real array with the default shape (no subclass / custom `@@species`)
- `itemCount <= actualDeleteCount`  so no new indices are created beyond `len - 1`, avoiding the prototype-chain-setter issue from #5076
- `len <= vec.len()` (storage matches length)

Falls through to the generic path otherwise.

~25% faster on a benchmark doing repeated splices on 500-element dense arrays.

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -2321,6 +2321,73 @@ impl Array {
                 .into());
         }
 
+        // Dense fast path: operate directly on the underlying storage vec.
+        // Only safe when itemCount <= actualDeleteCount, since inserting more
+        // items than deleted would create new indices beyond `len - 1` where
+        // prototype chain setters could fire (see #3407, #5076).
+        // Also requires the default array shape so that ArraySpeciesCreate
+        // would return a plain Array (no subclass with custom @@species).
+        if o.is_array() && item_count <= actual_delete_count {
+            let start = actual_start as usize;
+            let delete_count = actual_delete_count as usize;
+            let default_shape = context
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/array/mod.rs
```diff
@@ -2321,6 +2321,73 @@ impl Array {
...
+            } else {
+                None
+            };
+
+            if let Some(deleted) = maybe_deleted {
+                drop(o_borrow);
+                let new_len = len - actual_delete_count + item_count;
+                Self::set_length(&o, new_len, context)?;
+                let arr = Self::create_array_from_list(deleted, context);
+                return Ok(arr.into());
```
jedel1043, line 2387: Just put these in the matches themselves. It saves the trouble of having to `if let` afterwards and it looks a bit cleaner.

tkshsbcue: alright!

jedel1043:
Closing since the comments have not been addressed. Feel free to reopen whenever that is in place

PR #5329, opened 2026-04-13
perf(array): add dense array fast-path for Array.prototype.unshift

Exar04:
ref: #3407

### DIFF
--- a/core/engine/src/builtins/array/mod.rs
+++ b/core/engine/src/builtins/array/mod.rs
@@ -1291,6 +1291,48 @@ impl Array {
                     .with_message("length + number of arguments exceeds the max safe integer limit")
                     .into());
             }
+
+            if o.is_array() {
+                let mut o_borrow = o.borrow_mut();
+                let props = &mut o_borrow.properties_mut().indexed_properties;
+
+                let fast_path = match props {
+                    IndexedProperties::DenseI32(dense) if dense.len() as u64 == len => {
+                        let all_i32: Option<ThinVec<i32>> =
+                            args.iter().map(JsValue::as_i32).collect();
+                        if let Some(items) = all_i32 {
+                            dense.splice(0..0, items);
+                            true
+                        } else {
+                            false
+                        }
+                    }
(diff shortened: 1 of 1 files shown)


core/engine/src/builtins/array/mod.rs
```diff
@@ -1291,6 +1291,48 @@ impl Array {
                     .with_message("length + number of arguments exceeds the max safe integer limit")
                     .into());
             }
+
+            if o.is_array() {
+                let mut o_borrow = o.borrow_mut();
+                let props = &mut o_borrow.properties_mut().indexed_properties;
+
+                let fast_path = match props {
+                    IndexedProperties::DenseI32(dense) if dense.len() as u64 == len => {
```
Exar04, line 1300: Here we are iterating over values that will be added to array and converting it to type i32.
But what if we encounter string or f64? according to blog shared in issue, the type of array should get converted to lower type i.e DenseI32 -> DenseF64.
But we aren't doing that here (even before I added this code), is it getting handled somewhere else?


also if I am right about above, instead of getting out of match when encountering f64 instead of i32 it would be more performant to try again match.

something like this -> 
```rust
fn main() {
    let mut b = 1;
    let a = loop {
        match b {
        1 => {
            b = 2;
            if b == 2{
                continue
            }
            break "one"
        },
        2 => break "two",
        _ => break "other",
    }
    };
    println!("{}", a);
}
```
but idk... its just a thought

core/engine/src/builtins/array/mod.rs
```diff
@@ -1291,6 +1291,48 @@ impl Array {
...
+                    }
+                    _ => false,
+                };
+
+                drop(o_borrow);
+
+                if fast_path {
+                    let new_len = len + arg_count;
+                    Self::set_length(&o, new_len, context)?;
+                    return Ok(new_len.into());
```
jedel1043, line 1332: I would put these three lines on each fast path instead. Even though it is a bit verbose, it doesn't require having complex branches, it's just a "if dense and all types match, push to array, adjust the length and return"

jedel1043:
Closing since the review has not been addressed. Feel free to reopen when that is fixed

PR #5336, opened 2026-04-15
Align `Intl.DateTimeFormat` constructor with spec-mandated initialization order

### LINKED ISSUE #5335: `Intl.DateTimeFormat` constructor breaks the inheritance when passing wrong options
**Describe the bug**
The prototype lookup never happens when creating an `Intl.DateTimeFormat` with wrong options.

**To Reproduce**
```javascript
const proxyConstructor = new Proxy(Intl.DateTimeFormat, {
  get(target, prop) {
    if (prop === "prototype") {
      console.log("1. Prototype accessed");
    }
    return target[prop];
  }
});

try {
  new proxyConstructor("en", { timeZone: "Invalid/Zone" });
} catch (e) {
  console.log("2. Error thrown");
}
```

**Expected behavior**
Running this code, we should see this in console
```
1. Prototype accessed
2. Error thrown
```
But get this instead
```
2. Error thrown
```

Vellumic:
This Pull Request fixes/closes #5335.

It changes the following:

- Swap the prototype lookup and `DateTimeFormat` creation to match the spec steps order
- Add a regression test for this specific case
- Add a step 1 from spec to `create_date_time_format` as well as comment about why it's missing

**Validation**
- Run (all passed)
```
cargo test --package boa_engine --lib --features intl_bundled -- builtins::intl::date_time_format::tests --nocapture
```

### DIFF
--- a/core/engine/src/builtins/intl/date_time_format/mod.rs
+++ b/core/engine/src/builtins/intl/date_time_format/mod.rs
@@ -177,16 +177,16 @@ impl BuiltInConstructor for DateTimeFormat {
         let options = args.get_or_undefined(1);
 
         // 2. Let dateTimeFormat be ? CreateDateTimeFormat(newTarget, locales, options, any, date).
+        let prototype = get_prototype_from_constructor(
+            new_target_inner,
+            StandardConstructors::date_time_format,
+            context,
+        )?;
         let dtf = create_date_time_format(
             locales,
             options,
             FormatType::Any,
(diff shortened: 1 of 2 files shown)


Vellumic:
Hi everyone! Just wanted to gently bump this PR. Let me know if you need any changes or refactoring from my side. Thanks!

jedel1043 (approved):
LGTM!

PR #5344, opened 2026-04-22
fix eval delete bindings

### LINKED ISSUE #5333: eval-created local bindings survive delete
Direct `eval` inside a function creates local `var` and function bindings that should be deletable, but Boa keeps them alive after `delete`.

```js
(function () {
  var after;
  eval("delete x; after = () => x; var x;");
  try { after(); console.log("no throw"); } catch (e) { console.log(e.name); }
}());

(function () {
  var after;
  eval("delete f; after = () => f; function f() {}");
  try { after(); console.log("no throw"); } catch (e) { console.log(e.name); }
}());
```

Current Boa:
- `no throw`
- `no throw`

Expected:
- `ReferenceError`
- `ReferenceError`

This also matches the failing test262 cases:
- `test/language/eval-code/direct/var-env-var-init-local-new-delete.js`
- `test/language/eval-code/direct/var-env-func-init-local-new-delete.js`

Monti-27:
Fixes #5333.

Direct eval-created local `var` and function bindings were not treated as deletable at runtime, so `delete` left them visible and later `var` initialization could recreate them.

This tracks deleted eval-created local bindings in function environments and avoids recreating them after deletion.

Tests:
- cargo test -p boa_engine env -- --nocapture
- cargo test -p boa_engine unary_delete -- --nocapture
- cargo clippy -p boa_engine --all-features --all-targets -- -D warnings
- cargo make run-ci

### DIFF
--- a/core/engine/src/environments/runtime/declarative/function.rs
+++ b/core/engine/src/environments/runtime/declarative/function.rs
@@ -1,7 +1,15 @@
 use boa_ast::scope::Scope;
 use boa_gc::{Finalize, GcRefCell, Trace, custom_trace};
+use rustc_hash::FxHashSet;
+use std::cell::RefCell;
 
 use crate::{JsNativeError, JsObject, JsResult, JsValue, builtins::function::OrdinaryFunction};
+
+#[derive(Debug, Default)]
+struct BindingDeletionState {
+    deletable_bindings: FxHashSet<u32>,
+    deleted_bindings: FxHashSet<u32>,
+}
 
 #[derive(Debug, Trace, Finalize)]
 pub(crate) struct FunctionEnvironment {
@@ -11,6 +19,9 @@ pub(crate) struct FunctionEnvironment {
     // Safety: Nothing in `Scope` needs tracing.
     #[unsafe_ignore_trace]
     scope: Scope,
+
+    #[unsafe_ignore_trace]
+    binding_deletion_state: Box<RefCell<BindingDeletionState>>,
 }
 
 impl FunctionEnvironment {
@@ -20,6 +31,7 @@ impl FunctionEnvironment {
             bindings: GcRefCell::new(vec![None; bindings_count as usize]),
             slots: Box::new(slots),
             scope,
+            binding_deletion_state: Box::default(),
         }
     }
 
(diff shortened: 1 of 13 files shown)


zhuzhu81998 (commented):
i am not familiar with this part of the codebase, but this feels, kinda hacky?

are you sure this is not something that happens because of incompliance with spec somewhere?

Monti-27:
> i am not familiar with this part of the codebase, but this feels, kinda hacky?
> 
> are you sure this is not something that happens because of incompliance with spec somewhere?

Yeah, fair point. the current approach is not clean enough. The Test262 report also caught a regression around eval("var x=delete(x)") so I’ll rework this instead of trying to patch around it

hansl:
Let us know when this is ready to review, or close it and re-open a new PR :)

Monti-27:
this new changes moves deletability into the function environment itself, marks newly extended eval bindings as deletable, and re resolves stale deleted locators before assignment. this matches the EvalDeclarationInstantiation / CreateMutableBinding(name, true) behavior more directly

Monti-27:
Updated this now. The fix keeps deleted eval-created bindings in the function environment and uses a separate eval declaration var path, so normal var statements do not recreate deleted bindings.

Tests:
- cargo test -p boa_engine tests::env --lib
- cargo clippy -p boa_engine -- -D warnings
- cargo make run-ci from the pre-push hook
- targeted Test262 eval delete cases

Monti-27:
@hansl its ready for review !!

Monti-27:
@jedel1043 any nit picks here

core/engine/src/environments/runtime/declarative/function.rs
```diff
@@ -6,6 +6,8 @@ use crate::{JsNativeError, JsObject, JsResult, JsValue, builtins::function::Ordi
 #[derive(Debug, Trace, Finalize)]
 pub(crate) struct FunctionEnvironment {
     bindings: GcRefCell<Vec<Option<JsValue>>>,
+    deletable_bindings: GcRefCell<Vec<bool>>,
```
hansl: These could be `RefCell<...>` with `#[unsafe_ignore_trace]` as `Vec<bool>` does not need to be traced by the GC.

hansl (approved):
I'm good with this, assuming the GcRefCell (which is expensive to allocate/drop) is fixed.

Monti-27:
> I'm good with this, assuming the GcRefCell (which is expensive to allocate/drop) is fixed.

fixed !!!!

PR #5345, opened 2026-04-22
Fix: boa_cli returns non-zero exit code on uncaught exceptions

### LINKED ISSUE #4962: boa_cli exit code 0 on uncaught runtime exceptions

Pantkartik:
Description: Currently, the boa CLI returns an exit code of 0 even when a script fails with an uncaught runtime exception. This PR modifies evaluate_file and evaluate_expr in cli/src/main.rs to propagate errors back to the main function, ensuring a non-zero exit code is returned on failure.

Fixes: #4962

Verification:

Verified that boa returns exit code 1 for scripts with uncaught errors.
Verified that successful scripts still return exit code 0.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -410,33 +410,42 @@ fn evaluate_expr(
             Ok(v) => println!("{v}"),
             Err(v) => eprintln!("{v:?}"),
         }
-    } else {
-        let mut counters = Counters::new(args.time);
-        let script = {
-            let _timer = counters.new_timer("Parsing");
-            Script::parse(Source::from_bytes(line), None, context)
-        };
-
-        match script {
-            Ok(script) => {
-                let result = {
-                    let _timer = counters.new_timer("Execution");
-                    let result = script.evaluate(context);
-                    if let Err(err) = context.run_jobs() {
-                        printer.print(uncaught_job_error(&err));
-                    }
-                    result
-                };
-                match result {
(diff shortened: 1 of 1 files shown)


jedel1043:
Duplicate of #4964

PR #5346, opened 2026-04-22
Fix console grouping indentation and add compatibility methods

Pantkartik:
This PR fixes indentation in console.group() and adds console.exception(), console.timeStamp(), console.profile(), and console.profileEnd().


jedel1043:
Please don't push unrelated changes from a separate PR.

Also, no-op implementations of methods are not very useful; it gives zero notice for users that the methods won't return correct results.

PR #5356, opened 2026-05-06
perf: Simple optimization

hansl:
Prior to this change, a lot of assignments defaulted to something like `Move SRC, TMP`; `PushRegister TMP`. With TMP being a temporary register. These are redundant as we could simple PushRegister SRC. This change removes those duplicated temporary registers.

Another part of this change is removing the temporary register when building an array literal; each non-spread elements were assigned to a temp register instead of directly being pushed on the array.

Both changes are improving the benchmarks by about 1-4%.


jedel1043 (approved):
LGTM

PR #5379, opened 2026-05-18
fix(runtime): make timer IDs 1-based to avoid sentinel collision

### LINKED ISSUE #5378: [Bug]: `setTimeout`/`setInterval` return timer ID `0` for first timer, colliding with no-op sentinel
### Problem
`IntervalInnerState::next_id()` returns the internal counter before incrementing it. The counter starts at `0` via `Default`, so the first valid timer gets handle `0`:

```rust
// core/runtime/src/interval.rs:38-44
fn next_id(&mut self) -> JsResult<u32> {
    self.active_map.retain(|_, v| !v.revoked());
    let id = self.id;       // 0 on first call
    self.id = id.checked_add(1)...;
    Ok(id)                  // returns 0
}
```

Both `set_timeout` (L68) and `set_interval` (L123) return `Ok(0)` as a sentinel when no callback is provided:

```rust
let Some(function_ref) = function_ref else {
    return Ok(0);
};
```

The first real timer and the "nothing was scheduled" sentinel are now indistinguishable. This violates WHATWG HTML [§8.6 Timers](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers), which requires timer handles to be positive integers, and breaks the ubiquitous `if (timer) clearTimeout(timer)` idiom for the first timer in a context.

### Reproduction (`main` at `d2caa75c`):

```bash
cargo run -p boa_cli -- -e \
  "console.log(setTimeout(() => {}, 10)); \
   console.log(setTimeout(() => {}, 10)); \
   console.log(setTimeout())"
```

``` 
0     ← first valid timer (should be > 0)
1
0     ← no-callback sentinel (collides)
```
### Root Cause
Introduced in `8820b777` (#5289). The pre-refactor code incremented before returning, so the first ID was always `1`.

iammdzaidalam:
Closes #5378

## Problem

`IntervalInnerState::next_id()` returns the internal counter before incrementing it. Since the counter starts at `0` via `Default`, the first valid timer gets handle `0`:

```rust
fn next_id(&mut self) -> JsResult<u32> {
    self.active_map.retain(|_, v| !v.revoked());
    let id = self.id;
    self.id = id.checked_add(1)...;
    Ok(id)
}
```

At the same time, both `set_timeout` and `set_interval` return `Ok(0)` when no callback is provided. That makes the first real timer and the "nothing was scheduled" sentinel indistinguishable. It also violates WHATWG HTML timer semantics, which require positive timer handles, and breaks common JavaScript code that treats a returned timer ID as truthy.

## Fix

Update `next_id()` to increment the counter before returning it, so valid timer IDs become 1-based again. This preserves `0` for the no-callback sentinel and restores the pre-#5289 behavior.

## Tests

* `timer_ids_are_positive_and_unique` asserts that timer IDs returned by `setTimeout` and `setInterval` are greater than `0` and unique.
* `no_callback_returns_zero_sentinel` asserts that `setTimeout()` with no callback still returns `0` and does not collide with a valid timer ID.

## Verification

* `cargo test -p boa_runtime`
* `cargo run -p boa_cli -- -e "console.log(setTimeout(() => {}, 10)); console.log(setTimeout(() => {}, 10)); console.log(setTimeout())"`

Before the fix, the first valid timer returned `0`. After the fix, the output is `1`, `2`, `0`.


core/runtime/src/interval/tests.rs
jedel1043, line 1: I'm pretty sure these kinds of tests are already in our WPT suite, they're just disabled. It might be better to enable those instead.

core/runtime/src/interval.rs
```diff
@@ -37,11 +37,11 @@ impl IntervalInnerState {
...
     fn next_id(&mut self) -> JsResult<u32> {
         self.active_map.retain(|_, v| !v.revoked());
-        let id = self.id;
-        self.id = id
+        self.id = self
+            .id
             .checked_add(1)
             .ok_or_else(|| js_error!(Error: "Interval ID overflow"))?;
-        Ok(id)
+        Ok(self.id)
```
jedel1043: IMO the correct fix would be to use `NonZeroU32` then converting at the end of `setTimeout`, making zero as an id unrepresentable.

jedel1043 (changes requested):

core/runtime/src/interval/tests.rs
iammdzaidalam: Hi @jedel1043, Checked the WPT runner, the timers category isn't enabled yet so those tests aren't actually running. Only console, encoding, url and fetch are set up. Want me to enable those instead of keeping the custom tests here?

core/runtime/src/interval.rs
iammdzaidalam: Fair point, missed that angle. `NonZeroU32` is cleaner since zero-as-ID becomes impossible at the type level and we get a free zero-check in `clearTimeout` too. Should I go ahead and rework it around that?

core/runtime/src/interval/tests.rs
jedel1043: Yes... that's literally what I said...

core/runtime/src/interval.rs
jedel1043: yea

core/runtime/src/interval/tests.rs
iammdzaidalam: Quick question, at line 87 in `core/runtime/src/interval.rs`, why are we clamping the delay here instead of following the WebIDL `long` conversion behavior?

```rust
let delay = u64::from(delay.clamp_finite(0, u32::MAX));
```

I was trying the WPT timers suite locally, and tests like `type-long-settimeout` / `type-long-setinterval` fail because values like `Math.pow(2, 32)` seem to get clamped to `u32::MAX` instead of wrapping back to `0`, so Boa ends up scheduling a ~49 day timeout.

Feels like that’s a separate timer/WebIDL issue beyond the timer ID collision fix, but I wanted to check if the current behavior here was intentional.

iammdzaidalam: i mean if you run : `cargo run -p boa_cli -- -e "setTimeout(() => console.log('Should fire instantly!'), Math.pow(2, 32));"` ... hangs and does absolutely nothing  

while something like node : `node -e "setTimeout(() => console.log('Should fire instantly!'), Math.pow(2, 32));"` prints:
```
TimeoutOverflowWarning: 4294967296 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
Should fire instantly!
```

jedel1043: Not intentional IIRC. I think our implementation of timers is very ad-hoc, and many things will be non-conformant, so it would be good to take a look at the failing tests to fix up the bugs.

iammdzaidalam: Alright, I'll fix the delay conversion part first and rerun the WPTs

iammdzaidalam: btw, is there already any direction planned for the timers runtime?

jedel1043: No direction, since even the runtime itself has no direction :)
It started (and still is) as a testing environment to see if you can use Boa to build a runtime, but it shouldn't be used in production

hansl: Timeout is conformant because the spec just says that value is the number of milliseconds to wait to trigger the function. There's no mention of 32 bits.

hansl: The spec: https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html

core/runtime/src/interval.rs
```diff
@@ -45,7 +54,7 @@ impl IntervalInnerState {
     }
 
     /// Delete an interval ID from the active map.
-    fn clear_interval(&mut self, id: u32) -> Option<CancellationToken> {
+    fn clear_interval(&mut self, id: NonZeroU32) -> Option<CancellationToken> {
```
hansl: I don't think `clearInterval(0)` should error. This should be a special case in the code itself, still receiving a `u32`. Actually this should probably also not throw if `id` is not a number...

Image

iammdzaidalam: yes, clearInterval(0) doesnt throw error but returns null... but shouldnt it return undefined?
image
image

core/runtime/src/interval/tests.rs
iammdzaidalam: sorry for the late reply, i tried switching the delay conversion to `ToInt32` since the spec uses `long` there. And with that, `Math.pow(2, 32)` wraps to `0` and fires instantly locally. Timer tests pass too.

is it a right approach? should I keep this separate from the id fix or put it in this PR?

core/runtime/src/interval.rs
iammdzaidalam: I tried switching `clearTimeout` to normal js conversion... so stuff like `clearTimeout(0)`, `clearTimeout("foo")`, `clearTimeout(null)` etc dont throw anymore and behave like browsers 

tests too pass locally with this... does this approach look fine?

core/runtime/src/interval/tests.rs
hansl: Yes that would be the right approach.

iammdzaidalam:
hello @hansl @jedel1043, updated the pr.

* moved timer ids to `NonZeroU32`
* fixed `clearTimeout` / `clearInterval` behavior
* switched delay conversion to `to_i32()` (for webidl long behavior)
* enabled the WPT timer tests

also had to fix a windows WPT path issue, add `clear_jobs` for interval tests, and add the missing `experimental` cfg flags in `iterable/tests.rs`.

If needed i can split the last part.

tests/wpt/src/lib.rs
```diff
@@ -398,8 +385,8 @@ fn console(
 fn encoding(
     #[base_dir = "${WPT_ROOT}"]
     #[files("encoding/api-*.any.js")]
-    #[files("encoding/textencoder-constructor-non-utf.any.js")]
     // TODO: re-enable those when better encoding and options are supported.
+    // #[files("encoding/textencoder-constructor-non-utf.any.js")]
```
hansl: Is this not already included in line 391?

```rs
    // #[files("encoding/textencoder-*.any.js")]
```

hansl (commented):
I'm not convinced we need `clear_jobs`. We shouldn't need to clear jobs and should fail if jobs are not completed by the end of the run. Let me think a bit more about it.

tests/wpt/src/lib.rs
iammdzaidalam: my bad 😅 leftover from earlier testing, i'll clean it up in the next commit.

iammdzaidalam:
hi @hansl. i went with this because without `clear_jobs()` the WPT timer tests were hanging forever. some tests call `done()` but leave `setInterval`s running, so `run_jobs_async()` never exits because the jobs keep adding themselves back to the queue.

for now i used `clear_jobs()` to simulate browser teardown and get the tests passing. not sure if this is the best long term approach for Boa though, so would really love your guidance on handling this.

hansl:
@iammdzaidalam Hey Zaid. Sorry for the late reply I was out on a trip last weekend, and thinking about how to approach this.

What you actually want to do is not cancel jobs (there are jobs in there that aren't timers/intervals). You want to clear timers themselves and let the engine exit gracefully. Am I right?

In that case, what you could do is add a `clear_all()` (or `clear_timers` or similar) function to the runtime (public, in interval) that specifically cancel intervals and timers. Since you have access to the list of valid timers, it should be trivial to cancel all of them. Then the test can exit run_jobs and if there is a job hanging that is likely an actual bug we would need to fix.

What do you think?

iammdzaidalam:
Hey @hansl, hope the trip was good 😊.

tried a repro with a `Promise.resolve().then()` and an uncleared `setInterval`. With `clear_jobs()` the promise callback never ran since it clears more than just timers. With `clear_all()` the promise resolves normally and `run_jobs_async()` exits once the intervals are cancelled.

Also confirmed the hang directly. `run_jobs_async()` stayed `Pending` with a live interval until timers were cleared. 
All 7 WPT timer tests still pass. 

Also i am keeping the 10 second deadline in the harness so we notice if something else holds the loop open later.

seems like the right direction, thanks!

tests/wpt/src/lib.rs
```diff
@@ -398,7 +389,6 @@ fn console(
 fn encoding(
     #[base_dir = "${WPT_ROOT}"]
     #[files("encoding/api-*.any.js")]
-    #[files("encoding/textencoder-constructor-non-utf.any.js")]
```
hansl, line 401: No need to remove.

tests/wpt/src/lib.rs
```diff
@@ -285,6 +278,7 @@ impl TestCompletion {
 // in clippy.
 #[allow(unused)]
 fn execute_test_file(path: &Path) {
+    rustls::crypto::ring::default_provider().install_default().ok();
```
hansl, line 284: Which test requires this?

core/engine/src/context/mod.rs
```diff
@@ -500,6 +500,12 @@ impl Context {
         self.job_executor().run_jobs(self)
     }
 
+    /// Clears all queued jobs from the job executor.
+    #[inline]
```
hansl: If it's not needed, I'd rather like to leave it out of the PR.

hansl (changes requested):
Nothing major, but I'm curious why the `rustls` dependency was added.

tests/wpt/src/lib.rs
iammdzaidalam: Oops, accidental removal while fixing path handling. Restoring it.

tests/wpt/src/lib.rs
iammdzaidalam: The test runner needs it, not any specific test. Workspace reqwest is pinned with `rustls-no-provider` so `reqwest::blocking::Client` panics with `No provider set` without a `CryptoProvider` installed. `create_context()` builds a `WptFetcher` (and a reqwest Client) for every test so the suite won't run without it.

I removed the line locally to verify, encoding and timers both panic the same way. Checked main too and `cargo test encoding` fails identically there, so this is a latent harness issue ig, not something this PR introduced.

Same install already exists in `cli/src/main.rs` and `examples/src/bin/module_fetch_async.rs`. Happy to split it out or add a comment, whichever works.

core/engine/src/context/mod.rs
iammdzaidalam: yep not needed anymore after the switch to `interval::clear_all`. will remove in next push.

tests/wpt/src/lib.rs
hansl: Could you just add a comment above explaining this so we don't wonder in the future? Thanks. I think that would be good to go after this.

iammdzaidalam: ig i've already commented?

iammdzaidalam: yep... since u referred to the last commit's code. Kindly check the latest code 😊

iammdzaidalam:
added a test coverage for the clear_all function... missed it in last commit

hansl:
Waiting for #5389 to go in then I'll merge main and re-run CI.

hansl (approved):

hansl:
@jedel1043 need your approval, PTAL

jedel1043 (approved):
Looks good!

PR #5381, opened 2026-05-21
fix(vm): reserve jump-table entry 0 for try/finally fallthrough

### LINKED ISSUE #5369: Empty finally block causes function to return undefined when try block contains a conditional return
**Describe the bug**

(The bug report was written by GLM 5.1 with opencode)

When a function contains a `try/catch/finally` with an empty `finally` block, and the `try` block contains an `if` statement with a `return` that is not taken, control flow is lost after the `finally` block. The function returns `undefined` instead of continuing to subsequent statements.

**To Reproduce**

```javascript
function f(x) {
  try {
    if (x) return -1;
  } catch (e) {}
  return 42;
}
f(0); // → 42 ✓ (correct)
```

Adding an empty `finally {}` breaks it:

```javascript
function g(x) {
  try {
    if (x) return -1;
  } catch (e) {} finally {}
  return 42;
}
g(0); // → undefined ✗ (should be 42)
```

The only difference is the addition of `finally {}`. The `if (x)` condition is `false` (since `0` is falsy), so the `return -1` inside the `try` block is never executed. Control should fall through the empty `finally` and reach `return 42`, but instead the function returns `undefined`.

**Expected behavior**

Both functions should return `42`. The empty `finally {}` block should be a no-op, and control flow should continue to `return 42`.

**Build environment**

- OS: macOS
- Version: 15.7.3
- Target triple: x86_64-apple-darwin
- Rustc version: rustc 1.94.1 (e408947bf 2026-03-25)
- Boa version: git revision `3ce87f0fa03c5c037fb667f7f21a4996abbbc714` (Note: this appears to be a regression. The bug does not reproduce on the released `boa_engine` version `0.21.1`)

**Additional context**

This was discovered while running React 19's `react-reconciler` inside Boa. React's `dispatchSetStateInternal` function uses the exact pattern `if (cond) try { ... } catch (e) {} finally {}`, causing all React state updates to silently fail — the function returns `undefined`, and the code that schedules a re-render (after the `finally`) is never reached.

tkshsbcue:
The jump table emitted at the end of a `try/finally` block uses the `finally_throw_index` register to dispatch to a `break`/`continue`/`return` that was pending when control transferred into the finally. The index register is initialised to `0` and only updated by `HandleFinally`, so `0` represents "no jump record was selected — fall through past the table".

#4852 made `JumpTable` index its address array directly and dropped the explicit default address. The same PR also removed the `+1` offset that `HandleFinally` used to apply to its index, so the initial value `0` now collides with the first registered jump record. Any `return`, `break`, or `continue` syntactically present (but not executed) inside a `try` block whose `finally` runs is then taken as soon as the finally completes, even though control should fall through to the code after the `try` statement.

A minimal reproducer (which silently breaks React 19's `dispatchSetStateInternal`):

    function g(x) {
        try { if (x) return -1; } catch (e) {} finally {}
        return 42;
    }
    g(0); // returned `undefined`, should be `42`

Restore the `+1` offset in `HandleFinally` and size the emitted jump table at `N + 1` entries: entry `0` is patched to point past all of the jump-record handlers, and entries `1..=N` continue to dispatch to the registered records. A jump emitted right after the table skips the record handlers in the fallthrough case so that the new entry `0` target is only reachable through the table.
Fixes #5369 as well!

### DIFF
--- a/core/engine/src/bytecompiler/jump_control.rs
+++ b/core/engine/src/bytecompiler/jump_control.rs
@@ -117,7 +117,9 @@ impl JumpRecord {
                     finally_throw_flag,
                     finally_throw_index,
                 } => {
-                    let index = value as i32;
+                    // Note: +1 because 0 is reserved for the fallthrough entry of the
+                    // jump table emitted in `pop_try_with_finally_control_info`.
+                    let index = value as i32 + 1;
                     compiler
                         .bytecode
                         .emit_store_false(finally_throw_flag.into());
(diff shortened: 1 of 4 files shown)


nekevss (commented):
Please add a test for this to `tests/insta-bytecode`.

tkshsbcue:
hey @nekevss did that!

tkshsbcue:
hey @nekevss i hope you can review this now i have made those corrections 
Thanks!

nekevss (approved):
LGTM. Thanks for this and apologies for the delay

PR #5391, opened 2026-06-01
Fix bug in JsString ends_with

### LINKED ISSUE #5390: Investigate string ends_with() calling starts_with().
This is an extraction from #5370 by @qarmin . Note that issue was closed as being too broad and unverified, but here I'm looking at the very first potential issue in the Claude report.

First off, it's clearly incorrect for [ends_with()](https://github.com/boa-dev/boa/blob/8f5ef6542d641fd22320e51234e914b59e623717/core/string/src/lib.rs#L435) to be calling starts_with():
```
    pub fn ends_with(&self, needle: JsStr<'_>) -> bool {
        self.as_str().starts_with(needle)
    }
```
But as far as I can tell, this code is not called from anywhere. Running a JavaScript file that calls endsWith() works as expected:

ends_with.js:
```
let greeting = "Hello, world!";
let bang = "!";
console.log("Starts with a bang: ", greeting.startsWith(bang));
console.log("Ends with a bang: ", greeting.endsWith(bang));
```
```
$ cargo run -- ends_with.js
->
Starts with a bang:  false
Ends with a bang:  true
```
Should this code be removed? Fixed? Why is there no unit test? Why is it there at all?

I tried to answer why it is there. Note there are currently three definitions of ends_with():

1. core/engine/src/builtins/string/[mod.rs](https://github.com/boa-dev/boa/blob/8f5ef6542d641fd22320e51234e914b59e623717/core/engine/src/builtins/string/mod.rs#L893)
2. core/string/src/[str.rs](https://github.com/boa-dev/boa/blob/8f5ef6542d641fd22320e51234e914b59e623717/core/string/src/str.rs#L203)
3. core/string/src/[lib.rs](https://github.com/boa-dev/boa/blob/8f5ef6542d641fd22320e51234e914b59e623717/core/string/src/lib.rs#L435)

From what I can tell, (1) mod.rs is the actual implementation used for the engine, and has a [unit test](https://github.com/boa-dev/boa/blob/8f5ef6542d641fd22320e51234e914b59e623717/core/engine/src/builtins/string/tests.rs#L291). (2) str.rs is an implementation in a separate string crate, and (3) lib.rs is the incorrect wrapper implementation in the same crate.

This still leaves the question as to why there are three separate implementations instead of one shared implementation, which I have not figured out.

I also am unable to demonstrate a simple snippet that calls the broken path. I thought this would verify the problem, but it did not:

```
use boa_string::{JsStr};

fn main() {
    let greeting = JsStr::latin1(b"Hello, world!");
    let bang = JsStr::latin1(b"!");
    println!("{}", greeting.to_std_string().unwrap());
    println!("Starts with a bang: {}", greeting.starts_with(bang));
    println!("Ends with a bang: {}", greeting.ends_with(bang));
}
```
One thing I did find out is that the first definition of ends_with() outside of builtin/ was introduced in pull #3450: "This PR encodes strings that can be represented as latin1 as a byte array, instead of u16 array, this cuts latin1 strings size by 2x."

This concludes where I'm at. Hopefully somebody can fill in the missing pieces, and what should be done with the code.

nekevss:
This PR closes #5390

Looks like we're calling the wrong method in `ends_with`. Fixes the issue and adds a basic test for `starts_with` and `ends_with`

### DIFF
--- a/core/string/src/lib.rs
+++ b/core/string/src/lib.rs
@@ -432,7 +432,7 @@ impl JsString {
     // We check the size, so this should never panic.
     #[allow(clippy::missing_panics_doc)]
     pub fn ends_with(&self, needle: JsStr<'_>) -> bool {
-        self.as_str().starts_with(needle)
+        self.as_str().ends_with(needle)
     }
 
     /// Get the `u16` code unit at index. This does not parse any characters if there
--- a/core/string/src/tests.rs
+++ b/core/string/src/tests.rs
@@ -553,3 +553,15 @@ fn trim() {
     let base_str = JsString::from(" \u{000B} Hello World \t ");
     assert_eq!(base_str.trim(), JsString::from("Hello World"));
 }
+
+#[test]
+fn starts_with_and_ends_with_basic() {
+    let basic = JsString::from("abcdef");
+    let start_needle = JsStr::latin1("abc".as_bytes());
+    assert!(basic.starts_with(start_needle));
+    assert!(!basic.ends_with(start_needle));
+
+    let end_needle = JsStr::latin1("def".as_bytes());
+    assert!(!basic.starts_with(end_needle));
+    assert!(basic.ends_with(end_needle));
+}


hansl (approved):

hansl:
No test262 changes, uh.

PR #5398, opened 2026-06-16
fix: route StdClock through crate::sys::time to prevent wasm panics

### LINKED ISSUE #5397: StdClock (default Context clock) uses std::time directly, bypassing sys::time — Date panics on wasm32-unknown-unknown even with the js feature
boa_engine 0.21.1 (target wasm32-unknown-unknown, browser, non-emscripten/wasi)

Describe the bug

On wasm32-unknown-unknown (browser), any script that uses the Date builtin (Date.now(), new Date()) panics with:

panicked at library/std/src/sys/time/unsupported.rs:35:9:
time not implemented on this platform
This happens even with the js feature enabled, which is the documented way to run Boa in a browser.

Root cause: StdClock — the default Clock installed by ContextBuilder (src/context/mod.rs: self.clock.unwrap_or_else(|| Rc::new(StdClock))) — calls std::time directly instead of going through Boa's own crate::sys::time abstraction:

```
// src/context/time.rs
impl Clock for StdClock {
    fn now(&self) -> JsInstant {
        let now = std::time::SystemTime::now();             // <-- std, not sys::time
        let duration = now
            .duration_since(std::time::UNIX_EPOCH)          // <-- std, not sys::time
            .expect("System clock is before Unix epoch");
        JsInstant::new_unchecked(duration)
    }
}
```
Boa already has the right abstraction in src/sys/mod.rs, which maps time → web-time on wasm when js is enabled (and other modules like src/job.rs, src/vm/mod.rs, src/builtins/atomics/* correctly use crate::sys::time). StdClock just bypasses it. Since the Date builtin now reads time via context.clock().now() (src/builtins/date/mod.rs), Date inherits the panic. The js feature does pull in web-time/time/wasm-bindgen/getrandom, so getrandom and the deprecated HostHooks::utc_now (which uses time::OffsetDateTime::now_utc()) are wasm-safe — but the newer Clock/StdClock path was not updated to honor it.

To Reproduce
```
With boa_engine = { version = "0.21.1", features = ["js"] }, compiled to wasm32-unknown-unknown and run in a browser:

let mut ctx = boa_engine::Context::default();
ctx.eval(boa_engine::Source::from_bytes("Date.now()")).unwrap();
Equivalent JavaScript that triggers it:

Date.now();
// or
new Date();
```
Expected behavior

With feature = "js" on a wasm browser target, Date.now() should return the current time using web-time (the browser Date clock). Instead, the call panics (time not implemented on this platform) the first time a script touches Date.

Suggested fix — route StdClock through the existing abstraction (web_time re-exports SystemTime/UNIX_EPOCH, so it's a drop-in):

```
impl Clock for StdClock {
    fn now(&self) -> JsInstant {
        let now = crate::sys::time::SystemTime::now();
        let duration = now
            .duration_since(crate::sys::time::UNIX_EPOCH)
            .expect("System clock is before Unix epoch");
        JsInstant::new_unchecked(duration)
    }
}
```
Build environment (please complete the following information):

OS: Windows 11 Pro (10.0.26200)
Version: boa_engine 0.21.1
Target triple: wasm32-unknown-unknown (browser; non-emscripten/non-wasi)
Rustc version: rustc 1.96.0-nightly (bcf3d36c9 2026-03-19)
Additional context

Workaround we're using — install a custom Clock backed by web-time and build via the builder instead of Context::default():

```
use boa_engine::context::time::{Clock, JsInstant};

#[derive(Debug, Default, Clone, Copy)]
struct WebSafeClock;
impl Clock for WebSafeClock {
    fn now(&self) -> JsInstant {
        let d = web_time::SystemTime::now()
            .duration_since(web_time::UNIX_EPOCH)
            .unwrap_or_default();
        JsInstant::new(d.as_secs(), d.subsec_nanos())
    }
}

let ctx = boa_engine::Context::builder()
    .clock(std::rc::Rc::new(WebSafeClock))
    .build()
    .unwrap();
```
The doc comment on src/sys/fallback/mod.rs ("This could cause panics on platforms that don't support Instant::now()") already anticipates exactly this case — StdClock simply isn't going through sys::time. A grep for std::time:: outside src/sys/ would surface any other direct uses that should be sys::time. FixedClock is unaffected (it doesn't read the system clock).

mansiverma897993:
This Pull Request fixes/closes #5397.

It changes the following:
- Routes StdClock through crate::sys::time to prevent wasm panics.

### DIFF
--- a/core/engine/src/context/time.rs
+++ b/core/engine/src/context/time.rs
@@ -196,9 +196,9 @@ impl Clock for StdClock {
     }
 
     fn system_time_millis(&self) -> i64 {
-        let now = std::time::SystemTime::now();
+        let now = crate::sys::time::SystemTime::now();
         let duration = now
-            .duration_since(std::time::UNIX_EPOCH)
+            .duration_since(crate::sys::time::UNIX_EPOCH)
             .expect("System clock is before Unix epoch");
         duration.as_millis() as i64
     }


OBWANDO (approved):
This is a much safer path and should resolve the issue.

nekevss (approved):
Fine by me. Thanks for the contribution!

PR #5400, opened 2026-06-17
perf: remove GcRefCell from inline cache to avoid borrow checking overhead

### LINKED ISSUE #5399: [Perf] - can we remove usage of GcRefCell in the inline_cache?
**Describe the bug**
When running the "Richards" test, a high proportion of cache misses are coming from the inline cache. 
Image

This is because the IC goes through a GcRefCell borrow to get entries:
https://github.com/boa-dev/boa/blob/1af45b21c78187d145ced49502d7ab6415c8ce04/core/engine/src/vm/inline_cache/mod.rs#L34

When these are borrowed [here](https://github.com/boa-dev/boa/blob/1af45b21c78187d145ced49502d7ab6415c8ce04/core/engine/src/vm/inline_cache/mod.rs#L94), its nearly always a cache miss.

We have 2 options. 
1. As the hot path is [through the some branch](https://github.com/boa-dev/boa/blob/1af45b21c78187d145ced49502d7ab6415c8ce04/core/engine/src/vm/inline_cache/mod.rs#L100-L105) the mutable borrow isn't even needed, as thats only used when a weakShape has been collected, so most of these calls don't even need a mutable borrow, a shared one should be fine and offer less over head.
 
A quick fix could be something like this:
```rust
 pub(crate) fn get(&self, shape: &Shape) -> Option<(Shape, Slot)> {
      if self.megamorphic.get() { return None; }

      let entries = self.entries.borrow();   // shared borrow — no write to borrow flag
      let shape_addr = shape.to_addr_usize();

      for (i, entry) in entries.iter().enumerate() {
          if let Some(upgraded) = entry.shape.upgrade() {
              if upgraded.to_addr_usize() == shape_addr {
                  return Some((upgraded, entry.slot));
              }
          } else {
              // stale entry found — drop shared borrow, take exclusive, clean up
              drop(entries);
              let mut entries = self.entries.borrow_mut();
              entries.swap_remove(i);
              return None;  // shape didn't match anyway (it was stale)
          }
      }
      None
  }
```

2: Do we need to wrap in GcRefCell at all? CacheEntry only holds a weakShape, so probably doesn't require the wrapping in GcRefCell as by its very nature a weakShape wouldn't keep the value alive anyway. Could we just swap this to a `Cell` and have this still work?

mansiverma897993:
This Pull Request fixes/closes #5399.

It changes the following:
- Replaces `GcRefCell<ArrayVec<CacheEntry, PIC_CAPACITY>>` with `Cell<ArrayVec<CacheEntry, PIC_CAPACITY>>` in `InlineCache` structure.
- Uses `Cell::take()` and `Cell::set()` to access and modify the inline cache entries in a zero-overhead manner, avoiding all dynamic borrow-checking overhead on the hot path.
- Manually implements `Clone` and `Debug` for `InlineCache` because `Cell<T>` only auto-derives them when `T: Copy`, and `ArrayVec` is non-`Copy`.
- Replaces `.entries.borrow()` with a new test helper `.entries()` in `core/engine/src/vm/inline_cache/tests.rs` to allow unit testing without borrow checks.

### DIFF
--- a/core/engine/src/vm/inline_cache/mod.rs
+++ b/core/engine/src/vm/inline_cache/mod.rs
@@ -24,28 +24,32 @@ pub(crate) struct CacheEntry {
 }
 
 /// An inline cache entry for a property access.
+#[repr(C)]
 #[derive(Trace, Finalize)]
 pub(crate) struct InlineCache {
+    /// Whether this access site has seen too many shapes and should no longer be cached.
+    #[unsafe_ignore_trace]
+    pub(crate) megamorphic: Cell<bool>,
+
     /// The property that is accessed.
     pub(crate) name: JsString,
 
     /// Multiple cached shape-to-slot entries.
     pub(crate) entries: Cell<ArrayVec<CacheEntry, PIC_CAPACITY>>,
-
-    /// Whether this access site has seen too many shapes and should no longer be cached.
-    #[unsafe_ignore_trace]
-    pub(crate) megamorphic: Cell<bool>,
 }
 
 impl Clone for InlineCache {
     fn clone(&self) -> Self {
-        let entries = self.entries.take();
-        let cloned_entries = entries.clone();
-        self.entries.set(entries);
+        // SAFETY: `entries` is only ever accessed through `&self`/`&mut self`
+        // on this single-threaded cache, and cloning `CacheEntry` doesn't
+        // reenter this `Cell`, so it's safe to read through the raw pointer
+        // for the duration of this borrow without disturbing the cell's contents.
+        let cloned_entries = unsafe { (*self.entries.as_ptr()).clone() };
+
         Self {
+            megamorphic: self.megamorphic.clone(),
             name: self.name.clone(),
             entries: Cell::new(cloned_entries),
-            megamorphic: self.megamorphic.clone(),
         }
     }
 }
@@ -84,9 +88,9 @@ impl fmt::Display for InlineCache {
 impl InlineCache {
     pub(crate) fn new(name: JsString) -> Self {
         Self {
+            megamorphic: Cell::new(false),
             name,
             entries: Cell::new(ArrayVec::new()),
-            megamorphic: Cell::new(false),
         }
     }
 
--- a/core/engine/src/vm/inline_cache/mod.rs
+++ b/core/engine/src/vm/inline_cache/mod.rs
@@ -56,15 +56,15 @@ impl Clone for InlineCache {
 
 impl fmt::Debug for InlineCache {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let entries = self.entries.take();
-        let result = f
-            .debug_struct("InlineCache")
+        // SAFETY: `entries` is only ever accessed through `&self`/`&mut self`
+        // on this single-threaded cache, and printing doesn't reenter this `Cell`,
+        // so it's safe to read through the raw pointer.
+        let entries = unsafe { &*self.entries.as_ptr() };
+        f.debug_struct("InlineCache")
             .field("name", &self.name)
-            .field("entries", &entries)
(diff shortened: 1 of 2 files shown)


jasonwilliams:
Thanks for this @mansiverma897993, you need to run rust fmt

jasonwilliams:
For the clone it feels inefficient to take the value out, put a default one in then add the value back on a hotpath. We may need to get a pointer into the cell here as its safe (we are taking and resetting in the same operation). Something like this:

```rust
    fn clone(&self) -> Self {
        // SAFETY: `entries` is only ever accessed through `&self`/`&mut self`
        // on this single-threaded cache, and cloning `CacheEntry` doesn't
        // reenter this `Cell`, so it's safe to read through the raw pointer
        // for the duration of this borrow without disturbing the cell's contents.
        let cloned_entries = unsafe { (*self.entries.as_ptr()).clone() };

        Self {
            name: self.name.clone(),
            entries: Cell::new(cloned_entries),
            megamorphic: self.megamorphic.clone(),
        }
    }
 ```
 
 Also, while you're here, can you move the [megmorphic](https://github.com/mansiverma897993/boa/blob/d2d4cd231348adfdedf10d8744909dc687e1399a/core/engine/src/vm/inline_cache/mod.rs#L37) to be the first item in the struct? The `entries` value is so big it pushes the bool out of the cacheline so its a cache miss every time it's read, it may need the `#[repr(C)]` tag above

**Comment by mansiverma897993** [comment]
@jasonwilliams I have sucessfully update the PR as per your accordance !! look at once

**Comment by jasonwilliams** [comment]
@mansiverma897993 thanks, I'm not sure when I can take a look again. In the meantime are you able to compare the benchmarks between main and this? We have a data repo here: https://github.com/boa-dev/data

You can check that out in an adjacent folder and run boa against the bench/bench-v8/combined.js file

When I checked on cachegrind it looked like there were still a high amount of cache misses happening somewhere in that function, maybe @HalidOdat is better at hunting these things down.

File: core/engine/src/vm/inline_cache/mod.rs [hunk 1]
```diff
@@ -91,7 +133,7 @@ impl InlineCache {
             return None;
         }
 
-        let mut entries = self.entries.borrow_mut();
+        let mut entries = self.entries.take();
```
zhuzhu81998: `take` on `Cell` performs a copy i think. perhaps its not trivial to copy the entire cache twice (once here once below for setting) everytime you try to access or set it?
Given how often this is used.

mansiverma897993: "Thanks for pointing this out! You are absolutely right, Cell::take() does copy the entire ArrayVec which adds overhead on the hot path.

To avoid this copy while still avoiding GcRefCell's borrow checking, we can use Cell::as_ptr() to mutate the entries in-place using a raw pointer:

rust
let entries = unsafe { &mut *self.entries.as_ptr() };
Since this is single-threaded and the reference doesn't escape the function, this should be safe and provide zero-copy access. I'll update the PR with this change!"  I am updating PR accordingly I am working on it !!

mansiverma897993: @zhuzhu81998 @jasonwilliams I have updated PR accordingly now look at once!!

zhuzhu81998:
Do you have benchmark results for this newest version?

And also, if you are using raw pointers to avoid the copying etc., is the use of `Cell` actually necessary?

jedel1043:
Waiting on author for benchmark results. It doesn't make sense to complicate the code if we won't have perf wins
(remember to isolate the change that moved `megamorphic` to the start of the struct, because that might be what improves performance and not necessarily the removal of `GcRefCell`)

jedel1043:
Closing since the author doesn't seem to have benchmark results justifying the change.

PR #5401, opened 2026-06-17
fix: support HTML close comment on the first line of script

mansiverma897993:
This Pull Request fixes a bug where --> (HTML close comment) was unconditionally rejected as a syntax error.

Under Annex B, --> is a valid comment when it appears on the first line of a script. This change allows HTML close comments on the first line of scripts.


jedel1043:
The fix introduced here is also breaking some related tests

jedel1043:
Closing this since the PR hasn't addressed the broken trests. Feel free to reopen whenever that gets fixed

PR #5403, opened 2026-06-19
fix: allow deriving both TryFromJs and TryIntoJs with custom converters

### LINKED ISSUE #5361: Can't use `#[boa(from_js_with = "...")]` or `#[boa(into_js_with = "...")]` when deriving both `TryFromJs` and `TryIntoJs`
**Describe the bug**

If I am deriving both `TryFromJs` and `TryIntoJs`, I can't use `#[boa(from_js_with = "...")]` or `#[boa(into_js_with = "...")]` as each macro complains about the other's annotation.

**To Reproduce**

This example code:

```rust
#[derive(TryFromJs, TryIntoJs)]
struct Blah {
    #[boa(into_js_with = "my_custom_converter_to_js")]
    #[boa(from_js_with = "my_custom_converter_from_js")]
    x: i32,
}
```

Results in two error messages:

```
invalid syntax in the #[boa()] attribute. Note that this attribute only accepts the following syntax: #[boa(from_js_with = "fully::qualified::path")]
```

```
invalid syntax in the #[boa()] attribute. Note that this attribute only accepts the following syntax:
#[boa(into_js_with = "fully::qualified::path")]
#[boa(rename = "jsPropertyName")]
#[boa(skip)] 
```

**Expected behavior**
I expected to be able to specify one or more custom conversion functions when deriving both `TryIntoJs` and `TryFromJs` on the same type.

**Build environment:**

- OS: MacOS
- Version: 15.7.3
- Rustc version: rustc 1.92.0 (ded5c06cf 2025-12-08)

mansiverma897993:
Closes #5361

### DIFF
--- a/core/macros/src/lib.rs
+++ b/core/macros/src/lib.rs
@@ -524,10 +524,10 @@ fn generate_conversion(
 
         let mut from_js_with = None;
         let mut field_name = rename.rename(format!("{name}"));
-        if let Some(attr) = field
+        for attr in field
             .attrs
             .into_iter()
-            .find(|attr| attr.path().is_ident("boa"))
+            .filter(|attr| attr.path().is_ident("boa"))
         {
             attr.parse_nested_meta(|meta| {
                 if meta.path.is_ident("from_js_with") {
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Seems like the best thing we can do right now. We might want to change syntax a bit to be something like `boa_try_from_js(with = "function")` to avoid code duplication, but that's just a future enhancement IMO. Thanks!

mansiverma897993:
@jedel1043 
 Thanks for the suggestion! I've updated the syntax to use `#[boa_try_from_js(with = "...")]` and `#[boa_try_into_js(with = "...")]` helper attributes respectively. This eliminates the need to parse and ignore each other's custom attributes, keeping the two macro derivations decoupled.

PR #5408, opened 2026-06-20
Fix unsoundness in JsValueStore by replacing unsafe raw pointer mutation of Arc with safe OnceLock

### LINKED ISSUE #5392: Dubious use of unsafe in runtime::store::JsValueStore.replace()
This is another extraction from #5370. This is the potential problem: "C2. Unsoundness in JsValueStore::replace — mutates data through a shared Arc" as written in that issue.

This is the code in question: [core/runtime/src/store/mod.rs](https://github.com/boa-dev/boa/blob/aa72ad2671b8e41794162d2e828a8d3db54a1f71/core/runtime/src/store/mod.rs#L171):
```
unsafe fn replace(&mut self, other: ValueStoreInner) {
    let ptr = Arc::as_ptr(&self.0).cast_mut();

    assert!(!ptr.is_null());
    unsafe {
        assert!(
            matches!(*ptr, ValueStoreInner::Empty),
            "ValueStoreInner must be empty."
        );

        *ptr = other;
    }
}
```

It has the following comment justifying the unsafe block:
```
 /// # SAFETY
 /// This should only be done if the inner content is [`ValueStoreInner::Empty`]
 /// and only by the creator of the current [`JsValueStore`]. We enforce the first
 /// rule at runtime (and will panic), and the second rule by requiring a mutable
 /// reference. This is still unsafe and relies on unsafe pointer access.
```

This is the essence of Claude's complaint  (see #5370 for complete details):
> Mutating data behind a shared Arc(which is &T, not &mut T) is undefined behaviour even when single-threaded; the assert!(matches!(*ptr, ValueStoreInner::Empty)) does not rescue soundness because the other live clones currently observe Empty but their later Deref operations will read Object/Array/Map through a non-UnsafeCell shared reference.

Note the crucial point that even in a single-threaded context, this is still undefined behavior. I [asked](https://chatgpt.com/share/6a221ff7-8ae4-83ea-8615-b49466c3fc80) ChatGPT about this audit, and it says, in part:
> The concern raised in that audit is generally valid. The key issue is not threading; it's Rust's aliasing model.

and:
> ...writing through a raw pointer obtained from Arc::as_ptr() is mutating memory that is currently aliased by shared references. That is exactly the kind of operation Rust declares UB.

I don't have code to demonstrate this fails, and it is tested indirectly via [message/tests.rs](https://github.com/boa-dev/boa/blob/aa72ad2671b8e41794162d2e828a8d3db54a1f71/core/runtime/src/message/tests.rs), but that doesn't mean the code is valid or won't fail in all scenarios (optimizations for a particular architecture, future optimizations, etc). Ideally, the code in question would be replaced with safe code.

In my ChatGPT conversation, I asked for alternatives:
> The replace() function is trying to create a store with potential recursive data, and then never mutate it after. What would be the recommended way to achieve this safely without unnecessary overhead?

and it replied, in part:
> This is a classic problem: you need to construct a graph that may contain cycles, but you want the final structure to be immutable.

>The usual solution is not to mutate through a shared Arc<T>, but instead to introduce a temporary construction phase where mutation is permitted, and then "freeze" the structure.

with the following options:

1. Build with Arc::new_cyclic
2. Two-phase construction with OnceLock
3. Two-phase construction with UnsafeCell (lowest overhead)
4. Arena + freeze (often the cleanest)

with the concluding recommendation:
> OnceLock<ValueStoreInner> is almost certainly the closest drop-in replacement that preserves the existing algorithm while restoring soundness. The runtime cost is usually negligible compared to deserializing recursive structures.

But do read the chat link for all the details with pros and cons.

If unsafe is absolutely needed for performance, then it should be the appropriate unsafe, though I would caution against unsafe code in general WITHOUT A BENCHMARK TO SHOW IT'S NEEDED. Sorry for the all-caps, but unsafe is a nasty business, and we really should be avoiding premature optimization. Related issue: #4181.

mansiverma897993:
This Pull Request fixes/closes #5392.

It changes the following:

- Fix unsoundness in JsValueStore by replacing unsafe raw pointer mutation of Arc with safe OnceLock

### DIFF
--- a/core/engine/src/module/loader/mod.rs
+++ b/core/engine/src/module/loader/mod.rs
@@ -205,15 +205,15 @@ pub trait ModuleLoader: Any {
 pub struct IdleModuleLoader;
 
 impl ModuleLoader for IdleModuleLoader {
-    async fn load_imported_module(
+    fn load_imported_module(
         self: Rc<Self>,
         _referrer: Referrer,
         _request: ModuleRequest,
         _context: &RefCell<&mut Context>,
-    ) -> JsResult<Module> {
-        Err(JsNativeError::typ()
+    ) -> impl Future<Output = JsResult<Module>> {
+        std::future::ready(Err(JsNativeError::typ()
             .with_message("module resolution is disabled for this context")
-            .into())
+            .into()))
     }
 }
 
--- a/core/engine/src/vm/mod.rs
+++ b/core/engine/src/vm/mod.rs
@@ -756,7 +756,7 @@ impl Context {
         println!(
             "{:<TIME_COLUMN_WIDTH$} {:<OPCODE_COLUMN_WIDTH$} {operands:<OPERAND_COLUMN_WIDTH$} {stack}",
             format!("{}μs", duration.as_micros()),
-            format!("{}", opcode.as_str()),
+            opcode.as_str(),
             TIME_COLUMN_WIDTH = Self::TIME_COLUMN_WIDTH,
             OPCODE_COLUMN_WIDTH = Self::OPCODE_COLUMN_WIDTH,
             OPERAND_COLUMN_WIDTH = Self::OPERAND_COLUMN_WIDTH,
--- a/core/engine/tests/module.rs
+++ b/core/engine/tests/module.rs
@@ -11,16 +11,18 @@ use boa_engine::{Context, JsResult, JsString, Module, Source, js_string};
 fn test_json_module_from_str() {
     struct TestModuleLoader(JsString);
     impl ModuleLoader for TestModuleLoader {
-        async fn load_imported_module(
+        fn load_imported_module(
             self: Rc<Self>,
             _referrer: Referrer,
             request: boa_engine::module::ModuleRequest,
             context: &RefCell<&mut Context>,
-        ) -> JsResult<Module> {
-            assert_eq!(request.specifier().to_std_string_escaped(), "basic");
-            let src = self.0.clone();
+        ) -> impl Future<Output = JsResult<Module>> {
+            std::future::ready({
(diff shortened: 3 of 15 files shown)


hansl (changes requested):
Why is the macro change part of this PR? Feels like it should be its own PR, can you split them?

mansiverma897993:
Thanks for the review, @hansl.
I've split the PR as requested the macro change is no longer part of this PR. I removed the fix: allow deriving both TryFromJs and TryIntoJs with custom converters commit (that change was already merged separately in #5403), leaving only the JsValueStore OnceLock fix.
I also rebased onto the latest main, which resolves the conflicts from the store module move (core/runtime/src/store → core/wintertc/src/store).
All checks pass locally: boa_wintertc and boa_runtime tests, clippy with -D warnings, and cargo fmt --check.

mansiverma897993:
> Thanks for the review, @hansl. I've split the PR as requested the macro change is no longer part of this PR. I removed the fix: allow deriving both TryFromJs and TryIntoJs with custom converters commit (that change was already merged separately in #5403), leaving only the JsValueStore OnceLock fix. I also rebased onto the latest main, which resolves the conflicts from the store module move (core/runtime/src/store → core/wintertc/src/store). All checks pass locally: boa_wintertc and boa_runtime tests, clippy with -D warnings, and cargo fmt --check.

@jedel1043 can you plz review it too and let me know further changes !!

jedel1043:
This got a bit stale against main, and it also has a lot of unnecessary changes. Gonna close this PR, but feel free to reopen after cleaning up the unnecessary changes and rebasing against main.

PR #5409, opened 2026-06-20
Don't double-include line and column information in SyntaxError (fixes #5357)

### LINKED ISSUE #5357: Syntax error mentions location twice
```console
$ cat file.js
foo(); @
$ ./target/release/boa file.js
Error:
   0: SyntaxError: unexpected '@' at line 1, column 8 at line 1, col 8
```

MatrixFrog:
This Pull Request fixes #5357

The SyntaxError includes the location information already so it doesn't need to add the line and column number in the error message as well.

### DIFF
--- a/core/parser/src/lexer/mod.rs
+++ b/core/parser/src/lexer/mod.rs
@@ -342,11 +342,7 @@ impl<R> Lexer<R> {
                     NumberLiteral::new(next_ch as u8).lex(&mut self.cursor, start, interner)
                 }
                 _ => {
-                    let details = format!(
-                        "unexpected '{c}' at line {}, column {}",
-                        start.line_number(),
-                        start.column_number()
-                    );
+                    let details = format!("unexpected '{c}'");
                     Err(Error::syntax(details, start.position()))
                 }
             }?;
--- a/core/engine/src/tests/mod.rs
+++ b/core/engine/src/tests/mod.rs
@@ -480,6 +480,15 @@ fn strict_mode_reserved_name() {
 }
 
 #[test]
+fn issue5357() {
+    run_test_actions([TestAction::assert_native_error(
+        "foo(); @",
+        JsNativeErrorKind::Syntax,
+        "unexpected '@' at line 1, col 8",
+    )]);
+}
+
+#[test]
 fn empty_statement() {
     run_test_actions([TestAction::assert_eq(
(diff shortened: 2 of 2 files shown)


core/parser/src/lexer/mod.rs
nekevss, line 354: Does this potentially cause the same issue that is mentioned in the issue? Can we test this branch as well?

nekevss (commented):
Thanks for working on this! Just had a quick question 😄

core/parser/src/lexer/mod.rs
MatrixFrog: Probably. I'm not sure how to reach this branch though -- from what I can tell, even though next_ch is a u32 not a char, it will always be a valid char because the lexer ultimately uses a UTF8Input which will only yield valid UTF8 chars. But I may have missed something.

MatrixFrog: Ok, I dug in a little more, what seems to be happening is that ReadChar's next_char method is documented as "Returns an error if the next input in the input is not a valid unicode code point." but the implementations (both UTF8Input and UTF16Input) from what I can tell don't follow that and instead if the input is invalid they could yield u32s which are not valid chars. It seems to me that it would be nice to switch the implementations to match the docs. Then, next_char could yield chars instead of u32s, and a lot of the lexer/parser code could become much more readable. For instance instead of code like `(0x30..=0x39/* 0..=9 */).contains(c)` needing a comment to explain what char 0x30 corresponds to, it could just look like `('0'..='9').contains(c)`.

I started down this path a little bit but I don't know if that's the direction the project wants to go. It would also work to keep it as is but then it would be good to assemble some tests that feed in byte sequences which are not valid UTF8/UTF16, to make sure those code paths do what's expected.

In any case that would end up being a bigger change, happy to add it to this PR or start a separate one, whatever is preferred.

MatrixFrog: Added a small test to show what I mean, using cargo-fuzz to help produce an invalid input: https://github.com/MatrixFrog/boa/commits/invalid-utf8/

If the input is invalid, UTF8Input produces an Ok(Some(u32)), not an Err(), and then to determine that it's invalid, you have to try to convert the u32 to a char, which fails in this case.

nekevss (approved):

PR #5413, opened 2026-06-23
perf: add fast path to OrdinaryGet

Serdnad:
Adds a fast path to `OrdinaryGet` that reads a plain data property's value directly from object storage, skipping `PropertyDescriptor` construction.

Benchmarks on local machine:

| Benchmark | Δ |
|---|---|
| **strings/slice** | **−21.4%** |
| **intl/numberformat-different-options** | **−14.2%** |
| **strings/split** | **−12.4%** |
| **strings/concat** | **−12.3%** |
| **intl/segmenter-segment** | **−12.2%** |
| **strings/replace** | **−11.9%** |
| **properties/access** | **−10.4%** |
| prototypes/chain | −9.7% |
| intl/listformat-format | −8.8% |
| intl/datetimeformat-format | −7.7% |
| v8-benches/splay | −7.0% |
| v8-benches/regexp | −5.9% |
| basic/nested-loop | −4.8% |
| intl/collator-construction | −4.5% |
| intl/pluralrules-construction | −4.0% |
| intl/numberformat-construction | −3.9% |
| intl/segmenter-construction | −3.8% |
| v8-benches/earley-boyer | −3.7% |
| intl/listformat-construction | −3.4% |
| v8-benches/raytrace | −3.2% |
| json/stringify_circular | −2.9% |
| closures/invoke | −1.6% |
| closures/create | −1.3% |
| intl/datetimeformat-with-options | −0.9% |
| v8-benches/navier-stokes | −0.3% |
| intl/pluralrules-select | −0.1% |
| basic/call-loop | −0.0% |
| json/stringify_deep | +0.0% |
| v8-benches/deltablue | +0.1% |
| v8-benches/crypto | +0.1% |
| v8-benches/richards | +1.5% |
| **intl/collator-compare** | **+3.1%** |
| **intl/datetimeformat_resolved_options** | **+3.2%** |

### DIFF
--- a/core/engine/src/object/internal_methods/mod.rs
+++ b/core/engine/src/object/internal_methods/mod.rs
@@ -9,6 +9,7 @@ use std::ops::{Deref, DerefMut};
 
 use super::{
     JsPrototype, PROTOTYPE,
+    property_map::OwnDataLookup,
     shape::slot::{Slot, SlotAttributes},
 };
 use crate::{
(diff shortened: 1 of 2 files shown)


core/engine/src/object/internal_methods/mod.rs
```diff
@@ -709,6 +710,32 @@ pub(crate) fn ordinary_get(
...
+                    context.slot().set_not_cacheable_if_already_prototype();
+                    context.slot().attributes |= SlotAttributes::PROTOTYPE;
+                    return parent.__get__(key, receiver, context);
+                }
+                return Ok(JsValue::undefined());
+            }
+            // Accessor: fall through to invoke the getter.
+            OwnDataLookup::Accessor => {}
+        }
+    }
```
jedel1043, line 737: Wait, why does this improve performance, if it also needs to construct an `OwnDataLookup` struct? I'm not criticizing the approach, I want to understand this because we might have the same perf benefits of this with #5284, which doesn't need to specialize property lookups.

Serdnad: Hey @jedel1043 , thanks for taking a look at my PR, and apologies for the slow response.

I took a look at #5284 , and yes, the premise is effectively the same; the proposed enum `OwnDataLookup` is slightly smaller and cheaper to use than `PropertyDescriptor`, and since this is a hot path, it's enough to have an impact.

That said, I also took the liberty of benchmarking the other PR beside mine, and while they're very similar (similar improvements/regressions on the same benchmarks), which is to be expected, there are one or two v8 bench outliers where the other PR significantly beats mine. With that in mind, would be fully in support of closing my PR in favor of the other if you prefer - I hadn't seen it before opening mine.

jedel1043: Did you also test your changes on top of #5284? Because it might be cool if they composed together to give a bigger perf increase. If the difference is essentially nil, we can close this.

jedel1043:
Closing this since it seems like #5284 already covers the same optimization. However, feel free to reopen if you find that this improves performance on top of that.

PR #5414, opened 2026-06-26
expose fractionalSecondDigits as a property of resolvedOptions()

camilojd:
Hi boa-devs!

I have a small contribution for the "ECMA-402" effort. Please let me know if this looks good to you.
It changes the following:

- Exposes `fractionalSecondDigits` as a property inside the object returned by `resolvedOptions()`

### DIFF
--- a/core/engine/src/builtins/intl/date_time_format/mod.rs
+++ b/core/engine/src/builtins/intl/date_time_format/mod.rs
@@ -17,7 +17,9 @@ use crate::{
         },
         intl::{
             Service,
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
Looks good! Though you might need to run clippy on this, but preemptively approving.

camilojd:
@jedel1043 Thanks for the review! I rebased, as I saw that the clippy failures came from outside the change, and they seem to be addressed now

jedel1043 (approved):
Thanks! Looks great!

PR #5418, opened 2026-06-28
feat(wintertc): migrate atob/btoa from boa_runtime

KaustubhOG:
This Pull Request is part of the TC55 migration tracked in #4988, and builds on the skeleton crate from #5105.

**It changes the following:**

- **Moves the `base64` module** (`mod.rs` and `tests.rs`) into `boa_wintertc`. The implementation is **unchanged**; the only difference from the `boa_runtime` version is a short TC55 status note added to the module documentation.
- **Deletes `core/runtime/src/base64/`** and **re-exports the module from `boa_runtime`**: `boa_runtime::base64` is now `#[doc(inline)] pub use boa_wintertc::base64;`, so the existing public path keeps resolving and the public API is unchanged.
- **Delegates the extension**: `Base64Extension` now calls `boa_wintertc::base64::register`. It stays in the default extension set, so there is no behaviour change for existing users.
- **Drops the `base64` crate dependency** from `boa_runtime`, as the moved module was its only user, and adds `boa_wintertc` to `[workspace.dependencies]` and to `boa_runtime`.
- **Ports the test harness** (`TestAction` and `run_test_actions_with`) into `boa_wintertc` so the migrated unit tests run unchanged. This harness will be reused by later migrations.
- **Adds dependencies** to `core/wintertc/Cargo.toml`: the `base64` dependency, plus `indoc`, `textwrap`, and `futures-lite` as dev dependencies.

**Line counts (before and after, in both locations):**

| File | `boa_runtime` (before) | `boa_wintertc` before | `boa_wintertc` after |
| --- | --- | --- | --- |
| `base64/mod.rs` | 82 (deleted) | 23 (stub) | 86 |
| `base64/tests.rs` | 154 (deleted) | 0 (did not exist) | 154 |

The 4 extra lines in `mod.rs` are the TC55 status doc note; the code itself is identical. `tests.rs` is byte for byte the same.

**Approach:** following review, `boa_runtime` no longer keeps its own `base64` module. The module lives in `boa_wintertc` only, and `boa_runtime` re-exports it, so there is **no behaviour change** for existing users and the public API stays the same. The dependency direction is one way: `boa_runtime` -> `boa_wintertc` -> `boa_engine`. The remaining migration PRs will follow this same pattern.

**Testing:**

- `cargo test -p boa_wintertc`: all six base64 unit tests pass, plus the crate doctest.
- `cargo test -p boa_runtime`: passes, with base64 now exercised through the re-export.
- `cargo clippy -p boa_wintertc --all-features --all-targets` and `cargo clippy -p boa_wintertc --no-default-features`: clean, no warnings.
- `cargo fmt --all -- --check`: clean.
- Verified `atob` and `btoa` end to end through `boa_cli`, including forgiving decode, whitespace handling, and both error paths.

**Out of scope:**

- **WPT** is unchanged here: the runner targets `boa_runtime`, which now reaches this implementation through the re-export.
- **Test262** does not apply, as `atob` and `btoa` are Web APIs rather than ECMAScript.


jedel1043 (changes requested):

This seems to be a direct clone from `boa_runtime`, so to avoid code churn I think we should prioritize migrating `boa_runtime` to use `boa_wintertc`'s atob/btoa, and we should follow the same for the rest of the PRs. Can you integrate that change into this PR?

jedel1043:
btw our pipeline is slightly blocked by #5427, so just wait a bit until we merge the PR

KaustubhOG:
> btw our pipeline is slightly blocked by #5427, so just wait a bit until we merge the PR

yes

KaustubhOG:
> btw our pipeline is slightly blocked by #5427, so just wait a bit until we merge the PR

yeah just saw it merrged can you review it again

jedel1043 (approved):
Looks good!

PR #5419, opened 2026-07-01
feat(wintertc): migrate structuredClone, queueMicrotask and timers from boa_runtime

KaustubhOG:
**Summary**
Continues the boa_runtime -> boa_wintertc migration started with atob/btoa (#5418),moving three more modules with the same move + re-export pattern. Code lives only inboa_wintertc; boa_runtime re-exports it (public paths unchanged) and the *Extension structs delegate.

**Changes**
- **structuredClone**: `clone` -> `boa_wintertc::clone`. Its `store` dependency  (`JsValueStore`) also moves to `boa_wintertc::store`, since boa_wintertc cannot depend  on boa_runtime. `store` stays public and is re-exported, so `boa_runtime::message` keeps compiling. `rustc-hash` moves with it.
- **queueMicrotask**: `microtask` -> `boa_wintertc::microtask`.
- **timers**: `interval` -> `boa_wintertc::timers` (renamed to its TC55 category). Re-exported as `pub use boa_wintertc::timers as interval`, so `boa_runtime::interval` (incl. `clear_all`, used by the WPT harness) is unchanged.

**Notes / deviations**
- `store` is a non-TC55 helper that is now public in boa_wintertc. Net public surfac across both crates is unchanged (it was already public in boa_runtime); happy to make it  `pub(crate)` once `message` also migrates.
- The queueMicrotask unit test now records order in a global array instead of `console`,since `console` is still a stub in boa_wintertc.


KaustubhOG:
hey @jedel1043  Batched three trivial move + re-export migrations here (structuredClone,
queueMicrotask, timers).

core/runtime/src/lib.rs
```diff
@@ -119,16 +119,43 @@ pub use console::{Console, ConsoleState, DefaultLogger, Logger, NullLogger};
 
 #[cfg(feature = "fetch")]
 pub mod abort;
-pub mod clone;
+
+/// `structuredClone`, re-exported from [`boa_wintertc`].
+///
+/// This API is part of the `WinterTC` (TC55) Minimum Common Web API and is implemented in
+/// `boa_wintertc`. It is re-exported here so `boa_runtime` users keep a single import path.
```
jedel1043: All of these exports repeat basically the same thing: part of WinterTC and reexported here. I suggest unifying all these messages into a single doc description on the module root, then just in-lining the docs of each component here.

jedel1043 (changes requested):

core/runtime/src/lib.rs
KaustubhOG: yeah make sense it was unwanted noisee , let me remove it

KaustubhOG: done

jedel1043 (approved):
Looks great!

PR #5426, opened 2026-07-10
[oscars-integration] feat: setup initial oscars integration

shruti2522:
this is an early draft with the initial setup to integrate oscars in boa. We are currently using `null_collector_branded` as the baseline for integration

It changes the following:

- cleaned up `boa/core/gc/src/lib.rs` by moving the oldlogic into `boa_allocator.rs`
- added the `oscars_backend` feature flag to toggle between the current GC and the new oscars gc

(Note: building with the `oscars_backend` feature will currently fail since the engine still needs the new `'gc` lifetimes to be threaded. I will be adding those updates in the next few commits)


Cargo.toml
```diff
@@ -266,3 +266,7 @@ style = { level = "warn", priority = -1 }
 complexity = { level = "warn", priority = -1 }
 perf = { level = "warn", priority = -1 }
 pedantic = { level = "warn", priority = -1 }
+
+[patch."https://github.com/boa-dev/oscars.git"]
+oscars = { path = "../oscars/oscars" }
+oscars_derive = { path = "../oscars/oscars_derive" }
```
jedel1043: Might want to remove this for the upstream code, since it would make it a bit hard to test things using the CI (if we need to)

I think you can still configure this using the `$HOME/.cargo/config.toml` file

nekevss (approved):
Fine by me, let's go with it

PR #5431, opened 2026-07-16
perf: use mimalloc as the global allocator on macOS

hansl:
The CLI's default `fast-allocator` feature only installed a fast allocator on x86_64 Linux (jemalloc) and Windows (mimalloc); macOS silently fell back to the system `libmalloc`. Extend the existing mimalloc arm to cover macOS in both the CLI and the benchmark harness (`mimalloc-safe` already supports Apple Silicon).

Keeping the benchmarks on a fast allocator on macOS also stops local runs from measuring `libmalloc` overhead instead of engine work, so results track CI (Linux/jemalloc) far more closely. Measured locally, allocation-heavy workloads improve by up to ~8.6% (closures/create) versus libmalloc, while compute-bound benchmarks are unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -339,6 +339,7 @@ dependencies = [
  "boa_runtime",
  "criterion",
  "jemallocator",
+ "mimalloc-safe",
  "walkdir",
 ]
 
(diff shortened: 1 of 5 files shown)


jedel1043 (approved):
LGTM!

PR #5432, opened 2026-07-17
perf: fuse relational if/ternary conditions into a compare-branch 

hansl:
`if (a < b)` and `cond ? x : y` previously compiled the relational operator into a boolean in a scratch register (e.g. `LessThan`) and then tested it with `JumpIfFalse` — two dispatches plus a throwaway register. Route them through the existing `compile_condition_and_branch` helper (already used by `for`/`while`/`do-while` loops) so a relational condition emits a single fused `JumpIfNot{Less,Greater}Than[OrEqual]` opcode instead.

Also `flatten()` the condition first, so parenthesized forms like `(a < b) ? x : y` and `while ((i < n))` reach the fused path too. Non-relational conditions keep the previous `compile_expr` + `JumpIfFalse` fallback. Reuses opcodes already proven for loops; all boa_engine tests pass.

### DIFF
--- a/core/engine/src/bytecompiler/expression/mod.rs
+++ b/core/engine/src/bytecompiler/expression/mod.rs
@@ -52,9 +52,8 @@ impl ByteCompiler<'_> {
     }
 
     fn compile_conditional(&mut self, op: &Conditional, dst: &Register) {
-        self.compile_expr(op.condition(), dst);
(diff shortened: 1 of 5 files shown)


jedel1043 (approved):
Nice optimization!

PR #5437, opened 2026-07-19
feat(engine): add WeakJsObject weak reference type

### LINKED ISSUE #5420: No public weak reference API for `JsObject`
## Summary

There is currently no way for an embedder to hold a weak reference to a `JsObject`.
The GC types `WeakGc` and `Ephemeron` are public, but they require `&Gc<T>` to construct,
and the only path from `JsObject` to its inner `Gc<VTableObject<T>>` — the `inner()` method — is `pub(crate)`.
This leaves embedders without any mechanism to observe object liveness without keeping strong references that prevent collection.

## The problem

Consider an embedder that maintains a Rust-side cache keyed by JS objects.
The cache should not keep JS objects alive on its own. When a JS object becomes unreachable and is collected by the GC, the corresponding cache entry should become invalid.

This requires a weak reference to `JsObject`: the ability to check whether the object is still alive, and to upgrade back to a strong `JsObject` when it is.

### What is available

- `WeakGc<T>` and `Ephemeron<K, V>` are `pub` in `boa_gc`.
- `WeakGc::upgrade()` returns `Option<Gc<T>>` — exactly the right semantics.
- `WeakGc::new()` requires `&Gc<T>`.

### What is blocked

- `JsObject::inner()` is `pub(crate)`:

```rust
// core/engine/src/object/jsobject.rs:1041
pub(crate) fn inner(&self) -> &Gc<VTableObject<T>> {
    &self.inner
}
```

- `JsObject::from_inner()` is also `pub(crate)`:

```rust
// core/engine/src/object/jsobject.rs:1045
pub(crate) fn from_inner(inner: Gc<VTableObject<T>>) -> Self {
    Self { inner }
}
```

- `VTableObject` itself is `pub(crate)`.

So an external crate cannot write the type `Gc<VTableObject<...>>`, cannot call `inner()` to get it, and cannot call `from_inner()` to convert back. The entire `WeakGc` / `Ephemeron` API is unreachable for `JsObject`.

### Where weak references are used today

`FinalizationRegistry` and `WeakRef` both use weak references inside `boa_engine`, where `pub(crate)` is accessible:

```rust
// finalization_registry/mod.rs
WeakGc::new(obj.inner())        // accessible within boa_engine
Ephemeron::new(target_obj.inner(), ...)  // accessible within boa_engine
```

This means the existing weak reference infrastructure is exercised internally, but is not currently reachable from external crates.

## Why this matters

Without a weak reference API, embedders who need Rust-side structures keyed by JS objects have two options, neither of which is sound:

1. **Hold strong `JsObject` references.** This prevents GC collection, leading to memory leaks.

2. **Store raw pointers via `transmute` or similar.** This breaks GC safety: the pointer may dangle after collection, `ref_count` becomes inconsistent, and casting back to `JsObject` can trigger use-after-free or `ref_count` underflow.

## Questions

- Is the absence of a public weak reference API for `JsObject` an intentional design decision, or something that hasn't been needed yet?
- If it is intentional, what approach would you recommend for embedders who need to observe object liveness without preventing GC collection?
- Is there any thoughts on exposing this kind of functionality in the future, or should embedders plan to maintain a fork for now?

GordonYuanyc:
This Pull Request closes #5420.

It adds a public `WeakJsObject` type to `boa_engine`, giving embedders a weak reference to a `JsObject` that does not keep the object alive across garbage collections. This is the object-level counterpart of `boa_gc::WeakGc`.

It changes the following:

- Adds `WeakJsObject<T>` in `core/engine/src/object/weak.rs` with `new`, `upgrade` and `is_upgradable`, mirroring the surface of `WeakGc`, plus `Clone`, `Debug` and `From<&JsObject>`. The mechanism already existed internally (the `WeakRef` builtin builds a `WeakGc<VTableObject<T>>` via `WeakGc::new(target.inner())`), but `JsObject::inner()` is `pub(crate)`, so embedders had no public way to construct one.
- Deliberately does not implement `PartialEq`/`Eq`/`Hash`. Forwarding them to `WeakGc` compares and hashes by the live referent, so once the object is collected two references to it stop comparing equal and change their hash, which breaks `Eq` reflexivity and the `Hash`/`Eq` invariant for a collected key. These can be added later without a breaking change if a collection-stable identity is wanted.
- Adds unit tests covering upgrade while live, upgrade returning `None` after collection, an upgraded handle keeping the referent alive across a later collection, multiple weak references dropping together, and a typed (non-erased) object.


core/engine/src/object/weak.rs
```diff
@@ -0,0 +1,194 @@
...
+// a breaking change, if a collection-stable identity is designed.
+
+// `VTableObject` deliberately does not implement `Debug` to avoid recursing into the object graph
+// (which could overflow the stack), so we cannot derive `Debug` here. We provide a minimal,
+// non-recursive implementation instead.
+impl<T: NativeObject> Debug for WeakJsObject<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("WeakJsObject").finish_non_exhaustive()
+    }
+}
```
jedel1043: This was already solved by the `Debug` implementation on `JsObject`, which uses a recursion limiter

https://github.com/boa-dev/boa/blob/6ef5370c1cff770c51f3e4f8600590f11d66eeaf/core/engine/src/object/jsobject.rs#L1290-L1325

You should use that instead.

core/engine/src/object/weak.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    #[test]
+    fn built_from_reference_via_conversion() {
+        let object = JsObject::with_null_proto();
+        let weak: WeakJsObject = (&object).into();
+
+        assert_eq!(weak.upgrade().expect("live"), object);
+    }
+}
```
jedel1043: Instead of adding tests that are kind of already covered by the WeakGc tests on boa_gc, you should convert these into a nice usage example inside `examples`.

jedel1043 (changes requested):
Mostly looks good, I just have a couple of nitpicks.

core/engine/src/object/weak.rs
GordonYuanyc: Thanks for the pointer. Upgrades and delegates to `JsObject`'s Debug now in commit 2137fbf7.

core/engine/src/object/weak.rs
GordonYuanyc: Thanks for the suggestion! Added an example in commit 3355d720.

GordonYuanyc:
Hi! I think both nitpicks were addressed in 2137fbf7 and 3355d720 and CI is green. The Waiting On Author label is still on the PR, so it may not be showing up in your review queue.

Also, do I need to rebase onto the latest main?

jedel1043 (approved):
LGTM, thank you for the contribution!

PR #5440, opened 2026-07-20
Update example GIF in README

### LINKED ISSUE #2703: Update the example gif in the README.md file
The gif we use in the README shows how cool the engine was 3 years ago. Now that things have notably improved, we should be able to create a new animation that shows the current status of the REPL, ideally.

sanjanachitta:
This Pull Request closes #2703.

It changes the following:

- Updates the example GIF in `README.md`.
- Replaces the old demo with a newly recorded GIF showing the current behavior of Boa.


jasonwilliams:
I’m not sure what this brings over the older gif?
If anything the older gif looks better because it’s zoomed in in the source code and I don’t need to squint. 

You’re also running the same thing as before

sanjanachitta:
Thanks for the feedback! I'll update the GIF with a closer zoom level to improve readability and make sure it better demonstrates the intended behavior.

sanjanachitta:
Thanks for the feedback. The intention was to update the GIF for the current Boa version while keeping the same demonstration. Would you prefer a different example that highlights newer behavior?

jedel1043:
Closing this since it doesn't seem to improve on the demo that we already have.

